PR #16, opened 2017-05-23
Colour variables combined from existing repos

igloosi:
All colour variables from existing repos combined into a colours partials for now and prefixed with .govuk.
Later we can look and splitting and amending


gemmaleigh:
👍

PR #17, opened 2017-05-23
Import-once mixin to ensure common imports are not duplicated

igloosi:
An import-once mixin ensures that exported styles from globals that are imported into individual components are not duplicated in sass compilation.

### DIFF
--- a/src/components/component-example/_component-example.scss
+++ b/src/components/component-example/_component-example.scss
@@ -1,0 +1,7 @@
+// dependencies go here, import-once at the end
+@import "import-once";
+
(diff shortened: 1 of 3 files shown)


fofr:
I noticed the commit has that horrible ellipses expansion.

@igloosi we have some guidance on commit messages/descriptions here: https://github.com/alphagov/styleguides/blob/master/git.md

(eg keeping commit summary below 50 chars and detailing why a change is made in the commit body)

igloosi:
reopened https://github.com/alphagov/govuk-frontend/pull/18

PR #20, opened 2017-05-24
Preview components HTML

igloosi:
Utilising gulp-inject to grab html from all components and
injecting into a index.html preview file that is served with
gulp-live-server. Any change to component html will rebuild the
preview file. Added to `gulp dev` task.


igloosi:
issues discovered, so closing

PR #21, opened 2017-05-25
Preview components

igloosi:
Utilising gulp-inject to grab html from all components and
injecting into a index.html preview file that is served with
gulp-live-server. Any change to component html will rebuild the
preview file. Added to gulp dev task.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -61,7 +61,7 @@ gulp.task('serve', () => {
 // Combines all html files in components into a single  file
 // ---------------------------------------
 gulp.task('combine:html', () => {
-  gulp.src(paths.dist + 'index.html')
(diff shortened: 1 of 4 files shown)


gulpfile.js
```diff
@@ -61,7 +61,7 @@ gulp.task('serve', () => {
 // Combines all html files in components into a single  file
 // ---------------------------------------
 gulp.task('combine:html', () => {
-  gulp.src(paths.dist + 'index.html')
+  gulp.src(paths.src + 'index.html')
```
igloosi: true :)

gemmaleigh:
We'll need to inject the css for component too, but this can happen in a separate PR.

PR #24, opened 2017-05-26
Add a draft button component

igloosi:
This PR:
Adds documentation for usage of the button component to the readme.
Adds a button component with two variants, start and disabled
Copies example HTML from GOV.UK elements
Creates new mixins for the font sizes used by this component
Copies media query and ie conditionals to global partials
Adds copy assets task to show start button image icon
Adds lint rules for pseudo element
Modifies preview template to add component wrappers and spacing between them

**Screenshot:**
screen shot 2017-05-26 at 12 45 09


igloosi:
thanks for adding those. happy with sass-mq as a way to manage breakpoints

gemmaleigh:
They list the conditionals.scss file in ["Inspired By"](https://github.com/sass-mq/sass-mq#inspired-by) so we come full circle...

PR #25, opened 2017-05-30
Add a phase banner component

gemmaleigh:
Copy HTML from govuk_elements and styles from the govuk_frontend_toolkit.
Add a README.
Create a new bold-16 mixin
Import the styles for the phase banner and phase tag components.

Also split out the styles for the phase-tag and create a new component for the phase tag.

phase-banner-tag


gemmaleigh:
Good call, I'll split those out too!

igloosi:
👍

PR #37, opened 2017-06-05
Add an input component

gemmaleigh:
Copy styles from govuk-elements for input type="text".

Create a text input modifier.

The name here is up for grabs, I went for input as it is the shortest, but we might want to use `form-input`. If so, do we then want to prefix all other form components with `form-`?

### DIFF
--- a/src/components/input/README.md
+++ b/src/components/input/README.md
@@ -1,0 +1,9 @@
+# Input
+
+A single-line text field.
+
+Make field widths proportional to the input they take.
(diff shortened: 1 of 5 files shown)


gemmaleigh:
Ok, I think we can make the decision about it being a simple component later - once we have a few more form components to compare with. It may be best to use `.govuk-c-text-input` without the need for the modifier, if the base class isn't reused elsewhere.

PR #43, opened 2017-06-06
Add textarea component

igloosi:
This PR:
Adds base textarea styles
Adds error textarea styles
Adds a code example including error message and label

**Screenshot**
screen shot 2017-06-06 at 11 41 40

### DIFF
--- a/src/components/textarea/_textarea.scss
+++ b/src/components/textarea/_textarea.scss
@@ -1,0 +1,29 @@
+@import "import-once";
+@import "../../globals/scss/colours";
+@import "../../globals/scss/font-face";
+@import "../../globals/scss/typography";
+@import "../../globals/scss/vars";
+@import "../label/label";
+@import "../error-message/error-message";
+
+@include exports("textarea") {
+  .govuk-c-textarea {
+    display: block;
+    -webkit-appearance: none;
+    border-radius: 0;
+    border: $govuk-border-width-form-element solid $govuk-text-colour;
+    box-sizing: border-box; // should this be global?
+    font-family: $govuk-font-stack;
+    -webkit-font-smoothing: antialiased;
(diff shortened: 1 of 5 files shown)


src/components/textarea/_textarea.scss
```diff
@@ -0,0 +1,29 @@
...
+    width: 100%;
+    @include core-19;
+  }
+
+  .govuk-c-textarea:focus {
+    outline: $govuk-outline-width-focus solid $govuk-focus-colour;
+  }
+
+  .govuk-c-textarea--error {
+    border: $govuk-border-width-mobile solid $govuk-error-colour;
```
gemmaleigh: Think we should change this to `$govuk-border-width-error`? as the widths don't change here for mobile or tablet, it's only the width of the error summary border that uses a narrower width.

src/components/textarea/textarea.html
```diff
@@ -0,0 +1,12 @@
+<label class="govuk-c-label" for="govuk-c-textarea-a">
+  Why can't you provide a National Insurance number?
+</label>
+<textarea class="govuk-c-textarea" name="govuk-c-textarea-a" id="govuk-c-textarea-a" rows="5"></textarea>
+
+<label class="govuk-c-label" for="govuk-c-textarea-b">
+  hy can't you provide a National Insurance number?
```
gemmaleigh: Missing 'W'

gemmaleigh:
Looks ace, works really nicely combining the label and error components too! 💯 

I'll revise the `.form-input` component to use the same.

src/components/textarea/_textarea.scss
igloosi: ok will change variable name and add it to vars

PR #50, opened 2017-06-09
Add create package task

igloosi:
This PR:

Adds a gulp task that copies components to packages/ folder and amends scss import paths to match the npm package paths


gemmaleigh:
👍

PR #51, opened 2017-06-09
Prepare the packages directory for lerna boostrap

gemmaleigh:
Commit each of the components to `/packages`.

Add a `package.json` file for each component, include components dependencies on other components.

This means that `npm run lerna:bootstrap` can now be run and will output:
screen shot 2017-06-09 at 10 16 52

and `npm run lerna:ls`
screen shot 2017-06-09 at 10 17 09

Other things that'll need fixing after this PR:

* govuk-frontend and govuk-frontend-ie.scss are being copied into `/packages/globals`, they shouldn't end up in this package


gemmaleigh:
Thanks @igloosi 😄

igloosi:
shall i merge this in?

PR #59, opened 2017-06-12
Add select box component

igloosi:
This PR:
Adds select box component markup
Adds select box component styles
Adds select box readme

**Screenshot**
screen shot 2017-06-12 at 10 22 24

### DIFF
--- a/src/components/details/README.md
+++ b/src/components/details/README.md
@@ -1,0 +1,3 @@
+# Component example
+
+Description of component.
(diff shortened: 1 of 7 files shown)


src/components/select-box/_select-box.scss
```diff
@@ -0,0 +1,28 @@
+@import "../../globals/scss/import-once";
+@import "../../globals/scss/colours";
+@import "../../globals/scss/typography";
+
+@include exports("select-box") {
+  .govuk-c-select-box {
+    box-sizing: border-box; // should this be global?
+    font-family: $govuk-font-stack;
+    -webkit-font-smoothing: antialiased;
```
gemmaleigh: can we use the font-smoothing mixin here instead please? - @include font-smoothing

PR #61, opened 2017-06-12
Revise documentation

gemmaleigh:
Add documentation to explain directory structure.
Remove placeholder HTML docs.
Add links to license and contribution guidelines.


nickcolley:
Is it possible to have the main bundle without '/all'?

gemmaleigh:
@nickcolley I'm not clear on what you mean. 

You can pick and choose components - so you'd need `@govuk-frontend/global` and then `@govuk-frontend/component-name` - assuming you're using npm to consume them.

Or you could include `@govuk-frontend/all`, which would include all components as a dependency.

Or, if you're not using npm - you can consume the components via `/dist`, we've not quite figured out how that will work. We'll probably provide a minified stylesheet and js file including all the components.

nickcolley:
As in would 'npm install @govuk-frontend' be more straight forward? The '/all' is cool because it indicates there's other builds though

gemmaleigh:
We could, absolutely. None of this is set in stone, the only downside I can think of is wouldn't be able to use the `@govuk-frontend` scope, so it could be named like any other module, say `npm install --save govuk-frontend`. 

We'd stuck with the convention of using the @govuk-frontend scope for everything related to govuk-frontend, it requires us to then set a package name https://docs.npmjs.com/misc/scope but it seemed like a good way to keep all the modules together.

PR #71, opened 2017-06-13
Remove rem and pixel fallback, instead default to px

gemmaleigh:
It makes me sad to remove this, but having tried out including the breadcrumb component in the govuk-elements application, the font sizes are too small.

In our new world of isolated components - we don't have any control over the :root or html element of the document.

The font-size on the html element - set by govuk_template to 62.5%, is affecting the font size of the breadcrumb component, this component, which previously used font-size 1rem = 16px, instead is rendered at 10px.

In a future where we are no longer using govuk_template, we can revisit this decision. However, if our components are to exist in codebases we don't have control over - where the font size on the root or html element could vary, we will have to use pixels to prevent unexpected differences in the font-sizes of components.

### DIFF
--- a/src/globals/scss/_typography.scss
+++ b/src/globals/scss/_typography.scss
@@ -1,9 +1,3 @@
-$govuk-base-font-size: 16px !default;
-
-@function rem($px) {
-  @return ($px / $govuk-base-font-size) * 1rem;
-}
-
 // New Transport Light font family
 $govuk-nta-light: "nta", Arial, sans-serif;
 $govuk-nta-light-tabular: "ntatabularnumbers", $govuk-nta-light;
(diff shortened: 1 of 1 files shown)


nickcolley:
Can we add a temporary global override? This also makes me 😿

gemmaleigh:
I'm not sure how that would work... 💭  
If you've got an example, that would be handy!

igloosi:
@nickcolley  if we pulled in a global reset or use :root then we'd be affecting unknown stuff they might have setup with rem. PX is the safest option 😢

nickcolley:
I was thinking if our components at the top level are `rem/px` based and internals are `em` based, we'd be able to turn rems on and off depending on what project you're working in? 🤔

gemmaleigh:
@nickcolley let's come back to this when we start work on typography. For now it means we can test our packages of components and they will render as expected when used alongside govuk_template.

PR #81, opened 2017-06-20
Update to build & publish steps

igloosi:
This PR:
- updates gulp tasks to include `gulp build:packages`,  `gulp build:dist` and  `gulp build:demo`
to support the build & publish process
- updates package.json with new dependcies
- amend documentation with new steps
- adds `component-view` and `component-list` template to allow listing of all components and previewing
- injects component example html into the readme code example so that they'll always be in sync
- amends `config.paths` file with new paths
- amends `.gitignore` file with new paths
- appends package version of the 'all` package to minified js and css files
- updates VERSION.txt file with latest package version of the 'all` package in `dist` and `demo` folders
- adds `gulp serve:demo` to preview the demo folder

**Screenshots**
screen shot 2017-06-20 at 14 30 29
screen shot 2017-06-20 at 14 30 36
screen shot 2017-06-20 at 14 30 47


igloosi:
removed unwanted js files.
i will see how i can configure uglyfy, cssnano and template warp to respect/add new lines

gemmaleigh:
This is great, thanks @igloosi!

A couple of things once this has been merged - the order of components in index.html needs to be the same each time `gulp build:dist is run`, as this is showing up as a change (because of the ordering) - when no changes have been made to components.

We need to add an EOF to each of the generated files in dist and commit those, as this is also showing as a change to govuk-frontend.min.js.

PR #82, opened 2017-06-21
List desired components in alphabetical order

igloosi:
This PR:
- Removes _component-example from preview, dist and demo folders
- Ensures that components are always listed in alphabetical order (prevents making changes, when there are none - counteract gulp streams)

**Screenshot**
screen shot 2017-06-21 at 13 22 52


gemmaleigh:
This works nicely! 💯

PR #83, opened 2017-06-22
Copy existing grid styles to a grid component

gemmaleigh:
I haven't made any changes to these styles, other than to rename our existing grid placeholders and instead use mixins.

The grid component has the grid wrapper component as a dependency (as the grid wrapper must wrap a grid component). Alternatively, the grid-wrapper mixin can be used. 

Also update the cookie-banner component to use the grid-wrapper mixin.

### DIFF
--- a/src/components/cookie-banner/_cookie-banner.scss
+++ b/src/components/cookie-banner/_cookie-banner.scss
@@ -1,6 +1,7 @@
 @import "../../globals/scss/import-once";
 @import "../../globals/scss/colours";
 @import "../../globals/scss/typography";
+@import "../site-width-container/site-width-container";
 
 @include exports("cookie-banner") {
   .govuk-c-cookie-banner {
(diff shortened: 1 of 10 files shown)


nickcolley:
Do these need to be bundled together? Do they rely on each other to work?

If you look at frameworks like Bootstrap they have 'containers' which you can put a grid in if you please but otherwise can exist separate from the grid.

gemmaleigh:
Yup, it's the same here.

Bootstrap states: rows must be placed within a `.container` (fixed-width) or `.container-fluid` (full-width) for proper alignment and padding.

Here `.grid-wrapper` can exist on it's own (in which case maybe it needs a more generic name) and is similar to Bootstrap's `.container`. 

The grid has grid-wrapper as a dependency, as grid rows should sit inside a wrapper - in the same way as Bootstrap requires this.

nickcolley:
I didn't mean to suggest they depend on each other in the codebase, but rather the naming suggests they are dependant on each other.

I'll see myself out sorry :P 🚪 🏃

gemmaleigh:
No no! You're right, it needs a better name. As `.grid-wrapper` can be used on its own.

I'm going to revert and use `.site-width-container`. Gah, naming things.

PR #89, opened 2017-06-26
Enable Heroku review apps

gemmaleigh:
This PR:
- adds ExpressJS as a devDependency
- adds a new gulp task `review` to prepare assets for review apps
- adds a new npm script 'heroku' to run `gulp review && node review.js`
- amends the Procfile for Heroku to run the heroku npm script
- updates package-lock.json
- sets environment variables for Heroku review apps in app.json - these include setting the buildpack, ensuring that NPM_CONFIG_PRODUCTION is set to false so devDependencies are installed and gulp is available
- checks for environment variables in review.js so that the same script can be used to deploy both review apps and the demo folder, depending on the app the script is deployed to

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,1 +1,1 @@
-web: vendor/bin/heroku-php-apache2 demo/
+web: node review.js
--- a/Procfile
+++ b/Procfile
@@ -1,1 +1,1 @@
-web: node review.js
+web: npm run heroku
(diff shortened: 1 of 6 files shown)


gemmaleigh:
This will also mean we can deploy demo to: https://govuk-frontend.herokuapp.com/ (that should now be working) we may need to check we've run the correct gulp tasks to ensure demo is up-to-date, but that's for another PR.

PR #90, opened 2017-06-27
Use Travis build stages

gemmaleigh:
Split deployment via Travis into two stages:

1) Deploying to the review app
This will happen after a successful build on master.

2) Deploying to production 
This will only happen on tagged commits (after the
lerna publish step has been run).

https://travis-ci.org/alphagov/govuk-frontend/builds/247417470

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,11 +6,25 @@ notifications:
 sudo: false
 script:
 - npm test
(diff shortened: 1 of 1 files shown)


gemmaleigh:
I'll wait and see if Travis goes green!

PR #95, opened 2017-06-27
Add Prism.js and GOV.UK styles

igloosi:
This PR
Adds prism.js from cdn (quicker than npm)
Adds gov.uk list and site-width-container styles

Screenshot:

screen shot 2017-06-27 at 16 55 56
screen shot 2017-06-27 at 17 03 38

### DIFF
--- a/src/component-list-template.html
+++ b/src/component-list-template.html
@@ -14,12 +14,14 @@
   }
   </style>
 </head>
-<body>
+<body class="govuk-c-site-width-container">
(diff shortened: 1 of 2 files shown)


gemmaleigh:
Really nice! https://govuk-frontend-review-pr-95.herokuapp.com/
At some point we probably want to update this to use the prism theme we're using on govuk-elements, as it has a better colour contrast, but this looks great for now 👍 .

PR #101, opened 2017-07-03
Global icons as a package

igloosi:
This PR:
Updates gulp tasks that split `globals/icons` into a separate folder for npm package, demo and dist folders
Adds package.json for it in packages/icons
Replaces references from`/images/` to `/icons/` in scss files in required files
Adds @govuk-frontend/icons as a dependency to button and breadcrumb
Removes old `/images/` folders


gemmaleigh:
I think this needs a rebase against master to remove those extra commits.

PR #106, opened 2017-07-04
Insert newLine to files

igloosi:
Problem: 
Uglify and template wrap remove new lines from files and we want them back

This PR:
Adds gulp-eol as dependency
Pipes it in relevant tasks

🎉


gemmaleigh:
🎆

package.json
```diff
@@ -48,6 +48,7 @@
     "standard": "^10.0.2"
   },
   "dependencies": {
+    "gulp-eol": "^0.1.2",
```
gemmaleigh: should this be a devDependency?

igloosi: fair point, updated

PR #112, opened 2017-07-06
Ensure Travis is using npm5 for builds

gemmaleigh:
We can't be sure that npm5 is installed globally when dependencies are installed - in which case the package-lock.json file won't be updated. 

Use this check to ensure that if dependencies are added to
package.json, that the package-lock.json file has also been updated.

Also ensure npm5 is installed globally for Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,9 @@ jobs:
   include:
     - stage: Run npm test
       script:
+      # Ensure Travis aborts without running tests if the package-lock.json file needs updating
+      - set -e
+      - ./check-package-lock.sh
       - npm test
     - stage: Deploy to the review app
       script: echo "Deploying to govuk-frontend-review.herokuapp.com ..."
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,12 @@
 language: node_js
 node_js:
-- 6.0
+- 7
 notifications:
   email: false
 sudo: false
 before_install:
-  openssl aes-256-cbc -K $encrypted_ca4a154c2e6b_key -iv $encrypted_ca4a154c2e6b_iv -in config/tenon.json.enc -out config/tenon.json -d
+  - if [[ `npm -v | cut -d. -f1` -lt 5 ]]; then npm i -g npm@latest; fi
+  - openssl aes-256-cbc -K $encrypted_ca4a154c2e6b_key -iv $encrypted_ca4a154c2e6b_iv -in config/tenon.json.enc -out config/tenon.json -d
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -1,17 +1,21 @@
 language: node_js
 node_js:
-- 6.0
+- 7
 notifications:
   email: false
 sudo: false
 before_install:
-  openssl aes-256-cbc -K $encrypted_ca4a154c2e6b_key -iv $encrypted_ca4a154c2e6b_iv -in config/tenon.json.enc -out config/tenon.json -d
+  - if [[ `npm -v` != 5* ]]; then npm i -g npm@5.0.3; fi
```
tvararu: Should we try `npm i -g npm@latest`? If this does eventually end up biting us we can lock to a particular version.

check-package-lock.sh
```diff
@@ -0,0 +1,10 @@
+# Check to see whether the package-lock.json file needs updating.
+
+echo "checking package-lock.json for changes"
+
+if git diff --exit-code --quiet package-lock.json; then
+  echo "no changes have been made to package-lock.json"
+else
+  echo "package-lock.json needs updating, run npm install and commit"
```
tvararu: You need to run it on a new version of npm, so maybe this is better:

```
echo "package-lock.json needs updating, run npm install on npm@5+ and commit"
```

gemmaleigh: good point @tvararu, will update!

.travis.yml
```diff
@@ -1,17 +1,21 @@
 language: node_js
 node_js:
-- 6.0
+- 7
 notifications:
   email: false
 sudo: false
 before_install:
-  openssl aes-256-cbc -K $encrypted_ca4a154c2e6b_key -iv $encrypted_ca4a154c2e6b_iv -in config/tenon.json.enc -out config/tenon.json -d
+  - if [[ `npm -v` != 5* ]]; then npm i -g npm@latest; fi
```
tvararu: Sorry, didn't read the conditional when I suggested the `@latest` earlier; to make sure we don't reinstall `@latest` constantly when `npm@6` comes out we can do:

```bash
- if [[ `npm -v | cut -d. -f1` -lt 5 ]]; then npm i -g npm@latest; fi
```

`cut` will split the version number by the `.` delimeter and then give us the first field, which is the major number in the version. If this is less than 5, install the latest version of `npm`.

gemmaleigh:
This one needs merging first, then the package-lock.json file will need updating for open PRs.

PR #114, opened 2017-07-06
Fix string replace for .min.css | js links

igloosi:
This PR:
Fixes where we append version to static assets' names and also include external minifies files. Now we only append version number and correct suffix to out files

**Before:**
screen shot 2017-07-06 at 13 46 21

**After:**
screen shot 2017-07-06 at 13 45 23


gemmaleigh:
👍  good spot, I missed this!

PR #117, opened 2017-07-06
Remove pattern and min max attributes from date component

gemmaleigh:
Use input type=“number” to trigger a numeric keyboard.

The pattern attribute applies when the value of the type attribute is
text, search, tel, url, email, or password, otherwise it is ignored. 
So for input type="number" this will be ignored.

Also remove min and max, as these will not stop a user entering a
number manually outside of this range - this only works for the spinner
control buttons, which we aren’t displaying for this component.

### DIFF
--- a/src/components/date/date.html
+++ b/src/components/date/date.html
@@ -1,15 +1,15 @@
 <div class="govuk-c-date">
   <div class="govuk-c-date__item govuk-c-date__item--day">
     <label class="govuk-c-label govuk-c-date__label" for="dob-day">Day</label>
-    <input class="govuk-c-input govuk-c-date__input" id="dob-day" name="dob-day" type="number" pattern="[0-9]*" min="0" max="31">
+    <input class="govuk-c-input govuk-c-date__input" id="dob-day" name="dob-day" type="number">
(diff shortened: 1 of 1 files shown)


gemmaleigh:
I've asked @hannalaakso to investigate a better solution to this! We should probably update govuk-elements as well, since this is copied from there.

edwardhorsford:
We've also asked @hannalaakso to look at this for generic numeric inputs.

edwardhorsford:
Min and max also apply if you're using arrow keys to increment and detriment a number. This *might* be useful for dates and things like quantity fields, but not at all for things like passport numbers.

I think *avoiding* incrementing with arrow keys was why we used `type=text` for dates of birth on passports.

hannalaakso:
Re: what I discussed with Ed, we tend to turn off the browser native validation so the form should have "novalidate" added to it.

hannalaakso:
We should in fact use the `pattern` attribute on number inputs as it brings up the numeric keypad on iOS. 

No further action here as this markup no longer exists in GOV.UK Frontend - this is just a reference for best practise.

PR #118, opened 2017-07-06
Only show sassmq breakpoints on preview not dist or packages/

igloosi:
This PR:
Includes the `$mq-show-breakpoints: (mobile, tablet, desktop);` variable (that adds breakpoint content into body styles) only on preview so that these styles don't get compiled in dist and packages


demo/globals/scss/_media-queries.scss
```diff
@@ -55,10 +55,13 @@ $mq-breakpoints: (
 // (i.e. when $mq-responsive is set to 'false').
 $mq-static-breakpoint: desktop;
 
+// sass-lint:disable no-css-comments
+// for dev and demo only
```
gemmaleigh: for dev and preview only?

gemmaleigh:
Tested using `gulp serve:demo` (doesn't show mq breakpoints) and `gulp serve:preview` (does show mq breakpoints) 

👍 

Tiny change to the comment to reflect this and this is good to merge.

PR #121, opened 2017-07-06
Legal text component

gemmaleigh:
Preview:

screen shot 2017-07-06 at 17 17 17


igloosi:
design q: how different is the ! in the icon to the one in tabular or normal transport? if not much, we could just use that instead of .png or .svg

fofr:
Would this be better named: `legal-consequence`?

gemmaleigh:
The original version:
gov uk frontend - using icon

Using transport for the ! and the large step icon circle (38px x 38px):
gov uk frontend - transport

nickcolley:
Nice!

* How's this look cross browser?
* Does a assistive tech read the bang out?
* What happens when you zoom?

As a side note, what is the scope of components that should be on GOV.UK Frontend?

Is there a criteria that means it will be useful and reused across Government, my gut with this component would be that it seems pretty specific to GOV.UK Publishing?

gemmaleigh:
Hi Nick,

I'll need to take screenshots for the first and add `aria-hidden="true"` for the second.

GOV.UK Frontend's initial scope is to include components for all the examples we provide as part of GOV.UK elements, here's the legal text example: 
https://govuk-elements.herokuapp.com/typography/#typography-legal-text

This is used by services, often for declaration pages - for example, I confirm that the information I have given is true... and this component will be used underneath to reinforce that the declaration above has a legal consequence.

gemmaleigh:
Voiceover reads "Warning You can be fined up to £5,000 if you do not register".

gemmaleigh:
It looks like it should, apart from in IE8 which doesn't support border-radius so we get a big black square rather than a circle.

IE8:
win7_ie_8 0

IE11:
win8 1_ie_11 0

gemmaleigh:
Let's go with this for now. If we decide when we start the work on icons, that we would like to use an inline SVG for this icon to be consistent with icons used elsewhere - then we can change it at that point.

PR #127, opened 2017-07-10
Keep border width the same for error messages

### LINKED ISSUE #119: Should inputs with active error state be adjusted to remain the same size as inputs without an error?
**Potential problem:**
When an input has an error it's border is increased from 2px to 4px. This means that input will become physically bigger than it's previous state.

When this input is in-line or part of a tight lockup with another element the component parts will become mis-aligned.

Should we adjust the padding of the input error state to compensate for the increased border?

For example 

```
.govuk-c-input {
        padding: 5px 4px 4px;
}
.govuk-c-input--error {
        padding: 3px 2px 2px;
}
```

Before | After
screen shot 2017-07-06 at 16 15 39

Is there an alternative way of doing this with box-sizing?

gemmaleigh:
This avoids “jumping” form inputs and the difference in height between
inputs, when they sit next to each other (for example, for date of
birth).

@dashouse can you let me know if this fixes #119.

Examples are here:

Date of birth:
https://govuk-frontend-review-pr-127.herokuapp.com/components/date/date.html

Input:
https://govuk-frontend-review-pr-127.herokuapp.com/components/input/input.html

Textarea:
https://govuk-frontend-review-pr-127.herokuapp.com/components/textarea/textarea.html

### DIFF
--- a/src/components/input/_input.scss
+++ b/src/components/input/_input.scss
@@ -9,11 +9,11 @@
     box-sizing: border-box; // should this be global?
     width: 100%;
     padding: 5px 4px 4px;
-
     // setting any background-color makes text invisible when changing colours to dark backgrounds in Firefox (https://bugzilla.mozilla.org/show_bug.cgi?id=1335476)
     // as background-color and color need to always be set together, color should not be set either
     border: $govuk-border-width-form-element solid $govuk-text-colour;
     border-radius: 0;
+
 
     font-family: $govuk-font-stack;
     @include font-smoothing;
@@ -38,6 +38,7 @@
   }
 
   .govuk-c-input--error {
+    padding: 3px 2px 2px;
     border: $govuk-border-width-error solid $govuk-error-colour;
   }
 
--- a/src/components/textarea/_textarea.scss
+++ b/src/components/textarea/_textarea.scss
@@ -12,6 +12,7 @@
     display: block;
 
     width: 100%;
+    padding: 5px 4px 4px;
 
     border: $govuk-border-width-form-element solid $govuk-text-colour;
     border-radius: 0;
@@ -30,6 +31,7 @@
   }
 
   .govuk-c-textarea--error {
+    padding: 3px 2px 2px;
     border: $govuk-border-width-error solid $govuk-error-colour;
   }
 }


nickcolley:
In context how does this look?

We have consistent thickness here for example

screen shot 2017-07-10 at 16 10 26

gemmaleigh:
@edwardhorsford this is consistent with the styles for the character count prototype - are you happy with these changes?

gemmaleigh:
@nickcolley hold on a minute, I'll create a screenshot so we can compare the two.

gemmaleigh:
errors and validation gov uk elements

gemmaleigh:
Date of birth (from GOV.UK elements):


errors and validation gov uk elements

after example date form elements gov uk elements

nickcolley:
With that example, if you use `display: inline-block` rather than `float: left` (`display: block`) it'll be vertically aligned

screen shot 2017-07-10 at 16 38 32

fofr:
@gemmaleigh Is there sufficient difference in the last screenshot to indicate which of the date fields has an error? eg Viewing in black and white, high enough contrast difference, etc.

nickcolley:
screen shot 2017-07-10 at 16 41 11

gemmaleigh:
@fofr @nickcolley, you're both quite right - it isn't at all distinct now.

edwardhorsford:
I think the bigger border helps highlight the field - the issue is things getting out of alignment / the overall hight changing when it's added.

Is there a way (more / hidden borders) of getting them to line up / not move whilst keeping the existing widths? I'd suggest outline, but I know we're already using that. Second outline!

nickcolley:
@edwardhorsford yeah, if we use `display: inline-block` rather than floats in the date pattern we can have them centered as seen above.

edwardhorsford:
@nickcolley does that still change the overall height of the item (not including the extra line added)?

The issue we had on character count is that if you do inline/live styling, the overall height changes. I presume Pay will also have had this issue.

nickcolley:
Ah, I imagine so yes. Will investigate.

nickcolley:
screen shot 2017-07-10 at 17 39 10
screen shot 2017-07-10 at 17 39 24

nickcolley:
One alternative could be to do `box-shadow: 0 0 0 2px $error-warning-color inset;` ?

dashouse:
I like the border + box shadow (visually combining to create the 4px visual border)

Out of interest, in the original post I was just thinking about dropping the padding by 2px on the error class when the 2px is added to the border. Because the adding and subtracting 2px is equal the text remains in the same place. What was the drawback of this? Just that it's a bit weird?

```
.govuk-c-input {
        padding: 5px 4px 4px;
}
.govuk-c-input--error {
        padding: 3px 2px 2px;
}
```

nickcolley:
@dashouse that could be the nicest way if it renders the same cross-browser.

When adding `padding-top: 3px; padding-bottom: 2px;` to the error class only I got the following

screen shot 2017-07-11 at 10 09 46

igloosi:
⚡️  HEAD OF FRONTEND ⚡️ : do we mind that IE8 doesnt support box-shadow?

gemmaleigh:
We're not using box-shadow... will this solution (adjusting the padding) not work?
Probably need a bunch of browserstack screenshots to accompany this one.

igloosi:
adjust padding will work just fine, though feels a bit dirty

gemmaleigh:
Closing for now.

PR #128, opened 2017-07-10
Fix #518: stray line on checkbox on IE11

### LINKED ISSUE #518: Proposal: add a 'positive' colour
Copied over from [GOV.UK Frontend toolkit](https://github.com/alphagov/govuk_frontend_toolkit/issues/386)

There are cases where it's useful to have a colour with a semantic meaning of 'positive'.

We already have error_colour which is a good red for 'negative' - but we don't have a counterpart.

We could use an existing colour, or make a new colour. Either way, a semantic SASS name would be good.

The closest colour would be button_colour. An alternate would be #008800 which is the green used by petitions. It's warmer, which feels more positive to me.

image

alex-ju:
IE sometimes miscalculates the border width of a rotated element resulting in a thin border as shown in #518. Setting a transparent border will fix the issue without affecting semantics or rendering on any other browser.

### DIFF
--- a/src/components/checkbox/_checkbox.scss
+++ b/src/components/checkbox/_checkbox.scss
@@ -85,6 +85,7 @@
     transform: rotate(-45deg);
     border: solid;
     border-width: 0 0 5px 5px;
+    border-top-color: transparent;
 
     opacity: 0;
 


fofr:
Is there a before and after screenshot?

fofr:
Presumably this is a default IE11 style that we haven't normalised/reset?

src/components/checkbox/_checkbox.scss
```diff
@@ -85,6 +85,7 @@
     transform: rotate(-45deg);
     border: solid;
     border-width: 0 0 5px 5px;
+    border-top-color: transparent;
```
fofr: Feels like this line is worthy of a comment.

alex-ju:
Misplaced PR. It was suppose to be on [govuk_elements](https://github.com/alphagov/govuk_elements/pull/520). Screenshots added.

PR #132, opened 2017-07-14
Add aria-disabled to buttons to support older versions of JAWS

gemmaleigh:
JAWS 15 and below don't recognise the disabled attribute on buttons.

This adds the aria-disabled attribute to make disabled buttons
compatible with older screen readers.

See [govuk_elements #532](https://github.com/alphagov/govuk_elements/pull/532).


gemmaleigh:
No worries, Anika raised it as a pull request on element last week!

PR #137, opened 2017-07-17
Add global link styles

igloosi:
This PR:
Adds global link styles and it's states' styling
Allows rgba in lint rules
Adds link print styles

### DIFF
--- a/src/globals/scss/_colours.scss
+++ b/src/globals/scss/_colours.scss
@@ -56,6 +56,7 @@ $govuk-grey-2: #bfc1c3;
 $govuk-grey-3: #dee0e2;
 $govuk-grey-4: #f8f8f8;
 $govuk-white: #ffffff;
+$govuk-black: #000000;
 
 // Semantic colour names
 $govuk-link-colour: $govuk-blue;
--- a/src/globals/scss/_colours.scss
+++ b/src/globals/scss/_colours.scss
@@ -56,7 +56,6 @@ $govuk-grey-2: #bfc1c3;
 $govuk-grey-3: #dee0e2;
 $govuk-grey-4: #f8f8f8;
 $govuk-white: #ffffff;
-$govuk-black: #000000;
 
 // Semantic colour names
 $govuk-link-colour: $govuk-blue;
--- a/src/globals/scss/_links.scss
+++ b/src/globals/scss/_links.scss
@@ -1,7 +1,8 @@
 @import "import-once";
 @import "colours";
+@import "media-queries";
 
-@include exports("links") {
+@include exports("global-link-styles") {
   a:link {
     color: $govuk-link-colour;
   }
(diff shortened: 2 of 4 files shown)


gemmaleigh:
We already have $govuk-black https://github.com/alphagov/govuk-frontend/blob/a3ae2cbb5b22385128ff3d1ea94a203ef08fe841/src/globals/scss/_colours.scss#L53

src/globals/scss/_colours.scss
```diff
@@ -56,6 +56,7 @@ $govuk-grey-2: #bfc1c3;
 $govuk-grey-3: #dee0e2;
 $govuk-grey-4: #f8f8f8;
 $govuk-white: #ffffff;
+$govuk-black: #000000;
```
fofr: Is this intentionally different to `#0b0c0c` https://github.com/alphagov/govuk_frontend_toolkit/blob/master/stylesheets/colours/_palette.scss#L50

src/globals/scss/_links.scss
```diff
@@ -0,0 +1,29 @@
...
+    color: $govuk-link-active-colour;
+  }
+
+  a {
+    -webkit-tap-highlight-color: rgba($govuk-black, .3);
+  }
+
+  a:focus {
+    outline: 3px solid $govuk-focus-colour;
+    background-color: $govuk-focus-colour;
```
fofr: Should the colour be set to black on focus because of poor contrast between focus colour and link/visited colours. Vaguely recall us doing this somewhere already?

gemmaleigh: @fofr I think it was for the skiplink, so the styles will be in govuk_template.

igloosi: not familiar with swing that black. will check tmrw

igloosi: not familiar with switch that black. will check tmrw
not sure how I've missed that black ;(

gemmaleigh: No worries! The skiplink styles are here: https://github.com/alphagov/govuk_template/blob/master/source/assets/stylesheets/_accessibility.scss#L70. We can take a look tomorrow.

fofr:
Aside: Where should we include print styles for links? Eg showing the linked-to URL in the print view.

src/globals/scss/_links.scss
igloosi: i wonder if we're to  add skiplink styles in there, where to we add the skiplink container styles?

igloosi:
@fofr good call. i suggest here.

gemmaleigh:
If you remove the skiplink styles from this PR, I'll add them to #138.

We don't need to import links at the top of govuk-frontend.scss, those styles should be included in each component - in the same way as we do for other globals.

igloosi:
ok.

src/globals/scss/_links.scss
igloosi: @fofr dave mentioned it was brought up at the SDA show and tell (register to vote) accessibility audit that blue on yellow wasn't high enough contrast

PR #138, opened 2017-07-17
Replace the link back component with a more generic link component

gemmaleigh:
Add link variants - back, muted, download and skip.

add link variants - gov uk frontend

https://govuk-frontend-review-pr-138.herokuapp.com/components/link/link.html

To do before this can be merged:
- [x] Include global link styles from #137
- [x] Add skiplink variant


fofr:
Should these images be updated to be SVG?

gemmaleigh:
@fofr yes, we intend to replace all icons with inline SVG, with a png fallback. It's a bigger piece of work which we plan on doing later. 

https://github.com/alphagov/govuk-frontend/projects/2#card-3207524

gemmaleigh:
This PR needs updating to include the global link styles in #137, which will fix the colour of the download link and is also missing the skiplink variant.

PR #139, opened 2017-07-18
Prev/next pagination

igloosi:
This PR:
Adds prev/next (big arrow) pagination from https://github.com/alphagov/static/pull/1051
Adds meta viewport to the view template
Adds used typography mixin

Screnshots:
screen shot 2017-07-18 at 14 53 05
screen shot 2017-07-18 at 14 52 39
screen shot 2017-07-18 at 14 49 42

### DIFF
--- a/src/components/pagination/pagination.html
+++ b/src/components/pagination/pagination.html
@@ -1,0 +1,26 @@
+<nav class="govuk-c-pagination" role="navigation" aria-label="Pagination">
+  <ul class="govuk-c-pagination__list">
+      <li class="govuk-c-pagination__item govuk-c-pagination__item--previous">
+        <a class="govuk-c-pagination__link" href="#" rel="prev">
+          <span class="govuk-c-pagination__link-title">
+            <svg class="govuk-c-pagination__link-icon" xmlns="http://www.w3.org/2000/svg" height="13" width="17" viewBox="0 0 17 13">
+              <path fill="currentColor" d="m6.5938-0.0078125-6.7266 6.7266 6.7441 6.4062 1.377-1.449-4.1856-3.9768h12.896v-2h-12.984l4.2931-4.293-1.414-1.414z"></path>
+            </svg>
+            Previous
+          </span>
+          <span class="govuk-c-pagination__link-label">Overview</span>
+        </a>
+      </li>
+      <li class="govuk-c-pagination__item govuk-c-pagination__item--next">
+        <a class="govuk-c-pagination__link" href="#" rel="next">
+          <span class="govuk-c-pagination__link-title">
+            Next
+            <svg class="govuk-c-pagination__link-icon" xmlns="http://www.w3.org/2000/svg" height="13" width="17" viewBox="0 0 17 13">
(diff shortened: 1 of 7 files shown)


nickcolley:
What led to including this in GOV.UK Frontend? This is a GOV.UK Publishing Component, but there's no explanation why this has been promoted to GOV.UK Frontend.

src/components/pagination/pagination.html
```diff
@@ -0,0 +1,26 @@
+<nav class="govuk-c-pagination" role="navigation" aria-label="Pagination">
+  <ul class="govuk-c-pagination__list">
+      <li class="govuk-c-pagination__item govuk-c-pagination__item--previous">
```
fofr: There's a stray indentation here.

src/components/pagination/pagination.html
```diff
@@ -0,0 +1,26 @@
...
+        <a class="govuk-c-pagination__link" href="#" rel="prev">
+          <span class="govuk-c-pagination__link-title">
+            <svg class="govuk-c-pagination__link-icon" xmlns="http://www.w3.org/2000/svg" height="13" width="17" viewBox="0 0 17 13">
```
fofr: If you wanted to update this SVG, how would you do it?
As users will be copying/pasting this template, does that mean they'd have to copy a new SVG into their markup?

src/globals/scss/_typography.scss
```diff
@@ -65,6 +64,11 @@ $govuk-font-stack-tabular: $govuk-nta-light-tabular !default;
   line-height: (40 / 36);
 }
 
+@mixin govuk-core-27 {
```
fofr: Is the availability of this new font-size documented somehow?

fofr:
Following on from @nickcolley, what's the criteria for adding a component here?

src/components/pagination/pagination.html
nickcolley: It'd just mean updating the SVG would be a breaking change since GOV.UK Frontend can't hide the internals of it's markup.

src/globals/scss/_typography.scss
igloosi: no, not yet. still assessing which typography styles are being used before surfacing them

src/components/pagination/pagination.html
igloosi: support for external source svgs is not good so if the icon changes, yeah it's a breaking change

gemmaleigh:
@fofr we're working to combine what we have at the moment in govuk_template, govuk_frontend_toolkit and govuk_elements into a single repo. 

We're mostly creating components so we are able to build the equivalent code examples that we show as part of govuk_elements, but where there are components that we don't currently have in govuk_elements that are likely to be useful to service teams, we are also adding those. 

There isn't a way for external teams to consume GOV.UK Publishing Components and where there is a component we'd like to share, it will be included here.

This component is on a list of proposed components for the GOV.UK Design System created by @timpaul.

igloosi:
Updated with new spacing scale variables

gemmaleigh:
👍

PR #140, opened 2017-07-18
Order partials alphabetically for all components

gemmaleigh:
Reorder the list of imported partials for each component.
Order the partials alphabetically, after the import once partial.

The `src/global/_colours.scss` has been moved into `src/globals/vars/_colours`, as this is file contains vars and is now imported by the global vars file.

Rename the focus width variable and update where it is used.

Namespace the clearfix mixin, where this is used - update the mixin name.

Use renamed helpers partials, e.g. helpers-ie, or helpers-url.

Use import-once to wrap utility classes - if a utility has a concrete class, ensure this is wrapped using the import once mixin and the export name matches the filename.


gemmaleigh:
Can we merge this one first? Then the other two open PRs are gonna need rebasing and double checking that the imports follow the same conventions.

PR #142, opened 2017-07-20
Set a spacing scale and export spacing utility classes

gemmaleigh:
- Add test values to a spacing scale
- Set spacing properties for margin and padding
- Set spacing directions - top, right, bottom, left
- Set a spacing scale, using spacing scale vars. Include 0 to reset
paddings and margins.

Export classes in the format: `.govuk-u-mb-2` where 2 is the spacing scale, from 0-6.

Note: we're testing this to see how these classes might work to provide spacing around components, and how these spacing vars can be combined with our typography styles. This is *very* likely to change. 

Todo: 

- [x] Ensure if spacing is imported by each component, that the classes are only output once in the compiled css.

### DIFF
--- a/src/globals/scss/_spacing.scss
+++ b/src/globals/scss/_spacing.scss
@@ -1,0 +1,44 @@
+@import "vars";
+
+// Spacing properties
+$spacing-properties: (
+  "padding": "p",
+  "margin": "m",
+) !default;
+
+// Spacing directions
+$spacing-directions: (
+  null: null,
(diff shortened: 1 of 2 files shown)


fofr:
I like this a lot 🎉

fofr:
In government-frontend we found that our equivalent of these classes also needed to be responsive. eg Something that is scale-3 on wide viewports and scale-2 on thin.

PR #143, opened 2017-07-20
Update components to use spacing scale

gemmaleigh:
This PR:
- replaces px values with spacing-scale variables
- imports the spacing partial in each component that uses the spacing-scale variables

Things to decide:
In some cases em values have been swapped for pixel based values, where we have a values in em, would we like to keep this and use a px to em helper to convert between the two? this is currently inconsistent - see breadcrumb and button as examples. 

There are also a couple of places where we have 2px and 4px padding. I've added comments where hard coded px values still exist.

### DIFF
--- a/src/components/breadcrumb/_breadcrumb.scss
+++ b/src/components/breadcrumb/_breadcrumb.scss
@@ -1,5 +1,6 @@
 @import "../../globals/scss/import-once";
 @import "../../globals/scss/helpers-url";
+@import "../../globals/scss/spacing";
 @import "../../globals/scss/typography";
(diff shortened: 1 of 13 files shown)


fofr:
Should `$spacing-scale` be prefixed with `govuk-`?

gemmaleigh:
@fofr for consistency with the other vars, yes it should - i'll amend in this PR.

gemmaleigh:
I've also added a card so we can take a look at where we are using ems vs pixels. 
https://github.com/alphagov/govuk-frontend/projects/2#card-3865764

PR #148, opened 2017-07-21
Example pages

igloosi:
This PR:
Adds examples page(s) for showing combined components  
Adds examples gulp task that copies files and wraps them in a template 
Updates dev, review and watch tasks. 

Example page url: http://govuk-frontend-review-pr-148.herokuapp.com/examples/example-form.html

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -18,6 +18,7 @@ require('./tasks/gulp/demo-build.js')
 require('./tasks/gulp/preview-compile.js')
 require('./tasks/gulp/preview-component-list.js')
 require('./tasks/gulp/preview-docs.js')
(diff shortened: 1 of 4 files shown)


src/components/list/_list.scss
```diff
@@ -61,7 +61,8 @@
   }
 
   .govuk-c-list__icon {
-    @include govuk-bold-12;
+    @include govuk-bold-14;
+    // @include govuk-bold-12;
```
nickcolley: Commented out code can be removed since it'll be in version control

gemmaleigh:
Can we create govuk-bold-12 if we need it, rather than use the next mixin up?

PR #152, opened 2017-07-24
Form fieldset and legend elements

gemmaleigh:
Add a fieldset component. 

`.govuk-c-fieldset`
`.govuk-c-fieldset__legend`

Copy existing fieldset and legend styles from govuk_elements.

Update the example form to use the fieldset component.


gemmaleigh:
Yeah I wasn't sure either. I think it makes sense for now, but it might be more manageable in future to have a component that includes all of the smaller form components and shows how they can be used together.

PR #155, opened 2017-07-25
Fix height to show grid when scroll in on

igloosi:
Fixes the issue when grid is not visible beyond the window body height on longer pages

### DIFF
--- a/src/component-view-template.html
+++ b/src/component-view-template.html
@@ -37,8 +37,9 @@
     top: 0;
     bottom: 0;
     width: 100%;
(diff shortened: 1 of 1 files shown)


36degrees:
Is there a reason the grid can't be applied as a background by adding a class to the `body`? This would then become a lot simpler?

igloosi:
someone woul'd need to create the grid graphic that is transparent enough to that we can take opacity off. also body & html would need to be 100% and min-height: 100%

PR #160, opened 2017-07-26
Normalize form elements' heights

igloosi:
This PR adjusts properties of the following elements to match their heights across the breakpoints:
- input
- button
- select-box
- radio
- checkbox

Test page: http://govuk-frontend-review-pr-160.herokuapp.com/examples/form-elements-alignment.html

Screenshots

Chrome
screen shot 2017-07-26 at 15 32 13

Chrome mobile view
screen shot 2017-07-26 at 15 31 25

IE10
bs_win7_ie_10 0

IE 11
bs_win7_ie_11 0

Safari
safari

### DIFF
--- a/src/components/radio/_radio.scss
+++ b/src/components/radio/_radio.scss
@@ -13,7 +13,7 @@
     position: relative;
 
     margin-bottom: $govuk-spacing-scale-2;
-    padding: 0 0 0 38px;
+    padding: 0 0 0 40px;
 
     clear: left;
 
@@ -41,8 +41,8 @@
     top: 0;
     left: 0;
 
-    width: 38px;
-    height: 38px;
+    width: 40px;
+    height: 40px;
 
     cursor: pointer;
 
@@ -56,7 +56,7 @@
   .govuk-c-radio__label {
     display: block;
 
-    padding: 8px $govuk-spacing-scale-2 9px 12px;
+    padding: 12px $govuk-spacing-scale-2 8px 12px;
 
     cursor: pointer;
 
@@ -65,7 +65,7 @@
     touch-action: manipulation;
 
     @include mq($from: tablet) {
-      padding-top: 7px;
+      padding-top: 10px;
       padding-bottom: 7px;
     }
   }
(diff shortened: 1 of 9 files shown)


igloosi:
also should this be back-ported to elements?

nickcolley:
How does this look across browsers?

igloosi:
slighly broken... :(

edwardhorsford:
Will take a proper look later.

Something to mention - Tim and I have discussed having an alternate set of smaller scaled inputs - mostly checkboxes and radios, but could also apply to text inputs. This would be useful for things like admin interfaces. I feel like normalising might help with this.

edwardhorsford:
The boxes for select and inputs are aligned, but the text isn't. It looks to be 2px higher in the selects (on Chrome desktop).

screen shot 2017-07-27 at 13 07 48

Suggest future examples have the same text in all so we can see the alignment there too.

igloosi:
@edwardhorsford select box is the odd one as we can't nudge the font down (not our NTA). button and input text (input text and checkbox) should be aligned

igloosi:
I've redone them using height in ems.

src/components/radio/_radio.scss
```diff
@@ -89,13 +81,13 @@
...
-    top: 9px;
-    left: 9px;
+    top: em(10px, 19px);
+    left: em(10px, 19px);
 
     width: 0;
     height: 0;
 
-    border: $govuk-spacing-scale-2 solid;
+    border: em(10px, 19px) solid;
```
gemmaleigh: Can we keep the spacing scale var here? for 10px = $govuk-spacing-scale-2

src/components/radio/_radio.scss
```diff
@@ -45,7 +46,7 @@
 
     cursor: pointer;
 
-    // IE8 doesn’t support pseudoelements, so we don’t want to hide native elements there.
+    // IE8 doesn’t support pseudoelements, so we don’t want to hide native elements there. Double colons get ommited by IE8.
```
gemmaleigh: 👍

PR #166, opened 2017-08-02
Update demo folder

igloosi:
To publish latest examples of forms page (like equal height work) we need to get those changes into the demo/ folder.

This PR:
updates the task that copies files over to the demo folder and replaces asset names 
copies newly created files into the demo folder
updates minified assets with latest changes


36degrees:
I don't really understand what these changes are doing. Can you amend the commit message to provide a bit more context as to what you're trying to achieve?

igloosi:
updated the PR text and commit message

36degrees:
Thanks :)

PR #167, opened 2017-08-03
Replace path when importing from node_modules

igloosi:
We are importing sass-mq from node modules with the path of ./node_modules/sass-mq/mq. This path doesnt resolve when a 3rd party installs our component. This commit removes the node_modules part of the path from the published package

### DIFF
--- a/tasks/gulp/packages-update.js
+++ b/tasks/gulp/packages-update.js
@@ -34,6 +34,7 @@ gulp.task('packages:update', () => {
     '!' + paths.tmp + 'globals/scss/govuk-frontend-oldie.scss'
   ])
     .pipe(replace('../../components', '@govuk-frontend'))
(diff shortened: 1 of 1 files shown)


36degrees:
I'm a little nervous about the transformations going on in this file and the fact there's nothing explaining what any of them are doing or why…

Happy for this to merge but can we look at adding some inline comments at some point explaining why each line is needed and what it's doing?

36degrees:
Actually, didn't you have a solution that added ./node_modules to the local paths so that the include could just be `@import 'sass-mq/mq'`? That felt like a nicer solution than this, is there a reason it couldn't be done that way?

igloosi:
Have tried that first but that resulted in nodule_modules (with all gov.uk dependencies) being injected into each package.
Agree about the comments, will create a separate PR to annotate those

PR #170, opened 2017-08-11
Fix missing import and function parameters in _typography

igloosi:
This PR
Fixes missing import and missing parameters in em function usage

### DIFF
--- a/src/globals/scss/_typography.scss
+++ b/src/globals/scss/_typography.scss
@@ -358,12 +358,12 @@ $is-print: false !default;
   .heading-xlarge {
     @include govuk-bold-48;
 
-    margin-top: em(15, 32);
-    margin-bottom: em(30, 32);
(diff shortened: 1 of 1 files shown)


src/globals/scss/_typography.scss
```diff
@@ -424,8 +425,8 @@ $is-print: false !default;
     margin-bottom: em(20, 16);
 
     @include mq($from: tablet) {
-      margin-top: em(5);
-      margin-bottom: em(20);
+      margin-top: em(5, 19);
```
gemmaleigh: I think these need units, 5px = `$spacing-scale-1` and 19px.

igloosi: I'll go through and replace scale number. em(0 doesnt necessarily need units (but can make it consistent with other uses

PR #173, opened 2017-08-16
Use nunjucks for templating

igloosi:
This PR
creates the task to compile nunjucks files to HTML
shows example usage of a single and also nested macros
adds parameter and usage documentation for phase-tag and phase-banner

To do in future PR:
Replace current view wrapping with nunjucks
Replace preview tasks compilation of html  with express serving on .njk files

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,8 @@ before_install:
   - if [[ `npm -v | cut -d. -f1` -lt 5 ]]; then npm i -g npm@latest; fi
   - npm --version
   - openssl aes-256-cbc -K $encrypted_ca4a154c2e6b_key -iv $encrypted_ca4a154c2e6b_iv -in config/tenon.json.enc -out config/tenon.json -d
+install:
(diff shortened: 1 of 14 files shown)


gemmaleigh:
Well. That was fun.

We have had issues with this branch failing in Travis.

Travis CI breaks because the optional dependency, fs-events is installed. The platform used by Travis isn’t supported and it breaks the build.

Nunjucks uses chokidar, which has fs-events as a dependency. 

image

Setting fs-events to be optional in package.json, regenerating the package-lock.json file using `npm i --no-optional` and ensuring that the no-optional flag is passed when travis runs npm install fixes this.

Related issues:
https://github.com/npm/npm/issues/14042
https://github.com/npm/npm/issues/17671
https://github.com/npm/npm/issues/17910

Other people who have fixed it in the same way:
https://github.com/auth0/repo-supervisor/pull/3/files
https://github.com/angular/angular/pull/13945/files

PR #178, opened 2017-08-17
Add button macro

gemmaleigh:
Create a macro for the button component.

Add Nunjucks usage example and table of arguments to the README.

Display the button macro on the example page.

### DIFF
--- a/src/components/button/README.md
+++ b/src/components/button/README.md
@@ -18,6 +18,32 @@ Code example(s)
 @@include('button.html')
 ```
 
+## Nunjucks
+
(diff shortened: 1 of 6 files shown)


nickcolley:
Not sure if relevant but the interface to our button component changes internally if it's a link or not depending on if a href is passed:

https://github.com/alphagov/static/blob/master/app/views/govuk_component/button.raw.html.erb#L6

gemmaleigh:
Thanks @nickcolley! That's a better solution, have revised this to work in the same way.

PR #186, opened 2017-08-22
Generate code examples automatically

igloosi:
This PR:
Updates the readme files with  compiled HTML example markup and nunjucks usage
Updates tasks to insert example into README, fixes example urls and names
Deletes manual .html component example files where .njk files exist
Screenshots
screen shot 2017-08-22 at 10 21 59

screen shot 2017-08-22 at 10 17 43


gemmaleigh:
👍 

This works for me. 

A few things we might need to think about later - it would be nice if we could also preview the code examples when browsing the GitHub readme.

The table of arguments looks pretty borked at the moment, we should use the table component to style those.

I appear to have used "Usage" above the table of arguments, where it should be "Arguments" for some of the components, is there a way to create a template for all of these README files which has the same section headings - to ensure they are consistent?

tasks/gulp/preview-docs.js
```diff
@@ -9,7 +9,7 @@ const merge = require('merge-stream')
 const fileinclude = require('gulp-file-include')
 
 gulp.task('preview:docs', () => {
-  let original = gulp.src([paths.components + '**/*.html'])
+  let original = gulp.src(['!' + paths.components + '**/*.html', paths.components + '**/*.html'])
```
gemmaleigh: Should this have -example.html on the end?

igloosi: thanks

36degrees:
The code snippets would appear in the READMEs in the `dist` folder right?

igloosi:
yeah

PR #193, opened 2017-08-23
Add select box nunjucks template and macro

igloosi:
This PR 
Adds a select box macro, template file and nunjucks component file.
Displays how to call the component using Nunjucks and a table of arguments for the README.

### DIFF
--- a/src/components/select-box/README.md
+++ b/src/components/select-box/README.md
@@ -18,6 +18,63 @@ Code example(s)
 @@include('select-box.html')
 ```
 
+## Nunjucks
+```
+{{ govukSelectBox(
+  classes='',
+  id='select-box-1',
+  name='select-box-1',
+  options=[
+    {
+      value: '1',
+      label: 'GOV.UK frontend option 1'
+    },
+    {
+      value: '2',
+      label: 'GOV.UK frontend option 2'
+    },
+    {
+      value: '3',
+      label: 'GOV.UK frontend option 3'
+    }
+  ]
+)}}
+
+{{ govukSelectBox(
+  hasLabelWithText='Default label text',
+  labelClasses='select-box__label',
+  classes='',
+  id='select-box-2',
+  name='select-box-2',
+  options=[
+    {
+      value: 'a',
+      label: 'GOV.UK frontend option a'
+    },
+    {
+      value: 'b',
+      label: 'GOV.UK frontend option b',
+      selected: 'true'
+    },
+    {
+      value: 'c',
+      label: 'GOV.UK frontend option c'
+    }
+  ]
+)}}
+```
+
+## Arguments
+
+| Name             | Type    | Default | Required | Description
+|---               |---      |---      |---       |---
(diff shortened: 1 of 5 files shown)


src/components/select-box/macro.njk
```diff
@@ -0,0 +1,3 @@
+{% macro govukSelectBox(classes='', hasLabelWithText,labelClasses='', id, name, options) %}
```
gemmaleigh: I'm being annoying sorry, can we add a space before labelClasses.

src/components/select-box/select-box.njk
```diff
@@ -0,0 +1,43 @@
+{% from "select-box/macro.njk" import govukSelectBox %}
+
+
+{{ govukSelectBox(
+  id='select-box-1',
```
gemmaleigh: This is missing `classes=''`at the top

src/components/select-box/select-box.njk
```diff
@@ -0,0 +1,43 @@
...
+    {
+      value: '3',
+      label: 'GOV.UK frontend option 3'
+    }
+  ]
+)}}
+
+{{ govukSelectBox(
+  hasLabelWithText='Default label text',
+  labelClasses='select-box__label',
```
gemmaleigh: What does this class do? It isn't nested within .govuk-c-select-box. If it is for optional classes, it should probably be empty.

src/components/select-box/template.njk
```diff
@@ -0,0 +1,10 @@
+{% from "label/macro.njk" import govukLabel %}
+
+{% if hasLabelWithText %}
+{{ govukLabel(labelClasses, hasLabelWithText) }}
+{% endif %}
+<select class="govuk-c-select-box {{classes}}" id="{{id}}" name="{{name}}">
```
gemmaleigh: Can you please add spaces before and after the variable names, for consistency with the rest.

src/components/select-box/select-box.njk
igloosi: if you want to render a label with the select box thas has addidional custom class then you'd pass these two variables. I can put them at then end

src/components/select-box/select-box.njk
igloosi: i can make them consistent with the rest of components, but I set default empty here
```
{% macro govukSelectBox(classes=''
```
so that I dont need to specify classes when calling the macro. can change so that we show that classes is an available parameter

src/components/select-box/README.md
```diff
@@ -18,6 +18,63 @@ Code example(s)
...
+  ]
+)}}
+```
+
+## Arguments
+
+| Name             | Type    | Default | Required | Description
+|---               |---      |---      |---       |---
+| classes          | string  |         | No       | Optional additional classes
+| name             | string  |         | Yes      | Name of the group of checkboxes
```
gemmaleigh: These need 'checkbox' removing.

PR #194, opened 2017-08-23
Add panel macro

gemmaleigh:
Add panel nunjucks file, macro and template.

Display panel on the example page.

Make the confirmation page example the default, as we don't have an example of the information panel. We can add this variant later when we have a better idea of its usage.

#### Screenshot:
gov uk frontend - panel

### DIFF
--- a/src/components/panel/README.md
+++ b/src/components/panel/README.md
@@ -22,6 +22,28 @@ Code example(s)
 @@include('panel.html')
 ```
 
+## Nunjucks
+
+```
+{% from "panel/macro.njk" import govukPanel %}
+
+{{ govukInput(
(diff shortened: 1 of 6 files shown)


src/components/panel/README.md
```diff
@@ -22,6 +22,28 @@ Code example(s)
 @@include('panel.html')
 ```
 
+## Nunjucks
+
+```
+{% from "panel/macro.njk" import govukPanel %}
+
+{{ govukInput(
```
nickcolley: Was this meant to be govukPanel?

nickcolley: cc @gemmaleigh

PR #196, opened 2017-08-23
Add table macro

igloosi:
This PR:
Adds table template, macro and arguments table to README
Optional caption and table head parameters

Screenshots
screen shot 2017-08-23 at 17 06 47

### DIFF
--- a/src/components/table/README.md
+++ b/src/components/table/README.md
@@ -19,7 +19,142 @@ Table
 </a>
 </p>
 
-  <h2 class="govuk-u-heading-24">How to call this component</h2>
+<h2 class="govuk-u-heading-24">How to call this component</h2>
+```
+@@include('table.html')
+```
+## Nunjucks
+
+```
+{% from 'table/macro.njk' import govukTable %}
+
+{# table with a caption #}
+{{ govukTable(
+  classes='',
+  options = {
+    'caption': 'Months and rates',
+    'captionSize': 'small',
+    'isFirstCellHeader': 'true'
+  },
+  data = {
+    'rows' : [
+      [
+        {
+          text: 'January'
+        },
+        {
+          text: '£85',
+          type: 'numeric'
+        },
+        {
+          text: '£95',
+          type: 'numeric'
+        }
+      ],
+      [
+        {
+          text: 'February'
+        },
+        {
+          text: '£75',
+          type: 'numeric'
+        },
+        {
+          text: '£55',
+          type: 'numeric'
+        }
+      ],
+      [
+        {
+          text: 'March'
+        },
+        {
+          text: '£165',
+          type: 'numeric'
+        },
+        {
+          text: '£125',
+          type: 'numeric'
+        }
+      ]
+    ]
+  }
+)}}
+
+{# table with a head #}
+{{ govukTable(
+  classes='',
+  options = {
+    'isFirstCellHeader': 'true'
+  },
+  data = {
+    'head' : [
+      {
+        text: 'Month you apply'
+      },
+      {
+        text: 'Rate for bicycles',
+        type: 'numeric'
+      },
+      {
+        text: 'Rate for vehicles',
+        type: 'numeric'
+      }
+    ],
+    'rows' : [
+      [
+        {
+          text: 'January'
+        },
+        {
+          text: '£85',
+          type: 'numeric'
+        },
+        {
+          text: '£95',
+          type: 'numeric'
+        }
+      ],
+      [
+        {
+          text: 'February'
+        },
+        {
+          text: '£75',
+          type: 'numeric'
+        },
+        {
+          text: '£55',
+          type: 'numeric'
+        }
+      ],
+      [
+        {
+          text: 'March'
+        },
+        {
+          text: '£165',
+          type: 'numeric'
+        },
+        {
+          text: '£125',
+          type: 'numeric'
+        }
+      ]
+    ]
+  }
+)}}
+```
+
+## Arguments
+
+| Name             | Type    | Default | Required | Description
+|---               |---      |---      |---       |---
+| classes          | string  |         | No       | Optional additional classes
+| data             | object  |         | Yes      | Data for the table: rows, head and cells
+| data.options     | Object  |         | No       | Caption text , size and if first cell in bolder
+| data.head        | Array   |         | No       | Data for the table head cells
+| data.rows        | Array   |         | Yes      | Data for the table: rows, head and cells
 
   <pre><code></code></pre>
 
@@ -36,4 +171,3 @@ Table
 
 <h2 class="govuk-u-heading-24">Installation</h2>
 <pre><code>npm install --save @govuk-frontend/table</code></pre>
-
--- a/src/components/table/README.md
+++ b/src/components/table/README.md
@@ -152,9 +152,9 @@ Table
 |---               |---      |---      |---       |---
 | classes          | string  |         | No       | Optional additional classes
 | data             | object  |         | Yes      | Data for the table: rows, head and cells
-| data.options     | Object  |         | No       | Caption text , size and if first cell in bolder
+| options          | Object  |         | No       | Caption text, size and if first cell is bolder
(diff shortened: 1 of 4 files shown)


src/components/table/template.njk
```diff
@@ -0,0 +1,27 @@
...
+    {% endfor %}
+  </tr>
+  </thead>
+  {% endif %}
+  <tbody class="govuk-c-table__body">
+    {% for label, rows in data.rows %}
+      <tr class="govuk-c-table__row">
+        {% for label, cell in rows %}
+        {% if loop.first %}
+          <th class="govuk-c-table__header" scope="row">{{ label }}</th>
```
36degrees: - This assumes the first column will always be the heading – is that always going to be the case?
- The way this is structured means that the first row can never be of type `numeric`.
- Do we need to allow for other attributes to be specified per cell e.g. colspan, so that more complicated tables can be produced using this macro?

src/components/table/template.njk
```diff
@@ -0,0 +1,27 @@
+<table class="govuk-c-table {{ classes }}">
+  {% if caption %}
+  <caption class="govuk-c-table__caption {{captionSize}}">{{ caption }}</caption>
+  {% endif %}
+  {% if data.head %}
+  <thead class="govuk-c-table__head">
+    <tr class="govuk-c-table__row">
+    {% for label, type in data.head %}
```
36degrees: - `label` doesn't seem like the right name here when it could be e.g. `'£75.00'`.
- what you're referring to as `type` is currently either an empty hash or a hash containing {'type': 'numeric'}. Would it make more sense to call this `options`?

src/components/table/table.njk
```diff
@@ -0,0 +1,84 @@
...
+      'Month you apply': {},
+      'Rate for vehicles': {
+        'type': 'numeric'
+      },
+      'Rate for bicycles': {
+        'type': 'numeric'
+      }
+    },
+    'rows' : {
+      'row' : {
```
36degrees: Does this being a hash make sense? These labels (`row`, `row2`) don't seem to be used for anything. Should this be an array of rows instead?

src/components/table/template.njk
```diff
@@ -0,0 +1,27 @@
+<table class="govuk-c-table {{ classes }}">
+  {% if caption %}
+  <caption class="govuk-c-table__caption {{captionSize}}">{{ caption }}</caption>
```
36degrees: Are there issues if `captionSize` is not set?

Also, if caption size is `heading-small` as in the example docs, the class added will just be `heading-small`. Is that a valid class, or should it be a modifier? Either way, would it make sense to allow the user to set just `'captionSize': 'small'` and add the rest of the class name here?

igloosi: yeah can do if captionSize then add whole modifier class. we will have some global heading styles like we have now e.g heading-medium, heading small so we could either apply that class (and its modifiers) or a new modifier just for this caption

src/components/table/template.njk
igloosi: valid, will figure out what options can be there

src/components/table/template.njk
```diff
@@ -0,0 +1,27 @@
...
+  {% endif %}
+  {% if data.head %}
+  <thead class="govuk-c-table__head">
+    <tr class="govuk-c-table__row">
+    {% for label, type in data.head %}
+      <th class='govuk-c-table__header {% if type %} govuk-c-table__header--numeric {% endif %}' scope="col">{{ label }}</th>
```
igloosi: can do and we probably should do

src/components/table/template.njk
igloosi: hmm, need to look what ways we can structure the table, but yes, colspan could come in play

src/components/table/table.njk
igloosi: refactoring :)

igloosi:
Updated table data structure, where each row is an array and each cell is an object.
Options for caption and if cell is a header are in an object

src/components/table/template.njk
```diff
@@ -0,0 +1,29 @@
...
+  </thead>
+  {% endif %}
+  <tbody class="govuk-c-table__body">
+    {% for rows in data.rows %}
+    <tr class="govuk-c-table__row">
+    {% for cell in rows %}
+      {% if (loop.first) and (options.isFirstCellHeader) %}
+      <th class="govuk-c-table__header" scope="row"> {{ cell.text }}</th>
+      {% elseif loop.first %}
+      <td class="govuk-c-table__cell {% if cell.type %} govuk-c-table__cell--numeric {% endif %}" {% if cell.colspan %} colspan="{{ cell.colspan }}" {% endif %} {% if cell.rowspan %} rowspan="{{ cell.rowspan }}" {% endif %} scope="row">{{ cell.text }}</td>
```
36degrees: This is still hardcoded so if you set e.g. `'type': 'date'` you'll end up with the numeric class being applied - should be as with the headings which interpolate the type into the class name.

src/components/table/template.njk
```diff
@@ -0,0 +1,29 @@
...
+      <th class="govuk-c-table__header {% if item.type %} govuk-c-table__header--{{ item.type }} {% endif %}" {% if item.colspan %} colspan="{{ item.colspan }}" {% endif %} {% if item.rowspan %} rowspan="{{ item.rowspan }}" {% endif %} scope="col">{{ item.text }}</th>
+    {% endfor %}
+  </tr>
+  </thead>
+  {% endif %}
+  <tbody class="govuk-c-table__body">
+    {% for rows in data.rows %}
+    <tr class="govuk-c-table__row">
+    {% for cell in rows %}
+      {% if (loop.first) and (options.isFirstCellHeader) %}
```
36degrees: Mostly curious more than anything - are the brackets around the two conditions needed?

src/components/table/template.njk
```diff
@@ -0,0 +1,29 @@
...
+  <thead class="govuk-c-table__head">
+    <tr class="govuk-c-table__row">
+    {% for item in data.head %}
+      <th class="govuk-c-table__header {% if item.type %} govuk-c-table__header--{{ item.type }} {% endif %}" {% if item.colspan %} colspan="{{ item.colspan }}" {% endif %} {% if item.rowspan %} rowspan="{{ item.rowspan }}" {% endif %} scope="col">{{ item.text }}</th>
+    {% endfor %}
+  </tr>
+  </thead>
+  {% endif %}
+  <tbody class="govuk-c-table__body">
+    {% for rows in data.rows %}
```
36degrees: Would make more sense for this to be called `row`.

src/components/table/README.md
```diff
@@ -17,6 +17,138 @@ Code example(s)
...
+)}}
+```
+
+## Arguments
+
+| Name             | Type    | Default | Required | Description
+|---               |---      |---      |---       |---
+| classes          | string  |         | No       | Optional additional classes
+| data             | object  |         | Yes      | Data for the table: rows, head and cells
+| data.options     | Object  |         | No       | Caption text , size and if first cell in bolder
```
36degrees: Options is not nested under data?

src/components/table/README.md
```diff
@@ -17,6 +17,138 @@ Code example(s)
...
+| data.head        | Array   |         | No       | Data for the table head cells
+| data.rows        | Array   |         | Yes      | Data for the table: rows, head and cells
```
36degrees: Head goes in data.head?

src/components/table/template.njk
igloosi: yep, agree. done

src/components/table/template.njk
igloosi: it doesnt really, so updated

src/components/table/template.njk
igloosi: missed it, ta

src/components/table/README.md
igloosi: true, fixed

src/components/table/README.md
igloosi: changed

PR #206, opened 2017-09-05
Update link to design system

igloosi:
This PR:
changes block to variable with default url (base url + component name) that can be overridden in each component if necessary

### DIFF
--- a/src/components/breadcrumb/index.njk
+++ b/src/components/breadcrumb/index.njk
@@ -10,9 +10,8 @@
 {# componentNunjucks #}
 {# componentHtml #}
 
-{% block componentGuidanceLink %}
-  http://www.linktodesignsystem.com
-{% endblock %}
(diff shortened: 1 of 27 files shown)


src/components/breadcrumb/index.njk
```diff
@@ -10,9 +10,8 @@
 {# componentNunjucks #}
 {# componentHtml #}
 
-{% block componentGuidanceLink %}
-  http://www.linktodesignsystem.com
-{% endblock %}
+{# override link to design system here if it's diffrent to base url + componentName #}
```
gemmaleigh: Tiny thing, there's a typo - different, needs updating for all comments here. Apart from that, super!

igloosi:
Find&replace all :)

gemmaleigh:
💥

PR #212, opened 2017-09-06
Update component documentation

igloosi:
This PR:
amends the component nunjucks template to match sections outlined in the techwriter's template
updates content for button, breadcrumb and radio button components
generates updated readme

PR #214, opened 2017-09-06
Fix html include to readme

igloosi:
This PR:
fixes the wrong code example include in the html example section
updates all the readmes

This solution needs to be checked again if we still rely on html files being in the repo.


igloosi:
closing in favour of #212

PR #222, opened 2017-09-08
Add a component for entering dates

gemmaleigh:
Update the fieldset component, so that it can display legend text, optional hint text and an error message.

Create a date macro for the date component.

Regenerate the README for the date component.

Use the table component to create a table of arguments for the date component, displayed in the README.

### DIFF
--- a/src/components/date/README.md
+++ b/src/components/date/README.md
@@ -7,7 +7,7 @@ Date
 
 <h2 class="govuk-u-heading-24">Introduction</h2>
 <p class="govuk-u-core-24">
-  Date description.
+  A component for entering dates, for example - date of birth.
 </p>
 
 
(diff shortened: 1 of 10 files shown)


nickcolley:
Should the README have copy pasted versions of the markup?

Should this be generated from the templating?

Otherwise updating components will be very time consuming.

gemmaleigh:
Hi @nickcolley,

There's a Gulp task to generate the README files. The markup isn't copied and pasted into the READMEs.

Here's the Gulp task:
https://github.com/alphagov/govuk-frontend/blob/master/tasks/gulp/generate-readme.js

And the template we use for both the README.md file and the component "detail" page is here:
https://github.com/alphagov/govuk-frontend/blob/master/src/views/component.njk

nickcolley:
As a potential contributor having built files in `src` is really confusing, and I know from past experience people update generated files that are not clearly generated.

Maybe the READMEs can exist but link to the web docs or something else?

gemmaleigh:
We'd like to support those who want to browse the repo for component information via GitHub. 

These READMEs allow that, for example: https://github.com/alphagov/govuk-frontend/tree/master/src/components/button

Perhaps we need to make it clear that to edit a component's "detail" page or README, the "index.njk" file is the one to edit.

We're trialling this format for the docs, so we'll see how it goes in research.

PR #223, opened 2017-09-11
Compile component Nunjucks files to HTML

gemmaleigh:
Remove component HTML files in `/src/components` - these are no longer required, now that we have a `template.njk` files for each component.

Add a Gulp task to compile Nunjucks to HTML for each component. 

Amend the app, so that location of the HTML files for each component is now`/public/components/`

Use taskArguments so that the destination path for the compile-components task can be changed.


tasks/gulp/compile-components.js
```diff
@@ -0,0 +1,23 @@
...
+const paths = require('../../config/paths.json')
+const gulp = require('gulp')
+const nunjucks = require('gulp-nunjucks')
+const rename = require('gulp-rename')
+const taskArguments = require('./task-arguments')
+
+// Compile Nunjucks ----------------------
+// Compile Nunjucks to HTML
+// ---------------------------------------
+gulp.task('compile:components', cb =>
```
igloosi: is callback needed?

PR #225, opened 2017-09-12
NPM Build packages task

igloosi:
This PR
* creates a npm run build:packages task and then copies relevant files to packages/ folder
* removes old tasks


gemmaleigh:
When I run this locally, it removes the `package.json` file for each component - I think we need to keep these. 

I don't think we need to copy the index.njk file to `packages/componentname`, the content of this file (that is relevant to the package) should appear in the README. 

Otherwise, this appears to be working as it should!

igloosi:
strange, my package.json files remain unchanged and still present.
true, i will filter out index files.

Issue #226, opened 2017-09-12
Generated README files


igloosi:
To avoid duplicating code examples and documentation manually, as well as having full docs in the source directory for each component,  we generate the README markdown file and a dev preview page from the same set of nunjucks templates. 

This will allow us to test our assumptions with consumers.

It could be the case that we disable the generation task once we are confident in the docs content, but there is still the question creating and updating docs for component when a contribution happens.

36degrees:
@igloosi this has been done, right?

igloosi:
yeah done

PR #227, opened 2017-09-12
Compile READMEs to proper markdown

igloosi:
This PR
adds a gulp plugin to compile to markdown
amends generate:readme task to add a pipe step

### DIFF
--- a/tasks/gulp/generate-readme.js
+++ b/tasks/gulp/generate-readme.js
@@ -8,6 +8,7 @@ const data = require('gulp-data')
 const tap = require('gulp-tap')
 const path = require('path')
 const fs = require('fs')
+const toMarkdown = require('gulp-to-markdown')
 const vinylInfo = {}
 
 function getDataForFile (file) {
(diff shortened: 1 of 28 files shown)


tasks/gulp/generate-readme.js
```diff
@@ -34,6 +35,7 @@ gulp.task('generate:readme', () => {
     path: [paths.src + 'views', paths.components],
     manageEnv: manageEnvironment
   }))
+  .pipe(toMarkdown({ gfm: true }))
```
gemmaleigh: Can we add a comment here about where this came from and what it does.

gemmaleigh:
This is awesome and addresses @36degrees' concern about being able to read the markdown files in your editor. 

We'll still need to address #226, as we've got generated files in `/src/`, but while we're likely to be changing the content for many of these files - keeping the task to generate the documentation from a components source files will make it easier to keep component documentation up-to-date.

igloosi:
opened issue #226 for that

gemmaleigh:
:1st_place_medal:

36degrees:
👏

PR #231, opened 2017-09-13
NPM Build dist task

igloosi:
This PR
adds a task to move files to dist/ folder while:
* minifying css and js
* appending version to css and js
* updating VERSION.txt file

We're reusing tasks so they have if statements present to perform pipe actions based on flags passed to them

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -16,6 +16,7 @@ require('./tasks/gulp/generate-readme.js')
 require('./tasks/gulp/watch.js')
 // new tasks
 require('./tasks/gulp/copy-to-destination.js')
+require('./tasks/gulp/asset-version.js')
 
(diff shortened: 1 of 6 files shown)


gemmaleigh:
The import paths at the top of each file have changed from `@import "../../globals/scss/import-once";` to `@import @govuk-frontend/globals/import-once";`.

Should these remain relative paths to globals in the component scss files?

One tiny thing, I think that `dist/js/govuk-frontend0.0.11-alpha.min.js` should be `dist/js/govuk-frontend-0.0.11-alpha.min.js`, with a hyphen before the version number. The same for the minified css files.

igloosi:
true, missing couple of if statements. fixed now along with name output

gemmaleigh:
💯

PR #232, opened 2017-09-13
Replace gulp-tap with vinyPaths

igloosi:
Using it in #231  so it made sense to consolidate plugin usage


gemmaleigh:
If we're using `configPaths` rather than `paths` here, it should probably be updated in the other gulp tasks too, for consistency.

igloosi:
fair point, done

PR #234, opened 2017-09-14
Add a file upload component

gemmaleigh:
This should possibly be a variant of the input component (as the two
are almost identical). However if we intend to replace this with a
custom component in the future, it makes sense to keep the two
separate.


nickcolley:
I call dibs on building the custom one

gemmaleigh:
We welcome PRs!

PR #236, opened 2017-09-15
Remove whitespace and fix component indentation

gemmaleigh:
This PR removes the whitespace after each class attribute and ensures the indentation of the HTML output when the component template file is compiled is correct.


gemmaleigh:
No worries! I haven't generated the READMEs yet as it looks like they render slightly differently than the output HTML in /public/components.

I spotted a couple of errors with the label and input components, where we're importing another component first and need to be able to define additional classes for both the label and the input. Have fixed those for input and file upload.

gemmaleigh:
We will also be able to remove the additional (strip whitespace before and after)`{{-` and `-}}` for each component once we've split the variants into separate files.

PR #237, opened 2017-09-18
Link variants

igloosi:
This PR:
* splits link variable into separate files
* replaces markdown table syntax with govuk-c-table component, to be consistent with other components


igloosi:
Update to table macro  created in PR #238

PR #238, opened 2017-09-18
Escape html in cell text

igloosi:
This will allow the use of `<span>` or `<br>` in tables.
screen shot 2017-09-18 at 11 16 53

### DIFF
--- a/src/components/table/template.njk
+++ b/src/components/table/template.njk
@@ -2,7 +2,7 @@
   {%- if classes %} {{ classes }}{% endif %}">
   {% if options.caption %}
   <caption class="govuk-c-table__caption
-  {%- if options.captionSize %} {{ options.captionSize }}{% endif %}">{{ options.caption }}</caption>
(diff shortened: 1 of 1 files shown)


36degrees:
Firstly, this PR is **disabling** the escaping of HTML within tables, not escaping the HTML. The commit message is also a little misleading as it could imply that only `span` and `br` tags will work, which is not the case.

There's a couple of wider things going on here that I'd like us to step back and consider.

1. We're making changes to the Frontend components here as a result of _our needs_ in building the documentation. I like the _idea_ of using the components to build the documentation, but I'm concerned that it's causing us to forget what these components are and who they're for…

2. If there _are_ valid user needs to allow HTML to be passed into the table component, this should be clearly documented and ideally consistently applied across all components. Requiring the developer to manually inspect and remember which components escape HTML and which do not has security implications when dealing with user submitted data.

gemmaleigh:
Would it be possible to close this and instead re-write the documentation in #237 to remove the <br/> tags. These could be in a comma-separated list, rather than on separate lines. Then we wouldn't need to make this change to the table component.

igloosi:
can do.

PR #249, opened 2017-10-02
Update build & development task documentation to reflect changes in tooling

igloosi:
We have consolidated gulp tasks so we need to document those changes and describe what tasks do and how they're run.

This PR: 
* documents the main npm run tasks that are listed in `package.json`
* document important gulp tasks that are triggered in `gulpfile.js` and are defined in `tasks/gulp/`
* update references to old tasks


alex-ju:
Followed the documentation, tested the tasks.
Everything works fine.

PR #251, opened 2017-10-04
Button component variants from YAML file

igloosi:
This PR:
- adds a YAML file where default option and variants are defined
- updates macro, template and documentation to read and display those variants

### DIFF
--- a/src/components/button/button.yaml
+++ b/src/components/button/button.yaml
@@ -2,7 +2,7 @@ variants:
 - name: default
   data:
     text: Save and continue
-    classes: 'boom'
+    classes: 'example-class'
 - name: disabled
   data:
     text: Disabled button
--- a/src/components/button/button.yaml
+++ b/src/components/button/button.yaml
@@ -1,0 +1,14 @@
+variants:
+- name: default
+  data:
+    text: Save and continue
+    classes: 'boom'
+- name: disabled
+  data:
+    text: Disabled button
+    isDisabled: true
+- name: start
+  data:
+    text: Start now
+    url: /
+    isStart: true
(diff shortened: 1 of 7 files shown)


src/components/button/button.yaml
```diff
@@ -0,0 +1,14 @@
...
+    classes: 'boom'
+- name: disabled
+  data:
+    text: Disabled button
+    isDisabled: true
+- name: start
+  data:
+    text: Start now
+    url: /
+    isStart: true
```
fofr: Is `is` needed in `isStart` and `isDisabled`?
On GOV.UK we tried to mirror the DOM where possible, "disabled" would follow that approach.

src/components/button/button.yaml
```diff
@@ -0,0 +1,14 @@
+variants:
```
fofr: Are variants the same as examples?

After research with GOV.UK developers we changed our equivalent from "fixtures" to "examples". (https://github.com/alphagov/govuk_publishing_components/pull/37)

src/components/button/button.yaml
igloosi: I can do either way with this one. Seeing we're proposing is- prefix for component states I'd like to follow similar convention. We will be testing this on devs.

src/components/button/button.yaml
igloosi: Interesting. was fixtures more obvious than variants? 
yes, variants also referrers to examples

fofr: We didn't try variants. People referred to them as examples, so we used the term they chose.

There may be some subtle differences with how you're using the YAML files. I think it'd be useful to keep YAML structures the same to keep components interoperable.

PR #255, opened 2017-10-04
Remove demo folder and config

igloosi:
For now we don't need to publish the demo folder to Heroku so it can be removed


36degrees:
The linter is complaining because `herokuApp` is still being defined but never used.

igloosi:
oops, fixed

PR #252, opened 2017-10-04
Table component default and variants from YAML file

igloosi:
This PR:
- adds a YAML file where default option and variants are defined
- updates macro, template and documentation to read and display those variants

PR #265, opened 2017-10-06
Update components to use the YAML-based arguments table

alex-ju:
This PR updates YAML-based components to use the new YAML-based arguments table as per https://github.com/alphagov/govuk-frontend/pull/252#issuecomment-334455218


igloosi:
could you just amend "breadcrumbs" to "items" array in arguments table. otherwise all good

PR #271, opened 2017-10-06
Update label parameter definition when used in input

igloosi:
When we updated the label and input parameter definition, the label stopped showing when used in the input macro.
This PR removes unneeded specific parameter item declarations with the params object. Keys  are picked up by the label macro automacitally


igloosi:
actually this is solved by #267 
closing

PR #272, opened 2017-10-09
Use macro for default and variants

alex-ju:
This PR avoids redundancy by:
* Adding a macro for generating default and variants
* Replacing the current 'default and variants' block with a macro call

### DIFF
--- a/src/components/breadcrumb/index.njk
+++ b/src/components/breadcrumb/index.njk
@@ -1,5 +1,4 @@
 {% extends "component.njk" %}
-
 {% from "macros/showDefaultAndVariants.njk" import showDefaultAndVariants %}
 
 {# Commented out blocks below inherit from views/component.njk #}
@@ -9,8 +8,7 @@ A breadcrumb is a GOV.UK element that helps users to understand where they are w
 {% endblock %}
 
 {% block defaultAndVariants %}
-{# {% from "breadcrumb/macro.njk" import govukBreadcrumb %} #}
-{{ showDefaultAndVariants(importStatement, componentName, componentData) }}
+{{ showDefaultAndVariants(componentName, componentData) }}
 {% endblock %}
 
 {% block componentDependencies %}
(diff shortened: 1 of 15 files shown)


36degrees:
Is this a work in progress? Getting 'internal server error' on the review app when trying to view a variant.

src/components/breadcrumb/index.njk
```diff
@@ -8,44 +8,7 @@ A breadcrumb is a GOV.UK element that helps users to understand where they are w
...
-</code></pre>
-
-<h4 class="govuk-u-copy-19">Macro</h4>
-<pre><code>
-{% raw %}{{ govukBreadcrumb({% endraw %}
-  {{ item.data | dump(2) }}
-{% raw %}) }}{% endraw %}
-</code></pre>
-{% endfor %}
+{{ showDefaultAndVariants(componentName, componentData) }}
```
36degrees: Given this appears to be the same across all components, can this now just go into the `defaultAndVariants` block in `component.njk`?

src/views/macros/loadComponentTemplate.njk
```diff
@@ -0,0 +1,3 @@
+{% macro loadComponentTemplate(componentName, params) %}
```
36degrees: Where does `params` get used?

src/components/breadcrumb/index.njk
alex-ju: Fair point. I'll move it inside.

src/views/macros/loadComponentTemplate.njk
alex-ju: The parameters (`params`) are used inside each template file

alex-ju:
Not sure where the `internal server error` comes from, but happy to look when you have time.

alex-ju:
@36degrees 
Updated:
* the DefaultAndVariants macro is now being called from the main component template
* the `internal server error` will be fixed by #273.

PR #274, opened 2017-10-10
WIP: Document table macro parameters available for the table component

igloosi:
Existing table of arguments for the table component macro was not very descriptive, so this pull request rectifies that by listing available options.
Before:
http://govuk-frontend.herokuapp.com/components/table
After:
http://govuk-frontend-review-pr-274.herokuapp.com/components/table

### DIFF
--- a/src/components/table/index.njk
+++ b/src/components/table/index.njk
@@ -18,8 +18,6 @@
 {
   "data": {
     "options": {
-      "caption": "Caption 1 : Months and rates",
-      "captionSize": "small",
       "isFirstCellHeader": true
     },
     "head": [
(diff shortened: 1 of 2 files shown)


src/components/table/index.njk
```diff
@@ -37,20 +35,160 @@
...
+          text: 'data.rows'
+        },
+        {
+          text: 'Array'
+        },
+        {
+          text: 'Yes'
+        },
+        {
+          text: 'Array of arrays, where ieach array represents a row'
```
36degrees: each

src/components/table/index.njk
```diff
@@ -37,20 +35,160 @@
...
+          text: 'data.options.isFirstCellHeader'
+        },
+        {
+          text: 'Boolean'
+        },
+        {
+          text: 'No'
+        },
+        {
+          text: 'If true, the first cell in a row will be embolden'
```
36degrees: It's not about the styling – it's about the first cell in each row being marked up as a header for the rest of the row.

src/components/table/index.njk
```diff
@@ -37,20 +35,160 @@
...
+          text: 'data'
+        },
+        {
+          text: 'Object'
+        },
+        {
+          text: 'Yes'
+        },
+        {
+          text: 'Object containing all configuration options listed below'
```
36degrees: Why is everything inside `data`, when it isn't for other components?

igloosi:
yeah saw the card in the backlog. will revisit this one. in reality it doesnt necessary need data, i thought it was helpful to separate data from options, but in light of changing all other component parameters I should probably modify that

36degrees:
It possibly does make sense to separate data from options, but at the minute we have `data.options` and `data.rows` both inside data – which is not separating them.

igloosi:
proposed structure below. any thought on this @36degrees @alex-ju @hannalaakso 
```
{
  variants: [{
    name: 'table-with-caption-and-head',
    data: {
      options: {
        caption: 'Caption 1 : Months and rates',
        captionSize: 'small',
        isFirstCellHeader: true,
        classes: 'extra=dummy-class'
      },
      head: [{
          text: 'Month you apply'
        },
        {
          text: 'Rate for bicycles',
          type: 'numeric'
        },
        {
          text: 'Rate for vehicles',
          type: 'numeric'
        }
      ],
      rows: [
        [{
            text: 'January'
          },
          {
            text: '£85',
            type: 'numeric'
          },
          {
            text: '£95',
            type: 'numeric'
          }
        ],
        [{
            text: 'February'
          },
          {
            text: '£75',
            type: 'numeric'
          },
          {
            text: '£55',
            type: 'numeric'
          }
        ],
        [{
            text: 'March'
          },
          {
            text: '£165',
            type: 'numeric'
          },
          {
            text: '£125',
            type: 'numeric'
          }
        ]
      ]
    }
  }]
}
```

alternative structure i had in mind was for the options bit
```
 data: {
      caption: 'Caption 1 : Months and rates',
      captionSize: 'small',
      isFirstCellHeader: true,
      classes: 'extra=dummy-class'
      head: []...
      rows: []....
```

alex-ju:
I agree that would be good to separate actual data from options, but I'm also wandering if the caption shouldn't be a component itself, at the same level with the table header and table rows:
`{{ govukTable({options:{},data{caption:{},head:[],rows:[]}}) }}`

```
variants: [{
    name: 'table-with-caption-and-head',
    options: {
        isFirstCellHeader: true,
        classes: 'dummy-class',
    }
    data: {
        caption: {
            captionText: 'Caption 1 : Months and rates',
            captionSize: 'small'
        }
        head: [...],
        rows: [...]
    }
}]
```

I also think that keeping main parameters consistent across components at a minimum (e.g. having only options and data) will help us extending them without having to update templates, macros, etc.

igloosi:
closing as i will create a new PR for the updated table

PR #277, opened 2017-10-11
Fix autogenerated macro examples

alex-ju:
This PR:
- makes the `capitaliseComponentName()` function available to all Nunjucks templates and macros
- uses this function in `showDefaultAndVariants.njk` for autogenerating the macro examples

### DIFF
--- a/src/views/macros/showDefaultAndVariants.njk
+++ b/src/views/macros/showDefaultAndVariants.njk
@@ -27,15 +27,11 @@
 {% set componentHtml %}
 {{ loadComponentTemplate(componentName, item.data) }}
 {% endset %}
-<pre><code>
-  {{- componentHtml | e -}}
-</code></pre>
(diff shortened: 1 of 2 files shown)


src/views/macros/showDefaultAndVariants.njk
```diff
@@ -33,7 +33,7 @@
 
 <h4 class="govuk-u-copy-19">Macro</h4>
 <pre><code>
-{% raw %}{{ loadComponentTemplate(componentName, {% endraw %}
+{% raw %}{{ {% endraw %} govuk{{ componentName | capitaliseComponentName }}(
```
36degrees: We end up with a double space here – there's a space after the `{{` and another one before `govuk`.

36degrees:
LGTM. Suggest rebasing those commits into one if possible. 👍

PR #281, opened 2017-10-12
Grid component from Yaml and usage documentation

igloosi:
Grid component is the first component where any other content and component might live inside it. There needs to be a way to open the grid wrapper, specify grid item width, insert content and other component macros, and finally close the grid wrapper
In this PR I'm
* adding yaml file for data
* updating macro and template to accept parameters from yaml
* add extra documentation block to show real world usage of the grid component

To effectively open and close the grid wrapper, I'm using gridStart and gridEnd keys on relevant items.
There are also some styles to show the grid structure, which could be also be used for debug

### DIFF
--- a/src/components/grid/_grid.scss
+++ b/src/components/grid/_grid.scss
@@ -33,6 +33,13 @@
     width: percentage( 1 / 4);
   }
 
+  // visibly show grid layout
+  .preview-grid .govuk-c-grid__item {
+    background-color: $govuk-light-blue-25;
+    border: 1px dashed $govuk-purple;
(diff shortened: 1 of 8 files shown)


36degrees:
I find this macro really confusing…

I am wondering whether it actually makes sense to provide a macro for the grid, given it's not used it in the same way the other components will be.

igloosi:
hmm, in that case we'd need to move the grid sass into globals somewhere, possibly layout. and same for site-width-container

igloosi:
consensus is not have this as a component macro.
will move sass to globals/scss/layout and delete it from components folder.

will update the trellow ticket as well

PR #283, opened 2017-10-12
Date component default and variants from YAML file

alex-ju:
This PR:
* adds a YAML file where default option and variants are defined
* updates macro, template and documentation to read and display those variants
* updates table of accepted arguments
* updates component to use new fieldset and input macros

[Trello ticket](https://trello.com/c/nXTPuJdo)


alex-ju:
the formatting is affected by the `Label` macro. I'll make a PR for that one.

PR #284, opened 2017-10-12
Update error-summary component with variants and new API conventions

igloosi:
- template.njk file is updated to be able to read the YAML file
structure
- update macro.njk file to accept new paramaters from YAML
- update table of arguments in index.njk
- update example in error-summary.njk
- rename listofErrors to errors
- remove classes key from listofErrors
- replace content in errorList with either text or html keys and amend
template accordingly
- update title to accept either titleText or titleHtml and amend
template accordingly
- update description to accept either descriptionText or descriptionHtml
and amend template accordingly
- add functionality to specify html attributes that get
appended to the element

This component uses the list component and accepts the `text`,`html`and `href` arguments for it. The rest we do not want to expose as we're not allowing changes to the error-summary look.

[Trello ticket](https://trello.com/c/LoF1Zp4D/189-3-show-default-example-and-variants-of-the-error-summary-component)

### DIFF
--- a/src/components/error-summary/error-summary.njk
+++ b/src/components/error-summary/error-summary.njk
@@ -1,19 +1,17 @@
 {% from "error-summary/macro.njk" import govukErrorSummary %}
 
-{{- govukErrorSummary(
-  classes='',
-  title='Message to alert the user to a problem goes here',
-  description='Optional description of the errors and how to correct them',
-  listClasses='',
-  listItems=[
+{{ govukErrorSummary({
+  "titleText": "Message to alert the user to a problem goes here",
+  "descriptionText": "Optional description of the errors and how to correct them",
+  "classes": "optional-extra-class",
+  "errorList": [
     {
-      text: 'Descriptive link to the question with an error',
-      url: '#example-error-1'
+      "text": "Descriptive link to the question with an error",
+      "href": "#example-error-1"
     },
     {
-      text: 'Descriptive link to the question with an error',
-      url: '#example-error-2'
+      "text": "Descriptive link to the question with an error",
+      "href": "#example-error-1"
     }
   ]
-)
--}}
+}) }}
--- a/src/components/error-summary/error-summary.yaml
+++ b/src/components/error-summary/error-summary.yaml
@@ -1,0 +1,13 @@
+variants:
+  - name: default
+    data:
+      titleText: Message to alert the user to a problem goes here
+      descriptionText: Optional description of the errors and how to correct them
+      classes: optional-extra-class
+      errorList:
+        -
+          text: Descriptive link to the question with an error
+          href: '#example-error-1'
+        -
+          html: Descriptive link to the question with an error
+          href: '#example-error-1'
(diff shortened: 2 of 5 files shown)


src/components/error-summary/error-summary.njk
```diff
@@ -1,19 +1,20 @@
...
-      url: '#example-error-2'
-    }
-  ]
-)
--}}
+{{ govukErrorSummary({
+  "title": "Message to alert the user to a problem goes here",
+  "description": "Optional description of the errors and how to correct them",
+  "listOfErrors": {
+    "classes": "govuk-c-error-summary__list",
```
36degrees: I don't think users should have to remember to add this class here - it needs to be merged into the params when the list is included.

igloosi:
needs an update to list component. will submit a pr for that before updating this.

igloosi:
dependent on #288

src/components/error-summary/error-summary.njk
igloosi: now resolved with the updated list component merged in

src/components/error-summary/index.njk
```diff
@@ -55,86 +54,86 @@
...
-        },
-        {
-          text: 'No'
-        },
-        {
-          text: 'List options'
-        }
-      ]
+       {
+         text: 'listOfErrors.is (Bullet | Number | Step | isStepLarge)'
```
36degrees: This needs updating.

src/components/error-summary/error-summary.yaml
```diff
@@ -0,0 +1,28 @@
...
+      description: Optional description of the errors and how to correct them
+      listOfErrors:
+        items:
+        -
+          content: Descriptive link to the question with an error
+          href: '#example-error-1'
+        -
+          content: Descriptive link to the question with an error
+          href: '#example-error-1'
+  - name: error-list-is-bulleted
```
36degrees: We don't have any examples in Elements that use the bulleted list. Unless there's a good reason to encourage users to do this, I don't think we should have an example for it.

src/components/error-summary/error-summary.njk
36degrees: Can you update this example then please?

src/components/error-summary/error-summary.yaml
igloosi: i guess then it doesnt need a type after all ...

igloosi:
updated

igloosi:
back to work in progress as API needs to be updated

src/components/error-summary/error-summary.njk
```diff
@@ -1,19 +1,17 @@
...
-  classes='',
-  title='Message to alert the user to a problem goes here',
-  description='Optional description of the errors and how to correct them',
-  listClasses='',
-  listItems=[
+{{ govukErrorSummary({
+  "titleText": "Message to alert the user to a problem goes here",
+  "descriptionText": "Optional description of the errors and how to correct them",
+  "classes": "optional-extra-class",
+  "errors": [
```
36degrees: Should be `errorList`, as the arguments will be passed to the `list`component.

igloosi:
updated

igloosi:
@36degrees if you have some time today can you unblock it?

PR #288, opened 2017-10-12
Update to list component "type" parameter

igloosi:
Previously we had different boolean keys to identify different types of list:
isNumber, isBullet,  isStep and isStepLarge

This made it difficult to pass that parameter to the component from a parent component, like error-summary, as in #284 

This PR updates this to be a "type" key and we run a filter on the variable that converts it to lowercase (just in case)

### DIFF
--- a/src/components/list/template.njk
+++ b/src/components/list/template.njk
@@ -1,13 +1,14 @@
-{% if params.isBullet %}
+{% set type = params.type | lower %}
+{% if type == 'bullet' %}
 <ul class="govuk-c-list govuk-c-list--bullet
   {%- if params.classes %} {{ params.classes }}{% endif %}">
-{% elseif params.isNumber %}
+{% elseif type == 'number' %}
 <ol class="govuk-c-list govuk-c-list--number
   {%- if params.classes %} {{ params.classes }}{% endif %}">
-{% elseif params.isStep %}
+{% elseif type == 'step' %}
 <ol class="govuk-c-list govuk-c-list--icon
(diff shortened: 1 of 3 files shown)


src/components/list/template.njk
```diff
@@ -1,13 +1,14 @@
-{% if params.isBullet %}
+{% set type = params.type | lower %}
+{% if type == 'bullet' %}
```
36degrees: Would this be better as a switch statement? Could also inline the call to `lower` then.

src/components/list/template.njk
```diff
@@ -17,8 +18,8 @@
 
 {%- for item in params.items %}
   <li>
-    {% if (params.isStep) or (params.isStepLarge) %}
-    <span class="govuk-c-list__icon govuk-u-circle {% if (params.isStepLarge) %}govuk-c-list__icon--large{% endif %}">{{ loop.index }}</span> {{ item.content }}
+    {% if (type == 'step') or (type == 'step--large') %}
```
36degrees: A lot of these brackets seem unnecessary?

src/components/list/template.njk
igloosi: if only nunjucks had a switch statement...

src/components/list/template.njk
igloosi: oh yeah, not really. will remove

src/components/list/template.njk
36degrees: Dangit, nunjucks!

PR #290, opened 2017-10-13
Move grid & site-width-container styles to globals/scss/layout

igloosi:
It became apparent that grid component and site-width-container component are different in consumption to all other components in as much as they can contain nay user specified components. 

If there was any benefit of using a complicated macro vs  manually creating html grid code , it wasn't apparent.

Instead we're moving both sass files into global styles, and we're going to surface their usage some other way.
In addition we're renaming them to use our [own declared naming convention](https://github.com/alphagov/govuk-frontend/blob/master/docs/coding-standards/css.md) for layout: **govuk-o-**


nickcolley:
👍

36degrees:
Can you put some of the context into the commit message as well please?

igloosi:
done

PR #291, opened 2017-10-13
Update tag and phase banner to follow new convention

hannalaakso:
This PR:

- Ads phase tag component default and variants from YAML file. 
- Renames phase-tag to tag
- Move margin styling to phase banner
- Add inactive variant to tag
- Update phase banner and tag to use new data structure
- Add functionality to specify HTML attributes that get appended to tag and phase banner

[Trello ticket](https://trello.com/c/ZCP1a6Zo/196-2-show-default-example-and-variants-of-the-phase-tag-component)


igloosi:
this breaks phase banner. can you change phaseTagText to tag ?

igloosi:
could you rebase to 1 commit please?

hannalaakso:
Done!

igloosi:
https://github.com/alphagov/govuk-frontend/issues/292
maybe we can do this here as well.
tbc

hannalaakso:
This is now ready for review.

hannalaakso:
Review on hold - I've got a few more changes from yesterday's meeting to incorporate.

hannalaakso:
Latest changes incorporated, this is now ready for review.

src/components/tag/_tag.scss
```diff
@@ -20,5 +19,11 @@
 
     text-decoration: none;
     text-transform: uppercase;
+
+    &--inactive {
+      background-color: $govuk-grey-2;
+
```
alex-ju: There are two consecutive empty lines here resulting in a  `No trailing whitespace allowed` warning.

src/components/tag/_tag.scss
```diff
@@ -20,5 +19,11 @@
 
     text-decoration: none;
     text-transform: uppercase;
+
+    &--inactive {
+      background-color: $govuk-grey-2;
```
alex-ju: Do we have the concept of a tag being inactive? Are tags used to reflect an active/inactive status?
Using this grey shade as a background colour for a white text results in low contrast ratio (fails AA level). If we decide to use a grey maybe we should try `$govuk-grey-1` (which passes AAA level)?

src/components/phase-banner/index.njk
```diff
@@ -51,7 +51,7 @@
       ],
       [
         {
-          text: 'phaseTagText'
+          text: 'TagText'
```
alex-ju: The arguments table uses the old macro. We should update this file to use the new macro as follows:
- there is no need to import the macro on line 15 `{% from "table/macro.njk" import govukTable %}` as we already import it on the main component template
- `govukTable()` takes a single object argument

src/components/tag/_tag.scss
36degrees: Good questions – one for @timpaul perhaps?

Another place where we're using the nested / shorthand thing which makes it hard to search for things. (Should we add a standard or linting rule for this?)

src/components/tag/_tag.scss
36degrees: Where are you seeing that warning? Should it be failing CI?

src/components/tag/template.njk
```diff
@@ -0,0 +1,8 @@
+<strong class="govuk-c-tag
+  {%- if params.classes %} {{ params.classes }}{% endif %}">
+  {% if params.html %}
```
36degrees: Shorter if condition here too 🙂

36degrees:
I don't think 'add functionality to specify html attributes that get appended to the element' has been addressed in this PR?

i.e. being able to call

```
'tag' : {
    'attributes': {
        'data-variable': 'some-value',
        'aria-described-by': 'whatever'
    }
}
```

hannalaakso:
Thanks for your comments guys. This PR is now ready for review. 

RE: nesting of SASS+BEM classes, I've amended them as per your suggestion @36degrees but I've added a question to dev meetings board to discuss this so that we can settle on a convention.

src/components/phase-banner/index.njk
hannalaakso: This has been fixed in the latest commit.

36degrees:
FWIW, I don't think we need this much detail in the PR title - we should treat PRs a little like commits, a ~50 character PR title and all the detail in the PR body itself.

hannalaakso:
Good shout @36degrees. I've updated the PR name and description.

alex-ju:
I'm not sure about the commit message and the squashes. @36degrees any thoughts on this?

36degrees:
Yeah, that needs looking at. Probably better off using `fixup` than `squash` and rewording the commit message if required, using a 50 character summary for the first line summary and providing more detail as part of the description.

Give one of us a shout if you want a hand sorting that @hannalaakso 🙂

hannalaakso:
Done! 

I also removed the "inactive" tag variant to avoid introducing elements that haven't been reviewed by a designer.

I also removed the `with-html` variant of the phase banner as this functionality is now covered by the new convention.

src/components/phase-banner/phase-banner.njk
```diff
@@ -1,7 +1,13 @@
...
-    "tag": "alpha",
-    "message": "This part of GOV.UK Frontend is being built"
+    "tag":
+    {
+      "text": "alpha",
+      "html": "alpha",
+      "classes": "doing-something"
+    },
+    "text": "This part of GOV.UK Frontend is being built",
+    "html" : "This is a new service – your <a href='#''>feedback</a> will help us to improve it."
```
36degrees: It only makes sense to pass either `text` or `html`, but not both - if `html` is provided then `text` is ignored. Suggest using just the `html` argument for now.

src/components/phase-banner/phase-banner.njk
```diff
@@ -1,7 +1,13 @@
 {% from "phase-banner/macro.njk" import govukPhaseBanner %}
 {{ govukPhaseBanner(classes='',
   {
-    "tag": "alpha",
-    "message": "This part of GOV.UK Frontend is being built"
+    "tag":
+    {
+      "text": "alpha",
+      "html": "alpha",
```
36degrees: Just `text` – no need for the `html` too.

src/components/phase-banner/phase-banner.yaml
```diff
@@ -1,9 +1,9 @@
...
-  - name: with-html
-    data:
-      tag: beta
-      message: 'This is a new service – your <a href="#">feedback</a> will help us to improve it.'
+      tag:
+        text: alpha
+        html: alpha
+        classes: doing-something
+      text: This part of GOV.UK Frontend is being built
+      html: This is a new service – your <a href="#">feedback</a> will help us to improve it.
```
36degrees: Again, either `text` or `html` – not both.

hannalaakso:
Thanks @36degrees. The tag inside the phase banner reflects the structure of the actual tag component ie. the actual tag has both text and html so you can pass either through in the phase banner tag. Should the tag structure change as well?

Re: showing both options in examples, do you mean that we should have two variants (text and html variant) in yaml and print both out?

timpaul:
Hi @hannalaakso - I think grey-1 (#6F777B) with white text passes AA contrast (I tried it in the WebAIM checker).

image

hannalaakso:
@36degrees I've just clarified this with @igloosi, will update PR shortly

hannalaakso:
Thanks @timpaul you're correct. Have added the inactive tag variant back in.

@36degrees I'm now printing out both text and html variant of tags in tag itself and on phase banner. Removed the text option from phase banner so can now only use html.

igloosi:
@alex-ju  I have made some fixes, documented arguments. could you have a quick look as well?

PR #293, opened 2017-10-16
Fix generate:readme gulp task

igloosi:
When we used vinylPaths and js-yaml together in a pipe, it started
breaking with "UnhandledPromiseRejectionWarning: Unhandled promise
rejection (rejection id: 1): Error: write callback called multiple
times"

As we're already using gulp-data we can use it to pipe data from yaml to
nunjucks template, as gulp-data exposes file paths

Generate:readme task also requires `capitaliseComponentName` function as an environment filter, which is used in the nunjucks template macro

`capitaliseComponentName` is used both the the gulp task and app.js
to it makes sense to move it out to /lib folder as a helper function and 
then import it where it is needed.

### DIFF
--- a/app.js
+++ b/app.js
@@ -6,6 +6,7 @@ const path = require('path')
 const port = (process.env.PORT || 3000)
 const dto = require('directory-to-object')
 const yaml = require('js-yaml')
+const helperFunctions = require('./lib/helper-functions.js')
 
 // Set up views
 const appViews = [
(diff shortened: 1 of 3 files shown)


36degrees:
You've got some other changes bundled up in this commit (e.g. the addition of the `capitaliseComponentName` function).

igloosi:
yeah, generate: readme task won't work with out that, it's a filter in the template

36degrees:
Aah, OK. Can we avoid defining it in two places? It's already in `app.js`.

app.js
```diff
@@ -6,6 +6,7 @@ const path = require('path')
 const port = (process.env.PORT || 3000)
 const dto = require('directory-to-object')
 const yaml = require('js-yaml')
+const helperFunctions = require('./scripts/helper-functions.js')
```
36degrees: The `scripts` directory is currently used for deployment related scripts. I don't think it makes sense to put this file in there too. Suggest creating a `lib` directory and moving it there?

36degrees:
Are we still using `vinylPaths` elsewhere? If not, should we remove its dependency?

igloosi:
yeah it's being used elsewhere asset-version.js

PR #294, opened 2017-10-16
Update site-width-container use in the app

igloosi:
In #290 we changed the prefix to be `govuk-o` for layout elements (grid and side-width-container)
This PR updated the template to use the same class name. Without it the layout in the app is broken.


36degrees:
➡️ ⬅️

PR #295, opened 2017-10-16
Move variants for textarea component to YAML file

36degrees:
This PR:

- adds a YAML file where default option and variants are defined
- updates macro, template and documentation to read and display those variants

https://trello.com/c/z2wOWUFc/200-2-show-default-example-and-variants-of-the-textarea-component

### DIFF
--- a/src/components/textarea/template.njk
+++ b/src/components/textarea/template.njk
@@ -1,7 +1,13 @@
 {% from "label/macro.njk" import govukLabel %}
 
-{{- govukLabel(classes, labelText, hintText, errorMessage, id) -}}
+{{- govukLabel({
+  "labelText": params.label.labelText,
+  "hintText": params.label.hintText,
+  "errorMessage": params.label.errorMessage,
+  "classes": params.label.classes,
+  "id": params.id
+}) -}}
 
 <textarea class="govuk-c-textarea
-{%- if errorMessage %}govuk-c-input--error{% endif %}
-{%- if classes %} {{ classes }}{% endif -%}" id="{{ id }}" name="{{ name }}" rows="{%if rows %} {{- rows -}} {% else %}5{%endif %}"></textarea>
+{%- if params.label.errorMessage %} govuk-c-textarea--error{% endif %}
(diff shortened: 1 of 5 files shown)


src/components/textarea/template.njk
```diff
@@ -1,7 +1,13 @@
 {% from "label/macro.njk" import govukLabel %}
 
-{{- govukLabel(classes, labelText, hintText, errorMessage, id) -}}
+{{- govukLabel({
+  "labelText": params.label.labelText,
+  "hintText": params.label.hintText,
```
alex-ju: we might want to use safe here to avoid encoding?
`"hintText": params.label.hintText | safe`
this might be a valid question for most of the text content

src/components/textarea/template.njk
```diff
@@ -1,7 +1,13 @@
...
+  "errorMessage": params.label.errorMessage,
+  "classes": params.label.classes,
```
alex-ju: I like this way of referring attributes; we might want to extend this to other components as well.

src/components/textarea/template.njk
36degrees: I think we need to take a look at how we use safe… we're really inconsistent at the minute. I don't see the harm in escaping HTML entities.

src/components/textarea/template.njk
36degrees: It's annoying because ideally we would just be able to do e.g. `params.label + { "id": params.id }` to merge id in, and then we wouldn't need all this repetition.

36degrees:
> I would add a 'value'/content attribute as well.

Good call! How did we miss that?!

src/components/textarea/template.njk
alex-ju: Yes, we should act consistently on this matter.

src/components/textarea/template.njk
alex-ju: Yes, I thought of something similar, but I couldn't find a workaround yet.
Eventually we can create a js function and make it available across templates via addGlobal.

36degrees: I considered that – I even had a custom filter working that merged attributes in – but realised that means anyone who wanted to use our templates outside of the design system would need that extra context? Not sure it's worth it.

PR #300, opened 2017-10-19
Document how and what we use in compoment template and macros

igloosi:
In the session we identified inconsistencies in the use of parameters across all components. We identified common parameters and decided on their names and structure.
This PR is a summary of those changes and sets the guidance for future development.

### DIFF
--- a/docs/component-api.md
+++ b/docs/component-api.md
@@ -1,0 +1,128 @@
+
+# Component API definition and use
+We have chosen as Nunjucks as the templating language for GOV.UK Frontend components. We expose those templates as reusable chunks of code: macros. Developers import macros into their application, call them as per documentation and provide data to its parameters.
+
+To provide a level of consistency for developers we have standardised parameter names, their expected input, use and placement. There are expectations, and  if so they are documented accordingly.
+
+## Specifying content
+When providing *content* to a macro, say for a label or a button, we accept two parameter options:
+
+ - `text` accepts a plain string and is the default way of passing content
+ - `html` accepts html markup. In the template we will not escape html so it will be rendered. In a scenario where both text and html are set, html parameter will take precedence over text.
+
+Example:
+
+`{{ govukButton({"text": "Button text"}) }}`
+
+`{{ govukButton({"html": "Button <span class='bold'>text</span>"}) }}`
+
+Example of implementing logic in a component template:
+
+`{{ params.html | safe if params.html else params.text }}`
+
+Example shows that if `html` and `text` parameters are present, then `html` takes precedence over `text` and we are not escaping it.
+
+## Allow for text argument to be passed as a single parameter
+When the users don't want to specify extra attributes we should allow them to call a component without having to pass an entire object.
+
+Example:
+
+`govukErrorMessage("Full name must be provided")`
+
+This is possible for components that rely only on `text` and `html` attributes.
+
+We do this in the macro so that if in the future we transpile the templates we do not have to worry about mimicking this functionality in other languages.
+
+Example of implementing this logic in a component macro:
+```
+{% if params|string === params %}
+  {% set params = { text: params } %}
+{% endif %}
+```
+
+## Naming attributes
+We should use **camelCase** for naming attributes.
+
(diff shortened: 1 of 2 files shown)


docs/component-api.md
```diff
@@ -0,0 +1,112 @@
...
+When there is a need to specify html attributes, such as *checked, disabled, id, name*, etc, and they map directly, we use the same parameter name. We use boolean value to check and render the attribute.
+
+Example:
+
+`{{ govukButton({"disabled": true}) }}`
+
+`{{ govukCheckbox({"checked": true}) }}`
+
+
+## Defining additional HTML attributes
```
joelanman: Should we drop 'additional' here - don't these override any attributes set elsewhere?

36degrees: No, they'll be appended blindly to the end of the tag. If you used the `classes` argument and then provided `classes` in `attributes` as well then you'd end up with two `class` attributes on your tag.

igloosi: hopefully no one will add "class" in there, as we're making it clear what "classes" is for in macro arguments table, but :man_shrugging:

joelanman: I just tried adding two class definitions to an element

`<div class="a" class="b">`

Chrome only used the first class and ignored the second. Is that what we're expecting from the `attributes` feature?

igloosi: yes, safari, chrome and firefox just accept first class definition, which is great and works in a way that makes "atributes" work, even if they add "class" in there

joelanman: I'm just wondering if we should add attributes to the start of the tag instead of the end, so custom attributes always apply?

igloosi: I'd be inclined to say to, because then they can override our specific class, which i don't think we want. All other additional attributes will be applied.

36degrees: We can't do anything too clever here because of the limitations of nunjucks. Do we think there's a simple change we can make to the documentation to make this clearer?

docs/component-api.md
```diff
@@ -0,0 +1,112 @@
...
+When there is a need to add additional attributes to the component, we accept an ***"attributes"*** object with key : value pairs for each attribute.
+
```
igloosi: Like so?
These should not be used to overwrite already declared attributes, such as class. If there is a need for that, make use of "classes" parameter.

alex-ju:
As discussed, I would add the following section (mostly rephrased from @36degrees' commit message):

> ##Allow for text argument to be passed as a single parameter
> 
> When the users don't want to specify extra attributes we should allow them to call a component without having to pass an entire object.
> 
> Example:
> ``govukErrorMessage("Full name must be provided")``
> 
> This is only possible for component that only rely on `text` and `html` attributes.
> 
> We do this in the macro so that if in the future we transpile the templates we do not have to worry about mimicking this functionality in other languages.

igloosi:
@alex-ju  added

36degrees:
Pending any clarification about the `attributes` argument vs classes etc, I think this is looking good 👍

igloosi:
someone better with words check if this makes sense https://github.com/alphagov/govuk-frontend/pull/300#pullrequestreview-70578121

docs/component-api.md
36degrees: How about:

> You cannot use this to set attributes that are already defined, such as `class` – use the "classes" argument instead.

?

We use 'arguments' in the README for the table, so we should use that term consistently (vs parameters, attributes, etc)

igloosi:
updated

docs/component-api.md
igloosi: sounds good. will add it and check where else we use parameters and update it to arguments

docs/component-api.md
joelanman: ok yeh happy to see how this goes, I'm just wondering if there'll be any situation where someone using components really needs to override something we've done, in which case it might be better to add attributes to the start of the tag, rather than the end. But I can see that's also dangerous and can break things so lets see how it goes with current method.

timpaul: Could/should we create an explicit attribute for doing that? 'classOverride = true' kind of thing?

joelanman: thats true, we could investigate that if there's a need

PR #301, opened 2017-10-19
Update error message component to follow convention

36degrees:
- Allow error message content to be passed as either `text` or `html` rather than `errorMessage`. If HTML is provided, use with the safe filter, otherwise use text.
- Allow for extra attributes to be specified on the error message `<span>` through the `attributes` argument.
- Update the component definition (error-message.yaml) to use the new arguments
- Update the example nunjucks template (error-message.njk) to use the new arguments
- Allow for `text` to be passed as a single argument by itself without having to pass the entire object.

### DIFF
--- a/src/components/error-message/error-message.njk
+++ b/src/components/error-message/error-message.njk
@@ -2,7 +2,6 @@
 
 {{- govukErrorMessage(
   {
-    "classes": "" ,
-    "errorMessage": "Error message about full name goes here"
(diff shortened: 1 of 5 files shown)


igloosi:
LGTM, works as intended. 
are we aiming for a single commit still?

36degrees:
I think it makes sense as two commits – the first is just about improving consistency, whereas the second commit introduces a new feature.

PR #302, opened 2017-10-19
Update breadcrumb component to follow convention

igloosi:
- replace title parameter with text (as default) and html parameters, so
that either can be used.
- update template to accept text and html parameters
- replace url parameter with href and update template accordingly
- update yaml file to use new parameters
element
- update example in breadcrumb.njk file

Also update component view template where we use the breadcrumb
component, with the new syntax

### DIFF
--- a/src/components/breadcrumb/index.njk
+++ b/src/components/breadcrumb/index.njk
@@ -62,26 +62,12 @@ Please note, this component depends on @govuk-frontend/globals and @govuk-fronte
           text: 'Yes'
         },
         {
-          text: 'Breadcrumbs array with title and url keys'
+          text: 'Array of breadcrumb items'
         }
       ],
       [
         {
-          text: 'title'
-        },
-        {
-          text: 'string'
-        },
-        {
-          text: 'Yes'
-        },
-        {
-          text: 'Title of the breadcrumb item'
-        }
-      ],
-      [
-        {
-          text: 'url'
+          text: 'text'
         },
         {
           text: 'string'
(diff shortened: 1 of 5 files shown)


src/components/breadcrumb/index.njk
```diff
@@ -90,7 +90,21 @@ Please note, this component depends on @govuk-frontend/globals and @govuk-fronte
...
+          text: 'attributes'
+        },
+        {
+          text: 'object'
+        },
+        {
+          text: 'No'
+        },
+        {
+          text: 'Any extra HTML attributes (for example data attributes) to add to the breacrumb container.'
```
36degrees: brea**d**crumb

src/components/breadcrumb/index.njk
```diff
@@ -62,26 +62,26 @@ Please note, this component depends on @govuk-frontend/globals and @govuk-fronte
...
         {
           text: 'string'
         },
         {
-          text: 'Yes'
+          text: 'No'
         },
         {
-          text: 'Title of the breadcrumb item'
+          text: 'Text to use within the breadcumb item'
```
36degrees: breadc**r**umb

src/components/breadcrumb/index.njk
```diff
@@ -62,26 +62,26 @@ Please note, this component depends on @govuk-frontend/globals and @govuk-fronte
           text: 'Yes'
         },
         {
-          text: 'Breadcrumbs array with title and url keys'
+          text: 'Array of breadcumb items'
```
36degrees: Breadc**r**umb

36degrees:
Sorry, didn't pick up on this before – the `href` argument isn't documented.

igloosi:
done

PR #304, opened 2017-10-20
Update label component to follow convention

alex-ju:
- Allow label content to be passed as either `text` or `html` rather than `labelText`.
- Allow for extra attributes to be specified on the `<label>` through the `attributes` argument.
- Update the component definition (label.yaml) to use the new arguments
- Update the example nunjucks template (label.njk) to use the new arguments
- Allow for `text` to be passed as a single argument by itself without having to pass the entire object.

[Trello ticket](https://trello.com/c/bw1RwNDs)

### DIFF
--- a/src/components/label/index.njk
+++ b/src/components/label/index.njk
@@ -51,7 +51,35 @@
       ],
       [
         {
-          text: 'labelText'
+          text: 'text'
+        },
+        {
+          text: 'string'
+        },
+        {
+          text: 'No'
+        },
+        {
+          text: 'Text to use within the label'
+        }
+      ],
+      [
+        {
+          text: 'html'
+        },
+        {
+          text: 'string'
+        },
+        {
+          text: 'No'
+        },
+        {
+          text: 'HTML to use within the label. If this is provided, the text argument will be ignored.'
+        }
+      ],
+      [
+        {
+          text: 'for'
         },
         {
           text: 'string'
@@ -60,7 +88,7 @@
           text: 'Yes'
         },
         {
-          text: 'The label text'
+          text: 'The value of the for attribute, the id of the input the label is associated with'
         }
       ],
       [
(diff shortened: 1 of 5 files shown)


src/components/label/template.njk
```diff
@@ -1,14 +1,15 @@
...
+  {%- if params.for %} for="{{ params.for }}"{% endif %}>
+  {{ params.html | safe if params.html else params.text }}
   {% if params.hintText %}
   <span class="govuk-c-label__hint">
     {{ params.hintText }}
   </span>
   {% endif %}
   {% if params.errorMessage %}
-  {{ govukErrorMessage(params.errorMessage) -}}
+  {{ govukErrorMessage(params.errorMessage.text) -}}
```
36degrees: This should just be params.errorMessage so that you can pass `html` through etc.

src/components/label/template.njk
```diff
@@ -1,8 +1,9 @@
...
-<label class="govuk-c-label
-  {%- if params.classes %} {{ params.classes }}{% endif %}" for="{{ params.id }}">
-  {{ params.labelText }}
+<label class="govuk-c-label{%- if params.classes %} {{ params.classes }}{% endif %}"
+  {%- for attribute, value in params.attributes %} {{attribute}}="{{value}}"{% endfor %}
+  {%- if params.for %} for="{{ params.for }}"{% endif %}>
+  {{ params.html | safe if params.html else params.text }}
   {% if params.hintText %}
   <span class="govuk-c-label__hint">
     {{ params.hintText }}
```
36degrees: `hintHtml`?

src/components/label/index.njk
```diff
@@ -60,7 +88,7 @@
           text: 'Yes'
         },
         {
-          text: 'The label text'
+          text: 'The value of the for attribute, the id input the label is associated with'
```
36degrees: > the id **of the** input the label is associated with

?

src/components/label/template.njk
alex-ju: Yep, already fixed that :P

alex-ju:
@36degrees thanks! updated now.

PR #305, opened 2017-10-20
Update Pagination component to Previous-next and template to follow conventions

igloosi:
- replace url parameter with href and update template
- update name to PreviousNext and references to it in sass, etc
- rename previousPage and nextPage to just previous and next
- replace title parameter with text or html parameters so that either
can be used
- replace label parameter with labelText or labelHtml parameters so that
either can be used
- update template to to reflect changes above, so that text is set as
default and if html field is specified it supersedes it
- update example in previous-next.njk file
element


src/components/previous-next/template.njk
```diff
@@ -0,0 +1,37 @@
+
+<nav class="govuk-c-previous-next
+  {%- if params.classes %} {{ params.classes }}{% endif %}" role="navigation" aria-label="Previous-next"{% for attribute, value in params.attributes %} {{attribute}}="{{value}}"{% endfor %}>
```
36degrees: I don't think this is a good aria label for this component. We probably need to continue to call it 'Pagination' here?

src/components/previous-next/template.njk
```diff
@@ -0,0 +1,37 @@
...
+    <li class="govuk-c-previous-next__item govuk-c-previous-next__item--next">
+      <a class="govuk-c-previous-next__link" href="{{ params.next.href }}" rel="next">
+        <span class="govuk-c-previous-next__link-title">
+          {{ params.next.html | safe if params.next.html else params.next.text }}
+          <svg class="govuk-c-previous-next__link-icon" xmlns="http://www.w3.org/2000/svg" height="13" width="17" viewBox="0 0 17 13">
+            <path fill="currentColor" d="m10.107-0.0078125-1.4136 1.414 4.2926 4.293h-12.986v2h12.896l-4.1855 3.9766 1.377 1.4492 6.7441-6.4062-6.7246-6.7266z"></path>
+          </svg>
+        </span>
+        {% if params.next.labelText or params.next.labelHtml %}
+          <span class="govuk-c-previous-next__link-label">{{ params.next.labelHtml | safe if params.next.labelHtml else params.next.labelText }}</span>
```
36degrees: Indentation one too deep here.

igloosi:
updated

PR #306, opened 2017-10-20
Update button component to follow convention

36degrees:
- Fix a bug where buttons were actually text inputs in disguise (!) because the `type` attribute was not being set.
- Allow users to pass either html or text for both `a` and `button` type buttons.
- Allow users to choose the element they want to use for the button (input, button or a), but choose a sensible default based on usage of html or href.
- Allow users to set the name of the button or input.
- Allow users to set the type of the button or input (defaults to `submit`).
- Rename `isDisabled` argument to `disabled`.
- Rename `url` argument to `href`.
- Allow for additional attributes to be added to the button through the `attributes` argument.
- Update the component definition (button.yaml) to use the new arguments
- Update the example nunjucks template (button.njk) to use the new arguments
- Allow for text argument to be passed by itself

https://trello.com/c/R69h2v0n/260-update-button-component-api

### DIFF
--- a/src/components/button/template.njk
+++ b/src/components/button/template.njk
@@ -1,13 +1,37 @@
-{% if params.url %}
-<a class="govuk-c-button
-{%- if params.isStart %} govuk-c-button--start{% endif %}
-{%- if params.classes %} {{ params.classes }}{% endif %}" href="{{ params.url | safe }}" role="button">
-  {%- if params.text %}{{ params.text }}{% endif %}
-</a>
+{# Determine type of element to use, if not explicitly set -#}
+
+{% if params.element %}
+  {% set element = params.element %}
 {% else %}
-<input class="govuk-c-button
-{%- if params.isDisabled %} govuk-c-button--disabled{% endif %}
-{%- if params.classes %} {{ params.classes }}{% endif %}"
-{%- if params.text %} value="{{ params.text }}"{% endif %}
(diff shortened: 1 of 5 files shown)


src/components/button/template.njk
```diff
@@ -1,13 +1,37 @@
...
+  {% elseif params.html %}
+    {% set element = 'button' %}
+  {% else %}
+    {% set element = 'input' %}
+  {% endif %}
 {% endif %}
+
+{#- Define common attributes that we can use across all element types #}
+
+{%- set commonAttributes %} class="govuk-c-button{% if params.classes %} {{ params.classes }}{% endif %}{% if params.disabled %} govuk-c-button--disabled{% endif %}"{% for attribute, value in params.attributes %} {{attribute}}="{{value}}"{% endfor %}{% endset %}
```
igloosi: would it makes sense to add a check if param.href apply govuk-c-link--as-button so that in the link styles we can apply corrections? This is effectively a link variant:)

36degrees: We should be able to add corrections based on `a.govuk-c-button`?

36degrees: Sorry, just re-read this. I don't think we should see this a `link` variant, that just confuses things. It's a button that happens to be an `<a>`.

igloosi: yeah, just in terms of its BEM modifier. but can't think of any clever modifier names

PR #309, opened 2017-10-23
Update Select component to follow conventions

igloosi:
- update name to Select and references to it in sass, etc
- replace "options" array name with "items"
replace inputName parameter with name
- replace labelText and labelClasses with a label object that takes the
same parameters as the label component
add functionality to specify "selected" property in an item
- update template to to reflect changes above
- update example in select.njk file
update table of arguments to reflect changes
add functionality to specify html attributes that get appended to
the element

Trello ticket: https://trello.com/c/Hx0HRqIz/257-update-selectbox-component-api-and-name


36degrees:
You've committed all the changed READMEs as well – intentional?

igloosi:
oh no, reverting...

src/components/select/template.njk
```diff
@@ -0,0 +1,18 @@
+{% from "label/macro.njk" import govukLabel %}
+{% set labelArguments = {
+  text: params.label.text if params.label.text else params.label,
+  html: params.label.html,
+  hintText: params.hintText if params.hintText else params.hintHtml,
```
36degrees: This isn't quite right – it'd mean `hintHtml` would get passed through as `hintText` to the label component, which then wouldn't escape it.

src/components/select/template.njk
```diff
@@ -0,0 +1,18 @@
+{% from "label/macro.njk" import govukLabel %}
+{% set labelArguments = {
+  text: params.label.text if params.label.text else params.label,
```
36degrees: This won't work if `html` is set instead of `text`, as `params.label.text` will be `nil` and so the `params.label` object will be passed through as `text`.

src/components/select/template.njk
```diff
@@ -0,0 +1,18 @@
...
+  hintText: params.hintText if params.hintText else params.hintHtml,
+  for: params.id,
+  errorMessage: params.errorMessage
+} %}
+{%- if params.label %}
+{{- govukLabel(labelArguments) -}}
+{% endif -%}
+
+<select class="govuk-c-select
+  {%- if params.classes %} {{ params.classes }}{% endif %}{%- if params.errorMessage %} govuk-c-input--error{% endif %}" id="{{ params.id }}" name="{{ params.name }}{% for attribute, value in params.attributes %} {{attribute}}="{{value}}"{% endfor %}">
```
36degrees: Need a closing quote for the `name` attribute.

36degrees: Would also be good to have consistency around whitespace inside `{{` - `{{ params.id }}` vs `{{attribute}}`.

src/components/select/template.njk
igloosi: in the second example, i set html on the label, which seems to work fine

36degrees: So it does. I guess that works because `html` is set so we never output `text`.

However this still feels a little hacky to me – whilst it might work we're still doing odd things like dumping the whole label object into text, even if it's not used.

igloosi:
updated, with single argument option for label text and error message removed

src/components/select/template.njk
```diff
@@ -0,0 +1,19 @@
...
+{% set labelArguments = {
+  text: params.label.text,
+  html: params.label.html,
+  hintText: params.label.hintText,
+  hintHtml: params.label.hintHtml,
+  for: params.id,
+  errorMessage: params.errorMessage
+} %}
+{%- if params.label %}
+{{- govukLabel(labelArguments) -}}
```
36degrees: Not sure there's much benefit to defining the labelArguments outside of the call, given they're not used anywhere else.

You're also omitting some arguments from the `label` component - `classes` and `attributes` – as per https://github.com/alphagov/govuk-frontend/blob/master/src/components/file-upload/template.njk#L5-L14

src/components/select/template.njk
```diff
@@ -0,0 +1,19 @@
...
+{% endif -%}
+
+<select class="govuk-c-select
+  {%- if params.classes %} {{ params.classes }}{% endif %}{%- if params.errorMessage %} govuk-c-input--error{% endif %}" id="{{ params.id }}" name="{{ params.name }}"{% for attribute, value in params.attributes %} {{ attribute }}="{{ value }}"{% endfor %}>
+{% for item in params.items %}
+  <option value="{{ item.value }}"{%- if item.selected %} selected{% endif %} {%- if item.disabled %} disabled{% endif %}>{{ item.text }}</option>
```
36degrees: These could be e.g. `{{ ' selected' if item.selected }}` which IMO is cleaner – but less fussed about changing this.

src/components/select/template.njk
igloosi: goos shout, fixed

PR #310, opened 2017-10-23
Update link component to follow convention

alex-ju:
- Allow link content to be passed as either `text` or `html` rather than `content`.
- Allow for extra attributes to be specified on the `<a>` tag through the `attributes` argument.
- Update the component definition (link.yaml) to use the new arguments
- Update the example nunjucks template (link.njk) to use the new arguments

[Trello ticket](https://trello.com/c/XQBbTTwy)

### DIFF
--- a/src/components/link/index.njk
+++ b/src/components/link/index.njk
@@ -8,9 +8,8 @@
 {% block componentDescription %}
 Link component, with four variants:
 
-{{ govukList(
-  classes='',
-  [
+{{ govukList({
+  items: [
     {
       text: 'back link - a black underlined link with a left pointing arrow'
     },
@@ -24,10 +23,8 @@ Link component, with four variants:
       text: 'skip link - skip to the main page content'
     }
   ],
-  options = {
-    'isBullet': 'true'
-  }
-) }}
+  type: 'bullet'
+}) }}
 {% endblock %}
 
 {# defaultAndVariants #}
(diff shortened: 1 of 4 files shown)


src/components/link/index.njk
```diff
@@ -59,35 +56,35 @@ Link component, with four variants:
...
         {
           text: 'string'
         },
         {
-          text:'Yes'
+          text:'No'
         },
         {
-          text: 'The value of the link href attribute'
+          text: 'The available modifiers for the link type: --back, --muted, --download and --skip'
```
36degrees: This needs updating as they need to pass the entire class name e.g. `govuk-c-link--back`

36degrees:
The list of link types is also rendering without any text on [the preview](https://govuk-frontend-review-pr-310.herokuapp.com/components/link).

screen shot 2017-10-23 at 09 55 56

src/components/link/index.njk
alex-ju: Yes, it was the same before (need to pass the entire class), but for some reason only the modifiers were mentioned. I'll update to specify the entire class.

alex-ju:
@36degrees regarding to the list component: it's a breaking change, I know, but that's because the PR uses `text` attribute in anticipation of the soon coming list component.

alex-ju:
@36degrees updated.

src/components/link/link.njk
```diff
@@ -1,3 +1,3 @@
 {% from "link/macro.njk" import govukLink %}
 
-{{ govukLink({classes: 'govuk-c-link', href: '', content: 'Default link'}) }}
+{{ govukLink({href: '#', content: 'Default link'}) }}
```
36degrees: `content` should be `text`?

alex-ju: That's right. Corrected.

PR #314, opened 2017-10-23
Update fieldset component to follow conventions

igloosi:
- provide text and html variations for the legend and hint parameters,
so developers can use either and in the template we default to text and
mark html as safe
- update example in fieldset.njk file
- add functionality to accept an object of additional html attributes

Trello ticket: https://trello.com/c/9iL8IIzc/278-update-fieldset-component-api

### DIFF
--- a/src/components/fieldset/fieldset.njk
+++ b/src/components/fieldset/fieldset.njk
@@ -1,7 +1,9 @@
 {% from "fieldset/macro.njk" import govukFieldset %}
 
-{{- govukFieldset({
-  classes: '',
-  legendText: 'Legend text goes here'
-  })
--}}
+{{ govukFieldset({
+  "legendText": "Legend text goes here",
+  "legendHintText": "Legend hint text goes here",
+  "errorMessage": {
+    "text": "Error message goes here"
+  }
+}) }}
--- a/src/components/fieldset/fieldset.yaml
+++ b/src/components/fieldset/fieldset.yaml
@@ -1,16 +1,17 @@
 variants:
 - name: default
   data:
-    classes: ''
-    legendText: 'Legend text goes here'
-- name: with-hint-text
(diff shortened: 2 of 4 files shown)


hannalaakso:
Should there be a `classes` attribute in the component? Just noticed a small typo: `with-html-insted-of-text` should be `with-html-instead-of-text`

A more general note: What if you want to place a whole component like an input inside the fieldset? Would it better just to have fieldset as copy-paste markup for flexibility?

src/components/fieldset/fieldset.yaml
```diff
@@ -1,16 +1,17 @@
...
   data:
-    classes: ''
-    legendText: 'Legend text goes here'
-    hintText: 'Legend hint text goes here'
-    errorMessage: 'Error message goes here'
+    legendText: Legend text goes here
+    hintText: Legend hint text goes here
+    errorMessage:
+      text: 'Error message goes here'
+- name: with-html-insted-of-text
```
hannalaakso: Should be `with-html-instead-of-text`

src/components/fieldset/fieldset.njk
```diff
@@ -1,7 +1,9 @@
...
-  classes: '',
-  legendText: 'Legend text goes here'
-  })
--}}
+{{ govukFieldset({
+  "legendText": "Legend text goes here",
+  "hintText": "Legend hint text goes here",
+  "errorMessage": {
+    "text": "Error message goes here"
+  }
```
hannalaakso: Should there be a `classes` attribute?

igloosi:
you can specify classes, it's documented.
you can currently place a component inside the fieldset, as it's been done with date. there is a trello ticket to revisit this type of component (vs guidance around cop&paste markup)
I have amended the typo

PR #318, opened 2017-10-23
Update checkbox component to follow convention

igloosi:
- replace InputName parameter with name
- add idPrefix parameter which will be used if no specific id is set for
each item. if no id on items is then append loop index to idPrefix for
each item
- replace label parameter for each item with text or html parameters,
where either can be used, text is default and gets superseded by html
- for label use label component with additional govuk-c-checkbox__label
class (because of specific styling)
rename checkboxes array to items
- update template to reflect above changes
element
- update example in checkbox.njk file

### DIFF
--- a/src/components/checkbox/template.njk
+++ b/src/components/checkbox/template.njk
@@ -1,9 +1,16 @@
-{% for checkbox in params.checkboxes %}
+{% from "label/macro.njk" import govukLabel %}
+{% for item in params.items %}
 <div class="govuk-c-checkbox
-{%- if params.classes %} {{ params.classes }}{% endif %}">
-  <input class="govuk-c-checkbox__input" id="{{ params.id }}-{{ checkbox.id }}" name="{{ params.inputName }}" type="checkbox" value="{{ checkbox.value }}"
-  {%- if checkbox.checked %} checked{% endif %}
-  {%- if checkbox.disabled %} disabled{% endif %}>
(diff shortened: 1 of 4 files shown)


src/components/checkbox/template.njk
```diff
@@ -1,9 +1,17 @@
...
-{%- if params.classes %} {{ params.classes }}{% endif %}">
-  <input class="govuk-c-checkbox__input" id="{{ params.id }}-{{ checkbox.id }}" name="{{ params.inputName }}" type="checkbox" value="{{ checkbox.value }}"
-  {%- if checkbox.checked %} checked{% endif %}
-  {%- if checkbox.disabled %} disabled{% endif %}>
-  <label class="govuk-c-checkbox__label" for="{{ params.id }}-{{ checkbox.id }}">{{ checkbox.label }}</label>
+{%- if params.classes %} {{ params.classes }}{% endif %}"{% for attribute, value in params.attributes %} {{ attribute }}="{{ value }}"{% endfor %}>
+  <input class="govuk-c-checkbox__input" id="{% if item.id %}{{ item.id }}{% else %}{{params.idPrefix}}-{{loop.index}}{% endif %}" name="{{ params.name }}" type="checkbox" value="{{ item.value }}"
+  {{-" checked" if item.checked }}
+  {{-" disabled" if item.disabled }}>
+  {# <label class="govuk-c-checkbox__label" for="{{ params.id }}-{{ checkbox.id }}">{{ checkbox.label }}</label> #}
```
alex-ju: Shall we remove this line?

igloosi:
done

PR #323, opened 2017-10-24
WIP: Update table component to follow conventions

igloosi:
- remove options object and have items as individual key/value pairs
- replace cell text in head and row items with xxxText and xxxHtml
parameters, where either can be used (text as default and html marked as
safe)
element
- update example in table.njk file

Replace table of arguments structure for all components


igloosi:
@alex-ju i've update the table component and places where we use it, (including unmerged error -summary)

alex-ju:
Tested. Looks and works well. 👍 

Things that can be discussed:
- `isFirstCellHeader` is the only remaining (boolean) parameter prefixed with `is`
- we have to specify `type` (e.g. number) for each cell instead of an entire column, but I can see why that happens
- is `type` the right way to specify the content is a number? is `format` a better name

igloosi:
@alex-ju good spot:
- changed boolean to `firstCellIsHeader`
- changed `type` to` format`

For a single format definition i put the ticket in the backlog to investigate if/how it can be done

PR #325, opened 2017-10-27
Exclude .yaml files from being copied to /packages and /dist

igloosi:
They're configuration files, so there is no need for them to be in `/packages` or `/dist` folders.


36degrees:
Would it make sense to specify a whitelist rather than a blacklist at this point?

igloosi:
to say *.{js,yaml,scss.njk} and then do excludes?

36degrees:
Thinking about it a little more, I'm not sure a whitelist will work because of the fact we name all of the files using the component name. So maybe this is the right approach for now at least.

PR #327, opened 2017-10-30
Skip link component

igloosi:
As It is a unique "link" type with additional styling, it makes sense for it to be a standalone component.
This PR:
* remove the skip-link variant from the link component
* creates a separate skip-link component with a template, macro, yaml definition, styles and docs
* adds styles import to govuk-frontend.scss


src/components/skip-link/_skip-link.scss
```diff
@@ -0,0 +1,40 @@
...
+@import "../../globals/scss/vars";
+
+@include exports("skip-link") {
+  .govuk-c-skip-link {
+    background: $govuk-black;
+    text-align: center;
+  }
+
+  .govuk-c-skip-link__content-wrapper {
+    max-width: 1020px;
```
36degrees: I'm a little confused as to what this is meant to do – it sets a max width but because the link itself is positioned absolutely I don't think this actually does anything?

src/components/skip-link/_skip-link.scss
```diff
@@ -0,0 +1,40 @@
+@import "../../globals/scss/import-once";
+@import "../../globals/scss/font-face";
+@import "../../globals/scss/typography";
+@import "../../globals/scss/links";
+@import "../../globals/scss/vars";
+
+@include exports("skip-link") {
+  .govuk-c-skip-link {
+    background: $govuk-black;
```
36degrees: Why are these needed?

36degrees:
Right, sorry, they've come from govuk_template. But I don't think the styles actually make any sense there either.

src/components/skip-link/index.njk
```diff
@@ -0,0 +1,107 @@
...
+        text: 'href'
+      },
+      {
+        text: 'string'
+      },
+      {
+        text:'Yes'
+      },
+      {
+        text: 'The value of the skip link href attribute.'
```
36degrees: 'Defaults to #content'

igloosi: done

igloosi:
made it like DS, no flexbox though.

src/components/skip-link/template.njk
```diff
@@ -0,0 +1,3 @@
+<a class="govuk-c-skip-link{%- if params.classes %} {{ params.classes }}{% endif -%}"{%- for attribute, value in params.attributes %} {{ attribute }}="{{ value }}"{% endfor %} href="{%- if params.href %}{{ params.href }}{% else %}#content{% endif -%}">
```
36degrees: It's possibly a bit pedantic to call out argument ordering, but I'd usually expect `href` to be the first argument for a link – worth moving it to the front?

Could also simplify the href logic to just `{{ params.href if params.href else '#content' }}` or maybe even just `{{ params.href | default('#content') }}`

36degrees:
Also, the skip link in the Design System doesn't use flexbox…?

igloosi:
all in 
```
.tech-docs.flexbox .app-pane__header > * {
    -webkit-box-flex: 1;
    -ms-flex: 1 0 auto;
    flex: 1 0 auto;
}
```
I'm just using block :)

PR #330, opened 2017-10-31
Fix em fucntion to accept px or not

igloosi:
At the moment `em()` helper function falls down if px units are present in the argument value for either of the two arguments it accepts.

This PR fixes this issue, so the `em()` function now accepts arguments with px units and without.

See test here: https://codepen.io/anon/pen/NwqRZd


igloosi:
done

PR #332, opened 2017-11-01
Fix grid issues

igloosi:
There were discussions about making the grid more versatile and changing how it breaks on different viewports, so we held off making changes. 
This PR:
* reinstates current behaviour, where the grid columns become full width on mobile
* fixes incorrect padding between items
* creates an example grid layout page


hannalaakso:
@igloosi Is there an example in FE where we currently use the different grid classes?

igloosi:
@hannalaakso updated the PR with an example page

PR #333, opened 2017-11-01
Update generate:readme task behaviour

igloosi:
Currently the task defaults to amending files in `src/` which is not always convenient if latest changes to readme files don't exist there.

With this change, the `generate:readme` task checks for --destination flag and places the modified file to the correct directory.

### DIFF
--- a/tasks/gulp/generate-readme.js
+++ b/tasks/gulp/generate-readme.js
@@ -12,6 +12,9 @@ const nunjucks = require('nunjucks')
 const objectData = {}
 const yaml = require('js-yaml')
 const helperFunctions = require('../../lib/helper-functions.js')
+const taskArguments = require('./task-arguments')
+const gulpif = require('gulp-if')
+const isPackages = (taskArguments.destination === 'packages') || false
 
 // data variable to be passed to the nunjucks template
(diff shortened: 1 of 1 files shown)


36degrees:
I don't understand the use case for this – can you provide a little more context as to why this is needed? If we're fighting our build pipeline we should change the pipeline, not add flags to enable us to start bypassing stuff manually.

igloosi:
currently if you run generate:readme it will always update src, which is what it was designed to do so any change to docs is made. but what it doesn't do is make the change solely do dist in you run npm run build:packages as the task doesn't know about the flag passed to the build:packages task so it doesn't update files there; just in src

36degrees:
Just to check I'm understanding correctly:

- we expect just to copy the up to date README from `src` to `dist`
- we are currently not updating the READMEs in `src` because (reasons?)
- this _bypasses_ updating the files in the `src` directory so we can push the changes straight to dist?

If so, I think we should review our build pipeline so that it works for us, not start implementing workarounds that introduce complexity and different ways of building everything.

igloosi:
closing for now as major changes to build are taking place

PR #334, opened 2017-11-02
Amend image path variable definition

igloosi:
To create css image paths we use the  `file-url` helper Sass function. We mistakenly forgot to allow it to be overwritten in host app where Frontend is consumed.

This PR Fixes that.

### DIFF
--- a/src/globals/scss/_vars.scss
+++ b/src/globals/scss/_vars.scss
@@ -4,7 +4,7 @@
 @import "vars/colours";
 
 // Image paths
-$global-images: "/icons/";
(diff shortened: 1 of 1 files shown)


36degrees:
@hannalaakso there's [documentation around this at the minute](https://github.com/alphagov/govuk-frontend/blob/master/src/globals/icons/README.md), but we should explore this further.

It'd be great to look at putting assets onto a CDN, but there's a plethora of issues there around hosting, security, versioning… it's something we should be exploring but I don't think it's something we're going to be able to sort for private beta, for example.

PR #335, opened 2017-11-02
Update list component

igloosi:
When using this component in context, it became apparent that it has some styling that it too specific.

This PR:
*removes specific font-size declaration, taking the font-size of the parent where different
* removes margins, so that list has no opinions

### DIFF
--- a/src/components/list/_list.scss
+++ b/src/components/list/_list.scss
@@ -8,9 +8,7 @@
 @include exports("list") {
   .govuk-c-list {
     @include font-smoothing;
-    @include govuk-core-19;
-    // TODO: Make margins under components consistent
-    margin-bottom: $govuk-spacing-scale-4;
+    margin: 0;
     padding-left: 0;
     list-style-type: none;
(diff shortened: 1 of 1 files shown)


36degrees:
What was the context you were using this in? I think we need to revisit whether there's a difference between lists in the typographic context and how we're using them in other components.

igloosi:
in the footer.  to use it would need to override both the margin and font-size

36degrees:
I don't think that every list needs to use the list component… I _think_ the list component should be used for typographical lists within the body of the page.

Otherwise we'll end up with a generic component that doesn't actually do anything other than provide an unstyled `<ul>` or `<ol>`.

Would it make more sense to just define separately what lists look like when they are an element within the footer block?

igloosi:
yeah, definitely can be styled as part of footer. font and spacing on the list are part of the conversation with @dashouse where font was set too big for where it was used and defined spacing in has.

igloosi:
closing as it less relevant now

PR #336, opened 2017-11-03
Remove unused background image in back link component variant

igloosi:
Back link had an unused triangle background image declared, when it is actually using pseudo selectors for the left-pointing triangle.

This PR:
* removes the unused background image css property
* removes the image from the icons folder


36degrees:
I think the background image is needed for IE8 support – the back link in `master` currently renders correctly in IE8, but is broken in the preview for this PR.

This suggests we could explore wrapping it up inside an `ie-lte` mixin though?

igloosi:
hmm, need to check our linting rules, if we do : before instead of :: before then ie8 will pick it up

igloosi:
this is no longer relevant is back link is being [moved out](https://trello.com/c/8tAe1bNx/373-split-back-link-into-its-own-component) to its own component

PR #337, opened 2017-11-06
Setup test framework

alex-ju:
This PR sets up the test framework using [The GDS Way](https://github.com/maxcbc/gds-way/blob/600f30643be4438baf1b52d13d9f0896f2bfaf29/source/manuals/programming-languages/nodejs/testing.html.md.erb) recommended tools as follows:
- prepares `app.js` to be tested
- updates documentation on testing
- splits testing into 3 types:
  - App testing (Node.js/Express) `test-app`
  - Components testing `test-components`
  - Automation testing (Gulp tasks) `test-tasks`
- Adds helper functions to `/lib` for reading and comparing directories/files
- Adds testing tools dependencies
  - [Mocha](https://github.com/mochajs/mocha) as a test runner
  - [Chai](https://github.com/chaijs/chai) as assertion library
  - [Cheerio](https://github.com/cheeriojs/cheerio) for DOM parsing and manipulation
  - [Request](https://github.com/request/request) for HTTP requests
- Creates gulp tasks for testing the Node.js app (`test:app`) and the Gulp tasks (`test:tasks`)
- Adds basic app test to ensure `/` returns a list with all components
- Adds tasks tests ([Trello ticket](https://trello.com/c/hHnVtwtk)) to ensure:
  - All components are being copied to `dist` and `packages`
  - All the right files are being copied to `dist` and `packages` for each component

***Note to reviewer:** Some tests are failing. PRs will follow to fix the errors.*


docs/development-and-publish-tasks.md
```diff
@@ -59,6 +59,18 @@ This task will:
 This task will:
 - will run accessibility test (tenon and axe)
 
+**`gulp test:app`**
+
+This task will:
```
36degrees: Duplication of 'will' – which I notice is also a problem above! Would be good to fix this throughout.

lib/helper-functions.js
```diff
@@ -18,5 +21,36 @@ const componentNameToMacroName = componentName => {
...
+
+// Read components from various sources
+exports.SrcComponentList = fs.readdirSync(configPaths.src + 'components/')
+exports.DistComponentList = fs.readdirSync(configPaths.dist + 'components/')
+exports.PackagesComponentList = fs.readdirSync(configPaths.packages)
+
+// Compare component files from src with dist
+const compareComponentDist = (componentName) => {
+  let distComponentFiles = fs.readdirSync(configPaths.dist + 'components/' + componentName)
+  // We don't want to include the yaml file in dist
```
36degrees: or the readme template?

lib/helper-functions.js
```diff
@@ -18,5 +21,36 @@ const componentNameToMacroName = componentName => {
...
-
 exports.componentNameToMacroName = componentNameToMacroName
+
+// Read components from various sources
+exports.SrcComponentList = fs.readdirSync(configPaths.src + 'components/')
+exports.DistComponentList = fs.readdirSync(configPaths.dist + 'components/')
+exports.PackagesComponentList = fs.readdirSync(configPaths.packages)
+
+// Compare component files from src with dist
+const compareComponentDist = (componentName) => {
```
36degrees: I wonder about the naming of this function – it's not actually doing any 'comparison', right? It's returning an object with two arrays, that you can then compare back in the test.

Should it even be one function? I think we could split it into e.g.

```
const distFilesForComponent(componentName) => {
   return fs.readdirSync(configPaths.dist + 'components/' + componentName)
}

const expectedDistFilesForComponent(componentName) => {
  let srcComponentFiles = fs.readdirSync(configPaths.src + 'components/' + componentName)
                        .filter(file => (file !== componentName + '.yaml' && file !== 'index.njk'))
  // We expect dist to contain an html sample
  srcComponentFiles.push(componentName + '.html')
  return srcComponentFiles.sort()
}
```

Then the test would be something like:
```
let actualDistFiles = lib.distFilesForComponent(componentName)
let expectedDistFiles = lib.expectedDistFilesForComponent(componentName)
expect(actualDistFiles).to.have.same.members(expectedDistFiles)
```

Which I think is clearer?

test/app.spec.js
```diff
@@ -0,0 +1,51 @@
+/* eslint-env mocha */
+const expect = require('chai').expect
+const request = require('request')
+const cheerio = require('cheerio')
+
+const app = require('../app.js')
+const lib = require('../lib/helper-functions')
+
+const serverOptions = {
```
36degrees: I'm unclear what this is doing – is it options used to create a server used for the test, or parameters used when making a request?

test/tasks.spec.js
```diff
@@ -0,0 +1,38 @@
...
+
+      // Check component files in packages
+      it('should copy the right files to packages', () => {
+        let c = lib.compareComponentPackage(componentName)
+        expect(c.packagesComponentFiles).to.have.same.members(c.srcComponentFiles)
+      })
+    })
+  }
+
+  for (let i = 0; i < lib.SrcComponentList.length; i++) {
```
36degrees: Could this be done with a `forEach`?

lib/helper-functions.js
alex-ju: Sounds good. Will update.

test/app.spec.js
alex-ju: parameters used when making a request. I should probably make that more obvious through naming

test/tasks.spec.js
alex-ju: `for` is faster, `forEach` easier to read. If we're fine with `forEach` and use it consistently I can comply.

lib/helper-functions.js
alex-ju: Yes, if we don't want `Readme.md`s in `dist` I can exclude them. I was looking at what the tasks are supposed to be doing.

36degrees: Sorry, the readme _template_ (`index.njk`) – which is not mentioned in this comment.

test/tasks.spec.js
36degrees: I don't have strong opinions, I just worry that this bit of the test is hard to read at the minute. I'd also be tempted to rename `_Published` to make it clearer what's going on – e.g. `defineTestsForComponent` or something.

alex-ju:
@36degrees updated, please have a look when you get time.

lib/file-helper.js
```diff
@@ -0,0 +1,41 @@
...
+}
+exports.expectedDistFilesForComponent = expectedDistFilesForComponent
+
+const packagesFilesForComponent = componentName => {
+  return fs.readdirSync(configPaths.packages + componentName)
+}
+exports.packagesFilesForComponent = packagesFilesForComponent
+
+const expectedPackagesFilesForComponent = componentName => {
+  // We don't want to include the yaml file and index.njk in dist
```
36degrees: in **packages**?

lib/file-helper.js
```diff
@@ -0,0 +1,41 @@
...
+
+const fs = require('fs')
+const configPaths = require('../config/paths.json')
+
+// Read components from various sources
+exports.SrcComponentList = fs.readdirSync(configPaths.src + 'components/')
+exports.DistComponentList = fs.readdirSync(configPaths.dist + 'components/')
+exports.PackagesComponentList = fs.readdirSync(configPaths.packages)
+
+const distFilesForComponent = componentName => {
```
36degrees: I wonder if these functions would benefit from doc blocks to explain what they're doing and how they're used?

36degrees:
Would like make sure @hannalaakso has the opportunity to review this as well.

app.js
```diff
@@ -171,3 +171,5 @@ app.use(function (err, req, res, next) {
   res.status(err.status || 500)
   res.render('http-error', { error: 'Internal server error', message: err })
 })
+
+module.exports = server
```
hannalaakso: Is `server` used in `app.spec.js`? It looks like it might be defined again in that file.

lib/file-helper.js
```diff
@@ -0,0 +1,57 @@
...
+// This helper function takes a component name and returns the expected files
+// to be found in the packages folder of that component
+const expectedPackagesFilesForComponent = componentName => {
+  // We don't want to include the yaml file and index.njk in dist
+  let srcComponentFiles = fs.readdirSync(configPaths.src + 'components/' + componentName)
+                        .filter(file => (file !== componentName + '.yaml' && file !== 'index.njk'))
+  // We expect the package to contain a package.json and an html sample
+  srcComponentFiles.push('package.json')
+  srcComponentFiles.push(componentName + '.html')
+  srcComponentFiles.sort()
```
hannalaakso: This seems to sort the files twice.

lib/file-helper.js
```diff
@@ -0,0 +1,57 @@
...
+//
+// This helper function takes a component name and returns the expected files
+// to be found in the dist/components folder of that component
+const expectedDistFilesForComponent = componentName => {
+  // We don't want to include the yaml file and index.njk in dist
+  let srcComponentFiles = fs.readdirSync(configPaths.src + 'components/' + componentName)
+                        .filter(file => (file !== componentName + '.yaml' && file !== 'index.njk'))
+  // We expect dist to contain an html sample
+  srcComponentFiles.push(componentName + '.html')
+  return srcComponentFiles.sort()
```
hannalaakso: We're sorting the files here and in `tasks.spec.js`. I wonder if sorting is needed as `to.have.same.members` in Chai doesn't seem to check order, there's a separate `to.have.ordered.members`.

lib/file-helper.js
```diff
@@ -0,0 +1,57 @@
...
+// to be found in the packages folder of that component
+const expectedPackagesFilesForComponent = componentName => {
+  // We don't want to include the yaml file and index.njk in dist
+  let srcComponentFiles = fs.readdirSync(configPaths.src + 'components/' + componentName)
+                        .filter(file => (file !== componentName + '.yaml' && file !== 'index.njk'))
+  // We expect the package to contain a package.json and an html sample
+  srcComponentFiles.push('package.json')
+  srcComponentFiles.push(componentName + '.html')
+  srcComponentFiles.sort()
+  return srcComponentFiles.sort()
```
hannalaakso: This seems to sort the files twice.

app.js
alex-ju: We're exporting the server and import it in the spec as `app` on the following line
`const app = require('../app.js')`

we then save the reference in a local variable and start the server:
`server = app.listen(3000)`

and we're doing that to be able to stop the server after finishing the tests:
`server.close()`

lib/file-helper.js
alex-ju: There is no need to sort the files when looking for the same members, but it makes it a lot easier to figure the differences between ` + expected` and `- actual` when tests are failing.

lib/file-helper.js
alex-ju: Yep, sorted twice. Fixed that now.

app.js
hannalaakso: Cool that makes sense. Thanks for explaining 👍

test/tasks.spec.js
```diff
@@ -0,0 +1,42 @@
...
+    // Check to have all the components in dist and nothing else
+    it('should copy all components to dist', () => {
+      expect(lib.DistComponentList).to.deep.equal(lib.SrcComponentList)
+    })
+
+    // Check to have all the components (plus all, globals, icons) in packages and nothing else
+    it('should copy all components plus "all", "globals" and "icons" to packages', () => {
+      let expectedPackagesComponentList = lib.SrcComponentList.slice()
+      expectedPackagesComponentList.push('all', 'globals', 'icons')
+      expectedPackagesComponentList.sort()
```
hannalaakso: Could we add a comment here to say that there's no need to sort the files for `to.have.same.members` but it makes debugging failing tests a lot easier when the files are sorted?

lib/file-helper.js
hannalaakso: Aha that makes sense. I've left a comment below in the code to explain this.

test/tasks.spec.js
alex-ju: Sure. Comments added.

PR #338, opened 2017-11-07
Refactor directory-to-object in app.js

alex-ju:
This PR:
- Moves `directory-to-object` to `lib/helper-functions.js`
- Rewrites the `dto()` async calls as a generic function returning a Promises
- Waits for the `dto()` Promise to complete then passes the returned data to the view

The former helps passing [the *should display the list of components* test](https://github.com/alphagov/govuk-frontend/pull/337/files#diff-c01b3a6fb20de17b2b4e257759937692R43)

### DIFF
--- a/app.js
+++ b/app.js
@@ -4,9 +4,9 @@ const nunjucks = require('nunjucks')
 const fs = require('fs')
 const path = require('path')
 const port = (process.env.PORT || 3000)
-const dto = require('directory-to-object')
 const yaml = require('js-yaml')
 const helperFunctions = require('./lib/helper-functions.js')
+const dto = require('./lib/dto-helper.js')
 
 // Set up views
 const appViews = [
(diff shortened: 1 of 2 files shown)


app.js
```diff
@@ -38,29 +37,16 @@ app.listen(port, () => {
...
-})
-
 // Define routes
 
 // Index page - render the component list template
 app.get('/', function (req, res) {
-  res.render('component-list')
+  // Creates a componentsDirectory object representing the components in the src directory
+  // which is then passed to the view to list the components
+  helperFunctions.directoryToObject(path.resolve('./src/components'))
```
36degrees: My understanding is that the way the code previously worked was to define the directory object once when the app was started, and share that information between routes.

If I'm understanding this correctly, the directory object will now be regenerated every time the route is requested. Is that intentional, and do we understand the consequences of making this change? If generating the directory object is particularly intensive or time-consuming do we want to do this?

lib/helper-functions.js
```diff
@@ -20,3 +22,14 @@ const componentNameToMacroName = componentName => {
 }
 
 exports.componentNameToMacroName = componentNameToMacroName
+
+// Converts directory to object
+const directoryToObject = (path) => {
```
36degrees: This file's turning into a bit of a dumping ground (or at least it will have, once the test changes are merged in as well) – can we think about splitting it up or refactoring?

app.js
alex-ju: ```define the directory object once when the app was started, and share that information between routes.```
Yes that is right. There is a change that I haven't captured in my PR comment and that should be up for discussion.

My expectation is to have a list of directories that reflects the directory structure at any moment. So if I add/remove a component after the server is started I'd like the listing to reflect that, because otherwise, for example, I might click on a link for a component that no longer exists. Indeed that means reading the directory each time I access the page.

If we find that's too intensive/time-consuming and decide to read the directories only when starting the server, then I can update the code to do that.

The main changes here are actually about:
 - making that data available to specific pages
 - wait for the promise to be resolved before displaying a page

lib/helper-functions.js
alex-ju: I agree, we can split helper functions in different files inside `lib`.

app.js
joelanman: another way of approaching updating is to use something like nodemon in dev mode to restart everything when changes happen - that's the way the Prototype Kit works

36degrees: Happy with all of the above, just want to check we've thought it through.

We can always revisit if it becomes problematic.

lib/helper-functions.js
```diff
@@ -1,5 +1,3 @@
-'use strict'
```
36degrees: Intentional change?

alex-ju: No, mistake. Good catch!

alex-ju: - Re-added the `use strict` (I must have lost it when refactoring helpers 🙄).
- Added doc block

PR #340, opened 2017-11-09
Panel changes

dashouse:
**Design update**
Edited margins and padding on both the typography and the container.

### DIFF
--- a/src/components/panel/_panel.scss
+++ b/src/components/panel/_panel.scss
@@ -7,9 +7,12 @@
 @include exports("panel") {
 
   .govuk-c-panel {
-    // TODO: Make margins under components consistent
+    @include mq($until: tablet) {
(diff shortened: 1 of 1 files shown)


src/components/panel/_panel.scss
```diff
@@ -8,8 +8,12 @@
 
   .govuk-c-panel {
     // TODO: Make margins under components consistent
```
36degrees: Is this TODO still valid?

src/components/panel/panel.yaml
```diff
@@ -8,3 +8,4 @@ variants:
     classes: extra-dummy-class
     titleText: Application complete
     text: Thank you for your application
+
```
36degrees: I think your second commit is just removing the new line from the end of the file – which we don't want. Can we remove it?

PR #341, opened 2017-11-10
Include device-pixels mixin and update start button and breadcrumbs components 

dashouse:
The device-pixels mixin was left out of govuk frontend and assets were appearing pixelated on retina screens.

I have included the mixin, added the assets and changes the .scss to reference the correct files


36degrees:
Should either of these things be SVGs?

igloosi:
short of having svgs, could we avoid the device pixel mixin and always point to a 2x image?

nickcolley:
👍 for SVGs

36degrees:
Happy with this for now - let's revisit if and when we go down the SVG route. I think there will continue to be some assets where PNG makes more sense anyway.

PR #342, opened 2017-11-10
Clean dist

alex-ju:
This PR:
- Deletes orphan files (that are no longer in `src`) from renamed/removed components (date, grid, link-back, pagination, phase-tag, select-box, site-width-container)
- Removes index.html files from `dist`; we are not generating these index.html files anymore, so I figured we don't need them because we provide the same content through our express app
- Add the `clean` task to `build:dist` to prevent having orphan files in `dist`

This PR helps passing [the _should copy the right files to dist_ tests](https://github.com/alphagov/govuk-frontend/pull/337/files#diff-cca1edc01044fd5efe0013b2cdc25e64R24)


36degrees:
👏 💯 🗑 ➖ 🔻 🚫

PR #347, opened 2017-11-17
Update spacing scale and variables

dashouse:
This expands the spacing scale to cover all expected units at both mobile and desktop breakpoints.

I have also added an explicit `$govuk-spacing-scale-0` so that a utility class for applying 0 margin or padding is output.

There is no impact on components as the only breaking change was to `$govuk-spacing-scale-6`, the rest either remain the same or are additional units.

The only visible change will be on `heading-xl` (which used $govuk-spacing-scale-6`) within typography, which will be addressed in a full typography update

### DIFF
--- a/src/globals/scss/_spacing.scss
+++ b/src/globals/scss/_spacing.scss
@@ -17,13 +17,15 @@ $spacing-directions: (
 
 // Set spacing scale
 $govuk-spacing-scale: (
-  0: 0,
+  0: $govuk-spacing-scale-0,
   1: $govuk-spacing-scale-1,
   2: $govuk-spacing-scale-2,
   3: $govuk-spacing-scale-3,
   4: $govuk-spacing-scale-4,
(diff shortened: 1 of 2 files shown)


nickcolley:
Are there mockup pages that test this scale to see if it works for common use cases?

dashouse:
I've done some work done to look in to a largely uni-directional type system...From the experiments I've run with this 5 / 10 / 15 / 20 / 30 / 40 / 50 / 60 seem to cover all use cases for typographic spacing, default component margins and internal paddings for both breakpoints (that the type changes).

Part of the reason for this PR is to run further experimentation / get set up for updated typography style sheet, but as the existing variables and utility classes from GOV.UK Frontend are being used in the Design System build already I'd like to update them to the current best guess.

There will still be some need to use hard coded numbers for offsets, visual alignment or minute spacing within components (especially for adjustments for Transport) but I don't think these should become part of the scale.

PR #350, opened 2017-11-22
Fix visual regression of the error summary component

dashouse:
Update to use equal padding and tweak the spacing on the text content


36degrees:
@dashouse wanna merge this?

PR #351, opened 2017-11-24
Restructure 'globals' sass

36degrees:
The intention of this PR is to better structure the Sass we have within the globals folder to group it into layers, which will help us:

- understand and enforce what things are dependent on each other
- understand and enforce the specificity of the Sass we are writing
- have defined places for Sass with different responsibilities

We started by identifying the individual 'things' within the globals folder and putting them into logical groups. We then decided on seven layers loosely based on [ITCSS](http://www.creativebloq.com/web-design/manage-large-css-projects-itcss-101517528):

- settings
- tools
- helpers
- core
- objects
- components
- overrides

The descriptions for these layers can be found in the CSS standards documentation which is updated as part of this PR.

We still have work to do on the typography related Sass which is going to be re-done based on this new structure and the work on spacing and typography that @dashouse has done.

The font-face declaration is currently also violating the layering principles (it is included in the settings layer but relies on the `exports` mixin that lives above it in the `tools` layer) but we haven't been able to find a better solution yet.

We have also:

- updated all components to consistently import a 'common' Sass partial, which means that every component has access to all other layers. Once the typography work has been done, we should revisit this and only include the first three layers in `common`.
- removed some remaining usage of the `$gutter` variables within components
- reintroduced the concept of a gutter purely within the grid, which helps avoid coupling the grid too closely to the spacing scale (e.g. it previously 'relied' on the fact that `$govuk-spacing-scale-3` was half the size of `$govuk-spacing-scale-5`)
- rewritten the spacing overrides generation code to be easier to understand and remove a redundant `for` loop.
- updated some department colours to use govuk-blue rather than the link colour (which was govuk-blue, but this didn't make sense semantically).
- removed the border mixins which didn't seem to add any value

It's probably best to walk through this commit-by-commit when reviewing.


src/components/breadcrumb/_breadcrumb.scss
```diff
@@ -30,7 +24,7 @@
     float: left;
     background-image: file-url("separator.png");
 
-    @include device-pixel-ratio() {
+    @include govuk-h-device-pixel-ratio() {
```
alex-ju: Linter is complaining about the empty-args. We can update this line (and the other one is `_buttons.scss`) or we can adjust the linter if we're not happy with him picking that up.

src/components/button/_button.scss
```diff
@@ -112,7 +105,7 @@
     padding: em($govuk-spacing-scale-1, 24px) em($govuk-spacing-scale-6, 24px) em($govuk-spacing-scale-1, 24px) em($govuk-spacing-scale-4, 24px);
     background-image: file-url("icon-pointer.png");
 
-    @include device-pixel-ratio() {
+    @include govuk-h-device-pixel-ratio() {
```
alex-ju: Linter is complaining about the empty-args. We can update this line (and the other one is `_breadcrumbs.scss`) or we can adjust the linter if we're not happy with him picking that up.

src/globals/scss/_common.scss
```diff
@@ -0,0 +1,41 @@
...
+@import "settings/paths";
+
+// Tools
+@import "tools/exports";
+@import "tools/file-url";
+@import "tools/legacy-ie";
+@import "tools/px-to-em";
+
+// Vendor libraries
+@import "./node_modules/sass-mq/mq"; // We use sass-mq module for media queries
```
alex-ju: I have mixed feelings about this being here, but I don't see a better place right now.
I don't think the dependency on `sass-mq` is documented anywhere and it probably should.

docs/coding-standards/css.md
```diff
@@ -1,53 +1,89 @@
...
+Classes that override all other layers - for example to override the spacing or
+font size / line height on an element.
+
+- forms part of the @govuk-frontend/global package.
+- uses the prefix `govuk-!-` for its classes.
+
+# Class naming convention
 
-It can be reused and is not tied to any specific piece of UI.
+## Namespacing
```
alex-ju: I would place the Namespacing at the top, but maybe that's just me.

docs/coding-standards/css.md
```diff
@@ -1,53 +1,89 @@
...
 
-## JS hooks js-
+All classnames start with a `.govuk-` namespace to encapsulate component styling
+and prevent unnecessary style leaks. It also helps to identify ‘approved’
+components, that have been rewritten to meet the new coding conventions.
 
-Use a prefix for all JavaScript hooks.
```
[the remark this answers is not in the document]
36degrees: Sort of.

I think it's good practise for writing frontend code in general, but I am not aware that we actually use any class hooks for javascript at the minute, and _suspect_ that when we start writing Javascript we might use e.g. data attributes instead? I am wary of documenting conventions that we _might_ one day follow, but may never be relevant. But happy to revisit if we think this is a useful thing.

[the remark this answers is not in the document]
alex-ju: Sounds sensible; let's worry about it at that point.

src/components/breadcrumb/_breadcrumb.scss
36degrees: I think we should remove the empty parentheses, but I think it makes more sense to fix existing linting issues in another PR.

src/globals/scss/_common.scss
36degrees: Yeah, thinking is that it's _essentially_ part of our tools layer, it's just rather than rolling our own we're including another library. Open to suggestions, but I think we can work around this with documentation.

PR #353, opened 2017-11-27
Default component spacing

dashouse:
First pass of adding default margins to the components that require them.

The first commit is for components that belong between the `header` and the `main` content areas such as breadcrumb, back link and phase banner which all require a similar approach using both margin top **and** bottom.

The second commit is for the majority of components that will exist inside the `main` content area where a responsive margin-bottom only has been applied. Radio, Checkbox and Date Input components have been excluded as they will be affected by the update to fieldset

Regarding the change to `_breadcrumb.scss` that may look a little weird - The margin bottom is smaller than the margin-top because it is combined with the margin-bottom of `.govuk-c-breadcrumb__list-item` (required spacing for wrapping)

### DIFF
--- a/src/components/list/_list.scss
+++ b/src/components/list/_list.scss
@@ -3,8 +3,11 @@
 @include exports("list") {
   .govuk-c-list {
     @include govuk-font-regular-19;
-    // TODO: Make margins under components consistent
-    margin-bottom: $govuk-spacing-scale-4;
+    margin-top: $govuk-spacing-scale-0;
+    margin-bottom: $govuk-spacing-scale-3;
+    @include mq($from: tablet) {
+      margin-bottom: $govuk-spacing-scale-4;
+    }
     padding-left: 0;
     list-style-type: none;
     font-family: $govuk-font-stack;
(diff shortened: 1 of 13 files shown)


src/components/list/_list.scss
```diff
@@ -4,8 +4,11 @@
   .govuk-c-list {
     @include font-smoothing;
     @include govuk-core-19;
-    // TODO: Make margins under components consistent
-    margin-bottom: $govuk-spacing-scale-4;
+    margin-top: 0;
```
hannalaakso: Could this one use `$govuk-spacing-scale-0`?

src/components/phase-banner/_phase-banner.scss
```diff
@@ -3,14 +3,8 @@
 
 @include exports("phase-banner") {
   .govuk-c-phase-banner {
-    padding-top: em($govuk-spacing-scale-2, 16px);
-    padding-bottom: em($govuk-spacing-scale-2, 16px); //use to be 8
-    // Apply styling to tablet and upwards
-    // @include mq($from:tablet) {
-    //  padding-bottom: em(10px, 16px);
-    // }
+    padding: $govuk-spacing-scale-2 0;
```
hannalaakso: Similar to above about not setting left/right padding.

dashouse:
Thanks @hannalaakso - Updated

PR #354, opened 2017-11-27
Refactor typography

36degrees:
This refactors the way that typography works within Frontend to take into account Dave's work on styling prose.

As part of this work we have also:

- Adopted responsive typography 'maps' which define the way type should behave at different breakpoints in a single place, rather than multiple declarations of sizes and line heights we had in Frontend Toolkit
- Made font smoothing an integral part of the 'core' typography rather than relying on individual parts of the core or components applying the mixin
- Replaced usage of the word 'core' with 'regular' throughout
- Added new overrides for font size / line height and font weight.
- Removed 'font-size-adjust' (for now) – see commit ec4f497 for details.

## Still to do

- [x] Default line heights (and font sizes) for print media need to be reviewed by @dashouse


src/globals/scss/core/_typography.scss
```diff
@@ -1,435 +1,165 @@
...
-  }
-  font-weight: $font-weight;
-  text-transform: none;
+  // Headings
 
-  @if $is-print {
-    font-size: $font-size-print;
-    line-height: $line-height;
+  .govuk-heading-xl {
+    @include govuk-font-bold-48;
```
dashouse: I believe all of the headings, paragraph classes that use HTML elements need to be normalised. At the moment `<h1 class="govuk-heading-xl">govuk-heading-xl</h1>` will have the default margin-top applied to the `h1`

36degrees: Have pushed a further commit to remove top margins from all headings and paragraphs

src/globals/scss/core/_typography.scss
```diff
@@ -1,435 +1,173 @@
...
     }
-
   }
 
-  .heading-small {
-    @include govuk-bold-19;
-
-    margin-top: em($govuk-spacing-scale-2, 16);
-    margin-bottom: em($govuk-spacing-scale-1, 16);
+  .govuk-body-m  + .govuk-heading-l,
```
dashouse: We're missing the contextual adjustments for `govuk-body`, the alt class for `govuk-body-m`

36degrees: Good spot!

36degrees: Have pushed a commit to fix that.

src/globals/scss/core/_typography.scss
```diff
@@ -1,435 +1,173 @@
...
-      margin-top: em($govuk-spacing-scale-4, 19);
+      padding-top: $govuk-spacing-scale-4;
     }
   }
 
-  .body-text {
-    @include govuk-core-19;
-    margin-top: em($govuk-spacing-scale-1, 16);
-    margin-bottom: em($govuk-spacing-scale-4, 16);
+  .govuk-body-m + .govuk-heading-m,
```
dashouse: Apologies I missed this... this should also include `govuk-body-m` + `govuk-heading-s` and `govuk-body-s` + `govuk-heading-s`

36degrees: Pushed a fix

dashouse:
Been experimenting with this and not had any problems that haven't already been addressed in the additional commits.

Example usage:

gov uk frontend typography

36degrees:
@dashouse What do you want to do about the print styles? Specifically with regards to the line height which I've currently just defaulted to 1.

36degrees:
@hannalaakso I've rebased this on master to resolve some conflicts - the last 3 commits (
d224717, 50d469e, a4f5da1) are new since you last reviewed – could you check you're happy with them?

Then I think this is good to merge!

PR #356, opened 2017-11-27
Ensure breakpoints CSS only gets included once

36degrees:
Whilst this only appears in 'dev mode', currently this CSS is being included 38 times.

Unfortunately, we can't just wrap the import with the `exports` mixin because import directives cannot be used within control directives 😠

This is therefore a horrible, horrible hack which overrides the particular sass-mq setting if we record that we have previously imported it.

### DIFF
--- a/src/globals/scss/_common.scss
+++ b/src/globals/scss/_common.scss
@@ -20,13 +20,11 @@
 @import "tools/legacy-ie";
 @import "tools/px-to-em";
 
-// Vendor libraries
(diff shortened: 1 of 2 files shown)


36degrees:
Aah, where's the issue? I couldn't see one on their GitHub repo. Would be good to record it here for posterity.

hannalaakso:
This one: https://github.com/sass/sass/issues/1194

36degrees:
Ohh I see! Sorry, I thought you meant there was an issue open with sass-mq. That makes more sense. Still, useful to have it for reference – thanks!

PR #363, opened 2017-11-30
Remove 'Inset text' component

dashouse:
The Inset text component will be deprecated and will not be part of GOV.UK Frontend going forward. This has been discussed considerably with relevant stakeholders by @timpaul


hannalaakso:
@igloosi Are there any considerations for when we deprecate components?

igloosi:
doesn't make too much difference now as they packages are still private and under alpha no npm so they can be removed. 
as frontend now has a test to compare packages to src, i'm not sure at which point references to this component need to be removed from packages as well: if it's in this PR or before updating packages/publishing update to npm

I would guess in this PR.

igloosi:
just rebase to resolve conflict

stevenaproctor:
@dashouse Will this affect [conditionally revealing content](http://govuk-elements.herokuapp.com/form-elements/#form-toggle-content). This uses inset text in the revealed content too and includes a link to the inset text section.

dashouse:
@stevenaproctor We haven't replicated the the design of conditionally revealing content from Elements into GOV.UK Frontend yet. When we do it will be specific to form element, rather than the existing `panel` class that was used for `inset-text`.

A design suggestion we've had contributed is similar to the inset-text style but is visually more aligned to conveying the ownership of the revealed content to it's parent.

unnamed

Whether this is the final output or not, the style for the conditionally revealed content will be defined independently from the original concept of `inset-text`

PR #365, opened 2017-12-01
Fix visual regression of button component

dashouse:
This PR does fix the visual regression of the button component, <strike>however` *there are a significant number of issues with the code*.

Previously the padding / general spacing was set on the class `govuk-c-button` - however all of the button types were rendering different sizes.

This is because a button as a link `a`, an `input` and `<button>` all treat `box-sizing: border-box;` differently. Therefore each button type needs to have individual paddings defined to keep them visually aligned.

We will, therefore be unable to use the spacing scale for internal paddings of buttons as on many occasions we will need to compensate for the 2px border (albeit invisible) and create the required total height of the button (40px on normal buttons, 44px on Start buttons).</strike>

The linter has also stopped me adding `line-height: 1;` underneath the font mixin, so I have needed to add `!important`

I have removed the border bottom and reverted to box-shadow to avoid the bevelled effect on the border bottom.

### DIFF
--- a/src/components/button/_button.scss
+++ b/src/components/button/_button.scss
@@ -8,19 +8,23 @@
     display: inline-block;
     position: relative;
     width: 100%;
-    min-height: em(40px, 19px);
     margin-top: 0;
     margin-bottom: $govuk-spacing-scale-4;
     @include mq($from: tablet) {
       margin-bottom: $govuk-spacing-scale-5;
     }
-    padding: 0 em($govuk-spacing-scale-3, 19px) 0;
+    padding: ($govuk-spacing-scale-2 - $govuk-border-width-form-element) $govuk-spacing-scale-2;
     border-width: $govuk-border-width-form-element;
+    border-style: solid;
     border-radius: 0;
-    border-color: transparent transparent $govuk-button-colour-darken-15 transparent;
+    border-color: transparent;
     outline: 1px solid transparent; // keep some button appearance when changing colour settings in browsers
     outline-offset: -1px; // fixes bug in Safari that outline width on focus is not overwritten, is reset to 0 on focus in govuk_template
     background-color: $govuk-button-colour;
+    box-shadow: 0 2px 0 $govuk-button-colour-darken-15;
+    font-family: $govuk-font-stack;
+    // Had to use important because th linter won't let me put this under the font-size include
+    line-height: 1 !important;
     text-align: center;
     text-decoration: none;
     vertical-align: top;
@@ -62,7 +66,7 @@
     }
 
     &:active {
-      top: 2px;
+      top: $govuk-border-width-form-element;
       box-shadow: 0 0 0 $govuk-button-colour;
     }
 
@@ -103,9 +107,11 @@
 
   .govuk-c-button--start {
     @include govuk-font-bold-24;
-    // TODO: Check padding values
     min-height: auto;
-    padding: em($govuk-spacing-scale-1, 24px) em($govuk-spacing-scale-6, 24px) em($govuk-spacing-scale-1, 24px) em($govuk-spacing-scale-4, 24px);
+    padding-top: $govuk-spacing-scale-2 - $govuk-border-width-form-element;
+    padding-right: $govuk-spacing-scale-6;
+    padding-bottom: $govuk-spacing-scale-2 - $govuk-border-width-form-element;
+    padding-left: $govuk-spacing-scale-3;
     background-image: file-url("icon-pointer.png");
 
     @include govuk-h-device-pixel-ratio {
@@ -119,6 +125,7 @@
     @include ie-lte(8) {
       border-bottom: $govuk-border-width-form-element solid $govuk-button-colour-darken-15;
     }
+    line-height: 1 !important;
   }
 
   // making the click target bigger than the button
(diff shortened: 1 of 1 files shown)


nickcolley:
When disabling linting rules normally there's a way to do it with comments, looks like this might work for your use case. https://github.com/sasstools/sass-lint#disabling-linters-via-source

src/components/button/_button.scss
```diff
@@ -147,6 +153,26 @@
     }
   }
 
+  // Input needs less padding as does not use border-box
+
+  input[type="submit"].govuk-c-button ,
```
nickcolley: Will this introduce specificity problems which will make overriding these styles difficult?

nickcolley:
With these kinds of changes they're very precise, what stops this from breaking in the future? Should there be something like https://percy.io/ to ensure this doesn't regress?

src/components/button/_button.scss
dashouse: Right now they need to be ultra specific but really this branch was an experiment to work out why these were behaving weirdly.

Now we know why, I'm expecting all of this to be refactored. As you've mentioned below, perhaps `content-box` could align them all to have the same units.

src/components/button/_button.scss
```diff
@@ -119,6 +124,7 @@
     @include ie-lte(8) {
       border-bottom: $govuk-border-width-form-element solid $govuk-button-colour-darken-15;
     }
+    line-height: 1 !important;
```
nickcolley: Would this prevent wrapping buttons from working? If the button has a lot of text or is on a small screen?

See http://govuk-static.herokuapp.com/component-guide/button/extreme_text_start_now_button

dashouse: It's actually OK... At line-height: 1 it looks like this

screen shot 2017-12-01 at 13 46 10

I'm know seeing first hand some of the problems with having spacing on a 5px scale and font-sizes of 19 and 24. it means we have to make up the 1px somewhere. Probably from line-height.

screen shot 2017-12-01 at 13 54 34

dashouse:
@alex-ju Great stuff, might be best if you publish your branch and I'll delete this one

nickcolley:
I believe the `::before ` is used since boxshadow looks like it can be interacted with but is not, so it is an overlay to make up for that.

alex-ju:
@dashouse I pushed another commit on top of yours with the changes described above (I'll rebase it afterwords if is the right direction) so we can [test it](https://govuk-frontend-review-pr-365.herokuapp.com/components/button)

@nickcolley yep, you're right about that. now is commented, but I'll come up with with a version that won't affect `<a>`s

alex-ju:
Updated and rebased.

`.govuk-c-button:before` doesn't actually affect the height of the button, only the :hover height, which is what it's supposed to do in order to cover the border too.

36degrees:
> Should there be something like https://percy.io/ to ensure this doesn't regress?

Yes, we've been looking at Percy and similar solutions and this is likely to be on our roadmap for next quarter. We can unfortunately only do so much in one quarter!

36degrees:
I think we need to find a way to use the responsive typography mixins without setting line-height, or whilst passing custom line heights. Relying on the ordering of the declarations within the selector is really fragile.

Not suggesting that work needs to be done as this PR, but we should definitely re-visit it.

src/components/button/_button.scss
```diff
@@ -2,24 +2,30 @@
 
 @include exports("button") {
   .govuk-c-button {
+    // Box sizing does not work on the input
```
36degrees: I'm not sure what this comment is trying to tell me - what are the implications? It feels like it's missing a 'so we have to… [do something else as well]' clause.

src/components/button/_button.scss
```diff
@@ -103,9 +109,9 @@
 
   .govuk-c-button--start {
     @include govuk-font-bold-24;
-    // TODO: Check padding values
     min-height: auto;
-    padding: em($govuk-spacing-scale-1, 24px) em($govuk-spacing-scale-6, 24px) em($govuk-spacing-scale-1, 24px) em($govuk-spacing-scale-4, 24px);
+    // Total height should equal 44px
```
36degrees: Is that because of touch targets? Would be good to specify if so.

36degrees: Aah, sorry – hadn't noticed this was the start button, so it's just bigger? In which case, should we look at increasing it to 45px to fit with the general /5px rhythm that we've established?

src/components/button/_button.scss
alex-ju: It's coming from the initial commit, I should remove it now, as it's not longer the case.

36degrees: 👍

src/components/button/_button.scss
dashouse: Damn, yes it should... However because the font-size is not on the 5px scale we can't define equal spacing - same issue as the smaller buttons https://user-images.githubusercontent.com/23356842/33485592-d186ce34-d69e-11e7-8bb7-7511d3842b7f.png

So we'll need to make this 1px up with line-height which should distribute the 1px to 0.5px either side. However that means the unit-less line-height for start would be 1.0416666667 rather than 1

alex-ju:
Updated the PR so that
> compensations for the font baseline have been made in a separate, clearly marked section at the bottom of the CSS file.

nickcolley:
Nice work @alex-ju 👍 💯

PR #366, opened 2017-12-05
Update left: -9999em to use clip: rect(0 0 0 0); as visually-hidden

dashouse:
There was a thread about this on cross gov slack this morning. Noticed that this is handled differently on `govuk-h-visually-hidden` to `govuk-c-skip-link` so have swapped this out.

### DIFF
--- a/src/components/skip-link/_skip-link.scss
+++ b/src/components/skip-link/_skip-link.scss
@@ -3,14 +3,11 @@
 @include exports("skip-link") {
   .govuk-c-skip-link {
     position: absolute;
-    left: -9999em;
     padding: $govuk-spacing-scale-1 $govuk-spacing-scale-3;
+    clip: rect(0 0 0 0);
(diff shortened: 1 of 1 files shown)


36degrees:
Wonder if it's worth updating to [the version in HTML5 Boilerplate](https://github.com/h5bp/html5-boilerplate/blob/dbc3ed973573a77122f6b8a2aebd0a76a44ad6a6/dist/css/main.css#L128-L168) (which I _think_ is what inspired the version we have currently).

We could also add the `focussable` variant as a mixin and this could just `@include visuallyhidden`. That would have the advantage that we'd be have one place to update if we need to change the way we visually hide things in the future.

igloosi:
redundant due to https://github.com/alphagov/govuk-frontend/pull/369

36degrees:
Yep – forgot to come back and close this! Good call.

PR #367, opened 2017-12-05
Details component - Visual regression and design tweaks

dashouse:
e6cacf0 - Give the text a specific font colour as it's currently unspecified and appearing as red in the frontend review app.

9c759b3 - Design tweaks - Visually align the summary and details text. Align the border to the centre of the pointer when open.

**Before** 
screen shot 2017-12-05 at 13 31 58

**After**
screen shot 2017-12-05 at 13 31 28

### DIFF
--- a/src/components/details/_details.scss
+++ b/src/components/details/_details.scss
@@ -43,9 +43,8 @@
   }
 
   .govuk-c-details__text {
-    // TODO: Make margins underneath components consistent
-    margin-bottom: em($govuk-spacing-scale-3, 19px);
-    padding: em($govuk-spacing-scale-3, 19px);
+    margin-left: 3px;
+    padding: $govuk-spacing-scale-3;
     border-left: $govuk-border-width solid $govuk-border-colour;
     color: $govuk-black;
   }
--- a/src/components/details/_details.scss
+++ b/src/components/details/_details.scss
@@ -47,6 +47,7 @@
     margin-bottom: em($govuk-spacing-scale-3, 19px);
     padding: em($govuk-spacing-scale-3, 19px);
     border-left: $govuk-border-width solid $govuk-border-colour;
+    color: $govuk-black;
   }
 
   .govuk-c-details__text p {


edwardhorsford:
How does the move of the border look when combined with conditional questions (which also have a border)?

dashouse:
@edwardhorsford As the inset-text (`.panel` in Elements) component will be deprecated (https://github.com/alphagov/govuk-frontend/pull/363) the styling for how this will be applied will be defined individually on a component level, rather than using the existing `panel` class.

The suggestion of how this might be styled for conditional questions would be similar... (Design submitted by Mia)

unnamed

What do you think?

edwardhorsford:
@dashouse I like Mia's idea (possibly Kate's too).

Two things I think we'd want to consider:
- Do they look ok together? Currently nearly all our pages align to that left edge - will it look too ragged now?
- Insetting significantly more might cause issues for services who are insetting multiple times - though we'd normally advise against this (and perhaps should disallow it).

src/components/details/_details.scss
```diff
@@ -43,10 +43,10 @@
   }
 
   .govuk-c-details__text {
-    // TODO: Make margins underneath components consistent
-    margin-bottom: em($govuk-spacing-scale-3, 19px);
-    padding: em($govuk-spacing-scale-3, 19px);
+    margin-left: 3px;
```
36degrees: Has this change been tested across all browsers? I'm wary of trying to align with the toggle as I think it's part of the default browser styling for the details component, and will vary between browsers?

If it does work, it would be good to add a comment to explain what this is doing.

36degrees:
Good spot on the text colour BTW – must've missed it as I never expanded it! But shows the red text hack is doing its job!

dashouse:
@edwardhorsford I think the parent element still being aligned to the hard left wouldn't cause it to feel ragged (i'll do a test and stick it in here). In terms of conditional content, the intent would be help to define the ownership of the parent element to the conditional content, so the indent and margin-bottom would help with this.

If this change has too many potential issues I'll roll back on those commits and just fix the text colour and try to align the content.

dashouse:
@edwardhorsford 

Centred to component - Components look better individually but ragged when conditionally revealed content within radios is on the same screen as `<details>`
screen shot 2017-12-06 at 14 42 40

Left aligned - Components look "worse" individually but more consistent when on the same page
screen shot 2017-12-06 at 14 43 57

src/components/details/_details.scss
dashouse: Well it appears we've broken it in Firefox anyway...
screen shot 2017-12-06 at 17 01 47

edwardhorsford:
Thanks for sharing these @dashouse. Given some of these changes are non-controversial (text colour and aligning text), perhaps they should be broken out as a separate PR?

dashouse:
I'm deleting this branch and starting over as there's also an issue with firefox that needs addressing.

joelanman:
My preference: the centre aligned ones look better to me. It's not the majority case to have multiple different sections open like that. Even when they are, it might be technically aligned in the left aligned case, but it looks more aesthetically pleasing to me to have them centre aligned. They're always not 'aligned' since the radios and the details arrow are different widths, so the text next to them doesn't align, regardless of the opening panels.

PR #368, opened 2017-12-06
Add new point to spacing scale. Create new responsive spacing scale.

hannalaakso:
This PR:
- adds a new point 25px to spacing scale which was missing to make it consistent and updates globals and components accordingly to maintain previously set spacing
- sets spacing scale points in maps to allow us to map relationships between spacing points across different breakpoints
- updates `generate-spacing-overrides` mixin to use the maps and to produce spacing across viewpoints, essentially making the spacing overrides responsive
- creates `govuk-spacing-unit` function that allows a single point to be retrieved from a spacing map
- creates ` govuk-spacing-responsive` mixin to produce responsive spacing when given a spacing map - this will standardise spacing for components and global styles

I've worked with @dashouse to visually verify the accuracy of the new spacing styles on global and components. Also, we decided to comment out the negative margin styles for contextual headings in core/typography as they appear not to be needed after all. These might go back in later or get removed altogether.

### DIFF
--- a/src/components/panel/_panel.scss
+++ b/src/components/panel/_panel.scss
@@ -6,11 +6,11 @@
     @include govuk-font-regular-19;
 
     @include mq($until: tablet) {
-      padding: $govuk-spacing-scale-5;
+      padding: $govuk-spacing-scale-6;
     }
 
     margin-bottom: $govuk-spacing-scale-3;
-    padding: $govuk-spacing-scale-6;
+    padding: $govuk-spacing-scale-7;
 
     text-align: center;
   }
@@ -22,7 +22,7 @@
 
   .govuk-c-panel__title {
     margin-top: 0;
-    margin-bottom: $govuk-spacing-scale-5;
+    margin-bottom: $govuk-spacing-scale-6;
 
     @include govuk-font-bold-48;
   }
--- a/src/globals/scss/core/_typography.scss
+++ b/src/globals/scss/core/_typography.scss
@@ -7,11 +7,7 @@
     display: block;
 
     margin-top: 0;
-    margin-bottom: $govuk-spacing-scale-6;
-
-    @include mq($from: tablet) {
-      margin-bottom: $govuk-spacing-scale-8;
-    }
+    @include govuk-responsive-margin($govuk-spacing-responsive-8, "bottom");
   }
 
   .govuk-heading-l {
@@ -21,11 +17,8 @@
     display: block;
 
     margin-top: 0;
-    margin-bottom: $govuk-spacing-scale-4;
+    @include govuk-responsive-margin($govuk-spacing-responsive-6, "bottom");
 
-    @include mq($from: tablet) {
-      margin-bottom: $govuk-spacing-scale-6;
-    }
   }
 
   .govuk-heading-m {
@@ -35,11 +28,7 @@
     display: block;
 
     margin-top: 0;
-    margin-bottom: $govuk-spacing-scale-3;
-
-    @include mq($from: tablet) {
-      margin-bottom: $govuk-spacing-scale-4;
-    }
+    @include govuk-responsive-margin($govuk-spacing-responsive-4, "bottom");
   }
 
   .govuk-heading-s {
@@ -47,11 +36,7 @@
     @include govuk-font-bold-19;
 
     margin-top: 0;
-    margin-bottom: $govuk-spacing-scale-3;
-
-    @include mq($from: tablet) {
-      margin-bottom: $govuk-spacing-scale-4;
-    }
+    @include govuk-responsive-margin($govuk-spacing-responsive-4, "bottom");
   }
 
   // Captions to be used inside headings
@@ -93,11 +78,7 @@
     @include govuk-font-regular-24;
 
     margin-top: 0;
-    margin-bottom: $govuk-spacing-scale-4;
-
-    @include mq($from: tablet) {
-      margin-bottom: $govuk-spacing-scale-6;
-    }
+    @include govuk-responsive-margin($govuk-spacing-responsive-6, "bottom");
   }
 
   .govuk-body-m {
@@ -105,11 +86,7 @@
     @include govuk-font-regular-19;
 
     margin-top: 0;
-    margin-bottom: $govuk-spacing-scale-3;
-
-    @include mq($from: tablet) {
-      margin-bottom: $govuk-spacing-scale-4;
-    }
+    @include govuk-responsive-margin($govuk-spacing-responsive-4, "bottom");
   }
 
   .govuk-body-s {
@@ -117,11 +94,7 @@
     @include govuk-font-regular-16;
 
     margin-top: 0;
-    margin-bottom: $govuk-spacing-scale-3;
-
-    @include mq($from: tablet) {
-      margin-bottom: $govuk-spacing-scale-4;
-    }
+    @include govuk-responsive-margin($govuk-spacing-responsive-4, "bottom");
   }
 
   .govuk-body-xs {
@@ -129,11 +102,7 @@
     @include govuk-font-regular-14;
 
     margin-top: 0;
-    margin-bottom: $govuk-spacing-scale-3;
-
-    @include mq($from: tablet) {
-      margin-bottom: $govuk-spacing-scale-4;
-    }
+    @include govuk-responsive-margin($govuk-spacing-responsive-4, "bottom");
   }
 
   // Usage aliases
@@ -149,21 +118,6 @@
     @extend .govuk-body-m;
   }
 
-  // Contextual adjustments
-
-  // Add a negative top margin to headings that appear directly after other
-  // headings
-
-  .govuk-heading-l + .govuk-heading-m {
-    margin-top: -$govuk-spacing-scale-1;
-  }
-
-  .govuk-heading-m + .govuk-heading-s {
-    margin-top: -$govuk-spacing-scale-2;
-  }
-
-  // Add top padding to headings that appear directly after paragraphs
-
   .govuk-body-l  + .govuk-heading-l {
     padding-top: $govuk-spacing-scale-1;
 
@@ -174,11 +128,7 @@
 
   .govuk-body-m  + .govuk-heading-l,
   .govuk-body-s  + .govuk-heading-l {
-    padding-top: $govuk-spacing-scale-3;
-
-    @include mq($from: tablet) {
-      padding-top: $govuk-spacing-scale-4;
-    }
+    @include govuk-responsive-padding($govuk-spacing-responsive-4, "top");
   }
 
   .govuk-body-m + .govuk-heading-m,
--- a/src/globals/scss/core/_typography.scss
+++ b/src/globals/scss/core/_typography.scss
@@ -7,10 +7,10 @@
     display: block;
 
     margin-top: 0;
-    margin-bottom: $govuk-spacing-scale-5;
+    margin-bottom: $govuk-spacing-scale-6;
 
     @include mq($from: tablet) {
-      margin-bottom: $govuk-spacing-scale-7;
+      margin-bottom: $govuk-spacing-scale-8;
     }
   }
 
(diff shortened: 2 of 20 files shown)


src/globals/scss/settings/_spacing.scss
```diff
@@ -1,16 +1,67 @@
...
+// You can define different behaviour on tablet and desktop. The 'null'
+// breakpoint is for mobile.
+
+// To use the points on the spacing scale:
+//  Combine them with 'govuk-spacing-unit' or 'govuk-spacing-responsive' mixins
+// (see 'helpers/spacing').
+
+$govuk-spacing-scale-0: (
+  null: 0,
+  tablet: 0
```
36degrees: These could be omitted for 0 thru 3? Avoids generating media queries that don't do anything.

src/globals/scss/core/_typography.scss
```diff
@@ -154,41 +123,37 @@
...
   // headings
 
-  .govuk-heading-l + .govuk-heading-m {
-    margin-top: -$govuk-spacing-scale-1;
-  }
-
-  .govuk-heading-m + .govuk-heading-s {
-    margin-top: -$govuk-spacing-scale-2;
-  }
+  // .govuk-heading-l + .govuk-heading-m {
```
36degrees: Let's not leave commented out code in here – there's a good chance it'll just hang around forever. Delete it - if we need to bring it back, that's what version control is for 🙂

src/globals/scss/settings/_spacing.scss
36degrees: Although… that would break the `govuk-spacing-unit` function… 🤔

src/globals/scss/helpers/_spacing.scss
```diff
@@ -0,0 +1,42 @@
+// Get single spacing unit. We use "tablet" values from settings/spacing as they
+// cascade consistently. However, there is a naming disconnect here as it feels
+// odd to fetch "tablet" values specifically. We should discuss naming in the
+// spacing maps.
+// Arguments:
+// $negative == true returns a negative unit.
+@function govuk-spacing-unit($spacing-map, $negative: false) {
```
36degrees: I'm not sure how I feel about this. It still feels kinda weird to have a scale that's both responsive and non-responsive in this way. And as I commented in the settings file, it introduces a fragile convention that means you always have to set a tablet breakpoint, even if it doesn't change from mobile.

I'm not sure I have a better solution. Other than to be more explicit and have a scale and a responsive scale, or… something.

src/globals/scss/core/_typography.scss
```diff
@@ -6,12 +6,8 @@
 
     display: block;
 
-    margin-top: 0;
-    margin-bottom: $govuk-spacing-scale-5;
-
-    @include mq($from: tablet) {
-      margin-bottom: $govuk-spacing-scale-7;
-    }
+    margin-top: govuk-spacing-unit($govuk-spacing-scale-0);
```
36degrees: I think the benefit from having 0 in the 'scale' is that we make sure we generate overrides for it, but it seems kinda weird to add this extra complexity when `margin-top: 0` should (?) always have the same effect? Thoughts?

src/globals/scss/helpers/_spacing.scss
```diff
@@ -0,0 +1,42 @@
...
+  @return $unit;
+}
+
+
+// Get responsive spacing maps as set in settings/spacing. These maps should be
+// used wherever possible to standardise responsive spacing.
+// Arguments:
+// $type = margin | padding
+// $property = top \ right \ bottom \ left \ all
+@mixin govuk-spacing-responsive($spacing-map, $type, $property: "all", $important: false) {
```
36degrees: This is a really nice solution, but I wonder if it'd be 'friendlier' for contributors to have two mixins `govuk-margin-responsive` and `govuk-padding-responsive`? I think involving as few arguments as possible would make it easier to remember, but on the downside, we'd need to define two almost identical mixins.

src/globals/scss/overrides/_spacing.scss
```diff
@@ -14,25 +14,49 @@ $spacing-directions: (
...
+//
 // .govuk-\!-mt-1 {
 //   margin-top: [whatever spacing point 1 is...]
 // }
 
 @mixin generate-spacing-overrides($property, $property-shorthand) {
   // For each point in the spacing scale (defined in settings), create an
   // override that affects all directions...
-  @each $scale-point, $scale-value in $govuk-spacing-scale {
+  @each $scale-point, $scale-map in $new-govuk-spacing-scale {
```
36degrees: Can this function be update to use the `govuk-spacing-responsive` mixin? There's a lot of duplication between the two mixins.

36degrees:
Meant to add – comments and commit message etc are ace 👌

src/globals/scss/core/_typography.scss
hannalaakso: 👍  Code taken out.

src/globals/scss/overrides/_spacing.scss
hannalaakso: Good call! Done.

src/globals/scss/core/_typography.scss
hannalaakso: Agreed. I discussed with @dashouse and he saw no real benefit here either. Have removed all zero variables for margin/padding from style declarations.

src/globals/scss/settings/_spacing.scss
hannalaakso: Might you want to use a media query with zero padding/margin to override existing styling to be zero on all breakpoints? It's a weird use case. I've taken out the zero declaration from the base scale but left it in the responsive scale.

src/globals/scss/helpers/_spacing.scss
hannalaakso: Done. I think this was a really good shout because even though it makes the mixin names longer, it somewhat clarifies what they do.

hannalaakso:
The solution is in line with what we agreed with the team last week we’d try out: one scale and a mixin to extract single values from it. I agree though that `govuk-spacing-unit` feels like it’s hiding functionality. I’ve taken it out and replaced it with two scales.

Open to discussing the naming of `$govuk-spacing-scale-` and `$govuk-spacing-mq-` scales. I tried to make the latter scale sound similar to the mixins eg. `govuk-spacing-mq-margin` that go with it to make it easier to distinguish which scale is purely values on a scale and which needs a mixin to implement it.

Whether `$govuk-spacing-mq-` scale should use `$govuk-spacing-scale-` is up for discussion: it would introduce consistency but couple them really tightly. Also keeping the values separate makes `$govuk-spacing-mq-` a whole lot more readable. Thoughts? 

I’ll rebase these into a single (or so) commit but leaving them as separate for now to make it easier to track changes.

I agree that we should look at how the `em` now gets used. If it makes sense I can look at that that in a separate PR as it's about font-sizing.

src/globals/scss/helpers/_spacing.scss
hannalaakso: See my comment below :)

src/globals/scss/settings/_spacing.scss
36degrees: Ooh, interesting – I hadn't thought of that. I would imagine that because the overrides use `!important` they'd trump even media queries, but I don't actually know what the behaviour would be.

src/components/panel/_panel.scss
```diff
@@ -3,13 +3,13 @@
 @include exports("panel") {
 
   .govuk-c-panel {
+    margin-bottom:  $govuk-spacing-scale-3;
```
36degrees: Very minor point (I'm surprised the linter didn't pick it up) but there's an extra space here.

src/globals/scss/helpers/_spacing.scss
```diff
@@ -0,0 +1,39 @@
+// Create responsive spacing as set in settings/spacing.
+
+// Create responsive margins
+// Arguments:
+// $value = top \ right \ bottom \ left \ all
+@mixin govuk-spacing-mq-margin($scale-map, $value: "all", $important: false) {
```
36degrees: Do we think we need both `spacing` and `margin` in the mixin name? Wonder if `govuk-margin-mq` would be clear enough?

src/globals/scss/overrides/_spacing.scss
```diff
@@ -14,25 +14,30 @@ $spacing-directions: (
...
   // For each point in the spacing scale (defined in settings), create an
   // override that affects all directions...
-  @each $scale-point, $scale-value in $govuk-spacing-scale {
+  @each $scale-point, $scale-map in $govuk-spacing-mq-scale {
+
     .govuk-\!-#{$property-shorthand}-#{$scale-point} {
-      #{$property}: $scale-value !important;
+
+      @if $property == "all" {
+        @include govuk-spacing-mq($scale-map, $property, "all", true);
```
36degrees: 👌

hannalaakso:
Totally agree, and sorry if that wasn't reflected in the original review. Thanks for revisiting it. When we discussed it before it sounded OK in the abstract, but once actually in the code it just felt like it would be hard for other developers to follow. Are you happy with what we've ended up with?
> Yes - two scales create a bit more code in "settings/spacing" but the experience for the user will be simplified 👍 

The consistency between the scale and the mixins is great, but I do worry a little that we now have -responsive typography mixins and -mq spacing mixins doing broadly the same thing. The -mq suffix is nice and short, but the -responsive mixin is possibly more descriptive? Do you think we should standardise?
> I toyed with this idea too but `govuk-spacing-responsive-margin` felt like too long a mixin name - much happier with `govuk-responsive-margin` as per your comment above that we should get rid of the word "spacing" in the names. Have just done this.

Another thing we could look at doing in the future is throwing sass errors / warnings if e.g. you try to use the responsive mixins with a single value, so if users do confuse the two scales we can try and provide helpful feedback.
> Yes this is a good idea. Have raised a [ticket](https://trello.com/c/OjWgpYkt/463-throw-errors-when-fe-spacing-mixins-used-wrong).

I think it makes sense as it is… though if we did want to change the mq- scale to reference the static scale I'd suggest just using it for the tablet values which do actually correlate, and leave the mobile values in px? That'd make the relationship 'clearer' (the tablet breakpoints follow the static scale) whilst not coupling them too closely?
> Yes when I mentioned coupling them, that's what I meant too, ie. leave mobile values in px. But I think as well that they make sense better the way they are. I'll wait to see if @alex-ju has any thoughts on this.

src/components/panel/_panel.scss
hannalaakso: Done! Yes it's odd that the linter doesn't pick up on this type of spacing issues.

src/globals/scss/helpers/_spacing.scss
hannalaakso: Fixed! See my comment below.

alex-ju:
That was quite a long conversation to follow 😄 
I discussed the main point yesterday with @hannalaakso and I'm happy with how it looks now. We can discuss naming further, but as long as we keep a balance between 'descriptive', 'related' and 'length' - and I see we tried our best to do that - I'm happy with it.

src/globals/scss/overrides/_spacing.scss
```diff
@@ -14,25 +14,30 @@ $spacing-directions: (
...
 //   margin-top: [whatever spacing point 1 is...]
 // }
 
 @mixin generate-spacing-overrides($property, $property-shorthand) {
   // For each point in the spacing scale (defined in settings), create an
   // override that affects all directions...
-  @each $scale-point, $scale-value in $govuk-spacing-scale {
+  @each $scale-point, $scale-map in $govuk-spacing-responsive-scale {
+
     .govuk-\!-#{$property-shorthand}-#{$scale-point} {
```
36degrees: We should probably have something in the class names to mark these as responsive overrides?

hannalaakso: As just discussed, we now have an "r" to indicate responsiveness.

PR #369, opened 2017-12-06
Update visually hidden helper and skip link

36degrees:

Proposed as alternative to #366


dashouse:
Visually approved

PR #370, opened 2017-12-06
Remove font face and weight declarations

36degrees:
The font-face is being set twice in a lot of the components - once by the typography mixin and once in the component itself. This removes the duplication by removing all font-face declarations.

It also replaces font-weight declarations with the appropriate mixin, so that e.g if in the future we choose to re-introduce font-size-adjust we can ensure that it’s applied correctly according to the font-weight.

Finally, it moves all typography mixin includes to the top of their respective rulesets.

### DIFF
--- a/src/components/breadcrumb/_breadcrumb.scss
+++ b/src/components/breadcrumb/_breadcrumb.scss
@@ -2,21 +2,22 @@
 
 @include exports("breadcrumb") {
   .govuk-c-breadcrumb {
+    @include govuk-font-regular-16;
     @include govuk-text-colour;
+
     margin-top: $govuk-spacing-scale-3;
     margin-bottom: $govuk-spacing-scale-2;
-    font-family: $govuk-font-stack;
   }
 
   .govuk-c-breadcrumb__list {
     @include govuk-h-clearfix;
+
     margin: 0;
     padding: 0;
     list-style-type: none;
   }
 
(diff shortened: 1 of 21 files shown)


36degrees:
I think this is likely to conflict with #368 – happy to wait for that to be merged and deal with the conflicts in this branch if you prefer @hannalaakso?

36degrees:
One change that's worth being explicit about – I've started moving `includes` with no `content` (i.e. not `mq` to the top of the rulesets. I've done this because I think:

- it makes it more readable, rather than having them mixed in with declarations.
- it fits with my mental model of how components should be built – build first using the 'common stuff' (which will generally come from mixins) and then layer on the more custom styling after.

Doing a bit of reading a [number](https://sass-guidelin.es/#css-ruleset) [of](https://www.sitepoint.com/css-sass-styleguide/#sass-stuff) [styleguides](https://anotheruiguy.gitbooks.io/sassintherealworld_book-i/aLittleUnderTheHood/using-mixins.html) agree with grouping them in this way, though Airbnb [suggest they go last](https://github.com/airbnb/css#ordering-of-property-declarations).

Are we happy with this approach? If we want to adopt it formally, shall I make another PR to add it to our code conventions?

igloosi:
as the default font stack is set in the mixin this is fine. regarding moving to the top: it's useful and makes sense of you do need to override common stuff. do we need to update lint rules as well?

PR #372, opened 2017-12-08
Style details component correctly, use border-based triangles for marker.

36degrees:
This correctly sets the text colour for the disclosed text and adjusts the spacing so that the disclosed text is aligned with the summary.

It also styles the marker (arrow) using border-based triangles.

This allows us to consistently style the marker across browsers and resolves the issue with the Firefox marker disappearing because of our use of `display: inline-block`. For more details, see the implementation.md file added as part of this commit.

We also use clip-path to ensure that if browser colours are overridden the marker still appears as a triangle rather than turning into a square.

### Collapsed
screen shot 2017-12-08 at 09 18 01gmt

### Expanded 
screen shot 2017-12-08 at 09 18 04gmt

### Focussed
screen shot 2017-12-08 at 09 18 08gmt

## IE9

bs_win7_ie_9 0

## IE10

bs_win7_ie_10 0

## IE11

bs_win7_ie_11 0

## Edge

bs_win10_edge_16 0

## Android 
bs_realdroid_mobile_google pixel-8 0-1080x1920

## iOS
bs_realios_mobile_iphone 6s-9 0

## Windows Phone
bs_winphone_mobile_nokia lumia 930-8 1-1080x1920

## Firefox (custom colours)

screen shot 2017-12-08 at 09 19 32gmt

### DIFF
--- a/src/components/details/_details.scss
+++ b/src/components/details/_details.scss
@@ -1,24 +1,35 @@
 @import "../../globals/scss/common";
 
 @include exports("details") {
+
   .govuk-c-details {
     @include govuk-font-regular-19;
-
-    display: block;
+    @include govuk-text-colour;
     @include govuk-responsive-margin($govuk-spacing-responsive-6, "bottom");
 
-    clear: both;
+    display: block;
   }
 
   .govuk-c-details__summary {
+    // Make the focus outline shrink-wrap the text content of the summary
     display: inline-block;
 
+    // Absolutely position the marker against this element
     position: relative;
-    // TODO: Make margins under components consistent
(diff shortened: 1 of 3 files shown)


igloosi:
this looks very good. only point is if this shape can/should/will be used anywhere else, should we stick it in globals/objects/shapes?

36degrees:
We wouldn't be able to use it as an object in the traditional sense as you can't add a class to the pseudo element, but it might make a sensible helper mixin. However, I'd suggest we wait until we find ourselves duplicating the functionality and _then_ abstract it, rather than doing it prematurely based on that assumption? (I think it's quite likely – I think we can use this approach for the back link, for example)

PR #373, opened 2017-12-08
Add "Layout wrappers" to objects

dashouse:
This addresses a need that has come from the Typography changes (headings with margin-bottom only) and the `#content` wrapper that existed as part of GOV.UK Elements.

**Here's an illustration of what these wrappers are doing**
govuk-frontend-layout-wrapper


src/globals/scss/objects/_site-width-container.scss
```diff
@@ -20,7 +20,8 @@ $govuk-site-width: 960px;
 
 @include exports("site-width-container") {
 
-  .govuk-o-site-width-container {
+  .govuk-o-site-width-container,
+  .govuk-o-wrapper {
```
36degrees: Should we just get rid of `site-width-container` and rename the file?

src/globals/scss/objects/_content-wrapper.scss
```diff
@@ -0,0 +1,20 @@
...
+  // <div class="govuk-o-wrapper">
+  //   Breadcrumbs, Phase banners, Back links are placed in here.
+
+  //   <div class="govuk-o-content">
+  //       Wrapper for the main content of your page which applies padding to the top / bottom
+  //   </div>
+
+  // </div>
+
+  .govuk-o-content {
```
36degrees: Would be good to make this consistent with the filename (or make the filename consistent with this).

nickcolley:
Is there a use case where you don't want the headers to have margin between themselves and the content?

36degrees:
I think the issue is that multiple different 'things' comprise what we consider the 'header' margins wise – the header itself, cookie banner, phase banner, back link, breadcrumb, possibly secondary navigation… so I think wrapping the 'content' in some sort of wrapper is the only viable solution.

dashouse:
@nickcolley Not if you're using all of the new stuff... Because the typography update has removed all of the `margin-top`s on headings this wrapper is necessary to achieve the space that currently set by the h1.

As a collection of things it makes sense, if it doesn't make sense for a particular use case you just won't need to use the content wrapper, but to replicate current designs you will

nickcolley:
I was just wondering if the header itself should have the margins, but I guess because of the widths involved that'd be difficult to do, 👍

dashouse:
@nickcolley Initially tried putting the margin on the header only but the 3 or 4 things that sit in between the header and the main content need to butt up to the blue bar. I think this a relatively simple solution that will allow for breadcrumbs / back link etc or not without having to adjust spacing.

igloosi:
note: DS and PK currently have site-width-container so will need to change

src/views/component-list.njk
```diff
@@ -1,15 +1,17 @@
...
-  </h1>
-  <ul class="govuk-c-list">
-  {% for componentName, value in componentsDirectory %}
-    <li><a href="/components/{{ componentName }}">{{ componentName | replace("-", " ") | capitalize }}</a></li>
-  {% endfor %}
-  </ul>
-  <h2 class="govuk-heading-m">
-    <a href="/examples">View examples</a>
-  </h2>
+  <div class="govuk-o-content">
```
36degrees: This should probably be added to the layout rather than individual views?

PR #374, opened 2017-12-11
Update override classes in examples/typography.

hannalaakso:
This was missed out from #368 as the typography classes changed.

Also update docs on override classes.

### DIFF
--- a/docs/coding-standards/css.md
+++ b/docs/coding-standards/css.md
@@ -72,7 +72,8 @@ Classes that override all other layers - for example to override the spacing or
 font size / line height on an element.
 
 - forms part of the @govuk-frontend/global package.
-- uses the prefix `govuk-!-` for its classes.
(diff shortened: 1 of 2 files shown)


dashouse:
I think these might be using the previous values (as they weren't changed after the spacing scale changes) so anything above the 45px unit will need to go up one in the scale. That said, this is just for the review page so I'm not too fussed :)

hannalaakso:
That's a valid point. I've just pushed an amend if you don't mind reviewing that.

dashouse:
@hannalaakso Perfect, looks great

docs/coding-standards/css.md
```diff
@@ -72,7 +72,8 @@ Classes that override all other layers - for example to override the spacing or
 font size / line height on an element.
 
 - forms part of the @govuk-frontend/global package.
-- uses the prefix `govuk-!-` for its classes.
+- uses the prefix `govuk-!-` for its classes, eg. `govuk-!-mt-r8`; `r` indicates the class is responsive and `8` is the scale point; these are set in settings/spacing
```
36degrees: 👌

PR #375, opened 2017-12-11
Allow radio and checkboxes to have fieldsets

36degrees:
By passing a fieldset object as part of the component definition a fieldset will automatically be added that wraps the radio buttons or checkboxes.

To be consistent, this also makes fieldsets optional for the date input component.

Finally, it adds styling to reduce the margins of headings used when inside fieldset legends.

https://trello.com/c/HT8kvhSM/415-3-add-optional-fieldset-to-radio-checkbox-date-input-macro
https://trello.com/c/zOsvZn6P/384-3-allow-for-optional-larger-bolder-legend-text-within-the-fieldset-component

### DIFF
--- a/src/components/fieldset/_fieldset.scss
+++ b/src/components/fieldset/_fieldset.scss
@@ -20,12 +20,29 @@
     box-sizing: border-box; // 1
     display: table;         // 2
     max-width: 100%;        // 1
+    margin-bottom: $govuk-spacing-scale-4;
     padding: 0;
     // Hack to let legends or elements within legends have margins in webkit browsers
     overflow: hidden;
 
     color: $govuk-text-colour;
     white-space: normal;    // 1
+
+    .govuk-heading-s {
+      margin-bottom: 0;
+    }
+
+    .govuk-heading-m {
+      margin-bottom: $govuk-spacing-scale-1;
+    }
+
+    .govuk-heading-l {
+      margin-bottom: $govuk-spacing-scale-2;
+    }
+
+    .govuk-heading-xl {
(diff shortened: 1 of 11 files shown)


nickcolley:
Could this be a composition of a fieldset component and the radio group?

36degrees:
Sorry, I don't understand the question… this is using the fieldset component already?

nickcolley:
Oliver Byford in answer to my question on slack:
"
Partially because it makes a cleaner mapping between what we have in Frontend and what we have in the Design System (otherwise to have accessible, complete examples in the radio button component we’d be asking users to use other components too, which muddies things, and partly because we think the `call` syntax is less ‘friendly’ than the way you use other components, so us ‘doing the work’ simplifies things for users.
"

dashouse:
Approved from Design Perspective

src/components/fieldset/_fieldset.scss
```diff
@@ -20,12 +20,26 @@
...
+      margin-bottom: $govuk-spacing-scale-1;
+    }
+
+    .govuk-heading-l,
+    .govuk-heading-xl {
+      margin-bottom: $govuk-spacing-scale-3;
+    }
   }
 
   .govuk-c-fieldset__legend--bold {
```
hannalaakso: Not strictly part of the PR but I noticed that `govuk-c-fieldset__legend--bold` is not used anywhere so it could be removed from here. To make text bold, passing `govuk-!-w-bold` works well.

36degrees: I think it's a useful variant to have, even if we don't have an example of it.

PR #377, opened 2017-12-12
A few fixes for components with overridden colours

36degrees:
## Radio Buttons

Removes a transparent border from the radio button labels.

### Before

screen shot 2017-12-12 at 11 00 25

### After

screen shot 2017-12-12 at 11 00 30

## Checkboxes

Removes a transparent border from the checkbox labels.

### Before
screen shot 2017-12-12 at 11 00 35

### After
screen shot 2017-12-12 at 11 00 36

## Panel

Adds a transparent border which is made visible when users override colours.

### Before
screen shot 2017-12-12 at 11 00 18

### After
screen shot 2017-12-12 at 11 00 21

### DIFF
--- a/src/components/checkbox/_checkbox.scss
+++ b/src/components/checkbox/_checkbox.scss
@@ -40,7 +40,6 @@
   .govuk-c-checkbox__label {
     display: block;
     padding: em(8px, 19px) em($govuk-spacing-scale-3, 19px) em($govuk-spacing-scale-1, 19px);
-    border: 2px solid transparent;
     cursor: pointer;
     // remove 300ms pause on mobile
     -ms-touch-action: manipulation;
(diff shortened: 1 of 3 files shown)


nickcolley:
How does this impact hint text?
Did that invisible border set some spacing for this components that'll be missed?

36degrees:
> How does this impact hint text?

I don't believe it does?

> Did that invisible border set some spacing for this components that'll be missed?

My understanding is that the 2px border was originally added to align components to the same height, however removing it actually improves the alignment as seen on the form-alignment example:

### Before

screen shot 2017-12-12 at 11 07 56

### After

screen shot 2017-12-12 at 11 08 04

igloosi:
@nickcolley that border added part of the equalising heigh form elements a while ago. component have now changed so that is no longer relevant/workable. also looking at the firefox custom colours, it's not the best solution.

PR #378, opened 2017-12-12
Refactor lists into global scss

joelanman:
- move the govuk list styles (normal, bullet and number) to `globals/scss/_lists`
- remove the list component - update all components that use the list component
- remove all code and references to the steps variant - if it's MVP we can add again in future but it doesn't seem part of core styles

This work allows easy styling of lists without having to use the component. This is useful for example when we do a styled wrapper like govspeak.


36degrees:
I think keeping the list component is just going to muddy things and cause confusion - I think we probably need to refactor the error summary component to generate the list without using the list macro?

If we're going to move the steps to a new component I think that needs to be done at the same time.

joelanman:
Is steps really important to MVP? I would suggest not?

joelanman:
I've now removed the list component completely and updated the components that used it.

36degrees:
There's a few things within the packages folder that will need updating:

- The error summary component will need the dependency on the list component removing from its package.json
- The all package will need the dependency on the list component removing from its package.json

I'm unclear as to whether that should happen as part of this PR or when we publish the next release – @igloosi @alex-ju do you know what the right approach would be?

src/views/examples/typography/index.njk
```diff
@@ -111,7 +111,21 @@
       <p class="govuk-body-l govuk-!-w-bold">govuk-body-l + govuk-!-w-bold</p>
 
     </section>
-
+    <section class="govuk-!-mt-r8">
+      <h2 class="govuk-heading-l govuk-!-pb-r2" style="border-bottom: 4px solid;">Lists</h2>
```
36degrees: Should we have an example of a list with neither bullets nor numbers?

joelanman: I think only if we actually do that on gov.uk anywhere - I'll have a look around

36degrees: Lists of links in the related widget tend not to use bullets.

joelanman: thanks - adding now

joelanman:
I've added an example for a normal list.

I've removed references to the list component from packages.

I think we should do steps as another PR if needed, otherwise it's blocking the work on the prose scope.

36degrees:
@joelanman in rebasing this I think we lost the change of the list classes for the component and example lists, which were combined to create one page in #387. It's currently rendering incorrectly on http://govuk-frontend-review.herokuapp.com/.

joelanman:
sorry - will fix now

PR #381, opened 2017-12-13
Pluralise breadcrumb component

igloosi:
We've agreed to make component names plural where there they are always or nearly always used in groups (radios, breadcrumbs and checkboxes).

It doesn't make sense to refer to these things as singular, since they are always (or nearly always, in the case of checkboxes) used in groups.

This PR updates:
* Readme wording
* Class names
* Macro
* Folder name
* YAML file name
* Macro name where used

Part of [Trello ticket](https://trello.com/c/Vof8gDRz/362-2-pluralise-breadcrumb-radio-and-checkbox-components)


36degrees:
From a comment in #384:

> Are we sure on breadcrumb? In general usage I'd refer to the 'breadcrumb on the page' or 'breadcrumb trail' rather than the 'breadcrumbs on the page'.

I think, given the origin of the term, breadcrumbs makes more sense. To me at least, each link in the trail is a 'breadcrumb', so the component is generating multiple 'breadcrumbs'.

They are currently breadcrumbs in Elements and Frontend Toolkit.

Other design systems and frameworks are more divided than I'd expected:

- [Salesforce](https://www.lightningdesignsystem.com/components/breadcrumbs) use the plural.
- [IBM](http://carbondesignsystem.com/components/breadcrumb/code) use the singular but their guidance doesn't read that well IMO ("Breadcrumb enables users to quickly see their location within a path of navigation and move up to a parent level if desired.")
- [Bootstrap](http://getbootstrap.com/docs/4.0/components/breadcrumb/) uses the singular for the component name, but uses the plural in the guidance ("Since breadcrumbs provide a navigation…").

But generally in guidance the plural is being used, and if that's the case I think the component should be pluralised too.

joelanman:
I was surprised by the differences that are out there too, but my vote is with plural, for the reasons @36degrees gave.

Other references for plural:

[Nielsen](https://www.nngroup.com/articles/breadcrumb-navigation-useful/)

[Brad Frost](https://bradfrost.github.io/this-is-responsive/patterns.html#breadcrumbs)

[Wikipedia](https://en.wikipedia.org/wiki/Breadcrumb_(navigation)) (uses singular as title, for disambiguation, but refers to plural throughout the content)

edwardhorsford:
I think I'm half and half - but just wanted to point out this isn't _obviously_ a pluralised thing. I think it's both simultaneously - might just be something for a content designer to look at.

You might say 'the breadcrumb pattern' and also 'the breadcrumbs on the page'.

Unlike some of the other plural patterns, is this more likely one you might use this on it's own - to go up one level in a service?

PR #382, opened 2017-12-13
Add clipping back to back arrow, abstract into mixin

36degrees:
## Before
screen shot 2017-12-13 at 14 36 25

## After
screen shot 2017-12-13 at 14 36 27


36degrees:
@igloosi would you mind taking a look at the extra commit I've pushed?

igloosi:
@36degrees those look great.

PR #383, opened 2017-12-13
Update component naming guidance

igloosi:
We've agreed to make component names plural where there they are always
or nearly always used in groups (radios, breadcrumbs and checkboxes).

This PR updates previous guidance.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,7 +31,7 @@ For more detail, see our [coding standards for JavaScript](/docs/coding-standard
 
 Find components in `src/components`.
 
-Component folder and files should be **singular, not plural**.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -31,7 +31,7 @@ For more detail, see our [coding standards for JavaScript](/docs/coding-standard
 
 Find components in `src/components`.
 
-Component folder and files should be **singular, not plural**.
+Component folder and files should be singular, unless where they are always or nearly always used in groups (radios, breadcrumbs and checkboxes).
```
amyhupe: I would change to 'except in cases where they are more commonly used in groups, for example, radios, breadcrumbs and checkboxes.'

igloosi:
@amyhupe updated

PR #385, opened 2017-12-13
Fix IE9 full-width button on large devices

alex-ju:
This PR prevent IE9 from rendering the button in full-width by setting the `width` to `auto`

Before
screen shot 2017-12-13 at 17 08 34

After
screen shot 2017-12-13 at 17 10 10

Note: we still have it full-width on IE8 because it doesn't support media queries and we're doing mobile-first CSS, but we'll use `sass-mq` to generate an IE8 friendly stylesheet to deal with that.

### DIFF
--- a/src/components/button/_button.scss
+++ b/src/components/button/_button.scss
@@ -29,7 +29,7 @@
     -webkit-appearance: none;
 
     @include mq($from: tablet) {
-      width: initial;
+      width: auto;
     }
 
     // Set text colour depending on background colour


36degrees:
> Note: we still have it full-width on IE8 because it doesn't support media queries and we're doing mobile-first CSS.

`sass-mq` can rasterize our media queries for us to generate an IE8 friendly stylesheet, we just haven't got that set up for the Frontend review app at the minute.

PR #386, opened 2017-12-13
Correct grid behaviour on mobile

36degrees:
Floats should only be applied when the width is – otherwise, containers will shrink-wrap their content and may end up floated next to each other.

## Before
screen shot 2017-12-13 at 17 12 16

## After
screen shot 2017-12-13 at 17 12 43


36degrees:
Good call, have updated to do that instead.

PR #389, opened 2017-12-14
fix references to govuk-c-list

joelanman:
Fix a couple of references to `govuk-c-list`, which is now `govuk-list`

### DIFF
--- a/packages/error-summary/error-summary.html
+++ b/packages/error-summary/error-summary.html
@@ -7,7 +7,7 @@
     <p>
       Optional description of the errors and how to correct them
     </p>
-    <ul class="govuk-c-list govuk-c-error-summary__list">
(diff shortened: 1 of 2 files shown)


packages/error-summary/error-summary.html
```diff
@@ -7,7 +7,7 @@ <h2 class="govuk-c-error-summary__title" id="error-summary-title">
     <p>
       Optional description of the errors and how to correct them
     </p>
-    <ul class="govuk-c-list govuk-c-error-summary__list">
+    <ul class="govuk-list govuk-c-error-summary__list">
```
36degrees: This will be updated when we next publish the packages, using the changes that were made to the component in `src`. Whilst it's not the end of the world to include this change, it'd confuse things less if could revert it for now.

36degrees:
Superseded by #394

PR #388, opened 2017-12-14
Pluralise radio component

igloosi:
We've agreed to make component names plural where there they are always
or nearly always used in groups (radios, breadcrumbs and checkboxes).

This PR updates:

Readme wording
Class names
Macro
Folder name
YAML file name
Macro name where used
Updates the reference in the *all* package json and scss files

Part of [Trello ticket](https://trello.com/c/Vof8gDRz/362-2-pluralise-breadcrumb-radio-and-checkbox-components)

PR #390, opened 2017-12-15
Update fieldset component caller()

igloosi:
Controlling whitespace when fieldset is consumed in other components has become very tricky.

Removing `trim` and and `indent` will allow exact positioning to be handled where fieldset is called from.

This will fix whitespace in https://github.com/alphagov/govuk-frontend/pull/388
but break it in date-input, which is getting fixed [here](https://github.com/alphagov/govuk-frontend/pull/391)

### DIFF
--- a/src/components/fieldset/README.md
+++ b/src/components/fieldset/README.md
@@ -1,3 +1,7 @@
+<div class="govuk-o-width-container">
+
+<div class="govuk-o-main-wrapper">
+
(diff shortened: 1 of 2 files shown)


36degrees:
Would be good to understand this better, can we go through it on Monday? I thought indentation would be fine as long as the block passed by the caller isn't indented…

PR #391, opened 2017-12-15
Update date-input component

igloosi:
Update whitespace control to align with the fixed fieldset component, where [trim and indent have been removed](https://github.com/alphagov/govuk-frontend/pull/390)

### DIFF
--- a/src/components/date-input/README.md
+++ b/src/components/date-input/README.md
@@ -1,3 +1,7 @@
+<div class="govuk-o-width-container">
+
+<div class="govuk-o-main-wrapper">
+
 # Date input
 
 ## Introduction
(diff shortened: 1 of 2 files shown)


src/components/date-input/README.md
```diff
@@ -507,6 +529,14 @@ Guidelines can be found at [on our Github repository.](https://github.com/alphag
 
 ## Acknowledgements/credits
 
+*   GDS developers
```
alex-ju: There is a card/task in the backlog to remove Acknowledgements/credits section (with which I agree), but I don't understand why they appear/disappear from one update to another 🤔

36degrees: I think the usage of the list component in the README template was broken (it predated passing objects rather than kwargs) – when Joe removed the list component he updated the template to just use a `<ul>` and so now updates to READMEs are bringing them back.

alex-ju: Yep, that would make sense.

igloosi:
after pushing merged fieldset component changes:
screen shot 2017-12-22 at 13 58 09

36degrees:
Beautiful 👨‍🎨

PR #392, opened 2017-12-18
Button 'tweaks'

36degrees:
I suggest reviewing commit by commit.

This also fixes the 'form alignment' example and adds documentation to the typography helper.

### DIFF
--- a/src/components/button/_button.scss
+++ b/src/components/button/_button.scss
@@ -27,11 +27,33 @@
     text-align: center;
     text-decoration: none;
     vertical-align: top;
+
     cursor: pointer;
     -webkit-appearance: none;
 
     @include mq($from: tablet) {
       width: auto;
+    }
+
+    @include ie-lte(8) {
+      // Substitute shadow for border in IE8
+      padding: ($govuk-spacing-scale-2 - $govuk-border-width-form-element) $govuk-spacing-scale-2;
+      border-bottom: $button-shadow-size solid $button-shadow-colour;
+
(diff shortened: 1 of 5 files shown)


dashouse:
There's also an issue with reflow as the `input` will change size at mobile breakpoint when the font size changes

src/components/button/_button.scss
```diff
@@ -1,80 +1,109 @@
...
-    // Fix unwanted button padding in Firefox
-    &:-moz-focus-inner {
-      padding: 0;
-      border: 0;
+    // These adjustments do not work for <input type="button"> as non-container
+    // elements cannot include pseudo elements.
+
+    // Expand the click target to cover the shadow as well, in case users try
+    // to click it.
+    &::before {
```
nickcolley: Does this account for the `box-shadow`? In the original there's a `110%` definition since if you try and click the `box-shadow` it does not work.

src/components/button/_button.scss
```diff
@@ -1,80 +1,109 @@
...
-      }
-    } @else {
-      color: $govuk-text-colour;
-
-      &:link,
-      &:hover,
-      &:focus,
-      &:visited {
-        color: $govuk-text-colour;
+    @include ie-lte(8) {
```
nickcolley: Are the buttons functional without this change? Do we need to add this?

36degrees:
There's some more work to do with this and I'm conscious this PR is already pretty big, so I'm going to try and split it up a little bit.

PR #393, opened 2017-12-18
Add scoped prose styles

igloosi:
Styles for generic page elements that are wrapped in .govuk-prose-scope
container element class.

Includes: 
* headings
* paragraphs
* heading and paragraph combinations
* links
* list styles (ul, and ol)

[Trello ticket](https://trello.com/c/lJ6jgFcN/374-5-define-a-prose-scope)

### DIFF
--- a/src/globals/scss/core/_prose-scope.scss
+++ b/src/globals/scss/core/_prose-scope.scss
@@ -1,0 +1,48 @@
+
+@include exports("prose-scope") {
+  .govuk-prose-scope {
+
+    // @extend inheritance
+
+    // Contextual heading and paragraph combinations are inherited
+    // through the use of @extend
+
+    h1 {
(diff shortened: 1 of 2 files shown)


src/globals/scss/core/_prose-scope.scss
```diff
@@ -0,0 +1,48 @@
+
+@include exports("prose-scope") {
+  .govuk-prose-scope {
+
+    // @extend inheritance
+
+    // All contextual heading and paragraph combinations are getting inherited
+    // when .govuk-body is getting extended
```
36degrees: The contextual adjustments are not just related to `.govuk-body` so I think this should just say something like

> Contextual heading and paragraph combinations are inherited through the use of @extend

igloosi:
@36degrees updated

PR #401, opened 2017-12-21
Move and extend gulp:clean task

igloosi:
With recent changes to frontend when a lot of the files were moved and folders deleted, we'd need to clean the packages folder before publishing, so that old files didn't get published as well.

We didn't want to lose all the package.json files and we didn't have the clean task setup for `packages` folder.
This PR:
* moves the clean task from gulpfile.js to it's own partial 
* extends the task so that it checks for the destination folder that is passed from the parent `npm run build:packages` or  `npm run build:dist` and either clean the whole whole `dist/` folder or keeps `package.json` files and `README.md` in `packages/all` 

https://trello.com/c/dsb17NNi/499-clean-packages-folder-before-building-new-packages

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -6,9 +6,9 @@ const taskListing = require('gulp-task-listing')
 const runsequence = require('run-sequence')
 const taskArguments = require('./tasks/gulp/task-arguments')
 const nodemon = require('nodemon')
-const del = require('del')
 
 // Gulp sub-tasks
+require('./tasks/gulp/clean.js')
 require('./tasks/gulp/lint.js')
 require('./tasks/gulp/test-app.js')
 require('./tasks/gulp/test-components.js')
@@ -33,13 +33,6 @@ gulp.task('scripts', cb => {
 // --------------------------------------
 gulp.task('styles', cb => {
   runsequence('scss:lint', 'scss:compile', cb)
-})
-
-// Clean task for a specified folder ----
-// Removes all old files
-// --------------------------------------
-gulp.task('clean', function () {
-  return del([taskArguments.destination + '/**/*'])
 })
 
 // Copy icons task ----------------------
@@ -101,6 +94,7 @@ gulp.task('serve', ['watch'], () => {
 // -------------------------------------
 gulp.task('build:packages', cb => {
   runsequence(
+              'clean',
               'compile:components',
               'copy-files',
               'generate:readme',
--- a/tasks/gulp/clean.js
+++ b/tasks/gulp/clean.js
@@ -1,0 +1,28 @@
+'use strict'
+
+const gulp = require('gulp')
+const taskArguments = require('./task-arguments')
+const del = require('del')
+
+// Clean task for a specified folder --------------------
+// Removes all old files, except for package.json
+// and README in all package
+// ------------------------------------------------------
+
+gulp.task('clean', () => {
+  let destination = taskArguments.destination
+
+  if (destination === 'packages') {
+    return del.sync([
+      `${destination}/**`,
+      `!${destination}`,
+      `!${destination}/*`,
+      `!${destination}/**/package.json`,
+      `!${destination}/all/README.md`
+    ])
+  } else {
+    return del.sync([
+      `${destination}/**/*`
+    ])
+  }
+})


tasks/gulp/clean.js
```diff
@@ -0,0 +1,16 @@
...
+
+// Clean task for a specified folder --------------------
+// Removes all old files, except for package.json
+// and README in all package
+// '/**/**/*' is to remove folder deep down  in globals/
+// ------------------------------------------------------
+const isPackages = (taskArguments.destination === 'packages') || false
+
+gulp.task('clean', () => {
+  return del.sync(isPackages ? [taskArguments.destination + '/**/*.*', taskArguments.destination + '/**/**/*', '!' + taskArguments.destination, '!' + taskArguments.destination + '/**/package.json', '!' + taskArguments.destination + 'all/README.md'] : [taskArguments.destination + '/**/*'])
```
36degrees: This is really hard to follow. Ternary doesn't feel like the right choice when you end up with the `:` hidden in the middle of what's already a big list.

I'd suggest perhaps something like:

```
gulp.task('clean', () => {
  const destination = taskArguments.destination

  if (destination === 'packages') {
    # Clean packages folder
    return del.sync([
      `${destination}/**/*.*`,
      `${destination}/**/**/*`,
      `!${destination}`,
      `!${destination}/**/package.json`,
      `!${destination}/all/README.md`
    ])
  } else {
    # Clean dist
    return del.sync([`${destination}/**/*`])
  }
})
```

Alternatively, it might be better just to define separate clean tasks for `clean:packages` and `clean:dist`? That would allow for more specific documentation as well.

tasks/gulp/clean.js
```diff
@@ -0,0 +1,16 @@
...
+const gulp = require('gulp')
+const taskArguments = require('./task-arguments')
+const del = require('del')
+
+// Clean task for a specified folder --------------------
+// Removes all old files, except for package.json
+// and README in all package
+// '/**/**/*' is to remove folder deep down  in globals/
+// ------------------------------------------------------
+const isPackages = (taskArguments.destination === 'packages') || false
```
36degrees: You don't need the `|| false` here. You're evaluating a boolean expression – the outcome of which can only be `true` (in which case the expression after the `||` will never be evaluated) or `false` (in which case… it's already false).

tasks/gulp/clean.js
```diff
@@ -0,0 +1,16 @@
+'use strict'
+
+const gulp = require('gulp')
+const taskArguments = require('./task-arguments')
+const del = require('del')
+
+// Clean task for a specified folder --------------------
+// Removes all old files, except for package.json
+// and README in all package
+// '/**/**/*' is to remove folder deep down  in globals/
```
36degrees: I'm also not sure that `${destination}/**/**/*` is needed - as `**` allows for any numbers of directories so folders within globals should be included as part of `${destination}/**/*.*`? It definitely feels like if we get the syntax right we should only need one 'positive' path in the list.

gulpfile.js
```diff
@@ -35,13 +35,6 @@ gulp.task('styles', cb => {
   runsequence('scss:lint', 'scss:compile', cb)
 })
 
-// Clean task for a specified folder ----
-// Removes all old files
-// --------------------------------------
-gulp.task('clean', function () {
```
36degrees: Good call moving this to it's own file 👌

tasks/gulp/clean.js
igloosi: i can have a look, i found that **/* didnt delete folders 3 levels down packages/globals/var for example

tasks/gulp/clean.js
igloosi: if inside the gulp task isn't really a thing unfortunately

36degrees: We're inside a function and using `return` here so are you sure that's the case?

tasks/gulp/clean.js
36degrees: Is that because `**/*.*` should be `**/*`?

36degrees:
Just to check, is this relying on users specifying a `destination` as an argument, or is it inferred based on which task they're running?

igloosi:
destination flag are defined on npm run build:x tasks

36degrees:
So users can just run `gulp build:packages` and this will work as expected?

igloosi:
yeah

tasks/gulp/clean.js
igloosi: yeah tried both your example above and just using boolean.
I can try and move this up to a function outside and run the function on return

tasks/gulp/clean.js
36degrees: So I think the issue is that `**/*` deletes component folders even if they still contain a package.json file?

tasks/gulp/clean.js
36degrees: I've got an example working locally which uses an `if` within the function and it works just fine. Happy to share the code if helpful.

36degrees:
Trying this locally (without any changes) and it appears it's deleting everything under packages, not preserving the **/package.json or all/readme files as intended.

igloosi:
when running npm run build:packages?

36degrees:
Yes, and when running `gulp clean --destination packages` directly.

igloosi:
sorry read your comment above 
gulp build:package won't as it has no destination flag.
npm run build:packages has --destination 'packages'

atm were documenting to run npm run build:packages.
to run gulp build:packages you'd need to add -- destination 'packages' to it

36degrees:
Yes, even running the 'correct' `npm` command it's still not preserving the `package.json` files.

I'm going to put some time in Q4 to revisit the build pipeline. No matter what we document, iot's confusing and dangerous that you can do something that _looks like_ it builds the packages, but with undefined behaviour.

igloosi:
updated

36degrees:
Appears to work, though I don't _really_ understand how when `packages/**/**/*` is the only 'positive' path to remove.

I've pushed [another branch](https://github.com/alphagov/govuk-frontend/compare/modify-gulp-clean-task-ob?expand=1) with a further commit showing how it can be rewritten to be easier to read - I've tested it locally and it works. Would be good to do something similar (or adopt it verbatim) to improve the readability as at the minute it is very hard to follow.

igloosi:
updated to your approach and changed the glob pattern so there is no magic :)

PR #404, opened 2017-12-22
Allow line height to be overridden in typography helpers

36degrees:
This allows the line height to be overridden, and updates existing components to do that rather than to redeclare it.

As a result we should no longer end up with multiple line-height declarations within the same ruleset.


igloosi:
looks good to me, have someone else look at it

PR #406, opened 2018-01-02
Add button hover colour

joelanman:
This is the only colour on our documentation page that does not have a semantic name:

https://govuk-elements.herokuapp.com/colour/

### DIFF
--- a/src/components/button/_button.scss
+++ b/src/components/button/_button.scss
@@ -67,7 +67,7 @@
 
     &:hover,
     &:focus {
-      background-color: darken($govuk-button-colour, 5%);
+      background-color: $govuk-button-hover-colour;
     }
 
     &:active {
--- a/src/globals/scss/settings/_colours-applied.scss
+++ b/src/globals/scss/settings/_colours-applied.scss
@@ -4,6 +4,7 @@ $govuk-link-active-colour: $govuk-light-blue;
 $govuk-link-hover-colour: $govuk-light-blue;
 $govuk-link-visited-colour: #4c2c92;
 $govuk-button-colour: #00823b;
+$govuk-button-hover-colour: darken($govuk-button-colour, 5%);
 $govuk-button-colour-darken-15: darken($govuk-button-colour, 15%);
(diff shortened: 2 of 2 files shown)


36degrees:
At the minute the hover colour for the button is calculated within the component as `darken($govuk-button-colour, 5%);`. We should only be adding this if we're going to update the button component to actually use it.

joelanman:
That's odd - Elements documents it as $green (#006435) but the component is using #00692f instead. Was that a deliberate choice?

igloosi:
it's defined here https://github.com/alphagov/govuk_frontend_toolkit/blob/master/stylesheets/design-patterns/_buttons.scss#L56

joelanman:
I've updated to:
`$govuk-button-hover-colour: darken($govuk-button-colour, 5%);`
So it is the same as the colour used in the component.
This change means we can properly refer to this colour in the Colour page, with a semantic name.

36degrees:
Can you also update the button component to use the new variable?

joelanman:
@36degrees I'm not sure we would want that. The button component takes a single colour, and applies darken to get the hover colour. This way we can make other colour buttons. If I added the hover colour there, we wouldn't be able to do that. What this does do is make sure the colour listed here is exactly the same as the one calculated there.

36degrees:
The button component uses `$govuk-button-colour` in this same file to decide what colour the button should be - it's no longer a macro and there's no other way to override the colour.

If we have a variable called `$govuk-button-hover-colour` in the same file then I'd expect that the button component would use it.

joelanman:
ok I think I was confused because the button @igloosi linked above is a mixin, that takes a colour argument. The documentation tells you you can make a red version:

https://github.com/alphagov/govuk_frontend_toolkit/blob/master/stylesheets/design-patterns/_buttons.scss#L10-L23

but the button component doesn't use that - I'll update the button component

joelanman:
ugh, sorry my fault, didn't realise that was a link to the old frontend toolkit, not ours

PR #408, opened 2018-01-03
Add adjecent styles for headings after lists

dashouse:
This adds padding to the top of headings that follow a list

### DIFF
--- a/src/globals/scss/core/_typography.scss
+++ b/src/globals/scss/core/_typography.scss
@@ -139,14 +139,17 @@
   }
 
   .govuk-body-m  + .govuk-heading-l,
-  .govuk-body-s  + .govuk-heading-l {
+  .govuk-body-s  + .govuk-heading-l,
+  .govuk-list + .govuk-heading-l {
     @include govuk-responsive-padding($govuk-spacing-responsive-4, "top");
   }
 
   .govuk-body-m + .govuk-heading-m,
   .govuk-body-s + .govuk-heading-m,
+  .govuk-list + .govuk-heading-m,
   .govuk-body-m + .govuk-heading-s,
-  .govuk-body-s + .govuk-heading-s {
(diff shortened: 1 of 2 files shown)


36degrees:
Worth ensuring there's an example for this on the typography example page?

dashouse:
@36degrees Updated

src/globals/scss/core/_typography.scss
```diff
@@ -139,14 +139,17 @@
   }
 
   .govuk-body-m  + .govuk-heading-l,
-  .govuk-body-s  + .govuk-heading-l {
+  .govuk-body-s  + .govuk-heading-l,
```
hannalaakso: Does this rule somehow relate to lists?

src/globals/scss/core/_typography.scss
```diff
@@ -139,14 +139,17 @@
...
+  .govuk-list + .govuk-heading-l {
     @include govuk-responsive-padding($govuk-spacing-responsive-4, "top");
   }
 
   .govuk-body-m + .govuk-heading-m,
   .govuk-body-s + .govuk-heading-m,
+  .govuk-list + .govuk-heading-m,
   .govuk-body-m + .govuk-heading-s,
-  .govuk-body-s + .govuk-heading-s {
+  .govuk-body-s + .govuk-heading-s,
```
hannalaakso: Similar here, does this relate to lists styling?

src/globals/scss/core/_typography.scss
36degrees: The only thing that's changed is the opening parenthesis has been changed to a comma to allow the extra selectors to be added to the list…

hannalaakso: So it has, I must be tired 😅

PR #409, opened 2018-01-03
Improve documentation for publishing NPM modules

hannalaakso:
This PR improves the documentation on how to publish new Frontend modules to NPM and how to publish existing modules to NPM that have been updated, by:

- Adding more detail of what various commands do to make it easier to follow the documentation.
- Expand on the publishing process with Lerna, including screenshots.

Notes:

I think the publishing section should possibly move to its own file, it's a bit hidden inside "components" file. However I'm leaving that as it is for now as I think the docs could do with reorganising in general to make it easier to find stuff and this is something we should probably look at as a team.

Some of the commit messages syntax in the publishing docs uses conventions (eg.`chore`) listed [here](https://github.com/alphagov/govuk-frontend/blob/master/docs/components.md#commit-structure) that weren't part of our [commit message convention](https://github.com/alphagov/design-system-team-docs/pull/3). I've added a ticket to the dev-get-together-board to review this.


docs/components.md
```diff
@@ -21,62 +21,83 @@ We are using [lerna.js](https://lernajs.io/) to manage our packages and publish
...
+To do: We need to do more work to clarify how versioning, including managing breaking changes, will work.
+
 ### Publishing components manually (while in Alpha)
 
 In a new branch:
 
-- Run the `npm run build:packages` task to ensure any changes to `src/components/component-name` are copied to `packages/component-name`.
+1. Run the `npm run build:packages` task.
+
+2. Ensure all changes to `src/components/component-name` have been copied to `packages/component-name`. (Atom show changed directories in green).
```
nickcolley: Nit: Could we use `git status` instead of 'atom' since not everyone will be using atom as their text editor?

hannalaakso: Fair point, it was too specific to my particular set up. I took it out 👍

docs/components.md
```diff
@@ -21,62 +21,83 @@ We are using [lerna.js](https://lernajs.io/) to manage our packages and publish
...
 
+3. For any new components, add a `package.json` inside `packages/component-name`:
 ```
-npm run build:packages
+{
+    "name": "@govuk-frontend/component-name",
+    "version": "version-number",
+    "dependencies": {
+      "@govuk-frontend/globals": "version-number",
+      "@govuk-frontend/icons": "version-number"
```
36degrees: The dependencies for each component will be different – and I suspect the vast majority will not need the icons package. Suggest adding a sentence about making sure you include any components you depend on, and removing icons from here?

docs/components.md
```diff
@@ -21,62 +21,83 @@ We are using [lerna.js](https://lernajs.io/) to manage our packages and publish
...
 
-Open a pull request for these changes.
-
+6. Open a pull request for these changes.
 
-Once these have been merged into master.
+Once the PR has been merged into master:
 
-- Run lerna bootstrap to install all package dependencies and link any cross-dependencies
+1. Log in to `npm` as authorised user.
```
36degrees: As a user who has permission to publish to the govuk-frontend scope?

docs/components.md
```diff
@@ -21,62 +21,83 @@ We are using [lerna.js](https://lernajs.io/) to manage our packages and publish
...
 
-- Manually bump the version of each package
-
+4. Run
 ```
 lerna publish -m "chore(release): update packages and publish"
 ```
+Here, Lerna will prompt you to select the new version for each package. If you select
+`minor`/`major` etc., Lerna will complete the version number for you. We have so
+far selected `custom` and specified the new version number manually for each package
```
36degrees: So far = in private beta?

docs/components.md
```diff
@@ -21,62 +21,83 @@ We are using [lerna.js](https://lernajs.io/) to manage our packages and publish
 
 Lerna is a tool that optimizes the workflow around managing multi-package repositories with git and npm.
 
+It manages dependencies between our components so that if the version number of `globals` is bumped, all components dependent on `globals` will also have their version bumped.
+
+To do: We need to do more work to clarify how versioning, including managing breaking changes, will work.
```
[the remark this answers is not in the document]
hannalaakso: I guess this was an attempt to capture that versioning is complicated and that in the future there should be more info somewhere (hopefully here in this file but if not, the user might need to go looking for it) on how it works. But I agree it's not really workable to have these to-do bits around, better to have captured it in Trello which we have done.

docs/components.md
hannalaakso: Yup, good shout.

hannalaakso:
Thanks for the review @36degrees, see above for my comments.

I've now broken the publishing docs into their own file as a separate commit.

PR #416, opened 2018-01-09
Make buttons 40px high including box shadow

36degrees:
Because the shadow is visually 'part of' the button, we need to reduce the height of the button to compensate by adjusting its padding and increase the bottom margin to include it.

We also fix the line height at 40px which means that whilst the button font size reduces on a mobile breakpoint, the height of the element remains the same. This means that when we remove the use of `em` for the inputs we will end up with inputs and buttons having a consistent height of 40px at all breakpoints.

Until the use of `em` for inputs is removed, this currently looks worse on mobile because the inputs become shorter but the button now does not.

### Before
screen shot 2018-01-09 at 09 55 27

### After
screen shot 2018-01-09 at 09 56 08

### Before
screen shot 2018-01-09 at 09 55 38

### After
screen shot 2018-01-09 at 09 56 00

https://trello.com/c/03CyckKm/488-correct-button-height

### DIFF
--- a/src/components/button/_button.scss
+++ b/src/components/button/_button.scss
@@ -2,6 +2,9 @@
 
 @include exports("button") {
 
+  // Because the shadow (s0) is visually 'part of' the button, we need to reduce
+  // the height of the button to compensate by adjusting its padding (s1) and
+  // increase the bottom margin to include it (s2).
   $button-shadow-size: $govuk-border-width-form-element;
 
   .govuk-c-button {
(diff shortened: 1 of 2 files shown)


nickcolley:
How should the buttons align? At the moment they're different height to the input, is that intended?

36degrees:
They _should_ align on desktop and mobile, but at the minute they only align on desktop. Mobile will follow once https://trello.com/c/WpPKyqxl/493-define-size-of-inputs-etc-in-pixels-rather-than-em is done as inputs and buttons will all be 40px high at all breakpoints.

PR #417, opened 2018-01-09
Fix pseudo element click target corrections

36degrees:
The percentage based solution was not taking into account the borders which have been introduced since copying this from Elements, so it has been refactored to use calculated offsets instead.

Remove some IE6 specific logic as we no longer support this browser.

Nest the pseudo elements into their parent selector to group related items together.


nickcolley:
Tested this in Mac Chrome, looks good 👍

PR #419, opened 2018-01-10
Update padding on the tag component

dashouse:
Tiny PR to update the padding left and right to 8px.
Sorry for the magic number but it looks more balanced visually than using 5px or 10px.

Haven't looked at top / bottom as it's font related


nickcolley:
Feels like magic numbers that are internal to a component are more reasonable than those that are used in spacing between other components etc.

PR #420, opened 2018-01-10
RFC - Conditional reveal styles

dashouse:
**Request for comment**

This adds a style (only) for conditionally revealing content inside radios and checkboxes.

You can review what this looks like at the bottom of the "form-elements" review page.

I'm expecting this code might need to be refactored as it's duplicated in both components.

screen shot 2018-01-16 at 09 13 43

https://trello.com/c/ZO4zmupU/452-add-styles-for-conditionally-revealed-content

### DIFF
--- a/src/components/checkboxes/_checkboxes.scss
+++ b/src/components/checkboxes/_checkboxes.scss
@@ -102,7 +102,7 @@
   }
 
   .govuk-c-checkboxes__conditional {
-    margin-left: 15px;
+    margin-left: $govuk-spacing-scale-3;
     @include govuk-responsive-margin($govuk-spacing-responsive-4, "bottom");
     padding-left: 28px;
     border-left: $govuk-border-width-mobile solid $govuk-border-colour;
@@ -112,7 +112,7 @@
       padding-left: 33px;
     }
 
-    & [class^="govuk-c-"]:last-child {
+    & > [class^="govuk-c-"]:last-child {
       margin-bottom: 0;
     }
   }
--- a/src/components/checkboxes/_checkboxes.scss
+++ b/src/components/checkboxes/_checkboxes.scss
@@ -100,4 +100,20 @@
   .govuk-c-checkboxes__input:disabled + .govuk-c-checkboxes__label {
     opacity: .5;
   }
+
+  .govuk-c-checkboxes__conditional {
+    margin-left: 15px;
+    @include govuk-responsive-margin($govuk-spacing-responsive-4, "bottom");
+    padding-left: 28px;
+    border-left: $govuk-border-width-mobile solid $govuk-border-colour;
+
+    @include mq($from: tablet) {
+      margin-left: 18px;
+      padding-left: 33px;
+    }
+
+    & [class^="govuk-c-"]:last-child {
+      margin-bottom: 0;
+    }
+  }
 }
--- a/src/components/radios/_radios.scss
+++ b/src/components/radios/_radios.scss
@@ -103,4 +103,20 @@
   .govuk-c-radios__input:disabled + .govuk-c-radios__label {
     opacity: .5;
   }
+
+  .govuk-c-radios__conditional {
+    margin-left: 15px;
+    @include govuk-responsive-margin($govuk-spacing-responsive-4, "bottom");
+    padding-left: 28px;
+    border-left: $govuk-border-width-mobile solid $govuk-border-colour;
+
+    @include mq($from: tablet) {
+      margin-left: 18px;
+      padding-left: 33px;
+    }
+
+    & [class^="govuk-c-"]:last-child {
+      margin-bottom: 0;
+    }
+  }
 }
--- a/src/components/radios/_radios.scss
+++ b/src/components/radios/_radios.scss
@@ -105,7 +105,7 @@
   }
 
   .govuk-c-radios__conditional {
-    margin-left: 15px;
+    margin-left: $govuk-spacing-scale-3;
     @include govuk-responsive-margin($govuk-spacing-responsive-4, "bottom");
     padding-left: 28px;
     border-left: $govuk-border-width-mobile solid $govuk-border-colour;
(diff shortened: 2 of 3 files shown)


src/components/radios/_radios.scss
```diff
@@ -103,4 +103,20 @@
   .govuk-c-radios__input:disabled + .govuk-c-radios__label {
     opacity: .5;
   }
+
+  .govuk-c-radios__conditional {
+    margin-left: 15px;
```
igloosi: spacing scale? $govuk-spacing-scale-3;

src/components/checkboxes/_checkboxes.scss
```diff
@@ -100,4 +100,20 @@
   .govuk-c-checkboxes__input:disabled + .govuk-c-checkboxes__label {
     opacity: .5;
   }
+
+  .govuk-c-checkboxes__conditional {
```
igloosi: is it needed in checkboxes? can/should you have conditional content inside checkbox selection?

joelanman:
I'm thinking the second example with Legend H1 might need a second legend, to ask the question for the indented checkboxes starting checkbox 1.

src/components/radios/_radios.scss
dashouse: Fixed

src/components/checkboxes/_checkboxes.scss
dashouse: I'm not 100% sure, but I believe this might be used for contact preferences if the user wants to provide more than one method. Or an opt in checkbox with email input.

Not sure we should actively promote it, but not block it either?

dashouse:
@joelanman example was really just to get something in there to style. Have added a second fieldset and legend on the checkboxes. Don't know if the markup is right though :)

joelanman:
yep, just thinking that if having a legend there is the more commonly correct thing to do, it's useful to see how the alignment/spacing works with it.

src/components/checkboxes/_checkboxes.scss
```diff
@@ -100,4 +100,20 @@
...
+    border-left: $govuk-border-width-mobile solid $govuk-border-colour;
+
+    @include mq($from: tablet) {
+      margin-left: 18px;
+      padding-left: 33px;
+    }
+
+    & [class^="govuk-c-"]:last-child {
+      margin-bottom: 0;
+    }
```
hannalaakso: I wonder if the number of elements we want to apply this to is definite eg. `govuk-c-label`, `govuk-c-input` - the wildcard here feels like it could throw up unexpected results.

src/views/examples/form-elements/index.njk
```diff
@@ -107,5 +107,91 @@
...
+          No
+        </label>
+      </div>
+    </fieldset>
+  </div>
+
+  <div class="govuk-c-radios">
+    <fieldset class="govuk-c-fieldset">
+      <legend class="govuk-c-fieldset__legend">
+        <h1 class="govuk-heading-xl">Legend h1</h1>
```
hannalaakso: If this is going in the review app, can we use something more descriptive here. @dashouse I know you said the content hasn't been looked at - just flagging it 🙂

src/views/examples/form-elements/index.njk
```diff
@@ -107,5 +107,91 @@
...
+            </label>
+          </div>
+          <div class="govuk-c-radios__conditional">
+            <label class="govuk-c-label" for="previous-last-name">
+            Text input
+            </label>
+            <input class="govuk-c-input" id="text-input" name="text-input" type="text">
+          </div>
+          <div class="govuk-c-checkboxes__item">
+            <input class="govuk-c-checkboxes__input" id="checkbox-2" name="checkbox-2" type="checkbox" value="checkbox-2">
```
hannalaakso: The input `name` should match the one above

src/components/checkboxes/_checkboxes.scss
dashouse: Not really no, we know they will be components but could be anything. Input, text area, another fieldset etc 

I'm not sure the results would be unexpected, it will remove the margin-bottom of the last govuk component inside a conditional wrapper. If it did it on some and not others I'd be concerned that would be more unexpected?

hannalaakso: Thanks @dashouse, that sounds okay to me.

Let me know when this is ready to review, I think the input `name` flagged below is the only thing outstanding.

hannalaakso:
@dashouse Sorry I sent the previous message too quickly, didn't really think it through.

If you have a component that contains multiple nested classes that all start with `govuk-c-`, for instance `govuk-c-element` and nested inside of it `govuk-c-element_insides`, the `govuk-c-element_insides` would also be deemed to be `last-child`. I created a JS BIN to demonstrate: https://jsbin.com/yogubibapu/1/edit?html,css,js,output

We could use a direct descendant eg. `& [class^="govuk-c-"]:last-child`? Although it would not work if the user included a wrapper around them, but it would in most cases.

hannalaakso:
@dashouse Would you be able to update the Changelog with this work?

dashouse:
@hannalaakso I think there is a larger conversation that @alex-ju brought up. This will now fail in the review app because it uses HTML only.

Probably need to have a further discussion about what we can do.

src/components/checkboxes/_checkboxes.scss
joelanman: as @dashouse says, its good for contact preferences, so I can imagine its good in other situations.

hannalaakso:
There's an ongoing conversation about how to handle this in Nunjucks here: https://trello.com/c/VsHwPz3U/671-support-conditional-blocks#

nickcolley:
What is the status on this? Should we close in favour of the more complete solution or should we get this in as in?

alex-ju:
FYI: I have rebased my local version of this branch, but it's still html only, not a macro based solution. I was hoping to convert it into a macro-based solution using `form-group` after defining #519.

PR #407, opened 2018-01-02
Generate breadcrumb chevrons using pseudo-elements

36degrees:
This allows us to get rid of another two images, scales cleanly when zoomed, and ensures that the chevron is retained when users override the colours in their browser.

## Chrome 63
screen shot 2018-01-02 at 16 18 43gmt

## Firefox 57
screen shot 2018-01-02 at 16 11 08gmt

## Firefox 57 with custom colours
screen shot 2018-01-02 at 16 09 52gmt

## IE 8 (Windows 7)
win7_ie_8 0

## IE 9 (Windows 7)
win7_ie_9 0

## IE 10 (Windows 7)
win7_ie_10 0

## IE 11 (Windows 7)
win7_ie_11 0

## Edge (Windows 10)
win10_edge_15 0

## Opera 12.16 (Windows 8.1)
win8 1_opera_12 16

## Kindle Fire 2 (Android 4.0)
android_amazon-kindle-fire-2_4 0_portrait

## Samsung Galaxy S5 (Android 4.4)
android_samsung-galaxy-s5_4 4_portrait

## iPhone 5s (iOS 7.0)
ios_iphone-5s_7 0_portrait

## iPhone 6 (iOS 8.3)
ios_iphone-6_8 3_portrait

https://trello.com/c/0I5XLLwU/520-generate-breadcrumb-chevrons-using-pseudo-elements

PR #403, opened 2017-12-22
Remove contributors list from template, fix template markup and update README files

igloosi:
Contributors list is already out of date. We will need to explore a different way of crediting contributions.

This PR:
* removes the [list from the component ](https://trello.com/c/ZpBUWeym/263-remove-the-contributors-list-from-the-individual-component-readmes)view template
* updates template so no wrapper divs are present in the generated markdown files
* updates `pre` and `code` blocks in generated markdown files
* updates all the README files with latest changes

PR #421, opened 2018-01-10
chore(packages): copy changes to packages

hannalaakso:
Update with all the latest changes from source:

- Add `width` classes, which were based on `form-control` classes that were
specific to form control in Elements. (PR [#413](https://github.com/alphagov/govuk-frontend/pull/413))
- Make buttons 40px high including box shadow (PR [#416](https://github.com/alphagov/govuk-frontend/pull/416))
- Fix focus outline style in Chrome and Safari (PR [#414](https://github.com/alphagov/govuk-frontend/pull/414))
- Remove contributors list from template, fix template markup and update README files (PR [#403](https://github.com/alphagov/govuk-frontend/pull/403))
- Generate breadcrumb chevrons using pseudo-elements (PR [#407](https://github.com/alphagov/govuk-frontend/pull/407))
- Rename ‘govuk-body-lede’ to ‘govuk-body-lead’. (PR [#405](https://github.com/alphagov/govuk-frontend/pull/405)) **Breaking change**
- Fix undefined classes in date input macro (PR [#410](https://github.com/alphagov/govuk-frontend/pull/410))
- Allow line height to be overridden in typography helpers (PR [#404](https://github.com/alphagov/govuk-frontend/pull/404))
- Pluralise radio component (PR [#388](https://github.com/alphagov/govuk-frontend/pull/388)) **Breaking change**
- Pluralise checkbox component (PR [#384](https://github.com/alphagov/govuk-frontend/pull/384)) **Breaking change**
- Add documentation for typography helpers / core, simplify syntax (PR [#400](https://github.com/alphagov/govuk-frontend/pull/400))
- Add adjacent styles for headings after lists (PR [#408](https://github.com/alphagov/govuk-frontend/pull/408))

Note: this step of the publishing process is not perfect right now, it's hard to check that this contains the correct changes. We will have some tests eventually to verify.


packages/all/package.json
```diff
@@ -1,31 +1,31 @@
 {
   "name": "@govuk-frontend/all",
-  "version": "0.0.19-alpha",
+  "version": "0.0.20-alpha",
```
nickcolley: Is this normally done by lerna? Do we need to do this manually?

hannalaakso: 👍  After consulting Lerna documentation it appears it's not necessary to change package.json manually. Have taken this out.

packages/all/package.json
```diff
@@ -5,7 +5,7 @@
     "@govuk-frontend/back-link": "^0.0.19-alpha",
     "@govuk-frontend/breadcrumbs": "^0.0.19-alpha",
     "@govuk-frontend/button": "^0.0.19-alpha",
-    "@govuk-frontend/checkbox": "^0.0.19-alpha",
+    "@govuk-frontend/checkboxes": "^0.0.20-alpha",
```
36degrees: Is this bump intentional when renaming?

packages/button/_button.scss
```diff
@@ -55,6 +63,21 @@
         color: $govuk-text-colour;
       }
     }
+
+    // The following adjustments do not work for <input type="button"> as
```
36degrees: So I think `postcss-nested` is doing _crazy things_ here. It's removed the nesting of the `::before` and `::active::before` by shunting them down the file, taking the _immediately preceding_ comment _line_, but leaving the _rest_ of the comment here, divorced from its context.

Not sure what we can do about this in the short term, but… this is not great.

packages/button/_button.scss
```diff
@@ -81,6 +104,24 @@
...
+      position: absolute;
+
+      top: -$govuk-border-width-form-element;
+      right: -$govuk-border-width-form-element;
+      bottom: -($govuk-border-width-form-element + $button-shadow-size);
+      left: -$govuk-border-width-form-element;
+
+      background: transparent;
+    }
+    // 🎉
```
36degrees: :trollface:

src/components/radios/package.json
```diff
@@ -0,0 +1,7 @@
+{
```
36degrees: I don't think this is meant to be added? None of the other components have package.json files in the `src` directory.

hannalaakso: Yes indeed. I've taken it out.

packages/all/package.json
hannalaakso: As far as NPM is concerned, we've published it a new package, not renamed one.

@nickcolley was looking into this and added a note to dev time to discuss how we should handle deleted NPM packages (which govuk-frontend/checkbox will be)

packages/button/_button.scss
hannalaakso: As just discussed in person, we should review why we're using `postcss-nested` when @igloosi is back.

PR #422, opened 2018-01-10
Update Changelog for 0.0.20-alpha

hannalaakso:
This PR adds a Changelog that we can use to track version numbers and changes. It:

- Adds a note about semantic versioning
- Adds a section for Unreleased PRs - this will in future be completed by contributors of PRs
- Lists the features, breaking changes and fixes that form the release

Note: we've not yet settled on a system of semantic versioning, there's a Dev time ticket to discuss this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,28 @@
+# Changelog
+
+Note: We're not following semantic versioning yet, we are going to talk about this soon.
+
+## Unreleased
+
+## 0.0.20-alpha (Breaking release)
+
+Breaking changes:
+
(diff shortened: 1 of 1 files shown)


joelanman:
Not saying this is the best way, but Prototype Kit does it slightly differently:

https://github.com/alphagov/govuk_prototype_kit/releases

We try to list breaking changes first, then new features, then bug fixes

The format is like this

- [#430 Recommend Atom over Sublime text](#)

Would be nice to have a standard format, but maybe we can get to that over time

hannalaakso:
Thanks @joelanman 👍  I like the clear separation between types of changes in the Protype kit one. I've updated the format of this changelog to reflect the Protype kit one.

PR #427, opened 2018-01-12
Replace global link styling with core govuk-link class

36degrees:
This changes the core link styling so that it is applied via classes, rather than styling a elements directly. This brings it into line with the way other core styling is applied, and enforces our principle that GOV.UK Frontend should not have opinions about basic HTML elements.

As part of this, we also need to update a number of components which previously relied on these global link styles, adding back link styling and focus states.

The prose scope has also been updated to apply the link styling to any links.

We also need to update both the examples in the Frontend app and the Frontend app itself to ensure the govuk-link class is applied to all links.

The muted link variant has been copied from the link component.

These changes make the link component redundant, which will be removed in a separate PR.

https://trello.com/c/Bl7Sukkw/530-move-links-into-frontend-core

### DIFF
--- a/src/globals/scss/core/_links.scss
+++ b/src/globals/scss/core/_links.scss
@@ -2,6 +2,7 @@
 
   .govuk-link {
     @include govuk-font-regular;
+    @include govuk-focusable-fill;
 
     // Override the tap highlight colour (the color of the highlight that
     // appears when a link is tapped on some mobile devices). This is
@@ -22,11 +23,6 @@
 
     &:active {
       color: $govuk-link-active-colour;
-    }
-
-    &:focus {
-      outline: $govuk-focus-width solid $govuk-focus-colour;
-      background-color: $govuk-focus-colour;
     }
 
     @include mq($media-type: print) {
--- a/src/globals/scss/core/_links.scss
+++ b/src/globals/scss/core/_links.scss
@@ -1,45 +1,58 @@
 @include exports("links") {
-  a:link {
-    color: $govuk-link-colour;
-  }
 
-  a:visited {
-    color: $govuk-link-visited-colour;
-  }
+  .govuk-link {
+    @include govuk-font-regular;
 
-  a:hover {
-    color: $govuk-link-hover-colour;
-  }
-
-  a:active {
-    color: $govuk-link-active-colour;
-  }
-
-  a {
+    // Override the tap highlight colour (the color of the highlight that
+    // appears when a link is tapped on some mobile devices). This is
+    // ever-so-slightly darker than the default.
     -webkit-tap-highlight-color: rgba($govuk-black, .3);
-  }
 
-  a:focus {
-    outline: $govuk-focus-width solid $govuk-focus-colour;
-    background-color: $govuk-focus-colour;
-  }
-
-  @include mq($media-type: print) {
-    a,
-    a:visited {
-      word-wrap: break-word;
+    &:link {
+      color: $govuk-link-colour;
     }
 
-    a[href^="/"]:after,
-    a[href^="http://"]:after,
-    a[href^="https://"]:after {
-      content: " (" attr(href) ")";
-      font-size: 90%;
+    &:visited {
+      color: $govuk-link-visited-colour;
     }
 
-    a[href^="javascript:"]:after,
-    a[href^="#"]:after {
-      content: "";
+    &:hover {
+      color: $govuk-link-hover-colour;
+    }
+
+    &:active {
+      color: $govuk-link-active-colour;
+    }
+
+    &:focus {
+      outline: $govuk-focus-width solid $govuk-focus-colour;
+      background-color: $govuk-focus-colour;
+    }
+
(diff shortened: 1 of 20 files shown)


src/globals/scss/core/_links.scss
```diff
@@ -1,45 +1,66 @@
...
+    @include govuk-font-regular;
+    @include govuk-focusable-fill;
 
-  a:hover {
-    color: $govuk-link-hover-colour;
-  }
+    // Override the tap highlight colour (the color of the highlight that
+    // appears when a link is tapped on some mobile devices). This is
+    // ever-so-slightly darker than the default.
+    -webkit-tap-highlight-color: rgba($govuk-black, .3);
```
36degrees: I've put a ticket in Trello to revisit this because I'm not convinced it's needed: https://trello.com/c/UeF2Dex3/572-revisit-tap-highlighting-overrides-in-govuk-frontend

src/globals/scss/core/_links.scss
```diff
@@ -1,45 +1,66 @@
...
-    a[href^="/"]:after,
-    a[href^="http://"]:after,
-    a[href^="https://"]:after {
-      content: " (" attr(href) ")";
-      font-size: 90%;
+  // Muted link variant
+  //
+  // Used for secondary links on a page - the link will appear in muted colours
+  // regardless of visited state.
+  .govuk-link--muted {
```
36degrees: Here be dragons:

The intention is that this would be used as a modifier, in addition to the govuk-link class.

```
<a href="#" class="govuk-link govuk-link--muted">Muted link</a>
```

_However_, if a user were to omit the `govuk-link` class, as long as the link was inside another element that set the font they would end up with a link that looked correct _except_ it would not have the focus state.

I don't know what the best approach is here. We _could_ ensure the focus state etc is included in the variants, but we'll end up with a lot of duplication in the CSS and it seems a little weird to have a variant that re-defines everything? Or we could not use modifiers for this sort of thing, and change it to e.g. `.govuk-muted-link` or `.govuk-link-muted`? Thoughts?

nickcolley:
It feels like maybe a bad play to remove focus states as a default.
Many people don't really understand the importance of focus states, right now GOV.UK Elements you don't really need to understand them for focus states to work most of the time.

With this change people could forget to include these classes. It feels like these might be a good global that should be overridden when necessary?

src/globals/scss/core/_prose-scope.scss
```diff
@@ -44,5 +44,9 @@
     ul {
       @extend .govuk-list--bullet;
     }
+
+    a {
```
36degrees: More dragons - if a user wants to use an a tag to define an anchor within a markdown document, e.g.

```
# <a name="section-1">Section 1</a>
```

Then any styling we apply to the .govuk-link class that is not inside a pseudo-selector will be applied to the anchor, despite it not actually being a link.

Bootstrap tackle this by [_undoing_ their link styles using `a:not([href]):not([tabindex])`](https://github.com/twbs/bootstrap/blob/v4-dev/scss/_reboot.scss#L210-L228) which we could add here, but that means the prose scope would need to 'know' what styling the `govuk-link` class is applying in order to undo it.

Alternatively, we could try and ensure we only ever add styling to pseudo selectors within the `govuk-link` class, but that seems hard to document / enforce and also means if anyone wants to override that styling they'll need to apply it to e.g. `:link, :visited` within their own component.

Again, bootstrap has discussed this [at some length](https://github.com/twbs/bootstrap/issues/19402) so worth reading that to try and learn from what they've tried…

Thoughts on a postcard…

nickcolley:
Talked to Ollie and there's plans to revisit globals later when we have more confidence with releasing/testing etc, so my comments about defaults don't block anything in this PR 👍

36degrees:
Good catch! Have updated the READMEs.

36degrees:
Going to leave this open for a bit to see if @hannalaakso, @alex-ju or @nickcolley have any ideas on those thorny issues. I don't _think_ either of them are really blocking, but if we can think of any way to address them that'd be a bonus.

src/globals/scss/core/_links.scss
nickcolley: I think we could move this conversation to the 'should focuses be global' conversation later and not block this PR.

src/globals/scss/core/_prose-scope.scss
nickcolley: This could be better explored when we have good variant coverage of the thorny parts of the prose components?

36degrees: Created a card to track this [here](https://trello.com/c/E9qnX9Sf/603-work-out-how-to-handle-non-linking-anchors-within-the-prose-scope).

src/globals/scss/core/_links.scss
36degrees: Documented as a comment on this card – https://trello.com/c/etYO3raM/598-review-global-focus-stance

PR #429, opened 2018-01-12
Add prose scope example

36degrees:
Add an example of the prose scope being applied to some example HTML so that contributors can understand how their changes to Frontend might affect it.

https://trello.com/c/Jtron4KD/574-add-prose-scope-example

### DIFF
--- a/src/views/examples/prose-scope/index.njk
+++ b/src/views/examples/prose-scope/index.njk
@@ -1,0 +1,69 @@
+{% extends "layout.njk" %}
+
+{% block content %}
+
+<div class="govuk-o-grid govuk-!-mt-r9">
+  <div class="govuk-o-grid__item govuk-o-grid__item--two-thirds">
+
+    <div class="govuk-prose-scope">
+
+      <h1>H1 heading</h1>
+
+      <p>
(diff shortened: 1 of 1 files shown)


joelanman:
@amyhupe and I had a discussion previously about sample content. It's a bit tricky using real guidance, as it can easily get out of date. One solution we thought of was using 'historical' content that is much less likely to change, say an announcement. Alternatively just ipsum lorem

36degrees:
This is exactly the same content as we currently have in one of the typography examples, so this isn't introducing anything new. Agree that we should review what content we're using generally, but I don't think we need to address it as part of this PR.

joelanman:
ah ok, good point. However, this sample doesn't cover uls and ols, which is provided by prose-scope - do we want a sample that will cover that? The sample in the design system is here:

https://govuk-design-system-production.cloudapps.digital/styles/typography/prose-scope/index.html

36degrees:
Good call, will update.

36degrees:
@joelanman have updated with a new example based on the example in the design system, but with headings directly following each other, an example of a longer list item and a few links.

PR #432, opened 2018-01-16
Rename "legal text" to "warning text"

dashouse:
- Updates every reference to "legal text" to "warning text"
- Updates the template to take text or html
- A 1px change to icon position :)

https://trello.com/c/45Fnu1yr/254-2-update-legal-text-component-api-and-name

PR #434, opened 2018-01-17
Use macros for examples

nickcolley:
This updates hardcoded HTML to use the Nunjucks macros in the examples.

See https://trello.com/c/8IciWs48/272-2-update-examples-to-use-nunjucks-templates-in-fe-app


src/views/examples/all-components/index.njk
```diff
@@ -2,69 +2,37 @@
 
 {% block content %}
 
-  {% include "../components/breadcrumb/breadcrumb.njk" %}
+  {% include "../../../components/breadcrumbs/breadcrumbs.njk" %}
```
alex-ju: do we need full relative paths here? it can be `{% include "breadcrumbs/breadcrumbs.njk" %}` since we [include the components path in Nunjucks views](https://github.com/alphagov/govuk-frontend/blob/master/app.js#L14)

36degrees: I think there's a few components missing from this…

- Back link
- Date input
- Error summary
- File upload
- Previous next
- Skip link
- Table
- Tag

src/views/examples/all-components/index.njk
```diff
@@ -2,69 +2,37 @@
...
-
-  {% include "../components/panel/panel.njk" %}
-
-  {% include "../components/phase-banner/phase-banner.njk" %}
-
-  {% include "../components/radio/radio.njk" %}
+  {# include "../../../components/label/label.njk" #}
 
-  {% include "../components/textarea/textarea.njk" %}
+  {% include "../../../components/legal-text/legal-text.njk" %}
```
36degrees: Needs rebasing after https://github.com/alphagov/govuk-frontend/pull/432?

src/views/examples/all-components/index.njk
```diff
@@ -2,69 +2,37 @@
...
-  {% call govukSiteWidthContainer() %}
-    {% call govukGrid(
-      classes='',
-      gridItems=[ { width: 'full' } ]
-      )
-    %}
-    Example full width grid column
-    <!-- Insert macros to sit inside grid item here -->
-    {% endcall %}
+  {% include "../../../components/textarea/textarea.njk" %}
```
36degrees: `textarea`, `select` - this is not how you alphabet! 🔤

src/views/examples/form-elements/index.njk
```diff
@@ -14,98 +25,156 @@
...
-      <label class="govuk-c-checkbox__label" for="waste-type-2">Waste from mines or quarries</label>
-    </div>
-    <div class="govuk-c-checkbox">
-      <input class="govuk-c-checkbox__input" id="waste-type-3" name="waste-types" type="checkbox" disabled="disabled" value="waste-animal">
-      <label class="govuk-c-checkbox__label" for="waste-type-3">Waste from animal carcasses</label>
-    </div>
-  </fieldset>
+  {% call govukFieldset() %}
+    <span id="checkbox-stacked"></span>
+    {{ govukRadios({
```
36degrees: Checkboxes?

src/views/examples/form-elements/index.njk
```diff
@@ -14,98 +25,156 @@
...
-    <div class="govuk-c-radio govuk-c-radio--inline">
-      <input class="govuk-c-radio__input" id="radio-1" type="radio" name="radio-group" value="Yes">
-      <label class="govuk-c-radio__label" for="radio-1">Yes</label>
-    </div>
-    <div class="govuk-c-radio govuk-c-radio--inline">
-      <input class="govuk-c-radio__input" id="radio-2" type="radio" name="radio-group" value="No">
-      <label class="govuk-c-radio__label" for="radio-2">No</label>
-    </div>
-  </fieldset>
+  {% call govukFieldset() %}
```
36degrees: Do want want to use these components 'properly' by providing the fieldset options to the radio / checkbox macros, and including a legend?

nickcolley: Nesting of components feels like the best way to put components together, feels intuitive to me anyway?

src/views/examples/all-components/index.njk
nickcolley: I was just trying to get this page rendering as it was before, should we remove it if no one was using it?

36degrees: Fair.

src/views/examples/form-elements/index.njk
36degrees: I think I'd still prefer to use them in the way we've designed them, as in the first example [here](http://govuk-frontend-review.herokuapp.com/components/radios), but it's not a big issue.

src/views/examples/form-elements/index.njk
```diff
@@ -1,111 +1,180 @@
...
-      <label class="govuk-c-checkbox__label" for="waste-type-2">Waste from mines or quarries</label>
-    </div>
-    <div class="govuk-c-checkbox">
-      <input class="govuk-c-checkbox__input" id="waste-type-3" name="waste-types" type="checkbox" disabled="disabled" value="waste-animal">
-      <label class="govuk-c-checkbox__label" for="waste-type-3">Waste from animal carcasses</label>
-    </div>
-  </fieldset>
+  {% call govukFieldset() %}
+    <span id="checkboxes-stacked"></span>
+    {{ govukRadios({
```
36degrees: These still don't seem to be checkboxes…

nickcolley: Oh I thought you meant the id :man_facepalming:

36degrees: 😆

src/views/examples/form-elements/index.njk
nickcolley: Will update, what decisions led to overloading the API? Feels better to keep the components distinct, as it's more like HTML?

PR #439, opened 2018-01-18
[Do not merge] Include previews in iframe examples

nickcolley:
[Do note merge] This isn't as tidy as I'd like it:

I'm not sure how this impacts README generation since I think they share this template, which is very confusing.

I've put styles inline since the project is not really set up to separate these out yet, should we do this?

I've including the iframe-resizer project in a similar way via a CDN for the same reason.

desktop-collapsed
desktop-open
mobile-collapsed
mobile-open


nickcolley:
Spoke to Jani and we agreed we should move these assets into better structured directories when we move `/views` into it's own `app` directory.

src/views/layout.njk
```diff
@@ -3,12 +3,60 @@
...
+
+    .app-\!-di {
+      display: inline;
+    }
+
+    .app-c-component-preview {
+      outline: 5px solid #e4f2ff;
+    }
+
+    .app-c-component-preview__iframe {
```
alex-ju: may be good to have a `min-height` here, otherwise we won't be able to preview the skip link component.

app.js
```diff
@@ -107,7 +107,9 @@ app.get('/components/:component/:example*?/preview', function (req, res, next) {
     {{ ${macroName}(${macroParameters}) }}`
   )
 
-  res.render('component-preview')
+  res.render('component-preview', {
+    bodyClasses: req.query.iframe && 'app-iframe-in-component-preview'
```
36degrees: Is this conditionally adding the `app-iframe-in-component-preview` only if `req.query.iframe` is truthy? I've not seen this construct before and it's a bit hard to understand what's going on.

If that's what it's doing, would definitely prefer using ternary instead, i.e.

```js
bodyClasses: req.query.iframe ? 'app-iframe-in-component-preview' : '';
```

src/views/layout.njk
```diff
@@ -3,12 +3,60 @@
...
   <link rel="stylesheet" href="/css/govuk-frontend-oldie.css">
   <![endif]-->
+  <style>
+    /* Injected into the preview page, that is then put in an iframe */
+    .app-iframe-in-component-preview {
+      margin-top: 10px;
+      margin-bottom: 0;
+    }
+
+    .app-iframe-in-component-preview:before {
```
36degrees: I'm guessing this is nuking the breakpoint thing? Definitely worth a comment.

src/views/macros/showExamples.njk
```diff
@@ -17,27 +17,44 @@
...
+          <iframe
+          class="js-component-preview app-c-component-preview__iframe"
+          src="{{ previewLink }}?iframe=true"
+          frameBorder="0"
+          allowTransparency="true"
+          scrolling="auto"
+          ></iframe>
+        </div>
+
+        {# TODO: Allow details to use nesting #}
```
36degrees: Can we use the details macro for this?

36degrees:
Looking a bit funky with custom colours:

screen shot 2018-01-19 at 14 53 30gmt

Those thick lines are especially disorientating / headache inducing if you're scrolling the page.

src/views/macros/showExamples.njk
nickcolley: No since the details macro doesnt allow nesting? Only strings?

src/views/layout.njk
nickcolley: Good shout

app.js
nickcolley: Ah fair enough, I'm used to this but definitely more boring ways to do this, will update.

nickcolley:
@36degrees not sure it'll be easy to reduce the lines in firefox with changed colours without compromising the normal view here?

src/views/layout.njk
nickcolley: At the moment the examples size exactly to the size they are (including margin) so you know if you've added or removed too much spacing when developing.

Could we instead a variant for skip-link that uses the `:focus` state?

36degrees:
Do we think the blue outlines are useful, or could we just remove them everywhere? In Firefox they are full width and because you can't see the left or right edge I think they just confuse rather than help.

src/views/layout.njk
alex-ju: Yes, if we'll have the skip link focused we can leave it as is.

nickcolley:
Hmm noticed some issues with breadcrumbs being clipped. Might have to abandon getting the previews exactly clipped to the component 😢

nickcolley:
This is how firefox with overridden colours looks now:

screen shot 2018-01-22 at 11 58 50

PR #442, opened 2018-01-18
Improve publishing docs

hannalaakso:
This updates the docs, especially around taking care when symlinking directories and versioning new components.

Also adds a section for those consuming `govuk-frontend` and needing to update to the latest version.

The doc has also moved the `dist` step to happen in the same PR.

### DIFF
--- a/docs/publishing.md
+++ b/docs/publishing.md
@@ -1,10 +1,14 @@
-## Publishing components
+# Publishing components
 
-We are using [lerna.js](https://lernajs.io/) to manage our packages and publish to npm.
+We are using [lerna.js](https://lernajs.io/) to manage our packages and publish
+to npm.
 
-Lerna is a tool that optimizes the workflow around managing multi-package repositories with git and npm.
+Lerna is a tool that optimizes the workflow around managing multi-package
+repositories with git and npm.
 
-It manages dependencies between our components so that if the version number of `globals` is bumped, all components dependent on `globals` will also have their version bumped.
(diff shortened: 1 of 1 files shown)


docs/publishing.md
```diff
@@ -1,96 +1,126 @@
...
-11. If your changes included new packages, you need to grant the test user access to them:
+13. If your changes included new packages, you need to grant npm test user access to them:
 ```
-npm access grant read-only govuk-frontend:test @govuk-frontend/button`
+npm access grant read-only govuk-frontend:test @govuk-frontend/warning-text
 ```
 If you removed any packages, revoke the test user's access to them:
 ```
-npm access revoke govuk-frontend:test @govuk-frontend/button`
+npm access revoke govuk-frontend:test @govuk-frontend/link
```
36degrees: There's issues with this which we discovered today – anyone who's using an old version of Frontend can no longer deploy as they're reliant on packages they can no longer see.

I think realistically we'll need to continue to keep old packages around going forward. (Though we should probably clean up old packages we know no-one's using now, before people outside of the team start using it)

igloosi: shall be remove this line then?

hannalaakso:
@nickcolley It doesn't seem `link` and `ln -s` do the same thing (you'd think so wouldn't you). 

I could use `rm @govuk-frontend` for unlinking to make the commands look less inconsistent but I prefer to use `unlink` instead of `rm`.

PR #445, opened 2018-01-19
Allow extra attributes to be added to warning text

36degrees:
As a bonus, improve the description for the `html` argument to make it clearer how it relates to the `text` argument.

https://trello.com/c/MrcWCSKx/621-warning-text-component-should-accept-additional-attributes

### DIFF
--- a/src/components/warning-text/index.njk
+++ b/src/components/warning-text/index.njk
@@ -82,7 +82,7 @@
         text: 'Yes'
       },
       {
-        text: 'The text next to the icon'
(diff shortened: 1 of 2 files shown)


alex-ju:
This PR doesn't have the associated/generated README.md commited, wich results in an 'orphan' `src/components/warning-text/README.md` 'hanging' in git.

nickcolley:
@alex-ju could we add a test for this?

nickcolley:
We do have a test for this that would fail but it's not being run in CI, will fix.

PR #446, opened 2018-01-19
Add ‘no visited state' link variant

### LINKED ISSUE #297: Add a link style for links with no visited state
There are some cases where it’s not appropriate to distinguish between visited and non-visited links.

An example is navigation links to pages with dynamic content, like dashboards in admin systems. The content on the page is changing all the time, so the fact that you’ve visited it before is not that important.

36degrees:
Add an ‘no visited state' link variant which can be used where it is not helpful to distinguish between visited and non-visited links.

For example, navigation links to pages with dynamic content like admin dashboards. The content on the page is changing all the time, so the fact that you’ve visited it before is not important.

Closes #297

https://trello.com/c/BF3hYuXJ/622-add-link-variant-for-links-with-no-visited-state

### DIFF
--- a/src/globals/scss/core/_links.scss
+++ b/src/globals/scss/core/_links.scss
@@ -63,4 +63,18 @@
       color: $govuk-black;
     }
   }
+
+  // 'No visited state' link variant
+  //
+  // Used in cases where it is not helpful to distinguish between visited and
+  // non-visited links.
+  //
+  // For example, navigation links to pages with dynamic content like admin
+  // dashboards. The content on the page is changing all the time, so the fact
+  // that you’ve visited it before is not important.
+  .govuk-link--no-visited-state {
+    &:visited {
+      color: $govuk-link-colour;
+    }
+  }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,11 @@ Breaking changes:
 
 New features:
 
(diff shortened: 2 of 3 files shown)


src/globals/scss/core/_links.scss
```diff
@@ -63,4 +63,18 @@
...
+  // 'Unvisited' link variant
+  //
+  // Used in cases where it is not helpful to distinguish between visited and
+  // non-visited links.
+  //
+  // For example, navigation links to pages with dynamic content like admin
+  // dashboards. The content on the page is changing all the time, so the fact
+  // that you’ve visited it before is not important.
+  .govuk-link--unvisited {
+    &:visited {
```
nickcolley: Can we leave this unnested?

36degrees: We can, though I think there is some value in being consistent with the way the rest of the file is written, making it easier to scan. I also personally think it'd be easy to miss if it was on the same line as the class name.

nickcolley: Ah sorry, I didn't see the rest of the file in this diff.

Sounds good to me 👍

36degrees:
@nickcolley have changed to `govuk-link--no-visited-state` which I think makes more sense having talked it through with Tim. Would you mind just re-checking it?

nickcolley:
@36degrees makes sense to me, a bit verbose but hopefully very clear 👍

36degrees:
Mind approving again? I dismissed your previous review.

36degrees:
@nickcolley Good shout. Done 👍

PR #447, opened 2018-01-22
Error if responsive mixins don't get a map

36degrees:
Add validation to the responsive spacing and typography mixins that expects maps, and errors if a user incorrectly passes in another value (for example, a value from the non-responsive spacing scale)

GitHub and Sublime Text are both struggling to correctly syntax-highlight `@error` calls for some reason, but it does work…

https://trello.com/c/OjWgpYkt/463-throw-errors-when-fe-spacing-mixins-used-wrong


nickcolley:
Epic 👍

PR #448, opened 2018-01-22
Add notes about creating a package.json to consume GOV.UK Frontend NPM package

igloosi:
As part of spiking [this ticket](https://trello.com/c/dcjfPeC8/86-spike-integrate-govuk-frontend-with-whatever-submit-is-using) I've found we don't say how you can consume NPM packages if the application is not setup to do so.

This PR amends the installation notes to describe basic requirements.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,15 +9,36 @@ A single package containing everything needed to start building a GOV.UK service
 
 :rotating_light: **This project is in development and these instructions are likely to change.** :rotating_light:
 
-Guidance for building a GOV.UK service - [GOV.UK Design System](TODO: Insert link to GOV.UK Design System).
+Guidance for building a GOV.UK service - [GOV.UK Design System](https://govuk-design-system-production.cloudapps.digital/).
 
-## Quick start
+## Requirements
 
-:rotating_light: **This will not work until packages are published to npm.** :rotating_light:
+1) [NodeJS](https://nodejs.org/en/) installed
+2) [package.json](https://docs.npmjs.com/files/package.json) file
+
+This file is needed to install a GOV.UK Frontend NPM package in your application. Most often it resides in the root of your application. You can create this file interactively using a `npm init` command.
+
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -39,6 +39,10 @@ Include the CSS and JavaScript
 
 ## Installation
 
+To install GOV.UK Frontend NPM package in your application, need to create a [package.json](https://docs.npmjs.com/files/package.json) file in your application (if you haven't done so already) and add a package to it's devDependencies section.
```
36degrees: It'd need to be installed as a _non-dev_ dependency, right? You'd need Frontend to be included to actually use the application…

igloosi: true. updated

README.md
```diff
@@ -39,6 +39,10 @@ Include the CSS and JavaScript
 
 ## Installation
 
+To install GOV.UK Frontend NPM package in your application, need to create a [package.json](https://docs.npmjs.com/files/package.json) file in your application (if you haven't done so already) and add a package to it's dependencies section.
```
nickcolley: Should we be telling people to add packages to their package.json and then below telling them to run a command that would do this for them?

igloosi: duh :( fixed

README.md
```diff
@@ -39,6 +39,10 @@ Include the CSS and JavaScript
 
 ## Installation
 
+To install GOV.UK Frontend NPM package in your application, need to create a [package.json](https://docs.npmjs.com/files/package.json) file in your application (most often in the root of your application).
```
joelanman: '**you** need to create'

I'm not sure if this is the best way to talk about it. I *think* `npm install` would do this for you, or maybe `npm init` and then `npm install`

igloosi: they need to create it, but yes can can add a line than it can be created with `npm init`

igloosi:
@joelanman updated

PR #449, opened 2018-01-22
Expand legacy IE stylesheet documetation

igloosi:
As part of spiking [this ticket](https://trello.com/c/dcjfPeC8/86-spike-integrate-govuk-frontend-with-whatever-submit-is-using) I've found we don't say how you can generate your own legacy IE stylesheet when consuming GOV.UK Frontend partials.

This PR adds notes about generating own stylesheet from our partials.

### DIFF
--- a/docs/legacy-ie.md
+++ b/docs/legacy-ie.md
@@ -4,7 +4,36 @@ Unless stated otherwise the approach to browser support still follows the GDS pr
 
 We want to continue providing a functional experience for user of legacy browsers, but in a way that reduces complexity in this codebase.
 
-## Separate stylesheet
+## Generate your own legacy IE stylesheet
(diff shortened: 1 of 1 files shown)


docs/legacy-ie.md
```diff
@@ -4,7 +4,36 @@ Unless stated otherwise the approach to browser support still follows the GDS pr
...
+
+In our partials we use IE mixins that generate code for versions of IE:
+
+```
+@include ie-lte(8) {
+  border-bottom: $button-shadow-size solid $govuk-button-colour-darken-15;
+}
+```
+
+as well [SassMQ functionality](https://sass-mq.github.io/sass-mq/#undefined-variable-mq-responsive) that flattens media queries.
```
joelanman: Not sure this is as clear as it could be. Something like 'You need to set responsive to false, otherwise IE8 will get mobile styles, because it doesn't support media queries'?

PR #454, opened 2018-01-22
Add generated warning-text README

nickcolley:
Missed from https://github.com/alphagov/govuk-frontend/pull/445

Will follow up with a test that stops that will fail CI if this happens again


36degrees:
👍

PR #455, opened 2018-01-22
Replace Mocha/Chai with Jest, re-enable task tests, add back-link component

nickcolley:
This pull request:

- replaces mocha + chai with Jest
- re-enables the tests that check build:packages and build:dist work
- adds a test for the back-link component

Note: this approach avoids gulp to chain the commands, this is to avoid unnecessary abstraction which might be controversial main con is npm scripts is verbose. Comments welcome.

With special thanks to @htmlandbacon and @tyom for sharing their approaches.

As part of https://trello.com/c/UHYdyb4w/628-add-jest-testing-suite-to-govuk-frontend


lib/jest-helpers.js
```diff
@@ -0,0 +1,40 @@
...
+const cheerio = require('cheerio')
+
+const COMPONENT_DIR = 'src/components'
+
+nunjucks.configure(COMPONENT_DIR, {
+  trimBlocks: true,
+  lstripBlocks: true
+})
+
+function render (name, params) {
```
36degrees: Would be great to add some comments in this file. For example, from the point of view of someone new to JS testing who's writing their first test for something they're contributing, I think it'd be useful to understand what Cheerio is and what it is doing, and how you use the return value when it returns two things `{ output, $ }`.

I'd also suggest maybe using `componentName` and `arguments` as the parameters. I _think_ we're starting to standardise on arguments for macros, but we should maybe document our terminology somewhere.

lib/jest-helpers.js
```diff
@@ -0,0 +1,40 @@
...
+  const output = nunjucks.render(name + '/template.njk', { params })
+  const $ = cheerio.load(output)
+  return { output, $ }
+}
+
+const yaml = require('js-yaml')
```
36degrees: Is this conventional? I'd expect to see all includes at the top of the file.

lib/jest-helpers.js
```diff
@@ -0,0 +1,40 @@
...
+const path = require('path')
+const fs = require('fs')
+
+function getExamples (name) {
+  const file = fs.readFileSync(
```
36degrees: What happens if you try to get examples for a non-existent component? Is the error message clear? If not, can we add error handling to this function.

lib/jest-helpers.js
```diff
@@ -0,0 +1,40 @@
...
+
+const COMPONENT_DIR = 'src/components'
+
+nunjucks.configure(COMPONENT_DIR, {
+  trimBlocks: true,
+  lstripBlocks: true
+})
+
+function render (name, params) {
+  const output = nunjucks.render(name + '/template.njk', { params })
```
36degrees: What happens if you try to render a non-existent component? Is the error message clear? If not, can we add error handling to this function.

src/components/back-link/template.test.js
```diff
@@ -0,0 +1,74 @@
...
+  })
+
+  it('renders custom text correctly', () => {
+    const { $ } = render(COMPONENT_NAME, {
+      href: '#',
+      text: 'Home'
+    })
+
+    const $component = $('.govuk-c-back-link')
+    expect($component.text()).toEqual('Home')
```
36degrees: Can we have a test that ensures HTML is correctly escaped when passed as `text`?

src/components/back-link/template.test.js
```diff
@@ -0,0 +1,74 @@
...
+    // TODO: href is a required field but the component does not error,
+    // when it is not passed
+    expect(() => {
+      render(COMPONENT_NAME, {
+        text: 'Example'
+      })
+    }).toThrow()
+  })
+
+  it('renders the default with an anchor, href and text correctly', () => {
```
36degrees: the default what? "renders the default example"?

src/components/back-link/template.test.js
```diff
@@ -0,0 +1,74 @@
+/* globals describe, it, expect */
+
+const { render, getExamples } = require('../../../lib/jest-helpers')
+
+const COMPONENT_NAME = 'back-link'
```
36degrees: Not 100% sure this is worth the comprehension cost - seems like an extra connection for contributors to make?

36degrees: Worth noting, we also specify the class name in every test and it feels like if we continue down this path we would make that a constant too…?

e.g.

```
const { $ } = render(COMPONENT_NAME, examples.default)

const $component = $(COMPONENT_CLASS)
```

which I think could be a bit intimidating for new contributors.

lib/jest-helpers.js
nickcolley: Sounds good, do you think contributors will look at this file?

Isn't params what we're using everywhere? https://github.com/alphagov/govuk-frontend/search?utf8=%E2%9C%93&q=params&type=

lib/jest-helpers.js
nickcolley: Will check this, good shout.

I'll write a test for this...

lib/jest-helpers.js
nickcolley: I was grouping with where they're used, agree lets move these 👍

src/components/back-link/template.test.js
nickcolley: Agree let's remove this constant

lib/jest-helpers.js
nickcolley:
``` FAIL  src/components/back-link/template.test.js
  ● Test suite failed to run

    ENOENT: no such file or directory, open 'src/components/back-links/back-links.yaml'
```

Seems pretty clear to me, let me know what you think.

lib/jest-helpers.js
36degrees: …yep.

Good point, well made.

lib/jest-helpers.js
nickcolley:
``` 
FAIL  src/components/back-link/template.test.js
  back-link component
    ✕ renders the default with an anchor, href and text correctly (5ms)
    ✓ renders classes correctly (12ms)
    ✓ renders custom text correctly (3ms)
    ✓ renders html correctly (1ms)
    ✓ renders attributes correctly (3ms)
    ○ skipped 1 test

  ● back-link component › renders the default with an anchor, href and text correctly

    template not found: back-links/template.njk
```

lib/jest-helpers.js
```diff
@@ -0,0 +1,39 @@
+'use strict'
+
+const path = require('path')
+const fs = require('fs')
+const nunjucks = require('nunjucks')
+const cheerio = require('cheerio')
+const yaml = require('js-yaml')
+
+const COMPONENT_DIR = 'src/components'
```
alex-ju: Missed this one, would've been good to use `config/paths.json` so we have all the paths in one place.

nickcolley: Will do a PR

src/components/back-link/template.test.js
```diff
@@ -0,0 +1,82 @@
...
+        'data-test': 'attribute',
+        'aria-label': 'Back to home'
+      },
+      href: '#',
+      html: 'Back'
+    })
+
+    const $component = $('.govuk-c-back-link')
+    expect($component.attr('data-test')).toEqual('attribute')
+    expect($component.attr('aria-label')).toEqual('Back to home')
```
tvararu: If there is a way to get the entire list of attributes, you could combine that with a [jest snapshot](https://facebook.github.io/jest/docs/en/snapshot-testing.html) to very easily test the whole list without writing too many bulky test assertions. :)

This is very cool either way 🌟

nickcolley: At the moment the logic is pretty basic so I think it's covered by these assertions, definitely sounds like a good idea otherwise though.

36degrees: (Hi Theo! 👋)

PR #456, opened 2018-01-23
Decouple app from src

alex-ju:
This PR moves Nunjucks layouts and templates from `src` to `app` and is part of a process to separate development code from published code.

[Trello card](https://trello.com/c/l31FBTjx)


app.js
```diff
@@ -24,15 +27,15 @@ let env = nunjucks.configure(appViews, {
...
 env.addFilter('componentNameToMacroName', helperFunctions.componentNameToMacroName)
 
 // Set view engine
 app.set('view engine', 'njk')
 
 // Set up middleware to serve static assets
-app.use('/public', express.static(path.join(__dirname, '/public')))
-app.use('/icons', express.static(path.join(__dirname, '/public/icons')))
+app.use('/public', express.static(path.join(__dirname, configPaths.public)))
+app.use('/icons', express.static(path.join(__dirname, configPaths.public + 'icons')))
```
36degrees: Can we rely on path.join to join `configPaths.public` and `'icons'`?

i.e.

```
path.join(__dirname, configPaths.public, 'icons')
```

rather than

```
path.join(__dirname, configPaths.public + 'icons'))
```

app.js
```diff
@@ -5,13 +5,16 @@ const fs = require('fs')
 const path = require('path')
```
36degrees: Could/should this file move into `app` as well?

tasks/gulp/copy-to-destination.js
```diff
@@ -26,11 +26,11 @@ gulp.task('copy-files', () => {
     '!' + configPaths.src + 'globals/icons',
     '!' + configPaths.src + 'globals/scss',
     '!' + configPaths.src + 'globals',
-    '!' + configPaths.src + 'views',
     '!' + configPaths.src + 'components',
-    '!' + configPaths.src + 'views/**',
-    '!' + configPaths.src + 'examples',
-    '!' + configPaths.src + 'examples/**',
+    '!' + configPaths.app,
```
36degrees: We start out with `configPaths.src`, so all of app should be excluded already anyway?

tasks/gulp/generate-readme.js
```diff
@@ -33,7 +33,7 @@ gulp.task('generate:readme', () => {
     objectData.componentName = path.dirname(file.path).split(path.sep).slice(-1).toString()
     objectData.componentPath = objectData.componentName
     try {
-      let componentData = yaml.safeLoad(fs.readFileSync(`src/components/${objectData.componentName}/${objectData.componentName}.yaml`, 'utf8'), {json: true})
+      let componentData = yaml.safeLoad(fs.readFileSync(configPath.components + `${objectData.componentName}/${objectData.componentName}.yaml`, 'utf8'), {json: true})
```
36degrees: Can we use `path.join` here rather than trying to do it ourselves?

app.js
36degrees: If we did this, it might then make sense to put all of the views under `app/views`?

app.js
```diff
@@ -58,7 +61,7 @@ app.get('/', function (req, res) {
 // Whenever the route includes a :component parameter, read the component data
 // from its YAML file
 app.param('component', function (req, res, next, componentName) {
-  let yamlPath = `src/components/${componentName}/${componentName}.yaml`
+  let yamlPath = configPaths.components + `${componentName}/${componentName}.yaml`
```
36degrees: 👏

app.js
```diff
@@ -75,7 +78,7 @@ app.get('/components/:component', function (req, res, next) {
   // make variables available to nunjucks template
   res.locals.componentPath = req.params.component
 
-  res.render(`${req.params.component}/index`, function (error, html) {
+  res.render('../../' + configPaths.components + `${req.params.component}/index`, function (error, html) {
```
36degrees: Why do we need to specify the entire path if `configPaths.components` is in the appViews?

tasks/gulp/copy-to-destination.js
alex-ju: Ha, good one. Thanks!

app.js
```diff
@@ -5,13 +5,16 @@ const fs = require('fs')
 const path = require('path')
 const port = (process.env.PORT || 3000)
 const yaml = require('js-yaml')
-const helperFunctions = require('./lib/helper-functions.js')
-const dto = require('./lib/dto-helper.js')
+
+const helperFunctions = require('./lib/helper-functions.js')  // misc helpers
+const dto = require('./lib/dto-helper.js')                    // directoryToObject helper
```
nickcolley: Can we avoid single character variables?

E.g. `const directoryToObject = require('./lib/helpers/directory-to-object.js')`

app.js
nickcolley: Can we use a consistent way of concatenation here?

```js
`${configPaths.components}${componentName}/${componentName}.yaml`
```

nickcolley:
Looking good 👍 , agree with Ollie we could group views via `app/views/**/*`

app.js
alex-ju: Yep, you're right both should be moved. They were in the file structure, but I rushed into publishing the branch.

app.js
alex-ju: Sounds good, but we'll end up having `directoryToObject.directoryToObject()`

alex-ju:
Updated with requested changes.
Now `app.js` is in the `app` directory and all the templates are grouped in `views`

app/app.js
```diff
@@ -43,8 +47,8 @@ const server = app.listen(port, () => {
 // Index page - render the component list template
 app.get('/', function (req, res) {
   Promise.all([
-    dto.directoryToObject(path.resolve('./src/components')),
-    dto.directoryToObject(path.resolve('./src/views/examples'))
+    directoryToObject.directoryToObject(path.resolve(configPaths.components)),
```
36degrees: If we change

```
exports.directoryToObject = directoryToObject
```

to

```
exports = directoryToObject
```

in `lib/dto-helper.js` then I _think_ we can just make this `directoryToObject(path.resolve(configPaths.components))`

?

tasks/gulp/generate-readme.js
```diff
@@ -33,7 +33,8 @@ gulp.task('generate:readme', () => {
     objectData.componentName = path.dirname(file.path).split(path.sep).slice(-1).toString()
     objectData.componentPath = objectData.componentName
     try {
-      let componentData = yaml.safeLoad(fs.readFileSync(`src/components/${objectData.componentName}/${objectData.componentName}.yaml`, 'utf8'), {json: true})
+      let componentPath = path.join(configPath.components, `${objectData.componentName}/${objectData.componentName}.yaml`)
```
36degrees: Suggest letting path.join do it's work here
```
let componentPath = path.join(configPath.components, objectData.componentName, `${objectData.componentName}.yaml`)
```

alex-ju: There may be other places where we can use the `path.join` without passing a concatenated string.

app/app.js
alex-ju: mm..sounds good, but it has to be `module.exports`. updated.

alex-ju:
Rebased to master. Ready to be merged.

PR #457, opened 2018-01-24
Add tests for `govuk-c-tag` component

hannalaakso:
Add Jest tests for `govuk-c-tag` component.

These test if the component:

- renders the default example with strong tag and text correctly
- renders classes correctly
- renders custom text correctly
- renders escaped html when passed to text
- renders html correctly
- renders attributes correctly

Remove tag `--with-html` example from FE review app. This follows our decision that examples from the component definition that are now redundant and no longer useful in the review app are removed.

Trello card: https://trello.com/c/2Mv7onR8/637-automated-tests-for-tag-component

### DIFF
--- a/src/components/tag/template.test.js
+++ b/src/components/tag/template.test.js
@@ -1,0 +1,66 @@
+/* globals describe, it, expect */
+
+const { render, getExamples } = require('../../../lib/jest-helpers')
+
+const examples = getExamples('tag')
+
+describe('Tag', () => {
+  it('renders the default example with strong element and text', () => {
+    const { $ } = render('tag', examples.default)
+
+    const $component = $('.govuk-c-tag')
+    expect($component.get(0).tagName).toEqual('strong')
+    expect($component.text()).toContain('alpha')
+  })
+
+  it('renders classes', () => {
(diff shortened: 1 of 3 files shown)


igloosi:
this looks good from my POV

src/components/tag/template.test.js
```diff
@@ -0,0 +1,66 @@
...
+      text: 'alpha'
+    })
+
+    const $component = $('.govuk-c-tag')
+    expect($component.hasClass('govuk-c-tag--inactive')).toBeTruthy()
+  })
+
+  it('renders custom text correctly', () => {
+    const { $ } = render('tag', {
+      text: 'alpha'
```
36degrees: This is the same text as in the default example, so we could change the logic to hard code `alpha` and I think all of these tests would still pass? Suggest specifying some text that would not normally appear in any of our examples.

hannalaakso:
This is ready to review. 

I made the description of the tests resemble more closely the sort of natural language that we agreed on with the team.

src/components/tag/template.test.js
hannalaakso: Done!

PR #461, opened 2018-01-24
Add tests for `govuk-c-button` component

nickcolley:
We break this down into four groups (as represented by describe blocks).

1. Input buttons
2. Link buttons
3. Explicit buttons
4. Implicit buttons

Also removes examples in `button.yaml` that are now covered by tests.

As per https://trello.com/c/LzkwDRYe/639-automated-tests-for-button-component

### DIFF
--- a/src/components/button/template.test.js
+++ b/src/components/button/template.test.js
@@ -1,0 +1,242 @@
+/* globals describe, it, expect */
+
+const { render, getExamples } = require('../../../lib/jest-helpers')
+
+const examples = getExamples('button')
+
+describe('Button', () => {
+  describe('input[type=submit]', () => {
+    it('renders the default example', () => {
+      const { $ } = render('button', examples.default)
+
+      const $component = $('.govuk-c-button')
+      expect($component.get(0).tagName).toEqual('input')
+      expect($component.attr('type')).toEqual('submit')
+      expect($component.val()).toEqual('Save and continue')
+    })
+
+    it('renders with attributes', () => {
+      const { $ } = render('button', {
+        element: 'input',
+        attributes: {
+          'aria-controls': 'example-id',
+          'data-tracking-dimension': '123'
+        }
+      })
+
+      const $component = $('.govuk-c-button')
+      expect($component.attr('aria-controls')).toEqual('example-id')
+      expect($component.attr('data-tracking-dimension')).toEqual('123')
+    })
+
+    it('renders with classes', () => {
+      const { $ } = render('button', {
+        element: 'input',
+        classes: 'app-c-button--custom-modifier'
+      })
+
+      const $component = $('.govuk-c-button')
+      expect($component.hasClass('app-c-button--custom-modifier')).toBeTruthy()
+    })
+
+    it('renders with disabled', () => {
+      const { $ } = render('button', examples.disabled)
+
+      const $component = $('.govuk-c-button')
+      expect($component.attr('aria-disabled')).toEqual('true')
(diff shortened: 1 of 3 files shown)


nickcolley:
This PR includes duplicate testing for inputs based on an assumption that the forking elements could break 'attributes' 'classes' (etc) from being applied.

I'm not sure if we need to be so strict. Comments welcome.

nickcolley:
@36degrees I've reworded the tests so they read correctly, I'm not sure how to break them up as you described without losing track of the code coverage, if you want to do more than in this PR let's catch up and do it together.

src/components/button/template.test.js
```diff
@@ -0,0 +1,254 @@
...
+        text: 'Start now'
+      })
+
+      const $component = $('.govuk-c-button')
+      expect($component.attr('type')).toEqual('button')
+    })
+  })
+
+  describe('implicit button element', () => {
+    it('renders a link if you pass an href', () => {
```
36degrees: Context is 'implicit button element' but this is testing for an implicit link?

src/components/button/template.test.js
```diff
@@ -0,0 +1,254 @@
...
+
+    it('renders an input[type=submit] if you don\'t pass anything', () => {
+      const { $ } = render('button', {
+        text: 'Save and continue'
+      })
+
+      const $component = $('.govuk-c-button')
+      expect($component.get(0).tagName).toEqual('input')
+      expect($component.attr('type')).toEqual('submit')
+      expect($component.val()).toContain('Save and continue')
```
36degrees: I don't think this should be being expected as part of this test.

src/components/button/template.test.js
```diff
@@ -0,0 +1,254 @@
...
+
+  describe('implicit button element', () => {
+    it('renders a link if you pass an href', () => {
+      const { $ } = render('button', {
+        href: '/'
+      })
+
+      const $component = $('.govuk-c-button')
+      expect($component.get(0).tagName).toEqual('a')
+      expect($component.attr('href')).toEqual('/')
```
36degrees: This doesn't relate to the test description. I think we should just be testing that it's rendering a link.

src/components/button/template.test.js
```diff
@@ -0,0 +1,254 @@
...
+    })
+
+    it('renders a button if you pass html', () => {
+      const { $ } = render('button', {
+        html: 'Start <em>now</em>'
+      })
+
+      const $component = $('.govuk-c-button')
+      expect($component.get(0).tagName).toEqual('button')
+      expect($component.html()).toContain('Start <em>now</em>')
```
36degrees: Again, I think we should just be testing that we render a `button`.

src/components/button/template.test.js
```diff
@@ -0,0 +1,254 @@
...
+      })
+
+      const $component = $('.govuk-c-button')
+      expect($component.attr('type')).toEqual('button')
+    })
+  })
+
+  describe('link', () => {
+    it('renders with anchor, href and an accessible role of button', () => {
+      const { $ } = render('button', examples.link)
```
36degrees: The `link` example is an implicit link, because it specifies an `href` but not an `element`. So I think this is testing the same thing as the implicit link test on [`:223`](https://github.com/alphagov/govuk-frontend/pull/461/files#diff-4411da34c01789c52a5e7379b28380d5R223)?

nickcolley:
@36degrees updated based on your comments

nickcolley:
One last thing - I need to add an entry to the changelog.

PR #463, opened 2018-01-25
Use config/paths instead of constant variable

nickcolley:


lib/jest-helpers.js
```diff
@@ -21,7 +20,7 @@ function render (componentName, params) {
 
 function getExamples (componentName) {
   const file = fs.readFileSync(
-    path.join(COMPONENT_DIR, `${componentName}/${componentName}.yaml`),
+    path.join(configPaths.components, `${componentName}/${componentName}.yaml`),
```
36degrees: Suggest letting path.join work it's magic here:

```
path.join(configPaths.components, componentName, `${componentName}.yaml`)
```

nickcolley: Updated 👍

PR #464, opened 2018-01-25
Add spacing for nested lists

hannalaakso:
This PR:

- Add top margin for lists within lists
- Adds an example to frontend app
- Updates changelog

To review, view: `/examples/typography`

Trello card: https://trello.com/c/vuqWoun4/517-spacing-for-nested-lists-to-be-added-to-frontend

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,14 +21,15 @@ Fixes:
 
 - Remove double margin from Date input component
   (PR [#451](https://github.com/alphagov/govuk-frontend/pull/451))
+- Add top margin for nested lists (PR [#464](https://github.com/alphagov/govuk-frontend/pull/464))
+
 
 Internal:
 
 - Replace Mocha/Chai with Jest, re-enable task tests, add back-link component.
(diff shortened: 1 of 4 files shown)


nickcolley:
In your example this seems to put a margin-top of 10px on the root level component, instead of just the nested lists, is this intended?

nickcolley:
screen shot 2018-01-25 at 11 05 10

hannalaakso:
Good catch, not intended. Fixed now.

nickcolley:
In your amended commit, does this now only work for `.govuk-list` usage and not `.govuk-prose`?

hannalaakso:
This should work for `.govuk-list` and for lists within `.govuk-prose-scope`. I've just added an example to `/examples/prose-scope`

nickcolley:
Sweet 👍 

If anyone else is interested it gets mixed in here: https://github.com/alphagov/govuk-frontend/blob/0091fc8edd2521cff5bdd1addd2de38d2d533253/packages/globals/core/_prose-scope.scss#L35

PR #467, opened 2018-01-25
Fix test JavaScript appearing in compiled assets

nickcolley:


tasks/gulp/compile-assets.js
```diff
@@ -79,7 +79,10 @@ gulp.task('scss:compile', () => {
 // Compile js task for preview ----------
 // --------------------------------------
 gulp.task('js:compile', () => {
-  return gulp.src([configPaths.src + '**/*.js'])
+  return gulp.src([
+    '!' + configPaths.src + '**/*.test.js',
```
36degrees: Can we flip this so that we start out with a whitelist and then blacklist stuff? I imagine the outcome is the same, but I just think it makes the intent clearer.

nickcolley: Was keeping consistent with https://github.com/alphagov/govuk-frontend/blob/master/tasks/gulp/compile-components.js#L18

36degrees: Fair enough 👍

PR #472, opened 2018-01-29
Add tests for table component

igloosi:
We break this down into groups (as represented by describe blocks).

Trello https://trello.com/c/mdqq5RnP/645-automated-tests-for-table-component

### DIFF
--- a/src/components/table/template.test.js
+++ b/src/components/table/template.test.js
@@ -1,0 +1,371 @@
+/* globals describe, it, expect */
+
+const { render, getExamples } = require('../../../lib/jest-helpers')
+
+const examples = getExamples('table')
+describe('Table', () => {
+  it('renders with classes', () => {
+    const $ = render('table', {
+      'classes': 'custom-class-goes-here',
+      'rows': [
+        [
+          {
+            'text': 'January'
+          },
+          {
+            'text': '£85',
+            'format': 'numeric'
+          }
+        ],
+        [
+          {
+            'text': 'February'
+          },
+          {
+            'text': '£165',
+            'format': 'numeric'
+          }
+        ]
+      ]
+    })
+    const $component = $('.govuk-c-table')
+
+    expect($component.hasClass('custom-class-goes-here')).toBeTruthy()
+  })
+
+  it('renders with caption text', () => {
+    const $ = render('table', examples['table-with-caption-and-head'])
+    const $caption = $('.govuk-c-table__caption')
+
+    expect($caption.text()).toBe('Caption 1 : Months and rates')
+  })
+
+  it('renders the caption size class', () => {
+    const $ = render('table', examples['table-with-caption-and-head'])
+    const $caption = $('.govuk-c-table__caption')
+
+    expect($caption.hasClass('small')).toBeTruthy()
+  })
+
+  it('renders first cell in every row as a <th> element with correct `govuk-c-table__header` class and `scope=row` attribute', () => {
+    const $ = render('table', {
+      'firstCellIsHeader': true,
+      'rows': [
+        [
+          {
+            'text': 'January'
+          },
+          {
+            'text': '£85',
+            'format': 'numeric'
+          },
+          {
+            'text': '£95',
+            'format': 'numeric'
+          }
+        ],
+        [
+          {
+            'text': 'February'
+          },
+          {
+            'text': '£75',
+            'format': 'numeric'
+          },
+          {
+            'text': '£55',
+            'format': 'numeric'
+          }
+        ]
+      ]
+    })
+    const $component = $('.govuk-c-table')
+    const $tableBody = $component.find('.govuk-c-table .govuk-c-table__body')
+    const $firstTableRow = $tableBody.find('.govuk-c-table__row:first-child')
+    const $firstTableHeader = $firstTableRow.find('.govuk-c-table__header:first-child')
+
+    expect($firstTableHeader.get(0).tagName).toEqual('th')
+    expect($firstTableHeader.attr('scope')).toEqual('row')
+    expect($firstTableHeader.hasClass('govuk-c-table__header')).toBeTruthy()
+
+    const $lastTableRow = $tableBody.find('.govuk-c-table__row:last-child')
+    const $lastTableHeader = $lastTableRow.find('.govuk-c-table__header:first-child')
+
+    expect($lastTableHeader.get(0).tagName).toEqual('th')
+    expect($lastTableHeader.attr('scope')).toEqual('row')
+    expect($lastTableHeader.hasClass('govuk-c-table__header')).toBeTruthy()
+  })
+
+  it('renders with thead', () => {
+    const args = examples['table-with-head']
+    const $ = render('table', args)
+
+    const $component = $('.govuk-c-table')
+    const $tableHead = $component.find('.govuk-c-table__head')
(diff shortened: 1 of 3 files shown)


36degrees:
I think I would agree with Nick here. I find it really hard to follow the tests and to understand what the intent is. I think there's as much chance of introducing bugs in the logic of the tests as there is in having bugs in the code we're trying to test.

https://github.com/mawrkus/js-unit-testing-guide#avoid-logic-in-your-tests

igloosi:
was hoping to cut down on the number of loops for the default example for each cell and row, when checking data

nickcolley:
@igloosi I suppose since we only have three rows it'd look like

```
loop 3
  assertion
endloop
```

```
assertion
assertion
assertion
```

So the compromise is verbose tests, which for me personally is OK since looping behaviours are harder to grok and easier to hide bad assertions?

igloosi:
cool, i'll rewrite for more verbosity

36degrees:
How do we feel about the tests asserting against their own inputs?

i.e. doing things like
```
args = { "text": "bar" }
result = foo(args)

expect(result.text).toEqual(args.text)
```

rather than

```
result = doSomething({"text": "bar"})

expect(result.text).toEqual("bar")
```

I would generally prefer the latter, as I think: 

- it makes the intention clearer (as someone reading the tests, you don't have to follow and parse variable declarations to understand what we want to be happening)
- it is a lot harder to introduce bugs into your test
- it's how we're doing things in all (?) the other test cases so far

src/components/table/template.test.js
```diff
@@ -0,0 +1,332 @@
...
+          }
+        ]
+      ]
+    }
+    const { $ } = render('table', args)
+    const $component = $('.govuk-c-table')
+    const $componentBody = $component.children('tbody')
+    const $componentBodyRow = $componentBody.children('tr')
+
+    expect(parseInt($componentBodyRow.eq(0).children().eq(1).attr('colspan'), 10)).toBe(args.rows[0][1].colspan)
```
36degrees: If we used `toEqual` here, could we get rid of the casting?

igloosi:
will revise

src/components/table/template.test.js
igloosi: hmmm, with toEqual:

 Expected value to equal:
      2
 Received:
     "2"

src/components/table/template.test.js
```diff
@@ -0,0 +1,362 @@
...
+      expect($tableHead).toHaveLength(1)
+      expect($tableHeadRow).toHaveLength(1)
+      expect($tableHeadCell).toHaveLength(3)
+    })
+
+    it('renders a caption', () => {
+      const { $ } = render('table', examples['table-with-caption-and-head'])
+      const $caption = $('.govuk-c-table__caption')
+
+      expect($caption).toHaveLength(1)
```
nickcolley: I don't think we need this assertion since the next one down will fail if the caption does not exist.

src/components/table/template.test.js
```diff
@@ -0,0 +1,362 @@
...
+    it('renders correct text in cell 2 of row 1', () => {
+      expect($tableRow.eq(0).find($tableCell).eq(1).text()).toEqual('£85')
+    })
+
+    it('renders correct text in cell 3 of row 1', () => {
+      expect($tableRow.eq(0).find($tableCell).eq(2).text()).toEqual('£95')
+    })
+  })
+
+  describe('variant', () => {
```
nickcolley: Does this need to be in a describe block? What is it describing?

src/components/table/template.test.js
```diff
@@ -0,0 +1,362 @@
...
+            'format': 'numeric'
+          }
+        ]
+      ]
+    })
+    const $component = $('.govuk-c-table')
+    const $tableRow = $component.find('.govuk-c-table .govuk-c-table__body .govuk-c-table__row')
+
+    expect($tableRow.eq(0).children().eq(0).get(0).tagName).toEqual('th')
+    expect($tableRow.eq(0).children().eq(0).hasClass('govuk-c-table__header')).toBeTruthy()
```
nickcolley: We could swap this around and make this a bit easier to read by doing

Instead of 


```js
expect($tableRow.eq(0).children().eq(0).get(0).tagName).toEqual('th')
 expect($tableRow.eq(0).children().eq(0).hasClass('govuk-c-table__header')).toBeTruthy()
```

try

```js
const $tableBody = $component.find('.govuk-c-table .govuk-c-table__body')

const $firstTableRow = $tableBody.find('.govuk-c-table__row:first-child')
const $firstTableHeader = $firstTableRow.find('govuk-c-table__header:first-child')
expect($firstTableHeader.get(0).tagName).toEqual('th')

const $lastTableRow = $tableBody.find('.govuk-c-table__row:last-child')
const $lastTableHeader = $lastTableRow.find('govuk-c-table__header:last-child')
expect($lastTableHeader.get(0).tagName).toEqual('th')

```

src/components/table/template.test.js
alex-ju: could we get these two tests outside 'variant', even if they're using examples from yaml. also, and I know I'm being picky with this one, but can we order them somehow in a similar way we have them in the table or arguments (classes, caption..) would make it easier to follow and make sure we cover all the arguments.

igloosi: removed

igloosi:
@nickcolley  and @alex-ju  update based on comments

src/components/table/template.test.js
```diff
@@ -0,0 +1,374 @@
+/* globals describe, it, expect */
+
+const { render, getExamples } = require('../../../lib/jest-helpers')
+
+const examples = getExamples('table')
+describe('Table', () => {
+
```
alex-ju: _Block must not be padded by blank lines_

src/components/table/template.test.js
```diff
@@ -0,0 +1,374 @@
...
+    })
+    const $component = $('.govuk-c-table')
+    const $tableBody = $component.find('.govuk-c-table .govuk-c-table__body')
+    const $firstTableRow = $tableBody.find('.govuk-c-table__row:first-child')
+    const $firstTableHeader = $firstTableRow.find('.govuk-c-table__header:first-child')
+
+    expect($firstTableHeader.get(0).tagName).toEqual('th')
+    expect($firstTableHeader.attr('scope')).toEqual('row')
+    expect($firstTableHeader.hasClass('govuk-c-table__header')).toBeTruthy()
+
```
alex-ju: _More than one blank line not allowed_

src/components/table/template.test.js
```diff
@@ -0,0 +1,374 @@
...
+
+    const $lastTableRow = $tableBody.find('.govuk-c-table__row:last-child')
+    const $lastTableHeader = $lastTableRow.find('.govuk-c-table__header:first-child')
+
+    expect($lastTableHeader.get(0).tagName).toEqual('th')
+    expect($lastTableHeader.attr('scope')).toEqual('row')
+    expect($lastTableHeader.hasClass('govuk-c-table__header')).toBeTruthy()
+
```
alex-ju: _Block must not be padded by blank lines_

src/components/table/template.test.js
igloosi: fixed

src/components/table/template.test.js
igloosi: fixed

PR #473, opened 2018-01-29
Add tests for breadcrumbs macro

alex-ju:
This PR:
- Adds tests to breadcrumbs component to ensure the markup is rendered correctly when providing arguments to the macro.

*Note to reviewers*: if you feel like any example should be removed, let me know.

[Trello Card](https://trello.com/c/V9ZbJRRZ)

### DIFF
--- a/src/components/breadcrumbs/template.test.js
+++ b/src/components/breadcrumbs/template.test.js
@@ -1,0 +1,125 @@
+/* globals describe, it, expect */
+
+const { render, getExamples } = require('../../../lib/jest-helpers')
+
+const examples = getExamples('breadcrumbs')
+
+describe('Breadcrumbs', () => {
+  describe('by default', () => {
+    it('renders with classes', () => {
+      const { $ } = render('breadcrumbs', {
+        classes: 'app-c-breadcrumbs--custom-modifier'
+      })
+
+      const $component = $('.govuk-c-breadcrumbs')
+      expect($component.hasClass('app-c-breadcrumbs--custom-modifier')).toBeTruthy()
+    })
+
+    it('renders with attributes', () => {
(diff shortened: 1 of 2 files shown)


src/components/breadcrumbs/template.test.js
```diff
@@ -0,0 +1,112 @@
...
+          'role': 'navigation'
+        }
+      })
+
+      const $component = $('.govuk-c-breadcrumbs')
+      expect($component.attr('id')).toEqual('my-navigation')
+      expect($component.attr('role')).toEqual('navigation')
+    })
+
+    it('renders with items', () => {
```
nickcolley: Looks like items are covered below so we may not need this additional test.

src/components/breadcrumbs/template.test.js
```diff
@@ -0,0 +1,112 @@
...
+            'text': 'Sub-section'
+          }
+        ]
+      })
+
+      const $items = $('.govuk-c-breadcrumbs__list-item')
+      expect($items.length).toEqual(2)
+    })
+
+    it('renders item with text', () => {
```
nickcolley: Can we include a test that tries to put `html` in `text` that asserts it is escaped?

src/components/breadcrumbs/template.test.js
alex-ju: It kinda' is, but if `renders item with text` is failing is not explicit why. may have an item without content or you may not have the item at all. that was the logic behind it. happy to remove it if others feel the same about it being redundant.

src/components/breadcrumbs/template.test.js
alex-ju: Added `renders item with escaped entities in text`

PR #474, opened 2018-01-29
Add test to check readme generation

nickcolley:
This is a bit complicated, not sure if there's something better we could do here.

Maybe we should just remove dynamic building of READMEs and do something else?

Currently fails because Travis is being run in v7 and `util.promisify` requires 8.x see https://github.com/alphagov/govuk-frontend/pull/462

This should fail because of the lack of checked in READMEs once the above is addressed.

Note: Currently passes after the second run, since it is not aware of git changes.

### DIFF
--- a/app/__tests__/app.test.js
+++ b/app/__tests__/app.test.js
@@ -3,8 +3,8 @@
 const request = require('request')
 const cheerio = require('cheerio')
 
-const app = require('../app/app.js')
-const lib = require('../lib/file-helper')
+const app = require('../app.js')
+const lib = require('../../lib/file-helper')
 
 const requestParams = {
   url: 'http://localhost:3000/',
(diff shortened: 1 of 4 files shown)


nickcolley:
Solved better here: https://github.com/alphagov/govuk-frontend/pull/482

nickcolley:
Actually not sure @igloosi ? Doesnt seem to remove the HTML examples from the READMEs, is this PR useful?

nickcolley:
Updated this PR, you can see the build failing before I committed the generated README.

36degrees:
Is there any way to make the test output clearer and less verbose? Ideally it'd just report which files are dirty?

nickcolley:
Sure thing I'll try that.

nickcolley:
Updated so that if it errors we'll see only the component name.

screen shot 2018-02-02 at 14 42 30

36degrees:
That's ace, thanks!

I do wonder about the name and location of the test, because this isn't _really_ testing the `generate:readme` task - it's more of a sanity check for us?

Also, worth adding a card to set this up as a precommit / prepush action?

alex-ju:
Tested, works fine 👍 . I would probably asked if possible to put the logic (file operations) in functions in the file helper so we'll have clean tests, but I we need this tests sooner rather than later.

nickcolley:
Hopefully this test will be temporary once we move readme generation to only happen as part of releases.

Can I get an approve please?

PR #476, opened 2018-01-29
Add tests for `govuk-c-radios` component

nickcolley:
- includes new example disabled radios
- includes new example with all fieldset params set

Uses snapshot testing to test the dependant components - Label and Fieldset.
This ensures that all values that can be passed through the Radio components params through to
either component is tested, and if we change the behavour in either the Label or Fieldset component
we will see a change to confirm in this component.

Snapshot testing makes use of a [custom serializer](https://github.com/rayrutjes/jest-serializer-html) to make the diffs easy to read

As part of https://trello.com/c/G8rFbXWB/641-automated-tests-for-radios-component

### DIFF
--- a/lib/jest-helpers.js
+++ b/lib/jest-helpers.js
@@ -12,12 +12,24 @@ nunjucks.configure(configPaths.components, {
   lstripBlocks: true
 })
 
+/**
+ * Render a component's template for testing
+ * @param {string} componentName
+ * @param {string} params parameters that are used in the component template
+ * @returns {object} returns object that includes raw HTML (output) and
+ * also a cheerio (jQuery) instance of the template for easy DOM querying
+ */
 function render (componentName, params) {
   const output = nunjucks.render(componentName + '/template.njk', { params })
   const $ = cheerio.load(output)
   return { output, $ }
 }
 
+/**
+ * Get examples from a component's metadata file
+ * @param {string} componentName
+ * @returns {object} returns object that includes all examples at once
+ */
 function getExamples (componentName) {
   const file = fs.readFileSync(
     path.join(configPaths.components, componentName, `${componentName}.yaml`),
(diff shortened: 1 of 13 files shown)


src/components/radios/template.test.js
```diff
@@ -0,0 +1,256 @@
...
+  })
+
+  it('render extreme fieldset', () => {
+    const { $ } = render('radios', examples['with-extreme-fieldset'])
+
+    const $component = $('.govuk-c-radios')
+    expect($component.find('.govuk-c-error-message').text()).toContain('Please select an option')
+
+    const $fieldset = $('.govuk-c-fieldset')
+    expect($fieldset.hasClass('app-c-fieldset--custom-modifier')).toBeTruthy()
```
nickcolley: Lots of assertions here since we just want to know that the fieldset plays nice inside this component. Feels over the top to do individual test blocks (which should be done in the fieldset component tests), maybe a snapshot test would be better here?

nickcolley: See the same for label and error message.

src/components/radios/template.test.js
```diff
@@ -0,0 +1,261 @@
...
+        }
+      ]
+    })
+
+    const $component = $('.govuk-c-radios')
+
+    const $firstInput = $component.find('.govuk-c-radios__item:first-child input')
+    const $firstLabel = $component.find('.govuk-c-radios__item:first-child label')
+    expect($firstInput.attr('name')).toEqual('example-name')
+    expect($firstLabel.text()).toEqual('example-name')
```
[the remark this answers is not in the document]
nickcolley: Might be able to use `toContain`, will update

nickcolley:
Just setting this to WIP, need to update based on Alex's feedback

nickcolley:
Updated, I've added a new jest helper method that will allow us to only snapshot parts of the HTML we care about.

Would appreciate your thoughts, does it make sense?

nickcolley:
Talked to Alex, going to add some comments to document the jest-helpers

lib/jest-helpers.js
```diff
@@ -12,12 +12,24 @@ nunjucks.configure(configPaths.components, {
...
 
+/**
+ * Render a component's template for testing
+ * @param {string} componentName
+ * @param {string} params parameters that are used in the component template
+ * @returns {object} returns object that includes raw HTML (output) and
+ * also a cheerio (jQuery) instance of the template for easy DOM querying
+ */
 function render (componentName, params) {
   const output = nunjucks.render(componentName + '/template.njk', { params })
```
nickcolley: Wonder if we need this output anymore since it was originally used for snapshot testing...

nickcolley:
Will write up this approach for the docs

alex-ju:
@nickcolley, @hannalaakso also wants to update the docs on how to write tests, would be good to catch up or split the work

hannalaakso:
Sounds good to me. @nickcolley go ahead and write up your approach and I can contribute to the docs 🙂

PR #478, opened 2018-01-29
Add tests for input macro

alex-ju:
This PR:
- Adds tests to input component to ensure the markup is rendered correctly when providing arguments to the macro.

*Note to reviewers*: if you feel like any example should be removed, let me know.

[Trello Card](https://trello.com/c/brDoqQE0)

### DIFF
--- a/src/components/input/template.test.js
+++ b/src/components/input/template.test.js
@@ -1,0 +1,107 @@
+/* globals describe, it, expect */
+
+const { render, htmlWithClassName } = require('../../../lib/jest-helpers')
+
+describe('Input', () => {
+  describe('by default', () => {
+    it('renders with classes', () => {
+      const $ = render('input', {
+        classes: 'app-c-input--custom-modifier'
+      })
+
+      const $component = $('.govuk-c-input')
+      expect($component.hasClass('app-c-input--custom-modifier')).toBeTruthy()
+    })
+
+    it('renders with id', () => {
+      const $ = render('input', {
+        id: 'my-input'
+      })
+
+      const $component = $('.govuk-c-input')
+      expect($component.attr('id')).toEqual('my-input')
+    })
+
+    it('renders with name', () => {
+      const $ = render('input', {
+        name: 'my-input-name'
+      })
+
+      const $component = $('.govuk-c-input')
+      expect($component.attr('name')).toEqual('my-input-name')
+    })
+
+    it('renders with type="text"', () => {
+      const $ = render('input')
+
+      const $component = $('.govuk-c-input')
+      expect($component.attr('type')).toEqual('text')
+    })
+
+    it('renders with value', () => {
+      const $ = render('input', {
+        value: 'QQ 12 34 56 C'
+      })
+
+      const $component = $('.govuk-c-input')
+      expect($component.val()).toEqual('QQ 12 34 56 C')
+    })
+
+    it('renders with attributes', () => {
+      const $ = render('input', {
+        attributes: {
+          'data-attribute': 'my data value'
+        }
+      })
+
+      const $component = $('.govuk-c-input')
+      expect($component.attr('data-attribute')).toEqual('my data value')
+    })
+  })
+
+  describe('with dependant components', () => {
+    it('renders with label', () => {
+      const $ = render('input', {
+        id: 'my-input',
+        label: {
+          'text': 'National Insurance number'
+        }
+      })
+
+      expect(htmlWithClassName($, '.govuk-c-label')).toMatchSnapshot()
+    })
+
+    it('renders label with "for" attribute reffering the input "id"', () => {
+      const $ = render('input', {
+        id: 'my-input',
+        label: {
+          'text': 'National Insurance number'
+        }
+      })
+
+      const $label = $('.govuk-c-label')
+      expect($label.attr('for')).toEqual('my-input')
+    })
+
+    it('renders with error message', () => {
+      const $ = render('input', {
+        errorMessage: {
+          'text': 'Error message'
+        }
+      })
+
+      expect(htmlWithClassName($, '.govuk-c-error-message')).toMatchSnapshot()
+    })
+
+    it('has error class when rendered with error message', () => {
+      const $ = render('input', {
+        errorMessage: {
+          'text': 'Error message'
+        }
+      })
+
+      const $component = $('.govuk-c-input')
+      expect($component.hasClass('govuk-c-input--error')).toBeTruthy()
+    })
+  })
+})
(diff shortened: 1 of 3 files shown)


src/components/input/template.test.js
```diff
@@ -0,0 +1,120 @@
...
+      expect($component.hasClass('govuk-c-input--error')).toBeTruthy()
+      expect($errorMessage).toBeTruthy()
+    })
+  })
+
+  describe('default example', () => {
+    it('renders with label', () => {
+      const { $ } = render('input', examples.default)
+      const $label = $('.govuk-c-label')
+      expect($label.text()).toContain('National Insurance number')
```
36degrees: Should we have a test that the label is associated by `id` / `for`?

36degrees: Oh, we do, but it's further up. What's the thinking being splitting these up into a separate context?

src/components/input/template.test.js
```diff
@@ -0,0 +1,120 @@
...
+      })
+
+      const $component = $('.govuk-c-input')
+      expect($component.hasClass('app-c-input--custom-modifier')).toBeTruthy()
+    })
+
+    it('renders with id', () => {
+      const { $ } = render('input', {
+        id: 'my-input',
+        name: 'my-input-name'
```
36degrees: We don't test for name, and given that in the example above we omit both id and name, do we want to do the same here and just provide an `id`? Conversely, if we're providing `name` and `id` in the other examples because they are 'required', should we include them in the classes example also?

alex-ju: Indeed I started to add `id` and `name` because they are 'required', but I'm not sure it's the right approach and right now is inconsistent across examples. I would rather add the minimum number of attributes since we cannot really enforce the 'required' state.

src/components/input/template.test.js
alex-ju: From line 98 I tested the examples that we have defined in the `yaml` so somehow make sure we're not breaking them. Before that line are 'by default' tests that cover the arguments, logic and relationship.

src/components/input/template.test.js
alex-ju: Removed unused attributes for now, even if defined as 'required'.

src/components/input/template.test.js
```diff
@@ -0,0 +1,113 @@
...
+          'text': 'National Insurance number'
+        }
+      })
+
+      const $label = $('.govuk-c-label')
+      expect($label).toBeTruthy()
+      expect($label.attr('for')).toEqual('my-input')
+    })
+
+    it('renders with error message', () => {
```
36degrees: Would it make sense to split this test into two parts?

1) it applies the error variant class if an error message is passed
2) it renders an error message if passed

src/components/input/template.test.js
```diff
@@ -0,0 +1,113 @@
...
+        }
+      })
+
+      const $component = $('.govuk-c-input')
+      expect($component.attr('aria-describedby')).toEqual('my-label')
+    })
+
+    it('renders with label', () => {
+      const { $ } = render('input', {
+        id: 'my-input',
```
36degrees: This is redundant

src/components/input/template.test.js
```diff
@@ -0,0 +1,113 @@
...
+      })
+
+      const $component = $('.govuk-c-input')
+      expect($component.val()).toEqual('QQ 12 34 56 C')
+    })
+
+    it('renders with attributes', () => {
+      const { $ } = render('input', {
+        attributes: {
+          'aria-describedby': 'my-label'
```
36degrees: It might be worth using a different 'non-sensical' data attribute, only to avoid confusion for people scanning through who might mis-read this as inputs actually having or requiring some aria markup?

src/components/input/template.test.js
```diff
@@ -0,0 +1,113 @@
...
+      const { $ } = render('input', {
+        id: 'my-input',
+        label: {
+          'text': 'National Insurance number'
+        }
+      })
+
+      const $label = $('.govuk-c-label')
+      expect($label).toBeTruthy()
+      expect($label.attr('for')).toEqual('my-input')
```
36degrees: Again, suggest this is split into two parts - one that tests that the label is rendered correctly (probably worth checking the text itself rather than just ensuring the element exists?) and a second test that ensures the label is correctly associated with the input?

src/components/input/template.test.js
```diff
@@ -0,0 +1,113 @@
...
+      const $label = $('.govuk-c-label')
+      expect($label.text().trim()).toEqual('National Insurance number')
+    })
+  })
+
+  describe('with-hint-text example', () => {
+    it('renders with hint text', () => {
+      const { $ } = render('input', examples['with-hint-text'])
+      const $hintText = $('.govuk-c-label__hint')
+      expect($hintText.text().trim()).toEqual('It’s on your National Insurance card, benefit letter, payslip or P60. For example, ‘QQ 12 34 56 C’.')
```
36degrees: Would be good to wrap this within 80 chars.

src/components/input/template.test.js
alex-ju: I expect the label to be tested within the label macro. We don't add any logic about the label in the input macro, except for setting label's `for` attribute equal to the input's `id`.

36degrees: OK. I still think it's worth having an explicit test for that for/id relationship.

src/components/input/template.test.js
alex-ju: replaced `aria-describedby` with 'non-sensical' data attribute

src/components/input/template.test.js
alex-ju: removed the one that was testing the example and tested label too

src/components/input/template.test.js
alex-ju: split

src/components/input/template.test.js
alex-ju: split

src/components/input/template.test.js
```diff
@@ -0,0 +1,107 @@
...
+
+    it('renders with error message', () => {
+      const $ = render('input', {
+        errorMessage: {
+          'text': 'Error message'
+        }
+      })
+
+      const $errorMessage = $('.govuk-c-error-message')
+      expect($errorMessage).toBeTruthy()
```
nickcolley: Could we replace these truthy tests with snapshots?

alex-ju:
Updated with snapshots as discussed.

src/components/input/template.test.js
```diff
@@ -0,0 +1,105 @@
...
+
+    it('renders with label', () => {
+      const $ = render('input', {
+        id: 'my-input',
+        label: {
+          'text': 'National Insurance number'
+        }
+      })
+
+      expect(htmlWithClassName($, '.govuk-c-label')).toMatchSnapshot()
```
nickcolley: Could we wrap these snapshots tests with a description 'dependant components' or similar?

PR #479, opened 2018-01-29
Add tests for `govuk-c-warning-text` component

hannalaakso:
This PR adds Jest tests for the warning text component.

They test if the component:

- renders the default example with text and assistive text
- renders the default example with aria attribute for icon
- renders classes
- renders custom text and custom assistive text
- renders escaped html when passed to text
- renders html
- renders attributes

### DIFF
--- a/src/components/warning-text/template.test.js
+++ b/src/components/warning-text/template.test.js
@@ -1,0 +1,85 @@
+/* globals describe, it, expect */
+
+const { render, getExamples } = require('../../../lib/jest-helpers')
+
+const examples = getExamples('warning-text')
+
+describe('Warning text', () => {
+  it('renders the default example with text', () => {
+    const { $ } = render('warning-text', examples.default)
+
+    const $component = $('.govuk-c-warning-text')
+    expect($component.text()).toContain('You can be fined up to £5,000 if you don’t register.')
+  })
+
+  it('renders the default example with assistive text', () => {
+    const { $ } = render('warning-text', examples.default)
+
+    const $assistiveText = $('.govuk-c-warning-text__assistive')
+    expect($assistiveText.text()).toEqual('Warning')
+  })
+
+  it('hides the icon from screen readers using the aria-hidden attribute', () => {
+    const { $ } = render('warning-text', examples.default)
+
+    const $icon = $('.govuk-c-warning-text__icon')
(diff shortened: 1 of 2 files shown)


src/components/warning-text/template.test.js
```diff
@@ -0,0 +1,78 @@
...
+    const { $ } = render('warning-text', examples.default)
+
+    const $component = $('.govuk-c-warning-text')
+    expect($component.text()).toContain('You can be fined up to £5,000 if you don’t register.')
+
+    const $assistiveText = $('.govuk-c-warning-text__assistive')
+    expect($assistiveText.text()).toEqual('Warning')
+  })
+
+  it('renders the default example with aria attribute for icon', () => {
```
36degrees: We could be clearer with our intent here. How about something like…

> It hides the icon from screen readers using the aria-hidden attribute

?

src/components/warning-text/template.test.js
```diff
@@ -0,0 +1,78 @@
...
+    const { $ } = render('warning-text', {
+      classes: 'govuk-c-warning-text--custom-class',
+      text: 'Warning text'
+    })
+
+    const $component = $('.govuk-c-warning-text')
+    expect($component.hasClass('govuk-c-warning-text--custom-class')).toBeTruthy()
+  })
+
+  it('renders custom text and custom assistive text', () => {
```
36degrees: I think this would make more sense as two different tests.

src/components/warning-text/template.test.js
```diff
@@ -0,0 +1,78 @@
...
+    expect($component.html()).toContain('<span>Some custom warning text</span>')
+  })
+
+  it('renders attributes', () => {
+    const { $ } = render('warning-text', {
+      attributes: {
+        'data-test': 'attribute',
+        'id': 'my-warning-text'
+      },
+      html: '<span>Some custom warning text</span>'
```
36degrees: This isn't being tested here. Can it just be omitted?

src/components/warning-text/template.test.js
hannalaakso: A very good call 💯 Have amended.

src/components/warning-text/template.test.js
hannalaakso: I was in two minds about that but I think you're right. Have split it into two tests.

PR #480, opened 2018-01-29
Add tests for the details component

36degrees:
https://trello.com/c/lJzDNEvk/642-automated-tests-for-details-component

### DIFF
--- a/src/components/details/template.test.js
+++ b/src/components/details/template.test.js
@@ -1,0 +1,80 @@
+/* globals describe, it, expect */
+
+const { render, getExamples } = require('../../../lib/jest-helpers')
+
+const examples = getExamples('details')
+
+describe('Details', () => {
+  it('renders a details element', () => {
+    const { $ } = render('details', examples.default)
+
+    const $component = $('.govuk-c-details')
+    expect($component.get(0).tagName).toEqual('details')
+  })
+
+  it('includes a nested summary', () => {
+    const { $ } = render('details', examples.default)
+
+    // Look for the summary element _within_ the details element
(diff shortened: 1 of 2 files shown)


src/components/details/template.test.js
```diff
@@ -0,0 +1,80 @@
...
+
+    const $component = $('.govuk-c-details')
+    expect($component.get(0).tagName).toEqual('details')
+  })
+
+  it('includes a nested summary', () => {
+    const { $ } = render('details', examples.default)
+
+    // Look for the summary element within the details element
+    const $summary = $('.govuk-c-details__summary', '.govuk-c-details')
```
nickcolley: Didn't even know you could do this, I've been doing `$('...').find('...')` all this time.

36degrees: Yeah, I think it's called 'context' in jQuery-land. I've seen in the past people generally aren't that aware of it – hence the comment. Do you think it's clear enough what's it doing?

nickcolley: In this specific case I'd just rely on good ol' CSS selectors `$('.govuk-c-details > .govuk-c-details__summary')` which would hopefully be more widely understood.

36degrees: That's a good call, though not sure we'd want to use `>` as that would be testing that the summary is a direct descendant - which I don't think we care about?

nickcolley: I stealth removed the direct descendant, but wasn't quick enough 😢

36degrees: 😁

PR #482, opened 2018-01-30
Remove README content from review app

igloosi:
This PR
- removes the component description, guidance and table of arguments from
the component view within the review app. This just leaves the component title and the examples.

- creates a separate template in app/views for generating README files
- adds logic to each component's `index.njk` file to extend the appropriate templates based on the `isReadme: true` parameter that is we pass to  the nunjucks enviroment.

Trello: https://trello.com/c/rUJHVyt3/612-remove-readme-content-from-components-view-in-review-app

Note: this is until we move all content into YAML file and remove component `index.njk` files

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -111,6 +111,7 @@ Internal:
   (PR [#427](https://github.com/alphagov/govuk-frontend/pull/427))
 - Improve documentation around publishing (PR [#430](https://github.com/alphagov/govuk-frontend/pull/430))
 - Improve documentation around contributing (PR [#433](https://github.com/alphagov/govuk-frontend/pull/433))
+- Remove readme content from review app (PR [#482](https://github.com/alphagov/govuk-frontend/pull/482))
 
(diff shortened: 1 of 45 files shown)


36degrees:
Would it make more sense at this point to split the template for the README and the template for the component view entirely?

Given the only thing that's common between the two is the title and the examples which are already wrapped up in a macro.

igloosi:
can do.

alex-ju:
Discussed with @igloosi and @nickcolley. It's not ideal but i think looks good for now. It feels odd that we're passing data both downstream and upstream, which makes things confusing, especially for newcomers. Part of refectory would be to pass data only downstream, which will simply these views to be inherited from one layout without the need to check for `isReadme`.

igloosi:
shall we wait for the today's session outcome before doing anything with this?

nickcolley:
This seems like a good first step to me

igloosi:
will need to be reviewed after #528

igloosi:
rebased with latest changes

PR #483, opened 2018-01-30
Add govuk-section-break to core

dashouse:
Sometimes there is a need to visually separate parts of a page on top of the existing typographic hierarchy.

This class is added to an `<hr>`, adds margin (xl, l and m).
There is also an option to make the `<hr>` border visible or invisible. 

##Examples
### Not visible
screen shot 2018-01-30 at 11 53 16

###Visible
screen shot 2018-01-30 at 11 56 28

## Example usage
The section below the Start button on a start page is achieved by adding margin to the Start button.

This conceptually feels like a new section of the page.

screen shot 2018-01-30 at 12 01 38

### Related
How this might work with `<section>`

As part of https://trello.com/c/8LRu4psv/515-add-styling-for-horizontal-rules

### DIFF
--- a/src/globals/scss/core/_prose-scope.scss
+++ b/src/globals/scss/core/_prose-scope.scss
@@ -48,5 +48,10 @@
     a {
       @extend .govuk-link;
     }
+
+    hr {
+      @extend .govuk-section-break__visible;
+      @extend .govuk-section-break--xl;
+    }
   }
 }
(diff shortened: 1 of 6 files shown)


src/globals/scss/core/_prose-scope.scss
```diff
@@ -48,5 +48,10 @@
     a {
       @extend .govuk-link;
     }
+
+    hr {
```
nickcolley: Should this use govuk-section-break__visible ?

nickcolley: Can we also add an example to the prose page to show this in use?

dashouse: Updated

alex-ju:
After update, this looks technically good to me. Happy to approve if no one else wants to raise any questions.

hannalaakso:
This looks good to me too. Have just updated Changelog. 

@alex-ju if you're happy with the changelog could you approve 🙂

PR #487, opened 2018-01-30
Rename app.js to index.js for more idomatic entrypoint naming

nickcolley:
Generally people tend to use index.js or entry.js


joelanman:
app.js is consistent with Express
https://expressjs.com/en/starter/generator.html

36degrees:
Yeah, index to me suggests the file relates to a specific `route` – the 'index' page `/`. I think the current name better reflects what the file actually does at the minute?

nickcolley:
Feels off to me but no worries

PR #490, opened 2018-02-01
Update generated READMEs for components

nickcolley:
This PR commits accidentally forgotten about README updates.

I raised a PR to test for this so this doesn't happen again here https://github.com/alphagov/govuk-frontend/pull/474


igloosi:
commit in #482 addresses this as well

PR #491, opened 2018-02-01
Define size of inputs etc in pixels rather than em

hannalaakso:
This PR removes the calls to the `em` function from Frontend, so that all measurements
are defined in pixels. 

The `em` conversions were originally added in an attempt
to make form elements line up. However, they cause problems when you try to use
Frontend in an application that defines font sizes, especially when those font
sizes are nested.

This also removes the em() call from `govuk-c-details` as the above problems
apply to this component too. For now, `govuk-c-table` still calls `em` - I'll
find out from Anika if there's any particular reason tables in Elements used
them.

Note: I will review the visuals with @dashouse, there might be things we need to adjust now.

As part of https://trello.com/c/WpPKyqxl/493-5-define-size-of-inputs-etc-in-pixels-rather-than-em

### DIFF
--- a/src/components/checkboxes/_checkboxes.scss
+++ b/src/components/checkboxes/_checkboxes.scss
@@ -8,7 +8,7 @@
     position: relative;
 
     margin-bottom: $govuk-spacing-scale-2;
-    padding: 0 0 0 em(40px, 19px);
+    padding: 0 0 0 40px;
 
     clear: left;
   }
@@ -39,7 +39,7 @@
 
   .govuk-c-checkboxes__label {
     display: block;
-    padding: em(8px, 19px) em($govuk-spacing-scale-3, 19px) em($govuk-spacing-scale-1, 19px);
+    padding: 8px $govuk-spacing-scale-3 $govuk-spacing-scale-1;
     cursor: pointer;
     // remove 300ms pause on mobile
     -ms-touch-action: manipulation;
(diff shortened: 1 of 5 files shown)


nickcolley:
How do we want these components to look on small screens?

I've done a manual visual diff before and after

before-and-after

hannalaakso:
Thanks @igloosi that's now fixed.

hannalaakso:
Thanks @nickcolley 👍  I've requested @dashouse to review the visuals, there might be things we want to adjust. I just added that to the description for clarity.

nickcolley:
before-and-after

dashouse:
Neither of @nickcolley's GIFs look like what I'm seeing? I see them as perfectly lined up where as your screenshot still has a slight overlap?

Intention with these was to retain the physical size of the input at all breakpoints, even though the font-size gets smaller within.

nickcolley:
@dashouse I'm on Windows at the moment, so might be fine on OSX

hannalaakso:
@nickcolley What Windows OX/browser/browser version are you using?

nickcolley:
@hannalaakso do you want me to raise an issue so this isnt blocked?

hannalaakso:
Thanks @nickcolley that would be good. I'm not able to replicate this on Browserstack but happy for it to go in the backlog and I can pick it up and have another look in the accessibility lab when I'm in the office.

36degrees:
@hannalaakso Can you make sure we add something to the changelog for this? 🙂

hannalaakso:
Done: https://github.com/alphagov/govuk-frontend/pull/520 Thanks for the heads up @36degrees!

PR #494, opened 2018-02-02
WIP - Do not merge - Add error highlighting

hannalaakso:
This PR adds examples of error messages for form components. **It's not a complete proposed solution** but rather something for @dashouse to review before he's off for holibobs. I'm not working this afternoon so happy for @nickcolley to pick this up.

There are several alignment issues and as @dashouse said before, we might need to move the margins from inside the inputs to the `form-group` wrapper. It would be awesome @dashouse if you could decide our approach to those 🙌  

An issue I ran into was how to set an error state, not the actual error message, in components. I've tried an approach of resolving that here but this might well need a wider discussion with the team.

Features:
- Update fieldset to accept error state and render an error class
- Add error state to `input`, `textarea`, `select` and allow them to pass error state to `label`.
Also add outer error class wrapper (`form-group`) to add error style. Add error styling to form element error if error state is true, rather than checking for existence of error message.
if error message is set.
- Add error state to `radios`, `checkboxes` and `date-input` and allow them to pass to this to `fieldset`.
- Add the error styling with two different classes `fieldset` for those components that already use it (for `radios`, `checkboxes` and `date-input`) and `form-group` for those that didn't have a wrapper (`input`, `textarea`, `select`). It might be better to make the `form-group` class component specific but then again it might be good to abstract it out into something that the user can apply themselves to custom form errors. Up for discussion although I'm now actually leaning towards the former (the option I *didn't* implement 😛 ) because there are now are two places where error classes get added in `input`, `textarea`, `select` and a single wrapper class could manage them both.
- Add examples of the different error messages for form components to `/examples/error-messages`

This doesn't update all component previews, or any READMEs and component argument tables as we need to settle on an approach first before updating all those.

Note: Please review in [`/examples/error-messages`](https://govuk-frontend-review-pr-494.herokuapp.com/examples/error-messages). It might be easier to look at the commits which separate the work out into smaller chunks.

### DIFF
--- a/src/components/checkboxes/template.njk
+++ b/src/components/checkboxes/template.njk
@@ -31,6 +31,10 @@
     legendHtml: params.fieldset.legendHtml,
     legendHintText: params.fieldset.legendHintText,
     legendHintHtml: params.fieldset.legendHintHtml,
+    error: {
+      active: params.fieldset.error.active,
+      text: params.fieldset.error.text
+    },
     errorMessage: params.errorMessage
   }) %}
   {{ innerHtml | trim | safe }}
--- a/src/components/fieldset/_fieldset.scss
+++ b/src/components/fieldset/_fieldset.scss
@@ -11,6 +11,11 @@
     @include govuk-h-clearfix;
   }
 
+  .govuk-c-fieldset--error {
+    padding-left: 15px;
+    border-left: 5px solid $govuk-red;
+  }
+
   .govuk-c-fieldset__legend {
     @include govuk-font-regular-19;
 
--- a/src/components/fieldset/template.njk
+++ b/src/components/fieldset/template.njk
@@ -1,6 +1,6 @@
 {% from "error-message/macro.njk" import govukErrorMessage -%}
 
-<fieldset class="govuk-c-fieldset
+<fieldset class="govuk-c-fieldset {%- if params.error.active %} govuk-c-fieldset--error{% endif %}
   {%- if params.classes %} {{ params.classes }}{% endif %}"{% for attribute, value in params.attributes %} {{ attribute }}="{{ value }}"{% endfor %}>
   {% if params.legendHtml or params.legendText %}
   <legend class="govuk-c-fieldset__legend">
@@ -8,8 +8,8 @@
     {% if params.legendHintText or params.legendHintHtml %}
     <span class="govuk-c-fieldset__hint">{{ params.legendHintHtml | safe if params.legendHintHtml else params.legendHintText }}</span>
     {% endif %}
-    {% if params.errorMessage %}
-    {{ govukErrorMessage(params.errorMessage) | trim | indent(4) -}}
+    {% if params.error.active %}
+    {{ govukErrorMessage(params.error) | trim | indent(4) -}}
     {% endif %}
   </legend>
   {% endif %}
--- a/src/components/label/template.njk
+++ b/src/components/label/template.njk
@@ -9,7 +9,7 @@
     {{ params.hintHtml | safe if params.hintHtml else params.hintText }}
   </span>
   {% endif %}
-  {% if params.errorMessage %}
-  {{ govukErrorMessage(params.errorMessage) -}}
+  {% if params.error.active %}
+  {{ govukErrorMessage(params.error.message) -}}
   {% endif %}
 </label>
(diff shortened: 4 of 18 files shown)


dashouse:
The error should really be the full height of the fieldset, including the legend and the hint
screen shot 2018-02-02 at 12 56 39

I understand why the input looks like this...
screen shot 2018-02-02 at 13 00 41

BUT is there any way when `govuk-form-group--error` is applied that the `govuk-c-input` has it's margin-bottom removed? and that margin added to `govuk-form-group--error` container?
screen shot 2018-02-02 at 13 00 30

src/components/fieldset/template.njk
```diff
@@ -1,15 +1,15 @@
...
   {%- if params.classes %} {{ params.classes }}{% endif %}"{% for attribute, value in params.attributes %} {{ attribute }}="{{ value }}"{% endfor %}>
   {% if params.legendHtml or params.legendText %}
   <legend class="govuk-c-fieldset__legend">
     {{ params.legendHtml | safe if params.legendHtml else params.legendText }}
     {% if params.legendHintText or params.legendHintHtml %}
     <span class="govuk-c-fieldset__hint">{{ params.legendHintHtml | safe if params.legendHintHtml else params.legendHintText }}</span>
     {% endif %}
-    {% if params.errorMessage %}
-    {{ govukErrorMessage(params.errorMessage) | trim | indent(4) -}}
+    {% if params.error.active %}
```
alex-ju: This will prevent passing an `errorMessage` directly, without specifying an additional `error.active`. I think we should be flexible in this matter and allow an `errorMessage` without additional params.

src/components/checkboxes/template.njk
```diff
@@ -31,6 +31,10 @@
     legendHtml: params.fieldset.legendHtml,
     legendHintText: params.fieldset.legendHintText,
     legendHintHtml: params.fieldset.legendHintHtml,
+    error: {
```
36degrees: I am not sure about this change… previously errorMessage mapped directly to a component, consistent with the way we handle dependent components elsewhere. I don't think it's clear that you can pass arguments consistent with the errorMessage.

I'd suggest either:

- automatically adding the error classes if an error message is passed, or
- making error a single boolean (`error: true`) and retaining the existing `errorMessage`.

src/components/checkboxes/template.njk
```diff
@@ -31,6 +31,10 @@
...
+      active: params.fieldset.error.active,
+      text: params.fieldset.error.text
+    },
     errorMessage: params.errorMessage
```
36degrees: This is no longer being used in the fieldset component - if we decide that moving the text / html arguments to error _is_ the right approach, this should be removed.

src/globals/scss/core/_typography.scss
```diff
@@ -157,4 +157,10 @@
     }
   }
 
+  // Form group error style
+  .govuk-form-group--error {
```
36degrees: This is not related to typography in any way. I am not sure where it should live (sorry!) but it's definitely not here.

src/components/label/template.njk
```diff
@@ -9,7 +9,7 @@
     {{ params.hintHtml | safe if params.hintHtml else params.hintText }}
   </span>
   {% endif %}
-  {% if params.errorMessage %}
-  {{ govukErrorMessage(params.errorMessage) -}}
+  {% if params.error.active %}
+  {{ govukErrorMessage(params.error.message) -}}
```
36degrees: This is inconsistent with the way the errorMessage attributes are passed in the fieldset - they're not nested there.

src/components/fieldset/_fieldset.scss
```diff
@@ -11,6 +11,11 @@
     @include govuk-h-clearfix;
   }
 
+  .govuk-c-fieldset--error {
+    padding-left: 15px;
```
36degrees: Spacing scale?

src/components/fieldset/_fieldset.scss
```diff
@@ -11,6 +11,11 @@
...
+    border-left: 5px solid $govuk-red;
```
36degrees: `$govuk-border-width`?

hannalaakso:
Closing PR for posterity and trying a different approach in https://github.com/alphagov/govuk-frontend/pull/519

PR #497, opened 2018-02-05
Add tests for textarea macro

alex-ju:
This PR:
- Adds tests to textarea component to ensure the markup is rendered correctly when providing arguments to the macro;

[Trello Card](https://trello.com/c/OMxt1vUN)


igloosi:
how does the snapshot get generated?
all other assertions seem fine.

alex-ju:
@igloosi: when we have a snapshot assertion (e.g. `toMatchSnapshot()`), the first time this test is run, Jest creates a snapshot file that looks like the one in this PR. when continuing to work on the project, every time we run this test suite, Jest will do the process over again and compare the old snapshot to a new one. when we actually want to update the snapshot we will use `-u` flag (or `jest --updateSnapshot`).

PR #506, opened 2018-02-06
Add tests for select macro

alex-ju:
This PR:
- Adds tests to select component to ensure the markup is rendered correctly when providing arguments to the macro;

[Trello Card](https://trello.com/c/3R8USRGF)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,8 @@ Fixes:
 - Remove regular font weight from link styles (PR [#469](https://github.com/alphagov/govuk-frontend/pull/469))
 - Remove redundant 'govuk-c-border' div from the details component
   (PR [#481](https://github.com/alphagov/govuk-frontend/pull/481))
-
+- Add `govuk-c-select--error` modifier class to the select component instead of relying on `govuk-c-input--error` (PR [#506](https://github.com/alphagov/govuk-frontend/pull/506))
+- Allow error message and hint text to be passed to a select component without requiring a label parameter (PR [#506](https://github.com/alphagov/govuk-frontend/pull/506))
 
 Internal:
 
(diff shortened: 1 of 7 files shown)


src/components/select/__snapshots__/template.test.js.snap
```diff
@@ -0,0 +1,13 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Select with dependant components renders with error message 1`] = `""`;
```
nickcolley: Is this snapshot correct?

alex-ju: Hmm.. doesn't look right. I'll have a look.

alex-ju: It seems like `select` is the only component that requires a `label` param in order to render the message - that's why we have an empty string there.
```
{% if params.label %}
{{ govukLabel({...}) }}
{% endif %}
```
I will remove the conditional statement for now, for consistency, and we can discuss the approach and apply it consistently.

alex-ju: Updated. I'll leave the commits separated to ease the review. Will rebase afterwards to leave the history clean.

CHANGELOG.md
```diff
@@ -50,6 +50,7 @@ Internal:
 - Add tests for panel component (PR [#500](https://github.com/alphagov/govuk-frontend/pull/500))
 - Add tests for skip-link component (PR [#498](https://github.com/alphagov/govuk-frontend/pull/498))
 - Add tests for file-upload component (PR [#504](https://github.com/alphagov/govuk-frontend/pull/504))
+- Add tests for select component (PR [#506](https://github.com/alphagov/govuk-frontend/pull/506))
```
36degrees: Can we document the 'bug fixes' (select relying on input component) separately as changes?

alex-ju: Yep, that sounds good. Added.

igloosi: 👍

PR #512, opened 2018-02-08
Update flatten path to include _snapshots_ folder

igloosi:
If we want to publish test files and _snapshots_ with them then `flatten` path needs to be amended to include that folder as well.

This is primarily until we update the build pipeline where we wouldn't need to flatten the folders.

NOTE: If we don't want to include tests and _shapshots_ then the copy tasks needs to be amended to exclude those files and copy test to be update to account for the missing files


nickcolley:
Thanks for this Jani 👍

PR #519, opened 2018-02-09
Add error highlighting to form components

hannalaakso:
This PR adds the left error border from Elements to form components by:

- Adding another CSS variable for error border to settings. Also simplifies the border variables somewhat by removing the tablet specific ones and moving the default ones above the more specific border widths.
- Adding a `govuk-o-form-group` object that is used for adding the left error border to form elements, and also for grouping larger groups of form elements to add left error border.
- Wrapping all form elements in `govuk-o-form-group`
- Making checkbox and radios containers the height of children - @alex-ju, I’m aware you had some concerns about this approach. The issue is that the radios/checkboxes are absolutely positioned. We can’t adjust the line height of the text next to them to match the checkboxes height as the text might wrap. Instead we’re using padding to vertically centre the content which throws it off by a pixel or so. For that reason setting the height of the parent which can’t “see” the height of the absolutely positioned content inside fixes this (I know this a bit rambly, happy to talk about this in the office if I’m not making sense 🙂  )
- Adding example of error to preview app for checkboxes component
- Adding examples to preview app in examples/error-messages
- Replacing https://github.com/alphagov/govuk-frontend/pull/494 with a better approach

I used overrides to remove button margin (last two examples) as it doesn’t feel right to handle button margins in the `form-group` styles, having buttons inside form-groups seems like a bit of an edge case. But open to thoughts on this.

Note: Please review in `/examples/error-messages`. It might be easier to look at the commits which separate the work out into smaller chunks. 

_Optional reading, not related to this PR: If you view the page on mobile, there is an issue with margins in the last example. We remove the bottom margin of last-of-type on radios and checkboxes but if you place the item inside a grid item for instance, that rule will apply to all of them and on mobile this is an issue when things stack. I’ll raise another PR for this once I've spoken to @dashouse ._

https://trello.com/c/ioB6wNUq/670-5-add-highlighting-of-errors-to-form-components

### DIFF
--- a/src/components/checkboxes/_checkboxes.scss
+++ b/src/components/checkboxes/_checkboxes.scss
@@ -7,6 +7,8 @@
     display: block;
     position: relative;
 
+    min-height: $govuk-spacing-scale-7;
+
     margin-bottom: $govuk-spacing-scale-2;
     padding: 0 0 0 40px;
 
@@ -25,8 +27,8 @@
     top: 0;
     left: 0;
 
-    width: 40px;
-    height: 40px;
+    width: $govuk-spacing-scale-7;
+    height: $govuk-spacing-scale-7;
 
     cursor: pointer;
 
@@ -52,8 +54,8 @@
     position: absolute;
     top: 0;
     left: 0;
-    width: 40px;
-    height: 40px;
+    width: $govuk-spacing-scale-7;
+    height: $govuk-spacing-scale-7;
     border: $govuk-border-width-form-element solid $govuk-text-colour;
     background: transparent;
 
--- a/src/components/checkboxes/checkboxes.yaml
+++ b/src/components/checkboxes/checkboxes.yaml
@@ -72,3 +72,18 @@ examples:
         text: 'Irish'
       - value: 'other'
         text: 'Citizen of another country'
+
+- name: with-error
+  data:
+    errorMessage:
+      text: 'Please select an option'
+    fieldset:
+      legendHtml:
+        <h3 class="govuk-heading-m">Which types of waste do you transport regularly?</h3>
+    items:
+      - value: animal
+        text: Waste from animal carcasses
+      - value: mines
+        text: Waste from mines or quarries
+      - value: farm
+        text: Farm or agricultural waste
--- a/src/components/date-input/README.md
+++ b/src/components/date-input/README.md
@@ -16,47 +16,52 @@ Find out when to use the Date input component in your service in the [GOV.UK Des
 
 #### Markup
 
+      <div class="govuk-o-form-group govuk-o-form-group--error">
       <fieldset class="govuk-c-fieldset">
 
-      <legend class="govuk-c-fieldset__legend">
-        What is your date of birth?
+        <legend class="govuk-c-fieldset__legend">
+          What is your date of birth?
 
-        <span class="govuk-c-fieldset__hint">For example, 31 3 1980</span>
+          <span class="govuk-c-fieldset__hint">For example, 31 3 1980</span>
 
-        <span class="govuk-c-error-message">
+          <span class="govuk-c-error-message">
           Error message goes here
         </span>
 
-      </legend>
+        </legend>
 
       <div class="govuk-c-date-input" id="dob">
 
-        <div class="govuk-c-date-input__item govuk-c-date-input__item--day">
-          <label class="govuk-c-label govuk-c-date-input__label" for="dob-day">
+          <div class="govuk-c-date-input__item govuk-c-date-input__item--day">
+            <div class="govuk-o-form-group"><label class="govuk-c-label govuk-c-date-input__label" for="dob-day">
             Day
 
           </label>
           <input class="govuk-c-input govuk-c-date-input__input" id="dob-day" name="dob-day" type="text">
-        </div>
+          </div>
+          </div>
 
-        <div class="govuk-c-date-input__item govuk-c-date-input__item--month">
-          <label class="govuk-c-label govuk-c-date-input__label" for="dob-month">
+          <div class="govuk-c-date-input__item govuk-c-date-input__item--month">
+            <div class="govuk-o-form-group"><label class="govuk-c-label govuk-c-date-input__label" for="dob-month">
             Month
 
           </label>
           <input class="govuk-c-input govuk-c-date-input__input" id="dob-month" name="dob-month" type="text">
-        </div>
+          </div>
+          </div>
 
-        <div class="govuk-c-date-input__item govuk-c-date-input__item--year">
-          <label class="govuk-c-label govuk-c-date-input__label" for="dob-year">
+          <div class="govuk-c-date-input__item govuk-c-date-input__item--year">
+            <div class="govuk-o-form-group"><label class="govuk-c-label govuk-c-date-input__label" for="dob-year">
             Year
 
           </label>
           <input class="govuk-c-input govuk-c-date-input__input" id="dob-year" name="dob-year" type="text">
-        </div>
+          </div>
+          </div>
 
-      </div>
+        </div>
       </fieldset>
+    </div>
 
 #### Macro
 
@@ -89,47 +94,52 @@ Find out when to use the Date input component in your service in the [GOV.UK Des
 
 #### Markup
 
+      <div class="govuk-o-form-group govuk-o-form-group--error">
       <fieldset class="govuk-c-fieldset">
 
-      <legend class="govuk-c-fieldset__legend">
-        What is your date of birth?
+        <legend class="govuk-c-fieldset__legend">
+          What is your date of birth?
 
-        <span class="govuk-c-fieldset__hint">For example, 31 3 1980</span>
+          <span class="govuk-c-fieldset__hint">For example, 31 3 1980</span>
 
-        <span class="govuk-c-error-message">
+          <span class="govuk-c-error-message">
           Error message goes here
         </span>
 
-      </legend>
+        </legend>
 
       <div class="govuk-c-date-input" id="dob-errors">
 
-        <div class="govuk-c-date-input__item govuk-c-date-input__item--day">
-          <label class="govuk-c-label govuk-c-date-input__label" for="dob-errors-day">
+          <div class="govuk-c-date-input__item govuk-c-date-input__item--day">
+            <div class="govuk-o-form-group"><label class="govuk-c-label govuk-c-date-input__label" for="dob-errors-day">
             Day
 
           </label>
           <input class="govuk-c-input govuk-c-date-input__input govuk-c-input--error" id="dob-errors-day" name="undefined-day" type="text">
-        </div>
+          </div>
+          </div>
 
-        <div class="govuk-c-date-input__item govuk-c-date-input__item--month">
-          <label class="govuk-c-label govuk-c-date-input__label" for="dob-errors-month">
+          <div class="govuk-c-date-input__item govuk-c-date-input__item--month">
+            <div class="govuk-o-form-group"><label class="govuk-c-label govuk-c-date-input__label" for="dob-errors-month">
             Month
 
           </label>
           <input class="govuk-c-input govuk-c-date-input__input govuk-c-input--error" id="dob-errors-month" name="undefined-month" type="text">
-        </div>
+          </div>
+          </div>
 
-        <div class="govuk-c-date-input__item govuk-c-date-input__item--year">
-          <label class="govuk-c-label govuk-c-date-input__label" for="dob-errors-year">
+          <div class="govuk-c-date-input__item govuk-c-date-input__item--year">
+            <div class="govuk-o-form-group"><label class="govuk-c-label govuk-c-date-input__label" for="dob-errors-year">
             Year
 
           </label>
           <input class="govuk-c-input govuk-c-date-input__input govuk-c-input--error" id="dob-errors-year" name="undefined-year" type="text">
-        </div>
+          </div>
+          </div>
 
-      </div>
+        </div>
       </fieldset>
+    </div>
 
 #### Macro
 
@@ -164,47 +174,52 @@ Find out when to use the Date input component in your service in the [GOV.UK Des
 
 #### Markup
 
+      <div class="govuk-o-form-group govuk-o-form-group--error">
       <fieldset class="govuk-c-fieldset">
 
-      <legend class="govuk-c-fieldset__legend">
-        What is your date of birth?
+        <legend class="govuk-c-fieldset__legend">
+          What is your date of birth?
 
-        <span class="govuk-c-fieldset__hint">For example, 31 3 1980</span>
+          <span class="govuk-c-fieldset__hint">For example, 31 3 1980</span>
 
-        <span class="govuk-c-error-message">
+          <span class="govuk-c-error-message">
           Error message goes here
         </span>
 
-      </legend>
+        </legend>
 
       <div class="govuk-c-date-input" id="dob-day-error">
 
-        <div class="govuk-c-date-input__item govuk-c-date-input__item--day">
-          <label class="govuk-c-label govuk-c-date-input__label" for="dob-day-error-day">
+          <div class="govuk-c-date-input__item govuk-c-date-input__item--day">
+            <div class="govuk-o-form-group"><label class="govuk-c-label govuk-c-date-input__label" for="dob-day-error-day">
             Day
 
           </label>
           <input class="govuk-c-input govuk-c-date-input__input govuk-c-input--error" id="dob-day-error-day" name="dob-day-error-day" type="text">
-        </div>
+          </div>
+          </div>
 
-        <div class="govuk-c-date-input__item govuk-c-date-input__item--month">
-          <label class="govuk-c-label govuk-c-date-input__label" for="dob-day-error-month">
+          <div class="govuk-c-date-input__item govuk-c-date-input__item--month">
+            <div class="govuk-o-form-group"><label class="govuk-c-label govuk-c-date-input__label" for="dob-day-error-month">
             Month
 
           </label>
           <input class="govuk-c-input govuk-c-date-input__input" id="dob-day-error-month" name="dob-day-error-month" type="text">
-        </div>
+          </div>
+          </div>
 
-        <div class="govuk-c-date-input__item govuk-c-date-input__item--year">
-          <label class="govuk-c-label govuk-c-date-input__label" for="dob-day-error-year">
+          <div class="govuk-c-date-input__item govuk-c-date-input__item--year">
+            <div class="govuk-o-form-group"><label class="govuk-c-label govuk-c-date-input__label" for="dob-day-error-year">
             Year
 
           </label>
           <input class="govuk-c-input govuk-c-date-input__input" id="dob-day-error-year" name="dob-day-error-year" type="text">
-        </div>
+          </div>
+          </div>
 
-      </div>
+        </div>
       </fieldset>
+    </div>
 
 #### Macro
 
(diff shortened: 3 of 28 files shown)


src/globals/scss/core/_form-group.scss
```diff
@@ -0,0 +1,15 @@
+@include exports("form-group") {
+
+  .govuk-form-group {
```
36degrees: This is fine for a proof of concept, but I don't think we should be adding things like this to the 'core' styling. I think this might end up needing to become a component, like fieldset.

src/components/checkboxes/_checkboxes.scss
```diff
@@ -6,6 +6,7 @@
 
     display: block;
     position: relative;
+    min-height: 40px;
```
36degrees: Note that this does change the spacing between checkboxes, from:

screen shot 2018-02-09 at 14 38 24gmt


to: 

screen shot 2018-02-09 at 14 38 34gmt

I believe that the new behaviour is actually the 'correct' behaviour, but we should check with @dashouse when he's back.

src/components/fieldset/_fieldset.scss
```diff
@@ -9,6 +9,16 @@
...
+
+    .govuk-c-input:last-of-type,
+    .govuk-c-select:last-of-type {
+      margin-bottom: 0;
+    }
+  }
+
+  .govuk-c-fieldset--error {
+    padding-left: $govuk-spacing-scale-3;
+    border-left: $govuk-border-width-error-field-group solid $govuk-red;
```
36degrees: Should be using the semantic variable – `$govuk-error-colour`.

src/components/fieldset/_fieldset.scss
```diff
@@ -9,6 +9,16 @@
...
     @include govuk-h-clearfix;
+
+    .govuk-c-input:last-of-type,
+    .govuk-c-select:last-of-type {
+      margin-bottom: 0;
+    }
+  }
+
+  .govuk-c-fieldset--error {
+    padding-left: $govuk-spacing-scale-3;
```
36degrees: In Elements, this is responsive - 10px on mobile, 15px on desktop. We don't have an equivalent in our responsive spacing scale, so I think we need to check that this behaviour is correct (unless you already have?)

src/components/input/_input.scss
```diff
@@ -4,14 +4,22 @@
 
 @include exports("input") {
   .govuk-c-input {
+    @include govuk-responsive-margin($govuk-spacing-responsive-6, "bottom");
+  }
+
+  .govuk-c-input--error {
```
36degrees: Am I right in thinking that we're supporting two behaviours here – _either_ having an input in error state, or having a form-group in error state?

I think we should consider simplifying things such that inputs are always in a form-group, and the left-hand border always comes from the form-group?

src/components/radios/_radios.scss
```diff
@@ -8,14 +8,16 @@
...
+
     margin-bottom: $govuk-spacing-scale-2;
     padding: 0 0 0 $govuk-spacing-scale-7;
 
     clear: left;
   }
 
-  .govuk-c-radios__item :last-child,
-  .govuk-c-radios__item :last-of-type {
+  .govuk-c-radios__item:last-child,
```
36degrees: Good catch.

src/globals/scss/settings/_measurements.scss
```diff
@@ -8,7 +8,8 @@ $govuk-gutter-half: $govuk-gutter / 2;
 $govuk-border-width-mobile: 4px;
 $govuk-border-width-tablet: 5px;
 $govuk-border-width-form-element: 2px;
-$govuk-border-width-error: 4px;
```
36degrees: We've got so many border widths being defined here, and I'm not sure where they're all used. It might be good to sit down with Dave and try and work out whether we can simplify this.

src/components/input/_input.scss
hannalaakso: You could have both the larger form group in error state and the input itself in an error state. Like this:
screen shot 2018-02-12 at 17 54 44

Or do you mean that the wrapper (`govuk-c-input`) should actually be `govuk-form-group`? To me it feels more semantic to have the wrapper name sound like part of the component markup (like `govuk-c-input` is here), instead of `govuk-form-group` that might get missed out when copy pasting HTML as it sounds like something that you might want to use if you grouping inputs together. Although that's the way it is in Elements 🤔

src/components/fieldset/_fieldset.scss
hannalaakso: Done!

src/components/checkboxes/_checkboxes.scss
hannalaakso: Ran this past @dashouse and he's okay with it 👍

src/components/fieldset/_fieldset.scss
hannalaakso: Good for @dashouse

src/globals/scss/settings/_measurements.scss
hannalaakso: PR here: https://github.com/alphagov/govuk-frontend/pull/524

src/components/input/_input.scss
```diff
@@ -23,18 +31,18 @@
...
   }
 
-  .govuk-c-input[type="number"] {
+  .govuk-c-input__input[type="number"] {
     -moz-appearance: textfield;
   }
 
-  .govuk-c-input--error {
-    border: $govuk-border-width-error solid $govuk-error-colour;
+  .govuk-c-input__input--error {
```
36degrees: Do we have other places where we're applying a modifier to an element like this? This is probably something we need to try and be consistent on, as another approach would be to do `.govuk-c-input--error .govuk-c-input__input` – i.e. add the modifier to the block, and style elements as children.

src/components/input/_input.scss
36degrees: In that screenshot, the `input__input--error` class is being used to make the input border red, but the additional left-hand border is coming from the form-group, not the `input--error` class that this comment is referencing. If that class was added, you'd end up with two left-hand borders in that screenshot.

So… I guess I am basically saying the second thing? The benefit in my mind is that it's simpler to understand (the left-hand border always comes from the fieldset or form-group) and there's less repetition in the code. But as you say, users might find the form-group confusing.

src/globals/scss/core/_form-group.scss
```diff
@@ -0,0 +1,15 @@
...
+    @include govuk-responsive-margin($govuk-spacing-responsive-6, "bottom");
+
+    .govuk-form-group:last-of-type {
+      margin-bottom: 0; // Remove margin from last item in nested groups of form-groups
+    }
+  }
+
+  .govuk-form-group--error {
+    padding-left: $govuk-spacing-scale-3;
+    border-left: $govuk-border-width solid $govuk-red;
```
36degrees: Related to the whole border width conversation, this should probably be using either `$govuk-border-width-error` or `$govuk-border-width-error-group`, or those variables shouldn't exist and we should just be using `$govuk-border-width`?

src/globals/scss/settings/_measurements.scss
36degrees: Sorry, my original comment wasn't very clear… I was thinking more about trying to avoid introducing yet another variable than about removing any of the existing ones. For example, can the `$govuk-border-width-error-field-group` be avoided by just using `$govuk-border-width`? Or, if we think this variable does need to exist, can we do `$govuk-border-width-error-field-group: $govuk-border-width`?

(Also, can we name `$govuk-border-width-error-field` something like `$govuk-border-width-form-element-error` to make it clear it's the width of the element not the width of the left-hand border? It took me a couple of readings of the code to realise that it was related to the thickness of the border on the input rather than the thickness of the left-hand border.)

src/components/select/_select.scss
```diff
@@ -3,6 +3,15 @@
 
 @include exports("select") {
   .govuk-c-select {
+    @include govuk-responsive-margin($govuk-spacing-responsive-6, "bottom");
+  }
+
+  .govuk-c-select--error {
+    padding-left: $govuk-spacing-scale-3;
```
36degrees: If we decide to keep error states in here (as opposed to only in the form-group) we might want to think about abstracting this into a mixin or something.

alex-ju:
I had an in-person review of this with @hannalaakso earlier today (she kindly walked me through it). My take is that if we're going to use a `form-group` approach, because the `form-group` is essentially a wrapper I would define it as `.govuk-o-form-group` at a global level, as we already have all the wrappers as objects, then I'll use a single modifier `.govuk-o-form-group--error` to throw that error to all components.

Using this wrapper means we don't need additional nesting (e.g. `govuk-c-input__input`) and we don't have to define the `--error` for each component. Also this can potentially help us with conditional revealed content, by using it with a modifier (e.g. `.govuk-o-form-group--revealed`).

I can elaborate or add comments on specific lines, but being more about the approach I felt like I shouldn't hijack the review process.

hannalaakso:
Thanks @alex-ju. I think your approach makes sense 👍  I've amended the markup so that now everything uses `form-o-group` for displaying error messages. I've made it into an object as we have other wrappers in objects. Making `form-group` into a component means that literally the only argument it would be take would be `classes` so having it as a component seems like extra complexity for not much gain.

I spoke to @joelanman about use cases and he thinks having an errored field while highlighting the connected fields is not a thing:
screen shot 2018-02-14 at 09 03 31
He thinks only the fields that have errors in them should be highlighted, in this case the street. If there are multiple errors each of them should have a separate highlight. @dashouse agrees however he wants to do some digging around to see if there are exceptions.

src/globals/scss/core/_form-group.scss
hannalaakso: Thanks @36degrees. I've taken it out of `core` and made it into an object. More info here: https://github.com/alphagov/govuk-frontend/pull/519#issuecomment-365544439

src/components/input/_input.scss
hannalaakso: Thanks @36degrees. Now everything uses `form-o-group`. More info here: https://github.com/alphagov/govuk-frontend/pull/519#issuecomment-365544439

src/components/input/_input.scss
hannalaakso: Thanks @36degrees. This was trying to be consistent with `govuk-c-date-input__input` and `govuk-c-date-input__item--year` but it's no longer a thing as everything uses `form-o-group`.

We should possibly think about this so that we can provide guidance to contributors who want to create new components.

dashouse:
Don't disagree with @joelanman, reason for allowing it to happen (even if not promoted) were:

**Existing group errors, for example:**

screen shot 2018-02-14 at 09 42 09

**In-line input errors**
Any group of inputs that are inline (but not a component)...for example

Telephone number
[+44] [7875693355]

nickcolley:
What are our definitions for using component Vs objects? It's important we are consistent, we have lots of components that only have 'classes' and 'text'.

hannalaakso:
@nickcolley I guess objects act as wrappers for things and don't style the containing content (like `govuk-c-error-message` does which changes the colour of the containing text). In this definition,  `govuk-o-grid` and `govuk-o-width-container` are objects. `govuk-c-fieldset` is a wrapper but it contains a legend within it so it's no longer only a wrapper.

hannalaakso:
I've added the markup to macros (three latest commits) so this PR is now ready to review. 

I'll rebase the commits before merging.

src/globals/scss/settings/_measurements.scss
hannalaakso: Ahh cool. I'll look into this now.

hannalaakso: I've removed the tablet variable, renamed `$govuk-border-width-error-field` (that was a good call) and `$govuk-border-width-form-group-error` now uses `$govuk-border-width`.

joelanman:
@dashouse again, in that login example, do you really need it grouped? It seems very uncommon to have to highlight both being blank like this. I think two errors would be ok?

Also I think both inputs should be highlighted red to be consistent with our error styles.

src/components/file-upload/template.njk
```diff
@@ -1,16 +1,20 @@
 {% from "label/macro.njk" import govukLabel %}
 
-{#- Include label passing all label parameters, merging in `for` and `errorMessage` #}
+<div class="govuk-o-form-group">
```
nickcolley: Should this have the error modifier support?

src/components/input/_input.scss
```diff
@@ -34,7 +34,7 @@
   }
 
   .govuk-c-input--error {
-    border: $govuk-border-width-error solid $govuk-error-colour;
+    border: $govuk-border-width-form-element-error solid $govuk-red;
```
nickcolley: Should we be using $govuk-error-colour here not $red?

src/components/radios/_radios.scss
```diff
@@ -8,14 +8,16 @@
 
     position: relative;
 
+    min-height: 40px;
```
nickcolley: Non-blocking: I see 40px referenced in a few places, should they be a shared variable?

src/components/radios/template.njk
```diff
@@ -2,44 +2,46 @@
...
-      text: item.text,
-      classes: 'govuk-c-radios__label',
-      attributes: item.label.attributes,
-      for: id
-    }) | indent(4) | trim  }}
-  </div>
-  {% endfor %}
-  {% endset %}
+<div class="govuk-o-form-group {%- if params.errorMessage %} govuk-o-form-group--error{% endif %}">
+  <div class="govuk-c-radios
```
nickcolley: The fieldset is nested like

```html
<div class="govuk-o-form-group">
  <div class="govuk-c-radios">
    <div class="govuk-c-fieldset">
    </div>
  </div>
</div>
```

should it be


```html
<div class="govuk-o-form-group">
  <div class="govuk-c-fieldset">
    <div class="govuk-c-radios">
    </div>
  </div>
</div>
```

src/globals/scss/objects/_form-group.scss
```diff
@@ -0,0 +1,23 @@
+@include exports("form-group") {
+
+  .govuk-o-form-group {
+    @include govuk-responsive-margin($govuk-spacing-responsive-6, "bottom");
+
+    .govuk-o-form-group:last-of-type,
+    .govuk-c-input:last-of-type,
+    .govuk-c-select:last-of-type {
```
nickcolley: This seems brittle and incomplete

Is there a generic selector we can use here?

src/globals/scss/objects/_form-group.scss
```diff
@@ -0,0 +1,23 @@
...
+      margin-bottom: 0; // Remove margin from last item in nested groups
+    }
+  }
+
+  .govuk-o-form-group--error {
+    padding-left: $govuk-spacing-scale-3;
+    border-left: $govuk-border-width-form-group-error solid $govuk-red;
+
+    .govuk-o-form-group {
```
nickcolley: Is this the responsibility of the CSS or the use of markup / macros?

dashouse:
@joelanman I'm not sure what the conversation actually is here. I believe it's the same container whether you wrap it around one input or a group of inputs. So if this is more about what we present as best practice I agree absolutely...that example shouldn't be in the design system.

I just wouldn't want to somehow block using it on groups, just because of the unknowns about what it might need to be used for...

src/globals/scss/objects/_form-group.scss
hannalaakso: Yes absolutely, this was an attempt to mitigate the user making a mistake and/or nesting form groups in a way that makes checking for this difficult. In some ways this _is_ more of a visual concern as any form field that has an error class does in fact contain an error, it's just that we shouldn't be showing left error borders within left error borders. I don't know if I'm making any sense...

src/components/radios/template.njk
hannalaakso: I'd say that's not within the scope of this PR. Are you happy for me to raise an issue?

src/components/checkboxes/checkboxes.yaml
```diff
@@ -72,3 +72,18 @@ examples:
...
       - value: 'other'
         text: 'Citizen of another country'
+
+- name: with-error
+  data:
+    errorMessage:
+      text: 'Please select an option'
+    fieldset:
+      legendHtml:
+        <h3 class="govuk-heading-m">Which types of waste do you transport regularly?</h3>
```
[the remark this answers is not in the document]
hannalaakso: I don't think `legendClasses` is currently a thing...? Or is it?

src/components/file-upload/template.njk
hannalaakso: Sorry I should have mentioned this in the commit message. It's a weird one as we don't want to add a red border to this component (confirmed by @dashouse) but I was trying to keep it consistent with the other form components in case we want to do it in the future. Thinking about it now, it's confusing, I'll take it out.

src/components/input/_input.scss
hannalaakso: Yes, good catch.

src/components/radios/_radios.scss
hannalaakso: Good call. Fixed.

src/globals/scss/objects/_form-group.scss
hannalaakso: This was a good call - the bottom margins on input and select are no longer needed as they are no wrapped in form-group that handles the margins 👍

src/components/checkboxes/checkboxes.yaml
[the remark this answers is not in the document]
nickcolley: ah right, nevermind then

src/components/radios/template.njk
nickcolley: Yeah you're right sorry about that, issue would be good.

src/components/date-input/README.md
```diff
@@ -29,32 +29,37 @@ More information about when to use date-input can be found on [GOV.UK Design Sys
...
       </legend>
 
-      <div class="govuk-c-date-input" id="dob">
+      <div class="govuk-o-form-group govuk-o-form-group--error">
+        <div class="govuk-c-date-input" id="dob">
 
-        <div class="govuk-c-date-input__item govuk-c-date-input__item--day">
-          <label class="govuk-c-label govuk-c-date-input__label" for="dob-day">
+          <div class="govuk-c-date-input__item govuk-c-date-input__item--day">
+            <div class="govuk-o-form-group"><label class="govuk-c-label govuk-c-date-input__label" for="dob-day">
```
nickcolley: Is this html rendering as expected?

src/components/radios/template.njk
hannalaakso: Here it is: https://github.com/alphagov/govuk-frontend/issues/531

src/components/date-input/README.md
hannalaakso: I've just tried to improve the indentation of the readme. I think adding a wrapper to the component has thrown it out.

hannalaakso: As @nickcolley has just pointed out there are still some text indentation issues in the README. I don't think fiddling with the spacing and trimming and what not is a good use of my time 😞   if we're looking to implement Prettify. I think if we decide not to use Prettify, we should look at this issue again. Contributors will keep running into it as well.

nickcolley:
Think we'll need to squash the git commits to make the history easier to follow before we can merge.

nickcolley:
Pages to review:
- [Error messages](https://govuk-frontend-review-pr-519.herokuapp.com/examples/error-messages)
- [File upload](https://govuk-frontend-review-pr-519.herokuapp.com/components/file-upload#heading-file-upload--with-error-message)
- [Date input](https://govuk-frontend-review-pr-519.herokuapp.com/components/date-input#heading-date-input--with-errors)
- [Checkboxes](https://govuk-frontend-review-pr-519.herokuapp.com/components/checkboxes#heading-checkboxes--with-error)
- [Fieldset](https://govuk-frontend-review-pr-519.herokuapp.com/components/fieldset#heading-fieldset--with-error-message)
- [Input](https://govuk-frontend-review-pr-519.herokuapp.com/components/input#heading-input--with-error-message)
- [Radios](https://govuk-frontend-review-pr-519.herokuapp.com/components/radios#heading-radios--with-extreme-fieldset)
- [Select](https://govuk-frontend-review-pr-519.herokuapp.com/components/select#heading-select--with-hint-text-and-error)
- [Textarea](https://govuk-frontend-review-pr-519.herokuapp.com/components/textarea#heading-textarea--with error message)

src/components/radios/template.njk
```diff
@@ -2,44 +2,46 @@
...
-      html: item.html,
-      text: item.text,
-      classes: 'govuk-c-radios__label',
-      attributes: item.label.attributes,
-      for: id
-    }) | indent(4) | trim  }}
-  </div>
-  {% endfor %}
-  {% endset %}
+<div class="govuk-o-form-group {%- if params.errorMessage %} govuk-o-form-group--error{% endif %}">
```
nickcolley: I don't think we need this wrapper anymore since this can be applied with the fieldset.

This applies to any components that use a wrapping fieldset.

screen shot 2018-02-16 at 12 55 47

hannalaakso: Thanks @nickcolley, that's now fixed.

hannalaakso:
I've squashed the commits. Note that if you view the order of commits on Github, they don't seem to be in the most logical order as I rebased multiple time but if you do `git log` they are in the right order.

src/globals/scss/objects/_form-group.scss
```diff
@@ -0,0 +1,21 @@
+@include exports("form-group") {
+
+  .govuk-o-form-group {
+    @include govuk-responsive-margin($govuk-spacing-responsive-6, "bottom");
+
+    .govuk-o-form-group:last-of-type {
```
nickcolley: Still not sure about these resets in this file, I don't think having the additional nested classes is a great idea, since they seem to be a limitation of how we're doing our macros.

We could come back to this after this feature is merged though.

hannalaakso: I think this is a valid use case in case of the date-input for instance where you have outer and inner form-group. But happy to review this after.

hannalaakso:
Thanks @nickcolley 🎉 (Just to confirm: the date-input bug you picked up on is now fixed)

joelanman:
Well done @hannalaakso :)

hannalaakso:
Thank you @joelanman, a lot of sweat and tears went into this one 😅

PR #524, opened 2018-02-13
Remove tablet border variable

hannalaakso:
As discussed in #519, there are too many variables for setting borders.

Discussed with @dashouse - we can remove the one for tablet. It's only used by the error summary which can fall back on the standard border variable.


dashouse:
This is fine but might be worth thinking about all border widths being less specific in general and not worrying about specifying names based on usage?

`$govuk-border-width-10: 10px`
`$govuk-border-width-5: 5px`
`$govuk-border-width-4: 4px`
`$govuk-border-width-2: 2px`
`$govuk-border-width-1: 1px`

hannalaakso:
Closing this issue as it was addressed as part of https://github.com/alphagov/govuk-frontend/issues/519

@dashouse I've moved your suggestion into a new issue, hope that's okay: https://github.com/alphagov/govuk-frontend/issues/529

PR #525, opened 2018-02-13
Remove unused colour definitions

36degrees:
There are a load of applied colour definitions that are unused within Frontend.

There's a few things to consider here:

- The panel colour was used by the previous / next component which has been temporarily removed, but it would probably make more sense for that component to reference $govuk-grey-4 directly anyway.
- The colour page currently references a number of these variables, so we'd need to update that page as well (/cc @joelanman) to remove:
  - `$govuk-panel-colour`
  - `$govuk-highlight-colour`
  - `$govuk-page-colour`

This has been done in a few commits so that you can clearly see which variables are being removed amidst the re-structure.

Thoughts?


36degrees:
@nickcolley happy with the changelog entry?

PR #527, opened 2018-02-13
Remove scope=row attribute from non-th elements

### LINKED ISSUE #526: Investigate axe issue with scope attributes on <td>s
aXe fails when testing tables that have

```html<td scope="row></td>```

Steps to take:
1. Figure out if this is a real issue or something that should be ignored.
2. Unskip the test https://github.com/alphagov/govuk-frontend/pull/522/files#diff-2cfe07504444a7f34f0f9b6c29ccca3cR10
3. Update the test with ignored rules or template to remove the attribute depending on 1.

igloosi:
In HTML5 scope attribute should only be used on `<th>` elements according to https://dequeuniversity.com/rules/axe/1.1/scope

This PR removes the wrongly placed scope attribute on `<td>`s
Fixes #526


igloosi:
will need to be updated with #528

PR #532, opened 2018-02-19
Simplify folder structure

igloosi:
There are a lot of changes here and broadly they are:
- merging `packages` and `src` folders into `packages/@govuk-frontend`*
- updating all sass imports to `@govuk-frontend/*`
- updating paths in app to resolve all components an d variants while filtering out all,globals and icons packages
- creating an app specific .scss file that imports `all` for preview and a gulp task that compiles that
- removing some gulp tasks that aren't needed - **more tasks will be updated/remove with the next trello card**
- fix app, generate:readme tasks

*why `@govuk-frontend folder`? Mimicking npm consumption removes the need to rewrite sass imports before publishing to npm

Diff of new vs old compiled css file
screen shot 2018-02-19 at 15 43 43

Trello ticket: https://trello.com/c/kiNBlfHd/667-8-avoid-code-duplication-in-repository


packages/@govuk-frontend/date-input/_date-input.scss
```diff
@@ -21,6 +21,10 @@
     padding-bottom: 2px;
   }
 
+  .govuk-c-date-input__input {
```
nickcolley: Is this change meant to be here?

package.json
```diff
@@ -70,7 +69,6 @@
   },
   "dependencies": {
     "gulp-if": "^2.0.2",
-    "postcss-nested": "^2.0.2",
```
nickcolley: 👏

app/app.js
```diff
@@ -15,7 +15,7 @@ const appViews = [
   configPaths.layouts,
   configPaths.partials,
   configPaths.examples,
-  configPaths.components
+  configPaths.govukFrontend
```
nickcolley: Is there a reason this is not 'components' anymore? It'd allow easier reuse if we don't tie this code to govukFrontend?

packages/@govuk-frontend/date-input/_date-input.scss
igloosi: it's all latest changes i guess

hannalaakso: That's not from the latest round of changes: https://github.com/alphagov/govuk-frontend/commit/b48d3913a4b8f41fda60a0afa22641a3a54e99dd#diff-e94394b2ac1d4f73991af98e4fa34fa3

app/app.js
```diff
@@ -36,7 +36,7 @@ app.set('view engine', 'njk')
 
 // Set up middleware to serve static assets
 app.use('/public', express.static(configPaths.public))
-app.use('/icons', express.static(path.join(configPaths.public, 'icons')))
+app.use('/@govuk-frontend/icons', express.static(path.join(configPaths.components, '/icons/')))
```
nickcolley: Similar to the config paths, if we keep the paths generic 'component', others can use this code for their own review application easier.

igloosi: here i'm linking to the icons package directly. So there is no need to copy icons to the public folder as we do now

alex-ju: funny that we're actually using only one image from there.. the one for the start button, aren't we?

igloosi:
@nickcolley have the ticket to update the build pipeline.  it will be a marginally different step as we only publish sass files atm, but if we do decide to publish compiled and minified css and js in the `all` package, then we can do something similar to current compile task

app/app.js
igloosi: yes

nickcolley: Sorry, this is a bad example to reply to, I meant more widely the paths.

igloosi: oh that. i'll have a look at express url adaption

igloosi:
updated so url is now `/components/<component>`

config/paths.json
```diff
@@ -6,11 +6,8 @@
       "layouts": "app/views/layouts/",
   "config": "config/",
   "dist": "dist/",
-    "distComponents": "dist/components/",
   "packages": "packages/",
+    "components": "packages/@govuk-frontend/",
```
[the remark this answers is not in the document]
igloosi: "packagesRoot": ''
   "packages" :""
perhaps?

igloosi:
Having spent the afternoon drawing out and testing the release process, it became apparent that the process of having a single directory but managing "built" files and git tags would be rather unwieldy (see below)

I'm closing this in favour of a simpler, still two-folder approach, but with a identical tree structure and fewer transformations between them.
img_20180221_162108
img_20180221_162113
img_20180221_162118

PR #545, opened 2018-02-23
Simplify src structure and update tasks for preparing a release

igloosi:
Lost of files changed that results in:
- src has the same tree structure as packages
- 'copy-to-destination' task is slimmed down as it no longer needs to flatten folders, it's more of a straight copy
- app.js references to src have been updated to account for the new path
- views and partials have been updated to account for the new path
- tests have been updated to account for the new folders (all, globals, icons)
- paths in config/paths.json have been updated
- 'compile components' task remove
- package.json tasks update with removal of --production flag and implicit 'public' folder destination
- compilation of 'all' css has been updated do that review app and compile:assets task use the same entry point and import 'all' package scss
- review app uses icons from the 'icons' package directly
- other tasks have been updated with new paths
- package updated with new files
- dist updated with new files

Comparison of src and package scss file

screen shot 2018-02-23 at 12 23 06

Trello ticket
https://trello.com/c/kiNBlfHd/667-8-simplify-src-directory-and-tasks-that-build-packages

Symlinked to DS to check if relative paths from node_modules work fine


CHANGELOG.md
```diff
@@ -35,6 +35,8 @@ Internal:
 - The sass-lint config has been updated to prevent the use of `@extend` with
   concrete classes.
   (PR [#550](https://github.com/alphagov/govuk-frontend/pull/550)
+- Simplify `src/` folder structure, remove tree and string manipulations from
```
nickcolley: Believe this is a breaking change due to renaming 'oldie' to 'old-ie' so let's move this?

igloosi: that file was never published, so dunno. can move the log entry anyway

nickcolley: ah okay, no worries then.

igloosi:
will wait for #552 and then merge for real!

36degrees:
Don't wait for #552, I can rebase it once this has merged.

igloosi:
it's going to be a pain from your side. this one is easier

PR #547, opened 2018-02-23
Fix incorrect selector in 'section break'

36degrees:
The visible class is a modifier and so should be separated from the block using double dashes (--) rather than double underscores (__) which signifies an element.


36degrees:
Rebased to resolve merge conflict in changelog.

PR #551, opened 2018-02-23
Fix non-semantic use of text colour for borders.

36degrees:
At the minute we are using the `$govuk-text-colour` variable for the colour of form control borders, which doesn’t make much sense semantically.

This introduces a new variable, `$govuk-input-border-colour` which is set to the same `$govuk-black` that `$govuk-text-colour` currently refers to, and updates form components to use it for the border colour.

It also modifies the radio and checkbox components to consistently use `currentColor` for the pseudo-elements, and be explicit about doing so.

- [x] Changes documented in changelog


igloosi:
looking good, just the changelog

PR #552, opened 2018-02-23
Various improvements to Button component

36degrees:
I suggest walking through the individual commits to understand the changes.

- [x] Document changes in CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,29 @@
 
 Note: We're not following semantic versioning yet, we are going to talk about this soon.
 
+## Unreleased
+
+Fixes:
+- The transparent outline has been removed from the button, as it already has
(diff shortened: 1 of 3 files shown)


src/components/button/_button.scss
```diff
@@ -130,6 +99,10 @@
       cursor: default;
     }
 
+    :focus {
```
nickcolley: Does this need to be `&:focus`?

36degrees: Good catch! 👏

36degrees:
Rebased against master.

Pushed an additional commit which ensures that link styles are overridden, which was removed in 7fd770c.

Not sure how to document this in the changelog… the only 'effective' change now is the removal of the IE7 outline fix, everything else should just be removing redundant styling and tidying up the Sass. Thoughts?

PR #555, opened 2018-02-27
Remove show mq from dist compilation

igloosi:
Move `$mq-show-breakpoints` from globals to govuk-frontend scss entry point.
This way we can comment it out for dist compilation and keep it for dev/review site.
Importers can then also enable it for their app.

This is not the best solution at the moment, as we have a single entry point for review/dev, packages and dist compilation.

As mentioned in #556 and by Ollie on Slack, when we move to separate entry points for the app and distribution then we can set the variable that includes `$mq-show-breakpoints` for dev only.

### DIFF
--- a/src/globals/scss/govuk-frontend.scss
+++ b/src/globals/scss/govuk-frontend.scss
@@ -1,3 +1,11 @@
+// sass-lint:disable no-css-comments
+// start:devonly
+// If you want to display the currently active breakpoint in the top
+// right corner of your site during development, add the breakpoints
+// to this list, ordered by width, e.g. (mobile, tablet, desktop).
+$mq-show-breakpoints: (mobile, tablet, desktop);
+// end:devonly
+
 // Settings
 @import "../../globals/scss/common";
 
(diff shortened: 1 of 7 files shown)


src/globals/scss/govuk-frontend.scss
```diff
@@ -1,3 +1,11 @@
+// sass-lint:disable no-css-comments
+// start:devonly
```
nickcolley: Is there a better way to do this devonly thing?

Can we use a variable that is set to default off?

nickcolley: I've seen https://gist.github.com/Leimi/a9f125406c1efd916448

nickcolley: or https://www.npmjs.com/package/gulp-sass-variables

nickcolley: Not a big deal otherwise

nickcolley: I think the boring option is separate entry points

nickcolley: I'll raise an issue to think about this later, since we need this in ASAP for research.

nickcolley: https://github.com/alphagov/govuk-frontend/issues/556

igloosi:
@nickcolley can you check the additional commits to see if all is ok?

igloosi:
will need to redo

PR #557, opened 2018-02-28
Namespace all @functions and @mixin's

igloosi:
To avoid breaking compilation when Frontend is imported into an existing app that uses custom
functions or GOV.UK Elements or GOV.UK Frontend toolkit, as observed in user research.

May have gone a bit too far in namespacing all the things, but we can remove some.
Trello ticket: https://trello.com/c/B2tdqPg5/784-fix-em-function-conflict-with-govuk-elements


src/globals/scss/helpers/_arrow.scss
```diff
@@ -2,15 +2,15 @@
...
 // square root of three gives us its height. We use 1.732 as an approximation.
-@function equilateral-height($base) {
+@function govuk-equilateral-height($base) {
   $square-root-of-three: 1.732;
 
   @return ($base / 2) * $square-root-of-three;
 }
 
-// Arrow mixin
+// govuk-shape-arrow mixin
```
36degrees: Overly keen find and replacing throughout this doc block? 🙂

src/globals/scss/helpers/_arrow.scss
```diff
@@ -58,6 +58,6 @@
     border-width: $perpendicular $height $perpendicular 0;
     border-right-color: currentColor; // 2
   } @else {
-    @error "Invalid arrow direction: expected `up`, `right`, `down` or `left`, got `#{$direction}`";
+    @error "Invalid govuk-shape-arrow direction: expected `up`, `right`, `down` or `left`, got `#{$direction}`";
```
36degrees: Overly keen find and replacing here too.

src/globals/scss/tools/_iff.scss
```diff
@@ -1,5 +1,5 @@
 // Wrapper for Sass' built-in `if` function that does not require you to pass
 // a value for $if-false.
-@function iff($condition, $if-true) {
+@function govuk-iff($condition, $if-true) {
```
36degrees: I wonder if this function is worth keeping given the intention was to have a shorter way of writing

```
if($important, !important, null)
```

and now this is:
```
govuk-iff($important, !important)
```

Happy to address this in a later PR though.

src/globals/scss/helpers/_arrow.scss
```diff
@@ -19,10 +19,10 @@
 //
 // We need both because older browsers do not support clip-path.
 //
-// @param {Direction} $direction - Direction for arrow: up, right, down, left.
+// @param {Direction} $direction - Direction for govuk-shape-arrow: up, right, down, left.
 // @param {Base} $base - Length of the triangle 'base' side
 // @param {Height} $height - Height of triangle. Omit for equilateral.
-@mixin arrow($direction, $base, $height: null) {
+@mixin govuk-shape-arrow($direction, $base, $height: null) {
```
36degrees: If we're going to change the name of this to shape-arrow then can we update the filename to `_shape-arrow.scss` too please?

src/globals/scss/tools/_iff.scss
igloosi: happy to keep as `iff`

igloosi:
updated

hannalaakso:
Looks good to me too (meant to say that this morning)

Issue #556, opened 2018-02-27
Change 'devonly' behaviour to avoid magic


nickcolley:
At the moment we find / replace a string to force a portion of the Scss to be a comment which then gets stripped out as other comments do.

My proposal would be to change this so we have an `$enviroment` variable, that we can set to 'development'

https://github.com/alphagov/govuk-frontend/pull/555/files#r170951075

nickcolley:
Fixed as part of https://github.com/alphagov/govuk-frontend/pull/560

PR #560, opened 2018-02-28
Remove dev styling from compiled css and update pipeline

igloosi:
We don't want dev specific css in compiled output in `dist`, as observed in user research.
This PR addresses this by:
- creating separate app only sass entry points where `$mq-show-breakpoints` is included
- updates app layout template to reference app*.css files
- modifying `scss:compile` task to use either `app*.scss` or `govuk-frontend*.scss` files as source for compilation
- adding test to check for presence of that string in compiled stylesheets

Additionally:
 - removes magic commenting with `gulp-replace` and gulp-replaces from dependencies
- adds a file helper function to read contents of a file for a given path

the flow is now as follows:
- if `dev` then compile app*.scss with mq-show-breapointsand anthing else we might want to include in teh review app
- if `dist` then compile govuk-frontend*.scss without mq-show-breakpoints

This should also address #556 

https://trello.com/c/woTnvOTx/801-remove-active-breakpoint-debug-helper-from-dist-css

### DIFF
--- a/app/assets/scss/app-old-ie.scss
+++ b/app/assets/scss/app-old-ie.scss
@@ -1,0 +1,5 @@
+// If you want to display the currently active breakpoint in the top
+// right corner of your site during development, add the breakpoints
+// to this list, ordered by width, e.g. (mobile, tablet, desktop).
+$mq-show-breakpoints: (desktop);
+@import "../../../src/all/all-old-ie";
--- a/lib/file-helper.js
+++ b/lib/file-helper.js
@@ -54,3 +54,10 @@ const expectedPackagesFilesForComponent = componentName => {
   return srcComponentFiles.sort()
 }
 exports.expectedPackagesFilesForComponent = expectedPackagesFilesForComponent
+
+// Read the contents of a file from a given path
+const readFileContents = filePath => {
+  return fs.readFileSync(filePath, 'utf8')
+}
+
+exports.readFileContents = readFileContents
--- a/tasks/gulp/compile-assets.js
+++ b/tasks/gulp/compile-assets.js
@@ -31,9 +31,12 @@ const errorHandler = function (error) {
   this.once('finish', () => process.exit(1))
   this.emit('end')
 }
+// different entry points for both streams below and depending on destination flag
+const compileStyleshet = isDist ? configPaths.app + 'assets/scss/govuk-frontend.scss' : configPaths.app + 'assets/scss/app.scss'
+const compileOldIeStyleshet = isDist ? configPaths.app + 'assets/scss/govuk-frontend-old-ie.scss' : configPaths.app + 'assets/scss/app-old-ie.scss'
 
 gulp.task('scss:compile', () => {
-  let compile = gulp.src(configPaths.app + 'assets/scss/govuk-frontend.scss')
+  let compile = gulp.src(compileStyleshet)
     .pipe(plumber(errorHandler))
     .pipe(sass())
     // minify css add vendor prefixes and normalize to compiled css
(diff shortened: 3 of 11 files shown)


app/assets/scss/app-old-ie.scss
```diff
@@ -0,0 +1,5 @@
+// If you want to display the currently active breakpoint in the top
```
36degrees: Is there a reason these aren't using the [existing app stylesheets](https://github.com/alphagov/govuk-frontend/tree/ad2022ef044d2e67768199479e215f641e36cc5f/app/assets/scss), which are now not being used?

tasks/gulp/compile-assets.js
```diff
@@ -31,9 +30,20 @@ const errorHandler = function (error) {
   this.once('finish', () => process.exit(1))
   this.emit('end')
 }
+// used only for review app as they include showing of media queries
```
36degrees: The existing app assets seem to be getting compiled OK, so not sure why this is needed.

tasks/gulp/compile-assets.js
```diff
@@ -54,42 +64,6 @@ gulp.task('scss:compile', () => {
       })
     ))
     .pipe(gulp.dest(taskArguments.destination + '/css/'))
-
-  let compileOldIe = gulp.src(configPaths.app + 'assets/scss/govuk-frontend-old-ie.scss')
```
36degrees: I don't understand why this is being removed as part of this PR. Don't we still need an old-ie stylesheet?

lib/file-helper.js
```diff
@@ -54,3 +54,11 @@ const expectedPackagesFilesForComponent = componentName => {
   return srcComponentFiles.sort()
 }
 exports.expectedPackagesFilesForComponent = expectedPackagesFilesForComponent
+
+// Read the contents of a file from a given path
+const readFileContents = filePath => {
+  const contents = fs.readFileSync(filePath, 'utf8')
```
36degrees: Can just `return fs.readFileSync(filePath, 'utf8')`

igloosi: yeah could do

app/assets/scss/app-old-ie.scss
igloosi: the task to compile css for dist and review app is the same. tried using the same entry point for both and couldn't find a way to not include breakpoints in dist compiled and include them in review compiled

tasks/gulp/compile-assets.js
igloosi: we still have old ie stylesheet. oldie was not doing anything that we don't currently do at the momene (flatten mq's use rems)

app/assets/scss/app-old-ie.scss
igloosi: govuk-frontend*.scss fiels are used for compilation to dist

tasks/gulp/compile-assets.js
36degrees: Can that change be made in a separate PR? It doesn't seem related to the other changes.

(I think there are some things oldie might be doing or should be doing – patching `::before` into `:before` (which is currently disabled??), changing opacity to filters – but I don't want to hold this PR up)

app/assets/scss/app-old-ie.scss
36degrees: The `govuk-frontend*.scss` files **in the app folder** are used for compilation to dist?

igloosi: yes. happy to find a better place for them.
they used to be in globals and we were filtering them from copy task.

tasks/gulp/compile-assets.js
igloosi: ok, can leave it in and will add required merge-stream back in

igloosi:
reverted removing oldie compilation

igloosi:
@36degrees forgot to add changelog. still ok with it?

36degrees:
LGTM.

PR #561, opened 2018-03-01
Update publishing docs

hannalaakso:
This PR updates the docs to mention updating DS and PK to the latest release.

### DIFF
--- a/docs/publishing.md
+++ b/docs/publishing.md
@@ -59,6 +59,10 @@ scripts in the current directory.*
 
 6. Create a release in Github interface, ddd release notes to and attach the ZIP. 
 
+7. Add Trello cards to "This Sprint" column for
+  - updating GOV.UK Design System to use the latest release
+  - updating the Prototype Kit to use the latest release 
+
 #### Adding a new component
 
 1. In Frontend, for any new components inside `/packages`, add a `package.json`


36degrees:
I'm not sure about this. The 'next frontend release' column is intended to be used to track the things that _have already been done_ but not yet released. With this proposal, we're adding things to that column that _need to be done_ which I think confuses things.

It also implies that the person who did the release has to update both the Design System and the Prototype Kit.

Can I suggest that instead we add new cards to 'This Sprint' column, at the point the release goes out?

docs/publishing.md
```diff
@@ -117,9 +117,14 @@ These tests live in `after-build-dist.test.js`.
 
 16. Open a PR for these changes.
 
-17. Run `lerna clean` to remove `node_modules` directory created in each package. 
+17. Run `lerna clean` to remove `node_modules` directory created in each package.
 As documented in [Lerna docs](https://github.com/lerna/lerna/#clean)
 
+18. Create PRs to update the GOV.UK Design System and the Prototype Kit to use the new release of GOV.UK Frontend. Move the Trello cards to done when finished.
```
nickcolley: Can we setup greenkeeper to do this for us?

joelanman: can greenkeeper work with private npm packages?

nickcolley: Yes but only when paid for, our packages should be open long term.

So not worth thinking about atm true.

hannalaakso:
@36degrees I've reworded this, could you please review again.

PR #568, opened 2018-03-01
Allow input type to be overridden; Update date input to use type="number"

36degrees:

https://trello.com/c/gpdeKWhm/695-make-date-input-component-use-typenumber

### DIFF
--- a/src/date-input/README.md
+++ b/src/date-input/README.md
@@ -37,7 +37,7 @@ Find out when to use the Date input component in your service in the [GOV.UK Des
             Day
 
           </label>
-          <input class="govuk-c-input govuk-c-date-input__input" id="dob-day" name="dob-day" type="text">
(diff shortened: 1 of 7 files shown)


joelanman:
I have the same concerns here as with the `max` property. With type=number, some users will enter things and nothing happens at all - there's no messaging about what's gone wrong. However in this case it does have a significant advantage of displaying a more useful keyboard to touchscreen users. Is type="number" the only way to do that?

edwardhorsford:
From memory, `type=tel` more reliably produces a numeric keypad. I'd have to check our notes.

Nothing happens if a user types a non-numeric character, but I wonder if this is actually a problem in reality - it may help users catch issues quicker - and it's standard browser behavior.

36degrees:
This is just aligning Frontend with what we already have in Elements.

nickcolley:
This needs an addition to the changelog @36degrees

PR #571, opened 2018-03-02
Fix spelling of the colour fuchsia.

36degrees:
As raised in https://github.com/alphagov/govuk_elements/issues/561


nickcolley:
Could we reference the original issue, and if possible credit the original contributor?

36degrees:
Good call – done.

edwardhorsford:
Any thoughts on if this should / could go out with a depreciation warning and not as a breaking change? Unsure if that's possible or not.

36degrees:
It's something we'd like to explore in the future, but we're not set up for that yet.

As only a limited number of beta partners are using it, and I would be very surprised if any of them are using this colour in their applications, I'm happy we can do this as a breaking change.

PR #572, opened 2018-03-05
Add button shim and setup Puppeteer

alex-ju:
This PR:
- [x] adds [shim-links-with-button-role.js](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/javascripts/govuk/shim-links-with-button-role.js) from [govuk_frontend_toolkit](https://github.com/alphagov/govuk_frontend_toolkit/)
- [x] rewrites the code in plain javascript
- [x] updates the code to use common/generic functions (to deal with polyfill; e.g. `addEvent`, `preventDefault`). this functions would ideally sit in a `common.js`/`core.js`/`lib.js` that any component can reference.
- [x] sets up Puppeteer for integration testing and add tests to make sure the shim works in the browser

**Manual tests**
Browsers:
- [x] IE8+
- [x] Latest Safari, Chrome, Firefox, Edge

AT testing
- [x] VoiceOver
- [x] JAWS
- [x] NDVA
- [x] Dragon

**Automated tests**
- [x] Jest integration tests written

**Background on why we need this shim**
_For links that have role="button" we need to add an event handler so that it can be activated with the space bar_ https://github.com/alphagov/govuk_elements/pull/272#issuecomment-233028270

**Notes to reviewer(s)**
Left the commits separated so we can compare the govuk-frontend shim with the govuk_frontend_toolkit one.

[Trello card](https://trello.com/c/b49XDoyE)


src/components/button/button.js
```diff
@@ -0,0 +1,70 @@
...
+      // if the keyCode (which) is 32 it's a space, let's simulate a click.
+      if (GOVUK_FRONTEND.buttons.charCode(event) === 32) {
+        GOVUK_FRONTEND.buttons.preventDefault(event)
+        // trigger the target's click event
+        event.target.click()
+      }
+    },
+
+    init: function () {
+      var buttons = document.querySelectorAll('[role="button"]')
```
36degrees: This will only work for buttons that are present on document ready, not buttons that are loaded in afterwards (e.g. via an AJAX request). This would be a 'regression' from what we've got in Frontend Toolkit, where we use a delegated event handler.

This might be OK, but we might want to consider:

- documenting this limitation
- allow `init` to be run more than once (i.e. ensure we only ever bind to the buttons that we haven't already handled)?

src/components/button/button.js
```diff
@@ -0,0 +1,70 @@
+// javascript 'shim' to trigger the click event of element(s)
```
36degrees: Being a little bit fussy here, but can take this opportunity to wrap comments consistently and use the proper capitalisation?

src/components/button/button.js
```diff
@@ -0,0 +1,70 @@
...
+      if (e.preventDefault) {
+        e.preventDefault()
+      } else {
+        e.returnValue = false
+      }
+    },
+
+    eventHandler: function (event) {
+      // if the keyCode (which) is 32 it's a space, let's simulate a click.
+      if (GOVUK_FRONTEND.buttons.charCode(event) === 32) {
```
36degrees: Would be good to copy [what we've got in the details polyfill](https://github.com/alphagov/govuk-frontend/blob/master/src/details/details.js#L11-L12) and make the space character code a constant.

src/components/button/button.js
```diff
@@ -0,0 +1,70 @@
...
+      }
+    }
+
+  }
+
+  // hand back to global
+  global.GOVUK_FRONTEND = GOVUK_FRONTEND
+
+  // auto-initialise
+  GOVUK_FRONTEND.buttons.init()
```
36degrees: If we're immediately initialising, and this JS only binds to buttons present at the time the script loads, am I right in thinking we're then relying on the user including this JS at the end of their body, rather than in the `<head>`? That seems like it might be quite fragile in use.

Should we either run init on document ready, or use a delegated event handler?

src/components/button/button.js
nickcolley: I believe jQuery uses global document event handling to work around this.

e.g.
```js
document.addEventListener('click', event => {
  if (event.target === rolebutton) {
  }
})
```

alex-ju: @36degrees yep, it's something that I'm looking into.
I was thinking that we can either reinitialise on `DOMContentLoaded` on `document`
`addEvent(document, 'DOMContentLoaded', init)` or watch for the `click` event at the document level as @nickcolley described.

package.json
```diff
@@ -22,7 +22,8 @@
     "build:dist": "node bin/check-nvmrc.js && gulp build:dist --destination 'dist' && npm run test:build:dist",
     "test": "standard && gulp test && npm run test:app && npm run test:components && npm run test:generate:readme",
     "test:app": "jest app/__tests__/app.test.js --forceExit # express server fails to end process",
-    "test:components": "jest src/",
+    "test:components": "jest src/**/template.test.js --env=jsdom",
+    "test:components:integration": "jest src/button/button.test.js",
```
igloosi: how are we going to scale that to multiple components?

igloosi:
I get this
```
 FAIL  src/button/button.test.js
  /components/button
    /components/button/link
      ✕ renders a link button (15ms)
      ✕ triggers the click event when the space key is pressed (9ms)

  ● /components/button › /components/button/link › renders a link button

    net::ERR_CONNECTION_REFUSED

      at navigate (node_modules/puppeteer/lib/Page.js:520:37)

  ● /components/button › /components/button/link › triggers the click event when the space key is pressed

    Protocol error (Runtime.callFunctionOn): Cannot find context with specified idundefined

      at Promise (node_modules/puppeteer/lib/Connection.js:196:56)
      at CDPSession.send (node_modules/puppeteer/lib/Connection.js:195:12)
      at ExecutionContext.evaluateHandle (node_modules/puppeteer/lib/ExecutionContext.js:67:75)
      at ExecutionContext.evaluate (node_modules/puppeteer/lib/ExecutionContext.js:46:31)
      at Frame.evaluate (node_modules/puppeteer/lib/FrameManager.js:299:20)

Test Suites: 1 failed, 1 total
Tests:       2 failed, 2 total
Snapshots:   0 total
Time:        1.084s
Ran all test suites matching /src\/button\/button.test.js/i.
Teardown Puppeteer
Teardown Test Environment.
```

lib/puppeteer_environment.js
```diff
@@ -0,0 +1,33 @@
+const chalk = require('chalk')
+const NodeEnvironment = require('jest-environment-node')
+const puppeteer = require('puppeteer')
+const fs = require('fs')
+const os = require('os')
+const path = require('path')
+
+const DIR = path.join(os.tmpdir(), 'jest_puppeteer_global_setup')
+
+class PuppeteerEnvironment extends NodeEnvironment {
```
36degrees: Would be really good to get some comments throughout this file, especially in the setup function.

lib/puppeteer_environment.js
```diff
@@ -0,0 +1,33 @@
...
+  async setup () {
+    console.log(chalk.yellow('Setup Test Environment.'))
```
36degrees: Are these comments still useful, or should they be removed now?

lib/puppeteer_environment.js
```diff
@@ -0,0 +1,33 @@
...
+      browserWSEndpoint: wsEndpoint
+    })
+  }
+
+  async teardown () {
+    console.log(chalk.yellow('Teardown Test Environment.'))
+    await super.teardown()
+  }
+
+  runScript (script) {
```
36degrees: Do we need to define this function if all we're doing is calling the parent method? Does it not just inherit it?

lib/puppeteer_setup.js
```diff
@@ -0,0 +1,16 @@
...
+const fs = require('fs')
+const mkdirp = require('mkdirp')
+const os = require('os')
+const path = require('path')
+
+const DIR = path.join(os.tmpdir(), 'jest_puppeteer_global_setup')
+
+module.exports = async function () {
+  console.log(chalk.green('Setup Puppeteer'))
+  const browser = await puppeteer.launch({args: ['--no-sandbox']})
```
36degrees: Again, comments here would be good. For example, what is `--no-sandbox` doing? Why is it needed?

src/button/button.js
```diff
@@ -0,0 +1,85 @@
...
+    */
+    preventDefault: function (e) {
+      if (e.preventDefault) {
+        e.preventDefault()
+      } else {
+        e.returnValue = false
+      }
+    },
+
+    eventHandler: function (e) {
```
36degrees: Would be ace to document this in the same way the other functions are documented.

src/button/button.js
```diff
@@ -0,0 +1,85 @@
+/**
+ * JavaScript 'shim' to trigger the click event of element(s) when the space key is pressed.
```
36degrees: Comments in this file are generally 💯. Is the doc block style new? Are there new conventions we're establishing that need to be documented?

src/button/button.test.js
```diff
@@ -0,0 +1,38 @@
...
+      const button = await page.evaluate(() => document.body.getElementsByTagName('a'))
+      expect(button).toBeTruthy()
+    })
+
+    it('triggers the click event when the space key is pressed', async () => {
+      await page.goto(baseUrl + '/components/button/link/preview', { waitUntil: 'load' })
+
+      const href = await page.evaluate(() => document.body.getElementsByTagName('a')[0].getAttribute('href'))
+
+      await page.keyboard.press('Tab')
```
36degrees: Would it be clearer / more explicit to use `page.focus('a')` here?

alex-ju:
@igloosi the server wasn't synchronised with puppeteer. last commit fixes that.

src/button/button.js
nickcolley: We started using them on the lib jest helpers, seems like a good thing to try to do everywhere it makes sense.

alex-ju:
no we haven't looked into it since this PR doesn't rely on `expect-puppeteer` or `jest-puppeteer` as a dependency. we only use puppeteer, so the documentation that I've been looking at was https://github.com/GoogleChrome/puppeteer/blob/v1.1.1/docs/api.md
if we want to, we can change dependencies, rely on `jest-puppeteer` and update the tests accordingly.

lib/puppeteer_environment.js
alex-ju: It's not being used now, but gives the space to allowing extra code before running the script. Can remove.

lib/puppeteer_environment.js
alex-ju: I thought would be good to know what happens and when.
screen shot 2018-03-08 at 15 08 58
Can remove them all.

src/button/button.test.js
alex-ju: Sure. I was replicating my manual testing, but it makes sense to use focus for this.

lib/puppeteer_environment.js
36degrees: Happy to keep them in if they are useful :)

lib/puppeteer_environment.js
```diff
@@ -0,0 +1,33 @@
+const chalk = require('chalk')
+const NodeEnvironment = require('jest-environment-node')
+const puppeteer = require('puppeteer')
+const fs = require('fs')
+const os = require('os')
+const path = require('path')
+
+const DIR = path.join(os.tmpdir(), 'jest_puppeteer_global_setup')
```
nickcolley: Nit: Is underscores a directory thing, should it be kebab case?

lib/puppeteer_environment.js
```diff
@@ -0,0 +1,33 @@
+const chalk = require('chalk')
```
nickcolley: Nit: Can these files live in `/lib/puppeteer/*.*` ?

lib/puppeteer_environment.js
```diff
@@ -0,0 +1,33 @@
...
+const os = require('os')
+const path = require('path')
+
+const DIR = path.join(os.tmpdir(), 'jest_puppeteer_global_setup')
+
+class PuppeteerEnvironment extends NodeEnvironment {
+  async setup () {
+    console.log(chalk.yellow('Setup Test Environment.'))
+    await super.setup()
+    const wsEndpoint = fs.readFileSync(path.join(DIR, 'wsEndpoint'), 'utf8')
```
nickcolley: Can we use `async/await` (https://github.com/alphagov/govuk-frontend/blob/master/tasks/gulp/__tests__/check-generate-readme.test.js#L12)

lib/puppeteer_setup.js
```diff
@@ -0,0 +1,19 @@
...
+const mkdirp = require('mkdirp')
+const os = require('os')
+const path = require('path')
+const app = require('../app/app.js')
+
+const DIR = path.join(os.tmpdir(), 'jest_puppeteer_global_setup')
+
+module.exports = async function () {
+  console.log(chalk.green('\nStart server'))
+  global.__SERVER__ = app.listen(3000)
```
nickcolley: Is app.listen() async? Could this cause a race condition?

lib/puppeteer_setup.js
```diff
@@ -0,0 +1,19 @@
...
+  console.log(chalk.green('Setup Puppeteer'))
+  const browser = await puppeteer.launch({args: ['--no-sandbox']})
```
nickcolley: Can we comment why we're using --no-sandbox?

src/checkboxes/template.test.js
```diff
@@ -1,3 +1,6 @@
+/**
+ * @jest-environment jsdom
```
nickcolley: Is it possible to do this the other way around and opt into puppeteer since jsdom seems like a better default for us?

src/button/button.js
```diff
@@ -0,0 +1,85 @@
...
+      } else {
+        e.returnValue = false
+      }
+    },
+
+    eventHandler: function (e) {
+      // get the target element
+      var target = e.target || e.srcElement
+      // if the element has a role='button' and the pressed key is a space, we'll simulate a click
+      if (target.getAttribute('role') === 'button' && GOVUK_FRONTEND.buttons.charCode(e) === KEY_SPACE) {
```
nickcolley: Could we structure the code so we don't need to reference `GOVUK_FRONTEND.buttons` and rely on the encapsulation from the IIFE?

package.json
```diff
@@ -22,8 +22,8 @@
     "build:dist": "node bin/check-nvmrc.js && gulp build:dist --destination 'dist' && npm run test:build:dist",
     "test": "standard && gulp test && npm run test:app && npm run test:components && npm run test:generate:readme",
     "test:app": "jest app/__tests__/app.test.js --forceExit # express server fails to end process",
-    "test:components": "jest src/ && jest tasks/gulp/__tests__/check-individual-components-compile.test.js",
-    "test:generate:readme": "jest tasks/gulp/__tests__/check-generate-readme.test.js",
+    "test:components": "gulp copy-assets && jest src/ --forceExit && jest tasks/gulp/__tests__/check-individual-components-compile.test.js --forceExit",
```
hannalaakso: Could we document why we need to use `--forceExit` here?

lib/puppeteer_setup.js
alex-ju: yep, thought about that. this may be the cause of the CI failing now.

src/button/button.test.js
```diff
@@ -0,0 +1,38 @@
...
+let baseUrl = 'http://localhost:3000'
+
+beforeAll(async () => {
+  browser = await puppeteer.launch()
+  page = await browser.newPage()
+})
+
+describe('/components/button', () => {
+  describe('/components/button/link', () => {
+    it('renders a link button', async () => {
```
hannalaakso: Do we test for the anchor tag here, as well as in `button/template.test.js`, because here we're using Puppeteer with these tests? If we think we can't rely on `template.test.js`, should we also test for `href` inside this test as the next test will fail with a non helpful error if the `href` missing; this test would be the better place to catch that.

src/button/button.js
```diff
@@ -0,0 +1,85 @@
...
+;(function (global) {
+  'use strict'
+
+  var GOVUK_FRONTEND = global.GOVUK_FRONTEND || {}
+  var KEY_SPACE = 32
+
+  GOVUK_FRONTEND.buttons = {
+
+    /**
+    * Add event construct for modern browsers or IE
```
nickcolley: Anywhere we are polyfilling older browsers can we list the browser support?

This will help future people remove code if they know a browser is no longer needed for example.

nickcolley: jQuery does this like so

```js
// Support: IE 11
if (bunchOfCodeHere) {}
```

src/button/button.test.js
nickcolley: I'm not sure this test is necessary at all since the next test would fail if a link did not render?

lib/puppeteer_environment.js
alex-ju: done

lib/puppeteer_environment.js
alex-ju: done

src/checkboxes/template.test.js
alex-ju: done

src/button/button.test.js
alex-ju: removed

lib/puppeteer_setup.js
alex-ju: comment added

src/button/button.js
alex-ju: updated as suggested

src/button/button.js
alex-ju: I'm trying to follow the [existing js styleguide](https://github.com/alphagov/styleguides/blob/master/js.md) happy to chat about it if you feel we can improve it.

package.json
alex-ju: it is 'documented' on line 24. would be good to look into is and see if we can make it work without `--forceExit`, but maybe not part of this PR

.travis.yml
```diff
@@ -1,4 +1,8 @@
+os:
+- linux
 language: node_js
+node_js:
```
nickcolley: See if we can remove these changes and see if it still works?

alex-ju: sure thing. updated. works as expected.

PR #569, opened 2018-03-01
Add footer component

nickcolley:
👉link to review: https://govuk-frontend-review-pr-569.herokuapp.com/components/footer

New version of the GOV.UK Footer based off the work done in the GOV.UK Design System.

## Notable changes
- Macro interface that supports custom navigation and meta navigation.
- B.E.M.
- Uses flexbox for sizing sections
- Uses 'columns' for splitting the footer list (potential accessibility improvement)
- Uses inline SVG for OGL logo (can be see in inverse)
- Changes OGL logo from duplicate link to presentational element.
- Copyright crest has touch area that is more appropriate.

## Testing

<details>
<summary>Inverse colours ✅</summary>
screen shot 2018-03-05 at 17 12 49
</details>

<details>
<summary>Cross browser testing ✅</summary>

### Known issues:
- Note that some of the older browser have centered crest logo, this is now left aligned to fix an issue and section break. Has been tested with my eyes but not re-screenshotted.

<details>
<summary>IE8</summary>

ie8 top

ie8 bottom

Fixed (html5 shiv):

fixed ie8 top
fixed ie8 bottom

</details>

<details>
<summary>IE9</summary>

ie9 top

ie9 bottom
</details>

<details>
<summary>IE11 (IE10 looks similar)</summary>

ie11

</details>

<details>
<summary>Edge</summary>
edge

</details>

<details>
<summary>Windows, Chrome</summary>
windows chrome

</details>

<details>
<summary>Windows, Firefox</summary>
windows firefox

</details>

<details>
<summary>Mac, Safari</summary>
mac safari
</details>

<details>
<summary>Mac, Chrome</summary>
mac chrome
</details>

<details>
<summary>Mac, Firefox</summary>
mac firefox
</details>

<details>
<summary>iPhone 6s</summary>
iphone 6s

After fix:
iphone 6s

</details>

<details>
<summary>iPhone X</summary>
iphonex top
iphonex bottom
</details>
<details>
<summary>Android, Chrome</summary>
android chrome top
android chrome bottom
</details>

<details>
<summary>Android, Firefox</summary>
android firefox top
android firefox bottom
</details>

<details>
<summary>Windows phone</summary>
windows phone

After fix:
windows phone after fix

</details>

---

| Operating system | Browser                             | Tested                                                                                                         |
|------------------|-------------------------------------|----------------------------------------------------------------------------------------------------------------|
| Windows          | Internet Explorer 8-10              | ✅ Windows 7, IE 8 - 10                                                                                           |
[...]

PR #574, opened 2018-03-05
Create package.json for new components in `pre-release` step

igloosi:
If a new component has been added to `src`, it needs a package.json file when it gets copied to `packages/`

We use to do this manually and our tests that run after `build:packages` task would fail.

With this script, when `npm run pre-release` is run, the script will check if any new folders exist in package after the copy task. 
If so, it will create a sample package.json file with actual component name, version 0.0.0 and latest version of globals as dependency.

it will also remind the developer to amend `all/package.json` file with new dependencies.

new-component-prompt

When developer indicates they have done the above, the script will continue and the rest of the `npm run pre-release` tasks will run.

Next new component will be: #569


bin/check-and-create-package-json.sh
```diff
@@ -0,0 +1,46 @@
...
+
+        if ! [ -e ${D}/package.json ]; then
+          # ${D##*/} strips everything before /
+          COMPONENT_NAME=${D##*/}
+          echo "⚠️ 🆕  $COMPONENT_NAME 🆕  component is missing a package.json file.\n"
+          echo "⚠️ I've created a sample package.json file. Please amended it with correct data.\n"
+
+          create_package_json ${D}
+
+          echo "⚠️ Remember to add the 🆕  $COMPONENT_NAME 🆕  component to ALL/package.json\n"
```
alex-ju: should `all` here be lowercase?

PR #578, opened 2018-03-06
WIP: Toggle component full width for preview

igloosi:
Todo:

- [ ] support legacy ie?

screen shot 2018-03-06 at 11 06 39
screen shot 2018-03-06 at 11 06 31
screen shot 2018-03-06 at 11 07 38


36degrees:
Do we actually want to be able to toggle the full-width-ness for components, or do we just want to present some components differently?

Wonder if this would be better implemented as a flag in the component yaml file that allows us to present some components as 'full width'.

igloosi:
that would be useful, though we would need to find a way to affect the component view template that inherits `govuk-o-width-container` width restriction from `layout` view

igloosi:
closing this for now

PR #583, opened 2018-03-07
Add error summary examples, add margin to error summary

36degrees:
Add examples of how the error summary component interacts with other elements on the page, in both a ‘one thing per page’ context and where multiple form controls are provided.

Add a bottom margin to error summary component – the error summary should always be the first thing on the page, after any breadcrumbs or back link. This means it will always appear before a heading, in which case we know that we will always need to add a margin to separate it from the heading.

https://trello.com/c/hg7MaHJk/809-think-about-how-the-error-summary-works-in-context-especially-in-a-one-thing-per-page-context

- [x] Update CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ Note: We're not following semantic versioning yet, we are going to talk about th
 
 ## Unreleased
 
+Breaking changes:
+
+- The error summary component now has a default bottom margin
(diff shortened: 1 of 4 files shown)


36degrees:
@nickcolley Happy with the changelog entry?

nickcolley:
@36degrees would say it's a breaking change maybe? but private beta happy to approve

36degrees:
Fair – it changes the layout of that component considerably.

PR #585, opened 2018-03-07
make section break line 1px instead of 2px

### LINKED ISSUE #577: govuk-section-break line width is too large
Section breaks are 2px in width when they should only be 1px in width.

screen shot 2018-03-06 at 10 17 24

I imagine it just needs changing from 

```css
.class {
  border: 1px solid grey;
}
```

to

```css
.class {
  border: 0;
  border-bottom: 1px solid grey;
}
```

joelanman:
fixes #577


nickcolley:
Nevermind, I see they're reset above :)

PR #587, opened 2018-03-08
Ensure render function does not have undefined object

nickcolley:


igloosi:
don't all components need  `const $ = render('component', {})` change made?

PR #588, opened 2018-03-08
Fix components relying on global builds

nickcolley:


CHANGELOG.md
```diff
@@ -26,6 +26,8 @@ Internal:
 (PR [#574](https://github.com/alphagov/govuk-frontend/pull/574))
 - Ensure render function does not have undefined object
   (PR [#587](https://github.com/alphagov/govuk-frontend/pull/587))
+- Fix components relying on global builds
```
36degrees: Given footer hasn't been released yet, is this something we need to document? As the thing we're fixing was never actually released?

nickcolley: Agreed, that's why this change is under 'Internal' not 'Fix'

package.json
```diff
@@ -22,7 +22,7 @@
     "build:dist": "node bin/check-nvmrc.js && gulp build:dist --destination 'dist' && npm run test:build:dist",
     "test": "standard && gulp test && npm run test:app && npm run test:components && npm run test:generate:readme",
     "test:app": "jest app/__tests__/app.test.js --forceExit # express server fails to end process",
-    "test:components": "jest src/",
+    "test:components": "jest src/ && jest tasks/gulp/__tests__/check-individual-components-compile.test.js",
```
36degrees: Should this run all tests under `tasks/gulp/__tests__` rather than a specific case?

tasks/gulp/__tests__/check-individual-components-compile.test.js
```diff
@@ -0,0 +1,37 @@
...
+      if (error) {
+        return reject(error)
+      }
+      return resolve(result)
+    })
+  })
+}
+
+describe('Individual components', () => {
+  it('should compile individual scss files', done => {
```
36degrees: Can we have an expectation in this test (expect not to throw an exception?) to make it explicit what this is testing?

package.json
nickcolley: Some of the task tests run 'after' tasks.

36degrees: 👍

PR #595, opened 2018-03-09
Remove duplicated arrow helpers file

36degrees:
This file was renamed to _shape-arrow.scss but the original file was accidentally retained when resolving a merge conflict.


igloosi:
don't know if this is worth a changelog entry

PR #598, opened 2018-03-12
Create CODE_OF_CONDUCT.md

nickcolley:
This PR adds a code of conduct that we can use across our projects.

It's a fork to make sure we can enforce the Code of Conduct properly.

If this is contentious we could scope this individually to each project.

Thoughts welcome.

Alternatively I'm trying to track down who owns the current email address, we may be able to use that instead.


nickcolley:
Had a chat with @timblair and @36degrees is a member of the original mailing list, I'm going to simplify this to just pointing at the master CoC :)

PR #602, opened 2018-03-13
Update `govuk-main-wrapper` object and add a large varient

dashouse:
This PR:
- Updates the `govuk-main-wrapper` class to use the responsive spacing mixin, rather than the fixed spacing scale with media queries.
- Adds a large variant of `govuk-main-wrapper` (`govuk-main-wrapper--l`) for use on pages where there is no Back link, Breadcrumb or Phase banner. This applies some more space to the top and bottom of the page.

With breadcrumb
screen shot 2018-03-13 at 11 54 47

Without breadcrumb (standard `govuk-main-wrapper`
screen shot 2018-03-13 at 11 55 09

Without breadcrumb (using `govuk-main-wrapper--l`)
screen shot 2018-03-13 at 11 55 27

### DIFF
--- a/src/globals/objects/_main-wrapper.scss
+++ b/src/globals/objects/_main-wrapper.scss
@@ -1,4 +1,4 @@
-// Example usage:
+// Example usage with Breadcrumbs, phase banners, back links:
 // <div class="govuk-o-width-container">
 //   <!-- Breadcrumbs, phase banners, back links are placed in here. -->
 //   <div class="govuk-o-main-wrapper">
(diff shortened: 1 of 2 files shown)


nickcolley:
Nice approach.

Is this useful across services or just for us on the design system?

alex-ju:
Definitely a good approach. Did some testing (captures below).

My only question is if we should have the page title aligned regardless of having or not a back-link on the page. (e.g. when a service has the back button only on some pages to have the title in the same y-axis location).

Otherwise I'm happy for this to get merged - technically approved.

wrapper-desktop-back-link
wrapper-desktop-no-back-link
wrapper-mobile-back-link
wrapper-mobile-no-back-link

src/globals/objects/_main-wrapper.scss
```diff
@@ -1,22 +1,38 @@
...
 //   <!-- Breadcrumbs, phase banners, back links are placed in here. -->
 //   <div class="govuk-o-main-wrapper">
 //       <!-- Wrapper for the main content of your page which applies padding
 //            to the top / bottom -->
 //   </div>
 // </div>
+//
+// Example usage without Breadcrumbs, phase banners, back links:
+// <div class="govuk-o-width-container">
+//   <div class="govuk-o-main-wrapper--l">
```
36degrees: Modifiers (`--`) should always be used _with_ the classes that they are modifying – in which case this would be `class="govuk-o-main-wrapper govuk-o-main-wrapper--l"`.

src/globals/objects/_main-wrapper.scss
```diff
@@ -1,22 +1,38 @@
...
+  @include govuk-responsive-padding($govuk-spacing-responsive-6, "bottom");
+}
 
-  @include mq($from: tablet) {
-    padding: $govuk-spacing-scale-6 0;
-  }
+// Use govuk-main-wrapper--l when you page does not have Breadcrumbs, phase banners or back links
+@mixin govuk-main-wrapper--l {
+  @include govuk-responsive-padding($govuk-spacing-responsive-8, "top");
+  @include govuk-responsive-padding($govuk-spacing-responsive-8, "bottom");
```
36degrees: Not sure why the presence (or absence) of breadcrumbs / back links would affect the padding at the bottom of the page – should this still be `responsive-6`?

dashouse:
I've fixed @36degrees 's issues.
@alex-ju I originally intended for this to position the H1 in the same place, irregardless of back link or breadcrumb, however in practice 80px at the top looks too much.

I've messed up these commits slightly I think, had some issues with a rebase on the CHANGELOG

dashouse:
Merged

PR #603, opened 2018-03-13
Review app - Use layout wrappers and add back links

dashouse:
- use our layout wrappers in the review app
- add back links on example pages
- update some headings for consistency


36degrees:
I am seeing some funky behaviour with the example iframes as a result of this change (I think…)

screen shot 2018-03-14 at 09 32 15

screen shot 2018-03-14 at 09 33 08

screen shot 2018-03-14 at 09 33 12

dashouse:
Damn...

I removed the inline style `body { margin: 5px }` but it didn't look like this on my branch locally.

dashouse:
Just trying to find a fix now

PR #605, opened 2018-03-14
Fix margin on example iframes

dashouse:
This fixes the bug introduced by #603

iframe container previously had an inline style of `body { margin: 15px }` which was removed as it was applying a 15px margin to the `<body>` of all layouts.

This caused a bug where the contents flowed outside of their iframe so the inline margin declaration has been put back in using the specific class name of the container.


dashouse:
merged

PR #607, opened 2018-03-15
Fix inline radios

36degrees:
Turns out we do actually have a style for inline radios, it's just not documented or easy to apply. (h/t to @dashouse for spotting this…)

screen shot 2018-03-15 at 09 31 17

Introduce a modifier at the block level which makes all radios within that block inline. You can’t add classes to the individual radio items using the macro, and it doesn’t really make sense to require the modifier to be applied to every individual element.

Add an example of inline radios so that they are represented in the review app.

Update the readme.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 Note: We're not following semantic versioning yet, we are going to talk about this soon.
 
+## Unreleased
+
+Breaking changes:
+
+- Removed an (undocumented) modifier `govuk-c-radio__item--inline` which made
+  radio buttons inline, in favour of a new block-level modifier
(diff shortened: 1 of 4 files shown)


igloosi:
CH CH CH CHANGELOG :)

36degrees:
Updated.

🎵*Just gonna have to be a different man*
*Time may change me,*
*But I can't trace time* 🎵

36degrees:
Sorry @igloosi – one more commit to check.

igloosi:
it's good, though i feel like using inline radios is when you have shortly worded selection which should fit side-by-side on mobile.
but you never know, so better be safe ✅

36degrees:
Don't disagree, but from talking to @dashouse I think matching the existing behaviour in Elements is the right thing to do for now.

PR #609, opened 2018-03-16
Defend against a:link:focus in GOV.UK Template

36degrees:
GOV.UK Template includes [a specific `a:link:focus` selector](https://github.com/alphagov/govuk_template/blob/73cf50c02ae335159ae241387d2ddd61f50d1b23/source/assets/stylesheets/_accessibility.scss#L63-L66) designed to make unvisited links a slightly darker blue when focussed, so we need to override the text colour for that combination of selectors so that the links within the various components do not end up with dark blue text when focussed when that is not the desired behaviour.

https://trello.com/c/rD8KRQle/816-remove-blue-colour-on-unvisited-button-coming-from-govuk-template
https://trello.com/c/fOiHvIjh/817-remove-blue-colour-on-unvisited-back-link-coming-from-govuk-template
https://trello.com/c/N3sczGfq/818-remove-blue-colour-on-unvisited-muted-link-coming-from-govuk-template

### DIFF
--- a/src/back-link/_back-link.scss
+++ b/src/back-link/_back-link.scss
@@ -22,10 +22,18 @@
     // Underline is provided by a bottom border
     text-decoration: none;
 
+    // Override link colour to use text colour
+    //
+    // GOV.UK Template includes a specific a:link:focus selector designed to
+    // make unvisited links a slightly darker blue when focussed, so we need to
+    // override the text colour for that combination of selectors so so that
+    // the skip link does not end up with dark blue text when focussed. (1)
     &:link,
     &:visited,
     &:hover,
-    &:active {
+    &:active,
+    &:focus,
+    &:link:focus {
       @include govuk-text-colour;
     }
 
--- a/src/back-link/_back-link.scss
+++ b/src/back-link/_back-link.scss
@@ -4,7 +4,8 @@
 
   .govuk-c-back-link {
     @include govuk-font-regular-16;
-    @include govuk-focusable-fill;
+    @include govuk-link-common;
+    @include govuk-link-style-text;
 
     display: inline-block;
     position: relative;
@@ -21,21 +22,6 @@
 
     // Underline is provided by a bottom border
     text-decoration: none;
-
-    // Override link colour to use text colour
-    //
-    // GOV.UK Template includes a specific a:link:focus selector designed to
-    // make unvisited links a slightly darker blue when focussed, so we need to
-    // override the text colour for that combination of selectors so so that
-    // the skip link does not end up with dark blue text when focussed. (1)
-    &:link,
-    &:visited,
-    &:hover,
-    &:active,
-    &:focus,
-    &:link:focus {
-      @include govuk-text-colour;
-    }
 
     // Prepend left pointing arrow
     &:before {
--- a/src/breadcrumbs/_breadcrumbs.scss
+++ b/src/breadcrumbs/_breadcrumbs.scss
@@ -103,21 +103,7 @@
   }
 
   .govuk-c-breadcrumbs__link {
-    @include govuk-focusable-fill;
-
-    // Override link colour to use text colour
-    //
-    // GOV.UK Template includes a specific a:link:focus selector designed to
-    // make unvisited links a slightly darker blue when focussed, so we need to
-    // override the text colour for that combination of selectors so so that
-    // the skip link does not end up with dark blue text when focussed. (1)
-    &:link,
-    &:visited,
-    &:hover,
-    &:active,
-    &:focus,
-    &:link:focus {
-      color: $govuk-text-colour;
-    }
+    @include govuk-link-common;
+    @include govuk-link-style-text;
   }
 }
--- a/src/breadcrumbs/_breadcrumbs.scss
+++ b/src/breadcrumbs/_breadcrumbs.scss
@@ -105,10 +105,18 @@
   .govuk-c-breadcrumbs__link {
     @include govuk-focusable-fill;
 
+    // Override link colour to use text colour
+    //
+    // GOV.UK Template includes a specific a:link:focus selector designed to
+    // make unvisited links a slightly darker blue when focussed, so we need to
+    // override the text colour for that combination of selectors so so that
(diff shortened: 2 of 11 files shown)


alex-ju:
I would like to add something to this. The GOV.UK template selector was meant to make links slightly darker when focused - regardless of them being visited or not - as it says in the comment above that line `/* Make links slightly darker when focused to improve contrast. */` - but the implementation was wrong at that time - a `a:link:focus` selector being using instead of a `a:focus`, thinking that `:link` will only apply it to links (e.g. `a`s).

I raised [#325](https://github.com/alphagov/govuk_template/pull/326) months ago for this to be fixed, but it got lost in translation.

Otherwise defending the links from such situations (chaining pseudo-classes) is a good thing.

36degrees:
> The GOV.UK template selector was meant to make links slightly darker when focused - regardless of them being visited or not - as it says in the comment above that line /* Make links slightly darker when focused to improve contrast. */ - but the implementation was wrong at that time - a a:link:focus selector being using instead of a a:focus, thinking that :link will only apply it to links (e.g. as).

I _think_ in this case the implementation is actually right and the comment is wrong (or, at least, not clear) – from Ed's comment [here](https://github.com/alphagov/govuk_template/pull/272#discussion_r98440877) the intention was _just_ to make the unvisited links darker – leaving the visited links purple which already meets colour contrast - that being the case `:link:focus` is actually what is required.

> Defending the links from such situations (chaining pseudo-classes) is a good thing, but if we defend against :link:focus, why not defending against :link:focus:hover and all other pseudo-classes combinations?

Because this is a specific fix against an issue integrating Frontend into a service that uses GOV.UK Template, where these selectors are found. Other combinations (thankfully!) don't exist, and I wouldn't look to defend against those either.

> My opinion is that the first action should be to fix govuk_template issue, and then figure out if, how and in what degree we defend components, especially when chaining pseudo-classes.

I don't _think_ we can assume that those looking to move to GOV.UK Frontend will always be running the latest version of GOV.UK Template – some users might still be on the current version.

alex-ju:
If _the intention was just to make the unvisited links darker_ then my PRs not solving it so I'll close it now. The fact that other combinations don't exist, doesn't mean it won't be anytime soon.. but got you point.

edwardhorsford:
From memory, I was trying to specifically target unvisited links as these were the ones that failed contrast checks - not visited links.

src/breadcrumbs/_breadcrumbs.scss
```diff
@@ -105,10 +105,18 @@
...
+    //
+    // GOV.UK Template includes a specific a:link:focus selector designed to
+    // make unvisited links a slightly darker blue when focussed, so we need to
+    // override the text colour for that combination of selectors so so that
+    // the skip link does not end up with dark blue text when focussed. (1)
     &:link,
     &:visited,
     &:hover,
-    &:active {
+    &:active,
```
nickcolley: Could we break these out into their own sections and do something like

```scss
.class {
  &:link,
  &:visited,
  &:hover,
  &:active {
    color: $govuk-text-colour;
  }

  if ($with-govuk-template) {
    ... overrides here
  }
}

File: src/back-link/_back-link.scss [hunk 2]
```diff
@@ -22,10 +22,18 @@
     // Underline is provided by a bottom border
     text-decoration: none;
 
+    // Override link colour to use text colour
+    //
+    // GOV.UK Template includes a specific a:link:focus selector designed to
```
nickcolley: Instead of 'GOV.UK Template' should we use 'alphagov/govuk_template' which is less ambiguous for people in the future?

If we intend to deprecate GOV.UK Template there will be people in the future that don't even know what this is.

src/breadcrumbs/_breadcrumbs.scss
36degrees: I like this in principle, but not sure how it'll work in practise – especially for users that are not compiling the Sass themselves. I'm not sure what we'd do for the `dist` build, for example?

Also, I'm not sure that we have the tooling in place to have confidence in this when introducing yet more complexity into the builds. In the example above we'd need to set `color: $govuk-text-colour;` in two places, and I could e.g. foresee one being changed without the other which could lead to inconsistent and hard to reproduce problems.

36degrees: Pushed [a branch to show what this might look like in practise](https://github.com/alphagov/govuk-frontend/compare/override-link-focus-in-govuk-template...link-focus-compatibility-variation?expand=1)

nickcolley: For general builds we would set these to on by default, as is the same as your current code.

nickcolley: Low-fi is just to make sure each selector that overrides for govuk-template has a comment next to it, which most of these do?

igloosi:
this looks good. 👍 
want to wait for this to be approved to include it in #617

36degrees:
> It would be good if the useful code comments, especially from /globals/core/_links.scss, could at some point make their way into our docs as instructions on how to set up correctly styled links in your service.

Is https://govuk-design-system-production.cloudapps.digital/styles/typography/#links the sort of thing you're thinking of?

hannalaakso:
Yes that's capturing some of it but maybe on a slightly higher level 👍 I'm thinking about adding something about the compatibility modes and how to use things like the mixins together with each other such as this comment from `_links.scss`: 

> If you use this mixin in a component you must also include the govuk-link-common mixin in order to get the focus state.

However I wonder if some of this is too low level for the Design System and should live in GOV.UK Frontend `/docs` (or include a reference to go look in `_links.scss` etc.), I'm thinking that not everyone will go rooting around the core scss files but they might look at `/docs`. 

I might be making too many assumptions here and we should just take this to user research.

igloosi:
chnagelog :)

36degrees:
Which abstraction, sorry? `govuk_compatibility`?

nickcolley:
@36degrees the compatibility one makes sense, the other ones add more layers, we should merge this.

PR #615, opened 2018-03-22
Update documentation based on user research findings

hannalaakso:
This PR updates the main README with the following:
- Improve docs around Sass paths
- Improve docs around static assets; add guidance on how to use `$global-images` to set images path 
- Document the difference between the `govuk-frontend.min.css` and
`govuk-frontend-old-ie.min.css`. 
- Wrap ` [latest version].min.css` in the "Include assets" example in a conditional comment so that IE8 doesn't download two stylesheets - if we want to do this, `layout.njk` should be updated to do the same
- Document minimum version of Nunjucks in macros: 3.0.0 (2.5.2 and earlier throw up an error due to whitespacing being handled differently [thanks Alex!])
- Document mimimum version of node.js required: 8.3.0 (8.2.0 and earlier throw up console errors)
- General: When writing guidance don't specify a technology or make it clear it's an example
- Clarify what DS is and that it contains a demo

This PR updates the component READMEs with the following:
- Add Nunjucks import macro statement 

Other
- Prefix `$global-images` variable in `settings/_paths.scss`

Note: 
- I've split the main README changes into separate commits to make them easier to review but I'll squash them into one commit before merging this.
- I'll update the changelog once the changes have been approved.

https://trello.com/c/dTNV2ggW/833-update-govuk-frontend-readme-based-on-feedback-from-user-research

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,9 +9,7 @@ GOV.UK Frontend ·
 
 GOV.UK Frontend contains the code you need to start building a user interface for government platforms and services.
 
-## Before using GOV.UK Frontend
-
-You should start by familiarising yourself with the [GOV.UK Design System](https://govuk-design-system-production.cloudapps.digital/), which contains guidance and examples of components, styles and patterns you can use to design  government platforms and services.
+See live examples of GOV.UK Frontend components, and guidance on when to use them in your service, in the [GOV.UK Design System](https://govuk-design-system-production.cloudapps.digital/).
 
 ## How to use GOV.UK Frontend
 
@@ -24,8 +22,13 @@ There are 2 ways to use GOV.UK Frontend:
 ## Requirements
 To use GOV.UK Frontend with NPM you must:
 
-1. Install the long-term support (LTS) version of [Node.js](https://nodejs.org/en/), which includes NPM.
+1. Install the long-term support (LTS) version of [Node.js](https://nodejs.org/en/), which includes NPM. The minimum version of Node required is 8.6.0. (We recommend using [`nvm`](https://github.com/creationix/nvm) for managing versions of Node.)
 2. Create a [package.json file](https://docs.npmjs.com/files/package.json) if you don’t already have one. You can create a default `package.json` file by running `npm init` from the root of your application.
+3. If you want to use the GOV.UK Frontend Nunjucks macros, install Nunjucks - the minimum version required is 3.0.0.
+
+```
+npm install nunjucks --save
+```
 
 ## Installation
 GOV.UK Frontend is currently in private beta. You will need to log in to NPM using credentials provided by the Design System team.
@@ -44,7 +47,6 @@ To install all components, run:
 npm install --save @govuk-frontend/all
 ```
 
-
 To install individual components (for example, a button), run:
 ```
 npm install --save @govuk-frontend/button
@@ -55,32 +57,65 @@ After you have installed GOV.UK Frontend the `@govuk-frontend` package will appe
 
 ## Import assets
 
-Import styles into your main Sass file.
+### Import styles
 
-To import all components, add the line:
+You need to import the GOV.UK Frontend styles into the main Sass file in your project. You should place the below code before your own Sass rules (or Sass imports) if you want to override GOV.UK Frontend with your own styles.
+
+1. To import all components, add the below to your Sass file:
 ```CSS
-@import "@govuk-frontend/all/all";
+@import "node_modules/@govuk-frontend/all/all";
 ```
 
-To import an individual component (for example, a button), add the line:
+2. To import an individual component (for example a button), add the below to your Sass file:
+```CSS
+@import "node_modules/@govuk-frontend/button/button";
+```
+
+#### Optional: Resolve import paths
+
+If you wish to resolve the above `@import` paths in your build (in order to avoid prefixing paths with `node_modules`), you should add `node_modules` to your [Sass include paths](https://github.com/sass/node-sass#includepaths) (in Ruby, they should be added to [assets paths](http://guides.rubyonrails.org/asset_pipeline.html#search-paths)).
+
+For example, if your project uses Gulp, you would add the Sass include paths to your Gulp configuration file (for example `gulpfile.js`) with [gulp-sass](https://www.npmjs.com/package/gulp-sass). Below is an example:
+
+```JS
+gulp.task('sass', function () {
+  return gulp.src('./sass/**/*.scss')
+    .pipe(sass({
+      includePaths: 'node_modules'
+     }))
+    .pipe(gulp.dest('./css'));
+});
+
+```
+
+(If you compile Sass to CSS in your project, your build tasks will already include something similar to the above task - in that case, you will just need to include add `includePaths` to it.)
+
+After resolving the import paths you can import GOV.UK Frontend by using:
 ```CSS
 @import "@govuk-frontend/button/button";
 ```
 
+### Import images and icons
 
-To resolve your `@import` declarations you should add `node_modules` to your [Sass include paths](https://github.com/sass/node-sass#includepaths) or [assets paths](http://guides.rubyonrails.org/asset_pipeline.html#search-paths) in Ruby.
+In order to import GOV.UK Frontend images and icons to your project, you should configure your application to reference or copy the relevant GOV.UK Frontend assets.
 
-Below is a code sample you can add to your gulp configuration file using the gulp-sass package:
+1. Follow either [Recommended solution](#recommended-solution) or [Alternative solution](#alternative-solution).
+
+2. Set `$govuk-global-images` variable in your project Sass file to point to the images folder in your project. Make sure you do this in Sass before importing `@govuk-frontend` into your project - see [Import styles](#import-styles). (`$govuk-global-images` is defined by default in `/node_modules/frontend/global/settings/_paths.scss`.)
+
+#### Recommended solution:
+
+Make `/node_modules/@govuk-frontend/icons` available to your project by routing requests for your images folder there.
+
+For example, if your project uses [express.js](https://expressjs.com/), below is a code sample you could add to your configuration:
+
 ```JS
-.pipe(sass({ includePaths: 'node_modules/' }))
+app.use('/icons', express.static(path.join(__dirname, '/node_modules/@govuk-frontend/icons')))
 ```
 
-To import images, configure your application to reference or copy the icons assets.
+#### Alternative solution:
 
-Below is a code sample you can add to your [express.js](https://expressjs.com/) configuration:
-```JS
-app.use('/public', express.static(path.join(__dirname, '/node_modules/@govuk-frontend/icons')))
-```
+Manually copy the images from `/node_modules/@govuk-frontend/icons` into a public facing directory in your project. Ideally copying the files to your project should be an automated task or part of your build pipeline to ensure that the GOV.UK Frontend images and icons stay up-to-date.
 
 ## Usage
 
@@ -104,7 +139,9 @@ Add the CSS and JavaScript code to your HTML template:
 <!DOCTYPE html>
   <head>
     <title>Example</title>
-    <link rel="stylesheet" href="assets/govuk-frontend-[latest version].min.css">
+    <!--[if !IE 8]><!-->
+      <link rel="stylesheet" href="assets/govuk-frontend-[latest version].min.css">
+    <!--<![endif]-->
     <!--[if IE 8]>
       <link rel="stylesheet" href="assets/govuk-frontend-oldie-[latest version].min.css">
     <![endif]-->
@@ -117,6 +154,26 @@ Add the CSS and JavaScript code to your HTML template:
 </html>
 ```
 
+## Supporting Internet Explorer 8
+
+### Option 1: Add GOV.UK Frontend in your Sass build
+
+In order to support Internet Explorer 8, you should build a version of your application's stylesheet that targets IE8, and include it using conditional comments as shown in [Include assets](#include-assets).
+
+For instance, if your application Sass lives in `app.scss`, you might create an `app-ie8.scss` with the following:
+
+```SCSS
+// Target IE8
+$govuk-is-ie: true;
+$govuk-ie-version: 8;
+
+@import "app";
+```
+
+### Option 2: Use the CSS files provided in `dist`
+
+If you're not including GOV.UK Frontend styles in your Sass build but are using the generated files from `dist`, you should include both  `govuk-frontend.min.css` and `govuk-frontend-old-ie.min.css` in your project layout using conditional comments. The latter stylesheet adds support for IE8. See [Include assets](#include-assets) for an example.
+
 ## Usage
 
 Copy and paste code from the examples in the [GOV.UK Design System](https://govuk-design-system-production.cloudapps.digital/) to use GOV.UK Frontend in your service.
@@ -124,8 +181,8 @@ Copy and paste code from the examples in the [GOV.UK Design System](https://govu
 
 ## Licence
 
-This project is licensed under the [MIT License](LICENSE.txt).
+Unless stated otherwise, the codebase is released under the MIT License. This covers both the codebase and any sample code in the documentation. The documentation is &copy; Crown copyright and available under the terms of the Open Government 3.0 licence.
 
-## Contribution
+## Contribution guidelines
 
 If you want to help us build GOV.UK Frontend, view our [contribution guidelines](CONTRIBUTING.md).
--- a/app/views/partials/code.njk
+++ b/app/views/partials/code.njk
@@ -3,6 +3,8 @@
 <pre><code class="govuk-!-f-16">{{- componentHtml | e -}}</code></pre>
 
 <h4 class="govuk-heading-s">Macro</h4>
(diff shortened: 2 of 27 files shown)


joelanman:
Could we support old Nunjucks instead? Do we really need the whitespacing features, or other aspects?

nickcolley:
@joelanman likewise for the Node.js version, 8.x.x is very new

app/views/partials/code.njk
```diff
@@ -3,6 +3,12 @@
 <pre><code class="govuk-!-f-16">{{- componentHtml | e -}}</code></pre>
 
 <h4 class="govuk-heading-s">Macro</h4>
-<pre class="govuk-!-mb-r0 govuk-!-f-16"><code>{% raw %}{{ {% endraw %}{{ componentName | componentNameToMacroName }}(
-  {{- item.data | dump(2) -}}
-  ){% raw %} }}{% endraw %}</code></pre>
+<pre class="govuk-!-mb-r0 govuk-!-f-16">
```
36degrees: I think the new line here has changed the way the markdown is being generated, as all of the READMEs now include this div, rather than using the usual markdown indentation to signify a code block.

app/views/partials/code.njk
```diff
@@ -3,6 +3,12 @@
...
+  <code>
+    {% raw %}{% from '{% endraw %}{{ componentName }}/macro.njk' import {{ componentName | componentNameToMacroName }} {% raw %}%}{% endraw %}
```
36degrees: Is there a reason the first {% raw %} wraps `from '` as well as the opening tag?

README.md
```diff
@@ -9,9 +9,11 @@ GOV.UK Frontend ·
 
 GOV.UK Frontend contains the code you need to start building a user interface for government platforms and services.
 
+The guidance and demos of GOV.UK Frontend content live in the [GOV.UK Design System](https://govuk-design-system-production.cloudapps.digital/).
+
 ## Before using GOV.UK Frontend
 
-You should start by familiarising yourself with the [GOV.UK Design System](https://govuk-design-system-production.cloudapps.digital/), which contains guidance and examples of components, styles and patterns you can use to design  government platforms and services.
+You should start by familiarising yourself with the [GOV.UK Design System](https://govuk-design-system-production.cloudapps.digital/), which contains the guidance and examples of components, styles and patterns of the code that lives in GOV.UK Frontend.
```
36degrees: I don't think this is quite right - specifically 'patterns of the code that lives in GOV.UK Frontend'.

README.md
```diff
@@ -57,24 +57,36 @@ After you have installed GOV.UK Frontend the `@govuk-frontend` package will appe
...
+
+In order to resolve the above `@import` paths(s) in your build, you should add `node_modules` to your [Sass include paths](https://github.com/sass/node-sass#includepaths) (in Ruby, they should be added to [assets paths](http://guides.rubyonrails.org/asset_pipeline.html#search-paths)).
 
-To resolve your `@import` declarations you should add `node_modules` to your [Sass include paths](https://github.com/sass/node-sass#includepaths) or [assets paths](http://guides.rubyonrails.org/asset_pipeline.html#search-paths) in Ruby.
+For example, if your project uses Gulp, you would add the Sass include paths to your Gulp configuration file (for example `gulpfile.js`) with [gulp-sass](https://www.npmjs.com/package/gulp-sass). Below is an example:
 
-Below is a code sample you can add to your gulp configuration file using the gulp-sass package:
 ```JS
-.pipe(sass({ includePaths: 'node_modules/' }))
+gulp.task('sass', function () {
```
36degrees: Do we need to be even clearer that they will likely already have something like this in their gulp file already, and that the `includePaths` line is the only bit that's new?

README.md
```diff
@@ -89,11 +89,17 @@ gulp.task('sass', function () {
 
 ```
 
-To import images, configure your application to reference or copy the icons assets.
+### Import images and icons
 
-Below is a code sample you can add to your [express.js](https://expressjs.com/) configuration:
+In order to import GOV.UK Frontend images and icons to your project, you should configure your application to reference or copy the relevant GOV.UK Frontend assets.
+
+You can do this with one of the following:
```
36degrees: This isn't quite right – option 1 modifies the CSS to point to a different folder, but doesn't help with actually getting the images into a public facing directory. The use of 'one of the following' suggests that they should do either 1 _or_ 2 but it's not that simple.

README.md
```diff
@@ -89,11 +89,17 @@ gulp.task('sass', function () {
...
+
+1. Set `$global-images` variable in `/node_modules/frontend/global/settings` to point to your images folder.
```
36degrees: 1. This is pointing to the settings directory but we probably want to be referencing the `_paths.scss` file within that directory.
2. This makes it sound like they should _change the file in mode_modules_ but they can't and shouldn't do that – we want them to set the `$global-images` in their own SCSS file, _before_ they import Frontend.
3. That variable should probably be prefixed… 🤔

README.md
```diff
@@ -124,7 +124,9 @@ Add the CSS and JavaScript code to your HTML template:
 <!DOCTYPE html>
   <head>
     <title>Example</title>
-    <link rel="stylesheet" href="assets/govuk-frontend-[latest version].min.css">
+    <!--[if !IE 8]><!-->
```
36degrees: Good catch 👏

README.md
```diff
@@ -137,6 +139,14 @@ Add the CSS and JavaScript code to your HTML template:
 </html>
 ```
 
+### What is different in the `old-ie` stylesheet?
+
+The difference between `govuk-frontend.min.css` and `govuk-frontend-old-ie.min.css`
```
36degrees: Possibly an oversimplification – as well as rasterising media queries we also have IE8 specific fixes throughout the codebase, and do things like convert rgba to rgb.

This also only helps those using the `dist` versions – we also need to tell the user how to generate both versions from their own app stylesheet.

README.md
```diff
@@ -28,6 +28,10 @@ To use GOV.UK Frontend with NPM you must:
 
 1. Install the long-term support (LTS) version of [Node.js](https://nodejs.org/en/), which includes NPM.
 2. Create a [package.json file](https://docs.npmjs.com/files/package.json) if you don’t already have one. You can create a default `package.json` file by running `npm init` from the root of your application.
+3. If you want to use the GOV.UK Frontend Nunjucks macros, install Nunjucks - the minimum version required is 3.0.0.
```
36degrees: Is this definitely true?

app/views/partials/code.njk
hannalaakso: Thanks for catching 👍

README.md
hannalaakso: Sure. @amyhupe could you please review the wording. I was trying to make the relationship of FE and DS more explicit.

README.md
```diff
@@ -55,31 +61,49 @@ After you have installed GOV.UK Frontend the `@govuk-frontend` package will appe
...
-To import an individual component (for example, a button), add the line:
+2. To import an individual component (for example a button), add the below to your Sass file:
 ```CSS
 @import "@govuk-frontend/button/button";
 ```
 
+You should place the above code before your other Sass rules, or Sass imports, if you want to override
+GOV.UK Frontend with your own styles.
+
+In order to resolve the above `@import` paths(s) in your build, you should add `node_modules` to your [Sass include paths](https://github.com/sass/node-sass#includepaths) (in Ruby, they should be added to [assets paths](http://guides.rubyonrails.org/asset_pipeline.html#search-paths)).
```
nickcolley: Could we do this the other way around.

By default tell people to import 'node_modules/@govuk-frontend/button/button' and give them the option to remove the initial node_modules with path.

We saw many people copy pasting without reading so this could be a good way to make this foolproof.

README.md
nickcolley: I'm not sure we need to go into detail, all people really need to know is that this file contains additional CSS for IE8 support.

src/breadcrumbs/README.md
```diff
@@ -182,6 +190,8 @@ Find out when to use the Breadcrumbs component in your service in the [GOV.UK De
 
 #### Macro
 
+    from {% 'breadcrumbs/macro.njk' import govukBreadcrumbs %}
```
nickcolley: Is this syntax correct?

hannalaakso: No it wasn't, fixed now 👍

README.md
36degrees: Yep, I think I'd agree, but I think this is slightly misleading in that it suggests that the _only_ difference is media query rasterising.

README.md
hannalaakso: Please review again, I've added some more info to this section.

README.md
```diff
@@ -55,32 +61,57 @@ After you have installed GOV.UK Frontend the `@govuk-frontend` package will appe
...
 @import "@govuk-frontend/all/all";
 ```
 
-To import an individual component (for example, a button), add the line:
+2. To import an individual component (for example a button), add the below to your Sass file:
 ```CSS
 @import "@govuk-frontend/button/button";
 ```
 
+You should place the above code before your other Sass rules, or Sass imports, if you want to override
```
alex-ju: Wouldn't it be better to have this paragraph before the lines of code, after L66?

README.md
```diff
@@ -55,32 +61,57 @@ After you have installed GOV.UK Frontend the `@govuk-frontend` package will appe
...
+that case, you will just need to include add `includePaths` to it.
+
+### Import images and icons
+
+In order to import GOV.UK Frontend images and icons to your project, you should configure your application to reference or copy the relevant GOV.UK Frontend assets.
 
-Below is a code sample you can add to your [express.js](https://expressjs.com/) configuration:
+Recommended solution:
+
+Make `/node_modules/@govuk-frontend/icons` available to your project by routing requests for your images folder there. For example, if your project uses [express.js](https://expressjs.com/), below is a code sample you could add to your configuration:
```
alex-ju: Could we have the example on a new line, as we did on the _Import styles_ section?

README.md
```diff
@@ -55,32 +61,57 @@ After you have installed GOV.UK Frontend the `@govuk-frontend` package will appe
...
-To import images, configure your application to reference or copy the icons assets.
+Note: If you compile Sass to CSS in your project, your build tasks will already include something similar to the above task - in
+that case, you will just need to include add `includePaths` to it.
+
+### Import images and icons
+
+In order to import GOV.UK Frontend images and icons to your project, you should configure your application to reference or copy the relevant GOV.UK Frontend assets.
 
-Below is a code sample you can add to your [express.js](https://expressjs.com/) configuration:
+Recommended solution:
```
alex-ju: Would be good to 'highlight' the option titles. Either by using an h4 or bold.

README.md
```diff
@@ -55,32 +61,57 @@ After you have installed GOV.UK Frontend the `@govuk-frontend` package will appe
...
+Make `/node_modules/@govuk-frontend/icons` available to your project by routing requests for your images folder there. For example, if your project uses [express.js](https://expressjs.com/), below is a code sample you could add to your configuration:
 ```JS
-app.use('/public', express.static(path.join(__dirname, '/node_modules/@govuk-frontend/icons')))
+app.use('/icons', express.static(path.join(__dirname, '/node_modules/@govuk-frontend/icons')))
 ```
+Alternative solution:
+
+1. Manually copy the images from `/node_modules/@govuk-frontend/icons` into a public facing directory in your project. Ideally copying the files to your project should be an automated task or part of your build pipeline to ensure that the GOV.UK Frontend images and icons stay up-to-date.
+
+2. Set `$global-images` variable in your project Sass file to point to the images folder in your project. Make sure you do this before you import `@govuk-frontend` into your project - see "Import styles" above. (`$global-images` is defined by default in `/node_modules/frontend/global/settings/_paths.scss`.)
```
alex-ju: This is true, regardless of using the recommended or the alternative solution. So we can have it as a note for both solutions. 'Import styles' should be a link.

README.md
alex-ju: I think just stating that IE8 need a different build for the CSS would be fair enough. Otherwise we'll end up updating these documented differences each time we update the CSS that targets IE8.

README.md
hannalaakso: This sounds like it might be a good approach ie. do `@import node-modules/@govuk-frontend/button` in scss , with the import paths route as optional. It makes the imports a bit more verbose but will allow people to get started quicker if this is what participants in user research struggled with. @alex-ju @36degrees @igloosi do you have any thoughts?

README.md
hannalaakso: I think @alex-ju is right, this makes the file more maintainable. I've simplified the explanation.

README.md
hannalaakso: Done!

README.md
hannalaakso: Done!

README.md
hannalaakso: Done!

README.md
hannalaakso: I've prefixed $global-images and sneaked it into this PR 😁 Shout if you think it needs its own PR.

README.md
hannalaakso: Done!

README.md
hannalaakso: Yes it is, see PR description: 

> Document minimum version of Nunjucks in macros: 3.0.0 (2.5.2 and earlier throw up an error due to whitespacing being handled differently)

Also see this: https://github.com/alphagov/govuk-frontend/pull/615#issuecomment-375262011

README.md
```diff
@@ -55,33 +61,59 @@ After you have installed GOV.UK Frontend the `@govuk-frontend` package will appe
...
-To resolve your `@import` declarations you should add `node_modules` to your [Sass include paths](https://github.com/sass/node-sass#includepaths) or [assets paths](http://guides.rubyonrails.org/asset_pipeline.html#search-paths) in Ruby.
+For example, if your project uses Gulp, you would add the Sass include paths to your Gulp configuration file (for example `gulpfile.js`) with [gulp-sass](https://www.npmjs.com/package/gulp-sass). Below is an example:
 
-Below is a code sample you can add to your gulp configuration file using the gulp-sass package:
 ```JS
-.pipe(sass({ includePaths: 'node_modules/' }))
+gulp.task('sass', function () {
+  return gulp.src('./sass/**/*.scss')
+    .pipe(sass({
+      includePaths: 'node_modules/@govuk-frontend'
```
nickcolley: I'm not sure this includePath will work since the imports above reference @govuk-frontend.

Non-blocking suggestion: We could add 'node_modules' prefix to the imports above, and change this section to 'Don't want to include node_modules? Do this'

README.md
amyhupe: How about:

See live examples of GOV.UK Frontend components, and guidance on when to use them in your service, in the [GOV.UK Design System](https://govuk-design-system-production.cloudapps.digital/).

README.md
hannalaakso: Thanks @nickcolley I've fixed this.

README.md
```diff
@@ -55,33 +61,66 @@ After you have installed GOV.UK Frontend the `@govuk-frontend` package will appe
...
+    .pipe(gulp.dest('./css'));
+});
+
+```
+
+(If you compile Sass to CSS in your project, your build tasks will already include something similar to the above task - in that case, you will just need to include add `includePaths` to it.)
+
+After resolving the import paths you can import components from `node_modules/@govuk-frontend` by using:
+```CSS
+@import "button/button";
```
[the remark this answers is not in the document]
nickcolley: Agreed, I remember being massively confused when joining GOV.UK, not knowing where govuk_frontend_toolkit imports came from.

README.md
hannalaakso: This works with the example below. I was undecided whether to include `@govuk-frontend` in `includePaths` or in the Sass import. But I've switched these around now, that does make more sense.

README.md
```diff
@@ -24,8 +26,12 @@ There are 2 ways to use GOV.UK Frontend:
 ## Requirements
 To use GOV.UK Frontend with NPM you must:
 
-1. Install the long-term support (LTS) version of [Node.js](https://nodejs.org/en/), which includes NPM.
+1. Install the long-term support (LTS) version of [Node.js](https://nodejs.org/en/), which includes NPM. The minimum version of Node required is 8.6.0. (We recommend using [`nvm`](https://github.com/creationix/nvm) for managing versions of Node.)
 2. Create a [package.json file](https://docs.npmjs.com/files/package.json) if you don’t already have one. You can create a default `package.json` file by running `npm init` from the root of your application.
+3. If you want to use the GOV.UK Frontend Nunjucks macros, install Nunjucks - the minimum version required is 3.0.0.
+
+`npm install nunjucks --save`
```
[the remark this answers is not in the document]
hannalaakso: Done!

README.md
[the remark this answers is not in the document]
hannalaakso: Yes done, see above for convo.

README.md
hannalaakso: Done!

README.md
```diff
@@ -117,15 +158,20 @@ Add the CSS and JavaScript code to your HTML template:
 </html>
 ```
 
+### What is different in the `old-ie` stylesheet?
```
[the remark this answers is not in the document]
hannalaakso: Thanks @36degrees, a good shout, I've amended this accordingly.

Issue #514, opened 2018-02-08
Make it clear that the "html" attribute can be unsafe


paroxp:
## What

We had a discussion on a design system slack channel about the documentation not to encourage the use of `html` parameter in macros. We concluded, we shouldn't take away that functionality, but make it clear it's unsafe to use, causing people to rethink the strategy or apply necessary fixes before hand.

## Why

It's more efficient to copy and paste a block of text without realising what it's doing. It could easily bring some XSS vulnerability into the application by accident, which we'd all like to avoid.

## Proposal

Changing the name of the `html` attribute is a good start. This will need to be followed by some documentation changes on the Design System.

### Before

```nunjucks
{% from "panel/macro.njk" import govukPanel %}

{{ govukPanel({
  "titleText": "Application complete",
  "html": "Your reference number<br><strong>HDJ2123F</strong>"
})
}}
```

### After

```nunjucks
{% from "panel/macro.njk" import govukPanel %}

{{ govukPanel({
  "titleText": "Application complete",
  "unsafeHTML": "Your reference number<br><strong>HDJ2123F</strong>"
})
}}
```

## Connection

alphagov/govuk-design-system#175

edwardhorsford:
This suggested wording feels *scary*. I appreciate that's the point, but wonder if it's too strong. Used correctly, presumably it isn't unsafe.

If there aren't alternatives to providing html (such as the example given), do we really want such a scary key name?

nickcolley:
Another good example of this is React's [`dangerouslySetInnerHtml`](https://reactjs.org/docs/dom-elements.html#dangerouslysetinnerhtml) API.

36degrees:
Valid concerns – we need to balance the two carefully. FWIW, we have spent some time trying to address this before, and our solution was to mimick the javascript API by consistently using `text` and `html`. But I agree that it could definitely be clearer.

paroxp:
By all means, this only is a quick proposal. I imagine it will have to go through some user research before it's in stone.

paroxp:
How does `unprotectedHTML` sound? :D

/cc @edwardhorsford

nickcolley:
I've put together some examples of different ways to pass html to macros and the outputs:
https://nunjucks-passing-html-examples.glitch.me/

nickcolley:
Related: what impact does this have with autoescaping turned off? https://github.com/alphagov/govuk-frontend/blob/master/app/app.js#L27

joelanman:
[...]

PR #616, opened 2018-03-22
WIP: Add checkbox conditional reveal

nickcolley:


nickcolley:
This is blocked on modules, since polyfills overlap components and we'll have a massive amount of duplication between components. See https://github.com/alphagov/govuk-frontend/pull/619

src/checkboxes/template.njk
```diff
@@ -20,6 +30,11 @@
         for: id
       }) | indent(4) | trim }}
     </div>
+    {% if item.conditional %}
+      <div class="govuk-checkboxes__conditional" id="{{ conditionalId }}">
+        {{ item.conditional.html | safe }}
```
nickcolley: This feels like a big security risk considering we expect people to put html in here that contains form inputs which often replay user input...

nickcolley: see https://github.com/alphagov/govuk-frontend/issues/514

igloosi:
there are legacy -c prefixes in revealed labels
but it work in IE9

screen shot 2018-04-19 at 15 31 27
screen shot 2018-04-19 at 15 30 39

nickcolley:
@igloosi I thought I did unstyled examples and forgot about that, thanks for the spot. 👍

src/checkboxes/template.njk
```diff
@@ -1,17 +1,27 @@
 {% from "fieldset/macro.njk" import govukFieldset %}
 {% from "label/macro.njk" import govukLabel %}
 
+{% set isCondtional = false %}
```
alex-ju: typo: replace `isCondtional ` with `isConditional` in file

src/radios/radios.yaml
```diff
@@ -1,3 +1,13 @@
+accessibilityCriteria: |
+  ## Condtional reveals
```
alex-ju: Typo `conditional`

src/radios/template.njk
```diff
@@ -1,17 +1,27 @@
 {% from "fieldset/macro.njk" import govukFieldset %}
 {% from "label/macro.njk" import govukLabel %}
 
+{% set isCondtional = false %}
```
alex-ju: typo: replace `isCondtional ` with `isConditional` in file

src/checkboxes/checkboxes.test.js
```diff
@@ -0,0 +1,178 @@
...
+    it('with visible content', async () => {
+      await page.setJavaScriptEnabled(false)
+
+      await goToAndGetComponent('checkboxes', 'with-conditional')
+
+      const isVisible = await waitForVisibleSelector('.govuk-checkboxes__conditional')
+      expect(isVisible).toBeTruthy()
+    })
+  })
+  describe('must be hidden if JavaScript is avaliable and is collapsed', () => {
```
nickcolley: Missed this

src/checkboxes/checkboxes.test.js
```diff
@@ -0,0 +1,178 @@
...
+    it('checked input conditonal visible', async () => {
+      const $ = await goToAndGetComponent('checkboxes', 'with-conditional-checked')
+      const $component = $('.govuk-checkboxes')
+      const $checkedInput = $component.find('.govuk-checkboxes__input:checked')
+      const inputAriaControls = $checkedInput.attr('aria-controls')
+
+      const secondInputIsVisible = await waitForVisibleSelector(`[id="${inputAriaControls}"]`)
+      expect(secondInputIsVisible).toBeTruthy()
+    })
+    it('first input conditional hidden', async () => {
```
nickcolley: These test names could be better

src/checkboxes/checkboxes.test.js
```diff
@@ -0,0 +1,178 @@
...
+      const $lastInput = $component.find('.govuk-checkboxes__item:last-child .govuk-checkboxes__input')
+      const lastInputAriaControls = $lastInput.attr('aria-controls')
+
+      const lastInputIsHidden = await waitForHiddenSelector(`[id="${lastInputAriaControls}"]`)
+      expect(lastInputIsHidden).toBeTruthy()
+    })
+  })
+  describe('must indicate that there is collapsed content to interact with', () => {
+    it('with associations between input and collapsed area', async () => {
+      const $ = await goToAndGetComponent('checkboxes', 'with-conditional-checked')
```
nickcolley: Check if can reduce duplication with earlier tests.

src/radios/radios.js
```diff
@@ -0,0 +1,57 @@
+import '../globals/polyfills/Function/prototype/bind'
+import '../globals/polyfills/Event' // addEventListener and event.target normaliziation
+import { nodeListForEach } from '../globals/common'
+
+function Radios ($module) {
+  this.$module = $module
+  this.$inputs = $module.querySelectorAll('input[type="radio"]')
+}
+
+Radios.prototype.init = function () {
```
nickcolley: Keep consistent order with checkboxes

src/radios/radios.js
```diff
@@ -0,0 +1,57 @@
+import '../globals/polyfills/Function/prototype/bind'
+import '../globals/polyfills/Event' // addEventListener and event.target normaliziation
+import { nodeListForEach } from '../globals/common'
+
+function Radios ($module) {
+  this.$module = $module
+  this.$inputs = $module.querySelectorAll('input[type="radio"]')
```
nickcolley: Keep consistent with checkboxes

src/radios/radios.js
```diff
@@ -0,0 +1,57 @@
...
+  $input.setAttribute('aria-expanded', inputIsChecked)
+
+  var $content = document.querySelector('#' + $input.getAttribute('aria-controls'))
+  $content.setAttribute('aria-hidden', !inputIsChecked)
+}
+
+Radios.prototype.handleClick = function (event) {
+  nodeListForEach(this.$inputs, function ($input) {
+    // If a radio with aria-controls, handle click
+    var isCheckbox = $input.getAttribute('type') === 'radio'
```
nickcolley: isRadio

src/radios/radios.test.js
```diff
@@ -0,0 +1,178 @@
...
+    it('with visible content', async () => {
+      await page.setJavaScriptEnabled(false)
+
+      await goToAndGetComponent('radios', 'with-conditional')
+
+      const isVisible = await waitForVisibleSelector('.govuk-radios__conditional')
+      expect(isVisible).toBeTruthy()
+    })
+  })
+  describe('must be hidden if JavaScript is avaliable and is collapsed', () => {
```
nickcolley: Same issue as checkboxes

nickcolley:
@alex-ju just need to clean the tests up, and then this'll be ready to go.

igloosi:
TODO:
changelog

src/checkboxes/checkboxes.test.js
```diff
@@ -0,0 +1,178 @@
...
+  browser = global.__BROWSER__
+  page = await browser.newPage()
+})
+
+afterEach(async () => {
+  await page.close()
+})
+
+describe('Checkboxes', () => {
+  describe('must be visible as static content if JavaScript is unavaliable or fails', () => {
```
36degrees: It'd be good to nail these describe and it blocks – it's unclear to me reading this _what we're trying to test_ and what is _context_.

For example 'with no aria attributes' and 'with visible content' as currently written could easily be interpreted as context ('when the thing does not have aria attributes', 'when the thing has visible content').

I'd suggest this would be better as something _like_:

```
Describe checkboxes with conditional reveals
  Describe when Javascript is unavailable or fails
    It does not have any ARIA attributes
    It makes the conditional content visible
```

The benefit of this structure is that the nesting still 'works', but each line also stands on its own – `describe` blocks read as context and `it` blocks read as assertions.

src/checkboxes/checkboxes.test.js
```diff
@@ -0,0 +1,178 @@
...
+    it('with no aria attributes', async () => {
+      await page.setJavaScriptEnabled(false)
+
+      const $ = await goToAndGetComponent('checkboxes', 'with-conditional')
+      const $component = $('.govuk-checkboxes')
+
+      const hasAriaHidden = $component.find('[aria-hidden]').length
```
36degrees: The checkbox itself would never have `aria-hidden` right? That attribute would be on the conditional content? Should this be a separate test?

src/checkboxes/checkboxes.test.js
36degrees: Also, typo on unavaliable (appears a few times)

nickcolley: yeah like I said above, I'm cleaning these up after the review with alex and jani

src/checkboxes/checkboxes.test.js
nickcolley: `.find` will look for any child element with aria-hidden so this should catch any issues.

36degrees: Makes sense, though not sure that's clear. Can we be more specific and check the elements that we put the aria attributes on?

nickcolley: Sorted

PR #617, opened 2018-03-23
Make Sass imports less broad

igloosi:
This PR replaces `globals/common` import in each component with only the necessary partials.
There is already a test that checks if individual component Sass files compile.

Note: i've placed typography partials at the end of imports to make it a bit easier to do the work of making typography imports conditional.

https://trello.com/c/unO51Rno/830-make-sass-imports-dependencies-less-broad


nickcolley:
Seems fine to me, the main problem is that we were importing files that output CSS.

So could solve this by chunking the imports better too.

PR #618, opened 2018-03-23
Compile individual component Sass files

igloosi:
This PR 
- installs `node-sass-middleware` and `postcss-middleware` to compile component's scss files
- configure middleware in app.js and pass component name to `component-preview` view
- modify `component-preview` view to not inherit from `layout ` view and load in component-specific scss file
- move app iframe styles to app.scss file
- modify gulp dev task so that it no longer compiles scss to css 
- create a single place where css tranformation plugins are defined and export it for use in app.js and gulp tasks
- modify `component-preview` view so it no longer contains the wrapper (component is full width) 
- update iframe styles so that the component fits the frame

This is to aid https://trello.com/c/7ROgYy2X/828-ensure-sass-imports-match-npm-dependencies
and  https://trello.com/c/unO51Rno/830-make-sass-imports-dependencies-less-broad, so we can check if components scss file imports all required dependencies


igloosi:
hmmm
```
FAIL src/button/button.test.js (16.518s)
  ● /components/button › /components/button/link › triggers the click event when the space key is pressed
    Timeout - Async callback was not invoked within the 5000ms timeout specified by jest.setTimeout.
      
      at node_modules/jest-jasmine2/build/queue_runner.js:72:21
```

36degrees:
Does this mean the styles in the review app are no longer going through postcss?

igloosi:
they do now. autoprefix and postcsspseudo

app/app.js
```diff
@@ -34,6 +39,31 @@ env.addFilter('componentNameToMacroName', helperFunctions.componentNameToMacroNa
...
+app.use(sassMiddleware({
+  src: root,
+  response: false,
+  dest: destPath,
+  outputStyle: 'extended'
+}))
+
+// get the above request and tranform css
+app.use(postcssMiddleware({
+  plugins: [
```
nickcolley: I think we should have a single source of truth for how we build our files, this would be really easy to get out of sync with what we ship to users.

igloosi: created a single place for the plugins

igloosi:
I won't be able to make this work for IE8, e.g. compile <component>.scss with ie8 settings
```
$mq-responsive: false;
$govuk-is-ie: true;
$govuk-ie-version: 8;
```

Issue #612, opened 2018-03-19
Tests run on the same port as the review application.


nickcolley:
Our review application runs on port 3000, this is used in our integration tests which means you can't run both at the same time.

PR #620, opened 2018-03-27
Serve app from random port

igloosi:
Fixes https://github.com/alphagov/govuk-frontend/issues/612 and enables running the Design System and Frontend at the same time


igloosi:
tests will fail. there is a quicker way

Issue #614, opened 2018-03-20
Set up Nunjucks development environment with more meaningful errors


igloosi:
Nunjucks has a couple of undocumented features for developers to make it easier to debug issues
We should look into implementing some.

Examples of available options
https://github.com/mozilla/nunjucks/blob/6f3e4a36a71cfd59ddc8c1fc5dcd77b8c24d83f3/nunjucks/src/environment.js#L40
and 
https://github.com/mozilla/nunjucks/issues/412#issuecomment-97839292

Trello triage ticket: https://trello.com/c/3qAck5jL/841-configure-nunjucks-development-environment

nickcolley:
Since this is related to our review application I'm going to close this, we can still track this in trello.

PR #621, opened 2018-03-27
Configure nunjucks and express to give more meaningful errors

igloosi:
Addresses #614 

Before
screen shot 2018-03-27 at 15 53 35

After
screen shot 2018-03-27 at 15 55 19

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,10 @@ Internal:
 - Update check script for new components and tweak docs
   (PR [#589](https://github.com/alphagov/govuk-frontend/pull/589))
 
+- Expose more meaningful errors in dev
+  (PR [#621](https://github.com/alphagov/govuk-frontend/pull/621))
+  
+
 ## 0.0.26-alpha (Breaking release)
 
 Breaking changes:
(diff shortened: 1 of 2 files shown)


app/app.js
```diff
@@ -150,4 +152,13 @@ app.get('/robots.txt', function (req, res) {
   res.send('User-agent: *\nDisallow: /')
 })
 
+// // expose the error in combination with nunjucks env options
```
nickcolley: Nit: // //

Also, can you explain in the comment what this is doing, the setTimeout I'm not sure why it's required

CHANGELOG.md
```diff
@@ -19,6 +19,10 @@ Internal:
 - Update check script for new components and tweak docs
   (PR [#589](https://github.com/alphagov/govuk-frontend/pull/589))
 
+- Expose more meaniningful errors in dev
```
nickcolley: meaniningful -> meaningful

igloosi:
this won't work as it will throw endefined on cases where we have or, for example
`{{ params.html | safe if params.html else params.text }}`
and the macro calls
```
     {{ govukLabel({
        "text": "National Insurance number",
      }) }}
```
it will threat html as undefined

PR #622, opened 2018-03-27
Listen for development server on different port for tests

nickcolley:
Required some refactoring of how we listen since the tests were not actually,
adding a new listener.

Also fixes issues with tests not closing properly by allowing tests to turn off
nunjucks watch mode.

Used https://github.com/mafintosh/why-is-node-running to identify and fix this issue.

Use https://github.com/alphagov/govuk-frontend/pull/622/files?w=1 to see changes without whitespace updates, since I've indented a lot of code.

Fixes https://github.com/alphagov/govuk-frontend/issues/612


igloosi:
do you also need to change the heroku script in package.json?

igloosi:
could we have  `http://localhost:8888/` some where central, as we're repeating it in multiple places?

nickcolley:
I don't like using globals in Node.js projects, so tried that but it felt over-the-top.

Happy to put it back in though, I don't expect us to want to change this port programatically from now on though.

igloosi:
don't forget the changelog as well

PR #623, opened 2018-03-28
Fix date-input default example

igloosi:
BEFORE
screen shot 2018-03-28 at 10 33 56

AFTER
screen shot 2018-03-28 at 10 34 02

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,9 @@ Internal:
 - Update check script for new components and tweak docs
   (PR [#589](https://github.com/alphagov/govuk-frontend/pull/589))
 
+- Fix date-input default example
+(PR [#623](https://github.com/alphagov/govuk-frontend/pull/623))
+
 ## 0.0.26-alpha (Breaking release)
 
 Breaking changes:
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,11 @@ Breaking changes:
   within that block inline.
   (PR [#607](https://github.com/alphagov/govuk-frontend/pull/607))
 
+Fixes:
```
nickcolley: I think this is technically an internal change since we don't ship the yaml files.

src/date-input/template.test.js
```diff
@@ -205,7 +205,6 @@ describe('Date input', () => {
   it('passes through fieldset params without breaking', () => {
     const $ = render('date-input', examples['default'])
 
-    expect(htmlWithClassName($, '.govuk-c-error-message')).toMatchSnapshot()
```
36degrees: Should this have moved or been split out into a different test rather than just being removed?

nickcolley: Hmm, yeah, I'll do a PR.

nickcolley: See https://github.com/alphagov/govuk-frontend/pull/625

igloosi: sorry

nickcolley: S'all good, slipped my mind too 👍

PR #628, opened 2018-04-03
Add tests for mixins and functions

36degrees:
Add tests for the mixins and functions that include 'logic'.

### DIFF
--- a/src/globals/helpers/spacing.test.js
+++ b/src/globals/helpers/spacing.test.js
@@ -1,0 +1,265 @@
+/* eslint-env jest */
+
+const util = require('util')
+
+const configPaths = require('../../../config/paths.json')
+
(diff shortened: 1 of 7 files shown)


src/globals/helpers/spacing.test.js
```diff
@@ -0,0 +1,265 @@
...
+
+  it('throws an exception when passed anything other than a map', async () => {
+    const sass = `
+      ${sassBootstrap}
+
+      .foo {
+        @include govuk-responsive-spacing(14px, 'margin')
+      }`
+
+    expect.assertions(1)
```
nickcolley: May be able to do something like https://github.com/alphagov/govuk-frontend/blob/master/src/back-link/template.test.js#L20 to avoid this

PR #631, opened 2018-04-05
Simplify legacy IE logic

36degrees:
- Simplify the legacy IE logic now that we only need to support targeting IE8.
- Remove anything targeting IE6 or IE7.
- Remove the requirement for users to set `$mq-responsive: false;` in their IE8 stylesheet by setting this value based on `$govuk-is-ie8`.

Recommend reviewing commit by commit.

I have validated that this change does not affect the generated CSS by running `npm run build:dist` on both the master branch and this branch, copying the generated files from `dist/css` and comparing the files:

```
md5 **/*.css
MD5 (master/govuk-frontend-0.0.26-alpha.min.css) = 9515f244a0a79d1e7b307d0d696c3633
MD5 (master/govuk-frontend-old-ie-0.0.26-alpha.min.css) = 3dfa4089277139b33decb1af6512d126
MD5 (simplify-legacy-ie-logic/govuk-frontend-0.0.26-alpha.min.css) = 9515f244a0a79d1e7b307d0d696c3633
MD5 (simplify-legacy-ie-logic/govuk-frontend-old-ie-0.0.26-alpha.min.css) = 3dfa4089277139b33decb1af6512d126
```


alex-ju:
Should we update [legacy-ie.md](https://github.com/alphagov/govuk-frontend/blob/master/docs/legacy-ie.md) based on these changes?

PR #634, opened 2018-04-06
Make radio and checkbox labels inline blocks

### LINKED ISSUE #543: Radios labels are full width - should be the width of the label text?
Radio labels extend to the full width of the container. This means the clicks apparently in space on the righthand side activate radios.

This was not the case with Elements where labels only extended as far as the text.

36degrees:
Make radio and checkbox inline blocks so that the effective click area only extends to the label. Without this change, the labels extend to the full width of the container which means the clicks apparently in space on the right hand side activate the inputs.

## Before

before

## After

after

## Testing

Tested in:

- [X] Internet Explorer 8
- [X] Internet Explorer 9
- [X] Internet Explorer 10
- [X] Internet Explorer 11
- [X] Edge 16
- [X] Google Chrome 65 (Mac)
- [X] Mozilla Firefox 59 (Mac)
- [X] Safari 11 (Mac)
- [X] Internet Explorer, Windows Phone 8.1
- [X] Safari, iOS 9 (Simulated, iPhone 6S)
- [X] Safari, iOS 11.3 (Physical, iPhone X)
- [X] Android Google Chrome (Simulated, Samsung Galaxy S8)
- [X] Samsung Internet (Simulated, Samsung Galaxy S8)

Fixes #543

### DIFF
--- a/src/checkboxes/_checkboxes.scss
+++ b/src/checkboxes/_checkboxes.scss
@@ -44,7 +44,7 @@
   }
 
   .govuk-c-checkboxes__label {
-    display: block;
+    display: inline-block;
     padding: 8px $govuk-spacing-scale-3 $govuk-spacing-scale-1;
     cursor: pointer;
     // remove 300ms pause on mobile
--- a/src/radios/_radios.scss
+++ b/src/radios/_radios.scss
@@ -45,7 +45,7 @@
   }
 
   .govuk-c-radios__label {
-    display: block;
+    display: inline-block;
     padding: 8px $govuk-spacing-scale-3 $govuk-spacing-scale-1;
     cursor: pointer;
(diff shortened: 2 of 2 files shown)


joelanman:
I think on larger screen sizes its definitely a bug - you can trigger a radio button with a seemingly unrelated click. Also it's not how Elements works, so it's a change we'd need to do consciously.

On smaller screens it could possibly be ok, but I wonder if it's worth adding the complexity of it being different at different screen sizes for something that may or may not be helpful. I'd say no.

In user research the large majority of people try to click/tap the circle, which is why we made it really big.

nickcolley:
Ship it

36degrees:
@nickcolley if you're happy with this can you approve it?

Issue #636, opened 2018-04-09
Path to sass-mq is dependent on specific node_modules location


penx:
This line is causing us an issue:

https://github.com/alphagov/govuk-frontend/blob/ad2022ef044d2e67768199479e215f641e36cc5f/dist/components/globals/helpers/_media-queries.scss#L14

This assumes that `sass-mq` is installed in a `node_modules` folder relative to the root folder used by the Sass build system.

This isn't always the case, e.g.

1. If Sass is configured to resolve links relative to file location rather than parent project root
2. If the parent project uses a different version of sass-mq to govuk-frontend, npm won't flatten this dependency and the version needed by `govuk-frontend` actually lives in `./node_modules/govuk-frontend/node_modules/sass-mq`
 `govuk-frontend` 
3. if used in a monorepo (lerna, yarn workspaces) then node_modules are hoisted in to a parent directory.

I suspect this can be fixed by swapping it with `@import "sass-mq/mq"`, which will leave node to deal with the module resolution.

nickcolley:
Thanks for raising! Do you have an example of the build where this fails for us to reproduce? It may help us fix this quicker. 😄

penx:
failing that I could try create a simple example if that would help?

penx:
@nickcolley @igloosi I have created a failing example with instructions here https://github.com/penx/govuk-monorepo-example

penx:
Note that Webpack sass-loader [provides a mechanism to load dependencies from node_modules](https://github.com/webpack-contrib/sass-loader#imports):

```scss
@import "~sass-mq/mq";
```

However this doesn't seem to be a standard Sass convention, but personally I think it is preferred to importing from an assumed relative path to node_modules.

nickcolley:
@penx thanks so much for the failing example! 🙌

igloosi:
the other options would be to expose that path in config options

penx:
@igloosi that may work, but would require the parent project to know where the package was hoisted to (which could potentially vary) rather than defer this resolution to node.

Any reason for this instead of Eyeglass?

penx:
failing that we could have the govuk-frontend => JS variables go in to it's own project that wasn't a monorepo and do the conversion as part of its build, then have govuk-react depend on this interim package

nickcolley:
@penx (while we try and find other solutions) is it possible to avoid hoisting node_modules? Or avoiding yarn?

penx:
[...]

PR #637, opened 2018-04-09
Update SassMQ import path

igloosi:
In certain use cases as outlined in #636 the way we were importing SassMQ wasn't resolving correctly.
This change should fix that.
- Updated path in `src/globals/helpers/_media-queries.scss`
- `includePaths: 'node_modules'` added to gulp compilation task
- `includePaths: 'node_modules'` added to node-sass configuration in tests
- Changelog updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ Breaking changes:
   details.
   (PR [#631](https://github.com/alphagov/govuk-frontend/pull/631))
 - Rename captionSize table argument to captionClasses ([PR #643](https://github.com/alphagov/govuk-frontend/pull/643))
+- Update SassMQ import path (PR [#637](https://github.com/alphagov/govuk-frontend/pull/637))
 
 Fixes:
 - Link styles, as well as links within the  back-link, breadcrumbs, button,


src/globals/helpers/_media-queries.scss
```diff
@@ -11,6 +11,6 @@ $sass-mq-already-included: false !default;
   $mq-show-breakpoints: ();
 }
 
-@import "./node_modules/sass-mq/mq";
+@import "sass-mq/mq";
```
penx: the webpack sass-loader way of doing this would be `"~sass-mq/mq"` https://github.com/webpack-contrib/sass-loader#imports

nickcolley:
Added an example based on the reproduction @penx gave us. I'm not sure this PR is enough to fix it...

src/globals/helpers/_media-queries.scss
penx: Is there any reason not to use the `~` syntax here? I think the current implementation would mean that the parent project would need to manually ensure `sass-mq` was on the resolution path for sass files, wheras `~` should always work as long as the parent project used sass loader.

Or are you saying that compatibility with Sass build systems that don't support `~` is more important than manual setup?

Would be good to see how existing Sass libraries distributed via npm that have Sass sub-dependencies handle this.

36degrees:
In #615 we made adding `node_modules` optional in the documentation by making the default import relative to the project root. Unless I'm misunderstanding, this looks like it would rely on consumers having `node_modules` back in their Sass path definition again.

If this is the only way to do it that shouldn't be a blocker, but we'll need to make sure we update the documentation appropriately.

src/globals/helpers/_media-queries.scss
36degrees: Do you know what the support is for that syntax? We can make very few assumptions about the technologies involved our consumer's build pipelines, so unless that syntax is very well supported I would be wary of relying on it.

penx: I think the problem (as described at https://github.com/webpack-contrib/sass-loader#imports) is Sass doesn't have a default syntax for library imports.

The 3 options are:
1. `~`
2.`node_modules` or `./node_modules`
3.  mixing relative + lib paths (`@import "sass-mq/mq";`)

I think all 3 either infer an assumption or requirement from the parent project.

I'll take a look around and see if I can find examples in other Sass projects dealing with this.

igloosi:
yeah if they choose the option without needing to specify the path definition then it feels wrong in having to ask them to do it regardless for this to resolve.

- we can revert back to `@import "./node_modules/sass-mq/mq";` but wrap it in a variable so it can be modified for cases mentioned in the issues. Then there is no change for most users

src/globals/helpers/_media-queries.scss
penx: related, but not found an answer yet:

https://github.com/sass/node-sass/issues/1596
https://github.com/webpack-contrib/sass-loader/issues/466

igloosi: ~ is webpack specific, node-sass doesnt supoort that

penx: I made a couple comments on sass/node-sass#1596 and also found and commented on sass/sass/issues#2350

Is it ok to assume the parent project will use `node-sass` (given that it's distributed via npm)?

Looks like [eyeglass](https://github.com/sass-eyeglass/eyeglass) may be the recommended solution if so.

igloosi:
wrapping SassMQ import path to variable should solve 
https://github.com/alphagov/govuk-frontend/pull/637#issuecomment-380195066
as well as @penx's issue

igloosi:
won't fix, closing in favour of #657

PR #638, opened 2018-04-10
WIP Add description / definition list

dashouse:
Add basic styling for a description / definition list

### DIFF
--- a/src/globals/core/_lists.scss
+++ b/src/globals/core/_lists.scss
@@ -66,7 +66,7 @@
   }
 
   %govuk-list--description > dt {
-    @include govuk-font-bold-19
+    @include govuk-font-bold;
   }
 
   %govuk-list--description > dt {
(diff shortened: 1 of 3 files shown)


src/globals/core/_lists.scss
```diff
@@ -68,4 +68,25 @@
   .govuk-list--number {
     @extend %govuk-list--number;
   }
+
+  %govuk-list--description > dt {
+    @include govuk-font-bold-19
```
36degrees: We get the font size from `govuk-list`, so this can probably just be `govuk-font-bold`? Also, slight pedantry, but missing the semi-colon.

dashouse:
Thanks @36degrees - updated

dashouse:
Merged

hannalaakso:
NVDA reads out the number of items inside `dl` tags in a way that could be confusing to the user. This pattern is less complex than Check your answers but we should test it on NVDA just in case. Some more info in [backlog](https://github.com/alphagov/govuk-design-system-backlog/issues/36#issuecomment-380837671).

PR #644, opened 2018-04-11
Remove -c -o -h layer prefixes

igloosi:
In user research and in feedback from Private Beta partners we learned that users didn't really understand what -o, -c and -h prefixes stand for or why they're useful. 

They also introduced additional cognitive load when composing classes, having to remember which suffix the classname contains.

As a result of this, we're removing these prefixes from our codebase.

This PR:
- Removes relevant prefixes from `src/` files
- Removes references in documentation
- Updates internal app
- Updates changelog

Trello ticket: https://trello.com/c/vAitHnu7/863-2-remove-object-component-and-helper-layer-prefixes


hannalaakso:
`docs/component-api.md` and `docs/component.md` still refer to `govuk-c-button` so those refs should be changed before merging.

igloosi:
good stop. updated

PR #645, opened 2018-04-11
Revert "Add description / definition list"

hannalaakso:
This reverts the merge into master of https://github.com/alphagov/govuk-frontend/pull/638

Why? After some more discussions, it appears there's a need for a variant of the definition list from the above PR that reverses the bold styles. It makes sense to make definition list into a thing of its own that doesn't use `govuk-list`. We want to get this version of FE out so pulling this out of master and we'll investigate definition lists in more detail in due course.


36degrees:
Thanks @hannalaakso.

I think the changelog entry will need reverting too?

hannalaakso:
Thanks @36degrees, that's now done.

PR #619, opened 2018-03-26
Add Webpack as js bundler

alex-ju:
This PR:
- Setups up [Webpack](https://github.com/webpack/webpack) to build modular javascript using [UMD format](https://github.com/umdjs/umd)
- Rewrites existing scripts (`button.js`, `details.js` and a place for common functions called `common.js` for now) using ES6 modules; remove auto-initialisation;
- Updates JavaScript path in `app/views`; we're using `all.js` (which includes and initialises the JavaScripts for each component)
- Cleans Gulp pipes (re: JavaScript) and remove unused packages
- Update the build(preview app/packages/build) process to accommodate the Webpack build

Paired with @nickcolley on this one. Feel free to add commits on top of it.

[Trello card](https://trello.com/c/1xWsQvfm)

PR #646, opened 2018-04-11
Revert "Add Webpack as js bundler"

alex-ju:
Reverts alphagov/govuk-frontend#619


nickcolley:
👍 We're reverting this to do a release with the old JavaScript style

PR #652, opened 2018-04-12
Modules with rollup

nickcolley:
This work replaces an attempt to use Webpack before running into issues with IE8.

Main changes
- We now use ES6 modules to structure our code
- We now ship UMD bundles which work with module bundlers and in browsers
- We now have polyfills for 'bind' which uses an approach that mirrors the FT's Polyfill.io service
- We removed the redundant `gulp-standard` usage

- Reverted Webpack work: https://github.com/alphagov/govuk-frontend/pull/646
- Attempt at IE8 support: https://github.com/alphagov/govuk-frontend/tree/ie8-webpack-support

See testing done in https://github.com/alphagov/govuk-frontend/pull/646 as the main changes to the existing code has not changed.

Additional testing in IE8 has been done for this PR.

We intend to write up the findings from the Webpack work as documentation.

- https://trello.com/c/1xWsQvfm/846-build-modular-javascript
- https://trello.com/c/qhM3ClkX/894-3-make-sure-we-support-ie8-for-js


igloosi:
changelog!

igloosi:
this is much more understandable than webpack stuff. 
looks good to me.  anyone care to look

alex-ju:
Thanks for the review @igloosi. Rebased for the changelog.

PR #657, opened 2018-04-16
Vendor-in SassMQ functionality, write tests and remove external dependency

igloosi:
There are a number of scenarios  where assuming the sass-mq dependency lives in the same project breaks that have been outlined in #636.

As Sass does not allow dynamic imports we cannot provide a variable to override path to sass-mq stylesheet.

To fix this and avoid external dependencies of Frontend, we've moved the functionality into the project, wrote tests for its mixins and functions and remove dependency from package.json

### DIFF
--- a/src/globals/helpers/_media-queries.scss
+++ b/src/globals/helpers/_media-queries.scss
@@ -1,16 +1,318 @@
-// We use sass-mq module for media queries
+@charset "UTF-8";
+// Fixes an issue where Ruby locale is not set properly
+// See https://github.com/sass-mq/sass-mq/pull/10
 
-// This is a horrible, horrible hack to prevent the 'dev mode' CSS to display
-// the current breakpoint from being included multiple times. 
-// 
-// We can't use the `exports` mixin for this because import directives cannot be
-// used within control directives 😠
-$sass-mq-already-included: false !default;
+// This is a vendored version of SassMQ utility for composing media quesries
+// File is taken from https://github.com/sass-mq/sass-mq
+// Based on version 4.0.2
 
-@if $sass-mq-already-included {
-  $mq-show-breakpoints: ();
+@import "../tools/exports";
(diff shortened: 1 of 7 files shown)


nickcolley:
Not great to vendor but I imagine we won't really need to update this, so seems good to me.

Would be good to include a comment in the file that explains it's vendored, and what version it is.

igloosi:
added the comment about where its vendored from and version

src/globals/helpers/_media-queries.scss
```diff
@@ -1,8 +1,8 @@
 // We use sass-mq module for media queries
 
 // This is a horrible, horrible hack to prevent the 'dev mode' CSS to display
-// the current breakpoint from being included multiple times. 
-// 
+// the current breakpoint from being included multiple times.
+//
 // We can't use the `exports` mixin for this because import directives cannot be
```
nickcolley: Non-blocking but if we vendor this is this hack necessary any more? @36degrees

36degrees: No, not if we wrap the whole of sass-mq in an `exports`.

src/globals/helpers/_media-queries.scss
```diff
@@ -11,6 +11,6 @@ $sass-mq-already-included: false !default;
   $mq-show-breakpoints: ();
 }
 
-@import "./node_modules/sass-mq/mq";
+@import "../tools/sass-mq";
```
36degrees: It seems unnecessary to split media queries across these two layers… and because sass-mq 'outputs' CSS (even if it is just breakpoints) it would make more sense in here. Would suggest just moving sass-mq in here and using `exports` to simplify the hack as @nickcolley identified.

igloosi:
updated with comments by @nickcolley and @36degrees

PR #658, opened 2018-04-17
Opt-in global anchor and paragraph styles

igloosi:
To ease prototyping and new production instances we provided a way to include global styles for links and paragraphs, so that users do not have to add classes to every paragraph or link.

Inclusion is controlled by `$govuk-global-styles` variable in `globals/settings` folder.
 Documentation has been added to the main README.md file

If Elements or Template compatibility variables are set to true then we render fixes to those and also include global styles

Trello: https://trello.com/c/toKpf2vV/868-3-add-optional-styles-that-set-paragraph-and-link-styles

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,6 +138,23 @@ After resolving the import paths you can import GOV.UK Frontend by using:
 @import "@govuk-frontend/button/button";
 ```
 
+#### Global Styles
+
+[Global styles](src/globals/core/_global-styles.scss) are not included by default.
+
+This is to avoid the risk of these globals conflicting with any pre-existing globals, for example in GOV.UK Elements or GOV.UK Template.
+
+Hovever, we do include them in the [GOV.UK Prototype Kit](https://github.com/alphagov/govuk-prototype-kit-private-beta) to speed up prototyping.
+
+To include global styles, you can set `$govuk-global-styles` variable to `true`.
+```
+// application.scss
+
+$govuk-global-styles: true;
+
+@import "govuk-frontend/all/all";
+```
+
 ### Import JavaScript
 
 You need to import the GOV.UK Frontend scripts into the main JavaScript file in your project.
--- a/src/globals/core/_global-styles.scss
+++ b/src/globals/core/_global-styles.scss
@@ -1,0 +1,21 @@
+@import "../tools/exports";
+@import "links";
+@import "typography";
+@import "../settings/global-styles";
+
+@mixin govuk-global-styles {
+  a {
+    @extend %govuk-link;
+  }
+
+  p {
+    @extend %govuk-body-m;
+  }
+}
+
+@include govuk-exports("govuk-global-styles") {
+
+  @if $govuk-global-styles == true {
+    @include govuk-global-styles;
+  }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,17 @@ New features:
 - Focus Error Summary on window load
     (PR [#671](https://github.com/alphagov/govuk-frontend/pull/671))
 
+- Opt-in Global `<a>` and `<p>` styles
+  (PR [#658](https://github.com/alphagov/govuk-frontend/pull/658))
+
+  Global styles are not included by default.
+
+  This is to avoid the risk of these globals conflicting with any pre-existing globals, for example in GOV.UK Elements or GOV.UK Template.
+
+  Hovever, we do include them in the [GOV.UK Prototype Kit](https://github.com/alphagov/govuk-prototype-kit-private-beta) to speed up prototyping.
+
+  To include global styles, you can set `$govuk-global-styles` variable to `true`.
+
 Internal:
 
 - Update publishing docs (PR [#651](https://github.com/alphagov/govuk-frontend/pull/651))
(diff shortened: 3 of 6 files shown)


docs/global-styles.md
```diff
@@ -0,0 +1,20 @@
+# Global Styles
+
+In GOV.UK Frontend we want to provide a convenient way to have correct styles applied to your links (`<a>`) and paragraphs (`<p>`) without having to set required classes for each instance of an element.
+
+We have chosen 'Opt-in' as the default behaviour.
```
alex-ju: Shall we say 'Opted-in'. For me 'Opt-in' sound like I have the option to choose if I want it, which is the opposite of what we're doing by default. We can ask @amyhupe to help with wording.

src/all/_all.scss
```diff
@@ -1,5 +1,9 @@
+// optional setting of global styles for <a> and <p>
+$govuk-global-styles: true !default;
+
 // Settings
 @import "../globals/common";
+@import "../globals/optional/global-styles";
```
alex-ju: Given the context `optional/` may be the right name.
My view of this was that we'd have a `/core` in which we place the default style that you can choose not to use.

igloosi: was considering putting it in core first. i have no issue with it being in core tbh

hannalaakso: I quite like `optional/` as this clearly differentiates those styles from the rest. But not a blocker.

docs/global-styles.md
```diff
@@ -0,0 +1,20 @@
...
+You can set the `$govuk-global-styles` variable to `false` in your application Sass file.
+```
+// application.scss
+
+$govuk-global-styles: false;
+
+@import "govuk-frontend/all/all";
+```
+If you choose this option, you will need to either use classes on links (`<a>`) and paragraphs (`<p>`) or have a different way to apply correct styles.
+
```
hannalaakso: It would be good to document here, or link from here to another doc, what to look out for if choosing to set `$govuk-global-styles` to `false` and adding our link classes selectively, eg. `govuk-link` class must be added along with `govuk-link--muted`.

Instead of "use classes", maybe "add GOV.UK Frontend classes"?

docs/global-styles.md
```diff
@@ -0,0 +1,20 @@
+# Global Styles
```
nickcolley: I think this documentation is most likely to be useful if you're using the legacy stuff, with this in mind should we align this work with the compatibility mode stuff?

User scenario:
1. I have GOV.UK Template, and just want a build that works with that

igloosi: in that case you don't need global styles then, right? As they're coming from template?

igloosi: all of the *compatibillity vars are set to true in Frontend 🤔

src/globals/core/_global-styles.scss
```diff
@@ -0,0 +1,32 @@
...
+  //
+  // 2. By default $govuk-compatibility-govuktemplate is set to true.
+  // If so then include global styles, to combat govuk_template global styles
+  //
+  // 3. If global-styles is set to false we still want to combat govuk_template global styles.
+  // In this case we include global styles
+  //
+  // 4. If both $govuk-global-styles and $govuk-compatibility-govuktemplate are set to false
+  // then don't include global styles
+  @if global-variable-exists("govuk-global-styles") == false or $govuk-global-styles == true or $govuk-compatibility-govuktemplate == true {
```
nickcolley: should this be `$govuk-compatibility-govuktemplate == false`?

igloosi:
yes, let's hash out the scenarios together

igloosi:
updated PR with changes to logic and basic link styles (to defend against focus)

README.md
```diff
@@ -138,6 +138,27 @@ After resolving the import paths you can import GOV.UK Frontend by using:
...
+
+You can set the `$govuk-global-styles` variable to `false` in your application Sass file.
+```
+// application.scss
+
+$govuk-global-styles: false;
+
+@import "govuk-frontend/all/all";
+```
+If you choose this option, you will need to either use GOV.UK Frontend defined classes on links (`<a>`) and paragraphs (`<p>`). If you are also including GOV.UK Template, links and paragraph styles will be inherited from that stylesheet.
```
joelanman: this says 'either' but then just gives one course of action?

igloosi: yeah, ta. fixed now

igloosi:
so @nickcolley 's view (from slack conversation) is that we should set all compatibility to false and not include any overrides (with or without globals) for buttons, links, etc unless the consumers explicitly say  they use legacy products.

A flag, say `$legacy-styles: true` would then set `$govuk-compatibility` ones to true and include overrides in buttons, links, summary, etc for when they consume Frontend (with or without globals)

We would set globals and legacy to true for PK.

any opinions @joelanman @dashouse @hannalaakso @alex-ju @36degrees

joelanman:
it would be great to get a diagram to explain the logic for this work - happy to help out if needed

igloosi:
@dashouse made a point that if you are using template then you might need our globals as we have changed spacing in typography

igloosi:
updated to just include optional global styles based on a flag

PR #662, opened 2018-04-17
Exclude test files and snapshots folders from copy task

igloosi:
Currently tests that run after copy to `packages` or `dist` throw the following error:
```
 10 obsolete snapshot files found, run npm run test:build:packages -- -u to remove them..
```
As paths in tests are very specific and wouldn't work for consumers, we can avoid packaging them.


igloosi:
yeah, not related to the inciddent

PR #663, opened 2018-04-18
Re-instate code syntax highlighting and remove layout-debug view

igloosi:
Code highlight was accidentally removed, so this change brings it back.
As we don't really need the debug layout with the grid toggler,  that layout has also been removed

Before:
screen shot 2018-04-18 at 09 08 44

After:
screen shot 2018-04-18 at 09 09 25


igloosi:
@36degrees  forgot changelog :(

PR #665, opened 2018-04-19
Simplify grid syntax

igloosi:
Simplify grid classes based on user feedback
- Propose new class names:
```
.govuk-grid-row
.govuk-grid-column-   one-quarter, one-third, one-half, three-quarters, three-thirds, full
```
- created a map of common widths that is used to generate classes
- classes are using words, not numbers
- parameter `$at` describes when the fraction percentage kicks in

Default usage:
```
// row
@include govuk-grid-row;
// row with a custom class
@include govuk-grid-row("app-row);

// columns
@include govuk-grid-column(one-quarter);
@include govuk-grid-column(one-third);
@include govuk-grid-column(one-half);
@include govuk-grid-column(two-thirds);
@include govuk-grid-column(three-quarters);
@include govuk-grid-column(full); or  @include govuk-grid-column();
```
Define when percentage kicks in:
```
@include govuk-grid-column(three-quarters, $at: desktop)
```
Specify a custom class:
```
@include govuk-grid-column(one-quarter, $class: "app-grid");
```

- Updated example grid page: 
https://govuk-frontend-review-pr-665.herokuapp.com/examples/grid

Note: the column mixin is similar to the Frontend toolkit one
https://github.com/alphagov/govuk_frontend_toolkit/blob/master/stylesheets/_grid_layout.scss#L93

- Update all references in the review app

Trello ticket: https://trello.com/c/jjWZkDe0/865-5-simplify-the-grid-syntax

### DIFF
--- a/app/views/examples/grid/index.njk
+++ b/app/views/examples/grid/index.njk
@@ -11,69 +11,93 @@
 
 <main id="content" role="main" class="govuk-main-wrapper">
   <h1 class="govuk-heading-l">Example: Grid layout</h1>
-  <div class="govuk-grid">
-    <div class="govuk-grid__item govuk-grid__item--two-thirds">
+  <div class="govuk-layout-row">
+    <div class="govuk-layout-col-two-thirds">
       <h2 class="govuk-heading-m">Two thirds</h2>
       <p class="govuk-body">
         This guide shows how to make your service look consistent with the rest of GOV.UK. It includes example code and guidance for layout, typography, colour, images, icons, forms, buttons and data.
       </p>
     </div>
-    <div class="govuk-grid__item govuk-grid__item--one-third">
+    <div class="govuk-layout-col-one-third">
       <h2 class="govuk-heading-m">One third</h2>
       <p class="govuk-body">
         This guide shows how to make your service look consistent with the rest of GOV.UK. It includes example code and guidance for layout, typography, colour, images, icons, forms, buttons and data.
       </p>
     </div>
-    <div class="govuk-grid__item govuk-grid__item--one-third">
+    <div class="govuk-layout-col-one-third">
       <h2 class="govuk-heading-m">One third</h2>
       <p class="govuk-body">
         This guide shows how to make your service look consistent with the rest of GOV.UK. It includes example code and guidance for layout, typography, colour, images, icons, forms, buttons and data.
       </p>
     </div>
-    <div class="govuk-grid__item govuk-grid__item--one-third">
+    <div class="govuk-layout-col-one-third">
       <h2 class="govuk-heading-m">One third</h2>
       <p class="govuk-body">
         This guide shows how to make your service look consistent with the rest of GOV.UK. It includes example code and guidance for layout, typography, colour, images, icons, forms, buttons and data.
       </p>
     </div>
-    <div class="govuk-grid__item govuk-grid__item--one-third">
+    <div class="govuk-layout-col-one-third">
       <h2 class="govuk-heading-m">One third</h2>
       <p class="govuk-body">
         This guide shows how to make your service look consistent with the rest of GOV.UK. It includes example code and guidance for layout, typography, colour, images, icons, forms, buttons and data.
       </p>
     </div>
-    <div class="govuk-grid__item govuk-grid__item--one-half">
+    <div class="govuk-layout-col-one-half">
       <h2 class="govuk-heading-m">One half</h2>
       <p class="govuk-body">
         This guide shows how to make your service look consistent with the rest of GOV.UK. It includes example code and guidance for layout, typography, colour, images, icons, forms, buttons and data.
       </p>
     </div>
-    <div class="govuk-grid__item govuk-grid__item--one-half">
+    <div class="govuk-layout-col-one-half">
       <h2 class="govuk-heading-m">One half</h2>
       <p class="govuk-body">
         This guide shows how to make your service look consistent with the rest of GOV.UK. It includes example code and guidance for layout, typography, colour, images, icons, forms, buttons and data.
       </p>
     </div>
-    <div class="govuk-grid__item govuk-grid__item--one-quarter">
+    <div class="govuk-layout-col-one-quarter">
       <h2 class="govuk-heading-m">One quarter</h2>
       <p class="govuk-body">
         This guide shows how to make your service look consistent with the rest of GOV.UK. It includes example code and guidance for layout, typography, colour, images, icons, forms, buttons and data.
       </p>
     </div>
-    <div class="govuk-grid__item govuk-grid__item--one-quarter">
+    <div class="govuk-layout-col-one-quarter">
       <h2 class="govuk-heading-m">One quarter</h2>
       <p class="govuk-body">
         This guide shows how to make your service look consistent with the rest of GOV.UK. It includes example code and guidance for layout, typography, colour, images, icons, forms, buttons and data.
       </p>
     </div>
-    <div class="govuk-grid__item govuk-grid__item--one-quarter">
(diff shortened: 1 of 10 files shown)


src/globals/objects/_grid.scss
```diff
@@ -1,53 +1,78 @@
...
   }
 }
 
-@include govuk-exports("grid") {
+// A mixin for a grid column
+// Creates a cross browser grid column with a standardised gutter between the
+// columns. Widths should be defined as fractions of the full desktop width
+// they want to fill. By default they break to become full width at tablet size
+// but that can be configured to be desktop using the `$full-width` argument.
+//
```
hannalaakso: Hmmm I think `$full-width` sounds like 100% width which could be confusing as that's what the mobile column is. Maybe `$fraction-width`, `$unit-width`... I don't know, naming is hard.

src/globals/objects/_grid.scss
```diff
@@ -1,53 +1,78 @@
...
+    @include govuk-clearfix;
+    margin-right: - ($govuk-gutter-half);
+    margin-left: - ($govuk-gutter-half);
   }
 }
 
-@include govuk-exports("grid") {
+// A mixin for a grid column
+// Creates a cross browser grid column with a standardised gutter between the
+// columns. Widths should be defined as fractions of the full desktop width
```
hannalaakso: Suggestion: "units" instead of "fractions"

hannalaakso:
I like this approach too 🎉👍 Good that users can now define more grid classes if needed. Naming wise, maybe `.govuk-grid-col-1-4` (dash instead of slash), it's easier to remember and same as Bootstrap, slash is unusual.

src/globals/objects/_grid.scss
igloosi: i like units

src/globals/objects/_grid.scss
igloosi: went with $full-width as its the same param as in toolkit. maybe advanced users will be familiar with it. 
it's more like "take proportional width at this breakpoint".
`$full-width-from` ?

hannalaakso: Hmmm so the column is not "full width" from that point on. On mobile/tablet the column is full width (or 100% width). At this breakpoint the column width becomes a fraction/unit. So the name of the variable should be the opposite of "full-width" but I don't know what. `$fraction-width-from`, `$unit-width-from`?

I might not be explaining this very well 😅

hannalaakso: Or `$break-from`? I'm okay with either, we'll learn more in user research.

igloosi:
I've made changes.
- created a map of common widths that is used to generate classes
- classes are not words, not numbers
- paramater `$at` describes then the fraction percentage kicks in

Default usage:
```
  @include govuk-grid-col(one-quarter);
  @include govuk-grid-col(one-third);
  @include govuk-grid-col(one-half);
  @include govuk-grid-col(two-thirds);
  @include govuk-grid-col(three-quarters);
  @include govuk-grid-col(full);
```
Define when percentage kicks in:
```
@include govuk-grid-col(three-quarters, $at: desktop)
```
Specify a custom class:
```
@include govuk-grid-col(one-quarter, $class: "app-grid");
```

Created a custom example for 100% -> 66% -> 75% | 100% -> 33% -> 25% columns

app/views/examples/grid/index.njk
```diff
@@ -11,73 +11,97 @@
 
 <main id="content" role="main" class="govuk-main-wrapper">
   <h1 class="govuk-heading-l">Example: Grid layout</h1>
-  <div class="govuk-grid">
-    <div class="govuk-grid__item govuk-grid__item--two-thirds">
+  <div class="govuk-layout-row">
```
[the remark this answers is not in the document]
dashouse: I tend to agree, I think `govuk-grid-row` / `govuk-grid-column` would be more expected

app/views/examples/grid/index.njk
```diff
@@ -11,73 +11,97 @@
...
+        This guide shows how to make your service look consistent with the rest of GOV.UK. It includes example code and guidance for layout, typography, colour, images, icons, forms, buttons and data.
+      </p>
+    </div>
+    <div class="govuk-layout-col-one-third on-desktop-one-quarter">
+      <h2 class="govuk-heading-m">100% on mobile, 33% on tablet, 25% on desktop</h2>
+      <p class="govuk-body">
+        This guide shows how to make your service look consistent with the rest of GOV.UK. It includes example code and guidance for layout, typography, colour, images, icons, forms, buttons and data.
+      </p>
+    </div>
+    <div class="govuk-layout-col-two-thirds on-desktop-three-quarters">
```
dashouse: I love the functionality of `on-desktop-three-quarters` but I'm worried it doesn't feel like it's part of the grid syntax.

Did we discuss anything like: 
`govuk-grid-col-two-thirds` `d:govuk-grid-col-three-quarters`

src/globals/objects/_grid.scss
```diff
@@ -1,53 +1,16 @@
...
+  @include govuk-layout-row;
+  @include govuk-layout-col(one-quarter);
+  @include govuk-layout-col(one-third);
+  @include govuk-layout-col(one-half);
+  @include govuk-layout-col(two-thirds);
+  @include govuk-layout-col(three-quarters);
+  @include govuk-layout-col(full);
+
+  // custom 100,66,75%
+  @include govuk-layout-col(three-quarters, $at: desktop, $class: "on-desktop");
```
nickcolley: Should we have custom examples as part of the core library? It seems good to demonstrate but we shouldn't include them unless we know they'll be used?

igloosi: no for demo only. still WIP :)

app/views/examples/grid/index.njk
igloosi: on-desktop-three-quarters was an example class name. i could have just used 
`@include govuk-layout-col(three-quarters, $at: desktop, $class: "govuk-grid-col-desktop");`

igloosi:
updated classes and example page to
```
.govuk-grid-row
.govuk-grid-column
```

joelanman:
This looks great, very readable/predictable.

There was some concern earlier that this was moving away from BEM, is that still the case?

igloosi:
it's now kinda been, just two block elements

joelanman:
From a quick discussion with @igloosi it seems like `govuk-grid-row` isn't quite what I'd expect.

In our example page, it doesn't wrap each row, it wraps the whole grid.

Is it at all useful to wrap each row? If not maybe the previous name was actually better - just `govuk-grid`

igloosi:
i guess if you have wildly differing content length in columns you might want to wrap in multiple rows

igloosi:
updated the example page

dashouse:
@joelanman @igloosi I think the choice whether to use grid containers as individual rows or as a wrapper for all grid items is dependent on content to some degree...

For example I've used two rows here https://govuk-design-system-production.cloudapps.digital/styles/layout/common-two-thirds-two-thirds-one-third/index.html

But equally if you just fill 2/3 1/3 1/4 3/4 it wouldn't technically need a new row as they would wrap.

It gets a little more complex when you actually want it to wrap because your grid changes from 4 > 3 > 2 so it doesn't leave widowed columns

joelanman:
I think even if it would work in some situations if you didn't wrap every row, its an easier thing to get your head around if we always do?

dashouse:
Sure, so this would be just updating the example in the review app rather than anything else?

joelanman:
Seems so, I just checked the Design System layout page and it seems like we already wrap every row there.

igloosi:
and i have updated the review app grid example page as well

PR #668, opened 2018-04-20
Remove cursor: pointer from buttons as standard

penx:
Remove cursor: pointer from buttons as standard. This is common practice but against all best practices I'm aware of, including all browser vendor defaults, W3C and many OS HCI guidelines.

Source:

https://medium.com/simple-human/buttons-shouldnt-have-a-hand-cursor-b11e99ca374b

### DIFF
--- a/src/button/_button.scss
+++ b/src/button/_button.scss
@@ -50,7 +50,6 @@
     box-shadow: 0 $button-shadow-size 0 $govuk-button-shadow-colour; // s0
     text-align: center;
     vertical-align: top;
-    cursor: pointer;
     -webkit-appearance: none;
 
     @include mq($from: tablet) {


igloosi:
Hi @penx,
thanks for the contribution. 
`.govuk-c-button` class can be applied to `<button>`, `<a>` or `<input type=submit>` so we wouldn't want to remove the pointer from the class directly. 
To not display the cursor on `<button>` only, a more target approach is need.

penx:
`<a href="#" class="govuk-c-button">` would have cursor:pointer set on it because it's the browser default

penx:
`<input type=submit>` is a type of button and falls in to the same argument as linked above, it shouldn't have `cursor: pointer` in my view (and that of W3C, browser defaults and Microsoft and Apple's guidelines)

By putting the `cursor: pointer` in, you are explicitly stating you don't agree with the browser's default behaviour for elements.

igloosi:
Thanks,
i'll leave it her for the rest to review next week and hope we can get it into the next release.

penx:
Related:

https://medium.com/simple-human/but-sometimes-links-look-like-buttons-and-buttons-look-like-links-9b371c57b3d2
https://adamsilver.io/articles/buttons-shouldnt-have-a-hand-cursor-part-2/

joelanman:
Thanks for your PR!

Advantages of hand cursor on buttons:

- Slightly more affordance
- Consistency - all clickable elements use the same cursor

Disadvantages:

- Users can't use the hover cursor to see the difference between buttons and links

It seems to me that the disadvantage is really quite esoteric, and I don't see a clear user need. From Adam's article linked above:

Clicking a link opens a web page or resource.
(On desktop) I can right-click on a link and do many things (that I can’t do with a button). Open in new tab/window, save a link, copy address, add to reading list, bookmark it and more.
(On mobile devices) I can tap and hold on a link and get a similar context menu as per the previous point.
A link also tells me that I am just going somewhere else. I am not modifying any data or making changes in anyway (like a button is likely to do).

joelanman:
Thanks for your PR!

Advantages of hand cursor on buttons:

- Slightly more affordance
- Consistency - all clickable elements use the same cursor

Disadvantages:

- Users can't use the hover cursor to see the difference between buttons and links

It seems to me that the disadvantage is really quite esoteric, and I don't see a clear user need. From Adam's article linked above:

> Clicking a link opens a web page or resource.

Clicking a button also does this in many cases

> (On desktop) I can right-click on a link and do many things (that I can’t do with a button). Open in new tab/window, save a link, copy address, add to reading list, bookmark it and more.
(On mobile devices) I can tap and hold on a link and get a similar context menu as per the previous point.
A link also tells me that I am just going somewhere else. I am not modifying any data or making changes in anyway (like a button is likely to do).

joelanman:
Thanks for your PR!

Advantages of hand cursor on buttons:

- Slightly more affordance
- Consistency - all clickable elements use the same cursor

Disadvantages:

- Users can't use the hover cursor to see the difference between buttons and links

It seems to me that the disadvantage is really quite esoteric, and I don't see a clear user need to use the hover cursor to see the difference between buttons and links.

penx:
Disadvantages (to the user) 
- enforces the idea that if an element does not have `cursor: pointer` then it is not clickable.
- makes the user think that this is standard web behaviour, when it it is actually [seen by the W3C as misuse or abuse](http://lists.w3.org/Archives/Public/public-css-testsuite/2010Jul/0024.html).
- if a user sees a `cursor: pointer` and (correctly) expects this to mean it's a link to a location, they would expect to be able to right click the element to open in a new window.

Other disadvantages
- applies web-only tropes to any system that consumes this CSS e.g. React Native.

> all clickable elements use the same cursor

I'm not sure why this is an advantage for the user, nor do I think this is something that govuk-frontend intends to make consistent for all clickable elements. I don't want to repeat the points in Adam Silver's article, as he explains it much better detail, so I'll link to it again with the summary [The hand does not mean clickable](https://medium.com/simple-human/buttons-shouldnt-have-a-hand-cursor-b11e99ca374b).

> Slightly more affordance

Personally I don't think the design of the govuk button is at all lacking in affordance.

I'd also argue this phrasing the decision as 
> "use cursor pointer" vs. "don't use cursor pointer"

has a small but important difference from:
> "use cursor pointer" vs. "defer the decision to the browser"

i.e. if browser vendors decided to change the standard cursor for buttons in future, my proposed change would mean govuk-frontend would also change to use cursor pointer.

Also worth considering - is the behaviour of govuk-frontend intended to be specific to the web?

CSS is not web specific. E.g. these styles could be consumed with react-native. Native applications generally do not use the concept of "cursor:pointer" means clickable. You normally only see cursor: pointer in a native application when presented with a hyperlink.

PR #675, opened 2018-04-26
Disable code syntax highlighting in IE8

nickcolley:
Prism will throw syntax errors making it hard to test IE8 JavaScript.


igloosi:
code-wise it's all good.
Q: are we ok with IE8 users not getting code styling as prism doesn't support ie8?

36degrees:
IMO, for the review app, yes.

PR #682, opened 2018-05-04
Section break styles fix when used with Elements

igloosi:
`<hr>` styling in Elements has background and height
set, whereas in Frontend we're using border bottom.

To fix, `height:0` was added in a compatibility conditional include.

This fixes both the thickness and the fact that styles from Elements were making invisible section breaks visible
Before:
screen shot 2018-05-04 at 14 14 37

After:
screen shot 2018-05-04 at 14 14 55

Elements styles causing non-visible section break to be visible:

screen shot 2018-05-08 at 09 06 28

Ticket: https://trello.com/c/Sc7RtFZK/815-remove-section-break-border-coming-from-elements

### DIFF
--- a/src/globals/core/_section-break.scss
+++ b/src/globals/core/_section-break.scss
@@ -5,6 +5,12 @@
   %govuk-section-break {
     margin: 0;
     border: 0;
+
+    // fix double-width section break and forced visible section break
+    // when combined with styles from alphagov/elements
+    @include govuk-compatibility(govuk_elements) {
+      height: 0;
+    }
   }
 
   .govuk-section-break {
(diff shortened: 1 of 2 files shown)


src/globals/core/_section-break.scss
```diff
@@ -44,6 +44,9 @@
 
   %govuk-section-break--visible {
     border-bottom: 1px solid $govuk-border-colour;
+    @include govuk-compatibility(govuk_elements) {
```
nickcolley: What happens when elements overrides the invisible break?

igloosi: me being stupid and not moving it into where i said i was. fixed now.

src/globals/core/_section-break.scss
```diff
@@ -5,6 +5,12 @@
   %govuk-section-break {
     margin: 0;
     border: 0;
+
+    // fix doouble-width section break and forced visible section break
```
nickcolley: Typoe 'doouble'

PR #683, opened 2018-05-08
Make `button` element the default instead of `input` for button macro

nickcolley:
We are making `button` the default, however since the `button` element is buggy in IE6 / IE7, we are allowing input to be used if required.

Thanks @quis for the original issue on GOV.UK Elements :)

Context: https://github.com/alphagov/govuk_elements/pull/593

Trello: https://trello.com/c/gITiuXcS/765-use-button-by-default-for-buttons-rather-than-input-typebutton

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,10 @@ Breaking change:
   to a more consise naming structure. We have also introduced two mixins
   to help generate additional or custom grid styles and widths.
 
+- Make `button` element the default instead of `input` for button macro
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -23,6 +23,10 @@ Breaking change:
   to a more consise naming structure. We have also introduced two mixins
   to help generate additional or custom grid styles and widths.
 
+- Make `button` element the default instead of `input` for button macro #683
```
igloosi: extra PR #number

nickcolley: sorted

Issue #460, opened 2018-01-24
Macros Component API should not require internal classes for modifiers


NickColley:
Macros abstract our class names, if we were to change our class names, macro usage would break for our users.

In our button examples, we pass the internal button modifier class to make a button into a start button.

We could consider passing a boolean instead so that if we choose to change how our classes work, macros still function.

For example instead of

```yaml
- name: start
  data:
    text: Start now button
    classes: 'govuk-c-button--start'
```

We could do:

```yaml
- name: start
  data:
    text: Start now button
    start: true
```

See also the GOV.UK Publishing component: http://govuk-static.herokuapp.com/component-guide/button/start_now_button

Link to code: https://github.com/alphagov/static/blob/a9a040a72d475ef7015c1c9e690a5d77687fcfd8/app/views/govuk_component/button.raw.html.erb#L1

We constructed the class based on booleans, then finally merge with `classes` and inject into the markup.

Edit: 
I put together an example showing how you can restructure the macro to avoid the confusing forking logic.

This approach only forks on individual parts of the template, which is easier to reason about.

https://glitch.com/edit/#!/flag-based-macro-approach?path=views/index.html:15:26

igloosi:
we wen't through this discussion when we standardised component api, previous version of button macro had boolean (https://github.com/alphagov/govuk-frontend/commit/4559379446391bfcfb8ef620ff987c0b183ba63d#diff-c8d55831d340874284fc322fd388fca6R14)
That was clashing with other components where boolean wouldn't work

joelanman:
yeh at the time we thought MVP, classes will work. However @nickcolley makes a good point, that this is fragile - if we ever change our class names, it would break.

we also discussed something like

`variant: 'start'`

which is more flexible

nickcolley:
Interesting idea, we could consider `modifier: 'start'` if we wanted to align with BEM.

I think the reason I prefer Boolean flags is that you may want to do other things than just add a modifier class.

nickcolley:
@igloosi can you explain what you mean? The commit you linked to doesn't make it clear why that does not work.

36degrees:
A good example of where the convention 'breaks' is [the way the list component used to work](https://github.com/alphagov/govuk-frontend/blob/4559379446391bfcfb8ef620ff987c0b183ba63d/src/components/list/template.njk#L1-L13) - there are multiple boolean possibilities but only one of them is possible at a time and the order of precedence is undefined.

[...]

PR #684, opened 2018-05-10
Better define relationship between headings, error message and hints

36degrees:
This decouples the markup of the label as the page heading from the styling, by removing any styling from the heading and delegating all styling responsibilities to the label.

This: 

- introduces new 'variants' for legends and labels (e.g. `govuk-label--xl`, `govuk-fieldset__legend--xl`) that make the legend or label the same size and weight as the equivalent heading
- introduces a new parameter `legend.isPageHeading` for fieldsets, which defines whether the legend text should be wrapped in an `h1` – `<legend><h1>Have you changed your name?</h1></legend>`
- introduces a new parameter `isPageHeading` for the label component, which defines whether the label _itself_ should be wrapped in an `h1` – `<h1><label>What is your National Insurance number?</label></h1>`
- adds styling for `h1`s in both contexts which effectively 'reset' the h1 such that it has no effect – i.e. the `isPageHeading` flag should have no effect on the styling of the question.
- updates components that call the label or fieldset component to pass these parameters.

screen shot 2018-05-10 at 11 04 09-fullpage

You can see all combinations of [here](https://govuk-frontend-review-pr-684.herokuapp.com/examples/labels-legends-and-headings).

### DIFF
--- a/src/fieldset/fieldset.yaml
+++ b/src/fieldset/fieldset.yaml
@@ -7,4 +7,5 @@ examples:
   data:
     legend:
       text: What is your address?
+      classes: govuk-fieldset__legend--xl
       isPageHeading: true
--- a/src/fieldset/fieldset.yaml
+++ b/src/fieldset/fieldset.yaml
@@ -1,8 +1,10 @@
 examples:
 - name: default
   data:
-    legendText: What is your address?
+    legend:
+      text: What is your address?
 - name: as page heading
   data:
-    legendText: What is your address?
-    legendIsPageHeading: true
+    legend:
+      text: What is your address?
+      isPageHeading: true
--- a/src/fieldset/fieldset.yaml
+++ b/src/fieldset/fieldset.yaml
@@ -2,3 +2,7 @@ examples:
 - name: default
   data:
     legendText: What is your address?
+- name: as page heading
+  data:
+    legendText: What is your address?
+    legendIsPageHeading: true
--- a/src/label/_label.scss
+++ b/src/label/_label.scss
@@ -17,8 +17,25 @@
     display: block;
   }
 
-  .govuk-label--bold {
-    @include govuk-typography-weight-bold;
+  // Modifiers that make labels look more like their equivalent headings
+
+  .govuk-label--xl {
+    @include govuk-font-bold-48;
+    margin-bottom: $govuk-spacing-scale-3;
+  }
+
+  .govuk-label--l {
+    @include govuk-font-bold-36;
+    margin-bottom: $govuk-spacing-scale-3;
+  }
+
+  .govuk-label--m {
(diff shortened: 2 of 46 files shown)


src/fieldset/fieldset.yaml
```diff
@@ -7,4 +7,5 @@ examples:
   data:
     legend:
       text: What is your address?
+      classes: govuk-fieldset__legend--xl
```
nickcolley: Not blocking (since can be changed later) but this exposes internals (https://github.com/alphagov/govuk-frontend/issues/460)

src/label/_label.scss
```diff
@@ -15,9 +15,38 @@
     @include govuk-text-colour;
 
     display: block;
+
+    margin-bottom: $govuk-spacing-scale-1;
+  }
+
+  // Modifiers that make labels look more like their equivalent headings
+
+  .govuk-label--xl {
```
nickcolley: If we used the core heading classes would we need to do this?

src/fieldset/fieldset.yaml
36degrees: It's consistent with the way we do it everywhere else at the minute.

src/label/_label.scss
igloosi: i prefer that it's separate.

36degrees: The intention behind this is to decouple the idea of _what's a heading_ with what's important on the page - reframing the problem so that `heading-xl` is 'a heading which is extra large', and `label--xl` is 'a label that is extra large' – you can have an medium label which is not a heading, and you could (theoretically) have a label with no additional styling that is a heading. Using the core heading classes muddies the semantics again.

There are some additional benefits in that we don't actually want any of the margins that the core heading classes give us, so at the minute we have to override them - whereas this allows us to define them separately.

igloosi:
changelog

src/label/_label.scss
nickcolley: Think it's important to think about how the user views these, both `label--xl` and `heading--xl` is the same thing visually.

I'm mostly concerned about:

- how we keep the two in sync
- impact on file size
- another interface to learn

That said, if it's not easy to use the core classes, this seems a reasonable approach.

Other benefit is that decoupling them makes them easier to delete.

nickcolley: I should have prefixed my comment with that it wasnt blocking.

Just wanted to understand that using the core classes was not possible, since I think this is very complex to understand.

36degrees: It's totally possible, but I don't think it's actually desirable in this case, for the reasons outlined above.

36degrees:
@igloosi happy with c13793a?

igloosi:
👍

PR #687, opened 2018-05-10
Include gulp-cli to avoid required global packages.

matcarey:
## The Problem

Currently to follow the steps in the readme the user needs to have already run `npm install -g gulp-cli`.  This means that the `gulp-cli` version will vary between users and some users would rather not install anything globally.

## The Solution

This change is the result of running `npm install --save-dev gulp-cli`, any scripts defined in `package.json` can refer to local dependancies therefore with this change any user can clone, `npm install` and `npm start` without global dependencies.  This also means that the version of `gulp-cli` is defined in `package.json` and `package-lock.json` which gives a more consistent outcome of the `npm start` and any `npm run XX` commands by different users/boxes.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -208,6 +208,15 @@
         }
       }
     },
+    "ansi-colors": {
+      "version": "1.1.0",
+      "resolved": "https://registry.npmjs.org/ansi-colors/-/ansi-colors-1.1.0.tgz",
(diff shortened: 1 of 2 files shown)


matcarey:
When attempting to reproduce this I couldn't therefore I'm removing the PR.  Steps to reproduce if it is an issue for anyone else would be:

```
mkdir tmp && cd tmp
git clone git@github.com:alphagov/govuk-frontend.git
cd govuk-frontend
nvm install && nvm use
npm install
npm start
```

PR #689, opened 2018-05-10
RFC: Remove `r` in spacing overrides

dashouse:
As part of the "Make prototyping easier" work I think another easy fix would be to remove `r` from the spacing override classes.

The `r` stands for "responsive". With the way we have defined spacing for components and typographic elements I am not convinced we will need static spacing classes in the future as it makes sense for all spacing to follow the same rules.

We have also heard through research that the `r` can be confused with "right".

For example the expectation of adding `govuk-!-m-r9` would be applying `margin-right: 60px` not the actual output `margin: 60px`.

**Before:**
`govuk-!-mr-r9`
`govuk-!-p-r9`
etc

**After:**
`govuk-!-mr-9`
`govuk-!-p-9`
etc

### DIFF
--- a/app/views/examples/error-summary-with-messages/index.njk
+++ b/app/views/examples/error-summary-with-messages/index.njk
@@ -37,7 +37,7 @@
 
   {{ govukInput({
     label: {
-      "html": '<h3 class="govuk-heading-m govuk-!-mb-r1">Passport number</h3>',
+      "html": '<h3 class="govuk-heading-m govuk-!-mb-1">Passport number</h3>',
       "hintText": "For example, 502135326"
     },
     id: "passport-number",
--- a/app/views/examples/links/index.njk
+++ b/app/views/examples/links/index.njk
@@ -38,7 +38,7 @@
 
     {% for variant_description, variant_class in variants %}
 
-      <section class="govuk-!-mt-r8">
+      <section class="govuk-!-mt-8">
         <h2 class="govuk-heading-l">{{ variant_description }}</h2>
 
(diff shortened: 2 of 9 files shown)


nickcolley:
Main risk I can think of is it this could make it hard to introduce a static spacing scale.

Related: http://tachyons.io/docs/layout/spacing/

36degrees:
Yep, that. Though can definitely see confusion for 'right' being an issue. Wonder if there's any other change we could make that would solve that issue without making it harder to introduce non-responsive overrides later.

dashouse:
@nickcolley @36degrees Yep, that was the consideration for adding it in the first place, however with typography and component (outer most wrappers) having responsive vertical spacing, components with padding (like Error summary / Panel) having responsive internal spacing I can't think of a scenario where having a class based static spacing scale would be necessary.

If there is, I think the responsive spacing scale classes should be the default and the fixed scale overrides would be like a modifier e.g. `govuk-!-mb-9--fixed`.

I think there is a need for a static scale as a variable, I'm just not sure there's a need for these as class based overrides.

Happy to close, just thought it might be the last chance to change this.

timpaul:
This exact issue just happened to me. I spent a good few minutes assuming that the 'r' stood for 'right' and trying out things like `govuk-!-m-b9`, `govuk-!-m-l9` etc.

dashouse:
I have no idea if the thumbs up are to carry on or thumbs upping me deleting it

nickcolley:
If we have research that people keep running into this issue, feels reasonable to get in before 1.0.0

PR #694, opened 2018-05-14
Add display overrides

hannalaakso:
This PR:

- Adds override classes that allow the CSS `display` property to be set to `block`, `inline` or `inline-block`.
- Adds examples to the review app 

There are possibly more appropriate components to demonstrate how the classes should be used than the ones I picked - shout if you've got a suggestion.

Trello card: https://trello.com/c/haIqNCRH/664-fix-buttons-not-being-block-level-elements
Review app page: https://govuk-frontend-review-pr-694.herokuapp.com/examples/form-alignment

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,10 @@ Fixes:
 
 - Update component readme files content
   ([PR #704](https://github.com/alphagov/govuk-frontend/pull/704))
-  
+
 New features:
+
+- Add override classes to set `display` property to `block`, `inline` and `inline-block` (PR [#694](https://github.com/alphagov/govuk-frontend/pull/654))
 
 - Add option to set CSS display property for govuk-shape-arrow mixin
   ([PR #701](https://github.com/alphagov/govuk-frontend/pull/701))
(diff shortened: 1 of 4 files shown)


igloosi:
hmm, we do `govuk-\!-w` for weight, `.govuk-\!-f-` for font, `.govuk-\!-mt-` for margin-top
whereas here we do `.govuk-\!-display-`, aka the whole word.
should we change it to ``.govuk-\!-d-`?

nickcolley:
It would be good to have consistency but the width overrides already have long hand selectors.

joelanman:
Seems more readable to use full words. We generally use readable function/variable names in our code, so why not here?

igloosi:
consensus is to use long name. @hannalaakso can you fix conflict and let's ship it

igloosi:
conflict fixed and ready for review

CHANGELOG.md
```diff
@@ -302,6 +302,8 @@ New features:
 
   To include global styles, you can set `$govuk-global-styles` variable to `true`.
 
+- Add override classes to set `display` property to `block`, `inline` and `inline-block` (PR [#694](https://github.com/alphagov/govuk-frontend/pull/652))
```
nickcolley: Looks like the new feature is being added to the previous release

igloosi: yeah fixed now

PR #693, opened 2018-05-11
Restructure src directory

36degrees:
## Objectives
- Restructure the `src` directory to make the structure of Frontend clearer, now that we are not confined by needing to structure around multiple packages.
- Simplify imports within components
- Add tests to ensure that the `settings`, `helpers` and `tools` layers do not output CSS.

## Summary

- Move components into `./components`
- Move `/icons` into ./assets/image
- Move `globals/settings` into `./settings`
- Move `globals/tools` into `./tools`
- Split `globals/helpers` so that mixins are in `./tools`, and concrete classes are in `./utilities`
- Move `globals/core` into `./core`
- Move `globals/objects` into `./objects`
- Move `globals/overrides` into `./overides`
- Move `mq()` into `./vendor`
- Move `globals/font-face.scss` into `./helpers/font-faces.scss` and refactor as a mixin
- Move `globals/common.js` to `./common.js`
- Move polyfills into `./vendor`
- Add entry points to each layer e.g. `settings/_all.scss`
- Update documentation

I suggest reviewing commit-by-commit.

## Before
```
src
├── back-link
├── breadcrumbs
├── button
├── checkboxes
├── date-input
├── details
├── error-message
├── error-summary
├── fieldset
├── file-upload
├── footer
├── globals
│   ├── core
│   ├── helpers
│   ├── objects
│   ├── overrides
│   ├── polyfills
│   │   ├── Function
│   │   │   └── prototype
│   │   └── Object
│   ├── settings
│   └── tools
├── hint
├── icons
├── input
├── label
├── panel
├── phase-banner
├── radios
├── select
├── skip-link
├── table
├── tag
├── textarea
└── warning-text
```

## After
```
src
├── assets
│   └── images
├── components
│   ├── back-link
│   ├── breadcrumbs
│   ├── button
│   ├── checkboxes
│   ├── date-input
│   ├── details
│   ├── error-message
│   ├── error-summary
│   ├── fieldset
│   ├── file-upload
│   ├── footer
│   ├── hint
│   ├── input
│   ├── label
│   ├── panel
│   ├── phase-banner
│   ├── radios
│   ├── select
│   ├── skip-link
│   ├── table
│   ├── tag
│   ├── textarea
│   └── warning-text
├── core
├── helpers
├── objects
├── overrides
├── settings
├── tools
├── utilities
└── vendor
    └── polyfills
        ├── Function
        │   └── prototype
        └── Object
```
[...]

PR #695, opened 2018-05-15
Add header component

alex-ju:
Review link: https://govuk-frontend-review.herokuapp.com/components/header

New version of the GOV.UK Header based off the work done in the GOV.UK Design System.

## Notable changes
- Macro interface that supports service name and navigation
- B.E.M.
- Updates spacing based on the existing spacing scale and values
- Fixes alignments
- Uses inline SVG for the crown (can be see in inverse, prints sharp)
- Mobile navigation items on a single line with appropriate touch area

## Testing
[browser and AT tests were done; screen captures and notes coming soon]

## Notes
Although some experimental work was done around it, this PR doesn't include the search component and the macro interface for it.


src/header/_header.scss
```diff
@@ -0,0 +1,290 @@
...
+@import "../globals/helpers/focusable";
+@import "../globals/helpers/spacing";
+@import "../globals/helpers/device-pixels";
+@import "../globals/helpers/grid";
+@import "../globals/helpers/shape-arrow";
+
+@import "../globals/helpers/typography";
+@import "../globals/settings/typography-font-stacks";
+@import "../globals/settings/typography-font";
+@import "../globals/settings/typography-responsive";
```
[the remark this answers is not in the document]
alex-ju: yep. will probably be merged after the restructure anyway. will be easier to rebase on this one.

igloosi:
padding top on ie8 adds a black line 
screen shot 2018-05-15 at 10 29 16

igloosi:
firefox doesnt seem to like toggling
screen shot 2018-05-15 at 10 35 19

igloosi:
looks good and works in other supported browsers

src/header/template.njk
```diff
@@ -0,0 +1,61 @@
+{% from "phase-banner/macro.njk" import govukPhaseBanner %}
+
+<header class="govuk-header {{ params.classes if params.classes }}" role="banner" data-module="header"
+        {%- for attribute, value in params.attributes %} {{attribute}}="{{value}}"{% endfor %}>
+  <div class="govuk-header__container {{ params.containerClasses if params.containerClasses }}">
+
+    <div class="govuk-header__logo">
+      <a href="{{ params.homepageUrl }}" class="govuk-header__link govuk-header__link--homepage">
```
igloosi: could we consider having a default of / here?

src/header/_header.scss
```diff
@@ -0,0 +1,290 @@
...
+  .govuk-header__navigation--right {
+    @include mq ($from: desktop) {
+      margin: 0;
+      padding: $govuk-spacing-scale-1 0;
+      text-align: right;
+    }
+  }
+
+  .govuk-header__navigation--no-service-name {
+    padding-top: 40px; // 30px line-height of the service name + 10px for margin-bottom
```
igloosi: $govuk-spacing-scale-7 ?

src/header/_header.scss
```diff
@@ -0,0 +1,290 @@
...
+
+    @include mq ($from: desktop) {
+      display: none;
+    }
+  }
+
+  .govuk-header__menu-button--open {
+    &::after {
+      @include govuk-shape-arrow($direction: up, $base: 10px);
+      content: "";
```
36degrees: Shouldn't need to redeclare `content`, `display` or `margin-left`, they should be able to inherit from `govuk-header__menu-button`?

src/header/template.njk
alex-ju: Updated, added `/` as default value.

src/header/_header.scss
alex-ju: Updated, used the recommended `$govuk-spacing-scale-7` for padding.

src/header/template.njk
```diff
@@ -0,0 +1,61 @@
...
+        <span class="govuk-header__logotype">
+          <svg
+            role="presentation"
+            class="govuk-header__logotype-crown"
+            xmlns="http://www.w3.org/2000/svg"
+            viewbox="0 0 132 97"
+            height="32"
+            width="36"
+          >
+            <path
```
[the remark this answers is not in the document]
alex-ju: Updated. Added the comments from the Design System.

src/header/_header.scss
alex-ju: Updated, removed attributes that could be inherited. `display` is still needed, as it overwrites the style coming from the `govuk-shape-arrow` mixin.

src/header/_header.scss
igloosi: 👍

src/header/template.njk
igloosi: 👍

igloosi:
https://github.com/alphagov/govuk-frontend/pull/695/commits/fc0b5d93f8e93a9c3add9289dacd684184c98f40 fixed the Firefox issue

alex-ju:
Updated, addressing all comments.

@igloosi:
- reeeeally long service name was addressed in c38e15f
- padding top on ie8 adds a black line fixed with bce4888
- firefox on windows doesn't seem to like toggling fixed with fc0b5d9

alex-ju:
Restructured based on changes coming from #693.

src/components/header/_header.scss
```diff
@@ -0,0 +1,274 @@
...
+@import "../../tools/all";
+@import "../../helpers/all";
+
+@import "../../helpers/typography";
+
+@include govuk-exports("header") {
+
+  $govuk-header-background: $govuk-black;
+  $govuk-header-border-color: $govuk-blue;
+  $govuk-header-border-width: $govuk-spacing-scale-2;
```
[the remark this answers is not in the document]
alex-ju: We could, but that will make it inconsistent across components. (i.e. variables declaration in footer)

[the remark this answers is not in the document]
igloosi: let's keep it consistent then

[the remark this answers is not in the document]
36degrees: [We're](https://github.com/alphagov/govuk-frontend/blob/master/src/components/breadcrumbs/_breadcrumbs.scss#L7-L20) [already](https://github.com/alphagov/govuk-frontend/blob/master/src/components/button/_button.scss#L10) [inconsistent](https://github.com/alphagov/govuk-frontend/blob/master/src/components/footer/_footer.scss#L9-L14) 😞 

¯\\_(ツ)_/¯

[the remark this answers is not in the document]
igloosi: :shipit:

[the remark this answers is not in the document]
alex-ju: I'd rather raise a separate PR to address this inconsistencies, but as you wish..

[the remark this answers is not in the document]
36degrees: Totally happy for that to happen, just pointing out that the inconsistency is already there. Don't expect it to be addressed in this PR.

app/views/examples/header-and-footer/index.njk
```diff
@@ -0,0 +1,48 @@
...
+{% from "header/macro.njk" import govukHeader %}
+{% from "phase-banner/macro.njk" import govukPhaseBanner %}
+{% from "footer/macro.njk" import govukFooter %}
+
+{% block content %}
+
+{# {{ govukSkipLink({
+  href: '#content',
+  text: 'Skip to main content'
+}) }} #}
```
hannalaakso: Why is the code commented out? Is it opt in by uncommenting it? Could the code mention this?

src/components/header/_header.scss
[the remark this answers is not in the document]
alex-ju: Yep, you're right. It was probably introduced with the footer and I carried on with it.

src/components/header/template.njk
```diff
@@ -0,0 +1,79 @@
...
+            ></path>
+            {# Fallback PNG image for older browsers.
```
nickcolley: Excellent comments!

alex-ju: @36degrees' work, not mine

app/views/examples/header-and-footer/index.njk
alex-ju: Wasn't sure if the example should contain the skip-link or not. I added it back now. I think is useful when testing. Would be good to have the cookie-banner added to when it will be in.

Regarding the other comments: I tested with the ATs already. Mobile menu may be expanded by default, but that means there will be movement in the page when javascript kicks in and hides it.

src/components/header/_header.scss
```diff
@@ -0,0 +1,274 @@
...
+
+    @include mq ($from: desktop) {
+      display: none;
+    }
+  }
+
+  .govuk-header__menu-button--open {
+    &::after {
+      @include govuk-shape-arrow($direction: up, $base: 10px);
+      display: inline-block;
```
hannalaakso: It's a shame we have to set `display: inline-block` twice to override the mixin. Completely NOT a problem with the PR. I've raised an [issue.](https://github.com/alphagov/govuk-frontend/issues/698). It might worth linking to the issue in the code to say the second `display: inline-block` can be removed if the issue is fixed. Up to you @alex-ju 🙂

igloosi: https://github.com/alphagov/govuk-frontend/pull/701

alex-ju: @hannalaakso: @igloosi created a separate PR to deal with this

src/components/header/_header.scss
```diff
@@ -0,0 +1,274 @@
...
+      }
+    }
+  }
+
+  .govuk-header__navigation-item:last-child {
+    margin-right: 0;
+  }
+
+  @include mq($media-type: print) {
+    .app-header {
```
nickcolley: I think we may need the govuk- namespace

src/components/header/_header.scss
```diff
@@ -0,0 +1,274 @@
...
+    @include mq ($from: desktop) {
+      display: block;
+    }
+  }
+
+  .govuk-header__navigation--open {
+    display: block;
+  }
+
+  .govuk-header__navigation--right {
```
nickcolley: Could use `--end` instead of `--right` here

src/components/header/header.js
```diff
@@ -0,0 +1,57 @@
...
+}
+
+/**
+* Toggle class
+* @param {object} node element
+* @param {string} className to toggle
+*/
+Header.prototype.toggleClass = function (node, className) {
+  if (node.classList.contains(className)) {
+    node.classList.remove(className)
```
nickcolley: I think we need a polyfill for IE9 support.

I don't think too many people would notice since IE9 users typically is used on desktop machines but better to be safe.

alex-ju:
@nickcolley: thanks for the review, good findings. updated addressing them.

src/components/header/template.njk
```diff
@@ -0,0 +1,79 @@
+{% from "phase-banner/macro.njk" import govukPhaseBanner %}
```
nickcolley: I think we can remove this one

alex-ju:
@nickcolley Updated. Now the header relies on a `.js-enabled` class being added to the `<body>` (or actually any other parent element). Test for non-javascript behaviour added.

igloosi:
this looks very good now. can we ship it?

alex-ju:
@igloosi updated:
- Rebased to master
- Uses the new `govuk-shape-arrow` mixin
- Updated Changelog

P.S. I'm aware it needs some commit squashing

alex-ju:
Update: clean-up commit history

Issue #641, opened 2018-04-10
Theming: Components should alias variables used within their file


nickcolley:
When using variables we're inconsistent where they're defined and consumed.

In some places we rely on an [`applied` file](https://github.com/alphagov/govuk-frontend/blob/master/src/globals/settings/_colours-applied.scss), in other cases [we define them next to component definitions](https://github.com/alphagov/govuk-frontend/blob/master/src/footer/_footer.scss#L23).

Another example is the panel component which [references a top level variable directly](https://github.com/alphagov/govuk-frontend/blob/master/src/panel/_panel.scss#L37), making theming very difficult without changing colours for other components.

My proposal would be to consistently set variables at the top of each component file that references top level variables.

igloosi:
comment from slack

> Hi, I got an odd issue I am hoping you can help with.  I have overridden $govuk-brand-colour and the application.scss is updated (when inspecting it).  However this is used in the header indirectly via "$govuk-header-border-color: $govuk-brand-colour" ... however it still has this as "$govuk-header-border-color: $govuk-blue;"  (i.e no change in colour).  Any ideas?   [Also should it not be $govuk-header-border-*colour*]

nickcolley:
Going to close for now, we may look into whitelabelling in the future so we can re-open then.

Issue #640, opened 2018-04-10
Theming: Variables can not be overriden


nickcolley:
Most of our variables do not use the `!default` flag which means if someone wants to change colours or spacing variables for a theme they can not.

nickcolley:
Going to close for now, we may look into whitelabelling in the future so we can re-open then.

PR #696, opened 2018-05-16
WIP: Component-specific aliases for variables

igloosi:
Allow for easier theming / general overriding of component variables.

Addresses: https://github.com/alphagov/govuk-frontend/issues/641, some of https://github.com/alphagov/govuk-frontend/issues/640

### DIFF
--- a/src/components/back-link/_back-link.scss
+++ b/src/components/back-link/_back-link.scss
@@ -4,6 +4,8 @@
 
 @include govuk-exports("back-link") {
 
+  $backlink-border-colour: $govuk-black !default;
+
   .govuk-back-link {
     @include govuk-font-regular-16;
     @include govuk-link-common;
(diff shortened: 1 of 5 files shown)


src/components/back-link/_back-link.scss
```diff
@@ -4,6 +4,8 @@
 
 @include govuk-exports("back-link") {
 
+  $backlink-border-colour: $govuk-black !default;
```
nickcolley: Should these be namespaced with `govuk`?

36degrees: If we're expecting them to be referenced from outside of the component (e.g. by users customising the colours of a button before importing Frontend) then I think they probably should.

Is it worth holding off on doing this work until we've established our public API?

nickcolley: If these are not part of the public api we should consider not allowing them to be overridden?

igloosi: i'll stop then

PR #697, opened 2018-05-16
DO NOT MERGE: Inverted button modifier

dashouse:
First attempt at making a configurable inverted button

Not expecting a merge, just putting here for visibility / place to grab the code as we may use in the design system itself.

Sets variables inside the component

```
  $govuk-button-inverted-background-colour: $govuk-white;
  $govuk-button-inverted-foreground-colour: $govuk-blue;
  $govuk-button-inverted-shadow-colour: shade($govuk-button-inverted-foreground-colour, 30%);
  $govuk-button-inverted-icon-image: "icon-pointer--blue.png";
  $govuk-button-inverted-icon-image-2x: "icon-pointer--blue-2x.png";
```
screen shot 2018-05-16 at 14 11 09

screen shot 2018-05-16 at 14 14 40

Swapping out the foreground and background variables for different colours will also change the hover state and the box-shadow.

screen shot 2018-05-16 at 14 11 40

screen shot 2018-05-16 at 14 16 56

I have also added two new mixins from https://github.com/thoughtbot/bourbon to allow `tint()` and `shade()` in SCSS.

The results of `tint()` and `shade()` are much more desirable than `lighten()` and `darken()` and the effect of the percentage is more consistent when swapping colours.

Ideally the pointer icon would also be affected by setting the variable, but for now I just created blue and green icons until it the icon is recreated in CSS.


dashouse:
Closing, should have done it on the DS first.

PR #699, opened 2018-05-16
Remove normalize from /dist builds

### LINKED ISSUE #679: Remove normalize from dist builds
We current have normalize appended to dist CSS builds, this is not ideal since we do not test for these globals.

https://github.com/alphagov/govuk-frontend/blob/e5e2a5d308f60358fc203e9cf8b08b413cd943da/tasks/gulp/compile-assets.js#L46

nickcolley:
Normalize adds globals that we are not testing for.

Keep the /dist builds consistent with what is in npm by removing this layer.

Fixes #679

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ Breaking changes:
   - creating a new package.json
   - updating documentation
   - update any tests and build pipelines to support this
-  
+
   The single package will be published under a new name.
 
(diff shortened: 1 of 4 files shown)


36degrees:
Why can't this be merged until the release? It won't have any effect until we `build:dist` right?

nickcolley:
Sure, was just trying to make sure I didn't block the release. Can go in.

PR #707, opened 2018-05-21
Fix body element rendering with no text

### LINKED ISSUE #558: Panel component - Template / Macro
## Title

`titleText` adds content inside an `h2` with the class of `govuk-c-panel__title`

`titleHtml` also adds content inside the`h2`, meaning you cannot actually add HTML or add a class to this `h2`. If you try you end up with an empty `h2` above or around your new HTML.

## Empty div

The `text` argument is not conditional, therefore if a user wants to add only a title to their panel (it happens) they will end up with an empty div `govuk-c-panel__body`.

**Current code**

```
<div class="govuk-c-panel govuk-c-panel--confirmation
  {%- if params.classes %} {{ params.classes }}{% endif %}"
  {%- for attribute, value in params.attributes %} {{attribute}}="{{value}}"{% endfor %}>
  <h2 class="govuk-c-panel__title">
    {{ params.titleHtml | safe if params.titleHtml else params.titleText }}
  </h2>
  <div class="govuk-c-panel__body">
    {{ params.html | safe if params.html else params.text }}
  </div>
</div>
```

Suggested change (along the lines of)

```
<div class="govuk-c-panel
  {%- if params.classes %} {{ params.classes }}{% endif %}"
  {%- for attribute, value in params.attributes %} {{attribute}}="{{value}}"{% endfor %}>
  <h2 class="govuk-c-panel__title">
    {{ params.titleHtml | safe if params.titleHtml else params.titleText }}
  </h2>
  {% if params.html | safe if params.html else params.text %}
    <div class="govuk-c-panel__body">
      {{ params.html | safe if params.html else params.text }}
    </div>
  {% endif %}
</div>
```

I wonder if there is a broader review of components to check for other situations where empty divs are created when a user removes a component argument?

nickcolley:
Fixes #558

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ Fixes:
 - Update component readme files content
   ([PR #704](https://github.com/alphagov/govuk-frontend/pull/704))
 
+- Fix body element rendering with no text
+  ([PR #707](https://github.com/alphagov/govuk-frontend/pull/707))
+
 New features:
 
 - Add override classes to set `display` property to `block`, `inline` and `inline-block` (PR [#694](https://github.com/alphagov/govuk-frontend/pull/654))
--- a/src/components/panel/README.md
+++ b/src/components/panel/README.md
@@ -20,9 +20,11 @@ Find out when to use the Panel component in your service in the [GOV.UK Design S
       <h2 class="govuk-panel__title">
         Application complete
       </h2>
+
       <div class="govuk-panel__body">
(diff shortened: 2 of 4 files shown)


nickcolley:
titleHtml and titleText are both required params so I think it makes sense to always output them.

igloosi:
yeah, what Dave is saying if you use titleHTML you can't there is no way to specify classes as we're rendering it inside a predefined class. 
anywho, fixes main issus

PR #708, opened 2018-05-22
Bump jest timeout to 10s

nickcolley:


nickcolley:
Looks like even at 10s it still failed, which makes me think maybe something is hanging and not just slow.

nickcolley:
Closing this will put a card into the backlog to investigate this properly.

PR #709, opened 2018-05-22
Add inset text component

timpaul:
Inset text exists in GOV.UK Elements. We had originally decided not to reimplement it in GOV.UK Frontend as it was unclear that it was meeting user needs. However, during private beta it's become clear that inset text is needed.

[Trello card](https://trello.com/c/R1LbbBu6)


alex-ju:
Update:
- added component attributes
- added tests

igloosi:
📛 Call-out (text)

Issue #717, opened 2018-05-24
Allow checkbox items to have explicit names


solidgoldpig:
I would like the ability to set the names of checkbox items explicitly rather than have them all share the same name.

This would allow me not to have to map values from one variable to another at the backend, but also to target the checkbox items on the clientside using the name (the latter being of less importantance since I could always pass an explicit id to achieve much the same effect)

Proposed code changes here:

https://github.com/solidgoldpig/govuk-frontend/compare/master...checkboxes-items-can-have-explicit-names

igloosi:
Hi @solidgoldpig 
thanks for raising the issue. We're always looking to get the community's feedback as well as getting the community involved in the contribution.

If you're willing to raise a pull request for us to review it as a team, that'd be great.

igloosi:
Thanks. 
We'll aim to release this next week.

igloosi:
Hi @solidgoldpig,
this has now been released in [v0.0.30-alpha](https://github.com/alphagov/govuk-frontend/releases/tag/v0.0.30-alpha), so I will close the issue.

PR #719, opened 2018-05-24
Add ability to pass explicit names for checkbox items

solidgoldpig:
https://github.com/alphagov/govuk-frontend/issues/717

Also add test for explicitly passed ids for checkbox items


src/components/checkboxes/template.test.js
```diff
@@ -151,6 +151,59 @@ describe('Checkboxes', () => {
       expect($lastLabel.attr('for')).toEqual('custom-2')
     })
 
+    it('render explicitly passed item ids', () => {
```
nickcolley: 👏

PR #720, opened 2018-05-24
Ensure 'after' tests are not ignored

nickcolley:
Unblocks releases, `after` tests are only used to test the result of other tasks, for example
building the `dist/` folder.


igloosi:
changelog 🔏

PR #726, opened 2018-05-29
Use new font loading strategy

36degrees:
This is mirroring the work done by @Nooshu and @steventux in GOV.UK Template (alphagov/govuk_template#341):

> This PR has been raised to solve one of the major frontend performance issues that is currently live on GOV.UK. Four WOFF compressed web-fonts are base64 encoded and are sitting on the browsers critical path. This 200KB CSS file must be downloaded and parsed before page rendering can start. This single download is approximately 82% of the render blocking assets which stop the browser showing any content to the user.
>
> Base64 encoding into a CSS file is now considered an anti-pattern so the font loading method is being revisited.
>
> This PR removes the Base64 encoded font and reverts to a standard HTTP request to load the font files. This gives us the added advantage of being able to offer multiple font sources. The browser can then choose the most appropriate supported font format.
>
> The fonts being used are unchanged. We have created a new WOFF2 version of the font which offers approximately 25% better compression and no change in quality.
>
> This PR is also taking advantage of a new CSS property that is available in modern browsers: font-display. This property defines how the browser behaves during the font download. Browsers that don't support this property will ignore it.

Further details including an overview of the performance impact can be found in the original PR.

Additionally, this PR also:

- Renames the existing `govuk-file-url` helper to `govuk-image-url`, and adds a new `govuk-font-url` helper.
- Renames the existing `$govuk-global-images` to `$govuk-images-path` and adds two new variables – `$govuk-fonts-path` and `$govuk-assets-path`, which is used as the base for both `$govuk-images-path` and `$govuk-fonts-path`.


36degrees:
Will wait for #725 to merge and rebase to avoid conflicts before adding CHANGELOG entries.

36degrees:
@nickcolley happy with 83b9e53?

Nooshu:
Awesome work @36degrees! Very happy to see this in Frontend!

PR #731, opened 2018-05-29
Add boilerplate template

nickcolley:
I have tested this by integrating the template within our own review application, and in the [GOV.UK Design System](https://github.com/alphagov/govuk-design-system/pull/288).

## [Template area documentation](https://govuk-frontend-review-pr-731.herokuapp.com/examples/template-block-areas)

Changes from [govuk_template](https://github.com/alphagov/govuk_template):

Breaking changes:
- Removed stylesheets relating to `govuk_template.css`, we now recommend people bundle stylesheets in their application.
- Removed stylesheets relating to `print`, we do this at a component level.
- Removed stylesheets relating to `font`, we now have a [new font loading strategy](https://github.com/alphagov/govuk-frontend/pull/726) that does not require a different bundle.
- Removed `ie8` specific class being added to the `html` element, since we don't require functionality
- Remove `ie.js`, used to shim HTML5 files, add JSON2 support, we now recommend people bundle in their application.

- Removed functionality to remove `js-enabled` class if `window.GOVUK` is not defined, at the moment we can't guarantee that this global will be set with GOV.UK Frontend.
See https://github.com/alphagov/govuk_template/pull/248 for details on this feature.

Features:
- Added a headIcon block, which allows you to override icons, useful for unbranded layouts
- Allow changing the theme-color, useful for unbranded layouts.
- Added a main block, now skip link works by default, this block can be overridden.

Patch:
- Moved Skip link into it's own block, calling the [Skip link component](https://github.com/alphagov/govuk-frontend/tree/master/src/components/skip-link) by default instead of being inlined.
- Moved footer into it's own block, calling the [Footer component](https://github.com/alphagov/govuk-frontend/tree/master/src/components/footer) by default instead of being inlined.
- Moved header into it's own block, calling the [Header component](https://github.com/alphagov/govuk-frontend/tree/master/src/components/header) by default instead of being inlined.
- Changed the default `main` `id` to be `#main-content` to avoid being [styled by GOV.UK Elements](https://github.com/alphagov/govuk_elements/blob/8216538c1d2efb8b33372a22f28e1ea81d879ecd/assets/sass/elements/_layout.scss#L9).

https://trello.com/c/UcWg3ro2/290-epic-allow-a-user-to-set-up-an-app-without-govuk-template-eg-boilerplate


nickcolley:
GOV.UK Template sets a grey background so the footer does not cut off.

We could do the same, or do something like a box-shadow, ::after element that extends to the bottom of the page.

Alternatively, template could have it's own CSS?

screen shot 2018-05-30 at 13 26 46

nickcolley:
I'm planning on moving the template areas example into the design system when we document this for consumption.

src/template.njk
```diff
@@ -0,0 +1,59 @@
+{% from "./components/skip-link/macro.njk" import govukSkipLink %}
+{% from "./components/header/macro.njk" import govukHeader %}
+{% from "./components/footer/macro.njk" import govukFooter %}
+
+{% block pageStart %}{% endblock %}
+<!DOCTYPE html>
+<html lang="{{ htmlLang | default('en') }}" class="govuk-template">
```
igloosi: how does one extend this to add custom classes or ie specific classes?

nickcolley: Makes sense, I'll add that feature

igloosi:
changelog 📓

src/template.njk
```diff
@@ -0,0 +1,59 @@
...
+<!DOCTYPE html>
+<html lang="{{ htmlLang | default('en') }}" class="govuk-template {{ htmlClasses }}">
+  <head>
+    <meta charset="utf-8" />
+    <title>{% block pageTitle %}GOV.UK - The best place to find government services and information{% endblock %}</title>
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <meta name="theme-color" content="{{ themeColor | default('#0b0c0c') }}" /> {# Hardcoded value of $govuk-black #}
+
+    {% block headIcons %}
+      <link rel="shortcut icon" href="{{ assetPath | default('assets/') }}images/favicon.ico" type="image/x-icon" />
```
36degrees: I think all of the default asset paths should be `/assets`, with a leading slash so they resolve from the domain root. For example, at the minute the paths in the example resolve to e.g. `/examples/assets/images/favicon.ico`

app/views/examples/template-custom/index.njk
```diff
@@ -0,0 +1,124 @@
...
+          "href": "#1",
+          "text": "Aidez-moi"
+        },
+        {
+          "href": "#2",
+          "text": "Politique de confidentialité"
+        },
+        {
+          "href": "#3",
+          "text": "Contactez nous"
```
alex-ju: Not a blocker: Hyphenate (Contactez-nous) as in Aidez-moi

app/views/examples/template-custom/index.njk
```diff
@@ -0,0 +1,124 @@
...
+{% endblock %}
+
+{% block main %}
+  <!-- block:main -->
+  {{ govukPhaseBanner({
+    classes: "govuk-width-container",
+    tag: {
+      text: "alpha"
+    },
+    html: 'C\'est un nouveau service - vos <a class="govuk-link" href="#"> commentaires </a> nous aideront à l\'améliorer.'
```
alex-ju: Not a blocker: the anchor element has extra spaces before and after the text. If you update don't forget to update the test too.

Issue #724, opened 2018-05-29
Namespace all SCSS exports with `govuk`


nickcolley:
We have found that an application using the same export as us, it will conflict resulting in a broken build.

For example, when releasing the Header component, it conflicted with the application level export.

We should make sure our exports are namespaced.

https://github.com/alphagov/govuk-frontend/blob/master/src/components/button/_button.scss#L5

PR #732, opened 2018-05-30
Namespace exports with 'govuk'

igloosi:
We have found that an application using the same export as us, it will conflict resulting in a broken build.

For example, when releasing the Header component, it conflicted with the application level export.

This fixes: #724 

Trello ticket: https://trello.com/c/UaX5dWyO/1025-namespace-all-scss-exports-with-govuk

### DIFF
--- a/src/components/back-link/_back-link.scss
+++ b/src/components/back-link/_back-link.scss
@@ -2,7 +2,7 @@
 @import "../../tools/all";
 @import "../../helpers/all";
 
-@include govuk-exports("back-link") {
+@include govuk-exports("govuk/component/back-link") {
 
   .govuk-back-link {
     @include govuk-typography-responsive($govuk-font-16);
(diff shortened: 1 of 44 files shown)


36degrees:
Thought – do we want to standardise on how we construct these strings to avoid conflicts across layers?

e.g. `govuk/component/button`, `govuk/helpers/typography`, `govuk/overrides/typography`

igloosi:
@36degrees can do, i see no issue with it. will amend

src/components/back-link/_back-link.scss
```diff
@@ -2,7 +2,7 @@
 @import "../../tools/all";
 @import "../../helpers/all";
 
-@include govuk-exports("back-link") {
+@include govuk-exports("govuk/component/back-link") {
```
nickcolley: 'component' -> 'components'

igloosi: I'd argue for the use of singular

36degrees: It's a bit weird having component, but helper*s*, object*s*, utilitie*s*. But probably not going back and changing now.

nickcolley: It's not part of the public api, so not too much trouble

PR #733, opened 2018-05-30
Allow custom image and font URL helpers

### LINKED ISSUE #730: Using the Rails asset pipeline for images
In Rails we use helpers to refer to assets in CSS:

```css
background-image: image-url("icon-pointer.png");
background-image: asset-url("icon-pointer.png");
```

This makes the [asset pipeline in Rails](http://guides.rubyonrails.org/asset_pipeline.html#css-and-sass) work: instead of linking to `icon-pointer.png`, it will link to the [fingerprinted version](http://guides.rubyonrails.org/asset_pipeline.html#what-is-fingerprinting-and-why-should-i-care-questionmark) of the file.

Currently GOV.UK Frontend [just uses url("file.png") in the CSS](), so Rails won't point the correct assets:

https://github.com/alphagov/govuk-frontend/blob/5bb66d62c2f7ad5fb96363f9d54121b510b32ec1/package/tools/_file-url.scss#L4 

For GOV.UK to use GOV.UK Frontend, we'll need a way to override the `govuk-file-url` or to instruct it to use `asset-url`.

### ~Defining the function first~

I thought maybe defining the `govuk-file-url` function above the `@import` for GOV.UK Frontend would work, but the later definition overrides this. 

### ~Defining the function last~

This doesn't work because `govuk-file-url` is already used by the CSS.

### ~Conditionally defining the function~

I though maybe if we wrap the function in a conditional like below, our override might work. 

```scss
@if function-exists('govuk-file-url') == false {
  @function govuk-file-url($file) {
    @return url($govuk-global-images + $file);
  }
}
```

Unfortunately SCSS errors out with:

```
Functions may not be defined within control directives or other mixins.
```

### Conditional inside function

This works:

```scss
$govuk-use-rails: true;

@function govuk-file-url($file) {
  @if $govuk-use-rails {
    @return asset-url($file);
  } else {
    @return url($govuk-global-images + $file);
  }
}

```

---

This follows a spike into using GOV.UK Frontend on GOV.UK (https://github.com/alphagov/govuk_publishing_components/pull/328).

36degrees:
Allow users to set custom functions to resolve image and font urls.

Proposing this as a solution to #730.

Closes #730


36degrees:
I'd expect it to be documented as part of https://github.com/alphagov/govuk-frontend/blob/doc-updates/docs/installation/installing-with-npm.md#import-images, but waiting for that branch to be merged first.

36degrees:
Going to merge this and we can add the docs later.

PR #737, opened 2018-05-30
Fix a couple of IE8 issues

36degrees:
## Buttons

Fix missing button shadow by using a bottom-border in IE8.

### Before
bs_winxp_ie_8 0 2

### After
bs_winxp_ie_8 0 3

## Breadcrumbs

Fix missing chevrons by falling back to any sans-serif font. IE8 [seems to struggle to render pseudo-elements that use a font defined as a `@font-face`](https://www.google.co.uk/search?q=ie8+pseudo+element+font+face).

### Before
bs_winxp_ie_8 0 4

### After
bs_winxp_ie_8 0 5

### DIFF
--- a/src/components/breadcrumbs/_breadcrumbs.scss
+++ b/src/components/breadcrumbs/_breadcrumbs.scss
@@ -90,6 +90,10 @@
         height: auto;
         border: 0;
         color: $chevron-border-colour;
+
+        // IE8 doesn't seem to like rendering pseudo-elements using @font-faces,
+        // so fall back to using another sans-serif font to render the chevron.
+        font-family: Arial, sans-serif;
       }
     }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,11 @@ Note: We're not following semantic versioning yet, we are going to talk about th
 - Removed some duplicated CSS rules from the outputted CSS
   ([PR #727](https://github.com/alphagov/govuk-frontend/pull/727))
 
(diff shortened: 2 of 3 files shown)


src/components/breadcrumbs/_breadcrumbs.scss
```diff
@@ -90,6 +90,10 @@
         height: auto;
         border: 0;
         color: $chevron-border-colour;
+
+        // IE8 doesn't seem to like rendering pseudo-elements using @font-faces,
+        // so fall back to using any sans-serif font to render the chevron.
+        font-family: sans-serif;
```
nickcolley: Is this safe? Browser do weird things, I wonder if picking something stable like helvetica/arial would be good to start with.

igloosi: that would be forcing either a default instead of browser/os default combo

That said, our font stack definition is `$govuk-nta-light: "nta", Arial, sans-serif;`, so by going by that we either do `sans-serif` or `Arial, sans-serif`

Not sure how up to date this table is https://www.granneman.com/webdev/coding/css/fonts-and-formatting/web-browser-font-defaults/

nickcolley: Yeah, was thinking we use the current font stack minus nta

36degrees: Updated to use `Arial, sans-serif`.

Issue #714, opened 2018-05-22
Should label in Nunjucks be wrapped in a condition?


igloosi:
The following code
```
{{ govukInput({
   "id": "national-insurance-number",
   "name": "national-insurance-number",
 }) }}
```
will output
```
<div class="govuk-form-group">
  <label class="govuk-label" for="govuk-input-a">
    
  </label>
  <input class="govuk-input" id="govuk-input-a" name="national-insurance-number" type="text">
</div>
```
Empty label, that has a 5px bottom margin, which will effect any alignment

In the input component template we currently  don't wrap label in a condition, like we do hint and error message.

nickcolley:
Ideally not having a label input would result in a hard failure with a visible error.

If anything we could consider doing something like

```css
.govuk-label:empty::after {
  content: "Warning: No label for input";
  color: red;
  font-weight: bold;
}
```

joelanman:
Isn't it valid to provide a label separately from the input macro? We provide a label macro

igloosi:
@nickcolley label has whitespace so :empty won't work

nickcolley:
One of the best things about out input component is that we make it hard to forget an associated label by default.

If we can do this without compromising that 👍

joelanman:
It is valid to provide a label separately, so I think it should be possible not have a label in this macro.

We could possibly require `label: false` from the user. If label is simply missing, it could print an error (we can just use the macro to do this directly, not CSS). However this is a point on our general approach - do we show a similar error in any other macro?

nickcolley:
@joelanman `label: false` is great since it makes sure the user is making the decision.

Having general errors also sounds useful.

@igloosi was this raised based on an issue you noticed?

joelanman:
User-facing errors worries me a bit - on a dynamic site, a macro could theoretically be called that shows the end user an error that isn't for them - it's confusing and they can do nothing about it.

nickcolley:
Sure, this could be a development only thing.

PR #740, opened 2018-05-31
Omit empty labels

36degrees:
Fixes https://github.com/alphagov/govuk-frontend/issues/714

This ensures that the label element is not outputted if no text or HTML are provided. This allows for other form control components to be used with a separate label, e.g. if a user wants to have an input and a button on one line with a separate label above.

An example of where this is problematic is on the form-alignment page, where an empty label is included above each input:

## Before
```html
<fieldset class="govuk-fieldset">
    <label class="govuk-label" for="govuk-input-a">
        National Insurance number
    </label>

    <div class="app-u-w-half">
            <label class="govuk-label" for="govuk-input-a">

            </label>
            <input class="govuk-input" id="govuk-input-a" name="national-insurance-number" type="text">
        </div>

    </div>
    <div class="app-u-w-half">
        <button type="submit" class="govuk-button">
            Save and continue
        </button>
    </div>
</fieldset>
```
screen shot 2018-05-31 at 13 45 45bst

## After
```html
<fieldset class="govuk-fieldset">
    <label class="govuk-label" for="govuk-input-a">
        National Insurance number
    </label>

    <div class="app-u-w-half">
            <input class="govuk-input" id="govuk-input-a" name="national-insurance-number" type="text">
        </div>

    </div>
    <div class="app-u-w-half">
        <button type="submit" class="govuk-button">
            Save and continue
        </button>
    </div>
</fieldset>
```
screen shot 2018-05-31 at 13 45 48bst

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 Note: We're not following semantic versioning yet, we are going to talk about this soon.
 
+## Unreleased
+
+🔧 Fixes:
+
+- The `<label>` element will now be omitted for form controls where no label
+  text or html is provided. If you call the label component directly whilst
+  passing neither text nor html, no HTML will be outputted.
+  ([PR #740])(https://github.com/alphagov/govuk-frontend/pull/740)
+
 ## 0.0.31-alpha (Breaking release)
 
 💥 Breaking changes:
--- a/src/components/label/template.njk
+++ b/src/components/label/template.njk
@@ -1,3 +1,4 @@
+{% if params.html or params.text %}
 {% set labelHtml %}
 <label class="govuk-label{%- if params.classes %} {{ params.classes }}{% endif %}"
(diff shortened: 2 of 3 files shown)


36degrees:
Alternative approach would be update every component that uses the label (input, radios, checkboxes, textarea, file-upload, and select) to only call the label component if `params.label.html` or `params.label.text` is provided.

This would be consistent with [how we do error messages and hints](https://github.com/alphagov/govuk-frontend/blob/bfba045c082d907c409df9818f8e900017bae8cc/src/components/input/template.njk#L17-L37), but in retrospect it feels like the less components 'know about' the internals of other components (e.g. the fact they accept `html` and `text`) the better… thoughts? Another question is then whether we should make the error message and hint components work in a similar way.

36degrees:
Aah, of course we still need to know whether the error message and hint are going to be displayed within each form component so that we know whether to include their IDs in `aria-describedby`, which is _not_ the case for labels. So perhaps this does make some sense.

PR #741, opened 2018-05-31
Fix issues with canvas colour bleeding into main review page

nickcolley:
Sets the canvas colour to the body colour on review app pages, doesnt impact the template examples which do not inherit the _generic template.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,9 @@ Note: We're not following semantic versioning yet, we are going to talk about th
 - Reintroduce mistakenly deleted HTML5Shiv required for IE8
   ([PR #749](https://github.com/alphagov/govuk-frontend/pull/749))
 
+- Fix issues with canvas colour bleeding into main review page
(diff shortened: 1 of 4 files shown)


nickcolley:
Note that the reason it's doing this is because most of our containers allow overflow.

Another solution would look like:

```css
.govuk-template, .govuk-template__body {
    height: 100%;
}

/* Clearfix for containers */
.govuk-width-container::before,
.govuk-width-container::after {
    content: ' ';
    display: table;
    clear: both;
}

```

igloosi:
grey bg on variant preview pages...
why dont we give the `<html>` a custom app class inside _generic view and reset background colour, to fix all component pages?
template example pages inherit the template directly so they won't be affected

nickcolley:
@igloosi sure, but that felt like a hack, I thought we'd rather have a bit of gray and know how template really works, than having app specific overrides.

Willing to go with the override approach though if that's what people want.

PR #748, opened 2018-06-01
Document (non-contentious parts of) Sass settings layer

36degrees:
This is the first step in clarifying what comprises the public API of Frontend.

This adds Sassdocs for the following files:

- `settings/_assets.scss`
- `settings/_colours-applied.scss`
- `settings/_compatibility.scss`
- `settings/_global-styles.scss`
- `settings/_ie8.scss`
- `settings/_measurements.scss`
- `settings/_media-queries.scss`
- `settings/_typography-font-families.scss` (renamed from `_typography-font-stacks.scss`)
- `settings/_typography-font.scss`

That leaves some files in settings which are not using Sassdoc:

- `settings/_colours-organisations.scss`
- `settings/_colours-palette.scss`
- `settings/_spacing.scss`
- `settings/_typography-responsive.scss`

I've left these (for now) because they all contain a large number of variables, and we've discussed moving them to maps and adding helper functions which will give us more control over e.g. deprecating colours.

As part of these changes we've also:

- Defined private or public access for all settings
- Prefixed private variables with an underscore
- Made settings `!default` where it makes sense to do so
- Moved the button variables into the button component
- Refered to font families rather than font stacks consistently
- Renamed nta-light to font-family-nta
- Correctly used abstracted `$govuk-font-family-tabular` for tabular number mixins, rather than referencing nta-light-tabular directly
- Abstracted away `sass-mq`, and expose our own settings in our public api

https://trello.com/c/B8OCSU8R/759-agree-on-and-document-public-api-for-govuk-frontend

### DIFF
--- a/src/helpers/grid.test.js
+++ b/src/helpers/grid.test.js
@@ -14,11 +14,15 @@ const sassConfig = {
 }
 describe('grid system', () => {
   const sassImports = `
-    @import "tools/exports";
+    @import "settings/ie8";
+    @import "settings/media-queries";
     @import "settings/spacing";
     @import "settings/measurements";
+
     @import "helpers/grid";
     @import "helpers/media-queries";
+
+    @import "tools/exports";
   `
   describe('grid-width function', () => {
     it('outputs the specified key value from the map of widths', async () => {
@@ -115,7 +119,7 @@ describe('grid system', () => {
           box-sizing: border-box;
           width: 100%;
           padding: 0 15px; }
-          @media (min-width: 46.25em) {
+          @media (min-width: 40.0625em) {
             .govuk-grid-column-full {
               width: 100%;
               float: left; } }`)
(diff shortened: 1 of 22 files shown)


src/helpers/grid.test.js
```diff
@@ -115,7 +119,7 @@ describe('grid system', () => {
           box-sizing: border-box;
           width: 100%;
           padding: 0 15px; }
-          @media (min-width: 46.25em) {
+          @media (min-width: 40.0625em) {
```
alex-ju: What's the reason behind the change of these values?

igloosi:
this looks solid, happy to approve it now, but i guess others should have a look ( as they've been involved in the discussion)

src/helpers/grid.test.js
36degrees: Good question - it used to use the default `sass-mq` breakpoints, now it's using the ones defined in our settings.

alex-ju: Could we add this to the PR comments?

36degrees: Done.

36degrees:
Argh, CHANGELOG.

PR #763, opened 2018-06-05
Wrap sass-mq as govuk-media-query

36degrees:
‘Abstract away’ sass-mq so that we don’t expose it as part of our public API, which means we’d be able to remove it in the future if we wanted to roll our own or move to a different vendored solution.

I've added tests for the `govuk-media-query` mixin based off the existing tests for sass-mq, but I've removed the tests for the sass-mq functions that we're now not calling directly, as I don't think it makes sense to test the 'internals' of a vendored library when we can test its edges instead.


src/helpers/_media-queries.scss
```diff
@@ -13,10 +21,9 @@ $mq-show-breakpoints: ();
 // When building a stylesheet for IE8, set $mq-responsive to false in order to
 // 'rasterize' any media queries.
 
+$mq-responsive: true;
```
36degrees: Moving this because the way Sass variable scoping works means that unless `$mq-responsive` is defined before this point, the `$mq-responsive` declaration only works within the scope of the `@if` – which caused a lot of headaches in the tests.

src/helpers/_media-queries.scss
```diff
@@ -1,4 +1,12 @@
...
+////
+
+
+
+// =========================================================
+// Wrangle sass-mq config...
+// =========================================================
 
 // Pass our breakpoints and static breakpoint definitions through to sass-mq.
 $mq-breakpoints: if(variable-exists(govuk-breakpoints), $govuk-breakpoints, ());
```
36degrees: Because we do this configuration 'at import time', any changes to `$govuk-is-ie8`, `$govuk-breakpoints`, `$govuk-ie8-breakpoint` made after the import has been done will not have any effect on `govuk-media-query`. We _could_ move this wrangling into the `govuk-media-query` mixin, but it complicates the way we pass arguments somewhat, and relies on us being able to pass that config to `mq` which is an undocumented feature, and I can't think of any valid use cases where you'd want to change that config mid-way through your codebase.

Thoughts?

igloosi:
one think the abstraction is missing is adding new media-queries (somewhere up to desktop where we want them to rasterise for ie8)

you can still ad them via `@include mq-add-breakpoint(whatever, size);` but it might be a weird mix of govuk-media-query and this syntax

36degrees:
Is that something we think people actually want to use? I don't think we need to expose everything `mq` allows unless we think it's actually needed.

It's still possible to override `$govuk-breakpoints` to add new breakpoints globally?

igloosi:
true. people might not do it that way / know of it. most likely will just add it to the breakpoints map

36degrees:
One thing to consider is that if in the future we wanted to move away from using sass-mq we'd have to implement everything we've exposed as part of our API – if we implement every feature that mq offers, we're committing to providing those features or we'd be forced to introduce a breaking change to remove them.

36degrees:
@igloosi did you mean to mark that as 'request changes'?

PR #766, opened 2018-06-05
Fix govuk-equilateral-height function usage in shape-arrow helper

igloosi:
In #762 we changed the name of the `govuk-equilateral-height` function to `_govuk-equilateral-height` to signify that is is a private function.

Unfortunately we forgot to update the name where it was used to create the open/closed indicator for the details component.

This PR fixes this.

Before:
screen shot 2018-06-05 at 19 39 00

After: 
screen shot 2018-06-05 at 19 39 41

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,9 @@ Note: We're not following semantic versioning yet, we are going to talk about th
 
 🔧 Fixes:
 
+- Fix govuk-equilateral-height function usage in shape-arrow helper
+  ([PR ##766](https://github.com/alphagov/govuk-frontend/pull/766))
+
 - The `<label>` element will now be omitted for form controls where no label
   text or html is provided. If you call the label component directly whilst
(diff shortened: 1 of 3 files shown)


36degrees:
Argh, good catch - thanks for that. I think I understand why, but it's pretty annoying that Sass silently outputs the function call verbatim. I wonder if there's a way we could try and catch that if it happens.

igloosi:
i guess only if we wrap it in a custom mixin that check if function exist using sass's built-in `function-exists`

alternatively, if we use `get-function("myfunc")` when calling a function, that will throw an error in compilation

36degrees:
Added a test that greps the CSS for anything matching `/govuk-[\w-]+\(.*?\)/g`, which _should_ catch most function calls that slip through the net… unless they misspell govuk…

igloosi:
fixed the extra semicolon in the test

36degrees:
Added a big ole' comment around what the test is looking for

PR #769, opened 2018-06-06
Update components' tables of arguments

igloosi:
Ensuring table of arguments for each components is accurate.

Updates to  `index.njk` and `README.md` files for each component.

Trello ticket: https://trello.com/c/rRQl78KY/1038-make-sure-all-component-argument-tables-are-accurate

For reviewers: Might be easiest to go through each component page on Github in the [branch](https://github.com/alphagov/govuk-frontend/tree/update-tables-of-arguments) to check the rendered README file.

### DIFF
--- a/src/components/checkboxes/index.njk
+++ b/src/components/checkboxes/index.njk
@@ -81,20 +81,6 @@
     ],
     [
       {
-        text: 'classes'
-      },
-      {
-        text: 'string'
-      },
-      {
-        text: 'No'
-      },
-      {
-        text: 'Optional additional classes'
-      }
-    ],
-    [
-      {
         text: 'idPrefix'
       },
       {
@@ -104,7 +90,7 @@
         text: 'No'
       },
       {
-        text: 'String to prefix id for each checkbox item if no id is specified on each item.'
+        text: 'String to prefix id for each checkbox item if no id is specified on each item. If`idPrefix` is not passed, fallback to using the name attribute instead.'
       }
     ],
     [
@@ -132,12 +118,26 @@
         text: 'Yes'
       },
       {
-        text: 'Array of checkbox items.'
+        text: 'Array of checkbox items objects.'
       }
     ],
     [
       {
-        text: 'text'
+        text: 'items.{}.text (or) items.{}.html'
+      },
+      {
+        text: 'string'
+      },
+      {
+        text: 'Yes'
+      },
+      {
+        text: 'Text or HTML to use within each checkbox item label. If `html` is provided, the `text` argument will be ignored.'
+      }
+    ],
+    [
+      {
+        text: 'items.{}.id'
       },
       {
         text: 'string'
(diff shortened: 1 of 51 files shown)


src/components/footer/index.njk
```diff
@@ -75,12 +103,12 @@
...
       },
       {
-        text: 'Array of items for use in the navigation section of the footer'
+        text: 'Array of items for use in the navigation section of the footer.'
       }
     ],
     [
       {
-        text: 'navigation[].title'
+        text: 'navigation.title'
```
nickcolley: Should this be consistent with the with the other approach that uses `{}.prop`?

igloosi: updated, but it's tricky as its [].{}.prop - don't want to complicate it too much

src/components/footer/index.njk
```diff
@@ -117,7 +145,35 @@
...
+      {
+        text: 'No'
+      },
+      {
+        text: 'List item text in the navigation section of the footer.'
+      }
+    ],
+    [
+      {
+        text: 'navigation.items.[].href'
```
nickcolley: I think this is still inconsistent with other places using 'item.{}.prop'?

igloosi: yep, fixed it now

src/components/checkboxes/index.njk
```diff
@@ -202,7 +230,21 @@
         text: 'No'
       },
       {
-        text: 'If true, checkbox will be disabled'
+        text: 'If true, checkbox will be disabled.'
+      }
```
hannalaakso: I think we might need `items.{}.value` as well.

src/components/date-input/index.njk
```diff
@@ -133,6 +133,34 @@
       {
         text: 'Arguments for the fieldset component (e.g. legend). See fieldset component.'
       }
+    ],
+    [
```
hannalaakso: I think we might need `hint` object as well.

src/components/fieldset/index.njk
```diff
@@ -67,7 +67,7 @@
     ],
     [
       {
-        text: 'legend.text'
+        text: 'legend.text (or) legend.html'
```
hannalaakso: These should probably be `legend.{}.text` etc. to be consistent.

src/components/footer/index.njk
```diff
@@ -117,7 +145,35 @@
...
+        text: 'navigation.items.{}.href'
+      },
+      {
+        text: 'string'
+      },
+      {
+        text: 'No'
+      },
+      {
+        text: 'List item href atrribute in the navigation section of the footer. Both `text` and `href` attributes need to be present to create a link.'
```
hannalaakso: Nit: typo in "attribute".

src/components/radios/index.njk
```diff
@@ -204,6 +232,20 @@ Let users select a single option from a list.
         text: 'If true, radio will be disabled.'
       }
     ],
+    [
```
hannalaakso: I think we need `items.{}.value` as well.

src/components/select/index.njk
```diff
@@ -150,7 +136,7 @@ The HTML <code>&lt;select&gt;</code> element represents a control that provides
     ],
     [
       {
-        text: 'label'
+        text: 'item.label'
```
hannalaakso: Hmmm should this be just `label`? It doesn't look like `items` uses `label` in the template.

src/components/table/index.njk
```diff
@@ -104,40 +118,54 @@
...
-        text: 'string'
+        text: 'number'
+      },
+      {
+        text: 'No'
+      },
+      {
+        text: 'Specify how many columns a cell extends.'
+      }
+    ],
```
hannalaakso: I think this might also need `head.[].format`.

src/components/footer/index.njk
igloosi: can't see the typo

hannalaakso: It's in "atrribute" - sorry should have said that.

src/components/radios/index.njk
igloosi: added

src/components/date-input/index.njk
igloosi: added

igloosi:
@hannalaakso updated with comments

PR #772, opened 2018-06-07
Refactor typography mixin; remove shorthand helpers

36degrees:
The shorthand helpers are being removed in favour of a new mixin `govuk-font` throughout - this decouples the codebase from the typography scale (allowing it to be overridden by consumers in the future), reduces the number of mixins in our public API and gives us greater control over being able to e.g. warn if points in the scale are going to be deprecated.

Rather than passing font maps around, pass the 'key' from the scale instead.

These changes have been made whilst running a Jest test that expects against against a known-good snapshot of both the default and IE8 stylesheets from master, so the output should be exactly the same.

https://trello.com/c/B8OCSU8R/759-agree-on-and-document-public-api-for-govuk-frontend


igloosi:
looks good. looking forward to the changelog entry :) 📓

36degrees:
> looks good. looking forward to the changelog entry :) 📓

Changelog entries added!

> Nice work using the snapshots while developing!

It worked _really_ well – trying to think of a way we can turn it into a gulp task e.g. `gulp refactor:start` which would create a (non-committed) snapshot and run jest in watch mode. Maybe something for firebreak…

PR #773, opened 2018-06-07
Refactor organisation colours

36degrees:
- Define organisation colours in a map, rather than as multiple variables, allowing them to be overridden by consumers
- Add tests to ensure that all organisation colours meet colour contrast, based on the tests from Frontend Toolkit (https://github.com/alphagov/govuk_frontend_toolkit/blob/bf34b09a5e0433ef50781c68a0d3180a267d51d9/spec/stylesheets/_colour_contrast_spec.scss)
- Update DEFRA and HMRC organisation colours to match the latest colours in Frontend Toolkit (see https://github.com/alphagov/govuk_frontend_toolkit/pull/368)
- Add a new function govuk-organisation-colour to allow you to easily get an organisation colour

### DIFF
--- a/src/helpers/colour.test.js
+++ b/src/helpers/colour.test.js
@@ -1,0 +1,83 @@
+/* eslint-env jest */
+
+const util = require('util')
+
+const configPaths = require('../../config/paths.json')
+
+const sass = require('node-sass')
+const sassRender = util.promisify(sass.render)
+
+const sassConfig = {
+  includePaths: [ configPaths.src ],
+  outputStyle: 'compact'
+}
+
+const sassBootstrap = `
+  $govuk-colours-organisations: (
+    'floo-network-authority': (
+      colour: #EC22FF,
+      colour-websafe: #9A00A8
+    ),
+    'broom-regulatory-control': (
+      colour: #A81223
+    )
+  );
+
+  @import "helpers/colour";
+`
+
+describe('@function govuk-organisation-colour', () => {
+  it('returns the colour for a given organisation', async () => {
+    const sass = `
+      ${sassBootstrap}
+
+      .foo {
+        background: govuk-organisation-colour('floo-network-authority');
+      }`
+
+    const results = await sassRender({ data: sass, ...sassConfig })
+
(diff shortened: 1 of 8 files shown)


igloosi:
do we need to import `src/helpers/_colour.scss` into `helpers/all.scss` ?

36degrees:
> do we need to import src/helpers/_colour.scss into helpers/all.scss ?

Very good catch!

src/settings/colours.test.js
```diff
@@ -0,0 +1,41 @@
...
+      @import "sass-color-helpers/stylesheets/color-helpers";
+
+      $minimum-contrast: 4.5;
+
+      @each $organisation in map-keys($govuk-colours-organisations) {
+
+        $colour: govuk-organisation-colour($organisation, $websafe: true);
+        $contrast: ch-color-contrast($govuk-body-background-colour, $colour);
+
+        @if ($contrast < $minimum-contrast) {
```
nickcolley: Thanks @fofr :)

src/helpers/colour.test.js
```diff
@@ -0,0 +1,83 @@
...
+
+    expect(results.css.toString().trim()).toBe('.foo { background: #EC22FF; }')
+  })
+
+  it('returns the websafe colour for a given organisation if requested', async () => {
+    const sass = `
+      ${sassBootstrap}
+
+      .foo {
+        color: govuk-organisation-colour('floo-network-authority', $websafe: true);
```
nickcolley: I think websafe should be default true?

We don't want people using the non websafe unless they really need to.

GOV.UK uses the websafe for example.

36degrees: It depends on what you're using it for surely? 'Non-websafe' would be fine for background and border colours? Conceptually, I'd expect the function to return the 'true' / accurate colour by default.

nickcolley: Good defaults are more important than pure concepts, where we can make it harder for people to fail it's always a good idea.

36degrees:
@nickcolley happy with the changelog entry?

PR #776, opened 2018-06-08
Add tabs component

alex-ju:
## Review page
https://govuk-frontend-review-pr-776.herokuapp.com/components/tabs

## Contribution log
https://github.com/alphagov/govuk-design-system-backlog/issues/100#issuecomment-386542535

## Browser and AT testing
<details>
<summary>Chrome 67
</summary>
chrome67
</details>

<details>
<summary>Safari 11</summary>
safari11
</details>

<details>
<summary>Firefox 60 with inverted colours</summary>
ff60
</details>

<details>
<summary>IE11</summary>
ie11
</details>

<details>
<summary>IE10</summary>
ie10
</details>

<details>
<summary>IE9</summary>
ie9
</details>

<details>
<summary>IE8</summary>
ie8
</details>

<details>
<summary>NVDA with Firefox</summary>

_Tab control. 'Past day' tab, selected, 1 of 4._

</details>

<details>
<summary>JAWS with IE11</summary>

_'Past day' tab, selected, 1 of 4._

</details>

<details>
<summary>VoiceOver on Safari</summary>

_'Past day', selected, tab, 1 of 4._

</details>

## Progress
- [x] Update to follow the SCSS styleguide and uses govuk-frontend varibles and mixins
- [x] Define example in the yaml schema
- [x] Add Nunjucks macro and unit tests for the markup
- [x] Update to follow the JavaScript styleguide
- [x] Add polyfills for less-capable browser support or has a good fallback
- [x] Integrations tests for the JavaScript
- [x] Rebase and adjust based on recent govuk-frontend changes (i.e. mixin names and variables)
- [x] Update example
- [x] A new round of AT testing

[Trello card](https://trello.com/c/nMEvCgiT)

### DIFF
--- a/src/all.js
+++ b/src/all.js
@@ -5,6 +5,7 @@ import Checkboxes from './components/checkboxes/checkboxes'
 import ErrorSummary from './components/error-summary/error-summary'
 import Header from './components/header/header'
 import Radios from './components/radios/radios'
+import Tabs from './components/tabs/tabs'
 
 function initAll () {
   new Button(document).init()
@@ -31,6 +32,11 @@ function initAll () {
   nodeListForEach($radios, function ($radio) {
     new Radios($radio).init()
   })
+
+  var $tabs = document.querySelectorAll('[data-module="tabs"]')
+  nodeListForEach($tabs, function ($tabs) {
+    new Tabs($tabs).init()
+  })
 }
 
 export {
@@ -40,5 +46,6 @@ export {
   Checkboxes,
   ErrorSummary,
   Header,
-  Radios
+  Radios,
+  Tabs
 }
--- a/src/components/tabs/README.md
+++ b/src/components/tabs/README.md
@@ -1,0 +1,375 @@
+# Tabs
+
+## Introduction
+
+Component for conditionally revealing content, using tabs and tabs panels.
+
+## Guidance
+
+Find out when to use the Tabs component in your service in the [GOV.UK Design System](https://govuk-design-system-production.cloudapps.digital/components/tabs).
+
+## Quick start examples
+
+### Component default
+
+[Preview the tabs component](http://govuk-frontend-review.herokuapp.com/components/tabs/preview)
+
+#### Markup
+
+    <div class="govuk-tabs" data-module="tabs">
+      <h2 class="govuk-tabs__title">
+        Contents
+      </h2>
+
+      <ul class="govuk-tabs__list">
+
+          <li class="govuk-tabs__list-item">
+            <a class="govuk-tabs__tab" href="#past-day">
+              Past day
+            </a>
+          </li>
+
+          <li class="govuk-tabs__list-item">
+            <a class="govuk-tabs__tab" href="#past-week">
+              Past week
+            </a>
+          </li>
+
+          <li class="govuk-tabs__list-item">
+            <a class="govuk-tabs__tab" href="#past-month">
+              Past month
+            </a>
+          </li>
+
+          <li class="govuk-tabs__list-item">
+            <a class="govuk-tabs__tab" href="#past-year">
+              Past year
+            </a>
+          </li>
+
+      </ul>
+
+      <section class="govuk-tabs__panel" id="past-day">
+        <h2 class="govuk-heading-l">Past day</h2>
+    <table class="govuk-table">
+      <thead class="govuk-table__head">
+        <tr class="govuk-table__row">
+          <th class="govuk-table__header" scope="col">Case manager</th>
+          <th class="govuk-table__header" scope="col">Cases opened</th>
+          <th class="govuk-table__header" scope="col">Cases closed</th>
+        </tr>
+      </thead>
+      <tbody class="govuk-table__body">
+        <tr class="govuk-table__row">
+          <td class="govuk-table__cell">David Francis</td>
+          <td class="govuk-table__cell">3</td>
+          <td class="govuk-table__cell">0</td>
+        </tr>
+        <tr class="govuk-table__row">
+          <td class="govuk-table__cell">Paul Farmer</td>
+          <td class="govuk-table__cell">1</td>
+          <td class="govuk-table__cell">0</td>
+        </tr>
+        <tr class="govuk-table__row">
+          <td class="govuk-table__cell">Rita Patel</td>
+          <td class="govuk-table__cell">2</td>
+          <td class="govuk-table__cell">0</td>
+        </tr>
+      </tbody>
+    </table>
+
+      </section>
+
+      <section class="govuk-tabs__panel" id="past-week">
+        <h2 class="govuk-heading-l">Past week</h2>
+    <table class="govuk-table">
+      <thead class="govuk-table__head">
+        <tr class="govuk-table__row">
+          <th class="govuk-table__header" scope="col">Case manager</th>
+          <th class="govuk-table__header" scope="col">Cases opened</th>
+          <th class="govuk-table__header" scope="col">Cases closed</th>
+        </tr>
+      </thead>
+      <tbody class="govuk-table__body">
+        <tr class="govuk-table__row">
+          <td class="govuk-table__cell">David Francis</td>
+          <td class="govuk-table__cell">24</td>
+          <td class="govuk-table__cell">18</td>
+        </tr>
+        <tr class="govuk-table__row">
+          <td class="govuk-table__cell">Paul Farmer</td>
+          <td class="govuk-table__cell">16</td>
+          <td class="govuk-table__cell">20</td>
+        </tr>
+        <tr class="govuk-table__row">
+          <td class="govuk-table__cell">Rita Patel</td>
+          <td class="govuk-table__cell">24</td>
+          <td class="govuk-table__cell">27</td>
+        </tr>
+      </tbody>
+    </table>
+
+      </section>
+
+      <section class="govuk-tabs__panel" id="past-month">
+        <h2 class="govuk-heading-l">Past month</h2>
+    <table class="govuk-table">
+      <thead class="govuk-table__head">
+        <tr class="govuk-table__row">
+          <th class="govuk-table__header" scope="col">Case manager</th>
+          <th class="govuk-table__header" scope="col">Cases opened</th>
+          <th class="govuk-table__header" scope="col">Cases closed</th>
+        </tr>
+      </thead>
+      <tbody class="govuk-table__body">
+        <tr class="govuk-table__row">
+          <td class="govuk-table__cell">David Francis</td>
+          <td class="govuk-table__cell">98</td>
+          <td class="govuk-table__cell">95</td>
+        </tr>
+        <tr class="govuk-table__row">
+          <td class="govuk-table__cell">Paul Farmer</td>
+          <td class="govuk-table__cell">122</td>
+          <td class="govuk-table__cell">131</td>
+        </tr>
+        <tr class="govuk-table__row">
+          <td class="govuk-table__cell">Rita Patel</td>
+          <td class="govuk-table__cell">126</td>
+          <td class="govuk-table__cell">142</td>
+        </tr>
+      </tbody>
+    </table>
+
+      </section>
+
+      <section class="govuk-tabs__panel" id="past-year">
+        <h2 class="govuk-heading-l">Past year</h2>
+    <table class="govuk-table">
+      <thead class="govuk-table__head">
+        <tr class="govuk-table__row">
+          <th class="govuk-table__header" scope="col">Case manager</th>
+          <th class="govuk-table__header" scope="col">Cases opened</th>
+          <th class="govuk-table__header" scope="col">Cases closed</th>
+        </tr>
+      </thead>
+      <tbody class="govuk-table__body">
+        <tr class="govuk-table__row">
+          <td class="govuk-table__cell">David Francis</td>
+          <td class="govuk-table__cell">1380</td>
+          <td class="govuk-table__cell">1472</td>
+        </tr>
+        <tr class="govuk-table__row">
+          <td class="govuk-table__cell">Paul Farmer</td>
+          <td class="govuk-table__cell">1129</td>
+          <td class="govuk-table__cell">1083</td>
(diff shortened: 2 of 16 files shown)


src/components/tabs/tabs.js
```diff
@@ -0,0 +1,275 @@
...
+
+  this.keys = { left: 37, right: 39, up: 38, down: 40 }
+  this.jsHiddenClass = 'js-hidden'
+}
+
+Tabs.prototype.init = function () {
+  if (typeof window.matchMedia === 'function') {
+    this.setupResponsiveChecks()
+  } else {
+    this.setup()
```
adamsilver: Previously, this feature detect was around the whole component which meant that on small screens no tabs will be setup.

Am I right in thinking that in browsers that don't support matchMedia, they'll get tabs regardless of screen size?

nickcolley: Given that the mobile browsers that we support all have [good support for matchMedia](https://caniuse.com/#search=matchmedia) maybe this isn't too much of a worry.

With the assumption that most of the time IE9 and below are used by users on desktop machines with wide screens?

adamsilver: Yes, it's probably okay. Just to note two examples where it's not as robust:

An IE9 user who opens up two windows side by side will get tabs in a small-ish screen.

Similarly, Blackberry 7, for example, will get tabs on their small screen.

In both cases, they will experience some layout issues such as horizontal scrolling or wrapping of tabs.

joelanman:
Just to note I think the example needs changing - this example is a pattern that does not have clear labels, and is no longer used by gov.uk

The labels issue was brought up in the working group review: https://github.com/alphagov/govuk-design-system-backlog/issues/100#issuecomment-395774112

timpaul:
Hey @joelanman - how about this example from the Judicial Case Manager?:

image

(the respondent tab has a similar table in it)

joelanman:
It's ok, I'm sure in context for these users that the labels are clear, but it would be nice if we could find one where the labels are just clear to most people.

timpaul:
@joelanman - here is a fictional but I think plausible alternative:

image

Each tab has a similar table, but the numbers are different.

Incidentally (@adamsilver you might have an opinion on this) - do we think that each tab should have a section heading? I don't think the above example needs them, but they might be necessary for accessibility / non-JS...

adamsilver:
I don't think the Tab component needs to dictate whether each panel should be marked up with a heading or not.

The Tab component itself does provide an accessible label (labelled by the panel's related tab label) so that's non-sighted users sorted.

If there's a scenario whereby we want headings without JS but no (visual) headings with JS, then I'd imagine the heading would be given a class of `js-hidden` which provides that functionality?

src/components/tabs/README.md
```diff
@@ -0,0 +1,383 @@
+# Tabs
+
+## Introduction
+
+Component for conditionally revealing content, using the details HTML element.
```
igloosi: not sure this is right

src/components/tabs/index.njk
```diff
@@ -0,0 +1,168 @@
...
+      {
+        text: 'Yes'
+      },
+      {
+        text: 'Array of tab items'
+      }
+    ],
+    [
+      {
+        text: 'item.id'
```
igloosi: we changed the way we show item properties from `item.id` to `items.{}.id`

src/components/tabs/_tabs.scss
```diff
@@ -0,0 +1,120 @@
...
+    padding-bottom: govuk-spacing(2);
+
+    &[aria-current = "true"] {
+      color: govuk-colour("black");
+      text-decoration: none;
+    }
+  }
+
+  // JavaScript enabled
+  .js-enabled .govuk-tabs {
```
36degrees: To try and keep specificity to a minimum, should this just be `.js-enabled`?

src/components/tabs/tabs.js
```diff
@@ -0,0 +1,282 @@
...
+  }
+}
+
+Tabs.prototype.setupResponsiveChecks = function () {
+  this.mql = window.matchMedia('(min-width: 40.0625em)')
+  this.mql.addListener(this.checkMode.bind(this))
+  this.checkMode(this.mql)
+}
+
+Tabs.prototype.checkMode = function (mql) {
```
36degrees: `mql` is passed in but never used - within the function we use `this.mql`.

src/components/tabs/tabs.js
```diff
@@ -0,0 +1,282 @@
...
+    this.setupResponsiveChecks()
+  } else {
+    this.setup()
+  }
+}
+
+Tabs.prototype.setupResponsiveChecks = function () {
+  this.mql = window.matchMedia('(min-width: 40.0625em)')
+  this.mql.addListener(this.checkMode.bind(this))
+  this.checkMode(this.mql)
```
36degrees: Passing `this.mql` unnecessarily as `checkMode` never uses it.

src/components/tabs/tabs.js
```diff
@@ -0,0 +1,282 @@
...
+
+    // Handle events
+    $tab.boundTabClick = this.onTabClick.bind(this)
+    $tab.boundTabKeydown = this.onTabKeydown.bind(this)
+    $tab.addEventListener('click', $tab.boundTabClick, true)
+    $tab.addEventListener('keydown', $tab.boundTabKeydown, true)
+
+    // Remove old active panels
+    this.unhighlightTab($tab)
+    this.hidePanel($tab)
```
36degrees: Can we change these two calls for `hideTab` which wraps them both up together?

src/components/tabs/tabs.js
```diff
@@ -0,0 +1,282 @@
...
+  }.bind(this))
+
+  // If there's a tab that matches the hash
+  var $activeTab = this.getTab(window.location.hash)
+  if ($activeTab) {
+    this.highlightTab($activeTab)
+    this.showPanel($activeTab)
```
36degrees: Likewise I believe both `highlightTab` and `showPanel` can be replaced by `showTab`?

src/components/tabs/tabs.js
```diff
@@ -0,0 +1,282 @@
...
+  } else {
+    // or show the first
+    var firstTab = this.$tabs[0]
+    this.highlightTab(firstTab)
+    this.showPanel(firstTab)
```
36degrees: And these two can be showTab as well

src/components/tabs/tabs.js
```diff
@@ -0,0 +1,282 @@
...
+    $tab.addEventListener('keydown', $tab.boundTabKeydown, true)
+
+    // Remove old active panels
+    this.unhighlightTab($tab)
+    this.hidePanel($tab)
+  }.bind(this))
+
+  // If there's a tab that matches the hash
+  var $activeTab = this.getTab(window.location.hash)
+  if ($activeTab) {
```
36degrees: To avoid duplication I think this could be:

```js
var $activeTab = this.getTab(window.location.hash)
if (!$activeTab) {
  $activeTab = this.$tabs[0]
}
this.showTab($activeTab)
```

or even

```js
// Show either the active tab according to the URL's hash or the first tab
var $activeTab = this.getTab(window.location.hash) || this.$tabs[0]
this.showTab($activeTab)
```

src/components/tabs/tabs.js
```diff
@@ -0,0 +1,282 @@
...
+  nodeListForEach($tabListItems, function ($item) {
+    $item.setAttribute('role', 'presentation')
+  })
+
+  nodeListForEach($tabs, function ($tab) {
+    // Set HTML attributes
+    this.setAttributes($tab)
+
+    // Handle events
+    $tab.boundTabClick = this.onTabClick.bind(this)
```
36degrees: Worth a comment to explain why we need to store a reference to these.

src/components/tabs/tabs.js
```diff
@@ -0,0 +1,282 @@
...
+  // Remove hashchange event handler
+  window.removeEventListener('hashchange', $module.boundOnHashChange, true)
+}
+
+Tabs.prototype.onHashChange = function (e) {
+  var hash = window.location.hash
+  if (!this.hasTab(hash)) {
+    return
+  }
+  if (this.changingHash) {
```
36degrees: Worth an explaining comment I think

src/components/tabs/tabs.js
```diff
@@ -0,0 +1,282 @@
...
+    this.changingHash = false
+    return
+  }
+  var $hashTab = this.getTab(window.location.hash)
```
36degrees: Can use `hash` rather than `window.location.hash`.

Would also consider skipping `hasTab` and storing and using the result from `getTab` which would mirror what we do in setup (`var $activeTab = this.getTab(window.location.hash)`)

src/components/tabs/tabs.js
```diff
@@ -0,0 +1,282 @@
...
+  var $currentTab = this.getCurrentTab()
+  if ($hashTab) {
```
36degrees: This must be true if we got here? Because we can only reach this code if we didn't return as a result of `!this.hasTab`.

src/components/tabs/tabs.js
```diff
@@ -0,0 +1,282 @@
...
+    this.hideTab($currentTab)
+    this.showTab($hashTab)
+    $hashTab.focus()
+  } else {
+    var $firstTab = this.$tabs.firstElementChild
```
36degrees: I _think_ this is dead code that can be removed, but if not it'd be good to reduce the duplication.

src/components/tabs/tabs.js
```diff
@@ -0,0 +1,282 @@
...
+    this.hideTab($currentTab)
+    this.showTab($firstTab)
+    $firstTab.focus()
+  }
+}
+
+Tabs.prototype.hasTab = function (hash) {
+  return this.$module.querySelector(hash)
```
36degrees: Is there a reason this uses a different selector to `getTab`?

src/components/tabs/tabs.js
```diff
@@ -0,0 +1,282 @@
+import '../../vendor/polyfills/Function/prototype/bind'
+import '../../vendor/polyfills/Element/prototype/classList'
+import '../../vendor/polyfills/Event' // addEventListener and event.target normaliziation
+import { nodeListForEach } from '../../common'
+
+function Tabs ($module) {
+  this.$module = $module
+  this.$tabs = $module.querySelectorAll('.govuk-tabs__tab')
```
36degrees: Have we tested having multiple tab components on a single page?

alex-ju: Yes, did tested with multiple per page.

src/components/tabs/tabs.js
alex-ju: Not dead code, as when there's not hash it falls back to the first tab. Will try and reduce the logic.

src/components/tabs/tabs.js
alex-ju: Comment added.

src/components/tabs/tabs.js
alex-ju: Comment added.

alex-ju:
Updated addressing all comments.

src/components/tabs/tabs.js
```diff
@@ -0,0 +1,272 @@
...
+  var $currentTab = this.getCurrentTab()
+  this.hideTab($currentTab)
+  this.showTab($newTab)
+  this.createHistoryEntry($newTab)
+}
+
+Tabs.prototype.createHistoryEntry = function ($tab) {
+  var $panel = this.getPanel($tab)
+  var id = $panel.id
+  $panel.id = ''
```
36degrees: Why do we have to null and restore the panel's `id`? If this is behaviour we definitely need, can we add a comment to explain why?

adamsilver:
It is so the page doesn't jump when a user clicks a tab (which changes the
hash).

src/components/tabs/tabs.js
adamsilver: It is so the page doesn't jump when a user clicks a tab (which changes the
hash).

src/all.js
```diff
@@ -31,6 +32,11 @@ function initAll () {
   nodeListForEach($radios, function ($radio) {
     new Radios($radio).init()
   })
+
+  var $tabs = document.querySelectorAll('[data-module="tabs"]')
```
adamsilver: At the moment the macro template has the data attribute on the container. This means that if I use the macro, it will automatically get instantiated into tabs.

This could become problematic if I want to do something more custom. Like listen to an event fired by the instance (I know we can't do that right now).

Is GOV.UK Frontend continuing to instantiate automatically like this?

adamsilver:
Will `Tabs` constructor be a global? Or will it be assigned as `GOVUK.Tabs`?

src/all.js
alex-ju: You concern is valid, but this is consistent with how the other components are being initialised. Will make a note and look into that for all the components.

adamsilver: Yep, it might be okay, broadly speaking, for say tabs and conditional radio button reveals. But as a general approach it can be overly constraining.

Thanks @alex-ju

alex-ju:
@adamsilver: the Tabs constructor is attached to a GOVUKFrontend variable and every tab instance will be initialised separately using the `new` operator.

For example if we consider `$tabs` a queried tab element, the JavaScript initialisation will be `new GOVUKFrontend.Tabs($tabs).init()`

src/all.js
nickcolley: This use case is already accounted for: you can avoid `initAll()` and call the constructor yourself.

nickcolley: Since we did work to stop `initAll()` from executing automatically :)

alex-ju: Yes, not initialising or initialising separately a specific component is also an option.
When Adam mentioned 'something custom', I was thinking of a case when you want to initialise but hook different listeners or add custom functionality to a Tabs instance before doing that, which I think is still possible with the current model:
```
var myTabs = new Tabs($tabs)
myTabs.myCustomFunction = function () { ... }
myTabs.init()
```

PR #777, opened 2018-06-08
Refactor colours

36degrees:
- Define the colour palette in a map, rather than as multiple variables, allowing them to be overridden by consumers
- Add a new function `govuk-colour` to allow you to easily get an organisation colour
- Update other layers to use the `govuk-colour` function
- Remove tints
- Rename colours to be less opinionated (as raised in https://github.com/alphagov/govuk_elements/issues/561):
  - `mauve` → `light-purple`
  - `fuchsia` → `bright-purple`
  - `baby-pink` → `light-pink`
  - `mellow-red` → `bright-red`
  - `grass-green` → `light-green`

https://trello.com/c/B8OCSU8R/759-provide-a-defined-public-api-for-govuk-frontend


nickcolley:
Are there any restrictions where mixins can be used, or can they be composed the same way you would with variables?

igloosi:
getting rid of -25s and 50s as well?

36degrees:
> Are there any restrictions where mixins can be used, or can they be composed the same way you would with variables?

Do you mean functions? I've not come up against any issues so far… as far as I know you can use them anywhere.

36degrees:
> getting rid of -25s and 50s as well?

Yep. Discussed with @timpaul and we don't think they're useful at the minute - you can get them programatically and if we want people to stick to certain tints we should probably do that either via the mixin or through documentation.

PR #778, opened 2018-06-08
Remove prose scope

igloosi:
We don't have confidence that the prose scope is well understood – without further research / better documentation it's safer to remove it for now.
  
If you are currently using prose-scope, you should revert to applying classes to individual headings, lists and hr elements.

In version [0.0.29-alpha](https://github.com/alphagov/govuk-frontend/releases/tag/v0.0.29-alpha) release we have added the option to enable global link and paragraph styles. If you're using a recent version of the Private Beta Prototype kit, this is enabled by default.

Trello ticket: https://trello.com/c/v7tR99UO/955-remove-prose-scope


hannalaakso:
Looks like there's still a few references to the prose scope in `app/__test__/app.test.js` and in files in `src/core`.

igloosi:
good spot. updated

davidjamesstone:
FWIW, I thought this was a really neat feature. There are similar, generic classes in other css frameworks. I thought these were fairly well understood. e.g. [bulma](https://bulma.io/documentation/elements/content/).

A single class to handle WYSIWYG generated content, where only HTML tags are available is a common use case .

+1 for it's return!

36degrees:
Thanks for the feedback @davidjamesstone – that's really useful to know. We've not ruled out adding prose scope or at least something very similar back in in the future. It's also good to see examples of how this works in other systems, like in the Bulma documentation you linked to.

PR #779, opened 2018-06-12
Refactor spacing API

hannalaakso:
This PR:

- Defines the spacing variables in a map, rather than as multiple variables, allowing control over error messaging when an incorrect variable is used and allows them to be overridden by consumers.

In detail, it:
- Adds a new helper `govuk-spacing` to fetching single point spacing.
- Moves single point spacing variables inside a map.
- Moves responsive spacing maps inside a collective map `$govuk-spacing-responsive-scale` and updates the helper.
- Removes the old `$govuk-spacing-responsive-scale` map we created for generating overrides as the overrides can now use the new scale map 🎉 
- Possibly controversial: makes `govuk-responsive-spacing()` a **private** mixin as it should only be used via `govuk-responsive-margin` and `govuk-responsive-padding`.
- Updates other layers to use the `govuk-spacing` function and the new responsive spacing map.
- Updates tests accordingly and adds new tests for `govuk-spacing()`
- Updates comments for sass-docs.

Other:
- `$govuk-gutter` is now hardcoded as `govuk-spacing` cannot be accessed from inside the settings layer. However if we think it should be coupled more tightly with the spacing scale, we could maybe use `map-get` to fetch a specific spacing point in settings... I’ve created an [issue](https://github.com/alphagov/govuk-frontend/issues/780) to review this.

Note: the PR is easiest to review commit-by-commit.

https://trello.com/c/0IDTgRtF/1112-provide-a-defined-public-api-for-govuk-frontend-spacing

### DIFF
--- a/src/helpers/_spacing.scss
+++ b/src/helpers/_spacing.scss
@@ -2,34 +2,76 @@
 /// @group helpers
 ////
 
+/// Single point spacing
+///
+/// Returns measurement corresponding to the spacing point requested.
+///
+/// @param {Number} $spacing-point - Point on the spacing scale (set in `settings/_spacing.sccs`)
+///
+/// @returns {String} Spacing Measurement eg. 10px
+///
+/// @example scss
+///   .element {
+///     padding: govuk-spacing(5);
+///     top: govuk-spacing(2) !important; // if `!important` is required
+///   }
+/// @access public
+
+@function govuk-spacing($spacing-point) {
+
+  $actual-input-type: type-of($spacing-point);
+  @if $actual-input-type != "number" {
+    @error "Expected a number (integer), but got a "
+    + "#{$actual-input-type}.";
+  }
+
+  @if not map-has-key($govuk-spacing-points, $spacing-point) {
+    @error "Unknown spacing variable `#{$spacing-point}`. Make sure you are using a point from the spacing scale in `_settings/spacing.scss`.";
+  }
+
+  @return map-get($govuk-spacing-points, $spacing-point);
+}
+
 /// Responsive spacing
 ///
 /// Adds responsive spacing (either padding or margin, depending on `$property`)
-/// using a 'spacing map', which defines different spacing values at different
-/// breakpoints.
+/// by fetching a 'spacing map' from the responsive spacing scale, which defines
+/// different spacing values at different breakpoints.
 ///
-/// @param {Map} $scale-map - Map of breakpoints and spacing values
+/// To generate responsive spacing, use 'govuk-responsive-margin' or
+/// 'govuk-responsive-padding' mixins
+///
+/// @param {Number} $responsive-spacing-point - Point on the responsive spacing
+///   scale, corresponds to a map of breakpoints and spacing values
 /// @param {String} $property - Property to add spacing to (e.g. 'margin')
 /// @param {String} $direction [all] - Direction to add spacing to
 ///   (`top`, `right`, `bottom`, `left`, `all`)
 /// @param {Boolean} $important [false] - Whether to mark as `!important`
 /// @param {Number} $adjustment [false] - Offset to adjust spacing by
 ///
-/// @access public
+/// @access private
 
-@mixin govuk-responsive-spacing($scale-map, $property, $direction: "all", $important: false, $adjustment: false) {
+@mixin _govuk-responsive-spacing($responsive-spacing-point, $property, $direction: "all", $important: false, $adjustment: false) {
 
-  // Ensure that $scale-map is actually a map - it's really easy to accidentally
-  // pass in a point from the non-responsive scale, in which case this mixin
-  // would otherwise no-op silently.
-  $actual-input-type: type-of($scale-map);
-  @if $actual-input-type != "map" {
-    @error "Expected a map of breakpoints from the responsive scale, but got a "
-    + "#{$actual-input-type}. Make sure you are using a point from the "
-    + "responsive spacing scale.";
+  $actual-input-type: type-of($responsive-spacing-point);
+  @if $actual-input-type != "number" {
+    @error "Expected a number (integer), but got a " + "#{$actual-input-type}.";
   }
 
-  // Loop through each breakpoint
+  @if not map-has-key($govuk-spacing-responsive-scale, $responsive-spacing-point) {
+    @error "Unknown spacing point `#{$responsive-spacing-point}`. Make sure you are using a point from the "
+    + "responsive spacing scale in `_settings/spacing.scss`.";
+  }
+
+  // Make sure that the return value from `_settings/spacing.scss` is a map.
+  $scale-map: map-get($govuk-spacing-responsive-scale, $responsive-spacing-point);
+  $actual-map-type: type-of($scale-map);
+  @if $actual-map-type != "map" {
+    @error "Expected a number (integer), but got a "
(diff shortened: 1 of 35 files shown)


src/settings/_spacing.scss
```diff
@@ -1,88 +1,80 @@
...
+) !default;
 
-// To use the responsive spacing maps, combine them with
-// 'govuk-responsive-margin' or 'govuk-responsive-padding' mixins
-// (see 'helpers/spacing').
+/// Responsive spacing maps
+///
+/// These definitions are used to generate responsive spacing that adapts
+/// according to the breakpoints (see 'helpers/spacing'). These maps should be
+/// used wherever possible to standardise responsive spacing.
```
36degrees: Picky, but can we add the `///` to the blank lines to so that it reads as one connected comment block?

src/helpers/_spacing.scss
```diff
@@ -2,34 +2,77 @@
...
-/// breakpoints.
+/// by fetching a 'spacing map' from the responsive spacing scale, which defines
+/// different spacing values at different breakpoints.
+///
+/// To generate responsive spacing, use 'govuk-responsive-margin' or
+/// 'govuk-responsive-padding' mixins
 ///
-/// @param {Map} $scale-map - Map of breakpoints and spacing values
+/// @param {Number} $responsive-spacing-point - Point on the responsive spacing
+/// scale, corresponds to a map of breakpoints and spacing values
```
36degrees: Would be good to indent the second line like we have on :48

src/helpers/_spacing.scss
```diff
@@ -2,34 +2,77 @@
...
-  // pass in a point from the non-responsive scale, in which case this mixin
-  // would otherwise no-op silently.
-  $actual-input-type: type-of($scale-map);
-  @if $actual-input-type != "map" {
-    @error "Expected a map of breakpoints from the responsive scale, but got a "
-    + "#{$actual-input-type}. Make sure you are using a point from the "
-    + "responsive spacing scale.";
+  $actual-input-type: type-of($responsive-spacing-point);
+  @if $actual-input-type != "number" {
+    @error "Expected a number (integer), but got a "
```
36degrees: I _think_ this is short enough it'll fit on one line?

src/helpers/_spacing.scss
```diff
@@ -58,40 +101,53 @@
...
 ///
-/// @param {Map} $scale-map - Map of breakpoints and spacing values
+/// @param {Number} $responsive-spacing-point - Point on the responsive spacing
+/// scale, corresponds to a map of breakpoints and spacing values
 /// @param {String} $direction [all] - Direction to add spacing to
 ///   (`top`, `right`, `bottom`, `left`, `all`)
 /// @param {Boolean} $important [false] - Whether to mark as `!important`
 /// @param {Number} $adjustment [false] - Offset to adjust spacing by
 ///
+/// @example scss
```
36degrees: 👌

src/helpers/spacing.test.js
```diff
@@ -67,7 +110,7 @@ describe('@mixin govuk-responsive-spacing', () => {
             padding-top: 25px; } }`)
   })
 
-  it('throws an exception when passed anything other than a map', async () => {
+  it('throws an exception when passed anything other than a number', async () => {
```
36degrees: Technically '14px' is a number in Sass, so maybe 'throws an exception when passed a non-existent point' or similar?

src/settings/_spacing.scss
```diff
@@ -1,88 +1,80 @@
...
+/// Single point spacing variables. Access using `govuk-spacing()`
+/// (see `helpers/spacing`).
+///
+/// @type Map
+/// @access private
 
-// You can define different behaviour on tablet and desktop. The 'null'
-// breakpoint is for mobile.
+$govuk-spacing-points: (
+  0: 0,
```
igloosi: in there any point in having 0 in teh scale, since we've used 0 where needed instead of `govuk-spacing(0)?
```
.govuk-header--full-width {
   padding: 0 govuk-spacing(3);
}
```

36degrees: Yes, because it means there's an override generated for it, so you can remove margins.

src/helpers/_spacing.scss
```diff
@@ -2,34 +2,76 @@
...
+/// @param {Number} $responsive-spacing-point - Point on the responsive spacing
+///   scale, corresponds to a map of breakpoints and spacing values
 /// @param {String} $property - Property to add spacing to (e.g. 'margin')
 /// @param {String} $direction [all] - Direction to add spacing to
 ///   (`top`, `right`, `bottom`, `left`, `all`)
 /// @param {Boolean} $important [false] - Whether to mark as `!important`
 /// @param {Number} $adjustment [false] - Offset to adjust spacing by
 ///
-/// @access public
+/// @access private
```
36degrees: If this is private then the mixin name should be prefixed with an underscore (`_govuk-responsive-spacing`). Change makes sense I think as you can achieve everything with `govuk-responsive-margin` and `govuk-responsive-padding`?

src/helpers/_spacing.scss
```diff
@@ -58,40 +100,53 @@
...
+///   .element {
+///      @include govuk-responsive-margin(6, "left", $adjustment: 1px);
+///   }
+///
 /// @access public
 
-@mixin govuk-responsive-margin($scale-map, $direction: "all", $important: false, $adjustment: false) {
-  @include govuk-responsive-spacing($scale-map, "margin", $direction, $important, $adjustment);
+@mixin govuk-responsive-margin($responsive-spacing-point, $direction: "all", $important: false, $adjustment: false) {
+  @include govuk-responsive-spacing($responsive-spacing-point, "margin", $direction, $important, $adjustment);
```
36degrees: We could pass all arguments [like we do for mq](https://github.com/alphagov/govuk-frontend/blob/master/src/helpers/_media-queries.scss#L91-L95) if we wanted to avoid having to maintain the same signature in three places (though we'd need to update the docs anyway I guess…)

hannalaakso: Thank @36degrees. As discussed IRL, we need to pass the property (`margin` / `padding`) through to `_ govuk-responsive-spacing` so `$args...` wouldn't work in this scenario.

src/overrides/_spacing.scss
```diff
@@ -23,28 +23,30 @@ $spacing-directions: (
 // .govuk-\!-mt-r1 {
 //   margin-top: [whatever spacing point 1 is...]
 // }
+///
```
36degrees: Can the rest of this code block be updated to Sassdoc? (triple slash, use `@example`, document the parameters)?

hannalaakso: For sure, good shout @36degrees. Done now.

src/overrides/_spacing.scss
```diff
@@ -1,50 +1,62 @@
...
+/// for each point in the spacing scale.
+///
+/// These override classes are responsive - the 'r' before the scale point
+/// indicates that. We might want to add non-responsive ones later.
+///
+/// @param {String} $property - Property to add spacing to (e.g. 'margin')
+/// @param {String} $property-shorthand - Shorthand representation of property, eg. "m" for "margin"
+///
+/// @example scss
+///   .govuk-\!-m-r0 {
```
igloosi: as a note: this will need to be updated in #786

PR #782, opened 2018-06-12
Remove the ‘circle shape’ object

36degrees:
This is no longer a particularly useful abstraction now that list with step icons has been removed, as it’s only used in one place and includes a lot of very specific styles.


36degrees:
@igloosi @nickcolley happy with changelog?

igloosi:
all good

PR #784, opened 2018-06-12
Use the correct class name in bold label example

### LINKED ISSUE #783: Missing `govuk-label--bold`
we talk about this in the component variants http://govuk-frontend-review.herokuapp.com/components/label/

but we seem to have removed it

Shall we remove it totally or add the style?

igloosi:
Currently we talk about a bold label modifier class
https://govuk-frontend-review.herokuapp.com/components/label
which doesn't exist.

It's been replaced by `.govuk-label--s` class.

Updated to add the correct modifier class to the yaml file and README
https://govuk-frontend-review-pr-784.herokuapp.com/components/label

Fixes #783

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -165,6 +165,10 @@ Note: We're not following semantic versioning yet, we are going to talk about th
   and footer (the OGL logo) by marking them as non-focusable elements
   ([PR #774](https://github.com/alphagov/govuk-frontend/pull/774))
 
+- Use the correct class name in bold label example 
+  (govuk-label--s rather than govuk-label--bold)
+  ([PR #784](https://github.com/alphagov/govuk-frontend/pull/784))
+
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -165,6 +165,9 @@ Note: We're not following semantic versioning yet, we are going to talk about th
   and footer (the OGL logo) by marking them as non-focusable elements
   ([PR #774](https://github.com/alphagov/govuk-frontend/pull/774))
 
+- Update CSS modifier class for a bold label
```
36degrees: I don't think I'd know what this meant if I read this without looking at the PR.

What about something like "Use the correct class name in bold label example (`govuk-label--s` rather than `govuk-label--bold`)"

igloosi: yeah, was a bit optimistic with the title. updated

PR #785, opened 2018-06-13
Add notice about the use of `html` arguments in Nunjucks macros for production

igloosi:
We need to make developers aware that if they decide to use Nunjucks macros in production and use `html` arguments or ones ending in `Html` that render unescaped html, making it a [security risk](https://en.wikipedia.org/wiki/Cross-site_scripting)

This adds notice text before and after the table of arguments in the README file for each component.
Text is set as a variable, so it can be updated in a single place if required.

Part of this Trello ticket: 
https://trello.com/c/SruPShLz/1049-add-clearer-documentation-around-text-vs-html-in-nunjucks-macros

Adresses: #514

### DIFF
--- a/app/views/layouts/readme.njk
+++ b/app/views/layouts/readme.njk
@@ -1,3 +1,4 @@
+{% set nunjucksHtmlUsageMessage = '**If you’re using Nunjucks macros in production be aware that using  `html` arguments, or ones ending with `Html` can be a [security risk](https://en.wikipedia.org/wiki/Cross-site_scripting). More about it in the [Nunjucks documentation](https://mozilla.github.io/nunjucks/api.html#user-defined-templates-warning).**' %}
 {% set componentName = componentPath %}
 {% set componentNameHuman = componentName | replace("-", " ") | capitalize %}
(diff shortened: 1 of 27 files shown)


app/views/layouts/readme.njk
```diff
@@ -1,3 +1,4 @@
+{% set nunjucksHtmlUsageMessage = '**If you’re using Nunjucks macros in production be aware that using  `html` arguments, or ones ending with `Html` can be a [security risk](https://en.wikipedia.org/wiki/Cross-site_scripting).**' %}
```
36degrees: Are there any components where this notice isn't being added? If not, is there any reason we couldn't just add this to this layout above the `componentArguments` block and save including this in every component?

igloosi: no, it's everywhere. updated it. thanks

nickcolley: Could we instead link to the Nunjucks documentation: https://mozilla.github.io/nunjucks/api.html#user-defined-templates-warning ?

PR #789, opened 2018-06-14
Save changed package-lock

hannalaakso:
There seems to be multiple changes to `package-lock` at the moment with multiple changes to package version numbers (as well as some changes of integrity hashes which is as expected). This is despite the fact that last time the `package.json` was changed, we also committed the `package-lock` in the PR.

As we are doing a release of GOV.UK Frontend, I'm going to commit this into the repo.

We should keep an eye on this though going forward.


hannalaakso:
Don't want to do this. It seems that the source of the problem is that members of the team were on different versions of `npm`.

PR #796, opened 2018-06-15
Use the main README when deploying to npm

nickcolley:
When running the `npm run build:package` command now the main project README will be copied into the `package/` directory, bringing the two in sync for when it is published to npm.


igloosi:
not sure we want that.
for npm published readme we kinda want a mash of intro from here and then bits from here https://github.com/alphagov/govuk-frontend/blob/master/docs/installation/installing-with-npm.md

link to that page is relative so it wouldn't work on npm

nickcolley:
I'm worried about having duplicated content, since they go out of sync.

I'll raise an issue instead and close this PR

PR #806, opened 2018-06-20
Reduce margin-bottom on the hint when following a default or small label

dashouse:
The `margin-bottom` of the hint component is the same for all label sizes.

For the default (no class) or `govuk-label--s` class this looks unbalanced.

screen shot 2018-06-20 at 09 46 35

This PR reduces the margin-bottom of the hint by 5px after a default `<label>` or `<label class="govuk-label--s">`.

**Before**
screen shot 2018-06-20 at 09 45 05

**After**
screen shot 2018-06-20 at 09 45 34

This does not affect the hint when used with a legend (original spacing is fine) or when the `<label>` is inside an `<h1>` (isPageHeading: True) - however I'm ok with this because the default / `govuk-label--s` will never be used in this scenario.

### DIFF
--- a/src/components/hint/_hint.scss
+++ b/src/components/hint/_hint.scss
@@ -12,4 +12,18 @@
 
     color: $govuk-secondary-text-colour;
   }
+
+  // Reduces margin-bottom of hint when used after the default label (no class)
+  // or govuk-label--s for better vertical alignment.
+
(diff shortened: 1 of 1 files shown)


36degrees:
I'm happy with this but am not going to review it as I paired with Dave to come up with the solution.

PR #807, opened 2018-06-20
Add default container to header

alex-ju:
This PR adds the a default container to the header component to  simplify the initialisation and make it consistent with the footer.

### Before
```
{{ govukHeader({
  "homepageUrl": "/",
  "containerClasses": "govuk-width-container"
}) }}
```

### After  
```
{{ govukHeader() }}
```

### Notes
Re-tested with all browsers in our matrix.

### DIFF
--- a/src/components/header/_header.scss
+++ b/src/components/header/_header.scss
@@ -24,9 +24,13 @@
 
   }
 
-  .govuk-header--full-width {
+  .govuk-header__container--full-width {
     padding: 0 govuk-spacing(3);
     border-color: $govuk-header-border-color;
+
+    .govuk-header__menu-button {
+      right: govuk-spacing(3);
+    }
   }
 
   .govuk-header__container {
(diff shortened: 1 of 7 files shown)


src/components/header/_header.scss
```diff
@@ -24,9 +24,13 @@
 
   }
 
-  .govuk-header--full-width {
+  .govuk-header__container--full-width {
```
36degrees: I find it a _bit_ weird that the _container_ class adds a border and has a side effect on a contained element – I'd expect the container just to set margins and width… but might just be me. Anyone else got any thoughts on the matter?

alex-ju: Yep, it bugs me too, but otherwise I'll have to check the classes applied to the header and the ones applied to the container to make sure we don't have 'strange' mix (e.g. `classes: govuk-header--full-width` and `containerClasses: govuk-width-container`)

nickcolley: I'd go with what we have now, and reconsider modifiers again in the future as it allows you to do what you need to inside the component.

36degrees: 👍

PR #808, opened 2018-06-20
Update JavaScript docs

hannalaakso:
This PR:

- Updates the instructions for using JavaScript by grouping them under `include` or `import`
- Document including the scripts, instead of importing them, as an option
- Fixes out-of-date initialisation instructions.
- Fixes some heading levels.
- Remove a stray comment about FE being available to
install as multiple packages
- Updates Webpack example
- Add some comments for extra guidance to `src/all.js`

[Trello card](https://trello.com/c/aYoiCjGx)

### DIFF
--- a/docs/installation/installing-with-npm.md
+++ b/docs/installation/installing-with-npm.md
@@ -22,12 +22,9 @@ To use GOV.UK Frontend with NPM you must:
 npm install nunjucks --save
 ```
 
-### Installation
+## Installation
 
-You can install all components or one or more individual components depending on
-your needs.
-
-To install all components, run:
+To install, run:
 
 ```
 npm install --save govuk-frontend
@@ -36,7 +33,7 @@ npm install --save govuk-frontend
 After you have installed GOV.UK Frontend the `govuk-frontend` package will
 appear in your `node_modules` folder.
 
-### Importing styles
+## Importing styles
 
 You need to import the GOV.UK Frontend styles into the main Sass file in your
 project. You should place the below code before your own Sass rules (or Sass
@@ -55,7 +52,7 @@ your Sass file:
   @import "node_modules/govuk-frontend/components/button/button";
   ```
 
-#### Optional: Resolving import paths
+### Optional: Resolving SCSS import paths
 
 If you wish to resolve the above `@import` paths in your build (in order to
 avoid prefixing paths with `node_modules`), you should add `node_modules` to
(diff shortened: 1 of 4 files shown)


docs/installation/installing-with-npm.md
```diff
@@ -106,35 +103,48 @@ $govuk-global-styles: true;
...
+
+#### Option 1: Include JavaScript
+
+Include the `node_modules/govuk-frontend/all.js` script on your page. You might wish to copy the file into your project or reference it from `node_modules`.
+
+See below for an example:
 
 ```js
-import All from 'govuk-frontend/all'
+   <script src="govuk-frontend/all.js"></script>
```
igloosi: they still need to initialise it

docs/installation/installing-with-npm.md
```diff
@@ -145,7 +155,37 @@ instead use:
...
+
+You can initialise all the components with the below snippet:
+
+```js
+  window.GOVUKFrontend.initAll()
+```
+
+You can also initialise individual components using the `data-module` attribute set on the component. If you only have one instance of a component per page, you can follow the below example to initialise it:
+
+```js
```
alex-ju: The initialisation is dependent on the include/import method.

- If we use `import Radios from 'govuk-frontend/components/radios/radios'` we assign the exported constructor to the `Radios` variable at import.
- If we include the `all.js` script we need to refer the constructor from the `GOVUKFrontend` global variable `var Radios = window.GOVUKFrontend.Radios`

And then to the initialisation as shown below.

docs/installation/installing-with-npm.md
```diff
@@ -55,7 +52,7 @@ your Sass file:
   @import "node_modules/govuk-frontend/components/button/button";
   ```
 
-#### Optional: Resolving import paths
+### Optional: Resolving SASS import paths
```
nickcolley: SASS -> SCSS

docs/installation/installing-with-npm.md
```diff
@@ -106,45 +103,124 @@ $govuk-global-styles: true;
...
+The button component is initialised differently as the script polyfills globally for all anchor tags with `role="button"`; for this reason, there is no need to initialise using the `data-module` attribute. You can initialise the button component with the below:
 
 ```js
-import Button from 'govuk-frontend/components/button/button'
+var Button = window.GOVUKFrontend.Button
+new Button(document).init()
+```
 
-new Button().init()
+#### Initialise all included components
```
nickcolley: I think this would make sense to be above the other, since I imagine most people will use this by default.

nickcolley: We could consider merging this example with the place we tell people to include a script tag.

nickcolley:
Thanks Hanna, I think this'll help people a lot

PR #821, opened 2018-06-22
Make radio and checkbox labels inline-block

### LINKED ISSUE #543: Radios labels are full width - should be the width of the label text?

igloosi:
This fixes #543 regression, reintroduced with conditional reveal.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,11 +23,12 @@
 🔧 Fixes:
 
 - Update Crown copyright link
-
   Update the Crown copyright link on the National Archives so
   we don't send users on an unnecessary redirect.
-
   ([PR 824](https://github.com/alphagov/govuk-frontend/pull/824))
+
+- Fixes radio and checkbox labels extending full width of page
+  ([PR #821](https://github.com/alphagov/govuk-frontend/pull/821))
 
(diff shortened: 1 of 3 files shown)


36degrees:
@nickcolley any idea why the labels were made `display: block`? Was there an issue in a particular browser we need to test?

igloosi:
IE's seem to be fine with inline-block
ie8
ie9
ie10
ie11

nickcolley:
Only thing I can think of is potential wrapping, with the conditional element, but maybe I just made a mistake.

CHANGELOG.md
```diff
@@ -22,11 +22,8 @@
 
 🔧 Fixes:
 
-- Pull Request Title goes here
-
-  Description goes here (optional)
-
-  ([PR #N](https://github.com/alphagov/govuk-frontend/pull/N))
+- Make radio and checkbox labels `inline-block` instead of `block`
```
nickcolley: Can we reword this to explain what it fixes?

"Fixes radio and checkbox labels extending full width of page"

PR #824, opened 2018-06-22
Update link to Crown copyright

bevanloon:
Update the link to the Crown copyright so we don't send users on an
unnecessary redirect.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,11 +22,12 @@
 
 🔧 Fixes:
 
-- Pull Request Title goes here
+- Update Crown copyright link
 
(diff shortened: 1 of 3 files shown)


igloosi:
hi, @bevanloo thanks for this.

tests are failing as the footer component's readme file hasn't been updated.
could run `gulp generate:readme` and commit the Readme file as well?

the other tiny thing is to add the pr number where it says PR# in the CHANGELOG

bevanloon:
@igloosi - thanks for the help.

I don't have merge permissions on this one so if someone who does would do the honours please that would be grand :)

PR #826, opened 2018-06-23
Add support for customising item inputs’ field names in date-input

lhokktyn:
Add support for a new `item.{}.fieldName` parameter on the `date-input`
macro. This allows the author to explicitly set the field name on each
item input rather than rely on the `<param.name>-<item.name>` default
convention, which is not always convenient and couples the field name to
the displayed label too tightly.

Also fix a few minor missing/incorrect parts in the date-input README.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,10 +14,7 @@
 
 🆕 New features:
 
-- Pull Request Title goes here
-
-  Description goes here (optional)
-
+- Add support for specific input field names to date-input component
   ([PR #N](https://github.com/alphagov/govuk-frontend/pull/N))
 
 🔧 Fixes:
@@ -53,7 +50,7 @@
 🔧 Fixes:
 
 - Reduce margin-bottom on the hint when following a default or small labe
-  This reduces the margin-bottom of the hint by 5px after a default 
+  This reduces the margin-bottom of the hint by 5px after a default
   `<label>` or `<label class="govuk-label--s">`.
   ([PR #806](https://github.com/alphagov/govuk-frontend/pull/806))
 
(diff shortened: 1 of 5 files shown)


igloosi:
Hi James, thanks for raising this.
It looks like an interesting amendment.

As it involves modifying what we deem as public api, we'll have a discussion how to proceed.
Will keep you updated.

lhokktyn:
Thanks for the update @igloosi.

The primary reason for its introduction was down to our particular need to have date inputs named as `date[dd]`, `date[mm]`, `date[yyyy]` so that they could get automatically parsed into an object format by our service backend (Express, body-parser), but having control over the name is useful regardless.

Secondly, it seemed that the name was perhaps being too tightly coupled with the presentation by virtue of suffixing `-{{ item.name }}` to the names. If we wanted these to be different, there's currently no way to achieve that.

Look forward to the outcomes of your discussion 👍

36degrees:
That makes sense - I can see how this change would be beneficial.

I'd like to try and avoid adding another similarly-named argument if at all possible - I don't think the distinction between `name` and `fieldName` is obvious and I think it is likely to cause confusion for users.

If we were to make `params.name` optional and only prefix `item.name` if it's set instead would that meet your use case?

e.g.

```njk
  "name": (params.name + "-" + item.name) if params.name else item.name,
```

Ollie

lhokktyn:
It could even be made a little simpler, and align more with other components by just adding a new `item.{}.text` parameter for controlling the input label, and use `item.{}.name` as the input element's name.

This is the way it's done in other macros, such as `checkboxes` (https://github.com/alphagov/govuk-frontend/tree/master/package/components/checkboxes#component-arguments)

Nice and simple!

36degrees:
@lhokktyn I think #857 has addressed your needs, so I'm going to close this now. Please feel free to re-open or create a new issue if we've missed something 👍 

Thanks again for raising this and helping to improve GOV.UK Frontend.

PR #831, opened 2018-06-25
Test heroku build (as it's failing)

igloosi:
- remove private:true from package.json


nickcolley:
As far as I know this field was already there, I just moved it?

igloosi:
removed it :)

PR #832, opened 2018-06-25
Update installing-with-npm.md

MartinDM:
Possible mistake in this path name.
all.scss is directly in the root, not in a subdirectory.

Assuming this example refers to importing via node_modules not copying it into the project locally that is.


igloosi:
Hi Martin,
great spot and thanks for the contribution.

PR #834, opened 2018-06-25
Proposal: Remove internal changelog entry requirement

nickcolley:
The primary use case for the CHANGELOG is to make sure users can react to any new release.

For example if there are breaking changes, they need to be able to action those.

Internal changes do not impact the running of GOV.UK Frontend in our users applications.

This commit removes this requirement to document changes, so we do not have to require,
contributors to update the CHANGELOG for a trivial change.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -71,14 +71,15 @@ which describes how we prefer git history and commit messages to read.
 
 If you open a GitHub pull request on this repo, please update `CHANGELOG` to reflect your contribution.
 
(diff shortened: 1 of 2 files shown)


nickcolley:
I'm a little in two minds:

The nice thing about current way we do the changelog is that if we want to recognise contributions in the changelog, we can do so when we do a release.

Would this impact that, potentially we could include that in the summary?

nickcolley:
Also if we made a mistake in a release, users would have a chance of seeing what we mistakenly labelled as an internal change.

PR #837, opened 2018-06-27
Use relative line-height

dashouse:
- Converts line heights to relative unit-less value

This needs to be tested on Samsung Internet browser to see if it fixes issue brought up in Accessibility audit.

### DIFF
--- a/src/settings/_typography-responsive.scss
+++ b/src/settings/_typography-responsive.scss
@@ -26,11 +26,11 @@ $govuk-typography-scale: (
   80: (
     null: (
       font-size: 53px,
-      line-height: 55px
+      line-height: 1.0377358491 // 55px
     ),
     tablet: (
       font-size: 80px,
-      line-height: 80px
+      line-height: 1 // 80px
     ),
     print: (
       font-size: 53pt,
@@ -40,11 +40,11 @@ $govuk-typography-scale: (
   48: (
     null: (
       font-size: 32px,
-      line-height: 35px
+      line-height: 1.09375 // 35px
     ),
     tablet: (
       font-size: 48px,
-      line-height: 50px
+      line-height: 1.0416666667 // 50px
     ),
     print: (
       font-size: 32pt,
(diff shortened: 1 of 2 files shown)


nickcolley:
Thanks for this Dave!

Could we break these into a RFC for the rem stuff and a straightforward PR for the uncontentious stuff?

hannalaakso:
I've tested this on the following and it looks good:
Chrome (latest) on Mac
Chrome (Samsung Android) with text resized to 200% in Chrome settings
Samsung Internet (Samsung Android) with text resized to "Huge" in phone settings
Firefox (latest) on Mac
Safari 11.1 on Mac
iOS 11 Safari with text resized to max size in phone settings
IE8-11 on Windows

To test, I made the examples in `examples/typography` longer in order to check text wrapping.

Just need a changelog entry for the PR.

nickcolley:
Firefox before and after:

screen shot 2018-06-28 at 11 11 45
screen shot 2018-06-28 at 11 11 40

src/settings/_typography-responsive.scss
```diff
@@ -26,11 +26,11 @@ $govuk-typography-scale: (
   80: (
     null: (
       font-size: 53px,
-      line-height: 55px
+      line-height: 1.0377358491 // 55px
```
36degrees: I do wonder if these would have been clearer if written as:

```scss
line-height: 55px / 53px;
```

Thoughts?

36degrees: Alternatively, we could have left these as pixel values in the font-map and calculated them in the typography-responsive mixin.

dashouse: I tried various ways of doing this calculation but kept running into issues...Should probably refactor so it's more "automatic" / less prone to typos. If we haven't released can we carry on?

nickcolley: If we don't expect this to change often, might be worth doing less here?

36degrees: Yep, happy with this. I've been playing locally with this too – looks like our font-size and line-height are ending up as strings somehow, which means using them in calculations doesn't work. It'd be great to be able to define everything in pixels and everything 'just work', but happy with this for now 👍

dashouse: I kind of got something interesting when spitting it out as a string if we change the output to `font: 16 / 20` and not define line height independently... but this didn't play very nicely with the other PR that needed two font-size declarations.

PR #839, opened 2018-06-27
RFC - Use rem with px fallback

dashouse:
## Changes

- Converts font-size from px to rem with px fallback

## Rems
Due to defining font-size in px a users choice to change their default font size (not zooming) on a OS or browser level is not respected. This is acceptable as we do not disable zooming, however it would be good to know if we can do more.

I've intentionally **not** used the "62.5%" trick on the `<html>`, we don't need an easy way to remember 24px = 2.4rem as the users will never interact with the rem font size directly, only the mixins and classes. GOV.UK Frontend also steers away from setting global styles.

### Rem concerns
As we are aware using relative font sizes in frontend will have issues with users app's with a base font size that is **not** 100% / 16px / default. This means including frontend in an existing app with a different base font size would leak / affect their apps existing styles. We have seen governments services using a variety of units and measurements on their root, 62.5%, 160%, 19px and 16px.

For example:

#### Default HTML (works as intended with a 16px base)
screen shot 2018-06-27 at 11 51 43

#### `html { font-size: 62.5% }` `body { font-size: 160% }` (as on GOV.UK)
screen shot 2018-06-27 at 11 53 20

#### `html { font-size: 19px }` `body { font-size: 100% }`
screen shot 2018-06-27 at 11 54 59

We could potentially allow the user to not include the `$font-size-rem` on output, however there are some concerns that this may feel like an "accessible or not" switch.

(p.s. I have left the commit to convert line-height to a relative unit-less unit in the PR to show the end state of this change)

### DIFF
--- a/src/helpers/_typography.scss
+++ b/src/helpers/_typography.scss
@@ -99,6 +99,7 @@
 
   @each $breakpoint, $breakpoint-map in $font-map {
     $font-size: map-get($breakpoint-map, "font-size") iff($important, !important);
+    $font-size-rem: map-get($breakpoint-map, "font-size-rem") iff($important, !important);
     $line-height: map-get($breakpoint-map, "line-height") iff($important, !important);
 
     // Sometimes we need to use a custom non-responsive line height for
@@ -109,6 +110,8 @@
 
     @if $breakpoint == null {
       font-size: $font-size;
+      // sass-lint:disable no-duplicate-properties
+      font-size: $font-size-rem;
       line-height: $line-height;
     } @elseif $breakpoint == "print" {
       @include govuk-media-query($media-type: print) {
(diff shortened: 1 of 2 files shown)


nickcolley:
Should the spacing scale relative to the font size?

dashouse:
https://medium.com/@vamptvo/pixels-vs-ems-users-do-change-font-size-5cfb20831773

src/helpers/_typography.scss
```diff
@@ -109,6 +110,8 @@
 
     @if $breakpoint == null {
       font-size: $font-size;
+      // sass-lint:disable no-duplicate-properties
```
36degrees: This disables this linter for the rest of the file. If you want to disable it for just the one line, add the comment to the end of the line as per https://github.com/sasstools/sass-lint#disabling-linters-via-source

dashouse:
Closing in favour of https://github.com/alphagov/govuk-frontend/pull/858

Issue #842, opened 2018-06-28
Alternative font path requires trailing slash


lfdebrux:
## What I'm trying to do

I've copied the fonts to the folder `/public/fonts` and I want them to be accessible from there, so I set the path in my SASS file to

```
$govuk-fonts-path: '/public/fonts';
```

as documented [here](https://github.com/alphagov/govuk-frontend/blob/master/docs/installation/installing-with-npm.md#alternative-solution).

## What I expect

Clients performing successful GET requests such as 

```
GET /public/fonts/bold-a2452cb66f-v1.woff2 HTTP/1.1" 200
```

## What I actually see

```
GET /public/fontsbold-a2452cb66f-v1.woff2 HTTP/1.1" 404 -
```

## Workaround

Adding a trailing slash to the path so that `$govuk-fonts-path: '/public/fonts'` works as expected.

## Suggested fix

Either the documentation should be updated to reflect the need for a trailing slash, or the paths should be fixed so it's not necessary.

igloosi:
Hi, @lfdebrux 
thanks for raising this. 
I've created a pull request to fix this

PR #843, opened 2018-06-28
Fix fonts path in npm install documentation

igloosi:
Internally, the variable has a trailing slash, but in the docs it was missed out.
Fixes: #842


docs/installation/installing-with-npm.md
```diff
@@ -254,7 +254,7 @@ To use different asset paths, also complete the below step(s).
   ``` SCSS
   // Include images from /images/govuk-frontend and fonts from /fonts
   $govuk-images-path: “/images/govuk-frontend”;
```
36degrees: The same is also true for the images path.

igloosi: true, updated 👍

PR #844, opened 2018-06-28
Update static asset serving solution for Express.

igloosi:
To avoid conflicts if users already have `/assets` mounted in their server.
It also ties into the change in Protoype Kit
https://github.com/alphagov/govuk_prototype_kit/pull/531


36degrees:
I'm not sure about this. I think the examples should be kept as bare-bones as possible to give the user the best chance of understanding what's going on in them.

Do we think we can leave it as-is for now and keep an eye out for this coming up in support channels?

If not, can we add this as a second example and give a reason why you might need to use this over the former?

nickcolley:
@igloosi but that feedback came from someone on our team, not from our users.

PR #845, opened 2018-06-28
Define size of table in `px` rather than `em`

hannalaakso:
This mirrors the changes made in https://github.com/alphagov/govuk-frontend/pull/491. The change to tables was omitted from that PR as it wasn't clear whether there was an accessibility consideration for using `em` for tables in Elements but @selfthinker has said that this doesn't appear to be the case.

As discussed with @dashouse, the change in this PR does very slightly increase the padding of cells on mobile viewport as the use of `em` meant the font size set in the table was used to calculate padding. `top` and `bottom` of cells increase by 1.5px which slightly increases the height of the table on mobile. (`padding-right` also increases very slightly but it won't change the width of the table).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,6 +68,16 @@
   global styles are enabled
   ([PR #877}])https://github.com/alphagov/govuk-frontend/pull/877
 
+- Define size of table in `px` rather than `em`
+
+  This brings the styling of tables inline with rest of GOV.UK Frontend which no longer uses `em` for measurements.
+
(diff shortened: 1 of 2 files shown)


36degrees:
I don't think this is a breaking (major) change. The height of a table is not 'known' – it's entirely dependant on the content within it. I can't really see how someone would be using tables in such a way that changing that height of the table (especially by such a small amount) makes any meaningful difference.

hannalaakso:
Agreed. I've changed the Changelog entry and filed this under Fixes.

Issue #768, opened 2018-06-06
GOV.UK Radios (Checkboxes) Items do not support Hints


nickcolley:
Raised by Alex T.

It's currently it is not possible to use hints in items as seen in https://www.gov.uk/log-in-file-self-assessment-tax-return/sign-in/prove-identity

Current work around:

```
{{ govukRadios({
  "idPrefix": "example",
  "name": "example",
  "fieldset": {
    "legend": {
      "text": "Create a new organisation"
    }
  },
  "items": [
    {
      "value": "Create new organisation",
      "html": "test <span class="govuk-hint">hint here.</span>",
    },
    {
      "value": "Merge organisations",
      "html": "test <span class="govuk-hint">hint here</span>",
      "checked": true
    }
  ]
}) }}
```

igloosi:
for this it would be good to understand what the relationship is between the label text and hint text in the way they're read with screenreaders

timpaul:
This is a relatively common pattern, so we should support it. In the self assessment example above the hint text is not part of the label, it's in a span immediately following it. However, on Voiceover at least it does get read out when you navigate between the radio options, which is probably what we want to happen.

36degrees:
> In the self assessment example above the hint text is not part of the label, it's in a span immediately following it. However, on Voiceover at least it does get read out when you navigate between the radio options, which is probably what we want to happen.

This works because the span is associated with the input using `aria-describedby`, which mirrors how we now associate hints and error messages with inputs generally.

PR #846, opened 2018-06-29
Allow an optional hint for each radio and checkbox item

igloosi:
It's currently it is not possible to use hints in items as seen in https://www.gov.uk/log-in-file-self-assessment-tax-return/sign-in/prove-identity in Nunjucks macros

This work allows for optional hint object to be passed to each item that outputs a hint component. Id of that hint component is set as `aria-describedby` on radio/checkbox into so that it's read by AT.

- Update template to allow for hint component to be passed
- Add example to yaml
- Add test to check that hint is present
- Update table of arguments with the new option
- Regenerate README

Direct urls:
https://govuk-frontend-review-pr-846.herokuapp.com/components/checkboxes/with-hints-on-items/preview
https://govuk-frontend-review-pr-846.herokuapp.com/components/radios/with-hints-on-items/preview

**Browser tests:**
<details>
<summary>Chrome 67 and  Safari 11.1.1 ✅</summary>
safari
chrome
</details>
<details>
<summary>IE8 ✅</summary>
ie8
</details>
<details>
<summary>IE11 ✅</summary>
ie11
</details>
<details>
<summary>Safari 9.3  iOS and Samsung Internet browser ✅</summary>
safari9-3
samsung-internet
</details>
<hr/>
<details>
<summary>AT tests:</summary>
VoiceOver on maOS
voiceover

JAWS 16, 17, 18 in IE11: Hint read
NVDA: Hint read
Talkback on Android: Hint read
Voiceover on iOS: Hint read
</details>

This fixes: #768

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,12 @@
   colours in their browser, rather than it appearing underlined all the time
   ([PR #926](https://github.com/alphagov/govuk-frontend/pull/926))
 
+- Allow for optional hint for each radio and checkbox item
+
+  You can now pass a hint object (or add in html) to each radio
+  and checkbox item to display the hint
+  ([PR #846](https://github.com/alphagov/govuk-frontend/pull/846))
+
 🔧 Fixes:
 
 - Replace conflicting `js-hidden` class used within the tabs component with a new modifier class.
--- a/src/components/checkboxes/checkboxes.yaml
+++ b/src/components/checkboxes/checkboxes.yaml
@@ -43,6 +43,26 @@ examples:
         value: irish
         text: Irish
 
+- name: with hints on items
+  data:
+    fieldset:
+      legend:
+        text: How do you want to sign in?
+        isPageHeading: true
+    items:
+      - name: gateway
+        id: government-gateway
+        value: gov-gateway
+        text: Sign in with Government Gateway
+        hint:
+          text: You’ll have a user ID if you’ve registered for Self Assessment or filed a tax return online before.
+      - name: verify
+        id: govuk-verify
+        value: gov-verify
+        text: Sign in with GOV.UK Verify
+        hint:
+          text: You’ll have an account if you’ve already proved your identity with either Barclays, CitizenSafe, Digidentity, Experian, Post Office, Royal Mail or SecureIdentity.
+
 - name: with disabled item
   data:
     name: colours
--- a/src/components/checkboxes/template.njk
+++ b/src/components/checkboxes/template.njk
@@ -48,11 +48,14 @@
     {% set id = item.id if item.id else idPrefix + "-" + loop.index %}
     {% set name = item.name if item.name else params.name %}
     {% set conditionalId = "conditional-" + id %}
+    {% set hasHint = true if item.hint.text or item.hint.html %}
+    {% set itemHintId = id + '-item-hint' %}
     <div class="govuk-checkboxes__item">
       <input class="govuk-checkboxes__input" id="{{ id }}" name="{{ name }}" type="checkbox" value="{{ item.value }}"
       {{-" checked" if item.checked }}
       {{-" disabled" if item.disabled }}
-      {%- if item.conditional %} data-aria-controls="{{ conditionalId }}"{% endif -%}>
+      {%- if item.conditional %} data-aria-controls="{{ conditionalId }}"{% endif -%}
+      {%- if hasHint %} aria-describedby="{{ itemHintId }}"{% endif -%}>
       {{ govukLabel({
         html: item.html,
         text: item.text,
@@ -60,6 +63,15 @@
         attributes: item.label.attributes,
         for: id
       }) | indent(6) | trim }}
+      {%- if hasHint %}
+      {{ govukHint({
+        id: itemHintId,
+        classes: 'govuk-checkboxes__hint',
+        attributes: item.hint.attributes,
+        html: item.hint.html,
+        text: item.hint.text
+      }) | indent(6) | trim }}
+      {%- endif %}
     </div>
     {% if item.conditional %}
       <div class="govuk-checkboxes__conditional{% if not item.checked %} govuk-checkboxes__conditional--hidden{% endif %}" id="{{ conditionalId }}">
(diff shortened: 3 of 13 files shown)


src/components/checkboxes/template.test.js
```diff
@@ -255,6 +255,36 @@ describe('Checkboxes', () => {
     })
   })
 
+  it('render hint text', () => {
+    const $ = render('checkboxes', {
```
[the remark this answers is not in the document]
igloosi: I figure that checking for input and hint relationship are correct to that it gets announced.
I could just test that it renders, but that not that useful.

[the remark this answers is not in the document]
36degrees: It's definitely useful to test that, but there's still a lot going on here for one test:

- You could take out a lot of the lines here (for example the entire first item) and the test would still pass, which makes it harder for others coming in to try and understand what's _actually being tested_.
- There are three assertions being made, two of which aren't covered by the test description. If this test were to fail because the `aria-describedby` isn't what we expect, the description doesn't help you to understand why it's failing.

If you look at e.g. https://github.com/alphagov/govuk-frontend/blob/master/src/components/input/template.test.js#L85-L115:

- there's only one assertion per test
- the description of each test describes the assertion that's being made
- we're asserting on the behaviour we actually care about – we don't actually care what the `id` or `aria-describedby` values are – as long as the id is included in `aria-describedby`. For example, if we added another element which we wanted to add to `aria-describedby` then this test as written would break (incorrectly) and we'd have to come back and fix it.
- there's very little you could delete from each test without breaking it, so it's clear what the test is doing.

Does that make sense?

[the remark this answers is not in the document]
igloosi: yep. i've updated the tests

src/components/checkboxes/checkboxes.yaml
```diff
@@ -43,6 +43,26 @@ examples:
...
+  data:
+    fieldset:
+      legend:
+        text: How do you want to sign in?
+        isPageHeading: true
+    items:
+      - name: gateway
+        id: government-gateway
+        value: gov-gateway
+        html: Sign in with Government Gateway
```
36degrees: Should be text

src/components/checkboxes/checkboxes.yaml
```diff
@@ -43,6 +43,26 @@ examples:
...
+        hint:
+          text: You’ll have a user ID if you’ve registered for Self Assessment or filed a tax return online before.
+      - name: verify
+        id: govuk-verify
+        value: gov-verify
+        html: Sign in with GOV.UK Verify
```
36degrees: Should be text

src/components/radios/radios.yaml
```diff
@@ -100,6 +100,24 @@ examples:
...
+  data:
+    idPrefix: gov
+    name: gov
+    fieldset:
+      legend:
+        text: How do you want to sign in?
+        isPageHeading: true
+    items:
+      - value: gateway
+        html: Sign in with Government Gateway
```
36degrees: Should be text

src/components/radios/radios.yaml
```diff
@@ -100,6 +100,24 @@ examples:
...
+        hint:
+          text: You'll have a user ID if you've registered for Self Assessment or filed a tax return online before.
+      - value: verify
+        html: Sign in with GOV.UK Verify
```
36degrees: Should be text

src/components/radios/template.test.js
```diff
@@ -198,6 +198,60 @@ describe('Radios', () => {
       expect($lastInput.attr('checked')).toEqual('checked')
     })
 
+    describe('when they include a hint', () => {
+      it('it renders the hint text', () => {
+        const $ = render('radios', {
+          idPrefix: 'gov',
```
36degrees: Is this required (in all three tests)?

igloosi:
updated

src/components/checkboxes/template.njk
```diff
@@ -48,18 +48,30 @@
     {% set id = item.id if item.id else idPrefix + "-" + loop.index %}
     {% set name = item.name if item.name else params.name %}
     {% set conditionalId = "conditional-" + id %}
+    {% set hasHint = true if item.hint.text or item.hint.html %}
```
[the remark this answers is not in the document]
36degrees: We don't use the item prefix for anything else in this block (e.g. `name`, not `itemName`) – do we think this is different?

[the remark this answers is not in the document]
36degrees: Whatever we do, it'd be good to be consistent within this block – id, name and conditionalId _aren't_ prefixed. If we think it's useful to have that prefix, we should do it for all of them.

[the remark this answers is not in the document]
igloosi: we have `itemHintId` because we already have `hintId`, the whole group's hint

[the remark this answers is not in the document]
hannalaakso: In that case they should probably all be prefixed. It isn't clear from the context why `itemHintId` is named differently (to avoid clash with the group level hint) 🤔

alex-ju:
I was hoping to have this in the last release. Do you think it will be in v1.2.0?

nickcolley:
@alex-ju we can try to do a quicker release if you're blocked by this for sure.

alex-ju:
not a blocker, and even if so I can reproduce it before the release/updating to the latest version, but it's definitely something that we need short-term.

src/components/checkboxes/template.test.js
```diff
@@ -255,6 +255,60 @@ describe('Checkboxes', () => {
...
+            text: 'This is text',
+            id: 'item-id',
+            hint: {
+              text: 'This is a hint'
+            }
+          }
+        ]
+      })
+
+      expect($('.govuk-checkboxes__input').attr('aria-describedby')).toBe('item-id-item-hint')
```
nickcolley: What happens if a user does not supply an 'id'?

CHANGELOG.md
```diff
@@ -75,7 +81,7 @@
   `headingLevel: <number>` parameter. Default is `2`.
   ([PR #853](https://github.com/alphagov/govuk-frontend/pull/853))
 
-- Update date input component  
+- Update date input component
```
igloosi: can we keep this in here?

nickcolley: Yeah, we need to figure out how to get our editors stripping whitespace consistently, but that should not block this PR

Issue #813, opened 2018-06-21
Open graph image path needs to be customisable


igloosi:
At the moment the image in the template is set to
`<meta property="og:image" content="/assets/images/govuk-opengraph-image.png">`
and it doesnt resolve to the hostname.

As it's outside the `head` block so you have to define it in the head block again, but then you get two instances.

PR #847, opened 2018-06-29
Allow og:image meta tag to be customisable

igloosi:
Moving the meta tag into the `head` block so that it can be overwritten, as the image path for it needs to be an absolute path.

Fixes: #813

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,11 @@
   You can now provide a divider item (e.g "or") to separate items
   ([PR #849](https://github.com/alphagov/govuk-frontend/pull/849))
 
+- Allow og:image meta tag url to be set independantly
+  Image url for the opengraph image needs to be absolute and
+  can now be overwritten by setting the `assetUrl` variable.
+  ([PR #847](https://github.com/alphagov/govuk-frontend/pull/847))
+
 🔧 Fixes:
 
 - Replace conflicting `js-hidden` class used within the tabs component with a new modifier class.
@@ -75,7 +80,7 @@
   `headingLevel: <number>` parameter. Default is `2`.
   ([PR #853](https://github.com/alphagov/govuk-frontend/pull/853))
 
-- Update date input component  
+- Update date input component
 
   Allow the `name` and `id` attributes to be passed individually for each input item.
 
--- a/src/template.njk
+++ b/src/template.njk
@@ -1,6 +1,8 @@
 {% from "./components/skip-link/macro.njk" import govukSkipLink %}
 {% from "./components/header/macro.njk" import govukHeader %}
 {% from "./components/footer/macro.njk" import govukFooter %}
+{# specify absolute url for the static assets folder e.g. http://wwww.domain.com/assets #}
+{% set assetUrl = assetUrl | default(assetPath) %}
 <!DOCTYPE html>
 <html lang="{{ htmlLang | default('en') }}" class="govuk-template {{ htmlClasses }}">
   <head>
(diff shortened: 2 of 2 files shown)


src/template.njk
```diff
@@ -18,10 +18,11 @@
       <link rel="apple-touch-icon" href="{{ assetPath | default('/assets') }}/images/govuk-apple-touch-icon.png">
     {% endblock %}
 
-    {% block head %}{% endblock %}
-
+    {% block head %}
```
nickcolley: This is a breaking change since before even when you set head you'd get the default open graph image.

Could this be solved by something like

```html
<meta property="og:image" content="{{ opengraphAssetPath | default('') }}{{ assetPath | default('/assets') }}/images/govuk-opengraph-image.png">
```

igloosi: tru dat. updated so that the variable can just be set

src/template.njk
```diff
@@ -19,9 +19,9 @@
     {% endblock %}
 
     {% block head %}{% endblock %}
-
     {# The default og:image is added below head so that scrapers see any custom metatags first, and this is just a fallback #}
-    <meta property="og:image" content="{{ assetPath | default('/assets') }}/images/govuk-opengraph-image.png">
+    {# image path needs to absolute e.g. http://wwww.domain.com/.../govuk-opengraph-image.png #}
```
36degrees: Path refers to just the part after the domain - i.e. just the `/.../govuk-opengraph-image.png` bit. How about `assetUrl`?

igloosi: sure. updated

nickcolley:
Would you mind raising an issue after this is merged on the Design System repo to include this new feature in the guidance?

src/template.njk
```diff
@@ -19,9 +19,9 @@
     {% endblock %}
 
     {% block head %}{% endblock %}
-
     {# The default og:image is added below head so that scrapers see any custom metatags first, and this is just a fallback #}
-    <meta property="og:image" content="{{ assetPath | default('/assets') }}/images/govuk-opengraph-image.png">
+    {# image url needs to be absolute e.g. http://wwww.domain.com/.../govuk-opengraph-image.png #}
+    <meta property="og:image" content="{{ assetUrl | default('/assets') }}/images/govuk-opengraph-image.png">
```
nickcolley: I think the default needs to be 'assetPath' otherwise this is a breaking change still?

nickcolley: maybe `{{ openGraphImagePath | assertPath | default('/assets') }}`

not sure

igloosi: it doesn't work with assetPath anyway, so not sure it's a breaking change

nickcolley: Fair point, but consider if someone set their 'assetPath' to a absolute path, it would be working for those users?

igloosi: modified so that then default value for assetUrl is assetPath, insuring there it is not a breaking change if anyone specified an absolute url for the assets

CHANGELOG.md
```diff
@@ -28,6 +28,15 @@
   ([PR #837](https://github.com/alphagov/govuk-frontend/pull/837) and
    [PR #848](https://github.com/alphagov/govuk-frontend/pull/848))
 
+- Use relative line-height
```
36degrees: This shouldn't be part of this PR.

CHANGELOG.md
```diff
@@ -28,6 +28,15 @@
...
+  Update typography styles to use relative, unitless line-height
+  ([PR #837](https://github.com/alphagov/govuk-frontend/pull/837))
+
+- Allow og:image meta tag to be customisable
```
36degrees: It was already customisable by modifying `assetPath` – I think we need to be clearer as to what the change is here.

nickcolley:
Original code: https://github.com/alphagov/govuk_template/blob/master/source/views/layouts/govuk_template.html.erb#L36

igloosi:
i've updated the wording

nickcolley:
Thinking about this more, if this is solved by setting 'assetPath' to an absolute url do we need to do this at all?

igloosi:
we can't set the absolute path for example in the design system (as we have two different urls)
but if thats not of a concern and we just leave the declaration twice as we have now, then there is no need for this indeed

PR #848, opened 2018-06-29
Convert line-height from pixel values

36degrees:
Rather than having to declare the line-heights as relative values (such as 1.5) in the font map, allow them to be specified in pixels and convert them to relative unitless values within our typography mixin.

This provides a much clearer way to define line heights without the need for comments explaining how the values are calculated, whilst retaining the benefits of using relative unitless line-heights.

### DIFF
--- a/src/helpers/_typography.scss
+++ b/src/helpers/_typography.scss
@@ -103,7 +103,8 @@
 
     // Convert line-heights specified in pixels into a relative value, unless
     // they are already unit-less (and thus already treated as relative values)
-    @if not unitless($line-height) {
+    // or the units do not match the units used for the font size.
+    @if not unitless($line-height) and unit($line-height) == unit($font-size) {
       $line-height: $line-height / $font-size;
     }
 
--- a/src/helpers/_typography.scss
+++ b/src/helpers/_typography.scss
@@ -98,14 +98,26 @@
   $font-map: map-get($govuk-typography-scale, $size);
 
   @each $breakpoint, $breakpoint-map in $font-map {
-    $font-size: map-get($breakpoint-map, "font-size") iff($important, !important);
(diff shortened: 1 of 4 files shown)


src/helpers/_typography.scss
```diff
@@ -98,15 +98,27 @@
...
 
   @each $breakpoint, $breakpoint-map in $font-map {
-    $font-size: map-get($breakpoint-map, "font-size") iff($important, !important);
-    $line-height: map-get($breakpoint-map, "line-height") iff($important, !important);
+    $font-size: map-get($breakpoint-map, "font-size");
+    $line-height: map-get($breakpoint-map, "line-height");
+
+    // Convert line-heights specified in pixels into a relative value, unless
+    // they are already unit-less (and thus already treated as relative values)
+    @if not unitless($line-height) {
```
nickcolley: If you set your line-height to something other than 'px' would it this fail?

36degrees: Yep, because Sass doesn't know how to e.g. divide pixels by em.

We've got a few options…

- let Sass' error handling catch it
- detect if the units do not match and do nothing (output line-height as defined in the map)
- detect if the units do not match and throw our own error

Not sure what the best option is. Can you think of cases where you'd intentionally want to use define font-size and line-height in different units?

nickcolley: I suppose the only case I'm thinking of is this could be confusing for people making their own font maps, but if it's clearly documented maybe it'll be fine.

36degrees:
@nickcolley I've made the line-height stuff into its own function and attempted to make it handle those edge cases better. Can you please re-review?

nickcolley:
Good use of tests 👍

Issue #729, opened 2018-05-29
Add "Or" before the last radio button in a radio group


gavinwye:
The Marcos for radio buttons doesn't allow the commonly used pattern of using the word "Or" to separate the last value from the previous values as shown in the screenshot attached.

This should be added to the macro.

![image 1](https://user-images.githubusercontent.com/868772/40672237-1703fe10-6366-11e8-8be3-de5dde5f3d6b.png)

PR #849, opened 2018-07-01
Allow for an optional divider between radio items

igloosi:
It used to be a feature in [GOV.UK Elements](https://govuk-elements.herokuapp.com/form-elements/#form-radio-buttons) and it's not currently available in our macros.

Direct url: https://govuk-frontend-review-pr-849.herokuapp.com/components/radios/with-a-divider/preview

This PR:
- Adds an example
- Updates the component template
- Adds a test to check the divider item gets output
- Updates README

Fixes: #729 

Works fine with conditional reveal as well.

<details>
<summary>Screenshots</summary>

IE 8

IE 9

IE 11

Firefox 60

Safari 9.3

</details>

### DIFF
--- a/src/components/radios/_radios.scss
+++ b/src/components/radios/_radios.scss
@@ -139,6 +139,14 @@
     }
   }
 
+  .govuk-radios__divider {
+    $govuk-divider-size: $govuk-radios-size !default;
+    @include govuk-font($size: 19);
+    width: $govuk-divider-size;
+    margin-bottom: govuk-spacing(2);
+    text-align: center;
+  }
+
   $conditional-border-width: $govuk-border-width-mobile;
   // Calculate the amount of padding needed to keep the border centered against the radios.
   $conditional-border-padding: ($govuk-radios-size / 2) - ($conditional-border-width / 2);
--- a/src/components/radios/radios.yaml
+++ b/src/components/radios/radios.yaml
@@ -100,6 +100,22 @@ examples:
           <span class="govuk-heading-s govuk-!-margin-bottom-1">Part 3 of the Housing Act 2004</span>
           For properties that are within a geographical area defined by a local council
 
+- name: with a divider
+  data:
+    idPrefix: example-divider
+    name: example
+    fieldset:
+      legend:
+        text: How do you want to sign in?
+    items:
+      - value: governement-gateway
+        text: Use Government Gateway
+      - value: govuk-verify
+        text: Use GOV.UK Verify
+      - divider: or
+      - value: create-account
+        text: Create an account
+
 - name: without fieldset
   data:
     name: colours
--- a/src/components/radios/template.njk
+++ b/src/components/radios/template.njk
@@ -47,6 +47,9 @@
     {% for item in params.items %}
     {% set id = item.id if item.id else idPrefix + "-" + loop.index %}
     {% set conditionalId = "conditional-" + id %}
+    {%- if item.divider %}
+    <div class="govuk-radios__divider">{{ item.divider }}</div>
+    {%- else %}
     <div class="govuk-radios__item">
       <input class="govuk-radios__input" id="{{ id }}" name="{{ params.name }}" type="radio" value="{{ item.value }}"
       {{-" checked" if item.checked }}
@@ -64,6 +67,7 @@
       <div class="govuk-radios__conditional" id="{{ conditionalId }}">
         {{ item.conditional.html | safe }}
       </div>
+    {% endif %}
     {% endif %}
     {% endfor %}
   </div>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,11 @@
 
   ([PR #N](https://github.com/alphagov/govuk-frontend/pull/N))
 
+- Allow for optional divider between radio items
+
+  You can now provide a divider item (e.g "or") to separate items
+  ([PR #849](https://github.com/alphagov/govuk-frontend/pull/849))
+
 🔧 Fixes:
 
 - Replace conflicting `js-hidden` class used within the tabs component with a new modifier class.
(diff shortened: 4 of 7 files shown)


src/components/radios/radios.yaml
```diff
@@ -78,6 +78,22 @@ examples:
...
+    name: example
+    fieldset:
+      legend:
+        text: Have you changed your name?
+    items:
+      - value: governement-gateway
+        text: Use Government Gateway
+      - value: govuk-verify
+        text: 'Use GOV.UK Verify'
+      - divider: 'or'
```
igloosi: we could follow the same rules as with all our text params and allow for either txt or html, but surely it should only be text

igloosi:
@dashouse question about text alignment in relationship to the radio. in elements it's aligned to the left

dashouse:
Design wise 👍

src/components/radios/_radios.scss
```diff
@@ -130,6 +130,13 @@
     }
   }
 
+  .govuk-radios__divider {
+    @include govuk-font($size: 19);
+    display: block;
+    margin-bottom: govuk-spacing(2);
+    padding: 0 0 0 ($govuk-radios-size / 4);
```
36degrees: This is _visually_ centring the text, but it's relying on the string being 'or' - if you use a different string then it doesn't work. What about making this element the same width as the radio buttons / checkboxes and then use `text-align: center`? That way it should work correctly for short strings, and automatically fall back to left-aligning if the string is long enough that centring it doesn't make sense.

src/components/radios/radios.yaml
```diff
@@ -78,6 +78,22 @@ examples:
           <span class="govuk-heading-s govuk-!-margin-bottom-1">Part 3 of the Housing Act 2004</span>
           For properties that are within a geographical area defined by a local council
 
+- name: with divider
+  data:
+    idPrefix: example-divider
+    name: example
+    fieldset:
+      legend:
+        text: Have you changed your name?
```
36degrees: Can we make this example make sense? The answers don't relate to the question.

src/components/radios/template.njk
```diff
@@ -44,9 +44,12 @@
   <div class="govuk-radios {%- if params.classes %} {{ params.classes }}{% endif %}"
     {%- for attribute, value in params.attributes %} {{ attribute }}="{{ value }}"{% endfor %}
     {%- if isConditional %} data-module="radios"{% endif -%}>
-    {% for item in params.items %}
+    {% for item in params.items %}
```
36degrees: Trailing whitespace

src/components/radios/template.njk
```diff
@@ -59,12 +62,13 @@
         attributes: item.label.attributes,
         for: id
       }) | indent(6) | trim }}
-    </div>
+    </div>
```
36degrees: Trailing whitespace

src/components/radios/_radios.scss
igloosi: the aim is for the left alignment to be where the line is in the conditional reveal. i've update the calculation
screen shot 2018-07-03 at 09 14 23
screen shot 2018-07-03 at 09 14 18

igloosi: we agreed to go with Ollie's suggestions

src/components/radios/index.njk
```diff
@@ -190,6 +190,20 @@ Let users select a single option from a list.
...
+        text: 'items.{}.divider'
+      },
+      {
+        text: 'text'
+      },
+      {
+        text: 'No'
+      },
+      {
+        html: 'Optional divider to separate options, e.g "or".'
```
36degrees: I think it needs to be clearer how to use this – it's not obvious for example that specifying 'divider' means that all other options will be ignored.

src/components/radios/template.njk
```diff
@@ -47,6 +47,9 @@
     {% for item in params.items %}
     {% set id = item.id if item.id else idPrefix + "-" + loop.index %}
     {% set conditionalId = "conditional-" + id %}
+    {%- if item.divider %}
+    <span class="govuk-radios__divider">{{ item.divider }}</span>
```
hannalaakso: Should this just be a `div`, instead of `span` that has `display: block` set. It would make it consistent with the markup of the radios.

igloosi: it can be a div

src/components/radios/_radios.scss
```diff
@@ -130,6 +130,15 @@
     }
   }
 
+  .govuk-radios__divider {
+    $govuk-divider-size: $govuk-radios-size !default;
+    @include govuk-font($size: 19);
+    display: block;
```
36degrees: This is redundant now that the divider is a `div`.

src/components/radios/_radios.scss
```diff
@@ -138,6 +138,14 @@
     }
   }
 
+  .govuk-radios__divider {
+    $govuk-divider-size: $govuk-radios-size !default;
```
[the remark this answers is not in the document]
36degrees: That was the original approach – see related discussion in https://github.com/alphagov/govuk-frontend/pull/849#discussion_r199713140

When the divider is longer than or the text will overflow to the right, like this:

screen shot 2018-07-11 at 12 09 30

I don't think we'd tested with users changing their text size though – good call. It seems to work though, having enabled `$govuk-typography-use-rem: true;` in app.scss:

**Firefox, with size set to 28 in appearance settings:**

screen shot 2018-07-11 at 12 14 49

**Safari, with 'Never use font sizes smaller than 24' configured under 'Advanced':**
screen shot 2018-07-11 at 12 15 53

**Chrome, with 'Font size' set to very large in appearance settings:**

screen shot 2018-07-11 at 12 16 34

igloosi:
requires rebasing and changelog due to 1.1.1 release

PR #853, opened 2018-07-02
Allow the panel component title heading level to be customisable

igloosi:
This modification allows it panel component title to be any heading level by specifying `headingLevel` parameter
Default is `2`.

This PR
- Adds test
- Adds an example
- Updates component table of arguments
- Generates an updated README

Adresses: https://github.com/alphagov/govuk-design-system/issues/413

### DIFF
--- a/src/components/panel/template.njk
+++ b/src/components/panel/template.njk
@@ -1,9 +1,11 @@
+{#- TODO: Change default title heading level to 1 - https://github.com/alphagov/govuk-frontend/issues/864 -#}
+{% set headingLevel = params.headingLevel if params.headingLevel else 2 %}
 <div class="govuk-panel govuk-panel--confirmation
   {%- if params.classes %} {{ params.classes }}{% endif %}"
   {%- for attribute, value in params.attributes %} {{attribute}}="{{value}}"{% endfor %}>
-  <h2 class="govuk-panel__title">
+  <h{{ headingLevel }} class="govuk-panel__title">
     {{ params.titleHtml | safe if params.titleHtml else params.titleText }}
-  </h2>
+  </h{{ headingLevel }}>
   {% if params.html or params.text %}
   <div class="govuk-panel__body">
     {{ params.html | safe if params.html else params.text }}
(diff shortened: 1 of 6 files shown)


src/components/panel/template.njk
```diff
@@ -1,9 +1,10 @@
+{% set titleLevel = params.titleLevel if params.titleLevel else 1 %}
```
nickcolley: If the default is 1, this will be a breaking change, maybe we should default it to 2 and add a deprecation notice?

igloosi: In most cases it will (should) be an h1 so i guess a breaking change.

igloosi: filed under Breaking change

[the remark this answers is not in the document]
igloosi: thought about that. we have `titleText`, so 1st thought was `titleHeadingLevel` to keep naming related, but that is wordy.
I also toyed with the idea to actually move the title to an object, to have:
```
title: {
 text:
 html: 
 headingLevel: 
}
```

36degrees: If we're going to make this a breaking change we should probably hold this for now and batch it up with other breaking changes.

nickcolley: We can make do this work and default to 2, then update the examples in the Design System to explicitly set the level as 1.

We'd then deprecate this behaviour and release the default as 1 in a breaking release later on?

Maybe I'm overcomplicating it.

src/components/panel/template.test.js
```diff
@@ -30,6 +30,22 @@ describe('Panel', () => {
     expect(panelTitle).toEqual('Application &lt;strong&gt;not&lt;/strong&gt; complete')
   })
 
+  it('renders title as h1 (as the default heading level)', () => {
```
36degrees: h2

joelanman:
sorry to comment late, but didn't we want the default to be h1? As that's the common case for this panel? I thought I'd read we were going to do that somewhere?

igloosi:
@joelanman making it h1 in the next breaking release: https://github.com/alphagov/govuk-frontend/milestone/2

joelanman:
@igloosi ah cool thanks

PR #854, opened 2018-07-02
Show focus outlines when customising colours settings on Radios and Checkboxes

nickcolley:
Adds invisible outlines that will be shown when colours are overridden.

## Before overriden colours
before-checkboxes

## After overridden colours
after-checkboxes

## Before overridden colours
before-radios

## After overridden colours
after-radios

<details>
<summary>Cross browser testing</summary>

## Internet Explorer 11
screen shot 2018-07-09 at 10 58 20
screen shot 2018-07-09 at 10 59 07

## Internet Explorer 10
screen shot 2018-07-09 at 11 00 57
screen shot 2018-07-09 at 11 01 33

## Internet Explorer 9
screen shot 2018-07-09 at 11 02 54
screen shot 2018-07-09 at 11 02 24

## Internet Explorer 8
screen shot 2018-07-09 at 11 05 34
screen shot 2018-07-09 at 11 04 38

# Firefox latest, Chrome latest, Chrome Android latest, Safari latest, Safari iOS latest
No change

</details>

### DIFF
--- a/src/components/radios/_radios.scss
+++ b/src/components/radios/_radios.scss
@@ -10,6 +10,9 @@
 @include govuk-exports("govuk/component/radios") {
   $govuk-radios-size: govuk-spacing(7);
   $govuk-radios-label-padding-left-right: govuk-spacing(3);
+  // When the default focus width is used on a curved edge it looks visually smaller.
+  // So for the circular radios we bump the default to make it look visually consistent.
+  $govuk-radios-focus-width: $govuk-focus-width + 1px;
 
   .govuk-radios__item {
     @include govuk-font($size: 19);
@@ -99,7 +102,12 @@
 
   // Focused state
   .govuk-radios__input:focus + .govuk-radios__label::before {
-    box-shadow: 0 0 0 4px $govuk-focus-colour;
+    // Since box-shadows are removed when users customise their colours
+    // We set a transparent outline that is shown instead.
+    // https://accessibility.blog.gov.uk/2017/03/27/how-users-change-colours-on-websites/
+    outline: $govuk-focus-width solid transparent;
+    outline-offset: $govuk-focus-width;
+    box-shadow: 0 0 0 $govuk-radios-focus-width $govuk-focus-colour;
   }
 
   // Selected state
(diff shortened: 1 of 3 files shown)


src/components/radios/_radios.scss
```diff
@@ -10,6 +10,7 @@
 @include govuk-exports("govuk/component/radios") {
   $govuk-radios-size: govuk-spacing(7);
   $govuk-radios-label-padding-left-right: govuk-spacing(3);
+  $govuk-radios-focus-width: 4px;
```
36degrees: Assuming we _need_ to add 1 pixel can this be `$govuk-focus-width + 1` so that we understand where this value came from?

src/components/radios/_radios.scss
```diff
@@ -99,7 +100,12 @@
 
   // Focused state
   .govuk-radios__input:focus + .govuk-radios__label::before {
-    box-shadow: 0 0 0 4px $govuk-focus-colour;
+    top: -($govuk-radios-focus-width);
```
36degrees: I think this makes sense, but could definitely do with a comment to explain what's going on here.

36degrees:
This is ace 👏 – @dashouse are you happy with the focus behaviour from a design perspective?

dashouse:
nice!

src/components/radios/_radios.scss
```diff
@@ -99,7 +102,12 @@
 
   // Focused state
   .govuk-radios__input:focus + .govuk-radios__label::before {
-    box-shadow: 0 0 0 4px $govuk-focus-colour;
+    // Since box-shadows are removed when users customise their colours
+    // We set a transparent outline that is shown instead.
+    // https://accessibility.blog.gov.uk/2017/03/27/how-users-change-colours-on-websites/
+    outline: $govuk-focus-width solid transparent;
+    outline-offset: $govuk-focus-width;
```
timja-kainos: I assume you're aware but outline-offset doesn't exist in IE

nickcolley: Yeah, the offset is for mainly aesthetic reasons as this approach does not rely on the gap between the focus and control, having the offset just makes it a bit more clearer that this is a focus and not a thicker border.

PR #855, opened 2018-07-02
Add outline to tag for customised colours

nickcolley:
Adds a border to the tag element so it makes more sense when colours are overridden.

The screenshots I've included here show the tag component when used in the phase banner component.

Uses relative spacing to keep alignment when the label is on a smaller screen.

## Before overriden colours
screen shot 2018-07-02 at 12 48 15

## After overriden colours
screen shot 2018-07-02 at 12 48 10

## Before regular colours
screen shot 2018-07-02 at 12 47 18

## After regular colours
screen shot 2018-07-02 at 12 47 22

### Before overriden colours
screen shot 2018-07-02 at 12 53 56

### After overriden colours
screen shot 2018-07-02 at 12 53 52

### Before regular colours
screen shot 2018-07-02 at 12 53 47

### After regular colours
screen shot 2018-07-02 at 12 53 41

### DIFF
--- a/src/components/tag/_tag.scss
+++ b/src/components/tag/_tag.scss
@@ -9,6 +9,12 @@
     display: inline-block;
     padding: govuk-spacing(1) 8px 0;
 
+    // When a user customises their colours often the background is removed,
+    // by adding a outline we ensure that the tag component still keeps it's meaning.
+    // https://accessibility.blog.gov.uk/2017/03/27/how-users-change-colours-on-websites/
+    outline: 2px solid transparent;
+    outline-offset: -2px;
+
     color: govuk-colour("white");
     background-color: govuk-colour("blue");
     letter-spacing: 1px;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,6 +83,12 @@
   they should see a focus state on both Radios and Checkboxes.
   ([PR #854](https://github.com/alphagov/govuk-frontend/pull/854))
 
+- Add outline to tag for customised colour users
+
+  Now when a [user customises their colours](https://accessibility.blog.gov.uk/2017/03/27/how-users-change-colours-on-websites/),
+  the tag component still keeps it's meaning.
+  ([PR #855](https://github.com/alphagov/govuk-frontend/pull/855))
+
 🏠 Internal:
 
 - Fix Design System url in package READMEs and review app


src/components/tag/_tag.scss
```diff
@@ -4,10 +4,15 @@
 
 @include govuk-exports("govuk/component/tag") {
   .govuk-tag {
-    @include govuk-font($size: 16, $weight: bold, $line-height: 1.25);
+    @include govuk-font($size: 16, $weight: bold, $line-height: 1);
 
     display: inline-block;
-    padding: govuk-spacing(1) 8px 0;
+    padding-top: .3125em;
```
36degrees: Given we've got #845 open to try and remove our only current usage of `em` for padding, do we want to be introducing it again here, or should we be using pixels?

Also, can this be a single `padding` rule as the values are all the same?

nickcolley: I'd prefer to use relative units since it means when the tag is resized the padding is in proportion. Will show screenshots, before and after PX.

The padding results are not all the same, but I'll see if I can reduce the repeitition (I personally do not like shorthand since it's like doing `function(many, arguments, order, isimportant)` vs `function({ order: 'not important' })` etc)

36degrees: I think other components like the button keep the same dimensions when the font-size reduces (because the line-height is constant) so I'd expect this component to work in the same way… especially as we want to keep the /5 vertical rhythm.

Totally failed to notice that padding-bottom was different so… that's fine 👍

nickcolley: OK, good to know, thanks Ollie.

nickcolley:
@36degrees I've adjusted this PR based on your comments and ended up with something very similar to the original:

tag-colours-change

src/components/tag/_tag.scss
```diff
@@ -4,10 +4,12 @@
 
 @include govuk-exports("govuk/component/tag") {
   .govuk-tag {
-    @include govuk-font($size: 16, $weight: bold, $line-height: 1.25);
+    @include govuk-font($size: 16, $weight: bold, $line-height: 1);
 
     display: inline-block;
-    padding: govuk-spacing(1) 8px 0;
+    padding: 4px 6px 1px;
```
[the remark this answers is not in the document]
nickcolley: Could do yeah, would prefer to use borders since we can 'see' them. Up to you.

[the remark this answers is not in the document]
nickcolley: I'll change this, it also means the modifiers get outlines for free without additional CSS :)

nickcolley:
OK, went with the transparent outline approach in the end.

## Before outline

screen shot 2018-07-09 at 13 15 07

## After outline

screen shot 2018-07-09 at 13 15 03

36degrees:
👏

PR #856, opened 2018-07-02
Make the warning icon un-selectable

36degrees:
This prevents the exclamation mark from being included when the warning text is copied - for example if a user is copying a page that includes it.

## Before

before

## After

after


nickcolley:
👨‍🎨

PR #857, opened 2018-07-02
Update date input component

### LINKED ISSUE #468: govukDateInput macro could have default `items`
The majority of the time, you want Day, Month, Year, so we could just have that as the default without the user having to specify `items`

### LINKED ISSUE #836: Date input macro styling breaks if item name is changed
**Issue type:** Bug
**Component:** Date Input
**Location:** Macro - template.njk
**Version:** 1.0.0
**Prototype Kit Version:** 7.0.0-beta.9

**Details:**
When using the nunjucks macro to produce a date input component, changing the _name_ of the _year_ object inside _items_ with break the styling on the year input field that makes it 4 characters in width rather than the default 2 characters. This happens because the value of item.name for each input field provided via the macro is recycled to also set the CSS class on the inputs container div. This is a fixed class name that, if changed, will result in a loss of styles for that class.

**Reproduction:**
Use the default macro code from the docs to create a date input component. Change item.name for the year field from "year" to "Year" or any other string. 

**Code Sample:**
``` html
<div class="govuk-date-input__item govuk-date-input__item--year">
        <label class="govuk-label govuk-date-input__label" for="illness-end-year">
          Year
        </label>
        <input class="govuk-input govuk-date-input__input" id="illness-end-year" name="illnessEnd-year" type="number" pattern="[0-9]*">
      </div>
    </div>
```

alex-ju:
This PR:

Extends the input width classes to cover 2 to 5 characters so it can be reused for the date-input component:
   - defines a relationship between the number of characters and maximum ex width of an input
   - keeps the extra space for safari pre-fill icon on large inputs (10, 20, 30)

Updates the date-input component to:
   - decouple item name from style (fixes #836)
   - decouple the component name from the item name and item id (covers the work proposed by @lhokktyn in #826)
   - set the item width at input level in order to avoid label overlapping (see screenshot below)

Adds default date-input items if no items are being specified (fixes #468)

input-width
date-input-examples

### When users increase the font-size on labels
Before
screen shot 2018-07-04 at 12 14 55

After
screen shot 2018-07-04 at 12 15 17

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,15 @@
   `headingLevel: <number>` parameter. Default is `2`.
   ([PR #853](https://github.com/alphagov/govuk-frontend/pull/853))
 
+- Update date input component  
+
+  Allow the `name` and `id` attributes to be passed individually for each input item.
+
+  Rely on `classes` instead of item names to set the width of input items.
+
+  Add default (day, month, year) date input items if no items are being specified
+
+  ([PR #857](https://github.com/alphagov/govuk-frontend/pull/857))
 
 🔧 Fixes:
 
--- a/src/components/date-input/README.md
+++ b/src/components/date-input/README.md
@@ -29,33 +29,33 @@ Find out when to use the date input component in your service in the [GOV.UK Des
 
       <div class="govuk-date-input" id="dob">
 
-        <div class="govuk-date-input__item govuk-date-input__item--day">
+        <div class="govuk-date-input__item">
           <div class="govuk-form-group">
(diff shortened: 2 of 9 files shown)


igloosi:
awesome work! 🎆 

I wonder if input changes and date input changes could be separate PR's (dependency)?

alex-ju:
@igloosi: I didn't want to introduce dependencies between the PRs, but if that makes more sense I can split them up

igloosi:
true, if we we're going breaking at some point should be fine, as this PR has both new features and breaking changes

PR #858, opened 2018-07-02
Calculate font-sizes in rem

36degrees:
This PR supercedes https://github.com/alphagov/govuk-frontend/pull/839

Optionally enable automatically calculating font-sizes in rem by adding new settings `$govuk-typography-use-rem` and `$govuk-root-font-size`, which should match the font-size on the html (root) element.

In order to avoid this being a breaking change for users who use a different font-size on their root element (for example anyone using alphagov/govuk_template) this feature is going to be disabled by default for now. We'll [enable it for everyone by default](https://github.com/alphagov/govuk-frontend/issues/868) in the next major release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,21 @@
 
   ([PR #857](https://github.com/alphagov/govuk-frontend/pull/857))
 
+- The typography scale can now be converted from pixels to rem automatically,
+  with pixels also being provided as a fallback for older browsers.
+
+  This feature is disabled by default - in order to use it you will need to set
+  `$govuk-typography-use-rem: true` and ensure that `$govuk-root-font-size` is
+  set to the effective size of your root (html) element. For new projects, this
+  should be the default of 16px so you don't have to do anything. For projects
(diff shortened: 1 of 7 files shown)


igloosi:
excellent stuff.

hannalaakso:
Looks good to me 🙂 Just left a couple of small comments.

I enabled `$govuk-typography-use-rem` and tested whether the font size adjusts in the following browsers:

✅ Chrome 67.0 (set font size in Settings)
✅ Firefox 61 (set default font in Preferences)
✅ Safari 11.1 Mac OS X (set minimum font size in Advanced > Accessibility)
✅ Samsung Internet on Android Galaxy S8 (set font size in Settings)
✅ Chrome on Android Galaxy S8 (set font size in Settings)

These don't work but this is a limitation with iOS 😞 not our code:
🚫 Chrome iPhone 5s (set Larger text in General > Accessibility)
🚫 Safari iPhone 5s (set Larger text in General > Accessibility)

36degrees:
@hannalaakso are you happy with this now?

PR #859, opened 2018-07-02
Add govuk-visually-hidden-focusable class

alex-ju:
This PR introduces a new class `.govuk-visually-hidden-focusable` and deprecates `.govuk-visually-hidden-focussable`

[screen shot 2018-07-04 at 11 01 13](https://govuk-frontend-review-pr-859.herokuapp.com/docs/#undefined-css-.govuk-visually-hidden-focussable)

### DIFF
--- a/src/utilities/_visually-hidden.scss
+++ b/src/utilities/_visually-hidden.scss
@@ -3,7 +3,13 @@
     @include govuk-visually-hidden;
   }
 
+  /// @deprecated Deprecated as of release 1.1, replaced by `.govuk-visually-hidden-focusable`
   .govuk-visually-hidden-focussable {
     @include govuk-visually-hidden-focusable;
   }
+
+  .govuk-visually-hidden-focusable {
+    @include govuk-visually-hidden-focusable;
+  }
+
 }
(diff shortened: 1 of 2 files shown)


src/utilities/_visually-hidden.scss
```diff
@@ -3,7 +3,13 @@
     @include govuk-visually-hidden;
   }
 
+  /// @deprecated Deprecated as of release 1.1, replaced by `.govuk-visually-hidden-focusable`
   .govuk-visually-hidden-focussable {
     @include govuk-visually-hidden-focusable;
   }
+
```
nickcolley: Can we do

.className,
.secondClassName {
  @include ...
}

Or does that run into problems?

alex-ju: Tried, but Sass docs struggles with that when @ ing the deprecation

alex-ju: Also I think by generally defining the deprecated code separately make it easier to remove on major releases. We're doing the same with the font baseline adjustments.

nickcolley: Sounds good

Issue #862, opened 2018-07-02
<Main> element is missing padding in IE11


igloosi:
It's because IE11 doesn't set display block on `main` element
<img src="https://user-images.githubusercontent.com/3758555/42183378-80a5a39e-7e39-11e8-8a2f-f367dfbea21f.jpg">

PR #863, opened 2018-07-02
Set display:block on `main` wrapper

igloosi:
Because the main element is not recognised  – meaning it's not defined in IE's default style sheet,
so it uses CSS initial value, which is inline.

Fixes: #862

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,12 @@
 
 🔧 Fixes:
 
+- Apply `display:block` to `.govuk-main-wrapper`
+
+  In IE11 `main` element is set to `display:inline` so padding
+  and margins aren't applied.
+  ([PR #863](https://github.com/alphagov/govuk-frontend/pull/863)))
+
 - Line-heights are now converted from pixels to relative 'unit-less' values
   in order to prevent issues when resizing text in the browser.
   ([PR #837](https://github.com/alphagov/govuk-frontend/pull/837) and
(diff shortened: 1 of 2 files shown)


src/objects/_main-wrapper.scss
```diff
@@ -19,6 +19,8 @@
 @mixin govuk-main-wrapper {
   @include govuk-responsive-padding(6, "top");
   @include govuk-responsive-padding(6, "bottom");
+  // IE11 supports `main` but doesn't set it to block without HTML5shiv
```
36degrees: As far as I know html5shiv isn't involved here - IE11 does support the `<main>` element 'natively', it's just not styled correctly. The shiv is only invoked for Internet Explorer 6-9 according to their readme.

CHANGELOG.md
```diff
@@ -23,6 +23,11 @@
 
 🔧 Fixes:
 
+- Apply `display:block` to `.govuk-main-wrapper`
```
36degrees: If you're intending to add a line-break here, you need to leave a blank line. At the minute it wraps onto one line and reads like this:

> Apply `display:block` to `.govuk-main-wrapper` In IE11 `main` element is set to `display:inline` so padding an margins aren't applied.

Also, should be *and* instead of *an*.

igloosi: updated

nickcolley:
I would have thought this would have been fixed by the HTML5 shiv?

PR #865, opened 2018-07-04
Update dependencies

36degrees:
- Updated dependencies to the latest versions:

```
 autoprefixer         ^7.1.1  →   ^8.6.4
 gulp-debug           ^3.1.0  →   ^4.0.0
 gulp-eol             ^0.1.2  →   ^0.2.0
 gulp-nunjucks        ^3.0.0  →   ^4.0.0
 gulp-sass            ^3.1.0  →   ^4.0.1
 jest                ^22.4.4  →  ^23.2.0
 jest-axe             ^2.1.1  →   ^3.0.0
 standard            ^10.0.2  →  ^11.0.1
 yargs                ^8.0.2  →  ^12.0.1
 express             ^4.16.1  →  ^4.16.3
 gulp-better-rollup   ^3.1.0  →   ^3.3.0
 gulp-rename          ^1.2.2  →   ^1.3.0
 gulp-sass-lint       ^1.3.4  →   ^1.4.0
 gulp-task-listing    ^1.0.1  →   ^1.1.0
 js-yaml             ^3.10.0  →  ^3.12.0
 postcss-scss         ^1.0.0  →   ^1.0.6
 puppeteer            ^1.1.1  →   ^1.5.0
```

- Fixed linting issues raised by updated version of standard.
- Configured axe to ignore new skip-link rule which is not relevant when testing fragments of HTML


package.json
```diff
@@ -18,48 +18,48 @@
     "test:build:dist": "jest tasks/gulp/__tests__/after-build-dist.test.js"
   },
   "devDependencies": {
-    "autoprefixer": "^7.1.1",
+    "autoprefixer": "^8.6.4",
```
nickcolley: I don't expect anything weird, but with a major bump to autoprefixer should we check it's still outputting what we expect?

36degrees: Good call – checking.

igloosi: there are not output changes in package/ files :) 
from warning text: user-select:none addition

package.json
```diff
@@ -23,7 +23,7 @@
     "cssnano": "^3.10.0",
     "express": "^4.16.3",
     "gulp": "^3.9.1",
-    "gulp-better-rollup": "^3.3.0",
+    "gulp-better-rollup": "3.1.0",
```
nickcolley: Can we write a test that checks that this is always 3.1.0? Worried this is really easy to make a mistake on.

PR #869, opened 2018-07-04
Simplify readme generation

36degrees:
This PR is an attempt to start decoupling readme generation from the review app by using entirely different views for the component preview in the app and the README, and by refactoring the showExamples macro into two separate macros.

This also allows us to rename index.njk to README.njk to make the way that readmes are generated clearer.

There are some small changes to generated READMEs and to the review apps - trying to avoid using the word 'default' as I think this implies that the examples shows what would happen if you called the macro without passing any arguments which is not the case; and trying to make it clearer that clicking the preview link will take the user to the Frontend review app.

Other than that, changing the README content is not intended to be within the scope of this PR - but this will put in a good place to iterate the READMEs in the future to ensure that we're providing the right information to the user.


alex-ju:
Since we do CHANGELOG for internals I'll have to ask for an entry for this.

36degrees:
[Not any more](https://github.com/alphagov/govuk-frontend/pull/834)!

alex-ju:
Good to go then!

PR #875, opened 2018-07-05
[Experiment - DO NOT MERGE ]Macro in radios

igloosi:
this brings all sorts of issues of where the error message sits etc, but it works with any component
structure:
```
     component 
         name: details
         params: same params as the detail component takes
```

https://govuk-frontend-review-pr-875.herokuapp.com/components/radios/with-details-component/preview

### DIFF
--- a/src/components/radios/README.md
+++ b/src/components/radios/README.md
@@ -219,9 +219,7 @@ Find out when to use the radios component in your service in the [GOV.UK Design 
 
 #### Markup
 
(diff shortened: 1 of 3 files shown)


joelanman:
I was wondering if a 'block' approach might work, similar to the template. So we could have a `beforeItemsHtml` param, and it would insert whatever html there.

igloosi:
closing the experiment

Issue #879, opened 2018-07-06
Allow passing classes to govukRadio label


timja:
Currently there is `'govuk-radios__label'` applied but you can't pass your own class in, it would be useful to allow for a bold label and a nonbold hint underneath
https://github.com/alphagov/govuk-frontend/blob/master/src/components/radios/template.njk#L58

<img src="https://user-images.githubusercontent.com/3758555/42398697-c1950ee0-8162-11e8-91a3-21d559e17b82.png">

igloosi:
released in v1.2.0

PR #880, opened 2018-07-06
Allow additional label classes in radios and checkboxes macros

igloosi:
At the moment the label classes on radio and checkbox items are fixed. [There has been interest to be able to modify them](https://github.com/alphagov/govuk-frontend/issues/879), to say make the label bold.

When combined with ability to pass through then hint, the macros will be much more customisable

This PR:
- Updates the component template
- Adds a test to check for the additional label class
- Updates READMEs

Screenshot

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,12 @@
   and checkbox item to display the hint
   ([PR #846](https://github.com/alphagov/govuk-frontend/pull/846))
 
+- Allow additional classes to be added to the radio and checkbox items
+
+  You can now provide `label: { classes: 'extra-class' }` to each item.
+
+  ([PR #880](https://github.com/alphagov/govuk-frontend/pull/880))
+
 🔧 Fixes:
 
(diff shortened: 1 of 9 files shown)


joelanman:
I think we might need a quick discussion about whether bold radio labels is something with a definite need. For example you could differentiate with normal labels, and grey hint text.

igloosi:
Prior art:
https://www.gov.uk/log-in-file-self-assessment-tax-return/sign-in/prove-identity

https://govuk-frontend-review.herokuapp.com/components/radios/with-a-medium-legend/preview

joelanman:
the second example is interesting because it puts the hint inside the label. I think that just highlights that we should probably discuss what the best way to use radio labels is

igloosi:
sure, it puts in there as it was (is) the only way to do it

nickcolley:
We used the bold to contrast with the hints better I think.

igloosi:
we have decided to go ahead with this. code review needed

Issue #881, opened 2018-07-10
govuk-checkboxes__label and govuk-radios__label should 0 out margin on govuk-label


dashouse:
With the labels, legends and headings work `margin-bottom: 5px` was added to `govuk-label`.

This was intended for labels above text inputs but has been applied to checkbox and radios too.

So `govuk-checkboxes__label` and `govuk-radios__label` should have `margin-bottom: 0; ` applied to override this.

**Current**
<img width="619" alt="screen shot 2018-07-10 at 15 16 23" src="https://user-images.githubusercontent.com/23356842/42515945-dfab3e32-8454-11e8-9384-5a4935a6822a.png">

**Desired**
<img width="622" alt="screen shot 2018-07-10 at 15 16 42" src="https://user-images.githubusercontent.com/23356842/42515957-e464448c-8454-11e8-92b9-0fcf3e4f7963.png">

PR #883, opened 2018-07-10
Remove bottom margin on checkbox and radio labels 

igloosi:
With the labels, legends and headings work margin-bottom: 5px was added to `govuk-label`.
This was intended for labels above text inputs but has been applied to checkbox and radios  labels too.

So `govuk-checkboxes__label` and `govuk-radios__label` need to have `margin-bottom: 0;` applied to override this.

https://trello.com/c/aplfBBBF/1197-checkbox-and-radio-labels-should-not-have-a-bottom-margin

Fixes: #881

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,10 @@
 
 🔧 Fixes:
 
+- Remove mistakenly applied 5px bottom margin from radio and checkbox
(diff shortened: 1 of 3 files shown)


dashouse:
Looks good to me 👍

Issue #787, opened 2018-06-13
Conditional reveals result in a flash of unstyled content as the page loads


charge-valtech:
![fouc-conditional](https://user-images.githubusercontent.com/1433357/41355412-62cb582a-6f19-11e8-8a87-ac2b2bb63261.gif)

As the page loads, the hidden HTML shows briefly before being hidden. This is especially noticeable on slow connections. 

Ideally the hidden area should be hidden by default, and show only once triggered, or if no JS.

nickcolley:
I believe in the development of this feature we managed to avoid depending on the global 'js-enabled' class. This was because we were testing on a page with only the conditional reveal and on a fast connection.

In other components we've had to accept this dependency on a global 'js-enable' class so I think maybe all our components should just follow suit and avoid any issues like this.

dankmitchell:
@nickcolley now that you have fixed this for tabs should we re-open this issue?

nickcolley:
Hey @dankmitchell !

We have added two changes to GOV.UK Frontend to fix conditional reveals and then tabs for this similar issue.

The other issue for tabs is here: https://github.com/alphagov/govuk-frontend/issues/927

So I think both of these are resolved now, is that correct?

dankmitchell:
Ah yes, sorry @nickcolley. I was missing the `govuk-radios__conditional--hidden` class

Issue #815, opened 2018-06-21
Consider adding the `js-hidden` utility classes in the old GOV.UK Template


nickcolley:
In GOV.UK Template, there was a way to hide things conditionally if JavaScript is enabled:

https://github.com/alphagov/govuk_template/blob/78cbc105b564d2f4a83bb257abb98322bb851a93/source/assets/stylesheets/_accessibility.scss#L17

We might want to consider bringing these back.

nickcolley:
We've decided to avoid a global utility at the moment but potentially when we have more components this'd make more sense.

https://github.com/alphagov/govuk-frontend/issues/901#issuecomment-405872948

cc @adamsilver

PR #885, opened 2018-07-11
Fix flash of unstyled content with conditional reveals

nickcolley:
If the conditional reveal JavaScript is slow to execute it can result in showing the user their contents briefly which can be jarring.

See the original issue for an example: https://github.com/alphagov/govuk-frontend/issues/787

Fixes https://github.com/alphagov/govuk-frontend/issues/787

Tested in:
- IE 8 ✅
- IE 9 ✅
- IE 10 ✅
- IE 11 ✅
- Edge 17 ✅
- Firefox 61 ✅
- Chrome 67 ✅
- Safari 11.1 ✅

Assistive technologies:
I've tested this in VoiceOver and since `display: none` has the same behaviour as `aria-hidden=true` (styled with `display: none`), I've decided not to do a full test of assistive technologies.

Pages to review:
- https://govuk-frontend-review-pr-885.herokuapp.com/components/radios/with-conditional-item-checked/preview
- https://govuk-frontend-review-pr-885.herokuapp.com/components/checkboxes/with-conditional-item-checked/preview

https://trello.com/c/o9krC0SR/1195-conditional-reveals-result-in-a-flash-of-unstyled-content-as-the-page-loads

### DIFF
--- a/src/components/checkboxes/_checkboxes.scss
+++ b/src/components/checkboxes/_checkboxes.scss
@@ -140,7 +140,7 @@
     padding-left: $conditional-padding-left;
     border-left: $conditional-border-width solid $govuk-border-colour;
 
-    &[aria-hidden="true"] {
+    .js-enabled &--hidden {
       display: none;
     }
 
--- a/src/components/checkboxes/checkboxes.js
+++ b/src/components/checkboxes/checkboxes.js
@@ -40,7 +40,7 @@ Checkboxes.prototype.setAttributes = function ($input) {
   $input.setAttribute('aria-expanded', inputIsChecked)
 
   var $content = document.querySelector('#' + $input.getAttribute('aria-controls'))
-  $content.setAttribute('aria-hidden', !inputIsChecked)
+  $content.classList.toggle('govuk-checkboxes__conditional--hidden', !inputIsChecked)
 }
 
 Checkboxes.prototype.handleClick = function (event) {
--- a/src/components/checkboxes/checkboxes.js
+++ b/src/components/checkboxes/checkboxes.js
@@ -1,5 +1,6 @@
 import '../../vendor/polyfills/Function/prototype/bind'
 import '../../vendor/polyfills/Event' // addEventListener and event.target normaliziation
+import '../../vendor/polyfills/Element/prototype/classList'
 import { nodeListForEach } from '../../common'
 
 function Checkboxes ($module) {
--- a/src/components/checkboxes/template.test.js
+++ b/src/components/checkboxes/template.test.js
@@ -255,31 +255,82 @@ describe('Checkboxes', () => {
     })
   })
 
-  it('render conditional', () => {
-    const $ = render('checkboxes', {
-      name: 'example-conditional',
-      items: [
-        {
-          value: 'yes',
-          text: 'Yes'
-        },
-        {
-          value: 'no',
-          text: 'No',
-          checked: true,
-          conditional: {
-            html: 'Conditional content'
+  describe('render conditionals', () => {
+    it('hidden by default when not checked', () => {
+      const $ = render('checkboxes', {
+        name: 'example-conditional',
+        items: [
+          {
+            value: 'yes',
+            text: 'Yes',
+            conditional: {
+              html: 'Conditional content that should be hidden'
+            }
+          },
+          {
+            value: 'no',
+            text: 'No'
           }
-        }
-      ]
-    })
+        ]
+      })
 
-    const $component = $('.govuk-checkboxes')
-    const $lastInput = $component.find('.govuk-checkboxes__input').last()
-    expect($lastInput.attr('data-aria-controls')).toBe('conditional-example-conditional-2')
-    const $lastConditional = $component.find('.govuk-checkboxes__conditional').last()
-    expect($lastConditional.attr('id')).toBe('conditional-example-conditional-2')
-    expect($lastConditional.html()).toContain('Conditional content')
+      const $component = $('.govuk-checkboxes')
+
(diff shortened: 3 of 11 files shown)


src/components/radios/template.njk
```diff
@@ -61,7 +61,7 @@
       }) | indent(6) | trim }}
     </div>
     {% if item.conditional %}
-      <div class="govuk-radios__conditional" id="{{ conditionalId }}">
+      <div class="govuk-radios__conditional{% if not item.checked %} is-hidden{% endif %}" id="{{ conditionalId }}">
```
nickcolley: For users that do not use macros, how do we document this?

src/components/radios/_radios.scss
```diff
@@ -152,7 +152,7 @@
     padding-left: $conditional-padding-left;
     border-left: $conditional-border-width solid $govuk-border-colour;
 
-    &[aria-hidden="true"] {
+    .js-enabled &.is-hidden {
```
nickcolley: Should we use the 'is-hidden' class AND aria-hidden incase CSS fails to load but JS doesnt?

src/components/checkboxes/_checkboxes.scss
```diff
@@ -139,7 +139,7 @@
     padding-left: $conditional-padding-left;
     border-left: $conditional-border-width solid $govuk-border-colour;
 
-    &[aria-hidden="true"] {
+    .js-enabled &.is-hidden {
```
nickcolley: Should this class be called 'js-hidden'?

Could this be a common utility class?

https://github.com/alphagov/govuk-frontend/issues/815

hannalaakso: I think a common utility class would make sense.

(On a separate note, I think we should probably prefix our JS classes. But let's discuss that another time, I don't want to derail this PR 😁 )

hannalaakso: Yes I was thinking of `govuk-js-hidden`. I was even thinking of `js-hidden-govuk` as it's handy to recognise JS classes at a glance. But it just sounds too weird... 😅  

But `js-hidden` just seems like it would very easily conflict with something else.

nickcolley: Makes sense, looks like we already use js-hidden, so maybe raise an issue for us to think about it after all 😬 

https://github.com/alphagov/govuk-frontend/search?q=js-hidden&unscoped_q=js-hidden

src/components/radios/template.njk
hannalaakso: Actually I realised that this still wouldn't work because we need to we do `.js-enabled .js-hidden` in CSS so that style doesn't kick in until JS has loaded 😞 I'll keep on thinking.

nickcolley: That's currently how the JavaScript logic works. The logic in the template is for the initial render.

I think this might not be a big issue since the example we use in the Design System website has a pre-checked input which means you'd see it in the generated HTML.

src/components/checkboxes/_checkboxes.scss
nickcolley: I'm going to change the 'is-hidden' to 'js-hidden' to be consistent with tabs.

src/components/tabs/_tabs.scss
```diff
@@ -110,12 +110,12 @@
...
+
+        &.js-hidden {
+          display: none;
+        }
       }
     }
 
   }
-
-  .js-hidden {
```
nickcolley: good news is the tests found this selector being defined globally

i imagine I should fix this behaviour in a different PR first...

nickcolley: Makes me think having this class used with slightly different behaviours is indeed dangerous.

So we should either

1. Continue doing js-hidden for specific components but namespacing with their component names: `govuk-tabs-js-hidden`
2. Pull `js-hidden` out into a utility class that is used by tabs and conditional reveals

I feel 2. is probably the best way, since others can use this class too.

nickcolley: Pushed up some scrappy working to see what that might look like, will do something clean next week.

src/components/checkboxes/checkboxes.js
```diff
@@ -40,7 +41,7 @@ Checkboxes.prototype.setAttributes = function ($input) {
   $input.setAttribute('aria-expanded', inputIsChecked)
 
   var $content = document.querySelector('#' + $input.getAttribute('aria-controls'))
-  $content.setAttribute('aria-hidden', !inputIsChecked)
+  $content.classList.toggle('govuk-checkboxes__conditional--hidden', !inputIsChecked)
```
nickcolley: In the case that CSS fails but JS succeeds, should we also include 'aria-hidden'? Or is this a bit too over the top?

36degrees:
@nickcolley is there anything blocking this from being merged?

nickcolley:
@36degrees just wanted to get more eyes on it since it's a big change

src/components/checkboxes/template.test.js
```diff
@@ -255,31 +255,82 @@ describe('Checkboxes', () => {
...
           }
-        }
-      ]
+        ]
+      })
+
+      const $component = $('.govuk-checkboxes')
+
+      const $firstConditional = $component.find('.govuk-checkboxes__conditional').first()
+      expect($firstConditional.html()).toContain('Conditional content that should be hidden')
```
36degrees: It'd be good to split this out into its own test and just have the one assertion per test.

src/components/checkboxes/checkboxes.js
36degrees: I reckon this is fine as is.

PR #886, opened 2018-07-12
Update tabs spacing

dashouse:
Updates internal padding of tab content in the Tabs component (backlog https://github.com/alphagov/govuk-design-system-backlog/issues/100)

**Before** 
screen shot 2018-07-12 at 09 57 58

**After**
screen shot 2018-07-12 at 14 24 56

This required margin to be defined on the tab container at mobile breakpoint to separate content.
I also increased the space below the contents menu.

screen shot 2018-07-12 at 09 54 06

Have also used the `:last-child` solution from conditional radios and checkboxes to remove the margin-bottom of the last item that is inside the container to avoid double margin.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,8 +23,10 @@
 🔧 Fixes:
 
 - Update details behaviour to remove margin-bottom for all elements
-
   ([PR #900](https://github.com/alphagov/govuk-frontend/pull/900))
+
+- Update internal padding of tab content in the tabs component
+  ([PR #886](https://github.com/alphagov/govuk-frontend/pull/886))
 
 ## 1.1.0 (feature release)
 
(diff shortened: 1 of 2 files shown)


src/components/tabs/tabs.yaml
```diff
@@ -8,7 +8,7 @@ examples:
         panel:
           html: |
             <h2 class="govuk-heading-l">Past day</h2>
-            <table class="govuk-table">
+            <table class="govuk-table govuk-!-margin-bottom-0">
```
nickcolley: This works, just a question: In the conditional reveal I think we do this to any `:last-child`, should we consider doing that here too?

adamsilver:
The padding left and right was as it was so that the text in the first tab vertically aligned with the text (or whatever) inside the panel. This was something @joelanman originally asked me to do. I don't mind the change, but mentioning just in case.

dashouse:
@adamsilver @joelanman Ok great, sorry for the slight alignment issue (no pun intended). I'll keep the top/bottom margin and check in with Joe on the left and right...makes sense

nickcolley:
Thanks @dashouse 👍

PR #890, opened 2018-07-12
Remove example files govuk frontend

nickcolley:
This pull request removes the requirement for components to have a hardcoded example, and instead makes use of the first example as defined in the components 'yaml' config.

While these files are published to npm, as far as I know these are only used for internal documentation on the 'all components' page of the review applications.

It's also unlikely a user would use this functionality, since these files have hardcoded examples.

For this reason I'm determining this as not a breaking change, but a documentation related change.

https://trello.com/c/qM6PLrZj/264-spike-decide-whether-to-generate-example-html-files-from-the-examples-in-the-yaml-file-one-hour


36degrees:
I _think_ these three commits are largely unrelated – could they be split up into separate PRs?

nickcolley:
If the context of the pull request is to remove the examples, then If I do that without updating the 'all components' page our tests will fail, so they can't be split up.

If the context of the pull request is to use the yaml rather than hardcoding, we'd have lots of unused files which normally we would clean up in the same file if possible?

Happy to split up if you think that makes it easier to understand.

app/app.js
```diff
@@ -125,6 +116,27 @@ module.exports = (options) => {
     res.render('component-preview', { bodyClasses, previewLayout })
   })
 
+  // All components view
+  app.get('/examples/all-components', function (req, res, next) {
+    const components = fileHelper.allComponents
+
+    res.locals.componentData = components.map(componentName => {
+      let componentData = fileHelper.getComponentData(componentName)
+      let firstExample = componentData.examples[0]
```
36degrees: This is assuming that the _first_ example will be the 'default' example – to be consistent with what we do elsewhere can we re-use the logic from [here](https://github.com/alphagov/govuk-frontend/blob/61db9b3cec7249c72337cfd12f3c045876fa10e2/app/app.js#L90-L96) to pull out whichever example has the name 'default'?

nickcolley:
I like your suggestion about having an 'all' rather than in examples, will make that change.

app/app.js
nickcolley: I suppose I was not too concerned that this logic does get the default (even though it does by coincidence).

Let's go with your suggestion and have the same intent as elsewhere.

nickcolley:
Need to change the tests since they pointed at the old examples route.

app/__tests__/app.test.js
```diff
@@ -90,7 +90,10 @@ describe('frontend app', () => {
       request.get(requestParamsHomepage, (err, res) => {
         let $ = cheerio.load(res.body)
         let componentsList = $('li a[href^="/components/"]').get()
-        expect(componentsList.length).toEqual(lib.allComponents.length)
+        // Since we have an 'all' component link that renders the default example of all
+        // components, there will always be one more expected link.
```
36degrees: 👍

36degrees:
Good stuff.

Issue #899, opened 2018-07-13
Details should remove margin-bottom on the last child


nickcolley:
<img width="456" alt="screen shot 2018-07-13 at 16 38 12" src="https://user-images.githubusercontent.com/2445413/42700297-3a91512a-86bb-11e8-8333-a8073ba9650d.png">

```html
<details class="govuk-details" role="group" open="">
  <summary class="govuk-details__summary" role="button" aria-controls="details-content-0" aria-expanded="true">
    <span class="govuk-details__summary-text">
      I don't know the local council
    </span>
  </summary>
  <div class="govuk-details__text" id="details-content-0" aria-hidden="false">
    <div class="govuk-form-group">
      <label class="govuk-label" for="postcode">
        Enter your postcode
        <span class="govuk-hint">For example, SW1A 2AA</span>
      </label>
      <input class="govuk-input govuk-input--width-10" id="postcode" name="postcode" type="text">
    </div>                 
  </div>
</details>
```

```css
.govuk-details__text > :last-child {
    margin-bottom: 0;
}
```

<img width="435" alt="screen shot 2018-07-13 at 16 40 25" src="https://user-images.githubusercontent.com/2445413/42700377-7a4f9362-86bb-11e8-9193-a4e7290236a8.png">

PR #900, opened 2018-07-13
Update details behaviour to remove margin-bottom for all elements

nickcolley:
If we have a form-group it will not be caught by the logic that only targets paragraphs.

This is consistent with the tabs and conditional reveal logic.

Fixes https://github.com/alphagov/govuk-frontend/issues/899

### DIFF
--- a/src/components/details/_details.scss
+++ b/src/components/details/_details.scss
@@ -80,7 +80,7 @@
     margin-bottom: govuk-spacing(4);
   }
 
-  .govuk-details__text p:last-child {
+  .govuk-details__text > :last-child {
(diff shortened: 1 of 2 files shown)


src/components/details/_details.scss
```diff
@@ -75,12 +75,12 @@
     border-left: $govuk-border-width solid $govuk-border-colour;
   }
 
-  .govuk-details__text p {
+  .govuk-details__text > p {
```
nickcolley: Might be a breaking change actually 🤔

dashouse:
👍

PR #902, opened 2018-07-13
[Do not merge] Gitignore dist package

nickcolley:
This PR explores if we could add package/ and dist/ to gitignore.

If we were to do this it'd mean our editors would not index these folders, including our contributors.

I'm not sure how it would impact the Github interface, potentially it would remove them from the search which could be  aproblem for dist/

There are addition commits that show editing an existing file, and adding a new file.

Using `--force` with git add ensures that new files that are untracked are added regardless of gitignore

This feels maybe too clever, but thought I'd share before throwing it away.


nickcolley:
Going to close this but will bring up in a future 'developer discussion time' with the team.

PR #916, opened 2018-07-17
Replace global `js-hidden` class with tabs panel modifier

nickcolley:
`js-hidden` is a global class that is defined by GOV.UK Template (https://github.com/alphagov/govuk_template/blob/78cbc105b564d2f4a83bb257abb98322bb851a93/source/assets/stylesheets/_accessibility.scss#L17)

This could mean that if users include GOV.UK Frontend into their project it could break the JavaScript fallback
in their project, since the tabs' implementation of `js-hidden` does not check for the presense of the
`js-enabled` class.

To avoid this problem we avoid a global class name and instead create a modifier for the panel component which is being hidden.

Thankfully, since this class is only added at runtime by JavaScript, doing this is not a breaking change.


nickcolley:
Reviewers can you confirm that this is not a breaking change please.

CHANGELOG.md
```diff
@@ -22,11 +22,8 @@
 
 🔧 Fixes:
 
-- Pull Request Title goes here
-
-  Description goes here (optional)
-
-  ([PR #N](https://github.com/alphagov/govuk-frontend/pull/N))
+- Replace global `js-hidden` class with tabs panel modifier
```
36degrees: Just to be super clear, what about something like:

> Replace conflicting `js-hidden` class used within the tabs component with a new modifier class. Because this class is defined and used within the JavaScript, no markup changes are required.

Thoughts?

PR #917, opened 2018-07-17
Delay the loading of GOV.UK Frontend to simulate slow connections.

nickcolley:
With this change you can easily reproduce the issue described here: https://github.com/alphagov/govuk-frontend/issues/787

# Before
before-fast-execution-js-tabs
before-fast-execution-js-reveal

# After
after-fast-execution-js-tabs
after-fast-execution-js-reveal

### DIFF
--- a/app/views/layouts/_generic.njk
+++ b/app/views/layouts/_generic.njk
@@ -28,7 +28,14 @@
 
 {% block bodyEnd %}
   <script src="/public/all.js"></script>
-  <script>window.GOVUKFrontend.initAll()</script>
+  <script>
+    setTimeout(function () {
+      // In order to simulate a slow connection, we intentionally delay the intialisation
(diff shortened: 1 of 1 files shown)


nickcolley:
Hmm, we need to make sure this doesnt break the tests since they rely on the review application.

Should we decouple the two...

36degrees:
We've got https://trello.com/c/dhmbvwJZ/1223-decouple-component-integration-tests-from-the-review-app in triage and I've recorded this work as https://trello.com/c/vO3ycOTL/1230-make-it-easy-to-detect-eg-flashes-of-unstyled-content-in-the-frontend-review-app

Going to close this in the meantime as it doesn't feel like something we can merge soon.

nickcolley:
@36degrees thanks

Issue #892, opened 2018-07-13
Clarify JavaScript installation instructions that GOV.UK Frontend should be initialised at before the '<body>'


nickcolley:
Currently it's easy to interrupt the instructions to put your script in the `<head>` of your page, this will block render and may result in issues.

We instead want uses to put them at the end of their document, we should update the documentation to make this clear:

https://github.com/alphagov/govuk-frontend/blob/master/docs/installation/installing-with-npm.md#option-1-include-javascript

PR #923, opened 2018-07-23
Clarify where Javascript should be included

igloosi:
Update documentation to specify that Javascript should be included before the closing `</body>` tag and not in the `<head>`

Fixes: https://github.com/alphagov/govuk-frontend/issues/892
Trello ticket: https://trello.com/c/yrYI7eWq/1220-2-clarify-javascript-installation-instructions-that-govuk-frontend-should-be-initialised-at-before-the-body

### DIFF
--- a/docs/installation/installing-with-npm.md
+++ b/docs/installation/installing-with-npm.md
@@ -127,9 +127,14 @@ Include the `node_modules/govuk-frontend/all.js` script on your page. You might 
 
 To initialise all components, use the `initAll` function.
 
+JavaScript in GOV.UK Frontend requires HTML to be parsed first by the browser before it is initialised. Because of this, make sure you include the script before the closing `</body>` tag.
+Including the script elsewhere will stop components from functioning or displaying correctly.
+
 ```html
-<script src="path-to-assets/govuk-frontend/all.js"></script>
-<script>window.GOVUKFrontend.initAll()</script>
+    <script src="path-to-assets/govuk-frontend/all.js"></script>
+    <script>window.GOVUKFrontend.initAll()</script>
+  </body>
+</html>
 ```
 
 #### Initialise individual included components


igloosi:
thanks @amyhupe . updated. let's have devs look at it as well

docs/installation/installing-with-npm.md
```diff
@@ -127,9 +127,13 @@ Include the `node_modules/govuk-frontend/all.js` script on your page. You might
 
 To initialise all components, use the `initAll` function.
 
+JavaScript in GOV.UK Frontend requires HTML to be first parsed by the browser, otherwise it will block render. This may stop components from functioning or displaying correctly. To avoid this, make sure you include the script before the closing `</body>` tag.
```
36degrees: As I understand it, the main reason it's important for users to include it at the end of their document is because we're not wrapping our init up in a `documentReady`, so it'll only affect elements that are already part of the DOM when the script is run?

Do we want to be clearer about this?

At the minute, it sounds like the most important reason is that it affects performance and that it _might_ result in your components not working, but actually we _know_ that it _will_ result in components not being initialised, and that it _might_ impact performance.

igloosi: reworded to make it more explicit

docs/installation/installing-with-npm.md
```diff
@@ -127,9 +127,14 @@ Include the `node_modules/govuk-frontend/all.js` script on your page. You might
 
 To initialise all components, use the `initAll` function.
 
+JavaScript in GOV.UK Frontend requires the DOM to be parsed first by the browser before it is initialised. Because of this, make sure you include the script before the closing `</body>` tag.
```
nickcolley: Not blocking:

I think the DOM is a parsed representation of the HTML document, so the browser parses the HTML into the DOM, rather than parsing the DOM directly?

36degrees:
Makes sense to me, though it'd be good to address Nick's comment if possible. @amyhupe are you happy with the revised version?

docs/installation/installing-with-npm.md
igloosi: i've amended the wording

PR #926, opened 2018-07-23
Only underline logo on hover (with custom colours)

36degrees:
Because all borders will become text colour when a user overrides colour in their browser, the transparent border will become always visible.

By only applying the border on hover we avoid this which means we get the correct behaviour even when colours are customised. The layout is not affected because we offset the additional border with a negative margin.

We do however have to add an additional rule for IE8 which does not support customColor and so ignores the rule entirely - which meant it had no border before. This would be fine, except the negative margin _does_ kicks in on hover, which means without a border the header would shrink by 1px on hover in IE8.

## Before

screen shot 2018-07-23 at 16 55 05-fullpage

## After

jul-23-2018 16-57-16

### DIFF
--- a/src/components/header/_header.scss
+++ b/src/components/header/_header.scss
@@ -101,14 +101,16 @@
 
     &:link,
     &:visited {
-      margin-bottom: -1px; // Negate transparent bottom border
-      border-bottom: 1px solid transparent;
       text-decoration: none;
     }
 
     &:hover,
     &:active {
-      border-bottom-color: currentColor;
+      // Negate the added border
+      margin-bottom: -1px;
+      // Omitting colour will use default value of currentColor – if we
+      // specified currentColor explicitly IE8 would ignore this rule.
+      border-bottom: 1px solid;
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/components/header/_header.scss
```diff
@@ -101,14 +101,16 @@
...
-      border-bottom: 1px solid transparent;
       text-decoration: none;
     }
 
     &:hover,
     &:active {
-      border-bottom-color: currentColor;
+      margin-bottom: -1px; // Negate transparent bottom border
+      border-bottom: 1px solid currentColor;
+      @include govuk-if-ie8 {
```
nickcolley: could you include a comment that explains why the ie8 code is needed? the info in the PR is very good

src/components/header/_header.scss
```diff
@@ -108,6 +108,8 @@
     &:active {
       margin-bottom: -1px; // Negate transparent bottom border
       border-bottom: 1px solid currentColor;
```
nickcolley: Out of curiosity, I think that border's default colour is `currentColor` which could mean changing this statement to `border-bottom: 1px solid` may remove the need for the conditional IE8 logic?

36degrees: Great suggestion – that seems to work well.

36degrees:
@nickcolley would you mind taking another look?

nickcolley:
👍

PR #931, opened 2018-07-30
Add versioning guidance

nickcolley:
I've tried to make the sections based on the practical work we do.

I hope we can improve this document in time with more concrete examples, but for now I've left them out.


docs/contributing/versioning.md
```diff
@@ -0,0 +1,121 @@
...
+Bad reasons to make breaking changes:
+
+- you think you can sneak a change in with other breaking changes - breaking changes need to be migrated from, which is effort for our users we should try to avoid
+- changing the name of an API, based on a hunch
+- adopting a technology based on popularity and not because of the problems they solve
+- changing a components interface without a good reason to do so
+
+This is similar to how we try to tackle most problems, by focusing on user needs first.
+
+Ember.JS is a great example of prioritising stability see [Stability without Stagnation: Lessons Learned from Shipping Ember.js
```
hannalaakso: Could we consider moving this to the bottom of the section and marking it as "Optional reading" or something like that? I've just read through it, it's really interesting but it's quite a long read and doesn't have tons of resemblance to the way we work. We wouldn't want a potential contributor to feel they have to read it in order to understand how their contribution fits in with our releases. But definitely happy to keep it in this doc as it touches on a lot of important aspects of stability 🙂

docs/contributing/versioning.md
```diff
@@ -0,0 +1,121 @@
...
+](https://speakerdeck.com/wycats/stability-without-stagnation-lessons-learned-from-shipping-ember-dot-js)
+
+We do not want our users to be reliant on certain versions of our library with no way forward.
+
+Whenever we decide to make a breaking change we must ensure we do our best to try to first deprecate the feature and allow a migration path.
```
hannalaakso: Could this link to the Migration section?

docs/contributing/versioning.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Proposal
+
+Some breaking changes that need discussion may first be proposed in the [GOV.UK Design System architecture repository](https://github.com/alphagov/govuk-design-system-architecture/blob/master/proposals/README.md).
+
+This is to ensure the community can get involved with the decision.
+
+You should make sure to take steps to actively involved the community and not wait for feedback, this might be in the form of presentations or meetings. Since these issues may be difficult to understood without a conversation.
+
```
hannalaakso: Could this paragraph be clarified a little bit, I'm not sure what it's trying to say.

docs/contributing/versioning.md
```diff
@@ -0,0 +1,121 @@
...
+Note: Our users will may not know what 'deprecation' means, so it's important to also clarify in full what this means.
+
+Example 1: Fixing a typo in a CSS class name.
+
+1. We discover the class name `.govuk-visually-hidden-focussable` includes the typo 'focussable'
+2. We raise a pull request that renames the class to `.govuk-visually-hidden-focusable` while keeping
+the previous class available.
+3. We add a comment to the source code that indicates this is deprecated, and raise an issue to remove it in a future breaking release.
+4. When releasing the change we include a clear summary that indications what was the problem, what we've changed and how a user can migrate before the future breaking release.
```
hannalaakso: Could this mention that the summary/indication should live in the CHANGELOG?

docs/contributing/versioning.md
```diff
@@ -0,0 +1,121 @@
+# Versioning
+
+With an user interface (UI) library there are many ways that people can consume our code.
+
+We follow [semantic versioning](https://semver.org/) but a UI library often has subjective changes such as visual spacing changes.
+
+This document aims to give a basis to make decisions as a team.
```
hannalaakso: It might help to clarify if we add something like: "on how to version our releases".

docs/contributing/versioning.md
nickcolley: Yeah good idea

docs/contributing/versioning.md
nickcolley: Great idea

docs/contributing/versioning.md
bahmady: If this is an external link and external people are reading this (potentially) then it should be clear they'd be leaving GOV.UK

docs/contributing/versioning.md
```diff
@@ -0,0 +1,121 @@
...
+- changing the name of an API, based on a hunch
+- adopting a technology based on popularity and not because of the problems they solve
+- changing a components interface without a good reason to do so
+
+This is similar to how we try to tackle most problems, by focusing on user needs first.
+
+Ember.JS is a great example of prioritising stability see [Stability without Stagnation: Lessons Learned from Shipping Ember.js
+](https://speakerdeck.com/wycats/stability-without-stagnation-lessons-learned-from-shipping-ember-dot-js)
+
+We do not want our users to be reliant on certain versions of our library with no way forward.
```
bahmady: I don't think this is clear enough, and I would question whether it's absolutely necessary.

docs/contributing/versioning.md
bahmady: that _indicates_ what the problem was

docs/contributing/versioning.md
```diff
@@ -0,0 +1,121 @@
...
+
+1. We discover the class name `.govuk-visually-hidden-focussable` includes the typo 'focussable'
+2. We raise a pull request that renames the class to `.govuk-visually-hidden-focusable` while keeping
+the previous class available.
+3. We add a comment to the source code that indicates this is deprecated, and raise an issue to remove it in a future breaking release.
+4. When releasing the change we include a clear summary that indications what was the problem, what we've changed and how a user can migrate before the future breaking release.
+
+Sometimes it is not possible to deprecate code, this is OK but try to make this a last resort.
+
+Migrating takes a lot of effort for our users, and if done badly can result in a loss of trust.
```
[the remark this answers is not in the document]
bahmady: It seems like some of the ordering is slightly strange, for example this seems to be outside of the 'Migration' header/section.

docs/contributing/versioning.md
```diff
@@ -0,0 +1,121 @@
...
+
+Do not use a breaking release of deprecated features as an opportunity to make breaking changes based on the ‘bad reasons to make breaking changes’ in the ‘Stability’ section.
+
+## Migration
+
+Migration is the practice of a user moving from one approach to an equivalent approach.
```
bahmady: I would be specific that this is how you're defining it here. Some definitions would be more specific than simply "approach" e.g. "OS migration".

docs/contributing/versioning.md
```diff
@@ -0,0 +1,121 @@
...
+
+It is very important that we make it easy to migrate when we make deprecations.
+
+If possible include this directly in the CHANGELOG entry that introduces the deprecation.
```
bahmady: Include what? The decision behind the deprecation? I'm not clear on this part.

docs/contributing/versioning.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Public API
+
+As defined in the [Semantic Versioning specification](https://semver.org/) a defined public api is required to version against.
```
bahmady: Do you need the word 'defined' twice here? Can you just omit the second one? Better yet could you change this to something like:

> Application of the [Semantic Versioning specification](https://semver.org/) requires a public API. 

Or even just

> The [Semantic Versioning specification](https://semver.org/) requires a public API.

docs/contributing/versioning.md
```diff
@@ -0,0 +1,121 @@
...
+
+> For this system to work, you first need to declare a public API. This may consist of documentation or be enforced by the code itself.
+
+### Design System Website
+One of ways users interact with GOV.UK Frontend is through the [GOV.UK Design System website](https://design-system.service.gov.uk/).
+
+This includes:
+
+- HTML - Documented in examples
```
bahmady: Can you link to some of these?

docs/contributing/versioning.md
```diff
@@ -0,0 +1,121 @@
...
+- [JavaScript](https://github.com/alphagov/govuk-frontend/blob/master/docs/installation/installing-with-npm.md#using-javascript)
+- SCSS - https://govuk-frontend-review.herokuapp.com/docs/
+- Nunjucks Macros (Templates)
+
+## Accidental breaking changes
+If a backward-incompatible change is released unintentionally, we will follow the process outlined on semver.org: https://semver.org/#what-do-i-do-if-i-accidentally-release-a-backwards-incompatible-change-as-a-minor-version
+
+> As soon as you realize that you’ve broken the Semantic Versioning spec, fix the problem and release a new minor version that corrects the problem and restores backwards compatibility. Even under this circumstance, it is unacceptable to modify versioned releases. If it’s appropriate, document the offending version and inform your users of the problem so that they are aware of the offending version.
+
+If it makes sense to, setup an incident review that allows the team to see if there are
```
bahmady: > If appropriate, you can set up an incident review that allows the team to see ...

docs/contributing/versioning.md
```diff
@@ -0,0 +1,121 @@
...
+
+Some breaking changes that need discussion may first be proposed in the [GOV.UK Design System architecture repository](https://github.com/alphagov/govuk-design-system-architecture/blob/master/proposals/README.md).
+
+This is to ensure the community can get involved with the decision.
+
+You should make sure to take steps to actively involved the community and not wait for feedback, this might be in the form of presentations or meetings. Since these issues may be difficult to understood without a conversation.
+
+## Deprecation
+
+Deprecation is the practice of making breaking changes gradually, and is part of communicating upcoming changes
```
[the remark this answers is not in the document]
nickcolley: What is your mental model of deprecations?

The intention with this is to set deprecations up as a tool to make breaking changes in a way that's best for our users.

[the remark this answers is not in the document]
bahmady: In my mind, generally speaking it's simply the idea of recommending against using a particular service/thing. You could perhaps consider qualifying the sentence by saying how you define it, i.e. "We define deprecation..."

nickcolley:
@hannalaakso @bahmady thanks both for the great reviews, I'll update soon 😄

docs/contributing/versioning.md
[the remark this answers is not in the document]
nickcolley: Thanks, will try and clarify this :)

nickcolley:
I've updated the document based on your feedback, let me know what you think.

hannalaakso:
This addresses my comments, thanks @nickcolley 👍 It would be good to get @igloosi or @36degrees to look over this too.

docs/contributing/versioning.md
```diff
@@ -0,0 +1,102 @@
...
+
+- issues that are barriers for end-users (users of services) based on evidence (for example user research)
+- issues that are barriers for users (users of GOV.UK Frontend) based on evidence
+- accessibility issues
+- security issues
+- performance issues
+
+We would not make breaking changes for:
+
+- you think you can sneak a change in with other breaking changes
```
igloosi: perhaps replace "you think you can" with "thinking you can" for sake of consistent use of gerunds in this list

docs/contributing/versioning.md
```diff
@@ -0,0 +1,102 @@
...
+- changing the name of an API, based on a hunch
+- adopting a technology based on popularity and not because of the problems they solve
+- changing a components interface without a good reason to do so
```
igloosi: should be a possessive s not plural s: component's

PR #934, opened 2018-07-30
Add test coverage for caller interface

nickcolley:
Updates the render method to call the components in the same way our documentation asks users to.

Then adds the ability to pass children to the render function, which will be passed to the component via the [call](https://mozilla.github.io/nunjucks/templating.html#call) Nunjucks feature.

As a side note our render function has a similar signature as a virtual dom function now, such as React.

https://trello.com/c/FaKTvCYz/1289-add-tests-for-call-usage-for-components-that-support-it

### DIFF
--- a/src/components/fieldset/template.test.js
+++ b/src/components/fieldset/template.test.js
@@ -7,14 +7,14 @@ const { render, getExamples } = require('../../../lib/jest-helpers')
 const examples = getExamples('fieldset')
 
 describe('fieldset', () => {
-  it('default example passes accessibility tests', async () => {
+  it('passes accessibility tests', async () => {
     const $ = render('fieldset', examples.default)
 
     const results = await axe($.html())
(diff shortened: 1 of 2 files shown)


src/components/fieldset/template.test.js
```diff
@@ -96,16 +92,41 @@ describe('fieldset', () => {
...
+        'data-attribute': 'value'
       }
     })
 
     const $component = $('.govuk-fieldset')
     expect($component.attr('data-attribute')).toEqual('value')
-    expect($component.attr('data-another-attribute')).toEqual('another-value')
+  })
+
+  it('renders nested components using `call`', () => {
```
36degrees: Minor, but can we move this above the tests for additional classes and additional attributes? Trying to get the 'core' component tests (the things you need to be able to do use the component) at the top, followed by the tests that allow you to 'extend' a component.

PR #936, opened 2018-08-02
Remove panel title margin without body text

colinrotherham:
This pull request patches the styling for `govukPanel()`.

Currently:

1. The element **govuk-panel__body** `{ text }` is optional
2. Margin is always applied to **govuk-panel__title** `{ titleText }`

The title only needs a margin if the optional text is provided, otherwise vertical alignment is affected:

incorrect margin

Once patched, the margin will be optional 👍

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,12 @@
 
   ([PR #N](https://github.com/alphagov/govuk-frontend/pull/N))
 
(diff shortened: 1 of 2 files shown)


igloosi:
Hi @colinrotherham 

this looks like a good tweak. We just need a second confirmation from the team

36degrees:
I'd like to make sure @dashouse has the chance to review this.

PR #938, opened 2018-08-03
Date Input to use inline-block

dashouse:
Update Date Input to use inline block not floats.

When we give support to people wanting to add an element next to an input we often tell them to use `govuk-!-display-inline-block`. It would be good if we followed the same method for components too.

This should also make it easier to insert optional separators in between the fields in a future update.

I've also removed the extra 2px padding on the label, no longer needed.

### DIFF
--- a/src/components/date-input/_date-input.scss
+++ b/src/components/date-input/_date-input.scss
@@ -10,18 +10,18 @@
 @include govuk-exports("govuk/component/date-input") {
   .govuk-date-input {
     @include govuk-clearfix;
+    // font-size: 0 removes whitespace caused by inline-block
+    font-size: 0;
   }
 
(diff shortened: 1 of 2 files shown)


src/components/date-input/_date-input.scss
```diff
@@ -10,18 +10,17 @@
 @include govuk-exports("govuk/component/date-input") {
   .govuk-date-input {
     @include govuk-clearfix;
+    font-size: 0;
```
36degrees: I think this could do with a comment explaining why it's needed / what it's doing.

igloosi:
resolved CHANGELOG conflict

igloosi:
merged on behalf of @dashouse

PR #940, opened 2018-08-03
Change spacing relationship on default and small legends and hints 

dashouse:
Before the release of GOV.UK Frontend we changed the spacing of labels and hints when using the default or `govuk-label--s` class. (https://github.com/alphagov/govuk-frontend/pull/806)

We chose not update the legend and hint, in isolation they looked good.

However when a service uses a small label + hint and a small legend + hint on the same page the inconsistency is noticeable.

Amending the spacing on the `<legend>` is slightly different to the `<label>`, there was also a need to only reduce spacing on the `<legend>` when a hint was applied, hence the use of an adjacent selector and negative margin.

**Before**
screen shot 2018-08-03 at 10 44 36

**After**
screen shot 2018-08-03 at 10 40 04

### DIFF
--- a/src/components/fieldset/_fieldset.scss
+++ b/src/components/fieldset/_fieldset.scss
@@ -20,7 +20,7 @@
     box-sizing: border-box; // 1
     display: table;         // 2
     max-width: 100%;        // 1
-    margin-bottom: govuk-spacing(3);
+    margin-bottom: govuk-spacing(2);
     padding: 0;
     // Hack to let legends or elements within legends have margins in webkit browsers
     overflow: hidden;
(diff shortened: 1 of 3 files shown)


src/components/fieldset/_fieldset.scss
```diff
@@ -32,20 +32,25 @@
...
   .govuk-fieldset__legend--m {
     @include govuk-font($size: 24, $weight: bold);
+    margin-bottom: govuk-spacing(3);
   }
 
   .govuk-fieldset__legend--s {
     @include govuk-font($size: 19, $weight: bold);
   }
 
+  .govuk-fieldset__legend
```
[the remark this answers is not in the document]
dashouse: Thanks!

igloosi:
this looks but it needs a changelog entry as well. is this a bug fix or a new feature?

igloosi:
resolved conflict and comment line

src/components/hint/_hint.scss
```diff
@@ -16,14 +16,35 @@
...
-  // This adjustment will not work when the label is inside the <h1>, however
-  // it is unlikely that the default or govuk-label--s class would be used in
-  // this case.
+  // This adjustment will not work when the label is inside the <h1>, however it
+  // is unlikely that the default or govuk-label--s class would be used in this
+  // case.
 
-  // This adjustment will not work in browsers that do not support :not()
-  // Users with these browsers will see the default size margin (5px larger)
+  // This adjustment will not work in browsers that do not support :not() Users
```
36degrees: Can we add a full stop here to keep these sentences from running into each other like this?

igloosi: adjusted

igloosi:
merged on behalf of @dashouse

PR #942, opened 2018-08-06
Allow attributes on checkboxes/radios

colinrotherham:
Unlike other form controls, the checkbox/radio templates only allow attributes on the wrapping container markup.

Services like ours add data attributes to some checkboxes:

```yaml
items:
  - value: yes
    text: Yes
    attributes:
      data-example: '123'
  - value: no
    text: No
    attributes:
      data-example: '456'
```

Following the same pattern as the other input, textarea and select, details templates etc, the attributes are appended like this:

```html
<input ... value="yes" data-example="123">
<input ... value="no" data-example="456">
```

Hopefully this pull request will help other teams too 😊

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,13 @@
 
   ([PR #N](https://github.com/alphagov/govuk-frontend/pull/N))
 
+- Allow attributes on checkboxes/radios
+
+  You can now provide attributes on checkbox and radio items
+  `attributes: { 'data-attribute': 'value' }`
+
+  ([PR #942](https://github.com/alphagov/govuk-frontend/pull/942))
+
 🔧 Fixes:
 
 - Pull Request Title goes here
--- a/src/components/checkboxes/README.md
+++ b/src/components/checkboxes/README.md
@@ -911,6 +911,18 @@ If you are using Nunjucks,then macros take the following arguments
 
 <tr class="govuk-table__row">
 
+<th class="govuk-table__header" scope="row">items.{}.attributes</th>
+
+<td class="govuk-table__cell ">object</td>
+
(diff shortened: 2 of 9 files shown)


CHANGELOG.md
```diff
@@ -28,6 +28,13 @@
 
   ([PR #N](https://github.com/alphagov/govuk-frontend/pull/N))
 
+- Allow attributes on checkboxes/radios
```
nickcolley: Since this is adding a new feature this changelog entry should be under the feature title.


- https://github.com/alphagov/govuk-frontend/blob/master/CONTRIBUTING.md#updating-changelog

src/components/checkboxes/checkboxes.yaml
```diff
@@ -43,6 +43,32 @@ examples:
         value: irish
         text: Irish
 
+- name: with attributes on items
```
nickcolley: Not documented, we are trying to only add .yaml examples where we cannot write a unit test for it.

In this case could you remove this example and test that this functionality works by adding a test in `checkboxes/template.test.js` please?

CHANGELOG.md
colinrotherham: @nickcolley No problem, moved it to "New features"

src/components/checkboxes/checkboxes.yaml
colinrotherham: @nickcolley Ah that's fine, makes sense. Added two new tests!

colinrotherham:
@nickcolley I've made two two review changes.

Tests all passing 👍

colinrotherham:
Updated checkbox/radio readme to document new `items.{}.attributes`

dashouse:
Looks good to me 👍

PR #945, opened 2018-08-07
Update header docs

web-bert:
The navigation items can take an optional active param
This was missing from the docs so this adds it in


web-bert:
👍

PR #947, opened 2018-08-08
Add ability to specify the logo text

web-bert:
The default logo text is GOV.UK this change allows you to specify your own text

### DIFF
--- a/src/components/header/README.md
+++ b/src/components/header/README.md
@@ -389,6 +389,18 @@ If you are using Nunjucks,then macros take the following arguments
 
 <tr class="govuk-table__row">
 
+<th class="govuk-table__header" scope="row">logoText</th>
+
+<td class="govuk-table__cell ">string</td>
+
+<td class="govuk-table__cell ">No</td>
+
+<td class="govuk-table__cell ">The text next to the crown logo. Defaults to GOV.UK</td>
+
+</tr>
+
+<tr class="govuk-table__row">
+
 <th class="govuk-table__header" scope="row">productName</th>
 
 <td class="govuk-table__cell ">string</td>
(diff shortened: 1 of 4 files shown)


36degrees:
Hi Graham,

Thanks for your PR. It'd be good to understand your use case for this change. The crown icon should only be used together with the word 'GOV.UK', which is why the text is currently hard-coded.

If we were going to make this change, I think we'd need to make it so that providing your own text (or HTML) results in the crown icon being omitted.

Ollie

web-bert:
We have many internal services and so we don't need/want the GOV.UK logo text, but would rather our own "umbrella" name for our family of services - so users know they are all part of the same system. We then use the `serviceName` to specify which service they are using.

We still want the GOV.UK feel, just slightly customised

web-bert:
@36degrees I have done what you suggested and remove the crown logo if custom text is used

hannalaakso:
Hi Graham,

Thanks a lot for your work on this 🙏We know that internal or "service" headers have use cases for service teams. We have [roadmapped](https://design-system.service.gov.uk/roadmap/) to work on "making it easier for services that don’t look like GOV.UK to use the Design System" over the next 6-12 months. Internal headers are not something that we have explored yet at any length so we won't be accepting this PR for now as it's necessary to confirm that this is the right solution to the problem and covers other related use cases.  

Would you be happy to [raise an issue](https://design-system.service.gov.uk/community/propose-a-component-or-pattern#2-raise-an-issue) in the Community backlog for Internal headers to investigate this further? We can also do this on your behalf.

Thanks,
Hanna

nickcolley:
@web-bert thanks Graham 👍

Issue #546, opened 2018-02-23
Add test coverage for .form-group


nickcolley:
We need to add test coverage to include .form-group and .form-group--errors in our macros.

Probably makes sense to use snapshots tests for these.

PR #948, opened 2018-08-08
Add test coverage for form groups

nickcolley:
Closes https://github.com/alphagov/govuk-frontend/issues/546


src/components/checkboxes/template.test.js
```diff
@@ -93,6 +93,13 @@ describe('Checkboxes', () => {
     expect($component.attr('data-second-attribute')).toEqual('second-value')
   })
 
+  it('renders with a form group wrapper', () => {
+    const $ = render('radios', {})
```
36degrees: should be `checkboxes`

PR #949, opened 2018-08-08
Remove template whitespace

web-bert:
In my mind I always think it's not good to start an HTML document with whitespace
I think a brower in the past didn't like it, it's something that I have always done
This uses the nunjucks whitespace controls to strip the leading whitespace
And also some other whitespace in the header

### DIFF
--- a/src/template.njk
+++ b/src/template.njk
@@ -1,8 +1,8 @@
-{% from "./components/skip-link/macro.njk" import govukSkipLink %}
-{% from "./components/header/macro.njk" import govukHeader %}
-{% from "./components/footer/macro.njk" import govukFooter %}
+{% from "./components/skip-link/macro.njk" import govukSkipLink -%}
+{% from "./components/header/macro.njk" import govukHeader -%}
+{% from "./components/footer/macro.njk" import govukFooter -%}
 {# specify absolute url for the static assets folder e.g. http://wwww.domain.com/assets #}
-{% set assetUrl = assetUrl | default(assetPath) %}
+{%- set assetUrl = assetUrl | default(assetPath) -%}
 <!DOCTYPE html>
 <html lang="{{ htmlLang | default('en') }}" class="govuk-template {{ htmlClasses }}">
   <head>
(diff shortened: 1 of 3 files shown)


nickcolley:
Hey Graham :)

Doing a quick Google confirms that some older browsers will be forced into 'quirks mode' if there is whitespace before the doctype.

Although I would have thought we would have noticed the impact of this in our cross browser testing...

If the issues are around whitespace at the beginning of the document could we change this pull request so it only removes it before the doctype declaration?

I'd like to add a test since this could be easily re-introduced in the future, but we could do this after your pull request lands if you're not comfortable doing this.

web-bert:
Hey @nickcolley 

I have just pushed an update with a test file. I learnt about the `trimBlocks` and `lstripBlocks` config options which I haven't used before, so I've added a test to cover both scenarios as they alter the output - although technically the default one will cover both scenarios, I think it's good to highlight there can be differences?

nickcolley:
Ah so this is only a problem if you do not use those Nunjucks configurations? 

I wonder if we should also look to improve the guidance around those too? (Not in this Pull Request)

src/template.test.js
```diff
@@ -0,0 +1,24 @@
...
+    })
+  })
+  describe('with nunucks block trimming enabled', () => {
+    it('should not have any whitespace before the doctype', () => {
+      nunjucks.configure(configPaths.src, {
+        trimBlocks: true,
+        lstripBlocks: true
+      })
+      const output = nunjucks.render('./template.njk')
+      expect(output.charAt(0)).toEqual('<')
```
nickcolley: could we do `output.startsWith('<!DOCTYPE html>')`?

web-bert:
It does affect both, just you get more whitespace with one than the other. When using the block trimming the macro import statements don't need the `-%}` at the end, but without they do. Both require the `set` to have the trimming at the start and end.

It would be good to add guidance around the use of the trim config.

src/template.njk
```diff
@@ -18,7 +18,7 @@
       <link rel="apple-touch-icon" sizes="167x167" href="{{ assetPath | default('/assets') }}/images/govuk-apple-touch-icon-167x167.png">
       <link rel="apple-touch-icon" sizes="152x152" href="{{ assetPath | default('/assets') }}/images/govuk-apple-touch-icon-152x152.png">
       <link rel="apple-touch-icon" href="{{ assetPath | default('/assets') }}/images/govuk-apple-touch-icon.png">
-    {% endblock %}
+    {% endblock -%}
```
nickcolley: Is this required to pass the tests?

src/template.test.js
```diff
@@ -0,0 +1,24 @@
...
+
+describe('Template', () => {
+  describe('with default nunjucks configuration', () => {
+    it('should not have any whitespace before the doctype', () => {
+      nunjucks.configure(configPaths.src)
+      const output = nunjucks.render('./template.njk')
+      expect(output.charAt(0)).toEqual('<')
+    })
+  })
+  describe('with nunucks block trimming enabled', () => {
```
nickcolley: nunucks -> nunjucks

src/template.test.js
web-bert: I did wonder whether to check for more, I just thought that as long as it stars with an angle bracket then we know it has no leading whitespace. I didn't want to tie the test to a specific doctype as we don't really care here?

src/template.njk
web-bert: Sorry missed that one when reverting...

src/template.test.js
web-bert: oops

web-bert:
Update pushed to fix last two issues

src/template.test.js
nickcolley: Good enough for me 👍

nickcolley:
I've raised an issue for the guidance to investigate separately.

nickcolley:
@web-bert could you please add a CHANGELOG fix entry?

web-bert:
Sure, and then rebase to tidy up as well?

nickcolley:
Yeah, I think we'd be ready to merge then, pending a final review.

web-bert:
No problem, happy to help

src/template.test.js
```diff
@@ -0,0 +1,24 @@
+/* eslint-env jest */
+
+const nunjucks = require('nunjucks')
+const configPaths = require('../config/paths.json')
+
+describe('Template', () => {
```
36degrees: 👏

PR #951, opened 2018-08-09
Fix removal of dotted button outline in Firefox

36degrees:
Firefox adds a small dotted border on a button when focussed. This can be overridden by using the `::-moz-focus-inner { }` pseudo-element. We’re currently attempting to do this, but we’re accidentally omitting one of the `:` so this is not being applied.

## Before

screen shot 2018-08-09 at 14 15 40

## After

screen shot 2018-08-09 at 14 15 52

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,15 @@
   Some older browser will be forced into 'quirks mode' if there is whitespace before the doctype.
 
   ([PR #949](https://github.com/alphagov/govuk-frontend/pull/949))
+
+- Remove additional dotted outline from focussed buttons in Firefox
+
+  This was already the intended behaviour, but a minor typo (: rather than ::)
+  meant that it wasn't being applied.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -27,13 +22,8 @@
 
   ([PR #942](https://github.com/alphagov/govuk-frontend/pull/942))
 
-🔧 Fixes:
```
nickcolley: fyi I think people are leaving these in for others to easily copy and paste in the future, before we cut a release

not a big deal to remove them though

36degrees: is it useful to have one in every section?

nickcolley: maybe not, feel free to propose a change to https://github.com/alphagov/govuk-frontend/blob/master/docs/contributing/CHANGELOG_TEMPLATE.md

PR #955, opened 2018-08-10
Adjust tag component padding to compensate for font spacing

nickcolley:
This is in response to [an issue raised on the backlog](https://github.com/alphagov/govuk-design-system-backlog/issues/62#issuecomment-412112217).

Until we update GOV.UK Frontend with the new font, NTA
sits higher than common fonts, so we need to add adjustments
to compensate.

Links for review:

- [Current phase banner](https://govuk-frontend-review.herokuapp.com/components/phase-banner/preview)
- [Changed tag with phase banner](https://govuk-frontend-review-pr-955.herokuapp.com/components/phase-banner/preview)

# Screenshots

### Tag using default font with no changes
screen shot 2018-08-10 at 16 49 54

---

### Tag using default font with changes
screen shot 2018-08-10 at 16 50 59

https://trello.com/c/GV0PxHFL/1319-fix-vertical-text-alignment-in-tag-component

### DIFF
--- a/src/components/tag/_tag.scss
+++ b/src/components/tag/_tag.scss
@@ -7,7 +7,11 @@
     @include govuk-font($size: 16, $weight: bold, $line-height: 1.25);
 
     display: inline-block;
-    padding: govuk-spacing(1) 8px 0;
+    padding: 4px 8px;
+    // Since New Transport sits slightly higher than other common fonts.
+    // We use intentionally uneven padding to make it balanced, this can be
+    // removed using the version of the font that has a more common vertical spacing.
+    padding-bottom: 1px;
 
     // When a user customises their colours often the background is removed,
     // by adding a outline we ensure that the tag component still keeps it's meaning.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,9 @@
 - Change spacing relationship on default and small legends and hints
   ([PR #940](https://github.com/alphagov/govuk-frontend/pull/940))
 
+- Adjust tag component padding to compensate for font spacing
+  ([PR #955](https://github.com/alphagov/govuk-frontend/pull/955))
+
 ## 1.2.0 (feature release)
 
 🆕 New features:


nickcolley:
Note that this adjusts the text it sits beside slightly:

screen shot 2018-08-10 at 17 12 00

src/components/tag/_tag.scss
```diff
@@ -7,7 +7,13 @@
     @include govuk-font($size: 16, $weight: bold, $line-height: 1.25);
 
     display: inline-block;
-    padding: govuk-spacing(1) 8px 0;
+    padding: 4px 8px;
+    // Since New Transport sits slightly higher than other common fonts.
+    // We use intentionally uneven padding to make it balanced, this can be
+    // removed using the version of the font that has a more common vertical spacing.
+    @if $govuk-font-family == $govuk-font-family-nta {
```
36degrees: If this isn't true, then the alpha tag ends up with 4px on top and bottom, which makes it 28px – considerably larger than the 'with nta' variant and no longer following our /5 rhythm. I think if we're going to try and make this 'conditional' we should make sure that the component still follows our rules in both cases.

In other places we're just commenting and applying adjustments by taking from one side and adding to the other, e.g. like this: https://github.com/alphagov/govuk-frontend/blob/master/src/components/button/_button.scss#L164-L178

nickcolley: My thought process:

1. As far as I know the 5 pixels is a scale based on how NTA works (might be wrong)
2. The component being a little bit bigger is much better than it looking completely broken as seen in the screenshots.

I'll make this less clever and we'll figure this out when we do proper theming support.

dashouse: The 5px thing is not related to the font, it's a decision based on the font-size / line-height combination. Each font size has a line-height rounded to the nearest 5px increment. This is why the spacing scale is increments of 5px and 10px and the inputs, buttons etc have been standardised at 40px (45px with the start button)

I say this, but this has mainly been considered for the main content of the page, there were way too many variables to make it work properly for the header and beforeContent area...haven't checked it in a while.

PR #958, opened 2018-08-15
Accordion component

frankieroberto:
_Note for reviewers:_

Please leave any comments pertaining to the **guidance/documentation** or the **content** used for the accordion on https://github.com/alphagov/govuk-design-system/pull/635.

Please leave any comments for the accordion component itself (**design, code**) on this PR.

Preview here: https://deploy-preview-635--govuk-design-system-preview.netlify.com/components/accordion/

---

This is work-in-progress on adding an [Accordion](https://github.com/alphagov/govuk-design-system-backlog/issues/1) component to the Design System.

(Opening this Pull Request mainly as a place to collaborate and discuss the component implementation)

So far I've just done the most minimal lift-and-shift from https://github.com/frankieroberto/accordion to get it working, which has meant:

* [x] Renaming `this.element` to `this.$module` (following the convention of existing components)
* [x] Moving all the setup code from the constructor (`new Accordion()`) to an explicit function call (`new Accordion().init()`), following the convention of existing components.
* [x] Adding a super-basic Nunjucks macro+template (I'm not quite sure why these are separate files but that seems to be the convention).
* [x] Some minimal CSS tweaks to stop the CSS linter complaining 😄 

### To-do (prior to Working Group review):

* [x] Update the CSS class names to follow Design System conventions
* [x] Update the CSS syntax itself to follow the [coding standards](https://github.com/alphagov/govuk-frontend/blob/master/docs/contributing/coding-standards/css.md).
* [x] Update the javascript to follow the [coding standards](https://github.com/alphagov/govuk-frontend/blob/master/docs/contributing/coding-standards/js.md)
* [x] Investigate pros and cons of using the `<details>` element instead of divs, and the `.open` DOM API.
* [x] Consider making the SVG image a data-url to avoid a network request?
* [x] Move "+/-" image to markup (see https://github.com/alphagov/govuk-frontend/pull/958#issuecomment-429760943)
* [x] Ensure "+/-" image is visible if [user changes the default colours](https://accessibility.blog.gov.uk/2018/08/01/supporting-users-who-change-colours-on-gov-uk/)
* [x] Add unit tests
* [x] Add integration tests
* [x] Add Nunjucks macro
* [x] Add macro options as yaml, similar to `params` [here](https://github.com/alphagov/govuk-frontend/blob/f8b816f528ae95843f91f044f8cfc8959f004834/src/components/back-link/back-link.yaml#L1)
* [x] Browser testing (ongoing)
* [x] Assistive technology testing
* [x] Documentation - [draft](https://docs.google.com/document/d/1a41Z4-FSAEdfnGIHkMdmLs5qsmf3tSsHvgVKkQMKyDI/edit)
* [x] Save state of each accordion to session storage - @injms is working on this
* [x] Update the focus state to be yellow solid outline
* [x] Consider using the same language as W3C who use the terms "accordion header" and "accordion panel", ie call ours "header" and "panel" in the markup, instead of "section header" and "section body"
* [x] Make into a `<button>` that is inside a `<h3>`
* [x] Add `id` that is used by `aria-labelledby` of relevant panel
* [x] Add `aria-controls` that refers to `id` of relevant panel
* [x] Add `aria-expanded` that is toggled `false` / `true` based on state
* [x] Add `id`to panel that corresponds to `aria-controls` of relevant header
* [x] Add `role="region" attribute to panel
* [x] Add `aria-labelledby` to panel that refers to `id` of relevant header

### To-do (can be done after Working Group review)
* [x] Fix "+/-" svg not showing IE9-11
* [x] Fix IE8 issues - see https://github.com/alphagov/govuk-frontend/pull/958#pullrequestreview-174974815
* [x] Add template tests
* [x] Add visually hidden text "panels" to "Open all / Close all" button so it says "Open all / close all panels"
* [x] Make the heading level for the macro configurable by doing something similar to: https://github.com/alphagov/govuk-frontend/blob/3e7f8f1203bf54b71eed786b418db9d7b3a0bd0f/src/components/panel/template.njk#L5
* [x] Remove HTML default <button> padding from `govuk-accordion__section-header-button` to leave the text hard aligned to the left edge
* [x] add `@include govuk-responsive-margin(6, "bottom");` to `.govuk-accordion` as all components / wrappers on this "outer" layer have the same bottom margin applied.
* [x] `.govuk-accordion__section` currently use magic numbers for the spacing. Ideally we should use padding values from the spacing scale (spacing scale 3, 15px) and override with current magic values in a separate section at the bottom of the scss file. This is down to where New Transport sits on the baseline compared to the default sans-serif, we have a new version of this the font that fixes this so it would be great to quickly remove the "magic numbers" block of CSS when this version of the font is released.
* [x] Add padding top and bottom (spacing scale 3, 15px) to `.govuk-accordion__section-panel`
* [x] Remove the margin bottom from last item inside .govuk-accordion__section-panel we do this on tabs, panel etc using:
```
& > :last-child {
 margin-bottom: 0;
}
```
* [x] Add a "safe area" using padding to stop the icon clashing with the additional description when the text is long / on mobile
* [x] Align the icons fully to the right on mobile size, it's less likely the hover state will be seen and will allow more space for the "safe area"
* [x] Change the component to use a `description` (or `summary` as we now call it that in the Design System?) option for adding the summary line (at the moment we pass in markup to do this)
* [x] Fix background colour of buttons on Firefox inverted colours view
* [x] Fix colour of button text on Firefox inverted colours view > looks like FF makes text in buttons dark
* [x] Rename panel to content - or vice versa
* [x] Prefix header, heading, summary, content classes with "section"
* [x] Add template tests - IN PROGRESS
* [x] Tidy up no-js classes
* [x] Tidy up Sass in general - nice to have
* [x] Fix failing tests
* [x] Do a team review of public API of component
* [x] Do a team design/guidance review of preview in Design System
* [x] Do a dev team code review of Frontend PR
* [x] Decouple JS classes and style classes - not doing this for now, we'll have to discuss general approach to this with devs
* [x] Should we not print out component in template if user hasn't passed through both heading and content?> we can look at this later
* [x] Final cross browser testing
* [x] Final assistive tech testing

References:
[W3C Authoring Practises - Accordion](https://www.w3.org/TR/wai-aria-practices-1.1/#accordion)
[W3C Accordion example](https://www.w3.org/TR/2017/NOTE-wai-aria-practices-1.1-20171214/examples/accordion/accordion.html)


frankieroberto:
Hello. I've not used your recommended unit test framework, Jest, before, but it looks straightforward, so I'm not too worried about that. That said, I'm not sure how much of the `Accordion` class should be exposed as public properties / functions vs private?

Do you have a standardised way of doing integration testing at all, eg setting up an HTML page, adding the javascript, and then clicking the various elements and asserting what should happen? Selenium?

frankieroberto:
Just realised that Jest _does_ do full integration testing, so ignore that question...

frankieroberto:
Do you have any thoughts on how the javascript should be structured. Currently, both `Accordion`, `AccordionSection` and all their prototype functions are exposed as public functions. Not only that, but `AccordionSection` has to be passed an explicit reference to the parent `Accordion` object (so that it tell the parent object to update the state of the "Open/close all" button when one of the sections is opened/closed.

This feels a bit messy.

An alternative pattern might be something like this:

```javascript
function Accordion(element) {
  this.sections = [new Section()]

  function Section(sectionElement) {
    this.expanded = false
  }

  Section.prototype.setExpanded = function(expanded) {
    this.expanded = expanded
  };
}

Accordion.prototype.setExpandAll = function(expandAll) {
  for (section of this.sections) {
    section.setExpanded(expandAll)
  }
}
```

Which would only expose an `expanded` property on each of the sections, a `setExpanded()` function on the settings, and a `setExpandAll()` function on the parent:

```javascript
var accordion = new Accordion(document.querySelector('.govuk-accordion'))
accordion.sections.length
=> 2
accordion.sections[0].setExpanded(true)
accordion.sections[0].expanded
=> true
accordion.setExpandAll(true)
accordion.sections[1].expanded
=> true
```

(it'd be nice to be able to avoid having a separate setter function, but I'm not sure how easy that is with javascript right now).

Any thoughts?

src/components/accordion/_accordion.scss
```diff
@@ -0,0 +1,86 @@
...
+
+
+@include govuk-exports("govuk/component/accordion") {
+
+  .govuk-accordion__controls {
+    text-align: right;
+  }
+
+  // This is styled to look like a link not a button
+  .govuk-accordion__expand-all {
```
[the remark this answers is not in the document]
frankieroberto: Good idea. I think I was trying to be consistent with the naming of the `aria-expanded` attribute name, but "open" is probably more understandable than "expanded" (and is also more consistent with the `<details>` API).

src/components/accordion/_accordion.scss
```diff
@@ -0,0 +1,86 @@
...
+    @include govuk-font($size: 16);
+    display: inline;
+    border-width: 0;
+    color: $govuk-link-colour;
+    background: transparent;
+    cursor: pointer;
+  }
+
+  // Hide the body of collapsed sections
+  .govuk-accordion__section[aria-expanded="false"] .govuk-accordion__section-body {
```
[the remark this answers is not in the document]
frankieroberto: @hannalaakso so, I've ummed and ahhed on this.

My thinking was that there are some advantages to relying on the `aria-expanded` attribute. Namely that:

* it means that if someone uses alternative javascript (but the same CSS) they won't accidentally forget to include the `aria-expanded` attribute (which would otherwise be invisible, except to those using screen readers).
* it means the javascript doesn't have to update both the `aria-expanded` attribute and an `expanded` CSS class. (This is admittedly not that big a deal)

However there may be good arguments not to have the visual implementation dependent upon the accessibility API... (personally on balance I think it's probably a net positive)

The `govuk-accordion__section--expanded` class currently serves a single purpose, which is to indicate the "initial state" of the component (which defaults to collapsed but can be override to expanded). This class is then removed by the setup javascript and swapped to the `aria-expanded` role (at https://github.com/alphagov/govuk-frontend/pull/958/files#diff-75bb29fcad7b97e804de3fd00e8252d6R114) - the rationale being that it'd be bad to have an `aria-expanded` role on the initial page HTML before the javascript loads, as at that point we can't be certain whether the javascript will load or not.

src/components/accordion/_accordion.scss
```diff
@@ -0,0 +1,86 @@
...
+
+  // Border at the bottom of the whole accordtion
+  .govuk-accordion.with-js {
+    border-bottom-width: 1px;
+    border-bottom-style: solid;
+    border-bottom-color: $govuk-border-colour;
+  }
+
+  // Borders between accordion sections
+  .govuk-accordion.with-js .govuk-accordion__section {
```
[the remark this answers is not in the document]
frankieroberto: 👍 on switching to `js-enabled` from `with-js` for consistency with existing components.

I'm happy to make use of SASS nesting too – wasn't sure from your [coding standards](https://github.com/alphagov/govuk-frontend/blob/master/docs/contributing/coding-standards/css.md#nesting) whether that was encouraged or not.

src/components/accordion/_accordion.scss
```diff
@@ -0,0 +1,86 @@
...
+  // Headings in section headers have link colours as an additional affodance
+  .govuk-accordion.with-js .govuk-heading-m {
+    margin-top: 0;
+    margin-bottom: 0;
+    color: $govuk-link-colour;
+  }
+
+  // Section headers have a grey background on hover as an additional affodance
+  .govuk-accordion.with-js .govuk-accordion__section-header:hover {
+    background-color: govuk-colour("grey-4");
```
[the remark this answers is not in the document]
frankieroberto: @hannalaakso good question. Current practice on the [Service Manual](
https://www.gov.uk/service-manual/helping-people-to-use-your-servic) and at the DVSA is to wrap the `h2` in a button, so that gets focus and displays an orange border.

(The DVSA implementation also wraps the + and - icon in a `<button>` element, so that gets focus too, which seems redundant and not massively accessible)

My implementation currently assigns a `role=button` to the whole header element, on the basis that this better matches the actual clickable area.

However I think adding the same background colour on focus as on hover is a good idea.

src/components/accordion/_accordion.scss
```diff
@@ -0,0 +1,86 @@
...
+  // Section headers have a pointer cursor as an additional affordance
+  .govuk-accordion.with-js .govuk-accordion__section-header {
+    position: relative;
+    padding-top: 14px;
+    padding-bottom: 12px;
+    cursor: pointer;
+  }
+
+  // Headings in section headers have link colours as an additional affodance
+  .govuk-accordion.with-js .govuk-heading-m {
```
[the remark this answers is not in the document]
frankieroberto: 👍

src/components/accordion/_accordion.scss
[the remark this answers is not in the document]
hannalaakso: I think it would be good not to couple visuals with `aria` attributes as they serve two separate functions. I would be in favour of adding both `govuk-accordion__section--expanded` and `aria-expanded` to open sections. 

Definitely makes sense not to have an `aria-expanded` attribute on initial page load but add it with JS 👍

src/components/accordion/_accordion.scss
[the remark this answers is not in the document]
hannalaakso: Yeah we decided against SASS nesting for readability, it's documented [here](https://github.com/alphagov/govuk-frontend/blob/master/docs/contributing/coding-standards/css.md#nesting). It feels like it's one of those situations where there are benefits to both approaches. At the same time I think in the long run this will make the codebase easier to maintain.

hannalaakso:
Hi @frankieroberto, as discussed I've pushed some changes to the JS, these address the structure of the JS questions. We can keep discussing this obviously.

@joelanman Do you have any thoughts around using "+ or -" or arrow up/down as we use with details component? Another question: should we consider an enhancement of storing the open/closed state of sections? 

Review app [here](https://govuk-frontend-review-pr-958.herokuapp.com/components/accordion)

hannalaakso:
This [bug](https://github.com/alphagov/govuk-design-system-backlog/issues/44#issuecomment-421363962) means that it's not possible to expand `details` element with print styles. Until it's fixed we probably shouldn't use `details` element with accordion.

joelanman:
On the open/close control:

- Service manual and GOV.UK Pay use a plus/minus and say they didn't find any issues, so didn't try anything else.

- [DVSA say they tried plus/minus and up/down arrow and the arrows worked better](https://github.com/alphagov/govuk-design-system-backlog/issues/1#issuecomment-381904555)

- [Step by step navigation](https://www.gov.uk/learn-to-drive-a-car) use quite a different approach. They started with plus/minus. They ended up using the words Show and Hide, on the left under the heading. Being on the left means screen magnifier users can see the control.

I would say from this that we have some evidence that arrows are clearer than plus/minus.

The Step by step approach is probably less ambiguous, with words rather than images, and it has an accessibility advantage with the controls on the left. However it is not as widely used as the other accordion design and so I don't know if we can recommend it for all services.

frankieroberto:
DfE had an issue with the use of the Back button and the existing accordion, their fix for which is summarised here: https://github.com/DFE-Digital/search-and-compare-ui/pull/156

This makes me think we should do some research into whether/how different browsers restore state when using the back button – we've definitely observed some differences between browsers for this.

dankmitchell:
We'd also want to ensure that the accordions don't all flash open whilst the javascript is loading

frankieroberto:
@dankmitchell do the existing ones do that at all?

frankieroberto:
@hannalaakso any thoughts on how we should "inline" (via base64 encoding) the SVG image? I think it makes sense to keep the SVG file in the repo, but have the inlining be part of the sass compilation step? In which case we'd need to pick an npm package like https://www.npmjs.com/package/sass-inline-image to add a sass function?

frankieroberto:
Work-in-progress on the documentation is here: https://github.com/alphagov/govuk-design-system/pull/579

src/components/accordion/_accordion.scss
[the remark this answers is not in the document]
hannalaakso: @joelanman Would you have any thoughts on whether we should be applying the grey background colour when a section heading is in focus (similar to hover state)? Review app [here](https://govuk-frontend-review-pr-958.herokuapp.com/components/accordion).

src/components/accordion/template.njk
```diff
@@ -0,0 +1,14 @@
+<div class="govuk-accordion" data-module="accordion">
+
+  {% for item in params.items %}
+    <div class="govuk-accordion__section {% if item.expanded %}govuk-accordion__section--expanded" {% endif %}">
+      <div class="govuk-accordion__section-header">
```
hannalaakso: I'm wondering if we should consider making this a `button` element and also using `aria-controls` and `aria-labelledby` to indicate the relationship? I'm just looking at a [W3C example](https://www.w3.org/TR/wai-aria-practices/examples/accordion/accordion.html).

frankieroberto: @hannalaakso I think the primary semantics here are it being a heading (`h2`), so I wouldn't want to lose that, but we could do either `<h2><button>label</button</h2>` (via javascript), or add an `aria-role="button"` attribute?

hannalaakso:
Decisions from catch up with @frankieroberto and @joelanman:

- We’re going to go with +/- for the accordion heading button (as opposed to ^/v) as there are more examples of services using this pattern. 

- Re: how to handle the +/- SVG, we currently add the SVG directly into the markup for the header logo and have a fallback PNG in assets for browsers that don't support this so this would be the default way of doing this. The upside is that we can change the colour easily for users who change their colours. Having the base 64 encoded SVG in Sass could also be an option but it would need to support filling the colour when user changes their colours. @frankieroberto might do some more investigation if he has time.

- We’re going use a light grey hover colour for the accordion heading (focus state will be the standard GOV.UK orange border) - there didn't seem to be a good reason not to have this extra visual indication even though there aren't examples of services using it. It can be marked as requiring more research.

- We're working on draft guidance [here](https://docs.google.com/document/d/1a41Z4-FSAEdfnGIHkMdmLs5qsmf3tSsHvgVKkQMKyDI/).

hannalaakso:
Addition to the minutes from last week:
- We discussed storing of state and decided that since this is something Service Manual accordions use (example [here](https://www.gov.uk/service-manual/helping-people-to-use-your-service)), we should try to get it in if we can (but it shouldn't delay the initial release of the component if it comes to it). We’ll look into using session storage for this. We can either advise in guidance that service teams will need to make sure each accordion item has a unique ID or we can look into generating a unique ID based on a combination of page url and indexed ID for the accordion item.

hannalaakso:
Just a note to say that I've done some testing with Jaws and NVDA and have added some accessibility to-do's to the PR description.

src/components/accordion/template.njk
hannalaakso: Thanks @frankieroberto. Looking at the W3C specs, `<h2><button>label</button</h2>` is advised. I've updated the PR description.

src/components/accordion/accordion.yaml
```diff
@@ -34,6 +36,7 @@ examples:
 
 - name: with all sections already open
```
joelanman: is there a need for this? Unless we support 'maintain state' aren't accordions always going to start closed?

frankieroberto:
@hannalaakso I've made the accessibility changes you suggestion, except for a couple of differences, which I think are worth discussing:

* The section header is an `h2` instead of an `h3`. Possibly we need to make this configurable, but in my experience the accordions headers are usually the second-level heading rather than the third.
* I've used `role=button` rather than a `<button>` element within the `h2`. This is mainly because the javascript has to add the button semantics rather than the nunjucks macro (as we don't want a button to be present if javascript is disabled), and it's a bit easier to add an attribute rather than copying the node, removing it, and adding a replacement with a different element name. However we _could_ still do that, if it proves to be better to have a `<button>` element rather than an role.

Other comments:

* I've added the `role=region` to the panel. However I did note that this was optional within the W3C guidance, which also says:

> Avoid using the region role in circumstances that create landmark region proliferation, e.g., in an accordion that contains more than approximately 6 panels that can be expanded at the same time.

- should we add extra logic to only add it when there's less than < 6 panels? Or are we confident that in most circumstances, it's better to have the `region` role than to not?

src/components/accordion/accordion.yaml
frankieroberto: @joelanman not always, it seems. One service got in touch with me to say that they have different sections open by default depending upon answers the user has previously given.

frankieroberto:
@hannalaakso oh, another question I had was whether we should use `aria-hidden` or `hidden`? The example at https://www.w3.org/TR/wai-aria-practices-1.1/examples/accordion/accordion.html uses `hidden`, which I've not come across before, but seems to now be defined in the HTML spec: https://html.spec.whatwg.org/multipage/interaction.html#the-hidden-attribute

hannalaakso:
@frankieroberto I think `aria-hidden` is safer as it looks like [support](https://caniuse.com/#search=hidden) on old IE for `hidden` is poor.

frankieroberto:
@hannalaakso we wouldn't be using the `hidden` attribute to actually hide the element (as the class does that), just to signal to assistive technology software that that section is hidden, so it's probably more important to look at support for `hidden` vs `aria-hidden` in JAWS, etc?

I suspect that `aria-hidden` is more widely supported though – I was just surprised to see `hidden` in https://www.w3.org/TR/wai-aria-practices-1.1/examples/accordion/accordion.html

hannalaakso:
Interesting! So in my understanding it does matter whether the browser understands something as it builds an [accessibility tree](https://developers.google.com/web/fundamentals/accessibility/semantics-builtin/the-accessibility-tree) that acts as an API for the screen reader. 

I take your point though that it's intriguing that W3C have decided to use `hidden` in their example 👍

frankieroberto:
@hannalaakso we could add both, of course, to cover both bases? I'll do a bit more research.

src/components/accordion/_accordion.scss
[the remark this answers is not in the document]
hannalaakso: I think `expand-all` is actually okay, we talk about "expanded" state with aria.

src/components/accordion/_accordion.scss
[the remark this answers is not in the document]
hannalaakso: @frankieroberto Did you get a chance to look into switching this to use `js-enabled`? It would be good if it was consistent with the other components.

src/components/accordion/_accordion.scss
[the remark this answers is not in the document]
hannalaakso: This was resolved by https://github.com/alphagov/govuk-frontend/pull/958#issuecomment-429760943, we're keeping the background colour.

frankieroberto:
Here’s an interesting article comparing different ways to hide content, and their effect on accessibility: https://developer.paciellogroup.com/blog/2012/05/html5-accessibility-chops-hidden-and-aria-hidden/

src/components/accordion/template.njk
```diff
@@ -0,0 +1,16 @@
+{% set id = params.id %}
+
+<div class="govuk-accordion" data-module="accordion" id="{{ id }}">
+
+  {% for item in params.items %}
+    <div class="govuk-accordion__section {% if item.expanded %}govuk-accordion__section--expanded" {% endif %}">
+      <h2 class="govuk-accordion__section-header">
+        <div class="govuk-accordion__section-header-button govuk-heading-m" id="{{ id }}-heading-{{ loop.index0 }}">{{ item.label }}</div>
```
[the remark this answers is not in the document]
frankieroberto: @hannalaakso no good reason, I forgot there was an `index` alternative. I'll switch to that.

src/components/accordion/accordion.js
```diff
@@ -0,0 +1,169 @@
...
+  // always reflow after a setAttribute call.
+  this.$module.className = this.$module.className
+}
+
+Accordion.prototype.isExpanded = function ($section) {
+  return ($section.classList.contains('govuk-accordion__section--expanded'))
+}
+
+Accordion.prototype.setHeaderAttributes = function ($header, index) {
+  $header.setAttribute('tabindex', '0')
```
hannalaakso: `tabindex` is not needed here as the button itself is focusable.

src/components/accordion/accordion.js
```diff
@@ -0,0 +1,169 @@
...
+
+  var $button = $header.querySelector('.govuk-accordion__section-header-button')
+  $button.setAttribute('aria-controls', this.moduleId + '-panel-' + index)
```
hannalaakso: `index` now needs to start at 1 following the change from https://github.com/alphagov/govuk-frontend/pull/958#pullrequestreview-169372580. Apologies I didn't realise the change would affect this.

src/components/accordion/accordion.js
```diff
@@ -0,0 +1,169 @@
...
+  $button.setAttribute('role', 'button')
```
hannalaakso: Doing some testing and it looks like we should go with an actual `<button>` element. Jaws and NVDA don't announce when`aria-expanded` changes if using `role=button` whereas with `<button>` they do.

src/components/accordion/accordion.yaml
```diff
@@ -0,0 +1,99 @@
...
+- name: id
+  type: string
+  required: true
+  description: Used to assign an `id` in the HTML for the accordion as a whole, and also a prefix for the `id`s of the section panels and the buttons that open them, so that those `id`s can be the target of `aria-labelledby` and `aria-control` attributes.
+- name: items
+  type: array
+  required: true
+  descripion: An array of sections within the accordion.
+  params:
+    - name: label
```
hannalaakso: I wonder if calling this `label` risks confusing things with form elements - what about `headerText`?

src/components/accordion/accordion.yaml
hannalaakso: Since we are going to support maintaining state, I think this is okay. Resolving but feel free to unresolve if needed.

injms:
I've pushed the functionality for the accordion panels to remember which state they're in to [this branch](https://github.com/frankieroberto/govuk-frontend/tree/accordion-should-remember-open-closed-state), but I'm struggling through getting the tests up and running. 

Manipulating the page in an individual test seems to work okay, but reading the state of the `sessionStorage` in the headless browser is proving tricky. My current workaround is to read the state of the accordion after a refresh rather than the `sessionStorage` directly - we'll see how that goes.

@frankieroberto, would it be okay to merge the branch into your branch so that at least the functionality part can be reviewed?

frankieroberto:
@injms yep, let’s do that. I think refreshing the page and asserting that the state is maintained is fine? (Perhaps even preferable?)

I'll merge your branch in now. You can commit any further changes directly to this branch if needed.

frankieroberto:
@hannalaakso good idea. Should we make `h2` the default though, as that’s so common?

And would it be more intuitive to specify the `headingLevel` as `3` or `h3`, or should we accept either?

hannalaakso:
I think `h2` as default sounds good. I think we could go with numeric values for now? The panel component does this and we don't have any evidence that this is blocking users. We could always address it later if it becomes an issue.

src/components/accordion/accordion.js
hannalaakso: Did some testing and this looks good now with Jaws and NVDA 👍 I think the code for this looks okay but we'll do more reviews of it before publishing.

This *could* affect the details component, I'm going to have a look at that.

hannalaakso:
Did some testing of the component after the latest round of accessibility fixes with the following and it's looking good 🎉 
- Jaws 17 with IE11
- NVDA 2018 with latest Firefox
- Dragon Naturally Speaking 15 with IE11

Noticed that the component is throwing up an error in IE8 - this might be a recent regression, we should check. We might also just want to fall back to the no-JS experience for IE8 which is having the accordions open.

36degrees:
This is looking really good 👍 Thanks for all your work on this so far.

I just wanted to flag a couple of minor things I've noticed. I don't think any of them are blocking for the review session, but I think we should try and address the first two before this gets released.

You can't currently open or close the accordion using the space bar – or more precisely, it will open or close but it'll almost immediately then revert to the previous state.

The 'open all' / 'close all' button is still using the native browser focus state, without the yellow outline or fill:

govuk-frontend-review-pr-958 herokuapp com_components_accordion_preview

Also, the section headings currently use an 'outline' focus state, which matches the focus state for the accordions in the service manual. However, I think we've more or less standardised on a filled focus state (yellow background, black foreground) for all other 'text-like' interactive elements across frontend (e.g. links, breadcrumbs, details) – not sure if @dashouse has an opinion on this?

govuk-frontend-review-pr-958 herokuapp com_components_accordion_preview 1

Thanks!

frankieroberto:
@dashouse @36degrees I've fixed the focus style for the "Open all" button now.

I think the yellow outlined style works better than the yellow background colour here, as they're not links but buttons that have state? That's also the existing behaviour, eg on https://www.gov.uk/service-manual/the-team

frankieroberto:
@36degrees thanks for catching the spacebar bug. That was caused by switching the element to a `<button>` which fires a `click` event as well as a `keypress` event, and which isn't cancelable from within the `keypress` event handler (although enter is, and spacebar is from a `keyup` handler - I have no idea why).

I've remove the custom keyboard event handlers now. Hopefully the keyboard support for buttons is widespread enough not to require a custom implementation.

src/components/accordion/accordion.yaml
nickcolley: label mirrors the Tab component at least...

https://design-system.service.gov.uk/components/tabs/

frankieroberto: @nickcolley Yeah, I think I borrowed the terminology from that. Happy to switch it to `heading` though. I suggest we used `heading` rather than `headingText`, as that would allow us to support `heading.text` and `heading.html` if necessary.

src/components/accordion/accordion.js
```diff
@@ -0,0 +1,245 @@
...
+    var header = $section.querySelector('.govuk-accordion__section-header')
+    this.setHeaderAttributes(header, i)
+
+    this.setExpanded(this.isExpanded($section), $section)
+
+    // Handle events
+    header.addEventListener('click', this.onToggleExpanded.bind(this, $section))
+
+    header.addEventListener('click', function() {
+      console.log('clicked')
```
hannalaakso: We should remove all `console.log`'s as IE8 throws the error

```
 'console' is undefined
```
(unless Dev tools are open)

nickcolley:
@hannalaakso I think the polyfilling is actually an issue with this bug: https://github.com/alphagov/govuk-frontend/issues/1044

In terms of SVG support, we could consider not injecting them in IE8 or using CSS only: https://jsbin.com/pemifiq/4/edit?html,css,output

src/components/accordion/accordion.js
frankieroberto: @hannalaakso opps, well spotted, I should have removed those debugging lines earlier.

frankieroberto:
I've just pushed #838564a which keeps the `console.error` messages to warn developers that the state can't be stored if the aria attributes aren't present, as that feels useful, but I've updated it to feature-detect availability of the `console`, so that it doesn't break in IE8 (hopefully).

src/components/accordion/accordion.js
```diff
@@ -0,0 +1,241 @@
...
+  return ($section.classList.contains('govuk-accordion__section--expanded'))
+}
+
+Accordion.prototype.setHeaderAttributes = function ($header, index) {
+  var $button = $header.querySelector('.govuk-accordion__section-header-button')
+
+  // Copy existing div element to an actual button element, for improved accessibility.
+  // TODO: this probably needs to be more robust, and copy all attributes and child nodes?
+  var $buttonAsButton = document.createElement('button')
+  $buttonAsButton.setAttribute('class', $button.getAttribute('class'))
```
nickcolley: this needs the `type` attribute to be set to 'button' otherwise it could submit forms

frankieroberto: @nickcolley good spot. Dang, I thought that had been done already.

hannalaakso: @frankieroberto Thanks for adding this to the "Open all" button. This should be done for the accordion button as well.

adamsilver:
This looks **really** good—nice one, Frankie!

Here’s some notes/questions:

1. There’s a typo in the Nunjucks macro documentation: it says “Required. undefined See items”

2. Is “Open all” clear enough for screen reader users? I wonder if it should have “sections” (or similar) added to the end using visually hidden? Same for “Close all”

3. Is the “Open all/Close all” toggle button text configurable?

4. Hover state: is the subtle change of background colour obvious enough? We had a subtle grey used on Tabs (https://github.com/alphagov/govuk-design-system-backlog/issues/100#issuecomment-414043554)

5. Can it be configured so that particular sections start open?

6. Can it be configured so that all sections start open?

7. The focus outline only surrounds the button on the left even though the entire row is clickable. Is there a reason for this? I’d suggest the whole thing gets the focus outline.

8. Further to (7) could the `<button>` wrap the entire row contents including the H2 and SVG icon.

9. Could you add focusable=false to the SVG so that it’s not part of the tab sequence?

10. Could you add a visually-hidden class to the SVG so that it’s not read out for screen reader users as the `<button>` seems to provide enough information?

11. While forms shouldn’t be used with accordions, there is a potential need to have form controls inside an accordion like this for filtering controls. See the following links:
- http://hmcts-design-system.herokuapp.com/patterns/filter-a-list 
- http://nostyle.herokuapp.com/examples/filter-form (I’m not saying it’s always right or necessary to collapse filter controls, but there may be a need depending on the situation such as screen size or amount of options within a filter category)

12. With the filters, it's worth noting that the accordion label/heading is actually a legend. I wonder if we can make the accordion component work on legends etc.

13. There is a bug, at least in the second example where the `<h2>` contains the `<button>` and then another `<h2>` inside it.

14.  Can it be configured so that the accordion doesn't have the “Open all” button?

frankieroberto:
@adamsilver thanks for your comments!

Quick responses below:


1. Fixed, thanks. c9d0851

2. I’m not sure, to be honest. Existing accordions all seem to use "Open all" and I’ve not heard of any adverse findings. Whether "sections" is the right word, or would make it clearer, I can’t really say.

3. It’s not. It could be though, if there was a need. We’d lose consistency.

4. For tabs, it looks like you decided against a hover state altogether? My feeling here was that the subtle grey on hover was helpful, as it communicates the size of the hit area (ie that you don’t have to specifically hit the icon or the heading text), but non-essential.

5. Yes, via an `expanded` boolean paramter. I forgot to add this to the Nunjucks macro documentation though, which I’ve now fixed. eef03d4

6. Yes, by setting all sections to have `expanded` be `true`.

7. I just followed the [existing behaviour](https://www.gov.uk/world/albania) for this. It _could_ outline the whole row though, which would help to reinforce the larger hit area?

8. This seems sensible.

9. Are their any browsers which make SVG elements tabbable by default? I’ve not seen this pose any issues yet, but I may have missed something.

10. Similarly, are there any screen reader users which read out SVG elements (that don't contain any text) by default? If so, I’d be happy to add this.

11. Interesting use-case! Any thoughts on what the most intuitive way to specify this within the Nunjucks parameters might be?

12. That’s a bug with the documentation. The `label.html` param shouldn’t contain an h2. Will fix.

13. No. Are there any good reasons why this would be needed?

adamsilver:
@frankieroberto responses below

> 2. [On hidden text for open all button] I’m not sure, to be honest. Existing accordions all seem to use "Open all" and I’ve not heard of any adverse findings. Whether "sections" is the right word, or would make it clearer, I can’t really say.

> 3. [On configurable open all button text] It’s not. It could be though, if there was a need. We’d lose consistency.

I noticed in previous conversations there was a suggestion to use the word “Panel”. Either way, it would be good to make the text configurable I think for cases such as translating to Welsh.

> 4. [On hover] For tabs, it looks like you decided against a hover state altogether? My feeling here was that the subtle grey on hover was helpful, as it communicates the size of the hit area (ie that you don’t have to specifically hit the icon or the heading text), but non-essential.

I think that was an oversight for tabs. And I agree the hover state is helpful but it won't be seen by some users because of the lack of contrast.

> 7./8. [focus state] I just followed the existing behaviour for this. It could outline the whole row though, which would help to reinforce the larger hit area?

Yes, I'd say so using a `<button>` to wrap the whole thing.

> 9. Are their any browsers which make SVG elements tabbable by default? I’ve not seen this pose any issues yet, but I may have missed something.

Yes—IE.

> 10. Similarly, are there any screen reader users which read out SVG elements (that don't contain any text) by default? If so, I’d be happy to add this.

Not sure about this. If no text it makes sense that this is fine. Ignore me :)

> 11./12. Interesting use-case! Any thoughts on what the most intuitive way to specify this within the Nunjucks parameters might be?

I'm not sure it's something that the Nunjucks template needs to be considered with. It's more of a JavaScript thing.

Tell the JavaScript instance to operate on a legend (and sub container) instead by sprinkling classes onto things.

> 13. [My 14. on configuring the open all button to be excluded] No. Are there any good reasons why this would be needed?

I'm not sure. Perhaps if the accordion is made up of just a few sections it's not needed? As I read my argument it doesn't sound that convicing does it :). It's the whole “Versatile” point. I don't know what I don't know so allowing users to configure this seems sensible.

frankieroberto:
@adamsilver thanks your follow-up comments. Some thoughts below:

> I noticed in previous conversations there was a suggestion to use the word “Panel”. Either way, it would be good to make the text configurable I think for cases such as translating to Welsh.

I've discussed this briefly with the Design System team, and we felt this probably needs a bigger conversation about how we handle configurable/translatable text within javascript components (such as the existing Character Count one). Options could be data attributes within the HTML, or configuration files for the compiled javascript. We might need to postpone this decision for now though.

> I think that was an oversight for tabs. And I agree the hover state is helpful but it won't be seen by some users because of the lack of contrast.

I’d be open to increasing the contrast slightly by using a slightly darker grey (any suggestions?), but I wouldn't want to make it _much_ darker as then the contrast between the background colour and the text (which is way more important) might be too low.

>>  7./8. [focus state] I just followed the existing behaviour for this. It could outline the whole row though, which would help to reinforce the larger hit area?

> Yes, I'd say so using a `<button>` to wrap the whole thing.

There’s already a button wrapping the whole thing (apart from the SVG, which is absolutely positioned anyway), so it wouldn't be difficult to make it span the whole height and width of the header section. My only slight fear is that it'd make the focus outline harder to perceive, as the outline would be right next to the borders at the top and bottom. But it could be worth a try.

Re: the SVG element. Thanks for confirming that it's not read out by any screen readers (without containing any text). Are you saying that without text, it's still tabbable by IE though? And if so, which versions? I'd be happy to double-check and add `focusable=false` if necessary.

Re: your last example. This _should_ be possible with the existing javascript code (but not the macro), as that works on class names. I'll double-check though.

adamsilver:
> I’d be open to increasing the contrast slightly by using a slightly darker grey (any suggestions?)

Can you just try the next grey up?

> But it could be worth a try.

Cool

> Are you saying that without text, it's still tabbable by IE though?

I believe so.

> works on class names. I'll double-check though.

Cool

stevenaproctor:
@hannalaakso 'panel' seems a little jargon-y to me. Would 'sections' not be better? Or is the idea that it should be configurable to whatever you are using the accordion for?

stevenaproctor:
@hannalaakso Also, is this the best way to hide content. I thought there was a bug in JAWS that meant content could be read out of order. It was mentioned in an accessibility review a while ago. So `Open all <span class="govuk-visually-hidden">panels</span>` could be read out as 'panels open all'.

We have been recommending a visual span and a visually hidden span. So:
```<span aria-hidden="true">Open all</span>
<span class="govuk-visually-hidden">Open all panels</span>```

What do you think?

**Comment by hannalaakso** [comment]
Thanks @stevenaproctor. I think "sections" makes more sense, I'll change the wording.

Regarding the Jaws bug, would you have any more information on the device / browser / version of Jaws? I tested this with Jaws 15, 16, 18 and 2018 and couldn't replicate the issue. The pattern has been used for a while on [Check your answers](https://govuk-prototype-kit.herokuapp.com/docs/templates/check-your-answers) and it hasn't come up there.

**Comment by stevenaproctor** [comment]
@hannalaakso The JAWS bug was a while ago and it was something I was told rather than had happen to me. I will try to find out more.

**Comment by hannalaakso** [comment]
https://github.com/alphagov/govuk-frontend/pull/958/commits/0b4ca3eae92dfb02d019698fce42f8ff9cc113cf contains results from assistive tech testing done on the component. Specifically:

Summary is now specified as an option in Nunjucks and is placed after
the heading in the markup. This achieved the most consistent results when testing with
assistive tech. On encountering the heading, the assistive tech read out
the following:
- NVDA (Firefox): _clickable [heading text] button collapsed [summary text] heading level 2_
- Jaws 16 and Jaws 2018 (IE11): _[heading text] heading level 2 button collapsed [summary text]_
- iOS VoiceOver (Safari): _[heading text] button heading level 2 expanded [summary text] double tap to collapse_
- Mac Voiceover (Safari): _[heading text] you are currently on a heading level 2_

Mac VoiceOver doesn't give out a lot of information or read the summary and I can't find good
examples of this being done on collapsible elements so hoping someone else in the community might have ideas and I'm also going to run it past the Accessibility team at GDS. An alternative
(where Mac Voiceover does read out the summary) is to nest the summary inside the `<h2>` but this seriously limits what could be placed in the summary as `<h2>` can only contain [flow content](https://html.spec.whatwg.org/multipage/dom.html#flow-content) so it seems likely that this would lead to users inadvertently creating non-semantic markup.

**Comment by hannalaakso** [comment]
Thanks @stevenaproctor 🙂 

FYI after you flagged this we've discussed with the team whether we should show the same text ("Open all sections") visually as this would cognitively make more sense to sighted users who use screen readers, or whether we should not add the word "sections" at all as we don't have evidence that this particular pattern is helpful to users (and perhaps when the user encounters it they might not yet know what is meant by "sections" here). However as the "Check your answers" pattern (which does differ from the one used here) has tested well with users in the past, we're going to go with the current approach (ie. have "sections" as visually hidden) and add this information to "Known gaps and issues" in the guidance.

**Comment by stevenaproctor** [comment]
@hannalaakso I got my screen readers mixed up. I spoke to a colleague and they said it is a bug in VoiceOver on MacOS 10.12 and above. Does that help?

**Comment by nickcolley** [comment]
@stevenaproctor can you find out which browser and version they are using please? 👍

**Comment by stevenaproctor** [comment]
@nickcolley I will let you know as soon as I know.

**Comment by stevenaproctor** [comment]
@nickcolley It is not tied into any version of Safari, but the operating system itself as VoiceOver is embedded. The bug can be found in Mac OS 10.12 (Sierra) and above.

**Comment by nickcolley** [comment]
@stevenaproctor since that pattern is not new to this component I've raised a separate issue to investigate that, but it should not block this component using that pattern: https://github.com/alphagov/govuk-frontend/issues/1096

**Comment by timpaul** [comment]
accordion

Just noticed that the mouse doesn't switch to the hand icon when you hover over the heading.

File: src/components/accordion/_accordion.scss [hunk 15]
```diff
@@ -140,15 +141,31 @@
...
+      content: "";
+      position: absolute;
+      top: 0;
+      right: 0;
+      bottom: 0;
+      left: 0;
+      width: 25%;
+      height: 25%;
+      margin: auto;
+      background-color: govuk-colour("black");
```
nickcolley: could this be `currentColor`? That'll mean that when users change their colours it will override as expected.

adamsilver:
I just tested the latest version in IE11 (in Virtual Box) and the icons don't show.

image

And when I tab, you can see that I have to press the `tab` key twice in order to focus the next accordion heading. This might be because the SVG needs the `focusable="false"` attribute in IE.

accordion-keyboard

aliuk2012:
@adamsilver I have just replaced the SVG icons with a CSS version. Does this address the issues you have raised as I've tried this in Win 8.1, IE11 in Browserstack and I'm not able to replicate the issues you have raised.

src/components/accordion/_accordion.scss
hannalaakso: Thanks @nickcolley. `currentColor` is not supported by IE8. I've just added in a [fix](m/alphagov/govuk-frontend/pull/958/commits/2d7a663b01cc160aa6d1111d26837640b9f77b64) which makes sure that the icon displays for users who change their colours.

src/all.js
```diff
@@ -12,6 +13,12 @@ function initAll () {
   // Find all buttons with [role=button] on the document to enhance.
   new Button(document).init()
 
+  // Find all global accordion components to enhance.
+  var $accordions = document.querySelectorAll('[data-module="accordion"]')
```
aliuk2012: Should we not be name spacing data attributes? eg. `data-govuk-module`? Perhaps something to look at as a separate PR

src/components/accordion/template.test.js
```diff
@@ -0,0 +1,154 @@
...
+            }
+          },
+          {
+            heading: {
+              'text': 'Section B'
+            },
+            content: {
+              text: 'More content'
+            }
+          }
```
aliuk2012: I don't think we need this extra section as we are not testing its state

src/all.js
hannalaakso: Yes good point. We don't currently namespace data attributes in any of our components but it's been discussed before. I've added a card to Dev Time to discuss again as a team.

36degrees:
`src/assets/images/icon-plus-minus.png` does not appear to be used.

hannalaakso:
Thanks @36degrees, that's removed now.

src/components/accordion/accordion.js
```diff
@@ -0,0 +1,268 @@
...
+        console.log('Notice: sessionStorage not available.')
+      }
+    }
+  }
+}
+
+// Set the state of the accordions in sessionStorage
+Accordion.prototype.storeState = function ($section) {
+  if (this.browserSupportsSessionStorage) {
+    // We need a unique way of identifying each content in the accordion. Since
```
36degrees: Is `window.sessionStorage` scoped to the current page, or is it across domain?

If it's shared across a domain, and the same IDs are reused across multiple pages (e.g. `#section1`) then open sections will persist across different pages (e.g. the first section might be open when you navigate to a page you haven't seen before) which I think would be unexpected?

hannalaakso: They are persisted across a domain. We discussed this [here](https://github.com/alphagov/govuk-frontend/pull/958#issuecomment-431290759) and decided that we should advise in guidance that the service team will need to use unique-across-service IDs for their accordion items - we didn't feel that the onus of generating these with JS should be with the component code. 

It's a really good point since the guidance doesn't currently address this - this is quite a "code-y" point so I'm not sure whether the Design System component page is the best place for this guidance, maybe it should live in the component README 🤔

36degrees: 👍

hannalaakso: I've added an explanation to the [component options](https://github.com/alphagov/govuk-frontend/blob/bfb58c04ed1a7f9c368c75da16f1e8a06509d1d7/src/components/accordion/accordion.yaml#L5).

src/components/accordion/accordion.js
```diff
@@ -0,0 +1,268 @@
...
+  var expandedSectionCount = this.$module.querySelectorAll('.' + this.sectionExpandedClass).length
+  var areAllSectionsOpen = sectionsCount === expandedSectionCount
+
+  return areAllSectionsOpen
+}
+
+// Update "Open all" button
+Accordion.prototype.updateOpenAllButton = function (expanded) {
+  var newButtonText = expanded ? 'Close all' : 'Open all'
+  newButtonText += '<span class="govuk-visually-hidden"> sections</span>'
```
aliuk2012: Non-blocking suggestion,
These two lines and line 60 look like they could do with a function that returns the mark up.

Also if we wanted to add translation support to the component we could use the macro to determine the text and set a html data attribute which this js could pickup and use.

frankieroberto:
🎊

Issue #674, opened 2018-04-26
Review application can mislead if a component is polyfilled


nickcolley:
Our review application uses the 'all' JavaScript bundle.

If a polyfill is defined in one component it can also solve issues in another component, without explicitly importing them.

We could consider only initialising components on their respective pages.

This ties into https://github.com/alphagov/govuk-frontend/issues/660

PR #959, opened 2018-08-17
Add character count component

alex-ju:
This PR is a proposal for integrating the [character count component](https://github.com/alphagov/ds-character-count) to govuk-frontend.

#### Threshold value
- defined via `data-threshold="75"` (recommended) ~~or `.init( { threshold: 75 } )`~~
- if unset defaults to 0 — the message is displayed from the beginning
- if set to a certain value (e.g. 75) the message is displayed after 75% of the maximum limit is reached

#### Word count
- defined via `data-maxwords="200"` (recommended) ~~or `.init( { maxwords: 200 } )`~~
- if unset defaults to counting characters
- if set counts words instead of characters

#### Hint and Error message
- using the textarea component

### Dropped optional features
- highlight option (needs more user research and limits the resize of the textarea)
- validation border option (the validation border is part of the Error message component and is applied to the Character count component if an error occurs after the page is submitted)

### Known issues
JAWS repeats the label of the textarea each time an aria-live associated with the textarea is updated.

[Design System backlog](https://github.com/alphagov/govuk-design-system-backlog/issues/67)
[Trello card](https://trello.com/c/yn8X9iXN)


src/components/character-count/character-count.js
```diff
@@ -0,0 +1,226 @@
...
+    countMessage.classList.add('govuk-character-count__message--disabled')
+  } else {
+    countMessage.classList.remove('govuk-character-count__message--disabled')
+  }
+
+  // Update styles
+  if (remainingNumber < 0) {
+    countElement.classList.add('govuk-textarea--error')
+    if (options.validation) {
+      countElement.parentNode.parentNode.classList.add('govuk-form-group--error')
```
hannalaakso: I tested this locally be setting `validation: true` and the form group error class gets set on the wrong div so that just needs changing. Would it be possible to do this more robustly by looking for the form-group class and adding the class to that?

src/components/character-count/character-count.js
```diff
@@ -0,0 +1,226 @@
...
+  if (!$textarea) {
+    return
+  }
+
+  // Read options set using dataset ('data-' values) and merge them with options
+  var dataset = this.getDataset($module)
+  this.options = this.merge(options || {}, dataset)
+
+  // Determine the limit attribute (characters or words)
+  var countAttribute = (this.options.maxwords) ? this.defaults.wordCountAttribute : this.defaults.characterCountAttribute
```
hannalaakso: Hmmm we also allow setting `data-maxwords` in `template.njk`. What do you think about passing through all the `options` as data attributes from the markup instead?

alex-ju: You can pass any attribute/option via data attributes or options - a merge happens on line 25.
Updated the PR description to reflect that.

src/components/character-count/character-count.js
```diff
@@ -0,0 +1,226 @@
...
+
+  // Generate and reference message
+  var boundCreateCountMessage = this.createCountMessage.bind(this)
+  this.countMessage = boundCreateCountMessage()
+
+  // If there's a maximum length defined and the count message was successfully applied
+  if (this.maxLength && this.countMessage) {
+    // Replace maxlength attribute with data-maxlength to avoid hard limit
+    $module.setAttribute('maxlength', '')
+    $module.setAttribute('data-maxlength', $module.maxLength)
```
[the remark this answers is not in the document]
alex-ju: I'm skeptic about moving these to the textarea instead because in my opinion they are being used at the component level not at the textarea element level. Also we should keep it flexible and easier to extend in the future so that the field could be an input or a textarea.

src/components/character-count/character-count.js
hannalaakso: Thanks Alex. Is there an advantage to allowing user to do one or the other? I'm thinking that in terms of writing guidance it might be clearer if we just stick to one approach.

alex-ju: I would recommend the data attribute as it has more visibility for developers. Init could be missed or placed in a more obscure place inside the project.

hannalaakso: That sounds sensible 👍 Are you happy to amend the component so that options can no longer be passed through from `init`?

src/components/character-count/character-count.js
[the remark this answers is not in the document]
hannalaakso: My bad I assumed `maxlength` was set on `$textarea`. The maxlength value should be from the character count instance, not from the module as `$module.maxLength` is `undefined`. Think it should be:

```
$module.setAttribute('data-maxlength', this.maxLength)
```

src/components/character-count/character-count.njk
```diff
@@ -0,0 +1,10 @@
+{% from "character-count/macro.njk" import govukCharacterCount %}
+
+{{ govukCharacterCount({
+  id: "more-detail",
+  name: "more-detail",
+  label: {
+    text: "Can you provide more detail?",
+    hintText: "Don't include personal or financial information, eg your National Insurance number or credit card details."
+  }
+}) }}
```
hannalaakso: I don't think we have this file for other components. Does it help to demonstrate how you would use the component?

36degrees: Example files were removed in #890 – I imagine the work on this started before that change was made.

src/components/character-count/character-count.js
```diff
@@ -0,0 +1,192 @@
...
+
+  charVerb = (remainingNumber < 0) ? 'too many' : 'remaining'
+  displayNumber = Math.abs(remainingNumber)
+
+  countMessage.innerHTML = 'You have ' + displayNumber + ' ' + charNoun + ' ' + charVerb
+}
+
+CharacterCount.prototype.handleFocus = function () {
+  // Check if value changed on focus
+  this.valueChecker = setInterval(this.checkIfValueChanged.bind(this), 1000)
```
nickcolley: I wonder if we can take advantage of this possible alternative:

https://github.com/alphagov/accessible-autocomplete/issues/164#issuecomment-392875126

alex-ju: Good suggestion. I'll try.

nickcolley: Hmm, does not support IE8. But we could be OK with this since it'll likely only be used for users of assistive technologies that we can assume will be using a newer version of IE?

src/components/character-count/_character-count.scss
```diff
@@ -0,0 +1,28 @@
...
+@import "../../tools/all";
+@import "../../helpers/all";
+
+@import "../error-message/error-message";
+@import "../hint/hint";
+@import "../label/label";
+
+@include govuk-exports("govuk/component/character-count") {
+  .govuk-character-count {
+    .govuk-textarea {
```
nickcolley: Is this safe? We're relying on the textarea here.

I wonder if `.govuk-character-count__textarea` would be better, but I assume these overrides only make sense when this component is interacting with the other one's styles...

src/components/character-count/character-count.js
alex-ju: My personal choice would be to allow both and recommend one of them. Sometimes a dev may not have full access to how the markup is being rendered, in which case passing the attributes via javascript might come in handy.

src/components/character-count/character-count.njk
alex-ju: Wasn't aware. Will check and update.

src/components/character-count/character-count.js
alex-ju: Updated: not the options can only be passed via data attributes.

fofr:
A side note on this feature, might not be relevant to this PR directly: I think the logic that determines how many words are present should be surfaced somehow, as this will need to be reused in any server side validation so that both frontend and backend are consistent.

alex-ju:
Updated the component code based on the working group feedback. I think this is now good to be reviewed by the Design System team for integration

src/components/character-count/template.njk
```diff
@@ -0,0 +1,36 @@
...
+{% from "../hint/macro.njk" import govukHint %}
+{% from "../label/macro.njk" import govukLabel %}
+{% from "../textarea/macro.njk" import govukTextarea %}
+
+{#- a record of other elements that we need to associate with the input using
+   aria-describedby – for example hints or error messages -#}
+{% set describedBy = "" %}
+
+<div class="govuk-character-count" data-module="character-count"
+{% if params.maxlength %} data-maxlength="{{ params.maxlength }}"{% endif %}
```
nickcolley: Non-blocking, if we used the whitespace stripping syntax these would end up on the same line when the markup is rendered.

```html
<div class="govuk-character-count" data-module="character-count"
data-maxlength="blah"
></div>
```

vs


```html
<div class="govuk-character-count" data-module="character-count" data-maxlength="blah"></div>
```

src/components/character-count/character-count.js
```diff
@@ -0,0 +1,179 @@
...
+  }
+
+  // Read options set using dataset ('data-' values)
+  this.options = this.getDataset($module)
+
+  // Determine the limit attribute (characters or words)
+  var countAttribute = (this.options.maxwords) ? this.defaults.wordCountAttribute : this.defaults.characterCountAttribute
+
+  // Set the element limit
+  if ($module.getAttribute) {
```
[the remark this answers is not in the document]
alex-ju: Haha.. good catch. It was `$module.getAttribute(countAttribute)` to make sure an attribute is set and if not the script will stop, but got lost on it's way.. will update!

src/components/character-count/character-count.js
```diff
@@ -0,0 +1,179 @@
...
+  }
+
+  // Generate and reference message
+  var boundCreateCountMessage = this.createCountMessage.bind(this)
+  this.countMessage = boundCreateCountMessage()
+
+  // If there's a maximum length defined and the count message was successfully applied
+  if (this.maxLength && this.countMessage) {
+    // Replace maxlength attribute with data-maxlength to avoid hard limit
+    $module.setAttribute('maxlength', '')
```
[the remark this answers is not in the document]
joelanman: if this is what I think it is, it's relevant to something I was concerned about - if users don't use the macro they could well add maxlength to the texarea html tag themselves. We could remove it in code as it looks like we are doing here, and/or talk about this aspect in the guidance ('do not add 'maxlength' because ...)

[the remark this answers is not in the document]
alex-ju: Yes, just spoke about this with @joelanman. This is removing the hard limit if any set (needs a bit of refactor and I'll use `removeAttribute` instead of setting an empty string). How do you feel about this approach @nickcolley?

alex-ju:
@nickcolley thank you for reviewing this! update this PR with 4 commits addressing your comments.

src/components/character-count/character-count.js
```diff
@@ -0,0 +1,180 @@
...
+  if (!this.maxLength) {
+    return
+  }
+
+  // Generate and reference message
+  var boundCreateCountMessage = this.createCountMessage.bind(this)
+  this.countMessage = boundCreateCountMessage()
+
+  // If there's a maximum length defined and the count message exists
+  if (this.maxLength && this.countMessage) {
```
hannalaakso: Wondering if it's necessary to check for `this.maxLength` here as this is already done on line 33.

alex-ju: Thanks, Hanna! Removed the maxLength as we're checking now a few lines above.

alex-ju:
Rebased to master again 🤞

src/components/character-count/character-count.js
alex-ju: Tried it and behaves as expected (starts on focus, stops on blur). Could do a second test to be sure.

src/components/character-count/character-count.js
```diff
@@ -0,0 +1,186 @@
...
+
+// Generate count message and bind it to the input
+// returns reference to the generated element
+CharacterCount.prototype.createCountMessage = function () {
+  var countElement = this.$textarea
+  var elementId = countElement.id
+  // Check for existing info count message
+  var countMessage = document.getElementById(elementId + '-info')
+  // If there is no existing info count message we add one right after the field
+  if (elementId && !countMessage) {
```
nickcolley: I think we no longer need to create this count message, since we have a progressively enhanced way of doing this.

I don't think this is blocking but thought it was worth mentioning since it could simplify a lot of logic.

src/components/character-count/character-count.js
```diff
@@ -0,0 +1,186 @@
...
+  // Determine the remaining number of characters/words
+  var currentLength = this.count(countElement.value)
+  var maxLength = this.maxLength
+  var remainingNumber = maxLength - currentLength
+
+  // Set threshold if presented in options
+  var thresholdPercent = options.threshold ? options.threshold : 0
+  var thresholdValue = maxLength * thresholdPercent / 100
+  if (thresholdValue > currentLength) {
+    countMessage.classList.add('govuk-character-count__message--disabled')
```
nickcolley: Does this need to be polyfilled? (Maybe related to https://github.com/alphagov/govuk-frontend/issues/674)

src/components/character-count/character-count.js
nickcolley: I'll raise an issue after we merge this to investigate the approach defined in that issue.

src/components/character-count/template.njk
```diff
@@ -0,0 +1,36 @@
+{% from "../error-message/macro.njk" import govukErrorMessage -%}
+{% from "../hint/macro.njk" import govukHint %}
+{% from "../label/macro.njk" import govukLabel %}
+{% from "../textarea/macro.njk" import govukTextarea %}
+
+{#- a record of other elements that we need to associate with the input using
+   aria-describedby – for example hints or error messages -#}
+{% set describedBy = "" %}
```
nickcolley: I'm not following how describedBy gets set to something other than '', where does that happen?

nickcolley:
@alex-ju I've added some more review comments, I think the only one that is blocking is the polyfill one if we do need to polyfill classList

src/components/character-count/character-count.js
alex-ju: You're right about this, we can now remove the createCountMessage function and the associated logic; what I'm questioning thought is what will happen if they forget to add the default message. With the macro we can ensure that, but without it may get lost. I'm up for simplifying the code if we think people will add the default message.

src/components/character-count/character-count.js
alex-ju: Added the `classList` polyfill

src/components/character-count/template.njk
alex-ju: I added the id of the default message as a default value for describedBy - which can be amended with hint and error message ids in the textarea component.

src/components/character-count/character-count.js
nickcolley: Yeah I suppose this would force people to do a *more* progressive enhancement approach, but even as is I think it's still good to go.

nickcolley:
This is excellent work, thanks so much Alex this will helps lots of teams across GOV. I love the care that has gone into the research and efforts to make this component work well in assistive technologies like Dragon. 👏 

If we can release this today we might be able to publish it today 🚀 !

PR #960, opened 2018-08-20
Update select component width

dashouse:
This PR:

Creates a `govuk-!-width-full` override (previously we only had width to reduce the size of 100% elements

Removes the `width: 100%` declaration on the select so they are the size of the content by default

**Why**

The styling of the `<select>` component (dependant on browser) is very similar to a text input. The arrows on the right hand side indicate the behaviour of this component, however because we made it 100% wide by default, a select with a small amount of content, when zoomed in is not obvious.

For example:
screen shot 2018-08-20 at 12 00 17

This change reverts to the default `<select>` behaviour, being sized by it's content so the icon appears closer to the content.

screen shot 2018-08-20 at 12 01 31

If a user would like to resize their select, for example they are sizing it based on being 100% wide within a container this can be done with the override class

screen shot 2018-08-20 at 12 01 37

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -121,6 +121,12 @@
 
   ([PR #985](https://github.com/alphagov/govuk-frontend/pull/985))
 
+- Remove the width declaration from the `<select>` component
+
+  The `<select>` component’s width will now be defined by it’s content. This addresses some accessibility issues with the select being 100% wide by default. If you want to style your select to be 100% wide we have added a new override class `.govuk-!-width-full` to allow this.
+
+  ([PR #960](https://github.com/alphagov/govuk-frontend/pull/960))  
+
 - Pull Request Title goes here
 
   Description goes here (optional)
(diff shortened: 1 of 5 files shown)


dashouse:
Cool, I will do the changelog at the last minute if approved for minimal merge conflicts :)

Will discuss in channel about breaking change of not

nickcolley:
@dashouse can you please update the PR summary to explain what problem you're solving?

dashouse:
@nickcolley Hopefully this covers it...

CHANGELOG.md
```diff
@@ -72,8 +72,20 @@
 
   ([PR #969](https://github.com/alphagov/govuk-frontend/pull/969))
 
+- Remove the width declaration from the `<select>` component
+
+  The `<select>` component’s width will now be defined by it’s content. This addresses some accessibility issues with the select being 100% wide by default. If you want to style your select to be 100% wide we have added a new override class to allow this.
```
nickcolley: Non-blocking it would be good to explicitly reference the new class that can be used, not the end of the world though.

dashouse:
Updated changelog and merged, thanks

PR #961, opened 2018-08-21
Use production URL rather than cloudapps

fofr:
A small URL update to prevent an unnecessary redirect for users.

https://govuk-design-system-production.cloudapps.digital/ redirects to https://design-system.service.gov.uk/

Not sure if this warrants a changelog entry or not.


igloosi:
thanks, Paul.

ideally we would modify the link in 'src/components/input/README.njk' and regenerate the markdown file with `npm start` We can ignore the markdown file  in /packages as they get copied over in the release process.

Issue #896, opened 2018-07-13
Add unit tests for 'value' option in the date input


nickcolley:
# First Timers Only
This issue is reserved for people who never contributed to the GOV.UK Design System before, please only pick this up if you're a new contributor.

## Description
The [date input component](https://design-system.service.gov.uk/components/date-input/) has a [Nunjucks template](https://mozilla.github.io/nunjucks/) which allows us to output the HTML depending on what information is passed to it.

To ensure that this template is outputting the correct values we write automated tests using [Jest](https://jestjs.io/)

We are not testing that when a user passes the `value` property to a date input item.

This feature is often used to replay the users' input back to them, for example if they made a mistake.

```javascript
{{ govukDateInput({
  items: [
    {
      name: "day"
    },
    {
      name: "month"
    },
    {
      name: "year",
      value: "2018"
    }
  ]
}) }}
```

This code exists in the component's directory and is called [`template.njk`](https://github.com/alphagov/govuk-frontend/blob/f52e56a3de96457afb9b4497133ae365857027cf/src/components/date-input/template.njk#L85).

### 1. Claim this issue

Comment below to let other people know that you're working on this issue.

### 3. Add a new test

Update the file `src/components/date-input/template.test.js` with an additional test to check that the template `renders items with values`.

You can copy a previous example as a basis for this: https://github.com/alphagov/govuk-frontend/blob/f52e56a3de96457afb9b4497133ae365857027cf/src/components/date-input/template.test.js#L201-L218

To run this test you'll need to be able to [run the project locally](https://github.com/alphagov/govuk-frontend/blob/master/docs/contributing/running-locally.md) then run:

```bash
npm test
```

### 4. [Create a Pull Request](https://help.github.com/articles/creating-a-pull-request/)

Since this is an internal change, we do not need to add a changelog entry.

### 5. Done

We'll review the Pull Request and let you know if any changes are needed.

## Questions

If you’ve got a question, you can comment in this issue or email the Design System team on [govuk-design-system-support@digital.cabinet-office.gov.uk](mailto:govuk-design-system-support@digital.cabinet-office.gov.uk)

malydok:
Sorry for not claiming the task first, but I wasn't sure I'd be able to do it before actually doing it.
Here's the submitted PR https://github.com/alphagov/govuk-frontend/pull/965

PR #965, opened 2018-08-27
Add unit tests for 'value' option in the date input

malydok:
This is my first ever contribution to OSS, please be gentle.

Solution for https://github.com/alphagov/govuk-frontend/issues/896

Side questions:
1. is giving the PR the same name as the issue okay?
2. could I improve my commit message and branch name?

Thanks!

### DIFF
--- a/src/components/date-input/template.test.js
+++ b/src/components/date-input/template.test.js
@@ -237,6 +237,26 @@ describe('Date input', () => {
       expect($firstItems.attr('id')).toEqual('my-date-input-day')
     })
 
+    it('renders items with value', () => {
+      const $ = render('date-input', {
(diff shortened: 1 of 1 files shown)


igloosi:
Hi @malydok thank you for your contributions.

This looks good, but it does need a second review from a team member.

To answer you questions.
In this case the name of the PR is fine. Usually you'd give a title that represents the work done in the PR.
Branch names usually don't matter too much. What matters is the commit title and the commit message. These get retained in the history when the PR gets merged so it's useful to have a good explanation in there what the work was about.

More details on our commit style guide https://github.com/alphagov/styleguides/blob/master/git.md

Issue #864, opened 2018-07-04
Set panel component's title heading level to 1 as default


igloosi:
Default used to be `h2`, but in most cases it should be `h1` with rare exceptions.
In #853 we made it customisable, while leaving the default as `h2`.

Setting `h1` as default will be a breaking change

PR #967, opened 2018-08-29
Set panel component's title heading level to 1 as default

igloosi:
We think the panel title should be the h1 in the majority of cases.

This PR:
- updates the Panel component template
- updates the component tests
- updates the example YAMl file
- generates and updated README

Trello ticket: https://trello.com/c/CA0fql0w/1388-set-panel-components-title-heading-level-to-1-as-default

Fixes: https://github.com/alphagov/govuk-frontend/issues/864

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,13 +4,16 @@
 
 💥 Breaking changes:
 
-- Pull Request Title goes here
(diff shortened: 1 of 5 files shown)


36degrees:
Merged in @igloosi's absence.

Issue #894, opened 2018-07-13
Remove deprecated `govuk-visually-hidden-focussable` class name


nickcolley:
https://github.com/alphagov/govuk-frontend/blob/61d48a50a01ea5a30b59d0479dd07c37d76768e3/src/utilities/_visually-hidden.scss#L6-L13

PR #968, opened 2018-08-30
Remove deprecated `govuk-visually-hidden-focussable` class name

igloosi:
This is to remove the old, misspelt CSS class.
If users are still using this class in their codebase, they need to replace it with `.govuk-visually-hidden-focusable`

Fixes: #894 

Trello ticket: https://trello.com/c/gRLqoS4h/1390-1-remove-deprecated-govuk-visually-hidden-focussable-class-name

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,14 +6,26 @@
 
 - Set panel component's title heading level to 1 as default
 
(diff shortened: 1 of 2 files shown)


36degrees:
Merging in @igloosi's absence.

Issue #754, opened 2018-06-04
Conditional revealed content with inline radio buttons does not render correctly


gavinwye:
When using the macro for a conditionally revealing element under a radio button when the buttons are displayed inline the revel element should appear under the radio buttons as shown in [elements](http://govuk-elements.herokuapp.com/form-elements/example-form-elements).

![image](https://user-images.githubusercontent.com/868772/40907323-c23907ea-67e3-11e8-95b5-c42f7e10954e.png)

36degrees:
@adamsilver raised a good point about this pattern - when Javascript is not available the revealed thing that relates to Yes, for example, will be under No and out of context, and if there’s something to reveal for both Yes and No then similar problems arise.

This appears to be an issue with the current implementation in Elements - with Javascript enabled, the fields are associated with the radios because only one is visible at a time:

![jun-04-2018 09-55-50](https://user-images.githubusercontent.com/121939/40908018-81c81904-67dd-11e8-8e3c-3ebd4172068d.gif)

Whereas without Javascript, both revealed fields are visible at the same time and there is nothing to associate each field with the corresponding radio.
<img width="850" alt="screen shot 2018-06-04 at 09 46 27" src="https://user-images.githubusercontent.com/121939/40907938-3a6e3002-67dd-11e8-9ae6-22147cba2853.png">

One suggestion is that we could somehow make it so that the 'inline' modifier is not applied when javascript is not available and there is conditionally revealing content, meaning that 'Yes' and 'No' would appear stacked with the revealed content underneath each radio button.

timpaul:
Thanks for spotting this @gavinwye - we'll investigate to see what we can do to improve the layout of that use case.

In the meantime we'd recommend either not using the inline layout in combination with conditionally revealed content or moving the conditional content onto another screen. For the reason @adamsilver gave, these might be preferable approaches anyway.

adamsilver:
Perhaps there could be some guidance to say don't use conditionals with inline radios.

adamsilver:
Another little thing that we might want to consider is that if you select NO, the left border lines up with YES.

abbott567:
We have an interesting use case. The service I'm working on is for staff. As we all know, staff hate waiting for anything to load as rubbish equipment and network latency already slow everything down. So we use conditional reveals in a few places rather than loading different views.

[...]

PR #970, opened 2018-08-30
Address issue with conditional form content and inline form controls

igloosi:
While we don't recommend this combination, we need to address the broken styling.
This work forces 'inline' form controls to become 'block' when used in conjunction with conditional content when created with a Nunjucks macro.

Related to: https://github.com/alphagov/govuk-frontend/issues/754
Trello ticket: https://trello.com/c/bVn1hU8i/1311-2-fix-issue-with-conditional-form-content-and-inline-form-controls

**Before:** 
screen shot 2018-08-30 at 12 27 24

**After:**
screen shot 2018-09-20 at 10 30 56

### DIFF
--- a/src/components/radios/_radios.scss
+++ b/src/components/radios/_radios.scss
@@ -143,6 +143,14 @@
         clear: none;
       }
     }
+
+    // Prevent inline modifier being used with conditional reveals
+    &.govuk-radios--conditional {
+      .govuk-radios__item {
+        margin-right: 0;
+        float: none;
+      }
+    }
   }
 
   .govuk-radios__divider {
--- a/src/components/radios/template.njk
+++ b/src/components/radios/template.njk
@@ -41,7 +41,7 @@
     text: params.errorMessage.text
   }) | indent(2) | trim }}
 {% endif %}
-  <div class="govuk-radios {%- if params.classes %} {{ params.classes }}{% endif %}"
+  <div class="govuk-radios {%- if params.classes %} {{ params.classes }}{% endif %}{%- if isConditional %} govuk-radios--conditional{% endif -%}"
     {%- for attribute, value in params.attributes %} {{ attribute }}="{{ value }}"{% endfor %}
     {%- if isConditional %} data-module="radios"{% endif -%}>
     {% for item in params.items %}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,14 @@
   You can now pass additional attributes to links in header, footer, breadcrumbs, tabs and error-summary components
 
   ([PR #993](https://github.com/alphagov/govuk-frontend/pull/993))
+  
+- Fix issue with conditional form content and inline form controls
+
+  When inline variant of form controls is used with conditional content, we force
+  it to display block. This is to avoid breaking and confusing styling as it is
+  a combination we don't recommend.
+
+  ([PR #970](https://github.com/alphagov/govuk-frontend/pull/970))
 
 - Pull Request Title goes here
 
--- a/src/components/radios/radios.yaml
+++ b/src/components/radios/radios.yaml
@@ -234,6 +234,36 @@ examples:
             <label class="govuk-label" for="contact-text-message">Mobile phone number</label>
             <input class="govuk-input govuk-!-width-one-third" name="contact-text-message" type="text" id="contact-text-message">
 
+- name: inline with conditional items
+  readme: false
+  data:
+    classes: govuk-radios--inline
+    idPrefix: how-contacted
+    name: how-contacted
+    fieldset:
+      legend:
+        text: How do you want to be contacted?
+    items:
+      - value: email
+        text: Email
+        conditional:
+          html: |
+            <label class="govuk-label" for="context-email">Email address</label>
+            <input class="govuk-input govuk-!-width-one-third" name="context-email" type="text" id="context-email">
+      - value: phone
+        text: Phone
+        conditional:
+          html: |
+            <label class="govuk-label" for="contact-phone">Phone number</label>
+            <input class="govuk-input govuk-!-width-one-third" name="contact-phone" type="text" id="contact-phone">
+      - value: text
+        text: Text message
+        conditional:
+          html: |
+            <label class="govuk-label" for="contact-text-message">Mobile phone number</label>
+            <input class="govuk-input govuk-!-width-one-third" name="contact-text-message" type="text" id="contact-text-message">
+
+
 - name: with conditional item checked
   readme: false
   data:


36degrees:
Could we possibly add an example to the review app so that we have more chance of catching any issues with this combination in the future?

36degrees:
This still isn't working quite as I'd expect it to…

inline-conditional

igloosi:
@dashouse how do we expect conditional content with inline forms to behave?

dashouse:
This ticket only was labelled as build and bug so I hadn't looked at it properly

If we want to support this behaviour I'd want to spend some time designing it first, I'm ok with the glitch in the mean time as it means it won't be used.

36degrees:
We're limited somewhat by the source order here – the macro will always put the conditional HTML after the input, so the revealed content will always be 'between' the radios.

Arguably this is still an improvement over what was there before… I guess we need to work out what an 'acceptable' render is for a combination we don't really recommend and don't really want to support.

dashouse:
I think we need to consider if an unacceptable render is better in this instance?

If we zoom out slightly we may want to support conditionally revealing *something* based on *something else* outside of the individual component macro?

igloosi:
considering the comments above I'm tempted to close this as it won't be fine by the end of the sprint

igloosi:
the only way to fix this bug without a css hack is to do the work that toggles named IDs and them move conditional content in a block after all radio items.

nickcolley:
For what it's worth doing, allows you to do this without breaking the component out the flow of the document.

```css
.radio-group {
  display: flex;
  flex-wrap: wrap;
}

.condtional {
  width: 100%;
  order: 1;
}
```

But using `order` can have serious consequences to accessibility if not properly thought about, since it creates a mismatch between the markup ordering and the presentational ordering.

nickcolley:
The problem with using absolute positioning is that it does not take up space in the flow of the document, which means anything underneath it will occupy the space:

screen shot 2018-09-12 at 15 32 20

igloosi:
yeah, absolute positioning and flex and down-sides.

Last resort is to force inline radios  radios items to block if they're used with conditional content.

nickcolley:
With flexbox, if we were happy with the reordering, it would work. With older browsers falling back to everything stacked...

I think we should see what @dashouse thinks, since this seems like more of a 'should we even do this' rather than a technical problem.

igloosi:
done with flexbox as suggested. 
order: 2 only applied to the visible conditional item, so the tab order remains the same

dashouse:
This is truly worthy effort to try and make this render well but I am still on the site of not allowing this functionality as part of the radios and checkboxes components.

## Differences between GOV.UK Elements and GOV.UK Frontend

Elements re-used the "Inset text" (.panel) component for every type of reveal. The border was aligned to the far left like so

screen shot 2018-09-19 at 11 39 56

In GOV.UK Frontend each of the components had the conditional reveal style as part of the component itself. To create a better ownership between the radio / checkbox and the revealed element the border has been aligned to the centre of the radio / checkbox. This behaviour will follow down for multiple levels of reveals if desired.

screen shot 2018-09-19 at 11 41 11

### Current proposal

The current proposal in the PR, when mixed with these new styles could lead to some usability issues. For example, if a user selects the second radio button, the border is linked to the first radio button.

screen shot 2018-09-19 at 11 26 46

In this scenario I don't think this design works, the behaviour is actually a lot more likened to "Tabs" with radio controls. To give feedback on which radio the conditionally revealed content relates to a standard pattern is something more like this:

screen shot 2018-09-19 at 11 35 51

#### Mobile

When inline radios with conditional reveals wrap the ownership of the revealed content is lost as it now aligns with the first item to wrap.

screen shot 2018-09-19 at 11 26 35

#### No JavaScript

The vertical radios / checkboxes with conditional reveal still make a lot of sense without JS available:

screen shot 2018-09-19 at 11 53 21

Where as the inline ones will need a bit of work

screen shot 2018-09-19 at 11 37 43

### Meeting the needs

I think in the majority of cases we should recommend not using conditional reveals with inline radios or checkboxes.

However, I think allowing the user to reveal "some content" based on another interaction outside of the component is a valid use case, but not part of individual components themselves. Something a long the lines of a JavaScript utility.

igloosi:
nice one @dashouse 
while we can make inline conditional no-js styles work, we would require a redesign of the active state and the mobile treatment

shall we close this and mark as won't fix + add guidance to the Design system?

src/components/radios/template.njk
```diff
@@ -18,6 +18,11 @@
   {% endif %}
 {% endfor %}
 
+{% set userRefinedClasses = ' ' + params.classes if params.classes %}
```
nickcolley: From a user's perspective if they're using standard HTML this feels a bit weird.

1. start off with block level inputs
2. then want them inline so they add an inline modifier
3. then they run into this weird rendering so add additional modifier which essentially just undos 2.?

igloosi: if they set inline and get broken styling they might
a) clear floats in css directly 
b) add modifier and add styles
c) just use block

Not sure what we can do for them.

src/components/checkboxes/template.njk
```diff
@@ -18,6 +18,11 @@
   {% endif %}
 {% endfor %}
 
+{% set userDefinedClasses = ' ' + params.classes if params.classes %}
```
36degrees: Checkboxes don't currently support the inline modifier, only radios do.

src/components/radios/template.njk
```diff
@@ -18,6 +18,11 @@
   {% endif %}
 {% endfor %}
 
+{% set userDefinedClasses = ' ' + params.classes if (params.classes) %}
```
36degrees: Given we [already know whether there's a conditional or not](https://github.com/alphagov/govuk-frontend/pull/970/files#diff-5eee27b7e0c518d51503bac6513ad0edR17), what do we think about adding another class (e.g. `govuk-radios--conditional`) to the wrapper if `isConditional` is true?

We don't need to know if they're using the inline modifier or not, because that class wouldn't do anything by itself, which would allow us to simplify this a lot.

We could then just do something like…

```scss
  // Inline variant
  .govuk-radios--inline {
    @include mq ($from: tablet) {
      @include govuk-clearfix;

      .govuk-radios__item {
        margin-right: govuk-spacing(4);
        float: left;
        clear: none;
      }
    }
  }

  // Prevent inline modifier being used with conditional reveals
  .govuk-radios--conditional.govuk-radios--inline {
    .govuk-radios__item {
      float: none;
    }
  }
```

(Potentially we could just use `:not` if we make the inline modifier IE9+ only, and fall back in IE8?)

Thoughts?

36degrees: Alternatively, if we want a more rigorous approach that would work for users who aren't using the macros, we could style based on `data-module="radios"` 

```
  [data-module="radios"].govuk-radios--inline {
    .govuk-radios__item {
      float: none;
    }
  }
```

It means we're styling based on a data attribute, which is a bit odd… but might be worth considering in this case.

nickcolley: This seems like a better approach since it won't force us to document something we don't want people to do.

nickcolley: FWIW the data module attribute is not a requirement for our JavaScript since you can initialise based on any selector, so by doing that 2nd approach we'd force people to use it.

So I'm in favour of your first proposal

igloosi: ok, let me implement the first proposal

src/components/checkboxes/template.njk
igloosi: thanks, removed

src/components/radios/template.njk
igloosi: implemented proposed solution

src/components/radios/_radios.scss
```diff
@@ -143,6 +143,13 @@
         clear: none;
       }
     }
+
+    // Prevent inline modifier being used with conditional reveals
+    &.govuk-radios--conditional {
+      .govuk-radios__item {
+        float: none;
```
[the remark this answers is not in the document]
igloosi: it's minor, but i've reset the margin as well

36degrees:
🚢

abbott567:
Hey @dashouse these are all really good points, but it still doesn't address the issue I had in #754

The service I'm working on is for staff. Our research shows that poor equipment and network latency mean that 1 thing per page is not a good solution because they spend more time waiting for a page to load than completing the task on it. But, 1 'theme' per page works fine. So for example, we group fields such as first name, last name, date of birth under one theme page for claimant details.

So, with this in mind we have an issue when tried to move over to using the design system component. The whole design which tested great in research, now no longer works. The way the component is built always jams the revealed content into the dom after its selector pushing the remaining inline buttons down. (see screenshot).

The recommendation of not using conditional reveals on inline-radios is fine when it's one field and one thing per page. But if it's a collection of fields it pushes the remaining radios to the bottom where they lose context. 

We are left with a single 'No' radio right at the bottom above the green button which makes no sense as the user has now forgot the question they answered originally. It also makes a screen reader just read out 'No, radio button 1 of 2' after you've finished entering all of the information, which again has lost it's context. 

So, whilst I agree with all of the points in your original post. Ignoring the research and user needs of services that may already be using the component as it was originally implemented, feels like it might be short sighted. People are probably going to make work arounds, because we shouldnt be redesigning something that tested well to fit in with the techhnical restrictions of the new component. 

Instead, people are just going to hack the component, and that will will create accessibility issues or an inconsistent experience for the end user.

image

dashouse:
Hey @abbott567, sorry for the delayed response.

I wouldn't suggest changing your design, you should stick with it based on your research. However from a component design perspective, I think we would be introducing too many potential usability issues into the default behaviour of conditional radios...unless we spent some time specifically looking at conditional reveal styling of horizontal radios to address it.

There wasn't really an original implementation of this component as GOV.UK Elements only had a generic helper (that could be used with anything), and the design suggestion was to put the revealed items inside of an inset text (.panel) component which you could put this wherever you wanted to put it in the HTML. There wasn't really a component relationship between what was being revealed and the component revealing it as there is now with GOV.UK Frontend.

Based on the positioning of the reveal in your screenshots this looks like what you've done, which is perfectly fine as you have a specific use case where it makes sense. However I think allowing this as an option of the radios component as the design stands could affect the comprehension of what radio button "owns" the revealed content. Have you tried using the default conditional reveal of the radios yet?

It would be really good to understand more about the user losing context of the original question as this is our general approach to conditional reveals and hadn't considered changing this based on the amount of content within the reveal. As you have pointed out, normally in the this scenario we would break this question up over multiple pages and we have made the assumption that if the user has revealed this content themselves that they retain that understanding of the original question.

Out of interest, in your specific scenario does the user trigger the conditional reveal or does it start with "Yes" pre-selected already (as that might be the most common scenario for your user)? I imagine that would certainly affect the context of the "No" option.

PR #977, opened 2018-09-04
Ability to add attributes to select option

malcolmhire:
Within our service, we need the ability to add data attributes to a select option, I've used a very similar style to how radio and checkboxes were attributes were added by @colinrotherham to keep it simple.


src/components/select/select.yaml
```diff
@@ -17,6 +17,12 @@ examples:
         value: 3
         text: GOV.UK frontend option 3
         disabled: true
+      -
```
36degrees: Generally we're trying to only add to these examples when there's a visual difference that we need to be able to eyeball in the review app – we can omit this because we've got tests that cover this functionality.

It'd be good if we could remove this addition, and revert the corresponding changes to the README.

(sorry that this isn't documented anywhere – we must make this clearer)

src/components/select/template.test.js
```diff
@@ -151,6 +151,40 @@ describe('Select', () => {
     })
   })
 
+  describe('when they include option attributes', () => {
+    it('it renders the option attributes', () => {
```
36degrees: Minor – duplicated 'it', don't need to include it as part of the test name.

malcolmhire: Removed the duplicate, thanks.

36degrees:
@malcolmhire sorry, one more thing – would you be able to [add a entry for this to the changelog](https://github.com/alphagov/govuk-frontend/blob/master/CONTRIBUTING.md#updating-changelog)?

36degrees:
Thanks Malcolm – unfortunately there's a conflict because the CHANGELOG has changed on master. Would you be able to rebase against master?

(I am very happy to do this on your behalf if you'd prefer)

36degrees:
@nickcolley would you mind re-reviewing?

malcolmhire:
@36degrees yeah fine to squash my commits 🙂. No problem with help out, it helps me out in my project once this is released. Also working a very similar feature for the date component, so should that done sometime soon. 🙂

PR #981, opened 2018-09-06
[Do not merge] Turn off compatibility mode by default and turn on relative typography (rem) by default 

nickcolley:


src/all.test.js
```diff
@@ -149,4 +149,21 @@ describe('GOV.UK Frontend', () => {
 
     expect(functionCalls).toBeNull()
   })
+  describe('responsive typography', async () => {
+    it('is enabled by default', async () => {
```
36degrees: Rather than adding these, can we update the tests in `src/helpers/typography.test.js`? (We currently explicitly disable responsive mode, and then have a test for when it's enabled, but we should probably flip that logic?)

src/helpers/typography.test.js
```diff
@@ -273,8 +272,8 @@ describe('@mixin govuk-typography-responsive', () => {
     describe('and $important is set to true', () => {
       it('marks font size and line height as important', async () => {
         const sass = `
-          ${sassBootstrap}
           $govuk-typography-use-rem: false;
+          ${sassBootstrap}
```
nickcolley: I have moved these around to better demonstrate how our users would write their SCSS

nickcolley:
@36degrees updated the tests with your feedback

Issue #944, opened 2018-08-06
Details component fails WCAG AA standard for colour ratio whilst on focus


paroxp:
## What

I've been addressed by a user, regarding the contrast in our application where we're using the [GOV.UK Frontend - Details](https://design-system.service.gov.uk/components/details/) component.

The colours are:

Background: `#ffbf47`
Text: `#005ea5`
Text on hover: `#2b8cc4`

Running these past [snook](https://snook.ca/technical/colour_contrast/colour.html#fg=005ea5,bg=ffbf47) or [accessible colors](http://accessible-colors.com/) service, will incline the colours are failing the `AA` standard with a `Your contrast ratio: 4.07` warning.

Proposal from the user:

> Change the text colour to black.

I imagine you'd need a designer to look at it or perhaps have a small chat among yourselves. But perhaps, using the purple of your palette (`#2e358b`), would work fine for you? It would avoid the use of `darken()` function or setting it to black.

PR #982, opened 2018-09-06
Use text colour on focus for better contrast

nickcolley:
Closes https://github.com/alphagov/govuk-frontend/issues/944

This pull request updates the focus styles of links in GOV.UK Frontend to they pass WCAG contrast requirements.

You can compare https://govuk-frontend-review-pr-982.herokuapp.com/examples/links vs https://govuk-frontend-review.herokuapp.com/examples/links

### back link
| Before | After |
|-|-|
| image | image |

### breadcrumbs
| Before | After |
|-|-|
| image | image |

### details
| Before | After |
|-|-|
| image | image |

### skip link
| Before | After |
|-|-|
| image | image |

### header
| Before | After |
|-|-|
| screen shot 2018-09-06 at 15 07 20 | screen shot 2018-09-06 at 15 07 06 |

### govuk-link
| Before | After |
|-|-|
| govuk-frontend-review herokuapp com_examples_links | govuk-frontend-review-pr-982 herokuapp com_examples_links |

https://trello.com/c/XQPpwofN/1381-1-modify-focus-state-to-use-text-colour

### DIFF
--- a/src/components/details/_details.scss
+++ b/src/components/details/_details.scss
@@ -44,7 +44,7 @@
     outline-offset: -1px;
     // When focussed, the text colour needs to be darker to ensure that colour
     // contrast is still acceptable
-    color: $govuk-text-colour;
+    color: $govuk-focus-text-colour;
     background: $govuk-focus-colour;
   }
 
--- a/src/components/details/_details.scss
+++ b/src/components/details/_details.scss
@@ -42,6 +42,9 @@
     // -1px offset fixes gap between background and outline in Firefox
     outline: ($govuk-focus-width + 1px) solid $govuk-focus-colour;
     outline-offset: -1px;
+    // When focussed, the text colour needs to be darker to ensure that colour
+    // contrast is still acceptable
+    color: $govuk-text-colour;
     background: $govuk-focus-colour;
   }
 
--- a/src/components/error-summary/_error-summary.scss
+++ b/src/components/error-summary/_error-summary.scss
@@ -56,7 +56,7 @@
     // When focussed, the text colour needs to be darker to ensure that colour
     // contrast is still acceptable
     &:focus {
-      color: $govuk-text-colour;
+      color: $govuk-focus-text-colour;
     }
 
     // alphagov/govuk_template includes a specific a:link:focus selector
--- a/src/components/error-summary/_error-summary.scss
+++ b/src/components/error-summary/_error-summary.scss
@@ -49,10 +49,14 @@
     &:link,
     &:visited,
     &:hover,
-    &:active,
-    &:focus {
+    &:active {
       color: $govuk-error-colour;
-      text-decoration: underline;
+    }
+
+    // When focussed, the text colour needs to be darker to ensure that colour
+    // contrast is still acceptable
+    &:focus {
+      color: $govuk-text-colour;
     }
 
     // alphagov/govuk_template includes a specific a:link:focus selector
--- a/src/components/header/_header.scss
+++ b/src/components/header/_header.scss
@@ -76,8 +76,10 @@
       text-decoration: underline;
     }
 
+    // When focussed, the text colour needs to be darker to ensure that colour
+    // contrast is still acceptable
     &:focus {
-      color: govuk-colour("black");
+      color: $govuk-text-colour;
     }
 
     // alphagov/govuk_template includes a specific a:link:focus selector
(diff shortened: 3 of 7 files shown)


src/helpers/_links.scss
```diff
@@ -44,6 +44,12 @@
   &:active {
     color: $govuk-link-active-colour;
   }
+
+  // When focussed, the text colour needs to be darker to ensure that colour
+  // contrast is still acceptable
+  &:focus {
```
nickcolley: I tried changing the focusable mixin and that led to ordering problems since focus was not the last selector.

36degrees:
'Active' links in the header component are still showing as light blue on yellow:

screen shot 2018-09-06 at 13 44 06

36degrees:
> Through doing this I found that the .govuk-list core styles styles anchors, I have updated these to extend the govuk-link selector instead.
> 
> I have checked our documentation and we do not recommend our users do this, so maybe this should deprecated and removed instead, or in the future?

Yeah, we don't do it in any other component that might include links, so I agree we should be thinking about removing this.

src/helpers/_links.scss
36degrees: Do you think this would be easy to fix?

nickcolley: It just felt fragile having to move lots of code around to get the ordering right, we can explore it more, it would be nice to have it in the mixin.

I think one to do together if we decide to explore that.

nickcolley:
@36degrees I have added an additional commit to address your feedback about the header component

src/components/error-summary/_error-summary.scss
```diff
@@ -49,10 +49,14 @@
     &:link,
     &:visited,
     &:hover,
-    &:active,
-    &:focus {
+    &:active {
       color: $govuk-error-colour;
-      text-decoration: underline;
```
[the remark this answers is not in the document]
nickcolley: As far as I could tell, this did nothing?

src/components/header/_header.scss
```diff
@@ -76,8 +76,10 @@
       text-decoration: underline;
     }
 
+    // When focussed, the text colour needs to be darker to ensure that colour
+    // contrast is still acceptable
     &:focus {
-      color: govuk-colour("black");
+      color: $govuk-text-colour;
```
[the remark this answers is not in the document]
nickcolley: Yeah I like that, we should definitely do that.

src/components/details/_details.scss
```diff
@@ -42,6 +42,9 @@
     // -1px offset fixes gap between background and outline in Firefox
```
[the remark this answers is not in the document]
nickcolley: I think the width of the outline itself is different too... I think best for another pull request.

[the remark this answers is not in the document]
36degrees: 👍

src/components/header/_header.scss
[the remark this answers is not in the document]
nickcolley: Hmm, maybe we could do this later to be honest, I've done a commit but wonder if thinking about theming holistically with everything else would be better than adding it piecemeal.

src/settings/_colours-applied.scss
```diff
@@ -127,4 +136,4 @@ $govuk-link-hover-colour: govuk-colour("light-blue") !default;
 /// @type Colour
 /// @access public
 
-$govuk-link-active-colour: govuk-colour("light-blue") !default;
+$govuk-link-active-colour: govuk-colour("light-blue") !default;
```
36degrees: Minor, but looks like you've removed the new line from the end of this file.

src/settings/_colours-applied.scss
```diff
@@ -68,6 +68,15 @@ $govuk-secondary-text-colour: govuk-colour("grey-1") !default;
 
 $govuk-focus-colour: govuk-colour("yellow") !default;
 
+/// Focused text colour
+///
+/// Ensure that the contrast between the text and background colour passes
+/// WCAG Level AA contrast requirements.
```
36degrees: _Very_ minor, but there's a blank line between the description and any annotations elsewhere in this file, and I think generally through the codebase. Would be good to be consistent.

src/settings/_colours-applied.scss
36degrees: Looks like the linter's picked up on this as well, which I think is why the tests (and the review app) are now failing.

nickcolley: All hail the linter

nickcolley:
@36degrees updated as per your feedback, thanks Ollie.

nickcolley:
I think I've broken the 'no visited state', definitely have, so block merging until I resolve that.

nickcolley:
screen shot 2018-09-10 at 10 31 44

I managed to solve the issues by duplicating the pseudo selectors, which is unfortunate but I think is the simplest way for now, this is consistent with the other modifiers at least.

paroxp:
👏

Issue #909, opened 2018-07-16
Behaviour of date input's name argument is not clear


36degrees:
The `name` argument on the date component acts as a prefix for each input's `name`, but this behaviour is not clear or consistent with how we use `name` in other components.

We should consider either renaming it to e.g. `namePrefix` and updating the documentation accordingly, or removing it from the api entirely.

As raised by @davidjamesstone and discussed in https://github.com/alphagov/govuk-frontend/issues/903

PR #984, opened 2018-09-06
Rename `name` of date-input component to `namePrefix`

hannalaakso:
The `name` argument on the date-input component acts as a prefix for each input's name, but this behaviour is not clear or consistent with how we use `name` in other components.

This PR renames it to `namePrefix`. This is consistent with other components (like checkboxes and radios) which use the name `idPrefix` to explain similar functionality.

Fixes https://github.com/alphagov/govuk-frontend/issues/909

https://trello.com/c/CFN5bvev/1392-2-behaviour-of-date-inputs-name-argument-is-not-clear

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,6 +72,14 @@
 
   ([PR #969](https://github.com/alphagov/govuk-frontend/pull/969))
 
+- Rename `name` argument of date-input component to `namePrefix`.
+
+  This is better reflective of the purpose of the argument, which is to prefix the `name` attribute of `items`. This is consistent with other components which use the name `idPrefix` to explain similar functionality.
+
+  If your project currently uses this optional argument with the date-input macro, you need to rename all instances of it to `namePrefix` (NB: this argument shouldn't be confused with the `items.{}.name` attribute which hasn't changed.)
+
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -72,6 +72,14 @@
 
   ([PR #969](https://github.com/alphagov/govuk-frontend/pull/969))
 
+- Rename `name` argument of date-input component to `namePrefix`.
+
+  This is better reflective of the purpose of the argument, which is to prefix the `name` attribute of `items`. This is consistent with other components which use the name `idPrefix` to explain similar functionality.
```
36degrees: Great changelog entry – I think this gets the intention across well, and makes it clear what the user needs to do 👍

src/components/date-input/README.njk
```diff
@@ -55,7 +55,7 @@
         text: 'No'
       },
       {
-        text: 'Optional name. This is used to compose the name attribute for each item.'
+        text: 'Optional prefix. This is used to prefix each `item.name`.'
```
36degrees: I wonder if it's worth mentioning that they'll be joined using "-"?

hannalaakso: Done 👍  I thought initially about adding it but it felt like too much about the implementation. But it's probably useful information.

PR #985, opened 2018-09-06
Remove anchor styling in govuk-lists

nickcolley:
These styles partially implement the `.govuk-link` styles, but fail to implement the correct focus styles.

Additionally this need can be satisfied by enabling global styles, which is on by default in the GOV.UK Prototype Kit.

I have checked our [documentation](https://design-system.service.gov.uk/styles/typography/#lists) and we do not recommend our users do this.

With this both in mind this feels reasonable to remove this functionality as a breaking change, without a deprecation period.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -103,6 +103,16 @@
   ```
   ([PR #981](https://github.com/alphagov/govuk-frontend/pull/981))
 
+- Remove anchor styling in govuk-lists
+
+  This was an undocumented feature of lists.
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -103,6 +103,16 @@
   ```
   ([PR #981](https://github.com/alphagov/govuk-frontend/pull/981))
 
+- Remove anchor styling in govuk-lists
```
36degrees: 👍

PR #998, opened 2018-09-13
Add component options (args) to YAML, add as JSON to `package`

hannalaakso:
We want to expose the component options (arguments) to GOV.UK Design System. To do this, they need to be generated for inclusion in `package`.

This PR:
- Adds all component options in the component YAML as `params`. 
- Introduces `isComponent` option (so that the options for that component can be linked to in the Design System).
- Copies the component options as JSON files to `package` when `gulp build:package --destination 'package'` is run.
- Amends the tests for checking contents of `package` to allow filenames with wildcards in `additionalFilesNotInSrc`
- Adds a test to check all `package/components` have a `macro-options.json` that contains valid JSON with the expected attributes `name`, `type`, `required`, `description`

Created this as one PR so that the generated JSON files can be tested against https://github.com/alphagov/govuk-design-system/pull/518 

More on component options:
- We now refer to these as “options” in client facing docs as calling them “arguments” is technically incorrect and “parameters” was felt to be less clear than “options”.
- Regarding the "or" relationship between `text` and `html` attributes, we are continuing to explain this in the description of the attributes, ie. "If `html` is set, this is not required." There didn't seem to be a sensible way of trying to show this relationship in a simple table so using just text appeared to be  the most accessible method.
- This PR retains the convention of referring to them as “params” in component macro and template files - we might want to roll out renaming “params” to “options”  in all component macros and templates as a separate PR that can be quickly merged in without causing too many conflicts with other PRs.

To test, run `npm run build:package`

https://trello.com/c/QN0imhXt/722-define-arguments-for-components-in-the-component-yaml-file 
https://trello.com/c/00iDbQOU/1323-allow-argument-yamls-to-be-transformed-into-json-and-copied-to-package

### DIFF
--- a/src/components/breadcrumbs/breadcrumbs.yaml
+++ b/src/components/breadcrumbs/breadcrumbs.yaml
@@ -1,3 +1,30 @@
+params:
+- name: items
+  type: array
+  required: true
+  description: Array of breadcrumbs item objects.
+  params:
+  - name: text
+    type: string
+    required: true
+    description: If `html` is set, this is not required. Text to use within the breadcrumbs item. If `html` is provided, the `text` argument will be ignored.
+  - name: html
+    type: string
+    required: true
+    description: If `text` is set, this is not required. HTML to use within the breadcrumbs item. If `html` is provided, the `text` argument will be ignored.
+  - name: href
+    type: string
+    required: false
+    description: Link for the breadcrumbs item. If not specified, breadcrumbs item is a normal list item.
+- name: classes
+  type: string
+  required: false
+  description: Classes to add to the breadcrumbs container.
+- name: attributes
+  type: object
+  required: false
+  description: HTML attributes (for example data attributes) to add to the breadcrumbs container.
+
 examples:
 - name: default
   data:
--- a/src/components/date-input/date-input.yaml
+++ b/src/components/date-input/date-input.yaml
@@ -1,3 +1,61 @@
+params:
+- name: id
+  type: string
+  required: false
+  description: This is used for the main component and to compose id attribute for each item.
+- name: namePrefix
+  type: string
+  required: false
+  description: Optional prefix. This is used to prefix each `item.name` using `-`.
+- name: items
+  type: array
+  required: true
+  description: An array of input objects with name, value and classes.
+  params:
+  - name: id
+    type: string
+    required: false
+    description: Item-specific id. If provided, it will be used instead of the generated id.
+  - name: name
+    type: string
+    required: true
+    description: Item-specific name attribute.
+  - name: label
+    type: string
+    required: true
+    description: Item-specific label text. If provided, this will be used instead of `name` for item label text.
+  - name: value
+    type: string
+    required: false
+    description: If provided, it will be used as the initial value of the input.
+  - name: classes
+    type: string
+    required: false
+    description: Classes to add to date input item.
+- name: hint
+  type: object
+  required: false
+  description: Options for the hint component.
+  isComponent: true
+- name: errorMessage
+  type: object
+  required: false
+  description: Options for the error message.
+  isComponent: true
+- name: fieldset
+  type: object
+  required: false
+  description: Options for the fieldset component (e.g. legend).
+  isComponent: true
+- name: classes
+  type: string
+  required: false
+  description: Classes to add to the date-input container.
+- name: attributes
+  type: object
+  required: false
+  description: HTML attributes (for example data attributes) to add to the date-input container.
+
 examples:
 - name: default
   data:
--- a/src/components/details/details.yaml
+++ b/src/components/details/details.yaml
@@ -1,3 +1,37 @@
+params:
+- name: summaryText
+  type: string
+  required: true
+  description: If `summmaryHtml` is set, this is not required. Text to use within the summary element (the visible part of the details element). If `summaryHtml` is provided, the `summaryText` argument will be ignored.
+- name: summaryHtml
+  type: string
+  required: true
+  description: If `summmaryText` is set, this is not required. HTML to use within the summary element (the visible part of the details element). If `summaryHtml` is provided, the `summaryText` argument will be ignored.
+- name: text
+  type: string
+  required: true
+  description: If `html` is set, this is not required. Text to use within the disclosed part of the details element. If `html` is provided, the `text` argument will be ignored.
+- name: html
+  type: string
+  required: true
+  description: If `text` is set, this is not required. HTML to use within the disclosed part of the details element. If `html` is provided, the `text` argument will be ignored.
+- name: id
+  type: string
+  required: false
+  description: Id to add to the details element.
+- name: open
+  type: boolean
+  required: false
+  description: If true, details element will be expanded.
+- name: classes
+  type: string
+  required: false
+  description: Classes to add to the `<details>` element.
+- name: attributes
+  type: object
+  required: false
+  description: HTML attributes (for example data attributes) to add to the `<details>` element.
+
 examples:
 - name: default
   data:
--- a/src/components/error-summary/error-summary.yaml
+++ b/src/components/error-summary/error-summary.yaml
@@ -1,3 +1,51 @@
+params:
+- name: titleText
+  type: string
+  required: true
+  description: If `titleHtml` is set, this is not required. Text to use for the heading of the error summary block. If `titleHtml` is provided, `titleText` will be ignored.
+- name: titleHtml
+  type: string
+  required: true
+  description: If `titleText` is set, this is not required. HTML to use for the heading of the error summary block. If `titleHtml` is provided, `titleText` will be ignored.
+- name: descriptionText
+  type: string
+  required: false
+  description: If `descriptionHtml` is set, this is not required. Text to use for the description of the errors. If `descriptionHtml` is provided,  `descriptionText` will be ignored.
+- name: descriptionHtml
+  type: string
+  required: true
+  description: If `descriptionText` is set, this is not required. HTML to use for the description of the errors. If `descriptionHtml` is provided, `descriptionText` will be ignored.
+
+- name: errorList
+  type: array
+  required: true
+  description: Contains an array of error link items and all their available arguments.
+  params:
+  - name: href
+    type: string
+    required: false
+    description: Href attribute for the error link item. If provided item will be an anchor.
+  - name: text
+    type: string
+    required: true
+    description: If `html` is set, this is not required. Text for the error link item. If `html` is provided, the `text` argument will be ignored.
+  - name: html
+    type: string
+    required: true
+    description: If `text` is set, this is not required. HTML for the error link item. If `html` is provided, the `text` argument will be ignored.
+  - name: attributes
+    type: object
+    required: false
+    description: HTML attributes (for example data attributes) to add to the error link anchor.
+- name: classes
+  type: string
+  required: false
+  description: Classes to add to the error-summary container.
+- name: attributes
+  type: object
+  required: false
+  description: HTML attributes (for example data attributes) to add to the error-summary container.
+
 accessibilityCriteria: |
   - Must be focused when the page loads
 
--- a/src/components/fieldset/fieldset.yaml
+++ b/src/components/fieldset/fieldset.yaml
@@ -1,3 +1,42 @@
+params:
+- name: describedBy
+  type: string
+  required: false
+  description: Text or element id to add to the `aria-describedby` attribute to provide description of the group of fields for screenreader users.
+- name: legend
+  type: object
+  required: false
+  description: Options for the legend
+  params:
+  - name: text
+    type: string
+    required: true
+    description: If `html` is set, this is not required. Text to use within the legend. If `html` is provided, the `text` argument will be ignored.
+  - name: html
+    type: string
+    required: true
+    description: If `text` is set, this is not required. HTML to use within the legend. If `html` is provided, the `text` argument will be ignored.
+  - name: classes
+    type: string
+    required: false
+    description: Classes to add to the legend.
+  - name: isPageHeading
+    type: boolean
+    required: false
+    description: Whether the legend also acts as the heading for the page.
+- name: classes
+  type: string
+  required: false
+  description: Classes to add to the fieldset container.
+- name: attributes
+  type: object
+  required: false
+  description: HTML attributes (for example data attributes) to add to the fieldset container.
+- name: caller
+  type: nunjucks-block
+  required: false
+  description: Not strictly a parameter but [Nunjucks code convention](https://mozilla.github.io/nunjucks/templating.html#call). Using a `call` block enables you to call a macro with all the text inside the tag. This is helpful if you want to pass a lot of content into a macro. To use it, you will need to wrap the entire fielset component in a `call` block. See [checkboxes component](https://github.com/alphagov/govuk-frontend/blob/master/src/components/checkboxes/template.njk#L86) for an example.
+
 examples:
 - name: default
   data:
--- a/src/components/footer/footer.yaml
+++ b/src/components/footer/footer.yaml
@@ -1,3 +1,69 @@
+params:
+- name: meta
+  type: object
+  required: false
+  description: Object containing options for the meta navigation.
+  arguments:
+  - name: items
+    type: array
+    required: false
+    description: Array of items for use in the meta section of the footer.
+    params:
+    - name: text
+      type: string
+      required: false
+      description: List item text in the meta section of the footer.
+    - name: href
+      type: string
+      required: false
+      description: List item href attribute in the meta section of the footer.
+    - name: attributes
+      type: object
+      required: false
+      description: HTML attributes (for example data attributes) to add to the anchor in the footer meta section.
+- name: navigation
+  type: array
+  required: false
+  description: Array of items for use in the navigation section of the footer.
+  arguments:
+  - name: title
+    type: string
+    required: false
+    description: Title for a section
+  - name: columns
+    type: integer
+    required: false
+    description: Amount of columns to display items in navigation section of the footer.
+  - name: items
+    type: array
+    required: false
+    description: Array of items to display in the list in navigation section of the footer.
+    arguments:
+    - name: text
+      type: string
+      required: false
+      description: List item text in the navigation section of the footer.
+    - name: href
+      type: string
+      required: false
+      description: List item href attribute in the navigation section of the footer. Both `text` and `href` attributes need to be present to create a link.
+    - name: attributes
+      type: object
+      required: false
+      description: HTML attributes (for example data attributes) to add to the anchor in the footer navigation section.
+- name: containerClasses
+  type: string
+  required: false
+  description: Classes that can be added to the inner container, useful if you want to make the footer full width.
+- name: classes
+  type: string
+  required: false
+  description: Classes to add to the footer component container.
+- name: attributes
+  type: object
+  required: false
+  description: HTML attributes (for example data attributes) to add to the footer component container.
+
 previewLayout: full-width
 accessibilityCriteria: |
   Text and links in the Footer must:
--- a/src/components/header/header.yaml
+++ b/src/components/header/header.yaml
@@ -1,3 +1,62 @@
+params:
+- name: homepageUrl
+  type: string
+  required: false
+  description: The url of the homepage. Defaults to /
+- name: assetsPath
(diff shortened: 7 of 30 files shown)


tasks/gulp/copy-to-destination.js
```diff
@@ -25,5 +30,24 @@ gulp.task('copy-files', () => {
...
     ], {syntax: require('postcss-scss')}))
     .pipe(scssFiles.restore)
+    .pipe(yamlFiles)
+    .pipe(map(function (file, done) {
+      let componentName = path.dirname(file.path).split(path.sep).slice(-1).toString()
+      let componentPath = path.join(configPaths.components, componentName, `${componentName}.yaml`)
+
+      try {
+        let json = yaml.safeLoad(fs.readFileSync(componentPath), {encoding: 'utf8', json: true})
+        let paramsJson = {'params': json.params} // We only want the 'params' data from component yaml
```
igloosi: should we stop the task if the yaml file doesn't contain the `params` data

src/components/breadcrumbs/breadcrumbs.yaml
```diff
@@ -1,3 +1,30 @@
...
+  required: true
+  description: Array of breadcrumbs item objects.
+  params:
+  - name: text
+    type: string
+    required: true
+    description: If `html` is set, this is not required. Text to use within the breadcrumbs item. If `html` is provided, the `text` argument will be ignored.
+  - name: html
+    type: string
+    required: true
```
igloosi: did we postpone the 'what happens when text OR html are required?

hannalaakso: Good question, I've included that in the PR description. We decided with @joelanman to fall back using the description of the attributes to explain the relationship.

tasks/gulp/copy-to-destination.js
hannalaakso: This is in a `try` block so it will throw up an error if that happens and the tests will also fail. That's probably okay? This is in the middle of the `copy-files` task so if we exit here, there might already be files copied over to `package` and others deleted.

src/components/fieldset/fieldset.yaml
```diff
@@ -1,3 +1,38 @@
+params:
```
nickcolley: do we need to think about `caller` for fieldset, or can this be done later

hannalaakso: I've added an explanation of `caller` for fieldset, have a look. It strikes me that we could probably do with an explanation of `caller` in our docs that we could then link to from the fieldset table.

src/components/fieldset/fieldset.yaml
```diff
@@ -1,3 +1,42 @@
...
+  required: false
+  description: Classes to add to the fieldset container.
+- name: attributes
+  type: object
+  required: false
+  description: HTML attributes (for example data attributes) to add to the fieldset container.
+- name: caller
+  type: nunjucks-block
+  required: false
+  description: Not strictly a parameter but <a href="https://mozilla.github.io/nunjucks/templating.html#call">Nunjucks code convention</a>. Using a `call` block enables you to call a macro with all the text inside the tag. This is helpful if you want to pass a lot of content into a macro. To use it, you will need to wrap the entire fielset component in a `call` block. See <a href="https://github.com/alphagov/govuk-frontend/blob/master/src/components/checkboxes/template.njk#L86">checkboxes component</a> for an example.
```
joelanman: We're using html here for the links. We're also using ` backticks for option names, which is a markdown convention. We don't have to do anything with the backticks, but it's probably best to run them through a renderer to get <code> blocks. In which case maybe we should just use markdown for the links too?

hannalaakso: Good shout. I was avoiding markdown since I only needed to use HTML in this one place for the component options and we discovered we'll need to do some extra processing as part of the build in the Design System to render markdown from here. But since we've got the code blocks too then making it all markdown makes sense 🙂

tasks/gulp/copy-to-destination.js
```diff
@@ -25,5 +30,24 @@ gulp.task('copy-files', () => {
...
+    .pipe(map(function (file, done) {
+      let componentName = path.dirname(file.path).split(path.sep).slice(-1).toString()
+      let componentPath = path.join(configPaths.components, componentName, `${componentName}.yaml`)
+
+      try {
+        let json = yaml.safeLoad(fs.readFileSync(componentPath), {encoding: 'utf8', json: true})
+        let paramsJson = json.params // We only want the 'params' data from component yaml
+        file.contents = Buffer.from(JSON.stringify(paramsJson, null, 4))
+      } catch (e) {
+        console.log('ENOENT: no such file or directory: ', file)
```
joelanman: This should probably be `console.error`. Also the `try` section has 3 lines in it, so we don't know for sure that this is the error. I'd recommend trying to reduce the `try` to just opening the file.

tasks/gulp/copy-to-destination.js
```diff
@@ -25,5 +30,24 @@ gulp.task('copy-files', () => {
       autoprefixer
     ], {syntax: require('postcss-scss')}))
     .pipe(scssFiles.restore)
+    .pipe(yamlFiles)
+    .pipe(map(function (file, done) {
+      let componentName = path.dirname(file.path).split(path.sep).slice(-1).toString()
+      let componentPath = path.join(configPaths.components, componentName, `${componentName}.yaml`)
+
+      try {
+        let json = yaml.safeLoad(fs.readFileSync(componentPath), {encoding: 'utf8', json: true})
```
joelanman: the `readFileSync` could be pulled out into a line above to make it more readable and have fewer operations happening in a single line

joelanman:
Do we still have a frontend test app? Maybe we can read the macro options json and write to a page in that, so we can see it works?

tasks/gulp/copy-to-destination.js
hannalaakso: I've split up the error messages more, have a look.

hannalaakso:
@joelanman Thanks a lot for the review 🙌 

We test if the file contains JSON and if the JSON contains attributes (`name`, `type`, `required`, `description`) that we would expect it to have. We could use the frontend test app for testing the JSON but it seems like a slightly odd way of testing it (maybe put it in a table?) since the intention isn't to surface the JSON in GOV.UK Frontend review app in any way. Do you there are specific advantages to testing it in this way?

tasks/gulp/copy-to-destination.js
hannalaakso: Changed!

tasks/gulp/__tests__/after-build-package.test.js
```diff
@@ -89,4 +98,42 @@ describe('package/', () => {
...
+          return readFile(filePath, 'utf8')
+            .then((data) => {
+              var newData = JSON.parse(data)
+              // We expect the component JSON to contain at least "name" and "type"
+              expect(newData).toEqual(
+                expect.arrayContaining([
+                  expect.objectContaining({
+                    name: expect.anything(),
+                    type: expect.anything(),
+                    required: expect.anything(),
```
nickcolley: Non-blocking but I think these could be more strictly typed to bool for example in 'required'?

Not a big deal, this test is really nice.

tasks/gulp/__tests__/after-build-package.test.js
```diff
@@ -89,4 +98,42 @@ describe('package/', () => {
...
+              return { data }
+            }).catch(error => {
+              throw error
+            })
+        })
+      }
+      Promise
+        .all(getMacroOptionsContents())
+        .then(() => { done() })
+        .catch(error => { throw error })
```
igloosi: don't think this line is needed as the test fails higher up in the getMacroOptions function

tasks/gulp/__tests__/after-build-package.test.js
hannalaakso: Good point, I've made it stricter by checking for strings with `name`, `type` and `description`. For `type`, I don't know if there is a way to check if the value is a boolean? [jest-extended](https://www.npmjs.com/package/jest-extended) has `toBeBoolean()`.

nickcolley: does expect.any(Boolean) work?

tasks/gulp/__tests__/after-build-package.test.js
```diff
@@ -89,4 +98,42 @@ describe('package/', () => {
...
+            })
+            .then(data => {
+              return { data }
+            }).catch(error => {
+              throw error
+            })
+        })
+      }
+      Promise
+        .all(getMacroOptionsContents())
```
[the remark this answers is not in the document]
hannalaakso: `promise.all()` gets a single promise that resolves from the multiple promises that are looped through in `getMacroOptionsContents()`. I think this is the best way of doing it but happy to consider a different solution? 🙂

tasks/gulp/__tests__/after-build-package.test.js
hannalaakso: Removed!

tasks/gulp/__tests__/after-build-package.test.js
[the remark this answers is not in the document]
nickcolley: If this is true you might be able to simplify all this logic using `it.each` https://github.com/alphagov/govuk-frontend/blob/5b482293627363847545aa16a4f10b055bdbcb60/src/components/all.test.js#L18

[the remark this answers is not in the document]
nickcolley: Had a think about this on the way home yesterday, I think it's fine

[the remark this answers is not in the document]
hannalaakso: @nickcolley Actually I found what you said really helpful and I've rewritten the test, I think it's cleaner now, have a look. The only difference is that using `it.each` prints out the test result for each component in the console but that's probably okay, what do you think? Or I can revert back to the old solution but get rid of the unnecessary 
```
 .then(data => {
    return { data }
})
```

src/components/inset-text/inset-text.yaml
```diff
@@ -1,3 +1,25 @@
...
+  type: string
+  required: true
+  description: If `html` is set, this is not required. Text to use within the back link component. If `html` is provided, the `text` argument will be ignored.
+- name: html
+  type: string
+  required: true
+  description: If `text` is set, this is not required. HTML to use within the back link component. If `html` is provided, the `text` argument will be ignored.
+- name: id
+  type: string
+  required: true
```
igloosi: id is optional

tasks/gulp/__tests__/after-build-package.test.js
```diff
@@ -89,4 +98,39 @@ describe('package/', () => {
...
+                    name: expect.any(String),
+                    type: expect.any(String),
+                    required: expect.any(Boolean),
+                    description: expect.any(String)
+                  })
+                ])
+              )
+            })
+            .then(data => {
+              return { data }
```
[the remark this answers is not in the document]
hannalaakso: Agreed, this `then` is superflous

src/components/panel/panel.yaml
```diff
@@ -1,3 +1,33 @@
...
+  required: true
+  description: If `titleHtml` is set, this is not required. Text to use within the panel. If `titleHtml` is provided, the `titleText` argument will be ignored.
+- name: titleHtml
+  type: string
+  required: true
+  description: If `titleText` is set, this is not required. HTML to use within the panel. If `titleHtml` is provided, the `titleText` argument will be ignored.
+- name: headingLevel
+  type: integer
+  required: false
+  description: Heading level, from 1 to 6. Default is 2.
```
igloosi: default is now 1

nickcolley:
Arguments that have been checked:

- [x] hint @igloosi 
- [ ] input @igloosi 
- [ ] label  @igloosi

src/components/header/header.yaml
```diff
@@ -1,3 +1,58 @@
...
+    required: false
+    description: Text of the navigation item.
+  - name: href
+    type: string
+    required: false
+    description: Url of the navigation item anchor. Both `href` and `text` attributes for navigation items need to be provided to create an item.
+  - name: active
+    type: boolean
+    required: false
+    description: Flag to mark the navigation item as active or not.
```
igloosi: missing navigation.attributes

src/components/skip-link/skip-link.yaml
```diff
@@ -1,3 +1,25 @@
+params:
+- name: text
+  type: string
+  required: true
```
nickcolley: text and html are not required in the current table, since they have a default

src/components/footer/footer.yaml
```diff
@@ -1,3 +1,61 @@
...
+    arguments:
+    - name: text
+      type: string
+      required: false
+      description: List item text in the navigation section of the footer.
+    - name: href
+      type: string
+      required: false
+      description: List item href attribute in the navigation section of the footer. Both `text` and `href` attributes need to be present to create a link.
+- name: containerClasses
```
igloosi: missing navigation.items.attributes

src/components/footer/footer.yaml
```diff
@@ -1,3 +1,61 @@
...
+    description: Array of items for use in the meta section of the footer.
+    params:
+    - name: text
+      type: string
+      required: false
+      description: List item text in the meta section of the footer.
+    - name: href
+      type: string
+      required: false
+      description: List item href attribute in the meta section of the footer.
```
igloosi: missing meta.items.attributes

src/components/date-input/date-input.yaml
```diff
@@ -1,3 +1,57 @@
...
+  description: This is used for the main component and to compose id attribute for each item.
+- name: namePrefix
+  type: string
+  required: false
+  description: Optional prefix. This is used to prefix each `item.name` using `-`.
+- name: items
+  type: array
+  required: true
+  description: An array of input objects with name, value and classes.
+  params:
```
36degrees: This is missing `classes` – it wasn't in the README either 😢

joelanman:
@hannalaakso 

> Do you there are specific advantages to testing it in the test app or anything in particular we should look out for?

I think it could be useful to easily browse the macro option documentation in this project, instead of having to import it into the Design System to see it properly. But this can be another PR, if we want to do it

src/components/date-input/date-input.yaml
```diff
@@ -1,3 +1,57 @@
...
+  description: Options for the fieldset component (e.g. legend).
+  isComponent: true
+- name: classes
+  type: string
+  required: false
+  description: Classes to add to the anchor tag.
+- name: attributes
+  type: object
+  required: false
+  description: HTML attributes (for example data attributes) to add to the anchor tag.
```
36degrees:
```diff
- anchor tag
+ wrapper (?)
```

src/components/date-input/date-input.yaml
```diff
@@ -1,3 +1,57 @@
...
+  isComponent: true
+- name: fieldset
+  type: object
+  required: false
+  description: Options for the fieldset component (e.g. legend).
+  isComponent: true
+- name: classes
+  type: string
+  required: false
+  description: Classes to add to the anchor tag.
```
36degrees:
```diff
- anchor tag
+ wrapper (?)
```

src/components/details/details.yaml
```diff
@@ -1,3 +1,37 @@
+params:
+- name: summaryText
+  type: string
+  required: true
```
36degrees: Are we consistent with this throughout? We say it's `required` if you have to pass either one or the other?

src/components/details/details.yaml
```diff
@@ -1,3 +1,37 @@
...
+  description: If `summmaryHtml` is set, this is not required. Text to use within the summary element (the visible part of the details element). If `summaryHtml` is provided, the `summaryText` argument will be ignored.
+- name: summaryHtml
+  type: string
+  required: true
+  description: If `summmaryText` is set, this is not required. Text or HTML to use within the summary element (the visible part of the details element). If `summaryHtml` is provided, the `summaryText` argument will be ignored.
```
36degrees: Suggest we change 'Text or HTML' to just 'HTML' to reduce cases where people are using the HTML option unnecessarily?

src/components/details/details.yaml
```diff
@@ -1,3 +1,37 @@
...
+- name: text
+  type: string
+  required: true
+  description: If `html` is set, this is not required. Text or HTML to use within the disclosed part of the details element. If `html` is provided, the `text` argument will be ignored.
+- name: html
+  type: string
+  required: true
+  description: If `text` is set, this is not required. Text or HTML to use within the disclosed part of the details element. If `html` is provided, the `text` argument will be ignored.
```
36degrees: Suggest we change 'Text or HTML' to just 'HTML' to reduce cases where people are using the HTML option unnecessarily?

src/components/details/details.yaml
```diff
@@ -1,3 +1,37 @@
...
+  required: false
+  description: Id to add to the details element.
+- name: open
+  type: boolean
+  required: false
+  description: If true, details element will be expanded.
+- name: classes
+  type: string
+  required: false
+  description: Classes to add to the anchor tag.
```
36degrees:
```diff
- anchor tag
+ `<details>` element
```

src/components/details/details.yaml
```diff
@@ -1,3 +1,37 @@
...
+- name: attributes
+  type: object
+  required: false
+  description: HTML attributes (for example data attributes) to add to the anchor tag.
```
36degrees:
```diff
- anchor tag
+ `<details>` element
```

src/components/error-summary/error-summary.yaml
```diff
@@ -1,3 +1,38 @@
...
+- name: titleText
+  type: string
+  required: true
+  description: If `titleHtml` is set, this is not required. Text to use for the heading of the error summary block. If `titleHtml` is provided, the `titleText` argument will be ignored.
+- name: titleHtml
+  type: string
+  required: true
+  description: If `titleText` is set, this is not required. HTML to use for the heading of the error summary block. If `titleHtml` is provided, the `titleText` argument will be ignored.
+- name: errorList
+  type: object
```
36degrees: I think this should be `array`? Inconsistent with `items` on date input.

src/components/error-summary/error-summary.yaml
```diff
@@ -1,3 +1,38 @@
...
+  required: true
+  description: Contains an array of error link items and all their available arguments.
+  params:
```
36degrees: Missing `attributes`.

src/components/error-summary/error-summary.yaml
```diff
@@ -1,3 +1,38 @@
+params:
```
36degrees: Missing `descriptionText`, `descriptionHtml`.

src/components/fieldset/fieldset.yaml
```diff
@@ -1,3 +1,42 @@
...
+    required: true
+    description: If `html` is set, this is not required. Text to use within the legend. If `html` is provided, the `text` argument will be ignored.
+  - name: html
+    type: string
+    required: true
+    description: If `text` is set, this is not required. HTML to use within the legend. If `html` is provided, the `text` argument will be ignored.
+  - name: classes
+    type: string
+    required: false
+    description: Classes to add to the legend container.
```
36degrees:
```diff
- container
```
?

36degrees: Sorry, realise this is the same in the README – happy for you to leave these if you prefer.

src/components/skip-link/skip-link.yaml
nickcolley: Think I was thinking of the back-link, it probably should have a default... Disregard this comment :)

src/components/date-input/date-input.yaml
hannalaakso: Done!

src/components/details/details.yaml
hannalaakso: Yes 👍 See this in PR description:
> Regarding the "or" relationship between text and html attributes, we are continuing to explain this in the description of the attributes, ie. "If html is set, this is not required." There didn't seem to be a sensible way of trying to show this relationship in a simple table so using just text appeared to be the most accessible method.

src/components/error-summary/error-summary.yaml
hannalaakso: Done!

Issue #927, opened 2018-07-24
Tabs result in a flash of unstyled content as the page loads


nickcolley:
Similar to the [problem we had with conditional reveals](https://github.com/alphagov/govuk-frontend/issues/787) if the JavaScript is slow to load tabs will be shown stacked in a broken fashion. 

You can see a simulation of this in this closed PR https://github.com/alphagov/govuk-frontend/pull/917

nickcolley:
I have a WIP branch for this: https://github.com/alphagov/govuk-frontend/compare/fix-fouc-with-tabs-panels

Main thing I think we'd need to introduce is being able to set an initial active state for tabs.

adamsilver:
Originally I had it so that the default active state (unless in the location hash) is the first tab.

PR #1000, opened 2018-09-14
Fix flash of unstyled content in tabs component

nickcolley:
Closes https://github.com/alphagov/govuk-frontend/issues/927

On slow connections or devices users can see a flash of the unstyled version of the tabs.

This PR updates the tab component so that it's initial state can be set without the need for JavaScript.

I experimented with being able to set a specific panel, which I thought would allow rendering of selected tabs without any flashing, but since the hash set in the URL is not available to the server this would not be practical. - https://stackoverflow.com/a/318581

Since I could not think of an immediate need for someone to set a default tab other than the first one it seemed over the top and something we could add later on.

### Before, flash of unstyled content
before-tabs-unstyled-flash

### After, no flash with default tab
after-tabs-unstyled-flash

### Flash when changing tabs and reloading
after-tabs-unstyled-flash-selected-tab

https://trello.com/c/sNeuUPNH/1386-2-fix-flashes-of-unstyled-content-caused-by-the-tabs-component

### DIFF
--- a/src/components/tabs/README.md
+++ b/src/components/tabs/README.md
@@ -24,7 +24,7 @@ Find out when to use the tabs component in your service in the [GOV.UK Design Sy
       <ul class="govuk-tabs__list">
 
           <li class="govuk-tabs__list-item">
-            <a class="govuk-tabs__tab" href="#past-day">
+            <a class="govuk-tabs__tab govuk-tabs__tab--selected" href="#past-day">
               Past day
             </a>
           </li>
@@ -80,7 +80,7 @@ Find out when to use the tabs component in your service in the [GOV.UK Design Sy
 
       </section>
 
-      <section class="govuk-tabs__panel" id="past-week">
+      <section class="govuk-tabs__panel govuk-tabs__panel--hidden" id="past-week">
         <h2 class="govuk-heading-l">Past week</h2>
     <table class="govuk-table">
       <thead class="govuk-table__head">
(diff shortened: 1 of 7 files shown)


src/components/tabs/README.md
```diff
@@ -80,7 +80,7 @@ Find out when to use the tabs component in your service in the [GOV.UK Design Sy
 
       </section>
 
-      <section class="govuk-tabs__panel" id="past-week">
+      <section class="govuk-tabs__panel govuk-tabs__panel--hidden" id="past-week">
```
36degrees: Do we need both the hidden modifier _and_ the selected modifier? Can the tabs not be hidden by default unless they have the `--selected` modifier?

Only having one modifier simplifies the code a little bit and also avoids the possibility of a weird 'undefined' state where a tab has neither the selected modifier nor the hidden modifier.

36degrees: Possibly… for context ,I know that I ended up [refactoring the mobile nav in the Design System](https://github.com/alphagov/govuk-design-system/pull/483/files) so that it didn't have both `active` and `hidden` modifiers, because it was confusing (and broken, because sometimes a menu item had neither modifier, and that scenario was poorly defined – see the first commit) so a little wary of this pattern.

nickcolley: I agree it seems there could be a less error prone way of doing this, but given this is consistent with another component and making tab panels hidden by default would be a breaking change, could we raise that as a separate issue?

36degrees: 👍

Issue #755, opened 2018-06-04
Check for broken links in markdown files


hannalaakso:
It would be good to run tests against the project markdown files to ensure they don't contain broken links.

36degrees:
Closing this as I don't think we'd prioritise it now that most of the user-facing documentation lives in alphagov/govuk-frontend-docs.

There's a new issue (https://github.com/alphagov/govuk-frontend-docs/issues/8) that represents adding a link checker to that repo.

PR #1005, opened 2018-09-19
Check markdown links

nickcolley:
Closes https://github.com/alphagov/govuk-frontend/issues/755

Adds markdown internal link checking using the remark library, and a validator plugin.

I have fixed a bunch of errors after introduce this.

The errors it produces are really useful, giving you the path that it is actually resolving to:

screen shot 2018-09-19 at 17 59 34

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -912,7 +912,7 @@ You can now initialize individual components like so:
 
 - Restructure project to enable it to be published as a single package
 
-  The [new project structure]((./src/README.md)) matches our ITCSS inspired layers and is published as a single package as `@govuk-frontend/frontend`.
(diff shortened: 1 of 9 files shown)


nickcolley:
Will try and redo this with jest maybe, no idea why this is failing to be honest...

36degrees:
Trying a rebuild… can't see how those failures would relate to this change.

nickcolley:
This might be better done as a probot https://probot.github.io/apps/

nickcolley:
So I've done a lot of thinking and poking the code.

The programmatic interface to the remark library doesnt check links between files which is one of the biggest selling points...

nickcolley:
Will close this for now, and come back to it when I have some time.

Issue #891, opened 2018-07-13
"Header full width with navigation" example text slightly bunched up


hannalaakso:
The "Product name" in the example ends up nearly touching the "Menu" text. We should probably add a bit of margin to to the menu button so that it's clearly separate from the service name. 

Screen grab from Samsung Galaxy S8 (Chrome and Samsung Internet):

<img width="359" alt="screen shot 2018-07-13 at 09 53 50" src="https://user-images.githubusercontent.com/5007934/42682504-af6074ce-8682-11e8-80f4-60138ab5d500.png">

PR #1007, opened 2018-09-21
Prevent product name in header from wrapping and add some spacing

igloosi:
Currently the product name in the header wraps when the space shrinks which doesn't look great.

screen shot 2018-09-21 at 11 40 57
screen shot 2018-09-21 at 11 41 06

Adding `display:inline-table` prevents that so that the product name as a whole drops to a new line when space is shrunk.

screen shot 2018-09-21 at 11 44 25
screen shot 2018-09-21 at 11 45 03

<details>
<summary>GIFs</summary>

</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,13 +22,21 @@
 
 🔧 Fixes:
 
-- Apply max-width to the `<select> element
+- Apply max-width to the `<select>` element
 
   The `<select>` element's width is governed by the widths of its `<option>`'s.
 
   When the text in the options grows large, the element can grow to > 100% of the width of its container and break the layout.
 
   ([PR #1013](https://github.com/alphagov/govuk-frontend/pull/1013))
+
+- Prevent product name in header from wrapping
+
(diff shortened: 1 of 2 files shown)


igloosi:
@dashouse you're right, something like WEEE service will break. hmmm

dashouse:
@igloosi You'll have to remind me about WEEE, but if it's a service it would be using the service name not the product name.

igloosi:
@dashouse https://weee.service.gov.uk/ yeah true

dashouse:
screen shot 2018-09-21 at 12 12 26

Cool, looks like your change only affects product name anyway which is fine, services are less likely to be only two words anyway so wrapping is ok

igloosi:
@dashouse I've also added some spacing for #891 
does that look ok?

dashouse:
👍 LGTM

igloosi:
and Ipad Pro example if someone had both product and service name in there

before: 
screen shot 2018-09-23 at 08 53 27

after:
screen shot 2018-09-23 at 08 53 45

PR #986, opened 2018-09-07
Add 48px favicon

dashouse:
Add a single favicon with multiple sizes
Fixes #963

PR #1008, opened 2018-09-21
Revert 16x16 and 32x32 versions of the favicon

36degrees:
When adding the additional 48x48 size to the favicon, the 16x16 and 32x32 versions of the favicon were also updated to new versions based off the same source image used to generate the 48x48 version.

These versions do not ‘fit’ the underlying pixel grid as well as the old versions, which means the crown now appears blurry at these sizes.

This change restores the original 16x16 and 32x32 versions from ce755c7ee01e417585e286825cfb7d92f7c89fa3, whilst adding the new 48x48 version from d6e3949250f1eb6ef981bd7d759b5b01f6cade9c. Hopefully this gives us the best of both worlds.

**Images in the original version**

screen shot 2018-09-21 at 14 36 27

**Images in favicon.ico as of #986**

screen shot 2018-09-21 at 14 36 34

**Images in this version**

screen shot 2018-09-21 at 15 01 56


nickcolley:
The way they're displayed here they all look blurry, but if you're happy they're good 👍

36degrees:
There's so much detail in the crown we're never going to make it look sharp – regardless, the intent of this PR is not to make them perfect, just to make sure we don't _lose_ any sharpness in the smaller resolutions over what we used to have.

36degrees:
For reference, I've been testing using https://redketchup.io/icon/edit if you want to take a closer look at what's going on.

PR #1014, opened 2018-09-27
Ignore .DS_Store files in `after-build-dist` test

igloosi:
Currently the `after-build-dist` test fails as it includes .DS_Store files in the array of expected files.

This work adds an array of files to ignore, similarly to how the `after-build-package` test is set up.

Before:

screen shot 2018-09-27 at 08 41 53
screen shot 2018-09-27 at 08 42 01

After:
screen shot 2018-09-27 at 08 51 28


igloosi:
no, the file is already in .gitignore

Issue #915, opened 2018-07-17
HTML table examples have trailing spaces in class names


zuzak:
The generated HTML on the [Table](https://design-system.service.gov.uk/components/table/) examples have trailing spaces after some of the class names, which they probably shouldn't:

```html
<td class="govuk-table__cell ">£109.80 per week</td>
```

nickcolley:
I believe there's several places this happens, so potentially this should be done consistently.

igloosi:
fixed by https://github.com/alphagov/govuk-frontend/pull/1015

Issue #887, opened 2018-07-12
Make it easier to set widths on table columns


joelanman:
Sometimes you need to set widths on table columns. For example to set a good line width, or to have columns in two tables line up.

You can currently achieve this using the `format` argument but this is not documented

igloosi:
`format` attributes translates into a modifier, so we can't use override classes directly (with adding extra css)
we might have to bite the bullet and allow classes to be added to child elements like we did elsewhere

igloosi:
this is now possible to do with https://github.com/alphagov/govuk-frontend/pull/1015
after the release we can add an example to the Design System

PR #1015, opened 2018-09-28
Allow optional classes on table items

igloosi:
This functionality has been requested a number of times.

- allow classes to be passed to header and row cell items and 
- add additional test to cover new behaviour

Addresses: #887 

Doing this work also addresses: #915


36degrees:
Merging in @igloosi's absence.

PR #1016, opened 2018-10-01
Add character width support to select/textarea

colinrotherham:
I'm building a prototype at the moment where various `input` fields have the standard modifiers:

```css
.govuk-input--width-20
.govuk-input--width-30
```

Until https://github.com/alphagov/govuk-frontend/pull/1013 is fixed, all `select` menus on the page have varying widths based on their contents.

widths

Unfortunately I can't use the width modifier classes on both `select` menus here.

They don't exist!

E.g.

```css
.govuk-select--width-20
.govuk-select--width-30
```

Same with textareas:

```css
.govuk-textarea--width-20
.govuk-textarea--width-30
```

This pull request adds them.

**_measurements.scss**
The standard widths are now in this file
```scss
$govuk-field-widths: (
  30: 56ex + 3ex,
  20: 38ex + 3ex,
  10: 20ex + 3ex,
  5: 10.8ex,
  4: 9ex,
  3: 7.2ex,
  2: 5.4ex
);
```

**_select.scss**
They're looped in the various components they're shared with
```scss
@each $width, $value in $govuk-field-widths {
  .govuk-select--width-#{$width} {
    max-width: $value;
  }
}
```

What do you think?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,16 @@
 
   ([PR #N](https://github.com/alphagov/govuk-frontend/pull/N))
 
+- Add width class modifiers to <select> and <textarea>, for example:
+
+  .govuk-select--width-20
+  .govuk-select--width-30
+
+  .govuk-textarea--width-20
+  .govuk-textarea--width-30
+
+  ([PR #1016](https://github.com/alphagov/govuk-frontend/pull/1016))
+
 🔧 Fixes:
 
 - Pull Request Title goes here
--- a/src/components/input/_input.scss
+++ b/src/components/input/_input.scss
@@ -40,36 +40,9 @@
     border: $govuk-border-width-form-element-error solid $govuk-error-colour;
   }
 
-  // The ex measurements are based on the number of W's that can fit inside the input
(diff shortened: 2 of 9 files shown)


igloosi:
Hi Colin, thanks for raising this. We'll take some time to have a proper look at it.

colinrotherham:
Rebased now https://github.com/alphagov/govuk-frontend/pull/1013 has merged.

dashouse:
Thanks for looking at this, you can use the universal [width override classes](https://design-system.service.gov.uk/styles/spacing/#width-override-classes) on anything including `<select>`, but they are percentage based rather than character based. 

We were considering the fixed width input usage would be in the minority of cases when there is a known length so it would be less likely that you would need to align another form input with a fixed width input.

colinrotherham:
No worries. Feel free to close this if select/textarea shouldn't have the fixed character count widths.

It would be useful for us, but we can support it in our own CSS.

igloosi:
Hi @colinrotherham 
thanks again for taking the time to do the work and open the pull request. As reasoned above, this is something we're currently not looking to implement, so I'll close this.

PR #1018, opened 2018-10-02
Prevent multiple form submissions

quis:
_Although the problem described here is specific to Notify it feels like the solution could be more generally applicable. The code here is a rough first go at adapting it to GOV.UK Frontend. I might not have put it in the right place. It doesn’t have tests. I haven’t even checked that it works._

***

In Notify we’ve seen users complaining that they got sent an invitation email twice. This is probably because someone on their team clicked the ‘send invite’ button twice (even though they think they only clicked it once). Some users will just double click everything on a web page because Windows.

Double form submission is a common issue on web pages, and there are a number of different ways to prevent it. I chose to do it this way because:
- temporarily, not permanently disabling the button means that this addresses the double clicking issue without breaking things if the user did, really want to click the button again deliberately (eg an AJAX request times out and nothing happens)
- doing it with a `data` attribute, rather than the `disabled` attribute  means that the interaction behaviour of the button doesn’t change (`disabled` buttons can’t be focused, for example)

***

Adapted from https://github.com/alphagov/notifications-admin/blob/93e7f9213533790ffb52a8f3afadbace9145eea2/app/assets/javascripts/preventDuplicateFormSubmissions.js

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,29 @@
 
   ([PR #1178](https://github.com/alphagov/govuk-frontend/pull/1178))
 
+- Prevent accidental multiple submissions of forms
+
+  If a user double clicks a submit button in a form, we debounce this event and ignore the second click.
+
+  HTML data attribute:
+
+  ```html
+  <button class="govuk-button" data-prevent-double-click="true">
+    Submit
+  </button>
+  ```
+
+  Nunjucks macro:
+
+  ```js
+  {{ govukButton({
+    text: "Submit",
+    preventDoubleClick: true
+  }) }}
+  ```
+
+  ([PR #1018](https://github.com/alphagov/govuk-frontend/pull/1018))
+
 🔧 Fixes:
 
 - Pull Request Title goes here
--- a/src/components/button/button.js
+++ b/src/components/button/button.js
@@ -11,13 +11,14 @@
 import '../../vendor/polyfills/Event' // addEventListener and event.target normaliziation
 
 var KEY_SPACE = 32
+var DEBOUNCE_TIMEOUT_IN_SECONDS = 1
+var debounceFormSubmitTimer = null
 
 function Button ($module) {
   this.$module = $module
 }
 
 /**
-* Add event handler for KeyDown
 * if the event target element has a role='button' and the event is key space pressed
 * then it prevents the default event and triggers a click event
 * @param {object} event event
@@ -34,11 +35,35 @@ Button.prototype.handleKeyDown = function (event) {
 }
 
 /**
+* If the click quickly succeeds a previous click then nothing will happen.
+* This stops people accidentally causing multiple form submissions by
+* double clicking buttons.
+*/
+Button.prototype.debounce = function (event) {
+  var target = event.target
+  // Check the button that is clicked on has the preventDoubleClick feature enabled
+  if (target.getAttribute('data-prevent-double-click') !== 'true') {
+    return
+  }
+
+  // If the timer is still running then we want to prevent the click from submitting the form
+  if (debounceFormSubmitTimer) {
+    event.preventDefault()
+    return false
+  }
+
+  debounceFormSubmitTimer = setTimeout(function () {
+    debounceFormSubmitTimer = null
+  }, DEBOUNCE_TIMEOUT_IN_SECONDS * 1000)
+}
+
+/**
 * Initialise an event listener for keydown at document level
 * this will help listening for later inserted elements with a role="button"
 */
 Button.prototype.init = function () {
   this.$module.addEventListener('keydown', this.handleKeyDown)
+  this.$module.addEventListener('click', this.debounce)
 }
 
 export default Button
--- a/src/components/button/button.test.js
+++ b/src/components/button/button.test.js
@@ -32,7 +32,7 @@ describe('/components/button', () => {
 
       // we need to start the waitForNavigation() before the keyboard action
       // so we return a promise that is fulfilled when the navigation and the keyboard action are respectively fulfilled
-      // this is somewhat counter intuitive, as we humans expect to act and then wait for something
+      // this is somewhat counter intuitive, as we humans expect to act and then wait for something.
       await Promise.all([
         page.waitForNavigation(),
         page.keyboard.press('Space')
@@ -40,6 +40,86 @@ describe('/components/button', () => {
 
       const url = await page.url()
       expect(url).toBe(baseUrl + href)
+    })
+    describe('preventing double clicks', () => {
+      it('prevents unintentional submissions when in a form', async () => {
+        await page.goto(baseUrl + '/components/button/prevent-double-click/preview', { waitUntil: 'load' })
+
+        // Our examples don't have form wrappers so we need to overwrite it.
+        await page.evaluate(() => {
+          const $button = document.querySelector('button')
+          const $form = document.createElement('form')
+          $button.parentNode.appendChild($form)
+          $button.parentNode.removeChild($button)
+          $form.appendChild($button)
+
+          window.__SUBMIT_EVENTS = 0
+          $form.addEventListener('submit', event => {
+            window.__SUBMIT_EVENTS++
+            // Don't refresh the page, which will destroy the context to test against.
+            event.preventDefault()
+          })
+        })
+
+        await page.click('button')
+        await page.click('button')
+
+        const submitCount = await page.evaluate(() => window.__SUBMIT_EVENTS)
+
+        expect(submitCount).toBe(1)
+      })
+      it('does not prevent intentional multiple clicks', async () => {
+        await page.goto(baseUrl + '/components/button/prevent-double-click/preview', { waitUntil: 'load' })
+
+        // Our examples don't have form wrappers so we need to overwrite it.
+        await page.evaluate(() => {
+          const $button = document.querySelector('button')
+          const $form = document.createElement('form')
+          $button.parentNode.appendChild($form)
+          $button.parentNode.removeChild($button)
+          $form.appendChild($button)
+
+          window.__SUBMIT_EVENTS = 0
+          $form.addEventListener('submit', event => {
+            window.__SUBMIT_EVENTS++
+            // Don't refresh the page, which will destroy the context to test against.
+            event.preventDefault()
+          })
+        })
+
+        await page.click('button')
+        await page.click('button', { delay: 1000 })
+
+        const submitCount = await page.evaluate(() => window.__SUBMIT_EVENTS)
+
+        expect(submitCount).toBe(2)
+      })
+      it('does not prevent multiple submissions when feature is not enabled', async () => {
+        await page.goto(baseUrl + '/components/button/preview', { waitUntil: 'load' })
+
+        // Our examples don't have form wrappers so we need to overwrite it.
+        await page.evaluate(() => {
+          const $button = document.querySelector('button')
+          const $form = document.createElement('form')
+          $button.parentNode.appendChild($form)
+          $button.parentNode.removeChild($button)
+          $form.appendChild($button)
+
+          window.__SUBMIT_EVENTS = 0
+          $form.addEventListener('submit', event => {
+            window.__SUBMIT_EVENTS++
+            // Don't refresh the page, which will destroy the context to test against.
+            event.preventDefault()
+          })
+        })
+
+        await page.click('button')
+        await page.click('button')
+
+        const submitCount = await page.evaluate(() => window.__SUBMIT_EVENTS)
+
+        expect(submitCount).toBe(2)
+      })
     })
   })
 })
--- a/src/components/button/button.yaml
+++ b/src/components/button/button.yaml
@@ -39,6 +39,10 @@ params:
   type: object
   required: false
   description: HTML attributes (for example data attributes) to add to the button component.
+- name: preventDoubleClick
+  type: boolean
+  required: false
+  description: Prevent accidental double clicks on submit buttons from submitting forms multiple times
 
 examples:
 - name: default
@@ -72,6 +76,10 @@ examples:
     element: input
     text: Explicit input button disabled
     disabled: true
+- name: prevent double click
+  data:
+    text: Submit
+    preventDoubleClick: true
 - name: with active state
   description: Simulate triggering the :active CSS pseudo-class, not available in the production build.
   data:
--- a/src/components/button/template.njk
+++ b/src/components/button/template.njk
@@ -16,7 +16,7 @@
 
 {#- Define common attributes we can use for both button and input types #}
 
-{%- set buttonAttributes %}{% if params.name %} name="{{ params.name }}"{% endif %} type="{{ params.type if params.type else 'submit' }}"{% if params.disabled %} disabled="disabled" aria-disabled="true"{% endif %}{% endset %}
+{%- set buttonAttributes %}{% if params.name %} name="{{ params.name }}"{% endif %} type="{{ params.type if params.type else 'submit' }}"{% if params.disabled %} disabled="disabled" aria-disabled="true"{% endif %}{% if params.preventDoubleClick %} data-prevent-double-click="true"{% endif %}{% endset %}
 
 {#- Actually create a button... or a link! #}
 
--- a/src/components/button/template.test.js
+++ b/src/components/button/template.test.js
@@ -218,6 +218,15 @@ describe('Button', () => {
       const $component = $('.govuk-button')
       expect($component.attr('type')).toEqual('button')
     })
+
+    it('renders with preventDoubleClick attribute', () => {
+      const $ = render('button', {
+        preventDoubleClick: true
+      })
+
+      const $component = $('.govuk-button')
+      expect($component.attr('data-prevent-double-click')).toEqual('true')
+    })
   })
 
   describe('implicitly as no "element" param is set', () => {


nickcolley:
I love this idea but I'm worried about the impact it could have.

GOV.UK Frontend is intended to work in any service and we cannot be sure that buttons will always be used in a traditional form context, for example someone may build a number counter component that relies on being able to press 'increase' quickly.

It's also hard to know if this would result in any negative results since if this triggers in the wrong circumstance users are likely to blame themselves.

With this in mind, do you think it's possible to get what you have learned here and turn it into guidance?

I wonder if this is something that could be handled server side? Given that the server could know how quickly requests are coming in.

Failing this we could consider making this an opt in feature for buttons?

What do you think? Do you think I'm being overly cautious?

quis:
In the original implementation this was scoped to only elements with `type="submit"` inside a `<form>`.

Might be sensible to replicate that here because it would prevent this behaviour being applied to:
- links, which should submit idempotent `GET` requests 
- buttons which don’t submit forms but trigger Javascript events (like @nickcolley’s hypothetical number counter)

Thoughts?

nickcolley:
It does seem reasonable to only do this when a button is using type=submit, we should bare in mind that if a user does not supply a type to their button submit is the implicit default.

In terms of detecting if it is in a form I would be careful not to check any process that walks up the DOM is not slow for performance, it might be that this is not necessary...

Interested to hear what others think.

quis:
@nickcolley I think this:

```javascript
var isInsideAForm = event.target.form
```

should be something more like:

```javascript
var isSubmittingAForm = event.target.form && event.target.type === 'submit'
```

(I don’t know if that’s the right property names, but you get the idea)

---

Otherwise I think this is a solid implementation.

nickcolley:
Good shout

I've made an example page to test this: https://output.jsbin.com/deriwub

```js

var DEBOUNCE_TIMEOUT_IN_SECONDS = 0.5
var debounceFormSubmitTimer = null

button.addEventListener('click', function(event) {
    var $button = event.target
    var isSubmitButton = $button.type === 'submit'
    var isInsideAForm = $button.form

    if (!isSubmitButton || !isInsideAForm) {
      return;
    }
 
    if (debounceFormSubmitTimer) {
      event.preventDefault()
      return false;
    }

    debounceFormSubmitTimer = setTimeout(function () {
      debounceFormSubmitTimer = null
    }, DEBOUNCE_TIMEOUT_IN_SECONDS * 1000)
});
```

Needs cross browser testing, will see what other people think.

nickcolley:
Another thought, it feels like there's a few needs here:

1. If I accidentally double click since I'm used to that, don't submit the form
2. If the form takes long and I get impatient, dont submit the form if it's already in progress

I think this only covers the first case @quis maybe some sort of loading pattern would be necessary to satisfy both? Have you seen your complaints go down after you implemented that fix?

joelanman:
Might this ever prevent valid clicks? So if I click but my connection has gone wrong, and then my connection is fixed, I sometimes find I need to click again, this would prevent that?

nickcolley:
@joelanman at the moment the implementation only prevents clicks that happen immediately afterwards, in my example there's a 500millisecond window, Chris' example is a 1.5 second window

joelanman:
ah yes sorry, this PR is just about double clicks, not any click after the first one

quis:
Regarding the duration of the timeout, basing it on how quickly the user expects the UI to respond seems sensible. In other words it shouldn’t interfere with someone who re-clicks because they think that the system hasn’t registered their first click. The generally accepted figure seems to be 1 second<sup>1</sup>.

***

1. https://www.nngroup.com/articles/response-times-3-important-limits/

nickcolley:
@quis so do you think the timeout should be 1.5s as suggested to cover that case?

quis:
@nickcolley I think 1s. Not sure what I based the original 1.5s number on.

nickcolley:
I've been busy recently but I have pushed the work I have done, I think the only thing left to do is write tests for this, so will pick this up when I get some time.

penx:
I don’t disagree with this as a technique but would question whether it should be done by default. 

There may be edge cases where this functionality obstructs functionality - eg clicking a list of delete buttons in quick succession or a medium style “clap” button. 

Perhaps this could only be applied to buttons that explicitly request it, eg with a `data-prevent-doubleclick` attribute or a `js-prevent-doubleclick` class

nickcolley:
> There may be edge cases where this functionality obstructs functionality - eg clicking a list of delete buttons in quick succession or a medium style “clap” button.

In these cases we'd expect users to be using `<button type="button"></button>`.

The idea of having this as an explicit opt-in is interesting, I'll ask the team what they think, thanks. 👍

src/components/button/button.test.js
```diff
@@ -41,5 +41,102 @@ describe('/components/button', () => {
...
+        expect(url).toBe(baseUrl + href)
+      })
+      it('does not trigger on buttons with type=button', async () => {
+        await page.goto(baseUrl + '/components/button/preview', { waitUntil: 'load' })
+
+        // Our examples don't have form wrappers so we need to overwrite it.
+        await page.evaluate(() => {
+          document.querySelector('button').outerHTML = `
+            <form>
+              <button type="button">Save and continue</button>
```
36degrees: It'd be great to have a nicer way of doing this, but in the short term this does the job nicely 👍 

I wonder if we should wrap the existing button from the page rather than replace it though – there's a chance we could change the markup and break this functionality, but it wouldn't be picked up by the test because it'd be hardcoded to use the old markup.

CHANGELOG.md
```diff
@@ -20,6 +20,10 @@
 
   ([PR #N](https://github.com/alphagov/govuk-frontend/pull/N))
 
+- Prevent multiple form submissions
```
36degrees: I think we might want to be more specific here – we're really debouncing form submission so that you can't submit more than once a second, which isn't quite the same thing this implies.

src/components/button/button.js
```diff
@@ -11,6 +11,8 @@
 import '../../vendor/polyfills/Event' // addEventListener and event.target normaliziation
 
 var KEY_SPACE = 32
+var DEBOUNCE_TIMEOUT_IN_SECONDS = 1
+var debounceFormSubmitTimer = null
```
36degrees: Is there anything weird that could happen here if you had two forms / submit buttons on one page? Would it be better to scope this to the submitted form somehow?

src/components/button/button.js
```diff
@@ -33,12 +35,47 @@ Button.prototype.handleKeyDown = function (event) {
   }
 }
 
+/**
+* Add event handler for click
```
36degrees: Slightly pedantic, but this doesn't _add_ the event handler – it is the event handler.

src/components/button/button.test.js
```diff
@@ -41,5 +41,102 @@ describe('/components/button', () => {
...
+    describe('debouncing', () => {
+      it('does not block button links', async () => {
+        // Links should take a user to another part of the page or url, so debouncing it wouldnt have any impact.
+        // But we can check that we're not blocking any initial clicks.
+
+        await page.goto(baseUrl + '/components/button/link/preview', { waitUntil: 'load' })
+
+        const href = await page.evaluate(() => document.body.getElementsByTagName('a')[0].getAttribute('href'))
+
+        // we need to start the waitForNavigation() before the keyboard action
```
36degrees: keyboard action?

src/components/button/button.test.js
```diff
@@ -41,5 +41,102 @@ describe('/components/button', () => {
...
+
+        // Our examples don't have form wrappers so we need to overwrite it.
+        await page.evaluate(() => {
+          document.querySelector('button').outerHTML = `
+            <form>
+              <button type="button">Save and continue</button>
+            </form>
+          `
+
+          window.__BUTTON_PRESSED_COUNT = 0
```
36degrees: To make the intent super clear, I'd consider naming this something like `__BUTTON_CLICK_EVENTS` as we're counting the number of events fired, not the number of times the button was actually pressed?

src/components/button/button.test.js
```diff
@@ -41,5 +41,102 @@ describe('/components/button', () => {
...
+        })
+
+        await page.click('button')
+        await page.click('button')
+
+        const submitCount = await page.evaluate(() => window.__SUBMIT_COUNT)
+
+        expect(submitCount).toBe(1)
+      })
+      it('when a user clicks again intentionally it is not prevented', async () => {
```
36degrees: Again to make the intent really clear here (what does intentionally mean?) perhaps something like '(it) does not prevent additional clicks after one second'?

src/components/button/button.test.js
```diff
@@ -41,5 +41,102 @@ describe('/components/button', () => {
       const url = await page.url()
       expect(url).toBe(baseUrl + href)
     })
+    describe('debouncing', () => {
+      it('does not block button links', async () => {
```
36degrees: These tests are great 👍

src/components/button/button.js
nickcolley: While there's a single timer for all buttons I don't think it's humanly possible to click two buttons quick enough for it to matter?

🏃

src/components/button/button.test.js
nickcolley: recycled code 😬

src/components/button/button.js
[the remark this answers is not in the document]
nickcolley: Going through this scenario out loud...

From the server:

```html
<form action="/delete/1/">
  <button type="submit">Delete 1</button>
</form>
```

When JavaScript enhances the page, I suppose you might turn this into a "button" type, but more likely you'd just prevent the default submission.

```html
<form action="/delete/1/">
  <button type="button">Delete 1</button>
</form>
```

```js
document.querySelector('form').addEventListener('submit', event => {
  // Do an AJAX request
  // ...
  event.preventDefault()
})
```

edwardhorsford:
👋 I really like the spirit of this - I've seen quite a few services where the continue button goes disabled when you click it - I presume because the team / framework is disabling the button on submission. Besides the accessibility issues of disabled buttons I generally find it distracting.

---

With that said, I do wonder if this should be the concern of the server to deal with. As others have said, there may be cases where you do want to send data quickly, or connections might time out. Only the server knows what's actually been received - I'd hope there are frontend architecture patterns for recognising multiple submissions in a reliable way.

---

The server solution presumably being harder - if we can provide built-in solutions so we avoid teams having disabled buttons - all the better.

src/components/button/button.js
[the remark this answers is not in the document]
nickcolley: So I suppose there's definitely a chance that, you could want to rattle through deleting lots of things...

nickcolley:
@edwardhorsford one alternative is we turn this into guidance with a recommendation to do this server side for sure, I'm personally a little uncomfortable with how broad this JavaScript could apply based on @36degrees and @penx 's feedback.

src/components/button/button.js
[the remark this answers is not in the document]
nickcolley: If we wanted to have multiple timers we could change how this module works from

```
new Button(document).init()
```

to (pseudo code)

```js
Array.from(document.querySelector('button, input=button, [role=button]')).forEach($button => {
  new Button($button).init()
})
```

Then the timer instance could be stored on the button instance itself.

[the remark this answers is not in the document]
penx: alternative markup to consider coming from the server, not sure if this would affect your thinking

```
<form action="/delete">
  Item 1 <button name="id" value="1" type="submit">Delete 1</button>
  Item 2 <button name="id" value="2" type="submit">Delete 2</button>
  Item 3 <button name="id" value="3" type="submit">Delete 3</button>
</form>
```

I think it would be reasonable to support deleting multiple items without waiting for a response between each button click for both JS and non-JS users.

kellylee-gds:
@timpaul @36degrees @nickcolley agreed in planning we would book in a time boxed session to discuss this

nickcolley:
@timpaul @36degrees and I have chatted about this and we're going to resolve the remaining worries by doing the following:

1. Make this a opt-in feature
2. Add guidance to the design system, component page on how to enable this feature, and when.

We're going to try to have timers that do not conflict but if it's not possible we believe that having this as an opt-in feature will remove some of the risk.

nickcolley:
Think I need to pair with someone on it to talk through the compromises of the current approach vs refactored.

I've pushed up the less controversial version and the refactored version https://github.com/alphagov/govuk-frontend/compare/debounce-button-refactor?expand=1

The feature is called 'debounce' which can be changed to whatever we think is clearest for users.

nickcolley:
```html
<button class="govuk-button" data-prevent-double-click="true">
  Submit
</button>
```

```js
{{ govukButton({
  text: "Submit",
  preventDoubleClick: true
}) }}
```

src/components/button/button.js
```diff
@@ -33,12 +34,52 @@ Button.prototype.handleKeyDown = function (event) {
...
 
+/**
+* If the click quickly succeeds a previous click then nothing will happen.
+* This stops people accidentally causing multiple form submissions by
+* double clicking buttons.
+*/
+Button.prototype.debounce = function (event) {
+  var target = event.target
+  // Check the button that is clicked on has the preventDoubleClick feature enabled
+  if (target.attributes && target.getAttribute('data-prevent-double-click') !== 'true') {
```
36degrees: If the target doesn't have any attributes (not sure how likely that is), this won't evaluate the second statement. I think it needs to be flipped:

```suggestion
  if (!target.attributes || target.getAttribute('data-prevent-double-click') !== 'true') {
```

src/components/button/button.js
```diff
@@ -33,12 +34,52 @@ Button.prototype.handleKeyDown = function (event) {
...
+
+  // Only handle events on <input> and <button> elements
+  var tagName = event.target.tagName
+  if (tagName !== 'INPUT' && tagName !== 'BUTTON') {
+    return
+  }
+
+  var $button = event.target
+
+  // We only want to to handle submit buttons that are used in forms
```
36degrees: If this is now an opt-in behaviour, do we need to be as strict here?

If users add `data-prevent-double-click`, I think they'd expect it to work, regardless of whether it's in a form / a submit button?

src/components/button/button.js
36degrees: Actually, I think getAttribute is generally safe to use, and will just return `null` if the attribute isn't set, so not sure the first check is necessary? In which case this could just be

```suggestion
  if (target.getAttribute('data-prevent-double-click') !== 'true') {
```

src/components/button/button.test.js
```diff
@@ -41,5 +41,128 @@ describe('/components/button', () => {
       const url = await page.url()
       expect(url).toBe(baseUrl + href)
     })
+    describe('preventing double clicks', () => {
+      it('does not block button links', async () => {
+        // Links should take a user to another part of the page or url, so debouncing it wouldnt have any impact.
+        // But we can check that we're not blocking any initial clicks.
+
+        await page.goto(baseUrl + '/components/button/link/preview', { waitUntil: 'load' })
```
36degrees: The way the logic is currently written, this test is never going to reach the part of the function that tests that the element is an `input` / `button`, because it'll already have returned because it doesn't have `data-prevent-double-click`.

What are we trying to test here? Do we need to be testing clicking a link _that has `data-prevent-double-click`_ set to `true`? Or is this test now redundant?

src/components/button/button.test.js
```diff
@@ -41,5 +41,128 @@ describe('/components/button', () => {
...
+        // this is somewhat counter intuitive, as we humans expect to act and then wait for something.
+        await Promise.all([
+          page.waitForNavigation(),
+          await page.click('a[role="button"]')
+        ])
+
+        const url = await page.url()
+        expect(url).toBe(baseUrl + href)
+      })
+      it('does not trigger on buttons with type=button', async () => {
```
36degrees: Same thing applies here – this is just going to be returning because of the lack of `data-prevent-double-click`

src/components/button/button.yaml
```diff
@@ -39,6 +39,10 @@ params:
   type: object
   required: false
   description: HTML attributes (for example data attributes) to add to the button component.
+- name: preventDoubleClick
+  type: boolean
+  required: false
+  description: Prevent accidental double clicks from submitting forms multiple times
```
36degrees: Should we be clear that this doesn't affect links?

src/components/button/button.js
nickcolley: I've been guarding them since if you try to use 'getAttribute' on document it will throw.

Not sure this makes sense here though.

src/components/button/button.js
nickcolley: Yeah could probably simplify this now...

src/components/button/button.test.js
```diff
@@ -41,5 +41,59 @@ describe('/components/button', () => {
...
+        })
+
+        await page.click('button')
+        await page.click('button')
+
+        const submitCount = await page.evaluate(() => window.__SUBMIT_EVENTS)
+
+        expect(submitCount).toBe(1)
+      })
+      it('does not prevent intentional multiple clicks', async () => {
```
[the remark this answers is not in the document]
nickcolley: Fair

adamsilver:
Will this also stop users from submitting the form twice by pressing enter within a form field?

nickcolley:
@adamsilver no it does not.

nickcolley:
Thanks everyone that contributed to this, took us a long time to get it right but I'm happy with what we've managed here.

Special thanks to @quis for the original contribution.

PR #1020, opened 2018-10-03
Stop links styled as button from being dragged

quis:
In user research we have seen users’ attempts at clicking links styled as buttons not registered because they moved the mouse as they were clicking. This instead triggers the browsers drag behaviour.

d01

***

One user we were observing yesterday repeatedly encountered this problem and it took several attempts before they got to the next page. I reckon this is a combination of:
- having impaired motor skills
- being more familiar with a trackpad and touch screen

This only happens with links. With `<button>` elements you can move the mouse between depressing and releasing the button and a `click` event is still registered (as long as the cursor remains inside the button).

d02

***

Setting the `draggable` attribute<sup>1</sup> to `false` prevents the drag behaviour from being triggered. This makes links styled as buttons behave in the same way as buttons<sup>2</sup>.

d03

***

1. `draggable` is supported in all current browsers: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/draggable#Browser_compatibility

2. Buttons still have the problem of not registering a click if you drag outside the button’s area before releasing the mouse. It would be possible to use Javascript to trigger a `click` event even if the mouse left the region of the link or button before being released. This would be a more comprehensive fix, but would also involve significantly more engineering effort and potential side effects.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@
 🆕 New features:
 
 - Allow classes on table header and row cells
-  
+
   Optional classes attribute can now be used on table header and row cell item
   in the Nunjucks macro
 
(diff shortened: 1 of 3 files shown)


quis:
@nickcolley I’ve done the `CHANGELOG.md` and `README.md`

PR #1023, opened 2018-10-05
Set text colour for radios divider

kevindew:
Hello :wave:, I've enjoyed digging into some bits of GOV.UK Frontend over this firebreak, and I think I found a bug (or I don't properly understand isolation principles)

When testing component isolation on GOV.UK we noticed that text colour isn't applied to a radios_divider and instead it inherits this value. I'm assuming this behaviour isn't intentional as all the neighbouring fieldset components all have a colour set.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,10 @@
 
   ([PR #1007](https://github.com/alphagov/govuk-frontend/pull/1007))
 
+- Set text colour for radios divider
+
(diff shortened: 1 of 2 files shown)


kevindew:
Thanks @36degrees fyi I don't have write access

36degrees:
Sorry, I should have said – for contributions outside of the team we require reviews from two different people, so this'll get merged once someone else has had a chance to review it 🙂

Issue #1030, opened 2018-10-12
Anchor links inside tabs don't work


tijmenb:
When you click on an anchor link inside a tab, the first tab will be selected instead of the browser scrolling to the correct place in the page. Tested in Chrome & Firefox.

![](https://media.giphy.com/media/pyzHpxkGvviK0kh9oo/giphy.gif) 

## Steps to reproduce

1. Put a anchor link and anchor in a tab panel
2. Click the link
3. The tab is now selected
4. Click anchor link again
5. Browser now correctly scrolls to anchor

Glitch: 

https://govuk-frontend-tab-bug.glitch.me/

Code:

https://glitch.com/edit/#!/govuk-frontend-tab-bug?path=src/views/index.html:1:0

nickcolley:
This looks to be based on the implementation of the hash handling: https://github.com/alphagov/govuk-frontend/blob/master/src/components/tabs/tabs.js#L106

PR #1031, opened 2018-10-15
Fix anchor links in tabs component panels

nickcolley:
Fixes https://github.com/alphagov/govuk-frontend/issues/1030

When clicking an anchor in the tabs component panel, it will trigger the 'hashchange' event which the tabs component is listening to.
The component then checks to see if there's any tabs within the component that match this hash, but did so in a way that could also include regular anchor links.

This pull request ensures it is only checking that the hash matches available tabs only.

I've also tested the following scenarios manually:

1. Editing the URL directly
2. Clicking the anchor with no existing hash
3. Clicking the anchor when a hash already exists.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,10 @@
 
 🔧 Fixes:
 
+- Fix anchor links in tabs component panels
+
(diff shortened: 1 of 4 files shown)


alex-ju:
@nickcolley thanks a lot for looking into this! 🙏

PR #1038, opened 2018-10-18
Add optional Welsh language support to the Character Count component

robertpataki:
Add optional Welsh language support to the new Character Count component to comply with the UK Government language [guidelines](https://www.gov.uk/guidance/content-design/welsh-language-on-gov-uk).

### DIFF
--- a/package/components/character-count/macro-options.json
+++ b/package/components/character-count/macro-options.json
@@ -79,5 +79,11 @@
         "type": "object",
         "required": false,
         "description": "HTML attributes (for example data attributes) to add to the textarea."
+    },
+    {
+        "name": "welsh",
+        "type": "boolean",
+        "required": false,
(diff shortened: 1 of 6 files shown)


robertpataki:
Good morning all, I was just wondering - could someone please have a look at my PR? Thanks!

36degrees:
Hi @robertpataki,

This is a totally valid approach which solves your particular use case, but we need to find a way to support localisation consistently in a way that will work for different types of components and for other languages. For that reason, we’re not going to be able to accept this pull request.

We hope that we’ll be able to look properly at localisation early next year. In the mean time, we suggest that you fork and vendor in this component and make the adjustments in the service you’re working on.

Thanks for taking the time to propose this, and for your patience.

robertpataki:
Hey @36degrees, makes sense, thanks!

PR #1043, opened 2018-10-25
Allow form group classes on radios and checkboxes

igloosi:
## What

Make it possible to toggle the `govuk-form-group--error` class on the form group for radios and checkboxes without having to pass an error message.

## Why

Currently the error state for a conditionally revealed checkbox shows the 'error bar' against the revealed content only.
68747470733a2f2f7472656c6c6f2d6174746163686d656e74732e73332e616d617a6f6e6177732e636f6d2f3539616436653537653761636164313862643863623437362f3562613463663131396438346530343934353038353766352f6437336264323764623664363464363035626634643330316631

We want to move the error bar to the 'question' which will include the revealing checkboxes or radios themselves.

To do this we have in introduced a `formGroupClasses' option in the Nunjucks macro that allows a CSS class to be set directly on the form group.

Example of an error state set on the whole group

screen shot 2018-10-25 at 09 52 31
screen shot 2018-10-25 at 09 52 47

Trello ticket: https://trello.com/c/LggtR8s5/1455-make-it-possible-to-set-the-error-state-on-checkboxes-and-radios-without-passing-an-error-message

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,11 +27,11 @@
 
   ([PR #1045](https://github.com/alphagov/govuk-frontend/pull/1045))
 
-- Pull Request Title goes here
+- Allow form group classes on radios and checkboxes
 
-  Description goes here (optional)
+  We now provide a way to add classes to the radio and checkbox form-group wrapper
 
-  ([PR #N](https://github.com/alphagov/govuk-frontend/pull/N))
+  ([PR #1043](https://github.com/alphagov/govuk-frontend/pull/1043))
 
 🔧 Fixes:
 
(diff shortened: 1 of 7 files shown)


nickcolley:
I think if we want to add this potentially we want to make sure you can set formGroupClasses anywhere that formGroups are used? Or are these the only places?

Just thinking people will expect this to be consistent across components.

36degrees:
Are there other things that we can foresee users wanting to add form-groups in the future, like attributes? Just wondering if we should opt for `formGroup: { classes: '' }` instead.

I also wonder whether this is one area where providing a boolean flag (e.g. `error: true`) might be helpful (we may want both).

igloosi:
`formGroup: { classes: '' }` is more flexible, though more verbose. Not sure what else whey'd want to add to the form group, but i'm probably mistaken.

would error :true additional option be more confusing? what if you want both classes and the error state?

igloosi:
rewritten to accept  `formGroup: { classes: '' }` way of passing options to the macro

PR #1045, opened 2018-10-29
Add support for attributes on table cells

jonheslop:
It’s currently not possible to attach HTML attributes to table cells.
This PR makes it possible by using the same object pattern that is used
on other components (and also the `<table>` element).

I’ve updated the nunjucks template as well as the YAML and the README
to reflect the new functionality.

### Background

Whilst migrating https://github.com/alphagov/pay-selfservice I’ve tried to use macros where ever possible, but we regularly use ID’s on cells within table so that our various tests can easily get the value.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,13 @@
 
   ([PR #1037](https://github.com/alphagov/govuk-frontend/pull/1037))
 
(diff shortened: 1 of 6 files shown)


jonheslop:
Yay! Thanks :D

PR #1050, opened 2018-11-01
Add SRI

### LINKED ISSUE #1049: Use SRI for external scripts
External resources should use SubResource Integrity hashes to make sure that a malicious actor can't change their contents.

See https://www.srihash.org/

Re https://github.com/alphagov/govuk-frontend/blob/master/app/views/layouts/component-preview.njk#L16 & https://github.com/alphagov/govuk-frontend/blob/master/app/views/layouts/_generic.njk#L11

edent:
Fixes #1049

### DIFF
--- a/app/views/layouts/_generic.njk
+++ b/app/views/layouts/_generic.njk
@@ -9,8 +9,8 @@
 
   <!--[if !IE 8]><!-->
     <link rel="stylesheet" href="/public/app.css">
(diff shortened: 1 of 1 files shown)


edent:
If it is an internal app, that makes it an even juicier target :-)

But, yes, the hashes should be automatically generated.

PR #1054, opened 2018-11-05
Single field with error should have 'aria-describeby' attribute

aliuk2012:
## What

If you use a single checkbox without a `<fieldset>`, and that single checkbox has an error message or hint, the error or hint will not associated with the field – they would normally be associated with the `fieldset`, but that isn't part of the output because it's a single checkbox in isolation.

This PR adds the describeBy to the checkbox input if no fieldset exists

Example page: https://govuk-frontend-review-pr-1054.herokuapp.com/components/checkboxes/with-single-option-set-aria-describeby-on-input/preview

Trello ticket: https://trello.com/c/05zGTIPN/1453-associate-error-message-and-hint-with-checkbox-if-there-is-no-fieldset

### DIFF
--- a/src/components/checkboxes/checkboxes.yaml
+++ b/src/components/checkboxes/checkboxes.yaml
@@ -220,6 +220,16 @@ examples:
       - value: blue
         text: Blue
 
+- name: with single option set 'aria-describeby' on input
+  readme: false
+  data:
+    name: t-and-c
+    errorMessage:
+      text: Please accept the terms and conditions
+    items:
+      - value: yes
+        text: I agree to the terms and conditions
+
 - name: with all fieldset attributes
   data:
     idPrefix: example
--- a/src/components/checkboxes/template.njk
+++ b/src/components/checkboxes/template.njk
@@ -18,6 +18,9 @@
   {% endif %}
 {% endfor %}
 
+{#- fieldset is false by default -#}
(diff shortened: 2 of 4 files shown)


src/components/checkboxes/template.njk
```diff
@@ -18,6 +18,9 @@
   {% endif %}
 {% endfor %}
 
+{#- fieldset is false by default -#}
```
nickcolley: Non-blocking: I think we could remove these comments, or make them more descriptive what the logic is doing.

src/components/checkboxes/checkboxes.yaml
```diff
@@ -220,6 +220,15 @@ examples:
       - value: blue
         text: Blue
 
+- name: with single option set aria-describeby on input
```
nickcolley: Non-blocking: 'aria-describedby'

nickcolley: You can add `readme: false` to your example, since it's only needed for the test.

src/components/checkboxes/template.njk
aliuk2012: I added this comment only because it was not obvious at first that the component would have no fieldset by default.

igloosi:
changelog entry?

Issue #1039, opened 2018-10-22
Write integration tests for Character count component


nickcolley:

nickcolley:
@alex-ju could you push up your branch where you had WIP for this please?

alex-ju:
@nickcolley it seems that I've lost the tests in some branch 😞. Will try and write them today they were exhaustive anyway. Apologies again for not including them from the beginning.

aliuk2012:
I'm closing this issue as #1055 has been merged.

PR #1055, opened 2018-11-05
Add tests for character-count script

alex-ju:
This PR contains a basic set of tests for the character count component that were mistakenly missed from the [initial PR](https://github.com/alphagov/govuk-frontend/pull/959) by author's fault.

Issue: #1039

### DIFF
--- a/src/components/character-count/character-count.test.js
+++ b/src/components/character-count/character-count.test.js
@@ -1,0 +1,95 @@
+/**
+ * @jest-environment ./lib/puppeteer/environment.js
+ */
+/* eslint-env jest */
+
+const configPaths = require('../../../config/paths.json')
+const PORT = configPaths.ports.test
+
+let browser
+let page
+let baseUrl = 'http://localhost:' + PORT
+
+beforeEach(async () => {
+  browser = global.__BROWSER__
+  page = await browser.newPage()
+})
+
+afterEach(async () => {
+  await page.close()
+})
+
+describe('Character count', () => {
+  describe('when JavaScript is unavailable or fails', () => {
+    it('shows the static message', async () => {
+      await page.setJavaScriptEnabled(false)
+
+      await page.goto(baseUrl + '/components/character-count/preview', { waitUntil: 'load' })
+      const message = await page.$eval('.govuk-character-count__message', el => el.innerHTML.trim())
+
+      expect(message).toEqual('You can enter up to 10 characters')
+    })
+  })
+  describe('when JavaScript is available', () => {
+    it('shows the dynamic message', async () => {
+      await page.goto(baseUrl + '/components/character-count/preview', { waitUntil: 'load' })
+
+      const message = await page.$eval('.govuk-character-count__message', el => el.innerHTML.trim())
+
+      expect(message).toEqual('You have 10 characters remaining')
+    })
+
+    it('counts down while typing and within limit', async () => {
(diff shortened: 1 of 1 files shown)


src/components/character-count/character-count.test.js
```diff
@@ -0,0 +1,83 @@
...
+
+      expect(message).toEqual('You have 9 characters remaining')
+    })
+
+    it('counts up while typing and the limit is exceeded', async () => {
+      await page.goto(baseUrl + '/components/character-count/preview', { waitUntil: 'load' })
+
+      // Press 11 characters
+      await page.focus('.js-character-count')
+      for (let i = 0; i < 11; i++) {
```
36degrees: Not blocking: would `page.keyboard.type('A'.repeat(11))` have the same effect?

src/components/character-count/character-count.test.js
```diff
@@ -0,0 +1,83 @@
...
+  await page.close()
+})
+
+describe('Character count', () => {
+  describe('when JavaScript is unavailable or fails', () => {
+    it('shows the static message', async () => {
+      await page.setJavaScriptEnabled(false)
+
+      await page.goto(baseUrl + '/components/character-count/preview', { waitUntil: 'load' })
+      const message = await page.evaluate(() => document.body.querySelector('.govuk-character-count__message').innerHTML.trim())
```
36degrees: Non-blocking: Lines like this are very long, can we break them up?

```suggestion
      const message = await page.evaluate(() => {
        return document.body.querySelector('.govuk-character-count__message').innerHTML.trim()
      })
```

Alternatively, it looks like we might be able to use `page.$eval`? (not tested)

```suggestion
      const message = await page.$eval('.govuk-character-count__message', el => el.innerHTML.trim())
```

src/components/character-count/character-count.test.js
```diff
@@ -0,0 +1,83 @@
...
+  browser = global.__BROWSER__
+  page = await browser.newPage()
+})
+
+afterEach(async () => {
+  await page.close()
+})
+
+describe('Character count', () => {
+  describe('when JavaScript is unavailable or fails', () => {
```
36degrees: 👌 Great to see tests for this.

src/components/character-count/character-count.test.js
```diff
@@ -0,0 +1,83 @@
...
+      // Press 11 characters
+      await page.focus('.js-character-count')
+      for (let i = 0; i < 11; i++) {
+        await page.keyboard.press('A')
+      }
+
+      const textareaClasses = await page.evaluate(() => document.body.querySelector('.js-character-count').className)
+      const messageClasses = await page.evaluate(() => document.body.querySelector('.govuk-character-count__message').className)
+
+      expect(textareaClasses).toContain('govuk-textarea--error')
```
36degrees: Non-blocking: Would prefer one assertion per test.

alex-ju:
@36degrees thanks. will update with your suggestions. `.repeat(11)` feels better 😅

src/components/character-count/character-count.test.js
36degrees: Just tried `page.$eval` out in #1056 and it seems to work well for things like this 👍

src/components/character-count/character-count.test.js
alex-ju: Repeat doesn't seem to be working, unfortunately, so I'll have to leave the form.

PR #1056, opened 2018-11-05
Scroll to label or legend when linked from error summary

36degrees:
By default, the browser will scroll the target into view. Because our labels or legends appear above the input, this means the user will be presented with an input without any context, as the label or legend will be off the top of the screen.

screen shot 2018-11-14 at 12 51 49

Manually handling the click event, focussing the element and scrolling the question into view solves this.

screen shot 2018-11-14 at 12 52 51

## To do

- [x] Test in all supported browsers
  - [x] Internet Explorer 8 (Windows)
  - [x] Internet Explorer 9 (Windows)
  - [x] Internet Explorer 10 (Windows)
  - [x] Internet Explorer 11 (Windows)
  - [x] Edge (latest 2 versions) (Windows)
  - [x] Google Chrome (latest 2 versions) (Windows)
  - [x] Mozilla Firefox (latest 2 versions) (Windows)
  - [x] Safari 9+ (macOS)
  - [x] Google Chrome (latest 2 versions) (macOS)
  - [x] Mozilla Firefox (latest 2 versions) (macOS)
  - [x] Safari 9.3+ (iOS)
  - [x] Google Chrome (latest 2 versions) (iOS)
  - [x] Google Chrome (latest 2 versions) (Android)
  - [x] Samsung Internet (latest 2 versions) (Android)

- [x] Test in assistive technologies
  - [x] JAWS + Internet Explorer 11
  - [x] ZoomText + Internet Explorer 11
  - [x] Dragon NaturallySpeaking + Internet Explorer 11
  - [x] NVDA + Firefox
  - [x] VoiceOver (iOS)
  - [x] VoiceOver (macOS)

### DIFF
--- a/src/components/error-summary/error-summary.js
+++ b/src/components/error-summary/error-summary.js
@@ -1,4 +1,6 @@
+import '../../vendor/polyfills/Function/prototype/bind'
 import '../../vendor/polyfills/Event' // addEventListener
+import '../../vendor/polyfills/Element/prototype/closest'
 
 function ErrorSummary ($module) {
   this.$module = $module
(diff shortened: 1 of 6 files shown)


src/components/error-summary/error-summary.js
```diff
@@ -12,6 +14,103 @@ ErrorSummary.prototype.init = function () {
...
+  }
+
+  var $legendOrLabel = this.getAssociatedLegendOrLabel($input)
+  if (!$legendOrLabel) {
+    return false
+  }
+
+  window.location.hash = inputId
+  $input.focus({ preventScroll: true })
+  $legendOrLabel.scrollIntoView()
```
nickcolley: I was surprised this would be possible with standard web apis cool!

src/components/error-summary/error-summary.js
```diff
@@ -12,6 +14,103 @@ ErrorSummary.prototype.init = function () {
...
+
+  $module.addEventListener('click', this.onClick.bind(this))
+}
+
+/**
+* Click event handler
+*
+* @param {MouseEvent} event - Click event
+*/
+ErrorSummary.prototype.onClick = function (event) {
```
nickcolley: Non-blocking: I think we normally try to use 'handleClick', like the React convention, isn't written down so is OK if you ignore this.

src/components/error-summary/error-summary.js
```diff
@@ -1,4 +1,6 @@
+import '../../vendor/polyfills/Function/prototype/bind'
 import '../../vendor/polyfills/Event' // addEventListener
+import '../../vendor/polyfills/Element/prototype/closest' // addEventListener
```
nickcolley: Is this `// addEventListener` comment correct?

36degrees: Nope 😬

src/components/error-summary/error-summary.test.js
```diff
@@ -28,4 +28,44 @@ describe('Error Summary', () => {
...
+    ['a file upload', 'file', 'label[for="file"]'],
+    ['a group of radio buttons', 'radios', '#test-radios legend'],
+    ['a group of checkboxes', 'checkboxes', '#test-checkboxes legend'],
+    ['a single checkbox', 'single-checkbox', 'label[for="single-checkbox"]'],
+    ['a conditionally revealed input', 'yes-input', '#test-conditional-reveal legend']
+  ]
+
+  describe.each(inputTypes)('when linking to %s', async (_, inputId, legendOrLabelSelector) => {
+    beforeAll(async () => {
+      await page.goto(`${baseUrl}/examples/error-summary`, { waitUntil: 'load' })
```
36degrees: Not super-keen on hooking into the review app's examples, but couldn't think of a better way of doing this at the minute. This is something I'd definitely to revisit in the future.

nickcolley: Yeah this is an issue with all our examples like this.

igloosi:
IE11 + Jaws announces for example:
"Conditional?  "text input edit" "Problem with input type of text"

igloosi:
Zoom Text + IE11 doesn't move up to label on 
Checkboxes and Conditional revealing content
image
image

igloosi:
Samsung Internet on Android scrolls to Conditional reveal label, opens keyboard and scrolls just down a little. Others, including text input are ok.
image

36degrees:
There is still 'jumpy' scrolling in IE8 and IE9, where clicking the link takes you first to the input and then very quickly 'back up' to the legend or label:

nov-13-2018 15-33-21

However, I've been able to fix this in IE10 and above 🎉 

Whilst it's not ideal, I think it is acceptable for IE8 and IE9 to do this.

Alternatively, we could choose to do nothing in browsers that don't support history.pushState, which would mean in IE8 and IE9 you'd end up jumping to the input with the legend or label off the top of the screen.

src/components/error-summary/error-summary.js
```diff
@@ -12,6 +14,115 @@ ErrorSummary.prototype.init = function () {
...
+    return false
+  }
+
+  // Prefer using the history API where possible, as updating
+  // window.location.hash causes the viewport to jump to the input briefly
+  // before then scrolling to the label/legend in IE10, IE11 and Edge (as tested
+  // in Edge 17).
+  if (window.history.pushState) {
+    window.history.pushState(null, null, '#' + inputId)
+  } else {
```
nickcolley: One alternative might be to avoid setting a hash at all, I'm not sure what the user need is?

36degrees: TBH, it's just maintaining the existing behaviour.

36degrees:
It looks like this change also massively improves things for NVDA users:

## Text Input

Before: "Edit, has autocomplete"
After: "Label for input, edit, has autocomplete, problem with input, blank"

## Textarea

Before: "Edit, multiline"
After: "Label for textarea, edit, multiline, problem with input, blank"

## Select

Before: "Combobox, collapsed, option one."
After: "Label for select, Combobox, option one, collapsed, problem with select"

## Select

Before: "Spin button, edit"
After: "Label for date input, grouping, problem with date input. Day, spin button, edit, blank"

## File Upload

Before: "Browse, no file selected"
After: "Browse, button"

## Radios

Before: "Radio button, not checked"
After: "Legend for radios, grouping, problem with radios. Label for first radio button, radio button not checked, 1 of 2"

## Checkboxes

Before: "Checkbox, not checked"
After: "Legend for checkboxes, grouping. Label for first checkbox, checkbox not checked."

## Single checkbox

Before: "Checkbox, not checked"
After: "Label for single checkbox, checkbox not checked, problem with single checkbox"

36degrees:
I'm happy that this is good to review now 👍

src/components/error-summary/error-summary.js
colinrotherham: @nickcolley Using the hash means you can press "back" and be jumped up to the top of the page again. This is the current behaviour, I've seen people do this—it's pretty useful.

@36degrees For older browsers, could you use a timeout to allow _scrollIntoView()_ to complete before changing the hash? It likely won't scroll again as the input is already in view.

36degrees: @colinrotherham good call, but unfortunately changing the hash does cause IE8 and IE9 to scroll the target input to the top of the viewport, even if it is already in the viewport.

nickcolley:
Glad to see this get done, the NVDA improvements are awesome.

nickcolley:
Needs a changelog entry

36degrees:
Bugs filed as a result of this work:

- [NVDA does not announce label, legend or description when following a link to an input](https://github.com/nvaccess/nvda/issues/6209#issuecomment-428116805) (NVDA, Firefox, Windows)
- [Legend not announced when following a link to an input within a fieldset using Ctrl-Option-Space](https://bugs.webkit.org/show_bug.cgi?id=191742) (Voiceover, Safari, macOS)
- [Voiceover reads checkbox labels twice](https://bugs.webkit.org/show_bug.cgi?id=191743) (Voiceover, Safari, macOS)

colinrotherham:
Linking to this ticket https://github.com/alphagov/govuk-design-system-backlog/issues/155 so it can be documented how the error summary links to each form control by ID (not a legend/label).

I've found a mix of behaviours across our services (some linking to empty links `<a href="#something"></a>`) which we need to correct. This work is fab.

36degrees:
@colinrotherham there's a PR open on the Design System which adds guidance around this to the Error Summary – if you fancy checking it out, it'd be great to get feedback from someone outside the team.

https://github.com/alphagov/govuk-design-system/pull/634

PR #1057, opened 2018-11-05
Remove options from component READMEs

hannalaakso:
Remove tables of arguments from the component READMEs in GOV.UK Frontend, replacing them with links to the new tables of arguments in the Design System.

This uses `componentGuidanceLink` for setting the links to the options table in the Design System (takes care of the `input/text-input` difference).

`hint` and `label` don't have their own pages in the Design System so (for the want of a better solution) this links to another component page that displays their options.

https://trello.com/c/MarNP5Fl/723-1-remove-tables-of-arguments-from-frontend-readmes


src/components/hint/README.njk
```diff
@@ -14,79 +14,5 @@
...
-      {
-        text: 'No'
-      },
-      {
-        text: 'Any extra HTML attributes (for example data attributes) to add to the hint span tag.'
-      }
-    ]
-  ]
-})}}
+See <a href="https://design-system.service.gov.uk/components/file-upload/#options-example-default--hint">options table</a> for details.
```
nickcolley: I feel like the most vanilla hint would be using https://design-system.service.gov.uk/components/text-input/#options-example-default--hint but really doesnt matter.

PR #1058, opened 2018-11-06
Enable package installation from branch

nickcolley:
This is an implementation of the discussion for the [Enable package installation from branch proposal](https://github.com/alphagov/govuk-design-system-architecture/pull/10).

It uses your current branch name and prefixes it with 'pre-release', and then publishes the built package directory to this branch, which can then be installed via npm.

As an example I ran this script on this branch itself resulting in:

- Result from running the script: https://github.com/alphagov/govuk-frontend/tree/pre-release-publish-build-to-branch
- You can see this being consumed in: https://github.com/alphagov/govuk-design-system/pull/629

https://trello.com/c/QqsC621F/1556-work-out-how-to-preview-changes-from-frontend-in-another-app

### DIFF
--- a/bin/release-to-branch.sh
+++ b/bin/release-to-branch.sh
@@ -1,0 +1,47 @@
+#!/bin/sh
+set -e
+
+CURRENT_BRANCH_NAME=$(git rev-parse --abbrev-ref HEAD)
+BRANCH_NAME="pre-release-$CURRENT_BRANCH_NAME"
+
+# Check if there are files that need to be commited
+if [[ -n $(git status --porcelain) ]]; then
+  echo "⚠️ You have unstaged files, please commit these and then try again."
+  exit 1
+fi
+
+# Remove local branch is it already exists
+if [ `git branch --list $BRANCH_NAME` ]; then
+  echo "⚠️ Cleaning up branch $BRANCH_NAME that already exists."
+  git branch -D $BRANCH_NAME
+fi
+
+# Work on the new branch
+git checkout -b $BRANCH_NAME
+
+# Build the package as normal
+npm run build:package
+
+# Check if the new built package has anything new to commit
+if [[ -n $(git status --porcelain) ]]; then
+  echo "✍️ Commiting changed package"
+  git add package/
+
+  git commit -m "Release GOV.UK Frontend to '$BRANCH_NAME' for testing"
+
+  # Create a local branch containing the package directory
+  echo "✨ Filter the branch to only the package/ directory..."
+  git filter-branch --force --subdirectory-filter package
+
+  # Force the push of the branch to the remote Github origin
+  git push origin $BRANCH_NAME:$BRANCH_NAME --force
+
+  echo "⚠️ Branch pushed to '$BRANCH_NAME', do not edit this by hand."
+else
+  echo "⚠️ No new changes to the package."
+fi
+
+git checkout -
+
+echo
+echo "✅ To install the pushed branch release run 'npm install --save alphagov/govuk-frontend#$BRANCH_NAME'"
--- a/docs/contributing/publishing-a-pre-release.md
+++ b/docs/contributing/publishing-a-pre-release.md
@@ -4,7 +4,9 @@ We use pre-releases when:
 - We are working on a new contribution and want to see it in the Design System website
 - We want to try an experimental feature as if it's pushed to npm.
 
-This is done by pushing a release to a new branch which can be installed by npm.
+This is done by pushing the files used for a GOV.UK Frontend release (the contents of the `package` directory) to a new branch which can be installed by npm as if it was a released npm package.
+
+No changes get published to npm as part of the process.
 
 1. Checkout the branch you want to pre-release and pull latest changes.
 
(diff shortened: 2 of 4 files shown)


bin/release-to-branch.sh
```diff
@@ -0,0 +1,31 @@
...
+
+# Create a local branch containing the package directory
+echo "Split package/ directory commits into `$BRANCH_NAME` branch..."
+git subtree split --prefix package -b $BRANCH_NAME
+
+# Force the push of the branch to the remote Github origin
+git push origin $BRANCH_NAME:$BRANCH_NAME
+
+# Remove build commit to bring the original branch back to where it was pre-release
+git reset HEAD^ --hard
```
36degrees: If you ran this script with uncommitted changes, you would currently lose those changes.

Running `git subtree split` should have created a new branch, could we just `git checkout CURRENT_BRANCH_NAME` to go back to the original branch, rather than resetting?

It might also be good to detect uncommitted changes and warn the user?

bin/release-to-branch.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/sh
+set -e
+
+CURRENT_BRANCH_NAME=$(git rev-parse --abbrev-ref HEAD)
+BRANCH_NAME="pre-release-$CURRENT_BRANCH_NAME"
+
+# Remove local branch is it already exists
+if [ `git branch --list $BRANCH_NAME` ]; then
```
36degrees: Given the effect is the same, could we just force push the branch instead?

bin/release-to-branch.sh
```diff
@@ -0,0 +1,31 @@
...
+# Build the package as normal
+npm run build:package
+
+git add package/
+
+git commit -m "Release GOV.UK Frontend to Github branch testing"
+
+# Create a local branch containing the package directory
+echo "Split package/ directory commits into `$BRANCH_NAME` branch..."
+git subtree split --prefix package -b $BRANCH_NAME
```
36degrees: Great solution 👌

bin/release-to-branch.sh
```diff
@@ -0,0 +1,31 @@
...
+
+# Force the push of the branch to the remote Github origin
+git push origin $BRANCH_NAME:$BRANCH_NAME
+
+# Remove build commit to bring the original branch back to where it was pre-release
+git reset HEAD^ --hard
+
+echo "To install this branch release run 'npm install --save alphagov/govuk-frontend#$BRANCH_NAME'"
```
36degrees: Should we add a message to make it clear that if you start making changes on the `prerelease-` branch you're possibly gonna get yourself in a bit of a mess?

bin/release-to-branch.sh
nickcolley: I had issues at the subtree split stage when the branch already existed, but I can double check this.

bin/release-to-branch.sh
nickcolley: We are resetting the build commit before we split rather than resetting the split, so not sure checking the original branch would work?

Agreed with uncommitted stuff...

36degrees: Gotcha, that makes sense. Maybe the script just refuse to run if there are uncommitted changes?

nickcolley:
I have found another way to do this that seems to be a lot quicker than doing `git subtree split`:

https://help.github.com/articles/splitting-a-subfolder-out-into-a-new-repository/

nickcolley:
I have tested the following scenarios:

1. No branches exist, running for the first time: it succeeds
2. I have made a change locally and not committed it: it warns you and stops.
3. I have made a change and committed it with existing release pushed: it overwrites this release.
4. I have made no changes and run the release: it works

docs/contributing/publishing-a-pre-release.md
```diff
@@ -0,0 +1,21 @@
+# Publishing pre-release of GOV.UK Frontend
+
+We use pre-releases when:
+- We are working on a new contribution and want to see it in the Design System website
+- We want to try an experimental feature as if it's pushed to npm.
+
+This is done by pushing a release to a new branch which can be installed by npm.
```
hannalaakso: It might be obvious to some but should this make it clear that this is not an actual release but rather the equivalent of the files that get pushed to npm? 

Maybe something like `pushing the files used for a GOV.UK Frontend release (the contents of the `package` directory) to a new branch which can be installed by npm as if it was a released npm package. No changes get published to npm as part of the process.`

nickcolley: Yeah doesnt hurt to be more explicit since this is a bit of a weird process, I'll update.

bin/release-to-branch.sh
hannalaakso: I believe this is now fixed? I'm seeing the following if I have uncommitted changes:
```
⚠️ You have unstaged files, please commit these and then try again.
```

alex-ju:
Tested this on Friday too. @nickcolley's work entirely 👏 .
Thanks for this! Hopefully it will ease the contribution process, especially for components.

nickcolley:
@alex-ju we wouldnt have done it without you pushing to get this fixed, thanks Alex

PR #1061, opened 2018-11-09
Fix footer styling in IE8

igloosi:
At the moment footer looks broken in IE8. It doesn't like the padding being set on top and bottom of the container, resulting in an increased padding at the bottom and a mispositioned top border.

This work moves the responsive padding to the 'meta' element for IE8, but keeps it on the wrapper for IE9 and modern browsers.

Before:
bs_win7_ie_8 0

After: 
bs_win7_ie_8 0 1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,11 +28,9 @@
 
 🔧 Fixes:
 
-- Pull Request Title goes here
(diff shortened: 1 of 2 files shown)


igloosi:
realised this won't work when navigation in present.

Issue #1044, opened 2018-10-25
classList polyfill needs to import the Elements polyfill


nickcolley:
See https://github.com/alphagov/govuk-design-system/pull/596/commits/65d9e69a09ac02c8bc7ec125154640522c652d26#diff-9f975bd4d01eb71ff7e1ecad9f2a0484R2

nickcolley:
Related https://github.com/alphagov/govuk-frontend/issues/674

PR #1063, opened 2018-11-15
Remove implicit dependency on Element for classList

nickcolley:
Fixes https://github.com/alphagov/govuk-frontend/issues/1044

I think we should also consider https://github.com/alphagov/govuk-frontend/issues/674 since this will likely happen again.

From: https://cdn.polyfill.io/v2/polyfill.js?features=Element.prototype.classList&flags=always

 - Object.defineProperty, License: CC0 (required by "Element.prototype.classList", "_DOMTokenList", "DOMTokenList")
 - _DOMTokenList, License: ISC (required by "DOMTokenList", "Element.prototype.classList")
- DOMTokenList, License: CC0 (required by "Element.prototype.classList")
- Document, License: CC0 (required by "Element", "Element.prototype.classList")
- Element, License: CC0 (required by "Element.prototype.classList")
- Element.prototype.classList, License: ISC

We can remove the import from the design system after this ships: https://github.com/alphagov/govuk-design-system/pull/596/files#diff-9f975bd4d01eb71ff7e1ecad9f2a0484R2

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,10 @@
 
   ([PR #N](https://github.com/alphagov/govuk-frontend/pull/N))
 
+- Remove implicit dependency on Element for classList
+
+  ([PR #1063](https://github.com/alphagov/govuk-frontend/pull/1063))
+
 - Single field with error should have 'aria-describeby' attribute
 
   Although we discourage using checkboxes without fieldsets, this fix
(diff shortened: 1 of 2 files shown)


nickcolley:
Oops yes this needs a CHANGELOG entry

igloosi:
👍

PR #1065, opened 2018-11-19
Add summary list component

dashouse:
This component was initially developed to allow us to build the
'check your answers' pattern.

It is mostly the same as in the original pattern with some notable differences:

- On smaller screens it wraps by default
- It's possible to have multiple actions

## Review
- Review application: https://govuk-frontend-review-pr-1065.herokuapp.com/components/summary-list

- Original GOV.UK Prototype Kit implementation: https://govuk-prototype-kit.herokuapp.com/docs/templates/check-your-answers

## Todo

- [x] Consider visuallyHiddenText feature instead of using `html`
- [x] Improve the options descriptions
- [x] Assistive technology final run through

https://trello.com/c/kh2atSrd/1612-add-summary-list-component-to-govuk-frontend-pair

### DIFF
--- a/src/components/summary-list/summary-list.yaml
+++ b/src/components/summary-list/summary-list.yaml
@@ -108,6 +108,31 @@ examples:
               - href: '#'
                 text: Change
                 visuallyHiddenText: contact information
+  - name: no-actions
+    data:
+      rows:
+        - key:
+            text: Name
+          value:
+            text: Firstname Lastname
+        - key:
+            text: Date of birth
+          value:
+            text: 13/08/1980
+        - key:
+            text: Contact information
+          value:
+            html: |
+              <p class="govuk-body">
+                email@email.com
+              </p>
+              <p class="govuk-body">
+                Address line 1<br>
+                Address line 2<br>
+                Address line 3<br>
+                Address line 4<br>
+                Address line 5
+              </p>
   - name: no-border
     data:
       classes: govuk-summary-list--no-border
--- a/src/components/summary-list/summary-list.yaml
+++ b/src/components/summary-list/summary-list.yaml
@@ -1,0 +1,422 @@
+params:
+- name: rows
+  type: array
+  required: true
+  description: Array of row item objects
+  params:
+  - name: key.text
+    type: string
+    required: true
+    description: If `html` is set, this is not required. Text to use within the each key. If `html` is provided, the `text` argument will be ignored.
+  - name: key.html
+    type: string
+    required: true
+  - name: key.classes
+    type: string
+    required: false
+    description: Classes to add to the key wrapper
+  - name: value.text
+    type: string
+    required: true
+    description: If `html` is set, this is not required. Text to use within the each value. If `html` is provided, the `text` argument will be ignored.
+  - name: value.html
+    type: string
+    required: true
+    description: If `text` is set, this is not required. HTML to use within the each value. If `html` is provided, the `text` argument will be ignored.
+  - name: value.classes
+    type: string
+    required: false
+    description: Classes to add to the value wrapper
+  - name: actions.items
+    type: array
+    required: false
+    description: Array of action item objects
+    params:
+    - name: classes
+      type: string
+      required: false
+      description: Classes to add to the actions wrapper
+    - name: href
+      type: string
+      required: true
+      description: The value of the link href attribute for an action item
+    - name: text
+      type: string
+      required: true
+      description: If `html` is set, this is not required. Text to use within each action item. If `html` is provided, the `text` argument will be ignored.
+    - name: html
+      type: string
+      required: true
+      description: If `text` is set, this is not required. HTML to use within the each action item. If `html` is provided, the `text` argument will be ignored.
+    - name: visuallyHiddenText
+      type: string
+      required: false
+      description: Actions rely on context from the surrounding content so may require additional accessible text, text supplied to this option is appended to the end, use `html` for more complicated scenarios.
+- name: classes
+  type: string
+  required: false
+  description: Classes to add to the container.
+- name: attributes
+  type: object
+  required: false
+  description: HTML attributes (for example data attributes) to add to the container.
+
+examples:
+  - name: default
+    data:
+      rows:
+        - key:
+            text: Name
+          value:
+            text: Firstname Lastname
+          actions:
+            items:
+              - href: '#'
+                text: Edit
+                visuallyHiddenText: name
+              - href: '#'
+                text: Delete
+                visuallyHiddenText: name
+        - key:
+            text: Date of birth
+          value:
+            text: 13/08/1980
+          actions:
+            items:
+              - href: '#'
+                text: Change
+                visuallyHiddenText: date of birth
+        - key:
+            text: Contact information
+          value:
+            html: |
+              <p class="govuk-body">
+               email@email.com
+              </p>
+              <p class="govuk-body">
+                Address line 1<br>
+                Address line 2<br>
+                Address line 3<br>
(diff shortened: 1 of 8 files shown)


src/components/data-list/_data-list.scss
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    @include govuk-media-query($from: tablet) {
+      width: 25%;
+      padding-bottom: govuk-spacing(2);
+      border-bottom: 1px solid $govuk-border-colour;
+    }
+  }
+
+  .govuk-data-list__item-align-right {
```
igloosi:
```suggestion
  .govuk-data-list__item--align-right {
```
?

nickcolley:
I'm wondering if we should rename the classes to something like

```
.govuk-data-list__key {}
.govuk-data-list__value {}
.govuk-data-list__actions {}
```

@dashouse can we only include modifiers that we know are needed for check your answers?
Reducing the API surface is good if possible, but if you think some of these modifiers are definitely going to be used I won't block that.

dashouse:
I think there was some desire to get away from the "key value pairs" name, I'm not sure if that extends to class names too but I'm not precious about any of it.

Regarding the wrapping as default, let me talk to Joe as I think he designed it originally. The only reason I didn't do this was that the current pattern doesn't.

The modifiers were there to make it meet the needs for more than 'Check answers', although I'm not sure we need the text align one as that would be an easy modification to make for someone / could be a universal override.

p.s. Looking good, I think a few of the CSS changes you've made have changed the spacing a little. We can go through this when we're in the same room if that's easier.

nickcolley:
@dashouse I think we either go full hog and mirror the definition list terms, or intentionally try to avoid it.

joelanman:
Just to note `datalist` is an existing html element so we might not want to reuse the name:

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist

nickcolley:
Classes functionality, now:


```javascript
{{ govukDataList({
  "keyClasses": "govuk-!-width-one-half",
  "valueClasses": "govuk-!-width-one-quarter",
  "actionsClasses": "govuk-!-width-one-half",
  "rows": [
    {
      "key": "Name",
      "text": "Firstname Lastname",
      "actions": [
        {
          "href": "#",
          "text": "Edit"
        },
        {
          "href": "#",
          "text": "Delete"
        }
      ]
    }
  ]
}) }}
```

Alternatives:

```javascript
{{ govukDataList({
  "rows": [
    {
      "key": {
        classes: 'key-class',
        text: "Name"
      },
      "value": {
        classes: 'value-class',
        "text": "Sarah Philips"
      },
      "actions": {
        classes: 'actions-class',
        items: [
          {
            "href": "#",
            "text": "Change"
          }
        ]
      }
      ]
    }
  ]
}) }}
```

or maybe

```javascript
{{ govukDataList({
  "rows": [
    {
      "keyClasses": "govuk-!-width-one-half",
      "key": "Name",
      "classes": "govuk-!-width-one-quarter",
      "text": "Firstname Lastname",
      "actionsClasses": "govuk-!-width-one-half",
      "actions": [
        {
          "href": "#",
          "text": "Edit"
        },
        {
          "href": "#",
          "text": "Delete"
        }
      ]
    }
  ]
}) }}
```

fofr:
Current screenshot for interested onlookers:
screen shot 2018-11-26 at 12 13 11

https://govuk-frontend-review-pr-1065.herokuapp.com/components/data-list

src/components/data-list/data-list.yaml
```diff
@@ -0,0 +1,121 @@
...
+          <p class="govuk-body">
+            Address line 1<br>
+            Address line 2<br>
+            Address line 3<br>
+            Address line 4<br>
+            Address line 5
+          </p>
+ - name: overridden-widths
+   data:
+    keyClasses: govuk-!-width-one-half
```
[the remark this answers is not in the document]
igloosi: 👍

nickcolley:
Maybe we should only take the key if it's text, or overridable.

nickcolley:
I went through the API proposal with the dev team and we agreed with the current approach.

nickcolley:
@dashouse I've added a commit that makes the spacing more consistent with the prototype kit, just for conversation tomorrow.

nickcolley:
I wonder if we should keep it simple and just have guidance examples with actions.items.html = 'blah <span class="visually-hidden"> foo</span>`...

nickcolley:
Wondering if we should do `actions.items` -> `action.items` ?

nickcolley:
I don't want to block shipping this on a discussion on naming but it would be nice to have clarification to why this component could not be called 'summary list' instead of 'data summary list'?

It's not clear to me why we need the data prefix, would be good for future us to have on record I think.

nickcolley:
VoiceOver OSX:

Links read "name Change" when I expected "Change name", this might be consistent with https://github.com/alphagov/govuk-frontend/issues/1096

nickcolley:
Without more investigation and thinking I'm not sure I can make a decision on if this should ship as is, it is consistent with the GOV.UK Prototype Kit implementation.

Ideally I think we should talk to the accessibility team and try to come up with how we might resolve the general issue of our `govuk-visually-hidden` class (https://github.com/alphagov/govuk-frontend/issues/1096).

I don't believe there's any hard deadline for this component, so it could wait until we've figured out the best way to proceed.

Alternatively we could change how this component works to use ARIA, by replacing 'visuallyHiddenText' with 'accessibleName' as follows:

```javascript
{
  text: 'Change',
  accessibleName: 'Change name'  // see https://www.w3.org/TR/accname-1.1/
}
```

```html
<a href="#" aria-label="Change name">Change</a>
```

dashouse:
It looks like somewhere in the code refactoring we’ve lost the styling for the summary list, with borders but without actions. This means the spacing looks like this on small screens:

screen shot 2018-12-20 at 11 03 13

I'll give this a go

dashouse:
Latest commit adjusts small screen spacing for actions

screen shot 2018-12-20 at 11 54 30

And increases spacing when there is no actions (with a new class `govuk-summary-list--no-actions` .

screen shot 2018-12-20 at 11 54 42

dashouse:
@nickcolley Thanks, of course. This will just work now, just moved the margin https://github.com/alphagov/govuk-frontend/commit/cbee09e32a3a1858092a71447813acbfb8547ff9

src/components/summary-list/summary-list.yaml
```diff
@@ -0,0 +1,447 @@
...
+                text: Lock
+              - href: '#'
+                text: Fill
+              - href: '#'
+                text: Curl
+              - href: '#'
+                text: Find
+              - href: '#'
+                text: View
+              - href: '#'
```
36degrees: Minor – this last link doesn't have any text, so you get an empty link at the end of the actions.

src/components/summary-list/summary-list.yaml
```diff
@@ -0,0 +1,447 @@
...
+    data:
+      rows:
+        - key:
+            text: Name
+          value:
+            text: Barnaby Marmaduke Aloysius Benjy Cobweb Dartagnan Egbert Felix Gaspar Humbert Ignatius Jayden Kasper Leroy Maximilian Neddy Obiajulu Pepin Quilliam Rosencrantz Sexton Teddy Upwood Vivatma Wayland Xylon Yardley Zachary Usansky
+          actions:
+            items:
+              - href: '#'
+                text: Buy
```
36degrees: 😆👌

joelanman:
great work everyone :)

PR #1072, opened 2018-11-20
Simplify readmes

nickcolley:
These READMEs were originally built for when we had a monorepo.

Each component would be published to npm and the README would be the primary way to understand it.

We have now moved our macro options into the main GOV.UK Design System website, which also includes guidance and examples we want our users to use.

We have noticed some people sharing the review application links, potentially one way these have been shared is from these READMES.

This change makes the READMEs drastically simpler, and points to always up-to-date guidance, by making them simpler it removes the need for them to be templated so they can be regular markdown files.

Other benefits include:

- not having to generate READMEs for any changes to examples
- not failing contributors builds if they forget to update the README
- removes complexity from the build pipeline
- makes package/ closer to src/


tasks/gulp/watch.js
```diff
@@ -8,5 +8,5 @@ const configPaths = require('../../config/paths.json')
 gulp.task('watch', () => {
   gulp.watch([configPaths.src + '**/**/*.scss', configPaths.app + 'assets/scss/**/*.scss'], ['styles', 'sassdoc'])
   gulp.watch([configPaths.src + '**/**/*.js'], ['scripts'])
-  // gulp.watch([configPaths.src + '**/**/*.njk'], ['generate:readme', 'compile:components'])
+  // gulp.watch([configPaths.src + '**/**/*.njk'], ['compile:components'])
```
36degrees: Given this is commented out, should we just remove this line? (non-blocking)

nickcolley: Sure will do

PR #1078, opened 2018-11-22
Add 5px bottom margin to bullet-less list items

### LINKED ISSUE #1069: `govuk-list` needs margin adjustment
On small screens `govuk-list` > `<li>` has no bottom margin (with 5px being applied at tablet and above)

This works when the bullet or number modifier is added e.g. `class="govuk-list govuk-list--bullet"` because the bullet marker gives a clear indication of the start of a new list item.

When the `govuk-list` class is applied on it's own, such as a list of links the `<li>`'s appear to run into each other like so:

screen shot 2018-11-19 at 15 38 13

My suggestion would be to apply to spacing to the `govuk-list` class for all breakpoints, then remove when the modifiers are added.

```
  // add spacing to list on all breakpoints

  %govuk-list > li {
      margin-bottom: govuk-spacing(1);
  }

  %govuk-list--bullet > li,
  %govuk-list--number > li {
  // remove spacing from govuk-list
  margin-bottom: 0; 
  
  // add spacing back in from tablet
      margin-bottom: govuk-spacing(1);
    }
  }
```

**Before**
screen shot 2018-11-19 at 15 36 46

**After**
screen shot 2018-11-19 at 15 37 42

36degrees:
On small screens list items have no bottom margin (with a 5px bottom margin being applied at tablet and above)

When the govuk-list class is applied on it's own without any modifier, the list items appear to run into each other on mobile, as the spacing between the list items is the same as the spacing between each line.

(When a bullet or number modifier is added e.g. class="govuk-list govuk-list--bullet" the bullet or number indicates the start of a new list item, so the spacing does not need to be adjusted in that case.)

This change applies the spacing to the govuk-list class for all breakpoints, then adjusts the behaviour when bullet or number modifiers are added.

Closes #1069 

https://trello.com/c/pUKbZTbi/1629-add-bottom-margin-to-non-bulleted-list-items

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,11 +22,10 @@
 
 🔧 Fixes:
 
-- Pull Request Title goes here
+- Add 5px bottom margin to list items within lists that do not have bullets or
+  numbers on mobile breakpoints to make each item visually distinct.
 
-  Description goes here (optional)
-
-  ([PR #N](https://github.com/alphagov/govuk-frontend/pull/N))
+  ([PR #1078](https://github.com/alphagov/govuk-frontend/pull/1078))
 
 ## 2.4.0 (Feature release)
 
--- a/src/core/_lists.scss
+++ b/src/core/_lists.scss
@@ -15,9 +15,10 @@
   }
 
   %govuk-list > li {
-    @include govuk-media-query($from: tablet) {
-      margin-bottom: govuk-spacing(1);
-    }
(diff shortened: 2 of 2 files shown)


36degrees:
This is basically Dave's suggested implementation from https://github.com/alphagov/govuk-frontend/issues/1069, so I've added him as a co-author – but we didn't actually pair on this work.

Issue #1083, opened 2018-12-03
Unneeded padding on warning component?


nickcolley:
> Noticed that the govuk-warning-text__text element has padding left that includes an extra 15px to compensate for a negative margin left (see below, dotted outline added by me to show boundary of component). Seems like this could be removed and the padding decreased, but there must be a reason for it?

![screen shot 2018-12-03 at 11 30 12](https://user-images.githubusercontent.com/861310/49371364-2a2cdd00-f6ef-11e8-886a-edc4676dd188.png)

_Originally posted by @andysellick in https://github.com/alphagov/govuk-design-system-backlog/issues/71#issuecomment-443679660_

36degrees:
For reference, it looks like this was [copied from Elements](https://github.com/alphagov/govuk_elements/commit/7a77a95fcd285f068076f421bcec84f518734436).

36degrees:
Yep, but removing the negative margin and reducing the padding appears to be the same:

```diff
- margin-left: -$govuk-gutter-half;
- padding-left: 65px;
+ padding-left: 50px;
```

<img width="994" alt="screen shot 2018-12-03 at 13 44 55" src="https://user-images.githubusercontent.com/121939/49377275-a7614d80-f701-11e8-846f-0a24da1f6e02.png">

nickcolley:
That's weird, I've asked Andy if he wants to do a PR for this...

andysellick:
I am busy, but I'd like to contribute (if only to experience the process) so I'll try to raise a PR soon (also it's not a huge change). Thanks!

nickcolley:
@andysellick will leave this open for you when you get a second then.

stevenaproctor:
@nickcolley @36degrees Do you think the icon should be vertically aligned to the middle of the paragraph, like in the example, or the top or the paragraph?

stevenaproctor:
@nickcolley I just think it scans better if the icon is at the top of the paragraph. You would naturally scan left to right, not left, down, up, right...if you see what I mean.

PR #1084, opened 2018-12-03
Clean up warning text component styles

andysellick:
- remove negative margin left and reduce padding left to match

Fixes https://github.com/alphagov/govuk-frontend/issues/1083

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,10 @@
 
 🔧 Fixes:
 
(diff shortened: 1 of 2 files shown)


36degrees:
Thanks very much for picking this up, Andy – the change looks good 👍

Would you mind adding an entry to [the CHANGELOG](https://github.com/alphagov/govuk-frontend/blob/master/CHANGELOG.md) under 'Unreleased'? I think this would count as a fix.

PR #1085, opened 2018-12-03
Improve nvmrc check

nickcolley:
Through helping @andysellick get Node.js setup with the right version we found that when running `npm start` on an older version resulted in an error message that was confusing. This is consistent with what we do on the GOV.UK Design System repository.

This pull request adds a `prestart` and `preinstall` command that runs the Node.js verison check, which is more actionable.

It also adds instructions to try and use nvm, since this will be easier than trying to find the right Node.js version online.

### DIFF
--- a/bin/check-nvmrc.js
+++ b/bin/check-nvmrc.js
@@ -17,17 +17,25 @@ fs.readFile(path.join(__dirname, '../.nvmrc'), 'utf8', function (error, data) {
     process.exit()
   }
 
+  var nvmInstallText = 'To do this you can install nvm (https://github.com/creationix/nvm) then run `nvm install`.'
+
   if (versionMatchesMajor) {
     console.log('' +
       'Warning: You are using Node.js version ' + currentVersion + ' which we do not use. ' +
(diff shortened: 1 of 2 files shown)


bin/check-nvmrc.js
```diff
@@ -17,17 +17,25 @@ fs.readFile(path.join(__dirname, '../.nvmrc'), 'utf8', function (error, data) {
     process.exit()
   }
 
+  var nvmInstallText = 'To do this you can install nvm (https://github.com/creationix/nvm) then run `nvm install`.'
+
   if (versionMatchesMajor) {
     console.log('' +
       'Warning: You are using Node.js version ' + currentVersion + ' which we do not use. ' +
+      '\n\n' +
```
joelanman: wouldn't have to be done in this PR (non blocking) but would this be clearer using backticks? `

nickcolley: This code needs to be written in a way that'll run in older versions of Node.js, backticks are only avaliable in newer versions of Node.js

PR #1089, opened 2018-12-12
WIP: export tests as a 'spec'

penx:
This is based on a discussion I had recently with @36degrees 

For projects acting as a port of the nunjucks templates, in a different templating language, it would be useful to be able to run your unit tests as a 'spec' against a custom render function, to ensure the port has feature parity with the govuk-frontend nunjucks templates.

I have only done this for the button so far as a proof of concept.

I have introduced the naming convention:
- *.test.js is what is run by govuk-frontend
- *.spec.js is what is exported to be used by upstream projects

I also had to modify jest's testRegex so that it didn't try to run *.spec.js directly.

TODO:

- discuss whether this is a feasible approach to role out to all components
- publish to a temporary scoped module and prove this working in an upstream project
- discuss whether the render function provided upstream returns should a string rather than a cheerio object - and have the cheerio wrapper added inside the .spec.js files themselves


penx:
I'm having a few issues preventing this from working:

- the spec is reliant on jest-helpers and axe-helper which sit outside of src and are not part of the compiled package
- in turn, these need access to button.yaml is not copied across to the build
- jest-helpers.js tries to load button.yaml via 'configPaths.components' path, which points to src, which isn't part of the distributed package

At this point I'm pretty happy this would be a very useful feature for us, but would require some structural changes to govuk-frontend to achieve which I'd either need the maintainers to do, or support/instruct me in doing myself

36degrees:
Thanks for this @penx!

I think you're right that this needs more direction and/or work from us. We're in the middle of wrapping up a chunk of work before Christmas at the minute – can we revisit this properly in the new year?

penx:
sure

PR #1090, opened 2018-12-12
Simplify the grid mixins

36degrees:
This simplifies the grid to make it easier to add the [extended grid](https://github.com/alphagov/govuk-frontend/pull/1041).

- Move grid widths to the settings layer and make part of the public API
- Rename the `grid-width` function to `govuk-grid-width` to be consistent with other functions, retaining `grid-width` as as deprecated alias
- Copy `govuk-grid-row` mixin to create a new concrete class and mark the mixin as deprecated
- Update the `govuk-grid-column` mixin to allow class name to be suppressed by setting `$class` to `false`.
- Mark the `class` parameter of the `govuk-grid-column` mixin as deprecated
- Create grid classes programatically in objects layer by iterating over the `$govuk-grid-widths` map.

https://trello.com/c/1aBarhHU/1613-add-extended-grid-to-govuk-frontend-pair

### DIFF
--- a/src/settings/_measurements.scss
+++ b/src/settings/_measurements.scss
@@ -15,6 +15,20 @@
 
 $govuk-page-width: 960px !default;
 
+/// Map of grid column widths
+///
+/// @type Map
+/// @access public
+
+$govuk-grid-widths: (
+  one-quarter: 25%,
+  one-third: 33.3333%,
+  one-half: 50%,
+  two-thirds: 66.6666%,
+  three-quarters: 75%,
+  full: 100%
+) !default;
+
 /// Width of gutter between grid columns
 ///
 /// @type Number
--- a/src/helpers/_grid.scss
+++ b/src/helpers/_grid.scss
@@ -2,20 +2,6 @@
 /// @group helpers
 ////
 
-/// Map of grid column widths
-///
-/// @type Map
-/// @access private
-
-$_govuk-grid-widths: (
-  one-quarter: 25%,
-  one-third: 33.3333%,
-  one-half: 50%,
-  two-thirds: 66.6666%,
-  three-quarters: 75%,
-  full: 100%
-) !default;
-
 /// Grid width percentage
 ///
 /// @param {String} $key - Name of grid width (e.g. two-thirds)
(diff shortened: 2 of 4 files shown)


36degrees:
I think there would be three issues raised against 3.0 off the back of this work:

- Remove deprecated `grid-width` function
- Remove deprecated `govuk-grid-row` mixin
- Remove `$class` parameter and logic from the `govuk-grid-column` mixin

src/settings/_measurements.scss
```diff
@@ -15,6 +15,20 @@
...
 $govuk-page-width: 960px !default;
 
+/// Map of grid column widths
+///
+/// @type Map
+/// @access public
+
+$govuk-grid-widths: (
+  one-quarter: 25%,
+  one-third: 33.3333%,
```
Nooshu: Quick question. Is it worth letting the Sass language handle the [precision here](http://sass-lang.com/documentation/Sass/Script/Value/Number.html#precision-class_method)?

Swapping this out for:

```
one-third: (100/3) * 1%,
```

I believe this is possible in a map as according to the docs: "Both the keys and values in maps can be any SassScript object".

I doubt it would have any effect on the rounded result so may just be best left hardcoded.

nickcolley: Might be worth keeping it simple since it works and doesnt require thinking.

36degrees:
Verified that the CSS output is the same using both snapshots and a quick MD5 check:

```
➜  ~ curl https://govuk-frontend-review.herokuapp.com/public/app.css | md5
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  102k  100  102k    0     0   261k      0 --:--:-- --:--:-- --:--:--  261k
039d3275202aedb95bf0a5cbc92bafa3

➜  ~ curl https://govuk-frontend-review-pr-1090.herokuapp.com/public/app.css | md5
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  102k  100  102k    0     0   650k      0 --:--:-- --:--:-- --:--:--  651k
039d3275202aedb95bf0a5cbc92bafa3
```

PR #1094, opened 2018-12-14
Add desktop specific grid classes

### LINKED ISSUE #1115: Add desktop specific breakpoints to grid
## What

Provide an 'extended grid' which adds desktop-specific grid breakpoints, based on the WIP branch here:

https://github.com/alphagov/govuk-frontend/pull/1041

## Why

We want to be able to provide the code to help service teams implement the 'check answers' pattern.

As part of https://trello.com/c/y4t7YGwo/1549-spike-plan-approach-to-check-answers-pattern, as well as previous work on what additional features our users need from a grid, we've identified that breakpoint-specific grid widths will be useful in a number of scenarios, including 'check answers'.

## Who needs to be involved

@dashouse

36degrees:
This supersedes #1041 and builds on work done in #1090.

These classes allow users to create more advanced layouts, such as used in the ‘check answers’ pattern, where the list of answers users the entire width at tablet, but is restricted to the left two-thirds on desktop.

There may be a need for breakpoint constrained classes in the future (for example, an `at-tablet` class which applies only whilst within the tablet breakpoint) so we’re using `-from-` to signify that the class applies at desktop _and above_.

Closes #1115

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,14 @@
 
   ([PR #N](https://github.com/alphagov/govuk-frontend/pull/N))
 
+- Add desktop specific grid column widths
+
+  This allow you to define different grid behaviour for the tablet and desktop
+  breakpoints. For example, you can make a column two-thirds on desktop but
+  expand to full-width on smaller tablet sized screens.
+
+  ([PR #1094](https://github.com/alphagov/govuk-frontend/pull/1094))
+
 - Add summary list component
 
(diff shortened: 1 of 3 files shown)


dashouse:
This was tested as part of https://github.com/alphagov/govuk-frontend/pull/1097 and worked as expected

36degrees:
Updated with a changelog entry.

nickcolley:
👍

PR #1100, opened 2018-12-19
Clarify process for deploying a pre-release

aliuk2012:
Something I picked up while pairing with Nick to update the Accordion's preview link.

Changes are needed in two different repos and the existing README does
not explain this well.

### DIFF
--- a/docs/contributing/publishing-a-pre-release.md
+++ b/docs/contributing/publishing-a-pre-release.md
@@ -16,7 +16,9 @@ No changes get published to npm as part of the process.
 
 4. Run `npm run release-to-branch`.
 
-There should now be a branch `pre-release-[your-branch-name-here]` pushed to remote, which you can install with:
+There should now be a branch `pre-release-[your-branch-name-here]` pushed to the GOV.UK Frontend remote origin.
+
+To install this pre-release branch in another project, you can run the following command:
 
 ```bash
 npm install --save alphagov/govuk-frontend#pre-release-[your-branch-name-here]


docs/contributing/publishing-a-pre-release.md
```diff
@@ -16,8 +16,17 @@ No changes get published to npm as part of the process.
 
 4. Run `npm run release-to-branch`.
 
-There should now be a branch `pre-release-[your-branch-name-here]` pushed to remote, which you can install with:
+There should now be a branch `pre-release-[your-branch-name-here]` pushed to remote.
+
+5. In GOVUK Design System:
```
36degrees: I think this is useful, but I think there are use cases for pre-releases that _do not_ involve a change to the Design System, and this change will be confusing for those users.

For example, if a team wanted us to make a change to an existing component and we wanted to be able to share a pre-release with them so they could test it, they would install the pre-release branch in their own application, not the Design System.

I suggest making it clear that this step is optional, splitting it out entirely or keeping it generic.

(Minor: GOV.UK, not GOVUK)

docs/contributing/publishing-a-pre-release.md
```diff
@@ -16,8 +16,17 @@ No changes get published to npm as part of the process.
...
+- create a new branch and new PR if one does not already exist.
+- install the pre-release-branch by running the following command
 
 ```bash
 npm install --save alphagov/govuk-frontend#pre-release-[your-branch-name-here]
-```
\ No newline at end of file
+```
+
+6. Test the changes updates using `npm test`
```
36degrees: I'm not sure it's clear that this is in the Design System rather than back in Frontend – I certainly interpreted as running these in the Frontend repo initially.

nickcolley:
With Alistairs permission I've updated this based on @36degrees 's feedback.

It's simpler but I think takes time to make clear to run this command in another project, which I think could be the main cause for confusion.

Let me know if this is good to go @aliuk2012 @36degrees

PR #1102, opened 2018-12-20
Added config file for prototype kit.

matcarey:
To accompany the extensions proposal govuk-frontend this config allows govuk-frontend to be an extension in the prototype kit.

Implementation details can be found here https://github.com/alphagov/govuk-prototype-kit/pull/613
Proposal details can be found here https://github.com/alphagov/govuk-design-system-architecture/pull/9

### DIFF
--- a/package/govuk-prototype-kit.config.json
+++ b/package/govuk-prototype-kit.config.json
@@ -1,0 +1,15 @@
+{
+  "nunjucksPaths": [
+    "/",
+    "/components"
+  ],
+  "scripts": [
+    "/all.js"
+  ],
+  "assets": [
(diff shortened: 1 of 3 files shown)


joelanman:
I think from discussions, we were thinking of being able to list out extensions on the Prototype home page.

For this I think we agreed on adding a 'documentationURL' property, maybe a 'name' too - but maybe these could/should be in package.json and referenced there?

matcarey:
Those can easily be added.  I'd suggest using the package name to define the name but if having a `name` field is desirable I don't see any problems with it.

nickcolley:
:rocket:

PR #1107, opened 2019-01-03
Improve publishing workflow

nickcolley:
When @aliuk2012 and I went through the latest release I took notes and these are the improvements from what we discussed.

Most controversial thing is prompting to continue in the release steps, I'm not a bash genius so would appreciate a good eye on that, happy to remove those improvements too if it's too controversial.

Use https://github.com/alphagov/govuk-frontend/pull/1107/files?w=1 to see diff without whitespace changes.

### DIFF
--- a/bin/pre-release.sh
+++ b/bin/pre-release.sh
@@ -1,6 +1,22 @@
 #!/bin/sh
 set -e
 
+echo "Starting a pre-release..."
+echo " "
+echo "This will:"
+echo "- run the test suite"
+echo "- build GOV.UK Frontend into the 'package/' directory"
+echo "- build GOV.UK Frontend into the 'dist/' directory"
+echo "- commit all changes and push the branch to remote"
+echo " "
+
+read -r -p "Do you want to continue? [y/N] " continue_prompt
+
+if [[ $continue_prompt != 'y' ]]; then
+    echo "Cancelling pre-release, if this was a mistake, try again and use 'y' to continue."
+    exit 0
+fi
+
 npm run test
 npm run build:package
 npm run build:dist
@@ -15,7 +31,7 @@ if [ $(git tag -l "$TAG") ]; then
 else
     git add .
     git commit -m "Release $TAG"
-    #set upstream so that we can push the branch up
+    # set upstream so that we can push the branch up
     git push --set-upstream origin $CURRENT_BRANCH_NAME
     git push
     echo "🗒 All done. Ready to create a pull request. Once approved, run npm run release"
--- a/bin/release.sh
+++ b/bin/release.sh
@@ -1,6 +1,24 @@
 #!/bin/sh
 set -e
 
+echo "Starting a release..."
+echo " "
+echo "This will:"
+echo "- check that you're logged in to npm as the correct user"
+echo "- publish the package if it has not been published already"
+echo "- check that there is not already a tag published"
+echo "- create a new tag"
+echo "- push the tag to remote origin"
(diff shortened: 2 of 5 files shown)


bin/pre-release.sh
```diff
@@ -1,22 +1,37 @@
...
+
+    ALL_PACKAGE_VERSION=$(node -p "require('./package/package.json').version")
+    TAG="v$ALL_PACKAGE_VERSION"
+    CURRENT_BRANCH_NAME=$(git rev-parse --abbrev-ref HEAD)
+
+    if [ $(git tag -l "$TAG") ]; then
+        echo "⚠️ Git tag already exists."
+        exit 1;
+    else
+        # git add .
```
nickcolley: I need to uncomment these...

bin/pre-release.sh
```diff
@@ -1,22 +1,37 @@
...
 
-ALL_PACKAGE_VERSION=$(node -p "require('./package/package.json').version")
-TAG="v$ALL_PACKAGE_VERSION"
-CURRENT_BRANCH_NAME=$(git rev-parse --abbrev-ref HEAD)
+read -r -p "Do you want to continue? [y/N] " continue_prompt
 
-if [ $(git tag -l "$TAG") ]; then
-    echo "⚠️ Git tag already exists."
-    exit 1;
+if [[ $continue_prompt = 'y' ]]; then
```
36degrees: I'd consider flipping this logic and just exiting it `$continue_prompt` does _not_ equal `y` – then you can avoid indenting the whole block.

bin/release.sh
```diff
@@ -1,39 +1,55 @@
...
-    git tag $TAG -m "GOV.UK Frontend release $TAG"
-    git push --tags
-    echo "🗒 Tag $TAG created and pushed to remote."
-
-    echo "🗒 Creating a release artifact..."
-    git archive -o ./release-$TAG.zip HEAD:dist
-    echo "🗒 Artifact created. Now create a release on GitHub and attach this."
-fi
+    echo "Cancelling release, if this was a mistake, try again and use 'y' to continue."
+fi
```
36degrees: New line missing from the end of the file.

bin/release.sh
```diff
@@ -1,39 +1,55 @@
...
+echo "- check that you're logged in to npm as the correct user"
+echo "- publish the package if it has not been published already"
+echo "- create a new tag if the current git tag does not match the latest published tag"
+echo "- push the tag to remote origin"
+echo "- create a zip file of the 'dist/' directory locally"
+echo " "
+
+read -r -p "Do you want to continue? [y/N] " continue_prompt
+
+if [[ $continue_prompt = 'y' ]]; then
```
36degrees: Again, I'd consider flipping this logic.

bin/release.sh
```diff
@@ -1,39 +1,55 @@
...
-
-if [ $(git tag -l "$TAG") ]; then
-    echo "⚠️ Tag $TAG already exists"
-    exit 1
+echo "Starting a release..."
+echo " "
+echo "This will:"
+echo "- check that you're logged in to npm as the correct user"
+echo "- publish the package if it has not been published already"
+echo "- create a new tag if the current git tag does not match the latest published tag"
```
36degrees: This implies that if the tag already exists this will just do nothing, whereas from my reading it looks like it'll exit out with an error?

nickcolley:
@36degrees good shout with the early return, I've updated this.

PR #1109, opened 2019-01-07
Update visually hidden class to fix ordering issue in VoiceOver OSX

### LINKED ISSUE #1096: Investigate visually hidden text bug in OSX VoiceOver
We've had a report from @stevenaproctor that using visually hidden text results in bugs in VoiceOver.

https://github.com/alphagov/govuk-frontend/pull/958#issuecomment-446523349

We should try and reproduce this, I suspect it may be someone testing VoiceOver in Chrome rather than Safari, but since this is a very widespread pattern on GOV.UK (and GOV.UK Frontend) it's worth an investigation.

nickcolley:
I have investigated this and recorded the results in this gist: https://gist.github.com/nickcolley/19b80ed24d0364cfd3afd3b1b49c4014

We should update the govuk-visually-hidden class to remove the negative margin which will fix the ordering issue in VoiceOver for OSX.

We should recommend using aria-label or aria-labelledby where it makes sense to.

Visually hidden CSS can result in text not being read out when touching on iOS, this is regrettable but does not cause a hard barrier as users can explore the surrounding content to understand context.

Closes #1096

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,14 @@
 
   ([PR #N](https://github.com/alphagov/govuk-frontend/pull/N))
 
+- Update visually hidden class to fix ordering issue in VoiceOver OSX
+
+  We have fixed an issue in VoiceOver OSX where using the `govuk-visually-hidden` class in links would result in the text being announced out of order.
+
+  We also now recommend to use aria-label or aria-labelledby where appropriate.
+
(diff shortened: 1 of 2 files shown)


nickcolley:
Thanks to @stevenaproctor and the accessibility team at HMRC for letting us know about this issue.

PR #1119, opened 2019-01-08
update template.njk to add IE meta tag

bsushmith:
To ensure that older IE versions always render with the correct rendering engine, added a meta tag

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,12 @@
 
   ([PR #1109](https://github.com/alphagov/govuk-frontend/pull/1109))
 
+- Improve rendering in older Internet Explorer
+
+  Added a meta tag to ensure that older IE versions always render with the correct rendering engine
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -22,11 +22,11 @@
 
 🔧 Fixes:
 
-- Pull Request Title goes here
+- update template.njk to add IE meta tag
```
nickcolley: Thanks for picking this up so quickly, it looks really good. I have one small note about the CHANGELOG entry:

We try to keep the template in the CHANGELOG for future people to copy and paste, so maybe we could update this to be:

```suggestion
- Pull Request Title goes here

  Description goes here (optional)

  ([PR #N](https://github.com/alphagov/govuk-frontend/pull/N))

- Improve rendering in older Internet Explorer
```

That said if you want us to do this on your behalf let me know and I'll get this merged in.

bsushmith: Thanks, @nickcolley for the help. I have made the necessary changes. Please review.

PR #1131, opened 2019-01-14
Update summary list to simplify actions

### LINKED ISSUE #1128: Summary list - Can we avoid using a list for a single action?
The primary use case for summary list is Check Answers, which has one action - 'Change'

In the markup, this is wrapped in a list (ul, li), to allow for multiple actions. However in the case of one action, this is arguably not semantically correct, and makes the markup more complicated than it needs to be.

Can we avoid a list when there's only one action?

NickColley:
While my head is still in this component I thought I'd see if we can only output actions in a list when there's multiple actions.

This means in the most common case of a single action the markup is simpler.

Closes #1128

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,12 @@
 
   ([PR #N](https://github.com/alphagov/govuk-frontend/pull/N))
 
+- Update summary list to simplify actions
+
+  Only output actions in a list when there's multiple actions.
+
+  ([PR #1131](https://github.com/alphagov/govuk-frontend/pull/1131))
+
 ## 2.5.0 (Feature release)
 
 🆕 New features:


src/components/summary-list/template.njk
```diff
@@ -1,26 +1,33 @@
...
+{% endmacro -%}
 <dl class="govuk-summary-list {%- if params.classes %} {{ params.classes }}{% endif %}"{% for attribute, value in params.attributes %} {{attribute}}="{{value}}"{% endfor %}>
   {% for row in params.rows %}
     <div class="govuk-summary-list__row">
       <dt class="govuk-summary-list__key {%- if row.key.classes %} {{ row.key.classes }}{% endif %}">
         {{ row.key.html | safe if row.key.html else row.key.text }}
       </dt>
       <dd class="govuk-summary-list__value {%- if row.value.classes %} {{ row.value.classes }}{% endif %}">
-        {{ row.value.html | safe if row.value.html else row.value.text }}
+        {{ row.value.html | indent(8) | trim | safe if row.value.html else row.value.text }}
```
nickcolley: indent and trim filters is just to make the output markup easier to read

joelanman:
Yeh this is great work, thanks @nickcolley !

PR #1144, opened 2019-01-16
Add gutters to the header consistent with the grid

nickcolley:
This allows the header to line up with the grid properly.

### Desktop

<details>
<summary>Before</summary>

desktop-before

</details>

<details>
<summary>After</summary>

desktop-after

</details>

### Above tablet

<details>
<summary>Before</summary>

above-tablet-before

</details>

<details>
<summary>After</summary>

above-tablet-after

</details>

### Tablet

<details>
<summary>Before</summary>

tablet-before

</details>

<details>
<summary>After</summary>

tablet-after

</details>

## To do

- [ ] Test cross browser

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,14 @@
 
   ([PR #1147](https://github.com/alphagov/govuk-frontend/pull/1147))
 
+- Make gutters in the header consistent with the grid
+
+  This means that the header will now line up with the grid.
+
+  Thanks to @edwardhorsford for raising this issue.
+
+  ([PR #1144](https://github.com/alphagov/govuk-frontend/pull/1144))
+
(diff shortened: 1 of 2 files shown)


nickcolley:
Internet Explorer 8 is rendering oddly but I don't think it's related to this change, I'll try and fix that first so we can know that doing this work doesn't break Internet Explorer 8...

https://github.com/alphagov/govuk-frontend/issues/1145

36degrees:
This makes sense to me, but it'd be good if @dashouse could take a look as well.

36degrees:
)Needs a changelog entry)

36degrees:
🚢

nickcolley:
I'm happy this works well in IE8 now that the other bug has been fixed.

Issue #1136, opened 2019-01-14
Make it easy to set autocomplete attributes on form inputs


36degrees:
## What

Make it possible to pass an `autocomplete` option to:

- Text Inputs
- Textareas
- Character Count
- Date Input

When provided, the `autocomplete` option should be outputted as an attribute on the input. For example:

```js
{{ govukInput({
  label: {
    text: "Full name"
  },
  autocomplete: "name",
  id: "name",
  name: "name"
}) }}
```

would output:

```
<div class="govuk-form-group">
  <label class="govuk-label" for="name">
    Full name
  </label>
  <input class="govuk-input" id="name" name="name" type="text" autocomplete="name">
</div>
```

## Why

WCAG 2.1 introduces a new success criteria "[1.3.5 Identify Input Purpose](https://www.w3.org/WAI/WCAG21/Understanding/identify-input-purpose.html)" which "is to ensure that the purpose of a form input collecting information about the user can be programmatically determined, so that user agents can extract and present this purpose to users using different modalities".

Broadly speaking, we can help service teams meet this criteria by making it easier to pass autocomplete attributes to components.

At the minute this can be done by using the `attributes` option, but making it part of the component API promotes the idea that this should be used when it makes sense to do so.

## Done when…

- [x] users can use the autocomplete option with the text input component

  - [x] it's possible to pass an `autocomplete` option to the component
  - [x] there's a test that ensures that the autocomplete attribute is outputted correctly in the HTML
  - [x] the `autocomplete` option is documented in the component's YAML file

---

- [x] users can use the autocomplete option with the textarea component

  - [x] it's possible to pass an `autocomplete` option to the component
  - [x] there's a test that ensures that the autocomplete attribute is outputted correctly in the HTML
  - [x] the `autocomplete` option is documented in the component's YAML file

---

- [x] users can use the autocomplete option with the date input component

  - [x] it's possible to pass an `autocomplete` option to the component
  - [x] there's a test that ensures that the autocomplete attribute is outputted correctly in the HTML
  - [x] the `autocomplete` option is documented in the component's YAML file

PR #1146, opened 2019-01-16
Enable `autocomplete` attributes for input components

hannalaakso:
Set textarea, input and date input components to accept [`autocomplete`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete) option and add tests to ensure the attribute gets rendered.

This helps to surface the new WCAG 2.1 guidelines for "Identify Input Purpose" - see https://github.com/alphagov/govuk-frontend/issues/1136

There's also [a card to document this in the Design System](https://github.com/alphagov/govuk-design-system/issues/696).

Fixes https://github.com/alphagov/govuk-frontend/issues/1136

Edit: The initial PR also updated the character count to do this. This change was subsequently removed as it appears that of the [autofill attributes](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofill), only `street-address` can be multiline and character count wouldn't be used in this scenario.

### DIFF
--- a/src/components/date-input/template.test.js
+++ b/src/components/date-input/template.test.js
@@ -459,4 +459,25 @@ describe('Date input', () => {
     expect($monthInput.hasClass('undefined')).toBeFalsy()
     expect($yearInput.hasClass('undefined')).toBeFalsy()
   })
+
+  describe('when it includes autocomplete attributes', () => {
+    it('renders the autocomplete attribute', () => {
+      const $ = render('date-input', {
+        items: [
+          {
+            'autocomplete': 'bday-day'
+          },
+          {
+            'autocomplete': 'bday-month'
+          },
+          {
+            'autocomplete': 'bday-year'
+          }
+        ]
+      })
+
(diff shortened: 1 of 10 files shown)


src/components/date-input/template.test.js
```diff
@@ -459,4 +459,25 @@ describe('Date input', () => {
...
+            'autocomplete': 'bday-month'
+          },
+          {
+            'autocomplete': 'bday-year'
+          }
+        ]
+      })
+
+      const $firstItems = $('.govuk-date-input__item:first-child input')
+      expect($firstItems.attr('autocomplete')).toEqual('bday-day')
```
nickcolley: Non-blocking: I wonder if we should assert that all three are there...

hannalaakso: Thanks Nick 👍 Going to leave it as it is for now to keep things consistent, there's about 10 tests in this file that follow the same pattern. Maybe it's something we can look at separately.

PR #1147, opened 2019-01-17
Fix container not being centered in IE8

### LINKED ISSUE #1145: Header is misaligned in Internet Explorer 8
I'm not sure if this is a recent regression, but I found that the header doesnt render as I'd expect while working on the spacing (https://github.com/alphagov/govuk-frontend/issues/1122)

screen shot 2019-01-16 at 17 29 27

On the GOV.UK Prototype Kit site there also seems to be misrendering of the header:

screen shot 2019-01-16 at 17 30 21

In internet explorer 9 there are no issues with either.

nickcolley:
This allows browsers that do not support media queries to be centered without a separate stylesheet.

Since the header and the footer component use this container it also fixes centering for these components.

Fixes #1145 

---

# Screenshots

## Before
before

## After
after

### DIFF
--- a/src/objects/_width-container.scss
+++ b/src/objects/_width-container.scss
@@ -2,10 +2,6 @@
   // Limit the width of the container to the page width
   max-width: $govuk-page-width;
 
-  @include govuk-if-ie8 {
-    width: $govuk-page-width;
-  }
-
   // On mobile, add half width gutters
   margin: 0 $govuk-gutter-half;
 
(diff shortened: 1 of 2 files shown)


nickcolley:
Seeing a weird thing in IE8, will try and address...

nickcolley:
I tried to reorder the CSS desktop first but SassMQ smushes 'desktop' media queries into the output CSS which stops that from working...

src/objects/_width-container.scss
```diff
@@ -1,24 +1,27 @@
...
-
-  @include govuk-if-ie8 {
-    width: $govuk-page-width;
-  }
-
+  
   // On mobile, add half width gutters
   margin: 0 $govuk-gutter-half;
-
+
```
36degrees: unintentional whitespace changes?

PR #1156, opened 2019-01-21
Fix Checkboxes and Radios targeting selectors outside it's scope

### LINKED ISSUE #1142: Checkboxes component incorrectly hides elements in certain cases
The [Checkboxes component](https://github.com/alphagov/govuk-frontend/blob/master/src/components/checkboxes/checkboxes.js) adds the css class `govuk-checkboxes__conditional--hidden` to unrelated checkboxes in certain cases.

We found this in `finder-frontend` so have patched it for that repo there (see https://github.com/alphagov/finder-frontend/pull/782) but this could affect others.

### Description of issue

From a quick investigation it looks like the following is happening:

The `handleClick` event calls `setAttributes` with `event.target` as the `$input` argument. However, if jQuery `click` or `trigger('click')` is used, following a click event on a different element, then the `event.target` is the element clicked by the user, not the checkbox. This is a problem because `govuk-checkboxes__conditional--hidden` will then be added to the clicked element, not a checkbox.

Example pseudocode.

```
<checkbox>
<div class='uncheck-checkbox'>Uncheck box</div>
<script>
$('.uncheck-checkbox').on('click', () => {
    $('checkbox').trigger('click');
})
</script>
```

When the div is clicked, then this will hide the div. I haven't tested this in isolation, but I think this is the case given that removing the click event in https://github.com/alphagov/finder-frontend/pull/782 was able to fix this issue for us.

### LINKED ISSUE #1153: Checkbox conditional reveal is not confined to component
Checkboxes have a [conditional reveal option](https://design-system.service.gov.uk/components/checkboxes/conditional-reveal/index.html) that uses the value of the `aria-controls` attribute on the checkbox to apply a class of `govuk-checkboxes__conditional--hidden` to the content being shown or hidden, depending on the checked state of the checkbox.

In GOV.UK we're using the `aria-controls` attribute on finders to refer to a part of the page that shows the number of search results. This is so users of assistive tech have some association between their change (checking the checkbox) and the page updating dynamically. The checkboxes Javascript isn't limited to the scope of the component, so when we use it like this it's appling the conditional class to hide the element, which is not desirable.

Here it is in action:

screen shot 2019-01-21 at 15 06 13

The 'show only brexit results' checkbox has `aria-controls` pointing to the '5 results' part of the page.

Just after we've checked the checkbox, the results count and associated tags disappears:

screen shot 2019-01-21 at 15 06 23

The fix might be to check to see if the element referred to by `aria-controls` is within the component before applying the class.

The JS is here: https://github.com/alphagov/govuk-frontend/blob/master/src/components/checkboxes/checkboxes.js#L39

I need a fix for this fairly quickly so I'll write some temporary CSS to override the class being added by the checkbox JS, but will happily delete that once a solution is found.

Thanks!

nickcolley:
We only want conditional reveals that are within the module markup to be impacted by changes to inputs.

This scopes the querySelector only to this module.

I wasn't sure if I should add more test coverage for this, it feels as though the existing tests should cover this change and to add a test for what we're fixing would be odd since it'd break isolation.

It's also somewhat blocked by the issue that our tests are tied to the frontend review application.

Happy to hear others' thoughts on how/if you can test this.

Fixes #1153 
Fixes #1142

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,12 @@
 
   ([PR #N](https://github.com/alphagov/govuk-frontend/pull/N))
 
+- Fix Checkboxes and Radios targeting selectors outside it's scope
+
+  Thanks to [@andysellick](https://github.com/andysellick) and [@bilbof](https://github.com/bilbof) for helping us with this issue.
+
+  ([PR #1156](https://github.com/alphagov/govuk-frontend/pull/1156))
+
 - Fixes styling of the accordion component when there is no JavaScript or it has been turned off
 
   Thanks @dankmitchell for reporting this issue ([#1130](https://github.com/alphagov/govuk-frontend/issues/1130))
--- a/src/components/checkboxes/checkboxes.js
+++ b/src/components/checkboxes/checkboxes.js
@@ -40,8 +40,10 @@ Checkboxes.prototype.setAttributes = function ($input) {
   var inputIsChecked = $input.checked
   $input.setAttribute('aria-expanded', inputIsChecked)
 
-  var $content = document.querySelector('#' + $input.getAttribute('aria-controls'))
-  $content.classList.toggle('govuk-checkboxes__conditional--hidden', !inputIsChecked)
+  var $content = this.$module.querySelector('#' + $input.getAttribute('aria-controls'))
+  if ($content) {
+    $content.classList.toggle('govuk-checkboxes__conditional--hidden', !inputIsChecked)
+  }
 }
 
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -32,6 +32,10 @@
 
   ([PR #N](https://github.com/alphagov/govuk-frontend/pull/N))
 
+- Fix Checkboxes and Radios targeting selectors outside it's scope
```
nickcolley: Credit Andy and Bill

Issue #1079, opened 2018-11-23
Ability to group radio buttons / checkboxes within sub-headings


frankieroberto:
Hi.

I'm wondering how best to support radio buttons (and checkboxes) which have sub-groups with headings?

Currently, you can use multiple `govukRadios` macros which share the same `name` attribute, however because they use a `data-module` attribute to scope the conditional reveals, these don't work properly across the groups.

Alternatively, you can using a single macro and put headings in using `divider`s, but these don't currently support HTML markup.

On a related not, it's unclear whether the best output would be nested `<fieldset>`s with `<legend>`s, or a single `<fieldset>` with headings and either aria attributes or visually-hidden prefixes to associate labels with their headings.

nickcolley:
Coded example: https://tricky-fisher.glitch.me/

edwardhorsford:
I was looking to do similar today. I've got a set of 10 or so radios and I'd like to group in to 3 related groups. It's all one question - so tried to do it in one radio group with the divider option - but found it is hardcoded to be too narrow.

Also wanted to do similar for checkboxes, but found they don't support divider yet.

timpaul:
Hey there @edwardhorsford, @frankieroberto and @stevenaproctor. Would you be able to share a few use cases of this pattern here, so we can better understand how it's used in a service and what variations a macro would need to support?

Thanks!

edwardhorsford:
Here's an example from our service - radios roughly grouped in to individuals and teams:
<img width="497" alt="screenshot 2019-01-30 at 15 05 21" src="https://user-images.githubusercontent.com/2204224/51990231-89655e80-24a0-11e9-8f16-3e5f76f9fab9.png">

An older design we're not going with, but with a similar need:
<img width="429" alt="screenshot 2019-01-30 at 15 06 30" src="https://user-images.githubusercontent.com/2204224/51990309-aef26800-24a0-11e9-895a-98313dd93597.png">

For the first example, I don't think it's essential for 'teams' to be programmatically linked to the options. For the second it does, or the options would need hidden text to make the team they apply to clear.

frankieroberto:
@timpaul here's the example from the pattern I was looking at:

<img width="480" alt="screenshot 2019-01-30 at 15 12 51" src="https://user-images.githubusercontent.com/30665/51990725-9040a100-24a1-11e9-95d7-3a72797d080f.png">

[...]

PR #1163, opened 2019-01-30
Trigger setAttributes on all radio inputs

frankieroberto:
This enables the conditional reveals to work where two different macros are used with the same name attribute (eg to enable grouping of radio inputs).

One possible fix for #1079.

### DIFF
--- a/src/components/radios/radios.js
+++ b/src/components/radios/radios.js
@@ -45,7 +45,10 @@ Radios.prototype.setAttributes = function ($input) {
 }
 
 Radios.prototype.handleClick = function (event) {
-  nodeListForEach(this.$inputs, function ($input) {
+
+  var allRadioInputs = document.querySelectorAll('input[type="radio"]')
+
+  nodeListForEach(allRadioInputs, function ($input) {
     // If a radio with aria-controls, handle click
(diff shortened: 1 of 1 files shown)


frankieroberto:
@igloosi fixed, thanks.

nickcolley:
Hey Frankie,

Thanks for the suggestion.

I think we need to fix this in a way that doesn't look for inputs on the document, as that's created bugs in the past.

We're going to see what @colinrotherham has found out from his research and consider a fix based on that.

Are you OK for me to close this based on that?

frankieroberto:
@nickcolley can you explain what the issues are? Maybe there's a way to avoid them?

I did consider whether the event could be triggered on only the radio inputs which shared the same `name` attribute, but that was a bigger refactor, and triggering it on all radio buttons didn't seem to have any side-effects.

nickcolley:
You can see this bug: https://github.com/alphagov/govuk-frontend/pull/1156

Let me know if you have any ideas. 👍

nickcolley:
I've spoken to @frankieroberto and since this is not blocking him immediately anymore we're going to close this and action the corresponding issue when we have time https://github.com/alphagov/govuk-frontend/issues/1079.

Thanks @frankieroberto for helping us figure this out.

PR #1167, opened 2019-01-31
Allow describedBy parameter to be passed to the textarea component

### LINKED ISSUE #1106: Character count count message is not associated with textarea
The character count info message is (I believe) meant to be associated with the textarea using aria-describedby. However, this association is only made if there is no existing error message:

https://github.com/alphagov/govuk-frontend/blob/fe36aa381f79dc4ebd2f6cbea65381972179a696/src/components/character-count/character-count.js#L94-L105

We always provide an existing info message as of 7b534bf734cd0a1be2cff34d64018aa89abdabb6:

https://github.com/alphagov/govuk-frontend/blob/fe36aa381f79dc4ebd2f6cbea65381972179a696/src/components/character-count/template.njk#L26-L28

This means that the association will never be created.

From testing with Voiceover / Safari, the limit is still announced when it changes (because of the `aria-live` on the message), but it is no longer announced when focussing the field which I believe it should be.

igloosi:
In order to fix #1106 the textarea component needs to be able to accept the describedBy parameter.

This PR allows for that and the designed behaviour of that parameter is that if the hint text or error message or both exist, they take precedence over the custom describedBy parameter.

~As describedBy is already listed in the table of available options there is no need to add it here~
being removed in [#1181](https://github.com/alphagov/govuk-frontend/pull/1181/commits/7a705425be6b7767dc7f2ab87017e2d45059ed86)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,9 +14,11 @@
 
 🆕 New features:
 
-- Pull Request Title goes here
+- Allow describedBy parameter to be passed to the textarea component
 
-  Description goes here (optional)
+  Designed behaviour of that parameter is that if the hint text or error message or both exist, 
+  
+  they take precedence over the custom describedBy parameter.
 
   ([PR #N](https://github.com/alphagov/govuk-frontend/pull/N))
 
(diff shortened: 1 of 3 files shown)


igloosi:
closing as it will need a redo based on other PRs

Issue #995, opened 2018-09-12
Allow attributes on date input items 


malcolmhire:
Currently, there is no way of adding extra attributes to each of the date input fields, this would be a fairly easy update but the macro already sets attributes on [Line 67](https://github.com/alphagov/govuk-frontend/blob/28ca6d5f61df309c00490f22d809bb6d7aaec7ab/src/components/date-input/template.njk#L67) and as there no built in way to merge objects in Nunjucks (well not that I know of) this would require a filter, but this would need to added application side when initializing Nunjucks.

Any ideas how to achieve this?

igloosi:
Hi @malcolmhire , 
thanks for raising this. As you have correctly identified, there is no built-in way to merge objects in Nunjucks without creating a filter. 
Relying on our users to create a filter, as we can't ship one with a component, isn't something we'd want to do.

I have created a ticket in our backlog to look at this and work out if or how we can do and prioritise accordingly.

joelanman:
I wonder if objects could be merged using a for loop

malcolmhire:
> Hi @malcolmhire ,
> thanks for raising this. As you have correctly identified, there is no built-in way to merge objects in Nunjucks without creating a filter.
> Relying on our users to create a filter, as we can't ship one with a component, isn't something we'd want to do.
> 
> One possible solution would be to move the `pattern` attribute from the object and add it directly to the component, but that might not be as flexible if users wan't to create a custom pattern on data-input. I'm not sure how often that would happen though.
> 
> I have also created a ticket in our backlog so we can track this.

When you say 'directly to the component' do you mean adding it directly to the govukInput? I thought about adding it like that with the pattern as part of the govukInput, this would then leave open attributes object, but this wouldn't be consistent with all the other components so ditched that idea.

@igloosi is this the sort of thing you were meaning?

`{{ govukInput({
        label: {
          text: "foo",
        },
        id: "foo",
        name: "foo"
        value: "bar",
        type: "number",
        pattern: "[0-9]*"
      })`

malcolmhire:
> @malcolmhire something along those lines yes, but you're right, it lacks consistency
> 
> wondering if we can get user attributes object and component's internal object, push both to an array an then internally in the component loop over the array?
> 
> perhaps we might have to allow an array of objects to be passed as attributes
> 
> @malcolmhire
> there is a check in Nunjucks one can do check if array is empty
> 
> ```
> {% if array | length %}
> ```
[...]

PR #1172, opened 2019-02-04
Enable `pattern` attribute for input (+ custom date input attributes)

colinrotherham:
We're converting a service from GOV.UK Elements to GOV.UK Frontend and noticed we can't provide custom attributes to our day/month/year fields.

Here's a pull request to add the feature.

As there's no "object merge" nunjucks filter the `pattern: '[0-9]*'` default attribute must be added manually when the entire attributes object is overridden from the defaults.

### DIFF
--- a/src/components/date-input/template.test.js
+++ b/src/components/date-input/template.test.js
@@ -154,6 +154,20 @@ describe('Date input', () => {
       expect($firstInput.attr('pattern')).toEqual('[0-9]*')
     })
 
+    it('renders inputs with custom pattern attribute', () => {
+      const $ = render('date-input', {
+        items: [
+          {
+            'name': 'day',
+            'pattern': '[0-8]*'
+          }
+        ]
+      })
+
+      const $firstInput = $('.govuk-date-input__item:first-child input')
+      expect($firstInput.attr('pattern')).toEqual('[0-8]*')
+    })
+
     it('renders item with implicit class for label', () => {
       const $ = render('date-input', {
         items: [
--- a/src/components/date-input/template.test.js
+++ b/src/components/date-input/template.test.js
@@ -76,6 +76,39 @@ describe('Date input', () => {
       expect($firstItemInput.attr('name')).toEqual('day')
     })
 
+    it('renders with item attributes', () => {
+      const $ = render('date-input', {
+        items: [
+          {
+            'name': 'day',
+            'attributes': {
+              'data-example-day': 'day'
+            }
+          },
+          {
+            'name': 'month',
+            'attributes': {
+              'data-example-month': 'month'
+            }
+          },
+          {
+            'name': 'year',
+            'attributes': {
+              'data-example-year': 'year'
+            }
+          }
+        ]
+      })
+
+      const $input1 = $('.govuk-date-input__item:nth-of-type(1) input')
+      const $input2 = $('.govuk-date-input__item:nth-of-type(2) input')
+      const $input3 = $('.govuk-date-input__item:nth-of-type(3) input')
+
+      expect($input1.attr('data-example-day')).toEqual('day')
+      expect($input2.attr('data-example-month')).toEqual('month')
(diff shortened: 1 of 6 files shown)


36degrees:
This would fix #995, if we're happy with this as a solution.

I am a little concerned about users forgetting (or not realising) that they need to pass in the `pattern` attribute – would need to be careful about how we documented this.

colinrotherham:
@36degrees Yeah I thought the same.

I added a new global called `merge` which combined the custom attribute over the defaults, but then realised most people's nunjucks setups would need to re-add the global.

Would have been like:

```njk
{
  autocomplete: item.autocomplete,
  attributes: merge(item.attributes, { pattern: "[0-9]*" })
}
```

Sadly nunjucks doesn't provide object merging.

malcolmhire:
This is the same way I have done it in my forked version of `govuk-frontend`, will be interesting if this gets accepted as I was also concerned that users might forget to add the `pattern` attribute.

colinrotherham:
Rebased with latest `master`

nickcolley:
Hanna and I have reviewed this, we like this approach but want to consider the others too.

If we were to go with this approach we'd need to update the YAML description to be clear that setting this attributes requires you to add `pattern` back in.

You can see https://github.com/alphagov/govuk-frontend/issues/995#issuecomment-467469735 for more details on what we've talked about.

colinrotherham:
@nickcolley @hannalaakso If you'd like to go with this approach, I've updated both the yaml description and changelog entry to include `pattern: '[0-9]*'`.

Looking forward to seeing `__additionalAttributes` and/or a merge global filter in future 👍

nickcolley:
I've talked this through with the team and we think there's an alternative to this that avoids some of the negative aspects of this approach without making breaking changes: 
https://github.com/alphagov/govuk-frontend/issues/995#issuecomment-469630615

src/components/input/input.yaml
```diff
@@ -162,3 +166,11 @@ examples:
       id: input-with-autocomplete-attribute
       name: postcode
       autocomplete: postal-code
+  - name: with pattern attribute
```
nickcolley: Non-blocking for the future, we tend to only add examples to render if they can't be fully tested using the unit tests.

colinrotherham: Sorry, sort of! I wanted to confirm it popped up the number keyboard on iOS (to ensure the new attributes hadn't overridden the pattern).

I can remove it again? It works fine.

test number keypad

nickcolley: Yeah I think it's up to you, just letting you know when we tend to add examples. 👍

src/components/date-input/template.test.js
```diff
@@ -74,6 +74,40 @@ describe('Date input', () => {
 
       expect($items.length).toEqual(3)
       expect($firstItemInput.attr('name')).toEqual('day')
+      expect($firstItemInput.attr('pattern')).toEqual('[0-9]*')
```
36degrees: Non-blocking, but this feels like it might be important enough to be split out into its own test (I think we're also trying to have one expectation per test, but we're not super strict on that)

nickcolley: That seems to be tested already in "renders inputs with pattern="[0-9]*" to trigger numeric keypad on iOS", so we could consider removing this assertion here.

36degrees: Good spot @nickcolley! 👍Looks like this is duplicated in that case?

colinrotherham: No problem, now removed! Thanks both

colinrotherham:
Thanks @36degrees.

I've removed that extra assertion, it was definitely already covered.

All ready to merge if you're happy 😊

nickcolley:
Going to run out of superlatives to describe these pull requests thanks so much for your help Colin.

👏

PR #1173, opened 2019-02-05
WIP: Enable error summary autofocus disable

igloosi:
any idea how to test this scenario in the app?

`new ErrorSummary($errorSummary).init({ autofocus: false })`

cc: @nickcolley

### DIFF
--- a/src/components/error-summary/error-summary.js
+++ b/src/components/error-summary/error-summary.js
@@ -2,8 +2,9 @@ import '../../vendor/polyfills/Function/prototype/bind'
 import '../../vendor/polyfills/Event' // addEventListener
 import '../../vendor/polyfills/Element/prototype/closest'
 
-function ErrorSummary ($module) {
+function ErrorSummary ($module, options) {
   this.$module = $module
+  this.options = options || {}
 }
 
 ErrorSummary.prototype.init = function () {
@@ -11,9 +12,13 @@ ErrorSummary.prototype.init = function () {
   if (!$module) {
     return
   }
-  window.addEventListener('load', function () {
-    $module.focus()
-  })
+  var autofocus = this.options.autofocus === false ? this.options.autofocus : true
+
+  if (autofocus) {
+    window.addEventListener('load', function () {
+      $module.focus()
+    })
+  }
 
   $module.addEventListener('click', this.handleClick.bind(this))
 }


src/components/error-summary/error-summary.js
```diff
@@ -6,14 +6,18 @@ function ErrorSummary ($module) {
   this.$module = $module
 }
 
-ErrorSummary.prototype.init = function () {
+ErrorSummary.prototype.init = function (args = {}) {
```
nickcolley: This looks like ES2015 code to me...

nickcolley:
I thought it might make more sense to be something like

```
new ErrorSummary($errorSummary, { autofocus: false }).init()
```

Where the second argument is stored on the instance as 'options'

nickcolley:
I think bootstrap does a good job of having both with a data-api and programmatic api:

https://getbootstrap.com/docs/3.4/javascript/#js-data-attrs

So we could have a convention for components:

```js
new ErrorSummary($errorSummary, { autofocus: false }).init()
```

or

```html
<div data-module="error-summary" data-autofocus=false></div>
```

@alex-ju what do you think? Maybe too over the top?

alex-ju:
@nickcolley I'm all for keeping having both programatically and data-api for all components and all options – I initially did the same with the character count (and other components), but while contributing it I ended up giving up on one of them as we were debating on which one should superseed the other – so we only kept the data-api version.

36degrees:
Yeah, I think even if we support data attributes we'll need _some way_ to do this without changing the macros / HTML shown in the code examples.

alex-ju:
@igloosi I think we went into styling instead of answering your question. what do you mean by testing? having a visual example or unit/functional test it?

nickcolley:
You can test this by looking at `document.activeElement` which tells you which element is currently focused.

36degrees:
We have a [test that the ErrorSummary _is_ focussed already](https://github.com/alphagov/govuk-frontend/blob/master/src/components/error-summary/error-summary.test.js#L25-L30) so that would be a sensible place to start.

nickcolley:
I did a quick spike to explore having a data api that mirrors options...

I'd be happy to put together a proper architecture proposal for this, not sure if it should block this feature.

https://jsbin.com/gumibix/14/edit?html,js,console,output

igloosi:
I'll close the PR until it's decided in the team then

nickcolley:
Thanks Jani!

PR #1176, opened 2019-02-06
Accommodate camera notches on new devices (iPhone X, Google Pixel 3 etc)

colinrotherham:
Noticed this a while ago, thought I'd contribute a new feature for it.

**GOV.UK Frontend** doesn't support the dreaded "notch" (or Display Cutouts as they're known) on various new devices. See the unused white area on the left hand side here:

notch safe area

I did a quick check and we now have:

* iPhone X
* iPhone Xs
* Google Pixel 3
* OnePlus 6
* Huawei P20
* LG G7
* Nokia 6.1

The `safe-area-inset-*` variables were [added in iOS 11](https://webkit.org/blog/7929/designing-websites-for-iphone-x/), and recently Android P has added support via Chrome for Android 69: https://www.chromestatus.com/feature/5710044637167616

Firefox just added support in v65:
https://bugzilla.mozilla.org/show_bug.cgi?id=1462233

### Before
before

### After
after

This pull request adds the initial support to patch `.govuk-width-container`

### DIFF
--- a/src/components/skip-link/_skip-link.scss
+++ b/src/components/skip-link/_skip-link.scss
@@ -11,5 +11,16 @@
 
     display: block;
     padding: govuk-spacing(2) govuk-spacing(3);
+
+    // Respect 'display cutout' safe area (avoids notches and rounded corners)
+    @supports (padding: unquote("max(calc(0px))")) {
+      $padding-safe-area-right: calc(#{govuk-spacing(3)} + env(safe-area-inset-right));
+      $padding-safe-area-left: calc(#{govuk-spacing(3)} + env(safe-area-inset-left));
+
+      // Use max() to pick largest padding, default or with safe area
+      // Escaped due to Sass max() vs. CSS native max()
+      padding-right: unquote("max(#{govuk-spacing(3)}, #{$padding-safe-area-right})");
+      padding-left: unquote("max(#{govuk-spacing(3)}, #{$padding-safe-area-left})");
+    }
   }
 }
--- a/src/objects/_width-container.scss
+++ b/src/objects/_width-container.scss
@@ -5,15 +5,43 @@
   // On mobile, add half width gutters
   margin: 0 $govuk-gutter-half;
 
+  // Respect 'display cutout' safe area (avoids notches and rounded corners)
+  @supports (margin: unquote("max(calc(0px))")) {
+    $gutter-safe-area-right: calc(#{$govuk-gutter-half} + env(safe-area-inset-right));
+    $gutter-safe-area-left: calc(#{$govuk-gutter-half} + env(safe-area-inset-left));
+
+    // Use max() to pick largest margin, default or with safe area
+    // Escaped due to Sass max() vs. CSS native max()
+    margin-right: unquote("max(#{$govuk-gutter-half}, #{$gutter-safe-area-right})");
+    margin-left: unquote("max(#{$govuk-gutter-half}, #{$gutter-safe-area-left})");
+  }
+
   // On tablet, add full width gutters
   @include govuk-media-query($from: tablet) {
     margin: 0 $govuk-gutter;
+
+    // Respect 'display cutout' safe area (avoids notches and rounded corners)
+    @supports (margin: unquote("max(calc(0px))")) {
+      $gutter-safe-area-right: calc(#{$govuk-gutter-half} + env(safe-area-inset-right));
+      $gutter-safe-area-left: calc(#{$govuk-gutter-half} + env(safe-area-inset-left));
+
+      // Use max() to pick largest margin, default or with safe area
+      // Escaped due to Sass max() vs. CSS native max()
+      margin-right: unquote("max(#{$govuk-gutter}, #{$gutter-safe-area-right})");
+      margin-left: unquote("max(#{$govuk-gutter}, #{$gutter-safe-area-left})");
+    }
   }
 
   // As soon as the viewport is greater than the width of the page plus the
   // gutters, just centre the content instead of adding gutters.
   @include govuk-media-query($and: "(min-width: #{($govuk-page-width + $govuk-gutter * 2)})") {
     margin: 0 auto;
+
+    // Since a safe area may have previously been set above,
+    // we need to duplicate this margin that centers the page.
+    @supports (margin: unquote("max(calc(0px))")) {
+      margin: 0 auto;
+    }
   }
 
   @include govuk-if-ie8 {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,15 @@
 
   ([PR #N](https://github.com/alphagov/govuk-frontend/pull/N))
 
+- Accommodate camera notches on new devices (iPhone X, Google Pixel 3 etc)
+
+  On newer devices with "camera notches", browsers reserve a safe area in landscape orientation (known as pillarboxing) so content isn't obscured.
+
+  To avoid this, support has been added for `viewport-fit=cover` as shown here:
+  https://webkit.org/blog/7929/designing-websites-for-iphone-x/
+
+  ([PR #1176](https://github.com/alphagov/govuk-frontend/pull/1176))
+
 🔧 Fixes:
 
 - Pull Request Title goes here
--- a/docs/installation/examples/webpack/index.html
+++ b/docs/installation/examples/webpack/index.html
@@ -4,7 +4,7 @@
   <meta charset="utf-8">
   <meta http-equiv="X-UA-Compatible" content="IE=edge">
   <title>GOV.UK Frontend</title>
-  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <meta name="viewport" content="width=device-width, initial-scale=1, viewport-fit=cover">
   <!--[if !IE 8]><!-->
     <link rel="stylesheet" href="../../public/css/app.css">
   <!--<![endif]-->
--- a/src/template.njk
+++ b/src/template.njk
@@ -8,7 +8,7 @@
   <head>
     <meta charset="utf-8" />
     <title>{% block pageTitle %}GOV.UK - The best place to find government services and information{% endblock %}</title>
-    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <meta name="viewport" content="width=device-width, initial-scale=1, viewport-fit=cover">
     <meta name="theme-color" content="{{ themeColor | default('#0b0c0c') }}" /> {# Hardcoded value of $govuk-black #}
     {# Ensure that older IE versions always render with the correct rendering engine #}
     <meta http-equiv="X-UA-Compatible" content="IE=edge" />


src/components/skip-link/_skip-link.scss
```diff
@@ -11,5 +11,11 @@
 
     display: block;
     padding: govuk-spacing(2) govuk-spacing(3);
+
+    // Add safe area (e.g. iPhone X, Google Pixel 3)
+    @supports (padding: env(safe-area-inset-left)) {
```
nickcolley: Does this change impact the padding when the phone is portrait?

nickcolley:
Something to think about:

> The env() function shipped in iOS 11 with the name constant(). Beginning with Safari Technology Preview 41 and the iOS 11.2 beta, constant() has been removed and replaced with env(). You can use the CSS fallback mechanism to support both versions, if necessary, but should prefer env() going forward.

src/components/skip-link/_skip-link.scss
nickcolley: Realised this is what the 'max' is for 👍

colinrotherham: Yeah, although technically both portrait and landscape could have an inset.

For example, a circular Google Wear OS watch screen. It could be offset all around to avoid going to close to the edges of the circle.

colinrotherham: More info on that here: https://drafts.csswg.org/css-round-display/

colinrotherham:
@nickcolley Yeah Apple quickly backtracked on `constant()` as its quite rightly a variable! I chose not to support iOS 11.0 and 11.1 and only target the standard spec instead.

Thanks for the comment about `@supports` and variables though, I was taking my guidance from the [WebKit article](added in iOS 11) which suggests doing that.

I've added a new version with a simpler test:

```css
/* Add safe area (e.g. iPhone X, Google Pixel 3) */
@supports (padding: unquote("max(0px)")) {
  /* Use safe area or mobile margin, whichever is larger */
  /* Escaped due to Sass max() vs. CSS native max() */
  margin-right: unquote("max(#{$govuk-gutter-half}, env(safe-area-inset-right))");
  margin-left: unquote("max(#{$govuk-gutter-half}, env(safe-area-inset-left))");
}
```

Since Sass already has a `max()` we had to do some escaping to avoid it.

This example outputs the following CSS:

```css
@supports (padding: max(0px)) {
  .govuk-width-container {
    margin-right: max(15px, env(safe-area-inset-right));
    margin-left: max(15px, env(safe-area-inset-left));
  }
}
```

colinrotherham:
If anyone has a phone with a "notch" it's deployed here to test out:
https://govuk-frontend-review-pr-1176.herokuapp.com/examples/template-default

Should look like the **After** screenshot above.

Thanks

paulwaitehomeoffice:
Looks good to me (iPhone XS Max, iOS 12.1.2).

I’ve got an old iPhone X at home which I can probably restore to an earlier version of iOS tonight, if it'd help to test there too.

36degrees:
Over the next few weeks we're planning to add a number of example pages to the review app in preparation for an audit against WCAG 2.1.

If it's OK with you, I think it'd be worth waiting until we've got those example pages added, as I think they'll be an ideal way to test that this going to do what we want in different contexts.

colinrotherham:
@36degrees Sounds good to me. I'll keep this rebased with `master` in the meantime.

colinrotherham:
Rebased with latest `master`.

Good to see it in action with the new full page example @nickcolley.

start page

nickcolley:
@colinrotherham exciting, we will have a lot more examples in the review app by the end of the week, then we can merge this in once we're happy :)

nickcolley:
We've update the review application to include pages that cover as much as GOV.UK Frontend as possible so I think this is ready for review based on that, if you could bring this branch up to date with the current repository (master). 😄

colinrotherham:
@nickcolley Done 😊

nickcolley:
I'm wondering if it's a bit too tight at the moment? I've found one example where number lists can be blocked slightly...

Pinging @dashouse to see what he thinks on the spacing.

https://govuk-frontend-review.herokuapp.com/full-page-examples/announcements
screen shot 2019-02-26 at 17 29 00

colinrotherham:
@nickcolley These new example pages are so much better. Hadn't realised the numbered lists hang slightly off to the left—definitely looks too tight now! Should they hang off though?

I've chosen to use `$govuk-gutter-half` instead? Feels better (same size too?)

```scss
// Add safe area (e.g. iPhone X, Google Pixel 3)
@supports (padding: unquote("max(calc(0px))")) {
  // Use safe area or default padding, whichever is larger
  // Escaped due to Sass max() vs. CSS native max()
  padding-right: unquote("max(#{govuk-spacing(3)}, calc(#{$govuk-gutter-half} + env(safe-area-inset-right)))");
  padding-left: unquote("max(#{govuk-spacing(3)}, calc(#{$govuk-gutter-half} + env(safe-area-inset-left)))");
}
```

I've pushed up what you see above (added both max/calc into `@supports` but made sure the variable name isn't included) which you spotted in https://github.com/alphagov/govuk-frontend/pull/1176#issuecomment-461104250.

Same logic applied to the skip link should it be visible on other notched devices in future.

It's looking good! Thanks

colinrotherham:
Can we trigger a new Heroku build too please?

hannalaakso:
@colinrotherham Done 👍

dashouse:
Sorry I'm late to comment. The `<ol>` style doesn't look great to me in general (nothing to do with this PR).

However, without doing much of a deep dive the only way I can think to fix this would be to change the padding-left to a number that would stop the numeric bullets breaking out of the layout (when two digits).

I was playing with `list-style-position: inside;` and losing the padding all together but in that scenario the list content wraps under the number despite it fixing the overflow.

nickcolley:
@dashouse sounds good to raise as a separate issue regarding those specific list styles then...

Could you review the spacing that Colin has proposed originally and give any feedback on what you think we should be aiming for?

dashouse:
I think the changes that were made or this reason look really good (based on the screen shots). So I'd go with the bigger gutter as suggested anyway and we'll look at the <ol> style separately too.

nickcolley:
@dashouse sweet thanks for that, feel free to suggest any specific spacing as I kinda just fudged it and went with what felt OK.

dashouse:
Really happy with it feel wise in browser stack, can't really see the outputted units but for something like this I think it's as much about feel as anything...

colinrotherham:
@dashouse @nickcolley That's fab, glad you like it.

The current pull request includes Nick's suggestion (albeit using `$govuk-gutter-half` vs. `govuk-spacing(3)`) but let me know if you need anything else 😊

src/objects/_width-container.scss
```diff
@@ -5,15 +5,37 @@
...
+
   // On tablet, add full width gutters
   @include govuk-media-query($from: tablet) {
     margin: 0 $govuk-gutter;
+
+    // Add safe area (e.g. iPhone X, Google Pixel 3)
+    @supports (margin: unquote("max(calc(0px))")) {
+      // Use safe area or tablet margin, whichever is larger
+      // Escaped due to Sass max() vs. CSS native max()
+      margin-right: unquote("max(#{$govuk-gutter}, calc(#{$govuk-gutter-half} + env(safe-area-inset-right)))");
```
36degrees: Is this intentionally using `$govuk-gutter-half` rather than `$govuk-gutter` within the calc?

src/objects/_width-container.scss
```diff
@@ -5,15 +5,37 @@
...
+    // Escaped due to Sass max() vs. CSS native max()
+    margin-right: unquote("max(#{$govuk-gutter-half}, calc(#{$govuk-gutter-half} + env(safe-area-inset-right)))");
+    margin-left: unquote("max(#{$govuk-gutter-half}, calc(#{$govuk-gutter-half} + env(safe-area-inset-left)))");
+  }
+
   // On tablet, add full width gutters
   @include govuk-media-query($from: tablet) {
     margin: 0 $govuk-gutter;
+
+    // Add safe area (e.g. iPhone X, Google Pixel 3)
```
36degrees: AFAIK neither of the devices listed here would actually match this because it's at the tablet breakpoint – should the comment be clearer about this?

```suggestion
    // Respect 'safe area' on tablet devices to avoid placing content in e.g. notches and rounded corners
```

src/components/skip-link/_skip-link.scss
```diff
@@ -11,5 +11,13 @@
 
     display: block;
     padding: govuk-spacing(2) govuk-spacing(3);
+
+    // Add safe area (e.g. iPhone X, Google Pixel 3)
+    @supports (padding: unquote("max(calc(0px))")) {
+      // Use safe area or default padding, whichever is larger
+      // Escaped due to Sass max() vs. CSS native max()
+      padding-right: unquote("max(#{govuk-spacing(3)}, calc(#{$govuk-gutter-half} + env(safe-area-inset-right)))");
```
36degrees: Should we be using `govuk-spacing(3)` within the calc as well?

colinrotherham: Anything that ends in "px" triggers a match here, happy to if you prefer?

src/objects/_width-container.scss
colinrotherham: It is yeah. It's equals "notch safe area plus a little bit more" for aesthetics.

Adding the full gutter was too much.

src/objects/_width-container.scss
colinrotherham: Yeah, appreciate this is a hypothetical "future" notched tablet! Just in case.

Might have one arrive some day.

These devices might be long forgotten in future so shall I change all the comments if you prefer? Thanks

src/components/skip-link/_skip-link.scss
nickcolley: Non-blocking, but it might make it more obvious where these values come from if we set some private variables that can be shared between the first definition and the notch definitions?

src/objects/_width-container.scss
36degrees: Just to be clear, I'm asking if it's intentional that we're using _both_ `$govuk-gutter` and `$govuk-gutter-half` on this line – on the two different sides of the `max` calculation.

src/components/skip-link/_skip-link.scss
36degrees: Just to be clear, I'm asking if it's intentional that we're using _both_ `$govuk-spacing(3)` and `$govuk-gutter-half` on this line – on the two different sides of the `max` calculation.

src/objects/_width-container.scss
nickcolley: I think maybe clearer comments or new private variables may make this clearer...

src/components/skip-link/_skip-link.scss
nickcolley: I believe the logic will go something like this:

Does the device have an inset?

If yes, then we want that inset + a reduced gutter, since a regular gutter size is optically too large.

If there's no inset, then use the regular gutter.

src/components/skip-link/_skip-link.scss
colinrotherham: Resolving this, all questions answered

src/components/skip-link/_skip-link.scss
colinrotherham: I'm using `$govuk-spacing(3)` only for **_skip-link.scss**
I'm using `$govuk-gutter[-half]` only for **_width-container.scss**

Cheers for spotting where I'd used both in places.

This diff is outdated, so add a new comment if you like?

src/objects/_width-container.scss
colinrotherham: I've changed all the padding/margin overrides to use variables now:

```css
margin-right: unquote("max(#{$govuk-gutter}, #{$gutter-safe-area-right})");
margin-left: unquote("max(#{$govuk-gutter}, #{$gutter-safe-area-left})");
```

This conversation is on an outdated diff so open another comment if you've got some more feedback? I'm happy to fit in with both of you @nickcolley @36degrees.

src/objects/_width-container.scss
colinrotherham: Using a new comment for all notch @supports queries now:

```
// Respect 'display cutout' safe area (avoids notches and rounded corners)
```

colinrotherham:
I've resolved all the review comments.

They're on outdated diffs so add more if you need any other tweaks!

Thanks

36degrees:
Great stuff, thanks @colinrotherham 👍

nickcolley:
Will try to release this next week :+1:

PR #1181, opened 2019-02-08
Update documentation for component options

36degrees:
This corrects a bunch of issues with the documentation for our components, off the back of [some exploratory work into testing the documentation](https://github.com/alphagov/govuk-frontend/compare/update-options?expand=1) – which is really hard, and not something we're ready to merge yet.

See the individual commits for details.

### DIFF
--- a/src/components/footer/template.njk
+++ b/src/components/footer/template.njk
@@ -3,17 +3,17 @@
   <div class="govuk-width-container {{ params.containerClasses if params.containerClasses }}">
     {% if params.navigation %}
       <div class="govuk-footer__navigation">
-        {% for item in params.navigation %}
+        {% for nav in params.navigation %}
           <div class="govuk-footer__section">
-            <h2 class="govuk-footer__heading govuk-heading-m">{{ item.title }}</h2>
-            {% if item.items %}
+            <h2 class="govuk-footer__heading govuk-heading-m">{{ nav.title }}</h2>
+            {% if nav.items %}
               {% set listClasses %}
-                {% if item.columns %}
-                  govuk-footer__list--columns-{{ item.columns }}
+                {% if nav.columns %}
+                  govuk-footer__list--columns-{{ nav.columns }}
                 {% endif %}
               {% endset %}
               <ul class="govuk-footer__list {{ listClasses | trim }}">
-                {% for item in item.items %}
+                {% for item in nav.items %}
                   {% if item.href and item.text %}
                     <li class="govuk-footer__list-item">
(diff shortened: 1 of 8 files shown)


src/components/footer/template.njk
```diff
@@ -3,17 +3,17 @@
   <div class="govuk-width-container {{ params.containerClasses if params.containerClasses }}">
     {% if params.navigation %}
       <div class="govuk-footer__navigation">
-        {% for item in params.navigation %}
```
nickcolley: The tests have failed on this change, but it looks sensible so not sure why.

igloosi:
outputed class is `govuk-footer__list--columns-`
because
```
{% if nav.columns %}
      govuk-footer__list--columns-{{ **item**.columns }}
{% endif %}
```

36degrees:
Good catch, thanks @igloosi 👍

src/components/footer/template.njk
```diff
@@ -3,17 +3,17 @@
...
+        {% for nav in params.navigation %}
           <div class="govuk-footer__section">
-            <h2 class="govuk-footer__heading govuk-heading-m">{{ item.title }}</h2>
-            {% if item.items %}
+            <h2 class="govuk-footer__heading govuk-heading-m">{{ nav.title }}</h2>
+            {% if nav.items %}
               {% set listClasses %}
-                {% if item.columns %}
+                {% if nav.columns %}
                   govuk-footer__list--columns-{{ item.columns }}
```
hannalaakso: This will just need updating to `{{ nav.columns }}` and then the tests will pass again

src/components/footer/template.njk
```diff
@@ -3,17 +3,17 @@
...
-        {% for item in params.navigation %}
+        {% for nav in params.navigation %}
           <div class="govuk-footer__section">
-            <h2 class="govuk-footer__heading govuk-heading-m">{{ item.title }}</h2>
-            {% if item.items %}
+            <h2 class="govuk-footer__heading govuk-heading-m">{{ nav.title }}</h2>
+            {% if nav.items %}
               {% set listClasses %}
-                {% if item.columns %}
+                {% if nav.columns %}
```
hannalaakso: The below line will just need updating to `{{ nav.columns }}` and then the tests will pass again

igloosi:
technically I think this should be a fix in the changelog, but it depends on the notion of macro-options.json being part of the public API

nickcolley:
macro-options is not documented so it's not part of the public api, updating the template.njk did not change anything for users or the api.

So given that I'd argue these are all internal changes?

igloosi:
dilemma solved

PR #1184, opened 2019-02-13
Allow for full page examples

nickcolley:
In order to have more realistic examples to audit we want to put full page examples into the review application that's only used by the core team of GOV.UK Frontend.

This can also be used to test changes in the future for example focus states, and would be useful with a visual regression setup.

This initial pull request includes a start page example.

Since we're going to have full examples this also adds a big banner that can be dismissed:

screen shot 2019-02-13 at 13 54 44


nickcolley:
I'm just trying to add some test coverage for the banner, then I'll merge 👍

PR #1185, opened 2019-02-14
Prevent summary list from wrapping mid-word for most browsers

MoJ-Longbeard:
As discussed with [Colin Rotherham](https://github.com/colinrotherham)

[PR #1169](https://github.com/alphagov/govuk-frontend/pull/1169) added in some CSS:
- ``word-break: break-word;``, followed by
- ``word-break: break-all;``

``word-break: break-word;`` is the favoured behaviour, but isn't fully supported.  
``word-break: break-all;`` is the fallback behaviour, it sometimes results in unfavourable wrapping mid-word.  

The order for this CSS needs to be changed to, otherwise the less favourable `break-all` takes precedence.
```CSS
word-break: break-all;
word-break: break-word;
```

**caniuse** said:
> Chrome, Safari and other WebKit/Blink browsers also support the unofficial break-word value

The implication here being that it is not fully supported, so ``word-break: break-all;`` is still needed as a fallback.

Below screenshot shows the difference in Google Chrome, note the wrapping of the first column.
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,14 @@
 
   ([PR #N](https://github.com/alphagov/govuk-frontend/pull/N))
 
+
+- Prevent summary list from wrapping mid-word for most browsers
+
+  Thanks to [Malcolm Butler](https://github.com/MoJ-Longbeard) for fixing this issue!
+
+  ([PR #1185](https://github.com/alphagov/govuk-frontend/pull/1185))
+  
+
 ## 2.7.0 (Feature release)
 
 🆕 New features:
--- a/src/components/summary-list/_summary-list.scss
+++ b/src/components/summary-list/_summary-list.scss
@@ -58,8 +58,8 @@
   .govuk-summary-list__value {
     // sass-lint:disable no-duplicate-properties
     // Automatic wrapping for unbreakable text (e.g. URLs)
-    word-break: break-word; // WebKit/Blink only
     word-break: break-all; // Standards
+    word-break: break-word; // WebKit/Blink only
   }
 
   .govuk-summary-list__key {


nickcolley:
@MoJ-Longbeard happy to accept this but would be great to see examples of the impact of this changes if you can?

We also will need a CHANGELOG entry to describe this fix (make sure to put recognition for yourself as part of the notes too please :+1: )

MoJ-Longbeard:
@nickcolley 
Done - I think.

CHANGELOG.md
```diff
@@ -22,11 +22,13 @@
 
 🔧 Fixes:
 
-- Pull Request Title goes here
+- Update \_summary-list.scss
```
nickcolley:
```suggestion
- Pull Request Title goes here

  Description goes here (optional)

  ([PR #N](https://github.com/alphagov/govuk-frontend/pull/N))
 
- Minor CSS code change to prevent wrapping mid-word for most browsers
```

nickcolley: It's not super obvious but we try and keep the template around for contributors in the future to copy and paste.

Also I think the sentence below 'update summary list' is a more descriptive title we can use.

Happy to do this for you when it gets merged, let me know.

MoJ-Longbeard: @nickcolley , sure.  Please do that for me, happy with the description change.  I'm still getting to grips with GitHub etc...

nickcolley:
I've cleaned up the branch for you :)

Going to wait for an approval from @colinrotherham since he's been thinking about this a lot too.

colinrotherham:
For anyone else looking, we originally referred to https://caniuse.com/#search=word-break

>Chrome, Safari and other WebKit/Blink browsers also support the unofficial break-word value which is treated like word-wrap: break-word.

And the `non-standard` comment by [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/word-break):

```css
/* Keyword values */
word-break: normal; 
word-break: break-all; 
word-break: keep-all;
word-break: break-word; /* non-standard */ 
```

We made sure that the standard is preferred when available.

But… thanks to research by @MoJ-Longbeard we now know that `break-word` has better wrapping behaviour and is less likely to break up words, unless absolutely necessary.

We also found it's quite clearly listed in the latest CSS Text Module Level 3 working draft from December 2018: https://www.w3.org/TR/css-text-3/#overflow-wrap-property 😊 

@nickcolley So this is a 👍 from me!

PR #1186, opened 2019-02-14
Fix error summary linking on the example pages

colinrotherham:
I noticed these two pages don't adhere to https://github.com/alphagov/govuk-frontend/pull/1056 now @36degrees added in the focus handling.

http://govuk-frontend-review.herokuapp.com/examples/error-summary-with-messages
http://govuk-frontend-review.herokuapp.com/examples/error-summary-with-one-thing-per-page

The error summary links should go to specific inputs.

This is now fixed.

But, it opens up another question though:
**Can date inputs with two invalid fields have two error messages like this?**

date inputs

Looking at these notes:
https://design-system.service.gov.uk/components/error-summary/

It sounds like the date component can have multiple fields with errors.

>When a user has to enter their answer into multiple fields, such as the day, month and year fields in the date input component, link to the first field that contains an error.
>
>If you do not know which field contains an error, link to the first field.

Our content designer would love to know.

### DIFF
--- a/app/views/examples/error-summary-with-messages/index.njk
+++ b/app/views/examples/error-summary-with-messages/index.njk
@@ -21,12 +21,12 @@
     "classes": "optional-extra-class",
     "errorList": [
       {
-        "text": "Descriptive link to the question with an error",
-        "href": "#example-error-1"
+        "text": "You must provide your passport number",
+        "href": "#passport-number"
       },
       {
-        "text": "Descriptive link to the question with an error",
-        "href": "#example-error-1"
+        "text": "You must provide your expiry date",
+        "href": "#expiry-month"
       }
     ]
   }) }}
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -28,6 +28,13 @@
 
   ([PR #N](https://github.com/alphagov/govuk-frontend/pull/N))
 
+- Fix error summary linking on the example pages
```
nickcolley: We don't need to add a CHANGELOG since these are review pages for maintainers (us and you) not public guidance. :+1:

colinrotherham: Ah ha, good to know. Thanks.

colinrotherham:
@nickcolley Had a chat with some content designers on the cross-government Slack.

Since the date input should only have one error ([see here](https://design-system.service.gov.uk/components/date-input/#error-messages)), I've changed the `error-summary-with-one-thing-per-page` example to only have one error summary link.

Link hrefs are all fixed still.

updated example

colinrotherham:
Error summary links all work now:

https://govuk-frontend-review-pr-1186.herokuapp.com/examples/error-summary-with-messages
https://govuk-frontend-review-pr-1186.herokuapp.com/examples/error-summary-with-one-thing-per-page

nickcolley:
@colinrotherham I've moved your more general question into it's own issue, so we can merge this and then answer that separately :)

https://github.com/alphagov/govuk-frontend/issues/1199

PR #1188, opened 2019-02-14
Update 'classes' and 'attributes' argument descriptions for the inset-text component

igloosi:


nickcolley:
Nice one Jani thanks

PR #1190, opened 2019-02-14
Update 'classes' argument description for the select component

igloosi:


src/components/select/select.yaml
```diff
@@ -59,7 +59,7 @@ params:
 - name: classes
   type: string
   required: false
-  description: Classes to add to the anchor tag.
+  description: Optional classes to add to the select.
```
joelanman: Just for consistency - I don't think we normally say 'Optional' here

PR #1191, opened 2019-02-14
Check your answers, bank holidays and service manual topic example pages

hannalaakso:
Adds full page examples  for testing/auditing purposes.

### DIFF
--- a/app/views/full-page-examples/service-manual-topic/index.njk
+++ b/app/views/full-page-examples/service-manual-topic/index.njk
@@ -1,0 +1,274 @@
+{# This example is based of the live "Service Manual - Technology" page: https://www.gov.uk/service-manual/technology #}
+{% extends "_generic.njk" %}
+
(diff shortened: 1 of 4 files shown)


app/views/full-page-examples/service-manual-topic/index.njk
```diff
@@ -0,0 +1,274 @@
+{# This example is based of the live "UK bank holidays" page: https://www.gov.uk/service-manual/technology #}
```
36degrees: No it's not 😛

nickcolley:
We can also add these URLs' to the smoke tests in `app/__tests__` alongside the other examples.

hannalaakso:
Thanks @nickcolley 🙌  

I've added tests and now using `{% set %}` for HTML variables.

app/views/full-page-examples/service-manual-topic/index.njk
hannalaakso: 😄

PR #1195, opened 2019-02-15
Add Sign in, Confirm nationality, Confirm postcode, Announcement templates

hannalaakso:
This adds full page templates, and tests, for the purposes of accessibility audit:

Sign in page that demonstrates
- Radios
- Radios with hints

Confirm nationality that demonstrates:
- Checkboxes
- Error message on checkboxes
- Details
- Legend as page heading
- Does NOT test details with form elements inside

Confirm postcode:
- autocomplete with `postal-code`

Announcement:
- Header with lots of links
- Footer with lots of links
- Lead paragraph
- Unordered list
- Ordered list
- Heading caption - there are two different ways of implementing this but we use both in the Design System so that covers it


nickcolley:
Non blocking: would be good to use the error messages in the guidance https://design-system.service.gov.uk/components/error-message/#use-error-message-templates

nickcolley:
https://govuk-frontend-review-pr-1195.herokuapp.com/full-page-examples/what-is-your-nationality needs a conditional reveal

PR #1196, opened 2019-02-15
Add news and communications full page example

nickcolley:


app/app.js
```diff
@@ -170,6 +170,9 @@ module.exports = (options) => {
     })
   })
 
+  require('./views/full-page-examples/feedback-page')(app)
```
nickcolley: Had to move these to overwrite the general page handler

nickcolley:
@hannalaakso since it could make it confusing I think that's blocking, will fix, thanks for noticing.

PR #1198, opened 2019-02-18
Refactor full page example code

nickcolley:
The application main file was getting a bit big, likewise the tests for the app.

I've split out the new functionality into it's own file which makes it easier to delete and refactor more in the future.

- de-duplicates the utility that formats the errors for the views into a util file
- co-locate test files next to functions
- split banner tests into it's own file
- split full page example tests into it's own file


nickcolley:
@hannalaakso I think in the future we could better group the logic like we do with components, and then the top level would feel less busy.

This change makes it a lot easier to do that in the future so rather than making this a huge refactor I say we go with this and then make it better later on.

PR #1200, opened 2019-02-18
Add update account details page

nickcolley:
Somewhat blocked on https://github.com/alphagov/govuk-frontend/pull/1198

This example helps us test autocomplete and spellcheck.

### DIFF
--- a/app/full-page-examples.js
+++ b/app/full-page-examples.js
@@ -3,6 +3,7 @@ module.exports = (app) => {
   require('./views/full-page-examples/how-do-you-want-to-sign-in')(app)
   require('./views/full-page-examples/news-and-communications')(app)
   require('./views/full-page-examples/passport-details')(app)
+  require('./views/full-page-examples/update-your-account-details')(app)
   require('./views/full-page-examples/upload-your-photo')(app)
   require('./views/full-page-examples/what-is-your-nationality')(app)
   require('./views/full-page-examples/what-is-your-postcode')(app)
--- a/app/full-page-examples.test.js
+++ b/app/full-page-examples.test.js
@@ -15,6 +15,7 @@ const expectedPages = [
   'passport-details',
   'service-manual-topic',
   'start-page',
+  'update-your-account-details',
   'upload-your-photo',
   'what-is-your-nationality',
   'what-is-your-postcode'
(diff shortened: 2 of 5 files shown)


app/views/full-page-examples/update-account-details/index.njk
```diff
@@ -0,0 +1,73 @@
+{# This example is not based on a live example #}
+{% extends "_generic.njk" %}
+
+{% from "back-link/macro.njk" import govukBackLink %}
+{% from "error-summary/macro.njk" import govukErrorSummary %}
+{% from "input/macro.njk" import govukInput %}
+{% from "date-input/macro.njk" import govukDateInput %}
```
hannalaakso: I don't think we're using this

app/views/full-page-examples/update-account-details/index.njk
```diff
@@ -0,0 +1,73 @@
...
+				errorMessage: errors["email"],
+				autocomplete: "email",
+				attributes: {
+					spellcheck: false
+				}
+			}) }}
+
+			{{ govukInput({
+				label: {
+					text: "Password",
```
hannalaakso: This is non-blocking but I wonder if it it would be clearer to the use what's expected if this said "New password". Maybe that's implicit in the heading text, I don't know. Feel free to ignore.

nickcolley: Yeah fair that sounds good.

app/views/full-page-examples/update-account-details/index.njk
nickcolley: Ooops thanks for this.

app/views/full-page-examples/update-account-details/index.njk
```diff
@@ -0,0 +1,72 @@
...
+				errorMessage: errors["email"],
+				autocomplete: "email",
+				attributes: {
+					spellcheck: false
+				}
+			}) }}
+
+			{{ govukInput({
+				label: {
+					text: "Nw password",
```
hannalaakso: Typo

nickcolley: :man_facepalming:

PR #1202, opened 2019-02-18
Conditional reveal for nationality question template

hannalaakso:
This adds a conditional reveal to the nationality question example template as per the [GOV.UK page](https://www.registertovote.service.gov.uk/register-to-vote/nationality).

There could be multiple selected checkboxes. Store boolean on whether the searched value is contained in the array before passing it to `checked`.

### DIFF
--- a/app/views/full-page-examples/what-is-your-nationality/index.js
+++ b/app/views/full-page-examples/what-is-your-nationality/index.js
@@ -7,7 +7,20 @@ module.exports = (app) => {
     [
       body('confirm-nationality')
         .exists()
-        .not().isEmpty().withMessage('Select your nationality or nationalities')
+        .not().isEmpty().withMessage('Select your nationality or nationalities'),
+      body('country-name')
+        .custom((value, { req: request }) => {
+          // See https://github.com/express-validator/express-validator/pull/658
+          const confirmedNationality = request.body['confirm-nationality'] || []
+          // If the other country option is selected and there's no value.
+          if (
+            confirmedNationality.includes('other-country-nationality') &&
+              !value
+          ) {
+            throw new Error('Enter your country')
+          }
+          return true
+        })
     ],
     (request, response) => {
       const errors = formatValidationErrors(validationResult(request))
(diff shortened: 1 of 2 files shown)


app/views/full-page-examples/what-is-your-nationality/index.njk
```diff
@@ -47,17 +81,20 @@
...
               value: "british-nationality",
               text: "British (including English, Scottish, Welsh or from Northern Ireland)",
-              checked: values["confirm-nationality"] === "british-nationality"
+              checked: britishNationalityValue
             },
             {
               value: "irish-nationality",
               text: "Irish (including from Northern Ireland)",
-              checked: values["confirm-nationality"] === "irish-nationality"
+              checked: irishNationalityValue
```
nickcolley: This can be

```js
checked: (
    values["confirm-nationality"] and 
    values["confirm-nationality"].includes("british-nationality")
)
```

Same for the others.

app/views/full-page-examples/what-is-your-nationality/index.js
```diff
@@ -23,7 +23,22 @@ module.exports = (app) => {
     [
       body('confirm-nationality')
         .exists()
-        .not().isEmpty().withMessage('Select your nationality or nationalities')
+        .not().isEmpty().withMessage('Select your nationality or nationalities'),
+      body('country-name')
```
nickcolley: This could mirror the logic in `feedback-page`

```javascript
body('country-name')
    .custom((value, { req: request }) => {
        // See https://github.com/express-validator/express-validator/pull/658
        const otherCountryNationality = request.body['confirm-nationality']
        if (!otherCountryNationality.includes('other-country-nationality')) {
        return true
        }
        if (!value) {
        throw new Error('Enter your country')
        }
        return true
    })
```

nickcolley: This logic is not right one second...

nickcolley: Potentially this is slightly easier to read since it returns early rather than having a forking conditional.

Up to you though, I don't think this is blocking.

hannalaakso: Nice 👍 Have refactored accordingly.

hannalaakso:
Thanks for the review, comments addressed @nickcolley 👍

PR #1208, opened 2019-02-20
Update full page examples

hannalaakso:
Some updates to full page templates for the accessibility audit:
- Updated "What is your nationality page" to contain a form within details component for the accessibility audit.
- Added "What is the last country you visited?" example.

### DIFF
--- a/app/views/full-page-examples/what-is-your-nationality/index.njk
+++ b/app/views/full-page-examples/what-is-your-nationality/index.njk
@@ -5,6 +5,7 @@
 {% from "checkboxes/macro.njk" import govukCheckboxes %}
 {% from "input/macro.njk" import govukInput %}
 {% from "details/macro.njk" import govukDetails %}
+{% from "textarea/macro.njk" import govukTextarea %}
 {% from "button/macro.njk" import govukButton %}
 {% from "error-summary/macro.njk" import govukErrorSummary %}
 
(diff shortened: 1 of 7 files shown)


app/views/full-page-examples/what-was-the-last-country-you-visited/index.njk
```diff
@@ -0,0 +1,67 @@
...
+              value: "france",
+              text: "France"
+          },
+          {
+              value: "spain",
+              text: "Spain"
+          },
+          {
+              value: "italy",
+              text: "Italy"
```
nickcolley: Should we have an 'other' given that we don't have a complete list?

app/views/full-page-examples/what-is-your-nationality/index.njk
```diff
@@ -62,8 +63,11 @@
             {
               id: "confirm-nationality",
               value: "british-nationality",
-              text: "British (including English, Scottish, Welsh or from Northern Ireland)",
-              checked: values["confirm-nationality"]|length and "british-nationality" in values["confirm-nationality"]
+              text: "British",
+              checked: values["confirm-nationality"]|length and "british-nationality" in values["confirm-nationality"],
```
nickcolley: I think the second checkbox we can move the text in brackets into hint text too.

hannalaakso: 👍

app/views/full-page-examples/what-was-the-last-country-you-visited/index.njk
hannalaakso: 👍

Issue #1127, opened 2019-01-11
Investigate components being initialised more than once


joelanman:
We had a case of a user initialising the components twice by accident. This doesn't warn or error, but creates a broken state (the accordion for example opens then closes on click, and creates two 'Open all' buttons.

nickcolley:
We need to make sure that we allow for use cases where people inject new HTML into the DOM after the page has loaded and then running the initialise functions to enhance the new components.

Custom Elements seem to account for this, and the polyfills have code that could help us 

https://github.com/webcomponents/custom-elements/blob/7263e628f3d13b6a66e23c61147b299f31ff0964/src/Patch/HTMLElement.js#L48

nickcolley:
I've put together an example of how you could store instances on the classes for each component.

We'd likely also want to have consistent destroy methods too, which I've demonstrated.

https://jsbin.com/mapoqof/edit?js,console,output

nickcolley:
My approach I've shared already puts the instance information on the components themselves, but we could also consider an approach similar to web component custom elements.

Where we have a single registry (could be an array), that we update and check:

```
if (!window.customElements.get('x-component')) {
  window.customElements.define('x-component', ComponentElement)
}
```

nickcolley:
Seen this again from a Prototype Kit user.

Their tabs component was skipping two tabs at once since they accidentally included their `application.js` bundle twice in the page which meant the component was initialised twice.

vanitabarrett:
As discussed in the JavaScript effort/value session, this is happening because we don't have any guards against initialising a component more than once. This doesn't feel right, but it's also a bit of an unusual case which doesn't come up very often.

querkmachine:
Having this functionality may come in useful with our current work relating to localisation and programatic APIs. 

For example, a service developer may only want to pass configuration options to a specific instance of a component, but use the default options for all other components. In attempting to do so, they may write something like this:

```javascript
new window.GOVUKFrontend.Accordion($element, { config }).init()
window.GOVUKFrontend.initAll()
```

However, unless the developer takes care to remove the `data-module` attribute from the custom instance (which isn't possible if using our Nunjucks macros), this will initialise the same Accordion twice—resulting in multiple buttons and event handlers being created.

[...]

PR #1215, opened 2019-02-21
Remove error-summary dependence on document.onload

alex-ju:
The error-summary component is being initialised on an element using a `data-` attribute, which means that at the initialisation moment the element already exists is loaded in the DOM. Hence, we don't need to wait for `document.onload` to happen in order to set the focus on this element.

Background: We're using the error-summary component in a modal dialogue and we're planning to re-initialise the GOVUK Frontend components within the modal using `GOVUKFrontend.initAll()` (yeah, we need to scope it, but I'll raise a separate PR for that).
This component and `Tabs` are the only one relying on `document.onload` to happen.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,10 @@
 
   ([PR #N](https://github.com/alphagov/govuk-frontend/pull/N))
 
+- Remove error-summary dependence on document.onload
+
+  ([PR #1215](https://github.com/alphagov/govuk-frontend/pull/1215))
+
 
 ## 2.8.0 (Feature release)
 
--- a/src/components/error-summary/error-summary.js
+++ b/src/components/error-summary/error-summary.js
@@ -11,9 +11,7 @@ ErrorSummary.prototype.init = function () {
   if (!$module) {
     return
   }
-  window.addEventListener('load', function () {
-    $module.focus()
-  })
+  $module.focus()
 
   $module.addEventListener('click', this.handleClick.bind(this))
 }


nickcolley:
I'm not sure initialising GOV.UK Frontend twice is safe right now:

- https://github.com/alphagov/govuk-frontend/issues/1127

We also discussed a way to override this behaviour of focusing here:

https://github.com/alphagov/govuk-frontend/pull/1173

nickcolley:
@alex-ju what do you think about having a way to set this behaviour?

```js
new ErrorSummary($errorSummary, { autofocus: false }).init()
```

or

```html
<div data-module="error-summary" data-autofocus="false"></div>
```

then in your application code

```
// main.js
const errorSummaryInstance = new ErrorSummary($errorSummary, { autofocus: false }).init()

window.onload = () => {
  errorSummaryInstance.focus()
}

const modalErrorSummaryInstance = new ErrorSummary($errorSummary, { autofocus: false }).init()

modal.open = () => {
  modalErrorSummaryInstance.focus()
}
```

nickcolley:
If you're under time pressure there's also this hack you could consider: https://github.com/alphagov/govuk-design-system/blob/master/src/javascripts/govuk-frontend.js#L26

hannalaakso:
@alex-ju We're planning to pick up https://github.com/alphagov/govuk-frontend/issues/1223 to look into a potential solution. The outcome of the the related work would then help to fix this issue.

alex-ju:
I think I wasn't very clear in my PR message. But with this change I'm not aiming for custom initialisation of the error-message component, but I'm rather questioning the redundancy of [adding an onload event listener](https://github.com/alphagov/govuk-frontend/pull/1215/files#diff-8a8e15f0fe8b5d9236bef83a7c1642ceL14) on a component that's being initialised based on a `data-` attribute, hence after the document is loaded. #1238 might solve this, but it was more of a sanity check from me that the `addEventListener` is not required.

nickcolley:
If we removed this, the component would always steal the users' focus when added to the page which may be the correct behaviour...

It feels like fixing the underlying issues with initialising components will make these components more flexible in the long term.

I'm not a fan of over engineering so if you think this work: https://github.com/alphagov/govuk-frontend/issues/1223 is unnecessary let me know and we can park it.

hannalaakso:
I think that removing the event listener as @alex-ju proposes might actually make sense. I guess as @nickcolley said the question is whether anyone would add an error summary dynamically to a page and not want it to have focus. Can we think of any scenarios where that might be necessary? The behaviour when the error summary is present on page load won't change. I think being able to specify options programmatically would be a good change too but I don't think it's necessary for solving this problem (despite me proposing it earlier in this thread). 

I guess the question is also whether accepting this PR would be a breaking change. My hunch is "no", as relying on the current behaviour when adding an error summary to a page dynamically seems like such an odd edge case.

nickcolley:
OK chatted with @hannalaakso and @36degrees again on this one.

We think that your proposal is not a breaking change based on how GOV.UK Frontend needs to be used.

"JavaScript in GOV.UK Frontend requires HTML to be parsed first by the browser before it is initialised."

I have tested this locally and the error summary is still focused on page load as expected.

This means that the proposal I was suggesting is not a priority so we will close it and consider it again in the future if we get a clearer need for it.

nickcolley:
OK chatted with @hannalaakso and @36degrees again on this one.

We think that your proposal is not a breaking change based on how GOV.UK Frontend needs to be used.

"JavaScript in GOV.UK Frontend requires HTML to be parsed first by the browser before it is initialised."

This means that the proposal I was suggesting is not a priority so we will close it and consider it again in the future if we get a clearer need for it.

nickcolley:
I will add the CHANGELOG :)

alex-ju:
I was away for a few days. Yep I get your point @nickcolley - it's related to the way we do #1223, but I was wondering if they depend one on the other. Thanks for updating the branch!

PR #1216, opened 2019-02-21
Allow initAll to be scoped to a specific part of a page

alex-ju:
We currently initialise component at the document level, but there are cases when components are being injected after the document is loaded. I'm proposing to allow `initAll` function to be scoped, so it can be called for a specific container or part of the page.

**Note**
We can replace _scope_ with any name that makes more sense.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,12 @@
 
   ([PR #1233](https://github.com/alphagov/govuk-frontend/pull/1233))
 
+- Allow initAll to be scoped to a specific part of a page
+
+  See ["Initialise GOV.UK Frontend in only certain sections of a page"](docs/installation/installing-with-npm.md#initialise-govuk-frontend-in-only-certain-sections-of-a-page) for more information.
+
+  ([PR #1216](https://github.com/alphagov/govuk-frontend/pull/1216))
+
 🔧 Fixes:
 
 - Pull Request Title goes here
@@ -157,7 +163,7 @@
   Also thanks to [Malcolm Butler](https://github.com/MoJ-Longbeard) for exploring a [previous version of this fix](https://github.com/alphagov/govuk-frontend/pull/1185).
 
   ([PR #1220](https://github.com/alphagov/govuk-frontend/pull/1220))
-  
+
 
 ## 2.7.0 (Feature release)
 
--- a/docs/installation/installing-with-npm.md
+++ b/docs/installation/installing-with-npm.md
@@ -154,6 +154,21 @@ Including the script elsewhere will stop components from functioning or displayi
 </html>
 ```
 
+#### Initialise GOV.UK Frontend in only certain sections of a page
+
+By default, the `initAll` function from GOV.UK Frontend initialises all components scoped to an entire page with the `document` object.
+
+You can change this by passing the `scope` parameter to the `initAll` function.
+
+For example, if you have a modal dialog box that opens with new markup you could do the following:
+
+```js
+var $modal = document.querySelector('.modal')
+window.GOVUKFrontend.initAll({
+  scope: $modal
+})
+```
+
 #### Initialise individual included components
 
 GOV.UK Frontend components with JavaScript behaviour have the `data-module` attribute set in their markup.
--- a/src/all.test.js
+++ b/src/all.test.js
@@ -90,6 +90,28 @@ describe('GOV.UK Frontend', () => {
         }, component)
       })
     })
+    it('can be initialised scoped to certain sections of the page', async () => {
+      await page.goto(baseUrl + '/examples/scoped-initialisation', { waitUntil: 'load' })
+
+      // To test that certain parts of the page are scoped we have two similar components
+      // that we can interact with to check if they're interactive.
+
+      // Check that the conditional reveal component has a conditional section that would open if enhanced.
+      await page.waitForSelector('#conditional-not-scoped-1', { hidden: true })
+
+      await page.click('[for="not-scoped-1"]')
+
+      // Check that when it is clicked that nothing opens, which shows that it has not been enhanced.
+      await page.waitForSelector('#conditional-not-scoped-1', { hidden: true })
+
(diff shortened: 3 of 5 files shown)


nickcolley:
One thing to think about is we have had users report errors when running `initAll` twice, and we thought about solving that a different way by making it safe to call it multiple times:

https://github.com/alphagov/govuk-frontend/issues/1127

Doesn't necessarily block this, but I think if we solved that other issue it'd be fine to run initAll whenever new markup is injected into the page.

It's also worth noting that custom elements (web components) avoid this issue, and are enhanced whenever the browser detects the custom element.

hannalaakso:
I've chatted with @alex-ju about this. It seems that there might be value in building in both approaches:
1. Do a check to see if the component is already initialised (what @nickcolley drafted)
2. Initialising against a certain scope

The main value with 2. is initialising based on a scope with DOM elements that are added after `document.load` to avoid querying the whole DOM.

nickcolley:
Agree we can do both however...

>  avoid querying the whole DOM

Since the user will be running 'initAll' themselves I don't think there's much overhead, I'd understand if we were listening to DOM changes with mutation observers maybe...

nickcolley:
I've added done the following

1. Made the JavaScript work without ES2015 features
2. Add a test page that demonstrates the scoping and corresponding puppeteer tests.
3. Added some documentation to the installation instructions to explain how to use this feature.

alex-ju:
Thanks for taking the time to properly update this PR @nickcolley. This is very much aligned with the need for a scoped initialisation.

alex-ju:
@nickcolley changelog entry added. thanks a lot for the test and doc!

CHANGELOG.md
```diff
@@ -14,11 +14,22 @@
 
 🆕 New features:
 
-- Pull Request Title goes here
+- Enable GOV.UK Frontend components to be scoped the a specific container
```
nickcolley: @alex-ju would you mind keeping the template around instead of overwriting it, this is so future people can copy and paste it easier.

alex-ju: sure, added it back

nickcolley:
OK I think this is ready for review

docs/installation/installing-with-npm.md
```diff
@@ -154,6 +154,21 @@ Including the script elsewhere will stop components from functioning or displayi
 </html>
 ```
 
+#### Initialise GOV.UK Frontend in only certain sections of a page
```
36degrees: ❤️ 📖

src/all.test.js
```diff
@@ -90,6 +90,28 @@ describe('GOV.UK Frontend', () => {
...
       })
     })
+    it('can be initialised scoped to certain sections of the page', async () => {
+      await page.goto(baseUrl + '/examples/scoped-initialisation', { waitUntil: 'load' })
+
+      // To test that certain parts of the page are scoped we have two similar components
+      // that we can interact with to check if they're interactive.
+
+      // Check that the conditional reveal component has a conditional section that would open if enhanced.
+      await page.waitForSelector('#conditional-not-scoped-1', { visible: false })
```
36degrees: We use a combination of `visible: false`, `visible: true` and `hidden: true` in this test – is there a difference between them, or are they just opposites of each other?

It'd be good to be consistent with either the key (`visible: true`, `visible: false`) or the value (`visible: true`,  `hidden: true`) if so.

src/all.test.js
```diff
@@ -90,6 +90,28 @@ describe('GOV.UK Frontend', () => {
...
+
+      await page.click('[for="not-scoped-1"]')
+
+      // Check that when it is clicked that nothing opens, which shows that it has not been enhanced.
+      await page.waitForSelector('#conditional-not-scoped-1', { hidden: true })
```
36degrees: Does this definitely test what we want? It's 'waiting' for something was hidden to still be hidden (which it already was…), rather that asserting that it never becomes visible.

src/all.test.js
```diff
@@ -90,6 +90,28 @@ describe('GOV.UK Frontend', () => {
         }, component)
       })
     })
+    it('can be initialised scoped to certain sections of the page', async () => {
```
36degrees: This makes sense, but only tests the instantiation of a single component and feels like it could be relatively fragile if we e.g. changed the way that conditional reveals work.

Wondering if there's a better way to test this, maybe as a unit test rather than as an integration test, e.g. passing a stubbed `DomElement` as the scope and asserting that `querySelectorAll` is called, and/or stubbing `document` and asserting that it doesn't get called?

(Just worth thinking about, this isn't a blocking comment 👍)

docs/installation/installing-with-npm.md
nickcolley: @soupdragon99 has helped on this one

src/all.test.js
nickcolley: There's some nuance to it, you can have something in the DOM but not visible, or something not in the DOM and therefore not visible I believe.

I'll try and clarify this.

src/all.test.js
nickcolley: It was indeed testing it when I remove the scope functionality and ran these tests.

src/all.test.js
nickcolley: We are somewhat bound by the limitations of our testing setup.

I'll see if I can stub an element like you've suggested...

36degrees: Totally fair, don't spend too much time on it 👍

nickcolley: Yeah I think we'd have to run the test using JSDOM rather than in a browser, and it's not clear that you can spy on an element so I'll try to make the current setup clearer but not sure it's worth the time spent...

nickcolley: Let me know if you think the current approach is too risky

nickcolley:
@alex-ju thanks for your patience with these pull requests, sorry I did not fully understand your issue the first time.

Let us know when you need us to release these and we'll get it out for you. :+1:

alex-ju:
Thanks a lot for pairing on it @nickcolley! Would be great to have it in the next planned release.

PR #1217, opened 2019-02-21
Update full page examples based on automated testing

nickcolley:


app/views/full-page-examples/bank-holidays/index.njk
```diff
@@ -36,16 +36,18 @@
...
+          <div class='govuk-!-font-size-36 govuk-!-margin-bottom-4'>
+            The next bank holiday in England and Wales is
+          </div>
+          19 April
         {% endset %}
 
         {% set englandAndWalesPanelHtml %}
-          <span class='govuk-!-font-size-24 govuk-!-font-weight-bold'>Good Friday</span>
-        19 April
+          Good Friday
```
hannalaakso: As this content is no longer HTML, we could pass it directly as `text` to the component. Same for the others.

nickcolley: Good spot

Issue #1211, opened 2019-02-21
Revert fix for wrapping of long lines of text


edwardhorsford:
In #1169 a fix was added to force long lines of text to wrap - aiming for URLs.

@@leekowalkowski-hmrc [points out](https://github.com/alphagov/govuk-design-system-backlog/issues/61#issuecomment-465992040) that this forces *all* words to wrap - significantly reducing legibility.

Example:
![screenshot 2019-02-21 at 13 15 29](https://user-images.githubusercontent.com/2204224/53171582-1781c000-35db-11e9-8660-41ee284bac39.png)

I propose we revert #1169 until a better fix can be found. 

I think we've likely made a more common case worse (paragraphs of text) to improve a less common case (long urls).

I also wonder if we have test pages for some of these patterns that would have let us catch this - ideally a page with the extremes of the sort of data that might be used.

edwardhorsford:
It sounds like #1185 may partially fix this - but if I'm reading correctly, some browsers will still get broken words.

nickcolley:
This will be resolved when we release the new version with the fixes in #1185

edwardhorsford:
@nickcolley I'm wary of 'most browsers' - what browsers does it not fix? Is it worth not attempting to wrap those browsers? I'd argue wrapping paragraphs of text mid-word is probably worse than not wrapping urls.

nickcolley:
Looks like it'd be good to investigate  `overflow-wrap: break-word` but it's worth noting that it has no support in older browsers: https://caniuse.com/#feat=wordwrap

So we'd need to address that in the same way we have considered this in the fix which is currently merged.

leekowalkowski-hmrc:
We need `word-wrap: break-word`, too:

```
word-wrap: break-word;
overflow-wrap: break-word;
```

The property was renamed when it was adopted into the standards.

nickcolley:
We should also consider improving the extreme examples we have to include content that it's clearer the issues, since lorem ipsum always looks hard to read ;)

http://govuk-frontend-review.herokuapp.com/components/summary-list/extreme/preview

nickcolley:
I will be looking into this soon, hopefully release a fix for this next week, thanks again for your patience.

nickcolley:
@edwardhorsford @leekowalkowski-hmrc it would be good to get your feedback on this improvement I've proposed based on what you have been suggesting: https://github.com/alphagov/govuk-frontend/pull/1220

[...]

PR #1220, opened 2019-02-22
Improve word wrapping in summary list component

nickcolley:
Opening a draft of what I've explored so far, this seems like a problem where we have to make some compromises.

The summary list currently wraps text in all browsers, but only in blink/webkit does it wrap entire words, all other browsers it can wrap mid word which can make it more difficult to read.

To make this better we want to try and use `overflow-wrap` which works cross browsers, however...

In order to use `overflow-wrap` with table CSS we need to use a fixed layout since the summary list uses table CSS.

This means we can't rely on a table's automatic resizing, so we have to compromise
with a larger actions column that we have before, so that we can have at least two actions on one line without the actions wrapping.

Welcome any input to an alternative that I may have missed.

I have also considered that this might be best solved with CSS Grid but have discounted that for now based on the support for browsers.

You can take a look at the two by comparing:

## Before

- [Summary list review page](https://govuk-frontend-review.herokuapp.com/components/summary-list)
- [Check your answers review page](https://govuk-frontend-review.herokuapp.com/full-page-examples/check-your-answers)

### Screenshots

Without changes there's no gap

## After

- [Summary list review page](https://govuk-frontend-review-pr-1220.herokuapp.com/components/summary-list)
- [Check your answers review page](https://govuk-frontend-review-pr-1220.herokuapp.com/full-page-examples/check-your-answers)

## Pros

- Works more consistently across all browsers we support

## Cons

- Content cannot decide the size of columns anymore leading to smaller columns than before

### Screenshots

With changes, there's a gap between columns

## Related links

- [Handling Long Words and URLs (Forcing Breaks, Hyphenation, Ellipsis, etc)](https://css-tricks.com/snippets/css/prevent-long-urls-from-breaking-out-of-container/)
- [Overflow wrap in tables](https://stackoverflow.com/questions/1258416/word-wrap-in-an-html-table)
- [word-break support](https://caniuse.com/#feat=word-break) only Blink/Webkit browser support a word by word wrapping
- [overflow-wrap support](https://caniuse.com/#feat=wordwrap)

Aims to resolve https://github.com/alphagov/govuk-frontend/issues/1211

### DIFF
--- a/app/views/full-page-examples/check-your-answers/index.njk
+++ b/app/views/full-page-examples/check-your-answers/index.njk
@@ -177,6 +177,39 @@
               }
             ]
           }
+        },
+        {
+          key: {
+            text: "Additional details"
+          },
+          value: {
+            html: '
+              <p class="govuk-body">I have contacted my council for a Temporary Event Notice (<abbr title="Temporary Event Notice">TEN</abbr>) as I want to carry out a ‘licensable activity’ on unlicensed premises in England or Wales.</p>
+
+              <p class="govuk-body">Licensable activity includes:</p>
+
+              <ul class="govuk-list govuk-list--bullet">
+                <li>selling alcohol</li>
+                <li>serving alcohol to members of a private club</li>
+                <li>providing entertainment, such as music, dancing or indoor sporting events</li>
+                <li>serving hot food or drink between 11pm and 5am</li>
+              </ul>
+
+              <p class="govuk-body">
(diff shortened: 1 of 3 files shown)


nickcolley:
I got an approval from @dashouse over Slack for this so I'm going to push forwards and get it merged thanks again for the great review @36degrees

nickcolley:
Need to add a CHANGELOG entry for this, will follow up in a second.

PR #1221, opened 2019-02-25
Add hidden "Error:" prefix to error messages

36degrees:
We've standardised on using a simpler, more 'instructional' language for our messages, but this means that most error messages are less explicit about the fact that the field is in an error state.

I think we are therefore relying slightly more on the visual presentation of the error message to help the user understand that they have 'done something wrong'.

Thinking specifically about the way that a screen-reader user would encounter an error message:

> National Insurance number, text field. It’s on your National Insurance card, benefit letter, payslip or P60. For example, ‘QQ 12 34 56 C’. Enter a National Insurance number in the right format. You are currently inside a text field. To enter text in this field, type.

I don't think it's entirely clear that "Enter a National Insurance number in the right format" is an error.

This attempts to counter that by adding a visually hidden "Error:" prefix to the error message:

> National Insurance number, text field. It’s on your National Insurance card, benefit letter, payslip or P60. For example, ‘QQ 12 34 56 C’. Error: Enter a National Insurance number in the right format. You are currently inside a text field. To enter text in this field, type.

(As discussed in https://github.com/alphagov/govuk-design-system-backlog/issues/47#issuecomment-401793629)

### DIFF
--- a/src/components/error-message/template.test.js
+++ b/src/components/error-message/template.test.js
@@ -29,7 +29,7 @@ describe('Error message', () => {
     })
 
     const content = $('.govuk-error-message').html().trim()
-    expect(content).toEqual('Unexpected &gt; in body')
+    expect(content).toContain('Unexpected &gt; in body')
   })
 
   it('allows summary HTML to be passed un-escaped', () => {
@@ -38,7 +38,7 @@ describe('Error message', () => {
     })
 
     const content = $('.govuk-error-message').html().trim()
-    expect(content).toEqual('Unexpected <b>bold text</b> in body copy')
+    expect(content).toContain('Unexpected <b>bold text</b> in body copy')
   })
 
   it('allows additional attributes to be specified', () => {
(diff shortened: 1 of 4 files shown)


nickcolley:
This seems like a really nice improvement to make it super explicit that there's an error.

Does it need to be done at the same time as a new feature? Would you ever want to remove this?

In theory we could add this feature without extending the API if we were confident.

If not, this seems like a good way to do it :+1:

36degrees:
I figure you need to be able to customise the text for localisation if nothing else.

You might also want to disable it in some cases, if the error message you're using is more explicit.

src/components/error-message/template.test.js
```diff
@@ -53,4 +53,33 @@ describe('Error message', () => {
     expect($component.attr('data-test')).toEqual('attribute')
     expect($component.attr('id')).toEqual('my-error-message')
   })
+
+  it('includes a visually hidden "Error" prefix by default', () => {
```
nickcolley: These tests are very good :+1:

36degrees:
I think based on the work with the error summary and the conclusions reached in https://accessibility.blog.gov.uk/2017/12/18/what-working-on-gov-uk-navigation-taught-us-about-accessibility/ we can be pretty confident this will work as intended.

Would you agree?

PR #1226, opened 2019-02-28
Add required imports to form-group

penx:
When doing:

```scss
@import "~govuk-frontend/objects/form-group";
```

I get:

```
ERROR in ./src/objects/form-group/_form-group.module.scss (./node_modules/css-loader??ref--6-1!./node_modules/sass-loader/lib/loader.js!./src/objects/form-group/_form-group.module.scss)
Module build failed (from ./node_modules/sass-loader/lib/loader.js):

@include govuk-exports("govuk/objects/form-group") {
        ^
      No mixin named govuk-exports
      in /node_modules/govuk-frontend/objects/_form-group.scss (line 1, column 10)
```

Adding

```scss
@import "~govuk-frontend/settings/all";
@import "~govuk-frontend/helpers/all";
```

Makes it compile ok, and adding

```scss
@import "~govuk-frontend/tools/all";
```

Seems to be needed to get the correct CSS output.

In my opinion this would be better addressed in govuk-frontend direct.

### DIFF
--- a/src/objects/_form-group.scss
+++ b/src/objects/_form-group.scss
@@ -1,3 +1,7 @@
+@import "../settings/all";
+@import "../tools/all";
+@import "../helpers/all";
+
 @include govuk-exports("govuk/objects/form-group") {
 
   .govuk-form-group {


penx:
Related https://github.com/penx/govuk-frontend-react/pull/12/files#diff-55612d7a8f5985f4d26efb6d88e245ca

penx:
This also needs to be reviewed on other Scss files in objects, e.g. width-container.

nickcolley:
Hey @penx !

As far as I can tell, it does look like these files should be declaring their dependencies like you've described.

I've had a quick look and this looks to be the same case with most of the files, I'll ping the team to see if this was intentional.

36degrees:
@penx are you happy that this would be addressed by #1235?

penx:
Yep, great work thanks!

36degrees:
Closing in favour of #1235 – thanks again, @penx 👍

Issue #1204, opened 2019-02-18
Content on pages move horizontally depending on if a scrollbar is on screen or not


edwardhorsford:
Content in the template is centred on the page. If the page has a scrollbar, this will be slightly to the left. If the page doesn't have a scrollbar, the content is slightly to the right.

If a user moves between tabs / sections of the site, this makes the GOV.UK logo and other things appear to jump about. Ideally when using navigation links, everything would stay in position.

Example:
![govuk-template-scrollbar-or-not](https://user-images.githubusercontent.com/2204224/52967998-d5bb0480-33a3-11e9-8741-ee1f416cf4a6.gif)

On my personal site I've gotten around this by using `margin-left: calc(100vw - 100%);` on `body` - which adds a left margin equal to the width of a scrollbar. 

@nickcolley tells me that @colinrotherham also raised this recently and suggested this:
```
@media only screen {
  .container {
    overflow-y: scroll;
  }
}
```

nickcolley:
In GOV.UK Template (which we aim to replace) it has something to account for this:

- https://github.com/alphagov/govuk_template/blob/master/source/assets/stylesheets/_basic.scss#L60
- https://github.com/alphagov/govuk_template/blob/master/source/assets/stylesheets/_basic.scss#L111

36degrees:
Fixed in #1230

PR #1230, opened 2019-03-01
Prevent horizontal jump as scrollbars appear

colinrotherham:
As discussed previously with @nickcolley, we're seeing horizontal "jumps" in position as scroll bars appear and disappear in IE11—as page content grows.

Also spotted by @edwardhorsford here: https://github.com/alphagov/govuk-frontend/issues/1204

We've fixed the issue manually (screen-only, not affecting print) by using:

```css
@media only screen {
  .container {
    overflow-y: scroll;
  }
}
```

Also found in GOV.UK Template:
https://github.com/alphagov/govuk_template/blob/f2c10fae7ec7e371fd11c409cd71a399e3342e7f/source/assets/stylesheets/_basic.scss#L111

And inside GOV.UK Elements:
https://github.com/alphagov/govuk_elements/blob/84f485f688e7f796a4c44dd73b17177e8e39538c/assets/sass/elements/_govuk-template-base.scss#L50

Things to discuss:

**1. Do we need to restrict to `@media only screen`**
Does this prevent printable content from being hidden?

~~**2. Do we need the old `-ms-overflow-style: scrollbar` style**~~
~~To force the scrollbar to always display in IE10/11~~

Confirmed: Point 2) was previously required for "tablet mode" Internet Explorer that first came with Windows 8, but since scroll bars auto-hide by default it's not needed anymore.

### Browser testing

- [x] Internet Explorer 8 (Windows)
- [x] Internet Explorer 9 (Windows)
- [x] Internet Explorer 10 (Windows)
- [x] Internet Explorer 11 (Windows)
- [x] Edge (latest 2 versions) (Windows)
- [x] Google Chrome (latest 2 versions) (Windows)
- [x] Mozilla Firefox (latest 2 versions) (Windows)
- [x] Safari 9+ (macOS)
- [x] Google Chrome (latest 2 versions) (macOS)
- [x] Mozilla Firefox (latest 2 versions) (macOS)
- [x] Safari 9.3+ (iOS)
- [x] Google Chrome (latest version) (iOS)
- [x] Google Chrome (latest version) (Android)
- [x] Samsung Internet (latest version) (Android)

Specific screenshots:

### Edge 17 with fix
edge 17 with fix

### Edge 17 without fix
edge 17 without fix

### IE11 with fix
ie11 with fix

### IE11 without fix
ie11 without fix

### IE10 with fix
ie10 with fix

### IE10 without fix
ie10 without fix

### IE9 with fix
ie9 with fix

### IE9 without fix
ie9 without fix

### DIFF
--- a/src/core/_template.scss
+++ b/src/core/_template.scss
@@ -11,6 +11,12 @@
     // Prevent automatic text sizing, as we already cater for small devices and
     // would like the browser to stay on 100% text zoom by default.
     text-size-adjust: 100%;
+
+    // Force the scrollbar to always display in IE, to prevent horizontal page
+    // jumps as content height changes (e.g. autocomplete results open).
+    @include govuk-media-query($media-type: screen) {
+      overflow-y: scroll;
+    }
   }
 
   // Applied to the <body> element
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,16 @@
 
   ([PR #N](https://github.com/alphagov/govuk-frontend/pull/N))
 
+- Prevent horizontal jump as scrollbars appear
+
+  As content vertical height grows (e.g. autocomplete results appear), browsers
+  may add scroll bars causing the page to jump horizontally in position.
+
+  To avoid this, re-introduce fix from GOV.UK Template:
+  https://github.com/alphagov/govuk-frontend/issues/1204
+
+  ([PR #1230](https://github.com/alphagov/govuk-frontend/pull/1230))
+
 - Accommodate camera notches on new devices (iPhone X, Google Pixel 3 etc)
 
   On newer devices with "camera notches", browsers reserve a safe area in landscape orientation (known as pillarboxing) so content isn't obscured.


edwardhorsford:
Woop 🎉

colinrotherham:
Further testing on Windows 8 (using the non-desktop "tablet mode" Internet Explorer) shows that scroll bars will hide by default without:

```css
// Force IE10-11 scroll bar in "tablet" mode
-ms-overflow-style: scrollbar;
```

Added it back in.

src/core/_template.scss
```diff
@@ -11,6 +11,15 @@
     // Prevent automatic text sizing, as we already cater for small devices and
     // would like the browser to stay on 100% text zoom by default.
     text-size-adjust: 100%;
+
+    // Force the scrollbar to always display, to prevent horizontal page jumps
+    // as content height changes (e.g. autocomplete results open). Screen only.
+    @include govuk-media-query($media-type: screen) {
+      overflow-y: scroll;
+
+      // Force IE10-11 scroll bar in "tablet" mode
```
nickcolley: Your comments in the thread are really useful to understand why this is in here

"That said, there is no visual jump here as tablet mode IE draws the scroll bars on top of the content before fading away automatically"

I think it'd be good to add in the code too.

colinrotherham: But the more I read that, the more I think we shouldn't force scroll bars to be visible on "touch first" devices anymore.

Maybe at the launch of Windows 8 and the Microsoft Surface we were wrong to force them? We don't force them on Mac or iOS, or any other browser with auto-hiding scroll bars.

What do you think?

nickcolley: As far as I can tell setting the overflow to scroll forces scrollbars to always be visible? Even on OSX?

colinrotherham: Not that I've seen.

I can only get scrollbars to appear on macOS (Chrome and Safari) by plugging in a scroll-wheel mouse on a device without a trackpad or magic mouse or by changing the "Show scroll bars" setting in System Preferences.

scroll bars

nickcolley: OK I mistakenly though when you used that CSS they were always visible.

edwardhorsford: I get the sense it would be good to test this in browserstack against quite a few different devices. Do we need some test pages?

colinrotherham: Ah sorry, no, it's only Internet Explorer that seems to show a faded out disabled scroll bar once `overflow-y: scroll` is set. On macOS the default is also to auto-height, even with this set.

I'm going to remove this bit:

```scss
// Force IE10-11 scroll bar in "tablet" mode
-ms-overflow-style: scrollbar;
```

I'll add some more notes and a video below.

colinrotherham:
I've done some more testing today of "tablet mode" on Windows 10.

In Microsoft Edge (and the touch-optimised Internet Explorer on Windows 8/8.1) scroll bars are designed to auto-hide without taking up any horizontal space. Just like iOS and macOS Safari.

### Video demo showing auto-hiding scrollbars

[YouTube example](https://www.youtube.com/watch?v=PUWj6KHeQLU)

I'd prefer to adhere to the device's defaults and retain this behaviour.

For example, on touch-first hardware running iOS, Android, Windows 10 in S mode or "tablet mode" we should prefer auto-hiding scroll bars (same with macOS with a trackpad or magic mouse).

So I've removed:

```scss
// Force IE10-11 scroll bar in "tablet" mode
-ms-overflow-style: scrollbar;
```

nickcolley:
Edge 18 with fix

image

Edge 18 without fix

image
image

src/core/_template.scss
colinrotherham: Marking resolved as this line has been removed.

colinrotherham:
Additional browser testing completed above https://github.com/alphagov/govuk-frontend/pull/1230#issue-257529169

colinrotherham:
Since Firefox and Chrome on macOS will also obey the "Always" option when toggling "Show scroll bars", they also get the "visible but disabled" scroll bars—preventing the jump.

enable scrollbars

So we've prevented the visual jump in more than just IE:

#### Firefox 65 (with fix)
With scroll bars enabled, horizontal space is reserved

[YouTube example](https://www.youtube.com/watch?v=od9iKdrT73U)

#### Firefox 65 (without fix)
With scroll bars enabled, horizontal space is *not* reserved

[YouTube example](https://www.youtube.com/watch?v=WogaFg9Iy3U)

Browsers that auto-hide scroll bars are unaffected 👍

colinrotherham:
Thanks @nickcolley @dashouse @36degrees 😊

alex-ju:
Quick question: does anyone know if it's possible to achieve the same effect by moving this logic to the `<body>` instead of the `<html>` element? Could try it myself, but I'm curious to know if there's a reason for doing it on the special `<html>` element.

colinrotherham:
@alex-ju I think the `<body>` tag would work just as well.

By using `<html>` we copied what GOV.UK and the old **alphagov/govuk_elements** kit did which we know is tried and testing in various services already.

alex-ju:
@colinrotherham ok, thanks for clarifying this.

PR #1234, opened 2019-03-05
Avoid double-escaping example names

colinrotherham:
Just a minor one.

I noticed the example names are already HTML escaped before they hit the macro.

Due to Nunjucks `autoescape: true` they were being double-escaped.

#### Before
before

#### After
after


colinrotherham:
Pushed again, I fixed a typo `aria-describeby` which was in a test as well

PR #1235, opened 2019-03-05
Test that parts of GOV.UK Frontend can be imported individually

36degrees:
Ensure that files within certain layers render to CSS without error when imported individually:

- Core
- Objects
- Components
- Overrides

This will test that e.g. files within these layers import their own dependencies.

This builds on the work done by @penx in #1226.

Closes #1226

### DIFF
--- a/src/overrides/overrides.test.js
+++ b/src/overrides/overrides.test.js
@@ -1,15 +1,11 @@
 /* eslint-env jest */
 
-const util = require('util')
 const glob = require('glob')
-
+const { renderSass } = require('../../lib/jest-helpers')
 const configPaths = require('../../config/paths.json')
-
-const sass = require('node-sass')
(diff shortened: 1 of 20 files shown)


src/overrides/overrides.test.js
```diff
@@ -0,0 +1,11 @@
+/* eslint-env jest */
+
+const glob = require('glob')
+const { renderSass } = require('../../lib/jest-helpers')
+const configPaths = require('../../config/paths.json')
+
+const sassFiles = glob.sync(`${configPaths.src}/overrides/**/*.scss`)
+
+it.each(sassFiles)('%s renders to CSS without errors', (file) => {
+  return renderSass({ file: file })
```
nickcolley: Non-blocking you can use `func({ file })` here

penx:
Yay thanks!

PR #1244, opened 2019-03-08
Prevent Travis CI running when only docs change

nickcolley:
We trigger a lot of builds when we update CHANGELOGs.

We can detect if only markdown files have changed and avoid running Travis CI.

Based on https://reflectoring.io/skip-ci-build/

If we decide to do this we'd want to merge it and do a few quick test pull request branches to double check it still runs the tests as we expect...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,9 @@ jobs:
   include:
     - stage: Run npm test
       script:
+      # We can skip Travis if there are only changes to files like the CHANGELOG
+      - ./bin/check-skip-ci.sh
       # Ensure Travis aborts without running tests if the package-lock.json file needs updating
-      - set -e
       - ./bin/check-package-lock.sh
       # TravisCI is slower than our local machines, so results in intermittent timeouts
       # using the `--runInBand` flag we can force it to without requiring as much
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -2,6 +2,9 @@ language: node_js
 notifications:
   email: false
 sudo: false
+before_install:
+# We don't need to run Travis on files like the CHANGELOG
+- ./bin/check-skip-ci.sh
```
36degrees: I don't believe this will actually do anything here – this'll just exit like any other pre-install step (e.g. installing a dependency) and Travis will continue on to run the tests as usual.

I think this'll need to be part of the 'Run npm test' 'stage' – but it'll still need to exit cleanly to prevent Travis from marking the build as failed.

Happy to pair on this next week if that helps.

36degrees: Actually, it's more complicated than that, because I think `TRAVIS_COMMIT_RANGE` will always reflect the diff of the entire PR, not the diff since the last test was run. So this would help **PRs** that only change markdown files, but it won't prevent tests running for individual commits that only touch markdown files.

We also cancel pending builds when we push new commits to an existing branch, so I think it's probably safer just to always run the tests…

Issue #1236, opened 2019-03-05
Firefox visual bug with back button


BNewing:
This is a tiny issue, but on Firefox, the focus state of the back button has a little visual bug. It has a small white line at the top, which breaks the background highlighting:

<img width="96" alt="image" src="https://user-images.githubusercontent.com/16868713/53803120-a4157200-3f3b-11e9-8860-dda18bd741b4.png">

I found this on a project I was working on and the workaround I implemented was to add a border to the element and make it the same colour as the focus state when focused, and transparent when  not focused.

dashouse:
@BNewing Thanks Beverley, we're in the process of updating the focus style more generally for WCAG 2.1 compliance and the current proposal also fixes this issue.

We'll keep this open so we can refer to it when that work is complete and make sure that is addressed.

36degrees:
Just to confirm that this has been fixed as part of the focus state changes, and will be going out in GOV.UK Frontend 3.0.

<img width="124" alt="Screen Shot 2019-05-20 at 15 20 34" src="https://user-images.githubusercontent.com/121939/58028546-df731600-7b12-11e9-8a64-17aaada13395.png">

PR #1245, opened 2019-03-12
Focus iteration

dashouse:
WIP

### DIFF
--- a/src/components/back-link/_back-link.scss
+++ b/src/components/back-link/_back-link.scss
@@ -27,16 +27,7 @@
 
 
     &:focus {
-      // Current margin minus 3px padding added by focus
-      margin-top: 12px;
-      // 14px for icon + 5px from focus style
-      padding-left: 19px;
       border-bottom: 1px solid govuk-colour("yellow");
-
-      &:before {
-        top: 2px;
-        left: 5px;
-      }
     }
 
     // Prepend left pointing arrow
--- a/src/components/back-link/_back-link.scss
+++ b/src/components/back-link/_back-link.scss
@@ -25,6 +25,20 @@
     // Underline is provided by a bottom border
     text-decoration: none;
 
+
+    &:focus {
+      // Current margin minus 3px padding added by focus
+      margin-top: 12px;
+      // 14px for icon + 5px from focus style
+      padding-left: 19px;
+      border-bottom: 1px solid govuk-colour("yellow");
+
+      &:before {
+        top: 2px;
+        left: 5px;
+      }
+    }
+
     // Prepend left pointing arrow
     &:before {
       @include govuk-shape-arrow($direction: left, $base: 10px, $height: 6px);
--- a/src/components/checkboxes/_checkboxes.scss
+++ b/src/components/checkboxes/_checkboxes.scss
@@ -125,6 +125,7 @@
     // Since box-shadows are removed when users customise their colours, we set
     // a transparent outline that is shown instead.
     // https://accessibility.blog.gov.uk/2017/03/27/how-users-change-colours-on-websites/
+    border: 4px solid govuk-colour("black");
     outline: $govuk-focus-width solid transparent;
     outline-offset: $govuk-focus-width;
     box-shadow: 0 0 0 $govuk-focus-width $govuk-focus-colour;
--- a/src/components/footer/_footer.scss
+++ b/src/components/footer/_footer.scss
@@ -48,8 +48,7 @@
     // When focussed, the text colour needs to be darker to ensure that colour
     // contrast is still acceptable
     &:focus {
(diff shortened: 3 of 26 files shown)


src/components/footer/_footer.scss
```diff
@@ -7,11 +7,11 @@
...
   $govuk-footer-background: $govuk-canvas-background-colour;
-  $govuk-footer-border-top: #a1acb2;
-  $govuk-footer-border: govuk-colour("grey-2");
-  $govuk-footer-text: #454a4c;
+  $govuk-footer-border-top: $govuk-border-colour;
+  $govuk-footer-border: $govuk-border-colour;
+  $govuk-footer-text: $govuk-text-colour;
   $govuk-footer-link: $govuk-footer-text;
-  $govuk-footer-link-hover: #171819;
+  $govuk-footer-link-hover: $govuk-footer-text;
```
aliuk2012: All of the above has already been done.

src/components/checkboxes/_checkboxes.scss
```diff
@@ -125,6 +125,7 @@
     // Since box-shadows are removed when users customise their colours, we set
     // a transparent outline that is shown instead.
     // https://accessibility.blog.gov.uk/2017/03/27/how-users-change-colours-on-websites/
+    border: 4px solid govuk-colour("black");
     outline: $govuk-focus-width solid transparent;
```
36degrees: This might not be required any more, with the border-width change.

src/components/radios/_radios.scss
```diff
@@ -126,6 +126,7 @@
     // Since box-shadows are removed when users customise their colours we set a
     // transparent outline that is shown instead.
     // https://accessibility.blog.gov.uk/2017/03/27/how-users-change-colours-on-websites/
+    border: 4px solid govuk-colour("black");
     outline: $govuk-focus-width solid transparent;
```
36degrees: This may not be required any more either (as with checkboxes)

src/helpers/_links.scss
```diff
@@ -31,24 +31,29 @@
 @mixin govuk-link-style-default {
   &:link {
     color: $govuk-link-colour;
+    text-decoration-color: govuk-colour("light-blue");
```
36degrees: We're now not making this change (at least for now)

src/components/back-link/_back-link.scss
```diff
@@ -25,6 +25,11 @@
     // Underline is provided by a bottom border
     text-decoration: none;
 
+
+    &:focus {
```
36degrees: We may be able to close https://github.com/alphagov/govuk-frontend/issues/1236 with the focus state change here.

src/components/footer/_footer.scss
```diff
@@ -31,22 +31,24 @@
   }
 
   .govuk-footer__link {
-    @include govuk-focusable-fill;
+    @include govuk-focusable-text-link;
 
     &:link,
     &:visited {
       color: $govuk-footer-link;
+      text-decoration: none;
```
nickcolley: @dashouse are we sure about removing the underline in the footer, it makes all the links look like body content, there's no way of knowing something is clickable.

src/helpers/_focusable.scss
```diff
@@ -30,5 +29,28 @@
     outline: $govuk-focus-width solid $govuk-focus-colour;
     outline-offset: 0;
     background-color: $govuk-focus-colour;
+    text-decoration: none;
```
nickcolley: @dashouse can you confirm that you want to remove text-decoration here? It's think it's only used in the skip-link component after these changes, which I'm not sure is a good idea since the underline helps give it the affordance that it can be clicked?

We want to review these mixins once we've done all the work but just wondering here...

36degrees: This should only affect links that are focussed, I think? I believe the intention is to avoid a 'double underline' from the text decoration and the black box shadow together.

nickcolley: This mixin is no longer used with links, I think you're replying thinking of the mixin below this one potentially?

src/components/footer/_footer.scss
nickcolley: hmm im presuming this change it to make the footer consistent with the header...

src/helpers/_focusable.scss
nickcolley: Spoke to Dave and given that this is only used by tabbing to the component we've decided to leave the underline in for now.

src/components/footer/_footer.scss
nickcolley: Decided to go with this for now since it fits our navigation patterns.

I believe that this could fail WCAG though:

https://www.w3.org/TR/WCAG20-TECHS/F73.html

PR #1247, opened 2019-03-13
Fix tabs bullet point character encoding issue when not enhanced

### LINKED ISSUE #1246: Tabs with weird content rendering without JS
If I disable js on my computer (Chrome 72.036), I get this:

Screen Shot 2019-03-13 at 10 25 06

For [this](https://design-system.service.gov.uk/components/tabs/default/index.html) url.

nickcolley:
Will fix any issues with character encoding and is consistent with how we do this elsewhere (breadcrumbs)

I've not actually been able to replicate this, but it doesn't seem controversial to merge if the original contributor can confirm this fixes it.

Fixes #1246

### DIFF
--- a/src/components/tabs/_tabs.scss
+++ b/src/components/tabs/_tabs.scss
@@ -29,7 +29,7 @@
     margin-left: govuk-spacing(5);
 
     &::before {
-      content: "\2014 ";
+      content: "\2014 "; // "— "
       margin-left: - govuk-spacing(5);
       padding-right: govuk-spacing(1);
     }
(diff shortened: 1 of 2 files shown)


edwardhorsford:
I can't replicate the original issue - so I'm not sure I can confirm this fixes it. This renders fine for me though - but so does the original!

stevesims:
Checked the deployed version out, and this is working fine now when swapping character encoding in Safari.

Nice work 👍

src/components/tabs/_tabs.scss
```diff
@@ -29,7 +29,7 @@
     margin-left: govuk-spacing(5);
 
     &::before {
-      content: "— ";
+      content: "\2014 ";
```
36degrees: It'd be good to add a comment just so we know what this character is without having to look it up (and also in case anyone searches for it)

```suggestion
      content: "\2014 "; // "— "
```

Issue #1250, opened 2019-03-20
Date input component spacing problem at very narrow widths


andysellick:
![Screen Shot 2019-03-20 at 08 49 51](https://user-images.githubusercontent.com/861310/54671164-43666780-4aed-11e9-830e-9fc8a2fd3051.png)

If the date input is in a small enough space that it is forced to wrap, there's not enough vertical spacing between the label and the inputs above (see screenshot). Seems to occur if the component is in a container less than 200px wide.

This might seem unlikely to occur in the wild but we eventually want to use the date input component in finders to replace the existing single input field (for example on the [news and communication finder](https://www.gov.uk/search/news-and-communications)). Here the component would be in a narrow side column that can be reduced to a width less than 200px before hitting the mobile break point.

(there is some right margin on the 'year' input that could be removed, possibly, although that wouldn't completely solve the issue)

nickcolley:
We could consider using flexbox here...

dashouse:
Hey @andysellick,

This is a use case of this component that wasn't entirely expected, we tend to support something like this down to about 320px (inclusive of page gutter).

When something like this is used on a larger screen but in a small container the smaller font size won't have kicked in so the component will also be bigger than if it was 100% of a smaller screen.

There's a couple of options if you want to get this working for you now,

a) Depending on what you're doing with the other components in the refine panel you could reduce the font size to 16px which will also reduce the size of the inputs. I wouldn't imagine you'd want to do this unless all of the other form elements were also 16px

b) Create your own modifier for the input widths. Currently the widths are defines by the classes `govuk-input--width-2` and `govuk-input--width-4`. These set a width using the `ex` unit. 

They are a little bigger than they'd actually need to be so you could delete these classes, and make your own modifier of the `govuk-date-input__input` class for each input. 

A small tweak would be to change the year from `govuk-input--width-4` to `govuk-input--width-3` and see if that helps. You'll still be able to fit 4 characters in it, the extra space is there for a autocomplete icon that wouldn't be present in your use case.

andysellick:
Thanks @dashouse. @alex-ju has implemented a fix to remove the margin right from the year input and swap around the margin so each of the input/label groups has a bit of top margin, without changing the layout of anything else. Seems to work.

colinrotherham:
Since date input items use `display: inline-block` no flexbox needed.

Adding `white-space: nowrap;` will prevent wrapping.

timpaul:
[...]

PR #1257, opened 2019-03-27
Prevent wrapping of individual date inputs

colinrotherham:
Date inputs should group together as if they're a single field.

This fix keeps them together using `white-space: nowrap`, and removes the stray right margin from the year (reducing component width to *200px*).

Fixes @andysellick's issue on https://github.com/alphagov/govuk-frontend/issues/1250

Issue

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,12 @@
 
   ([PR #1265](https://github.com/alphagov/govuk-frontend/pull/1265))
 
+- Prevent wrapping of individual date inputs
+
+  Date inputs should group together as if they're a single field, this fix keeps them together using `white-space: nowrap`.
+
+  ([PR #1257](https://github.com/alphagov/govuk-frontend/pull/1257))
+
 ## 2.9.0 (Feature release)
 
 🆕 New features:
--- a/src/components/date-input/_date-input.scss
+++ b/src/components/date-input/_date-input.scss
@@ -12,12 +12,16 @@
     @include govuk-clearfix;
     // font-size: 0 removes whitespace caused by inline-block
     font-size: 0;
+    white-space: nowrap;
   }
 
   .govuk-date-input__item {
     display: inline-block;
-    margin-right: govuk-spacing(4);
     margin-bottom: 0;
+  }
+
+  .govuk-date-input__item + .govuk-date-input__item {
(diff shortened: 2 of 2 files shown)


colinrotherham:
Sorry just spotted `:last-of-type` is IE9+ so moved margin to the left of siblings instead:

```scss
.govuk-date-input__item + .govuk-date-input__item {
  margin-left: govuk-spacing(4);
}
```

colinrotherham:
Rebased with `master`

hannalaakso:
Thanks @colinrotherham  🙂  It looks like this will prevent the wrapping of the items even if they don't fit in the container. Here, the date input is inside one of three `.govuk-grid-column-one-third` and overflows the container on smaller screens. 

Screen Shot 2019-04-09 at 16 28 50

I wonder if [the fix by GOV.UK Publishing](https://govuk-publishing-components.herokuapp.com/component-guide/date_input) to https://github.com/alphagov/govuk-frontend/issues/1250 might work for us. It doesn't prevent wrapping but it looks a bit neater.

colinrotherham:
Hi @hannalaakso yeah that was intentional. Just like the other fixed-width inputs I thought it shouldn't wrap.

Same if we added the `--width-20` modifier.

If it's supposed to wrap then I'm mistaken! Sorry.

I'll amend this pull request to support wrapping with the appropriate vertical spacing.

Thanks

hannalaakso:
Thanks @colinrotherham. We can try to get some design clarification to make sure I'm not making wrong assumptions here 😄 

@dashouse Would you have any thoughts on how this this should be behave from a design perspective?

dashouse:
Thanks for raising this (as always). Sorry to be a blocker, but I don't believe we have a solid decision on how we want the component to behave in this scenario.

The solution to the original issue on GOV.UK will be to "make it fit" in that instance using a few techniques that were discussed in the comment thread. This is because we have some knowns about that layout, for example the minimum width of the filter column.

As global change to this component I think we need to make a few decisions based on unknowns...for example, should we stack inputs, reduce space between inputs, change the input size, consider mobile breakpoint design vs limited width container design, should we be using flexbox for this...etc

For now it's probably best to close this PR, make sure this is properly recorded as an issue and potentially add some options.

PR #1259, opened 2019-03-28
Don't output an empty ul if 0 items provided

edwardhorsford:
I need to provide an empty action option so that I don't get weird gaps like this: 
Screenshot 2019-03-28 at 12 23 55

You can fix this by providing an empty action.items - but right now this results in an empty ul. This pr doesn't spit out the ul if there's 0 items.

NB: untested!

### DIFF
--- a/src/components/summary-list/_summary-list.scss
+++ b/src/components/summary-list/_summary-list.scss
@@ -119,12 +119,39 @@
     border: 0;
   }
 
+  // No border on entire summary list
   .govuk-summary-list--no-border {
-    .govuk-summary-list__key,
-    .govuk-summary-list__value,
-    .govuk-summary-list__actions,
-    .govuk-summary-list__row {
+    @include govuk-media-query($until: tablet) {
+      .govuk-summary-list__row {
+        border: 0;
+      }
+    }
+
+    @include govuk-media-query($from: tablet) {
+      .govuk-summary-list__key,
+      .govuk-summary-list__value,
+      .govuk-summary-list__actions {
+        // Remove 1px border, add 1px height back on
+        padding-bottom: govuk-spacing(2) + 1px;
+        border: 0;
+      }
+    }
+  }
+
+  // No border on specific rows
+  .govuk-summary-list__row--no-border {
+    @include govuk-media-query($until: tablet) {
       border: 0;
+    }
+
+    @include govuk-media-query($from: tablet) {
+      .govuk-summary-list__key,
+      .govuk-summary-list__value,
+      .govuk-summary-list__actions {
+        // Remove 1px border, add 1px height back on
+        padding-bottom: govuk-spacing(2) + 1px;
+        border: 0;
+      }
     }
   }
 }
--- a/src/components/summary-list/_summary-list.scss
+++ b/src/components/summary-list/_summary-list.scss
@@ -80,6 +80,13 @@
     }
   }
 
+  // Expand width when value is last column (no action)
+  .govuk-summary-list__value:last-child {
+    @include govuk-media-query($from: tablet) {
+      width: 70%;
+    }
+  }
+
   .govuk-summary-list__value > p {
     margin-bottom: govuk-spacing(2);
   }
--- a/src/components/summary-list/template.njk
+++ b/src/components/summary-list/template.njk
@@ -15,7 +15,7 @@
 
 <dl class="govuk-summary-list {%- if params.classes %} {{ params.classes }}{% endif %}"{% for attribute, value in params.attributes %} {{attribute}}="{{value}}"{% endfor %}>
   {% for row in params.rows %}
-    <div class="govuk-summary-list__row">
+    <div class="govuk-summary-list__row {%- if row.classes %} {{ row.classes }}{% endif %}">
       <dt class="govuk-summary-list__key {%- if row.key.classes %} {{ row.key.classes }}{% endif %}">
         {{ row.key.html | safe if row.key.html else row.key.text }}
       </dt>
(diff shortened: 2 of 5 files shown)


timpaul:
Thanks for this Ed. Looks good, but we'll need to write some tests for it before we merge.

colinrotherham:
Hi @timpaul I've added some tests for @edwardhorsford as one of "Check answers" pages has a row without an actions column and this will affect us too—might as well help out!

This includes a few other changes as well:

**1. Allows custom classes on entire rows**
To support custom styling of whole rows at a time

**2. Fixes the 1px row height change when borders are removed**
i.e. Uses `border-bottom-color: transparent` rather than `border: 0`

**3. Fixes The text alignment when the actions column isn't added**
See below, as the 30% and 50% column widths didn't add up to 100%

Alignment

36degrees:
> **2. Fixes the 1px row height change when borders are removed**
> i.e. Uses transparent border-color rather than removing it entirely:
> 
> **Before**
> This removes 1px from each summary list row
> 
> ```css
> border: 0;
> ```
> 
> **After**
> The height of the row stays the same
> 
> ```css
> border-bottom-color: transparent;
> ```

This'll affect anyone using custom colours in their browser – they'll see a border when others don't. What's the impact of the 1px smaller row height?

36degrees:
As far as I'm aware, there's no way to stop borders from becoming opaque when colours are overridden, even using `!important`.

This screenshot illustrates the problem:

Screenshot 2019-04-24 at 13 24 44BST

I think in this case the best thing to do is roll back that change until we can find a solution that works for everyone, and we can get the rest of these changes merged.

Ta!

src/components/summary-list/template.njk
```diff
@@ -6,20 +6,27 @@
     {% endif -%}
   </a>
 {% endmacro -%}
+
+{# Determine if we need 2 or 3 columns #}
+{% set hasActions = false %}
+{% for row in params.rows %}
+  {% set hasActions = row.actions.items if row.actions.items else hasActions %}
```
36degrees: As this is a boolean, should we set it to `true if row.actions.items`, rather than the contents of `row.actions.items`?

src/components/summary-list/template.njk
```diff
@@ -6,20 +6,27 @@
...
       <dd class="govuk-summary-list__value {%- if row.value.classes %} {{ row.value.classes }}{% endif %}">
         {{ row.value.html | indent(8) | trim | safe if row.value.html else row.value.text }}
       </dd>
-      {% if row.actions.items %}
+      {% if row.actions.items.length %}
         <dd class="govuk-summary-list__actions {%- if row.actions.classes %} {{ row.actions.classes }}{% endif %}">
           {% if row.actions.items.length == 1 %}
             {{ _actionLink(row.actions.items[0]) | indent(12) | trim }}
-          {% else %}
+          {% elseif row.actions.items.length > 1 %}
```
36degrees: What else could it be, if `row.actions.items.length` is truthy and it's not `1`?

src/components/summary-list/template.njk
```diff
@@ -6,20 +6,27 @@
     {% endif -%}
   </a>
 {% endmacro -%}
+
+{# Determine if we need 2 or 3 columns #}
+{% set hasActions = false %}
```
36degrees: I wonder if this could be clearer, perhaps something like

```suggestion
{% set anyRowHasActions = false %}
```

src/components/summary-list/template.test.js
```diff
@@ -314,6 +330,66 @@ describe('Data list', () => {
...
+            }
+          ]
+        })
+
+        const $component = $('.govuk-summary-list')
+        const $action = $component.find('.govuk-summary-list__actions')
+
+        expect($action.length).toEqual(0)
+      })
+      it('adds dummy action columns when only some rows have actions', async () => {
```
36degrees: 👌

src/components/summary-list/_summary-list.scss
```diff
@@ -80,6 +80,12 @@
     }
   }
 
+  .govuk-summary-list__value:last-child {
```
36degrees: This could do with a comment I think

src/components/summary-list/template.njk
colinrotherham: You're right, this got refactored slightly higher up so can go now.

Was part of the first commit: https://github.com/alphagov/govuk-frontend/pull/1259/commits/f0df010ac8fec2047431b3223b8e32a2dc81670a

colinrotherham:
@36degrees Regarding the borders, usefully each cell has 10px bottom padding.

So with borders turned off, could this be the fix? i.e. Add 1px back on:

```scss
.govuk-summary-list__key,
.govuk-summary-list__value,
.govuk-summary-list__actions {
  padding-bottom: govuk-spacing(2) + 1px;
  border: 0;
}
```

src/components/summary-list/template.njk
colinrotherham: Thanks, sorted

36degrees:
Makes sense to me 👍

src/components/summary-list/template.njk
colinrotherham: Yep I'm happy with this

colinrotherham:
@36degrees Review comments all addressed. Thanks again.

On mobile "no-border" has margin collapsing between the rows, so I've not addressed the height jump here as I imagine that's intentional.

36degrees:
This is looking good I think, thanks again.

It just needs documenting in the changelog and then I can approve it.

colinrotherham:
@36degrees Thanks.

Changelog added, split into both fixes/features.

Issue #1248, opened 2019-03-13
Checkbox can end up with multiple `aria-describedby` attributes when used without a fieldset but with a hint


36degrees:
The checkbox component has some logic to move the association with the error message (`aria-describedby`) to the input _if there is no fieldset_ (e.g. for a single input).

It _also_ has logic to associate the hint for the individual checkbox if a hint is provided.

We don't handle the (admittedly fairly unlikely) case where both of these things are true, so the checkbox will end up with two `aria-describedby` attributes (`aria-describedby="id-of-error-message" aria-describedby="id-of-hint"`)

https://github.com/alphagov/govuk-frontend/blob/abb0c36f9900afc2093361acd3ac5363a0ac219b/src/components/checkboxes/template.njk#L63-L65

You can reproduce this with the following macro call:

```js
{% from "checkboxes/macro.njk" import govukCheckboxes %}

{{ govukCheckboxes({
  "name": "t-and-c",
  "errorMessage": {
    "text": "Please accept the terms and conditions"
  },
  "items": [
    {
      "value": "yes",
      "text": "I agree to the terms and conditions",
      "hint": {
        "text": "Go on, you know you want to!"
      }
    }
  ]
}) }}
```

which produces this HTML:

```html
<div class="govuk-form-group govuk-form-group--error">
  <span id="t-and-c-error" class="govuk-error-message">
    Please accept the terms and conditions
  </span>
  <div class="govuk-checkboxes">
    <div class="govuk-checkboxes__item">
      <input class="govuk-checkboxes__input" id="t-and-c-1" name="t-and-c" type="checkbox" value="yes" aria-describedby="t-and-c-error" aria-describedby="t-and-c-1-item-hint">
      <label class="govuk-label govuk-checkboxes__label" for="t-and-c-1">
        I agree to the terms and conditions
      </label>      <span id="t-and-c-1-item-hint" class="govuk-hint govuk-checkboxes__hint">
        Go on, you know you want to!
      </span>    </div>
  </div>
</div>
```

instead of:

```html
<div class="govuk-form-group govuk-form-group--error">
  <span id="t-and-c-error" class="govuk-error-message">
    Please accept the terms and conditions
  </span>
  <div class="govuk-checkboxes">
    <div class="govuk-checkboxes__item">
      <input class="govuk-checkboxes__input" id="t-and-c-1" name="t-and-c" type="checkbox" value="yes" aria-describedby="t-and-c-error t-and-c-1-item-hint">
[...]

PR #1265, opened 2019-04-01
Prevent duplicate checkbox aria-describedby

colinrotherham:
I've spent a bit of time in the checkbox component recently (passing `describedBy` around) so I've fixed https://github.com/alphagov/govuk-frontend/issues/1248 as it'll benefit our work as well.

### DIFF
--- a/src/components/checkboxes/template.njk
+++ b/src/components/checkboxes/template.njk
@@ -54,15 +54,15 @@
     {% set name = item.name if item.name else params.name %}
     {% set conditionalId = "conditional-" + id %}
     {% set hasHint = true if item.hint.text or item.hint.html %}
-    {% set itemHintId = id + '-item-hint' %}
+    {% set itemHintId = id + "-item-hint" if hasHint else "" %}
+    {% set itemDescribedBy = describedBy if not hasFieldset else "" %}
+    {% set itemDescribedBy = (itemDescribedBy + " " + itemHintId) | trim %}
     <div class="govuk-checkboxes__item">
       <input class="govuk-checkboxes__input" id="{{ id }}" name="{{ name }}" type="checkbox" value="{{ item.value }}"
       {{-" checked" if item.checked }}
       {{-" disabled" if item.disabled }}
-      {#- fieldset is false by default -#}
-      {%- if (not hasFieldset) and ((describedBy | length) > 0) %} aria-describedby="{{ describedBy }}"{% endif -%}
       {%- if item.conditional %} data-aria-controls="{{ conditionalId }}"{% endif -%}
-      {%- if hasHint %} aria-describedby="{{ itemHintId }}"{% endif -%}
+      {%- if itemDescribedBy %} aria-describedby="{{ itemDescribedBy }}"{% endif -%}
       {%- for attribute, value in item.attributes %} {{ attribute }}="{{ value }}"{% endfor -%}>
       {{ govukLabel({
         html: item.html,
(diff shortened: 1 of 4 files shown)


src/components/checkboxes/template.njk
```diff
@@ -55,14 +55,14 @@
     {% set conditionalId = "conditional-" + id %}
     {% set hasHint = true if item.hint.text or item.hint.html %}
     {% set itemHintId = id + '-item-hint' %}
+    {% set itemDescribedBy = describedBy if not hasFieldset else "" %}
+    {% set itemDescribedBy = (describedBy + " " + itemHintId) | trim if hasHint else itemDescribedBy %}
```
hannalaakso: Do you think we should consider breaking this out a bit to might make it more readable. Something like:
```
{% set itemDescribedBy = "" %}
{% if not hasFieldset %}
   {% set itemDescribedBy = describedBy %}
{% endif %}
{% if hasHint %}
   {% set itemDescribedBy = (describedBy + " " + itemHintId) | trim %}
{% endif %}
```

aliuk2012: I agree with @hannalaakso this was a bit difficult to read. 

Could this not do the same thing you are trying to achieve?

```js
    {% set itemHintId = id + '-item-hint' if hasHint %}
    {% set itemDescribedBy = (describedBy + " " + itemHintId) | trim %}
```

You shouldn't need line 58 if the `itemHintId` is only set if there is a fieldset. If it doesn't have a hint then the last lines `trim` should remove any trailing spaces.

colinrotherham:
Thanks for the feedback. I've been in Nunjucks land a while now, will make it simpler!

src/components/checkboxes/template.njk
colinrotherham: Sorry @hannalaakso @aliuk2012 both of your versions would affect fieldsets of checkboxes which have an error message, giving us two IDs in `aria-describedby`.

**Original version**
```html
waste-1-item-hint
```

**Our suggestions**
```html
waste-error waste-1-item-hint
```

i.e. The fieldset hint would be applied to both the fieldset and the input, which isn't intended!

Admittedly there is a gap in your test coverage here, so I've added a new example.

Leave it with me.

colinrotherham: Is this alright?

```njk
{% set itemHintId = id + "-item-hint" if hasHint else "" %}
{% set itemDescribedBy = describedBy if not hasFieldset else "" %}
{% set itemDescribedBy = (itemDescribedBy + " " + itemHintId) | trim %}
```

All tests pass, including a new one here https://github.com/alphagov/govuk-frontend/pull/1265/commits/d9c4957179324727b261add2345d1324f680f320 that tests for the _checkboxes with error message and hints on items_ edge case.

colinrotherham: I've kept `else ""` so we don't end up with `undefined` as a string in `aria-describedby`

colinrotherham:
@aliuk2012 I needed to support this new hybrid case (previously untested) so I've pushed a slightly revised version up: https://govuk-frontend-review-pr-1265.herokuapp.com/components/checkboxes/with-error-message-and-hints-on-items/preview

As mentioned on the diffs, it's to cater for:

1. Shows error message, so `aria-describedby` on fieldset
2. Hint on checkboxes, so `aria-describedby` on inputs
3. Error message ID should not be in input `aria-describedby`

Example

@hannalaakso If you'd still prefer those Nunjucks if/else simplifying further let me know.

colinrotherham:
Rebased with `master`

Issue #1264, opened 2019-03-29
Select element with long option text is forced outside the container when placed in a fieldset


kr8n3r:
See here https://picturesque-chair.glitch.me/ on mobile (screeshot below)

I _believe_ this is because browsers apply `min-inline-size: min-content;` to the `fieldset`, which [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/min-inline-size) says:
> ...defines the horizontal or vertical minimal size of an element's block, depending on its writing mode. It corresponds to either the min-width or the min-height property, depending on the value of writing-mode.

Applying `min-width: 0;` to `.govuk-fieldset` fixes it, but I haven't tested cross-browsers extensively (see screnshots at the end)

![picturesque-chair glitch me_(iPhone 5_SE) (1)](https://user-images.githubusercontent.com/3758555/55258087-c314d480-5259-11e9-9f06-c7922434134e.png)

Some tests:

iphone 6s
<img width="457" alt="Screen Shot 2019-03-29 at 19 42 41" src="https://user-images.githubusercontent.com/3758555/55258515-ef7d2080-525a-11e9-9b93-1b42076145cb.png">

iphone X
<img width="501" alt="Screen Shot 2019-03-29 at 19 44 16" src="https://user-images.githubusercontent.com/3758555/55258564-1cc9ce80-525b-11e9-9f26-35ebd5d25ced.png">

galaxy note 9
<img width="457" alt="Screen Shot 2019-03-29 at 19 45 05" src="https://user-images.githubusercontent.com/3758555/55258598-3703ac80-525b-11e9-9925-c99f397f32bb.png">

colinrotherham:
@kr8n3r Thanks for the online demo.

As soon as you said `fieldset` I knew we'd hit this ancient Firefox bug from 2009!
https://bugzilla.mozilla.org/show_bug.cgi?id=504622

Your fix works great, but for Firefox only in v53 onwards.

There was a hacky workaround documented here:
https://thatemil.com/blog/2015/01/03/reset-your-fieldset/

```css
body:not(:-moz-handler-blocked) fieldset {
  display: table-cell;
}
```

But changing the `fieldset` display mode from _block_ might have accessibility implications.

### Firefox 52

![Firefox 52](https://user-images.githubusercontent.com/415517/55389160-9ba95a80-552c-11e9-8309-02152853de3c.png)

kr8n3r:
@colinrotherham 😢 
[...]

PR #1269, opened 2019-04-05
Ensure fieldset never exceeds max-width

colinrotherham:
This fix ensures that both WebKit/Blink and Firefox are prevented from expanding their fieldset widths to the content's minimum size.

**Online demo**
https://tourmaline-spoonbill.glitch.me/

This was preventing `max-width: 100%` from being applied to select menus inside a fieldset. There's further discussion in ["Reset your fieldset"](https://thatemil.com/blog/2015/01/03/reset-your-fieldset/) and raised by @kr8n3r in [issue #1264](https://github.com/alphagov/govuk-frontend/issues/1264).

Due to Firefox's legacy fieldset behaviour, I've also added a workaround for Firefox < 53.

```scss
// Fix for Firefox < 53
// https://bugzilla.mozilla.org/show_bug.cgi?id=504622
@supports not (caret-color: auto) {
  .govuk-fieldset,
  x:-moz-any-link {
    display: table-cell;
  }
}
```

It uses `x:-moz-any-link` to make the snippet Firefox-only, but with `@supports not (caret-color: auto)` to exclude Firefox >= 53 when _caret-color_ support was added.

Firefox 52

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,16 @@
 
   ([PR #1265](https://github.com/alphagov/govuk-frontend/pull/1265))
 
+- Ensure fieldset never exceeds max-width
+
+  This fix ensures that both WebKit/Blink and Firefox are prevented from expanding their fieldset widths to the content's minimum size.
+
+  This was preventing `max-width: 100%` from being applied to select menus inside a fieldset.
+
+  See discussion in ["Reset your fieldset"](https://thatemil.com/blog/2015/01/03/reset-your-fieldset/) and raised by [issue #1264](https://github.com/alphagov/govuk-frontend/issues/1264)
+
+  ([PR #1269](https://github.com/alphagov/govuk-frontend/pull/1269))
+
 ## 2.9.0 (Feature release)
 
 🆕 New features:
--- a/src/components/fieldset/_fieldset.scss
+++ b/src/components/fieldset/_fieldset.scss
@@ -4,10 +4,20 @@
 
 @include govuk-exports("govuk/component/fieldset") {
   .govuk-fieldset {
+    min-width: 0;
     margin: 0;
     padding: 0;
     border: 0;
     @include govuk-clearfix;
+  }
+
+  // Fix for Firefox < 53
+  // https://bugzilla.mozilla.org/show_bug.cgi?id=504622
+  @supports not (caret-color: auto) {
+    .govuk-fieldset,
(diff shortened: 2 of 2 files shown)


colinrotherham:
Rebased with `master`

src/components/fieldset/_fieldset.scss
```diff
@@ -4,12 +4,22 @@
...
 @include govuk-exports("govuk/component/fieldset") {
   .govuk-fieldset {
+    min-width: 0;
     margin: 0;
     padding: 0;
     border: 0;
     @include govuk-clearfix;
   }
 
+  // Fix for Firefox < 53
```
36degrees: Given Firefox 53 was released in April 2017 and we're now on version 66 I _think_ we can probably omit this? On GOV.UK, 6.18% of Firefox users were using a version older than 53, which represents about 0.29% of the total traffic.

Thoughts?

colinrotherham: @36degrees Honestly I'll leave this one down to you and the team. I was stung many times in my career by Firefox fieldset oddities so maybe I'm emotionally attached 😁

colinrotherham: Actually, whilst the default browser was changed to IE11 a few months ago, I believe DWP Jobcentre work coaches and other staff still have access to Firefox 45 ESR.

It's included in various automated tests.

Is that reason to keep it? Thanks

CHANGELOG.md
```diff
@@ -34,6 +34,16 @@
 
   ([PR #1265](https://github.com/alphagov/govuk-frontend/pull/1265))
 
+- Ensure fieldset never exceeds max-width
+
+  This fix ensures that both WebKit/Blink and Firefox are prevented from expanding their fieldset widths to the content's minimum size.
+
+  This was preventing `max-width: 100%` from being applied to select menus inside a fieldset.
+
+  See discussion in ["Reset your fieldset"](https://thatemil.com/blog/2015/01/03/reset-your-fieldset/) and raised by [issue #1264](https://github.com/alphagov/govuk-frontend/pull/N)
```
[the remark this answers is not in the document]
colinrotherham: Fixed this, thanks

colinrotherham:
@36degrees Just the last review point to address now: https://github.com/alphagov/govuk-frontend/pull/1269#discussion_r275388327

Let me know if I should remove the Firefox < 53 fix.

PR #1276, opened 2019-04-16
Fix review app banner for IE8

hannalaakso:
- Pull banner styles into the IE8 stylesheet 
- Fix a link style as IE8 doesn't support `:not` 
- Fix font size for IE8 as it doesn't support `rem`
- Update code so it passes `Standard` linting


36degrees:
It'd be good to fix the font size (I think it just needs a px fallback, or maybe just update it to use the proper responsive mixin) but otherwise this looks good to me 👍

Issue #1262, opened 2019-03-29
Review app should be able to include code from legacy repos


hannalaakso:
Spiked in https://github.com/alphagov/govuk-frontend/issues/1158

## Problem 
We want to be able to switch including the legacy repos in review app on and off to give us confidence that we're not breaking things in various states (compability mode, legacy code included) when we make the WGAG 2.1 changes to styles and components. Various people in the team need to be able to test the changes in various states.

This work will also be valuable going forward when testing changes to components. We should ideally look at adding automated visual regression testing soon.

## Solution
We discussed toggling the states in three different ways:
1. UI component toggle
 - Pro: Not having to know what values/pages are needed for toggling
- Con: Can't link to pages, either rely on JS or pass this to backend in form. 
2. Separate page to have values on/off
- Pro: Linkable. Would limit to certain number of pages.
- Con: Would have to limit number of pages this would work on. Manually having to add this to page templates.
3. Query string in URL to set eg. `/?legacy=on`
- Pro: Linkable. Easy to use with range of pages.
- Con: Having to know what values to use. Not knowing what default is unless you read docs. 

We decided to go with 3 as this is the most flexible on. We will allow both legacy code and compatibility mode to be toggled on/off with a query string.

## States

There will therefore be four scenarios to test:
 ~1. Compatibility mode on, legacy repos not included
       - Use the existing colour palette
       - Use the new focus styles~
  2. Compatibility mode off, legacy repos not included
       - Use the new colour palette
       - Use the new focus styles
  3. Compatibility mode on, legacy repos included
       - Use the existing colour palette
       - Use the new focus styles
       - Optional: Include some components from legacy repos on the page
~4. Compatibility mode off, legacy repos included
       - Use the new colour palette
       - Use the new focus styles~

Irrelevant ones ~crossed out~

## Definitions

By compatibility mode we mean [this](https://github.com/alphagov/govuk-frontend/blob/master/docs/installation/installing-with-npm.md#compatibility-mode).

By legacy repos we mean:
- https://github.com/alphagov/govuk_frontend_toolkit
- https://github.com/alphagov/govuk_elements
- https://github.com/alphagov/govuk_template

## Checklist
- [x] Add legacy styles to review app
- [x] Create legacy stylesheets
- [x] Add tasks to compile legacy sass stylesheets
- [x] Allow template CSS to be included in layout
- [x] Configure Express to process query strings and pass these into view if compatibility/legacy
[...]

PR #1279, opened 2019-04-17
Conditionally load in legacy stylesheets along with govuk-frontend

aliuk2012:
Legacy mode is a means to test legacy css (GOV.UK Elements, GOV.UK template
etc.) along side govuk-frontend.  The review app will by default only use
govuk-frontend. However, if you add the query string of `?legacy=1` to
the url, the application should load govuk-elements, govuk-template and
govuk-frontend. eg `https://govuk-frontend-review-pr-1279.herokuapp.com/?legacy=1`

It should also remember the mode as you navigate through
the review app.

Part of https://github.com/alphagov/govuk-frontend/issues/1262

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -71,6 +71,17 @@ module.exports = (options) => {
   // Handle the banner component serverside.
   require('./banner.js')(app)
 
+  // Define middleware for all routes
+  app.use('*', function (request, response, next) {
+    response.locals.legacy = request.query['legacy'] === '1'
+    if (response.locals.legacy) {
+      response.locals.legacy_query = '?legacy=1'
+    } else {
+      response.locals.legacy_query = ''
+    }
+    next()
+  })
+
   // Define routes
 
   // Index page - render the component list template
--- a/app/app.js
+++ b/app/app.js
@@ -75,9 +75,9 @@ module.exports = (options) => {
   app.use('*', function (request, response, next) {
     response.locals.legacy = request.query['legacy'] === '1'
     if (response.locals.legacy) {
-      response.locals.legacy_query = '?legacy=1'
+      response.locals.legacyQuery = '?legacy=1'
     } else {
-      response.locals.legacy_query = ''
+      response.locals.legacyQuery = ''
     }
     next()
   })
--- a/app/views/layouts/_generic.njk
+++ b/app/views/layouts/_generic.njk
@@ -13,28 +13,26 @@
     <link rel="stylesheet" media="all" href="/vendor/govuk_template/fonts.css"/>
     <!--[if !IE 8]><!-->
       <link rel="stylesheet" href="/public/app-legacy.css">
-      <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/prism/1.6.0/themes/prism.min.css" integrity="sha256-vtR0hSWRc3Tb26iuN2oZHt3KRUomwTufNIf5/4oeCyg=" crossorigin="anonymous" />
-      <script src="https://cdnjs.cloudflare.com/ajax/libs/prism/1.6.0/prism.min.js" integrity="sha256-Zb9yKJ/cfs+jG/zIOFL0QEuXr2CDKF7FR5YBJY3No+c=" crossorigin="anonymous"></script>
     <!--<![endif]-->
     <!--[if IE 8]>
       <link rel="stylesheet" href="/public/app-legacy-ie8.css">
     <![endif]-->
-    <!--[if lt IE 9]>
-      <script src="/vendor/html5-shiv/html5shiv.js"></script>
-    <![endif]-->
   {% else %}
     <!--[if !IE 8]><!-->
       <link rel="stylesheet" href="/public/app.css">
-      <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/prism/1.6.0/themes/prism.min.css" integrity="sha256-vtR0hSWRc3Tb26iuN2oZHt3KRUomwTufNIf5/4oeCyg=" crossorigin="anonymous" />
-      <script src="https://cdnjs.cloudflare.com/ajax/libs/prism/1.6.0/prism.min.js" integrity="sha256-Zb9yKJ/cfs+jG/zIOFL0QEuXr2CDKF7FR5YBJY3No+c=" crossorigin="anonymous"></script>
     <!--<![endif]-->
     <!--[if IE 8]>
       <link rel="stylesheet" href="/public/app-ie8.css">
     <![endif]-->
-    <!--[if lt IE 9]>
-      <script src="/vendor/html5-shiv/html5shiv.js"></script>
-    <![endif]-->
   {% endif %}
+
+  <!--[if !IE 8]><!-->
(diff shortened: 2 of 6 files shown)


app/views/layouts/_generic.njk
```diff
@@ -1,17 +1,32 @@
...
-    <script src="https://cdnjs.cloudflare.com/ajax/libs/prism/1.6.0/prism.min.js" integrity="sha256-Zb9yKJ/cfs+jG/zIOFL0QEuXr2CDKF7FR5YBJY3No+c=" crossorigin="anonymous"></script>
-  <!--<![endif]-->
-  <!--[if IE 8]>
-    <link rel="stylesheet" href="/public/app-ie8.css">
-  <![endif]-->
-  <!--[if lt IE 9]>
-    <script src="/vendor/html5-shiv/html5shiv.js"></script>
-  <![endif]-->
+
+  {% if legacy %}
```
36degrees: You also need to pull in the stylesheets from GOV.UK Template separately:

- https://govuk-frontend-review-pr-1279.herokuapp.com/vendor/govuk_template/govuk-template.css
- https://govuk-frontend-review-pr-1279.herokuapp.com/vendor/govuk_template/govuk-template-print.css
- https://govuk-frontend-review-pr-1279.herokuapp.com/vendor/govuk_template/govuk-template-ie8.css)

app/app.js
```diff
@@ -71,6 +71,17 @@ module.exports = (options) => {
   // Handle the banner component serverside.
   require('./banner.js')(app)
 
+  // Define middleware for all routes
+  app.use('*', function (request, response, next) {
+    app.locals.legacy = request.query['legacy'] === '1'
```
36degrees: I think we want these to be [`res.locals`](https://expressjs.com/en/api.html#res.locals) as this is scoped to the specific request that's being made.

app/views/macros/showExamples.njk
```diff
@@ -17,7 +17,7 @@
     <div class="govuk-width-container">
       <div class="govuk-heading-m">
         <h3 id="heading-{{ exampleSlug }}" class="app-!-di">{{ heading | safe }}</h3>
-        <a href="{{ path }}" class="govuk-link govuk-!-margin-left-1 govuk-!-font-size-16">
+        <a href="{{ path }}{{ legacy_query }}" class="govuk-link govuk-!-margin-left-1 govuk-!-font-size-16">
```
36degrees: These don't appear to be working – I think because the macro doesn't have access to local variables in the same way, only variables that are passed to the macro?

app/views/macros/showExamples.njk
```diff
@@ -31,7 +31,7 @@
     <div class="app-component-preview">
       <iframe
       class="js-component-preview app-component-preview__iframe"
-      src="{{ path }}?iframe=true"
+      src="{{ path }}?{{ legacy_query }}iframe=true"
```
36degrees: This will result in e.g. `/components/accordion/preview?legacy=1iframe=true` – there's no `&` joining the two parameters together.

aliuk2012: thanks it is on my todo I just wanted to push up what I have done so far ;)

app/views/layouts/_generic.njk
aliuk2012: Ah this might explain why there fonts appear to be different when you toggle between the two modes

36degrees: If you mean the component page, I believe that's happening because the markup for the headings is weird (the heading class is on a parent div, not the heading, so the reset on the heading wins)

app/views/macros/showExamples.njk
aliuk2012: I'm now passing the local variables to the macro and this should now be working

app/views/layouts/_generic.njk
```diff
@@ -1,17 +1,40 @@
...
+      <link rel="stylesheet" href="/vendor/govuk_template/govuk-template.css">
+    <!--<![endif]-->
+    <!--[if IE 8]>
+      <link rel="stylesheet" href="/vendor/govuk_template/govuk-template-ie8.css">
+    <![endif]-->
+    <link rel="stylesheet" media="print" href="/vendor/govuk_template/govuk-template-print.css">
+    <link rel="stylesheet" media="all" href="/vendor/govuk_template/fonts.css"/>
+    <!--[if !IE 8]><!-->
+      <link rel="stylesheet" href="/public/app-legacy.css">
+      <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/prism/1.6.0/themes/prism.min.css" integrity="sha256-vtR0hSWRc3Tb26iuN2oZHt3KRUomwTufNIf5/4oeCyg=" crossorigin="anonymous" />
```
36degrees: There's a fair bit of duplication between these two flows (Prism, HTML5shiv) – can we split that stuff out so that it's only in one place?

app/views/layouts/all-components.njk
```diff
@@ -22,6 +22,6 @@
   {% for data in componentData %}
     {% set componentName = data.componentName %}
     {% set componentNameHuman = componentName | replace("-", " ") %}
-    {{ showExamples(componentName, componentNameHuman, data) }}
+    {{ showExamples(componentName, componentNameHuman, data, legacy_query) }}
```
36degrees: We seem to use camel case pretty consistently for view variables – should we update this to match?

app/views/layouts/component.njk
```diff
@@ -33,6 +33,6 @@
   </div>
 
   {% block examples %}
-    {{ showExamples(componentName, componentNameHuman, componentData) }}
+    {{ showExamples(componentName, componentNameHuman, componentData,legacy_query) }}
```
36degrees: It'd be good if the breadcrumbs in this view also preserved the legacy state (by using `legacy_query`)

aliuk2012: Well spotted, I've added the mode to this breadcrumb and the all-components breadcrumb too.

app/views/layouts/_generic.njk
```diff
@@ -1,17 +1,40 @@
...
+      <link rel="stylesheet" href="/public/app-legacy-ie8.css">
+    <![endif]-->
+    <!--[if lt IE 9]>
+      <script src="/vendor/html5-shiv/html5shiv.js"></script>
+    <![endif]-->
+  {% else %}
+    <!--[if !IE 8]><!-->
+      <link rel="stylesheet" href="/public/app.css">
+      <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/prism/1.6.0/themes/prism.min.css" integrity="sha256-vtR0hSWRc3Tb26iuN2oZHt3KRUomwTufNIf5/4oeCyg=" crossorigin="anonymous" />
+      <script src="https://cdnjs.cloudflare.com/ajax/libs/prism/1.6.0/prism.min.js" integrity="sha256-Zb9yKJ/cfs+jG/zIOFL0QEuXr2CDKF7FR5YBJY3No+c=" crossorigin="anonymous"></script>
```
hannalaakso: These two script tags are repeated. I wonder if we can always include them regardless of `legacy`?

hannalaakso: Looks like my comment had already been addressed

PR #1282, opened 2019-04-23
Add views that use govuk-frontend and legacy elements/template

hannalaakso:
This PR
- adds start page view which extends govuk_template
- adds applicant details view which extends govuk_template
- exposes `node_modules` to review app for the purposes of extending govuk_template in `app/views`
- fixes template assets path 
- sets assets path for legacy scss

Part of https://github.com/alphagov/govuk-frontend/issues/1262


nickcolley:
I think it might be useful to be able to differentiate these from regular views in the title without having to read the source code.

PR #1207, opened 2019-02-20
Add 'secondary' and 'warning' action buttons

dashouse:
This PR:
- Adds a `tint` and `shade` mix function - This is more controllable than `darken` and `lighten`
- Converts existing button to use `shade` (Therefore a slight change to the hover colour)
- Adds a secondary button modifier
- Adds a warning button modifier

**Secondary**
I need to consult with this design regarding WCAG 2.1 non-text contrast. The text and the shadow meet contrast which I believe will be enough (the background of the button does not).
screen shot 2019-02-20 at 08 46 47

**Warning**
screen shot 2019-02-20 at 08 47 01

PR #1283, opened 2019-04-24
Document button changes in CHANGELOG

36degrees:
This was missed in #1207

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,15 @@
 
 🆕 New features:
 
+- Add new govuk-shade and govuk-tint functions for creating shades and tints of
+  colours.
+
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,15 @@
...
 🆕 New features:
 
+- Add new govuk-shade and govuk-tint functions for creating shades and tints of
+  colours.
+
+  ([PR #1207](https://github.com/alphagov/govuk-frontend/pull/1207))
+
+- Add new secondary and warning button variants
+
+  ([PR #1207](https://github.com/alphagov/govuk-frontend/pull/1207))
```
aliuk2012: Do we need to separate entries?

36degrees: Whilst one relies on the other, I think they're two distinct things that should be called out separately?

Issue #1240, opened 2019-03-06
The $govuk-border-width-mobile variable doesn't make any sense


36degrees:
We have a variable called `$govuk-border-width-mobile`.

It's documented as 'Border width on mobile'

However, it's only used in a couple of places:

- The width of the conditional reveal on checkboxes and radio buttons **at all breakpoints**
- The width of the error summary **at all breakpoints**

We should consider renaming or removing this variable.

dashouse:
The 4px border is used in the conditional reveals because it has to be an even number in order to be centred under the 40px checkbox or radio. 5px would put it 1px too far left or right and it was noticeable.

The error summary change is probably not needed.

So I'd say we will need a 4px border, it should just have a different name.

PR #1287, opened 2019-04-24
Rename border-width-mobile to reflect how it's used

hannalaakso:
Despite its current name, the variable isn't used on mobile.

Standardise the border width on error summary on mobile to use the standard border width.

From @dashouse)
> The 4px border is used in the conditional reveals because it has to be an even number in order to be centred under the 40px checkbox or radio. 5px would put it 1px too far left or right and it was
noticeable. The error summary change is probably not needed.

Fixes https://github.com/alphagov/govuk-frontend/issues/1240

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,14 @@
 # Changelog
 
-## Unreleased
+## 3.0.0 (Breaking release)
 
 💥 Breaking changes:
+
+- Rename `$govuk-border-width-mobile` to `$govuk-border-width-narrow`
+
+  To migrate: If you are using `$govuk-border-width-mobile` in your own custom code, you need to rename any instances to `$govuk-border-width-narrow`.
+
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,16 +1,18 @@
...
 💥 Breaking changes:
 
-- Pull Request Title goes here
+- Rename $govuk-border-width-mobile to $govuk-border-width-narrow
 
-  Description goes here (optional)
+  This better reflects how the variable is used.
 
-  To migrate you need to change: X
+  This PR also makes the error summary border the standard width on mobile.
```
nickcolley: Non-blocking in future could split the 'fix' out into a separate changelog entry

hannalaakso: Thanks @nickcolley I've split this into breaking change and fix 👍

PR #1288, opened 2019-04-24
Update colour palette

36degrees:
This updates GOV.UK Frontend to use the new palette, as designed by the GOV.UK graphic design community.

This is part of a broader series of work which will be released as part of v3.0, which also includes changes to the hover and focus states of many components.

## Changes to the palette

| Colour        | Before    | After     | Notes                                           |
|---------------|-----------|-----------|-------------------------------------------------|
| purple        | `#2e358b` | `#4c2c92` | updated to match visited link colour            |
| light-purple  | `#6f72af` | `#6f72af` |                                                 |
| bright-purple | `#912b88` | `#912b88` |                                                 |
| pink          | `#d53880` | `#d53880` |                                                 |
| light-pink    | `#f499be` | `#f499be` |                                                 |
| red           | `#b10e1e` | `#d4351c` | updated                                         |
| bright-red    | `#df3034` |           | removed, as it was very similar to the new red  |
| orange        | `#f47738` | `#f47738` |                                                 |
| brown         | `#b58840` | `#b58840` |                                                 |
| yellow        | `#ffbf47` | `#ffdd00` | updated                                         |
| light-green   | `#85994b` | `#85994b` |                                                 |
| green         | `#006435` | `#00703c` | updated                                         |
| turquoise     | `#28a197` | `#28a197` |                                                 |
| light-blue    | `#2b8cc4` | `#5694ca` | updated                                         |
| blue          | `#005ea5` | `#1d70b8` | updated                                         |
| dark-blue     |           | `#003078` | added                                           |
| black         | `#0b0c0c` | `#0b0c0c` |                                                 |
| grey-1        | `#6f777b` |           | removed                                         |
| dark-grey     |           | `#6f777b` | added                                           |
| grey-2        | `#bfc1c3` |           | removed                                         |
| mid-grey      |           | `#b1b4b6` | added                                           |
| grey-3        | `#dee0e2` |           | removed                                         |
| grey-4        | `#f8f8f8` |           | removed                                         |
| light-grey    |           | `#f3f2f1` | added                                           |
| white         | `#ffffff` | `#ffffff` |                                                 |

## Changes to components and styles

It also makes changes to the way that some of these colours are used, including:

- the button component now uses the green from the colour palette, instead of a custom green previously only used for the button
- the confirmation panel now uses a green background rather than a turquoise background
- links now get _darker_ when hovered, rather than lighter
- visited links remain the same colour, but the purple in the palette has been updated to match

## Maintaining compatibility with GOV.UK Template, Frontend Toolkit and Elements

When GOV.UK Frontend is run in 'compatibility mode', the legacy palette will be used and components will continue to use the existing colours (for example, the panel will continue to be turquoise).

This is so that services can continue to adopt GOV.UK Frontend gradually, without having a mix of the two different palettes appearing within a service or a single page.

To support this:

- a new setting `$govuk-use-legacy-palette` has been added, which by default will be `true` if any of `$govuk-compatibility-govukfrontendtoolkit`, `$govuk-compatibility-govuktemplate` or `$govuk-compatibility-govukelements` are `true`.
- the `govuk-colour` function has been updated to add a `$legacy` argument, which specifies the colour (either a literal, or a name of a colour from the legacy palette) to use when `$govuk-use-legacy-palette` is `true`.
- in some places we are conditionally setting variables based on the value of `$govuk-use-legacy-palette`

Any new components that are implemented whilst this arrangement is in place will need to be designed for both contexts.

You can see how 'legacy mode' looks in the review app by appending `?legacy=1` to the URL:
https://govuk-frontend-review-pr-1288.herokuapp.com/?legacy=1

### DIFF
--- a/src/settings/_colours-palette.scss
+++ b/src/settings/_colours-palette.scss
@@ -61,6 +61,7 @@ $_govuk-colour-palette-modern: (
   "blue": #1d70b8,
   "dark-blue": #003078,
   "light-blue": #5694ca,
+  "purple": #4c2c92,
 
   "black": #0b0c0c,
   "dark-grey": #6f777b,
@@ -68,7 +69,6 @@ $_govuk-colour-palette-modern: (
   "light-grey": #f3f2f1,
   "white": #ffffff,
 
-  "purple": #2e358b,
   "light-purple": #6f72af,
   "bright-purple": #912b88,
   "pink": #d53880,
--- a/src/settings/_colours-palette.scss
+++ b/src/settings/_colours-palette.scss
@@ -11,7 +11,11 @@
 /// @type Boolean
 /// @access public
 
-$govuk-use-legacy-palette: true !default;
+$govuk-use-legacy-palette: if((
+    $govuk-compatibility-govukfrontendtoolkit or
+    $govuk-compatibility-govuktemplate or
+    $govuk-compatibility-govukelements
+  ), true, false) !default;
 
 /// Legacy colour palette
 ///
--- a/src/settings/_colours-palette.scss
+++ b/src/settings/_colours-palette.scss
@@ -73,7 +73,6 @@ $_govuk-colour-palette-modern: (
   "bright-purple": #912b88,
   "pink": #d53880,
   "light-pink": #f499be,
-  "bright-red": #df3034,
   "orange": #f47738,
   "brown": #b58840,
   "light-green": #85994b,
--- a/src/settings/_colours-palette.scss
+++ b/src/settings/_colours-palette.scss
@@ -2,16 +2,26 @@
 /// @group settings/colours
 ////
 
-/// Colour palette
+/// Use 'legacy' colour palette
 ///
-/// @type Map
+/// Whether or not to use the colour palette from GOV.UK Elements / Frontend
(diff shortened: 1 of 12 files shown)


src/settings/_colours-palette.scss
```diff
@@ -34,4 +48,52 @@ $govuk-colours: (
...
+  "white": #ffffff,
+
+  "light-purple": #6f72af,
+  "bright-purple": #912b88,
+  "pink": #d53880,
+  "light-pink": #f499be,
+  "orange": #f47738,
+  "brown": #b58840,
+  "light-green": #85994b,
+  "turquoise": #28a197
```
aliuk2012: I thought Turquoise was being dropped in the new palette. https://github.com/alphagov/govuk-frontend/pull/1245/files#diff-76794ff1589bb2e535aa6b43c4902fb9L27

36degrees: After discussion with @dashouse, we're going to keep most of the 'secondary' colours in the palette as they're useful for teams doing e.g. data visualisation that need a broader palette to work with.

36degrees:
Rebased against master and changelog entry added.

@nickcolley @aliuk2012 thanks for the review – let me know if you're happy with it 👍

PR #1266, opened 2019-04-01
Ability to group radio buttons / checkboxes within sub-headings

colinrotherham:
This feature adds a new macro `govukFormGroup()`.

This lets us put radios/checkboxes inside a `<div class="govuk-form-group">` wrapper, nested to multiple levels with the `describedBy` variable to be persisted and appended to at each level.

It also allows the component examples to show both regular and fieldset-wrapped variations using the optional `params.fieldset`, useful to confirm bugs like https://github.com/alphagov/govuk-frontend/issues/1264

Form group

### Why else is it needed?
I've also addressed what looks like a bit of technical debt.

Duplicated inside almost all form control macros is Nunjucks code to generate a `govukLabel()` label, a `govukHint()` hint an a `govukErrorMessage()` error message.

This code here is pretty much copy and pasted 7 times:

```nunjucks
  {{ govukLabel({
    html: params.label.html,
    text: params.label.text,
    classes: params.label.classes,
    isPageHeading: params.label.isPageHeading,
    attributes: params.label.attributes,
    for: params.id
  }) | indent(2) | trim }}
{% if params.hint %}
  {% set hintId = params.id + '-hint' %}
  {% set describedBy = describedBy + ' ' + hintId if describedBy else hintId %}
  {{ govukHint({
    id: hintId,
    classes: params.hint.classes,
    attributes: params.hint.attributes,
    html: params.hint.html,
    text: params.hint.text
  }) | indent(2) | trim }}
{% endif %}
{% if params.errorMessage %}
  {% set errorId = params.id + '-error' %}
  {% set describedBy = describedBy + ' ' + errorId if describedBy else errorId %}
  {{ govukErrorMessage({
    id: errorId,
    classes: params.errorMessage.classes,
    attributes: params.errorMessage.attributes,
    html: params.errorMessage.html,
    text: params.errorMessage.text,
    visuallyHiddenText: params.errorMessage.visuallyHiddenText
  }) | indent(2) | trim }}
{% endif %}
```

It's used by textarea, select, radios, checkboxes, text input, file upload and date input.

All this boilerplate code is now handled by `govukFormGroup()`.

PR #1297, opened 2019-04-25
Trigger setAttributes on radio inputs with shared wrapper

colinrotherham:
This change allows radio input events (for conditional reveals) to be scoped higher to parent `.govuk-form-group` wrappers, allowing groups with multiple lists to behave correctly.

Originally proposed by @frankieroberto in https://github.com/alphagov/govuk-frontend/pull/1163 and has been separated from https://github.com/alphagov/govuk-frontend/pull/1266.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,12 @@ See the [versioning documentation for how to update this changelog](./docs/contr
 
   ([PR #1347](https://github.com/alphagov/govuk-frontend/pull/1347))
 
+- Trigger setAttributes on radio inputs with shared wrapper
+
+  This change allows radio input events (for conditional reveals) to be scoped higher to parent `.govuk-form-group` wrappers, allowing groups with multiple lists to behave correctly.
+
+  ([PR #1297](https://github.com/alphagov/govuk-frontend/pull/1297))
+
 🔧 Fixes:
 
 - Update colour for MHCLG
--- a/src/components/radios/radios.js
+++ b/src/components/radios/radios.js
@@ -1,15 +1,23 @@
 import '../../vendor/polyfills/Function/prototype/bind'
 import '../../vendor/polyfills/Event' // addEventListener and event.target normaliziation
 import '../../vendor/polyfills/Element/prototype/classList'
+import '../../vendor/polyfills/Element/prototype/closest'
 import { nodeListForEach } from '../../common'
 
 function Radios ($module) {
   this.$module = $module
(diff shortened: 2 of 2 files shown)


nickcolley:
Hey Colin,

We are currently prioritising work to make[GOV.UK Frontend pass WCAG 2.1 requirements](https://github.com/alphagov/govuk-design-system/issues/677) and we have an external audit soon.

This work looks really well done, we want to discuss it as a team and make sure it's the right approach, and then give some proper feedback.

We're not going to be able to follow up on this until we've done the audit so thanks a lot for your patience. :)

(Sorry for the copy paste :))

Nick

nickcolley:
Hey Colin,

We've discussed this as a team, thanks for waiting.

Have you considered instead of looking at a parent class, to rely on inputs with the same `name` attribute? This would allow us to keep the logic within the component itself without relying on something that might not exist.

Nick

colinrotherham:
Hi @nickcolley, yeah I did but noticed the original didn't do that?

Trying to also apply this feedback: https://github.com/alphagov/govuk-frontend/pull/1163#issuecomment-463223512

Can you suggest an approach the team would be happiest with?

nickcolley:
Colin and I spoke about this and thought if we could instead look for inputs with the same name that were scoped within a radios component.

Because there's no way (yet https://github.com/alphagov/govuk-frontend/issues/1127) of knowing what elements are initialised (data-module is optional), I think Colin's current approach might be the best way to do it without drastically changing how this feature works. (For example, decoupling conditional hide/show logic from the component).

nickcolley:
Hey @colinrotherham !

Firstly, thanks for your patience with us taking so long on this pull request.

I sat down with Ollie and we think we have an alternative that will fix this behaviour without relying on the `.govuk-form-group` parent.

https://github.com/alphagov/govuk-frontend/compare/radio-trigger-setattributes

If you think this approach is good, we intend on optimising it to only run code on inputs with the same name, and within the same form owner.

Let me know what you think.

frankieroberto:
@nickcolley sounds good to me. Querying on the same name attribute and the same form parent sounds like a reasonable thing to do too.

colinrotherham:
@nickcolley Yep this is great, using `document.querySelector()` was @frankieroberto's plan originally. I'm happy if you start with this and narrow it down a little. Glad we found a way, thanks.

nickcolley:
@frankieroberto @colinrotherham yeah, we've gone round in circles a bit trying to avoid making this a more global change but it seems like the right approach in the end, thanks a lot for your patience again.

Issue #1302, opened 2019-04-30
Update links (and things that look like links) to use the new focus style


36degrees:
## What

Update links, and things that look like links, to use the new focus style as per the [preview branch](https://github.com/alphagov/govuk-frontend/pull/1245).

That includes:

- Links (including text and muted variants)
- Back link
- Details
- Links within the error summary
- Skip link

## Why

We're updating the focus style as part of [making the Design System meet WCAG 2.1](https://github.com/alphagov/govuk-design-system/issues/677) – specifically with regard to WCAG 2.1 1.4.11 Non-Text Contrast (#1137).

## Done when…

- [ ] Links updated to use new focus state
- [ ] Back link updated to use new focus state
- [ ] Details updated to use new focus state
- [ ] Error summary links updated to use new focus state
- [ ] Skip link updated to use new focus state
- [ ] Tested in all [browsers that we support](https://github.com/alphagov/govuk-frontend#browser-support)
- [ ] Tested with zoom up to 400%
- [ ] Tested with ZoomText
- [ ] Tested in Windows High Contrast mode
- [ ] Tested with overridden colours in the browser

PR #1309, opened 2019-04-30
Update links (and things that look like links) to use the new focus style

nickcolley:
Related issue: https://github.com/alphagov/govuk-frontend/issues/1302

### DIFF
--- a/src/components/details/_details.scss
+++ b/src/components/details/_details.scss
@@ -13,6 +13,7 @@
   }
 
   .govuk-details__summary {
+    @include govuk-focusable-text-link;
     // Make the focus outline shrink-wrap the text content of the summary
     display: inline-block;
 
@@ -34,18 +35,13 @@
     text-decoration: underline;
   }
 
-  .govuk-details__summary:hover {
-    color: $govuk-link-hover-colour;
+  // Remove the underline when focussed to avoid duplicate borders
+  .govuk-details__summary:focus .govuk-details__summary-text {
+    text-decoration: none;
   }
 
-  .govuk-details__summary:focus {
-    // -1px offset fixes gap between background and outline in Firefox
-    outline: ($govuk-focus-width + 1px) solid $govuk-focus-colour;
-    outline-offset: -1px;
-    // When focussed, the text colour needs to be darker to ensure that colour
-    // contrast is still acceptable
-    color: $govuk-focus-text-colour;
-    background: $govuk-focus-colour;
+  .govuk-details__summary:hover {
+    color: $govuk-link-hover-colour;
   }
 
   // Remove the default details marker so we can style our own consistently and
@@ -59,7 +55,7 @@
     content: "";
     position: absolute;
 
-    top: 0;
+    top: -1px;
     bottom: 0;
     left: 0;
 
--- a/src/helpers/_focusable.scss
+++ b/src/helpers/_focusable.scss
@@ -32,3 +32,31 @@
     background-color: $govuk-focus-colour;
   }
 }
+
+/// Focusable with box-shadow
+///
+/// Removes the visible outline and replace with box-shadow and background colour.
+/// Used for interactive text-based elements.
+///
+/// @access public
+
+@mixin govuk-focusable-text {
+  // When colours are overridden, for example when users have a dark mode,
+  // backgrounds and box-shadows disappear, so we need to ensure there's a
+  // transparent outline which will be set to a visible colour.
+  outline: $govuk-focus-width solid transparent;
+  outline-offset: 0;
+  color: $govuk-text-colour;
+  background-color: $govuk-focus-colour;
+  box-shadow:
+  -5px -1px 0 1px govuk-colour("yellow"),
+  5px -1px 0 1px govuk-colour("yellow"),
+  -3px 1px 0 3px govuk-colour("black"),
+  3px 1px 0 3px govuk-colour("black");
+}
+
+@mixin govuk-focusable-text-link {
+  &:focus {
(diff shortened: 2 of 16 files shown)


src/components/details/_details.scss
```diff
@@ -34,18 +35,13 @@
...
-  .govuk-details__summary:focus {
-    // -1px offset fixes gap between background and outline in Firefox
-    outline: ($govuk-focus-width + 1px) solid $govuk-focus-colour;
-    outline-offset: -1px;
-    // When focussed, the text colour needs to be darker to ensure that colour
-    // contrast is still acceptable
-    color: $govuk-focus-text-colour;
-    background: $govuk-focus-colour;
+  .govuk-details__summary:hover {
+    color: $govuk-link-hover-colour;
```
36degrees: This is affecting focus styles too, which I don't believe it should be.

Screenshot 2019-05-03 at 09 35 16BST

src/helpers/_links.scss
```diff
@@ -10,7 +10,7 @@
 
 @mixin govuk-link-common {
   @include govuk-typography-common;
-  @include govuk-focusable-fill;
+  @include govuk-focusable-text-link;
```
36degrees: ~~Links don't seem to be getting the focus text colour when in compatibility mode – happening on master as well so I don't think it was fixed by #1310 either?~~

Screenshot 2019-05-03 at 09 41 10BST

Realised this was fixed by #1310, but on the v3 branch – can we rebase against `v3` to bring that in now it's been merged?

src/helpers/_focusable.scss
```diff
@@ -32,3 +32,38 @@
...
+  // transparent outline which will be set to a visible colour.
+
+  // Since Internet Explorer 8 does not support box-shadow, we want to force the user-agent outlines
+  @include govuk-not-ie8 {
+    outline: $govuk-focus-width solid transparent;
+    outline-offset: 0;
+  }
+  color: $govuk-text-colour;
+  background-color: $govuk-focus-colour;
+  box-shadow:
```
36degrees: Very much personal preference, but I'd probably disable the linter for this line and align the values:

```
box-shadow: -5px -1px 0 1px govuk-colour("yellow"),
             5px -1px 0 1px govuk-colour("yellow"),
            -3px  1px 0 3px govuk-colour("black"),
             3px  1px 0 3px govuk-colour("black");
```

It feels like this could do with a comment as well, but I don't even know where to start trying to explain how all these shadows interact with each other… 🤪

src/helpers/_focusable.scss
```diff
@@ -32,3 +32,38 @@
...
+/// Used for interactive text-based elements.
+///
+/// @access public
+
+@mixin govuk-focusable-text {
+  // When colours are overridden, for example when users have a dark mode,
+  // backgrounds and box-shadows disappear, so we need to ensure there's a
+  // transparent outline which will be set to a visible colour.
+
+  // Since Internet Explorer 8 does not support box-shadow, we want to force the user-agent outlines
```
36degrees: Should we consider adding e.g. a black border all round for IE8? Or are we happy with just user agent defaults?

nickcolley: Adding a black border may make reading the text difficult, so the user-agent styles (screenshot) above seemed better. Happy to go with whatever people think is best.

src/helpers/_focusable.scss
nickcolley: Yeah lets do this

src/helpers/_focusable.scss
```diff
@@ -32,3 +32,39 @@
...
+              5px -1px 0 1px govuk-colour("yellow"),
+              -3px 1px 0 3px govuk-colour("black"),
+              3px 1px 0 3px govuk-colour("black");
+  // sass-lint:enable indentation
+  // When link is focussed, hide the default underline since the
+  // box shadow adds the "underline"
+  text-decoration: none;
+}
+
+@mixin govuk-focusable-text-link {
```
hannalaakso: This mixin seems to be performing a task very similar to `@mixin govuk-focusable-fill`. Do we need both? It would be good to have a comment to indicate what the difference between the two is.

nickcolley: As discussed on Slack we're going to wait until we have everything merged before deciding how the mixins should look.

src/components/skip-link/_skip-link.scss
```diff
@@ -5,7 +5,8 @@
 @include govuk-exports("govuk/component/skip-link") {
   .govuk-skip-link {
     @include govuk-visually-hidden-focusable;
-    @include govuk-link-common;
+    @include govuk-typography-common;
+    @include govuk-focusable-fill;
```
aliuk2012: Why is this using govuk-focusable-fill and why is the skip link styled differently to other links?

nickcolley: This is a design decision from @dashouse to not use the heavy underline focus state for skip-link.

aliuk2012:
Looks good from a code POV, I'm now going to run through a quick xbrowser test before approving

src/components/skip-link/_skip-link.scss
aliuk2012: 🤔 ok, currently this component and Footer are the only two components still using `govuk-focusable-fill` mixin. When we come to review the public mixing perhaps we can just hard code the mixins output here instead of having it hanging around

nickcolley: Yeah once we've got everything production ready we can (after the audit potentially) decide on how the mixins should work, with the new guidance in mind...

PR #1310, opened 2019-05-01
Fixes an issue with styling links

aliuk2012:
When visiting the links example page, if you had previously visited the
home page then the unvisited link and visited link example looked identical.
This commit passes a random number to the view. The view adds the randomId
to the href as a hash.

👉 https://govuk-frontend-review-pr-1310.herokuapp.com/examples/links

# Before
image

# After

image

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -177,7 +177,9 @@ module.exports = (options) => {
 
   // Example view
   app.get('/examples/:example', function (req, res, next) {
-    res.render(`${req.params.example}/index`, function (error, html) {
+    // Passing a random number used for the links so that they will be unique and not display as "visited"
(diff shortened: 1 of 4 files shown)


app/app.js
```diff
@@ -177,7 +177,9 @@ module.exports = (options) => {
 
   // Example view
   app.get('/examples/:example', function (req, res, next) {
-    res.render(`${req.params.example}/index`, function (error, html) {
+    // Passing a random number used for the links so that they will be unique and not display as "visited"
+    const randomId = (Math.random() * 1000000).toFixed()
```
hannalaakso: Would `randomPageid` or `unvisitedPageId`  be more descriptive?

PR #1312, opened 2019-05-01
Update form inputs focus to comply with WCAG 2.1 

hannalaakso:
Iterate on https://github.com/alphagov/govuk-frontend/pull/1245 to update the GOV.UK Frontend form inputs.

Affected components:
- Text input
- Textarea
- Select
- File upload - WIP as still working on getting the spacing work both with and without error styles applied 

Method:
- Double the border width on focus using `box-shadow`.
- Cater for IE8 which doesn't support `box-shadow` by setting `border-width`.
- The component error style already has a double width border so don’t apply `box-shadow` on focus there.

Closes: #1302


src/components/file-upload/_file-upload.scss
```diff
@@ -7,13 +7,66 @@
...
+    // As `upload--error` has border, it needs to have the same padding as
+    // the standard focused element.
+    margin-right: -$spacing-unit;
+    margin-left: -$spacing-unit;
+    padding-right: $spacing-unit;
+    padding-left: $spacing-unit;
     border: $govuk-border-width-form-element-error solid $govuk-error-colour;
+
+    &:focus {
+      border-color: $govuk-input-border-colour;
```
36degrees: I don't think this is working quite as intended in firefox, I suspect because of the `:focus-within` hack.

In Chrome the file input, in error state, when focused, looks like this:

govuk-frontend-review-pr-1312 herokuapp com_components_file-upload_with-error-message_preview (3)

This is consistent with the other form input types – when in an error state, the red border is _replaced_ with a black border.

But in Firefox, you always end up with a 'triple ring' like this:

Screenshot_2019-05-03 GOV UK Frontend(1)

You also see the triple border in Chrome if you click on the 'Choose file' button (I assume because the button is focused, so this counts as 'focus within' the input.

govuk-frontend-review-pr-1312 herokuapp com_components_file-upload_with-error-message_preview (4)

src/components/file-upload/_file-upload.scss
```diff
@@ -7,13 +7,66 @@
 @import "../label/label";
 
 @include govuk-exports("govuk/component/file-upload") {
+  $spacing-unit: govuk-spacing(1);
```
36degrees: Can we think of a more descriptive name for this?

src/components/input/_input.scss
```diff
@@ -24,6 +24,19 @@
 
     // Disable inner shadow and remove rounded corners
     appearance: none;
+
+    &:focus {
+      // Double the border by adding its width again. Use `box-shadow` to do
+      // this instead of changing `border-width` (which changes element size) and
```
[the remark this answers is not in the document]
dashouse: When I was testing the design branch with overridden colours it changed in a meaningful way as the focus state for this is using a visible outline so the yellow turns to black

focus-override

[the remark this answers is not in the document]
hannalaakso: Thanks @36degrees I've updated the comment to more accurately explain what's going on.

I think this branch matches the focus state on @dashouse's branch when colours are overridden 

nameinput

So there is a visible change. @36degrees what do you think?

src/components/file-upload/_file-upload.scss
hannalaakso: Fixed. I'm adding the same resets to `:focus-within` as to `:focus` to prevent the error styles from being applied when the focus is within.

src/components/file-upload/_file-upload.scss
hannalaakso: Updated!

Issue #1291, opened 2019-04-25
Add ability to override or remove OGL logo in Footer


aliuk2012:
The footer hardcodes the Open Government Licence text. It should be possible for the user to change the text or even remove the entire thing.

User need explained here: https://github.com/alphagov/govuk_template/issues/188

Split from: https://github.com/alphagov/govuk-frontend/issues/971

Part of https://github.com/alphagov/govuk-design-system/issues/777

aliuk2012:
Needed as part of https://github.com/alphagov/govuk-design-system-backlog/issues/185

HughePaul:
This is needed for Welsh translations, which is required for a site to go into public beta.

Issue #1290, opened 2019-04-25
Add ability to override or remove Crown Copyright in Footer


aliuk2012:
At the moment `© Crown copyright` in Footer is hardcoded text. User should be able to override this or remove it altogether.

Split from: https://github.com/alphagov/govuk-frontend/issues/971

Part of : https://github.com/alphagov/govuk-frontend/issues/971

aliuk2012:
Needed as part of https://github.com/alphagov/govuk-design-system-backlog/issues/185

HughePaul:
This is needed for Welsh translations, which is required for a site to go into public beta.

matthewmascord:
Hello, is there an ETA for this? Considering whether it's worth waiting or implementing our own fork of govukFooter.

36degrees:
You can now override the 'crown copyright' text used alongside the crest in the footer, but you still can't remove it entirely.

ch-ncolley:
I believe that this can be closed in a future release once this feature is released:

https://github.com/alphagov/govuk-frontend/pull/6191

romaricpascal:
@ch-ncolley Thanks for bringing that to our attention. I think the issue will remain relevant after your PR is merged, as if I understand correctly, it's about the crest and the `Crown Copyright` link at the very end of the footer, rather than the content licence part.

I'll double check with the team if we want to keep this open as we've made a point that we were not [keen to allow the link's removal in this PR](https://github.com/alphagov/govuk-frontend/pull/3876).

ch-ncolley:
Ah my mistake sorry!

PR #1314, opened 2019-05-02
Update footer to take licence text and copyright text

georgeeker:


georgeeker:
See: 
https://github.com/alphagov/govuk-frontend/issues/1291
https://github.com/alphagov/govuk-frontend/issues/1290

src/components/footer/footer.yaml
```diff
@@ -125,6 +147,16 @@ examples:
     meta:
       text: GOV.UK Prototype Kit v7.0.1
 
+- name: with custom licence text
+  description: Custom licence section
+  data:
+    licence: GOVUK Licence
```
colinrotherham: Hi @georgeeker, this will need to have a `.text` or `.html` property to work.

See the preview: https://govuk-frontend-review-pr-1314.herokuapp.com/components/footer/with-custom-licence-text/preview

It's not triggering `{% if params.licence.html or params.licence.text %}`

src/components/footer/footer.yaml
```diff
@@ -125,6 +147,16 @@ examples:
...
+
+- name: with custom copyright text
+  description: Custom copyright section
+  data:
+    logoText: Copyright
```
colinrotherham: Can this be added to Nunjucks template too?

36degrees:
There's a couple of things that still need thinking about here.

The OGL logo is included as an SVG as part of the footer. With the change as it currently stands, even if you override the copyright text you still get the OGL logo, because it's outside of the `{% if %}` block:

localhost_3000_components_footer_with-custom-licence-text_preview

We could move the `{% if %}` block to wrap the SVG as well, but:

- this would mean that users overriding the HTML would need to know to wrap their licence in `govuk-footer__licence-description`
- I think there may be instances where users want to overwrite the copyright text whilst still preserving the OGL logo

Secondly, removing the crest makes the footer a lot smaller, so I think we probably need some design input as to whether that's an issue or not.

localhost_3000_components_footer_with-custom-copyright-text_preview

There _might_ also again be cases where a user wants to override the copyright text without removing the crest, although it's less of an issue as they only have to include a specific class name on the link, rather than having to redeclare an entire SVG.

36degrees:
Hi George,

Thanks for your work on this so far. I just wanted to check whether you were interested in carrying on working on this?

If we don't hear from you in the next 7 days then we'll close this PR (keeping things tidy helps us keep track of what's being worked on) but please feel free to re-open if you do want to pick this up again.

Addressing this is something we plan to do, but we've not got capacity to pick up ourselves right now.

Thanks,

Ollie

36degrees:
As mentioned in the last comment we're going to close this for now, but please feel free to re-open if you want to pick this up again.

Thanks again for your contribution.

PR #1315, opened 2019-05-03
Add new focus style to buttons

### LINKED ISSUE #1305: Update buttons to use the new styles and the new focus state
## What

Update buttons to use the new focus style as per the [preview branch](https://github.com/alphagov/govuk-frontend/pull/1245).

That includes:

- Default buttons
- Start buttons
- Secondary buttons
- Warning buttons

All as both buttons and links.

## Why

We're updating the focus style as part of [making the Design System meet WCAG 2.1](https://github.com/alphagov/govuk-design-system/issues/677) – specifically with regard to WCAG 2.1 1.4.11 Non-Text Contrast (#1137).

## Done when…

- [x] Buttons updated to use the new focus state
- [x] Tested in all [browsers that we support](https://github.com/alphagov/govuk-frontend#browser-support)
- [x] Tested in legacy mode (using the deprecated libraries)
- [x] Tested with zoom up to 400%
- [x] Tested with ZoomText
- [ ] Tested in Windows High Contrast mode
- [x] Tested with overridden colours in the browser

aliuk2012:
### Before:
image

### After: 
image

### Browser testing:

- [x] Chrome 73 & 74
- [x] Firefox 65 & 66
- [x] OS Safari 10, 11, 12
- [x] Android Samsung Galaxy (Chrome)
- [x] iOS XS / 8 (Safari) 
- [x] Edge 17 & 18
- [x] IE 8 - 11

### Others tests:
- [x] Zoomed at 400%
- [x] Legacy mode - focus is darker colour. If a button doesn't use govuk-button class then the focus style is not updated to the new style. For example https://govuk-frontend-review-pr-1315.herokuapp.com/full-page-examples/renew-driving-licence

### Issues:
- Start button icon is still a two toned white png. 

👉 **Before:** https://govuk-frontend-review.herokuapp.com/components/button
👉 **After:** https://govuk-frontend-review-pr-1315.herokuapp.com/components/button
Closes #1305

### DIFF
--- a/src/components/button/_button.scss
+++ b/src/components/button/_button.scss
@@ -27,7 +27,6 @@
 
   .govuk-button {
     @include govuk-font($size: 19, $line-height: 19px);
-    @include govuk-focusable;
 
     box-sizing: border-box;
     display: inline-block;
@@ -63,35 +62,46 @@
       text-decoration: none;
     }
 
-    // alphagov/govuk_template includes a specific a:link:focus selector
-    // designed to make unvisited links a slightly darker blue when focussed, so
-    // we need to override the text colour for that combination of selectors so
-    // so that unvisited links styled as buttons do not end up with dark blue
-    // text when focussed.
-    @include govuk-compatibility(govuk_template) {
-      &:link:focus {
-        color: $govuk-button-text-colour;
-      }
-    }
-
     // Fix unwanted button padding in Firefox
     &::-moz-focus-inner {
       padding: 0;
       border: 0;
     }
 
-    &:hover,
-    &:focus {
+    &:hover {
       background-color: $govuk-button-hover-colour;
     }
 
     &:active {
+      // Bump the button down so it looks like its being pressed in
       top: $button-shadow-size;
-      box-shadow: none;
 
       @include govuk-if-ie8 {
         border-bottom-width: 0;
       }
+    }
+
+    &:focus {
+      border-color: $govuk-focus-colour;
+      // When colours are overridden, for example when users have a dark mode,
+      // backgrounds and box-shadows disappear, so we need to ensure there's a
+      // transparent outline which will be set to a visible colour.
+      @include govuk-not-ie8 {
+        outline: $govuk-focus-width solid transparent;
+        outline-offset: 0;
+      }
+      @include govuk-if-ie8 {
+        color: $govuk-text-colour;
+        background-color: $govuk-focus-colour;
+      }
+      box-shadow: inset 0 0 0 1px $govuk-focus-colour;
+    }
+
+    &:focus:not(:active):not(:hover) {
+      border-color: $govuk-focus-colour;
+      color: $govuk-text-colour;
+      background-color: $govuk-focus-colour;
+      box-shadow: 0 2px 0 $govuk-focus-text-colour;
     }
 
     // The following adjustments do not work for <input type="button"> as
@@ -175,8 +185,7 @@
       }
     }
 
-    &:hover,
-    &:focus {
+    &:hover {
       background-color: $govuk-secondary-button-hover-colour;
     }
   }
@@ -204,8 +213,7 @@
       }
     }
 
-    &:hover,
-    &:focus {
+    &:hover {
       background-color: $govuk-warning-button-hover-colour;
     }
   }
--- a/src/components/button/_button.scss
+++ b/src/components/button/_button.scss
@@ -86,10 +86,12 @@
       // When colours are overridden, for example when users have a dark mode,
       // backgrounds and box-shadows disappear, so we need to ensure there's a
       // transparent outline which will be set to a visible colour.
+      // Since Internet Explorer 8 does not support box-shadow, we want to force the user-agent outlines
       @include govuk-not-ie8 {
         outline: $govuk-focus-width solid transparent;
         outline-offset: 0;
       }
+      // Since Internet Explorer does not support `:not()` we set a clearer focus style to match user-agent outlines.
       @include govuk-if-ie8 {
         color: $govuk-text-colour;
         background-color: $govuk-focus-colour;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,12 @@
 
   ([PR #1309](https://github.com/alphagov/govuk-frontend/pull/1309))
 
+- Update buttons to use new focus style
+
+  To migrate: [TODO add migration instructions before we ship v3.0.0]
+
+  ([PR #1315](https://github.com/alphagov/govuk-frontend/pull/1315))
+
 - Rename `$govuk-border-width-mobile` to `$govuk-border-width-narrow`
 
   To migrate: If you are using `$govuk-border-width-mobile` in your own custom code, you need to rename any instances to `$govuk-border-width-narrow`.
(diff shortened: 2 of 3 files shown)


src/components/button/_button.scss
```diff
@@ -80,14 +81,26 @@
...
     }
 
     &:active {
-      top: $button-shadow-size;
-      box-shadow: none;
+
+      border: $govuk-border-width-form-element solid govuk-colour("yellow");
+      // !important is used to overwrite the :focus !important colour
+      color: govuk-colour("white") !important;
+      background-color: $govuk-button-hover-colour !important;
```
nickcolley: Are you sure we need to put these !important statements in? I wonder if these pseudo selectors could be rearranged to avoid this.

aliuk2012: Good spot 👍

src/components/button/_button.scss
```diff
@@ -38,6 +38,7 @@
     padding: (govuk-spacing(2) - $govuk-border-width-form-element - ($button-shadow-size / 2)) govuk-spacing(2); // s1
     border: $govuk-border-width-form-element solid transparent;
     border-radius: 0;
+    outline: $govuk-focus-width solid transparent !important;
```
nickcolley: Do we need to put !important here?

src/components/button/_button.scss
```diff
@@ -80,14 +81,26 @@
       border: 0;
     }
 
-    &:hover,
-    &:focus {
+    // Ordering of this pseudo selector is important to ensure
+    // It replaces hover state when element is focussed and mouse is hovering
+    &:hover {
+      color: govuk-colour("white");
```
nickcolley: I believe that the white here is `$govuk-button-text-colour`, do we need to set it again?

I imagine if the focus styles were at the end we could avoid having to set the text colour again...

src/components/button/_button.scss
```diff
@@ -80,14 +81,26 @@
...
       background-color: $govuk-button-hover-colour;
     }
 
+    // Ordering of this pseudo selector is important to ensure
+    // It replaces hover state when element is focussed and mouse is hovering
+    &:focus {
+      color: govuk-colour("black");
+      background-color: govuk-colour("yellow");
```
nickcolley: We need to use the applied variables rather than the palette variables.

nickcolley:
Looking good so far, added some comments to make it more consistent with the rest of the codebase. 👍

dashouse:
Looking good so far

The secondary `:focus` state that is applied at the same time as `:active` is causing the button to get smaller. I believe I may have added some padding on the prototype, ideally the button will remain the same size in every state.

It also appears to have lost the margin-top that is applied in this state to give the button it's "pushed down" feel

36degrees:
I don't think this is working correctly for secondary or warning buttons at the minute. They're not getting the yellow highlight, and in the case of the warning buttons the text colour is changing to black (on red)

aliuk2012:
@36degrees Ah yes I see that, I think thats because I removed the important and that was forcing everything to work. I'll take a look at it as @nickcolley suggested perhaps moving the focus/hover state to the bottom

src/components/button/_button.scss
```diff
@@ -70,6 +70,7 @@
     // text when focussed.
     @include govuk-compatibility(govuk_template) {
       &:link:focus {
+        outline-color: transparent;
```
nickcolley: In a different part of the codebase I found I could just remove this compatibility stuff since it isnt needed anymore, might be the same here, worth checking.

src/components/button/_button.scss
```diff
@@ -80,14 +81,30 @@
       border: 0;
     }
 
-    &:hover,
-    &:focus {
+    // Ordering of this pseudo selector is important to ensure
```
nickcolley: Not sure we should have comments about the ordering, since it's just how CSS works?

nickcolley:
It looks like we can make the highlight border thicker to be consistent with Dave's original design:

Screen Shot 2019-05-09 at 16 00 05

Original:

Screen Shot 2019-05-09 at 16 00 27

src/components/button/_button.scss
```diff
@@ -204,10 +228,17 @@
...
 
-    &:hover,
-    &:focus {
+    &:hover {
       background-color: $govuk-warning-button-hover-colour;
     }
+
+    // Ordering of this pseudo selector is important to ensure
+    // It replaces hover state when element is focussed and mouse is hovering
+    &:focus {
```
nickcolley: Good work reordering the CSS to avoid specificity issues

nickcolley:
Buttons with modifiers have inconsistent focus state compared with the original.

Let me know if you wanna go over any of this feedback together.

aliuk2012:
It was decided that the final version of the buttons would not have a double thick border like links and the most noticeable changes would be the background colour changing to yellow and the bottom border colour changing to black

src/components/button/_button.scss
```diff
@@ -204,10 +221,15 @@
...
-    &:hover,
-    &:focus {
+    &:hover {
       background-color: $govuk-warning-button-hover-colour;
     }
+
+    &:focus {
+      color: $govuk-focus-text-colour;
+      background-color: $govuk-focus-colour;
+      box-shadow: 0 2px 0 $govuk-focus-text-colour;
```
hannalaakso: Similarly to above comment these are also inherited from `.govuk-button`

src/components/button/_button.scss
```diff
@@ -175,10 +187,15 @@
...
-    &:hover,
-    &:focus {
+    &:hover {
       background-color: $govuk-secondary-button-hover-colour;
     }
+
+    &:focus {
+      color: $govuk-focus-text-colour;
+      background-color: $govuk-focus-colour;
+      box-shadow: 0 2px 0 $govuk-focus-text-colour;
```
hannalaakso: It looks like these styles are inherited from .`govuk-button` so maybe we won't need to set them here.

src/components/button/_button.scss
aliuk2012: The reason there are duplicates is because the secondary buttons :hover state overrides `.govuk-button` focus because its further down the page.

src/components/button/_button.scss
aliuk2012: The reason there are duplicates is because the secondary buttons :hover state overrides `.govuk-button` focus because its further down the page.

src/components/button/_button.scss
hannalaakso: Thanks Al. It seems that they set the same styles so I don't think they need to be overridden?

Screen Shot 2019-05-10 at 15 16 41

hannalaakso: Thanks for the IRL @aliuk2012 🙌 -  so, if we don't set the focus states here, the hover state that's being set for secondary buttons overrides it if button is being hovered while focused.

Probably worth leaving a comment, maybe something like
```
Set .govuk-button :focus styles to avoid :hover styles overriding them if button is both focused and hovered.
```

We could also consider moving the `.govuk-button` focus styles to the bottom of `button.scss` to avoid having to rewrite styles but let's think about that later.

aliuk2012: I think we have a better solution now that we have a separate `:focus:hover` state. I have since removed this code in the last commit.

PR #1317, opened 2019-05-03
Allow banner to be hidden using query parameter

36degrees:
This is helpful for e.g. Browserstack screenshots

### DIFF
--- a/app/banner.js
+++ b/app/banner.js
@@ -6,6 +6,11 @@ module.exports = function (app) {
   // Detect if banner should be shown based on cookies set
   app.use(cookieParser())
   app.use(function (request, response, next) {
+    if ('hide-banner' in request.query) {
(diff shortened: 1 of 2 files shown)


app/banner.js
```diff
@@ -6,6 +6,11 @@ module.exports = function (app) {
   // Detect if banner should be shown based on cookies set
   app.use(cookieParser())
   app.use(function (request, response, next) {
+    if ('hide_banner' in request.query) {
```
nickcolley: would prefer kebab case but doesnt matter tbf

nickcolley: Should it set a cookie so you dont require it for subsequent navigations?

36degrees: Re cookies, I don't need that functionality for my use case (Browserstack) so reckon we keep it simple for now.

PR #1321, opened 2019-05-08
Update footer links to use new focus style

nickcolley:
## V3 focus styles

footer-new-focus-styles

## V3 focus styles with legacy mode turned on

footer-new-focus-styles-legacy

## Pre-v3 focus styles

footer-old-focus-styles

Closes https://github.com/alphagov/govuk-frontend/issues/1320

### DIFF
--- a/src/components/footer/_footer.scss
+++ b/src/components/footer/_footer.scss
@@ -44,7 +44,12 @@
   }
 
   .govuk-footer__link {
-    @include govuk-focusable-fill;
+    text-decoration: none;
+
+    &:hover,
+    &:active {
+      text-decoration: underline;
+    }
 
     @if ($govuk-use-legacy-palette) {
       &:link,
@@ -65,11 +70,7 @@
       }
     }
 
-    // When focussed, the text colour needs to be darker to ensure that colour
-    // contrast is still acceptable
-    &:focus {
-      color: $govuk-focus-text-colour;
-    }
+    @include govuk-focusable-text-link;
 
     // alphagov/govuk_template includes a specific a:link:focus selector
     // designed to make unvisited links a slightly darker blue when focussed, so
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,12 @@
 
   ([PR #1312](https://github.com/alphagov/govuk-frontend/pull/1312))
 
+- Update footer links to use new focus style
+
+  To migrate: [TODO add migration instructions before we ship v3.0.0]
+
+  ([PR #1321](https://github.com/alphagov/govuk-frontend/pull/1321))
+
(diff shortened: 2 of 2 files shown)


src/components/footer/_footer.scss
```diff
@@ -44,7 +44,12 @@
   }
 
   .govuk-footer__link {
-    @include govuk-focusable-fill;
+    text-decoration: none;
```
36degrees: @dashouse are you happy for this change to be made in 'compatibility mode' as well?

src/components/footer/_footer.scss
```diff
@@ -65,11 +70,7 @@
       }
     }
 
-    // When focussed, the text colour needs to be darker to ensure that colour
-    // contrast is still acceptable
-    &:focus {
-      color: $govuk-focus-text-colour;
-    }
+    @include govuk-focusable-text-link;
```
36degrees: Is the positioning of this important (coming after the pseudo-classes)? If so, might be worth a comment as I think we generally try and do our includes together at the top of the ruleset.

nickcolley: Yeah we tend to put focus styles at the end of the block, I would prefer to have the mixin only contain properties to avoid this confusion.

It's on my list of things to consider once we've done all the work and know what the mixins need to do...

src/components/footer/_footer.scss
nickcolley: Good shout, will see what @dashouse thinks.

dashouse:
Sorry if this wasn't clear but the links that are inline within a paragraph should have an underline. In this case that would be Government Digital Service and Open Government Licence v3.0

Regarding the legacy mode, should probably keep the underlines if it's not too much trouble.

nickcolley:
Spoke to people on Slack, since people using govuk_template wont be using this component, it seems like it's not a big importance to add the old behaviour.

dashouse:
Did you manage to make the inline link change? the preview is dead now

nickcolley:
Note that this style was updated in this pull request: https://github.com/alphagov/govuk-frontend/pull/1491

nickcolley:
Some context on why we removed underlines:

> I believe the intention was to treat the links in the footer in the same way we treat other navigational elements – like the header, or in the case of the Design System the side navigation. Where the ‘thing’ is a ‘list of links’, the thinking is/was that they don’t need the underlining because there’s no wider context in which they need to be distinguished. The exception is any link that does appear within e.g. a paragraph, in which case it does need the underline to distinguish it. It has been questioned a few times on support recently though.
> 
> The benefit, I believe, is that it allows us to have a clear hover state for those links, which we didn’t previously have

Issue #1325, opened 2019-05-08
Accordion 'open/close all sections' cannot be customised or translated


HughePaul:
https://github.com/alphagov/govuk-frontend/blob/1eef89c59c7120050d97cec283d0435cdbb00a45/src/components/accordion/accordion.js#L60

The accordion has hardcoded English text in javascript file, meaning it can't be translated to Welsh. Without being able to translate this our site can't enter Public Beta.

36degrees:
Hi Paul,

Thanks for raising. This has come up before a few times before – most recently in #1038.

Unfortunately at the minute we don't have a decent solution for localisation / internationalisation.

For now, we suggest that you copy the Javascript for this component into your service's codebase and make the adjustments required for the service you’re working on. I recognise this isn't ideal –  in #1038 I mentioned that we were hoping to look at this properly in 'early [2019],' but we haven't got to it yet.

Hope that helps,

Ollie

HughePaul:
I've penned an MR to try to pull hardcoded text  from data attributes and fall back to the existing English - could this be a suitable route for hardcoded text in the javascript?

36degrees:
Thanks for doing that! 👏 It sounds like a sensible option, but as it introduces new conventions to the component API we need to spend a little bit of time understanding how it might work for other components – both existing, and new ones – and to make sure it's the right approach.

Our team are currently prioritising work to make [GOV.UK Frontend pass WCAG 2.1 requirements](https://github.com/alphagov/govuk-design-system/issues/677) and we have an external audit coming up soon. Because of this, it may take a little while before we can review your PR.

If this is blocking you, I still recommend copying the JS into your service as the best short-term solution.

Thanks for your understanding.

HughePaul:
Given the new design of the accordion and the predominant English text for "Show all sections", "Hide all sections", "Show" and "Hide" compared to the old component, I think this has become a bit more important.

New accordion on page in welsh:
![new-accordion-welsh](https://user-images.githubusercontent.com/128149/150115593-9a9c6a0b-c3ab-4442-9e9a-47a8c39be3a3.png)

Old accordion on same page:
![old-accordion-welsh](https://user-images.githubusercontent.com/128149/150115598-c1d83134-f87c-4806-8b12-1a362f9037da.png)

[...]

PR #1328, opened 2019-05-09
Allow custom accordion open and close text

HughePaul:
Try to read custom open and close button text from a data attribute and fall back to English text to enable localisation.

fix for #1325

### DIFF
--- a/src/components/accordion/accordion.js
+++ b/src/components/accordion/accordion.js
@@ -55,9 +55,15 @@ Accordion.prototype.init = function () {
 // Initialise controls and set attributes
 Accordion.prototype.initControls = function () {
   // Create "Open all" button and set attributes
+  this.buttonText = {
+    openAll: this.$module.getAttribute('data-open-all-html') ||
+      'Open all <span class="govuk-visually-hidden">sections</span>',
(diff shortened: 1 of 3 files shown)


frankieroberto:
Nice! Feels a _bit_ weird to be including html fragments within attributes (as you'd have to remember to escape any `"` marks) but maybe that’s ok.

The alternative might be to bake multi-language support (at least for Welsh?) into the javascript, and then use a `data-lang` (or just `lang`?) attribute to select the right language.

hannalaakso:
Hi @HughePaul 

Thanks a lot for your work on this 🙌 This is a totally valid approach which solves your particular use case, but we need to find a way to support localisation consistently in a way that will work for different types of components and for other languages. For that reason, we’re not going to be able to accept this pull request. 

There's some related discussion here for context: https://github.com/alphagov/govuk-frontend/pull/958#issuecomment-440981361.

We've captured your PR as part of a larger epic https://github.com/alphagov/govuk-frontend/issues/1389 to improve how Javascript works in GOV.UK Frontend. For now, we suggest that you fork and vendor in this component and make the adjustments in the service you’re working on.

Thanks for taking the time to propose this, and for your patience.

PR #1329, opened 2019-05-09
New start button icon

aliuk2012:
Replacing two tones png file for an inline SVG.

## Before
old-start-button

## After
new-start-button

## TODO:
- [ ] add changelog entry
- [ ] test in all the browsers we support
- [ ] test text size 400%


src/components/button/template.njk
```diff
@@ -10,9 +21,14 @@
   {% endif %}
 {% endif %}
 
+{% if params.isStartButton %}
+  {% set iconMarkUp = '<svg data-name="Layer 1" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 20 20" height="19px" width="30px"><path fill="currentColor" d="M7.5 0H0l10 10L0 20h7.5l10-10-10-10z"/></svg>' %}
```
nickcolley: I think we can improve the inline SVG by doing the following:

- remove 'data-name' attribute
- add `.govuk-button__icon` class instead of styling the 'svg' element
- change `height` and `width` to be relative units (`em`)

src/components/button/template.njk
```diff
@@ -10,9 +21,14 @@
   {% endif %}
 {% endif %}
 
+{% if params.isStartButton %}
```
nickcolley: Based on our other API options like 'expanded' on the accordion, this should probably be something like `params.start`?

src/components/button/_button.scss
```diff
@@ -204,29 +228,39 @@
...
-    background-position: 100% 50%;
+    svg {
+      margin-top: -2px;
+      vertical-align: middle;
+    }
 
-    @include govuk-device-pixel-ratio {
-      background-image: govuk-image-url("icon-pointer-2x.png");
-      background-size: 30px 19px;
+    @include govuk-media-query($until: tablet) {
```
nickcolley: Might be a way to do this without using flexbox

nickcolley:
Screen Shot 2019-05-13 at 10 41 12

Worth checking the sizes for height and width so that they're consistent with the underlying icon, this'll make sure the padding for the button looks consistent with the previous version.

src/components/button/_button.scss
nickcolley: Do we need to do this with a breakpoint?

I think that `justify-content` might be covered by `align-items` too.

src/components/button/template.njk
aliuk2012: I based this on fieldset's `isPageHeading` option. I also like `is` as it helps the user understand that the API is expecting a Boolean.  We also had a chat about this offline and both agreed this would be alright.

fofr:
Will there be a solid black chevron for a focussed start button?

Screen Shot 2019-05-13 at 13 42 39

src/components/button/template.njk
aliuk2012: First two points have been done.

nickcolley:
@fofr yep :)

src/components/button/template.njk
```diff
@@ -10,9 +21,14 @@
...
+  {% set iconMarkUp = '<svg class="govuk-button__start-icon" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 20 20" height="19px" width="30px"><path fill="currentColor" d="M7.5 0H0l10 10L0 20h7.5l10-10-10-10z"/></svg>' %}
```
nickcolley: We should check if we need `focusable=false` on internet explorer here

aliuk2012: I've tested this in IE and it doesn't appear to be an issue. I think this might because the svg is already in a focusable element like `a` and `button`

aliuk2012:
Closing in favour of #1341

PR #1330, opened 2019-05-09
Spacing of tabs list updated to be more inline with similar lists on GOV.UK and the Design System

hannalaakso:
See https://github.com/alphagov/govuk-frontend/pull/1330#issuecomment-491233294

This PR:
- Adjusts the padding/margin so that with the the list version of component that’s used on mobile and no-js we use have `govuk-spacing(2)` as bottom margin. For the desktop/js-enabled version of component remove bottom margin and use top and bottom padding instead.
- Adjusts bottom margin of list item so that it’s applied on no-js version
- Adds some more spacing under the tabs title.

Split from https://github.com/alphagov/govuk-frontend/pull/1326

### DIFF
--- a/src/components/tabs/_tabs.scss
+++ b/src/components/tabs/_tabs.scss
@@ -13,16 +13,14 @@
 
   .govuk-tabs__title {
     @include govuk-font($size: 19);
-    margin-bottom: govuk-spacing(1);
+    margin-bottom: govuk-spacing(2);
   }
 
   .govuk-tabs__list {
     margin: 0;
     padding: 0;
     list-style: none;
-    @include mq($until: tablet) {
-      @include govuk-responsive-margin(6, "bottom");
-    }
+    @include govuk-responsive-margin(6, "bottom");
   }
 
   .govuk-tabs__list-item {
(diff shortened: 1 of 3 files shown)


dashouse:
Spacing of list updated to be more inline with similar lists on GOV.UK and the Design System. 

GOV.UK Example
Screen Shot 2019-05-10 at 10 54 43

GOV.UK Design System example
Screen Shot 2019-05-10 at 10 55 07

src/components/tabs/_tabs.scss
```diff
@@ -41,8 +39,10 @@
     @include govuk-font($size: 19);
 
     display: inline-block;
-    padding-top: govuk-spacing(2);
-    padding-bottom: govuk-spacing(2);
+    // Use bottom margin on mobile and when JS is not enabled.
+    // Consistent with what we do with on-page navigation in GOV.UK Design
```
36degrees: I think this comment can probably just be removed? I don't think we'd have written it if we'd built the component this way from scratch, and I think it's likely just to go out of date.

Issue #1308, opened 2019-04-30
Add contextual example with secondary and warning buttons


36degrees:
## What

Add one of more 'full-page' examples that show the secondary and warning buttons in a realistic context.

## Why

We're updating the focus style as part of [making the Design System meet WCAG 2.1](https://github.com/alphagov/govuk-design-system/issues/677) – specifically with regard to WCAG 2.1 1.4.11 Non-Text Contrast (#1137).

As part of this we are getting our full-page examples audited. We want to ensure that these newly introduced variants are included in the audit

## Done when…

- [ ] Realistic example defined
- [ ] Example(s) added to the review app which show the secondary and warning buttons in context

PR #1331, opened 2019-05-10
Add example using secondary and warning button

dashouse:
Adds an example using Warning and Secondary Button. I don't know if it's the most realistic example but It should be enough for the audit?

Screen Shot 2019-05-10 at 09 32 04


36degrees:
This looks great Dave 👍I think this'll do fine.

_Ideally_ this needs to be rebased against the `v3` branch so that it uses the new styles, and the 'base' edited so that this PR will get merged into `v3` (you can do that by clicking the edit button up by the title, and changing the base branch from the dropdown)

Let me know if I can help with any of that.

36degrees:
Closes #1308

PR #1332, opened 2019-05-10
Underline meta information links in footer

nickcolley:
See Dave's comments on https://github.com/alphagov/govuk-frontend/pull/1321#issuecomment-491220070

### DIFF
--- a/src/components/footer/_footer.scss
+++ b/src/components/footer/_footer.scss
@@ -44,13 +44,6 @@
   }
 
   .govuk-footer__link {
-    text-decoration: none;
-
-    &:hover,
-    &:active {
-      text-decoration: underline;
-    }
-
     @if ($govuk-use-legacy-palette) {
       &:link,
       &:visited {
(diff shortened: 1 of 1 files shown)


src/components/footer/_footer.scss
```diff
@@ -150,6 +154,11 @@
     margin-bottom: govuk-spacing(4);
   }
 
+
+  .govuk-footer__meta-custom .govuk-footer__link {
```
nickcolley: Something weird has happened here, let me investigate

36degrees:
I don't really have a preference either way, but wondering if it'd make sense to only take the underline away for links inside `govuk-footer__list` or `govuk-footer__inline-list`, rather than taking them away and then adding them back again.

It'd also mean if we added additional elements to the footer in the future they'd get underlines unless we explicitly removed them.

nickcolley:
Yeah lets do that

PR #1333, opened 2019-05-10
Remove deprecated `@else-if` statement, replace with `@else if`

gunjam:
When compiling with the [sass](https://www.npmjs.com/package/sass) npm module I'm getting the following deprecation warning:
```
DEPRECATION WARNING on line 145, column 7 of node_modules/govuk-frontend/helpers/_typography.scss:
@elseif is deprecated and will not be supported in future Sass versions.
Use "@else if" instead.
    ╷
145 │     } @elseif $breakpoint == "print" {
    │       ^^^^^^^
    ╵
```
This should fix the issue, while still sucessfully building with the `node-sass` package.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,10 @@
 
 🔧 Fixes:
 
+- Remove deprecated `@else-if` statement, replace with `@else if`
+
+  ([PR #1333](https://github.com/alphagov/govuk-frontend/pull/1333))
+
 - Pull Request Title goes here
 
   Description goes here (optional)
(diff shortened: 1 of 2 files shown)


36degrees:
Thanks for this, @gunjam! 

Would you mind [adding an entry to the changelog](https://github.com/alphagov/govuk-frontend/blob/master/docs/contributing/versioning.md#updating-changelog)?

This will then need another review from someone else on the team before we can get it merged. It's also worth mentioning that we're currently working towards a 3.0 release and so it may take a while before this actually goes out.

gunjam:
Hi @36degrees, I updated `CHANGELOG.md`, did I do it right? :)

36degrees:
Bang on 👌– thanks!

PR #1334, opened 2019-05-10
Add index page for full page examples with scenarios and notes

36degrees:
This adds an index page that lists out all full-page examples, along with scenarios and notes that are populated from front-matter.

localhost_3000_full-page-examples

👉 https://govuk-frontend-review-pr-1334.herokuapp.com/full-page-examples

### DIFF
--- a/lib/file-helper.js
+++ b/lib/file-helper.js
@@ -3,6 +3,7 @@
 const fs = require('fs')
 const path = require('path')
 const yaml = require('js-yaml')
+const fm = require('front-matter')
 
 const configPaths = require('../config/paths.json')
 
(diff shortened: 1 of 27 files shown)


nickcolley:
Should make these pages easier to maintain too, nice 👍

lib/file-helper.js
```diff
@@ -34,3 +35,13 @@ const getComponentData = componentName => {
...
 exports.getComponentData = getComponentData
+
+exports.fullPageExamples = () => {
+  return childDirectories(path.resolve(configPaths.fullPageExamples))
+    .map(folderName => ({
+      name: folderName,
+      path: folderName,
+      ...fm(readFileContents(path.join(configPaths.fullPageExamples, folderName, 'index.njk'))).attributes
+    }))
+    .sort((a, b) => (a.name.toLowerCase() > b.name.toLowerCase()) ? 1 : -1)
```
hannalaakso: Non-blocking: Could leave a comment to say `Case insensitive sort by name`

PR #1336, opened 2019-05-10
Fix app banner contrast for IE8

hannalaakso:
I think this fix wasn't meant to be taken out. See original code https://github.com/alphagov/govuk-frontend/commit/ea6ac782caac605403fc896b81cc9827a2be6a01#diff-fa4461f64f4ca9ed30f64c87616594fd

This PR fixes the link style in the review app banner to avoid poor contrast in IE8 (which doesn't support `:not`).

Before:
Screen Shot 2019-05-10 at 15 42 05

After:
Screen Shot 2019-05-10 at 15 42 24

### DIFF
--- a/app/assets/scss/partials/_banner.scss
+++ b/app/assets/scss/partials/_banner.scss
@@ -9,8 +9,12 @@
     font-size: inherit;
   }
 
-  .govuk-link:not(:focus) {
-    color: inherit;
+  .govuk-link {
+    color: govuk-colour("white");
+   }
+
+  .govuk-link:focus {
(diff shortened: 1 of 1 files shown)


app/assets/scss/partials/_banner.scss
```diff
@@ -9,8 +9,12 @@
     font-size: inherit;
   }
 
-  .govuk-link:not(:focus) {
-    color: inherit;
+  .govuk-link {
+    color: govuk-colour("white");
```
nickcolley: I think this should have been inherit still but works for the most part.

PR #1338, opened 2019-05-13
Fix HTML validation errors in review app 

hannalaakso:
This fixes the following:
- A document must not include more than one visible `main` element.
- Element `div` not allowed as child of element `h2` in this context

### DIFF
--- a/app/views/full-page-examples/bank-holidays/index.njk
+++ b/app/views/full-page-examples/bank-holidays/index.njk
@@ -50,9 +50,9 @@ notes: >-
       {% set englandAndWalesHTML %}
 
         {% set englandAndWalesPanelTitleHtml %}
-          <div class='govuk-!-font-size-36 govuk-!-margin-bottom-4'>
+          <span class='govuk-!-font-size-36 govuk-!-margin-bottom-4 govuk-!-display-block'>
             The next bank holiday in England and Wales is
-          </div>
+          </span>
           19 April
         {% endset %}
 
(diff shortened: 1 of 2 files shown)


app/views/full-page-examples/bank-holidays/index.njk
```diff
@@ -50,9 +50,10 @@ notes: >-
       {% set englandAndWalesHTML %}
 
         {% set englandAndWalesPanelTitleHtml %}
-          <div class='govuk-!-font-size-36 govuk-!-margin-bottom-4'>
+          <span class='govuk-!-font-size-36'>
             The next bank holiday in England and Wales is
-          </div>
+          </span>
+          <br><br>
```
nickcolley: Is there a way to do this without the <br>? do we have a `govuk-!-display-block`? Could make an `app-` version maybe.

36degrees: Wondering if we could/should just put `display: block` on the span? Not pretty, but neither is random line breaks.

app/views/full-page-examples/confirm-delete/index.njk
```diff
@@ -9,40 +9,38 @@
 {% set mainClasses = "govuk-main-wrapper--l" %}
 
 {% block content %}
-  <main id="content" role="main">
```
36degrees: Good catch 👏

app/views/full-page-examples/bank-holidays/index.njk
hannalaakso: Yes a good idea 👍

PR #1341, opened 2019-05-13
Add new start button icon

nickcolley:
V3:
V3

V3 Focused:
V3 Focused

V3 Legacy mode:
V3 Legacy mode

V3 Legacy mode focused:
V3 Legacy mode focused

V2:
V2

V2 Focused:
V2 Focused

Changed colours:

Screen Shot 2019-05-15 at 12 54 14

## Issues: 
<details>
<summary>SVG:</summary>

Internet Explorer 8 does not support SVG and will therefore not display the icon. We did not feel that it warranted a fallback as it did not affect how the user would interact with the button and the icon was purely used for presentation. 
</details>

<details>
<summary>Flexbox</summary>

As this implementation uses flexbox not all browsers support it. I've checked with @dashouse and this shouldn't be a problem as the icon will wrap instead of floating to the right side.

### Example where browser doesn't support flexbox
Screen Shot 2019-05-15 at 09 50 20
</details>


nickcolley:
@aliuk2012 I've explored some changes in this branch, will chat with you when we're both in the office. 👍

src/components/button/template.njk
```diff
@@ -10,9 +21,18 @@
   {% endif %}
 {% endif %}
 
+{% if params.isStartButton %}
+  {% set iconHtml %}
+    <svg class="govuk-button__start-icon" xmlns="http://www.w3.org/2000/svg" width="17.5" height="19" viewBox="0 0 33 40" focusable="false">
+      <path d="M0 0h13l20 20-20 20H0l21-19z" fill="currentColor"></path>
```
dashouse:
```suggestion
      <path d="M0 0h13l20 20-20 20H0l20-20z" fill="currentColor"></path>
```

dashouse: Before
Screen Shot 2019-05-15 at 11 23 42

After
Screen Shot 2019-05-15 at 11 23 53

src/components/button/_button.scss
```diff
@@ -224,20 +224,24 @@
...
+  }
 
-    @include govuk-device-pixel-ratio {
-      background-image: govuk-image-url("icon-pointer-2x.png");
-      background-size: 30px 19px;
-    }
+  .govuk-button__start-icon {
+    margin-top: -.125em;
+    margin-right: -.08333333333333333em;
+    margin-left: .4166666666666667em;
```
hannalaakso: Could these use `govuk-em`?

src/components/button/template.njk
```diff
@@ -1,18 +1,39 @@
...
     {% set element = 'button' %}
   {% endif %}
+{% endif -%}
+
+{% if params.isStartButton %}
+  {% set iconHtml %}
+{#- The SVG needs `focusable="false"` so that Internet Explorer does not
+treat it as an interactive element - without this it will be
+'focusable' when using the keyboard to navigate. #}
+<svg class="govuk-button__start-icon" xmlns="http://www.w3.org/2000/svg" width="17.5" height="19" viewBox="0 0 33 40" focusable="false">
```
hannalaakso: Non-blocking: It would be nice to indent the svg similarly to https://github.com/alphagov/govuk-frontend/blob/840d019e0412309d6f8ab1032ec3b17b30fcd955/src/components/header/template.njk#L20 in the template for readability

aliuk2012: This was done intentional because nunjucks was not producing neat example markup in the review app.

src/components/button/_button.scss
nickcolley: When I initially did this the icon scaled alongside the font size, but to keep it consistent with the current implementation it no longer does that.

So with that in mind we don't need to use relative units (em), so I've changed it back to pixels.

dashouse: I think this should be:

```
margin-top: -3px;
margin-right: 0; (or not there)
@include govuk-media-query($from: desktop) {
    margin-left: govuk-spacing(2);
}
```
The spacing scale does not change below `govuk-spacing(4)` so this change from 1 to 2 would need to be done with media queries

nickcolley:
@hannalaakso there's not really a great way to keep the icon right aligned on smaller screens that I know of with flexbox, so this felt like an OK compromise for the benefits we get from the inline SVG.

Issue #1091, opened 2018-12-13
Remove deprecated grid-width function


36degrees:
In https://github.com/alphagov/govuk-frontend/pull/1090 we renamed the `grid-width` function to `govuk-grid-width` to be consistent with the other public functions which are in the `govuk` namespace.

We retained `grid-width` as an alias, but should remove it in the next breaking release (v3.0).

PR #1342, opened 2019-05-14
Remove grid-width mixin

aliuk2012:
https://github.com/alphagov/govuk-frontend/pull/1090 deprecated this mixin
and left it in as an alias to govuk-grid-width.

ACTION: Replace all instances of grid-width with govuk-grid-width

Closes https://github.com/alphagov/govuk-frontend/issues/1091

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -135,6 +135,15 @@
 
   ([PR #1343](https://github.com/alphagov/govuk-frontend/pull/1343))
 
+- Remove grid-width mixin
+
+  https://github.com/alphagov/govuk-frontend/pull/1090 deprecated this mixin
+  and left it in as an alias to govuk-grid-width.
+
+  To migrate you'll need to replace any reference to the `grid-width` mixin with `govuk-grid-width`
+
+  ([PR #1342](https://github.com/alphagov/govuk-frontend/pull/1342))
+
 - Pull Request Title goes here
 
   Description goes here (optional)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -126,6 +126,17 @@
...
   ([PR #1330](https://github.com/alphagov/govuk-frontend/pull/1330))
 
+- Remove grid-width mixin
+
+  https://github.com/alphagov/govuk-frontend/pull/1090 deprecated this mixin
+  and left it in as an alias to govuk-grid-width.
+
+  **ACTION TO BE TAKEN**: Replace all instances of `grid-width` mixin with `govuk-grid-width`
+
+Closes https://github.com/alphagov/govuk-frontend/issues/1091
```
36degrees: We've never put this in before. Do we think this is useful for end users?

CHANGELOG.md
```diff
@@ -126,6 +126,17 @@
 
   ([PR #1330](https://github.com/alphagov/govuk-frontend/pull/1330))
 
+- Remove grid-width mixin
+
+  https://github.com/alphagov/govuk-frontend/pull/1090 deprecated this mixin
+  and left it in as an alias to govuk-grid-width.
+
+  **ACTION TO BE TAKEN**: Replace all instances of `grid-width` mixin with `govuk-grid-width`
```
36degrees: Again, not sure about this.

```suggestion
  To migrate you'll need to replace any references to the `grid-width` mixin with `govuk-grid-width`
```

CHANGELOG.md
aliuk2012: Ah, thats not meant to be in there. 👍 good spot

Issue #1092, opened 2018-12-13
Remove deprecated govuk-grid-row mixin


36degrees:
In https://github.com/alphagov/govuk-frontend/pull/1090 we moved the rules from the `govuk-grid-row` mixin and used them to create a concrete `.govuk-grid-row` class.

We retained the existing `govuk-grid-row` mixin to avoid a breaking change, but should remove it in the next breaking release (v3.0).

PR #1343, opened 2019-05-14
Removes govuk-grid-row mixin

aliuk2012:
https://github.com/alphagov/govuk-frontend/pull/1090 copied govuk-grid-row
mixin to create a new concrete `.govuk-grid-row` class and marked the mixin as deprecated.

ACTION TO BE TAKEN: Remove any references to `govuk-grid-row` mixin and
                    use `.govuk-grid-row` class.

closes: https://github.com/alphagov/govuk-frontend/issues/1092

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -126,6 +126,15 @@
 
   ([PR #1330](https://github.com/alphagov/govuk-frontend/pull/1330))
 
+- Removes `govuk-grid-row` mixin
+
+  https://github.com/alphagov/govuk-frontend/pull/1090 copied govuk-grid-row
+  mixin to create a new concrete `.govuk-grid-row` class and marked the mixin as deprecated.
+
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -126,6 +126,15 @@
 
   ([PR #1330](https://github.com/alphagov/govuk-frontend/pull/1330))
 
+- Removes `govuk-grid-row` mixin
+
+  https://github.com/alphagov/govuk-frontend/pull/1090 copied govuk-grid-row
+  mixin to create a new concrete `.govuk-grid-row` class and marked the mixin as deprecated.
+
+  **ACTION TO BE TAKEN**: Remove any references to `govuk-grid-row` mixin and use `.govuk-grid-row` class.
```
36degrees: This seems a little shouty considering I'd be very surprised if anyone's actually using it. Can we consider using the same format we used for the colour palette change?

> To migrate you'll need to remove any references to `govuk-grid-row` mixin and use `.govuk-grid-row` class instead.

PR #1346, opened 2019-05-14
Update header focus styles so the menu button is consistent

nickcolley:
Before:
Screen Shot 2019-05-14 at 17 06 53

After:
Screen Shot 2019-05-14 at 17 07 12

## Todo

- [x] Cross browser testing


nickcolley:
@aliuk2012 we need to improve the changelog anyway, I don't think me adding one for this would be useful given we already have a header changelog...

PR #1347, opened 2019-05-15
Support initial aria-describedby on all form fields

colinrotherham:
We've noticed the `describedBy` param used internally can't be configured.

This pull request supports an initial `aria-describedby` value on all form fields, populated before the optional hint and error message IDs are appended.

For form inputs:

```js
{{ govukInput({
  describedBy: 'parent-fieldset-error-id'
}) }}
```

Or for those wrapped by fieldsets:

```js
{{ govukDateInput({
  fieldset: {
    describedBy: 'parent-fieldset-error-id'
  }
}) }}
```

Useful when fields are described by errors or hints on parent fieldsets such as:

Parent fieldset error

When moving between these nested checkboxes, we can now use `params.fieldset.describedBy` to repeat the error message "Select at least one reason" to screen readers as a usability improvement.

### DIFF
--- a/src/components/checkboxes/checkboxes.yaml
+++ b/src/components/checkboxes/checkboxes.yaml
@@ -1,4 +1,8 @@
 params:
+- name: describedBy
+  type: string
+  required: false
+  description: One or more element IDs to add to the input `aria-describedby` attribute without a fieldset, used to provide additional descriptive information for screenreader users.
 - name: fieldset
   type: object
   required: false
@@ -240,6 +244,22 @@ examples:
         text: I agree to the terms and conditions
         hint:
           text: Go on, you know you want to!
+
+- name: with fieldset and error message
+  data:
+    name: colours
+    errorMessage:
+      text: Please accept the terms and conditions
+    fieldset:
+      legend:
+        text: What is your nationality?
+    items:
+      - value: british
+        text: British
+      - value: irish
+        text: Irish
+      - value: other
+        text: Citizen of another country
 
 - name: with all fieldset attributes
   data:
--- a/src/components/checkboxes/template.njk
+++ b/src/components/checkboxes/template.njk
@@ -9,7 +9,10 @@
 
 {#- a record of other elements that we need to associate with the input using
    aria-describedby – for example hints or error messages -#}
-{% set describedBy = "" %}
+{% set describedBy = params.describedBy if params.describedBy else "" %}
+{% if params.fieldset.describedBy %}
+   {% set describedBy = params.fieldset.describedBy %}
+{% endif %}
 
 {% set isConditional = false %}
 {% for item in params.items %}
--- a/src/components/checkboxes/template.test.js
+++ b/src/components/checkboxes/template.test.js
@@ -71,6 +71,30 @@ describe('Checkboxes', () => {
     expect($component.hasClass('app-checkboxes--custom-modifier')).toBeTruthy()
   })
 
+  it('renders initial aria-describedby on fieldset', () => {
+    const describedById = 'some-id'
+
+    const $ = render('checkboxes', {
+      name: 'example-name',
+      fieldset: {
+        describedBy: describedById
+      },
+      items: [
+        {
+          value: '1',
+          text: 'Option 1'
+        },
+        {
+          value: '2',
+          text: 'Option 2'
+        }
+      ]
+    })
+
+    const $fieldset = $('.govuk-fieldset')
+    expect($fieldset.attr('aria-describedby')).toMatch(describedById)
+  })
+
   it('render attributes', () => {
     const $ = render('checkboxes', {
       name: 'example-name',
@@ -521,7 +545,7 @@ describe('Checkboxes', () => {
     })
 
     it('associates the fieldset as "described by" the error message', () => {
-      const $ = render('checkboxes', examples['with all fieldset attributes'])
+      const $ = render('checkboxes', examples['with fieldset and error message'])
 
       const $fieldset = $('.govuk-fieldset')
       const $errorMessage = $('.govuk-error-message')
@@ -532,6 +556,27 @@ describe('Checkboxes', () => {
 
       expect($fieldset.attr('aria-describedby'))
         .toMatch(errorMessageId)
+    })
+
+    it('associates the fieldset as "described by" the error message and parent fieldset', () => {
+      const describedById = 'some-id'
+      const params = examples['with fieldset and error message']
+
+      params.fieldset.describedBy = describedById
+
+      const $ = render('checkboxes', params)
+
+      const $fieldset = $('.govuk-fieldset')
+      const $errorMessage = $('.govuk-error-message')
+
+      const errorMessageId = new RegExp(
+        WORD_BOUNDARY + describedById + WHITESPACE + $errorMessage.attr('id') + WORD_BOUNDARY
+      )
+
+      expect($fieldset.attr('aria-describedby'))
+        .toMatch(errorMessageId)
+
+      delete params.fieldset.describedBy
     })
 
     it('does not associate each input as "described by" the error message', () => {
@@ -576,6 +621,24 @@ describe('Checkboxes', () => {
 
       expect($fieldset.attr('aria-describedby')).toMatch(hintId)
     })
+
+    it('associates the fieldset as "described by" the hint and parent fieldset', () => {
+      const describedById = 'some-id'
+      const params = examples['with all fieldset attributes']
+
+      params.fieldset.describedBy = describedById
+
+      const $ = render('checkboxes', params)
+      const $fieldset = $('.govuk-fieldset')
+      const $hint = $('.govuk-hint')
+
+      const hintId = new RegExp(
+        WORD_BOUNDARY + describedById + WHITESPACE + $hint.attr('id') + WORD_BOUNDARY
+      )
+
+      expect($fieldset.attr('aria-describedby')).toMatch(hintId)
+      delete params.fieldset.describedBy
+    })
   })
 
   describe('when they include both a hint and an error message', () => {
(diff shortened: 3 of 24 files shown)


colinrotherham:
One suggestion, is perhaps we rename it `describedByPrefix` so it's more obvious that we're only setting the initial `aria-describedby` attribute value before hints and error message IDs are appended.

Similar to `idPrefix` and `namePrefix` used by checkboxes and dates.

src/components/checkboxes/template.njk
```diff
@@ -9,7 +9,10 @@
 
 {#- a record of other elements that we need to associate with the input using
    aria-describedby – for example hints or error messages -#}
-{% set describedBy = "" %}
+{% set describedBy = params.describedBy if params.describedBy else "" %}
+{% if params.fieldset.describedBy %}
```
36degrees: Am I right in thinking there's no difference between passing `{ fieldset : { describedBy: 'foo' }}` and `{ describedBy: 'foo' }`? Wondering if we need to support both of those, or if we should just support passing it on the fieldset.

src/components/checkboxes/checkboxes.yaml
```diff
@@ -241,6 +241,22 @@ examples:
         hint:
```
36degrees: _If_ we're allowing describedBy to be passed as a 'root' option for the checkboxes component, we should document it in `params`.

src/components/fieldset/template.test.js
```diff
@@ -61,6 +61,15 @@ describe('fieldset', () => {
     expect($legend.text().trim()).toEqual('What is your address?')
   })
 
+  it('allows you to set the initial aria-describedby attribute', () => {
```
36degrees: 'Initial' here suggests that `describedBy` may still change, but it's outputted as it is.

```suggestion
  it('allows you to set the aria-describedby attribute', () => {
```

src/components/fieldset/template.test.js
```diff
@@ -61,6 +61,15 @@ describe('fieldset', () => {
...
+    const $ = render('fieldset', {
+      describedBy: 'parent-fieldset-hint-id'
```
36degrees: In most cases this _won't_ be the parent fieldset hint ID, so can we make this more generic? Maybe even just…

```suggestion
      describedBy: 'some-id'
```

src/components/fieldset/template.test.js
```diff
@@ -61,6 +61,15 @@ describe('fieldset', () => {
...
+    })
+
+    const $component = $('.govuk-fieldset')
+    expect($component.attr('aria-describedby')).toEqual('parent-fieldset-hint-id')
```
36degrees: Very much non-blocking, but as `$component` only gets used once and there's nothing complicated going on I'd probably just chain it directly

```suggestion
    expect($('.govuk-fieldset').attr('aria-describedby')).toEqual('parent-fieldset-hint-id')
```

src/components/checkboxes/template.njk
36degrees: This would also be consistent with the date input and radio components.

src/components/file-upload/file-upload.yaml
```diff
@@ -11,6 +11,10 @@ params:
   type: string
   required: false
   description: Optional initial value of the input
+- name: describedBy
+  type: string
+  required: false
+  description: Optional text or element id to add to the `aria-describedby` attribute to provide a description for screenreader users.
```
36degrees: Only element IDs will work with `aria-describedby` – text can't be passed. How about something like…

```suggestion
  description: One or more element IDs to add to the `aria-describedby` attribute, used to provide additional descriptive information for screenreader users.
```

?

src/components/input/input.yaml
```diff
@@ -15,6 +15,10 @@ params:
   type: string
   required: false
   description: Optional initial value of the input.
+- name: describedBy
+  type: string
+  required: false
+  description: Optional text or element id to add to the `aria-describedby` attribute to provide a description for screenreader users.
```
36degrees: Again, this needs correcting as you cannot pass text.

src/components/select/template.test.js
```diff
@@ -135,6 +135,19 @@ describe('Select', () => {
...
+    it('renders with aria-describedby', () => {
+      const describedById = 'parent-fieldset-hint-id'
+
+      const $ = render('select', {
+        describedBy: describedById
+      })
+
+      const $component = $('.govuk-select')
+
+      expect($component.attr('aria-describedby'))
```
36degrees: Again non-blocking, but I'd just chain `$('govuk-select')`

src/components/textarea/textarea.yaml
```diff
@@ -15,6 +15,10 @@ params:
   type: string
   required: false
   description: Optional initial value of the textarea.
+- name: describedBy
+  type: string
+  required: false
+  description: Optional text or element id to add to the `aria-describedby` attribute to provide a description for screenreader users.
```
36degrees: Again, this needs correcting as you cannot pass text.

src/components/select/select.yaml
```diff
@@ -32,6 +32,10 @@ params:
     type: object
     required: false
     description: HTML attributes (for example data attributes) to add to the option.
+- name: describedBy
+  type: string
+  required: false
+  description: Optional text or element id to add to the `aria-describedby` attribute to provide a description for screenreader users.
```
36degrees: Again, this needs correcting as you cannot pass text.

src/components/checkboxes/template.test.js
```diff
@@ -576,22 +621,62 @@ describe('Checkboxes', () => {
...
   })
 
   describe('when they include both a hint and an error message', () => {
     it('associates the fieldset as described by both the hint and the error message', () => {
       const $ = render('checkboxes', examples['with all fieldset attributes'])
 
       const $fieldset = $('.govuk-fieldset')
-      const $errorMessageId = $('.govuk-error-message').attr('id')
-      const $hintId = $('.govuk-hint').attr('id')
+      const errorMessageId = $('.govuk-error-message').attr('id')
```
36degrees: Good spot with the `$` removal 👀

src/components/checkboxes/checkboxes.yaml
36degrees: (Having reviewed the rest of the PR, I suspect we're not so this is probably irrelevant)

colinrotherham:
Thanks @36degrees, I'll fix the `fieldset` param description, I copied it from there:

```yaml
params:
- name: describedBy
  type: string
  required: false
  description: Text or element id to add to the `aria-describedby` attribute to provide description of the group of fields for screenreader users.
```

I'll remove the "Text or" bit from all mine too, so it's just the element id.

36degrees:
Hah, of course it came from us 🤦‍♂

src/components/select/template.test.js
colinrotherham: @36degrees Mind if we keep it but I'll group it better?

It's done like this throughout a lot of tests:

```js
const $component = $('.govuk-select')
expect($component.attr('aria-describedby')).toMatch(describedById)
```

colinrotherham: So if the `expect()` comes after the `$component` declaration it's now as above.
i.e. Two lines, variable first, with the assertion all on one line.

If it's following other variable declarations like `errorMessageId` and `hintId` it's written like so:

```js
expect($component.attr('aria-describedby'))
  .toMatch(errorMessageId)
```

I think I understand your code style now 😁

src/components/file-upload/file-upload.yaml
colinrotherham: Perfect, I've gone with this one in the latest push

src/components/fieldset/template.test.js
colinrotherham: It's like this in lots of tests, so I copied this pattern. Shall we leave it?

src/components/select/template.test.js
36degrees: TBH, I don't think we're terribly good as being consistent on this, so this is more my personal preference than anything 😬

Happy for you to leave it as it is and we'll try and firm up our style a little and go through and tidy up… one day…

src/components/checkboxes/template.njk
colinrotherham: @36degrees It's because checkboxes are special, unlike radios.

>Checkboxes with single option set 'aria-describedby' on input

http://govuk-frontend-review.herokuapp.com/components/checkboxes/with-single-option-set-'aria-describedby'-on-input/preview

So the `describedBy` attribute can be without a fieldset.

src/components/fieldset/template.test.js
colinrotherham: Going to leave this as it matches other assertions where `$component` is only used once.

See:

https://github.com/alphagov/govuk-frontend/blob/master/src/components/fieldset/template.test.js#L28

https://github.com/alphagov/govuk-frontend/blob/master/src/components/fieldset/template.test.js#L39

https://github.com/alphagov/govuk-frontend/blob/master/src/components/fieldset/template.test.js#L50

etc

src/components/checkboxes/checkboxes.yaml
colinrotherham: So this relates to https://github.com/alphagov/govuk-frontend/pull/1347#discussion_r285124421, but I forgot to add the root param.

When used without a fieldset, this is where `describedBy` can be set.

I'd make this the description?

>One or more element IDs to add to the input `aria-describedby` attribute without a fieldset, used to provide additional descriptive information for screenreader users.

Unless you'd prefer not to allow `describedBy` to be customised when no fieldset is set?

colinrotherham:
@36degrees Thanks for the thorough review 👍 

I've left two of the conversations open regarding the single checkbox without a fieldset. You might not want to support customising `describedBy` here.

src/components/checkboxes/template.njk
36degrees: Gotcha, in which case I think this makes sense, thanks for talking me through it.

I can't see a test that covers this functionality (setting aria-describedby on a single checkbox without a fieldset) – could we add one?

colinrotherham: You're right! Thanks.

I've extended the existing single checkbox tests and made two new ones.

You can see the new bit pushed here: https://github.com/alphagov/govuk-frontend/compare/8e23ce86235c93a77d54b7ccccb0edad923ef7e5..09a9700814d433925b52a9f6d25848093cacdd61

src/components/checkboxes/checkboxes.yaml
colinrotherham: I'll resolve this one as we're chatting on https://github.com/alphagov/govuk-frontend/pull/1347#discussion_r286029798

colinrotherham:
Two new tests added to exercise the single checkbox variations:

* `with single option set 'aria-describedby' on input`
* `with single option (and hint) set 'aria-describedby' on input`

The new tests show how `params.describedBy` works here (instead of `params.fieldset.describedBy`) as neither variation has a wrapping fieldset.

colinrotherham:
@36degrees Perfect, thank you.

No immediate urgency. Would be a great one to have alongside 3.x 👍

Issue #1252, opened 2019-03-22
Child elements in a tab label break when clicked


fofr:
In the [Tabs component](https://design-system.service.gov.uk/components/tabs/), if you include HTML within a tab label – eg an abbreviation tag. When you try and click that element the javascript events don’t bubble up correctly and the component breaks.

It's probably this line causing it:
https://github.com/alphagov/govuk-frontend/blob/master/src/components/tabs/tabs.js#L172

![example-of-bug](https://user-images.githubusercontent.com/319055/54821295-9d0b9500-4c99-11e9-9d85-a911897972cf.gif)

## Example nunJucks:

```
{{ govukTabs({
    items: [
      {
        label: 'Tab with child element, eg <strong>strong</strong>' | safe,
        id: "first",
        panel: {
          html: 'First tab'
        }
      },
      {
        label: 'Normal tab',
        id: "second",
        panel: {
          html: 'Second tab'
        }
      }
    ]
  }) }}
```

```html
<li class="govuk-tabs__list-item">
  <a class="govuk-tabs__tab govuk-tabs__tab--selected" href="#first">
    Tab with child element, eg <strong>strong</strong>
  </a>
</li>
```

PR #1351, opened 2019-05-15
Fix tabs label child elements breaking

hannalaakso:
Using HTML elements inside the tabs label breaks as `event.target` is looking at the child element instead of the tab label. 

This fix checks for the class of the target to allow the event to bubble up to the parent which is the label.

(Looks like this can also be fixed by adding

```
    & * {
      pointer-events: none;
    }
```

to https://github.com/alphagov/govuk-frontend/blob/840d019e0412309d6f8ab1032ec3b17b30fcd955/src/components/tabs/_tabs.scss#L46 but it makes more sense to fix this in JS rather than couple the fix with the stylesheet).

~Question: **Should we change the public API to accommodate adding markup inside the label by having a text and html option?** It seems like it would be useful. At the moment it's not possible to add a test (I don't think) to check for this as you can't add Nunjucks from the example yaml to set `| safe`, it needs to be done [in the template](https://github.com/alphagov/govuk-frontend/blob/840d019e0412309d6f8ab1032ec3b17b30fcd955/src/components/tabs/template.njk#L17).  I've tested this locally by amending one of the full page examples following [@fofr's example Nunjucks](https://github.com/alphagov/govuk-frontend/issues/1252).~ Edit: Had a discussion with the team. We currently can't see a user need to make larger changes to the API.

Also created a [tech debt issue](https://trello.com/c/yaKqbyQg) to look at refactoring the tabs Javascript at some point.

### Browser testing:

- [x] Chrome 74
- [x] Firefox 65
- [x] OS Safari 11
- [x] Android Samsung Galaxy (Chrome)
- [x] iOS XS / 8 (Safari) 
- [x] Edge 18
- [x] IE 8 - 11

Fixes https://github.com/alphagov/govuk-frontend/issues/1252

### DIFF
--- a/src/components/tabs/tabs.test.js
+++ b/src/components/tabs/tabs.test.js
@@ -78,6 +78,27 @@ describe('/components/tabs', () => {
         })
         expect(secondTabPanelIsHidden).toBeFalsy()
       })
+
+      describe('when the tab contains a DOM element', () => {
+        it('should display the tab panel associated with the selected tab', async () => {
+          await page.goto(baseUrl + '/components/tabs/preview', { waitUntil: 'load' })
+
+          await page.evaluate(() => {
+            // Replace contents of second tab with a DOM element
+            const secondTab = document.body.querySelector('.govuk-tabs__list-item:nth-child(2) .govuk-tabs__tab')
+            secondTab.innerHTML = '<span>Tab 2</span>'
+          })
+
+          // Click the DOM element inside the second tab
+          await page.click('.govuk-tabs__list-item:nth-child(2) .govuk-tabs__tab span')
+
+          const secondTabPanelIsHidden = await page.evaluate(() => {
+            const secondTabAriaControls = document.body.querySelector('.govuk-tabs__list-item:nth-child(2) .govuk-tabs__tab').getAttribute('aria-controls')
+            return document.body.querySelector(`[id="${secondTabAriaControls}"]`).classList.contains('govuk-tabs__panel--hidden')
+          })
+          expect(secondTabPanelIsHidden).toBeFalsy()
+        })
+      })
     })
 
     describe('when first tab is focused and the right arrow key is pressed', () => {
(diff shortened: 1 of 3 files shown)


36degrees:
Seems like a great way to fix this issue. I wonder if we should have a test for this? I know when I was looking at this it was difficult, because we don't have a way to pass HTML for tab labels (yet) – but perhaps we could fake it in the tests by injecting an extra element?

hannalaakso:
Thanks for the review @36degrees 👍 I've added a test where the contents of the tab label are replaced with a DOM element.

src/components/tabs/tabs.test.js
```diff
@@ -78,6 +78,27 @@ describe('/components/tabs', () => {
...
+          await page.goto(baseUrl + '/components/tabs/preview', { waitUntil: 'load' })
+
+          await page.evaluate(() => {
+            // Replace contents of second tab with a DOM element
+            const secondTab = document.body.querySelector('.govuk-tabs__list-item:nth-child(2) .govuk-tabs__tab')
+            secondTab.innerHTML = '<span>Tab 2</span>'
+          })
+
+          // Click the second tab
+          await page.click('.govuk-tabs__list-item:nth-child(2) .govuk-tabs__tab')
```
fofr: Should the click event be on the new `<span>` element?

36degrees: To test the bubbling, don't we need to click on the span?

36degrees: …jinx? 😆

hannalaakso: Okay I've added `span` to the selector. As the span covers the whole label, the test was actually correctly failing without the bubbling fix as the span was always clicked on - whereas adding more text inside the label made the test pass as the span wasn't clicked on anymore. But agree that it's more correct / explicit clicking on the span.

hannalaakso: See comment above 🙏

PR #1353, opened 2019-05-15
Fix improperly indented html output in Design System examples

nickcolley:
From support, I've been told a few times that we ask users to take care to indent their code when prototyping.

This pull request fixes a few issues where the HTML we ask people to copy was indented incorrectly.

Note that this doesnt fix all whitespace and indent issues, just the ones that will be shown up in the design system website...


36degrees:
Tests aren't passing because some snapshots need updating.

PR #1354, opened 2019-05-16
Spike mixin update override current mixins

nickcolley:
This is a spike of one option for what we could do with the mixins, which is to update the current ones.

I will update this pull request as I go.

Pros:
- Users don't need to update their components in some cases.

Cons:
- Users might not notice an odd behaviour with the new focus styles.

### DIFF
--- a/src/components/accordion/_accordion.scss
+++ b/src/components/accordion/_accordion.scss
@@ -202,7 +202,7 @@
     // We also want to avoid styling when the secttion is pressed or `:active`
     // to avoid the different focus styles from flashing quickly while the user interacts with the section.
     .govuk-accordion__section:not(.govuk-accordion__section--expanded) .govuk-accordion__section-button:focus:not(:active) {
-      @include govuk-focusable-text;
+      @include govuk-focusable-fill-css;
     }
 
     .govuk-accordion__controls {
--- a/src/components/error-summary/_error-summary.scss
+++ b/src/components/error-summary/_error-summary.scss
@@ -13,6 +13,11 @@
     @include govuk-focusable;
 
     border: $govuk-border-width solid $govuk-error-colour;
+
+    &:focus {
+      // Remove `box-shadow` inherited from `:focus`.
+      box-shadow: none;
+    }
   }
 
   .govuk-error-summary__title {
--- a/src/components/error-summary/_error-summary.scss
+++ b/src/components/error-summary/_error-summary.scss
@@ -38,7 +38,7 @@
   }
 
   .govuk-error-summary__list a {
-    @include govuk-focusable-text-link;
+    @include govuk-focusable-fill;
     @include govuk-typography-weight-bold;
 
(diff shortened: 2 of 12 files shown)


src/components/accordion/_accordion.scss
```diff
@@ -202,7 +202,7 @@
     // We also want to avoid styling when the secttion is pressed or `:active`
     // to avoid the different focus styles from flashing quickly while the user interacts with the section.
     .govuk-accordion__section:not(.govuk-accordion__section--expanded) .govuk-accordion__section-button:focus:not(:active) {
-      @include govuk-focusable-text;
+      @include govuk-focusable-fill-css;
```
nickcolley: The original mixins assume you always want the state on a regular `:focus` but you can see here It is useful to be able to customise when to show the focus.

At the moment we'd have to have an additional mixin without the `:focus` selector.

You can also see this in the file upload component.

src/components/error-summary/_error-summary.scss
```diff
@@ -13,6 +13,11 @@
     @include govuk-focusable;
 
     border: $govuk-border-width solid $govuk-error-colour;
+
+    &:focus {
```
nickcolley: Some components we have to override the new focusable state

src/components/header/_header.scss
```diff
@@ -176,7 +176,7 @@
       margin-left: govuk-spacing(1);
     }
 
-    @include govuk-focusable;
+    @include govuk-focusable-fill;
```
nickcolley: See https://github.com/alphagov/govuk-frontend/pull/1346

src/components/skip-link/_skip-link.scss
```diff
@@ -23,5 +22,11 @@
       padding-right: unquote("max(#{govuk-spacing(3)}, #{$padding-safe-area-right})");
       padding-left: unquote("max(#{govuk-spacing(3)}, #{$padding-safe-area-left})");
     }
+
+    &:focus {
```
nickcolley: Skip link is different enough we remove the mixin entirely...

src/helpers/_focusable.scss
```diff
@@ -10,10 +10,18 @@
...
 
+$govuk-focus-width-inner: 2px;
+
 @mixin govuk-focusable {
   &:focus {
     outline: $govuk-focus-width solid $govuk-focus-colour;
     outline-offset: 0;
+    box-shadow: inset 0 0 0 $govuk-focus-width-inner $govuk-text-colour;
+
+    @include govuk-if-ie8 {
```
nickcolley: This feels too specific to the input components, and would likely break anything else.

Perhaps IE8 could use user-agent styles instead?

nickcolley: Radios and checkboxes set the width for their styles so it'd be a bit weird to not do it on inputs.

PR #1356, opened 2019-05-16
SPIKE - Add v2 font

dashouse:
This PR uses v2 of the font.

- Adds new font files, reducing the amount of files from 12 to 8
- Significantly better alignment with the baseline of Helvetica / Arial*
- Tabular numbers are now rolled up in to the Light or Bold weights and set with `font-feature-settings` (CSS applied to existing mixin so `$tabular: true` still works)
- NTA specific overrides on "Button" and "Tag" have been removed

*This is somewhat dependent on font size. Frustratingly for us 19px is still not entirely aligned. 

**Tag before**
before

**Tag after**
after

**Header before** (Note position of Service name - Component may need adjustment of crown and top padding)
Screen Shot 2019-05-16 at 14 05 27

**Header after**
Screen Shot 2019-05-16 at 14 05 38

### Discoveries

- FIXED, see comment below. The header appears to have got larger (65px to 66.16px (1.78%)) suggesting there is a change to how the font takes up space. This was to do with a mixture of the SVG height and vertical position. The header is now 60px regardless of font.

**Overlay of example page to show the difference.**
Note general page scaffolding is intact, images, footer etc in exactly the same position despite text shift.

overlayv2


src/components/button/_button.scss
```diff
@@ -230,20 +230,4 @@
       background-size: 30px 19px;
     }
   }
-
```
nickcolley: I think we'd likely tie the old font to legacy mode, so we'd keep the old stuff around but only when compatibility flags are set.

src/helpers/_font-faces.scss
```diff
@@ -21,47 +21,26 @@
 
       @font-face {
         font-family: "nta";
```
nickcolley: I wonder if there's a chance that since the font-family is named the same thing that a service could accidentally load the old font into GOV.UK Frontend?

Might be safer to update the font-family name... something to test.

src/helpers/_font-faces.scss
```diff
@@ -21,47 +21,26 @@
...
-        src: govuk-font-url("light-2c037cf7e1-v1.eot?#iefix") format("embedded-opentype"),
-             govuk-font-url("light-f38ad40456-v1.woff2") format("woff2"),
-             govuk-font-url("light-458f8ea81c-v1.woff") format("woff");
+        src: govuk-font-url("light-df2b968d71-v2.eot");
+        src: govuk-font-url("light-df2b968d71-v2.eot?#iefix") format("embedded-opentype"),
+             govuk-font-url("light-68f0c84f0e-v2.woff2") format("woff2"),
+             govuk-font-url("light-222368e53d-v2.woff") format("woff");
         font-weight: normal;
         font-style: normal;
         font-display: fallback;
```
nickcolley: Matt recommended to change this to `swap` in his initial work

Nooshu: Yes this still looks like the recommended setting. Google Fonts now allows passing in the `display` parameter and it by default is set to `swap`:

[https://www.zachleat.com/web/google-fonts-display/](https://www.zachleat.com/web/google-fonts-display/)

src/helpers/_typography.scss
```diff
@@ -171,7 +171,13 @@
 
 @mixin govuk-font($size, $weight: regular, $tabular: false, $line-height: false) {
   @if $tabular {
-    @include govuk-typography-common($font-family: $govuk-font-family-tabular);
+    @include govuk-typography-common;
```
nickcolley: I think this might need the compatibility stuff here as well

src/helpers/_font-faces.scss
```diff
@@ -21,47 +21,26 @@
 
       @font-face {
```
nickcolley: I believe that we don't need to have the old fonts around here since GOVUK_Template will supply these?

This font-face is already automatically excluded when certain compat flags are present, something to double check.

src/settings/_typography-font-families.scss
```diff
@@ -9,14 +9,3 @@
...
-/// List of font families to use if using the 'tabular numbers' subset of NTA
-/// (the default font 'stack' for GOV.UK)
-///
-/// Because ntatabularnumbers only includes the digits 0-10, all other glyphs
-/// will 'fall-through' the stack to NTA.
-///
-/// @type List
-/// @access public
-
-$govuk-font-family-nta-tabular: "ntatabularnumbers", $govuk-font-family-nta;
```
nickcolley: Again probably will need to keep this around for the legacy mode...?

src/helpers/_font-faces.scss
dashouse: The old fonts are already gone, unless you mean get rid of the .eot versions etc?

nickcolley: When legacy mode is enabled, GOV.UK Frontend font-face declarations are not output and GOV.UK Frontend relies on GOVUK_Template being loaded alongside.

So this means we don't need to keep the old fonts around in GOV.UK Frontend, but we should check this.

dashouse:
Header issue resolved:

With NTA (60px)
Screen Shot 2019-05-20 at 08 51 07

With Arial (60px)
Screen Shot 2019-05-20 at 08 50 55

GOV.UK's header is 53px (3px padding top is to compensate for transport)
Current header in the design system is 65px (-10 for blue border = 55px)

This change makes the header 60px (50px without border) regardless of font. Which is what I believe to be the desired size.

Nooshu:
I'd recommend removing the EOT versions of the fonts from the `@font-face` declaration. Only IE support this font format (including IE11). All but IE8 also support other font formats that are available. According to the last 30 days GOV.UK analytics, IE8 has had 9,054 "users" out of 50,566,007 (0.018%). 

EOT files can be compressed with Gzip (or Zopfli for 5% better compression), but even with this compression we are asking old browsers on legacy devices to download a fairly large amount of data (60KB) for unhinted versions of the font (only Vista and XP require hinted versions). These browsers and devices are the least equipped to manage these assets.

Font | Size (bytes) | Saving (bytes) | Saving (%)
------------ | ------------- | ------------- | -------------
Bold EOT uncompressed | 58932 | 0 | 0
Bold EOT gzipped | 30366 | 28566 | 48.4
Bold EOT zopfli | 28115 | 30817 | 52.3
Light EOT uncompressed | 58708 | 0 | 0
Light EOT gzipped | 29885 | 28823 | 49.1
Light EOT zopfli | 27589 | 31119 | 53

**Totals**: Uncompressed download - 118KB, Compressed (gzip) - 60KB, Compressed (Zopfli) - 56KB.

I'd say it's a better strategy for this small number of devices is to use the fallback font of Arial, and not have them require to download the additional EOT assets. Since the new font now corrects the glyph placement in the bounding box, the fallback will render correctly without the need for additional adjustments to the vertical placement.

If it is we decide to keep the EOT files in the `@font-face` declaration it's worth noting the container order will prevent IE9-11 from receiving the WOFF version of the font. Browsers pick the first font format it supports from the list. It will ignore any declarations after, (even if they are also supported). In the current setup IE9, 10 and 11 will download the EOT format even though they all support WOFF. 

Since WOFF files are already compressed using Gzip compression, developers don't need to worry about enabling compression on the server. If developers forget to enable Gzip compression on the server for EOT files (since we can't provide compressed versions of the EOT files) these users will be downloading 118KB vs 83KB (WOFF). Due to [all the hacks required](https://www.paulirish.com/2009/bulletproof-font-face-implementation-syntax/) to make fonts work properly in older versions of IE there looks to be no way around this. This seems like another good reason to remove the EOT sources from the declaration.

Recommended declaration
```css
@font-face {
    font-family: "nta";
    src: govuk-font-url("light-68f0c84f0e-v2.woff2") format("woff2"),
         govuk-font-url("light-222368e53d-v2.woff") format("woff");
    font-weight: normal;
    font-style: normal;
    font-display: swap;
}

@font-face {
    font-family: "nta";
    src: govuk-font-url("bold-9561e2d007-v2.woff2") format("woff2"),
         govuk-font-url("bold-5d3836b441-v2.woff") format("woff");
    font-weight: bold;
    font-style: normal;
    font-display: swap;
}
```

nickcolley:
It won't block this work since it's already an issue, but we should consider making the logo entirely an image which includes the crown and the text, so that in browsers that don't have the font they still get the correct branding.

aliuk2012:
This PR will be split into two different PR.

1. Replacing v1 font with v2 (PR #1434)
2. Removing NTA specific overrides on "Button" and "Tag" etc

aliuk2012:
Closing this as #1434 and #1441 have been merged

Issue #1327, opened 2019-05-09
Tabs are not navigable using the keyboard in IE8


36degrees:
The left / right arrow shortcuts to change between tabs are not working in IE8 (as tested using Browserstack, Windows 7).

Because we remove inactive tabs from the tab order, this means you cannot navigate between tabs using the keyboard in IE8.

kellylee-gds:
If this isn't an easy fix we should consider falling back to the non-JS version for IE8

PR #1359, opened 2019-05-17
fixes tabs keyboard navigation bug in IE8

aliuk2012:
In IE8, the browser could not find the next/previous tab because it does
not support `nextElementSibling` and `previousElementSibling` DOM traversal
methods. To fix it I applied a polyfill for it.

Once the browser could find the tab we then had to find the `firstChildElement`
(i.e the anchor element) to add/edit the various data attributes to show/hide
the tab panel. Again IE8 doesn't support it and instead of introducing another
polyfill I used `querySelector instead to look up the "a". I assumed the
first element would always be anchor for navigation purposes and also the
nunjucks template uses an anchor with the class name that I'm looking up.

### Before:

ie8-tab-nav-bug

### After
ie8-tab-nav-fix

fixes: #1327 
👉 https://govuk-frontend-review-pr-1359.herokuapp.com/components/tabs
👉 https://govuk-frontend-review-pr-1359.herokuapp.com/components/tabs/preview

### DIFF
--- a/src/components/tabs/tabs.js
+++ b/src/components/tabs/tabs.js
@@ -204,10 +204,25 @@ Tabs.prototype.onTabKeydown = function (e) {
 }
 
 Tabs.prototype.activateNextTab = function () {
+  // IE doesn't support 'nextElementSibling' this code polyfills IE8
+  // https://developer.mozilla.org/en-US/docs/Web/API/NonDocumentTypeChildNode/nextElementSibling#Polyfill_for_Internet_Explorer_8
+  // Source: https://github.com/Alhadis/Snippets/blob/master/js/polyfills/IE8-child-elements.js
+  if (!('nextElementSibling' in document.documentElement)) {
+    Object.defineProperty(Element.prototype, 'nextElementSibling', { // eslint-disable-line no-undef
+      get: function () {
+        var e = this.nextSibling
+        while (e && e.nodeType !== 1) {
+          e = e.nextSibling
+        }
+        return e
+      }
+    })
+  }
+
   var currentTab = this.getCurrentTab()
   var nextTabListItem = currentTab.parentNode.nextElementSibling
   if (nextTabListItem) {
-    var nextTab = nextTabListItem.firstElementChild
+    var nextTab = nextTabListItem.querySelector('.govuk-tabs__tab')
   }
   if (nextTab) {
     this.hideTab(currentTab)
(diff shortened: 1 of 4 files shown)


src/components/tabs/tabs.js
```diff
@@ -218,10 +233,25 @@ Tabs.prototype.activateNextTab = function () {
...
+        while (e && e.nodeType !== 1) {
+          e = e.previousSibling
+        }
+        return e
+      }
+    })
+  }
+
   var currentTab = this.getCurrentTab()
   var previousTabListItem = currentTab.parentNode.previousElementSibling
```
nickcolley: out of curiosity, if we changed `previousElementSibling` to `previousSibling` would this resolve the issue?

src/components/tabs/tabs.js
```diff
@@ -218,10 +233,25 @@ Tabs.prototype.activateNextTab = function () {
 }
 
 Tabs.prototype.activatePreviousTab = function () {
+  // IE doesn't support 'previousElementSibling' this code polyfills IE8
```
nickcolley: If we do go down the polyfill route, this is how we tend to do it: https://github.com/alphagov/govuk-frontend/blob/master/docs/contributing/polyfilling.md

src/components/tabs/tabs.js
aliuk2012: Nice one, I can remove the polyfill 😄

36degrees: Should test this change carefully – I wasn't previously aware of the difference but it looks like next/previousSibling [may return the whitespace between elements](https://stackoverflow.com/questions/24226571/what-is-the-difference-between-node-nextsibling-and-childnode-nextelementsibling)

aliuk2012: Yes I did run into so problems while testing and its also one of the reason the automated tests were failing too. I have since reintroduced the polyfill.

src/vendor/polyfills/Element/prototype/previousElementSibling.js
```diff
@@ -0,0 +1,43 @@
...
+(function(undefined) {
+
+    // Detection from https://github.com/Financial-Times/polyfill-service/pull/1062/files#diff-a162235fbc9c0dd40d4032265f44942e
+    var detect = (
+      'Element' in this && 'previousElementSibling' in document.documentElement
+    )
+
+    if (detect) return
+
+    (function (global) {
```
nickcolley: I think this define bit might not be needed since you're importing it above.

PR #1367, opened 2019-05-21
Allow for classes, rowspan, colspan and attributes on row headers

### LINKED ISSUE #1261: Can't pass classes to first item in row if first cell is header
If the `firstCellIsHeader` perimeter is true, classes on that row item aren't passed to the element.

I think it just needs `{%- if cell.classes %} {{ cell.classes }}{% endif %}` added to the `th` to get it to work?

36degrees:
This change allows you to pass classes, rowspan, colspan and attributes to the first cell in each row when `firstCellIsHeader` is `true`.

~~It also simplifies the logic of the table template, and tidies up the tests to use a single assertion per test, improve coverage and use a consistent format.~~ (Split into a future PR)

Fixes #1261

### DIFF
--- a/src/components/table/template.njk
+++ b/src/components/table/template.njk
@@ -21,14 +21,21 @@
     {% for row in params.rows %}
     <tr class="govuk-table__row">
     {% for cell in row %}
+      {% set commonAttributes %}
+        {%- if cell.colspan %} colspan="{{ cell.colspan }}"{% endif %}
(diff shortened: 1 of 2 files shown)


36degrees:
Happy to split the last two commits out into a separate PR if you prefer?

nickcolley:
I think it's up to you depending on how confident you are about the new tests.

36degrees:
:+1: I've split that work out and will raise another PR once this has been merged.

PR #1368, opened 2019-05-21
Clarify productName vs serviceName

### LINKED ISSUE #1286: Incorrect description for serviceName option in header component
The header component Nunjucks macro options description has the same text repeated for `productName` and `serviceName` - I suspect service name is meant to have text talking about common service names?

Screenshot 2019-03-21 at 10 07 21

36degrees:
At the minute the serviceName description incorrectly duplicates the productName description.

In most cases, we expect people to use service name, so this is an attempt to clarify what each is for and when to use them.

Fixes #1286

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -242,6 +242,13 @@
 
   ([PR #1310](https://github.com/alphagov/govuk-frontend/pull/1310))
 
+- Fixes the description for serviceName option in header component (used in the
+  GOV.UK Design System)
+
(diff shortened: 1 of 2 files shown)


edwardhorsford:
I know this is merged, and I don't know if it's helpful - but my service has opted to use `productName` rather than `serviceName` - at least in part because we've got a noun based name rather than a verb based name. When noun based, it sits better next to gov.uk - see Digital Marketplace, Verify, etc.

alex-ju:
Arguably Verify and Pay are verbs.. they all become common nouns if we use them to name products, but they are verbs at origin.

Issue #1218, opened 2019-02-22
Namespacing [data-module] to avoid conflicts


nickcolley:
One of the main aims for GOV.UK Frontend is to ensure we do not break services code.

Namespacing is a big part of this, by prefixing certain parts of the codebase with `govuk-` and encouraging services to use `departmentName-` or `app-` we can have more confidence that we will not break their code.

GOV.UK Publishing ran into an issue where their version of accordion code was conflicting with ours: https://github.com/alphagov/collections/pull/1042

I've raised this issue to discuss the impact of this and if we should consider changing the namespace.

Changing the namespace without keeping backwards compatibility would be a breaking change, as always we need to weigh up the risk of not doing this with the risk of making it too hard for people to upgrade in the future.

aliuk2012:
Regarding HTML data attributes there are potentially two options:

`data-govuk-module="accordion"` or `data-module="govuk-accordion`

I prefer the first one because I think we should namespace the key not the value. Its also easier to find with css attribute selectors `[data-govuk-module]` vs `[data-module="govuk-modules"]`

nickcolley:
The way I think about our current setup is it is a halfway house towards custom elements (web components).

In that world, you register a component with a single string so in my mind [data-module="govuk-accordion"] aligns closest to that.

nickcolley:
We asked the community about this idea and it has been accepted: https://github.com/alphagov/govuk-design-system-architecture/pull/16

Rendered: https://github.com/alphagov/govuk-design-system-architecture/blob/master/proposals/006-initialise-javascript-using-govuk-namespaced-data-module-attributes.md

aliuk2012:
Closed #1443

PR #1370, opened 2019-05-22
Set debounceFormSubmitTimer at instance level instead of global

alex-ju:
When there are multiple form submit buttons on the page with `prevent-double-click` enabled they will all reference to the same `debounceFormSubmitTimer` which means that in a sequence of actions only the first one will be called causing automated end-to-end tests to fail and potentially preventing 'quick' users to be efficient.

In this PR we're attaching the `debounceFormSubmitTimer` to each instance, so a specific action is stopped from being called multiple times without interfering with other submit buttons on the page.

This issue was found while implementing the `prevent-double-click` feature in govuk_publishing_components so we [reverted the change](https://github.com/alphagov/govuk_publishing_components/pull/865) hoping to fix this at the root.

P.S. I'm aware WCAG2.1 is a priority for the team now, but I want to add this to the backlog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -508,6 +508,13 @@
 
 [compatibility mode]: https://github.com/alphagov/govuk-frontend/blob/master/docs/installation/installing-with-npm.md#compatibility-mode
 
+- Allow distinct buttons with `prevent-double-click` enabled to be clicked subsequently within one second
+
+  For buttons with `prevent-double-click` enabled we set the `debounceFormSubmitTimer` for each instance, so a specific action is stopped from being called multiple times without interfering with other submit buttons on the page.
+
+  ([PR #1370](https://github.com/alphagov/govuk-frontend/pull/1370))
+
+
 ## 2.11.0 (Feature release)
 
 🆕 New features:
--- a/src/components/button/button.js
+++ b/src/components/button/button.js
@@ -1,11 +1,12 @@
 import '../../vendor/polyfills/Event' // addEventListener and event.target normaliziation
+import '../../vendor/polyfills/Function/prototype/bind'
 
 var KEY_SPACE = 32
 var DEBOUNCE_TIMEOUT_IN_SECONDS = 1
-var debounceFormSubmitTimer = null
 
 function Button ($module) {
   this.$module = $module
+  this.debounceFormSubmitTimer = null
 }
 
 /**
@@ -41,14 +42,14 @@ Button.prototype.debounce = function (event) {
   }
 
   // If the timer is still running then we want to prevent the click from submitting the form
-  if (debounceFormSubmitTimer) {
+  if (this.debounceFormSubmitTimer) {
     event.preventDefault()
     return false
   }
 
-  debounceFormSubmitTimer = setTimeout(function () {
-    debounceFormSubmitTimer = null
-  }, DEBOUNCE_TIMEOUT_IN_SECONDS * 1000)
+  this.debounceFormSubmitTimer = setTimeout(function () {
+    this.debounceFormSubmitTimer = null
+  }.bind(this), DEBOUNCE_TIMEOUT_IN_SECONDS * 1000)
 }
 
 /**
--- a/src/components/button/button.test.js
+++ b/src/components/button/button.test.js
@@ -104,6 +104,34 @@ describe('/components/button', () => {
 
         expect(submitCount).toBe(2)
       })
+      it('does not prevent subsequent clicks on different buttons', async () => {
+        await page.goto(baseUrl + '/components/button/prevent-double-click/preview', { waitUntil: 'load' })
+
+        // Our examples don't have form wrappers so we need to overwrite it.
+        await page.evaluate(() => {
+          const $button = document.querySelector('button')
+          const $buttonPrime = $button.cloneNode()
+          const $form = document.createElement('form')
+          $button.parentNode.appendChild($form)
+          $button.parentNode.removeChild($button)
+          $form.appendChild($button)
+          $form.appendChild($buttonPrime)
+
+          window.__SUBMIT_EVENTS = 0
+          $form.addEventListener('submit', event => {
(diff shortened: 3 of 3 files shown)


src/components/button/button.js
```diff
@@ -47,14 +47,14 @@ Button.prototype.debounce = function (event) {
...
     event.preventDefault()
     return false
   }
 
-  debounceFormSubmitTimer = setTimeout(function () {
-    debounceFormSubmitTimer = null
-  }, DEBOUNCE_TIMEOUT_IN_SECONDS * 1000)
+  this.debounceFormSubmitTimer = setTimeout(function () {
+    this.debounceFormSubmitTimer = null
+  }.bind(this), DEBOUNCE_TIMEOUT_IN_SECONDS * 1000)
```
nickcolley: We need to import the polyfill for bind here I think.

nickcolley:
Hmm actually on second thoughts because the button component is instantiated at the document level, does this actually fix the issue? @alex-ju 🤔

alex-ju:
good point, I forgot we're initialising it at the document level. I may need to change that as I don't see other options

src/all.js
```diff
@@ -17,8 +17,11 @@ function initAll (options) {
   // Defaults to the entire document if nothing is set.
   var scope = typeof options.scope !== 'undefined' ? options.scope : document
 
-  // Find all buttons with [role=button] on the scope to enhance.
-  new Button(scope).init()
+  // Find all links with role="button" and all buttons to enhance.
+  var $buttons = scope.querySelectorAll('a[role="button"], button')
```
nickcolley: I think you could remove the `a` since `[role="button"]` on it's own would be fine.

Would this need to include `input` elements that are buttons?

When I initially did this work I was worried that we'd end up initialising this JavaScript on a massive amount of buttons if it were on a case working system screen for example.

Should we be thinking about the performance impact of this?

nickcolley:
@alex-ju for reference I did explore this way of implementing the feature originally: https://github.com/alphagov/govuk-frontend/compare/debounce-button-refactored?expand=1

See the discussion in the original pull request: https://github.com/alphagov/govuk-frontend/pull/1018#pullrequestreview-207410175

src/all.js
alex-ju: Right, input is a valid case too, I've updated to contain it.

I share your concerns, but the current solution (which I consider smart) doesn't seem to work in all cases. And thanks for sharing the initial work, seems to be very considerate.

Would you mind dismissing the initial review, since the following updates are slightly expanding outside the initial scope?

nickcolley: We could also consider removing the feature, not likely we'd want to do this but it's an option.

CHANGELOG.md
```diff
@@ -4,6 +4,27 @@
 
 💥 Breaking changes:
 
+- Set `debounceFormSubmitTimer` at instance level instead of global
+
+  For buttons with `prevent-double-click` enabled we set the `debounceFormSubmitTimer` to each instance, so a specific action is stopped from being called multiple times without interfering with other submit buttons on the page.
+
+  To migrate you need to update the way you initialise the Button components from:
```
36degrees: This also means that if any buttons are added to the page _after_ the module has been initialised they'll no longer get this behaviour without being initialised separately.

Are we happy with that change? If so, can we be explicit about it in the changelog?

src/components/button/button.js
```diff
@@ -47,14 +48,14 @@ Button.prototype.debounce = function (event) {
...
-  if (debounceFormSubmitTimer) {
+  if (this.debounceFormSubmitTimer) {
     event.preventDefault()
     return false
   }
 
-  debounceFormSubmitTimer = setTimeout(function () {
-    debounceFormSubmitTimer = null
-  }, DEBOUNCE_TIMEOUT_IN_SECONDS * 1000)
+  this.debounceFormSubmitTimer = setTimeout(function () {
```
36degrees: Could we write a test to cover the behaviour we want from this change?

CHANGELOG.md
alex-ju: That is right and it's missing from the changelog. Will update to reflect this. It is however aligned with how the rest of the components enhanced with JavaScript are being initialised.

nickcolley:
Just to update, we will be doing changes to namespacing as part of 3.0.0.

We met and discussed the implementation here, and we think we can do a safer implementation alongside the updates we'll have to make for [namespacing the data-module syntax](https://github.com/alphagov/govuk-frontend/issues/1218).

We're going to [write up an architecture proposal](https://github.com/alphagov/govuk-design-system-architecture) with the full details.

alex-ju:
Thanks for the update, @nickcolley! Sounds like a sensible approach.

nickcolley:
Now that https://github.com/alphagov/govuk-frontend/pull/1443 is merged, this work will need to be adjusted and rebased.

@alex-ju would you like to do this? I am happy to pick this up and keep your existing work, to finish it off if you do not have time. :)

alex-ju:
@nickcolley thanks for the work on #1443. I'll update this one. Will need you around for a second review if that' ok with you.

nickcolley:
@alex-ju of course :)

src/components/button/button.js
alex-ju: Test added

CHANGELOG.md
```diff
@@ -358,6 +358,12 @@
 
   ([PR #1443](https://github.com/alphagov/govuk-frontend/pull/1443))
 
+- Set `debounceFormSubmitTimer` at instance level instead of global
```
nickcolley: Should this be in the 'fix' category?

Could we name the title so it's clear what we're fixing rather than how we fixed it?

nickcolley:
Will need a second review too and we can get it merged.

CHANGELOG.md
alex-ju: Makes sense, now it became a fix. Updated title, good point about it.

Issue #888, opened 2018-07-12
Can we make it an easier user experience when beforeContent is empty?


joelanman:
In this situation, the spacing at the top of the page is wrong.

The current solution is to add this to your page:

```{% set mainClasses = "govuk-main-wrapper--l" %}```

Wondering if we can do anything that makes this simpler/clearer or that, ideally, our users don't need to do anything

joelanman:
We could make the beforeContent block have default content that would provide the correct spacing:

```{% block beforeContent %}<div class="empty-before-content"></div>{% endblock %}```

nickcolley:
If we contributed the 'self' feature to the Nunjucks project we could do this in a non-hacky way:

https://github.com/mozilla/nunjucks/issues/164

36degrees:
I wonder if we could do this in CSS instead, perhaps by applying an additional margin to main if it's the `:first-child` (and thus has no preceding siblings)

joelanman:
Is this arguably a bug, if a user doesn't know to do the 'right' thing, it will be wrong?

nickcolley:
We can definitely make it better and I hope we do, but the current feature works as we designed it to, so it's not a bug.

36degrees:
I believe this was addressed in #1371, so I'm going to close it.

PR #1371, opened 2019-05-22
Simplify main wrapper spacing logic

nickcolley:
By using a `:first-child` selector we can avoid the need for a modifier class, which is often missed.

This also removes the mixins and modifier class, as there's no clear need for these things.

This work would also allow us to simplify the guidance around main wrapper, since the modifier class is no longer needed: https://design-system.service.gov.uk/styles/layout/#page-wrappers

You can compare pages, which used to require the modifier class which no longer do.

- [Confirm delete page, current v3](https://govuk-frontend-v3.herokuapp.com/full-page-examples/confirm-delete)
- [Confirm delete page, this pull request](https://govuk-frontend-review-pr-1371.herokuapp.com/full-page-examples/confirm-delete)

Closes https://github.com/alphagov/govuk-frontend/issues/888

Update: I've changed the approach based on the feedback so it's not a breaking change.

It includes deprecations instead tracked by this issue: https://github.com/alphagov/govuk-frontend/issues/1379

We still may want to update the guidance but it might not be as drastic: https://github.com/alphagov/govuk-design-system/issues/887

### DIFF
--- a/src/objects/_main-wrapper.scss
+++ b/src/objects/_main-wrapper.scss
@@ -50,6 +50,10 @@
     @include govuk-main-wrapper;
   }
 
+  // For most instances of main wrapper, :first-child will give the correct spacing without
+  // the need for the `l` modifier, but it is available in instances where there may be
+  // empty elements that would be difficult to remove.
+  .govuk-main-wrapper:first-child,
   .govuk-main-wrapper--l {
     @include govuk-main-wrapper--l;
   }
--- a/src/objects/_main-wrapper.scss
+++ b/src/objects/_main-wrapper.scss
@@ -20,6 +20,7 @@
 // </div>
 
 
+/// @deprecated Replace this mixin with more direct references to the [spacing mixins](https://design-system.service.gov.uk/styles/spacing/#spacing-on-custom-components).
 @mixin govuk-main-wrapper {
   // In IE11 the `main` element can be used, but is not recognized  –
   // meaning it's not defined in IE's default style sheet,
@@ -37,7 +38,9 @@
   }
 }
 
-// Use govuk-main-wrapper--l when you page does not have Breadcrumbs, phase banners or back links
+/// Use govuk-main-wrapper--l when you page does not have Breadcrumbs, phase banners or back links.
+///
+/// @deprecated Replace this mixin with more direct references to the [spacing mixins](https://design-system.service.gov.uk/styles/spacing/#spacing-on-custom-components).
 @mixin govuk-main-wrapper--l {
   @include govuk-responsive-padding(8, "top");
 }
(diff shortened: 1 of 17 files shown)


src/objects/_main-wrapper.scss
```diff
@@ -2,52 +2,27 @@
...
-// Example usage without Breadcrumbs, phase banners, back links:
-// <div class="govuk-width-container">
-//   <div class="govuk-main-wrapper govuk-main-wrapper--l">
-//       <!-- Wrapper for the main content of your page which applies padding
-//            to the top / bottom -->
-//   </div>
-// </div>
-
-
-@mixin govuk-main-wrapper {
```
nickcolley: Maybe I'm being too trigger happy and we should just leave these, I'm just not sure why they're part of the public api.

src/objects/_main-wrapper.scss
```diff
@@ -2,52 +2,27 @@
...
+
+    @include govuk-media-query($from: tablet) {
+      // This spacing is manually adjusted to replicate the margin of
+      // govuk-heading-xl (50px) minus the spacing of back link and
+      // breadcrumbs (10px)
+      padding-top: govuk-spacing(7);
+      padding-bottom: govuk-spacing(7);
+    }
 
-  .govuk-main-wrapper--l {
```
nickcolley: We could in theory leave this modifier around, there might be cases where it could be useful?

My reasoning for removing it is, that this modifier should only be applied when the main wrapper is the first child. cc @dashouse is this correct?

nickcolley:
We could consider doing this after v3.0.0 and deprecating things rather than adding this as a breaking change, but I figured the risk was low since migration is automatic?

src/objects/_main-wrapper.scss
36degrees: Yeah, I'm not sure either…

By the looks of it, I added the mixins in https://github.com/alphagov/govuk-frontend/commit/4c01202842352f6f8488065cab4d9dd1db976b1c but I've no idea why.

src/objects/_main-wrapper.scss
36degrees: I guess there might be edge cases where for _reasons_ a service ends up with an empty or non-visible element before the main wrapper? For example, a script tag, or an empty container that's always there even if there's no phase banner / breadcrumbs / whatever.

I _think_ I'd be tempted to keep the class just so that there's an option available in those cases?

dashouse: Based on existing implementations of Alerts (https://github.com/alphagov/govuk-design-system-backlog/issues/2) I would second keeping the class

nickcolley: Makes sense to me lets keep it.

nickcolley:
Could get this into 2.12.0 if we were so inclined...

PR #1372, opened 2019-05-22
Support custom attributes on summary list action links

colinrotherham:
We've spotted a requirement in our service to add attributes to the "Check Answers" action links.

I've added support via this pull request. Thanks!

```js
{{ govukSummaryList({
  rows: [
    {
      actions: {
        items: [
          {
            text: 'Edit',
            attributes: {
              'data-attribute-1': 'ABC',
              'data-attribute-2': '123'
            }
          }
        ]
      }
    }
  ]
}) }}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,12 @@
 
 🆕 New features:
 
+- Support custom attributes on summary list action links
+
+  You can now use the `attributes` macro option to add additional HTML attributes to summary list action links.
+
+  ([PR #1372](https://github.com/alphagov/govuk-frontend/pull/1372))
+
 - Support aria-describedby on all form fields
 
   All form fields now support an initial `aria-describedby` value, populated before the optional hint and error message IDs are appended.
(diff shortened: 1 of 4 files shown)


src/components/summary-list/template.njk
```diff
@@ -1,5 +1,5 @@
 {%- macro _actionLink(action) %}
-  <a class="govuk-link {%- if action.classes %} {{ action.classes }}{% endif %}" href="{{ action.href }}">
```
nickcolley: This is looking really good, do we need to add some documentation in for this new option? In the 'yaml' file?

CHANGELOG.md
```diff
@@ -14,6 +14,12 @@
 
 🆕 New features:
 
+- Support custom attributes on summary list action links
+
+  The `attributes: {}` pattern can now be used to add `id` or `data-attr` attributes to summary list action links.
```
36degrees: Somewhat minor, but this is oddly specific in that it makes it sound a little bit like attributes can only be used for `id` / `data-*` attributes, when actually it can be used for any attributes.

How about something like:

```suggestion
  You can now use the `attributes` macro option to add additional HTML attributes to summary list action links.
```

?

src/components/summary-list/template.njk
colinrotherham: Ah good point! Added

CHANGELOG.md
colinrotherham: Sounds good to me, I've added it

colinrotherham:
@36degrees @nickcolley All sorted. Thank for the quick review

PR #1373, opened 2019-05-22
Document table improvements in CHANGELOG

36degrees:
This was accidentally removed from #1367 when I removed the test refactoring commits.


aliuk2012:
Good spot

Issue #1362, opened 2019-05-20
Tab has bevelled edge


dashouse:
Because the selected tab has the blue (box-shadow) border style applied it does not have the grey border the top.

This implementation has kept the border set to transparent, creating an angled edge where the borders join (doesn't matter that one of them is invisible).

Ideally this would be set to 0px width to remove this effect, however I'm aware that for users that override their colours or can't see the box shadow the selected tab would be open topped.

<img width="396" alt="Screen Shot 2019-05-13 at 15 57 18" src="https://user-images.githubusercontent.com/23356842/58026794-2ced8400-7b0f-11e9-9002-c72465e205a5.png">

Is there anything we can do to keep the clean look of the 0px border-top while still retaining a good looking tab when overriding or inverting colours?

36degrees:
Not quite what you're asking for, but adding a 1px blue border and removing 1px from the box-shadow gets _closer_…

<img width="348" alt="Screen Shot 2019-05-21 at 15 27 35" src="https://user-images.githubusercontent.com/121939/58104671-10b81880-7bdd-11e9-8d20-0ebfe3a61f6d.png">

😕

timpaul:
Does this help?

https://jsfiddle.net/ThinkingStiff/TvGC5/

36degrees:
Unfortunately not, because when a user has customised their colours box-shadows become invisible, which means rather than looking like this:

![Screenshot_2019-05-21 Stackoverflow - set css border to end in a 90 instead of a 45 degree angle](https://user-images.githubusercontent.com/121939/58110902-5201f580-7be8-11e9-88ed-f5bde7d2ac9c.png)

We end up with this:

![Screenshot_2019-05-21 Stackoverflow - set css border to end in a 90 instead of a 45 degree angle(1)](https://user-images.githubusercontent.com/121939/58110912-562e1300-7be8-11e9-9ab0-f2259bc797f6.png)

And in the case of our tabs, we end up with something like this:

<img width="1019" alt="Screen Shot 2019-05-21 at 16 50 02" src="https://user-images.githubusercontent.com/121939/58110999-7d84e000-7be8-11e9-8475-fda33635dba2.png">

nickcolley:
I managed to get it 90% working by using floating pseudo elements, but I got proper into the weeds on it and got scared...

gunjam:
How about this? 😅
https://codepen.io/anon/pen/eaeybB

gunjam:
[...]

PR #1375, opened 2019-05-22
Fix tab bevelled edge

nickcolley:
Update the tabs component to use an absolutely positioned element above the tab, which allows for 90 degree crisp rendering.

# Screenshots

| Before | After |
| - | - |
|  |  |

## Focused

| Before | After |
| - | - |
|  |  |

## Overridden colours

| Before | After |
| - | - |
|  |  |

## Internet Explorer 8

| Before | After |
| - | - |
|  |  |

- [x] Cross browser testing: IE 8-11, Edge, Firefox, Safari, Chrome (latest for all)

Fixes https://github.com/alphagov/govuk-frontend/issues/1362

### DIFF
--- a/src/components/button/_button.scss
+++ b/src/components/button/_button.scss
@@ -106,37 +106,40 @@
       box-shadow: 0 2px 0 $govuk-focus-text-colour;
     }
 
-    // The following adjustments do not work for <input type="button"> as
-    // non-container elements cannot include pseudo elements (i.e. ::before).
+    // Internet Explorer 8 does not support box-shadow, which this CSS is used for.
+    @include govuk-not-ie8 {
+      // The following adjustments do not work for <input type="button"> as
+      // non-container elements cannot include pseudo elements (i.e. ::before).
 
-    // Use a pseudo element to expand the click target area to include the
-    // button's shadow as well, in case users try to click it.
-    &::before {
-      content: "";
-      display: block;
+      // Use a pseudo element to expand the click target area to include the
+      // button's shadow as well, in case users try to click it.
+      &::before {
+        content: "";
+        display: block;
 
-      position: absolute;
+        position: absolute;
 
-      top: -$govuk-border-width-form-element;
-      right: -$govuk-border-width-form-element;
-      bottom: -($govuk-border-width-form-element + $button-shadow-size);
-      left: -$govuk-border-width-form-element;
+        top: -$govuk-border-width-form-element;
+        right: -$govuk-border-width-form-element;
+        bottom: -($govuk-border-width-form-element + $button-shadow-size);
+        left: -$govuk-border-width-form-element;
 
-      background: transparent;
-    }
+        background: transparent;
+      }
 
-    // When the button is active it is shifted down by $button-shadow-size to
-    // denote a 'pressed' state. If the user happened to click at the very top
-    // of the button, their mouse is no longer over the button (because it has
-    // 'moved beneath them') and so the click event is not fired.
-    //
-    // This corrects that by shifting the top of the pseudo element so that it
-    // continues to cover the area that the user originally clicked, which means
-    // the click event is still fired.
-    //
-    // 🎉
-    &:active::before {
-      top: -($govuk-border-width-form-element + $button-shadow-size);
+      // When the button is active it is shifted down by $button-shadow-size to
+      // denote a 'pressed' state. If the user happened to click at the very top
+      // of the button, their mouse is no longer over the button (because it has
+      // 'moved beneath them') and so the click event is not fired.
+      //
+      // This corrects that by shifting the top of the pseudo element so that it
+      // continues to cover the area that the user originally clicked, which means
+      // the click event is still fired.
+      //
+      // 🎉
+      &:active::before {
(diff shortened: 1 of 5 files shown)


src/components/tabs/_tabs.scss
```diff
@@ -90,57 +91,73 @@
...
-          $top-border-adjustment: 1px;
-
-          @include govuk-if-ie8 {
-            // IE8 already has a top border so no adjustment is needed
-            $top-border-width: 4px;
-            $top-border-adjustment: 0;
-          }
-
-          margin-top: - govuk-spacing(1);
+          margin-top: - (govuk-spacing(1) - 1px);
```
nickcolley: I've had to adjust the spacing since we now always have a border.

src/components/tabs/_tabs.scss
```diff
@@ -90,57 +91,78 @@
...
+          border: 1px solid $govuk-border-colour;
+          border-bottom: 0;
 
           color: $govuk-text-colour;
           background-color: govuk-colour("white");
-          box-shadow: inset 0 4px 0 0 $govuk-link-colour;
           text-decoration: none;
 
-          @include govuk-if-ie8 {
+          &:after {
```
nickcolley: I've had to use `:after` instead of `::after` since Internet Explorer 8 does not support the non deprecated selector, anyone know why our build steps don't convert these automatically? 🤔

hannalaakso: I *think* we currently only do prefixing with postcss. We'd probably want this: https://github.com/timelsass/postcss-pseudo-element-colons

src/components/tabs/_tabs.scss
```diff
@@ -90,57 +91,78 @@
...
-            // IE8 already has a top border so no adjustment is needed
-            $top-border-width: 4px;
-            $top-border-adjustment: 0;
-          }
-
-          margin-top: - govuk-spacing(1);
+          margin-top: - (govuk-spacing(1) - 1px);
           margin-bottom: -1px;
 
-          // 1px is compensation for border (otherwise we get a 1px shift)
```
hannalaakso: Non-blocking: this comment still feels useful (maybe minus the reference to 1px) for why we're reducting pixels from the padding.

src/components/tabs/_tabs.scss
```diff
@@ -90,57 +91,78 @@
...
+        //
+        &:after {
+          content: "";
+          position: absolute;
+          top: -1px;
+          right: 0;
+          left: 0;
+          border-width: 0;
+          border-style: solid;
+          border-color: transparent;
```
hannalaakso: Totally not blocking but this could be shortened to 
```
border: 0 solid transparent;
```

src/components/tabs/_tabs.scss
36degrees: The 'oldie' plugin already supports this, but for some reason [it's disabled](https://github.com/alphagov/govuk-frontend/blob/master/tasks/gulp/compile-assets.js#L78).

src/components/tabs/_tabs.scss
nickcolley: This comment no longer applies since the border is always around.

src/components/tabs/_tabs.scss
nickcolley: Probably more consistent with the rest of the codebase so I'll change it.

src/components/tabs/_tabs.scss
nickcolley: Looks like we may have turned this off so that the radios/checkboxes component can fallback easier... https://github.com/alphagov/govuk-frontend/blame/fdde74e1887b81e3419cd82dc55dc793398117d3/tasks/gulp/compile-assets.js

We might be able to guard this CSS with the IE8 mixins and turn this back on...

src/components/button/_button.scss
```diff
@@ -123,6 +123,11 @@
       left: -$govuk-border-width-form-element;
 
       background: transparent;
+
+      // IE8 doesn’t support box-shadow
+      @include govuk-if-ie8 {
```
36degrees: Rather than 'setting' and then 'unsetting', would it make more sense to wrap the whole `::before` block in `govuk-not-ie8`? That way the IE8 stylesheet doesn't end up with lots of rules that it doesn't use, and hopefully it's easier to follow? Thoughts?

nickcolley: I did think of this, but to be super honest I just didn't want to take the risk so took the easy way out 😬

nickcolley: I'll just wrap these things 😇

36degrees:
Needs a changelog entry.

I think we should also list the changes to postcss separately, as the changes to _our_ config only affect the `dist` build? As far as I can tell, anyone currently building their own IE8 stylesheets will need to be using postcss with the same config as us in order for tabs to render correctly in IE8?

nickcolley:
@36degrees how would you feel if I raise an issue to document that on the IE8 page?

nickcolley:
Closing this for now since we'll be changing the focus state, but will take parts of this once the focus work is done.

PR #1377, opened 2019-05-23
Add latest package-lock

hannalaakso:
Even though we're all using the same versions on npm and node, there seems to be some back and forth with the sha's in `package-lock`. Committing the latest version we now get in this branch. 

I'm running npm v6.4.1 and node v10.15.1.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -573,7 +573,7 @@
     "abbrev": {
       "version": "1.1.1",
       "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.1.tgz",
-      "integrity": "sha1-+PLIh60Qv2f2NPAFtph/7TF5qsg=",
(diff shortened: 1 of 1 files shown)


colinrotherham:
Hi @hannalaakso,

Try changing all the references of `npm install` in the various docs to `npm ci` (unless installing a specific dependency)? You'll get a nice clean install via the lockfile only.

It usefully stops each new version of npm rewriting it with changes like this.

I checked, and it was my **npm@6.9.0** version that added all the `"optional": true` entries, sorry! 😭 

I'd followed the npm recommendation to use the latest version:
https://docs.npmjs.com/about-npm-versions#the-latest-release-of-npm

hannalaakso:
Thanks @colinrotherham for the `npm ci` suggestion, we should look into it. I'm not going to be around for a bit so going to leave this open for other people in the team to review or close if appropriate. I was keen to clear the decks for https://github.com/alphagov/govuk-frontend/pull/1378 to avoid committing a `package-lock` with unrelated changes in that PR but I've now pushed up a `package-lock` to it.

hannalaakso:
Cleanup: I'm going to close this as we update this in https://github.com/alphagov/govuk-frontend/pull/1378

Issue #1360, opened 2019-05-17
Conditionally include legacy scripts in review app


hannalaakso:
Following on from the work done in https://github.com/alphagov/govuk-frontend/pull/1279, we should also conditionally load in the Javascript from GOV.UK Template and Frontend Toolkit to ensure that they don't conflict with the GOV.UK Frontend scripts.

This would help us spot issues such as https://github.com/alphagov/govuk-frontend/issues/1270.

hannalaakso:
Closed by https://github.com/alphagov/govuk-frontend/pull/1378

Issue #1270, opened 2019-04-10
Header component script is not compatible with govuk_template


alex-ju:
### Context
govuk_template has [an IIFE for header navigation toggle](https://github.com/alphagov/govuk_template/blob/f164b2825b1479d7c7d4bdbec4ba9c8e171f504c/source/assets/javascripts/core.js#L11) which adds functionality to elements that have a `.js-header-toggle` class attached to it.

The header component in govuk-frontend [uses the same `.js-header-toggle` class](https://github.com/alphagov/govuk-frontend/blob/master/src/components/header/header.js#L16) (applied to a button element) to toggle the header.

### Problem
Because the govuk_template script is expecting an anchor instead of a button, it tries to get the `href` attribute and fails to do so if the header component from govuk-frontend is used.

<img width="459" alt="Screen Shot 2019-04-10 at 15 37 29" src="https://user-images.githubusercontent.com/788096/55888136-c471c580-5ba6-11e9-86af-e5fb7ce37e06.png">

### Potential solution
Prefix or change `js-` classes to avoid clashed with other scripts (including the deprecated ones).

I'm happy to raise a PR myself since I think I introduced this bit of code, but I need a decision at the team level on what the solution should be.

nickcolley:
Adding this to the 3.0.0 milestone as a point of discussion, potentially we should be reviewing more than just this class and considering if they should be using a 'js-govuk' prefix.

nickcolley:
We have decided to change our uses of 'js-' to be 'govuk-js-' as per @hannalaakso 's suggestion.

aliuk2012:
Should be fixed in v3.0.0. Closing this as #1444 has been merged.

PR #1378, opened 2019-05-23
Include legacy scripts in review app legacy mode

hannalaakso:
This PR includes the GOV.UK Template and frontend toolkit scripts on the page in the legacy mode.

Specifically, it 
- Includes [`GOVUK.modules.start()`](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/docs/javascript.md#usage) which initiates any scripts with `start()`. However, the modules (apart from `auto-track-event.js` for which this is needed) use `init()` so they have to be instantiated separately.
- Doesn't include Selection buttons which are [deprecated](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/docs/javascript.md#selection-buttons).
- Doesn't include any of the tracking scripts from Frontend Toolkit as it seems unlikely these will conflict with GOV.UK Frontend which doesn't include any tracking scripts.
- Includes `auto-track-event.js` as this seems "special", being incldued in the `/modules` folder instead of `/analytics`. 

I've tested the modules to work on the pages but haven't included the markup in this PR. We could consider adding them to some pages if we thought this was useful.

This can also now be used to reproduce https://github.com/alphagov/govuk-frontend/issues/1270 by going to `/full-page-examples/announcements` and clicking on mobile Menu button.

Includes latest package-lock (see https://github.com/alphagov/govuk-frontend/pull/1377)

Fixes https://github.com/alphagov/govuk-frontend/issues/1360

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -62,8 +62,10 @@ module.exports = (options) => {
   // serve html5-shiv from node modules
   app.use('/vendor/html5-shiv/', express.static('node_modules/html5shiv/dist/'))
 
-  // serve legacy code from node node modules
+  // serve legacy code from node modules
   app.use('/vendor/govuk_template/', express.static('node_modules/govuk_template_jinja/assets/'))
+  app.use('/vendor/govuk_frontend_toolkit/', express.static('node_modules/govuk_frontend_toolkit/javascripts/govuk/'))
+  app.use('/vendor/jquery/', express.static('node_modules/jquery/dist'))
 
   app.use('/assets', express.static(path.join(configPaths.src, 'assets')))
 
--- a/app/views/layouts/_generic.njk
+++ b/app/views/layouts/_generic.njk
@@ -43,6 +43,39 @@
 {% endblock %}
 
 {% block bodyEnd %}
+
+  {% if legacy %}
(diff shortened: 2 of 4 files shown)


app/app.js
```diff
@@ -62,8 +62,10 @@ module.exports = (options) => {
   // serve html5-shiv from node modules
   app.use('/vendor/html5-shiv/', express.static('node_modules/html5shiv/dist/'))
 
-  // serve legacy code from node node modules
+  // serve legacy code from node modules
   app.use('/vendor/govuk_template/', express.static('node_modules/govuk_template_jinja/assets/'))
+  app.use('/vendor/govuk_frontend_toolkit/', express.static('node_modules/govuk_frontend_toolkit/javascripts/govuk/'))
+  app.use('/vendor/jquery/', express.static('node_modules/jquery/dist'))
```
aliuk2012: # 😱   jQuery !!!

hannalaakso: Yeah... it's a Frontend Toolkit dependency needed for testing. Rest assured, it will only be included in the review app legacy mode and won't be 🚢

36degrees:
Merged in Hanna's absence.

PR #1387, opened 2019-05-28
removing hard coded text and replacing with title param

htmlandbacon:
I noticed when I was doing some Welsh language conversion that this is currently hard coded.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,9 @@ See the [versioning documentation for how to update this changelog](./docs/contr
 
 🆕 New features:
 
-- Meta title
+- Support custom text on meta title 
 
-  Hard coded 'Support Links' is replaced by `params.meta.title`
+  Hard coded 'Support Links' is replaced by `params.meta.visuallyHiddenTitle` this has a default off Support links.
 
   ([PR #1387](https://github.com/alphagov/govuk-frontend/pull/1387))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,19 @@
 
 See the [versioning documentation for how to update this changelog](./docs/contributing/versioning.md#updating-changelog).
 
+🆕 New features:
+
+- Meta title
+
+  Hard coded 'Support Links' is replaced by `params.meta.title`
+
+  ([PR #1387](https://github.com/alphagov/govuk-frontend/pull/1387))
+
 🔧 Fixes:
 
 - Fixes the color of disabled secondary and warning button variants
+
+  ([PR #1392](https://github.com/alphagov/govuk-frontend/pull/1392))
 
 ## 2.12.0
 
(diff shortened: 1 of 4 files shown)


src/components/footer/template.njk
```diff
@@ -32,8 +32,8 @@
     <div class="govuk-footer__meta">
       <div class="govuk-footer__meta-item govuk-footer__meta-item--grow">
         {% if params.meta %}
+          <h2 class="govuk-visually-hidden">{{ params.meta.title }}</h2>
```
nickcolley: At the moment, if we did this it'd be a breaking change since there is no default.

You could consider adding a default using the Nunjucks filter.

`{{ params.meta.title | default("Support links") }}`

You could then keep the existing test the same, and add a new test that sets this new feature.

src/components/footer/footer.yaml
```diff
@@ -4,6 +4,9 @@ params:
   required: false
   description: Object containing options for the meta navigation.
   params:
+  - name: title
+    type: string
+    description: Title for a meta item section
```
nickcolley: Should we add something in the description about this title being hidden?

If we choose to keep a default around, we'd also want to say that this overrides the default title for the meta item section...

CHANGELOG.md
```diff
@@ -4,6 +4,14 @@
 
 See the [versioning documentation for how to update this changelog](./docs/contributing/versioning.md#updating-changelog).
 
+🆕 New features:
+
+- Meta title
```
nickcolley: The title you have in your pull request might be more descriptive for the changelog.

src/components/footer/footer.yaml
36degrees: `params.meta.visuallyHiddenText` or `params.meta.visuallyHiddenTitle` would be consistent with the approach we use elsewhere.

src/components/footer/template.njk
htmlandbacon: I did debate having fallback text, largely I didn't because if you are switching lang, it might be there by default and not noticed.

Happy to make this change if required though.

nickcolley: Yeah I think we want to keep this default since most people will not know they need to set this as it's invisible.

src/components/footer/footer.yaml
htmlandbacon: I went with visuallyHiddenTitle, I've not updated the description, as I think the parameter name does that, but have mentioned the default.

htmlandbacon:
Cheers for the prompt review @nickcolley @36degrees 

 I've reflected the changes 🎉

aliuk2012:
Hi Colin,

Looks great, I've just rebase your work with the latest master and resolved the minor conflict in the changelog. Once the tests have passed, we can merge this pr.

Thanks again for this contribution 😄

PR #1393, opened 2019-05-29
Spike nested folder namespace

nickcolley:
See a spiked installation of this branch in the design system repository: https://github.com/alphagov/govuk-design-system/pull/889


nickcolley:
As the proposal has been accepted I will close this but leave the branch around for future reference.

PR #1421, opened 2019-06-03
Fix footer links hover focus state

nickcolley:
I think I unintentionally introduced this when changing the link underline behaviour to only be present on certain footer links...

When a link is focused and a pointer is hovering of it, it should not show an underline.

### DIFF
--- a/src/components/footer/_footer.scss
+++ b/src/components/footer/_footer.scss
@@ -77,13 +77,16 @@
     }
   }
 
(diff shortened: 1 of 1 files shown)


nickcolley:
I'm not adding another changelog entry since we already have one for adding the new focus styles: https://github.com/alphagov/govuk-frontend/pull/1321

Issue #929, opened 2018-07-26
Make it easy to generate the error summary with correct href values


adamsilver:
When you submit a form containing groups of fields such as radio buttons, populating the error summary so that focus is moved when the error link is clicked is problematic.

Say you have this:

	<a href="#name-of-field">Enter blah</a>

And you have this:

	<input type="radio" name="name-of-field" id="name-of-field-1" value="1">
	<input type="radio" name="name-of-field" id="name-of-field-2" value="2">

When you click the link focus won't move to the first `<input>` in the group. For that to work, we'd have to suffix the href with `-1` to match the input's `id` attribute value.

This means the server has to be told that a request that contains `name-of-field` is in fact a group of radios and therefore when the error summary is rendered for this particular error message, be sure to suffix it with `-1`.

This is a pain, and an unnecessary burden on the developer. What I've always done, is make sure the first input's `id` in the group matches the name of the group. Like this:

	<input type="radio" name="name-of-field" id="name-of-field" value="1">
	<input type="radio" name="name-of-field" id="name-of-field-2" value="2">

This works nicely on the server side and on the client side (should a service implement client-side form validation).

36degrees:
Thanks for raising Adam!

We should go through and check that all of the components work well with the error summary. Am I right in thinking this applies to at least the radios, checkboxes and date input component?

I think there might be some things to explore here, especially whether we should link from the error summary to the first field or to the parent fieldset? Have you done any testing with assistive technologies that shows that linking to the first field is the best thing to do?

adamsilver:
No worries.

I think it should do this for the radios and the checkboxes. The date input might be a little different.

You see, with the date input, a problem could be that the user left just one of the three input's empty. So the error might read “Enter the day of your date of birth” (or similar). And so the error should link to  the day field (not the fieldset/wrapper). But what if the user leaves two of the three input's blank? Is that two errors (or one)? Interesting stuff.

I took this approach (with the radios/checkboxes) back in ~2008 after the RNIB audited a site I was working on back then. In terms of functionality, it works and I think this is reasonable behaviour for say a screen reader. It's the group that's erroneous and by focusing the first input, the legend would be read out along with the first radio (in much the same way as they would generally traverse and enter the field).

With that said, I would love to learn more from deeper testing with people who use AT.

igloosi:
technically the change is simple
```
{% set id = item.id if item.id else idPrefix + ("-" + loop.index if not loop.first)  %}
```
[...]

PR #1426, opened 2019-06-04
Make radios and checkboxes components easier to link to from error summary

nickcolley:
Updates the radios, checkboxes and date input components so it's easier to link to their first items from the error summary component.

I have tried to simplify the date input component in a similar way but the complexity was the same after I changed it, if not worse. So if we want to make the date input component easier to use in the future we'll need to think about something else.

Closes https://github.com/alphagov/govuk-frontend/issues/929

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,38 @@
 
 💥 Breaking changes:
 
+- Make radios and checkboxes components easier to link to from the error summary component
+
+  You should check that clicking a link from the error summary to radios or checkboxes results in the first input being focused.
+
+  The `id` of the first input in a set of radios or checkboxes will no longer be suffixed with `-1`. This means that it will be the same as your `idPrefix`, or if `idPrefix` is not set it'll be the same as the `name`.
+
+  We've made this change to make it easier to link from the error summary to radios or checkboxes, if you we're working around this before by setting the first input's `item.id` to the same as the `name` you can now remove this.
+
+  To migrate you can update your error summary so it links to the first element, for example if you are using an `idPrefix` option of `question`, your error summary can now link to `#question` instead of `#question-1`.
+
+  If you cannot change your error summary, you could change your component to set the `item.id` option for the first input, for example:
+
+  ```javascript
+  {{ govukRadios({
+    idPrefix: "question",
+    name: "question",
+    items: [
+      {
+        id: "question-1",
+        value: "yes",
+        text: "Yes"
+      },
+      {
+        value: "no",
+        text: "No"
+      }
+    ]
+  }) }}
+  ```
+
+  ([PR #1426](https://github.com/alphagov/govuk-frontend/pull/1426))
+
 - Remove `govuk-focusable`, `govuk-focusable-fill` mixins, introduce `govuk-focus-text` mixin.
 
   To migrate:
--- a/src/components/checkboxes/template.njk
+++ b/src/components/checkboxes/template.njk
@@ -50,7 +50,17 @@
     {%- for attribute, value in params.attributes %} {{ attribute }}="{{ value }}"{% endfor %}
     {%- if isConditional %} data-module="checkboxes"{% endif -%}>
     {% for item in params.items %}
-    {% set id = item.id if item.id else idPrefix + "-" + loop.index %}
+    {#- If the user explicitly sets an id, use this instead of the regular idPrefix -#}
+    {%- if item.id -%}
+      {%- set id = item.id -%}
+    {%- else -%}
+      {#- The first id should not have a number suffix so it's easy to link to from the error summary component -#}
+      {%- if loop.first -%}
+        {%- set id = idPrefix %}
+      {% else %}
+        {%- set id = idPrefix + "-" + loop.index -%}
+      {%- endif -%}
+    {%- endif -%}
     {% set name = item.name if item.name else params.name %}
     {% set conditionalId = "conditional-" + id %}
     {% set hasHint = true if item.hint.text or item.hint.html %}
--- a/app/views/full-page-examples/have-you-changed-your-name/index.njk
+++ b/app/views/full-page-examples/have-you-changed-your-name/index.njk
@@ -55,7 +55,6 @@ scenario: >-
       },
       items: [
         {
-          id: "changed-name",
           value: "yes",
           text: "Yes"
         },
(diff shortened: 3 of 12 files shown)


src/components/date-input/template.njk
```diff
@@ -52,16 +52,25 @@
   }) | indent(2) | trim }}
 {% endif %}
   <div class="govuk-date-input {%- if params.classes %} {{ params.classes }}{% endif %}"
-    {%- for attribute, value in params.attributes %} {{attribute}}="{{value}}"{% endfor %}
-    {%- if params.id %} id="{{ params.id }}"{% endif %}>
```
nickcolley: I'm not sure this ID was doing anything, since it's not focusable? So potentially we should remove this regardless?

app/views/full-page-examples/passport-details/index.js
```diff
@@ -8,7 +8,7 @@ module.exports = (app) => {
       body('passport-number')
         .exists()
         .not().isEmpty().withMessage('Enter your passport number'),
-      body('expiry-day')
+      body('expiry')
```
nickcolley: I'm not convinced making these changes on the date input component actually makes it any easier.

You can see that I was not able to simplify the logic in the validation section: https://github.com/alphagov/govuk-frontend/blob/d41467da8dafd2b259cd06daa66bf4a7266a9d5d/app/views/full-page-examples/passport-details/index.js#L28-L52

Is it worth changing the date input considering this?

CHANGELOG.md
```diff
@@ -4,6 +4,17 @@
 
 💥 Breaking changes:
 
+- Make components easier to link to from error summary
+
+  Updates the radios, checkboxes and date input components so it's easier to link to their first items from the error summary component.
+
+  To migrate:
+
+  You will need to ensure that your error summary still links correctly to your first input by removing any logic that adds a suffix to your id.
```
nickcolley: I don't think this is very good, would welcome some ideas to make this better. Potentially some before and after examples?

app/views/full-page-examples/passport-details/index.js
nickcolley: Noticed that the full page example that features the date input doesnt work anymore, so maybe I will have to change this logic, will see if it makes it simpler or not, I suspect it'll be more complex than it already is...

CHANGELOG.md
```diff
@@ -4,6 +4,35 @@
 
 💥 Breaking changes:
 
+- Make radios and checkboxes components easier to link to from the error summary component
+
+  The first input in the radios and checkboxes will have an `id` and `name` attribute that match, allowing for easier generation of error summary links.
```
36degrees: This is only true if you haven't specified your own `idPrefix`, I think?

CHANGELOG.md
```diff
@@ -4,6 +4,35 @@
...
+
+  You should check that clicking an error link for a radios or checkboxes results in the first input being focused.
```
36degrees: Is it clear enough what we mean by 'error link'? Would it be better to use something like:

```suggestion
  You should check that clicking a link from the error summary to radios or checkboxes results in the first input being focused.
```

CHANGELOG.md
36degrees: I wonder if we should be more explicit about what's actually changed here?

Perhaps something like…

> The ID of the first input in a set of radios or checkboxes will no longer be suffixed with `-1`. This means that it will be the same as your `idPrefix`, or if `idPrefix` is not set it'll be the same as the `name`
>
> We've made this change to make it easier to link from the error summary to radios or checkboxes.

CHANGELOG.md
```diff
@@ -4,6 +4,35 @@
...
+
+  To migrate you can update your error summary so it links to the first element, for example if you are using an `idPrefix` option of `question`, your error summary can now link to `#question` instead of `#question-1`.
```
36degrees: Users might also be overriding the `id` of the first radio/checkbox, which I think is what we suggested in #929. In that case, they could remove the `id` as it won't be doing anything.

CHANGELOG.md
```diff
@@ -4,6 +4,35 @@
...
+
+  If you cannot change your error summary, you could change your component to set the `item.id` option for the first input, for example:
+
+  ```javascript
+  {{ govukRadios({
```
36degrees: This example is missing a `name` attribute, which is required.

src/components/checkboxes/template.njk
```diff
@@ -50,7 +50,16 @@
     {%- for attribute, value in params.attributes %} {{ attribute }}="{{ value }}"{% endfor %}
     {%- if isConditional %} data-module="checkboxes"{% endif -%}>
     {% for item in params.items %}
-    {% set id = item.id if item.id else idPrefix + "-" + loop.index %}
+    {#- If the user explicitly sets an id, use this instead of the regular idPrefix -#}
+    {%- if item.id -%}
+      {%- set id = item.id -%}
+    {%- else -%}
+      {%- set id = idPrefix -%}
+      {#- The first id should not have a prefix so it's easy to link to from the error summary component -#}
```
36degrees: The number is a suffix, not a prefix 🙂 

```suggestion
      {#- The first id should not have a number suffix so it's easy to link to from the error summary component -#}
```

src/components/checkboxes/template.njk
```diff
@@ -50,7 +50,16 @@
...
+      {%- if not loop.first -%}
+        {%- set id = id + "-" + loop.index -%}
```
36degrees: My brain tripped over this a little bit because I think it would end up using `item.id` as the prefix, rather than `idPrefix`. Now that I've read it again, I get what's going on, but I think the multiple `{% set %}` calls combined with the very similarly named variables could be confusing.

Very much non-blocking, but it _might_ be better to do e.g.

```
{%- if loop.first -%}
  {%- set id = idPrefix %}
{% else %}
  {%- set id = id + "-" + loop.index -%}
{%- endif -%}
```

CHANGELOG.md
nickcolley: I figured since this wasnt required that I'd leave the up to people themselves rather than potentially confusing things... maybe should just have this in

src/components/checkboxes/template.njk
nickcolley: I'll give this a go thanks for the comment

src/components/checkboxes/template.njk
nickcolley: i always get these mixed up lol :)

CHANGELOG.md
nickcolley: I tried this and I'm worried that there's just a lot of steps being described that could confuse people so I wanted to focus on the things they definitely need to do...

nickcolley: nevermind I've included it after all..

Issue #1012, opened 2018-09-26
Add version 2 of the font to frontend


Nooshu:
*Note*: Moving these details over from [PR 357](https://github.com/alphagov/govuk_template/pull/357) on govuk_template.

Now that the [new font loading strategy](https://github.com/alphagov/govuk_template/pull/341) is live in govuk_template and GOV.UK we can now move onto the next iteration of improving frontend performance.

# Issue
The first version of the custom font wasn't 100% optimised for the web. This has created four main issues:

1. the filesize for all versions of the font is larger than it needs to be
2. we require a separate font for the tabular numbers used in data tables
3. the glyphs sit too high in the fonts bounding box which leads to the need for correction in the CSS
4. hinting requirement for oldIE (EOT) increases the filesize even further

Version 2 of the font has been created to solve three of these issues. See below for information around each of the issues listed. I've also proposed a solution to issue 4.

## Filesize
The filesize of the current font is bigger than it needs to be. The new font has been optimised to fix this issue. See the table below comparing the unhinted version of the font.

Font | v1 | v2 | Difference
------------ | ------------- | ------------- | -------------
Bold WOFF2 | 55KB | 32KB | -42%
Bold WOFF | 73KB | 41KB | -44%
Bold EOT | 72KB | 57KB | -21%
Light WOFF2 | 68KB | 33KB | -51%
Light WOFF | 96KB | 43KB | -55%
Light EOT | 92KB | 57KB | -38%

The new fonts will reduce the download requirement by 58KB for browsers that support WOFF2 and by 85KB for WOFF browsers. For EOT usage see below.

## Tabular numbers
Version 1 of the font was missing the details required to enable embedded tabular numbers. Version 2 now allows us to control this tablular number setting using CSS, thus removing the need for two other fonts (tabular-light, tabular-bold). This tabular setting can be enabled using the following example CSS:

````CSS
.tabular {
    font-feature-settings: "tnum" 1;
}

@supports(font-variant-numeric: tabular-nums) {
    .tabular {
        font-feature-settings: normal;
        font-variant-numeric: tabular-nums;
    }
}
````
The difference this setting makes can be seen in the screenshot below. Older browser will fallback to `font-feature-settings`, where as newer browser will see the `@supports` rule and use `font-variant-numeric`.

![screen shot 2018-04-05 at 13 07 54](https://user-images.githubusercontent.com/1223960/38364941-7474059a-38d2-11e8-971d-eeb61de31676.png)

[...]

PR #1434, opened 2019-06-06
Update font to use v2 of GOV.UK Transport font.

aliuk2012:
Based off @dashouse's PR https://github.com/alphagov/govuk-frontend/pull/1356

This PR uses v2 of the font.

- Adds new font files, reducing the amount of files from 12 to 4
- Significantly better alignment with the baseline of Helvetica / Arial*
- Tabular numbers are now rolled up in to the Light or Bold weights and set with font-feature-settings (CSS applied to existing mixin so $tabular: true still works)

- All browsers apart from IE8 and older support WOFF or WOFF2. IE8 and older will not download any font but will fallback to `Arial`

This PR does not remove any of the adjustments that were needed for v1. 

## Old font: 
👉 Home page: https://govuk-frontend-review-pr-1434.herokuapp.com/?legacy=1
👉 Table component: https://govuk-frontend-review-pr-1434.herokuapp.com/components/table?legacy=1
👉 Tabular number example: https://govuk-frontend-review-pr-1434.herokuapp.com/components/table/table-with-head-and-caption/preview?legacy=1

## New Font: 
👉  Home page: https://govuk-frontend-review-pr-1434.herokuapp.com/
👉  Table Component: https://govuk-frontend-review-pr-1434.herokuapp.com/components/table
👉  Tabular number example: https://govuk-frontend-review-pr-1434.herokuapp.com/components/table/table-with-head-and-caption/preview

Thanks to @Nooshu for creating the original issue (https://github.com/alphagov/govuk_template/pull/357 & #1012) and the recommendation to drop EOT files for older IE https://github.com/alphagov/govuk-frontend/pull/1356#issuecomment-494727525.


src/helpers/_font-faces.scss
```diff
@@ -7,61 +7,35 @@
...
+    }
 
-      @font-face {
-        font-family: "ntatabularnumbers";
-        src: govuk-font-url("bold-tabular-357fdfbcc3-v1.eot");
-        src: govuk-font-url("bold-tabular-357fdfbcc3-v1.eot?#iefix") format("embedded-opentype"),
-             govuk-font-url("bold-tabular-b89238d840-v1.woff2") format("woff2"),
-             govuk-font-url("bold-tabular-784c21afb8-v1.woff") format("woff");
+    @font-face {
+        font-family: "GOVUKTransport";
```
nickcolley: `font-family: "GOV.UK Transport"`? 👼

src/settings/_typography-font.scss
```diff
@@ -16,14 +16,20 @@
 /// @type List
 /// @access public
 
-$govuk-font-family: $govuk-font-family-nta !default;
+$govuk-font-family: if($govuk-use-legacy-font,
```
nickcolley: Really like how this refactor has panned out nice work.

src/settings/_typography-font.scss
```diff
@@ -34,16 +40,15 @@ $govuk-font-family-print: sans-serif !default;
 
 /// Include the default @font-face declarations
 ///
-/// If you have set $govuk-font-family to something other than $govuk-font-family-nta this option is disabled by default.
+/// If you have set $govuk-font-family to something other than $govuk-font-family-govuk-transport this option is disabled by default.
 ///
 /// If you are using $govuk-compatibility-govuktemplate this option is disabled by default.
```
36degrees: I think this line can probably go?

src/settings/_typography-font-families.scss
```diff
@@ -1,12 +1,38 @@
...
+/// Use 'legacy' fonts
+///
+/// Whether or not to use v1 nta font from GOV.UK Elements / Frontend
+/// Toolkit, for teams that are migrating to GOV.UK Frontend and may be using
+/// components from both places in a single application.
+///
+/// @type Boolean
+/// @access public
+
+$govuk-use-legacy-font: if((
```
36degrees: This doesn't get used in this file – would it make more sense in `typography-font`?

src/helpers/_font-faces.scss
```diff
@@ -7,61 +7,35 @@
...
+        src: govuk-font-url("bold-9561e2d007-v2.woff2") format("woff2"),
+             govuk-font-url("bold-5d3836b441-v2.woff") format("woff");
         font-weight: bold;
         font-style: normal;
-        font-display: fallback;
-      }
+        font-display: swap;
+    }
```
36degrees: Indentation issue?

src/helpers/_font-faces.scss
```diff
@@ -7,61 +7,35 @@
...
-        src: govuk-font-url("light-tabular-498ea8ffe2-v1.eot");
-        src: govuk-font-url("light-tabular-498ea8ffe2-v1.eot?#iefix") format("embedded-opentype"),
-             govuk-font-url("light-tabular-851b10ccdd-v1.woff2") format("woff2"),
-             govuk-font-url("light-tabular-62cc6f0a28-v1.woff") format("woff");
-        font-weight: normal;
-        font-style: normal;
-        font-display: fallback;
-      }
+        font-display: swap;
+    }
```
36degrees: Don't think this in indented correctly.

src/helpers/_typography.scss
```diff
@@ -171,7 +172,18 @@
 
 @mixin govuk-font($size, $weight: regular, $tabular: false, $line-height: false) {
   @if $tabular {
-    @include govuk-typography-common($font-family: $govuk-font-family-tabular);
+    // if govuk-font-family-tabular is set use $govuk-font-family-tabular
+    @if $govuk-font-family-tabular != false {
```
36degrees: Non-blocking, but I think this can just be

```suggestion
    @if $govuk-font-family-tabular {
```

CHANGELOG.md
```diff
@@ -4,6 +4,27 @@
 
 💥 Breaking changes:
 
+- Update the font to v2
```
36degrees: What do we think about splitting this out into multiple bullet points? (e.g. the IE8 change could be it's own thing) and possibly trying to give a TL;DR of how v2 is different and/or why we're doing this?

36degrees:
P.S. super commit structure and messages 👍

aliuk2012:
@36degrees makes it super easy to merge and rebase https://thoughtbot.com/blog/autosquashing-git-commits

src/helpers/_font-faces.scss
hannalaakso: Not sure it's safe to have you a full stop in `font-family` 🤔 This seems to caution against things other than letters/numbers/spaces: https://developer.mozilla.org/en-US/docs/Web/CSS/font-family#Valid_family_names

36degrees:
Oh, I meant the earlier commits with the detailed commit messages :+1:

src/helpers/_font-faces.scss
nickcolley: > This means that punctuation characters and digits at the start of each token must be escaped in unquoted font family names.

I think this means that punctuation is fine as long as it's with quote marks?

hannalaakso: Yes good spot 👍

nickcolley: @aliuk2012 what do you think? We could change it based on talking to Dave later?

aliuk2012: @nickcolley I'd be up for changing it to 'GOVUK Transport` but lets get it through now and I'll catch up with @dashouse on Monday and see what he thinks

src/settings/_typography-font-families.scss
```diff
@@ -1,12 +1,23 @@
+@import "compatibility";
```
36degrees: I don't believe this import is required here any more.

src/helpers/_font-faces.scss
```diff
@@ -7,60 +7,34 @@
...
-        src: govuk-font-url("light-2c037cf7e1-v1.eot?#iefix") format("embedded-opentype"),
-             govuk-font-url("light-f38ad40456-v1.woff2") format("woff2"),
-             govuk-font-url("light-458f8ea81c-v1.woff") format("woff");
+        font-family: "GOVUKTransport";
+        src: govuk-font-url("light-68f0c84f0e-v2.woff2") format("woff2"),
+             govuk-font-url("light-222368e53d-v2.woff") format("woff");
         font-weight: normal;
         font-style: normal;
-        font-display: fallback;
+        font-display: swap;
```
nickcolley: @hannalaakso has noticed the behaviour might not be better than fallback.

We should leave this as fallback, and do a proper test that includes slower network conditions and an optimised server before changing this.

src/helpers/_font-faces.scss
```diff
@@ -7,60 +7,34 @@
...
-      @font-face {
-        font-family: "ntatabularnumbers";
-        src: govuk-font-url("bold-tabular-357fdfbcc3-v1.eot");
-        src: govuk-font-url("bold-tabular-357fdfbcc3-v1.eot?#iefix") format("embedded-opentype"),
-             govuk-font-url("bold-tabular-b89238d840-v1.woff2") format("woff2"),
-             govuk-font-url("bold-tabular-784c21afb8-v1.woff") format("woff");
-        font-weight: bold;
-        font-style: normal;
-        font-display: fallback;
+        font-display: swap;
```
nickcolley: Same as above

src/helpers/_font-faces.scss
nickcolley: https://github.com/alphagov/govuk-frontend/issues/1437

Nooshu:
Excellent work @aliuk2012! Very excited to see this in Frontend. I'd say leave `font-display` with `fallback` for the moment. There's no reason for it to become blocker. We can create a few test pages and see which setting works best. Should be an easy change for a point release post v3 release.

PR #1436, opened 2019-06-06
Spike an all.njk file

nickcolley:
If we were interested in this, I would want to check it wouldnt block an eventual 'exploded' components approach down the road for example.

radios.fieldset
radios.item
radios.item
radios.item

etc


nickcolley:
Closing this for now

PR #1439, opened 2019-06-07
Document how to use the legacy palette

### LINKED ISSUE #1432: Document colour palette sass variable settings
In #1288 we introduced a new setting `$govuk-use-legacy-palette`, but we haven't documented this setting anywhere.

## Done when

- [ ] we've added documentation that explains what the `$govuk-use-legacy-palette` setting is for and how to use it

36degrees:
It’s hard to talk about this option without also talking about compatibility mode, so this also splits the documentation on compatibility mode out from the installation instructions and adds more detail about what it actually does.

There is some duplication between this and the compatibility mode Sass documentation updated in #1438 – it feels like it does need to be explained in both places. Open to ideas on how to reduce that duplication.

Closes #1432

### DIFF
--- a/docs/installation/compatibility.md
+++ b/docs/installation/compatibility.md
@@ -1,0 +1,53 @@
+# Use GOV.UK Frontend with old frameworks or colours
+
+You can configure GOV.UK Frontend to work with the following 'legacy'
+frameworks:
+
+- [GOV.UK Frontend toolkit]
+- [GOV.UK Template]
+- [GOV.UK Elements]
+
+You cannot use the following features if you installed GOV.UK Frontend from
+`dist`. [Install with node package manager](installing-from-npm.md) instead.
+
+If you use these features, your service will not have GOV.UK Frontend's many
+accessibility and usability improvements.
+
+## Turn on 'compatibility mode'
+
+You can turn on 'compatibility mode' if you want to use both GOV.UK Frontend
+components and a legacy framework in a service.
+
+GOV.UK Frontend will:
+
+- use the old colour palette
+- use the old GOV.UK font from GOV.UK Template
+- override some of the CSS in the legacy frameworks
+- no longer use `rem` for font sizes
+
+To turn on compatibility mode for all 3 legacy frameworks, add the following
+variables to your project's Sass file, above `@import "govuk-frontend/all"`:
+
+```SCSS
+$govuk-compatibility-govukfrontendtoolkit: true;
+$govuk-compatibility-govuktemplate: true;
+$govuk-compatibility-govukelements: true;
+```
+
+You can leave out a variable to turn off compatibility mode for that framework.
+
+## Use the old colour palette
+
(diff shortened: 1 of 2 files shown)


36degrees:
@m-green would you have any time to review this? Happy to talk it through in person.

docs/installation/compatibility.md
```diff
@@ -0,0 +1,71 @@
...
+$govuk-compatibility-govukelements: true;
+
+@import "govuk-frontend/all";
+```
+
+If compatibility mode is enabled for any framework, GOV.UK Frontend will:
+
+- use the legacy colour palette
+- no longer use rem for font sizes
+- include adjustments to the CSS to 'counter' specific CSS rules in the legacy
```
aliuk2012: I think it might be worth mentioning how these variables affect the font that is loaded.

m-green:
@36degrees I've rejigged for style and conciseness - let me know what you think. Happy to chat through!

docs/installation/compatibility.md
```diff
@@ -1,64 +1,46 @@
...
 
-In order to enable either of these options you'll need to have [installed GOV.UK
-Frontend using npm](installing-with-npm.md). These features are not available if
-you are using the 'dist' version.
+## Turn on 'compatibility mode'
 
+You can turn on 'compatibility mode' if you want to use both GOV.UK Frontend components and a legacy framework in a service.
 
-## Compatibility mode
+GOV.UK Frontend components in your pages will:
```
36degrees: It's not just the components – this would also apply to anything styled using GOV.UK Frontend, for example headings, paragraphs and links.

docs/installation/compatibility.md
```diff
@@ -1,64 +1,46 @@
...
 ```
 
-If compatibility mode is enabled for any framework, GOV.UK Frontend will:
-
-- use the legacy colour palette
-- no longer use rem for font sizes
-- include adjustments to the CSS to 'counter' specific CSS rules in the legacy
-  frameworks
-- use the previous version of the font included with GOV.UK Template
+Set a variable to `false` to turn off compatibility mode for that framework.
```
36degrees: I _think_ we'd generally encourage people to omit the variable entirely if they're not changing it from the default.

docs/installation/compatibility.md
m-green: Ah good to know, thanks - tweaked in https://github.com/alphagov/govuk-frontend/pull/1439/commits/f6214966822764e76468e3e5d4b672f9c4796c7f.

docs/installation/compatibility.md
m-green: Thank you - fixed in https://github.com/alphagov/govuk-frontend/pull/1439/commits/f6214966822764e76468e3e5d4b672f9c4796c7f.

36degrees:
Squashed into one commit and wrapped at 80 characters.

PR #1440, opened 2019-06-07
Move aria-label to nav element in header

### LINKED ISSUE #1340: Review if `aria-label` in header navigation is necessary
The [header with navigation](https://design-system.service.gov.uk/components/header/with-service-name-and-navigation/index.html) throws the following warning in https://validator.w3.org/:
> Possible misuse of aria-label.

This relates to https://github.com/alphagov/govuk-frontend/blob/fea29aaa833d5f8233c71e0d076923a697e752cd/src/components/header/template.njk#L68 which was inherited from [govuk_template](https://alphagov.github.io/govuk_template/example-proposition-menu.html).

The [WCAG specs ](https://www.w3.org/TR/WCAG20-TECHS/ARIA6.html#ARIA6-examples) state that `aria-label` can be used for:
- ` Distinguishing navigation landmarks`
- `Identifying region landmarks`

We should review whether having `aria-label` in this context provides useful information for assistive technologies. 

There is also a related issue where we hide the list inside the navigation element when showing/hiding the navigation. This means that a user can end up with an empty `<nav></nav>` element, which may cause issues/confusion for screenreader users. See [comment](https://github.com/alphagov/govuk-frontend/issues/1340#issuecomment-501318558) for more context.

## Done when

- [ ] TBC
- [ ] "Details of breaking change" section below completed to help with writing release notes

## Details of breaking change

- which users are affected: TBC
- the change that’s been made: TBC
- changes users will have to make: TBC
- impact of users not making those changes: TBC

36degrees:
HTML validator flagged this as a ‘possible misuse of aria-label.’

This pattern of applying aria-label to the `<ul>` was [copied from GOV.UK Template][1].

Best practise is to [add the aria-label to the element with the navigation role][2], which in this case would be the `<nav>`.

When a label is used on a navigation element, [the term ‘navigation’ should be omitted][3], as ‘the screen reader will read both the role and the contents of the label’, so I’ve also removed the word ’navigation’ from the label.

Closes #1340.

[1]: https://github.com/alphagov/govuk_template/blob/f164b2825b1479d7c7d4bdbec4ba9c8e171f504c/docs/usage.md#propositional-title-and-navigation
[2]: https://www.w3.org/TR/wai-aria-practices/examples/landmarks/navigation.html
[3]: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Navigation_Role#Associated_WAI-ARIA_Roles_States_and_Properties

### DIFF
--- a/src/components/header/template.njk
+++ b/src/components/header/template.njk
@@ -61,8 +61,8 @@
     {% endif %}
     {% if params.navigation %}
     <button type="button" role="button" class="govuk-header__menu-button js-header-toggle" aria-controls="navigation" aria-label="Show or hide Top Level Navigation">Menu</button>
-    <nav>
-      <ul id="navigation" class="govuk-header__navigation {{ params.navigationClasses if params.navigationClasses }}" aria-label="Top Level Navigation">
+    <nav aria-label="top level">
+      <ul id="navigation" class="govuk-header__navigation {{ params.navigationClasses if params.navigationClasses }}">
         {% for item in params.navigation %}
           {% if item.href and item.text %}
             <li class="govuk-header__navigation-item{{ ' govuk-header__navigation-item--active' if item.active }}">
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -268,6 +268,27 @@
 
   ([PR #1376](https://github.com/alphagov/govuk-frontend/pull/1376))
 
+- The 'aria-label' in the header is now correctly applied to the `<nav>` element
+  rather than the `<ul>` within it, and has been changed from 'Top Level
+  Navigation' to 'top level' as screen readers will already announce the role.
+
+  If you are using navigation within your header, and you are not using the
+  nunjucks templates, you should make this change in your own code, replacing
+
+  ```html
+  <nav>
+    <ul id="navigation" class="govuk-header__navigation" aria-label="Top Level Navigation">
+  ```
+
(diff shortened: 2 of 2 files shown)


src/components/header/template.njk
```diff
@@ -61,8 +61,8 @@
     {% endif %}
     {% if params.navigation %}
     <button type="button" role="button" class="govuk-header__menu-button js-header-toggle" aria-controls="navigation" aria-label="Show or hide Top Level Navigation">Menu</button>
```
nickcolley: Are we sure about this change? To me it seems like the label exists to compliment the `aria-controls` for the button on line 63...

36degrees: Which bit of the change are you not sure about?

nickcolley: It seems from reading the source all together like the aria-label on the `ul` element exists to give it an accessible name when expanded by the button.

Since the `button` and `ul` have very similar accessible names.

36degrees: I do think the aria-label should be on the nav, rather than the ul. Perhaps though the aria-controls should point to the nav rather than the ul too?

nickcolley: I think if we test this with assistive technologies it might give us an idea of the interaction.

36degrees:
Needs more investigation.

Issue #1433, opened 2019-06-05
Prevent double clicks on individual buttons rather than globally


36degrees:
This was started by @alex-ju in #1370 but we blocked in whilst we worked on namespacing.

Once we've understood how we intend to namespace JS components going forward, we should unblock and update that PR.

PR #1443, opened 2019-06-11
Ensure GOV.UK Frontend component selectors cannot conflict when initialised

nickcolley:
This change introduces to main changes:

1. `govuk-` prefix to all `[data-module]` selectors to avoid conflicts.
We have found users using the same name without a prefix which results in a conflicts that break components: https://github.com/alphagov/govuk-frontend/issues/1218

2. Initialising the button and details component with a `[data-module]` selector so all components are initialised consistently.

Doing this for the button is important as it unblocks a bugfix for the 'prevent double click' feature: https://github.com/alphagov/govuk-frontend/issues/1433

It will allow us to fix this issue without adding JavaScript behaviour to all buttons on pages, which could result in performance or conflict issues.

Closes https://github.com/alphagov/govuk-frontend/issues/1218

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -268,6 +268,77 @@
 
   ([PR #1376](https://github.com/alphagov/govuk-frontend/pull/1376))
 
+- Ensure GOV.UK Frontend component selectors cannot conflict when initialised
+
+  Components which have JavaScript functionality include a data attribute called `[data-module]`,
+  we have added the `govuk-` prefix as a namespace.
+
+  This ensures the component's name does not conflict with other service level components
+  with the same name, which is consistent with our CSS class name convention.
+
+  If you're using Nunjucks macros, and are [using the `initAll` function](https://github.com/alphagov/govuk-frontend/blob/master/docs/installation/installing-with-npm.md#option-1-include-javascript) you will not be affected.
+
+  If you are using HTML, you will need to add the `govuk-` prefix to any `[data-module]` attributes.
+
+  For example the accordion component selector has changed from `[data-module="accordion"]` to `[data-module="govuk-accordion"]`.
+
+  ```html
+  <div class="govuk-accordion" data-module="govuk-accordion">
+    <!-- ... -->
+  </div>
+  ```
+
+  If you are initialising components manually, you will need to add the `govuk-` prefix to any CSS selectors used to find GOV.UK Frontend components:
+
+  ```javascript
+  var $accordions = document.querySelector('[data-module="govuk-accordion"]')
+  ```
+
+  ([PR #1443](https://github.com/alphagov/govuk-frontend/pull/1443))
+
+
+- Button and details components are now initialised consistently
+
+  The button and details components now use the `[data-module]` selector to initialise them.
+
+  If you're using Nunjucks macros, and are [using the `initAll` function](https://github.com/alphagov/govuk-frontend/blob/master/docs/installation/installing-with-npm.md#option-1-include-javascript) you will not be affected.
+
+  ### Button component
+
+  The button component now is initialised with the `[data-module="govuk-button"]` selector instead of globally based on the document.
+
+  If you are using HTML, you need to add the `[data-module="govuk-button"]` attribute to your buttons:
+
+  ```html
+  <button ... data-module="govuk-button">...</button>
+  ```
+
+  If you are initialising the button component manually, you will need to reference this `[data-module]` attribute:
+
+  ```javascript
+  var $buttons = document.querySelector('[data-module="govuk-button"]')
+  ```
+
+  ### Details component
+
+  The details component now is initialised with the `[data-module="govuk-details"]` selector instead of `details` globally.
+
+  If you are using the Nunjucks macro no changes is needed.
+
+  If you are using HTML, you need to add the `[data-module="govuk-details"]` attribute to your details:
+
+  ```html
+  <details ... data-module="govuk-details">...</details>
+  ```
+
+  If you are initialising the button component manually, you will need to reference this `[data-module]` attribute:
+
+  ```javascript
+  var $details = document.querySelector('[data-module="govuk-details"]')
+  ```
+
+  ([PR #1443](https://github.com/alphagov/govuk-frontend/pull/1443))
+
 - Pull Request Title goes here
 
   Description goes here (optional)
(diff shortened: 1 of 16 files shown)


CHANGELOG.md
```diff
@@ -268,6 +268,77 @@
...
+  <button ... data-module="govuk-button">...</button>
+  ```
+
+  If you are initialising the button component manually, you will need to reference this `[data-module]` attribute:
+
+  ```javascript
+  var $buttons = document.querySelector('[data-module="govuk-buttons"]')
+  ```
+
+  ### Details component
```
nickcolley: Is there a way to avoid the duplication here? I duplicated it to be really clear but it makes it lengthy

CHANGELOG.md
```diff
@@ -268,6 +268,77 @@
 
   ([PR #1376](https://github.com/alphagov/govuk-frontend/pull/1376))
 
+- Ensure GOV.UK Frontend component selectors cannot conflict when initialised
```
nickcolley: @m-green would be good to get your thoughts on this

CHANGELOG.md
```diff
@@ -268,6 +268,77 @@
...
+    <!-- ... -->
+  </div>
+  ```
+
+  This ensures the component's name does not conflict with other service level components
+  with the same name, which is consistent with our CSS class name convention.
+
+  If you're using Nunjucks macros, and are [using the `initAll` function](https://github.com/alphagov/govuk-frontend/blob/master/docs/installation/installing-with-npm.md#option-1-include-javascript) you will not be affected.
+
+  If you are using HTML, you will need to add the `govuk-` prefix to any `[data-module]` attributes referencing GOV.UK Frontend components.
```
hannalaakso: I wonder if 
> referencing GOV.UK Frontend components.

is clear? Could we say 
> any `[data-module]` attributes on GOV.UK Frontend component HTML

We could add
> and any `[data-module]` attributes referencing GOV.UK Frontend scripts in your custom components

although it seems like a fairly unlikely scenario.

hannalaakso: Or alternatively
> any `[data-module]` attributes in HTML which reference GOV.UK Frontend Javascript 
🤷‍♀

CHANGELOG.md
```diff
@@ -268,6 +268,77 @@
...
+
+  If you are initialising components manually, you will need to add the `govuk-` prefix to any selectors referencing GOV.UK Frontend components:
```
hannalaakso: > any **Javascript** selectors

to make it super clear.

CHANGELOG.md
nickcolley: Will try and make this clearer thanks Hanna :)

nickcolley: I've moved the example around and adjusted the wording so it hopefully makes it clearer, lemme know what you think.

m-green:
Hi @nickcolley - I think I've managed to use sections to combine this in a way that avoids duplication (and means that each type of user - HTML/Javascript - gets their guidance in one go). Let me know what you think!
I think some of the terms I've used might not be quite right - let me know if so.

CHANGELOG.md
```diff
@@ -268,72 +268,47 @@
 
   ([PR #1376](https://github.com/alphagov/govuk-frontend/pull/1376))
 
-- Ensure GOV.UK Frontend component selectors cannot conflict when initialised
+- Update and add the data-module attribute
```
nickcolley: I think this title is too vague now

CHANGELOG.md
```diff
@@ -268,72 +268,47 @@
...
 
-  Components which have JavaScript functionality include a data attribute called `[data-module]`,
-  we have added the `govuk-` prefix as a namespace.
+  The `[data-module]` attribute in components now has a `gov-uk` namespace prefix. This is to avoid conflicts between attribute names, and make the attributes consistent with our style for class names.
```
nickcolley: `gov-uk` -> `govuk-`

nickcolley: Conflicts with what? Before it was clear what it was conflicting with I think.

CHANGELOG.md
```diff
@@ -268,72 +268,47 @@
...
 
-  This ensures the component's name does not conflict with other service level components
-  with the same name, which is consistent with our CSS class name convention.
+  The [button](https://design-system.service.gov.uk/components/button/#start-buttons) and [details](https://design-system.service.gov.uk/components/details/) components now also have the `[data-module]` attribute.
```
nickcolley: 'now have' -> 'now use'?

CHANGELOG.md
```diff
@@ -268,72 +268,47 @@
...
 
-  If you're using Nunjucks macros, and are [using the `initAll` function](https://github.com/alphagov/govuk-frontend/blob/master/docs/installation/installing-with-npm.md#option-1-include-javascript) you will not be affected.
+  You'll need to update your CSS or JavaScript code, to make sure that JavaScript can find and initalise the components.
```
nickcolley: I'm not sure CSS is relevant to this.

'JavaScript', 'your JavaScript'

CHANGELOG.md
```diff
@@ -268,72 +268,47 @@
...
 
   ```html
   <div class="govuk-accordion" data-module="govuk-accordion">
-    <!-- ... -->
+  ...
   </div>
   ```
 
-  If you are initialising components manually, you will need to add the `govuk-` prefix to any CSS selectors used to find GOV.UK Frontend components:
+  You'll also need to add `data-module="govuk-button"` to each:
```
nickcolley: This is incorrect since we don't want to add 'govuk-button' to the details component

CHANGELOG.md
```diff
@@ -268,72 +268,47 @@
...
-  If you're using Nunjucks macros, and are [using the `initAll` function](https://github.com/alphagov/govuk-frontend/blob/master/docs/installation/installing-with-npm.md#option-1-include-javascript) you will not be affected.
+  You'll need to update your CSS or JavaScript code, to make sure that JavaScript can find and initalise the components.
 
-  If you are using HTML, you will need to add the `govuk-` prefix to any `[data-module]` attributes.
+  You do not need to do anything if you're using Nunjucks macros and the [`initAll`](https://github.com/alphagov/govuk-frontend/blob/master/docs/installation/installing-with-npm.md#option-1-include-javascript) function.
 
-  For example the accordion component selector has changed from `[data-module="accordion"]` to `[data-module="govuk-accordion"]`.
+  -- If you're using HTML and CSS
+
+  You'll need to add the `govuk-` prefix to the `[data-module]` attribute in each CSS selector.
```
nickcolley: Simplifying this to 'CSS' selector means this is no longer correct, as we want to help people using JavaScript not CSS

nickcolley:
Spoke to Mark and we're going to try and simplify the CHANGELOG later in the release notes so we can unblock other work. We need to pair on this to get it right...

PR #1444, opened 2019-06-11
Renames `js-` css prefix to `govuk-js-`

aliuk2012:
Currently two components(character count & header) have `js-` interaction
hooks. Unfortunately we have found somes instances where other 3rd party js or
legacy js (govuk-elements etc) had some unexpected behaviour with
the components and in some cases raised js errors.

We decided as a team that we should instead namespace (using `govuk-js`)
these hooks so that we have confidence in the initialising and interaction of the components on a page with other scripts.

fixes https://github.com/alphagov/govuk-frontend/issues/1270

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,23 @@
 
 💥 Breaking changes:
 
+- Renamed css class names `js-character-count` and `js-header-toggle`
+
+  Fixes issues with govuk-fronted javascript clashing with govuk-template/govuk-element and potentially other third-party javascript libraries using `js-` class names as hooks.
+
+  HTML markup will only need to change if nunjucks macros are not being used.
+  Update it use `govuk-js-character-count` or `govuk-js-header-toggle`.
+
+  Eg. `<textarea class="govuk-textarea govuk-js-character-count"></textarea>`
+
+  if you have any custom JavaScript relying on either `js-character-count` or `js-header-toggle`, you will
+  need to update it use `govuk-js-character-count` or `govuk-js-header-toggle`.
+
+  Eg. `document.querySelectorAll('.govuk-js-character-count')`
+  Eg. `document.querySelectorAll('.govuk-js-header-toggle')`
+
+  ([PR #1444](https://github.com/alphagov/govuk-frontend/pull/1444))
+
 - Make radios and checkboxes components easier to link to from the error summary component
 
   You should check that clicking a link from the error summary to radios or checkboxes results in the first input being focused.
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,12 @@
 
 💥 Breaking changes:
 
+- Rename css class names prefixed with `js-` to `govuk-js-`
+
+  Fixes issues with govuk-fronted javascript clashing with govuk-template/govuk-element and potentially other third-party javascript libraries using `js-` class names as hooks.
```
nickcolley: How do people migrate?

kr8n3r:
does this change extend to the template? https://github.com/alphagov/govuk-frontend/blob/master/src/template.njk#L31

aliuk2012:
Hi @kr8n3r, 

No it doesn't,  the team decided that `js-enabled` was to widely used for styling components and we could not think of a use case were you would use 'js-enabled' as a hook to add interactions. 

As you pointed out with this line of code, if JavaScript fails or is disabled then 'js-enabled' would not be added to the `body` element and no JavaScript would run anyway.

nickcolley:
One big reason to avoid changing `js-enabled` is that it'd make it harder for users using GOV.UK Template with GOV.UK Frontend

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@
 
   Fixes issues with govuk-fronted javascript clashing with govuk-template/govuk-element and potentially other third-party javascript libraries using `js-` class names as hooks.
 
+  if you have any custom JavaScript relying on any GOV.UK Frontend components having class name prefixed with `.js-`(eg. `.js-header-toggle`) class name then you will need to find and add `.govuk-js-` to it.
```
nickcolley: Could people confuse this advice and add `.govuk-js-` to `.js-enabled` sass?

I wonder since we have only two classes we just make it explicit on how to update specifically those classes...

aliuk2012: Thats a really good point. We are only updating two css classes. I've pushed up another commit that should address this problem and also add instructions for non-nunjuck usage.

CHANGELOG.md
```diff
@@ -4,6 +4,21 @@
...
+
+  Fixes issues with govuk-fronted javascript clashing with govuk-template/govuk-element and potentially other third-party javascript libraries using `js-` class names as hooks.
+
+  if you have any custom JavaScript relying on either `js-character-count` or `js-header-toggle`, you will
+  need to update it use `govuk-js-character-count` or `govuk-js-header-toggle`.
+
+  Eg. `document.querySelectorAll('.govuk-js-character-count')`
+  Eg. `document.querySelectorAll('.govuk-js-header-toggle')`
+
+  ([PR #1444](https://github.com/alphagov/govuk-frontend/pull/1444))
```
nickcolley: I think this should be below the next line.

Issue #1410, opened 2019-05-31
User focus was not taken to the error summary heading as expected in GOV.UK Frontend review application legacy pages


nickcolley:
This issue is from a May 2019 external accessibility audit report.

**WCAG Reference**: Usability feedback only, there is no WCAG related guidelines.
**Issue ID**: DAC_Issue26
**URL**: http://govuk-frontend-v3.herokuapp.com/full-page-examples/applicant-details
**Page**: Applicants details

## Screen Shot
![](https://user-images.githubusercontent.com/2445413/58715320-816dea80-83be-11e9-920d-7394b55b0509.png)

When an error was committed, user focus was not taken to the error summary heading as expected. Screen reader users were required to navigate the page to look for the information.

## Current Code Ref(s)
```html
<h2 class="heading-medium error-summary-heading" id="error-summary-heading-example-1"> There is a problem </h2>
```

## Screen reader comments
> “I committed an error and the error heading appeared on the page as expected. However, my focus was not taken to the error heading above the form. While using JAWS my focus was taken to the top of the page. With NVDA I landed on the first error field meaning I was situated below the error heading. While using VoiceOver for iPhone the page refreshed and my focus was taken to the top of the page. For VoiceOver with Mac, focus landed in the address sidebar. I would find it beneficial if my focus could land on the error heading consistently to ensure that users are immediately aware of the error.”

## Keyboard only comments
> “I found that when I made an error, my focus was not taken to the list of errors at the top of the form as is usually the case with GOV.UK services.”

## Solution
Ensure that keyboard focus is taken to the error summary when an error has been committed. This will enable users to quickly determine that an error has occurred and locate the skip links to jump to the error in question.

nickcolley:
I think this is because we did not include the JavaScript for GOV.UK Elements in this page, and this page includes the GOV.UK Elements version of the error summary component.

Potentially we can retest this when we have the Elements JavaScript running on this page: https://github.com/alphagov/govuk_elements/blob/master/assets/javascripts/application.js#L22

Or we could close this since it's not an issue in the GOV.UK Frontend codebase.

nickcolley:
Confirmed as fixed in a re-test by DAC on 7th October 2019

PR #1459, opened 2019-06-17
Fix legacy error summary component autofocus on full page examples

nickcolley:
This pull request:

1. Ensures legacy javascript is run on the legacy pages, regardless if the flag is set
2. Ensures legacy error summary works
3. Fixes images not being loaded for legacy pages

Fixes https://github.com/alphagov/govuk-frontend/issues/1410

### DIFF
--- a/app/views/full-page-examples/applicant-details/index.njk
+++ b/app/views/full-page-examples/applicant-details/index.njk
@@ -8,25 +8,13 @@ notes: >-
   known accessibility issues.
 ---
 
-{% extends "govuk_template_jinja/views/layouts/govuk_template.html" %}
+{% extends "legacy.njk" %}
 
 {% from "back-link/macro.njk" import govukBackLink %}
 {% from "date-input/macro.njk" import govukDateInput %}
 {% from "button/macro.njk" import govukButton %}
 
-{% set asset_path = "/vendor/govuk_template/" %}
-{% set homepage_url = "/" %}
 {% set page_title = "Applicant details" %}
-{% block page_title %}{{ "Error: " if errors }}{{ page_title }} - GOV.UK{% endblock %}
-
-{% block head %}
-  <!--[if !IE 8]><!-->
-    <link rel="stylesheet" href="/public/app-legacy.css">
-  <!--<![endif]-->
-  <!--[if IE 8]>
-    <link rel="stylesheet" href="/public/app-legacy-ie8.css">
-  <![endif]-->
-{% endblock %}
 
 {% block content %}
   <main id="content" role="main">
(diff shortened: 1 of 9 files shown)


36degrees:
Can you update the PR title so that it describes the changes?

app/views/full-page-examples/applicant-details/index.njk
```diff
@@ -38,7 +38,7 @@ notes: >-
         }) }}
 
        {% if errorSummary.length > 0 %}
-          <div class="error-summary" role="alert" aria-labelledby="error-summary-heading-example-1" tabindex="-1">
+          <div class="error-summary js-error-summary" role="alert" aria-labelledby="error-summary-heading-example-1" tabindex="-1">
```
36degrees: Would it make sense just to remove the extra check as it's not relevant in our review app?

(https://github.com/alphagov/govuk-frontend/pull/1459/files#diff-5c3d7cace66f0b0e3079ba612882dbfcR30)

app/views/full-page-examples/applicant-details/index.njk
```diff
@@ -135,6 +135,7 @@ notes: >-
 {% endblock %}
 
 {% block body_end %}
+  {% include "partials/legacyJavaScript.njk" %}
```
36degrees: Do you think it's worth exploring having a layout for legacy examples?

app/views/partials/legacyJavaScript.njk
```diff
@@ -0,0 +1,46 @@
...
+    GOVUK.primaryLinks.init()
+
+    var showHideContent = new GOVUK.ShowHideContent()
+    showHideContent.init()
+
+    GOVUK.stickAtTopWhenScrolling.init()
+    GOVUK.shimLinksWithButtonRole.init()
+
+    $(window).load(function () {
+      if ($('.js-error-summary').length) {
```
36degrees: Worth adding a comment to explain where this code came from, or putting it in the commit message?

app/views/full-page-examples/applicant-details/index.njk
nickcolley: Yeah we could definitely refactor the views in the demo apps...

36degrees: Non-blocking, happy to leave it for now if you think it makes sense 👍

app/views/full-page-examples/applicant-details/index.njk
nickcolley: Removed the check

app/views/full-page-examples/applicant-details/index.njk
nickcolley: Refactored it a bit, still more could do I imagine.

app/views/partials/legacyJavaScript.njk
nickcolley: Added a comment good shout

Issue #1401, opened 2019-05-31
Input Purpose not specified in GOV.UK Review application


nickcolley:
This issue is from a May 2019 external accessibility audit report.

**WCAG Reference**: [1.3.5 Identify Input Purpose (Level AA)](https://www.w3.org/WAI/WCAG21/Understanding/identify-input-purpose.html)
**Issue ID**: DAC_Issue17
**URL**: Global

## Summary

The autocomplete attribute had not been used on input fields where it was appropriate to do so

## Screen Shot
![](https://user-images.githubusercontent.com/2445413/58711135-c17c9f80-83b5-11e9-84e8-8f260d19a019.png)

The autocomplete attribute had not been used on input fields to enable users’ data to be ‘autofilled’ by the browser. This helps users to complete forms more quickly.

## Current Code Ref(s)
```html
<input class="govuk-input govuk-input--width-10" id="address-postcode" name="address-postcode" type="text">
```

## Cognitive user comments
> "When I was filling out some forms, I noticed that suggestions were not available. I like these as they make forms easy for me to complete.”

## Solution
Use the autocomplete attribute on text inputs to help users’ complete forms more quickly.
Please refer to the following source for a list of examples: https://www.w3.org/TR/WCAG21/#input-purposes

nickcolley:
I have given this a low priority since this is an issue in the review application and not GOV.UK Frontend, we give guidance on this in the design system website, this is something we can update quickly however.

nickcolley:
Confirmed as fixed in a re-test by DAC on 7th October 2019

PR #1461, opened 2019-06-17
Add input purposes to full page examples

nickcolley:
Fixes https://github.com/alphagov/govuk-frontend/issues/1401

### DIFF
--- a/app/views/full-page-examples/what-is-your-address/index.njk
+++ b/app/views/full-page-examples/what-is-your-address/index.njk
@@ -54,7 +54,8 @@ scenario: >-
             id: "address-line-1",
             name: "address-line-1",
             value: values["address-line-1"],
-            errorMessage: errors["address-line-1"]
+            errorMessage: errors["address-line-1"],
+            autocomplete: "address-line1"
           }) }}
 
           {{ govukInput({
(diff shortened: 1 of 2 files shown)


app/views/full-page-examples/what-is-your-address/index.njk
```diff
@@ -97,7 +99,8 @@ scenario: >-
             id: "address-postcode",
             name: "address-postcode",
             value: values["address-postcode"],
-            errorMessage: errors["address-postcode"]
+            errorMessage: errors["address-postcode"],
+            autocomplete: "postcode"
```
aliuk2012: Shouldn't this be `postal-code`?

app/views/full-page-examples/what-is-your-postcode/index.njk
```diff
@@ -46,7 +46,8 @@ scenario: >-
       id: "address-postcode",
       name: "address-postcode",
       autocomplete: "postal-code",
-      errorMessage: errors["address-postcode"]
+      errorMessage: errors["address-postcode"],
+      autocomplete: "postcode"
```
aliuk2012: I don't think this is needed because on line 48 there is already an autocomplete declaration.

Issue #1318, opened 2019-05-03
Document how to update version of govuk-frontend


hannalaakso:
We had a query from a user who was unsure how to update their version of govuk-frontend to the latest version.

Our docs show [how to install govuk-frontend for the first time](https://github.com/alphagov/govuk-frontend/blob/master/docs/installation/installing-with-npm.md#installation) but not how to update to a new version via npm.

Suggested steps we could add to docs:

1. Check [release notes](https://github.com/alphagov/govuk-frontend/releases). 
 <img width="116" alt="Screen Shot 2019-05-03 at 15 53 22" style="float: right;" src="https://user-images.githubusercontent.com/5007934/57145664-9f463080-6dbb-11e9-811a-e7166fae2b37.png"> marks the latest release. Compare this against your version of govuk-frontend, defined in `package.json`, to find out what changes have been introduced since then. See [example](https://github.com/alphagov/govuk-design-system/blob/master/package.json#L34) for where this information lives in GOV.UK Design System.

2. Check if the release is a breaking one (generally most releases won't be). Example of a breaking release title: 

<img width="775" alt="Screen Shot 2019-05-03 at 15 47 13" src="https://user-images.githubusercontent.com/5007934/57145256-c3554200-6dba-11e9-89b8-46cb89ce89e9.png">

 For breaking releases, read the release notes carefully as you might be required to make code changes to your project for `govuk-frontend` to work correctly.

2. To install a new version of govuk-frontend, run

```
npm install govuk-frontend@2.11.0
```
where `2.11.0` is the version you require.

3. _Optional but recommended step_: Watch [`govuk-frontend`](https://github.com/alphagov/govuk-frontend) on Github for "Releases only" to get notified when new releases come out. 

<img width="327" alt="Screen Shot 2019-05-03 at 15 44 29" src="https://user-images.githubusercontent.com/5007934/57145091-62c60500-6dba-11e9-824c-774c159d65c5.png">

PR #1463, opened 2019-06-18
Add docs for updating version of GOV.UK Frontend

hannalaakso:
Fixes https://github.com/alphagov/govuk-frontend/issues/1318

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,7 +105,7 @@ https://accessibility.blog.gov.uk/2017/03/27/how-users-change-colours-on-website
 To be notified when there’s a new release, you can either:
 
 - [watch the govuk-frontend Github repository](https://help.github.com/en/articles/watching-and-unwatching-repositories)
-- subscribe to the [#govuk-design-system channel on cross-government Slack](https://ukgovernmentdigital.slack.com/app_redirect?channel=govuk-design-system)
+- join the [#govuk-design-system channel on cross-government Slack](https://ukgovernmentdigital.slack.com/app_redirect?channel=govuk-design-system)
 
 Find out how to [update with npm](https://github.com/alphagov/govuk-frontend/blob/master/docs/installation/updating-with-npm.md).
 
--- a/README.md
+++ b/README.md
@@ -33,6 +33,10 @@ We recommend [installing GOV.UK Frontend using node package manager
 You can also [download the compiled and minified assets (CSS, JavaScript) from
 GitHub](docs/installation/installing-from-dist.md).
 
+## Getting updates
+
+See [updating with npm](https://github.com/alphagov/govuk-frontend/blob/master/docs/installation/updating-with-npm.md).
+
 ## Browser support
 
 GOV.UK Frontend will allow you to build services that comply with the [guidance
--- a/README.md
+++ b/README.md
@@ -33,10 +33,6 @@ We recommend [installing GOV.UK Frontend using node package manager
 You can also [download the compiled and minified assets (CSS, JavaScript) from
 GitHub](docs/installation/installing-from-dist.md).
 
-## Getting updates
-
-See [updating with npm](https://github.com/alphagov/govuk-frontend/blob/master/docs/installation/updating-with-npm.md).
-
 ## Browser support
 
 GOV.UK Frontend will allow you to build services that comply with the [guidance
@@ -103,6 +99,15 @@ https://support.microsoft.com/en-gb/help/13862/windows-use-high-contrast-mode
 
 [how-users-change-colours-on-websites]:
 https://accessibility.blog.gov.uk/2017/03/27/how-users-change-colours-on-websites/
+
+## Getting updates
+
+To be notified when there’s a new release, you can either:
+
+- [watch the govuk-frontend Github repository](https://help.github.com/en/articles/watching-and-unwatching-repositories)
+- subscribe to the [#govuk-design-system channel on cross-government Slack](https://ukgovernmentdigital.slack.com/app_redirect?channel=govuk-design-system)
+
+Find out how to [update with npm](https://github.com/alphagov/govuk-frontend/blob/master/docs/installation/updating-with-npm.md).
 
 ## Licence
 
--- a/docs/installation/updating-with-npm.md
+++ b/docs/installation/updating-with-npm.md
@@ -1,0 +1,31 @@
+# Updating GOV.UK Frontend with node package manager (NPM)
+
+Follow these instructions if you have previously installed the GOV.UK Frontend npm package following the [installation instructions](https://github.com/alphagov/govuk-frontend/blob/master/docs/installation/installing-with-npm.md#requirements), and want to update to a more recent version.
+
+## Steps to update
+
+### 1. Find out changes introduced in recent versions.
+
+To do this, view `package.json` at the root of your project to find out the version of govuk-frontend that your app is currently running, for instance "2.11.0". See [example](https://github.com/alphagov/govuk-design-system/blob/master/package.json#L34) for where this information lives in GOV.UK Design System.
+
+Review [Release notes](https://github.com/alphagov/govuk-frontend/releases) to see what changes were introduced since this version was released.
+
+### 2. Choose the version to update to.
+
+You can update to the latest release, or to an earlier version.
+
+If you update to the last non-breaking version of GOV.UK Frontend, you shouldn't need to make any changes to your app. If you update to a breaking release, you might need to make changes to your app to make it work correctly with GOV.UK Frontend. [Release notes](https://github.com/alphagov/govuk-frontend/releases) will have information in the "Breaking changes" section of that release about the changes you might need to make.
+
+### 3. Update the version
+
+Run
+
+```
+npm install govuk-frontend@2.13.0
+```
+
+where 2.13.0 is the version you require.
+
+## Find out about new releases
+
+You can watch [govuk-frontend](https://github.com/alphagov/govuk-frontend) for "Releases only" to get notified of new releases.
--- a/docs/installation/updating-with-npm.md
+++ b/docs/installation/updating-with-npm.md
@@ -1,4 +1,4 @@
-# Updating GOV.UK Frontend with node package manager (NPM)
+# Updating GOV.UK Frontend using node package manager (NPM)
 
 You can update with node package manager (NPM) if you [originally installed
 GOV.UK Frontend with NPM](https://github.com/alphagov/govuk-frontend/blob/master/docs/installation/installing-with-npm.md#requirements).
@@ -25,15 +25,17 @@ notes](https://github.com/alphagov/frontend/releases) in the GOV.UK Frontend Git
 
 You can update to any version.
 
-You'll need to progressively update to each version following your current
-version. You will not need to make code changes for GOV.UK Frontend to keep working in
-your project, unless the release note for the version has a ‘Breaking changes’
-section.
+You'll need to make code changes to make sure GOV.UK Frontend keeps working in your project, if you’re updating either:
 
-To update, run:
+- to a major version - for example you’re updating to 2.0.0
+- past a major version - for example you’re updating from 1.2.0 to 2.3.0
+
+A version is major ('breaking') if there's a 'Breaking changes' section in that version's release notes.
+
+To update to the most recent version, run:
 
 ```shell
-npm install govuk-frontend@2.13.0
+npm install govuk-frontend@latest
 ```
 
-Replace `2.13.0` with the version you want to update to.
+If you want to install a specific version, replace `@latest` with the version that you want to update to. For example `2.13.0`.
--- a/docs/installation/updating-with-npm.md
+++ b/docs/installation/updating-with-npm.md
@@ -1,7 +1,7 @@
-# Updating GOV.UK Frontend using node package manager (NPM)
+# Updating GOV.UK Frontend using node package manager (npm)
 
-You can update with node package manager (NPM) if you [originally installed
-GOV.UK Frontend with NPM](https://github.com/alphagov/govuk-frontend/blob/master/docs/installation/installing-with-npm.md#requirements).
+You can update with node package manager (npm) if you [originally installed
+GOV.UK Frontend with npm](https://github.com/alphagov/govuk-frontend/blob/master/docs/installation/installing-with-npm.md#requirements).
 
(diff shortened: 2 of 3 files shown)


docs/installation/updating-with-npm.md
```diff
@@ -0,0 +1,31 @@
+# Updating GOV.UK Frontend with node package manager (NPM)
+
+Follow these instructions if you have previously installed the GOV.UK Frontend npm package following the [installation instructions](https://github.com/alphagov/govuk-frontend/blob/master/docs/installation/installing-with-npm.md#requirements), and want to update to a more recent version.
+
+## Steps to update
+
+### 1. Find out changes introduced in recent versions.
+
+To do this, view `package.json` at the root of your project to find out the version of govuk-frontend that your app is currently running, for instance "2.11.0". See [example](https://github.com/alphagov/govuk-design-system/blob/master/package.json#L34) for where this information lives in GOV.UK Design System.
```
nickcolley: If we want to include an example could consider inlining a snippet rather than pointing to a file that can change.

However, If we keep pointed to the design system code base it should be permalinked.

docs/installation/updating-with-npm.md
```diff
@@ -0,0 +1,31 @@
...
+
+```
+npm install govuk-frontend@2.13.0
+```
+
+where 2.13.0 is the version you require.
+
+## Find out about new releases
+
+You can watch [govuk-frontend](https://github.com/alphagov/govuk-frontend) for "Releases only" to get notified of new releases.
```
nickcolley: Could we add a screenshot here? Not sure people will know how to do this.

docs/installation/updating-with-npm.md
36degrees: What about asking people to run `npm list govuk-frontend` instead?

```
$ npm list govuk-frontend
govuk_design_system@ /Users/oliverbyford/Code/govuk-design-system
└── govuk-frontend@2.11.0
```

m-green:
I've committed the edited version that @hannalaakso and I put together yesterday. Please let me know if we've lost any needed detail.

docs/installation/updating-with-npm.md
```diff
@@ -0,0 +1,39 @@
...
+````
+
+## Update
+
+You can check the [release
+notes](https://github.com/alphagov/frontend/releases) in the GOV.UK Frontend Github repository to find out the latest version of GOV.UK Frontend.
+
+You can update to any version.
+
+You'll need to progressively update to each version following your current
```
36degrees: This reads to me as though if e.g. you're on `v2.1` and you want to update to `v2.4`, you have to first update to `v2.2`, then `v2.3` and then finally `v2.4`, which shouldn't be the case.

What is it we're trying to convey here?

README.md
```diff
@@ -100,6 +100,15 @@ https://support.microsoft.com/en-gb/help/13862/windows-use-high-contrast-mode
 [how-users-change-colours-on-websites]:
 https://accessibility.blog.gov.uk/2017/03/27/how-users-change-colours-on-websites/
 
+## Getting updates
+
+To be notified when there’s a new release, you can either:
+
+- [watch the govuk-frontend Github repository](https://help.github.com/en/articles/watching-and-unwatching-repositories)
+- subscribe to the [#govuk-design-system channel on cross-government Slack](https://ukgovernmentdigital.slack.com/app_redirect?channel=govuk-design-system)
```
36degrees: Slack tends to talk about '[joining channels](https://get.slack.help/hc/en-gb/articles/205239967-Join-a-channel)' rather than 'subscribing to a channel', and I'm not aware of that phrase really being used.

Should we align with how Slack talks about these things, to avoid potential confusion? (for example, a user interpreting this to mean that you can subscribe to get channel updates by email)

docs/installation/updating-with-npm.md
```diff
@@ -0,0 +1,39 @@
...
+version. You will not need to make code changes for GOV.UK Frontend to keep working in
```
36degrees: Do we want to talk about semantic versioning here? It _might_ be easier for the user if they know that e.g. a 2.x to 2.y change will not include breaking changes, rather than requiring them to read the release notes to find that out?

docs/installation/updating-with-npm.md
```diff
@@ -0,0 +1,39 @@
...
+your project, unless the release note for the version has a ‘Breaking changes’
+section.
+
+To update, run:
+
+```shell
+npm install govuk-frontend@2.13.0
```
36degrees: Should we use `@latest` here, as that'll be the most common use case?

The sentence below could then be something like:

> If you want to install a specific version, replace `@latest` with the version that you want to update to, for example `npm install govuk-frontend@2.13.0`

docs/installation/updating-with-npm.md
```diff
@@ -0,0 +1,39 @@
+# Updating GOV.UK Frontend with node package manager (NPM)
```
36degrees: I don't know why, but '_with_ node package manager' just doesn't sound quite right to me. Have you considered 'using' instead?

```suggestion
# Updating GOV.UK Frontend using node package manager (NPM)
```

README.md
m-green: Make sense! I'll tweak.

docs/installation/updating-with-npm.md
m-green: Good point - I'll tweak this.

docs/installation/updating-with-npm.md
m-green: So something like: "You will not need to make code changes for GOV.UK Frontend to keep working in your project, unless you're updating to a major version - for example 2.0 rather than 2.13.0"?

docs/installation/updating-with-npm.md
m-green: Sounds good - will amend.

docs/installation/updating-with-npm.md
hannalaakso: Thanks @36degrees. We decided to keep both methods in case someone doesn't have access to the terminal.

docs/installation/updating-with-npm.md
nickcolley: Could I propose for our style that we avoid 

'major', 'minor', 'patch'

and instead always use

'breaking', 'feature', 'fix'?

I think these are more understandable and reflect the CHANGELOG

36degrees: I like that idea a lot when we're talking about an individual release (e.g. 3.0.0 is a 'breaking release'), but I don't know how well it works when we're talking about how _version numbers_ change?

For example, someone updating from v2.13.0 to v3.1.1 is updating to a 'fix' release, but because the major version number changes they know there'll be breaking changes anyway.

We could talk about 'if the major version number changes' (having explained that term somewhere) but I'm not sure we'd ever say 'if the breaking version number changes'

nickcolley: I think that's just a matter of how it's worded specifically what I was responding to:

> unless you're updating to a major version

would be

> unless you're updating to a version with breaking changes

etc

something to talk about later probs

m-green: How about combining the two, so we clarify it for anyone who might not be clear on either term? Something like: "...unless you're updating to a major version. A version is major if it has a breaking changes section in its release notes."

36degrees: >"...unless you're updating to a major version. A version is major if it has a breaking changes section in its release notes."

It's slightly more nuanced than that though, because the specific version they're updating to is irrelevant unless they happen to be running the immediate preceding release. In the example I gave above, going from v2.13.0 to v3.1.1 is _not_ 'updating to a major version' (the version they're updating to is a patch release!) but the major version number does change.

I guess we need to work out what's easier for the user – going through the release notes to understand whether there was a breaking change _at some point_, or understanding the basics of how major, minor and patch version numbers work?

m-green: How about this (to cover this discussion and Ollie's comment about 'progressive' updates too):

You will need to make code changes to make sure GOV.UK keeps working in your project, if you’re updating:

- to a major version - for example you’re updating to 2.0.0
- past a major version - for example you’re updating from 1.2.0 to 2.3.0

A version is major ('breaking') if there's a 'Breaking changes' section in the release notes.

36degrees: That sounds ace, thanks @m-green 👏 

I wonder if it should say 'You *may* need to make code changes' (sometimes, the 'breaking' part of the release might not affect a user) and whether we need to point the user back to the release notes so that they know what those code changes might be?

docs/installation/updating-with-npm.md
```diff
@@ -0,0 +1,41 @@
...
+
+```shell
+npm list govuk-frontend
+```
+
+If you do not have command line access, you can see the version number in the
+`package.json` file in the root of your project directory. For example:
+
+```json
+'gov-uk frontend': ‘2.11.0’
```
nickcolley: `govuk-frontend`

docs/installation/updating-with-npm.md
```diff
@@ -0,0 +1,41 @@
...
+````
+
+## Update
+
+You can check the [release
+notes](https://github.com/alphagov/frontend/releases) in the GOV.UK Frontend Github repository to find out the latest version of GOV.UK Frontend.
```
nickcolley: Git**H**ub

docs/installation/updating-with-npm.md
```diff
@@ -0,0 +1,41 @@
...
+
+You can update to any version.
```
nickcolley: Not sure this sentence is needed?

docs/installation/updating-with-npm.md
```diff
@@ -0,0 +1,41 @@
...
+
+You may need to make code changes to make sure GOV.UK Frontend keeps working in your project, if you’re updating either:
+
+- to a major version - for example you’re updating to 2.0.0
```
nickcolley: These two bullet points seem to duplicate themselves.

Sorry to bring it up again but I don't think there's value saying 'major' here.

I would simplify this to something like

> If the release has breaking changes, you may need to make code changes to your project for it to work.

36degrees: Can we talk through this in person on Monday?

docs/installation/updating-with-npm.md
```diff
@@ -0,0 +1,43 @@
...
+`package.json` file in the root of your project directory. For example:
+
+```json
+'govuk-frontend': ‘2.11.0’
+````
+
+## Update
+
+You can check the [release
+notes](https://github.com/alphagov/frontend/releases) in the GOV.UK Frontend GitHub repository to find out the latest version of GOV.UK Frontend.
```
nickcolley:
```suggestion
notes](https://github.com/alphagov/govuk-frontend/releases) in the GOV.UK Frontend GitHub repository to find out the latest version of GOV.UK Frontend.
```

docs/installation/updating-with-npm.md
```diff
@@ -18,26 +16,19 @@ If you do not have command line access, you can see the version number in the
...
+You may need to make code changes to keep GOV.UK Frontend working in your project, if the major version number will change when you update. The major version number is the first digit in the version number.
 
 To update to the most recent version, run:
 
 ```shell
 npm install govuk-frontend@latest
 ```
 
-If you want to install a specific version, replace `@latest` with the version that you want to update to. For example `2.13.0`.
+If you want to install an earlier version, replace `latest` with the version that you want to update to. For example `2.13.0`.
```
m-green: @hannalaakso In retrospect I thought it was best in the end to remove the 'You can update to any version' line. Because with a tweak to this line 34 - to say "earlier" rather than "specific", I think we cover off the idea that you can update to any version?

hannalaakso: @m-green I'm happy with that 👍

docs/installation/updating-with-npm.md
```diff
@@ -0,0 +1,34 @@
+# Updating GOV.UK Frontend using node package manager (NPM)
+
+You can update with node package manager (NPM) if you [originally installed
```
bravegrape: What's the reasoning behind the "NPM" abbreviation?
The product is branded as "npm" (https://www.npmjs.com/) so we could probably do away with the all caps abbreviation.
Also since the product is branded as "npm", I'm wondering whether we should have it as "node package manager, npm" or similar 🤔

docs/installation/updating-with-npm.md
```diff
@@ -0,0 +1,34 @@
...
+GOV.UK Frontend with NPM](https://github.com/alphagov/govuk-frontend/blob/master/docs/installation/installing-with-npm.md#requirements).
+
+## Find out which version you're using
+
+You can find out which version of GOV.UK Frontend your project is using with:
```
bravegrape:
```suggestion
You can find out which version of GOV.UK Frontend your project is using by running:
```
Just a suggestion - it's a common docs pattern for commands that need to be run in the Terminal.

docs/installation/updating-with-npm.md
nickcolley: Since the audience for this guidance is web developers, the majority of them will not know npm by it's extended name.

The tool itself is nearly always referred to as `npm`, for example when you use it `npm install`

You can see on Google Trends when you compare the two, there is nearly no activity for the extended version. https://trends.google.com/trends/explore?geo=GB&q=npm,node%20package%20manager

So in this case I would only expand it once and then always use 'npm' for the rest of the guidance.

It's also worth nothing that npm should be lowercase not uppercase.

m-green: Thanks both. Sounds like we're in agreement - expand the first time (keeping to style), use lowercase 'npm' (as that's the name). I'll fix that up.

@bravegrape, the brackets/comma question is an interesting one! I think our style is still brackets even if the acronym is the known name / product name. A similar-ish GOV.UK example is https://www.gov.uk/make-a-sorn - most people know it as SORN and definitely not 'Statutory Off Road Notification', but we keep the brackets. What do you think?

docs/installation/updating-with-npm.md
m-green: Thanks Andrea - amended.

docs/installation/updating-with-npm.md
bravegrape: Sounds good to me - let's follow the established pattern 👍

PR #1466, opened 2019-06-19
[As part of release] Updates the docs to point to the right files

aliuk2012:
As part of namespacing, we need to update the links the docs point to
so that they are still valid and do not break.

### DIFF
--- a/docs/contributing/coding-standards/components.md
+++ b/docs/contributing/coding-standards/components.md
@@ -1,6 +1,6 @@
 # Components
 
-Find components in `src/components`.
+Find components in `src/govuk/components`.
 
 Components must use the `.govuk-` namespace.
 
@@ -19,10 +19,6 @@ Components must:
 ### Component folder structure and naming
 
 Component folder and files should be singular, except in cases where they are more commonly used in groups, for example, radios, breadcrumbs and checkboxes.
-
-An example component exists in `src/components/component-example`.
-
-Use this as the basis for creating new components.
 
 The folder structure should be:
 
--- a/docs/installation/installing-with-npm.md
+++ b/docs/installation/installing-with-npm.md
@@ -42,14 +42,14 @@ imports) if you want to override GOV.UK Frontend with your own styles.
 1. To import all components, add the below to your Sass file:
 
   ```SCSS
-  @import "node_modules/govuk-frontend/all";
+  @import "node_modules/govuk-frontend/govuk/all";
   ```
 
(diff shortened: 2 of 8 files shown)


docs/installation/installing-with-npm.md
```diff
@@ -283,7 +283,7 @@ To use different asset paths, also complete the below step(s).
   @import "govuk-frontend/all";
```
nickcolley: Nice work so far, I think we need to make sure the imports now have the directory too?

docs/installation/installing-with-npm.md
```diff
@@ -200,6 +198,12 @@ If you're using a bundler such as Webpack, use the `import` syntax to import a c
 import { Radios } from 'govuk-frontend'
 ```
 
+or
+
+```JS
+import Button from 'govuk-frontend/govuk/components/button/button'
```
36degrees: What does this offer over `import { Button } from 'govuk-frontend'`? Are there scenarios when `import { Button } from 'govuk-frontend'` wouldn't work and the longer form is needed?

Just wondering if we should keep this simpler – by providing two options we're giving users another decision to make. If we need both, it'd be good to be clear when you might use each option.

nickcolley: Agreed, let's not have two ways here, and keep the original.

nickcolley:
Unless we change the default branch to a v2.X tag merging this would make the installation docs incorrect until we do the v3.0.0 release so I've put a blocked label on this.

docs/installation/installing-with-npm.md
aliuk2012: 👍 done

aliuk2012:
I've addressed @36degrees comment on not having two options to import js. Hopefully this is now in a state to be ready to merge once we do a v3 release

docs/contributing/coding-standards/components.md
```diff
@@ -20,8 +20,6 @@ Components must:
 
 Component folder and files should be singular, except in cases where they are more commonly used in groups, for example, radios, breadcrumbs and checkboxes.
 
-An example component exists in `src/components/component-example`.
-
 Use this as the basis for creating new components.
```
m-green: Should this line be removed as well? 'This' seems to refer to the example component that we've now removed reference to.

36degrees: Yep, I'll update this now.

Great spot 👀

PR #1468, opened 2019-06-21
Namespace the assets

aliuk2012:


aliuk2012:
We decided not to do this change for the following reasons:

- We don't know its an actual issue that users are experiencing
- We would need to add further guidance on how to update
- It shouldn't be a problem as users can update the sass variables to include `govuk` if they do run into difficulties.

PR #1481, opened 2019-07-08
Improve reasoning for 'tabIndex' property usage

nickcolley:
We normally use `setAttribute` but if we were to set the attribute to the correct `tabindex` value it fails in IE6/7

I consider this a documentation pull request, as it's only really useful for maintainers so I won't be adding a CHANGELOG entry.

### DIFF
--- a/src/govuk/components/details/details.js
+++ b/src/govuk/components/details/details.js
@@ -89,7 +89,9 @@ Details.prototype.init = function () {
   $summary.setAttribute('aria-controls', $content.id)
 
   // Set tabIndex so the summary is keyboard accessible for non-native elements
-  // http://www.saliences.com/browserBugs/tabIndex.html
+  //
+  // We have to use the camelcase `tabIndex` property as there is a bug in IE6/IE7 when we set the correct attribute lowercase:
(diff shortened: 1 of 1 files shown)


36degrees:
According to the linked page, `setAttribute('tabIndex', …)` _does_ work, as long as you use camel-case (`tabIndex`, not `tabindex`)?

Should we be using `setAttribute` for consistency?

nickcolley:
`setAttribute('tabIndex', …)` is not correct though, the attribute should be lowercase

36degrees:
When you say it's _not correct_, what do you mean? According to the [linked page](http://web.archive.org/web/20170120194036/http://www.saliences.com/browserBugs/tabIndex.html), using `setAttribute('tabIndex')` does work 'in all browsers' (though that may be out of date).

There's two things going on here (setting the property directly rather than using `setAttribute`; using camel-case for `tabIndex`) but they seem to be conflated.

nickcolley:
Tab index can be set in two ways:

1. By the property which is camelcase 'tabIndex'
2. By the attribute which is lowercase 'tabindex'

To set the attribute to 'tabIndex' is incorrect and feels like we'd be prioritising convention over valid HTML?

PR #1458, opened 2019-06-17
Namespace nunjucks and components

aliuk2012:
To namespace GOV.UK Frontend components, a new `govuk` directory has been added to the `package` and `src` directories. This means that users should now be able to run GOV.UK Frontend along side their own departmental frontend without any conflicts. There was no need to add `govuk` directory to `dist` or `public` as they would only contain compiled code and be used differently from `package` and `src`

## Changes to `src` directory

<details>
<summary>Before</summary>

```
src
├── assets
├── components
├── core
├── helpers
├── objects
├── overrides
├── settings
├── tools
├── utilities
├── vendor
├── README.md
├── all-ie8.scss
├── all.js
├── all.scss
├── all.test.js
├── common.js
├── template.njk
└── template.test.js
```

</details>

<details>
<summary>After</summary>

```
src
└── govuk
    ├── assets
    ├── components
    ├── core
    ├── helpers
    ├── objects
    ├── overrides
    ├── settings
    ├── tools
    ├── utilities
    ├── vendor
    ├── README.md
    ├── all-ie8.scss
    ├── all.js
    ├── all.scss
    ├── all.test.js
    ├── common.js
    ├── template.njk
    └── template.test.js
```

</details>

## Changes to `package`

<details>
<summary>Before</summary>

```
package
├── assets
├── components
├── core
├── helpers
├── objects
├── overrides
├── settings
├── tools
├── utilities
├── vendor
├── README.md
├── all-ie8.scss
├── all.js
├── all.scss
├── common.js
├── govuk-prototype-kit.config.json
├── package.json
└── template.njk
```

</details>

<details>
<summary>After</summary>

```
package
├── govuk
│   ├── assets
│   ├── components
│   ├── core
│   ├── helpers
│   ├── objects
│   ├── overrides
│   ├── settings
│   ├── tools
│   ├── utilities
│   ├── vendor
│   ├── all-ie8.scss
│   ├── all.js
│   ├── all.scss
│   ├── common.js
│   └── template.njk
├── README.md
├── govuk-prototype-kit.config.json
└── package.json
```

</details>

## Tested:
- `npm start` still works
- `npm run pre-release` still works
- `npm run release-to-branch` still works
- check no changes to `dist` and `public` directories.

<details>
<summary>Details</summary>

</details>

PR #1492, opened 2019-07-12
Update with edited version of v3.0.0 (unreleased) changelog

m-green:
This is the edited version of the v3.0.0 release note in CHANGELOG.md. (The same content - with a summary and updating links - will be used in the GitHub release notes this time).

It's already been 2i'd for style by another technical writer.

Please review for:
- factual accuracy
- any missing changes in this release.

Following review and before publication we need to:
- add missing links
- copy to the GitHub release note and add the summary

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,510 +2,321 @@
 
 ## Unreleased 3.0.0 (Breaking release)
 
-See the [versioning documentation for how to update this
-changelog](./docs/contributing/versioning.md#updating-changelog).
+## Breaking changes
+You must make the following changes when you migrate to this release, or your service may break.
 
-- Renamed css class names `js-character-count` and `js-header-toggle`
+### Update file paths, attributes and class names
+To make sure your project's files do not conflict with GOV.UK Frontend, we've moved our package files into a directory called `govuk`.
 
-  Fixes issues with govuk-fronted javascript clashing with govuk-template/govuk-element and potentially other third-party javascript libraries using `js-` class names as hooks.
+#### If you’re using Sass
 
-  HTML markup will only need to change if nunjucks macros are not being used.
-  Update it use `govuk-js-character-count` or `govuk-js-header-toggle`.
+Add `govuk/` to `@import` paths in your [Sass](https://sass-lang.com/) file that have a `govuk-frontend/` path.
 
-  Eg. `<textarea class="govuk-textarea govuk-js-character-count"></textarea>`
+For example:
 
-  if you have any custom JavaScript relying on either `js-character-count` or `js-header-toggle`, you will
-  need to update it use `govuk-js-character-count` or `govuk-js-header-toggle`.
+```scss
+@import "node_modules/govuk-frontend/govuk/all";
+```
+If you’ve [added `node_modules/govuk-frontend` as a Sass include path](https://github.com/alphagov/govuk-frontend/blob/268b1e1b962d4bd293c5681868a2119f8b4ece60/docs/installation/installing-with-npm.md#optional-resolving-scss-import-paths), add `govuk/` to your `@import` paths:
 
-  Eg. `document.querySelectorAll('.govuk-js-character-count')`
-  Eg. `document.querySelectorAll('.govuk-js-header-toggle')`
+```scss
+@import "govuk/all";
+```
 
-  ([PR #1444](https://github.com/alphagov/govuk-frontend/pull/1444))
+####  If you’re using Javascript
 
-- Make radios and checkboxes components easier to link to from the error summary component
+You must do the following.
 
-  You should check that clicking a link from the error summary to radios or checkboxes results in the first input being focused.
+1. Update file paths.
+2. Update and add `data-module` attributes.
+3. Update CSS class names.
 
-  The `id` of the first input in a set of radios or checkboxes will no longer be suffixed with `-1`. This means that it will be the same as your `idPrefix`, or if `idPrefix` is not set it'll be the same as the `name`.
+##### Update file paths
 
-  We've made this change to make it easier to link from the error summary to radios or checkboxes, if you we're working around this before by setting the first input's `item.id` to the same as the `name` you can now remove this.
+You must add `govuk/` to your import file paths.
 
-  To migrate you can update your error summary so it links to the first element, for example if you are using an `idPrefix` option of `question`, your error summary can now link to `#question` instead of `#question-1`.
+If you're importing `node_modules/govuk-frontend/all.js` in your project, change this file path to `node_modules/govuk-frontend/govuk/all.js`.
 
-  If you cannot change your error summary, you could change your component to set the `item.id` option for the first input, for example:
+If you’re importing a specific path, add `govuk/` after 'govuk-frontend/`. For example, if you're importing the button component:
 
-  ```javascript
-  {{ govukRadios({
-    idPrefix: "question",
-    name: "question",
-    items: [
-      {
-        id: "question-1",
-        value: "yes",
-        text: "Yes"
-      },
-      {
-        value: "no",
-        text: "No"
-      }
-    ]
-  }) }}
-  ```
+```nunjucks
+import Button from 'govuk-frontend/govuk/components/button/button'
+```
 
-  ([PR #1426](https://github.com/alphagov/govuk-frontend/pull/1426))
+##### Update and add data-module attributes
 
-- Update the font to v2
+You do not need to do anything if you're using Nunjucks macros and the [`initAll`](https://github.com/alphagov/govuk-frontend/blob/268b1e1b962d4bd293c5681868a2119f8b4ece60/docs/installation/installing-with-npm.md#option-1-include-javascript) function.
 
-  The new version of the font is significantly smaller and its baseline has been
-  optimised for the web, which means it's more consistent with other fonts.
+If you are not using the Nunjucks macros, add a `govuk-` prefix to `data-module` attribute values. For example:
 
-  This means we can remove the adjustments that were previously required to
-  vertically align text visually within boxes.
+```html
+<div class="govuk-accordion" data-module="govuk-accordion">
+...
+</div>
+```
 
-  For more details see the original issue logged by [@Nooshu](https://github.com/Nooshu) (https://github.com/alphagov/govuk_template/pull/357 & https://github.com/alphagov/govuk-frontend/issues/1012).
+The [button](https://design-system.service.gov.uk/components/button/) and [details](https://design-system.service.gov.uk/components/details/) components now also use the `data-module` attribute for initialisation. If you are not using Nunjucks macros, add:
 
-  The original version of the font is no longer included. If you're using GOV.UK
-  Frontend alongside GOV.UK Template then the version of the font from GOV.UK
-  Template will be used, as long as you've enabled [compatibility mode].
+- `data-module="govuk-button"` to each `<button>` HTML tag
+- `data-module="govuk-details"` to each `<details>` HTML tag
 
-  To migrate: if you have any custom components you may want to check their vertical
-  alignment
+If you're using your own JavaScript code to initialise components, add a `govuk-` prefix to any selectors that find components.
 
-  ([PR #1434](https://github.com/alphagov/govuk-frontend/pull/1434))
+[Pull request #1443: Ensure GOV.UK Frontend component selectors cannot conflict when initialised](https://github.com/alphagov/govuk-frontend/pull/1443)
 
-- Remove `govuk-focusable`, `govuk-focusable-fill` mixins, introduce `govuk-focus-text` mixin.
+##### Update CSS class names
 
-  To migrate:
+You do not need to do anything if you're using Nunjucks.
 
-  For many use cases where `govuk-focusable-fill` was used you can now use `govuk-focused-text`.
+If you're using HTML or custom JavaScript, change:
 
-  Before
-  ```scss
-  @include govuk-focusable-fill;
-  ```
+- `js-character-count` to `govuk-js-character-count`
+- `js-header-toggle` in the GOV.UK Frontend header component to `govuk-js-header-toggle`
 
-  After
-  ```scss
-  &:focus {
-    @include govuk-focused-text;
-  }
-  ```
+[Pull request #1444: Renames `js-` css prefix to `govuk-js-`](https://github.com/alphagov/govuk-frontend/pull/1444)
 
-  For the `govuk-focusable` mixin, there is no general purpose mixin to replace this,
-  please read the new [TODO focus how to guide on the Design System website](#[TODO]).
+#### If you’re using Nunjucks
 
-  ([PR #1361](https://github.com/alphagov/govuk-frontend/pull/1361))
+1. Change your `nunjucks.configure` property so it only includes the  `node_modules/govuk-frontend/` directory:
 
-- Update positioning on table headers and cells to help improve readability
+```nunjucks
+nunjucks.configure([
+  "node_modules/govuk-frontend/"
+])
+```
 
-  To migrate: If you rely on a centered certain vertical alignment, you could add
-  a custom class to your application.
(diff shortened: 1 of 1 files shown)


nickcolley:
Keeping track of pull requests we'll need to include as links:

- focus updates: https://github.com/alphagov/govuk-frontend/pull/1491
- main wrapper release note will need updating based on https://github.com/alphagov/govuk-frontend/pull/1493

CHANGELOG.md
```diff
@@ -2,510 +2,321 @@
...
+[Pull request #1367: Allow for classes, rowspan, colspan and attributes on row headers](https://github.com/alphagov/govuk-frontend/pull/1367). Thanks to [edwardhorsford](https://github.com/edwardhorsford).
 
-- Fixes govuk-template focussed link colour overriding govuk-frontends
+### Use page wrappers without a modifier
 
-  govuk-template uses specific css selector`a:link:focus` whereas
-  govuk-frontend uses `.govuk-link:focus`. This fix makes govuk-frontend
-  selector more specific `.govuk-link:link:focus` but only when compatibility
-  mode is being used.
+If your page wrapper does not have components above the main section, you no longer need to use the [`govuk-main-wrapper--l` modifier class](https://design-system.service.gov.uk/styles/layout/#exploded-view-of-page-wrappers-without-a-back-link-breadcrumbs-or-phase-banner) to increase the vertical space above the content.
```
nickcolley: This needs to be updated based on https://github.com/alphagov/govuk-frontend/pull/1493

CHANGELOG.md
```diff
@@ -2,510 +2,321 @@
 
 ## Unreleased 3.0.0 (Breaking release)
```
nickcolley: We need to do a link check...

CHANGELOG.md
```diff
@@ -2,510 +2,321 @@
...
 
-  To migrate, you'll need to update any places in your code where you use the
-  `@govuk-grid-column` mixin.
+- `data-module="govuk-button"` to each `<button>` HTML tag
+- `data-module="govuk-details"` to each `<details>` HTML tag
 
-  If you are currently passing a class name (`$class: 'foo'`) you should remove
-  the `$class` argument and wrap your call to the mixin in the class name
-  instead:
+If you're using your own JavaScript code to initialise components, add a `govuk-` prefix to any selectors that find components.
```
hannalaakso: I guess they only need to do this if they've been hooking onto the `data-module` to initialise. If they hook onto the CSS class that hasn't changed.

CHANGELOG.md
```diff
@@ -2,510 +2,321 @@
...
 
-  If you are currently passing `$class: false` then you can just remove the
-  `$class` argument, as this is now the default behaviour.
+##### Update CSS class names
 
-  ([PR #1376](https://github.com/alphagov/govuk-frontend/pull/1376))
+You do not need to do anything if you're using Nunjucks.
 
-- Ensure GOV.UK Frontend component selectors cannot conflict when initialised
+If you're using HTML or custom JavaScript, change:
```
hannalaakso: What would be an example of using custom Javascript in this context?

CHANGELOG.md
```diff
@@ -2,510 +2,321 @@
...
 
-  Components which have JavaScript functionality include a data attribute called `[data-module]`,
-  we have added the `govuk-` prefix as a namespace.
+- `js-character-count` to `govuk-js-character-count`
```
hannalaakso: Nit picky but for consistency should this say in the GOV.UK Frontend character count component

CHANGELOG.md
```diff
@@ -2,510 +2,321 @@
...
-  If you are initialising the button component manually, you will need to reference this `[data-module]` attribute:
+### Migrate to the new accessible focus states
 
-  ```javascript
-  var $buttons = document.querySelector('[data-module="govuk-button"]')
-  ```
+The focus state of components now meets the new WCAG 2.1 level AA requirements.
 
-  ### Details component
+You must [update your component’s focus state](xxxx) to make your design consistent with our new focus styles.
```
hannalaakso: Completely non-blocking: "our" sounds kind of proprietary. Maybe "GOV.UK Frontend focus styles"

CHANGELOG.md
```diff
@@ -2,510 +2,321 @@
...
+We've also changed the background of the following components:
 
-  ([PR #1316](https://github.com/alphagov/govuk-frontend/pull/1316))
+- buttons - `green` instead of a custom green
+- confirmation panels - `green` instead of `turquoise`
+- links in their hover state - `dark-blue` instead of `light-blue`
 
-- A new setting `$govuk-use-legacy-palette` has been added, which by default
-  will be true if any of the `$govuk-compatibility-*` settings are true.
+If you're using legacy projects like GOV.UK Elements, you can keep your current colours by [turning on compatibility mode](https://github.com/alphagov/govuk-frontend/docs/installation/compatibility.md).
```
hannalaakso: Would it be worth adding "while you migrate your service to use GOV.UK Frontend". We probably don't want to make it sound a viable option to continue using legacy codebases as they don't meet WCAG 2.1.

CHANGELOG.md
```diff
@@ -2,510 +2,321 @@
...
-changelog](./docs/contributing/versioning.md#updating-changelog).
+## Breaking changes
+You must make the following changes when you migrate to this release, or your service may break.
 
-- Renamed css class names `js-character-count` and `js-header-toggle`
+### Update file paths, attributes and class names
+To make sure your project's files do not conflict with GOV.UK Frontend, we've moved our package files into a directory called `govuk`.
 
-  Fixes issues with govuk-fronted javascript clashing with govuk-template/govuk-element and potentially other third-party javascript libraries using `js-` class names as hooks.
+#### If you’re using Sass
```
hannalaakso: Could we say "GOV.UK Frontend Sass".

CHANGELOG.md
```diff
@@ -2,510 +2,321 @@
...
+```
+If you’ve [added `node_modules/govuk-frontend` as a Sass include path](https://github.com/alphagov/govuk-frontend/blob/268b1e1b962d4bd293c5681868a2119f8b4ece60/docs/installation/installing-with-npm.md#optional-resolving-scss-import-paths), add `govuk/` to your `@import` paths:
 
-  ([PR #1288](https://github.com/alphagov/govuk-frontend/pull/1288))
+```scss
+@import "govuk/all";
+```
 
-- Spacing of tabs list updated to be more inline with similar lists on GOV.UK and the Design System
+####  If you’re using Javascript
```
hannalaakso: Could we say "GOV.UK Frontend Javascript"

CHANGELOG.md
```diff
@@ -2,510 +2,321 @@
...
 
-  This ensures the component's name does not conflict with other service level components
-  with the same name, which is consistent with our CSS class name convention.
+[Pull request #1444: Renames `js-` css prefix to `govuk-js-`](https://github.com/alphagov/govuk-frontend/pull/1444)
 
-  If you're using Nunjucks macros, and are [using the `initAll` function](https://github.com/alphagov/govuk-frontend/blob/master/docs/installation/installing-with-npm.md#option-1-include-javascript) you will not be affected.
+#### If you’re using Nunjucks
 
-  If you are using HTML, you will need to add the `govuk-` prefix to any `[data-module]` attributes.
+1. Change your `nunjucks.configure` property so it only includes the  `node_modules/govuk-frontend/` directory:
```
hannalaakso: I guess they might still want to import the entire `node_modules` for some other purpose 🤷‍♀ 

The meaning of "only" is a bit opaque here. The array could contain other values.

CHANGELOG.md
```diff
@@ -2,510 +2,321 @@
...
-  will be used.
+Read our [blog post about why we changed the colour palette](xxxx)
 
-  ([PR #1288](https://github.com/alphagov/govuk-frontend/pull/1288))
+[Pull request #1288: Update colour palette](https://github.com/alphagov/govuk-frontend/pull/1288).
 
-- The `govuk-colour` function has been updated to add a `$legacy` argument,
-  which allows you to specify a colour (either a literal, or a name of a colour
-  from the legacy palette) to use when `$govuk-use-legacy-palette` is true.
+### Check the new font
```
hannalaakso: Would it be worth saying that the font itself is unchanged, this is just a new version of it.

CHANGELOG.md
```diff
@@ -2,510 +2,321 @@
...
+```javascript
+govukButton({
+  text: "Start now",
+  href: "#",
+  isStartButton: true
+})
+```
 
-- Fix issue where link underlines sit too low in Firefox
+If you're using HTML, add the SVG code from the [start button example in the Design System](https://design-system.service.gov.uk/components/button/#start-buttons).
```
hannalaakso: Maybe "HTML `<svg>` element"

CHANGELOG.md
nickcolley: Since there's no legacy character count component we thought the distinction was not necessary

CHANGELOG.md
```diff
@@ -2,510 +2,321 @@
...
 ## Unreleased 3.0.0 (Breaking release)
 
-See the [versioning documentation for how to update this
-changelog](./docs/contributing/versioning.md#updating-changelog).
+## Breaking changes
+You must make the following changes when you migrate to this release, or your service may break.
 
-- Renamed css class names `js-character-count` and `js-header-toggle`
+### Update file paths, attributes and class names
+To make sure your project's files do not conflict with GOV.UK Frontend, we've moved our package files into a directory called `govuk`.
```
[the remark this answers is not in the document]
m-green: I think we link to #1458 further down the changenote for those who need it. It's probably best to avoid 'further info' links in the actual migration instructions, because it tempts users to click through looking for more actual instructions.

CHANGELOG.md
hannalaakso: Yeah I didn't see this as legacy vs non-legacy thing, it just looked like something might have been missed.

But I don't feel that strongly about it.

CHANGELOG.md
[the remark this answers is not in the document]
hannalaakso: Yeah I was thinking that might be the case. Cool 👍

CHANGELOG.md
nickcolley: Will leave this up to @m-green to decide

CHANGELOG.md
m-green: Thanks Hanna - so technically it's '... add a `gov-uk` prefix to any selectors that find components using the `data-module` attribute.'?

CHANGELOG.md
m-green: Ah good point - would '### Check the new version of the font' work better?

CHANGELOG.md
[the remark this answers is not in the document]
m-green: Thanks Hanna - as this is non-blocking I've added it to a list of potential post-release changes.

CHANGELOG.md
m-green: I feel it should be ok. If users should only replace this in certain files, we should definitely clarify. But if I understand it right, this is a simple search-and-replace for everywhere you have `js-character-count`?

CHANGELOG.md
m-green: I agree but I think we've covered this in the new compatibility page this links to, where it says "If you use these features, your service will not have GOV.UK Frontend's many accessibility and usability improvements."

CHANGELOG.md
hannalaakso: Yeah that makes sense 👍

CHANGELOG.md
hannalaakso: Ok no worries this is not blocking 👍

CHANGELOG.md
hannalaakso: Sounds good to me 👍

CHANGELOG.md
36degrees: I don't know whether this is important enough to warrant a change, but I think our motivation for this was more to avoid GOV.UK Frontend conflicting with the users code, than it is to defend our code against theirs.

Part of me thinks it's nit-picking, and part of me thinks it's actually quite important as it frames the change in a very different light ('we're doing this to make _our_ lives easier, but you have to do some work as a result' vs 'we've made this change because it helps you')…

I know we want to get this merged, so happy to leave it as-is, but might be worth spending 30 seconds thinking about.

CHANGELOG.md
36degrees: +1 - I think it's very likely that the user will be using nunjucks partials / layouts from elsewhere in their own application.

m-green: Thanks both - so would it be more like this: "In your `nunjucks.configure` property, change `/frontend/` to `/govuk-frontend/`"?

CHANGELOG.md
m-green: That seems fair to me - so reversing it to say "To make sure GOV.UK Frontend's files do not conflict with your project's files..." would work?

36degrees: I think so, yep 👍

CHANGELOG.md
```diff
@@ -2,510 +2,321 @@
...
 
-  Fixes issues with govuk-fronted javascript clashing with govuk-template/govuk-element and potentially other third-party javascript libraries using `js-` class names as hooks.
+#### If you’re using Sass
 
-  HTML markup will only need to change if nunjucks macros are not being used.
-  Update it use `govuk-js-character-count` or `govuk-js-header-toggle`.
+Add `govuk/` to `@import` paths in your [Sass](https://sass-lang.com/) file that have a `govuk-frontend/` path.
```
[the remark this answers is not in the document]
m-green: Yep, makes sense to be consistent - I'll add that.

CHANGELOG.md
m-green: I've added it to a list of potential post-release style changes.

CHANGELOG.md
m-green: I've added this to a list of potential post-release style changes.

CHANGELOG.md
m-green: I've added this to our list of potential post-release style questions.

CHANGELOG.md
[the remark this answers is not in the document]
m-green: Added in https://github.com/alphagov/govuk-frontend/pull/1492/commits/217e12a762789d8040c56dc7831cebc0c9282801

CHANGELOG.md
m-green: Added in https://github.com/alphagov/govuk-frontend/pull/1492/commits/217e12a762789d8040c56dc7831cebc0c9282801

CHANGELOG.md
m-green: Updated in https://github.com/alphagov/govuk-frontend/pull/1492/commits/217e12a762789d8040c56dc7831cebc0c9282801

CHANGELOG.md
m-green: Thanks - updated in https://github.com/alphagov/govuk-frontend/pull/1492/commits/217e12a762789d8040c56dc7831cebc0c9282801

CHANGELOG.md
```diff
@@ -77,7 +77,7 @@ If you're using HTML or custom JavaScript, change:
 
 #### If you’re using Nunjucks
 
-1. Change your `nunjucks.configure` property so it only includes the  `node_modules/govuk-frontend/` directory:
+1. In your `nunjucks.configure` property, change `/frontend/` to `/govuk-frontend/`:
```
nickcolley: This is no longer correct so I think we need to change this to something like:

When configuring Nunjucks views, remove any references to the `node_modules/govuk-frontend/components` directory.

nickcolley: People's configuration file will look like

```js
nunjucks.configure([
 'path/to/their/views',
 'node_modules/govuk-frontend', // Used for the template.njk
 'node_modules/govuk-frontend/components' // Used for components
])
```

And now need to be:

```js
nunjucks.configure([
 'path/to/their/views',
 'node_modules/govuk-frontend' // Used for the everything
])
```

PR #1493, opened 2019-07-12
Add automatic vertical spacing modifier for main wrapper

### LINKED ISSUE #1479: 'Automatic' main-wrapper margin may be applied incorrectly in some situations
For most services, the spacing for the main wrapper depends on whether there is an element in the 'beforeContent' block (such as breadcrumbs or a phase banner)

In #1371 we made a change to try and make it easier to get this spacing correct, by automatically applying a greater margin if the `main-wrapper` is the `:first-child` of its parent.

There are situations where this doesn't apply though which we may not have considered. For example, some admin or case-working systems may have additional navigation inside the main block, which they want to remain close to the header.

In the case of GOV.UK Pay, who tested a pre-release of 3.0 for us, they have an interface that looks like this:

Screenshot_2019-06-27 Dashboard - Jon Heslop’s test service sandbox test - GOV UK Pay

When they first upgraded, the new `:first-child` rule made it look like this:

Screenshot_2019-07-04 Dashboard - Purchase a positron projection permit sandbox test - GOV UK Pay

In order to defend against this, they had to [add their own `:first-child` rule to their stylesheet](https://github.com/alphagov/pay-selfservice/blob/19ec9d6ba59cc1123b24fef2c229c88800542039/app/assets/sass/components/navigation.scss#L1-L3):

```
.govuk-main-wrapper--with-phase-banner:first-child {
  padding-top: 5px;
}
```

We should consider reverting this 'automatic' behaviour or perhaps making it something that can be opted into or opted out of through the use of a modifier.

Alternatively, in this specific case the secondary navigation should not be inside the `main-wrapper` (and thus the `main-wrapper` would not be a `:first-child` and would not get the increased padding, which seems correct) – in which case maybe we just need to add better guidance around what should go inside the `main-wrapper`.

nickcolley:
We have found relying on `:first-child` alone for automatic spacing can be a a breaking change.

So we have changed this implementation slightly to be opt-in.

We intend to change the guidance (non-blocking v3) https://github.com/alphagov/govuk-design-system/issues/887, users can then remove this class easily if they do not need the custom spacing.

Closes #1479

Previous pull request: https://github.com/alphagov/govuk-frontend/pull/1371

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -298,13 +298,18 @@ You can now add attributes like classes, rowspan and colspan to table row header
 
 [Pull request #1367: Allow for classes, rowspan, colspan and attributes on row headers](https://github.com/alphagov/govuk-frontend/pull/1367). Thanks to [edwardhorsford](https://github.com/edwardhorsford).
 
-#### Use page wrappers without a modifier
 
-If your page wrapper does not have components above the main section, you no longer need to use the [`govuk-main-wrapper--l` modifier class](https://design-system.service.gov.uk/styles/layout/#exploded-view-of-page-wrappers-without-a-back-link-breadcrumbs-or-phase-banner) to increase the vertical space above the content.
+#### Use page wrapper auto spacing
+
+You can now use the the `.govuk-main-wrapper--auto-spacing` modifier on the main wrapper.
+
+This will apply the correct spacing depending on whether there are any elements (such the back link, breadcrumbs or phase banner components) before the `.govuk-main-wrapper` in the `.govuk-width-container`.
+
+If you need to control the spacing manually, use the `.govuk-main-wrapper--l` modifier instead.
 
 The `govuk-main-wrapper` and `govuk-main-wrapper--l` mixins are now deprecated. [Contact us](https://design-system.service.gov.uk/get-in-touch/) if you need to continue using these mixins.
 
-[Pull request #1371: Simplify main wrapper spacing logic](https://github.com/alphagov/govuk-frontend/pull/1371)
+[Pull request #1493: Add automatic vertical spacing modifier for main wrapper](https://github.com/alphagov/govuk-frontend/pull/1493)
 
 #### GDS Transport now falls back to Arial in Internet Explorer 8 (IE8)
 
--- a/src/govuk/objects/_main-wrapper.scss
+++ b/src/govuk/objects/_main-wrapper.scss
@@ -57,7 +57,7 @@
   // spacing without the need for the `l` modifier, but it is available in
   // instances where there may be empty elements that would be difficult to
   // remove.
-  .govuk-main-wrapper:first-child,
+  .govuk-main-wrapper--auto-spacing:first-child,
   .govuk-main-wrapper--l {
     @include govuk-main-wrapper--l;
   }
--- a/src/govuk/objects/_main-wrapper.scss
+++ b/src/govuk/objects/_main-wrapper.scss
@@ -53,10 +53,13 @@
     @include govuk-main-wrapper;
   }
 
-  // For most instances of main wrapper, :first-child will give the correct
-  // spacing without the need for the `l` modifier, but it is available in
-  // instances where there may be empty elements that would be difficult to
-  // remove.
+  // Using the `.govuk-main-wrapper--auto-spacing` modifier should apply the
+  // correct spacing depending on whether there are any elements
+  // (such the back link, breadcrumbs or phase banner components) before the
+  // `.govuk-main-wrapper` in the `govuk-width-container`.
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -408,14 +408,14 @@ changelog](./docs/contributing/versioning.md#updating-changelog).
 
   ([PR #1367](https://github.com/alphagov/govuk-frontend/pull/1367))
 
-- Simplify `.govuk-main-wrapper` logic to avoid the need for large modifier in most cases.
+- Add new main wrapper modifier to avoid the need for large modifier in most cases.
```
nickcolley: This change note will be removed so I've just done the bare minimum to make this history make sense.

nickcolley:
I talked to Mark and we think to make the 'opt-in' `--auto-spacing` modifier worthwhile we should update every example containing the main wrapper to use the modifier class, then the guidance will explain what to do if you need to remove it.

@oliver.byford my gut feeling is that this feels really clunky and I'm had me double guessing if we should have this additional class, maybe it is OK for people like Pay to have to override the unwanted spacing since they need to have custom CSS classes anyway.

src/govuk/objects/_main-wrapper.scss
```diff
@@ -57,7 +57,7 @@
   // spacing without the need for the `l` modifier, but it is available in
   // instances where there may be empty elements that would be difficult to
   // remove.
-  .govuk-main-wrapper:first-child,
+  .govuk-main-wrapper--auto-spacing:first-child,
```
hannalaakso: Non-blocking: We could expand on the comment to say that this bit of code ensures the correct margin when there are no other elements like breadcrumbs or back link above the main container

36degrees: I think the comment is slightly out of date as it still talks about the `:first-child` as if it's always applied.

What about:

```scss
  // Using the `govuk-main-wrapper--auto-spacing` modifier should apply the
  // correct spacing depending on whether there are any elements (such as a back
  // link or breadcrumbs) before the `.govuk-main-wrapper` in the 
  // `govuk-width-container`.
  //
  // If you need to control the spacing manually, use the
  // `govuk-main-wrapper--l` modifier instead.
```

?

CHANGELOG.md
```diff
@@ -298,13 +298,17 @@ You can now add attributes like classes, rowspan and colspan to table row header
 
 [Pull request #1367: Allow for classes, rowspan, colspan and attributes on row headers](https://github.com/alphagov/govuk-frontend/pull/1367). Thanks to [edwardhorsford](https://github.com/edwardhorsford).
 
-### Use page wrappers without a modifier
```
nickcolley: @m-green this updates how the page wrapper works

CHANGELOG.md
```diff
@@ -298,13 +298,17 @@ You can now add attributes like classes, rowspan and colspan to table row header
...
+### Use page wrapper auto spacing
 
-If your page wrapper does not have components above the main section, you no longer need to use the [`govuk-main-wrapper--l` modifier class](https://design-system.service.gov.uk/styles/layout/#exploded-view-of-page-wrappers-without-a-back-link-breadcrumbs-or-phase-banner) to increase the vertical space above the content.
+You can now use the the `.govuk-main-wrapper--auto-spacing` modifier on the main wrapper.
```
m-green: Tweaked this to match the language we use on https://design-system.service.gov.uk/styles/layout/.

```suggestion
You can now use the `.govuk-main-wrapper--auto-spacing` modifier class in your `<main>` element to add responsive padding to the top and bottom of the page.
```

m-green: Tweak to match the language in https://design-system.service.gov.uk/styles/layout/.

```suggestion
You can now add the `.govuk-main-wrapper--auto-spacing` modifier class to your `<main>` element to add responsive padding to the top and bottom of the page.
```

CHANGELOG.md
```diff
@@ -298,13 +298,17 @@ You can now add attributes like classes, rowspan and colspan to table row header
...
+
+This will apply the correct spacing depending on whether there are any elements (such the back link, breadcrumbs or phase banner components) before the `.govuk-main-wrapper` in the `.govuk-width-container`.
```
m-green: More tweaks to match the language. Also removed brackets, as we should try to avoid those wherever poss. Let me know if still reads ok.

```suggestion
This will add the correct amount of padding depending on whether there are elements above the `<main>` element inside the `govuk-width-container` wrapper, like the back link, breadcrumbs or phase banner components.
```

nickcolley: Good to know about brackets thanks Mark

PR #1496, opened 2019-07-16
Update tab focus state

### LINKED ISSUE #1472: Tabs focus state cropped in IE11
IE11 in Browserstack shows the focus correctly but in IE11 on native Windows 7 and 10 the focus appears cropped:

tabs-ie11

Edge on native Windows 10 displays the focus state correctly.

36degrees:
The ‘tab styling’ is currently applied to the link, with the list item doing very little in terms of presentation.

This is problematic when it comes to presenting the focus state, which we want to be tight around the tab text, and thus visually smaller than the tab. Our current approach, which creates the focus state using an `::after` pseudo-element, [doesn't render correctly in IE 11](https://github.com/alphagov/govuk-frontend/issues/1472).

By moving the tab styling to the list item, we can then make the link element smaller, and use an `::after` pseudo element to make the target area for the link fill the entire tab (the ['pseudo-content trick'](https://inclusive-components.design/cards/#thepseudocontenttrick)).

This also:

- generally tidies up the Sass a bit
- removes an unused `aria-current` ruleset in the Sass
- simplifies the focus state as we can just use the default link styles for IE8 as we do for accordions
- avoids setting the font on the entire tab panel content (it was being set on `govuk-tabs`)
- matches the tab colour to the background colour of the page rather than hardcoding to white
- tears down and restores the `aria-selected` attribute when moving to and from mobile

Fixes #1472. I suggest reviewing commit-by-commit.

## Screenshots

<details>
<summary>Chrome (macOS)</summary>
Chrome (macOS)
</details>

<details>
<summary>Safari (macOS)</summary>
Safari (macOS)
</details>

<details>
<summary>Firefox (macOS, custom colours)</summary>
Firefox (macOS, custom colours)
</details>

<details>
<summary>IE 8</summary>
IE 8
</details>

<details>
<summary>IE 9</summary>
IE 9
</details>

<details>
<summary>IE 10</summary>
IE 10
</details>

<details>
<summary>IE 11</summary>
IE 11
</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -207,6 +207,22 @@ If there are links back to radios or checkboxes components in your error summary
 
 [Pull request #1426: Make radios and checkboxes components easier to link to from error summary](https://github.com/alphagov/govuk-frontend/pull/1426)
 
+### Update the markup for tabs
+
+You do not need to do anything if you're using the Nunjucks macros.
+
+If you are not using the Nunjucks macros, remove the `govuk-tabs__tab--selected` class from the first tab's link, then add the `govuk-tabs__list-item--selected` class to the link's parent list item.
+
+```html
+<li class="govuk-tabs__list-item govuk-tabs__list-item--selected">
+  <a class="govuk-tabs__tab" href="#tab1">
+    Tab 1
+  </a>
+</li>
+```
+
+[Pull request #1496: Update the focus state for tabs](https://github.com/alphagov/govuk-frontend/pull/1443)
+
 ### Update start button icon
 
 [Start buttons](https://design-system.service.gov.uk/components/button/#start-buttons) have a new icon. Your start buttons will lose their current icons unless you replace the old icon with the new one.
--- a/src/govuk/components/tabs/_tabs.scss
+++ b/src/govuk/components/tabs/_tabs.scss
@@ -22,6 +22,7 @@
   }
 
   .govuk-tabs__list-item {
+    @include govuk-font($size: 19);
     margin-left: govuk-spacing(5);
 
     &::before {
@@ -33,7 +34,6 @@
 
   .govuk-tabs__tab {
     @include govuk-link-style-default;
-    @include govuk-font($size: 19);
 
     display: inline-block;
     margin-bottom: govuk-spacing(2);
--- a/src/govuk/components/tabs/_tabs.scss
+++ b/src/govuk/components/tabs/_tabs.scss
@@ -83,27 +83,6 @@
         text-align: center;
         text-decoration: underline;
 
-        &--selected {
-          $border-width: 1px;
-
-          position: relative;
-
-          margin-top: - govuk-spacing(1);
-
-          // Compensation for border (otherwise we get a shift)
-          margin-bottom: -$border-width;
-          padding-top: govuk-spacing(3) - $border-width;
-          padding-right: govuk-spacing(4) - $border-width;
-          padding-bottom: govuk-spacing(3) + $border-width;
-          padding-left: govuk-spacing(4) - $border-width;
-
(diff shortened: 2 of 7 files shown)


36degrees:
Suggested release note:

> ### Update the markup for tabs
> 
> You do not need to do anything if you're using Nunjucks macros.
>
> If you are not using the Nunjucks macros, remove the `govuk-tabs__tab--selected` from the first tab's link, and add the new class `govuk-tabs__list-item--selected` to its parent list item instead.
>
> ```html
> <li class="govuk-tabs__list-item govuk-tabs__list-item--selected">
>  <a class="govuk-tabs__tab" href="#tab1">
>   Tab 1
>  </a>
> </li>
> ```
>

36degrees:
Rebased and release note added.

CHANGELOG.md
```diff
@@ -207,6 +207,22 @@ If there are links back to radios or checkboxes components in your error summary
 
 [Pull request #1426: Make radios and checkboxes components easier to link to from error summary](https://github.com/alphagov/govuk-frontend/pull/1426)
 
+### Update the markup for tabs
```
nickcolley: This looks good to me. @m-green what do you think?

CHANGELOG.md
```diff
@@ -207,6 +207,22 @@ If there are links back to radios or checkboxes components in your error summary
...
+
+You do not need to do anything if you're not using using Nunjucks macros.
```
36degrees: Should we also say that the user doesn't need to do anything if they're not using tabs, or are we taking that as a given?

(We didn't seem to do it anywhere else)

nickcolley: People seem to search for things and check if they have them, and ignore things they dont use, that we saw...

CHANGELOG.md
m-green: Would this heading still be correct?

```suggestion
### Update the classes in tabs
```

m-green: Or 'Update tab classes'

CHANGELOG.md
m-green: Yep we should be ok not mentioning it. (Unless `govuk-tabs__tab--selected` appears in other non-tab components, but given the name I guess it wouldn't?)

m-green:
```suggestion
You do not need to do anything if you are not using Nunjucks macros.
```

CHANGELOG.md
```diff
@@ -207,6 +207,22 @@ If there are links back to radios or checkboxes components in your error summary
...
+
+If you are not using the Nunjucks macros, remove the `govuk-tabs__tab--selected` from the first tab's link, and add the new class `govuk-tabs__list-item--selected` to its parent list item instead:
```
m-green: Just a style tweak. I've assumed this should say if you *are* using Nunjucks macro, because line 212 already says 'if you are not'.

```suggestion
If you're using Nunjucks macros, remove the `govuk-tabs__tab--selected` class from the first tab's link, then add the `govuk-tabs__list-item--selected` class to the link's parent list item.

For example:
```

36degrees: D'oh! It's entirely the other way around, good spot.

36degrees: Line 212 also says 'using using'!

CHANGELOG.md
36degrees: Sort of… the class has also moved onto another element which makes it slightly more than just updating classes? 🤔

m-green: Ah yes, good point. How about just 'Update tabs'? Or stick with the original - this one's definitely not a dealbreaker.

36degrees: In which case I think I'll leave it as-is 👍

PR #1497, opened 2019-07-16
Users can now conditionally reveal content on pages with multiple grouped radios

nickcolley:
This pull request is a continuation of the work that @colinrotherham and @frankieroberto have proposed: https://github.com/alphagov/govuk-frontend/pull/1297

It is similar to Frankie's original suggestion but keeps the scope as small as possible to avoid running JavaScript unnecessarily.

Fixes https://github.com/alphagov/govuk-frontend/issues/1079
Closes #1297

### DIFF
--- a/src/govuk/components/radios/radios.js
+++ b/src/govuk/components/radios/radios.js
@@ -36,11 +36,13 @@ Radios.prototype.init = function () {
 }
 
 Radios.prototype.setAttributes = function ($input) {
-  var inputIsChecked = $input.checked
-  $input.setAttribute('aria-expanded', inputIsChecked)
-
   var $content = document.querySelector('#' + $input.getAttribute('aria-controls'))
-  if ($content) {
+
+  if ($content && $content.classList.contains('govuk-radios__conditional')) {
+    var inputIsChecked = $input.checked
+
+    $input.setAttribute('aria-expanded', inputIsChecked)
+
     $content.classList.toggle('govuk-radios__conditional--hidden', !inputIsChecked)
   }
 }
--- a/src/govuk/components/radios/radios.js
+++ b/src/govuk/components/radios/radios.js
@@ -52,14 +52,11 @@ Radios.prototype.handleClick = function (event) {
   }
   // Because checking one radio can uncheck a radio in another $module,
   // we need to call set attributes on all radios in the document
-  var $allInputs = document.querySelectorAll('input[type="radio"]')
+  //
+  // We also only want radios which have aria-controls, as they support conditional reveals.
+  var $allInputs = document.querySelectorAll('input[type="radio"][aria-controls]')
   nodeListForEach($allInputs, function ($input) {
-    // If a radio with aria-controls, handle click
-    var isRadio = $input.getAttribute('type') === 'radio'
-    var hasAriaControls = $input.getAttribute('aria-controls')
-    if (isRadio && hasAriaControls) {
-      this.setAttributes($input)
-    }
+    this.setAttributes($input)
   }.bind(this))
 }
 
--- a/src/govuk/components/radios/radios.js
+++ b/src/govuk/components/radios/radios.js
@@ -5,12 +5,11 @@ import { nodeListForEach } from '../../common'
 
 function Radios ($module) {
   this.$module = $module
-  this.$inputs = $module.querySelectorAll('input[type="radio"]')
 }
 
 Radios.prototype.init = function () {
   var $module = this.$module
-  var $inputs = this.$inputs
+  var $inputs = $module.querySelectorAll('input[type="radio"]')
 
   /**
   * Loop over all items with [data-controls]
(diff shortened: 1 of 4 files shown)


src/govuk/components/radios/radios.js
```diff
@@ -55,8 +55,12 @@ Radios.prototype.handleClick = function (event) {
   //
   // We also only want radios which have aria-controls, as they support conditional reveals.
   var $allInputs = document.querySelectorAll('input[type="radio"][aria-controls]')
+  var $clickedInput = event.target
   nodeListForEach($allInputs, function ($input) {
-    this.setAttributes($input)
+    // In radios, only radios with the same name will affect each other.
```
nickcolley: Can we have a test for this? :thinking:

src/govuk/components/radios/radios.js
```diff
@@ -40,18 +39,27 @@ Radios.prototype.setAttributes = function ($input) {
   var inputIsChecked = $input.checked
   $input.setAttribute('aria-expanded', inputIsChecked)
 
-  var $content = this.$module.querySelector('#' + $input.getAttribute('aria-controls'))
+  var $content = document.querySelector('#' + $input.getAttribute('aria-controls'))
```
nickcolley: Just realised I've confused myself.

This line is the one that's important that it isnt global, not the others.

https://github.com/alphagov/govuk-frontend/pull/1156

Going to have to figure this out...

nickcolley: I think, we can't assume that 'aria-controls' on a radio means that the user has only conditional reveals.

In this case that we fixed, they put their own aria-controls on the radio for something unrelated.

I think we'll need to do something at the initialisation stage to avoid this...

36degrees: We either need to have a way of querying the `$module` in which the given `$input` exists, or query the whole `document`, as `this.$module` will only search within the scope of the module of the _clicked_ radio.

In terms of checking for aria-controls on something unrelated, yesterday you suggested checking for `govuk-radios__conditional` in `$content.classList`, which sounds like a sane defence?

If we do that, then scoping this to the document seems OK to me as we'll always just be syncing the `--hidden` class with the checked state of the input?

src/govuk/components/radios/radios.js
```diff
@@ -40,18 +39,27 @@ Radios.prototype.setAttributes = function ($input) {
...
+  // We only want to handle clicks for radio inputs
+  if ($clickedInput.type !== 'radio') {
+    return
+  }
+  // Because checking one radio can uncheck a radio in another $module,
+  // we need to call set attributes on all radios in the same form, or document if they're not in a form.
+  //
+  // We also only want radios which have aria-controls, as they support conditional reveals.
+  var $scope = ($clickedInput.form || document)
+  var $allInputs = $scope.querySelectorAll('input[type="radio"][aria-controls]')
```
36degrees: This is probably something of an edge-case but this won't query any inputs that are outside of the form but associated with it using `form="form-id"` as they won't be children of the form.

I'd suggest instead checking the form attribute of both within the forEach, i.e. `if ($input.form === $clicked.form)`

src/govuk/components/radios/radios.js
```diff
@@ -40,18 +39,27 @@ Radios.prototype.setAttributes = function ($input) {
...
+  nodeListForEach($allInputs, function ($input) {
+    // In radios, only radios with the same name will affect each other.
+    if ($clickedInput.name === $input.name) {
```
36degrees: This check seems backwards to me, I think?

```suggestion
    if ($input.name === $clickedInput.name) {
```

src/govuk/components/radios/radios.js
36degrees: Because all this will do is sync the conditional reveals with the checked state of their inputs, we'd have to do something really weird, like having another set of radios with a conditional reveal that's 'out of sync' with its radios, and asserting that it doesn't sync them back up again.

It's such a synthetic scenario I'm not sure it makes much sense as a test.

src/govuk/components/radios/radios.js
nickcolley: Yeah I forgot about the idea of looking for that class, that'll be the ticket thanks Ollie.

nickcolley:
Blocking this on the release notes being merged, so I can add a CHANGELOG entry...

nickcolley:
I wasn't up to date with my branch on my work computer and force pushed over some of my changes, so I'll sort this out when I get home...

nickcolley:
@36degrees suggested that I use the GitHub info log in this pull request to target the commit before I forced pushed which worked fine...

src/govuk/components/radios/radios.js
```diff
@@ -37,21 +36,35 @@ Radios.prototype.init = function () {
 }
 
 Radios.prototype.setAttributes = function ($input) {
-  var inputIsChecked = $input.checked
-  $input.setAttribute('aria-expanded', inputIsChecked)
+  var $content = document.querySelector('#' + $input.getAttribute('aria-controls'))
```
hannalaakso: Non-blocking: It could be worth leaving a comment here to say that we use `document` as the scope to cater for the edge case when other radios with conditionals are part of the same group. It might not be immediately clear that there could be other radios that are in fact related even though they're outside `$module`.

nickcolley: I originally put a comment in but @36degrees thought that if we were doing this from scratch that we probably wouldnt have commented it. The case you're talking about is described in the test suite, so maybe that's enough? What do you think?

nickcolley: I think if someone changed this the tests would fail so I'm going to go with what Ollie and I discussed originally so can get this merged 🐎

edwardhorsford:
@nickcolley #1079 is about being able to group related radios together - is that made possible from this PR?

nickcolley:
@edwardhorsford this allows you to do the conditional reveals when you have groups of radios with the same name, so I think so! :)

This'll go out very soon in 3.0.0.

If you need a different use case feel free to open an new issue for us to consider.

PR #1498, opened 2019-07-17
Minor fixes to header component

36degrees:
- Fix a bug where hovering the GOV.UK logo whilst it is focused results in the header shrinking by 1px
- Reduce the line-height on the product name, which was causing the height of the header to increase by 2px (from 60px to 62px) when a product name was added.

### DIFF
--- a/src/govuk/components/header/_header.scss
+++ b/src/govuk/components/header/_header.scss
@@ -61,7 +61,7 @@
   }
 
   .govuk-header__product-name {
-    @include govuk-font($size: 24);
+    @include govuk-font($size: 24, $line-height: 1);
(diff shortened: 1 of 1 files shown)


nickcolley:
We need a changelog entry for this...

36degrees:
I believe the first fix is fixing something that was introduced with the new focus state, which hasn't shipped yet.

Not sure about the second fix – I believe as part of 3.0 we've generally reduced the size of the header to align it closer to what's currently on GOV.UK, but that's not mentioned in the changelog already?

nickcolley:
alright sounds fair :)

PR #1499, opened 2019-07-17
Update description of serviceName macro option

frankieroberto:
Fixes what looks like a copy-and-paste mistake with the macro option descriptions for the Header component.

### DIFF
--- a/package/components/header/macro-options.json
+++ b/package/components/header/macro-options.json
@@ -21,7 +21,7 @@
         "name": "serviceName",
         "type": "string",
         "required": false,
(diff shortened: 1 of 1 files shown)


nickcolley:
Heya Frankie,

Thanks for this pull request, I think this'll get resolved once version 3.0.0 of frontend goes out based on this pull request: https://github.com/alphagov/govuk-frontend/pull/1368

Files within the `package` directory are generated at release, which is why this has not been updated yet.

So with that in mind I'm going to close this out.

PR #1501, opened 2019-07-18
Add example of fixed-width header w/ product name

36degrees:
The only examples we currently have for headers with product names are all full width.

We know that in many cases a product name will be used without the header being full width, so we should support this.

At the minute, this example does not render as we’d want it to – the ‘Product Name’ wraps onto a new line – so we’ll need to look at fixing this in the future.

localhost_3000_components_header_with-product-name_preview


36degrees:
I've recorded the issue in #1502

Issue #1494, opened 2019-07-12
Test for empty items in radios and checkboxes


edwardhorsford:
I'd like to propose testing if items inside the `items` array for checkboxes and radios are empty. If they are, don't output markup.

## My use case
I want to conditionally show a radio based on some data.

In Nunjucks I can alter a value of an item in an array by adding a conditional after it - but it still outputs an empty item and I've not worked out how to not provide it.

I could prepare my data in advance - but for a simple thing like this, doing it directly is more readable IMHO. You can see the logic directly.

Example macro:
```js
{{ govukRadios({
  idPrefix: "where-do-you-live",
  name: "where-do-you-live",
  fieldset: {
    legend: {
      text: "Where do you live?",
      isPageHeading: true,
      classes: "govuk-fieldset__legend--xl"
    }
  },
  items: [
    {
      value: "england",
      text: "England"
    },
    {
      value: "scotland",
      text: "Scotland"
    },
    {
      value: "wales",
      text: "Wales"
    },
    {
      value: "northern-ireland",
      text: "Northern Ireland"
    },
    {
      divider: "or"
    } if data.showFantasy == true,
    {
      value: "narnia",
      text: "Narnia"
    } if data.showFantasy == true
  ]
}) }}
```

This _mostly_ works - but right now outputs a blank space and random html doing nothing:
<img width="631" alt="Screenshot 2019-07-12 at 15 23 36" src="https://user-images.githubusercontent.com/2204224/61135180-0c133080-a4b9-11e9-9d2a-3542121e440a.png">
<img width="486" alt="Screenshot 2019-07-12 at 15 24 12" src="https://user-images.githubusercontent.com/2204224/61135217-20572d80-a4b9-11e9-8faf-4f046799591e.png">

## Suggested fix
Either strip out empty items from the `items` array or else test whether they're empty when iterating over them.

This feels like a safer thing to do / more defensive. We don't output markup where there's no input.

I had a quick test of this by editing the radios macro directly in my node_modules, and it appears to work well.

nickcolley:
I've put together a live demonstration to show what Ed is talking about here: https://glitch.com/edit/#!/govuk-frontend-issue-1494

nickcolley:
Should we also consider doing something similar for:

- https://design-system.service.gov.uk/components/accordion/ (items)
- https://design-system.service.gov.uk/components/breadcrumbs/ (items)
- https://design-system.service.gov.uk/components/error-summary/ (error list)
[...]

PR #1512, opened 2019-07-24
Update components to only output items when they are defined

htmlandbacon:
Just a follow up to https://github.com/alphagov/govuk-frontend/issues/1494

I've had a similar issue on the summary list where peopled ended up adding classes to hide content.

This currently covers:

- Radios items
- Checkbox items
- summary items
- accordion items
- tab items
- Select options
- table rows

Not 100% where else it would be needed to be added, but this at least makes the higher level ones consistent .

Also I wasn't sure how best to describe this in the docs - "An array of sections within the accordion, undefined values are skipped." doesn't seem quite right but opening as WIP for eyes.

### DIFF
--- a/src/govuk/components/accordion/template.njk
+++ b/src/govuk/components/accordion/template.njk
@@ -4,22 +4,24 @@
 <div class="govuk-accordion {%- if params.classes %} {{ params.classes }}{% endif -%}" data-module="govuk-accordion" id="{{ id }}"
 {%- for attribute, value in params.attributes %} {{attribute}}="{{value}}"{% endfor %}>
   {% for item in params.items %}
-    <div class="govuk-accordion__section {% if item.expanded %}govuk-accordion__section--expanded{% endif %}">
-      <div class="govuk-accordion__section-header">
-        <h{{ headingLevel }} class="govuk-accordion__section-heading">
-          <span class="govuk-accordion__section-button" id="{{ id }}-heading-{{ loop.index }}">
-            {{ item.heading.html | safe if item.heading.html else item.heading.text }}
-          </span>
-        </h{{ headingLevel }}>
-        {% if item.summary.html or item.summary.text %}
-          <div class="govuk-accordion__section-summary govuk-body" id="{{ id }}-summary-{{ loop.index }}">
-            {{ item.summary.html | safe if item.summary.html else item.summary.text }}
-          </div>
-        {% endif %}
+    {% if item %}
+      <div class="govuk-accordion__section {% if item.expanded %}govuk-accordion__section--expanded{% endif %}">
+        <div class="govuk-accordion__section-header">
+          <h{{ headingLevel }} class="govuk-accordion__section-heading">
+            <span class="govuk-accordion__section-button" id="{{ id }}-heading-{{ loop.index }}">
+              {{ item.heading.html | safe if item.heading.html else item.heading.text }}
+            </span>
+          </h{{ headingLevel }}>
+          {% if item.summary.html or item.summary.text %}
+            <div class="govuk-accordion__section-summary govuk-body" id="{{ id }}-summary-{{ loop.index }}">
+              {{ item.summary.html | safe if item.summary.html else item.summary.text }}
+            </div>
+          {% endif %}
+        </div>
+        <div id="{{ id }}-content-{{ loop.index }}" class="govuk-accordion__section-content" aria-labelledby="{{ id }}-heading-{{ loop.index }}">
+          {{ item.content.html | safe if item.content.html else item.content.text }}
+        </div>
       </div>
-      <div id="{{ id }}-content-{{ loop.index }}" class="govuk-accordion__section-content" aria-labelledby="{{ id }}-heading-{{ loop.index }}">
-        {{ item.content.html | safe if item.content.html else item.content.text }}
-      </div>
-    </div>
+    {% endif %}
   {% endfor %}
 </div>
--- a/src/govuk/components/select/template.test.js
+++ b/src/govuk/components/select/template.test.js
@@ -63,6 +63,24 @@ describe('Select', () => {
       expect($items.length).toEqual(2)
     })
 
+    it('renders without falsely items', () => {
+      const $ = render('select', {
+        items: [
+          {
+            text: 'Option 1'
+          },
+          undefined,
+          null,
+          false,
+          {
+            text: 'Options 2'
+          }
+        ]
+      })
+      const $items = $('.govuk-select option')
+      expect($items.length).toEqual(2)
+    })
+
     it('renders item with value', () => {
       const $ = render('select', {
         value: '2',
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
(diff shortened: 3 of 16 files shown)


nickcolley:
Hey Colin,

Thanks for helping us out here, we'll be able to review your work properly once we've got 3.0.0 shipped (soon).

I'll leave some initial review comments right now though, hope that's OK.

src/govuk/components/accordion/accordion.yaml
```diff
@@ -18,7 +18,7 @@ params:
 - name: items
   type: array
   required: true
-  description: An array of sections within the accordion.
+  description: An array of sections within the accordion, undefined values are skipped.
```
nickcolley: >  I wasn't sure how best to describe this in the docs

We may not need to update these descriptions, if we think that this is an expected behaviour from our users, which I'd argue it is.

src/govuk/components/accordion/template.njk
```diff
@@ -4,22 +4,24 @@
...
-            {{ item.heading.html | safe if item.heading.html else item.heading.text }}
-          </span>
-        </h{{ headingLevel }}>
-        {% if item.summary.html or item.summary.text %}
-          <div class="govuk-accordion__section-summary govuk-body" id="{{ id }}-summary-{{ loop.index }}">
-            {{ item.summary.html | safe if item.summary.html else item.summary.text }}
-          </div>
-        {% endif %}
+    {% if item %}
+      <div class="govuk-accordion__section {% if item.expanded %}govuk-accordion__section--expanded{% endif %}">
```
nickcolley: By indenting the code it makes it easier to read but introduces more whitespace to the output HTML.

Should we consider keeping the if statement at the same level as the for loop?

This frustrates me very much I wish we could have best of both worlds.

src/govuk/components/select/template.test.js
```diff
@@ -63,6 +63,24 @@ describe('Select', () => {
...
+          undefined,
+          null,
+          false,
+          {
+            text: 'Options 2'
+          }
+        ]
+      })
+      const $items = $('.govuk-select option')
+      expect($items.length).toEqual(2)
```
nickcolley: Nice clear tests for all these thanks Colin

nickcolley:
The reason the build has failed is because there are JavaScript linting errors, you can see these by running `npm test`. Let me know if you need any help getting that working.

src/govuk/components/accordion/accordion.yaml
htmlandbacon: Yeah, that is why I opened it as WIP - it seems like it needs to be documented, but wasn't sure the best way.

This is certain a way!

src/govuk/components/accordion/template.njk
htmlandbacon: Yeah, I've got a similar issue with zoomtext where there is a value vs [space]value[space] issue to look at, I'll re-look at them.

edwardhorsford:
@colinrotherham I don't know if it could be done, but I've also got a use case for hiding a table column.

src/govuk/components/accordion/template.njk
edwardhorsford: Presumably we could also filter the source items for empty - and then skip the if statement entirely.

htmlandbacon: Is there already a filter for this? I'd be interested in seeing some render performance tests!

colinrotherham:
@edwardhorsford This is my partner in crime, the arguably better Colin, @htmlandbacon 😁

htmlandbacon:
> @edwardhorsford This is my partner in crime, the arguably better Colin, @htmlandbacon 😁

All dem Colins.

So, you could do this but it starts getting messier and messier, specially as there are things like this where it is detecting the first row and multi colspans.

```
{% for cell in row %}
          {% set commonAttributes %}
            {%- if cell.colspan %} colspan="{{ cell.colspan }}"{% endif %}
            {%- if cell.rowspan %} rowspan="{{ cell.rowspan }}"{% endif %}{% for attribute, value in cell.attributes %} {{ attribute }}="{{ value }}"{% endfor %}
          {% endset %}
          {% if loop.first and params.firstCellIsHeader %}
          <th scope="row" class="govuk-table__header{%- if cell.classes %} {{ cell.classes }}{% endif %}"
            {{- commonAttributes | safe -}}
          >{{ cell.html | safe if cell.html else cell.text }}</th>
          {% else %}
          <td class="govuk-table__cell
            {%- if cell.format %} govuk-table__cell--{{ cell.format }}{% endif %}
            {%- if cell.classes %} {{ cell.classes }}{% endif %}"
            {{- commonAttributes | safe -}}
          >{{ cell.html | safe if cell.html else cell.text }}</td>
          {% endif %}
        {% endfor %}
```

Wonder if it is worth opening up a separate issue with some use cases you have?

src/govuk/components/accordion/template.njk
nickcolley: I don't think there is a filter for this unfortunately, but if there were this would be the way to go.

htmlandbacon: I've done some stuff with the mighty dash to clean up this example

base is current
new is from this pr
refactored is with - to clean up the input

Files, results and some janky performance 
https://gist.github.com/htmlandbacon/26c9119d6f0609b7e420f6a9a072e01e

nickcolley: What do you think about this:

To get this merged let's do the simplest approach that doesn't impact whitespace and then consider refactoring separately as most of these templates could use some love regardless of Ed's original issue.

edwardhorsford:
Many apologies @colinrotherham and @htmlandbacon. You're both excellent and held in equally high colin-esteem.

@htmlandbacon yeah, disabling a column feels like an unrelated thing. Please ignore.

src/govuk/components/accordion/template.njk
htmlandbacon: I guess the easiest thing would be to leave, as is, with bonus white pace but if its going to get blocked till I fix them I can do!

src/govuk/components/accordion/accordion.yaml
htmlandbacon: Just reflecting on this, with some arrays now working like this, I feel like it should be explicit?

Happy to take the review comment and change, else I can this can be resolved?

src/govuk/components/accordion/template.njk
nickcolley: I don't think it's a hard blocker since the Design System website *should* reformat the HTML nicely, but I'd prefer if we tried to keep the output HTML from getting more funky even if it's a bit rough around the edges right now.

Will see what the rest of the team thinks later :)

nickcolley:
@htmlandbacon I've chatted about this pull request with the team.

We've chatted about the document updates and think it's better to keep it simpler without talking about undefined values, so we can remove the documentation updates from this pull request.

Once that's done we're happy to accept this and get it in :)

htmlandbacon:
> @htmlandbacon I've chatted about this pull request with the team.
> 
> We've chatted about the document updates and think it's better to keep it simpler without talking about undefined values, so we can remove the documentation updates from this pull request.
> 
> Once that's done we're happy to accept this and get it in :)

Probably worth a check over, but I've removed these, cheers!

edwardhorsford:
Thanks so much @htmlandbacon! 👏

htmlandbacon:
Thanks for sorting the change log @nickcolley!

PR #1513, opened 2019-07-24
Fix style in the v3 (unreleased) changelog

m-green:
This pull request contains fixes some minor style issues and inconsistencies in the v3 (unreleased) changelog, including from 2i.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,9 +33,9 @@ You must do the following.
 
 ###### Update file paths
 
-You must add `govuk/` to your import file paths.
+You must add `govuk/` to your import paths.
 
-If you're importing `node_modules/govuk-frontend/all.js` in your code, change this file path to `node_modules/govuk-frontend/govuk/all.js`.
+If you're importing `node_modules/govuk-frontend/all.js`, change this import path to `node_modules/govuk-frontend/govuk/all.js`.
 
 If you’re importing a specific path, add `govuk/` after `govuk-frontend/`. For example, if you're importing the button component:
 
@@ -47,7 +47,7 @@ import Button from 'govuk-frontend/govuk/components/button/button'
 
 You do not need to do anything if you're using Nunjucks macros and the [`initAll`](https://github.com/alphagov/govuk-frontend/blob/master/docs/installation/installing-with-npm.md#option-1-include-javascript) function.
 
-If you are not using the Nunjucks macros, add a `govuk-` prefix to `data-module` attribute values. For example:
+If you are not using Nunjucks macros, add a `govuk-` prefix to `data-module` attribute values. For example:
 
 ```html
 <div class="govuk-accordion" data-module="govuk-accordion">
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -77,7 +77,7 @@ If you're using HTML or custom JavaScript, change:
 
 ##### If you’re using Nunjucks
 
-1. Change your `nunjucks.configure` property so that the only GOV.UK Frontend directory it contains is `node_modules/govuk-frontend/`:
+1. Change your `nunjucks.configure` object so that the only GOV.UK Frontend directory it contains is `node_modules/govuk-frontend/`:
```
nickcolley: I believe nunjucks.configure is a 'function' to be specific.

To be accurate this should probably be `Change your `nunjucks.configure` options`

CHANGELOG.md
```diff
@@ -117,9 +117,9 @@ Pull requests:
 
 The focus state of components now meets the new WCAG 2.1 level AA requirements.
 
-You must update your component’s focus state to make your design consistent with our new focus styles.
+You must update your component’s focus state to make your design is consistent with our new focus styles.
```
nickcolley: I think this 'is' might be a mistake.

CHANGELOG.md
m-green: Ah great, thanks. I think 'function' works best, if that works for you?

CHANGELOG.md
m-green: Oops, yes, sorry - I introduced this error. Will fix.

CHANGELOG.md
36degrees: You're not really changing the function, you're just changing what you pass to it.

How about something like:

```suggestion
1. Change the list of paths passed to `nunjucks.configure` so that the only GOV.UK Frontend path is `node_modules/govuk-frontend/`
```

?

For reference, this is how Nunjucks documentation talks about the configure function:
https://mozilla.github.io/nunjucks/api.html#configure

m-green: Sounds good, but could we tweak to 'Change the list of paths in...'?

36degrees: For sure, happy for you to revise further as you see fit! :+1:

m-green: Ace, thanks - this is amended in https://github.com/alphagov/govuk-frontend/pull/1513/commits/46b32a88c1b7da1af2ba697a405dec8a242431ca.

CHANGELOG.md
m-green: Fixed in https://github.com/alphagov/govuk-frontend/pull/1513/commits/46b32a88c1b7da1af2ba697a405dec8a242431ca.

m-green:
https://github.com/alphagov/govuk-frontend/pull/1513/commits/46b32a88c1b7da1af2ba697a405dec8a242431ca adds one more 2i tweak that I forgot about yesterday - sorry!

PR #1516, opened 2019-07-31
Add more hex colour before and after to CHANGELOG

nickcolley:
Someone on support was looking for a before and after for the grey hex codes.

Since people may have hardcoded colours this makes it easier to migrate.

I'll update the release note if this is accepted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -171,8 +171,12 @@ purple|#2e358b|#4c2c92
 red|#b10e1e|#d4351c
 yellow|#ffbf47|#ffdd00
 green|#006435|#00703c
+grey-2|#bfc1c3|#b1b4b6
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -171,8 +171,13 @@ purple|#2e358b|#4c2c92
 red|#b10e1e|#d4351c
 yellow|#ffbf47|#ffdd00
 green|#006435|#00703c
+grey-1|#6f777b|#6f777b
```
m-green: Looks good 👍 
Because we've framed this as 'change these colours', I'd probably remove `grey-1`. Otherwise users might think we've made a mistake and meant to include a different #rgb value.

nickcolley: ok 👍

PR #1517, opened 2019-07-31
Add del package as devDependency

LBHELewis:
Included this in devDependencies as it's used by gulp tasks

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -615,6 +615,32 @@
         "@types/yargs": "^12.0.9"
       }
     },
+    "@nodelib/fs.scandir": {
+      "version": "2.1.1",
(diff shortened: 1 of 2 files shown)


nickcolley:
Thanks for sorting this!

I've pulled this down locally and ran `npm install` and it's looking good.

This is used by https://github.com/alphagov/govuk-frontend/blob/master/tasks/gulp/clean.js#L5

I'll need a second review from someone else on the team.

36degrees:
@nickcolley any idea how this hasn't tripped us up in the past? Is `del` an implicit dependency already?

nickcolley:
When I spoke to @LBHELewis this came up when removing the `package-lock.json` and running `npm install`.

It's a bit of a weird one, maybe one of our dependencies has `del`, I'm honestly not sure.

Issue #1253, opened 2019-03-25
Disabled radios and checkboxes do not look disabled in IE9 thru IE11


36degrees:
Disabled radios and checkboxes do not look disabled in IE9, IE10 and IE11 – their labels are 'greyed out', but the controls themselves are black.

![Screenshot showing disabled radios and checkboxes with controls retaining the black borders associated with non-disabled state](https://user-images.githubusercontent.com/121939/54930056-7e620400-4f0e-11e9-838b-be6ae2f02132.png)

They should look like this (screenshot from Edge 18):

![Screenshot showing the expected 'greyed out' controls](https://user-images.githubusercontent.com/121939/54930032-74400580-4f0e-11e9-9f6b-2f3370c501b3.png)

I believe this is because `:before` and `:after` pseudo-elements do not inherit the parent elements opacity in IE.

nickcolley:
Had a look into this, seems to be the same issue as described here: https://www.jacklmoore.com/notes/ie-opacity-inheritance/

So we can solve this by having relative positioning on the label, but then we have to add more complexity to the absolute positioned pseudo elements.

I've also tried changing the label font colour, which works as expected but I imagine would not work as well as opacity for users who change their colours...

36degrees:
Based on the discussion in #1522, we think that the correct way to fix this involves a breaking change, as we would need to change the way that checkboxes and radios are positioned, which may break things for anyone using them in unexpected ways, or where users have created their own modifiers.

If this was only an issue in e.g. IE 8 and 9 it might be appropriate to close as a 'won't fix', but given this affects IE11 we think it's still worth doing, and the approach used in #1522 does seem like the best fix.

We're assigning this to the 4.0 milestone to pick up again then.

vanitabarrett:
As discussed in the V4 scoping session, it's possible there's a way of doing this without a breaking change that we should investigate. Rather than changing the positioning of the elements (which is a breaking change), we might be able to set the colour. We [previously had concerns](https://github.com/alphagov/govuk-frontend/pull/1522#issuecomment-526249860) about how this approach would work for controls on background colours other than white (e.g: if a user changed their colour settings), but we might be able to use [queries for forced-colors mode to help](https://github.com/alphagov/govuk-frontend/pull/2264/files#diff-975a8be111c331e3a0404d9a11e68b81a526c58ce9ab3177f5e94b898d359250R135)

querkmachine:
[...]

PR #1522, opened 2019-08-05
 Fix disabled radios and checkboxes not being greyed out in IE9-11

nickcolley:
This feels like a pretty big change to the positioning of these elements so would welcome some thoughts of how this could break things for users...

See https://github.com/alphagov/govuk-frontend/issues/1253#issuecomment-518200978 for more details

Fixes https://github.com/alphagov/govuk-frontend/issues/1253

### DIFF
--- a/src/govuk/components/checkboxes/_checkboxes.scss
+++ b/src/govuk/components/checkboxes/_checkboxes.scss
@@ -69,6 +69,8 @@
 
   .govuk-checkboxes__label {
     display: inline-block;
+    // In order for Internet Explorer 9 - 11 to inherit opacity on pseudo elements the parent needs to be relative to it's direct children.
+    position: relative;
     margin-bottom: 0;
     padding: 8px $govuk-checkboxes-label-padding-left-right govuk-spacing(1);
     cursor: pointer;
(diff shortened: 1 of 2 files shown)


nickcolley:
@36degrees what's your feeling on this?

36degrees:
Seems to work really well. I guess the scenarios where this change might cause problems might be things like users who are using checkboxes inside tables, or have created their own modifiers?

Maybe we could try and find some of those users and do a pre-release to see whether it does affect them?

nickcolley:
I think given that this is a uncommon pattern (we do not recommend people disable inputs) I'm wary of making this change due to the risk of breaking people's modifiers like you've suggested.

Given this I would push for closing this as a will not fix as the labels themselves do indicate a change in state.

However if we feel strongly that this could cause barriers to users then let's do the testing.

36degrees:
I spent some time talking this through with @timpaul – we discussed other alternative approaches, including using grey for either the controls or the text, but came to the conclusion that this would cause issues if the component is used on backgrounds other than white, which would be more of an issue as and when we create inverse versions of the controls.

If this was only an issue in e.g. IE 8 and 9 it might be easier to close as a 'won't fix', but given this affects IE11 I think it's still worth doing, and this does seem like the right approach.

I'd suggest we do this but treat it as a breaking change, so… close this for now, document this in the issue, add the breaking change label, and assign it to a 4.0 milestone? Thoughts?

nickcolley:
I agree with your conclusions :+1:

PR #1526, opened 2019-08-08
Add isStartButton (which is used in the examples) to the params in bu…

LBHELewis:
…tton.yaml

isStartButton is used in button.yaml in the data for the examples (and then used in the template file) but it is never defined in the params. This PR adds it to the params.

### DIFF
--- a/src/govuk/components/button/button.yaml
+++ b/src/govuk/components/button/button.yaml
@@ -46,7 +46,7 @@ params:
 - name: isStartButton
   type: boolean
   required: false
-  description: If set to true this button will have an arrow icon added to it. Use for the main call to action on your service's start page.
+  description: Use for the main call to action on your service's start page.
 
 examples:
 - name: default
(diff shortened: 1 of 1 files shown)


src/govuk/components/button/button.yaml
```diff
@@ -43,6 +43,10 @@ params:
   type: boolean
   required: false
   description: Prevent accidental double clicks on submit buttons from submitting forms multiple times
+- name: isStartButton
+  type: boolean
+  required: false
+  description: If set to true this button will have an arrow icon added to it. Use for the main call to action on your service's start page.
```
nickcolley: Generally with a few exceptions we try to avoid explaining how to set options and what the internal implementation results from that.

I think we could make this simpler and just keep the latter bit of the sentence, something like:
```suggestion
  description: Use for the main call to action on your service's start page.
```

What do you think?

LBHELewis: @nickcolley yes - that's better 👍

nickcolley:
Ooo super fast 🚤 💨 

Will need a second reviewer from the team to approve this and we'll get it merged.

We'll only see this improvement when we do a release so I've attached this to the ['Next' milestone](https://github.com/alphagov/govuk-frontend/milestone/21).

LBHELewis:
Great, thank you!

PR #1531, opened 2019-08-12
Add templates for issues and pull requests

36degrees:
Between 13-26 August, the GOV.UK Design System team is assisting with urgent and critical Brexit-related work being undertaken by GDS and will not be able to respond to support requests.

We're adding issue and pull request templates to let users know what's going on, and to set expectations around how quickly we'll be able to respond.

### DIFF
--- a/.github/issue_template.md
+++ b/.github/issue_template.md
@@ -1,0 +1,9 @@
+<!--
+Just to let you know, the GOV.UK Design System team are assisting with urgent Brexit-related work being undertaken by GDS between 13 and 26 August.
+
+During this time, we will not be able to review your issue. Sorry about that.  
+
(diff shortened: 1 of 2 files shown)


36degrees:
Yep, that's intentional so that we don't show the 'type of issue' picker interface on the new issue page (at least as far as I understand it from reading the docs)

nickcolley:
Will work for what we need it for but just for your information:

> This is the legacy workflow to create an issue template. We recommend using the upgraded multiple issue template builder to create multiple issue templates. For more information, see "About issue and pull request templates."

https://help.github.com/en/articles/manually-creating-a-single-issue-template-for-your-repository

PR #1541, opened 2019-08-28
Allow date input fieldset attributes to be set

### LINKED ISSUE #1478: govukDateInput not allowing "attribute" property to change for fieldset option
Hi,
I've been working using the govukDateInput component and was looking to disable the input boxes; I am using the `<fieldset>` variant.  When passing in the property "attribute", I found that it was being ignored.  e.g. 

      {{ govukDateInput({
        value: 'the value',
        fieldset: {
          attributes: { 
            disabled: true 
          },
          legend: {
            text: t('common:dob'),
            isPageHeading: false,
            classes: 'govuk-fieldset__legend--s'
          }
        },
        hint: {
          text: t('common:dobHint')
        },
      })}}

I've looked at your codeset and the issue is with the attributes value you've had to add.  This causes any attribute property passed in to be ignored:
```
  {% call govukFieldset({
    describedBy: describedBy,
    classes: params.fieldset.classes,
    attributes: {
      role: "group"
    },
    legend: params.fieldset.legend
  }) %}
  {{ innerHtml | trim | safe }}
  {% endcall %}
```

I think the solution is quite simple and should have no impact.  A very rough example solution could be: 

```
  {% call govukFieldset({
    describedBy: describedBy,
    classes: params.fieldset.classes,
    attributes: { ...params.fieldset.attributes, role: 'group' },
    legend: params.fieldset.legend
  }) %}
  {{ innerHtml | trim | safe }}
  {% endcall %}
```
I've decided against disabling these now, but this still seems worth fixing.
Thanks

36degrees:
- Promote the `role` attribute to a ‘first-class’ option of the fieldset component, as we currently pass it in within ‘attributes’ from the date input component.
- Update the date input component to pass `role` directly rather than as part of `attributes`, meaning that `params.fieldset.attributes` can be set, which is consistent with other components.

Fixes #1478

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,21 @@
 
 ## Unreleased
 
-###  Inputmode
+### New features
+
+#### Add attributes to the date input fieldset
+
+You can now pass attributes to add to the fieldset on the date input component.
+
+[Pull request #1541: Allow date input fieldset attributes to be set](https://github.com/alphagov/govuk-frontend/pull/1541). Thanks to [andrew-mcgregor](https://github.com/andrew-mcgregor) for raising this.
+
+#### Add ARIA role to the fieldset component
+
+You can now pass an ARIA role to the fieldset component.
+
+[Pull request #1541: Allow date input fieldset attributes to be set](https://github.com/alphagov/govuk-frontend/pull/1541).
+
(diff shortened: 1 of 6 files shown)


36degrees:
Needs a changelog entry.

Issue #1539, opened 2019-08-28
Footer layout isn't aligned with content layout


dashouse:
As noticed on GovWifi, the flexbox auto layout used in the footer means that the columns do not lock into the grid.

**Example from GovWifi**
<img width="1044" alt="Screen Shot 2019-08-28 at 11 53 40" src="https://user-images.githubusercontent.com/23356842/63849764-a278b400-c98a-11e9-9d1b-fdc871c3ebc2.png">

With equal content, the content is not distributed evenly, the first column takes up more space. Adding content to the second or third column will take away space from the first, which is why the GovWifi example looks **almost** right, but it's coincidental.

**Recreated in the Prototype Kit**
<img width="1055" alt="Screen Shot 2019-08-28 at 11 52 15" src="https://user-images.githubusercontent.com/23356842/63849793-b1f7fd00-c98a-11e9-91f3-1ed3e452de62.png">

Is there anything we can add to the flexbox properties to ensure min or max width of content?

## Done when

- [ ] TBC
- [ ] "Details of breaking change" section below completed to help with writing release notes

## Details of breaking change

- which users are affected: TBC
- the change that’s been made: TBC
- changes users will have to make: TBC
- impact of users not making those changes: TBC

nickcolley:
I've had a look into this, I think the way flexbox works means that even though the proportions are correct in certain scenarios the footer will not align with the main grid system.

I have proposed a fix to improve the specific issues raised by Dave here for three columns: https://github.com/alphagov/govuk-frontend/pull/1542

I believe to fix this we need to improve how the footer works:
https://github.com/alphagov/govuk-frontend/tree/spike-only-default-two-thirds-footer-for-two-sections

This will require a small but breaking change, so will put this against the 4.0.0 milestone

36degrees:
We may want to look at #1284 at the same time as this.

vanitabarrett:
As discussed in v4 prioritisation, we think this work is important but it's quite broad at the moment. We should probably do a timeboxed investigation into this issue and [this related one](https://github.com/alphagov/govuk-frontend/issues/1726) to explore a few different layout systems for the footer and see how they perform.

PR #1542, opened 2019-08-28
Only default two thirds/one-third layout when footer has two sections

nickcolley:
This approach is from: http://lea.verou.me/2011/01/styling-children-based-on-their-number-with-css3/

Changes the default layout of two-thirds/one-third only when there's two sections.

The first section can only be also the the 2nd last section (counted backwards) if there's two overall sections.

### Before
Before equal columns

### After
After equal columns

Improves https://github.com/alphagov/govuk-frontend/issues/1539

### DIFF
--- a/src/govuk/components/footer/_footer.scss
+++ b/src/govuk/components/footer/_footer.scss
@@ -196,9 +196,11 @@
     }
   }
 
-  // Sections two-third:one-third on desktop
+  // If there are only two sections, set the layout to be two-third:one-third on desktop
   @include mq ($from: desktop) {
-    .govuk-footer__section:first-child {
+    // We match the first section with `:first-child`.
+    // To ensure the section is one of two, we can count backwards using `:nth-last-child(2)`.
+    .govuk-footer__section:first-child:nth-last-child(2) {
       flex-grow: 2; // Support: Flexbox
     }
   }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ govukInput({
 - [Pull request #1523: Improve accessibility of details component by polyfilling only where the native element is not available](https://github.com/alphagov/govuk-frontend/pull/1523).
 - [Pull request #1512: Update components to only output items when they are defined](https://github.com/alphagov/govuk-frontend/pull/1512).
 - [Pull request #1538: Simplify button types to avoid unnecessary type attribute](https://github.com/alphagov/govuk-frontend/pull/1538).
+- [Pull request #1542: Only default two thirds/one-third layout when footer has two sections](https://github.com/alphagov/govuk-frontend/pull/1542).
 
 ## 3.0.0 (Breaking release)
 
--- a/src/govuk/components/footer/footer.yaml
+++ b/src/govuk/components/footer/footer.yaml
@@ -251,3 +251,53 @@ examples:
         - href: '/cymraeg'
           text: Rhestr o Wasanaethau Cymraeg
       html: Built by the <a class="govuk-footer__link" href="#">Government Digital Service</a>
+
+- name: Three equal columns
+  description: A full example to demonstrate three equal width columns
+  data:
+    navigation:
+      - title: Single column list 1
+        columns: 1
+        items:
+          - href: '#1'
+            text: Navigation item 1
+          - href: '#2'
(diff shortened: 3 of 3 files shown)


src/govuk/components/footer/_footer.scss
```diff
@@ -196,9 +196,9 @@
     }
   }
 
-  // Sections two-third:one-third on desktop
+  // If the there are only two sections and set the layout to be two-third:one-third on desktop
```
ChrisBAshton:
```suggestion
  // If there are only two sections, set the layout to be two-third:one-third on desktop
```

src/govuk/components/footer/_footer.scss
```diff
@@ -196,9 +196,9 @@
...
   @include mq ($from: desktop) {
-    .govuk-footer__section:first-child {
+    .govuk-footer__section:first-child:nth-last-child(2) {
```
ChrisBAshton: Won't this apply a `flex-grow: 2` to the 2nd-to-last child of the footer section _regardless_ of how many columns there are?

I.e. a three-column layout will set the middle column to `flex-grow: 2` - is that expected?

nickcolley: nth-last-child counts backwards, so it'll only target the first element and the 2nd from last element.

if you had three columns then the 2nd element would not also be the first element

ChrisBAshton: Ahhhhhh that makes sense. Clever! 👏

src/govuk/components/footer/_footer.scss
nickcolley: This reads loads better thanks Chris

src/govuk/components/footer/_footer.scss
nickcolley: I'm not a fan of clever code but feels appropriate here

nickcolley: I'll try and write a comment that explains this better...

ChrisBAshton: Thanks, this is much better!

36degrees:
I think there's still a couple of places where this isn't doing what we want – I think we're hardcoding some logic to try and meet a couple of specific examples (the GOV.UK Footer, three columns and one column) but there's a couple of other scenarios that are breaking the grid too.

For example, if you put the 2-column section on the right, it does not fit with the grid (you end up with a ⅖ column and a ⅗ column:

localhost_3000_components_footer_GOV UK-Flipped_preview (1)

If you have two columns, it should either go to two ½ columns, or two ⅓ columns and a third blank ⅓ column, but at the minute we get two uneven columns:

localhost_3000_components_footer_Two-equal-columns_preview

If we have only one column it kind of works as-is as it fills the entire footer, though I question whether there might be cases where a user might want to restrict it to a ⅓ column:

localhost_3000_components_footer_One-column_preview

36degrees:
I've currently these as examples on top of this branch – would it be useful for me to commit them and push them as something to work from?

nickcolley:
@36degrees yes please, we may be able to make use of the column modifier classes that already exist to tighten this up even more.

nickcolley:
Actually I've changed my mind I think those examples look how I'd expect, given the context of a fall page where most layouts are two thirds / one thirds it feels balanced. Having an empty slot feels worse?

36degrees:
Neither of the examples align with the grid – the first example is more like a ⅖ column and a ⅗ column, and the second example is two uneven columns.

nickcolley:
I think I know what you're saying, I wonder if it's always been a bit off...

nickcolley:
I've pushed up a change that uses CSS Grid instead of Flexbox.

This would mean IE11 would use the fallback layout (simple inline block that does not align with the grid).

Would also be a breaking change since I've moved the modifier...

What do you think?

- https://caniuse.com/#feat=flexbox
- https://caniuse.com/#feat=css-grid

nickcolley:
I think we can make a small breaking change to sort this properly, so have pushed the main issues around the grid alignment back into the main issue for the next milestone: https://github.com/alphagov/govuk-frontend/issues/1539#issuecomment-527123152

I will reduce this pull request back to the original intent, so we can ship a small improvement to three column layouts only.

PR #1543, opened 2019-08-29
Minor improvements to the changelog

nickcolley:


36degrees:
I'm not 100% sure about this change. I was aiming to be consistent with [the new features as listed in the 3.0 release](https://github.com/alphagov/govuk-frontend/blob/master/CHANGELOG.md#new-features-1) which seem to be framed from the point of view of 'things *you*, the user, can now do' rather than 'things we have done' / 'changes we have made' – i.e. 'you can now add attributes to the date input fieldset' vs 'We have add[ed the] attributes option to the date input fieldset'.

I know you were more involved with the release notes for 3.0, so if you're sure this is the intended format then I'm happy to go with it, but otherwise it'd be good to try and get clarity from @m-green.

nickcolley:
Yeah I find the user centred action stuff odd with features since it's more of a 'you can if you want' rather than a 'you should definitely do this'

m-green:
Yes it's a bit of a difficult fit for new feature headings as Nick says.

The "X now Y" format works in some cases (eg "GDS Transport now falls back to Arial in Internet Explorer 8 (IE8)" example from v3).

Otherwise I tend to use the "You can now..." format, but strip out the actual "You can now" bit. So "~You can now~ Add attributes to table headings"

nickcolley:
I'll update with this in mind thanks for the feedback both.

PR #1549, opened 2019-09-03
Change case of 'Tag' component

dashouse:
This PR:
- Creates modifiers for `.govuk-tag--uppercase` and `.govuk-tag--lowercase` (which also removes the letter-spacing)
- Adds `.govuk-tag--uppercase` modifier to Phase banner component

I think we need to remove the `text-decoration` / `letter-spacing` declaration from the component all together, however to avoid waiting for a breaking change I have put in deprecation warnings into the component SCSS and used the modifier in the phase banner even though it's not currently needed.

Screen Shot 2019-09-03 at 14 27 07

Screen Shot 2019-09-03 at 14 27 12

The lowercase modifier, because it will be used to override lowercase on something that is already uppercase uses `:first-letter { text-transform: capitalize; }`. This wouldn't be needed if we did the breaking change straight away as we wouldn't need a lowercase modifier.

Let me know what you think is the best approach for this is?

### DIFF
--- a/src/govuk/components/tag/_tag.scss
+++ b/src/govuk/components/tag/_tag.scss
@@ -38,6 +38,20 @@
     }
   }
 
+  .govuk-tag--uppercase {
+    letter-spacing: 1px;
+    text-transform: uppercase;
+  }
+
+  .govuk-tag--lowercase {
+    letter-spacing: normal;
+    text-transform: lowercase;
+
+    &:first-letter {
+      text-transform: capitalize;
+    }
+  }
+
   .govuk-tag--inactive {
     background-color: govuk-colour("dark-grey", $legacy: "grey-1");
   }
--- a/src/govuk/components/tag/_tag.scss
+++ b/src/govuk/components/tag/_tag.scss
@@ -15,9 +15,13 @@
 
     color: govuk-colour("white");
     background-color: govuk-colour("blue");
+    // Deprecation warning: This will be removed in v4. 
+    // Use the .govuk-tag--uppercase modifier to make tag uppercase / spaced out if required.
     letter-spacing: 1px;
 
     text-decoration: none;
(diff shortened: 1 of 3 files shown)


src/govuk/components/tag/_tag.scss
```diff
@@ -38,6 +42,21 @@
     }
   }
 
+  .govuk-tag--uppercase {
+    letter-spacing: 1px;
+    text-transform: uppercase;
+  }
+
+  .govuk-tag--lowercase {
```
nickcolley: will this modifier be removed in v4.0.0?

dashouse: I guess so, this is why I'm debating whether or not this just needs to be a breaking change anyway?

dashouse: Unless of course we want this class for anyone that has their existing data in uppercase?

36degrees:
It's maybe a little messy, but the other option might be to add the uppercase transformations to the `govuk-phase-banner__content__tag` instead – this class already exists, so there would be no markup changes required when users update to retain the uppercase-ness.

Then we would remove them from the tag component, adding a modifier for those that want to continue using uppercase. However:

- this wouldn't help for cases where the tag is used for alpha / beta tags outside of the phase banner (e.g. after the product name in the header)
- I'm unclear whether making other tags lowercase (excluding the phase banner) would still constitute a breaking change.

(I'm assuming that we want to keep the uppercase transformation for alpha / beta tags – have/should we explored title-casing them too?)

nickcolley:
I would vote for something similar to Dave's proposal:

- Breaking change
- Default is lowercase and no spacing
- Uppercase modifier for phase banner usage
- Phase banner Nunjucks is updated

Since using the phase banner modifier would work, but as you say there's other places where you'd want the uppercase modifier.

PR #1553, opened 2019-09-04
Include formGroup on character count and pass through to textarea to …

LBHELewis:
…allow class to be added to character count form group

Hopefully this should make sense. Found myself in a situation where I wanted to add a class to the formGroup but couldn't when using character count so included it in the params that get passed through to the textarea (where it was already present).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+### New features
+
+#### Add formGroup parameter to character count component
+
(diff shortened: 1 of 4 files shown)


LBHELewis:
@36degrees Great, thank you. I've updated the changelog now. Put it as a New Feature, but I think you could also argue that it's a Fix, so let me know if you'd like me to change it.

36degrees:
No, I think that makes sense – thanks! 🙂

Issue #1550, opened 2019-09-03
IE8 causes 404 errors when using Frontend 3.0


petergphillips:
We recently upgraded our app to Frontend 3.0.

As a result we have noticed that we are getting lots of 404 errors in our server logs from IE8 users requesting invalid URLs e.g. 
https://gateway.prod.nomis-api.service.hmpps.dsd.io:443/auth/fonts/bold-b542beb274-v2.woff2)%20format(%22woff2%22),%20url(/auth/fonts/bold-affa96571d-v2.woff)%20format(%22woff%22

It looks like IE8 doesn't understand the `@font-face` directive and treats `format("woff2"),url(` as part of the URL too.

As a side issue woff2 isn't supported in IE8 so shouldn't be included in the IE8 css (https://github.com/alphagov/govuk-frontend/blob/master/dist/govuk-frontend-ie8-3.1.0.min.css).

36degrees:
Thanks for reporting, @petergphillips 

I've been able to reproduce this in the review app as well – seeing requests for `/assets/fonts/light-94a07e06a1-v2.woff2) format("woff2"), url(/assets/fonts/light-f591b13f7d-v2.woff) format("woff"` there.

hannalaakso:
I did some investigation. The font paths themselves could be naively fixed by [adding `?#iefix` to the end of the URLs](https://stackoverflow.com/questions/8050640/how-does-iefix-solve-web-fonts-loading-in-ie6-ie8/) which is what we used to do in [v2](https://github.com/alphagov/govuk-frontend/blame/88f28d949daedcf11d2cac2f2ff323d7c23a51c9/src/helpers/_font-faces.scss#L25).  However this would just mean that the woff2 files get downloaded by IE8 which is redundant as IE8 doesn't support the format.

We could put a `govuk-if-ie8` check in [font-faces](https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/helpers/_font-faces.scss#L18) which would stop them being outputted for IE8. This would mean that the font files aren't downloaded by IE8 and we fall back to Arial. This could be an okay fix? However it would mean that CSS for IE8 would still say:
` font-family: "GDS Transport", Arial, sans-serif;`
which is somewhat confusing but then this is what v3 already does anyway. 

[...]

PR #1559, opened 2019-09-06
Stop IE8 from downloading GDS Transport font

hannalaakso:
Internet Explorer 8 was making [unsuccessful requests](https://github.com/alphagov/govuk-frontend/issues/1550) for the new font files because 3.0.0 no longer [postfixes them with `?#iefix`](https://stackoverflow.com/questions/8050640/how-does-iefix-solve-web-fonts-loading-in-ie6-ie8/) as we did in [v2](https://github.com/alphagov/govuk-frontend/blame/88f28d949daedcf11d2cac2f2ff323d7c23a51c9/src/helpers/_font-faces.scss#L25).

3.0.0 [dropped support for rendering GDS Transport for IE8](https://github.com/alphagov/govuk-frontend/pull/1434). This PR removes the `font-face` declarations from the IE8 stylesheet to stop IE8 trying to unsuccessfully to download the font files. This means that IE8 will fall back to system fonts (`Arial` by default). Our previous approach of postfixing the paths with `?#iefix` wouldn't work here as IE8 would still download the redundant WOFF files.

The downside of this new approach is that `GDS Transport` is still shown in `font-family` in the IE8 stylesheet.

I've also tried an [alternative approach](https://github.com/alphagov/govuk-frontend/compare/fix-ie8-fonts-alt-approach?expand=1) that sets `font-family` for IE8 on a higher level and removes the above issue, but has other downsides.

This also imports the IE8 tools for `govuk-typography-common` tests, and tests that `font-face` is not declared for IE8.

Fixes https://github.com/alphagov/govuk-frontend/issues/1550

### DIFF
--- a/src/govuk/helpers/typography.test.js
+++ b/src/govuk/helpers/typography.test.js
@@ -131,6 +131,28 @@ describe('@mixin govuk-typography-common', () => {
     expect(resultsString).toContain(`font-family: "GDS Transport"`)
     expect(resultsString).toContain(`font-family: "GDS Transport"`)
   })
+  it('should not output a @font-face declaration when the browser is IE8', async () => {
+    const sass = `
+    $govuk-is-ie8: true;
+
+    @import "settings/all";
+    @import "helpers/all";
+    @import "tools/ie8";
+
+    :root {
+      @include govuk-typography-common;
+    }
+    :root {
+      @include govuk-typography-common($font-family: $govuk-font-family-tabular);
+    }
+    `
+
+    const results = await renderSass({ data: sass, ...sassConfig })
+    const resultsString = results.css.toString()
+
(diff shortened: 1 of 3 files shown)


src/govuk/helpers/typography.test.js
```diff
@@ -130,6 +131,28 @@ describe('@mixin govuk-typography-common', () => {
     expect(resultsString).toContain(`font-family: "GDS Transport"`)
     expect(resultsString).toContain(`font-family: "GDS Transport"`)
   })
+  it('should not output a @font-face declaration when the browser is IE8', async () => {
```
nickcolley: Nice test :+1:

36degrees: 💯

Issue #1562, opened 2019-09-09
Some components active states are broken in compatibility mode


nickcolley:
<img width="307" alt="Screen Shot 2019-09-09 at 17 36 36" src="https://user-images.githubusercontent.com/2445413/64549968-c98b9a00-d329-11e9-8d69-054c05500560.png">

We have run into this issue before here: https://github.com/alphagov/govuk-frontend/pull/1310

This seems to be a problem for other components too, you can see the impact live on GOV.UK: https://www.gov.uk/register-to-vote

You can reproduce this by going to:
http://localhost:3000/components/button/start-link/preview?legacy=1

And changing the href to something unique in the example yaml file.

This may be an issue with other components too.

nickcolley:
Not sure if there's a more holistic approach we can have here.

Should all hrefs get automatically be cache busted?

If we did that how do we test visited states?

PR #1564, opened 2019-09-10
Ensure legacy styles do not override button active state

nickcolley:
**This bug only impacts users that are using the legacy project GOV.UK Template alongside GOV.UK Frontend.**

Add additional selector only when GOV.UK Template is loaded to ensure it does not override GOV.UK Frontend's active state.

This was missed because the links are all visited which gives the selectors a higher specificity.

To reproduce this you will need to change the link href in the example yaml file so that is not a visited link.

## All link styles tested

- Button start link (fixed)
- Tabs unenhanced mode (good)
- Back link (good)
- Breadcrumbs (good)
- Error summary links (good)
- Footer links (good)
- Header links (good)
- Skip-link (good)

Fixes https://github.com/alphagov/govuk-frontend/issues/1562

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,12 +6,13 @@
 
 #### Add formGroup parameter to character count component
 
-- [Pull request #1553: Include formGroup on character count and pass through to textarea to allow class to be added to character count form group](https://github.com/alphagov/govuk-frontend/pull/1553)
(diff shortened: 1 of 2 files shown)


nickcolley:
@36degrees yeah, I'm not sure how we fixed it there but not in the default variation, is a bit of a weird one.

PR #1570, opened 2019-09-13
Prevent inputs ending up off screen or obscured by keyboards when linking from the error summary to inputs within a large fieldset

### LINKED ISSUE #1544: Error Summary link handling can result in obscured inputs
As reported by a user on Slack:

> I’ve got a bit of a weird one that only seems to affect iOS users. We have an address page that follows the standard pattern. When you click continue it will show the error messages which link to the appropriate inputs. If you click on the error message for the post code it seems to scroll down a bit but the input is covered my the iOS keyboard (please see attached screenshot).
>
> 
> image

The user confirmed that the address fields were inside a fieldset (as [we recommend in the addresses pattern](https://design-system.service.gov.uk/patterns/addresses/#multiple-text-inputs)).

I believe this is happening because the logic used when clicking on the error summary links is to scroll to the legend of the parent fieldset, if there is one. This makes sense for radios, checkboxes and the date input, but maybe less so in this situation.

36degrees:
Only scroll from error summary to legend if the input is a radio or checkbox, or the input would end up in the top half of the screen.

This is to prevent linking to inputs which then end up either off the bottom of the screen or obscured by a software keyboard.

Fixes #1544

### DIFF
--- a/src/govuk/components/error-summary/error-summary.js
+++ b/src/govuk/components/error-summary/error-summary.js
@@ -112,10 +112,16 @@ ErrorSummary.prototype.getAssociatedLegendOrLabel = function ($input) {
 
     if (legends.length) {
       var $candidateLegend = legends[0]
-      
-      // Only scroll to the fieldset’s legend (instead of the label associated
-      // with the input) if the input would end up in the top half of the
-      // screen.
+
+      // If the input type is radio or checkbox, always use the legend if there
+      // is one.
+      if ($input.type === 'checkbox' || $input.type === 'radio') {
+        return $candidateLegend
+      }
+
+      // For other input types, only scroll to the fieldset’s legend (instead of
+      // the label associated with the input) if the input would end up in the
+      // top half of the screen.
       //
       // This should avoid situations where the input either ends up off the
       // screen, or obscured by a software keyboard.
--- a/src/govuk/components/error-summary/error-summary.js
+++ b/src/govuk/components/error-summary/error-summary.js
@@ -94,7 +94,9 @@ ErrorSummary.prototype.getFragmentFromUrl = function (url) {
  *
  * Returns the first element that exists from this list:
  *
- * - The `<legend>` associated with the closest `<fieldset>` ancestor
+ * - The `<legend>` associated with the closest `<fieldset>` ancestor, as long
+ *   as the top of it is no more than half a viewport height away from the
+ *   bottom of the input
  * - The first `<label>` that is associated with the input using for="inputId"
  * - The closest parent `<label>`
  *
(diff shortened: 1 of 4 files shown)


src/govuk/components/error-summary/error-summary.js
```diff
@@ -109,7 +111,32 @@ ErrorSummary.prototype.getAssociatedLegendOrLabel = function ($input) {
     var legends = $fieldset.getElementsByTagName('legend')
 
     if (legends.length) {
-      return legends[0]
+      var $candidateLegend = legends[0]
```
nickcolley: In theory we could write tests for this logic, where we do using puppeteer on the demo pages you've been testing, this could be useful in the future if we (hopefully not) encounter other edge cases.

nickcolley:
For reference there's also 'IntersectionObservers' which have concepts of thresholds, I bit overkill for this I think since we're only doing it once per link click but worth sharing: https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API

src/govuk/components/error-summary/error-summary.js
```diff
@@ -109,7 +111,32 @@ ErrorSummary.prototype.getAssociatedLegendOrLabel = function ($input) {
...
+      // screen, or obscured by a software keyboard.
+      var legendTop = $candidateLegend.getBoundingClientRect().top
+      var inputRect = $input.getBoundingClientRect()
+
+      // If the browser doesn't support Element.getBoundingClientRect().height
+      // or window.innerHeight (like IE8), bail and just link to the label.
+      if (inputRect.height && window.innerHeight) {
+        var inputBottom = inputRect.top + inputRect.height
+
+        if (inputBottom - legendTop < window.innerHeight / 2) {
```
nickcolley: When testing this on my machine I find that I'm always linked to the errored input for the last example despite the fieldset being able to fit all on the screen at once.

Is the issue that the keyboard opens after the navigation?

src/govuk/components/error-summary/error-summary.js
36degrees: I've added to the existing tests to try and cover the two new examples that have been added to the review app.

nickcolley: Epic! Thank you

src/govuk/components/error-summary/error-summary.js
36degrees: More or less – the logic checks to see whether the input would end up entirely in the top half of the viewport, to allow for keyboards.

36degrees:
@nickcolley happy with this changelog entry?

nickcolley:
@36degrees 👍 looks peng mate

Issue #1519, opened 2019-08-01
Form elements do not scale well if only font size increased


andysellick:
If you `zoom text only` in Firefox, select elements retain their 40px height but the text within them becomes larger and therefore cropped from the bottom edge.

<img width="678" alt="Screen Shot 2019-08-01 at 14 30 10" src="https://user-images.githubusercontent.com/861310/62297312-e4d0d300-b468-11e9-9e98-390152b7ae23.png">

Doesn't seem to be as much of a problem for text inputs.

Can't reproduce the problem in Chrome because the zoom text only option isn't available.

Noticed this during user testing on a phone of GOV.UK's search, specifically where we have select elements [in finders](https://www.gov.uk/search/news-and-communications).

(can't remember what kind but possibly an iPhone 7?)

Side note: doing this on the Design System site causes the currently active menu item to be crossed out instead of underlined:

<img width="1199" alt="Screen Shot 2019-08-01 at 14 31 28" src="https://user-images.githubusercontent.com/861310/62297443-2497ba80-b469-11e9-84e0-b0dfb5e73727.png">

andysellick:
This appears to be fixable by changing the height from using `px` to `em` but I think we already discussed this and there was a reason not to do it? Can someone remind me?

nickcolley:
There seems to be some concern to nested font sizing: https://github.com/alphagov/govuk-frontend/commit/5f4d20c2ef9549c0e4596239f77067e0462ff89a

But perhaps `rem` would be fine?

andysellick:
Interestingly on testing it seems `rem` is far less predictable than `em`. `rem` in the component guide produces one height but the same thing tested inside `finder-frontend`'s search results in a much smaller height (actually creating the original issue in a different way). `em` produces a consistent height in both interfaces.

andysellick:
Looks like the reason for that is the body font size isn't set in the component guide (?) - we might need to look into that.

nickcolley:
The main issue with 'em' and 'rem' when using it with the legacy GOV.UK Template is that GOV.UK Template messes with the root font size, which is likely what you're seeing.

nickcolley:
We do have ways to only use relative units when compatibility mode is not turned on that we could consider.

36degrees:
It looks like this _can_ affect text inputs, but only at extreme values (I had to modify the values for `toolkit.zoomManager.zoomValues` and `zoom.maxPercent` in Firefox's `about:config` to do this):

[...]

Issue #882, opened 2018-07-10
Investigate Dynamic Type to improve accessibility for iOS users


hannalaakso:
This came out of the work on https://github.com/alphagov/govuk-frontend/pull/858.

We currently don't support adjusting the base font size on iOS devices due to iOS constraints.

However @36degrees spotted that [Dynamic Type ](https://ind.ie/labs/blog/dynamic-type-support/)could solve this issue.

We should investigate this as it would be great to support iOS users who needs to adjust their base font size.

## Who needs to know about this

Developers, Tim or Joe

nickcolley:
I believe it would require us to not use the GOV.UK font?

36degrees:
From what I can tell, `font: -apple-system-body;` is doing more than just setting the font-family, and the dynamic font-sizing persists even if we override the `font-family` later - in other words, we can still use the font.

From a quick spike it seems to work, with the exception that Laura's approach of 'scaling' the root font-size by setting it on the body won't work for us because everything is using rem. Frustratingly, this means that the 'base' font-size ends up being 17px. Setting the font-size on the html element stops the dynamic sizing from working, so that's not an option either.

I _think_ we'd have to modify the `typography-responsive` helper to spit out another font-size declaration for iOS:

```scss
@supports (font: -apple-system-body) {
  font-size: $font-size / 17px * 1rem;
}
```

This seems to work, but will increase the size of our CSS and increase the complexity of that helper.

nickcolley:
Saw this which is related, which seems to suggest setting the system font on the root element then changing it allows for the resizing to work: https://twitter.com/eriksen_dk/status/1075555147856732161?s=17

nickcolley:
OK so need to do an investigation to see the impact on file size by adding that helper min/gzip, and if it's too large consider if it's a problem if the base size is 17px...

nickcolley:
Looks like 11.4 kB with Ollie's suggestion and 10.4 kB without, so a decent chunk.

36degrees:
Going to try and get a working example set up so we can see how it works, and what difference it makes to asset size.

36degrees:
Master branch:
```
$ wc -c dist/govuk-frontend-3.5.0.min.css
   88829 dist/govuk-frontend-3.5.0.min.css

$ gzip-size dist/govuk-frontend-3.5.0.min.css
10.4 kB
```

This branch:
```
wc -c dist/govuk-frontend-3.5.0.min.css
[...]

PR #1574, opened 2019-09-13
Make form elements scale correctly when text resized by user

hannalaakso:
## Requirement

WCAG 2.1 requires that user can resize text to at least 200%.

## Fix
The current versions of `<select>` and `<input>` do not scale well when text is resized by user as they're constrained by `height` set in pixels. Changing the height to use `rem` relative units makes them scalable.

Thanks @nickcolley and @andysellick for the suggestion to use relative units.

### Before
chrome-before
chrome-before-2

### After
Screen Shot 2019-09-13 at 14 48 10
Screen Shot 2019-09-13 at 14 48 20

### Before
ff-before
firefox-before-2

### After
Screen Shot 2019-09-13 at 15 20 04
Screen Shot 2019-09-13 at 15 23 20

### Before
safari-before
safari-before-2

### After
Screen Shot 2019-09-13 at 15 20 47
Screen Shot 2019-09-13 at 15 20 41

### Before
edge-before
ie11-before-2

### After
Screen Shot 2019-09-13 at 15 21 52
Screen Shot 2019-09-13 at 15 22 06

Fixes https://github.com/alphagov/govuk-frontend/issues/1519

<details>
<summary>
How text was resized in browsers
</summary>
### Chrome 
Settings > Appearance > Customise fonts

## Firefox
Set View > Zoom > Zoom text only. Press +/-
or Preferences > Language and appearance > Font size

## Safari
To increase the font size, press Option-Command-Plus sign (+)
To decrease the font size, press Option-Command-Minus sign (-)

## Edge 
Can't resize text apart from "Reading view" which removes form elements from the view.

## IE 8-11
Page > Text size > Largest

## iOS 
**Text can't resized in browser. See https://github.com/alphagov/govuk-frontend/issues/882**

## Android Chrome/Firefox
Settings > Accessibility > Text scaling
</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+### Fixes
+- [Pull request #1574: Make form elements scale correctly when text resized by user](https://github.com/alphagov/govuk-frontend/pull/1574).
+
 ## 3.2.0 (Feature release)
 
 ### New features
--- a/src/govuk/components/input/_input.scss
+++ b/src/govuk/components/input/_input.scss
@@ -13,6 +13,9 @@
     box-sizing: border-box;
     width: 100%;
     height: 40px;
+    @if $govuk-typography-use-rem {
+      height: govuk-px-to-rem(40px);
+    }
     margin-top: 0;
 
     padding: govuk-spacing(1);
--- a/src/govuk/components/select/_select.scss
+++ b/src/govuk/components/select/_select.scss
@@ -13,6 +13,9 @@
     box-sizing: border-box; // should this be global?
     max-width: 100%;
     height: 40px;
+    @if $govuk-typography-use-rem {
+      height: govuk-px-to-rem(40px);
+    }
     padding: govuk-spacing(1); // was 5px 4px 4px - size of it should be adjusted to match other form elements
     border: $govuk-border-width-form-element solid $govuk-input-border-colour;
 


nickcolley:
I think we may need to consider compatibility mode.

Our font scaling approach does something like this:

Compatibility mode turned off:
```css
.input {
  font-size: 16px;
}
```

Compatibility mode turned on:
```css
.input {
  font-size: 16px;
  font-size: 1rem;
}
```

If you have a look with legacy mode turned on everything is a lil squsho since GOV.UK Template sets it's root font size to 50% percent (or something close)

https://govuk-frontend-review-pr-1574.herokuapp.com/examples/form-alignment?legacy=1

36degrees:
> I think we may need to consider compatibility mode.

Good shout. I think we could mirror what we do in the typography helpers:

https://github.com/alphagov/govuk-frontend/blob/882333f762d91b0bcd4b1ec275e3ead51b34f5b6/src/govuk/helpers/_typography.scss#L146-L148

We could probably tie it to the same flag, as in theory we only need to resize the box if the font-size is in em. So maybe something like:

```scss
.whatever {
    height: 40px;
    @if $govuk-typography-use-rem {
        height: govuk-px-to-rem(40px);
    }
}
```

Thoughts?

nickcolley:
It feels like we're stretching that variable a bit, but considering this is all related to the scaling of the typography, I think we'll get away with it... :pray:

hannalaakso:
Thanks both for the good suggestions 👍

nickcolley:
The resizing of text docs is very good to know too thanks

Issue #1292, opened 2019-04-25
Add ability to override language for page title and `main` in Template


aliuk2012:
It is already possible to set the `lang` attribute on the `html` in Template. However according to https://github.com/alphagov/govuk_template/issues/342   there is a need to be able to set language of the page title and `main` separately. 

Split from: https://github.com/alphagov/govuk-frontend/issues/971

Part of https://github.com/alphagov/govuk-design-system/issues/777

aliuk2012:
Needed as part of https://github.com/alphagov/govuk-design-system-backlog/issues/185

PR #1576, opened 2019-09-17
Allow language to be set on <title> and <main> of page template

hannalaakso:
This allows for cases when some of the page elements, such as navigation and footer, are in eg. English whereas the title and page content are in a different language.

From [W3C](https://www.w3.org/International/questions/qa-html-language-declarations):  
> Always use a language attribute on the html tag to declare the default language of the text in the page. When the page contains content in another language, add a language attribute to an element surrounding that content.

Also discussed [here](https://github.com/alphagov/govuk_template/issues/342).

Fixes https://github.com/alphagov/govuk-frontend/issues/1292

See also [proposed guidance](https://github.com/alphagov/govuk-design-system/pull/1046).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 ## Unreleased
 
+### New features
+
+#### Allow `lang` to be set on `<title>` and `<main>` of template
+
+You can now set the [lang attribute](https://www.w3.org/International/questions/qa-html-language-declarations) in the title and main of page template.
+
+This will help with scenarios where some of the elements, such as navigation and footer, are in English whereas the title and page content are in a different language.
+
+- [Pull request #1576: Allow `lang` to be set on title and main of template](https://github.com/alphagov/govuk-frontend/pull/1576).
+
 ### Fixes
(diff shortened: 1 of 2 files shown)


src/govuk/template.njk
```diff
@@ -7,7 +7,7 @@
 <html lang="{{ htmlLang | default('en') }}" class="govuk-template {{ htmlClasses }}">
   <head>
     <meta charset="utf-8" />
-    <title>{% block pageTitle %}GOV.UK - The best place to find government services and information{% endblock %}</title>
+    <title {% if pageTitleLang %}lang="{{ pageTitleLang }}"{% endif %}>{% block pageTitle %}GOV.UK - The best place to find government services and information{% endblock %}</title>
```
nickcolley: Nit: to avoid a default of `<title >` could considering putting the space within the if statement.

hannalaakso: 👍

CHANGELOG.md
```diff
@@ -2,6 +2,16 @@
 
 ## Unreleased
 
+### New features
+
+#### Allow `lang` to be set on `<title>` and `<main>` of template
```
nickcolley: Nice CHANGELOG entry 👍

Issue #1384, opened 2019-05-28
Investigate flakey banner test [TIMEBOX: 1 hour]


nickcolley:
Maybe to do with cookies being left around, I'm not sure but it is skipped for now: https://github.com/alphagov/govuk-frontend/pull/1383/commits/c44602d578a8a4a8933028823d1516d5ad6a99f0#diff-bac8f4ef40104d619366dd4794cacd26L32

PR #1577, opened 2019-09-18
Remove skipped tests

nickcolley:
Since these tests add limited value we should remove them.

Closes https://github.com/alphagov/govuk-frontend/issues/1384

### DIFF
--- a/app/banner.test.js
+++ b/app/banner.test.js
@@ -44,22 +44,4 @@ describe('Banner', () => {
       done(err)
     })
   })
-
-  it.skip('should be dismissable', done => {
-    requestPath.post('/hide-banner', {
-      followAllRedirects: true,
-      jar: true // enable cookies
-    }, (err, res) => {
-      const $ = cheerio.load(res.body)
-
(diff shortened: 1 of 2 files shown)


app/banner.test.js
```diff
@@ -44,22 +44,4 @@ describe('Banner', () => {
       done(err)
     })
   })
-
-  it.skip('should be dismissable', done => {
```
nickcolley: This test is not really that useful since we'd notice if the button didnt work anymore, and if it broke it would just impact us on the dev team and would be a minor annoyance.

src/govuk/components/back-link/template.test.js
```diff
@@ -16,16 +16,6 @@ describe('back-link component', () => {
     expect(results).toHaveNoViolations()
   })
 
-  it.skip('fails to render if the required fields are not included', () => {
```
nickcolley: this test we wont bring back since the ship has sailed on doing this sort of thing...

Issue #1352, opened 2019-05-15
Consider a align-top modifier for Warning Text component


dashouse:
The Warning Text icon is always vertically centred within it's container. With a long warning message, the icon could appear unrelated to the text, for example appearing off screen on mobile.

We could add a modifier that changed the `top` value to ~20px to be used in this scenario.

**Default:**
<img width="221" alt="Screen Shot 2019-05-15 at 12 11 06" src="https://user-images.githubusercontent.com/23356842/57771470-b3b6f100-770a-11e9-9f70-7daeb1dbe8aa.png">

**Suggested modifier:**
<img width="213" alt="Screen Shot 2019-05-15 at 12 11 29" src="https://user-images.githubusercontent.com/23356842/57771482-bb769580-770a-11e9-9ba7-3e17f291d256.png">

PR #1578, opened 2019-09-19
Change position and spacing relationship of warning text icon

dashouse:
This PR addresses - https://github.com/alphagov/govuk-frontend/issues/1352

The primary purpose of the this PR was to align the icon with the top line, so that when the content wraps the icon doesn't vertically align with the entire block and potentially end up out of sight when the user starts reading the warning text.

To do this I had to also reduce the size of the icon slightly, and reduce the gap between the icon and the text accordingly.

It may seem a random size (29px) but this is helped it align with the first line, 30px created an unequal distribution of space.

### Before: Single line
Screen Shot 2019-09-19 at 09 32 44
 
### Before: Wrapping
Screen Shot 2019-09-19 at 09 33 10

### After: Single line
Screen Shot 2019-09-19 at 09 28 02

### After: Wrapping
Screen Shot 2019-09-19 at 09 28 53

I also removed some of the (I think) New Transport adjustments.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,12 @@ This will help with scenarios where some of the elements, such as navigation and
 
 - [Pull request #1576: Allow `lang` to be set on title and main of template](https://github.com/alphagov/govuk-frontend/pull/1576).
 
+#### Visual updates to the warning text component
+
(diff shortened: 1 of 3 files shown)


36degrees:
Browserstack screenshots:
https://www.browserstack.com/screenshots/a6b58c821d93372cac1412f7338fec30286b858d

36degrees:
This makes sense to me – just needs a changelog entry.

PR #1587, opened 2019-09-23
Fix height and alignment issue within header in Chrome 76+

### LINKED ISSUE #1575: Crown logo is not always aligned with logotype
Speaking with Dave it seems you can't reproduce this on his machine but I have been able to on Browserstack.

Screen Shot 2019-09-16 at 14 04 28

Safari shows the logo lower instead of higher.

36degrees:
A change in Chrome 76 means that the header is currently being rendered 64px high (rather than the expected 60px), with mis-alignment between the crown and the GOV.UK logotype.

Adding `display: inline-block` to the __logotype element fixes it ([h/t @trevorsaint](https://github.com/alphagov/govuk-frontend/issues/1575#issuecomment-533610517)).

(We are not sure what exactly changed in Chrome 76, nor why this fixes it.)

Fixes #1575 

👉 [Browser Screenshots](https://www.browserstack.com/screenshots/38a4e8401a166947d4c22804c12293d78105a22c)
👉 [Preview](https://govuk-frontend-review-pr-1587.herokuapp.com/components/header)

## Before
Header Before

## After
Header After


36degrees:
I've double checked and it renders exactly the same as the current header does in Chrome 75.

PR #1594, opened 2019-09-26
Refactor handling of count message in character count JS

hannalaakso:
The “count message” refers to the “You have x characters remaining” message in the character count (`govuk-character-count__message`).

## Original issue

The original issue #1106 raised by @36degrees was fixed by https://github.com/alphagov/govuk-frontend/pull/1347. This allowed a value for `aria-describedby` to be passed through to the textarea component. This means that the character count component can correctly pass the id of the count message to the textarea component to associate the two.

## Fixes in this PR

There was a bit of refactoring to do in terms of how the JS of the character count handles defining the count message so this PR does that. I’ve tried to break the changes into smaller commits to make it clear what’s going on but can squash them before merging if that makes more sense. 

### Making the count message markup required

The count message markup is now always present (since https://github.com/alphagov/govuk-frontend/commit/7b534bf734cd0a1be2cff34d64018aa89abdabb6). This PR makes the markup for it required, instead of giving the option to add it with Javascript - the message markup should always exist so that user is notified of the character limit, even if JS doesn’t run. 

This PR removes the branching logic and the code that added the count message markup dynamically in `createCountMessage()`. It’s worth nothing that the code inside the block hasn’t been executed since https://github.com/alphagov/govuk-frontend/commit/7b534bf734cd0a1be2cff34d64018aa89abdabb6.  As using the character count without  the count message markup and relying it to be added with JS is not part of the public API, removing this functionality is not considered a breaking change.

### Defining the count message

The count message DOM element is now defined at the top of the module with the other required DOM elements. 

[We check whether the textarea is present](https://github.com/alphagov/govuk-frontend/compare/character-count-associate-textarea?expand=1#diff-4243571b7fc266600817b0d8eb6a392bR8) before getting its `id` to get the count message DOM element - the check prevents an error with getting `$textarea.id` if the script is run without component markup being present present on the page. Alternatively we could define the count message in `init()` after doing a check for textarea which would require another check for it and wouldn’t make it as clear that it’s required as when placed at the top of module. See below for the code. Thoughts?

```
CharacterCount.prototype.init = function () {
  var $module = this.$module

  if (!$textarea) {
    return
  }

  this.$countMessage = $module.querySelector('[id=' + this.$textarea.id + '-info]')

  if (!$countMessage) {
    return
  }
```

### Moving the count message markup inside `govuk-form-group`

The original code for checking/creating the count message markup [moves it in the DOM](https://github.com/alphagov/govuk-frontend/blob/f4818118a71462972d4ec5d4b6409e89c8fbc56c/src/govuk/components/character-count/character-count.js#L104 ) to be inside `govuk-form-group`. This was probably useful when the message markup was created dynamically. This PR retains the bit of code that moves the element for backwards compatibility. It’s important that the count message markup is inside `govuk-form-group` so that it gets wrapped within the error block (see screen grab below).

Screen Shot 2019-09-26 at 17 57 26

However we should look to remove moving the markup with JS and instead change the component markup so that the count message is inside `govuk-form-group`. This would be a breaking change so shall leave it for now and create an issue.

### DIFF
--- a/src/govuk/components/character-count/character-count.js
+++ b/src/govuk/components/character-count/character-count.js
@@ -26,6 +26,10 @@ CharacterCount.prototype.init = function () {
     return
   }
 
+  // We move count message right after the field
+  // Kept for backwards compatibility
+  $textarea.insertAdjacentElement('afterend', $countMessage)
+
   // Read options set using dataset ('data-' values)
   this.options = this.getDataset($module)
 
@@ -43,9 +47,6 @@ CharacterCount.prototype.init = function () {
     return
   }
 
-  // Generate and reference message
-  var boundCreateCountMessage = this.createCountMessage.bind(this)
-  this.countMessage = boundCreateCountMessage()
 
   // If there's a maximum length defined and the count message exists
   if (this.countMessage) {
@@ -88,27 +89,6 @@ CharacterCount.prototype.count = function (text) {
     length = text.length
   }
   return length
-}
-
-// Generate count message and bind it to the input
-// returns reference to the generated element
-CharacterCount.prototype.createCountMessage = function () {
-  var countElement = this.$textarea
-  var elementId = countElement.id
-  // Check for existing info count message
-  var countMessage = document.getElementById(elementId + '-info')
-  // If there is no existing info count message we add one right after the field
-  if (elementId && !countMessage) {
-    countElement.insertAdjacentHTML('afterend', '<span id="' + elementId + '-info" class="govuk-hint govuk-character-count__message" aria-live="polite"></span>')
-    this.describedBy = countElement.getAttribute('aria-describedby')
-    this.describedByInfo = this.describedBy + ' ' + elementId + '-info'
-    countElement.setAttribute('aria-describedby', this.describedByInfo)
-    countMessage = document.getElementById(elementId + '-info')
-  } else {
-  // If there is an existing info count message we move it right after the field
-    countElement.insertAdjacentElement('afterend', countMessage)
-  }
-  return countMessage
 }
 
 // Bind input propertychange to the elements and update based on the change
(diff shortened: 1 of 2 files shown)


36degrees:
Couple of minor typos in the commit messages:

3b52b27a34d03de78645cf6f7ab1aed43b2e3f0e
> We check whether the textarea is present before **trying to its id** to find the count message

814935a1afc2925bd9278f896160f414431a261e
> It’s worth **nothing** that the code inside the block hasn’t been executed since **[]**.

hannalaakso:
Thanks @36degrees I've just fixed those

src/govuk/components/character-count/character-count.js
```diff
@@ -5,6 +5,7 @@ import '../../vendor/polyfills/Element/prototype/classList'
 function CharacterCount ($module) {
   this.$module = $module
   this.$textarea = $module.querySelector('.govuk-js-character-count')
+  if (this.$textarea) this.$countMessage = $module.querySelector('[id=' + this.$textarea.id + '-info]')
```
36degrees: I think we tend to avoid inline `if` statements like this

```suggestion
  if (this.$textarea) {
    this.$countMessage = $module.querySelector('[id=' + this.$textarea.id + '-info]')
  }
```

src/govuk/components/character-count/character-count.js
```diff
@@ -17,10 +18,16 @@ CharacterCount.prototype.init = function () {
...
-  if (!$textarea) {
+  var $countMessage = this.$countMessage
+
+  if (!$textarea || !$countMessage) {
     return
   }
 
+  // We move count message right after the field
+  // Kept for backwards compatibility
+  $textarea.insertAdjacentElement('afterend', this.$countMessage)
```
36degrees: Can we just use `$countMessage` here? Alternatively, as we're currently only using `$countMessage` in one place, can we remove the local variable and just use `this.$countMessage`?

hannalaakso: Yeps good catch 👍

PR #1595, opened 2019-09-27
Do not output conditionally revealed content for radios or checkboxes when it's empty

### LINKED ISSUE #1546: Radios and checkboxes with empty conditionals causes odd spacing
Raising as separate ticket from #1494

If you have an item with a conditional and the conditional content is empty, when you select that item, you get weird spacing:
61715646-7186dd00-ad55-11e9-9e97-32f9d7ddad17

I think this is because conditional items are given `margin-bottom: 20px`. When the item is selected, the conditional is made visible so the margin is added.

Similar to #1494 I might check if conditional contains any items - if not, don't render it.

36degrees:
If `item.conditional.html` is falsy, perhaps because it’s referencing a variable that is conditionally assigned, the macro currently outputs an empty `div.govuk-checkboxes__conditional` and associates it with the input using aria-controls.

By testing for `item.conditional.html` being truthy (which is the only property in the item.conditional object used within the macro) we can avoid this.

Fixes #1546


edwardhorsford:
Woop! 🎉

Issue #1449, opened 2019-06-12
Investigate alternative to [type=number] for the date input component


nickcolley:
We are aware of two issues with assistive technologies and their compatibility with [type=number] 
- https://github.com/alphagov/govuk-frontend/issues/1408
- https://github.com/alphagov/govuk-frontend/issues/1101

We should investigate an alternative to this as we're aware of general issues with [type=number].

@hannalaakso has done a lot of investigation into this.

hannalaakso:
If collecting dates as numbers, the biggest problem with `type="number"` is that it has native input validation in Chrome (desktop and Android) which prevents the typing of non-numeric values. This is not a bug but an implementation of the [HTML spec](https://www.w3.org/TR/html5/sec-forms.html#number-state-typenumber).  Unfortunately the user experience is poor when it comes to handling errors. We recommend the [use of clear error messages](https://design-system.service.gov.uk/components/error-message/), whereas there is no information given to user about what they’ve done wrong if they attempt to type the month as a number, for example. 

`type="number"` also allows the user to increment numbers by pressing the up/down arrows. This could improve usability for some users (as long as the number should be incrementable - dates generally are) but also introduces the risk of some users accidentally incrementing a number using the arrow keys when trying to navigate a form. The incrementing behaviour can be disabled using Javascript but we prefer to avoid scripting solutions if possible.

There are other issues with `type="number"` that concern long numeric values being validated by browsers and making sure that users can enter "-" and "," signs; but these aren’t a concern for dates. 

Alternative solution (we should test this across the board) is to use `inputmode="numeric"` combined with `pattern`:

```
<input type="text" pattern="[0-9]*" id="sometext" name="sometext" inputmode="numeric">
```

There is also `type="tel"` which has no browser validation, and brings up a numeric keypad on all mobile devices (the input should set `autocomplete=off` to stop the browser trying to complete it from the user’s address book). It's often used by payment providers to collect credit card numbers. However on Android devices the keyboard includes additional buttons for "Pause" and "Wait" which could be confusing for some users. Additionally, there is the (admittedly very slim) possibility that browser vendors might in the future add validation to `type="tel"`, for which reason if it was used on a service, there should be a team available to monitor the behaviour.

nickcolley:
Just a note on the pattern attribute suggested, if someone has HTML5 validation enabled (something we do not recommend), it will show a message and stop the form from being submitted:

[...]

PR #1599, opened 2019-10-02
WIP: Updating date component to be type text and allow inputmode

htmlandbacon:
Opening for discussion, following on from [pull request 1527](https://github.com/alphagov/govuk-frontend/pull/1527) - I've applied this to the date component.

Before I go too far I figured there is a couple that probably needs talking through.

- I've allowed type and inputmode to be overridden - should this be the case?

- Given we recommend not using html5 validation, should we remove the pattern params?

- We should update the example with maxlength attribute?

My thoughts where no, yes and yes.

Happy to take onboard any addtional comments 👍 

I've done manual testing in on Browserstack and had a Dragon user test it.

Following on from https://github.com/alphagov/govuk-frontend/issues/1449#issuecomment-504006087

### DIFF
--- a/src/govuk/components/date-input/date-input.yaml
+++ b/src/govuk/components/date-input/date-input.yaml
@@ -20,6 +20,14 @@ params:
     type: string
     required: true
     description: Item-specific name attribute.
+  - name: type
+    type: string
+    required: false
+    description: Sets the input type, default is text.
+  - name: inputmode
+    type: string
+    required: false
+    description: Sets the input inputmode, default is numric.
   - name: label
     type: string
     required: true
--- a/src/govuk/components/date-input/template.njk
+++ b/src/govuk/components/date-input/template.njk
@@ -65,7 +65,8 @@
         classes: "govuk-date-input__input " + (item.classes if item.classes),
         name: (params.namePrefix + "-" + item.name) if params.namePrefix else item.name,
         value: item.value,
-        type: "number",
+        type: item.type if item.type else "text",
+        inputmode: item.inputmode if item.inputmode else "numeric",
         autocomplete: item.autocomplete,
         pattern: item.pattern if item.pattern else "[0-9]*",
         attributes: item.attributes
--- a/src/govuk/components/date-input/template.test.js
+++ b/src/govuk/components/date-input/template.test.js
@@ -131,7 +131,7 @@ describe('Date input', () => {
       expect($firstItems.text().trim()).toEqual('Day')
     })
 
-    it('renders inputs with type="number"', () => {
+    it('renders inputs with type="text"', () => {
       const $ = render('date-input', {
         items: [
           {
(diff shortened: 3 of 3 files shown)


nickcolley:
Heya Colin,

> I've allowed type and inputmode to be overridden - should this be the case?

I think unless you have a very clear reason why someone wants to do this, we should not allow them to be overridden for now as it expands the API surface area without a need.

> Given we recommend not using html5 validation, should we remove the pattern params?

We should consider this but it would be a breaking change so would need to be against the 4.0.0 milestone, so if we want to do this we should raise a new issue to track removing that.

> We should update the example with maxlength attribute?

Sorry if I've missed some important context but we typically avoid using this attribute as it is less usable, copy pasting, typing without looking, autofill etc.

> I've done manual testing in on Browserstack and had a Dragon user test it.

Nice! We're the results positive?

htmlandbacon:
Thanks for the info, I'll make them changes, and work out what else to raise!

In terms of the below, is this documented any where? Be a great thing to point people to if so!

> Sorry if I've missed some important context but we typically avoid using this attribute as it is less usable, copy pasting, typing without looking, autofill etc.

Also yes, which shouldn't be a surprise since it has wide support https://caniuse.com/#search=inputmode

nickcolley:
I was talking about you comment on `maxlength` .

> We should update the example with maxlength attribute?

I'm not sure what you're referring by this actually, I think I got confused haha.

It looks like to ship this we might want to do some work alongside it as noted in the original comment: https://github.com/alphagov/govuk-frontend/issues/1449#issuecomment-504006087

So I've put awaiting triage on this for us to chat as a team about how we go about doing that.

hannalaakso:
@htmlandbacon Thanks for your patience. There are some [accompanying actions/guidance](https://github.com/alphagov/govuk-frontend/issues/1449#issuecomment-504006087) we want to do alongside this work so we're not able to merge this yet. We're going to look at this in a few weeks' time when we'll be doing accessibility fixes.

edwardhorsford:
> Given we recommend not using html5 validation, should we remove the pattern params?

I might be wrong, but I think this is here so iOS displays a numeric keypad.

36degrees:
> I might be wrong, but I think this is here so iOS displays a numeric keypad.

Just to clarify, in iOS 12.2 and above `inputmode` dictates the keyboard used. I _believe_ `pattern` can still be used, but has no effect if `inputmode` is set.

IMO we should base our decision on dropping `pattern` on the amount of traffic we're seeing from devices running iOS < 12.2.

nickcolley:
Hey Colin,

Just an update: we were hoping to push forwards with this soon but have had to re-prioritise some stuff to focus on making our contribution and support more sustainable, so we may not get back to this for a while longer.

Thanks for your patience 👍

htmlandbacon:
> Hey Colin,
> 
> Just an update: we were hoping to push forwards with this soon but have had to re-prioritise some stuff to focus on making our contribution and support more sustainable, so we may not get back to this for a while longer.
> 
> Thanks for your patience 👍

I've been away so no issue. 😂

36degrees:
Do we need to make `inputmode` and `type` configurable? I can't think of a situation where a user would want to deviate from `numeric` and `text` respectively.

36degrees:
> > I've allowed type and inputmode to be overridden - should this be the case?
> 
> I think unless you have a very clear reason why someone wants to do this, we should not allow them to be overridden for now as it expands the API surface area without a need. If we then get a clear need to override this in the future we can add this functionality without a breaking change.

Agreed, can we remove this?

> > Given we recommend not using html5 validation, should we remove the pattern params?
> 
> We should consider this but it would be a breaking change so would need to be against the 4.0.0 milestone, so if we want to do this we should raise a new issue to track removing that.

I've raised #1701 to address this.

@htmlandbacon we're looking to pick this up this sprint – would you have time to remove the ability to customise the input `type` and `inputmode`, or shall we take that on? Thanks again for your patience with this one.

36degrees:
@htmlandbacon thanks very much for your work on this so far – we can pick this up and take it forward.

htmlandbacon:
> @htmlandbacon thanks very much for your work on this so far – we can pick this up and take it forward.

No worries, apologies for not replying - I was trying to work out if I could progress it this week!

Let me know if there is anything I can support on 🎈

Issue #1611, opened 2019-10-11
All sentences should be inside <p> (or the relevant semantic elements)


colinrotherham:
I'm going to add a few points that came out of a recent developer review, before three of our services are sent for an external accessibility audit.

### Inset text and Panel components
We're currently using the `govukInsetText()` and `govukPanel()` macros and we've been flagged during a developer review for putting text inside `<div>` tags.

All three services were caught out by this.

The non-semantic HTML is the recommended usage on the GOV.UK Design System examples shown here:

https://design-system.service.gov.uk/components/inset-text/
https://design-system.service.gov.uk/components/panel/

You can see in these screenshots:

![Inset text](https://user-images.githubusercontent.com/415517/66648305-24632a80-ec23-11e9-8fa0-17627317eeb0.png)
![Panel](https://user-images.githubusercontent.com/415517/66648307-24632a80-ec23-11e9-9119-df05dd82e050.png)

Since other service teams may copy from the examples and make the same mistakes as us, I'd prefer it if the examples included semantic HTML by default! 😊 

Suggested change:

**Before**
```js
{{ params.html | safe if params.html else params.text }}
```

**After**
```js
{% if params.html %}{{ params.html | safe }}{% else %}<p>{{ params.text }}</p>{% endif %}
```

But other macros might be affected too.

It's been raised before in https://github.com/alphagov/govuk-design-system/issues/822

Thanks

nickcolley:
Pre-triage comment: I believe that there are other components we'd want to review as part of considering this.

colinrotherham:
Sorry just realised we're flagged on one service only with **warning text**.

I'd expect a wrapping paragraph tag here too.

![Warning text](https://user-images.githubusercontent.com/415517/66650144-cbe25c00-ec27-11e9-8918-28d20f5a1004.png)

Thanks @nickcolley.

hannalaakso:
@colinrotherham We'd be interested in hearing whether this gets flagged in your external accessibility audit.

Have added this to 4.0 milestone as this would be a breaking change.

adamliptrot-oc:
WHATWG deem the`<p>` element as only one way of markup up a paragraph. Semantically it seems not to have any impact. There could be styling implications of course.
https://html.spec.whatwg.org/multipage/dom.html#paragraph

nickcolley:
Thanks @adamliptrot-oc, that's good context, I've had a chat about this with some people on the team...

[...]

Issue #1385, opened 2019-05-28
GOV.UK Frontend accessibility improvements


timpaul:
## Must
- [ ] Review if `aria-label` in header navigation is necessary #1340
- [x] Conditional reveals with an input not being announced on JAWS #935
- [x] Conditionally revealing content with checkboxes not announcing collapsed/expanded state on macOS with VoiceOver #989
- [x] Conditional reveal radio component has aXe errors #979
- [x] Fieldset legend text can be clipped when using a custom or fallback font #1239
- [x] Descenders clipped from letters such as qyp when select boxes are in an error state #1534 
- [x] Form elements do not scale well if only font size increased #1519
- [x] Error Summary link handling can result in obscured inputs #1544

## Should
- [ ] All caps text in the tag component is harder to read #1504
- [x] Dragon Accessibility issue with DOB field #1101
- [x] Summary list action links visually hidden text are read out without spaces in screen readers #1643
- [x] Investigate adding 'aria-invalid' attribute on form inputs that have errors #1168
- [x] Hint for fieldset not announced if first input has hint #1696
- [ ] Link colour change on hover not clear #1417
- [x] Screen-readers may not 'announce' lists if 'list-style: none' is used #1471
- [x] Investigate Dynamic Type to improve accessibility for iOS users #882
- [x] Details component not announcing collapsed/expanded state on iOS with VoiceOver #988 
- [x] Investigate why role="group" is applied to details component automatically #1483 
- [x] Character count count message is not associated with textarea #1106
- [x] Character count remaining characters message not visible #1406
- [x] Investigate colour contrast issues with secondary text colour #1134

## Could
- [x] Consider making tabs `<section>` into `<div>` #1339
- [ ] Tabs interface needs better way to get to panel content #860
- [x] Wrapping component text with `<p>` where semantics are missing #1611
- [x] Investigate using more robust method for detecting updates to input fields in Dragon #1028

automaticdreams:
automated testing with axe flags .govuk-footer__copyright-logo as an issue because it cannot calculate contrast as a background image is used without a background colour

36degrees:
Closing in favour of the accessibility label.

PR #1613, opened 2019-10-11
Wrap component text with <p> where semantics are missing

colinrotherham:
The **Inset text** and **Panel** components are styled `<div>` tags.

Plain text via `params.text` should be wrapped with `<p></p>` to improve semantics, and it's better to have sensible defaults than to assume developers will use `params.html`.

I've also included a fix for **Warning text** but due to enclosing `<strong>` and `<span>` it's a slightly different fix to the other two.

Fixes https://github.com/alphagov/govuk-frontend/issues/1611 and https://github.com/alphagov/govuk-design-system/issues/822

**Before**
Before

**After**
After

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixes
+
+- [Pull request #1613: Wrap component text with <p> where semantics are missing](https://github.com/alphagov/govuk-frontend/pull/1574).
+
 ## 3.3.0 (Feature release)
 
 ### New features
--- a/src/govuk/components/accordion/_accordion.scss
+++ b/src/govuk/components/accordion/_accordion.scss
@@ -41,7 +41,13 @@
     margin-bottom: 0;
   }
 
-  // Remove the bottom margin from the last item inside the content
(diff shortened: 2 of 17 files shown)


nickcolley:
Thanks for jumping on this, I believe there may be other components (such as accordions), that we would want to consider this.

I've put the awaiting triage label on this so the GOV.UK Design System team can get together and decide if this is the direction we want to go in.

hannalaakso:
We've added this to our list of [Accessibility Improvements for GOV.UK Frontend](https://github.com/alphagov/govuk-frontend/issues/1385) under "Could" as we're not sure yet if this is the correct approach. Thanks for raising @colinrotherham 👍

hannalaakso:
As this would be a breaking change which could break our users' projects we're not able to merge this right now. We would like to look at this but will have to wait until we release GOV.UK Frontend 4.0 which is not yet imminent.

Thanks for all your work on this @colinrotherham we'll use this PR as the basis of exploring this for 4.0. We're going to close this PR and have added the issue you raised to the [4.0 milestone](https://github.com/alphagov/govuk-frontend/milestone/22).

PR #1614, opened 2019-10-14
aria hide summary list action text if visually hidden text is set

gunjam:
Currently the summary-list component offers a `visuallyHiddenText` property to add hidden text to action links for use with screen readers, the only problem is it is designed on the assumption that the text will follow on naturally from the action text. While fine in English, the gramatical order of the sentence may be completely different once translated - where a 'change' verb could be in the middle of a sentence rather at the beginning, or when written as sentence may be a completely different word from that which you'd use as a singular call to action.

Because of this I currently have locales files full of spans which I'm passing to the html text prop, which is a bit messy.

With this pull request I've refactored the component to wrap the normal action text in an aria hidden span if visually hidden text is set, so you can use entirely separate pieces of content for both sighted and screen reader users.

The only downside is it would be a breaking change as you'd have to refactor your use of the component from:

```javascript
{
  actions: {
    items: [
      {
        text: "Change",
        visuallyHiddenText: "name"
      }
    ]
  }
}
```
to
```javascript
{
  actions: {
    items: [
      {
        text: "Change",
        visuallyHiddenText: "Change name"
      }
    ]
  }
}
```

### DIFF
--- a/src/govuk/components/summary-list/template.njk
+++ b/src/govuk/components/summary-list/template.njk
@@ -1,8 +1,10 @@
 {%- macro _actionLink(action) %}
   <a class="govuk-link {%- if action.classes %} {{ action.classes }}{% endif %}" href="{{ action.href }}" {%- for attribute, value in action.attributes %} {{attribute}}="{{value}}"{% endfor %}>
-    {{ action.html | safe if action.html else action.text }}
+    {% set actionText = action.html | safe if action.html else action.text %}
     {%- if action.visuallyHiddenText -%}
-      <span class="govuk-visually-hidden"> {{ action.visuallyHiddenText }}</span>
+      <span aria-hidden="true">{{ actionText }}</span><span class="govuk-visually-hidden"> {{ action.visuallyHiddenText }}</span>
+    {%- else -%}
+      {{ actionText }}
     {% endif -%}
   </a>
 {% endmacro -%}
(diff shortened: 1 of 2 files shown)


hannalaakso:
@gunjam Thanks a lot for your work on this 🙌 It looks like this would be a breaking change for our users. I've added this to our team triage next week so that we can discuss how to best proceed.

hannalaakso:
@gunjam As this would be a breaking change we're not able to merge this right now. We would like to look at this but will have to wait until we release GOV.UK Frontend 4.0 which is not yet imminent. We're going to close use this PR as it's going to fall behind but we'll use it as the basis for  exploring a solution when working on 4.0 fixes.

As you mentioned you can currently get around this by including the content with `action.html` although it's not an ideal solution.

Are you happy to create an issue about this which we can assign to the [4.0 milestone](https://github.com/alphagov/govuk-frontend/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+milestone%3Av4.0.0) or shall I do this on your behalf?

Thanks again for working on this.

nickcolley:
@gunjam it can be dangerous but have you considered using `aria-label` on the anchor?

```json
{
  actions: {
    items: [
      {
        text: "Change",
        attributes: {
            "aria-label": "Change name"
       }
      }
    ]
  }
}
```

nickcolley:
I spoke to @gunjam on Slack and they said that it's OK to 'shelve' this for now given there's workarounds but that we should consider this sort of thing when looking into translations for example Welsh support.

Issue #1280, opened 2019-04-18
List known validation errors/warnings


hannalaakso:
## Description

Once we've finished the [WCAG 2.1 compliance work](https://github.com/alphagov/govuk-design-system/issues/677), we should consider increasing visibility of any known validation errors with our code due to accessibility, legacy etc fixes. This would help users to know which errors/warnings are known / expected.

An example would be the `role` attribute we add to `<header>`, `<footer>`, `<button>` for older browsers that don't support HTML5 elements, and the [`src` on SVG issue](https://github.com/alphagov/govuk-frontend/blob/master/src/components/header/template.njk#L33). 

## Checklist
- [x] Review and list existing / known issues -> [resulting list](https://github.com/alphagov/govuk-frontend/issues/1280#issuecomment-509588851) 
- [x] Triage [list](https://github.com/alphagov/govuk-frontend/issues/1280#issuecomment-509588851) with team to decide which ones need to be investigated / fixed and which ones should be documented in the Design System / somewhere as known issues.

aliuk2012:
Linked to https://github.com/alphagov/govuk-design-system/issues/677

hannalaakso:
WIP here: https://docs.google.com/spreadsheets/d/1bK127XvQJ72m8Xtn5006BKDmQlT5d-JIGM_a05w1hIA/

hannalaakso:
This could be useful https://stevefaulkner.github.io/Articles/On%20HTML%20belts%20and%20ARIA%20braces%20(The%20Default%20Implicit%20ARIA%20semantics%20they%20didn%E2%80%99t%20want%20you%20to%20know%20about).html (thanks @leekowalkowski-hmrc)

hannalaakso:
## Known validation errors / warnings

We need to investigate/document these.

| Component (s)                              | Message                                                                                              | Explanation / issue raised                                                                                                                                                                                                     | Flagged by                   |
|--------------------------------------------|------------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|------------------------------|
| Button - disabled                          | Attribute aria-disabled is unnecessary for elements that have attribute disabled                     |                                                                                                                                                                                                                                | https://validator.w3.org     |
[...]

PR #1616, opened 2019-10-21
Update GOV.UK Frontend accessibility documentation

m-green:
This updates the Frontend README and compatibility mode documentation with information about Frontend's accessibility.

To keep the README as concise as possible and reduce duplication, I've also linked to content (such as https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices#browsers-to-test-in) rather than repeat it here.

Ready to merge when:

- [ ] Nick and Hanna have checked the facts - in particular please check that the list of issues and accessibility checker warnings is accurate
- [ ] Amy has 2i'd
- [ ] A technical writer has 2i'd

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,72 +33,33 @@ We recommend [installing GOV.UK Frontend using node package manager
 You can also [download the compiled and minified assets (CSS, JavaScript) from
 GitHub](docs/installation/installing-from-dist.md).
 
-## Browser support
+## Browser and assistive technology support
 
-GOV.UK Frontend will allow you to build services that comply with the [guidance
-in the Service Manual][service-manual-browsers].
+GOV.UK Frontend supports:
 
-Currently, GOV.UK Frontend officially supports the following browsers:
+- [recommended browsers](https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices#browsers-to-test-in)
+- [recommended assistive technologies](https://www.gov.uk/service-manual/technology/testing-with-assistive-technologies#what-to-test)
+- Internet Explorer 8, 9 and 10, although components may not look perfect
+- your users overriding colours in Windows, Firefox and Chrome
 
-| Operating system | Browser                                | Support     |
-|----------------- |----------------------------------------|-------------|
-| Windows          | Internet Explorer 8-10                 | functional  |
-| Windows          | Internet Explorer 11                   | compliant   |
-| Windows          | Edge (latest 2 versions)               | compliant   |
-| Windows          | Google Chrome (latest 2 versions)      | compliant   |
-| Windows          | Mozilla Firefox (latest 2 versions)    | compliant   |
-| macOS            | Safari 9+                              | compliant   |
-| macOS            | Google Chrome (latest 2 versions)      | compliant   |
-| macOS            | Mozilla Firefox (latest 2 versions)    | compliant   |
-| iOS              | Safari for iOS 9.3+                    | compliant   |
-| iOS              | Google Chrome (latest 2 versions)      | compliant   |
-| Android          | Google Chrome (latest 2 versions)      | compliant   |
-| Android          | Samsung Internet (latest 2 versions)   | compliant   |
+## Accessibility
 
-‘Compliant’ means that the components must look as good and function as well as
-they do in other modern browsers.
+The GOV.UK Design System team works hard to ensure that GOV.UK Frontend is accessible.
 
-'Functional' means the components may not look perfect in the given browser, but
-must still be usable without errors and without 'looking broken'.
+Using Frontend will help your service meet [level AA of WCAG 2.1](https://www.gov.uk/service-manual/helping-people-to-use-your-service/understanding-wcag). But you must still [check that your service meets accessibility requirements](https://www.gov.uk/service-manual/helping-people-to-use-your-service/making-your-service-accessible-an-introduction), especially if you extend or modify components.
 
-If you are including GOV.UK Frontend as part of a stylesheet that you are
-generating in your application's build pipeline, you will need to [generate and
-include a separate stylesheet in order to support Internet Explorer
-8](docs/installation/supporting-internet-explorer-8.md).
+You should also use:
 
-[service-manual-browsers]: https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices#browsers-to-test-in
+- [the JavaScript from GOV.UK Frontend](https://github.com/alphagov/govuk-frontend/blob/master/docs/installation/installing-with-npm.md#using-javascript)
+- [a separate stylesheet](https://github.com/alphagov/govuk-frontend/blob/master/docs/installation/supporting-internet-explorer-8.md) if you support Internet Explorer 8
 
-## Assistive technology support
+^ Your service will not meet level AA of WCAG 2.1 if you use [compatibility mode](https://github.com/alphagov/govuk-frontend/blob/master/docs/installation/compatibility.md) to use GOV.UK Frontend with old frameworks or the old colour palette.
 
-GOV.UK Frontend will allow you to build services that comply with the [guidance
-in the Service Manual][service-manual-assistive-technologies].
+You can also read the [accessibility statement for the GOV.UK Design System](https://design-system.service.gov.uk/accessibility/).
 
-Currently, GOV.UK Frontend officially supports the following assistive technologies:
+### Accessibility warnings
 
-| Software                 | Version        | Type               | Browser                   |
-|--------------------------|----------------|--------------------|---------------------------|
-| JAWS                     | 15 or later    | screen reader      | Internet Explorer 11      |
-| ZoomText                 | 10.11 or later | screen magnifier   | Internet Explorer 11      |
-| Dragon NaturallySpeaking | 11 or later    | speech recognition | Internet Explorer 11      |
-| NVDA                     | 2016 or later  | screen reader      | Firefox (latest versions) |
-| VoiceOver                | 7.0 or later   | screen reader      | Safari on iOS10 and OS X  |
-
-In addition, we test that all content is accessible with keyboard only.
-
-We aim to support [users who adjust or override the colours of websites they visit][how-users-change-colours-on-websites]. We test this by [changing colours in Firefox][changing-colours-in-firefox], by [enabling 'High Contrast' mode in Windows][enabling-high-contrast-mode-in-windows] and by using the [High Contrast plugin for Chrome][high-contrast-plugin-for-chrome].
-
-[service-manual-assistive-technologies]: https://www.gov.uk/service-manual/technology/testing-with-assistive-technologies#what-to-test
-
-[changing-colours-in-firefox]:
-https://support.mozilla.org/en-US/kb/change-fonts-and-colors-websites-use
-
-[enabling-high-contrast-mode-in-windows]:
-https://support.microsoft.com/en-gb/help/13862/windows-use-high-contrast-mode
-
-[high-contrast-plugin-for-chrome]: https://chrome.google.com/webstore/detail/high-contrast/djcfdncoelnlbldjfhinnjlhdjlikmph?hl=en-US
-
-[how-users-change-colours-on-websites]:
-https://accessibility.blog.gov.uk/2017/03/27/how-users-change-colours-on-websites/
+If you get a warning from a linter or accessibility checker, check our list of [issues you should not need to fix](https://github.com/alphagov/govuk-frontend/issues/1280#issuecomment-509588851).
 
 ## Getting updates
 
--- a/docs/installation/compatibility.md
+++ b/docs/installation/compatibility.md
@@ -10,8 +10,7 @@ frameworks:
 You cannot use the following features if you installed GOV.UK Frontend from
 `dist`. [Install with node package manager](installing-from-npm.md) instead.
 
-If you use these features, your service will not have GOV.UK Frontend's many
-accessibility and usability improvements.
+If you use these features, your service will not meet [level AA of WCAG 2.1](https://www.gov.uk/service-manual/helping-people-to-use-your-service/understanding-wcag) because your users will not be able to change the font size.
 
 ## Turn on 'compatibility mode'
 


README.md
```diff
@@ -33,72 +33,50 @@ We recommend [installing GOV.UK Frontend using node package manager
...
+You should also use:
 
-‘Compliant’ means that the components must look as good and function as well as
-they do in other modern browsers.
+- [the JavaScript from GOV.UK Frontend](https://github.com/alphagov/govuk-frontend/blob/master/docs/installation/installing-with-npm.md#using-javascript)
+- [a separate stylesheet](https://github.com/alphagov/govuk-frontend/blob/master/docs/installation/supporting-internet-explorer-8.md) if you support Internet Explorer 8
 
-'Functional' means the components may not look perfect in the given browser, but
-must still be usable without errors and without 'looking broken'.
+Your service will not meet level AA of WCAG 2.1 if you use [compatibility mode](https://github.com/alphagov/govuk-frontend/blob/master/docs/installation/compatibility.md).
```
nickcolley: Do we need to say why? Will this put people off migrating if we're not reassuring that this is still a good thing?

m-green: Yep definitely - I've added the 'why' to the docs on the compatibility mode itself.

README.md
```diff
@@ -33,72 +33,50 @@ We recommend [installing GOV.UK Frontend using node package manager
...
-https://support.mozilla.org/en-US/kb/change-fonts-and-colors-websites-use
-
-[enabling-high-contrast-mode-in-windows]:
-https://support.microsoft.com/en-gb/help/13862/windows-use-high-contrast-mode
-
-[high-contrast-plugin-for-chrome]: https://chrome.google.com/webstore/detail/high-contrast/djcfdncoelnlbldjfhinnjlhdjlikmph?hl=en-US
-
-[how-users-change-colours-on-websites]:
-https://accessibility.blog.gov.uk/2017/03/27/how-users-change-colours-on-websites/
+- `aria-disabled` is invalid in disabled elements.
```
nickcolley: Again I don't think we should list these out if possible, I'd prefer to keep them in the issue tracker where they'll always be up to date.

We could consider breaking that issue into multiple issues that we then close and give a consistent label.

README.md
```diff
@@ -33,72 +33,50 @@ We recommend [installing GOV.UK Frontend using node package manager
...
-| iOS              | Safari for iOS 9.3+                    | compliant   |
-| iOS              | Google Chrome (latest 2 versions)      | compliant   |
-| Android          | Google Chrome (latest 2 versions)      | compliant   |
-| Android          | Samsung Internet (latest 2 versions)   | compliant   |
+You should also use:
 
-‘Compliant’ means that the components must look as good and function as well as
-they do in other modern browsers.
+- [the JavaScript from GOV.UK Frontend](https://github.com/alphagov/govuk-frontend/blob/master/docs/installation/installing-with-npm.md#using-javascript)
+- [a separate stylesheet](https://github.com/alphagov/govuk-frontend/blob/master/docs/installation/supporting-internet-explorer-8.md) if you support Internet Explorer 8
```
nickcolley: Internet Explorer 8 is stuff is important but not accessibility

README.md
```diff
@@ -33,72 +33,50 @@ We recommend [installing GOV.UK Frontend using node package manager
...
 
-[service-manual-browsers]: https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices#browsers-to-test-in
+GOV.UK Frontend supports the Government Digital Service’s recommended [browsers](https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices#browsers-to-test-in) and [assistive technologies](https://www.gov.uk/service-manual/technology/testing-with-assistive-technologies#what-to-test).
 
-## Assistive technology support
+GOV.UK Frontend also:
 
-GOV.UK Frontend will allow you to build services that comply with the [guidance
-in the Service Manual][service-manual-assistive-technologies].
+- is accessible for users who only use a keyboard
```
nickcolley: i think this keyboard section doesnt need explicitly calling out?

README.md
```diff
@@ -33,72 +33,50 @@ We recommend [installing GOV.UK Frontend using node package manager
 You can also [download the compiled and minified assets (CSS, JavaScript) from
 GitHub](docs/installation/installing-from-dist.md).
 
-## Browser support
+## Accessibility
 
-GOV.UK Frontend will allow you to build services that comply with the [guidance
-in the Service Manual][service-manual-browsers].
+The GOV.UK Design System team regularly tests GOV.UK Frontend with a range of users to make sure it’s as accessible as possible. You can read [the results of our most recent test for accessibility](https://design-system.service.gov.uk/accessibility/).
```
nickcolley: 'the results of our most recent test for accessibility' 

this wont always be an audit, I think we could give a more general statement that'd cover other things like internal reviews etc

README.md
```diff
@@ -33,72 +33,50 @@ We recommend [installing GOV.UK Frontend using node package manager
...
 
-Currently, GOV.UK Frontend officially supports the following browsers:
+However, your service will not automatically meet [level AA of WCAG 2.1](https://www.gov.uk/service-manual/helping-people-to-use-your-service/understanding-wcag) if you use GOV.UK Frontend. You must still do research to check that your service meets accessibility requirements, especially if you extend or modify components.
```
nickcolley: should we link to the service manual for what they need to do?

README.md
hannalaakso: This list doesn't currently tell you which component the warnings relate to (it's not clear for many of them), or the reason for why it happens.

Like Nick suggests we should probably link to these in issue tracker so that the context/reasoning is available.

README.md
```diff
@@ -33,72 +33,50 @@ We recommend [installing GOV.UK Frontend using node package manager
 You can also [download the compiled and minified assets (CSS, JavaScript) from
 GitHub](docs/installation/installing-from-dist.md).
 
-## Browser support
+## Accessibility
```
hannalaakso: I'm not entirely sure about getting rid of the "Browser support" heading as browser support doesn't only relate to accessibility but is wider than that 🤔

README.md
```diff
@@ -33,72 +33,50 @@ We recommend [installing GOV.UK Frontend using node package manager
...
-| macOS            | Mozilla Firefox (latest 2 versions)    | compliant   |
-| iOS              | Safari for iOS 9.3+                    | compliant   |
-| iOS              | Google Chrome (latest 2 versions)      | compliant   |
-| Android          | Google Chrome (latest 2 versions)      | compliant   |
-| Android          | Samsung Internet (latest 2 versions)   | compliant   |
+You should also use:
 
-‘Compliant’ means that the components must look as good and function as well as
-they do in other modern browsers.
+- [the JavaScript from GOV.UK Frontend](https://github.com/alphagov/govuk-frontend/blob/master/docs/installation/installing-with-npm.md#using-javascript)
```
hannalaakso: At the moment the reason we call out Javascript in this way but not CSS isn't clear. Maybe we could say that in many cases the JS contains "hidden" accessibility fixes eg. the button is like this, even though it might not seen like a component that comes with JS.

The component Javascript also needs to be initialised. Maybe that's too much detail but I suspect that step is often missed out.

README.md
m-green: Ah good point. How about "You can read about [our most recent accessibility tests](https://design-system.service.gov.uk/accessibility/)."?

README.md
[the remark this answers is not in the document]
m-green: Good plan - will add!

m-green: Good idea - will add a link to https://www.gov.uk/service-manual/helping-people-to-use-your-service/making-your-service-accessible-an-introduction if that sounds ok?

README.md
m-green: No problem - I can remove this.

README.md
m-green: Good point... it's a bit tricky because the theme of these sections overlaps.
Would it work if I moved the 'Browser and assistive technology support' section into its own section that come first?

So that would look like this:
## Browser and assistive technology support
## Accessibility
### Accessibility issues and warnings

README.md
m-green: See my comment here - https://github.com/alphagov/govuk-frontend/pull/1616#discussion_r337024332. Would that proposed heading structure work better?

README.md
m-green: The initialisation part is covered in the content in the link (https://github.com/alphagov/govuk-frontend/blob/master/docs/installation/installing-with-npm.md#using-javascript) - we could look at improving that content separately if it's currently being missed by users?

Could we also add content about the 'hidden' accessibility fixes on that page too? I'd like to keep all the usage/installation instructions in that page, rather than get into too much detail here.

README.md
```diff
@@ -33,72 +33,50 @@ We recommend [installing GOV.UK Frontend using node package manager
...
+### Accessibility issues and warnings
 
-| Software                 | Version        | Type               | Browser                   |
-|--------------------------|----------------|--------------------|---------------------------|
-| JAWS                     | 15 or later    | screen reader      | Internet Explorer 11      |
-| ZoomText                 | 10.11 or later | screen magnifier   | Internet Explorer 11      |
-| Dragon NaturallySpeaking | 11 or later    | speech recognition | Internet Explorer 11      |
-| NVDA                     | 2016 or later  | screen reader      | Firefox (latest versions) |
-| VoiceOver                | 7.0 or later   | screen reader      | Safari on iOS10 and OS X  |
+We're working on fixing the following issues from our [issues list in GitHub](https://github.com/alphagov/govuk-frontend/issues/).
```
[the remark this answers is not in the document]
m-green: I'm in two minds because it feels like it might be helpful for users to have a concise summary list. It'd be good to see what the issues list would look like - are you able to create the label so I can see what the user journey would be?

README.md
m-green: For this second list, it might be a bit odd to have separate GitHub issues for things that aren't issues we'll be working on fixing.

I feel that (unlike the issues-we're-working-on list) this summary list is important, because it helps us meet the user need (essentially "I need to know whether to worry about these warnings") without asking them to read the table / issues in detail. There's still the link if users want to know more, but hopefully some (most?) users can get by with this summary list. What do you think?

m-green: I'm not sure linking alone will work here for a couple of reasons - please see my comment here: https://github.com/alphagov/govuk-frontend/pull/1616#discussion_r337034374

README.md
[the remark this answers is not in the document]
hannalaakso: @m-green As an example, here's link for the the issues with the label "accessibility": https://github.com/alphagov/govuk-frontend/labels/accessibility

README.md
hannalaakso: I'm actually okay with having issues that we won't fix in Github for the record but we can have the `wont-fix` label on them. That way users can find them in Github by searching for them.

README.md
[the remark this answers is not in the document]
m-green: Ok - I think this could work (for this first list), if we can add the label Nick mentioned so that users see only issues that do not meet WCAG 2.1 AA. 
Perhaps we could also tweak the titles of the GitHub issues so they're as easy to possible to understand in the list alone?

README.md
hannalaakso: Thanks @m-green I've responded in #1616 (comment)

README.md
hannalaakso: That sounds good, thanks @m-green  👍 Non-blocking but I'd then prefer to see the link to the supported browsers under that main heading, rather than one level down under `### Accessibility`.

README.md
m-green: Ah ok, makes sense - thanks! I'd still like to keep this summary list here though - so we can help as many users as possible with their accessibility tests, without them needing to click and read the full explanation. They can still click through to the `wont-fix` issues list if they'd like the full background.

README.md
hannalaakso: Okay sounds good 👍

README.md
hannalaakso: If we're keeping the list in the README then we need to lift the component name from the "component name" column of the [table](https://github.com/alphagov/govuk-frontend/issues/1280#issuecomment-509588851).

Because for instance these warnings don't make sense on their own:
```
- A `group` role does not have a name.
- The `contentinfo` role is not necessary.
- The `banner` role is not necessary.
```
I also worry that since we've rephrased the warning message wording, users might not be able to find the error they've seen generated in the README as they'll probably be searching by the exact wording they've seen (that's what I'd do).

nickcolley: > For this second list, it might be a bit odd to have separate GitHub issues for things that aren't issues we'll be working on fixing.

Having closed issues with 'wont-fix' attached to them means they're recorded and visible to users (and importantly searchable from Google)

README.md
nickcolley: I feel like there is a more positive 'migration' framed way of introducing compatibility mode, this wording assumes the reader already knows about this when it's likely they will not?

README.md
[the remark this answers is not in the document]
nickcolley: Yeah I think it's really important that we focus the content where the workflow of the team is.

If we don't we'll not keep the README up-to-date, so even if it may be easier to scan the summary it'll be incorrect.

README.md
m-green: On your positive framing/migration point, it's best for users if we keep to GOV.UK style of just the concise facts - in this case, that your service won't meet AA if you use this mode. I appreciate plain words can feel a bit 'harsh' to us as we closely read and reread the words - but it doesn't sound that way to users, promise!

But I see what you mean on the familiarity with compatibility mode. How about we add "... to use GOV.UK Frontend with old frameworks or colours" at the end to make it clearer for new users what compatibility mode is?

README.md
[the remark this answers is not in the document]
m-green: Definitely see where you're coming from, but that approach can lead us to prioritise our needs (eg ease of updating or time to update) over the user need for plain language. We can always put content processes in place to ensure we keep guidance content in sync with changes.

All ok in this instance though! Are you or Hanna ok to split everything out and add the appropriate labels? Then I'll add the link here before we merge.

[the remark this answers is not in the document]
m-green: @hannalaakso mentioned at standup that she should be able to split out into separate issues - I'll update everything else we discussed in the meantime.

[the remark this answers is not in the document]
nickcolley: I've tried retroactively applying a 'accessibility-regulations-failure' label: 

https://github.com/alphagov/govuk-frontend/labels/accessibility-regulations-failure

As you can see there are no open issues, I may have missed some of the closed ones too.

README.md
nickcolley: > but it doesn't sound that way to users, promise!

I think it's fair to worry that our users could misunderstand this statement given what we know about people's attitude to migration.

I'm definitely not suggesting we write some flowery advertising copy, more to just ensure our users understand the feature will help them migrate to a more compliant framework, before we say something negative about it.

Through talking about this I think the bigger issue is just that we don't really have much to say about migration that's useful for our users, so it's always going to be a bit weird hearing about this for the first time in an accessibility statement.

So probably OK to leave as is and think about migration another time.

nickcolley:
For other reviewers, we're going to discuss the process to add issues to this statement in the future, but for now it's simplified as there are no open WCAG related issues.

README.md
```diff
@@ -33,72 +33,33 @@ We recommend [installing GOV.UK Frontend using node package manager
...
-[service-manual-browsers]: https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices#browsers-to-test-in
+You should also use:
 
-## Assistive technology support
+- [the JavaScript from GOV.UK Frontend](https://github.com/alphagov/govuk-frontend/blob/master/docs/installation/installing-with-npm.md#using-javascript)
+- [a separate stylesheet](https://github.com/alphagov/govuk-frontend/blob/master/docs/installation/supporting-internet-explorer-8.md) if you support Internet Explorer 8
 
-GOV.UK Frontend will allow you to build services that comply with the [guidance
-in the Service Manual][service-manual-assistive-technologies].
+Your service will not meet level AA of WCAG 2.1 if you use [compatibility mode](https://github.com/alphagov/govuk-frontend/blob/master/docs/installation/compatibility.md) to use GOV.UK Frontend with old frameworks or colours.
```
hannalaakso: I know "old colours" is lifted from the heading of the compatibility mode docs but "old colour palette" would be clearer to me (and we use it too in the docs).

README.md
```diff
@@ -33,72 +33,33 @@ We recommend [installing GOV.UK Frontend using node package manager
...
-https://support.mozilla.org/en-US/kb/change-fonts-and-colors-websites-use
-
-[enabling-high-contrast-mode-in-windows]:
-https://support.microsoft.com/en-gb/help/13862/windows-use-high-contrast-mode
-
-[high-contrast-plugin-for-chrome]: https://chrome.google.com/webstore/detail/high-contrast/djcfdncoelnlbldjfhinnjlhdjlikmph?hl=en-US
-
-[how-users-change-colours-on-websites]:
-https://accessibility.blog.gov.uk/2017/03/27/how-users-change-colours-on-websites/
+If your accessibility checker warns you about accessibility issues, you will not need to do anything if they're on our list of [issues that should not affect users](https://github.com/alphagov/govuk-frontend/issues/1280#issuecomment-509588851).
```
hannalaakso: "issues that should not affect users" feels a bit random here since the label for them in GOV.UK Frontend will most likely say something like "tooling-warnings". I think it would be good to establish a common term to refer to these issues if that makes sense. Open to discussing what that term might be but I think "issues that should not affect users" is probably too opaque.

README.md
m-green: Well spotted - looks like I got the style for this wrong on the compatibility page. Will fix!

README.md
hannalaakso: Thanks for the IRL @m-green! This looks great now 👍

Issue #1171, opened 2019-02-04
back-link is underlined when no href is provided


penx:
Consider the following HTML:

```html

      <div>
        <a href="#">Standard anchor with href</a>
      </div>
      <div>
        <a>Standard anchor with no href</a>
      </div>
      <div>
        <a class="govuk-back-link" href="#">
          Back link with href
        </a>
      </div>
      <div>
        <a class="govuk-back-link">Back link with no href</a>
      </div>
```
https://codesandbox.io/s/4wylrj5w

Browser default behaviour is to underline anchors only when an href is provided. This is to ensure [users can identify links on a page](https://webaim.org/techniques/hypertext/link_text#underlining) (an anchor without an href is not considered a link by default).

This feature also has the side effect of helping developers realise that they may have missed an href attribute - as they can visually see on the page that a link is not underlined.

The way the back-link is coded uses border-bottom instead of an underline, but this is not implemented as per normal anchor styles.

The old way browser default styles did this was as follows:

```css
a[href] {
border-bottom: none
}
```

In newer browsers, the `:any-link` pseudo-selector can be used, though it has limited support
https://developer.mozilla.org/en-US/docs/Web/CSS/:any-link

I would recommend adding `a[href]` or ``:any-link` to the back-link component and wherever else you may be overriding anchor underline styles, so that you mirror browser default behaviour for anchors, help users identify links and help developers identify errors.

PR #1620, opened 2019-10-25
Only add underline to back link when href exists

nickcolley:
"Browser default behaviour is to underline anchors only when an href is provided. This is to ensure users can identify links on a page (an anchor without an href is not considered a link by default).

This feature also has the side effect of helping developers realise that
they may have missed an href attribute - as they can visually see on the
page that a link is not underlined."

Fixes https://github.com/alphagov/govuk-frontend/issues/1171

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+### Fixes
+- [Pull request #1620: Only add underline to back link when href exists ](https://github.com/alphagov/govuk-frontend/pull/1620).
+
 ## 3.3.0 (Feature release)
 
 ### New features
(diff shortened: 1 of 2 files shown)


36degrees:
> The underline is also added if `href` is empty. Do we care about that?

Looking at [the spec](https://www.w3.org/TR/html52/textlevel-semantics.html#the-a-element), '[if] the `a` element has an `href` attribute, then it represents a hyperlink (a hypertext anchor)' – whether the `href` attribute is empty or not doesn't appear to be a consideration. I'd suggest we follow the spec here.

nickcolley:
@hannalaakso how do you feel about following the spec?

Issue #1567, opened 2019-09-12
Extending govuk-width-container mixin


Dawdre:
I am currently attempting to modify the `$govuk-page-width` variable. It is `960px !default`. 

**My goal is to have a BEM modifier on the `govuk-width-container`. Something like `govuk-width-container--wide`. I am conditionally applying this via our spring freemarker templating engine. Our desire is to have a wider content space for our large tables with lots of data. Things are becoming squished.**

I am looking for a way, whilst avoiding completely copy pasting the code inside your own `_width-container.scss` and creating a whole new "page-width" variable and mixin that does exactly the same thing apart from passing in a different base value.
Without being able to override the current "baked-in" mixin that compiles to `min-width: 1020px` it appears to be very tricky to do.

**Solution 1**
I have created my own "measurements" file in my project's directory where I am importing `settings/all` to gain access to your own `_measurements.scss` and also importing `objects/all` to gain access to the `govuk-width-container` mixin. I then reassign `$govuk-page-width`:  

This only partly works. The new value gets passed through to the `max-width` property, but it does not get passed into: 

`govuk-media-query($and: "(min-width: #{($govuk-page-width + $govuk-gutter * 2)})")` 

As suspected this breaks the expected behaviour. The content pushes up against the window due to the new `max-width` being wider than the original `1020px` breakpoint.

**Solution 2**
Setting the `$govuk-page-width` variable in my `main.scss`.
e.g.  
`$govuk-assets-path: "../../govuk-frontend/govuk/assets/";`
`$govuk-page-width: 1200px; `
`// import ALL GOVUK components`
`@import '../public/assets/govuk-frontend/govuk/all';`

But I feel this goes against modular Sass and doesn't feel maintainable. This also doesn't completely solve my problem as I have now set 1200px across my whole app. I only want it for my modifier class - `govuk-width-container--wide`.

I could just make a new variable and wrap it in some logic and only apply it when my modifier class is present but that will still not change the fact that if I **don't** reassign the base `$govuk-page-width: 960px !default` at some point I will always get the `1020px` breakpoint and it will break the expected behaviour. 

Are there any suggestions for extending or modifying your mixins and functions for this specifically and generally speaking?

Many thanks

nickcolley:
We're going to make it so you can use the `govuk-width-container` mixin and specify a width, this will be documented in Sass docs for now.

This'll also help us solve this issue too: https://github.com/alphagov/govuk-design-system/issues/798

Thanks for raising.

[...]

PR #1626, opened 2019-10-31
Allow creating custom width containers and using them with template

hannalaakso:
This PR:

1. Makes it possible for `govuk-width-container` mixin to accept a width. It defaults to `$govuk-page-width` so that `govuk-width-container` class will continue to function as before. The mixin defaults to pixels if no unit is specified.
2. Makes it possible for classes to be set on template container. This is necessary for setting custom width classes when needed, in addition to using `.govuk-width-container` by default. I called the new template option `containerClasses` following the naming in the header and footer which have `containerClasses`. These are "Classes that can be added to the inner container, useful if you want to make the header/footer full width."
3. Adds an example of a page with the template with custom width for template, header and footer containers. I’ve used BEM for the custom class name in `app.scss` as doing so feels clearer and in the spirit of what this is doing (creating something based on the default width class), although actually nothing is being inherited from `.govuk-width-container`. Note that there’s a difference in how the header and footer add the `govuk-width-container` class: footer hardcodes the class whereas header overrides it with `containerClasses`. As nothing is being inherited from `.govuk-width-container` I'm not passing it again unnecessarily to the header.

Tested on:
🍏 Chrome
🍏 FF
🍏 IE11
🍏 IE8

Fixes https://github.com/alphagov/govuk-frontend/issues/1567

Additionally, we'll be able to fix https://github.com/alphagov/govuk-design-system/issues/798 when this PR is released.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,34 @@
 
 ### New features
 
-#### Allow attributes to be set on `<body>` of template
+#### Create custom width container classes
+
+You can now create custom page width container classes using the `govuk-width-container` mixin. You do this by passing in the required maximum width of the container.
+
+For example:
+
+```scss
+.app-width-container--wide {
+  @include govuk-width-container(1200px);
+}
+```
+
+You can use the generated classes to set the width of:
+- template container
+- header container
+- footer container
+
+It was already possible to set the page app width with the `$govuk-page-width` variable. This new feature is useful when creating additional custom page width classes.
+
+- [Pull request #1626: Allow creating custom width containers and using them with template](https://github.com/alphagov/govuk-frontend/pull/1626).
+
+#### Set custom container classes on template
+
+You can now set classes on `.govuk-width-container` in the template with `containerClasses`. This is useful if you want to set a custom width class on the template container.
+
+- [Pull request #1626: Allow creating custom width containers and using them with template](https://github.com/alphagov/govuk-frontend/pull/1626).
+
+#### Set attributes on the `<body>` of template
 
 You can now set attributes in the `<body>` element of page template.
 
--- a/app/assets/scss/partials/_app.scss
+++ b/app/assets/scss/partials/_app.scss
@@ -2,6 +2,10 @@
   background-color: $govuk-body-background-colour;
 }
 
+.app-width-container--wide {
+  @include govuk-width-container(1200px);
+}
+
 .app-iframe-in-component-preview {
   margin: 15px 0;
 }
--- a/src/govuk/objects/_width-container.scss
+++ b/src/govuk/objects/_width-container.scss
@@ -2,9 +2,27 @@
 @import "../tools/all";
 @import "../helpers/all";
 
-@mixin govuk-width-container {
-  // Limit the width of the container to the page width
-  max-width: $govuk-page-width;
+////
+/// @group objects
+////
+
+/// Width container mixin
+///
+/// Used to create page width and custom width container classes.
+///
+/// @param {String} $width [$govuk-page-width] - Width in pixels
+///
+/// @example scss - Creating a 1200px wide container class
+///  .app-width-container--wide {
+///    @include govuk-width-container(1200px);
+///  }
+///
+/// @access public
+
+@mixin govuk-width-container($width: $govuk-page-width) {
+
+  // By default, limit the width of the container to the page width
+  max-width: $width;
 
   // On mobile, add half width gutters
   margin: 0 $govuk-gutter-half;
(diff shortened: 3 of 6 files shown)


app/views/examples/template-wide/index.njk
```diff
@@ -0,0 +1,32 @@
+{# To change the app page width use $govuk-page-width. To create additional custom page widths use the `govuk-width-container` mixin to create a class to do this, and set the class on the template, header and footer containers as shown below. #}
```
nickcolley: What're your thoughts on having a Sass test rather than an app page for this?

CHANGELOG.md
```diff
@@ -4,6 +4,14 @@
 
 ### New features
 
+#### Allow creating custom width containers and using them with template
+
+You can now create custom page width container class (this in in addition to already being able to modify the app page width with `$govuk-page-width` variable).
```
nickcolley: If we reference the mixin itself will it make clearer that we've introduced new behaviour to an existing mixin?

src/govuk/objects/_width-container.scss
```diff
@@ -2,9 +2,30 @@
...
+///
+/// @example scss - Creating a 1200px wide container class
+///  .govuk-width-container--wide {
+///    @include govuk-width-container(1200px);
+///  }
+///
+/// @access public
+
+@mixin govuk-width-container($width: $govuk-page-width) {
+  @if (unitless($width)) {
```
nickcolley: Is this necessary to handle? Should we expect our users to explicitly set their units like they would elsewhere in their CSS?

app/assets/scss/partials/_app.scss
```diff
@@ -2,6 +2,10 @@
   background-color: $govuk-body-background-colour;
 }
 
+.govuk-width-container--wide {
```
nickcolley: If we do go down this route the application specific class should use the namespace `app-` since this class is not in core GOV.UK Frontend.

https://design-system.service.gov.uk/get-started/extending-and-modifying-components/

src/govuk/template.njk
```diff
@@ -43,7 +43,7 @@
     {% endblock %}
 
     {% block main %}
-      <div class="govuk-width-container">
+      <div class="govuk-width-container {{ containerClasses }}">
```
nickcolley: Just a note that we'll need to add documentation to the Design System page template page.

CHANGELOG.md
nickcolley: Likewise with the containerClass feature should we be specific what the new feature and how you can use it?

app/views/examples/template-wide/index.njk
nickcolley: If you were keen on this idea we could still add a custom container class to the existing 'custom template' example, if you wanted somewhere to visually test this class.

src/govuk/template.njk
hannalaakso: The issues says to document this in just the SassDocs for now: https://github.com/alphagov/govuk-frontend/issues/1567#issuecomment-532679373 Maybe documenting it in the Design System will be part of the SassDocs work.

CHANGELOG.md
hannalaakso: Thanks @nickcolley. I've updated this section, have a look now.

src/govuk/template.njk
hannalaakso: Yeah that's a good point. Created an issue: https://github.com/alphagov/govuk-design-system/issues/1075

app/views/examples/template-wide/index.njk
hannalaakso: @nickcolley I've added a test for the output of `@mixin govuk-width-container`.

src/govuk/objects/objects.test.js
```diff
@@ -1,11 +1,58 @@
...
+const sassConfig = {
+  outputStyle: 'nested'
+}
+
 const sassFiles = glob.sync(`${configPaths.src}/objects/**/*.scss`)
 
-it.each(sassFiles)('%s renders to CSS without errors', (file) => {
-  return renderSass({ file: file })
+describe('The objects layer ', () => {
+  describe('@mixin govuk-width-container', () => {
```
36degrees: Ideally I think this should be in its own file – `width-container.test.js`. In the other layers, we only put tests that apply to the whole layer in the layer test file.

app/views/examples/template-wide/index.njk
hannalaakso: I've also moved the custom container classes to the custom template example as per your suggestion.

src/govuk/objects/objects.test.js
hannalaakso: Yes that's a good point, I've moved it 👍

CHANGELOG.md
```diff
@@ -4,7 +4,34 @@
 
 ### New features
 
-#### Allow attributes to be set on `<body>` of template
+#### Create custom width container classes
```
nickcolley: Very good work on the CHANGELOG, nice 👍

src/govuk/objects/_width-container.scss
nickcolley: I think we should do the minimum the solve the problem but I don't think this is a problem to keep around if you think it's useful.

hannalaakso: Thanks for the IRL 🙌 Decided to do the very minimum here so have removed the unit conversion.

PR #1629, opened 2019-11-05
Move #1609 from 3.3.0 to unreleased

36degrees:
This was accidentally added to the changelog for 3.3.0 after the release had been made – we've yet to release it, so it should be included in the unreleased section.


nickcolley:
My mistake not noticing this, thanks Ollie 👍

36degrees:
Easily done :)

PR #1637, opened 2019-11-12
Update with edited version of v3.4.0 (unreleased) changelog

m-green:
This is the edited version of the v3.4.0 release note in CHANGELOG.md.

Please review for:

- factual accuracy
- any missing changes in this release.

Following review and before publication we need to:

- copy to the GitHub release note

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,21 @@
 
 ### New features
 
-#### Create custom width container classes
+#### Add classes to the page wrapper
 
-You can now create custom page width container classes using the `govuk-width-container` mixin. You do this by passing in the required maximum width of the container.
+If you're using Nunjucks, you can now add classes to the page wrapper using the [`containerClasses` variable](https://design-system.service.gov.uk/styles/page-template/#variables).
+
+[Pull request #1626: Allow creating custom width containers and using them with template](https://github.com/alphagov/govuk-frontend/pull/1626).
+
+#### Set a custom width for wrappers
+
+If you're using Sass, you can now use the `govuk-width-container` mixin to create a custom wrapper class with a specific width. You can then add that class to the following wrappers to override the width of the `govuk-width-container` class:
+
+- [page wrapper](https://design-system.service.gov.uk/styles/layout/#page-wrappers)
+- [header container](https://design-system.service.gov.uk/components/header/)
+- [footer container](https://design-system.service.gov.uk/components/footer/)
+
+To create your custom wrapper class, include the `govuk-width-container` mixin. Pass in the width in pixels.
 
 For example:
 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -4,44 +4,56 @@
...
-You can now create custom page width container classes using the `govuk-width-container` mixin. You do this by passing in the required maximum width of the container.
+You can now:
+
+- create your own Sass classes with specific widths
+- add those classes to the `.govuk-width-container` wrapper using Nunjucks
+
+This means you can set specific widths for the:
+
+- [page wrapper](https://design-system.service.gov.uk/styles/layout/#page-wrappers)
+- [header component](https://design-system.service.gov.uk/components/header/)
```
nickcolley: It's the containers within the component not the component itself, you can apply this with the Nunjucks option 'containerClasses'

CHANGELOG.md
```diff
@@ -4,44 +4,56 @@
...
 ### New features
 
-#### Create custom width container classes
+#### Set a specific width for wrappers
 
-You can now create custom page width container classes using the `govuk-width-container` mixin. You do this by passing in the required maximum width of the container.
+You can now:
+
+- create your own Sass classes with specific widths
+- add those classes to the `.govuk-width-container` wrapper using Nunjucks
```
nickcolley: The ".govuk-width-container" class is shared by the page, header and footer component, so this there is no single 'width container', could consider removing this last bullet point and clarifying the use of 'containerClasses' in the template, header, and footer.

CHANGELOG.md
```diff
@@ -4,44 +4,56 @@
...
+
+This means you can set specific widths for the:
+
+- [page wrapper](https://design-system.service.gov.uk/styles/layout/#page-wrappers)
+- [header component](https://design-system.service.gov.uk/components/header/)
+- [footer component](https://design-system.service.gov.uk/components/footer/)
+
+Include the `govuk-width-container` mixin inside your Sass class definition, and pass in the width in pixels.
```
nickcolley: > Sass class definition

I've not heard 'Sass class definition' before, is this correct terminology for this code?

Could consider something simpler like

"Include the `govuk-width-container` mixin inside your code"

CHANGELOG.md
```diff
@@ -4,44 +4,56 @@
...
 
 For example:
 
 ```scss
-.app-width-container--wide {
+.YOURAPP-width-container--wide {
```
nickcolley: I think it's fine to keep this as 'app', not sure if it's required to explain namespaces here, since for most people 'app' will be fine.

CHANGELOG.md
m-green: I think the overall user task though is that you're setting the width for the wrapper / component? This bit's about what the user's aiming to achieve overall, rather than the details on exactly how to do it (which come further down).

CHANGELOG.md
m-green: Clarifying `containerClasses` might be a bit involved for the release note. Something to add to the relevant bit in the variables table on the Design System page maybe?
Could we make this bullet more accurate by saying "add those classes to your layout", or something similar? Then the next bit talks about the template, header and footer specifically.

CHANGELOG.md
m-green: Sounds good! Just for future reference - what would be the best term for one of these blocks in a Sass file?

CHANGELOG.md
nickcolley: There's a difference between a component's width and the container widths which are inside the component, if we don't think that's important that's fine but it may confuse people.

CHANGELOG.md
nickcolley: The two new additions are:

1. the govuk-width-container mixin now can have a different width if you pass a width to it
2. the page template can have additional classes applied to it using 'containerClasses', which is consistent with the header and footer component

CHANGELOG.md
m-green: @hannalaakso and I discussed this quite a bit - there's a big risk of confusion here (I've tripped up on it myself!) because there's no clear distinction between placeholder names and copy-and-paste code. It's especially problematic here because you have an `app` folder in Frontend. So it's worth clarifying - and it won't hold up users who already know about namespaces.

nickcolley: >  big risk of confusion

There's no risk that I know of people copying 'app-width-container--wide' into their codebase, and this is what we expect a large majority of people to do, there's only a small part of our user base that require something different to 'app'.

nickcolley: This mirrors the precedent set in the https://design-system.service.gov.uk/get-started/extending-and-modifying-components/ guidance where we use 'app-' classes

m-green: No probs, but if we use `app` I'd like to be specific in https://github.com/alphagov/govuk-frontend/pull/1637#discussion_r345129313 by naming the element of the Sass file.

m-green:
@nickcolley @36degrees We've rejigged this a bit following @nickcolley's comments - please take a look!

CHANGELOG.md
```diff
@@ -16,32 +28,28 @@ For example:
...
-It was already possible to set the page app width with the `$govuk-page-width` variable. This new feature is useful when creating additional custom page width classes.
+- `containerClasses` in the [page template](https://design-system.service.gov.uk/styles/page-template/#variables)
+- `containerClasses` option in the [header](https://design-system.service.gov.uk/components/header/) or [footer](https://design-system.service.gov.uk/components/footer/)
 
-- [Pull request #1626: Allow creating custom width containers and using them with template](https://github.com/alphagov/govuk-frontend/pull/1626).
-
-#### Set custom container classes on template
-
-You can now set classes on `.govuk-width-container` in the template with `containerClasses`. This is useful if you want to set a custom width class on the template container.
+Use the `$govuk-page-width` Sass variable instead if all your pages are the same width.
```
36degrees: I wonder if this is clear enough that this is _instead_ of doing all of the above? And do we think we need to front load this information?

CHANGELOG.md
```diff
@@ -4,9 +4,21 @@
 
 ### New features
 
-#### Create custom width container classes
+#### Add classes to the page wrapper
 
-You can now create custom page width container classes using the `govuk-width-container` mixin. You do this by passing in the required maximum width of the container.
+If you're using Nunjucks, you can now add classes to the page wrapper using the [`containerClasses` variable](https://design-system.service.gov.uk/styles/page-template/#variables).
+
+Pull request #1626: Allow creating custom width containers and using them with template.
```
nickcolley: Non-blocking, should this be a link? No worries if not.

CHANGELOG.md
m-green: It's a good point and @hannalaakso and I discussed it - it probably *does* belong further up, but moving it up might make it a bit confusing because it will get in the way of the main guidance about the custom classes. So we think it's best to stick with this, but keep an eye on it and change it if any users get confused.

PR #1638, opened 2019-11-12
Check component item arrays are not empty before outputting markup

natcarey:
Overview
===

When using the Nunjucks components there are a number of optional arrays, there's one behaviour when you provide something and another when you provide nothing.  At HMRC we've found some confusion over the difference between providing nothing and providing an empty array.

This change standardises that behaviour so that an empty array is treated as nothing being provided.

Scope
---

This only affects the Nunjucks interface in its handling of empty arrays, users who aren't providing empty arrays will not be affected by this change.

There are no CSS or JS changes

Tests
---

We have added tests for the new behaviour.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Fixes
 
 - [Pull request #1655: Ensure components use public `govuk-media-query` mixin](https://github.com/alphagov/govuk-frontend/pull/1655).
+- [Pull request #1638: Check component item arrays are not empty before outputting markup](https://github.com/alphagov/govuk-frontend/pull/1638).
 
 ## 3.4.0 (Feature release)
 
@@ -57,7 +58,6 @@ You can now add attributes to the `<body>` element of a page, by using the [`bod
 - [Pull request #1609: Update hex value for secondary text to improve contrast](https://github.com/alphagov/govuk-frontend/pull/1609).
 - [Pull request #1620: Only add underline to back link when href exists ](https://github.com/alphagov/govuk-frontend/pull/1620).
 - [Pull request #1631: Fix classes on character count when in error state](https://github.com/alphagov/govuk-frontend/pull/1631).
-
 
 ## 3.3.0 (Feature release)
 
--- a/src/govuk/components/date-input/template.njk
+++ b/src/govuk/components/date-input/template.njk
@@ -7,7 +7,7 @@
    aria-describedby – for example hints or error messages -#}
 {% set describedBy = params.fieldset.describedBy if params.fieldset.describedBy else "" %}
 
-{% if params.items %}
+{% if params.items | length %}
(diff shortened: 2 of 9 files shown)


nickcolley:
Spoke to Nat and they've said that they're going to be away for a week but @silasl (hi 👋 ) can help out in the time.

36degrees:
It'd be good to remove the change to package-lock.json, which I believe is unrelated and just down to different versions of npm being used.

(I've not had a chance to review the rest of the PR yet)

36degrees:
Thanks @natcarey!

nickcolley:
We're going to release 3.4.0 then do a small patch release with this once you've had time to look at this :)

natcarey:
I've updated with that feedback and I'm working as usual this week.

Thanks all.

natcarey:
I just spotted this comment "we are wondering if there are more places we need to do this: for instance accordion, breadcrumbs, select also have an array of items", I haven't addressed that yet.  I'll look at that this morning.

natcarey:
@nickcolley said:
> we are wondering if there are more places we need to do this: for instance accordion, breadcrumbs, select also have an array of items

I don't think these are affected by the same problem.

These examples are going to behave strangely when an empty array or no array is provided but the behaviour of an empty array matches the behaviour of no array.  In the components I've updated there is a difference between the handling of "nothing" and an empty array.

Adding conditions around the template would help avoid a nonsense output but it isn't needed for consistency between these two behaviours.  For example `govukAccordion` has a `for` loop which iterates the items and checks each one is not empty; If there's an empty array or no array the outer div (`.govuk-accordion`) is still output but it will be empty.  A condition of `{% if params.items | length %}` would help with this but it changes the behaviour for both an empty array and no array.

CHANGELOG.md
```diff
@@ -49,6 +49,7 @@ You can now add attributes to the `<body>` element of a page, by using the [`bod
 - [Pull request #1609: Update hex value for secondary text to improve contrast](https://github.com/alphagov/govuk-frontend/pull/1609).
 - [Pull request #1620: Only add underline to back link when href exists ](https://github.com/alphagov/govuk-frontend/pull/1620).
 - [Pull request #1631: Fix classes on character count when in error state](https://github.com/alphagov/govuk-frontend/pull/1631).
+- [Pull request #1638: Check component item arrays are not empty before outputting markup](https://github.com/alphagov/govuk-frontend/pull/1638).
```
nickcolley: Something I missed, this changelog entry needs to be in an 'unreleased section' instead of being in the previous release

nickcolley: Yeah that looks great but 'New features' should be 'Fixes'

36degrees: Sorry, this probably needs updating again as v3.4.0 has now been released.

PR #1646, opened 2019-11-14
Add custom classes to character count

LBHELewis:
`params.classes` isn't currently being added to the character count component, this PR fixes that.

### DIFF
--- a/src/govuk/components/character-count/template.njk
+++ b/src/govuk/components/character-count/template.njk
@@ -1,6 +1,6 @@
 {% from "../textarea/macro.njk" import govukTextarea %}
 
-<div class="govuk-character-count" data-module="govuk-character-count"
+<div class="govuk-character-count {%- if params.classes %} {{ params.classes }}{% endif -%}" data-module="govuk-character-count"
 {%- if params.maxlength %} data-maxlength="{{ params.maxlength }}"{% endif %}
 {%- if params.threshold %} data-threshold="{{ params.threshold }}"{% endif %}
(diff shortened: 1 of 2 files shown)


src/govuk/components/character-count/template.njk
```diff
@@ -1,6 +1,6 @@
 {% from "../textarea/macro.njk" import govukTextarea %}
 
-<div class="govuk-character-count" data-module="govuk-character-count"
+<div class="govuk-character-count {%- if params.classes %} {{ params.classes }}{% endif -%}" data-module="govuk-character-count"
```
nickcolley: I believe that the `classes` are used by the nested textarea already, which is unfortunate as it breaks the convention for components, perhaps it should have been `textareaClasses`.

Do you need to add custom classes to this or will the current functionality of classes on the textarea element work for your use case?

LBHELewis: Oh I see, sorry - I missed that it was being used elsewhere. It's a bit of a problem for us. We can get around it by directly styling `.govuk-character-count`, but until now we've tried to avoid styling any `govuk` classes, instead using our own. Shall I close this PR and add an issue instead?

nickcolley: Yes please, could you give detail on why you need to style the wrapper, it'll help us figure out what the right thing to do is.

In the meantime you could consider duplicating the template to allow for additional classes, to avoid styling `govuk-` classes.

nickcolley:
Closing this as we'll discuss next steps here: https://github.com/alphagov/govuk-frontend/issues/1649

PR #1648, opened 2019-11-18
Update checkboxes and radio buttons to include item hint classes on i…

LBHELewis:
…tem hint

We need to be able to add a custom class to all hints. Currently the checkbox and radio item loop outputs most hint properties, but not the classes, instead using only `govuk-checkboxes__hint` or `govuk-radios__hint`. This PR adds the hint.classes value to this so that we can add custom classes.

### DIFF
--- a/src/govuk/components/checkboxes/__snapshots__/template.test.js.snap
+++ b/src/govuk/components/checkboxes/__snapshots__/template.test.js.snap
@@ -47,6 +47,17 @@ exports[`Checkboxes when they include a hint renders the hint 1`] = `
 >
   If you have dual nationality, select all options that are relevant to you.
 </span>
+<span id="example-item-hint"
+      class="govuk-hint govuk-checkboxes__hint"
+>
+  Hint for british option here
+</span>
+<span id="example-3-item-hint"
+      class="govuk-hint govuk-checkboxes__hint app-checkboxes__hint-other"
+      data-test-attribute="true"
+>
+  Hint for other option here
+</span>
 `;
 
 exports[`Checkboxes when they include an error message renders the error message 1`] = `
--- a/src/govuk/components/checkboxes/checkboxes.yaml
+++ b/src/govuk/components/checkboxes/checkboxes.yaml
@@ -261,29 +261,6 @@ examples:
       - value: other
         text: Citizen of another country
 
-- name: with all fieldset attributes
-  data:
-    idPrefix: example
-    name: example
-    fieldset:
-      classes: app-fieldset--custom-modifier
-      attributes:
-        data-attribute: value
-        data-second-attribute: second-value
-      legend:
-        text: What is your nationality?
-    hint:
-      text: If you have dual nationality, select all options that are relevant to you.
-    errorMessage:
-      text: Please select an option
-    items:
-      - value: british
-        text: British
-      - value: irish
-        text: Irish
-      - value: other
-        text: Citizen of another country
-
 - name: with error message
   data:
     name: waste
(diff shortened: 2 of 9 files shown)


nickcolley:
Thanks for spotting this and helping us fix it!

Can confirm that in our documentation we indicate you can add additional classes to the hints: https://design-system.service.gov.uk/components/radios/#options-inline-radios-example--hint

I don't think we add tests for components within components as we expect to be able to pass through all options but perhaps we should change this given we missed this option, will see what other reviewers thing.

36degrees:
The changelog entry in this pull request will need updating as v3.4.0 has now been released.

nickcolley:
Hey again :)

I think we can get this merged if we update a test fixture and update the jest snapshots.

When we have nested components we render them and take a snapshot, you can see this with the test named 'renders the hint', in `template.test.js`.

This test should represent all the bits you could pass through to a component, just so we know it's doing what we expected.

So to test this functionality we can add hints to the `with all fieldset attributes` example in `radios.yaml` and then run the tests updating the snapshot:

`node_modules/.bin/jest src/components/radios/template.test.js --updateSnapshot`

You'll need to do the same for checkboxes component.

Let me know if you run into any problems.

src/govuk/components/checkboxes/template.njk
```diff
@@ -88,7 +88,7 @@
           {% if hasHint %}
           {{ govukHint({
             id: itemHintId,
-            classes: 'govuk-checkboxes__hint',
+            classes: 'govuk-checkboxes__hint' + (' ' + item.hint.classes if item.hint.classes),
```
nickcolley: Non-blocking I wonder if we could avoid the extra whitespace by doing something like

`'govuk-checkboxes__hint' + ((' ' + item.hint.classes) if item.hint.classes)`

Hopefully this'd output the space only when there are classes.

nickcolley: This doesnt add whitespace after all so works well.

src/govuk/components/checkboxes/checkboxes.yaml
```diff
@@ -279,10 +279,21 @@ examples:
     items:
       - value: british
         text: British
+        hint:
```
36degrees: I think this would be better served through adding new tests, as this fixture is meant to test checkboxes with 'all fieldset attributes' and these changes don't really fit that description.

nickcolley: I've refactored the tests so the fixtures are specific for each test.

src/govuk/components/checkboxes/template.test.js
```diff
@@ -777,7 +854,25 @@ describe('Checkboxes', () => {
     })
 
     it('passes through fieldset params without breaking', () => {
-      const $ = render('checkboxes', examples['with all fieldset attributes'])
+      const $ = render('checkboxes', {
```
hannalaakso: "with all fieldset attributes" specified `idPrefix` but didn't test for it as its value in the example was the same as `name` - so you couldn't really tell if it was working or falling back to `name`. 

It would be nice to have a test for `idPrefix` too, what do you think @nickcolley? Or we can do it later.

src/govuk/components/checkboxes/__snapshots__/template.test.js.snap
```diff
@@ -47,6 +47,17 @@ exports[`Checkboxes when they include a hint renders the hint 1`] = `
 >
   If you have dual nationality, select all options that are relevant to you.
 </span>
+<span id="example-item-hint"
+      class="govuk-hint govuk-checkboxes__hint"
+>
+  Hint for british option here
```
hannalaakso: Nit: capitalise "british".

src/govuk/components/checkboxes/template.test.js
nickcolley: As far as I know idPrefix is only needed for items, which is why the snapshot has not changed, so I think that'd be something to do later.

Issue #1645, opened 2019-11-14
Make character count use hint component


LBHELewis:
Currently the character count component contains a hint with markup (`<span class="govuk-hint">`) rather than using the hint component. Could this use the hint component instead? Happy to put in a PR myself, but wasn't sure whether this was a decision rather than a bug, because of the javascript taking place in this component?

nickcolley:
I can't be certain it seems like it could be using the hint component macro and still output everything it needs to, if you wanted to raise a pull request to change this that sounds good!

LBHELewis:
I've added a PR here: https://github.com/alphagov/govuk-frontend/pull/1650

36degrees:
Closed in #1650

PR #1650, opened 2019-11-18
Make Character Count use hint component for message and allow custom …

LBHELewis:
…classes to be added

Character count was using the hint markup but not the component for its count message. This PR changes that to use the hint component. It doesn't allow custom text or id as these are referred to in the javascript, but it does allow custom classes to be added for styling.

Relevant issue: https://github.com/alphagov/govuk-frontend/issues/1645

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@
 
 ### New features
 
+#### Add classes to the character count component's count message
+
+If you're using Nunjucks, you can now add classes to the character count component's count message using the `countMessage.classes` option.
+
+- [Pull request #1650: Make Character Count use hint component for message and allow custom classes to be added](https://github.com/alphagov/govuk-frontend/pull/1650)
+
 ### Fixes
 
 - [Pull request #1655: Ensure components use public `govuk-media-query` mixin](https://github.com/alphagov/govuk-frontend/pull/1655).
--- a/src/govuk/components/character-count/character-count.yaml
+++ b/src/govuk/components/character-count/character-count.yaml
@@ -59,6 +59,10 @@ params:
   type: object
   required: false
   description: HTML attributes (for example data attributes) to add to the textarea.
+- name: countMessageClasses
+  type: string
+  required: false
+  description: Classes to add to the count message.
 
 
 examples:
--- a/src/govuk/components/character-count/character-count.yaml
+++ b/src/govuk/components/character-count/character-count.yaml
@@ -59,11 +59,15 @@ params:
   type: object
   required: false
   description: HTML attributes (for example data attributes) to add to the textarea.
-- name: countMessageClasses
-  type: string
+- name: countMessage
+  type: object
   required: false
-  description: Classes to add to the count message.
-
+  description: Options for the count message
+  params:
+  - name: classes
+    type: string
+    required: false
+    description: Classes to add to the count message
 
 examples:
   - name: default
--- a/src/govuk/components/character-count/template.njk
+++ b/src/govuk/components/character-count/template.njk
@@ -28,7 +28,7 @@
   {{ govukHint({
     text: 'You can enter up to ' + (params.maxlength or params.maxwords) + (' words' if params.maxwords else ' characters'),
     id: params.id + '-info',
-    classes: 'govuk-character-count__message' + (' ' + params.countMessageClasses if params.countMessageClasses),
+    classes: 'govuk-character-count__message' + (' ' + params.countMessage.classes if params.countMessage.classes),
     attributes: {
       'aria-live': 'polite'
     }
--- a/src/govuk/components/character-count/template.njk
+++ b/src/govuk/components/character-count/template.njk
@@ -1,4 +1,5 @@
 {% from "../textarea/macro.njk" import govukTextarea %}
+{% from "../hint/macro.njk" import govukHint %}
 
 <div class="govuk-character-count" data-module="govuk-character-count"
 {%- if params.maxlength %} data-maxlength="{{ params.maxlength }}"{% endif %}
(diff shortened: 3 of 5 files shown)


36degrees:
The changelog entry in this pull request will need updating as v3.4.0 has now been released.

LBHELewis:
@36degrees Thank you - that's updated now.

CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
...
 
+## Unreleased
+
+### New features
+
+#### Add custom classes to the count message in the character count component
+
+You can now add custom classes to the count message in the character count component, which has been changed to use the hint component rather than custom markup.
+
+- [Pull request #:1650: Make Character Count use hint component for message and allow custom classes to be added](https://github.com/alphagov/govuk-frontend/pull/1650)
```
nickcolley: Tiny style thing: "Pull request #:1650" -> "Pull request: #1650"

CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
 # Changelog
 
+## Unreleased
+
+### New features
+
+#### Add custom classes to the count message in the character count component
+
+You can now add custom classes to the count message in the character count component, which has been changed to use the hint component rather than custom markup.
```
nickcolley: When writing CHANGELOG notes we try to focus on what the user needs to know, so how we fixed or updating something is not required to understand the new feature.

So this could instead focus on how to use the new `countMessageClasses` option rather than the internal details on how we implemented this change.

src/govuk/components/character-count/template.njk
```diff
@@ -24,7 +25,14 @@
     errorMessage: params.errorMessage,
     attributes: params.attributes
   }) }}
-  <span id="{{ params.id }}-info" class="govuk-hint govuk-character-count__message" aria-live="polite">
-    You can enter up to {{ params.maxlength or params.maxwords }} {{'words' if params.maxwords else 'characters' }}
-  </span>
+  {{
```
nickcolley: Tiny style thing: We keep the macro on the same line as the curly

```javascript
{{ govukHint({
  ...
}) }}
```

vs

```javascript
{{
  govukHint({
    ...
  })
}}
```

src/govuk/components/character-count/character-count.yaml
```diff
@@ -141,3 +145,12 @@ examples:
       threshold: 75
       label:
         text: Full address
+
+  - name: with count message class
```
nickcolley: Where possible we try to add tests rather than review application demonstrations for non-visible functionality. So we remove this example and test this with Jest rather than having to check the review application.

src/govuk/components/character-count/character-count.yaml
```diff
@@ -59,6 +59,10 @@ params:
   type: object
   required: false
   description: HTML attributes (for example data attributes) to add to the textarea.
+- name: countMessageClasses
```
nickcolley: Should this be

`countMessage` or `message` with a nested `classes` option?

Could allow for updates without breaking changes in the future.


```javascript
{{ govukCharacterCount({
  label: {
    text: "Can you provide more detail?"
  },
  countMessageClasses: 'app-extra-class'
}) }}
```
vs

```javascript
{{ govukCharacterCount({
  label: {
    text: "Can you provide more detail?"
  },
  countMessage: {
    classes: 'app-extra-class'
  }
}) }}
```

src/govuk/components/character-count/template.test.js
```diff
@@ -99,9 +99,11 @@ describe('Character count', () => {
     it('renders with hint', () => {
       const $ = render('character-count', {
         id: 'character-count-with-hint',
+        maxlength: 10,
         hint: {
           text: 'It’s on your National Insurance card, benefit letter, payslip or P60. For example, ‘QQ 12 34 56 C’.'
-        }
+        },
+        countMessageClasses: 'app-extra-class'
```
36degrees: I think this should probably be tested in a separate assertion – this test is looking at whether a separate actual 'hint' can be added to the character count ("'It’s on your National Insurance card, benefit letter, payslip or P60. For example, ‘QQ 12 34 56 C’.'") whereas this class is added to the count message, which is unrelated to the rest of the test. I'd also assert that the class is on the element, rather than using a snapshot for it.

src/govuk/components/character-count/template.njk
```diff
@@ -24,7 +25,12 @@
...
   }) }}
-  <span id="{{ params.id }}-info" class="govuk-hint govuk-character-count__message" aria-live="polite">
-    You can enter up to {{ params.maxlength or params.maxwords }} {{'words' if params.maxwords else 'characters' }}
-  </span>
+  {{ govukHint({
+    text: 'You can enter up to ' + (params.maxlength or params.maxwords) + (' words' if params.maxwords else ' characters'),
+    id: params.id + '-info',
+    classes: 'govuk-character-count__message' + (' ' + params.countMessageClasses if params.countMessageClasses),
+    attributes: {
+      'aria-live': 'polite'
```
36degrees: I know this isn't 'new', but we're changing the implementation and it doesn't look like there's a test checking that this attribute is set correctly. Given how important this is for assistive tech, could we add one?

src/govuk/components/character-count/character-count.yaml
36degrees: I think I'd tend towards the nested object, as it's likely we'll need to make the text customisable in the future (somehow) so that it can be overridden if you're using a different language.

src/govuk/components/character-count/template.njk
```diff
@@ -24,7 +25,12 @@
     errorMessage: params.errorMessage,
     attributes: params.attributes
   }) }}
-  <span id="{{ params.id }}-info" class="govuk-hint govuk-character-count__message" aria-live="polite">
-    You can enter up to {{ params.maxlength or params.maxwords }} {{'words' if params.maxwords else 'characters' }}
-  </span>
+  {{ govukHint({
```
36degrees: I am still a little unsure about this change – it generally feels a bit strange to me that we're using the hint component at all here (as opposed to just having a 'character count message' which happens to look a bit like a hint…) – however, I don't think this makes it any harder for us to change this later, so I think it's OK for now.

nickcolley:
@36degrees Updated the tests so there is coverage for the count message functionality and include new functionality.

PR #1661, opened 2019-11-22
Ensure all *.scss files are linted

colinrotherham:
Morning 👋 

I've been working on a different pull request to this one, but I've split this out:

1. My code editor (and others) don't _see_ the **.sass-lint.yaml** config
Why? In this project it has a custom path `./config/.sass-lint.yml`

2. Only `./src` _*.scss_ files get run through **sass-lint**, not `./app`
I've updated the paths and fixed the linting issues.

There was a "magic hex" value used by the iframe component previews:

```css
.app-iframe-in-component-preview {
  box-shadow: 0 0 0 5px #e4f2ff;
}
```

Which I've moved to a variable, it's not quite a variation on a palette colour.

### DIFF
--- a/app/assets/scss/partials/_app.scss
+++ b/app/assets/scss/partials/_app.scss
@@ -6,13 +6,12 @@
   @include govuk-width-container(1200px);
 }
 
+// Injected into the preview page, that is then put in an iframe
 .app-iframe-in-component-preview {
   margin: 15px 0;
 }
-/* Injected into the preview page, that is then put in an iframe */
-.app-iframe-in-component-preview {}
 
-/* Removes the breakpoint sass-mq debug display */
+// Removes the breakpoint sass-mq debug display
 .app-iframe-in-component-preview:before {
   display: none;
 }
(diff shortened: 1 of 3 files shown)


app/assets/scss/partials/_app.scss
```diff
@@ -22,34 +21,35 @@
 }
 
 .app-component-preview {
-  position: relative;
```
nickcolley: Can you explain the reasoning behind removing this property please?

package.json
```diff
@@ -93,6 +93,7 @@
     "ie 8-11",
     "iOS >= 9"
   ],
+  "sasslintConfig": "config/.sass-lint.yml",
```
nickcolley: This is a great change thanks!

app/assets/scss/partials/_app.scss
colinrotherham: I wasn't supposed to commit this until a further merge request, but the immediate child `.app-component-preview__iframe` already has `position: relative`.

It doesn't appear necessary to have it on both.

I've put it back, it's not the right commit to remove lines—only lint them.

Issue #1662, opened 2019-11-22
Problem with fsevents@1.2.7 and Node.js 12


colinrotherham:
Since the Node.js 12 update I've been unable to install GOV.UK Frontend via `npm ci` (clean install from lockfile) without errors:

```
node-pre-gyp WARN Using request for node-pre-gyp https download
node-pre-gyp WARN Tried to download(404): https://fsevents-binaries.s3-us-west-2.amazonaws.com/v1.2.7/fse-v1.2.7-node-v72-darwin-x64.tar.gz
node-pre-gyp WARN Pre-built binaries not found for fsevents@1.2.7 and node@12.13.1 (node-v72 ABI, unknown) (falling back to source compile with node-gyp)
node-pre-gyp WARN Pre-built binaries not installable for fsevents@1.2.7 and node@12.13.1 (node-v72 ABI, unknown) (falling back to source compile with node-gyp)
```

```
npm WARN deprecated fsevents@1.2.9: One of your dependencies needs to upgrade to fsevents v2: 1) Proper nodejs v10+ support 2) No more fetching binaries from AWS, smaller package size
```

No pre-built binary exists for **fsevents@1.2.7** on Node.js 12
https://fsevents-binaries.s3-us-west-2.amazonaws.com/v1.2.7/fse-v1.2.7-node-v72-darwin-x64.tar.gz

The current lockfile in the project requires **fsevents@1.2.7** for Gulp, Jest and Nodemon.

Version 1.2.9+ is required for Node.js 12 compatibility.

colinrotherham:
With minor updates I managed to get it to install.

But now I'm seeing high CPU usage from `gulp nodemon`. I think we need to go through the outdated packages and ensure we have Node.js 12 support for all.

For example **nodemon@2.0.1** adds chokidar v3 for file watching (including Node.js 12 fixes via **fsevents@~2.1.1**) but this hasn't been applied yet.

![Nodemon CPU usage](https://user-images.githubusercontent.com/415517/69446116-a333a300-0d4b-11ea-8698-f732ebd1a54c.png)

colinrotherham:
Slightly related, other older dependencies still want Node.js 4:

```
npm WARN notsup Unsupported engine for to-markdown@2.1.0: wanted: {"node":"^4"} (current: {"node":"12.13.1","npm":"6.12.1"})
npm WARN notsup Not compatible with your version of node/npm: to-markdown@2.1.0
```

[...]

PR #1663, opened 2019-11-22
Fix problem with fsevents@1.2.7 and Node.js 12

colinrotherham:
Updates:

1. All non-breaking child dependencies.
2. Some carefully-checked breaking changes

This pull fixes `node-gyp` build issues in https://github.com/alphagov/govuk-frontend/issues/1662

I've also pinned `postcss-pseudo-classes@0.2.0` as the **0.2.1** update throws an “Expected closing parenthesis” error due to the current blacklist for the `:not()` selector hack.

**Breaking but not breaking changes**
Also included because they're major semver bumps, but aren't breaking for GOV.UK Frontend:

>**acorn@^7.1.0**
>Raises default ES2018 to ES2019, supports dynamic `import()` from Node.js 10
>
>**gulp-if@^3.0.0**
>Now only supports Node.js 8+
>
>**jest-serializer-html@^7.0.0**
>Untrimmed whitespace now matters in matched HTML
>
>**nodemon@^2.0.1**
>Now only supports Node.js 8+ (including chokidar fixes for Node.js 12)
>
>**vinyl-paths@^3.0.1**
>Now only supports Node.js 8+
>
>**yargs@^15.0.2**
>Now only supports Node.js 8+
>Throws errors for invalid arguments config


colinrotherham:
No problem!

Also held back `puppeteer@2.0.0` as we're waiting on https://github.com/smooth-code/jest-puppeteer/pull/289 too

tasks/gulp/watch.js
```diff
@@ -5,7 +5,7 @@ const configPaths = require('../../config/paths.json')
 // Watch task ----------------------------
 // When a file is changed, re-run the build task.
 // ---------------------------------------
-gulp.task('watch', () => {
```
nickcolley: I don't believe this change is necessary as it was a problem with my computer, sorry!

colinrotherham: If an error does happen in `watch` in future you'll want the promise to reject and be fed back to gulp so it can handle it correctly. Sure you don't want to leave it in?

nickcolley: Sounds useful to me so can leave it in sure.

nickcolley:
Thanks for tidying this all up, very good improvements.

PR #1238, opened 2019-03-05
Proposal programmatic and data attribute api interface for components.

nickcolley:
This is a scrappy branch to explore the practical impact of adding a data api alongside an programmatic api.

I've opened this draft so I can share code examples and review application links in the official proposal.

Inspired by [Bootstraps data API](https://getbootstrap.com/docs/3.4/javascript/#js-data-attrs)

This will go into an fleshed out architecture proposal and aims to solve issues https://github.com/alphagov/govuk-frontend/pull/1215 in a more long term way.

PR #1668, opened 2019-11-27
Add new Modal Dialogue component

colinrotherham:
Afternoon 👋 

As discussed with @hannalaakso, @timpaul, @amyhupe our team has been working on contributing a new "Modal dialogue" component to the Design System.

Sadly I'm going to be handing over work on this to @FeyAgape and @Ash-Wilson so this is still a draft for now, but more work to come!

## Acceptance criteria

See https://github.com/alphagov/govuk-design-system-backlog/issues/30#issuecomment-553932293

## Changes to GOV.UK Frontend
To support a modal component I've had to enhance GOV.UK Frontend slightly:

* Adjusted iframe previews for `position: fixed` content
* Adjusted iframe previews for custom variant styling (fixed height for modals)
* Add support for Nunjucks macro call blocks `{% call macro() %}{% endcall %}`

But here's a draft pull request where we're up to anyway.

Modal preview

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -162,12 +162,23 @@ module.exports = (options) => {
 
     // Construct and evaluate the component with the data for this example
     const macroName = helperFunctions.componentNameToMacroName(componentName)
-    const macroParameters = JSON.stringify(exampleConfig.data, null, '\t')
+    const macroContext = { data: exampleConfig.data, caller: exampleConfig.caller }
 
-    res.locals.componentView = env.renderString(
-      `{% from '${componentName}/macro.njk' import ${macroName} %}
-      {{ ${macroName}(${macroParameters}) }}`
-    )
+    // Build macro view
+    let macroString = `{% from '${componentName}/macro.njk' import ${macroName} %}`
+
+    // Macro is a caller
+    if (typeof exampleConfig.caller === 'object') {
+      macroString += `
+        {% call ${macroName}(data) %}
+          {{ caller.html | safe if caller.html else caller.text }}
+        {% endcall %}
+      `
+    } else {
+      macroString += `{{ ${macroName}(data) }}`
+    }
+
+    res.locals.componentView = env.renderString(macroString, macroContext)
 
     // Optional preview layout
     let previewLayout = res.locals.componentData.previewLayout
--- a/app/app.js
+++ b/app/app.js
@@ -152,8 +152,6 @@ module.exports = (options) => {
     const componentName = req.params.component
     const requestedExampleName = req.params.example || 'default'
 
-    const previewLayout = res.locals.componentData.previewLayout || 'layout'
-
     const exampleConfig = res.locals.componentData.examples.find(
       example => example.name.replace(/ /g, '-') === requestedExampleName
     )
@@ -171,9 +169,19 @@ module.exports = (options) => {
       {{ ${macroName}(${macroParameters}) }}`
     )
 
+    // Optional preview layout
+    let previewLayout = res.locals.componentData.previewLayout
     let bodyClasses = ''
+
+    // Customise iframe preview
     if (req.query.iframe) {
       bodyClasses = 'app-iframe-in-component-preview'
+
+      if (previewLayout) {
+        bodyClasses += ` app-iframe-in-component-preview--${previewLayout}`
+      }
+
+      previewLayout = 'component-iframe'
     }
 
     res.render('component-preview', { bodyClasses, previewLayout })
--- a/app/assets/scss/partials/_app.scss
+++ b/app/assets/scss/partials/_app.scss
@@ -8,7 +8,12 @@
 
 // Injected into the preview page, that is then put in an iframe
 .app-iframe-in-component-preview {
-  margin: 15px 0;
+  margin: 5px;
+}
+
+.app-iframe-in-component-preview--full-width {
+  margin-right: 0;
+  margin-left: 0;
 }
 
 // Removes the breakpoint sass-mq debug display
@@ -22,9 +27,15 @@
 
 .app-component-preview {
   position: relative;
-  width: 100%;
-  margin-top: -15px;
-  margin-bottom: 15px;
+  @include govuk-responsive-margin(5, "top");
+  @include govuk-responsive-margin(5, "bottom");
+}
+
+// Full width for header/footer etc
+.app-component-preview--full-width {
+  max-width: none;
+  margin-right: 0;
+  margin-left: 0;
 }
 
 .app-component-preview__iframe {
--- a/app/assets/scss/partials/_app.scss
+++ b/app/assets/scss/partials/_app.scss
@@ -34,6 +34,7 @@
 // Full width for header/footer etc
 .app-component-preview--full-width,
 .app-component-preview--modal {
+  width: 100%;
   max-width: none;
   margin-right: 0;
   margin-left: 0;
--- a/app/assets/scss/partials/_app.scss
+++ b/app/assets/scss/partials/_app.scss
@@ -32,7 +32,8 @@
 }
 
 // Full width for header/footer etc
-.app-component-preview--full-width {
+.app-component-preview--full-width,
+.app-component-preview--modal {
   max-width: none;
   margin-right: 0;
   margin-left: 0;
@@ -57,11 +58,21 @@
     content: " ";
     display: table;
     width: 100%;
+    height: 100%;
     clear: both;
     box-shadow: 0 0 0 5px $app-preview-border-colour;
 
     @include govuk-if-ie8 {
       outline: 5px solid $app-preview-border-colour;
     }
+  }
+}
+
+// Allow fixed height for "position: fixed" modal
+.app-iframe-in-component-preview--modal {
+  height: 400px;
+
+  @include govuk-media-query ($from: tablet) {
+    height: 500px;
   }
 }
--- a/src/govuk/components/modal-dialogue/_modal-dialogue.scss
+++ b/src/govuk/components/modal-dialogue/_modal-dialogue.scss
@@ -1,0 +1,161 @@
+@import "../../settings/all";
+@import "../../tools/all";
+@import "../../helpers/all";
+
+@include govuk-exports("govuk/component/modal-dialogue") {
+  $govuk-dialogue-width: 640px;
+
+  .govuk-modal-dialogue,
+  .govuk-modal-dialogue__backdrop {
+    position: fixed;
+    z-index: 0;
+    top: 0;
+    left: 0;
+    width: 100%;
+    height: 100%;
+  }
+
+  // Hide dialogue when closed
+  .govuk-modal-dialogue {
+    display: none;
+  }
+
+  // Show dialogue when opened
+  .govuk-modal-dialogue--open {
+    display: block;
+  }
+
+  // Wrapper to handle overflow scrolling
+  .govuk-modal-dialogue__wrapper {
+    box-sizing: border-box;
+    display: flex;
+    height: 100%;
+    @include govuk-responsive-padding(7, "top");
+    @include govuk-responsive-padding(7, "bottom");
+    overflow-y: auto;
+    align-items: flex-start; // sass-lint:disable no-duplicate-properties
+    align-items: safe center;
+  }
+
+  // HTML5 dialogue component
+  .govuk-modal-dialogue__box {
+    box-sizing: border-box;
+    display: block;
+    position: relative;
+    z-index: 1;
+    width: 90%;
+    margin: auto;
+    padding: 0;
+    overflow-y: auto;
+    border: $govuk-focus-width solid govuk-colour("black");
+    background: govuk-colour("white");
+
+    // Add focus outline to dialogue
+    &:focus {
+      outline: $govuk-focus-width solid $govuk-focus-colour;
+    }
+
+    // Hide browser backdrop
+    &::backdrop {
+      display: none;
+    }
+  }
+
+  // Header with close button
+  .govuk-modal-dialogue__header {
+    @include govuk-clearfix;
+    @include govuk-responsive-margin(5, "bottom");
+    padding-bottom: $govuk-focus-width;
+    color: govuk-colour("white");
+    background-color: govuk-colour("black");
+    text-align: right;
+  }
+
+  // Inner content
+  .govuk-modal-dialogue__content {
+    @include govuk-font($size: 16);
+    @include govuk-responsive-padding(6);
+    padding-top: 0;
+  }
+
+  .govuk-modal-dialogue__description {
+    @include govuk-responsive-margin(4, "bottom");
+  }
+
+  // Remove bottom margins
+  .govuk-modal-dialogue__description:last-child,
+  .govuk-modal-dialogue__description > :last-child,
+  .govuk-modal-dialogue__content > :last-child {
+    margin-bottom: 0;
+  }
+
+  // Custom backdrop
+  .govuk-modal-dialogue__backdrop {
+    opacity: .8;
+    background: govuk-colour("black");
+    pointer-events: none;
+    touch-action: none;
+  }
+
+  // Crown icon
+  .govuk-modal-dialogue__crown {
+    display: block;
+    margin: 6px 0 0 6px;
+    @include govuk-responsive-margin(5, "left");
+    float: left;
+  }
+
+  // Heading
+  .govuk-modal-dialogue__heading:last-child {
+    margin-bottom: 0;
+  }
+
+  // Close button
+  .govuk-modal-dialogue__close {
+    $font-size: 36px;
+    $line-height: 1;
+
+    display: block;
+    width: auto;
+    min-width: 44px;
+    margin: 0;
+    padding: 2px 5px;
+    float: right;
+    color: govuk-colour("white");
+    background-color: govuk-colour("black");
+    box-shadow: none !important;
+    font-size: $font-size;
+    @if $govuk-typography-use-rem {
+      font-size: govuk-px-to-rem($font-size);
+    }
+    @include govuk-typography-weight-bold;
+    line-height: $line-height;
+
+    &:hover {
+      color: govuk-colour("black");
+      background-color: govuk-colour("yellow");
+    }
+
+    &:active {
+      top: 0;
+    }
+  }
+
+  // New dialogue width, inline button + link
+  @include govuk-media-query($from: tablet) {
+    .govuk-modal-dialogue__content {
+      padding-top: 0;
+    }
+
+    .govuk-modal-dialogue__box {
+      width: percentage($govuk-dialogue-width / map-get($govuk-breakpoints, desktop));
+    }
+  }
+
+  // Fixed width
+  @include govuk-media-query($from: desktop) {
+    .govuk-modal-dialogue__box {
+      width: $govuk-dialogue-width;
+    }
+  }
+}
--- a/src/govuk/components/modal-dialogue/modal-dialogue.js
+++ b/src/govuk/components/modal-dialogue/modal-dialogue.js
@@ -1,0 +1,186 @@
+import { nodeListForEach } from '../../common'
+import '../../vendor/polyfills/Element/prototype/classList'
+import '../../vendor/polyfills/Function/prototype/bind'
+import '../../vendor/polyfills/Event' // addEventListener and event.target normaliziation
+
+function ModalDialogue ($module) {
+  this.$module = $module
+  this.$dialogBox = $module.querySelector('dialog')
+  this.$container = document.documentElement
+
+  // Check for browser support
+  this.hasNativeDialog = 'showModal' in this.$dialogBox
+
+  // Allowed focussable elements
+  this.focussable = [
+    'button',
+    '[href]',
+    'input',
+    'select',
+    'textarea'
+  ]
+}
+
+// Initialize component
+ModalDialogue.prototype.init = function (options) {
+  this.options = options || {}
+
+  this.open = this.handleOpen.bind(this)
+  this.close = this.handleClose.bind(this)
+  this.focus = this.handleFocus.bind(this)
+  this.focusTrap = this.handleFocusTrap.bind(this)
+  this.boundKeyDown = this.handleKeyDown.bind(this)
+
+  // Modal elements
+  this.$buttonClose = this.$dialogBox.querySelector('.govuk-modal-dialogue__close')
+  this.$focussable = this.$dialogBox.querySelectorAll(this.focussable.toString())
+  this.$focusableLast = this.$focussable[this.$focussable.length - 1]
+  this.$focusElement = this.options.focusElement || this.$dialogBox
+
+  if (this.$dialogBox.hasAttribute('open')) {
+    this.open()
+  }
+
+  this.initEvents()
+
+  return this
+}
+
+// Initialize component events
+ModalDialogue.prototype.initEvents = function (options) {
+  if (this.options.triggerElement) {
+    this.options.triggerElement.addEventListener('click', this.open)
+  }
+
+  // Close dialogue on close button click
+  this.$buttonClose.addEventListener('click', this.close)
+}
+
+// Open modal
+ModalDialogue.prototype.handleOpen = function (event) {
+  if (event) {
+    event.preventDefault()
+  }
+
+  // Save last-focussed element
+  this.$lastActiveElement = document.activeElement
+
+  // Disable scrolling, show wrapper
+  this.$container.classList.add('govuk-!-scroll-disabled')
+  this.$module.classList.add('govuk-modal-dialogue--open')
+
+  // Close on escape key, trap focus
+  document.addEventListener('keydown', this.boundKeyDown, true)
+
+  // Optional 'onOpen' callback
+  if (typeof this.options.onOpen === 'function') {
+    this.options.onOpen.call(this)
+  }
+
+  // Skip open if already open
+  if (this.$dialogBox.hasAttribute('open')) {
+    return
+  }
+
+  // Show modal
+  this.hasNativeDialog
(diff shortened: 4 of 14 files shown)


colinrotherham:
Can we please have this on a Heroku preview URL so I can share with the team?

Thanks

colinrotherham:
Also any design feedback would be appreciated.

We're aware we added another embedded "crown" SVG, so you might want a better way of adding this to the page without duplicating the SVG markup.

I added the crown as it's already in use here:
https://components.publishing.service.gov.uk/component-guide/modal_dialogue

But not sure what the user need is.

soniaturcotte:
@alex-ju and I worked on one here, not sure if you've seen it: 
https://content-publisher.publishing.service.gov.uk/component-guide/modal_dialogue

colinrotherham:
> @alex-ju and I worked on one here, not sure if you've seen it:
> https://content-publisher.publishing.service.gov.uk/component-guide/modal_dialogue

Ahh is this the same code as used on as https://components.publishing.service.gov.uk/component-guide/modal_dialogue? If you're the original designer thank you 😁 

We've also improved the focus trapping so if you manage to click out of the modal, you're pulled back in again, not just looping from last/first items.

We've added `onOpen` and `onClose` callbacks, added native HTML5 `<dialog>` support, and turned it into a Nunjucks macro.

It's been great to improve one that's already in production.

nickcolley:
I have some questions on the design:

1. what is the user need for the crown logo on this modal?
2. we try to avoid using icons instead of words if possible as it is more accessible, has a word label been considered here?
3. is it clear that the close button is a button? should it have more visual affordance?

colinrotherham:
@nickcolley Maybe @soniaturcotte can help?

We used the design from https://components.publishing.service.gov.uk/component-guide/modal_dialogue and it tested well when did user research.

colinrotherham:
I've removed the vertical alignment fallback for older browsers using `display: table-cell` as it doesn't cater for overflowing content at all.

In IE10+ we've got basic flexbox support so that's good enough.

Older browsers like IE8 + IE9 will have the modal at the top of the screen.

colinrotherham:
@nickcolley Did you mean like this for the close button text?

Close button text

Maybe it's something @FeyAgape can explore?

You're in her capable hands now 😊

src/govuk/components/modal-dialogue/modal-dialogue.yaml
```diff
@@ -0,0 +1,120 @@
...
+      id: modal
+      open: true
+      heading:
+        text: This is a modal dialogue
+      description:
+        html: |
+          <p class="govuk-body">
+            This is some very long content.
+          </p>
+    call:
```
hannalaakso: Could we name this `caller` to be consistent with [`caller` in the fieldset](https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/components/fieldset/fieldset.yaml#L39)?

app/app.js
```diff
@@ -152,7 +152,7 @@ module.exports = (options) => {
     const componentName = req.params.component
     const requestedExampleName = req.params.example || 'default'
 
-    const previewLayout = res.locals.componentData.previewLayout || 'layout'
+    let previewLayout = res.locals.componentData.previewLayout
```
hannalaakso: Non-blocking but as this variable is now used in the [iframe if-block](https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/components/fieldset/fieldset.yaml#L39) we could consider moving it above that block.

app/views/layouts/iframe.njk
```diff
@@ -0,0 +1,21 @@
...
+
+{% block main %}
+  {% block beforeContent %}{% endblock %}
+  {% block content %}{% endblock %}
+{% endblock %}
+
+{% block bodyEnd %}
+  {{ super() }}
+  {% block scripts %}{% endblock %}
+{% endblock %}
```
hannalaakso: Would it make sense to call this file `component-iframe.njk`? We already have `component-preview.njk`.

app/assets/scss/partials/_app.scss
```diff
@@ -46,6 +53,7 @@
     content: " ";
     display: table;
     width: 100%;
+    height: 100%;
```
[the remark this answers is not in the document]
colinrotherham: @hannalaakso Ah I thought that was a bug!

The blue outline now goes all the way around the iframe.

Because `iFrameResizer` can't calculate the height of a `position: fixed` modal, the fixed height comes from this bit that I've hard-coded:

```css
// Allow fixed height for "position: fixed" modal
.app-iframe-in-component-preview--modal {
  height: 400px;

  @include govuk-media-query ($from: tablet) {
    height: 500px;
  }
}
```

app/assets/scss/partials/_app.scss
```diff
@@ -54,3 +62,12 @@
...
   }
 }
+
+// Allow fixed height for "position: fixed" modal
+.app-iframe-in-component-preview--modal {
+  height: 400px;
+
+  @include govuk-media-query ($from: tablet) {
+    height: 500px;
+  }
```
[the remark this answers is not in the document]
colinrotherham: Thanks Hannah. If you could find a way to get `iFrameResizer` work for the modal that would be even better than a fixed height.

Looks like there's a `heightCalculationMethod` API method for iframe child content that could be customised? Currently it's not working due to `position: fixed`.

src/govuk/components/modal-dialogue/_modal-dialogue.scss
```diff
@@ -0,0 +1,167 @@
+// Disables linting for this file only
+// sass-lint:disable no-duplicate-properties
```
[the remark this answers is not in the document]
colinrotherham: @hannalaakso Annoyingly this feature was removed from `sass-lint`, but the documentation wasn't reverted. See https://github.com/sasstools/sass-lint/issues/1292

What looks like per-line disable rules actually affect the whole file! 🙈 

I'll move it anyway, as it's good to see which line it's trying to apply to, hopefully they enable it again in a future release.

app/assets/scss/partials/_app.scss
```diff
@@ -22,9 +22,16 @@
 
 .app-component-preview {
   position: relative;
-  width: 100%;
```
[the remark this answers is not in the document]
colinrotherham: @hannalaakso Thanks I've put this back. It was the `margin: 5px` change that was causing the blue outline to appear on the left/right of full-width previews.

After

src/govuk/components/modal-dialogue/modal-dialogue.yaml
colinrotherham: @hannalaakso Sounds good, I've changed it 👍

app/app.js
colinrotherham: @hannalaakso Yep sounds good too, all changed

app/views/layouts/iframe.njk
colinrotherham: @hannalaakso Done

src/govuk/components/modal-dialogue/modal-dialogue.js
```diff
@@ -0,0 +1,186 @@
...
+function ModalDialogue ($module) {
+  this.$module = $module
+  this.$dialogBox = $module.querySelector('dialog')
+  this.$container = document.documentElement
+
+  // Check for browser support
+  this.hasNativeDialog = 'showModal' in this.$dialogBox
+
+  // Allowed focussable elements
+  this.focussable = [
```
nickcolley: There is an 'inert' feature we could consider referring to when reviewing the focus trap functionality in this proposal:

https://github.com/WICG/inert

It ships in chromium, I'm not sure if it's a proper standard yet but worth thinking about.

src/govuk/components/modal-dialogue/modal-dialogue.js
```diff
@@ -0,0 +1,186 @@
...
+    'select',
+    'textarea'
+  ]
+}
+
+// Initialize component
+ModalDialogue.prototype.init = function (options) {
+  this.options = options || {}
+
+  this.open = this.handleOpen.bind(this)
```
nickcolley: I wonder if these method behinds should be in the constructor instead?

src/govuk/components/modal-dialogue/modal-dialogue.js
```diff
@@ -0,0 +1,186 @@
...
+
+  // Hide wrapper, enable scrolling
+  this.$module.classList.remove('govuk-modal-dialogue--open')
+  this.$container.classList.remove('govuk-!-scroll-disabled')
+
+  // Restore focus to last active element
+  this.$lastActiveElement.focus()
+
+  // Optional 'onClose' callback
+  if (typeof this.options.onClose === 'function') {
```
nickcolley: I wonder if we need to consider how we're documenting the JavaScript options for components, since we use data-attributes elsewhere instead of programmatic options.

See also: [Proposal programmatic and data attribute api interface for components](https://github.com/alphagov/govuk-frontend/pull/1238)

colinrotherham: @nickcolley Yeah good point.

Here's how we currently use this Modal API on our prototype:

```js
var $dialogue = document.querySelector('[data-module="govuk-modal-dialogue"]')
var $dialogueButtonResume = document.querySelector('.govuk-modal-dialogue__continue')

if ($dialogue && $dialogueButtonResume) {
  var modal = new window.ModalDialogue($dialogue)
    .init({
      focusElement: $dialogueButtonResume,
      onClose: function () {
        // Extend session timeout here
      }
    })
}
```

src/govuk/components/modal-dialogue/modal-dialogue.js
colinrotherham: Yeah sounds like a good enhancement.

You'd have to be _very_ careful and keep the modal outside `govuk-width-container` or whatever the page wrapper element was.

```html
<body>
    <div class="govuk-width-container" inert>…</div>
    <div data-module="govuk-modal-dialogue">…</div>
</body>
```

It could go horribly wrong if the modal was inside the `[inert]` wrapper.

src/govuk/components/modal-dialogue/modal-dialogue.js
colinrotherham: @nickcolley I try to defer anything that uses more CPU cycles until it's actually needed (like when `.init()` gets run) but it's probably negligible so don't mind.

app/assets/scss/partials/_app.scss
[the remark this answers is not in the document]
hannalaakso: Going to resolve this as it doesn't concern the component itself. Good to have the conversation here for reference though.

app/assets/scss/partials/_app.scss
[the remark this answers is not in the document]
hannalaakso: Going to resolve this as it doesn't concern the component itself. Good to have the conversation here for reference though.

src/govuk/components/modal-dialogue/modal-dialogue.js
hannalaakso: The early prototype version of the modal that GOV.UK based theirs on [used both the `inert` attribute and `aria-hidden`](https://github.com/hannalaakso/accessible-timeout-warning/blob/0991356bd5ceba342e810e494e567fa61c1fe9b9/src/govuk/components/timeout-warning/timeout-warning.js#L228) to make the `<main>` wrapper non-actionable and this tested well with screen reader users. The modal markup was outside `<main>` and below `<body>`. 

Colin does have a valid point: we need to be clear in the guidance about where the modal markup is placed in the page content.

src/govuk/components/modal-dialogue/modal-dialogue.js
hannalaakso: Thanks Colin 👍 We believe that the performance impact is negligible so we should make the change for consistency with other components.

src/govuk/components/modal-dialogue/_modal-dialogue.scss
```diff
@@ -0,0 +1,161 @@
+@import "../../settings/all";
+@import "../../tools/all";
+@import "../../helpers/all";
+
+@include govuk-exports("govuk/component/modal-dialogue") {
+  $govuk-dialogue-width: 640px;
```
hannalaakso: We probably want to use [`!default`](https://sass-lang.com/documentation/variables#default-values) here so that the value can overridden.

src/govuk/components/modal-dialogue/_modal-dialogue.scss
```diff
@@ -0,0 +1,161 @@
...
+    align-items: flex-start; // sass-lint:disable no-duplicate-properties
+    align-items: safe center;
+  }
+
+  // HTML5 dialogue component
+  .govuk-modal-dialogue__box {
+    box-sizing: border-box;
+    display: block;
+    position: relative;
+    z-index: 1;
```
hannalaakso: We might want to increase this, just in the (non-ideal) case that the page has other competing elements. I notice that the [GOV.UK Publishing equivalent sets this to 1000](https://github.com/alphagov/govuk_publishing_components/blob/34315777fabc6ac53348e636c7d3006b62228fcd/app/assets/stylesheets/govuk_publishing_components/components/_modal-dialogue.scss#L3).

src/govuk/components/modal-dialogue/_modal-dialogue.scss
```diff
@@ -0,0 +1,161 @@
...
+    width: 90%;
+    margin: auto;
+    padding: 0;
+    overflow-y: auto;
```
hannalaakso: This might be okay as it is but just for reference, I notice that the GOV.UK Publishing example currently sets these to:
```
overflow-x: hidden;
overflow-y: scroll;
```
I don't think we currently have an example with page content in the main area so we need to add one in order to check that the main page content is not scrollable and the modal is.

src/govuk/components/modal-dialogue/_modal-dialogue.scss
```diff
@@ -0,0 +1,161 @@
...
+    margin-bottom: 0;
+  }
+
+  // Custom backdrop
+  .govuk-modal-dialogue__backdrop {
+    opacity: .8;
+    background: govuk-colour("black");
+    pointer-events: none;
+    touch-action: none;
+  }
```
hannalaakso: I think this block should live below https://github.com/alphagov/govuk-frontend/pull/1668/files#diff-40be96fa6566e06e9d9d9ab4d1d3d831R8 to in order to keep the backdrop styles together.

src/govuk/components/modal-dialogue/_modal-dialogue.scss
```diff
@@ -0,0 +1,161 @@
...
+      color: govuk-colour("black");
+      background-color: govuk-colour("yellow");
+    }
+
+    &:active {
+      top: 0;
+    }
+  }
+
+  // New dialogue width, inline button + link
```
hannalaakso: I'm not sure why these styles are set separately from the other ones in this file. Should this be a modifier class?

src/govuk/components/modal-dialogue/_modal-dialogue.scss
```diff
@@ -0,0 +1,161 @@
...
+      padding-top: 0;
+    }
+
+    .govuk-modal-dialogue__box {
+      width: percentage($govuk-dialogue-width / map-get($govuk-breakpoints, desktop));
+    }
+  }
+
+  // Fixed width
+  @include govuk-media-query($from: desktop) {
```
hannalaakso: I'm not sure why these styles are set separately from the other ones in this file. Should this be a modifier class?

src/govuk/components/modal-dialogue/modal-dialogue.js
```diff
@@ -0,0 +1,186 @@
...
+
+// Initialize component events
+ModalDialogue.prototype.initEvents = function (options) {
+  if (this.options.triggerElement) {
+    this.options.triggerElement.addEventListener('click', this.open)
+  }
+
+  // Close dialogue on close button click
+  this.$buttonClose.addEventListener('click', this.close)
+}
```
hannalaakso: Not a big deal but I wonder if the events inside this block could live inside `init()`, it'll be tidier once we move the variable definitions to the constructor.

src/govuk/components/modal-dialogue/modal-dialogue.js
```diff
@@ -0,0 +1,186 @@
...
+ModalDialogue.prototype.handleOpen = function (event) {
+  if (event) {
+    event.preventDefault()
+  }
+
+  // Save last-focussed element
+  this.$lastActiveElement = document.activeElement
+
+  // Disable scrolling, show wrapper
+  this.$container.classList.add('govuk-!-scroll-disabled')
```
hannalaakso: Does `govuk-!-scroll-disabled` do anything? I can't see any styles created for it.

src/govuk/components/modal-dialogue/template.njk
```diff
@@ -0,0 +1,69 @@
...
+          interpreted as such, displaying the fallback image. #}
+          <image src="{{ params.assetsPath | default('/assets/images') }}/govuk-logotype-crown.png" xlink:href="" class="govuk-header__logotype-crown-fallback-image" width="36" height="32"></image>
+        </svg>
+
+        {{ govukButton({
+          text: "×",
+          type: "button",
+          classes: "govuk-modal-dialogue__close",
+          attributes: {
+            "aria-label": "Close modal dialogue"
```
hannalaakso: I wonder if all users will know what is meant by "modal dialogue"? The users of GOV.UK Publishing (which this is based on) would mainly be repeat users with at least a moderate level of computer literacy and are therefore not representative of the wider population.

If we could test with an alternative wording that would be great.

src/govuk/components/modal-dialogue/template.njk
```diff
@@ -0,0 +1,69 @@
...
+        'focusable' when using the keyboard to navigate. #}
+        <svg
+          role="presentation"
+          focusable="false"
+          class="govuk-modal-dialogue__crown"
+          xmlns="http://www.w3.org/2000/svg"
+          viewbox="0 0 132 97"
+          height="30"
+          width="36"
+        >
```
hannalaakso: Had a chat with @FeyAgape. We should consider removing the crown logo as there is no clear user need. Its usage here is not consistent with our other components and it adds complexity. 

Its usage in public facing sites in this manner might also be problematic with regards to the branding guidelines; the [Service Manual guidance](https://www.gov.uk/service-manual/design/making-your-service-look-like-govuk#if-your-service-isnt-on-govuk) only mentions the use of the crown logo in the header.

src/govuk/components/modal-dialogue/template.njk
```diff
@@ -0,0 +1,69 @@
...
+          image it has no effect. It's important to include the empty xlink:href
+          attribute as this prevents versions of IE which support SVG from
+          downloading the fallback image when they don't need to.
+          In other browsers <image> is synonymous for the <img> tag and will be
+          interpreted as such, displaying the fallback image. #}
+          <image src="{{ params.assetsPath | default('/assets/images') }}/govuk-logotype-crown.png" xlink:href="" class="govuk-header__logotype-crown-fallback-image" width="36" height="32"></image>
+        </svg>
+
+        {{ govukButton({
+          text: "×",
```
hannalaakso: Had a chat with @FeyAgape. If possible, we should explore using a text equivalent here as icons might not be understood by all users.

src/govuk/components/modal-dialogue/modal-dialogue.yaml
```diff
@@ -0,0 +1,120 @@
...
+  description: The modal dialog’s prompt text. If `description.html` is supplied, this is ignored.
+- name: description.html
+  type: string
+  required: true
+  description: The HTML content for the modal dialog’s prompt text.
+
+previewLayout: modal
+
+examples:
+  - name: default
```
hannalaakso: Could we make this example to open the modal with a button? Similar to the [GOV.UK Publishing modal example](https://components.publishing.service.gov.uk/component-guide/modal_dialogue/preview). This would allow us to test the transition of the modal being closed to it being open, and how the main page content is made actionable / non-actionable.

Doing this might require some tweaking of the review app.

src/govuk/components/modal-dialogue/modal-dialogue.yaml
```diff
@@ -0,0 +1,120 @@
...
+  - name: open
+    data:
+      id: modal
+      open: true
+      heading:
+        text: This is a modal dialogue
+      description:
+        html: <p class="govuk-body">Prompt text for the modal dialogue component goes here</p>
+
+  - name: open with scrolling content
```
hannalaakso: Could we make this example to have some page content in the main page area?  Similar to the [GOV.UK Publishing modal example](https://components.publishing.service.gov.uk/component-guide/modal_dialogue/preview). This would allow us to test that the background is not scrollable / actionable when the modal is open.

src/govuk/components/modal-dialogue/modal-dialogue.yaml
```diff
@@ -0,0 +1,120 @@
+params:
+- name: id
+  type: string
+  required: true
+  description: The id of the modal dialogue.
+- name: open
+  type: boolean
+  required: true
+  description: If true, modal dialogue element will be visible.
```
hannalaakso: Should we consider calling the component "modal dialog" instead? Most of the industry docs talk about "modal dialog" or "dialog element":
https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/dialog.html
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog

nickcolley: https://trends.google.com/trends/explore?cat=422&q=modal,dialog,modal%20dialog,modal%20dialogue,dialogue

src/govuk/components/modal-dialogue/modal-dialogue.js
```diff
@@ -0,0 +1,186 @@
...
+}
+
+// Lock scroll, focus modal
+ModalDialogue.prototype.handleFocus = function () {
+  this.$dialogBox.scrollIntoView()
+  this.$focusElement.focus({ preventScroll: true })
+}
+
+// Ensure focus stays within modal
+ModalDialogue.prototype.handleFocusTrap = function (event) {
```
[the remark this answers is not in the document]
nickcolley: To add to this point, what if a user navigates via landmarks? What is the impact on them when the modal is open.

nickcolley:
@hannalaakso I can help review this when you're happy later on. I think we should do some real device testing at some point too.

NogginBox:
What is the status of this PR?

We're adding a timeout modal in our next sprint and I would like to use this component.

hannalaakso:
@NogginBox Thanks for reaching out. This work depends on https://github.com/alphagov/govuk-frontend/issues/1708 especially, and also https://github.com/alphagov/govuk-frontend/issues/1722. It's unfortunately not possible to give a timeline on the work at the moment but we will pick it up when we're able to prioritise it.

NogginBox:
I've used the code from this PR to get a modal dialogue timeout warning in our site.

I've made one change to it, which which is to increase the z-index to 100 to make sure that it is above all other elements of the site.

src/govuk/components/modal-dialogue/_modal-dialogue.scss
```diff
@@ -0,0 +1,161 @@
...
+@import "../../tools/all";
+@import "../../helpers/all";
+
+@include govuk-exports("govuk/component/modal-dialogue") {
+  $govuk-dialogue-width: 640px;
+
+  .govuk-modal-dialogue,
+  .govuk-modal-dialogue__backdrop {
+    position: fixed;
+    z-index: 0;
```
NogginBox, line 11: Increase to 1000 so that this will site above all other elements already in the site.

I had to do this when I've used this code as several elements floated on top of the backdrop.

src/govuk/components/modal-dialogue/_modal-dialogue.scss
NogginBox: I've made this 1001 so it is above the backdrop which also needs the z-index increasing.
```suggestion
    z-index: 1001;
```

src/govuk/components/modal-dialogue/_modal-dialogue.scss
```diff
@@ -0,0 +1,161 @@
...
+    @include govuk-responsive-margin(5, "bottom");
+    padding-bottom: $govuk-focus-width;
+    color: govuk-colour("white");
+    background-color: govuk-colour("black");
+    text-align: right;
+  }
+
+  // Inner content
+  .govuk-modal-dialogue__content {
+    @include govuk-font($size: 16);
```
NogginBox, line 76: I was able to safely remove this font definition as I wanted the site's standard fonts to be used. We're a non gov body so can't use the standard GDS font and have overrides in place to not use them. This line being included broke our GDS font overrides.
```suggestion
```

zaizous:
Hi @hannalaakso, do you guys have any update on the timeline? 
We are looking into implementing time out in the coming sprints and would be good to make use of the GDS pattern.

hannalaakso:
@zaizous No update at the moment. You might find the [acceptance criteria which this component will need to meet](https://github.com/alphagov/govuk-design-system-backlog/issues/30#issuecomment-553932293) helpful if you're researching this topic.

hannalaakso:
I’m closing this PR as it’s currently blocked by #1722 and #1708.

Note for teams who are researching modal dialogues: the work done in this PR could be used as the basis for creating a modal dialogue. We would encourage teams to review: 
- [acceptance criteria related to this PR](https://github.com/alphagov/govuk-design-system-backlog/issues/30#issuecomment-553932293), especially the points about design and accessibility
- conversations on the [modal dialogue backlog issue
](https://github.com/alphagov/govuk-design-system-backlog/issues/30)

Issue #1417, opened 2019-05-31
Link hover states are not clear


nickcolley:
This issue is from a May 2019 external accessibility audit report.

**WCAG Reference**: Usability feedback only, there is no WCAG related guidelines.
**Issue ID**: DAC_Issue33
**URL**: Throughout

### Before hover
![](https://user-images.githubusercontent.com/2445413/58717364-f93e1400-83c2-11e9-8a4d-54d25643938e.png)

### On hover
![](https://user-images.githubusercontent.com/2445413/58717369-fb07d780-83c2-11e9-9b5d-729d0d040515.png)

The colour change when a user hovers over a link is not clear and this was especially difficult for low vision users to determine.

Ensuring that the state change is distinctive would benefit low vision users in particular, while benefiting all mouse users in general.

## Current Code Ref(s)
```$govuk-link-colour #005ea5 $govuk-link-hover-colour #2b8cc4```

## Low vision user comments
> “I had difficulty navigating the links on all pages because little visual feedback was presented when using mouse hover. While they did change to purple, this change was difficult to notice due to the links then looking similar to the regular text on mouse hover.”

## Solution
Offer a clearer colour change i.e. similar to that of the keyboard focus, or add/remove an underline when users hover over the links/buttons.

nickcolley:
I think it'd good to think about https://github.com/alphagov/govuk-frontend/issues/1205 as well if we did this.

nickcolley:
Hover states in general could do with a review for example the error summary links and tabs component do not have hover states (perhaps more components too).

nickcolley:
Noticed the NHS seem to have responded to this challenge too (perhaps by DAC as well!).

## Main NHS.UK website
### No interaction
![](https://user-images.githubusercontent.com/2445413/73543384-c014e280-442e-11ea-9eb1-a4e6889e1e1d.png)

### Hover with a mouse
![](https://user-images.githubusercontent.com/2445413/73543392-c73bf080-442e-11ea-9535-72044eba9f8d.png)

### Focus with a keyboard
![](https://user-images.githubusercontent.com/2445413/73543400-cdca6800-442e-11ea-93ac-0c33529c4061.png)

I think this approach could potentially be problematic because the hover and the focus style are so similar a user could be confused what they're focused on and accidentally interact with wrong element.

![](https://user-images.githubusercontent.com/2445413/73543463-eb97cd00-442e-11ea-974f-9f1316f687d4.png)

[...]

PR #1672, opened 2019-12-04
Ensure footer links look clickable

nickcolley:
We originally removed the underlines from these links to make them consistent with the approach we have for the header component.

Many community members have asked why we have made this change, and justification seems to be related to allowing a better hover state.

This makes sense but is not consistent with our other components, for example:

- back link
- breadcrumbs
- error summary links
- tabs component 

do not have a distinct hover state.

I think we should [consider hover states (and touch states) at a system level](https://github.com/alphagov/govuk-frontend/issues/1417#issuecomment-561601401) separately.

In the meantime we can make sure that the links in the footer have the affordance of links, as we did not have any evidence from users that removing these underlines is helpful.

Original reasoning for removing underlines: https://github.com/alphagov/govuk-frontend/pull/1321#issuecomment-549364442

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ If you're using Nunjucks, you can now add classes to the character count compone
 ### Fixes
 
 - [Pull request #1676: Fix skip link component focus style with global styles enabled](https://github.com/alphagov/govuk-frontend/pull/1676).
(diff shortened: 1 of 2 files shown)


nickcolley:
From @timpaul :

> I chatted to Ste McC about this today [...]. On reflection we both agree that we should reinstate them. The context of the header is pretty ambiguous compared to other navigation - some things are links, some things aren't, so we need to provide additional context with underlines.

edwardhorsford:
👏

PR #1673, opened 2019-12-05
Fix typo and link

frankieroberto:
The README for the hint component referred to a "label". The jump-link also needed to be updated.

### DIFF
--- a/src/govuk/components/hint/README.md
+++ b/src/govuk/components/hint/README.md
@@ -6,7 +6,7 @@ See the [main README quick start guide](https://github.com/alphagov/govuk-fronte
 
 ## Guidance and Examples
 
-The label component is used in other input components, to see an example of it in use see the [text input component](https://design-system.service.gov.uk/components/text-input/#using-hint-text).
+The hint component is used in other input components. To see an example of it in use, see the [text input component](https://design-system.service.gov.uk/components/text-input/#hint-text).
 
 ## Component options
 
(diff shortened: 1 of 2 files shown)


src/govuk/components/hint/README.md
```diff
@@ -6,7 +6,7 @@ See the [main README quick start guide](https://github.com/alphagov/govuk-fronte
 
 ## Guidance and Examples
 
-The label component is used in other input components, to see an example of it in use see the [text input component](https://design-system.service.gov.uk/components/text-input/#using-hint-text).
+The hint component is used in other input components. To see an example of it in use, see the [text input component](https://design-system.service.gov.uk/components/text-input/#hint-text).
```
nickcolley: I'd suggest only updating the word 'hint' or updating the 'label' README to if you think there's worth improving the grammar, since this same format is used elsewhere:

https://github.com/alphagov/govuk-frontend/tree/master/src/govuk/components/label#guidance-and-examples

nickcolley:
Thanks @frankieroberto :)

src/govuk/components/hint/README.md
frankieroberto: @nickcolley I've updated the label README too in 69dd3b7aa78eac610345ef0e636b440d5e1d0193. Happy to revert and just change "label" to "hint" (and fix the link anchor) though. I felt it would be slightly more readable as separate sentences.

PR #1697, opened 2020-01-16
Improve npm install documentation

m-green:
This updates the Frontend npm install documentation to improve the content's readability and accuracy. This is a draft PR because updating the docs in the GitHub repo is 'plan B'. 'Plan A' is to publish these docs in the new Tech Docs Template site. 

Please review for factual accuracy. 

While reviewing, please bear in mind that:

- this is more of a tidy up than a top-to-bottom improvement, and we're taking it into research to test style, structure, possible gaps, etc. So please only suggest factual fixes rather than new additions or larger changes.
- the current docs are somewhat based on a ‘starting from scratch’ install approach. We've theorised that users would benefit from docs that did more to take their current setups into account. That's out of scope for now - but we’ll explore which is the better approach through our research.
- it's likely that we'll split all this into 'Install' and 'Get started' at some point - because having a doc this long that's titled 'Install' is potentially making installation look off-puttingly long and complicated.

### DIFF
--- a/docs/installation/get-started.md
+++ b/docs/installation/get-started.md
@@ -42,7 +42,7 @@ Your component will not use the right font or images until you've added GOV.UK F
 
 2. Run your application, then use [the Fonts tab in Firefox Page Inspector](https://developer.mozilla.org/en-US/docs/Tools/Page_Inspector/How_to/Edit_fonts#The_Fonts_tab) to check the accordion is using the GDS Transport font.
 
-In your live application, we recommend [routing requests to `node-modules`](/docs/installation/importing-css-assets-and-javascript.md#fonts-and-image-assets) instead of copying files.
+In your live application, we recommend [routing requests to `node-modules`](/docs/installation/importing-css-assets-and-javascript.md#font-and-image-assets) instead of copying files.
 
 ## Get the JavaScript working
 
--- a/docs/installation/get-started.md
+++ b/docs/installation/get-started.md
@@ -39,7 +39,9 @@ Paste the HTML into a page or template in your application.
       </head>
     ```
 
-3. Run your application and check that the accordion renders using styles that start with `govuk-`.
+3. Run your application and check that the accordion displays correctly.
+
+The accordion will use a generic font until you get the font and images working, and will not be interactive until you get the JavaScript working.
 
 There are also different ways you can [import GOV.UK Frontend's CSS](/docs/installation/importing-css-assets-and-javascript.md#css).
 
--- a/docs/installation/get-started.md
+++ b/docs/installation/get-started.md
@@ -1,24 +1,36 @@
 # Get started
 
-Get one GOV.UK Frontend component working in your application.
+Get one GOV.UK Frontend component working in your application, so you can test everything works correctly before you add more components or styles.
 
-You must [install GOV.UK Frontend using npm](/docs/installation/installing-with-npm.md) first. If you've installed from `dist`, get started with [a basic page](/docs/installation/installing-from-dist.md#2-include-resources) instead.
+You will need to do the following.
 
-## Add a component to your application
+1. Install GOV.UK Frontend.
+2. Add the HTML for a component to your application.
+3. Get the CSS working.
+4. Get the font and images working.
+5. Get the JavaScript working.
+
+## 1. Install
+
+[Install GOV.UK Frontend using npm](/docs/installation/installing-with-npm.md).
+
+If you've installed from `dist`, get started with [a basic page](/docs/installation/installing-from-dist.md#2-include-resources) instead.
+
+## 2. Add the HTML for a component to your application
 
 Go to the [example accordion component](https://design-system.service.gov.uk/components/accordion/#accordion-example) on the GOV.UK Design System website, then copy the HTML.
 
-Paste the HTML into your application.
+Paste the HTML into a page or template in your application.
 
-## Get the CSS working
+## 3. Get the CSS working
 
-1. Add the following line to the main Sass file in your project, so your Sass compiler adds all GOV.UK Frontend styles to your CSS file.
+1. Add the following to the main Sass file in your project, so your Sass compiler adds all of GOV.UK Frontend's styles to your CSS file.
 
     ```Scss
     @import "node_modules/govuk-frontend/govuk/all";
     ```
 
-2. Add your CSS file to your page layout. For example:
+2. Add your CSS file to your page layout if you need to. For example:
 
     ```html
       <head>
@@ -27,36 +39,34 @@ Paste the HTML into your application.
       </head>
     ```
 
-3. Run your application and check that the accordion is rendering using styles that start with `govuk-`.
+3. Run your application and check that the accordion renders using styles that start with `govuk-`.
 
 There are also different ways you can [import GOV.UK Frontend's CSS](/docs/installation/importing-css-assets-and-javascript.md#css).
 
-## Get the font and images working
+## 4. Get the font and images working
 
 Your component will not use the right font or images until you've added GOV.UK Frontend's assets to your application.
 
-1. Use an automated task or your build pipeline to copy the:
+1. Copy the:
 
     - `/node_modules/govuk-frontend/govuk/assets/images` folder to `<YOUR-APP>/assets/images`
     - `/node_modules/govuk-frontend/govuk/assets/fonts` folder to `<YOUR-APP>/assets/fonts`
 
 2. Run your application, then use [the Fonts tab in Firefox Page Inspector](https://developer.mozilla.org/en-US/docs/Tools/Page_Inspector/How_to/Edit_fonts#The_Fonts_tab) to check the accordion is using the GDS Transport font.
 
-In your live application, we recommend [routing requests to `node-modules`](/docs/installation/importing-css-assets-and-javascript.md#font-and-image-assets) instead of copying files.
+In your live application, we recommend [using an automated task or your build pipeline](/docs/installation/importing-css-assets-and-javascript.md#font-and-image-assets) instead of copying the files manually.
 
-## Get the JavaScript working
+## 5. Get the JavaScript working
 
-If you do not initialise GOV.UK Frontend's JavaScript with `initAll`, some components will not work correctly for disabled users who use assistive technologies.
-
-1. Add the `.js-enabled` class to the opening `<body>` tag of your page template. For example:
+1. Add the following JavaScript to the top of the `<body>` section of your page template:
 
     ```JavaScript
     document.body.className = ((document.body.className) ? document.body.className + ' js-enabled' : 'js-enabled');
     ```
 
-2. Use an automated task or your build pipeline to copy the `/node-modules/govuk-frontend/govuk/all.js` file to `<YOUR-JAVASCRIPT-FOLDER>/govuk.js`
+2. Copy the `/node-modules/govuk-frontend/govuk/all.js` file to `<YOUR-JAVASCRIPT-FOLDER>/govuk.js`.
 
-3. Import the Javascript file before the closing `</body>` tag of your page template, then run the `initAll` function to initialise all the components. For example:
+3. Import the file before the closing `</body>` tag of your page template, then run the `initAll` function to initialise all the components. For example:
 
     ```html
     <body>
@@ -70,4 +80,9 @@ If you do not initialise GOV.UK Frontend's JavaScript with `initAll`, some compo
 
 4. Run your application and check you can show and hide sections of the accordion by selecting the buttons.
 
-There are also different ways you can [import GOV.UK Frontend's JavaScript](/docs/installation/importing-css-assets-and-javascript.md#javascript).
+In your live application:
+
+- you must use `initAll` to initialise all components that use GOV.UK Frontend's JavaScript, or some components will not work correctly for disabled users who use assistive technologies
+- we recommend [using an automated task or your build pipeline](/docs/installation/importing-css-assets-and-javascript.md#javascript) instead of copying the files manually
+
+You can now get the full code for page layouts and other components from the [Design System website](https://design-system.service.gov.uk/).
--- a/docs/installation/get-started.md
+++ b/docs/installation/get-started.md
@@ -1,0 +1,73 @@
+# Get started
+
+Get one GOV.UK Frontend component working in your application.
+
+You must [install GOV.UK Frontend using npm](/docs/installation/installing-with-npm.md) first. If you've installed from `dist`, get started with [a basic page](/docs/installation/installing-from-dist.md#2-include-resources) instead.
+
+## Add a component to your application
+
+Go to the [example accordion component](https://design-system.service.gov.uk/components/accordion/#accordion-example) on the GOV.UK Design System website, then copy the HTML.
+
+Paste the HTML into your application.
+
+## Get the CSS working
+
+1. Add the following line to the main Sass file in your project, so your Sass compiler adds all GOV.UK Frontend styles to your CSS file.
+
+    ```Scss
+    @import "node_modules/govuk-frontend/govuk/all";
+    ```
+
+2. Add your CSS file to your page layout. For example:
+
+    ```html
+      <head>
+        ...
+        <link rel="stylesheet" href="<YOUR-CSS-FILE>.css">
+      </head>
+    ```
+
+3. Run your application and check that the accordion is rendering using styles that start with `govuk-`.
+
+There are also different ways you can [import GOV.UK Frontend's CSS](/docs/installation/importing-css-assets-and-javascript.md#css).
+
+## Get the font and images working
+
+Your component will not use the right font or images until you've added GOV.UK Frontend's assets to your application.
+
+1. Use an automated task or your build pipeline to copy the:
+
+    - `/node_modules/govuk-frontend/govuk/assets/images` folder to `<YOUR-APP>/assets/images`
+    - `/node_modules/govuk-frontend/govuk/assets/fonts` folder to `<YOUR-APP>/assets/fonts`
+
+2. Run your application, then use [the Fonts tab in Firefox Page Inspector](https://developer.mozilla.org/en-US/docs/Tools/Page_Inspector/How_to/Edit_fonts#The_Fonts_tab) to check the accordion is using the GDS Transport font.
+
+In your live application, we recommend [routing requests to `node-modules`](/docs/installation/importing-css-assets-and-javascript.md#fonts-and-image-assets) instead of copying files.
+
+## Get the JavaScript working
+
+If you do not initialise GOV.UK Frontend's JavaScript with `initAll`, some components will not work correctly for disabled users who use assistive technologies.
+
+1. Add the `.js-enabled` class to the opening `<body>` tag of your page template. For example:
+
+    ```JavaScript
+    document.body.className = ((document.body.className) ? document.body.className + ' js-enabled' : 'js-enabled');
+    ```
+
+2. Use an automated task or your build pipeline to copy the `/node-modules/govuk-frontend/govuk/all.js` file to `<YOUR-JAVASCRIPT-FOLDER>/govuk.js`
+
+3. Import the Javascript file before the closing `</body>` tag of your page template, then run the `initAll` function to initialise all the components. For example:
+
+    ```html
+    <body>
+      ...
+      <script src="<YOUR-JAVASCRIPT-FOLDER>/govuk.js"></script>
+      <script>
+        window.GOVUKFrontend.initAll()
+      </script>
+    </body>
+    ```
+
+4. Run your application and check you can show and hide sections of the accordion by selecting the buttons.
+
+There are also different ways you can [import GOV.UK Frontend's JavaScript](/docs/installation/importing-css-assets-and-javascript.md#javascript).
--- a/docs/installation/get-started.md
+++ b/docs/installation/get-started.md
@@ -1,8 +1,8 @@
 # Get started
 
-Get one GOV.UK Frontend component working in your application, so you can test everything works correctly before you add more components or styles.
+Get one GOV.UK Frontend component working in your application, so you can test everything works before you add more components or styles.
 
-You will need to do the following.
+You will need to do all the following steps to get your component working.
 
 1. Install GOV.UK Frontend.
 2. Add the HTML for a component to your application.
@@ -49,7 +49,7 @@ There are also different ways you can [import GOV.UK Frontend's CSS](/docs/insta
 
 Your component will not use the right font or images until you've added GOV.UK Frontend's assets to your application.
 
-1. Copy the:
+1. Copy the following 2 folders:
 
     - `/node_modules/govuk-frontend/govuk/assets/images` folder to `<YOUR-APP>/assets/images`
     - `/node_modules/govuk-frontend/govuk/assets/fonts` folder to `<YOUR-APP>/assets/fonts`
@@ -80,7 +80,7 @@ In your live application, we recommend [using an automated task or your build pi
     </body>
     ```
 
-4. Run your application and check you can show and hide sections of the accordion by selecting the buttons.
+4. Run your application and check it works the same way as the Design System accordion example, by selecting the buttons and checking the accordion shows and hides sections.
 
 In your live application:
 
--- a/docs/installation/importing-css-assets-and-javascript.md
+++ b/docs/installation/importing-css-assets-and-javascript.md
@@ -4,7 +4,7 @@
 
 ### Import all the CSS
 
-To import all the CSS styles from GOV.UK Frontend, add the following to your Sass file:
+To import all the Sass rules from GOV.UK Frontend, add the following to your Sass file:
 
 ```Scss
 @import "node_modules/govuk-frontend/govuk/all";
@@ -25,7 +25,7 @@ If you want to make Sass import paths shorter, add `node_modules/govuk-frontend`
 - [Sass load paths](https://sass-lang.com/documentation/at-rules/import#finding-the-file)
 - [assets paths](http://guides.rubyonrails.org/asset_pipeline.html#search-paths) if you use Ruby in your project
 
-You can then import only the [button](https://design-system.service.gov.uk/components/button/) component for example, by adding the following to your Sass file:
+You can then import without using `node_modules/govuk-frontend/` in your import path. For example:
 
 ```SCSS
 @import "govuk/components/button/button";
@@ -43,12 +43,12 @@ If your project uses GOV.UK Frontend toolkit, GOV.UK Template or GOV.UK Elements
 
 To use the font and image assets from GOV.UK Frontend, you can either:
 
-- route requests to the GOV.UK Frontend package - recommended
-- copy the font and images into your application
+- serve the assets from the GOV.UK Frontend assets folder - recommended
+- copy the font and image files into your application
 
-### Route requests - recommended
+### Serve the assets from the GOV.UK Frontend assets folder - recommended
 
-Route requests from `<YOUR-SITE-URL>/assets` to `/node_modules/govuk-frontend/govuk/assets`.
+Set up your routing so that requests for files in `<YOUR-SITE-URL>/assets` are served from `/node_modules/govuk-frontend/govuk/assets`.
 
 For example if you're using [express.js](https://expressjs.com/), add the following to your `app.js` file:
 
@@ -57,18 +57,18 @@ var path = require('path');
 app.use('/assets', express.static(path.join(__dirname, '/node_modules/govuk-frontend/govuk/assets')))
 ```
 
-### Copy the assets into your project folder
+### Copy the font and image files into your application
 
 If you decide to copy the assets instead, copy the:
 
 - `/node_modules/govuk-frontend/govuk/assets/images` folder to `<YOUR-APP>/assets/images`
--  `/node_modules/govuk-frontend/govuk/assets/fonts` folder to `<YOUR-APP>/assets/fonts`
+- `/node_modules/govuk-frontend/govuk/assets/fonts` folder to `<YOUR-APP>/assets/fonts`
 
 You should use an automated task or your build pipeline to copy the files, so your project folder stays up to date when we update GOV.UK Frontend.
 
 #### If you have your own folder structure
 
-If you copy the asset files but use a different folder structure than `<YOUR-APP>/assets/images` and `<YOUR-APP>/assets/images`, you can set Sass variables to make Sass build the CSS to point to your folders.
+If you use a different folder structure than `<YOUR-APP>/assets/images` and `<YOUR-APP>/assets/images`, you can set Sass variables so that Sass builds the CSS to point to your folders.
 
 Set one of the following before the `@import` line in your Sass file:
 
@@ -88,7 +88,7 @@ $govuk-images-path: "/<YOUR-IMAGES-FOLDER>/";
 $govuk-fonts-path: "/<YOUR-FONTS-FOLDER>/";
 ```
 
-You can also use your own function to generate paths, for example if you're using `asset-pipeline` in [sass-rails](https://github.com/rails/sass-rails). Set the `$govuk-image-url-function` variable to the name of your function.
+You can also use your own function to generate paths, for example if you're using `asset-pipeline` in [sass-rails](https://github.com/rails/sass-rails). Set the `$govuk-image-url-function` and `$govuk-font-url-function` variables to the name of your function.
 
 ## JavaScript
 
@@ -101,10 +101,10 @@ To import the JavaScript from GOV.UK Frontend, you can either:
 
 If you decide to add the JavaScript to your HTML, first either:
 
-- route requests to the `node_modules/govuk-frontend/govuk/all.js` file
-- copy the `node_modules/govuk-frontend/govuk/all.js` file into your project folder
+- set up your routing so that requests for the JavaScript file are served from  `node_modules/govuk-frontend/govuk/all.js`
+- copy the `node_modules/govuk-frontend/govuk/all.js` file into your application
 
-Then import GOV.UK Frontend's Javascript file before the closing `</body>` tag of your HTML page or page template, and run the `initAll` function to initialise all the components.
+Then import the JavaScript file before the closing `</body>` tag of your HTML page or page template, and run the `initAll` function to initialise all the components.
 
 ```html
 <body>
--- a/docs/installation/importing-css-assets-and-javascript.md
+++ b/docs/installation/importing-css-assets-and-javascript.md
@@ -81,7 +81,7 @@ Set the `$govuk-assets-path` variable if your `font` and `image` folders have th
 $govuk-assets-path: '/<YOUR-ASSETS-FOLDER>';
 ```
 
-Set the the `$govuk-images-path` and `$govuk-fonts-path` variables if your `font` and `image` folders have different parent folders. For example:
+Set the `$govuk-images-path` and `$govuk-fonts-path` variables if your `font` and `image` folders have different parent folders. For example:
 
 ```SCSS
 $govuk-images-path: "/<YOUR-IMAGES-FOLDER>/";
--- a/docs/installation/importing-css-assets-and-javascript.md
+++ b/docs/installation/importing-css-assets-and-javascript.md
@@ -1,0 +1,162 @@
+# Import CSS, assets and JavaScript
+
+## CSS
+
+### Import all the CSS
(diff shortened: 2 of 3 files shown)


docs/installation/installing-with-npm.md
```diff
@@ -1,303 +1,196 @@
...
 
-   (We recommend using [`nvm`](https://github.com/creationix/nvm) for managing
-   versions of Node.)
+2. `cd` to the root of your project and check if you have a [`package.json` file](https://docs.npmjs.com/files/package.json). If you do not have the file, create it by running:
 
-2. Create a [package.json file](https://docs.npmjs.com/files/package.json) if
-   you don’t already have one. You can create a default `package.json` file by
-   running `npm init` from the root of your application.
+    ```
+    npm init -y
```
nickcolley: Can you explain the decision to say `-y`, this will mean they accept a code licence that they may not agree with.

docs/installation/installing-with-npm.md
```diff
@@ -1,303 +1,196 @@
...
-
-To initialise the first radio component on a page, use:
-
-```js
-var $radio = document.querySelector('[data-module="govuk-radios"]')
-if ($radio) {
-  new Radios($radio).init()
-}
+```html
+...
```
nickcolley: This section confuses the scope feature with individual component initialisation, they should be separately explained as they solve different issues and are different features.

I don't think this example would work as given.

Scope is used to determine a part of the page to look for components to initialise, for example if you've added some new markup that needs to be initialised.

docs/installation/installing-with-npm.md
```diff
@@ -1,303 +1,196 @@
...
+```javascript
+const GOVUKFrontend = require('govuk-frontend')
+GOVUKFrontend.initAll()
+```
 
-## Include CSS and JavaScript
+### 5. Check it works
 
-Add the CSS and JavaScript code to your HTML template:
+Create an HTML file with the following code:
```
nickcolley: Should this be 'a HTML file'?

docs/installation/installing-with-npm.md
```diff
@@ -1,303 +1,196 @@
...
-    <link rel="stylesheet" href="assets/application.css">
+    <title>Example title</title>
+    <link rel="stylesheet" href="/{PATH-TO-CSS}/application.css">
   </head>
   <body>
-    <!-- Copy and paste component HTML-->
-    <button class="govuk-button">This is a button component</button>
-    <script src="assets/application.js"></script>
+    <button class="govuk-button">Example button</button>
+    <script src="/{PATH-TO-JAVASCRIPT}/all.js"></script>
```
nickcolley: Should we consider using `main.js` or `application.js` instead of `all.js` so we avoid potential confusion with govuk-frontend/govuk/all.js and their application file

docs/installation/installing-with-npm.md
m-green: If I understand right, this was to avoid people being presented with all the package.json questions that aren't very relevant for them and potentially hard to answer/explain. @36degrees - have I got that right?

docs/installation/installing-with-npm.md
```diff
@@ -1,303 +1,196 @@
...
-After resolving the import paths you can import GOV.UK Frontend by using:
+You can also import individual components. For example to import the [button](https://design-system.service.gov.uk/components/button/) component:
 
 ```SCSS
-@import "govuk-frontend/govuk/components/button/button";
+@import "node_modules/govuk-frontend/govuk/components/button/button";
 ```
 
-### Global Styles
+You can also [use styles from our old frameworks](https://github.com/alphagov/govuk-frontend/blob/master/docs/installation/compatibility.md) GOV.UK Frontend Toolkit, GOV.UK Template and GOV.UK Elements.
```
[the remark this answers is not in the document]
m-green: Yep that's right. This is my attempt to keep the [compatibility mode section at the bottom of the 'Importing styles' section](https://github.com/alphagov/govuk-frontend/blob/master/docs/installation/installing-with-npm.md#using-govuk-frontend-with-old-frameworks), but make it clearer why it's part of 'Importing styles'.

Maybe something like: "You can also configure GOV.UK Frontend to [use the colours from our legacy frameworks](https://github.com/alphagov/govuk-frontend/blob/master/docs/installation/compatibility.md)"?

docs/installation/installing-with-npm.md
```diff
@@ -1,303 +1,196 @@
...
+If your `image` and `font` folders are together in the same folder, set the `$govuk-assets-path` variable to your folder. For example:
 
-```js
-var Radios = window.GOVUKFrontend.Radios
-var $radio = document.querySelector('[data-module="govuk-radios"]')
-if ($radio) {
-  new Radios($radio).init()
-}
+```SCSS
+$govuk-assets-path: '/{YOUR-FOLDER}';
```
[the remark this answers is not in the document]
m-green: Thanks for spotting those quotes! I'll tweak the placeholder style too.

docs/installation/installing-with-npm.md
```diff
@@ -1,303 +1,196 @@
...
-GOVUKFrontend.initAll()
-```
+You can either:
 
-#### Import individual components
+- add GOV.UK Frontend's JavaScript file to your HTML
+- import the JavaScript using a bundler like [Webpack](https://webpack.js.org/)
 
-If you're using a bundler such as Webpack, use the `import` syntax to import a component:
+You must also initialise GOV.UK Frontend's components by calling the `initAll` function. If you do not initialise, some components will not work correctly for users who have accessible needs or use older browsers.
```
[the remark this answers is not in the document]
m-green: Sorry, this sentence definitely needs improving. Your suggestion is great, thanks!

docs/installation/installing-with-npm.md
m-green: Ah yes, sorry - it seems like I've merged two different sections into one somehow. I'll rework.

docs/installation/installing-with-npm.md
m-green: It can look odd but "an HTML" is correct. It's because 'an' is used when the next word begins with a vowel *sound* ('*a*itch'). It's the same reason we say "a user" (starts with a 'y' consonant sound), not "an user".

docs/installation/installing-with-npm.md
m-green: The docs further up (and the current docs) recommend users copy or route to the `all.js` file, so I'm not sure changing the name would work? Unless we tweaked the docs further up to recommend changing the filename? Or add `{YOUR-PROJECT-PATH}` to the hrefs in the example? @36degrees - what do you think?

docs/installation/installing-with-npm.md
nickcolley: People should make the decision what licence their code is in as it has legal implications and can impact how the code is shared.

docs/installation/installing-with-npm.md
```diff
@@ -1,303 +1,196 @@
...
 
-initAll()
+```SCSS
+$govuk-images-path: "/{YOUR-IMAGES-FOLDER}/";
+$govuk-fonts-path: "/{YOUR-FONTS-FOLDER}/";
+@import "govuk-frontend/govuk/all";
 ```
 
-If you're using a bundler such as [Browserify](http://browserify.org/), you may need to use the CommonJS `require`:
+You can also use your own function to generate paths, for example if you're using `asset-pipeline` in [sass-rails](https://github.com/rails/sass-rails). Set the `$govuk-image-url-function` variable to the name of your function.
```
hannalaakso: I wonder if we should have an example of how to do this? Since we've now removed:
> See `src/govuk/settings/_assets.scss` for more information and examples.

docs/installation/installing-with-npm.md
```diff
@@ -1,303 +1,196 @@
...
-})
-```
+- `/node_modules/govuk-frontend/govuk/assets/images`
+- `/node_modules/govuk-frontend/govuk/assets/fonts`
 
-#### Initialise individual included components
+You should use an automated task or your build pipeline to copy the files, so your project folder stays up to date when we update GOV.UK Frontend.
 
-GOV.UK Frontend components with JavaScript behaviour have the `data-module` attribute set in their markup.
+#### If you use different folders
```
hannalaakso: I think that this section is a subsection [Copy files](https://github.com/alphagov/govuk-frontend/pull/1697/files#diff-0eb006911fb60032aa38031190ad4869R80).

docs/installation/installing-with-npm.md
```diff
@@ -1,303 +1,196 @@
...
 ```
 
-### Polyfills
-A JavaScript polyfill provides functionality on older browsers or assistive technology that do not natively support it.
-
-The polyfills provided with GOV.UK Frontend aim to fix usability and accessibility issues. If there is a JavaScript included in the component directory, it is important to import and initialise it in your project to ensure that all users can properly use the component (see [Polyfilling](/docs/contributing/polyfilling.md)).
+If you add markup to a page after it's loaded, you can also use the `data-module` attribute to select specific components. For example, you can select radios using their `data-module` attribute of `govuk-radios`.
 
-### How GOV.UK Frontend is bundled
-The JavaScript included in GOV.UK Frontend components are in [UMD (Universal Module Definition)](https://github.com/umdjs/umd) format which makes it compatible with AMD (Asynchronous module definition) and CommonJS.
```
hannalaakso: Would it be worth saying this somewhere else as it's deleted from here? Some users might find this useful when deciding which bundler to use.

docs/installation/installing-with-npm.md
```diff
@@ -1,303 +1,196 @@
...
+#### Import JavaScript using a bundler
 
-See [JavaScript Coding Standards](/docs/contributing/coding-standards/js.md) for more details of how JavaScript is used in the project.
+Use `import` to import all of GOV.UK Frontend's components, then run the `initAll` function to initialise them:
 
-#### Using GOV.UK Frontend with Webpack 4
-Here's an example of setting up [`webpack.config.js`](examples/webpack/webpack.config.js) in your project
```
hannalaakso: I think this is the only place from where we link the Webpack config file. Should we mention it somewhere else instead?

docs/installation/installing-with-npm.md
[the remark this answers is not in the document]
m-green: Thanks - I'll add the suggested first line. We should probably cover the second line in the compatibility mode docs themselves I think?

docs/installation/installing-with-npm.md
m-green: Ah good point, thanks. This takes us down to an h5, which is probably getting a bit much... may need to rejig after research.

docs/installation/installing-with-npm.md
```diff
@@ -1,303 +1,196 @@
...
+- `/node_modules/govuk-frontend/govuk/assets/fonts`
 
-#### Initialise individual included components
+You should use an automated task or your build pipeline to copy the files, so your project folder stays up to date when we update GOV.UK Frontend.
 
-GOV.UK Frontend components with JavaScript behaviour have the `data-module` attribute set in their markup.
+#### If you use different folders
 
-You can use this attribute to initialise the component manually. This may be useful if you are adding markup to a page after it has loaded.
+If you use different folders for images and the font, you can override GOV.UK Frontend's default paths with your paths.
```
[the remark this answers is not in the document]
m-green: I've had a look and I think that with the existing bullet list above this, plus the 2 examples below, users should have enough context to understand this. But we could see how things pan out in research?

docs/installation/installing-with-npm.md
m-green: I wonder whether we need an example - the description seems fairly clear on its own, and I'm keen to avoid going too deep into advice about specific frameworks (at least until we've seen from research whether that's a major need). Could we leave as is and see if Rails users hit issues?

docs/installation/installing-with-npm.md
```diff
@@ -1,303 +1,196 @@
...
-2. Create a [package.json file](https://docs.npmjs.com/files/package.json) if
-   you don’t already have one. You can create a default `package.json` file by
-   running `npm init` from the root of your application.
+    ```
+    npm init -y
+    ```
 
-3. If you want to use the GOV.UK Frontend Nunjucks macros, install Nunjucks -
-   the minimum version required is 3.0.0.
+3. [Install Nunjucks v3.0.0 or later](https://www.npmjs.com/package/nunjucks), if you want to [use GOV.UK Frontend's Nunjucks macros](https://design-system.service.gov.uk/get-started/production/#using-nunjucks-macros).
```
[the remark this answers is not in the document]
m-green: Happy to add this as an optional step. I'll try moving it out of the list and to the bottom, and tweak to start with "You can also..." to make it's clearer it's optional. If our research users think it's still compulsory, we can look at adding "optional" more explicitly?

docs/installation/installing-with-npm.md
m-green: I spoke to @nickcolley about this and we did keep a version of it at one point, but if I remember right we chose to remove it?

docs/installation/installing-with-npm.md
m-green: This was kept at one stage, but @36degrees felt it was best to remove it during an earlier review.

docs/installation/installing-with-npm.md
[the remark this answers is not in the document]
nickcolley: This is a good point, it may make people think they can't use frontend because of Nunjucks

docs/installation/installing-with-npm.md
```diff
@@ -1,303 +1,196 @@
...
 
-After resolving the import paths you can import GOV.UK Frontend by using:
+You can also import individual components. For example to import the [button](https://design-system.service.gov.uk/components/button/) component:
 
 ```SCSS
-@import "govuk-frontend/govuk/components/button/button";
+@import "node_modules/govuk-frontend/govuk/components/button/button";
 ```
 
-### Global Styles
```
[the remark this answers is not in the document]
nickcolley: Would be good to have this documented somewhere but since it's not something we recommend feels good to leave out the main install journey.

docs/installation/installing-with-npm.md
36degrees: Yeah, I hadn't twigged that one of the questions was about the license – that probably shouldn't be something that gets accepted automatically. Good spot, @nickcolley. I think we probably need to revert this and provide guidance around how to answer the questions that they're asked (which I guess will include providing _something_ to help people understand how to answer the licensing question… 😬)

Do package.json files require a license field? Is it something users can skip without specifying?

docs/installation/installing-with-npm.md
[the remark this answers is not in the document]
hannalaakso: The examples below don't show any file paths and the two ones above start with `/node_modules` so it might not occur to the user that they're meant to be taking the subdirectory structure from those and reflecting the structure in their project. 

But if you feel confident about it as is then yes keep an eye on it in research.

docs/installation/installing-with-npm.md
hannalaakso: Okay that sounds good 👍

docs/installation/installing-with-npm.md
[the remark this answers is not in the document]
36degrees: Alternatively, could we front load the if statement?

```suggestion
3. If you want to [use GOV.UK Frontend's Nunjucks macros](https://design-system.service.gov.uk/get-started/production/#using-nunjucks-macros), [install Nunjucks v3.0.0 or later](https://www.npmjs.com/package/nunjucks).
```

docs/installation/installing-with-npm.md
nickcolley: We might want to just say `npm init` and link to the official guidance: https://docs.npmjs.com/creating-a-package-json-file

docs/installation/installing-with-npm.md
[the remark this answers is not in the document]
m-green: In an imminent commit, I've tried it in the way I mentioned in https://github.com/alphagov/govuk-frontend/pull/1697/files#r369471500 - @36degrees, take a look and let me know if you prefer the way you've suggested above (ie keep it as optional step 3).

m-green:
Hi all - I've updated with a couple of commits that I think address all of the comments so far. Let me know what you think!

docs/installation/installing-with-npm.md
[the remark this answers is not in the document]
hannalaakso: Would we consider having an additional section `### Optional requirements` for the Nunjucks sentence? I'm not sure that removing the numbering from it makes it clear that it's an optional step. There's nothing in the sentence that suggests that. If I didn't know what Nunjucks or macros were I might get stuck on this sentence, trying to figure out what it means and whether I need to do anything.

docs/installation/installing-with-npm.md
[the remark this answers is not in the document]
hannalaakso: I'm still having doubts about this one, sorry to come back to it. Could we at least say:
> If you use folders for images and the font that are different from the above ones, ....

I don't think it's necessarily clear what it's meant by "different" here. My first thought was "Different from what?" It's only because I know what we talk about here that I was able to figure out it's talking about the folder paths above.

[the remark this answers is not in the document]
m-green: @hannalaakso From what we've seen in the pilot version of the research, I think you're probably right, so I'm planning some tweaks based on your thoughts. But I'll see how the rest of the research goes so we can use that to get the final wording right.

docs/installation/installing-with-npm.md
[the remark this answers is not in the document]
m-green: Definitely a good option! We didn't see any issues with the separate-paragraph approach in the pilot version of the research - but let's keep an eye on it in research, and we can use that extra heading if developers miss that it's optional

m-green:
Adding the 'blocked' label while we research this version.

m-green:
In https://github.com/alphagov/govuk-frontend/pull/1697/commits/0f5ba783af8ea349116735987b593ff402253dcf, I've done the rejig I presented last week, which is based on the findings from our research. I've:

- split the install guide into 1) install, 2) get started and 3) importing
- kept things as simple and linear as possible in 'Get started', but linked to other options in the 'Importing' page
- used a 'try-then-test' approach for 'Get started', to better match the user behaviour we saw
- tweaked language in headings and copy to help convey that steps are not optional

Temporarily, I've put all the pages in the 'installation/' folder. This isn't necessarily a perfect structure - but as mentioned above, we should hopefully be able to go with 'Plan A' of moving these docs into the Tech Docs Template, rather than publish these docs here.

I haven't yet added any guidance about adding install lines in different places if you're migrating.

docs/installation/importing-css-assets-and-javascript.md
```diff
@@ -0,0 +1,162 @@
...
+## Font and image assets
+
+To use the font and image assets from GOV.UK Frontend, you can either:
+
+- route requests to the GOV.UK Frontend package - recommended
+- copy the font and images into your application
+
+### Route requests - recommended
+
+Route requests from `<YOUR-SITE-URL>/assets` to `/node_modules/govuk-frontend/govuk/assets`.
```
36degrees: I think this section could still do with improving…

- I'm not sure that people understand what we mean by 'route requests' – even with Express, which is the example we're providing, that's not terminology they use on https://expressjs.com/en/starter/static-files.html
- If the user is using webpacker then the best solution doesn't look like either of these (as I understand it, they'd want to use `require.context` and override `$govuk-font-url-function` and `$govuk-image-url-function`)

This could be something to iterate on in the future though.

docs/installation/importing-css-assets-and-javascript.md
```diff
@@ -0,0 +1,162 @@
...
+If you decide to copy the assets instead, copy the:
+
+- `/node_modules/govuk-frontend/govuk/assets/images` folder to `<YOUR-APP>/assets/images`
+-  `/node_modules/govuk-frontend/govuk/assets/fonts` folder to `<YOUR-APP>/assets/fonts`
+
+You should use an automated task or your build pipeline to copy the files, so your project folder stays up to date when we update GOV.UK Frontend.
+
+#### If you have your own folder structure
+
+If you copy the asset files but use a different folder structure than `<YOUR-APP>/assets/images` and `<YOUR-APP>/assets/images`, you can set Sass variables to make Sass build the CSS to point to your folders.
```
36degrees: You could also 'route requests' and this would still be relevant, I believe.

```suggestion
If you use a different folder structure than `<YOUR-APP>/assets/images` and `<YOUR-APP>/assets/images`, you can set Sass variables to make Sass build the CSS to point to your folders.
```

docs/installation/importing-css-assets-and-javascript.md
```diff
@@ -0,0 +1,162 @@
...
+```
+
+Set the the `$govuk-images-path` and `$govuk-fonts-path` variables if your `font` and `image` folders have different parent folders. For example:
+
+```SCSS
+$govuk-images-path: "/<YOUR-IMAGES-FOLDER>/";
+$govuk-fonts-path: "/<YOUR-FONTS-FOLDER>/";
+```
+
+You can also use your own function to generate paths, for example if you're using `asset-pipeline` in [sass-rails](https://github.com/rails/sass-rails). Set the `$govuk-image-url-function` variable to the name of your function.
```
36degrees: `$govuk-font-url-function` as well.

docs/installation/installing-with-npm.md
```diff
@@ -1,303 +1,34 @@
...
-    <button class="govuk-button">This is a button component</button>
-    <script src="assets/application.js"></script>
-  </body>
-</html>
-```
+When the installation finishes, the `govuk-frontend` package will be inside your `node_modules` folder.
 
-If your service supports Internet Explorer 8, you will need to [generate and
-include a separate stylesheet](supporting-internet-explorer-8.md) as well.
+You can now [get started by adding one GOV.UK Frontend component](/docs/installation/get-started.md).
```
36degrees: Is there anything we can / should do here to encourage users to read this guidance and understand the things they need to do, rather than just blindly trying to hook it up and stopping as soon as they get the styles working?

docs/installation/get-started.md
```diff
@@ -0,0 +1,73 @@
...
+
+Get one GOV.UK Frontend component working in your application.
+
+You must [install GOV.UK Frontend using npm](/docs/installation/installing-with-npm.md) first. If you've installed from `dist`, get started with [a basic page](/docs/installation/installing-from-dist.md#2-include-resources) instead.
+
+## Add a component to your application
+
+Go to the [example accordion component](https://design-system.service.gov.uk/components/accordion/#accordion-example) on the GOV.UK Design System website, then copy the HTML.
+
+Paste the HTML into your application.
```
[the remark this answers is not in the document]
m-green: Good idea. Would "a page or layout in your application" work ok?

docs/installation/get-started.md
```diff
@@ -0,0 +1,73 @@
+# Get started
```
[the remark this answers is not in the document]
m-green: Good plan - will add!

docs/installation/importing-css-assets-and-javascript.md
```diff
@@ -0,0 +1,162 @@
+# Import CSS, assets and JavaScript
+
```
hannalaakso: Should we say that this page is more relevant for production sites and for ensuring that your assets stay up to date? It feels that the relationship of this page to the "Get started" isn't really clear. A user might feel a bit overwhelmed if they hit this first 😅 

We could also include a link to the "Get started" guidance? 

Possibly both of these things will be taken care of by the navigation in tech docs template.

docs/installation/importing-css-assets-and-javascript.md
```diff
@@ -0,0 +1,162 @@
...
+## CSS
+
+### Import all the CSS
+
+To import all the CSS styles from GOV.UK Frontend, add the following to your Sass file:
```
hannalaakso: Strictly speaking this imports SCSS which get compiled into CSS by the compiler. It might help to make this distinction so that user knows they'll be working with SCSS which they can override, extend etc. You can also import pure CSS files into SCSS.

docs/installation/get-started.md
```diff
@@ -0,0 +1,73 @@
...
+      <script src="<YOUR-JAVASCRIPT-FOLDER>/govuk.js"></script>
+      <script>
+        window.GOVUKFrontend.initAll()
+      </script>
+    </body>
+    ```
+
+4. Run your application and check you can show and hide sections of the accordion by selecting the buttons.
+
+There are also different ways you can [import GOV.UK Frontend's JavaScript](/docs/installation/importing-css-assets-and-javascript.md#javascript).
```
hannalaakso: We could suggest an onward journey from here? As a user I might expect these pages in the tech docs template to flow in a linear / tutorial like order where I think "I've completed this page, where do I go next?" 

But feel free to ignore this if it doesn't feel relevant.

docs/installation/get-started.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Get the CSS working
+
+1. Add the following line to the main Sass file in your project, so your Sass compiler adds all GOV.UK Frontend styles to your CSS file.
+
+    ```Scss
+    @import "node_modules/govuk-frontend/govuk/all";
+    ```
+
+2. Add your CSS file to your page layout. For example:
```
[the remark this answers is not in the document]
m-green: I've added "if you need to" - hoping this will be enough to prod users that they don't need to do this if the file's already included?

docs/installation/get-started.md
```diff
@@ -0,0 +1,73 @@
...
+
+3. Run your application and check that the accordion is rendering using styles that start with `govuk-`.
+
+There are also different ways you can [import GOV.UK Frontend's CSS](/docs/installation/importing-css-assets-and-javascript.md#css).
+
+## Get the font and images working
+
+Your component will not use the right font or images until you've added GOV.UK Frontend's assets to your application.
+
+1. Use an automated task or your build pipeline to copy the:
```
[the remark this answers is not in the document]
m-green: Good plan!

docs/installation/get-started.md
```diff
@@ -0,0 +1,73 @@
...
+    - `/node_modules/govuk-frontend/govuk/assets/images` folder to `<YOUR-APP>/assets/images`
+    - `/node_modules/govuk-frontend/govuk/assets/fonts` folder to `<YOUR-APP>/assets/fonts`
+
+2. Run your application, then use [the Fonts tab in Firefox Page Inspector](https://developer.mozilla.org/en-US/docs/Tools/Page_Inspector/How_to/Edit_fonts#The_Fonts_tab) to check the accordion is using the GDS Transport font.
+
+In your live application, we recommend [routing requests to `node-modules`](/docs/installation/importing-css-assets-and-javascript.md#font-and-image-assets) instead of copying files.
+
+## Get the JavaScript working
+
+If you do not initialise GOV.UK Frontend's JavaScript with `initAll`, some components will not work correctly for disabled users who use assistive technologies.
```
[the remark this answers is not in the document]
m-green: I've moved this to the end and made it a bit firmer - let me know what you think.

docs/installation/get-started.md
```diff
@@ -0,0 +1,73 @@
...
+
+1. Add the `.js-enabled` class to the opening `<body>` tag of your page template. For example:
```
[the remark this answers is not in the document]
m-green: Ah gotcha. Added!

docs/installation/get-started.md
```diff
@@ -0,0 +1,73 @@
...
+
+    ```JavaScript
+    document.body.className = ((document.body.className) ? document.body.className + ' js-enabled' : 'js-enabled');
+    ```
+
+2. Use an automated task or your build pipeline to copy the `/node-modules/govuk-frontend/govuk/all.js` file to `<YOUR-JAVASCRIPT-FOLDER>/govuk.js`
```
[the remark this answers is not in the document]
m-green: Good plan - tweaked!

docs/installation/importing-css-assets-and-javascript.md
m-green: Good point! Following our chat I've improved the content to address your first point. Let's iterate on the second bullet in future if that's ok?

docs/installation/importing-css-assets-and-javascript.md
m-green: Ah good to know, thanks!

docs/installation/importing-css-assets-and-javascript.md
m-green: 👍

docs/installation/installing-with-npm.md
m-green: Hmm good point. I've tweaked to "You should now get started by [getting the CSS, assets and JavaScript working](/docs/installation/get-started.md) with one GOV.UK Frontend component." - does that sound ok?

docs/installation/get-started.md
```diff
@@ -0,0 +1,73 @@
+# Get started
+
+Get one GOV.UK Frontend component working in your application.
```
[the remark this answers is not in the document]
m-green: I like it - added!

docs/installation/get-started.md
```diff
@@ -0,0 +1,73 @@
...
+
+You must [install GOV.UK Frontend using npm](/docs/installation/installing-with-npm.md) first. If you've installed from `dist`, get started with [a basic page](/docs/installation/installing-from-dist.md#2-include-resources) instead.
```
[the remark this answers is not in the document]
m-green: Very good point, and this was also mentioned by the tech writers earlier too. I've tweaked so it's very much step 1.

docs/installation/importing-css-assets-and-javascript.md
m-green: This page is trying to do a lot, it's true! It's also aiming to be something of a 'fix issues with CSS, assets and JavaScript' page. In a future iteration we need to tease the content apart a little (possibly into separate pages).

I think you're right that the navigation should hopefully help here. Get started will be nearer the top (and immediately under install) in the nav, so I hope users will be drawn back to Get started if this looks overwhelming. 

So would we be ok to leave as is for now, but aim to test a little in Q1?

docs/installation/importing-css-assets-and-javascript.md
m-green: Ah ok - I've tweaked to "import all the SCSS rules" - does that sound ok?

docs/installation/get-started.md
m-green: Ah very good point. I've borrowed a line we lost from the last iteration: "You can get the full code for page layouts and components from the [Design System website](https://design-system.service.gov.uk/)."

docs/installation/importing-css-assets-and-javascript.md
hannalaakso: Yes sounds good, thanks @m-green 👍

docs/installation/importing-css-assets-and-javascript.md
hannalaakso: 👍

docs/installation/get-started.md
hannalaakso: 👍

m-green:
@36degrees @hannalaakso Thank you for your great comments! I've pushed a new commit that hopefully addresses almost all of them. There are a couple of outstanding ones that I'll chat to you about tomorrow morning if that's ok?

docs/installation/importing-css-assets-and-javascript.md
36degrees: I love everything about this change.

docs/installation/get-started.md
```diff
@@ -0,0 +1,90 @@
...
+2. Add the HTML for a component to your application.
+3. Get the CSS working.
+4. Get the font and images working.
+5. Get the JavaScript working.
+
+## 1. Install
+
+[Install GOV.UK Frontend using npm](/docs/installation/installing-with-npm.md).
+
+If you've installed from `dist`, get started with [a basic page](/docs/installation/installing-from-dist.md#2-include-resources) instead.
```
hannalaakso: I'm not sure it's clear what this refers to. A user might look at this and wonder what `dist` is and whether it's something to do with their own set up and something they should know about, instead of realising that it's an alternative to the first sentence in this section.

We could say something like this (or make it clear in some other way that these are alternative options):

> If you've [installed with the compiled files](/docs/installation/installing-from-dist) instead of npm, get started with [a basic page](/docs/installation/installing-from-dist.md#2-include-resources) and skip step 3 in this guidance.

docs/installation/get-started.md
```diff
@@ -0,0 +1,90 @@
...
+Your component will not use the right font or images until you've added GOV.UK Frontend's assets to your application.
+
+1. Copy the:
+
+    - `/node_modules/govuk-frontend/govuk/assets/images` folder to `<YOUR-APP>/assets/images`
+    - `/node_modules/govuk-frontend/govuk/assets/fonts` folder to `<YOUR-APP>/assets/fonts`
+
+2. Run your application, then use [the Fonts tab in Firefox Page Inspector](https://developer.mozilla.org/en-US/docs/Tools/Page_Inspector/How_to/Edit_fonts#The_Fonts_tab) to check the accordion is using the GDS Transport font.
+
+In your live application, we recommend [using an automated task or your build pipeline](/docs/installation/importing-css-assets-and-javascript.md#font-and-image-assets) instead of copying the files manually.
```
hannalaakso: Do we mean "production application" by "live application"? If so, I would prefer that wording over the word "live" which we haven't really used in this context I don't think. But totally non-blocking.

docs/installation/get-started.md
m-green: Sorry @hannalaakso, I've just realised I didn't respond to your comments!

Definitely agree here (and as you saw it came up in the migration docs research too). It's something I'm planning to improve across our docs in Q4, so I'm tempted to leave this as is for now to avoid confusion by having inconsistency with how we talk about installing from dist/compiled files. But I'm wavering, so I may well use your suggestion while this is getting 2i'd.

docs/installation/get-started.md
m-green: Unfortunately 'production' (and 'prototyping') caused some confusion in the research, so I'd like to try a different term for now. I do appreciate that 'live' and 'production' don't quite mean the same thing - happy to take another look if users are confused by it after publication.

docs/installation/get-started.md
```diff
@@ -0,0 +1,90 @@
...
+    ```Scss
+    @import "node_modules/govuk-frontend/govuk/all";
+    ```
+
+2. Add your CSS file to your page layout if you need to. For example:
+
+    ```html
+      <head>
+        ...
+        <link rel="stylesheet" href="<YOUR-CSS-FILE>.css">
```
hannalaakso: Cross referencing a conversation from another PR: https://github.com/alphagov/govuk-frontend-docs/pull/2#discussion_r380771988 We should mention here that if they use the legacy repos, GOV.UK Frontend stylesheet should be included after the legacy files in their page template.

docs/installation/get-started.md
```diff
@@ -0,0 +1,90 @@
...
+
+## 5. Get the JavaScript working
+
+1. Add the following JavaScript to the top of the `<body>` section of your page template:
+
+    ```JavaScript
+    document.body.className = ((document.body.className) ? document.body.className + ' js-enabled' : 'js-enabled');
+    ```
+
+2. Copy the `/node-modules/govuk-frontend/govuk/all.js` file to `<YOUR-JAVASCRIPT-FOLDER>/govuk.js`.
```
hannalaakso: Cross referencing a conversation from another PR: https://github.com/alphagov/govuk-frontend-docs/pull/2#discussion_r380771988 We should mention here that if they use the legacy repos, GOV.UK Frontend script should be included after the legacy files in their page template.

docs/installation/get-started.md
m-green: I'm not sure we should mention this in Get started, as this doc is meant as a very basic 'get started from scratch' guide. If we started to talk about using legacy repos, we'd also need to mention compatibility mode etc.

Instead, I think it would work if we add this in both:

a) the migration docs (for users migrating)
b) the compatibility mode docs (for users running FE alongside legacy longer-term)

What do you think?

hannalaakso: Yes adding it to the migration docs would make sense thanks @m-green 👍

docs/installation/get-started.md
hannalaakso: We're doing this in the migration docs instead, see https://github.com/alphagov/govuk-frontend-docs/pull/2#discussion_r400326184

m-green:
Closing because this has now become part of the Tech Docs Template

Issue #1557, opened 2019-09-06
date-input: id is documented as optional but used as required to create hintId and errorId


dabd:
In the `date-input` component [`id` is documented as optional](https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/components/date-input/date-input.yaml#L4) but [treated as a required parameter when creating the `hintId`](https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/components/date-input/template.njk#L32) and [`errorId`](https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/components/date-input/template.njk#L43) (if it is not provided it will generate an id starting with hyphen, which is probably not the intended result).
It is treated as optional [elsewhere](https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/components/date-input/template.njk#L56).

Should it be considered as a required parameter instead?

In the `radios` component when `idPrefix` is not supplied [it falls back to using `name`](https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/components/radios/template.njk#L8).  Does it make sense to use the same approach here?

Same problem here: https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/components/tabs/template.njk#L3

StheffanyHadlich:
Can I help with this issue?

nickcolley:
Heya @StheffanyHadlich!

Thanks for offering to help out, we think this issue is more than just documentation so the GOV.UK Design System team will pick this up.

We have some first timer issues over on the GOV.UK Design System website repository if you're interested in contributing there: https://github.com/alphagov/govuk-design-system/issues?q=is%3Aopen+is%3Aissue+label%3A%22help+wanted%22

StheffanyHadlich:
Oh, thank you

kellylee-gds:
Review from @m-green

PR #1698, opened 2020-01-16
Fix date input documentation

nickcolley:
I think this should ideally be called `idPrefix` instead of `id` but as that would be a breaking change we should consider this in another change.

I'm not sure the potential confusion from inconsistency between components for users warrants making this breaking change.

Closes https://github.com/alphagov/govuk-frontend/issues/1557

### DIFF
--- a/src/govuk/components/date-input/date-input.yaml
+++ b/src/govuk/components/date-input/date-input.yaml
@@ -1,7 +1,7 @@
 params:
 - name: id
   type: string
-  required: false
+  required: true
   description: This is used for the main component and to compose id attribute for each item.
 - name: namePrefix
   type: string


CHANGELOG.md
```diff
@@ -12,7 +12,8 @@ If you're using Nunjucks, you can now add classes to the character count compone
 
 ### Fixes
 
-- [Pull request #1690: Don't unneccesarily self-close tags](https://github.com/alphagov/govuk-frontend/pull/1690)
+- [Pull request #1698: Fix documentation so that `id` option in date input is required](https://github.com/alphagov/govuk-frontend/pull/1698)
```
36degrees: Trying to work out if this is useful to add to the release notes for Frontend – it feels like it might be an internal thing, as the value is ultimately 'delivered' in the Design System? Thoughts? /cc @m-green

m-green: It seems best to leave it out of the release note - because we're not changing how Frontend behaves, we're just updating the documentation to match the reality.

nickcolley: Yeah we don't update the CHANGELOG for documentation, my mistake

PR #1699, opened 2020-01-16
Remove unused id date input

nickcolley:
We link to the inputs from the error summary so I think this attribute is not used for anything.

For reviewers:

Am I right in thinking this isn't being used for anything?

Is this a breaking change? My gut feeling is no if this was not doing anything but wonder what your thoughts are...

Related https://github.com/alphagov/govuk-frontend/issues/1557

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ If you're using Nunjucks, you can now add classes to the character count compone
 
 ### Fixes
 
+- [Pull request #1699: Remove unused id in date input component wrapper](https://github.com/alphagov/govuk-frontend/pull/1699)
 - [Pull request #1690: Don't unneccesarily self-close tags](https://github.com/alphagov/govuk-frontend/pull/1690)
 - [Pull request #1678: Fix tabs component throwing JavaScript errors in Internet Explorer 8](https://github.com/alphagov/govuk-frontend/pull/1678).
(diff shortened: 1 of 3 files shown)


36degrees:
We use params.id to create the IDs for the hint and error message, so I _think_ it makes sense to keep it and to mark it as required as described in #1557?

https://github.com/alphagov/govuk-frontend/blob/b3741f611e364788551b726bbf7ceb49232dc578/src/govuk/components/date-input/template.njk#L32

https://github.com/alphagov/govuk-frontend/blob/b3741f611e364788551b726bbf7ceb49232dc578/src/govuk/components/date-input/template.njk#L43

We also use it to generate IDs for each input, if they're not otherwise provided:

https://github.com/alphagov/govuk-frontend/blob/b3741f611e364788551b726bbf7ceb49232dc578/src/govuk/components/date-input/template.njk#L63

I think if we did choose to remove it it would be a breaking change – if anyone's using that ID to hook into their own application javascript, or even just using it as a selector in tests, the ID no longer being present could definitely break things.

36degrees:
Sorry, having now reviewed https://github.com/alphagov/govuk-frontend/pull/1698 I guess the intention was to keep `id` around but just not to use it directly.

I still think it makes sense to keep this – in retrospect, `idPrefix` may have been more appropriate, but whilst it's called `id` it seems strange (and inconsistent) not to actually use it as an `id` anywhere.

I hope that at some point we'll revisit the date input component entirely and we can look at addressing issues like this then?

nickcolley:
Yeah maybe not worth the breaking change at this point then.

PR #1703, opened 2020-01-20
Update error summary macro options descriptions

frankieroberto:
Neither `descriptionText` nor `descriptionHtml` are required, so this updates the documentation to make this clearer.

### DIFF
--- a/src/govuk/components/error-summary/error-summary.yaml
+++ b/src/govuk/components/error-summary/error-summary.yaml
@@ -10,11 +10,11 @@ params:
 - name: descriptionText
   type: string
   required: false
-  description: Text to use for the description of the errors. If `descriptionHtml` is provided,  `descriptionText` will be ignored.
+  description: Text to use for the description of the errors. If you set `descriptionHtml`, the component will ignore `descriptionText`.
 - name: descriptionHtml
   type: string
   required: false
-  description: HTML to use for the description of the errors. If `descriptionHtml` is provided, `descriptionText` will be ignored.
+  description: HTML to use for the description of the errors. If you set this option, the component will ignore `descriptionText`.
 
 - name: errorList
   type: array
--- a/src/govuk/components/error-summary/error-summary.yaml
+++ b/src/govuk/components/error-summary/error-summary.yaml
@@ -10,11 +10,11 @@ params:
 - name: descriptionText
   type: string
   required: false
-  description: If `descriptionHtml` is set, this is not required. Text to use for the description of the errors. If `descriptionHtml` is provided,  `descriptionText` will be ignored.
+  description: Text to use for the description of the errors. If `descriptionHtml` is provided,  `descriptionText` will be ignored.
 - name: descriptionHtml
   type: string
-  required: true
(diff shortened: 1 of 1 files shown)


nickcolley:
I think... that these should mirror the `text` / `html` options common in components. In those components they're both required.

https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/components/button/button.yaml#L6-L13

Perhaps worth us looking how we describe options with `text` / `html` consistently.

frankieroberto:
@nickcolley I think the button component requires _either_ the `text` or `html` param though, whereas here it is perfectly valid for neither to be present?

nickcolley:
@frankieroberto yeah I think you're right there, good point.

src/govuk/components/error-summary/error-summary.yaml
```diff
@@ -10,11 +10,11 @@ params:
 - name: descriptionText
   type: string
   required: false
-  description: If `descriptionHtml` is set, this is not required. Text to use for the description of the errors. If `descriptionHtml` is provided,  `descriptionText` will be ignored.
+  description: Text to use for the description of the errors. If `descriptionHtml` is provided,  `descriptionText` will be ignored.
```
m-green:
```suggestion
  description: Text to use for the description of the errors. If you set `descriptionHtml` instead, `descriptionText` will be ignored.
```

Slight tweak to avoid passive voice (and fix the double space).

m-green:
```suggestion
  description: Text to use for the description of the errors. If you set `descriptionHtml`, the component will ignore `descriptionText`.
```
Slight tweak to avoid passive voice (and fix the double space).

src/govuk/components/error-summary/error-summary.yaml
```diff
@@ -10,11 +10,11 @@ params:
...
 - name: descriptionHtml
   type: string
-  required: true
-  description: If `descriptionText` is set, this is not required. HTML to use for the description of the errors. If `descriptionHtml` is provided, `descriptionText` will be ignored.
+  required: false
+  description: HTML to use for the description of the errors. If `descriptionHtml` is provided, `descriptionText` will be ignored.
```
m-green:
```suggestion
  description: HTML to use for the description of the errors. If you set `descriptionText`, the component will ignore `descriptionHTML`.
```
Same tweak as above. I think the original also accidentally duplicated the `descriptionText` descripton - I've reversed it to say that if you use *this* variable, the other will be ignored.

36degrees: I think the original was correct. If you've passed both `descriptionHTML` and `descriptionText`, `descriptionHTML` will always 'win' and `descriptionText` will always be ignored.

m-green: Ah sorry - I've tweaked my suggestion accordingly.

frankieroberto:
@m-green thanks!

36degrees:
Thanks @frankieroberto!

PR #1711, opened 2020-01-24
Additional Tag modifier classes for different colours

adamsilver:
I spoke with @36degrees and then @timpaul about contributing additional Tag component styles via modifier classes. A lot of services (citizen-facing and internal) use the Tag component for different statuses.

Most of the time, different colours are used as an enhancement to spot those different states easily. Most of the designs I've seen put white text on coloured backgrounds. At DfE we did the same but they fail contrast requirements.

So we now uses these styles which are accessible and use the tints and shades of the GOV.UK colour palette.

More information here:
https://github.com/alphagov/govuk-design-system-backlog/issues/62#issuecomment-570575887

Note:

1. I would usually look for meaningful names for modifier classes like ‘urgent’ but statuses are often unique to the system and so I used colours for the class names giving users flexibility.

2. I still need to add guidance / examples to the design system repo. I'm not sure what the process is on that i.e. do I wait for this to be merged or something more elaborate?

Screenshot:

image

### DIFF
--- a/src/govuk/components/tag/_tag.scss
+++ b/src/govuk/components/tag/_tag.scss
@@ -41,4 +41,50 @@
   .govuk-tag--inactive {
     background-color: govuk-colour("dark-grey", $legacy: "grey-1");
   }
+
+  .govuk-tag--grey {
+    color: govuk-shade(govuk-colour("dark-grey", $legacy: "grey-1"), 30);
+    background: govuk-tint(govuk-colour("dark-grey", $legacy: "grey-1"), 90);
+  }
+
+  .govuk-tag--purple {
+    color: govuk-shade(govuk-colour("purple"), 20);
+    background: govuk-tint(govuk-colour("purple"), 80);
+  }
+
+  .govuk-tag--turquoise {
+    color: govuk-shade(govuk-colour("turquoise"), 60);
+    background: govuk-tint(govuk-colour("turquoise"), 70);
+  }
+
+  .govuk-tag--blue {
+    color: govuk-shade(govuk-colour("blue"), 30);
+    background: govuk-tint(govuk-colour("blue"), 80);
+  }
+
+  .govuk-tag--yellow {
+    color: govuk-shade(govuk-colour("yellow"), 65);
+    background: govuk-tint(govuk-colour("yellow"), 75);
+  }
+
+  .govuk-tag--orange {
+    color: govuk-shade(govuk-colour("orange"), 55);
+    background: govuk-tint(govuk-colour("orange"), 70);
+  }
+
+  .govuk-tag--red {
+    color: govuk-shade(govuk-colour("red"), 30);
+    background: govuk-tint(govuk-colour("red"), 80);
+  }
+
+  .govuk-tag--pink {
+    color: govuk-shade(govuk-colour("pink"), 40);
+    background: govuk-tint(govuk-colour("pink"), 80);
+  }
+
+  .govuk-tag--green {
+    color: govuk-shade(govuk-colour("green"), 20);
+    background: govuk-tint(govuk-colour("green"), 80);
+  }
+
 }
--- a/src/govuk/components/tag/_tag.scss
+++ b/src/govuk/components/tag/_tag.scss
@@ -38,6 +38,7 @@
     }
   }
 
+  // Deprecated. We'll remove this class in a future release. Use `.govuk-tag--grey` instead.
   .govuk-tag--inactive {
     background-color: govuk-colour("dark-grey", $legacy: "grey-1");
   }
--- a/src/govuk/components/tag/tag.yaml
+++ b/src/govuk/components/tag/tag.yaml
@@ -24,3 +24,39 @@ examples:
    data:
     text: alpha
     classes: govuk-tag--inactive
+ - name: grey
+   data:
+    text: Grey
+    classes: govuk-tag--grey
+ - name: blue
+   data:
+    text: Blue
+    classes: govuk-tag--blue
+ - name: turquoise
+   data:
+    text: Turquoise
+    classes: govuk-tag--turquoise
+ - name: green
+   data:
+    text: Green
+    classes: govuk-tag--green
+ - name: purple
+   data:
+    text: Purple
+    classes: govuk-tag--purple
+ - name: pink
+   data:
+    text: Pink
+    classes: govuk-tag--pink
+ - name: red
+   data:
+    text: Red
+    classes: govuk-tag--red
+ - name: orange
+   data:
+    text: Orange
+    classes: govuk-tag--orange
+ - name: yellow
+   data:
+    text: Yellow
+    classes: govuk-tag--yellow


src/govuk/components/tag/template.test.js
```diff
@@ -75,4 +75,94 @@ describe('Tag', () => {
     expect($component.attr('data-test')).toEqual('attribute')
     expect($component.attr('id')).toEqual('my-tag')
   })
+
+  it('grey example passes accessibility tests', async () => {
```
36degrees: The `render` function just returns the HTML from the macro itself, without the layout that brings in the stylesheet. Unfortunately this means there's little benefit to running this for every example, as it won't be checking the contrast between the colours that will actually be used.

src/govuk/components/tag/_tag.scss
```diff
@@ -20,6 +20,8 @@
     text-decoration: none;
     text-transform: uppercase;
 
+    white-space: nowrap;
```
36degrees: This change seems unrelated to the commit description – if it's required, can we make this change in another commit that explains why it's being added, or possibly make it part of a separate PR?

src/govuk/components/tag/_tag.scss
```diff
@@ -41,4 +43,50 @@
   .govuk-tag--inactive {
     background-color: govuk-colour("dark-grey", $legacy: "grey-1");
   }
+
+  .govuk-tag--grey {
```
36degrees: We should consider whether there's a need for `inactive` and `grey`, or whether they can be the same thing.

src/govuk/components/tag/tag.yaml
```diff
@@ -24,3 +24,39 @@ examples:
    data:
     text: alpha
     classes: govuk-tag--inactive
+ - name: grey
```
36degrees: Thanks for adding these 👏

nickcolley:
I think it's worth carefully considering the modifier names, if we name them based on their colour and decide the colours in the future need to change they may not make sense anymore or require a breaking change?

Good ol' boostrap has this sort of thing with their 'badges' and they use broad semantic names which might be worth considering:
https://getbootstrap.com/docs/4.4/components/badge/#contextual-variations

src/govuk/components/tag/template.test.js
nickcolley: We've talked about running axe on the review application itself in the future, which would achieve what you're looking to do.

adamsilver: Shall I remove all of the tests I added @36degrees ?

src/govuk/components/tag/_tag.scss
adamsilver: Yeah. I considered deleting it, but I thought that would delay the contribution as I assumed it would be a breaking change.

36degrees: We could consider marking it as deprecated, for example.

src/govuk/components/tag/template.test.js
36degrees: I think so – unfortunately we don't have a good way to test this sort of thing right now.

adamsilver:
> I think it's worth carefully considering the modifier names, if we name them based on their colour and decide the colours in the future need to change they may not make sense anymore or require a breaking change?

@nickcolley 

Yeah—I did consider this as I said in the PR description. 

I think it would be too difficult to come up with useful names (I checked Bootstrap) and I also think it could mean code within a service could be confusing...

For example, let's say we call the orange variant ‘warning-2’ and the yellow being ‘warning-1’. (The reason for the numbers is because there aren't enough generic class names.) 

Let's imagine I use ‘warning-1’ because in Apply for teaching training we want to the ‘withdrawn’ status to be yellow. That would indicate “withdrawn’ is some sort of warning when it isn't.

src/govuk/components/tag/_tag.scss
adamsilver: Yeah, and actually, it's not documented on the Tag component page. Is it documented publically elsewhere?

nickcolley:
@adamsilver I think you might be right that it's too difficult to have semantics that can anticipate what context these are used in and the colours dont change often.

src/govuk/components/tag/_tag.scss
adamsilver: I've left it in, and maybe the deprecation can be a release note? Or do I need to do something in the code to mark it as deprecated?

nickcolley: Since it's not been documented we could consider dropping it in 4.0.0

timpaul:
> @adamsilver I think you might be right that it's too difficult to have semantics that can anticipate what context these are used in and the colours dont change often.

Good discussion! I'm comfortable with the colour-based class names, as they're aligned with the names in our colour palette. 

I totally agree that we can't anticipate all future semantic states that the tags will be used to represent. But we might converge on a few one day (eg. 'success', 'warning', 'fail' etc.) - does this approach block us from doing that in the future?

adamsilver:
> But we might converge on a few one day (eg. 'success', 'warning', 'fail' etc.) - does this approach block us from doing that in the future?—@timpaul 

I think that would be fine because we could just add an additional govuk-tag--success class name as a minor release.

peteryates:
I don't know if this has been brought up elsewhere but, to me at least, the red and orange look _very_ similar. To the point where I might not notice they were different. To a slightly lesser extent, the red and pink do too.

adamsilver:
@peteryates some of us at DfE thought the same thing, but we weren't overly concerned because:

1. Colour is only used as an enhancement
2. You don't have to use both colours if you don't want to
3. They may not appear right next to each in the interface

pripley-ea:
tag-colour-vision
I think these will be very useful and I like what you have done with matching the text colour with the background hue.

I've just run a quick test to see how a user with Protanopia would see them as shown above (red colour vision deficiency) and they are still usable as the text is clear but some of the colours cross over so the only thing is to be careful if multiple tags are used together. 

Really nice addition to the design system :)

src/govuk/components/tag/_tag.scss
hannalaakso: Created an issue for dropping it in 4.0.0: https://github.com/alphagov/govuk-frontend/issues/1757  

I'll chat to @m-green about whether we want to add a deprecation notice for `govuk-tag--inactive` somewhere, seeing that we don't document the class anywhere currently.

hannalaakso: @adamsilver Would you be okay to add something like
```
Deprecated. We will be removing this class in a future release, use `.govuk-tag--grey` instead.
```
above the `govuk-tag--inactive` class in the scss file? 🙏

m-green: No probs - added in https://github.com/alphagov/govuk-frontend/pull/1756/commits/ed8534b156ad53c32dd71240b81a12b243f747a5

adamsilver: @hannalaakso where would you like me to put this line?

adamsilver: @hannalaakso done.

hannalaakso:
@adamsilver Thanks for updating the PR. 

We're now thinking that https://github.com/alphagov/govuk-frontend/pull/1711/commits/1d79100f2db22cd6c83904624399a0c4f6b5d590 should be moved to its own PR as it could possibly be a breaking change as it could change your layout (esp on mobile) by forcing a longer tag to stay on its own line. We should talk about it separately. Apologies for the back and forth on this 🙁 

If you were able to remove the commit from this PR (let me know if you need help rebasing) we can get this one merged. Thank you!

adamsilver:
@hannalaakso thank you—I'd be delighted to take you up on your offer to do it for me pls :)

hannalaakso:
_On 13 February 2020 the [Design System working group approved](https://github.com/alphagov/govuk-design-system-backlog/issues/62#issuecomment-590800378) the addition of different coloured variants of this component._ 

(Just mentioning this for reference as I didn't see it explicitly mentioned in the comments on this PR.)

Issue #573, opened 2018-03-05
Components need print mode


nickcolley:
When simulating print mode on components most don't have the necessary adjustments.

<img width="1006" alt="screen shot 2018-03-05 at 13 53 50" src="https://user-images.githubusercontent.com/2445413/36978549-afbe3cba-207c-11e8-887d-528b02fdfcc5.png">

nickcolley:
@36degrees I think our users such as GOV.UK Publishing will run into this pretty soon.

nickcolley:
@36degrees I think our users such as GOV.UK Publishing will run into this pretty soon.

igloosi:
Trello ticket: https://trello.com/c/mlBn52Wg/1395-design-print-styles-for-components

andysellick:
I'm currently looking at print stylesheets for components using GOV.UK Frontend on GOV.UK, so this would be great 👍

injms:
✋ me too - I've put together something to make the [accordion component](https://github.com/alphagov/govuk_publishing_components/pull/714) readable when printed out, but it's got some not-so-nice `!important`s.

frankieroberto:
👍  on this.

When developing the accordion component, we had an interesting debate about whether, when printing a page, all of the accordions should be "opened" or not. My conclusion was that it probably made sense to keep them in the same state as the page, as the user may only want to print some of the sections (and if they really wanted to print them all then that’s fairly easy to do via the "Open all" button).

crosslandwa:
The HMRC PAYE (classic) service has recently switched to using the design system GOV UK. The application offers users the ability to view-and-print previous years submissions - we use the [check answers](https://design-system.service.gov.uk/patterns/check-answers/) pattern to "play back" submitted details and provide a (JS) print button to open a print dialogue _* note screenshots show test data*_:

Page screenshot:
<img width="378" alt="p11db-govuk-print-screenshot" src="https://user-images.githubusercontent.com/4037089/58697886-17d7e700-8392-11e9-8ea0-6f5eefc1c513.png">

Printing in chrome:
<img width="378" alt="p11d-print-screenshot" src="https://user-images.githubusercontent.com/4037089/58698467-3c808e80-8393-11e9-9d0a-738132ca5e91.png">

We've provided a form for users to provide feedback, and had several comments about printing. Here are some quotes:

- Can you make reports printer friendly please. The large bold letters are using plenty of ink!
[...]

PR #1723, opened 2020-02-04
Add display override for hiding content when printing

fofr:
When printing users often need to print the content without the surrounding site furniture (eg menus, breadcrumbs, back links and footers).

Introduce a helper so that a class can be used to hide any given element from the print view:
```
govuk-!-print-display-none
```

This is explicit and preferable to listing all the selectors for things that shouldn't be printed. These lists grow unwieldy over time. ([An example of one of these lists](https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training-prototype/blob/bdd9c0a51dc89a1ef8aca04b3faa2354227ebafd/app/assets/sass/application.scss#L32-L41))

(We've been using this pattern on [Apply for teacher training](https://www.apply-for-teacher-training.education.gov.uk/candidate) and [Becoming a teacher: Design History](https://bat-design-history.herokuapp.com/))

### DIFF
--- a/src/govuk/overrides/_display.scss
+++ b/src/govuk/overrides/_display.scss
@@ -19,4 +19,10 @@
   .govuk-\!-display-none {
     display: none !important;
   }
+
+  @include govuk-media-query($media-type: print) {
+    .govuk-\!-display-none-print {
+      display: none !important;
+    }
+  }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@
 
 ### New features
 
+#### Add display override for hiding content when printing
+
+Introduce a CSS class that can be used to hide any given element from the print view: `govuk-!-display-none-print`
+
+[Pull request #1723: Add display override for hiding content when printing](https://github.com/alphagov/govuk-frontend/pull/1723).
+
 ### Fixes
 
 ## 3.5.0 (Feature release)


nickcolley:
@fofr what do you think about components such as the footer having sensible print styles: https://github.com/alphagov/govuk-frontend/issues/573 ? Do you think you'd want full control over that with a helper instead of them being baked in?

fofr:
@nickcolley I think we need both.

Components should have good print styles out of the box. But sometimes it's not clear if a component should be printed or not – for instance a breadcrumb may sometimes be useful to print as it shows the hierarchy or context of some content, whereas at other times it's simply a navigational aid and can be hidden.

In our case we are printing an application and all its completed fields. Which means hiding a lot from the print view so that it's clean and appears similar to a completed paper form.

nickcolley:
We're happy to accept this, would you mind adding a feature CHANGELOG entry that introduces the new class? You can look at other feature CHANGELOG entries for examples.

I can do this on your behalf if you'd prefer that.

fofr:
I've added a changelog entry. Happy for you to amend if it's not quite right.

src/govuk/overrides/_display.scss
```diff
@@ -19,4 +19,10 @@
   .govuk-\!-display-none {
     display: none !important;
   }
+
+  @include govuk-media-query($media-type: print) {
+    .govuk-\!-print-display-none {
```
nickcolley: Sorry for this after approving things.

Can we be careful to consider this naming vs all the other overrides?

Perhaps this should be `.govuk-\!-display-none-print` as the format seems to be `.govuk-\!-<PROPERTY>-<VALUE/SCALE-MODIFIER>`

nickcolley: I've pinged the team on this...

fofr: We put print first so that it was more easily differentiated from `govuk-!-display-none`.

`print-display-none` matches the order in the stylesheet:
```
@media print {
  .blah {
    display: none
  }
}
```

nickcolley: @36degrees @hannalaakso thoughts?

36degrees: As far as I can see, we don't have any other overrides that are conditional on media queries. Probably the closest existing class would be e.g. `govuk-grid-column-one-third-from-desktop`, where the 'media query' part is last?

If we did go with `govuk-`!-display-none-print` that means it would also match if you searched your code base for `govuk-\!-display-none`, but that's also true for `govuk-grid-column-one-third-from-desktop` and `govuk-grid-column-one-third` and I'm not aware of that having been an issue for anyone.

fofr:
What's happening with this PR now?

nickcolley:
@fofr sorry I think I got us confused and left this in an unresolved place.

I'll work with the team to get consensus and reply to you, thanks for your patience.

nickcolley:
@fofr sorry for the delay, we decided that in order to keep consistency with the existing `govuk-grid-column-one-third-from-desktop` class that we want to name this new override `govuk-!-display-none-print`. If you want we can do this on your behalf, let us know :)

fofr:
@nickcolley PR updated.

Issue #1636, opened 2019-11-12
'Unlabelled graphic' read out within header using NVDA


36degrees:
From https://github.com/alphagov/govuk-design-system-backlog/issues/97#issuecomment-535417997:

> On reviewing the header https://design-system.service.gov.uk/components/header/default/index.html# in NVDA for Windows the png  is declared as "Unlabelled graphic" (text from NVDA text viewer below)
>
> Header – Example – GOV.UK Design System document
> visited link
> **Unlabelled graphic
> To get missing image descriptions, open the context menu.**
> GOV.UK
>
> To make NVDA ignore the  tag I'd suggest giving the  a role="presentation" too as the same role attribute in the tag doesn't seem to carry through.

I've been able to reproduce this using Chrome + NVDA on Windows 10.

nickcolley:
Is this relating to the SVG that already has `[role="presentation"]` on it? I wonder if `[aria-hidden="true"]` has the same issue.

36degrees:
As per the original comment, I think this might relate to the fallback png image. We'd need to do a bit of investigating to know for sure.

nickcolley:
It seems to be an accessibility feature on Chrome since you can right click then 'Get image descriptions from Google > Always | Just once'

When pressing 'just once'.

Now when I visit the logo it says the same message but 'Getting description...', but never succeeds.

When putting an empty alt attribute on the fallback image it stops the image from being detected for a description.

Changing `role=presentation` to `aria-hidden=true` also stops the image from being detected for a description.

My suggestion would be to always use `aria-hidden=true` when we intend for something to be removed from the accessibility tree as it also stops the children of the element from being shown.

nickcolley:
Here is the feature explained: https://support.google.com/chrome/answer/9311597?hl=en

nickcolley:
> Using aria-hidden="true" will completely remove all elements, child elements, and content from the accessibility tree regardless of the default semantic nature and they will remain removed until the aria state is changed to false it functions in much the same way as CSS’s display:none. This is where it differs from role="presentation".

https://timwright.org/blog/2016/11/19/difference-rolepresentation-aria-hiddentrue/

PR #1724, opened 2020-02-05
Fix fallback logo being detected by chromes image description feature

nickcolley:
role="none"/"presentational" does not do enough to hide the child fallback image used in older browsers.

We have found that Chrome's new "Get image descriptions from Google" feature will notice this fallback image.

This means when screenreaders such as NVDA visit the logo, it'll announce "Unlabelled graphic, To get missing image descriptions, open the context menu." for this fallback image.

By using aria-hidden="true" we can ensure that the SVG and it's children are all hidden from the accessibility tree, which I have confirmed in NVDA + Chrome.

I have also made this change to the other places we're using role="presentation" in this way for consistency.

- [Get image descriptions on Chrome](https://support.google.com/chrome/answer/9311597?hl=en)
- [The Difference Between role=”presentation” and aria-hidden=”true”](https://timwright.org/blog/2016/11/19/difference-rolepresentation-aria-hiddentrue/)

Closes https://github.com/alphagov/govuk-frontend/issues/1636

### DIFF
--- a/src/govuk/components/button/template.njk
+++ b/src/govuk/components/button/template.njk
@@ -24,7 +24,7 @@
 {#- The SVG needs `focusable="false"` so that Internet Explorer does not
 treat it as an interactive element - without this it will be
 'focusable' when using the keyboard to navigate. #}
-<svg class="govuk-button__start-icon" xmlns="http://www.w3.org/2000/svg" width="17.5" height="19" viewBox="0 0 33 40" role="presentation" focusable="false">
+<svg class="govuk-button__start-icon" xmlns="http://www.w3.org/2000/svg" width="17.5" height="19" viewBox="0 0 33 40" aria-hidden="true" focusable="false">
   <path fill="currentColor" d="M0 0h13l20 20-20 20H0l20-20z"/>
 </svg>
   {% endset %}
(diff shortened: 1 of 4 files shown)


36degrees:
Makes sense to me, thanks Nick 👏 How confident are you that this will work consistently everywhere? Do you think we need to test this change across all of the assistive technologies we support?

nickcolley:
@36degrees I have not done a full run through testing with all assistive technologies and browsers as I have confidence that this aria attribute has very broad support and we use it elsewhere (for example the accordion component).

I can do some testing soon to double check.

zaccolley:
Well done Nick, proud of you.

nickcolley:
@36degrees I've tested this in JAWS, NVDA in Edge, Chrome, IE11, Firefox and VoiceOver in Safari and it dont hear any icons being announced when tabbing to the GOV.UK link in the header.

nickcolley:
@36degrees I've tested this in JAWS, NVDA in Edge, Chrome, IE11, Firefox and VoiceOver in Safari and it dont hear any icons being announced when tabbing to the GOV.UK link in the header.

36degrees:
If you're confident with the changes to the other components as well then I think this is good to merge 👍

nickcolley:
@36degrees I also tested the start link button

nickcolley:
Needs a changelog,oops

PR #1727, opened 2020-02-07
Tidy contribution docs

m-green:
This tidies the docs/contributions/ folder based on the 'contributions docs quick wins' doc we discussed in December.

It:

- remove `docs/contributions/README.md`, as it's an artefact of trying to create hierarchical docs navigation in GitHub (which doesn’t work well), and only links to the root `CONTRIBUTING.md` file
- remove `docs/contributions/coding-standards/README.md` for the same reason as above
- moves the content from `docs/contributions/deploying.md` into `running-locally.md`, as it's too small to justify its own file
- moves `docs/contributions/publishing-a-prerelease.md` and `docs/contributions/publishing.md` into a new `docs/releases` folder - as they meet the needs of our team more than contributors


36degrees:
This looks great @m-green. It'd be good to try and tidy up the commit history so we don't have a merge commit in here – would be happy to help with that.

36degrees:
@m-green I think this is good to merge

PR #1730, opened 2020-02-10
Update skip link macro param documentation

frankieroberto:
The `#content` is currently not displaying on https://design-system.service.gov.uk/components/skip-link/ as it is being interpreted as a comment within the YAML file.

The `href` param also doesn't seem to be strictly "required", as a default value is specified.

### DIFF
--- a/src/govuk/components/skip-link/skip-link.yaml
+++ b/src/govuk/components/skip-link/skip-link.yaml
@@ -10,7 +10,7 @@ params:
 - name: href
   type: string
   required: false
-  description: The value of the skip link's `href` attribute. Defaults to `#content` if you do not provide a value.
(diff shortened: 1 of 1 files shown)


src/govuk/components/skip-link/skip-link.yaml
```diff
@@ -9,8 +9,8 @@ params:
   description: If `text` is set, this is not required. HTML to use within the skip link component. If `html` is provided, the `text` argument will be ignored.
 - name: href
   type: string
-  required: true
-  description: The value of the skip link href attribute. Defaults to #content
+  required: false
+  description: The value of the skip link href attribute. Defaults to `#content` if not provided.
```
m-green:
```suggestion
  description: The value of the skip link's `href` attribute. Defaults to `#content` if you do not provide a value.
```

PR #1741, opened 2020-02-21
Add Sass compiler requirements to npm install docs

m-green:
This updates our npm install documentation to add a Sass install step in 'Requirements', and list which Sass compilers we support.

Please check for factual accuracy.

(I'll also add this to [our overall npm install documentation improvement](https://github.com/alphagov/govuk-frontend/pull/1697) when we iterate it after our research analysis).

### DIFF
--- a/docs/installation/installing-with-npm.md
+++ b/docs/installation/installing-with-npm.md
@@ -15,7 +15,14 @@ To use GOV.UK Frontend with NPM you must:
    you don’t already have one. You can create a default `package.json` file by
    running `npm init` from the root of your application.
 
-3. If you want to use the GOV.UK Frontend Nunjucks macros, install Nunjucks -
+3. Install either:
+
+   - [LibSass](https://sass-lang.com/libsass) - version 3.3.0 or higher
+   - [Dart Sass](https://www.npmjs.com/package/sass) - version 1.0.0 or higher
+
+   GOV.UK Frontend also works with Ruby Sass version 3.4.0 or higher, but we recommend you do not use Ruby Sass because it's been deprecated.
+
+4. If you want to use the GOV.UK Frontend Nunjucks macros, install Nunjucks -
    the minimum version required is 3.0.0.
 
 ```


docs/installation/installing-with-npm.md
```diff
@@ -15,7 +15,14 @@ To use GOV.UK Frontend with NPM you must:
    you don’t already have one. You can create a default `package.json` file by
    running `npm init` from the root of your application.
 
-3. If you want to use the GOV.UK Frontend Nunjucks macros, install Nunjucks -
+3. Install either:
```
nickcolley: Annoying pedant comment :nerd_face: :

I'm not a pro on when to use 'hyphens' vs endash vs emdash but should we be using a hyphen here?

m-green: Well spotted! The wider world may well use a dash here (and I would have used a dash in previous roles at other orgs). But the GOV.UK style is to use a hyphen. Neither Stephen nor I have the history on why exactly...

docs/installation/installing-with-npm.md
```diff
@@ -15,7 +15,14 @@ To use GOV.UK Frontend with NPM you must:
...
+
+   - [LibSass](https://sass-lang.com/libsass) - version 3.3.0 or higher
```
36degrees: Do we think it's worth documenting in some way the issue that GOV.UK have seen with LibSass and Rails? https://github.com/alphagov/finder-frontend/issues/1124

nickcolley: Seems like it's a common issue people will run into :+1:

36degrees: Happy for that to be done later if it's going to block this otherwise.

m-green: Happy to add something! Would this work?

"If your app uses Rails, we recommend not using SassC::Rails because it may cause errors. Install Dart Sass instead." 

I appreciate there are other options, but installing Dart Sass seems like the simplest route?

nickcolley: Maybe we should raise an issue and do it later just cos it's a bit nuanced...

m-green: Sounds good - let's go with what we have and look at the LibSass and Rails issue later.

36degrees: In which case I think this is good to 🚀

PR #1745, opened 2020-02-25
Update vendor polyfills to match upstream

nickcolley:
We have contributed these [polyfills back upstream to
polyfill-library](https://github.com/Financial-Times/polyfill-library/pull/484).

Removes Immediately Invoked Function Expression [IFFE](https://developer.mozilla.org/en-US/docs/Glossary/IIFE)
as it's not required since there's no reference to 'global' and the
overall file has an IFFE.

Updates links to the upstream repository.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 ### Fixes
 
+- [Pull request #1745: Update vendor polyfills to match upstream](https://github.com/alphagov/govuk-frontend/pull/1745).
 - [Pull request #1724: Fix fallback logo being detected by Google Chrome's image description feature](https://github.com/alphagov/govuk-frontend/pull/1724).
 
 ## 3.5.0 (Feature release)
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 ### Fixes
 
+- [Pull request #1745: Update vendor polyfills to match upstream](https://github.com/alphagov/govuk-frontend/pull/1745).
```
36degrees: Do we think this is useful for users to have this in the release notes, as it's mostly an internal change?

nickcolley: I err on the side of strict most of the time, we'd probably get away with not having an entry though. Up to you.

36degrees: I have no particular preference, was just unsure about it myself. Maybe @m-green can look at this when we come to cut the next release.

Issue #1339, opened 2019-05-13
Consider making tabs <section> into <div> 


hannalaakso:
On encountering [`<section>` in tabs](https://github.com/alphagov/govuk-frontend/blob/fea29aaa833d5f8233c71e0d076923a697e752cd/src/components/tabs/template.njk#L26), HTML validator gives the following warning: 
> Section lacks heading. Consider using h2-h6 elements to add identifying headings to all sections.

if it doesn't contain a heading.

According to the [HTML specs](https://www.w3.org/TR/html52/sections.html#the-section-element)
> Each section should be identified, typically by including a heading (h1-h6 element) as a child of the section element.

We should consider making this element a `<div>`.

PR #1746, opened 2020-02-25
Use generic div element for tabspanel

nickcolley:
Sections when they have an accessible name create landmarks.

> The use of region landmarks should be limited, as too many landmarks can dilute their usefulness in quickly navigating to important areas of a page. If you find there are many landmarks, especially generic regions existing in a single page, then the page’s structure should be reexamined.

- https://www.scottohara.me/blog/2018/03/03/landmarks.html

Also, when the component has not be enhanced these sections do not have
accessible names so it means they are the equivalent of a div already.

When they're enhanced they're given a role of tabpanel, so they're no
longer a section.

Given this I'm not sure it was ever useful and by making them generic
divs we can avoid HTML validators warning about this component.

Closes https://github.com/alphagov/govuk-frontend/issues/1339

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 ### Fixes
 
+- [Pull request #1746: Use generic div element for tabspanel](https://github.com/alphagov/govuk-frontend/pull/1746).
 - [Pull request #1745: Update vendor polyfills to match upstream](https://github.com/alphagov/govuk-frontend/pull/1745).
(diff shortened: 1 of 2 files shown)


alex-ju:
As I remember, on the initial implementation of tabs component a (potentially visually hidden) heading was meant to be automatically generated for each section acting as identifier ([as exemplified in the Design System](https://design-system.service.gov.uk/components/tabs/)) – hence passing the HTML validation.

PR #1747, opened 2020-02-25
Test Sass compilation against supported compilers

### LINKED ISSUE #1684: Fail the build if compilation fails with any of our supported Sass compilers
## What

Compile the GOV.UK Frontend Sass using the minimum version of each of the Sass compilers that we support.

Fail the build if it fails to compile using any one of them.

## Why

So that we have the confidence to ship releases whilst knowing that they should work with any of the Sass compilers that we support

## Questions / prompts

- Should we also consider compare the output of the different compilers to ensure they're effectively equivalent? (May be very difficult if there are big differences in order / whitespace etc?)
- Check to see if any other Sass frameworks are already doing this.
- What impact will this have on our CI run time?

## Dependencies

- [x] Be explicit about which Sass compilers we support #1683

36degrees:
Compile the GOV.UK Frontend Sass using the minimum version of each of the Sass compilers that we support:

- LibSass v3.3.0
- Dart Sass v1.0.0
- Ruby Sass v3.4.0 (deprecated; not recommended)

Fail the build if it fails to compile using any one of them.

We're doing this so that we have the confidence to ship releases whilst knowing that they should work with any of the Sass compilers that we support.

This also removes a before_deploy step (because we no longer have a deploy step) and the sudo key as according to Travis' config linter it has no effect any more.

Closes #1684

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,20 +1,46 @@
 language: node_js
 notifications:
   email: false
-sudo: false
-before_deploy:
-- test $TRAVIS_TEST_RESULT = 0
+
 jobs:
   include:
-    - stage: Run npm test
+    - name: 🧪 Run test suite
       script:
-      # Ensure Travis aborts without running tests if the package-lock.json file needs updating
+      # Ensure Travis aborts without running tests if the package-lock.json file
+      # needs updating
       - set -e
       - ./bin/check-package-lock.sh
-      # TravisCI is slower than our local machines, so results in intermittent timeouts
-      # using the `--runInBand` flag we can force it to without requiring as much
-      # resources (https://facebook.github.io/jest/docs/en/troubleshooting.html#tests-are-extremely-slow-on-docker-and-or-continuous-integration-ci-server)
+      # TravisCI is slower than our local machines, so results in intermittent
+      # timeouts using the `--runInBand` flag we can force it to without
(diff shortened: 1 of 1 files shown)


36degrees:
Example build output: https://travis-ci.com/alphagov/govuk-frontend/builds/150477666

36degrees:
Example failed build: https://travis-ci.com/alphagov/govuk-frontend/builds/150478107

.travis.yml
```diff
@@ -1,20 +1,46 @@
...
+      install: npm install -g node-sass@v3.4.0 # node-sass v3.4.0 = libsass v3.3.0
+      script: 
+        - node-sass -v
+        - time node-sass src/govuk/all.scss > /dev/null
+
+    - name: 🎯 Test Sass compiles using Dart Sass v1.0.0
+      node_js: 8 # Assumed based on release date of dart-sass v1.0.0
+      install: npm install -g sass@v1.0.0
+      script: 
+        - sass --version
```
nickcolley: nit I prefer expanded flags e.g. --global vs -g

we're using them interchangably, not a blocker but could be consistent either way

36degrees: For some reason `dart-sass` doesn't like `-v`, so I had to use the expanded version. Will expand consistently.

PR #1748, opened 2020-02-25
Update `errorMessage` option documentation with falsy value behaviour

m-green:
Updates the description of the `errorMessage` option (in all the component pages it appears) so that it's clearer that a falsy value will be ignored.

Please check if factually correct.

### DIFF
--- a/src/govuk/components/character-count/character-count.yaml
+++ b/src/govuk/components/character-count/character-count.yaml
@@ -40,7 +40,7 @@ params:
 - name: errorMessage
   type: object
   required: false
-  description: Options for the errorMessage component (e.g. text).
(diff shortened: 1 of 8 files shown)


hannalaakso:
Looks great! Should we have an example of how to do this eg. 
```
{% if error %}
  {% set errorMessage = { text: "error!" } %}
{% endif %}
...
{{ govukDateInput({
  errorMessage: errorMessage,
}) }}
```

I'm not sure if it's immediately clear how you should do this without an example.

What do you think @edwardhorsford?

edwardhorsford:
@m-green thanks for adding.

@hannalaakso I'm tempted to think the description added is the most important part - as this is more relevant to developers implementing.

If you ever get to doing full worked pages, it might be helpful though.

PR #1749, opened 2020-02-26
Add note about override naming for media queries

nickcolley:


src/govuk/README.md
```diff
@@ -70,6 +70,6 @@ and override all other layers. Rules in this layer will generally be marked as
 Class names within this layer use an additional prefix `-!-`, for example
 `.govuk-!-font-width-regular`.
 
-
+If an override is applied conditionally the class name will indicate that at the end, for example: `.govuk-!-display-none-print`.
```
m-green: Looks good! Could we expand the example a little bit (and tweak to active voice)?

```suggestion
If a class only overrides when a condition is met, the condition will be at the end of the class name. For example `.govuk-!-display-none-print` will only override with `display: none` in a print layout.
```

nickcolley: I hope to one day understand and be able to spot active voice in my work. 😭 

Very nice improvement, thank you.

PR #1751, opened 2020-02-26
Mobile breadcrumbs show first and last only

RobBond-GDS:
On mobile devices collapse the breadcrumb to first and last items instead of showing all breadcrumb items.

## What
Apply mobile enhancement designs to the GOV.UK Breadcrumb component, specifically:

- Collapse breadcrumb to show only the first and last item (the parent item) on devices less than 640px wide
- Increase the touch target padding on mobile devices
- Collapse on Mobile to be de facto behaviour
- Update the sample yaml file to show how the breadcrumb would handle long taxons

## Why

The reasoning is taken from this snippet in the trello card:

> This [article by NNG](https://www.nngroup.com/articles/breadcrumbs/) lays out the pros and cons of showing a full breadcrumb path on mobile. In short, the drawbacks of having a multi-line breadcrumb, and the need to save space on mobile, make it less useful to users. By truncating the breadcrumb we at least allow users to go backwards by one level, or home.

## Visual Changes
Screenshot 2020-02-05 at 14 57 28
Before Change

Screenshot 2020-02-05 at 14 56 20
After Change

### DIFF
--- a/src/govuk/components/breadcrumbs/_breadcrumbs.scss
+++ b/src/govuk/components/breadcrumbs/_breadcrumbs.scss
@@ -111,4 +111,45 @@
     @include govuk-link-common;
     @include govuk-link-style-text;
   }
+
+  @include govuk-media-query($until: tablet) {
+
+    .govuk-breadcrumbs__list {
+      display: flex;
+    }
+
+    .govuk-breadcrumbs__list-item {
+      display: none;
+    }
+
+    .govuk-breadcrumbs__list-item {
+      &:first-child,
+      &:last-child {
+        display: inherit;
+      }
+    }
+
+    .govuk-breadcrumbs__list-item {
+      padding-top: 14px;
+      padding-bottom: 14px;
+    }
+
+    .govuk-breadcrumbs__list-item:before {
+      top: 18px;
+      margin: 0;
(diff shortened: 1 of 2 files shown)


vanitabarrett:
If this is contributed, it's worth stressing that this is a breaking change for anyone using the breadcrumbs component to display page hierarchy without providing links to all pages. 

Example:
This breadcrumb on desktop...
Screenshot 2020-02-27 at 16 57 10

...becomes this breadcrumb on mobile, which provides less navigation options to mobile users:
Screenshot 2020-02-27 at 16 57 18

36degrees:
Thanks for the heads up @vanitabarrett 👍 

We were talking about this yesterday, and think that this behaviour might make more sense applied as a modifier, e.g. `govuk-breadcrumbs--collapse-on-mobile`, rather than replacing the default behaviour.

timpaul:
Agreed - there may be some people using breadcrumbs to show a strict
hierarchy, for example product export codes - where removing some of the
links might be misleading.

vanitabarrett:
Hi all - this has been superseded by https://github.com/alphagov/govuk-frontend/pull/1754 . Feel free to close this PR 🙂

PR #1752, opened 2020-02-27
Improve Sass compilation performance when importing all components

### LINKED ISSUE #1671: Slow compilation of SCSS files in Rails apps
Hello 👋 from GOV.UK Publishing. We've been having some problems with speed of CSS compilation of GOV.UK apps for a while and after doing some detective work we've traced that a large volume of the time is due to importing files from govuk-frontend.

## Backstory

GOV.UK apps use the [Rails asset pipline](https://guides.rubyonrails.org/asset_pipeline.html) to transpile SCSS to CSS. They mostly continue to use the deprecated [Ruby SASS](https://github.com/sass/ruby-sass) library, but are likely to start switching to [sassc](https://github.com/sass/sassc-ruby) in the forseeable future. Nearly all GOV.UK apps use the [govuk_publishing_components](https://github.com/alphagov/govuk_publishing_components) gem which pulls in the govuk-frontend node module and [imports all components](https://github.com/alphagov/govuk_publishing_components/blob/530de08061c54d670f857375c9eeaae00ba5674a/app/assets/stylesheets/govuk_publishing_components/_all_components.scss#L9).

Gradually over the last few months or years we've been seeing the time that is needed to compile assets for GOV.UK has increased and once we switched to a [docker based development environment](https://github.com/alphagov/govuk-docker) we started to investigate as it seemed abnormally poor. Developers are finding that frequently an app won't seem to work the first time you try run it locally as the CSS could be taking more than a minute to compile. We did start to wonder why other Rails devs weren't going crazy about the slowness of CSS precompilation as it felt increasingly broken.

Looking into this further we found that most of the time was spent around the process of using `@import` to include files. Looking at a particular app, [Finder Frontend](https://github.com/alphagov/finder-frontend), we could see that 7054 files were being imported, despite less than 400 distinct files being used, and every one of these files requires processing which seemed to be where the majority of the time was spent. Following this further we could see that from the point where we import [all govuk-frontend components](https://github.com/alphagov/govuk_publishing_components/blob/530de08061c54d670f857375c9eeaae00ba5674a/app/assets/stylesheets/govuk_publishing_components/_all_components.scss#L9) was where we were seeing the big increase in imported files.

Some basic benchmarks from my MBP 2015 on time to generate [Finder Frontend's application.css](https://github.com/alphagov/finder-frontend/blob/a4ee821fc6be49f1ab429d390f27d4afae0ca057/app/assets/stylesheets/application.scss) file:

| Environment   | SASS      | Cache | Time (s) |
|---------------|-----------|-------|----------|
| govuk-docker  | Ruby SASS | Cold  | 59.983   |
| govuk-docker  | Ruby SASS | Warm  | 0.228    |
| govuk-docker  | sassc     | Cold  | 32.693   |
| govuk-docker  | sassc     | Warm  | 17.478   |
| Local machine | Ruby SASS | Cold  | 31.926   |
| Local machine | Ruby SASS | Warm  | 0.063    |
| Local machine | sassc     | Cold  | 11.805   |
| Local machine | sassc     | Warm  | 2.413    |

## Cause and effect

The problem of slow CSS compilation for GOV.UK seems to stem from the `@import` approach used in govuk-frontend and how this can affect Rails apps.

If we follow [govuk/all.scss](https://github.com/alphagov/govuk-frontend/blob/c5b75de1bfe1d3d75b16151e05f7fe066b3d7618/src/govuk/all.scss) we import "settings/all" (>= 13 files), "tools/all" (>= 9 files), "helpers/all" (>= 12 files). Then for each component the [same files are then imported](https://github.com/alphagov/govuk-frontend/blob/c5b75de1bfe1d3d75b16151e05f7fe066b3d7618/src/govuk/components/accordion/_accordion.scss#L1-L3). As sass [executes each `@import`](https://sass-lang.com/documentation/at-rules/import) whether or not it's been included before this can exponentially increase the amount of work done. Some basic maths would have the 30 components adding up to at least 384 files imported but I think the number ends up being quite significantly greater.

We're theorising that the sheer quantity of imported files affects the Rails asset pipeline particularly badly as everyone one of these files is checked to see if it uses any Rails specific functions ([asset-url, asset-path](https://github.com/rails/sass-rails#asset-helpers)). It could well be paired with IO operation pain too, as it probably searches a whole bunch of directories for every single file.

I put together some [isolated test cases](https://github.com/kevindew/govuk-frontend-rails-performance) to show the effects in a Rails app free from the rest of the GOV.UK infrastructure. As a basis for comparison I used [a version of govuk-frontend I put together](https://github.com/kevindew/govuk-frontend-import-test) that was optimised for the all.scss file where all other imports were removed.

Some benchmarks were:

| SASS      | Cache | Time (s)  |
|-----------|-------|-----------|
| sassc     | Cold  | 7.494180  |
| sassc     | Warm  | 1.667473  |
| Ruby SASS | Cold  | 15.556202 |
| Ruby SASS | Warm  | 0.007234  |

Where we can see this performs relatively consistently with Finder Frontend app, just clearly has a lot less to do being isolated.

And compared to the version with duplicate `@import` calls removed:

| SASS      | Cache | Time (s) | Decrease |
|-----------|-------|----------|----------|
| sassc     | Cold  | 0.552520 | 92.6 %   |
| sassc     | Warm  | 0.007781 | 99.5 %   |
| Ruby SASS | Cold  | 2.913921 | 81.3 %   |
| Ruby SASS | Warm  | 0.005758 | 20.4 %   |

Here we can see quite how dramatic the effects of `@import` can be.

I did ponder whether things would be quicker using webpacker with Rails. There I saw it take [~5s](https://github.com/kevindew/govuk-frontend-rails-performance#with-govuk-frotend-using-webpacker) to compile with `@imports` duplicated and [~2.5s](https://github.com/kevindew/govuk-frontend-rails-performance#with-govuk-frontend-using-webpacker-with-imports-culled) with them removed. So it seemed difference is still significant here but as the scale is much reduced the impact is less. I'll also add that I really didn't put much time into setting up webpacker so these numbers could be flawed.

Finally, I tried running the sassc command line utility to see if this is slow I found that it took [0.27s](https://github.com/kevindew/govuk-frontend-rails-performance#with-govuk-frotend-using-sassc) to compile with `@imports` duplicated against [0.21s](https://github.com/kevindew/govuk-frontend-rails-performance#with-govuk-frontend-with-imports-slimmed-and-sassc) with them removed. Which isn't a very significant difference.

## Impact on GOV.UK

Some of the effects this has had on GOV.UK have been:

- Apps never seem to work on first request for local dev, as Rails hangs waiting for assets to preload. Unaccustomed developers assume the app is broken, abort it and ask for help
- First test was frequently failing on apps due to timeout. We [switched to running a manual pre-compile before tests](https://github.com/alphagov/govuk-jenkinslib/pull/50) to resolve it.
- Switching to sassc is concerning because of the heavy performance impact on even a warm cache.

In general we seem to have ended up with a perfect storm on GOV.UK by having some apps needing to precompile [multiple files with the same contents](https://github.com/alphagov/whitehall/tree/76c64d04d06909977d7fd31a2a89103aacbb5aaa/app/assets/stylesheets/frontend), importing all of govuk-frontend and then switching to the slower docker environment for development.

We're [currently pondering](https://github.com/alphagov/govuk-rfcs/pull/110) about switching away from the approach of including all govuk-frontend components which would likely offer some performance improvement. Though we'd still be paying a reasonable performance penalty for duplicate `@import` calls.

Although we could see performance improve a reasonable amount by switching to use webpacker with Rails this seems like a long way off for GOV.UK given the quantity of apps and the amount of existing asset-pipeline based code we have. We're still a reasonable distance from being able to migrate one app to webpacker

## Things we're wondering about

- Whether something is broken in [sassc-rails](https://github.com/sass/sassc-rails) for the warm cache performance to perform so badly
- Quite what is going on with each import that impacts time so badly
- [`@use`](https://sass-lang.com/documentation/at-rules/use) seems the ideal solution for resolving this problem but it's not available on libsass.
- Whether other teams on different stacks are experiencing same performance issues
- Whether the govuk-frontend style of `@import` usage is common or not - I notice it's not an approach tachyons or bootstrap use - as to whether wider world is affected.

It'd be great if we can try work out if there are things that can be done to reduce this problem. Although aspects of this seem quite specific Rails it does seem to be a problem that is likely to affect other environments (as multiple execution is a property of `@import`) and is likely to only increase in severity as more components are added.

36degrees:
We currently import the settings, tools and helpers layers within each component so that users can import just a single component (e.g. `@import govuk/components/button/button`) and it'll import its own dependencies.

However, Ruby Sass in particular is slow to resolve these imports, and so when importing `govuk/all`, the fact that every component imports the dependencies again has a big impact on the compilation time.

By moving everything except the imports into a separate file (`_style.scss`) and including that in the original component .scss file, we can then import just the styles when importing all and shortcut all of the imports.

We can use the same logic to remove imports of dependent components (e.g. character count imports the styles for textarea).

Finally, we can remove two redundant imports of the typography helper, as we already import all helpers.

## Ruby Sass

**Before**: 30s
**After**: 7.8s

```
➜ sass --version
Ruby Sass 3.7.4
```

### Baseline (`master`)

```
➜ time sass -C src/govuk/all.scss > /dev/null
sass -C src/govuk/all.scss > /dev/null  30.02s user 0.99s system 98% cpu 31.614 total

➜ time sass -C src/govuk/all.scss > /dev/null
sass -C src/govuk/all.scss > /dev/null  30.00s user 0.96s system 98% cpu 31.363 total

➜ time sass -C src/govuk/all.scss > /dev/null
sass -C src/govuk/all.scss > /dev/null  29.29s user 0.91s system 98% cpu 30.524 total
```

### After changes (`sass-perf`)

```
➜ time sass -C src/govuk/all.scss > /dev/null
sass -C src/govuk/all.scss > /dev/null  7.80s user 0.35s system 98% cpu 8.304 total

➜ time sass -C src/govuk/all.scss > /dev/null
sass -C src/govuk/all.scss > /dev/null  7.89s user 0.35s system 98% cpu 8.321 total

➜ time sass -C src/govuk/all.scss > /dev/null
sass -C src/govuk/all.scss > /dev/null  7.68s user 0.32s system 99% cpu 8.049 total
```

## LibSass

**Before**: 0.2s
**After**: 0.18s

```
➜ sassc -v
sassc: 3.6.1
libsass: 3.6.3
sass2scss: 1.1.1
sass: 3.5
```

### Baseline (`master`)

```
➜ time sassc src/govuk/all.scss > /dev/null
sassc src/govuk/all.scss > /dev/null  0.20s user 0.02s system 98% cpu 0.224 total

➜ time sassc src/govuk/all.scss > /dev/null
sassc src/govuk/all.scss > /dev/null  0.20s user 0.02s system 98% cpu 0.224 total

➜ time sassc src/govuk/all.scss > /dev/null
sassc src/govuk/all.scss > /dev/null  0.20s user 0.02s system 99% cpu 0.224 total
```

### After changes (`sass-perf`)

```
➜ time sassc src/govuk/all.scss > /dev/null
sassc src/govuk/all.scss > /dev/null  0.18s user 0.03s system 87% cpu 0.242 total

➜ time sassc src/govuk/all.scss > /dev/null
sassc src/govuk/all.scss > /dev/null  0.17s user 0.02s system 93% cpu 0.198 total

➜ time sassc src/govuk/all.scss > /dev/null
sassc src/govuk/all.scss > /dev/null  0.16s user 0.02s system 98% cpu 0.183 total
```

## Dart Sass

**Before**: 0.5s
**After**: 0.25s

```
➜ /usr/local/Cellar/sass/1.23.7/bin/sass --version
1.23.7
```

### Baseline (`master`)

```
➜ time /usr/local/Cellar/sass/1.23.7/bin/sass src/govuk/all.scss > /dev/null
/usr/local/Cellar/sass/1.23.7/bin/sass src/govuk/all.scss > /dev/null  0.53s user 0.22s system 100% cpu 0.743 total

➜ time /usr/local/Cellar/sass/1.23.7/bin/sass src/govuk/all.scss > /dev/null
/usr/local/Cellar/sass/1.23.7/bin/sass src/govuk/all.scss > /dev/null  0.50s user 0.21s system 99% cpu 0.714 total

➜ time /usr/local/Cellar/sass/1.23.7/bin/sass src/govuk/all.scss > /dev/null
/usr/local/Cellar/sass/1.23.7/bin/sass src/govuk/all.scss > /dev/null  0.53s user 0.23s system 100% cpu 0.754 total
```

### After changes (`sass-perf`)

```
➜ time /usr/local/Cellar/sass/1.23.7/bin/sass src/govuk/all.scss > /dev/null
/usr/local/Cellar/sass/1.23.7/bin/sass src/govuk/all.scss > /dev/null  0.23s user 0.10s system 101% cpu 0.331 total

➜ time /usr/local/Cellar/sass/1.23.7/bin/sass src/govuk/all.scss > /dev/null
/usr/local/Cellar/sass/1.23.7/bin/sass src/govuk/all.scss > /dev/null  0.24s user 0.10s system 101% cpu 0.332 total

➜ time /usr/local/Cellar/sass/1.23.7/bin/sass src/govuk/all.scss > /dev/null
/usr/local/Cellar/sass/1.23.7/bin/sass src/govuk/all.scss > /dev/null  0.24s user 0.10s system 101% cpu 0.340 total
```

Closes #1671


nickcolley:
Apart from it being a bit weird it seems like a good way forward, nice thinking.

Think we could think about the following:

- what people should do to get a permanent custom build. Should we deprecate the current `button/button` import pattern entirely?
- documentation explaining why we have two files? I don't think we do this for Nunjucks perhaps so maybe this is a nice to have.
- seems like big improvements on ruby sass, but we know the issues are more to do Rails wrappers running ruby code regardless of the compiler, since even the libsass sassc-rails implementation seemed to have issues. Should we run this benchmark against something more real world?

36degrees:
Just on the last point whilst I think about the others…

> * seems like big improvements on ruby sass, but we know the issues are more to do Rails wrappers running ruby code regardless of the compiler, since even the libsass sassc-rails implementation seemed to have issues. Should we run this benchmark against something more real world?

I think the next step might be to publish this as a pre-release and ask GOV.UK to do some benchmarking to understand if it improves things for them?

36degrees:
> This might be a breaking change for any folks (low risk) that import only `@components/all`? I think it's a low impact breaking change but may need to go into 4.0.0

I guess we could add a single import of the three layers to `components/all` to mitigate that, if we wanted to? It'd reduce performance slightly again, but should still have a significant positive impact overall.

36degrees:
> What should people do to get a performant custom build.
> Should we deprecate the current `button/button` import pattern entirely?
> Maybe not worth the trouble unless people run into it.

Good question. I think I'd expect people to start with a copy of `govuk/all.scss` and customise it to be more specific / remove the bits they don't need?

```scss
@import "settings/all";
@import "tools/all";
@import "helpers/all";

@import "core/all";
@import "objects/all";

@import "components/button/style";
@import "components/input/style";
@import "components/header/style";
@import "components/footer/style";

@import "utilities/all";
// Removed overrides as not using them
```

The use of `/style` does seem a bit strange, but maybe no more so than `button/button`…

Another option might be just to make this a proper breaking change, and remove the ability to import a single component without having to also import the `settings`, `tools` and `helpers` layers. I'm not sure what we'd do about dependent components though?

I guess we should also bear in mind that at some point in the next 12 - 18 months we'll probably want to look at moving to the `@uses` syntax anyway?

nickcolley:
I think we should be taking more time to consider deprecation, some things we can't communicate deprecation that easily but Sass we often can do so pretty well with warnings.

36degrees:
Having reviewed this as a team, we think that although it should be a non-breaking change for most users, it:

- increases the complexity of the codebase
- adds yet another way that users could import GOV.UK Frontend
- does nothing to improve compilation performance for users that are importing individual components, e.g. @import govuk/components/button/button
- would be a breaking change for anyone importing govuk/components/all without also importing the underlying layers (which we think would be an unusual way to import it, but should currently work)

On that basis, we now think that the best approach would be to remove the settings, tools and helpers imports from the components entirely, and update our documentation to tell users to import those layers if they are importing individual components.

There's more information on next steps in https://github.com/alphagov/govuk-frontend/issues/1671#issuecomment-599429977.

PR #1753, opened 2020-03-02
Make back link arrow consistent with breadcrumb component

vanitabarrett:
## What
This PR contains a few small changes to the back link component:

- updates the back link arrow to be consistent with the breadcrumbs component
- increase the touch target of the back link

Note: this change has been [implemented on GOVUK](https://github.com/alphagov/govuk_publishing_components/pull/1299) and we wish to contribute this change back to the design system.

## Why

- Back links use an inconsistent icon from the breadcrumbs and there should be consistent design patterns across GOV.UK.
- The touch target of the back link should be a minimum of 44x44px as recommended by [WCAG guidelines](https://www.w3.org/WAI/WCAG21/Understanding/target-size.html)
- The touch target has been implemented regardless of device to handle people interacting with the link, for example, on touchscreen laptops.

## Before
Screenshot 2020-03-02 at 13 57 42
Screenshot 2020-03-02 at 13 57 54

## After
Screenshot 2020-03-02 at 13 58 02
Screenshot 2020-03-02 at 13 58 08

### DIFF
--- a/src/govuk/components/back-link/_back-link.scss
+++ b/src/govuk/components/back-link/_back-link.scss
@@ -28,19 +28,13 @@
     padding-left: 14px;
   }
 
-  // Only add a custom underline if the component is linkable
+  // Only underline if the component is linkable
   .govuk-back-link[href] {
-    // Use border-bottom rather than text-decoration so that the arrow is
-    // underlined as well.
-    border-bottom: 1px solid govuk-colour("black");
-
-    // Underline is provided by a bottom border
-    text-decoration: none;
+    text-decoration: underline;
 
     // When the back link is focused, hide the bottom link border as the
     // focus styles has a bottom border.
     &:focus {
-      border-bottom-color: transparent;
       text-decoration: none;
 
       &:before {
--- a/src/govuk/components/back-link/_back-link.scss
+++ b/src/govuk/components/back-link/_back-link.scss
@@ -4,6 +4,15 @@
 
 @include govuk-exports("govuk/component/back-link") {
 
+  // Size of chevron (excluding border)
+  $chevron-size: 7px;
+
+  // Size of chevron border
+  $chevron-border-width: 1px;
+
+  // Colour of chevron
+  $chevron-border-colour: $govuk-secondary-text-colour;
+
   .govuk-back-link {
     @include govuk-typography-responsive($size: 16);
     @include govuk-link-common;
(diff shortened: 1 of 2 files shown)


timpaul:
For reference, [this change has been requested before via our backlog](https://govuk-frontend-review-pr-1753.herokuapp.com/components/back-link).

src/govuk/components/back-link/_back-link.scss
```diff
@@ -4,6 +4,15 @@
...
 @include govuk-exports("govuk/component/back-link") {
 
+  // Size of chevron (excluding border)
+  $chevron-size: 7px;
+
+  // Size of chevron border
+  $chevron-border-width: 1px;
+
+  // Colour of chevron
+  $chevron-border-colour: $govuk-text-colour;
```
christopherthomasdesign, line 14: The chevron border colour here is different to that in the breadcrumbs. If we want to be consistent with the breadcrumbs this should be `$govuk-secondary-text-colour`

36degrees: This sort of makes sense to me – in the breadcrumbs, the arrow is a separator and isn't part of the link.

image

With the back link, it is treated as part of the link.

image

So in that sense, the arrow in the back link is interactive, and in the breadcrumbs it's not.

Thoughts?

christopherthomasdesign: Fair point yeah. I suppose I'd ask why the arrow should be interactive in the back link?

timpaul: Hmm, this is a subtle one. I think on balance the chevrons should both be the same colour, for visual consistency. What they signify in each case is very slightly different, and for that reason I think it's OK to keep the active areas as they currently are.

vanitabarrett: Thanks both! I've changed the colour for the chevron to be `$govuk-secondary-text-colour` by default, and `$govuk-text-colour` on focus.

36degrees:
Something slightly strange going on with the changelog – seeing duplicated entries for two unrelated changes. @vanitabarrett are you able to take a look?

vanitabarrett:
Oops, probably a rebase issue. I'll look into it

vanitabarrett:
@36degrees Fixed 👍

36degrees:
Thanks, @vanitabarrett 👏

PR #1754, opened 2020-03-02
Add collapseOnMobile breadcrumbs flag

vanitabarrett:
## What
Adds a collapseOnMobile flag to the breadcrumbs component. When set to true, this flag means the breadcrumb component will:

- Collapse to only showing the first and last item on tablet breakpoint and below
- Long breadcrumb text will wrap over multiple lines instead of the whole breadcrumb moving to a new line

Note: this change has been implemented on GOVUK and we wish to contribute this change back to the design system.

## Why
Screenshot 2020-03-02 at 16 14 10
Screenshot 2020-03-02 at 16 14 15

Screenshot 2020-03-02 at 16 07 54
Screenshot 2020-03-02 at 16 08 02

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,12 +4,17 @@
 
 ### New features
 
+#### Add collapseOnMobile breadcrumbs flag
+
+Add a collapseOnMobile flag to breadcrumbs which, when set to true, means the breadcrumb collapses to the first and last items only on mobile and doesn't wrap onto a new line.
+
+[Pull request #1754: Add collapseOnMobile breadcrumbs flag](https://github.com/alphagov/govuk-frontend/pull/1754)
+
 #### New list modifier to add extra spacing between list items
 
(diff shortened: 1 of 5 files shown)


36degrees:
As discussed on Slack, from talking to the team we’re not 100% sure about the touch target size changes at the minute because:

- it changes the position of the breadcrumb – because it has a lot more padding, there’s more space between the header and the breadcrumb when using collapseOnMobile, compared to the ‘normal’ breadcrumb or the back link
- this change wasn't called out when it was put to the working group (which is our fault, sorry!)

We've agreed to just ship the behaviour that hides everything between the first and last breadcrumb, and create an issue to review the touch target size separately (with yourself and the designer that worked on it).

36degrees:
Raised a separate issue to look at the breadcrumb touch target size: #1817

vanitabarrett:
Updated PR to remove touch target change but keep:

- collapsing to first and last item only on mobile
- Long breadcrumb text will wrap over multiple lines instead of the whole breadcrumb moving to a new line

PR #1767, opened 2020-03-16
Use aria-describedby relationship for count message

nickcolley:
Since aria-describedby is required for the hint message to announced by assistive technologies when interacting with the textarea we can use it to find the count message.

This makes the JavaScript slightly less reliant on the HTML for the component, for example if someone wants to use a different ID on their count message.


36degrees:
If the text message has a hint or is in an error state, I think `aria-describedby` will contain the IDs of the hint and error message respectively, as well as the count message?

nickcolley:
Good point, wonder if we could consider targetting the '.govuk-character-count__message' class, it is presentational though...

nickcolley:
Closing for now

PR #1774, opened 2020-03-30
Group helpers and objects into logical groups

36degrees:
Improve the navigation within Sass Docs by grouping helpers and objects into logical groups.

### DIFF
--- a/src/govuk/all.test.js
+++ b/src/govuk/all.test.js
@@ -1,5 +1,7 @@
 /* eslint-env jest */
 
+const sassdoc = require('sassdoc')
+
 const configPaths = require('../../config/paths.json')
 const PORT = configPaths.ports.test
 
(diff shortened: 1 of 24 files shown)


m-green:
Seems to make sense - will I be able to see how this looks in the preview of the frontend docs?

nickcolley:
@m-green I'll make a preview branch for you

nickcolley:
https://deploy-preview-19--govuk-frontend-docs-prototype.netlify.com/sass_api_reference/#sass-api-reference

m-green:
Thanks for the preview Nick! It wonder whether the style is a bit hard to understand and navigation. The '/' could be read as an 'or' (as in normal text), and in each h2 the pertinent bit is at the end of the heading rather than the start, which makes it tricky to glance down the nav list for what you need.

Would it be possible to reverse the order on each h2 and use lowercase for the parent name to make it more natural-sounding? For example:

Helpers
  Accessibility helpers
  Colour helpers
  Layout helpers
  ...

Objects
  Layout objects
  ...

Settings
  Assets settings
  Colours settings
  ...

nickcolley:
@m-green I wonder if we should consider something like 'General helpers' instead of 'helpers' as there isnt a hierarchy, those entries under the current 'helpers' section  are just those that do not fit in a specific category.

The heading structure is like this:

```markdown
## Helpers
## Accessibility helpers
## Colour helpers
```

Not 
```markdown
# Helpers
## Accessibility helpers
## Colour helpers
```

So perhaps this makes more sense from a hierarchy point of view? 
```markdown
## General helpers
## Accessibility helpers
## Colour helpers
```

m-green:
> So perhaps this makes more sense from a hierarchy point of view?
> 
> ```gfm
> ## General helpers
> ## Accessibility helpers
> ## Colour helpers
> ```

Yep that'd work! Longer-term, to help with navigation, I think it'd be good to have 4 h2s - 'Helpers', 'Objects', 'Tools', 'Settings' - and group under those. But I can see how that will shift all the headings down by one and possibly lead us to having h7s? So let's discuss that another day!

36degrees:
There's an example of what this would look like with the changes from https://github.com/alphagov/govuk-frontend-docs/pull/52 here:

https://sass-api-reference-groups-preview--govuk-frontend-docs-preview.netlify.app/sass-api-reference/#sass-api-reference

PR #1775, opened 2020-03-31
Add list--spaced modifier

### LINKED ISSUE #1272: More spacing between bulleted list items containing longer text
Our services uses bulleted lists for summaries of facts and figures, eg:

Screenshot 2019-04-12 at 13 08 02

These have tested well, and it's a common pattern also used by ONS in their statistical bulletins (see [example](https://www.ons.gov.uk/economy/nationalaccounts/balanceofpayments/bulletins/uktrade/february2019)).

We are spacing the bullets out more than the default, as otherwise they are harder to scan, eg:

Screenshot 2019-04-12 at 13 17 16

We are currently using this custom modifier to space the bullets out a bit more:

```css
.eff-list--sparse > li {
  @include govuk-responsive-margin(4, "bottom");
}
```

Is this something we could contribute back to the Design System perhaps? I feel like it may be useful for others too.

36degrees:
This modifier is designed to make lists easier to scan when the list items are long, and span multiple lines.

Closes #1272

### DIFF
--- a/src/govuk/core/_lists.scss
+++ b/src/govuk/core/_lists.scss
@@ -50,11 +50,23 @@
     }
   }
 
+  %govuk-list--spaced > li {
+    margin-bottom: govuk-spacing(3);
+
+    &:last-child {
+      margin-bottom: govuk-spacing(0);
+    }
+  }
+
   .govuk-list--bullet {
     @extend %govuk-list--bullet;
   }
 
   .govuk-list--number {
     @extend %govuk-list--number;
+  }
+
+  .govuk-list--spaced {
+    @extend %govuk-list--spaced;
   }
 }
--- a/src/govuk/core/_lists.scss
+++ b/src/govuk/core/_lists.scss
@@ -56,10 +56,6 @@
     @include govuk-media-query($from: tablet) {
       margin-bottom: govuk-spacing(3);
     }
-
-    &:last-child {
-      margin-bottom: govuk-spacing(0);
-    }
   }
 
   .govuk-list--bullet {
--- a/src/govuk/core/_lists.scss
+++ b/src/govuk/core/_lists.scss
@@ -51,7 +51,11 @@
   }
 
   %govuk-list--spaced > li {
-    margin-bottom: govuk-spacing(3);
+    margin-bottom: govuk-spacing(2);
+
+    @include govuk-media-query($from: tablet) {
+      margin-bottom: govuk-spacing(3);
+    }
 
     &:last-child {
       margin-bottom: govuk-spacing(0);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@
 
 ### New features
 
+#### New list modifier to add extra spacing between list items
+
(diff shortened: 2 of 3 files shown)


src/govuk/core/_lists.scss
```diff
@@ -48,11 +48,23 @@
     }
   }
 
+  %govuk-list--spaced > li {
+    margin-bottom: govuk-spacing(3);
+
+    &:last-child {
```
nickcolley: Does the original govuk-list remove the last child margin?

timpaul: The original list items don't have any bottom margins. But if they did then I think we would remove it from the last one.

timpaul:
> Makes sense, think we'll need to add some new guidance to the Design System too.

Absolutely - I'm going to work on that later

src/govuk/core/_lists.scss
36degrees: Hmm, they do on tablet and above:

https://github.com/alphagov/govuk-frontend/blob/5fcf41bc32df9f892f4df938994067c13433f041/src/govuk/core/_lists.scss#L42-L49

I think changing that behaviour might be a breaking change, but if we thought it made sense to bring them in line we could do that in 4.0?

Which raises another point – are we happy with the same spacing being applied on mobile?

nickcolley: I would say let's figure out what the correct behaviour is and have it consistent, and do it in 4.0.0

36degrees:
On mobile, we do not include any margin after bulleted or numbered list items – the margin is only added on tablet and above.

I'm wondering if we should be using a slightly smaller margin on mobile to be consistent and to better associate each bullet point with the lead-in paragraph:

```scss
  %govuk-list--spaced > li {
    margin-bottom: govuk-spacing(2);

      margin-bottom: govuk-spacing(3);
    }

    &:last-child {
      margin-bottom: govuk-spacing(0);
    }
  }
```

To my (non-designer) eyes the 15px on mobile otherwise seems excessive.

Thoughts @timpaul or @christopherthomasdesign?

| Current (15px) | Proposed (10px) |
| --- | --- |
| current | proposed |

src/govuk/core/_lists.scss
36degrees: Looking at this again, the `ul` has a bottom margin of 15px (mobile) or 20px (tablet, desktop) and the bottom margin of the last `li` collapses into it. As far as I can tell this means that zero-ing out the last list item has no effect and so this can probably be removed to simplify things?

timpaul: Sounds good!

timpaul:
I think that typically the internal padding and margins of textual components should shrink by a similar proportion to the text shrinkage, so yep - I'd agree with your suggestion. @christopherthomasdesign - what do you think? There's no crazy vertical rhythm thing we're doing that I've forgotten about is there?

christopherthomasdesign:
Looks good to me @timpaul, good suggestion @36degrees 👍

timpaul:
This proposal was approved by the Design System working group on 14 May 2020

frankieroberto:
Nice work! 🎊  I've used this style a few times in the past.

PR #1777, opened 2020-03-31
Remove sassdocs from the core repository

nickcolley:
This will be available at alphagov/govuk-frontend-docs.

I've put the existing version in the internet archive for reference:

https://web.archive.org/web/20200331131136/https://govuk-frontend-review.herokuapp.com/docs/

I wonder if we need specific developer docs for how to test your changes to sassdocs with the new repository to merge this?


36degrees:
I'd prefer to leave the sass-docs functionality in the repo somehow if possible, but only include it when running locally?

This will allow us to check that e.g. parameters and examples are being included correctly, without having to publish a pre-release and pull it into the frontend-docs repo.

Also, I don't think we should merge this until we know where the sass docs are going to live long-term, and ideally we should set up redirects so that anyone that's got this bookmarked can find the new documentation.

nickcolley:
Alright lets close this for now

PR #1778, opened 2020-03-31
Fix accordion underline hover state being removed when hovering plus/minus symbol

### LINKED ISSUE #1769: Accordion underline on hover is removed when over plus/minus symbol
## Steps to reproduce

1. Mouse over the plus/minus symbol for a section in the accordion.

## Expected result

The underline text-decoration on the section title should remain visible.

## Actual result

The underline text-decoration on the section title is removed.

accordion-hover-bug

36degrees:
At present, when you hover over a section the section title gets the affordance of an underline to show which element is being interacted with – except if you hover over the plus/minus 'control', in which case the underline disappears again.

This happens because the icon is appended to the heading, rather than the button, and so the button and icon are siblings, and the icon being absolutely positioned overlays the button.

(The only reason that you can click the control to toggle the section is that the click handler is bound to the heading, rather than the button. If the click handler was bound to the button, clicking the icon would not work.)

Append the icon to the button rather than the heading, which makes it part of the button, and thus hovering the icon will correctly trigger the button hover styles.

## Browser testing

- [x] IE8 (Windows) – does not get the hover state, but this matches the existing behaviour
- [x] IE9 (Windows)
- [x] IE10 (Windows)
- [x] IE11 (Windows)
- [x] Edge (Windows)
- [x] Chrome (Windows)
- [x] Firefox (Windows)
- [x] Safari (macOS)
- [x] Chrome (macOS)
- [x] Firefox (macOS)
- [x] Safari (iOS)
- [x] Chrome (iOS)
- [x] Chrome (Android)

Fixes #1769

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Fixes
 
 - [Pull request #1765: Import textarea from character count](https://github.com/alphagov/govuk-frontend/pull/1765).
+- [Pull request #1778: Fix accordion underline hover state being removed when hovering plus/minus symbol](https://github.com/alphagov/govuk-frontend/pull/1778).
 
 ## 3.6.0 (Feature release)
 
--- a/src/govuk/components/accordion/accordion.js
+++ b/src/govuk/components/accordion/accordion.js
@@ -135,7 +135,7 @@ Accordion.prototype.initHeaderAttributes = function ($headerWrapper, index) {
   icon.className = this.iconClass
   icon.setAttribute('aria-hidden', 'true')
 
-  $heading.appendChild(icon)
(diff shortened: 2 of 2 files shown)


36degrees:
This does change the behaviour in Firefox if using a custom colour scheme – the controls end up black (no matter what foreground colour is chosen, it seems) like the rest of the `button` (which is already an issue):

## Before



## After

nickcolley:
@36degrees that's unfortunate but I would argue if the darker text was a usability issue in general then that text is still the most important clickable part of the component. So given that we're improving the usability elsewhere feels like a decent compromise to me...

Issue #1789, opened 2020-04-23
Legacy code affects accordion height and touch area


hannalaakso:
## What
As noted in https://github.com/alphagov/govuk_publishing_components/pull/1386, the height and therefore the touch area in the accordion headings is smaller when there is legacy code on the page. 

This is because of a [reset](https://github.com/alphagov/govuk_elements/blob/84f485f688e7f796a4c44dd73b17177e8e39538c/assets/sass/elements/_reset.scss#L115) inherited from GOV.UK Elements: 

`font-size: inherit` 

changes the  `<h2>` font-size from 24px to 16px (16px is inherited from Elements `<body>` styles). This in turn reduces the height of the container and the touch area. 

This isn't visible in the font size of the headings as that's set in the enclosed `<button>` element.

To fix, we could explicitly set the font-size on the accordion `<h2>` elements if we detect Elements is included. 

## Why 
It might be harder for users to open the sections on mobile due to the reduced touch area.

## Who needs to know about this
Hanna, Ollie, GOV.UK

## Further detail
You can also see this in the frontend review app if you append `?legacy=1` to the URL to include the legacy styles.

Note that this isn't an issue with `<h2>` elements in general if they use the `govuk-heading-l` class. We should double check our components but I don't think this is an issue anywhere else.

PR #1796, opened 2020-05-04
Standardise accordion section headings font size (reduce height of section headings on mobile)

hannalaakso:
As reported in https://github.com/alphagov/govuk_publishing_components/pull/1386, the height of accordion sections on mobile doesn't match GOV.UK Design System if Elements stylesheet is included. This is because of the inclusion of GOV.UK Elements which resets `<h2>` to inherit font-size and line-height from `<body>`.   

We currently rely on browser defaults for font size on the accordion heading sections (although the text font size is actually set in the enclosed button and the font size in the section headings impacts only the height of the sections).  

This PR sets the font size on the section headings to match our standard font sizing for `<h2>` elements so that we don't rely on browser defaults. However this means that the height of the accordion sections is reduced by ~9px on mobile devices. 

<table role="table">
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td><a target="_blank" rel="noopener noreferrer" href="https://user-images.githubusercontent.com/5007934/81842295-af9ef880-9543-11ea-91e4-1eb53658c666.png">Before change</a></td>
<td><a target="_blank" rel="noopener noreferrer" href="https://user-images.githubusercontent.com/5007934/81842340-c3e2f580-9543-11ea-9356-737085e6a388.png">After change</a></td>
</tr>
</tbody>
</table>

#### Exact changes on different OS

| OS / browser        | Before | After |
|---------------------|--------|-------|
| iOS 13 / Safari     | 60px   | 50px  |
| iOS 12 / Safari     | 60px   | 49px  |
| Android 10 / Chrome | 58px   | 50px  |
| Android 9 / Chrome  | 58px   | 50px  |

WCAG 2.1 states that a [minimum target size must be 44 by 44 CSS pixels](https://www.w3.org/WAI/WCAG21/Understanding/target-size.html) which this will meet.

Fixes https://github.com/alphagov/govuk-frontend/issues/1789

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ For example:
 - [Pull request #1753: Make back link arrow consistent with breadcrumb component](https://github.com/alphagov/govuk-frontend/pull/1753)
 - [Pull request #1765: Import textarea from character count](https://github.com/alphagov/govuk-frontend/pull/1765).
 - [Pull request #1778: Fix accordion underline hover state being removed when hovering plus/minus symbol](https://github.com/alphagov/govuk-frontend/pull/1778).
+- [Pull request #1796: Standardise accordion section headings font size (reduce height of section headings on mobile)](https://github.com/alphagov/govuk-frontend/pull/1796).
 
 ## 3.6.0 (Feature release)
 
--- a/src/govuk/components/accordion/_index.scss
+++ b/src/govuk/components/accordion/_index.scss
@@ -19,6 +19,10 @@
   }
 
   .govuk-accordion__section-heading {
+    // Override browser defaults to ensure consistent element height
+    // Font size is set in .govuk-accordion__section-button
+    @include govuk-font(24);
+
     margin-top: 0; // Override browser default
     margin-bottom: 0; // Override browser default
   }


hannalaakso:
@andysellick Could you please take a look at this PR in relation to https://github.com/alphagov/govuk_publishing_components/pull/1386? It standardises the heading height to be consistent with GOV.UK Frontend `<h2>` styles. This means that the section heading has reduced on mobile because we don't have a clear user need or an accessibility requirement to have them taller than the GOV.UK Frontend `<h2>` style.

andysellick:
@hannalaakso if this standardises things and it's still above the recommended size then that's fine. 

We did have someone on our team who struggled to use the accordion at something like this size (which was also above the recommended 44x44) so maybe it's worth checking when you next have a round of user testing?

hannalaakso:
Thanks @andysellick. Talking IRL it sounds like this might have been more of an opinion, the user was able to complete the task. But it could be something to look out for if there's any research done. Would you be okay to leave a comment on the [accordion in the backlog](https://github.com/alphagov/govuk-design-system-backlog/issues/1) about this? I can also do it on your behalf.

PR #1802, opened 2020-05-05
Remove unused dependencies

36degrees:
Remove `gulp-data`, `gulp-to-markdown`, `gulp-nunjucks`, and `gulp-changed` as dependencies.

- `gulp-data` was used to generate README files automatically, introduced in 894ba63.
- `gulp-to-markdown` was then used to create the README files as markdown, introduced in 5ac9d18.
- `gulp-nunjucks` was added in dd0bd6b in order to compile the Nunjucks templates to HTML for the READMEs.

The functionality to create READMEs automatically was removed in b7dcac2, but the dependencies were not.

`gulp-changed` was introduced in 2b8df9d, but is unused since d44d522.


36degrees:
This also removes lodash.template as a dependency, which is currently being flagged as a security alert:

```
$ npm ls "lodash.template"
govuk-frontend-repository
└─┬ gulp-to-markdown@1.0.0
  └─┬ gulp-util@3.0.8
    └── lodash.template@3.6.2
```

Issue #1800, opened 2020-05-05
Remove dependency imports from files in the core and overrides layers


36degrees:
## What

Remove the imports of the settings, helpers and tools layers from files in core and overrides.

## Why

As proposed in https://github.com/alphagov/govuk-design-system-architecture/pull/22, we’re making changes to the way that the base layers (settings, tools and helpers) are imported within different parts of GOV.UK Frontend’s Sass, to reduce the time it takes to compile to CSS.

## Done when…

- [x] Imports of the settings, helpers and tools layers from files in core and overrides have been removed
- [x] Deprecation warnings introduced in #1798 have been removed
- [x] "Details of breaking change" section below completed to help with writing release notes

## Details of breaking change

- which users are affected: Anyone that is importing files from the core or overrides layers without first importing base.
- the change that’s been made: The files in these layers no longer import 'base' themselves
- changes users will have to make: Import 'base' before importing the files in core or overrides
- impact of users not making those changes: their Sass will not compile – they will likely see the error "Undefined mixin 'govuk-exports'" or "no mixin named govuk-exports"

PR #1807, opened 2020-05-12
Warn if importing core, overrides w/o dependencies

### LINKED ISSUE #1798: Deprecate importing files from the core and overrides layers without importing dependencies
## What

Add deprecation warnings if users import files from the core and overrides layers without first importing the settings, helpers and tools layers.

## Why

As proposed in https://github.com/alphagov/govuk-design-system-architecture/pull/22, we’re making changes to the way that the base layers (settings, tools and helpers) are imported within different parts of GOV.UK Frontend’s Sass, to reduce the time it takes to compile to CSS.

36degrees:
Add deprecation warnings if users import files from the core and overrides layers without first importing the settings, helpers and tools layers.

We'll [remove the imports of "base" from files in these layers in 4.0][1], so this will no longer work.

Closes #1798 

[1]: https://github.com/alphagov/govuk-frontend/issues/1800

### DIFF
--- a/src/govuk/core/_global-styles.scss
+++ b/src/govuk/core/_global-styles.scss
@@ -1,3 +1,7 @@
+@if not mixin-exists("govuk-exports") {
+  @warn "Importing items from the core layer without first importing `base` is deprecated, and will no longer work as of GOV.UK Frontend v4.0.";
+}
+
 @import "../base";
 
 @import "links";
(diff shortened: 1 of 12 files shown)


36degrees:
@m-green it'd be good to get your thoughts on the deprecation notice we're using:

> Importing items from the [core | overrides] layer without first importing `base` is deprecated, and will no longer work as of GOV.UK Frontend v4.0.

36degrees:
I think that's _probably_ OK as an edge case, as it should also be communicated in the release notes (which should probably be done as part of this PR) and in the release notes for 4.0 when it goes out? Thoughts?

PR #1816, opened 2020-05-26
Add input prefix and suffix

simonwhatley:
## What

Adds prefix and suffix elements to the input component.

## Why

A number of local and departmental design systems ([GOV.UK Publishing Components](https://components.publishing.service.gov.uk/component-guide/input/with_prefix_and_suffix), [HMRC](https://design.tax.service.gov.uk/hmrc-design-patterns/currency-input/), [MOJ](https://moj-design-system.herokuapp.com/components/currency-input) and [ONS](https://ons-design-system.netlify.app/patterns/numeric-values/)) are using prefixes and suffixes to indicate currencies and measurements.

#### HTML

```html
  <label class="govuk-label" for="input-with-prefix-attribute">
    Amount
  </label>
  <div class="govuk-input__wrapper">
    <span class="govuk-input__prefix" aria-hidden="true">£</span>
    <input class="govuk-input" id="input-with-prefix-attribute" name="amount" type="number">
  </div>
</div>
```

#### Macro

```
{% from "input/macro.njk" import govukInput %}

{{ govukInput({
  "label": {
    "text": "Amount"
  },
  "id": "input-with-prefix-attribute",
  "name": "amount",
  "type": "number",
  "prefix": {
    "text": "£"
  }
}) }}
```

image

#### HTML

```html
  <label class="govuk-label" for="input-with-suffix-attribute">
    Weight
  </label>
  <div class="govuk-input__wrapper">
    <input class="govuk-input" id="input-with-suffix-attribute" name="weight" type="number">
    <span class="govuk-input__suffix" aria-hidden="true">kg</span>
  </div>
</div>
```

#### Macro

```
{% from "input/macro.njk" import govukInput %}

{{ govukInput({
  "label": {
    "text": "Weight"
  },
  "id": "input-with-suffix-attribute",
  "name": "weight",
  "type": "number",
  "suffix": {
    "text": "kg"
  }
}) }}
```

image

Fixes https://github.com/alphagov/govuk-frontend/issues/1915

### DIFF
--- a/src/govuk/components/input/_index.scss
+++ b/src/govuk/components/input/_index.scss
@@ -8,6 +8,7 @@
 
     box-sizing: border-box;
     width: 100%;
+    min-width: 0;
     height: 40px;
     @if $govuk-typography-use-rem {
       height: govuk-px-to-rem(40px);
@@ -129,9 +130,15 @@
     border: $govuk-border-width-form-element solid $govuk-input-border-colour;
     background-color: govuk-colour("light-grey", $legacy: "grey-3");
 
+    @include govuk-media-query($until: tablet) {
+      line-height: 1.6;
+    }
+
     text-align: center;
     white-space: nowrap;
     cursor: default; // emphasise non-editable status of prefixes and suffixes
+
+    flex: 0 0 auto;
 
   }
 
--- a/src/govuk/components/input/_index.scss
+++ b/src/govuk/components/input/_index.scss
@@ -102,6 +102,16 @@
       // Hack to stop focus style being overlapped by the suffix
       z-index: 1;
     }
+
+    // Split prefix/suffix onto separate lines on narrow screens
+    @include govuk-media-query($until: mobile) {
+      display: block;
+
+      .govuk-input {
+        // Set max-width to override potential width override class on the input
+        max-width: 100%;
+      }
+    }
   }
 
   .govuk-input__prefix,
@@ -133,13 +143,31 @@
     cursor: default;
 
     flex: 0 0 auto;
+
+    // Split prefix/suffix onto separate lines on narrow screens
+    @include govuk-media-query($until: mobile) {
+      display: block;
+      height: 100%;
+      white-space: normal;
+    }
   }
 
   .govuk-input__prefix {
-    border-right: 0;
+    @include govuk-media-query($until: mobile) {
+      border-bottom: 0;
+    }
+    @include govuk-media-query($from: mobile) {
+      border-right: 0;
+    }
   }
 
+  // Split prefix/suffix onto separate lines on narrow screens
   .govuk-input__suffix {
-    border-left: 0;
+    @include govuk-media-query($until: mobile) {
+      border-top: 0;
+    }
+    @include govuk-media-query($from: mobile) {
+      border-left: 0;
+    }
   }
 }
--- a/src/govuk/components/input/_index.scss
+++ b/src/govuk/components/input/_index.scss
@@ -91,4 +91,56 @@
     max-width: 5.4ex;
   }
 
+  .govuk-input__wrapper {
+    display: flex;
+    display: -ms-flexbox; // sass-lint:disable-line no-duplicate-properties
+
+    .govuk-input {
+      display: inline-block;
+      flex: 0 1 auto;
+    }
+
+    .govuk-input:focus {
+      z-index: 1; // increase z-index on focus so it appears over the suffix
+    }
+  }
+
+  .govuk-input__prefix,
+  .govuk-input__suffix {
+    @include govuk-font($size: 19);
+
+    box-sizing: border-box;
+    display: inline-block;
+    width: auto;
+
+    min-width: 40px;
+    @if $govuk-typography-use-rem {
+      min-width: govuk-px-to-rem(40px);
+    }
+
+    height: 40px;
+    @if $govuk-typography-use-rem {
+      height: govuk-px-to-rem(40px);
+    }
+    margin-top: 0;
+
+    padding: govuk-spacing(1);
+
+    border: $govuk-border-width-form-element solid $govuk-input-border-colour;
+    background-color: govuk-colour("light-grey", $legacy: "grey-3");
+
+    text-align: center;
+    white-space: nowrap;
+    cursor: default; // emphasise non-editable status of prefixes and suffixes
+
+  }
+
+  .govuk-input__prefix {
+    border-right: 0;
+  }
+
+  .govuk-input__suffix {
+    border-left: 0;
+  }
+
 }
--- a/src/govuk/components/input/_index.scss
+++ b/src/govuk/components/input/_index.scss
@@ -157,7 +157,9 @@
       border-bottom: 0;
     }
     @include govuk-media-query($from: mobile) {
-      border-right: 0;
+      @include govuk-not-ie8 {
+        border-right: 0;
+      }
     }
   }
 
@@ -167,7 +169,9 @@
       border-top: 0;
     }
     @include govuk-media-query($from: mobile) {
-      border-left: 0;
+      @include govuk-not-ie8 {
+        border-left: 0;
+      }
     }
   }
 }
--- a/src/govuk/components/input/_index.scss
+++ b/src/govuk/components/input/_index.scss
@@ -8,7 +8,6 @@
 
     box-sizing: border-box;
     width: 100%;
-    min-width: 0;
     height: 40px;
     @if $govuk-typography-use-rem {
       height: govuk-px-to-rem(40px);
@@ -94,15 +93,14 @@
 
   .govuk-input__wrapper {
     display: flex;
-    display: -ms-flexbox; // sass-lint:disable-line no-duplicate-properties
 
     .govuk-input {
-      display: inline-block;
       flex: 0 1 auto;
     }
 
     .govuk-input:focus {
-      z-index: 1; // increase z-index on focus so it appears over the suffix
+      // Hack to stop focus style being overlapped by the suffix
+      z-index: 1;
     }
   }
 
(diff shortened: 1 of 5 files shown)


hannalaakso:
This work was reviewed by the Working Group in June, see https://github.com/alphagov/govuk-design-system/pull/1269#issuecomment-654282576 for their comments.

hannalaakso:
Implements the [design for very narrow viewports](https://github.com/alphagov/govuk-frontend/issues/1915#issuecomment-677705663). It only kicks in at 320px so I think that it's most likely to be seen by users who zoom in on a mobile device. I've tested this using the 'Aa' zoom button on iOS Safari: 

Screenshot 2020-09-01 at 10 45 22
Screenshot 2020-09-01 at 10 45 59
Screenshot 2020-09-01 at 10 46 41
Screenshot 2020-09-01 at 11 09 25
Screenshot 2020-09-01 at 10 47 44
Screenshot 2020-09-01 at 11 35 27

I've also tried to increase text scaling on Android Chrome but it mainly seems to work for text like paragraphs and headings, it doesn't do anything for form elements for instance. I'll raise a separate issue to investigate that.

Tested on 
✅ IE8 (See [commit](https://github.com/alphagov/govuk-frontend/pull/1816/commits/6f8a451b395b5aa9cdbd449eb4edb4b099688335) for details)
❓ IE9 (Doesn't support flex so some borders are incorrectly removed, however number of IE9 users is abysmally small so I'm not worried about it) 
✅ IE10
✅ IE11
✅ Windows 10 Edge 84
✅ Android 10 - Samsung galaxy S20: FF, Chrome, Samsung Internet
✅ Android 6 - Galaxy S7: FF, Chrome
✅ iOS 13 - iPhone XS: Safari
✅ iOS 11 - iPhone 6S: Safari and FF
✅ iOS 9 - iPhone 6S: Safari and FF
❓ iOS 8 - iPhone 6S: Safari and FF (Things aren't quite correctly aligned, probably a flex issue)
❌ iOS 7 - iPhone 5: Safari (Things aren't quite correctly aligned and the suffix drops onto next line - but this is way outside our support matrix and the component is still usable) 
✅ MacOS Safari 13.1
✅ MacOS Firefox 80.0 with inverted colours
✅ MacOS Chrome 84
✅ Windows 10 Firefox 80.0
✅ Windows 10 Edge 84

src/govuk/components/input/_index.scss
```diff
@@ -91,4 +91,88 @@
...
+  .govuk-input__wrapper {
+    display: flex;
+
+    .govuk-input {
+      flex: 0 1 auto;
+    }
+
+    .govuk-input:focus {
+      // Hack to stop focus style being overlapped by the suffix
+      z-index: 1;
```
hannalaakso, line 103: I'm not keen on this fix but I also can't find another way of stopping the suffix overlapping the input focus border.

src/govuk/components/input/input.yaml
```diff
@@ -54,15 +54,15 @@ params:
     - name: classes
       type: string
       required: false
-      description: Classes to add to the prefix tag
+      description: Classes to add to the prefix.
     - name: attributes
       type: object
       required: false
-      description: HTML attributes (for example data attributes) to add to the prefix tag.
+      description: HTML attributes (for example data attributes) to add to the prefix elementf.
```
vanitabarrett, line 61:
```suggestion
      description: HTML attributes (for example data attributes) to add to the prefix element.
```

src/govuk/components/input/template.test.js
```diff
@@ -382,4 +389,107 @@ describe('Input', () => {
...
     })
   })
+
+  describe('when it includes a prefix', () => {
+    it('renders the input wrapper', () => {
+      const $ = render('input', examples['with prefix'])
+
+      const $wrapper = $('.govuk-form-group > .govuk-input__wrapper')
+      expect($wrapper.length).toBeTruthy()
+    })
```
vanitabarrett, line 399: Minor, but the pattern in this file (and I've seen in other test files) is to leave a new line between each test

src/govuk/components/input/template.test.js
```diff
@@ -382,4 +389,107 @@ describe('Input', () => {
...
+    it('renders the prefix inside the wrapper', () => {
+      const $ = render('input', examples['with prefix'])
+
+      const $prefix = $('.govuk-form-group > .govuk-input__wrapper > .govuk-input__prefix')
+      expect($prefix.length).toBeTruthy()
```
vanitabarrett, line 404: Might be worth checking the prefix text is what we expect too

src/govuk/components/input/template.test.js
```diff
@@ -382,4 +389,107 @@ describe('Input', () => {
...
+      const $ = render('input', examples['with suffix'])
+
+      const $wrapper = $('.govuk-form-group > .govuk-input__wrapper')
+      expect($wrapper.length).toBeTruthy()
+    })
+    it('renders the suffix inside the wrapper', () => {
+      const $ = render('input', examples['with suffix'])
+
+      const $suffix = $('.govuk-form-group > .govuk-input__wrapper > .govuk-input__suffix')
+      expect($suffix.length).toBeTruthy()
```
vanitabarrett, line 451: Same comment here about checking the suffix text is as expected

vanitabarrett (changes requested):

src/govuk/components/input/template.njk
```diff
@@ -5,6 +5,18 @@
...
+
+{%- set inputTag %}
+<input class="govuk-input {%- if params.classes %} {{ params.classes }}{% endif %} {%- if params.errorMessage %} govuk-input--error{% endif %}" id="{{ params.id }}" name="{{ params.name }}" type="{{ params.type | default('text') }}"
+{%- if (params.spellcheck === false) or (params.spellcheck === true) %} spellcheck="{{ params.spellcheck }}"{% endif %}
+{%- if params.value %} value="{{ params.value}}"{% endif %}
+{%- if describedBy %} aria-describedby="{{ describedBy }}"{% endif %}
+{%- if params.autocomplete %} autocomplete="{{ params.autocomplete}}"{% endif %}
+{%- if params.pattern %} pattern="{{ params.pattern }}"{% endif %}
+{%- if params.inputmode %} inputmode="{{ params.inputmode }}"{% endif %}
+{%- for attribute, value in params.attributes %} {{ attribute }}="{{ value }}"{% endfor -%}>
```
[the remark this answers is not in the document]
hannalaakso: This block isn't modified in this PR (I think you were looking at part of commit [`be2c2de` (#1816)](https://github.com/alphagov/govuk-frontend/pull/1816/commits/be2c2debc2d3fec8e7a3b11c8b242ff1d44be18c) which is outdated).

src/govuk/components/input/_index.scss
```diff
@@ -91,4 +91,56 @@
...
+    }
+  }
+
+  .govuk-input__prefix,
+  .govuk-input__suffix {
+    @include govuk-font($size: 19);
+
+    box-sizing: border-box;
+    display: inline-block;
+    width: auto;
```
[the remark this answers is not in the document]
hannalaakso: Good spot 👍

src/govuk/components/input/template.test.js
hannalaakso: In my I had covered this with `allows prefix text to be passed whilst escaping HTML entities` but I've added a new test to check the text in the default example, it's more explicit 👍

src/govuk/components/input/template.test.js
hannalaakso: Same as above but with `allows suffix text to be passed whilst escaping HTML entities`.

vanitabarrett (approved):

36degrees:
We want to hold off merging this until we're ready to start on 4.0.

hannalaakso:
I'll also need to draft something for the changelog, will do that shortly

36degrees:
Minor, but can we have at least one example that does not use an input width modifier? This will allow us to verify that the input is unconstrained and 'full-width' by default.

Possibly most of the examples should omit the width modifier? In the same way the National Insurance Number examples are 'full-width' even though in real world usage you'd expect them to be constrained… most of the examples should probably combine the minimum number of 'features' possible? Although we would then want some examples that show that they do combine properly…

hannalaakso:
@36degrees I don't know if you saw the [feedback](https://github.com/alphagov/govuk-design-system/pull/1269#issuecomment-684966546) from @christopherthomasdesign which the change with the width modifiers was in response to? 

But maybe we mainly shouldn't use the width modifiers in the review app as the examples are out of context anyway..? Let me know if you have any thought on that. Definitely happy to include some examples without width modifiers in either case.

36degrees:
> @36degrees I don't know if you saw the [feedback](https://github.com/alphagov/govuk-design-system/pull/1269#issuecomment-684966546) from @christopherthomasdesign which the change with the width modifiers was in response to?

Yep, I did – and completely agree for the examples in the Design System.

> But maybe we mainly shouldn't use the width modifiers in the review app as the examples are out of context anyway..? Let me know if you have any thought on that. Definitely happy to include some examples without width modifiers in either case.

I think _generally_ we use examples that are focused on only the specific feature, and don't 'enable' anything else – for example, the National Insurance Number examples don't include the spellcheck attribute, the example that enables isPageHeading doesn't also add a class to style it (though I think we might be missing some examples for each label size!)

hannalaakso:
That all makes sense 👍 

However if you remove the width modifiers altogether you end up with this that to my eye looks pretty strange (I hadn't realised this up until point either):

Screenshot 2020-09-03 at 18 20 17

Or is it just me and they actually look okay like that? I've been staring at them for so long over the past weeks that I can't tell anymore 😆

edwardhorsford:
> That all makes sense 👍
> 
> However if you remove the width modifiers altogether you end up with this that to my eye looks pretty strange (I hadn't realised this up until point either):
> 
> Screenshot 2020-09-03 at 18 20 17
> 
> Or is it just me and they actually look okay like that? I've been staring at them for so long over the past weeks that I can't tell anymore 😆

Do these end up showing in a 2/3 column grid?

hannalaakso:
> Do these end up showing in a 2/3 column grid?

They get constrained by the grid so like I said I think we're okay.

hannalaakso:
Massive thanks for all the amazing work by @simonwhatley and his team on this component 🎉🙏

PR #1819, opened 2020-05-28
Add the ability to specify HTML for a navigation item

adamsilver:


36degrees:
@adamsilver Thanks for raising this.

Before this can be merged, we'll need to:

- add `html` to the `params` list in `/src/govuk/components/header/header.yaml`:

  https://github.com/alphagov/govuk-frontend/blob/9a77beeac392fa5bc63cf409bef971b935d2dc7c/src/govuk/components/header/header.yaml#L22-L30

- add a test for this behaviour to `/src/govuk/components/header/template.test.js`
- document this as an unreleased new feature in `/CHANGELOG.md`

Are those things that you're comfortable doing?

src/govuk/components/header/template.test.js
```diff
@@ -107,6 +107,20 @@ describe('header', () => {
       expect($firstItem.text()).toContain('Navigation item 1')
     })
 
+    it('renders navigation item with html', () => {
+      const $ = render('header', {
+        navigation: [
+          {
+            href: "#1",
```
36degrees, line 114: The tests are failing because the linter's being fussy and wants single quotes here.

```suggestion
            href: '#1',
```

PR #1827, opened 2020-06-03
Make dark grey darker

christopherthomasdesign:
The grey hint text continues to cause issues for users, with many struggling to read it or missing it because it's too pale. For an example see this [comment on the text input component](https://github.com/alphagov/govuk-design-system-backlog/issues/51#issuecomment-584223652).

This PR changes `govuk-colour("dark-grey")` from `#626a6e` to `#505a5f`. This is the one of the palest greys I could find that meets AAA contrast for body copy:

Screenshot 2020-05-20 at 11 07 15

I appreciate we have darkened this grey before and that we probably shouldn't go any darker than this because it'll get so close to being black as to be pointless. This grey has a bit less contrast with the black than before, though I'd argue that this shouldn't affect the usability of form elements for those that can't detect the difference in colour.

This is the simplest change we can make now with the fewest repercussions across the system. If we continue to see problems after this change, we clearly can't make it any darker so will look into an alternative approach like bolding labels and making hint text black.

This change will need to be accompanied by changes to the [colour](https://design-system.service.gov.uk/styles/colour/) page in the Design System.

Relates to https://github.com/alphagov/govuk-design-system/issues/1222

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+### New features
+
+#### Darkened secondary text colour
+
+We darkened `govuk-colour("dark-grey")` to improve the readability of hint text. It now has a contrast ratio of 7:1 and helps hint text meet the WCAG 2.1 (AAA) accessibility standard.
+
 ### Fixes
 
 We’ve made fixes to GOV.UK Frontend in the following pull requests:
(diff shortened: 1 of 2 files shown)


36degrees:
> This change will need to be accompanied by changes to the [colour](https://design-system.service.gov.uk/styles/colour/) page in the Design System.

The palette on the colour page is magically* generated from the Sass, so will automatically change once updated to use the latest version of frontend 💥 

<sub>* It's not really magic, it's just code, some gaffa tape and crossed fingers</sub>

Issue #1826, opened 2020-06-03
Prevent creating `<a>` without `href` in header service name


hannalaakso:
## What
As discussed in https://github.com/alphagov/govuk-frontend/pull/1825, `serviceName` has an implicit dependency on `serviceUrl`. If `serviceName` is set but `serviceUrl` isn't, the macro renders an `<a>` tag without `href`. 

If `serviceName` is set, we could:
- Check if `serviceUrl` is set and if it's not, render a `<span>` in place of `<a>` OR have a default value `/` for `href` 
- Improve the macro docs around this to make the dependency explicit.

## Why
To help our users to avoid outputting invalid markup / miss including the href. 

## Who needs to know about this
Devs, Mark?

## Done when
- [ ] It's not possible to output an `<a>` tag without `href` for service name using the macro 
- [ ] We've improved the macro documentation to make the dependency clearer

andymantell:
I'd be quite happy to pick this up following on from #1825 if you know which way you want to go. I'd probably be inclined to go for the `<span>` when the `serviceUrl` isn't present for flexibility - but I don't know whether you may want to discourage that as an option in the name of consistency? I.e. when a service name is present, it is always a link. But perhaps the documentation would be sufficient to encourage people to set a `serviceUrl`

As for making `serviceUrl` have a default value of `/` - this would be a breaking change I think? Anyone who's currently using the header without a `serviceUrl` will merrily be rendering an `<a>` with no href. Adding a `/` would change their output more significantly than rendering a `<span>`.

hannalaakso:
@andymantell We've discussed this and we think that the most sensible thing to do is to hold this back until the next breaking release since we don't have evidence that it's causing users problems. We will then go with the default value of `/` for the `href` which seems sensible given that the home page of the service will often be the root url. We will also flag this change in the release notes to ensure that teams are aware of the change.

The alternative to output a `<span>` is complicated by the fact that the existing class that styles the text is `govuk-header__link--service-name` which doesn't really makes sense for a span element. We would have therefore had to rename the class which is also a breaking change, and also to introduce some extra functionality in the template to swap the element out, even though we don't have any evidence that a service name without a link is something that teams need to do.

But thanks for offering to work on this, it's much appreciated 🙌

matthewmascord:
Hello, we have some interesting edge cases where there is a need to provide a serviceName in the header without providing a serviceUrl to that service.  Was there any rationale originally around forcing the serviceName to be hyperlinked?

[...]

PR #1834, opened 2020-06-11
Add default value of / for Header component's serviceUrl param

andymantell:
Add default value of `/` for Header component's `serviceUrl` param. Prior to this if you passed `serviceName` to the Header component but omitted `serviceUrl`, it would be rendered with an empty href.

Original issue at #1826

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+### Breaking changes
+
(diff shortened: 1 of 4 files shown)


andymantell:
Probably worth closing this now as the changes contradict @timpaul's comment here:

https://github.com/alphagov/govuk-frontend/issues/1826#issuecomment-759488404

PR #1842, opened 2020-06-19
Preserve the state of conditional reveals when navigating 'back' in the browser

### LINKED ISSUE #1552: Checked radio button with conditional reveal does not expand on page refresh in IE11
When interacting with the radio group and selecting the first time the conditional reveal is opened.

When you refresh the page, the first input is remembered as a checked input but the conditional reveal is not open.

I have only reproduced this on IE11 so far, but potentially might be an issue with error playback so we should construct an example of that too.

## Steps to reproduce

1. Use IE11 + JAWs 2018 (might be optional)
2. Go to https://govuk-frontend-review.herokuapp.com/components/radios/with-conditional-items/preview
3. Tab to first input
4. Press space to activate
5. Refresh page

### LINKED ISSUE #1794: Conditionals don’t show when using Back to return to a form with radio buttons (Chrome)
Steps to reproduce:

Using Chrome 80 or 81:

- Go to a form with radio buttons and a conditional (like this: https://design-system.service.gov.uk/components/radios/conditional-reveal/index.html)
- Select the radio button, show the field
- Hit continue
- Go back using the browser's back button
- The radio button is selected, but the field isn't showing

You can see this on:
https://www.find-postgraduate-teacher-training.service.gov.uk/

bug-radio-not-open-on-back-button

36degrees:
When the user navigates back to a previous page that includes a conditional reveal, the visible state of the conditionally revealed content is currently only preserved in some browsers.

Firefox / Safari
----------------

Recent versions of Firefox and Safari [use a Back-Forward Cache ('bfcache')][1] which preserves the entire page, which means the JavaScript is not re-evaluated but the browser 'remembers' that the conditional reveal was visible.

Internet Explorer
-----------------

In Internet Explorer the state of the form controls has not been restored at the point we currently call the init function, and so the conditional reveal state is not preserved.

To fix this, wait for the `DOMContentLoaded` event before syncing the visibility of the conditional reveals to their checkbox / radio state. As the checkbox state in IE8-11 has been restored before the `DOMContentLoaded` event fires, this fixes the preservation of the reveal state in Internet Explorer.

We already polyfill document.addEventListener and the DOMContentLoaded event for IE8, so we don't have to treat it as a special case.

Edge Legacy
-----------

In Edge 18, the state of the form controls does not seem to be restored at all when navigating back, so there is nothing to sync the conditional reveal state to 😢

Chromium based browsers (Chrome, Edge, Opera)
---------------------------------------------

In browsers based on Chromium 78 or older, the state of the form controls has been restored at the point we invoke the init function, and so the reveal state currently displays correctly, even without waiting for the `DOMContentLoaded` event.

In Chromium 79, the ['timing of restoring control state was changed so that it is done as a task posted just after DOMContentLoaded'][2]. This means that even after the `DOMContentLoaded` event the form state has not yet been restored. The recommended approach seems to be to wait for the `pageshow` event, however in Chromium 79 the form control state [has not been restored at the point this event fires either][3]! This was [fixed in Chromium 80][4].

So:

- in Chrome ≤ 78, the form control state is restored before the script is evaluated, before both the `DOMContentLoaded` and `pageshow` events.
- in Chrome = 79, the form control state is restored after both the `DOMContentLoaded` and `pageshow` events.
- in Chrome ≥ 80, the form control state is restored after the `DOMContentLoaded` but before the `pageshow` event.

Syncing the conditional reveal state after the `pageshow` event preserves the conditional reveal state except for Chromium 79 where it remains broken. Given that [Chrome 79's usage is already trending towards 0][5] (0.29% in May 2020) and there's seemingly no other event we can listen for that'll guarantee the state is restored, we'll accept that it'll remain broken in this specific version (affecting Chrome 79, Edge 79 and Opera 66).

The `pageshow` event is not supported in older browsers including IE8-10 so we need to listen for both. This means that we 'sync' twice in browsers that support `pageshow`, but the performance impact should be minimal.

[1]: https://www.chromestatus.com/feature/5815270035685376
[2]: https://chromium.googlesource.com/chromium/src.git/+/069ad65654655b7bdfb9b760f188395840bc4be4
[3]: https://bugs.chromium.org/p/chromium/issues/detail?id=1035662
[4]: https://crrev.com/069ad65654655b7bdfb9b760f188395840bc4be4
[5]: https://gs.statcounter.com/browser-version-market-share

## Browser testing (against 82c5bf0)

- [x] IE8 (Windows)
- [x] IE9 (Windows)
- [x] IE10 (Windows)
- [x] IE11 (Windows)
- [x] Edge (Windows)
- [x] Chrome (Windows)
- [x] Firefox (Windows)
- [x] Safari (macOS)
- [x] Chrome (macOS)
- [x] Firefox (macOS)
- [x] Safari (iOS)
- [x] Chrome (iOS)
- [x] Chrome (Android)

Fixes #1794
Fixes #1552

### DIFF
--- a/src/govuk/components/checkboxes/checkboxes.js
+++ b/src/govuk/components/checkboxes/checkboxes.js
@@ -52,9 +52,7 @@ Checkboxes.prototype.init = function () {
 }
 
 Checkboxes.prototype.syncAll = function () {
-  nodeListForEach(this.$inputs, function ($input) {
-    this.syncWithInputState($input)
-  }.bind(this))
+  nodeListForEach(this.$inputs, this.syncWithInputState.bind(this))
 }
 
 Checkboxes.prototype.syncWithInputState = function ($input) {
--- a/src/govuk/components/checkboxes/checkboxes.js
+++ b/src/govuk/components/checkboxes/checkboxes.js
@@ -32,19 +32,20 @@ Checkboxes.prototype.init = function () {
     $input.removeAttribute('data-aria-controls')
   })
 
-  if (document.readyState === 'loading') {
-    window.addEventListener('DOMContentLoaded', this.syncState.bind(this))
-  } else {
-    this.syncState()
-  }
-
   // When the page is restored after navigating 'back' in some browsers the
   // state of form controls is not restored until *after* the DOMContentLoaded
-  // event is fired, so we need to sync after the pageshow event is fired as
-  // well.
-  //
-  // (Older browsers don't have a pageshow event, so we do both.)
-  window.addEventListener('pageshow', this.syncState.bind(this))
+  // event is fired, so we need to sync after the pageshow event in browsers
+  // that support it.
+  if ('onpageshow' in window) {
+    window.addEventListener('pageshow', this.syncState.bind(this))
+  } else {
+    window.addEventListener('DOMContentLoaded', this.syncState.bind(this))
+  }
+
+  // Although we've set up handlers to sync state on the pageshow or
+  // DOMContentLoaded event, init could be called after those events have fired,
+  // for example if they are added to the page dynamically, so sync now too.
+  this.syncState()
 
   // Handle events
   $module.addEventListener('click', this.handleClick.bind(this))
--- a/src/govuk/components/checkboxes/checkboxes.js
+++ b/src/govuk/components/checkboxes/checkboxes.js
@@ -37,27 +37,27 @@ Checkboxes.prototype.init = function () {
   // event is fired, so we need to sync after the pageshow event in browsers
   // that support it.
   if ('onpageshow' in window) {
-    window.addEventListener('pageshow', this.syncState.bind(this))
+    window.addEventListener('pageshow', this.syncAll.bind(this))
   } else {
-    window.addEventListener('DOMContentLoaded', this.syncState.bind(this))
+    window.addEventListener('DOMContentLoaded', this.syncAll.bind(this))
   }
 
   // Although we've set up handlers to sync state on the pageshow or
   // DOMContentLoaded event, init could be called after those events have fired,
   // for example if they are added to the page dynamically, so sync now too.
-  this.syncState()
+  this.syncAll()
 
   // Handle events
   $module.addEventListener('click', this.handleClick.bind(this))
 }
 
-Checkboxes.prototype.syncState = function () {
+Checkboxes.prototype.syncAll = function () {
   nodeListForEach(this.$inputs, function ($input) {
-    this.setAttributes($input)
+    this.syncWithInputState($input)
   }.bind(this))
 }
 
-Checkboxes.prototype.setAttributes = function ($input) {
+Checkboxes.prototype.syncWithInputState = function ($input) {
   var inputIsChecked = $input.checked
   $input.setAttribute('aria-expanded', inputIsChecked)
 
@@ -74,7 +74,7 @@ Checkboxes.prototype.handleClick = function (event) {
   var isCheckbox = $target.getAttribute('type') === 'checkbox'
   var hasAriaControls = $target.getAttribute('aria-controls')
   if (isCheckbox && hasAriaControls) {
-    this.setAttributes($target)
+    this.syncWithInputState($target)
   }
 }
 
--- a/src/govuk/components/checkboxes/checkboxes.js
+++ b/src/govuk/components/checkboxes/checkboxes.js
@@ -1,5 +1,6 @@
 import '../../vendor/polyfills/Function/prototype/bind'
-import '../../vendor/polyfills/Event' // addEventListener and event.target normaliziation
+// addEventListener, event.target normalization and DOMContentLoaded
+import '../../vendor/polyfills/Event'
 import '../../vendor/polyfills/Element/prototype/classList'
 import { nodeListForEach } from '../../common'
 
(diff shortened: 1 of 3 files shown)


src/govuk/components/checkboxes/checkboxes.js
```diff
@@ -29,13 +30,30 @@ Checkboxes.prototype.init = function () {
     // If we have content that is controlled, set attributes.
     $input.setAttribute('aria-controls', controls)
     $input.removeAttribute('data-aria-controls')
-    this.setAttributes($input)
-  }.bind(this))
+  })
+
+  if (document.readyState === 'complete') {
```
hannalaakso, line 35: Should this be 
```
document.addEventListener('readystatechange', this.syncState.bind(this))
```
```
Checkboxes.prototype.syncState = function () {
  if (document.readyState === 'complete') {
```
because at the moment it's only evaluated once (when it's `loading`) so `complete` is never hit. But I might be misunderstanding the reason for this check.

36degrees: The intention of this logic is to handle cases where the script is being evaluated after the `DOMContentLoaded` event has already fired (in which case `document.readyState` would be `complete`).

This might be unnecessary – I am not sure if the way the script is included means this is guaranteed to run before the `DOMContentLoaded`event fires. However, given this code is probably going to end up being included in ways we can't anticipate (e.g. as part of a React component) I'm thinking it might be safer to try and cater for these scenarios?

Having done a bit more reading, I think that I possibly need to flip this in order to handle readyState being interactive, like in [this example in the MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Document/DOMContentLoaded_event#Checking_whether_loading_is_already_complete):

```js
if (document.readyState === 'loading') {
  window.addEventListener('DOMContentLoaded', this.syncState.bind(this))
} else {
  this.syncState()
}
```

36degrees:
> Testing on Chrome 79 (on Browserstack) I'm still seeing the bug persisting.

Ah, sorry, I didn't make this clear at all! There's a bug in Chrome 79 where the form control state is not guaranteed to have been restored before the `pageshow` event fires – see https://bugs.chromium.org/p/chromium/issues/detail?id=1035662. It was fixed in Chrome 80.

So:

- In Chrome 78 and below, the form control state is restored before the script is evaluated
- In Chrome 79, the form control state is restored some time after the script is evaluated, _and_ (potentially?) after the `pageshow` event fires.
- In Chrome 80, [a fix is introduced](https://chromium.googlesource.com/chromium/src/+/069ad65654655b7bdfb9b760f188395840bc4be4) which means that the form control state is restored after the script is evaluated but _before_ the `pageshow` event fires.

Based on the comments in the bug report, I don't think there's a reliable event we can listen for to fix the behaviour in Chrome 79. My current thinking is that we'd accept that it's broken in Chrome 79, but that [Chrome 79's usage is already trending towards 0](https://gs.statcounter.com/browser-version-market-share) (0.29% in May).

I'll update the commit message to make this clearer.

36degrees:
> I'll update the commit message to make this clearer.

Commit and PR description updated.

36degrees:
> * Go to `/components/checkboxes` and then `/components/checkboxes/with-conditional-items/preview`
> * Tick one of the checkboxes, fill in the conditional input.
> * Press back button and then forward again. The item is checked but not expanded.

I've been using the [feedback form](https://govuk-frontend-review-pr-1842.herokuapp.com/full-page-examples/feedback) ('Do you want a reply?') and [what is your nationality](https://govuk-frontend-review-pr-1842.herokuapp.com/full-page-examples/what-is-your-nationality) ('Citizen of a different country') to test radios and checkboxes respectively, as you can submit the form and then hit back 👍

src/govuk/components/checkboxes/checkboxes.js
hannalaakso: Thanks for digging into this. It seems that [older IE only support `complete` for `readyState`](https://caniuse.com/#search=readystate) so according to that `loading` wouldn't ever be true for those browsers.

36degrees: That page suggests that IE9-10 can fire the 'interactive' state too early, but also that they only support 'complete' – how can both of those be true?

It also associates the bug about Internet Explorer 9 and 10 with the IE11 block?

MDN seems to suggest that it's only IE8 that only supports `complete`: https://developer.mozilla.org/en-US/docs/Web/API/Document/readyState

36degrees: Raised an issue about the misleading notes – https://github.com/Fyrd/caniuse/issues/5489

src/govuk/components/checkboxes/checkboxes.js
```diff
@@ -29,13 +30,32 @@ Checkboxes.prototype.init = function () {
...
+  //
+  // (Older browsers don't have a pageshow event, so we do both.)
+  window.addEventListener('pageshow', this.syncState.bind(this))
 
   // Handle events
   $module.addEventListener('click', this.handleClick.bind(this))
 }
 
+Checkboxes.prototype.syncState = function () {
+  nodeListForEach(this.$inputs, function ($input) {
```
nickcolley, line 54: I can't unsubscribe from all notifications so sometimes I see interesting things, excellent write up, super interesting...

You might be able to simplify this to:

`nodeListForEach(this.$inputs, this.setAttributes($input).bind(this))`

36degrees: Thanks, @nickcolley! 👋

36degrees: I think it'd have to be:

```js
nodeListForEach(this.$inputs, this.setAttributes.bind(this))
```

which is possibly a little less clear – any thoughts @hannalaakso?

src/govuk/components/checkboxes/checkboxes.js
```diff
@@ -29,13 +30,32 @@ Checkboxes.prototype.init = function () {
...
     $input.removeAttribute('data-aria-controls')
-    this.setAttributes($input)
-  }.bind(this))
+  })
+
+  if (document.readyState === 'loading') {
+    window.addEventListener('DOMContentLoaded', this.syncState.bind(this))
+  } else {
+    this.syncState()
+  }
```
alex-ju, line 39: I like how these cover asynchronous loading too 👏

src/govuk/components/checkboxes/checkboxes.js
36degrees: So because of the bug in IE9-10 the readyState as being reported as `interactive` too early, which means that the state is not being restored correctly in those browsers.

I don't think it's safe to just wait for the `DOMContentLoaded` because we can't guarantee it hasn't already fired – an example might be if a user was initialising the checkboxes on a page that had already loaded, e.g. if they were added dynamically.

We could consider syncing within `init`, on `DOMContentLoaded` _and_ on the `pageshow` event, but wondering if it's worth it given that we're only aiming for 'functional' support of IE9 and IE10 and this is something of an edge case?

36degrees: At the very least, I'll need to revise the commit message and PR description as this it currently implies that it fixes IE8-11, which is not the case.

36degrees: Actually, maybe the correct approach is to:

- always sync within the `init` function
- sync on the `pageshow` event OR the `DOMContentLoaded` if the browser does not support the `pageshow` event

Trying that now.

36degrees:
I think this is good to go, pending a review from @hannalaakso 🙏

src/govuk/components/checkboxes/checkboxes.js
hannalaakso: This sounds like a good solution, 🙌 for the investigation and write up @36degrees.

Issue #1809, opened 2020-05-15
Character count fails if ID contains characters that have a special meaning in CSS


36degrees:
Because we interpolate the textarea ID into a querySelector call without escaping it, the character count can fail if the ID contains characters that have a special meaning in CSS:

https://github.com/alphagov/govuk-frontend/blob/387c33d5280745523a9ee0fd1da961be584137e6/src/govuk/components/character-count/character-count.js#L8-L10

```html
<div class="govuk-character-count" data-module="govuk-character-count" data-maxlength="10">
  <div class="govuk-form-group">
  <label class="govuk-label" for="docs[more-detail]">
    Can you provide more detail?
  </label>
  <textarea class="govuk-textarea govuk-js-character-count" id="docs[more-detail]" name="more-detail" rows="5" aria-describedby="docs[more-detail]-info"></textarea>
</div>

  <span id="docs[more-detail]-info" class="govuk-hint govuk-character-count__message" aria-live="polite">
  You can enter up to 10 characters
</span>

</div>
```

```
Uncaught DOMException: Failed to execute 'querySelector' on 'Element': '[id=docs[more-detail]-info]' is not a valid selector.
    at new CharacterCount (http://localhost:3000/public/all.js:1501:34)
    at http://localhost:3000/public/all.js:2366:5
    at NodeList.forEach (<anonymous>)
    at nodeListForEach (http://localhost:3000/public/all.js:14:18)
    at Object.initAll (http://localhost:3000/public/all.js:2365:3)
    at http://localhost:3000/components/character-count/special-chars/preview:69:32
```

36degrees:
This is similar to #1808, and the solution is probably the same – wrapping the call with CSS.escape – although we should be able to use a simpler ID selector rather than using the attribute syntax:

```js
this.$countMessage = $module.querySelector('#' + CSS.escape(this.$textarea.id) + '-info') 
```

36degrees:
Based on #1843 fixing this, changing the label from 'Days' to 'Hours' as I think we have a solution that works that doesn't involve polyfilling `CSS.escape`.

36degrees:
I think this might actually have been fixed by quoting the ID attribute which was done in #2080, except _possibly_ if the ID itself contains double quotes?

PR #1843, opened 2020-06-19
Fix conditional reveals failing if the id of the revealed element contains special characters

### LINKED ISSUE #1808: Conditional reveals fail silently if id contains characters characters that have a special meaning in CSS

36degrees:
If the conditionally revealed element has an ID that contains characters that have a special meaning in CSS, such as `[]`, the call to `querySelector` fails because as it's being evaluated as a selector the ID needs to be escaped:

```
Uncaught DOMException: Failed to execute 'querySelector' on 'Element': '#conditional-person-1[how-contacted]-2' is not a valid selector.
    at Radios.<anonymous> (http://localhost:3000/public/all.js:1978:31)
    at NodeList.forEach (<anonymous>)
    at nodeListForEach (http://localhost:3000/public/all.js:14:18)
    at Radios.init (http://localhost:3000/public/all.js:1973:3)
    at http://localhost:3000/public/all.js:2384:24
    at NodeList.forEach (<anonymous>)
    at nodeListForEach (http://localhost:3000/public/all.js:14:18)
    at Object.initAll (http://localhost:3000/public/all.js:2383:3)
    at http://localhost:3000/components/radios/with-conditional-items-and-square-brackets/preview:98:32
```

In some cases the selector may be valid but still incorrect, like `#conditional-rejectionReasons[candidate-actions]` which would select any element with ID conditional-rejectionReasons, also having the attribute `candidate-actions`.

We can avoid all of this by using `document.getElementById` instead, avoiding the ID being evaluated as a selector. The downside is that we can't constrain the scope to the `$module`. However, I don't think this should cause any issues (especially as the ID should be unique) and this avoids alternatives like using `CSS.escape` to escape the ID, which we'd then need to polyfill.

Fixes #1808

### DIFF
--- a/src/govuk/components/checkboxes/checkboxes.js
+++ b/src/govuk/components/checkboxes/checkboxes.js
@@ -41,7 +41,7 @@ Checkboxes.prototype.setAttributes = function ($input) {
   $input.setAttribute('aria-expanded', inputIsChecked)
 
   var $content = document.getElementById($input.getAttribute('aria-controls'))
-  if ($content) {
+  if ($content && $content.classList.contains('govuk-checkboxes__conditional')) {
     $content.classList.toggle('govuk-checkboxes__conditional--hidden', !inputIsChecked)
   }
 }
--- a/src/govuk/components/checkboxes/checkboxes.js
+++ b/src/govuk/components/checkboxes/checkboxes.js
@@ -22,7 +22,7 @@ Checkboxes.prototype.init = function () {
 
     // Check if input controls anything
     // Check if content exists, before setting attributes.
-    if (!controls || !$module.querySelector('#' + controls)) {
+    if (!controls || !document.getElementById(controls)) {
       return
     }
 
@@ -40,7 +40,7 @@ Checkboxes.prototype.setAttributes = function ($input) {
   var inputIsChecked = $input.checked
   $input.setAttribute('aria-expanded', inputIsChecked)
 
-  var $content = this.$module.querySelector('#' + $input.getAttribute('aria-controls'))
+  var $content = document.getElementById($input.getAttribute('aria-controls'))
   if ($content) {
     $content.classList.toggle('govuk-checkboxes__conditional--hidden', !inputIsChecked)
   }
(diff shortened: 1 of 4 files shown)


36degrees:
Raised as a draft as I'm worried that my Friday afternoon brain is missing a reason why changing the scope from `$module` to the `document` would be an issue…

36degrees:
If we're happy with this approach it'd probably make sense to apply the same change to the character count, which should fix #1809:

https://github.com/alphagov/govuk-frontend/blob/4a9d392a11ad89bde25f45bac880a0a37814f7ef/src/govuk/components/character-count/character-count.js#L9

hannalaakso:
> Raised as a draft as I'm worried that my Friday afternoon brain is missing a reason why changing the scope from $module to the document would be an issue…

I think this would be okay but I have a couple of small reservations.

It is possible that a third party plugin etc. could happen to be using the same ID in the page markup and our script would then grab that element 🤷 Should we at least trust these elements that use `getElementById` less and do some checking in our script to make sure that it's the element we think it is? Our component wouldn't work correctly but at least we could catch a script error possibly resulting from trying to manipulate a DOM element that's the wrong one. 

Also, `querySelector` is more flexible than `getElementById`. This sentence (from [O'Reilly](https://www.oreilly.com/library/view/dom-enlightenment/9781449344498/ch04.html#selecting_a_specific_element_node)) explains it 
> The querySelector() method permits a parameter in the form of a CSS selector syntax. Basically, you can pass this method a CSS3 selector (e.g., #score>tbody>tr>td:nth-of-type(2)), which it will use to select a single element in the DOM.

but I believe that with the current implementation you wouldn't be able to pass through `score>tbody>tr>td:nth-of-type(2)` for the ID since it contains special characters without an error being thrown? Is there a CSS selector that could have previously worked with `querySelector` that now won't if we use `getElementById`? The answer is probably no.

We should consider changing [the code comment about scope](https://github.com/alphagov/govuk-frontend/blob/063cd8e2470b62b824c6e50ca66342ac7a95d2d8/src/govuk/all.js#L16) to say that wherever IDs are concerned they are queried from the whole document (I don't think we have any other docs about `scope` but I might be having brain freeze).

36degrees:
Let's play it safe and make this change in 4.0.

PR #1847, opened 2020-06-26
Improve documentation and consistency in radios and checkboxes components

36degrees:
Building on top of #1842:

- Improve consistency between the radios and checkboxes components
- Simplify and improve code comments

### DIFF
--- a/src/govuk/components/checkboxes/checkboxes.js
+++ b/src/govuk/components/checkboxes/checkboxes.js
@@ -53,7 +53,7 @@ Checkboxes.prototype.syncAll = function () {
 Checkboxes.prototype.syncWithInputState = function ($input) {
   var $content = this.$module.querySelector('#' + $input.getAttribute('aria-controls'))
 
-  if ($content) {
+  if ($content && $content.classList.contains('govuk-checkboxes__conditional')) {
     var inputIsChecked = $input.checked
 
     $input.setAttribute('aria-expanded', inputIsChecked)
--- a/src/govuk/components/checkboxes/checkboxes.js
+++ b/src/govuk/components/checkboxes/checkboxes.js
@@ -51,13 +51,13 @@ Checkboxes.prototype.syncAll = function () {
 }
 
 Checkboxes.prototype.syncWithInputState = function ($input) {
-  var $content = this.$module.querySelector('#' + $input.getAttribute('aria-controls'))
+  var $target = this.$module.querySelector('#' + $input.getAttribute('aria-controls'))
 
-  if ($content && $content.classList.contains('govuk-checkboxes__conditional')) {
+  if ($target && $target.classList.contains('govuk-checkboxes__conditional')) {
     var inputIsChecked = $input.checked
 
     $input.setAttribute('aria-expanded', inputIsChecked)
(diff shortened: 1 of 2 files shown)


src/govuk/components/checkboxes/checkboxes.js
```diff
@@ -56,11 +51,12 @@ Checkboxes.prototype.syncAll = function () {
 }
 
 Checkboxes.prototype.syncWithInputState = function ($input) {
-  var inputIsChecked = $input.checked
-  $input.setAttribute('aria-expanded', inputIsChecked)
-
   var $content = this.$module.querySelector('#' + $input.getAttribute('aria-controls'))
+
   if ($content) {
```
hannalaakso, line 56: Additional suggestion for increasing consistency: [radios](https://github.com/alphagov/govuk-frontend/blob/7680cb300ffdab8cf5ae7b2e4fdf01fcdcc08d09/src/govuk/components/radios/radios.js#L61) have a check `$content.classList.contains('govuk-radios__conditional')` on this line.

src/govuk/components/checkboxes/checkboxes.js
```diff
@@ -13,22 +13,18 @@ Checkboxes.prototype.init = function () {
   var $module = this.$module
   var $inputs = this.$inputs
 
-  /**
-  * Loop over all items with [data-controls]
-  * Check if they have a matching conditional reveal
-  * If they do, assign attributes.
-  **/
```
hannalaakso, line 20: Does this comment definitely need removing? It still feels useful to me - maybe excluding the line 
```
Loop over all items with [data-controls]
```

Or a different solution might be to rename
```
var target 
```
to
```
var conditionalReveal 
```

36degrees: I removed it as there's a lot of duplication between this comment and the comments within the callback function.

I've clarified the comments within the function, and I think they cover everything that was in this comment except the fact that it's a loop?

In terms of renaming the variable, the target element that's being associated is _probably_ (hopefully!) a conditional reveal, but I don't think we know that for sure at this point.

src/govuk/components/checkboxes/checkboxes.js
36degrees: Good spot. I think we were going to do this in #1843 but can include that here now instead.

src/govuk/components/checkboxes/checkboxes.js
hannalaakso: As discussed earlier, we could add a general comment to indicate that the JavaScript here sets up the relationship of the `aria-controls` and conditional reveals.  

>  the target element that's being associated is probably (hopefully!) a conditional reveal, but I don't think we know that for sure at this point.

True. We could have a comment above that line to convey the _intention_ to extract the conditional reveal attribute (but it might not be necessary to add that if we add a comment elsewhere about the relationship).

36degrees: I've added JSDoc comments to each function, aiming to explain what the JS is doing.

@hannalaakso I'd value any thoughts or feedback – do you think this addresses your concerns?

hannalaakso: Yes the comments and clarified function names are super helpful 👍

PR #1850, opened 2020-07-01
Add example of Slack release message

m-green:
This adds an example of the Slack message we send when we publish a release, in the release documentation.

### DIFF
--- a/docs/releasing/publishing.md
+++ b/docs/releasing/publishing.md
@@ -67,7 +67,7 @@
 npm logout
 ```
 
-17. Publish a short summary of the release in the cross-government and GDS #govuk-design-system Slack channels. For example:
+17. Post a short summary of the release in the cross-government and GDS #govuk-design-system Slack channels. For example:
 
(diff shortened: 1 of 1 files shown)


docs/releasing/publishing.md
```diff
@@ -67,8 +67,9 @@
 npm logout
 ```
 
-17. Send a message to our users in both the X-GOV and GDS #govuk-design-system slack channels that indicates
-there's a new release with a short summary.
+17. Post a short summary of the release in the cross-government and GDS #govuk-design-system Slack channels. For example:
+
+    🚀 We’ve just released GOV.‌UK Frontend v3.7.0. It's now easier and faster to use our Sass. We've also made improvements to back links, breadcrumbs, lists and the header. [https://github.com/alphagov/govuk-frontend/releases/tag/v3.7.0](https://github.com/alphagov/govuk-frontend/releases/tag/v3.7.0)
```
hannalaakso, line 72: Could we also add something about acknowledging the help we got from different contributors.

m-green: Ah yes, sorry, I forgot about that - will add! Thanks Hanna.

m-green: Added in f92358b! Just in the Slack comms section for now so we keep this update in scope.

Issue #1835, opened 2020-06-11
Consider changing govukHint to wrap its content in a DIV rather than a SPAN


matthewmascord:
Services may have a valid reason to include hint copy that includes block level elements such as P or UL. For example, on the Gov.UK cookies page, https://www.gov.uk/help/cookies

At the moment the govukHint component wraps the markup it accepts in a SPAN and this causes HTML validation errors if developers include block level elements because block level elements are not allowed within a SPAN.

https://github.com/alphagov/govuk-frontend/tree/master/src/govuk/components/hint

36degrees:
This also came up in https://github.com/alphagov/govuk-design-system/issues/778#issuecomment-561130513

PR #1855, opened 2020-07-08
Hint component can render block-level elements as valid HTML

vanitabarrett:
## What

- Switch the hint component wrapping element from span to div

## Why
We allow HTML to be passed to the component, and it seems reasonable that someone may want to pass (for example) 2 paragraphs in. At the moment this would render correctly but would be invalid HTML.

Preview: https://govuk-frontend-review-pr-1855.herokuapp.com/components/hint (no visual changes expected)

Fixes https://github.com/alphagov/govuk-frontend/issues/1835

### DIFF
--- a/src/govuk/components/hint/_index.scss
+++ b/src/govuk/components/hint/_index.scss
@@ -16,7 +16,7 @@
   // is unlikely that the default or govuk-label--s class would be used in this
   // case.
 
-  // This adjustment will not work in browsers that do not support :not(). 
+  // This adjustment will not work in browsers that do not support :not().
   // Users with these browsers will see the default size margin (5px larger).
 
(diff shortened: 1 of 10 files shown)


src/govuk/components/hint/_index.scss
```diff
@@ -2,8 +2,6 @@
   .govuk-hint {
     @include govuk-font($size: 19);
 
-    display: block;
```
36degrees, line 5: I think we might want to keep this in for now – if we don't, users will need to make the change to their HTML markup (changing hints from spans to divs) which would then count as a breaking change – otherwise their hints will appear inline.

What do you think about leaving this in, but raising an issue to remove this later, and attach it to the 4.0 milestone?

vanitabarrett: @36degrees How's this https://github.com/alphagov/govuk-frontend/issues/1856 ?

Issue #1832, opened 2020-06-09
Add 'spellcheck' option to text macros


joelanman:
# What

We could add a 'spellcheck' option to text macros

# Why

We recommend turning it off in some contexts, but we do it via `attributes`. If this is a common option people should be setting, might be nicer to set it direct like we do other common options like `id`, `name` etc.

Example with spellcheck=false:

https://design-system.service.gov.uk/components/text-input/#numbers

PR #1859, opened 2020-07-10
Add spellcheck option

vanitabarrett:
Resolves https://github.com/alphagov/govuk-frontend/issues/1832

## What
Adds a Boolean `spellcheck` param to the input and textarea components

## Why
Currently, users can add a spellcheck attribute to these components by using the `attributes` param and passing the String value, for example:

```
attributes: {
    spellcheck: "true"
}
```

We want to make the spellcheck attribute more explicit. Making it a Boolean also leaves less room for human error.

Screenshot 2020-07-10 at 10 58 54

Examples:

- [input](https://govuk-frontend-review-pr-1859.herokuapp.com/components/input)
- [textarea](https://govuk-frontend-review-pr-1859.herokuapp.com/components/textarea)
- [full page example using spellcheck:false](https://govuk-frontend-review-pr-1859.herokuapp.com/full-page-examples/update-your-account-details)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@
 
 We darkened `govuk-colour("dark-grey")` to improve the readability of hint text. It now has a contrast ratio of 7:1 and helps hint text meet the WCAG 2.1 (AAA) accessibility standard.
 
+#### Add spellcheck parameter to input and textarea components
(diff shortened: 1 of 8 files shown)


joelanman:
nice work!

PR #1860, opened 2020-07-14
Update dependencies

36degrees:
A little housekeeping to try and stay on top of our dependencies 🧹 


36degrees:
> Out of interest, what are we using/including govuk_frontend_toolkit for?

If you add `?legacy=1` to the end of the URL in the review app ([like this](http://govuk-frontend-review.herokuapp.com/components/button/?legacy=1)) it'll enable [compatibility mode](https://frontend.design-system.service.gov.uk/compatibility-mode/#use-with-our-old-frameworks-or-colours) and include the stylesheets from GOV.UK Elements (which relies on GOV.UK Frontend Toolkit) and GOV.UK Template.

This allows us to check that the components in GOV.UK Frontend look and behave correctly when included 'alongside' the old frameworks, for example if a service is gradually migrating from one to the other.

Some of the full page examples (like '[Applicant details](http://govuk-frontend-review.herokuapp.com/full-page-examples/applicant-details)') also use a mix of 'components' from the old repos and new components from GOV.UK Frontend.

vanitabarrett:
@36degrees Awesome! Thanks for explaining, that's really useful to know about

Issue #1848, opened 2020-06-26
Character count can show incorrect count when going 'back', or if page is discarded and restored


36degrees:
The character count can show the incorrect count when navigating back in the browser, or if the page is discarded and restored.

This is because the component is initialised before the form control state is restored, and is very similar to the issue described in #1794 and fixed in #1842.

Because of the similarity with those issues, I suspect this will affect Chromium browsers > 79 and Internet Explorer 8-11.

PR #1868, opened 2020-07-16
Preserve the state of character count when navigating 'back'

vanitabarrett:
Fixes https://github.com/alphagov/govuk-frontend/issues/1848

## What
When the user navigates back to a previous page that includes a character count, the character count message should reflect the text currently in the textarea.

## Why
Fix based on https://github.com/alphagov/govuk-frontend/pull/1842
Previously, when navigating back to a page with a character count, the character count message wouldn't update (reflect the text currently in the textarea) until the textarea received focus.

Current behaviour: http://govuk-frontend-review.herokuapp.com/full-page-examples/feedback
Preview new behaviour: https://govuk-frontend-review-pr-1868.herokuapp.com/full-page-examples/feedback

## Notes
On slow connections there may be a flash of the default character count text before it updates. From testing with a 4x CPU slowdown, the default text may show for 0.03s before it updates. Given how small this is even with slowdown, we've decided to leave it as-is. If we decide this is an issue in the future, we'd also need to look at other components (e.g: radios and checkboxes) which are affected in a similar way.

## Browser testing
 
- [x] IE8 (Windows)
- [x] IE9 (Windows)
- [x] IE10 (Windows)
- [x] IE11 (Windows)
- [x] Edge (Windows)
- [x] Chrome (Windows)
- [x] Firefox (Windows)
- [x] Safari (macOS)
- [x] Chrome (macOS)
- [x] Firefox (macOS)
- [x] Safari (iOS)
- [x] Chrome (iOS)
- [x] Chrome (Android)

### DIFF
--- a/src/govuk/components/character-count/character-count.js
+++ b/src/govuk/components/character-count/character-count.js
@@ -50,6 +50,20 @@ CharacterCount.prototype.init = function () {
   // Remove hard limit if set
   $module.removeAttribute('maxlength')
 
+  // When the page is restored after navigating 'back' in some browsers the
+  // state of the character count is not restored until *after* the DOMContentLoaded
+  // event is fired, so we need to sync after the pageshow event in browsers
+  // that support it.
+  if ('onpageshow' in window) {
+    window.addEventListener('pageshow', this.sync.bind(this))
+  } else {
+    window.addEventListener('DOMContentLoaded', this.sync.bind(this))
+  }
+
+  this.sync()
+}
+
+CharacterCount.prototype.sync = function () {
   this.bindChangeEvents()
   this.updateCountMessage()
 }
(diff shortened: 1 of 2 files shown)


src/govuk/components/character-count/character-count.js
```diff
@@ -50,6 +50,20 @@ CharacterCount.prototype.init = function () {
...
+  if ('onpageshow' in window) {
+    window.addEventListener('pageshow', this.syncAll.bind(this))
+  } else {
+    window.addEventListener('DOMContentLoaded', this.syncAll.bind(this))
+  }
+
+  this.syncAll()
+}
+
+CharacterCount.prototype.syncAll = function () {
```
36degrees, line 66: Wondering if `syncAll` is the right name for this function, as unlike the radios and checkboxes there is only one character count per 'module', so there's only really one thing to 'sync'… thoughts?

src/govuk/components/character-count/character-count.js
```diff
@@ -50,6 +50,20 @@ CharacterCount.prototype.init = function () {
...
+  if ('onpageshow' in window) {
+    window.addEventListener('pageshow', this.sync.bind(this))
+  } else {
+    window.addEventListener('DOMContentLoaded', this.sync.bind(this))
+  }
+
+  this.sync()
+}
+
+CharacterCount.prototype.sync = function () {
```
36degrees, line 66: 👍

Issue #1139, opened 2019-01-16
Test that documented component options match the templates


36degrees:
## What

Add tests that validate the documented options within our component YAML files (like [the button YAML file](https://github.com/alphagov/govuk-frontend/blob/f704053f8544e8d88acec68ca448b4cb992a76d2/src/components/button/button.yaml)) match the actual options used within the template / macro.

If the two do not match, the tests should fail.

There is as a ['work-in-progress' branch](https://github.com/alphagov/govuk-frontend/tree/test-definitions) for this work which currently does this, but needs some refactoring / documentation before it can be merged.

## Why

Currently we rely on human review to ensure that the documented options match reality. During a recent [firebreak](https://insidegovuk.blog.gov.uk/2018/05/03/firebreaks-on-gov-uk/) project, we explored whether we could do anything to catch these in our CI process.

The work found a number of inaccuracies in the documentation options:

- `attributes` not documented on the breadcrumbs and select components
- `describedBy` mistakenly documented on textarea and character count components
- `footer` component YAML misformed using `arguments` instead of `params`, which means the `meta` object is not documented correctly in the Design System
- `items[].name` mistakenly documented on radios – unlike checkboxes, this can only be set globally as having a set of radios with different names doesn't make any sense
- `title` not documented on the tabs component

This highlights the need to automate these checks, so we can make sure our documentation is correct.

colinrotherham:
Whilst not fully related, we've started seeing failures on our Design System v5 preview:
https://github.com/alphagov/govuk-design-system/actions/runs/5956674573/job/16157972971

Design System HTML generates duplicate slugs for nested option with the same name

E.g. [**Summary list**](https://design-system.service.gov.uk/components/summary-list/) `rows[].actions` and `card.actions` are duplicated in v5

PR #1869, opened 2020-07-17
Add missing spellcheck param to character count

vanitabarrett:
## What
Add missing spellcheck param for the character count component.

## Why
We added a new spellcheck param to the textarea component in https://github.com/alphagov/govuk-frontend/pull/1859 However, we forgot to add the param to the character count component, which meant it couldn't be passed through to the textarea.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,9 +10,10 @@ We darkened `govuk-colour("dark-grey")` to improve the readability of hint text.
 
 #### Add spellcheck parameter to input and textarea components
 
-Optional parameter added to the input and textarea components to enable or disable the spellcheck attribute
+Optional parameter added to the input, textarea and character count components to enable or disable the spellcheck attribute
 
 ([PR #1859](https://github.com/alphagov/govuk-frontend/pull/1859))
+([PR #1869](https://github.com/alphagov/govuk-frontend/pull/1869))
 
 ### Fixes
 
--- a/src/govuk/components/character-count/template.njk
+++ b/src/govuk/components/character-count/template.njk
@@ -10,6 +10,7 @@
     name: params.name,
     describedBy: params.id + '-info',
     rows: params.rows,
+    spellcheck: params.spellcheck,
     value: params.value,
     formGroup: params.formGroup,
     classes: 'govuk-js-character-count' + (' govuk-textarea--error' if params.errorMessage) + (' ' + params.classes if params.classes),
--- a/src/govuk/components/character-count/character-count.yaml
+++ b/src/govuk/components/character-count/character-count.yaml
@@ -59,6 +59,10 @@ params:
   type: object
   required: false
   description: HTML attributes (for example data attributes) to add to the textarea.
+- name: spellcheck
+  type: boolean
+  required: false
+  description: Optional field to enable or disable the spellcheck attribute on the character count.
 - name: countMessage
   type: object
   required: false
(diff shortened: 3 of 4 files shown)


src/govuk/components/character-count/template.njk
```diff
@@ -10,6 +10,7 @@
     name: params.name,
     describedBy: params.id + '-info',
     rows: params.rows,
+    spellcheck: params.spellcheck,
```
36degrees, line 13: Because the character count sort of 'extends' the textarea component rather than 'uses it', `spellcheck` needs to be added as a parameter in `character-count.yaml` as well (just like you added it to the textarea component):

https://github.com/alphagov/govuk-frontend/blob/ce8e22355e32af4d8f3c876779acd26128401842/src/govuk/components/textarea/textarea.yaml#L10-L13

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ We darkened `govuk-colour("dark-grey")` to improve the readability of hint text.
 Optional parameter added to the input and textarea components to enable or disable the spellcheck attribute
```
36degrees, line 13: Do you think it'd be worth adding character count to this list too?

```suggestion
Optional parameter added to the input, textarea and character count components to enable or disable the spellcheck attribute
```

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ We darkened `govuk-colour("dark-grey")` to improve the readability of hint text.
...
 
 ([PR #1859](https://github.com/alphagov/govuk-frontend/pull/1859))
+([PR #1869](https://github.com/alphagov/govuk-frontend/pull/1869))
```
36degrees, line 16: 👍

src/govuk/components/character-count/template.njk
vanitabarrett: That seems correct - although I managed to test the change worked without this, which I guess makes me question what these params do...? Is it more documentation than anything which will fail if something isn't present?

36degrees: Great question!

These parameters are [included as a JSON file in the released package](https://github.com/alphagov/govuk-frontend/blob/ce8e22355e32af4d8f3c876779acd26128401842/package/govuk/components/textarea/macro-options.json) and then they're used to automatically generate the ['Nunjucks macro options' in the Design System](https://design-system.service.gov.uk/components/character-count/#options-character-count-example).

We've looked at trying to test that the params file is correct in the past – there's an issue and a really old spike at https://github.com/alphagov/govuk-frontend/issues/1139 – but didn't really get anywhere with it 😢

vanitabarrett:
@36degrees Added tests now - are you ok to re-review please?

src/govuk/components/character-count/character-count.test.js
```diff
@@ -143,6 +145,36 @@ describe('Character count', () => {
       })
     })
 
+    describe('when it has the spellcheck attribute', () => {
```
36degrees, line 148: Sorry, these tests should be in `src/govuk/components/character-counte/template.js` – components that have javascript have two test files:

- [component-name].test.js – integration tests that Puppeteer to test that the JavaScript behaves as expected
- template.test.js – unit tests (sort of) that test the HTML outputted by the template given different options

vanitabarrett: 🤦‍♀️ Not sure how I didn't spot that, thanks!

36degrees: It's not at all obvious! Is there anywhere you looked where you think it'd make sense to have documented this?

PR #1870, opened 2020-07-17
Reduce border width of form inputs in the error state

### LINKED ISSUE #1820: Focused inputs, textareas and file uploads may not have enough contrast when in the 'error' state
For the input and textarea components, we [rely on the input border getting thicker in order to meet WCAG 2.1 1.4.11: Non-text Contrast](https://design-system.service.gov.uk/get-started/focus-states/).

When the text input or textarea is in the error state, it already has a thicker border, and so the border only changes from red to black (in addition to the yellow outline). The contrast between the red and the black is only 2.75:1.

May-29-2020 16-09-11

36degrees:
Reduce border width of form inputs in error state.

We rely on the [input border getting thicker][1] in order to meet [WCAG 2.1 1.4.11: Non-text Contrast][2].

However, when the text input or textarea is in the error state, it already has a thicker border, and so the border only changes from red to black (alongside the addition of the yellow outline). The contrast between the red and the black is only 2.75:1, so the change in state from error to focused may not be perceivable by some users.

Reduce the error border width on inputs down to 2px, which is the same as the normal border width for a form input.

As the border width is now no thicker than the normal border width, remove the overrides that suppress the focus state's box shadow from errored inputs.

For components with multiple inputs, like the date input,this change does make it even more important to ensure that the error message clearly explains which field has the error. Although fields with an error continue to be differentiated from others by the thicker left border and the bold error text, the red border around the specific input can now only be seen as an enhancement. We are [updating the documentation in the Design System to stress the importance of being explicit about which field an error message relates to][3].

Remove the border from the file input in its error state, aligning it with radios / checkboxes, where the error state is indicated by the red left-hand border on the form group and the error message.

Finally, deprecate `$govuk-border-width-form-element-error` to be removed in v4.0.

Fixes #1820

[1]: https://design-system.service.gov.uk/get-started/focus-states/
[2]: https://www.w3.org/TR/WCAG21/#non-text-contrast
[3]: https://www.github.com/alphagov/govuk-design-system/pull/1302

| Before | After |
|---|---|
| localhost_3000_examples_error-summary  | localhost_3000_examples_error-summary (1) |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,11 +8,33 @@
 
 We darkened `govuk-colour("dark-grey")` to improve the readability of hint text. It now has a contrast ratio of 7:1 and helps hint text meet the WCAG 2.1 (AAA) accessibility standard.
 
+#### Updated error styling for text input, select, textarea and file upload components
+
+The error styling for the text input, select and textarea components no longer makes the border thicker.
+
+The error styling for the file upload component no longer includes a border around the file upload input.
+
+We're making these changes to ensure that the focus state is always perceivable.
+
+When an error message relates to multiple fields, ensure you are explicit about which field has the error. You must not rely on users being able to perceive which field has the error styling.
+
+([PR #1870](https://github.com/alphagov/govuk-frontend/pull/1870))
+
 #### Add spellcheck parameter to input and textarea components
 
 Optional parameter added to the input and textarea components to enable or disable the spellcheck attribute
 
 ([PR #1859](https://github.com/alphagov/govuk-frontend/pull/1859))
+
+### Deprecated features
+
+#### The `$govuk-border-width-form-element-error` setting
+
+The `$govuk-border-width-form-element-error` setting has been deprecated, and will be removed in GOV.UK Frontend v4.0.
+
+Update any references to `$govuk-border-width-form-element-error` in your application's styles to reference `$govuk-border-width-form-element` instead.
+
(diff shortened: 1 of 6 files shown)


EoinShaughnessy:
@vanitabarrett Was `$govuk-border-width-form-element-error` the thing that changed borders from red to black/an override/something else entirely?

vanitabarrett:
@EoinShaughnessy No, it set the width of the border rather than the colour. 

We used to make the border of an element thicker when it was in an error state, but we now don't (it keeps it's original thinner border size) - you can see the difference in the screenshots in the PR description. Ollie's explanation in the PR covers the reason why we've made this change. I think to really summarise it down to one or two lines, it's because we need the focus state to be made clear to users - a change from red border to black border isn't enough, but a change from red to black PLUS a change in border thickness is. Because the focus state already has a thick border, we need to reduce the size of the error border so there's a visible change in thickness when someone focuses on an input in error state. 

Not sure if that helps or is more confusing! Happy to jump on a call to explain.

EoinShaughnessy:
@vanitabarrett It helps, thanks! I drew on your explanation to update our [release notes](https://docs.google.com/document/d/1zXF8tRgbRJm8863-laNLB9ZcB4AbWdkNI-IGDtCth2U/edit?pli=1#heading=h.icv1a7bx25ly) about why we made these changes. Not sure what I've added is definitely what we want, though - probably a good one to chat about at our catchup tomorrow!

PR #1875, opened 2020-07-21
Switch linting to stylelint

kevindew:
Hello from GOV.UK 👋 , as forewarned this is a draft PR to propose a change in linter from the abandoned [sass-lint](https://github.com/sasstools/sass-lint) to [stylelint](https://stylelint.io/). This also abstracts the existing linting rules from this repo into a new package, [stylelint-config-gds](https://github.com/alphagov/stylelint-config-gds), which is intended to provide a stylelint configuration that can be recommended to all GDS projects via the [GDS way](https://gds-way.cloudapps.digital/).

## Motivation for GOV.UK to work on this 

GOV.UK is currently trying to migrate away from the deprecated Ruby SASS and instead use Sassc in projects. One of the blockers we have is that the linter we use, [scss-lint](https://github.com/sds/scss-lint), introduces Ruby SASS into projects and makes it cumbersome to migrate to Sassc. 

We therefore want to upgrade our linter. We don't think GOV.UK needs to go it's own route with a linter and prefer our SCSS to be consistent with the design system, thus ideally we'd like to adopt a similar configuration as the design system. However since sass-lint is now abandoned it doesn't make much sense to put work into applying that to ~40 applications. Thus we felt it'd be beneficial to try extract the design system linting rules into a configuration and then see if that could be adopted across GDS. Stylelint felt like the obvious choice given it's ubiquity in CSS/SCSS linting. 

The alternative approach - should this GDS attempt prove too complex, divisive or time consuming - is to have a stylelint configuration that is just for the GOV.UK programme and then to consider merging things at a later date.

## stylelint-config-gds

We've put together a repository, [stylelint-config-gds](https://github.com/alphagov/stylelint-config-gds), that provides SCSS or CSS rules that can be applied to projects. The basic goals of this project were:

1. Have as few rules as possible, instead leaning on community standards. In a similar vein to how [standard JS](https://standardjs.com/) resolves the need for lots of rule discussion
2. Be consistent with govuk-frontend styles to offer the least painful path for initial adoption forward
3. Can be applied to plain CSS as well as SCSS, since vanilla CSS has been regaining popularity for years
4. Be sufficiently agreeable that the frontend community of GDS can, or aspire to, adopt the rules

The approach taken involved building rules from [stylelint-config-standard](https://github.com/stylelint/stylelint-config-standard) and then tweaking any rules where govuk-frontend's existing ruleset disagreed. These rules can be seen (and are annotated) in [scss.js](https://github.com/alphagov/stylelint-config-gds/blob/master/scss.js) and [css.js](https://github.com/alphagov/stylelint-config-gds/blob/master/css.js). One thing to note when reviewing these rules is that the stylelint project is renaming it's whitelist/blacklist rules in their next release to neutral terminology (I could understand this producing reservations in adopting this tool).

The project is named stylelint-config-gds to represent that this is intended for the GDS organisation. It may well be that we decide later we want to make this broader and then give it a different name. This seems like a future step unless anyone has clear ideas now.

## Applying to govuk-frontend

The approach taken in this PR is to apply stylelint at the same time of scss-lint and then disable the occasional line/rule were there are minor disagreements between the linters. Although this is not a long term plan this did seem to offer a possible short term approach that two linters could run in parallel to aid adoption. I also broke down the applying of each rule to a single commit so that the effects of individual rules could be seen. It's likely easiest to review this PR commit by commit.

When establishing the rules for stylelint-config-gds it felt like the property ordering rule in govuk-frontend was somewhat more project specific than the other rules so I decided to employ that at a project level rather than in stylelint-config-gds. This was consistent to how [GOV.UK adopted the scss-lint](https://github.com/alphagov/govuk_publishing_components/pull/758) rules for one project but didn't use the property ordering rule. Do let me know if that raises eyebrows.

Anyway, here is a breakdown of rule differences from scss-lint:

[bem-depth](https://github.com/alphagov/govuk-frontend/blob/2f73bd6d639db10c61d3b0da6f79f08150a4e530/config/.sass-lint.yml#L22) - there isn't an equivalent core stylelint, there is a [potential plugin route](https://github.com/simonsmith/stylelint-selector-bem-pattern) but doesn't appear to be like-for-like. Could potentially also be achieved with specifying class name regex.

[border-zero](https://github.com/alphagov/govuk-frontend/blob/2f73bd6d639db10c61d3b0da6f79f08150a4e530/config/.sass-lint.yml#L26) - An equivalent rule doesn't exists in stylelint-core as it's [rather divisive topic](https://github.com/stylelint/stylelint/issues/948). We could apply this by a [plugin](https://github.com/timothyneiljohnson/stylelint-value-border-zero) but I'm inclined to not have a rule for it given the divisive aspects.

[class-name-format](https://github.com/alphagov/govuk-frontend/blob/2f73bd6d639db10c61d3b0da6f79f08150a4e530/config/.sass-lint.yml#L37) - there isn't the same BEM option available in stylelint core. Instead gone for a [relatively loose regexp](https://github.com/alphagov/stylelint-config-gds/blob/39e82c4197f26c303137787f1318599253b991b9/css.js#L78) that's roughly similar.

[no-color-literals](https://github.com/alphagov/govuk-frontend/blob/2f73bd6d639db10c61d3b0da6f79f08150a4e530/config/.sass-lint.yml#L153) - there isn't an equivalent rule in stylelint core that forces colors to always be hex. Could consider doing something with a regex if this is a concern.

[no-important](https://github.com/alphagov/govuk-frontend/blob/2f73bd6d639db10c61d3b0da6f79f08150a4e530/config/.sass-lint.yml#L190) - rather than allowing important values I left stylelint to raise an error on an important. This feels like the sort of rule developers should have to disable themselves for a single declaration.

## Next steps

The things we're looking for at this point are:

- Feedback on the rules in stylelint-config-gds. Are we missing any rules? or are some of the rules wrong? Are there any govuk-frontend specific preferences that can be deleted to fallback to stylelint-config-standard? 
- Identify if there is a road to adoption. Is this approach welcome? are there blockers? do any things need to happen before this can be adopted?
- Feedback on this specific PR. Are there are things that should be done differently to work towards adopting this?

Do feel free to raise issues on https://github.com/alphagov/stylelint-config-gds too for more in-depth conversations about aspects than can be covered in this PR. 

Thanks for your time :-) I'll look into arrange a meeting so we can chat more.


kevindew:
@vanitabarrett @hannalaakso I've updated this now with the inclusion of the declaration-no-important rule

vanitabarrett (approved):

kevindew:
Thanks for the approval @vanitabarrett. I've now created the initial public release of [stylelint-config-gds](https://github.com/alphagov/stylelint-config-gds) and rebased against this. Unfortunately I've learnt that there is a regression in the latest version of stylelint (which we depend on), which [they're working on fixing quickly](https://github.com/stylelint/stylelint/issues/4918) so I've rebased this with the failing branch. I'll ping you once this is sorted and this is ready to merge.

I also noticed that earlier on you were experimenting with seeing about keeping some rules (I recall disallowing hsl function) do you want to consider raising an issue on stylelint-config-gds about those things if they're still a concern?

vanitabarrett:
Thanks for the update @kevindew 🙂 We're just digging into https://github.com/alphagov/govuk-frontend/pull/1875/commits/448f78eb4b5df22ea2674ffd3266f66a09d5b96e to make sure we're happy with the media features being dropped in that commit, but all good otherwise. 

We decided against the other rules I was experimenting with (hsl and a stricter class regex for nesting), so nothing to raise against stylelint-config-gds there 👍

kevindew:
> Thanks for the update @kevindew 🙂 We're just digging into [448f78e](https://github.com/alphagov/govuk-frontend/commit/448f78eb4b5df22ea2674ffd3266f66a09d5b96e) to make sure we're happy with the media features being dropped in that commit, but all good otherwise.
> 
> We decided against the other rules I was experimenting with (hsl and a stricter class regex for nesting), so nothing to raise against stylelint-config-gds there 👍

That's great @vanitabarrett - did you come to a decision on the commit (now ba6641d)? 

I've updated to stylelint 13.7.1 so this is now a passing build. 

Would it be helpful if I cherry picked https://github.com/alphagov/govuk-frontend/pull/1902 onto this so it is just a single PR that goes in and applies both?

vanitabarrett:
Thanks @kevindew - all ok to leave https://github.com/alphagov/govuk-frontend/commit/ba6641d05db94a5812c55cea6449018ba81e4e02 as is 👍 

I'll go ahead and merge this, and then the PR to remove sass-lint. Thank you for all your hard work getting this completed! 🎉 👏

kevindew:
Yay, thanks @vanitabarrett 👍

PR #1879, opened 2020-07-22
Explicitly set outline-offset to remove 1px transparent border in chrome v84

rorywilsonros:
Chrome appears to have changed the default outline-offset to 1px, which then adds a transparent border to the skip link. This explicitly sets it to 0 to hide that border.


vanitabarrett:
Hi @rorywilsonros ! Thank you, really speedy work raising this. I'm happy to approve this, just a couple of tidy-up tweaks needed:

- can you squash the 2 commits into one
- would you mind adding a CHANGELOG entry (under "fixes") for this change? You can look at other CHANGELOG entries for examples. I'm also happy to add the entry on your behalf, if you'd prefer.

rorywilsonros:
Hi @vanitabarrett - thanks for the review and comments. I've squashed the commits, but would prefer you to add the changelog if that's ok? Best

Issue #1882, opened 2020-07-23
Test if inline script tag in page template has changed and alert of next steps


hannalaakso:
## What
We're going to document how teams should configure Content Security Policy (CSP) to work with GOV.UK Frontend page template: https://github.com/alphagov/govuk-frontend/issues/1811#issuecomment-663171834

As a result, any changes to the [inline script tag in Template](https://github.com/alphagov/govuk-frontend/blob/2339e8a1c13ad5eaa532d5b6e87a65de3405d78e/src/govuk/template.njk#L31) should be treated as a breaking change so we want to be alerted of any changes to it.

We need to add a test to check that a hash of the inline script tag matches 
```
sha256-+6WnXIl4mbFTCARd8N3COQmT3bJJmo32N8q8ZSQAIcU=
  ```

If the test fails, the test should output an error message saying that 
- a change to the script tag needs must be treated as a breaking change
- the relevant docs in Frontend docs need to be updated with the new hash of the script tag

## Why
We want to guard against accidental changes of the inline script tag because teams will be relying on the hash being correct for their CSP. 

We also want to be reminded that if there is an intended change, we should treat it as a breaking change and that we also need to update the relevant developer docs.

## Who needs to know about this
Developers

Issue #1811, opened 2020-05-19
Inline script in template.njk blocked by CSP


chrift:
We are looking to add a Content Security Policy to our application, but are currently unable to do so without specifying the `unsafe-inline` policy, due to the inline script tag in template.njk which adds the `js-enabled` class to the body. 

Could this instead be part of the `GOVUKFrontend.initAll` function to avoid having an inline script?

chrift:
Found [this issue](https://github.com/alphagov/govuk-frontend/issues/1657) and have added the sha hash to our CSP

36degrees:
Hi Chris,

Thanks for raising this. I'm going to re-open your issue and stick a triage label on it for our team to discuss, as I think we should provide better documentation around this so you don't have to go digging through past issues to find the solution!

hannalaakso:
To resolve this, we’re going to: 
- document the recommendation that teams allow the hash of the inline script tag in their CSP under JavaScript in GOV.UK Frontend developer docs 
- treat any changes to the hash as a breaking change 
- add a test to GOV.UK Frontend to check that the script tag hasn’t changed and if it does, what to do about it

This approach seems low risk since we’re not introducing any breaking changes, and we don’t expect the script tag to change much, if ever, so the hash should stay fairly static over time. 

We could in the future consider supporting nonces if we had evidence that it would be useful to teams.

We also considered publishing the hash in a file in the GOV.UK Frontend npm package but at the moment it isn’t clear how doing so would benefit teams.

hannalaakso:
Resolved by https://github.com/alphagov/govuk-frontend-docs/pull/70

PR #1884, opened 2020-07-28
Test if page template inline script matches hash we will publish in docs

hannalaakso:
We're going to [document how to configure Content Security Policy (CSP) to work with the inline 'js-enabled' script](https://github.com/alphagov/govuk-frontend-docs/issues/68) in Frontend page template.

This PR adds a test that
- uses Node.js Crypto to create a base64 encoded hash of the contents of the inline script tag in Frontend page template
- compares the generated hash it to the hash we're going to publish in developer docs
- adds a comment about the changes required if the script tag changes

We don't yet have the docs URL that needs to be updated if the hash changes, so I've used a placeholder. 

Fixes https://github.com/alphagov/govuk-frontend/issues/1882

### DIFF
--- a/src/govuk/template.test.js
+++ b/src/govuk/template.test.js
@@ -2,6 +2,7 @@
 
 const nunjucks = require('nunjucks')
 const configPaths = require('../../config/paths.json')
+const crypto = require('crypto')
 
 const { renderTemplate } = require('../../lib/jest-helpers')
 
@@ -150,6 +151,20 @@ describe('Template', () => {
       const $ = renderTemplate({}, { bodyEnd })
 
       expect($('body > div:last-of-type').text()).toEqual('bodyEnd')
+    })
+
+    describe('inline script that adds "js-enabled" class', () => {
+      it('should match the hash published in docs', () => {
+        const $ = renderTemplate()
+        const script = $('body > script').first().html()
+
+        // Create a base64 encoded hash of the contents of the script tag
+        const hash = crypto.createHash('sha256').update(script).digest('base64')
+
+        // A change to the inline script would be a breaking change, and it would also require
+        // updating the hash published in https://frontend.design-system.service.gov.uk/importing-css-assets-and-javascript/#if-your-javascript-isn-t-working-properly
+        expect('sha256-' + hash).toEqual('sha256-+6WnXIl4mbFTCARd8N3COQmT3bJJmo32N8q8ZSQAIcU=')
+      })
     })
 
     describe('skip link', () => {


src/govuk/template.test.js
```diff
@@ -152,6 +153,18 @@ describe('Template', () => {
...
 
+    describe('inline script that adds "js-enabled" class', () => {
+      it('should match the hash published in docs. A change to the script would be a breaking change, and it would also require updating the hash in https://frontend.design-system.service.gov.uk/importing-css-assets-and-javascript/#javascript', () => {
+        const $ = renderTemplate()
+        const script = $('body > script').first().html()
+
+        // Create a base64 encoded hash of the contents of the script tag
+        const hash = crypto.createHash('sha256').update(script).digest('base64')
+
+        expect('sha256-' + hash).toEqual('sha256-+6WnXIl4mbFTCARd8N3COQmT3bJJmo32N8q8ZSQAIcU=')
```
36degrees, line 164: Just because this is so long, I wonder if we should consider splitting out into a comment directly above the hash? If someone edited the test, they'd still see the comment.

Thoughts?

```suggestion
      it('should match the hash published in docs', () => {
        const $ = renderTemplate()
        const script = $('body > script').first().html()

        // Create a base64 encoded hash of the contents of the script tag
        const hash = crypto.createHash('sha256').update(script).digest('base64')

        // A change to the script would be a breaking change, and it would also
        // require updating the hash in https://frontend.design-system.service.gov.uk/importing-css-assets-and-javascript/#javascript
        expect('sha256-' + hash).toEqual('sha256-+6WnXIl4mbFTCARd8N3COQmT3bJJmo32N8q8ZSQAIcU=')
```

hannalaakso:
> I'm wondering if there's a way we can distribute the hash in the package we build so we don't have to manually update it in the docs each time it changes

@vanitabarrett This is a good idea, there's discussion about it [here](https://github.com/alphagov/govuk-frontend/issues/1811#issuecomment-663171834) but basically we're not going to publish the hash in the package for now as it's not clear at the moment how/if teams should use it from the package (but they _could_ come to rely on it being in the package if we publish it there). It might also be good to have that additional safeguard for ourselves of having to update it separately in the docs as it'll be a breaking change.

But we might find out in due course from teams that it would be useful to publish it in the package in which case we should definitely consider it. The current approach feels quite low risk and will allow us to add other things later if there's evidence that they would be useful (another example is supporting nonces which we could consider doing later).

src/govuk/template.test.js
hannalaakso: @36degrees Good idea, I'll change it 👍

hannalaakso:
Blocked until we can add the correct URL for the documentation: https://github.com/alphagov/govuk-frontend-docs/issues/68

Issue #1906, opened 2020-08-12
Visual focus glitch when clicking on checkbox and radio labels


edwardhorsford:
There's a minor visual glitch you get with focus when clicking on a checkbox or label. A bit like kerning, once you spot it, you can't un see it.

Issue: when clicking on the label for a checkbox or radio, focus on the checkbox itself flashes off and then on.

Example of clicking on label:
![flashing-focus](https://user-images.githubusercontent.com/2204224/90017120-f61b0500-dca2-11ea-9284-b0f696d0b9f6.gif)

Example of clicking on checkbox:
![non-flashing-focus](https://user-images.githubusercontent.com/2204224/90017196-0cc15c00-dca3-11ea-84cf-56d47bb9b978.gif)

I believe there are two main causes for this:
* Checkboxes don't normally get visual focus when clicked.
* Labels aren't focussable - when you click on the label, you unf-ocus the checkbox, and whatever is the next parent element that's focusable gets focused. When you mouse up, focus moves to the checkbox.

This [issue can also cause artefacts](https://github.com/DFE-Digital/apply-for-teacher-training/pull/2640) where checkboxes are placed in a focusable element - like in a modal or popover.

---

A prototyped a fix - pr incoming - but I don't necessarily expect it to merge - but it does help to illustrate one possible approach to fixing it. And the review app will help for testing.

After fix:
![flashing-focus-fixed](https://user-images.githubusercontent.com/2204224/90017459-78a3c480-dca3-11ea-8865-dd5689c7ef37.gif)

owenatgov:
After discussing this as a team we’ve decided we’re going to close this issue and the associated PR; and do not plan to  fix this because:

- All of the solutions that we can come up with create risk. Pure CSS solutions involve “tricking” the CSS into giving something that isn’t focused a focus state. Potential JS solutions involve mutating events like `mousedown` to prevent defaults. Both these examples and adjacent solutions do things that could impact us in unpredictable ways later.
- We don’t feel the issue is impactful enough to warrant said risk. Visual focus briefly moving off of the checkbox when the label is active is a purely aesthetic issue and doesn’t pose a serious user impact.
- The one instance where this could be an issue, when the checkboxes are in a container that allows for visual focus which is triggered as a result of this bug, meaning that when a checkbox label is active that the visual focus styles are triggered for the focusable container, can be solved at the service level.

[...]

PR #1907, opened 2020-08-12
Demo of possible fix for checkbox label focus glitch

edwardhorsford:
This PR demonstrates a possible fix to #1906.

I don't necessarily expect it to get merged - but the review app can help demonstrate how it works and possibly be used for testing.

I've tried it in Voiceover on mac and it seems to have no impact. 

As it's fiddling with focus it would want to be well tested - but you can't directly focus these items anyway so it mostly makes changes only when mouse-down on a label.

### DIFF
--- a/src/govuk/components/checkboxes/_index.scss
+++ b/src/govuk/components/checkboxes/_index.scss
@@ -30,6 +30,7 @@
   }
   
   .govuk-checkboxes__item:focus {
+    
     outline: none;
 
     & .govuk-checkboxes__label:before {
--- a/src/govuk/components/checkboxes/_index.scss
+++ b/src/govuk/components/checkboxes/_index.scss
@@ -31,7 +31,8 @@
   
   .govuk-checkboxes__item:focus {
     outline: none;
-    & .govuk-checkboxes__label:before {
+    
+    & .govuk-checkboxes__label:before { 
       border-width: 4px;
       box-shadow: 0 0 0 $govuk-focus-width $govuk-focus-colour;
     }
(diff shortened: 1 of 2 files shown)


edwardhorsford:
This fix is a variation of what I did to fix a [similar issue](https://github.com/DFE-Digital/apply-for-teacher-training/pull/2640) on my service - there using a parent element as a 'trap' for the focus.

I actually think it's better if the checkbox input retains visual focus, so I prefer this approach.

One other approach could be a pure js solution looking for mousedown on a label and moving focus.

owenatgov:
Hi @edwardhorsford. After 2+ years someone's getting back to you on this 😅 

Thanks for investigating this, the solution looks good. It has however prompted me to investigate. the issue and I think I've found a slightly neater solution that involves targeting the label with `:active`. By changing the focus state CSS that starts on on line 123 of the checkbox sass file so that it also targets `.govuk-checkboxes__label:active:before`, it applies the correct focus styling when the label is clicked ("active") even though focus technically isn't on the checkbox input anymore, as per your hypothesis.

I've done some very rudimentary testing and it seems to work well but I'm also interested in your opinion on this, if you can remember working on this. Depending on if you think this solution works we can do a couple of things to move it forward:

1. Tidy up this PR to reflect the `active` solution
2. Close this PR and open a new one with the `active` solution

Let me know what you think.

edwardhorsford:
@owenatgov thanks for looking at this.

I didn't immediately remember it specifically, but do recall fixing a bug related to this on our service. As my memory is patchy, it might be better if you just open a new pr with your solution. I'd be happy to give it a test though.

owenatgov:
I've made a PR with my proposal https://github.com/alphagov/govuk-frontend/pull/2827

As this PR is recorded in the new PR, I'm going to close this one. We can always reopen this one if we discover that it's a more suitable solution.

PR #1918, opened 2020-08-17
Add new brand colour for FCDO

deborahchua:
This new brand colour will then be used in https://github.com/alphagov/whitehall/commit/81898b4e6644d3eb847d4f811d02e0363c4a765b

Trello card - https://trello.com/c/s0NySL7Q/2098-add-new-brand-colour-for-fcdo

Co-authored-by: Rebecca Pearce <rebecca.pearce@digital.cabinet-office.gov.uk>


src/govuk/settings/_colours-organisations.scss
```diff
@@ -70,6 +70,10 @@ $govuk-colours-organisations: (
     colour: #00ad93,
     colour-websafe: #39836e
   ),
+  "foreign-commonwealth-development-office": (
+    colour: #012169,
+    colour-websafe: #003366
```
36degrees, line 75: We (confusingly) use 'websafe' here to mean 'meets the required contrast requirements to allow it to be used for text' (which would be at least 4.5:1). This is designed to be used when brand colours would not normally meet contrast, so that a darker alternative can be specified which does.

`#012169` has a contrast ratio of 14.75:1 against white, so already meets contrast requirements just fine.

You can omit the `colour-websafe` property entirely (like has been done for `"government-equalities-office"`) and the [`govuk-organisation-colour`](https://frontend.design-system.service.gov.uk/sass-api-reference/#govuk-organisation-colour) mixin will always return `#012169` for the FCDO whether `$websafe` is true or false.

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Add new brand colour for FCDO
```
36degrees, line 5: Can this be moved under the 'New features' heading, and include a H4 and a link to the pull request?

e.g.

```
#### Add new brand colour for The Foreign, Commonwealth and Development Office (FCDO)

This was added in [pull request #1918: Add new brand colour for FCDO](https://github.com/alphagov/govuk-frontend/pull/1918).
```

src/govuk/settings/_colours-organisations.scss
deborahchua: Ah I see, thanks @36degrees ! I've removed the `colour-websafe` property now.

CHANGELOG.md
deborahchua: Thanks @36degrees ! I've updated the changelog file now 👍

Issue #1573, opened 2019-09-13
Header Component - Custom HTML option for links


aliuk2012:
I am trying to use the header component but running into a couple of issues while adding a logout link. 

1. The current component does not support custom html only text and url for the navigation links. If I use html in `text` it still works but obviously it wraps `<a>` around the html. This raises issues with keyboard navigation (tabbing twice to go to the form element) and style focus.

eg.
```
{{
    govukHeader({
      'serviceName': 'Digital Marketplace',
      'navigation': { 'text': '<form method="post" action="google.com">
              <input type="submit" class="logout-button" value="Log out" />
              <input type="hidden" name="csrf_token"/>
            </form>',
      'href': url_for('external.render_login')
    })
  }}
```

![Sep-13-2019 13-42-10](https://user-images.githubusercontent.com/3441519/64864488-1d022e80-d62f-11e9-9b8a-3f877438adf2.gif)

To avoid this I could change it to a link but this changes the request from a `post` to `get` which is not considered good practice.

nickcolley:
We decided as a team that we'd support the simplest case which is to allow HTML alongside text

hannalaakso:
Some notes from team discussion on this:
- Seems like a legitimate use case
- We should consider not breaking existing functionality but adding a new option

hannalaakso:
I just looked into this. You **can** already include HTML for the navigation links but what you can't do is to include HTML and:
- not have it wrapped in `<a class="govuk-header__link" href="#">`
- have it render without passing through `href` 

neither of which you want if you want to render a form.

Even if we allowed rendering of HTML without `href` being set and removing the `<a>` tag, to style the form correctly `.govuk-header__link` would need to be on the input eg.

```html
<li class="govuk-header__navigation-item">
	<!-- Omitted <a class="govuk-header__link" href="#"> from here-->
	<form method="post" action="google.com">
	  <input type="submit" class="logout-button **govuk-header__link**" value="Log out" />
	  <input type="hidden" name="csrf_token"/>
	</form>
    <!-- Omitted </a> from here-->
</li>
```

## Potential solutions

1. Do nothing code wise and advise teams wanting to add a "Log out" button etc. that they should use the HTML version of header.
[...]

PR #1921, opened 2020-08-18
Make it possible to exclude link from header navigation item

hannalaakso:
This PR:
- makes rendering the link in the header navigation item conditional on `item.href` being set
- adds tests to check that `item.text` and `item.html` can render a navigation item without a link

We would expect this to be mainly useful for rendering a form for instance for adding a log out button. The team would need to manually style any enclosed form elements or buttons.

I considered not allowing this behaviour for `item.text` as it seems unlikely you'd want to omit the link when passing through simple text content. However I decided against in order to keep the API simpler, the fork would require documentation under `navigation.item.href` eg.
> Both `href` and `text` attributes for navigation items need to be provided to create an item. Using `html` attribute does not require `href` to create an item.

But happy to re-consider this.

Fixes https://github.com/alphagov/govuk-frontend/issues/1573

### DIFF
--- a/src/govuk/components/header/header.yaml
+++ b/src/govuk/components/header/header.yaml
@@ -35,7 +35,7 @@ params:
   - name: href
     type: string
     required: false
-    description: Url of the navigation item anchor. Both `href` and `text` attributes for navigation items need to be provided to create an item.
+    description: Url of the navigation item anchor.
   - name: active
     type: boolean
     required: false
@@ -232,3 +232,23 @@ examples:
         html: <em>Navigation item 2</em>
       - href: '#3'
         html: <em>Navigation item 3</em>
+
+- name: navigation item with text without link
+  data:
+    serviceName: Service Name
+    serviceUrl: '/components/header'
+    navigation:
+      - text: Navigation item 1
+        active: true
+      - text: Navigation item 2
+      - text: Navigation item 3
+
+- name: navigation item with html without link
+  data:
+    serviceName: Service Name
+    serviceUrl: '/components/header'
+    navigation:
+      - html: <em>Navigation item 1</em>
+        active: true
+      - html: <em>Navigation item 2</em>
+      - html: <em>Navigation item 3</em>
--- a/src/govuk/components/header/template.test.js
+++ b/src/govuk/components/header/template.test.js
@@ -139,6 +139,20 @@ describe('header', () => {
       expect($navigationLink.html()).toContain('<em>Nav item</em>')
     })
 
+    it('renders navigation item with text without a link', () => {
+      const $ = render('header', examples['navigation item with text without link'])
+
+      const $navigationItem = $('.govuk-header__navigation-item')
+      expect($navigationItem.html()).toContain('Navigation item 1')
+    })
+
+    it('renders navigation item with html without a link', () => {
+      const $ = render('header', examples['navigation item with html without link'])
+
+      const $navigationItem = $('.govuk-header__navigation-item')
+      expect($navigationItem.html()).toContain('<em>Navigation item 1</em>')
+    })
+
     it('renders navigation item anchor with attributes', () => {
       const $ = render('header', {
         navigation: [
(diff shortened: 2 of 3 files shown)


src/govuk/components/header/template.test.js
```diff
@@ -139,6 +139,20 @@ describe('header', () => {
       expect($navigationLink.html()).toContain('<em>Nav item</em>')
     })
 
+    it('renders navigation item with text without a link', () => {
+      const $ = render('header', examples['navigation item with text without link'])
+
+      const $navigationItem = $('.govuk-header__navigation-item')
+      expect($navigationItem.html()).toContain('Navigation item 1')
```
36degrees, line 146: I don't think this is testing quite what the test description says it's testing, because of the use of `.toContain()` – for example, even if the HTML looked like this:

```html
<li class="govuk-header__navigation-item">
    <a href="/super-sneaky-link">Navigation item 1</a>
</li>
```

then `$('.govuk-header__navigation-item').html()` would contain 'Navigation item 1' and the test would pass.

src/govuk/components/header/template.test.js
```diff
@@ -139,6 +139,20 @@ describe('header', () => {
       expect($navigationLink.html()).toContain('<em>Nav item</em>')
     })
 
+    it('renders navigation item with text without a link', () => {
```
36degrees, line 142: It might also be a good idea to test that when the navigation item is defined with `text` that entities are escaped like we expect them to be.

We do something similar here:

https://github.com/alphagov/govuk-frontend/blob/4a4d9342999cd559317724a6054b1b89da0f8d1e/src/govuk/components/label/template.test.js#L55-L62

src/govuk/components/header/template.test.js
hannalaakso: Good spot 👍

src/govuk/components/header/template.test.js
hannalaakso: Good point about there not being a test to check that navigation item text is escaped although it's slightly outside of what this PR was meant to. But I've added a commit to improve the existing test coverage.

CHANGELOG.md
```diff
@@ -4,6 +4,26 @@
...
+
+You can do this by setting `text` or `html` and not setting `href`.
+
+For example:
+
+```javascript
+{{ govukHeader({
+    navigation: [
+    {
+      html: "<em>Navigation item 1</em>"
```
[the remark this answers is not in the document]
hannalaakso: I've added a more specific example, let me know what you think.

src/govuk/components/header/header.yaml
```diff
@@ -232,3 +232,23 @@ examples:
...
       - href: '#3'
         html: <em>Navigation item 3</em>
+
+- name: navigation item with text without link
+  data:
+    serviceName: Service Name
+    serviceUrl: '/components/header'
+    navigation:
+      - text: Navigation item 1
+        active: true
```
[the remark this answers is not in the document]
hannalaakso: Good spot 👍

CHANGELOG.md
```diff
@@ -13,6 +13,28 @@ This was added in [pull request #1925: Generate fixtures.json files for componen
...
+```javascript
+{{ govukHeader({
+    navigation: [
+    {
+      html: "<form method='post' action='url.com'>
+              <input type='submit' class='app-logout-button-style' value='Log out' />
+            </form>"
+    }
+  ]
+}) }}
```
36degrees, line 33: Minor indentation tweak:

```suggestion
{{ govukHeader({
  navigation: [
    {
      html: "<form method='post' action='url.com'>
              <input type='submit' class='app-logout-button-style' value='Log out' />
            </form>"
    }
  ]
}) }}
```

CHANGELOG.md
```diff
@@ -13,6 +13,28 @@ This was added in [pull request #1925: Generate fixtures.json files for componen
...
+
+You can do this by setting `text` or `html` and not setting `href`.
+
+For example:
+
+```javascript
+{{ govukHeader({
+    navigation: [
+    {
+      html: "<form method='post' action='url.com'>
```
36degrees, line 28: Do we want to flip the quotes here so we're using double quotes for HTML attributes?

36degrees: Alternatively, we could provide an example using `{% set %}`?

36degrees (approved):

CHANGELOG.md
hannalaakso: I think we probably want to keep the example as simple as possible so instead of `{% set %}` I've flipped the quotes 👍

Issue #1830, opened 2020-06-09
Provide a way for maintainers of ports to check that they're using equivalent HTML markup


36degrees:
## What

Provide a way for maintainers of ports to ensure that, for given scenarios, their port generates markup equivalent to that of GOV.UK Frontend.

## Why

Whilst anyone can easily re-use the stylesheets and JavaScript from GOV.UK Frontend, most of our users have to maintain their own abstractions of the HTML markup used for a component.

It's important for that markup to be equivalent to the markup used in GOV.UK Frontend so that it renders the same and has the same accessibility. This is especially true when those abstractions are published as packages to be re-used by others.

## Further detail

Possible implementations might include:

- example ‘fixtures’ that provide the expected HTML output for a given set of inputs, optionally with a test suite that makes assertions using those fixtures and some sort of test harness implemented by the maintainer – a little like snapshot testing
- a test suite that makes assertions about the HTML for a given input, for example asserting that elements have specific attributes or relationships

Questions:

- How would fixtures relate to the component YAML files?
- Do these supplement or replace any of our existing tests? (Related: If using fixtures, do we generate them as part of the release process or as we make changes?)
- Do we release tests as a separate package or as part of the existing GOV.UK Frontend package?
- How thoroughly do we test components that are included by other components?
- What about less widely supported features, like {% call %}?
- How do we support ports that have deviated from the API (e.g. if they chose to implement sanitisation differently)
- How do we ensure coverage?
- What about scenarios where class names are changed (e.g. using local identifiers in React)

matthew-shaw:
This would be really useful for a Jinja port I've recently started: https://github.com/LandRegistry/govuk-frontend-jinja

The code in the repo above is generated by an automated script that creates Jinja equivalent macros for each of the Nunjucks ones. This is currently run within each frontend app itself at build time. I'd rather this was an installable package from PyPi to help manage version upgrades, remove some additional complexity from the app, and have a robust set of tests to prove compatibility. Ideally, I want to render the HTML output using a single set of inputs, for both the original Nunjucks and Jinja ported macros, then diff and fail tests if they're not identical.

As you can see, not got very far with this yet so feedback/contributions very welcome!

andymantell:
[...]

PR #1922, opened 2020-08-19
Add note for contributors on using hidden:true in yaml files

m-green:
Part of https://github.com/alphagov/govuk-frontend/issues/1830

As part of [adding docs on testing with HTML fixtures](https://github.com/alphagov/govuk-frontend-docs/pull/74), this adds a note in our contributing docs about using `hidden: true` in yaml files if you want your example to appear in the fixtures files but not the review app.

### DIFF
--- a/docs/contributing/testing.md
+++ b/docs/contributing/testing.md
@@ -60,6 +60,10 @@ You should write new tests if you’ve created a new component, or changed the w
 - changing or adding to the component's Nunjucks macro
 - creating or updating a Sass mixin or function
 
+Test files use examples from each component’s `.yaml` file, for example `src/govuk/components/button/button.yaml`. When you add or update tests, you can use the existing examples or add new ones.
+
+Use `hidden: true` in a new example if you do not want to include the example in the review app. The example will still appear in our [test fixtures](http://frontend.design-system.service.gov.uk/testing-your-html/).
+
 ### If you created a component
 
 Create the following files in the `src/govuk/components` folder:


docs/contributing/testing.md
```diff
@@ -37,6 +37,8 @@ You should add an example to the review app if the existing examples do not refl
 
 If you've created a new component, create a new `src/govuk/<COMPONENT>/<COMPONENT>.yaml` file instead, where `<COMPONENT>` is the name of the component you've created.
 
+Use `hidden: true` in an example if you want to include the example in our [test fixtures](http://frontend.design-system.service.gov.uk/testing-your-html/) but not the review app.
```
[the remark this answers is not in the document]
m-green: Good point! I've clarified this with a rejig in 51a6942.

docs/contributing/testing.md
```diff
@@ -37,7 +37,7 @@ You should add an example to the review app if the existing examples do not refl
 
 If you've created a new component, create a new `src/govuk/<COMPONENT>/<COMPONENT>.yaml` file instead, where `<COMPONENT>` is the name of the component you've created.
 
-Use `hidden: true` in an example if you want to include the example in our [test fixtures](http://frontend.design-system.service.gov.uk/testing-your-html/) but not the review app.
+Use `hidden: true` in your example if you do not want to include the example in the review app. The example will still appear in our [test fixtures](http://frontend.design-system.service.gov.uk/testing-your-html/).
```
hannalaakso, line 40: Would it be helpful to say that they would probably want the example to appear in the review app if there's a visual element to it?

m-green: Is that already covered in the first line of this section maybe? "You should add an example to the review app if the existing examples do not reflect the changes you've made."

hannalaakso: We should actually update that line too as not all examples appear in the review app with this new change. Maybe something along the lines of:
> You should add a new example if the existing examples do not reflect the changes you've made.

And then 

> If the example doesn't involve a visual change you can use `hidden: true` in your example to hide the example in the review app.

m-green: Ah I see what you mean. I'm thinking this through... is it likely that an external contributor would add an example to the yaml file that would need `hidden: true`? Or is it mainly us that would be doing that? 

I'm just wondering in retrospect whether to get into `hidden: true`, because including it might mean turning this section upside-down (because it's currently primarily focused on the review app).

hannalaakso: I think the rest of this section is still good in terms of the references to the review app. 

If we added the above suggestions and the contributor isn't sure what we mean by 'visual change' they could ask us.

docs/contributing/testing.md
```diff
@@ -37,6 +37,8 @@ You should add an example to the review app if the existing examples do not refl
 
 If you've created a new component, create a new `src/govuk/<COMPONENT>/<COMPONENT>.yaml` file instead, where `<COMPONENT>` is the name of the component you've created.
 
+Use `hidden: true` in your example if you do not want to include the example in the review app. The example will still appear in our [test fixtures](http://frontend.design-system.service.gov.uk/testing-your-html/).
```
hannalaakso, line 40: I wonder if this new paragraph would make more sense being positioned after
> 2. Add or update examples in the examples list.

as it's also about adding examples. 

The 'If you've created a new component' paragraph feels like it starts a new topic as it's about adding a new file

m-green: That makes senses, thanks - will tweak.

m-green:
Rejigged this in 99267dc following a chat with @36degrees - let me know if it's all ok!

docs/contributing/testing.md
m-green: Thanks Hanna - I've rejigged in 99267dc following your feedback and a chat with @36degrees. Hope it all makes sense now.

vanitabarrett (approved):
Looks good to me 👍 Ready to merge once @hannalaakso has approved too

hannalaakso (approved):
Nice, a great tweak to the original content @m-green 📝

Issue #1895, opened 2020-08-03
Generate test fixtures for components


vanitabarrett:
Part of https://github.com/alphagov/govuk-frontend/issues/1830

## Done when:

- A `fixture.json` file is generated for each component when building the package
- The `fixture.json` file is an object with the following properties: `name`, `examples`.

PR #1925, opened 2020-08-21
Generate fixtures.json files for components on build:package

vanitabarrett:
Resolves https://github.com/alphagov/govuk-frontend/issues/1895

## What
Adds steps to the `build:package` task to generate fixtures.json files for each component in the following format:

```
{
  "component": "button",
  "examples": [
      {
        "name": "default-example",
        "options": {
           "text": "Save and continue"
         },
         "html": "<button class=\"govuk-button\" data-module=\"govuk-button\">\n  Save and continue\n</button>"
      }
   ]
}
```
Fixtures files contain the component name  and all examples with the output html to allow maintainers to test their own HTML against ours

### DIFF
--- a/tasks/gulp/copy-to-destination.js
+++ b/tasks/gulp/copy-to-destination.js
@@ -1,5 +1,6 @@
 'use strict'
 
+const nunjucks = require('nunjucks')
 const gulp = require('gulp')
 const configPaths = require('../../config/paths.json')
 const postcss = require('gulp-postcss')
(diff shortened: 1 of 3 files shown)


hannalaakso (commented):
This looks really good to me 👍  Agree with @36degrees that it would be good to generate the two files in a similar manner.

vanitabarrett:
@hannalaakso @36degrees Thanks both - I think I've addressed those comments now. I've also added a second commit to create separate functions for the fixtures and macro-options gulp tasks - I *think* it makes it a bit easier to quickly scan the gulp task and see what's happening, but happy to leave that change out if people don't find it helpful!

tasks/gulp/copy-to-destination.js
```diff
@@ -62,3 +51,70 @@ gulp.task('copy-files', () => {
...
+function generateFixtures (file) {
+  const json = convertYamlToJson(file)
+  const componentName = path.dirname(file.path).split(path.sep).slice(-1).toString()
+  const componentTemplatePath = path.join(configPaths.components, componentName, 'template.njk')
+
+  if (json) {
+    const examplesJson = json.examples
+    const paramsJson = json.params
+
+    if (examplesJson && paramsJson) {
```
36degrees, line 64: Minor – does it matter if `json.params` is not truthy? We don't need it to create the fixtures file, and we don't use `paramsJson` anywhere.

tasks/gulp/copy-to-destination.js
```diff
@@ -62,3 +51,70 @@ gulp.task('copy-files', () => {
...
+      const fixtures = {
+        component: componentName,
+        examples: []
```
36degrees, line 67: To avoid muddling terminology should this be `fixtures` as well?

36degrees (approved):
Looks good to me, with a couple of minor comments 👍🏻 

I think it might be revisiting this task some point soon as it feels like it's becoming a bit of a 'monolith' task with too many responsibilities – the two JSON-generating functions could be separate dependent tasks?

That would also make it possible to update just the JSON files without having to run all of the other changes, which feels like it might be useful.

hannalaakso (approved):
Nice work 👍

vanitabarrett:
Needs rebasing against master once https://github.com/alphagov/govuk-frontend/pull/1924 has been merged

PR #1947, opened 2020-09-07
Add print styles for the panel component

### LINKED ISSUE #1946: Define print styles for panel component
As raised by a user on Slack, the lack of print styles means that the panel component prints with light grey text on white:

image

I believe this is the browser’s ‘best effort’ to make it printable – by default, browsers will remove background colours from things when printing them, and because the text would then be white-on-white it’s making the text grey.

_Originally posted by @36degrees in https://github.com/alphagov/govuk-frontend/issues/573#issuecomment-617212704_

36degrees:
Because the panel component is currently white text on a green background, it prints without the background with with light grey text (at least in Chrome).

Screenshot 2020-09-07 at 15 40 46

Add print styles to remove the background, set the text colour to black and add a border (like we do when colours are overridden).

Screenshot 2020-09-07 at 15 40 36

Fixes #1946


christopherthomasdesign (approved):
Looks completely sensible to me 👍

CharlotteDowns:
I agree I like it 👍

hannalaakso (approved):
Makes sense 👍

Issue #1871, opened 2020-07-17
Remove deprecated `$govuk-border-width-form-element-error` setting


36degrees:
## What

Remove the `$govuk-border-width-form-element-error` setting, and update any references in the code to use `$govuk-border-width-form-element`, or to just set the `border-color` as appropriate.

## Why

We deprecated `$govuk-border-width-form-element-error` in #1870 the border width of form elements in their error state (`$govuk-border-width-form-element-error`) is now the same as form elements in their normal state (`$govuk-border-width-form-element`) – both are 2px.

This isn't something we think users need to be able to configure, and we don't really want the complexity of trying to support different border widths for the error state (for example, the box shadow override that we removed in 06623f8 would need to be conditionally added if the user changed $govuk-border-width-form-element in order to maintain a consistent focus state).

## Who needs to know about this
Developers, Tech Writer

## Done when

- [x] `$govuk-border-width-form-element-error` has been removed

Issue #1958, opened 2020-09-17
Decide and document how to release a patch fix for GOV.UK Frontend if there are unreleased changes on main


hannalaakso:
## What
In the run up to Frontend v4.0.0 (a breaking release) we should consider how we manage the source code in case we need to do a hotfix release **when there are unreleased changes on the main branch that are not in a releasable state.**

## Why
If we needed to do a hotfix release prior to v4.0.0 we should know in advance the exact steps so that the process is quick and straightforward.

## Who needs to know about this
Developers, Eoin

## Done when
- [x] We've reviewed the different options and their pros and cons
- [x] We've met again and decided on an approach (see next steps)
- [x] We've mobbed on setting up a test project to run through the steps for the two approaches
- [x] We've drafted a document detailing how to publish a new version when there are unreleased changes on the main branch that are not in a releasable state
- [x] We've tested the process in the draft document and made any necessary changes/tweaks
- [x] Document has been reviewed by Eoin
- [ ] Document has been published
- [x] We've set up branch protection for the new support branches we would need to create

hannalaakso:
We should probably also consider how we manage backwards compatibility, for instance if **after** releasing 4.0.0 we need to release an urgent patch fix 4.0.1 and **3.9.2**.

hannalaakso:
Following a review, we've documented what we want to test in https://docs.google.com/document/d/1avJPHEJ4hENt5LEcMwxp90zl1WS72Tx6qdoN2JOhXCE/edit#heading=h.5s859hln1k0l

hannalaakso:
Following a conversation with Ollie and Vanita, the next steps are:

- Turn ‘Problem 1’ into a draft doc and get it reviewed by devs/Eion. Decide how this doc will work with the existing Publishing docs
- Investigate branch protection rules
- Investigate ‘latest’ tag question on ‘Problem 3’?
- Ask PM to help us define the scope of solving 3?

hannalaakso:
I've spoken to Eion about the draft doc. We've agreed that it would be good to avoid massive duplication between the 'Publishing' doc and this new one. I'm going to put together a very rough draft to see what the two publishing processes would like if compiled into one doc and Eion will then review it to see if it's viable to keep everything in that one doc. We should also bear in mind that there might yet be a third process (solution to Problem 3) added to that doc in the future too.

EoinShaughnessy:
@hannalaakso Nice going on the draft doc! It was useful to see the two processes beside each other.

[...]

PR #1963, opened 2020-09-18
Remove deprecated `$govuk-border-width-form-element-error`

vanitabarrett:
Fixes https://github.com/alphagov/govuk-frontend/issues/1871

Depends on https://github.com/alphagov/govuk-frontend/issues/1958 being decided first.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,17 @@
 
 ## Unreleased
 
+### Breaking changes
+You must make the following changes when you migrate to this release, or your service may break.
+
+#### Remove deprecated $govuk-border-width-form-element-error
+To make sure the border width of form elements in their error state is the same as form elements in their normal state, we've deprecated the `$govuk-border-width-form-element-error` setting.
+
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+### Breaking changes
+You must make the following changes when you migrate to this release, or your service may break.
+
+#### Deprecated: $govuk-border-width-form-element-error
```
hannalaakso, line 8: Only a small thing but the wording of the heading suggests that we're only now deprecating the variable even though we did it in https://github.com/alphagov/govuk-frontend/pull/1870. So maybe 'Removed' instead of 'Deprecated'?

hannalaakso (approved):

36degrees:
Minor wording thing: We already deprecated `$govuk-border-width-form-element-error` in #1870, so this PR is now removing the thing which was _previously_ deprecated ('marked for removal').

I think this mostly affects the release note, but it'd be good to update the PR title. I don't think it's enough of an issue to worry about the branch name?

PR #1975, opened 2020-10-02
Enable custom inputmode for date input component

foaly-nr1:
Some government websites allow the input of `X` characters in the day and month fields of a date when the values are unknown. In this case the `pattern` would need to be changed to `[0-9X]*` along with changing the `inputmode` to `text`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 ## Unreleased
 
+### New features
+
+#### Customise input mode in the date component
+
+You can now customise the input mode of each input in the date component to trigger a keyboard of your choice to appear on mobiles.
+
+This was added in [#1975 Enable custom inputmode for date input component](https://github.com/alphagov/govuk-frontend/pull/1975) – thanks to [@foaly-nr1](https://github.com/foaly-nr1) for contributing this.
+
 ## 3.9.1 (Fix release)
 
 ### Fixes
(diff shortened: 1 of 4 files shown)


foaly-nr1:
@36degrees what are your thoughts on this?

36degrees:
Just to note – Nick has left GDS so please don't at-mention him.

I think this change makes sense, especially given that you can already override pattern. Thank you for adding a test for as well – that's really helpful.

I would note that the pattern attribute is only used to force the numeric keypad on older iOS devices, which relies on the `pattern` being `[0-9]*`. Given that [you should have HTML5 validation turned off in your service](https://design-system.service.gov.uk/patterns/validation/#turn-off-html5-validation) it has no other purpose, so I'd recommend omitting the pattern attribute entirely or leaving it blank if you need to trigger the full alphanumeric keypad.

Would you mind adding an entry to [the changelog](https://github.com/alphagov/govuk-frontend/blob/master/CHANGELOG.md) under 'Unreleased' and as a new feature? I can do this for you, if you prefer.

foaly-nr1:
@36degrees done. feel free to change the wording if you'd like.

36degrees (approved):
Looks good to me – thanks for adding the changelog entry.

This just needs one other review from another member of the team and then we can get this merged.

vanitabarrett (approved):

foaly-nr1:
Thanks for the quick turnaround!

Issue #1935, opened 2020-09-01
Create notification banner component


36degrees:
## What

Create the [notification banner component](https://deploy-preview-1313--govuk-design-system-preview.netlify.app/components/notification-banner/) in GOV.UK Frontend.

## Why

Having reviewed the feedback from the working group, we now need to build the component in Frontend ready for release.

## Depends on

- #1933 
- #1934 

## Further detail

Some specific bits of feedback worth considering from the working group and [accessibility clinic findings](https://github.com/alphagov/govuk-design-system-backlog/issues/2#issuecomment-673356072):
- it’s recommended we remove `aria-labelledby` from the markup for the red and the green banners as they’re not landmarks (https://www.w3.org/TR/wai-aria-1.1/#landmark_roles) and therefore the label will never be read out (the same actually applies to our current error summary)
- we should explain the accessibility considerations – for example what the `role=` does – in comments in the code and/or tests.

## Done when…

- [x] Nunjucks template / macro created based on the API agreed in #1933 
- [x] Tests for Nunjucks template
- [x] Tech docs reviewed by tech writer
- [x] Add styling for notification banner
- [x] Component auto focuses when using role="alert"
- [x] Tests for component auto focus JS
- [x] Finalise notification banner macro options docs #1997 
- [x] Cross-browser tested
- [x] Create some realistic example pages to present at the accessibility clinic #1999
- [x] Take finalised version to accessibility clinic for testing #1999
- [x] Test notification banner with assistive technologies #2010
- [x] Address notification banner accessibility concerns #2011
- [x] Remove error type banner #2030

hannalaakso:
I'm thinking that one thing we should highlight at the accessibility clinic is the approach of making the `<p>` tags look like headings when we have just a single line of content in the banner. We do this as we want to draw attention to the message in the banner and semantically you should haven't a heading tag that doesn't act as a heading for some content below it. However it'd be good to confirm that this isn't a concern as we're making an element look like another element.

Issue #1934, opened 2020-09-01
Work out what to do with styled links in the notification banner component


36degrees:
## What

Links in the error summary are red, because we target them in the component CSS:

https://github.com/alphagov/govuk-frontend/blob/2b7a162b0f915483190ad56b5b15c543b8cd36c9/src/govuk/components/error-summary/_index.scss#L39-L53

We've flagged in #1732 that this seems to break the BEM principles applied elsewhere, where we style things explicitly as elements using classes, rather than using tags in the selector.

From talking to @christopherthomasdesign, when we build the notification banner, we want to make the links red for the error variant and green for the success variant, so we'll have a similar issue.

We should consider:

- that this is a valid use case to bend the BEM principles and using the same approach we use in the error summary
- creating error and success a link element and modifiers, scoped within the component (for example `govuk-notification-banner__link govuk-notification-banner__link--error`)
- creating error and success link modifiers (for example `govuk-link govuk-link--error`)

vanitabarrett:
## Option 1: style by element (`a`)
Follows the same pattern that we use for the error summary component. 

```
.govuk-notification-banner a { 
  // link styles
}
```

Pros: follows the existing approach we use for error summary; happens for the user magically - they don't need to remember to add a class.

Cons: not following BEM principles; potential duplication of CSS between error summary and notification banner error state.


## Option 2: New notification banner class
Add new classes within the notification banner, e.g: `govuk-notification-banner__link`.

```
# HTML
<a href="#" class="govuk-notification-banner__link“>Enter a postcode, like AA1 1AA</a>

# CSS
.govuk-notification-banner--success .govuk-notification-banner__link {
  // link styles
}
```

This could work in 2 ways:

a. Rely on the parent element having a class that reflects the state (error/success/info), e.g: `.govuk-notification-banner--success .govuk-notification-banner__link` (see above)

b. Have it's own modifiers, e.g: `.govuk-notification-banner__link--success`

Pros: scoped to just notification banner, so we can easily change this if needed without affecting anything else;

Cons: option b requires a bit more work from the user and potentially opens up the possibility of "mix and match" styles; 

## Option 3: New `govuk-link` modifiers
Add modifiers to the existing `govuk-link` class, e.g: `govuk-link--error`.

```
# HTML
<a href="#" class="govuk-link govuk-link—error govuk-notification-banner__link“>Enter a postcode, like AA1 1AA</a>

# CSS
@mixin govuk-link-style-error {
[...]

PR #1977, opened 2020-10-06
Add notifications banner template macro and tests

hannalaakso:
This PR adds
- a template macro for the notification banner component
- tests for the template logic
- examples of the component

Please don't review the stylesheets for now, they are from the demo version and need to be worked on separately; this could be done as part of https://github.com/alphagov/govuk-frontend/issues/1934.

Addresses parts of https://github.com/alphagov/govuk-frontend/issues/1935

### DIFF
--- a/src/govuk/components/notification-banner/README.md
+++ b/src/govuk/components/notification-banner/README.md
@@ -1,0 +1,15 @@
+# Notification banner
+
+## Installation
+
+See the [main README quick start guide](https://github.com/alphagov/govuk-frontend#quick-start) for how to install this component.
+
+## Guidance and Examples
+
+Find out when to use the notification banner component in your service in the [GOV.UK Design System](https://design-system.service.gov.uk/components/notification-banner).
+
+## Component options
+
+Use options to customise the appearance, content and behaviour of a component when using a macro, for example, changing the text.
+
+See [options table](https://design-system.service.gov.uk/components/notification-banner/#options-notification-banner-example) for details.
--- a/src/govuk/components/notification-banner/notification-banner.yaml
+++ b/src/govuk/components/notification-banner/notification-banner.yaml
@@ -1,0 +1,177 @@
+params:
+- name: text
+  type: string
+  required: true
+  description: If `html` is set, this is not required. Text to use within the notification banner. If `html` is provided, the `text` argument will be ignored.
+- name: html
+  type: string
+  required: true
+  description: If `text` is set, this is not required. HTML to use within the notification banner. If `html` is provided, the `text` argument will be ignored.
+- name: title
+  type: string
+  required: false
+  description: Title text to use within the notification banner. Defaults to 'Important' ('Success' for success type and 'Error' for error type). If `titleHtml` is supplied, the `title` argument will be ignored.
+- name: titleHtml
+  type: string
+  required: false
+  description: Title HTML to use within the notification banner. If `titleHtml` is provided, the `title` argument will be ignored.
+- name: titleHeadingLevel
+  type: string
+  required: false
+  description: Heading level, from 1 to 6. Default is `2`.
+- name: type
+  type: string
+  required: false
+  description: If `type` is set to `success` or `error`, the notification banner sets `role` to `alert` and `tabindex` to `-1`, and JavaScript moves the keyboard focus moves to the notification banner when the page loads. If `type` is not set, the notification banner defaults to setting `role` to `region` and using `aria-labelledby` to provide information for users of assistive technologies.
+- name: role
+  type: string
+  required: false
+  description: Overrides the value of the `role` attribute for the notification banner. Defaults to `region`. If `type` is set to `success` or `error`, defaults to `alert`.
+- name: tabindex
+  type: string/boolean
+  required: false
+  description: Overrides the value of the `tabindex` attribute for the notification banner. Not set by default.  If `type` is set to `success` or `error`, defaults to `-1`; the attribute can be removed by setting `tabindex` to `false`;
+- name: titleId
+  type: string
+  required: false
+  description: Overrides the value of the `id` attribute for the title. `id` used by the `aria-labelledby` attribute on the notification banner to provide information to users of assistive technologies. `id` defaults to `govuk-notification-banner-title` if `role` is set to `region`. If `type` is set to `success` or `error`, `id` is not rendered by default.
+- name: autoFocus
+  type: boolean
+  required: false
+  description: Moves keyboard focus to the notification banner when the page loads. Defaults to `false`. If `type` is set to `success` or `error`, defaults to `true`.
+- name: classes
+  type: string
+  required: false
+  description: Classes to add to the notification banner.
+- name: attributes
+  type: object
+  required: false
+  description: HTML attributes (for example data attributes) to add to the notification banner.
+
+examples:
+- name: default
+  data:
+    text: This publication was withdrawn on 7 March 2014.
+- name: with text as html
+  data:
+    html: |
+      <h3 class="govuk-heading-m">This publication was withdrawn on 7 March 2014</h3><p>Archived and replaced by the <a href="#">new planning guidance</a> launched 6 March 2014 on an external website</p>
+- name: with type as success
+  data:
+    type: success
+    text: Email sent to example@email.com
+- name: with type as error
+  data:
+    type: error
+    text: There was a problem uploading your file. Please try again.
+- name: with a list
+  data:
+    html: |
+      <h3 class="govuk-heading-m">4 files uploaded</h3>
+      <ul class="govuk-list govuk-list--bullet">
+          <li><a href="#">government-strategy.pdf</a></li>
+          <li><a href="#">government-strategy-v2.pdf</a></li>
+          <li><a href="#">government-strategy-v3-FINAL.pdf</a></li>
+          <li><a href="#">government-strategy-v4-FINAL-v2.pdf</a></li>
+      </ul>
+
+# Hidden examples are not shown in the review app, but are used for tests and HTML fixtures
+
+- name: custom title
+  hidden: true
+  data:
+    title: Important information
+    text: This publication was withdrawn on 7 March 2014.
+- name: title as html
+  hidden: true
+  data:
+    titleHtml: <span>Important information</span>
+    text: This publication was withdrawn on 7 March 2014.
+- name: title html as text
+  hidden: true
+  data:
+    title: <span>Important information</span>
+    text: This publication was withdrawn on 7 March 2014.
+- name: custom title heading level
+  hidden: true
+  data:
+    titleHeadingLevel: 3
+    text: This publication was withdrawn on 7 March 2014.
+- name: custom title id
+  hidden: true
+  data:
+    titleId: my-id
+    text: This publication was withdrawn on 7 March 2014.
+- name: custom title id with type as success
+  hidden: true
+  data:
+    type: success
+    titleId: my-id
+    text: Email sent to example@email.com
+- name: custom title id with type as error
+  hidden: true
+  data:
+    type: error
+    titleId: my-id
+    text: There was a problem uploading your file. Please try again.
+
+- name: custom text
+  hidden: true
+  data:
+    text: This publication was withdrawn on 7 March 2014.
+- name: html as text
+  hidden: true
+  data:
+    text: <span>This publication was withdrawn on 7 March 2014.</span>
+
+- name:  custom role
+  hidden: true
+  data:
+    role: banner
+    text: This publication was withdrawn on 7 March 2014.
+- name:  role overridden to region
+  hidden: true
+  data:
+    type: success
+    role: region
+    text: Email sent to example@email.com-
+- name: custom tabindex
+  hidden: true
+  data:
+    tabindex: 0
+    text: This publication was withdrawn on 7 March 2014.
+- name: tabindex as false and type as success
+  hidden: true
+  data:
+    type: success
+    tabindex: false
+    text: Email sent to example@email.com
+- name: autoFocus as true
+  hidden: true
+  data:
+    autoFocus: true
+    text: This publication was withdrawn on 7 March 2014.
+- name: autoFocus as false and type as success
+  hidden: true
+  data:
+    type: success
+    autoFocus: false
+    text: Email sent to example@email.com
+
+- name: classes
+  hidden: true
+  data:
+    text: This publication was withdrawn on 7 March 2014.
+    classes: app-my-class
+- name: attributes
+  hidden: true
+  data:
+    text: This publication was withdrawn on 7 March 2014.
+    attributes:
+      my-attribute: value
+
+- name: with invalid type
+  hidden: true
+  data:
+    type: some-type
+    text: This publication was withdrawn on 7 March 2014.
--- a/src/govuk/components/notification-banner/template.njk
+++ b/src/govuk/components/notification-banner/template.njk
@@ -1,0 +1,65 @@
+{%- if params.type == "success" or params.type == "error" %}
+  {% set successOrError = true %}
+{% endif %}
+
+{%- if successOrError %}
+  {% set typeClass = "govuk-notification-banner--" + params.type %}
+{% endif %}
+
+{% if params.role %}
+  {% set role = params.role %}
+{% elif successOrError %}
+  {# If type is success or error, add `role="alert"` to prioritise the information in the notification banner to users of assistive technologies #}
+  {% set role = "alert" %}
+{% else %}
+  {# Otherwise add `role="region"` to make the notification banner a landmark to help users of assistive technologies to navigate to the banner #}
+  {% set role = "region" %}
+{% endif %}
+
+{# Check whether to add `data-auto-focus="true"` which focuses the notification banner on page load #}
+{% if params.autoFocus is defined %}
+  {% set autoFocus = params.autoFocus %}
+{% elif successOrError %}
+  {% set autoFocus = true %}
+{% endif %}
+
+{% if params.tabindex is defined %}
+  {% set tabindex = params.tabindex %}
+{# Make sure that success or error banner should be focused on page load #}
+{% elif successOrError and autoFocus %}
+  {# Make the notification banner focusable #}
+  {% set tabindex = "-1" %}
+{% endif %}
+
+{%- if params.titleHtml %}
+  {% set title = params.titleHtml | safe %}
+{%- elif params.title %}
+  {% set title = params.title %}
+{%- elif params.type == "success" %}
+  {% set title = "Success" %}
+{%- elif params.type == "error" %}
+  {% set title = "Error" %}
+{%- else %}
+  {% set title = "Important" %}
+{%- endif -%}
+
+<div class="govuk-notification-banner {{ typeClass }} {%- if params.classes %} {{ params.classes }}{% endif -%}" role="{{ role }}"
+  {%- if tabindex or tabindex === 0 %} tabindex="{{ tabindex }}" {% endif %}
+  {%- if (role == "region") %} aria-labelledby="{{ params.titleId | default('govuk-notification-banner-title') }}" {%- endif -%}
+  data-module="govuk-notification-banner"
+  {%- if autoFocus -%} data-auto-focus="true"{% endif %}
+  {%- for attribute, value in params.attributes %} {{attribute}}="{{value}}"{% endfor %}>
+  <div class="govuk-notification-banner__header">
+    <h{{ params.titleHeadingLevel | default(2) }} class="govuk-notification-banner__title"{% if (role == "region" or params.titleId) %} id="{{ params.titleId | default('govuk-notification-banner-title') }}" {%- endif %}>
+      {{ title }}
+    </h{{ params.titleHeadingLevel | default(2) }}>
+  </div>
+  <div class="govuk-notification-banner__content">
+    {%- if params.html -%}
+      {{ params.html | safe }}
+    {%- elif params.text -%}
+      {# Set default style for single line content #}
+      <p class="govuk-notification-banner__heading">{{ params.text }}</p>
+    {%- endif -%}
+  </div>
+</div>
--- a/src/govuk/components/notification-banner/template.test.js
+++ b/src/govuk/components/notification-banner/template.test.js
@@ -1,0 +1,348 @@
+/**
+ * @jest-environment jsdom
+ */
+/* eslint-env jest */
+
+const axe = require('../../../../lib/axe-helper')
+
+const { render, getExamples } = require('../../../../lib/jest-helpers')
+
+const examples = getExamples('notification-banner')
+
+describe('Notification-banner', () => {
+  describe('default example', () => {
+    it('passes accessibility tests', async () => {
+      const $ = render('notification-banner', examples.default)
+
+      const results = await axe($.html())
+      expect(results).toHaveNoViolations()
+    })
+
+    it('aria-labelledby attribute matches the title id', () => {
+      const $ = render('notification-banner', examples.default)
+      const ariaAttr = $('.govuk-notification-banner').attr('aria-labelledby')
+      const titleId = $('.govuk-notification-banner__title').attr('id')
+
+      expect(ariaAttr).toEqual(titleId)
+    })
+
+    it('has role=region attribute', () => {
+      const $ = render('notification-banner', examples.default)
+      const $component = $('.govuk-notification-banner')
+
+      expect($component.attr('role')).toEqual('region')
+    })
+
+    it('does not have tabindex set', () => {
+      const $ = render('notification-banner', examples.default)
+
+      const $component = $('.govuk-notification-banner')
+      expect($component.attr('tabindex')).toBeUndefined()
+    })
+
+    it('has data-module attribute to initialise JavaScript', () => {
+      const $ = render('notification-banner', examples.default)
+      const $component = $('.govuk-notification-banner')
+
+      expect($component.attr('data-module')).toEqual('govuk-notification-banner')
+    })
+
+    it('does not have data-auto-focus attribute to focus component on page load', () => {
+      const $ = render('notification-banner', examples.default)
+      const $component = $('.govuk-notification-banner')
+
+      expect($component.attr('data-auto-focus')).not.toEqual('govuk-auto-focus')
+    })
+
+    it('renders header container', () => {
+      const $ = render('notification-banner', examples.default)
+      const $header = $('.govuk-notification-banner__header')
+
+      expect($header.length).toBeTruthy()
+    })
+
+    it('renders default heading level', () => {
+      const $ = render('notification-banner', examples.default)
+      const $title = $('.govuk-notification-banner__title')
+
+      expect($title.get(0).tagName).toEqual('h2')
+    })
+
+    it('renders default title text', () => {
+      const $ = render('notification-banner', examples.default)
+      const $title = $('.govuk-notification-banner__title')
+
+      expect($title.html().trim()).toEqual('Important')
+    })
+
+    it('renders content', () => {
+      const $ = render('notification-banner', examples.default)
+      const $content = $('.govuk-notification-banner__content')
+
+      expect($content.html().trim()).toEqual('<p class="govuk-notification-banner__heading">This publication was withdrawn on 7 March 2014.</p>')
+    })
+  })
+
+  describe('custom options', () => {
+    it('renders custom title', () => {
+      const $ = render('notification-banner', examples['custom title'])
+      const $title = $('.govuk-notification-banner__title')
+
+      expect($title.html().trim()).toEqual('Important information')
+    })
+
+    it('renders custom content', () => {
+      const $ = render('notification-banner', examples['custom text'])
+      const $content = $('.govuk-notification-banner__content')
+
+      expect($content.html().trim()).toEqual('<p class="govuk-notification-banner__heading">This publication was withdrawn on 7 March 2014.</p>')
+    })
+
+    it('renders custom heading level', () => {
+      const $ = render('notification-banner', examples['custom title heading level'])
+      const $title = $('.govuk-notification-banner__title')
+
+      expect($title.get(0).tagName).toEqual('h3')
+    })
+
+    it('renders custom role', () => {
+      const $ = render('notification-banner', examples['custom role'])
+      const $component = $('.govuk-notification-banner')
+
+      expect($component.attr('role')).toEqual('banner')
+    })
+
+    it('renders aria-labelledby attribute matching the title id when role overridden to region', () => {
+      const $ = render('notification-banner', examples['role overridden to region'])
+      const ariaAttr = $('.govuk-notification-banner').attr('aria-labelledby')
+      const titleId = $('.govuk-notification-banner__title').attr('id')
+
+      expect(ariaAttr).toEqual(titleId)
+    })
+
+    it('renders custom tabindex', () => {
+      const $ = render('notification-banner', examples['custom tabindex'])
+      const $component = $('.govuk-notification-banner')
+
+      expect($component.attr('tabindex')).toEqual('0')
+    })
+
+    it('renders custom title id', () => {
+      const $ = render('notification-banner', examples['custom title id'])
+      const $title = $('.govuk-notification-banner__title')
+
+      expect($title.attr('id')).toEqual('my-id')
+    })
+
+    it('has an aria-labelledby attribute matching the title id', () => {
+      const $ = render('notification-banner', examples['custom title id'])
+      const ariaAttr = $('.govuk-notification-banner').attr('aria-labelledby')
+
+      expect(ariaAttr).toEqual('my-id')
+    })
+
+    it('renders data-auto-focus attribute to focus component on page load', () => {
+      const $ = render('notification-banner', examples['autoFocus as true'])
+
+      const $component = $('.govuk-notification-banner')
+      expect($component.attr('data-auto-focus')).toEqual('true')
+    })
+
+    it('removes tabindex attribute so component is not focusable', () => {
+      const $ = render('notification-banner', examples['tabindex as false and type as success'])
+
+      const $component = $('.govuk-notification-banner')
+      expect($component.attr('tabindex')).toBeUndefined()
+    })
+
+    it('removes data-auto-focus attribute so component is not focused on page load', () => {
+      const $ = render('notification-banner', examples['autoFocus as false and type as success'])
+
+      const $component = $('.govuk-notification-banner')
+      expect($component.attr('data-auto-focus')).toBeUndefined()
+    })
+
+    it('renders classes', () => {
+      const $ = render('notification-banner', examples.classes)
+
+      const $component = $('.govuk-notification-banner')
+      expect($component.hasClass('app-my-class')).toBeTruthy()
+    })
+
+    it('renders attributes', () => {
+      const $ = render('notification-banner', examples.attributes)
+
+      const $component = $('.govuk-notification-banner')
+      expect($component.attr('my-attribute')).toEqual('value')
+    })
+  })
+
+  describe('html', () => {
+    it('renders title as escaped html when passed as text', () => {
+      const $ = render('notification-banner', examples['title html as text'])
+      const $title = $('.govuk-notification-banner__title')
+
+      expect($title.html().trim()).toEqual('&lt;span&gt;Important information&lt;/span&gt;')
+    })
+
+    it('renders title as html', () => {
+      const $ = render('notification-banner', examples['title as html'])
+      const $title = $('.govuk-notification-banner__title')
+
+      expect($title.html().trim()).toEqual('<span>Important information</span>')
+    })
+
+    it(' as escaped html when passed as text', () => {
+      const $ = render('notification-banner', examples['html as text'])
+      const $content = $('.govuk-notification-banner__content')
+
+      expect($content.html().trim()).toEqual('<p class="govuk-notification-banner__heading">&lt;span&gt;This publication was withdrawn on 7 March 2014.&lt;/span&gt;</p>')
+    })
+
+    it('renders content as html', () => {
+      const $ = render('notification-banner', examples['with text as html'])
+      const $contentHtml = $('.govuk-notification-banner__content')
+
+      expect($contentHtml.html().trim()).toEqual('<h3 class="govuk-heading-m">This publication was withdrawn on 7 March 2014</h3><p>Archived and replaced by the <a href="#">new planning guidance</a> launched 6 March 2014 on an external website</p>')
+    })
+  })
+
+  describe('when success type is passed', () => {
+    it('renders with appropriate class', () => {
+      const $ = render('notification-banner', examples['with type as success'])
+
+      const $component = $('.govuk-notification-banner')
+      expect($component.hasClass('govuk-notification-banner--success')).toBeTruthy()
+    })
+
+    it('has role=alert attribute', () => {
+      const $ = render('notification-banner', examples['with type as success'])
+
+      const $component = $('.govuk-notification-banner')
+      expect($component.attr('role')).toEqual('alert')
+    })
+
+    it('does not render aria-labelledby', () => {
+      const $ = render('notification-banner', examples['with type as success'])
+      const $component = $('.govuk-notification-banner')
+
+      expect($component.attr('aria-labelledby')).toBeUndefined()
+    })
+
+    it('does not render a title id for aria-labelledby', () => {
+      const $ = render('notification-banner', examples['with type as success'])
+      const $component = $('.govuk-notification-banner')
+
+      expect($component.attr('id')).toBeUndefined()
+    })
+
+    it('renders custom title id', () => {
+      const $ = render('notification-banner', examples['custom title id with type as success'])
+      const $title = $('.govuk-notification-banner__title')
+
+      expect($title.attr('id')).toEqual('my-id')
+    })
+
+    it('has the correct tabindex attribute to be focused', () => {
+      const $ = render('notification-banner', examples['with type as success'])
+
+      const $component = $('.govuk-notification-banner')
+      expect($component.attr('tabindex')).toEqual('-1')
+    })
+
+    it('renders data-auto-focus attribute to focus component on page load', () => {
+      const $ = render('notification-banner', examples['with type as success'])
+
+      const $component = $('.govuk-notification-banner')
+      expect($component.attr('data-auto-focus')).toEqual('true')
+    })
+
+    it('renders default success title text', () => {
+      const $ = render('notification-banner', examples['with type as success'])
+      const $title = $('.govuk-notification-banner__title')
+
+      expect($title.html().trim()).toEqual('Success')
+    })
+  })
+
+  describe('when error type is passed', () => {
+    it('renders with appropriate class', () => {
+      const $ = render('notification-banner', examples['with type as error'])
+
+      const $component = $('.govuk-notification-banner')
+      expect($component.hasClass('govuk-notification-banner--error')).toBeTruthy()
+    })
+
+    it('has role=alert attribute', () => {
+      const $ = render('notification-banner', examples['with type as error'])
+
+      const $component = $('.govuk-notification-banner')
+      expect($component.attr('role')).toEqual('alert')
+    })
+
+    it('does not render aria-labbelledby', () => {
+      const $ = render('notification-banner', examples['with type as error'])
+      const $component = $('.govuk-notification-banner')
+
+      expect($component.attr('aria-labbelledby')).toBeUndefined()
+    })
+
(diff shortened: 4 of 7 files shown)


src/govuk/components/notification-banner/notification-banner.yaml
```diff
@@ -0,0 +1,134 @@
...
+
+- name:  custom role
+  hidden: true
+  data:
+    role: banner
+    content: You have 9 days to send a response.
+- name: custom tabindex
+  hidden: true
+  data:
+    tabindex: "0"
```
hannalaakso, line 104: I marked this as a string because otherwise it seems to get interpreted as a falsy value from yaml [when the template checks whether the value is set](https://github.com/alphagov/govuk-frontend/pull/1977/files#diff-8ba6864c62c415e3869d6d4bb972023bR9). We could add a Nunjucks filter that does the same as 
```
if (value or value === 0) 
```

Generally speaking, 0 not being interpreted as a truthy value for things like attributes etc. is maybe something that we should watch out for with our other yaml examples as well 🤔

src/govuk/components/notification-banner/notification-banner.yaml
```diff
@@ -0,0 +1,134 @@
...
+  description: Heading level, from 1 to 6. Default is `2`.
+- name: content
+  type: string
+  required: true
+  description: Content for the notification banner. If `contentHtml` is supplied, this is ignored.
+- name: contentHtml
+  type: string
+  required: true
+  description: HTML content for the notification banner.
+- name: type
```
hannalaakso, line 22: Wondering whether "variant" might make more sense than "type"?

src/govuk/components/notification-banner/notification-banner.yaml
36degrees: I think this would mean that users would also have to ensure they passed the `tabindex` parameter as a string?

We can't add filters as we'd then rely on those filters being loaded in our user's Nunjucks environments, but I think something like this might do what we want:

```jinja
{% if 'tabindex' in params %}
  {% set tabindex = params.tabindex %}
{% elif params.type == "success" or params.type == "error" %}
  {% set tabindex = "-1" %}
{% else %}
  {% set tabindex = false %}
{% endif %}
```

Then:

```jinja
{% if tabindex !== false %} tabindex="{{ tabindex }}" {% endif %}
```

It'd also be good to have a test that shows that the user can pass `tabindex: false` to remove the tabindex entirely.

src/govuk/components/notification-banner/template.test.js
```diff
@@ -0,0 +1,276 @@
...
+const axe = require('../../../../lib/axe-helper')
+
+const { render, getExamples } = require('../../../../lib/jest-helpers')
+
+const examples = getExamples('notification-banner')
+
+describe('Notification-banner', () => {
+  describe('default example', () => {
+    it('passes accessibility tests', async () => {
+      const $ = render('notification-banner', examples.default)
```
36degrees, line 15: If we wanted to save rendering the same template multiple times (and remove some duplication), we could render the example once in a `beforeAll` function for the block, and then just make the assertions in each test.

(Also applies to some of the other describe blocks in this file)

src/govuk/components/notification-banner/template.njk
```diff
@@ -0,0 +1,40 @@
...
+
+{% if params.tabindex %}
+  {% set tabindex = params.tabindex %}
+{% elif params.type == "success" or params.type == "error" %}
+  {% set tabindex = "-1" %}
+{% endif %}
+
+{% set titleHeadingLevel = params.titleHeadingLevel if params.titleHeadingLevel else 2 %}
+
+<div class="govuk-notification-banner
```
36degrees, line 17: It'd be good to look at the whitespace control in this template, as it's including extra whitespace and line breaks in a few places:

```html
<div class="govuk-notification-banner"
  role="region"
 aria-labelledby="govuk-notification-banner-title" >
  <div class="govuk-notification-banner__header">
    <h2 class="govuk-notification-banner__title"  id="govuk-notification-banner-title"     >Important</h2>
  </div>
  <div class="govuk-notification-banner__content">
    <h3 class="govuk-heading-m">This publication was withdrawn on 7 March 2014</h3><p>Archived and replaced by the <a href="#">new planning guidance</a> launched 6 March 2014 on an external website</p>

  </div>
</div>
```

src/govuk/components/notification-banner/template.njk
```diff
@@ -0,0 +1,40 @@
...
+  {%- if params.type == "success" %} govuk-notification-banner--success{% endif %}
+  {%- if params.type == "error" %} govuk-notification-banner--error{% endif %}
+  {%- if params.classes %} {{ params.classes }}{% endif %}"
+  role="{{ role }}"
+  {% if tabindex %} tabindex="{{ tabindex }}" {% endif %}
+  {% if not(params.type == "success") and not(params.type == "error") %} aria-labelledby="govuk-notification-banner-title" {% endif %}
```
36degrees, line 23: Should this be tied to role being `region`, rather than the `type`? If the user overrides the role for some reason, the use of `aria-labelledby` should probably be adjusted accordingly.

src/govuk/components/notification-banner/template.njk
```diff
@@ -0,0 +1,40 @@
...
+  {% if ((params.type == "success" or params.type == "error") and not(params.initialFocus === false)) or params.initialFocus === true %} data-module="govuk-initial-focus"{% endif %}
+  {% for attribute, value in params.attributes %} {{attribute}}="{{value}}"{% endfor %}
+>
+  <div class="govuk-notification-banner__header">
+    <h{{ titleHeadingLevel }} class="govuk-notification-banner__title" {% if not(params.type == "success") and not(params.type == "error") %} id="govuk-notification-banner-title" {% endif %}
```
36degrees, line 28: I know we're planning to provide guidance about not having more than one banner on the same page, but at the minute there's no way to make the banner title IDs unique. Should there be?

src/govuk/components/notification-banner/template.njk
```diff
@@ -0,0 +1,40 @@
...
+    >
+      {%- if params.titleHtml %} {{ params.titleHtml | safe }}
+      {%- elif params.title %} {{ params.title }}
+      {%- elif params.type == "success" %}Success
+      {%- elif params.type == "error" %}Error
+      {%- else %}Important{%- endif -%}
+    </h{{ titleHeadingLevel }}>
+  </div>
+  <div class="govuk-notification-banner__content">
+    {{ params.contentHtml | safe if params.contentHtml else params.content }}
```
36degrees, line 38: I think to be consistent with elsewhere this should either be 

```suggestion
    {{ params.contentHtml | safe if params.contentHtml else params.contentText }}
```

or maybe as this is the 'main block' in this component, just:

```suggestion
    {{ params.html | safe if params.html else params.text }}
```

Should we make this component callable?

src/govuk/components/notification-banner/template.njk
```diff
@@ -0,0 +1,40 @@
...
+  role="{{ role }}"
+  {% if tabindex %} tabindex="{{ tabindex }}" {% endif %}
+  {% if not(params.type == "success") and not(params.type == "error") %} aria-labelledby="govuk-notification-banner-title" {% endif %}
+  {% if ((params.type == "success" or params.type == "error") and not(params.initialFocus === false)) or params.initialFocus === true %} data-module="govuk-initial-focus"{% endif %}
+  {% for attribute, value in params.attributes %} {{attribute}}="{{value}}"{% endfor %}
+>
+  <div class="govuk-notification-banner__header">
+    <h{{ titleHeadingLevel }} class="govuk-notification-banner__title" {% if not(params.type == "success") and not(params.type == "error") %} id="govuk-notification-banner-title" {% endif %}
+    >
+      {%- if params.titleHtml %} {{ params.titleHtml | safe }}
```
36degrees, line 30: I'd be tempted to set this sort of thing up at the top of the template, rather than inlining it. What do you think?

src/govuk/components/notification-banner/template.njk
```diff
@@ -0,0 +1,40 @@
...
+{% set titleHeadingLevel = params.titleHeadingLevel if params.titleHeadingLevel else 2 %}
+
+<div class="govuk-notification-banner
+  {%- if params.type == "success" %} govuk-notification-banner--success{% endif %}
+  {%- if params.type == "error" %} govuk-notification-banner--error{% endif %}
+  {%- if params.classes %} {{ params.classes }}{% endif %}"
+  role="{{ role }}"
+  {% if tabindex %} tabindex="{{ tabindex }}" {% endif %}
+  {% if not(params.type == "success") and not(params.type == "error") %} aria-labelledby="govuk-notification-banner-title" {% endif %}
+  {% if ((params.type == "success" or params.type == "error") and not(params.initialFocus === false)) or params.initialFocus === true %} data-module="govuk-initial-focus"{% endif %}
```
36degrees, line 24: This is quite hard to follow. Could we set it up at the top of the file? Maybe something like

```jinja
{% if 'initialFocus' in params %}
  {% set initialFocus = params.initialFocus %}
{% elif params.type == "success" or params.type == "error" %}
  {% set initialFocus = true %}
{% else %}
  {% set initialFocus = false %}
{% end %}
```

36degrees: I think the data-module should always be the component name, for example `data-module="govuk-notification-banner"`.

In case we want to add any other JS functionality in the future (like making banners dismissable) I'd suggest we always include the data-module attribute, and pass the initial focus as a separate data attribute – otherwise we're coupling the autofocusing behaviour with initialising the module.

```jinja
  data-module="govuk-initial-focus" {%- if initialFocus %} data-initial-focus="true"{% endif %}
```

src/govuk/components/notification-banner/notification-banner.yaml
```diff
@@ -0,0 +1,134 @@
...
+  type: string
+  required: false
+  description: Sets classes and attributes to create a type of notification banner. Accepts `success` or `alert`.
```
36degrees, line 25: I know setting this to 'notice' (or similar) wouldn't actually do anything, but it seems a little strange to only list the 2 'success' or 'alert' types here, when really there's 3 'types' and one of them just happens to be the default.

src/govuk/components/notification-banner/notification-banner.yaml
```diff
@@ -0,0 +1,134 @@
...
+  required: false
+  description: Custom role for the notification banner. Defaults to `region` (`alert` for success and error banners).
+- name: tabindex
+  type: string
+  required: false
+  description: Custom tabindex for the notification banner. Not set by default (defaults to `-1` for success and error banners).
+- name: initialFocus
+  type: string
+  required: false
+  description: Sets `data-module="govuk-initial-focus"` for the notification banner which focuses the banner on page load. Not set by default (on by default for success and error banners).
```
36degrees, line 37: Is it useful to mention the data attribute here? Should this just describe the outcome ("Focus the banner when the page loads")?

src/govuk/components/notification-banner/template.test.js
```diff
@@ -0,0 +1,276 @@
...
+
+      const results = await axe($.html())
+      expect(results).toHaveNoViolations()
+    })
+
+    it('aria-labelledby attribute matches the title id', () => {
```
36degrees, line 21: Being fussy, but it'd be good to fix this so it reads as a sentence.

```suggestion
    it('has an aria-labelledby attribute matching the title id', () => {
```

src/govuk/components/notification-banner/template.test.js
```diff
@@ -0,0 +1,276 @@
...
+    })
+
+    it('renders default title text', () => {
+      const $ = render('notification-banner', examples.default)
+      const $title = $('.govuk-notification-banner__title')
+
+      expect($title.html().trim()).toEqual('Important')
+    })
+
+    it('title id matches the aria-labelledby attribute on the component', () => {
```
36degrees, line 70: At the minute, you could just change the ID, fix the hardcoded ID in the test, and end up with an aria-labelledby that points to an invalid ID. 

Would it make sense to have one test that ties the two together? I think that'd also mean we could cover this relationship with one test rather than two.

```js
    it('aria-labelledby attribute matches the title id', () => {
      const $ = render('notification-banner', examples.default)
      const ariaAttr = $('.govuk-notification-banner').attr('aria-labelledby')
      const titleId = $('.govuk-notification-banner__title').attr('id')

      expect(ariaAttr).toEqual(titleId)
    })
```

We've done something sort of similar in the checkboxes template tests:

https://github.com/alphagov/govuk-frontend/blob/abe1672b68fe08a55fb3562a684b408a5d452bb0/src/govuk/components/checkboxes/template.test.js#L336-L347

src/govuk/components/notification-banner/README.md
```diff
@@ -0,0 +1,15 @@
+# Notification banner
+
+## Installation
+
+See the [main README quick start guide](https://github.com/alphagov/govuk-frontend#quick-start) for how to install this component.
+
+## Guidance and Examples
+
+Find out when to use the accordion component in your service in the [GOV.UK Design System](https://design-system.service.gov.uk/components/notification-banner).
```
36degrees, line 9:
```suggestion
Find out when to use the notification banner component in your service in the [GOV.UK Design System](https://design-system.service.gov.uk/components/notification-banner).
```

src/govuk/components/notification-banner/README.md
```diff
@@ -0,0 +1,15 @@
+# Notification banner
+
+## Installation
+
+See the [main README quick start guide](https://github.com/alphagov/govuk-frontend#quick-start) for how to install this component.
```
36degrees, line 5: We should probably update these to point to the Frontend docs at some point…

36degrees (commented):
Looking really good so far 👍🏻

src/govuk/components/notification-banner/notification-banner.yaml
vanitabarrett, line 25: Perhaps we could add something brief to the description about what this actually affects? E.g: the role and visual styling of the banner?

src/govuk/components/notification-banner/template.njk
```diff
@@ -0,0 +1,40 @@
...
+  {% set tabindex = params.tabindex %}
+{% elif params.type == "success" or params.type == "error" %}
+  {% set tabindex = "-1" %}
+{% endif %}
+
+{% set titleHeadingLevel = params.titleHeadingLevel if params.titleHeadingLevel else 2 %}
+
+<div class="govuk-notification-banner
+  {%- if params.type == "success" %} govuk-notification-banner--success{% endif %}
+  {%- if params.type == "error" %} govuk-notification-banner--error{% endif %}
```
vanitabarrett, line 19: Could we simplify this to something like:

```
{%- if params.type %} govuk-notification-banner--{{params.type}}{% endif %}
```

It does mean if someone passes in something other than `success` or `error` then it will generate an unused class, but not sure that's an issue?

vanitabarrett (changes requested):

src/govuk/components/notification-banner/notification-banner.yaml
hannalaakso: Thanks @36degrees this all makes sense. I'll add another test too.

src/govuk/components/notification-banner/template.njk
hannalaakso: Yes that's a good point. I'll also add something to the docs to explain this relationship between `role=region` and `aria-labelledby`.

src/govuk/components/notification-banner/template.njk
hannalaakso: Yes I did think about that. I think I'm going to make it customisable as the team might wish to link to the banner if it contains useful information so it'd be good if the id was meaningful in such case.

src/govuk/components/notification-banner/template.njk
hannalaakso: I had it as `text` initially but then went with `content` to help separate it more clearly from the title. But I guess since setting a title is optional it makes the text content clearly the main thing so `text` makes sense. I think @vanitabarrett had a similar comment as well. I'm going to change it.

src/govuk/components/notification-banner/notification-banner.yaml
hannalaakso: @36degrees I know what you mean. If this said 
> Accepts `success` or `alert`. Defaults to `notification`.

would that make more sense? (I could check with Chris if he's got thoughts on what the default `type` should be called.

src/govuk/components/notification-banner/notification-banner.yaml
hannalaakso: Yes it's on my list to make these explanations more helpful 👍

src/govuk/components/notification-banner/template.njk
vanitabarrett, line 38: Should we render`content.text`inside a `<p>` tag? I've just been looking at the styling, and having `content.text` inside it's own paragraph would enable us to style it without then automatically applying those styles to any `content.html` a user passes in (and therefore forcing them to override our styling)

src/govuk/components/notification-banner/template.njk
hannalaakso: @36degrees We've usually name spaced the attribute value (`govuk-notification-banner`). If we're setting the value to a boolean I wonder if the attribute should be name spaced instead eg.
```
data-govuk-initial-focus="true"
```
 This feels like something we should have possibly considered when we decided on the name spacing syntax 🤔 

On another note, I'm thinking that this attribute should only be rendered if it's true to avoid a redundant 
```
data-govuk-initial-focus="false"
```
on the HTML of the default component. I think this could confuse teams unnecessarily and there's probably not any benefit on having it copied over. Or what do you think?

src/govuk/components/notification-banner/notification-banner.yaml
hannalaakso: @36degrees I think I've now resolved this, please have a look.

src/govuk/components/notification-banner/template.test.js
hannalaakso: Chatted to @36degrees about this offline, it doesn't apply here so going to resolve comment.

src/govuk/components/notification-banner/template.njk
hannalaakso: @36degrees This should be resolved now.

src/govuk/components/notification-banner/template.njk
hannalaakso: Done!

src/govuk/components/notification-banner/template.njk
hannalaakso: Done!

src/govuk/components/notification-banner/template.njk
hannalaakso: Done!

src/govuk/components/notification-banner/notification-banner.yaml
hannalaakso: I've improved the documentation here but I'm also going to have a look at it with Eoin.

src/govuk/components/notification-banner/notification-banner.yaml
hannalaakso: Again I've tried to clarify the documentation here but will review it with Eoin next.

src/govuk/components/notification-banner/template.test.js
hannalaakso: Done!

src/govuk/components/notification-banner/notification-banner.yaml
```diff
@@ -0,0 +1,134 @@
...
+  required: false
+  description: Heading level, from 1 to 6. Default is `2`.
+- name: content
+  type: string
+  required: true
+  description: Content for the notification banner. If `contentHtml` is supplied, this is ignored.
+- name: contentHtml
+  type: string
+  required: true
+  description: HTML content for the notification banner.
```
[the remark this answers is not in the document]
hannalaakso: Good point, done!

src/govuk/components/notification-banner/notification-banner.yaml
hannalaakso: I've tried to improve this documentation and will review it with Eoin next.

src/govuk/components/notification-banner/template.njk
hannalaakso: I don't think we should add a redundant class but I've tried to simplify this logic and moved it to the top of the template so it's now not cluttering the markup, have a look.

src/govuk/components/notification-banner/template.njk
hannalaakso: @vanitabarrett I think wrapping the default text in `<p>` tags would make sense semantically. From a styling point of view though what if the user passes `<p>` tags as part of the `html` option? Would it be unexpected that those get styled in a certain way? Let me know what you think as you have a better idea of how the styles are going to work.

hannalaakso:
Thanks for the reviews @36degrees and @vanitabarrett 🙌 I think I've now resolved most of the comments, mainly apart from https://github.com/alphagov/govuk-frontend/pull/1977#discussion_r503412762 where it'd be good to decide how we name space boolean data attributes. I've left 2 TO DO's in the template code pending a decision.

I'm going to do a review of the documentation side of things with Eoin next.

src/govuk/components/notification-banner/template.njk
vanitabarrett: @hannalaakso I was thinking if we wrap `contentText` in a `<p>` tag that gets a specific class, e.g: `govuk-notification-banner__text`, then we can target that class specifically and leave any `contentHtml` unstyled. That would make it the responsibility of the user to style any html that they pass into the component.

src/govuk/components/notification-banner/notification-banner.yaml
```diff
@@ -0,0 +1,165 @@
...
+  required: false
+  description: Title HTML to use within the notification banner. If `titleHtml` is provided, the `title` argument will be ignored.
+- name: titleHeadingLevel
+  type: string
+  required: false
+  description: Heading level, from 1 to 6. Default is `2`.
+- name: type
+  type: string
+  required: false
+  description: If `type` is set to `success` or `alert`, the notification banner sets `role` to `alert` and `tabindex` to `-1`, and JavaScript moves the keyboard focus moves to the notification banner when the page loads. If `type` is not set, the notification banner defaults to setting `role` to `region` and using `aria-labelledby` to provide information for users of assistive technologies.
```
vanitabarrett, line 25: I think this should say:

```suggestion
  description: If `type` is set to `success` or `error`, the notification banner sets `role` to `alert` and `tabindex` to `-1`, and JavaScript moves the keyboard focus to the notification banner when the page loads. If `type` is not set, the notification banner defaults to setting `role` to `region` and using `aria-labelledby` to provide information for users of assistive technologies.
```

src/govuk/components/notification-banner/template.njk
36degrees: I don't think we need to namespace the attributes. The `data-module` pattern is used elsewhere so it make sense to namespace that to avoid conflicts, but we only do anything with `data-initial-focus` when `data-module="govuk-notification-banner"` and at that point the element is already 'in our scope' and so we shouldn't have to worry about namespacing every option individually.

Assuming `data-initial-focus` should default to `false` (which it probably should do?), agree the macro can omit data-govuk-initial-focus entirely if it's false. I think that's what would happen in the example I gave in my previous comment?

src/govuk/components/notification-banner/template.njk
```diff
@@ -0,0 +1,53 @@
...
+  {% set role = "region" %}
+{% endif %}
+
+{% if params.tabindex %}
+  {% set tabindex = params.tabindex %}
+{# TODO:  Link below logic to check whether js focus behaviour is enabled (when decision has been made in  https://github.com/alphagov/govuk-frontend/pull/1977#discussion_r500833933 #}
+{% elif successOrError and not(params.tabindex === false)%}
+  {# If type is success or error, add `tabindex="-1"` to make the notification banner focusable #}
+  {% set tabindex = "-1" %}
+{% endif %}
```
36degrees, line 25: I think this can be simplified slightly by checking for the presence of `tabindex` (with _any_ value) in the params object, rather than checking if `params.tabindex` is truthy?

```suggestion
{% if 'tabindex' in params %}
  {% set tabindex = params.tabindex %}
{# TODO:  Link below logic to check whether js focus behaviour is enabled (when decision has been made in  https://github.com/alphagov/govuk-frontend/pull/1977#discussion_r500833933 #}
{% elif successOrError %}
  {# If type is success or error, add `tabindex="-1"` to make the notification banner focusable #}
  {% set tabindex = "-1" %}
{% endif %}
```

src/govuk/components/notification-banner/notification-banner.yaml
36degrees: It'd be good to make the tabindex in the YAML file an integer again.

src/govuk/components/notification-banner/template.test.js
```diff
@@ -0,0 +1,327 @@
...
+
+      const $component = $('.govuk-notification-banner')
+      expect($component.attr('data-module')).toEqual('govuk-initial-focus')
+    })
+
+    it('removes tabindex attribute so component is not focusable', () => {
+      const $ = render('notification-banner', examples['tabindex as false and type as success'])
+
+      const $component = $('.govuk-notification-banner')
+      expect($component.attr('tabindex')).toBeFalsy()
```
36degrees, line 148: Does checking if this is undefined work? This would pass if tabindex="0", for example.

```suggestion
      expect($component.attr('tabindex')).toBeUndefined()
```

src/govuk/components/notification-banner/template.test.js
```diff
@@ -0,0 +1,327 @@
...
+      const $component = $('.govuk-notification-banner')
+
+      expect($component.attr('aria-labelledby')).toBeFalsy()
+    })
+
+    it('does not render a title id for aria-labelledby', () => {
+      const $ = render('notification-banner', examples['with type as success'])
+      const $component = $('.govuk-notification-banner')
+
+      expect($component.attr('id')).not.toEqual('govuk-notification-banner-title')
```
36degrees, line 229: Should this be checking for undefined, as we don't expect an id at all now?

src/govuk/components/notification-banner/template.njk
```diff
@@ -0,0 +1,53 @@
...
+{%- endif -%}
+
+<div class="govuk-notification-banner {{ typeClass }} {%- if params.classes %} {{ params.classes }}{% endif -%}" role="{{ role }}"
+  {%- if tabindex %} tabindex="{{ tabindex }}" {% endif %}
+  {%- if (role == "region") -%} aria-labelledby="{{ params.titleId | default('govuk-notification-banner-title') }}" {%- endif -%}
+  {# TODO: Sort out the below logic when decision has been made in discussion in https://github.com/alphagov/govuk-frontend/pull/1977#discussion_r500833933 is concluded #}
+  {%- if ((params.type == "success" or params.type == "error") and not(params.initialFocus === false)) or params.initialFocus === true -%} data-module="govuk-initial-focus"{% endif %}
+  {%- for attribute, value in params.attributes %} {{attribute}}="{{value}}"{% endfor %}>
+  <div class="govuk-notification-banner__header">
+    <h{{ params.titleHeadingLevel | default(2) }} class="govuk-notification-banner__title"{% if (role == "region" or params.id) %} id="{{ params.titleId | default('govuk-notification-banner-title') }}" {%- endif %}>
```
36degrees, line 46: There isn't a `params.id` listed in the YAML file – should this be `params.titleId`?

```suggestion
    <h{{ params.titleHeadingLevel | default(2) }} class="govuk-notification-banner__title"{% if (role == "region" or params.titleId) %} id="{{ params.titleId | default('govuk-notification-banner-title') }}" {%- endif %}>
```

src/govuk/components/notification-banner/template.njk
36degrees: Looking really good, although I think there's an errant `-` which is causing the space before `aria-labelledby` to get stripped:

```html
<div class="govuk-notification-banner " role="region"aria-labelledby="govuk-notification-banner-title">
  [..]
</div>
```

I think:
```
{%- if (role == "region") -%} aria-labelledby="{{ params.titleId | default('govuk-notification-banner-title') }}" {%- endif -%}
```

wants to be:

```
{%- if (role == "region") %} aria-labelledby="{{ params.titleId | default('govuk-notification-banner-title') }}" {%- endif -%}
```

src/govuk/components/notification-banner/notification-banner.yaml
36degrees: The new documentation is really good, but focuses only on the `role` and `tabindex` attributes and the focus behaviour. The `type` also affects the style of the banner, and the default title.

36degrees: I'll mark this as resolved though as I think that's separate.

src/govuk/components/notification-banner/notification-banner.yaml
hannalaakso: Done!

src/govuk/components/notification-banner/template.njk
hannalaakso: Done!

src/govuk/components/notification-banner/template.njk
hannalaakso: Okay that makes sense, thanks @36degrees. I think I've resolved this now, have a look.

src/govuk/components/notification-banner/notification-banner.yaml
hannalaakso: The tech docs are going to be reviewed by Eoin so will resolve this for now.

src/govuk/components/notification-banner/template.njk
hannalaakso: Thanks @vanitabarrett. Following our chat offline I've wrapped the `<p>` tag in a class. That class can then be used for styling the `<p>` tags in html that need to look like headings.

src/govuk/components/notification-banner/template.njk
hannalaakso: This should be resolved now @36degrees.

src/govuk/components/notification-banner/template.test.js
hannalaakso: Could you double check this one please? 🙏 I had the test failing with `.toBeFalsy()` if I set `tabindex=0` or `tabindex="0"`.

src/govuk/components/notification-banner/template.test.js
hannalaakso: Done!

src/govuk/components/notification-banner/template.njk
hannalaakso: Good catch. I've added a test which would have caught this.

src/govuk/components/notification-banner/notification-banner.yaml
```diff
@@ -0,0 +1,177 @@
...
+      <h3 class="govuk-heading-m">This publication was withdrawn on 7 March 2014</h3><p>Archived and replaced by the <a href="#">new planning guidance</a> launched 6 March 2014 on an external website</p>
+- name: with type as success
+  data:
+    type: success
+    text: Email sent to example@email.com
+- name: with type as error
+  data:
+    type: error
+    text: There was a problem uploading your file. Please try again.
+- name: with a list
```
vanitabarrett, line 67: Is it intentional that examples that don't explictly set `tabindex` still get a blank `tabindex` value in the HTML?

src/govuk/components/notification-banner/template.njk
vanitabarrett: @hannalaakso Let me know if I'm interpreting these comments incorrectly, but I think the suggestion was not namespacing the initial-focus attribute because it should already be scoped inside the `data-module="govuk-notification-banner"` - but I think that attribute is missing?

The relevant bit from Ollie's comment that I'm looking at:
> we only do anything with data-initial-focus when data-module="govuk-notification-banner"

src/govuk/components/notification-banner/template.test.js
vanitabarrett: @hannalaakso I get the same, the test fails if tabindex is set to 0, so I think it's ok 👍

src/govuk/components/notification-banner/notification-banner.yaml
hannalaakso: Good catch 👍

vanitabarrett (approved):

src/govuk/components/notification-banner/notification-banner.yaml
```diff
@@ -0,0 +1,177 @@
+params:
+- name: text
+  type: string
+  required: true
+  description: If `html` is set, this is not required. Text to use within the notification banner. If `html` is provided, the `text` argument will be ignored.
```
EoinShaughnessy, line 5: Does "text to use" suggest that users do something with the text once it's inside the notification banner? Suggested tweak: "The text to display within the notification banner."

Also, should this info display at the start of the description, i.e., after Required? This would inform the user what the text option is before they learn about the effect html has on it.

src/govuk/components/notification-banner/notification-banner.yaml
```diff
@@ -0,0 +1,177 @@
...
+- name: html
+  type: string
+  required: true
+  description: If `text` is set, this is not required. HTML to use within the notification banner. If `html` is provided, the `text` argument will be ignored.
```
EoinShaughnessy, line 9: Could personalise this and phrase it in the active voice: If you set 'text,' this...

EoinShaughnessy, line 9: Could personalise and phrase in active voice: If you provide 'html,' this...

src/govuk/components/notification-banner/notification-banner.yaml
EoinShaughnessy, line 5: Could personalise and phrase in active voice: If you set 'html,' this...

EoinShaughnessy, line 5: Could personalise and phrase in active voice: If you provide 'html,' this...

src/govuk/components/notification-banner/notification-banner.yaml
EoinShaughnessy, line 9: Do say 'parameter' instead of 'argument'?

src/govuk/components/notification-banner/notification-banner.yaml
```diff
@@ -0,0 +1,177 @@
+params:
+- name: text
+  type: string
+  required: true
+  description: If `html` is set, this is not required. Text to use within the notification banner. If `html` is provided, the `text` argument will be ignored.
+- name: html
```
EoinShaughnessy, line 6: Do the table options display in order of importance? If not, then is there a case for displaying them alphabetically? (I understand, though, if we wouldn't want to change the order because it would make this doc inconsistent with other docs.)

src/govuk/components/notification-banner/notification-banner.yaml
EoinShaughnessy, line 9: Should "HTML to use within the notification banner" display at the start of the description, i.e., after Required? This would inform the user what the option is before they learn anything else about it.

src/govuk/components/notification-banner/notification-banner.yaml
```diff
@@ -0,0 +1,177 @@
...
+  type: string
+  required: true
+  description: If `text` is set, this is not required. HTML to use within the notification banner. If `html` is provided, the `text` argument will be ignored.
+- name: title
```
EoinShaughnessy, line 10: So this option is for the title text in the banner, and the text option (above) is for non-title text in the banner? In the table, should this option display above the text option, to reflect their order in the banner?

src/govuk/components/notification-banner/notification-banner.yaml
```diff
@@ -0,0 +1,177 @@
...
+  type: string
+  required: false
+  description: Title text to use within the notification banner. Defaults to 'Important' ('Success' for success type and 'Error' for error type). If `titleHtml` is supplied, the `title` argument will be ignored.
```
EoinShaughnessy, line 13: Re first sentence — could just say "The title of the notification banner" (if that sounds ok).

What do we mean by “Defaults to ‘Important’” and ‘“‘Success’ for success type and ‘Error’ for error type’”? Also, is “‘Success’ for success type and ‘Error’ for error type’” self-evident (success for success, error for error)?

I think we can delete the parentheses as the style guide says not to use them.

Re last sentence — we could personalise this and make it active, for example, "If you provide 'titleHTML,' the 'title' argument..." (The options above this say "provide" instead of "supply," so I'd use the same wording here.)

src/govuk/components/notification-banner/notification-banner.yaml
```diff
@@ -0,0 +1,177 @@
...
+- name: titleHtml
+  type: string
+  required: false
+  description: Title HTML to use within the notification banner. If `titleHtml` is provided, the `title` argument will be ignored.
```
EoinShaughnessy, line 17: Could we personalise and replace the passive? So: “If you provide 'titleHtml',…”

src/govuk/components/notification-banner/notification-banner.yaml
```diff
@@ -0,0 +1,177 @@
...
+- name: titleHeadingLevel
+  type: string
+  required: false
+  description: Heading level, from 1 to 6. Default is `2`.
```
EoinShaughnessy, line 21: Mark's suggestion: "Sets the heading level of the title. The minimum is 1 and the maximum is 6. The default is 2."

How do we feel about beginning each description with a verb, like in Mark's suggestion? Would it be a good way to communicate the purpose of each option?

src/govuk/components/notification-banner/notification-banner.yaml
```diff
@@ -0,0 +1,177 @@
...
+- name: type
+  type: string
+  required: false
+  description: If `type` is set to `success` or `error`, the notification banner sets `role` to `alert` and `tabindex` to `-1`, and JavaScript moves the keyboard focus moves to the notification banner when the page loads. If `type` is not set, the notification banner defaults to setting `role` to `region` and using `aria-labelledby` to provide information for users of assistive technologies.
```
EoinShaughnessy, line 25: The description states what happens when you set type. Do we need to specify what type actually is, though?

Is this option to denote whether the notification banner is purely informational or else an alert box which the screen reader picks up on immediately?

Typo to remove: second "moves" in "Javascript moves the keyboard focus moves to the notification banner..."

Could personalise and make active the second last and last sentences. So, "If you set type to success..." and "If you do not set type..."

I'd also suggest changing "using aria-labelledby" to "uses aria-labelledby." This would be to parallel the active voice used in the previous clause ("...the notification banner defaults...").

src/govuk/components/notification-banner/notification-banner.yaml
```diff
@@ -0,0 +1,177 @@
...
+- name: role
+  type: string
+  required: false
+  description: Overrides the value of the `role` attribute for the notification banner. Defaults to `region`. If `type` is set to `success` or `error`, defaults to `alert`.
```
EoinShaughnessy, line 29: The description for the type option (above) mentions the role attribute, but doesn’t explain it. Would it help users if this role description displayed before they read about the type option?

“If type is set to success or error, defaults to alert.” It’s the role that defaults (and not the type), is that right? If so, should we explicitly call that out?

Replace passive in last sentence? So: “If you set type to…”

EoinShaughnessy, line 29: The description for the type option (above) mentions the role attribute, but doesn’t explain it. Would it help users if this role description displayed before they read about the type option?

“If type is set to success or error, defaults to alert.” It’s the role that defaults (and not the type), is that right? If so, should we explicitly call that out?

Replace passive in last sentence? So: “If you set type to…”

src/govuk/components/notification-banner/notification-banner.yaml
```diff
@@ -0,0 +1,177 @@
...
+- name: tabindex
+  type: string/boolean
+  required: false
+  description: Overrides the value of the `tabindex` attribute for the notification banner. Not set by default.  If `type` is set to `success` or `error`, defaults to `-1`; the attribute can be removed by setting `tabindex` to `false`;
```
EoinShaughnessy, line 33: The description for the type option above mentions the tabindex attribute, but doesn’t explain it. Would it help users if this tabindex description displayed before they read about the type option?

Are options set by default unless we state otherwise? If so, do we need to call this out somewhere?

Instead of "Not set by default," could we reword the first sentence to "Overrides the value that the browser sets for the tabindex attribute’ (or similar)?

“If type is set to success or error, defaults to alert.” Is it the tabindex or the type that defaults?

Suggested tweak for last sentence: “If you set type to success or error, tabindex defaults to -1. To remove the attribute, set tabindex to false.”

src/govuk/components/notification-banner/notification-banner.yaml
```diff
@@ -0,0 +1,177 @@
...
+- name: titleId
+  type: string
+  required: false
+  description: Overrides the value of the `id` attribute for the title. `id` used by the `aria-labelledby` attribute on the notification banner to provide information to users of assistive technologies. `id` defaults to `govuk-notification-banner-title` if `role` is set to `region`. If `type` is set to `success` or `error`, `id` is not rendered by default.
```
EoinShaughnessy, line 37: Could break up the second sentence. So: “...on the notification banner. This provides information…”

Should this display first? “id used by the aria-labelledby attribute on the notification banner.” It defines the option, which might be the first info a user wants to know.

Replace passive? So, “if you set role to region” instead of “if role is set to region” and “If you set type…" instead of "If type is set..."

src/govuk/components/notification-banner/notification-banner.yaml
```diff
@@ -0,0 +1,177 @@
...
+- name: initialFocus
+  type: boolean
+  required: false
+  description: Moves keyboard focus to the notification banner when the page loads. Defaults to `false`. If `type` is set to `success` or `error`, defaults to `true`.
```
EoinShaughnessy, line 41: Is it necessary to say "Defaults to false"? Maybe we could emphasise what setting it to true does. So, "If you set initialFocus to true, the keyboard focus moves to the notification banner when the page loads" (or similar).

Could make last sentence active: "If you set type,…"

src/govuk/components/notification-banner/notification-banner.yaml
```diff
@@ -0,0 +1,177 @@
...
+- name: classes
+  type: string
+  required: false
+  description: Classes to add to the notification banner.
```
EoinShaughnessy, line 45: I wonder could we personalise this (e.g. "Classes you want to add to the notification banner").

src/govuk/components/notification-banner/notification-banner.yaml
```diff
@@ -0,0 +1,177 @@
...
+- name: attributes
+  type: object
+  required: false
+  description: HTML attributes (for example data attributes) to add to the notification banner.
```
EoinShaughnessy, line 49: Could edit this to remove parentheses (as per style guide advice) and personalise it. So, "HTML attributes that you want to add to the notification banner, for example, data attributes."

hannalaakso:
Thanks @EoinShaughnessy for the super useful comments! 🙏 As discussed, we're going to incorporate your changes in a new pull request in order to unblock this PR as this is now blocking another PR from being merged.

src/govuk/components/notification-banner/notification-banner.yaml
hannalaakso: We're going to do a separate wider review of the tech docs here as this touches on most of the component docs.

src/govuk/components/notification-banner/notification-banner.yaml
hannalaakso: We have a loose order of importance (very loose in fact) and we seem to have settled on a bit of a hierarchy where text/html come first and the classes option is always last. We could definitely review that though.

src/govuk/components/notification-banner/notification-banner.yaml
hannalaakso: Good question. We've tended to have the text/html options first as those could be considered to be the main content. In this particular case, title text is also not required (a placeholder text is used if you don't supply one) whereas text/html are.

src/govuk/components/notification-banner/notification-banner.yaml
hannalaakso: > Do we say 'parameter' instead of 'argument'?

We actually say 'option' 😅  but this might not always be consistent throughout the rest of the tech docs.

src/govuk/components/notification-banner/notification-banner.yaml
hannalaakso: > What do we mean by “Defaults to ‘Important’” and ‘“‘Success’ for success type and ‘Error’ for error type’”?

I agree, this is definitely not clear, will be good to chat about it next week.

Issue #1965, opened 2020-09-22
Add required component options to yaml examples


vanitabarrett:
Part of https://github.com/alphagov/govuk-frontend/issues/1962

## Problem
For example: the date input component requires items to have a name or label, but this example has neither:
```
- name: with values
  hidden: true
  data:
    items:
      -
        id: day
      -
        id: month
      -
        id: year
        value: 2018
```

## Checklist

- [x] check the options we mark as required are correct (https://github.com/alphagov/govuk-frontend/pull/1979)
- [ ] add any required options to component examples

Issue #1930, opened 2020-08-26
Unclear if phase banner `tag` is required


vanitabarrett:
The phase banner [params](https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/components/phase-banner/phase-banner.yaml#L12) state that the `tag` is not required. However, if you leave out the `tag` param, you get an empty tag:

<img width="1016" alt="Screenshot 2020-08-26 at 16 54 42" src="https://user-images.githubusercontent.com/29889908/91326962-dadce900-e7bc-11ea-85b2-1aefaf8d3f1a.png">

Possible options:

- Make it required, and it renders an empty tag if you don’t pass one (similar to screenshot above)
- Only output a tag if tag.text or tag.html are passed, and leave it ‘optional’
- Only output a tag if tag.text or tag.html are passed, and say it’s required

terrysimpson99:
See also discussion at: https://github.com/alphagov/govuk-design-system-backlog/issues/57

MalcolmVonMoJ:
We have kept the phase banner, but without a tag.  

![image](https://user-images.githubusercontent.com/32877315/94676557-71984a80-0313-11eb-8c4c-72091ffd5b84.png)

We do, however, keep the tag for internal development, testing and staging environments.  

![image](https://user-images.githubusercontent.com/32877315/94676670-9987ae00-0313-11eb-89cf-b843d6d7207f.png)

nabeelamir-defra:
Hi would it be possible to "Only output a tag if tag.text or tag.html are passed, and leave it ‘optional’". Similar to the above screenshot we would like to keep the phase banner, but without the tag. At the moment we're using pure html to achieve this, but would like to use the component.

hannalaakso:
@nabeelamir-defra Thanks for your message. Making the phase banner component officially support your use case would be changing the purpose of the component. As the component relates to service assessments, status of projects etc., doing so would need to be a conversation involving many people from across government.

There is a a separate component for [feedback link for asking users for feedback about a service,](https://github.com/alphagov/govuk-design-system-backlog/issues/117) it might be worth reading through the comments on that issue and leaving a comment on the thread explaining your use case and how the feedback link component should meet them. Any user research you might have done in this area would be particularly valuable.

querkmachine:
As has been noted, the Phase Banner is used to convey details of the development and assessment phase of a service. We would expect it to include a Tag as part of this use, and do not intend for it to be used as a generic site-wide messaging component. 

[...]

PR #1979, opened 2020-10-07
Update required options for components

vanitabarrett:
Part of https://github.com/alphagov/govuk-frontend/issues/1965

## What

- Updates some component yaml files to mark certain options as `required: false`
- Add any missing `required` attributes to options 
- Attempt to clarify option descriptions where certain options depend on/override the other

Not included:

- Phase banner - tag is marked as `required: false` but failing to pass a tag means that a blank tag (with background colour) is rendered. See [issue here](https://github.com/alphagov/govuk-frontend/issues/1930).

### DIFF
--- a/src/govuk/components/footer/footer.yaml
+++ b/src/govuk/components/footer/footer.yaml
@@ -6,12 +6,15 @@ params:
   params:
   - name: visuallyHiddenTitle
     type: string
+    required: false
     description: Title for a meta item section, which defaults to Support links
   - name: html
     type: string
+    required: false
     description: HTML to add to the meta section of the footer, which will appear below any links specified using meta.items.
   - name: text
     type: string
+    required: false
     description: Text to add to the meta section of the footer, which will appear below any links specified using meta.items. If meta.html is specified, this option is ignored.
   - name: items
     type: array
--- a/src/govuk/components/footer/footer.yaml
+++ b/src/govuk/components/footer/footer.yaml
@@ -53,11 +53,11 @@ params:
     params:
     - name: text
       type: string
-      required: false
+      required: true
       description: List item text in the navigation section of the footer.
     - name: href
       type: string
-      required: false
+      required: true
       description: List item href attribute in the navigation section of the footer. Both `text` and `href` attributes need to be present to create a link.
     - name: attributes
       type: object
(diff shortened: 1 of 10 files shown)


src/govuk/components/footer/footer.yaml
```diff
@@ -6,12 +6,15 @@ params:
   params:
   - name: visuallyHiddenTitle
     type: string
+    required: false
     description: Title for a meta item section, which defaults to Support links
   - name: html
     type: string
+    required: true
```
36degrees, line 13: I don't think this is true for `meta.html` or `meta.text` – this is only used if you want to add extra HTML or text near the links. If you don't provide either of them we just omit that section entirely:

https://github.com/alphagov/govuk-frontend/blob/abe1672b68fe08a55fb3562a684b408a5d452bb0/src/govuk/components/footer/template.njk#L47-L51

src/govuk/components/input/input.yaml
```diff
@@ -13,7 +13,7 @@ params:
   description: Type of input control to render. Defaults to "text".
 - name: inputmode
   type: string
-  require: false
+  required: false
```
36degrees, line 16: I love it when you improve your test coverage and it catches stuff like this! 👏🏻 

Great job.

src/govuk/components/footer/footer.yaml
vanitabarrett: @36degrees Thanks, I think I confused it with the `meta.items.text`! Have updated `meta.text` and `meta.html` to be `required: false`

tasks/gulp/__tests__/after-build-package.test.js
```diff
@@ -107,17 +107,19 @@ describe('package/', () => {
       return readFile(filePath, 'utf8')
         .then((data) => {
           var parsedData = JSON.parse(data)
+
           // We expect the component JSON to contain "name", "type", "required", "description"
-          expect(parsedData).toEqual(
-            expect.arrayContaining([
+          expect(Array.isArray(parsedData)).toBeTruthy()
```
36degrees, line 112: I _think_ this could be…

```suggestion
          expect(parsedData).toBeInstanceOf(Array)
```

36degrees (approved):
This is great. Thanks for putting the rationale in the commit messages – it made it really easy to review.

PR #1983, opened 2020-10-12
Update examples for conditional reveals on radios and checkboxes

36degrees:
Correctly label the conditionally revealed email address field ‘email address’ rather than ‘mobile phone number’ in the conditional reveals example for checkboxes and radios.

### DIFF
--- a/src/govuk/components/checkboxes/checkboxes.yaml
+++ b/src/govuk/components/checkboxes/checkboxes.yaml
@@ -346,7 +346,7 @@ examples:
         text: Email
         conditional:
           html: |
-            <label class="govuk-label" for="context-email">Mobile phone number</label>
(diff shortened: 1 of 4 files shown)


vanitabarrett (approved):

vanitabarrett:
Just realised we have the same problem with radios. Probably my fault for copy+pasting - happy to fix myself separately if you want (example: https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/components/radios/radios.yaml#L309)

36degrees:
Ah sorry, I had spotted that and added the changes to this PR (I think?) but never updated the PR title or description!

vanitabarrett:
Looks good! 👍

Issue #1604, opened 2019-10-08
Axe issue: "All page content must be contained by landmarks"


colinrotherham:
Hello 👋 

We’ve noticed for a while that prototype kit and GOV.UK Frontend examples keep the “Back link” and “Phase banner” outside landmark regions.

i.e. They’re not in a `<header>`, `<footer>` or `<main>`.

This throws up moderate issues in axe:
https://dequeuniversity.com/rules/axe/3.3/region?application=AxeChrome

### Components not in `<main>` or `<header>`
This example shows showing both components outside regions:  
http://govuk-frontend-review.herokuapp.com/full-page-examples/upload-your-photo

![Components outside  main](https://user-images.githubusercontent.com/415517/66395735-18cbf580-e9d0-11e9-978b-cdd0ca4fa496.png)

### But the "Breadcrumbs" component _is_ in `<main>`
This example shows breadcrumbs inside a region:  
http://govuk-frontend-review.herokuapp.com/full-page-examples/renew-driving-licence

![Breadcrumbs component inside  main](https://user-images.githubusercontent.com/415517/66395830-4dd84800-e9d0-11e9-889c-8eb3ab62f0d9.png)

Here's what we see in the axe sidebar when follow the examples (with “Back link” and “Phase banner” outside landmark regions):

![Axe report](https://user-images.githubusercontent.com/415517/66396125-f686a780-e9d0-11e9-81b3-fb411141704c.png)

It would be great to see the research behind when to use a landmark or not.

Should this be logged under https://github.com/alphagov/govuk-frontend/issues/1280#issuecomment-509588851?

Thanks

nickcolley:
Just to clarify, the 'govuk-frontend-review' application is not intended to be guidance and contains intentionally legacy and therefore out of date patterns for test the 'compatibility mode'.

So that's the reason why one of the examples in there is wrong.

In the official guidance on the Design System website we intentionally do not have the phase banner, back links, or breadcrumbs in the 'main' element.

This is for two reasons I know of:
- the skip link is less useful if it does not skip the phase banner, back links and breadcrumbs
- too many navigation landmarks on GOV.UK have been reported to be an issue so we intentionally have removed some navigation landmarks in order to make it clear what the main navigation is.

I don't have the original research for the latter decision, but will ask around and see if I can dig it up.

Will leave this in triage for us to decide as a team what direction we want to take, thanks a lot for raising this I'm sure others will run into this too.

nickcolley:
[...]

PR #1984, opened 2020-10-14
Move phase banner inside header to resolve aXe landmark issue

robertparkinson:
Non breaking implementation so low impact on existing users.
See https://github.com/alphagov/govuk-frontend/issues/1604 for details

### DIFF
--- a/app/app.test.js
+++ b/app/app.test.js
@@ -157,6 +157,9 @@ describe(`http://localhost:${PORT}`, () => {
         const $ = cheerio.load(res.body)
         const $header = $('.govuk-header')
         const $serviceName = $header.find('.govuk-header__link--service-name')
+        const $phaseBanner = $header.find('.govuk-phase-banner')
+
+        expect($phaseBanner.length).toBe(1)
         expect($serviceName.html()).toContain('Nom du service')
         done(err)
       })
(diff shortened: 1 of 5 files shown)


hannalaakso:
Thanks a lot for your work on this @robertparkinson 👍 

As discussed in https://github.com/alphagov/govuk-frontend/issues/1604#issuecomment-539518514, some open questions remain around accessibility if the phase banner is moved inside the header. Have you been able to test this approach with users of assistive technologies and were there any findings around how the navigation elements and the skip link were perceived? (Our team also has some accessibility user research coming up so we might be able to test some variations of including the phase banner and the back links inside the header.)

robertparkinson:
Hi @hannalaakso I have not no as this change should have no impact on a users experience. I specifically chose to only move the phase banner as it doesn't run into the problems of the nav tags mentioned in that link. The tools used by most users from an accessibility point of view should behave the same as they will read out the feedback link in the phase banner even if it is in the header. 

I was hoping by making this small change that will remove one of the aXe errors it will help reduce the amount of time spent by testers and projects identifying this issue and either coding around it or getting it signed off as a known issue which is what we had to do.

vanitabarrett:
Hi @robertparkinson 👋  Thank you for this contribution. There are multiple ways that users of assistive tech may be navigating pages, and we can’t confidently say for certain that this change would have no impact on their experience without testing. We’re planning to include this as part of some upcoming user research our team is planning and work from there.

As a team I think we also want to consider whether this is the right approach technically, as this is probably something we'd be applying to other components, e.g: breadcrumbs, back link. Another potential approach would be decoupling the header component from the `<header>` element and making it part of the page template rather than fixed to a specific component (like we do for `<main>`). Obviously, this would be a much bigger change! We have some other header improvements planned, so probably need to consider how this change would fit in with those.

But as @hannalaakso mentioned, please let us know if you do any testing or come across any relevant research!

owenatgov:
We've decided to close this PR as it's been open for over 2 years following a comment from @vanitabarrett that this isn't something we would want to include in govuk-frontend in its current state. More details on the state of the linked issue: https://github.com/alphagov/govuk-frontend/issues/1604#issuecomment-1258201663

PR #1990, opened 2020-10-16
Add required fields to examples

vanitabarrett:
Closes https://github.com/alphagov/govuk-frontend/issues/1965

## What

- Update component YAML examples to include required attributes for all examples
- The only exception to the above is when we want to intentionally test how a component behaves with invalid data/in certain extremes, e.g: when no data is passed in at all, or when falsey values are provided.

## Why
In a previous PR we reviewed the `required` attributes on all our component options to double check that they're correct (https://github.com/alphagov/govuk-frontend/pull/1979). Following on from that work, we want to make sure the examples in the YAML files (used for tests and to generate `macro-options.json` and `fixtures.json`) properly reflect the required attributes.

### DIFF
--- a/src/govuk/components/date-input/date-input.yaml
+++ b/src/govuk/components/date-input/date-input.yaml
@@ -270,9 +270,6 @@ examples:
           data-example-year: year
 
 # Hidden examples are not shown in the review app, but are used for tests and HTML fixtures
-- name: no data
-  hidden: true
-  data: {}
 - name: classes
   hidden: true
   data:
--- a/src/govuk/components/date-input/date-input.yaml
+++ b/src/govuk/components/date-input/date-input.yaml
@@ -276,19 +276,23 @@ examples:
 - name: classes
   hidden: true
   data:
+    id: with-classes
     classes: app-date-input--custom-modifier
 - name: attributes
   hidden: true
   data:
+    id: with-attributes
     attributes:
       data-attribute: my data value
 - name: with empty items
   hidden: true
   data:
+    id: with-empty-items
     items: []
 - name: custom pattern
   hidden: true
   data:
+    id: with-custom-pattern
     items:
       -
         name: day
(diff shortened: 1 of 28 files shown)


src/govuk/components/date-input/date-input.yaml
```diff
@@ -276,26 +276,31 @@ examples:
 - name: classes
```
36degrees, line 276: I think the 'no data' example also needs an `id`.

src/govuk/components/warning-text/warning-text.yaml
```diff
@@ -34,6 +34,8 @@ examples:
   - name: attributes
```
36degrees, line 34: I think the 'text only' example in this file needs an `iconFallbackText` or possibly should just be removed, along with the test?

36degrees (commented):
The label component also seems to mark the `for` param as required, but most of the examples do not include it (and the template handles it being omitted) so possibly it shouldn't be marked as required?

src/govuk/components/date-input/date-input.yaml
vanitabarrett: Addressed https://github.com/alphagov/govuk-frontend/pull/1990/commits/d0d625ad8eaea1421887eb422f0e64f5b5a56726

Removed the 'no data' example, as adding an id makes it a duplicate of the default example.

src/govuk/components/warning-text/warning-text.yaml
vanitabarrett: Addressed in https://github.com/alphagov/govuk-frontend/pull/1990/commits/da7b4f6369886dfe4f8e31e6095789ef88e0238a

vanitabarrett:
@36degrees Thanks for the review - I think I've addressed them all now in the last few commits. I've marked the label for attribute as not required (https://github.com/alphagov/govuk-frontend/pull/1990/commits/49c530882a7379bd87315faf5332bb0f36910c26)

36degrees (approved):
I think this looks good to me. I haven't manually verified every single example - but until we've got a way to programatically validate all the examples I think we have to treat this as a 'best effort' exercise anyway.

PR #1993, opened 2020-10-19
Allow certain (header) component variables to be overridden.

codebymikey:
This issue also affects other components which might need certain variables to be overridden.

Unsure about the use of the `!global` flag as that assigns the variable globally if it's not already defined.

This quick regex `^\h+\$[\w-]+:` may be used to find locally defined component variables that can potentially be made to be overridden as needed.

### DIFF
--- a/src/govuk/components/header/_index.scss
+++ b/src/govuk/components/header/_index.scss
@@ -1,13 +1,13 @@
 @include govuk-exports("govuk/component/header") {
 
-  $govuk-header-background: govuk-colour("black");
-  $govuk-header-border-color: $govuk-brand-colour;
-  $govuk-header-border-width: govuk-spacing(2);
-  $govuk-header-text: govuk-colour("white");
-  $govuk-header-link: govuk-colour("white");
-  $govuk-header-link-hover: govuk-colour("white");
-  $govuk-header-link-active: #1d8feb;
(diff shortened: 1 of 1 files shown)


codebymikey:
Hi @36degrees,

I'm essentially creating a theme for potential government platforms and using the govuk-frontend framework as its foundation over a framework like Bootstrap 4 mainly for the look-and-feel and accessibility features.

However not all clients might want to use the typical colour schemes and it'd be nice if the framework was flexible enough to allow certain component styles to be overridden without having to replace/override the whole component.

colinrotherham:
Hi @codebymikey

Thanks again for opening this one, appreciate it's been a long time since we've replied

We've met to discuss whether we'd want the [Header component](https://design-system.service.gov.uk/components/header/) to have colour overrides but _really_ we see this component as the header for all GOV.UK services—always displayed as a "black bar with blue underline"

At the top of the [**Header** component](https://design-system.service.gov.uk/components/header/) (Design System page) we say:

>The GOV.UK header shows users that they are on GOV.UK and which service they are using.

So definitely do continue to use our component if you're running a GOV.UK service, but if you need customisable colour themes for your clients we'd rather you made your own.

Appreciate the contribution though, and it's good to document these things so I'll link to:

* https://github.com/alphagov/govuk-design-system/issues/1174

PR #1994, opened 2020-10-22
Auto-focus the notification banner, add tests

hannalaakso:
This PR:
- Auto-focuses the notification banner when the page loads if it has `role="alert"` and adds `tabindex="1"` so it can be fosused.
- Tests the auto-focus behaviour:
  - If a notification banner has `role=alert`, it should receive `tabindex="1"` so it can be focused with JavaScript and the keyboard focus should move to it on page load
  - If a notification banner has the `data-disable-auto-focus` attribute, or it doesn't have `role="alert"` it should not have a tabindex or be focused 

Tested in:
✅ Chrome 86 (Mac)
✅ Firefox 81 (Mac)
✅ Safari 14 (Mac)
✅ Galaxy S20 (Android 10)
✅ Galaxy Note 10 (Android 9)
✅ iPhone XS 12 
✅ IE11
✅ IE10 
✅ IE9
✅ IE8

Addresses part of https://github.com/alphagov/govuk-frontend/issues/1935

### DIFF
--- a/src/govuk/all.js
+++ b/src/govuk/all.js
@@ -5,6 +5,7 @@ import Details from './components/details/details'
 import CharacterCount from './components/character-count/character-count'
 import Checkboxes from './components/checkboxes/checkboxes'
 import ErrorSummary from './components/error-summary/error-summary'
+import NotificationBanner from './components/notification-banner/notification-banner'
 import Header from './components/header/header'
 import Radios from './components/radios/radios'
 import Tabs from './components/tabs/tabs'
@@ -49,6 +50,11 @@ function initAll (options) {
   // Find first header module to enhance.
   var $toggleButton = scope.querySelector('[data-module="govuk-header"]')
   new Header($toggleButton).init()
+
+  var $notificationBanners = scope.querySelectorAll('[data-module="govuk-notification-banner"]')
+  nodeListForEach($notificationBanners, function ($notificationBanner) {
+    new NotificationBanner($notificationBanner).init()
+  })
 
   var $radios = scope.querySelectorAll('[data-module="govuk-radios"]')
   nodeListForEach($radios, function ($radio) {
--- a/src/govuk/components/notification-banner/notification-banner.js
+++ b/src/govuk/components/notification-banner/notification-banner.js
@@ -1,0 +1,47 @@
+function NotificationBanner ($module) {
+  this.$module = $module
+}
+
+/**
+ * Initialise the component
+ */
+NotificationBanner.prototype.init = function () {
+  var $module = this.$module
+  // Check for module
+  if (!$module) {
+    return
+  }
+
+  this.setFocus()
+}
+
+/**
+ * Focus the element
+ *
+ * If `role="alert"` is set, focus the element to help some assistive technologies
+ * prioritise announcing it.
+ *
+ * You can turn off the auto-focus functionality by setting `data-disable-auto-focus="true"` in the
+ * component HTML. You might wish to do this based on user research findings, or to avoid a clash
+ * with another element which should be focused when the page loads.
+ */
+NotificationBanner.prototype.setFocus = function () {
+  var $module = this.$module
+
+  if ($module.getAttribute('data-disable-auto-focus') === 'true') {
+    return
+  }
+
+  if ($module.getAttribute('role') !== 'alert') {
+    return
+  }
+
+  // Set tabindex to -1 to make the element focusable with JavaScript.
+  if (!$module.getAttribute('tabindex')) {
+    $module.setAttribute('tabindex', '-1')
+  }
+
+  $module.focus()
+}
+
+export default NotificationBanner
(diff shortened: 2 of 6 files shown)


src/govuk/all.js
```diff
@@ -50,6 +51,11 @@ function initAll (options) {
   var $toggleButton = scope.querySelector('[data-module="govuk-header"]')
   new Header($toggleButton).init()
 
+  var $notificationBanners = scope.querySelectorAll('[data-module="govuk-notification-banner"]')
```
hannalaakso, line 54: We could just initialise the first instance of the component here. This might also remove the need to check in the JavaScript whether there's an already initialised notification banner on the page. However:
1. The initialisation is tied to the module, not the auto-focus behaviour so in that sense maybe we should keep the two separate and initialise all notification banners (although we haven't done this for the error summary but then that is less likely to have duplicates on the page)
2. We don't control how teams initialise components so they could initialise them individually and for that reason we should maybe have the check in JS. 
Thoughts welcome.

src/govuk/components/notification-banner/notification-banner.js
```diff
@@ -0,0 +1,82 @@
...
+
+  // Check if the attribute that determines whether the element should be auto-focused is present
+  if ($module.getAttribute('data-auto-focus') !== 'true') {
+    return
+  }
+
+  // Check if role="alert" is set
+  if ($module.getAttribute('role') !== 'alert') {
+    return
+  }
```
vanitabarrett, line 45: Could this be simplified to:

```js
if ($module.getAttribute('data-auto-focus') !== 'true' && $module.getAttribute('role') !== 'alert') {
    return
  }
```

src/govuk/components/notification-banner/notification-banner.js
```diff
@@ -0,0 +1,82 @@
...
+NotificationBanner.prototype.checkIfAnotherElementFocused = function () {
+  if (!document.activeElement) {
+    return
+  }
+
+  var errorSummaryPresent = document.querySelector('[data-module="govuk-error-summary"]')
+  var anotherNotificationFocused = document.activeElement.classList.contains('govuk-notification-banner')
+
+  return errorSummaryPresent || anotherNotificationFocused
+}
```
vanitabarrett, line 80: I'm in two minds about this - I'm not sure if we should be guarding against this in code or whether it's really something for our guidance instead?

hannalaakso: We can have different types of notification banners on the page and the top one is the most important so should be focused.

From the draft guidance:
> If you need to show notification banners of different types on the same page, place the more immediately relevant banner at the top:

src/govuk/all.js
hannalaakso: I thinks this has been resolved by https://github.com/alphagov/govuk-frontend/pull/1994#discussion_r512074290

src/govuk/components/notification-banner/notification-banner.test.js
```diff
@@ -0,0 +1,26 @@
+/* eslint-env jest */
+
+const configPaths = require('../../../../config/paths.json')
+const PORT = configPaths.ports.test
+
+const baseUrl = 'http://localhost:' + PORT
+
+describe('/components/notification-banner/with-type-as-success', () => {
```
36degrees, line 8: These are great – well-written tests 👍🏻 

Do you think it'd be worth also having tests for the other two code paths?

- when the notification banner has a `data-disable-auto-focus` attribute, it's not focused on page load
- when the notification banner does not have role="alert", it's not focused on page load

src/govuk/components/notification-banner/notification-banner.test.js
```diff
@@ -0,0 +1,26 @@
...
+  describe('when JavaScript is available', () => {
```
36degrees, line 9: We could probably remove this `describe` – I don't think we generally include this unless we also have tests for the non-JS fallback behaviour.

src/govuk/components/notification-banner/notification-banner.js
```diff
@@ -0,0 +1,47 @@
...
+  if ($module.getAttribute('role') !== 'alert') {
+    return
+  }
+
+  // Set tabindex to -1 to make the element focusable with JavaScript.
+  if (!$module.getAttribute('tabindex')) {
+    $module.setAttribute('tabindex', '-1')
+  }
+
+  this.$module.focus()
```
36degrees, line 44: Really minor, but to be consistent with the preceding lines:

```suggestion
  $module.focus()
```

36degrees (commented):
Looking really good – just a couple of minor comments 👍🏻

36degrees (approved):
👏🏻

src/govuk/components/notification-banner/notification-banner.test.js
hannalaakso: Good point, removed.

PR #2002, opened 2020-10-30
Adding guidance for use of sans-serif print styles

CharlotteDowns:
Added comment line [line 55] with guidance as to why sans-serif font is set as default and why we don't use govuk-font-family in print styles.

### DIFF
--- a/src/govuk/settings/_typography-font.scss
+++ b/src/govuk/settings/_typography-font.scss
@@ -52,6 +52,8 @@ $govuk-font-family-tabular: if(
 
 /// Font families to use for print media
 ///
+/// This option avoids font-related issues that occur with some printer drivers and operating systems.
+///
 /// @type List
 /// @access public
 
--- a/src/govuk/settings/_typography-font.scss
+++ b/src/govuk/settings/_typography-font.scss
@@ -52,7 +52,8 @@ $govuk-font-family-tabular: if(
 
 /// Font families to use for print media
 ///
-/// This option avoids font-related issues that occur with some printer drivers and operating systems.
+/// We recommend that you use system fonts when printing. This will avoid issues
+/// with some printer drivers and operating systems.
 ///
 /// @type List
 /// @access public


src/govuk/settings/_typography-font.scss
```diff
@@ -52,6 +52,8 @@ $govuk-font-family-tabular: if(
 
 /// Font families to use for print media
 ///
+/// This option avoids font-related issues that occur with some printer drivers and operating systems.
```
36degrees, line 55: I think it might help to be clear what we mean by 'this' and to explain a bit more about what issues we're avoiding – what about something like:

```suggestion
/// We recommend falling back to system fonts when printing to avoid issues with
/// garbled output seen with some printer drivers / operating systems.
```

EoinShaughnessy: Hello, not sure how to suggest a change like Ollie has done above. (Enlightenment welcome!)

I agree that it would be good to mention the issues.

Another suggestion: "We recommend that you use system fonts when you print. This will avoid issues with some printer drivers and operating systems."

36degrees: @EoinShaughnessy check out step 6 of https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/commenting-on-a-pull-request#adding-line-comments-to-a-pull-request 🙂

EoinShaughnessy, line 55:
```suggestion
/// We recommend that you use system fonts when you print. This will avoid issues with some printer drivers and operating systems.
```

36degrees:
The linter was complaining about trailing whitespace (extra spaces at the end of a line), which is why the tests were failing – I've updated the comment to remove the extra whitespace and wrap it rather than breaking it over 2 lines.

36degrees (approved):
LGTM.

Issue #1656, opened 2019-11-20
Limiting content width when using wider page layouts


edwardhorsford:
Raising this as it's something we've come across and perhaps other teams would too.

My service is about to move to a wider default page width - 1200px. We do this to accommodate our service's needs and take advantage of our users having more screen real estate.

In moving to 1200px width, it means lines of content inside a 2/3 column now become too long.

The design system site itself solves this by [setting max width on specific elements](https://github.com/alphagov/govuk-design-system/blob/master/src/stylesheets/main.scss#L69).

We've done similar, but set a fixed width of `722px` for all elements.

Should these max-widths on content lines be set by default?

## Suggestion:

Given the design system supports other widths, I wonder if one of the following might happen:

1. Review guidance on width in [grid system](https://design-system.service.gov.uk/styles/layout/#grid-system) - would be good to mention explicitly that services will need to limit content lines.
2. Include styles to limit specific items to no more than x width.

NB - the design system has 'Limiting width of content' - but this is not specific about textual content - which could possibly be covered separately.

## Detail

As well as the elements the design system site limits, we're also considering following elements:
* legend,
* label,
* govuk-hint,
* input,
* textarea,
* select,
* govuk-details__summary,
* govuk-details__text,
* govuk-inset-text,
* govuk-panel__body,
* govuk-warning-text__text

AdamChrimes:
In the NHS.UK frontend, we have something known as a "reading width" utility class, which does this. We did some research a while back and found the optimum amount of characters for reading was around 70-80 characters per line, so calculated the max-width with our typography sizing.

You can see the reading width guidance and example here: https://beta.nhs.uk/service-manual/styles-components-patterns/layout#utility-classes

The code can be found here: https://github.com/nhsuk/nhsuk-frontend/blob/master/packages/core/utilities/_reading-width.scss and https://github.com/nhsuk/nhsuk-frontend/blob/master/packages/core/tools/_mixins.scss

Don't know if you'd find this approach useful, rather than having apply a max width to each of the components. I suppose this gives a bit more control to the user.

nickcolley:
Thanks for sharing that @AdamChrimes, very useful.

A related feature request has come up before too: https://github.com/alphagov/govuk-frontend/issues/760

edwardhorsford:
@AdamChrimes thanks, good to see I'm not alone in wanting this.

[...]

PR #2012, opened 2020-11-10
Update issue templates

vanitabarrett:
Closes https://github.com/alphagov/design-system-team-internal/issues/349

We want to trial the use of issue templates in `govuk-frontend`. If the trial goes well, we will consider rolling them out to our other repos.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.md
+++ b/.github/ISSUE_TEMPLATE/bug-report.md
@@ -1,0 +1,33 @@
+---
+name: "🐛 Bug report"
+about: Report a bug or regression
+title: ''
+labels: "\U0001F41B bug, awaiting triage"
+assignees: ''
+
+---
+
+<!--
+    Please fill in as much of the template below as you’re able to. If you're unsure whether the issue already exists or how to fill in the template, open an issue anyway. Our team will help you to complete the rest.
+
+    Your issue might already exist. If so, add a comment to the existing issue instead of creating a new one. You can find existing issues here:
+    - an existing Github issue: https://github.com/alphagov/govuk-frontend/issues
+    - our known validation error/warning list: https://github.com/alphagov/govuk-frontend/issues/1280#issuecomment-509588851
+-->
+
+## Description of the issue
+<!-- A clear and concise summary of what the bug is. -->
+
+## Steps to reproduce the issue
+<!-- How can we reproduce this issue? If you think it will be helpful, please provide a small code snippet and/or screenshots. -->
+
+## Actual vs expected behaviour
+<!-- What is happening vs what would you expect to happen instead? -->
+
+## Environment (where applicable)
+<!-- Details of your operating system, browser and the version of GOVUK Frontend you’re using may help us to reproduce your issue. -->
+
+- Operating system:
+- Browser:
+- Browser version:
+- GOV.UK Frontend Version:
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,5 @@
+blank_issues_enabled: true
+contact_links:
+  - name: Get in touch another way
+    url: https://design-system.service.gov.uk/get-in-touch/
+    about: Find out how to get in touch via email or Slack
--- a/.github/ISSUE_TEMPLATE/feature-request.md
+++ b/.github/ISSUE_TEMPLATE/feature-request.md
@@ -1,0 +1,30 @@
+---
+name: "✨ Feature request"
+about: Suggest a new feature or idea
+title: ''
+labels: "feature request, awaiting triage"
+assignees: ''
+
+---
+
+<!--
+  Please fill in as much of the template below as you’re able to. If you're unsure whether the issue already exists or how to fill in the template, open an issue anyway. Our team will help you to complete the rest.
+
+  Your issue might already exist. If so, add a comment to the existing issue instead of creating a new one. You can find existing issues here:
+  - the community backlog: https://design-system.service.gov.uk/community/backlog/
+  - an existing Github issue: https://github.com/alphagov/govuk-frontend/issues
+
+  If you are proposing a new component or pattern, please follow the instructions here: https://design-system.service.gov.uk/community/propose-a-component-or-pattern/
+-->
+
+## Related component
+<!-- Does this feature relate to an existing component? -->
+
+## Context
+<!-- What are you trying to do? Is this something you think should behave differently, or something that you currently cannot do? Is this related to an existing issue/bug? -->
+
+## Alternatives
+<!-- Are you currently using a workaround / alternative solution instead? -->
+
+## Additional information (if applicable)
+<!-- What evidence do you have that this meets the needs of users? It’s useful for us to know of any user research/testing you’ve done with this feature. -->
--- a/.github/ISSUE_TEMPLATE/internal-story.md
+++ b/.github/ISSUE_TEMPLATE/internal-story.md
@@ -1,0 +1,24 @@
+---
+name: "Internal story template"
+about: For internal use only
+title: ''
+labels: "awaiting triage"
+assignees: ''
+
+---
+
+<!--
+
+  This is a template for any issues that aren’t bug reports or new feature requests. The headings in this section provide examples of the information you might want to include, but feel free to add/delete sections where appropriate.
+
+-->
+
+## What
+
+## Why
+
+## Who needs to know about this
+
+## Done when
+
+- [ ] Thing to do


.github/ISSUE_TEMPLATE/bug-report.md
```diff
@@ -0,0 +1,33 @@
...
+assignees: ''
+
+---
+
+<!--
+    Before raising this issue, make sure you have checked that the issue isn’t already captured in:
+    - an existing Github issue: https://github.com/alphagov/govuk-frontend/issues
+    - our known validation error/warning list: https://github.com/alphagov/govuk-frontend/issues/1280#issuecomment-509588851
+
+    Please fill in as much of the template below as you’re able to. If you're not sure whether the issue already exists or you're not sure how to fill in the issue template, feel free to open an issue anyway and our team will help you to complete the rest.
```
EoinShaughnessy, line 15:
```suggestion
    Please fill in as much of the template below as you’re able to. If you're unsure whether the issue already exists or how to fill in the template, open an issue anyway. Our team will help you to complete the rest.
```

.github/ISSUE_TEMPLATE/bug-report.md
```diff
@@ -0,0 +1,33 @@
...
+name: "🐛 Bug report"
+about: Report a bug or regression
+title: ''
+labels: "\U0001F41B bug, awaiting triage"
+assignees: ''
+
+---
+
+<!--
+    Before raising this issue, make sure you have checked that the issue isn’t already captured in:
```
EoinShaughnessy, line 11:
```suggestion
    Before you raise this issue, please check that the issue isn’t already captured in:
```

.github/ISSUE_TEMPLATE/feature-request.md
```diff
@@ -0,0 +1,32 @@
...
+about: Suggest a new feature or idea
+title: ''
+labels: "feature request, awaiting triage"
+assignees: ''
+
+---
+
+<!--
+
+  Before raising this feature request, please make sure you have checked that it isn’t already captured in:
```
EoinShaughnessy, line 12:
```suggestion
  Before you raise this feature request, please check that it isn’t already captured in:
```

.github/ISSUE_TEMPLATE/feature-request.md
```diff
@@ -0,0 +1,32 @@
...
+  - the community backlog: https://design-system.service.gov.uk/community/backlog/
+  - an existing Github issue: https://github.com/alphagov/govuk-frontend/issues
+
+  If you are proposing a new component or pattern, please follow the instructions here: https://design-system.service.gov.uk/community/propose-a-component-or-pattern/
+
+  Please fill in as much of the template below as you’re able to. If you're not sure whether the issue already exists or you're not sure how to fill in the issue template, feel free to open an issue anyway and our team will help you to complete the rest.
```
EoinShaughnessy, line 18:
```suggestion
  Please fill in as much of the template below as you’re able to. If you're unsure whether the issue already exists or how to fill in the template, open an issue anyway. Our team will help you to complete the rest.
```

EoinShaughnessy (commented):
This all made sense and was easy to read. Just left a few small comments. :)

.github/ISSUE_TEMPLATE/feature-request.md
```diff
@@ -0,0 +1,32 @@
...
+title: ''
+labels: "feature request, awaiting triage"
+assignees: ''
+
+---
+
+<!--
+
+  Before raising this feature request, please make sure you have checked that it isn’t already captured in:
+  - the community backlog: https://design-system.service.gov.uk/community/backlog/
```
36degrees, line 13: Sometimes we do ask people to promote issues from the backlog to the frontend repo, and if a feature has been 'lost' in the backlog raising it in the frontend repo seems like a good way to try and drive it forward. Is this definitely something we want to discourage?

.github/ISSUE_TEMPLATE/feature-request.md
```diff
@@ -0,0 +1,32 @@
...
+-->
+
+## Related component
+<!-- Does this feature relate to an existing component? -->
+
+## Context
+<!-- What are you trying to do? Is this something you think should behave differently, or something that you currently cannot do? Is this related to an existing issue/bug? -->
+
+## Alternatives
+<!-- Can you achieve the same result with an alternative solution? -->
```
36degrees, line 29: What are we trying to ask for here? Are we asking for the user to try and come up with other ways we could solve the same problem? Are we trying to understand if the user has a workaround they can make do with?

36degrees (commented):
These are looking great – super excited to see how well they work.

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,4 @@
+blank_issues_enabled: true
+contact_links:
+  - name: Support request
```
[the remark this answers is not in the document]
vanitabarrett: @36degrees Do you think this would be better if it was less specific about getting 'support'? We could do something like this to mirror what the design system page says:

```
name: Get in touch
url: https://design-system.service.gov.uk/get-in-touch/
about: If you’ve got a question, idea or suggestion get in touch with the Design System team.
```

.github/ISSUE_TEMPLATE/bug-report.md
```diff
@@ -0,0 +1,33 @@
...
+    Please fill in as much of the template below as you’re able to. If you're not sure whether the issue already exists or you're not sure how to fill in the issue template, feel free to open an issue anyway and our team will help you to complete the rest.
+-->
+
+## Description of the issue
+<!-- A clear and concise summary of what the bug is. -->
+
+## Steps to reproduce the issue
+<!-- How can we reproduce this issue? If you think it will be helpful, please provide a small code snippet and/or screenshots. -->
+
+## Expected behaviour
```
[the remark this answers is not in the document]
vanitabarrett: Changed to:

```
<!-- What is happening vs what would you expect to happen instead? -->
```

.github/ISSUE_TEMPLATE/feature-request.md
vanitabarrett: I think this was about understanding if the user has a workaround that they're using to 1) understand what a different solution might look like, and 2) to help us prioritise feature suggestions. But I can see it's not very clear from that description. I've changed it to this to hopefully make it clear that we don't expect a user to try and come up with other solutions if they're not already using one:

```
## Alternatives
<!-- Are you currently using a workaround / alternative solution instead? -->
```

.github/ISSUE_TEMPLATE/feature-request.md
vanitabarrett: That's a good point, I think that's an oversight - we don't actually we want to discourage that. 

I wonder if we can say something along the lines of making sure to check the issue hasn't been raised already, but if it has to add your comment to that issue to help us understand when multiple users are experiencing the same problems/need the same features.

Do you have any ideas on this @EoinShaughnessy ?

.github/ISSUE_TEMPLATE/config.yml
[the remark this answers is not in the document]
36degrees: Yep, although raising an issue is also a valid way of 'getting in touch' and we don't want to mis-direct people who would otherwise make a suggestion via an issue so… what about something like…

```
name: Get in touch another way
url: https://design-system.service.gov.uk/get-in-touch/
about: Find out how to get in touch via email or Slack
```

.github/ISSUE_TEMPLATE/bug-report.md
```diff
@@ -0,0 +1,33 @@
...
+assignees: ''
+
+---
+
+<!--
+    Before you raise this issue, please check that the issue isn’t already captured in:
+    - an existing Github issue: https://github.com/alphagov/govuk-frontend/issues
+    - our known validation error/warning list: https://github.com/alphagov/govuk-frontend/issues/1280#issuecomment-509588851
+
+    Please fill in as much of the template below as you’re able to. If you're unsure whether the issue already exists or how to fill in the template, open an issue anyway. Our team will help you to complete the rest.
```
hannalaakso, line 15: How would we feel about putting this section in bold in each of the templates? I'd be really keen to emphasise the points here, especially during the trial phase when we don't know whether  the issue templates could discourage some users from raising issues. Anything that could potentially lower that barrier would be good.

.github/ISSUE_TEMPLATE/feature-request.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Related component
+<!-- Does this feature relate to an existing component? -->
+
+## Context
+<!-- What are you trying to do? Is this something you think should behave differently, or something that you currently cannot do? Is this related to an existing issue/bug? -->
+
+## Alternatives
+<!-- Are you currently using a workaround / alternative solution instead? -->
+
```
hannalaakso, line 30: Just an idea but I wonder if we should add one more section to ask whether the person raising this thinks that any documentation for the feature might be required. Perhaps especially in terms of whether the feature will need guidance to be added to the Design System website? 

The section name could be followed by "(if applicable)" as we can't expect users to necessarily be aware of how we document things but it would be useful insight if they had any thoughts in this area.

.github/ISSUE_TEMPLATE/bug-report.md
vanitabarrett: @hannalaakso Do you mean moving this out of a comment within the template, into bold text that's actually rendered when the issue is previewed/created?

.github/ISSUE_TEMPLATE/feature-request.md
vanitabarrett: @hannalaakso That's an interesting idea - can you give an example of the kind of thing a user might note down in this section, so I get a better idea of what you mean?

hannalaakso: For instance, with a feature request such as https://github.com/alphagov/govuk-frontend/issues/1656 it would be interesting to understand how the person raising the issue sees the feature being documented in the Design System. This particular feature request touches on all the components and other things too, so it might not be immediately clear to our team where/how we would tell our users about it. Asking questions about documentation could possibly prompt the person raising the card to expand on whether the feature they've proposed would be opt-in or on by default too. 

But I'm happy to leave this out at least for now, we'd probably want to think about it a bit more..! Maybe once we have a better idea of how the issue templates are received we can think of further tweaks like this.

.github/ISSUE_TEMPLATE/bug-report.md
hannalaakso: Ignore my comment about bolding (that won't work, will it 😂 ), I was a bit tired when I wrote this. What I meant was that if there's a way of making this section more prominent that would be great. I wonder if we could consider swapping the two paragraphs in the comment as I'd personally much rather that the person raising the issue saw the first paragraph rather than the second one.

.github/ISSUE_TEMPLATE/feature-request.md
vanitabarrett: Thanks @hannalaakso , really helpful to see that issue! From my point of view, I think I'd prefer to leave that out for now (at least for the trial) and see how things go... My gut instinct is that for the majority of issues raised it'll be clear how any change affects our documentation and it'll be something we can deal with internally as a team (I guess that's something we can test during the trial), so I'm leaning towards leaving that section out and either expecting more advanced users to pro-actively add details if they think it requires a big documentation change __or__ we have to go back to the user and ask.

.github/ISSUE_TEMPLATE/bug-report.md
vanitabarrett: Haha, no worries! That's a good point, I think it will work better with that paragraph at the top 👍

.github/ISSUE_TEMPLATE/feature-request.md
EoinShaughnessy: Sorry for the late response! This is what me and Vanita discussed adding:

"Your issue might already exist. If so, add a comment to the existing issue instead of creating a new one. You can find existing issues here:
  - all GOV.UK Frontend issues: https://github.com/alphagov/govuk-frontend/issues
- known validation error/warning list: https://github.com/alphagov/govuk-frontend/issues/1280#issuecomment-509588851"

Let me know if I should edit the window above to include this (or similar)!

vanitabarrett: @36degrees I've updated the templates with Eoin's suggestion - do you think that makes it clearer?

36degrees: I think this file (`feature-request.md`) still needs updating?

vanitabarrett: Done 👍

36degrees (approved):
🎉

Issue #2011, opened 2020-11-09
Address notification banner accessibility concerns


vanitabarrett:
## What
Follow up on the [feedback](https://github.com/alphagov/govuk-frontend/issues/1999#issuecomment-723020026) we got from the accessibility clinic and our own accessibility testing.

## Done when
We've taken some action on all of the concerns below or documented why we don't think changes are needed:

- [x] Users may miss a notification banner at the top of a page if they skip straight to the H1
- [x] Positioning of banner and length of content may mean more scrolling for users on smaller devices/screen magnifiers to reach the main page content
- [x] Autofocus behaviour: remove tabindex on blur so the banner is not focusable on click
- [x] Should an error summary and notification banner ever be used together? An action link in the error summary means it's more likely a user will skip past the banner
- [x] Styling a paragraph as a heading may cause confusion for users who override styles or disable CSS

hannalaakso:
One possible course of action is to add some of these under the
> We need more research to understand ...

section in the [guidance](https://docs.google.com/document/d/1SNRBfwvZGS0Iqa_IfMMKwBXSC4bR_D3XrWCQPcbM9gI/edit#heading=h.kvth2ctbmufd).

hannalaakso:
### Update:
- Users may miss a notification banner at the top of a page if they skip straight to the H1. Positioning of banner and length of content may mean more scrolling for users on smaller devices/screen magnifiers to reach the main page content
 **Action**: We'll be including an example of a banner with long content in our upcoming accessibility research prototype. We'll document any relevant findings in the research section of the guidance.
- Autofocus behaviour: remove tabindex on blur so the banner is not focusable on click
  **Action:** Addressed by https://github.com/alphagov/govuk-frontend/pull/2014.
- Should an error summary and notification banner ever be used together? An action link in the error summary means it's more likely a user will skip past the banner
**Action**: We're going to specify in the guidance that teams shouldn't place an alert notification banner and an error summary on the same page.
- Styling a paragraph as a heading may cause confusion for users who override styles or disable CSS
**Action:** We'll be including an example of a banner with paragraph styled as a heading in our upcoming accessibility research prototype. We'll document any relevant findings in the research section of the guidance.

Issue #1999, opened 2020-10-27
Present notification banner at the accessibility clinic


hannalaakso:
## What
We're taking the notification banner to the GDS Accessibility clinic for testing it with different assistive technologies. We've also got a few specific questions we want to ask at the clinic (see below) (part of https://github.com/alphagov/govuk-frontend/issues/1935)

## Why
To get feedback on accessibility related points and to test against different AT.

## Who needs to know about this
Hanna; Ollie's kindly offered to present what we have at the clinic but other people from the team might also want to join the call.

## Done when
- [x] We've created realistic example pages to demonstrate the component 
- [x] We've presented what we have at the clinic
- [x] We've made notes of any interesting findings and documented follow up actions 
- [x] We've posted a summary of the feedback on https://github.com/alphagov/govuk-design-system-backlog/issues/2

### Questions to ask
- What happens if there is more than one element on the page with `role="alert"`?
- Which screen readers require the focus to be set on the component, in addition to already setting `role="alert"`? (We'd quite like to [document the detail in the code](https://github.com/alphagov/govuk-frontend/pull/1994/files#diff-bd2c727342e9eb043d2adf1d61d802419c192a6c1cb5da317642c0aec413dbd0R22))
- Is it okay on occasion to apply the DS heading styles to a `<p>` tag? (link to example)
- Does it make sense that we set the `tabindex` to `-1`, not `0`? What's the user experience like if the notification banner is focused on page load and then the user wants to tab back to it as an interactive element but it's no longer there? (The error summary contains links so those can still be navigated to later on)
- Would it make sense to remove the `tabindex` after focusing?

CharlotteDowns:
### Addition to the accordion prototype for testing
For the accessibility clinic we are going to test an accordion with description text as well as without.

vanitabarrett:
The Design System team took the latest iteration of the notification banner for an internal accessibility review. We were hoping to have addressed some of the [feedback we had from the last accessibility review](https://github.com/alphagov/govuk-design-system-backlog/issues/2#issuecomment-673356072).

As we've recently got AssistivLabs, we focused more on the specific questions we had rather than general testing. It might be a good idea to do some general testing of the banners with other screenreaders. In this review, we mainly used JAWS.

## Differentiating between the types of banners
[...]

PR #2014, opened 2020-11-11
Remove tabindex attribute from notification alert banner on blur

hannalaakso:
There is no need for the notification banner component to remain focusable after the page has loaded and the focus has moved away from it. This is following some feedback from the GDS Accessibility clinic: https://github.com/alphagov/govuk-frontend/issues/1999#issuecomment-723020026. 

This PR:
- Removes the tabindex attribute from the notification alert banner on blur 
- Tests that the attribute gets removed correctly
- Also tests that if the tabindex has been set as an attribute in the HTML, it doesn't get removed by the script.

Partially addresses https://github.com/alphagov/govuk-frontend/issues/2011

### DIFF
--- a/src/govuk/components/notification-banner/notification-banner.js
+++ b/src/govuk/components/notification-banner/notification-banner.js
@@ -37,8 +37,14 @@ NotificationBanner.prototype.setFocus = function () {
   }
 
   // Set tabindex to -1 to make the element focusable with JavaScript.
+  // Remove the tabindex on blur as the component doesn't need to be focusable after the page has
+  // loaded.
   if (!$module.getAttribute('tabindex')) {
     $module.setAttribute('tabindex', '-1')
+
+    $module.addEventListener('blur', function () {
+      $module.removeAttribute('tabindex')
+    })
   }
 
   $module.focus()
--- a/src/govuk/components/notification-banner/notification-banner.test.js
+++ b/src/govuk/components/notification-banner/notification-banner.test.js
@@ -6,7 +6,7 @@ const PORT = configPaths.ports.test
 const baseUrl = 'http://localhost:' + PORT
 
 describe('/components/notification-banner/with-type-as-success', () => {
-  it('has the correct tabindex to be focused with JavaScript', async () => {
+  it('has the correct tabindex attribute to be focused with JavaScript', async () => {
     await page.goto(baseUrl + '/components/notification-banner/with-type-as-success/preview', { waitUntil: 'load' })
 
     const tabindex = await page.$eval('.govuk-notification-banner', el => el.getAttribute('tabindex'))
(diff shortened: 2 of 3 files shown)


src/govuk/components/notification-banner/notification-banner.test.js
```diff
@@ -21,6 +21,15 @@ describe('/components/notification-banner/with-type-as-success', () => {
...
     expect(activeElement).toBe('govuk-notification-banner')
   })
+
+  it('removes the tabindex attribute on blur', async () => {
+    await page.goto(baseUrl + '/components/notification-banner/with-type-as-success/preview', { waitUntil: 'load' })
+
+    await page.$eval('.govuk-notification-banner', el => el.blur())
+
+    const tabindex = await page.$eval('.govuk-notification-banner', el => el.getAttribute('tabindex'))
+    expect(tabindex).toBeFalsy()
```
36degrees, line 31: I think we could probably use `toBeUndefined ` here – `toBeFalsy` wouldn't fail if `tabindex` was 0, for example.

```suggestion
    expect(tabindex).toBeUndefined()
```

src/govuk/components/notification-banner/notification-banner.test.js
```diff
@@ -62,3 +71,16 @@ describe('components/notification-banner/role=alert-overridden-to-role=region,-w
     })
   })
 })
+
+describe('/components/notification-banner/custom-tabindex', () => {
+  describe('when custom tabindex set', () => {
```
36degrees, line 76: Really good idea to have a test for this 👍🏻

src/govuk/components/notification-banner/notification-banner.js
```diff
@@ -37,8 +37,14 @@ NotificationBanner.prototype.setFocus = function () {
   }
 
   // Set tabindex to -1 to make the element focusable with JavaScript.
+  // Remove the tabindex on blur as the component doesn't need to be focusable after the page has
+  // loaded.
   if (!$module.getAttribute('tabindex')) {
     $module.setAttribute('tabindex', '-1')
+
+    $module.addEventListener('blur', function () {
+      $module.removeAttribute('tabindex')
```
36degrees, line 46: I'm curious if removing the `tabindex` attribute immediately after focusing it does the same job, although I don't think there's anything wrong with this approach – it just feels like it would be _ever so slightly_ simpler, and avoids having an event listener hanging around.

Not sure it's worth spending any more time on though – what do you think?

36degrees (commented):
Nice 👍🏻

src/govuk/components/notification-banner/notification-banner.js
hannalaakso: Good question! Do you mean that we would just have:
```
 $module.focus()
 $module.removeAttribute('tabindex')
```

In my testing the focus isn't set if you do that. I'd be interested in understanding why the focus hasn't finished setting by the time you remove the tabindex 🤓  but like you said it might not be worth spending more time on.

src/govuk/components/notification-banner/notification-banner.test.js
hannalaakso: `toBeFalsy()` does fail here if `tabindex` is set to 0. 

Puppeteer returns `null` for things that don't exist in the DOM so for this reason `toBeUndefined()` can't used here. 

I think we're okay to leave this as is but let me know if you think further digging is needed 🙏

36degrees: Ah, I guess because it's the string "0" rather the number 0?

I think [`toBeNull`](https://jestjs.io/docs/en/expect#tobenull) might be what we want, in that case?

```suggestion
    expect(tabindex).toBeNull()
```

36degrees (approved):

PR #2027, opened 2020-11-19
Change visited state for error summary and notification banner

vanitabarrett:
We recommend not having a different visited link state where it is not helpful to distinguish between visited and unvisited states, for example when linking to pages with frequently-changing content such as the dashboard for an admin interface.

It doesn't seem useful to highlight visited links in the notification banner, and the previous behaviour for the error summary also matches this.

Therefore this commit updates the error and success link style mixins so that the visited colour is the same as the original link colour.

### DIFF
--- a/src/govuk/helpers/_links.scss
+++ b/src/govuk/helpers/_links.scss
@@ -82,12 +82,9 @@
 /// @access public
 
 @mixin govuk-link-style-error {
-  &:link {
-    color: $govuk-error-colour;
-  }
-
+  &:link,
   &:visited {
-    color: $govuk-link-visited-colour;
+    color: $govuk-error-colour;
   }
 
   &:hover {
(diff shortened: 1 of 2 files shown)


src/govuk/helpers/_links.scss
```diff
@@ -87,7 +87,7 @@
   }
 
   &:visited {
-    color: $govuk-link-visited-colour;
+    color: $govuk-error-colour;
```
36degrees, line 90: Worth merging these together, do you think?

```scss
&:link,
&:visited {
  color: $govuk-error-colour;
}
```

vanitabarrett: I was following what we do for the no-visited mixin, but happy to combine - makes sense to me 👍 https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/helpers/_links.scss#L265

36degrees: Ooh, I hadn't noticed that. Happy either way!

36degrees (approved):

PR #2028, opened 2020-11-19
Remove stray dash from banner nunjucks examples

vanitabarrett:
There was a stray dash after some of the email addresses in the notification banner examples


36degrees (approved):
👍🏻

PR #2030, opened 2020-11-20
Remove error notification banner type

vanitabarrett:
## What
Remove 'error' type notification banner

## Why
We struggled to find a use case for the error banner that wasn't already covered by the error summary, an information notification banner or another component/pattern. For now, we're going to release the success and information banner and add the error banner later if we identify a strong use case.


EoinShaughnessy (approved):
Text changes look good!

Nooshu (approved):

PR #2035, opened 2020-11-25
Fix hairline gap between notification banner header and outer border on high resolution screens in Chrome/Edge (Blink)

### LINKED ISSUE #2034: White line around notification banner title
## Description of the issue
The new notification banner component's title has a thin white line around it. I've had a quick look and couldn't see any obvious causes but I haven't studied it at any length.

Windows/Chrome
image

Android/Chrome
image

Assuming the white line shouldn't be there...

- Operating system: Windows 10
- Browser: Chrome
- Browser version:  86.0.4240.198
- GOV.UK Frontend Version: 3.10.0

And:

- Operating system: Android 10 (Google Pixel XL)
- Browser: Chrome
- Browser version:  86.0.4240.198
- GOV.UK Frontend Version: 3.10.0

andymantell:
To fix #2034 (White line around notification banner heading on high DPI screens in chrome) I have flipped the colouring of the notification banner around. This means the background colour of the entire notification banner is now blue/green, but the content container inside it is white - meaning there is no gap between the two any more.

I had to remove bottom margin from the last child inside the content div - reason being I had to switch the content container to padding instead of margins but that causes it to contain the bottom margin of the last child.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+### Fixes
+
+We’ve made fixes to GOV.UK Frontend in the following pull requests:
+
+- [#2035: Fix hairline gap between notification banner header and outer border on high resolution screens in Chrome/Edge (Blink)](https://github.com/alphagov/govuk-frontend/pull/2035)
+
 ## 3.10.0 (Feature release)
 
 ### New features
--- a/src/govuk/components/notification-banner/_index.scss
+++ b/src/govuk/components/notification-banner/_index.scss
@@ -5,6 +5,8 @@
 
     border: $govuk-border-width solid $govuk-brand-colour;
 
+    background-color: $govuk-brand-colour;
+
     &:focus {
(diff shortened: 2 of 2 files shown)


36degrees (approved):
Looks good to me – thanks @andymantell 

My only concern was around High Contrast Mode but visually it seems to behave exactly the same:

## Before

Screenshot 2020-11-25 at 11 50 34

## After

Screenshot 2020-11-25 at 11 50 13

This just needs another review from another member of the team and then we can get this merged.

vanitabarrett:
Thanks @andymantell , good spot and this fix looks good to me!

Would you be able to add a quick entry in the Changelog too? Something similar to [this example](https://github.com/alphagov/govuk-frontend/blob/master/CHANGELOG.md#fixes).

andymantell:
Thanks both.

Will changelog it now...

vanitabarrett (approved):

36degrees:
Thanks for the really clear update to the PR title 👍🏻

PR #2036, opened 2020-11-25
Explicitly set the text colour for notification banner content, rather than using the user-agent default text colour

### LINKED ISSUE #2037: Content within the notification banner uses the default user-agent text colour
## Description of the issue

The notification banner content currently uses the user agent default text colour.

This means that for users who have not changed their user agent default text colour, the text will be true black (`#000000`) rather than the ‘GOV.UK’ black we specify in the colour palette (`#0b0c0c`).

It also means that if a user has changed their default text colour, but has not chosen to override colours, depending on the user’s prefered colour scheme the text may have low contrast or be invisible.

As I understand it, this is as a [‘failure of Success Criterion 1.4.3, 1.4.6 and 1.4.8 due to specifying foreground colors without specifying background colors or vice versa’][1].

For example, my Firefox instance is configured to use white text on a blue background:

Firefox colour preferences, with 'Text' set to white, 'Background' set to blue, and Unvisited/Visited links set to yellow, and 'Override the colors specified by the page with your selections above' set to 'Never'

The notification banner is therefore white-on-white, and thus invisible:

The notification banner showing invisible text

Rather than:

The fixed notification banner with visible text

[1]: https://www.w3.org/TR/WCAG20-TECHS/F24.html

36degrees:
The notification banner content currently uses the user agent default text colour.

This means that for users who have not changed their user agent default text colour, the text will be true black (`#000000`) rather than the ‘GOV.UK’ black we specify in the colour palette (`#0b0c0c`).

It also means that if a user has changed their default text colour, but has not chosen to override colours, depending on the user’s prefered colour scheme the text may have low contrast or be invisible.

For example, my Firefox instance is configured to use white text on a blue background:

Firefox colour preferences, with 'Text' set to white, 'Background' set to blue, and Unvisited/Visited links set to yellow, and 'Override the colors specified by the page with your selections above' set to 'Never'

The notification banner is therefore white-on-white, and thus invisible:

The notification banner showing invisible text

As I understand it, this is as a [‘failure of Success Criterion 1.4.3, 1.4.6 and 1.4.8 due to specifying foreground colors without specifying background colors or vice versa’][1].

Set the text colour of the `govuk-notification-banner__content` element using the `govuk-text-colour` mixin to fix this:

The fixed notification banner with visible text

[1]: https://www.w3.org/TR/WCAG20-TECHS/F24.html

Fixes #2037

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 We’ve made fixes to GOV.UK Frontend in the following pull requests:
 
 - [#2035: Fix hairline gap between notification banner header and outer border on high resolution screens in Chrome/Edge (Blink)](https://github.com/alphagov/govuk-frontend/pull/2035)
+- [#2036: Explicitly set the text colour for notification banner content, rather than using the user-agent default text colour](https://github.com/alphagov/govuk-frontend/pull/2036)
 
 ## 3.10.0 (Feature release)
 
--- a/src/govuk/components/notification-banner/_index.scss
+++ b/src/govuk/components/notification-banner/_index.scss
@@ -34,6 +34,7 @@
   }
 
   .govuk-notification-banner__content {
+    @include govuk-text-colour;
     padding: govuk-spacing(3);
 
     background-color: $govuk-body-background-colour;


vanitabarrett (approved):
Think this just needs a Changelog entry, then it's good to go

PR #2039, opened 2020-11-25
Expand git tag error in build-release script

vanitabarrett:
## What
Expands the git tag error message in the build-release script to include:

- The name of the tag which already exists
- A suggestion for what might be causing the error

## Why
This tripped me up when I was doing the release - I had forgotten to update the package.json file, but thought the error was about the new release tag already existing. Hopefully adding the tag itself to the error message and suggesting the package.json check will help if anyone else gets this error in the future.


36degrees (approved):
Nice improvement! 🏷️

PR #2042, opened 2020-11-26
Wrap notification banner content at 2/3 grid width

### LINKED ISSUE #2033: Suggestion: set max-width on notification banner content
The notification banner doesn’t set any limits on content width. Unlike other components, I think it's often likely to be used in full-width columns.

The current guidance says:
> The notification banner should be the same width as the page content.

Which is a bit ambiguous about whether the banner should be 2/3 or whether it should match the other page content. If it matches the other page content, then the content of the banner should be constrained.

My suggestion would be to set `max-width` on `govuk-notification-banner__content` to `660px` or `44ex` or similar.

Example without max width:
Screenshot 2020-11-24 at 11 53 59

Example with max width:
Screenshot 2020-11-24 at 11 56 15

If the banner should only ever be 2/3 width, then the guidance could be more explicit to say that.

36degrees:
Maintain shorter line-lengths when the notification banner is full width by wrapping at the same point a 2/3 column ends.

👉🏻  &nbsp; [See examples](https://govuk-frontend-review-pr-2042.herokuapp.com/components/notification-banner)

Screenshot 2020-11-26 at 11 57 29

With increased root text size:

Screenshot 2020-11-26 at 11 57 54

With a custom page size and gutter width:

Screenshot 2020-11-26 at 11 59 42

Fixes #2033

### DIFF
--- a/src/govuk/components/notification-banner/_index.scss
+++ b/src/govuk/components/notification-banner/_index.scss
@@ -34,13 +34,14 @@
   }
 
   .govuk-notification-banner__content {
+    $padding-tablet: govuk-spacing(4);
     @include govuk-text-colour;
     padding: govuk-spacing(3);
 
     background-color: $govuk-body-background-colour;
 
     @include govuk-media-query($from: tablet) {
-      padding: govuk-spacing(4);
+      padding: $padding-tablet;
     }
 
     // Wrap content at the same place that a 2/3 grid column ends, to maintain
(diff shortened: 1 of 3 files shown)


edwardhorsford:
@36degrees do you think a similar thing should be applied to the error summary?

src/govuk/components/notification-banner/_index.scss
```diff
@@ -43,6 +43,21 @@
...
+    > * {
+      // When elements have their own padding (like lists), include the padding
+      // in the max-width calculation
+      box-sizing: border-box;
+
+      // Calculate the internal width of a two-thirds column...
+      $two-col-width: ($govuk-page-width * 2 / 3) - ($govuk-gutter * 1 / 3);
+
+      // ...and then take the border width and internal padding into account
+      $banner-exterior: (govuk-spacing(4) + $govuk-border-width);
```
vanitabarrett, line 57: Should we save `govuk-spacing(4)` in a variable so we don't accidentally change the internal padding of the banner and forget to update the formula?

christopherthomasdesign (approved):
Checked out locally and tried out on a few different page layouts. Looks great, top work 💥

src/govuk/components/notification-banner/_index.scss
36degrees: Good idea 👍🏻  Done in 54ca621561401bd87ae7df07a2b4e371793c8cab.

vanitabarrett (approved):

PR #2043, opened 2020-11-26
Fixes for a couple of component examples / fixtures

andymantell:
I have spotted a couple of issues with the current fixtures whilst using them in govuk-react-jsx. This PR is to fix them...

1) on the `fieldset params` example in the checkboxes component, the `legend` param was not nested underneath the `fieldset` param.

2) On the select component, `value` is not marked as required underneath the items / individual options. If not supplied, the component will render every option with an empty value which I think doesn't make much sense. I have therefore marked it as required, and fixed two of the hidden examples to match (All the other examples already had `value` specified. This was causing problems in my react port since React cannot have two `<option>` elements with the same underlying `value`, even if that value is blank.

### DIFF
--- a/src/govuk/components/checkboxes/__snapshots__/template.test.js.snap
+++ b/src/govuk/components/checkboxes/__snapshots__/template.test.js.snap
@@ -6,6 +6,9 @@ exports[`Checkboxes nested dependant components passes through fieldset params w
           data-attribute="value"
           data-second-attribute="second-value"
 >
+  <legend class="govuk-fieldset__legend">
+    What is your nationality?
+  </legend>
 </fieldset>
 `;
 
--- a/src/govuk/components/checkboxes/checkboxes.yaml
+++ b/src/govuk/components/checkboxes/checkboxes.yaml
@@ -710,9 +710,9 @@ examples:
     name: example-name
     errorMessage:
       text: Please select an option
-    legend:
-      text: What is your nationality?
     fieldset:
+      legend:
+        text: What is your nationality?
       classes: app-fieldset--custom-modifier
       attributes:
         data-attribute: value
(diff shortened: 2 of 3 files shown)


andymantell:
Ah, I spy a failing snapshot, will sort after lunch

andymantell:
I've marked this as a draft for a second as I've got more changes coming...

andymantell:
That should be all the changes now so this is ready for review. I've got govuk-react-jsx passing 100% against the govuk-frontend fixtures now, with the changes made in this pull request hacked into my library as overrides.

vanitabarrett:
Thank you for raising this PR @andymantell ! I think on the whole we're happy with these changes and it's great to see the examples being made more consistent! 

However, we're not sure about marking `value` as a required option for selects. At the moment, not providing a `value` means you end up with an empty string as a value - this is valid HTML. It's possible that in the future we may want to revisit the macro behaviour so that not passing `value` actually results in what you would perhaps expect (no value on the option), and therefore marking it as required would be correct. But this would be a breaking change, so not something we're in a position to look at yet.

Would you be happy with updating this PR to remove the change that marks `value` as required? If so, I'll also ask our tech writer to take a look to see if we can update the documentation for that option to make it clearer that you get an empty string if not provided.

andymantell:
Ok that's fair enough. I'll remove the required option. I suppose even though I would consider that output with duplicate empty values to be somewhat weird, there may be some use case that I haven't envisaged...

From my perspective, so long as the actual examples contain values that are unique and non empty then that works.

andymantell:
@vanitabarrett That's done now

vanitabarrett (approved):

36degrees (approved):
Thanks @andymantell 🙏🏻

PR #2045, opened 2020-11-27
Stop same-site cookies from being wiped when printing in Internet Explorer 11

gunjam:
**tl;dr** this fixes an issue where a service with `SameSite: Strict` set on its session cookies will have them wiped in IE when opening the print dialog if they serve a page on the root of their domain.

We were experiencing an odd issue in our service where some users were unexpectantly getting sent back to the beginning of the service from the Check Your Answers page. After some research with users this appeared to only happen after they attempted to print their answers. A lot of head scratching later I figured out what was going on: when a user in IE11 (and presumably lower) opens the print dialog, the browser re-requests all uncached assets and in doing will make a GET request for `xlink:href` value in the crown SVG image tag, even though it's blank. It'll resolve this to the root of the domain which for us is the first page of our service. The problem is that if you have `SameSite: Strict` set on your cookies, IE won't send them when making the image request to `/` so our service will act as if this is a new session and send a new session cookie with the response, which IE will then propagate back into the page resetting the session cookie. On submission, the user is now trying to access a page at the end of the service with a blank session, so gets sent back to the start.

A pretty odd issue, typical of IE, but will affect any service with a session based page being served on `/`, unless you set `SameSite` to something other than `Strict`, which we don't really want to do.

The simplest solution to this I've come up with, is to set the empty `xlink:href` value to a blank data URI so that the browser will not initiate a network request. I've also set the SVG `display` attribute to `none` so browsers do no render a small blank SVG in place.

### DIFF
--- a/src/govuk/components/header/template.njk
+++ b/src/govuk/components/header/template.njk
@@ -33,9 +33,9 @@
 
             The <image> element is a valid SVG element. In SVG, you would specify
             the URL of the image file with the xlink:href – as we don't reference an
-            image it has no effect. It's important to include the empty xlink:href
-            attribute as this prevents versions of IE which support SVG from
-            downloading the fallback image when they don't need to.
+            image it has no effect. It's important to include an empty data URI
+            as the xlink:href attribute as this prevents versions of IE which
+            support SVG from downloading the fallback image when they don't need to.
 
             In other browsers <image> is synonymous for the <img> tag and will be
             interpreted as such, displaying the fallback image. #}
--- a/src/govuk/components/header/template.njk
+++ b/src/govuk/components/header/template.njk
@@ -39,7 +39,7 @@
 
             In other browsers <image> is synonymous for the <img> tag and will be
             interpreted as such, displaying the fallback image. #}
-            <image src="{{ params.assetsPath | default('/assets/images') }}/govuk-logotype-crown.png" xlink:href="" class="govuk-header__logotype-crown-fallback-image" width="36" height="32"></image>
+            <image src="{{ params.assetsPath | default('/assets/images') }}/govuk-logotype-crown.png" xlink:href="data:," display="none" class="govuk-header__logotype-crown-fallback-image" width="36" height="32"></image>
           </svg>
           <span class="govuk-header__logotype-text">
             GOV.UK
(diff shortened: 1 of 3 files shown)


36degrees:
Thanks for contributing this @gunjam – really amazing investigative work getting to the bottom of that!

I've been able to verify the issue by tracking network requests in ngrok.

When the ['check your answers' page in our review app](http://govuk-frontend-review.herokuapp.com/full-page-examples/check-your-answers) is first loaded in IE11 (using BrowserStack), the following requests are made:

before-pre-print

When pressing 'Print' in the browser several additional network requests are made, including a request for the fallback PNG but also an unexpected request to `/full-page-examples/` which as you've identified appears to be caused by the empty `xlink:href`.

before-post-print

With this branch checked out, the request to `/full-page-examples/` does not happen:

after-post-print

> I've also set the SVG `display` attribute to `none` so browsers do no render a small blank SVG in place.

This fixes a tiny bordered square in IE10 and IE11 that I hadn't noticed before.

| Browser | Before | After |
| - | - | - |
| IE 9 | IE9-before | IE9-after |
| IE 10 | IE10-before | IE10-after |

We do need to rebase this to pick up the latest changes to our CI setup, which I can do for you now. I'll also add an entry to the changelog.

Finally, this will need another review from another member of the team and then we can get it merged.

36degrees:
I've rebased this, added a test for the `display` attribute (as I think someone could easily try and remove it, not understanding what it does), and added an entry to the changelog.

@gunjam are you happy with the changelog entry and the updated PR title? I've tried to focus it around the issue it fixes, rather than the change that was made.

36degrees (approved):
I'm happy with this, pending another review from the team.

gunjam:
More than happy, thanks @36degrees .

src/govuk/components/header/template.njk
```diff
@@ -39,7 +39,7 @@
```
hannalaakso, line 39: It would be good to correct the above comment to say 
> an empty data URI xlink:href

instead of
> empty xlink:href attribute

hannalaakso (approved):
Thanks for your work on this @gunjam 🕵️

src/govuk/components/header/template.njk
gunjam: Hi @hannalaakso I've made the changes you suggested. Thanks!

calvin-lau-gds:
Hi @gunjam, thanks again for your contribution.

We'd like to add your name to a list of contributors in some of our release comms. Can you let me know your full name and where you work?

Thanks kindly!

PR #2048, opened 2020-11-30
Add modifiers for table captions

frankieroberto:
This makes table captions consistent with fieldset legends and labels, in having modifiers which apply the typographic styles equivalent to headings.

Resolves [issue 1405](https://github.com/alphagov/govuk-design-system/issues/1405).

### DIFF
--- a/src/govuk/components/table/_index.scss
+++ b/src/govuk/components/table/_index.scss
@@ -47,4 +47,25 @@
     display: table-caption;
     text-align: left;
   }
+
+  // Modifiers that make captions look more like their equivalent headings
+
+  .govuk-table__caption--xl {
+    @include govuk-font($size: 48, $weight: bold);
+    margin-bottom: govuk-spacing(3);
+  }
+
+  .govuk-table__caption--l {
+    @include govuk-font($size: 36, $weight: bold);
+    margin-bottom: govuk-spacing(3);
+  }
+
+  .govuk-table__caption--m {
+    @include govuk-font($size: 24, $weight: bold);
+    margin-bottom: govuk-spacing(3);
+  }
+
+  .govuk-table__caption--s {
+    @include govuk-font($size: 19, $weight: bold);
+  }
 }


src/govuk/components/table/_index.scss
```diff
@@ -47,4 +47,25 @@
     display: table-caption;
     text-align: left;
   }
+
+  // Modifiers that make captions look more like their equivalent headings
+
+  .govuk-table__caption--xl {
+    @include govuk-font($size: 48, $weight: bold);
+    margin-bottom: govuk-spacing(3);
```
CharlotteDowns, line 55: Do we need to also include a line-height style for each modifier?

src/govuk/components/table/_index.scss
```diff
@@ -47,4 +47,25 @@
...
+  }
+
+  .govuk-table__caption--m {
+    @include govuk-font($size: 24, $weight: bold);
+    margin-bottom: govuk-spacing(3);
+  }
+
+  .govuk-table__caption--s {
+    @include govuk-font($size: 19, $weight: bold);
```
CharlotteDowns, line 69: Do we also need to include a margin-bottom for `govuk-table__caption--s`, I'm not sure?

src/govuk/components/table/_index.scss
```diff
@@ -47,4 +47,25 @@
...
+  }
+
+  .govuk-table__caption--l {
+    @include govuk-font($size: 36, $weight: bold);
+    margin-bottom: govuk-spacing(3);
+  }
+
+  .govuk-table__caption--m {
+    @include govuk-font($size: 24, $weight: bold);
+    margin-bottom: govuk-spacing(3);
```
CharlotteDowns, line 65: In the [typography section of the Design System](https://design-system.service.gov.uk/styles/typography/#headings), the margin-bottom for --m and --s is slightly smaller than the --l heading. Does it look odd to follow these margin styles on the table component?

CharlotteDowns (commented):
This looks like a really good addition. I've left a few comments but you may have chosen to not do them intentionally based on the context. Let me know what you think and thanks again :).

src/govuk/components/table/_index.scss
36degrees: The `govuk-font` mixin will automatically include the `line-height` appropriate for the specified font size as defined in the typography settings (although it can be overridden by passing a specific `$line-height`)

https://github.com/alphagov/govuk-frontend/blob/9b320cece2177f3a9749d8150e250fe8978e7eef/src/govuk/settings/_typography-responsive.scss#L85-L98

src/govuk/components/table/_index.scss
frankieroberto: Good question. I followed the code from the [fieldset legend modifiers](https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/components/fieldset/_index.scss#L37-L54), which only add a margin-bottom of `3` for the `xl`, `l` and `m` modifiers.

CharlotteDowns: Thanks so much for clarifying @frankieroberto I'm happy to approve :)

CharlotteDowns (approved):

36degrees:
Do we need to update the Design System tables guidance to document these new classes?

frankieroberto:
@36degrees I think it makes sense to mention these options on the [table component page](https://design-system.service.gov.uk/components/table/)?

I'd be tempted to make the default example on that page using a larger caption size too.

36degrees (approved):

PR #2049, opened 2020-12-01
Rename notification banner full page example

vanitabarrett:
## What
Rename the notification banner full page example to refer to 'success' rather than 'error'

## Why
This example originally used an error notification banner, but this style of banner was removed as we could not find a realistic use case. We switched the example to use the success type notification banner when we did the release, but forgot to rename the files


36degrees (approved):
👍🏻

PR #2052, opened 2020-12-02
Move CI to use GitHub Actions

### LINKED ISSUE #2051: Move CI to GitHub Actions
## What

Move GOV.UK Frontend's CI from Travis to GitHub Actions

## Why

Travis' changes to our plan means that builds of GOV.UK Frontend are getting stuck in a queue, which means they can take around 30 minutes to run.

Given that the direction of travel for the organisation seems to be towards GitHub Actions, and to be consistent with the approach we're taking with the Prototype Kit, move to GitHub Actions.

Developers

## Further detail

The current [Travis config](https://github.com/alphagov/govuk-frontend/blob/master/.travis.yml) defines four jobs:

- Run the test suite
- Check that the Sass compiles using LibSass v3.3.0
- Check that the Sass compiles using Dart Sass v1.0.0
- Check that the Sass compiles using Ruby Sass v3.4.0

Equivalent jobs need to be run using GitHub Actions.

## Done when

- [ ] Existing open PRs merged / closed where possible to avoid needing to rebase
- [x] GitHub Actions config added
- [x] Travis config removed
- [ ] Required status checks updated
- [x] Documentation updated
  - [x] [Testing](https://github.com/alphagov/govuk-frontend/blob/master/docs/contributing/testing.md#if-github-shows-the-build-failed)

36degrees:
Add config for 2 new GitHub workflows:

- `.github/workflows/tests.yaml` – runs the tests
- `.github/workflows/sass.yml` – check that the Sass compiles with the minimum supported version of Dart Sass, LibSass & Ruby Sass

These are broadly equivalent to jobs previously run by Travis as defined in `.travis.yml`.

Remove `.travis.yml` as it's no longer required.

Other changes:

- Ruby 1.9.3 is not easily available on GitHub Actions, so update the Ruby Sass compilation test to run under Ruby 2.1.9 instead – it still uses Ruby Sass v3.4.0 which is the minimum version we support.
- Remove `bin/check-package-lock.sh` in favour of an inline `run` step in the workflow YAML file
- Update the 'badge' in the README
- Update testing documentation to explain what to do when checks on a pull request fail

This PR will show '1 expected check' and not be mergeable until we change the required status checks to require the GitHub Actions checks, and remove the Travis checks.

Closes #2051


vanitabarrett (approved):

36degrees:
@EoinShaughnessy would you be able to review the [change to the contributing 'testing' doc](https://github.com/alphagov/govuk-frontend/pull/2052/files#diff-f0dd422f285e6679f617ea7de2b688bbc149f3532f4042f9e98cb169e8eb4b25)?

docs/contributing/testing.md
```diff
@@ -109,10 +109,12 @@ When you create the pull request for your contributions, list what you’ve test
 
 If your contribution changes how a component looks, include before and after screenshots if you can.
 
-### If GitHub shows the build failed
+### If GitHub shows that some checks were not successful
```
EoinShaughnessy, line 112:
```suggestion
### If GitHub shows that some checks failed
```

docs/contributing/testing.md
```diff
@@ -109,10 +109,12 @@ When you create the pull request for your contributions, list what you’ve test
...
 
-1. At the bottom of your pull request in GitHub, find the line that says **Build Failed**.
-2. Select **Details**.
-3. Under **Build Failed**, select **The build**.
-4. Wait for the build log to appear.
-5. Use the information in the build log to find the problem and fix your code.
+At the bottom of your pull request in GitHub, find the checks that are marked as **Failing**.
```
EoinShaughnessy, line 114:
```suggestion
At the bottom of your pull request in GitHub, find the checks marked as **Failing**.
```

docs/contributing/testing.md
```diff
@@ -109,10 +109,12 @@ When you create the pull request for your contributions, list what you’ve test
...
+
+For each failing check:
+
+1. Select **Details**.
+2. Wait for the build log to appear.
```
EoinShaughnessy, line 119: Is it always the same amount of time a user waits? Or could we write something like "Wait a few moments for the build log to appear" / "Wait about 10 seconds for the build log to appear"?

docs/contributing/testing.md
```diff
@@ -109,10 +109,12 @@ When you create the pull request for your contributions, list what you’ve test
...
+### If GitHub shows that some checks were not successful
 
-1. At the bottom of your pull request in GitHub, find the line that says **Build Failed**.
-2. Select **Details**.
-3. Under **Build Failed**, select **The build**.
-4. Wait for the build log to appear.
-5. Use the information in the build log to find the problem and fix your code.
+At the bottom of your pull request in GitHub, find the checks that are marked as **Failing**.
+
+For each failing check:
```
EoinShaughnessy, line 116:
```suggestion
For each failed check:
```

EoinShaughnessy (commented):
Just added a few suggestions and had one question.

docs/contributing/testing.md
36degrees: Might be some context missing here – when the build fails the 'headline' shows 'Some checks were not successful', then lists all of the checks, with their various statuses ('Success', 'Failing')

Screenshot 2020-12-03 at 17 18 48

There's an example of a failing build here, if you want to try the steps: https://github.com/alphagov/govuk-frontend/pull/2053

So 'some checks were not successful' is trying to mimic the terms used in the GitHub interface. Happy to change it if you still think it makes sense to do so though!

docs/contributing/testing.md
36degrees: Good question. It actually appears pretty much straight away so I do wonder if we need a step that says 'Wait for the thing to appear'.

Maybe this could just be

> For each failing check, select **Details** and use the build log to find the problem and fix your code.

Or even we could consider removing this section entirely? I think it used to be helpful because there was an extra step that was not obvious, but now that extra step has been removed maybe the process is simple enough that this is redundant?

@vanitabarrett @hannalaakso any thoughts?

docs/contributing/testing.md
EoinShaughnessy: @36degrees That makes sense! Thanks for explaining. Sounds like the original wording is fine then. I'll delete this suggestion.

docs/contributing/testing.md
EoinShaughnessy: @36degrees I'll delete this suggestion as it makes sense for "failing" on line 116 to mirror "find the checks marked as **Failing**" on line 114.

EoinShaughnessy (approved):

hannalaakso (approved):
Looks great, thanks @36degrees 🚢  (The fact that we've already rolled this out in the kit repo in https://github.com/alphagov/govuk-prototype-kit/pull/956 gives me extra confidence)

36degrees:
This is blocked as for some reason trying to change which status checks are required currently results in a 500 server error. I've reached out to GitHub support.

vanitabarrett (approved):

36degrees:
Removing the emojis from the start of the job names (33882e4) has allowed us to update the required status checks, so this can now be merged.

PR #2064, opened 2020-12-12
Use non-bool type strings in test fixtures

matthew-shaw:
When using boolean cast-able string values in test fixtures, these can be interpreted as actual boolean types by ports of govuk-frontend, such as the [govuk-frontend-jinja](https://github.com/LandRegistry/govuk-frontend-jinja) Python port. For example, three tests fail due to casing differences between the expected string values of "true" and "false" and their boolean equivalents True and False: https://github.com/LandRegistry/govuk-frontend-jinja/pull/10/checks?check_run_id=1538392387#step:6:1365

Other options:
- I could code around this in my port by type-checking all param values and lower casing only bool cast-able values as strings. However, it seems better to improve the upstream test fixtures to use more explicit string type values instead.
- I could blanket lower case all attribute values, but this could break things with unexpected behaviour and would deviate from how the upstream original behaves.

Changing these test fixture params to non-bool type strings will make maintaining ports of govuk-frontend in other frameworks and languages easier.

### DIFF
--- a/docs/releasing/publishing.md
+++ b/docs/releasing/publishing.md
@@ -1,17 +1,34 @@
-# Publishing GOV.UK Frontend
+# Steps you must complete before you release a new version of GOV.UK Frontend
 
-1. Make sure you have cards in the "Backlog" column to:
-  - Update the GOV.UK Design System to use the latest release
-  - Update the GOV.UK Prototype Kit to use the latest release
-  - Update the GOV.UK Frontend Docs to use the latest release
+1. Raise new issues in the team GitHub repositories ([govuk-frontend](https://github.com/alphagov/govuk-frontend), [govuk-frontend-docs](https://github.com/alphagov/govuk-frontend-docs), [govuk-prototype-kit](https://github.com/alphagov/govuk-prototype-kit)) to:
+  - create release notes for the new release (example card: [#1986](https://github.com/alphagov/govuk-frontend/issues/1986))
+  - create release notes for the new version of GOV.UK Prototype Kit (example card: [#958](https://github.com/alphagov/govuk-prototype-kit/issues/958))
+  - create a card for the new release of GOV.UK Frontend (example card: [#1987](https://github.com/alphagov/govuk-frontend/issues/1987))
+  - update the GOV.UK Design System to use the new release of GOV.UK Frontend (example card: [#1347](https://github.com/alphagov/govuk-design-system/issues/1347))
+  - create a card for the new release of GOV.UK Prototype Kit (example card: [#917](https://github.com/alphagov/govuk-prototype-kit/issues/917))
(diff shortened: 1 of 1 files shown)


matthew-shaw:
Also closes #2063

36degrees:
Hey Matt,

Thanks for raising this. I'll need to run through this with the rest of the team, but it makes sense to me on the face of it – attributes will always really end up as strings anyway.

A couple of minor things:

- it looks like the bump to the ini package is covered by #2063, and generally we'd avoid making unrelated changes like that and pull them out into a separate PR
- we only update the `package` folder when we're cutting a new release

I think that means we'd just take the changes in d37ec8e – if you're able to rebase to just include that single commit that'd be helpful.

Thanks again!

matthew-shaw:
Thanks @36degrees, fair points and I've reverted those commits, leaving just the test fixture updates.

Don't suppose there's a point fix release planned soon is there? This is kind of blocking me from doing a new release of `govuk-frontend-jinja` due to the failing tests. https://github.com/LandRegistry/govuk-frontend-jinja/pull/10 I _could_ code around it, but don't want to do that then strip it out again if there's a new `govuk-frontend` release right around the corner.

36degrees:
@matthew-shaw as those commits effectively 'cancel out' would you mind rebasing to remove those commits entirely, leaving just the one commit with the changes in d37ec8e?

Alternatively, if you can [allow edits from maintainers](https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests) on this PR I'd be happy to do this for you – I don't seem to have permission at the minute.

matthew-shaw:
@36degrees Ooops sorry yes you did say rebase originally, my bad. Have enabled edits by maintainers so feel free to rebase my commit 👍

36degrees:
Thanks – have accidentally managed to force push this to equivalent with master so it's closed itself… will get that fixed for you! 🤦🏻

matthew-shaw:
Haha no problem, this is exactly the kind of mistake I was worried about making myself 😆 Need me to re-raise a new PR?

36degrees:
Unfortunately, I think you might need to! I think because the PR has closed itself, I no longer have the permission to affect your fork, so I can't undo my mistake! 😬 

I'm still not entirely sure what actually happened there – sorry!

36degrees:
For my own future reference as much as anything, I _think_ this might have happened because @matthew-shaw's branch is called `master` and so when I force-pushed it pushed my local `master` branch rather than the tracking branch I was on. The same steps have worked previously for other PRs, but they tend to use a different branch name.

Maybe. Unless it was something else.

matthew-shaw:
@36degrees Recreated in #2071

PR #2071, opened 2020-12-15
Use non-bool type strings for attributes in test fixtures

matthew-shaw:
Follow on from #2064:

> When using boolean cast-able string values in test fixtures, these can be interpreted as actual boolean types by ports of govuk-frontend, such as the govuk-frontend-jinja Python port. For example, three tests fail due to casing differences between the expected string values of "true" and "false" and their boolean equivalents True and False: https://github.com/LandRegistry/govuk-frontend-jinja/pull/10/checks?check_run_id=1538392387#step:6:1365
>
> Other options:
>
> - I could code around this in my port by type-checking all param values and lower casing only bool cast-able values as strings. However, it seems better to improve the upstream test fixtures to use more explicit string type values instead.
> - I could blanket lower case all attribute values, but this could break things with unexpected behaviour and would deviate from how the upstream original behaves.
> 
> Changing these test fixture params to non-bool type strings will make maintaining ports of govuk-frontend in other frameworks and languages easier.

### DIFF
--- a/src/govuk/components/error-summary/error-summary.yaml
+++ b/src/govuk/components/error-summary/error-summary.yaml
@@ -137,8 +137,8 @@ examples:
   data:
     titleText: There is a problem
     attributes:
-      first-attribute: 'true'
-      second-attribute: 'false'
+      first-attribute: 'foo'
+      second-attribute: 'bar'
     errorList:
       -
         text: Invalid username or password
--- a/src/govuk/components/error-summary/template.test.js
+++ b/src/govuk/components/error-summary/template.test.js
@@ -115,8 +115,8 @@ describe('Error-summary', () => {
       const $ = render('error-summary', examples.attributes)
 
       const $component = $('.govuk-error-summary')
-      expect($component.attr('first-attribute')).toEqual('true')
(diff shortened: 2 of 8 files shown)


matthew-shaw:
Ping @36degrees 👍

36degrees (approved):
Looks good to me. Thanks for contributing and for your patience, @matthew-shaw 🙏🏻 

This just needs another review from someone else on the team and then we can get this merged for you.

vanitabarrett (approved):

matthew-shaw:
@36degrees Unfortunately when I recreated this PR from the original I also missed one of the changes required. There's a fixture with the same issue for the details component. I've created a new PR (#2079) to resolve that one. So sorry about all the hassle! 😞

36degrees:
No worries. Unfortunately we've already shipped the 3.10.2 release with the previous changes in it (we haven't announced it yet as we're having some issues getting the docs updated), and we're unlikely to be doing another release for a few weeks now.

matthew-shaw:
That's ok, I'll wait for the next fix release before doing a new release of the Jinja port. Cheers 🎄

PR #2082, opened 2020-12-21
Add GitHub Action to diff changes to dist

### LINKED ISSUE #2054: Automate posting a diff of changes to the dist folder when creating a new release
## What

Using GitHub Actions, automatically post a prettified git diff of the `dist` folder on pull requests that update the `dist` folder.

## Why

We've got in the habit of doing this as part of our release process, as it allows us to review the (minfied) changes to the `dist` folder, checking that they're what we expect.

For example, see [the diff for the dist folder for v3.10.0](https://github.com/alphagov/govuk-frontend/pull/2032#issuecomment-732056928).

However, this requires us to remember that we need to do it, and is a fairly manual process.

Developers

## Further detail

The current process uses the approach outlined here to prettify the minified JS / CSS before diffing it:
https://scripter.co/git-diff-minified-js-and-css/

1. Install `js-beautify`
2. Define two new [diff drivers](https://git-scm.com/docs/gitattributes#_defining_an_external_diff_driver) in the user's .gitconfig:

    ```
    [diff "minjs"]
        textconv = js-beautify
        cachetextconv = true
    [diff "mincss"]
        textconv = js-beautify --css
        cachetextconv = true
    ```

3. Configure git to use the diff driver for minified CSS and JS by adding to the `.gitattributes` file in the repo:

    ```
    *.min.js diff=minjs
    *.min.css diff=mincss
    ```

4. Run `git diff` against the `master` branch for the `dist` folder ([using -M to detect renames](https://git-scm.com/docs/git-diff#Documentation/git-diff.txt--Mltngt) with a similarity of 10% seems to work reliably so that git doesn't see the rename as a deletion and an addition)

    ```
    git diff -M1 master dist
    ```

5. Paste the output on the PR, using `diff` for syntax highlighting.

We should try and automate this, whilst ensuring that `git diff` continues to work for users without having to install or configure anything. For example, we should check whether committing the `.gitattributes` file to the repo would cause issues for users without a `minjs` or `mincss` diff driver.

## Done when

- [ ] Prettified diffs are automatically posted on PRs that update files in `diff`

36degrees:
When a pull request makes changes to files in the `dist` directory, automatically post a prettified git diff of the changes made, which we can use to ensure that the changes are what we expect (versus trying to review the minified change on GitHub).

This broadly follows the same steps that we've been doing manually, as outlined in https://scripter.co/git-diff-minified-js-and-css/:

1. Install js-beautify
2. Define two new [diff drivers][1] using `git config` for minified JS and CSS
3. Configure Git to use the diff driver for minified CSS and JS in .gitattributes - if the driver is not defined (for example, because it's not in a contributor's local Git config) it seems to be ignored.
4. Run git diff against the base branch for the PR for the dist folder, whilst [allowing for renames using `-M`][2] with a similarity index of 10% which has worked well previously.

The diff is then added to the PR as a comment using a GitHub Script action.

There is an [example comment below](https://github.com/alphagov/govuk-frontend/pull/2082#issuecomment-749046038) based on some test changes made in d1f449be460ea9c0ad4a5a3cc1cf7b10d423d1c4 – that commit has since been rebased out of this branch.

[1]: https://git-scm.com/docs/gitattributes#_defining_an_external_diff_driver
[2]: https://git-scm.com/docs/git-diff#Documentation/git-diff.txt--Mltngt

Closes #2054

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+*.min.js diff=minjs
+*.min.css diff=mincss
--- a/.github/workflows/diff-change-to-dist.yaml
+++ b/.github/workflows/diff-change-to-dist.yaml
@@ -8,6 +8,8 @@ jobs:
   generate-diff:
     name: Generate Diff
     runs-on: ubuntu-latest
+    # Abort if run from a fork, as token won't have write access to post comment
+    if: github.event.pull_request.head.repo.full_name == github.repository
     steps:
       - uses: actions/checkout@v2
         with:
--- a/.github/workflows/diff-change-to-dist.yaml
+++ b/.github/workflows/diff-change-to-dist.yaml
@@ -1,0 +1,47 @@
+name: Diff changes to dist
+
+on:
+  pull_request:
+    paths: ['dist/**']
+
+jobs:
+  generate-diff:
+    name: Generate Diff
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0 # Need to also checkout the base branch to compare
+      - uses: actions/setup-node@v1
+      - name: Set up diff drivers
+        run: |
+          npm install -g js-beautify
+          git config diff.minjs.textconv js-beautify
+          git config diff.mincss.textconv js-beautify
+      - name: Generate diff
+        id: diff
+        run: |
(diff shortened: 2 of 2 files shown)


vanitabarrett (approved):
Nice! 🎉

Issue #2074, opened 2020-12-16
Upgrade Heroku stack


36degrees:
## What

GOV.UK Frontend is using the Heroku-16 stack, which is deprecated. Update it to use the current Heroku stack (likely Heroku-20)

## Why

From Heroku's automated email to our Heroku account:

> From May 1st 2021, Heroku-16 will be end-of-life, and apps using it will no longer receive security updates, and be run at your own risk.
> 
> From June 1st 2021, builds will no longer be allowed for Heroku-16 apps.

## Further detail

- [Upgrading to the latest stack](https://devcenter.heroku.com/articles/upgrading-to-the-latest-stack)
- [Heroku-16 End-Of-Life FAQ](https://help.heroku.com/0S5P41DC/heroku-16-end-of-life-faq)

## Who needs to know about this

Developers

## Done when

- [x] Production 'Review app' is running on latest Heroku stack
- [x] New review apps spun up for pull requests run on the latest Heroku stack

36degrees:
As part of this I've also deleted the `govuk-frontend` 'production' app, which hadn't been deployed since Jun 25, 2018 and only showed an application error anyway.

PR #2086, opened 2020-12-23
Run review apps on Heroku 20 stack

36degrees:
The current Heroku-16 stack will be end-of-life (EOL) from May 1st 2021. From June 1st 2021, builds will no longer be allowed for Heroku-16 apps.

Updating to the latest Heroku stack now means we won’t get an email every time a review app is deployed.

Changing the stack definition in app.json only changes the stack for new review apps. Once this has been tested, we’ll also need to change the stack for the current running ‘production’ app using the Heroku CLI or the Heroku Dashboard.

https://devcenter.heroku.com/articles/upgrading-to-the-latest-stack

Part of #2074

### DIFF
--- a/app.json
+++ b/app.json
@@ -7,6 +7,7 @@
     "NPM_CONFIG_PRODUCTION": "false",
     "NODE_MODULES_CACHE": "false"
   },
+  "stack": "heroku-20",
   "formation": {
   },
   "addons": [


vanitabarrett (approved):
Looks good to me! 👍

PR #2089, opened 2021-01-04
[WIP] Make links and hover states clearer

36degrees:
(not actually Ollie's words – added by @christopherthomasdesign  later)

Some changes to link styles and hover states, put together for the working group review in February 2021.

There are 2 problems we're trying to solve here:

- link hover states are not clear (outlined in https://github.com/alphagov/govuk-frontend/issues/1417)
- link text could be easier to read – especially when used in long lists or at larger sizes (a problem raised by designers on the GOV.UK team)

The change adds some extra CSS to links to make underlines lighter and a bit further away from the text:

```
text-decoration-thickness: 1px;
text-underline-offset: .1em;
```

Browser support for this is about to improve massively when it's adopted by Chrome around April this year. We see it as an enhancement – for browsers that don't support these properties, the normal underline will still work.

On hover the underlines become thicker, to make the hover state clearer:

```
text-decoration-thickness: unquote("max(2px, .18em)");
```

Some of the rationale for this is included in a [comment around the time of the previous working group review](https://github.com/alphagov/govuk-frontend/issues/1417#issuecomment-590970918). The broad intention is that the thicker underline is a state change that can work across any combination of text colour and background colour, as long as the link has enough contrast in its default state.

### DIFF
--- a/app/views/examples/links/index.njk
+++ b/app/views/examples/links/index.njk
@@ -58,7 +58,39 @@
         </p>
       {% endfor %}
       </section>
+
     {% endfor %}
+    </div>
+  </div>
+
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-one-half">
+
+      <h2 class="govuk-heading-l govuk-!-margin-top-6">Multi-line links</h2>
+
+      <p class="govuk-heading-xl">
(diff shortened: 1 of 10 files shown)


36degrees:
Closing as this is now being addressed in #2183

PR #2092, opened 2021-01-07
Use tabular numbers for character count message

quis:
By default different numbers take up different amounts of horizontal space, for example ‘1’ is narrower than ‘0’, just like an ‘i’ is narrower than an ‘m’.

In the character count component the amount of space take up by the message changes as the user types. This causes the words after the count (‘characters remaining’) to jump around in a distracting way.

Tabular numbers take up a fixed amount of space, which means the words after the count do not change position as long as the count maintains the same number of digits (which will be true for most keystrokes).

# Before 

cc-lining

# After 

cc-tabular


36degrees (approved):
Thanks for this @quis – seems like a really sensible change and shows a great attention to detail 👍🏻  Would you mind adding an entry to the changelog?

Then we just need a second approval from someone else on the team and we can get this merged.

quis:
Thanks @36degrees, I’ve updated the changelog in 7e75d73e

vanitabarrett (approved):

PR #2093, opened 2021-01-11
Only output space after breadcrumbs class if there's an additional class

frankieroberto:
This change prevents the breadcrumb component from outputting a space after the html class if there are no custom classes to include. This has no impact on end users, but improves consistency with other components, and makes it easier to port the macros to other languages using the test fixtures. 

## Before

```html
<div class="govuk-breadcrumbs ">
```

## After

```html
<div class="govuk-breadcrumbs">
```


36degrees (approved):
Makes sense to me – thanks @frankieroberto. Would you mind adding this fix to the changelog?

frankieroberto:
@36degrees done. Let me know if you need a rebase and force push, or if you’re happy to squash-merge.

36degrees:
Thanks @frankieroberto. This just needs a second review and then we can get this merged for you.

vanitabarrett (approved):

PR #2114, opened 2021-01-22
Add button groups for use in cookie banner

### LINKED ISSUE #2107: Create additional link styles to align with buttons on cookie banner
## What
We now have three inline actions in the cookie banner. New styles may need to be introduced to keep all actions aligned, equally spaced and consistent.

The 'accept...' and 'reject...' actions are buttons but the 'View and set cookie preferences' action is a link and is visually different from the two buttons.
## Why
Users need to understand that the three actions are related and that their cookie preferences can be altered to selectively opt in and out of additional cookie types.
Designers, developers
## Further detail
- The focus styles for links are also different and should be looked at as part of this work.
- The primary button padding may also need to be reviewed if it is to have consistency with the currently proposed button for rejecting cookies.
## Done when

- [x] Decided on content for buttons and links in the cookie banner.
- [x] Decided on styling for the 'reject...' button within the banner.
- [x] Researched and documented use of button alongside links in services across gov.
- [x] Decided if this is an adjustment to the styles in the component or more universally to inline links.
- [x] Adjustments to the styles of the links and buttons to create consistency across desktop and mobile.

36degrees:
Add a new button groups object which can be used to group buttons and links together.

Within a button group:
 - links are styled to line up visually with the buttons, including being centre-aligned on mobile
 - spacing between the buttons and links is handled automatically, including when they wrap across multiple lines

The button group object will then be used to group the 'Accept cookies', 'Reject cookies' and 'View cookies' buttons and links in the cookie banner component.

The button group object uses flexbox and align-items to align the buttons and links along the 'cross axis' – centre-aligning on mobile and aligning rows along the baseline on tablet and above.

`align-items` is in theory supported in IE11 and above, but our use of autoprefixer means that IE10-compatible flexbox properties are also included, and in testing it works as well in IE10 as it does in IE11.

In browsers that do not support [flexbox][1] or [align-items][2] the spacing between buttons and links is different (as whitespace between inline blocks is not handled) and they are not aligned along the cross-axis, but the general layout is still applied (see screenshots for IE9, IE8 and iOS 7 for examples).

There's a corresponding PR to add guidance to the Design System here: https://github.com/alphagov/govuk-design-system/pull/1472

Closes #2107 

[1]: https://caniuse.com/flexbox
[2]: https://caniuse.com/mdn-css_properties_align-items_flex_context

### Firefox 84 (macOS)
Firefox 84

### Safari 14 (macOS)
Safari 14

### Chrome 88 (macOS)
Chrome 88

### Android 11
Android 11

### Edge 87
Edge

### IE11
IE11

### IE10
IE10

### IE10 ('mobile')
IE10 mobile

### IE9
IE9

### IE8
IE8

### iOS 7 (Simulated iPhone 5S)
iOS 7 (Simulated iPhone 5S)

### Android 4.4 (Simulated Galaxy S4)
Android 4 4 (Simulated Galaxy S4)

## Example of use in the cookie banner

https://user-images.githubusercontent.com/121939/105830147-0e66ae00-5fbd-11eb-8d5b-0b498e427d0a.mov

### DIFF
--- a/app/views/examples/button-groups/index.njk
+++ b/app/views/examples/button-groups/index.njk
@@ -1,0 +1,111 @@
+{% extends "layout.njk" %}
+
+{% from "back-link/macro.njk" import govukBackLink %}
+{% from "button/macro.njk" import govukButton %}
+
+{% block beforeContent %}
+  {{ govukBackLink({
+    "href": "/"
+  }) }}
+{% endblock %}
+
+{% block content %}
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+
+  <div class="govuk-button-group">
+    {{ govukButton({
+      "text": "Accept analytics cookies"
+    }) }}
+
+    {{ govukButton({
+      "text": "Reject analytics cookies"
+    }) }}
+
+    <a href="#" class="govuk-link govuk-link--align-button">View cookies</a>
+  </div>
+
+  <div class="govuk-button-group">
+    {{ govukButton({
+      "text": "Derbyn cwcis dadansoddeg"
+    }) }}
+
+    {{ govukButton({
+      "text": "Gwrthod cwcis dadansoddeg"
+    }) }}
+
+    <a href="#" class="govuk-link govuk-link--align-button">Gweld cwcis</a>
+  </div>
+
+  <div class="govuk-button-group">
+    {{ govukButton({
+      "text": "Accept really important analytics cookies"
+    }) }}
+
+    {{ govukButton({
+      "text": "Reject really important analytics cookies but not essential cookies"
+    }) }}
+
+    <a href="#" class="govuk-link govuk-link--align-button">View cookies and set preferences to do with whether we can set cookies</a>
+  </div>
+
+  <div class="govuk-button-group">
+    {{ govukButton({
+      "text": "Delete account",
+      "classes": "govuk-button--warning"
+    }) }}
+
+    <a href="#" class="govuk-link govuk-link--muted govuk-link--align-button">Cancel</a>
+    <a href="#" class="govuk-link govuk-link--muted govuk-link--align-button">Cancel and go back</a>
+  </div>
+
+  <div class="govuk-button-group">
+    <a href="#" class="govuk-link govuk-link--align-button">Buy it</a>
+
+    {{ govukButton({
+      "text": "Use it"
+    }) }}
+
+    {{ govukButton({
+      "text": "Break it"
+    }) }}
+
+    <a href="#" class="govuk-link govuk-link--align-button">Fix it</a>
+    <a href="#" class="govuk-link govuk-link--align-button">Trash it</a>
+    <a href="#" class="govuk-link govuk-link--align-button">Change it</a>
+
+    {{ govukButton({
+      "text": "Mail it"
+    }) }}
+
+    <a href="#" class="govuk-link govuk-link--align-button">Upgrade it</a>
+    
+    <a href="#" class="govuk-link govuk-link--align-button">Charge it</a>
+
+    {{ govukButton({
+      "text": "Point it"
+    }) }}
+
+    {{ govukButton({
+      "text": "Zoom it"
+    }) }}
+
+    {{ govukButton({
+      "text": "Press it"
+    }) }}
+  </div>
+
+  <div class="govuk-button-group">
+    {{ govukButton({
+      text: "Start now",
+      href: "#",
+      isStartButton: true
+    }) }}
+
+    <a href="#" class="govuk-link">Back</a>
+  </div>
+
+  </div>
+</div>
+{% endblock %}
--- a/src/govuk/objects/_button-group.scss
+++ b/src/govuk/objects/_button-group.scss
@@ -1,0 +1,75 @@
+@import "../base";
+
+@include govuk-exports("govuk/objects/button-group") {
+  // Button groups can be used to group buttons and links together as a group.
+  //
+  // Within a button group:
+  //
+  // - links are styled to line up visually with the buttons, including being
+  //   centre-aligned on mobile
+  // - spacing between the buttons and links is handled automatically, including
+  //   when they wrap across multiple lines
+  .govuk-button-group {
+    $horizontal-gap: govuk-spacing(3);
+    $vertical-gap: govuk-spacing(3);
+
+    // These need to be kept in sync with the button component's styles
+    $button-padding: govuk-spacing(2);
+    $button-shadow-size: $govuk-border-width-form-element;
+
+    $link-spacing: govuk-spacing(1);
+
+    @include govuk-responsive-margin(6, "bottom", $adjustment: $vertical-gap * -1);
+
+    // Flexbox is used to center-align links on mobile, align everything along
+    // the baseline on tablet and above, and to removes extra whitespace that
+    // we'd get between the buttons and links because they're inline-blocks.
+    //
+    // Ideally we'd use `gap` with flexbox rather than having to do it all with
+    // margins, but unfortunately the support isn't there (yet) and @supports
+    // doesn't play nicely with it
+    // (https://github.com/w3c/csswg-drafts/issues/3559)
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+
+    // Give links within the button group the same font-size and line-height
+    // as buttons.
+    //
+    // Because we want the focus state to be tight around the link text, we use
+    // margins where the buttons would use padding.
+    .govuk-link {
+      @include govuk-font($size: 19, $line-height: 19px);
+      display: inline-block;
+      margin-top: $link-spacing;
+      margin-bottom: $link-spacing + $vertical-gap;
+      text-align: center;
+    }
+
+    // Reduce the bottom margin to the size of the vertical gap (accommodating
+    // the button shadow) – the 'lost' margin is moved to the button-group.
+    .govuk-button {
+      margin-bottom: $vertical-gap + $button-shadow-size;
+    }
+
+    // On tablet and above, we also introduce a 'column gap' between the
+    // buttons and links in each row and left align links
+    @include govuk-media-query($from: tablet) {
+      // Cancel out the column gap for the last item in each row
+      margin-right: ($horizontal-gap * -1);
+
+      flex-direction: row;
+      flex-wrap: wrap;
+      align-items: baseline;
+
+      .govuk-button,
+      .govuk-link {
+        margin-right: $horizontal-gap;
+      }
+
+      .govuk-link {
+        text-align: left;
+      }
+    }
+  }
+}
--- a/src/govuk/objects/_button-group.scss
+++ b/src/govuk/objects/_button-group.scss
@@ -41,6 +41,9 @@
     .govuk-link {
       @include govuk-font($size: 19, $line-height: 19px);
       display: inline-block;
+      // Prevent links overflowing their container in IE10/11 because of bug
+      // with align-items: center
+      max-width: 100%;
       margin-top: $link-spacing;
       margin-bottom: $link-spacing + $vertical-gap;
       text-align: center;
--- a/app/views/full-page-examples/confirm-delete/index.njk
+++ b/app/views/full-page-examples/confirm-delete/index.njk
@@ -26,17 +26,17 @@
       }) }}
 
       <form method="post" novalidate>
+        <div class="govuk-button-group">
+          {{ govukButton({
+            text: "Delete account",
+            classes: "govuk-button--warning"
+          }) }}
 
-        {{ govukButton({
-          text: "Delete account",
-          classes: "govuk-button--warning govuk-!-margin-right-3"
-        }) }}
-
-        {{ govukButton({
-          text: "Cancel",
-          classes: "govuk-button--secondary"
-        }) }}
-
+          {{ govukButton({
+            text: "Cancel",
+            classes: "govuk-button--secondary"
+          }) }}
+        </div>
       </form>
 
     </div>
--- a/src/govuk/objects/_all.scss
+++ b/src/govuk/objects/_all.scss
@@ -1,3 +1,4 @@
+@import "button-group";
 @import "form-group";
 @import "grid";
 @import "main-wrapper";


src/govuk/objects/_button-group.scss
```diff
@@ -0,0 +1,89 @@
+@import "../base";
+
+@include govuk-exports("govuk/objects/form-group") {
```
vanitabarrett, line 3: @36degrees Guessing this should be `button-group` instead of `form-group`?

36degrees: Well, that explains why form-groups aren't being styled any more 🤦🏻‍♂️  Thanks!

christopherthomasdesign:
Looks good! 

The things that stood out to me are:
- left aligning of links on iOS 7 (Simulated iPhone 5S)
- higher baseline of links on IE9 & IE8 (meaning they don't align with the button text)

Those are pretty minor issues tbh and I don't think they'd impact usability much. Are they easily fixable or not?

36degrees:
> Those are pretty minor issues tbh and I don't think they'd impact usability much. Are they easily fixable or not?

## Alignment of links on iOS 7 (Simulated iPhone 5S)

In theory this could be an easy fix, but I don't know if we'd want to do it…

Flexbox is only supported on iOS 7 through 8.4 with a `-webkit-` prefix (https://caniuse.com/flexbox). We use autoprefixer to automatically apply prefixes based on [our browserslist](https://github.com/alphagov/govuk-frontend/blob/master/package.json#L85-L94) which currently includes `iOS >= 9` and any browser with > 0.1% marketshare.

If we changed the browserslist to include `iOS >= 7` instead autoprefixer would then include the required prefixes to make it display correctly in iOS 7 through 8.4. This would also affect (although hopefully in a positively way) breadcrumbs, start buttons, the footer, input prefixes and suffixes.

For reference, according to https://iosref.com/ios the last iPhone that did not support iOS 9 or above was the iPhone 4 which Apple stopped selling in September 2013.

Stats from visits from iOS to GOV.UK in December:

| Major Version | SUM of Users | SUM of Sessions | SUM of Sessions |
|---------------|--------------|-----------------|-----------------|
| 15 👀            | 8            | 19              | 0.00%           |
| 14            | 20,282,530   | 43,515,853      | 81.50%          |
| 13            | 2,633,560    | 5,639,239       | 10.56%          |
| 12            | 1,519,371    | 3,134,580       | 5.87%           |
| 11            | 113,025      | 285,591         | 0.53%           |
| 10            | 203,831      | 485,339         | 0.91%           |
| 9             | 120,034      | 260,733         | 0.49%           |
| 8             | 4,973        | 11,973          | 0.02%           |
| 7             | 4,176        | 8,780           | 0.02%           |
| 6             | 1128         | 2586            | 0.00%           |
| 5             | 978          | 2,134           | 0.00%           |
| 4             | 33           | 37              | 0.00%           |
| 3             | 93           | 241             | 0.00%           |
| 0             | 35,715       | 47,012          | 0.09%           |
| Grand Total   | 24,919,455   | 53,394,117      | 100.00%         |

I might create a separate issue to track this as it affects more than this component…

<details>
<summary>Diff of changes to package by changing <code>iOS >= 9</code> to <code>iOS >= 7</code></summary>

```diff
diff --git a/package.json b/package.json
index 8b75d4f3..6f7fca5a 100644
--- a/package.json
+++ b/package.json
@@ -90,7 +90,7 @@
     "last 2 Samsung versions",
     "Safari >= 9",
     "ie 8-11",
-    "iOS >= 9"
+    "iOS >= 7"
   ],
   "standard": {
     "globals": [
diff --git a/package/govuk/components/breadcrumbs/_index.scss b/package/govuk/components/breadcrumbs/_index.scss
index 3a4745b4..918e654a 100644
--- a/package/govuk/components/breadcrumbs/_index.scss
+++ b/package/govuk/components/breadcrumbs/_index.scss
@@ -129,6 +129,7 @@
 
       .govuk-breadcrumbs__list {
         display: -webkit-box;
+        display: -webkit-flex;
         display: -ms-flexbox;
         display: flex;
       }
diff --git a/package/govuk/components/button/_index.scss b/package/govuk/components/button/_index.scss
index 73cf764d..0ba74b24 100644
--- a/package/govuk/components/button/_index.scss
+++ b/package/govuk/components/button/_index.scss
@@ -239,6 +239,8 @@
 
     display: -webkit-inline-box;
 
+    display: -webkit-inline-flex;
+
     display: -ms-inline-flexbox;
 
     display: inline-flex;
@@ -246,6 +248,8 @@
 
     -webkit-box-pack: center;
 
+    -webkit-justify-content: center;
+
         -ms-flex-pack: center;
 
             justify-content: center;
@@ -258,10 +262,12 @@
       margin-left: govuk-spacing(2);
     }
     vertical-align: middle;
-    -ms-flex-negative: 0;
-        flex-shrink: 0;
-    -ms-flex-item-align: center;
-        align-self: center;
+    -webkit-flex-shrink: 0;
+        -ms-flex-negative: 0;
+            flex-shrink: 0;
+    -webkit-align-self: center;
+        -ms-flex-item-align: center;
+            align-self: center;
   }
 
   @if $govuk-use-legacy-font {
diff --git a/package/govuk/components/footer/_index.scss b/package/govuk/components/footer/_index.scss
index dbb314f7..b0c94772 100644
--- a/package/govuk/components/footer/_index.scss
+++ b/package/govuk/components/footer/_index.scss
@@ -79,16 +79,20 @@
 
   .govuk-footer__meta {
     display: -webkit-box;
+    display: -webkit-flex;
     display: -ms-flexbox;
     display: flex; // Support: Flexbox
     margin-right: -$govuk-gutter-half;
     margin-left: -$govuk-gutter-half;
-    -ms-flex-wrap: wrap;
-        flex-wrap: wrap; // Support: Flexbox
+    -webkit-flex-wrap: wrap;
+        -ms-flex-wrap: wrap;
+            flex-wrap: wrap; // Support: Flexbox
     -webkit-box-align: end;
+    -webkit-align-items: flex-end;
         -ms-flex-align: end;
             align-items: flex-end; // Support: Flexbox
     -webkit-box-pack: center;
+    -webkit-justify-content: center;
         -ms-flex-pack: center;
             justify-content: center; // Support: Flexbox
   }
@@ -101,11 +105,13 @@
 
   .govuk-footer__meta-item--grow {
     -webkit-box-flex: 1;
+    -webkit-flex: 1;
         -ms-flex: 1;
             flex: 1; // Support: Flexbox
     @include govuk-media-query ($until: tablet) {
-      -ms-flex-preferred-size: 320px;
-          flex-basis: 320px; // Support: Flexbox
+      -webkit-flex-basis: 320px;
+          -ms-flex-preferred-size: 320px;
+              flex-basis: 320px; // Support: Flexbox
     }
   }
 
@@ -165,12 +171,14 @@
 
   .govuk-footer__navigation {
     display: -webkit-box;
+    display: -webkit-flex;
     display: -ms-flexbox;
     display: flex; // Support: Flexbox
     margin-right: -$govuk-gutter-half;
     margin-left: -$govuk-gutter-half;
-    -ms-flex-wrap: wrap;
-        flex-wrap: wrap; // Support: Flexbox
+    -webkit-flex-wrap: wrap;
+        -ms-flex-wrap: wrap;
+            flex-wrap: wrap; // Support: Flexbox
   }
 
   .govuk-footer__section {
@@ -181,15 +189,18 @@
     vertical-align: top;
     // Ensure columns take up equal width (typically one-half:one-half)
     -webkit-box-flex: 1;
+    -webkit-flex-grow: 1;
         -ms-flex-positive: 1;
             flex-grow: 1; // Support: Flexbox
-    -ms-flex-negative: 1;
-        flex-shrink: 1; // Support: Flexbox
+    -webkit-flex-shrink: 1;
+        -ms-flex-negative: 1;
+            flex-shrink: 1; // Support: Flexbox
     @include govuk-media-query ($until: desktop) {
       // Make sure columns do not drop below 200px in width
       // Will typically result in wrapping, and end up in a single column on smaller screens.
-      -ms-flex-preferred-size: 200px;
-          flex-basis: 200px; // Support: Flexbox
+      -webkit-flex-basis: 200px;
+          -ms-flex-preferred-size: 200px;
+              flex-basis: 200px; // Support: Flexbox
     }
   }
 
@@ -199,6 +210,7 @@
     // To ensure the section is one of two, we can count backwards using `:nth-last-child(2)`.
     .govuk-footer__section:first-child:nth-last-child(2) {
       -webkit-box-flex: 2;
+      -webkit-flex-grow: 2;
           -ms-flex-positive: 2;
               flex-grow: 2; // Support: Flexbox
     }
diff --git a/package/govuk/components/input/_index.scss b/package/govuk/components/input/_index.scss
index 93fe3808..288b3102 100644
--- a/package/govuk/components/input/_index.scss
+++ b/package/govuk/components/input/_index.scss
@@ -95,11 +95,13 @@
 
   .govuk-input__wrapper {
     display: -webkit-box;
+    display: -webkit-flex;
     display: -ms-flexbox;
     display: flex;
 
     .govuk-input {
       -webkit-box-flex: 0;
+      -webkit-flex: 0 1 auto;
           -ms-flex: 0 1 auto;
               flex: 0 1 auto;
     }
@@ -150,6 +152,8 @@
 
     -webkit-box-flex: 0;
 
+    -webkit-flex: 0 0 auto;
+
         -ms-flex: 0 0 auto;
 
             flex: 0 0 auto;
diff --git a/package/govuk/objects/_button-group.scss b/package/govuk/objects/_button-group.scss
index 21c58767..1af99593 100644
--- a/package/govuk/objects/_button-group.scss
+++ b/package/govuk/objects/_button-group.scss
@@ -30,13 +30,16 @@
     // doesn't play nicely with it
     // (https://github.com/w3c/csswg-drafts/issues/3559)
     display: -webkit-box;
+    display: -webkit-flex;
     display: -ms-flexbox;
     display: flex;
     -webkit-box-orient: vertical;
     -webkit-box-direction: normal;
+    -webkit-flex-direction: column;
         -ms-flex-direction: column;
             flex-direction: column;
     -webkit-box-align: center;
+    -webkit-align-items: center;
         -ms-flex-align: center;
             align-items: center;
 
@@ -70,12 +73,16 @@
 
       -webkit-box-direction: normal;
 
+      -webkit-flex-direction: row;
+
           -ms-flex-direction: row;
 
               flex-direction: row;
-      -ms-flex-wrap: wrap;
-          flex-wrap: wrap;
+      -webkit-flex-wrap: wrap;
+          -ms-flex-wrap: wrap;
+              flex-wrap: wrap;
       -webkit-box-align: baseline;
+      -webkit-align-items: baseline;
           -ms-flex-align: baseline;
               align-items: baseline;
 
```
</details>

## Baseline of links in IE8 / IE9

The higher baseline of links on IE9 & IE8 is harder to fix, because those browsers just don't support flexbox at all. We _could_ set the spacing on links within a button group to be the same as buttons, which would make them line up neatly in IE8 and IE9, **but** that would mean the links would have to have the same 'total height' as buttons which they intentionally don't at the minute.

| With different spacing for links and buttons | With consistent spacing for links and buttons |
| -- | -- |
| Screenshot 2021-01-26 at 13 33 28 |  Screenshot 2021-01-26 at 13 34 19 |

app/views/examples/button-groups/index.njk
```diff
@@ -0,0 +1,111 @@
+{% extends "layout.njk" %}
```
hannalaakso, line 1: We could also update could `/full-page-examples/confirm-delete` to use the new button group.

src/govuk/objects/_button-group.scss
```diff
@@ -0,0 +1,76 @@
...
+      max-width: 100%;
+      margin-top: $link-spacing;
+      margin-bottom: $link-spacing + $vertical-gap;
+      text-align: center;
+    }
+
+    // Reduce the bottom margin to the size of the vertical gap (accommodating
+    // the button shadow) – the 'lost' margin is moved to the button-group.
+    .govuk-button {
+      margin-bottom: $vertical-gap + $button-shadow-size;
```
hannalaakso, line 53: It might be good to make the bottom margin of links and buttons inside the button group to be consistent with each other, at the moment there's a small difference between them.

We could also consider making the combined bottom margin of the button group to be consistent with `govuk-form-group` - but that's just a suggestion.

hannalaakso (commented):
@36degrees This is looking great 🎉  Just left a couple of small comments.

I'll take a look at your comment about iOS7/IE8-9 support.

src/govuk/objects/_button-group.scss
```diff
@@ -0,0 +1,76 @@
...
+
+    // Give links within the button group the same font-size and line-height
+    // as buttons.
+    //
+    // Because we want the focus state to be tight around the link text, we use
+    // margins where the buttons would use padding.
+    .govuk-link {
+      @include govuk-font($size: 19, $line-height: 19px);
+      display: inline-block;
+      max-width: 100%;
```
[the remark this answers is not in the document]
36degrees: IE10 and IE11 have [a bug where column flex items set to `align-items: center` overflow their container](https://github.com/philipwalton/flexbugs#2-column-flex-items-set-to-align-itemscenter-overflow-their-container) – see c98bd684c5a0a4d591c291c694ed9896307faad8. Setting `max-width: 100%` stops this from happening.

I split it out as a separate commit so that it would show in a git blame, but could add a comment if we think it warrants it.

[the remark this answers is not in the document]
hannalaakso: Ahh my bad, I missed that. A separate comment might be good so that down the line we could consider removing it when we drop support for certain browser combinations.

src/govuk/objects/_button-group.scss
36degrees: I think this is probably one for @CharlotteDowns to check, but I think the difference is intentional, based on how we want them to appear on mobile and when they wrap onto multiple lines. On tablet and above, everything should be aligned along the baseline so in theory the vertical margins of links are irrelevant unless they appear on a line by themselves.

The intent is that the combined bottom margin of the group matches the bottom margin on a button – so if there's only a single row of buttons then it should have the exact same spacing as if there was a single button not in a group.

app/views/examples/button-groups/index.njk
36degrees: Ah great call, thanks :+1:

36degrees: Done in f85eb47

src/govuk/objects/_button-group.scss
hannalaakso: The main reason I mentioned it was that on mobile the button group ends up having a slightly different bottom margin depending on whether the last item is a link or a button.

CharlotteDowns: Hey both, we want the link to look equally spaced between the 'visual' bottom of the 'reject' button and the bottom of the banner on mobile view. To follow the 5px baseline grid, I had originally tried to aim for 20px above and 20px below (as I think that was the padding on the banner container, this did mean some odd number margin values to account for the 3px box-shadow). I guess the aim is that the vertical spacing can all be divisible by 5px on all size viewports and that the text is on the same vertical line on sizes greater than mobile.

hannalaakso:
@36degrees Re: alignment of links on iOS 7, I would be tempted to leave it as it is since it's outside our support matrix and the user experience is still functional. If there was a blocker to user completing an action then we should consider fixing it. Interesting investigation into the potential fix though 👍  Similarly with IE8/9, the user experience is still functional.

hannalaakso (approved):
Thanks for addressing my comments @36degrees!

christopherthomasdesign:
Just to reply to [Ollie's comment](#issuecomment-767559883) – agree with what you proposed: 
* if alignment of links on iOS 7 is an issue that affects more than this component, consider it separately to this
* don't do anything about the baseline weirdness on IE8&9 – I don’t think it really changes the meaning of the links in any big way, just doesn’t look as neat

Issue #2099, opened 2020-12-03
Add Nunjucks macro template, tests and examples for cookie banner


hannalaakso:
## What
Create Nunjucks template / macro created based on the API agreed in https://github.com/alphagov/govuk-frontend/issues/2101. Add appropriate examples and tests for it.

If https://github.com/alphagov/govuk-frontend/issues/2103 is not finalised, we can add some draft Nunjucks macro options in this PR. These will be reviewed in https://github.com/alphagov/govuk-frontend/issues/2103

## Who needs to know about this
Developers

## Done when

- [x] Macro template created
- [x] Examples added
- [x] Tests added

PR #2119, opened 2021-01-27
Add Nunjucks macro template, tests and examples for cookie banner 

vanitabarrett:
Closes https://github.com/alphagov/govuk-frontend/issues/2099

Adds the macro, YAML examples and tests for the cookie banner component.

Depends on https://github.com/alphagov/govuk-frontend/pull/2114/files for the button grouping to work.

### DIFF
--- a/src/govuk/components/cookie-banner/cookie-banner.yaml
+++ b/src/govuk/components/cookie-banner/cookie-banner.yaml
@@ -100,8 +100,6 @@ examples:
           actions:
             - text: Hide this message
               type: button
-              attributes:
-                "data-hide-button": "true"
 
   - name: rejected confirmation banner
     data:
@@ -111,8 +109,36 @@ examples:
           actions:
             - text: Hide this message
               type: button
-              attributes:
-                "data-hide-button": "true"
+
+  - name: client-side implementation
+    data:
+      banners:
+        - headingText: Cookies on this service
+          text: We use cookies to help understand how users use our service.
+          actions:
+            - text: Accept analytics cookies
+              type: submit
+              name: cookies
+              value: accept
+            - text: Reject analytics cookies
+              type: submit
+              name: cookies
+              value: reject
+            - text: View cookie preferences
+              href: /cookie-preferences
+        - text: Your cookie preferences have been saved. You have accepted cookies.
+          role: alert
+          hidden: true
+          actions:
+            - text: Hide this message
+              type: button
+        - text: Your cookie preferences have been saved. You have rejected cookies.
+          role: alert
+          hidden: true
+          actions:
+            - text: Hide this message
+              type: button
+
 
 # Hidden examples are not shown in the review app, but are used for tests and HTML fixtures
   - name: heading html
--- a/src/govuk/components/cookie-banner/cookie-banner.yaml
+++ b/src/govuk/components/cookie-banner/cookie-banner.yaml
@@ -114,3 +114,107 @@ examples:
               attributes:
                 "data-hide-button": "true"
 
+# Hidden examples are not shown in the review app, but are used for tests and HTML fixtures
+  - name: heading html
+    hidden: true
+    data:
+      banners:
+        - headingHtml: Cookies on <span>my service</span>
+  - name: heading html as text
+    hidden: true
+    data:
+      banners:
+        - headingText: Cookies on <span>my service</span>
+  - name: html
+    hidden: true
+    data:
+      banners:
+        - html: <p class="govuk-body">We use cookies in <span>our service</span>.</p>
+  - name: classes
+    hidden: true
+    data:
+      banners:
+        - classes: "app-my-class"
+  - name: attributes
+    hidden: true
+    data:
+      banners:
+        - attributes:
+            "data-attribute": "my-value"
+  - name: custom aria label
+    hidden: true
+    data:
+      ariaLabel: "Cookies on GOV.UK"
+      banners:
+        - text: "We use cookies on GOV.UK"
+  - name: hidden
+    hidden: true
+    data:
+      banners:
+        - hidden: true
+  - name: hidden false
+    hidden: true
+    data:
+      banners:
+        - hidden: false
+  - name: default action
+    hidden: true
+    data:
+      banners:
+        - actions:
+          - text: This is a button
+  - name: link
+    hidden: true
+    data:
+      banners:
+        - actions:
+          - text: This is a link
+            href: /link
+  - name: link with button options
+    hidden: true
+    data:
+      banners:
+        - actions:
+          - text: This is a link
+            href: /link
+            value: cookies
+            name: link
+  - name: type
+    hidden: true
+    data:
+      banners:
+        - actions:
+          - text: Button
+            type: button
+  - name: button classes
+    hidden: true
+    data:
+      banners:
+        - actions:
+          - text: Button with custom classes
+            classes: "my-button-class app-button-class"
+  - name: button attributes
+    hidden: true
+    data:
+      banners:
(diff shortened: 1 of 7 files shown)


src/govuk/components/cookie-banner/template.njk
```diff
@@ -0,0 +1,57 @@
+{% from "../button/macro.njk" import govukButton %}
+
+<div class="govuk-cookie-banner" role="region" aria-label="{{ params.ariaLabel | default("Cookie banner") }}">
+  {% for banner in params.banners %}
+    {%- set classNames = "govuk-cookie-banner__container govuk-width-container" -%}
+    {% if banner.classes %}
+      {%- set classNames = classNames + " " + banner.classes -%}
+    {% endif -%}
+
+    <div class="{{classNames}}" {% if banner.role %}role="{{banner.role}}"{% endif %}
```
36degrees, line 10: Minor, but it'd be good to use Nunjucks' [whitespace control](https://mozilla.github.io/nunjucks/templating.html#whitespace-control) to avoid outputting extra spaces where they're not needed. They can be a bit confusing to get the hang of – give me a shout if you want to pair on it or anything.

src/govuk/components/cookie-banner/template.njk
```diff
@@ -0,0 +1,57 @@
...
+    {%- for attribute, value in banner.attributes %}{{attribute}}="{{value}}"{% endfor %}
```
36degrees, line 11: I think this needs a space to separate multiple attributes

```suggestion
    {%- for attribute, value in banner.attributes %} {{attribute}}="{{value}}"{% endfor %}
```

src/govuk/components/cookie-banner/template.njk
```diff
@@ -0,0 +1,57 @@
...
+        {%- elif banner.text -%}
+          <p class="govuk-body">{{ banner.text }}</p>
+        {%- endif -%}
+      </div>
+
+      {% if banner.actions %}
+      <div class="govuk-button-group">
+        {% for action in banner.actions %}
+            {% if action.href %}
+              {% set linkClasses = "govuk-link govuk-link--align-button" %}
```
36degrees, line 36: Don't need the `govuk-link--align-button` as the button group takes care of all that for us now.

```suggestion
              {% set linkClasses = "govuk-link" %}
```

src/govuk/components/cookie-banner/template.test.js
```diff
@@ -0,0 +1,219 @@
...
+    })
+  })
+
+  describe('action', () => {
+    it('renders as button by default', () => {
+      const $ = render('cookie-banner', examples['default action'])
+
+      const $actions = $('.govuk-cookie-banner .govuk-button')
+      expect($actions.get(0).tagName).toEqual('button')
+      expect($actions.text().trim()).toEqual('This is a button')
```
36degrees, line 143: There's a few tests in this describe block that have multiple expectations, and they don't always directly relate to the test description. 

Are some of these expectations repeated? Could any of them be removed, or move into their own tests?

src/govuk/components/cookie-banner/template.test.js
```diff
@@ -0,0 +1,219 @@
...
+ */
+/* eslint-env jest */
+
+const axe = require('../../../../lib/axe-helper')
+
+const { render, getExamples } = require('../../../../lib/jest-helpers')
+
+const examples = getExamples('cookie-banner')
+
+describe('Cookie Banner', () => {
```
36degrees, line 12: Really thorough tests – great stuff 👍🏻

src/govuk/components/cookie-banner/template.test.js
```diff
@@ -0,0 +1,219 @@
...
+
+  describe('confirmation banner', () => {
+    it('passes accessibility tests', async () => {
+      const $ = render('cookie-banner', examples['accepted confirmation banner'])
+
+      const results = await axe($.html())
+      expect(results).toHaveNoViolations()
+    })
+
+    it('set role attribute to alert', () => {
```
36degrees, line 106: This test sort of implies that there's some logic that means confirmation banners automatically get a role of `alert` – wonder if the distinction between question and confirmation banners is helpful as it's not really a distinction that exists in the code.

src/govuk/components/cookie-banner/template.test.js
```diff
@@ -0,0 +1,219 @@
...
+      const $component = $('.govuk-cookie-banner')
+      expect($component.attr('aria-label')).toEqual('Cookie banner')
+    })
+
+    it('renders a custom aria label', () => {
+      const $ = render('cookie-banner', examples['custom aria label'])
+
+      const $component = $('.govuk-cookie-banner')
+
+      // expect($component.attr('role')).toEqual('region')
```
36degrees, line 93: Can this be deleted?

src/govuk/components/cookie-banner/cookie-banner.yaml
```diff
@@ -0,0 +1,220 @@
...
+  - name: accepted confirmation banner
+    data:
+      banners:
+        - text: Your cookie preferences have been saved. You have accepted cookies.
+          role: alert
+          actions:
+            - text: Hide this message
+              type: button
+              attributes:
+                "data-hide-button": "true"
```
36degrees, line 104: Are these useful to have in the review app?

src/govuk/components/cookie-banner/cookie-banner.yaml
```diff
@@ -0,0 +1,220 @@
...
+        - name: classes
+          type: string
+          required: false
+          description: The additional classes that you want to add to the button or link.
+        - name: attributes
+          type: object
+          required: false
+          description: The additional attributes that you want to add to the button or link. For example, data attributes.
+
+examples:
```
36degrees, line 77: Can we add an example of the three banners together we can use in the review app? I think we might also want a test that ensures that when you pass multiple banners they are all actually included as I think all of the tests only pass one at the minute.

src/govuk/components/cookie-banner/template.test.js
vanitabarrett: I did wonder that, but the pattern of 'default' options that we normally use for our other components didn't really seem to fit here either. I've reworded the name of the test to `sets role attribute when role provided` which hopefully makes it clearer that it's not automatic?

src/govuk/components/cookie-banner/template.njk
```diff
@@ -0,0 +1,56 @@
+{% from "../button/macro.njk" import govukButton -%}
+
+<div class="govuk-cookie-banner" role="region" aria-label="{{ params.ariaLabel | default("Cookie banner") }}">
+  {%- for banner in params.banners %}
```
hannalaakso, line 4: I think we want to rename any `banner` parameters to `message`.

src/govuk/components/cookie-banner/template.test.js
```diff
@@ -0,0 +1,242 @@
...
+      const $ = render('cookie-banner', examples.attributes)
+
+      const $banner = $('.govuk-cookie-banner .govuk-cookie-banner__message')
+
+      expect($banner.attr('data-attribute')).toEqual('my-value')
+    })
+  })
+
+  describe('role and aria attributes', () => {
+    it('has role=region by default', () => {
```
36degrees, line 74: Really minor, but this perhaps implies you can override or remove the role, which you can't. Maybe just 

```suggestion
    it('has a role of region', () => {
```

36degrees (approved):
Looking really good – just one really minor test description thing which I noticed.

hannalaakso (commented):
🎉

PR #2122, opened 2021-01-27
Fix typos in Table Nunjucks

EoinShaughnessy:
Change "will be ignore" to "will be ignored."


36degrees (approved):
Great spot @EoinShaughnessy  👀  👍🏻

PR #2125, opened 2021-01-27
Add full page examples of cookie banners

### LINKED ISSUE #2097: Create some realistic example pages with cookie banner and page to present at the accessibility clinic
## What
We're going to [take the cookie banner to the accessibility clinic](https://github.com/alphagov/design-system-team-internal/issues/386) for a review. 

We should be able to amend the prototype we used for the accessibility research to include the new cookie banner and page and deploy it to a new herokup app.

## Why
We need some realistic examples of pages and a service flow to take to the accessibility clinic. 

Designers, developers

## Done when 
- [ ] We've identified how we're going to include the cookie banner and page in the prototype 
- [ ] We've added the cookie banner and page in the prototype
- [ ] We've deployed the prototype to a new heroku app

36degrees:
Add 'minimum viable' server-side and client-side implementations of a cookie banner 'journey' that we can use for acceptance and manual accessibility testing. These are based on the versions built by @vanitabarrett and used in initial prototypes in #2081.

As the cookie banner component will not include any JavaScript, the full page examples include (very basic) JavaScript that does roughly what we'd expect service teams to do (bar actually doing anything with the user's preference).

~~This is a draft as it's built on top of the `cookie-banner-macro` branch and will need rebasing once that's merged into the `feature/cookie-banner` branch.~~

Closes #2097

### DIFF
--- a/app/views/full-page-examples/cookie-banner-client-side/index.njk
+++ b/app/views/full-page-examples/cookie-banner-client-side/index.njk
@@ -1,0 +1,226 @@
+---
+name: Cookie banner (client side)
+scenario: >-
+  You need to make a choice about whether to accept cookies or not.
+
+notes: >-
+  For this example, when the user makes a choice to accept or reject cookies
+  their preference is handled entirely on the client side with no page
+  navigation, and JavaScript is used to swap out the cookie banner for the
+  'confirmation' banner.
+  
+  The choice to accept or reject cookies will not be remembered.
+
+  The content of the page is not important for this scenario.
+---
+
+{# This example is based of the live "Apply online for a UK passport" start page: https://www.gov.uk/apply-renew-passport #}
+{% extends "_generic.njk" %}
+
+{% from "breadcrumbs/macro.njk" import govukBreadcrumbs %}
+{% from "button/macro.njk" import govukButton %}
+{% from "cookie-banner/macro.njk" import govukCookieBanner %}
+{% from "tabs/macro.njk" import govukTabs %}
+
+{% set pageTitle = "Apply online for a UK passport" %}
+{% block pageTitle %}{{ pageTitle }} - GOV.UK{% endblock %}
+
+{% block bodyStart %}
+    {{ govukCookieBanner({
+    "banners": [
+        {
+            "headingText": "Cookies on this government service",
+            "text": "We use analytics cookies to help understand how users use our service.",
+            "actions": [
+                {
+                    "text": "Accept analytics cookies",
+                    "type": "button",
+                    "name": "cookies",
+                    "value": "accept",
+                    "classes": "js-cookies-button-accept"
+                },
+                {
+                    "text": "Reject analytics cookies",
+                    "type": "button",
+                    "name": "cookies",
+                    "value": "reject",
+                    "classes": "js-cookies-button-reject"
+                },
+                {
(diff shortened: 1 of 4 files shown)


app/views/full-page-examples/cookie-banner-server-side/index.njk
```diff
@@ -0,0 +1,193 @@
...
+              <a class="govuk-link" href="#/government/news/changes-to-passport-applications-for-british-nationals-living-abroad">Changes to passport applications for British nationals living abroad</a>
+            </li>
+          </ul>
+        </nav>
+      </aside>
+
+    </div>
+  </div>
+
+{# As the cookie banner component does not currently include JavaScript, we have
```
vanitabarrett, line 183: I think we also need to include the JavaScript for setting `tabindex` and `focus` so we have a server-side example of what we're recommending services implement for accessibility testing?

app/views/full-page-examples/cookie-banner-client-side/index.njk
```diff
@@ -0,0 +1,226 @@
...
+    "banners": [
+        {
+            "headingText": "Cookies on this government service",
+            "text": "We use analytics cookies to help understand how users use our service.",
+            "actions": [
+                {
+                    "text": "Accept analytics cookies",
+                    "type": "button",
+                    "name": "cookies",
+                    "value": "accept",
```
vanitabarrett, line 39: Do these buttons need a `name` and `value`?

app/views/full-page-examples/cookie-banner-client-side/index.njk
```diff
@@ -0,0 +1,226 @@
...
+                    "classes": "js-cookies-button-accept"
+                },
+                {
+                    "text": "Reject analytics cookies",
+                    "type": "button",
+                    "name": "cookies",
+                    "value": "reject",
+                    "classes": "js-cookies-button-reject"
```
vanitabarrett, line 47: I know this is just an internal example, so perhaps not important here, but do we normally hook in JS using classes or data-attributes? I know on GOV.UK we were quite keen on using data attributes to separate out styling from behaviour (although obviously the `js-` prefix helps too).

app/views/full-page-examples/cookie-banner-client-side/index.njk
vanitabarrett: Ah, I think this is because we're demonstrating how you might progressively enhance a server side banner? Makes sense 👍

app/views/full-page-examples/cookie-banner-client-side/index.njk
36degrees: We tend to use classes with a `govuk-js-` namespace, and only use data attributes for configuration rather than as hooks.

app/views/full-page-examples/cookie-banner-server-side/index.njk
36degrees: Good call – addressed in 3913131.

app/views/full-page-examples/cookie-banner-client-side/index.njk
36degrees: Not really, they were just copied and pasted from the server-side example to be honest. Removed in 823bff3.

vanitabarrett (approved):

PR #2130, opened 2021-02-01
Allow the whole cookie banner component to be hidden

vanitabarrett:
## What
Allows a user to pass `hidden: true` to the cookie banner component and for the whole component to be hidden. This is separate to the `hidden` flag which allows each individual cookie message to be hidden.

## Why
For a client-side implementation, when a user selects Accept/Reject and then hides the replacement message, the page is still left with a cookie banner region with nothing visible inside. In some AT we tested, this is listed as 'Cookie banner empty region', which at least tells the user there is no visible content, but isn't great. Ideally service teams would completely hide the whole banner when a user chooses to hide the replacement message. This gives them the ability to do that.

We also need to allow service teams to pass classes/attributes to the cookie banner so they can hook in and toggle the `hidden` attribute.


36degrees (approved):
LGTM 👍🏻

PR #2136, opened 2021-02-05
Tweaks to macro option documentation

36degrees:
Based on feedback from @EoinShaughnessy after reviewing the [Design System preview](https://deploy-preview-1484--govuk-design-system-preview.netlify.app/components/cookie-banner/):

- remove duplicated ‘required’ on `messages` and `actions.text` – ‘Required’ is automatically added as a prefix to the description in the Design System when `required` is `true`, so it does not need to be included in the description as well.
- mark `messages.text`, `messages.html` as required and remove ‘Required’ from their descriptions
- mark `false` as inline code in description for `hidden` and `messages.hidden`
- tweak the description for top-level `hidden` option to remove brackets

### DIFF
--- a/src/govuk/components/cookie-banner/cookie-banner.yaml
+++ b/src/govuk/components/cookie-banner/cookie-banner.yaml
@@ -6,7 +6,7 @@ params:
 - name: hidden
   type: boolean
   required: false
-  description: Defaults to `false`. If you set it to `true`, the whole cookie banner (including all messages within) is hidden. You can use `hidden` for client-side implementations where the cookie banner HTML is present, but hidden until it is shown using JavaScript.
+  description: Defaults to `false`. If you set it to true, the whole cookie banner is hidden, including all messages within it. You can use `hidden` for client-side implementations where the cookie banner HTML is present, but hidden until it is shown using JavaScript.
 - name: classes
   type: string
   required: false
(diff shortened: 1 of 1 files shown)


src/govuk/components/cookie-banner/cookie-banner.yaml
```diff
@@ -6,7 +6,7 @@ params:
 - name: hidden
   type: boolean
   required: false
-  description: Defaults to false. If you set it to `true`, the whole cookie banner (including all messages within) is hidden. You can use `hidden` for client-side implementations where the cookie banner HTML is present, but hidden until it is shown using JavaScript.
+  description: Defaults to `false`. If you set it to true, the whole cookie banner is hidden, including all messages within it. You can use `hidden` for client-side implementations where the cookie banner HTML is present, but hidden until it is shown using JavaScript.
```
EoinShaughnessy, line 9:
```suggestion
  description: Defaults to `false`. If you set this option to `true`, the whole cookie banner is hidden, including all messages within the banner. You can use `hidden` for client-side implementations where the cookie banner HTML is present, but hidden until the cookie banner is shown using JavaScript.
```

EoinShaughnessy: @vanitabarrett @36degrees Your thoughts, please? :)

vanitabarrett (approved):

PR #2137, opened 2021-02-05
Cookie banner accessibility tweaks

36degrees:


36degrees:
Cherry picked c0744a1 into #2144.

We decided not to change the focus-shifting approach (f42157b) as it made the confirmation message alert unreliable on iOS, and other users testing with JAWS 2021 are not seeing the same issues that we are.

PR #2139, opened 2021-02-08
[TEST] Explicitly set display: block

36degrees:


36degrees:
Behaves the same as the original implementation –  when clicking accept or reject cookies it reads out the link from the original cookie message instead.

> Enter. View cookie preferences link

PR #2141, opened 2021-02-08
Fix quotes in description for cookie banner role

36degrees:


EoinShaughnessy (approved):
@36degrees Thanks Ollie, looks good

PR #2157, opened 2021-02-24
Use pointer cursor for 'Menu' button in header

MalcolmVonMoJ:
Currently, the cursor is `default` when hovering over the menu button in the header.   An edge case perhaps because this will presumably mostly be for mobile.  But users on desktops with a narrow window, or when heavily zoomed in, will have an inconsistent experience.

This change adds one line of CSS to change the cursor to `pointer` when hovered over, as per any other link, including other links in the header.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Fixes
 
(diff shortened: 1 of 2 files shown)


36degrees (approved):
Makes sense to me – this is consistent with the cursor on both links and buttons elsewhere.

I've taken the liberty of squashing the commits down to 1 and adding a changelog entry – hope that's OK.

This just needs a review from a second member of the team and then we can get this merged – maybe @christopherthomasdesign?

christopherthomasdesign (approved):
👉 LGTM

PR #2163, opened 2021-03-08
Fix inaccessible text in components macros

EoinShaughnessy:
Style guide says to avoid 'e.g' - screen readers can misread it.


timpaul (approved):
LGTM :-)

PR #2164, opened 2021-03-08
Enable cookie banner to set link styled as a button

hannalaakso:
It should be possible to render a link styled as a button in the cookie banner. This is an edge case but could be useful for rendering the 'Hide' button if: 
- The cookie banner needs to work without JavaScript.
- The user has already accepted/rejected cookies. This means that when the user presses 'Hide' and the page reloads, the banner will be hidden.
 
This PR:
- Enables an action in the cookie banner to be a link styled as a button, with `role="button"` and `draggable="false"`, if both `href` and `type="button"` are provided. The resulting markup matches [Start buttons](https://design-system.service.gov.uk/components/button/#start-buttons).
- Add a test.
- Rename another example that renders a link but also sets additional button attributes to test that they aren't rendered, to distinguish it from this new example. 
 
We need to add some extra documentation to the macro options for this change. There are a few concepts mixed up in this change so I've started a [doc](https://docs.google.com/document/d/1BhH2VGCUkbltc5k9hAjSPBCxG7emdX12xxeNinJwOiU/edit) for @EoinShaughnessy to help us finalise the wording.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+## New features
+
+- [#2164: Enable cookie banner to set link styled as a button](https://github.com/alphagov/govuk-frontend/pull/2164)
+
 ### Fixes
 
 - [#2132 Improve vertical alignment of phase banner tag on mobile devices](https://github.com/alphagov/govuk-frontend/pull/2132) – thanks to [@matthewmascord](https://github.com/matthewmascord) for contributing this.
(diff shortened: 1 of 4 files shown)


EoinShaughnessy:
@hannalaakso Have commented on the [Google Doc for the content change](https://docs.google.com/document/d/1BhH2VGCUkbltc5k9hAjSPBCxG7emdX12xxeNinJwOiU/edit). :)

vanitabarrett (approved):

hannalaakso:
Merging this as @EoinShaughnessy has pulled the docs changes into https://github.com/alphagov/govuk-frontend/pull/2168, including updates to the changelog.

PR #2171, opened 2021-03-23
Fix padding on GOV.UK logo affecting hover and focus states

36degrees:
Address two minor issues with the hover and focus states for the GOV.UK logo (including product name, when present):

## 1. With product name

When the header includes a product name, the homepage link’s underline (on hover) and the focus state take up an additional 10px on the right hand side, because of the 10px right padding on `govuk-header__product-name`.

| Before | After |
| --- | --- |
| Screenshot 2021-03-23 at 17 20 09 | Screenshot 2021-03-23 at 17 20 19 |
| Screenshot 2021-03-23 at 17 20 13 | Screenshot 2021-03-23 at 17 20 22 |

This [padding is there to ensure there is a gap left between the product name and the menu button][1]. Remove it, and add a 10px margin to `govuk-header__link--homepage` instead which achieves the same result but without adding an additional 10px to the hover and focus states.

## 2. Without product name

When the header does not include a product name, the homepage link’s underline (on hover) and the focus state take up an additional 5px on the right hand side, because of a 5px margin on `govuk-header__logotype`.

| Before | After |
| --- | --- |
| Screenshot 2021-03-23 at 17 19 17 | Screenshot 2021-03-23 at 17 19 33 |
| Screenshot 2021-03-23 at 17 19 42 | Screenshot 2021-03-23 at 17 19 37 |

This margin has been present since the component was added to GOV.UK Frontend, and I believe it’s to ensure there’s a gap between GOV.UK and a product name. Remove this margin when the `__logotype` element is the last child of its descendent, so that the margin is only included when there is a product name to show.

[1]: https://github.com/alphagov/govuk-frontend/commit/df413cf6363b537fe37b1b0db5bc40b97424422f

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,8 +14,9 @@ This was added in [pull request #2164: Enable cookie banner to set link styled a
 
 ### Fixes
 
-- [#2132 Improve vertical alignment of phase banner tag on mobile devices](https://github.com/alphagov/govuk-frontend/pull/2132) – thanks to [@matthewmascord](https://github.com/matthewmascord) for contributing this.
-- [#2157 Use pointer cursor for 'Menu' button in header](https://github.com/alphagov/govuk-frontend/pull/2157) – thanks to [@MalcolmVonMoJ](https://github.com/MalcolmVonMoJ) for contributing this.
+- [#2132: Improve vertical alignment of phase banner tag on mobile devices](https://github.com/alphagov/govuk-frontend/pull/2132) – thanks to [@matthewmascord](https://github.com/matthewmascord) for contributing this.
+- [#2157: Use pointer cursor for 'Menu' button in header](https://github.com/alphagov/govuk-frontend/pull/2157) – thanks to [@MalcolmVonMoJ](https://github.com/MalcolmVonMoJ) for contributing this.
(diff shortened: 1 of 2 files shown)


CharlotteDowns (approved):
I really like this change!

The class `govuk-header__logo` also seems to have some right padding for menu items. Do we know why there was right padding on both the `govuk-header__logo` and `govuk-header__product-name` previously, is it for responsive views or another use case?

christopherthomasdesign (approved):
Checked out and tested locally, all looks good to me!

hannalaakso (changes requested):
This is a nicely documented tidy-up @36degrees 👍 

However, it seems that there's now some extra margin to the left of the product name when it's on a new line. The GOV.UK text also seems to have slightly moved. [Preview](https://govuk-frontend-pr-2171.herokuapp.com/components/header/with-product-name/preview).

36degrees:
@hannalaakso I've changed the approach for the second fix, by removing the margin if the `__logotype` element is the last child – in which case, there is no product name and so there's no need for the margin.

I think that addresses the issue you raised?

hannalaakso (approved):
@36degrees Looks great 🎉  I opened https://github.com/alphagov/govuk-frontend/issues/2181 for the other issues I spotted to do with the product name hover/focus states (nothing to do with the changes in this PR).

PR #2183, opened 2021-04-09
Updates to link styles and link hover states

36degrees:
This is a tidied up version of #2089.

We're making changes to link styles in order to improve their readability and to make the hover state clearer.

These changes will be behind a feature flag and 'opt-in' for their initial release. This is because:

- there's a a rendering bug in Chromium that affects links inside CSS columns (see 'known issues')
- users that have implemented their own link styles will likely need to make changes to their CSS in order for their links behave consistently.

In a future major version, we'll flip the feature flag so that this behaviour is opt-out, and then deprecate and remove the feature flag.

## Improve readability of links

Reduce the thickness of link underlines and offset them from the text in order to improve readability. This should help especially when multiple links are included in long lists (for example, search results pages) or at larger sizes.

| Before | After |
|----|----|
Screenshot before changes | Screenshot after changes |

Allow the thickness and offset to be overridden, so that anyone using GOV.UK Frontend outside of GOV.UK can opt out of these changes.

## Make link hover state clearer with thick underline

We have a long-standing issue (#1417) from an audit in May 2019 about the link hover state not being clear:

> The colour change when a user hovers over a link is not clear and this was especially difficult for low vision users to determine.
>
> Ensuring that the state change is distinctive would benefit low vision users in particular, while benefiting all mouse users in general.

This was raised as a usability issue rather than a WCAG failure.

We’ve been keeping an eye on browser support for the `text-decoration-thickness` property, and its introduction into Chrome 87 means that all evergreen browsers now support it. We can therefore use `text-decoration-thickess` as a way to differentiate the link on hover.

| Before | After |
|----|----|
| Screenshot before changes | Screenshot after changes |

https://user-images.githubusercontent.com/121939/114198950-c9abab80-994b-11eb-8f4a-7ecb8d3a8938.mov

Some of the rationale for this is included in [a comment around the time of the previous working group review][2]. The broad intention is that the thicker underline is a state change that can work across any combination of text colour and background colour, as long as the link has enough contrast in its default state.

We set the underline to a minimum width to 3px with a max to 0.12em. The aim is to make the state change clear whilst avoiding the underline crashing into the text below at larger sizes. A minimum 3px width ensures at least a 2px change from the default 1px thickness. It’s a bit close with 16px body-s text, but if you set the minimum width too low, the hover thickness at 19px stays too thin ([original comment][3]).

We're retaining the existing colour change on hover, as without the colour change there would be no state change in browsers that don't support `text-decoration-thickness`, including IE11. We can revisit this decision as and when our browser support requirements change ([original comment][4]).

[2]: #1417 (comment)
[3]: alphagov/govuk-design-system#1578 (comment)
[4]: alphagov/govuk-design-system#1578 (comment)

## Browser testing
- [x] IE 11
- [x] Edge
- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] Safari (iOS)
- [x] Chrome (iOS)
- [x] Chrome (Android)
- [x] Edge (Windows High Contrast Mode)
- [x] Firefox (Forced colours)
- [x] Firefox (Zoomed to 400%, zoom text only)

## Known issues

- In Safari, [links above 19px which do not have a colour change do not show the underline thickness change on hover](https://bugs.webkit.org/show_bug.cgi?id=224483). All of the link styles in GOV.UK Frontend have a colour change (even if imperceptible) to force the hover state to render correctly, but this could affect custom link styles.
- In Chrome, there's [a rendering bug that affects links inside CSS columns](https://bugs.chromium.org/p/chromium/issues/detail?id=1190987). This would affect the multi-column lists of links in the footer, except we've disabled the hover style for those links as we don't think they're really used outside of GOV.UK, and GOV.UK have their own implementation that does not use CSS columns. We think the risk here is relatively low, and will flag this issue in the release notes and advise anyone using CSS columns to opt out of the new hover state.

### DIFF
--- a/src/govuk/helpers/_links.scss
+++ b/src/govuk/helpers/_links.scss
@@ -47,7 +47,7 @@
 /// Provides the default unvisited, visited, hover and active states for links.
 ///
 /// If you use this mixin in a component you must also include the
-/// govuk-link-common mixin in order to get the focus state.
+/// govuk-link-common mixin in order to get the correct focus and hover states.
 ///
 /// @example scss
 ///   .govuk-component__link {
@@ -97,7 +97,7 @@
 /// Provides the error unvisited, visited, hover and active states for links.
 ///
 /// If you use this mixin in a component you must also include the
-/// govuk-link-common mixin in order to get the focus state.
+/// govuk-link-common mixin in order to get the correct focus and hover states.
 ///
 /// @example scss
 ///   .govuk-component__link {
@@ -144,7 +144,7 @@
 /// Provides the success unvisited, visited, hover and active states for links.
 ///
 /// If you use this mixin in a component you must also include the
-/// govuk-link-common mixin in order to get the focus state.
+/// govuk-link-common mixin in order to get the correct focus and hover states.
 ///
 /// @example scss
 ///   .govuk-component__link {
@@ -192,7 +192,7 @@
 /// regardless of visited state.
 ///
 /// If you use this mixin in a component you must also include the
-/// govuk-link-common mixin in order to get the focus state.
+/// govuk-link-common mixin in order to get the correct focus and hover states.
 ///
 /// @example scss
 ///   .govuk-component__link {
@@ -232,7 +232,7 @@
 /// navigation components, such as breadcrumbs or the back link.
 ///
 /// If you use this mixin in a component you must also include the
-/// govuk-link-common mixin in order to get the focus state.
+/// govuk-link-common mixin in order to get the correct focus and hover states.
 ///
 /// @example scss
 ///   .govuk-component__link {
@@ -266,7 +266,7 @@
 /// Overrides the colour of links to work on a dark background.
 ///
 /// If you use this mixin in a component you must also include the
-/// govuk-link-common mixin in order to get the focus state.
+/// govuk-link-common mixin in order to get the correct focus and hover states.
 ///
 /// @example scss
 ///   .govuk-component__link {
@@ -308,7 +308,7 @@
 /// that you’ve visited it before is not important.
 ///
 /// If you use this mixin in a component you must also include the
-/// govuk-link-common mixin in order to get the focus state.
+/// govuk-link-common mixin in order to get the correct focus and hover states.
 ///
 /// @example scss
 ///   .govuk-component__link {
--- a/src/govuk/helpers/_links.scss
+++ b/src/govuk/helpers/_links.scss
@@ -25,17 +25,19 @@
 @mixin govuk-link-decoration {
   text-decoration: underline;
 
-  @if ($govuk-link-underline-thickness) {
-    text-decoration-thickness: $govuk-link-underline-thickness;
-  }
+  @if ($govuk-new-link-styles) {
+    @if ($govuk-link-underline-thickness) {
+      text-decoration-thickness: $govuk-link-underline-thickness;
+    }
 
-  @if ($govuk-link-underline-offset) {
-    text-underline-offset: $govuk-link-underline-offset;
-  }
+    @if ($govuk-link-underline-offset) {
+      text-underline-offset: $govuk-link-underline-offset;
+    }
 
-  @if ($govuk-link-hover-underline-thickness) {
-    &:hover {
-      text-decoration-thickness: $govuk-link-hover-underline-thickness;
+    @if ($govuk-link-hover-underline-thickness) {
+      &:hover {
+        text-decoration-thickness: $govuk-link-hover-underline-thickness;
+      }
     }
   }
 }
--- a/src/govuk/helpers/_links.scss
+++ b/src/govuk/helpers/_links.scss
@@ -202,10 +202,13 @@
 
 @mixin govuk-link-style-muted {
   &:link,
-  &:visited,
+  &:visited {
+    color: $govuk-secondary-text-colour;
+  }
+
   &:hover,
   &:active {
-    color: $govuk-secondary-text-colour;
+    color: $govuk-text-colour;
   }
 
   // When focussed, the text colour needs to be darker to ensure that colour
@@ -242,8 +245,16 @@
 
 @mixin govuk-link-style-text {
   &:link,
-  &:visited,
-  &:hover,
+  &:visited {
+    @include govuk-text-colour;
+  }
+
+  // Force a colour change on hover to work around a bug in Safari
+  // https://bugs.webkit.org/show_bug.cgi?id=224483
+  &:hover {
+    color: rgba($govuk-text-colour, .99);
+  }
+
   &:active,
   &:focus {
     @include govuk-text-colour;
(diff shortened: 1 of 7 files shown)


src/govuk/helpers/_links.scss
```diff
@@ -10,18 +10,48 @@
...
+  );
+
+  text-decoration: solid underline $thickness;
+
+  @if ($govuk-link-underline-offset) {
+    text-underline-offset: $govuk-link-underline-offset;
+  }
+}
+
+@mixin govuk-link-hover-decoration {
```
36degrees, line 43: Needs Sass docs

src/govuk/helpers/_links.scss
```diff
@@ -236,6 +266,44 @@
...
+///
+/// @example scss
+///   .govuk-component__link {
+///     @include govuk-link-common;
+///     @include govuk-link-style-inverse;
+///   }
+///
+/// @access public
+
+@mixin govuk-link-style-inverse {
```
36degrees, line 284: I wonder if we should provide a class for this as well as a mixin?

src/govuk/settings/_measurements.scss
```diff
@@ -103,3 +103,27 @@ $govuk-focus-width: 3px !default;
 /// @access public
 
 $govuk-hover-width: 10px !default;
+
+/// Link underline thickness
```
36degrees, line 107: Is measurements the right place for these settings? I also considered creating a new `settings/_links.scss` file for them.

src/govuk/helpers/_links.scss
```diff
@@ -10,18 +10,48 @@
 
 @mixin govuk-link-common {
   @include govuk-typography-common;
+  @include govuk-link-decoration;
+
+  &:hover {
+    @include govuk-link-hover-decoration;
```
36degrees, line 16: This separation is just to support details, which may be changing soon anyway – the alternative would be to duplicate more of the CSS in the details component? But this might be a helpful separation for anyone building their own components where the hover interaction is a little more complex.

src/govuk/settings/_measurements.scss
36degrees: Worth thinking about the effect this has on the Sass docs as well – at the minute it'd be under [measurements](https://frontend.design-system.service.gov.uk/sass-api-reference/#measurements) which doesn't feel right. Equally if we did have a `settings/links` it feels a little odd that all of the link colour settings will still be under `settings/colours`.

src/govuk/components/accordion/_index.scss
```diff
@@ -147,7 +132,12 @@
     }
 
     .govuk-accordion__section-button:hover:not(:focus) {
+      color: $govuk-link-hover-colour;
       text-decoration: underline;
+
+      // This needs to come after the text-decoration property
```
vanitabarrett, line 138: Should we make it clear why?

src/govuk/components/footer/_index.scss
```diff
@@ -37,27 +31,20 @@
...
       &:link,
       &:visited {
-        color: $govuk-footer-link;
+        color: $govuk-footer-text;
       }
 
       &:hover,
       &:active {
-        color: $govuk-footer-link-hover;
+        color: #171819; // Only used with the legacy palette
```
vanitabarrett, line 44: Not a big issue, but not sure how I feel about hardcoding the value rather than declaring it above. I know we're doing it because it's only ever used once now, with the legacy palette, but I wonder if someone looking at this file might see the values defined at the top and assume those are the only legacy values and miss this one?

src/govuk/helpers/_links.scss
```diff
@@ -219,8 +259,16 @@
 @mixin govuk-link-style-text {
   // Override link colour to use text colour
   &:link,
-  &:visited,
-  &:hover,
+  &:visited {
+    @include govuk-text-colour;
+  }
+
+  // Force a colour change on hover to work around a bug in Safari
```
vanitabarrett, line 266: Nice 👌

src/govuk/helpers/_links.scss
```diff
@@ -236,6 +284,49 @@
...
+/// @example scss
+///   .govuk-component__link {
+///     @include govuk-link-common;
+///     @include govuk-link-style-inverse;
+///   }
+///
+/// @access public
+
+@mixin govuk-link-style-inverse {
+  // Override link colour to use text colour
```
vanitabarrett, line 303: I'm not sure what this comment means, as we're setting the link colour to white?

36degrees: Copy and paste fail! Great spot 👀

36degrees: I've removed this comment from this mixin and the text-colour mixin because I don't think it's particularly useful.

36degrees:
Need to revisit setting underline thickness in rem – it doesn't work so well in compatibility mode with GOV.UK Template when the root font size is 10px. Probably need to tie it to `$govuk-typography-use-rem`.

src/govuk/components/back-link/_index.scss
```diff
@@ -24,21 +24,6 @@
...
-    // focus styles has a bottom border.
-    &:focus {
-      text-decoration: none;
-
-      &:before {
-        border-color: $govuk-text-colour;
-      }
-    }
-  }
-
```
lfdebrux, line 41: Could this be done in a separate PR?

src/govuk/components/tabs/_index.scss
```diff
@@ -30,16 +30,11 @@
...
     @include govuk-link-style-default;
 
     display: inline-block;
     margin-bottom: govuk-spacing(2);
-
-    // Focus state for mobile and when JavaScript is disabled
-    // It is removed for JS-enabled desktop styles
-    &:focus {
-      @include govuk-focused-text;
-    }
```
lfdebrux, line 42: Could this be done in a separate PR, before the changes to the link styles? It seems to me that it could be, but I might be misunderstanding.

chao-xian:
@36degrees FYI we're currently applying these (by me, hacked in in a terrible way) to a WIP [prototype](https://explore-prot-new-link-d-dponqd.herokuapp.com/browse/births-deaths-marriages) we've got, and any feedback we find we'll let you know here. Likely next week.

36degrees:
This PR has been reworked so that it now contains only the changes to the core link styles.

Commits that refactored components to use the link mixins, removed redundant code removed or made other minor design changes have been split out into separate PRs that have already been merged:

- Simplify back link styles #2208
- Update tabs to use core link mixin #2209
- Update footer to use link mixins #2210
- Tidy up redundant code in accordion #2211
- Add underline to crown copyright link in footer #2212

The two new link mixins and modifier classes were also split out:

- Add no-underline mixin and modifier class #2214
- Add inverse link mixin and modifier class #2213

Finally, changes to components that rely on the changes in this PR (because they update the design, or because they rely on the new 'decoration' mixins introduced) have been split out into https://github.com/alphagov/govuk-frontend/pull/2215.

chao-xian:
I've spotted that the offset doesn't seem to be working in FF...? 
Screenshot 2021-05-01 at 01 04 31

vanitabarrett (approved):

36degrees:
> I've spotted that the offset doesn't seem to be working in FF...?

@chao-xian I think it is working, but the baseline for the offset is not consistent across browsers so it looks different in Firefox and Chrome. Is the hover state working correctly? And where is that screenshot from?

I think actually it's Chrome that's wrong (at least according to http://crbug.com/1172623) but we've had to find a value that looks alright in all browsers. As and when Chrome fix their baseline, we'll probably tweak the offset slightly to work better in all browsers.

hannalaakso (approved):
Awesome work @36degrees and @christopherthomasdesign, these new link styles are a big improvement in readability and usability 🎉

Issue #1863, opened 2020-07-15
Warning text icon clipped in Firefox v73+ when using high contrast mode or overriding colours


36degrees:
In the warning text icon, we use a 3px border (which is usually visually indistinct from the background) in order to better support users who use High Contrast Mode or override colours:

<img width="528" alt="Screenshot 2020-07-15 at 14 02 20" src="https://user-images.githubusercontent.com/121939/87548114-f4f6b800-c6a3-11ea-95b9-bd567fb76515.png">

https://github.com/alphagov/govuk-frontend/blob/36edb2f5eb9447a82e195bbbc0ee31c2f6ee7bff/src/govuk/components/warning-text/_index.scss#L29-L31

Firefox 73 [introduced a new 'readability backplate' feature](https://www.mozilla.org/en-US/firefox/73.0/releasenotes/):

> Many users with low vision rely on Windows' High Contrast Mode to make websites more readable. Traditionally, to increase the readability of text, Firefox has disabled background images when High Contrast Mode is enabled. With today’s release of Firefox 73, we introduce a “readability backplate” solution which places a block of background color between the text and background image. Now, websites in High Contrast Mode are more readable without disabling background images.

The new solid 'block' of background colour causes a clipping effect around the warning text icon in Firefox 73+.

<img width="528" alt="Screenshot 2020-07-15 at 14 03 14" src="https://user-images.githubusercontent.com/121939/87548147-fd4ef300-c6a3-11ea-97d4-5be069e2e1c4.png">

36degrees:
We could consider using an SVG for the icon.

36degrees:
After #2186 is merged, this should be fixed as and when / if Firefox implements the [`forced-color-adjust`](https://caniuse.com/mdn-css_properties_forced-color-adjust) property – for which [there are positive signals](https://github.com/mozilla/standards-positions/issues/463).

36degrees:
Firefox 89 [introduces support for the `forced-colors` CSS media](https://hacks.mozilla.org/2021/06/looking-fine-with-firefox-89/), but seemingly not `forced-color-adjust`.

Looks like https://bugzilla.mozilla.org/show_bug.cgi?id=1591210 is the bug to watch for that property being implemented.

36degrees:
Fixed in Firefox 113 Beta:

[...]

PR #2186, opened 2021-04-13
Fix display of warning text in Edge when Windows High Contrast Mode is enabled

36degrees:
When Windows High Contrast Mode (HCM) is enabled for Edge, the warning text icon is affected by the [readability backplate][1] which draws a box around the exclamation mark, obscuring parts of the circle.

Opt out of the readability backplate by setting `forced-color:adjust: none` and handle it ourselves, by overriding the colors to system colours in a media query which detects if forced colors are active.

Use `forced-color-adjust` and `forced-colors` over the MS specific `-ms-high-contrast-adjust: none;` and `-ms-high-contrast: active` to allow for future support from other browsers (e.g. Firefox, which allows users to set their own colour scheme for pages).

HCM with IE11 does not seem to be affected by the same issue – I don’t think it implements a readability backplate.

| Theme | Before | After |
| --- | --- | --- |
| High Contrast Black | Screenshot 2021-04-13 at 09 12 00|  Screenshot 2021-04-13 at 09 12 04 |
| High Contrast White | Screenshot 2021-04-13 at 09 12 12 | Screenshot 2021-04-13 at 09 12 17 |
| High Contrast # 1 |  Screenshot 2021-04-13 at 09 12 26 | Screenshot 2021-04-13 at 09 12 30 |
| High Contrast # 2 | Screenshot 2021-04-13 at 09 12 40 | Screenshot 2021-04-13 at 09 12 43 |

[1]: https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/6208eaffa0362ca107644c99b902c0b06bd5a0c8/Accessibility/HighContrast/explainer.md#ensuring-readability

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ This was added in [pull request #2164: Enable cookie banner to set link styled a
 - [#2132: Improve vertical alignment of phase banner tag on mobile devices](https://github.com/alphagov/govuk-frontend/pull/2132) – thanks to [@matthewmascord](https://github.com/matthewmascord) for contributing this.
 - [#2157: Use pointer cursor for 'Menu' button in header](https://github.com/alphagov/govuk-frontend/pull/2157) – thanks to [@MalcolmVonMoJ](https://github.com/MalcolmVonMoJ) for contributing this.
 - [#2171: Fix padding on GOV.UK logo affecting hover and focus states](https://github.com/alphagov/govuk-frontend/pull/2171)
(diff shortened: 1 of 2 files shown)


36degrees:
We have [the same issue in Firefox](https://github.com/alphagov/govuk-frontend/issues/1863), which this unfortunately doesn't fix, because Firefox doesn't support the [`forced-color-adjust`](https://caniuse.com/mdn-css_properties_forced-color-adjust) property (and doesn't have any other prefixed, non-standard or similar property we could use instead).

vanitabarrett (approved):
Looks good to me 👍 Is it worth expanding the Changelog entry, at the moment it sounds like it fixes the error in all browsers?

PR #2192, opened 2021-04-16
Add data-nosnippet to prevent cookie banner text appearing in Google Search snippets

vanitabarrett:
## What
Add data-nosnippet to the cookie banner component

## Why
Taken inspiration from https://github.com/alphagov/govuk_publishing_components/pull/1185
Apply to `data-nosnippet` attribute to the cookie banner component to stop its text content appearing in Google Search snippets.

We think we can make this change without adding a macro option to remove the attribute, as we can't think of a scenario where you would want the cookie banner text to appear in Google Search snippets.


36degrees (commented):
I think this makes sense, but as I sort of prompted this PR I think it'd be worth @hannalaakso reviewing in case she can think of any reason why this would ever be problematic.

I do wonder if it'd be useful to cross-reference the PR from GOV.UK where they added this in the commit message? Just as 'prior art'.

Great spot adding the missing README as well.

hannalaakso (approved):
This looks good to me 🎉 The NHS also seem to have done this in their cookie banner for nearly a year now which gives me confidence.

Issue #2194, opened 2021-04-16
Safari browser choose file text not aligned correctly when page maximised


moylesp:
## Description of the issue

When viewing option to upload a file using the Choose File button, the associated text of no file selected does not align correctly.
## Steps to reproduce the issue

open link in Safari browser https://design-system.service.gov.uk/components/file-upload/

![image](https://user-images.githubusercontent.com/82668346/115050734-f6c00700-9ed3-11eb-9bd6-a20485b9a660.png)

## Actual vs expected behaviour

I would expect that the text would align correctly.  This does happen if you reduce the size of the browser window and can display correctly  
![image](https://user-images.githubusercontent.com/82668346/115051028-43a3dd80-9ed4-11eb-9cb2-8db127880905.png)

## Environment (where applicable)

I am testing on the Land Registry web site, but I have included a link to the file uploader example https://design-system.service.gov.uk/components/file-upload/

- Operating system: Mac Catalina
- Browser: Safari
- Browser version:
![image](https://user-images.githubusercontent.com/82668346/115051409-b1e8a000-9ed4-11eb-9ef8-6938a4038df5.png)

- GOV.UK Frontend Version:

vanitabarrett:
Thanks for raising this @moylesp . I've had a look, and it looks like this is happening because the font-size of the file upload button isn't changing as it should do (and as it does in other browsers). On mobile, the button and the text are both 16px and therefore align correctly. However, on desktop they should both be 19px but in Safari this only applies to the text which pushes it out of alignment.

To me, this looks like a Safari bug as it doesn't seem possible to change the font-size of the file upload button. Strangely enough, this only seems to work if you also set the background color using `::-webkit-file-upload-button`. I've raised an issue with Safari, so we'll have to see what comes back from that.

PR #2201, opened 2021-04-22
Set `-webkit-appearance: button` on file upload so text is aligned in Safari

vanitabarrett:
Fixes https://github.com/alphagov/govuk-frontend/issues/2194

Screenshot 2021-04-23 at 10 36 29

## What
In Safari, the text of the file upload component was mis-aligned with the upload button on desktop, due to the font-size not being properly applied to the button. Safari say:

> We've had this behavior for a long time, since the native rounded button style does not support variable sizes. It works when you specify a "background-color", since that drops us out of the native appearance. A workaround, if you do not require a native appearance, would be to specify "-webkit-appearance: none".

## Setting -webkit-appearance: none
Setting `-webkit-appearance: none` fixes the alignment issue as the font-size can now be set correctly. However, setting appearance to none suggests the element has no "special" appearance - "No special styling is applied. This is the default." (https://developer.mozilla.org/en-US/docs/Web/CSS/appearance).

Screenshot 2021-04-23 at 10 35 16

## Setting -webkit-appearance: button
Setting appearance to `button` means that "the element is drawn like a button", which is more appropriate for our use case.

Screenshot 2021-04-23 at 10 48 23

However, the button goes grey with white text when pressed, which fails colour contrast recommendations (2.09:1 instead of 4.5:1). So we need to also set the text colour for the button.

Screenshot 2021-04-23 at 10 51 31

## Cross-browser testing
Because we're nesting `-webkit-appearance` within a `::-webkit-file-upload-button` selector, this doesn't have any effect in the majority of browsers. The only change I observed was in Safari (Mac). Full selection of screenshots [here](https://docs.google.com/presentation/d/1V1b0rwgjyTH6EwfkYn1NLwLzwnTFymcuAgsr0VJF73w/edit#slide=id.p)

- [x] IE11
- [x] Edge
- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] Safari (iOS)
- [x] Chrome (iOS)
- [x] Chrome (Android)
- [x] Windows High Contrast Mode (Edge)

### DIFF
--- a/src/govuk/components/file-upload/_index.scss
+++ b/src/govuk/components/file-upload/_index.scss
@@ -11,6 +11,17 @@
     padding-top: $component-padding;
     padding-bottom: $component-padding;
 
+    // The default file upload button in Safari does not
+    // support setting a custom font-size. Set `-webkit-appearance`
+    // to `button` to drop out of the native appearance so the
+    // font-size is set to 19px
+    // https://bugs.webkit.org/show_bug.cgi?id=224746
+    &::-webkit-file-upload-button {
+      -webkit-appearance: button;
+      color: inherit;
+      font: inherit;
+    }
+
     &:focus {
       // "Yank" the padding with negative margin to avoid a jump
       // when element is focused
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ This was added in [pull request #2164: Enable cookie banner to set link styled a
 - [#2171: Fix padding on GOV.UK logo affecting hover and focus states](https://github.com/alphagov/govuk-frontend/pull/2171)
 - [#2186: Fix display of warning text in Edge when Windows High Contrast Mode is enabled](https://github.com/alphagov/govuk-frontend/pull/2186)
 - [#2192: Add data-nosnippet to prevent cookie banner text appearing in Google Search snippets](https://github.com/alphagov/govuk-frontend/pull/2192)
+- [#2201: Set -webkit-appearance: button on file upload so text is aligned in Safari](https://github.com/alphagov/govuk-frontend/pull/2201)
 
 ## 3.11.0 (Feature release)
 


src/govuk/components/file-upload/_index.scss
```diff
@@ -11,6 +11,16 @@
...
     padding-bottom: $component-padding;
 
+    // The default file upload button in Safari does not
+    // support setting a custom font-size. Set `-webkit-appearance`
+    // to `button` to drop out of the native appearance so the
+    // font-size is set to 19px
+    // https://bugs.webkit.org/show_bug.cgi?id=224746
+    &::-webkit-file-upload-button {
+      -webkit-appearance: button;
+      color: $govuk-text-colour;
```
36degrees, line 21: Would inheriting from the parent work here? Just thinking that'd mean less things to keep in sync, and the print styles in the `govuk-text-colour` also get applied correctly.

```suggestion
      color: inherit;
```

src/govuk/components/file-upload/_index.scss
```diff
@@ -11,6 +11,16 @@
     padding-top: $component-padding;
     padding-bottom: $component-padding;
 
+    // The default file upload button in Safari does not
+    // support setting a custom font-size. Set `-webkit-appearance`
+    // to `button` to drop out of the native appearance so the
+    // font-size is set to 19px
+    // https://bugs.webkit.org/show_bug.cgi?id=224746
+    &::-webkit-file-upload-button {
```
36degrees, line 19: Looking at [what normalize.css does](https://github.com/necolas/normalize.css/blob/fc091cce1534909334c1911709a39c22d406977b/normalize.css#L303-L311) for file uploads, it looks like they set `font: inherit` as well, which might be worth doing here too?

36degrees (approved):
Great fix, and really thorough write up – thanks! Got a couple of suggestions, but also happy for this to merge as is.

PR #2205, opened 2021-04-26
Stop file upload component 'jumping' on focus

vanitabarrett:
## What
Stop file upload component 'jumping' on focus.

## Why
While working on https://github.com/alphagov/govuk-frontend/issues/2194 , we noticed that the file upload component 'jumps' slightly to the left on focus. Example (in Safari, but this affects other browsers too):

https://user-images.githubusercontent.com/29889908/116105923-b0279500-a6a9-11eb-8ad7-d6b041652d67.mov

This is related to some CSS and a comment in the code, which is applied on focus:

```scss
// Yank the padding with negative margin to avoid a jump
// when element is focused
margin-right: -$component-padding;
margin-left: -$component-padding;
padding-right: $component-padding;
padding-left: $component-padding;
```

We can understand why this code was added by observing the effect of removing it:

https://user-images.githubusercontent.com/29889908/116106552-1d3b2a80-a6aa-11eb-89ee-6b8d8f6f96b8.mov

When this CSS is removed, the focus state and the file upload button overlap on the left-hand side.

### So... we need to add some padding-left?
The component has bottom and top padding applied by default, but left and right padding is only added on focus (see the code above). What happens if we add padding by default instead of waiting for focus, to prevent anything moving around on focus?

Screenshot 2021-04-26 at 16 07 03
Screenshot 2021-04-26 at 16 07 09

This initially looks good on focus, but we can see that when the focus state is removed the file upload button is out of line with the text above it. This explains why we were initially adding a negative margin.

### Setting padding and negative margin by default
Instead of waiting until focus, we can add padding and a negative left margin by default:

https://user-images.githubusercontent.com/29889908/116108031-72c40700-a6ab-11eb-93a6-b30bfe11f035.mov

## Cross Browser testing
- [x] IE11
- [x] Edge
- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] Safari (iOS)
- [x] Chrome (iOS)
- [x] Chrome (Android)
- [x] Windows High Contrast Mode (Edge)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ This was added in [pull request #2164: Enable cookie banner to set link styled a
 - [#2186: Fix display of warning text in Edge when Windows High Contrast Mode is enabled](https://github.com/alphagov/govuk-frontend/pull/2186)
 - [#2192: Add data-nosnippet to prevent cookie banner text appearing in Google Search snippets](https://github.com/alphagov/govuk-frontend/pull/2192)
 - [#2201: Set -webkit-appearance: button on file upload so text is aligned in Safari](https://github.com/alphagov/govuk-frontend/pull/2201)
+- [#2205: Stop file upload component 'jumping' on focus](https://github.com/alphagov/govuk-frontend/pull/2205)
 
 ## 3.11.0 (Feature release)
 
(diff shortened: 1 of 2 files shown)


36degrees (approved):
Looks good – neat solution and great write up 👍🏻

PR #2208, opened 2021-04-29
Simplify back link styles

36degrees:
We used to use a bottom border to create the underline so that the ‘triangle’ and the link were both underlined.

This was changed in ad141f4 to make it consistent with breadcrumbs, which means the special link treatment isn’t required any more.

Remove it, in preparation for the new link styles which it will otherwise override.

Split out from #2183.

### DIFF
--- a/src/govuk/components/back-link/_index.scss
+++ b/src/govuk/components/back-link/_index.scss
@@ -24,21 +24,6 @@
     padding-left: 14px;
   }
 
-  // Only underline if the component is linkable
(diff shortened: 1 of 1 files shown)


lfdebrux (approved):
Thanks for splitting it out.

This looks good to me (tested focus state in Chrome, Firefox and Safari on macOS)!

hannalaakso (approved):
Looks good to me too 👍

PR #2210, opened 2021-04-29
Update footer to use link mixins

36degrees:
Simplify the footer link styles by using the `govuk-link-style-text` mixin, except when using the legacy palette where we want to maintain the 'old' footer link styles.

Split out from #2183.


hannalaakso (approved):
Nice tidy up 👍

PR #2213, opened 2021-04-30
Add inverse link mixin and modifier class

### LINKED ISSUE #1916: Add inverse mixin and modifier for links
Links on dark backgrounds (like blue 'heros') are generally white.

Because of the way that the different states interact with each other, and the importance of the ordering of the selectors in the CSS, it's relatively easy to end up with the focus state being incorrectly applied – for example, with focused visited links being white on yellow:

image (5)

We should consider adding an `govuk-link-style-inverse` mixin and an `govuk-link--inverse` modifier for links that sets the text colour to white and correctly applies the different states.

36degrees:
Inverse links are a fairly common thing for service teams to need to implement, and we often see issues with the focus state being implemented incorrectly.

The upcoming changes to link styles will add even more nuance, so providing a reusable mixin and modifier class will save users time and help to make it more likely that link styles are implemented well.

We can also use this new mixin when we update the header component.

Closes #1916 

Screenshot 2021-04-30 at 09 21 43

### DIFF
--- a/app/views/examples/links/index.njk
+++ b/app/views/examples/links/index.njk
@@ -14,7 +14,8 @@
   'Link': '',
   'Link with no visited state': 'govuk-link--no-visited-state',
   'Text link': 'govuk-link--text-colour',
-  'Muted link': 'govuk-link--muted'
+  'Muted link': 'govuk-link--muted',
+  'Inverse': 'govuk-link--inverse'
 } %}
 
 {% set states = {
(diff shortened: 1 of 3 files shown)


vanitabarrett:
@36degrees Change looks good, but I'm not sure https://github.com/alphagov/govuk-frontend/pull/2213/commits/6199cfd364965842a55656530a535064db14ddf0 belongs in this PR?

36degrees:
> @36degrees Change looks good, but I'm not sure [6199cfd](https://github.com/alphagov/govuk-frontend/commit/6199cfd364965842a55656530a535064db14ddf0) belongs in this PR?

Good shout – rebased to remove it.

vanitabarrett (approved):

PR #2215, opened 2021-04-30
Update components to use new link styles

36degrees:
Most components that use the link mixins will automatically benefit from the new link styles introduced in #2183, but some components need special treatment.

Update those components to use the new link styles.

## Accordion

Updated to use new link styles, including the colour change which was missing previously.

### Before

Accordion before

### After (new link styles disabled)

Accordion after off

### After (new link styles enabled)

Accordion after on

## Details

Updated to use new link styles

## Before

govuk-frontend-pr-2183 herokuapp com_components_details_preview (1)
govuk-frontend-pr-2183 herokuapp com_components_details_preview

## After

govuk-frontend-pr-2215 herokuapp com_components_details_preview
govuk-frontend-pr-2215 herokuapp com_components_details_preview (1)

## Footer

Remove the hover state from links in the multi-column lists due to a bug in Chromium browsers affecting links within columns.

## Header

Updated to use new link styles, with a consistent 3px underline on all links on hover. These changes are applied whether `$govuk-new-link-styles` is enabled or not.

### Before

Header before

### After

Header after

### Known issues

- In Firefox, when colours are overridden, the underline on the 'GOV.UK' logo in the header when hovered is partially obscured by the readability backplate, and is only visible below the crown.

## Skip link

Updated to use underline offset

### Before

govuk-frontend-pr-2183 herokuapp com_components_skip-link_with-focus_preview

### After

govuk-frontend-pr-2215 herokuapp com_components_skip-link_with-focus_preview

### DIFF
--- a/src/govuk/components/accordion/_index.scss
+++ b/src/govuk/components/accordion/_index.scss
@@ -1,8 +1,4 @@
 @include govuk-exports("govuk/component/accordion") {
-
-  $govuk-accordion-link-colour: $govuk-link-colour;
-  $govuk-accordion-link-hover-colour: govuk-colour("light-blue");
-
   .govuk-accordion {
     @include govuk-responsive-margin(6, "bottom");
   }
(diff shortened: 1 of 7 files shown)


vanitabarrett (approved):

hannalaakso (approved):
Really good work Ollie 🙌

PR #2219, opened 2021-05-07
Improve Sass docs for link helpers and settings

### LINKED ISSUE #2217: Revise Sass API docs for link helpers and settings
## What

Review and update the Sass API docs for the link helpers, ensuring that:

-  mixins within link sections have consistent names
-  mixins within link sections have examples where appropriate
-  references to mixins within descriptions are marked as inline code with backticks

Review the Sass API docs for the new link settings introduced in [#2183](https://github.com/alphagov/govuk-frontend/pull/2183).

## Why
To tell users about the new link settings and make the existing content easier to read.

@EoinShaughnessy, @36degrees 

## Done when

- [x] Developer drafts content additions
- [x] Tech writer reviews content additions
- [x] Content passes 2i

36degrees:
Improve the SassDoc comments for link settings and helpers, which are used to generate the [Sass API reference docs](https://preview-link-sass-docs--govuk-frontend-docs-preview.netlify.app/sass-api-reference/).

Closes #2217 

---

👉🏻 Preview what this would look like in the Frontend Docs for:

- [link settings](https://preview-link-sass-docs--govuk-frontend-docs-preview.netlify.app/sass-api-reference/#links) – everything up to the 'Measurements' h2
- [link helpers](https://preview-link-sass-docs--govuk-frontend-docs-preview.netlify.app/sass-api-reference/#helpers-links) – everything up to the 'Shapes' h2

### DIFF
--- a/src/govuk/helpers/_links.scss
+++ b/src/govuk/helpers/_links.scss
@@ -2,7 +2,7 @@
 /// @group helpers/links
 ////
 
-/// Common link mixin
+/// Common link styles
 ///
 /// Provides the typography and focus state, regardless of link style.
 ///
@@ -21,9 +21,10 @@
   }
 }
 
-/// Link decoration mixin
+/// Link decoration
 ///
-/// Provides the text decoration for links, including any underline offset
+/// Provides the text decoration for links, including thickness and underline
+/// offset. Use this mixin only if you cannot use the `govuk-link-common` mixin.
 ///
 /// @access public
 @mixin govuk-link-decoration {
@@ -40,10 +41,11 @@
   }
 }
 
-/// Link hover decoration mixin
+/// Link hover decoration
 ///
-/// Provides the text decoration for links in their hover state, intended to
-/// be used within a `:hover` pseudo-selector
+/// Provides the text decoration for links in their hover state, for you to use
+/// within a `:hover` pseudo-selector. Use this mixin only if you cannot use the
+/// `govuk-link-common` mixin.
 ///
 /// @access public
 
@@ -53,12 +55,12 @@
   }
 }
 
-/// Default link style mixin
+/// Default link styles
 ///
-/// Provides the default unvisited, visited, hover and active states for links.
+/// Makes links use the default unvisited, visited, hover and active colours.
 ///
-/// If you use this mixin in a component you must also include the
-/// govuk-link-common mixin in order to get the correct focus and hover states.
+/// If you use this mixin in a component, you must also include the
+/// `govuk-link-common` mixin to get the correct focus and hover states.
 ///
 /// @example scss
 ///   .govuk-component__link {
@@ -103,12 +105,13 @@
   }
 }
 
-/// Error link style mixin
+/// Error link styles
 ///
-/// Provides the error unvisited, visited, hover and active states for links.
+/// Makes links use the error colour. The link will darken if it's active or a
+/// user hovers their cursor over it.
 ///
-/// If you use this mixin in a component you must also include the
-/// govuk-link-common mixin in order to get the correct focus and hover states.
+/// If you use this mixin in a component, you must also include the
+/// `govuk-link-common` mixin to get the correct focus and hover states.
 ///
 /// @example scss
 ///   .govuk-component__link {
@@ -150,12 +153,13 @@
   }
 }
 
-/// Success link style mixin
+/// Success link styles
 ///
-/// Provides the success unvisited, visited, hover and active states for links.
+/// Makes links use the success colour. The link will darken if it's active or a
+/// user hovers their cursor over it.
 ///
-/// If you use this mixin in a component you must also include the
-/// govuk-link-common mixin in order to get the correct focus and hover states.
+/// If you use this mixin in a component, you must also include the
+/// `govuk-link-common` mixin to get the correct focus and hover states.
 ///
 /// @example scss
 ///   .govuk-component__link {
@@ -197,13 +201,13 @@
   }
 }
 
-/// Muted style link mixin
+/// Muted link styles
 ///
-/// Used for secondary links on a page - the link will appear in muted colours
-/// regardless of visited state.
+/// Makes links use the secondary text colour. The link will darken if it's
+/// active or a user hovers their cursor over it.
 ///
-/// If you use this mixin in a component you must also include the
-/// govuk-link-common mixin in order to get the correct focus and hover states.
+/// If you use this mixin in a component, you must also include the
+/// `govuk-link-common` mixin to get the correct focus and hover states.
 ///
 /// @example scss
 ///   .govuk-component__link {
@@ -240,13 +244,13 @@
   }
 }
 
-/// Text style link mixin
+/// Text link styles
 ///
-/// Overrides the colour of links to match the text colour. Generally used by
+/// Makes links use the primary text colour, in all states. Use this mixin for
 /// navigation components, such as breadcrumbs or the back link.
 ///
-/// If you use this mixin in a component you must also include the
-/// govuk-link-common mixin in order to get the correct focus and hover states.
+/// If you use this mixin in a component, you must also include the
+/// `govuk-link-common` mixin to get the correct focus and hover states.
 ///
 /// @example scss
 ///   .govuk-component__link {
@@ -283,12 +287,13 @@
   }
 }
 
-/// Inverse style link mixin
+/// Inverse link styles
 ///
-/// Overrides the colour of links to work on a dark background.
+/// Makes links white, in all states. Use this mixin if you're displaying links
+/// against a dark background.
 ///
-/// If you use this mixin in a component you must also include the
-/// govuk-link-common mixin in order to get the correct focus and hover states.
+/// If you use this mixin in a component, you must also include the
+/// `govuk-link-common` mixin to get the correct focus and hover states.
 ///
 /// @example scss
 ///   .govuk-component__link {
@@ -325,17 +330,17 @@
   }
 }
 
-/// No visited state link mixin
+/// Default link styles, without a visited state
 ///
-/// Used in cases where it is not helpful to distinguish between visited and
-/// non-visited links.
+/// Makes links use the default unvisited, hover and active colours, with no
+/// distinct visited state.
 ///
-/// For example, navigation links to pages with dynamic content like admin
-/// dashboards. The content on the page is changing all the time, so the fact
-/// that you’ve visited it before is not important.
+/// Use this mixin when it's not helpful to distinguish between visited and
+/// non-visited links. For example, when you link to pages with
+/// frequently-changing content, such as the dashboard for an admin interface.
 ///
-/// If you use this mixin in a component you must also include the
-/// govuk-link-common mixin in order to get the correct focus and hover states.
+/// If you use this mixin in a component, you must also include the
+/// `govuk-link-common` mixin to get the correct focus and hover states.
 ///
 /// @example scss
 ///   .govuk-component__link {
@@ -369,10 +374,18 @@
   }
 }
 
-/// No underline link mixin
+/// Remove underline from links
 ///
-/// Removes underlines from links (except on hover, or when the link is active).
-/// This does not work in IE8, because IE8 does not support `:not`.
+/// Remove underlines from links unless the link is active or a user hovers
+/// their cursor over it. This has no effect in Internet Explorer 8 (IE8),
+/// because IE8 does not support `:not`.
+///
+/// @example scss
+///   .govuk-component__link {
+///     @include govuk-link-common;
+///     @include govuk-link-style-default;
+///     @include govuk-link-style-no-underline;
+///   }
 ///
 /// @access public
 
@@ -382,10 +395,10 @@
   }
 }
 
-/// Print friendly link mixin
+/// Include link destination when printing the page
 ///
-/// When printing, append the the destination URL to the link text, as long
-/// as the URL starts with either `/`, `http://` or `https://`.
+/// If the user prints the page, add the destination URL after the link text, if
+/// the URL starts with `/`, `http://` or `https://`.
 ///
 /// @access public
 
--- a/src/govuk/settings/_links.scss
+++ b/src/govuk/settings/_links.scss
@@ -4,43 +4,57 @@
 
 /// Enable new link styles
 ///
+/// If enabled, the link styles will change. Underlines will:
+///
+/// - be consistently thinner and a bit further away from the link text
+/// - have a clearer hover state, where the underline gets thicker to make the
+///   link stand out to users
+///
+/// You should only enable the new link styles if both:
+///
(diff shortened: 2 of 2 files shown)


lfdebrux:
Changes look good to me, just one question: it says that for the style mixins you "must" use them with `govuk-link-common`, but what if you're using `govuk-link-decoration`?

36degrees:
> Changes look good to me, just one question: it says that for the style mixins you "must" use them with `govuk-link-common`, but what if you're using `govuk-link-decoration`?

Good question… there are exceptions but in the common case users should use that mixin if they can. Maybe it's as simple as changing must to should?

src/govuk/helpers/_links.scss
```diff
@@ -369,10 +375,18 @@
   }
 }
 
-/// No underline link mixin
+/// Remove underline from links
 ///
-/// Removes underlines from links (except on hover, or when the link is active).
-/// This does not work in IE8, because IE8 does not support `:not`.
+/// Underlines are still visible when the link is hovered over, or active. This
```
m-green, line 380:
```suggestion
/// Underlines are still visible when the link is hovered over or active. This
```
This is passive - can you use active voice? Probably something like: "Removes underlines from links, unless a link is active or a user is hovering over it."

src/govuk/helpers/_links.scss
```diff
@@ -283,12 +288,13 @@
   }
 }
 
-/// Inverse style link mixin
+/// Inverse link styles
 ///
-/// Overrides the colour of links to work on a dark background.
+/// Makes links white, in all states. Use for links displayed against a dark
```
m-green, line 293:
```suggestion
/// Makes links white, in all states. Use this mixin if you're displaying links against a dark
```

src/govuk/helpers/_links.scss
```diff
@@ -21,9 +21,11 @@
   }
 }
 
-/// Link decoration mixin
+/// Link decoration
 ///
-/// Provides the text decoration for links, including any underline offset
+/// Provides the text decoration for links, including thickness and underline
```
m-green, line 26: Just a general observation: do you need to have both the short summary and the description? It's particularly notable here, where 'Link decoration' on its own doesn't add much. For this and the other mixins, I think you could bring the summary and description together, to avoid duplication - and to avoid the short summary feeling a bit 'floaty'. 

(It's just an idea - I appreciate that this would lead to inconsistency between these mixins and the ones that haven't been updated.)

src/govuk/helpers/_links.scss
```diff
@@ -103,12 +106,13 @@
   }
 }
 
-/// Error link style mixin
+/// Error link styles
 ///
-/// Provides the error unvisited, visited, hover and active states for links.
+/// Makes links use the error colour, darkening when the link is hovered over or
```
m-green, line 111:
```suggestion
/// Makes links use the error colour. Links will darken when the link is hovered over or
```

This might be better as two sentences.

m-green: Could you make this more active too? For example: "A link will darken if it's active or a user is hovering over it."

src/govuk/helpers/_links.scss
```diff
@@ -150,12 +154,13 @@
   }
 }
 
-/// Success link style mixin
+/// Success link styles
 ///
-/// Provides the success unvisited, visited, hover and active states for links.
+/// Makes links use the success colour, darkening when the link is hovered over
```
m-green, line 159: Same comments as `govuk-link-style-error`

src/govuk/helpers/_links.scss
```diff
@@ -197,13 +202,13 @@
   }
 }
 
-/// Muted style link mixin
+/// Muted link styles
 ///
-/// Used for secondary links on a page - the link will appear in muted colours
-/// regardless of visited state.
+/// Makes links use the secondary text colour, switching to the primary text
```
m-green, line 207: Same comments as `govuk-link-style-error`.

src/govuk/helpers/_links.scss
```diff
@@ -240,13 +245,13 @@
   }
 }
 
-/// Text style link mixin
+/// Text link styles
 ///
-/// Overrides the colour of links to match the text colour. Generally used by
-/// navigation components, such as breadcrumbs or the back link.
+/// Makes links use the primary text colour, in all states. Use for navigation
```
m-green, line 250:
```suggestion
/// Makes links use the primary text colour, in all states. Use this mixin for navigation
```
To keep consistent with updates further up the docs, and to help it flow more nicely.

src/govuk/helpers/_links.scss
```diff
@@ -325,17 +331,17 @@
...
 ///
-/// Used in cases where it is not helpful to distinguish between visited and
-/// non-visited links.
+/// Makes links use the default unvisited, hover and active colours, with no
+/// distinct visited state.
 ///
-/// For example, navigation links to pages with dynamic content like admin
-/// dashboards. The content on the page is changing all the time, so the fact
-/// that you’ve visited it before is not important.
+/// Use in cases where it is not helpful to distinguish between visited and
```
m-green, line 339:
```suggestion
/// Use this mixin when it's not helpful to distinguish between visited and
```

src/govuk/helpers/_links.scss
```diff
@@ -325,17 +331,17 @@
...
+/// non-visited links. [Read more about links without visited states in the
```
m-green, line 340: There's not much 'more' to read on the Design System page unfortunately - there's only one extra contextualising example sentence. Could you add that example sentence here instead?

src/govuk/helpers/_links.scss
```diff
@@ -382,10 +396,10 @@
   }
 }
 
-/// Print friendly link mixin
+/// Include link destination when printing the page
 ///
-/// When printing, append the the destination URL to the link text, as long
-/// as the URL starts with either `/`, `http://` or `https://`.
+/// If the user prints the page, add the destination URL after the link text, as
```
m-green, line 401:
```suggestion
/// If the user prints the page, this mixin adds the destination URL after the link text, as
```

m-green: Also suggest 'if' rather than 'as long as' so it's more in GOV.UK style

src/govuk/helpers/_links.scss
```diff
@@ -382,10 +396,10 @@
...
+/// long as the URL starts with either `/`, `http://` or `https://`.
```
m-green, line 402:
```suggestion
/// long as the URL starts with `/`, `http://` or `https://`.
```
I think we only use 'either' with binary 'either/or' choices.

src/govuk/settings/_links.scss
```diff
@@ -4,43 +4,62 @@
 
 /// Enable new link styles
 ///
+/// If enabled, the link styles will change. Underlines will:
```
m-green, line 7: This uses a slightly different format to the other mixins ("If you enable this, a thing will happen" here vs "Does something to a thing" elsewhere). Best to make it consistent one way or the other.

src/govuk/settings/_links.scss
```diff
@@ -4,43 +4,62 @@
...
+///
+/// These changes are an opt-in feature because you may:
+///
+/// - need to make changes to your service to introduce these new link styles
+///   consistently
+/// - want to avoid opting in if you have links within a multiple-column layout
+///   in your CSS - there is [a Chromium bug that affects such
+///   links](https://github.com/alphagov/govuk-frontend/issues/2204)
+///
+/// These new link styles will be enabled by default in a future major version
```
m-green, line 21: It's probably best to avoid promising this - you never know what might happen. It's also probably not useful to a developer using the documentation right now (unless you're expecting developers to write their code to accommodate the potential change) - it's better to communicate this in a release note at the time.

src/govuk/settings/_links.scss
```diff
@@ -4,43 +4,62 @@
 
 /// Enable new link styles
 ///
+/// If enabled, the link styles will change. Underlines will:
+///
+/// - be consistently thinner and a bit further away from the link text
+/// - have a clearer hover state, where the underline gets thicker to make the
+///   link stand out to users
+///
+/// These changes are an opt-in feature because you may:
```
m-green, line 13: This feels a little like we're explaining the feature from our perspective rather than our users'. Can you make it more task-based?

For example, something like:

You should only enable the new link styles if both:

- you've made sure your whole service will use the new style consistently
- you do not have links in a multi-column CSS layout - there's [a Chromium bug that affects links]

src/govuk/settings/_links.scss
```diff
@@ -4,43 +4,62 @@
...
+/// These new link styles will be enabled by default in a future major version
+/// of GOV.UK Frontend.
+///
 /// @type Boolean
 /// @access public
 
 $govuk-new-link-styles: false !default;
 
-/// Link underline thickness
+/// Thickness of link underlines
```
m-green, line 29: See my comment about whether these summaries are needed up top but - would it be better if we described these consistently with a verb? For example the one above has 'Enable'. So could this - and similar ones - use 'Set'?

src/govuk/settings/_links.scss
```diff
@@ -4,43 +4,62 @@
...
+///
+/// The default is whichever is calculated to be thicker:
```
m-green, line 31: Bit of a GOV.UK style-related tweak needed here, because the bullets need to follow on from the lead-in sentence.

"The default will be either:
- 1px
- 0.0625rem, if it's thicker than 1px because the user has zoomed in their text"

src/govuk/settings/_links.scss
```diff
@@ -4,43 +4,62 @@
...
 ///
 /// @type Number
 /// @access public
 
 $govuk-link-underline-offset: .1em !default;
 
-/// Link hover state underline thickness
+/// Thickness of link underlines in hover state
+///
+/// The default is, for a given link, whichever is calculated to be thicker:
```
m-green, line 55: Suggest 'The default for each link is' rather than 'The default is, for a given link,".

This also needs a tweak to make the bullet lead-in work, as above.

Maybe: "The default for each link will be the thickest of the following:"

m-green (commented):
Good stuff @EoinShaughnessy! Mostly minor comments I hope. Happy to chat through if you'd like!

src/govuk/helpers/_links.scss
EoinShaughnessy: That's a really good idea. Ok if I add it to the list of improvements we plan to make in a future iteration?

lfdebrux:
> > Changes look good to me, just one question: it says that for the style mixins you "must" use them with `govuk-link-common`, but what if you're using `govuk-link-decoration`?
> 
> Good question… there are exceptions but in the common case users should use that mixin if they can. Maybe it's as simple as changing must to should?

Works for me.

src/govuk/settings/_links.scss
EoinShaughnessy: @m-green Yeah, we're communicating this in the upcoming release notes.

@36degrees I'll remove the text here about what we're doing in future, and I'll make a note to update these Sass docs when we're enabling link styles by default.

src/govuk/settings/_links.scss
EoinShaughnessy:
```suggestion
/// Makes underlines:
```

EoinShaughnessy: @m-green How's this ^ suggestion - is it a bit bare? Also, should we delete "be" from the first bullet so the sentence as a whole does not read awkwardly ("Makes underlines be consistently thinner...")?

m-green: 'Makes underlines' seems to work! But yep you'd need to delete 'be' from the bullet.

src/govuk/settings/_links.scss
EoinShaughnessy: @m-green I wondered about this too...

So, there are 2 files in this PR, each representing different sections of the Sass docs:

- src/govuk/helpers/_links.scss - this documents mixins
- src/govuk/settings/_links.scss - this documents variables

As I understand it, mixins make your code do something, for example make links use a certain colour.

Whereas variables are settings that you change to apply to all of GOV.UK Frontend. So, variables are almost more like switches or toggles. I think if you use a variable, you're effectively just setting something to 'on' or 'off,' like choosing to include or not include the new link styles.

((In the future,  as we revise this content further, we'll hopefully also do something to differentiate mixins from variables.)

I think the rationale here (correct me if I'm wrong, @36degrees) for not prefacing summaries with a verb is: users will know that they have to set/enable a variable. In the same way that you see the word 'Volume' above a stereo knob, but not instructional text on what to do.

It's true that the variable above this one has the summary "Enable new link styles." The reason we included the verb there is because just saying "New link styles" is the only variable we believe users are likely to customise.

If we wanted to be consistent with the verb-first approach of "Enable new link styles" throughout the other variables, we'd have content like:

> **$govuk-link-underline-thickness**
> Sets offset of link underlines from text baseline
> Set this variable to `false` to avoid setting an offset.

How does that compare with the verb-less version?

> **$govuk-link-underline-thickness**
> Offset of link underlines from text baseline
> Set this variable to `false` to avoid setting an offset.

EoinShaughnessy: If we wanted to be consistent with the verb-first approach of "Enable new link styles" throughout the other variables, we'd have content like:

> **$govuk-link-underline-thickness**
> Sets offset of link underlines from text baseline
> Set this variable to `false` to avoid setting an offset.

Is that potentially a bit confusing? Because we're saying 'Sets' (as in, "This is what the variable does") and then 'Set' (as in "This is what you, the user, should do").

src/govuk/helpers/_links.scss
m-green: Yep that's totally fine!

src/govuk/settings/_links.scss
m-green: Hi Eoin - following our chat, 'enable' makes sense but I'd try to make it active if possible - eg "If you enable this variable...". (Or even better "If you set this variable to `true`..." to make it super-explicit).

EoinShaughnessy: @36degrees Just talked to @m-green and we're going to revisit this issue another day. We're going to leave the wording as you originally suggested.

src/govuk/settings/_links.scss
EoinShaughnessy: @36degrees Just talked to @m-green and we're going to revisit this issue another day. For now, we'll leave the wording as you originally suggested.

EoinShaughnessy:
This content has passed pre-i, so I've now sent it for 2i.

jonathanglassman (approved):

EoinShaughnessy:
This content has passed 2i.

EoinShaughnessy:
Can we get a commit-squash, please? :)

jonathanglassman (approved):

EoinShaughnessy (approved):

PR #2228, opened 2021-05-14
Fix display of checkboxes in Internet Explorer 8

### LINKED ISSUE #2227: Checkboxes display incorrectly in IE8
## Description of the issue

The 'checkmark' is not hidden in IE8, appearing un-rotated to the left of the checkbox input.

Screenshot 2021-05-14 at 10 19 50

Look at the checkbox examples in Internet Explorer 8

The checkbox should look like a standard HTML checkbox.

Instead, the checkbox has an unexpected mark to the left of it.

- Operating system: Windows 7
- Browser: Internet Explorer
- Browser version: 8
- GOV.UK Frontend Version: v3.12.0

36degrees:
We use the `before` and `after` pseudo-elements on the label for checkboxes to provide stylised checkboxes in modern browsers.

Internet Explorer 8 (IE8) does not support some of the features we use to do this, so we ‘fall back’ to the native HTML inputs instead.

In 09ef79d2e4ed5ea824d1f09526f902a4c8edc643 we updated the before and after pseudo-elements in the checkboxes sass to use a single colon, which was newly enforced as part of a switch to stylelint.

IE8 [only supports the single-colon syntax][1], and it looks like we were actually relying on the double-colon syntax _not_ being interpreted by IE8 in order to prevent them being displayed in that browser.

As a result, changing to a single-colon means that IE8 now interprets the pseudo-selectors, and as a result ends up with a malformed checkmark visible next to the native HTML input.

Wrap the before and after pseduo-elements with the `@govuk-not-ie8 mixin` to exclude these rules from being included in the IE8 specific stylesheet.

## Before

Screenshot 2021-05-14 at 11 00 42

## After

Screenshot 2021-05-14 at 11 00 46

[1]: https://caniuse.com/css-gencontent

Fixes #2227

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixes
+
+- [#2228: Fix display of checkboxes in Internet Explorer 8](https://github.com/alphagov/govuk-frontend/pull/2228)
+
 ## 3.12.0 (Feature release)
 
 ### New features
--- a/src/govuk/components/checkboxes/_index.scss
+++ b/src/govuk/components/checkboxes/_index.scss
@@ -72,43 +72,45 @@
     touch-action: manipulation;
   }
 
-  // [ ] Check box
-  .govuk-checkboxes__label:before {
-    content: "";
-    box-sizing: border-box;
-    position: absolute;
-    top: 0;
-    left: 0;
-    width: $govuk-checkboxes-size;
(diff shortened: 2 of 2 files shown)


lfdebrux (approved):
Works for me with the review app and BrowserStack IE8. Also checked the checkboxes in Chrome on macOS to confirm there was no change there.

Issue #1758, opened 2020-03-04
Consider documenting fallback logo fix in next breaking release


hannalaakso:
We made a markup change to the header and footer in https://github.com/alphagov/govuk-frontend/pull/1724 to ensure that screen readers don't announce the fallback image. We could consider documenting the steps to make this change in people's projects in the release notes of the next breaking release.

36degrees:
This was included as a recommended change in the release notes for v4.0.

PR #2229, opened 2021-05-17
Change approach to fallback PNG in the header to fix blank data URI from triggering Content Security Policy error

### LINKED ISSUE #2203: xlink:href="data:," attribute on crown image is triggering Content Security Policy error
## Description of the issue
Previous releases of govuk-frontend included the crown logo using this HTML:

```
<image src="/assets/images/govuk-logotype-crown.png" xlink:href="" class="govuk-header__logotype-crown-fallback-image" width="36" height="32"></image>
```

A recent update (I'm not sure which, but I'm getting the issue in 3.11.0, having upgraded from 3.9.1) has inserted the content `data:,` in the `xlink:href` attribute:

```
<image src="/assets/images/govuk-logotype-crown.png" xlink:href="data:," display="none" class="govuk-header__logotype-crown-fallback-image" width="36" height="32"></image>
```

Our Content Security Policy for does not allow data: URL images, and it seems like this attribute is being interpreted (by Chrome and Safari at least) as an attempt to load a data: URL image.

I'm not sure if this is just the browsers being over-zealous, but I'm also not clear why the `xlink:href` attribute is needed.

**Edit:** I forgot I could just look at the source to see if there are comments about the `Xlink:href` attribute, and indeed there are: https://github.com/alphagov/govuk-frontend/blob/ffe72e5daba72901362c187934d1fae1d58f33e4/src/govuk/components/header/template.njk#L34

So I think it's there to prevent SVG-supporting versions of Internet Explorer from unnecessarily downloading the fallback PNG image?

1. Create a site using govuk-frontend 3.11.0
2. Implement a Content Security Policy on the site that does not allow data: URL images  (for example, `img-src 'self' www.google-analytics.com`)
3. Load the site in a browser that supports Content Security Policy

Expected: no console errors.
Actual:

> Refused to load the image 'data:,' because it violates the following Content Security Policy directive: "img-src 'self' www.google-analytics.com".

- Operating system: macOS Catalina 10.15.7
- Browser: Chrome
- Browser version: 90.0.4430.85 (Official Build) (x86_64)
- GOV.UK Frontend Version: 3.11.0

36degrees:
The current fallback uses an `<image>` element, which is interpreted differently depending on whether it’s parsed in the context of SVG or HTML.

We include an `xlink:href` attribute on the `<image>` tag to prevent versions of IE which support SVG from downloading the fallback image when they don't need to.

However, this approach has proved problematic – IE11 still requests the fallback PNG when printing, which can cause issues with sessions (#2045) and an attempted fix for that issue by using a blank data uri has now caused further issues with Content Security Policies (#2203).

Switch to the simpler approach of using conditional comments for the header fallback PNG, targeting Internet Explorer 8 specifically. We can then use a good old-fashioned `` tag.

This does mean that browsers that do not support SVG other than Internet Explorer 8 will not be able to see the crown emblem.

According to [Can I Use][1], the other main browser that does not support SVG that we may need to consider is Android 2.1 - 2.3.

Between 2020-05-14 and 2021-04-30 analytics for GOV.UK shows 6,792 (0.00041%) sessions from 4,343 (0.00119%) users using Android 2.x, out of 577,929,096 sessions from 77,144,107 users.

By comparison, in the same period we saw 36,075 sessions (0.00217%) from 28,963 users (0.00791%) using IE8.

It appears that Android 2.x browsers just omit the SVG entirely – there’s no ‘broken image’ displayed – these users just see ‘GOV.UK’ without the crown:

Screenshot of header component on a simulated Galaxy S2 without the crown emblem visible

Given the very low volume of traffic from these browsers, this seems like an acceptable thing to do.

We could have removed the fallback PNG entirely, however our browser support explicitly states support for Internet Explorer 8 (‘although components may not look perfect’) and we know that some service teams do have a disproprtionate number of IE8 users – which is unlikely to be the case for Android 2.

This approach also mirrors the change that GOV.UK Notify have made for their own SVG fallback images in https://github.com/alphagov/notifications-govuk-alerts/pull/39, where they came to the same conclusions.

This PR also tweaks the alignment of the crown with the text, which works independently of the change to the markup:

### Before
Before

### With the CSS change only
CSS change only

### With CSS and markup changes
H After

[1]:https://caniuse.com/svg

Fixes #2203

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 ### Fixes
 
 - [#2228: Fix display of checkboxes in Internet Explorer 8](https://github.com/alphagov/govuk-frontend/pull/2228)
+- [#2229: Change approach to fallback PNG in the header to fix blank data URI from triggering Content Security Policy errors](https://github.com/alphagov/govuk-frontend/pull/2229)
+- [#2229: Fix alignment of fallback PNG in the header](https://github.com/alphagov/govuk-frontend/pull/2229)
 
 ## 3.12.0 (Feature release)
 
--- a/src/govuk/components/header/_index.scss
+++ b/src/govuk/components/header/_index.scss
@@ -62,7 +62,7 @@
     width: 36px;
     height: 32px;
     border: 0;
-    vertical-align: middle;
+    vertical-align: bottom;
   }
 
   .govuk-header__product-name {
--- a/src/govuk/components/header/template.njk
+++ b/src/govuk/components/header/template.njk
@@ -4,6 +4,7 @@
     <div class="govuk-header__logo">
       <a href="{{ params.homepageUrl | default('/') }}" class="govuk-header__link govuk-header__link--homepage">
         <span class="govuk-header__logotype">
+          <!--[if gt IE 8]><!-->
           {#- We use an inline SVG for the crown so that we can cascade the
           currentColor into the crown whilst continuing to support older browsers
           which do not support external SVGs without a Javascript polyfill. This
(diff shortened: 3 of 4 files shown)


36degrees:
Opened as draft for now – hoping we can get some data on requests for the fallback PNG from the GOV.UK CDN logs just to check that there are no other browsers that don't support SVG we need to be considering that are not listed on Can I Use.

36degrees:
The CDN logs only shows requests for the fallback image from IE 10, IE11 and Spartan Edge – all of which support SVG and are making requests for the fallback image when they don't need to.

lfdebrux:
Looks good to me, tested using IE8,9 on BrowserStack and Chrome on macOS.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Fixes
 
 - [#2228: Fix display of checkboxes in Internet Explorer 8](https://github.com/alphagov/govuk-frontend/pull/2228)
+- [#2229: Change approach to fallback PNG in the header to fix blank data URI from triggering Content Security Policy error](https://github.com/alphagov/govuk-frontend/pull/2229)
```
lfdebrux, line 8: Should the tweak to positioning also be in the changelog?

lfdebrux (approved):

tombye:
I checked this with the https://github.com/alphagov/notifications-govuk-alerts pages and Content Security Policy (CSP) and seeing no CSP errors.

Note: we use the excellent https://github.com/LandRegistry/govuk-frontend-jinja instead of this (also excellent) code for templating so I made the same changes listed here to our local copy of that code to test it.

hannalaakso (approved):
This makes sense to me and it simplifies the header code too 👍 

We could wrap this change into https://github.com/alphagov/govuk-frontend/issues/1758 to tell users in 4.0 to make the corresponding change in their markup.

paulwaitehomeoffice:
Thanks so much all! Our logs will be so much cleaner.

Also IE 8: truly the gift that keeps on giving, however much you try to make it stop.

PR #2230, opened 2021-05-18
Add extra letter spacing modifier for inputs

36degrees:
This is a re-implementation of the changes in #1222.

---

Add an `--extra-letter-spacing` modifier for inputs which increases the spacing between letters and turns on [tabular numbers](https://www.fonts.com/content/learning/fontology/level-3/numbers/proportional-vs-tabular-figures).

The idea is that this modifier would be used for inputs asking for things like security codes, reference or phone numbers, where the user would be checking back one character at a time.

This was inspired by https://github.com/alphagov/notifications-admin/pull/1545

## Without modifier

localhost_3000_components_input_extra-tracking-reference_preview 1

## With modifier

localhost_3000_components_input_extra-tracking-reference_preview 2

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,9 +53,10 @@ The [Button](https://design-system.service.gov.uk/components/button/) Nunjucks m
 ```
 
 This was added in [pull request #3344: Adding optional id attribute to button component](https://github.com/alphagov/govuk-frontend/pull/3344). Thanks to [Tom Billington](https://github.com/TomBillingtonUK) for this contribution.
+
 ### Added a modifier for text input styles that accept sequences of digits
 
-A new `.govuk-input--extra-letter-spacing` class for [Text Input](https://design-system.service.gov.uk/components/text-input/) was added to increase readability of text inputs meant to receive sequences of digits (like security codes, references or phone numbers).
+We've added a new `.govuk-input--extra-letter-spacing` class for [Text Input](https://design-system.service.gov.uk/components/text-input/). This increases readability of text inputs that receive sequences of digits (like security codes, references or phone numbers).
 
 You can add it through the `classes` option when using Nunjucks, or directly in the `class` attribute of the `<input>` when using HTML.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,13 @@ The [Button](https://design-system.service.gov.uk/components/button/) Nunjucks m
 ```
 
 This was added in [pull request #3344: Adding optional id attribute to button component](https://github.com/alphagov/govuk-frontend/pull/3344). Thanks to [Tom Billington](https://github.com/TomBillingtonUK) for this contribution.
+### Added a modifier for text input styles that accept sequences of digits
+
+A new `.govuk-input--extra-letter-spacing` class for [Text Input](https://design-system.service.gov.uk/components/text-input/) was added to increase readability of text inputs meant to receive sequences of digits (like security codes, references or phone numbers).
+
+You can add it through the `classes` option when using Nunjucks, or directly in the `class` attribute of the `<input>` when using HTML.
+
+This was added in [pull request #2230: Add extra letter spacing modifier for inputs](https://github.com/alphagov/govuk-frontend/pull/2230)
 
 ### Deprecated features
 
--- a/src/govuk/components/input/_index.scss
+++ b/src/govuk/components/input/_index.scss
@@ -66,6 +66,11 @@
     }
   }
 
+  .govuk-input--extra-letter-spacing {
+    @include govuk-font(false, $tabular: true);
+    letter-spacing: .05em;
+  }
+
   // em measurements are based on the point size of the typeface
   // Extra space is added on the right hand side to allow for the Safari prefill icon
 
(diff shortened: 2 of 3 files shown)


frankieroberto:
@36degrees 🙌

frankieroberto:
Is this still planned to get merged at some point?

romaricpascal:
Hi @frankieroberto! Yup, the plan is to try and get this revived, reviewed by designers and add some guidance so we can get it in eventually 😄

romaricpascal:
Added a changelog entry for the new class. Necessary, but also a way to get this merged and shipped without waiting to have more complete guidance in the Design System docs, which can come at a later stage (and be tracked in its own issue). 

@36degrees @claireashworth Two questions:
1. Does the changelog read OK?
2. As I was writing it, occured to me that a better name for the modifier may be `--digit-sequence` (especially as it does more than adding letter spacing). Would it make sense to rename it?

frankieroberto:
> 2. As I was writing it, occured to me that a better name for the modifier may be `--digit-sequence` (especially as it does more than adding letter spacing). Would it make sense to rename it?

I initially agreed with this, but then I thought you might also use it when asking people to input an alphanumeric code (or purely alphabetic?) in which case it might be confusing? And perhaps the letter spacing is more significant and obvious than the tabular numerals?

That said, I like the idea of having a more semantic name rather than one that just describes the styles applied (which might change). `—code-sequence` maybe?

CHANGELOG.md
```diff
@@ -31,6 +31,14 @@ Only use disabled form controls if research shows it makes the user interface ea
 
 This was added in [pull request #3187: Add disabled styles for form controls](https://github.com/alphagov/govuk-frontend/pull/3187).
 
+### Added a modifier for text input styles that accept sequences of digits
+
+A new `.govuk-input--extra-letter-spacing` class for [Text Input](https://design-system.service.gov.uk/components/text-input/) was added to increase readability of text inputs meant to receive sequences of digits (like security codes, references or phone numbers).
```
claireashworth, line 36:
```suggestion
We've added a new `.govuk-input--extra-letter-spacing` class for [Text Input](https://design-system.service.gov.uk/components/text-input/). This increases readability of text inputs that receive sequences of digits (like security codes, references or phone numbers).
```

claireashworth (approved):

romaricpascal:
> I initially agreed with this, but then I thought you might also use it when asking people to input an alphanumeric code (or purely alphabetic?) in which case it might be confusing? And perhaps the letter spacing is more significant and obvious than the tabular numerals?
> 
> That said, I like the idea of having a more semantic name rather than one that just describes the styles applied (which might change). `--code-sequence` maybe?

Yeah, `digit` is not really accurate there, that's true. Would `code` encompass phone numbers as well? Otherwise, maybe we could refer to "characters" (would it be understood as "non-digits" though?): `--character-sequence` or `--character-by-character`. Or maybe just `--sequence` without mentioning which kind of sequence it is: it's not really important and  you're typing "characters" anyways, after all.

querkmachine:
Of those options, I'd go for `--character-sequence`. 

Hopefully, the folks using this may already be familiar with the idea that a character is just any one glyph—be it letter, digit, punctuation or whitespace—from the character count component. `--sequence` alone feels like it might be too vague?

frankieroberto:
My thinking behind `--code-sequence` was to describe where this could be used - ie when the user is entering a code rather than a number or word. `--character-sequence` feels quite technical to me, and perhaps too vague (isn't everything a character sequence, including words?)

Language is hard though! Maybe `--extra-letter-spacing` was the clearest after all... 😂

querkmachine:
I'm more inclined towards naming things after the purpose they serve, rather than what they are, but that isn't a hard and fast rule by any means.

However, as this style change is not _only_ applying additional letter-spacing, but is also explicitly enabling tabular number formatting, I think it should be named something akin to `--code-sequence` in favour of `--extra-letter-spacing`. The former feels more accurate and more future-proof, in my mind.

calvin-lau-sig7:
Suggestions:
If it's fixed spacing (monospacing, fixed-character-spacing)
It it's just a little extra space being added in there (extra-spacing, distinct-spacing, separated-spacing)

querkmachine:
> Suggestions: 
> If it's fixed spacing (monospacing, fixed-character-spacing) 
> It it's just a little extra space being added in there (extra-spacing, distinct-spacing, separated-spacing)

I think that's part of the contention, @calvin-lau-sig7 — it's both!

owenatgov:
Jumping in with the opinion that we should keep the name as simple as possible. I think `--extra-letter-spacing` is a fine choice. Says what it does in the broad sense. I also like `--monospacing` or `--monospaced` as a catch-all.

At the end of the day though there is a ceiling to what value we can get out of picking the perfect word or phrase to explain what it's doing. I don't think we need to split hairs on precision of what we choose (appreciate I just contributed to this!). What it is now seems to do the job, in my opinion.

romaricpascal:
Let's keep it `--extra-letter-spacing` then 😄

querkmachine (approved):

frankieroberto:
@querkmachine woohoo! 🙌 

Can you add this modifier to [confirm a phone number inputs](https://github.com/alphagov/govuk-design-system/blob/main/src/patterns/confirm-a-phone-number/default/index.njk#L21)?

querkmachine:
@frankieroberto I think this might warrant a bit more documentation outside of updating certain examples, so I've opened an issue for it here: https://github.com/alphagov/govuk-design-system/issues/2665

PR #2237, opened 2021-05-27
Fix GOV.UK logo disappearing on light background in Windows High Contrast Mode

36degrees:
In 6f90027 I updated the logo in the header to use `forced-color-adjust: none;` in order to fix an issue where the readability backplate obscured the underline on the link.

However, I neglected to then handle the color adjustment correctly – the logo is now always white in high contrast mode, regardless of the user’s colour preferences. This means that the logo can be invisible or hard to perceive when the user’s colour scheme has a lighter background.

To fix this, set the colour of the logo to the system color `linktext` when forced colors are active, which means it will appear in the user’s preferred colour for unvisited links, matching the appearance of the other links in the header.

This only affects Chrome and Edge (and presumably Opera as that's also built on Chromium, but that's not available to test with in Assistiv Labs) – as Firefox doesn't support `forced-color-adjust`.

| Browser / Theme | Before | After
| -- | -- | -- |
| Chrome / #1 | Chrome Before 1 | Chrome After 1 |
| Chrome / #2 | Chrome Before 2 | Chrome After 2 |
| Chrome / Dark | Chrome Before Dark | Chrome After Dark |
| Chrome / White | Chrome Before White | Chrome After White |
| Edge / #1 | Edge Before 1 | Edge After 1 |
| Edge / #2 | Edge Before 2 | Edge After 2 |
| Edge / Dark | Edge Before Dark | Edge After Dark |
| Edge / White | Edge Before White | Edge After White |

### DIFF
--- a/src/govuk/components/header/_index.scss
+++ b/src/govuk/components/header/_index.scss
@@ -39,9 +39,12 @@
     // Add a gap between logo and any product name
     margin-right: govuk-spacing(1);
 
-    // Prevent readability backplate from obscuring underline in Windows
-    // High Contrast Mode
-    forced-color-adjust: none;
+    // Prevent readability backplate from obscuring underline in Windows High
+    // Contrast Mode
+    @media (forced-colors: active) {
+      forced-color-adjust: none;
+      color: linktext;
+    }
 
     // But remove it if there's nothing after the logo to keep hover and focus
     // states neat
--- a/src/govuk/components/header/_index.scss
+++ b/src/govuk/components/header/_index.scss
@@ -36,7 +36,8 @@
   .govuk-header__logotype {
     display: inline-block;
 
-    // Add a gap between logo and any product name
+    // Add a gap after the logo in case it's followed by a product name. This
+    // gets removed later if the logotype is a :last-child.
     margin-right: govuk-spacing(1);
 
     // Prevent readability backplate from obscuring underline in Windows High
@@ -46,8 +47,8 @@
       color: linktext;
     }
 
-    // But remove it if there's nothing after the logo to keep hover and focus
-    // states neat
+    // Remove the gap after the logo if there's no product name to keep hover
+    // and focus states neat
     &:last-child {
       margin-right: 0;
     }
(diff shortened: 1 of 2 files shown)


src/govuk/components/header/_index.scss
```diff
@@ -43,6 +43,10 @@
     // High Contrast Mode
     forced-color-adjust: none;
 
+    @media (forced-colors: active) {
+      color: linktext;
+    }
+
     // But remove it if there's nothing after the logo to keep hover and focus
     // states neat
```
lfdebrux, line 51: I know it wasn't changed in this PR, but this comment is kind of confusing now that there is stuff between setting and removing `margin-right`

lfdebrux:
I'm not sure if this is worth fixing, but I tested this with IE11 and the visited state had a different colour to the unvisited state 🤔

src/govuk/components/header/_index.scss
36degrees: Great catch. I think there's a couple of options…

We could move the color adjust properties to after the pseudo-selector. However, I think we generally try and avoid having nested selectors in the middle of a ruleset.

```scss
  .govuk-header__logotype {
    display: inline-block;

    // Add a gap between logo and any product name
    margin-right: govuk-spacing(1);

    // But remove it if there's nothing after the logo to keep hover and focus
    // states neat
    &:last-child {
      margin-right: 0;
    }

    // Prevent readability backplate from obscuring underline in Windows
    // High Contrast Mode
    forced-color-adjust: none;

    @media (forced-colors: active) {
      color: linktext;
    }
  }
```

We could also just update the comment to make it less confusing, but the two related declarations are still a little divorced from each other:

```scss
  .govuk-header__logotype {
    display: inline-block;

    // Add a gap after the logo in case it's followed by a product name. This
    // gets removed later if the logotype is a :last-child.
    margin-right: govuk-spacing(1);

    // Prevent readability backplate from obscuring underline in Windows
    // High Contrast Mode
    forced-color-adjust: none;

    @media (forced-colors: active) {
      color: linktext;
    }

    // Remove the gap after the logo if there's no product name to keep hover
    // and focus states neat
    &:last-child {
      margin-right: 0;
    }
  }
```

Or… something else I haven't thought of.

Thoughts?

36degrees:
> I'm not sure if this is worth fixing, but I tested this with IE11 and the visited state had a different colour to the unvisited state 🤔

I think that's probably fine, and the behaviour in browsers other than IE matches the non-HCM version where the logo in the header does not have a distinct visited state either.

src/govuk/components/header/_index.scss
lfdebrux: Second option looks good to me.

lfdebrux: Is the rule about nested declarations purely a stylistic thing, or does the location affect the generated CSS?

lfdebrux (approved):

PR #2244, opened 2021-06-04
Fix diff changes to dist GitHub Action

### LINKED ISSUE #2225: GitHub Action to add a diff of the changes to dist to release PRs fails with `Error: Argument list too long`
## Description of the issue

For example, https://github.com/alphagov/govuk-frontend/pull/2224/checks?check_run_id=2575562845

When raising a release PR, we have a GitHub Action that is meant to add a diff of the changes to dist (what a mouthful!) as a comment on the PR.

This failed when we were trying to release v3.12.0 with `Error: Argument list too long`

Possibly because the diff output was too long to be passed as an environment variable to the script?

The GitHub Action should add the diff to the PR, but failed.

N/A

lfdebrux:
Fixes #2225.

Following a suggestion from another person who had the same issue [[1]], this commit saves the diff text to a file in the GITHUB_WORKSPACE folder instead. The GitHub Actions states actions can modify this directory and have subsequent actions access those changes [[2]].

This also has the advantage of removing the need to percent-encode the diff text, so the script to generate the diff can now be simpler.

---

I tested this by temporarily adding a commit to change a file in the `dist` folder; you can see the comment(s) from the GitHub action below. Once I was happy that the new action works I deleted the commit changing the dist folder.

[1]: https://github.community/t/maximum-length-for-the-comment-body-in-issues-and-pr/148867/5
[2]: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#file-systems

---

This PR also tweaks the [git diff detect renames similarity index threshold](https://git-scm.com/docs/git-diff#Documentation/git-diff.txt--Mltngt) to make sure the diff is always showing the changes to the generated CSS as a rename.

### DIFF
--- a/.github/workflows/diff-change-to-dist.yaml
+++ b/.github/workflows/diff-change-to-dist.yaml
@@ -23,22 +23,21 @@ jobs:
       - name: Generate diff
         id: diff
         run: |
-          diff=$(git diff -M1 origin/$GITHUB_BASE_REF -- dist)
-          # Escape new lines
-          diff="${diff//'%'/'%25'}"
-          diff="${diff//$'\n'/'%0A'}"
-          diff="${diff//$'\r'/'%0D'}"
-          echo "::set-output name=diff::$diff"
+          git diff -M1 origin/$GITHUB_BASE_REF -- dist \
+            > $GITHUB_WORKSPACE/dist.diff
       - name: Add comment to PR
         uses: actions/github-script@v3
-        env:
-          DIFF: ${{ steps.diff.outputs.diff }}
         with:
           github-token: ${{secrets.GITHUB_TOKEN}}
           script: |
-            const commentText = '## Changes to dist\n' + 
+            const fs = require('fs').promises
+            const diff = await fs.readFile(
+              process.env.GITHUB_WORKSPACE + '/dist.diff', 'utf8'
+            )
+
+            const commentText = '## Changes to dist\n' +
               '```diff\n' +
-              process.env.DIFF +
+              diff +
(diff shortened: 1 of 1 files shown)


.github/workflows/diff-change-to-dist.yaml
```diff
@@ -23,22 +23,21 @@ jobs:
       - name: Generate diff
         id: diff
         run: |
-          diff=$(git diff -M1 origin/$GITHUB_BASE_REF -- dist)
-          # Escape new lines
-          diff="${diff//'%'/'%25'}"
-          diff="${diff//$'\n'/'%0A'}"
-          diff="${diff//$'\r'/'%0D'}"
-          echo "::set-output name=diff::$diff"
+          git diff -M1 origin/$GITHUB_BASE_REF -- dist \
```
36degrees, line 26: I think in the last release we also had an issue because the similarity was too low, so it saw the change as a deletion and addition rather than a rename.

Looking at my history, it looks like when I generated the diff manually I had to use `-M05`, so I'm wondering if we should make that change here too?

36degrees (commented):
Thanks for picking this up 🙌🏻  The changes look good – just one query about a related change we might also want to make.

.github/workflows/diff-change-to-dist.yaml
lfdebrux: Oh yes, I remember that too, I will add that change in an additional commit.

36degrees (approved):
Great stuff, thanks 🙌🏻

Issue #2162, opened 2021-03-05
Update node dependency version in package.json


gunjam:
The current LTS version of Node.js is 14, although this repo's package.json is pinned to version 12.13.1. 12 is still in active support but this restrictive versioning is causing an issue for my project (more info in issue [711](https://github.com/alphagov/govuk-frontend/issues/711#issuecomment-791419464)).

lhokktyn:
I'd second this request to loosen this versioning restriction, if at all possible. [Contribution guidance](https://github.com/alphagov/govuk-frontend/blob/main/docs/contributing/running-locally.md) suggests the project aims to track the latest active LTS too.

On manually updating the `.nvmrc` to v15, the module appears to compile and run ok, but that's a very dirty, high-level check. Are there any issues that document the reasons we cannot go beyond v12 at the moment?

36degrees:
@lhokktyn how are you currently consuming GOV.UK Frontend?

The [published package](https://www.npmjs.com/package/govuk-frontend) does not have any upper restrictions on node version – it's only the private top level package (which includes the dependencies for our tooling etc) in the repository that has those restrictions.

lhokktyn:
> @lhokktyn how are you currently consuming GOV.UK Frontend?

We're pulling straight from the git source as described by @gunjam above (we're colleagues) in order to optimise the asset bundle served up to the browser.

This issue has become more prominent in our workflow since we've adopted some of the newer syntax available in the active LTS (14) elsewhere in our project. We work around it with clunky usage of `nvm`, but it'd be great if the govuk-frontend source could track the active LTS as it's baseline.

As I say, it _appears_ to compile ok up to v15 on the face of it so perhaps the `.nvmrc` has remained unchanged (last updated Nov 2019) because it's not breaking build pipelines?

georgeherby:
I am also referencing this repo from the source so facing the same issues mentioned here where I am running version 14.17.0 and the version check stops `npm install` from completing

jenbutongit:
Direct installation via GH aside, there are two groups that would benefit from v14 (although, it’s more of a venn intersection)

* You (GDS/maintainers) v14 will get security updates and fixes.
  * I would suggest that you also specify ^14 to force consumers to use node 14 in their own projects
* You / maintainers + consumers of the package(s) - will be able to leverage new ECMAScript features now available in v14 - without the need for a transpiler like babel.
[...]

Issue #711, opened 2018-05-22
Cannot import standard ES6 modules from npm


nickcolley:
At the moment if you try to import using our [documentation](https://github.com/alphagov/govuk-frontend#import-javascript) using a standard ES6 bundler such as Rollup it will fail.

This is because we ship our bundles as UMD, which as great for browser and CommonJs support but not ES6 Modules.

This means our [current guidance around rollup](https://github.com/alphagov/govuk-frontend#include-with-gulp-and-rollup) is misleading and only works when compiling from source.

We could consider adding more real world examples that import the package directly to ensure what we recommend is tested.

I'm not sure if we should be publishing these as `.mjs` files, which seems to be what Node.js is going towards.

We might also want to consider exporting all components from the main entry point which would allow something like:

```
import { Button } from '@govuk/frontend/all'

// new Button()
```

- Webpack example: https://github.com/alphagov/govuk-frontend/tree/master/examples/webpack
- Rollup advice on modules https://github.com/rollup/rollup/wiki/pkg.module
- Microbundle https://github.com/developit/microbundle

matthewblewitt:
Do you lose tree shaking by using 'rollup-plugin-commonjs'/UMD? I'm having to use`import Button from 'govuk-frontend/components/button/button';` rather than `import { Button } from 'govuk-frontend'` to only get the code I want.

nickcolley:
Yeah I believe that's the only way currently to achieve what you'd get for free with tree shaking.

nickcolley:
A possible way to test tree shaking: https://twitter.com/Rich_Harris/status/1031220091940204546?s=17

nickcolley:
https://philipwalton.com/articles/deploying-es2015-code-in-production-today/

nickcolley:
http://exploringjs.com/es6/ch_modules.html

gunjam:
In our project we use very few govuk-frontend js modules: `Button`, `Radios`, `Checkboxes`, `ErrorSummary`, that's about it. So to save on the bundle size I made a custom `all.js` file that only imports the components I need. The other reason we do this is that we have some custom components which use the same polyfills (and one that reuses Button) which can also be added to the same `all.js` file. The tree shaking will avoid bundling the same polyfills or components twice, resulting in a single small js bundle.

[...]

PR #2249, opened 2021-06-08
Update node version to latest LTS Fermium

gunjam:
This updates the node version in the `.nvmrc` file to the latest LTS release of 14.17.0 in order to resolve https://github.com/alphagov/govuk-frontend/issues/2162.

Node 14 is the active LTS version, with 16 being the 'current' version. 12 is now in maintenance.

Thanks.


36degrees:
Hi,
  
Thanks for raising. I'm going to approve this, but only on the basis that we’ve been meaning to move on to Node 14 for a while.
  
We really don’t want to encourage installation 'from source' by pointing directly at the GitHub repo.

The supported methods for installing GOV.UK Frontend are to:

- install the govuk-frontend package from npm
- use the pre-compiled files included with every release

I understand that you're doing this because it allows you to deliver an optimised bundle for users. It's great to see this focus on performance, and we're keen to make this easier as part of #711.

However, the contents of the GitHub repo, including the `src` folder, are considered outside of the 'contract' we have with our users. We may make changes to the repo structure or our release processes, for example, but these would not be considered breaking changes as long as the published package and pre-compiled files are not affected. We're also unlikely to make further changes to enable this unsupported installation method.

If you are decide to install from source anyway, please ensure you're pinning to a tag or commit hash that references a specific release, rather than pointing at the main branch. This is going to become especially important in the coming months as we start work on 4.0, during which time the `src` folder will contain unreleased breaking changes.
  
This'll also need a review from a second person from the team before we merge it.

36degrees (approved):

36degrees:
Note for the rest of the team – this would be an 'internal' change which doesn't need to be scheduled as part of a release or documented in release notes.

lfdebrux (approved):

PR #2255, opened 2021-06-18
Fix bugs with radios and checkboxes where multiple modules have the same name

36degrees:
We have an [open issue to provide examples of grouping radios and checkboxes under headings](https://github.com/alphagov/govuk-frontend/issues/1079).

Whilst we don’t provide any guidance on this at the minute, we’ve previously suggested that users can achieve this by making multiple calls to the `govukCheckboxes` macro and grouping them together inside a single call to `govukFieldset`. However, this means that each ‘set’ of checkboxes is inside its own module, which can cause issues with things like conditional reveals.

Add an example of this to the review app, so that we can test that everything works as expected, and fix a few issues identified that can leave conditional reveals 'out of sync' with their controls in this context.

### DIFF
--- a/src/govuk/components/checkboxes/checkboxes.js
+++ b/src/govuk/components/checkboxes/checkboxes.js
@@ -76,7 +76,7 @@ Checkboxes.prototype.syncAllConditionalReveals = function () {
  * @param {HTMLInputElement} $input Checkbox input
  */
 Checkboxes.prototype.syncConditionalRevealWithInputState = function ($input) {
-  var $target = this.$module.querySelector('#' + $input.getAttribute('aria-controls'))
+  var $target = document.getElementById($input.getAttribute('aria-controls'))
 
   if ($target && $target.classList.contains('govuk-checkboxes__conditional')) {
     var inputIsChecked = $input.checked
@@ -99,10 +99,9 @@ Checkboxes.prototype.unCheckAllInputsExcept = function ($input) {
     var hasSameFormOwner = ($input.form === $inputWithSameName.form)
     if (hasSameFormOwner && $inputWithSameName !== $input) {
       $inputWithSameName.checked = false
+      this.syncConditionalRevealWithInputState($inputWithSameName)
     }
-  })
-
-  this.syncAllConditionalReveals()
+  }.bind(this))
 }
 
 /**
@@ -121,10 +120,9 @@ Checkboxes.prototype.unCheckExclusiveInputs = function ($input) {
     var hasSameFormOwner = ($input.form === $exclusiveInput.form)
     if (hasSameFormOwner) {
       $exclusiveInput.checked = false
+      this.syncConditionalRevealWithInputState($exclusiveInput)
     }
-  })
-
-  this.syncAllConditionalReveals()
+  }.bind(this))
 }
 
 /**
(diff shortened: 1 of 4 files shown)


36degrees:
Should we include this in 3.13.0 or hold this and merge after the release has gone out?

We have 'frozen' the release, but arguably these are things that could have been fixed as part of #2151. Equally, these are existing bugs that are unlikely to come up in most contexts.

lfdebrux:
> Should we include this in 3.13.0 or hold this and merge after the release has gone out?

I vote for holding!

src/govuk/components/checkboxes/checkboxes.js
```diff
@@ -99,10 +99,9 @@ Checkboxes.prototype.unCheckAllInputsExcept = function ($input) {
     var hasSameFormOwner = ($input.form === $inputWithSameName.form)
     if (hasSameFormOwner && $inputWithSameName !== $input) {
       $inputWithSameName.checked = false
+      this.syncConditionalRevealWithInputState($inputWithSameName)
     }
-  })
-
-  this.syncAllConditionalReveals()
+  }.bind(this))
```
lfdebrux, line 104: I don't really understand why this is necessary... is this a stylistic thing or did something break without it?

src/govuk/components/radios/radios.js
```diff
@@ -77,7 +77,7 @@ Radios.prototype.syncAllConditionalReveals = function () {
  * @param {HTMLInputElement} $input Radio input
  */
 Radios.prototype.syncConditionalRevealWithInputState = function ($input) {
-  var $target = document.querySelector('#' + $input.getAttribute('aria-controls'))
+  var $target = document.getElementById($input.getAttribute('aria-controls'))
```
lfdebrux, line 80: 🙌 🙌 🙌

src/govuk/components/checkboxes/checkboxes.js
36degrees: We're now referencing `this` from within a nested function (on line 102).

By default, [`this` in the context of a function will be undefined](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this#function_context). The call to `bind` allows us to keep the context of the outer function so that `this` resolves correctly.

Once upon a time, this is the sort of thing we used to work around by aliasing `this` in the outer scope…

```js
  var that = this;

  var allInputsWithSameName = document.querySelectorAll('input[type="checkbox"][name="' + $input.name + '"]')

  nodeListForEach(allInputsWithSameName, function ($inputWithSameName) {
    var hasSameFormOwner = ($input.form === $inputWithSameName.form)
    if (hasSameFormOwner && $inputWithSameName !== $input) {
      $inputWithSameName.checked = false
      that.syncConditionalRevealWithInputState($inputWithSameName)
    }
  })
```

lfdebrux: Ah, I understand now, thanks!

lfdebrux:
I think it might be good to have tests for this behaviour to catch possible future regressions; something like 32f116d867c78e0524602613d63761a1499e1c34 for example?

36degrees:
I've realised that the (necessary) change from `this.$module.querySelector` to `document.getElementById` in bf0041a129b8aee2f0dc09c550fa9b672fa0018b is the same change we've previously flagged as a breaking change, so we should treat this as breaking as well.

Attached to the 4.0 milestone.

lfdebrux:
@36degrees I've added more tests in 6ad421d, I think covering everything this PR fixes. I have run these tests before the fixes and they fail as expected, so I think they should make good regression tests. Feel free to leave comments/suggestions on them!

vanitabarrett (approved):
Looks good to me 👍 Think it just needs a changelog entry?

PR #2257, opened 2021-06-21
Update Accordion design

chris-gds:
## What

Update the Design of the Accordion component to improve accessibility.
 
## Why

The existing iteration of the Accordion: the section headings in accordions can be mistaken for links, but are treated as buttons - this fails [WCAG 2.1 success criterion 1.3.1 Info and Relationships
](https://www.w3.org/TR/UNDERSTANDING-WCAG20/content-structure-separation-programmatic.html) this update addresses this.

Also, there was initially 3 different Accordion Designs on different parts of GOV.UK Publishing. This impacted [WCAG 3.2.4](https://www.w3.org/WAI/WCAG21/Understanding/consistent-identification.html) as the differences were not consistent they were ultimately confusing for users.

(This iterative update within the Design System (GOV.UK Frontend) will also be updated on [GOV.UK Publishing Gem](https://components.publishing.service.gov.uk/component-guide/accordion))

## Visuals 
(Before: Left, After: Right)

https://github.com/alphagov/govuk-design-system/issues/1706

## Browser testing
- [x]  IE8
- [x]  IE9
- [x]  IE10
- [x]  IE11
- [x] Edge
- [x] Chrome
- [x] Chrome 400% zoom
- [x] Firefox
- [x] Firefox 300% zoom text only 
- [x] Safari
- [x] Safari (iOS)
- [x] Chrome (iOS)
- [x] Chrome (Android)

<details>
  <summary>Selected Results:</summary>

(Trigger warning: GIFs to follow [do not pause after 3 loops](https://www.w3.org/TR/WCAG20-TECHS/G152.html))

IE8
IE8

IE9
IE9

IE10
IE10

IE11
IE11

Edge

Edge

Chrome zoom

https://user-images.githubusercontent.com/71266765/128702962-93db59b9-5a33-4675-9a95-5bacc5748d58.mov

Firefox Zoom-text only

https://user-images.githubusercontent.com/71266765/128702994-06028a6b-2f78-499b-a170-6f4e97f305dc.mov

Android Pixel 5 Chrome

pixel-chrome

Android Samsung 20 Chrome

samsung-20-chrome

Android Samsung 20 Samsung Internet

samsung-20-samsung

Safari macOS Catalina

safari-13 1-macOS-catalina

Safari macOS BigSur

safari-macOS-bigSur

iPadPro
iOS-iPadPro11-2020

iPhone 11 Pro Max
iOS-iPhone11ProMax

iPhone 12
iOS-iPhone12

iPhone SE 2020
iOS-iPhoneSE2020

Chrome - iPhone 12
chrome-ios

</details>

## Assistive Technology testing
- [x] JavaScript Disabled
- [x] CSS Disabled
- [x] CSS + JS Disabled
- [x] Voiceover / Safari (macOS Big Sur)
- [x] Voiceover / Safari (iOS 14.6)
- [x] NVDA 2020.3 / Firefox 88
- [x] NVDA 2020.3 / Chrome
- [x] JAWS 2020 / Chrome
- [x] JAWS 2020 / IE11
- [x] Windows Magnifier
- [x] Firefox (User preferred colours)
- [x] Chrome - Blurred vision, protanopia, deuteranopia, tritanopia and achromatopsia
- [x] Chrome (User preferred Colours (Increase contrast, Greyscale, Inverted colour, Inverted Greyscale, Yellow on black))
- [x] Windows High Contrast Modes - Chrome, IE, Edge

## Known issues

Untested
- [x] Talkback / Chrome 86 (Android 10)  
- [x] Dragon (speech recognition) \| 15 or later \| Internet Explorer 11*
- [x] Voiceover / Safari (iOS 12.1)

<details>
  <summary>Results:</summary>
  
JavaScript Disabled

no-JS

CSS Disabled

no-CSS

CSS + JS  Disabled

no-CSS-no-JS

VoiceOver Safari

https://user-images.githubusercontent.com/71266765/128705909-0bc64625-b261-4c31-9c99-c021017fd49e.mov

VoiceOver iOS (14.6)*
*`main` landmark present in testing environment

https://user-images.githubusercontent.com/71266765/128704394-1677774e-ce8f-465d-8507-09a5a8be04d7.mp4

NVDA 2020.4 / Firefox 88

https://user-images.githubusercontent.com/71266765/128702866-fca26121-1966-4c1a-8a97-988c366d5696.mov

NVDA 2020.4 / Chrome

https://user-images.githubusercontent.com/71266765/128702849-3773d32e-2bc1-4f6e-b254-c70352cd5b4a.mov

JAWs / Chrome  

https://user-images.githubusercontent.com/71266765/128719445-c18b226a-8819-4f37-8e79-39761c8fec09.mov

JAWs / IE  

https://user-images.githubusercontent.com/71266765/128702743-7bba985d-1290-49de-8030-e41079e0717c.mov

Chrome (Blurred vision, protanopia, deuteranopia, tritanopia and achromatopsia)

https://user-images.githubusercontent.com/71266765/128707921-d4182bad-0597-4509-a2aa-5770f520a654.mov

Chrome (Increase contrast, Greyscale, Inverted colour, Inverted Greyscale, Yellow on black)

https://user-images.githubusercontent.com/71266765/128708277-2eb79045-8b67-4c05-9b8c-9273338138df.mov

Windows Magnifier

Windows-magnifier

Chrome (Windows High Contrast Modes)

https://user-images.githubusercontent.com/71266765/128722205-50df7a61-5660-4d8e-a024-a1e2458d2a7d.mov

Edge (Windows High Contrast Mode)

https://user-images.githubusercontent.com/71266765/128721473-748f30a9-20ce-48c8-87bd-c16cce52e439.mov

IE11 (Windows High Contrast Mode)

Windows-high-contrast

</details>

## Anything else

**Breaking Change** as the [ID has now been removed from the summary element within the markup](https://github.com/alphagov/govuk-frontend/pull/2257/commits/fa572d331f030685df53923951cc5e0286b9c863)

### User preferred colours

[User's may adjust their colour schemes](https://accessibility.blog.gov.uk/2017/03/27/how-users-change-colours-on-websites/).  Testing with different colour schemes revealed some problems, having nested elements within a button [required additional adjustments](https://github.com/alphagov/govuk-frontend/commit/f258ce4e3cddc653e502193bedd32dab9a44787a) in order to improve the experience. 

There is further conversation about this use and [an opinion piece can be found here](https://kilianvalkhof.com/2021/css-html/prefers-contrast-forced-is-a-mistake/)
https://github.com/w3c/csswg-drafts/issues/5433#issuecomment-785251576

### NVDA fix

<details>
  <summary>Testing videos of NVDA</summary>

Before - the sections are broken up as user navigates.
Within reader-mode the user requires multiple keypresses to navigate, also the audio is heard in an disjointed manner.

128227998-ea003a53-9492-4a86-863c-5303950a8783

After - read out as one whole block.

https://user-images.githubusercontent.com/71266765/128356759-b3be940f-bfd2-4aff-85cf-807a4e3a8807.mov

</details>

<details>
  <summary>Additional Details and screenshots:</summary>

The issue before:
before

After:

firefox-color2

firefox-color1

---

To replicate:

`Firefox > preferences > Language and appearance > Colours`:
  
Update to high contrast scheme
Screenshot 2021-06-22 at 11 02 43

Update to reduced contrast scheme

image

</details>

## Details of breaking change

- which users are affected: TBC
- the change that’s been made: TBC
- changes users will have to make: TBC
- impact of users not making those changes: TBC

### DIFF
--- a/src/govuk/components/accordion/_index.scss
+++ b/src/govuk/components/accordion/_index.scss
@@ -84,12 +84,12 @@
         }
 
         .govuk-accordion-nav__chevron {
-          border-color: $govuk-accordion-base-colour;
+          color: $govuk-accordion-base-colour;
           background: $govuk-accordion-base-colour;
         }
 
         .govuk-accordion-nav__chevron:after {
-          border-color: $govuk-accordion-hover-colour;
+          color: $govuk-accordion-hover-colour;
         }
       }
 
@@ -101,7 +101,7 @@
         }
 
         .govuk-accordion-nav__chevron:after {
-          border-color: $govuk-accordion-icon-focus-colour;
+          color: $govuk-accordion-icon-focus-colour;
         }
       }
     }
@@ -218,7 +218,7 @@
         }
 
         .govuk-accordion-nav__chevron:after {
-          border-color: $govuk-accordion-hover-colour;
+          color: $govuk-accordion-hover-colour;
         }
       }
 
@@ -234,12 +234,12 @@
         }
 
         .govuk-accordion-nav__chevron {
-          border-color: $govuk-accordion-base-colour;
+          color: $govuk-accordion-base-colour;
           background: $govuk-accordion-base-colour;
         }
 
         .govuk-accordion-nav__chevron:after {
-          border-color: $govuk-accordion-icon-focus-colour;
+          color: $govuk-accordion-icon-focus-colour;
         }
       }
 
--- a/src/govuk/components/accordion/_index.scss
+++ b/src/govuk/components/accordion/_index.scss
@@ -118,10 +118,10 @@
 
       position: relative;
 
-      width: 20px;
-      height: 20px;
+      width: govuk-em(20, 19);
+      height: govuk-em(20, 19);
 
-      border: 1px solid;
+      border: govuk-em(1, 16) solid;
       border-radius: 50%;
 
       vertical-align: text-top;
@@ -140,16 +140,16 @@
         display: block;
 
         position: absolute;
-        bottom: 5px;
-        left: 6px;
+        bottom: govuk-em(5, 19);
+        left: govuk-em(6, 19);
 
-        width: 6px;
-        height: 6px;
+        width: govuk-em(6, 19);
+        height: govuk-em(6, 19);
 
         transform: rotate(-45deg);
 
-        border-top: 2px solid $govuk-link-colour;
-        border-right: 2px solid $govuk-link-colour;
+        border-top: govuk-em(2, 16) solid;
+        border-right: govuk-em(2, 16) solid;
 
         // IE8 fallback of icon with HTML symbol
         @include govuk-if-ie8 {
--- a/src/govuk/components/accordion/_index.scss
+++ b/src/govuk/components/accordion/_index.scss
@@ -256,6 +256,16 @@
       }
     }
 
+    // As Chevron icon is vertically aligned it overlaps with the focus state bottom border
+    // Styling adds some spacing
+    .govuk-accordion__section-button:focus .govuk-accordion__section-toggle {
+      padding-bottom: 3px;
+
+      @include govuk-media-query ($from: desktop) {
+        padding-bottom: 2px;
+      }
+    }
+
     // NVDA fix
     //
     // NVDA treats any form of line break as a different announcement landmark and will therefore announce it separately.
--- a/src/govuk/components/accordion/_index.scss
+++ b/src/govuk/components/accordion/_index.scss
@@ -275,4 +275,36 @@
     .govuk-accordion__section-toggle-text {
       margin-left: 5px;
     }
+
+    // High contrast / user colour adjustments
+    // When adjustments are active the inner button content is not visible.
+    // On hover add underline to toggle text to add extra affordance as icon colour changes on pseudo element are not respected.
+    // On focus add transparency to background so content is visible
+    @media screen and (forced-colors: active), (-ms-high-contrast: active) {
+      .govuk-accordion__section-button:hover,
+      .govuk-accordion__show-all:hover {
+        background: transparent;
+
+        .govuk-accordion-nav__chevron {
+          background-color: transparent;
+        }
+
+        .govuk-accordion__show-all-text,
+        .govuk-accordion__section-toggle {
+          text-decoration: underline;
+          text-decoration-thickness: $govuk-header-link-underline-thickness;
+        }
+      }
+
+      .govuk-accordion__show-all:focus,
+      .govuk-accordion__section-button:focus {
+        .govuk-accordion__section-heading-focus-wrapper,
+        .govuk-accordion__section-summary,
+        .govuk-accordion__section-toggle,
+        .govuk-accordion-nav__chevron {
+          background: transparent;
+          background-color: transparent;
+        }
+      }
+    }
 }
--- a/src/govuk/components/accordion/_index.scss
+++ b/src/govuk/components/accordion/_index.scss
@@ -125,7 +125,6 @@
 
     .govuk-accordion__section-heading {
       padding: 0;
-      border-top: $govuk-accordion-bottom-border-width solid $govuk-border-colour;
     }
 
     // Create Chevron icon aligned with text
@@ -198,9 +197,14 @@
     .govuk-accordion__section-button {
       width: 100%;
 
-      padding: govuk-spacing(2) 0 govuk-spacing(3);
+      padding: govuk-spacing(2) 0 govuk-spacing(1) 0;
 
-      border-width: 0;
+      border: 0;
+
+      border-top: $govuk-accordion-bottom-border-width solid $govuk-border-colour;
+      // Visually separate the section from the one underneath when user changes colours in their
+      // browser.
+      border-bottom: govuk-spacing(2) solid transparent;
 
       color: $govuk-text-colour;
       background: none;
@@ -259,6 +263,14 @@
         padding: 0;
         border: 0;
       }
+    }
+
+    // Remove the transparent border when the section is expanded to make it clear that the heading
+    // relates to the content below. Adjust padding to maintain the height of the element.
+    // See https://github.com/alphagov/govuk-frontend/pull/2257#issuecomment-951920798
+    .govuk-accordion__section--expanded .govuk-accordion__section-button {
+      padding-bottom: govuk-spacing(4);
+      border-bottom: 0;
     }
 
     // As Chevron icon is vertically aligned it overlaps with the focus state bottom border
--- a/src/govuk/components/accordion/_index.scss
+++ b/src/govuk/components/accordion/_index.scss
@@ -1,5 +1,10 @@
 @include govuk-exports("govuk/component/accordion") {
+  $govuk-accordion-base-colour: govuk-colour("black");
+  $govuk-accordion-hover-colour: govuk-colour("light-grey", $legacy: "grey-3");
   $govuk-accordion-icon-focus-colour: govuk-colour("yellow");
+  $govuk-accordion-border-width: 3px;
+  $govuk-accordion-bottom-border-width: 1px;
+  $govuk-header-link-underline-thickness: 3px;
 
   .govuk-accordion {
     @include govuk-responsive-margin(6, "bottom");
@@ -10,11 +15,6 @@
     padding-top: govuk-spacing(3);
   }
 
-  .govuk-accordion__section-header {
-    padding-top: govuk-spacing(3);
-    padding-bottom: govuk-spacing(3);
-  }
-
   .govuk-accordion__section-heading {
     // Override browser defaults to ensure consistent element height
     // Font size is set in .govuk-accordion__section-button
@@ -22,19 +22,18 @@
 
     margin-top: 0; // Override browser default
     margin-bottom: 0; // Override browser default
+
+    padding-top: govuk-spacing(3);
+    padding-bottom: govuk-spacing(3);
   }
 
   // Buttons within the sections don’t need default styling
   .govuk-accordion__section-button {
-    @include govuk-font($size: 24, $weight: bold);
-    display: inline-block;
+    @include govuk-text-colour;
+    @include govuk-font($size: 24, $weight: bold, $line-height: 1.25);
+    display: block;
     margin-bottom: 0;
     padding-top: govuk-spacing(3);
-  }
-
-  .govuk-accordion__section-summary {
-    margin-top: govuk-spacing(2);
-    margin-bottom: 0;
   }
 
   // Remove the bottom margin from the last item inside the content
@@ -46,47 +45,70 @@
   .js-enabled {
     .govuk-accordion {
       // Border at the bottom of the whole accordion
-      border-bottom: 1px solid $govuk-border-colour;
+      border-bottom: $govuk-accordion-bottom-border-width solid $govuk-border-colour;
     }
 
     // Borders between accordion sections
     .govuk-accordion__section {
-      padding-top: 0;
+      padding: 0;
     }
 
-    // Hide the body of collapsed sections
-    .govuk-accordion__section-content {
-      display: none;
-      @include govuk-responsive-padding(3, "top");
-      @include govuk-responsive-padding(3, "bottom");
-    }
-
-    // Show the body of expanded sections
-    .govuk-accordion__section--expanded .govuk-accordion__section-content {
-      display: block;
-    }
-
-    // This is styled to look like a link not a button
-    .govuk-accordion__open-all {
-      @include govuk-font($size: 16);
+    .govuk-accordion__show-all {
+      @include govuk-font($size: 19);
       position: relative;
       z-index: 1;
-      margin: 0;
-      padding: 0;
+
+      margin: 0 0 govuk-spacing(2) 0;
+      padding: 7px 2px 7px 0;
+
       border-width: 0;
+
       color: $govuk-link-colour;
       background: none;
+
       cursor: pointer;
       -webkit-appearance: none;
-
-      @include govuk-link-common;
-      @include govuk-link-style-default;
 
       // Remove default button focus outline in Firefox
       &::-moz-focus-inner {
         padding: 0;
         border: 0;
       }
+
+      &:hover {
+        color: $govuk-accordion-base-colour;
+        background: $govuk-accordion-hover-colour;
+
+        .govuk-accordion__section-toggle-text {
+          color: $govuk-accordion-base-colour;
+        }
+
+        .govuk-accordion-nav__chevron {
+          border-color: $govuk-accordion-base-colour;
+          background: $govuk-accordion-base-colour;
+        }
+
+        .govuk-accordion-nav__chevron:after {
+          border-color: $govuk-accordion-hover-colour;
+        }
+      }
+
+      &:focus {
+        @include govuk-focused-text;
+
+        .govuk-accordion-nav__chevron {
+          background: $govuk-accordion-base-colour;
+        }
+
+        .govuk-accordion-nav__chevron:after {
+          border-color: $govuk-accordion-icon-focus-colour;
+        }
+      }
+    }
+
+    // Removing padding from once JavaScript loaded
+    .govuk-accordion__section-heading {
+      padding: 0;
     }
 
     // Create Chervon icon aligned with text
@@ -154,31 +176,71 @@
       }
     }
 
-    // Section headers have a pointer cursor as an additional affordance
     .govuk-accordion__section-header {
-      position: relative;
-      // Safe area on the right to avoid clashing with icon
-      padding-right: 40px;
-      border-top: 1px solid $govuk-border-colour;
-      cursor: pointer;
+      padding: 0;
     }
 
     // Buttons within the headers don’t need default styling
     .govuk-accordion__section-button {
-      @include govuk-typography-common;
-      margin-top: 0;
-      margin-bottom: 0;
-      margin-left: 0;
-      padding: 0;
-      border-width: 0;
-      color: $govuk-link-colour;
+      position: relative;
+
+      width: 100%;
+
+      padding: govuk-spacing(2) 0 govuk-spacing(3);
+
+      border-top: $govuk-accordion-bottom-border-width solid $govuk-border-colour;
+      border-width: $govuk-accordion-bottom-border-width 0 0 0;
+
+      color: $govuk-text-colour;
       background: none;
+
       text-align: left;
+      // Section headers have a pointer cursor as an additional affordance
       cursor: pointer;
       -webkit-appearance: none;
 
+      &:active {
+        color: $govuk-link-active-colour;
+        background: none;
+      }
+
+      &:hover {
+        color: $govuk-accordion-base-colour;
+        background: $govuk-accordion-hover-colour;
+
+        .govuk-accordion__section-toggle-text {
+          color: $govuk-accordion-base-colour;
+        }
+
+        .govuk-accordion-nav__chevron {
+          color: $govuk-accordion-base-colour;
+          background: $govuk-accordion-base-colour;
+        }
+
+        .govuk-accordion-nav__chevron:after {
+          border-color: $govuk-accordion-hover-colour;
+        }
+      }
+
       &:focus {
-        @include govuk-focused-text;
+        // Remove default focus border around button as
+        // styling is being applied to inner text elements that receive focus
+        outline: 0;
+
+        .govuk-accordion__section-heading-focus-wrapper,
+        .govuk-accordion__section-summary,
+        .govuk-accordion__section-toggle {
+          @include govuk-focused-text;
+        }
+
+        .govuk-accordion-nav__chevron {
+          border-color: $govuk-accordion-base-colour;
+          background: $govuk-accordion-base-colour;
+        }
+
+        .govuk-accordion-nav__chevron:after {
+          border-color: $govuk-accordion-icon-focus-colour;
+        }
       }
 
       // Remove default button focus outline in Firefox
@@ -188,77 +250,29 @@
       }
     }
 
-    // Extend the touch area of the button to span the section header
-    .govuk-accordion__section-button:after {
-      content: "";
-      position: absolute;
-      top: 0;
-      right: 0;
-      bottom: 0;
-      left: 0;
+    // Add toggle link with Chevron icon on left.
+    .govuk-accordion__section-toggle {
+      @include govuk-font($size: 19, $line-height: 1);
+
+      color: $govuk-link-colour;
     }
 
-    .govuk-accordion__section-button:hover:not(:focus) {
-      color: $govuk-link-hover-colour;
-      text-decoration: underline;
-
-      // This needs to come after the text-decoration property otherwise
-      // text-decoration, as a shorthand property, resets it to auto
-      @include govuk-link-hover-decoration;
-      text-underline-offset: $govuk-link-underline-offset;
-    }
-
-    // For devices that can't hover such as touch devices,
-    // remove hover state as it can be stuck in that state (iOS).
-    @media (hover: none) {
-      .govuk-accordion__section-button:hover {
-        text-decoration: none;
-      }
-    }
-
-    .govuk-accordion__controls {
-      text-align: right;
-    }
-
-    // Display an icon to the right of each header to indicate open/closed status,
-    // and as an additional affordance.
-    .govuk-accordion__icon {
-      position: absolute;
-      top: 50%;
-      right: 15px;
-      width: 16px;
-      height: 16px;
-      margin-top: -8px;
-    }
-
-    .govuk-accordion__icon:after,
-    .govuk-accordion__icon:before {
-      content: "";
-      box-sizing: border-box;
-      position: absolute;
-      top: 0;
-      right: 0;
-      bottom: 0;
-      left: 0;
-      width: 25%;
-      height: 25%;
-      margin: auto;
-      border: 2px solid transparent;
-      background-color: govuk-colour("black");
-    }
-
-    .govuk-accordion__icon:before {
-      width: 100%;
-    }
-
-    .govuk-accordion__icon:after {
-      height: 100%;
-    }
-
-    // Vertical bar should be hidden when section is open, to display a '-' icon
-    .govuk-accordion__section--expanded .govuk-accordion__icon:after {
-      content: " ";
+    // Hide body of expanded sections
+    .govuk-accordion__section-content {
       display: none;
+      @include govuk-responsive-padding(0, "top");
+      @include govuk-responsive-padding(8, "bottom");
     }
-  }
+
+    // Show the body of expanded sections
+    .govuk-accordion__section--expanded .govuk-accordion__section-content {
+      display: block;
+    }
+
+    // Space between icon and text,
+    // Avoid placing on icon due to transform altering placement of margin
+    .govuk-accordion__show-all-text,
+    .govuk-accordion__section-toggle-text {
+      margin-left: 5px;
+    }
 }
--- a/src/govuk/components/accordion/_index.scss
+++ b/src/govuk/components/accordion/_index.scss
@@ -287,37 +287,11 @@
       }
     }
 
-    // 1. A fix to improve NVDA announcements.
-    //
-    // NVDA treats any form of line break as a different announcement landmark and will therefore announce it separately.
-    // As you move through NVDA using arrow keys, this announces each element in turn (rather than a whole block) which could cause confusion for users
-    // This line break can come from both the semantics via br tags or two block-level tags, or via CSS using display: block or display: table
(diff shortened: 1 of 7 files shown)


src/govuk/components/accordion/accordion.js
```diff
@@ -93,49 +105,55 @@ Accordion.prototype.initSectionHeaders = function () {
...
-    $button.setAttribute(attr.nodeName, attr.nodeValue)
-  }
-
-  $button.addEventListener('focusin', function (e) {
-    if (!$headerWrapper.classList.contains($module.sectionHeaderFocusedClass)) {
-      $headerWrapper.className += ' ' + $module.sectionHeaderFocusedClass
+    // Add all attributes but not ID as this is being add directly on $wrapperFocusHeading
+    if (attr.nodeName !== 'id') {
+      $button.setAttribute(attr.nodeName, attr.nodeValue)
+      console.log(attr.nodeName, attr.nodeValue)
```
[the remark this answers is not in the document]
chris-gds: 🙈

src/govuk/components/accordion/accordion.js
```diff
@@ -93,49 +105,55 @@ Accordion.prototype.initSectionHeaders = function () {
...
 // Set individual header attributes
 Accordion.prototype.initHeaderAttributes = function ($headerWrapper, index) {
-  var $module = this
   var $span = $headerWrapper.querySelector('.' + this.sectionButtonClass)
   var $heading = $headerWrapper.querySelector('.' + this.sectionHeadingClass)
   var $summary = $headerWrapper.querySelector('.' + this.sectionSummaryClass)
 
   // Copy existing span element to an actual button element, for improved accessibility.
   var $button = document.createElement('button')
-  $button.setAttribute('type', 'button')
```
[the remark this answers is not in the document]
36degrees: This is setting the button's`type`, not `role`.

I think there's _potential_ for this change to cause issues for accordions inside forms, as the default `type` for a button is `submit` and I can't see anything in the JS to prevent the default click behaviour on the button which would be to submit the form.

src/govuk/components/accordion/accordion.test.js
```diff
@@ -118,13 +118,24 @@ describe('/components/accordion', () => {
...
+        })
+      })
+
+      describe('hidden text', () => {
+        it('should contain hidden text " this section" for assistive technology', async () => {
+          await page.goto(baseUrl + '/components/accordion/preview', { waitUntil: 'load' })
+
+          const numberOfExampleSections = 2
+
+          const numberOfIcons = await page.evaluate(() => document.body.querySelectorAll('.govuk-accordion .govuk-accordion__section .govuk-visually-hidden').length)
```
hannalaakso, line 138: It would be good to check that the visually hidden div contains the " this section" text. Similar to what we do with the "Show all sections" button.

src/govuk/components/accordion/accordion.test.js
```diff
@@ -118,13 +118,24 @@ describe('/components/accordion', () => {
...
 
           const numberOfExampleSections = 2
 
-          const numberOfIcons = await page.evaluate(() => document.body.querySelectorAll('.govuk-accordion .govuk-accordion__section .govuk-accordion__icon').length)
+          const numberOfIcons = await page.evaluate(() => document.body.querySelectorAll('.govuk-accordion .govuk-accordion__section .govuk-accordion-nav__chevron').length)
+          expect(numberOfIcons).toEqual(numberOfExampleSections)
+        })
+      })
+
+      describe('hidden text', () => {
```
hannalaakso, line 132: It would be great if we could add a test for the "Show" text in the button as well.

src/govuk/components/accordion/accordion.js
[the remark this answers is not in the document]
frankieroberto: Yes, agreed – this would be a breaking bug for anywhere accordions are used inside forms. My service current does this, although we’re in the process of moving away from using the accordion in this way.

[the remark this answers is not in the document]
chris-gds: All makes sense, restored. 🙌🏼

src/govuk/components/accordion/accordion.test.js
chris-gds: Theres's now a test for the text in it's entirely - hopefully enough?

src/govuk/components/accordion/_index.scss
```diff
@@ -42,76 +35,214 @@
...
-      margin-bottom: 0;
-      margin-left: 0;
-      padding: 0;
-      border-width: 0;
-      color: $govuk-link-colour;
+      position: relative;
+
+      width: 100%;
+
+      padding: govuk-spacing(2) 0 govuk-spacing(6);
```
frankieroberto, line 191: Could the padding at the bottom of the button be reduced, to say `govuk-spacing(2)` to match the padding on top (10px instead of 30px)? The padding on the top of the expanded content section could then be increased slightly to maintain the spacing.

The rationale for reducing the padding would be to reduce the height that each section takes up when collapsed, as this new design is significantly taller than the current one. This might impact usability, as it is harder to scan a list of sections when you can't see as many of them at once.

The version on the [Coronavirus page](https://www.gov.uk/coronavirus) currently has 96px high sections whereas this version has 107px high sections. Existing component has 61px high sections.

chris-gds: 👋🏼  Hey @frankieroberto, so at this stage this current Design has been reviewed and finalised by multiple teams however the observations you've raised would be worth adding to user-testing to see the impact.

> as this new design is significantly taller than the current one. This might impact usability, as it is harder to scan a list of sections when you can't see as many of them at once.

Also, there is a [condensed version](https://components.publishing.service.gov.uk/component-guide/accordion/condensed_layout/preview) of the layout that exists for GOV.UK that might be of some interest.  I'm aware that the Accordion's guidance centres around this element being used a "last resort" and _if_ the Accordion houses too much Content then it _might_ be worth a review

frankieroberto: Fair enough! Will the condensed version be an option within the component?

chris-gds: AFAIK not within the Design System however this option will persist on GOV.UK.  However, there _might_ be more of a case to remove this as well.

I suppose it prompts an upstream question, if there are lot of sections and the larger Design potentially impacts how a user scans  - why are there that many sections?  Does the Content need review? Also discovering the details of the scanning.. impact.
 
I've be curious to discover all the instances of this being used cross-gov to get a better understanding of use.  Equally the reserve could be true + once this is in situ / user research this might be adjusted.

frankieroberto: @chris-gds I suspect the accordion is used (abused?) in all sorts of different ways!

One of the primary ways I’ve seen it used though is for navigation, where each section is essentially a list of links instead of content in paragraphs.

Examples:

* https://explore-education-statistics.service.gov.uk/find-statistics
* https://www.ethnicity-facts-figures.service.gov.uk/education-skills-and-training

Perhaps there’s better ways to do this though.

src/govuk/components/accordion/accordion.js
```diff
@@ -44,30 +51,36 @@ Accordion.prototype.init = function () {
...
+  // Create "Show all" button and set attributes
   this.$openAllButton = document.createElement('button')
-  this.$openAllButton.setAttribute('type', 'button')
-  this.$openAllButton.innerHTML = 'Open all <span class="govuk-visually-hidden">sections</span>'
   this.$openAllButton.setAttribute('class', this.openAllClass)
   this.$openAllButton.setAttribute('aria-expanded', 'false')
-  this.$openAllButton.setAttribute('type', 'button')
+
+  // Create icon, add to element
+  var icon = document.createElement('span')
```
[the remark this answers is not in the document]
chris-gds: Makes sense, this should be reflected now

src/govuk/components/accordion/_index.scss
chris-gds: >I suspect the accordion is used (abused?) in all sorts of different ways!

+1, thanks for these - it's useful to see how it's being used elsewhere (I've not seen these before)
Out of scope to comment on this use within this PR but having more examples is useful to understand how this update might interact

hannalaakso: Thanks @frankieroberto @chris-gds for your thoughts!

I think you make a really good point about accordions often being misused when another component/pattern might be more suitable. We have actually have it in our backlog to update the component guidance to be clearer on this point when we publish this new version. Have a look [here](https://github.com/alphagov/govuk-design-system/issues/1684#issuecomment-846946557) (it's point 2) and feel free to leave any comments on that issue if you think there is anything else we should take into account when updating the guidance.

lfdebrux:
Browser testing - Edge - Windows High Contrast Mode

https://user-images.githubusercontent.com/503614/127344634-8907dc4f-dc12-430f-a2c9-bdd88da1c5f5.mp4

src/govuk/components/accordion/_index.scss
frankieroberto: @hannalaakso thanks, added a comment to that issue. Whether accordions are being misused or not is a thorny debate!

src/govuk/components/accordion/accordion.test.js
```diff
@@ -118,13 +116,46 @@ describe('/components/accordion', () => {
...
+        await page.goto(baseUrl + '/components/accordion/preview', { waitUntil: 'load' })
+
+        const ShowOrHideButtonText = await page.evaluate(() => document.body.querySelector('.govuk-accordion__section-toggle-text').textContent)
+
+        await page.click('.govuk-accordion .govuk-accordion__section:nth-of-type(2) .govuk-accordion__section-header')
+        await page.click('.govuk-accordion .govuk-accordion__section:nth-of-type(3) .govuk-accordion__section-header')
+
+        expect(ShowOrHideButtonText).toEqual('Hide this section')
+      })
+
```
lfdebrux, line 151: There's a bit of a (pre-existing) gap in the tests, currently there are no tests for an accordion with additional description/summary text.

Given that this PR adds a bit more sophisticated handling of the additional description, it might be worth adding tests such as `it('should move the additional description to the button text')`; would that be okay?

src/govuk/components/accordion/accordion.test.js
```diff
@@ -118,13 +116,46 @@ describe('/components/accordion', () => {
...
+          await page.goto(baseUrl + '/components/accordion/preview', { waitUntil: 'load' })
+
+          const numberOfExampleSections = 2
+          const numberOfCommas = await page.evaluate(() => document.body.querySelectorAll('.govuk-accordion .govuk-accordion__section .govuk-visually-hidden:first-child').length)
+
+          expect(numberOfCommas).toEqual(numberOfExampleSections)
+        })
+      })
+
+      it('should change the (combined) Show[ this section] text to Hide[ this section ] when sections are opened', async () => {
```
hannalaakso, line 141: This test should not pass, but is passing because the accordion is still open from another test scenario.

We should possibly be scoping the tests differently in general, @lfdebrux will investigate this and will raise a PR against main branch to fix.

src/govuk/components/accordion/accordion.js
```diff
@@ -50,7 +50,6 @@ Accordion.prototype.init = function () {
   }
```
hannalaakso, line 50: As moving the summary inside the button is a slightly unexpected change (which then requires the change in this commit), we discussed that it would be good to 
- move code that moves the summary inside the button into this commit to have a self contained commit of all the related changes
- add a bit more detail to the commit message about why we’re moving the summary inside the button (mainly to do with design as we didn’t want the summary text below the button element)
- why moving the summary means that it can no longer be a span

src/govuk/components/accordion/_index.scss
```diff
@@ -190,77 +252,57 @@
...
-      width: 100%;
+    // Show the body of expanded sections
+    .govuk-accordion__section--expanded .govuk-accordion__section-content {
+      display: block;
     }
 
-    .govuk-accordion__icon:after {
-      height: 100%;
+    // Remove the bottom margin from the last item inside the content
+    .govuk-accordion__section-content > :last-child {
```
hannalaakso, line 307: We moved this rule here but didn't delete the same rule from line 32, the duplicate should be removed.

src/govuk/components/accordion/_index.scss
```diff
@@ -263,6 +263,15 @@
       &:active {
         background: $govuk-link-active-colour;
       }
+
```
hannalaakso, line 266: We discussed that this commit https://github.com/alphagov/govuk-frontend/pull/2257/commits/4d44935136f9470d851fb6f7e6e37998f3980176# 
might need 
- a bit more detail on the original problem that we're fixing
- the desired outcome
- to be split in two as there are two slightly different changes here

It would also be good to dig out the original bug report for NVDA - @chris-gds let me know if I can help with that.

src/govuk/components/accordion/accordion.test.js
```diff
@@ -118,13 +116,46 @@ describe('/components/accordion', () => {
...
+          await page.goto(baseUrl + '/components/accordion/preview', { waitUntil: 'load' })
+
+          const numberOfExampleSections = 2
+          const numberOfIcons = await page.evaluate(() => document.body.querySelectorAll('.govuk-accordion .govuk-accordion__section .govuk-accordion-nav__chevron').length)
+
+          expect(numberOfIcons).toEqual(numberOfExampleSections)
+        })
+      })
+
+      describe('hidden comma', () => {
```
hannalaakso, line 130: It would be good to update this test so that we can check that the div content equals the comma symbol.

src/govuk/components/accordion/_index.scss
```diff
@@ -44,74 +37,212 @@
...
+      width: 20px;
+      height: 20px;
+
+      border: 1px solid;
+      border-radius: 50%;
+
+      vertical-align: text-top;
+
+      // IE8 fallback of icon
+      @include govuk-if-ie8 {
```
hannalaakso, line 128: The arrow lines for IE8 are very narrow (might not be visible to all users) and also slightly overlaps the borders of the box. We don't need to do anything too fancy here for IE8 but there might some slightly improvements we could make, I think @chris-gds is going to experiment with bolding the icon.

src/govuk/components/accordion/accordion.js
```diff
@@ -104,6 +119,18 @@ Accordion.prototype.initHeaderAttributes = function ($headerWrapper, index) {
...
+  // Create show / hide icons with text.
+  var $showIcons = document.createElement('span')
+  $showIcons.classList.add(this.sectionShowHideToggleClass)
+
+  // Build additional wrapper for toggle text, place icon before wrapped text.
+  var $wrapperShowHideIcon = document.createElement('span')
+  var $icon = document.createElement('span')
+  $icon.classList.add(this.upChevronIconClass)
+  $showIcons.appendChild($icon)
+  $wrapperShowHideIcon.classList.add(this.sectionShowHideTextClass)
```
hannalaakso, line 139: Only a small thing but the variable name is slightly confusing as it contains the toggle text.

src/govuk/components/accordion/accordion.test.js
```diff
@@ -118,13 +116,46 @@ describe('/components/accordion', () => {
...
+
+        expect(ShowOrHideButtonText).toEqual('Hide this section')
+      })
+
+      describe('hidden text', () => {
+        it('should contain hidden text " this section" for assistive technology', async () => {
           await page.goto(baseUrl + '/components/accordion/preview', { waitUntil: 'load' })
 
           const numberOfExampleSections = 2
+          const numberOfIcons = await page.evaluate(() => document.body.querySelectorAll('.govuk-accordion .govuk-accordion__section .govuk-accordion__section-toggle-text .govuk-visually-hidden').length)
```
hannalaakso, line 157: Not a big deal but a slightly confusing variable name here.

src/govuk/components/accordion/_index.scss
```diff
@@ -190,77 +252,57 @@
...
-      text-underline-offset: $govuk-link-underline-offset;
-    }
+      color: $govuk-link-colour;
 
-    // For devices that can't hover such as touch devices,
-    // remove hover state as it can be stuck in that state (iOS).
-    @media (hover: none) {
-      .govuk-accordion__section-button:hover {
-        text-decoration: none;
+      &:active {
```
hannalaakso, line 263: This is probably redundant as the container is a div.

src/govuk/components/accordion/_index.scss
```diff
@@ -87,6 +89,73 @@
...
+        position: absolute;
+        bottom: 5px;
+        left: 6px;
+
+        width: 6px;
+        height: 6px;
+
+        transform: rotate(-45deg);
+
+        border-top: 2px solid;
```
hannalaakso, line 149: A small thing but we could explicitly set the colour for the borders, at the moment it's being inherited from the toggle element which could change.

hannalaakso (commented):
Thanks @chris-gds for all your work on this, it's looking good 🎉 

I left some comments, mainly to do with adding some more code comments and clarifying a few things.

A couple of other things that we spotted:
- [x] iOS Voiceover announces summary text twice on 12.1
- [ ] iOS Voiceover announces the show all button as a main landmark - I think you correctly said @chris-gds that this is to do with the context we're testing in. It would be good to see up a full page example with the accordion which should fix this, let me know if I can help with that @chris-gds 
- [x] NVDA read mode (Insert + Arrow Up/Down) splits the heading/summary/button/whole element into separate entities, not sure if this is an issue for users but I think you said @chris-gds that you might know why this is and will investigate
nvda
- [x] Picked up as part of the design review: the headings should be the same as `govuk-heading-m` and the summary should be the same as `govuk-body`
- [x] We know that there are some really useful contextual comments for the design changes [here](https://docs.google.com/document/d/1CG0OUT1tzy7q9juK4O0Iv974854hiBK5GMGF9_0ma08/edit#heading=h.1i6pnax09yoi) - hopefully we could move those into the commit messages to explain the design rationale 
- [x] There are a couple of things we could move around in the commit messages to make it easier to review the changes - listed [here](https://docs.google.com/document/d/1CG0OUT1tzy7q9juK4O0Iv974854hiBK5GMGF9_0ma08/edit#heading=h.i43eyiccsig0), but our team could also pick those up later

src/govuk/components/accordion/_index.scss
chris-gds: Update to alter icon to improve this (Before / After) 

IE8-update

(Possible action point: Design icon AC: include IE8)

src/govuk/components/accordion/_index.scss
chris-gds: Agreed, updating commit with details splitting out commit thank contains pause 

-----

Additional Details for context 
Original issue flagged:

Not sure if this is a quirk of NVDA, but if you navigate through the accordion (arrow keys), it reads out the heading (together with ‘Show’) and ‘this section’ as 2 separate items, which could be confusing:

_clickable  heading    level 2  button  collapsed    Know your audience,Show
heading    level 2  button  collapsed    this section_

---- 

Clinic Notes:
"Notably, we didn’t take issue with when NVDA split the button into separate blocks. It’ll be good regardless to get some detailed rationale as to why this could be acceptable and how theoretically impactful the lack of pausing is.

The rationale was NVDA has been seen to treat headings / buttons in a similar fashion elsewhere on the web. There would be a degree of expectation that the user might be used to this."

---

Action -  agreed to fix issue.

src/govuk/components/accordion/_index.scss
```diff
@@ -44,74 +37,212 @@
...
+    }
+
+    // Create Chervon icon aligned with text
+    .govuk-accordion-nav__chevron {
+      box-sizing: border-box;
+      display: inline-block;
+
+      position: relative;
+
+      width: 20px;
```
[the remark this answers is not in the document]
chris-gds: Agreed - the `govuk_publishing_components` version of this iteration contains relative values[1] in order to scale as expected, in the context of this contribution it was noted this pattern is perhaps not present within the Design System at present so didn't wish to introduce a change in convention.  This could be worth exploring at a later date for all icons to meet a zoom text-only AC.

[1]https://github.com/alphagov/govuk_publishing_components/commit/192ff22f1332c18d710e42889431b6c92eeda3e1

src/govuk/components/accordion/_index.scss
chris-gds: In response - @christopherthomasdesign feedback:

> The component being taller generally is a necessary side effect of having the show all controls on a new line. I actually agree with frankie on making the padding equal on the top and bottom – need to agree this with designers on gov.uk because there was deffo a reason for it Nikin happy with 15px rather than 30 under each row

Design @nnagewad being altered to include this change.

src/govuk/components/accordion/accordion.test.js
chris-gds: As a backup to that action - updated the test and added a reset as well to tackle this

src/govuk/components/accordion/_index.scss
chris-gds: - Added more detail to commits about the history of being spotted, feedback, investigation and fix also,
- Added inline comments to further outline issue.
- PR also updated with before and after videos for visual reference of the issue
- Pause split out of original commit for clarity

src/govuk/components/accordion/accordion.js
chris-gds: Updated the naming to be clearer to `wrapperToggleText`

src/govuk/components/accordion/accordion.test.js
chris-gds: Make sense, i've added a tests to check for the span being changed. However it might be worth filling that gap to add the missing check of the summary text just existing.

src/govuk/components/accordion/accordion.js
```diff
@@ -93,49 +105,78 @@ Accordion.prototype.initSectionHeaders = function () {
...
+  $heading.removeChild($span)
+  $heading.appendChild($button)
+  $button.appendChild($wrapperFocusHeading)
+  $button.appendChild($srPause)
+  // span could contain HTML elements (see https://www.w3.org/TR/2011/WD-html5-20110525/content-models.html#phrasing-content)
+  $wrapperFocusHeading.innerHTML = $span.innerHTML
 
+  // If summary content exists add to DOM in correct order
   if (typeof ($summary) !== 'undefined' && $summary !== null) {
-    $button.setAttribute('aria-describedby', this.moduleId + '-summary-' + (index + 1))
```
hannalaakso, line 124: If we remove this, we're no longer linking the button to the summary text. That might be okay as the summary is now inside the button but we should probably re-test on screen readers to see if the summary is still announced -  I think you were anticipating this @chris-gds?

If we do retain this change, we should also delete the id from the summary, I don't think it serves any other purpose.

chris-gds:
> A couple of other things that we spotted:
> * [x]  iOS Voiceover announces summary text twice on 12.1

Removed `aria-describedby` to resolve this -  as the summary has moved inside the button this additional aria attribute is no longer needed 🎉  so now prevents the duplicated announcement.

> * [ ]  iOS Voiceover announces the show all button as a main landmark 

I'm 90% certain this due to the testing environment being wrapped in a `main` as this also announced on other elements like breadcrumbs. Will confirm during deeper testing once accordion is placed on a different page.

> * [x]  NVDA read mode (Insert + Arrow Up/Down) splits the heading/summary/button/whole element into separate entities, 

The fix wasn't in place correctly, this has been resolved.

> * [x]  Picked up as part of the design review: the headings should be the same as `govuk-heading-m` and the summary should be the same as `govuk-body`

This has been updated however it's worth noting as a result this is a **breaking change** as it update the template file. 

Also if we move ahead with his update (might need talk a bit more) it might be worth exploring changing the `<div>` to `<section>` which was omitted to reduce this impact

> * [x]  We know that there are some really useful contextual comments for the design changes [here](https://docs.google.com/document/d/1CG0OUT1tzy7q9juK4O0Iv974854hiBK5GMGF9_0ma08/edit#heading=h.1i6pnax09yoi) - hopefully we could move those into the commit messages to explain the design rationale

This has been done, I've added a lot more detail about the change rationale within the commit from Design(ers) to help outline why the changes were done.

src/govuk/components/accordion/accordion.js
chris-gds: >we're no longer linking the button to the summary 

Exactly, as the summary text is now _inside_ the button the additional `aria-describedby` is not longer required as it's not needed to be referenced by something outside.  The heading + summary naturally is announced should be present.

> re-test on screen readers to see if the summary is still announced - I think you were anticipating this @chris-gds?

Yes, there's a fresh round of testing that's needed to just make sure this works as intended.

> If we do retain this change, we should also delete the id from the summary, I don't think it serves any other purpose.

This is a good point 🤔  - as we're updating the template (mentioned below might need a talk) to add `govuk-heading-` this can be removed as well.

src/govuk/components/accordion/_index.scss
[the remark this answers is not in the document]
chris-gds: After some thought decided it was more favourable to introduce relative values in this context. [This has been applied here:](https://github.com/alphagov/govuk-frontend/pull/2257/commits/aae5ab9486383189d1450057d1f49bd0e18579e8)

Updated visuals (zoom text only firefox):

Aug-09-2021 10-58-59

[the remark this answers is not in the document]
hannalaakso: Brilliant, thanks @chris-gds 🎉

src/govuk/components/accordion/accordion.js
hannalaakso: A small thing but it seems that this comment from https://github.com/alphagov/govuk-frontend/pull/2257/commits/408b8cdd85e7abcd2a2152f14e130d1cad263815

> (@nnagewad) "We’ve decided to move the summary in order to better visually illustrate the additional context of accordion heading by placing it directly beneath the heading as opposed to placing it directly beneath the reveal/expand button"[4]

should probably be in https://github.com/alphagov/govuk-frontend/pull/2257/commits/61020eb0de84361ff2a2b9da3dc0ca53d106b223. 

We could also signpost to https://html.spec.whatwg.org/multipage/form-elements.html#the-button-element re: 
> Content model: 
> [Phrasing content](https://html.spec.whatwg.org/multipage/dom.html#phrasing-content-2)

chris-gds: Moved the justification into the other commit & added the additional sign post to phrasing.

chris-gds:
> > * [x]   iOS Voiceover announces the show all button as a main landmark
> 
> I'm 90% certain this due to the testing environment being wrapped in a `main` as this also announced on other elements like breadcrumbs. Will confirm during deeper testing once accordion is placed on a different page.

Testing in situ below, non-issue - no action

https://user-images.githubusercontent.com/71266765/128730749-27674af6-2b7e-460f-8f19-4c75b3d58833.mp4

christopherthomasdesign:
Just spotted a couple of small details!

The shape of the circle around the chevron is currently changing between states, so that when the section is open it gets narrower and isn't a perfect circle. You can see in the second image here it looks a bit squished:

Screenshot 2021-08-09 at 16 53 49
Screenshot 2021-08-09 at 16 53 54


It should stay the same, a perfect circle with only the rotation of the chevron changing.

---

To me, on smaller screens the 'show all sections' looks a little bit far from the sections below. 

Screenshot 2021-08-09 at 16 26 42

It currently uses a margin-bottom of 14px. Assuming that's 14 in order to leave 1px for the line, and therefore keep the 5px vertical rhythm. If that's so, I'd probably reduce it to 9px at 'mobile' and back up to 14px for larger screens. Small difference but I think it will tie it together a bit better.

chris-gds:
> isn't a perfect circle. You can see in the second image here it looks a bit squished:

Agreed, this new conversion from `pixels` to `em` doesn't quite seem... exact. Will sort.

---

> It currently uses a margin-bottom of 14px. Assuming that's 14 in order to leave 1px for the line, and therefore keep the 5px vertical rhythm.

[taking a look at the Design](https://www.figma.com/file/3yTE9s4EOkvjcqRnqaIzgG/Accordion-Redesign?node-id=343%3A726) it's currently reflective of the 10px across both views however [I've a feeling some adjustments to the Hover state](https://github.com/alphagov/govuk-frontend/pull/2257/commits/b1c57bb81b8121aa90112a8dd6a53b06bffd918c) aren't that obvious.

Screenshot 2021-08-09 at 17 35 35

Screenshot 2021-08-09 at 17 35 25

>To me, on smaller screens the 'show all sections' looks a little bit far from the sections below.

> If that's so, I'd probably reduce it to 9px at 'mobile' and back up to 14px for larger screens. Small difference but I think it will tie it together a bit better.

In this context I think changing it to a relative value will feel a smidge tighter on mobile and give that 1px 
reduction feel.

chris-gds:
As the NVDA fix changes the element to `display:inline` there's less control in styling which results in the summary copy - while focused-  visually displaying a 1px line due the `box-shadow`. A potential fix is to increase the font from `16px` to `17px` however then strays aways from `govuk-body` values.

Screenshot-2021-08-10-at-18 50 04

@hannalaakso will confirm with Design

PR #2261, opened 2021-06-24
WIP: Task list component collaboration

frankieroberto:


frankieroberto:
One of the questions for this component is what do we call the param that holds the array of things with statuses?

Our initial instinct was to use `rows` to match `Summary List` and because they look like rows (horizontally divided).

However I've done a quick check of what existing components use:

Accordion: `items`
Breadcrumbs: `items`
Checkboxes: `items`
Error summary: `errorList`
Footer: `items`, `navigation`
Header: `navigation`
Radios: `items`
Summary List: `rows`, `actions`
Table: `rows`, `head`
Tabs: `items`

Based on this, I wonder whether it would be more consistent to use `items` instead? That would also not tie us to a particular visual style, should the component evolve in future to render differently.

frankieroberto:
GDS wrote a summary back in 2017 of some of the [pros and cons of using Tables, Definition lists and Lists](https://github.com/alphagov/govuk-prototype-kit/pull/365#issuecomment-288417322).

frankieroberto:
We might want to add `role=list` to the `<ul>`, see https://github.com/alphagov/govuk-prototype-kit/issues/1009 cc: @edwardhorsford

frankieroberto:
Here’s the [original PR which added the task list to the Prototyping Kit](https://github.com/alphagov/govuk-prototype-kit/pull/361).

frankieroberto:
> One of the questions for this component is what do we call the param that holds the array of things with statuses?
> [...]
> Based on this, I wonder whether it would be more consistent to use `items` instead? That would also not tie us to a particular visual style, should the component evolve in future to render differently.

I've switched to `items` as the param name in 181273d78778ca14a7e5b923607dd483f101c426, based on the following rationale:

* 6 components use `items` vs 2 that use `rows`
* Consistency with the HTML tag name (`<li>` - "list item") and CSSS class name (`govuk-task-list__item`)
* More generic, not tied to visual appearance

owenatgov (commented):
At the request of the team and community members looking after the task list contribution I've given this an initial review. Fully conscious that this is still in draft so take my comments with a pinch of salt as I imagine most of them would be addressed anyway. We for example need some more tests once this is ready to contribute in full but that's not something we need to worry about right now.

I've also done some browser and assistive tech testing and this is looking good. Great stuff overall 💪🏻

frankieroberto:
@owenatgov many thanks for reviewing this! It is still work-in-progress, but now is definitely a good time for us to start to get detailed reviews and to refine the code.

src/govuk/components/task-list/_index.scss
```diff
@@ -0,0 +1,103 @@
...
+    margin-top: 0;
+    margin-bottom: 0;
+    padding: 0;
+    list-style-type: none;
+    @include govuk-media-query($from: tablet) {
+      min-width: 550px;
+    }
+  }
+
+  .govuk-task-list__items {
```
Dawdre, line 13: What element does this class correspond to? I can't seem to see it in the template.

src/govuk/components/task-list/_index.scss
```diff
@@ -0,0 +1,103 @@
...
+
+  .govuk-task-list__status {
+    display: inline-block;
+    margin-top: govuk-spacing(2);
+    margin-bottom: govuk-spacing(1);
+    margin-left: 0;
+    padding-top: 2px;
+    padding-right: 4px;
+    padding-bottom: 2px;
+    padding-left: 0;
```
Dawdre, line 56: What is the reason for setting this to `0`? I have attached a screenshot to show the padding being lost on smaller screens creating an imbalance.

image

Dawdre (commented):
Couple minor comments. Looks good otherwise 👍🏻

src/govuk/components/task-list/_index.scss
```diff
@@ -0,0 +1,103 @@
...
+    }
+  }
+
+  .govuk-task-list__hint {
+    display: block;
+    margin-top: 5px;
+    color: $govuk-secondary-text-colour;
+  }
+
+  .govuk-task-list__status {
```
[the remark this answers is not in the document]
frankieroberto: Yeah, it's tricky! One the one hand it looks quite similar to the existing tag, and perhaps the changes ought to apply to that component too. But on the other hand, that'd then involve updating the other component too and tying the two together. So far we decided to keep them separate, as that makes this new component a non-breaking change, and would potentially allow the "status" sub-component and the "tag" component to deviate further in future.

But it's one we should keep under review!

src/govuk/components/task-list/_index.scss
frankieroberto: Well spotted - this is a bug, will fix!

src/govuk/components/task-list/_index.scss
```diff
@@ -0,0 +1,103 @@
...
+    display: inline-block;
+    margin-top: govuk-spacing(2);
+    margin-bottom: govuk-spacing(1);
+    margin-left: 0;
+    padding-top: 2px;
+    padding-right: 4px;
+    padding-bottom: 2px;
+    padding-left: 0;
+    @include govuk-media-query($from: 450px) {
+      margin-top: 0;
```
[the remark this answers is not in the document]
frankieroberto: See other comment - I'm not sure whether we should explicitly set margin left and right to 0 or not.

src/govuk/components/task-list/task-list.yaml
```diff
@@ -0,0 +1,207 @@
+params:
+- name: items
+  type: array
+  required: true
+  description: Array of item objects.
+  params:
+- name: classes
+  type: string
+  required: false
+  description: Classes to add to the item `div`
```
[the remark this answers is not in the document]
frankieroberto: yes, although this reminds me to add support for `classes` on the top-level `<ul>` element... 👍

src/govuk/components/task-list/task-list.yaml
```diff
@@ -0,0 +1,207 @@
+params:
+- name: items
+  type: array
+  required: true
+  description: Array of item objects.
+  params:
+- name: classes
```
owenatgov, line 7: A very small nitpick that these params of `items` should be indented one more so that it's communicated in the yml and therefore in the nunjucks docs via guidance when we get around to writing it up that these relate to the `items` params, not the params for the root component.

src/govuk/components/task-list/task-list.yaml
```diff
@@ -0,0 +1,207 @@
...
+  description: Classes to add to the container.
+- name: attributes
+  type: object
+  required: false
+  description: HTML attributes (for example data attributes) to add to the container.
+
+examples:
+  - name: default
+    data:
+      idPrefix: "task-list-example"
```
owenatgov, line 59: `idPrefix` isn't documented above.

src/govuk/components/task-list/_index.scss
```diff
@@ -0,0 +1,120 @@
...
+  // This class is added to the <li> elements where the task name is a link.
+  // The background hover colour is added to help indicate that the whole row is clickable, rather
+  // than just the visible link text.
+  .govuk-task-list__item--with-link:hover {
+    background: $govuk-task-list-hover-colour;
+  }
+
+  // This uses table layout so that the task name and status always appear side-by-side, with the width of
+  // each 'column' being flexible depending upon the length of the task names and statuses.
+  .govuk-task-list__item-container {
```
owenatgov, line 35: How come this styling can't go on `govuk-task-list__item` instead?

src/govuk/components/task-list/_index.scss
```diff
@@ -0,0 +1,120 @@
...
+    display: table-cell;
+    padding-left: govuk-spacing(2);
+    text-align: right;
+    vertical-align: top;
+  }
+
+  // This adds an empty transparent box covering the whole row, including the task status and
+  // any hint text. Because this is generated within the link element, this allows the whole area
+  // to be clickable.
+  .govuk-task-list__task-link:after {
```
owenatgov, line 58: Naming nitpick: as we already have the context that this is part of the task list component and there aren't any other links as part of this component (?) could this be changed to:

```suggestion
  .govuk-task-list__link:after {
```

owenatgov (commented):
Finally gotten around to re-reviewing this.

I'm conscious that this is still a WIP and worry that my nitpicky comments could come across as a bit snipey. Take them all with a big scoop of salt. 

Huge props to Frankie and everyone for comnig together on this! Really feels like it's coming together.

src/govuk/components/task-list/task-list.yaml
frankieroberto: Opps! Fixed in b7d5ccbe82e36d1353b9a2ed9e571301a3254130, thanks!

src/govuk/components/task-list/task-list.yaml
frankieroberto: Well spotted! Fixed in 1388997e582b530fe2ffb45ae62d0335e5ee0694 thanks.

src/govuk/components/task-list/_index.scss
frankieroberto: Good idea! Fixed in e9d97ef2e4e6afdbf31c9e4e52ff5f37dd9577d6.

frankieroberto:
@owenatgov thanks for the follow-up review! (Sorry it's take a while to respond, I was on holiday).

And no worries about coming across snipey, these comments are great and I really value them! Let me know if you spot anything else!

src/govuk/components/task-list/_index.scss
```diff
@@ -0,0 +1,103 @@
+@include govuk-exports("govuk/component/task-list") {
+  .govuk-task-list {
+    @include govuk-font($size: 19);
+    margin-top: 0;
```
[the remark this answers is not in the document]
sussexrick: GOV.UK components generally have a bottom margin by default. I'd expect `@include govuk-responsive-margin(6, "bottom");` here.

src/govuk/components/task-list/template.njk
```diff
@@ -0,0 +1,28 @@
...
+    {% if item.hint %}
+      {%- set hintId = idPrefix + "-" + loop.index + "-hint" -%}
+    {% endif %}
+    {%- set statusId = idPrefix + "-" + loop.index + "-status" -%}
+    <li class="govuk-task-list__item{% if item.href %} govuk-task-list__item--with-link{% endif %}">
+      <span class="govuk-task-list__task-name-and-hint">
+        {% if item.href %}
+          <a class="govuk-link govuk-task-list__link" href="{{ item.href }}" aria-describedby="{{ hintId + " " if item.hint }}{{ statusId }}">{{ item.title.html | safe if item.title.html else item.title.text }}</a>
+        {% else %}
+          <div class="govuk-task-list__task-no-link">{{ item.title.html | safe if item.title.html else item.title.text }}</div>
```
sussexrick, line 13: `div` inside `span` is invalid HTML (for both `task-no-link` and `task-hint`)

src/govuk/components/task-list/_index.scss
```diff
@@ -0,0 +1,120 @@
...
+
+  .govuk-task-list__status--text-colour {
+    color: $govuk-text-colour;
+  }
+
+  .govuk-task-list__status--secondary-text-colour {
+    color: $govuk-secondary-text-colour;
+  }
+
+  .govuk-task-list__status--purple {
```
sussexrick, line 101: I would expect to see class names representing statuses (eg `--incomplete` and `--completed`) rather than colours.

sussexrick:
On the task list pattern page the task list summary (ie 'Application incomplete. You've completed x of x tasks.') is just an image. It would be useful if this was a code sample at least.

sussexrick:
On the task list pattern page these three links are all broken:

image

colinrotherham:
Hi @frankieroberto would you mind rebasing this PR?

It's missing some recent commits organising our `dependencies` from `devDependencies` for Heroku review apps. Don't think the PR has **“Allow edits and access to secrets by maintainers”** ticked so can't do it for you (sorry)

For now though I've pushed a temporary copy of your branch (but rebased) to → https://github.com/alphagov/govuk-frontend/tree/task-list-component which gets Heroku up and running again

frankieroberto:
@colinrotherham done. I've done a merge commit from main instead of a rebase for the time being, to avoid having to force-push and cause complications for anyone else who's checked out the branch.  We can rebase and squash the commits when we’re ready to merge. That ok?

I can't actually see where the "Allow edits and access to secrets by maintainers" checkbox is...?

src/govuk/components/task-list/_index.scss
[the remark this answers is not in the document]
frankieroberto: Thanks – I wasn’t sure what the convention was on this. Fixed in e67b84fb0bc429c4e0ef2eb86b98c744dec97dd9.

src/govuk/components/task-list/_index.scss
frankieroberto: @sussexrick Interesting! We deliberated a bit on this, and in the end decided to use colour based names to allow for maximum flexibility for services to use them as they see fit (without it clashing with the class name) but to give some recommendations of how to pair colours with labels in the guidance.

The exception was the red colour, which we decided to give the class name of `govuk-task-list__status--error` to more strongly discourage people from using it for anything other than an error.

All still open to discussion though!

src/govuk/components/task-list/template.njk
```diff
@@ -0,0 +1,28 @@
+{% set idPrefix = params.idPrefix if params.idPrefix else "task-list" %}
+<ul class="govuk-task-list">
+  {% for item in params.items %}
+    {% if item.hint %}
+      {%- set hintId = idPrefix + "-" + loop.index + "-hint" -%}
+    {% endif %}
+    {%- set statusId = idPrefix + "-" + loop.index + "-status" -%}
+    <li class="govuk-task-list__item">
+      <span class="govuk-task-list__task-name">
```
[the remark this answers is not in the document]
frankieroberto: Updated now

src/govuk/components/task-list/template.njk
frankieroberto: Updated now

src/govuk/components/task-list/task-list.yaml
```diff
@@ -0,0 +1,212 @@
+params:
```
simonwhatley, line 1: @frankieroberto I think you're missing the `hint` params from the docs.

src/govuk/components/task-list/template.njk
```diff
@@ -0,0 +1,28 @@
...
+          <div class="govuk-task-list__task-no-link">{{ item.title.html | safe if item.title.html else item.title.text }}</div>
+        {% endif %}
+        {% if item.hint %}
+          <div id="{{ hintId }}" class="govuk-task-list__task_hint">
+            {{ item.hint.html | safe if item.hint.html else item.hint.text }}
+          </div>
+        {% endif %}
+      </div>
+      <div class="govuk-task-list__status-container">
+        <span id="{{ statusId }}" class="govuk-task-list__status {{ item.status.classes }}">
```
simonwhatley, line 22:
```suggestion
        <span id="{{ statusId }}" class="govuk-task-list__status {%- if item.status.classes %} {{ item.status.classes }}{% endif %}">
```

src/govuk/components/task-list/template.njk
```diff
@@ -0,0 +1,28 @@
+{% set idPrefix = params.idPrefix if params.idPrefix else "task-list" %}
+<ul class="govuk-task-list">
+  {% for item in params.items %}
+    {% if item.hint %}
+      {%- set hintId = idPrefix + "-" + loop.index + "-hint" -%}
+    {% endif %}
+    {%- set statusId = idPrefix + "-" + loop.index + "-status" -%}
+    <li class="govuk-task-list__item{% if item.href %} govuk-task-list__item--with-link{% endif %}">
```
simonwhatley, line 8:
```suggestion
    <li class="govuk-task-list__item {%- if item.href %} govuk-task-list__item--with-link{% endif %}">
```

src/govuk/components/task-list/template.njk
```diff
@@ -0,0 +1,28 @@
...
+      <div class="govuk-task-list__task-name-and-hint">
+        {% if item.href %}
+          <a class="govuk-link govuk-task-list__link" href="{{ item.href }}" aria-describedby="{{ hintId + " " if item.hint }}{{ statusId }}">{{ item.title.html | safe if item.title.html else item.title.text }}</a>
```
simonwhatley, line 11:
```suggestion
          <a class="govuk-link govuk-task-list__link" href="{{ item.href }}" aria-describedby="{{ hintId if item.hint else statusId }}">{{ item.title.html | safe if item.title.html else item.title.text }}</a>
```

src/govuk/components/task-list/task-list.yaml
simonwhatley: For example:

```
  - name: hint.text
    type: string
    required: true
    description: If `html` is set, this is not required. Text to use within the hint. If `html` is provided, the `text` argument will be ignored.
  - name: hint.html
    type: string
    required: true
    description: If `text` is set, this is not required.
```

src/govuk/components/task-list/task-list.yaml
```diff
@@ -0,0 +1,212 @@
...
+  description: Array of item objects.
+  params:
+  - name: classes
+    type: string
+    required: false
+    description: Classes to add to the item `div`
+  - name: title.text
+    type: string
+    required: true
+    description: If `html` is set, this is not required. Text to use within the each title. If `html` is provided, the `text` argument will be ignored.
```
simonwhatley, line 14:
```suggestion
    description: If `html` is set, this is not required. Text to use within the title. If `html` is provided, the `text` argument will be ignored.
```

src/govuk/components/task-list/task-list.yaml
```diff
@@ -0,0 +1,212 @@
...
+    required: true
+    description: If `text` is set, this is not required.
+  - name: title.classes
+    type: string
+    required: false
+    description: Classes to add to the title wrapper
+  - name: status.text
+    type: string
+    required: true
+    description: If `html` is set, this is not required. Text to use within the each value. If `html` is provided, the `text` argument will be ignored.
```
simonwhatley, line 26:
```suggestion
    description: If `html` is set, this is not required. Text to use within the value. If `html` is provided, the `text` argument will be ignored.
```

src/govuk/components/task-list/task-list.yaml
```diff
@@ -0,0 +1,212 @@
...
+  - name: status.html
+    type: string
+    required: true
+    description: If `text` is set, this is not required. HTML to use within the each value. If `html` is provided, the `text` argument will be ignored.
```
simonwhatley, line 30:
```suggestion
    description: If `text` is set, this is not required. HTML to use within the value. If `html` is provided, the `text` argument will be ignored.
```

src/govuk/components/task-list/task-list.yaml
```diff
@@ -0,0 +1,212 @@
...
+  - name: status.classes
+    type: string
+    required: false
+    description: Classes to add to the status wrapper
+  - name: url.text
+    type: string
+    required: false
+    description: If `html` is set, this is not required. Text to use within the each title. If `html` is provided, the `text` argument will be ignored.
```
simonwhatley, line 38:
```suggestion
    description: If `html` is set, this is not required. Text to use within the title. If `html` is provided, the `text` argument will be ignored.
```

src/govuk/components/task-list/task-list.yaml
```diff
@@ -0,0 +1,212 @@
...
+  - name: url.html
+    type: string
+    required: false
+    description: If `text` is set, this is not required. HTML to use within the each value. If `html` is provided, the `text` argument will be ignored.
```
simonwhatley, line 42:
```suggestion
    description: If `text` is set, this is not required. HTML to use within the value. If `html` is provided, the `text` argument will be ignored.
```

src/govuk/components/task-list/task-list.yaml
frankieroberto: Opps, well spotted! 🙌

src/govuk/components/task-list/template.njk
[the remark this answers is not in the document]
frankieroberto: @simonwhatley we debated this quite a bit. The thinking was that we wanted to deviate from the existing Tag component styles (ie uppercase text with bright blue background) as there was a fair amount of concern (and some research evidence) that the tag component looked too clickable when used within the task list pattern, as the design is quite close to the button design, and that the uppercase text may be less accessible.

There are a few different approaches we could take:

1. also update the styling of the Tag component at the same time 
2. use the Tag component but add some extra modifiers to completely change the style
3. not use the tag component at all, and instead treat it as a custom 'status' element of the Task list

We went with 3 mainly for pragmatism, to avoid having to make backwards-incompatible changes to the Tag component (which is also used by the Phase banner component). The second option feels a bit pointless to me, as we’d be changing so much of the styling it wouldn't share that much in common...?

It’s also worth noting that one of the styles for the status component is just plain black text (for "Complete") - which differs quite a lot from the Tag component.

The downside with option 3 though is that we’d have to live with some visual inconsistency between the Tag component and the status bit of the task list component, at least in the short term whilst any changes to the Tag component are debated...

See [previous comment on this](https://github.com/alphagov/govuk-frontend/pull/2261/files#r837395296).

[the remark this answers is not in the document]
frankieroberto: @philsherry yep - we discussed that design (and the blog posts) a fair bit!  It helped convince us to both make the statuses look less clickable AND to make them actually clickable. Belt and braces... 😄

src/govuk/components/task-list/template.njk
```diff
@@ -0,0 +1,28 @@
...
+          <div class="govuk-task-list__task-no-link">{{ item.title.html | safe if item.title.html else item.title.text }}</div>
+        {% endif %}
+        {% if item.hint %}
+          <div id="{{ hintId }}" class="govuk-task-list__task_hint">
+            {{ item.hint.html | safe if item.hint.html else item.hint.text }}
+          </div>
+        {% endif %}
+      </div>
+      <div class="govuk-task-list__status-container">
+        <span id="{{ statusId }}" class="govuk-task-list__status {{ item.status.classes }}">
```
owenatgov, line 22: A quick question about the use of `span` here that was prompted by the [tag redesign](https://github.com/alphagov/govuk-frontend/pull/3502). The tag currently uses a `strong` as its containing element. The history around why this was chosen is very wooly as it was pulled from the legacy frameworks but the tea are confident that this is for progressive enhancement reasons to maintain some form of visual delegation when css isn't available ie: if you turn css off, the tag will still be bold. If you were to use the redesigned tag component, this would turn into a `strong`.

My assumption here is that a `span` was chosen here just because it semantically makes sense and swapping it out for a `strong` won't disrupt the implementation (they're both inline elements so it doesn't make a difference) but I wanted to prompt on that to make sure this wasn't going to be disruptive. FWIW my personal take is that plonking the tag in as a `strong` is perfectly fine and we don't need to make any destructive changes to the markup of the tag component at present.

frankieroberto: @owenatgov hi - yep I swapped it to a span just because it no longer visually appeared bold.

However with our decision to also update the Tag component, I'm going to update the Task list component to actually use the Tag component for the statuses (as a sub-component), so it'll swap to using a `<strong>` element. I'd agree with your take that that's fine, and possibly even preferable where the CSS isn't available.

colinrotherham:
Afternoon 👋

Just a heads up. We've moved our source code into [**npm workspaces**](https://docs.npmjs.com/cli/v9/using-npm/workspaces?v=true) package directories via https://github.com/alphagov/govuk-frontend/commit/102601ec1d7a7294b601e7f0c1141e785c0207dd

`src` → [`packages/govuk-frontend/src`](https://github.com/alphagov/govuk-frontend/tree/main/packages/govuk-frontend/src)
`app/src` → [`packages/govuk-frontend-review/src`](https://github.com/alphagov/govuk-frontend/tree/main/packages/govuk-frontend-review/src)

Hopefully it's not too much of a tricky `git rebase` but let us know if you need a hand

### Publishing a pre-release or running `npm link`
If you're linking `govuk-frontend` locally into another project (or `git filter-branch`) also note:

`package` → [`packages/govuk-frontend-review/dist`](https://github.com/alphagov/govuk-frontend/tree/main/packages/govuk-frontend-review/dist)

frankieroberto:
Now that the [Tag design changes have been merged into main](https://github.com/alphagov/govuk-frontend/pull/3502), I’ve rebased this branch from main, so it now uses the new Tag design. cc @owenatgov @colinrotherham

Is there anything left to do?

packages/govuk-frontend/src/govuk/components/task-list/accessibility.test.mjs
```diff
@@ -0,0 +1,22 @@
...
+    beforeAll(async () => {
+      exampleNames = Object.keys(await getExamples('task-list'))
+    })
+
+    it('passes accessibility tests', async () => {
+      for (const name of exampleNames) {
+        const exampleName = name.replace(/ /g, '-')
+
+        // Navigation to example, create report
+        await goToComponent(page, 'task-list', { exampleName })
```
colinrotherham, line 17: Thanks for adding these new tests 😊

frankieroberto: Looks like they've started failing though (since the merge from main) for some reason. Will investigate!

packages/govuk-frontend/src/govuk/components/task-list/template.test.js
```diff
@@ -0,0 +1,73 @@
+/**
+ * @jest-environment jsdom
```
colinrotherham, line 2: Can get rid of all these lines now 😊

```js
/**
 * @jest-environment jsdom
 */
/* eslint-env jest */
```

We automatically match filenames for Jest and ESLint environments

packages/govuk-frontend/src/govuk/components/task-list/template.test.js
```diff
@@ -0,0 +1,73 @@
...
+    it('associates the task name link with the status using aria', async () => {
+      const $ = render('task-list', examples.default)
+
+      const $itemLink = $('.govuk-task-list__link')
+      expect($itemLink.get(0).tagName).toEqual('a')
+      expect($itemLink.attr('href')).toEqual('#')
+
+      expect($itemLink.attr('aria-describedby')).toEqual('task-list-example-1-status')
+
+      const $statusWithId = $('#' + $itemLink.attr('aria-describedby'))
```
colinrotherham, line 39: Just looking into this test failure for you

We have type checking enabled and Cheerio expects selector strings to match:

```js
type SelectorType = `${SelectorSpecial}${AlphaNumeric}${string}` | `${AlphaNumeric}${string}`;
```

Otherwise it can't infer the element type and thinks `.tagName` might not exist 🤦‍♂️

Give this a try? It keeps the selector as a string but still checks `aria-describedby` matches

```js
const statusId = 'task-list-example-1-status'
expect($itemLink.attr('aria-describedby')).toEqual(statusId)

const $statusWithId = $(`#${statusId}`)
expect($statusWithId.get(0).tagName).toEqual('div')
```

packages/govuk-frontend/src/govuk/components/task-list/accessibility.test.mjs
colinrotherham: For the Cheerio type checks in `template.test.js` [try these ideas](https://github.com/alphagov/govuk-frontend/pull/2261#discussion_r1203793390)

You can run them locally with:

```console
npm run build:types
```

Or for accessibility tests in particular we find [opening the browser](https://github.com/alphagov/govuk-frontend/blob/main/docs/releasing/testing-and-linting.md#component-tests) is helpful:

```console
HEADLESS=false npx jest task-list/accessibility.test.mjs --watch
```

But for tests in a loop like these you'll need to pause Jest or the browser will close:

```js
await expect(axe(page)).resolves.toHaveNoViolations()
await jestPuppeteer.debug() // Pause the test run for debugging
```

colinrotherham:
Thanks @frankieroberto

I've added some comments and @owenatgov's already picked this up

packages/govuk-frontend/src/govuk/components/task-list/template.test.js
colinrotherham: You can run the type checks locally with:

```console
npm run build:types
```

packages/govuk-frontend/src/govuk/components/task-list/template.test.js
frankieroberto: Done, thanks.

packages/govuk-frontend/src/govuk/components/task-list/template.test.js
frankieroberto: @colinrotherham that didn't fix it, I got the same error. But changing from ``$(`#${statusId}`)`` to ``$component.find(`#${statusId}`)`` does. I have no idea why. 🤷

packages/govuk-frontend/src/govuk/components/task-list/template.njk
```diff
@@ -0,0 +1,36 @@
...
+    {% if item.hint %}
+      {%- set hintId = idPrefix + "-" + loop.index + "-hint" -%}
+    {% endif %}
+    {%- set statusId = idPrefix + "-" + loop.index + "-status" -%}
+    <li class="govuk-task-list__item {%- if item.href %} govuk-task-list__item--with-link{% endif %}">
+      <div class="govuk-task-list__task-name-and-hint">
+        {% if item.href %}
+          <a class="govuk-link govuk-task-list__link" href="{{ item.href }}" aria-describedby="{{ hintId + " " if item.hint }}{{ statusId }}">{{ item.title.html | safe if item.title.html else item.title.text }}</a>
+        {% else %}
+          <div class="govuk-task-list__task-no-link">{{ item.title.html | safe if item.title.html else item.title.text }}</div>
```
owenatgov, line 15: Separately, this `div` doesn't include the `aria-desscribedby` setup. I can't remember 100% but I suspect this is because `aria-describedby` doesn't like non-interactive elements. Do we need ot think about the `aria-describedby` relationship?

packages/govuk-frontend/src/govuk/components/task-list/template.njk
```diff
@@ -0,0 +1,36 @@
...
+        {% endif %}
+        {% if item.hint %}
+          <div id="{{ hintId }}" class="govuk-task-list__task_hint">
+            {{ item.hint.html | safe if item.hint.html else item.hint.text }}
+          </div>
+        {% endif %}
+      </div>
+      <div class="govuk-task-list__status {{ item.status.classes }}" id="{{ statusId}}">
```
owenatgov, line 23: The most pedantic that ever pedanticed

```suggestion
      <div class="govuk-task-list__status {{ item.status.classes }}" id="{{ statusId }}">
```

packages/govuk-frontend/src/govuk/components/task-list/task-list.yaml
```diff
@@ -0,0 +1,222 @@
...
+        description: If `text` is set, this is not required.
+      - name: title.classes
+        type: string
+        required: false
+        description: Classes to add to the title wrapper
+      - name: tag.text
+        type: string
+        required: true
+        description: If `html` is set, this is not required. Text to use within the value. If `html` is provided, the `text` argument will be ignored.
+      - name: tag.html
```
owenatgov, line 27: These docs are out of date. Below, you include a `status` attribute where you store either the tag macro data or the flat data for the status `div`. The `status` attribute isn't documented here, and this documentation reflections a previous implementation where the `tag` attribute include the information for rendering the status which used the new tag design.

packages/govuk-frontend/src/govuk/components/task-list/task-list.yaml
```diff
@@ -0,0 +1,222 @@
...
+  - name: items
+    type: array
+    required: true
+    description: Array of item objects.
+    params:
+      - name: classes
+        type: string
+        required: false
+        description: Classes to add to the item `div`
+      - name: title.text
```
owenatgov, line 11: Whilst we do use the parent.child notation for documenting objects and specs for arrays of objects elsewhere in govuk-frontend, this is complex enough that I'm finding it quite difficult to ingest. Could we please change it to be indented by object?

owenatgov (commented):
A few comments, mainly on the yaml docs. We're looking strong.

I ran a quick test against VoiceOver and found that the `aria-describedby` setup wasn't working at all. The screen reader would declare the list item number, that it was a link and then I would have to step down the accessibility tree to hear the status and the hint, both the latter announced without any context.

I think this is fine on the surface as context is provided by the fact the user is within a list item, although I can't work out why `describedby` isn't working. I'll investigate this more. Generally I'm not 100% sold on using `describedby` here, mainly because I'm fairly confident you can't plug more than one `id` into it and it means the content of the thing being `describedby` is lost. As said above, I feel like letting the user step through the content of the list item is fine, but I'll verify that with the team.

frankieroberto:
> I ran a quick test against VoiceOver and found that the `aria-describedby` setup wasn't working at all. The screen reader would declare the list item number, that it was a link and then I would have to step down the accessibility tree to hear the status and the hint, both the latter announced without any context.
> 
> I think this is fine on the surface as context is provided by the fact the user is within a list item, although I can't work out why `describedby` isn't working. I'll investigate this more. Generally I'm not 100% sold on using `describedby` here, mainly because I'm fairly confident you can't plug more than one `id` into it and it means the content of the thing being `describedby` is lost. As said above, I feel like letting the user step through the content of the list item is fine, but I'll verify that with the team.

Thanks for testing this. I think we mostly adopted the `aria-describedby` setup from [the existing code in the prototyping kit](https://github.com/alphagov/govuk-prototype-kit-task-list/blob/main/templates/task-list.html) - does it work ok there? I think the only thing we added was the option for the hint text.

The ARIA specification seems to suggest that [multiple IDs are allowed](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-describedby) but maybe that isn’t supported in practice?

owenatgov:
So, after conferring with the team and re-testing, I was mistaken 🙈 

`describedby` isn't actually intended to be announced when navigating using the arrow keys and only activates when navigating through sequentially focusable items on a page aka; navigating with the tab key. The basis for this is that you can get the context of the link being described by using the arrow keys to navigate to the surrounding content, which you should include nearby the link itself (which we do). The `describedby` is for when users are tabbing through links on a page and will therefore possibly miss the contextual content which we provide in `describedby`.

Re-testing with this in mind against NVDA and JAWS as well as VoiceOver, everything works as expected 🎉 I'm therefore happy to call this little issue resolved.

packages/govuk-frontend/src/govuk/components/task-list/task-list.yaml
```diff
@@ -0,0 +1,222 @@
...
+            text: Registered company details
+          href: '#'
+          status:
+            tag:
+              text: Not started
+              classes: govuk-tag--light-blue
+        - title:
+            text: Business plan
+          href: '#'
+          hint:
```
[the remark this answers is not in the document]
frankieroberto: Fixed now, thanks!

packages/govuk-frontend/src/govuk/components/task-list/template.njk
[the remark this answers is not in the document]
frankieroberto: Thanks, I've removed this. I still needed a way to find that item within the tests, so I’ve added a custom class (using the `app-` prefix) for now. As a bonus, this also tests that the custom classes work, which they previously weren’t as I’d forgotten to add them to the template.

packages/govuk-frontend/src/govuk/components/task-list/task-list.yaml
frankieroberto: @owenatgov  Thanks - this all needed updating. I've updated it now to show that `tag` is within `status`. It’s a bit tricky to describe which things are required or not, as the tag overrides the status `text` or `html` if present.

Could you take another look and let me know if what I’ve added makes sense?

frankieroberto:
@owenatgov I’ve updated the param documentation, which was well out-of-date. Would you mind taking another look? I couldn't see any easy way to preview how this documentation would appear within the Design System website itself - I guess that'd need another pre-release?

I also updated the template to allow custom classes to be added at a few different levels: the `<ul>`, the `<li>` and the `<a>` or `<div>` containing the title text. They can also be added to the tag and the status container. The only elements which don’t currently accept custom classes are `<div class="govuk-task-list__task-name-and-hint">` (container for the title + hint) and the hint element (`<div class="govuk-task-list__task_hint">`). Do you think those ought to support custom classes too, or is that too much?

packages/govuk-frontend/src/govuk/components/task-list/template.njk
```diff
@@ -0,0 +1,36 @@
...
+    {% if item.hint %}
+      {%- set hintId = idPrefix + "-" + loop.index + "-hint" -%}
+    {% endif %}
+    {%- set statusId = idPrefix + "-" + loop.index + "-status" -%}
+    <li class="govuk-task-list__item {%- if item.href %} govuk-task-list__item--with-link{% endif %}{%- if item.classes %} {{ item.classes }}{% endif %}">
+      <div class="govuk-task-list__task-name-and-hint">
+        {% if item.href %}
+          <a class="govuk-link govuk-task-list__link {%- if item.title.classes %} {{ item.title.classes }}{% endif %}" href="{{ item.href }}" aria-describedby="{{ hintId + " " if item.hint }}{{ statusId }}">{{ item.title.html | safe if item.title.html else item.title.text }}</a>
+        {% else %}
+          <div class="{%- if item.title.classes %} {{ item.title.classes }}{% endif %}">{{ item.title.html | safe if item.title.html else item.title.text }}</div>
```
owenatgov, line 15: This could result in an empty `class` attribute

```suggestion
          <div{% if item.title.classes %} class="{{ item.title.classes }}" {%- endif %}>{{ item.title.html | safe if item.title.html else item.title.text }}</div>
```

owenatgov (commented):
A few more comments. Most of them fairly petty code hygiene things. We're very close.

packages/govuk-frontend/src/govuk/components/task-list/template.njk
```diff
@@ -0,0 +1,36 @@
+{% from "../tag/macro.njk" import govukTag -%}
+
+{% set idPrefix = params.idPrefix if params.idPrefix else "task-list" %}
+<ul class="govuk-task-list {%- if params.classes %} {{ params.classes }}{% endif %}"{% for attribute, value in params.attributes %} {{attribute}}="{{value}}"{% endfor %}>
+  {% for item in params.items %}
+    {% if item.hint %}
+      {%- set hintId = idPrefix + "-" + loop.index + "-hint" -%}
```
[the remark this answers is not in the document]
frankieroberto: I wondered how the accordion does this, and it seems to take the opposite approach of not setting variables for the IDs at all, and instead just [repeating the same logic inline](https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/src/govuk/components/accordion/template.njk) within both the `id=""` and `aria-labelledby=""` attributes...

I can't tell which approach is better? Arguably repeating the logic is more readable, but setting the variables feels less likely to get accidentally broken (although I guess that's what tests are for?).... 🤷

frankieroberto:
@owenatgov:
> A few more comments. Most of them fairly petty code hygiene things. We're very close.

Thanks - fixed them all I think, apart from discussion around how to set the IDs.  Let me know if you spot anything else!

packages/govuk-frontend/src/govuk/components/task-list/template.njk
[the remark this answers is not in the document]
owenatgov: I think you've answered this with your parenthesised comment. I'll drop this as a suggestion and double check that the tests cover this functionality.

packages/govuk-frontend/src/govuk/components/task-list/_index.scss
```diff
@@ -0,0 +1,72 @@
+@include govuk-exports("govuk/component/task-list") {
+  $govuk-task-list-focus-colour: govuk-colour("black");
+  $govuk-task-list-focus-background-colour: govuk-colour("yellow");
```
36degrees, line 3: These aren't used anywhere in the code.

packages/govuk-frontend/src/govuk/components/task-list/_index.scss
```diff
@@ -0,0 +1,72 @@
...
+
+  .govuk-task-list {
+    @include govuk-font($size: 19);
+    margin-top: 0;
+    @include govuk-responsive-margin(6, "bottom");
+    padding: 0;
+    list-style-type: none;
+    @include govuk-media-query($from: tablet) {
+      min-width: 550px;
+    }
```
36degrees, line 14: This can force the task list to overflow it's container, which can cause issues if for example it's used in grid column where there is anything else in the same row.

Screenshot 2023-06-08 at 11 03 53

owenatgov (commented):
Cheers @frankieroberto!

There are a few tiny bits left from me:

- The `hint` param in the docs is still missing a description
- We're getting test failures which I suspect can be fixed with a rebase

I'm going to open this up to the team to get some extra opinions. Once they've looked at it, could we also please squash the commits? I think one is fine here again, like the tag redesign.

packages/govuk-frontend/src/govuk/components/task-list/task-list.yaml
```diff
@@ -0,0 +1,250 @@
...
+        params:
+          - name: text
+            type: string
+            required: true
+            description: Text to use within the hint. If `html` is provided, the `text` argument will be ignored.
+          - name: html
+            type: string
+            required: true
+            description: HTML to use within the hint. If `html` is provided, the `text` argument will be ignored.
+      - name: status
```
36degrees, line 36: `item.status.classes` is missing from the docs

https://github.com/alphagov/govuk-frontend/blob/2bb6d8e154454b84b9bf50d9348a8c4cecded5ab/packages/govuk-frontend/src/govuk/components/task-list/template.njk#L23

packages/govuk-frontend/src/govuk/components/task-list/task-list.yaml
```diff
@@ -0,0 +1,250 @@
...
+                description: Classes to add to the tag.
+          - name: text
+            required: false
+            type: string
+            description: Text to use for the status, as an alternative to using a tag. If `html` or `tag` is provided, the `text` argument will be ignored.
+          - name: html
+            required: false
+            type: string
+            description: HTML to use for the status, as an alternative to using a tag. If `html` or `tag` is provided, the `text` argument will be ignored.
+      - name: url
```
36degrees, line 66: I believe this should be `href`:

https://github.com/alphagov/govuk-frontend/blob/2bb6d8e154454b84b9bf50d9348a8c4cecded5ab/packages/govuk-frontend/src/govuk/components/task-list/template.njk#L10-L12

packages/govuk-frontend/src/govuk/components/task-list/_index.scss
36degrees: I'd also suggest adding a full-page example to the review app that shows how the task list works in the context of a page, which would make it easier to catch things like this.

packages/govuk-frontend/src/govuk/components/task-list/_index.scss
```diff
@@ -0,0 +1,72 @@
+@include govuk-exports("govuk/component/task-list") {
```
36degrees, line 1: Task names that aren't links and statuses that aren't tags do not currently have a text colour set on them, which means they're using the default user agent foreground colour.

This is usually black – which would mean that these elements are `#000000` rather than GOV.UK's `#0b0c0c`, but the default text colour can be overridden by the user:

Screenshot 2023-06-08 at 11 50 52

I think this would fail WCAG based on [Failure 24](https://www.w3.org/TR/WCAG20-TECHS/F24.html).

The easiest fix is either to `@include govuk-text-colour;` on both `.govuk-task-list__task-name-and-hint` and `govuk-task-list__status`, or on the root `.govuk-task-list`.

36degrees (commented):
This is looking really good, but there are a few minor things I think need addressing.

frankieroberto:
@36degrees thanks for spotting all these! I'll get onto them tomorrow!

packages/govuk-frontend/src/govuk/components/task-list/task-list.yaml
```diff
@@ -0,0 +1,250 @@
...
+              text: Error
+              classes: govuk-tag--red
+        - classes: app-task-list__item--no-link
+          title:
+            text: Payment
+          hint:
+            text: It will cost between £15 and £75
+          status:
+            text: Cannot start yet
+            classes: govuk-hint
```
36degrees, line 192: Don't use a class from another component that's not being imported, because:

1. there's no guarantee that the user will be including the styles for the hint component in the future
2. we might change the styles for the hint component in the future (for example, adding a margin, or making other changes) that would affect this component as well.

Suggest instead adding a modifier class:

```scss
.govuk-task-list__status--muted { // or similar, maybe .govuk-task-list__status--secondary
  color: $govuk-secondary-text-colour;
}
```

packages/govuk-frontend/src/govuk/components/task-list/task-list.yaml
frankieroberto: 😳

packages/govuk-frontend/src/govuk/components/task-list/_index.scss
frankieroberto: Opps! 😳 

Good idea about adding a full page example to the review app!

36degrees (commented):
Noticed something else, sorry!

packages/govuk-frontend/src/govuk/components/task-list/task-list.yaml
frankieroberto: Good point! I think the `govuk-hint` class is widely abused for non-field-hints (we might even do that in production, opps), but we should defo avoid it here.

How about `--cannot-start-yet` as the modifier, if we want to be more prescriptive? I don’t think we've uncovered any other use-cases for grey text in the status?

packages/govuk-frontend/src/govuk/components/task-list/_index.scss
frankieroberto: Opps, thanks!

packages/govuk-frontend/src/govuk/components/task-list/task-list.yaml
frankieroberto: Added, thanks!

packages/govuk-frontend/src/govuk/components/task-list/task-list.yaml
frankieroberto: Added now, thanks!

packages/govuk-frontend/src/govuk/components/task-list/_index.scss
frankieroberto: Thanks - added `@include govuk-text-colour;` to both `govuk-task-list__task-name-and-hint` and `govuk-task-list__status`.

frankieroberto:
@36degrees I think I’ve made all the changes - thanks for spotting them all! 👏 

Would you mind approving the test workflows, and checking and re-reviewing the code? Thanks!

36degrees (approved):
This looks good – thanks for all your hard work on this @frankieroberto 🙌🏻 

A few final things to sort:

- We need to add a changelog entry to tell users know about this exciting new component!
- Ideally we'd improve the commit structure, ideally reducing to a small number of meaningful commits and removing the merges. Is that something you'd be comfortable doing? I'm happy to do this for you if you'd prefer.
- It'd be good to add some more tests to cover the various different macro options to ensure they're working as intended. This is a little laborious, so I'm happy to create an issue for our team to do this in a follow up PR unless you're particularly keen to be involved?

36degrees:
> It'd be good to add some more tests to cover the various different macro options to ensure they're working as intended. This is a little laborious, so I'm happy to create an issue for our team to do this in a follow up PR unless you're particularly keen to be involved?

Discussed this with @frankieroberto and have raised an issue for us to do this (#3828).

I believe the other points have been addressed. This just needs a second review and then we can get it merged – @owenatgov would you be happy to take another look?

PR #2262, opened 2021-06-24
Clarify which npm user to login with for releases

vanitabarrett:
## What
Make it clearer in the release docs which npm user we should be logging in as.

## Why
The original documentation linked to the relevant user/folder in the team credentials repo, but it's easy to miss and choose a different user (e.g: alphagov) by mistake.


lfdebrux (approved):
That's very helpful, thank you!

PR #2264, opened 2021-07-01
Improve focus state for radio and checkbox controls in forced color mode (for example Windows High Contrast Mode)

### LINKED ISSUE #2258: Radios and checkboxes do not follow native colours in Windows High Contrast
## Description of the issue
There seems to be a discrepancy between how native radios are displayed in Windows High Contrast and how GOVUK ones are - the highlight colour used on native inputs (including the text boxes in the Design System) is missing from the focus and selected states of radios (and checkboxes), probably down to the way in which the pseudo control is built. This makes the focus state somewhat less visible than the native version.

Screenshot shows Design System with focussed (top), selected and unselected/unfocussed vs native elements on the right (focussed at top, selected below).
Screenshot 2021-06-24 at 08 06 01

Screenshot to show Design System text input focus state
Screenshot 2021-06-24 at 10 15 48

Open https://design-system.service.gov.uk/components/radios/ in MS Edge on Windows and enable Windows High Contrast Mode

Expectation is that the Design System radios and checkboxes to follow the native colouring in line with the Design System's text inputs, but they use white instead of the blue highlight which makes them much less obvious as the default state is also white.

- Operating system: Windows 10
- Browser: MS Edge
- Browser version: 91.0.864.54
- GOV.UK Frontend Version: 3.12.0

36degrees:
The focus state for radios and checkboxes is currently not very visible in high contrast mode or when colours are overridden in Firefox – the only visible change is that the outer border of the radio or checkbox control thickens from 2px to 4px. This is the same as the part of the focus state change that meets contrast, but omits the yellow ‘ring’ created by the box shadow.

Add an additional transparent outline to the control. In forced colors mode, transparent borders become solid (matching the text / foreground color) and so we get an additional outline around the focused control making it clearly visible.

When explicitly in forced-color mode, we can enhance the outline further by using the Highlight system color, which matches the color used by the focus state of native controls – at least in in Edge / Chrome / Opera which are the only other browsers to support forced color mode.

Fixes #2258

| Browser | Before | After |
|--|--|--|
| Internet Explorer 11 | ie-before | ie-after |
| Chrome | chrome-before | chrome-after |
| Edge | edge-before | edge-after
| Firefox (High Contrast Mode) | ff-hcm-before | ff-hcm-after
| Firefox ([Custom Colours](https://support.mozilla.org/en-US/kb/change-fonts-and-colors-websites-use)) | ff-custom-before | ff-custom-after |

A native focused text input in Edge in High Contrast Mode, for comparison:

Screenshot 2021-07-01 at 17 45 05

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+### Fixes
+
+We’ve made fixes to GOV.UK Frontend in the following pull requests:
+
+- [#2264: Improve focus state for radio and checkbox controls in forced color mode (for example Windows High Contrast Mode)](https://github.com/alphagov/govuk-frontend/pull/2264) – thanks to [@adamliptrot-oc](https://github.com/adamliptrot-oc) for reporting this issue.
+
 ## 3.13.0 (Feature release)
 
 ### New features
--- a/src/govuk/components/checkboxes/_index.scss
+++ b/src/govuk/components/checkboxes/_index.scss
@@ -122,6 +122,20 @@
   // Focused state
   .govuk-checkboxes__input:focus + .govuk-checkboxes__label:before {
     border-width: 4px;
+
+    // When colours are overridden, the yellow box-shadow becomes invisible
+    // which means the focus state is less obvious. By adding a transparent
+    // outline, which becomes solid (text-coloured) in that context, we ensure
+    // the focus remains clearly visible.
+    outline: $govuk-focus-width solid transparent;
+    outline-offset: 1px;
+
+    // When in an explicit forced-color mode, we can use the Highlight system
+    // color for the outline to better match focus states of native controls
+    @media screen and (forced-colors: active) {
+      outline-color: Highlight;
+    }
+
     box-shadow: 0 0 0 $govuk-focus-width $govuk-focus-colour;
   }
 
--- a/src/govuk/components/radios/_index.scss
+++ b/src/govuk/components/radios/_index.scss
@@ -120,6 +120,20 @@
   // Focused state
   .govuk-radios__input:focus + .govuk-radios__label:before {
     border-width: 4px;
+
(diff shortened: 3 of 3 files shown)


vanitabarrett (approved):
Seems like a sensible change. Tested in Windows and Firefox high contrast modes and works well.

vanitabarrett:
@36degrees Can we add a Changelog entry for this?

EoinShaughnessy:
Is it 'forced color mode' or 'forced colors mode'? I thought it might be the latter because 'forced colors mode' is the one I've seen more of online. For example, in [Chrome](https://www.chromestatus.com/feature/5757293075365888) and [Mozilla](https://developer.mozilla.org/en-US/docs/Web/CSS/forced-color-adjust). If it _is_ the latter, then we probably need to tweak the title of this ticket and any related tickets...

36degrees:
Good catch!

The Windows blog post (https://blogs.windows.com/msedgedev/2020/09/17/styling-for-windows-high-contrast-with-new-standards-for-forced-colors/) uses both terms interchangeably, but 'forced color mode' more often – which I think influenced my usage of the term originally.

However, I think the canonical resource to refer to for this is the [CSS Color Adjustment Module Level 1](https://www.w3.org/TR/css-color-adjust-1/) specification, which includes ['forced **colors** mode'](https://www.w3.org/TR/css-color-adjust-1/#forced-colors-mode) as a defined term.

EoinShaughnessy:
Thanks, @36degrees. I've updated this PR's title, along with those for [#2265](https://github.com/alphagov/govuk-frontend/pull/2265),  [#2273](https://github.com/alphagov/govuk-frontend/pull/2273) and [#2277](https://github.com/alphagov/govuk-frontend/pull/2277). I'll also update the Changelog entries when I raise a PR for the notes.

PR #2265, opened 2021-07-02
Don’t remove focus outline from disabled link buttons in forced color mode

### LINKED ISSUE #2127: SortSite flags removal of focus outline from disabled buttons
## Description of the issue

SortSite flags the way we remove the outline from the focus state of disabled button as a potential issue:

> The CSS outline or border style on this element makes it difficult or impossible to see the dotted link focus outline.
> Change the style to avoid obscuring the focus outline around focusable elements.
> Using a border or outline style that obscures the focus ring causes problems for keyboard-only users in:
> 
> - Chrome (obscures the focus indicator on links, buttons, dropdowns and range controls)
> - Firefox (obscures the focus indicator on links)
> - Internet Explorer (obscures the focus indicator on links, buttons and range controls)
> - Internet Explorer does not support the `outline-offset:` property, so using this does not reliably avoid focus outline overlap.
>
> Line 375 [WCAG 2.1 AA F78](https://www.w3.org/TR/WCAG20-TECHS/F78.html) [Section 508 (2017) AA F78](https://www.w3.org/TR/WCAG20-TECHS/F78.html)
> ```
> 375: .govuk-button--disabled:focus,.govuk-button[disabled="disabled"]:focus,.govuk-button[disabled]:focus{outline:none}
> ```

When the disabled attributed is used on a `<button>` the button should no longer be focusable, so I don't think this would be an issue.

However, if the `govuk-button--disabled` class is used on a link this could be problematic as the link might still be focusable but the focus indicator would not be visible.

36degrees:
SortSite flags this as a potential issue, referencing [WCAG 2.0 F78][1]:

> The CSS outline or border style on this element makes it difficult or impossible to see the dotted link focus outline.

We apply a 3px transparent outline to focused buttons, which becomes the visible focus indicator in forced color mode when the other focus state changes to the background and box shadow are not visible.

Adding the `--disabled` modifier to link buttons (`<a>` elements styled as buttons) does not remove them from the tab index or make them non-interactive. They are still focusable and outside of forced color mode retain the rest of the focus state. Removing the outline means that users in High Contrast Mode who focus the 'disabled' link button will not have a visible focus indicator.

This change should not affect buttons, as disabled buttons are not focusable and so the `:focus` pseudo-selector should never apply to them.

Outside of forced color mode there should be no visible change, as the outline is transparent.

## Before

Screenshot 2021-07-09 at 10 15 45

http://govuk-frontend-review.herokuapp.com/components/button/link-disabled/preview

## After

Screenshot 2021-07-09 at 10 16 06

https://govuk-frontend-pr-2265.herokuapp.com/components/button/link-disabled/preview

Fixes #2127

[1]: https://www.w3.org/TR/WCAG20-TECHS/F78.html

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 We’ve made fixes to GOV.UK Frontend in the following pull requests:
 
 - [#2264: Improve focus state for radio and checkbox controls in forced color mode (for example Windows High Contrast Mode)](https://github.com/alphagov/govuk-frontend/pull/2264) – thanks to [@adamliptrot-oc](https://github.com/adamliptrot-oc) for reporting this issue.
+- [#2265: Don’t remove focus outline from disabled link buttons in forced color mode](https://github.com/alphagov/govuk-frontend/pull/2265)
 
 ## 3.13.0 (Feature release)
 
--- a/src/govuk/components/button/_index.scss
+++ b/src/govuk/components/button/_index.scss
@@ -158,10 +158,6 @@
       cursor: default;
     }
 
-    &:focus {
-      outline: none;
-    }
-
     &:active {
       top: 0;
       box-shadow: 0 $button-shadow-size 0 $govuk-button-shadow-colour; // s0


36degrees:
Wondering if we should also deprecate the `--disabled` modifier class (leaving the attribute selectors) – although that's not something we've done before I don't think – and create an issue to remove them in a future breaking release?

Styling link buttons to look disabled without being able to effectively make them non-interactive and non-focusable seems like a bad idea generally.

36degrees:
Written up an issue to remove the ability to style link buttons as disabled: https://github.com/alphagov/govuk-frontend/issues/2271

vanitabarrett:
@36degrees Can you add this to the changelog and tag the issue to the [NEXT] release milestone?

vanitabarrett (approved):

EoinShaughnessy:
Have changed "Don't" in title to "Do not." We'll be linking to this PR from our next release notes and [many users have difficulties with negative contractions](https://readabilityguidelines.co.uk/grammar-points/contractions/#1-avoid-negative-contractions).

PR #2285, opened 2021-07-29
Fix font-size when Chrome autofills text inputs

quis:
For some reason Chrome decides that using its own font declaration is preferable to keeping the input looking as it would without autofill.

This overrides that with GOV.UK Frontend’s bigger, better, default font.

Before | After
---|---
image | image

Copied from https://github.com/alphagov/notifications-admin/pull/3602

### DIFF
--- a/src/govuk/components/input/_index.scss
+++ b/src/govuk/components/input/_index.scss
@@ -39,6 +39,10 @@
         border-width: $govuk-border-width-form-element * 2;
       }
     }
+
+    &::-webkit-autofill::first-line {
+      @include govuk-font($size: 19);
+    }
   }
 
   .govuk-input::-webkit-outer-spin-button,


36degrees:
Digged into this a little bit to try and understand why the font is forced in Chrome, and it's got a fairly interesting background.

When hovering over an option in the autofill menu, Chrome 'pre-populates' the field to be autofilled with that value. However, the value is not meant to be programatically available at this point (e.g. by looking at the field's value). The value should only be available once the user has made an explicit choice to autofill the value. This is to prevent sites from being able to access a user's personal information (especially credit card details) without their consent.

[Chromium bug #916838](https://bugs.chromium.org/p/chromium/issues/detail?id=916838) outlines a security issue where a malicious user sets the font of the autofilled text, generating multiple fonts with different ligatures and then looking to see how the scroll width of the input is affected, allowing them to 'fuzz' the autofilled value.

In order to fix this security issue, the Chromium team [opted to pin the autofilled text to use the system font](https://bugs.chromium.org/p/chromium/issues/detail?id=916838#c27).

However, [users have found a way to work around this fix, by using the `::first-line` pseudo-selector exactly as we're doing here](https://stackoverflow.com/a/60379185). This effectively bypasses the original fix. This has [also been reported as a follow up issue (Chromium bug #1035058)](https://bugs.chromium.org/p/chromium/issues/detail?id=1035058).

There are also reportedly other ways to work around the forced system font (like [#951487](https://bugs.chromium.org/p/chromium/issues/detail?id=951487), [#1013882](https://bugs.chromium.org/p/chromium/issues/detail?id=1013882)), so the Chromium team have now implemented a different fix. They're [now mocking out the `scrollWidth` and `scrollHeight` values when in the suggest state](https://chromium.googlesource.com/chromium/src/+/7a9e7a968aa77eae1410b40dbddd8b0dab44406a).

This means the original fix – forcing a system font – is no longer needed (especially as it can be bypassed). In their own commit message, they mention that 'this [...] may allow us to disable the font overriding again.'

This is a very long winded way of saying that I would expect/hope that Chromium will disable the font overriding in the relatively near future as it doesn't appear to be an effective way to mitigate the security issue, and has no other benefits as far as I can tell.

With that in mind, we could:

- accept this change and keep an eye on the related Chromium bugs, with a view to removing it once this is fixed in Chromium
- accept this change and do nothing
- hold this change whilst we push Chromium to revert the font overriding, or at least understand their intent

Thoughts?

36degrees:
I've added a comment to a related Chromium bug – https://bugs.chromium.org/p/chromium/issues/detail?id=953689#c43

quis:
Looks like the `::-webkit-autofill::first-line` hack is no longer working in Chrome 95 🙄 

image

vanitabarrett:
I'm going to close this ticket for now. It looks like this workaround no longer works in newer versions of Chrome and I'm not sure it's a change we'd want to make as it bypasses something intentionally added to fix a security issue - although it looks like Chrome have implemented a different fix now, it still doesn't feel quite right to me. Hopefully Chrome will stop pinning the autofill font in future versions, we can keep an eye on the Chromium bug report(s).

PR #2290, opened 2021-07-30
Improve support for IE11 with Windows High Contrast Mode 

lfdebrux:
This PR adds support for IE11 to the previous Windows High Contrast Mode fixes in PR https://github.com/alphagov/govuk-frontend/pull/2264.

Internet Explorer 11 doesn't support the [`forced-colors` media feature](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/forced-colors) or the [`forced-color-adjust` property](https://developer.mozilla.org/en-US/docs/Web/CSS/forced-color-adjust), but it has equivalents in [`-ms-high-contrast`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/-ms-high-contrast) and [`-ms-high-contrast-adjust`](https://docs.microsoft.com/en-us/previous-versions/hh771863(v=vs.85)). We can use those CSS features to extend fixes to high contrast mode that previously only worked on more modern browsers.

| Fix | Before | After |
|--|--|--|
| [Form control focus states](https://github.com/alphagov/govuk-frontend/pull/2264) | Screenshot of focus controls in Internet Explorer 11 before change | Screenshot of focus controls in Internet Explorer 11 before change |

---

There are a couple of other places in our codebase where we have fixes for forced color mode (see PRs https://github.com/alphagov/govuk-frontend/pull/2186, https://github.com/alphagov/govuk-frontend/pull/2237), but the problems they fix don't seem to be present in IE11 (I think because it doesn't have a readability backplate), so I left those lines alone.


vanitabarrett (approved):

lfdebrux:
Just added a changelog line to this commit!

PR #2302, opened 2021-08-10
Add tech debt issue template

lfdebrux:
Adds a GitHub issue template [[1]] for tracking tech debt in the manner encouraged by the GDS Way [[2]].

This should make it slightly easier to add tech debt tickets to this repo.

[1]: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/about-issue-and-pull-request-templates
[2]: https://gds-way.cloudapps.digital/standards/technical-debt.html#example


vanitabarrett:
@lfdebrux Is this something we need on other repos (e.g: design system, I know we have some tech debt cards there too)?

lfdebrux:
> @lfdebrux Is this something we need on other repos (e.g: design system, I know we have some tech debt cards there too)?

Yeah, that would be cool I think, if you're happy with this one I can add it to the other repos. I don't think we have a tool we have for syncing issue templates do we?

vanitabarrett:
@lfdebrux Nope, we don't have an automated tool for that at the moment. Sounds like a good plan 👍 This looks good to me, but it's probably a good idea to loop in @EoinShaughnessy as they reviewed the other issue templates across all the repos.

36degrees:
Good idea, thanks for doing this @lfdebrux! 👍🏻 

Just a thought – if we wanted, this could be an interesting opportunity to try out [issues forms](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms) – it seems like a good fit given how the impact, effort and overall ratings map quite neatly to dropdowns, and it also gives us the ability to provide additional guidance on how to fill it in.

lfdebrux:
I've changed it to an issue form as suggested by @36degrees, but I can't find a way to preview what it would look like 🤔

36degrees:
@lfdebrux for things like this in the past I've created a temporary repo on my personal account, added the YAML file and tested it there.

lfdebrux:
Ah, I've found you get a preview if you view the file in blob mode:

https://github.com/alphagov/govuk-frontend/blob/ldeb-add-tech-debt-template/.github/ISSUE_TEMPLATE/tech-debt.yaml

vanitabarrett (approved):

lfdebrux:
> Small question (which also applies to the [GDS Way example](https://gds-way.cloudapps.digital/standards/technical-debt.html#example)): can the 'Impact of debt' rating and 'Overall rating' ever be different from one another? Or are they always likely to be the same?

Great question! The overall impact is usually what I've used in the past to prioritise tech debt. So I don't know if this is how you're supposed to do it, but I've always treated the overall impact as being some kind of function of the impact of debt and the effort to pay down, but it's not very straightforward. Usually that means I make the overall impact be the highest or the average of the two. For example, an issue that is low impact and high effort to pay down usually has a low overall rating, but an issue that is high impact and high effort to pay down it is harder to say what the overall rating is.

Anyway I guess this is a long-winded way of saying no, but we should probably agree as a team what it is supposed to be...

EoinShaughnessy:
@lfdebrux Thanks for the thoughtful answer! Really helped me understand what informs the overall rating.

EoinShaughnessy (approved):

PR #2306, opened 2021-08-11
add width to file upload

### LINKED ISSUE #2007: File upload component at 400% zoom does not resize
Component: https://design-system.service.gov.uk/components/file-upload/

At 400% zoom (screenshot on Chrome), the file input does not resize to fit the viewport and horizontal scrolling appears:
screenshot-fileinput-400%zoom

Having style `width: 100%` for the file input does rescale it correctly (as done for text inputs, e.g. https://design-system.service.gov.uk/components/text-input/default/index.html)

For reference https://github.com/alphagov/govuk-design-system-backlog/issues/49#issuecomment-722362355

CharlotteDowns:
## What we've changed
This PR adds a width attribute to the file-upload component.

Fixes #2007 .
## Why we've changed it
When a user magnifies to 400% zoom the component does not resize to fit the viewport and horizontal scrolling appears.


vanitabarrett:
@CharlotteDowns Is it intentional that we're making the file upload component always full width now, so it'll take up the full width of the container it's placed in? This does seem to match what we do for the text input component, but I know Ollie's [original comment](https://github.com/alphagov/govuk-design-system-backlog/issues/49#issuecomment-722415301) in the backlog seems to imply that's something we might not want to do for the file upload component.

Example at default zoom on desktop the focus area for the component now looks like this:

Screenshot 2021-08-11 at 13 26 58

Whereas before it looked like this:

Screenshot 2021-08-11 at 13 37 27

CharlotteDowns:
@vanitabarrett I had the same concern. 

I'll do some further investigation work around the original width and the other implications of making the file-upload component `100% width`.

36degrees:
Bear in mind the file upload input type is displayed very differently across different browsers and operating systems. I think we'd want to test this pretty thoroughly across all of our supported combinations to be sure. Ideally it'd be good to get before/after screenshots for all of them.

Most of our other form inputs will grow to fill the available space, so I can see the argument for doing the same here. Aside from the way the focus state displays, it also means there can be quite a lot of invisible 'clickable area' in some browsers – which might also be an issue on mobile when scrolling?

(Now it's been pointed out, I've no idea why the file upload in Chrome doesn't shrink to tightly wrap the button and filename label – I've no idea why it's the width it is.)

CharlotteDowns:
### Testing across browsers, before and after
We've had a look into the before behaviour of the focused file upload component compared with the focused file upload component at `max-width: 100%`.

On Chrome version 92.0.4515.159 without max-width the component requires right scrolling
Screenshot of file-upload on chrome version 92.0.4515.159 without max-width

On Chrome version 92.0.4515.159 with max-width the component is contained within the width of the browser size Screenshot of file-upload on chrome version 92.0.4515.159 with max-width

[More examples across the other browsers that we support can be found in this slide deck](https://docs.google.com/presentation/d/1CRK-BSwggVhVypWSGjzRS2gyZNhEfGnn2DCRUk69TCE/edit?usp=sharing).

PR #2312, opened 2021-08-13
Remove padding-right from details element

edwardhorsford:
The details element doesn't need `padding-right` - having it will mean contents don't line up nicely with the rest of the page.

| Browser | Before |	After |
|---|---|---|
| Chrome | govuk-frontend-review herokuapp com_full-page-examples_what-is-your-nationality(Chrome) | govuk-frontend-pr-2312 herokuapp com_full-page-examples_what-is-your-nationality(Chrome) |
| Safari mobile | govuk-frontend-review herokuapp com_full-page-examples_what-is-your-nationality(iPhone 6_7_8) | govuk-frontend-pr-2312 herokuapp com_full-page-examples_what-is-your-nationality(iPhone 6_7_8)|

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ We’ve made fixes to GOV.UK Frontend in the following pull requests:
 - [#2273: Fix invisible footer OGL logo in forced color mode](https://github.com/alphagov/govuk-frontend/pull/2273) – thanks to [@oscarduignan](https://github.com/oscarduignan) for reporting this issue.
 - [#2277: Fix invisible start button chevron in forced color mode](https://github.com/alphagov/govuk-frontend/pull/2277)
 - [#2290: Improve support for IE11 with Windows High Contrast Mode](https://github.com/alphagov/govuk-frontend/pull/2290)
+- [#2312: Remove padding-right from details element](https://github.com/alphagov/govuk-frontend/pull/2312)
 
 ## 3.13.0 (Feature release)
 
--- a/src/govuk/components/details/_index.scss
+++ b/src/govuk/components/details/_index.scss
@@ -71,7 +71,8 @@
   }
 
   .govuk-details__text {
-    padding: govuk-spacing(3);
+    padding-top: govuk-spacing(3);
+    padding-bottom: govuk-spacing(3);
     padding-left: govuk-spacing(4);
     border-left: $govuk-border-width solid $govuk-border-colour;
   }


lfdebrux:
This was discussed in a [thread asking why the right padding is present in the #govuk-design-system channel in x-gov Slack](https://ukgovernmentdigital.slack.com/archives/C6DMEH5R6/p1628864105076900). I shared the question with designers in an [internal Slack thread](https://gds.slack.com/archives/G011V7RAAKY/p1628864433321600), @joelanman and @christopherthomasdesign agreed that it didn't need to be there. Looking at the git history, the right padding has been there since this component was added (see commit https://github.com/alphagov/govuk-frontend/blame/35aa6906eeb2a536a82ec29b3c3c1fa248715d61/src/components/details/_details.scss), with no reason given as to why it is there.

36degrees:
> Looking at the git history, the right padding has been there since this component was added (see commit [2e6769a](https://github.com/alphagov/govuk-frontend/commit/2e6769add82133147e4de408ce1e508b191008aa)), with no reason given as to why it is there.

The details component came from Elements, where it was known as [hidden text](https://govuk-elements.herokuapp.com/typography/#typography-hidden-text). In Elements, it used the 'panel' styles which were shared with inset text and conditional reveals, so changes to those panel styles affected all three things.

At one point, [there was no right padding on the panel](https://github.com/alphagov/govuk_elements/blob/e301e3e9acd6e157c32b65fe43b10ec20cefdb00/public/sass/elements/_panels.scss#L13). However, this was changed in https://github.com/alphagov/govuk_elements/commit/af2375b6ad37fddb618867f02c2f313ede56d17c (part of https://github.com/alphagov/govuk_elements/pull/106), when the spacing was adjusted 'so the padding is 15px on all sides'. As far as I can tell, there's been padding on all sides ever since.

I do think this change makes sense though – it makes it consistent with the conditional reveals which don't have right padding either.

36degrees:
Just to record a conversation from Slack – we discussed whether the right padding should be retained at the mobile breakpoint as it can look a little 'unbalanced', and adding padding on both sides emphasises the ‘inset’ nature:

| Before | After |
| -- | -- | 
| govuk-frontend-review herokuapp com_full-page-examples_what-is-your-nationality(iPhone 6_7_8) | govuk-frontend-pr-2312 herokuapp com_full-page-examples_what-is-your-nationality(iPhone 6_7_8) |

However, we ultimately agreed to remove the right padding on mobile (which is the change already proposed in this PR) because:

- it's consistent with the way padding works on conditional reveals
- it maximises the horizontal space available to users to see what they're writing, especially on smaller screens

36degrees (approved):
Thanks for this @edwardhorsford – are you able to add an entry to the changelog?

It'd also be really useful to include before/after screenshots in the PR description, if possible.

edwardhorsford:
@36degrees updated description with your screenshots and added an entry in the changelog.

christopherthomasdesign (approved):

36degrees:
Rebased against `main` to resolve a conflict in the changelog.

Issue #2310, opened 2021-08-12
Update docs to explain that accordion summary line can only contain phrasing content


hannalaakso:
## What

As explained in https://github.com/alphagov/govuk-frontend/issues/2309, we’re changing the design of the accordion which means that the summary line can only contain [phrasing content](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#phrasing_content).

We should update the relevant docs about this restriction: https://github.com/alphagov/govuk-frontend/blob/1c57eab2f74c4873ee5685cc6ad822a36f824352/src/govuk/components/accordion/accordion.yaml#L38 

## Why

To help our users understand that having elements other than phrasing content inside the summary line is not valid HTML, and also that the summary content should be simple content (elements other than phrasing content inside the summary would be likely not be an appropriate use of the accordion component)

## Who needs to know about this
Technical writer, developer

## Done when

- [x]  We’ve updated the Nunjucks documentation

EoinShaughnessy:
The Nunjucks update has passed 2i. See [#2313](https://github.com/alphagov/govuk-frontend/pull/2313).

hannalaakso:
@EoinShaughnessy Thanks for updating the docs. We should make this information available to users of the HTML version too. I guess that would fall somewhere between this ticket and https://github.com/alphagov/govuk-design-system/issues/1684 - I've left a note on that issue.

PR #2313, opened 2021-08-16
Update accordion macros about summary line change

EoinShaughnessy:
Addresses [#2310](https://github.com/alphagov/govuk-frontend/issues/2310).

**What we've added**

This PR adds content to the following parameters' Descriptions in the macros table for the [accordion component](https://design-system.service.gov.uk/components/accordion/):

- `summary.html`
- `heading.html`

**Why we've added it**

This new content tells users the summary line can only contain [phrasing content](https://html.spec.whatwg.org/#phrasing-content). This is because using non-phrasing content will result in invalid HTML.

We've also advised users to:

- keep content simple
- reconsider their use of the accordion, if they'd planned to use non-phrasing content

### DIFF
--- a/src/govuk/components/accordion/accordion.yaml
+++ b/src/govuk/components/accordion/accordion.yaml
@@ -23,11 +23,11 @@ params:
     - name: heading.text
       type: string
       required: true
-      description: The title of each section. If `heading.html` is supplied, this is ignored. This is used both as the title for each section, and as the button to open or close each section.
+      description: The title of each section. If `heading.html` is supplied, this is ignored.
     - name: heading.html
       type: string
       required: true
-      description: The HTML content of the header for each section which is used both as the title for each section, and as the button to open or close each section.
+      description: The HTML content of the header for each section. Used as the title for each section. The header is inside the HTML `<button>` element, so you can only add [phrasing content](https://html.spec.whatwg.org/#phrasing-content) to it.
     - name: summary.text
       type: string
       required: false
@@ -35,7 +35,7 @@ params:
     - name: summary.html
       type: string
       required: false
-      description: HTML content for summary line.
+      description: The HTML content for the summary line. The summary line is inside the HTML `<button>` element, so you can only add [phrasing content](https://html.spec.whatwg.org/#phrasing-content) to it.
     - name: content.text
       type: string
       required: true


src/govuk/components/accordion/accordion.yaml
```diff
@@ -35,7 +35,7 @@ params:
     - name: summary.html
       type: string
       required: false
-      description: HTML content for summary line.
+      description: HTML content for summary line. You can only include [phrasing content](https://html.spec.whatwg.org/#phrasing-content) inside the summary line. This content should be simple. Do not use other elements besides phrasing content, as they will not be valid HTML. Any other elements are not likely to be an appropriate use of the accordion component.
```
hannalaakso, line 38: Could we add something about how this restriction is due to the fact that the summary line is inside `<button>` element and buttons can only contain phrasing content.

EoinShaughnessy: @hannalaakso Good shout! How does this look?

```suggestion
      description: HTML content for summary line. The summary line is inside the button element, so it can only contain [phrasing content](https://html.spec.whatwg.org/#phrasing-content). This content should be simple. Do not use other elements besides phrasing content, as they will not be valid HTML. Any other elements are not likely to be an appropriate use of the accordion component.
```

hannalaakso: Thanks @EoinShaughnessy, that looks good.

A couple more small suggestions:
> The summary line is inside the HTML `<button>` element

to clarify it's the HTML element we're talking about here, not a CSS class name.


> as they will not be valid HTML.

I would change this to say that the resulting HTML (ie. having non phrasing content inside a button) will not be valid (it's not that the non phrasing content elements by themselves are not valid).

> Any other elements are not likely to be an appropriate use of the accordion component. 

I'm unsure about this line (I just wrote it quickly to try and describe the issue). I would try to clarify it to something like

> The content should be simple. If you need to use non phrasing content inside the summary line, you should consider simplifying your content or reviewing the [guidance](https://design-system.service.gov.uk/components/accordion/#when-not-to-use-this-component) to understand whether the accordion is the correct component to use.

EoinShaughnessy:
```suggestion
      description: HTML content for summary line. You can only include [phrasing content](https://html.spec.whatwg.org/#phrasing-content) inside the summary line. This content should be simple. Do not use other elements besides phrasing content, as they will not be valid HTML. Any other elements are not likely to be an appropriate use of the accordion component.
```

EoinShaughnessy: @hannalaakso Thanks! That sounds good.

> If you have considered using non-phrasing content inside the summary line, you should consider simplifying your content or reviewing the guidance to understand whether the accordion is the correct component to use.

Even if they need to, isn't it the case that they won't be able to (because the summary line can only contain phrasing content)?

```suggestion
      description: HTML content for summary line. The summary line is inside the HTML <button> element, so it can only contain [phrasing content](https://html.spec.whatwg.org/#phrasing-content). This content should be simple. Do not use other elements besides phrasing content, as the resulting HTML will not be valid. If you had planned to use non-phrasing content inside the summary line, you should instead:

- consider simplifying your content
- review the [accordion component's guidance](https://design-system.service.gov.uk/components/accordion/#when-not-to-use-this-component) to understand whether it's the right component to use

```

EoinShaughnessy:
Content has passed internal review. Will now send it for pre-i.

PippaClarkGDS (approved):

EoinShaughnessy:
Content has passed pre-i. Will now send it for 2i.

jonathanglassman (approved):

EoinShaughnessy:
@hannalaakso Content has passed 2i. Can you approve? I won't merge until we're ready to release the iteration. :)

hannalaakso (approved):

src/govuk/components/accordion/accordion.yaml
```diff
@@ -23,19 +23,19 @@ params:
     - name: heading.text
       type: string
       required: true
-      description: The title of each section. If `heading.html` is supplied, this is ignored. This is used both as the title for each section, and as the button to open or close each section.
+      description: The title of each section. If `heading.html` is supplied, this is ignored.
     - name: heading.html
       type: string
       required: true
-      description: The HTML content of the header for each section which is used both as the title for each section, and as the button to open or close each section.
+      description: The HTML content of the header for each section which is used both as the title for each section, and as the button to open or close each section. Can only include [phrasing content](https://html.spec.whatwg.org/#phrasing-content).
```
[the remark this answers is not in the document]
hannalaakso: @EoinShaughnessy Yes, I think removing `and as the button to open or close each section` makes sense (we _could_ expand on it to explain that it now forms **one part** of the button text - but the guidance already covers this).

But I think we should still keep 
`which is used both as the title for each section`

src/govuk/components/accordion/accordion.yaml
```diff
@@ -23,19 +23,19 @@ params:
...
     - name: summary.text
       type: string
       required: false
       description: Text content for summary line. If `summary.html` is supplied, this is ignored.
     - name: summary.html
       type: string
       required: false
-      description: HTML content for summary line.
+      description: HTML content for summary line. The summary line is inside the HTML `<button>` element, so you can only add [phrasing content](https://html.spec.whatwg.org/#phrasing-content) to it. Keep this content simple. Do not use any non-phrasing content, as the resulting HTML will not be valid. Instead, consider simplifying your content and [reviewing the accordion guidance to confirm whether it's the right component to use](https://design-system.service.gov.uk/components/accordion/#when-not-to-use-this-component).
```
hannalaakso, line 38: Now that we know what the guidance will contain, it feels that 
```
Keep this content simple. Do not use any non-phrasing content, as the resulting HTML will not be valid. Instead, consider simplifying your content and [reviewing the accordion guidance to confirm whether it's the right component to use](https://design-system.service.gov.uk/components/accordion/#when-not-to-use-this-component).
```
duplicates what the guidance says. 

If we did still want to keep it, we should probably also include it for `heading.html`. But it might be easier just to remove it.

src/govuk/components/accordion/accordion.yaml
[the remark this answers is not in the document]
hannalaakso: Something else (possibly slightly picky, sorry!): The structure of `heading.html` could more closely mirror `summary.html` content as both have the same content restriction.

src/govuk/components/accordion/accordion.yaml
```diff
@@ -23,19 +23,19 @@ params:
     - name: heading.text
       type: string
       required: true
-      description: The title of each section. If `heading.html` is supplied, this is ignored. This is used both as the title for each section, and as the button to open or close each section.
+      description: The title of each section. If `heading.html` is supplied, this is ignored.
```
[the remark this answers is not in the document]
hannalaakso: Looks good, thanks @EoinShaughnessy 👍

hannalaakso (changes requested):
Just blocking this from being merged until we get a chance to chat about my comments above (although they aren't necessarily blocking).

src/govuk/components/accordion/accordion.yaml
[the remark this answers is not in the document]
EoinShaughnessy: @hannalaakso Those are both good points!

```suggestion
      description: The HTML content of the header for each section. Used as the title for each section. The header is inside the HTML `<button>` element, so you can only add [phrasing content](https://html.spec.whatwg.org/#phrasing-content) to it.
```

src/govuk/components/accordion/accordion.yaml
EoinShaughnessy: @hannalaakso Yeah, agreed!

```suggestion
      description: HTML content for summary line. The summary line is inside the HTML `<button>` element, so you can only add [phrasing content](https://html.spec.whatwg.org/#phrasing-content) to it.
```

src/govuk/components/accordion/accordion.yaml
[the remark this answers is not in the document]
hannalaakso: This looks good 👍

src/govuk/components/accordion/accordion.yaml
hannalaakso: Likewise, this looks good.

hannalaakso (approved):
@EoinShaughnessy Thanks for addressing my feedback, those two proposed commits look good.

EoinShaughnessy:
@hannalaakso Thanks for reviewing! Do we need to wait for 4.0 release before merging?

hannalaakso:
@EoinShaughnessy We've started merging 4.0 things to main now so this can be merged in 👍 (for anyone's who's interested, we're putting a [separate process](https://github.com/alphagov/govuk-frontend/issues/1958) in place if we ever find ourselves in a situation where we need to publish any more 3.x.x releases when 4.0 changes have already been merged into main).

PR #2326, opened 2021-08-25
Document how to initialise JS and target elements

hannalaakso:


EoinShaughnessy (commented):
Wording looks good to me. Do we just need to replace the mentions of `js` with `HTML`, as per Vanita's comments?

EoinShaughnessy (approved):

Issue #1955, opened 2020-09-15
Improve the banner on GOV.UK Frontend review app


hannalaakso:
## What
A user on support ran the frontend repo locally and was confused by the banner on the review app:
> This is not a real service, and is used for testing purposes only. See the [GOV.UK Design System] (https://design-system.service.gov.uk/) for recommended guidance.

They thought that the above meant that GOV.UK Frontend shouldn't be used in production.

The above text is shown on all the pages of the review app but really only applies to the full page examples. For any other pages, could we change the text to something like:
> This is an internal development app. See [Get started] (https://design-system.service.gov.uk/get-started/) for how to use GOV.UK Design System in your project.

## Why
New users might come to the frontend repo from somewhere else than the Design System and think they need to run it locally to use it. We should improve the messaging they see if that happens.

## Who needs to know about this
Developers, Technical Writer

PR #2333, opened 2021-09-02
Add wording and Get started link to Heroku banner

EoinShaughnessy:
Fixes [#1955](https://github.com/alphagov/govuk-frontend/issues/1955).

This PR replaces the previous wording and link in the banner for our [review app](https://govuk-frontend-review.herokuapp.com/).

The previous wording made some users think they should not use GOV.UK Frontend in production.

### DIFF
--- a/app/views/full-page-examples/announcements/index.njk
+++ b/app/views/full-page-examples/announcements/index.njk
@@ -2,7 +2,7 @@
 scenario: You want to read this article about '2018’s oddest requests from Brits abroad'.
 notes: Based on https://www.gov.uk/government/news/strictly-vampires-and-sausages-2018s-oddest-requests-from-brits-abroad
 ---
-{% extends "_generic.njk" %}
+{% extends "full-page-example.njk" %}
 
 {% from "breadcrumbs/macro.njk" import govukBreadcrumbs %}
 {% from "inset-text/macro.njk" import govukInsetText %}
@@ -11,7 +11,6 @@ notes: Based on https://www.gov.uk/government/news/strictly-vampires-and-sausage
 {% block pageTitle %}{{ pageTitle }} - GOV.UK{% endblock %}
 
 {% block header %}
-{% include "../../partials/banner.njk" %}
 {{ govukHeader({
   navigation: [
     {
--- a/app/views/full-page-examples/applicant-details/confirm.njk
+++ b/app/views/full-page-examples/applicant-details/confirm.njk
@@ -1,4 +1,4 @@
-{% extends "legacy.njk" %}
+{% extends "full-page-example-legacy.njk" %}
 
 {% set page_title = "Applicant details submitted" %}
 
--- a/app/views/full-page-examples/applicant-details/index.njk
+++ b/app/views/full-page-examples/applicant-details/index.njk
@@ -8,7 +8,7 @@ notes: >-
   known accessibility issues.
 ---
 
-{% extends "legacy.njk" %}
+{% extends "full-page-example-legacy.njk" %}
 
 {% from "back-link/macro.njk" import govukBackLink %}
 {% from "date-input/macro.njk" import govukDateInput %}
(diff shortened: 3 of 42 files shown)


EoinShaughnessy:
This is the original / current text:

This is not a real service, and is used for testing purposes only. See the [GOV.UK Design System] (https://design-system.service.gov.uk/) for recommended guidance.

As per [@hannalaakso's comment in the linked issue](https://github.com/alphagov/govuk-frontend/issues/1955#issue-702119419), can we make this PR apply to all pages except for the 'Full page examples' ones?

app/views/partials/banner.njk
```diff
@@ -2,9 +2,9 @@
 <div class="app-banner app-banner--warning" data-module="app-banner">
   <div class="govuk-width-container">
     <p class="govuk-body">
-      This is not a real service, and is used for testing purposes only.
+      This is an internal development app.
       <br>
-      See the <a class="govuk-link govuk-link--inverse" href="https://design-system.service.gov.uk/">GOV.UK Design System</a> for recommended guidance.
+      To learn how to use the GOV.UK Design System in your project, see [Get started](https://design-system.service.gov.uk/get-started/) .
```
EoinShaughnessy, line 7:
```suggestion
      To learn how to use the GOV.UK Design System in your project, see [Get started](https://design-system.service.gov.uk/get-started/).
```

app/views/partials/banner.njk
```diff
@@ -2,9 +2,9 @@
 <div class="app-banner app-banner--warning" data-module="app-banner">
   <div class="govuk-width-container">
     <p class="govuk-body">
-      This is not a real service, and is used for testing purposes only.
+      This is an internal development app.
       <br>
-      See the <a class="govuk-link govuk-link--inverse" href="https://design-system.service.gov.uk/">GOV.UK Design System</a> for recommended guidance.
+      To learn how to use the GOV.UK Design System in your project, see [Get started](https://design-system.service.gov.uk/get-started/).
```
lfdebrux, line 7: I don't think there you can use Markdown in this context, so this link should be hardcoded:

```suggestion
      To learn how to use the GOV.UK Design System in your project, see <a class="govuk-link govuk-link--inverse" href="https://design-system.service.gov.uk/get-started/">Get started</a>.
```

EoinShaughnessy: Ah interesting, did not know that - thanks, @lfdebrux 👍🏻

lfdebrux:
> @lfdebrux As per @hannalaakso's comment in the linked issue, can we make this PR apply to all pages except for the 'Full page examples' ones?

If I understand what Hanna is saying, we want to have a different banner for full page examples and for other pages?

EoinShaughnessy:
@lfdebrux Yeah, exactly. I took it to mean we want to:

- keep the existing banner-text for [Full page examples](https://govuk-frontend-review.herokuapp.com/full-page-examples)
- have different banner-text (as suggested in this PR)for all other pages

Is that right, @hannalaakso?

lfdebrux:
@EoinShaughnessy @hannalaakso Okay, I've done a bit of moving things around to make this work. I spent a while trying different solutions and I think this is the easiest and best solution; by moving the banner out of the `header` block and into the `bodyStart` block it means there are fewer places where we need to remember to include the correct banner template. I then also added two new full page examples templates, to reduce repetition even more.

The upshot is full page examples now have a different banner as we wanted (yay), and the banner is now before the skip link instead of after it (but I think this is a good thing). There's more details in the commit messages if you want it... Sorry this slightly bloats your PR Eoin 😅 .

EoinShaughnessy:
@lfdebrux Looks great, thanks! Really appreciate all your thought and effort. 👍🏻

hannalaakso (approved):
Thanks both, this is a nice tidy up of the various templates we have 👍  It's also useful that the banner is now in the layout whereas before we could have missed including it when we set up a new example page.

EoinShaughnessy:
Thanks @hannalaakso!

Issue #1824, opened 2020-06-01
Create override classes for alignment


hannalaakso:
## What
Re: @edwardhorsford's [comment](https://github.com/alphagov/govuk-design-system-backlog/issues/61#issuecomment-625183121), it would be useful to have override classes for aligning text and table cell content to left, right and middle. 

```
.govuk-!-align-left
.govuk-!-align-right
.govuk-!-align-center
```

These classes would respectively set `text-align` to `left`, `right` and `center`. The naming is inline with the `display` classes which are `.govuk-!-display-block`, `.govuk-!-inline` etc.

## Why
This is a very common and generic requirement and utility classes like this exist in other popular frameworks like Bootstrap. 

## Who needs to know about this
Devs, Stephen

## Done when
- [x] We've added the classes
- [x] We've created a card for relevant guidance in the Design System

hannalaakso:
Was going to pick this up but have a couple of questions on the language used. Will poll other devs for thoughts.

1. Should the class be called `.govuk-!-align-left` or `.govuk-!-text-align-left`? 
The first one is more succinct and text-align is sometimes used for aligning things other than text (although it works most reliably for text).
The second one is more accurate and I notice [this is the approach that Bootstrap take](https://getbootstrap.com/docs/5.1/utilities/text/).

2. Should we use the British spelling 'centre' in the class name (ie. `.govuk-!-text-align-centre`), even though the CSS rule it contains is `text-align: center`. We already go with the British spelling in `govuk-colour` so this would be consistent (and [follows the style guide](https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#american-and-uk-english)). In this particular instance it might not be what teams would expect (whereas they’re probably more used to the spelling of the colour mixins/classes as they are used everywhere) but I think we need to be consistent with this.

36degrees:
> 1. Should the class be called `.govuk-!-align-left` or `.govuk-!-text-align-left`?
>    The first one is more succinct and text-align is sometimes used for aligning things other than text (although it works most reliably for text).
>    The second one is more accurate and I notice [this is the approach that Bootstrap take](https://getbootstrap.com/docs/5.1/utilities/text/).

[...]

PR #2339, opened 2021-09-07
Add text align override classes

hannalaakso:
This PR:

- Adds text align override classes that set the CSS `text-align` property to `left`, `right` or `center`.
- Adds examples to the review app 

The review app example used for `govuk-!-text-align-left` is a bit odd, feel free to suggest a better one (this would also be useful for when we add an example to the Design System guidance).

Related: https://github.com/alphagov/govuk-design-system/issues/1863

Fixes https://github.com/alphagov/govuk-frontend/issues/1824

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### New features
+
+- [#2339: Add text align override classes](https://github.com/alphagov/govuk-frontend/pull/2339)
+
 ## 3.13.1 (Fix release)
 
 ### Fixes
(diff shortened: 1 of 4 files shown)


lfdebrux:
Hmm, I see what you're saying about the examples being odd, maybe it would be more useful to show examples using them to align text in a table?

36degrees (approved):
Changes make sense to me – thanks @hannalaakso! Agree with Laurence that it'd be useful to have examples that show how the classes work with tables – especially as that's the original use case that was identified.

Personally not sure how important it is to have the panel example – I'd be happy with just text overridden, even though the 'natural' alignment would be left anyway. You can still use dev tools to check that the alignment is coming explicitly from the override.

lfdebrux (approved):
Looks great! I appreciate the table examples :)

Issue #2173, opened 2021-03-29
Text within panel component can overflow the container


36degrees:
## Description of the issue

As originally raised in https://github.com/hmrc/accessibility/issues/20, and flagged in the backlog [here](https://github.com/alphagov/govuk-design-system-backlog/issues/40#issuecomment-729058084), text within the panel component can overflow the container, for example if viewed on a mobile with text resized to 200%.

![](https://user-images.githubusercontent.com/272769/82329510-53bbf000-99d9-11ea-8c4a-128de706ac17.png)

## Steps to reproduce the issue

In Safari on iOS

1. Open http://govuk-frontend-review.herokuapp.com/components/panel/preview on a mobile device
2. Resize the text to 200%

## Actual vs expected behaviour

The text should remain readable, but the words 'Application' and 'completed' are truncated (because they are white text on a white background where they overflow the container).

vanitabarrett:
This came up on support today, again flagged by an internal accessibility audit at HMRC as failing WCAG 1.4.4.

> Issue: Text is clipped (AA)
> WCAG 2.1 violation: 1.4.4 Resize text
> Page: Application Submitted
> URL:
> Journey/Step: J1 S6
> Description: A long string of text not breaking to accommodate the width of the parent container. The text gets clipped. Tested with increased page zoom 200% on ios using iPhone11
> Resolve: Text getting clipped due to text doesn’t wrap. Apply appropriate CSS.

vanitabarrett:
Had a quick look at this, using the solution mentioned above:

```scss
    word-wrap: break-word; // Fallback for older browsers only
    overflow-wrap: break-word;
```

It's definitely an improvement, as the text no longer overflows (white on white) so you can read it. On very small width screens it can be difficult to read because the text is so broken up.

<img width="144" alt="Screenshot 2021-09-09 at 16 16 43" src="https://user-images.githubusercontent.com/29889908/132713751-b3ac4724-25cf-462d-b98a-9a798ec50a76.png">

The other alternative would probably be combining the above solution with something like `min-content` for newer browsers, e.g:

```scss
    word-wrap: break-word; // Fallback for older browsers only
    overflow-wrap: break-word;

    @include govuk-media-query($until: tablet) {
      padding: govuk-spacing(6) - $govuk-border-width;
    }

    @supports (width: min-content) {
      min-width: min-content;
    }
```

Text is more readable, but it means the panel overflows the page which looks a bit broken and you have to scroll to see all the content.

[...]

PR #2347, opened 2021-09-14
Fix panel text overflowing when zoomed in on mobile

hannalaakso:
As discussed in https://github.com/alphagov/govuk-frontend/issues/2173, the panel heading text can overflow its container when a user increases text size on a mobile eg. using iOS Safari text resize controls. The overflowing is a particular problem with the panel component since it uses white text. When the text overflows the container, it is invisible on the white (page) background. When the text in our other components overflow, the user might have to scroll horizontally to view it but the the text remains legible.

We should aim to meet the [WCAG Text resize Success criterion](https://www.w3.org/WAI/WCAG21/Understanding/resize-text.html) which requires that
> text can be scaled up to 200%

## How we fix this

To help fix this, this PR:
- Decreases the padding of the panel component on mobile to allow the text flow better.
- As an if-all-else-fails attempt to stop very long words from overflowing, forces them to break and wrap instead. Hopefully this is not something that would happen often since we now also reduce the padding to help with this.

## Possible future improvements

We additionally considered reducing the font size of the heading on a mobile to avoid the text wrapping at very narrow viewports. However, the [`govuk-font` mixin](https://frontend.design-system.service.gov.uk/sass-api-reference/#govuk-font) uses the responsive font size map which fixes font sizes for different breakpoints to make our components behave consistently. There currently isn’t a way to override those set font sizes for `govuk-font`. Since the other fixes in this commit go a long way towards solving this issue and we haven’t come across this requirement with our other components so far, we’re hesitant to add this new feature to `govuk-font` at this point, but we’ll keep an eye out for similar requirements in the future.

We could consider adding something to the panel guidance about the need to keep the panel heading wording concise.

### Before
Screenshot 2021-09-14 at 11 34 26

### After

Screenshot 2021-09-23 at 13 35 17

### Before
Screenshot 2021-09-14 at 11 34 47

### After

Screenshot 2021-09-23 at 15 33 46

### Before
Screenshot 2021-09-14 at 11 37 39

### After

Screenshot 2021-09-23 at 15 34 01

## We decided not to hyphenate words that wrap

We also considered progressively enhancing any words that wrap by using `hyphens: auto` to hyphenate them but this seemed to potentially introduce more problems than it was solving. For instance, we found that iOS Safari can omit the dash from some words but not others, see below:

Screenshot 2021-09-23 at 15 52 48

iOS Safari can also hyphenate words unnecessarily, see below:
Screenshot 2021-09-23 at 15 53 36

Fixes https://github.com/alphagov/govuk-frontend/issues/2173

### DIFF
--- a/src/govuk/components/panel/_index.scss
+++ b/src/govuk/components/panel/_index.scss
@@ -12,7 +12,19 @@
     text-align: center;
 
     @include govuk-media-query($until: tablet) {
-      padding: govuk-spacing(6) - $govuk-border-width;
+      padding: govuk-spacing(3) - $govuk-border-width;
+
+      // This is an if-all-else-fails attempt to stop long words from overflowing the container
+      // on very narrow viewports by forcing them to break and wrap instead. This
+      // overflowing is more likely to happen when user increases text size on a mobile eg. using
+      // iOS Safari text resize controls.
+      //
+      // The overflowing is a particular problem with the panel component since it uses white
+      // text: when the text overflows the container, it is invisible on the white (page)
+      // background. When the text in our other components overflow, the user might have to scroll
+      // horizontally to view it but the the text remains legible.
+      overflow-wrap: break-word;
+      word-wrap: break-word; // Support IE (autoprefixer doesn't add this as it's not a prefix)
     }
   }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 - [#2339: Add text align override classes](https://github.com/alphagov/govuk-frontend/pull/2339)
 
+### Fixes
+
+- [#2347: Fix panel text overflowing when zoomed in on mobile #2347](https://github.com/alphagov/govuk-frontend/pull/2347)
+
 ## 3.13.1 (Fix release)
 
 ### Fixes
--- a/app/views/full-page-examples/passport-details/confirm.njk
+++ b/app/views/full-page-examples/passport-details/confirm.njk
@@ -2,14 +2,15 @@
 
 {% from "panel/macro.njk" import govukPanel %}
 
-{% set pageTitle = "Passport details submitted" %}
+{% set pageTitle = "We have emailed you a confirmation" %}
 {% block pageTitle %}{{ pageTitle }} - GOV.UK{% endblock %}
 
 {% block content %}
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
       {{ govukPanel({
-        titleText: pageTitle
+        titleText: pageTitle,
+        html: "Your reference number<br><strong>HDJ2123F</strong>"
       }) }}
     </div>
   </div>


src/govuk/components/panel/_index.scss
```diff
@@ -12,7 +12,15 @@
...
 
     @include govuk-media-query($until: tablet) {
-      padding: govuk-spacing(6) - $govuk-border-width;
+      padding: govuk-spacing(3) - $govuk-border-width;
+
+      // This is an if-all-else-fails attempt to stop long words from clipping on very narrow
+      // viewports by forcing them to break and wrap instead. The clipping is more likely to happen
+      // when user increases text size on a mobile eg. using iOS Safari text resize controls, and is
+      // more applicable to the panel than other components since the panel padding limits the width
+      // of text more than our other components do.
```
36degrees, line 21: I think the main reason this is an issue specifically for the panel component is the fact that it uses white text on a coloured background, so when text overflows outside of the container you end up with white text on the white (page) background which is then invisible.

When most (all?) other components overflow, it can look a bit ugly (and create horizontal scrolling) but the text remains legible.

Is there anything else this component does that means it limits the width of text more than other components? I'm not aware of anything, but might be missing something.

src/govuk/components/panel/_index.scss
```diff
@@ -32,6 +40,13 @@
     margin-bottom: govuk-spacing(6);
 
     @include govuk-font($size: 48, $weight: bold);
+
+    // Decrease font size on mobile to stop long words from clipping. The clipping is most likely
+    // to happen when user increases text size on a mobile eg. using iOS Safari text resize
+    // controls.
+    @include govuk-media-query($until: tablet) {
+      font-size: 24px;
+    }
```
36degrees, line 49: I don't think there's anywhere else in the codebase we override the typography scale in this way (?), and it feels like we're running into issues with the way the helpers are currently designed to work – specifically:

- defining the tablet/desktop font size first and then the mobile feels a little 'backwards' compared to how we define our typography 'mobile first' everywhere else
- we end up with redundant CSS outputted  by the `govuk-font` mixin (the font size on mobile is defined twice)
- we're missing some of the other important things that come from the `govuk-typography-responsive` mixin (the font size in rem as well as px, and the line-height)

Maybe we need to look at making some changes to the helpers to better support this, if we're sure this is the right approach?

src/govuk/components/panel/_index.scss
hannalaakso: Thank you, these are good points 👍  I've updated the code comment to reflect, have a look.

src/govuk/components/panel/_index.scss
hannalaakso: @36degrees Extending the mixins sounds like something we could consider. One for the discussion list? 

How do you feel me about creating a separate card for it, if I switch these media queries to use `govuk-font` to set the font for mobile? 

Or if this whole approach needs further discussion, let's chat about it next week.

hannalaakso: Thanks @36degrees. Since the other fixes in this commit go a long way towards solving this issue and we haven’t come across this requirement with our other components so far, I've removed the alternative font sizes. I've added the same to the PR description and we can keep an eye out for any requests to to add this feature to `govuk-font` in the future.

36degrees: Sounds sensible!

36degrees (approved):
Needs a changelog entry, but otherwise looks good to me! 👏🏻

36degrees:
Although I think the PR description needs updating now that we're no longer using `hyphens: auto`:

> * The wrapped words will also hyphenate using `hyphens: auto` which  isn’t fully supported by Edge or by [languages other than English](https://developer.mozilla.org/en-US/docs/Web/CSS/hyphens#browser_compatibility) (and it requires that the language of the document is set) but it is a progressive enhancement in this scenario to improve the experience if the browser and page language support it.

PR #2348, opened 2021-09-17
Allow `govuk-spacing` to output negative spacing

### LINKED ISSUE #1760: It’s not clear what the best way to define negative spacing is
You can add spacing to a CSS property using mixins from GOV.UK Frontend, for example:
```scss
.example {
  margin-bottom: govuk-spacing(1)
}
```
The number (hereafter `n`) can be anything from `0`–`9`, which outputs values from `0px` to `60px` (on large screens).

This works fine for increasing the spacing between stuff. However in CSS it’s also possible to give properties like `margin` or `top` negative values to move stuff in the opposite direction. It’s not clear what the best way to express this is using GOV.UK Frontend.

# What doesn’t work

## `-govuk-spacing(n)`

Presuming this is a limitation of SASS’s syntax. Results in `margin-top: -govuk-spacing(1)` in the outputted CSS.

# What does work

## `- govuk-spacing(n)` (note the space)

Feels fragile and not very explicit

## `-1 * govuk-spacing(n)`

Feels a bit messy.

## `govuk-spacing(-n)`

Feels nicest, but isn’t documented, so could break in the future.

36degrees:
Occasionally, users need to use the `govuk-spacing` class to output negative spacing, for example to add a negative margin to bring elements closer together.

There are currently a few ways to do this.

You can use `- govuk-spacing(n)` but the extra space is required, and it looks like a calculation (and may be interpreted as such in some contexts).

You can also multiply the result of the function call by -1, for example `govuk-spacing(n) * -1`, but this feels a little clunky.

To make it easier for users to do this, allow govuk-spacing to accept a negative version of a point on the spacing scale, and output the equivalent spacing from the scale multiplied by -1.

Closes #1760

### DIFF
--- a/src/govuk/helpers/_spacing.scss
+++ b/src/govuk/helpers/_spacing.scss
@@ -8,15 +8,26 @@
 ///
 /// Returns measurement corresponding to the spacing point requested.
 ///
-/// @param {Number} $spacing-point - Point on the spacing scale (set in `settings/_spacing.sccs`)
+/// @param {Number} $spacing-point - Point on the spacing scale
+///  (set in `settings/_spacing.scss`)
 ///
-/// @returns {String} Spacing Measurement eg. 10px
+/// @returns {String} Spacing measurement eg. 10px
 ///
 /// @example scss
 ///   .element {
 ///     padding: govuk-spacing(5);
-///     top: govuk-spacing(2) !important; // if `!important` is required
 ///   }
+///
+/// @example scss Using negative spacing
+///   .element {
+///     margin-top: govuk-spacing(-1);
+///   }
+///
+/// @example scss Marking spacing declarations as important
+///   .element {
+///     margin-top: govuk-spacing(1) !important;
+///   }
+///
 /// @access public
 
 @function govuk-spacing($spacing-point) {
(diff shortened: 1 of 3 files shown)


src/govuk/helpers/_spacing.scss
```diff
@@ -8,15 +8,26 @@
 ///
 /// Returns measurement corresponding to the spacing point requested.
 ///
-/// @param {Number} $spacing-point - Point on the spacing scale (set in `settings/_spacing.sccs`)
+/// @param {Number} $spacing-point - Point on the spacing scale
+///  (set in `settings/_spacing.sccs`)
```
vanitabarrett, line 12: Typo, should be settings/_spacing.scss

src/govuk/helpers/spacing.test.js
```diff
@@ -66,6 +82,36 @@ describe('@function govuk-spacing', () => {
         'Expected a number (integer), but got a string.'
       )
   })
+
+  it('throws an error when passed a non-existent point', async () => {
```
vanitabarrett, line 86: Do you think it's worth adding a test for being passed a non-existent negative point?

36degrees: Good suggestion, thanks! 👍🏻

vanitabarrett (approved):
Looks good, just needs a rebase for the Changelog

36degrees:
I've rebased to resolve a conflict with the CHANGELOG. I also attempted to write some more descriptive release notes for #2339 as it looked odd having one bulleted entry and one with an h4 and description.

vanitabarrett:
Might be worth asking @EoinShaughnessy to take a look at the release notes?

EoinShaughnessy:
@vanitabarrett @36degrees Where can I review the release notes draft? :)

vanitabarrett:
Decided to look at the release notes as a separate piece of work in https://github.com/alphagov/govuk-frontend/issues/2364

36degrees:
@vanitabarrett in which case, am I good to merge this?

vanitabarrett:
@36degrees Yep, go for it! 🚢 🚀

PR #2353, opened 2021-09-21
Add cookie banner examples using `headingHtml` and `html`

vanitabarrett:
## What
All of the examples we currently have in the review app used `headingText` and `text` only.
We should have visible examples which use HTML too. This PR adds a visible example to the component examples and also changes a full page example to use `html`.

## Why
Prompted by https://github.com/alphagov/govuk-design-system/issues/1886, I noticed we don't have any visible examples of the cookie banner component in the review app which use HTML.


36degrees (approved):
👍🏻

PR #2370, opened 2021-09-29
Prevent issues with conditionally revealed content when the ID of the revealed content includes characters that have a special meaning in CSS

### LINKED ISSUE #1808: Conditional reveals fail silently if id contains characters characters that have a special meaning in CSS

36degrees:
Because the string passed to `querySelector` is evaluated as a CSS selector, this can fail (throwing a JavaScript error) if the ID contains characters that have a special meaning in CSS, such as `.` or `[]` – the ID would need to be escaped for it to be evaluated correctly.

Avoid this by using `document.getElementById` instead, so the ID is no longer evaluated as a selector.

The downside is that we can't constrain the scope to the $module. However, we don't think this should cause any issues as the ID _should_ be unique within the document.

Closes #1808

### DIFF
--- a/src/govuk/components/checkboxes/checkboxes.js
+++ b/src/govuk/components/checkboxes/checkboxes.js
@@ -32,7 +32,7 @@ Checkboxes.prototype.init = function () {
 
     // Skip checkboxes without data-aria-controls attributes, or where the
     // target element does not exist.
-    if (!target || !$module.querySelector('#' + target)) {
+    if (!target || !document.getElementById(target)) {
       return
     }
 
--- a/src/govuk/components/checkboxes/checkboxes.test.js
+++ b/src/govuk/components/checkboxes/checkboxes.test.js
@@ -49,6 +49,7 @@ describe('Checkboxes with conditional reveals', () => {
       expect(hasAriaExpanded).toBeFalsy()
       expect(hasAriaControls).toBeFalsy()
     })
+
     it('falls back to making all conditional content visible', async () => {
       await goToAndGetComponent('checkboxes', 'with-conditional-items')
 
@@ -56,6 +57,7 @@ describe('Checkboxes with conditional reveals', () => {
       expect(isContentVisible).toBeTruthy()
     })
   })
+
   describe('when JavaScript is available', () => {
     it('has conditional content revealed that is associated with a checked input', async () => {
       const $ = await goToAndGetComponent('checkboxes', 'with-conditional-item-checked')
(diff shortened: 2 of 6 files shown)


vanitabarrett (approved):
Thanks Ollie, this change makes sense to me. Probably worth updating either the PR or the issue to explain which users will be affected by this change and what they might need to do, to help Eoin draft some release notes.

36degrees:
Needs to be merged after #2255 once the target branch has automagically updated.

36degrees:
First attempt at a changelog entry:

> The JavaScript for radios and checkboxes with conditionally-revealed content will now look for the conditionally-revealed content anywhere on the page rather than only looking within the same set of radios or checkboxes.
> 
> If you re-use the same `id` multiple times on a single page, this change could cause a radio or checkbox to hide and reveal the wrong element on the page.
> 
> Make sure that any radios or checkboxes with conditionally revealed content behave as expected. If you are seeing unexpected behaviour, make sure that the `id` for the revealed content is unique within the page it appears on.

Or, an alternative version which tries less hard to abstract technical details and language (which may be clearer for users who are more familiar with JavaScript, but is arguably harder to understand for everyone else):

> The JavaScript for radios and checkboxes with conditionally-revealed content will now look for the conditionally-revealed content using `document.getElementById` rather than being scoped to the same set of radios or checkboxes.
> 
> If you re-use the same `id` multiple times on a single page, this change could cause a radio or checkbox to hide and reveal the wrong element on the page.
> 
> Make sure that any radios or checkboxes with conditionally revealed content behave as expected. If you are seeing unexpected behaviour, make sure that the `id` for the revealed content is unique within the page it appears on.

I don't know if it's also worth mentioning that using the same `id` for multiple elements is also invalid HTML?

EoinShaughnessy:
@36degrees Changelog entry is taking shape nicely! I'd lean towards using the first, simpler version, to help any users who may be less familiar with JS. (Plus, the final PR will mention `document.getElementById`, right? So we're not short-changing our more JS-savvy users.)

> I don't know if it's also worth mentioning that using the same id for multiple elements is also invalid HTML?

Depends, I guess. This sounds like a borderline thing between writing Design System-specific instructions and telling people how to use HTML. Maybe, by saying multiple uses could display the wrong element, we've already given enough caution.

> The JavaScript for radios and checkboxes with conditionally-revealed content will now look for the conditionally-revealed content anywhere on the page rather than only looking within the same set of radios or checkboxes.
> 
> If you re-use the same id multiple times on a single page, this change could cause a radio or checkbox to hide and reveal the wrong element on the page.
> 
> Make sure that any radios or checkboxes with conditionally revealed content behave as expected. If you are seeing unexpected behaviour, make sure that the id for the revealed content is unique within the page it appears on.

First paragraph is about JS looking anywhere on the page. Second and third paragraphs are about reuse of ID. Is that ok?

Maybe combine the second and third paragraphs, and reverse their order? Something like in the revision below:

<h4>JavaScript now looks across whole page for conditional reveals</h4>

On [radios](/components/radios/) and [checkboxes](/components/checkboxes/) with conditionally revealed content, the JavaScript now looks for the content anywhere on the page. Before, it only looked within the same set of radios or checkboxes.

Make sure any radios or checkboxes with conditionally revealed content behave as expected. If you see unexpected behaviour, make sure the `id` for the revealed content is unique within the page the content is on. If you reuse the same `id` within a page, this duplication could cause a radio or checkbox to reveal the wrong element.

PR #2369, opened 2021-09-29
Add release notes for Frontend v3.14.0

EoinShaughnessy:
Closes [#2364](https://github.com/alphagov/govuk-frontend/issues/2364).

This PR contains release notes that tell our users about:

- [new text-alignment classes](https://github.com/alphagov/govuk-frontend/pull/2339)
- [negative spacing functionality we've added](https://github.com/alphagov/govuk-frontend/pull/2348)
- [an accessibility fix to prevent panel text spilling over when zoomed on mobile](https://github.com/alphagov/govuk-frontend/pull/2347)

PR #2378, opened 2021-10-04
Update package.json to 3.14.0

vanitabarrett:
We released 3.14.0 from the support/3.x branch. We now need to make sure `main` is in sync and includes all of the changes that were released in 3.14.0.

The release notes were updated in a separate PR: https://github.com/alphagov/govuk-frontend/pull/2369

The changes themselves were already merged into `main` as we were trialling the support branch release process in a slightly artificial scenario and we decided to do this only after the changes were already in `main`.

That therefore leaves updating the package/package.json file so it's in sync with the latest release version - this is done in this PR

### DIFF
--- a/package/package.json
+++ b/package/package.json
@@ -1,7 +1,7 @@
 {
   "name": "govuk-frontend",
   "description": "GOV.UK Frontend contains the code you need to start building a user interface for government platforms and services.",
-  "version": "3.13.1",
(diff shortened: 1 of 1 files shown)


lfdebrux (approved):

36degrees (approved):
👍🏻

Issue #1379, opened 2019-05-28
Remove deprecated main-wrapper mixins


nickcolley:
Placeholder issue to track the removal of a deprecated feature, will update soon.

nickcolley:
Wonder if we should use `@warn` directives on these before 4.0.0 @36degrees ?

EoinShaughnessy:
Will we need to update the [spacing mixins guidance](https://design-system.service.gov.uk/styles/spacing/#spacing-on-custom-components)? Or just tell users in the release notes?

vanitabarrett:
Thanks @EoinShaughnessy , good shout. I don't think the spacing mixins guidance needs updating, but it looks like we do reference these mixins in https://design-system.service.gov.uk/styles/layout/#page-wrappers

PR #2385, opened 2021-10-15
Remove deprecated `govuk-main-wrapper` and `govuk-main-wrapper--l` mixins

vanitabarrett:
Fixes https://github.com/alphagov/govuk-frontend/issues/1379 (also see here for details of how this is a breaking change for users)

In https://github.com/alphagov/govuk-frontend/commit/ade03d4ba46a0479ad51880112d5c7fd036a3255 we deprecated the `main-wrapper` and `main-wrapper--l` mixins because we didn't think they were useful.

This PR removes those mixins. Note: the classes remain, but the mixins are being removed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,15 @@
 ### Breaking changes
 You must make the following changes when you migrate to this release, or your service may break.
 
+#### Remove deprecated `govuk-main-wrapper` and `govuk-main-wrapper--l` mixins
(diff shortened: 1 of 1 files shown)


vanitabarrett:
@EoinShaughnessy I've added a draft entry in the Changelog for this change. Do you want to take a look here or should I copy it into the v4.0.0 draft release notes doc?

EoinShaughnessy:
@vanitabarrett Content looks good! I can copy it into the draft doc once we've resolved the 2 small comments I've added.

PR #2417, opened 2021-11-03
Remove deprecated `govuk-tag--inactive` class

### LINKED ISSUE #1757: Remove `govuk-tag--inactive` class
## What
Remove the `govuk-tag--inactive` class as it now duplicates what the `govuk-tag--grey` class does.

## Why
As discussed in https://github.com/alphagov/govuk-frontend/pull/1711#discussion_r370634942, https://github.com/alphagov/govuk-frontend/pull/1711 adds a new class `govuk-tag--grey`, the naming of which is inline with the other new tag colour classes.

Developers, Tech Writer

## Done when

- [ ] `govuk-tag--inactive` class has been removed
- [x] "Details of breaking change" section below completed to help with writing release notes

## Details of breaking change

- which users are affected: If you use `govuk-tag--inactive`
- the change that’s been made: `govuk-tag--inactive` has been removed
- changes users will have to make: Replace `govuk-tag--inactive` with `govuk-tag--grey`
- impact of users not making those changes: Tag will use default style instead of greyed-out style

lfdebrux:
Fixes #1757 (also see here for details of how this is a breaking change for users).

Class `govuk-tag--inactive` was replaced by `govuk-tag--grey` and deprecated in PR #1711.

This commit removes the class completely. Users should replace any uses of this class with `.govuk-tag--grey`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,14 @@ If you’re calling `iff` from your own Sass, you should use [Sass's native `if`
 
 This change was introduced in [pull request #2409: Remove deprecated `iff` Sass function](https://github.com/alphagov/govuk-frontend/pull/2409).
 
+#### Remove deprecated `govuk-tag--inactive` class
+
+We've removed the `govuk-tag--inactive` class which we deprecated in [GOV.UK Frontend v3.6.0](https://github.com/alphagov/govuk-frontend/releases/tag/v3.6.0).
+
+Replace any use of this class with the `govuk-tag--grey` class.
+
+This change was introduced in [pull request #2417: Remove deprecated `govuk-tag--inactive class`](https://github.com/alphagov/govuk-frontend/pull/2417).
+
 ### Fixes
 
 We’ve made fixes to GOV.UK Frontend in the following pull requests:
--- a/src/govuk/components/phase-banner/phase-banner.yaml
+++ b/src/govuk/components/phase-banner/phase-banner.yaml
@@ -60,6 +60,6 @@ examples:
     data:
       text: This is a new service – your feedback will help us to improve it
       tag:
-        classes: govuk-tag--inactive
+        classes: govuk-tag--grey
         text: alpha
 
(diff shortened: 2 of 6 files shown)


lfdebrux:
I've had a go at writing the changelog message, @EoinShaughnessy you might want to take a look as well?

CHANGELOG.md
```diff
@@ -57,6 +57,14 @@ If you’re calling `iff` from your own Sass, you should use [Sass's native `if`
 
 This change was introduced in [pull request #2409: Remove deprecated `iff` Sass function](https://github.com/alphagov/govuk-frontend/pull/2409).
 
+#### Remove deprecated `.govuk-tag--inactive` class
+
+We've removed the `.govuk-tag--inactive` class which we deprecated in [GOV.UK Frontend v3.6.0](https://github.com/alphagov/govuk-frontend/releases/tag/v3.6.0).
+
+Replace any use of this class with the `.govuk-tag--grey` class.
+
+This change was introduced in [pull request #2417: Remove deprecated `.govuk-tag--inactive class`](https://github.com/alphagov/govuk-frontend/pull/2417).
```
36degrees, line 66: Really minor, but suggest we avoid prefixing the class names with a period in the release notes because:

- technically the dot is part of the CSS selector rather than the class name
- wherever users are 'using' the class (in the classes in the HTML, or when passing classes to the Nunjucks macro) they wouldn't be including the `.`, so including it in the instructions may be confusing.

Appreciate this is inconsistent with, for example, the release notes for 3.6.0, but still think it's worth changing.

```suggestion
#### Remove deprecated `govuk-tag--inactive` class

We've removed the `govuk-tag--inactive` class which we deprecated in [GOV.UK Frontend v3.6.0](https://github.com/alphagov/govuk-frontend/releases/tag/v3.6.0).

Replace any use of this class with the `govuk-tag--grey` class.

This change was introduced in [pull request #2417: Remove deprecated `govuk-tag--inactive class`](https://github.com/alphagov/govuk-frontend/pull/2417).
```

src/govuk/components/phase-banner/phase-banner.yaml
```diff
@@ -60,6 +60,6 @@ examples:
     data:
       text: This is a new service – your feedback will help us to improve it
       tag:
-        classes: govuk-tag--inactive
```
36degrees, line 63: Good catch 👍🏻

36degrees (commented):
💥 🗑️ 🎉

36degrees (approved):

EoinShaughnessy:
@lfdebrux Changelog update looks great!

EoinShaughnessy (approved):

PR #2419, opened 2021-11-03
Add pagination component

timpaul:
A draft PR for a new contribution: a pagination component.

It's based partly on [this one in the MoJ Pattern Library](https://design-patterns.service.justice.gov.uk/components/pagination/) and partly on [this one in the GOV.UK Component Library](https://components.publishing.service.gov.uk/component-guide/previous_and_next_navigation) - but it takes into account guidance in the [Home Office Design System](https://design.homeoffice.gov.uk/components?name=Pagination), feedback in the [GOV.UK Design System backlog](https://github.com/alphagov/govuk-design-system-backlog/issues/77) and conversations with @adamsilver and @gregtyler (esp. [this issue in moj-frontend](https://github.com/ministryofjustice/moj-frontend/issues/161)). Guidance for the component is being developed in [this branch of govuk-design-system](https://github.com/alphagov/govuk-design-system/tree/add-pagination).

**[Preview this component in the GOV.UK Design System](https://add-pagination--govuk-design-system-preview.netlify.app/components/pagination/)**

### DIFF
--- a/src/govuk/components/pagination/_index.scss
+++ b/src/govuk/components/pagination/_index.scss
@@ -97,6 +97,7 @@
     padding-right: 0;
     padding-left: 0;
     color: $govuk-secondary-text-colour;
+    pointer-events: none;
   }
 
   .govuk-pagination__item--prev .govuk-pagination__link:before {
--- a/src/govuk/components/pagination/_index.scss
+++ b/src/govuk/components/pagination/_index.scss
@@ -54,6 +54,7 @@
     .govuk-pagination__icon {
       margin-top: 6px;
       margin-right: govuk-spacing(3);
+      fill: $govuk-secondary-text-colour;
     }
   }
 
@@ -81,6 +82,7 @@
     min-width: 45px;
     min-height: 45px;
     padding: 10px 18px;
+    color: $govuk-secondary-text-colour;
     font-weight: bold;
     text-align: center;
   }
--- a/src/govuk/components/pagination/_index.scss
+++ b/src/govuk/components/pagination/_index.scss
@@ -30,10 +30,6 @@
     }
   }
 
-  .govuk-pagination__list.govuk-pagination__list--block {
-    display: block;
-  }
-
   .govuk-pagination__results {
     @include govuk-font(19);
     margin-top: 0;
@@ -53,6 +49,14 @@
     @include govuk-media-query($until: tablet) {
       display: none;
     }
+
+    .govuk-pagination__icon {
+      margin-top: 6px;
+    }
+  }
+
+  .govuk-pagination__item--next .govuk-pagination__icon {
+    float: right;
   }
 
   // Only show previous, next, first, last and current items on mobile
@@ -67,17 +71,29 @@
     }
   }
 
-  // When link descriptions are used in previous and next links, align them vertically
-  .govuk-pagination__item.govuk-pagination__item--block {
-    display: block;
+  // Block mode - position previous and next links above and below numbers
+  .govuk-pagination__list--block {
+    .govuk-pagination__list {
+      display: block;
+      text-align: justify;
+    }
+
+    .govuk-pagination__item--prev,
+    .govuk-pagination__item--next {
+      display: block;
+    }
+
+    .govuk-pagination__item--next {
+      border-top: 1px solid $govuk-border-colour;
+
+      .govuk-pagination__icon {
+        float: none;
+      }
+    }
 
     .govuk-pagination__link {
       padding-left: 0;
     }
-  }
-
-  .govuk-pagination__item.govuk-pagination__item--next.govuk-pagination__item--block {
-    border-top: 1px solid $govuk-border-colour;
   }
 
   .govuk-pagination__item--active,
@@ -115,17 +131,25 @@
     min-width: 45px;
     min-height: 45px;
     padding: 10px 18px;
-    //text-align: center;
+
     &:hover {
       background: #f3f2f1;
+    }
+
+    .govuk-pagination__link-label {
+      display: block;
+      padding-left: govuk-spacing(5);
+      @include govuk-font($size: 16, $weight: "regular");
     }
   }
 
   .govuk-pagination__item--prev .govuk-pagination__link {
     padding-left: 0;
+    font-weight: bold;
   }
 
   .govuk-pagination__item--next .govuk-pagination__link {
     padding-right: 0;
+    font-weight: bold;
   }
 }
--- a/src/govuk/components/pagination/_index.scss
+++ b/src/govuk/components/pagination/_index.scss
@@ -126,13 +126,13 @@
       text-decoration: underline;
     }
 
-    &:hover {
-      color: $govuk-link-colour;
-      background: #f3f2f1;
+    &:focus {
+      box-shadow: 0 0 $govuk-focus-colour, 0 4px $govuk-focus-text-colour;
+      text-decoration: underline;
+    }
 
-      .govuk-pagination__link-label {
-        @include govuk-link-hover-decoration;
-      }
+    &:hover:not(:focus) {
+      background-color: govuk-colour("light-grey", $legacy: "grey-4");
     }
   }
 
--- a/src/govuk/components/pagination/_index.scss
+++ b/src/govuk/components/pagination/_index.scss
@@ -26,18 +26,18 @@
     @include govuk-media-query($from: tablet) {
       display: inline-block;
       margin-bottom: 0;
-      vertical-align: middle;
     }
   }
 
   .govuk-pagination__results {
     @include govuk-font(19);
     margin-top: 0;
+    margin-bottom: govuk-spacing(4);
     padding: govuk-spacing(1) 0;
     @include govuk-media-query($from: tablet) {
       display: inline-block;
       margin-bottom: 0;
-      vertical-align: middle;
+      //vertical-align: middle;
     }
   }
 
@@ -53,7 +53,7 @@
 
     .govuk-pagination__icon {
       margin-top: 6px;
-      margin-right: govuk-spacing(2);
+      margin-right: govuk-spacing(3);
     }
   }
 
@@ -103,18 +103,18 @@
 
   .govuk-pagination__link {
     @include govuk-link-common;
+    @include govuk-link-style-no-underline;
     box-sizing: border-box;
     display: block;
     min-width: 45px;
     min-height: 45px;
     padding: 10px 18px;
     color: $govuk-link-colour;
-    text-decoration: none;
 
     .govuk-pagination__link-label {
       @include govuk-font($size: 16, $weight: "regular");
       display: block;
-      padding-left: 31px;
+      padding-left: 32px;
       text-decoration: underline;
     }
 
@@ -124,10 +124,6 @@
 
       .govuk-pagination__link-label {
         @include govuk-link-hover-decoration;
-      }
-
-      .govuk-pagination__link-title {
-        //@include govuk-link-hover-decoration;
       }
     }
   }
@@ -166,6 +162,7 @@
 
     .govuk-pagination__link {
       padding-left: 0;
+      text-decoration: none;
     }
   }
 }
--- a/src/govuk/components/pagination/_index.scss
+++ b/src/govuk/components/pagination/_index.scss
@@ -89,23 +89,10 @@
   .govuk-pagination__link {
     @include govuk-link-common;
     @include govuk-link-style-default;
+    @include govuk-link-style-no-underline;
     display: block;
     min-width: 25px;
     padding: govuk-spacing(1);
     text-align: center;
-    text-decoration: none;
-
-    &:link,
-    &:visited {
-      color: $govuk-link-colour;
-    }
-
-    &:hover {
-      color: govuk-tint($govuk-link-colour, 25);
-    }
-
-    &:focus {
-      color: govuk-colour("black");
-    }
   }
 }
--- a/src/govuk/components/pagination/_index.scss
+++ b/src/govuk/components/pagination/_index.scss
@@ -1,7 +1,7 @@
 @include govuk-exports("govuk/component/pagination") {
   .govuk-pagination {
 
-    @include govuk-media-query($from: desktop) {
+    @include govuk-media-query($from: tablet) {
 
       // Alignment adjustments
       margin-right: - govuk-spacing(1);
@@ -25,7 +25,7 @@
     margin: 0;
     padding: 0;
     list-style: none;
-    @include govuk-media-query($from: desktop) {
+    @include govuk-media-query($from: tablet) {
       display: inline-block;
       margin-bottom: 0;
       vertical-align: middle;
@@ -36,7 +36,7 @@
     @include govuk-font(19);
     margin-top: 0;
     padding: govuk-spacing(1);
-    @include govuk-media-query($from: desktop) {
+    @include govuk-media-query($from: tablet) {
       display: inline-block;
       margin-bottom: 0;
       vertical-align: middle;
--- a/src/govuk/components/pagination/_index.scss
+++ b/src/govuk/components/pagination/_index.scss
@@ -5,8 +5,8 @@
     @include govuk-media-query($from: desktop) {
 
       // Alignment adjustments
-      margin-left: - govuk-spacing(1);
       margin-right: - govuk-spacing(1);
+      margin-left: - govuk-spacing(1);
 
       // Hide whitespace between elements
       font-size: 0;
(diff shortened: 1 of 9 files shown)


timpaul:
### Decisions still to make

- [ ] Elipses or dots?
- [ ] Allow html in links?
- [ ] What should the default for previous/next links be?

timpaul:
### Design decisions

The design is an attempt to balance 3 competing needs: 
- consistent spacing of numbers
- conservation of horizontal space
- large enough targets

To do this...

The number elements have a minimum width and height of 45px, to meet WCAG target size criteria. They also have horizontal padding applied (based on the perceived padding that a single-digit element has, due to the min-width setting). This is so that digits wider than 45px still receive the padding users expect.

The elipses characters are given negative margins, so that the numbers are evenly spaced, and horizontal space is conserved.

Finally, at mobile resolutions all but the first, last, current, previous and next characters are shown, again, to conserve horizontal space.

src/govuk/components/pagination/_index.scss
```diff
@@ -0,0 +1,170 @@
...
+      font-size: 0;
+
+      // Trick to remove the need for floats
+      text-align: justify;
+
+      &:after {
+        content: '';
+        display: inline-block;
+        width: 100%;
+      }
```
frankieroberto, line 18: 🤯 

It took me a while to work out how this trick works...  As I now understand it, this is so that the pagination list appears on the left and the results sentence on the right.

It works by using the "justify text" mode, and then setting the 2 elements as `inline-block`, so that the browser lays them out as if it was a line with 2 words on it. Only, the words in the last line of a justified paragraph aren't justified, so a third, invisible `inline-block` element is generated with a 100% wide width, forcing it to appear on its own line, making the 2 visible elements no longer be on the last line.

Does that sound about right? I think we’d need some comments to remind ourselves how this works...?

Incidentally, what’s the reason for not using floats (like the grid-columns)? Is it to enable the widths of the 2 elements to be flexible? If so, would another alternative be CSS table layout (`display: table-row` and `display: table-cell`)?

Another option might be to just stack them on the left, which might be better for screen magnifier users, and doesn’t look too bad to my eye...

Screenshot 2021-11-18 at 23 11 31

src/govuk/components/pagination/_index.scss
```diff
@@ -0,0 +1,170 @@
...
+    color: $govuk-secondary-text-colour;
+    font-weight: bold;
+    text-align: center;
+  }
+
+  .govuk-pagination__item--elipses {
+    margin: 0 -22.5px;
+    padding-right: 0;
+    padding-left: 0;
+  }
```
frankieroberto, line 94: Should the current page number appear black rather than dark grey?

Screenshot 2021-11-18 at 23 21 05

vs

Screenshot 2021-11-18 at 23 14 35


```suggestion
    font-weight: bold;
    text-align: center;
  }

  .govuk-pagination__item--elipses {
    margin: 0 -22.5px;
    padding-right: 0;
    padding-left: 0;
    color: $govuk-secondary-text-colour;
  }
```

src/govuk/components/pagination/template.njk
```diff
@@ -0,0 +1,62 @@
...
+
+  <ul class="govuk-pagination__list">
+    {%- if params.previous %}
+    <li class="govuk-pagination__item govuk-pagination__item--prev">
+      <a class="govuk-pagination__link" href="{{ params.previous.href }}">
+        <span class="govuk-pagination__link-title">
+          <svg class="govuk-pagination__icon" xmlns="http://www.w3.org/2000/svg" height="13" width="17">
+            <path d="m6.5938-0.0078125-6.7266 6.7266 6.7441 6.4062 1.377-1.449-4.1856-3.9768h12.896v-2h-12.984l4.2931-4.293-1.414-1.414z"></path>
+          </svg>{%- if params.previous.text %}{{ params.previous.text }}{% else %}Previous
+          <span class="govuk-visually-hidden">page</span>
```
frankieroberto, line 13: I kinda think we should avoid having default text within the templates? Partly as it makes translation harder, and partly as it just feels a bit unexpected to be able to call the template without specifying the text and have it magically appear.

Could we follow the pattern from the Summary List component here, and have `visuallyHiddenText` as a separate param?

```suggestion
          </svg>{{ params.previous.text }}{% if params.previous.visuallyHiddenText %} 
          <span class="govuk-visually-hidden">{{ params.previous.visuallyHiddenText }}</span>{% endif %}
```

src/govuk/components/pagination/template.njk
```diff
@@ -0,0 +1,62 @@
...
+        <span class="govuk-pagination__link-divider govuk-visually-hidden">:</span>
+        <span class="govuk-pagination__link-label">{{ params.previous.label }}</span>
+        {% endif -%}
+      </a>
+    </li>
+    {% endif -%}
+
+    {%- for item in params.items %}
+    {%- if item.type == 'elipses' %}
+    <li class="govuk-pagination__item govuk-pagination__item--elipses">&hellip;</li>
```
frankieroberto, line 26: I think this could be encoded directly, to save a few characters, and make it a bit clearer what it is?

```suggestion
    <li class="govuk-pagination__item govuk-pagination__item--elipses">…</li>
```

src/govuk/components/pagination/template.njk
```diff
@@ -0,0 +1,62 @@
...
+        <span class="govuk-pagination__link-divider govuk-visually-hidden">:</span>
+        <span class="govuk-pagination__link-label">{{ params.next.label }}</span>
+        {% endif -%}
+      </a>
+    </li>
+    {% endif -%}
+  </ul>
+
+  {%- if params.results %}
+  <p class="govuk-pagination__results">Showing <b>{{ params.results.from }}</b> to <b>{{ params.results.to }}</b> of <b>{{params.results.count}}</b> {{type}}</p>
```
frankieroberto, line 59: Add some extra classes here, just in case we (or anyone) want to style them differently in future?

```suggestion
  <p class="govuk-pagination__results">Showing <b class="govuk-pagination__results__from">{{ params.results.from }}</b> to <b class="govuk-pagination__results__to">{{ params.results.to }}</b> of <b class="govuk-pagination__results__count">{{params.results.count}}</b> {{type}}</p>
```

Alternatively, maybe the whole line should be passed in to the macro as html (or text)?

```suggestion
  <p class="govuk-pagination__results">{{ params.results.html }}</p>
```

frankieroberto (commented):
This is looking great! Really like how it combines the two pagination styles (previous/next vs page numbers).

My initial instinct was that these were different enough to be two separate components, but there’s enough similarity that bringing them together makes sense – especially as the arrow styles have been unified.

src/govuk/components/pagination/template.njk
```diff
@@ -0,0 +1,62 @@
...
+        {%- if params.previous.label %}
+        <span class="govuk-pagination__link-divider govuk-visually-hidden">:</span>
+        <span class="govuk-pagination__link-label">{{ params.previous.label }}</span>
+        {% endif -%}
+      </a>
+    </li>
+    {% endif -%}
+
+    {%- for item in params.items %}
+    {%- if item.type == 'elipses' %}
```
frankieroberto, line 25: Could `type` be avoided by just having items that have a `text` param but no `href`, and then passing in the elipses character within the text param?

src/govuk/components/pagination/template.njk
```diff
@@ -0,0 +1,62 @@
...
+    <li class="govuk-pagination__item govuk-pagination__item--elipses">&hellip;</li>
+    {% else %}
+    {%- if item.selected %}
```
frankieroberto, line 28: `item.active` maybe (to match the class name), or `item.current` (to match the text)?

src/govuk/components/pagination/_index.scss
```diff
@@ -0,0 +1,170 @@
...
+      margin-top: 6px;
+      margin-right: govuk-spacing(3);
+      fill: $govuk-secondary-text-colour;
+    }
+  }
+
+  .govuk-pagination__item--next .govuk-pagination__icon {
+    margin-right: 0;
+    margin-left: govuk-spacing(3);
+    float: right;
```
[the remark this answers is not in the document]
gregtyler: Could also put the SVG _after_ the text, and forgo any floats, but I assume it's before the text for a reason.

src/govuk/components/pagination/_index.scss
```diff
@@ -0,0 +1,170 @@
...
+      display: inline-block;
+      margin-bottom: 0;
+      //vertical-align: middle;
+    }
+  }
+
+  .govuk-pagination__item {
+    @include govuk-font(19);
+    display: inline-block;
+    margin-bottom: govuk-spacing(4);
```
gregtyler, line 47: Could we move the bottom margin up to `.govuk-pagination__list`? I believe that would mostly work the same, but it would remove what I think of as an awkward gap when the items are stacked:

| Before | After |
| ----- | ----- |
| image | image |

src/govuk/components/pagination/_index.scss
```diff
@@ -0,0 +1,170 @@
...
+    min-width: 45px;
+    min-height: 45px;
+    padding: 10px 18px;
+    color: $govuk-secondary-text-colour;
+    font-weight: bold;
+    text-align: center;
+  }
+
+  .govuk-pagination__item--elipses {
+    margin: 0 -22.5px;
```
gregtyler, line 91: This is a really neat way of handling ellipses spacing, but it does interfere with the touch area of the preceding link:
image

You can see this when hovering over the right-hand-side of the preceding page link.

I believe this can be avoided by adding `pointer-events: none;`, which ultimately causes any hovers/clicks/touches to apply to the element beneath.

src/govuk/components/pagination/pagination.yaml
```diff
@@ -0,0 +1,198 @@
...
+    description: The next page URL.
+- name: items
+  type: array
+  required: true
+  description: Array of link objects.
+  params:
+  - name: type
+    type: string
+    required: false
+    description: Set to 'elipses' to indicate where a list has been truncated.
```
gregtyler, line 40: "Ellipses" is spelt with two `l`s, which I think needs to be corrected here, in the CSS and in the Nunjucks template.

src/govuk/components/pagination/pagination.yaml
```diff
@@ -0,0 +1,198 @@
...
+  - name: type
+    type: string
+    required: false
+    description: Used to describe the component to screen-readers and in the results text. Defaults to 'results'.
+
+examples:
+- name: default
+  data:
+    previous:
+      href: '/page=1'
```
gregtyler, line 79: This is picky—and I know it's what the MOJ Pattern Library does too—but I'd personally prefer these to be a more representative URL, like `/?page=1`. I worry that because `/page=1` isn't a common URL pattern, it might look like it has some special meaning.

src/govuk/components/pagination/template.njk
gregtyler: Could we allow `params.previous.html` so that people have more control over how the hidden and visible text work together?

For example, if translating into Spanish you'd want to use `<span class="govuk-visually-hidden">página</span> anterior`, which isn't possible with either of the suggested patterns but could be done if you could provide custom HTML.

On the other point: I like the idea of sensible defaults for these kinds of values because it helps keep the components accessible-by-default. I've seen a few accessibility issues in services where someone's used the Design System components and failed to set a necessary parameter. If we can help them out-of-the-box, but provide enough configurability to make choices, I think we should.

src/govuk/components/pagination/template.njk
gregtyler: I understand the main reason to use `&hellip;` is to ensure it works if ever in a non-UTF8 encoded context, though I don't know how common that is these days.

src/govuk/components/pagination/template.njk
gregtyler: I like the idea of custom HTML, since those parameters aren't used for anything else and it would allow for customisation.

If you do keep the embedded parameters though, they should use `<strong>` tags rather than `<b>`

src/govuk/components/pagination/_index.scss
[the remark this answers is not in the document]
timpaul: Thanks both! @gregtyler you're right - the only reason it's before the text is so we can use the same markup for both variants of the component. But we don't _have_ to do that - we could adjust the order of the HTML elements in the macro instead.

For now, I'll try @frankieroberto's suggestion, but we can always go down the macro route if we want to.

src/govuk/components/pagination/_index.scss
timpaul: Good detective work! This part of the code was inherited from the MoJ Design System - I'd not stopped to figure out what it was doing. I'd have to get a second opinion about whether table layout could work too - I wonder if @adamsilver or @edwardhorsford have a view on the idea of a stacked layout?

src/govuk/components/pagination/template.njk
timpaul: Really good point about translation order - I think you're right that we should offer html here, as we do with most components.

The issue of sensible defaults is a thorny one! I think I'll leave this thread here as a prompt to have that conversation with the team - feels like something we should agree as a general principle. I do agree though, that the current defaults, whilst handy, are a bit too 'magic'.

src/govuk/components/pagination/template.njk
timpaul: Thanks both - again good points. I'll leave this thread here for when we finalise the macro design with the team.

src/govuk/components/pagination/template.njk
timpaul: Not as it stands I think - the elipses <li> has its own tag and custom styles.

src/govuk/components/pagination/template.njk
timpaul: Good spot - I've gone with `current` for the yaml and sass

src/govuk/components/pagination/_index.scss
timpaul: Hey Greg - I'd ha a similar thought, but at the moment that's how it works on GOV.UK guidance pages, and I wanted this first iteration to track that design as closely as possible.

src/govuk/components/pagination/_index.scss
timpaul: Ooh, great point, thanks. I'll try your suggestion. Any idea how to get the z-index of the elipses above the hover states of both neighbouring links?

src/govuk/components/pagination/pagination.yaml
timpaul: Good spot! Fixed

src/govuk/components/pagination/pagination.yaml
timpaul: Thanks - actually, there's another issue too, which is that if you use these examples directly in the Design System website, it fails our link checker. For this reason, all the examples in the design system currently use `/#`. I think, for that reason, I'll update them to # - but perhaps a future update would be to adjust our link checker

src/govuk/components/pagination/_index.scss
gregtyler: I had a little try with z-index, but I think it only works on absolutely-positioned elements. `pointer-events` seemed a little neater, though I've just realised it's not supported in IE < 11.

Jon-Rowe-HMRC:
Hi,

We’ve also been working on a pagination pattern at HMRC and were close to publishing. It makes more sense for it to be in the main design system though, so we’ve paused for now.

Here’s the draft HMRC pattern: https://github.com/hmrc/design-patterns/issues/172 

Our main driver for taking on the pattern was to share accessibility improvements we’d made to pre-existing patterns shared by other departments. We also wanted to centralise the logic for link list truncation so this wouldn’t be redeveloped by separate teams.

We’re happy to discuss (our front-end dev is Tim Bryan, and Adam Liptrot has been looking at this from an accessibility perspective). Our main question though, is when do you think the Design System pattern will be published? If it’s unlikely to be soon, we might still publish the HMRC version. But if it is soon, we’ll probably not publish and remove the HMRC pattern from our backlog.

Best, Jon

src/govuk/components/pagination/_index.scss
```diff
@@ -0,0 +1,181 @@
+@include govuk-exports("govuk/component/pagination") {
+  .govuk-pagination {
+    border-top: 1px solid $govuk-border-colour;
```
simonwhatley, line 3: @timpaul why is there a top border in the component? I don't think it's needed. It also feels like something that would be frequently overridden depending on the needs of the service.

owenatgov:
Prototype here superseded by production PR for pagination: https://github.com/alphagov/govuk-frontend/pull/2610 closing this PR.

PR #2422, opened 2021-11-08
Add footer alignment examples

lfdebrux:
We want the columns in our footers and our grids to align, however currently they do not in several scenarios.

This commit adds an example at [/examples/footer-alignment](https://govuk-frontend-pr-2422.herokuapp.com/examples/footer-alignment) with several combinations of grid and footer layouts to act as a quick visual reference for whether the alignment is correct or not. The examples were based on a [comment in PR #1542][1].  This should aid in developing a new footer layout.

[1]: https://github.com/alphagov/govuk-frontend/pull/1542#issuecomment-526201629


vanitabarrett (approved):
This is so helpful! 🎉

Issue #1729, opened 2020-02-10
Run installation docs research


RosieClayton:
## Why
Testing the install docs work 
## Who needs to know about this
Mark green and the devs 
## Done when
We have tested the docs with a range of users 
- [x] run research session 
- [x] book in analysis session

PR #2423, opened 2021-11-08
Update footer to use CSS grid instead of flexbox

lfdebrux:
This PR resurrects a spike by @nickcolley in 2019 and rebases it on top of our code today.

I've done a little testing and I think it fixes issues #1539 and #1729 (with the addition of a span parameter).

I'm opening this draft PR now as I'd like to pair with another dev to explore it further and find any pitfalls early.

### DIFF
--- a/src/govuk/components/footer/_index.scss
+++ b/src/govuk/components/footer/_index.scss
@@ -145,10 +145,16 @@
   }
 
   .govuk-footer__navigation {
-    display: flex; // Support: Flexbox
     margin-right: -$govuk-gutter-half;
     margin-left: -$govuk-gutter-half;
-    flex-wrap: wrap; // Support: Flexbox
+  }
+
+  @supports (display: grid) {
+    .govuk-footer__navigation {
+      $minimum-column-size: 250px;
+      display: grid; // Support: Grid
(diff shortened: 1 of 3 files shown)


lfdebrux:
Screenshot of some previous GOV.UK footer layouts with the new CSS, based on comments from #1729

Screenshot 2021-11-08 at 11 04 52

Screenshot 2021-11-08 at 11 06 51

lfdebrux:
One of the immediate disadvantages of using CSS grids is the same CSS won't work with IE11: this screenshot of the code in d0b4bd6a668f28f1ffcaf69a55f83d7eb11b159a in IE11 shows that the footer sections are all bunched up to the left and mis-aligned. It might be possible however to add some IE11 specific CSS that uses the version of grid that IE11 does support, but it won't look perfect in IE8 or 9.

Screenshot 2021-11-08 at 11 32 48

lfdebrux:
It looks like support for a grid-based footer in IE11 is going to be severely constrained by the fact that Internet Explorer doesn't support auto-placement:

> The main thing to know right now is that if you want to use CSS grid in an IE-friendly way, you should only ever use it if there are a known number of cells for a known number of rows and columns.

So we'd probably need totally different CSS for IE if we wanted to get the alignment right.

Issue #1836, opened 2020-06-12
JavaScript modules are attached to the same global variable when imported individually


alex-ju:
### Context

[When we compile the JavaScript assets to UMD we attach them to `GOVUKFrontend`](https://github.com/alphagov/govuk-frontend/blob/master/tasks/gulp/compile-assets.js#L160-L167), which [works well when importing all components](https://github.com/alphagov/govuk_publishing_components/pull/1472/files#diff-b7cab78200d934a1e9398c273d9f5be5L3), but not so great [when importing individual scripts](https://github.com/alphagov/govuk_publishing_components/pull/1472/files#diff-3fd0e8584efb3f8d58f4656b5dc2c618R5).

### Problem
For example, in GOV.UK Publishing Components, when importing multiple components individually from `govuk-frontend` only the last one will be available.
```ruby
//= require govuk/components/accordion/accordion.js
# accordion.js is exposed via window.GOVUKFrontend; not GOVUKFrontend.Accordion as expected
//= require govuk/components/accordion/button.js
# button.js is exposed via window.GOVUKFrontend; accordion.js is lost
```
As a workaround, [we attached the component to another global](https://github.com/alphagov/govuk_publishing_components/pull/1472/files#diff-3fd0e8584efb3f8d58f4656b5dc2c618R5) `GOVUK.Modules.[ComponentName]` right after importing it.

### Potential solution
Update [compile-assets.js](https://github.com/alphagov/govuk-frontend/blob/master/tasks/gulp/compile-assets.js) script to attach components to `GOVUKFrontend.[ComponentName]` instead of `GOVUKFrontend`.

## Done when

- [ ] Importing multiple individual component JavaScript modules no longer overwrites the previous one
- [x] "Details of breaking change" section below completed to help with writing release notes

## Details of breaking change

- which users are affected: users who import component JavaScript individually
- the change that’s been made: component JavaScript is now attached to `window.GOVUKFrontend.[ComponentName]` rather than `window.GOVUKFrontend` to avoid multiple imports for individual components overwriting each other
- changes users will have to make: any users who important component JavaScript individually may have written their own code to work around this problem or may be relying on the JavaScript being available with `window.GOVUKFrontend`. Users must update references to account for the new naming, for example switching calls to `window.GOVUKFrontend` to `window.GOVUKFrontend.[ComponentName]`
- impact of users not making those changes: their JavaScript code may error as it's unable to find the correct module

36degrees:
[...]

PR #2426, opened 2021-11-09
Rename exported JavaScript modules to include component name

vanitabarrett:
Closes https://github.com/alphagov/govuk-frontend/issues/1836

## What
Change the `window` global and UMD export name for component JavaScript to include the component name, for example: `GOVUKFrontend.CharacterCount`.

The way this PR currently stands, this new naming applies only to component JavaScript files - not to`all.js`, `common.js` or any vendor/polyfills. Because `all.js` already exports multiple modules, we can keep the original naming of `GOVUKFrontend` and modules will already be automatically nested underneath, e.g: `window.GOVUKFrontend.initAll()`. If the new naming system was also applied to `all.js`, you would end up having to call `window.GOVUKFrontend.All.initAll()`.

This means that if a user imports a specific component JavaScript (e.g: character-count) and then`all.js` -  `all.js` will overwrite any existing modules attached to `window.GOVUKFrontend`. This feels like an unlikely scenario, but I'm not sure I know enough about how users are importing our JavaScript to know if applying this change only to component JavaScript is likely to cause problems - **welcome any thoughts/insights on this**.

## Why
When importing individual scripts, each one was attached to `window.GOVUKFrontend`. This meant that if you imported (for example) the accordion and then imported the button JavaScript, one overwrote the other so that only the button JavaScript would be attached to `window.GOVUKFrontend`.

## Breaking Changes
**Again - I welcome any thoughts/insights on this as I'm not 100% confident I know enough about how users are using our JavaScript to account for all changes that might need to be made**

If a user is importing all our JavaScript, I don't think they'll need to make any changes.

If a user is importing individual component JavaScript, then they will need to switch any calls to `window.GOVUKFrontend` to be to `window.GOVUKFrontend.[ComponentName]`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,6 +79,29 @@ If you see unexpected behaviour, make sure the `id` for the  textarea is unique 
 
 This change was introduced in [pull request #2408: Prevent issues with character count when textarea `id` includes CSS syntax characters](https://github.com/alphagov/govuk-frontend/pull/2408).
 
+#### Make sure individually imported JavaScript modules work as expected
+
+You do not need to do anything if you have either:
+
+- followed our [Getting Started guide](https://frontend.design-system.service.gov.uk/get-started/#5-get-the-javascript-working) and are importing all of the GOV.UK Frontend JavaScript in one go via `all.js`
+- installed GOV.UK Frontend using precompiled files
+
+We've changed the naming of our components' JavaScript modules so that individual imports are now attached to
+`window.GOVUKFrontend.[ComponentName]` instead of `window.GOVUKFrontend`.
+
+You can now import multiple modules without overwriting the previous one, for example:
+
+```
+//= require govuk/components/accordion/accordion.js
+//= require govuk/components/button/button.js
+
+# These modules are available under window.GOVUKFrontend.Accordion and window.GOVUKFrontend.Button respectively
+```
+
+If you're importing JavaScript modules individually, you should check any references to `window.GOVUKFrontend` in your code and update them to point to the correct component, `window.GOVUKFrontend.[ComponentName]`. You can now remove any workarounds you may have implemented.
+
+This change was introduced in [pull request #1836: Rename exported JavaScript modules to include component name](https://github.com/alphagov/govuk-frontend/issues/1836)].
+
 #### Remove calls to deprecated `iff` Sass function
 
 We've removed the `iff` function which we deprecated in [GOV.UK Frontend version 3.6.0](https://github.com/alphagov/govuk-frontend/releases/tag/v3.6.0).
@@ -149,6 +172,7 @@ We’ve made fixes to GOV.UK Frontend in the following pull requests:
 - [#2370: Prevent issues with conditionally revealed content when content `id` includes CSS syntax characters](https://github.com/alphagov/govuk-frontend/pull/2370)
 - [#2408: Prevent issues with character count when textarea `id` includes CSS syntax characters](https://github.com/alphagov/govuk-frontend/pull/2408)
 - [#2434: Add brand colour for Department for Levelling Up, Housing and Communities (DLUHC)](https://github.com/alphagov/govuk-frontend/pull/2434)
+- [#1836: Rename exported JavaScript modules to include component name](https://github.com/alphagov/govuk-frontend/issues/1836))
 
 ## 3.14.0 (Feature release)
 
--- a/tasks/gulp/__tests__/after-build-package.test.js
+++ b/tasks/gulp/__tests__/after-build-package.test.js
@@ -9,11 +9,13 @@ var glob = require('glob')
 
 const configPaths = require('../../../config/paths.json')
 const lib = require('../../../lib/file-helper')
+const { componentNameToJavaScriptModuleName } = require('../../../lib/helper-functions')
 
 const { renderSass } = require('../../../lib/jest-helpers')
 
 const readFile = util.promisify(fs.readFile)
 const componentNames = lib.allComponents.slice()
+const componentsWithJavaScript = glob.sync(configPaths.package + 'govuk/components/' + '**/!(*.test).js')
 
 describe('package/', () => {
   it('should contain the expected files', () => {
@@ -101,6 +103,19 @@ describe('package/', () => {
     })
   })
 
+  describe('all.js', () => {
+    it('should have correct module name', async () => {
+      const allJsFile = path.join(configPaths.package, 'govuk', 'all.js')
+      return readFile(allJsFile, 'utf8')
+        .then((data) => {
+          expect(data).toContain("typeof define === 'function' && define.amd ? define('GOVUKFrontend', ['exports'], factory)")
+        })
+        .catch(error => {
+          throw error
+        })
+    })
+  })
+
   describe('component', () => {
     it.each(componentNames)('\'%s\' should have macro-options.json that contains JSON', (name) => {
       const filePath = path.join(configPaths.package, 'govuk', 'components', name, 'macro-options.json')
@@ -120,6 +135,20 @@ describe('package/', () => {
               })
             )
           })
+        })
+        .catch(error => {
+          throw error
+        })
+    })
+  })
+
+  describe('components with JavaScript', () => {
+    it.each(componentsWithJavaScript)('\'%s\' should have component JavaScript file with correct module name', (javaScriptFile) => {
+      const moduleName = componentNameToJavaScriptModuleName(path.parse(javaScriptFile).name)
+
+      return readFile(javaScriptFile, 'utf8')
+        .then((data) => {
+          expect(data).toContain("typeof define === 'function' && define.amd ? define('" + moduleName + "', factory)")
         })
         .catch(error => {
           throw error
--- a/tasks/gulp/compile-assets.js
+++ b/tasks/gulp/compile-assets.js
@@ -1,6 +1,6 @@
 'use strict'
 
-const helperFunctions = require('../../lib/helper-functions')
+const { componentNameToJavaScriptModuleName } = require('../../lib/helper-functions')
 
 const path = require('path')
 
@@ -189,18 +189,28 @@ gulp.task('scss:compile', function (done) {
 // Compile js task for preview ----------
 // --------------------------------------
 gulp.task('js:compile', (done) => {
-  // for dist/ folder we only want compiled 'all.js'
+  // For dist/ folder we only want compiled 'all.js'
   const fileLookup = isDist ? configPaths.src + 'all.js' : configPaths.src + '**/!(*.test).js'
+
+  // Perform a synchronous search and return an array of matching file names
   const srcFiles = glob.sync(fileLookup)
 
   srcFiles.forEach(function (file) {
-    const currentDirectory = path.dirname(file)
-    const newDirectory = currentDirectory.replace('src/govuk', '')
+    // This is combined with desinationPath in gulp.dest()
+    // so the files are output to the correct folders
+    const newDirectoryPath = path.dirname(file).replace('src/govuk', '')
+
+    // We only want to give component JavaScript a unique module name
+    let moduleName = 'GOVUKFrontend'
+    if (path.dirname(file).includes('/components/')) {
+      moduleName = componentNameToJavaScriptModuleName(path.parse(file).name)
+    }
 
     return gulp.src(file)
       .pipe(rollup({
-        // Used to set the `window` global and UMD/AMD export name.
-        name: 'GOVUKFrontend',
+        // Used to set the `window` global and UMD/AMD export name
+        // Component JavaScript is given a unique name to aid individual imports, e.g GOVUKFrontend.Accordion
+        name: moduleName,
         // Legacy mode is required for IE8 support
         legacy: true,
         // UMD allows the published bundle to work in CommonJS and in the browser.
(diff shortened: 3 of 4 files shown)


package.json
```diff
@@ -33,7 +33,7 @@
     "express": "^4.16.4",
     "express-validator": "^6.6.1",
     "front-matter": "^4.0.2",
-    "glob": "^7.1.4",
+    "glob": "^7.2.0",
```
lfdebrux, line 36: Why is the bump to gulp needed?

lfdebrux:
I'm not sure I understand what is going on in the first commit... would you mind expanding a bit on why you're making the change and what it means? 😅

tasks/gulp/compile-assets.js
```diff
@@ -183,33 +184,51 @@ gulp.task('scss:compile', function (done) {
   done()
 })
 
+function convertFileNameToModuleName (filename) {
```
lfdebrux, line 187: I think a doc string with an example would be really helpful here

package.json
vanitabarrett: This is a bump to `glob` - not necessary for this work, it just bumps it to the latest version

lfdebrux: Oh yeah, force of habit 😅 

It might be good to split out the bump into a separate commit or PR, just in case it happens to break something?

vanitabarrett: Good shout, I've removed it from this PR and I'll raise a separate PR for that

tasks/gulp/compile-assets.js
vanitabarrett: I've added some more code comments to this function - is that helpful?

https://github.com/alphagov/govuk-frontend/pull/2426/files#diff-ded529645ff40394f7d2c28c16f2907dc01a2760fdd702a2aecae639838c8fc1R187

lfdebrux: Ah yeah that's very helpful, thanks!

vanitabarrett:
@lfdebrux I'll try to explain it more here, let me know if it's any more helpful! If not, happy to jump on a call to chat about it.

The current gulp task as it is on `main` sets up an array of JavaScript files, and gulp uses streams to pass that data through the rest of the transformations (rollup, uglify etc). We need to make a change to the rollup config so that the `name` contains the component name and isn't just "GOVUKFrontend" for every single module. In order to do that, we need some way of accessing what current file we're "on"/"processing" so we can pull out the relevant component name.

As far as I can tell, there's no easy way of doing this with the current set-up as gulp streams mean you don't get that concept of 'this is the thing that's currently being processed'. I might be wrong about that, welcome to other suggestions!

So in order to get more "control" over knowing what file is currently being processed, I've introduced a `forEach` loop which loops through every JavaScript file and passes that through the same gulp process as before - but this time, gulp is being passed one file at a time rather than an array of files at the start. That means we know what file we're on and we have access to the pathname and filename etc.

lfdebrux:
> @lfdebrux I'll try to explain it more here, let me know if it's any more helpful! If not, happy to jump on a call to chat about it.
> 
> The current gulp task as it is on `main` sets up an array of JavaScript files, and gulp uses streams to pass that data through the rest of the transformations (rollup, uglify etc). We need to make a change to the rollup config so that the `name` contains the component name and isn't just "GOVUKFrontend" for every single module. In order to do that, we need some way of accessing what current file we're "on"/"processing" so we can pull out the relevant component name.
> 
> As far as I can tell, there's no easy way of doing this with the current set-up as gulp streams mean you don't get that concept of 'this is the thing that's currently being processed'. I might be wrong about that, welcome to other suggestions!
> 
> So in order to get more "control" over knowing what file is currently being processed, I've introduced a `forEach` loop which loops through every JavaScript file and passes that through the same gulp process as before - but this time, gulp is being passed one file at a time rather than an array of files at the start. That means we know what file we're on and we have access to the pathname and filename etc.

Ah, I get it now, thank you!

CHANGELOG.md
```diff
@@ -65,6 +65,26 @@ Replace any use of this class with the `govuk-tag--grey` class.
 
 This change was introduced in [pull request #2417: Remove deprecated `govuk-tag--inactive class`](https://github.com/alphagov/govuk-frontend/pull/2417).
 
+#### Component JavaScript modules have a unique name when imported individually
```
36degrees, line 68: I _think_ we've sort of settled on using a description of 'what the user needs to do' for breaking changes (for example 'Make sure character counts still work as expected') – is there an equivalent we could use for this?

I also think we might want to list this under 'Fixes' as well, as it feels like a bug?

36degrees (commented):
It'd be nice to have tests to cover at least some of this, but no idea how to make that happen…

Maybe we could add a test to `after-build-package.test.js` that looks at the component JS and asserts something about [the exported name](https://github.com/alphagov/govuk-frontend/blob/f281cd247c6b3b4f1ca689a0e92d14c31aa95519/package/govuk/components/accordion/accordion.js#L1-L5)?

tasks/gulp/compile-assets.js
```diff
@@ -183,33 +184,61 @@ gulp.task('scss:compile', function (done) {
...
-    .pipe(eol())
-    .pipe(gulp.dest(destinationPath))
+gulp.task('js:compile', (done) => {
+  // For dist/ folder we only want compiled 'all.js'
+  const fileLookup = isDist ? configPaths.src + 'all.js' : configPaths.src + '**/!(*.test).js'
+
+  // Perform a synchronous search and return an array of matching file names
+  const srcFiles = glob.sync(fileLookup)
+
+  srcFiles.forEach(function (file) {
```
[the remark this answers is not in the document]
lfdebrux: Yeah I'd like to see us using less gulp rather than more :/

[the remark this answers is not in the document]
vanitabarrett: Yeah, I came across a few things when I was researching but most gulp plugins seem very unloved and outdated now. Coupled with the fact that I don't think we have great gulp knowledge on the team and there's probably other ways of doing things now, I'm trying to avoid plugging in any new gulp things.

CHANGELOG.md
vanitabarrett: @EoinShaughnessy I've had a go at the release notes for this (including Ollie's suggestion here), but I think they could do with some love! Would it be helpful if I put some time in to talk through it together?

EoinShaughnessy: @vanitabarrett That'd be great, thanks!

vanitabarrett: @EoinShaughnessy Let's have a chat about it at our catch-up tomorrow 👍

tasks/gulp/__tests__/after-build-package.test.js
```diff
@@ -9,11 +9,13 @@ var glob = require('glob')
 
 const configPaths = require('../../../config/paths.json')
 const lib = require('../../../lib/file-helper')
+const helperFunctions = require('../../../lib/helper-functions')
```
36degrees, line 12: Would it be worth importing just the function we need, so we can remove the namespacing?

```suggestion
const { componentNameToJavaScriptModuleName } = require('../../../lib/helper-functions')
```

tasks/gulp/compile-assets.js
```diff
@@ -1,5 +1,7 @@
 'use strict'
 
+const helperFunctions = require('../../lib/helper-functions')
```
36degrees, line 3: Likewise here:

```suggestion
const { componentNameToJavaScriptModuleName } = require('../../lib/helper-functions')
```

36degrees (commented):
Looks good to me, just one suggestion re how we import the `componentNameToJavaScriptModuleName` method.

tasks/gulp/__tests__/after-build-package.test.js
vanitabarrett: Good idea 👍

36degrees (approved):
Looks good to me.

Have manually run the `build:dist` and `build:package` tasks and the output looks like what I'd expect – although it looks like using the `dist` version of the JS the modules are not attached to GOVUKFrontend, but I guess that's expected? It doesn't appear to be a change from the existing behaviour, anyway.

CHANGELOG.md
```diff
@@ -65,6 +65,26 @@ Replace any use of this class with the `govuk-tag--grey` class.
 
 This change was introduced in [pull request #2417: Remove deprecated `govuk-tag--inactive class`](https://github.com/alphagov/govuk-frontend/pull/2417).
 
+#### Make sure any individually imported JavaScript modules work as expected
```
EoinShaughnessy, line 68:
```suggestion
#### Make sure individually imported JavaScript modules work as expected
```

CHANGELOG.md
```diff
@@ -65,6 +65,26 @@ Replace any use of this class with the `govuk-tag--grey` class.
...
+
+We've changed the naming of our component JavaScript modules so that individual imports are now attached to
+`window.GOVUKFrontend.[ComponentName]` instead of `window.GOVUKFrontend`.
+
+This means you can now import multiple modules without overwriting the previous one, for example:
```
EoinShaughnessy, line 73:
```suggestion
You can now import multiple modules without overwriting the previous one, for example:
```

CHANGELOG.md
```diff
@@ -65,6 +65,26 @@ Replace any use of this class with the `govuk-tag--grey` class.
...
+
+```
+//= require govuk/components/accordion/accordion.js
+//= require govuk/components/button/button.js
+
+# The above are available under window.GOVUKFrontend.Accordion and window.GOVUKFrontend.Button respectively
```
EoinShaughnessy, line 79:
```suggestion
# These modules are available under window.GOVUKFrontend.Accordion and window.GOVUKFrontend.Button respectively
```

CHANGELOG.md
```diff
@@ -65,6 +65,26 @@ Replace any use of this class with the `govuk-tag--grey` class.
...
+```
+
+If you are importing indvidual component JavaScript, you should check any references to `window.GOVUKFrontend` in your code and update them to point to the correct component: `window.GOVUKFrontend.[ComponentName]`. You may have implemented workarounds that can now be removed.
```
EoinShaughnessy, line 82:
```suggestion
If you're importing individual component JavaScript, you should check any references to `window.GOVUKFrontend` in your code and update them to point to the correct component: `window.GOVUKFrontend.[ComponentName]`. You can now remove any workarounds you may have implemented.
```
Wonder if we could change "If you're importing individual component JavaScript" to "If you're importing JavaScript modules individually" (to somewhat mirror wording in heading title)? 
Also, should we tell users "you can" or "you should" now remove any workarounds?

CHANGELOG.md
```diff
@@ -65,6 +65,26 @@ Replace any use of this class with the `govuk-tag--grey` class.
...
+
+This change does not affect you if you followed our [Getting Started guide](https://frontend.design-system.service.gov.uk/get-started/#5-get-the-javascript-working) and are importing all of the GOV.UK Frontend JavaScript in one go via `all.js`, or if you installed GOV.UK Frontend using precompiled files.
```
EoinShaughnessy, line 84: Wonder should this display right under the heading? Just to help users who do not need to do anything skip over this content.

```suggestion
You do not need to do anything if you have either:

- followed our [Getting Started guide](https://frontend.design-system.service.gov.uk/get-started/#5-get-the-javascript-working) and are importing all of the GOV.UK Frontend JavaScript in one go via `all.js`
- installed GOV.UK Frontend using precompiled files

```

CHANGELOG.md
```diff
@@ -65,6 +65,26 @@ Replace any use of this class with the `govuk-tag--grey` class.
...
+
+This change was introduced in [pull request #1836: Rename exported JavaScript modules to include component name](https://github.com/alphagov/govuk-frontend/issues/1836)]
```
EoinShaughnessy, line 86:
```suggestion
This change was introduced in [pull request #1836: Rename exported JavaScript modules to include component name](https://github.com/alphagov/govuk-frontend/issues/1836)].
```

EoinShaughnessy (commented):
@vanitabarrett Looks great! Just a few tiny comments.

CHANGELOG.md
vanitabarrett: Good idea, I've moved this to the top and made the suggestions you mentioned

CHANGELOG.md
```diff
@@ -65,6 +65,26 @@ Replace any use of this class with the `govuk-tag--grey` class.
 
 This change was introduced in [pull request #2417: Remove deprecated `govuk-tag--inactive class`](https://github.com/alphagov/govuk-frontend/pull/2417).
 
+#### Make sure any individually imported JavaScript modules work as expected
+
+We've changed the naming of our component JavaScript modules so that individual imports are now attached to
```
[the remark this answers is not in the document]
vanitabarrett: Yeah, I found it a bit difficult to find the right wording here! At the moment, I can't think of alternative though 🤔

EoinShaughnessy (approved):
@vanitabarrett Looks good!

vanitabarrett:
Thanks for your reviews @EoinShaughnessy and @36degrees ! I'm going to hold off merging this for a bit as I've asked @alex-ju to test this change in the GOV.UK Publishing Components to double check it works as expected.

alex-ju:
Many thanks for the work on this, Vanita! 🙌 
Tested this change in https://github.com/alphagov/govuk_publishing_components/pull/2457 and it works as expected.
This approach – with `window.GOVUKFrontend.initAll()` (when importing everything) and `window.GOVUKFrontend.Button(element).init()` (for individual imports) – is what I would expect as a consumer.

Issue #1726, opened 2020-02-06
Improve the layout options for the footer component.


maxgds:
## What

Let the layout for the contents of a footer be more flexible to support more potential footer layouts. 

## Why
I'm currently trying to rebuild the GOV.UK website's existing footer using this component but the desired layout doesn't seem to be possible. 

## Further detail
The live site (http://www.gov.uk) looks like this:

![Screenshot 2020-02-06 at 10 00 09](https://user-images.githubusercontent.com/31649453/73952775-4af65100-48f7-11ea-9ba7-bb89ca351349.png)

During different iterations of this footer the first section has contained a Brexit header with many list items in 2 columns, and as now, a Transition header with a single list item. In both cases this section should take up 2/3 of the space and line up with the next section, which also takes 2/3 of the space and is followed by a 3rd section taking the remainder of that pseudo-row. The problem is that using the footer component classes Section 1 and Section 2 don't align properly:

![Screenshot 2020-02-06 at 10 01 21](https://user-images.githubusercontent.com/31649453/73954337-9578cd00-48f9-11ea-94e2-60dc574aa3e4.png)

This layout could be easily achieved using the design system layout classes, but it seems wrong to try to do that when the design system has implemented specific footer classes. Is there a reason why it is using flexbox instead of the custom classes? 

Having sought feedback from design they are keen to keep the layout as-is in our design so we may be forced into using the non-footer approach.

## Done when

- [ ] "Details of breaking change" section below completed to help with writing release notes

## Details of breaking change

- which users are affected: TBC
- the change that’s been made: TBC
- changes users will have to make: TBC
- impact of users not making those changes: TBC

nickcolley:
We're aware of some [other footer layout issues](https://github.com/alphagov/govuk-frontend/issues/1539) that we want to address so this is helpful.

The footer doesnt use the regular grid layout classes so that it flows better between breakpoints but I think this has made it more difficult to extend and use outside how we expected it to be used.

So I think when we address the other footer layout issues we'll want to consider taking a step away from flexbox and doing something simpler like you have suggested, perhaps exploring CSS Grid.

It would be nice to ensure that content in the footer columns have opportunity to determine how this component responds when on smaller screens but our priority is making sure it aligns with the overall grid so we'll keep this in mind.

nickcolley:
One option is what you suggested is to use the grid classes:

```html
<div class="govuk-width-container">
[...]

Issue #1680, opened 2019-12-19
Improvements to the footer component


36degrees:
The footer has a number of (~minor) layout issues that are hard to address without making changes to the markup. It also has a number of hardcoded English strings which we should address, aligning with supporting multiple languages.

Some of these fixes will involve making breaking changes.

## Stories

- [ ] Long text in footer columns do not wrap #1647
- [ ] Footer layout isn't aligned with content layout #1539
- [ ] Safari alignment issue with Footer list containing odd number of items #1284
- [ ] Add ability to override or remove OGL logo in Footer #1291
- [ ] Add ability to override or remove Crown Copyright in Footer #1290

DouglasCaven:
The new version of the Axe Review accessibility tool has highlighted a potential issue suggesting the ‘Crown copyright’ logo in the footer must have sufficient colour contrast. The failure is actually the logo and the colour of text passed and thought it should be noted.

CharlotteDowns:
## The problem
![current Crown copyright logo in the footer](https://user-images.githubusercontent.com/56260216/106613192-d7b30980-6561-11eb-8f5b-9ffa8df5c47d.jpg)
The crown copyright logo has a hex code colour of `#6E777A`, which against the footer background of `#F3F2F1` has a AA Large 4.09:1 colour contrast ratio. 


## A potential fix
![Crown copyright logo in the footer using the darker text colour](https://user-images.githubusercontent.com/56260216/106612745-578ca400-6561-11eb-9b1e-a2255c12ea56.jpg)
Here is a mock-up of the crown copyright logo using the footer text colour `#454A4C` which has a contrast ratio of AAA 8.03:1 against the footer background `#F3F2F1`.

This approach may not meet the brand use of the Crown copyright logo and should be looked into.

36degrees:
Closing this as having an epic for the component isn't currently useful for us.

PR #2428, opened 2021-11-10
Fix footer alignment with grid classes

lfdebrux:
Fixes https://github.com/alphagov/govuk-frontend/issues/1539. It also partly addresses https://github.com/alphagov/govuk-frontend/issues/1726, along with https://github.com/alphagov/govuk-frontend/pull/2446

Unfortunately this PR does not fix any of the other bugs reported in https://github.com/alphagov/govuk-frontend/issues/1680

## What
Adjust the styles of the footer layout to better match up with the grid classes. The screenshot below shows the grid classes and the footer not aligning properly:

mac-chrome-left-before

**Note:** It's worth comparing and contrasting these changes with those in branch [`nickcolley-ldeb-spike-footer-grid`](https://github.com/alphagov/govuk-frontend/tree/nickcolley-ldeb-spike-footer-grid), which achieves much the same effect with fewer lines and fewer commits by using CSS grids. The CSS grid version also behaves slightly better as the viewport changes size. However, CSS grids are not supported by all browsers we want to support, so I think this PR that uses flexbox should take precedence.

Browsers to test in:

- [x] Internet Explorer 8
- [x] Internet Explorer 9
- [x] Internet Explorer 10
- [x] Internet Explorer 11
- [x] Chrome on macOS
- [x] Safari on macOS
- [x] Chrome on Windows
- [x] Firefox on macOS
- [x] Edge on Windows
- [x] Firefox on Windows
- [x] Safari on iOS
- [x] Chrome on iOS
- [x] Chrome on Android
- [x] Samsung Internet on Android

For each browser, test the two main footer examples:
 
 - https://govuk-frontend-pr-2428.herokuapp.com/examples/footer-alignment
  - https://govuk-frontend-pr-2428.herokuapp.com/components/footer
 
and zoom in to 400%.

## Screenshots
Note: this is just a sample of screenshots where the change is most visible

### IE8 GOV.UK example - Before vs After
ie8-govuk-before
ie8-govuk-after

### IE8 Two Column Left example - Before vs After
ie8-left-before
ie8-left-after

### IE8 Two Column Right example - Before vs After
ie8-right-before
ie8-right-after

### IE8 Equal Three Column example - Before vs After
ie8-three-before
ie8-three-after

### IE10 Two Column Right example - Before vs After
ie10-right-before
ie10-right-after

### DIFF
--- a/src/govuk/components/footer/_index.scss
+++ b/src/govuk/components/footer/_index.scss
@@ -145,10 +145,16 @@
   }
 
   .govuk-footer__navigation {
-    display: flex; // Support: Flexbox
     margin-right: -$govuk-gutter-half;
     margin-left: -$govuk-gutter-half;
-    flex-wrap: wrap; // Support: Flexbox
+  }
+
+  @supports (display: grid) {
+    .govuk-footer__navigation {
+      $minimum-column-size: 250px;
+      display: grid; // Support: Grid
+      grid-template-columns: repeat(auto-fit, minmax($minimum-column-size, 1fr)); // Support: Grid
+    }
   }
 
   .govuk-footer__section {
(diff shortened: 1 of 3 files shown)


vanitabarrett:
**Note:** this PR changes the behaviour slightly of what happens on screen sizes smaller than desktop. The previous behaviour gradually dropped the width of each column, e.g: from thirds, to halves, to full width. This PR changes that behaviour so it drops straight to full width. This means you can end up with quite a large full width footer on tablet, e.g:

## Example - before (tablet)
Screenshot 2021-12-01 at 14 40 21

## Example - after (tablet)
Screenshot 2021-12-01 at 14 40 05

We could make a change to drop down to 50% width on tablet. However, this would mean that columns that have perhaps been given more visual importance (see example below of the 'Coronavirus' section in the GOV.UK footer) would in some ways have less importance on tablet as the column below would sit alongside it.

## Desktop - Coronavirus column sits on its own row
Screenshot 2021-12-01 at 14 44 39

## Tablet - Coronavirus column sits alongside "Services and Information"
Screenshot 2021-12-01 at 14 44 53

vanitabarrett:
Note: because https://github.com/alphagov/govuk-frontend/pull/2446 builds off the changes made in this PR, I'll add the Changelog entry in https://github.com/alphagov/govuk-frontend/pull/2446 rather than in this PR.

lfdebrux:
@vanitabarrett 

> Note: this PR changes the behaviour slightly of what happens on screen sizes smaller than desktop. 

Do we need a designer to review this change?

vanitabarrett:
Superseded by https://github.com/alphagov/govuk-frontend/pull/2462

PR #2430, opened 2021-11-12
Select with search

frankieroberto:
This draft builds upon the Select component, enhancing it by allowing users to search for an option, and also by displaying the dropdown options in a more accessible way than the default HTML rendering.

It is based in part upon the legacy [accessible autocomplete](https://github.com/alphagov/accessible-autocomplete), but focuses only on the use case where:

* users have to select one of the options (or none)
* the number of options is small enough (under 500?) that users could realistically scan through them, and that they can all be included in the HTML without performance issues

See also [Input with suggestions](https://github.com/alphagov/govuk-frontend/pull/2453) for a similar draft component focusing on a different use case.

➡️  [Working demo](https://govuk-frontend-pr-2430.herokuapp.com/components/select/with-auto-select/preview)
➡️  [Draft guidance](https://deploy-preview-1991--govuk-design-system-preview.netlify.app/components/select/) ([Pull Request](https://github.com/alphagov/govuk-design-system/pull/1991))

## Design notes

The primary consideration for this component is designing it so that it clearly communicates to users that they must select an option from the list. The secondary consideration is making it clear that users can type some characters to filter the suggestions, making it easier to find the relevant option. This is secondary, as if it it is missed, users can still use the component by scanning through the complete list, whereas if users perceive the component as a regular free text input, they may get an error if their input does not match an option.

### Comparison with default Select behaviour

The design tries to emulate the default Select behaviour as closely as possible, by:

* having a 'down' arrow icon on the right
* revealing all the options when clicking into the component (but not when focusing it using the keyboard)
* showing a tick on the left of the option that is currently selected
* highlighting options when you navigate the list using either mouse hover or keyboard arrows, but not selecting it until you click or press enter

Deliberate differences from default Select are:

* the text in the options list uses the default font and text size, on a white background
* long options in the list wrap onto a second line instead of being truncated
* the options list pushes the rest of the content down the page, rather than appearing on top of it
* a scrollbar appears on the right of the options list if required
* the highlighted option uses the Design System focus style (yellow background) – Select default is a blue background (on Macs)

Select

### Filtering the options using text search

In addition, the component allows you to filter the options list by searching using keywords. This improves upon the default Select behaviour, where if you type some characters it will jump to the first match, but the characters you’ve entered are not shown, and will reset after a short break.

When typing to filter the list of options, the list updates after every character entered. The filtering ignores capitalisations and punctuation, and matches the start of words.

In addition, synonyms can be added for each option, which are also used for filtering. For example, this can be used to add previous names, official acronyms, or common mis-spellings.

If you type something which doesn’t match any of the options, a "No results" message is displayed (this can be customised), prompting the user to change your answer. If you move to another field or click away ('blurring' the field), then the input reverts to blank, or the previously selected option. This is to avoid misleading the user into thinking you can enter an answer which does not match an option.

Filtering

### Differences from accessible-autocomplete

The are a few ways in which this component deviates from the existing accessible-autocomplete design:

* the addition of the tick to indicate the currently selected item
* it is no longer possible to highlight two items, one using keyboard arrow and one using mouse hover – instead, only one item can be highlighted at the time
* when returning to the component where an item has been selected, clicking or using the arrows to reveal the options list shows all the options, rather than only the matching one - this makes it easier to change to a different answer
* when hovering or using the keyboard to go through the options, the highlighted option (which is focused) uses the yellow focus background common to other Design System components, and the text input does not have a yellow outline

## Accessibility considerations

This is a complex component which will need careful research to make it as accessible as possible.

So far, the code does the following:

* sets a `role` of `combobox` on the parent container
* creates an `aria-owns` relationship between the parent container and the text input
* sets `aria-haspopup` with a value of `listbox`
* updates `aria-exanded` on the container to either `true` or `false`, depending on whether the options are revealed
* sets `aria-autocomplete` to `list` on the text input
* creates a `aria-controls` relationship between the text input and the list of options
* sets `role` of `listbox` on the list of options (a `<ul>`)
* sets `role` of `option` on each of the options (`<li>`s)
* updates `aria-selected` to `true` on the option that is currently selected

The following aren’t done yet, but could be added if research shows they would be beneficial:

* A hidden (or visible?) status message indicating the number of matching options (eg "3 options match") with `aria-live` set to `polite`. The [accessible autocomplete javascript](https://github.com/alphagov/accessible-autocomplete/blob/master/src/status.js) updates this after 1400 milliseconds of no input, and alternates between two elements (A and B), presumably to workaround screen reader bugs.
* A hidden (or visible?) hint which tells screen reader users how to user the component. The [default text](https://github.com/alphagov/accessible-autocomplete/blob/master/src/autocomplete.js#L52) in accessible-autocomplete is "When autocomplete results are available use up and down arrows to review and enter to select. Touch device users, explore by touch or with swipe gestures."

## Other considerations

The filtered options list currently always appears in the same order (assumed to be alphabetical) that it appears as in the underlying `<select>` component. It might make sense to re order this so that matches closer to the start of the string, or exact matches of a synonym (such as an acronym) appear higher.
  
Unlike the pop-up on the default Select, the list of options doesn't automatically scroll if you hover near the last visible item on the list. Instead, the user has to manually scroll using mouse scroll wheel, trackpad, keyboard or clicking within the scrollbar on the right. It may make to implement some sort of auto-scrolling, but equally this could prove to be unintuitive, or pose accessibility issues for those with motor conditions.

I’ve not given this component a separate name, instead choosing to treat it as an optional enhancement of the existing Select component. My thinking is that this would help when describing it on the Design System website, as it would appear on the same page. However there are pros and cons, and it could be considered a separate component.

One other option for the future may be a mode where it enhances a Select using the more accessibile list of options, but does not offer the text filtering feature. This might be useful for scenarios where the list of options is very short, but where the compactness of the Select makes more sense than Radios.

## To do

* [ ] full accessibility testing
* [ ] add javascript tests
* [ ] copy any additional CSS modifier classes (such as `govuk-select--error`) from the `<select>` element to the generated `<input type="text">`
* [ ] add support for hint text – including supporting different hint text wording for the enhanced version vs no-js fallback
* [ ] consider changing mouse cursor when hovering over an option


36degrees:
As discussed with Frankie, I'm going to close this as it's not something we can merge right now and hasn't been touched in over a year.

‘Autocomplete’ is one of [our next priorities](https://design-system.service.gov.uk/community/upcoming-components-patterns/). When we start working on it, we'll revisit this PR and see if there’s anything we can reuse. It's already referenced from our backlog issue on autocomplete.

Issue #2358, opened 2021-09-22
Difficult to notice when cookie banner HTML classes are missing


vanitabarrett:
## What
Prompted by https://github.com/alphagov/govuk-design-system/issues/1886

When a user passes `html` to the cookie banner, they are required to add their own classes to correctly style that text (normally the `govuk-body` class). At the moment, we have some [default text styling](https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/components/cookie-banner/_index.scss#L7) on the cookie banner component which means that the text mostly looks correct even if a user forgets to add the `govuk-body` classes.

We should consider making it more obvious to a user if they have forgotten these classes, by removing the default text styling and therefore making the component look "a bit broken".

## Why
To help users identify when they have forgotten to add the appropriate classes to any HTML they pass into the cookie banner component
To make sure users are styling their cookie banner HTML correctly and avoiding accessibility issues, e.g: https://github.com/alphagov/govuk-design-system/issues/1886

## Who needs to know about this
Developers; Tech Writer

## Done when

- [ ] Default text styles are removed from the cookie banner component
- [ ] Tested how the component looks when `text` is passed in vs `html` - we should check the main message and also the confirmation messages

## Breaking Change
This is a breaking change because any users who are relying on the default styles will see unstyled text in their cookie banner when they update. Users will need to add the correct classes to style the text - this will probably involve adding the `govuk-body` class to any HTML paragraph tags they're passing through to the component.

PR #2432, opened 2021-11-16
Remove default font styles from cookie banner Sass

vanitabarrett:
Closes https://github.com/alphagov/govuk-frontend/issues/2358

## What
Remove the default font styles (setting the font and font-size) from the cookie banner Sass.

## Why
With the default font styles present, it's very difficult to spot if you forget to add the relevant `govuk-body` styles to any `html` you pass into the component. We had [missed this ourselves when we added the cookie banner to the design system](https://github.com/alphagov/govuk-design-system/issues/1886).

### Before change - cookie banner with `html`, without `govuk-body` classes
Screenshot 2021-11-16 at 15 13 59

### After change - cookie banner with `html`, without `govuk-body` classes
Screenshot 2021-11-16 at 13 02 32

### After change - cookie banner with `html`, with the correct `govuk-body` classes
Screenshot 2021-11-16 at 13 02 53

### After change - in the Design System
This work was prompted by when we forgot to [add our own styling classes to the cookie banner on the design system site](https://github.com/alphagov/govuk-design-system/issues/1886), which resulted in slightly incorrect text spacing. This is what the design system cookie banner would now look like if we forgot to add the relevant classes:

Screenshot 2021-11-16 at 13 58 39

## Details of breaking change
This is a breaking change because any users who are relying on the default styles will see unstyled text in their cookie banner when they update. Users will need to add the correct classes to style the text - this will probably involve adding the govuk-body class to any HTML paragraph tags they're passing through to the component.

- which users are affected: users who are using the cookie banner component and passing their own `html` into the cookie banner component
- the change that’s been made: we've removed the default font styles from the cookie banner component
- changes users will have to make: users will need to make sure that any HTML they pass into the cookie banner component has the appropriate classes/styles (e.g: `govuk-body`)
- impact of users not making those changes: cookie banners will appear with unstyled text

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -128,6 +128,14 @@ Replace any use of this class with the `govuk-tag--grey` class.
 
 This change was introduced in [pull request #2417: Remove deprecated `govuk-tag--inactive class`](https://github.com/alphagov/govuk-frontend/pull/2417).
 
+#### Style any custom HTML in your cookie banner
+
+We've removed the default font styles from the cookie banner Sass. This change makes it more obvious when classes and styles have not been added to any custom HTML.
+
+If you're passing custom HTML into the cookie banner component (for example, using the `html` Nunjucks parameter), you must make sure you're applying the appropriate classes and styles to that HTML, such as adding the `govuk-body` class to any paragraph tags. You must do this to prevent your cookie banner displaying with unstyled text.
+
+This change was introduced in [pull request #2432: Remove default font styles from cookie banner Sass](https://github.com/alphagov/govuk-frontend/pull/2432).
+
 ### Optional changes
 
 We've recently made some other changes to GOV.UK Frontend. While these are not breaking changes, implementing them will make your service work better.
--- a/src/govuk/components/cookie-banner/_index.scss
+++ b/src/govuk/components/cookie-banner/_index.scss
@@ -4,8 +4,6 @@
   $border-bottom-width: govuk-spacing(2);
 
   .govuk-cookie-banner {
-    @include govuk-font($size: 19);
-
     padding-top: govuk-spacing(4);
     // The component does not set bottom spacing.
     // The bottom spacing should be created by the items inside the component.


hannalaakso (approved):
Makes sense, thanks @vanitabarrett 👍

CHANGELOG.md
```diff
@@ -65,6 +65,14 @@ Replace any use of this class with the `govuk-tag--grey` class.
 
 This change was introduced in [pull request #2417: Remove deprecated `govuk-tag--inactive class`](https://github.com/alphagov/govuk-frontend/pull/2417).
 
+#### Style the `html` in your cookie banner
```
EoinShaughnessy, line 68: Are we referring here to HTML-the-thing or to the `html` parameter? If it's the thing, then we could just write HTML here.

```suggestion
#### Style the HTML in your cookie banner
```

CHANGELOG.md
```diff
@@ -65,6 +65,14 @@ Replace any use of this class with the `govuk-tag--grey` class.
...
+
+We have removed the default font styles from the cookie banner Sass, to make it more obvious when classes/styles have been forgotten.
```
EoinShaughnessy, line 70:
```suggestion
We've removed the default font styles from the cookie banner Sass, to make it more obvious when classes and styles have been forgotten.
```
Forgotten by what? (Sorry if daft question - let me know if the meaning would be obvious to users!)

CHANGELOG.md
```diff
@@ -65,6 +65,14 @@ Replace any use of this class with the `govuk-tag--grey` class.
...
+
+If you are passing `html` into the cookie banner component, you must make sure you are applying the appropriate classes/styles to that HTML, such as adding the `govuk-body` class to any paragraph tags. You must do this to prevent your cookie banner displaying with unstyled text.
```
EoinShaughnessy, line 72:
```suggestion
If you're passing `html` into the cookie banner component, make sure you're applying the appropriate classes and styles to that HTML. For example, add the `govuk-body` class to any paragraph tags. You must do this to prevent your cookie banner displaying with unstyled text.
```
Wonder could we delete "to that HTML" from the second sentence? Or is there a risk some users might think it's the cookie banner component they have to apply classes and styles to?

CHANGELOG.md
vanitabarrett: Hm, kind of both 😅 It's if someone uses the `html` parameter to pass HTML code in, rather than using the `text`parameter and just passing in a string of text.

CHANGELOG.md
vanitabarrett: This is talking about classes/styling being forgotten on the HTML that's being passed into the cookie banner. Some users may assume that the HTML they pass in is somehow automatically styled. This isn't the case so we want to make it really obvious if users have forgotten to add their own styles.

CHANGELOG.md
vanitabarrett: Although actually, thinking about it, I think the same applies if a user is using the HTML version of the component rather than the macros. So maybe we need to be more generic and just talk about any custom HTML a user provides... 

I'll have a go at tweaking it and see what I come up with.

vanitabarrett: @EoinShaughnessy I've had a go at rewording this a bit to be more about "custom HTML" and less about how the user is providing that custom HTML. Would you mind taking another look?

CHANGELOG.md
```diff
@@ -128,6 +128,14 @@ Replace any use of this class with the `govuk-tag--grey` class.
 
 This change was introduced in [pull request #2417: Remove deprecated `govuk-tag--inactive class`](https://github.com/alphagov/govuk-frontend/pull/2417).
 
+#### Style any custom HTML in your cookie banner
+
+We have removed the default font styles from the cookie banner Sass, to make it more obvious when classes/styles haven't been added to any custom HTML.
```
EoinShaughnessy, line 133:
```suggestion
We've removed the default font styles from the cookie banner Sass. This change makes it more obvious when classes and styles have not been added to any custom HTML.
```
Style guide encourages us to use positive contractions, but to avoid negative contractions (some users find them hard to read). Also says to replace '/' with 'or'.

CHANGELOG.md
```diff
@@ -128,6 +128,14 @@ Replace any use of this class with the `govuk-tag--grey` class.
...
+
+If you are passing custom HTML into the cookie banner component (for example, using the `html` Nunjucks param), you must make sure you are applying the appropriate classes/styles to that HTML, such as adding the `govuk-body` class to any paragraph tags. You must do this to prevent your cookie banner displaying with unstyled text.
```
EoinShaughnessy, line 135:
```suggestion
##### If you're passing custom HTML (for example, using the `html` Nunjucks parameter) into the cookie banner component

Make sure you're applying the appropriate classes and styles to that HTML. For example, add the `govuk-body` class to any paragraph tags. You must do this to prevent your cookie banner displaying with unstyled text.
```
Just suggesting this to break up the sentences a little. Also took out the first "You must", in case it suffices to link the remaining "You must" with what happens if you users do not do this.

EoinShaughnessy: Is there any other way users could pass custom HTML without using the `html` param?

CHANGELOG.md
vanitabarrett: Thanks @EoinShaughnessy , really useful to know! I'll try to remember for future 😬

CHANGELOG.md
vanitabarrett: @EoinShaughnessy If the user was using the HTML version of the component (not the Nunjucks macros) they can edit the HTML however they want and put whatever they want in there. Hopefully these users would already be more aware of styling their HTML, but it's possible someone could have added some custom HTML which previously would have looked "correct" but following this update won't look styled any more. They'd need to take the same actions as anyone using the Nunjucks macro though.

vanitabarrett: I'm not 100% sure I see the benefit of splitting this out into its own heading? At least to me, it kind of suggests that there might be another heading later on with different actions if you do something else, which isn't the case?

CHANGELOG.md
```diff
@@ -128,6 +128,14 @@ Replace any use of this class with the `govuk-tag--grey` class.
 
 This change was introduced in [pull request #2417: Remove deprecated `govuk-tag--inactive class`](https://github.com/alphagov/govuk-frontend/pull/2417).
 
+#### Style any custom HTML in your cookie banner
+
+We've removed the default font styles from the cookie banner Sass. This change makes it more obvious when classes and styles have not been added to any custom HTML.
+
+If you're passing custom HTML into the cookie banner component (for example, using the `html` Nunjucks parameter), you must make sure you are applying the appropriate classes/styles to that HTML, such as adding the `govuk-body` class to any paragraph tags. You must do this to prevent your cookie banner displaying with unstyled text.
```
EoinShaughnessy, line 135:
```suggestion
If you're passing custom HTML into the cookie banner component (for example, using the `html` Nunjucks parameter), you must make sure you are applying the appropriate classes and styles to that HTML, such as adding the `govuk-body` class to any paragraph tags. You must do this to prevent your cookie banner displaying with unstyled text.
```

CHANGELOG.md
EoinShaughnessy: That's a good point - yeah, best not to split out then.

CHANGELOG.md
```diff
@@ -128,6 +128,14 @@ Replace any use of this class with the `govuk-tag--grey` class.
 
 This change was introduced in [pull request #2417: Remove deprecated `govuk-tag--inactive class`](https://github.com/alphagov/govuk-frontend/pull/2417).
 
+#### Style any custom HTML in your cookie banner
+
+We've removed the default font styles from the cookie banner Sass. This change makes it more obvious when classes and styles have not been added to any custom HTML.
+
+If you're passing custom HTML into the cookie banner component (for example, using the `html` Nunjucks parameter), you must make sure you are applying the appropriate classes and styles to that HTML, such as adding the `govuk-body` class to any paragraph tags. You must do this to prevent your cookie banner displaying with unstyled text.
```
EoinShaughnessy, line 135:
```suggestion
If you're passing custom HTML into the cookie banner component (for example, using the `html` Nunjucks parameter), you must make sure you're applying the appropriate classes and styles to that HTML, such as adding the `govuk-body` class to any paragraph tags. You must do this to prevent your cookie banner displaying with unstyled text.
```

EoinShaughnessy (commented):
Just suggesting a last tiny change, otherwise content is golden. :)

Issue #1856, opened 2020-07-08
Remove `display:block` hint style as we're switching from span to div


vanitabarrett:
## What
Remove `display: block` from the hint component styles - this is a breaking change.

## Why
In https://github.com/alphagov/govuk-frontend/pull/1855 we switched the hint component from having a wrapping span to having a wrapping div. This was to allow block-level HTML elements to be passed in and render as valid HTML. However, we left the `display: block` CSS on the wrapping element as [otherwise this would be a breaking change that required users to update their HTML.](https://github.com/alphagov/govuk-frontend/pull/1855#pullrequestreview-444757832)

## Further details
Related to https://github.com/alphagov/govuk-frontend/issues/2083, but the outcome of that card shouldn't impact this work unless we decide to switch back to an inline element.

## Who needs to know about this
Developers, Tech Writer

## Done when

- [ ] `display: block` has been removed from the [hint component styles](https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/components/hint/_index.scss#L5)
- [x] "Details of breaking change" section below completed to help with writing release notes

## Details of breaking change

- which users are affected: users using the hint component HTML
- the change that’s been made: (in 3.8.0) we switched the hint component to use `div` instead of `span` to allow block elements inside hints. We’re removing the styling which made the `span` behave like a block element
- changes users will have to make: update the hint HTML to use a <div> rather than a <span>
- impact of users not making those changes: their hints may not be styled correctly

Issue #2083, opened 2020-12-22
Consider updating hints and error messages to use paragraphs


36degrees:
## What

Consider updating hints and error messages to use paragraphs rather than spans and divs.

https://github.com/alphagov/govuk-frontend/blob/d5a3fa8c5e61dbc7ca0f7a8429cc8e6716ee7c32/src/govuk/components/error-message/template.njk#L3-L6

https://github.com/alphagov/govuk-frontend/blob/d5a3fa8c5e61dbc7ca0f7a8429cc8e6716ee7c32/src/govuk/components/hint/template.njk#L1-L4

## Why

As raised by a user on x-gov Slack:

> the error message is in a <span> tag. Is there a reason why a `<span>` is used, rather than a `<div>` or `<p>`, for example? The reason I ask is because I’ve been doing some testing with JAWS and I’ve found that there’s no pause between reading the hint text and the error message. So JAWS reads “For example 12 11 2007 error colon the date your passport was issued must be in the past”. Using a `<div>` didn’t make a difference, but using a `<p>` resulted in a pause between the hint text and the error message, so potentially making it a little clearer.

It looks like hints used to be paragraphs, but [were changed to spans when the hints were moved inside the label](https://github.com/alphagov/govuk_elements/commit/5f631c113840bfd4cc3ed81de8fb311638457a92#diff-1968332cde003a71f281b5d33445bcb3a43d2ac1cee565f6bc0e0a0144cfb846) – presumably as [only phrasing content is permitted within a `label`](https://html.spec.whatwg.org/multipage/forms.html#the-label-element). Error messages, when they were added later, were also spans inside the label.

We recently [updated hints to use divs rather than spans](https://github.com/alphagov/govuk-frontend/commit/d92d3c8a990730213430b686d239fc7b278530a9) so that they can be used with flow content (for example multiple paragraph tags, or a list)

Since [we moved hints and error messages out of the labels](https://github.com/alphagov/govuk-frontend/pull/681), we should consider whether they should now by marked up as paragraphs instead.

## Who needs to know about this

Developers, designers

## Done when

- [x] Tested alternate form controls with error messages and hints marked up as paragraphs in assistive technologies
- [x] Identified other potential issues cause by changing hints and error messages to paragraphs
- [ ] Decision made and code change made
- [ ] Check if the changelog for https://github.com/alphagov/govuk-frontend/issues/1856 needs updating as a result of this work

36degrees:
[...]

PR #2437, opened 2021-11-19
Remove `display:block` on hint component

vanitabarrett:
Closes https://github.com/alphagov/govuk-frontend/issues/1856

## What
Remove the `display: block` on the hint component

## Why
In v3.8.0 we added the ability for block level elements to be rendered inside hints by switching from using a `span` to a `div`.

However, we kept the `display: block` styling (which would now be redundant as this is the default for divs). This meant we weren't forcing users to update their HTML and we could avoid a breaking change.

We now want to tidy up that redundant CSS as part of v4.0.0.

**Note:** this work is related to https://github.com/alphagov/govuk-frontend/issues/2083. I think we can probably safely make this change first, without making a decision on https://github.com/alphagov/govuk-frontend/issues/2083. I've added a checklist item to that card to make sure we check the Changelog for this change still makes sense alongside whatever we decide in #2083.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -102,6 +102,16 @@ If you're importing JavaScript modules individually, you should check any refere
 
 This change was introduced in [pull request #1836: Rename exported JavaScript modules to include component name](https://github.com/alphagov/govuk-frontend/issues/1836)].
 
+#### Update the HTML for hints
+
+In GOV.UK Frontend v3.8.0, we [updated the HTML for hints to use a `<div>` rather than a `<span>` element, to allow the use of block elements](https://github.com/alphagov/govuk-frontend/issues/1835). We've now removed the styles which made the hint `<span>` display as block.
+
+To make sure your hints are styled correctly, update your hint HTML to use a `<div>` rather than a `<span>`. For example: `<div id="input-hint" class="govuk-hint"></div>`.
+
+You do not need to do anything if you're using Nunjucks macros.
+
+This change was introduced in [pull request #2437: Remove `display:block` on hint component](https://github.com/alphagov/govuk-frontend/pull/2437).
+
 #### Remove calls to deprecated `iff` Sass function
 
 We've removed the `iff` function which we deprecated in [GOV.UK Frontend version 3.6.0](https://github.com/alphagov/govuk-frontend/releases/tag/v3.6.0).
--- a/src/govuk/components/hint/_index.scss
+++ b/src/govuk/components/hint/_index.scss
@@ -2,8 +2,6 @@
   .govuk-hint {
     @include govuk-font($size: 19);
 
-    display: block;
-
     margin-bottom: govuk-spacing(3);
 
     color: $govuk-secondary-text-colour;


CHANGELOG.md
```diff
@@ -102,6 +102,17 @@ If you're importing JavaScript modules individually, you should check any refere
 
 This change was introduced in [pull request #1836: Rename exported JavaScript modules to include component name](https://github.com/alphagov/govuk-frontend/issues/1836)].
 
+#### Update the HTML for hints
+
+In v3.8.0, we updated the HTML for hints to use a <div> rather than a <span> element, to allow block elements to be used.
```
EoinShaughnessy, line 107:
```suggestion
In [GOV.UK Frontend v3.8.0](https://github.com/alphagov/govuk-frontend/releases/tag/v3.8.0), we updated the HTML for hints to use a `<div>` rather than a `<span>` element, to allow the use of block elements.
```
In the v3.8.0 release notes, the link for this change is in the 'Fixes' section, which might make it easy to miss. Wonder is there a case for linking to [#1855](https://github.com/alphagov/govuk-frontend/pull/1855) from the text "updated the HTML for hints...use of block elements"?

CHANGELOG.md
```diff
@@ -102,6 +102,17 @@ If you're importing JavaScript modules individually, you should check any refere
...
+We have now removed the styles which made the hint <span> display as block.
```
EoinShaughnessy, line 108:
```suggestion
We've now removed the styles which made the hint `<span>` display as block.
```

CHANGELOG.md
```diff
@@ -102,6 +102,17 @@ If you're importing JavaScript modules individually, you should check any refere
...
+
+Update your hint HTML to use a <div> rather than a <span>, for example: `<div id="input-hint" class="govuk-hint"></div>`, to ensure your hints are styled correctly.
```
EoinShaughnessy, line 110:
```suggestion
To make sure your hints are styled correctly, update your hint HTML to use a `<div>` rather than a `<span>`. For example: `<div id="input-hint" class="govuk-hint"></div>`.
```

EoinShaughnessy (commented):
Content looks great! Made a few suggestions.

lfdebrux (approved):

CHANGELOG.md
```diff
@@ -102,6 +102,16 @@ If you're importing JavaScript modules individually, you should check any refere
 
 This change was introduced in [pull request #1836: Rename exported JavaScript modules to include component name](https://github.com/alphagov/govuk-frontend/issues/1836)].
 
+#### Update the HTML for hints
+
+In GOV.UK Frontend v3.8.0, we [updated the HTML for hints to use a `<div>` rather than a `<span>` element, to allow the use of block elements](https://github.com/alphagov/govuk-frontend/issues/1835). We've now removed the styles which made the hint <span> display as block.
```
EoinShaughnessy, line 107:
```suggestion
In GOV.UK Frontend v3.8.0, we [updated the HTML for hints to use a `<div>` rather than a `<span>` element, to allow the use of block elements](https://github.com/alphagov/govuk-frontend/issues/1835). We've now removed the styles which made the hint `<span>` display as block.
```

EoinShaughnessy (commented):
Sorry! I'd overlooked one code mention (line 107) that needs backticks. Content looks perfect otherwise.

PR #2438, opened 2021-11-22
Move focus to content area on skip link activation

vanitabarrett:


hannalaakso:
Used to generate a preview for the original investigation - closed in favour of https://github.com/alphagov/govuk-frontend/pull/2450

PR #2441, opened 2021-11-24
Update legend/label/heading examples to have unique names and ids

vanitabarrett:
I came across this when using these examples to test https://github.com/alphagov/govuk-frontend/issues/2083

The current examples for legends, labels and headings have the same `name` and `id` (or `idPrefix`). This meant for things using checkboxes and radios (for example), the checkboxes were appearing as one list of checkboxes for screenreader users, e.g: tickbox 2 of 60.

It also meant that hints were being read incorrectly for some questions, because all hints on the page had the same `id` (e.g: `example-hint`).

This commit gives each question a unique name, and either removes the `idPrefix` so it falls back to the `name` or gives each question a unique `id` too. 

Each individual commit contains more details on the issues this was causing when testing with assistive technologies and the change that's been made to fix it.

https://govuk-frontend-pr-2441.herokuapp.com/examples/labels-legends-and-headings

### DIFF
--- a/app/views/examples/labels-legends-and-headings/index.njk
+++ b/app/views/examples/labels-legends-and-headings/index.njk
@@ -266,7 +266,7 @@
     "hint": {
       "html": "It’s on your National Insurance card, benefit letter, payslip or P60. For example, ‘QQ 12 34 56 C’."
     },
-    "id": "input-with-error-message",
+    "id": "text-input-heading-hint-error-xl",
     "name": "text-input-heading-hint-error-xl",
     "errorMessage": {
       "text": "Error message goes here"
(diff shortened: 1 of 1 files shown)


lfdebrux:
For some reason the radio buttons are now pre-selected on `no` 🤔

vanitabarrett:
@lfdebrux Really good spot! I've addressed this in https://github.com/alphagov/govuk-frontend/pull/2441/commits/714bece0192f3bd379fb62deea16ddb30237fe4f

app/views/examples/labels-legends-and-headings/index.njk
```diff
@@ -2838,12 +2694,11 @@
     })
   }}
 
-
   <h3 class="govuk-heading-s">isPageHeading: false</h3>
 
   {{ govukDateInput({
     id: 'dob',
```
lfdebrux, line 2700: I think this `id` should also be updated? There are a couple of date input components with the id `dob` still.

lfdebrux (approved):
W3C validator reports no errors or warnings 🙌

PR #2445, opened 2021-11-29
Footer alignment example should use full width layout

vanitabarrett:
This commit updates the footer alignment examples to use the full width layout, as the component would normally be full width on the page. This mimics the behaviour of the `footer` and `content` blocks, where the content is normally wrapped in `govuk-width-container` but the footer is not.

The previous implementation nested everything inside a parent with `govuk-width-container` class which pushed the alignment slightly out of place on tablet/mobile.

## Before
Screenshot 2021-11-29 at 09 46 45

## After
Screenshot 2021-11-29 at 09 46 16


lfdebrux (approved):
Ah, smart, I love it!!

Issue #2187, opened 2021-04-13
After clicking "Skip to main content", focus isn't sent to the content so screenreader can't read from there


jonnywyatt:
## Description of the issue
gov.uk "Skip to main content" links to the https://www.gov.uk/#content anchor but doesn't send focus to that element, so if I then try to read from that point using Voiceover with VO + right arrow, I'm still in the top banner. It feels to me like a screen reader should be able to start reading again from the anchor point. You can press tab to go to the next focusable element, but there might not be any in the content area, you might just want to read from that point.

I noticed some discussion about adding then reverting tabindex on <main />  - https://github.com/alphagov/govuk-design-system-backlog/issues/66#issuecomment-411673528

Do you consider that the two problems with tabindex identified by @fofr are bigger issues than not being able to read from the anchor point, if tabindex isn't used?

## Steps to reproduce the issue
1. on www.gov.uk, tab to "Skip to main content" and press it
2. Use VO + right arrow to attempt to read the main content

## Actual vs expected behaviour
Expected - Voiceover reads the first element in the content panel
Actual - Voiceover continues reading from the banner

## Environment (where applicable)

- Operating system: Mac OS Catalina 10.15.6
- Browser: Safari
- Browser version: 14.0 (15610.1.28.1.9, 15610)
- GOV.UK Frontend Version: live site

## Done when

- [ ] TBC
- [ ] "Details of breaking change" section below completed to help with writing release notes

## Details of breaking change

- which users are affected: TBC
- the change that’s been made: TBC
- changes users will have to make: TBC
- impact of users not making those changes: TBC

jonnywyatt:
https://user-images.githubusercontent.com/471250/114539888-d894a580-9c4c-11eb-8d92-f1b95c24f98d.mov

36degrees:
I *think* this might involve a breaking change, so attaching this to the 4.0 milestone for consideration.

vanitabarrett:
Relevant article: https://axesslab.com/skip-links/ which also links to [Anika's solution](https://github.com/selfthinker/dokuwiki_template_writr/blob/master/js/skip-link-focus-fix.js)

vanitabarrett:
As discussed during v4 prioritisation, we should investigate the solution to this and see if 1) it involves a breaking change and what changes users will need to make; 2) if it tests well in different browsers.

hannalaakso:
I can replicate the Mac VoiceOver issue raised by @jonnywyatt 👍 

[...]

PR #2450, opened 2021-11-29
Set focus to skip link target to improve screen reader announcements

hannalaakso:
When user activates the skip link, Mac VoiceOver doesn't announce the main content or continue reading
from it as reported in https://github.com/alphagov/govuk-frontend/issues/2187.

To improve the experience for Mac VoiceOver users:
- make the main content element focusable by adding a `tabindex` attribute
- move focus to it programmatically
- override the native focus outline to none whilst `tabindex` is present
- remove the `tabindex` attribute and the style override on blur

This follows the pattern we already use in the error summary and notification banner components, and follows [the approach recommended by @selfthinker.](https://github.com/alphagov/govuk-design-system-backlog/issues/66#issuecomment-415758358)

### Results from screen reader testing
This solves the issue in Mac VoiceOver: when the user activates the skip link, VoiceOver now announces the main element. If the user then tries to read the next element, VoiceOver now correctly continues reading from the main content. There is also a smaller improvement to the announcements on JAWS (both with Chrome and IE11): JAWS now announces it's on the main region, before it starts reading the main content.

Behaviour | Announces skip link when it's navigated to | Announces main has been navigated to | Reads out content of main
-- | -- | -- | --
JAWS 2020 / Chrome | same page link skip to main content | enter **main** region **main** region page has 4 regions and 8 headings `<main content>` | enter **main** region **main** region page has 4 regions and 8 headings `<main content>`
JAWS 2020 / IE 11 | skip to main content same page link | enter **main** region page has 4 regions and 8 headings `<main content>` | enter main region page has 4 regions and 8 headings `<main content>`
NVDA 2021.1/ Firefox 90 | skip to main content link | main landmark  `<main content>` | main landmark  `<main content>`
Voiceover / Safari (macOS Mojave) | skip to main content, link | **main you are currently on a text area** | **`<main content>`**
Voiceover / Safari (iOS 15) | Skip to main content, in-page link | After double-tapping to follow link: "Press release, main landmark" | After double-tapping to follow link: "Press release, main landmark"
Talkback / Chrome 96 (Android 11) | skip to main content, link | After double-tapping to follow link: "main, Press release, <main content>" | After double-tapping to follow link: "main, Press release, <main content>"

(Announcements that have changed from the live version are in **bold** ^)

See https://github.com/alphagov/govuk-frontend/issues/2187#issuecomment-973172184 for more details.

Closes https://github.com/alphagov/govuk-frontend/issues/2187

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -135,6 +135,24 @@ You do not need to do anything if you're using Nunjucks macros.
 
 This change was introduced in [pull request #2437: Remove `display:block` on hint component](https://github.com/alphagov/govuk-frontend/pull/2437).
 
+#### Include JavaScript for skip link to improve screen reader announcements
+
+We've added JavaScript for the skip link component to set focus to the linked element, for example, the main content on the page. This helps screen readers read the linked content when users use the skip link.
+
+If you're not using Nunjucks macros, add a `data-module="govuk-skip-link"` attribute to the component HTML. For example:
+
+```html
+<div class="govuk-skip-link" data-module="govuk-skip-link">
+...
+</div>
+```
+
+If you're [importing JavaScript for individual components](https://frontend.design-system.service.gov.uk/importing-css-assets-and-javascript/#select-and-initialise-an-individual-component), import the skip link JavaScript.
+
+Once you've made the changes, check that the skip link JavaScript works. To make sure, click the skip link and check that the linked element (usually the `<main>` element) in the browser has a `tabindex` attribute.
+
+This change was introduced in [pull request #2450: Set focus to skip link target to improve screen reader announcements](https://github.com/alphagov/govuk-frontend/pull/2450).
+
 #### Remove calls to deprecated `iff` Sass function
 
 We've removed the `iff` function which we deprecated in [GOV.UK Frontend version 3.6.0](https://github.com/alphagov/govuk-frontend/releases/tag/v3.6.0).
--- a/src/govuk/components/skip-link/skip-link.js
+++ b/src/govuk/components/skip-link/skip-link.js
@@ -1,3 +1,7 @@
+import '../../vendor/polyfills/Function/prototype/bind'
+import '../../vendor/polyfills/Element/prototype/classList'
+import '../../vendor/polyfills/Event' // addEventListener and event.target normalization
+
 function SkipLink ($module) {
   this.$module = $module
 }
@@ -11,6 +15,79 @@ SkipLink.prototype.init = function () {
   if (!$module) {
     return
   }
+
+  $module.addEventListener('click', this.handleClick.bind(this))
+}
+
+/**
+* Click event handler
+*
+* @param {MouseEvent} event - Click event
+*/
+SkipLink.prototype.handleClick = function (event) {
+  var target = event.target
+
+  if (this.focusLinkedElement(target)) {
+    event.preventDefault()
+  }
+}
+
+/**
+ * Focus the linked element
+ *
+ * @param {HTMLElement} $target - Event target
+ * @returns {boolean} True if the linked element was able to be focussed
+ */
+SkipLink.prototype.focusLinkedElement = function ($target) {
+  // If the element that was clicked does not have a href, return early
+  if ($target.href === false) {
+    return false
+  }
+
+  var linkedElementId = this.getFragmentFromUrl($target.href)
+  var $linkedElement = document.getElementById(linkedElementId)
+  if (!$linkedElement) {
+    return false
+  }
+
+  if (!$linkedElement.getAttribute('tabindex')) {
+    // Set the content tabindex to -1 so it can be focused with JavaScript.
+    $linkedElement.setAttribute('tabindex', '-1')
+    $linkedElement.classList.add('govuk-skip-link-focused-element')
+
+    $linkedElement.addEventListener('blur', this.removeFocusProperties.bind(this, $linkedElement))
+  }
+  $linkedElement.focus()
+}
+
+/**
+ * Remove the tabindex that makes the linked element focusable because the content only needs to be
+ * focusable until it has received programmatic focus and a screen reader has announced it.
+ *
+ * Remove the CSS class that removes the native focus styles.
+ *
+ * @param {HTMLElement} $linkedElement - DOM element linked to from the skip link
+ */
+SkipLink.prototype.removeFocusProperties = function ($linkedElement) {
+  $linkedElement.removeAttribute('tabindex')
+  $linkedElement.classList.remove('govuk-skip-link-focused-element')
+}
+
+/**
+ * Get fragment from URL
+ *
+ * Extract the fragment (everything after the hash) from a URL, but not including
+ * the hash.
+ *
+ * @param {string} url - URL
+ * @returns {string} Fragment from URL, without the hash
+ */
+SkipLink.prototype.getFragmentFromUrl = function (url) {
+  if (url.indexOf('#') === -1) {
+    return false
+  }
+
+  return url.split('#').pop()
 }
 
 export default SkipLink
--- a/src/govuk/components/skip-link/skip-link.js
+++ b/src/govuk/components/skip-link/skip-link.js
@@ -1,0 +1,16 @@
+function SkipLink ($module) {
+  this.$module = $module
+}
+
+/**
+ * Initialise the component
+ */
+SkipLink.prototype.init = function () {
+  var $module = this.$module
+  // Check for module
+  if (!$module) {
+    return
+  }
+}
+
+export default SkipLink
--- a/src/govuk/components/skip-link/skip-link.test.js
+++ b/src/govuk/components/skip-link/skip-link.test.js
@@ -1,0 +1,50 @@
+/* eslint-env jest */
+
+const configPaths = require('../../../../config/paths.json')
+const PORT = configPaths.ports.test
+
+const baseUrl = 'http://localhost:' + PORT
+
+describe('/examples/template-default', () => {
+  describe('skip link', () => {
+    beforeAll(async () => {
+      await page.goto(`${baseUrl}/examples/template-default`, { waitUntil: 'load' })
+      await page.keyboard.press('Tab')
+      await page.keyboard.press('Enter')
+    })
+
+    it('focuses the linked element', async () => {
+      const activeElement = await page.evaluate(() => document.activeElement.id)
+
+      expect(activeElement).toBe('main-content')
+    })
+
+    it('adds the tabindex attribute to the linked element', async () => {
+      const tabindex = await page.$eval('.govuk-main-wrapper', el => el.getAttribute('tabindex'))
+
+      expect(tabindex).toBe('-1')
+    })
+
+    it('adds the class for removing the native focus style to the linked element', async () => {
+      const cssClass = await page.$eval('.govuk-main-wrapper', el => el.classList.contains('govuk-skip-link-focused-element'))
+
+      expect(cssClass).toBeTruthy()
+    })
+
+    it('removes the tabindex attribute from the linked element on blur', async () => {
+      await page.$eval('.govuk-main-wrapper', el => el.blur())
+
+      const tabindex = await page.$eval('.govuk-main-wrapper', el => el.getAttribute('tabindex'))
+
+      expect(tabindex).toBeNull()
+    })
+
+    it('removes the class for removing the native focus style from the linked element on blur', async () => {
+      await page.$eval('.govuk-main-wrapper', el => el.blur())
+
+      const cssClass = await page.$eval('.govuk-main-wrapper', el => el.getAttribute('class'))
+
+      expect(cssClass).not.toContain('govuk-skip-link-focused-element')
+    })
+  })
+})
--- a/app/views/examples/template-custom/index.njk
+++ b/app/views/examples/template-custom/index.njk
@@ -130,5 +130,6 @@
 {% block bodyEnd %}
   <!-- block:bodyEnd -->
   <script src="/public/all.js"></script>
+  <script>window.GOVUKFrontend.initAll()</script>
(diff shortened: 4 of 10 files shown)


src/govuk/components/skip-link/skip-link.test.js
```diff
@@ -0,0 +1,50 @@
...
+    it('adds the tabindex attribute to the target', async () => {
+      const tabindex = await page.$eval('.govuk-main-wrapper', el => el.getAttribute('tabindex'))
+
+      expect(tabindex).toBe('-1')
+    })
+
+    it('adds the class for removing the native focus style to the target', async () => {
+      const cssClass = await page.$eval('.govuk-main-wrapper', el => el.getAttribute('class'))
+
+      expect(cssClass).toContain('govuk-skip-link-focused-element')
```
36degrees, line 31: I'd suggest using `classlist.contains` here, because it's a slightly more robust check.

For example, this would currently pass if there was a class that contained this string  `class="govuk-skip-link-focused-element-foo"` or if there was a missing space between two class names (`class="another-classgovuk-skip-link-focused-element"`)

```suggestion
      const hasClass = await page.$eval('.govuk-main-wrapper',
        el => el.classList.contains('govuk-skip-link-focused-element')
      )

      expect(hasClass).toBeTruthy()
```

src/govuk/components/skip-link/skip-link.test.js
```diff
@@ -0,0 +1,50 @@
...
+const PORT = configPaths.ports.test
+
+const baseUrl = 'http://localhost:' + PORT
+
+describe('/examples/template-default', () => {
+  describe('skip link', () => {
+    beforeAll(async () => {
+      await page.goto(`${baseUrl}/examples/template-default`, { waitUntil: 'load' })
+      await page.keyboard.press('Tab')
+      await page.click('.govuk-skip-link')
```
36degrees, line 13: Would it make more sense to use `page.keyboard.press('Enter')`, as that's a better simulation of what someone navigating using the keyboard would do?

src/govuk/components/skip-link/skip-link.js
```diff
@@ -0,0 +1,81 @@
...
+    return false
+  }
+
+  var contentId = this.getFragmentFromUrl($target.href)
+  var $content = document.getElementById(contentId)
+  if (!$content) {
+    return false
+  }
+
+  if (!$content.getAttribute('tabindex')) {
```
36degrees, line 49: To make this easier to follow, what about making the variable names consistent with the method name and refer to the thing we want to focus as the `target`, rather than `content`?

```suggestion
  var targetId = this.getFragmentFromUrl($target.href)
  var $target = document.getElementById(contentId)
  if (!$target) {
    return false
  }

  if (!$target.getAttribute('tabindex')) {
```

etc…

src/govuk/components/skip-link/skip-link.js
```diff
@@ -0,0 +1,81 @@
+function SkipLink ($module) {
```
36degrees, line 1: I think we're missing some polyfill imports in this file… at least `Element.classList`, but possibly others?

(It'll still work in older browsers because of #674, unfortunately… not sure what the best way is to test this)

src/govuk/components/skip-link/skip-link.test.js
```diff
@@ -0,0 +1,50 @@
+/* eslint-env jest */
+
+const configPaths = require('../../../../config/paths.json')
+const PORT = configPaths.ports.test
+
+const baseUrl = 'http://localhost:' + PORT
+
+describe('/examples/template-default', () => {
```
36degrees, line 8: Nice tests 👍🏻

CHANGELOG.md
```diff
@@ -75,6 +75,24 @@ If you see unexpected behaviour, make sure the `id` for the  textarea is unique
...
+<div class="govuk-skip-link" data-module="govuk-skip-link">
+...
+</div>
+```
+
+If you're [importing JavaScript for individual components](https://frontend.design-system.service.gov.uk/importing-css-assets-and-javascript/#select-and-initialise-an-individual-component), import the skip link JavaScript.
+
+Once you've made the changes, you can check that the skip link works correctly by clicking it and   inspecting the target element (commonly the `<main>` element) in the browser to ensure it has a `tabindex` attribute.
+
+This change was introduced in [pull request #2450: ](https://github.com/alphagov/govuk-frontend/pull/2450).
```
36degrees, line 94: Missing PR title?

src/govuk/components/skip-link/skip-link.test.js
hannalaakso: @36degrees Could you elaborate on this, I don't think just pressing 'Enter' will navigate to the skip link?

36degrees: It should do, once the link is focussed by pressing tab? I'd expect in most cases someone who is relying on the skip link and navigating the site using their keyboard isn't going to tab to the skip link only then to use their mouse to click it.

CHANGELOG.md
```diff
@@ -75,6 +75,24 @@ If you see unexpected behaviour, make sure the `id` for the  textarea is unique
 
 This change was introduced in [pull request #2408: Prevent issues with character count when textarea `id` includes CSS syntax characters](https://github.com/alphagov/govuk-frontend/pull/2408).
 
+#### Include JavaScript for the skip link to improve screen reader announcements
```
EoinShaughnessy, line 78:
```suggestion
#### Include JavaScript for skip link to improve screen reader announcements
```

CHANGELOG.md
```diff
@@ -75,6 +75,24 @@ If you see unexpected behaviour, make sure the `id` for the  textarea is unique
...
+
+We've added JavaScript for the skip link component to set focus to the target element. This improves the announcements especially for Mac VoiceOver users and also for JAWS users.
```
EoinShaughnessy, line 80: Should we give an example of what we mean by 'target element' (if there could be any inexperienced users who wouldn't understand this right away)? "The main content on a page"?

How does it improve the announcements? By reading the main element when users activate the skip link - which is different to what happened before we made this change?

Last comment: we had some feedback from release notes testing that some of our users are junior devs who are inexperienced when it comes to accessibility. Do we need to spell out the JAWS acronym for them, or provide any links?

```suggestion
We've added JavaScript for the skip link component to set focus to the target element, for example, the main content on a page. This change improves the announcements, especially for users of Mac VoiceOver and JAWS.
```

CHANGELOG.md
```diff
@@ -75,6 +75,24 @@ If you see unexpected behaviour, make sure the `id` for the  textarea is unique
...
+
+If you are not using Nunjucks macros, add a `data-module="govuk-skip-link"` attribute to the component HTML. For example:
```
EoinShaughnessy, line 82:
```suggestion
If you're not using Nunjucks macros, add a `data-module="govuk-skip-link"` attribute to the component HTML. For example:
```

CHANGELOG.md
EoinShaughnessy:
```suggestion
This change was introduced in [pull request #2450: Set focus to skip link target to improve screen reader announcements](https://github.com/alphagov/govuk-frontend/pull/2450).
```

src/govuk/components/skip-link/skip-link.js
hannalaakso: I tried to find a better name for the variable but we're constrained by the existing naming conventions here. `$target` is already defined, it contains the skip link. This function duplicates logic in the error summary (ideally we should be extracting this to be a reusable function but as just discussed in a different context doing that will be part a bigger piece of work) so I'm loath to rename the function or its parameter in order to keep that linkage. 

In the error summary we call the elements to focus by the HMTL element name (`$input` and `$legendOrLabel`) but `$main` in this context felt too specific, so I settled on `$content`.

CHANGELOG.md
```diff
@@ -75,6 +75,24 @@ If you see unexpected behaviour, make sure the `id` for the  textarea is unique
...
+
+```html
+<div class="govuk-skip-link" data-module="govuk-skip-link">
+...
+</div>
+```
+
+If you're [importing JavaScript for individual components](https://frontend.design-system.service.gov.uk/importing-css-assets-and-javascript/#select-and-initialise-an-individual-component), import the skip link JavaScript.
+
+Once you've made the changes, you can check that the skip link works correctly by clicking it and   inspecting the target element (commonly the `<main>` element) in the browser to ensure it has a `tabindex` attribute.
```
EoinShaughnessy, line 92:
```suggestion
Once you've made the changes, check that the skip link works. To make sure, click the skip link and check that the target element (usually the `<main>` element) in the browser has a `tabindex` attribute.
```

src/govuk/components/skip-link/skip-link.js
36degrees: That's a good point – but I think points to a bigger naming issue (which is absolutely already present in the error summary) and which is potentially even more confusing… `target` means two different things in two different parts of the code. Within the context of `handleClick` target refers to the 'event target' (the thing that was clicked), but `focusTarget` at least suggests that the target is the element that we want to focus.

Base on the function's signature (`focusTarget($target)`) you would definitely expect it to focus whatever `$target` is. And that's also what the doc block says it does! But it **doesn't** do that – it focuses whatever `$target` links to. The target of the `$target`, I guess…?

I think this is worth trying to fix rather than introducing the issue to another component.

src/govuk/components/skip-link/skip-link.js
```diff
@@ -0,0 +1,81 @@
...
+    // Set the content tabindex to -1 so it can be focused with JavaScript.
+    $content.setAttribute('tabindex', '-1')
+    $content.classList.add('govuk-skip-link-focused-element')
+
+    // Remove the tabindex attribute on blur because the content only needs to be focusable until it
+    // has received programmatic focus and a screen reader has announced it.
+    $content.addEventListener('blur', function () {
+      $content.removeAttribute('tabindex')
+      $content.classList.remove('govuk-skip-link-focused-element')
+    })
```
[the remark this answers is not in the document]
hannalaakso: I don't think this is true, it's valid to use an anonymous function with an event listener: https://developer.mozilla.org/en-US/docs/Web/API/EventListener#javascript I don't think a named function would change the actual behaviour in any way. But I'm going to leave this open, we can compare notes on Monday 👀

CHANGELOG.md
hannalaakso: Thanks @EoinShaughnessy, this is a nice improvement 👍  

I've gone with your wording but I've added the word 'JavaScript'. This is because strictly speaking we want to them to check that the skip link _JavaScript_ works. You might get the impression that the skip link has worked since the browser has shifted the page content but the JavaScript might have not fired - checking for the `tabindex` ensures that it did.

CHANGELOG.md
hannalaakso: Thanks for the feedback Eoin, all really good suggestions. I've gone with something that's quite close to what you suggested including removing some of the jargon. Have a look and let me know if you think it needs further tweaking.

src/govuk/components/skip-link/skip-link.js
hannalaakso: Good catch, this is a bit of a blind spot 🤕  I've added the imports and tested this works by removing all the other components from `all.js` in checking that the skip link works in older IE and there are no errors in the console.

src/govuk/components/skip-link/skip-link.test.js
hannalaakso: Oh I see what you mean. Yes, that seems like a better simulation. I've switched it up.

src/govuk/components/skip-link/skip-link.js
hannalaakso: You're right, it could be quite confusing. I've just tried to improve the wording just for this component, have a look.

CHANGELOG.md
```diff
@@ -75,6 +75,24 @@ If you see unexpected behaviour, make sure the `id` for the  textarea is unique
 
 This change was introduced in [pull request #2408: Prevent issues with character count when textarea `id` includes CSS syntax characters](https://github.com/alphagov/govuk-frontend/pull/2408).
 
+#### Include JavaScript for skip link to improve screen reader announcements
+
+We've added JavaScript for the skip link component to set focus to the linked element, for example, the main content on the page. This helps screen readers read the linked content when users activate the skip link.
```
EoinShaughnessy, line 80:
```suggestion
We've added JavaScript for the skip link component to set focus to the linked element, for example, the main content on the page. This helps screen readers read the linked content when users turn on the skip link.
```

EoinShaughnessy (commented):
@hannalaakso Changelog update is great! Just suggesting one word-change, but otherwise this looks good to go. 👍🏻

src/govuk/components/skip-link/skip-link.js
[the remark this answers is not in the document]
hannalaakso: My bad, I've had another look and this was actually a good catch 💯  I'm now binding the function on the event listener which stops the event listener being created more than once.

vanitabarrett (approved):
This PR looks really good, thanks @hannalaakso !

As discussed in dev catchup, this PR is good apart from one issue Ollie noticed where the event listener is added each time a user interacts with the skip link. It feels unlikely that this would happen or cause many problems, but it's something we should fix.

However, we want to include this change in the pre-release so I'm going to approve it for now so it can be merged. We'll revisit this to fix the multiple event listeners issue before the actual v4.0.0 release.

src/govuk/components/skip-link/skip-link.js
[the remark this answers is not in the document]
hannalaakso: This still isn't working as expected - I want to refactor the code so that we set the event listener at the point of initialisation but going to pick this up later so that we can get the actual change to the component behaviour tested as part of the pre-release.

PR #2452, opened 2021-12-01
Update error messages to use paragraph tags instead of spans

vanitabarrett:
Part of https://github.com/alphagov/govuk-frontend/issues/2083

Updates the error messages parent element to use paragraph tags instead of spans, as it's more semantically correct.

Read https://github.com/alphagov/govuk-frontend/issues/2083#issuecomment-979874842 for full details of accessibility testing. There should be no visual difference following this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -102,6 +102,14 @@ If you're importing JavaScript modules individually, you should check any refere
 
 This change was introduced in [pull request #2426: Rename exported JavaScript modules to include component name](https://github.com/alphagov/govuk-frontend/pull/2426).
 
+#### Update the HTML for error messages
+
+We’ve updated the HTML for the error message component to use a paragraph element instead of a span element, as this is more semantically correct.
+
+If you’re not using Nunjucks macros, swap the `<span class="govuk-error-message">` for a `<p class="govuk-error-message">`.
+
+This change was introduced in [pull request #2452: Update error messages to use paragraph tags instead of spans](https://github.com/alphagov/govuk-frontend/pull/2452).
+
 #### Update the HTML for hints
 
 In GOV.UK Frontend v3.8.0, we [updated the HTML for hints to use a `<div>` rather than a `<span>` element, to allow the use of block elements](https://github.com/alphagov/govuk-frontend/issues/1835). We've now removed the styles which made the hint `<span>` display as block.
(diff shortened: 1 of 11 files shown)


CHANGELOG.md
```diff
@@ -102,6 +102,14 @@ If you're importing JavaScript modules individually, you should check any refere
 
 This change was introduced in [pull request #2426: Rename exported JavaScript modules to include component name](https://github.com/alphagov/govuk-frontend/pull/2426).
 
+#### Update the HTML for error messages
+
+We’ve updated the HTML for the error message component to use a paragraph element instead of a span element.
+
+If you’re not using Nunjucks macros, swap the `<span class="govuk-error-message">` for a `<p class="govuk-error-message">`.
+
+This change was introduced in [pull request #2452: Update error messages from span to paragraph](https://github.com/alphagov/govuk-frontend/pull/2452).
```
EoinShaughnessy, line 111: Wonder should we change the PR title to 'Update error messages to use paragraph tags instead of spans'?

CHANGELOG.md
```diff
@@ -102,6 +102,14 @@ If you're importing JavaScript modules individually, you should check any refere
 
 This change was introduced in [pull request #2426: Rename exported JavaScript modules to include component name](https://github.com/alphagov/govuk-frontend/pull/2426).
 
+#### Update the HTML for error messages
+
+We’ve updated the HTML for the error message component to use a paragraph element instead of a span element.
```
EoinShaughnessy, line 107: Do we need to say why we've done this? The PR says because it's more semantically correct. Will users know what we mean by that, or do we need to specify how this change advantages them?

lfdebrux (approved):

36degrees:
> I think I can see a visual change

Yep, good catch. I think there's an additional top margin coming from the user agent default styles for paragraphs?

lfdebrux:
I've done a visual inspection in:

- macOS / Safari 
- macOS / Chrome 
- macOS / Firefox with forced colours
- Android / Samsung Internet (BrowserStack)
- Windows / IE8 (BrowserStack)
- Windows / IE11 with High Contrast Mode

lfdebrux (approved):
Looks good to me

vanitabarrett:
As a note, @lfdebrux has also spotted that this slightly changes how this component looks when styles are disabled. As it's now a paragraph, the error message renders on its own line, but we think this is an improvement.

## Before
screenshot_2021-12-03_at_13 10 29

## After
screenshot_2021-12-03_at_13 10 52

PR #2453, opened 2021-12-01
Input with suggestions

frankieroberto:
This draft builds upon the Text input feature, by allowing an option where "suggestions" are shown to the user as they type into the box. This can save the user time by allowing them to select a suggestion rather than having to type out a full answer.  It can also help to avoid typos or other minor inconsistencies.

It should only be used where the input is not constrained, and the user is able to type in their own answer. If the user must pick an option from a pre-defined list, the alternative [Select with search](https://github.com/alphagov/govuk-frontend/pull/2430) could be used.

## Design considerations

The design is partly based on the existing accessible-autocomplete component, however it aims to make it clearer that users do not have to select one of the options by:

* having some space between the Input box and the suggestions
* adding a "Suggestions" header above the list
* not showing anything if there are no matching options, so that it looks like a regular text input 

Screenshot 2021-12-06 at 17 40 38

The suggestions search is case insensitive, and matches the start of each word. A minimum of 2 characters must be entered before the suggestions appear (this could be configurable?).

Synonyms are supported, which can be used for alternative names, acronyms, previous names or common mis-spellings.

➡️ [Working demo](https://govuk-frontend-pr-2453.herokuapp.com/components/input/with-suggestions/preview)
➡️ [Draft guidance](https://deploy-preview-1993--govuk-design-system-preview.netlify.app/components/text-input/#if-there-is-a-list-of-common-answers) ([Pull request](https://github.com/alphagov/govuk-design-system/pull/1991))

# TODO

* [ ] full accessibility testing
* [ ] add javascript tests
* [ ] consider whether to use `datalist` as a no-javascript fallback, or whether the browser implementations are too usable for it to be worth it
* [ ] consider whether suggestions should be sorted in "relevance" order (based upon matching input) rather than always in source order


simonwhatley:
In lieu of the accessible autocomplete being included in the design system, I think this has the potential to serve a similar purpose elegantly.

I have a few thoughts.

**Proximity of suggestions**
The spacing between the input and the suggestions is sufficiently large enough that some users may not perceive the suggestions as related. 

I recommend reducing the space from `30px` to `10px`.

**Clickable suggestions**
On macOS, it‘s not clear whether the suggestions can be chosen, or whether the user needs to continue typing the name of the suggestion.

**Scrollable suggestions**
On macOS, it‘s not clear that the suggestions list can be scrolled.

I recommend adding CSS rules to resolve click and scroll, for example:

```
::-webkit-scrollbar {
  width: 7px;
  -webkit-appearance: none;
}

::-webkit-scrollbar-thumb {
  border-radius: 4px;
  background-color: rgba(0, 0, 0, .5);
  -webkit-box-shadow: 0 0 1px rgba(255, 255, 255, .87);
}
```

**Further questions/considerations**
- Have you considered how typeahead could be incorporated into this solution or is your thinking that this should form part of the ‘Select with search‘?
- Do you have research into how this feature has performed with users?

frankieroberto:
@simonwhatley:

> In lieu of the accessible autocomplete being included in the design system, I think this has the potential to serve a similar purpose elegantly.

Thanks! It’s hugely inspired by the accessible autocomplete, but one key difference is to try and focus on the "these are only suggestions, you can type your own answer" use case, where the non-js fallback is a plain text input rather than a `<select>` dropdown. This isn't always clear with the autocomplete, and so the use of the gap and the 'suggestions' header is to try and make it more obvious that you can ignore the suggestions and add your own answer (rather than have to pick the closest match).

The design is also inspired by the [ONS autosuggest](https://ons-design-system.netlify.app/components/autosuggest/). 

> **Proximity of suggestions** The spacing between the input and the suggestions is sufficiently large enough that some users may not perceive the suggestions as related.
> 
> I recommend reducing the space from `30px` to `10px`.

Nice idea, will try this. There’s of a trade-off between having too big of a gap, and it looking too much like a dropdown where you have to select one. But a 10px gap might be enough.
 
> **Clickable suggestions** On macOS, it‘s not clear whether the suggestions can be chosen, or whether the user needs to continue typing the name of the suggestion.

I'm not quite sure I follow this. Can you say more?

> **Scrollable suggestions** On macOS, it‘s not clear that the suggestions list can be scrolled.
> 
> I recommend adding CSS rules to resolve click and scroll

Good point, will add.

> * Have you considered how typeahead could be incorporated into this solution or is your thinking that this should form part of the ‘Select with search‘?

By 'typeahead' do you mean displaying the first suggestion in the input box itself, to the right of the cursor, with the remaining text selected so that it can be overtyped?

If so, I'm inclined not to do this, at least initially, as I think it can introduce issues with accessibility and people accidentally submitting an answer (eg it's not always obvious how to clear the selected text if your intended answer is a shorter version of a suggestion).

> * Do you have research into how this feature has performed with users?

Yes – we have done some initial testing of this within the "Apply for teacher training" service, in the context of entering degree details (subject and university) and so far it has worked well, with users being both able to select a suggestion, and also typing in their own answer when there were no relevant suggestions.

36degrees:
As discussed with Frankie, I'm going to close this as it's not something we can merge right now and hasn't been touched in over a year.

‘Autocomplete’ is one of [our next priorities](https://design-system.service.gov.uk/community/upcoming-components-patterns/). When we start working on it, we'll revisit this PR and see if there’s anything we can reuse. It's already referenced from our backlog issue on autocomplete.

PR #2458, opened 2021-12-03
Use native way to output negative spacing in hint

36degrees:
Update the hint component to use the new 'native' way to output negative spacing introduced in ca7a126 (#2348).

### DIFF
--- a/src/govuk/components/hint/_index.scss
+++ b/src/govuk/components/hint/_index.scss
@@ -37,6 +37,6 @@
 
   // Reduces visual spacing of legend when there is a hint
   .govuk-fieldset__legend + .govuk-hint {
-    margin-top: -(govuk-spacing(1));
+    margin-top: govuk-spacing(-1);
   }
 }


36degrees:
Given this is a small 'internal' change that has no impact to the outputted CSS, I don't think this needs a changelog entry?

lfdebrux:
There are also two places in the tabs component where we use `- govuk-spacing()`, do we want to change those as well?

https://github.com/alphagov/govuk-frontend/blob/0d7165626312ddb105a6514d7e98322c92c538a5/src/govuk/components/tabs/_index.scss#L27

https://github.com/alphagov/govuk-frontend/blob/0d7165626312ddb105a6514d7e98322c92c538a5/src/govuk/components/tabs/_index.scss#L80

36degrees:
Good catch – I did search for `govuk-spacing` and skim the results, but must have missed them.

I do now regret my choice of branch name 😆  Could follow up with a second PR?

lfdebrux:
> I do now regret my choice of branch name 😆 Could follow up with a second PR?

Whatever you prefer!

lfdebrux (approved):

PR #2459, opened 2021-12-03
Use native way to output negative spacing in tabs

36degrees:
Update the tabs component to use the new 'native' way to output negative spacing introduced in ca7a126 – as spotted by @lfdebrux in #2458.


36degrees:
As in #2458, given this is a small 'internal' change that has no impact to the outputted CSS I don't think this needs a changelog entry.

lfdebrux (approved):

PR #2460, opened 2021-12-06
Add npm configuration file to read token from environment

lfdebrux:
Rather than logging in to npm with a username and password we want to log in using an access token with limited privileges. This commit adds an `.npmrc` file that tells npm to read the access token from the `$NPM_TOKEN` environment variable, following instructions in the npm docs [[1]].

[1]: https://docs.npmjs.com/using-private-packages-in-a-ci-cd-workflow

### DIFF
--- a/.github/workflows/diff-change-to-dist.yaml
+++ b/.github/workflows/diff-change-to-dist.yaml
@@ -4,6 +4,9 @@ on:
   pull_request:
     paths: ['dist/**']
 
+env:
+  NPM_TOKEN: none  # needed for npm@6
+
 jobs:
   generate-diff:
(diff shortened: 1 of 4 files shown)


lfdebrux:
I thought adding the `.npmrc` file to the repo would be easier than having everyone have to manage it on their own machines, however there are a few annoying wrinkles:

Older versions of npm (including v6, which comes with Node 14) will raise an error on start if the config refers to an envvar which is not set. This means you can't run things like `npm install` without the envvar being set, which is super unhelpful.

36degrees:
I'm going to close this – we'll look at it again as part of #3526.

PR #2461, opened 2021-12-06
Update publishing instructions to use npm access token

lfdebrux:
We want to avoid using the password for our npm account, as if this leaks it can be used to create a lot of havoc. Instead we want to log in using an access token with limited privileges. Unfortunately the npm cli does not allow logging in with a token interactively, so instead we configure the project to look for a token in the `$NPM_TOKEN` environment variable (see #2460) and add instructions to developers on how to add the token to the environment from the credentials store.

---

This PR is a draft so we can test the instructions before reviewing.

Note that the credentials repo does not yet have an access token for npm.


36degrees:
I'm going to close this – we'll look at it again as part of #3526.

PR #2446, opened 2021-11-29
Add support for footer sections that span multiple columns

vanitabarrett:
Partly fixes https://github.com/alphagov/govuk-frontend/issues/1726

**Note:** this PR is based off the changes made in https://github.com/alphagov/govuk-frontend/pull/2428

Allow having a footer section that spans multiple columns but does not have multiple columns of list items. Useful if you have multiple rows of sections and want to align a section with only a few items with a section with many items (for instance the current GOV.UK footer layout).

Screenshot 2021-12-02 at 11 22 06
Screenshot 2021-12-02 at 11 21 19

Browsers to test in:

- [x] Internet Explorer 8
- [x] Internet Explorer 9
- [x] Internet Explorer 10
- [x] Internet Explorer 11
- [x] Chrome on macOS
- [x] Safari on macOS
- [x] Chrome on Windows
- [x] Firefox on macOS
- [x] Edge on Windows
- [x] Firefox on Windows
- [x] Safari on iOS
- [x] Chrome on iOS
- [x] Chrome on Android
- [x] Samsung Internet on Android

https://govuk-frontend-pr-2446.herokuapp.com/full-page-examples/campaign-page
https://govuk-frontend-pr-2446.herokuapp.com/components/footer/Full-GDS-example/preview

PR #2462, opened 2021-12-06
Fix footer alignment with grid classes and add `width` macro option

vanitabarrett:
Closes https://github.com/alphagov/govuk-frontend/issues/1539 and https://github.com/alphagov/govuk-frontend/issues/1726

## Why
This PR attempts to rework what was originally spiked in https://github.com/alphagov/govuk-frontend/pull/2446 and https://github.com/alphagov/govuk-frontend/pull/2428 to have the same outcome, but without using flexbox. When reviewing the previous PRs, we realised that due to the fallback provided for older browsers, we were 1) duplicating a lot of what our existing grid does, and 2) flexbox was only really being used for 'dynamic' half/half layouts. 

One positive of using flexbox was that we could support more 'dynamic' layouts where the footer sections could shrink/grow accordingly. However, the downsides are that it makes the code difficult to follow and it was potentially also a bit of "magic" for users.

## What
This PR makes the following changes:

- Strips out flexbox, uses our grid instead
- Removes certain assumptions we'd made about the layout of a footer dependent on the number of sections
- Introduces a `width` macro option to explicitly set the width of the section if needed, using our grid widths
- Sets the default width of a footer section to full width - this is most similar to the current behaviour where if you pass only one section it renders as full width

## Links for testing
https://govuk-frontend-pr-2462.herokuapp.com/components/footer
https://govuk-frontend-pr-2462.herokuapp.com/examples/footer-alignment
https://govuk-frontend-pr-2462.herokuapp.com/full-page-examples/campaign-page

## Screenshots
Screenshot 2021-12-06 at 15 49 38

## IE8
Screenshot 2021-12-06 at 15 50 41

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -198,6 +198,20 @@ To fix this error, replace any imports of `node_modules/govuk-frontend/govuk/cor
 
 This change was introduced in [pull request #22463: Move template styles from `core` to `objects` layer](https://github.com/alphagov/govuk-frontend/pull/2463).
 
+#### Check your footer displays as expected
+
+We’ve made some fixes to the alignment of columns within the footer component, so they align with our grid. We've also removed the logic that assumes a 2-section layout displays as a 'two-thirds and one-third' layout. Footer sections now display as full-width by default.
+
+If you're using the Nunjucks macros, check your footer displays as expected and use the `width` macro option to set the width you want for each section.
+
+If you're not using the Nunjucks macros, check your footer displays as expected and use the standard Design System grid classes on the `govuk-footer__section` element to set the width. For example:
+
+```html
+<div class="govuk-footer__section govuk-grid-column-two-thirds">...</div>
+```
+
+This change was introduced in [pull request #2462: Fix footer alignment with grid classes and add `width` macro option](https://github.com/alphagov/govuk-frontend/pull/2462).
+
 ### Optional changes
 
 We've recently made some other changes to GOV.UK Frontend. While these are not breaking changes, implementing them will make your service work better.
--- a/app/views/examples/footer-alignment/index.njk
+++ b/app/views/examples/footer-alignment/index.njk
@@ -123,6 +123,7 @@
     "navigation": [
       {
         "title": "Two column list",
+        "width": "two-thirds",
         "columns": 2,
         "items": [
           {
--- a/app/views/examples/footer-alignment/index.njk
+++ b/app/views/examples/footer-alignment/index.njk
@@ -39,8 +39,8 @@
   {{ govukFooter({
     "navigation": [
       {
-        "title": "Single column list 1",
-        "columns": 1,
+        "title": "One-third",
+        "width": "one-third",
         "items": [
           {
             "href": "#1",
@@ -57,8 +57,8 @@
         ]
       },
       {
-        "title": "Single column list 2",
-        "columns": 1,
+        "title": "One-third",
+        "width": "one-third",
         "items": [
           {
             "href": "#1",
@@ -75,8 +75,8 @@
         ]
       },
       {
-        "title": "Single column list 3",
-        "columns": 1,
+        "title": "One-third",
+        "width": "one-third",
         "items": [
           {
             "href": "#1",
@@ -96,7 +96,7 @@
   }) }}
 
   <div class="govuk-width-container">
-    <h2 class="govuk-heading-m govuk-!-margin-top-6">Two column list on the left</h2>
+    <h2 class="govuk-heading-m govuk-!-margin-top-6">Two-thirds (two columns) / one-third layout</h2>
     <div class="govuk-grid-row">
       <div class="govuk-grid-column-one-third">
         <h3 class="govuk-heading-m">One third</h3>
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -198,6 +198,20 @@ To fix this error, replace any imports of `node_modules/govuk-frontend/govuk/cor
 
 This change was introduced in [pull request #22463: Move template styles from `core` to `objects` layer](https://github.com/alphagov/govuk-frontend/pull/2463).
 
+#### Check your footer displays as expected
+
+We’ve made some fixes to the alignment of columns within the footer component, so they align with our grid. We have also removed the logic which assumes a two-section layout displays as two-thirds / one-third. Footer sections now display as full-width by default.
```
EoinShaughnessy, line 203:
```suggestion
We’ve made some fixes to the alignment of columns within the footer component, so they align with our grid. We've also removed the logic that assumes a 2-section layout displays as a 'two-thirds and one-third' layout. Footer sections now display as full-width by default.
```
Reasons for suggestion:

- style guide say to write out whole numbers unless they're one/1
- style guide says to avoid dashes when they're standing in for a word
- in our Layout guidance, we've referred to 'two-thirds and one third layout'*

*There should be a hyphen between 'one' and 'third' in the guidance - I'll fix that

CHANGELOG.md
```diff
@@ -198,6 +198,20 @@ To fix this error, replace any imports of `node_modules/govuk-frontend/govuk/cor
...
+
+If you are using the Nunjucks macros, check your footer displays as expected and use the `width` macro option to set the desired width of each section.
```
EoinShaughnessy, line 205:
```suggestion
If you're using the Nunjucks macros, check your footer displays as expected and use the `width` macro option to set the width you want for each section.
```

CHANGELOG.md
```diff
@@ -198,6 +198,20 @@ To fix this error, replace any imports of `node_modules/govuk-frontend/govuk/cor
...
+
+If you are not using the Nunjucks macros, check your footer displays as expected and use the standard design system grid classes on the `govuk-footer__section` element to set the width, for example:
```
EoinShaughnessy, line 207:
```suggestion
If you're not using the Nunjucks macros, check your footer displays as expected and use the standard Design System grid classes on the `govuk-footer__section` element to set the width. For example:
```

EoinShaughnessy (commented):
Content looks nice and clear! Just added a few comments.

36degrees (commented):
Generally think the approach is sound and the changes are looking good – a few minor comments.

Something of an edge case, but long words can overflow across other columns – but this is also true of the grid elsewhere. Not sure if it's worth worrying about as I couldn't think of any realistic examples, other than a long email address which seems an like unlikely thing to put in a section.

Screenshot 2021-12-08 at 14 57 01

src/govuk/components/footer/template.test.js
```diff
@@ -178,6 +178,30 @@ describe('footer', () => {
...
+      expect($section.hasClass('govuk-grid-column-full')).toBeTruthy()
+    })
+
+    it('renders two-column section full width by default', () => {
+      const $ = render('footer', examples['with default width navigation (two columns)'])
+
+      const $component = $('.govuk-footer')
+      const $section = $component.find('.govuk-footer__section')
+      expect($section.hasClass('govuk-grid-column-full')).toBeTruthy()
+    })
```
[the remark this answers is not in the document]
vanitabarrett: Yep, you're right, that is the only difference. I think I added it just to double check that columns being set doesn't have any impact on the width class being set because originally they were linked. But maybe that's not really worth testing now?

app/views/examples/footer-alignment/index.njk
```diff
@@ -39,8 +39,8 @@
   {{ govukFooter({
```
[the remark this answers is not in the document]
vanitabarrett: Good shout, have updated the examples to include examples for `one-half` and `one-quarter`

src/govuk/components/footer/template.test.js
```diff
@@ -178,6 +178,30 @@ describe('footer', () => {
       const $list = $component.find('ul.govuk-footer__list')
       expect($list.hasClass('govuk-footer__list--columns-2')).toBeTruthy()
     })
+
+    it('renders one-column section full width by default', () => {
+      const $ = render('footer', examples['with default width navigation (one column)'])
+
+      const $component = $('.govuk-footer')
+      const $section = $component.find('.govuk-footer__section')
```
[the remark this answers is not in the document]
vanitabarrett: Thanks, good point! I've added a new commit to remove these in the footer tests.

CHANGELOG.md
vanitabarrett: Thanks @EoinShaughnessy ! Quick question, I'm not sure how changing from `two-section` to `2-section` fits in with "style guide say to write out whole numbers unless they're one/1"? Or am I mis-understanding what that means?

EoinShaughnessy: No, totally my bad - sorry! What I should have said was, one is usually the only number we write as a word - all others we can write as numbers. Gah, sorry for the brain fog.

vanitabarrett: Thanks for clarifying!

EoinShaughnessy (approved):
Looks great! 👍🏻

36degrees (approved):
Ship it! 🚢

PR #2467, opened 2021-12-08
Set skip link linked element in init and put event listener behind flag 

hannalaakso:
This PR solves the issue mentioned in https://github.com/alphagov/govuk-frontend/pull/2450#pullrequestreview-825124287.

Improve the readability and performance of the skip link script by:
- checking for the linked element early as part of the initialisation 
- moving the event listener for the main content behind a flag so it's only set once
- simplifying `init()` and `getFragmentFromUrl()` 
- removing redundant `clickEventHandler()`

~This also doesn’t seem to play nice with the `exported Components can be initialised` test we have in `all.test.js`, I don’t fully understand right now whether it’s picked up on an issue or if the test is not working correctly. Was going to do some test debugging tomorrow anyway so could look at this then.~ Solved in https://github.com/alphagov/govuk-frontend/pull/2470

Tested in 

Operating system | Browser | Supports
-- | -- | --
Windows | Internet Explorer 11 | x
  | Edge (latest versions) | x
  | Google Chrome (latest versions) | x
  | Mozilla Firefox (latest versions) | x
macOS | Safari 12 and later | x
  | Google Chrome (latest versions) | x
  | Mozilla Firefox (latest versions) | x
iOS | Safari for iOS 12.1 and later |  
  | Google Chrome (latest versions) |  
Android | Google Chrome (latest versions) |  
  | Samsung Internet (latest versions) |

### DIFF
--- a/src/govuk/components/skip-link/skip-link.js
+++ b/src/govuk/components/skip-link/skip-link.js
@@ -4,90 +4,91 @@ import '../../vendor/polyfills/Event' // addEventListener and event.target norma
 
 function SkipLink ($module) {
   this.$module = $module
+  this.$linkedElement = null
+  this.linkedElementListener = false
 }
 
 /**
  * Initialise the component
  */
 SkipLink.prototype.init = function () {
-  var $module = this.$module
   // Check for module
-  if (!$module) {
+  if (!this.$module) {
     return
   }
 
-  $module.addEventListener('click', this.handleClick.bind(this))
+  // Check for linked element
+  this.$linkedElement = this.getLinkedElement()
+  if (!this.$linkedElement) {
+    return
+  }
+
+  this.$module.addEventListener('click', this.focusLinkedElement.bind(this))
 }
 
 /**
-* Click event handler
+* Get linked element
 *
-* @param {MouseEvent} event - Click event
+* @returns {HTMLElement} $linkedElement - DOM element linked to from the skip link
 */
-SkipLink.prototype.handleClick = function (event) {
-  var target = event.target
+SkipLink.prototype.getLinkedElement = function () {
+  var linkedElementId = this.getFragmentFromUrl()
 
-  if (this.focusLinkedElement(target)) {
-    event.preventDefault()
+  if (!linkedElementId) {
+    return false
   }
+
+  return document.getElementById(linkedElementId)
 }
 
 /**
  * Focus the linked element
  *
- * @param {HTMLElement} $target - Event target
- * @returns {boolean} True if the linked element was able to be focussed
+ * Set tabindex and helper CSS class. Set listener to remove them on blur.
  */
-SkipLink.prototype.focusLinkedElement = function ($target) {
-  // If the element that was clicked does not have a href, return early
-  if ($target.href === false) {
-    return false
-  }
-
-  var linkedElementId = this.getFragmentFromUrl($target.href)
-  var $linkedElement = document.getElementById(linkedElementId)
-  if (!$linkedElement) {
-    return false
-  }
+SkipLink.prototype.focusLinkedElement = function () {
+  var $linkedElement = this.$linkedElement
 
   if (!$linkedElement.getAttribute('tabindex')) {
-    // Set the content tabindex to -1 so it can be focused with JavaScript.
+    // Set the element tabindex to -1 so it can be focused with JavaScript.
     $linkedElement.setAttribute('tabindex', '-1')
     $linkedElement.classList.add('govuk-skip-link-focused-element')
 
-    $linkedElement.addEventListener('blur', this.removeFocusProperties.bind(this, $linkedElement))
+    // Add listener for blur on the focused element (unless the listener has previously been added)
+    if (!this.linkedElementListener) {
+      this.$linkedElement.addEventListener('blur', this.removeFocusProperties.bind(this))
+      this.linkedElementListener = true
+    }
   }
   $linkedElement.focus()
 }
 
 /**
- * Remove the tabindex that makes the linked element focusable because the content only needs to be
+ * Remove the tabindex that makes the linked element focusable because the element only needs to be
  * focusable until it has received programmatic focus and a screen reader has announced it.
  *
  * Remove the CSS class that removes the native focus styles.
- *
- * @param {HTMLElement} $linkedElement - DOM element linked to from the skip link
  */
-SkipLink.prototype.removeFocusProperties = function ($linkedElement) {
-  $linkedElement.removeAttribute('tabindex')
-  $linkedElement.classList.remove('govuk-skip-link-focused-element')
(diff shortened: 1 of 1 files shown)


36degrees:
> This also doesn’t seem to play nice with the `exported Components can be initialised` test we have in `all.test.js`, I don’t fully understand right now whether it’s picked up on an issue or if the test is not working correctly. Was going to do some test debugging tomorrow anyway so could look at this then.

I think I've managed to get to the bottom of this…

We're bailing out of the `getLinkedElement` if `href` strictly equals false.

https://github.com/alphagov/govuk-frontend/blob/91f6254b1507f3ac74636d5eca9e8eff09999640/src/govuk/components/skip-link/skip-link.js#L39-L42

If you try to get the `href` property of an element that does not have an `href` attribute, you'll get `undefined` – and `undefined` != `false` – so the function does not return and continues to try and find the `indexOf('#')` against a variable which is undefined.

BUT – why would the skip link be initialised on an element that doesn't have an `href`?! Especially in the context of that test, where we're only initialising modules within a particular part of the page – which does not include a skip link.

It looks like the error is actually coming from the test immediately before it:

https://github.com/alphagov/govuk-frontend/blob/429af6119018b0829af37e90085641568c0a5c02/src/govuk/all.test.js#L61-L76

There are a few unusual things going on in that test:

1. The test initialises each module whilst passing in `document.documentElement` as the `$module`, whereas the module JavaScript expects the element with the `data-module` attribute – so the skip link JS is being initialised against the `<body>` which obviously does not have an `href`!
2. If we wanted to initialise with the correct HTML for each component, the tests are being run against the 'home page' of the review app, which doesn't include the majority of them.
3. We do not wait for the calls to `page.evaluate` on line 70 to complete – those calls are 'fired and forgotten', so by the time the errors in the JS occur we've moved on to the next test – which is why the tests are surfacing in the wrong place.

You can make the error surface in the correct test by changing lines 69 to 75 so that we `await` the evaluation:

```diff
       // Check that all the components on the GOV.UK Frontend global can be initialised
-      components.forEach(component => {
-        page.evaluate(component => {
+      await page.evaluate((components) => {
+        components.forEach(component => {
           const Component = window.GOVUKFrontend[component]
           const $module = document.documentElement
           new Component($module).init()
-        }, component)
-      })
+        })
+      }, components)
     })
```

However, I am still not sure what we should do with the test… it's not testing what we think it does, and at the minute seems to offer little value. From talking to @lfdebrux, the suggestions so far are to either delete it, or to alter it so that it only checks that there is an `init` function for each component, and not try to execute it. Thoughts?

36degrees:
I've raised a PR to update the test (#2470)

src/govuk/components/skip-link/skip-link.js
```diff
@@ -83,6 +94,11 @@ SkipLink.prototype.removeFocusProperties = function ($linkedElement) {
  * @returns {string} Fragment from URL, without the hash
  */
 SkipLink.prototype.getFragmentFromUrl = function (url) {
+  // Bail if the url is not a string (for example if it's undefined)
```
36degrees, line 97: Just noticed that there's a method [`HTMLAnchorElement.hash`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLAnchorElement/hash) which we might be able to use here to simplify things a little.

36degrees (commented):
I think this approach makes sense – nice work 👍🏻  I pushed a commit to fix the issue in my previous comment, hope that's OK.

There's a couple of other things that I think we might need to look at, and a couple of suggestions.

src/govuk/components/skip-link/skip-link.js
```diff
@@ -4,71 +4,82 @@ import '../../vendor/polyfills/Event' // addEventListener and event.target norma
...
 /**
  * Remove the tabindex that makes the linked element focusable because the content only needs to be
  * focusable until it has received programmatic focus and a screen reader has announced it.
  *
  * Remove the CSS class that removes the native focus styles.
- *
- * @param {HTMLElement} $linkedElement - DOM element linked to from the skip link
  */
-SkipLink.prototype.removeFocusProperties = function ($linkedElement) {
+SkipLink.prototype.removeFocusProperties = function () {
```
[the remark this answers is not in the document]
hannalaakso: Good point. I've now moved the event listener to sit inside the `tabindex` check so it's now only removed if it's set.  

I'm going to leave the `tabindex` check in place though as we allow the linked element to be overridden so we don't know what it contains. I've seen mentions of some services containing a secondary skip link to skip to specific parts of the page so it's not always used for the main content.

src/govuk/components/skip-link/skip-link.js
```diff
@@ -4,71 +4,82 @@ import '../../vendor/polyfills/Event' // addEventListener and event.target norma
...
   if (!$linkedElement.getAttribute('tabindex')) {
     // Set the content tabindex to -1 so it can be focused with JavaScript.
     $linkedElement.setAttribute('tabindex', '-1')
     $linkedElement.classList.add('govuk-skip-link-focused-element')
-
-    $linkedElement.addEventListener('blur', this.removeFocusProperties.bind(this, $linkedElement))
   }
   $linkedElement.focus()
+
+  return $linkedElement === document.activeElement
```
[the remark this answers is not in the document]
hannalaakso: Yes, that behaviour is not very useful here, good point 👍

src/govuk/components/skip-link/skip-link.js
hannalaakso: I've had a go at using it, have a look. I removed the check for the type of `string` as the native check for the hash on an anchor link should handle that.

src/govuk/components/skip-link/skip-link.js
```diff
@@ -4,90 +4,91 @@ import '../../vendor/polyfills/Event' // addEventListener and event.target norma
...
-    // Set the content tabindex to -1 so it can be focused with JavaScript.
+    // Set the element tabindex to -1 so it can be focused with JavaScript.
     $linkedElement.setAttribute('tabindex', '-1')
     $linkedElement.classList.add('govuk-skip-link-focused-element')
 
-    $linkedElement.addEventListener('blur', this.removeFocusProperties.bind(this, $linkedElement))
+    // Add listener for blur on the focused element (unless the listener has previously been added)
+    if (!this.linkedElementListener) {
+      this.$linkedElement.addEventListener('blur', this.removeFocusProperties.bind(this))
+      this.linkedElementFocused = true
```
36degrees, line 60: If I understand the intention correctly, I think this needs to be `linkedElementListener` for the guard to work?

```suggestion
      this.linkedElementListener = true
```

hannalaakso: Thanks for the eagle eye, I kept fluctuating on the variable name and this slipped through.

36degrees (approved):

PR #2470, opened 2021-12-09
Fix flawed 'components can be initialised' test

36degrees:
The test currently passes `document.documentElement` (the root `<html>` element) as the `$module` when initialising each component. This is not what the component expects – `$module` should be the element with the corresponding `data-module` attribute, for example `<div data-module="govuk-button">`.

If we did want to initialise with the correct HTML for each component, we'd need to make substantial changes to the test. At the minute it runs in the context of the 'home page' of the review app, which doesn't include the majority of the components that have JavaScript.

Given these constraints, replace the test with one that does not try to instantiate or initialise the components, but only check that they all have an `init` function.

We were also not waiting for the Promises returned by `page.evaluate` to by fulfilled, so any errors thrown by the JS could cause unrelated tests that run later to fail. We were seeing this occur in #2467 when adding tests for the new skip link JavaScript.

Fix this by moving the bulk of the test logic into a single `page.evaluate` call and using `await` to pause execution until the returned Promise is fulfilled.

(It would likely make more sense to run these tests against the classes directly, rather than using Puppeteer to assert things about the `window.GOVUKFrontend` global object in a browser. However, we don't currently have a good way to test the component JavaScript directly. When we do, we should revisit this test.)

### DIFF
--- a/src/govuk/all.test.js
+++ b/src/govuk/all.test.js
@@ -58,21 +58,20 @@ describe('GOV.UK Frontend', () => {
         'Tabs'
       ])
     })
-    it('exported Components can be initialised', async () => {
+    it('exported Components have an init function', async () => {
       await page.goto(baseUrl + '/', { waitUntil: 'load' })
 
-      const GOVUKFrontendGlobal = await page.evaluate(() => window.GOVUKFrontend)
+      var componentsWithoutInitFunctions = await page.evaluate(() => {
+        var components = Object.keys(window.GOVUKFrontend)
+          .filter(method => method !== 'initAll')
 
-      var components = Object.keys(GOVUKFrontendGlobal).filter(method => method !== 'initAll')
-
-      // Check that all the components on the GOV.UK Frontend global can be initialised
-      components.forEach(component => {
-        page.evaluate(component => {
-          const Component = window.GOVUKFrontend[component]
-          const $module = document.documentElement
-          new Component($module).init()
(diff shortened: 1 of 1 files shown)


lfdebrux (approved):
This works for me; I managed to make the test fail in the expected way, by changing the name of the init function for the button component:

```
 FAIL  src/govuk/all.test.js (13.514 s)
  ● GOV.UK Frontend › javascript › exported Components have an init function

    expect(received).toEqual(expected) // deep equality

    - Expected  - 1
    + Received  + 3

    - Array []
    + Array [
    +   "Button",
    + ]

      72 |       })
      73 |
    > 74 |       expect(componentsWithoutInitFunctions).toEqual([])
         |                                              ^
      75 |     })
      76 |     it('can be initialised scoped to certain sections of the page', async () => {
      77 |       await page.goto(baseUrl + '/examples/scoped-initialisation', { waitUntil: 'load' })

      at Object.<anonymous> (src/govuk/all.test.js:74:46)
```

I'm still not totally certain that it makes sense, but I think it makes the same amount of sense as what was there before; I think the idea is we're ensuring that we keep our side of the API contract, and have an `init` function for each component.

36degrees:
Yeah, I was in two minds as to whether to just delete the test entirely to be honest… but I think this is 'good enough for now' and we can revisit this as part of the upcoming JavaScript work.

Issue #2456, opened 2021-12-03
Allow `govuk-font` mixin to bypass responsive typography


owenatgov:
## Proposal
Add an option to the [`govuk-font` sass mixin](https://frontend.design-system.service.gov.uk/sass-api-reference/#govuk-font) to allow for bypassing of the design system responsive typography.

The desired outcome would be a way to use `govuk-font` in such a way that it includes all the desired helpers like anti-aliasing and print styles but doesn't produce additional media queries where the font size goes down at lower screen sizes. How this might look:

```scss
.my-element {
    @include govuk-font($size: 16, $responsive: false);
}
```

The above element would have a font size of `16px` and the `em` equivalent at all screen sizes.

## Context
On govuk, we've recently been doing some work to set a new design tone for govuk that provides a better experience on smaller devices. We'll be able to see this soon in the launch of the new homepage design. Something we've run into however is that there are instances where we want to maintain font size across screen sizes in order to both create a consistent experience for users and to adhere to good designs across screen sizes.

In order to meet our needs, we've had to bypass `govuk-font` and manually enter font values and associated extra attributes.

Examples of govuk components where we've done this:

- https://components.publishing.service.gov.uk/component-guide/layout_super_navigation_header
- https://components.publishing.service.gov.uk/component-guide/big_number

Pull requests that illustrate the problem:

- https://github.com/alphagov/govuk_publishing_components/pull/2493
- https://github.com/alphagov/govuk_publishing_components/pull/2494

We're conscious that there are mixins like [`govuk-typography-common`](https://frontend.design-system.service.gov.uk/sass-api-reference/#govuk-typography-common) that help with this approach, however the ideal would be to use `govuk-font` as it reduces risk of maintenance chain reactions in our code if `govuk-font` or any of it's settings ever change and means our code is cleaner.

There is also an accessibility concern as `govuk-font` encourages font size going below `16px` on mobile if you want a small font size on desktop.

This issue is tangentially related https://github.com/alphagov/govuk-frontend/issues/2418 but feels separate enough to this ask not to combine then, in my opinion.

36degrees:
You can currently pass `$size: false` to `govuk-font` which will give you the basic font properties without any size declaration:

```scss
.app-my-class {
  @include govuk-font($size: false);
}
```

```css
.app-my-class {
  font-family: "GDS Transport", arial, sans-serif;
  -webkit-font-smoothing: antialiased;
  -moz-osx-font-smoothing: grayscale;
[...]

PR #2474, opened 2021-12-13
Add tests for the `govuk-font` mixin

36degrees:
I noticed when looking into #2456 that we had no test coverage for the `govuk-font` mixin. The `govuk-font` mixin does quite a lot of things – most of it handing off to the other mixins in this file – but does have responsibility for setting up things like tabular numbers.

Add tests to try and cover all of the execution paths within the mixin.

### DIFF
--- a/src/govuk/helpers/typography.test.js
+++ b/src/govuk/helpers/typography.test.js
@@ -60,6 +60,7 @@ describe('@mixin govuk-typography-common', () => {
     expect(resultsString).toContain('font-family: "GDS Transport"')
     expect(resultsString).toContain('font-family: "GDS Transport"')
   })
+
   it('should not output a @font-face declaration when the user has changed their font', async () => {
     const sass = `
     $govuk-font-family: Helvetica, Arial, sans-serif;
@@ -82,6 +83,7 @@ describe('@mixin govuk-typography-common', () => {
     expect(resultsString).not.toContain('font-family: "GDS Transport"')
     expect(resultsString).not.toContain('font-family: "ntatabularnumbers"')
   })
+
   it('should not output a @font-face declaration when the user wants compatibility with GOV.UK Template', async () => {
     const sass = `
     $govuk-compatibility-govuktemplate: true;
@@ -103,6 +105,7 @@ describe('@mixin govuk-typography-common', () => {
     expect(resultsString).toContain('font-family: "nta"')
     expect(resultsString).toContain('font-family: "ntatabularnumbers"')
   })
+
   it('should not output a @font-face declaration when the user has turned off this feature', async () => {
     const sass = `
     $govuk-include-default-font-face: false;
@@ -124,6 +127,7 @@ describe('@mixin govuk-typography-common', () => {
     expect(resultsString).toContain('font-family: "GDS Transport"')
     expect(resultsString).toContain('font-family: "GDS Transport"')
   })
+
   it('should not output a @font-face declaration when the browser is IE8', async () => {
     const sass = `
     $govuk-is-ie8: true;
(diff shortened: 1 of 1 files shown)


src/govuk/helpers/typography.test.js
```diff
@@ -419,4 +417,145 @@ describe('@mixin govuk-typography-responsive', () => {
...
+      $govuk-include-default-font-face: false;
+      ${sassBootstrap}
+
+      .foo {
+        @include govuk-font($size: 14)
+      }`
+
+      const results = await renderSass({ data: sass, ...sassConfig })
+
+      expect(results.css.toString().trim()).toBe(outdent`
```
36degrees, line 434: This is basically a 'snapshot test', but I couldn't think of a better way of doing it… this does mean that whenever anything changes in the govuk-font mixin we'll need to update this test accordingly. It's not ideal, but I think that's OK?

src/govuk/helpers/typography.test.js
```diff
@@ -419,4 +417,145 @@ describe('@mixin govuk-typography-responsive', () => {
...
+      ${sassBootstrap}
+
+      .foo {
+        @include govuk-font($size: 12)
+      }`
+
+      const results = await renderSass({ data: sass, ...sassConfig })
+
+      expect(results.css.toString()).toContain('font-size: 12px')
+      expect(results.css.toString()).not.toContain('font-size: 14px')
```
36degrees, line 500: This feels a little clunky, but because `$size: 14` _also_ outputs `font-size: 12px` (for mobile) it'd pass anyway… suggestions welcome.

src/govuk/helpers/typography.test.js
```diff
@@ -419,4 +417,145 @@ describe('@mixin govuk-typography-responsive', () => {
...
+      const results = await renderSass({ data: sass, ...sassConfig })
+      const css = results.css.toString()
+
+      expect(css).toContain('font-feature-settings: "tnum" 1;')
+      expect(css).toContain(outdent`
+      ${outdent}
+        @supports (font-variant-numeric: tabular-nums) {
+          .foo {
+            font-feature-settings: normal;
+            font-variant-numeric: tabular-nums; } }`)
```
36degrees, line 470: Thought it was worth testing for both of these, but not worth writing individual tests for them.

src/govuk/helpers/typography.test.js
```diff
@@ -419,4 +413,145 @@ describe('@mixin govuk-typography-responsive', () => {
...
+      .foo {
+        @include govuk-font($size: 14, $weight: bold)
+      }`
+
+      const results = await renderSass({ data: sass, ...sassConfig })
+
+      expect(results.css.toString()).toContain('font-weight: 700')
+    })
+
+    it('does not output font-weight if $weight: false', async () => {
```
vanitabarrett, line 531: @36degrees do you think it's worth having a test for an unrecognised $weight?

36degrees: Hmm, good question… at the minute it'll basically ignore `$weight if it's not `regular` or `bold`. What would the assertion be for that test? Just check that the output does not contain `font-weight`?

vanitabarrett: I think so, yep. If that's the behaviour we want to keep!

36degrees: Added one here – https://github.com/alphagov/govuk-frontend/pull/2474/files#diff-1aafb81d87bfc00119f85ee5e8f77ebdb31964d61786c89d2b51746be0cd943dR546-R559

vanitabarrett (approved):

PR #2475, opened 2021-12-14
Tweak whitespace in input component HTML for improved readability

vanitabarrett:
Tweak the input component nunjucks macro slightly to improve the whitespace of the output HTML so that:

- the closing `div` for the form group is on a new line
(https://github.com/alphagov/govuk-design-system/issues/1939)
- there is no unecessary whitespace between the `govuk-input__wrapper`
and `govuk-input__prefix`

## Before vs After (default input example)
Before:
Screenshot 2021-12-14 at 14 36 57

After:
Screenshot 2021-12-14 at 14 37 03

## Before vs After (input with prefix and suffix)
Before:
Screenshot 2021-12-14 at 14 36 29

After:
Screenshot 2021-12-14 at 14 36 37

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixes
+
+- [#2475: Tweak whitespace in input component HTML for improved readability](https://github.com/alphagov/govuk-frontend/pull/2475)
+
 ## 4.0.0 (Breaking release)
 
 ### Breaking changes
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+## Fixes
```
36degrees, line 5:
```suggestion
### Fixes
```

36degrees (approved):
I think this makes sense – just a minor issue re heading levels in the changelog.

We should keep an eye out for feedback from those using our fixtures – hopefully most will be testing in a way that's forgiving of whitespace changes, but otherwise this sort of 'tweaking' might create extra work for them.

PR #2479, opened 2021-12-15
Remove unnecessary use of `aria-hidden` in the character count

36degrees:
The `govuk-character-count__message--disabled` class hides the message using `visibility: hidden`, so there is no need to also use the `aria-hidden` attribute.

> If an interactive element is hidden using `display:none` or `visibility:hidden` (either on the element itself, or any of the element's ancestors), it won't be focusable, and it will also be removed from the accessibility tree. This makes the addition of `aria-hidden="true"` or explicitly setting tabindex="-1" unnecessary.
>
> – https://www.w3.org/TR/using-aria/#4thrule

### DIFF
--- a/src/govuk/components/character-count/character-count.js
+++ b/src/govuk/components/character-count/character-count.js
@@ -133,12 +133,8 @@ CharacterCount.prototype.updateCountMessage = function () {
   var thresholdValue = maxLength * thresholdPercent / 100
   if (thresholdValue > currentLength) {
     countMessage.classList.add('govuk-character-count__message--disabled')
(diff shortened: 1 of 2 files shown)


36degrees:
Everything I've read suggests that nothing will be announced by the live region (as intended) when the element is hidden using `visibility: hidden`, but I will double check that just in case.

36degrees:
Closing this as this needs further investigation… from testing there are a few issues with the current implementation that should be addressed first:

- #2485 
- #2486
- #2487
- #2488

PR #2489, opened 2022-01-04
Set granular token permissions for GitHub Actions

36degrees:
Following best practise, [grant the GITHUB_TOKEN only the required permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token) – which in this case is write access on pull requests in order to be able to add the comment to the PR.

This also allows us to change the default permissions for workflows to 'Read repository contents' only.

### DIFF
--- a/.github/workflows/diff-change-to-dist.yaml
+++ b/.github/workflows/diff-change-to-dist.yaml
@@ -4,6 +4,9 @@ on:
   pull_request:
     paths: ['dist/**']
 
+permissions:
+  pull-requests: write
+
 jobs:
   generate-diff:
     name: Generate Diff


36degrees:
Permissions on the last workflow run according to [the workflow run log](https://github.com/alphagov/govuk-frontend/runs/4535812483?check_suite_focus=true#step:1:14) (with default workflow permissions set to 'Read and write permissions'):

```
GITHUB_TOKEN Permissions
  Actions: write
  Checks: write
  Contents: write
  Deployments: write
  Discussions: write
  Issues: write
  Metadata: read
  Packages: write
  Pages: write
  PullRequests: write
  RepositoryProjects: write
  SecurityEvents: write
  Statuses: write
```

Permissions with this change according to the [workflow run log](https://github.com/alphagov/govuk-frontend/runs/4704397701?check_suite_focus=true#step:1:15) (with default workflow permissions set to 'Read repository contents' only):

```
GITHUB_TOKEN Permissions
  Metadata: read
  PullRequests: write
```

lfdebrux (approved):

Issue #1068, opened 2018-11-19
Allow ErrorSummary component to disable autofocus


nickcolley:
This may only be a need when using the ErrorSummary in the Design System website, but there might be cases where someone's application had different load events.

https://github.com/alphagov/govuk-design-system/pull/643

36degrees:
We need to establish conventions for passing options to components to be able to pick this up.

nickcolley:
I got pretty far down the road to figuring this out here: https://github.com/alphagov/govuk-frontend/pull/1238

We decided not to do it since Alex did not need this work to solve this problem in the end.

36degrees:
Closed by #2494

PR #2491, opened 2022-01-05
Prevent error summary from being re-focused after it has been initially focused on page load

36degrees:
Always adding `tabindex="-1"` to the error summary means that as well as being programatically focusable the summary will also take focus if a user clicks within it, which may be confusing as the focus indicator will appear.

Instead, add the tabindex just before focusing the element, and then remove the tabindex attribute as soon as focus is lost, preventing the element from taking focus again afterwards.

This also makes the approach consistent with the notification banner and skip link, which also remove tabindex on blur.

It would make sense for users to remove the tabindex attribute from their own markup as we are now adding it using JavaScript, but the component should work fine either way so I don't believe this is a breaking change.

This work is being done as part of https://github.com/alphagov/govuk-frontend/issues/1068 – a future PR will make it possible to disable the autofocus functionality.

### DIFF
--- a/src/govuk/components/error-summary/error-summary.test.js
+++ b/src/govuk/components/error-summary/error-summary.test.js
@@ -13,6 +13,15 @@ describe('Error Summary', () => {
     expect(moduleName).toBe('govuk-error-summary')
   })
 
+  it('removes the tabindex attribute on blur', async () => {
+    await page.goto(baseUrl + '/components/error-summary/preview', { waitUntil: 'load' })
+
+    await page.$eval('.govuk-error-summary', el => el.blur())
+
+    const tabindex = await page.$eval('.govuk-error-summary', el => el.getAttribute('tabindex'))
+    expect(tabindex).toBeNull()
+  })
+
   const inputTypes = [
     // [description, input id, selector for label or legend]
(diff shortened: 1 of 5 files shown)


src/govuk/components/error-summary/error-summary.test.js
```diff
@@ -13,6 +13,15 @@ describe('Error Summary', () => {
     expect(moduleName).toBe('govuk-error-summary')
   })
 
+  it('removes the tabindex attribute on blur', async () => {
```
vanitabarrett, line 16: Do we need to check that the `tabindex` is being set correctly too? I think this test could pass if `tabindex` was never set in the first place?

36degrees: Good question. This test does pass, but the test before it fails – because tabindex isn't set, focus doesn't move to the error summary. So it's sort of there but implicit at the moment. Happy to add an explicit test for it though.

vanitabarrett: @36degrees I think it'd be good to have an explicit test so we don't unintentionally change some behaviour/test without realising 🙂

vanitabarrett (approved):

PR #2492, opened 2022-01-06
Clarify how `expanded` param works on accordion

EoinShaughnessy:
Partly addresses [#1443](https://github.com/alphagov/govuk-design-system/issues/1443).

This PR updates the Description for the `expanded` param in the accordion Nunjucks. The previous wording was unclear about when the section expanded.

### DIFF
--- a/src/govuk/components/accordion/accordion.yaml
+++ b/src/govuk/components/accordion/accordion.yaml
@@ -47,7 +47,7 @@ params:
     - name: expanded
       type: boolean
       required: false
-      description: Whether the section should be expanded upon initial load or not. Defaults to `false`.
+      description: Sets whether the section should be expanded when the page loads for the first time. Defaults to `false`.
 
 examples:
 - name: default


EoinShaughnessy:
We've had user feedback about this content:

> I feel like the guidance (particularly the nunjucks option for expanded) could be clearer - it gave me confidence that there was an option to set the expanded state, but this didn't work, and doesn't work as expected.
> 
> I also feel like there should be a way to set the state on load - but this doesn't seem possible.

[See issue-comment for full context.](https://github.com/alphagov/govuk-design-system/issues/1443#issuecomment-756105435)

Are we certain the current wording is accurate about what users have to do?

36degrees:
@EoinShaughnessy there's a 'gotcha' with the accordion that might be worth trying to mention here.

The accordion remembers which sections are open (using [session storage](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage)), so that for example if you go 'back' in your browser you'll see the accordion with whatever sections you had previously open visible.

But that also means that if you were to:

1. Load the page in your browser
2. Make changes to the `expanded` option for a section
3. Refresh the page in your browser

A user would probably expect to see that section now expanded, but it won't be because it will just restore whatever sections were open in step 1.

However, if they opened a new tab and viewed the same page then they would see the changes they expect.

How we cover that succinctly in the docs I've no idea!

EoinShaughnessy:
@36degrees That's great info and context, thanks.

Ok, first pass at Nunjucks update:

> Sets whether the section should be open upon initial load. Defaults to `false`. The accordion uses [session storage](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage) to 'remember' which sections users have opened. This means, if the user goes back in their browser, any sections they've opened earlier will still be open. However, if they load the page, open sections, and then refresh, the accordion will not remember which sections they've opened. To make the accordion remember, they need to open the page in a new tab after they've made their changes.

36degrees:
> @36degrees That's great info and context, thanks.
> 
> Ok, first pass at Nunjucks update:
> 
> > Sets whether the section should be open upon initial load. Defaults to `false`. The accordion uses [session storage](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage) to 'remember' which sections users have opened. This means, if the user goes back in their browser, any sections they've opened earlier will still be open. However, if they load the page, open sections, and then refresh, the accordion will not remember which sections they've opened. To make the accordion remember, they need to open the page in a new tab after they've made their changes.

I think that's all factually correct, but I don't think it makes sense to talk about users here. This should only really affect developers (or, for example, people using the Prototype Kit) who are making changes to the `expanded` option and not seeing the changes reflected in their browser. An 'end user' should never need to know or care about any of this.

I also wonder if we're giving too much information here – there must be a way to summarise this… perhaps we could try and find some time to chat about it tomorrow?

EoinShaughnessy:
@36degrees Thanks - have emailed invite, feel free to reschedule.

src/govuk/components/accordion/accordion.yaml
```diff
@@ -47,7 +47,7 @@ params:
     - name: expanded
       type: boolean
       required: false
-      description: Whether the section should be expanded upon initial load or not. Defaults to `false`.
+      description: Sets whether the section should be expanded upon initial load. Defaults to `false`.
```
EoinShaughnessy, line 50:
```suggestion
      description: Sets whether the section should be expanded when the page loads for the first time. Defaults to `false`.
```

EoinShaughnessy:
@36degrees Can I get a review? :)

36degrees (approved):
Thanks @EoinShaughnessy 👏🏻

EoinShaughnessy:
@36degrees Thanks for the reviews and pair-writes!

PR #2494, opened 2022-01-07
Allow disabling autofocus on error summary

36degrees:
Make it possible to disable the autofocus behaviour on the error summary by setting a data attribute on the module (or by setting `disableAutoFocus` if using the Nunjucks macro). This is consistent with the approach currently used for the notifcation banner.

The main use case for this is on the Design System, where the auto focus is problematic for the examples provided within the guidance – it causes the browser to scroll down to the first example on the page. At the minute, we are working around this by ['monkey-patching' the init method to remove the call to `$module.focus`][1].

[1]: https://github.com/alphagov/govuk-design-system/pull/643

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ This change was introduced in [#2491: Prevent error summary from being re-focuse
 
 - [#2475: Tweak whitespace in input component HTML for improved readability](https://github.com/alphagov/govuk-frontend/pull/2475)
 - [#2491: Prevent error summary from being re-focused after it has been initially focused on page load](https://github.com/alphagov/govuk-frontend/pull/2491)
+- [#2494: Allow disabling autofocus on error summary](https://github.com/alphagov/govuk-frontend/pull/2494)
 
 ## 4.0.0 (Breaking release)
 
--- a/src/govuk/components/error-summary/error-summary.js
+++ b/src/govuk/components/error-summary/error-summary.js
@@ -22,6 +22,10 @@ ErrorSummary.prototype.init = function () {
 ErrorSummary.prototype.setFocus = function () {
   var $module = this.$module
 
+  if ($module.getAttribute('data-disable-auto-focus') === 'true') {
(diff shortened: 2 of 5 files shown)


36degrees:
This PR builds on the changes in #2491, which needs to be merged first.

lfdebrux:
Does this need a changelog entry?

36degrees:
Give me a chance – don't know the PR number until I've opened the PR 😉

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ This change was introduced in [#2491: Prevent error summary from being re-focuse
 
 - [#2475: Tweak whitespace in input component HTML for improved readability](https://github.com/alphagov/govuk-frontend/pull/2475)
 - [#2491: Prevent error summary from being re-focused after it has been initially focused on page load](https://github.com/alphagov/govuk-frontend/pull/2491)
+- [#2494: Allow disabling autofocus on error summary](https://github.com/alphagov/govuk-frontend/pull/2494)
```
vanitabarrett, line 17: Is this a fix or a new feature? 🤔

vanitabarrett (approved):

CHANGELOG.md
vanitabarrett: Ah sorry, I hadn't spotted that!

Ollie and I had a chat about this in our catch-up. We discussed it being a bit of an odd case where it's technically a new feature, but we're almost only adding it to fix our situation in the Design System. It may be beneficial to services (perhaps if they're showing/hiding the error summary on the client side) but we're not sure of any use cases like that for certain.

I'm happy to approve this as it is (listed as a fix). We did discuss the possibility of listing it as a feature if there are other new features in this release, but it seems weird to highlight this in a feature release if it's the only feature we're shipping but we don't recommend using it. We're not sure what else will be included in this release yet. but we should review the release notes when we are ready to release, so we'll have an opportunity to change this then.

PR #2515, opened 2022-01-25
Add explicit width to summary list row with no actions pseudoelement

### LINKED ISSUE #2495: Incorrect content width on summary list rows when not all rows have actions
## Description of the issue

When using the summary-list nunjucks macro, when some rows have actions and some do not, the width given to the values in the row becomes too thin. This is most notable when using long content.

If all rows have actions the content has the correct width.

This commit could be the culprit: https://github.com/alphagov/govuk-frontend/pull/2323/commits/29358635a77efa5a81b923ad644e42bc8481ce45

| v3.14 | v4.0 | v4.0 with just one row |
|--|--|--|
|v3 14 0 | v4 0 0 | v4 0 0-no-other-rows

| v4 | If I add width 50% to the container |
|--|--|
| Screenshot 2022-01-12 at 16 05 13 | Screenshot 2022-01-12 at 16 05 07 |

### Design system examples

This bug is visible on the design system examples too (https://design-system.service.gov.uk/components/summary-list/#adding-actions-to-each-row) – look at the two contact details rows:

Screenshot 2022-01-12 at 16 18 57

See design system example

Expected content to remain the same width

- Operating system: Mac
- Browser: Chrome
- Browser version: 96
- GOV.UK Frontend Version: 4.0.0

domoscargin:
When a summary list row has no actions, it gets an `after` pseudoelement that adds an empty table cell, so that row borders are drawn correctly.

If the first row has no actions, it spaces itself automatically unless a width is specified for the pseudoelement. This then cascades to the rest of the summary list, and the actions column on every row takes up too much space, clipping content in other columns.

This can be solved by explicitly setting a width on the pseudoelement.

Additionally, we don't need this pseudoelement on widths below tablet, as the summary list cells split onto separate rows.

Closes #2495

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ This change was introduced in [#2491: Prevent error summary from being re-focuse
 - [#2475: Tweak whitespace in input component HTML for improved readability](https://github.com/alphagov/govuk-frontend/pull/2475)
 - [#2491: Prevent error summary from being re-focused after it has been initially focused on page load](https://github.com/alphagov/govuk-frontend/pull/2491)
 - [#2494: Allow disabling autofocus on error summary](https://github.com/alphagov/govuk-frontend/pull/2494)
+- [#2515: Add explicit width to summary list row with no actions pseudoelement](https://github.com/alphagov/govuk-frontend/pull/2515)
 
 ## 4.0.0 (Breaking release)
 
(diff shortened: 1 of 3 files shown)


fofr:
Thank you @domoscargin 🎉

36degrees:
Needs a changelog entry, but otherwise this is looking good to me 👍🏻

domoscargin:
> Needs a changelog entry, but otherwise this is looking good to me 👍🏻
Done!

36degrees (approved):

PR #2516, opened 2022-01-26
Removed unreachable code

JoPintoPaul:
The setting of `isConditional` was removed in an earlier commit, so this class will never be set.


domoscargin (approved):
Thanks @JoPintoPaul Nice catch.

I'm just going to wait for somebody else on my team to approve before merging.

owenatgov (approved):
Looks good to me 👍🏻  Good spot @JoPintoPaul

Issue #2435, opened 2021-11-17
Select elements have grey background in Firefox


matteason:
## Description of the issue
In the latest Firefox release (94.0.1), `select` elements have a grey background by default. This makes them look disabled:

![image](https://user-images.githubusercontent.com/1935173/142191720-4fc81853-9000-454f-bf41-5fb8105096bb.png)

This also affects the main GOV.UK site:

![image](https://user-images.githubusercontent.com/1935173/142191617-ce27b6aa-426f-4ccc-ac6c-0d488c1d4b89.png)

I think strictly this is a Firefox issue, but from https://bugzilla.mozilla.org/show_bug.cgi?id=1715064 it sounds like they consider it an expected change.

There was a similar issue with Ubuntu a couple of years ago (https://github.com/alphagov/govuk-frontend/issues/1659). The issue was closed off because setting a background colour on `select` caused issues in Firefox if the user has changed their browser default colours, but that issue now [appears to be fixed](https://bugzilla.mozilla.org/show_bug.cgi?id=1335476#c18)

## Steps to reproduce the issue
1. Download the latest Firefox release
2. Visit https://design-system.service.gov.uk/components/select/

## Actual vs expected behaviour
I would expect the background of the select to match the background colour of the page

## Environment (where applicable)

- Operating system: Windows 10 and macOS
- Browser: Firefox
- Browser version: 94.0.1
- GOV.UK Frontend Version: 3.14.0

matteason:
Sorry, for some reason GitHub isn't rendering the images in my initial post, but it will in comments 🤷‍♀️

Image 1:
![image](https://user-images.githubusercontent.com/1935173/142191720-4fc81853-9000-454f-bf41-5fb8105096bb.png)

Image 2:
![image](https://user-images.githubusercontent.com/1935173/142191617-ce27b6aa-426f-4ccc-ac6c-0d488c1d4b89.png)

vanitabarrett:
A bit of testing to confirm when this was introduced:

Windows 10, Firefox 94: Grey background
Windows 10, Firefox 95 (beta): Grey background
Windows 10, Firefox 93: White background

Mac, Firefox 94: Grey background
Mac, Firefox 95 (beta): Grey background
Mac, Firefox 93: White background

matteason:
Thanks @vanitabarrett - I was just about to say, I've reinstalled Firefox 93 on Win10 and the `select` background is white

vanitabarrett:
[...]

PR #2524, opened 2022-01-31
Fix select component renders with grey background in Firefox and Safari

vanitabarrett:
Fixes https://github.com/alphagov/govuk-frontend/issues/2435

## What
Sets background colour and text colour on the select component so that we have more control over the background and foreground colour and don't get influenced by browser defaults which sometimes default to a grey background.

## Why
We have had two separate reports from users of the select component appearing with a grey background in Firefox and iOS Safari. In iOS Safari, the text within the select also appears as blue - the blue text on grey background does not meet colour contrast (ratio 2.19:1). We should probably raise a WebKit bug for this too, but alongside the issue in Firefox it probably also makes sense for us to look at getting more control over our component styling too.

## Screenshots (default component)

Chrome:
Screenshot 2022-01-31 at 15 25 18
Screenshot 2022-01-31 at 15 25 25

iOS:
Screenshot 2022-01-31 at 15 33 37

Firefox:
Screenshot 2022-01-31 at 15 45 55
Screenshot 2022-01-31 at 15 45 52

IE11:
Screenshot 2022-01-31 at 15 49 24
Screenshot 2022-01-31 at 15 49 28

## Screenshots (colours changed)

Windows High Contrast, Edge:
Screenshot 2022-01-31 at 15 17 29

Windows High Contrast, Chrome:
Screenshot 2022-01-31 at 15 18 55

Windows High Contrast, Firefox:
Screenshot 2022-01-31 at 15 19 42

Windows High Contrast, IE11:
Screenshot 2022-01-31 at 15 20 47
Screenshot 2022-01-31 at 15 20 31

Firefox (changed colours):
Screenshot 2022-01-31 at 15 46 22
Screenshot 2022-01-31 at 15 46 25

### DIFF
--- a/src/govuk/components/select/_index.scss
+++ b/src/govuk/components/select/_index.scss
@@ -15,6 +15,11 @@
     padding: govuk-spacing(1); // was 5px 4px 4px - size of it should be adjusted to match other form elements
     border: $govuk-border-width-form-element solid $govuk-input-border-colour;
 
+    // Default user agent colours for selects can have low contrast,
+    // and may look disabled (#2435)
+    color: $govuk-text-colour;
+    background-color: govuk-colour("white");
+
     &:focus {
       outline: $govuk-focus-width solid $govuk-focus-colour;
       // Ensure outline appears outside of the element
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ This change was introduced in [#2491: Prevent error summary from being re-focuse
 - [#2494: Allow disabling autofocus on error summary](https://github.com/alphagov/govuk-frontend/pull/2494)
 - [#2515: Add explicit width to summary list row with no actions pseudoelement](https://github.com/alphagov/govuk-frontend/pull/2515)
 - [#2514: Fix accordion heading style while JavaScript is disabled](https://github.com/alphagov/govuk-frontend/pull/2514)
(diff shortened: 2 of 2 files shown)


src/govuk/components/select/_index.scss
```diff
@@ -15,6 +15,11 @@
     padding: govuk-spacing(1); // was 5px 4px 4px - size of it should be adjusted to match other form elements
     border: $govuk-border-width-form-element solid $govuk-input-border-colour;
 
+    // Sets the background and foreground colours
```
36degrees, line 18: Minor point, but given this is a relatively 'expected' thing to do, and we have the Git history, I do wonder how useful the comment is.

If we do want to include a comment, it might be worth including a summary of the issue we're trying to fix, rather than requiring the reader to go to the GitHub issue for the context?

Perhaps something like:

```scss
    // Default user agent colours for selects can have low contrast, and may
    // look disabled (#2435)
```

36degrees (approved):
Makes sense to me, and a big improvement in Firefox and iOS – although I haven't tested in any other browsers.

My only comment is a minor one, about the comment!

src/govuk/components/select/_index.scss
vanitabarrett: I think I just wanted a comment in there because otherwise it felt easy for someone to remove (for example) the text color without realising why it was being set. But that's a good point, I've added your suggestion 👍

PR #2545, opened 2022-02-14
Add exit this page component

owenatgov:
## What/Why
Adds the Exit This Page component to GOV.UK Frontend. Details about this component including functionality, history of work and decisions can be found in the epic: https://github.com/alphagov/govuk-design-system/issues/1755

### Standard view
Large screen:
Screenshot 2023-03-28 at 13 59 20

Small screen:
Screenshot 2023-03-28 at 13 59 44

### With indicator light active
Large screen:
Screenshot 2023-03-28 at 14 00 02

Small screen:
Screenshot 2023-03-28 at 14 00 40

### Forced colours mode (Windows High Contrast)
Large screen:
Screenshot 2023-03-28 at 14 02 58

Small screen:
Screenshot 2023-03-28 at 14 03 36

### Forced colours with indicator light active
Large screen:
Screenshot 2023-03-28 at 14 05 04

Small screen:
Screenshot 2023-03-28 at 14 04 29

## Testing
This component has been tested extensively via the following means:

- user research with users with experience of domestic abuse
- user research with users with assistive tech needs
- auditing with disabled users via [DAC](https://digitalaccessibilitycentre.org/)

## Notes
This component has been built across the following pull requests:
- https://github.com/alphagov/govuk-frontend/pull/2942
- https://github.com/alphagov/govuk-frontend/pull/2943
- https://github.com/alphagov/govuk-frontend/pull/2940
- https://github.com/alphagov/govuk-frontend/pull/2949
- https://github.com/alphagov/govuk-frontend/pull/3059
- https://github.com/alphagov/govuk-frontend/pull/3081
- https://github.com/alphagov/govuk-frontend/pull/3176
- https://github.com/alphagov/govuk-frontend/pull/3209
- https://github.com/alphagov/govuk-frontend/pull/3210
- https://github.com/alphagov/govuk-frontend/pull/3247
- https://github.com/alphagov/govuk-frontend/pull/3257
- https://github.com/alphagov/govuk-frontend/pull/3221
- https://github.com/alphagov/govuk-frontend/pull/3353
- https://github.com/alphagov/govuk-frontend/pull/3305

### DIFF
--- a/app/src/views/examples/exit-this-page-with-skiplink/index.njk
+++ b/app/src/views/examples/exit-this-page-with-skiplink/index.njk
@@ -1,0 +1,19 @@
+{% extends "layout.njk" %}
+
+{% from "exit-this-page/macro.njk" import govukExitThisPage %}
+{% from "skip-link/macro.njk" import govukSkipLink %}
+
+{% block skipLink %}
+  {{ super() }}
+  {{ govukSkipLink({
+    href: "https://www.gov.uk/",
+    text: "Exit this page",
+    classes: "govuk-js-exit-this-page-skiplink"
+  }) }}
+{% endblock %}
+
+{% block content %}
+  {{ govukExitThisPage({
+    redirectUrl: "https://www.gov.uk/"
+  }) }}
+{% endblock %}
--- a/app/src/views/full-page-examples/child-maintenance/index.njk
+++ b/app/src/views/full-page-examples/child-maintenance/index.njk
@@ -1,0 +1,166 @@
+---
+scenario: >-
+  You are a user looking for information on child maintenance.
+  There is a chance that as a user of this service, you are a victim of domestic
+  abuse, therefore this page includes an Exit this Page component.
+
+  Things to try:
+
+  1. Activate the Exit this Page button by clicking it
+
+  2. Activate the Exit this Page button by pressing the Esc key 3 times
+
+  3. Activating the Exit this Page button on a small screen view
+
+  4. Scoll down the page to see if the Exit this Page button remains sticky
+
+  5. Tabbing through the page to find the Exit this Page "skip link" at the top
+  of the accessibility tree
+---
+
+{% extends "full-page-example.njk" %}
+
+{% from "breadcrumbs/macro.njk" import govukBreadcrumbs %}
+{% from "button/macro.njk" import govukButton %}
+{% from "details/macro.njk" import govukDetails %}
+{% from "exit-this-page/macro.njk" import govukExitThisPage %}
+{% from "inset-text/macro.njk" import govukInsetText %}
+{% from "phase-banner/macro.njk" import govukPhaseBanner %}
+
+{% set pageTitle = "Child maintenance" %}
+{% block pageTitle %}{{ pageTitle }} - GOV.UK{% endblock %}
+
+{% block skipLink %}
+  {{ super() }}
+  {{ govukSkipLink({
+    href: "https://bbc.co.uk/weather/",
+    classes: "govuk-js-exit-this-page-skiplink"
+  }) }}
+{% endblock %}
+
+{% block beforeContent %}
+  {{ govukPhaseBanner({
+    tag: {
+      text: "beta"
+    },
+    html: "This is a new service – your <a class="govuk-link" href="#">feedback</a> will help us to improve it."
+  }) }}
+
+  {{ govukBreadcrumbs({
+    classes: "",
+    collapseOnMobile: true,
+    items: [
+      {
+        html: "Home",
+        href: "/"
+      },
+      {
+        html: "Births, deaths, marriages and care",
+        href: "/"
+      },
+      {
+        html: "Having a child, parenting and adoption",
+        href: "/"
+      }
+    ]
+  }) }}
+{% endblock %}
+
+{% block content %}
+  {{ govukExitThisPage({
+    text: "Exit this page",
+    redirectUrl: "https://bbc.co.uk/weather/"
+  }) }}
+
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <h1 class="govuk-heading-xl">Get help arranging child maintenance</h1>
+
+      <p class="govuk-body-m">This service will give you information about the different options available to make a child maintenance arrangement. These are:</p>
+
+      <ul class="govuk-list govuk-list--bullet">
+        <li>making your own arrangement</li>
+        <li>using the Child Maintenance Service</li>
+      </ul>
+
+      <p class="govuk-body-m">
+        <a class="govuk-link" href="contact-child-maintenance-choices">Contact Child Maintenance Choices if you live in Northern Ireland</a>
+      </p>
+
+      <h2 class="govuk-heading-m">Using the Child Maintenance Service</h2>
+      <h3 class="govuk-heading-s">If you already have an active case</h3>
+      <p class="govuk-body-m">If you have a case with the same parent you should <a href="https://www.gov.uk/child-maintenance-service/change-of-circumstances" class="govuk-link"> report a change of circumstances </a>instead of making a new arrangement. You can use this service if you want to set up a new arrangement with a different parent.</p>
+
+      <h3 class="govuk-heading-s">Fees</h3>
+      <p class="govuk-body-m">If you decide to use the Child Maintenance Service the application fee is £20 and cannot be refunded. You’ll need a debit or credit card to pay this. It does not guarantee an arrangement.</p>
+      <p class="govuk-body-m">You will not have to pay the application fee if you:</p>
+      <ul class="govuk-list govuk-list--bullet">
+        <li>are under 19 years old</li>
+        <li>have experienced domestic abuse</li>
+      </ul>
+      <p class="govuk-body-m">
+        <a class="govuk-link" href="https://www.gov.uk/guidance/domestic-abuse-how-to-get-help">Get help if you have experienced domestic abuse</a>
+      </p>
+
+      <h3 class="govuk-heading-s">If you’re worried about contacting the other parent</h3>
+      <p class="govuk-body-m">If you decide to use the Child Maintenance Service, we will try to contact the other parent about setting up child maintenance – you do not need to.</p>
+      <p class="govuk-body-m">We will share the following details with the other parent:</p>
+      <ul class="govuk-list govuk-list--bullet">
+        <li>your name</li>
+        <li>your child’s name and date of birth</li>
+        <li>the number of days your child stays with the parent who will be making payments - the other parent will need to confirm this</li>
+      </ul>
+
+      <p class="govuk-body-m">If you are the parent who will be paying child maintenance, we will also share the following details with the other parent:</p>
+      <ul class="govuk-list govuk-list--bullet">
+        <li>your income information</li>
+        <li>whether any additional children are taken into account in the calculation - we will not share identifying details about these children, only how many have been taken into account</li>
+      </ul>
+
+      {{ govukInsetText({
+        text: "We will not share any other information with the other parent as a part of your application."
+      }) }}
+
+      {% set doIHaveAnActiveCase %}
+        You will need to <a href="/get-help-arranging-child-maintenance/contact-child-maintenance-service" target="_blank" class="govuk-link">contact child maintenance (opens in new tab)</a>.
+      {% endset %}
+
+      {% set payingToReceiving %}
+        <p class="govuk-body">You will need to close your existing Child Maintenance case and open a new one.</p>
+        <p class="govuk-body">If you were the original applicant, you can close your case in <a href="https://childmaintenanceservice.direct.gov.uk/onlinerevive/public/landing-screen" target="blank" class="govuk-link">MCMC</a> , or you can <a href="/get-help-arranging-child-maintenance/contact-child-maintenance-service" target="blank" class="govuk-link">contact child maintenance (opens in new tab)</a>.</p>
+        <p class="govuk-body">If you were not the original applicant, you will need to <a href="/get-help-arranging-child-maintenance/contact-child-maintenance-service" target="_blank" class="govuk-link">contact child maintenance (opens in new tab)</a>.</p>
+      {% endset %}
+
+      {% set receivingToPaying %}
+        <p class="govuk-body">You will need to close your existing Child Maintenance case and open a new one.</p>
+        <p class="govuk-body">If you were the original applicant, you can close your case in <a href="https://childmaintenanceservice.direct.gov.uk/onlinerevive/public/landing-screen" target="blank" class="govuk-link">MCMC</a> , or you can <a href="/get-help-arranging-child-maintenance/contact-child-maintenance-service" target="blank" class="govuk-link">contact child maintenance (opens in new tab)</a>.</p>
+        <p class="govuk-body">If you were not the original applicant, you will need to <a href="/get-help-arranging-child-maintenance/contact-child-maintenance-service" target="_blank" class="govuk-link">contact child maintenance (opens in new tab)</a>.</p>
+      {% endset %}
+
+      {{ govukDetails({
+        summaryText: "I do not know if I have an active case",
+        html: doIHaveAnActiveCase
+      }) }}
+
+      {{ govukDetails({
+        summaryText: "I need to change from paying to receiving child maintenance",
+        html: payingToReceiving
+      }) }}
+
+      {{ govukDetails({
+        summaryText: "I need to change from receiving to paying child maintenance",
+        html: receivingToPaying
+      }) }}
+
+      {{ govukButton({
+        text: "Start now",
+        href: "#",
+        isStartButton: true
+      }) }}
+    </div>
+    <div class="govuk-grid-column-one-third">
+      <p class="govuk-body">There could be other content here.</p>
+      <p class="govuk-body">In ea ex nulla veniam commodo Lorem labore dolor et ea in deserunt qui. Enim adipisicing dolore est amet magna. Cupidatat adipisicing labore consectetur exercitation Lorem dolor ipsum et deserunt est. Sit laborum dolore sit id eiusmod.</p>
+    </div>
+  </div>
+{% endblock %}
--- a/src/govuk/all.mjs
+++ b/src/govuk/all.mjs
@@ -6,6 +6,7 @@ import CharacterCount from './components/character-count/character-count.mjs'
 import Checkboxes from './components/checkboxes/checkboxes.mjs'
 import Details from './components/details/details.mjs'
 import ErrorSummary from './components/error-summary/error-summary.mjs'
+import ExitThisPage from './components/exit-this-page/exit-this-page.mjs'
 import Header from './components/header/header.mjs'
 import NotificationBanner from './components/notification-banner/notification-banner.mjs'
 import Radios from './components/radios/radios.mjs'
@@ -58,6 +59,11 @@ function initAll (config) {
     new ErrorSummary($errorSummary, config.errorSummary).init()
   }
 
+  var $exitThisPageButtons = $scope.querySelectorAll('[data-module="govuk-exit-this-page"]')
+  nodeListForEach($exitThisPageButtons, function ($button) {
+    new ExitThisPage($button, config.exitThisPage).init()
+  })
+
   // Find first header module to enhance.
   var $header = $scope.querySelector('[data-module="govuk-header"]')
   if ($header) {
@@ -97,6 +103,7 @@ export {
   CharacterCount,
   Checkboxes,
   ErrorSummary,
+  ExitThisPage,
   Header,
   NotificationBanner,
   Radios,
@@ -113,5 +120,6 @@ export {
  * @property {import('./components/button/button.mjs').ButtonConfig} [button] - Button config
  * @property {import('./components/character-count/character-count.mjs').CharacterCountConfig} [characterCount] - Character Count config
  * @property {import('./components/error-summary/error-summary.mjs').ErrorSummaryConfig} [errorSummary] - Error Summary config
+ * @property {import('./components/exit-this-page/exit-this-page.mjs').ExitThisPageConfig} [exitThisPage] - Exit This Page config
  * @property {import('./components/notification-banner/notification-banner.mjs').NotificationBannerConfig} [notificationBanner] - Notification Banner config
  */
--- a/src/govuk/components/exit-this-page/exit-this-page.mjs
+++ b/src/govuk/components/exit-this-page/exit-this-page.mjs
@@ -1,0 +1,405 @@
+/* eslint-disable es-x/no-function-prototype-bind -- Polyfill imported */
+
+import { normaliseDataset } from '../../common/normalise-dataset.mjs'
+import { nodeListForEach, mergeConfigs, extractConfigByNamespace } from '../../common.mjs'
+import { I18n } from '../../i18n.mjs'
+import '../../vendor/polyfills/Element/prototype/classList.mjs'
+import '../../vendor/polyfills/Event.mjs' // addEventListener, event.target normalization and DOMContentLoaded
+import '../../vendor/polyfills/Element/prototype/dataset.mjs'
+import '../../vendor/polyfills/Function/prototype/bind.mjs'
+
+/**
+ * @constant
+ * @type {ExitThisPageTranslations}
+ * @see Default value for {@link ExitThisPageConfig.i18n}
+ * @default
+ */
+var EXIT_THIS_PAGE_TRANSLATIONS = {
+  activated: 'Loading.',
+  timedOut: 'Exit this page expired.',
+  pressTwoMoreTimes: 'Shift, press 2 more times to exit.',
+  pressOneMoreTime: 'Shift, press 1 more time to exit.'
+}
+
+/**
+ * Exit This Page component
+ *
+ * @class
+ * @param {HTMLElement} $module - HTML element that wraps the Exit This Page button
+ * @param {ExitThisPageConfig} [config] - Exit This Page config
+ */
+function ExitThisPage ($module, config) {
+  /** @type {ExitThisPageConfig} */
+  var defaultConfig = {
+    i18n: EXIT_THIS_PAGE_TRANSLATIONS
+  }
+
+  if (!($module instanceof HTMLElement)) {
+    return this
+  }
+
+  var $button = $module.querySelector('.govuk-exit-this-page__button')
+  if (!($button instanceof HTMLElement)) {
+    return this
+  }
+
+  /**
+   * @deprecated Will be made private in v5.0
+   * @type {ExitThisPageConfig}
+   */
+  this.config = mergeConfigs(
+    defaultConfig,
+    config || {},
+    normaliseDataset($module.dataset)
+  )
+
+  this.i18n = new I18n(extractConfigByNamespace(this.config, 'i18n'))
+
+  /** @deprecated Will be made private in v5.0 */
+  this.$module = $module
+
+  /** @deprecated Will be made private in v5.0 */
+  this.$button = $button
+
+  /** @deprecated Will be made private in v5.0 */
+  this.$skiplinkButton = document.querySelector('.govuk-js-exit-this-page-skiplink')
+
+  /** @deprecated Will be made private in v5.0 */
+  this.$updateSpan = null
+
+  /** @deprecated Will be made private in v5.0 */
+  this.$indicatorContainer = null
+
+  /** @deprecated Will be made private in v5.0 */
+  this.$overlay = null
+
+  /** @deprecated Will be made private in v5.0 */
+  this.keypressCounter = 0
+
+  /** @deprecated Will be made private in v5.0 */
+  this.lastKeyWasModified = false
+
+  /** @deprecated Will be made private in v5.0 */
+  this.timeoutTime = 5000 // milliseconds
+
+  // Store the timeout events so that we can clear them to avoid user keypresses overlapping
+  // setTimeout returns an id that we can use to clear it with clearTimeout,
+  // hence the 'Id' suffix
+
+  /** @deprecated Will be made private in v5.0 */
+  this.keypressTimeoutId = null
+
+  /** @deprecated Will be made private in v5.0 */
+  this.timeoutMessageId = null
+}
+
+/**
+ * Create the <span> we use for screen reader announcements.
+ *
+ * @deprecated Will be made private in v5.0
+ */
+ExitThisPage.prototype.initUpdateSpan = function () {
+  this.$updateSpan = document.createElement('span')
+  this.$updateSpan.setAttribute('role', 'status')
+  this.$updateSpan.className = 'govuk-visually-hidden'
+
+  this.$module.appendChild(this.$updateSpan)
+}
+
+/**
+ * Create button click handlers.
+ *
+ * @deprecated Will be made private in v5.0
+ */
+ExitThisPage.prototype.initButtonClickHandler = function () {
+  // Main EtP button
+  this.$button.addEventListener('click', this.handleClick.bind(this))
+
+  // EtP skiplink
+  if (this.$skiplinkButton) {
+    this.$skiplinkButton.addEventListener('click', this.handleClick.bind(this))
+  }
+}
+
+/**
+ * Create the HTML for the 'three lights' indicator on the button.
+ *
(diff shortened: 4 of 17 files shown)


src/govuk/components/hide-this-page/README.md
```diff
@@ -0,0 +1,15 @@
+# Hide this page
```
vanitabarrett, line 1: Sorry to be picky as I know this isn't ready for production yet, but I've noticed in the past that it's really easy to forget to update this file to talk about the correct component (it's normally copied + pasted) and so sometimes we notice it a way down the line. So just an early heads up that this file still references the inset text component.

src/govuk/components/hide-this-page/_index.scss
```diff
@@ -0,0 +1,31 @@
...
+      width: auto;
+      float: right;
+    }
+  }
+
+  .govuk-hide-this-page__button {
+    margin-bottom: 0;
+  }
+
+  @media only print {
```
vanitabarrett, line 26: Nice work on this and thinking about print styles 👏

src/govuk/components/hide-this-page/hide-this-page.yaml
```diff
@@ -0,0 +1,54 @@
...
+  required: false
+  description: HTML attributes (for example data attributes) to add to the hide this page container.
+
+examples:
+- name: default  
+  data:
+    text: Hide this page
+    currentTabURL: "https://www.google.com"
+    newTabURL: "https://www.bbc.co.uk/weather"
+    fakePageTitle: How to prevent the spread of Coronavirus - GOV.UK
```
vanitabarrett, line 37: I'd suggest the default  example should only contain any required fields (so in this case, maybe nothing?)

src/govuk/all.js
```diff
@@ -52,6 +53,11 @@ function initAll (options) {
   var $toggleButton = scope.querySelector('[data-module="govuk-header"]')
   new Header($toggleButton).init()
 
+  var $hideThisPageButtons = scope.querySelectorAll('[data-module="govuk-hide-this-page"]')
+  nodeListForEach($hideThisPageButtons, function ($hideThisPageButtons) {
+    new HideThisPage($hideThisPageButtons).init()
+  })
+
```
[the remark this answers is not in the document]
owenatgov: In the current implementation we're planning on trying out, there would technically be 2 on a page. See the draft branch on the design system website: https://github.com/alphagov/govuk-design-system/tree/hide-this-page-component I agree that if we change it to be only one, we would only look for and initialise that 1 button.

src/govuk/components/hide-this-page/hide-this-page.yaml
owenatgov: I need to work out how best to handle this. Right now if I pass nothing, literally nothing gets applied ie: you get a button with not text in. We should probably set default values for this in the component itself but I've seen different implementations across the repo.

vanitabarrett: Agreed, I'm not sure if I've seen one implementation we always use for this. Potentially it depends on how complex the logic is. I think I've seen some where we do `if params.text else 'default' }}`, but there are instances where we pull out the logic to the top of the template, e.g:

```
{%- if params.text %}
    {% set buttonText = params.text %}
{%- else %}
    {% set buttonText = "default" %}
{%- endif -%}
```

Potentially there's an argument for setting defaults at the top to avoid cluttering the actual HTML code with that logic and hopefully making it easier to read.

src/govuk/components/hide-this-page/hide-this-page.js
```diff
@@ -0,0 +1,44 @@
...
+
+HideThisPage.prototype.init = function () {
+  // We put the loop here instead of in all.js because we want to have both listeners on the individual buttons for clicks
+  // and a single listener for the keyboard shortcut
+  nodeListForEach(this.$module, function ($button) {
+    $button.querySelector('.govuk-js-hide-this-page-button').addEventListener('click', this.openNewTab)
+  }.bind(this))
+
+  window.addEventListener('keydown', function (e) {
+    if (e.key === 'Escape') {
```
querkmachine, line 38: [KeyboardEvent.key isn't properly supported in Internet Explorer.](https://caniuse.com/keyboardevent-key) and will need supplementing with KeyboardEvent.keyCode and/or KeyboardEvent.which. 

Something like `e.key === 'Escape' || e.keyCode === 27 || e.which === 27` outta work, I think?

src/govuk/components/hide-this-page/_index.scss
```diff
@@ -0,0 +1,31 @@
...
+  .govuk-hide-this-page {
+    position: fixed;
+    z-index: 1000;
+    top: 0;
+    left: 0;
+    width: 100%;
+
+    @include govuk-media-query($from: tablet) {
+      display: inline-block;
+      position: sticky;
```
querkmachine, line 13: Does this need to be progressively enhanced for browsers without support for `position: sticky`?

src/govuk/components/hide-this-page/hide-this-page.js
```diff
@@ -0,0 +1,44 @@
...
+}
+
+HideThisPage.prototype.init = function () {
+  // We put the loop here instead of in all.js because we want to have both listeners on the individual buttons for clicks
+  // and a single listener for the keyboard shortcut
+  nodeListForEach(this.$module, function ($button) {
+    $button.querySelector('.govuk-js-hide-this-page-button').addEventListener('click', this.openNewTab)
+  }.bind(this))
+
+  window.addEventListener('keydown', function (e) {
```
querkmachine, line 37: There are potentially some edge cases where we don't want this escape key functionality to fire, in particular, with popover elements, such as those that appear when certain form elements are in use—datepickers, select lists and the like. These could lead to users inadvertently exiting a form and losing data or progress. 

I did a quick whip around and, luckily, browser vendors seem to mostly handle that for us. (✅ = event did not propagate to `window`; ❌ = event _did_ propagate to `window`). 

|Browser|`input[type="date"]` datepicker|`select` menu|
|:-|:-|:-|
|Chrome 99|✅|✅|
|Edge 99|✅|✅|
|Opera 84|✅|✅|
|Firefox 98|✅|✅|
|Safari 15.4 (macOS)|❌|✅|
|Safari 15.4 (iOS/iPadOS)|✅|✅|
|IE 11|N/A|✅|

It's also notable that both the accessible-autocomplete and the country-and-territory-autocomplete packages currently use the Escape key to close an open popover menu, and neither prevent propagation of that event to the `window` at this time. 

Given we don't encourage the use of `input[type="date"]` anyway, the Safari issue is perhaps not the highest priority at this time. I don't know if we want to simply document these situations or attempt to resolve these in code.

src/govuk/components/hide-this-page/_index.scss
owenatgov: It already is! (sort of). Currently on browser not supporting `sticky` it falls back to `fixed` found on line 5. There's a question about the tradeoff of having a button that follows the user but is outside the bounds of the width container and a button that doesn't follow the user but is easier to notice and access visually when it is on the screen. I'll revisit this with a designer once this has gone through some research.

src/govuk/components/hide-this-page/hide-this-page.js
owenatgov: This is incredibly thorough! Many thanks, I'll link this to the [keyboard shortcut issue](https://github.com/alphagov/govuk-design-system/issues/2116).

We're very aware of the risk of using `esc` and how it may conflict with a number of existing functionalities, however we're implementing it for now for the purpose of having an established convention for this pattern so we an test if keyboard shortcuts are actually useful or not. See [my comment on the keyboard issue](https://github.com/alphagov/govuk-design-system/issues/2116#issuecomment-1077837908) for further details.

querkmachine:
(Realise this is an old prototype, but wanted to get this down for future reference.) 

It appears that the functionality tied to pressing `Esc` three times is blocked by Firefox's default popup blocking functionality, at least as of Firefox 105. Instead a banner is shown across the top of the viewport, indicating that the action has been blocked by the browser. No new tab is opened, nor is the current tab directed away from.

This seems to be tied to the new tab functionality. The same issue is seemingly not present on [The Trevor Project website](https://www.thetrevorproject.org/), which uses the same method. The banner warning that a new tab was blocked from opening still appears, however the current tab is then successfully navigated away from. Chromium browsers open the new tab without complaint.

Ensuring that any escape functionality works as expected with default browser settings is probably something we need to keep in mind.

owenatgov:
@querkmachine I might have some context on this. Something we found when we were building this is that browser popup functionality to block what it asserts is malicious code includes blocking any attempts to at least open new views from a page. What turns this off is the user interacting with the page in some way ie: tapping, clicking, scrolling etc. This seems to not include keyboard interaction, meaning that if a user hasn't interacted with a page but they try to use the esc key it'll popup block. I can't tell if this is a bug or not and didn't investigate this during the prototype building.

I'm coming into this making some assumptions about what I encountered when I was last looking at this so this may not be relevant. I agree that let's keep an eye on this and see if it's still an issue when we remove the more complex tab functionality.

src/govuk/components/hide-this-page/hide-this-page.mjs
```diff
@@ -0,0 +1,65 @@
...
+      this.escTimerActive = false
+      this.updateSpan.innerText = ''
+    }.bind(this), 2000)
+  }
+}
+
+HideThisPage.prototype.init = function () {
+  this.initUpdateSpan()
+  this.initButtonClickHandler()
+  window.addEventListener('keydown', this.handleEscKeypress.bind(this))
```
colinrotherham, line 62: Nice work with all this 😊 

Just a tiny tiny tiny implementation detail

Default phase for listeners is to grab events "bubbling up" (like we have here)
https://www.w3.org/TR/DOM-Level-3-Events/#event-flow

1. Capturing phase – the event goes down to the element.
2. Target phase – the event reached the target element.
3. Bubbling phase – the event bubbles up from the element.

But we're listening at the top on `window`

The `(e)` event might have already "bubbled up" through hundreds of DOM elements—potentially being cancelled/stopped by other listeners via [`e.preventDefault()`](https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault) or [`e.stopPropagation()`](https://developer.mozilla.org/en-US/docs/Web/API/Event/stopPropagation) along the way

I'd recommend changing to the `useCapture` capturing phase (still on `window`) to ensure we grab the event on its way down, before other listeners get chance to cancel it once it bubbles back up

colinrotherham: The [GOV.UK Publishing Components **Modal dialogue**](https://github.com/alphagov/govuk_publishing_components/blob/main/app/assets/javascripts/govuk_publishing_components/components/modal-dialogue.js#L57) (good example) also listens for the escape key, but you can see they add the `useCapture` flag

They're good citizens and don't stop the event, but other listeners might do

owenatgov: Aha I missed this, why did I set this to listen on window 🤔

src/govuk/components/exit-this-page/exit-this-page.test.js
```diff
@@ -0,0 +1,169 @@
...
+    })
+
+    afterAll(async () => {
+      await page.setJavaScriptEnabled(true)
+    })
+
+    it('navigates to the href of the button', async () => {
+      await goToComponent(page, 'exit-this-page')
+
+      const href = await page.evaluate((buttonClass) => document.querySelector(buttonClass).href, buttonClass)
```
36degrees, line 24: Suggestion: Puppeteer has a handy [`$eval` function](https://pptr.dev/api/puppeteer.page._eval) that could be used to simplify this (and similar statements throughout these tests) slightly:

```suggestion
      const href = await page.$eval(buttonClass, el => el.href)
```

src/govuk/components/exit-this-page/exit-this-page.mjs
```diff
@@ -0,0 +1,343 @@
...
+
+/**
+ * @constant
+ * @type {ExitThisPageTranslations}
+ * @see Default value for {@link ExitThisPageConfig.i18n}
+ * @default
+ */
+var EXIT_THIS_PAGE_TRANSLATIONS = {
+  activated: 'Loading.',
+  timedOut: 'Exit this page expired.',
```
36degrees, line 18: Suggestion – would it make sense for these translation key to be more aligned with their English defaults?

```suggestion
  loading: 'Loading.',
  expired: 'Exit this page expired.',
```

src/govuk/components/exit-this-page/exit-this-page.mjs
```diff
@@ -0,0 +1,343 @@
...
+  pressTwoMoreTimes: 'Shift, press 2 more times to exit.',
+  pressOneMoreTime: 'Shift, press 1 more time to exit.'
+}
+
+/**
+ * JavaScript functionality for the Exit this Page component
```
36degrees, line 24: Can we mark everything in this class other than the constructor and init function as private? In all other classes we've deprecated everything and plan to make them private in v5 – we should  avoid these ever being 'public'.

app/src/views/full-page-examples/child-maintenance/index.njk
```diff
@@ -0,0 +1,167 @@
...
+  {{ govukBreadcrumbs({
+    classes: '',
+    collapseOnMobile: true,
+    items: [
+      {
+        html: 'Home',
+        href: '/'
+      },
+      {
+        html: ' Births, deaths, marriages and care',
```
colinrotherham, line 59: Tiny little stray space slipped in

```suggestion
        html: 'Births, deaths, marriages and care',
```

app/src/views/full-page-examples/child-maintenance/index.njk
```diff
@@ -0,0 +1,167 @@
...
+{% block beforeContent %}
+  {{ govukPhaseBanner({
+    tag: {
+      text: "beta"
+    },
+    html: 'This is a new service – your <a class="govuk-link" href="#">feedback</a> will help us to improve it.'
+  }) }}
+
+  {{ govukBreadcrumbs({
+    classes: '',
```
colinrotherham, line 51: We don't lint Nunjucks, but should we stick to double quotes everywhere still?

(Can use single quotes when they enclose double quotes)

```suggestion
    classes: "",
```

src/govuk/components/exit-this-page/exit-this-page.mjs
```diff
@@ -0,0 +1,343 @@
...
+    $light.classList.remove('govuk-exit-this-page__indicator-light--on')
+  })
+
+  // Turn on the ones we want on
+  var $lightsQueried = this.$indicatorContainer.querySelectorAll('.govuk-exit-this-page__indicator-light')
+  nodeListForEach($lightsQueried, function ($light, index) {
+    if (index < this.keypressCounter) {
+      $light.classList.add('govuk-exit-this-page__indicator-light--on')
+    }
+  }.bind(this))
```
36degrees, line 130: Suggestion: This could potentially be done in a single pass by using [`DOMTokenList.toggle`](https://developer.mozilla.org/en-US/docs/Web/API/DOMTokenList/toggle):

```suggestion
  // Turn on only the indicators we want on
  var $indicators = this.$indicatorContainer.querySelectorAll(
    '.govuk-exit-this-page__indicator-light'
  )

  nodeListForEach($indicators, function ($indicator, index) {
    $indicator.classList.toggle(
      'govuk-exit-this-page__indicator-light--on',
      index < this.keypressCounter
    )
  }.bind(this))
```

app/src/views/full-page-examples/child-maintenance/index.njk
```diff
@@ -0,0 +1,167 @@
...
+{% from "details/macro.njk" import govukDetails %}
+{% from "button/macro.njk" import govukButton %}
+
+{% set pageTitle = "Child maintenance" %}
+{% block pageTitle %}{{ pageTitle }} - GOV.UK{% endblock %}
+
+{% block skipLink %}
+  {{ super() }}
+  {{ govukSkipLink({
+    href: 'https://bbc.co.uk/weather/',
```
colinrotherham, line 36: Did we want both "exit" URLs to be different? (Button uses Google versus skip link using BBC)

src/govuk/components/exit-this-page/exit-this-page.mjs
```diff
@@ -0,0 +1,343 @@
...
+ *
+ * @param {MouseEvent} [event] - mouse click event
+ */
+ExitThisPage.prototype.exitPage = function (event) {
+  var redirectUrl = this.$button.getAttribute('href')
+
+  if (typeof event !== 'undefined' && event.target instanceof HTMLElement) {
+    event.preventDefault()
+    redirectUrl = event.target.getAttribute('href')
+  }
```
36degrees, line 147: I don't understand what this is doing – can you explain?

src/govuk/components/exit-this-page/exit-this-page.mjs
```diff
@@ -0,0 +1,343 @@
...
+  // As well as creating an overlay with text, we also set the body to hidden
+  // to prevent screen reader and sequential navigation users potentially
+  // navigating through the page behind the overlay during loading
+  document.body.classList.add('govuk-exit-this-page-hide-content')
+  this.$overlay = document.createElement('div')
+  this.$overlay.className = 'govuk-exit-this-page-overlay'
+  this.$overlay.setAttribute('role', 'alert')
+
+  document.body.appendChild(this.$overlay)
+  this.$overlay.innerText = this.i18n.t('activated')
```
36degrees, line 161: Not 100% sure but might be able to avoid an extra paint by doing these the other way around?If it needs to be this way round so that it's announced it might be worth adding a comment.

src/govuk/all.mjs
```diff
@@ -58,6 +59,13 @@ function initAll (config) {
     new ErrorSummary($errorSummary, config.errorSummary).init()
   }
 
+  var $exitThisPageButtons = $scope.querySelectorAll('[data-module="govuk-exit-this-page"]')
+  if ($exitThisPageButtons.length > 0) {
```
colinrotherham, line 63: Shall we ditch the length check? For the other components we go straight to `nodeListForEach()`

```suggestion
```

src/govuk/components/exit-this-page/exit-this-page.mjs
```diff
@@ -0,0 +1,343 @@
...
+ */
+ExitThisPage.prototype.setKeypressTimer = function () {
+  // Clear any existing timeout. This is so only one timer is running even if
+  // there are multiple keypresses in quick succession.
+  clearTimeout(this.keypressTimeoutId)
+
+  // Set a fresh timeout
+  this.keypressTimeoutId = setTimeout(function () {
+    this.resetKeypressTimer()
+  }.bind(this), this.timeoutTime)
```
36degrees, line 238: Suggestion: Think you can avoid the extra wrapper function here:

```suggestion
    // Set a fresh timeout
  this.keypressTimeoutId = setTimeout(
    this.resetKeypressTimer.bind(this),
    this.timeoutTime
  )
```

src/govuk/components/exit-this-page/exit-this-page.mjs
```diff
@@ -0,0 +1,343 @@
...
+ * JavaScript functionality for the Exit this Page component
+ *
+ * @class
+ * @param {HTMLElement} $module - HTML element that wraps the EtP button
+ * @param {ExitThisPageConfig} [config] - Exit this Page config
+ */
+function ExitThisPage ($module, config) {
+  this.$module = $module
+
+  var defaultConfig = {
```
colinrotherham, line 33:
```suggestion
  /** @type {ExitThisPageConfig} */
  var defaultConfig = {
```

src/govuk/components/exit-this-page/template.test.js
```diff
@@ -0,0 +1,83 @@
...
+
+  describe('Custom options', () => {
+    it('renders with text', () => {
+      const $ = render('exit-this-page', examples.testing)
+      const $button = $('.govuk-exit-this-page').find('.govuk-button')
+
+      expect($button.text()).toContain('Exit this test')
+    })
+
+    it('renders with a current tab URL', () => {
```
36degrees, line 43: What does this mean?

src/govuk/components/exit-this-page/exit-this-page.mjs
```diff
@@ -0,0 +1,343 @@
...
+    i18n: EXIT_THIS_PAGE_TRANSLATIONS
+  }
+
+  this.config = mergeConfigs(
```
colinrotherham, line 37: Mind adding all the JSDoc types like we do in the other components?

You'll start to see all the JavaScript type linting come to life and reveal problems if they exist

```suggestion  
  /**
   * @private
   * @type {ExitThisPageConfig}
   */
  this.config = mergeConfigs(
```

colinrotherham (changes requested):

app/src/views/examples/exit-this-page-with-skiplink/index.njk
```diff
@@ -0,0 +1,20 @@
+{% extends "layout.njk" %}
```
36degrees, line 1: Thanks for adding this and the full page example! Would also be good to add one to the 'Translated' example that shows the component in its translated state.

src/govuk/objects/_template.scss
```diff
@@ -12,6 +12,23 @@
     // would like the browser to stay on 100% text zoom by default.
     text-size-adjust: 100%;
 
+    // Add scroll padding to the top of govuk-template unless the exit this page
```
36degrees, line 15: I don't follow – if this is to prevent a potential issue with exit this page, why are we only adding the padding when exit this page is _not_ present?

36degrees (commented):
There's a couple of things I don't quite understand, but otherwise this is looking really good. Most of my comments are just suggestions on ways that the code could be simplified slightly – I'll leave it up to you to decide if they're worth doing.

src/govuk/components/exit-this-page/exit-this-page.mjs
owenatgov: Good idea. This, as with a few things in the ETP code, are likely hangovers for when the content we included was something else.

src/govuk/components/exit-this-page/template.test.js
owenatgov: Stale wording from when we did fancy tab stuff on click. Have since updated it.

src/govuk/components/exit-this-page/exit-this-page.mjs
owenatgov: [Toggle has fairly meagre browser support](https://caniuse.com/?search=toggle). Are we ok putting in a polyfill?

src/govuk/components/exit-this-page/exit-this-page.mjs
owenatgov: This is to handle the different ways that the button can be activated. If the button was clicked, we `preventDefault` and do some extra stuff before navigating. If it was activated via the keypress, we don't want to `preventDefault` as the keypress activation doesn't pass a `MouseEvent` to the the function.

We probably don't need to separate the ways we define `redirectUrl` here but we do need to have the `preventDefault` check. I'll add a clarifying comment.

src/govuk/objects/_template.scss
owenatgov: I think I got the wording confused here. It should be the other way around, so don't have `scroll-padding` if ETP isn't present. It's a complicated implementation because of how we're trying to cast as wide a net as possible and mitigate as much hassle for folks as possible. I'll update the comment and extend it a bit.

src/govuk/components/exit-this-page/exit-this-page.mjs
querkmachine: I think I flagged this at an earlier point and it was the case that the element and text need to be inserted separately for screen readers to reliably pick it up.

src/govuk/components/exit-this-page/exit-this-page.mjs
owenatgov: On a second look I'm going to object to these changes as they most accurately represent what the text is for. "Loading", IMO, wouldn't communicate to folks using the translation API that this is for when the user activates the button. Expired I could be swayed on. Unless you have strong opinions I'm gonna skip over this one.

src/govuk/components/exit-this-page/exit-this-page.mjs
owenatgov: I'll add a comment with Beep's point

src/govuk/components/exit-this-page/exit-this-page.mjs
36degrees: Gotcha, so `exitPage` can be called either:

- directly as an event handler for the `click` event
- indirectly from the event handler for the `keypress` event (`handleKeypress`)

Have you considered adding a `handleClick` function as an intermediary?

```js
ExitThisPage.prototype.handleClick = function (event) {
    event.preventDefault()
    this.exitPage()
}
```

That'd make the purpose of the `exitPage` method clearer and mean that you never need to pass it the `event` (which at the minute is optional).

src/govuk/components/exit-this-page/exit-this-page.mjs
owenatgov: So sadly it looks like we can't do this. A note from @colinrotherham on slack:

>We mainly can’t use `@private` in v4 as the `this` keyword (inside a constructor function or methods) can be `window` until it’s been instantiated. The TypeScript compiler isn’t clever enough at this sort of thing until ES2015 classes come along

The chat's continuing but we might not have a suitable workaround at the moment.

src/govuk/components/exit-this-page/exit-this-page.mjs
owenatgov: Doh! My fault for not checking my work 🤦🏻 I'll do this then!

src/govuk/components/exit-this-page/exit-this-page.mjs
```diff
@@ -0,0 +1,393 @@
...
+ * pointer.
+ *
+ * We do this to differentiate it from the keyboard activation event because we
+ * need to run `e.preventDefault` as the button or skiplink are both links and we
+ * want to apply some additional logic in `exitPage` before navigating.
+ *
+ * @deprecated Will be made private in v5.0
+ * @param {MouseEvent} e - mouse click event
+ */
+ExitThisPage.prototype.handleClick = function (e) {
```
colinrotherham, line 203: Another super tiny thing

Did we want to use ~`e`~ `event` like the other components?

```suggestion
ExitThisPage.prototype.handleClick = function (event) {
```

src/govuk/components/exit-this-page/exit-this-page.mjs
```diff
@@ -0,0 +1,393 @@
...
+
+/**
+ * Initiates the redirection away from the current page.
+ * Includes the loading overlay functionality, which covers the current page with a
+ * white overlay so that the contents are not visible during the loading
+ * process. This is particularly important on slow network connections.
+ *
+ * @deprecated Will be made private in v5.0
+ */
+ExitThisPage.prototype.exitPage = function (redirectUrl) {
```
colinrotherham, line 171: Just double checking, was `redirectUrl` supposed to be an optional override?

It's not used at the moment

```suggestion
ExitThisPage.prototype.exitPage = function () {
```

src/govuk/components/exit-this-page/exit-this-page.mjs
```diff
@@ -0,0 +1,393 @@
...
+    normaliseDataset($module.dataset)
+  )
+
+  this.i18n = new I18n(extractConfigByNamespace(this.config, 'i18n'))
+
+  /** @deprecated Will be made private in v5.0 */
+  this.$module = $module
+
+  /** @deprecated Will be made private in v5.0 */
+  this.$button = $module.querySelector('.govuk-exit-this-page__button')
```
colinrotherham, line 52: We redirect to the button's `href` a bit later on:

```js
window.location.href = this.$button.getAttribute('href')
```

Think we should check `instanceof HTMLAnchorElement` then can safely access `.href` instead

src/govuk/components/exit-this-page/exit-this-page.mjs
```diff
@@ -0,0 +1,393 @@
...
+    // Clear the timeout for the keypress timeout message clearing itself
+    if (this.timeoutMessageId !== null) {
+      clearTimeout(this.timeoutMessageId)
+      this.timeoutMessageId = null
+    }
+
+    if (this.keypressCounter >= 3) {
+      this.keypressCounter = 0
+
+      clearTimeout(this.keypressTimeoutId)
```
colinrotherham, line 242: For the timeout `this.timeoutMessageId` we did a "not null" check first

Do we need one here?

colinrotherham: Although we clear timers and intervals without checking in a few places, so maybe we're being [too strict above](https://github.com/alphagov/govuk-frontend/pull/2545/files#diff-73815616f3d4872193e2c150dfa54c35ee4232882d6b0441f749c833636a3d56R234)

src/govuk/components/exit-this-page/exit-this-page.mjs
owenatgov: Ah whoops. This was a carryover from when I was trying out splitting how the URL we're passing to the navigation was defined. I'll get rid of it.

src/govuk/components/exit-this-page/exit-this-page.mjs
owenatgov: I wonder if I should check this earlier on in the constructor function? We currently don't do any `$module` verification like we do for other components. I could check `$module` and `$button` at least at the outset?

colinrotherham: Sounds brill. Yeah try and check them both before adding to `this.$xxx`

Then the compiler knows exactly what they are later 👍

src/govuk/components/exit-this-page/exit-this-page.mjs
```diff
@@ -0,0 +1,404 @@
...
+    normaliseDataset($module.dataset)
+  )
+
+  this.i18n = new I18n(extractConfigByNamespace(this.config, 'i18n'))
+
+  /** @deprecated Will be made private in v5.0 */
+  this.$module = $module
+
+  /** @deprecated Will be made private in v5.0 */
+  this.$button = $module.querySelector('.govuk-exit-this-page__button')
```
colinrotherham, line 61: Liking the new checks ✅

Just need to swap in `$button` here

```suggestion
  this.$button = $button
```

src/govuk/components/exit-this-page/exit-this-page.mjs
```diff
@@ -0,0 +1,404 @@
+/* eslint-disable es-x/no-function-prototype-bind -- Polyfill imported */
+
+import { normaliseDataset } from '../../common/normalise-dataset.mjs'
+import { nodeListForEach, mergeConfigs, extractConfigByNamespace } from '../../common.mjs'
+import { I18n } from '../../i18n.mjs'
+import '../../vendor/polyfills/Element/prototype/classList.mjs'
```
colinrotherham, line 6: Can you just double check if we need this?

Other components that use `addEventListener()` import it

If a service team JUST imports `exit-this-page.mjs`, do they have everything they need polyfilled?

```suggestion
import '../../vendor/polyfills/Element/prototype/classList.mjs'
import '../../vendor/polyfills/Event.mjs' // addEventListener, event.target normalization and DOMContentLoaded
```

owenatgov: Let's import it just in case. Great spot.

src/govuk/components/exit-this-page/template.njk
```diff
@@ -0,0 +1,17 @@
+{% from "../button/macro.njk" import govukButton -%}
+
+<div
+  {%- if params.id %} id="{{ params.id }}"{% endif %}
+  class="govuk-exit-this-page {%- if params.classes %} {{ params.classes }}{% endif %}"
```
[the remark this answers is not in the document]
owenatgov: I've added some extra whitespace management. I think the whitespace management is already pretty ok though. Lemme know what you think.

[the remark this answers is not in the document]
colinrotherham: I think the new `-%}` were already cared for by the next line’s `{%-` but no harm

We haven’t handled the new line before `class` or `data-module` so it’s the same as it was, sorry 😬 

That alright?

Issue #2181, opened 2021-04-06
Focus and hover states of product name in header are much wider than the content


hannalaakso:
## Description of the issue
The focus and hover states on the GOV.UK product name in the header are much wider than the enclosed content **when the product name breaks onto a new line**.

**_Chrome 89 on Mac OS_**
![Screenshot 2021-04-06 at 10 51 32](https://user-images.githubusercontent.com/5007934/113693048-12523300-96c6-11eb-8187-9ca406f77e4c.png)
![Screenshot 2021-04-06 at 10 58 49](https://user-images.githubusercontent.com/5007934/113693966-1af73900-96c7-11eb-91ba-cde21acba59f.png)

**_Chrome on Galaxy S21_**
![Screenshot 2021-04-06 at 10 43 30](https://user-images.githubusercontent.com/5007934/113692209-36614480-96c5-11eb-960c-a14716efc19b.png)

Generally, the focus/hover states on the product name (whether on one or two lines) are inconsistent with the service name focus/hover states which only wrap the GOV.UK text. 

![Screenshot 2021-04-06 at 10 42 11](https://user-images.githubusercontent.com/5007934/113693512-960c1f80-96c6-11eb-9656-b455e0183075.png)

In contrast, with the product name both the GOV.UK text and the product name itself are wrapped in the focus/hover state.

There might be a reason for this inconsistency (ie. we think that services often link to GOV.UK from the header, whereas products link to the product homepage), in which case we should consider documenting it.

## Actual vs expected behaviour
Focus and hover states should neatly wrap around the enclosed content, unless 

Focus and hover states between the service and product names should probably be consistent, or the reason for the inconsistency should be documented.

vanitabarrett:
Note: the PR mentioned above will partly fix the issue on desktop, but won't address it on mobile. We should make sure this issue stays open even after the PR is merged, as it might be something to revisit when we look at other changes to the header.

domoscargin:
This has been [resolved in the rebranded component](https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/src/govuk/components/header/_index.scss#L144-L159).

We do not plan to resolve it for the older brand style.

Issue #1502, opened 2019-07-18
Product name wraps onto new line with non full-width header


36degrees:
The product name wraps when it doesn't need to, when used in a non full-width header.

![localhost_3000_components_header_with-product-name_preview](https://user-images.githubusercontent.com/121939/61457889-32294c80-a961-11e9-9350-3e30f339f243.png)

I believe this is because of the `width: 33%` declaration on `govuk-header__logo`, which is necessary for correct layout when the header has a service name and/or navigation:

https://github.com/alphagov/govuk-frontend/blob/edec2e977b64c2b54332c1a97a1ec1b2461f84a6/src/govuk/components/header/_header.scss#L136-L146

nickcolley:
I've investigated this issue and https://github.com/alphagov/govuk-frontend/issues/1635 and think that we probably want to try and explore flexbox/grid with an approach that allows the content in the header dictate space.

hannalaakso:
Some notes from team discussion:
-  Test with longest possible product name: “Platform as a Service”?

kr8n3r:
this is what I've ended up doing on paas product pages
![paas-header](https://user-images.githubusercontent.com/3758555/85371965-088d8380-b529-11ea-949b-00c9ba844071.gif)

vanitabarrett:
As discussed in V4 scoping session, we probably need to get a better idea of the changes GOV.UK are making to the header and any changes we may need to make to match before we look into this.

PR #2549, opened 2022-02-16
Fix header with product name focus and hover state length

owenatgov:
## What
Changes the desktop display properly of the header homepage link from `inline-block` to `inline`.

## Why
Fixes https://github.com/alphagov/govuk-frontend/issues/2181

You can test this by starting a local review app, going to the header component and looking at the focus and hover states of "header with product name".

### Focus state
Before:
Screenshot 2022-02-16 at 11 18 10

After:
Screenshot 2022-02-16 at 11 17 34

### Hover state
Before:
Screenshot 2022-02-16 at 11 18 04

After:
Screenshot 2022-02-16 at 11 17 28


owenatgov:
@vanitabarrett Good spot, how strange! It looks like this is coming from the `box-shadow` applied by the focus state. I'm just trying to work out _why_ it's doing that; I don't really want to turn off a feature of our focus state.

owenatgov:
@vanitabarrett I've found a solution that involved amending the `box-shadow` of the focus state. I'm not 100% comfortable with it but it's the best solution that didn't involve changing the height of the header, which I want to avoid. Let me know what you think.

vanitabarrett:
Hm, I'm not 100% sure about this change. I think the change from `inline-block` to `inline` means that on super narrow screens things start getting cut off, whereas before they'd maintain the right size and overflow the container instead.

Example:

## Before (234px wide)
Screenshot 2022-02-16 at 15 14 59

## After (234px wide)
Screenshot 2022-02-16 at 15 15 04

Unfortunately I don't really have any other suggestions... I am wondering if this work would get any easier if we solved https://github.com/alphagov/govuk-frontend/pull/1501, but maybe not.

owenatgov:
@vanitabarrett Bugger. I think you're right that the root of this is elsewhere. I'm pretty sure that issue is linked to this so I'll investigate if there's a way to solve both of these.

owenatgov:
@vanitabarrett After assessing this a bit more, I reckon https://github.com/alphagov/govuk-frontend/issues/1502 is going to take a more thorough investigation. For now I've found a quick fix that solves the small screen problem by using a media query to control the `display: inline` change. Let me know what you think.

vanitabarrett:
@owenatgov Thanks for digging into it a bit! I agree, it might need a more thorough investigation. I quite like this quick fix, nice work 👍 I'm wondering if the box-shadow change should also be inside the media query, as we're essentially not changing the focus state at all in lower breakpoints?

owenatgov:
@vanitabarrett 100%. Applied!

vanitabarrett (approved):
As mentioned in Slack, this just needs a Changelog entry and then it should be ready to go

vanitabarrett:
@owenatgov Would you also mind adding the linked issue to the [next] milestone (indicating it will be release in the next version of GOV.UK Frontend)?

Issue #2070, opened 2020-12-15
Add visual regression testing


36degrees:
## What

Add visual regression testing to GOV.UK Frontend

## Why

This came up as part of a ['tech & ops health check' exercise](https://github.com/alphagov/govuk-design-system/issues/1395) the team ran in November 2020, as part of the 'testing – pre-deployment' practise.

We scored ourselves as 'good' for that practise, but said that we were 'moderately concerned' about it as we know there are some areas that could be improved.

Visual regression testing was highlighted as one area we'd like to add to GOV.UK Frontend. We believe that it will help us to avoid shipping unintended changes to the design or layout of components, especially in non-default contexts (for example, changes that are only visible when using compatibility mode, or if the user is using Windows High Contrast Mode).

## Done when

- [x] Trial Percy in GOV.UK Frontend repo - in progress until end of Q4
- [x] Revisit how trial went and decide next steps

vanitabarrett:
Percy seems to be working pretty well for our team. People are either looking at the screenshots and finding them useful, or they haven't needed to do so yet - probably due to the nature of the things they've been working on.

Given we're only using ~43,000 screenshots per month in total (across the GDS account), we're not near our monthly limit so it makes sense to keep Percy running.

I've raised a new card for exploring other scenarios in which we may want to generate screenshots https://github.com/alphagov/govuk-frontend/issues/2593 which we may want to pick up in future.

PR #2551, opened 2022-02-18
Add Percy visual regression testing

vanitabarrett:
Part of https://github.com/alphagov/govuk-frontend/issues/2070

## What
This PR adds Percy visual regression testing to GOV.UK Frontend. Screenshots are generated for each default example for every component, both with JavaScript enabled and disabled.

## Proposal
As discussed in dev catch-up, we generally think visual regression tools may help us catch unintended visual changes when we're working. There were some concerns around hitting screenshot limits, Percy checks being ignored if approvals aren't required (PRs can get rushed through into `main` without checking the screenshots), and generally it not being a replacement for manual testing as it doesn't cover older browsers, high contrast mode etc.

I'm proposing a trial period with this PR. I suggest we:

1. add Percy to GOV.UK Frontend until the end of the quarter (~ early April) and revisit our experience then
2. ~initially make Percy a required check~ **Edit to add: unfortunately, because PRs from forks don't have access to Github secrets, Percy won't run on these PRs. We therefore can't make it a required check, as it'll never run/pass for these PRs.**
3. put the responsibility for approving Percy runs onto the person reviewing the pull request

With all of the above, we should revisit decisions earlier than April if they're causing us issues and slowing down development.

## Why
Visual regression tools can help us catch visual changes which we weren't expecting, for example: if we make changes to one area of the codebase, without realising that it will have an effect elsewhere.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -23,4 +23,6 @@ jobs:
         run: npm ci
 
       - name: Run tests
-        run: npm test
+        run: node_modules/.bin/percy exec -- npm test
+        env:
+          PERCY_TOKEN: ${{ secrets.PERCY_TOKEN }}
--- a/docs/releasing/testing-and-linting.md
+++ b/docs/releasing/testing-and-linting.md
@@ -1,6 +1,6 @@
 # Testing and linting
 
-The CI lints SASS and JavaScript, runs unit and functional tests with Node, and generates snapshots for visual regression testing (Percy.io).
+The CI lints SASS and JavaScript, runs unit and functional tests with Node, and generates snapshots for [visual regression testing](https://www.browserstack.com/percy/visual-testing).
 
 ## Running all tests locally
 
@@ -50,20 +50,20 @@ This will update the snapshot file. Commit this file separately with a commit me
 
 ## Visual regression testing with Percy
 
-We use [Percy](https://percy.io/), a visual regression testing tool, to generate and store screenshots of our components to help us check for any unintended visual changes.
+[Percy](https://percy.io/) is a visual regression testing tool. We use it to generate and store screenshots of our components to help us check for any unintended visual changes.
 
-We generate two screenshots for each default example of every component: one with JavaScript enabled and one with JavaScript disabled. Screenshots are not taken for all the different variations of each component. This tool is not a replacement for manual testing.
+We generate 2 screenshots for each default example of every component. One example has JavaScript enabled, the other has JavaScript disabled. Screenshots are not taken for all the different variations of each component. This tool is not a replacement for manual testing.
 
-The screenshots are public, so they can be checked without logging in. A BrowserStack account is needed to approve or reject any changes. It's the responsibility of the person reviewing the pull request code to approve any visual changes that Percy highlights.
+The screenshots are public, so you can check them without logging in. A BrowserStack account is needed to approve or reject any changes (if you don't have access, ask your tech lead for help). If you're the reviewer of the pull request code, it's your responsibility to approve or request changes for any visual changes Percy highlights.
 
-When running the tests locally (e.g: via `npm test`), Percy commands are ignored and no screenshots are generated. You will see the following message in your command line output: `[percy] Percy is not running, disabling snapshots`.
+When you run the tests locally (for example, using `npm test`), Percy commands are ignored and Percy does not generate any screenshots. You will see the following message in your command line output: `[percy] Percy is not running, disabling snapshots`.
 
 ### PRs from forks
-When Github Actions is running against a PR from a fork, the Percy secret is not available and therefore no screenshots are generated. Other tests will continue to run as normal. You will see the following messages in the output:
+When Github Actions is running against a PR from a fork, the Percy secret is not available and Percy does not generate any screenshots. Other tests will continue to run as normal. You will see the following messages in the output:
 
 ```
 [percy] Skipping visual tests
 [percy] Error: Missing Percy token
 ```
 
-This is the reason why we are unable to make Percy a required check for this repo. However, we should continue to act as if it is required. Percy runs should be approved before merging, and we should treat any failures as blocking.
+Being unable to run Percy on PRs from forks is the reason we're unable to make Percy a required check for this repo. However, we should continue to act as if a Percy check is required. If the Percy build fails, do not merge the pull request even though GitHub would let you. Continue to review the changes, and approve or reject them as you would normally.
--- a/docs/releasing/testing-and-linting.md
+++ b/docs/releasing/testing-and-linting.md
@@ -1,6 +1,6 @@
 # Testing and linting
 
-The CI lints SASS and JavaScript, and runs unit and functional tests with Node.
+The CI lints SASS and JavaScript, runs unit and functional tests with Node, and generates snapshots for visual regression testing (Percy.io).
 
 ## Running all tests locally
 
@@ -47,3 +47,13 @@ If a snapshot test fails, review the difference in the console. If the change is
 `npm test -- -u src/govuk/components/button`
 
 This will update the snapshot file. Commit this file separately with a commit message that explains you're updating the snapshot file and an explanation of what caused the change.
+
+## Visual regression testing with Percy
+
+We use [Percy](https://percy.io/), a visual regression testing tool, to generate and store screenshots of our components to help us check for any unintended visual changes.
+
+We generate two screenshots for each default example of every component: one with JavaScript enabled and one with JavaScript disabled. Screenshots are not taken for all the different variations of each component. This tool is not a replacement for manual testing.
+
+The screenshots are public, so they can be checked without logging in. A BrowserStack account is needed to approve or reject any changes. It's the responsibility of the person reviewing the pull request code to approve any visual changes that Percy highlights.
+
+Note: when running the tests locally via `npm test`, Percy commands are ignored and no screenshots are generated. You will see the following message in your command line output: `[percy] Percy is not running, disabling snapshots`.
--- a/docs/releasing/testing-and-linting.md
+++ b/docs/releasing/testing-and-linting.md
@@ -56,4 +56,14 @@ We generate two screenshots for each default example of every component: one wit
 
 The screenshots are public, so they can be checked without logging in. A BrowserStack account is needed to approve or reject any changes. It's the responsibility of the person reviewing the pull request code to approve any visual changes that Percy highlights.
 
-Note: when running the tests locally via `npm test`, Percy commands are ignored and no screenshots are generated. You will see the following message in your command line output: `[percy] Percy is not running, disabling snapshots`.
+When running the tests locally (e.g: via `npm test`), Percy commands are ignored and no screenshots are generated. You will see the following message in your command line output: `[percy] Percy is not running, disabling snapshots`.
+
+### PRs from forks
+When Github Actions is running against a PR from a fork, the Percy secret is not available and therefore no screenshots are generated. Other tests will continue to run as normal. You will see the following messages in the output:
+
+```
+[percy] Skipping visual tests
+[percy] Error: Missing Percy token
+```
+
+This is the reason why we are unable to make Percy a required check for this repo. However, we should continue to act as if it is required. Percy runs should be approved before merging, and we should treat any failures as blocking.
--- a/package-lock.json
+++ b/package-lock.json
@@ -1934,6 +1934,24 @@
       "integrity": "sha512-cxCMLQ6uYkfVXIb4qgDnxnYWlgtiLr64LEE6BKQUG6d+3D5TTzfexVXGFc+miB37ntD4aCEvW9WWG8kPswYSSg==",
       "dev": true
     },
+    "@percy/puppeteer": {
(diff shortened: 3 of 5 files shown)


fofr:
Nice! Will the Percy screenshots and diffs be open to the public?

docs/releasing/testing-and-linting.md
```diff
@@ -47,3 +47,13 @@ If a snapshot test fails, review the difference in the console. If the change is
...
 
 This will update the snapshot file. Commit this file separately with a commit message that explains you're updating the snapshot file and an explanation of what caused the change.
+
+## Visual regression testing with Percy
+
+We use [Percy](https://percy.io/), a visual regression testing tool, to generate and store screenshots of our components to help us check for any unintended visual changes.
+
+We generate two screenshots for each default example of every component: one with JavaScript enabled and one with JavaScript disabled. Screenshots are not taken for all the different variations of each component. This tool is not a replacement for manual testing.
+
+The screenshots are public, so they can be checked without logging in. A BrowserStack account is needed to approve or reject any changes. It's the responsibility of the person reviewing the pull request code to approve any visual changes that Percy highlights.
```
domoscargin, line 57: I can access these when logged in, but am struggling to figure out how in an incognito window.

Created a dummy account for Github so I could view the Github Action logs and head off to the Percy build. Created a Browserstack account, but am getting "Access Denied" on the build at https://percy.io/55ba5e1a/govuk-frontend/builds/15872008

Screenshot 2022-02-23 at 09 57 56

domoscargin:
I love this so, so much. Left a comment about findability.

docs/releasing/testing-and-linting.md
vanitabarrett: @domoscargin Doh, hadn't enabled that setting. Do you want to give it a try now?

domoscargin: That works! 🎉  No login necessary.

.github/workflows/tests.yml
```diff
@@ -23,4 +23,6 @@ jobs:
         run: npm ci
 
       - name: Run tests
-        run: npm test
+        run: node_modules/.bin/percy exec -- npm test
+        env:
+          PERCY_TOKEN: ${{ secrets.PERCY_TOKEN }}
```
36degrees, line 28: This secret won't be available to PRs from forks, so we'll need to think about how this works for external contributors.

At the very least, we'll need to make sure this fails gracefully (the tests still run and pass) – based on the documentation you added I'm guessing it will still run the tests with '[percy] Percy is not running, disabling snapshots' but would be good to check this.

We'll also need to think about how this works if we aren't able to check for visual regressions in external PRs until after they've been merged.

docs/releasing/testing-and-linting.md
```diff
@@ -47,3 +47,13 @@ If a snapshot test fails, review the difference in the console. If the change is
 `npm test -- -u src/govuk/components/button`
 
 This will update the snapshot file. Commit this file separately with a commit message that explains you're updating the snapshot file and an explanation of what caused the change.
+
+## Visual regression testing with Percy
```
36degrees, line 51: Thanks for adding these 🎉

36degrees (commented):
Exciting stuff! Have you by any chance looked into how many snapshots you think we might end up taking in an average month?

domoscargin (approved):
This is great and should hopefully give us a bit more confidence that our changes aren't breaking unexpected things.

One concern is that Percy.io appears to be fully broken when JavaScript is disabled, so I could foresee potential issues in reaching and approving changes during patchy internet times (and also JS is just more likely to fail).

Happy to see this being trialled!

.github/workflows/tests.yml
vanitabarrett: That's a good point. I've checked that behaviour, and you're right, the tests are run without Percy generating snapshots and you see this in the output:

```
[percy] Skipping visual tests
[percy] Error: Missing Percy token
[percy] Running "npm test"
.... [normal test output here]
```

I'll add some documentation for that, and have a think about what the process might look like for that..

vanitabarrett: @36degrees I've been digging into this a bit more, and I think the fact that the checks can't run on a PR from a fork means that we can't make Percy a required check either (it'll never run for PRs from forks, and I can't find a way to manually trigger any kind of Percy build from our side).

I think that means our only option is to not mark Percy as a required check and just be disciplined in trying to review the builds when we review. Can you think of any other alternatives?

36degrees: It would be significantly more work, but if we wanted to we could explore using the `pull_request_target` or `workflow_run` events, which run in the context of the base of the pull request (likely `main`) and have access to secrets.

A workflow might do something like:

1) When a PR is opened, a workflow triggered by the `pull_request` event compile the JS and CSS from that branch and saves them as an artefact
2) Kick off another workflow using the `workflow_run` event, which runs in the context of the base branch and has access to secrets
3) That workflow downloads the built JS and CSS and runs the review app using the compiled JS and CSS, then runs the tests to take snapshots
4) Pass the context of the original PR to Percy to allow it to associate it with the PR, add the status checks etc.

We'd need to be very careful to ensure that we're not introducing potential vulnerabilities to our workflows – there's a good article on this here: https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

My initial thought is we should probably try the simple thing first and explore this route in the future if we find it's causing friction?

docs/releasing/testing-and-linting.md
```diff
@@ -1,6 +1,6 @@
 # Testing and linting
 
-The CI lints SASS and JavaScript, and runs unit and functional tests with Node.
+The CI lints SASS and JavaScript, runs unit and functional tests with Node, and generates snapshots for visual regression testing (Percy.io).
```
EoinShaughnessy, line 3: Will all our users know what 'visual regression testing' is? If not, do we need to link to an explanation?

Not sure about having 'Percy.io' in brackets at the end - there's no explanation of what it is (unless Percy.io = visual regression testing), and it isn't hyperlinked...maybe it would be ok to delete it from here, since we both hyperlink and explain it in line 53...

docs/releasing/testing-and-linting.md
```diff
@@ -47,3 +47,23 @@ If a snapshot test fails, review the difference in the console. If the change is
 `npm test -- -u src/govuk/components/button`
 
 This will update the snapshot file. Commit this file separately with a commit message that explains you're updating the snapshot file and an explanation of what caused the change.
+
+## Visual regression testing with Percy
+
+We use [Percy](https://percy.io/), a visual regression testing tool, to generate and store screenshots of our components to help us check for any unintended visual changes.
```
EoinShaughnessy, line 53:
```suggestion
[Percy](https://percy.io/) is a visual regression testing tool. We use it to generate and store screenshots of our components to help us check for any unintended visual changes.
```

docs/releasing/testing-and-linting.md
```diff
@@ -47,3 +47,23 @@ If a snapshot test fails, review the difference in the console. If the change is
...
+
+We generate two screenshots for each default example of every component: one with JavaScript enabled and one with JavaScript disabled. Screenshots are not taken for all the different variations of each component. This tool is not a replacement for manual testing.
```
EoinShaughnessy, line 55:
```suggestion
We generate 2 screenshots for each default example of every component. One example has JavaScript enabled, the other has JavaScript disabled. Screenshots are not taken for all the different variations of each component. This tool is not a replacement for manual testing.
```

docs/releasing/testing-and-linting.md
```diff
@@ -47,3 +47,23 @@ If a snapshot test fails, review the difference in the console. If the change is
...
+
+The screenshots are public, so they can be checked without logging in. A BrowserStack account is needed to approve or reject any changes. It's the responsibility of the person reviewing the pull request code to approve any visual changes that Percy highlights.
```
EoinShaughnessy, line 57:
```suggestion
The screenshots are public, so you can check them without logging in. A BrowserStack account is needed to approve or reject any changes. If you're the reviewer of the pull request code, it's your responsibility to approve any visual changes Percy highlights.
```

Does the user need to have a BrowserStack account? If they do, do we need to link to BrowserStack signup?

docs/releasing/testing-and-linting.md
```diff
@@ -47,3 +47,23 @@ If a snapshot test fails, review the difference in the console. If the change is
...
+
+When running the tests locally (e.g: via `npm test`), Percy commands are ignored and no screenshots are generated. You will see the following message in your command line output: `[percy] Percy is not running, disabling snapshots`.
```
EoinShaughnessy, line 59:
```suggestion
When you run the tests locally (for example, using `npm test`), Percy commands are ignored and Percy does not generate any screenshots. You will see the following message in your command line output: `[percy] Percy is not running, disabling snapshots`.
```

docs/releasing/testing-and-linting.md
```diff
@@ -47,3 +47,23 @@ If a snapshot test fails, review the difference in the console. If the change is
...
+
+### PRs from forks
+When Github Actions is running against a PR from a fork, the Percy secret is not available and therefore no screenshots are generated. Other tests will continue to run as normal. You will see the following messages in the output:
```
EoinShaughnessy, line 62:
```suggestion
When Github Actions is running against a PR from a fork, the Percy secret is not available and Percy does not generate any screenshots. Other tests will continue to run as normal. You will see the following messages in the output:
```

docs/releasing/testing-and-linting.md
```diff
@@ -47,3 +47,23 @@ If a snapshot test fails, review the difference in the console. If the change is
...
+
+```
+[percy] Skipping visual tests
+[percy] Error: Missing Percy token
+```
+
+This is the reason why we are unable to make Percy a required check for this repo. However, we should continue to act as if it is required. Percy runs should be approved before merging, and we should treat any failures as blocking.
```
EoinShaughnessy, line 69:
```suggestion
These messages are the reason we're unable to make Percy a required check for this repo. However, we should continue to act as if a Percy check is required. You should approve Percy runs before merging pull requests. Treat any failures as blocking.
```
Will it be obvious to users what action they should take in order to "treat any failures as blocking"?

docs/releasing/testing-and-linting.md
```diff
@@ -47,3 +47,23 @@ If a snapshot test fails, review the difference in the console. If the change is
 `npm test -- -u src/govuk/components/button`
 
 This will update the snapshot file. Commit this file separately with a commit message that explains you're updating the snapshot file and an explanation of what caused the change.
+
+## Visual regression testing with Percy
```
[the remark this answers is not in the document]
EoinShaughnessy: @vanitabarrett Content looks really good! Left a few comments.

docs/releasing/testing-and-linting.md
vanitabarrett: @EoinShaughnessy They need to have a BrowserStack account to approve or reject changes. I'm not sure about adding a link to BrowserStack signup - I think an existing member needs to add them to the Design System team. Maybe I'll add a line about asking the tech lead in the team to get you access...

docs/releasing/testing-and-linting.md
vanitabarrett: I guess what I mean there is, if the Percy build fails, don't just merge the PR because Github lets you. You should continue to review the changes and approve/reject them as necessary. There probably is a clearer way of writing that...

EoinShaughnessy: Thanks for explaining! Your wording looks good - maybe we could add in something like the following:

```suggestion
These messages are the reason we're unable to make Percy a required check for this repo. However, we should continue to act as if a Percy check is required. If the Percy build fails, do not merge the pull request even though GitHub would let you. Continue to review the changes, and approve or reject them as you would normally.
```

vanitabarrett:
Thanks @EoinShaughnessy , I think I've addressed those comments if you want to take a second look.

docs/releasing/testing-and-linting.md
```diff
@@ -47,3 +47,23 @@ If a snapshot test fails, review the difference in the console. If the change is
...
+
+### PRs from forks
+When Github Actions is running against a PR from a fork, the Percy secret is not available and Percy does not generate any screenshots. Other tests will continue to run as normal. You will see the following messages in the output:
+
+```
+[percy] Skipping visual tests
+[percy] Error: Missing Percy token
+```
+
+Being unable to run Percy on PRs from forks is the reason why we are unable to make Percy a required check for this repo. However, we should continue to act as if a Percy check is required. If the Percy build fails, do not merge the pull request even though GitHub would let you. Continue to review the changes, and approve or reject them as you would normally.
```
EoinShaughnessy, line 69:
```suggestion
Being unable to run Percy on PRs from forks is the reason we're unable to make Percy a required check for this repo. However, we should continue to act as if a Percy check is required. If the Percy build fails, do not merge the pull request even though GitHub would let you. Continue to review the changes, and approve or reject them as you would normally.
```

EoinShaughnessy (approved):
@vanitabarrett Left one last tiny (optional!) suggestion - but yeah, content looks great!

PR #2567, opened 2022-03-02
Upgrade to stylelint-config-gds 0.2.0

kevindew:
This re-purposes a draft PR to solicit feedback on the changes from stylelint-config-gds 0.2.0 into actually applying the changes - now that we've released it.

Thanks @querkmachine, @36degrees and @vanitabarrett for chiming in. Apologies if the subsequent comments look a little out of context.


kevindew:
Thanks @querkmachine for taking a look and your thoughts.

I can understand your concern, it makes sense. My own opinion is that it's a bit heavy handed in all the places it applies to, I.e. it makes sense to me that it'd want to be consistent with something simple like padding and wanting consistency in the approach (and I think the fixes in here are, imo, an improvement) but with the quantity of shorthand options you have for background, animation and grid it'd seem rather hard to actually trigger it and could be an unwelcome surprise to a dev.

As it's a community rule, and not one we have introduced, it's consistent with our linting principles to try work with it so long as it doesn't cause pain - if it does cause pain then it should be easier to justify an exclusion as we can link to an example in our documentation.

As an aside, the ideal situation if this causes pain would be to open a PR to change it in the community standards themselves: https://github.com/stylelint/stylelint-config-standard

36degrees:
I can't see where the rule is coming from – looking at `stylelint-config-standard` the only reference to `declaration-block-no-redundant-longhand-properties` is [in the changelog where it mentions the rule being removed](https://github.com/stylelint/stylelint-config-recommended/search?q=declaration-block-no-redundant-longhand-properties&type=code).

That removal [appears to have come from a PR in 2017 which was justified by similar reasons](https://github.com/stylelint/stylelint-config-recommended/issues/14) to those raised by @querkmachine.

😕

kevindew:
@36degrees looks like you're in stylelint-config-recommended there which is what stylelint-config-standard builds on. This rule was added to stylelint-config-standard in version 23: https://github.com/stylelint/stylelint-config-standard/blob/main/CHANGELOG.md#2300. It unfortunately lacks a solid justification for it being an added last info seems to be on: https://github.com/stylelint/stylelint-config-standard/issues/113 where opinions have seemingly since changed.

kevindew:
@36degrees @querkmachine just wanted to see if you had any further thoughts on this? do you have a stance of whether you'd prefer the rule removed in our standards, happy for it to stay or ambivalent? Thanks

querkmachine:
My personal preference would be that the rule isn't enforced, yes. I find it strange that stylelint-config-standard has seemingly added it without any real justification and with historical resistance to it being enabled, and it only seems to exist to enforce brevity over readability, which isn't a problem we have to solve. 

My main weariness is that if the historical convention has been to align with what stylelint-config-standard does, I don't think my concerns alone are reason enough to break that convention.

vanitabarrett:
I've had a quick look through the changes and agree with the concerns @querkmachine has about `declaration-block-no-redundant-longhand-properties` and the effects it has on readability (all the other changes seem fairly minor, I'm happy with those). My personal preference would also be to disable that rule.

kevindew:
Now that stylelint-config-gds 0.2.0 is released I've rebased this PR to use the actual release. So this PR should be ready for merging.

querkmachine:
This looks good to me 👍 

The [coding standards documentation](https://github.com/alphagov/govuk-frontend/blob/main/docs/contributing/coding-standards/css.md) will need reviewing to ensure that the code standards prescribed there are aligned with the new linting rules.

querkmachine (approved):

vanitabarrett (approved):

Issue #2231, opened 2021-05-19
Use of rgba() limits values for $govuk-text-colour


daniel-ac-martin:
## Description of the issue
>         SassError: $color: inherit is not a color.
>             ╷
>         272 │     color: rgba($govuk-text-colour, .99);

Until the most recent version it has been possible to set `$govuk-text-colour` to `inherit`. This is useful when you wish to set the colour elsewhere, for example on the body tag.

(I do this here: https://not-gov.uk/components?name=Page#Dark%20mode-example
But others may have other reasons for doing this sort of thing.)

This seems to have been introduced here: https://github.com/alphagov/govuk-frontend/commit/683591e5980fcc706caf21a9c7fe9bfe46fe85cf

## Steps to reproduce the issue
My setup is a bit complicated but I think the following sass code should do it:
```scss
$govuk-text-colour: inherit;
@import "govuk-frontend/govuk/all";
```

## Thoughts
I'm not sure what can be done if this Safari workaround is deemed important. Are there no other workarounds besides changing the colour?

Would it be possible to detect the values that won't be accepted by `rgba()` and avoid applying the function in that case?

PR #2573, opened 2022-03-13
Better handle cases where govuk-text-colour is set to a non-colour value

daniel-ac-martin:
Was: Reinstate support for 'inherit' as a text colour

The use of the `rgba` function, which was added to work around a bug in
Safari, prevents users from setting `$rgba-text-colour` to inherit.
This is useful if you wish to control the text colour by cascading a
style down from, say, a body tag, which in turn is useful in order to
apply this control at run-time rather than build-time.

This fixes the problem by first testing for the value `inherit`. When
found we simply evaluate to `inherit`, which is probably the best we
can do in the circumstance. When it is not found we apply the `rgba`
trick.

Addresses: #2231

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 ### Fixes
 
 - [#2549: Fix header with product name focus and hover state length](https://github.com/alphagov/govuk-frontend/pull/2549)
-- [#2573: Reinstate support for 'inherit' as a text colour](https://github.com/alphagov/govuk-frontend/pull/2573)
+- [#2573: Better handle cases where govuk-text-colour is set to a non-colour value](https://github.com/alphagov/govuk-frontend/pull/2573)
 
 ## 4.0.1 (Fix release)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,8 @@
 
 ### Fixes
 
-[#2549: Fix header with product name focus and hover state length](https://github.com/alphagov/govuk-frontend/pull/2549)
-[#2573: Reinstate support for 'inherit' as a text colour](https://github.com/alphagov/govuk-frontend/pull/2573)
+- [#2549: Fix header with product name focus and hover state length](https://github.com/alphagov/govuk-frontend/pull/2549)
+- [#2573: Reinstate support for 'inherit' as a text colour](https://github.com/alphagov/govuk-frontend/pull/2573)
 
 ## 4.0.1 (Fix release)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Fixes
 
 [#2549: Fix header with product name focus and hover state length](https://github.com/alphagov/govuk-frontend/pull/2549)
+[#2573: Reinstate support for 'inherit' as a text colour](https://github.com/alphagov/govuk-frontend/pull/2573)
 
 ## 4.0.1 (Fix release)
 
--- a/src/govuk/helpers/_links.scss
+++ b/src/govuk/helpers/_links.scss
@@ -273,7 +273,7 @@
   // Force a colour change on hover to work around a bug in Safari
   // https://bugs.webkit.org/show_bug.cgi?id=224483
   &:hover {
-    color: rgba($govuk-text-colour, .99);
+    color: if($govuk-text-colour == inherit, inherit, rgba($govuk-text-colour, .99));
   }
 
   &:active,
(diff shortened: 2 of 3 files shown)


daniel-ac-martin:
I couldn't find any tests for this mixin so I've tried to add some very simple ones to demonstrate the cases that this PR caters to.

When I ran the tests locally, I would get sporadic failures, due to time outs, in random tests. I'm assuming that is a known issue and is nothing to do with these changes.

src/govuk/helpers/_links.scss
```diff
@@ -273,7 +273,7 @@
   // Force a colour change on hover to work around a bug in Safari
   // https://bugs.webkit.org/show_bug.cgi?id=224483
   &:hover {
-    color: rgba($govuk-text-colour, .99);
+    color: if($govuk-text-colour == inherit, inherit, rgba($govuk-text-colour, .99));
```
36degrees, line 276: Given that the `:hover` state is effectively redundant when `$govuk-text-colour` is `inherit`, I wonder if we should just omit it entirely?

We can also check the type of `$govuk-text-colour` rather than testing for `inherit` specifically. Although I can't think of any other value that we would expect users to set it to, it's a better guard against `rgba()` being called with something that is not a colour.

Thoughts?

```suggestion
    @if (type-of($govuk-text-colour) == color) {
      color: rgba($govuk-text-colour, .99);
    }
```

src/govuk/helpers/links.test.js
```diff
@@ -156,3 +156,41 @@ describe('@mixin govuk-link-hover-decoration', () => {
     })
   })
 })
+
+describe('@mixin govuk-link-style-text', () => {
```
36degrees, line 160: Thanks for including tests 🙌🏻

src/govuk/helpers/_links.scss
daniel-ac-martin: Hi @36degrees.

I think you are right about testing for the type. I did think about whether there were other values we might want to guard against but wasn't sure how to achieve this.

I wasn't sure whether it was safe to omit the state entirely. You're knowledge of the CSS precedence is probably better than mine, so I'll defer to you on this point. (If it can be safely removed then, let's remove it.)

Thanks for the suggestion.

36degrees (commented):
Thanks @daniel-ac-martin – this looks good to me. Are you able to add this to the [list of unreleased fixes in the changelog](https://github.com/alphagov/govuk-frontend/blob/main/CHANGELOG.md)? We can do it, if you prefer.

daniel-ac-martin:
@36degrees: How's that?

CHANGELOG.md
```diff
@@ -4,7 +4,8 @@
 
 ### Fixes
 
-[#2549: Fix header with product name focus and hover state length](https://github.com/alphagov/govuk-frontend/pull/2549)
+- [#2549: Fix header with product name focus and hover state length](https://github.com/alphagov/govuk-frontend/pull/2549)
+- [#2573: Reinstate support for 'inherit' as a text colour](https://github.com/alphagov/govuk-frontend/pull/2573)
```
vanitabarrett, line 8: @daniel-ac-martin 👋 I'm reviewing this as Ollie is on leave. I'm wondering if the Changelog entry and this PR title needs adjusting given the changes we've made? From a look at the change and the comment Ollie left, it sounds like if someone passes 'inherit', nothing is output (feel free to correct me if I've read that wrong!). I'm wondering if it's therefore misleading to say we've "reinstated support for inherit". What do you think?

daniel-ac-martin: Hi @vanitabarrett.

The wording was from the perspective of, "Can I set `$govuk-text-colour` to `inherit` and have it work?"
In the past, this was possible but was later broken (the sass wouldn't build) when a the `rgba` function was introduced to work around an issue with Safari.

I think the title is okay, as `inherit` hasn't been working since the safari workaround.
But I'm more than happy to change it, if you can suggest a better one.

vanitabarrett: Hi @daniel-ac-martin , I've spoken with our technical writer and we're wondering about this instead:

`Better handle cases where govuk-text-colour is set to a non-colour value`

How does that sound?

daniel-ac-martin: @vanitabarrett: Sounds good to me, I'll make the change...

daniel-ac-martin: (@vanitabarrett: I've just pushed the change.)

vanitabarrett (approved):

36degrees (approved):

daniel-ac-martin:
@vanitabarrett: Any chance of a (patch?) release for this?

vanitabarrett:
@daniel-ac-martin the Design System team are working on a couple of things at the moment which might be ready for release soon, so we're going to see how they progress. It may make sense for us to wait and include them in a feature release in the next couple of weeks.

Issue #2239, opened 2021-06-01
Migrate GOV.UK Frontend build pipeline to use Dart Sass


36degrees:
## What

Migrate GOV.UK Frontend's tooling to use Dart Sass rather than LibSass.

**This does not include migrating to the module system or dropping support for LibSass or Ruby Sass – we'll continue to test against and support all three compilers for the time being.**

This only affects which compiler we use internally to generate the stylesheets used in the review app, and when compiling minified stylesheets for 'dist'.

## Why

Dart Sass is the only one of the three compilers we support that is not deprecated. Increasingly, new projects should and likely will be using Dart Sass.

Changes are being made to Dart Sass including deprecations (like #2238) which may cause warnings and other issues for our users. Using Dart Sass as our primary compiler will help us to identify these issues earlier.

## Who needs to know about this

Developers

## Done when

- [ ] GOV.UK Frontend build pipeline updated to use Dart Sass rather than LibSass

querkmachine:
We may want to re-evaluate priority on this as we're starting to run into compatibility issues running node-sass with Node 16 on Apple Silicon Macs (#2574).

node-sass has apparently been updated to work on Node 16, though some commenters seem to dispute the efficacy of those fixes (sass/node-sass#3077, sass/node-sass#3090). It does not have any support for arm64 architecture as of writing (open issue sass/node-sass#3033), and this is likely not a priority given the package has been deprecated.

PR #2574, opened 2022-03-15
Update Node to Active LTS version v16 (Gallium)

### LINKED ISSUE #2504: Update Node.js to active LTS version
## What

Update the project to use the latest active Long Term Support (LTS) version of Node.js for development, testing, and deployment.

## Why

This project currently uses Node.js version 14, which has been in maintenance since last October (2021-10-19) and won't receive new features.

## Who needs to work on this

Developer

## Who needs to review this

Developer, Tech lead

## Done when

- [ ] Update `.nvmrc`
- [ ] Update `package-lock.json` to use lockfile v2

domoscargin:
Node.js v14 has been in maintenance mode from 19 October 2021. v16 is the active LTS version.

This in turns updates NPM to v8, meaning our lockfile has changed to [`lockfileVersion` 2](https://docs.npmjs.com/cli/v8/configuring-npm/package-lock-json#lockfileversion)

This is an internal change that affects development, testing and deployment.

Closes #2504


domoscargin:
> There are no plans to support M1 with the node-sass compiler. You can switch to the sass compiler instead.

https://github.com/dlmanning/gulp-sass/issues/803#issuecomment-869580225

domoscargin:
I got this working locally by running:

`npm uninstall node-sass -g && npm cache clean -force && npm install node-sass`

36degrees:
I have a new M1 mac so tried switching to this branch immediately after cloning the repo, installing and using 16.14.0 _without_ having installed or run GOV.UK Frontend under v14, and was able to install and start without any issues. So wondering if this is 'only' an issue for users migrating from 14 to 16.

36degrees:
@vanitabarrett Yes, I think so.

domoscargin:
In that case, I can add some troubleshooting documentation here: https://github.com/alphagov/govuk-frontend/blob/main/docs/contributing/running-locally.md

Though this does seem like a pretty temporary problem that affects a small group, so we'd probably want to review it down the line.

EoinShaughnessy:
@domoscargin Hi!

Yeah, I think adding this content is probably a good idea - should help some users, even if only a few.

Do we have an idea of when this issue will no longer affect users?

Also, what sort of problems might they encounter? Would giving an example help users confirm the issue is to do with using Node v14 or earlier?

domoscargin:
> Do we have an idea of when this issue will no longer affect users?
> 
> Also, what sort of problems might they encounter? Would giving an example help users confirm the issue is to do with using Node v14 or earlier?

This would stop affecting users once everybody who's installed `govuk-frontend` locally on their M1 with Node v14 has successfully moved to Node v16. This only affects devs/contributors, since most users don't need to install this repo locally. 

Once this change is in place, anybody who installs locally for the first time should have no problems. Additionally, if we migrate our build process to Dart Sass (#2239), this problem should go away for anybody who installs after that change. So it's probably a small(ish) pool of users, but with a potentially long tail depending on when they update.

The problem is specifically with the `node-sass` package. Users will get `node-sass` related errors when trying to run `npm install` or `npm start` (https://github.com/alphagov/govuk-design-system/pull/2110#issuecomment-1072333844)

EoinShaughnessy:
@domoscargin Great explanation, thanks!

Content looks really good! Just suggesting a few tweaks*:

```
### Fixing update errors
If you've previously installed `govuk-frontend` locally using Node v14 or earlier, you may see `node-sass`-related errors when updating to Node v16.

To get rid of these errors, delete the `node_modules` folder from the `node-sass` package, then run:

nvm use
npm uninstall node-sass -g && npm cache clean -force && npm install node-sass
npm install
```

* These might be based on my misunderstanding stuff - if so, sorry!

lfdebrux (approved):
1️⃣ 6️⃣ 🎉

Issue #2539, opened 2022-02-09
Investigate how assistive technologies interact with character count 


davidc-gds:
## What
We're looking to investigate how screen reader assistive technologies interact with the character count component.

https://design-system.service.gov.uk/components/character-count/
https://govuk-frontend-review.herokuapp.com/components/character-count

The investigation will focus primarily on gathering information to help resolve four related bugs, listed below.

### Related bug tickets
This investigation will help us understand the bugs listed in the four tickets below.

- https://github.com/alphagov/govuk-frontend/issues/2485
- https://github.com/alphagov/govuk-frontend/issues/2486
- https://github.com/alphagov/govuk-frontend/issues/2487
- https://github.com/alphagov/govuk-frontend/issues/2488

## Why
The character count component was initially designed to play well with screen readers, but sometime after the launch of the component something changed. Now there are four bugs that make interacting with the character count component via 
 a screen reader frustrating and cumbersome.

This might be an accessibility compliance issue, but more investigation would be needed to confirm. For now, we're approaching it as a severe usability issue and poor experience for those using a screen reader (and potentially other assistive technologies as well).

## Who needs to work on this

- @davidc-gds will handle the accessibility portions of the investigation and test with various browser and assistive technology combinations.
- @owenatgov has offered to support from a developer perspective if capacity allows, since the character count component includes some JavaScript.

## Who needs to review this
Unknown

## Done when

- [ ] Character count component variants are tested on various screen readers and browsers
- [ ] Research is compiled into a summary
- [ ] All 4 identified bugs have been explored
- [ ] Action items and approaches are identified and proposed for each of the four identified bugs

36degrees:
@davidc-gds @querkmachine is there anything else to do here or can we call this one done?

querkmachine:
@36degrees We're still waiting on the guidance in the design system to be updated to reflect the changes made. We kept this issue open to accommodate for that but having a new issue in that repo might make more sense? @davidc-gds

vanitabarrett:
@querkmachine A new issue in the Design System repo would make sense to me - we could attach it to the v4.1.0 milestone then too

davidc-gds:
I've set up the new issue for the guidance content updates. Feel free to edit or modify it if I've built it incorrectly:
https://github.com/alphagov/govuk-design-system/issues/2155

@querkmachine would this mean that this issue can now be closed, and we'll hop over to that new issue?

querkmachine:
Sounds like a plan!

Issue #2488, opened 2021-12-22
Character count messages are announced by screen readers even when the count is below a set threshold


36degrees:
## Description of the issue

The character count has a ['threshold' feature](https://design-system.service.gov.uk/components/character-count/#avoid-narrow-limits) which allows you to hide the remaining count unless the count reaches a set percentage of the total limit of characters or words.

However, some screen readers announce the count message even when it is hidden.

## Steps to reproduce the issue

1. Go to https://govuk-frontend-review.herokuapp.com/components/character-count/with-threshold/preview
2. Enter a single character in the box – the count message should only appear after 7 characters.

## Actual vs expected behaviour

Expected behaviour: The count message should not be announced, as it is not visible.

Actual behaviour: The count message is announced in:

- JAWS 2021.2103.174 / Chrome 96
- JAWS 2021.2103.174 / IE 11
- NVDA 2021.2 / Chrome 96

It seems to be announced very rarely in VoiceOver / Safari 15.2 (macOS Big Sur).

Firefox behaves as expected with both NVDA and JAWS.

There is also something very odd going on in VoiceOver / Safari (iOS 15.2) where all of the information about the field is announced on every key press. This needs further testing – it might just be an issue with the way I was testing it, or it could be a bug in WebKit that may or may not be related to what we're doing with the character count…

## Environment (where applicable)

- Operating system: As above
- Browser: As above
- Browser version: As above
- GOV.UK Frontend Version: v4.0.0

Issue #2487, opened 2021-12-22
Multiple outdated announcements can be made when the user stops typing into the character count.


36degrees:
## Description of the issue

Multiple count messages can get 'queued up' and announced when the user stops typing into the character count. Some of these messages will be outdated, and do not reflect the latest 'count'.

https://user-images.githubusercontent.com/121939/147126319-a96b7f62-bc78-49a9-be23-42f032045d6d.mov

## Steps to reproduce the issue

1. Go to https://govuk-frontend-review.herokuapp.com/components/character-count/preview
2. Type multiple characters into the box quickly, then pause

## Actual vs expected behaviour

Expected behaviour: Only one count message should be announced, and it should reflect the current number of remaining characters in the textarea.

Actual behaviour: Multiple count messages are announced in:

- JAWS 2021.2103.174 / Chrome 96
- JAWS 2021.2103.174 / Firefox 95
- NVDA 2021.2 / Chrome 96

This also happens occasionally in JAWS 2021.2103.174 / IE 11 and with VoiceOver / Safari 15.2 (macOS Big Sur).

It did not seem to occur with NVDA 2021.2 / Firefox 94 or VoiceOver / Safari (iOS 15.2).

## Environment (where applicable)

- Operating system: As above
- Browser: As above
- Browser version: As above
- GOV.UK Frontend Version: v4.0.0

36degrees:
I think this is also related to, and compounded by, #2485.

querkmachine:
I've looked at this as part of #2539 and made some attempts to debounce the update of the counter until after the user has stopped typing for a short period of time. 

Unfortunately this is being undermined by the bugfix we have in place for Dragon Naturally Speaking, which forces an update to happen every one second if the textarea value has been changed. This plus debouncing does reduce the number of announcements being queued up, but it doesn't resolve it completely or prevent stuttering in the readout.

36degrees:
> Unfortunately this is being undermined by the bugfix we have in place for Dragon Naturally Speaking, which forces an update to happen every one second if the textarea value has been changed. This plus debouncing does reduce the number of announcements being queued up, but it doesn't resolve it completely or prevent stuttering in the readout.

Could we keep a timestamp of the last time the count message was updated because of a key event, and then only trigger an update from the one second interval if that timestamp is more than one second ago? That way the interval timer should be a no-op if the keyboard events are firing as you'd expect when Dragon isn't involved… I think?

Issue #2486, opened 2021-12-22
Hint text for a character count is included by screen readers as part of the count message


36degrees:
## Description of the issue

If a character count also has a hint associated with the textarea, some screen readers also include the hint text whenever the count message is announced.

## Steps to reproduce the issue

1. Go to https://govuk-frontend-review.herokuapp.com/components/character-count/with-hint/preview
2. Focus the textarea and wait for the accessible name, type and description to be read
3. Enter characters in the text area

## Actual vs expected behaviour

Expected: Only the count message should be announced, for example "You have 9 characters remaining."

Actual: The hint text is also announced in:

- JAWS 2021.2103.174 / Chrome 96
- JAWS 2021.2103.174 / Firefox 95
- NVDA 2021.2 / Chrome 96
- NVDA 2021.2 / Firefox 94

"You have 5 characters remaining. Don't include personal or financial information, eg your National Insurance Number or credit card details."

This hint text was also announced _very occasionally_ in VoiceOver / Safari 15.2 (macOS Big Sur) with no discernible pattern.

JAWS 2021.2103.174 / IE11 and VoiceOver / Safari (iOS 15.2) behave as expected, announcing only the remaining count.

## Environment (where applicable)

- Operating system: As above
- Browser: As above
- Browser version: As above
- GOV.UK Frontend Version: v4.0.0

36degrees:
I believe this is likely because changes to the accessible description are announced a little like live regions (the character count message is associated with the textarea using `aria-describedby`).

See also https://www.davidmacd.com/blog/test-aria-describedby-errormessage-aria-live.html

Some overlap with #2485.

PR #2577, opened 2022-03-21
Refactor character count to inject new element

querkmachine:
Modifies the character count JavaScript to compose and inject new elements that are live-updated, instead of overwriting the content of the initial HTML hint.

There are now three elements responsible for character count announcements:

- A static element that describes the character limit (e.g. "You can use up to 200 characters"). 
  - This is present in HTML, and serves as the fallback if JavaScript is unavailable. 
  - It is visible if JavaScript is unavailable, and visually hidden if it is.
  - It is announced by screen readers when the user first focuses on the textarea, before they type. 
  - Screen readers will always read this out, even if a counter utilises a threshold.
- An injected, live-updating element showing the characters remaining (e.g. "You have 200 characters remaining"). 
  - This is visible, updates immediately in response to user input, and uses `aria-hidden` to hide itself from screen readers.
  - If the counter has a threshold, this is invisible until after the user has passed the threshold. 
- An injected, debounced element showing the characters remaining (e.g. "You have 200 characters remaining"). 
  - This is invisible, only updates after one second of inactivity, and is announced by screen readers following input using `aria-live="polite"`.
  - If the counter has a threshold, this is only announced once the user has passed the threshold. 

Only the first element must be present in the initial HTML. The other two are injected when JavaScript runs. 

## Upgrading

Users of the govuk-frontend Nunjucks macros do not have to make any changes. 

Users of static HTML can remove the hardcoded `aria-live="polite"` attribute from `.govuk-character-count__message`. We haven't found any negative repercussions to it still being present, but it is not needed as this element's content no longer updates. 

For backwards compatibility, custom classes that have been added to the HTML count element are copied onto the visible, injected counter. If you previously used a custom class to change the visibility of the counter outside of the existing threshold functionality, this may no longer work as expected. The screen reader only counter does not inherit any classes. 

## Why these changes have been made

These changes are intended to resolve a number of issues that came up in a recent investigation into how the character count interacted with more recent versions of screen readers: #2539 

Should resolve:
- #2485
- #2486
- #2488 
- #2487 

## Current known issues

- JAWS 2021.2103.174 / IE 11 will always read the character count, even if the input length is below a set threshold. 
- JAWS 2021.2103.174 / Chrome 99 will refuse to announce input hints if the input has a prefilled value. This is not an issue introduced by this PR, but was found during testing for it. Raised as a new issue: #2585.

### DIFF
--- a/src/govuk/components/character-count/character-count.js
+++ b/src/govuk/components/character-count/character-count.js
@@ -6,7 +6,7 @@ function CharacterCount ($module) {
   this.$module = $module
   this.$textarea = $module.querySelector('.govuk-js-character-count')
   if (this.$textarea) {
-    this.$countMessage = document.getElementById(this.$textarea.id + '-info')
+    this.$fallbackLimitMessage = document.getElementById(this.$textarea.id + '-info')
   }
 }
 
@@ -20,15 +20,27 @@ CharacterCount.prototype.init = function () {
   // Check for module
   var $module = this.$module
   var $textarea = this.$textarea
-  var $countMessage = this.$countMessage
+  var $fallbackLimitMessage = this.$fallbackLimitMessage
 
-  if (!$textarea || !$countMessage) {
+  if (!$textarea) {
     return
   }
 
-  // We move count message right after the field
+  // We move fallback count message right after the field
   // Kept for backwards compatibility
-  $textarea.insertAdjacentElement('afterend', $countMessage)
+  $textarea.insertAdjacentElement('afterend', $fallbackLimitMessage)
+
+  // Create our live-updating counter element, copying the classes from the
+  // fallback element for backwards compatibility as these may have been configured
+  var $countMessage = document.createElement('div')
+  $countMessage.className = $fallbackLimitMessage.className
+  $countMessage.classList.add('govuk-character-count__status')
+  $countMessage.setAttribute('aria-live', 'polite')
+  this.$countMessage = $countMessage
+  $fallbackLimitMessage.insertAdjacentElement('afterend', $countMessage)
+
+  // Hide the fallback limit message
+  $fallbackLimitMessage.classList.add('govuk-visually-hidden')
 
   // Read options set using dataset ('data-' values)
   this.options = this.getDataset($module)
--- a/src/govuk/components/character-count/character-count.js
+++ b/src/govuk/components/character-count/character-count.js
@@ -6,6 +6,8 @@ function CharacterCount ($module) {
   this.$module = $module
   this.$textarea = $module.querySelector('.govuk-js-character-count')
   this.$countMessage = null
+  this.lastInputTimestamp = null
+  this.debouncedInputTimer = null
 }
 
 CharacterCount.prototype.defaults = {
@@ -60,21 +62,17 @@ CharacterCount.prototype.init = function () {
   // Remove hard limit if set
   $module.removeAttribute('maxlength')
 
+  this.bindChangeEvents()
+
   // When the page is restored after navigating 'back' in some browsers the
   // state of the character count is not restored until *after* the DOMContentLoaded
-  // event is fired, so we need to sync after the pageshow event in browsers
-  // that support it.
+  // event is fired, so we need to manually update it after the pageshow event
+  // in browsers that support it.
   if ('onpageshow' in window) {
-    window.addEventListener('pageshow', this.sync.bind(this))
+    window.addEventListener('pageshow', this.updateCountMessage.bind(this))
   } else {
-    window.addEventListener('DOMContentLoaded', this.sync.bind(this))
+    window.addEventListener('DOMContentLoaded', this.updateCountMessage.bind(this))
   }
-
-  this.sync()
-}
-
-CharacterCount.prototype.sync = function () {
-  this.bindChangeEvents()
   this.updateCountMessage()
 }
 
@@ -109,7 +107,7 @@ CharacterCount.prototype.count = function (text) {
 // Bind input propertychange to the elements and update based on the change
 CharacterCount.prototype.bindChangeEvents = function () {
   var $textarea = this.$textarea
-  $textarea.addEventListener('keyup', this.checkIfValueChanged.bind(this))
+  $textarea.addEventListener('keyup', this.handleKeyUp.bind(this))
 
   // Bind focus/blur events to start/stop polling
   $textarea.addEventListener('focus', this.handleFocus.bind(this))
@@ -177,9 +175,25 @@ CharacterCount.prototype.updateCountMessage = function () {
   countMessage.innerHTML = 'You have ' + displayNumber + ' ' + charNoun + ' ' + charVerb
 }
 
+// Debounce updating the character counter until after a user has stopped typing
+// for a short period of time. This helps prevent screen readers from queuing up
+// multiple text updates in rapid succession.
+CharacterCount.prototype.handleKeyUp = function () {
+  this.lastInputTimestamp = Date.now()
+  clearTimeout(this.debouncedInputTimer)
+  this.debouncedInputTimer = setTimeout(this.updateCountMessage.bind(this), 250)
+}
+
 CharacterCount.prototype.handleFocus = function () {
-  // Check if value changed on focus
-  this.valueChecker = setInterval(this.checkIfValueChanged.bind(this), 1000)
+  // If the field is focused, and a keyup event hasn't been detected for at
+  // least 1000 ms (1 second), then run the manual change check.
+  // This is so that the update triggered by the manual comparison doesn't
+  // conflict with debounced KeyboardEvent updates.
+  this.valueChecker = setInterval(function () {
+    if (!this.lastInputTimestamp || (Date.now() - 1000) >= this.lastInputTimestamp) {
+      this.checkIfValueChanged.bind(this)
+    }
+  }.bind(this), 1000)
 }
 
 CharacterCount.prototype.handleBlur = function () {
--- a/src/govuk/components/character-count/character-count.js
+++ b/src/govuk/components/character-count/character-count.js
@@ -167,7 +167,7 @@ CharacterCount.prototype.updateVisibleCountMessage = function () {
   }
 
   // Update message
-  $visibleCountMessage.innerHTML = this.formatUpdateMessage()
+  $visibleCountMessage.innerHTML = this.formattedUpdateMessage()
 }
 
 // Update screen reader-specific counter
@@ -183,11 +183,11 @@ CharacterCount.prototype.updateScreenReaderCountMessage = function () {
   }
 
   // Update message
-  $screenReaderCountMessage.innerHTML = this.formatUpdateMessage()
+  $screenReaderCountMessage.innerHTML = this.formattedUpdateMessage()
 }
 
 // Format update message
-CharacterCount.prototype.formatUpdateMessage = function () {
+CharacterCount.prototype.formattedUpdateMessage = function () {
   var $textarea = this.$textarea
   var options = this.options
   var remainingNumber = this.maxLength - this.count($textarea.value)
--- a/src/govuk/components/character-count/character-count.js
+++ b/src/govuk/components/character-count/character-count.js
@@ -5,9 +5,7 @@ import '../../vendor/polyfills/Element/prototype/classList'
 function CharacterCount ($module) {
   this.$module = $module
   this.$textarea = $module.querySelector('.govuk-js-character-count')
-  if (this.$textarea) {
-    this.$fallbackLimitMessage = document.getElementById(this.$textarea.id + '-info')
-  }
+  this.$countMessage = null
 }
 
 CharacterCount.prototype.defaults = {
@@ -20,7 +18,7 @@ CharacterCount.prototype.init = function () {
   // Check for module
   var $module = this.$module
   var $textarea = this.$textarea
-  var $fallbackLimitMessage = this.$fallbackLimitMessage
+  var $fallbackLimitMessage = document.getElementById(this.$textarea.id + '-info')
 
   if (!$textarea) {
     return
--- a/src/govuk/components/character-count/character-count.js
+++ b/src/govuk/components/character-count/character-count.js
@@ -17,16 +17,17 @@ CharacterCount.prototype.defaults = {
 
 // Initialize component
 CharacterCount.prototype.init = function () {
-  // Check for module
-  var $module = this.$module
-  var $textarea = this.$textarea
-  var $fallbackLimitMessage = document.getElementById(this.$textarea.id + '-info')
-
-  if (!$textarea) {
+  // Check that required elements are present
+  if (!this.$textarea) {
     return
   }
 
-  // We move fallback count message right after the field
+  // Check for module
+  var $module = this.$module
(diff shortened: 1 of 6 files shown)


querkmachine:
Regarding #2487:

My intended approach for this was to debounce the update of the character counter text until after the user had stopped typing for a period of time (like 300 milliseconds). 

However, this approach conflicts with the existing Dragon Naturally Speaking bugfix; which runs a `setInterval` loop every one second and pushes an update to the counter if the textarea's value has been changed. This undermines the debouncing approach, as although the character counter is updating less frequently, it is still being updated, and thus queuing things to read out. It reduces the problem, but it doesn't remove it completely. 

Debouncing the character counter update also makes for a bit more of a jarring experience for sighted users, as the counter text will appear to "lag" behind their typing.

src/govuk/components/character-count/character-count.js
```diff
@@ -20,15 +20,27 @@ CharacterCount.prototype.init = function () {
   // Check for module
   var $module = this.$module
   var $textarea = this.$textarea
-  var $countMessage = this.$countMessage
+  var $fallbackLimitMessage = this.$fallbackLimitMessage
```
owenatgov, line 23: Is this declaration necessary? You could replace the instances of `$fallbackLimitMessage` below with `this.$fallbackLimitMessage`. It doesn't look like this is being used anywhere else in the page so the risk of mutating the var right in the `this` is pretty low as far as I can tell.

You could go one step further and remove the `if` from the module declaration on line 9 and only scope the variable to the `init` function.

src/govuk/components/character-count/character-count.js
```diff
@@ -6,7 +6,7 @@ function CharacterCount ($module) {
   this.$module = $module
   this.$textarea = $module.querySelector('.govuk-js-character-count')
   if (this.$textarea) {
-    this.$countMessage = document.getElementById(this.$textarea.id + '-info')
```
owenatgov, line 9: Is it worth adding a rudimentary `this.$countMessage = null` here? This would primarily be a readability thing as otherwise the declaration is hidden on line 39 which might be a hassle for devs to find if they're trying to trace how this module works. Not a blocker.

owenatgov (commented):
Just had a look at the code and left 2 little comments. Looking sharp 👍🏻 

I'd also add to consider squashing your commits either now or once it's out of draft as they're close enough to sit under a single commit and we're otherwise knowingly introducing a test failure into our commit history.

owenatgov:
@querkmachine It's. tricky one. I generally play it by ear depending on how many "things" the code I'm writing includes. Things in this context being changes to the code that we could reliably ship, big or small, even if they're part of a wider bit of work so it makes sense to chunk them into one PR.

In this case my advice would be if you end up adding anything to this that isn't covered by the original commit message, make it a new commit. I think how you've split it out so far makes sense.

querkmachine:
Again regarding https://github.com/alphagov/govuk-frontend/issues/2487, I've considered a couple of approaches to trying to fix this: 

- Debouncing the text update.
  - See https://github.com/alphagov/govuk-frontend/pull/2577#issuecomment-1074998075
- Adding a second, screen reader-only counter element with debounced text in it.
  - David and I think this might be verging on overkill, and isn't something we want to do if we can help it.
- Adding debounced text to an `aria-label` on the counter, in an effort to see if screen readers will opt to read that over the innerText.
  - It does not. In fact, VoiceOver doesn't read out the counter at all if there's an `aria-label` on it. Weird!
- Entirely remove the functionality that updates the counter on keyboard input, and rely entirely on interval-based checks. 
  - This feels like a poor practice to follow, and makes the counter inherently less responsive to user input. 

I'm open to any other approaches that people come up with

querkmachine:
Testing this PR in the AT/browser combinations these issues were initially identified in. Appears to be fixed in:

- #2485
  - [x] JAWS 2021.2103.174 / Chrome 96
  - [x] JAWS 2021.2103.174 / Firefox 95
  - [x] NVDA 2021.2 / Chrome 96
  - [x] NVDA 2021.2 / Firefox 94
- #2486
  - [x] JAWS 2021.2103.174 / Chrome 96
  - [x] JAWS 2021.2103.174 / Firefox 95
  - [x] NVDA 2021.2 / Chrome 96
  - [x] NVDA 2021.2 / Firefox 94
- #2488
  - [x] JAWS 2021.2103.174 / Chrome 96
  - [x] JAWS 2021.2103.174 / IE 11
  - [x] NVDA 2021.2 / Chrome 96

All of the above have also been tested and working in VoiceOver/Firefox 98, which is the combination I used while devtesting.

querkmachine:
I've applied the suggestion from @36degrees (here: https://github.com/alphagov/govuk-frontend/issues/2487#issuecomment-1084283243) for debouncing the character counter update. The counter is now not updated until 250 milliseconds after the user has stopped typing, which (mostly) successfully stops screen readers from queuing up multiple updates or attempting to read them whilst a user is still typing. 

The delay before the update is made may still need tweaking based on testing. There is a need to strike a balance between immediacy of feedback for users of the visual counter and that of screen reader users. 

The debounced nature of the counter now means that the tests fail too. Gotta fix that.

querkmachine:
Have discussed some of the issues that came up in testing and we have decided that none of these are blockers (they are all equivalent to or improvements on the existing component, despite the flaws. Having these be documented as known issues would be useful. 

The JAWS/Chrome issue with textareas with default values has been opened as #2585.

src/govuk/components/character-count/template.njk
```diff
@@ -29,9 +29,6 @@
   {{ govukHint({
     text: 'You can enter up to ' + (params.maxlength or params.maxwords) + (' words' if params.maxwords else ' characters'),
     id: params.id + '-info',
-    classes: 'govuk-character-count__message' + (' ' + params.countMessage.classes if params.countMessage.classes),
-    attributes: {
-      'aria-live': 'polite'
-    }
+    classes: 'govuk-character-count__message' + (' ' + params.countMessage.classes if params.countMessage.classes)
```
vanitabarrett, line 32: @querkmachine Do you know what would happen if someone updated to the latest version of govuk-frontend with this change, but failed to update their HTML (assuming they're using HTML, not Nunjucks)? Following on from what David said in the planning meeting, I'm wondering if this change makes it a breaking change.

querkmachine: My assumption has been that nothing will happen and the attribute is just redundant, as the inner text of this element is never updated, and thus `aria-live` is never 'activated'. Happy to do a quick whip around to test that assumption, though.

querkmachine: Having quickly ran through JAWS/NVDA/VoiceOver in a couple of browsers each, there doesn't seem to be any side effect to the `aria-live` attribute being present.

As the attribute was previously included in HTML even if the character count was never initialised, I don't think that it still being present in an upgrading service represents a breaking change and is more akin to a recommended change (as `tabindex` was for error summaries in the last release).

edwardhorsford:
> I've applied the suggestion from @36degrees (here: [#2487 (comment)](https://github.com/alphagov/govuk-frontend/issues/2487#issuecomment-1084283243)) for debouncing the character counter update. The counter is now not updated until 250 milliseconds after the user has stopped typing, which (mostly) successfully stops screen readers from queuing up multiple updates or attempting to read them whilst a user is still typing.
> 
> The delay before the update is made may still need tweaking based on testing. There is a need to strike a balance between immediacy of feedback for users of the visual counter and that of screen reader users.
> 
> The debounced nature of the counter now means that the tests fail too. Gotta fix that.

@querkmachine do the visual updates and screen reader updates need to be tied together? Perhaps they should just be two independent elements? One hidden from screen readers, the other visually hidden.

querkmachine:
> @querkmachine do the visual updates and screen reader updates need to be tied together? Perhaps they should just be two independent elements? One hidden from screen readers, the other visually hidden.

@edwardhorsford They don't have to be tied, strictly speaking. I discussed the idea previously with David, our accessibility specialist (mentioned [in this comment](https://github.com/alphagov/govuk-frontend/pull/2577#issuecomment-1079181501)), but we were worried that having multiple updating elements—some visible and others not, some debounced and others not—would be making the component JS more complex than is strictly necessary.

The change to how the visual counter updates hasn't undergone any user testing, and we were considering splitting it out if we felt it was necessary. Do you feel like it might?

src/govuk/components/character-count/template.njk
vanitabarrett: Brill, thanks for checking that! 👍

edwardhorsford:
@querkmachine is there a review app I could try? My worry is it might impact faster typists - as 250ms would be quite a while for and you might end up with an extended period without update.

querkmachine:
@edwardhorsford Yep! You can demo it here: https://govuk-frontend-pr-2577.herokuapp.com/components/character-count

vanitabarrett:
@querkmachine I've not had a chance to review the code yet, but agree that the slow visual updating doesn't feel... great. I  think it's quite noticeable and if we release like that we're probably going to see an increase in users (i.e: service teams and end users) asking why the component feels 'slow' or 'hangs'. Shall we chat about the options in dev catch-up this morning?

querkmachine:
I've updated the component to have two live counters. This approach seems to work in the widest array of browser and screen reader combinations, whilst not degrading the experience for sighted users. 

There are now three elements responsible for character count announcements:
- A static element that describes the character limit (e.g. "You can use up to 200 characters"). This is present in HTML, and serves as the fallback if JavaScript is unavailable. It is announced by screen readers when the user first focuses on the textarea, before they type. 
  - Screen readers will read this out even if a counter utilises a threshold.
- A live element showing the characters remaining (e.g. "You have 200 characters remaining"). This is visible, updates immediately in response to user input, and is hidden to screen readers. 
- A debounced element showing the characters remaining (e.g. "You have 200 characters remaining"). This is invisible, only updates after one second of inactivity, and is announced by screen readers following input.

---

There appears to be a conflict with some part of the default VoiceOver functionality, where VO overrules the character count announcement with its own announcement of the last word that was written. For example, if a user is typing "hello world", VoiceOver will announce something like this:

> H-E-L-L-O-space-W-O-R-L-D. You have — world. 

I'm not sure what can be done about this, as this 'recap' is a feature of VoiceOver, so we don't control if or when it happens. Neither JAWS nor NVDA have this issue, as neither of them repeat the user input in this way.

Fenwick17:
When using dictation to type into the textarea on iOS 15 I have noticed an issue with how the character counter is updated.  
The number value correctly updates as dictation inserts characters, but there is no audio update to the user. 
After completing the dictation it will read out the text area including the text it has inserted, but will still read out the original value for character limit. 'You have 10 characters remaining' even if the words on the screen say 'You have 11 characters too many'

Swiping through with VoiceOver gives the correct value, but the automatic VoiceOver/dictation does not.

src/govuk/components/character-count/character-count.js
```diff
@@ -153,6 +166,31 @@ CharacterCount.prototype.updateCountMessage = function () {
...
 
   // Update message
+  countMessage.innerHTML = this.formatUpdateMessage()
+}
+
+// Update screen reader-specific counter
+CharacterCount.prototype.updateScreenReaderCountMessage = function () {
+  var countMessage = this.$screenReaderCountMessage
+
+  // If other the threshold, remove the aria-hidden attribute, allowing screen
```
vanitabarrett, line 176:
```suggestion
  // If over the threshold, remove the aria-hidden attribute, allowing screen
```

src/govuk/components/character-count/character-count.js
```diff
@@ -153,6 +166,31 @@ CharacterCount.prototype.updateCountMessage = function () {
...
+    countMessage.setAttribute('aria-hidden', true)
+  }
+
+  // Update message
+  countMessage.innerHTML = this.formatUpdateMessage()
+}
+
+// Format update message
+CharacterCount.prototype.formatUpdateMessage = function () {
+  var countElement = this.$textarea
```
vanitabarrett, line 190: Looks like this is something that's been around for a while and i wonder if this is a good time to update the name of this variable - not sure why it needs to be referred to as 'countElement' when we refer to it elsewhere as 'textArea'?  Any thoughts?

src/govuk/components/character-count/character-count.js
```diff
@@ -153,6 +166,31 @@ CharacterCount.prototype.updateCountMessage = function () {
...
+  } else {
+    countMessage.setAttribute('aria-hidden', true)
+  }
+
+  // Update message
+  countMessage.innerHTML = this.formatUpdateMessage()
+}
+
+// Format update message
+CharacterCount.prototype.formatUpdateMessage = function () {
```
vanitabarrett, line 189: Really nitpicky, but I wonder if this function name could be clearer. At the moment, it kind of sounds like we're passing it a bunch of things which it's formatting. Maybe something like `formattedUpdateMessage` is clearer?

vanitabarrett (dismissed):
Some minor comments, mostly about improving the readability of the javascript

src/govuk/components/character-count/character-count.js
```diff
@@ -20,15 +21,35 @@ CharacterCount.prototype.init = function () {
...
+  // fallback element for backwards compatibility as these may have been configured
+  var $visibleCountMessage = document.createElement('div')
+  $visibleCountMessage.className = $fallbackLimitMessage.className
+  $visibleCountMessage.classList.add('govuk-character-count__status')
+  $visibleCountMessage.setAttribute('aria-hidden', 'true')
+  this.$visibleCountMessage = $visibleCountMessage
+  $fallbackLimitMessage.insertAdjacentElement('afterend', $visibleCountMessage)
+
+  // Hide the fallback limit message
+  $fallbackLimitMessage.classList.add('govuk-visually-hidden')
```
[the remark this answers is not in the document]
querkmachine: I considered trying to genericise and abstract the element creation into a new function, but they only have three lines of code in common between them so there didn't seem to be much to gain in doing so.

src/govuk/components/character-count/character-count.js
querkmachine: Agree. I also noticed that some variables referencing elements are prefixed with `$` while others aren't, so I did a general whip around to try and make the formatting more consistent.

src/govuk/components/character-count/character-count.js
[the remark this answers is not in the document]
vanitabarrett: As discussed with @querkmachine on Slack, we're going to leave this as-is for now. My initial comment was around splitting this code into separate smaller functions rather than trying to write one generic function. If we get to a point where we can write proper unit tests for our JavaScript code, then smaller functions are easier to test. However, @querkmachine made the good point that it may be harder to understand if we split the logic out of execution order, so we're going to leave it like this for now and consider refactoring later if we need to.

vanitabarrett:
@querkmachine It would be good if we could get this tested in Android and iOS and record it on that spreadsheet. You might need to ask around for someone with an Android device to test 🤔

querkmachine:
Have updated testing spreadsheet with results for VoiceOver / iOS 15.3 and Talkback / Android 11.

edwardhorsford:
Checking the latest review app, the responsiveness (for a fast typist) is looking much better.

vanitabarrett (approved):
Amazing work, this looks really good 🎉

Issue #1791, opened 2020-04-28
Migrate to Sass module system (@use)


36degrees:
## What

Update GOV.UK Frontend to use the new `@use` syntax internally to import dependencies, and to be consumed using `@use`.

## Why

The Sass [module system](https://sass-lang.com/blog/the-module-system-is-launched) has been introduced, aiming to address some of the issues with the `@import` rule.

The `@import` rule, and most global functions will eventually be deprecated, and then removed from Sass.

Moving to `@use` will allow us to revisit and simplify the way that GOV.UK Frontend is implemented, without the limitations of `@import – for example namespacing, and the [govuk-exports](https://github.com/alphagov/govuk-frontend/blob/42742381f5f231b6d6a9962f6baddfcf15710efb/src/govuk/tools/_exports.scss) mixin.

## Depends on

Stable support for [`@uses` in LibSass](https://github.com/sass/libsass/issues/2807) – which probably means a stable version with support having been out for 3-6 months.

(The ['future plans' section of the module system blog post](https://sass-lang.com/blog/the-module-system-is-launched#future-plans) sets a hard limit of October 2021 for deprecating `@import`, which we could also do, but we'd need to have a better understanding of the impact it'll have on our users.)

## Further detail

Our [supported Sass compilers](https://frontend.design-system.service.gov.uk/installing-with-npm/#requirements) would then be:

- Dart Sass 1.23.0 or higher
- LibSass 5.0 or higher (based on the tracked release in https://github.com/sass/libsass/issues/2807)

We would drop support for Ruby Sass, which is deprecated (as of 26 March 2019) and will not be updated to use the new module system.

36degrees:
It looks as though LibSass is now being deprecated:
https://github.com/sass/libsass/issues/3123

36degrees:
The beta of 2.10.0 of the US Web Design System migrates to the module system:

https://designsystem.digital.gov/about/releases/#version-2100-beta-0

thisisdano:
From the USWDS side, we're finding that the modules support is very patchy — perhaps too patchy to support in the near term. When we can control the compile chain, we can feel confident using modules, but we have a lot of project users that compile using, say, the built-in compiler in Jekyll, and this cannot parse modules at all. What I believe we've found — though this is not yet final — is that while we intend to move to modules when we can, we may strip this functionality from the beta and go as far as we can using the old syntax, waiting to fully move to modules if and when we find better overall support.

[...]

PR #2583, opened 2022-03-30
Support Dart Sass (part 1)

natcarey:
This PR is a step towards supporting Dart Sass.  I made the codebase more compatible and then moved back to libsass.  These changes would help prototype kit team and are a step closer to supporting Dart Sass on frontend project.

Dart sass is deprecating `1248 / 2` style division in favour of `math.div(1248, 2)` which is not supported by libsass.  This creates a logic problem where we can either fully support Dart Sass and break libsass or support Dart Sass with warnings while fully supporting libsass.

I have consolidated the issue into one `divideby` function so it only gets flagged once when `govuk-frontend` is used by the Prototype Kit - that is a massive improvement for our users.  I have also included a `.dart-scss` file which I think we might be able to configure so we use the Dart Sass version inside the Prototype Kit.


daniel-ac-martin:
@natcarey: I wonder whether there is anyway to detect whether we are in dart-sass or libsass?
Failing that, would it work to have a variable that the user can set to tell govuk-frontend which style of division to use?

36degrees:
Thanks for picking this up @natcarey.

To avoid confusion, I do want to clarify that GOV.UK Frontend has explicitly supported Dart Sass v1.0.0 or above since February 2020 and [we recommend using it over LibSass and Ruby Sass in our installation instructions](https://frontend.design-system.service.gov.uk/installing-with-npm/#requirements). We also have a [documented approach to silencing the deprecation warnings](https://frontend.design-system.service.gov.uk/importing-css-assets-and-javascript/#silence-deprecation-warnings-from-dependencies-in-dart-sass) which is [in line with the approach recommended by the Sass maintainers](https://github.com/sass/sass/issues/3131#issuecomment-907585432).

Having talked with @vanitabarrett and @querkmachine, we're going to close this as [the immediate issue with the Prototype Kit has been resolved](https://github.com/alphagov/govuk-prototype-kit/pull/1269) and released in [v12.0.2](https://github.com/alphagov/govuk-prototype-kit/releases/tag/v12.0.2). There is also [a follow up PR](https://github.com/alphagov/govuk-prototype-kit/pull/1272) open to replace the silent logger with the quiet deps approach we recommend. 

We still intend to move the GOV.UK Frontend build pipeline to use Dart Sass as part of #2239. This will only affect the CSS used in the review app and the compiled CSS we include in the ZIP file attached to each release (the `dist` folder). When we do that, we will need to work out how to deal with the deprecation warnings. However, we anticipate the Sass team may go on to deprecate further features and so we should work on a generic solution rather than introducing a new layer of abstraction that focuses specifically on division.

Finally, it's worth noting that _at some point_ in the future we plan to migrate to the Sass module system as part of #1791. This isn't something we're looking at in the short term as we know lots of our users are still using LibSass, including the GOV.UK team here at GDS.

PR #2599, opened 2022-04-22
Remove pattern attribute from date input component

### LINKED ISSUE #1701: Consider removing the pattern attribute from date input once traffic from iOS < 12.2+ drops off
## What

We use the pattern attribute to force the inputs within the date input component to trigger a numeric keypad on iOS < 12.2+, (and possibly some older Android browsers?)

## Why

Once the number of users that require the `pattern` attribute in order to trigger a numeric keypad has dropped to a low enough percentage, we can remove it and rely exclusively on `inputmode`.

querkmachine:
`pattern="[0-9]*"` was added to these inputs to prompt iOS to display the numeric keypad, which [was considered the best approach at the time](https://css-tricks.com/finger-friendly-numerical-inputs-with-inputmode/). This has been unnecessary since Safari 12.2, when support for [the standardised `inputmode` attribute](https://html.spec.whatwg.org/multipage/interaction.html#attr-inputmode) was added to Safari.

As of March 2022, the proportion of GOV.UK visitors using versions of iOS 12.x and below has fallen to 0.08% of total traffic (0.19% of iOS traffic), so it seems like a reasonable time to consider removing the default `pattern` attribute.

Would resolve #1701.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,14 @@ If you're not using the Nunjucks macros, remove the `aria-live` attribute from t
 
 This change was introduced in [pull request #2577: Refactor character count to inject new element](https://github.com/alphagov/govuk-frontend/pull/2577)
 
+#### Remove `pattern` from the date input component
+
+If you're not using the Nunjucks macros, remove `pattern="[0-9]*"` from each of the date input component's elements.
+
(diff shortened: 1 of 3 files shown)


36degrees (approved):
👍🏻

PR #2601, opened 2022-04-22
Test Sass compiles using the latest compilers

### LINKED ISSUE #2589: Ensure GOV.UK Frontend compiles with the latest Sass compilers as part of CI
## What

We have an existing GitHub Action which is designed to check that GOV.UK Frontend compiles successfully when using the oldest versions of [the Sass compilers we support](https://frontend.design-system.service.gov.uk/installing-with-npm/#requirements).

Add additional jobs to the workflow to ensure GOV.UK Frontend also compiles with the latest compatible versions of:

- Dart Sass (^v1.0.0)
- LibSass (^v3.3.0)
- Ruby Sass (^3.4.0)

The latest versions of Dart Sass output deprecation warnings when compiling. These should be silenced by treating GOV.UK Frontend as a dependency (including it using `loadPaths`) and [setting the `quietDeps` flag](https://frontend.design-system.service.gov.uk/importing-css-assets-and-javascript/#silence-deprecation-warnings-from-dependencies-in-dart-sass).

## Why

We want to ensure that we maintain compatibility with the latest versions of the Sass compilers. As the compilers should be following semantic versioning we don't anticipate this to be a problem, but this will give us confidence and flag any issues that may come up in new versions (of Dart Sass, in particular).

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [x] The latest compatible versions of Dart Sass, LibSass and Ruby Sass are tested as part of the existing CI workflow
- [x] Status checks clearly reported through the PR interface
- [ ] The status checks are configured as 'required status checks', causing a PR to become un-mergeable if they fail

36degrees:
We have existing checks to ensure that GOV.UK Frontend compiles successfully when using the oldest versions of the Sass compilers we support.

We want to ensure that we maintain compatibility with the latest versions of the Sass compilers as well. As the compilers should be following semantic versioning we don't anticipate this to be a problem, but this will give us confidence and flag any issues that may come up in new versions (of Dart Sass, in particular).

Add additional jobs to the GitHub Actions workflow to ensure GOV.UK Frontend also compiles with the latest compatible versions of:

- Dart Sass (v1)
- LibSass (v3)
- Ruby Sass (v3)

Because these jobs are able to run in parallel, this thankfully doesn't have much impact on the overall time it takes for the action to run. The action has run a few times whilst putting this PR together, and tends to complete in 36-38s. For comparison, the most recent runs on the main branch look to have taken between 30-33s (very unscientifically based on glancing at the first page of runs on that branch).

Closes #2589

### DIFF
--- a/.github/workflows/sass.yaml
+++ b/.github/workflows/sass.yaml
@@ -27,7 +27,13 @@ jobs:
       - run: |
           npm install -g sass@v1
           sass --version
-      - run: time sass src/govuk/all.scss > /dev/null
+      # Treat GOV.UK Frontend as a dependency by importing it via load paths,
+      # allowing us to mimic the way we recommend our users silence deprecation
+      # warnings using the `quiet-deps` flag.
+      #
+      # Run the command through a shell to ensure `time` measures the time
+      # taken by the entire pipeline, as we are now piping input into `sass`.
+      - run: time sh -c 'echo "@import "\""govuk/all"\"";" | sass --stdin --quiet-deps --load-path=src > /dev/null'
 
 
   # Node Sass v3.4.0 = LibSass v3.3.0
--- a/.github/workflows/sass.yaml
+++ b/.github/workflows/sass.yaml
@@ -16,6 +16,21 @@ jobs:
           sass --version
       - run: time sass src/govuk/all.scss > /dev/null
 
+  dart-sass-latest:
+    name: Dart Sass v1 (latest)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 14
+      - run: |
+          npm install -g sass@v1
+          sass --version
+      - run: time sass src/govuk/all.scss > /dev/null
+
+
+  # Node Sass v3.4.0 = LibSass v3.3.0
   lib-sass:
     name: LibSass v3.3.0 (deprecated)
     runs-on: ubuntu-latest
@@ -29,6 +44,20 @@ jobs:
           node-sass --version
       - run: time node-sass src/govuk/all.scss > /dev/null
 
+  # Node Sass v7.x = LibSass v3 latest
+  lib-sass-latest:
+    name: LibSass v3 (latest, deprecated)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 14
+      - run: |
+          npm install -g node-sass@v7
+          node-sass --version
+      - run: time node-sass src/govuk/all.scss > /dev/null
+
   ruby-sass:
     name: Ruby Sass v3.4.0 (deprecated)
     runs-on: ubuntu-latest
@@ -39,5 +68,19 @@ jobs:
           ruby-version: 2.1.9 # Oldest version available on ruby/setup-ruby
       - run: |
           gem install sass -v 3.4.0
+          sass --version
+      - run: time sass src/govuk/all.scss > /dev/null
+
+  # Ruby Sass is EOL so we can pin to the last release
+  ruby-sass-latest:
+    name: Ruby Sass v3.7.4 (latest, deprecated)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: 2.6 # Latest Ruby when Ruby Sass was EOL'ed (April 2019)
+      - run: |
+          gem install sass -v 3.7.4
           sass --version
       - run: time sass src/govuk/all.scss > /dev/null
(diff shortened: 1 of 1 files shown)


36degrees:
In acafa4ffabdaa91e4e8b74117c097b601a6ab66f I've also silenced the deprecation warnings by importing GOV.UK Frontend as a dependency and using the `quiet-deps` flag. I don't know if we really need the changes in this commit – all it's really doing is preventing deprecation warnings from appearing in the CI logs? I think I was originally thinking that the warnings would cause the build to fail, but obviously they don't – because they're warnings!

I guess it does give us some confidence that [the approach to silencing the deprecation warnings that we recommend in our docs](https://frontend.design-system.service.gov.uk/importing-css-assets-and-javascript/#silence-deprecation-warnings-from-dependencies-in-dart-sass) works if we go in and check the build output?

Thoughts?

lfdebrux:
Nice!

I have a couple of thoughts, but otherwise this all looks good to me. I think the `--quiet-deps` flag is a nice idea too.

Thought number 1, what do you anticipate we should do if/when a new major release of one of the compilers comes out?

Thought number 2, I know it says in the comments that Ruby Sass is EOL so we're not expecting any more updates, but I can't see why we wouldn't allow for updates anyway, for the latest test, just in case?

Thought number 3, I wonder if there is a way we can make this more DRY? Right now the amount of code is increasing, and I can imagine it increasing further, it would be nice if adding new versions to the text matrix didn't involve a lot of copying and pasting...

36degrees:
> Thought number 1, what do you anticipate we should do if/when a new major release of one of the compilers comes out?

Good question. I think `dart-sass` is the most likely compiler to bring out a major release. I _suspect_ that we won't be able to support v2 of dart-sass whilst maintaining support for LibSass and Ruby Sass, so at that point we'd need to stop and take a broader look at compiler support anyway.

You have made me realise that although it's unlikely we'd see a new major version of LibSass, we could see a new major version of `node-sass` – for example, dropping support for an older version of Node. So we might need to either change that to install `latest`? I've tried to pin to major versions generally to avoid new releases of a Sass compiler resulting in unrelated build failures on PRs, so the alternative would be to keep an eye on new major releases of node-sass some other way?

> Thought number 2, I know it says in the comments that Ruby Sass is EOL so we're not expecting any more updates, but I can't see why we wouldn't allow for updates anyway, for the latest test, just in case?

Fair, although it's worth noting that the last release was in April 2019 and [the repo](https://github.com/sass/ruby-sass) for the library is archived, so I'd be _very_ surprised if there was a new release at this point – especially one that would affect compilation. I don't actually know how `gem install` works with constraints / major versions, but will try it and see.

> Thought number 3, I wonder if there is a way we can make this more DRY? Right now the amount of code is increasing, and I can imagine it increasing further, it would be nice if adding new versions to the text matrix didn't involve a lot of copying and pasting...

Maybe? There are quite a few things that vary between each job (ruby vs node, version of ruby or node, version of the compiler, the command that needs to be run) and I think any attempt to define that in a matrix is going to result in just as much 'configuration' and potentially make things harder to read? The only truly common parts of the job that I can see are the fact that it runs on ubuntu-latest and that it checks out the code 😅 

I also can't see us adding many (any?) more versions, and don't think these jobs are likely to be around for _that long_  as we'll need to drop support for LibSass and Ruby Sass before too long… (famous last words, I know)

Having said all of that, if you have any idea for something neater or want to spike an alternate approach, I'm definitely open to it!

lfdebrux:
> Good question. I think dart-sass is the most likely compiler to bring out a major release. I suspect that we won't be able to support v2 of dart-sass whilst maintaining support for LibSass and Ruby Sass, so at that point we'd need to stop and take a broader look at compiler support anyway.

Yeah, I think you're very likely right about that, let's hope it doesn't happen too soon!

> You have made me realise that although it's unlikely we'd see a new major version of LibSass, we could see a new major version of node-sass – for example, dropping support for an older version of Node. So we might need to either change that to install latest? I've tried to pin to major versions generally to avoid new releases of a Sass compiler resulting in unrelated build failures on PRs, so the alternative would be to keep an eye on new major releases of node-sass some other way?

Yeah, keeping an eye on updates is a general issue, but I think it should be fine for now... 🤞 

> Fair, although it's worth noting that the last release was in April 2019 and [the repo](https://github.com/sass/ruby-sass) for the library is archived, so I'd be very surprised if there was a new release at this point – especially one that would affect compilation. I don't actually know how gem install works with constraints / major versions, but will try it and see.

Right, _very_ EOL then 😅 Thanks for changing it anyway!

> Maybe? There are quite a few things that vary between each job (ruby vs node, version of ruby or node, version of the compiler, the command that needs to be run) and I think any attempt to define that in a matrix is going to result in just as much 'configuration' and potentially make things harder to read? The only truly common parts of the job that I can see are the fact that it runs on ubuntu-latest and that it checks out the code 😅

Fair point. Sometimes verbose is better than clever ;)

lfdebrux (approved):
LGTM! 👍

PR #2600, opened 2022-04-22
Update publishing doc card examples

vanitabarrett:
Update the publishing docs to point to newer examples of our release cards. I noticed when drafting the cards for the v4.1.0 release that some of these were outdated and didn't follow our latest way of doing things, e.g:

- splitting out comms into separate cards for drafting and posting
- having one overarching epic card for the release, which all release cards are linked to
- mention of working with the new separate GOV.UK Prototype team

PR #2605, opened 2022-04-25
Replace card with issue in support branch doc

EoinShaughnessy:
Updates our 'Publishing from a support branch' doc.

Replaces 'card' with 'issue' throughout.

Previously, we used these terms interchangeably, so for consistency's sake we should stick with the more appropriate one.

That's probably 'issue', as not only does it reflect GitHub's UI, but also 'card' can have multiple meanings.

[We've made similar changes to our 'Publishing' doc](https://github.com/alphagov/govuk-frontend/pull/2600#discussion_r857439664).


36degrees (approved):
👍🏻

PR #2609, opened 2022-05-03
Update pattern removal entry in changelog

EoinShaughnessy:
Suggests a few edits to our changelog entry about removing the `pattern` attribute from the date input component.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,11 +53,11 @@ This change was introduced in [pull request #2577: Refactor character count to i
 
 #### Remove `pattern` from the date input component
 
-If you're not using the Nunjucks macros, remove `pattern="[0-9]*"` from each of the date input component's elements.
+If you're not using the Nunjucks macros, remove `pattern="[0-9]*"` from each of the date input's elements.
 
-This attribute was originally added to enforce numeric keyboard functionality in Safari on iOS. The `inputmode` attribute provides this functionality for newer Safari versions, so we no longer need to use the `pattern` attribute. 
+We originally added the `pattern` attribute to make numeric keypad functionality the default functionality in Safari on iOS. However, we no longer need to use it, as the `inputmode` attribute provides this functionality for newer Safari versions.
 
-This change was introduced in [pull request #2599: Remove pattern attribute from date input component](https://github.com/alphagov/govuk-frontend/pull/2599).
+This change was introduced in [pull request #2599: Remove `pattern` attribute from date input component](https://github.com/alphagov/govuk-frontend/pull/2599).
 
 ### Fixes
 


CHANGELOG.md
```diff
@@ -53,11 +53,11 @@ This change was introduced in [pull request #2577: Refactor character count to i
 
 #### Remove `pattern` from the date input component
 
-If you're not using the Nunjucks macros, remove `pattern="[0-9]*"` from each of the date input component's elements.
+If you're not using the Nunjucks macros, remove `pattern="[0-9]*"` from each of the date input's elements.
 
-This attribute was originally added to enforce numeric keyboard functionality in Safari on iOS. The `inputmode` attribute provides this functionality for newer Safari versions, so we no longer need to use the `pattern` attribute. 
+We originally added the `pattern`attribute to enforce numeric keyboard functionality in Safari on iOS. However, we no longer need to use it, as the `inputmode` attribute provides this functionality for newer Safari versions.
```
EoinShaughnessy, line 58: Does 'enforce' have a specific meaning for users, or could we say something like "...to make sure numeric keyboard functionality works in..."?

EoinShaughnessy, line 58:
```suggestion
We originally added the `pattern` attribute to enforce numeric keyboard functionality in Safari on iOS. However, we no longer need to use it, as the `inputmode` attribute provides this functionality for newer Safari versions.
```

CHANGELOG.md
```diff
@@ -53,11 +53,11 @@ This change was introduced in [pull request #2577: Refactor character count to i
 
 #### Remove `pattern` from the date input component
 
-If you're not using the Nunjucks macros, remove `pattern="[0-9]*"` from each of the date input component's elements.
+If you're not using the Nunjucks macros, remove `pattern="[0-9]*"` from each of the date input's elements.
 
-This attribute was originally added to enforce numeric keyboard functionality in Safari on iOS. The `inputmode` attribute provides this functionality for newer Safari versions, so we no longer need to use the `pattern` attribute. 
+We originally added the `pattern` attribute to enforce numeric keyboard functionality in Safari on iOS. However, we no longer need to use it, as the `inputmode` attribute provides this functionality for newer Safari versions.
```
EoinShaughnessy, line 58: Does 'enforce' have a specific meaning for users, or could we say something like "...to make sure numeric keyboard functionality works in..."?

CHANGELOG.md
querkmachine: Probably not either. It's more like it's a "suggestion" that defaulted the on-screen keyboard to a numeric input layout—but the user could still change it to alphanumeric if they wanted to, so it's not really enforced, nor was it functionally necessary for the user to provide valid input.

EoinShaughnessy: Ah cool, thanks for explaining. Could something like the below work?

> We originally added the `pattern` attribute to make numeric keypad functionality the default functionality in Safari on iOS.

querkmachine: Yeah, that makes sense, I think!

EoinShaughnessy: Cheers! :)

CHANGELOG.md
EoinShaughnessy, line 58:
```suggestion
We originally added the `pattern` attribute to make numeric keypad functionality the default functionality in Safari on iOS. However, we no longer need to use it, as the `inputmode` attribute provides this functionality for newer Safari versions.
```

querkmachine (approved):

PR #2610, opened 2022-05-03
Add pagination component

owenatgov:
## What
Adds the pagination component to govuk-frontend. See below epic for background on this addition .

### Related issues/pull requests
- Epic: https://github.com/alphagov/govuk-design-system/issues/1985
- Design system website pull request: https://github.com/alphagov/govuk-design-system/pull/1972
- Supersedes existing prototype pull request: https://github.com/alphagov/govuk-frontend/pull/2419

### Standard view
Screenshot 2022-05-03 at 17 04 58

### With a large number of pages
Teams have the option to hide several page numbers behind ellipses
Screenshot 2022-05-03 at 16 59 46

### Without numbers
Alternative "block" view for pre/next navigation without page numbers
Screenshot 2022-05-03 at 16 59 57


edwardhorsford:
Is there a review app for this? From the screenshots it looks like there might be errant whitespace in the previous and next links - the underlines look oddly sized for the text content.

querkmachine:
The title and label lines on the "block" view are also not aligned with one another, possibly due to unruly whitespace again.

image

@edwardhorsford We don't have review apps at the moment due to the GitHub/Heroku outage. You need to pull the branch and run it locally for the time being.

edwardhorsford:
@querkmachine that's out of alignment too - but what I was talking about was this:
Screenshot 2022-05-04 at 13 28 01

src/govuk/components/pagination/pagination.yaml
```diff
@@ -0,0 +1,169 @@
...
+- name: previous
+  type: object
+  required: false
+  description: A link to the previous page, if there is one.
+  params:
+  - name: text
+    type: string
+    required: false
+    description: The previous page link text. Defaults to 'Previous page', where 'page' is visually hidden.
+  - name: label
```
36degrees, line 11: Can we use `labelText` here (and for next) so that we can introduce `labelHtml` later if we find there's a need for it?

src/govuk/components/pagination/template.test.js
```diff
@@ -0,0 +1,72 @@
...
+      const $ = render('pagination', examples['with prev and next only and labels'])
+      const prevLabel = $('.govuk-pagination__item--prev .govuk-pagination__link-label')
+      const nextLabel = $('.govuk-pagination__item--next .govuk-pagination__link-label')
+
+      expect(prevLabel.text()).toEqual('1 of 3')
+      expect(nextLabel.text()).toEqual('3 of 3')
+    })
+  })
+
+  describe('custom attributes', () => {
```
36degrees, line 63: What does custom attributes mean in this context?

src/govuk/components/pagination/template.test.js
```diff
@@ -0,0 +1,72 @@
...
+ */
+/* eslint-env jest */
+
+const axe = require('../../../../lib/axe-helper')
+
+const { render, getExamples } = require('../../../../lib/jest-helpers')
+
+const examples = getExamples('pagination')
+
+describe('Pagination', () => {
```
36degrees, line 12: Would expect to see a few more tests that provide better coverage of the API of the component. It's probably worth having a read through the tests for a couple of other components to get an idea of the sort of things we'd tend to include.

Generally tests are better at covering the things that are not visually obvious in the review app. For example, ensuring that when users pass in additional classes they are applied correctly, text can be customised, text is sanitised and HTML is not, etc.

It might also make sense to add some tests to ensure that SVGs are `aria-hidden` / `focusable=false`; that rel=prev/next are applied correctly – anything that has semantic meaning or is important for accessibility but could easily be missed in a future change.

src/govuk/components/pagination/_index.scss
```diff
@@ -0,0 +1,204 @@
...
+  }
+
+  .govuk-pagination__item--ellipses {
+    color: $govuk-secondary-text-colour;
+  }
+
+  .govuk-pagination__link {
+    display: block;
+    min-width: govuk-spacing(3);
+    text-align: center;
```
36degrees, line 105: Might be worth adding a comment on why we're doing this

src/govuk/components/pagination/_index.scss
```diff
@@ -0,0 +1,204 @@
...
+  .govuk-pagination__item:nth-child(2),
+  .govuk-pagination__item:nth-last-child(2) {
+    display: block;
+  }
+
+  .govuk-pagination__item,
+  .govuk-pagination__item--current,
+  .govuk-pagination__item--ellipses {
+    box-sizing: border-box;
+    min-width: govuk-spacing(8);
```
36degrees, line 87: Are these _really_ tied to the spacing scale? If not, might make more sense to just use `50px`.

src/govuk/components/pagination/pagination.yaml
```diff
@@ -0,0 +1,169 @@
+params:
```
36degrees, line 1: I haven't reviewed these fully – ideally the descriptions at least should be reviewed by @EoinShaughnessy before shipping.

src/govuk/components/pagination/pagination.yaml
```diff
@@ -0,0 +1,169 @@
...
+  type: string
+  required: false
+  description: The type of navigation. Goes in the aria label for the encapsulating `nav` element. Defaults to "results".
+- name: items
+  type: array
+  required: false
+  description: Array of link objects.
+  params:
+  - name: number
+    type: number
```
36degrees, line 46: Is this really a string? We don't treat it as a number in any meaningful sense, and we just output it as text.

36degrees (commented):
Thanks for this @owenatgov – looking really good so far. I have added quite a few comments but most of them are relatively minor. Very happy to talk through it if that'd be useful.

src/govuk/components/pagination/pagination.yaml
EoinShaughnessy: Thanks for the heads up, @36degrees! :)

Since preview is unavailable, I'll copy this content into a Google Doc to get a better idea of what it'll look like.

EoinShaughnessy: Also, after I've reviewed the descriptions, they'll need to go for 2i (external tech writer review).

src/govuk/components/pagination/pagination.yaml
```diff
@@ -0,0 +1,169 @@
+params:
+- name: previous
+  type: object
+  required: false
+  description: A link to the previous page, if there is one.
+  params:
+  - name: text
+    type: string
+    required: false
+    description: The previous page link text. Defaults to 'Previous page', where 'page' is visually hidden.
```
EoinShaughnessy, line 10: Does "previous page link text" mean the link text ON the previous page, or the link text TO the previous page?

If the former, then could change this to:

> The previous page's link text.

If the latter, then could change this to:

> The link text to the previous page.

src/govuk/components/pagination/pagination.yaml
```diff
@@ -0,0 +1,169 @@
...
+  - name: label
+    type: string
+    required: false
+    description: Optional label that goes underneath the previous page link providing further context for where it goes.
```
EoinShaughnessy, line 14:
```suggestion
    description: Optional label that goes underneath the previous page's link, providing further context for the user about where it goes.
```

src/govuk/components/pagination/pagination.yaml
```diff
@@ -0,0 +1,169 @@
...
+  - name: href
+    type: string
+    required: true
+    description: The previous page URL.
```
EoinShaughnessy, line 18:
```suggestion
    description: The previous page's URL.
```

src/govuk/components/pagination/pagination.yaml
```diff
@@ -0,0 +1,169 @@
...
+    required: false
+    description: The link URL.
+  - name: current
+    type: boolean
+    required: false
+    description: Set to 'true' to indicate the current page.
+  - name: ellipsis
+    type: boolean
+    required: false
+    description: For specifying if you want an ellipsis at a given point between numbers. If you set this as true, the other attributes aren't required for the same item.
```
EoinShaughnessy, line 60:
```suggestion
    description: For specifying if you want an ellipsis at a given point between numbers. If you set this option as `true`, the other attributes are not required for the same item.
```

src/govuk/components/pagination/pagination.yaml
```diff
@@ -0,0 +1,169 @@
...
+    required: false
+    description: Optional label that goes underneath the next page link providing further context for where it goes.
+  - name: href
+    type: string
+    required: true
+    description: The next page URL.
+- name: type
+  type: string
+  required: false
+  description: The type of navigation. Goes in the aria label for the encapsulating `nav` element. Defaults to "results".
```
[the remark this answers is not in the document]
EoinShaughnessy: Will all our users know what 'encapsulating' means, or is there a plainer way of saying this?

src/govuk/components/pagination/_index.scss
```diff
@@ -0,0 +1,204 @@
...
+      flex-basis: 100%;
+    }
+
+    &:before {
+      order: 1;
+    }
+
+    &:after {
+      order: 3;
+    }
```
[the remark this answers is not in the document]
owenatgov: I'm drawn more to the first option. I'm wondering now if the prev/next don't necessarily need to be in the list, although I'm not sure if that creates a weird experience for screen reader users. The second option I could probably achieve with some clever `aria-hidden` use but it feels messy. Let me confer with design and accessibility on this.

src/govuk/components/pagination/_index.scss
```diff
@@ -0,0 +1,204 @@
...
+    .govuk-pagination__list {
+      display: block;
+    }
+
+    .govuk-pagination__item {
+      padding: govuk-spacing(3);
+      float: none;
+    }
+
+    // Set the after pseudo element to a block which makes the title visually display as block level whilst programmatically being inline
```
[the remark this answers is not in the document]
owenatgov: I'll sort this out but is it written down anywhere?

src/govuk/components/pagination/_index.scss
```diff
@@ -0,0 +1,204 @@
...
+
+      .govuk-pagination__link-label {
+        text-decoration: none;
+      }
+    }
+  }
+
+  .govuk-pagination__link-label {
+    // Don't set the font via govuk font as currently this will mean the font size is 14px on mobile
+    // which is an inaccessible font size
```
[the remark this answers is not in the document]
owenatgov: This to me is more credence to fix the issues with typography more urgently. Whilst the issues here are issues, not passing the font is the lesser evil compared to shipping an inaccessible font size.

Something I can try here is still pass the font so that it adds the helper but maintain the font-size declaration on line 147. I'm not 100% sure that's going to work and it's messy but could be worth a shot?

src/govuk/components/pagination/template.test.js
owenatgov: This was intended to be "passing custom content on the macro renders it properly". I thought I'd copied this from an existing test but when I search `custom attributes` in govuk-frontend very little comes up. It might've already been there when I copied the prototype? As per your other comment, these tests need some work so I'll either remove this or make it much clearer.

src/govuk/components/pagination/pagination.yaml
owenatgov: I'll take your recommendations as they are now @EoinShaughnessy but I'm going to be making some changes to these following Ollie's review. I probably should've said, sorry. Be prepared for these to change a bit.

EoinShaughnessy: @owenatgov No worries, give me a shout when you need a review!

christopherthomasdesign:
Just had a look through, a few more comments...

1. +1 to Ed's question about the whitespace before 'Previous' and after 'Next' – is this a workaround for something? It's quite noticeable that there's an extra space in there, it should deffo be flush with the line below
2. Should the bold 'current page' numbers have a hover state? To me that indicates you can interact with them...
3. Think the pagination link labels might be a bit small – IMO the bold of the 'previous / next' bit is enough of a visual differentiator between the two. So I'd probs bump up the `govuk-pagination__link-label` to 19px on bigger screens to match the rest of the link. GOV.UK do it this way too, see https://www.gov.uk/adoption-records. If we do this, it might help to add an extra bit of spacing between the two rows so they're not bunched up too closely – no more than 5px

edwardhorsford:
I suspect the whitespace issues are caused by the new lines. That can likely be fixed using [Nunjucks whitespace control](https://mozilla.github.io/nunjucks/templating.html#whitespace-control).

owenatgov:
@edwardhorsford was spot on that this was a whitespace issue. All sorted now thanks to Ed's helpful link and some light flexbox styling (thanks again @36degrees).

@christopherthomasdesign I believe I've covered all your points. Please take another look at let me know if you spot anything else.

I've reorganised the sass so it's a bit more sensible and tidy so some of the comments above will be listed as outdated even though they're still relevant. I'll still work through them in order.

src/govuk/components/pagination/_index.scss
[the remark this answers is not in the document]
36degrees: Good shout – have raised #2613.

src/govuk/components/pagination/template.njk
```diff
@@ -0,0 +1,76 @@
...
+
+        {%- if params.previous.label -%}
+          <span class="govuk-visually-hidden">:</span>
+          <span class="govuk-pagination__link-label">{{ params.previous.label }}</span>
+        {%- endif -%}
+      </a>
+    </div>
+  {% endif %}
+
+  <ul class="govuk-pagination__list">
```
36degrees, line 28: At the minute this is outputted even if there are no items, so e.g. in the `with-prev-and-next-only` example there's an empty `<ul></ul>`

owenatgov: I forgot to put an `if` there! Will sort out.

src/govuk/components/pagination/pagination.yaml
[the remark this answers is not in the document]
vanitabarrett: Feel free to ignore this as I've had 0 involvement in any pagination discussions up to this point, but I'm wondering if 'type' is the best name for this param 🤔 I'm guessing we don't want to use 'label' because it could be confused with the previous/next label options, but I'm not sure if it's clear that 'type' -> 'aria-label'. Feels like it might change the behaviour of the component, but I don't think it does?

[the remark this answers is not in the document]
36degrees: What about something like `navLabel` or `regionLabel`?

I also think the description could be simpler, perhaps something like:

```suggestion
  description: The label for the navigation landmark that wraps the pagination. Defaults to "results".
```

src/govuk/components/pagination/template.njk
```diff
@@ -0,0 +1,76 @@
...
+      </li>
+    {% endif %}
+
+    {% for item in params.items %}
+      {% if item.ellipsis %}
+        <li class="govuk-pagination__item govuk-pagination__item--ellipses">&ctdot;</li>
+      {% elif item.number %}
+        {% if item.current %}
+          <li class="govuk-pagination__item govuk-pagination__item--current">
+            <span class="govuk-visually-hidden">Page </span>{{ item.number }}<span class="govuk-visually-hidden"> (current page)</span>
```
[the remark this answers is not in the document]
owenatgov: I've tried applying a series of overrides for the content here, going `html > text > number`. It feels a bit complicated to me so I'm not wedded to it. It all sort of depends how attached we are to those visually hidden elements, which is the only reason to keep `number` around. I suppose we could do something clever where we default to the item's count in the array if no link text is provided in an item, but that also feels overcomplicated to me. I'm gonna defer to the team.

vanitabarrett:
@owenatgov Could we add this to a full page example, e.g: http://govuk-frontend-review.herokuapp.com/full-page-examples/search ?

src/govuk/components/pagination/template.njk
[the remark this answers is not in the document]
36degrees: I think the main reason for passing customising is for translation, where as you rightly say we can't predict where the number is going to go. Hence needing to override the entire string.

I can't really follow what you're suggesting – if you're only setting the visually hidden text then how does the visible number get defined? We don't always want to just have 1…n using the loop index as numbers as the user might be on page 140 of 300, in which case we need to be able to output `[1 … 138 139 140 141 142 … 300]`.

Having said that, your suggestion has got me wondering if we could simplify things a little by using `aria-label` – which gets rid of the user needing to pass HTML but does mean that the number is duplicated (therefore potentially allowing for a mismatch between the visible label and the visually hidden label):

```njk
<a class="govuk-link govuk-pagination__link" href="{{ item.href }}" aria-label="{{ visuallyHiddenLabel | default("Page " + item.number) }} {%- for attribute, value in item.attributes %} {{ attribute }}="{{ value }}" {%- endfor -%}>
    {{ item.number }}
</a>
```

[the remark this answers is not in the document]
owenatgov: @lfdebrux Some general thoughts on this approach:

- I worry we're enforcing latin/romance language grammar by putting the hidden "page" before the number and not giving teams any way around this.
- We would need 2 attributes for this: `params.item.visuallyHiddenPageText` and `params.item.visuallyHiddenCurrentPageText`. The latter being for the "(current page)" after current items. We'd need to find some way to articulate when these get used which I worry is going to simply replicate the complexity of the macro in a different form.

Having a hierarchy of overrides, whilst a bit awkward as I've articulated, does solve the problem at the risk of service teams potentially doing whacky things. Something I could do to simplify this would be to remove the `text` attribute and only have `html` > `number` which defaults to the visually hidden stuff.

@36degrees Good point on the looping 🤦🏻 I do quite like the visually hidden label idea. It feels like the simplest approach in terms of the macro API whilst giving us some nice looseness. It also feels less risky than letting folks pass html which I feel is more likely to cause problems than a text only aria label. I'm gonna give that a go.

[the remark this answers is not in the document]
owenatgov: @36degrees Do you reckon the aria label approach could apply to the prev and next buttons as well? So we could remove the need for `params.previous.html`?

src/govuk/components/pagination/pagination.yaml
owenatgov: When I say the same item I meant the instance of the pagination item, as part of the `items` array within the pagination macro, that the (presumed) dev will be constructing to call the macro. To illustrate this, here's what a normal pagination `item` might look like:

```
- number: '1'
  href: '/page/1'
```

...whereas here's what an ellipsis `item` looks like:

```
- ellipsis: true
```

What we're trying to convey here is that if the `ellipsis` attribute is set in the macro call of an `item`, then anything that get's passed to that one `item` call will be ignored and so it's a waste of code. For example if we did something like this:

```
- number: '1'
  href: '/page/1'
  ellipsis: true
```

...then the `number` and `href` would be a waste of code as they will be ignored during rendering in favour of just rendering an ellipsis list item. That also hopefully answers your other question (not annoying) that the other attributes are the other `item` params: `number`, `href`, `current` etc.

Does that make sense? I'm wondering if the confusion comes from the fact that I'm using `attributes` to describe thing that go into a macro call and not `params`. Still getting used to our tech docs etymology.

src/govuk/components/pagination/pagination.yaml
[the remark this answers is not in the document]
owenatgov: I've followed Ollie's suggestion and changed the attribute to `landmarkLabel`.

owenatgov:
@EoinShaughnessy Taken your review into consideration. There have been some changes, notably new and renamed params (right word?). Please when you've got a spare mo, run your eye over the tech docs on this PR again. Happy to discuss anything.

owenatgov:
@vanitabarrett Added pagination to the full page search example 👍🏻

EoinShaughnessy:
@owenatgov Cheers for the heads up! Will review at 3.30.

src/govuk/components/pagination/pagination.yaml
EoinShaughnessy: That's a great explanation - thanks! Re 'attributes', actually I thinking you're using the right term, that's [what we say throughout the component macro options](https://www.google.com/search?q=%3Csite%3Adesign-system.service.gov.uk%2F+%22attributes%22%3E&rlz=1C5CHFA_enGB970GB970&oq=%3Csite%3Adesign-system.service.gov.uk%2F+%22attributes%22%3E&aqs=chrome..69i57j69i64.4296j0j4&sourceid=chrome&ie=UTF-8).

owenatgov: I've applied Ollie's recommendation. I agree that "option" is a suitable word and if we want to create consistency, I'm all for it.

src/govuk/components/pagination/pagination.yaml
```diff
@@ -0,0 +1,243 @@
...
+    required: true
+    description: The previous page's URL.
+  - name: attributes
+    type: object
+    required: false
+    description: HTML attributes (for example data attributes) to add to the anchor.
+- name: next
+  type: object
+  required: false
+  description: A link to the next page, if there is one.
```
EoinShaughnessy, line 30:
```suggestion
  description: A link to the next page, if there is a next page.
```

src/govuk/components/pagination/pagination.yaml
```diff
@@ -0,0 +1,243 @@
...
+    required: false
+    description: The visually hidden label for the pagination item which will be applied to an `aria-label` and announced by screen readers on the next page link.
+  - name: labelText
+    type: string
+    required: false
+    description: Optional label that goes underneath the next page link providing further context for where it goes.
+  - name: href
+    type: string
+    required: true
+    description: The next page URL.
```
EoinShaughnessy, line 47:
```suggestion
    description: The next page's URL.
```

src/govuk/components/pagination/pagination.yaml
```diff
@@ -0,0 +1,243 @@
...
+  - name: attributes
+    type: object
+    required: false
+    description: HTML attributes (for example data attributes) to add to the anchor.
```
EoinShaughnessy, line 51:
```suggestion
    description: The HTML attributes (for example, data attributes) you want to add to the anchor.
```

src/govuk/components/pagination/pagination.yaml
```diff
@@ -0,0 +1,243 @@
...
+  description: Array of link objects.
+  params:
+  - name: number
+    type: string
+    required: false
+    description: The pagination item text - usually a page number.
+  - name: visuallyHiddenText
+    type: string
+    required: false
+    description: The visually hidden label for the pagination item which will be applied to an `aria-label` and announced by screen readers on the pagination item link.
```
EoinShaughnessy, line 68:
```suggestion
    description: The visually hidden label (for the pagination item) which will be applied to an `aria-label` and announced by screen readers on the pagination item link.
```
In those last few words, are we referring to a link ON the pagination item item, or a link to the pagination item. If it's the latter, could change "pagination item link" to "link to the pagination item".

src/govuk/components/pagination/pagination.yaml
```diff
@@ -0,0 +1,243 @@
...
+  - name: href
+    type: string
+    required: false
+    description: The link URL.
+  - name: current
+    type: boolean
+    required: false
+    description: Set to 'true' to indicate the current page.
```
EoinShaughnessy, line 76:
```suggestion
    description: Set to `true` to indicate the current page.
```

src/govuk/components/pagination/pagination.yaml
```diff
@@ -0,0 +1,243 @@
...
+    required: false
+    description: HTML attributes (for example data attributes) to add to the anchor.
+- name: classes
+  type: string
+  required: false
+  description: The classes you want to add to the pagination `nav` parent.
+- name: attributes
+  type: object
+  required: false
+  description: HTML attributes (for example data attributes) to add to the pagination `nav` parent.
```
EoinShaughnessy, line 92:
```suggestion
  description: The HTML attributes (for example, data attributes) you want to add to the pagination `nav` parent.
```

EoinShaughnessy (commented):
Hey @owenatgov , cracking job on this here macros content! Just a couple of final questions:

- Are the options in any particular order, as in, does it reflect how we assume the user would work through them? Asking in case the 2i reviewer asks.

- Every time it says ‘params’ amid the macro options, does that indicate the end of one table and the beginning of another?

owenatgov:
@EoinShaughnessy Before I address your review, your questions:

1. The order doesn't make any difference to the rendering. They can be in any order.
2. Yes!

src/govuk/components/pagination/pagination.yaml
```diff
@@ -0,0 +1,243 @@
...
+  description: A link to the previous page, if there is one.
+  params:
+  - name: text
+    type: string
+    required: false
+    description: The link text to the previous page. Defaults to 'Previous page', where 'page' is visually hidden.
+  - name: visuallyHiddenText
+    type: string
+    required: false
+    description: The visually hidden label for the pagination item which will be applied to an `aria-label` and announced by screen readers on the previous page link.
```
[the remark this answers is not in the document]
owenatgov: Yes, 'previous page link' is the link being specified to go back to the previous page in the context of a UI that uses pagination to organise its content.

src/govuk/components/pagination/pagination.yaml
```diff
@@ -0,0 +1,243 @@
...
+  - name: labelText
+    type: string
+    required: false
+    description: Optional label that goes underneath the previous page's link, providing further context for the user about where it goes.
```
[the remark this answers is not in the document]
owenatgov: I think this is a good idea, I'll do it!

src/govuk/components/pagination/pagination.yaml
```diff
@@ -0,0 +1,243 @@
...
+  description: A link to the next page, if there is one.
+  params:
+  - name: text
+    type: string
+    required: false
+    description: The link text to the next page. Defaults to 'Next page', where 'page' is visually hidden.
+  - name: visuallyHiddenText
+    type: string
+    required: false
+    description: The visually hidden label for the pagination item which will be applied to an `aria-label` and announced by screen readers on the next page link.
```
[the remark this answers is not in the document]
owenatgov: YES

src/govuk/components/pagination/pagination.yaml
owenatgov: It's the former. We're not referring to the target of the link here, we're referring to the link html element itself.

src/govuk/components/pagination/pagination.yaml
```diff
@@ -0,0 +1,243 @@
...
+    required: false
+    description: The link URL.
+  - name: current
+    type: boolean
+    required: false
+    description: Set to 'true' to indicate the current page.
+  - name: ellipsis
+    type: boolean
+    required: false
+    description: For specifying if you want an ellipsis at a given point between numbers. If you set this option as `true`, any other options for the item are ignored.
```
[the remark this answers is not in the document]
owenatgov: In this context I used it to mean "tell the macro that we want this item to just be an ellipsis instead of a link/number like most of the other pagination items".

The use case for ellipses is when you want to portray a large number of pages in a pagination without printing every single page in this element. There are details on the [website guidance preview](https://add-pagination--govuk-design-system-preview.netlify.app/components/pagination/#for-larger-numbers-of-pages) (although awkwardly at present it's using the wrong code example).

src/govuk/components/pagination/template.njk
```diff
@@ -0,0 +1,84 @@
...
+          </svg>
+          {%- if params.previous.text -%}
+            {{- params.previous.text -}}
+          {%- else -%}
+            Previous
+          {%- endif -%}
+        </span>
+
+        {%- if params.previous.labelText -%}
+          <span class="govuk-visually-hidden">:</span>
```
peteryates, line 19:
```suggestion
          <span class="govuk-visually-hidden govuk-pagination__link-divider">:</span>
```

src/govuk/components/pagination/template.njk
```diff
@@ -0,0 +1,84 @@
...
+          {%- else -%}
+            Next
+          {%- endif -%}
+          {%- if not blockLevel -%}
+            {{- nextArrow | safe -}}
+          {%- endif -%}
+        </span>
+
+        {%- if params.next.labelText -%}
+          <span class="govuk-pagination__link-divider govuk-visually-hidden">:</span>
```
peteryates, line 78:
```suggestion
          <span class="govuk-visually-hidden">:</span>
```

As discussed, I think `govuk-pagination__link-divider` is now superfluous.

owenatgov:
@peteryates pointed out via x-gov slack that without whitespace control, the whitespace next to the prev/next links persists. I'm going to experiment with removing the svg from the `link-title` span to see if that solves this issue.

src/govuk/components/pagination/pagination.yaml
```diff
@@ -0,0 +1,243 @@
...
+    required: false
+    description: HTML attributes (for example data attributes) to add to the anchor.
+- name: landmarkLabel
+  type: string
+  required: false
+  description: The label for the navigation landmark that wraps the pagination. Defaults to "results".
+- name: items
+  type: array
+  required: false
+  description: Array of link objects.
```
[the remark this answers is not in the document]
owenatgov: I had a look at other arrays in govuk-frontend and we're kinda inconsistent. For example for the accordion tech docs, we add very little elaboration for the `items` array option:

```
An array of sections within the accordion.
```

However we elaborate much more in places like the cookie banner option `messages`:

```
The different messages you can pass into the cookie banner. For example, the cookie message and the confirmation message.
```

In this case I personally don't think we need to add an example because we're not doing anything unusual with this option. We're just assembling the pagination numbers, this feels quite easy to grasp to me.

src/govuk/components/pagination/_index.scss
```diff
@@ -0,0 +1,243 @@
...
+  }
+
+  .govuk-pagination__prev,
+  .govuk-pagination__next {
+    @include govuk-typography-weight-bold;
+    flex-basis: 100%;
+
+    @include govuk-media-query($from: tablet) {
+      flex-basis: auto;
+    }
```
36degrees, line 60: I don't _think_ this is needed any more. Before we switched to using `flex-direction: column;` on `.govuk-pagination` I think this was making the previous / next links expand to full width so that they were on separate lines to the page numbers, but now I don't think it's doing anything.

src/govuk/components/pagination/_index.scss
```diff
@@ -0,0 +1,243 @@
...
+  .govuk-pagination__item--ellipses,
+  .govuk-pagination__item:first-child,
+  .govuk-pagination__item:last-child {
+    display: block;
+  }
+
+  .govuk-pagination__item--current,
+  .govuk-pagination__item--ellipses {
+    @include govuk-typography-weight-bold;
+  }
```
36degrees, line 88: This is personal preference more than anything, so feel free to ignore, but not sure this optimisation is worth it versus having all of the styles 'in one place'.

I'd be inclined to add `@include govuk-typography-weight-bold;` to the existing `govuk-pagination__item--current` and `govuk-pagination__item--ellipses` rulesets and get rid of this, even though it means there's a little more duplication.

src/govuk/components/pagination/_index.scss
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    .govuk-pagination__link {
+      color: govuk-colour("white");
+
+      &:focus {
+        color: $govuk-focus-text-colour;
+      }
+    }
+  }
```
36degrees, line 104: Would suggest using the [`govuk-link-style-inverse`](https://frontend.design-system.service.gov.uk/sass-api-reference/#govuk-link-style-inverse) mixin here:

```suggestion
    .govuk-pagination__link {
      @include govuk-link-style-inverse;
    }
```

src/govuk/components/pagination/_index.scss
```diff
@@ -0,0 +1,243 @@
...
+
+    @include govuk-media-query($from: tablet) {
+      flex-basis: auto;
+    }
+
+    // Use flex to get around a whitespace issue between the arrow svg and the link text
+    // without having to rely on whitespace control from backend tooling
+    .govuk-pagination__link {
+      display: flex;
+    }
```
36degrees, line 66: Probably one for @christopherthomasdesign or @Ciandelle to weigh in on, but to my eyes the arrows on mobile are sitting too low relative to the text:

Screenshot 2022-05-20 at 11 17 21

I think removing the 5px top margin from the icon (only on mobile?) and possibly adding `align-items: baseline` to `govuk-pagination__link` gets us closer to something that looks right:

Screenshot 2022-05-20 at 11 19 58

src/govuk/components/pagination/template.njk
```diff
@@ -0,0 +1,71 @@
+{% set blockLevel = not params.items and (params.next or params.previous) %}
+
+<nav class="govuk-pagination {{- ' ' + params.classes if params.classes }} {{- ' govuk-pagination--block' if blockLevel }}" role="navigation" aria-label="{{ params.landmarkLabel | default("results") }} {%- for attribute, value in params.attributes %} {{ attribute }}="{{ value }}" {%- endfor -%}">
+  {%- if params.previous and params.previous.href -%}
+    <div class="govuk-pagination__prev">
+      <a class="govuk-link govuk-pagination__link" href="{{ params.previous.href }}" rel="prev" aria-label="{{ params.previous.visuallyHiddenText | default("Previous page") }}" {%- for attribute, value in params.previous.attributes %} {{ attribute }}="{{ value }}" {%- endfor -%}>
```
36degrees, line 6: I've just realised that because we're now using `aria-label` on the link, the `labelText` is no longer included in the accessible name for the link.

Over-simplifying somewhat, we currently end up with something like this:

```html
<a href="#" aria-label="Previous page">
    Previous<span class="govuk-visually-hidden">:</span> Title of previous page
</a>
```

So in the accessibility tree the label is just 'Previous page' 😕 

A couple of suggestions to get around this…

1. Don't use an `aria-label` on the previous/next links, as it's arguably less in need of the additional 'page' context than the page numbers are:
    ```html
    <a href="#">
        Previous<span class="govuk-visually-hidden">:</span> Title of previous page
    </a>
    ```

2. Change the default so that includes the `labelText` as well, which also means we can get rid of the visually hidden colon, but means that anyone customising `visuallyHiddenText` must also include the `labelText` themselves:

    ```html
    <a href="#" aria-label="Previous page: Title of previous page">
        Previous Title of previous page
    </a>
    ```

src/govuk/components/pagination/template.njk
```diff
@@ -0,0 +1,71 @@
...
+        <svg class="govuk-pagination__icon govuk-pagination__icon--prev" xmlns="http://www.w3.org/2000/svg" height="13" width="15" aria-hidden="true" focusable="false" viewBox="0 0 15 13">
+          <path d="m6.5938-0.0078125-6.7266 6.7266 6.7441 6.4062 1.377-1.449-4.1856-3.9768h12.896v-2h-12.984l4.2931-4.293-1.414-1.414z"></path>
+        </svg>
+        <span class="govuk-pagination__link-title {{- ' govuk-pagination__link-title--decorated' if blockLevel and not params.previous.labelText }}">
+          {%- if params.previous.text -%}
+            {{- params.previous.text -}}
+          {%- else -%}
+            Previous
+          {%- endif -%}
```
36degrees, line 15: Could use the `default` filter here to simplify:

```suggestion
          {{- params.previous.text | default("Previous") -}}
```

src/govuk/components/pagination/template.njk
```diff
@@ -0,0 +1,71 @@
...
+      <a class="govuk-link govuk-pagination__link" href="{{ params.next.href }}" rel="next" aria-label="{{ params.next.visuallyHiddenText | default("Next page") }}" {%- for attribute, value in params.next.attributes %} {{ attribute }}="{{ value }}" {%- endfor -%}>
+        {%- if blockLevel -%}
+          {{- nextArrow | safe -}}
+        {%- endif -%}
+        <span class="govuk-pagination__link-title {{- ' govuk-pagination__link-title--decorated' if blockLevel and not params.next.labelText }}">
+          {%- if params.next.text -%}
+            {{- params.next.text -}}
+          {%- else -%}
+            Next
+          {%- endif -%}
```
36degrees, line 59: Could use the `default` filter here to simplify:

```suggestion
          {{- params.next.text | default("Next") -}}
```

src/govuk/components/pagination/template.njk
```diff
@@ -0,0 +1,71 @@
...
+  {% endif %}
+
+  {%- if params.items -%}
+    <ul class="govuk-pagination__list">
+      {%- for item in params.items -%}
+        {%- if item.ellipsis -%}
+          <li class="govuk-pagination__item govuk-pagination__item--ellipses">&ctdot;</li>
+        {%- elseif item.number -%}
+          <li class="govuk-pagination__item {{- ' govuk-pagination__item--current' if item.current }}">
+            <a class="govuk-link govuk-pagination__link" href="{{ '' if item.current else item.href }}" aria-label="{{ item.visuallyHiddenText | default("Page " + item.number) }}" {%- if item.current %} aria-current="page" {% endif -%} {%- for attribute, value in item.attributes %} {{ attribute }}="{{ value }}" {%- endfor -%}>
```
36degrees, line 33: Have we decided on where the current link should go to? 

Generally I think if an `href` is set we should probably always use if rather than hard-coding the empty string for the current item.

36degrees, line 33: Really nit-picky but you end up with extra whitespace after `aria-current="page"` at the minute:

```html
<a class="govuk-link govuk-pagination__link" href="" aria-label="Page 1" aria-current="page" >
```


```suggestion
            <a class="govuk-link govuk-pagination__link" href="{{ '' if item.current else item.href }}" aria-label="{{ item.visuallyHiddenText | default("Page " + item.number) }}" {%- if item.current %} aria-current="page"{% endif -%} {%- for attribute, value in item.attributes %} {{ attribute }}="{{ value }}" {%- endfor -%}>
```

src/govuk/components/pagination/template.njk
```diff
@@ -0,0 +1,71 @@
...
+
+  {%- if params.next and params.next.href -%}
+    {%- set nextArrow -%}
+      <svg class="govuk-pagination__icon govuk-pagination__icon--next" xmlns="http://www.w3.org/2000/svg" height="13" width="15" aria-hidden="true" focusable="false" viewBox="0 0 15 13">
+        <path d="m8.107-0.0078125-1.4136 1.414 4.2926 4.293h-12.986v2h12.896l-4.1855 3.9766 1.377 1.4492 6.7441-6.4062-6.7246-6.7266z"></path>
+      </svg>
+    {%- endset -%}
+
+    <div class="govuk-pagination__next">
+      <a class="govuk-link govuk-pagination__link" href="{{ params.next.href }}" rel="next" aria-label="{{ params.next.visuallyHiddenText | default("Next page") }}" {%- for attribute, value in params.next.attributes %} {{ attribute }}="{{ value }}" {%- endfor -%}>
```
36degrees, line 50: This has the same issue with `aria-label` we have with the previous link.

src/govuk/components/pagination/template.test.js
```diff
@@ -0,0 +1,158 @@
...
+    it('passes accessibility tests', async () => {
+      const $ = render('pagination', examples.default)
+
+      const results = await axe($.html())
+      expect(results).toHaveNoViolations()
+    })
+
+    it('renders the correct URLs for each link', () => {
+      const $ = render('pagination', examples.default)
+      const previous = $('.govuk-pagination__prev .govuk-pagination__link')
```
36degrees, line 23: Generally when a variable references a Cheerio HTML element we prefix it with `$` in our tests, so this would be `$previous` for example.

src/govuk/components/pagination/template.test.js
```diff
@@ -0,0 +1,158 @@
...
+    // Instead of the aria-label solution used for the links in the pagination because of issues caused
+    // by aria-label on non-interactive elements like li's
+    it('marks up the current item correctly', () => {
+      const $ = render('pagination', examples.default)
+      const currentNumber = $('.govuk-pagination__item--current')
+      const currentNumberLink = currentNumber.find('.govuk-pagination__link')
+
+      expect(currentNumber).toBeTruthy()
+      expect(currentNumberLink.attr('aria-current')).toEqual('page')
+    })
```
36degrees, line 57: Not really sure what we get from testing for `currentNumber` being truthy – if it was falsey then `currentNumberLink` would also have to be falsey and getting the attribute would fail?

src/govuk/components/pagination/template.test.js
```diff
@@ -0,0 +1,158 @@
...
+      expect(previousSvg.attr('height')).toEqual('13')
+      expect(previousSvg.attr('width')).toEqual('15')
+      expect(previousSvg.attr('aria-hidden')).toEqual('true')
+      expect(previousSvg.attr('focusable')).toEqual('false')
+      expect(nextLink.attr('rel')).toEqual('next')
+      expect(nextSvg.attr('height')).toEqual('13')
+      expect(nextSvg.attr('width')).toEqual('15')
+      expect(nextSvg.attr('aria-hidden')).toEqual('true')
+      expect(nextSvg.attr('focusable')).toEqual('false')
+    })
```
36degrees, line 76: I'd suggest mimicking the tests we have in the header for the focusable and aria-hidden attributes and do something like this:

```
    it('sets focusable="false" so that IE does not treat it as an interactive element', () => {
      expect($svg.attr('focusable')).toEqual('false')
    })

    it('sets aria-hidden="true" so that it is ignored by assistive technologies', () => {
      expect($svg.attr('aria-hidden')).toEqual('true')
    })
```

I really like that we're testing the rel attributes, but not sure what we gain from testing width and height TBH.

36degrees (changes requested):
This is looking really good – thanks for all your hard work in.

Quite a lot of my comments are just suggestions or just tidying up minor things, although I have spotted an accessibility issue we've accidentally introduced in our refactor which needs addressing – hence requesting changes.

src/govuk/components/pagination/_index.scss
edwardhorsford: Agree with @36degrees - the arrows look of centre on the first image.

owenatgov: I've gone with removing the top margin altogether and using `align-items: center` as to my eye it centres it a little better.
Screenshot 2022-05-20 at 15 38 48

src/govuk/components/pagination/template.njk
owenatgov: We discussed this on slack and we've agreed to set `items.href` on all items including the current item to required and have no logic on this `href` as all items are now links and it is the logical option to ask teams to specify the link for each item, current or not.

owenatgov: Discussed this on slack. I've made the `href` option for all `item`'s required and we will be asking teams to pass `href` to all items, `current` or otherwise. We're doing this because now that the current item is a link, it makes sense for teams to treat it like a link and mange the link to the page they are currently on as part of the implementation.

src/govuk/components/pagination/template.test.js
owenatgov: I've re-worked it so that there are some more `describe`'s and it's segmented into more specific testing suites. LEt me know what you think.

src/govuk/components/pagination/pagination.yaml
```diff
@@ -0,0 +1,233 @@
...
+  description: A link to the next page, if there is a next page.
+  params:
+  - name: text
+    type: string
+    required: false
+    description: The link text to the next page. Defaults to 'Next page', where 'page' is visually hidden.
+  - name: labelText
+    type: string
+    required: false
+    description: The optional label that goes underneath the link to the next page providing further context for the user about where the link goes.
```
EoinShaughnessy, line 35:
```suggestion
    description: The optional label that goes underneath the link to the next page, providing further context for the user about where the link goes.
```

src/govuk/components/pagination/pagination.yaml
```diff
@@ -0,0 +1,233 @@
...
+  description: The array of link objects.
+  params:
+  - name: number
+    type: string
+    required: false
+    description: The pagination item text - usually a page number.
+  - name: visuallyHiddenText
+    type: string
+    required: false
+    description: The visually hidden label (for the pagination item() which will be applied to an `aria-label` and announced by screen readers on the pagination item link.
```
EoinShaughnessy, line 60:
```suggestion
    description: The visually hidden label (for the pagination item) which will be applied to an `aria-label` and announced by screen readers on the pagination item link.
```

EoinShaughnessy:
@owenatgov Macros content looks great! Before I send it off for 2i, just wanna check one small thing...

So, this content consists of 3 tables: macros-tables-in-google-doc

Will the second and third tables have titles of their own, like in the macros content for other components? sample-table-heading-from-cookie-banner

src/govuk/components/pagination/_index.scss
```diff
@@ -0,0 +1,234 @@
...
+    display: inline-block;
+    padding-left: govuk-spacing(6);
+  }
+
+  .govuk-pagination__icon {
+    // Set size using rems to make the icon scale with text if user resizes text in their browser
+    width: govuk-px-to-rem(15px);
+    height: govuk-px-to-rem(13px);
+    color: $govuk-secondary-text-colour;
+    fill: currentcolor;
```
36degrees, line 158: I think annoyingly you also need `forced-color-adjust: auto` to work around https://bugs.chromium.org/p/chromium/issues/detail?id=1164162.

The bug report suggests it might be fixed as of Chromium 102 (currently in beta) but in my testing it still has the same issue…

Screenshot 2022-05-23 at 11 57 34
Screenshot 2022-05-23 at 11 57 37

36degrees (commented):
Really nice to see this component coming together. I just have one question about the bottom margin and a couple of suggestions for improvements to the experience in forced colors mode.

src/govuk/components/pagination/_index.scss
```diff
@@ -0,0 +1,234 @@
+@include govuk-exports("govuk/component/pagination") {
+  // Flexbox enhancement for small screen visual design
+  // Falls back to a float: left layout on non-flex browsers
+  .govuk-pagination {
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    flex-wrap: wrap;
+    margin-bottom: govuk-spacing(4);
```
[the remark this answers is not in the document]
owenatgov: I decided to go for `govuk-responsive-margin(4, "bottom")` but I need to run this past designers. Will keep this issue unresolved until I get their insight.

owenatgov:
@EoinShaughnessy I've just updated the guidance PR to test this (https://github.com/alphagov/govuk-design-system/pull/2079) and the tables do have titles!

Screenshot 2022-05-24 at 10 48 37
Screenshot 2022-05-24 at 10 48 44
Screenshot 2022-05-24 at 10 48 29

src/govuk/components/pagination/_index.scss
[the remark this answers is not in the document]
owenatgov: Discussed with designers and yourself on slack and we've agreed on your original suggestion on the grounds that this is a consistent margin bottom convention we use on other components so I've used that.

Side note: does this convention need to be written down somewhere?

EoinShaughnessy:
@owenatgov Ahh, magic! Thanks for sharing that. Cool, I'll send the macro options Google Doc for 2i and will keep you posted about when it passes review/if the reviewer has questions I can't answer.

Oh, by the way, should the 'Options for items' table display after the 'Options for next' and 'Options for previous' tables? In my head, the 'next' and 'previous' tables' seem like sibling things. Whereas, the 'items' table seems like, I dunno (exposing my ignorance here), an extra level of config options? Also, I've seen 'Options for items' tables display as the last table in other components, like the accordion...

36degrees (approved):
Thanks @owenatgov! 🙌🏻

EoinShaughnessy:
Macro options still need to be 2i'd

src/govuk/components/pagination/template.njk
```diff
@@ -0,0 +1,62 @@
...
+  {% endif %}
+
+  {%- if params.items -%}
+    <ul class="govuk-pagination__list">
+      {%- for item in params.items -%}
+        {%- if item.ellipsis -%}
+          <li class="govuk-pagination__item govuk-pagination__item--ellipses">&ctdot;</li>
+        {%- elseif item.number -%}
+          <li class="govuk-pagination__item {{- ' govuk-pagination__item--current' if item.current }}">
+            <a class="govuk-link govuk-pagination__link" href="{{ item.href }}" aria-label="{{ item.visuallyHiddenText | default("Page " + item.number) }}" {%- if item.current %} aria-current="page" {%- endif -%} {%- for attribute, value in item.attributes %} {{ attribute }}="{{ value }}" {%- endfor -%}>
```
frankieroberto, line 28: Does this mean that the `visuallyHiddenText` for items should include the actual page number (eg `Result 1`) rather than just the word to replace "Page"? If so, it might be worth making this clear in the macro guidance?

src/govuk/components/pagination/pagination.yaml
```diff
@@ -0,0 +1,233 @@
...
+  description: The array of link objects.
+  params:
+  - name: number
+    type: string
+    required: false
+    description: The pagination item text - usually a page number.
+  - name: visuallyHiddenText
+    type: string
+    required: false
+    description: The visually hidden label (for the pagination item) which will be applied to an `aria-label` and announced by screen readers on the pagination item link.
```
EoinShaughnessy, line 60: Suggesting change based on [Frankie's feedback](https://github.com/alphagov/govuk-frontend/pull/2610/files#r881414308):

```suggestion
    description: The visually hidden label (for the pagination item) which will be applied to an `aria-label` and announced by screen readers on the pagination item link. Should include page number.
```

frankieroberto: 👍  thanks.

owenatgov: Good suggestion, implemented!

owenatgov:
@EoinShaughnessy I've managed to reorder the tables on the website by re-ordering the options in the tech docs in this repo. Tech docs are all sorted!

EoinShaughnessy (approved):

CHANGELOG.md
```diff
@@ -12,6 +12,14 @@ We've deprecated the `govuk-header__link--service-name` class, and will remove i
 
 This change was introduced in [pull request #2617: Do not make the service name in the header a link if no `serviceUrl` is provided](https://github.com/alphagov/govuk-frontend/pull/2617).
 
+### New features
+
+#### Help users navigate through pages with pagination
+
+You can now use [pagination](https://design-system.service.gov.uk/components/pagination/) to help users navigate forwards and backwards through a series of pages. For example, in search results or guidance that's divided ito multiple website pages.
```
owenatgov, line 19: The linked guidance here doesn't exist yet so this is a broken link, but this is the page it is going to be once it gets published. There's a decision here about if we want to merge this now or once the guidance is deployed.

vanitabarrett: We can merge this in, even with the broken link. While it's unreleased, no one should really be looking at / expect the Changelog to be fully working anyway. When we do the release, it does mean the release notes will contain a broken link for a short period of time until the guidance has been published, but we try to do them in quick succession to limit the length of time this happens. Ideally, most people will become aware of the release via our comms too, which we'll only put out once the guidance has also been published.

TLDR: it's all good to be merged, *after v4.1.1 has been released.*

vanitabarrett, line 19: Small typo @owenatgov  

```suggestion
You can now use [pagination](https://design-system.service.gov.uk/components/pagination/) to help users navigate forwards and backwards through a series of pages. For example, in search results or guidance that's divided into multiple website pages.
```

Issue #2559, opened 2022-02-28
Spike adding unit testing for our component JavaScript


vanitabarrett:
## What
Investigate how we might unit testing our component JavaScript

## Why
We don't currently have unit tests for our JavaScript, in the sense that we can't check/assert the output of specific functions within our component JavaScript code. Adding this level of testing would give us more confidence in the JavaScript code we write and add an extra level of checks if we're changing existing JavaScript code.

## Who needs to work on this
Developers

## Who needs to review this
Developers

## Done when

- [x] Investigate the best tool to use for writing unit tests
- [x] Investigate how these new tests would fit in with our current tests and testing strategy
- [ ] Write up follow-up cards

vanitabarrett:
Written up some thoughts from my initial investigation here: https://github.com/alphagov/govuk-frontend/pull/2611

vanitabarrett:
Moving this into the Backlog for now as I don't really have time to work on it this week.

Things I was hoping to do next (in case anyone else wants to pick this up):

- Adjust the spike for the new i18n singleton
- Think about our testing strategy if we adopt this, e.g: what we use unit tests vs integration tests for. Do we need the snapshots still?

The spike is very much 'what can we add with minimal changes to the codebase'. I'd like to think about what we'd do in an ideal world too, where we didn't have any existing tests restricting us, but that's probably a completely different spike.

vanitabarrett:
Going to convert this card into a 'spike' card, and then write up a new set of cards for follow-up work

vanitabarrett:
In the end, I've only written up one follow-up card to [define and document what we use the different tests for](https://github.com/alphagov/govuk-frontend/issues/2721). An acceptance criteria on that card will be to write up cards to add tests for our existing code.

We're also planning to try out using Jest for unit tests for the i18n logic, so I've made sure that's captured as acceptance criteria on the relevant localisation cards.

PR #2611, opened 2022-05-05
[SPIKE][DO NOT MERGE] Investigate using Jest for unit testing

vanitabarrett:
Investigation into https://github.com/alphagov/govuk-frontend/issues/2559

## What
Spike to see how we can add unit tests for our component JavaScript. This spike explores using tooling we already have (i.e: Jest) rather than introducing a new library.

## Findings / Thoughts

- Jest has [built-in support for ESM](https://jestjs.io/docs/27.x/ecmascript-modules) - however, it seems to rely on using the `.mjs` file extension (among other things) which complicates the current testing we have. You can also [use Babel](https://jestjs.io/docs/27.x/getting-started#using-babel) (Jest ships with `babel-jest`) to compile code, which is the approach used in this spike.
- Because we're initialising the component in each test, that gives us greater control over what it's initialised with. This lets us test different scenarios, which is a bit more tricky with our current test setup as it relies on the review app.
- A lot of our component JS does some sort of DOM manipulation, e.g: adding/removing attributes, creating new elements and appending them. It can be quite lengthy testing that, as we have to set up the right HTML elements in the DOM before running the test (as seen in this spike). An alternative approach might be making more use of mocks and spies to ensure the right things are being called with the right arguments, rather than actually checking the DOM.
- I'm not sure how well this approach works for something like the accordion JS at the moment, where a lot of functions create HTML elements but don't append them to the DOM, and probably also rely on previous execution of other functions (i.e: we're not passing things through as arguments to functions, but using globals). We'd probably want to do some refactoring before adding unit tests here.
- I think there's also a bit of overlap now between these tests and the other tests we have in place, so we would probably want to consider what exactly we want to use unit tests vs our existing JavaScript tests for.


vanitabarrett:
Update:

I've pulled in some commits from the i18n spike to see how this testing strategy fits with that. Broadly it seems to work fine - we can pull in and set translations within unit tests so have control over testing how component JS handles different translations. We can also add tests for the i18n function itself (note: this has been tested with the previous iteration of the i18n function, not the new version using the Singleton pattern).

Adding any kind of translations to our example app will require us to change existing tests that require the review app as the text will be different. Probably another reason why having tests based on our review app is perhaps not ideal.

vanitabarrett:
This spike is superseded by https://github.com/alphagov/govuk-frontend/pull/2680 which works with the latest (and hopefully more "stable" i18n approach) and the new `.mjs` component JS which affects how unit testing is set up.

Issue #1720, opened 2020-01-30
Prototype ways to access to "public" JavaScript methods and variables and to define methods/variables as "private" 


hannalaakso:
## What
Prototype different ways to add methods and variables to JavaScript public API or to set them private.

## Why
So we can get a better understanding of what would work in practice we want to build some examples to try before writing any proposals.

## Who needs to know about this
Developers

## Depends on

- [ ] [Understand how JavaScript public/private definitions work in existing projects (libraries, design systems))](https://github.com/alphagov/govuk-frontend/issues/1719)

## Done when

- [ ] We've tried viable approaches
- [ ] We're confident enough to write up a proposal

36degrees:
Closing, although we should make sure we're including this as part of our 'done when' criteria other JavaScript related work

PR #2614, opened 2022-05-11
[DO NOT MERGE] Internationalisation support

querkmachine:
A tech spike into internationalisation (i18n) support in GOV.UK Frontend, particularly within our JavaScript files, which are currently hardcoded. This spike specifically explores a bespoke solution for Frontend, which was the conclusion I came to following [the research and tech spikes into third-party solutions](https://docs.google.com/document/d/1zNjFKAZr2OJhH0HFeoVCI2qeaYU1e-dw0KST0V7XDOY/edit) (internal document).

## Specification

The spike was undertaken based on this rough, self-imposed specification:

- Must be able to pass translation strings from outside of Frontend (e.g. via JS inside of HTML).
- Must store localisation strings as key-value pairs.
- Must support interpolation of dynamic content within translation strings. 
- Must support alternative strings for plural forms.
  - Must support different pluralisation rules of different languages.
- Must support fallback ('default') text.
- Must support JS modules being individually initialised (not via `initAll`). 
- Optionally, allow localisation strings to be passed as nested JS objects.
- Optionally, allow implementors to customise which language's pluralisation rules are used.
- Optionally, allow implementors to customise how interpolated strings are detected (defaulting to `%{key_name}` otherwise, as GOV.UK uses).

## How it works

This is a rough, step-by-step of how the example implemention in this spike works. The actual i18n.js file is chock full of documentation too. It's worth a read for how individual methods work. 

### When using `initAll`

`initAll` accepts a configuration object where child objects are separated by the component name. This object is split up and each component passed the relevant part of the configuration.

Currently only two component namespaces are implemented:
- `accordion` for the Accordion component
- `character_count` for the Character Count component

Each component's configuration object may contain a nested `i18n` object, which includes information on the locale to use (used for pluralisation rules and number formatting) and a key-value based list of translations (the 'translation map'). 

This configuration object may also include other configuration in future (though that is outside of the scope of this work).

```js
window.GOVUKFrontend.initAll({
  accordion: {
    i18n: {
      locale: "cy",
      translations: {
        show_all_sections: "Dangoswch bob adran",
        hide_all_sections: "Cuddiwch bob adran"
      }
    }
  },
  character_count: {
    maxlength: 500
  }
})
```

### When initialising components separately

Individually initialised components use the same configuration object format as `initAll`, but with the component namespace removed.

```js
new window.GOVUKFrontend.Accordion(document.getElementById("accordion"), {
  i18n: {
    locale: "cy",
    translations: {
      show_all_sections: "Dangoswch bob adran",
      hide_all_sections: "Cuddiwch bob adran"
    }
  }
}).init()
```

### Continue...

1. Each component takes the provided configuration object and performs a _shallow merge_ with a default configuration included within the component. 
1. If internationalisation information has been provided (via an `i18n` key in the config object), each component creates a new instance of the I18n class and passes in the i18n object. 
2. `i18n.mjs` — Reads in the information given. 
3. `i18n.mjs` — If no locale information is given, it attempts to read this from `html[lang]`. If that is null or empty, it falls back to `"en"`. This is made available as `this.locale`. 
4. `i18n.mjs` — The translation map is made available as `this.translations`.  
5. Components may then call `t()` with a lookup key from the previously provided translation map and, optionally, an object containing additional placeholder data. The lookup key is used to find the associated translation string, which may be processed, before being returned to the calling component. See examples. 
 
## Examples 

Assume we've instantiated a non-existent example component with the following configuration for the Pirate language:

```js
new window.GOVUKFrontend.Example(
  $element,
  {
    animate_flags: false,
    i18n: {
      locale: "en",
      translations: {
        page_name: "Pirate's Cove Antique Treasures",
        welcome_user: "Welcome aboard, Cap'n %{name}",
        ship_count_one: "Ye got %{count} fine ship under ye command",
        ship_count_other: "Ye have %{count} ships under ye command",
      }
    }
  }
)
```

In the component's code, we pass the `i18n` information through to the I18n class, and assign the resulting instance to a variable we can use. 

```js
this.i18n = new I18n(options.i18n)
```

Using `t()` later in the component will result in these outputs. Note how in the last example that the number has been formatted appropriate to the given locale. 

```js
this.i18n.t("page_name")
// => Pirate's Cove Antique Treasures

this.i18n.t("welcome_user", { name: "Picard" })
// => Welcome aboard, Cap'n Picard

this.i18n.t("welcome_user", { count: 1 })
// => Ye got 1 fine ship under ye command

this.i18n.t("welcome_user", { count: 2000 })
// => Ye have 2,000 ships under ye command
```

## Currently...

Currently the spike supports:

- Localisations being passed into Nunjucks macros. 
- Localisations being passed globally to all components, via `initAll`.
- Localisations being passed to individual components, via each component's constructor.
- Key-value based localisation strings. 
- A specific method (`t`) for selecting a lookup key and interpolating dynamic content.
- Pluralisation support.
  - Uses unique, explicitly named keys for different plural forms (based on [Unicode CLDR's specification](https://cldr.unicode.org/index/cldr-spec/plural-rules)).
  - Has an internal reference of the pluralisation rules used by different languages. 
  - Also supports multiple plural forms in fallbacks.
- Fallback text defined within the component.
- Overriding the locale used to determine which pluralisation rules to use.

It doesn't support:

- Alternative grammar forms, gendered forms, or switching strings based on an arbitrary context.
- Bi-directional (Bidi) text strings (mixing left-to-right and right-to-left strings) without munging up formatting.
  - This can be handled manually by implementors by including the appropriate control characters ([LTR](https://en.wikipedia.org/wiki/Left-to-right_mark), [RTL](https://en.wikipedia.org/wiki/Right-to-left_mark)).
- Pluralisation for numbers that are negative, decimals or part of a range ("3–5 dogs"). 

Improvements that could still be made:

- Fallbacks that are 'mismatched' to what is expected aren't checked for (e.g. a pluralised fallback object for a non-pluralised string).
- The list of languages currently supported is pretty much arbitrary, and ones we don't need to support can be removed.


lfdebrux:
I had a good conversation with @querkmachine about this the other day. Noting the gist of it here in case it's useful/interesting for others.

I asked about whether she'd considered using [gettext](https://en.wikipedia.org/wiki/Gettext) style calling, where the key is the string in English. She said she had, but decided against it, as she pointed out that if we changed the text for a component, that would force a team with a translation to update that translation, which seems like an unnecessary pain for users, which I agree with.

That does though tangentially raise interesting questions about how we encourage conformity; are there ever cases where we will want to make a 'breaking change' to our content, to force teams with translations to update theirs? Of course, there is the wider point that if every team is doing their own translation, then we can expect their to be much consistency anyway.

Anyway, I think after discussing it, how @querkmachine has done it using a lookup table with simple keys is in fact the right way to go.

I've also been thinking of ways to reduce the need to pass around configuration, but I need to do some experiments with that first...

querkmachine:
It's been suggested that this version of the i18n code (or something similar to it) could be done in an IE-compatible v4 release. On this basis, I've made some minor changes and added polyfills to enable functionality back to IE8. 

After we stop shipping JavaScript to IE, we can then do a revision to remove polyfills and most of the pluralisation code, as we could then start to leverage the [Internationalization API's PluralRules](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/PluralRules) instead.

app/views/layouts/_generic.njk
```diff
@@ -51,5 +51,18 @@
   {% endif %}
 
   <script src="/public/all.js"></script>
-  <script>window.GOVUKFrontend.initAll()</script>
+  <script>window.GOVUKFrontend.initAll({
```
vanitabarrett, line 54: So I know I'm understanding correctly, can you explain/describe how a service that has to provide multiple different translations of it's site use this in practice? Would they need to swap out this translation object depending on the locale of the site?

vanitabarrett: Nice. Hopefully if there's any issues with that approach they'll be flagged when we get service teams to test this. It might be a good idea to get GOV.UK to test this (as a good example of a team needing to translate into many many languages)

querkmachine:
Based on a comment @lfdebrux made during a demo, I've tried modifying the function to use a singleton pattern. 

In this case, the 'default' I18n instance (which is usually the one created within `initAll`, otherwise the first one created) is made available to other components via a static `getInstance` method. This simplifies things a little, as we no longer need to pass a specific I18n instance around between components—the components can access the instance themselves. 

I've kept the ability to manually pass I18n instances into components on the assumption that a service might want to customise specific uses of a component differently to other uses of the same component.

querkmachine:
I've spent some time making the code a little more robust and working on some of the "Improvements that could still be made" bits. 

---

A component neglecting to give a translation string a key, or giving it a key that is falsy in JavaScript lingo, will now output an error in the console. 
  - This won't stop code execution however, as it'll continue to try and use the fallback text. If none is defined, it'll just output "UNDEFINED" (this could probably be better). 
  - This is a situation that is internal to the Design System JS and is very unlikely to occur in production unless there's some massive failure in our code review process—it's more just about making sure things doesn't fail quietly should the eventuality somehow arise. 

The `count` option is now checked to make sure that it's actually a number. 
  - As our current code can only handle integers and positive numbers at the moment, it also forces `count` through `abs` and `floor` to make sure it's a positive integer.
  - We can stop forcing it when we start using Intl.PluralRules, as that supports decimals and negative numbers. 
  - As far as I'm aware, no languages have plural rules specific to negative numbers anyway. 

Translation strings are now checked if they have placeholders before being passed through `replacePlaceholders`. 
  - [Stand back! I know regular expressions!](https://xkcd.com/208/)

I've removed the ability to customise placeholder separators. 
  - During testing I realised that, if a user customises the placeholder separators, that means that all of the placeholders in the fallback text would cease to work, which isn't ideal.
  - Having a different set of separators defined for passed-through translation strings versus fallback translation strings seemed like it would add a lot of complexity (in code and also in mentally parsing what's happening) for little benefit. 
  - Having customisable placeholder format was a 'nice to have' in my self-imposed spec anyway, so removing it doesn't seem terribly detrimental.

nickcolley (commented):
Non-blocking feedback:
1. Should we expand the public API? Recommendation: keep it private until a) there is evidenced user need b) it's been in production for a while
2. Is a consistent API for individually initialized components worthwhile?

(Feel free to disregard any and all feedback as I am a community member and just was interested in this work)

src/govuk/all.js
```diff
@@ -74,6 +81,7 @@ function initAll (options) {
 
 export {
   initAll,
+  I18n,
```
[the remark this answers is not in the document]
querkmachine: This makes sense to me. The reason `I18n` is currently exposed is because it's needed to support localising individually-initialised modules; but in conjunction with your other suggestion, I see no reason for making it private. 

Thanks for the suggestion!

src/govuk/components/character-count/character-count.js
```diff
@@ -1,9 +1,12 @@
+import I18n from '../../i18n'
 import '../../vendor/polyfills/Function/prototype/bind'
 import '../../vendor/polyfills/Event' // addEventListener and event.target normaliziation
 import '../../vendor/polyfills/Element/prototype/classList'
 
-function CharacterCount ($module) {
+function CharacterCount ($module, i18nInstance) {
```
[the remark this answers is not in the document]
querkmachine: I really dig the idea of having each component take a generic options object. Probably something to tie into #1720.

querkmachine:
@vanitabarrett and I had a discussion about how this work will be covered by automated tests. Currently tests are ran against the review app, however the app currently uses the same global `initAll` call for all of the examples, and it's not simple to have some examples use this and others not within the current structure. 

Instead, we will unit test the internationalisation JavaScript independently from the review app (as per Vanita's spike, #2611), and provide an example page in the app for visual checks and manual tests.

querkmachine:
The PR has been rebased to include the Nunjucks localisation work from #2648. 

An example of localised components in action (including localisation performed both via Nunjucks and via JavaScript) can be found here: https://govuk-frontend-pr-2614.herokuapp.com/full-page-examples/localisation

app/views/full-page-examples/localisation/index.njk
```diff
@@ -0,0 +1,233 @@
...
+
+      <!-- The Related items component is not part of GOV.UK Frontend but will be styled if used in the Prototype Kit -->
+
+      <aside class="app-related-items" role="complementary">
+        <h2 class="govuk-heading-m" id="subsection-title">
+          Cynnwys cysylltiedig
+        </h2>
+        <nav role="navigation" aria-labelledby="subsection-title">
+          <ul class="govuk-list govuk-!-font-size-16">
+            <li class="gem-c-related-navigation__link">
```
domoscargin, line 177: Think these gem-c classes have snuck in.

src/govuk/components/footer/footer.yaml
```diff
@@ -67,6 +67,22 @@ params:
       type: object
       required: false
       description: HTML attributes (for example data attributes) to add to the anchor in the footer navigation section.
+- name: contentLicence
+  type: string
+  required: false
+  description: HTML or text to replace the default Open Government Licence notice with.
```
domoscargin, line 73: Is it potentially a slight issue that we split HTML and text into their own properties on other fields, but not here?

app/views/full-page-examples/localisation/index.njk
querkmachine: I copied this page from one of the other full-page examples, though I'm happy to remove given it's pretty much decorative here.

src/govuk/components/footer/footer.yaml
```diff
@@ -67,6 +67,22 @@ params:
...
+  required: false
+  description: Removes the Open Government Licence logo, if the licence being used isn't the OGL.
+- name: copyright
+  type: string
+  required: false
+  description: Text to replace the default Crown Copyright notice with.
+- name: hideCrownCopyrightArms
+  type: boolean
+  required: false
+  description: Removes the Crown Copyright notice and royal coat of arms.
```
domoscargin, line 85: Would be good to clarify what will be required and suggested in this case. Does the Design System/GDS care if users don't add a new copyright/license to replace these?

src/govuk/components/footer/footer.yaml
querkmachine: I figured splitting it up might not be necessary as it's fairly likely to be HTML. There are other parameters that aren't split (like `visuallyHiddenText`), though that does include 'text' in the name. 

Would it be better named as `contentLicenceHtml` or do you think it's better to split it up?

src/govuk/i18n.js
```diff
@@ -0,0 +1,418 @@
...
+}
+
+/**
+ * The most used function - looks up translation strings and sorts what
+ * operations to perform on them.
+ *
+ * ALL properties passed to `options` will be exposed as options for string
+ * interpolation, but only `count` and `fallback` initiate special functionality.
+ *
+ * TODO: Do we want to put string interpolation options in their own object (e.g. `options.data`) to avoid potential conflict in future?
```
domoscargin, line 55: Useful from the perspective of users not introducing breaking changes in the future, though p'raps a bit more mental load for users to take on right now. I wouldn't be against this.

src/govuk/components/footer/footer.yaml
querkmachine: I'm not sure how necessary that is, might be a question for policy? 

For public-facing GOV.UK services, the default (or a translation of the default) is expected, I imagine. 

For internal/caseworking systems, they're more likely to remove or change the content licence and potentially remove the coat of arms for space purposes. 

For third-parties that use Frontend as a boilerplate, they'll probably remove or replace all of it.

src/govuk/components/footer/footer.yaml
querkmachine: `copyright` is a more awkward example of this, I think.

The coat of arms is set as the background image of the crown copyright link. As such, a user cannot use HTML to set a custom link unless `hideCrownCopyrightArms: true` (which removes the link and thus the arms), in which case they can. The typical input then switches between being HTML or plain text depending on that context.

querkmachine:
I've made a fork of this work based on a suggestion made by @36degrees, for consideration as an alternative approach.

You can compare it against this PR here: https://github.com/alphagov/govuk-frontend/compare/spike-i18n-support...spike-i18n-support-alt

The end result is identical, however the way that localisation strings are configured and scoped is different.

Instead of strings being defined within a `i18n` object which is separated by component, `i18n` is instead now a possible parameter within each component's individual options object, essentially reversing the prior relationship. 

```diff
- { i18n: { accordion: { translations: { show_all_sections: "Show all sections" } } } }
+ { accordion: { i18n: { translations: { show_all_sections: "Show all sections" } } } }
```

Localisations are now, essentially, part of a component's own configuration, along with any other configuration options. This allows for future configuration options to be nested within the same component object, or for the wrapping object to be discarded entirely when it's unnecessary (for example, when passing a configuration directly into a component).

```diff
new Accordion($element, {
  remember_state: false,
  i18n: {
-   accordion: {
-     translations: {
-       show_all_sections: "Show all sections"
-     }
-   }
+   translations: {
+     show_all_sections: "Show all sections"
+   }
  }
})
```

Other changes:
- The singleton pattern has been removed, as each component now instantiates its own I18n instance. Similarly, an I18n instance is no longer created by `initAll`.
- The `fallback` parameter has been removed, as these are now part of a component's default configuration options. They're now defined as part of `defaultOptions`. 
- The 'flattening' functionality for nested translation objects has been removed as it provides less benefit in this pattern.
- Previously, the `locale` parameter was only provided once, which propagated to all components. This now has to be specified for each component (unless the document language is being used).

## Comparisons

<details>
<summary>Using `initAll` to pass translation strings to multiple components at once</summary>

Old:

```html
<script>
  window.GOVUKFrontend.initAll({
    i18n: {
      locale: "cy",
      translations: {
        accordion: {
          show_all_sections: "Dangos pob adran",
          hide_all_sections: "Cuddiwch bob adran",
          show_this_section: "Dangoswch yr adran hon",
          hide_this_section: "Cuddio'r adran hon"
        },
        character_count: {
          characters_over_limit_zero: "Rydych chi %{count} llythyrau drosodd",
          characters_over_limit_one: "Rydych chi %{count} llythyr drosodd",
          characters_over_limit_two: "Rydych chi %{count} lythyren drosodd",
          characters_over_limit_few: "Rydych chi %{count} llythyren drosodd",
          characters_over_limit_many: "Rydych chi %{count} llythyren drosodd",
          characters_over_limit_other: "Rydych chi %{count} o lythyrau drosodd",
          characters_under_limit_zero: "Mae gennych %{count} llythyren ar ôl",
          characters_under_limit_one: "Mae gennych %{count} llythyr ar ôl",
          characters_under_limit_two: "Mae gennych %{count} lythyren ar ôl",
          characters_under_limit_few: "Mae gennych %{count} llythyren ar ôl",
          characters_under_limit_many: "Mae gennych %{count} llythyr ar ôl",
          characters_under_limit_other: "Mae gennych %{count} llythyren ar ôl"
        }
      }
    }
  })
</script>
```

New:

```html
<script>
  window.GOVUKFrontend.initAll({
    accordion: {
      i18n: {
        locale: "cy",
        translations: {
          show_all_sections: "Dangos pob adran",
          hide_all_sections: "Cuddiwch bob adran",
          show_this_section: "Dangoswch yr adran hon",
          hide_this_section: "Cuddio'r adran hon"
        }
      }
    },
    character_count: {
      i18n: {
        locale: "cy",
        translations: {
          characters_over_limit_zero: "Rydych chi %{count} llythyrau drosodd",
          characters_over_limit_one: "Rydych chi %{count} llythyr drosodd",
          characters_over_limit_two: "Rydych chi %{count} lythyren drosodd",
          characters_over_limit_few: "Rydych chi %{count} llythyren drosodd",
          characters_over_limit_many: "Rydych chi %{count} llythyren drosodd",
          characters_over_limit_other: "Rydych chi %{count} o lythyrau drosodd",
          characters_under_limit_zero: "Mae gennych %{count} llythyren ar ôl",
          characters_under_limit_one: "Mae gennych %{count} llythyr ar ôl",
          characters_under_limit_two: "Mae gennych %{count} lythyren ar ôl",
          characters_under_limit_few: "Mae gennych %{count} llythyren ar ôl",
          characters_under_limit_many: "Mae gennych %{count} llythyr ar ôl",
          characters_under_limit_other: "Mae gennych %{count} llythyren ar ôl"
        }
      }
    }
  })
</script>
```

</details>

<details>
<summary>Passing translation strings when individually instantiating a component</summary>

Old:

```html
<script>
  new window.GOVUKFrontend.Accordion($element, {
    i18n: {
      locale: "cy",
      translations: {
        accordion: {
          show_all_sections: "Dangos pob adran",
          hide_all_sections: "Cuddiwch bob adran",
          show_this_section: "Dangoswch yr adran hon",
          hide_this_section: "Cuddio'r adran hon"
        }
      }
    }
  }).init()
</script>
```

New:

```html
<script>
  new window.GOVUKFrontend.Accordion($element, {
    i18n: {
      locale: "cy",
      translations: {
        show_all_sections: "Dangos pob adran",
        hide_all_sections: "Cuddiwch bob adran",
        show_this_section: "Dangoswch yr adran hon",
        hide_this_section: "Cuddio'r adran hon"
      }
    }
  }).init()
</script>
```

</details>

vanitabarrett:
Closing this PR as it contains code developed for the internationalisation spike, which we've completed. The branch will remain so we can refer back to it and/or generate pre-releases as needed. I'm going to draft some new issues that break down the work into sizeable chunks we can start working on implementing this with a release in mind.

PR #2615, opened 2022-05-12
Fix hints for disabled checkboxes/radios appearing darker than the associated labels

### LINKED ISSUE #2581: Hints on disabled checkboxes and radios are styled darker than the label
If you set a checkbox to `disabled` the label of the checkbox is made a light grey. However I'd also expect the hint to be made at least the same matching grey. At the moment it doesn't seem to be changed - making a weird appearance as seen below.

Set a checkbox item to `disabled` that also includes a hint.

What the design system produces:
Screenshot 2022-03-29 at 16 13 12

What I'd expect:
Screenshot 2022-03-29 at 16 31 26

or if we want to make the hint 50% transparent:
Screenshot 2022-03-29 at 16 28 59

36degrees:
When checkboxes or radios with hints are disabled, the hint text ends up being darker than the label because the opacity of the label is reduced to 50%, but the opacity of the hint is not changed and so remains a darker grey.

Reduce the opacity of the hints associated with disabled radios and checkboxes so that it is also ‘greyed out’ proportional to the label.

Update the examples for disabled radios and checkboxes to include hints so that we have a visual example of this in the review app.

## Before

Screenshot 2022-05-12 at 17 11 55

Screenshot 2022-05-12 at 17 11 49

## After

Screenshot 2022-05-12 at 17 12 10

Screenshot 2022-05-12 at 17 12 08

Fixes #2581.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,6 +73,7 @@ We’ve also made fixes in the following pull requests:
 - [#2549: Fix header with product name focus and hover state length](https://github.com/alphagov/govuk-frontend/pull/2549)
 - [#2573: Better handle cases where `$govuk-text-colour` is set to a non-colour value](https://github.com/alphagov/govuk-frontend/pull/2573)
 - [#2590: Remove `maxlength` attribute from `textarea` after character count JavaScript has been initialised](https://github.com/alphagov/govuk-frontend/pull/2590)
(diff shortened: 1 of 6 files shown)


querkmachine (approved):
This all makes sense to me 👍

Ciandelle (approved):
Happy with the change

PR #2616, opened 2022-05-13
Allow selecting options by passing current values

### LINKED ISSUE #2449: Allow selecting / checking values in checkboxes, radios and select by passing current value(s)
## Related component

Checkboxes, radios and select

## Context

We should consider allowing the user to pass the current value (or array of values, for checkboxes?), and using that to work out which item should be checked / selected.

For example, by introducing a new top level `value` param:

```
{{ govukSelect({
    id: "sort",
    name: "sort",
    label: {
    text: "Sort by"
    },
    items: [
        {
            value: "published",
            text: "Recently published"
        },
        {
            value: "updated",
            text: "Recently updated"
        },
        {
            value: "views",
            text: "Most views"
        },
        {
            value: "comments",
            text: "Most comments"
        }
    ],
    value: data['sort']
}) }}
```

## Alternatives

At the minute the only way to pre-select or check specific items when using the Nunjucks macros for these components is to pass a boolean flag for each item, which then maps directly to the `checked` / `selected` HTML attribute for the `<input>` / `<option>`.

This means the user has to do the work to map the previous form data back to a boolean, resulting in code that might look something like this:

```njk
{{ govukSelect({
    id: "sort",
    name: "sort",
    label: {
        text: "Sort by"
    },
    items: [
        {
            value: "published",
            text: "Recently published",
            selected: (data['sort'] == "published")
        },
        {
            value: "updated",
            text: "Recently updated",
            selected: (data['sort'] == "updated")
        },
        {
            value: "views",
            text: "Most views",
            selected: (data['sort'] == "views")
        },
        {
            value: "comments",
            text: "Most comments",
            selected: (data['sort'] == "comments")
        }
    ]
}) }}
```

This is prone to introducing errors (especially when prototyping) – for example, any changes to the `name` of the field or the `value` of the `item` also need corresponding changes made to the boolean logic, potentially in multiple places.

## Additional information (if applicable)

We should consider providing this in addition to the existing boolean attribute – if so, we'll need to be clear how the two interact with each other.

We should consider how this interacts with the `checked` helper in the Prototype Kit.

36degrees:
At the minute the only way to pre-select or check specific items when using the Nunjucks macros for these components is to pass a boolean flag for each item, which then maps directly to the checked / selected HTML attribute for the <input> / <option>.

This means the user has to do the work to map the previous form data back to a boolean, resulting in code that might look something like this:

```njk
{{ govukSelect({
    id: "sort",
    name: "sort",
    label: {
        text: "Sort by"
    },
    items: [
        {
            value: "published",
            text: "Recently published",
            selected: (data['sort'] == "published")
        },
        {
            value: "updated",
            text: "Recently updated",
            selected: (data['sort'] == "updated")
        },
        {
            value: "views",
            text: "Most views",
            selected: (data['sort'] == "views")
        },
        {
            value: "comments",
            text: "Most comments",
            selected: (data['sort'] == "comments")
        }
    ]
}) }}
```

This is prone to introducing errors (especially when prototyping) – for example, any changes to the name of the field or the value of the item also need corresponding changes made to the boolean logic, potentially in multiple places.

Allow the user to pass the current value (or array of values, for checkboxes), and using that to work out which item(s) should be checked or selected.

The above example can then be done like this:

```njk
{{ govukSelect({
    id: "sort",
    name: "sort",
    label: {
        text: "Sort by"
    },
    items: [
        {
            value: "published",
            text: "Recently published"
        },
        {
            value: "updated",
            text: "Recently updated"
        },
        {
            value: "views",
            text: "Most views"
        },
        {
            value: "comments",
            text: "Most comments"
        }
    ],
    selectedValue: data[‘sort’]
}) }}
```

Naming the new option has presented a few challenges – I wanted to avoid using `checked` or `selected` so that we don’t have options at different ‘levels’ with the same name but accepting different types of values – booleans for individual items and strings / arrays at the top level.

I also felt that `value` was not an accurate description for what the option did, and may also be confused with `item.value` (and the top-level `value` option on other form controls such as text inputs).

I’ve ended up combining the two whilst also trying to be consistent with the underlying HTML attribute, and also reflecting whether the option accepts a string or an array. This does unfortunately mean that the option name is different for each of the three components:

- `checkedValue` for Radios, where a string is passed to check a single radio button
- `checkedValues` for Checkboxes, where an array of strings is passed to check one or more checkboxes
- `selectedValue` for the Select, where a string is passed to select a single option.

I think on balance, although I can see how this may be confusing for users, it is logical and consistent within each individual component.

Closes #2449.


CHANGELOG.md
```diff
@@ -4,6 +4,48 @@
...
+
+```
+checkedValue: 'foo'
+```
+
+This is an alternative to setting the `checked` option on each individual radio item.
+
+This change was introduced in [pull request #2616: Allow selecting options by passing current values](https://github.com/alphagov/govuk-frontend/pull/2616).
+
+#### Select an option in a Select by using the `selectedValue` option
```
36degrees, line 35: THIS IS A NIGHTMARE. Two of these words have two different meanings and we use both meanings in the same heading.

**Select** (verb) an **option** (an `<option>` HTML element within a `<select>`) in a **Select** (component name / HTML element name) using the `selectedValue` **option** (a 'Nunjucks macro option')

Suggestions welcome.

CHANGELOG.md
```diff
@@ -4,6 +4,48 @@
 
 ### New features
 
+#### Check checkboxes by using the `checkedValues` option
```
36degrees, line 7: I've opted to document this as three separate (but very similar) features because the words (checked / selected) and the options (`checkedValues` / `checkedValue` / `selectedValue`) are different in each case and trying to write it is a single feature got messy quickly.

36degrees:
I've raised a PR with a proof of concept of how this could be extended in the Prototype Kit to allow pre-selecting radios and checkboxes from the `data` object without the user having to do anything, which seems kinda neat: https://github.com/alphagov/govuk-prototype-kit/pull/1316

CHANGELOG.md
EoinShaughnessy: Makes sense!

Can see from line 35 that 'option' is a bit of a problem term in this content...

Wonder could we replace 'option' with 'parameter' throughout, whenever we refer to `checkedValues`, etc.? Or would we not want to do that, because of the PR a few days ago that made 'option' the word we use for these things?

CHANGELOG.md
```diff
@@ -4,6 +4,48 @@
...
+
+When using the Checkboxes nunjucks macro you can now use `checkedValues` option to indicate which checkboxes should be checked when the page loads.
+
+For example, this would check the checkboxes that have the values 'foo' and 'bar':
+
+```
+checkedValues: ['foo', 'bar']
+```
```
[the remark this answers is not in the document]
EoinShaughnessy: Maybe this suggestion is overkill considering it would just be for one update, _but..._could we add a full example, plus the existing content, to a Google Doc, then ask a few frontend users for their thoughts?

CHANGELOG.md
```diff
@@ -4,6 +4,48 @@
...
+
+This is an alternative to setting the `checked` option on each individual checkbox item.
```
EoinShaughnessy, line 17:
```suggestion
Use this kind of code so you do not have to set the `checked` parameter on each individual checkbox item.
```
Is the word 'item' adding anything here?

CHANGELOG.md
```diff
@@ -4,6 +4,48 @@
...
+
+This change was introduced in [pull request #2616: Allow selecting options by passing current values](https://github.com/alphagov/govuk-frontend/pull/2616).
+
+#### Check a radio button by using the `checkedValue` option
+
+When using the Radios nunjucks macro you can now use `checkedValue` option to indicate which radio should be checked when the page loads.
```
EoinShaughnessy, line 23:
```suggestion
When using the Nunjucks macro for radios, you can now use the `checkedValue` parameter to indicate which radio should be checked when the page loads.
```

CHANGELOG.md
```diff
@@ -4,6 +4,48 @@
...
+
+For example, this would check the radio that has the values 'foo':
```
EoinShaughnessy, line 25:
```suggestion
For example, this code would check the radio that has the value 'foo':
```

CHANGELOG.md
```diff
@@ -4,6 +4,48 @@
...
+
+```
+checkedValue: 'foo'
+```
+
+This is an alternative to setting the `checked` option on each individual radio item.
```
EoinShaughnessy, line 31:
```suggestion
Use this kind of code so you do not have to set the `checked` parameter on each individual radio item.
```
Do we have to keep 'item'?

CHANGELOG.md
EoinShaughnessy: Tricky one alright! Ok, first pass:

> Choose an option in a select component by using the `selectedValue` parameter

CHANGELOG.md
```diff
@@ -4,6 +4,48 @@
...
+
+This change was introduced in [pull request #2616: Allow selecting options by passing current values](https://github.com/alphagov/govuk-frontend/pull/2616).
+
+#### Select an option in a Select by using the `selectedValue` option
+
+When using the Select nunjucks macro you can now use `selectedValue` option to indicate which option should be selected when the page loads.
```
EoinShaughnessy, line 37:
```suggestion
When using the Nunjucks macro for the select component, you can now use the `selectedValue` parameter to indicate which option should be selected when the page loads.
```

CHANGELOG.md
```diff
@@ -4,6 +4,48 @@
...
+
+For example, this would select the option that has the value 'foo':
+
+```
+selectedValue: 'foo'
+```
+
+This is an alternative to setting the `select` option on each individual option item.
```
EoinShaughnessy, line 45:
```suggestion
Use this kind of code so you do not have to set the `select` parameter on each individual option item.
```
Could we remove the word 'item', or does it add something that 'individual option' doesn't?

EoinShaughnessy (commented):
Left a few small comments. Content looks good though! Separating it out was a neat idea.

lfdebrux:
> Naming the new option has presented a few challenges – I wanted to avoid using checked or selected so that we don’t have options at different ‘levels’ with the same name but accepting different types of values – booleans for individual items and strings / arrays at the top level.
> 
> I also felt that value was not an accurate description for what the option did, and may also be confused with item.value (and the top-level value option on other form controls such as text inputs).

I'd argue strongly for calling it `value`. I agree that it is not a completely accurate description, because users will be used to `value` from other macros, using `value` here will be the [least surprising](https://en.wikipedia.org/wiki/Principle_of_least_astonishment), in my opinion. I think having the parameter named different for different macros will add friction, and I'm not sure whether there would be a benefit to it.

CHANGELOG.md
36degrees: It might be to save the user time, or more likely it might be because the user has gone back to change an answer they previously gave.

For example, if a user was in the middle of a transaction and went 'back' to a previous page that used a select, the answer they previously gave should be selected when the page loads.

Likewise, if the user was [checking their answers](https://design-system.service.gov.uk/patterns/check-answers/) and decided to change an answer they previously gave, when they click change the page that loads should pre-select the answer that they'd previously given.

CHANGELOG.md
```diff
@@ -4,6 +4,48 @@
 
 ### New features
 
+#### Check checkboxes by using the `checkedValues` option
+
+When using the Checkboxes nunjucks macro you can now use `checkedValues` option to indicate which checkboxes should be checked when the page loads.
```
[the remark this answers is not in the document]
36degrees: I think 'include' is too passive in this instance – the user can't just add the code `checkedValues` to the macro call and it'd work, they need to also pass a value as well. Does that make sense?

CHANGELOG.md
```diff
@@ -4,6 +4,48 @@
...
+
+For example, this would check the checkboxes that have the values 'foo' and 'bar':
```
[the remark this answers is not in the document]
36degrees: Possibly not. I always struggle to come up with neat, simple and self-contained examples that don't rely on nonsense placeholder values… will have a think!

CHANGELOG.md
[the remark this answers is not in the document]
EoinShaughnessy: It does! Thanks. Let's leave as is.

CHANGELOG.md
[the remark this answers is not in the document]
36degrees: I've replaced 'foo' and 'bar' with 'red' and 'blue' so although the examples are still a little contrived they're at least terms that a user should be familiar with!

CHANGELOG.md
36degrees: I've tweaked these slightly to put the code inline, like:

> For example, `value: 'red'` would check the radio that has a `value` of 'red'.

src/govuk/components/radios/template.njk
```diff
@@ -56,11 +56,12 @@
         {%- if item.divider %}
         <div class="govuk-radios__divider">{{ item.divider }}</div>
         {%- else %}
+        {% set isChecked = item.checked or (params.value and item.value == params.value) %}
```
edwardhorsford, line 59: I'd recommend checking `item.checked` for undefined, not truthiness. If it's explicitly false you want that stored and it not to fall to the other conditions.

src/govuk/components/select/template.njk
```diff
@@ -42,7 +42,7 @@
   {% for item in params.items %}
     {% if item %}
       <option value="{{ item.value }}"
-        {{-" selected" if item.selected }}
+        {{-" selected" if item.selected or (params.value and item.value == params.value) }}
```
edwardhorsford, line 45: I'd recommend checking `item.selected` for undefined, not truthiness. If it's explicitly false you want that stored and it not to fall to the other conditions.

edwardhorsford (commented):
Nice work! This is a bit similar to a [filter we use](https://github.com/DFE-Digital/register-trainee-teachers-prototype/blob/main/app/filters/utility.js#L72) to make the macros simpler.

Note you probably want the ability for a single item to have `checked: false` - at the moment the logic will overwrite these.

src/govuk/components/radios/template.njk
edwardhorsford: In js that would be the equivalent of the [Nullish coalescing operator (??)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) but in Nunjucks I *think* this could be:
```suggestion
        {% set isChecked = (item.checked if (item.checked !== undefined) else (params.value and item.value == params.value) %}
```

CHANGELOG.md
```diff
@@ -2,6 +2,38 @@
...
+
+For example, `value: ['red', 'blue']` would check any checkboxes that have a `value` of 'red' or 'blue'.
+
+This is an alternative to setting the boolean `checked` option on each individual checkbox.
+
+This change was introduced in [pull request #2616: Allow selecting options by passing current values](https://github.com/alphagov/govuk-frontend/pull/2616).
+
+#### Check a radio button by using the `value` Nunjucks option
+
+When using the `govukRadios` Nunjucks macro you can now use `value` option to indicate which radio should be checked when the page loads.
```
EoinShaughnessy, line 19:
```suggestion
When using the `govukRadios` Nunjucks macro, you can now use the `value` option to indicate which radio should be checked when the page loads.
```

CHANGELOG.md
```diff
@@ -2,6 +2,38 @@
...
+
+For example, `value: 'red'` would check the radio that has a `value` of 'red'.
+
+This is an alternative to setting the boolean `checked` option on each individual radio.
+
+This change was introduced in [pull request #2616: Allow selecting options by passing current values](https://github.com/alphagov/govuk-frontend/pull/2616).
+
+#### Select an option in a select by using the `value` Nunjucks option
+
+When using the `govukSelect` Nunjucks macro you can now use `value` option to indicate which option should be selected when the page loads.
```
EoinShaughnessy, line 29:
```suggestion
When using the `govukSelect` Nunjucks macro, you can now use the `value` option to indicate which option should be selected when the page loads.
```

src/govuk/components/select/select.yaml
```diff
@@ -32,6 +32,10 @@ params:
     type: object
     required: false
     description: HTML attributes (for example data attributes) to add to the option.
+- name: value
+  type: string
+  required: false
+  description: Value for the option which should be selected. This can be used as an alternative to setting the `selected` option on each individual item.
```
EoinShaughnessy, line 38:
```suggestion
  description: Value for the option which should be selected. Use this as an alternative to setting the `selected` option on each individual item.
```

src/govuk/components/checkboxes/checkboxes.yaml
```diff
@@ -98,6 +98,10 @@ params:
     type: object
     required: false
     description: HTML attributes (for example data attributes) to add to the checkbox input tag.
+- name: values
+  type: array
+  required: false
+  description: Array of values for checkboxes which should be checked when the page loads. This can be used as an alternative to setting the `checked` option on each individual item.
```
EoinShaughnessy, line 104:
```suggestion
  description: Array of values for checkboxes which should be checked when the page loads. Use this as an alternative to setting the `checked` option on each individual item.
```

CHANGELOG.md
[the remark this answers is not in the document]
36degrees: Are you happy with the examples using red, green and blue?

CHANGELOG.md
36degrees: I've removed the word item, but have kept the overall sentence structure as I think the suggested wording pushes users too much towards using `values` when both approaches are equally valid. Very happy to talk this one through though if we do want to change it.

EoinShaughnessy (approved):

src/govuk/components/checkboxes/checkboxes.yaml
```diff
@@ -77,7 +77,7 @@ params:
   - name: checked
     type: boolean
     required: false
-    description: If `true`, checkbox will be checked.
+    description: If `true`, the checkbox will be checked when the page loads.
```
owenatgov, line 80: Is it worth noting that `checked` takes precedent over `values` if set? Same applies to radios and selects. Maybe something like:

```suggestion
    description: If `true`, the checkbox will be checked when the page loads. Setting this option in the checkbox item itself will override any data related to this checkbox item's value in the `values` option, if `values` is set.
```

A bit of a mouthful, I'm sure @EoinShaughnessy can do a better job.

src/govuk/components/radios/radios.yaml
```diff
@@ -69,7 +69,7 @@ params:
   - name: checked
     type: boolean
     required: false
-    description: If `true`, radio will be checked.
+    description: If `true`, the radio will be checked when the page loads.
```
owenatgov, line 72: Same comment on communicating the override

src/govuk/components/radios/radios.yaml
```diff
@@ -380,6 +383,46 @@ examples:
           <label class="govuk-label" for="contact-text-message">Mobile phone number</label>
           <input class="govuk-input govuk-!-width-one-third" name="contact-text-message" type="text" id="contact-text-message">
 
+- name: prechecked
```
owenatgov, line 386: Is it worth adding an example just for demonstrating the `value` option? You've added a couple here but my read of this is that they're for demonstrating different things alongside the `value` option over demonstrating the.. er... value of the `value` option on its own as you have done in the checkboxes example.

src/govuk/components/select/select.yaml
```diff
@@ -23,7 +23,7 @@ params:
   - name: selected
     type: boolean
     required: false
-    description: Sets the option as the selected.
+    description: Sets the option as selected when the page loads.
```
owenatgov, line 26: Same comment on communicating the override

owenatgov (commented):
Looking really nice! I've left a couple of comments about tech docs. The solution itself is very slick 👍🏻

src/govuk/components/checkboxes/checkboxes.yaml
36degrees: What about something like this?

```suggestion
    description: Whether the checkbox should be checked when the page loads. If this option is set the checked state of the checkbox will not be affected by the `value` option.
```

36degrees: Thanks Eoin, that makes sense.

This is one place where I really dislike that we now have a `value` option at the top-level and a `value` option on each item, and they do different things. I think in this context it's not clear which one we're talking about.

I'm wondering if we said 'the top-level `value` option will not affect…' – would that make sense? Is there a better way to differentiate between the two options?

EoinShaughnessy: Yeah, I see what you mean - would the user know what we meant by 'top-level'?

Sorry for daft question - where in the checkboxes macro options here is the 'top-level' `value` option? The only `value` option I can see is the one with the description, "Value for the checkbox input" on line 62.

Also, are we talking about the `value` or `values` option? The changelog update mentions `values`.

36degrees: Not a daft question at all – what you're asking is pretty much the crux of the issue.

Selects and radios use `value` in two different places to mean two different things:

```
{{ govukRadios({
  name: "colours",
  items: [
    {
      value: "red",
      text: "Red"
    },
    {
      value: "green",
      text: "Green"
    },
    {
      value: "blue", ← [1]
      text: "Blue",
      checked: false ← [3]
    }
  ],
  value: "blue" ← [2]
}) }}
```

(Checkboxes is slightly less ambiguous as it uses `values` at the top level – because multiple checkboxes can be checked when the page first loads so it accepts one or more values.)

We use `item.value` (with a `[1]` in the example) to mean 'the value that should be submitted IF the user chooses this option'.

But we also use `value` at the top-level (`[2]`) to mean 'the value of the checkbox that should be checked when the page loads'.

When we're talking about `value` from the context of the `item.checked` option (`[3]`), I think it's ambiguous whether we're talking about `[1]` or `[2]`. Logically only `[2]` would really makes sense, but it's definitely confusing at the minute.

Hence wondering if 'top-level' is a good way to distinguish between the two…

EoinShaughnessy: That's a great explanation - thanks! Ok, another pass at rewording:

```suggestion
    description: Whether the checkbox should be checked when the page loads. If you set this option, the `value` option for the checkbox will not affect the checked state.
```

36degrees:
I think I've addressed the comments from @owenatgov but it could do with another once over. Then I believe this is ready to merge **once v4.1.1 is out**.

owenatgov (approved):
Looking great 👍🏻

PR #2617, opened 2022-05-13
Do not make the service name in the header a link if no `serviceUrl` is provided

### LINKED ISSUE #1826: Prevent creating `<a>` without `href` in header service name

36degrees:
The header macro currently assumes that if there is a serviceName then there will always be a serviceUrl. If no serviceUrl is provided then the service name is wrapped with a link with an empty `href` attribute.

There’s not always a logical place to link the service name in the header to, so it makes sense to allow for service names that don’t link anywhere.

Update the macro so that when serviceUrl is omitted from the options a span is used to wrap the service name rather than a link.

As the current class used on the service name is a modifier for links (`govuk-header__link--service-name`) but will not be used on a link in this context, rename it to `govuk-header__service-name` to make it more generic.

Keep the old class name around as an alias to prevent this being a breaking change, but mark it as deprecated.

Fixes #1826

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,22 @@
 
 ## Unreleased
 
+### Recommended changes
+
+#### Replace deprecated `govuk-header__link--service-name` class in the header
+
+If you're not using the Nunjucks macros, replace any instances in the header of the class `govuk-header__link--service-name` with `govuk-header__service-name`.
+
+We've deprecated the `govuk-header__link--service-name` class, and will remove it in a future major release.
+
+This change was introduced in [pull request #2617: Do not make the service name in the header a link if no `serviceUrl` is provided](https://github.com/alphagov/govuk-frontend/pull/2617).
+
+### Fixes
+
+We’ve made fixes to GOV.UK Frontend in the following pull requests:
+
+- [#2617: Do not make the service name in the header a link if no `serviceUrl` is provided](https://github.com/alphagov/govuk-frontend/pull/2617)
+
 ## 4.1.0 (Feature release)
 
 ### New features
--- a/src/govuk/components/header/_index.scss
+++ b/src/govuk/components/header/_index.scss
@@ -142,6 +142,9 @@
     }
   }
 
+  // The govuk-header__link--service-name class is deprecated - use
+  // govuk-header__service-name instead.
+  .govuk-header__service-name,
   .govuk-header__link--service-name {
     display: inline-block;
     margin-bottom: govuk-spacing(2);
--- a/app/app.test.js
+++ b/app/app.test.js
@@ -165,7 +165,7 @@ describe(`http://localhost:${PORT}`, () => {
       requestPath(templatePath, (err, res) => {
         const $ = cheerio.load(res.body)
         const $header = $('.govuk-header')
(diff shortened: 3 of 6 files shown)


src/govuk/components/header/_index.scss
```diff
@@ -142,6 +142,9 @@
     }
   }
 
+  // The govuk-header__link--service-name class is deprecated - use
+  // govuk-header__service-name instead.
+  .govuk-header__service-name,
```
36degrees, line 147: Are we happy with this approach? If so, will need to create an issue to remove the `govuk-header__link--service-name` class and attach it to the milestone for v5.0.

The alternative I think is to either:

- live with having a link modifier class on something that isn't a link (which feels icky)
- keep both classes around forever

owenatgov (approved):
Looking fab. I'm happy with the approach and am in favour of us removing `govuk-header__link--service-name`. It makes sense to keep our styles clean at least in the long term.

owenatgov:
Afterthought: Does this need a changelog entry?

src/govuk/components/header/_index.scss
36degrees: I've created an issue to remove the deprecated class here: https://github.com/alphagov/govuk-frontend/issues/2636

CHANGELOG.md
```diff
@@ -2,6 +2,22 @@
 
 ## Unreleased
 
+### Recommended changes
+
+#### Replace deprecated `govuk-header__link--service-name` class in the header
+
+If you're not using the Nunjucks macros, replace any instances of the class `govuk-header__link--service-name` with `govuk-header__service-name` in the header.
```
EoinShaughnessy, line 9:
```suggestion
If you're not using the Nunjucks macros, replace any instances in the header of the class `govuk-header__link--service-name` with `govuk-header__service-name`.
```

CHANGELOG.md
```diff
@@ -2,6 +2,22 @@
...
+
+The `govuk-header__link--service-name` class has been deprecated and will be removed in a future major release.
```
EoinShaughnessy, line 11:
```suggestion
We've deprecated the `govuk-header__link--service-name` class, and will remove it in a future major release.
```

EoinShaughnessy (commented):
Content looks good! Made 2 small suggestions, but not feeling strongly about either so feel free to accept/ignore.

CHANGELOG.md
EoinShaughnessy: 🤣

PR #2627, opened 2022-05-16
Use ‘option’ instead of ‘argument’ in macro options documentation

36degrees:
To be consistent with the language we use elsewhere in the documentation and in the Design System, use ‘option’ instead of ‘argument’ in the macro options documentation.


EoinShaughnessy (approved):
This is very neat, thanks for doing it, @36degrees 👍🏻

Issue #2623, opened 2022-05-16
Update browser support documentation


vanitabarrett:
## What
Update any documentation in the GOV.UK Frontend repo that references our browser support, e.g: the README.

We should include:

- what our new level of support is
- that we recommend using v4.x.x if you need to support IE 
- that our level of support for v4.x.x updates is reduced and will be revisited in future
- what we mean by 'functional' and 'compliant' (if we use that terminology)

## Why
We are changing our browser support, so should make sure our documentation is updated. See our [browser proposal](https://github.com/alphagov/govuk-frontend/discussions/2607) for more details.

## Who needs to work on this
Tech writer

## Who needs to review this
Developers

## Done when

- [ ] GOV.UK Frontend README is updated with our new browser support 
- [ ] We have documentation on what to do if your service needs to support Internet Explorer
- [ ] We have documentation on what updates v4 will receive and for how long
- [ ] We have updated any other references to browser support or Internet Explorer as necessary

EoinShaughnessy:
Other docs in repo that mention IE:

- https://github.com/alphagov/govuk-frontend/blob/main/docs/contributing/testing.md
- https://github.com/alphagov/govuk-frontend/blob/main/docs/contributing/resize-text-in-browsers.md
- https://github.com/alphagov/govuk-frontend/blob/main/docs/contributing/coding-standards/css.md
- https://github.com/alphagov/govuk-frontend/blob/main/CONTRIBUTING.md

claireashworth:
Closing this issue as the work related to it is occurring as part of the v5 release. specifically issue 2186

PR #2628, opened 2022-05-17
Update info about browser support in repo docs

EoinShaughnessy:
Fixes [#2623](https://github.com/alphagov/govuk-frontend/issues/2623).

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,8 +39,27 @@ GOV.UK Frontend supports:
 
 - [recommended browsers](https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices#browsers-to-test-in)
 - [recommended assistive technologies](https://www.gov.uk/service-manual/technology/testing-with-assistive-technologies#which-assistive-technologies-to-test-with)
-- Internet Explorer 8, 9 and 10, although components may not look perfect
 - your users overriding colours in Windows, Firefox and Chrome
+
+### Internet Explorer (IE)
+
+We're planning to [reduce our level of support for all versions of IE](https://github.com/alphagov/govuk-frontend/discussions/2607).
+
+In a future major release, we'll reduce the level of support for IE11 from 'compliant' to 'functional'.
+
+We'll also be ending support for IE versions 8, 9 and 10.
+
+For now, you should keep using [GOV.UK Frontend v4.0.0](https://github.com/alphagov/govuk-frontend/releases/tag/v4.0.0) if:
+
+- you need to continue providing a ‘compliant’ experience in IE11
+- your service needs to support IE versions 8, 9 or 10
+
+We’ll continue to provide limited support for GOV.UK Frontend v4.0.0 for the next 12 months by publishing:
+
+- urgent security updates
+- major bug fixes
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -39,9 +39,28 @@ GOV.UK Frontend supports:
...
 - [recommended browsers](https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices#browsers-to-test-in)
 - [recommended assistive technologies](https://www.gov.uk/service-manual/technology/testing-with-assistive-technologies#which-assistive-technologies-to-test-with)
-- Internet Explorer 8, 9 and 10, although components may not look perfect
 - your users overriding colours in Windows, Firefox and Chrome
 
+### Internet Explorer (IE)
+
+We're planning to [reduce our level of support for all versions of IE](https://github.com/alphagov/govuk-frontend/discussions/2607).
+
+In a future major release, we'll reduce the level of support for IE11 from 'compliant' to 'functional'.
```
EoinShaughnessy, line 48: Should we add some detail here about what 'functional' and 'compliant' mean, or would that just be repeating what's in Kim's proposal (which we're linking to)?

EoinShaughnessy:
@vanitabarrett Also saw some other IE mentions in the repo...

https://github.com/alphagov/govuk-frontend/blob/main/docs/contributing/resize-text-in-browsers.md#resize-text-in-internet-explorer-versions-8-to-11
Should we leave this as it is for now, in case anyone wants to know how to resize text in IE? If so, should I create an issue so we know to change/remove it when the time comes (i.e., when we've reduced support)?

https://github.com/alphagov/govuk-frontend/blob/main/docs/contributing/coding-standards/css.md#use-single-colons-for-pseudo-element-selectors
Line: "This is to ensure compatibility with Internet Explorer 8, which doesn't support the double colon syntax."
Can we delete this section?

https://github.com/alphagov/govuk-frontend/blob/main/CONTRIBUTING.md#supported-browsers
Line: Your contribution needs to work with certain browsers as set out in README. See also supporting Internet Explorer 8."
Should we delete the second sentence, if the plan is to deprecate the [Frontend Docs page about supporting IE8](https://github.com/alphagov/govuk-frontend/blob/main/docs/contributing/resize-text-in-browsers.md#resize-text-in-internet-explorer-versions-8-to-11)?
Comment in PR: "Should we leave this as it is for now, in case anyone needs it? If so, should I create an issue so we know to change/remove it when the time comes (i.e., when we've reduced support)?"

https://github.com/alphagov/govuk-frontend/blob/main/docs/contributing/coding-standards/css.md#use-single-colons-for-pseudo-element-selectors
"This is to ensure compatibility with Internet Explorer 8, which doesn't support the double colon syntax."
Comment in PR: "Can we delete this section?"


https://github.com/alphagov/govuk-frontend/blob/main/CONTRIBUTING.md#supported-browsers
"Your contribution needs to work with certain browsers as set out in [README](https://github.com/alphagov/govuk-frontend/blob/main/README.md#browser-and-assistive-technology-support). See also [supporting Internet Explorer 8](https://frontend.design-system.service.gov.uk/supporting-ie8/)."
Comment in PR: Should we delete the second sentence, if the plan is to deprecate the Frontend Docs page about supporting IE8?

Issue #2004, opened 2020-11-04
Decide and document how we release fixes for earlier major versions


hannalaakso:
## What
Split from https://github.com/alphagov/govuk-frontend/issues/1958.

We currently haven't decided or documented how we would issue a patch fix for an earlier major version. 

This is both a product issue (how far back in time does our backwards compatibility go?) and a technical one.

An example scenario would be if we discovered a serious bug in GOV.UK Frontend 3.x.x after we release v4.0.

## Why
It would be good contingency to decide (and document if appropriate) how we should handle a scenario like the one above.

## Further detail
We think we have [documented how we think we would handle this from a technical point of view ](https://docs.google.com/document/d/1avJPHEJ4hENt5LEcMwxp90zl1WS72Tx6qdoN2JOhXCE/edit#heading=h.sxm5dic2qz4c) but a question around handling the `latest` npm tag remains (see below),

## Done when
- [x] We've discussed with the product owner how we want to solve this problem- the browser support work has forced our hand on this a bit as we'll need to maintain v4 and v5+
- [x] If necessary, we've investigated and decided how we avoid moving the `latest` tag on npm 
- [x] If necessary, we've created docs for the process of releasing a fix for an earlier major version

lfdebrux:
> I think the biggest question we've thought of is how we avoid moving the latest tag on npm (e.g: we don't want v4.2.0 to show up as the latest release if v5 is already out)

This is fairly straightforward to do, `npm publish` accepts a `--tag` argument [[1](https://docs.npmjs.com/cli/v8/commands/npm-publish#tag)] that lets you specify a tag other than `latest`.

There does need to be a tag [[2](https://stackoverflow.com/questions/68230173/publishing-a-previous-major-version-to-npm-with-no-tag-and-avoiding-latest)], and there isn't any documented convention on what the tag should be if not latest [[3](https://docs.npmjs.com/cli/v8/commands/npm-dist-tag)], but npm themselves use the pattern `latest-<major version>` [[4](https://www.npmjs.com/package/npm?activeTab=versions)], which makes sense to me.

We'd need to modify the [publish-release script](github.com/alphagov/govuk-frontend/tree/main/bin/publish-release.sh) to allow overriding the tag, but we might be able to add logic to generate the tag for the dev, from the major version.

PR #2635, opened 2022-05-20
Generate NPM tag within publish-release bash script

vanitabarrett:
Part of https://github.com/alphagov/govuk-frontend/issues/2004

## What
Add logic to existing publish-release bash script to work out the correct NPM tag for a release before publishing it.
This also adds the option to provide a tag yourself if the suggested one doesn't suit.

## Why
If we want to release versions of GOV.UK Frontend older than the current major version, e.g: releasing a version 4.5.0. when we've already released version 5.0.0, then we need to be able to override NPM's default "tag" which tags the most recent release as "latest".

Tagging v4.5.0 as "latest" would mean people doing `npm install govuk-frontend` would get v4.5.0 instead of v5.0.0.

Unfortunately, the only way to NOT tag something as "latest" is to provide a different tag to tag it with. This script compares the tags on Github with the version being proposed. If the version number being proposed is smaller than the highest version tag on Github, the NPM tag will be `latest-[MAJOR-VERSION-NUMBER]` e.g: `latest-4`

## Examples

| Current release | New release | Tag      |
|-----------------|-------------|----------|
| 4.1.0           | 4.1.1       | latest   |
| 4.1.0           | 4.2.0       | latest   |
| 4.1.0           | 3.14.0      | latest-3 |
| 4.1.0           | 5.0.0       | latest   |
| 5.0.0           | 4.1.5       | latest-4 |

### DIFF
--- a/bin/generate-npm-tag.sh
+++ b/bin/generate-npm-tag.sh
@@ -1,0 +1,20 @@
+#!/bin/sh
+set -e
+
+# Get highest tag published on Github
+HIGHEST_PUBLISHED_VERSION=$(git tag --list 2>/dev/null | sort -V | tail -n1 2>/dev/null | sed 's/v//g')
+
+# Extract tag version from ./package/package.json
+CURRENT_VERSION=$(node -p "require('./package/package.json').version")
+
+function version() { echo "$@" | awk -F. '{ printf("%d%03d%03d\n", $1,$2,$3); }'; }
+
+if [ $CURRENT_VERSION == $HIGHEST_PUBLISHED_VERSION ]; then
+  echo "⚠️ Git tag $TAG already exists. Check you have updated the version in package/package.json correctly."
+  exit 1
+elif [ $(version $CURRENT_VERSION) -ge $(version $HIGHEST_PUBLISHED_VERSION) ]; then
+  NPM_TAG="latest"
+else
+  major_version_num="${CURRENT_VERSION:0:1}"
+  NPM_TAG="latest-$major_version_num"
+fi
--- a/docs/releasing/publishing-from-a-support-branch.md
+++ b/docs/releasing/publishing-from-a-support-branch.md
@@ -1,11 +1,10 @@
 # Publishing from a support branch
 
-This document is for GOV.UK Design System developers who need to publish a support branch of GOV.UK Frontend when the `main` branch contains unreleasable changes. For example, you might need to release a fix as part of a:
+This document is for GOV.UK Design System developers who need to publish a support branch of GOV.UK Frontend. For example, you might need to release a fix as part of a:
 
 - patch release, after the team has started to merge changes for a new feature release into the `main` branch - for example, a 3.14.x release once we've started merging changes for 3.15.0
-- release, after the team has started to merge changes for a new breaking release into the `main` branch - for example, a 3x.x release once we've started merging changes for 4.0.0
-
-Do not use this document to release changes for previous major releases. We have not tested the document in this scenario, and extra work would be needed to tell npm not to mark the release as the 'latest'.
+- release, after the team has started to merge changes for a new breaking release into the `main` branch - for example, a 3.x.x release once we've started merging changes for 4.0.0
+- release for a previous major version - for example, a 4.x.x release after we've released 5.0.0
 
 If you want to publish the `main` branch for GOV.UK Frontend, [follow the steps in Publishing GOV.UK Frontend](/docs/releasing/publishing.md).
 
@@ -55,7 +54,7 @@ Note: Before you go on annual leave, tell the delivery manager who will be looki
 
 ### Change the code
 
-1. To check out the support branch for the current major release, run `git checkout support/<CURRENT MAJOR VERSION NUMBER>.x`. If the branch does not exist, follow these steps to create it:
+1. Find out which major version this release will be targeting, for example, if you're releasing v4.x.x, the major version is version 4. To check out the support branch for that major version, run `git checkout support/<MAJOR VERSION NUMBER>.x`. If the branch does not exist, follow these steps to create it:
 
     - make sure you have all tags locally by running `git fetch --all --tags --prune`
     - run `git checkout tags/v<LAST RELEASED VERSION NUMBER> -b support/<CURRENT MAJOR VERSION NUMBER>.x` - for example, `git checkout tags/v3.9.1 -b support/3.x`
@@ -64,7 +63,7 @@ Note: Before you go on annual leave, tell the delivery manager who will be looki
 
 3. Push the support branch to GitHub. The branch will automatically have branch protection rules applied.
 
-4. To fix the issue, create a new branch (for example, `git checkout -b fix-the-thing`) from the `support/<CURRENT MAJOR VERSION NUMBER>.x` branch.
+4. Create a new branch for your code changes (for example, `git checkout -b fix-the-thing`) from the `support/<MAJOR VERSION NUMBER>.x` branch.
 
 5. Run `npm install` to make sure you have the latest dependencies installed.
 
@@ -72,13 +71,13 @@ Note: Before you go on annual leave, tell the delivery manager who will be looki
 
 7. Update the changelog with details of the fix.
 
-8. Commit your changes, then push your new branch (see step 4) to GitHub and raise a pull request, with `support/<CURRENT MAJOR VERSION NUMBER>.x` as the base branch to merge into.
+8. Commit your changes, then push your new branch (see step 4) to GitHub and raise a pull request, with `support/<MAJOR VERSION NUMBER>.x` as the base branch to merge into.
 
-9. Once a developer approves the pull request, merge it into `support/<CURRENT MAJOR VERSION NUMBER>.x`. It’s usually a developer who reviews the pull request, but sometimes pull requests need an extra review from another role. For example, if the pull request involves a design change, you may need a review from a designer.
+9. Once a developer approves the pull request, merge it into `support/<MAJOR VERSION NUMBER>.x`. It’s usually a developer who reviews the pull request, but sometimes pull requests need an extra review from another role. For example, if the pull request involves a design change, you may need a review from a designer.
 
 ### Build a new release
 
-1. Check out `support/<CURRENT MAJOR VERSION NUMBER>.x`.
+1. Check out `support/<MAJOR VERSION NUMBER>.x`.
 
 2. Create and check out a new branch, `support-release-[version-number]`. The version number of the new release depends on the type of release. New features correspond to a minor (X.1.X) change - for example, '3.14.0 (Feature release)'. Fixes correspond to a patch (X.X.1) change - for example, '3.13.1 (Patch release)'. In either case, refer to the previous release of that kind, and give the new release the logical next number. To learn more about our versioning, see our [guidance on updating the changelog](/docs/contributing/versioning.md#updating-changelog).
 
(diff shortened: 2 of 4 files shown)


bin/generate-npm-tag.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
+set -e
+
+# Get highest tag published on Github
+HIGHEST_PUBLISHED_TAG=$(git tag --list 2>/dev/null | sort -V | tail -n1 2>/dev/null | sed 's/v//g')
```
lfdebrux, line 5: I think there's a bit of potential of confusion here between Git tags and npm tags 😅 Could we rename this to `HIGHEST_PUBLISHED_VERSION`?

```suggestion
HIGHEST_PUBLISHED_VERSION=$(git tag --list 2>/dev/null | sort -V | tail -n1 2>/dev/null | sed 's/v//g')
```

36degrees:
This is really neat – great idea 🙌🏻 

At the risk of bike-shedding, I think e.g. `v4-latest` might be clearer than `latest-4`… although having said that it looks like [bootstrap](https://www.npmjs.com/package/bootstrap) do something very similar so maybe there's a convention here I'm not aware of…

lfdebrux:
> At the risk of bike-shedding, I think e.g. `v4-latest` might be clearer than `latest-4`… although having said that it looks like [bootstrap](https://www.npmjs.com/package/bootstrap) do something very similar so maybe there's a convention here I'm not aware of…

I don't think npm will let you use `v4-latest`, as they say a tag can't be a valid semantic version, and they say 'the simplest way to avoid semver problems with tags is to use tags that do not begin with a number or the letter v.' [[1](https://docs.npmjs.com/cli/v8/commands/npm-dist-tag)].

bin/generate-npm-tag.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
+set -e
+
+# Get highest tag published on Github
+HIGHEST_PUBLISHED_VERSION=$(git tag --list 2>/dev/null | sort -V | tail -n1 2>/dev/null | sed 's/v//g')
+
+# Extract tag version from ./package/package.json
+CURRENT_TAG=$(node -p "require('./package/package.json').version")
```
lfdebrux, line 8: As above comment about HIGHEST_PUBLISHED_TAG

```suggestion
CURRENT_VERSION=$(node -p "require('./package/package.json').version")
```

vanitabarrett:
> This is really neat – great idea 🙌🏻
> 
> At the risk of bike-shedding, I think e.g. `v4-latest` might be clearer than `latest-4`… although having said that it looks like [bootstrap](https://www.npmjs.com/package/bootstrap) do something very similar so maybe there's a convention here I'm not aware of…

This was my first though, but what @lfdebrux said above - looks like npm advise against doing this. We could do `latest-v4` if that's any better 🤷

36degrees:
Ahah! That all makes sense. Personally I think `latest-v4` makes it a little clearer what the number represents, but not precious about if we'd rather just follow what bootstrap are doing.

docs/releasing/publishing-from-a-support-branch.md
```diff
@@ -55,7 +54,7 @@ Note: Before you go on annual leave, tell the delivery manager who will be looki
 
 ### Change the code
 
-1. To check out the support branch for the current major release, run `git checkout support/<CURRENT MAJOR VERSION NUMBER>.x`. If the branch does not exist, follow these steps to create it:
+1. Find out which major version this release will be targeting, for example: if you're releasing v4.x.x, the major version is version 4. To check out the support branch for that major version, run `git checkout support/<MAJOR VERSION NUMBER>.x`. If the branch does not exist, follow these steps to create it:
```
EoinShaughnessy, line 57:
```suggestion
1. Find out which major version this release will be targeting, for example, if you're releasing v4.x.x, the major version is version 4. To check out the support branch for that major version, run `git checkout support/<MAJOR VERSION NUMBER>.x`. If the branch does not exist, follow these steps to create it:
```

docs/releasing/publishing-from-a-support-branch.md
```diff
@@ -100,17 +99,24 @@ Note: Before you go on annual leave, tell the delivery manager who will be looki
...
 
 2. Sign in to npm (`npm login`), using the npm/govuk-patterns-and-tools team [credentials](https://github.com/alphagov/design-system-team-credentials/tree/main/npm/govuk-patterns-and-tools).
 
-3. Run `npm run publish-release`, which will prompt you to either continue or cancel. Enter `y` to continue.
+3. Run `npm run publish-release`, which will prompt you to check whether the NPM tag looks as expected.
+
+  If you're following these instructions, it's likely that you're publishing an out-of-order release which should
+  *not* be tagged as the 'latest' release on NPM. In this scenario, the tag should take the format `latest-[major-version-number`, for example: `latest-4`.
+
+  Enter 'y' to continue. If you think the tag should be different, enter `N` to have the option to set your own NPM tag.
```
EoinShaughnessy, line 117:
```suggestion
  Enter `y` to continue. If you think the tag should be different, enter `N` to have the option to set your own npm tag.
```

docs/releasing/publishing-from-a-support-branch.md
```diff
@@ -100,17 +99,24 @@ Note: Before you go on annual leave, tell the delivery manager who will be looki
...
 
-1. Check out `support/<CURRENT MAJOR VERSION NUMBER>.x` and pull the latest changes.
+1. Check out `support/<MAJOR VERSION NUMBER>.x` and pull the latest changes.
 
 2. Sign in to npm (`npm login`), using the npm/govuk-patterns-and-tools team [credentials](https://github.com/alphagov/design-system-team-credentials/tree/main/npm/govuk-patterns-and-tools).
 
-3. Run `npm run publish-release`, which will prompt you to either continue or cancel. Enter `y` to continue.
+3. Run `npm run publish-release`, which will prompt you to check whether the NPM tag looks as expected.
+
+  If you're following these instructions, it's likely that you're publishing an out-of-order release which should
```
EoinShaughnessy, line 114:
```suggestion
  If you're following these instructions, you're probably publishing an out-of-order release.
```

docs/releasing/publishing-from-a-support-branch.md
```diff
@@ -100,17 +99,24 @@ Note: Before you go on annual leave, tell the delivery manager who will be looki
...
+  *not* be tagged as the 'latest' release on NPM. In this scenario, the tag should take the format `latest-[major-version-number`, for example: `latest-4`.
```
EoinShaughnessy, line 115:
```suggestion
  Do not tag this release as the 'latest' release on npm. Instead, give the tag the format `latest-[major-version-number`, for example,`latest-4`.
```

- Does there need to be a ']' after 'number'?
- What's an 'out-of-order' release? Will the users of this doc know?

docs/releasing/publishing.md
```diff
@@ -78,17 +78,24 @@ Developers should pair on releases. When remote working, it can be useful to be
 
 2. Sign in to npm (`npm login`), using the npm/govuk-patterns-and-tools team [credentials](https://github.com/alphagov/design-system-team-credentials/tree/main/npm/govuk-patterns-and-tools).
 
-3. Run `npm run publish-release`, which will prompt you to either continue or cancel. Enter `y` to continue.
+3. Run `npm run publish-release`, which will prompt you to check whether the NPM tag looks as expected.
 
-4. View the created tag in the [Github interface](https://github.com/alphagov/govuk-frontend/releases) as follows:
+  If you're following these instructions, you're probably doing a sequential release and therefore
+  the tag should be "latest".
```
EoinShaughnessy, line 84:
```suggestion
  the tag should be 'latest'.
```

docs/releasing/publishing.md
```diff
@@ -78,17 +78,24 @@ Developers should pair on releases. When remote working, it can be useful to be
...
+
+  Enter 'y' to continue. If you think the tag should be different, enter `N` to have the option to set your own NPM tag.
+
+4. You will now be prompted to continue or cancel the release. Check the details and enter `y` to continue.
```
EoinShaughnessy, line 88: What if they check the details and see something they need to sort before continuing?

docs/releasing/publishing-from-a-support-branch.md
vanitabarrett: > Does there need to be a ']' after 'number'?

Thanks, good spot!

> What's an 'out-of-order' release? Will the users of this doc know?

Very good question. I struggled with how to describe this kind of release. I don't think it's clear from 'out-of-order' but wasn't sure how else to describe it. There's a thread on Slack here about it: https://gds.slack.com/archives/CAF8HTJES/p1653049737575209 Happy to hear any thoughts/suggestions!

I'm also going to change the wording because actually I don't think it is likely that the person will be doing one - it's just a lot more likely than if they were following the 'publishing.md' docs.

docs/releasing/publishing.md
vanitabarrett: As above, have added a line about pressing `N` to cancel the release.

docs/releasing/publishing-from-a-support-branch.md
```diff
@@ -100,21 +99,28 @@ Note: Before you go on annual leave, tell the delivery manager who will be looki
...
 
-1. Check out `support/<CURRENT MAJOR VERSION NUMBER>.x` and pull the latest changes.
+1. Check out `support/<MAJOR VERSION NUMBER>.x` and pull the latest changes.
 
 2. Sign in to npm (`npm login`), using the npm/govuk-patterns-and-tools team [credentials](https://github.com/alphagov/design-system-team-credentials/tree/main/npm/govuk-patterns-and-tools).
 
-3. Run `npm run publish-release`, which will prompt you to either continue or cancel. Enter `y` to continue.
+3. Run `npm run publish-release`, which will prompt you to check whether the npm tag looks as expected.
+
+  If you're following these instructions, you may be publishing an out-of-order release. Do
```
EoinShaughnessy, line 114: Just looked at the [Slack convo](https://gds.slack.com/archives/CAF8HTJES/p1653049737575209)...

Can see the logic for each of the suggestions, but I'd probably lean towards 'a release for the current major version', if that's appropriate. I think with each of the others, a new user of this content might wonder "What does 'backport'/'maintenance release'/'support release' mean?"

vanitabarrett: @EoinShaughnessy I've just had a brainwave and realised that actually I think the only time someone may not want to tag as "latest" is when they're doing a release for a previous major version... I've adjusted the text to reflect this.

docs/releasing/publishing-from-a-support-branch.md
```diff
@@ -100,21 +99,27 @@ Note: Before you go on annual leave, tell the delivery manager who will be looki
...
 2. Sign in to npm (`npm login`), using the npm/govuk-patterns-and-tools team [credentials](https://github.com/alphagov/design-system-team-credentials/tree/main/npm/govuk-patterns-and-tools).
 
-3. Run `npm run publish-release`, which will prompt you to either continue or cancel. Enter `y` to continue.
+3. Run `npm run publish-release`, which will prompt you to check whether the npm tag looks as expected.
+
+  If you're publishing a release for a previous major version, do not tag this release as the 'latest' release on npm. Instead, give the tag the format `latest-[major-version-number]`, for example, `latest-4`.
+
+  Enter `y` to continue. If you think the tag should be different, enter `N` to have the option to set your own npm tag.
+
+4. You will now be prompted to continue or cancel the release. Check the details and enter `y` to continue. If something doesn't look right, press `N` to cancel the release.
```
EoinShaughnessy, line 118: Sorry, being nitpicky :p (Style guide says negative contractions give some users trouble)

```suggestion
4. You will now be prompted to continue or cancel the release. Check the details and enter `y` to continue. If something does not look right, press `N` to cancel the release.
```

docs/releasing/publishing.md
```diff
@@ -78,17 +78,24 @@ Developers should pair on releases. When remote working, it can be useful to be
...
-3. Run `npm run publish-release`, which will prompt you to either continue or cancel. Enter `y` to continue.
+3. Run `npm run publish-release`, which will prompt you to check whether the NPM tag looks as expected.
 
-4. View the created tag in the [Github interface](https://github.com/alphagov/govuk-frontend/releases) as follows:
+  If you're following these instructions, you're probably doing a sequential release, meaning
+  the tag should be 'latest'.
+
+  Enter `y` to continue. If you think the tag should be different, enter `N` to have the option to set your own NPM tag.
+
+4. You will now be prompted to continue or cancel the release. Check the details and enter `y` to continue. If something doesn't look right, press `N` to cancel the release.
```
EoinShaughnessy, line 88:
```suggestion
4. You will now be prompted to continue or cancel the release. Check the details and enter `y` to continue. If something does not look right, press `N` to cancel the release.
```

PR #2638, opened 2022-05-20
Use node-version-file input to specify node version in GitHub Actions

lfdebrux:
actions/setup-node v2.5.0 and above support reading a .nvmrc file itself [[1]]. This lets us simplify our code.

[1]: https://github.com/actions/setup-node/releases/tag/v2.5.0


36degrees (approved):
A thing of beauty ❤️

Issue #2608, opened 2022-04-29
Accordion touch targets are too close together for touch users


christopherthomasdesign:
## Related component
Accordion – raised by a user on Slack.

## Context
On accordions, the open section background sits very very close to the first link in the list. This might be a usability issue because there isn't much space between the two touch targets.

<img width="320" alt="Screenshot 2022-04-29 at 12 15 12" src="https://user-images.githubusercontent.com/19834460/165955075-e25f970e-1bce-4b1c-bcd6-eaa0416d3137.png">

## Alternatives
To fix it we could:
* add a margin to the bottom of the section
* add some padding to the top of the first link
* something else?

## Additional information (if applicable)
N/A

PR #2640, opened 2022-05-24
Add top padding to accordion section

owenatgov:
## What/Why
Fixes https://github.com/alphagov/govuk-frontend/issues/2608

Adds top padding to the toggleable accordion section so that the touch area between the accordion heading button and any potential touch targets within that accordion section eg: a link aren't too close together.

### Before
Open:
Screenshot 2022-05-24 at 09 49 22

Open with hover state:
Screenshot 2022-05-24 at 09 49 28

### After
Open:
Screenshot 2022-05-24 at 09 48 48

Open with hover state:
Screenshot 2022-05-24 at 09 48 56


36degrees (approved):
I think this makes sense – I did have a brief chat with @christopherthomasdesign as the section heading does feel quite far away from the content now, but given the other constraints at play there wasn't an obvious solution.

Issue #1892, opened 2020-07-31
Redirect Sass docs in review app to Sass API reference in Frontend Docs


36degrees:
## What

Redirect from http://govuk-frontend-review.herokuapp.com/docs to https://frontend.design-system.service.gov.uk/sass-api-reference/#sass-api-reference

We may need to provide a way to run the 'generic' Sass docs locally so that we can use it to preview documentation before we do a release.

## Why

Avoid the confusion of having two versions of the Sass docs available, and point anyone who may have the existing docs bookmarked to the new docs.

EoinShaughnessy:
Are we linking from the Heroku app to the Sass docs?

Or are we linking to the Sass docs, instead of to the Heroku app? (If so, where are we linking _from_?)

EoinShaughnessy:
@36degrees Our ['Coding standards' doc](https://github.com/alphagov/govuk-frontend/blob/main/docs/contributing/coding-standards/css.md) contains a link to http://govuk-frontend-review.herokuapp.com/docs:

> The syntax is used to generate a [SassDoc application](http://govuk-frontend-review.herokuapp.com/docs/) that documents SCSS in a readable format.

Should the link in above quote go to https://frontend.design-system.service.gov.uk/sass-api-reference/#sass-api-reference instead?

36degrees:
@EoinShaughnessy yes, I think it probably should! Great catch. I don't think we should call it a 'SassDoc application' if we change the link though, and the wording could be clearer… maybe something like:

> The SassDoc comments are used to generate the [Sass API reference in the GOV.UK Frontend docs](https://frontend.design-system.service.gov.uk/sass-api-reference/).

36degrees:
It's also odd that we link 'SassDoc' to the specific SassDoc documentation on file level annotations. That doesn't make sense to me – it should probably just link to the SassDoc homepage?

EoinShaughnessy:
@36degrees Thanks! That wording looks good. I'll raise a PR, separate to this issue, that includes both your wording and a link to the SassDoc homepage.

PR #2642, opened 2022-05-24
Add link from review app to Frontend Sass docs

EoinShaughnessy:
Partly addresses [#1892](https://github.com/alphagov/govuk-frontend/issues/1892).

The front page of the [review app](https://govuk-frontend-review.herokuapp.com/) contains a link ('Sass documentation') to http://govuk-frontend-review.herokuapp.com/docs/.

Seeing the same Sass content in 2 places might be confusing for users.

This PR replaces the current link with one to the [Sass content on our Frontend docs site](https://frontend.design-system.service.gov.uk/sass-api-reference/#sass-api-reference).

### DIFF
--- a/app/views/layouts/index.njk
+++ b/app/views/layouts/index.njk
@@ -40,7 +40,7 @@
       <h2 class="govuk-heading-m">Misc</h2>
 
       <ul class="govuk-list">
-        <li><a href="/docs" class="govuk-link">Sass Documentation</a></li>
+        <li>[Sass Documentation](https://frontend.design-system.service.gov.uk/sass-api-reference/#sass-api-reference)</li>
       </ul>
     </div>
   </div>


querkmachine:
I'm not totally sure of the value of this. Whilst de-duplicating documentation makes sense—especially as both are generated from the same source—the two are potentially distinct enough in function that they make sense being separated.

The SassDoc docs are a little easier to use by virtue of having navigation and search that is specific to Sass code, as well as showing extra information that isn't currently in the Frontend Docs version. Conversely the Docs have usage examples, which SassDoc doesn't. 

Given the change doesn't stop SassDocs being created (it just removes the link to it) there isn't any significant reduction in 'effort' happening. If we wanted to consistently direct people to the Frontend Docs over SassDocs, we should probably ensure the two have feature parity before switching SassDoc off entirely.

vanitabarrett:
@36degrees Are you still planning to take a look at this, or should we put this back into the Backlog?

Issue #2094, opened 2021-01-13
Summary list links are hard to localise


timpaul:
## Related component
Summary panel

## Context
Trying to localise the 'Change' links into another language.

## Additional information (if applicable)
https://github.com/alphagov/govuk-design-system-backlog/issues/36#issuecomment-694167301

36degrees:
I'm going to close this as from reading the linked backlog comment, I think it's the same thing as #2649 which covers this in more detail.

Issue #1904, opened 2020-08-11
Make 'menu' text in header customisable


hannalaakso:
## What
Split out from https://github.com/alphagov/govuk-frontend/issues/1682

The 'menu' text in the header navigation toggle button currently cannot be customised using the header Nunjucks macro. We intended to fix this in https://github.com/alphagov/govuk-frontend/issues/1682, however doing so raises the issue of how should longer text in the menu button behave on smaller viewports (should the text wrap, move the button to a new line, how should the arrow in the button be aligned etc.)

I chatted with @christopherthomasdesign about this who thinks that there's potentially a couple of different scenarios to cover here which would create extra design and dev work. 

I'm thinking that it would be good to see some use cases of services needing to use different text here to understand the use case properly.

## Why
We might still want to do this but it would be good to triage it as a team.

## Who needs to know about this
Chris, devs

christopherthomasdesign:
Just adding a comment to illustrate the issue above and share a very brief bit of exploration into it. 

The current header menu button works fine if the text says something short:

<img width="321" alt="Screenshot 2020-08-12 at 11 51 23" src="https://user-images.githubusercontent.com/19834460/90126773-165aca80-dd5c-11ea-8dba-b47c8c61ef2d.png">

But breaks in its current configuration if you use longer text:

<img width="317" alt="Screenshot 2020-08-12 at 11 59 37" src="https://user-images.githubusercontent.com/19834460/90126939-54f08500-dd5c-11ea-9e9f-5d98b0064aa9.png">

The obvious solution here would be to move the menu button to below the service name:

<img width="318" alt="Screenshot 2020-08-12 at 11 59 00" src="https://user-images.githubusercontent.com/19834460/90127158-a436b580-dd5c-11ea-949b-ad583524816b.png">

There are some drawbacks to this:
- it makes for quite a tall header on small screens (especially for services with long names and when you factor in things like cookie banners)
- as far as I'm aware, CSS doesn't let you position something absolute/block depending on how long it is, meaning that even menus with short text would need to be on a new line (see below)

<img width="317" alt="Screenshot 2020-08-12 at 11 54 32" src="https://user-images.githubusercontent.com/19834460/90127769-a0576300-dd5d-11ea-80da-ff93fb7c134f.png">

[...]

Issue #1681, opened 2019-12-19
Character Count 'count message' cannot be customised or translated


36degrees:
The Character Count contains hardcoded English strings in both the template:

https://github.com/alphagov/govuk-frontend/blob/12f0d6cbe1a5650b8bd61d5ade9e502684f2a167/src/govuk/components/character-count/template.njk#L29

… and the JavaScript:

https://github.com/alphagov/govuk-frontend/blob/12f0d6cbe1a5650b8bd61d5ade9e502684f2a167/src/govuk/components/character-count/character-count.js#L147-L159

This means this text cannot be customised or translated (for example into Welsh) without forking the JavaScript for the component.

matthewmascord:
Have there been any developments on this issue recently? We need to be able to supply a version of this component with the message translated to Welsh and this is not currently possible.

natcarey:
The HMRC approach is to have `language` as a parameter which can be set to `en` or `cy` (for example https://design.tax.service.gov.uk/hmrc-design-patterns/open-links-in-a-new-window-or-tab/)

Being able to set the text directly would be another solution but it de-centralises the default text in Welsh.

If you need Welsh content provided I'm sure HMRC can help.

natcarey:
If we raise a PR to allow this content to be set would that be appropriate?  That feels consistent with the other components.

hannalaakso:
@natcarey We're going to need to consider internalisation as part of a wider piece of work to ensure the approach will work for all Javascript components including more complex ones in the future. This is related to https://github.com/alphagov/govuk-frontend/issues/1708 and https://github.com/alphagov/govuk-frontend/issues/1722. 

So for now we won't be able to accept a contribution on this but thanks so much for offering to work on it, it's much appreciated 🙌 If you wanted to leave any thoughts on either of the above linked issues please feel free to do so.

natcarey:
Thanks, I'll take a look at those other issues.

RickyWelch:
Application Platform Services also have the need for this to be translatable for applications such as the SA-Filing years where a Welsh equivalent of the app is a requirement.

simonneb:
The solution will need to be able to accurately count _characters_ and not _bytes_. 

The GDS Notify team recently updated their internal character counting to more accurately reflect the number of non English characters, so might be worth chatting with @quis ?

36degrees:
@simonneb there's a separate issue for that – #1104. Would you be able to add a comment there instead?

[...]

PR #2648, opened 2022-05-27
[WIP][DO NOT MERGE] Nunjucks internationalisation support

querkmachine:
In addition to updating our JavaScript to allow for localisation (#2614), we also have a few instances of hardcoded strings and fixed word order in our Nunjucks code that need to made configurable. There isn't a 'one size fits all' approach to doing this and we'll probably need to use slightly different methods for enabling localisation of each component. 

## Identified issues

- [x] Character count: Fallback hint text is not editable within Nunjucks. Nunjucks parameters are interpolated into the hint in a fixed content order.
  - #1681
- [ ] Error message: Some visually hidden text (which defaults to ‘Error’) and a colon (:) are hardcoded to appear before the error message, creating a fixed content order.
  - #2650
- [x] Footer: Content licence and Crown copyright notices are hardcoded. 
  - #1290 
  - #1291
- [x] Header: The visible 'menu' link text is hardcoded. Currently, it will overlap the GOV.UK branding if the replacement label is quite long.
  - #1904
- [ ] Summary list: The row key is interpolated alongside action text in a fixed content order.
  - #2094
  - #2649

## Possible improvements

- Date input: Each input's label is localisable, but only by providing the entire `items` array, requiring service users to also re-define the default `name` and `classes` if they don't want these to change. There may be a nicer way of doing this.
- Error summary: The description is always be interpolated between paragraph tags, even if the description is provided via the `html` parameter. This is the only component to do this; other components only add paragraph tags if the content is provided via `text`.

None of the above changes are to enable RTL language support.

### DIFF
--- a/src/govuk/components/header/header.yaml
+++ b/src/govuk/components/header/header.yaml
@@ -51,11 +51,15 @@ params:
 - name: navigationLabel
   type: string
   required: false
-  description: Text for the `aria-label` attribute of the navigation. Defaults to 'Navigation menu'.
+  description: Text for the `aria-label` attribute of the navigation. Defaults to 'Menu'.
 - name: menuButtonLabel
   type: string
   required: false
-  description: Text for the `aria-label` attribute of the button that toggles the navigation. Defaults to 'Show or hide navigation menu'.
+  description: Text for the `aria-label` attribute of the button that toggles the navigation. Defaults to 'Show or hide menu'.
+- name: menuButton
+  type: string
+  required: false
+  description: Text for the button that toggles the navigation. Defaults to 'Menu'.
 - name: containerClasses
   type: string
   required: false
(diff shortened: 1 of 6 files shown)


edwardhorsford:
> Date input: Each input's label is localisable, but only by providing the entire items array, requiring service users to also re-define the default name and classes if they don't want these to change. There may be a nicer way of doing this.

I *suspect* in most cases teams might already have to do this? Don't you need to provide these where you're pre-setting a date? If a team has a check your answers page they'll likely be needing to pre-set.

(I think it would be great for you to be able to pre-set a date without requiring providing the whole array)

src/govuk/components/header/header.yaml
```diff
@@ -51,11 +51,15 @@ params:
 - name: navigationLabel
   type: string
   required: false
-  description: Text for the `aria-label` attribute of the navigation. Defaults to 'Navigation menu'.
+  description: Text for the `aria-label` attribute of the navigation. Defaults to 'Menu'.
```
nickcolley, line 54: Small one, I think the default is "Show or hide menu" instead of "Menu"

querkmachine: You might be mixing `navigationLabel` with `menuButtonLabel` — the latter parameter does default to 'Show or hide menu'.

They're pretty similarly named parameters 😅

edwardhorsford:
I reckon it's probably good enough in the first instance. Agree on documentation as I don't think it's obvious you could do this unless you were rather experienced with the deisgn system - I wonder if it would suit for the design system to have a specific page all about localisation which talks through the various things you might do.

querkmachine:
Consensus within the team also seems to be that the above proposal will do for the time being. We'll revisit it if it proves to be too manual or too inflexible for service teams needs.

querkmachine:
The contents of this branch/PR have now been incorporated into https://github.com/alphagov/govuk-frontend/pull/2614, so I'm going to close this PR.

PR #2651, opened 2022-05-31
Change cookie hide button text examples

vanitabarrett:
Related to https://github.com/alphagov/govuk-design-system/issues/2087

The original hide button text 'Hide this message' was raised in audits as being unclear to users of assistive tech. This isn't hardcoded into the component, but we do have examples which we should update with our [new recommended text](https://github.com/alphagov/govuk-design-system/pull/2207): Hide cookie message.


domoscargin (approved):
👍

Issue #2633, opened 2022-05-19
ES Modules can't be imported without a bundler


colinrotherham:
Hello! 👋 

Loved seeing the ES Modules support land yesterday 😊 

## Description of the issue
I'm unfortunately seeing an [`Error [ERR_PACKAGE_PATH_NOT_EXPORTED]`](https://nodejs.org/api/errors.html#err_package_path_not_exported) thrown _without_ a bundler

## Steps to reproduce the issue
This happens with a plain native Node.js `import` from another `*.mjs` file

**Current** (see [package.json](https://github.com/alphagov/govuk-frontend/blob/137b806d7c308f98f75f8c78ecfdb7f760b27d39/package/package.json#L7))
Missing the `main` or default export

```json
"exports": {
  "import": "./govuk-esm/all.mjs"
}
```

**Fixed** (see [exports documentation](https://nodejs.org/api/packages.html#conditional-exports))
Provided a `main` or default export for both `import` and `require()`

```json
"exports": {
  ".": {
    "import": "./govuk-esm/all.mjs",
    "require": "./govuk/all.js"
  }
}
```

## Actual vs expected behaviour
Node's resolver `node:internal/modules/esm/resolve` doesn't have enough info to import `'govuk-frontend'`

Unlike the [**Import JavaScript using a bundler**](https://frontend.design-system.service.gov.uk/importing-css-assets-and-javascript/#import-javascript-using-a-bundler) guidance regarding webpack, I can't teach Node.js to break the ESM rules.

With the fix above, I sadly hit another challenge. Node.js wouldn't attempt to resolve `.mjs` files as there's no **[package.json](https://github.com/alphagov/govuk-frontend/blob/137b806d7c308f98f75f8c78ecfdb7f760b27d39/package/package.json)** `"type": "module"` and extensions are mandatory for CommonJS and ESM compatibility 😬 

>[**Mandatory file extensions**](https://nodejs.org/api/esm.html#mandatory-file-extensions)
>A file extension must be provided when using the `import` keyword. Directory indexes (e.g. `'./startup/index.js'`) must also be fully specified.
>
>This behavior matches how import behaves in browser environments, assuming a typically configured server.

I wasn't fully out of options though.

[...]

PR #2658, opened 2022-06-10
Add missing mandatory file extensions for ESM JavaScript

vanitabarrett:
Fixes https://github.com/alphagov/govuk-frontend/issues/2633 

> A file extension must be provided when using the import keyword to resolve relative or absolute specifiers. Directory indexes (e.g. './startup/index.js') must also be fully specified.
https://nodejs.org/api/esm.html#mandatory-file-extensions

This PR has been updated to go with Option 2, where our JavaScript files are `.mjs` by default.

--- 

Two approaches here at the moment:

- Option 1: minimal changes to the pre-built files, just adding the missing `.js` file extensions. This means we need to update these file extensions when we copy the files in `package/govuk-esm/` so that they reference `.mjs` rather than `.js`
- Option 2: pre-built files use `.mjs` as they're already in ESM format. These can be copied directly over to `package/govuk-esm/` with no changes. UMD build (using rollup) gets rid of (resolves) import statements anyway and transform to `.js`.


36degrees:
Agreed that option 2 seems like the more 'honest' approach, and also puts us on the right track to eventually move entirely to ESM.

CHANGELOG.md
```diff
@@ -12,6 +12,19 @@ We've deprecated the `govuk-header__link--service-name` class, and will remove i
...
+#### File extensions added for JavaScript ES Module imports
+In 4.1.0, we started providing our component JavaScript in ES module format. This JavaScript used import statements without [mandatory file extensions](https://nodejs.org/api/esm.html#mandatory-file-extensions), as required by Node.js.
+
+We have now updated our component JavaScript to include these missing file extensions. If you previously received an error similar to the below, this fix should resolve that issue:
+
+```
+Cannot find module '../node_modules/govuk-frontend/govuk-esm/common' imported from ../node_modules/govuk-frontend/govuk-esm/all.mjs
+```
+
+If you are already successfully importing our JavaScript as ES modules, you should not need to make any changes, but there may be config you can remove. For example, removing `fullySpecified: false` from your Webpack config file.
```
domoscargin, line 24: My only thought is that this is under "Recommended changes", and we get right up to this line before we suggest an actual change.

Are we able to expand on an entry in "Fixes"? Or is that not really something we do? We could possibly include this action here, and the rest of the info under fixes.

domoscargin:
Small update needed in the Gulp Tasks doc (.js to .mjs)

https://github.com/alphagov/govuk-frontend/blob/da32bb0e9700da1aeb020f83a94d28f3393bfbcb/docs/contributing/tasks.md?plain=1#L62

domoscargin (approved):
I've left a very minor comment, but otherwise this all looks good to me, and still works as expected locally. 🎉

vanitabarrett:
Great spot @domoscargin , thanks! Will change that now

PR #2668, opened 2022-06-22
Fix Summary List action link alignment

### LINKED ISSUE #2281: Multiple actions in Summary list don't align when they go over one line
## Description of the issue
(From support) As seen here, when there are multiple action that go over one line, they don't line up with each other, and the spacing is inconsistent - especially on the right.
https://govuk-frontend-review.herokuapp.com/components/summary-list/extreme/preview

Use multiple actions as in the example above

Actions align/spaced more consistently

- GOV.UK Frontend Version: 3.13.0

querkmachine:
When there are several action links in the Summary List component, the links become misaligned on large screens, due to the links being right-aligned and the separator being appended to the right of each link. Links are aligned correctly on narrow screens, as the links are left-aligned in that context.

Additionally, the spacing on each side of the separator is uneven due to extra whitespace being counted between each link.

This PR intends to fix these issues by:
- Using media queries to swap which side of the link the separator line, margin and padding are applied to, depending on the viewport size.
- Using flexbox on the link container, so that whitespace isn't included in layout calculations. 

Fixes #2281. 

### Small screens (modern browsers, IE11, IE10)

|Before|After|
|:-|:-|
| image | image |

### Large screens (modern browsers, IE11, IE10)

|Before|After|
|:-|:-|
| image | image |

### Large screens (IE9)

Flexbox is not supported by IE9, so the uneven spacing between separators is still present.

|Before|After|
|:-|:-|
| image | image |

### Large screens (IE8)

Neither flexbox, nor the `:first-child` and `:last-child` pseudo-selectors, are supported in IE8.

|Before|After|
|:-|:-|
| image | image |

### DIFF
--- a/src/govuk/components/summary-list/_index.scss
+++ b/src/govuk/components/summary-list/_index.scss
@@ -93,21 +93,43 @@
   }
 
   .govuk-summary-list__actions-list-item {
-    display: inline;
-    margin-right: govuk-spacing(2);
-    padding-right: govuk-spacing(2);
+    display: inline-block;
   }
 
   // In older browsers such as IE8, :last-child is not available,
   // so only show the border divider where it is available.
-  .govuk-summary-list__actions-list-item:not(:last-child) {
-    border-right: 1px solid $govuk-border-colour;
+  @include govuk-media-query($until: tablet) {
+    .govuk-summary-list__actions-list-item {
+      margin-right: govuk-spacing(2);
+      padding-right: govuk-spacing(2);
+    }
+
+    .govuk-summary-list__actions-list-item:not(:last-child) {
+      border-right: 1px solid $govuk-border-colour;
+    }
+
+    .govuk-summary-list__actions-list-item:last-child {
+      margin-right: 0;
+      padding-right: 0;
+      border: 0;
+    }
   }
 
-  .govuk-summary-list__actions-list-item:last-child {
-    margin-right: 0;
-    padding-right: 0;
-    border: 0;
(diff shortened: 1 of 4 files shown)


src/govuk/components/summary-list/_index.scss
```diff
@@ -87,27 +89,54 @@
   }
 
   .govuk-summary-list__actions-list {
+    display: flex;
```
vanitabarrett, line 92: I'm wondering if we can avoid the flex change by removing the errant whitespace in the template instead, using the [minus symbol to strip out whitespace](https://mozilla.github.io/nunjucks/templating.html#whitespace-control) within the list?

E.g:

```
{%- for action in row.actions.items -%}
    <li class="govuk-summary-list__actions-list-item">{{ _actionLink(action) | indent(18) | trim }}</li>
{%- endfor -%}
```

vanitabarrett: Regardless, I think it'd be good for the whitespace fix to be pulled out as a separate commit on this PR as it's not necessary to fix the alignment issue.

vanitabarrett (approved):
Looks good to me! Would be good to get this in a Changelog entry before merging.

vanitabarrett:
@querkmachine I *think* it would just be listed as a regular fix? If we want to tell our users how to remove whitespace if they're not using Nunjucks macros I think we should consider either 1) having the flex fix in combination with the Nunjucks whitespace fix and/or documenting it somewhere a bit more permanent than the release notes

Issue #2170, opened 2021-03-23
Select min-width


Nosfistis:
## Related component

Select

## Context

The select component's width is determined by the max width of its options, as per browser behavior. However, the select box does not specify a min-width, which means that selects with no options, or options with a few characters of text do not appear well.

![εικόνα](https://user-images.githubusercontent.com/3009076/112156158-a6b09600-8bee-11eb-8aaa-1e8d6b6b217f.png)

This is particularly common in JS frameworks that might load the options array lazily.

## Alternatives

We are enforcing a min-width of 5 characters to mitigate for such cases.

PR #2670, opened 2022-06-23
Define mininimum width for select component

querkmachine:
Adds a minimum width of the select component to account for situations where the value list may be loaded asyncronously, as requested in #2170.

As there is no 'standard' minimum width for selects, I opted to use 23 ex—equivalent to about 184 pixels wide on mobile, and 218 pixels wide on tablet and above. 

I used 23 ex for a few reasons:
- It is wide enough that it looks 'as expected' for a select, in the sense that it is noticably wider than it is tall. 
- It's the same width as a text input with the `.govuk-input--width-10` class applied.
- It can sit within `.govuk-width-container` and still comfortably fit on screens 240 pixels wide, which is the narrowest screen resolution with any significant presence in GOV.UK's analytics (the narrowest to be in triple digits, anyway), and one apparently [still used on modern feature phones](https://www.gsmarena.com/nokia_2720_v_flip-10910.php)(!)

### Mobile

|Before|After|
|:-|:-|
| image | image |

### Tablet/desktop

|Before|After|
|:-|:-|
| image | image |

### DIFF
--- a/src/govuk/components/select/_index.scss
+++ b/src/govuk/components/select/_index.scss
@@ -7,6 +7,12 @@
     @include govuk-font($size: 19, $line-height: 1.25);
 
     box-sizing: border-box; // should this be global?
+
+    // This min-width was chosen because:
+    // - it makes the Select noticeably wider than it is taller (which is what users expect)
+    // - 23ex matches the 'length-10' variant of the input component
+    // - it fits comfortably on screens as narrow as 240px wide
+    min-width: 23ex;
     max-width: 100%;
     height: 40px;
     @if $govuk-typography-use-rem {
--- a/src/govuk/components/select/_index.scss
+++ b/src/govuk/components/select/_index.scss
@@ -10,9 +10,9 @@
 
     // This min-width was chosen because:
     // - it makes the Select noticeably wider than it is taller (which is what users expect)
-    // - 23ex matches the 'length-10' variant of the input component
+    // - 11.5em matches the 'length-10' variant of the input component
     // - it fits comfortably on screens as narrow as 240px wide
-    min-width: 23ex;
+    max-width: 11.5em;
     max-width: 100%;
     height: 40px;
     @if $govuk-typography-use-rem {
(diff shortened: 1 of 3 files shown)


src/govuk/components/select/_index.scss
```diff
@@ -7,6 +7,7 @@
     @include govuk-font($size: 19, $line-height: 1.25);
 
     box-sizing: border-box; // should this be global?
+    min-width: 23ex;
```
owenatgov, line 10: I think your reasoning in the PR is sound. Is it worth trying to summarise that in a comment here?

owenatgov (commented):
One nitpick, otherwise looking sharp 👍🏻 This also needs a changelog entry (I think?)

src/govuk/components/select/_index.scss
querkmachine: Added a code comment parroting my comment above.

Also added a changelog entry, currently listed as a fix, though I suppose this could also be a new feature, in a sense??

owenatgov: Great, cheers! I see what you mean about it being a new feature potentially but my instinct is to keep it a fix as it's technically fixing the problem of the select not having a min-width.

owenatgov (approved):
✅

vanitabarrett (changes requested):
Marking this as 'request changes' only because we haven't reached agreement on using `ex` yet.

querkmachine:
We've since merged a change to the Text Input component to use em units instead of ex. I've updated this PR to reflect that change.

vanitabarrett (changes requested):

src/govuk/components/select/_index.scss
```diff
@@ -7,6 +7,12 @@
     @include govuk-font($size: 19, $line-height: 1.25);
 
     box-sizing: border-box; // should this be global?
+
+    // This min-width was chosen because:
+    // - it makes the Select noticeably wider than it is taller (which is what users expect)
+    // - 11.5em matches the 'length-10' variant of the input component
+    // - it fits comfortably on screens as narrow as 240px wide
+    max-width: 11.5em;
```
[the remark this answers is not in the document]
querkmachine: 🤦
Thanks!

vanitabarrett (approved):

PR #1277, opened 2019-04-17
Package updates (Gulp v4 + Jest v24 upgrades)

colinrotherham:
As discussed in https://github.com/alphagov/govuk-frontend/issues/1273 there were various missing `peerDependencies` and 290 npm vulnerabilities found after install.

It did involve a few breaking changes:

**1. Gulp v4**
Lots of minor API changes

**2. Jest v24**
Can't use `__BROWSER__` or any globals so moved to `jest-puppeteer`

You get a nice config file to configure puppeteer `./jest-puppeteer.config.js` to handle starting the application, and means we can remove all the `./lib/puppeteer` boilerplate.

PR #2671, opened 2022-06-23
Remove acorn as a dev dependency

domoscargin:
We added `acorn` as part of [updating Gulp and Jest](https://github.com/alphagov/govuk-frontend/pull/1277) back in 2019.

I've done a brief investigation using that PR's node/npm version and package.json file, and I can't recreate any peer dependency warnings for packages that need `acorn`. It's not used anywhere in our code, and removing it doesn't seem to break anything - it pops up as a sub-dependency here and there with no problem. I think we're mostly putting this down to "legacy npm nonsense".


lfdebrux (approved):
👏 🗑️

Issue #2279, opened 2021-07-26
Tell contributors they need to validate HTML when testing


EoinShaughnessy:
## What

Add step to [our Frontend testing doc](https://github.com/alphagov/govuk-frontend/blob/main/docs/contributing/testing.md) about checking that HTML is valid.

## Why

While working on [accessibility criteria content](https://github.com/alphagov/govuk-design-system/issues/1792), we realised that our testing doc lacked content about checking markup (also, no link to the validator site). That could mean that contributors do not know they will need to check their HTML when they test their contribution.

## Who needs to know about this

@hannalaakso, @EoinShaughnessy  

## Done when

- [x] Update added to doc
- [ ] Update PR reviewed and approved by dev
- [ ] Update 2i'd

PR #2672, opened 2022-06-24
Add static spacing override classes

### LINKED ISSUE #2457: Add static spacing override classes
## Proposal
Add spacing override classes (eg: `govuk-!-margin-bottom-6`) that use the static spacing scale rather than the [responsive spacing scale](https://design-system.service.gov.uk/styles/spacing/).

The desired outcome would be a way to use the classes in such a way that they don't default to the smaller spacing size when moving to mobile. How this might look:

```html
<div class="govuk-!-margin-bottom-6--non-responsive">
...
</div>
```

The above element would have a bottom margin of `30px` across both desktop and mobile instead of moving down to `20px` on mobile.

## Context
On govuk, we've recently been doing some work to set a new design tone that provides a better experience on smaller devices. We'll be able to see this soon in the launch of the new homepage design. Something we've run into however is that there are instances where we want to maintain spacing across screen sizes in order to both create a consistent experience for users and to adhere to good designs across screen sizes.

We are struggling to meet our needs in a positive way presently. For example, we try to use our [govuk heading component](https://components.publishing.service.gov.uk/component-guide/heading) in as many places as possible. The bottom margin for this component is handled by a helper in our govuk components repo to create consistent spacing across our components. This spacing helper uses the margin override classes, as seen above, to minimise CSS weight sent to our users, however that means that we're stuck with the responsive spacing by default when it's not always what we want. The best way we've been able to work out to manage this presently is to create custom classes that use the non-responsive `govuk-spacing` mixin, something we want to avoid doing as we'd be serving extra CSS to our users.

This solution would give us more flexibility in our designs, allowing us to build on top of the design system to create designs that are suitable to the level of information density on govuk whilst still consuming the design system in a positive way.

patrickpatrickpatrick:
New spacing override classes which use the same spacing at all viewports instead of what the current override spacing classes use which is the responsive spacing.

Resolves #2457.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,22 @@ Components that have been updated to support this syntax are:
 
 This was added in [pull request #2734: Update various components to be callable](https://github.com/alphagov/govuk-frontend/pull/2734).
 
+#### Use new override classes to apply static spacing
+
+You can now use static spacing override classes to apply spacing from [the static spacing scale](https://design-system.service.gov.uk/styles/spacing/#static-spacing) to elements of your design.
+
+The new classes start with: `govuk-!-static` followed by either `margin-` or `padding-`, and then a spacing unit number.
+
+To apply spacing in a single direction, include `left-`, `right-`, `top-`, or `bottom-` just before the spacing unit.
+
+For example:
+
+- `govuk-!-static-margin-9` will apply a 60px margin to all sides of the element at all screen sizes
+- `govuk-!-static-padding-right-5` will apply 25px of padding to the right side of the element at all screen sizes
+- `govuk-!-static-margin-0` will remove all margins at all screen sizes
+
+This was added in [pull request #2672: Add static spacing override classes](https://github.com/alphagov/govuk-frontend/pull/2672). Thanks to [Patrick Cartlidge](https://github.com/patrickpatrickpatrick) for this contribution.
+
 ### Deprecated features
 
 #### Remove deprecated `govuk-header__navigation--no-service-name` class in the header
--- a/src/govuk/overrides/_spacing.scss
+++ b/src/govuk/overrides/_spacing.scss
@@ -2,6 +2,8 @@
 /// @group overrides
 ////
 
+// stylelint-disable declaration-no-important
+
 /// Directions for spacing
 ///
 /// @type Map
@@ -14,25 +16,27 @@ $_spacing-directions: (
   "left"
 ) !default;
 
-/// Spacing override classes
+/// Generate responsive spacing override classes
 ///
 /// Generate spacing override classes for the given property (e.g. margin)
-/// for each point in the spacing scale.
+/// for each point in the responsive spacing scale.
 ///
 /// @param {String} $property - Property to add spacing to (e.g. 'margin')
 ///
-/// @example scss
-///   .govuk-\!-margin-0 {
-///      margin: 0;
-///   }
+/// @example css
+///  .govuk-\!-margin-4 {
+///    margin: 15px !important;
+///  }
 ///
-///  .govuk-\!-margin-top-1 {
-///     margin-top: [whatever spacing point 1 is...]
-///   }
+///  @media (min-width: 40.0625em) {
+///    .govuk-\!-margin-4 {
+///      margin: 20px !important;
+///    }
+///  }
 ///
 /// @access private
 
-@mixin _govuk-generate-spacing-overrides($property) {
+@mixin _govuk-generate-responsive-spacing-overrides($property) {
   // For each point in the spacing scale (defined in settings), create an
   // override that affects all directions...
   @each $scale-point, $scale-map in $govuk-spacing-responsive-scale {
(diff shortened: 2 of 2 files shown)


36degrees (commented):
Thanks for picking this up, @patrickpatrickpatrick!

Looking back at the original issue, we think that GOV.UK should probably be writing more CSS for things like headings in components, and relying on override classes less. We'll pick up that conversation with the frontend developers on GOV.UK in the next few weeks.

However, we can still see how the override classes would be useful, and would be happy to see this contribution merged. We have sort of shot ourselves in the foot a little bit with the naming of the existing responsive spacing override classes, but have suggested an approach that we think works.

We'd also prefer to avoid making changes to the helpers, which we think we can do by making a few changes to the way that the overrides have been written. Hopefully the suggestions make sense, but I'm happy to talk through anything if you want clarity. I've also pushed a [branch with a potential alternative approach](https://github.com/alphagov/govuk-frontend/compare/static-spacing-overrides?expand=1) that incorporates the suggestions for your consideration.

> A concern I had about this card is that it adds 76 extra classes that might not necessarily be used by the application that is importing govuk/overrides/all (which is also imported by component_support and used for govuk_publishing_components components). I was wondering if there was a way to import only the override classes that an application needs instead of importing all of the override files in their entirety.

I've taken a look at the file size of the built minified CSS file in dist with and without this change. Thankfully because there's a lot of repetition it compresses reasonably well, and only adds 653 bytes gzipped which I don't think we need to be too concerned about.

**Before:** 110.21kB (13.172kB gzipped)
**After:** 115.85kB (13.825kB gzipped)
**Difference:** 5,640 bytes (653 bytes gzipped)

Generally, users can 'pick and choose' the bits of GOV.UK Frontend that they want to use, but you're correct that there's no way to include only the responsive or non-responsive override classes with this approach.

We can keep an eye on this and, if needed, split the spacing overrides across two files in the future so that they can be imported individually?

36degrees:
As the changes are now based on a branch that I put together, I'm going to request that someone else from the team now reviews this.

querkmachine (approved):
All makes sense to me

36degrees:
This still needs a changelog entry – suggest it'd make sense to do it at the same time as we work on the Design System guidance (https://github.com/alphagov/govuk-design-system/issues/2244) so that we can ensure the terminology we use is consistent.

src/govuk/overrides/_spacing.scss
```diff
@@ -52,7 +56,38 @@ $_spacing-directions: (
...
+///
+/// @example css
+///   .govuk-\!-static-margin-4 {
+///      margin: 20px !important;
+///   }
+///
+/// @access private
+@mixin _govuk-generate-static-spacing-overrides($property) {
+  @each $spacing-point in map-keys($govuk-spacing-points) {
+    .govuk-\!-#{$property}-static-#{$spacing-point} {
```
vanitabarrett, line 74: This doesn't match up with the suggested class name format. For example, it outputs `.govuk-\!-margin-static-1` rather than `.govuk-\!-static-margin-1` (as suggested it should be in the code comment above)

```suggestion
    .govuk-\!-static-#{$property}-#{$spacing-point} {
```

src/govuk/overrides/_spacing.scss
```diff
@@ -52,7 +56,38 @@ $_spacing-directions: (
...
+      #{$property}: govuk-spacing($spacing-point) !important;
+    }
+
+    @each $direction in $_spacing-directions {
+
+      .govuk-\!-#{$property}-#{$direction}-static-#{$spacing-point} {
```
vanitabarrett, line 80: Same as above, I think this should be:

```suggestion
      .govuk-\!-static-#{$property}-#{$direction}-#{$spacing-point} {
```

vanitabarrett (changes requested):

src/govuk/overrides/_spacing.scss
36degrees: Great catch – and apologies to @patrickpatrickpatrick as that was my mistake, introduced in the branch that I pushed.

src/govuk/overrides/_spacing.scss
patrickpatrickpatrick: have made both of the changes requested

src/govuk/overrides/_spacing.scss
```diff
@@ -52,7 +56,38 @@ $_spacing-directions: (
...
 
+/// Generate static spacing override classes
+///
+/// Generate spacing override classes for the given property (e.g. margin)
+/// for each point in the non-responsive spacing scale.
+///
+/// @param {String} $property - Property to add spacing to (e.g. 'margin')
+///
+/// @example css
+///   .govuk-\!-margin-static-4 {
```
vanitabarrett, line 67: Sorry to be nit-picky! I think this should be:

```suggestion
///   .govuk-\!-static-margin-4 {
```

patrickpatrickpatrick: ah whoops, fixed that now

vanitabarrett:
Thanks @patrickpatrickpatrick , this all looks good to me! I'm going to leave my 'request changes' review there for now until the team have figured out what we want to include in the Changelog entry, just so it doesn't get merged too early.

36degrees:
I've added a first attempt at a changelog entry for this.

@patrickpatrickpatrick are you happy with the credit?

I've ended up replicating a lot of the content from the proposed Design System guidance (https://github.com/alphagov/govuk-design-system/pull/2282) – the alternative would probably be to keep the changelog entry really short and just link to the Design System? My thinking however is that without the examples it's hard to understand what it is that we're talking about, so copying the guidance provides useful context.

CHANGELOG.md
```diff
@@ -20,6 +20,22 @@ Visit The National Archives' [documentation on OGL and Crown copyright](https://
 
 This was added in [pull request #2702: Allow localisation of content licence and copyright notices in Footer](https://github.com/alphagov/govuk-frontend/pull/2702).
 
+#### Use new override classes to apply static spacing
+
+You can now use static spacing override classes to apply spacing from [the static spacing scale](https://design-system.service.gov.uk/styles/spacing/#static-spacing) to elements of your design.
```
36degrees, line 25: The linked section doesn't exist yet but should be added by https://github.com/alphagov/govuk-design-system/pull/2279

patrickpatrickpatrick:
> I've added a first attempt at a changelog entry for this.
> 
> @patrickpatrickpatrick are you happy with the credit?
> 
> I've ended up replicating a lot of the content from the proposed Design System guidance ([alphagov/govuk-design-system#2282](https://github.com/alphagov/govuk-design-system/pull/2282)) – the alternative would probably be to keep the changelog entry really short and just link to the Design System? My thinking however is that without the examples it's hard to understand what it is that we're talking about, so copying the guidance provides useful context.

oh yep that looks good to me!

CHANGELOG.md
```diff
@@ -20,6 +20,22 @@ Visit The National Archives' [documentation on OGL and Crown copyright](https://
...
+
+The new classes start with: `govuk-!-static` followed by either `margin-` or `padding-`, and then a spacing unit number.
+
+To apply spacing in a single direction, include `left-`, `right-`, `top-`, or `bottom-` just before the spacing unit.
+
+For example:
+
+- `govuk-!-static-margin-9` will apply a 60px margin to all sides of the element at all screen sizes
+- `govuk-!-static-right-5` will apply 25px of padding to the right side of the element at all screen sizes
```
domoscargin, line 34:
```suggestion
- `govuk-!-static-padding-right-5` will apply 25px of padding to the right side of the element at all screen sizes
```

36degrees: Good catch. I'd spotted this and fixed in #2279 where I'd copied this from but forgot to make the same change here…

Issue #2634, opened 2022-05-19
4.1.0 does not bundle with esbuild


tvararu:
## Description of the issue

`4.1.0` does not bundle successfully with [esbuild](https://esbuild.github.io/).

## Steps to reproduce the issue

Minimal repro repository: https://github.com/tvararu/govuk-frontend-esbuild-test

```js
// index.js
import { initAll } from "govuk-frontend";

initAll();
```

```sh
# build script
"build": "esbuild index.js --bundle"
```

## Actual vs expected behaviour

See https://github.com/tvararu/govuk-frontend-esbuild-test/actions for a successful build with 4.0.1 vs a failing one with 4.1.0.

Example failing build output:

```sh
$ node -v
v18.1.0
$ yarn
yarn install v1.22.19
[1/5] Validating package.json...
[2/5] Resolving packages...
success Already up-to-date.
Done in 0.08s.
$ yarn build
yarn run v1.22.19
$ esbuild index.js --bundle
✘ [ERROR] Could not resolve "./common"
    node_modules/govuk-frontend/govuk-esm/all.mjs:1:32:
      1 │ import { nodeListForEach } from './common'
        ╵                                 ~~~~~~~~~~
✘ [ERROR] Could not resolve "./components/accordion/accordion"
    node_modules/govuk-frontend/govuk-esm/all.mjs:2:22:
      2 │ import Accordion from './components/accordion/accordion'
        ╵                       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
✘ [ERROR] Could not resolve "./components/button/button"
    node_modules/govuk-frontend/govuk-esm/all.mjs:3:19:
      3 │ import Button from './components/button/button'
        ╵                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
✘ [ERROR] Could not resolve "./components/details/details"
    node_modules/govuk-frontend/govuk-esm/all.mjs:4:20:
      4 │ import Details from './components/details/details'
        ╵                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
✘ [ERROR] Could not resolve "./components/character-count/character-count"
    node_modules/govuk-frontend/govuk-esm/all.mjs:5:27:
      5 │ import CharacterCount from './components/character-count/character-count'
        ╵                            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[...]

PR #2674, opened 2022-06-27
Release v4.2.0

domoscargin:
#### Help users navigate through pages with pagination

You can now use [pagination](https://design-system.service.gov.uk/components/pagination/) to help users navigate forwards and backwards through a series of pages. For example, in search results or guidance that's divided into multiple website pages.

This was added in [pull request #2610: Add pagination component](https://github.com/alphagov/govuk-frontend/pull/2610).

#### Check checkboxes by using the `values` Nunjucks option

When using the `govukCheckboxes` Nunjucks macro, you can now use the `values` option to determine which checkboxes should already be checked when the page loads.

For example, `values: ['red', 'blue']` would check any checkboxes that have a `value` of 'red' or 'blue'.

You can use this option instead of setting the boolean `checked` option on each individual checkbox.

This change was introduced in [pull request #2616: Allow selecting options by passing current values](https://github.com/alphagov/govuk-frontend/pull/2616).

#### Check a radio button by using the `value` Nunjucks option

When using the `govukRadios` Nunjucks macro, you can now use the `value` option to determine which radio should already be checked when the page loads.

For example, `value: 'red'` would check the radio that has a `value` of 'red'.

You can use this option instead of setting the boolean `checked` option on each individual radio.

This change was introduced in [pull request #2616: Allow selecting options by passing current values](https://github.com/alphagov/govuk-frontend/pull/2616).

#### Select an option in a select by using the `value` Nunjucks option

When using the `govukSelect` Nunjucks macro, you can now use the `value` option to determine which option should already be selected when the page loads.

For example, `value: 'red'` would select the option that has a `value` of 'red'.

You can use this option instead of setting the boolean `selected` option on each individual option.

This change was introduced in [pull request #2616: Allow selecting options by passing current values](https://github.com/alphagov/govuk-frontend/pull/2616).

#### Replace deprecated `govuk-header__link--service-name` class in the header

If you're not using the Nunjucks macros in the header, replace any instances of the class `govuk-header__link--service-name` with `govuk-header__service-name`.

We've deprecated the `govuk-header__link--service-name` class, and will remove it in a future major release.

This change was introduced in [pull request #2617: Do not make the service name in the header a link if no `serviceUrl` is provided](https://github.com/alphagov/govuk-frontend/pull/2617).

#### File extensions added for JavaScript ES Module imports

We have updated our component ES module JavaScript to include [missing file extensions](https://nodejs.org/api/esm.html#mandatory-file-extensions) not provided in release 4.1.0. If you have received an error similar to the following, for example when running or building your application, this fix should resolve the issue.

```
Cannot find module '../node_modules/govuk-frontend/govuk-esm/common' imported from ../node_modules/govuk-frontend/govuk-esm/all.mjs
```

You should not need to make any changes if you are successfully importing our JavaScript as ES modules with version 4.1.0, but there still might be config you can remove. For example, removing `fullySpecified: false` from your Webpack config file.

This change was introduced in [pull request #2658: Add missing mandatory file extensions for ESM JavaScript](https://github.com/alphagov/govuk-frontend/pull/2658). Thanks to @colinrotherham for reporting issues and suggesting and testing fixes.

### Fixes

We’ve made fixes to GOV.UK Frontend in the following pull requests:

- [#2617: Do not make the service name in the header a link if no `serviceUrl` is provided](https://github.com/alphagov/govuk-frontend/pull/2617)
- [#2640: Add top padding to accordion section](https://github.com/alphagov/govuk-frontend/pull/2640)
- [#2644: Allow users to use `require.resolve` to import GOV.UK Frontend JavaScript](https://github.com/alphagov/govuk-frontend/pull/2644) - thanks to @HughePaul for reporting this issue and testing the fix
- [#2647: Allow users to import sass files via Webpack `sass-loader`](https://github.com/alphagov/govuk-frontend/pull/2647) - thanks to @Garethp for reporting this issue, and to @Garethp and @richpjames for testing the fix
- [#2659: Add missing label and legend classes to HTML fixtures](https://github.com/alphagov/govuk-frontend/pull/2659)


36degrees (approved):
All looks good to me 👍🏻

colinrotherham:
@36degrees If you want another thank you, @tvararu spotted the ES Modules issue too via https://github.com/alphagov/govuk-frontend/issues/2634

For anyone else searching for [esbuild](https://esbuild.github.io/) answers it's likely the `--resolve-extensions=.mjs,.js` fix isn't needed anymore either once **v4.2.0** lands 🎉

PR #2675, opened 2022-06-27
Update changelog to add thank you before release

domoscargin:
It's been pointed out that we missed a thank you in our release notes, so just adding that here.

https://github.com/alphagov/govuk-frontend/pull/2674#issuecomment-1167207038


36degrees (approved):
👍🏻

Issue #2657, opened 2022-06-06
Error summary list not read out in JAWS (2022)


tvararu:
Discussion in cross-gov Slack: https://ukgovernmentdigital.slack.com/archives/C6DMEH5R6/p1654079614172949

Similar issue: https://github.com/alphagov/govuk-frontend/issues/2072 (no longer applicable in my own testing)

## Description of the issue

List of error messages within the error summary is not announced when using JAWS 2022 with Edge or Chrome.

NVDA, VoiceOver on iPad were also tested and they correctly recited the content.

## Steps to reproduce the issue

- Open this link http://govuk-frontend-review.herokuapp.com/components/error-summary/preview
- Observe what JAWS reads out

Here is an example video on JAWS on another service (https://find-a-lost-trn.education.gov.uk/). The h2 is read out at 00:36, the messages are not. Many thanks to Adam Liptrot on the cross-government Slack who recorded it.

https://user-images.githubusercontent.com/1650875/172218411-a408b8d9-8e03-4810-b498-dab166879537.mov

## Actual vs expected behaviour

JAWS is only reading out the heading. It should read out the entire summary.

## Environment (where applicable)

- Operating system: Windows 11
- Browser: Chrome and Edge
- Browser version: Chrome 102 and Edge 101
- GOV.UK Frontend Version: 4.1.0

querkmachine:
Hi @tvararu, thanks for the bug report. 

Are you able to specify which versions of Chrome and Edge you've reproduced this on? This is useful in case we end up coming back to this later, and there can be differences between browser versions over time. 

Thanks!

tvararu:
@querkmachine updated! :+1:

tvararu:
# Investigation and analysis

@fofr, @adamliptrot-oc, and I have been looking into this for the past couple of days. Here's what we've found so far.

## Investigation

The bug can be observed by opening or refreshing [a page containing the error summary](http://govuk-frontend-review.herokuapp.com/components/error-summary/preview), while running JAWS 2022.2204.20 (Trial) with Edge 102 on Windows 10:

https://user-images.githubusercontent.com/1650875/172676846-32a4c6b3-6f98-4b5b-8af8-76e0de5596de.mov

In the first video, JAWS is reading out the following (comments added):

```
Refreshing page, Loading page
GOV.UK Frontend                             // Page title
Heading level 2 There is a problem          // The first element inside the focussed error summary
Page has 1 region, 1 heading, and 2 links
```

[...]

Issue #2055, opened 2020-12-03
Error Summary is not focused correctly on second submission with NVDA


domoscargin:
## Description of the issue

While using NVDA 2020.3 on Firefox 81, if an invalid form is submitted twice, focus is not set on the Error Summary.

This means that when a user tabs to the next element, it is _not_ the first error message link, but a later link on the page.

## Steps to reproduce the issue

* Using NVDA 2020.3 on Firefox 81 (I ran it through assistivlabs)
* Go to this full page example: https://github.com/alphagov/govuk-frontend/tree/master/app/views/full-page-examples/update-your-account-details
* Enter nothing, submit
* The page reloads with an Error Summary which is correctly focused
* Again, enter nothing and submit again
* The page reloads with focus set elsewhere. Pressing tab once moves focus to a footer link

## Actual vs expected behaviour

Focus does not appear to be set on the Error Summary - I would expect it to have focus on each form submission beyond the first.

## Environment (where applicable)

- Operating system: assistivLabs NVDA 2020.3 (presumably on Windows 10)
- Browser: Firefox
- Browser version: 81
- GOV.UK Frontend Version: 3.10

36degrees:
Re-tested and still happening with NVDA 2021.3.1 and Firefox 97.

Don't think there's anything we can do here, but we might want to report this as a bug to NVDA if we can work out a minimal test case.

domoscargin:
This has now been reported to NVDA: https://github.com/nvaccess/nvda/issues/13919

domoscargin:
NVDA will not process without a Codepen or similar. Given that this is a bit of a flaky issue (see https://github.com/alphagov/govuk-frontend/pull/2677) and might need page reload behaviour to test properly, I'm not sure how to get it done.

vanitabarrett:
Might make sense to wait until we find out the outcome of https://github.com/alphagov/govuk-frontend/pull/2677, but we could timebox half a day or so to look into creating a minimal example

PR #2677, opened 2022-06-28
Amend error summary markup to fix page load focus bug in JAWS 2022

owenatgov:
Fixes https://github.com/alphagov/govuk-frontend/issues/2657

## What/Why
Makes the following changes to the [error summary component](https://design-system.service.gov.uk/components/error-summary/):

- Moves `role="alert"` from the parent error summary element to a child container as it was causing a race condition 
- Removes the `aria-labelledby` and the associated `id` from the error summary title to reduce unnecessary verbosity

## Testing
As well as the latest version of JAWS 2022, this was tested against screen reader/combinations as outlined in [the service manual's assistive technology testing guidance](https://www.gov.uk/service-manual/technology/testing-with-assistive-technologies#which-assistive-technologies-to-test-with). I applied the following criteria when testing:

- JAWS on Chrome and Edge should, at a minimum, read out the entire contents of the error summary on load
- All other screen readers don't have to be perfect but any changes should not create regressions in the experience

[Testing results](https://docs.google.com/spreadsheets/d/15gxrw2zRhHdkdYq8YZ9xeL1C0H7FPAIJV8w_DGzOS20/edit?usp=sharing)

 ## Credits
 Thank you to @tvararu, @fofr and their team at DfE for their continued support on solving this issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,20 @@
 
 ## Unreleased
 
+### Recommended changes
+
+#### Remove `aria-labelledby`, remove `id="error-summary-title"` from title and move `role="alert"` to child container on the error summary component
+
+If you're not using the Nunjucks macros, you can improve the experience for screen reader users by making these changes to the error summary markup:
+
+- Remove `aria-labelledby="error-summary-title"` and `role="alert"` from the parent element (`govuk-error-summary`)
+- Add a `div` wrapper around the contents of `govuk-error-summary` with the attribute `role="alert"`
+- Remove `id="error-summary-title"` from the error summary `h2` (`govuk-error-summary__title`)
+
+This will enable screen reader users to have a better, more coherent experience with the error summary. Most notably it will ensure that users of JAWS 2022 or later will hear the entire contents of the error summary on page load and therefore have further context on why there is an error on the page they're on.
+
+This was added in [pull request #2677: Amend error summary markup to fix page load focus bug in JAWS 2022](https://github.com/alphagov/govuk-frontend/pull/2677)
+
 ## 4.3.1 (Patch release)
 
 ### Recommended changes
--- a/src/govuk/components/error-summary/template.njk
+++ b/src/govuk/components/error-summary/template.njk
@@ -1,26 +1,30 @@
 <div class="govuk-error-summary
-  {%- if params.classes %} {{ params.classes }}{% endif %}" aria-labelledby="error-summary-title" role="alert"
+  {%- if params.classes %} {{ params.classes }}{% endif %}" aria-labelledby="error-summary-title"
   {%- if params.disableAutoFocus %} data-disable-auto-focus="true"{% endif %}
   {%- for attribute, value in params.attributes %} {{ attribute }}="{{ value }}"{% endfor %} data-module="govuk-error-summary">
-  <h2 class="govuk-error-summary__title" id="error-summary-title">
-    {{ params.titleHtml | safe if params.titleHtml else params.titleText }}
-  </h2>
-  <div class="govuk-error-summary__body">
-    {% if caller or params.descriptionHtml or params.descriptionText %}
-    <p>
-      {{ caller() if caller else (params.descriptionHtml | safe if params.descriptionHtml else params.descriptionText) }}
-    </p>
-    {% endif %}
-    <ul class="govuk-list govuk-error-summary__list">
-    {% for item in params.errorList %}
-      <li>
-      {% if item.href %}
-        <a href="{{ item.href }}"{% for attribute, value in item.attributes %} {{attribute}}="{{value}}"{% endfor %}>{{ item.html | safe if item.html else item.text }}</a>
-      {% else %}
-        {{ item.html | safe if item.html else item.text }}
+  {# Keep the role="alert" in a seperate child container to prevent a race condition between
+  the focusing js at the alert, resulting in information getting missed in screen reader announcements #}
+  <div role="alert">
+    <h2 class="govuk-error-summary__title" id="error-summary-title">
+      {{ params.titleHtml | safe if params.titleHtml else params.titleText }}
+    </h2>
+    <div class="govuk-error-summary__body">
+      {% if caller or params.descriptionHtml or params.descriptionText %}
+        <p>
+          {{ caller() if caller else (params.descriptionHtml | safe if params.descriptionHtml else params.descriptionText) }}
+        </p>
       {% endif %}
-      </li>
-    {% endfor %}
-    </ul>
+      <ul class="govuk-list govuk-error-summary__list">
+        {% for item in params.errorList %}
+          <li>
+          {% if item.href %}
+            <a href="{{ item.href }}"{% for attribute, value in item.attributes %} {{attribute}}="{{value}}"{% endfor %}>{{ item.html | safe if item.html else item.text }}</a>
+          {% else %}
+            {{ item.html | safe if item.html else item.text }}
+          {% endif %}
+          </li>
+        {% endfor %}
+      </ul>
+    </div>
   </div>
 </div>
--- a/src/govuk/components/error-summary/template.njk
+++ b/src/govuk/components/error-summary/template.njk
@@ -1,11 +1,11 @@
 <div class="govuk-error-summary
-  {%- if params.classes %} {{ params.classes }}{% endif %}" aria-labelledby="error-summary-title"
+  {%- if params.classes %} {{ params.classes }}{% endif %}"
   {%- if params.disableAutoFocus %} data-disable-auto-focus="true"{% endif %}
   {%- for attribute, value in params.attributes %} {{ attribute }}="{{ value }}"{% endfor %} data-module="govuk-error-summary">
   {# Keep the role="alert" in a seperate child container to prevent a race condition between
   the focusing js at the alert, resulting in information getting missed in screen reader announcements #}
   <div role="alert">
-    <h2 class="govuk-error-summary__title" id="error-summary-title">
+    <h2 class="govuk-error-summary__title">
       {{ params.titleHtml | safe if params.titleHtml else params.titleText }}
     </h2>
     <div class="govuk-error-summary__body">
--- a/src/govuk/components/error-summary/template.test.js
+++ b/src/govuk/components/error-summary/template.test.js
@@ -25,11 +25,11 @@ describe('Error-summary', () => {
       expect(ariaAttr).toEqual('error-summary-title')
     })
 
-    it('has role=alert attribute', () => {
+    it('has a child container with the role=alert attribute', () => {
       const $ = render('error-summary', examples.default)
-      const roleAttr = $('.govuk-error-summary').attr('role')
+      const childRoleAttr = $('.govuk-error-summary div:first-child').attr('role')
 
-      expect(roleAttr).toEqual('alert')
+      expect(childRoleAttr).toEqual('alert')
     })
 
     it('renders title text', () => {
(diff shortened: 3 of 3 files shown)


36degrees (commented):
Thanks for all your hard work investigating this @owenatgov – glad we've got a solution that works for JAWS!

Given this change is likely to affect other assistive technologies as well, it'd be really useful if this PR included:

- how you have been testing the change
- the exact behaviour in JAWS before / after the change (mostly helps the reviewer, but also useful if a similar issue comes up in the future and we want to compare the behaviour)
- what happens in each of the main assistive tech / browser combinations that we support, ideally compared to the current behaviour (so we can be confident that we're not making things worse in e.g. VoiceOver on iOS)

It'd also be great if we can capture a bit more detail in the commit messages – in particular [_why_ the change is being made](https://gds-way.cloudapps.digital/standards/source-code/working-with-git.html#commit-messages) in 40d777e260a5d9416b02540e4fe1bc93e9e1084d.

It might be worth taking a look at PRs for some of the other recent screenreader related changes to see what we've done in the past. (I know this isn't something we're particularly consistent at yet, and ideally this should be better documented!)

Very happy to talk things through if you want clarity on anything!

owenatgov:
@36degrees I've updated the PR description with further testing details and updated the commit history with a more logical flow of work and more details per change. I'll be doing some final testing of my solution on other screen readers to ensure it's all working correctly soon. Let me know if there's more I can do to tidy this up.

tvararu (approved):
Code looks good to me, the commits are logical and well explained. There's a small typo in the first commit ("remvoed").

Do we have a feel for which milestone this could go into? Quite excited to use this in our services :)

owenatgov:
# Results of final pass browser testing

Testing done using this branch and the [passport details page in the review app](https://govuk-frontend-review.herokuapp.com/full-page-examples/passport-details).

| Screen reader/browser combination | Result |
| - | - |
| JAWS 2022 (version available in GDS AssistivLabs account), Chrome and Edge | Upon submitting a blank form, JAWS will read out the error summary heading ("there is a problem") along with the error summary links without the need for me to navigate down. Upon page refresh, only the heading is read out and I need to navigate to reach the links. |
| NVDA, Chrome and Firefox | Error summary heading and links are consistently read out on empty form submission and page refresh |
| VoiceOver, Safari | Initially didn't read out the links on empty form submission as well as the error heading, however did eventually reach them when a full page read out was triggered. Same behaviour for page refresh. |

Personal analysis of this testing is that there are some imperfections, notably in VoiceOver, however we are making a marked improvements so recommend we proceed with solution. I'm keen to compare testing results from a reviewer.

querkmachine:
Trying the same browser/AT combinations as above, on the same page in this branch's review app.

NVDA returned the same results as @owenatgov's testing, as did VoiceOver (with caveats), however my experience with JAWS was markedly different and incredibly inconsistent when making multiple runs. 

## JAWS 2022 / Chrome 103

I received some wildly different results each time I tried JAWS. 

When using a mouse, the result seemed to be contingent on how quickly I navigated. If I submitted the form before JAWS had completed announcing 'Continue button, to activate press enter' then the error summary on the next page wouldn't be announced. If I let the announcement finish first, then it would read 'heading level 2, There is a problem', but not the list of errors. 

When using a keyboard, 'heading level 2, There is a problem' was read out consistently, regardless of whether the previous announcement had finished or not. In one instance which I couldn't consistently reproduce, it read out 'There is a problem' and the text of the error list, without announcing the elements. 

On page refresh, the error summary was never read out.

In all situations keyboard focus was moved to the error summary, so navigating downwards would announce the heading and list of errors.

## JAWS 2022 / Edge 103

I couldn't get Edge to read out the error summary on page load by any method. 

Like Chrome, the keyboard focus was moved to the error summary so it was possible to read the error summary by manually navigating downwards. 

## NVDA 2022.1 / Chrome 103, Firefox 102

The 'There is a problem' and list of errors was read out consistently on page load or reload. 

The error summary consistently received keyboard focus. 

## VoiceOver / Safari 15.5

When navigating using VoiceOver controls (VO key + arrows), the error summary was not announced on page load or reload, and was only read in the context of the entire page. The **heading** of the error summary received focus, and that the user was currently on a 'heading level 2' was announced, but the text of the heading was not.[^1] I was able to announce the heading and navigate to the list of errors using VO controls.

[^1]: VoiceOver's on-screen output did display the text briefly, but it appears to have been overridden by the 'heading level 2' announcement before it could be read.

When using the tab key to navigate, the 'There is a problem' heading's text _was_ announced, and the error summary container received focus. However, it was not possible to navigate into the list of errors using the tab key.

domoscargin:
I was hoping this would solve this problem with NVDA: #2055 but it looks like it's still an issue. Just popping it here for completeness.

owenatgov:
I've been playing around with this again in response to @querkmachine and @domoscargin's testing and have some odd findings:

Firstly to verify the above review testing, JAWs is indeed very inconsistent. There doesn't seem to be any rhyme or reason to when it reads out the full error summary and when it just reads out the heading or just nothing at all ("blank"). I have a hypothesis that either there is a definite bug in JAWs that we've been boxing with this whole time or there's something else in the component that's causing problems.

Secondly, I've been playing around with the component js a bit more and I tried setting the programatic `tabindex` to 0 instead of -1 again. I don't know why but this seems to make all screen readers much more consistent in reading out the error summary in full. It doesn't really make sense that this would work over -1 as the whole point of `tabindex="-1"` is to stop something from being focusable via tabbing but allow it to be programatically focusable via js, which is what we're doing here. I'm gonna do some more thorough testing and try to find out what's going on here.

owenatgov:
Have updated the solution to use 0 on the tabindex instead of -1. Testing again using the [passport details page](https://govuk-frontend-pr-2677.herokuapp.com/full-page-examples/passport-details).

# Results
## JAWS 2022 / Chrome 103
JAWS continues to be flakey. It read out the entire contents of the error summary more often than in didn't and when it didn't, screen reader focus consistently remained in the correct place, allowing users to reach the links next.

## JAWS 2022 / Edge 103
Seems to be more flakey than Chrome. Read out the entire error summary about 25% of the time. Would never have focus in the correct place, often it was either on the continue button or one of the questions.

## NVDA 2022.1 / Firefox 102
Reads out the entire error summary on load consistently. It also maintained screen reader focus on the error summary on refresh every time. @domoscargin It may be worth you retesting the issue you linked to make sure that I haven't had an exceptionally lucky session.

## VoiceOver / Safari 15.5
Unexpectedly chaotic. Wouldn't move focus to the error summary on page reload when using VO navigation. The user would eventually receive the error summary content via the entire page getting announced, which seemed to be triggered 75% of the time in testing.

# Thoughts
There are obviously still problems however from extensive testing I can't find anything that could be causing this on our end. I'll leave this to those reviewing and testing this to decide if this is worth merging regardless. I feel like it's still a positive addition as it cleans up the markup and provides a broadly better experience than we had previously. I'm planning on updating the original issue and finally raising issues to screen reader vendors about this shortly.

domoscargin:
Thanks for the continued picking apart of this stuff! I can still get flaky NVDA behaviour on the [deployment](https://govuk-frontend-pr-2677.herokuapp.com/full-page-examples/passport-details), but again, I wouldn't let that stop this PR in any way.

If, as part of raising issues to screen reader vendors you come up with a good minimal test case for these related issues, please do let me know and I can update [our issue with NVDA](https://github.com/nvaccess/nvda/issues/13919) so they can take a look.

owenatgov:
Following a bit of a return to centre after updating the version of JAWS in the GDS AssistivLabs account, I've updated this PR along with the description with an amended solution and some more methodical testing results as per our [accessibility testing guidance](https://github.com/alphagov/govuk-frontend/blob/main/docs/contributing/test-components-using-accessibility-acceptance-criteria.md#how-to-test-components-using-accessibility-acceptance-criteria).

vanitabarrett:
This looks good to me 👍 The testing results show that the experience has been improved compared to what's live now, and I tested it out in JAWS and NVDA and it seemed to read it out fairly reliably.

This will need a changelog entry - I imagine we'll want to file this under a 'recommended change' as anyone using the HTML components will need to update their markup.

owenatgov:
@vanitabarrett Changelog entry added. Let me know what you think, although it in theory is gonna get reviewed by content folks before the next release.

CHANGELOG.md
```diff
@@ -2,6 +2,20 @@
 
 ## Unreleased
 
+### Recommended changes
+
+#### Change accessible attributes in the markup of the error summary component
```
vanitabarrett, line 7: Looking at [other times we've written up recommended changes](https://github.com/alphagov/govuk-frontend/blob/main/CHANGELOG.md#remove-aria-live-from-the-character-count-component), I think we try to be as explicit as possible about what needs to be changed.

owenatgov: I've updated this to try to be more descriptive, hopefully it isn't too overwhelming now. Let me know what you think.

vanitabarrett: It's definitely clearer now! Not sure if it's too long, but hopefully Calvin or Claire might be able to help condense it down a little too (or reassure us that it's ok)

CHANGELOG.md
```diff
@@ -2,6 +2,20 @@
 
 ## Unreleased
 
+### Recommended changes
+
+#### Remove `aria-labelledby`, remove `id="error-summary-title"` from title and move `role="alert"` to child container on the error summary component
+
+If you're not using the Nunjucks macros, we recommend making the following changes to the error summary markup:
```
calvin-lau-sig7, line 9: Yeah, this looks alright to me. I'd suggest giving the reason for the recommendation earlier on.

Also, is "the Nunjucks macros" clear enough? I can't tell from this if it's 'all Nunjucks macros in GOV.UK Frontend' or just the ones in the error summary component.

```suggestion
If you're not using the Nunjucks macros, you can help screen reader users by making these changes to the error summary markup:
```

owenatgov: Will make the change in a mo. I took the nunjucks macros line as a convention in our changelog. Generally if a team aren't using one of our macros then it's very likely they aren't using our macros at all as typically the reason to not use our macros is a question of codebase compatibility eg: govuk using ruby on rails which doesn't fit with nunjucks. It's a bit tricky to convey though so I'm open to suggestions.

CHANGELOG.md
```diff
@@ -2,6 +2,20 @@
...
+
+#### Remove `aria-labelledby`, remove `id="error-summary-title"` from title and move `role="alert"` to child container on the error summary component
+
+If you're not using the Nunjucks macros, you can improve the experience for screen reader users by making these changes to the error summary markup:
+
+- Remove `aria-labelledby="error-summary-title"` and `role="alert"` from the parent element (`govuk-error-summary`)
+- Add a `div` wrapper around the contents of `govuk-error-summary` with the attribute `role="alert"`
+- Remove `id="error-summary-title"` from the error summary `h2` (`govuk-error-summary__title`)
+
+This will enable screen reader users to have a better, more coherent experience with the error summary. Most notably it will ensure that users of JAWS 2022 and up will hear the entire contents of the error summary on page load and therefore have further context on why there is an error on the page they're on.
```
calvin-lau-sig7, line 15: Looks good here, thanks! I don't feel too concerned about length, seems pretty typical for the changelog?

Did a quick ctrl-F and looks like we say "or later" when referring to software versions.

```suggestion
This will enable screen reader users to have a better, more coherent experience with the error summary. Most notably it will ensure that users of JAWS 2022 and later will hear the entire contents of the error summary on page load and therefore have further context on why there is an error on the page they're on.
```

calvin-lau-sig7 (approved):

CHANGELOG.md
owenatgov: Thanks Calv! I've made this change locally and I'm just waiting for the 4.3.1 release before I push and merge

tvararu:
🙌 Amazing work all!!

PR #2679, opened 2022-07-01
Add docs on managing change

36degrees:
Split out the documentation on deprecations from the versioning documentation, as the new documentation supercedes it.

### DIFF
--- a/docs/contributing/managing-change.md
+++ b/docs/contributing/managing-change.md
@@ -1,0 +1,173 @@
+# Managing change
+
+Follow the processes in this document to make it easier for users to migrate to new major versions.
+
+## Deprecating and removing features
+
+Wherever possible, deprecate features as part of a minor release, before removing them in the next major release.
+
+This gives users the option of migrating away from using the deprecated features at their own pace, minimising the amount of work that needs to be done when new major versions are released.
+
+Features should not be deprecated while they are relied on by other parts of GOV.UK Frontend.
+
+Deprecations should not be made in patch releases.
+
+### Tell users about the deprecation
+
+Tell users about the deprecation in the release notes.
+
+Do not assume that users will understand the term 'deprecation'. Wherever possible, explain why we are deprecating the feature and what we want users to do if they are currently using it.
+
+Depending on what is being deprecated, we might want to throw deprecation warnings from the code to help users find where they are using the deprecated feature in their own codebase. See the guidance on deprecating specific types of features below.
+
+You can mention the warnings in the release notes to help users understand if they are affected.
+
+For example:
+
+> If you import specific files from the core or overrides layers, you’ll now see a deprecation warning when compiling Sass if you do not import `node_modules/govuk-frontend/govuk/base` first.
+> 
+> To fix the warning, import `node_modules/govuk-frontend/govuk/base` first. For example:
+> 
+> ```scss
+> @import "node_modules/govuk-frontend/govuk/base";
+> @import "node_modules/govuk-frontend/core/typography";
+> ```
+> 
+> If you do not import `node_modules/govuk-frontend/govuk/base` first, your service will no longer work from GOV.UK Frontend v4.0.0.
+
+### Make sure we remember to remove the deprecated feature
+
+Create a GitHub issue for removing the deprecated feature. If there are multiple related things that should be removed at the same time then you can create a single issue that covers all of them.
+
+The issue should include:
+
+- clear instructions for removal
+- the reason for the deprecation and removal
+
+Attach the GitHub issue to the milestone for the next major version.
+
+### Deprecating a Sass function or mixin
+
+Include the [`@deprecated` annotation](http://sassdoc.com/annotations/#deprecated) in the Sassdoc comment block for the function, variable or mixin.
+
+In the annotation description include:
+
+- the suggested alternative, if there is one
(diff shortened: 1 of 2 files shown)


docs/contributing/managing-change.md
```diff
@@ -0,0 +1,171 @@
+# Managing change
+
+When making changes to the codebase that could affect our users, we want to minimise the number of breaking changes in a major version that cannot be made before updating to the major version.
```
vanitabarrett, line 3: Slightly unsure what we mean by the second half of this sentence, "...in a major version that cannot be made before updating to the major version". Are we talking about minimising the number of breaking changes that a user might be unable to make in their own service/port?

docs/contributing/managing-change.md
```diff
@@ -0,0 +1,171 @@
...
+///
+/// @deprecated Use govuk-multiply(number, 2) instead.
+///   See https://github.com/alphagov/govuk-frontend/issues/1234
+@function govuk-double($number) {
+  @warn "govuk-double($number) is deprecated. Use govuk-multiply($number, 2) instead.";
+  @return govuk-multiply($number, 2);
+}
+```
+
+### Deprecating an parameter for a Sass mixin or function
```
vanitabarrett, line 73:
```suggestion
### Deprecating a parameter for a Sass mixin or function
```

docs/contributing/versioning.md
```diff
@@ -41,32 +41,6 @@ This is to ensure the community can get involved with the decision.
 
 Make an active effort to involve the community, this might be in the form of presentations or meetings.
 
-## Deprecation
```
vanitabarrett, line 44: There is an anchor link to this section [further up in the file](https://github.com/alphagov/govuk-frontend/pull/2679/files#diff-86a4fd213ed466a2707fbf35c759f4bf8fdf9a24139864757994ef8bc22dcdb1R34) which will break now the section has been removed.

docs/contributing/versioning.md
```diff
@@ -41,32 +41,6 @@ This is to ensure the community can get involved with the decision.
...
-
-### Sass doc deprecation annotations
-
-When deprecating Sass, you can use the [deprecated annotation](http://sassdoc.com/annotations/#deprecated).
-
-If there is an alias, for example if you are renaming something, you can use the [alias annotation](http://sassdoc.com/annotations/#alias).
-
-See [an example of deprecating a Sass mixin on GitHub](https://github.com/alphagov/govuk-frontend/blob/9424d87ed54764d2d8afe35d6e0077ee43d231e1/src/helpers/_grid.scss#L20-L26).
-
 ## Migration
```
vanitabarrett, line 44: Should this be part of the new doc too?

vanitabarrett (changes requested):

docs/contributing/managing-change.md
36degrees: We want as many breaking changes as possible to be 'actionable' in minor releases that come prior to the major release.

So if a user has followed the release notes by the time they upgrade to the next major version there should be few changes they actually have to make at that point.

Does that make sense? Agree it's clumsily worded at the minute – maybe @claireashworth can help?

docs/contributing/versioning.md
36degrees: I've resolved this by removing the migration section (which didn't really add anything over what was in the managing change doc) and updating the link to point to the managing change doc instead.

docs/contributing/managing-change.md
36degrees: Good catch!

docs/contributing/versioning.md
36degrees: I don't think there's anything in there that's not said in the new doc in more detail, so have removed.

docs/contributing/managing-change.md
claireashworth: How about something like this:

Breaking changes should, wherever possible, be actionable by the users. By doing this the users will have fewer actions to do to complete a major release.

vanitabarrett: Ah I see, thanks for explaining. So it's more about, where possible, telling users about changes we're making in patch/feature releases so they're given a heads up earlier and are given the opportunity to make changes to their service in advance to prepare?

36degrees: Exactly that 👍🏻

36degrees: @claireashworth I think the missing bit is that the changes should ideally be actionable in prior minor or patch versions.

That's not always going to be possible but we should aim to do it where it makes sense. The rest of the doc mostly provides strategies for doing so.

vanitabarrett: Could we say something like:

```Where possible, users should be notified about upcoming breaking changes in a patch or feature release. By doing this, users have the opportunity to make the relevant changes in advance, so they have fewer actions to complete when we do a major release.```

36degrees: I've pushed a change that tries to simplify the 'intro' down to the bare minimum, and let the other sections go in to more detail on how it helps.

Thoughts?

claireashworth: The intro looks good. A couple of style guide things - change 'whilst' to 'while' and expand 'don't' (negative verbs are written in full for gov.uk style)

vanitabarrett (approved):

PR #2680, opened 2022-07-01
[SPIKE] Investigate using Jest for unit testing

vanitabarrett:
Investigation into https://github.com/alphagov/govuk-frontend/issues/2559

## What
Spike to see how we can add unit tests for our component JavaScript. This spike explores using tooling we already have (i.e: Jest) rather than introducing a new library.

## Existing component testing 

At the moment, we have a lot of different testing strategies with a bit of overlap, but each intended to serve a useful and distinct (if only slightly distinct) purpose. Our snapshot tests check the full HTML rendering of a component, so can pick up on things like whitespace issues in attributes. Our template.test.js files also check HTML rendering of each component, but focuses more on making sure the parameters being passed into each component have the expected result. Note, these tests also run aXe for accessibility testing against the ‘default’ example of a component.

So far, none of this testing covers any of the JavaScript functionality our components might have. We also have *.test.js files which use the review app to test that component JS functionality works correctly. Because it uses the review app, this is more about testing the JS from the user’s POV. 

So we don’t currently have any tests properly testing the internals of our components. Hopefully any bugs in our JS would be surfaced in our *.test.js tests, but it can be tricky to debug and isn’t well suited for ‘helper’ things, e.g:  functions shared around in `common.mjs` and internationalisation. This is where unit testing comes in.

### Other tests to note:
- all.test.js
- template.test.js
- tests for our helper sass
- package.json test to check version of rollup
- full page example tests

## Proposal
Add unit tests for our component JS, to give us greater control over what components are initialised with. This lets us test different scenarios (e.g: initialising in different states, passing i18n config), which is a bit more tricky with our current test setup as it relies on the review app.

I started this spike assuming we’d write unit tests for all our component JS code and aim for ~100% code coverage. However, given the other tests we already have (see above), there may not be value in doing that.

A mid-way suggestion would be: add unit tests for JavaScript functions that are more “background logic”, e.g functions that:

- don’t have a visible effect on the component
- don’t change anything in the DOM

This might involve splitting out logic from long functions to make them easier to test in isolation.

### Example
Test out the above suggestion with the error summary and accordion components:

Error Summary:

|  `*.test.js`  | `*.unit.test.mjs`  |
|---|---|
| setFocus() - check the right attributes are added/removed and focus is set <br> focusTarget scrolling and focus behaviour |  init() <br> handleClick() <br> setFocus() when data-disable-auto-focus false <br> focusTarget() return value <br> getFragmentFromUrl() <br> getAssociatedLegendOrLabel() |

Accordion:

|  `*.test.js`  | `*.unit.test.mjs`  |
|---|---|
| initControls() - make sure the "right stuff" appears when JS is available <br> initSectionHeaders() <br> constructHeaderMarkup() <br> setExpanded() <br> updateShowAllButton() - would suggest moving the session storage stuff out into a separate function if possible, and unit test that instead  | init() <br> onSectionToggle() <br> onShowOrHideToggle() <br> isExpanded() <br> checkIfAllSectionsOpen() <br> storeState() <br> setInitialState() <br> getButtonPunctuationEl() |

## Questions left to answer

- [x] Are we happy with this approach, or do we want to investigate any other tools?
- [x] Do we think the proposal above seems sensible?
- [ ] Roll-out approach for unit testing, e.g: adding them to all our existing components now vs a gradual rollout, maybe starting with the internationalisation work?


domoscargin:
This looks good! Got a few "I am not a Jest/testing expert" type questions.

- Do we need to consider whether we test private methods or just the public methods that use them? `ErrorSummary.getAssociatedLegendOrLabel` and `ErrorSummary.getFragmentFromUrl`, for example, both feel like things we wouldn't expose (though perhaps both would be moved to a separate helper file).
- Just to clarify your handy tables - in the `*.test.js` files, would we be replacing the existing behavioural tests with ones based on the component methods? For example, would there be a `describe('setFocus()'` test suite in `error-summary.test.js`? And would we be shifting to `jsdom` vs `puppeteer`?
- Could we do some initial work to define and document exactly what we consider goes in the "unit testing" bucket, integration, e2e, snapshots, template, etc so it's very clear for contributors (and devs!)
- Would it be useful to do some thinking/spiking/research about more radical approaches/other testing suites/blank slate ideals before implementing? I'm of two minds whether it would just slow down this good work unnecessarily, or give us a clearer idea on our full testing strategy.

domoscargin:
I think it makes sense to just test public functions, since the private stuff should just be implementation details which could change. Was thinking about it from the perspective of avoiding writing tests for stuff we later decide doesn't get exposed - but we can probably make those calls in tandem with writing the tests. At a glance, the SkipLink has a `getFragmentFromUrl()` which looks pretty similar to the ErrorSummary, so I definitely foresee some kind of shared helper in our future!

Thanks for clarifying the tables for me - I was barking up the wrong tree entirely, and that all makes total sense now. I am also totally swayed by your arguments against ripping everything up.

In other words, I think this is a winner, and I think a gradual approach makes sense with i18n as the guinea pig.

vanitabarrett:
Closing this now as the spike has been completed. Going to write up follow-up cards and add them to https://github.com/alphagov/govuk-frontend/issues/2559

PR #2682, opened 2022-07-05
Update docs for typography helpers

36degrees:
As part of #772 we updated the `govuk-typography-responsive` mixin to accept a point from the spacing scale rather than a map (1ea1353, 93a0f33).

However the associated SassDocs were not updated at the same time, and still talk about accepting a `$font-map` param which is incorrect.

Update the SassDocs to reflect the way the mixin works.

Also update the SassDocs for the `govuk-font` mixin to use the same terminology, and correct ‘as it would appear on desktop’ to ‘as it would appear on tablet’ which is the breakpoint where the font-size actually changes.

### DIFF
--- a/src/govuk/helpers/_typography.scss
+++ b/src/govuk/helpers/_typography.scss
@@ -86,12 +86,13 @@
 
 /// Responsive typography helper
 ///
-/// Takes a 'font map' as an argument and uses it to create font-size and
-/// line-height declarations for different breakpoints, and for print.
+/// Takes a point from the responsive 'font map' as an argument (the size as it
+/// would appear on tablet and above), and uses it to create font-size and
+/// line-height declarations for different breakpoints, and print.
 ///
 /// Example font map:
 ///
-/// $my-font-map: (
+/// 19: (
 ///   null: (
 ///     font-size: 16px,
 ///     line-height: 20px
@@ -106,11 +107,14 @@
 ///   )
 /// );
 ///
-/// @param {Map} $font-map - Font map
+/// @param {Number} $size - Point from the spacing scale (the size as it would
+///   appear on tablet and above)
 /// @param {Number} $override-line-height [false] - Non responsive custom line
 ///   height. Omit to use the line height from the font map.
 /// @param {Boolean} $important [false] - Whether to mark declarations as
 ///   `!important`.
+///
+/// @throw if `$size` is not a valid point from the spacing scale
 ///
 /// @access public
 
@@ -167,11 +171,14 @@
 
 /// Font helper
 ///
-/// @param {Number} $size - Size of the font as it would appear on desktop -
-///   uses the responsive font size map
+/// @param {Number | Boolean} $size Point from the spacing scale (the size as it
+///   would appear on tablet and above). Use `false` to avoid setting a size.
 /// @param {String} $weight [regular] - Weight: `bold` or `regular`
 /// @param {Boolean} $tabular [false] - Whether to use tabular numbers or not
-/// @param {Number} $line-height [false] - Line-height, if overriding the default
+/// @param {Number} $line-height [false] - Line-height, if overriding the
+///   default
+///
+/// @throw if `$size` is not a valid point from the spacing scale (or false)
 ///
 /// @access public
 


src/govuk/helpers/_typography.scss
```diff
@@ -106,12 +107,15 @@
 ///   )
 /// );
 ///
-/// @param {Map} $font-map - Font map
+/// @param {Number} $size - Point from the spacing scale (typically the size as
```
owenatgov, line 110: You've written "**typically** the size..." where as below (lines 174 and 175) you've omitted the "typically". I'm wondering if we should either make these consistent or if we should be more specific here. I'm leaning towards the latter as my understanding is that the only instance when the `$size` wouldn't be a number on the font scale is if you wanted to pass `false` and not use the scale. This also contradicts the `@throw` on like 117.

Can you think of other scenarios besides a font scale value of `false`? If not, I'd suggest changing this and the similar line on 174 to something like:

>Point from the spacing scale (the size in px as it would appear on tablet and above) or `false`

I'd similarly either omit the `@throw` on line 117 or change it to:

>if `$size` is not a valid point from the spacing scale or `false`

src/govuk/helpers/_typography.scss
```diff
@@ -167,11 +171,12 @@
...
-/// @param {Number} $size - Size of the font as it would appear on desktop -
-///   uses the responsive font size map
+/// @param {Number} $size Point from the spacing scale (the size as it would
+///   appear on tablet and above)
 /// @param {String} $weight [regular] - Weight: `bold` or `regular`
 /// @param {Boolean} $tabular [false] - Whether to use tabular numbers or not
-/// @param {Number} $line-height [false] - Line-height, if overriding the default
+/// @param {Number} $line-height [false] - Line-height, if overriding the
+///   default
 ///
```
owenatgov, line 180: Do we want a `@throw` here or is `govuk-font` expected to fail silently? _Should_ it fail silently if so?

owenatgov (commented):
Thanks for tidying this up Ollie! A few little nitpicks, happy to discuss them as I may not have all the context. I also think it's worth running this past Claire for the words.

Also changelog update pls.

src/govuk/helpers/_typography.scss
36degrees: > Can you think of other scenarios besides a font scale value of false?

The other scenario where a point from the spacing scale wouldn't be the size on tablet and above is if the user has overridden `$govuk-typography-scale` – the fact that the 'key' is the same as the tablet size is only a convention and is not enforced in code, so a user could choose to use a different set of keys, e.g:

```scss
$govuk-typography-scale: (
    1: (
        null: (
        font-size: 53px,
        line-height: 55px
        ),
        tablet: (
        font-size: 80px,
        line-height: 80px
        ),
        print: (
        font-size: 53pt,
        line-height: 1.1
        )
    ),
    2: (
        null: (
        font-size: 32px,
        line-height: 35px
        ),
        tablet: (
        font-size: 48px,
        line-height: 50px
        ),
        print: (
        font-size: 32pt,
        line-height: 1.15
        )
    )
    // etc
)
```

However, this is a rather theoretical edge case and, because of the way we use the typography scale throughout the components, in reality would result in lots of things breaking.

> This also contradicts the @throw on like 117.

> I'd suggest changing this and the similar line on 174 to something like:
> 
>> Point from the spacing scale (the size in px as it would appear on tablet and above) or false

You can't call `govuk-typography-responsive` with `false` – it doesn't make sense as setting the font size is the only thing the mixin does, so it'll throw an error. You **can** call `govuk-font` with `$size: false` as it does lots of other things.

src/govuk/helpers/_typography.scss
36degrees: I've pushed [a change](https://github.com/alphagov/govuk-frontend/compare/42e46e13fd619b33d33c0b54cf68182c17e19462..9ffed11085ed9509bfd80a377f54e1b83d22e4de) that adds a `@throw` annotation to the `govuk-font` mixin as suggested.

owenatgov:
Re: the changelog, with a fresh head I agree that this is a relatively abstract docs change and isn't necessary to record

owenatgov (approved):
Great stuff! Still think it's worth running this past @claireashworth before merging to be belt and braces.

claireashworth (approved):
Looks good to me too.

PR #2685, opened 2022-07-11
Add missing ID to select example

### LINKED ISSUE #2683: The govuk select component may have an invalid example
## Description of the issue

PlatUI has recently upgraded our **play-frontend-hmrc** service with the new **govuk-frontend** version(**v4.2.0**)
after the upgrade, we had a failing view test around the newly added [select component example](https://github.com/alphagov/govuk-frontend/blob/8456499d3d7b08953e2abb092b17f020e0f4c29b/src/govuk/components/select/select.yaml#L309). In our [GovukSelect wrapper](https://github.com/hmrc/play-frontend-hmrc/blob/3436091343a0ddbc935ac28b45ac70fd0b9401fa/src/main/twirl/uk/gov/hmrc/govukfrontend/views/components/GovukSelect.scala.html#L21) we ensure that both the **name** and **id** attributes are required when using a **GovukSelect** component, as the **name** and **id** attributes are required by the select element.

The [select component example](https://github.com/alphagov/govuk-frontend/blob/8456499d3d7b08953e2abb092b17f020e0f4c29b/src/govuk/components/select/select.yaml#L309) is missing the **id** attribute. For now, we have patched the example to pass our tests but we feel that the **id** should be included in the [select component example](https://github.com/alphagov/govuk-frontend/blob/8456499d3d7b08953e2abb092b17f020e0f4c29b/src/govuk/components/select/select.yaml#L309).

- GOV.UK Frontend Version: v4.2.0

36degrees:
The example added in f321a1c is missing an id.

The id Nunjucks option is required, as a valid id attribute is required on the select element in order to create the association with the label.

Add an appropriate id to the example.

Fixes #2683.

### DIFF
--- a/src/govuk/components/select/select.yaml
+++ b/src/govuk/components/select/select.yaml
@@ -310,6 +310,7 @@ examples:
     hidden: true
     data:
       name: colors
+      id: colors
       items:
         - value: red
           text: Red


querkmachine (approved):
🌈

PR #2686, opened 2022-07-12
Update header macro documentation for `navigationLabel` and `menuButtonLabel`

vanitabarrett:
We removed the redundant 'navigation' text in the navigationLabel and
menuButtonLabel in https://github.com/alphagov/govuk-frontend/commit/aed4b3e60f991d47595f4af7a7d03c8ff7f22455

However, we didn't update the macro documentation to match.

This commit updates the macro documentation to the correct defaults.

### DIFF
--- a/src/govuk/components/header/header.yaml
+++ b/src/govuk/components/header/header.yaml
@@ -51,11 +51,11 @@ params:
 - name: navigationLabel
   type: string
   required: false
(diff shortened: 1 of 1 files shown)


36degrees (approved):
Great catch 👍🏻

vanitabarrett:
All credit goes to @querkmachine for noticing this in the internationalisation spike! 💯 🙏  I'm simply moving that fix out into a separate PR as it's not tied to that work

PR #2688, opened 2022-07-13
Update link to contribution guidance in readme

owenatgov:
The link to the contribution guidance in the readme was linking to a specific version of the guidance. This means that updates to the guidance when navigated to from the readme would be lost. This change ensures it stays up to date.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,7 +85,7 @@ Open Government 3.0 licence.
 
 ## Contributing
 
-[To learn how to help us build GOV.UK Frontend, see our contribution guidelines.](https://github.com/alphagov/govuk-frontend/blob/f07569c9e7724606970c0d2764f1e9ce7fa14092/CONTRIBUTING.md)
+[To learn how to help us build GOV.UK Frontend, see our contribution guidelines.](CONTRIBUTING.md)
 
 The govuk-frontend repository is public and we welcome contributions from anyone.
 


README.md
```diff
@@ -85,7 +85,7 @@ Open Government 3.0 licence.
 
 ## Contributing
 
-[To learn how to help us build GOV.UK Frontend, see our contribution guidelines.](https://github.com/alphagov/govuk-frontend/blob/f07569c9e7724606970c0d2764f1e9ce7fa14092/CONTRIBUTING.md)
+[To learn how to help us build GOV.UK Frontend, see our contribution guidelines.](https://github.com/alphagov/govuk-frontend/blob/main/CONTRIBUTING.md)
```
36degrees, line 88: This works fine as a solution, but if we wanted to avoid tying it to a specific branch we could also make it [a relative link](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#relative-links):

```suggestion
[To learn how to help us build GOV.UK Frontend, see our contribution guidelines.](CONTRIBUTING.md)
```

querkmachine (approved):
Nice catch 🎣

README.md
vanitabarrett, line 88: Should this be a relative link (e.g: `/CONTRIBUTING.md`) so it also works when you run it locally, on a branch etc?

36degrees (approved):
Suggested an alternative approach that might be slightly more robust, but equally happy with the change you've already made so I'll leave it to you to decide.

owenatgov:
@vanitabarrett @36degrees I've made the link the relative path as recommended. Gonna merge.

Issue #2216, opened 2021-04-30
Deprecate `govuk-header__navigation--no-service-name` class


36degrees:
## What

Deprecate the `govuk-header__navigation--no-service-name` class for removal in a future major release.

## Why

This class adds padding to the top of the navigation to compensate for a header that doesn't have a service name:

![Screenshot 2021-04-30 at 13 17 15](https://user-images.githubusercontent.com/121939/116728795-43c8d080-a9de-11eb-8ca2-ba2894c813a5.png)

However, there are no examples of this class, and it's not used in the template. It's not clear why you'd ever want to use it.

(If we do change our mind and decide to keep this class around for some reason, we should add an example for it to the review app)

## Who needs to know about this

Designers, developers

## Done when

- [x] Class annotated as deprecated
- [x] Release notes make it clear the class is deprecated and will be removed in a future major version.
- [x] Add a card to remove this in a breaking release

vanitabarrett:
As discussed in v4 scoping, the work to deprecate this isn't a breaking change. We should deprecate this in a smaller release and then remove it in the next breaking release.

PR #2694, opened 2022-07-14
Deprecate `.govuk-header__navigation--no-service-name`

querkmachine:
Marks the `.govuk-header__navigation--no-service-name` as having been deprecated in our Sass and changelog, as per #2216.

The class is not documented in our docs or review app, serves a dubiously useful use case, and I couldn't find any instances of it being used when searching public GitHub repos, with it only appearing in forks of Frontend and automatically-generated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 ## Unreleased
 
+### Deprecated features
+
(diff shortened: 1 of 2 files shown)


owenatgov (approved):
Does the job 👍🏻

PR #2702, opened 2022-07-15
Allow localisation of content licence and copyright notices in Footer

querkmachine:
This are the bare minimum changes needed for enable localisation of the Footer component's content licence and copyright notices. In this case, it is assumed that an Open Government Licence (OGL) and Crown copyright are still being used, and that the notices are only being translated into another language.

This differs slightly from the internationalisation spike work, as I've split the `contentLicence` and `copyright` parameters into `html` and `text` versions to facilitate extending them with more options, as described below. 

---

Issues #1290 and #1291 originally asked that it be possible to remove the OGL and Crown graphics. This hasn't been done in this PR, as it isn't strictly needed for the current localisation work and (for the most part) public-facing GOV.UK services will have been made available under OGL and Crown copyright. 

I've created a version of this work with more customisation options, including changing/removing the Crown copyright URL and removing the licence/copyright graphics, as a separate branch. You can [compare those branches here](https://github.com/alphagov/govuk-frontend/compare/kg-footer-i18n..kg-footer-i18n-extended). Whether we consider these changes necessary or not is up for discussion!

### DIFF
--- a/src/govuk/components/footer/footer.yaml
+++ b/src/govuk/components/footer/footer.yaml
@@ -67,6 +67,32 @@ params:
       type: object
       required: false
       description: HTML attributes (for example data attributes) to add to the anchor in the footer navigation section.
+- name: contentLicence
+  type: object
+  required: false
+  description: Object containing options for the content licence.
+  params:
+  - name: text
+    type: string
+    required: false
+    description: If `html` is set, this is not required. If `html` is provided, the `text` option will be ignored. If neither are provided, the text for the Open Government Licence is used.
+  - name: html
+    type: string
+    required: false
+    description: If `text` is set, this is not required. If `html` is provided, the `text` option will be ignored. If neither are provided, the text for the Open Government Licence is used. The content licence is inside a `<span>` element, so you can only add [phrasing content](https://html.spec.whatwg.org/#phrasing-content) to it.
+- name: copyright
+  type: object
+  required: false
+  description: Object containing options for the copyright notice.
+  params:
+  - name: text
+    type: string
+    required: false
+    description: If `html` is set, this is not required. If `html` is provided, the `text` option will be ignored. If neither are provided, Crown copyright is used.
+  - name: html
+    type: string
+    required: false
+    description: If `text` is set, this is not required. If `html` is provided, the `text` option will be ignored. If neither are provided, Crown copyright is used. The copyright notice is inside an `<a>` element, so you can only use text formatting elements within it.
 - name: containerClasses
   type: string
   required: false
@@ -116,6 +142,22 @@ examples:
 - name: default
   data:
     {}
+
+- name: with custom HTML content licence and copyright notice
+  description: Open Government Licence and Crown copyright notice translated into Welsh
+  data:
+    contentLicence:
+      html: 'Mae’r holl gynnwys ar gael dan <a class="govuk-footer__link" href="https://www.nationalarchives.gov.uk/doc/open-government-licence-cymraeg/version/3/" rel="license">Drwydded y Llywodraeth Agored v3.0</a>, ac eithrio lle nodir yn wahanol'
+    copyright:
+      html: '<span>Hawlfraint y Goron</span>'
+
+- name: with custom text content licence and copyright notice
+  description: Open Government Licence and Crown copyright notice translated into Welsh
+  data:
+    contentLicence:
+      text: 'Mae’r holl gynnwys ar gael dan Drwydded y Llywodraeth Agored v3.0, ac eithrio lle nodir yn wahanol'
+    copyright:
+      text: '© Hawlfraint y Goron'
 
 - name: with meta
   description: Secondary navigation with meta information relating to the site
@@ -378,6 +420,13 @@ examples:
   hidden: true
   data:
     containerClasses: app-width-container
+- name: with HTML passed as text content
+  hidden: true
+  data:
+    contentLicence:
+      text: 'Mae’r holl gynnwys ar gael dan <a class="govuk-footer__link" href="https://www.nationalarchives.gov.uk/doc/open-government-licence-cymraeg/version/3/" rel="license">Drwydded y Llywodraeth Agored v3.0</a>, ac eithrio lle nodir yn wahanol'
+    copyright:
+      text: '<span>Hawlfraint y Goron</span>'
 - name: with empty meta
   hidden: true
   data:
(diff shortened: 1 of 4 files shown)


src/govuk/components/footer/footer.yaml
```diff
@@ -67,6 +67,32 @@ params:
...
       description: HTML attributes (for example data attributes) to add to the anchor in the footer navigation section.
+- name: contentLicence
+  type: object
+  required: false
+  description: Object containing options for the content licence.
+  params:
+  - name: text
+    type: string
+    required: false
+    description: If `html` is set, this is not required. Text to use within the disclosed part of the details element. If `html` is provided, the `text` option will be ignored. If neither are provided, the Open Government Licence is used.
```
vanitabarrett, line 78: I wonder if this description implies that if you pass in your own text (or html), the Open Government Licence isn't used? Which isn't quite correct, because the logo will still be there.

querkmachine: Perhaps. Most other examples of this kind include the actual text string that it falls back to, but in the case of the content licence this is fairly long and contains an embedded link, so including it wholesale didn't seem like a good idea.

vanitabarrett: Maybe just a tweak needed then, e.g: `If neither are provided, the text for he Open Government Licence is used.`?

vanitabarrett:
This PR makes sense to me (although needs tests and a Changelog entry to make it 'release ready' - aware it's still in draft). I'm happy to do just what's needed to enable internationalisation, although can see benefits of the other approach too. If we go with this PR, do you think it would then involve a breaking change to introduce the ability to completely override the logo etc too?

querkmachine:
@vanitabarrett I don't think anything in this PR would block us expanding it to include the other options at a later date.

querkmachine:
I've added a changelog entry but was a little unsure of how to focus it (as 'customisation', as 'localisation', as 'translation'), as it can now be changed, but still practically has to be OGL and Crown copyright. 

Can you give an expert opinion, @claireashworth?

claireashworth (commented):
Having the focus as localisation seems sensible to me. If the focus is customisation we should refer to the information from the National Archives as they own/manage the licence and copyright.

src/govuk/components/footer/footer.yaml
```diff
@@ -117,6 +143,14 @@ examples:
   data:
     {}
 
+- name: with custom content licence and copyright notice
```
vanitabarrett, line 146: Would be good to have text and html examples of each, so we're (hopefully) more like to catch [issues like this](https://github.com/alphagov/govuk-frontend/issues/2703).

src/govuk/components/footer/footer.yaml
```diff
@@ -67,6 +67,32 @@ params:
...
+  description: Object containing options for the content licence.
+  params:
+  - name: text
+    type: string
+    required: false
+    description: If `html` is set, this is not required. If `html` is provided, the `text` option will be ignored. If neither are provided, the text for the Open Government Licence is used.
+  - name: html
+    type: string
+    required: false
+    description: If `text` is set, this is not required. If `html` is provided, the `text` option will be ignored. If neither are provided, the text for the Open Government Licence is used.
```
vanitabarrett, line 82: It looks to me like any HTML content provided here will be inserted within a `span` element, e.g:

```
<span class="govuk-footer__licence-description">[contentLicence.html here]</span>
```

Do we need to do something [similar to the accordion component yaml ](https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/components/accordion/accordion.yaml#L30)and explain that only [phrasing content](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-span-element) should be passed in?

src/govuk/components/footer/template.test.js
```diff
@@ -207,4 +207,36 @@ describe('footer', () => {
...
+
+  describe('content licence', () => {
+    it('is visible', () => {
+      const $ = render('footer', examples.default)
+
+      const $licenceMessage = $('.govuk-footer__licence-description')
+      expect($licenceMessage.text().trim()).toContain('Open Government Licence v3.0')
+    })
+
+    it('can be customised', () => {
```
vanitabarrett, line 219: Similar to the examples comment above, might be good to have tests for both `text` and `html`. We've also previously written tests to check that HTML passed into the `text` option renders as text.

vanitabarrett (changes requested):
All works well, just a few comments about expanding the documentation/examples

src/govuk/components/footer/footer.yaml
querkmachine: Makes sense, I wasn't sure if that was a thing covered by the parameter docs. 

I'll add it for `copyright.html` too, as that's presented within a link that isn't configurable/removable in this version of the work.

src/govuk/components/footer/footer.yaml
querkmachine: What would be the best way to present this? Have two examples in the review app with the same text content but one passed through `html` and the other with `text`?

querkmachine: Done 👍

src/govuk/components/footer/footer.yaml
querkmachine: Have added text similar to what appears on the accordion.

The copyright instance was a little funky as there are various phrasing content elements that don't make sense within the link there (a, button, iframe, input (and other form elements), label). I've tried to make this a little clearer by making it "non-interactive phrasing content", but it might be better put as "you can only add text formatting elements".

vanitabarrett: Maybe one for @claireashworth to have a quick look at?

vanitabarrett:
Looks good to me! Would be great to get the commits tidied up/squashed a bit before merging.

src/govuk/components/footer/footer.yaml
claireashworth: Are you referring to lines 78 and 82? If so they read clearly to me.

claireashworth: If you are adding more to the description, the second option of only adding text formatting elements is better.

querkmachine: I've updated the sentence to read "The copyright notice is inside an `<a>` element, so you can only use text formatting elements within it."

vanitabarrett (approved):
Looks good to me 🎉

PR #2704, opened 2022-07-19
Fix variable name in accordion JavaScript

36degrees:
The `newButtonText` variable is a string and not a DOM element, and should therefore not be prefixed with `$`.

I don't believe this needs a changelog entry as it's not a part of the public API for the component.


domoscargin:
@36degrees did you mean to edit the `/package` file?

36degrees:
🤦🏻

domoscargin (approved):
💲

PR #2705, opened 2022-07-19
Tidy up the views used by the review app

36degrees:


owenatgov (approved):
All makes sense to me!

PR #2706, opened 2022-07-19
Redirect Sass docs when running on Heroku

### LINKED ISSUE #1892: Redirect Sass docs in review app to Sass API reference in Frontend Docs

36degrees:
When running on Heroku (based on the presence of the `HEROKU_APP` environment variable set in `app.json`):

- link to the live API reference docs from the homepage of the review app
- redirect requests for `/docs` to the live API reference docs

This avoids the confusion of having two ‘live’ versions of the Sass docs available, whilst redirecting anyone who may have the existing docs bookmarked.

However, it’s still useful to be able to run the 'generic' Sass docs locally so that we can use it to preview documentation before we do a release. As such, when running locally:

- link to both the live API reference docs and the ‘local’ docs from the homepage of the review app
- preserve the existing behaviour of serving the locally generate Sass docs when visiting `/docs`

Closes #1892.

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -13,6 +13,8 @@ const helperFunctions = require('../lib/helper-functions')
 const fileHelper = require('../lib/file-helper')
 const configPaths = require('../config/paths.json')
 
+const isDeployedToHeroku = !!process.env.HEROKU_APP
+
 const appViews = [
   configPaths.layouts,
   configPaths.views,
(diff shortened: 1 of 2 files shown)


36degrees:
Built on top of #2705 which should be merged first. Once that PR is merged, the target branch for this PR should automatically update to `main`.

vanitabarrett (approved):

PR #2719, opened 2022-07-21
Move #2673 out of v4.2.0 release in change log

36degrees:
#2673 was merged *after* v4.2.0 was released and remains an unreleased feature. There was no conflict in 536edad as the surrounding lines had not changed.


domoscargin (approved):
Whelp! Thanks for catching.

Issue #2703, opened 2022-07-15
Accordion content.text doesn't render with the right font


joelanman:
## Related component

Accordion

I wasn't sure if this is a bug or a feature request!

If you use the `content.text` option in accordion, the text doesn't look right - it doesn't use the normal gov.uk typography. As such, I'm not sure if `content.text` is useful? I think you'd have to add your own css to style it.

Maybe `content.text` could be wrapped in `<p class="govuk-body">` by default?

vanitabarrett:
Agree that doesn't seem right. I'm surprised it's not been raised as an issue before now, but maybe a lot of people use the `html` option instead!

querkmachine:
Note: This is also the case if when using `text` in the tabs component's panels.

PR #2723, opened 2022-07-22
Style accordion and tabs text content with `govuk-body` class

vanitabarrett:
Fixes https://github.com/alphagov/govuk-frontend/issues/2703

## What
Wraps `content.text` and `panel.text` in a paragraph with the `govuk-body` class when passed into the accordion and tabs component respectively. A user can pass in `content.html` or `panel.html` if they need something more custom.

## Why
Passing in `content.text` or `panel.text` meant displaying text in the default font as it wasn't being styled.

## Before vs After (accordion)
Screenshot 2022-07-22 at 10 54 40
Screenshot 2022-07-22 at 10 23 18

## Before vs After (tabs)
Screenshot 2022-07-22 at 10 27 54
Screenshot 2022-07-22 at 10 28 50

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,12 +22,13 @@ This change was introduced in [pull request #2694: Deprecate `.govuk-header__nav
 
 ### Fixes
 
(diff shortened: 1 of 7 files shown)


owenatgov (commented):
Code looks sharp. Needs a changelog entry then I'll approve.

owenatgov (approved):
✅  Also approved the Percy build

PR #2724, opened 2022-07-22
Remove redundant `aria-hidden` attribute from the content when using the Details polyfill

36degrees:
As discussed in https://github.com/alphagov/govuk-design-system/pull/2261#issuecomment-1192385380, when `$content` is hidden it is removed from the accessibility tree through the use of `display: none;`. The `aria-hidden` attribute is redundant and can be removed:

> If an interactive element is hidden using display:none or visibility:hidden (either on the element itself, or any of the element's ancestors), it won't be focusable, and it will also be removed from the accessibility tree. This makes the addition of aria-hidden="true" or explicitly setting tabindex="-1" unnecessary.
>
> https://www.w3.org/TR/using-aria/#4thrule

I've also tried to simplify the function slightly by:

- setting all of the attributes based on the `open` attribute to avoid the possibility of the different attributes getting out of sync
- setting everything in one logical flow rather than setting some things using ternary and some things in an if/else

(It's probably easier to review the 2 commits individually)


owenatgov (commented):
One little comment on efficiency. Looking tidy.

owenatgov (approved):
✅

PR #2725, opened 2022-07-22
Remove padding-right from last column in summary list row

### LINKED ISSUE #1928: Suggestion: remove right padding from last column in summary list
The summary list removes the right padding from the actions column - presumably so things can right align nicely.

If you don't have an actions column, I think the `value` column should also have no right padding.

In my own service, I've done this:
``` css
.govuk-summary-list__row dd:last-of-type {
  padding-right: 0;
}
```

My use case is this `dl`, which doesn't have an actions column:
Screenshot 2020-08-26 at 12 17 50

With the above fix:
Screenshot 2020-08-26 at 12 18 10

querkmachine:
Tiny change to CSS so that the last column in a summary list always has right padding removed, not only if the last column is an actions column. Resolves #1928. 

The line that removes right padding from the actions column has been maintained for IE8, which doesn't support `:last-child`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ We’ve also made fixes in the following pull requests:
 - [#2670: Define mininimum width for select component](https://github.com/alphagov/govuk-frontend/pull/2670)
 - [#2723: Style accordion and tabs text content with `govuk-body` class](https://github.com/alphagov/govuk-frontend/pull/2723)
 - [#2724: Remove redundant `aria-hidden` attribute from the content when using the Details polyfill](https://github.com/alphagov/govuk-frontend/pull/2724)
+- [#2725: Remove padding-right from last column in summary list row](https://github.com/alphagov/govuk-frontend/pull/2725)
 
 ## 4.2.0 (Feature release)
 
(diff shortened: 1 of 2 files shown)


36degrees (commented):
Neat solution 👍🏻 

One thing to be aware of is that when only some rows in a summary list have actions, there will only be padding on the 'value' column when there are actions.

(The rows that don't have actions use a pseudo-element to fill the void where the actions would be be, and TIL that pseudo-elements ['do not affect the interpretation of structural pseudo-classes'](https://www.w3.org/TR/selectors/#pseudo-elements).)

I don't think this is necessarily worth worrying about, as there's no need for a 'gap' before the non-existent actions, but the text in the rows with and without actions will wrap at different points:

Screenshot 2022-07-22 at 14 07 49

@christopherthomasdesign or @Ciandelle thoughts? Any concerns with this approach?

christopherthomasdesign:
I think ideally all items in the same column would align with one another, but I don't think it's super critical here. I've seen summary cards used for large blocks of text but they're more commonly used for individual lines or lists, where you wouldn't really notice the difference. Even for blocks of text, the ragged right edge of it will obscure the lack of alignment a bit.

So unless there's a simple way to make something like Ollie's screenshot above line up, I'm pretty happy to go with this approach tbh

36degrees (approved):
👏🏻 👍🏻

Issue #775, opened 2018-06-07
Header component has visible unusable menu button when CSS fail to load. 


nickcolley:
Not a big deal but might be nice to use the 'hidden' attribute, or a hardcoded 'display: none'.

<img width="325" alt="Screenshot 2021-09-01 at 10 34 39" src="https://user-images.githubusercontent.com/29889908/131648607-0d52fbe3-bd80-4431-a13f-d1e83a91608d.png">

## Done when

- [ ] TBC
- [ ] "Details of breaking change" section below completed to help with writing release notes

## Details of breaking change

- which users are affected: TBC
- the change that’s been made: TBC
- changes users will have to make: TBC
- impact of users not making those changes: TBC

igloosi:
button is also visible if CSS succeeds and JS fails

PR #2727, opened 2022-07-25
Make use of `hidden` in header navigation functionality

owenatgov:
Fixes https://github.com/alphagov/govuk-frontend/issues/775

## What
Replaces the header menu's default CSS based show/hide functionality with a reliance on the `hidden` attribute. Lifted from [a similar change to the design system website navigation](https://github.com/alphagov/govuk-design-system/pull/2106).

Additionally does not load the mobile menu UI if `matchMedia` is not present and instead presents the "no JS" view to those browsers (IE9 and below).

## Why
To stop an unusable menu button appearing if CSS fails to load.

The choice to make `matchMedia` a required function comes from us needing a way to move `hidden` attributes around between screen sizes without having to rely on CSS overrides, which kicks the problem down the road for users without CSS. As our IE9 and below users are low, it's unlikely that they'll be on screens that would require the mobile menu and they can still use the menu regardless, this is a low risk change.

## Upgrade notes
This change introduces a `hidden` attribute to the menu button on the header markup, meaning that teams that use the header but don't use our macros will need to update their markup. However the component still operates without the hidden attribute thanks to fallbacks in the CSS, this just means that teams won't get the benefit of `hidden` on no CSS and no JS views. In this respect we aren't introducing a regression so I recommend this go into the next release non-breaking release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,18 @@ We've deprecated the `govuk-header__navigation--no-service-name` class, and will
 
 This change was introduced in [pull request #2694: Deprecate `.govuk-header__navigation--no-service-name`](https://github.com/alphagov/govuk-frontend/pull/2694).
 
+### Recommended changes
+
+We've recently made some non-breaking changes to GOV.UK Frontend. Implementing these changes will make your service work better.
+
+#### Add `hidden` to the mobile menu button in the header component
+
+If you're not using the Nunjucks macros, add the `hidden` attribute to the mobile menu button (button with class `govuk-header__menu-button`) in the header component.
+
+We've changed the header's mobile menu functionality to use the `hidden` attribute over using CSS to show/hide the mobile menu. Adding `hidden` to the mobile menu button by default will ensure that it does not display for users when javascript doesn't load.
+
+This change was introduced in [pull request 2727: Make use of hidden in header navigation functionality](https://github.com/alphagov/govuk-frontend/pull/2727)
+
 ### Fixes
 
 In [pull request 2678: Replace ex units with ems for input lengths](https://github.com/alphagov/govuk-frontend/pull/2678), we changed how we define input lengths in our CSS. Browsers might now display these inputs as being slightly wider than before. The difference is usually fewer than 3 pixels.
--- a/src/govuk/components/header/_index.scss
+++ b/src/govuk/components/header/_index.scss
@@ -210,11 +210,14 @@
     @include govuk-media-query ($from: tablet) {
       top: govuk-spacing(3);
     }
-  }
 
-  .govuk-header__menu-button--open {
-    &:after {
-      @include govuk-shape-arrow($direction: up, $base: 10px, $display: inline-block);
+    .js-enabled & {
+      display: block;
+    }
+
+    &[hidden],
+    .js-enabled &[hidden] {
+      display: none;
     }
   }
 
@@ -229,25 +232,9 @@
     margin: 0;
     padding: 0;
     list-style: none;
-  }
 
-  .js-enabled {
-    .govuk-header__menu-button {
-      display: block;
-      @include govuk-media-query ($from: desktop) {
-        display: none;
-      }
-    }
-
-    .govuk-header__navigation-list {
+    &[hidden] {
       display: none;
-      @include govuk-media-query ($from: desktop) {
-        display: block;
-      }
-    }
-
-    .govuk-header__navigation-list--open {
-      display: block;
     }
   }
 
--- a/src/govuk/components/header/_index.scss
+++ b/src/govuk/components/header/_index.scss
@@ -207,6 +207,10 @@
       margin-left: govuk-spacing(1);
     }
 
+    &[aria-expanded="true"]:after {
+      @include govuk-shape-arrow($direction: up, $base: 10px, $display: inline-block);
+    }
+
     @include govuk-media-query ($from: tablet) {
       top: govuk-spacing(3);
     }
--- a/src/govuk/components/header/header.mjs
+++ b/src/govuk/components/header/header.mjs
@@ -8,6 +8,17 @@ function Header ($module) {
   this.$menu = this.$menuButton && $module.querySelector(
     '#' + this.$menuButton.getAttribute('aria-controls')
   )
+
+  // Save the opened/closed state for the nav in memory so that we can
+  // accurately maintain state when the screen is changed from small to
+  // big and back to small
+  this.menuIsOpen = false
+
+  // A global const for storing a matchMedia instance which we'll use to
+  // detect when a screen size change happens. We set this later during the
+  // init function and rely on it being null if the feature isn't available
+  // to initially apply hidden attributes
+  this.mql = null
 }
 
 /**
@@ -15,27 +26,58 @@ function Header ($module) {
  *
  * Check for the presence of the header, menu and menu button – if any are
  * missing then there's nothing to do so return early.
+ * Feature sniff for and apply a matchMedia for desktop which will
+ * trigger a state sync if the browser viewport moves between states. If
+ * matchMedia isn't available, hide the menu button and present the "no js"
+ * version of the menu to the user.
  */
 Header.prototype.init = function () {
   if (!this.$module || !this.$menuButton || !this.$menu) {
     return
   }
 
-  this.syncState(this.$menu.classList.contains('govuk-header__navigation-list--open'))
-  this.$menuButton.addEventListener('click', this.handleMenuButtonClick.bind(this))
+  if ('matchMedia' in window) {
+    // Set the matchMedia to the govuk-frontend desktop breakpoint
+    this.mql = window.matchMedia('(min-width: 48.0625em)')
+
+    if ('addEventListener' in this.mql) {
+      this.mql.addEventListener('change', this.syncState.bind(this))
+    } else {
+      // addListener is a deprecated function, however addEventListener
+      // isn't supported by IE or Safari. We therefore add this in as
+      // a fallback for those browsers
+      this.mql.addListener(this.syncState.bind(this))
+    }
+
+    this.syncState()
+    this.$menuButton.addEventListener('click', this.handleMenuButtonClick.bind(this))
+  } else {
+    this.$menuButton.setAttribute('hidden', '')
+  }
 }
 
 /**
  * Sync menu state
  *
- * Sync the menu button class and the accessible state of the menu and the menu
- * button with the visible state of the menu
- *
- * @param {boolean} isVisible Whether the menu is currently visible
+ * Uses the global variable menuIsOpen to correctly set the accessible and
+ * visual states of the menu and the menu button.
+ * Additionally will force the menu to be visible and the menu button to be
+ * hidden if the matchMedia is triggered to desktop.
  */
-Header.prototype.syncState = function (isVisible) {
-  this.$menuButton.classList.toggle('govuk-header__menu-button--open', isVisible)
-  this.$menuButton.setAttribute('aria-expanded', isVisible)
+Header.prototype.syncState = function () {
+  if (this.mql.matches) {
+    this.$menu.removeAttribute('hidden')
+    this.$menuButton.setAttribute('hidden', '')
+  } else {
+    this.$menuButton.removeAttribute('hidden')
+    this.$menuButton.setAttribute('aria-expanded', this.menuIsOpen)
+
+    if (this.menuIsOpen) {
+      this.$menu.removeAttribute('hidden')
+    } else {
+      this.$menu.setAttribute('hidden', '')
+    }
+  }
 }
 
 /**
@@ -45,8 +87,8 @@ Header.prototype.syncState = function (isVisible) {
  * sync the accessibility state and menu button state
  */
 Header.prototype.handleMenuButtonClick = function () {
-  var isVisible = this.$menu.classList.toggle('govuk-header__navigation-list--open')
-  this.syncState(isVisible)
+  this.menuIsOpen = !this.menuIsOpen
+  this.syncState()
 }
 
 export default Header
--- a/src/govuk/components/header/header.test.js
+++ b/src/govuk/components/header/header.test.js
@@ -25,10 +25,19 @@ describe('Header navigation', () => {
     })
 
     it('shows the navigation', async () => {
-      await expect(page).toMatchElement('.govuk-header__navigation', {
-        visible: true,
-        timeout: 1000
-      })
+      const navDisplay = await page.$eval('.govuk-header__navigation-list',
+        el => window.getComputedStyle(el).getPropertyValue('display')
+      )
+
+      expect(navDisplay).toBe('block')
+    })
+
+    it('does not show the mobile menu button', async () => {
+      const buttonDisplay = await page.$eval('.govuk-js-header-toggle',
+        el => window.getComputedStyle(el).getPropertyValue('display')
+      )
+
+      expect(buttonDisplay).toBe('none')
     })
   })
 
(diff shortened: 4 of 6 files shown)


CHANGELOG.md
```diff
@@ -24,6 +24,8 @@ This change was introduced in [pull request #2694: Deprecate `.govuk-header__nav
 
 In [pull request 2678: Replace ex units with ems for input lengths](https://github.com/alphagov/govuk-frontend/pull/2678), we changed how we define input lengths in our CSS. Browsers might now display these inputs as being slightly wider than before. The difference is usually fewer than 3 pixels.
 
+In [pull request 2727: Make use of hidden in header navigation functionality](https://github.com/alphagov/govuk-frontend/pull/2727), we updated header component's mobile navigation functionality to better account for if CSS or javascript don't load. This feature doesn't require changes to your component code to continue working, however we recommend adding the attribute `hidden` to the mobile menu button (button element with class `govuk-header__menu-button`) to get the full benefits of the change.
```
owenatgov, line 27: Made a best guess with this. It may not be the right place to put this and may not even need this level of detail.

src/govuk/components/header/header.mjs
```diff
@@ -8,20 +8,38 @@ function Header ($module) {
...
+ * Additionally feature sniff for and apply a matchMedia for desktop which will
+ * trigger a state sync if the browser viewport moves between states
  */
 Header.prototype.init = function () {
   if (!this.$module || !this.$menuButton || !this.$menu) {
     return
   }
 
-  this.syncState(this.$menu.classList.contains('govuk-header__navigation-list--open'))
+  if (typeof window.matchMedia === 'function') {
```
querkmachine, line 37: This could alternatively be `if ('matchMedia' in window)` to return a boolean without needing `typeof`.

src/govuk/components/header/header.mjs
```diff
@@ -30,12 +48,23 @@ Header.prototype.init = function () {
...
- * @param {boolean} isVisible Whether the menu is currently visible
  */
-Header.prototype.syncState = function (isVisible) {
-  this.$menuButton.classList.toggle('govuk-header__menu-button--open', isVisible)
-  this.$menuButton.setAttribute('aria-expanded', isVisible)
+Header.prototype.syncState = function () {
+  if (this.mql === null || !this.mql.matches) {
+    this.$menuButton.removeAttribute('hidden')
+
+    this.$menuButton.classList.toggle('govuk-header__menu-button--open', this.menuIsOpen)
```
querkmachine, line 56: As a thought (maybe outside of the scope of this work), would it be an idea to hook the styles onto a selector like `.govuk-header__menu-button[aria-expanded="true"]` and get rid of having to add/remove an additional class?

CHANGELOG.md
querkmachine: The 'recommended changes' from 4.1.0 might be good examples to pull from for this? 
https://github.com/alphagov/govuk-frontend/releases/tag/v4.1.0

querkmachine (approved):
Left some suggestions. Other than the changelog entry probably needing rewriting this looks good and does what it says on the tin!

src/govuk/components/header/_index.scss
```diff
@@ -210,6 +210,22 @@
...
+      display: none;
+    }
+
+    .js-enabled & {
+      display: block;
+
+      @include govuk-media-query ($from: desktop) {
+        display: none;
+      }
+    }
```
36degrees, line 228: I'm a little bit lost with all of the `display: none` and `display: block` rules and how they interact with the `hidden` attribute.

It feels like we should be able to simplify things quite a lot as we should be able to rely on the hidden attribute? The only exception I think being where the service has not added the hidden attribute to the HTML and the JS doesn't run?

src/govuk/components/header/header.mjs
```diff
@@ -30,12 +48,23 @@ Header.prototype.init = function () {
  *
  * Sync the menu button class and the accessible state of the menu and the menu
  * button with the visible state of the menu
- *
- * @param {boolean} isVisible Whether the menu is currently visible
  */
-Header.prototype.syncState = function (isVisible) {
-  this.$menuButton.classList.toggle('govuk-header__menu-button--open', isVisible)
-  this.$menuButton.setAttribute('aria-expanded', isVisible)
+Header.prototype.syncState = function () {
```
36degrees, line 52: Think the comment for this function could do with updating.

src/govuk/components/header/header.test.js
```diff
@@ -30,6 +30,13 @@ describe('Header navigation', () => {
         timeout: 1000
       })
     })
+
+    it('does not show the mobile menu button', async () => {
+      await expect(page).toMatchElement('.govuk-js-header-toggle', {
+        visible: false,
```
36degrees, line 36: I tried to do the same thing recently and came to the conclusion that I don't think this does what you think it does – setting `visible: false` doesn't mean 'the element must not be visible in order to match'. Instead it means 'the element doesn't _have_ to be visible in order to match'. So I think this would pass even if the element was visible.

I also tried negating it (`expect(page).not.toMatchElement`), but that doesn't work either – see https://github.com/smooth-code/jest-puppeteer/issues/459.

Please do double check and see if you reach the same conclusion! But I ended up getting the computed style instead – see https://github.com/alphagov/govuk-frontend/commit/e6aa87a14302c1fae741b39757c367f2f48a43df for an example.

src/govuk/components/header/header.test.js
```diff
@@ -49,6 +56,32 @@ describe('Header navigation', () => {
         })
       })
 
+      it('reveals the menu button', async () => {
+        const hidden = await page.$eval('.govuk-js-header-toggle',
+          el => el.getAttribute('hidden')
```
36degrees, line 61: Suggest using `hasAttribute` and expecting it to be false, not falsey.

src/govuk/components/header/header.test.js
```diff
@@ -49,6 +56,32 @@ describe('Header navigation', () => {
...
+
+        await expect(page).toMatchElement('.govuk-js-header-toggle', {
+          visible: true,
+          timeout: 1000
+        })
+      })
+
+      it('hides the menu via the hidden attribute', async () => {
+        const hidden = await page.$eval('.govuk-header__navigation-list',
+          el => el.getAttribute('hidden')
```
36degrees, line 74: Suggest using `hasAttribute` and expecting it to be true, not empty string.

src/govuk/components/header/header.test.js
```diff
@@ -49,6 +56,32 @@ describe('Header navigation', () => {
...
+        )
+
+        expect(hidden).toEqual('')
+
+        await expect(page).toMatchElement('.govuk-header__navigation-list', {
+          visible: false,
```
36degrees, line 80: Same issue as described in my previous comment about `visible: false`.

src/govuk/components/header/_index.scss
owenatgov: There are 2 reasons that we've got any `display: none`'s at all in the CSS:

1. To account for older browser which don't support `hidden` (-IE11)
2. To account for when teams haven't added the initial `hidden` and we don't want to introduce a regression on their no js version of the header ie: the menu button being visible with CSS but without JS.

Do you think these are overcautious? I'm gonna see if there's a way we can slim this down anyway, maybe relying on the initial `display: block`'s without the styling attached to the `hidden`'s.

src/govuk/components/header/header.mjs
owenatgov: This isn't something we do currently anywhere else in the codebase, but that's not a reason not to do it as it's a neat trick.

src/govuk/components/header/_index.scss
36degrees: The `display: none`s are probably fine, but I think we can get rid of at least some of the `display: block` rules? The element should 'naturally' be `display: block` as long as the `hidden` attribute isn't present and the media query doesn't match.

owenatgov: I believe I've managed to slim this down (not pushed yet) however I think I've discovered why it initially seemed confusing. The awkwardness comes from the fact that services teams might not have updated their markup to include the `hidden` attribute on the menu by default, meaning that if we push this out without accounting for it in CSS then we could introduce a regression. This creates an awkward dynamic where we can reliably use `&[hidden]` to handle display in CSS for the nav list as we're only ever adding it via js but we can't do this for the menu button as it relies on `hidden` in the markup to manage default behaviour and there's a small chance that `hidden` wont be there.

This feels dodgy to me as it means we have 2 different approaches to fallback hiding in the same sass file. Some options:

1. Leave it as is and create an issue to use the `&[hidden]` rule for the menu button as well as the nav list
2. Make it consistent but save this for a breaking change
3. Make it consistent and just ship it

My instinct is 2 but I'm also not sure how serious an issue this if we went with 3. Interested to get some alternative perspectives.

36degrees: Doesn't the JS add the hidden attribute to the menu?

owenatgov: To the menu, yes. In theory the menu never needs the hidden attribute without js and it's far less risk if it's visible compared to if an unusable menu button is visible so we can reliably use only `hidden` and fallback CSS to manage it's visibility.

CHANGELOG.md
owenatgov: Updated. It'll still need a review fro ma content designer anyway and hinges a bit on the outcome of[ the discussion below](https://github.com/alphagov/govuk-frontend/pull/2727#discussion_r931077527) but we've got this ready to go if needs be.

src/govuk/components/header/header.mjs
owenatgov: Updated. Let me know if it covers what the function is doing.

src/govuk/components/header/_index.scss
```diff
@@ -207,9 +207,21 @@
       margin-left: govuk-spacing(1);
     }
 
+    &[aria-expanded="true"]:after {
```
36degrees, line 210: Does this mean we can get rid of the `--open` modifier class?

src/govuk/components/header/header.mjs
```diff
@@ -8,34 +8,63 @@ function Header ($module) {
...
+
+    if (this.menuIsOpen) {
+      this.$menu.removeAttribute('hidden')
+    } else {
+      this.$menu.setAttribute('hidden', '')
+    }
+  } else if (this.mql === null || this.mql.matches) {
+    this.$menu.removeAttribute('hidden')
+    this.$menuButton.setAttribute('hidden', '')
+  }
```
36degrees, line 67: I find it a little hard to follow what's going on with the logic here, and I'm not sure it's doing what you want.

- `this.mql === null` is included in both the `if` and `if else` part of the conditional, and so can never be true in the `else if` (because it `this.mql === null` we'd always go into the `if` part of the conditional and never get to the `else`)
- `this.mql.match` can also only be `true` or `false`, so there's no need for the `else if` – it should just be `else`.

It might be easier to make this understandable if we either flip the logic or update the media query to use `max-width` so it matches on mobile/tablet.

What is it that we actually want to happen when the browser doesn't support `matchMedia`?

src/govuk/components/header/_index.scss
owenatgov: Whoops! Yes it does. Cheers for flagging.

src/govuk/components/header/header.mjs
owenatgov: The idea is that if `matchMedia` isn't supported it will only ever run the contents of that first `if` block which manages changes in state for mobile view. If it is supported and `syncState` is triggered via the `mql` `change` event, then it'll run the contents of the `else` which essentially clear the UI, hiding the mobile menu button and forcing the menu to be visible.

This has prompted me to check this on an older browser which don't support `matchMedia` and the menu isn't visible because it never reaches that attribute clearing stage 😬 Also when I look at this again I agree that there's no reason for that `else if` to be there. I'm gonna clear this up and switch it as suggested.

It's also probably worth making this change on the design system website where I lifted this solution from which I can do separately.

owenatgov: I've figured it out. The `this.mql === null` was a way to let the user use the menu and let the syncing function work without having `matchMedia` enabled and without having to repeat code for if the user had `matchMedia` or not since `this.mql.matches` errors if `mql` is null. There's obviously a tidier, less confusing way to do this.

owenatgov: This ended up being a bit of a can of worms. In trying to account for the user not having `matchMedia` available, we create a risk that users of those browsers wont be able to use the header navigation. For example if a user is on IE9 (doesn't support `matchMedia`) and JS does load but CSS doesn't, then the JS will add `hidden` to the menu, as we don't check screen size by any means other than match media so assume they're on mobile to let them use the mobile menu, and without the CSS overrides to force the menu to display, the menu will be hidden for no CSS users.

I can see 2 ways around this whilst still solving the linked issue:
1. Present a "no JS" view to users who don't have `matchMedia` available
2. Don't user `matchMedia`

I've opted for 1 as the number of IE9 and below users are very low, a pertinent point as we're planning to reduce support for IE11 and below in v5, they are unlikely to be on screens small enough to require the mobile menu UI and they can still access the links in the menu this way.

src/govuk/components/header/_index.scss
owenatgov: Following changes to this in another thread, I feel like the CSS is tidy enough to go back into review. I leave it to the review to decide if this needs further work or not. We can always return to this and fix it in later releases when we're more confident that teams will have added the `hidden` attribute to their mobile menu buttons.

src/govuk/components/header/header.mjs
```diff
@@ -8,34 +8,67 @@ function Header ($module) {
...
  */
 Header.prototype.init = function () {
   if (!this.$module || !this.$menuButton || !this.$menu) {
     return
   }
 
-  this.syncState(this.$menu.classList.contains('govuk-header__navigation-list--open'))
-  this.$menuButton.addEventListener('click', this.handleMenuButtonClick.bind(this))
+  if ('matchMedia' in window) {
+    this.mql = window.matchMedia('(max-width: 48.0625em)')
```
36degrees, line 40: If you set the viewport to exactly 769px you get the menu button showing with the desktop layout:

Screenshot 2022-08-01 at 14 57 33

owenatgov: This seemed to be solved by changing the `em` value to `px`.

owenatgov:
I've made an issue to handle the inconsistent `hidden` CSS fallback discussed in the above discussions https://github.com/alphagov/govuk-frontend/issues/2746

src/govuk/components/header/header.mjs
owenatgov: After a chat offline with @36degrees we've managed to get this working whilst maintaining `em`'s by switching it to `min-width`.

src/govuk/components/header/header.mjs
```diff
@@ -8,34 +8,68 @@ function Header ($module) {
...
   if (!this.$module || !this.$menuButton || !this.$menu) {
     return
   }
 
-  this.syncState(this.$menu.classList.contains('govuk-header__navigation-list--open'))
-  this.$menuButton.addEventListener('click', this.handleMenuButtonClick.bind(this))
+  if ('matchMedia' in window) {
+    // Set the matchMedia to the govuk-frontend desktop breakpoint
+    this.mql = window.matchMedia('(min-width: 48.0625em)')
+    this.mql.addEventListener('change', this.syncState.bind(this))
```
36degrees, line 42: Unfortunately it looks like [IE11 and older versions of Safari (< 14.0)](https://caniuse.com/mdn-api_mediaquerylist) don't support `MediaQueryList.addEventListener` even though they support `window.matchMedia`, so at the minute this doesn't work in those browsers (and throws errors).

We need to use the non-standard `MediaQueryList.addListener` instead, [like we do for tabs](https://github.com/alphagov/govuk-frontend/blob/159455d6c049f99242569afc87302337308ee8ff/src/govuk/components/tabs/tabs.mjs#L24-L28):

```suggestion
    this.mql.addListener(this.syncState.bind(this))
```

Or, if we wanted to future-proof the code against the deprecation:

```suggestion
    if (this.mql.hasOwnProperty('addEventListener')) {
      this.mql.addEventListener('change', this.syncState.bind(this))
    } else {
      this.mql.addListener(this.syncState.bind(this))
    }
```

Either way, the `addListener` method is deprecated (and this gets flagged by my editor), so might also be worth adding a comment to explain what's going on?

owenatgov: I amended the solution very slightly as js apparently doesn't like `this.mql.hasOwnProperty('addEventListener')` but this should now be working 👍🏻

owenatgov: After a chat on slack, this is now using `'addEventListener in this.mql` as it's tidier and more consistent with what I'm doing to sniff for `matchMedia`. The reason the codebase didn't like `hasOwnProperty` was due to a linting concern.

36degrees (approved):

Issue #941, opened 2018-08-03
Consider making more components callable


frankieroberto:
It feels fairly common to include html (eg multiple paragraphs), and possibly even other components, with a Details component, or other components such as Panel or Inset Text.

To this end, it'd be nice to make these component callable, so you could do this:

```
{% call govukDetails({summaryText: 'Test'}) -%}
  <p>hello</p>
  <p>world</p>
{% endcall %}
```

as an alternative to this:

```
{% set detailsHtml %}
  <p>hello</p>
  <p>world</p>
{% endset %}

{% call govukDetails({summaryText: 'Test', html: detailsHtml}) -%}
```

vanitabarrett:
This came up again today, suggested by a new user https://github.com/alphagov/govuk-frontend/issues/2299

> GDS components support passing HTML content inline or via a variable with separate set block. Often the structure of this content means that passing HTML content inline is messy or impossible. This results in many set blocks that are disconnected from their components, which is difficult to read.

> A solution to this would be to support the caller() block within GDS macros. From Nunjucks documentation: A call block enables you to call a macro with all the text inside the tag. This is helpful if you want to pass a lot of content into a macro. The content is available inside the macro as caller().

PR #2734, opened 2022-07-26
Update various components to be callable

querkmachine:
Updates various components to be [callable](https://mozilla.github.io/nunjucks/templating.html#call). 

Callers make it easier for users of the Design System to pass large amounts of HTML directly into a component, without having to first assign it to a variable and then pass it to the `html` parameter. 

Suggested in #941, and based on the list of components I commented there, with some changes. 

---

I intend to make these components callable. In all situations, the called content will replace and take priority over the `html` and `text` parameters. 

- [x] Details 
- [ ] Error summary (replacing the `descriptionHtml` and `descriptionText` parameters)
- [x] Inset text
- [ ] Notification banner
- [x] Panel

I'm no longer considering the following components:

- Cookie banner: This component accepts looping content to allow for segregating consent into multiple groups. This is incompatible with using a caller, which only supports one slot of called content per component. 
- Warning text: The main text of this component is presented inside of a `<strong>` element, negating the usefulness of using a caller, as it's unlikely to ever be more than one line of text.

### DIFF
--- a/src/govuk/components/details/details.yaml
+++ b/src/govuk/components/details/details.yaml
@@ -15,6 +15,10 @@ params:
   type: string
   required: true
   description: If `text` is set, this is not required. HTML to use within the disclosed part of the details element. If `html` is provided, the `text` option will be ignored.
+- name: caller
+  type: nunjucks-block
+  required: false
+  description: Not strictly a parameter but [Nunjucks code convention](https://mozilla.github.io/nunjucks/templating.html#call). Using a `call` block enables you to call a macro with all the text inside the tag. This is helpful if you want to pass a lot of content into a macro. To use it, you will need to wrap the entire details component in a `call` block.
 - name: id
   type: string
   required: false
--- a/src/govuk/components/details/template.njk
+++ b/src/govuk/components/details/template.njk
@@ -5,6 +5,6 @@
     </span>
   </summary>
   <div class="govuk-details__text">
-    {{ params.html | safe if params.html else params.text }}
+    {{ caller() if caller else (params.html | safe if params.html else params.text) }}
   </div>
 </details>
--- a/src/govuk/components/details/template.test.js
+++ b/src/govuk/components/details/template.test.js
@@ -53,6 +53,12 @@ describe('Details', () => {
     expect($summary.get(0).tagName).toEqual('summary')
   })
 
+  it('renders nested components using `call`', () => {
+    const $ = render('details', {}, '<div class="app-nested-component"></div>')
+
+    expect($('.govuk-details .app-nested-component').length).toBeTruthy()
+  })
+
   it('allows text to be passed whilst escaping HTML entities', () => {
     const $ = render('details', examples['html as text'])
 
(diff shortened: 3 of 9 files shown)


CHANGELOG.md
```diff
@@ -20,6 +20,20 @@ Visit The National Archives' [documentation on OGL and Crown copyright](https://
 
 This was added in [pull request #2702: Allow localisation of content licence and copyright notices in Footer](https://github.com/alphagov/govuk-frontend/pull/2702).
 
+#### Pass HTML directly into compatible components
+
+If using the Nunjucks macros, you can now pass HTML content directly into compatible components using [the Nunjucks `call` syntax](https://mozilla.github.io/nunjucks/templating.html#call). In these instances, the called HTML is treated the same as the `html` parameter.
```
domoscargin, line 25: Two thoughts:

1. I know this is very unlikely to be used by somebody who isn't familiar with Nunjucks, but should we make it clear how this interacts with `html` and `text` parameters? I could potentially see an edge case where somebody was passing in those options AND the called HTML and getting confused (but this does feel like an edge case!)
2. Should we add a little "this is in addition to the Fieldset component, which is already callable"? I can't see that note in the rest of the changelog, and might be useful for folks interested in this to see if they don't know about it.

CHANGELOG.md
```diff
@@ -20,6 +20,20 @@ Visit The National Archives' [documentation on OGL and Crown copyright](https://
...
+
+Components that have been updated to support this syntax are:
+
+- Details
+- Error summary (mapped to `descriptionHtml` parameter)
```
domoscargin, line 30: I'm now very superstitious about the Error Summary and any changes that might affect screenreaders, but I don't think there should be any issue (from our side) on this change.

domoscargin (approved):
This looks good to me - and anecdotally would've been useful in my previous work using `govuk-frontend`!

I've left two small notes on the Changelog, which aren't blocking.

CHANGELOG.md
querkmachine: The resulting HTML from the Nunjucks changes shouldn't be different between the call syntax and the `html` parameter, so I can't imagine there'll be a problem. Can always give @owenatgov a poke so he has to think about the error summary for a bit longer, though. 😜

CHANGELOG.md
domoscargin: 1. Yeah, that sounds good - no harm in being really clear I think.
2. Basically just using the shiny new things to help highlight the sorta shiny old thing...

CHANGELOG.md
domoscargin: Don't you put that evil on him!

owenatgov: NO MORE.

In all seriousness this looks fine to me. The problems with the error summary I'm pretty sure are how our focusing on load via js is interacting with screen readers and less so the markup. This will impact my proposed error summary change, i it goes in, but that's something for me to deal during the rebase process.

CHANGELOG.md
querkmachine: I've updated the changelog.

domoscargin: I think you've formatted the entire CHANGELOG file in that latest push.

querkmachine: 🤦 Ack! I was fiddling with extensions last week and forgot to turn that one off. Fixed.

domoscargin (approved):
Looks grand!

PR #2741, opened 2022-08-02
Remove prototype kit as a dependancy from release docs

owenatgov:
The prototype kit team and design system team are no longer syncing releases so the mentions of them are no longer necessary in our release docs.

- [Current docs](https://github.com/alphagov/govuk-frontend/blob/main/docs/releasing/publishing.md)
- [New docs from this branch](https://github.com/alphagov/govuk-frontend/blob/remove-prototype-kit-from-release-docs/docs/releasing/publishing.md)


querkmachine (approved):

domoscargin (approved):
😢 

This looks good. One small thought is that we might want to keep a line in about letting the Kit team know about releases in the interim, but that might just be unnecessary noise.

owenatgov:
@domoscargin I've added an optional step to let the prototype kit team know, explaining that it's not necessary because they do their own releasing. I anticipate it's the sort of thing we can remove in a years time, depending on the state of our teams by then, but for now it's a reminder that we don't need to be in sync. If you're happy I'll hit merge.

domoscargin:
Go for it!

PR #2742, opened 2022-08-02
Add ability to customise character count fallback text

### LINKED ISSUE #2687: Add ability to customise character count text that announces character/word limit / no-js fallback
## What
The character count has a 'count message' which is visually hidden when JavaScript is available, used by screenreader users. It tells the user the character or word limit for the character count, e.g: "You can enter up to 10 characters". It is visible when JavaScript is disabled, example:

Screenshot 2022-07-12 at 13 15 50

This issue is for adding the ability for users to customise that text. At the moment, it is hardcoded in the component template as "You can enter up to [x] [words/characters]".

## Why
This functionality is needed for users to be able to fully translate the character count text.

## Who needs to work on this
Developers

## Who needs to review this
Developers

## Done when

- [x] Character count macro options allow a user to pass translated text for the count message
- [x] This works whether they're setting maxwords or maxlength
- [x] We're not adding any hardcoded content order (i.e: expecting the 'count' number to appear in a certain place in the string
- [x] There is an example of this in the review app
- [x] There are tests for this

querkmachine:
Adds a new Nunjucks parameter (`fallbackHintText`) which allows the user to customise the fallback message used on the character count in situations where JavaScript is not available. This text is also exposed to assistive technologies, regardless of JavaScript availability. 

It uses [Nunjucks' `replace` filter](https://mozilla.github.io/nunjucks/templating.html#replace) to swap instances of the string `%{count}` with the value of `maxwords`, if defined, or `maxlength`. 

Closes #2687.

## Questions
1. Does `fallbackHintText` make sense as the name for this parameter? It was named `limitHint` in the i18n tech spike, but I changed it here as this piece of text is commonly referred to as being the fallback and it helps avoid confusion with the live limit counter that the JS adds. 
2. Currently only one parameter is made available for customisation. Our English language default switches strings dynamically depending on whether we're counting characters or words, however this would only be possible with localisation if we provided two separate parameters—should we do so? 
3. Currently this doesn't support HTML. Does we want it to? 

> I'm keeping this in draft for the time being as this work doesn't really make sense without the equivalent JavaScript portion (a user is unlikely to customise one and not the other), and we probably want to merge them into main for the same release.

### DIFF
--- a/src/govuk/components/character-count/character-count.yaml
+++ b/src/govuk/components/character-count/character-count.yaml
@@ -40,7 +40,7 @@ params:
 - name: fallbackHintText
   type: string
   required: false
-  description: Text describing the maximum number of characters or words that can be entered, which is announced to screen readers and displayed as a fallback if the character count JavaScript does not run. Instances of `%{count}` within the hint will be replaced by the value of `maxwords` if configured, otherwise it uses `maxlength`. By default, fallback text is provided in English based on what parameters are provided to the component.
+  description: Text describing the maximum number of characters you can enter, which is announced to screen readers. The text is displayed as a fallback if the character count JavaScript does not run. Depending on how you configure this component and what parameters you add, instances of `%{count}` are replaced by the value of `maxwords`. If not configured, it uses `maxlength`. By default, fallback text is provided in English.
 - name: errorMessage
   type: object
   required: false
--- a/src/govuk/components/character-count/character-count.yaml
+++ b/src/govuk/components/character-count/character-count.yaml
@@ -37,6 +37,10 @@ params:
   required: false
   description: Options for the hint component.
   isComponent: true
+- name: fallbackHintText
+  type: string
+  required: false
+  description: Text describing the maximum number of characters or words that can be entered, which is announced to screen readers and displayed as a fallback if the character count JavaScript does not run. Instances of `%{count}` within the hint will be replaced by the value of `maxwords` if configured, otherwise it uses `maxlength`. By default, fallback text is provided in English based on what parameters are provided to the component.
 - name: errorMessage
   type: object
   required: false
@@ -81,6 +85,14 @@ examples:
       maxlength: 10
       label:
         text: Can you provide more detail?
+  
+  - name: with custom fallback text
+    description: with no-js fallback text translated into Welsh
+    data:
+      name: custom-fallback
+      id: custom-fallback
+      maxlength: 10
+      fallbackHintText: Gallwch ddefnyddio hyd at %{count} nod
 
   - name: with hint
     data:
--- a/src/govuk/components/character-count/character-count.test.js
+++ b/src/govuk/components/character-count/character-count.test.js
@@ -26,7 +26,7 @@ describe('Character count', () => {
       await page.setJavaScriptEnabled(true)
     })
 
-    it('shows the static message', async () => {
+    it('shows the fallback message', async () => {
       await goToExample()
       const message = await page.$eval('.govuk-character-count__message', el => el.innerHTML.trim())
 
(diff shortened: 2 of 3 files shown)


vanitabarrett:
My thoughts:

1. I like `fallbackHintText` - compared to `limitHint` it better explains what the text is being used for, whereas in the original spike I had to do some digging to figure out what `limitHint` was.
2. I’m not 100% sure I understand this point. At the point someone is using the component, they’ll have to provide a limit for either characters or words, and therefore the translation string to match that - is that right? So I’m unsure where they’d need to be able to switch dynamically.
3. While this uses the hint component behind the scenes (which does accept HTML), I think this use case is a bit different to normal and it’s unlikely you’d need to pass HTML, like a link. I would lean towards not being able to pass HTML in. I think the way we’ve implemented this means we could easily add `fallbackHintHtml` if we needed to.

src/govuk/components/character-count/character-count.yaml
```diff
@@ -37,6 +37,10 @@ params:
   required: false
   description: Options for the hint component.
   isComponent: true
+- name: fallbackHintText
+  type: string
+  required: false
+  description: Text describing the maximum number of characters or words that can be entered, which is announced to screen readers and displayed as a fallback if the character count JavaScript does not run. Instances of `%{count}` within the hint will be replaced by the value of `maxwords` if configured, otherwise it uses `maxlength`. By default, fallback text is provided in English based on what parameters are provided to the component.
```
vanitabarrett, line 43: Might be worth running this past @claireashworth when she's back - it's quite long!

querkmachine:
@vanitabarrett Regarding number 2: This was me trying to think of a solution to a possible source of confusion: In English, setting the `maxwords` parameter will automatically change the text, but in other languages it won't.

I realise this isn't a big deal, just trying to cover off the bases!

vanitabarrett:
@querkmachine Hm, that's true. That might be ok, because it's the default? Tempted to try it out as-is for now and then consider adding something in if people found it confusing to use.

vanitabarrett (approved):
This change looks good to me - just left one comment about the macro option description. I think it'll need a changelog entry too.

querkmachine:
@vanitabarrett Do you think we should wait until the JavaScript equivalent of this work is closer to completion before merging? It seems weird to release one without the other, even if they are functionally independent.

vanitabarrett:
@querkmachine I don't really have a strong opinion on it. On one side, we already have components where you can end up with mixed languages, whereas on the other side this feels different because it's letting you translate only some of the component 'furniture'. I'm happy for this to sit here until we're more confident in knowing when we're shipping the internationalisation work.

src/govuk/components/character-count/character-count.yaml
claireashworth: I've made the assumption that all this info is needed so it's still long, but the sentences are broken up a bit more. The suggestion is below, but I want to check:
- whether this is specifically a limit on the number of characters that is read out by a screen reader, rather than a general limit that is also read out by screen readers. 
- also, when saying 'if configured', what is the context? Is it something like - 'Depending on how you configure the component and what parameters you add...' ?
Suggested tweaks:
Text describing the maximum number of characters you can enter which will be announced to screen readers. The text is displayed as a fallback if the character count of JavaScript does not run.
Depending on how you configure the component and what parameters you add, instances of `%{count}` are replaced by the value of `maxwords`. If not configured it uses `maxlength`. By default, fallback text is in English.

querkmachine: @claireashworth 
1. This is a limit on how much can be entered into the textarea. The description of the limit is announced by screen readers, or is displayed visually as a fallback if JavaScript is unavailable. 
2. Pretty much. The counter uses characters and `maxlength` by default, and will only switch to words mode if the `maxwords` parameter has been explicitly set.

claireashworth:
```suggestion
  description: Text describing the maximum number of characters you can enter, which is announced to screen readers. The text is displayed as a fallback if the character count JavaScript does not run. Depending on how you configure this component and what parameters you add, instances of `%{count}` are replaced by the value of `maxwords`. If not configured, it uses `maxlength`. By default, fallback text is provided in English.
```

Issue #2721, opened 2022-07-21
Expand existing testing documentation to define what we use unit tests (and other types) for


vanitabarrett:
## What
We currently have integration, end to end, snapshot, template tests and we're proposing adding unit tests too. We should make sure that our existing testing documentation covers what all of these things are and when it's appropriate to use each one.

Yes, definitely! I would love to do this, seems like a good first step before we introduce any tests for our component JS. It might be tricky to articulate in clear terms, but we can have a go. If in doubt, more tests is probably safer too, so as long as we make that clear somewhere we should be ok.

## Why
To help developers in the team and contributors understand what tests they should be writing when they make changes or add new features.

## Who needs to work on this
Developers, Tech Writer

## Who needs to review this
Developers, Tech Writer

## Done when

- [x] Look at how other open source repos deal with this, for inspiration?
- [ ] We have testing docs which explain the different types of testing we do
- [ ] We have some rough guidelines on when it's appropriate to use certain types of testing
- [ ] We might want to say something along the lines of 'more tests are better than none' in case anyone gets confused
- [ ] Follow up card(s) written for adding unit tests for our existing component JS, where appropriate

vanitabarrett:
I've drafted something up for this - going to move it into Blocked for now while Claire is on leave.

PR #2745, opened 2022-08-03
Expand testing docs

vanitabarrett:
Closes https://github.com/alphagov/govuk-frontend/issues/2721

Expand our existing testing documentation to talk about the different types of testing we do and when to use them.
Make reference to unit testing, which we're introducing for JavaScript logic.

### DIFF
--- a/docs/contributing/coding-standards/components.md
+++ b/docs/contributing/coding-standards/components.md
@@ -21,7 +21,7 @@ When creating your component, you should create the following files in the compo
 If your component uses JavaScript, you must also create the following files in the component’s folder:
 
       - `_[component-name].mjs`
-      - `_[component-name].test.js`
+      - `[component-name].test.js`
 
 ## Building your components
 
--- a/docs/contributing/coding-standards/components.md
+++ b/docs/contributing/coding-standards/components.md
@@ -20,7 +20,8 @@ When creating your component, you should create the following files in the compo
 
 If your component uses JavaScript, you must also create the following files in the component’s folder:
 
-      - `_[component-name].mjs`
+      - `[component-name].mjs`
+      - `[component-name].unit.test.mjs`
       - `[component-name].test.js`
 
 ## Building your components
--- a/docs/contributing/testing.md
+++ b/docs/contributing/testing.md
@@ -77,6 +77,8 @@ You should write new tests if you’ve created a new component, or changed the w
 - changing or adding to the component's Nunjucks macro
 - creating or updating a Sass mixin or function
 
+If you're new to testing, see existing test files for examples of things to do. Do not let the tests keep you from submitting your contribution! If you're not sure which tests are needed or are having trouble updating them, submit your pull request anyway. We will help you create the tests and solve problems during code review.
+
 Test files use examples from each component’s `.yaml` file, for example `src/govuk/components/button/button.yaml`. When you add or update tests, you can use the existing examples or add new ones.
 
 Use `hidden: true` in a new example if you do not want to include the example in the review app. The example will still appear in our [test fixtures](http://frontend.design-system.service.gov.uk/testing-your-html/).
--- a/docs/contributing/testing.md
+++ b/docs/contributing/testing.md
@@ -88,6 +88,8 @@ Some test files use examples from each component’s `.yaml` file, for example `
 
 Use `hidden: true` in a new example if you do not want to include the example in the review app. The example will still appear in our [test fixtures](http://frontend.design-system.service.gov.uk/testing-your-html/).
 
+All tests should try and meet [our testing conventions](../releasing/testing-and-linting#conventions)
+
 ### If you created a component
 
 Create the following files in the `src/govuk/components` folder:
--- a/docs/contributing/testing.md
+++ b/docs/contributing/testing.md
@@ -65,11 +65,16 @@ You should test that your contribution works:
 
 ## 5. Run the automated tests
 
+Automated testing helps ensure that the code changes we make do not unintentionally break functionality.
+Tests are automatically run against a branch, pull request, or when the project is built for release, and will notify us if a test has failed.
+
 In your project folder, run `npm test` to run the automated tests, including linting.
 
 If a test fails, you should check your code for any errors, then update any tests you need to.
 
 ## 6. Write new tests
+
+You can [read more about the different types of tests in this project](../releasing/testing-and-linting).
 
 You should write new tests if you’ve created a new component, or changed the way a component works by:
 
--- a/docs/contributing/testing.md
+++ b/docs/contributing/testing.md
@@ -84,7 +84,7 @@ You should write new tests if you’ve created a new component, or changed the w
 
 If you're new to testing, see existing test files for examples of things to do. Do not let the tests keep you from submitting your contribution! If you're not sure which tests are needed or are having trouble updating them, submit your pull request anyway. We will help you create the tests and solve problems during code review.
 
-Test files use examples from each component’s `.yaml` file, for example `src/govuk/components/button/button.yaml`. When you add or update tests, you can use the existing examples or add new ones.
+Some test files use examples from each component’s `.yaml` file, for example `src/govuk/components/button/button.yaml`. When you add or update these tests, you can use the existing examples or add new ones.
 
 Use `hidden: true` in a new example if you do not want to include the example in the review app. The example will still appear in our [test fixtures](http://frontend.design-system.service.gov.uk/testing-your-html/).
 
@@ -93,6 +93,7 @@ Use `hidden: true` in a new example if you do not want to include the example in
 Create the following files in the `src/govuk/components` folder:
 
 - `<COMPONENT>/<COMPONENT>.test.js` - to test functionality if the component uses JavaScript
+- `<COMPONENT>/<COMPONENT>.unit.test.mjs` - to unit test any JavaScript logic
 - `<COMPONENT>/template.test.js` - to test the Nunjucks macro
 
 Where `<COMPONENT>` is the name of the component you created.
@@ -104,6 +105,7 @@ You can use the existing files in the `src/govuk/components` folder as templates
 In the `src/govuk/components` folder, update or add tests to:
 
 - `<COMPONENT>/<COMPONENT>.test.js` - if you updated functionality
+- `<COMPONENT>/<COMPONENT>.unit.test.mjs` - if you updated JavaScript logic
 - `<COMPONENT>/template.test.js` - if you updated the Nunjucks macro
 
 Where `<COMPONENT>` is the name of the component you changed or added to.
--- a/docs/releasing/testing-and-linting.md
+++ b/docs/releasing/testing-and-linting.md
@@ -68,7 +68,11 @@ You should also test component Javascript logic with unit tests, in a `[componen
 We write functional tests for checking our JavaScript exports and our global sass variables - see [all.test.js](../../src/govuk/all.test.js) for examples of global tests we run.
 
 ### Conventions
-We aim to write the description of our tests in as "natural language" as possible, for instance "back-link component fails to render if the required fields are not included".
+We aim to write the test descriptions in everyday language. For example, "back-link component fails to render if the required fields are not included".
+
+Keep all tests separate from each other. It should not matter the order or amount of tests you run from a test suite.
+
+Try and keep assertions small, so each test only checks for one thing. This makes tests more readable and makes it easier to see what's happening if a test is failing.
 
 ## Updating component snapshots
 For components, the snapshots are stored in `[component-name directory]/_snapshots_`.
--- a/docs/releasing/testing-and-linting.md
+++ b/docs/releasing/testing-and-linting.md
@@ -83,7 +83,7 @@ If a snapshot test fails, review the difference in the console. If the change is
 
 This will update the snapshot file. Commit this file separately with a commit message that explains you're updating the snapshot file and an explanation of what caused the change.
 
-## Visual regression testing with Percy
+## Visual regression testing with [Percy](https://percy.io/)
 
 We generate 2 screenshots for each default example of every component. One example has JavaScript enabled, the other has JavaScript disabled. Screenshots are not taken for all the different variations of each component. This tool is not a replacement for manual testing.
 
--- a/docs/releasing/testing-and-linting.md
+++ b/docs/releasing/testing-and-linting.md
@@ -31,11 +31,11 @@ See [Tasks](tasks.md) for details of what `npm test` does.
 ### Running individual tests
 You can run a subset of the test suite that only tests components by running:
 
-    `npm test -- --testPathPattern=src/govuk/components/button`
+    `npx jest src/govuk/components/button`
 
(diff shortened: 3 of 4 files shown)


docs/releasing/testing-and-linting.md
```diff
@@ -2,45 +2,80 @@
...
 
 See [JavaScript Coding Standards](/docs/contributing/coding-standards/js.md#formatting-and-linting) for details.
 
 ## Unit and functional tests with Node.js
 
 We use [Jest](https://jestjs.io/), an automated testing platform with an assertion library, and [Puppeteer](https://pptr.dev/) that is used to control [headless Chrome](https://developers.google.com/web/updates/2017/04/headless-chrome).
 
-We test individual components and for instance global Sass files. See [all.test.js](../../src/govuk/all.test.js). for examples of global tests we run, and  [checkboxes.test.js](../../src/govuk/components/checkboxes/checkboxes.test.js) for an example of component tests.
+### Component tests
+We write tests for every component to check the output of our Nunjucks code. These are found in `template.test.js` files in each component directory. These Nunjucks tests call `render` on the component examples defined in the component yaml files, and asserts the HTML tags, attributes and classes are as expected. For example: checking that when you pass in an `id` to the component using the Nunjucks macro, it outputs the component with an `id` attribute equal to that value.
```
owenatgov, line 61: What type of test would component tests be? Unit or functional? I'm wondering if we should sus this out and jot it up as a conscious choice. I _think_ this is functional as we're doing less logic based tests compared to making sure stuff renders properly.

docs/releasing/testing-and-linting.md
```diff
@@ -2,45 +2,80 @@
...
 
-We aim to write the description of our tests in as "natural language" as possible, for instance "back-link component fails to render if the required fields are not included".
+If a component uses JavaScript, we also write tests in a `[component name].test.js` file, for example [checkboxes.test.js](../../src/govuk/components/checkboxes/checkboxes.test.js). These component tests check that interactions, e.g: a mouse click, have the expected result, e.g: the component gaining focus or the addition of a class.
 
-### Running individual tests
-You can run a subset of the test suite that only tests components by running:
+Component JavaScript logic should also be tested using unit tests, in a `[component name].unit.test.mjs` file.
```
owenatgov, line 65: Once we have a unit test, we should provide an example as we have done on line 63.

owenatgov (commented):
I really like this. Being explicit about how we test stuff at worst can't hurt but more likely is going to help us stay consistent and sensible. I've left a couple of thoughts.

Once this is in a more cohesive state it's worth running this past a tech writer.

docs/releasing/testing-and-linting.md
vanitabarrett: I think functional too. I'll add that in here somewhere...

vanitabarrett:
@claireashworth It would be great to get your thoughts and comments on this when you have time (it's not urgent)! Give me a shout if you want me to run you through it too.

docs/contributing/coding-standards/components.md
```diff
@@ -21,7 +21,8 @@ When creating your component, you should create the following files in the compo
 If your component uses JavaScript, you must also create the following files in the component’s folder:
 
       - `_[component-name].mjs`
-      - `_[component-name].test.js`
+      - `[component-name].unit.test.mjs`
```
claireashworth, line 24: Just checking that the underscores aren't needed before the square brackets here.

docs/contributing/testing.md
```diff
@@ -65,27 +65,36 @@ You should test that your contribution works:
 
 ## 5. Run the automated tests
 
+Automated testing helps ensure that our code is as free of bugs as it can be. It also lets us know immediately if a change we make breaks any other part of the code.
```
claireashworth, line 68: Should we add how it lets us know? Perhaps - It also provides immediate notifications if a change we make...etc
or
It also immediately displays where a change we make...etc

docs/contributing/testing.md
```diff
@@ -65,27 +65,36 @@ You should test that your contribution works:
...
+
 In your project folder, run `npm test` to run the automated tests, including linting.
 
 If a test fails, you should check your code for any errors, then update any tests you need to.
```
claireashworth, line 72: are they only updating the tests that failed?

claireashworth, line 72: Are they only updating the tests that fail?

docs/contributing/testing.md
```diff
@@ -65,27 +65,36 @@ You should test that your contribution works:
...
+You can [read more about the different types of tests in this project](../releasing/testing-and-linting).
+
 You should write new tests if you’ve created a new component, or changed the way a component works by:
 
 - changing or adding to the component's JavaScript code
 - changing or adding to the component's Nunjucks macro
 - creating or updating a Sass mixin or function
 
-Test files use examples from each component’s `.yaml` file, for example `src/govuk/components/button/button.yaml`. When you add or update tests, you can use the existing examples or add new ones.
+If you're new to testing, see existing test files for examples of things to do. Don't let the tests keep you from submitting your contribution! If you're not sure how to do this or are having trouble, submit your pull request anyway. We will help you create the tests and sort out any kind of problem during code review.
```
claireashworth, line 84: Some minor things:
- Change 'Don't' to 'Do not'  
- "If you're not sure how to do this..." Is 'this' how to do the pull request or the testing?  
- Alternative text tweak for the last line - 'We will help you create the tests and solve problems during code review.'

docs/releasing/testing-and-linting.md
```diff
@@ -2,55 +2,88 @@
 
 The CI lints SASS and JavaScript, runs unit and functional tests with Node, and generates snapshots for [visual regression testing](https://www.browserstack.com/percy/visual-testing).
 
-## Running all tests locally
+## Testing terminology
+
+We use different types of tests to check different areas of our code are working as expected.
+
+Unit tests are small, modular tests that verify a "unit" of code. We write unit tests to check our JavaScript logic, particularly 'background logic' that doesn't heavily rely on DOM interaction (where functional tests may be better suited).
```
claireashworth, line 9: Expand 'doesn't' into 'does not'.
Also, does DOM need to be expanded?

docs/releasing/testing-and-linting.md
```diff
@@ -2,55 +2,88 @@
...
+
+Functional tests verify the output of an action and do not check the intermediate states of the system when performing that action. We write functional tests to check component interactions have the expected results, so we also refer to these as component tests. We also write functional tests to check that our Nunjucks code outputs expected HTML, and we tend to refer to these as our Nunjucks tests.
```
claireashworth, line 11: Suggest removing 'tend to'. If we use more than one terminology for the test, we should say under what circumstances we call them Nunjucks tests or [some other tests] under other circumstances.

docs/releasing/testing-and-linting.md
```diff
@@ -2,55 +2,88 @@
...
+### Component tests
+We write functional tests for every component to check the output of our Nunjucks code. These are found in `template.test.js` files in each component directory. These Nunjucks tests call `render` on the component examples defined in the component yaml files, and asserts the HTML tags, attributes and classes are as expected. For example: checking that when you pass in an `id` to the component using the Nunjucks macro, it outputs the component with an `id` attribute equal to that value.
 
-We aim to write the description of our tests in as "natural language" as possible, for instance "back-link component fails to render if the required fields are not included".
+If a component uses JavaScript, we also write functional tests in a `[component name].test.js` file, for example [checkboxes.test.js](../../src/govuk/components/checkboxes/checkboxes.test.js). These component tests check that interactions, e.g: a mouse click, have the expected result, e.g: the component gaining focus or the addition of a class/HTML attribute.
```
claireashworth, line 63: Remove e.g. - These component tests check that interactions, such as a mouse click, have the expected result. 
I don't think the last example (the component gaining focus) is needed.

docs/releasing/testing-and-linting.md
```diff
@@ -2,55 +2,88 @@
...
 
-### Running individual tests
-You can run a subset of the test suite that only tests components by running:
+Component JavaScript logic should also be tested using unit tests, in a `[component name].unit.test.mjs` file. These tests are better suited for testing behind-the-scenes logic or in cases where the final output of some logic isn't a change to the component markup.
```
claireashworth, line 65: Suggest removing the passive voice by changing to - 'You should also test Component Javascript logic with unit tests...etc' 
Also expand 'isn't' to 'is not'. (this is only because the style guide says not to contract negative statements)
Add a comma after 'logic'.

docs/releasing/testing-and-linting.md
```diff
@@ -2,55 +2,88 @@
...
 
-    `npm test -- src/govuk/components/button`
+### Global tests
+We write functional tests for checking our JavaScript exports and our global sass variables - see [all.test.js](../../src/govuk/all.test.js) for examples of global tests we run.
 
-Note: There's a watch mode that keeps a testing session open waiting for changes that can be used with:
+### Conventions
+We aim to write the description of our tests in as "natural language" as possible, for instance "back-link component fails to render if the required fields are not included".
```
claireashworth, line 71: Would changing 'natural language' still work if it was changed to 'plain English'? Something like this - We aim to write the test descriptions in plain English. For example, "back-link...etc

docs/releasing/testing-and-linting.md
```diff
@@ -2,55 +2,88 @@
...
 
-    `npm test -- --watch src/govuk/components/button`
+Each test should be independent and isolated from all other tests. It shouldn't matter what order tests are run in, or how many tests from a test suite are run.
```
claireashworth, line 73: Suggest changing to something like - Keep all tests separate from each other. It should not matter the order or amount of tests you run from a test suite.

docs/releasing/testing-and-linting.md
```diff
@@ -2,55 +2,88 @@
...
+
+Try and keep assertions small, so they only check for one thing. This makes tests more readable and makes it easier to see what's happening if a test is failing.
```
claireashworth, line 75: amend to ...so each test only checks for one thing.

docs/contributing/coding-standards/components.md
vanitabarrett: The underscores aren't needed - we don't use underscores in these filenames. I've actually just realised we don't use them in the `.mjs` files either (the top bullet), so I'll remove that too!

docs/contributing/testing.md
vanitabarrett: I think there probably is a scenario where someone may need to update tests that *didn't* fail. They might spot some tests which should have failed, but didn't - they should also update those to make sure the tests are specific enough that they will catch code changes in future.

docs/releasing/testing-and-linting.md
vanitabarrett: I'm not sure - do you think that has the same meaning? I think we're trying to say that test names should make sense as a sentence when read out loud, which may be slightly different to plain English. Although I'm struggling to think of an example 🤔

vanitabarrett:
Thanks @claireashworth - I think I've made all your suggested changes now, apart from two comments I wasn't sure about, where I've replied to your comments instead. You should be able to see my replies in the comment threads above this message.

docs/contributing/testing.md
claireashworth: Well of we get any feedback that people are not sure what we mean then we can revisit.

docs/contributing/testing.md
```diff
@@ -65,27 +65,37 @@ You should test that your contribution works:
 
 ## 5. Run the automated tests
 
+Automated testing helps ensure that the code changes we make don't unintentionally break functionality.
```
claireashworth, line 68:
```suggestion
Automated testing helps ensure that the code changes we make do not unintentionally break functionality.
```

docs/releasing/testing-and-linting.md
claireashworth: Ah I see. Well 'colloquial' would be an accurate word, but not particularly user friendly. How about 'everyday language' or 'conversational language'?

vanitabarrett: @claireashworth Silly question - how would you word that as a sentence? E.g: "we aim to write the test descriptions in everyday language" - would that be ok?

claireashworth: yup that sounds good

docs/releasing/testing-and-linting.md
```diff
@@ -2,55 +2,88 @@
...
-This will trigger [standard](https://github.com/standard/standard) and [stylelint](https://github.com/stylelint/stylelint) for linting, and [Jest](https://github.com/facebook/jest) for unit and functional tests.
+This will trigger [standard](https://github.com/standard/standard) and [stylelint](https://github.com/stylelint/stylelint) for linting, and [Jest](https://github.com/facebook/jest) for our testing.
 
 See [Tasks](tasks.md) for details of what `npm test` does.
 
-## SASS linting
+### Running individual tests
+You can run a subset of the test suite that only tests components by running:
+
+    `npm test -- --testPathPattern=src/govuk/components/button`
```
36degrees, line 34: Interesting! I didn't know you could do it this way. I use `npx jest src/govuk/components/button` which _might_ be easier to remember 🤔

docs/releasing/testing-and-linting.md
```diff
@@ -2,55 +2,88 @@
...
 
 See [JavaScript Coding Standards](/docs/contributing/coding-standards/js.md#formatting-and-linting) for details.
 
 ## Unit and functional tests with Node.js
 
 We use [Jest](https://jestjs.io/), an automated testing platform with an assertion library, and [Puppeteer](https://pptr.dev/) that is used to control [headless Chrome](https://developers.google.com/web/updates/2017/04/headless-chrome).
 
-We test individual components and for instance global Sass files. See [all.test.js](../../src/govuk/all.test.js). for examples of global tests we run, and  [checkboxes.test.js](../../src/govuk/components/checkboxes/checkboxes.test.js) for an example of component tests.
+### Component tests
+We write functional tests for every component to check the output of our Nunjucks code. These are found in `template.test.js` files in each component directory. These Nunjucks tests call `render` on the component examples defined in the component yaml files, and asserts the HTML tags, attributes and classes are as expected. For example: checking that when you pass in an `id` to the component using the Nunjucks macro, it outputs the component with an `id` attribute equal to that value.
```
36degrees, line 61: Minor suggestion for readability

```suggestion
We write functional tests for every component to check the output of our Nunjucks code. These are found in `template.test.js` files in each component directory. These Nunjucks tests render the component examples defined in the component yaml files, and assert that the HTML tags, attributes and classes are as expected. For example: checking that when you pass in an `id` to the component using the Nunjucks macro, it outputs the component with an `id` attribute equal to that value.
```

36degrees (approved):
This looks great, thanks @vanitabarrett! I have a couple of minor comments but nothing blocking.

docs/releasing/testing-and-linting.md
vanitabarrett: Nice, didn't know about that 👍 I'll update this bit of the docs.

PR #2759, opened 2022-08-09
Provide more details in our publishing docs

owenatgov:
## What
Makes the following changes to our [publishing documentation](https://github.com/alphagov/govuk-frontend/blob/main/docs/releasing/publishing.md):

- Explains that merging a release branch will automatically create a tag in Github
- Separates out instructions on publishing to npm and creating a release on Github
- Explains that the npm publishing process produces a ZIP file which is used to create a Github release

## Why
So that we are clear about what is happening per step in the publishing process so that those carrying out the publishing aren't tripped up by unexpected elements.

## Reviewing
- [Current publishing docs](https://github.com/alphagov/govuk-frontend/blob/main/docs/releasing/publishing.md)
- [Publishing docs after this change](https://github.com/alphagov/govuk-frontend/blob/update-publishing-docs/docs/releasing/publishing.md)

### DIFF
--- a/docs/releasing/publishing.md
+++ b/docs/releasing/publishing.md
@@ -64,7 +64,7 @@ Developers should pair on releases. When remote working, it can be useful to be 
 9. Create a pull request and copy the changelog text.
    When reviewing the PR, check that the version-numbers have been updated and that the compiled assets use this version-number.
 
-10. Once a reviewer approves the pull request, merge it to **main**.
+10. Once a reviewer approves the pull request, merge it to **main**. This will automatically create a tag in Github which you can later use to create a Github release.
 
 ## Publish a release to npm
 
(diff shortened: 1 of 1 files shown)


docs/releasing/publishing.md
```diff
@@ -75,21 +75,25 @@ Developers should pair on releases. When remote working, it can be useful to be
...
-  - set 'GOV.UK Frontend v[version-number]' as the title
-  - add release notes from changelog
-  - attach the generated ZIP that has been generated at the root of this project
-  - publish release
+5. Run `npm logout` to log out from npm.
 
-6. Run `npm logout` to log out from npm.
+## Create a release on Github
+
+You can view the tag created during step 10 of creating the new version in the [Github interface](https://github.com/alphagov/govuk-frontend/releases). To create a new Github release, do the following:
```
querkmachine, line 89:
```suggestion
You can view the tag created during step 10 of creating the new version in the [Github interface](https://github.com/alphagov/govuk-frontend/tags). To create a new Github release, do the following:
```

Link directly to tags page.

docs/releasing/publishing.md
```diff
@@ -75,21 +75,25 @@ Developers should pair on releases. When remote working, it can be useful to be
 3. Run `npm run publish-release`, which will prompt you to check whether the npm tag looks as expected.
 
   If you're following these instructions, you're probably doing a sequential release, meaning
-  the tag should be 'latest'.
+  the tag should be 'latest'. This step will also create a ZIP file containing the entirety of the release in the root
+  of your local instance of govuk-frontend and will be referenced by the command line in its prompt ("...and attach this").
```
querkmachine, line 79: It feels to me like this should be two paragraphs, though I realise having a paragraph about the ZIP generation might be a bit awkward when both the previous and next paragraphs are about npm tags.

querkmachine (changes requested):

docs/releasing/publishing.md
vanitabarrett: I think this should probably go with/after step 4. At this point, pressing `y` simply means accepting that the tag looks correct - it won't publish the release until you press `y` again as prompted to in step 4.

owenatgov: I've moved this into its own paragraph and moved it under what is now step 5.

docs/releasing/publishing.md
```diff
@@ -77,19 +77,25 @@ Developers should pair on releases. When remote working, it can be useful to be
...
   the tag should be 'latest'.
 
-  Enter `y` to continue. If you think the tag should be different, enter `N` to have the option to set your own npm tag.
+4. Enter `y` to continue. If you think the tag should be different, enter `N` to have the option to set your own npm tag.
 
-4. You will now be prompted to continue or cancel the release. Check the details and enter `y` to continue. If something does not look right, press `N` to cancel the release.
+5. You will now be prompted to continue or cancel the release. Check the details and enter `y` to continue. If something does not look right, press `N` to cancel the release.
+  
+  This step will also create a ZIP file containing the entirety of the release in the root of your local instance of 
+  govuk-frontend and will be referenced by the command line ("...and attach this").
```
querkmachine, line 85: This sentence feels a bit cumbersome to me — maybe something more like this?

```suggestion
  This step will create a ZIP file containing the release in the root of your govuk-frontend git directory. You will need this file when creating the GitHub release.
```

Otherwise happy with this 👍

querkmachine (approved):

Issue #2760, opened 2022-08-10
Deprecate smaller font sizes on mobile in the font scale


owenatgov:
## What
Deprecate the following as options for our font mixins (`govuk-font` and `govuk-typography-responsive`) from the typography scale:

- 14 (14px on large screens, 12px on small screens) as an option at all
- 16 dropping down to 14px on small screens
- 19 dropping down to 16px on small screens

## Why
To communicate to service teams that we will be removes these as options in the next breaking release (5.0).

Part of https://github.com/alphagov/govuk-design-system/issues/2289

## Who needs to work on this
@owenatgov 

## Who needs to review this
Developers

owenatgov:
Closing this PR as we're reviewing when and how we deprecate these parts of the font scale as part of a wider release plan for the font scale work.

PR #2761, opened 2022-08-10
Deprecate smaller font sizes in the font scale

owenatgov:
Fixes https://github.com/alphagov/govuk-frontend/issues/2760

Part of https://github.com/alphagov/govuk-design-system/issues/2289

Deprecates the following as options for `$size` in `govuk-typography-responsive` and as possible values for `govuk-!-font-size={value}`:

- 14 as an option at all
- 16 dropping down to 14 on smaller screen sizes
- 19 dropping down to 16 on smaller screen sizes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,19 @@
 
 ## Unreleased
 
+### Deprecated features
+
+#### Using `14` on the font scale and `16` and `19` dropping to a different size on mobile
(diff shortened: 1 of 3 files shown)


owenatgov:
Closing this PR as we're reviewing when and how we deprecate these parts of the font scale as part of a wider release plan for the font scale work.

Issue #2243, opened 2021-06-04
Frontend gulp tasks are complicated and doing too many things


lfdebrux:
## Cause

- There are loads of sequences that build on top of each other.
- There are tasks called both `copy:assets` and `copy-assets`.
- Tasks are sometimes in files with different names
- The `scss:compile` task is responsible for creating both the app CSS and the dist CSS, depending on a flag. The pipelines for these files are similar, but not identical, but completely different files are built depending on the flag.
- The `js:compile` task is responsible for creating both the app JS and the dist JS, depending on a flag. The pipelines for these files are similar, but not identical, but completely different files are built depending on the flag.

## Consequences 

- It's hard to make changes to our build pipeline
- It's hard for contributors to understand how our build pipeline works
- It's easy to run the wrong task because there are many similarly named tasks or ways of running things (e.g. gulp directly vs npm scripts)
- We have a lot of dependencies, which means a greater attack surface, and more maintenance effort is required

## Impact of debt

Medium, it works most of the time, but there is a high risk of breaking things when changes are needed.

## Effort to pay down

High, we don't know enough about Gulp on the team, and we would want to be sure that any changes made don't break the build or lose functionality we need

## Overall rating:

Medium

lfdebrux:
Copied from https://trello.com/c/YCZz7wWY/19-frontend-gulp-tasks-are-complicated-and-doing-too-many-things

lfdebrux:
We'd probably need to pay this down before we can make changes to the build pipeline, i.e. if we want to change how we bundle JavaScript

vanitabarrett:
As discussed in the JavaScript effort/value sessions, one approach might be to tidy up our current Gulp tasks to make them easier to follow and so we have more confidence in making changes to them. However, this won't address the issues with the gulp dependencies we rely on being outdated.

We could also consider switching to an alternative to Gulp. This will require learning something new and moving all our existing tasks to that, and as a team we don't feel we know any other task runners any better than we do Gulp. There may be a steep learning curve, but there may be less maintenance overhead for us if this alternative is better maintained and supported.

vanitabarrett:
Just mentioning this issue here where we attempted to switch `gulp-stylelint` for `stylelint` but it was more complex than we originally though due to needing linting while running `npm start` which watches some of our Sass files. https://github.com/alphagov/govuk-frontend/pull/2512

domoscargin:
[...]

PR #2763, opened 2022-08-12
Refactor asset-version build task to remove Gulp dependency

### LINKED ISSUE #2715: Refactor asset version task to remove Gulp
## What
Update `tasks/gulp/asset-version.js` to replace Gulp with our own asset version update script. This task is responsible for adding the current version to filenames.

## Why
We have [made the decision to move away from Gulp](https://github.com/alphagov/govuk-frontend/issues/2558) and write our own build scripts.

## Who needs to work on this
Developers

## Who needs to review this
Developers

## Done when

- [x] tasks/gulp/asset-version.js no longer depends on Gulp
- [x] Dependencies are as up to date as is feasible
- [x] asset-version.js has been relocated to tasks folder

querkmachine:
Removes the Gulp dependency from the asset-version build task. Resolves #2715.

Specifically, this PR:

- Rewrites the `update-assets-version` Gulp task to be an exported JavaScript function, renamed as `updateDistAssetsVersion`.
- Moves `asset-version.js` to the `tasks` directory.
- Removes dependency on Gulp and Gulp plugins from the task.
- Removes use of the `destination` variable from `task-arguments.js` to determine where compiled files go, as per [this comment](https://github.com/alphagov/govuk-frontend/issues/2243#issuecomment-1214774873). This task only ever seems to compile to `dist`, so this has been hardcoded. 
- Removes the `vinyl-paths` package, as this is no longer used anywhere. 
- Modifies `gulpfile.js` to integrate the renamed task.

Related changes: 

- Moves `task-arguments.js` to the `tasks` directory.
- Modifies the paths in some unrelated task files to account for moving `task-arguments.js`.

These changes appear to produce the expected compiled files, filenames, and passes our automated tests. 

Note: The `updateDistAssetsVersion` function is required to accept and return a callback function so that it can be integrated with the existing Gulp tasks. This could be removed once we've gotten rid of Gulp entirely.


tasks/gulp/asset-version.js
```diff
@@ -1,35 +1,49 @@
...
-const rename = require('gulp-rename')
-const del = require('del')
-const vinylPaths = require('vinyl-paths')
 
 // check for the flag passed by the task
-
 const isDist = taskArguments.destination === 'dist' || false
 
 // Update assets' versions ----------
 // Add all.package.json version
```
domoscargin: Is this still accurate? We're writing to `VERSION.txt`

Might also be nice to be more accurate with the description here, something like "append new asset version to file names".

tasks/gulp/asset-version.js
```diff
@@ -1,35 +1,49 @@
...
-      })
-    ))
-    .pipe(gulp.dest(taskArguments.destination + '/'))
-})
+  ]
+
+  // Loop through files and move them to their new locations
+  assetFiles.map(srcFilename => {
+    let destFilename = srcFilename
+    if (isDist) {
```
domoscargin: Does this function ever run outside of `dist`?

Wondering if we should get really specific and call this something like `updateDistAssetsVersion` and get rid of this `if` clause. We could probably then drop the need to copy the files to the same location.

domoscargin (commented):
Looking good. This seems to produce the same dist folder and files for me.

A coupla comments with some _reckons_ but very excited to be leaving Gulp behind!

tasks/gulp/asset-version.js
querkmachine: I think it's accurate, just poorly explained? 

I interpreted it as it adding the version *from* `package.json`. We should probably rewrite it so that it's obvious, though.

tasks/gulp/asset-version.js
querkmachine: It doesn't *seem* like this task is used for anything else in our own build:dist task, and that would seem to be supported by it not really doing anything but write `VERSION.txt` if the if-statement fails.

Changing the destination folder is an option we provide through the command line though, so I'm not 100% confident that the option is never used.

tasks/gulp/asset-version.js
domoscargin: Ah, right - I read it as "Add the version to package.json" at a glance 🤦 , so count me as the reason for spelling it out!

tasks/gulp/asset-version.js
domoscargin: Yeah, as far as I can see, we just use that command line option to separate the package and dist builds.

I think we should simplify our tasks and get rid of the `destination` option, but maybe that's a step for after we've updated most of these.

From what I can see:

- The only task in common for `build:package` and `build:dist` is `clean`, and that does different things depending on which destination is set, so it probably makes sense just to make them two separate tasks.
- The `copy-to-destination` file uses `destination`, but its `copy-files` task is only used in `build:package`
- The `compile-assets` file uses `destination`, but again, that whole file is a bit of a mess and we could simplify it.
- `copy:assets` is, supposedly, supposed to be used to copy to `public`, but it's actually only used for the `dist` build task
- The only times we run a destination-reliant thing using the default of `public` are in the `heroku` and `test` scripts, so it feels like we'd probably be able to simplify that as well.

querkmachine: I've removed the destination if-statement. 👍

domoscargin (approved):
🌲 Breathe the sweet air of this post-gulp world! 🌳 

Nice work!

PR #2766, opened 2022-08-15
Remove unused `console.log` calls from accordion JavaScript

36degrees:
Because of the way the if statements are written, these attempts to call `console.log` and `console.error` will only ever run if either `console` or `console.log` is undefined (and therfore _not_ a function). They will never successfully write anything to the console log.

Given that they’ve been like this for several years (since 027373d), we can safely assume they’re not doing anything useful, and can remove them.


owenatgov (approved):
Does the job 👍🏻 

Whilst the impact to the user will be minor, this could also do with a quick `fixes` changelog entry.

I've also been pedantic and added this to the [NEXT] milestone.

PR #2767, opened 2022-08-15
Hotfix Sassdoc not being able to build for Heroku

querkmachine:
Hotfixes the Heroku script failing as Sassdoc cannot be built. This is because Sassdoc is being called through Gulp, but the Sassdoc build script no longer uses it as of #2764.

Fixes this by creating a single-use Gulp task that calls the Sassdoc function. 

In the long term, we will need to change the Heroku script to call Sassdoc through some other method.


domoscargin (approved):
Works for me locally!

PR #2773, opened 2022-08-15
Refactor clean task to remove Gulp dependency

### LINKED ISSUE #2714: Refactor clean task to remove Gulp
## What
Update `tasks/gulp/clean.js` to replace Gulp with our own clean script. This task is responsible for watching cleaning a destination directory.

## Why
We have [made the decision to move away from Gulp](https://github.com/alphagov/govuk-frontend/issues/2558) and write our own build scripts.

## Who needs to work on this
Developers

## Who needs to review this
Developers

## Done when

- [ ] tasks/gulp/clean.js no longer depends on Gulp
- [ ] tasks/gulp/clean.js no longer depends on `taskArguments.destination`
- [ ] `del` package is as up to date as is feasible or this task does not rely on it
- [ ] docs/contributing/tasks.md has been updated
- [ ] clean.js has been relocated to tasks folder

domoscargin:
- Rewrites the `clean` Gulp task to be 3 exported JavaScript functions: `cleanPublic`, `cleanDist` and `cleanPackage`
  - Currently `cleanPublic` and `cleanDist` contain the same code, but I think it's worth the extra clarity to duplicate them in this instance, and there's a possibility they diverge in the future.
  - This removes the dependency on task arguments.
- Moves clean.js to the tasks directory
- Removes dependency on Gulp from the task
- Modifies `gulpfile.js` to integrate the renamed task
- Bumps `del` to 6.1.1 to remove a [race condition bug that could cause errors unless using `del.sync`](https://github.com/sindresorhus/del/issues/68). `del` 7.0.0+ is ESM only - we'll get there eventually!
- Updates `doc/contributing/tasks.md` with clean tasks

Fixes #2714 

## Note
I considered removing `del` and using `fs` instead, but I don't think that's in scope, since `del` is a nice, safe, cross-platform way to handle this and we'd need to consider the pros and cons of tossing it.

### DIFF
--- a/docs/contributing/tasks.md
+++ b/docs/contributing/tasks.md
@@ -29,12 +29,14 @@ NPM scripts are defined in `package.json`. These trigger a number of gulp tasks.
 - starts up Express
 
 **`npm run build:package` will do the following:**
+- clean the `package` folder
 - compile component nunjucks to HTML
 - copy template, macro and component.njk files for each component
 - copy Sass files, add vendor prefixes and replace path to be node_modules consumption compliant
 - runs `npm run test:build:package` (which will test the output is correct)
 
 **`npm run build:dist` will do the following:**
+- clean the `dist` folder
 - copy JS
 - copy icons
 - copy SASS and add vendor prefixes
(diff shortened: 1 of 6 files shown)


querkmachine (approved):
Looks good and all works locally 👍

PR #2778, opened 2022-08-16
fix(): Replace `console.log` with `info()` and `error`

abhi-markan:
## Introduction

`console.log()` should only be used for development purpose (ideally on local dev environment). This PR aims to replace `.log()` calls with `.info()` and `error()` where appropriate. This also ensures any linting rules which prohibits usage of `.log()` calls are also compatible.

* Packages have also been updated here (purely for flagging purpose).
* `typeof console.error` have been used when an error is being thrown using `console.error()`.

### DIFF
--- a/app/start.js
+++ b/app/start.js
@@ -4,5 +4,5 @@ const PORT = process.env.PORT || configPaths.ports.app
 const app = require('./app.js')()
 
 app.listen(PORT, () => {
-  console.log('Server started at http://localhost:' + PORT)
+  console.info('Server started at http://localhost:' + PORT)
 })
--- a/bin/check-nvmrc.js
+++ b/bin/check-nvmrc.js
@@ -20,7 +20,7 @@ fs.readFile(path.join(__dirname, '../.nvmrc'), 'utf8', function (error, data) {
   var nvmInstallText = 'To do this you can install nvm (https://github.com/creationix/nvm) then run `nvm install`.'
 
   if (versionMatchesMajor) {
-    console.log('' +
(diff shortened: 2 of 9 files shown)


NickColley:
In a recent pull request most of these console.logs were removed so this might mean this work is no longer needed as the remaining logs are not part of the published code for end users.

https://github.com/alphagov/govuk-frontend/pull/2766

owenatgov:
Hi @abhi-markan. Thank you for the contribution but @NickColley hit the nail on the head that this change is now no longer necessary.

To expand further, you identified a couple of places where changes could be made however not all of them are appropriate:

- We don't edit the `/package` directly in normal pull requests as said directory is used as a source for creating releases. We only update package with the latest changes from main when we run a release.
- Many of the places you edited such as `/app/start.js` and `/tasks/sassdoc.js` are our own build scripts and will never be run in a browser, therefore `console.log` is acceptable.

The only place that could've been updated was in the accordion component's js and, as Nick has already mentioned, we recently added a change to remove this js as we discovered it was incorrectly configured and wasn't adding anything to the user's experience. With that in mind, I'm going to close this PR as it looks like your concerns about misusing `console.log` are resolved as we aren't using `console.log` in any user-facing js as of 4.3.1.

I'm sorry that your work has been undone by pure coincidence that we removed the offending code just before you made this change. We appreciate you putting the time and effort into do this and hope you will consider contributing again in the future.

abhi-markan:
Thank you for a detailed explanation!! 💯

PR #2800, opened 2022-08-19
Improve Pagination component print styles

NickColley:
By default `.govuk-link` print styles will output the `href` attribute
in brackets after the link. For example: "Home (/homepage)".

This is done with the `::after` pseudo selector.
So when this was styled to increase the touch area of the pagination link
it resulted in the bracketed information overlapping the component in
a difficult to read way.

By only applying this touch area improvement in the `screen` media context
we avoid overwriting this inherited print style and allow the default
`.govuk-link` behaviour to apply as usual.

In practice I think many GOV.UK Frontend users might hide this component from their print views so impact on end-users is low.

I'd be down for doing something more holistic as part of https://github.com/alphagov/govuk-frontend/issues/573 if you're interested in that contribution (including thinking about it from a more high level design pov), but I noticed this just at a quick glance.

## Screenshots

| Before | After |
|-|-|
| Print mode shows overlapping links | Print mode has pagination links are shown inline without overlap |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,12 @@ This will enable screen reader users to have a better, more coherent experience 
 
 This was added in [pull request #2677: Amend error summary markup to fix page load focus bug in JAWS 2022](https://github.com/alphagov/govuk-frontend/pull/2677)
 
+### Fixes
+
+We’ve made fixes to GOV.UK Frontend in the following pull requests:
+
+- [#2800: Improve Pagination component print styles](https://github.com/alphagov/govuk-frontend/pull/2800)
+
 ## 4.3.1 (Patch release)
 
 ### Recommended changes
--- a/src/govuk/components/pagination/_index.scss
+++ b/src/govuk/components/pagination/_index.scss
@@ -106,13 +106,16 @@
     display: block;
     min-width: govuk-spacing(3);
 
-    &:after {
-      content: "";
-      position: absolute;
-      top: 0;
-      right: 0;
-      bottom: 0;
-      left: 0;
+    // Increase the touch area for the link to the parent element.
+    @media screen {
+      &:after {
+        content: "";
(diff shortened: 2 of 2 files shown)


owenatgov:
Hi @NickColley! Finally getting into this.

Very much appreciate the bugfix, deffo something we should've spotted during development 😅 However I'm going to be irritating and say this has prompted me to wonder if we should even show pagination on print and I wonder if it makes more sense to amend this PR to just hide it outright. What are your thoughts? If we wanted to just hide the whole thing, do you have capacity to amend this PR?

NickColley:
@owenatgov my high level thought at this point is this:

1. Show all components with print friendly styles by default
2. Allow the user to then decide which components should be hidden themself.
Perhaps an opt in default for the second option. I'd need to spend more time thinking about it.

That's what I was thinking by the extended piece of work. But what I've shared here is sort of how people are doing it in practice if you look at GOV.UK Publishing for example.

NickColley:
Maybe something like:

1. All components have functionally working print styles
2. By default we have some best practice defaults for components we think are most likely to be hidden
3. User can override those per-component basis
5. Write some guidance to explain all of this alongside Sass docs

NickColley:
@querkmachine that reminded me of something Frankie said about how they intentionally made accordions keep the state they are on the page when printed with the idea that the user might want to control that themselves.

owenatgov:
@NickColley I think this makes sense. Let's get this in for now to clear it up and start thinking about a more consistent print strategy soon. I'll review this next.

owenatgov (approved):
📄

domoscargin (approved):
Yup, this looks good to me, and agree we should have a think about print styles later.

NickColley:
Thanks for your time reviewing this, I'll leave it with you to merge as I have no power to merge I am a humble contributor :P

PR #2806, opened 2022-08-22
fix(code-quality): Variable conflict & typing error rectification

abhi-markan:
## Introduction
`compile-assets.js` is in a need of a code review and have noticed discrepancies with regards to the code quality.
This PR aims to triage those quick wins.

## Updates
* Variable scope conflict: `path` variable has already been defined at the outermost scope, which could be conflicting with an inner-scope `path` variable. This has been renamed to `location`.
* `var` keyword has been updated with `let`, as it suffice block scope usage thus making `var` usage redundant.
*  `desinationPath` has been corrected to `destinationPath`.

### DIFF
--- a/tasks/gulp/compile-assets.js
+++ b/tasks/gulp/compile-assets.js
@@ -163,9 +163,9 @@ function compileFullPageStyles (done) {
   gulp.src(compileFullPageExampleStylesheets)
     .pipe(plumber(errorHandler))
     .pipe(sass())
-    .pipe(rename(function (path) {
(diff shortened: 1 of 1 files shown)


domoscargin (approved):
Hi @abhi-markan,

Thanks for this. We're in the process of updating our tasks to move away from Gulp, so this file will be overhauled in the near future, but I think these changes are worth making now anyway.

Just need another approval from the team before merging.

querkmachine (approved):
LGTM 👍

Issue #2692, opened 2022-07-14
Add full page example(s) for translated components (via macro options) in the review app


vanitabarrett:
## What
Add a new [full page example](http://govuk-frontend-review.herokuapp.com/full-page-examples) in the review app, which shows components which have been passed translated content via their macro options. This card does not cover adding examples for translating content in JavaScript.

## Why
We want to see all translated components in one place, so we can see how they interact together, view them in a more realistic example, and spot any issues. 

## Who needs to work on this
Developers

## Who needs to review this
Developers

## Done when

- [x] Realistic example(s) chosen (could be a translated version of an existing full page example)
- [x] Full page example added

PR #2809, opened 2022-08-24
Add new example page of all translated components to review app

vanitabarrett:
**Note: this relies on @querkmachine 's change to [allow translation of the character count fallback message](https://github.com/alphagov/govuk-frontend/pull/2742). Please only review the second commit.**

Closes https://github.com/alphagov/govuk-frontend/issues/2692

## What
Adds a new example page containing all GOV.UK Frontend components, translated into Welsh. This includes translation of text in Nunjucks/HTML, not strings in component JavaScript code as we haven't begun our implementation to allow that yet.

This should also include translation of any HTML attributes, e.g: aria labels.

Note: the card was originally written with the end goal of having a full page realistic example of translated components, but it didn't feel possible to have a realistic page with all components in, so I've instead opted for just a "list" of all components on a page instead.

## Why

- So we can view examples of all translated components in one place.
- We can spot components which may be missing abilities to translate.
- If we're looking to make design changes in future and we want to check how it affects components in a different language.

➡️ Preview link: https://govuk-frontend-pr-2809.herokuapp.com/examples/translated

### DIFF
--- a/app/views/examples/translated/index.njk
+++ b/app/views/examples/translated/index.njk
@@ -1,0 +1,872 @@
+{% from "accordion/macro.njk" import govukAccordion %}
+{% from "back-link/macro.njk" import govukBackLink %}
+{% from "breadcrumbs/macro.njk" import govukBreadcrumbs %}
+{% from "button/macro.njk" import govukButton %}
+{% from "character-count/macro.njk" import govukCharacterCount %}
+{% from "checkboxes/macro.njk" import govukCheckboxes %}
+{% from "cookie-banner/macro.njk" import govukCookieBanner %}
+{% from "date-input/macro.njk" import govukDateInput %}
+{% from "details/macro.njk" import govukDetails %}
+{% from "error-message/macro.njk" import govukErrorMessage %}
+{% from "error-summary/macro.njk" import govukErrorSummary %}
+{% from "file-upload/macro.njk" import govukFileUpload %}
+{% from "footer/macro.njk" import govukFooter %}
+{% from "header/macro.njk" import govukHeader %}
+{% from "inset-text/macro.njk" import govukInsetText %}
+{% from "notification-banner/macro.njk" import govukNotificationBanner %}
+{% from "pagination/macro.njk" import govukPagination %}
+{% from "panel/macro.njk" import govukPanel %}
+{% from "phase-banner/macro.njk" import govukPhaseBanner %}
+{% from "radios/macro.njk" import govukRadios %}
+{% from "select/macro.njk" import govukSelect %}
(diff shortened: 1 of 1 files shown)


app/views/examples/translated/index.njk
```diff
@@ -0,0 +1,871 @@
...
+          href: "#",
+          text: "Cysylltwch"
+        },
+        {
+          href: "#",
+          text: "Telerau ac Amodau"
+        }
+      ],
+      html: 'Adeiladwyd gan y <a href="#" class="govuk-footer__link">Government Digital Service</a>'
+    },
```
36degrees, line 326: I think you also want to set `meta.visuallyHiddenTitle` here to change the hidden H2 from it's default 'Support Links'

https://github.com/alphagov/govuk-frontend/blob/aafbb16079a53730ba9a8a0e2cc0aed647237fe8/src/govuk/components/footer/template.njk#L35

app/views/examples/translated/index.njk
```diff
@@ -0,0 +1,871 @@
...
+      html: "Mae’r holl gynnwys ar gael o dan y <a href='#' class='govuk-footer__link'>Drwydded Llywodraeth Agored v3.0</a>, ac eithrio lle nodir yn wahanol"
+    },
+    copyright: {
+      html: "<span>Hawlfraint y Goron</span>"
+    }
+  }) }}
+
+  {{ govukHeader({
+    homepageUrl: "#",
+    serviceName: "Service name",
```
36degrees, line 337: Do we want to translate this too?

app/views/examples/translated/index.njk
```diff
@@ -0,0 +1,871 @@
...
+  }) }}
+
+  {{ govukPanel({
+    titleText: "Cais wedi'i gwblhau",
+    html: "Eich rhif cyfeirnod<br><strong>HDJ2123F</strong>"
+  }) }}
+
+  {{ govukPhaseBanner({
+    tag: {
+      text: "alpha"
```
36degrees, line 414: Do we want to translate this?

vanitabarrett: Unsure about this one, as they're like official "statuses" for a service. Examples I've found don't tend to translate it, e.g:

https://www.manage-student-loan-balance.service.gov.uk/ors?locale=cy
https://www.passport.service.gov.uk/filter/overseas?lang=cy&_ga=2.191870766.509248048.1661854651-1299619278.1657288777

36degrees: Yep, fair enough. I think we already have an example of a translated tag elsewhere on the page anyway.

querkmachine (approved):
Looks good here 👍

PR #2811, opened 2022-08-26
Use Element.id to get module id for accordion

36degrees:
Use the native `Element.id` IDL property ([supported by all modern browsers including IE6+](https://caniuse.com/mdn-api_element_id)) rather than `Element.getAttribute('id')`

As `this.moduleId` is only used once, replace it with a call to `this.$module.id`.


querkmachine (approved):

36degrees:
Just to record that I've tested this in IE8 through 11 (by inspecting the button and ensuring that `aria-controls` is set as expected).

I've not tested in any other browsers other than Chrome 104 as my day-to-day browser.

PR #2807, opened 2022-08-24
Tidy up and refactor the Character Count JavaScript

36degrees:
Tidying this up with the aim of making it easier to set config (e.g. maxwords, threshold) by passing a config object at initialisation in the future.

Suggest reviewing commit by commit as quite a lot of the code has been moved around.

PR #2812, opened 2022-08-26
Fix incorrect type in JSDoc

36degrees:
JavaScript doesn't have an Integer type, only Numbers.

I don't think this needs a changelog entry as it's a tweak to changes introduced in a PR that's already listed in the changelog (#2807)?


lfdebrux (approved):
Makes sense to me, do we want to specify in the description though that it should be an integer?

36degrees:
I guess it's implied in that you can't count a fractional number of characters or words? 🤔

PR #2814, opened 2022-08-26
Add initial i18n lookup function

### LINKED ISSUE #2801: Add internationalisation logic to lookup translation from key
## What
Write the JavaScript logic needed for the most minimal internationalisation use case, i.e: replacing a default English string with a new string in a different language. The logic will need to be able to:

- take a translations object
- take a translation ‘key’ and look up the relevant string in the translations object

This should not cover more complex translations, e.g: pluralisation or string interpolation.

## Why
So we can begin to allow hardcoded strings in our component JS to be translated
So we have a solution for translating hardcoded strings in future component JS

## Who needs to work on this
Developers

## Who needs to review this
Developers

## Done when

- [ ] Internationalisation logic written that takes a JavaScript object of translations
- [ ] Ability to lookup a key within that translations object, and return a string
- [ ] Nicely handles instances where the key cannot be found in the translations object
- [ ] Unit tests written for the above

querkmachine:
Adds the most basic version of the i18n function used to lookup and return translation strings, and implements it for the Accordion component's default strings. 

This doesn't touch the Character Count component, as this requires placeholder and pluralisation support, which is being added later.

Will close #2801 upon completion.

### DIFF
--- a/src/govuk/components/accordion/accordion.mjs
+++ b/src/govuk/components/accordion/accordion.mjs
@@ -16,6 +16,7 @@
 */
 
 import { nodeListForEach } from '../../common.mjs'
+import I18n from '../../i18n.mjs'
 import '../../vendor/polyfills/Function/prototype/bind.mjs'
 import '../../vendor/polyfills/Element/prototype/classList.mjs'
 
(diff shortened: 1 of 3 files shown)


vanitabarrett (approved):
🎉 👏

PR #2815, opened 2022-08-30
Use empty strings for visually hidden text

### LINKED ISSUE #2803: summary-list.yaml contains invalid value for property visuallyHiddenText
## Description of the issue

PlatUI has recently upgraded our play-frontend-hmrc service with the new govuk-frontend version(**v4.3.1**)
after the upgrade, we had a failing view test around the **summary-list** component.

The problem was that the **visuallyHiddenText** [property](https://github.com/alphagov/govuk-frontend/blob/9fc06138fca9bdb16350910360234441c79d3c10/src/govuk/components/summary-list/summary-list.yaml#L55) which is a string is being set with a [boolean](https://github.com/alphagov/govuk-frontend/blob/9fc06138fca9bdb16350910360234441c79d3c10/src/govuk/components/summary-list/summary-list.yaml#L159) value.

- GOV.UK Frontend Version: v4.3.1

36degrees:
The visuallyHiddenText option is expected to be a string (according to the params definition) but we're setting it to `false` in cases where we don't want to include visually hidden text.

Set it to an empty string instead to avoid using a boolean where we should be using a string.

Fixes #2803.

### DIFF
--- a/src/govuk/components/summary-list/summary-list.yaml
+++ b/src/govuk/components/summary-list/summary-list.yaml
@@ -156,10 +156,10 @@ examples:
             items:
               - href: '#'
                 html: Golygu<span class="govuk-visually-hidden"> enw</span>
-                visuallyHiddenText: false
+                visuallyHiddenText: ''
               - href: '#'
                 html: Dileu<span class="govuk-visually-hidden"> enw</span>
-                visuallyHiddenText: false
+                visuallyHiddenText: ''
         - key:
             text: Dyddiad geni
           value:
@@ -168,7 +168,7 @@ examples:
             items:
               - href: '#'
                 html: Golygu<span class="govuk-visually-hidden"> dyddiad geni</span>
-                visuallyHiddenText: false
+                visuallyHiddenText: ''
         - key:
             text: Gwybodaeth cyswllt
           value:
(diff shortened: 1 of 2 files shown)


src/govuk/components/summary-list/summary-list.yaml
```diff
@@ -156,10 +156,10 @@ examples:
             items:
               - href: '#'
                 html: Golygu<span class="govuk-visually-hidden"> enw</span>
-                visuallyHiddenText: false
+                visuallyHiddenText:
```
36degrees, line 159: Unlike the error message component, there's no default for the visually hidden text for each action:

https://github.com/alphagov/govuk-frontend/blob/dc6fc67d4dca1574194b8337ff0e3cfa07b14dd1/src/govuk/components/summary-list/template.njk#L4-L6

So we _could_ just omit this option entirely, and it'd have the same effect.

vanitabarrett:
The PR/commit message suggests we're setting these to empty strings instead, but it looks like leaving a field blank in yaml is a null value (as opposed to `''` which is an empty string)?

36degrees:
Good catch!

I did intend for it to be an empty string, but I think using `null` is probably fine too? 🤔 Would just need to update the commit message and PR description if so.

querkmachine:
I'm personally indifferent as to whether we use `null` or empty strings, as it seems both would be acceptable from a validation point of view, but if we do opt to use empty strings, should we be more explicit about them?

To me, just having `visuallyHiddenText:` seems like it's an authoring mistake, rather than a purposeful use of an empty string. `visuallyHiddenText: ''` would be clearer, I think.

owenatgov (approved):
Looks sharp as a tack to me 👍🏻

PR #2817, opened 2022-08-30
Avoid copying tests ending in `.mjs` to package folder

36degrees:
Without this change e.g. `src/govuk/i18n.unit.test.mjs` is copied to `package/govuk-esm/i18n.unit.test.mjs`, as it ends in `.test.mjs` rather than `.test.js`.


vanitabarrett (approved):
Good spot, thanks! 👍

PR #2818, opened 2022-08-30
Add support for localisation via data-* attributes to Accordion component

### LINKED ISSUE #2699: Add ability to pass translation strings to accordion component via data-attributes
## What
Adds ability to pass translation strings into the accordion component, using the Nunjucks macro. 
Needs to work with the i18n logic added in https://github.com/alphagov/govuk-frontend/issues/2801. There will need to be some kind of logic to look up the data translation attributes set on a component and return them as an object

## Why
So users can translate the hardcoded strings in our accordion JS component

## Who needs to work on this
Developers

## Who needs to review this
Developers

## Done when

- [x] The accordion component has new Nunjucks macro options for passing in translation strings (following the format decided on in https://github.com/alphagov/govuk-frontend/issues/2780)
- [x] Implemented JS code to lookup translations set using data attributes
- [x] Translations provided via data attributes are in a format that the internationalisation logic understands, i.e: a JS object
- [x] The accordion component JS uses the internationalisation logic to swap out hardcoded strings
- [x] The accordion component still works if no/only some translations are passed (falls back to defaults)
- [x] The accordion component has tests which cover the new ability to translate
- [x] Added to or created a full page example which shows full translation of the accordion component

querkmachine:
Raising early as a draft PR for visibility and those sweet, sweet automated deployments. Will resolve #2699 on completion. 

- Adds the configuration merging function and modified `Element.prototype.dataset` polyfill from the config spike (#2810). 
- Adds a new helper function, `extractConfigByNamespace`, which non-destructively filters a flattened configuration object by a given namespace (e.g. retrieving all key-value pairs where the key starts with 'i18n.') 
- Adds Nunjucks parameters for setting i18n related data-* attributes on the Accordion component and associated documentation.
- Adds support for retrieving and merging configuration objects to the Accordion JS.
- Adds an example of a localised Accordion component using data-* attributes. 
- Adds unit tests for the configuration merging and value extraction functions.
- Adds integration tests for the Accordion component, checking that changes to the data-* attributes are reflected in the UI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,21 @@ This text is announced by screen readers when the character count input is focus
 
 This was added in [pull request #2742: Add ability to customise character count fallback text](https://github.com/alphagov/govuk-frontend/pull/2742).
 
+#### Localise the accordion's toggle buttons
+
+You can now translate the text of the [accordion](https://design-system.service.gov.uk/components/character-count/) component's show and hide toggle buttons.
+
+When using the Nunjucks macro, you can use the new `showSectionHtml` and `hideSectionHtml` parameters to customise the text of the 'show' and 'hide' toggles in each section. You can also use `showAllSectionsHtml` and `hideAllSectionsHtml` parameters to customise the text of the toggle at the top of the accordion.
+
+If you're not using the Nunjucks macro, you can customise these using data-* attributes. Any HTML appearing within the attributes must have quotation marks and brackets converted into their HTML entity equivalents. 
+
+- `data-i18n.show-section`
+- `data-i18n.hide-section`
+- `data-i18n.show-all-sections`
+- `data-i18n.hide-all-sections`
+
+This was added in [pull request #2818: Add support for localisation via data-* attributes to Accordion component](https://github.com/alphagov/govuk-frontend/pull/2818).
+
 ### Recommended changes
 
 #### Remove `aria-labelledby`, remove `id="error-summary-title"` from title and move `role="alert"` to child container on the error summary component
@@ -50,7 +65,7 @@ If you're not using the Nunjucks macros, you can improve the experience for scre
 
 This will enable screen reader users to have a better, more coherent experience with the error summary. Most notably it will ensure that users of JAWS 2022 or later will hear the entire contents of the error summary on page load and therefore have further context on why there is an error on the page they're on.
 
-This was added in [pull request #2677: Amend error summary markup to fix page load focus bug in JAWS 2022](https://github.com/alphagov/govuk-frontend/pull/2677)
+This was added in [pull request #2677: Amend error summary markup to fix page load focus bug in JAWS 2022](https://github.com/alphagov/govuk-frontend/pull/2677).
 
 ### Fixes
 


src/govuk/components/accordion/accordion.yaml
```diff
@@ -169,6 +185,29 @@ examples:
...
           html: <a class="govuk-link" href="#">Link B</a>
 
+- name: with translations
+  data:
+    id: with-translations
+    hideAllSectionsHtml: Collapse all sections
+    showAllSectionsHtml: Expand all sections
+    hideSectionHtml: |
+      Collapse <span class="govuk-visually-hidden">this section</span>
+    showSectionHtml: |
```
domoscargin, line 195:
```suggestion
    showSectionHtml: |-
```

Incredibly minor thing, but the HTML for the translated accordion example ends up with endquotes on newlines:

```html
<div
  class="govuk-accordion"
  data-module="govuk-accordion"
  id="with-translations" data-i18n.hide-all-sections="Collapse all sections" data-i18n.hide-section="Collapse &lt;span class=&quot;govuk-visually-hidden&quot;&gt;this section&lt;/span&gt;
" data-i18n.show-all-sections="Expand all sections" data-i18n.show-section="Expand &lt;span class=&quot;govuk-visually-hidden&quot;&gt;this section&lt;/span&gt;
">
```

src/govuk/components/accordion/template.njk
```diff
@@ -1,8 +1,15 @@
...
-<div class="govuk-accordion {%- if params.classes %} {{ params.classes }}{% endif -%}" data-module="govuk-accordion" id="{{ id }}"
-{%- for attribute, value in params.attributes %} {{attribute}}="{{value}}"{% endfor %}>
+<div
+  class="govuk-accordion {%- if params.classes %} {{ params.classes }}{% endif -%}"
+  data-module="govuk-accordion"
+  id="{{ id }}"
+  {%- if params.hideAllSectionsHtml %} data-i18n.hide-all-sections="{{ params.hideAllSectionsHtml | escape }}"{% endif %}
+  {%- if params.hideSectionHtml %} data-i18n.hide-section="{{ params.hideSectionHtml | escape }}"{% endif %}
+  {%- if params.showAllSectionsHtml %} data-i18n.show-all-sections="{{ params.showAllSectionsHtml | escape }}"{% endif %}
+  {%- if params.showSectionHtml %} data-i18n.show-section="{{ params.showSectionHtml | escape }}"{% endif %}
```
36degrees, line 11: Can we add tests for these to template.test.js?

src/govuk/components/accordion/template.njk
```diff
@@ -1,8 +1,15 @@
 {% set id = params.id %}
 {% set headingLevel = params.headingLevel if params.headingLevel else 2 %}
 
-<div class="govuk-accordion {%- if params.classes %} {{ params.classes }}{% endif -%}" data-module="govuk-accordion" id="{{ id }}"
-{%- for attribute, value in params.attributes %} {{attribute}}="{{value}}"{% endfor %}>
+<div
+  class="govuk-accordion {%- if params.classes %} {{ params.classes }}{% endif -%}"
```
36degrees, line 5: I'm also not 100% sure about splitting attributes over multiple lines, in terms of the readability of the HTML that's outputted. (It's fine to split the Nunjucks wrapped attributes out, as the whitespace control puts them back on one line anyway)

We use a prettifier on the HTML examples in the Design System, but I don't think it removes extra new lines like these ones. Worth checking with a pre-release to see what it looks like?

src/govuk/common.unit.test.mjs
```diff
@@ -0,0 +1,103 @@
...
+      b: 'bee',
+      c: { a: 'cat', o: 'cobra' }
+    }
+    const config3 = {
+      b: 'bat',
+      c: { o: 'cow' },
+      d: 'dog'
+    }
+
+    test('flattens a single object', () => {
```
36degrees, line 27: Minor, but is there a reason we're using `test` here when we use `it` everywhere else?

36degrees (commented):
A few very minor things but generally this is looking ace. Great work on the tests!

src/govuk/common.unit.test.mjs
vanitabarrett: Ollie also [left the same comment on my PR](https://github.com/alphagov/govuk-frontend/pull/2819/files#r960376063), so I thought I'd copy in my response here because I'm pretty sure this originated from my unit test spike!

> Good point 🤔 There's a bit of info on [the difference here](https://stackoverflow.com/a/56072272/1669682), but I think `it` makes the most sense in terms of 1) what we're doing here, and 2) to be consistent with our other tests. 

> I think the only reason it's crept in is because it's what [Jest uses in it's Getting Started guide](https://jestjs.io/docs/getting-started). Will amend now!

querkmachine: Yeah, I was just copying the tests from the spike. Happy to make these into `it`s instead.

src/govuk/components/accordion/template.njk
querkmachine: Can I check what the template tests need to cover? Making sure the values are escaped? 
I've already added tests to ensure that the UI text matches the data-* attributes to integration tests.

src/govuk/components/accordion/accordion.yaml
querkmachine: Nice catch 👍 Thanks

src/govuk/components/accordion/template.njk
querkmachine: I've added tests to check for the existence and values of the data-* attributes. I couldn't find a way to test that their values were being properly escaped by the Nunjucks filter, however, as the HTML seems to have already been 'transformed' back into regular HTML by the time the Jest is looking at it.

36degrees: Aah of course, that makes sense. I guess if they weren't being escaped correctly the attributes would be malformed and the values would be incorrect?

querkmachine: Yep, that's the best we can hope for, I think.

36degrees (approved):

PR #2819, opened 2022-08-31
Add unit tests for helpers

vanitabarrett:
This PR adds some unit tests for `lib/file-helpers.js` and `lib/helper-functions.js`.

Things I considered in scope for this testing:

- error handling
- checking expected results from any data manipulation

Things I considered out of scope for this testing:

- functions which essentially act as a wrapper around built-in functions, e.g: `readFileContents` which just calls `fs.readFileSync`
- axe-helper.js, as it simply sets up aXe with a set of config
- jest-helpers.js - I *think* if any of these changed/broke, our component tests would break. These also felt tricky to unit test, as they mostly render out HTML or CSS

### DIFF
--- a/lib/file-helper.unit.test.js
+++ b/lib/file-helper.unit.test.js
@@ -1,0 +1,70 @@
+/* eslint-env jest */
+
+const path = require('path')
+const fileHelper = require('../lib/file-helper')
+
+describe('getComponentData', () => {
+  it('returns an error if unable to load component data', () => {
+    expect(() => { fileHelper.getComponentData('not-a-real-component') }).toThrow(Error)
+  })
+
+  it('looks up the correct component path', () => {
+    jest.spyOn(path, 'join')
+
+    fileHelper.getComponentData('accordion')
+
+    expect(path.join).toHaveBeenCalledWith('src/govuk/components/', 'accordion', 'accordion.yaml')
+  })
+
+  it('outputs objects with an array of params and examples', () => {
+    var componentData = fileHelper.getComponentData('accordion')
+
+    expect(componentData).toEqual(expect.objectContaining({
+      params: expect.any(Array),
+      examples: expect.any(Array)
+    }))
+  })
+
+  it('outputs a param for each object with the expected attributes', () => {
+    var componentData = fileHelper.getComponentData('accordion')
+
+    componentData.params.forEach((param) => {
+      expect(param).toEqual(
+        expect.objectContaining({
+          name: expect.any(String),
+          type: expect.any(String),
+          required: expect.any(Boolean),
+          description: expect.any(String)
+        })
+      )
+    })
+  })
+
+  it('contains example objects with the expected attributes', () => {
+    var componentData = fileHelper.getComponentData('accordion')
+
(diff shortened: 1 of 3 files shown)


lib/file-helper.unit.test.js
```diff
@@ -0,0 +1,61 @@
+/* eslint-env jest */
+
+const fileHelper = require('../lib/file-helper')
+
+describe('getComponentData', () => {
```
36degrees, line 5: I like the idea of these tests, but it feels like they're mostly testing the contents of the YAML files rather than the `getComponentData` function itself.

Potentially they also overlap quite a lot with the package tests?

https://github.com/alphagov/govuk-frontend/blob/1abe3e3efb38aad972cc0644387b130dbea8f86f/tasks/gulp/__tests__/after-build-package.test.js#L133-L157

If we think they're useful, I wonder if we need to tweak the descriptions or something to make this clearer?

lib/helper-functions.unit.test.js
```diff
@@ -0,0 +1,31 @@
+/* eslint-env jest */
+
+const helperFunctions = require('../lib/helper-functions')
+
+describe('componentNameToMacroName', () => {
```
36degrees, line 5: These here are some real nice tests 🎉

lib/helper-functions.unit.test.js
```diff
@@ -0,0 +1,31 @@
...
+
+    expect(macroName).toBe('govukCharacterCount')
+  })
+})
+
+describe('componentNameToJavaScriptModuleName', () => {
+  test('transfoms a single word component name', () => {
+    var macroName = helperFunctions.componentNameToJavaScriptModuleName('button')
+
+    expect(macroName).toBe('GOVUKFrontend.Button')
```
36degrees, line 23: These aren't macro names any more, so possibly need a different variable name? (Same for the function below)

```suggestion
    var moduleName = helperFunctions.componentNameToJavaScriptModuleName('button')

    expect(moduleName).toBe('GOVUKFrontend.Button')
```

(Alternatively, you could get rid of the variables entirely?)

```suggestion
    expect(helperFunctions.componentNameToJavaScriptModuleName('button'))
      .toBe('GOVUKFrontend.Button')
```

lib/file-helper.unit.test.js
```diff
@@ -0,0 +1,61 @@
...
+  test('returns an error if unable to load component data', () => {
```
36degrees, line 6: Minor, but is there a reason we're using `test` here when we use `it` everywhere else?

36degrees (commented):
👏🏻

lib/file-helper.unit.test.js
vanitabarrett: Good point 🤔 There's a bit of info on [the difference here](https://stackoverflow.com/a/56072272/1669682), but I think `it` makes the most sense in terms of 1) what we're doing here, and 2) to be consistent with our other tests. 

I think the only reason it's crept in is because it's what [Jest uses in it's Getting Started guide](https://jestjs.io/docs/getting-started). Will amend now!

36degrees (approved):

PR #2821, opened 2022-09-02
Avoid duplicated `--error` class on character count

36degrees:
The character count is adding the `--error` modifier class when there is an error message passed, but as part of a call to the textarea component which is already handling this for us using similar logic:

https://github.com/alphagov/govuk-frontend/blob/8f8fa6f9a64753f4bc3c36328c6fdb0539cbb52c/src/govuk/components/textarea/template.njk#L40

As a result, when an error message is passed we currently end up with duplicated `--error` modifier classes on the textarea:

```html
<textarea class="govuk-textarea govuk-textarea--error govuk-js-character-count govuk-textarea--error" id="exceeding-characters" name="exceeding" rows="5" aria-describedby="exceeding-characters-info exceeding-characters-error">
</textarea>
```

Remove the logic to add the `--error` modifier class from the character count component and leave it to the textarea component to handle this.


owenatgov (approved):
LGTM!

Issue #2700, opened 2022-07-15
Expand internationalisation logic to add string interpolation feature


vanitabarrett:
## What
In some components, we'll need to be able to insert a dynamic string or value inside a translated string, e.g: "You have [x] characters remaining". We need to give users a way to pass us a string and indicate where the value should go, e.g: `You have %{count} character remaining`

Expand the existing internationalisation logic to handle this kind of string interpolation.

## Why
We cannot hardcode where in the string that value will sit, because it will be different for different languages. 

## Who needs to work on this
Developers

## Who needs to review this
Developers

## Done when

- [ ] Internationalisation logic has been expanded to handle string interpolation
- [ ] Unit tests added for new string interpolation logic

PR #2822, opened 2022-09-02
Add string interpolation functionality to i18n

vanitabarrett:
Closes https://github.com/alphagov/govuk-frontend/issues/2700

## What
Adds logic to `i18n.mjs` to allow strings with placeholders to be passed in and replaced with dynamic data.

For example:
```
{
    "nameString": "My name is ${name}"
}

I18n.t('nameString', { name: "Vanita" })
```

Note: at the moment, the logic throws an error if it can't replace placeholder strings because it can't find the relevant data in the options object. Unsure if this is the behaviour we want, or if we just want to return the string with placeholders present, but I've gone with error in this PR so far. Happy to hear other thoughts. 🤷‍♀️

### DIFF
--- a/src/govuk/i18n.mjs
+++ b/src/govuk/i18n.mjs
@@ -27,7 +27,7 @@ I18n.prototype.t = function (lookupKey, options) {
     var translationString = this.translations[lookupKey]
 
     // Check for ${} placeholders in the translation string
-    if (translationString.match(/\${(.\S+)}/)) {
+    if (translationString.match(/%{(.\S+)}/)) {
       if (!options) {
         throw new Error('i18n: cannot replace placeholders in string if no option data provided')
       }
@@ -51,7 +51,7 @@ I18n.prototype.t = function (lookupKey, options) {
  * @returns  {String}             - The translation string to output, with ${} placeholders replaced
  */
 I18n.prototype.replacePlaceholders = function (translationString, options) {
-  var placeholderRegex = RegExp(/\${(.\S+)}/, 'g')
+  var placeholderRegex = RegExp(/%{(.\S+)}/, 'g')
   var placeholderMatch
 
   // Use `exec` for fetching regex matches, as matchAll() is not supported in IE
--- a/src/govuk/i18n.mjs
+++ b/src/govuk/i18n.mjs
@@ -15,19 +15,63 @@ function I18n (translations) {
  * returns the appropriate string.
  *
  * @param    {String}  lookupKey  - The lookup key of the string to use.
+ * @param    {Object}  options    - Any options passed with the translation string, e.g: for string interpolation.
  * @returns  {String}             - The appropriate translation string.
  */
-I18n.prototype.t = function (lookupKey) {
+I18n.prototype.t = function (lookupKey, options) {
   if (!lookupKey) {
     // Print a console error if no lookup key has been provided
     throw new Error('i18n lookup key missing.')
   } else if (lookupKey in this.translations) {
-    // Return the translation string if it exists in the object
-    return this.translations[lookupKey]
+    // Fetch the translation string for that lookup key
+    var translationString = this.translations[lookupKey]
+
+    // Check for ${} placeholders in the translation string
+    if (translationString.match(/\${(.\S+)}/)) {
+      if (!options) {
+        throw new Error('i18n: cannot replace placeholders in string if no option data provided')
+      }
+
+      return this.replacePlaceholders(translationString, options)
+    } else {
+      return translationString
+    }
   } else {
     // Otherwise, return the lookup key itself as the fallback
     return lookupKey
   }
+}
+
+/**
+ * Takes a translation string with placeholders, and replaces the placeholders
+ * with the provided data
+ *
+ * @param    {String}  translationString  - The translation string
+ * @param    {Object}  options    - Any options passed with the translation string, e.g: for string interpolation.
+ * @returns  {String}             - The translation string to output, with ${} placeholders replaced
+ */
+I18n.prototype.replacePlaceholders = function (translationString, options) {
+  var placeholderRegex = RegExp(/\${(.\S+)}/, 'g')
+  var placeholderMatch
+
+  // Use `exec` for fetching regex matches, as matchAll() is not supported in IE
+  while ((placeholderMatch = placeholderRegex.exec(translationString)) !== null) {
+    var placeholderIncludingBraces = placeholderMatch[0]
+    var placeholderKey = placeholderMatch[1]
+    if (Object.prototype.hasOwnProperty.call(options, placeholderKey)) {
+      // If a user has passed `false` as the value for the placeholder
+      // treat it as though the value should not be displayed
+      if (options[placeholderKey] === false) {
+        translationString = translationString.replace(placeholderIncludingBraces, '')
+      }
+
+      translationString = translationString.replace(placeholderIncludingBraces, options[placeholderKey])
+    } else {
+      throw new Error('i18n: no data found to replace ' + placeholderMatch[0] + ' placeholder in string')
+    }
+  }
+
+  return translationString
 }
 
 export default I18n
--- a/src/govuk/i18n.unit.test.mjs
+++ b/src/govuk/i18n.unit.test.mjs
@@ -3,8 +3,6 @@
  */
 /* eslint-env jest */
 
-/* eslint-disable no-template-curly-in-string */
-
 import I18n from './i18n.mjs'
 
 describe('I18n', () => {
@@ -44,7 +42,7 @@ describe('I18n', () => {
 
   describe('string interpolation', () => {
     const config = {
-      nameString: 'My name is ${name}'
+      nameString: 'My name is %{name}'
     }
 
     it('throws an error if the options data is not present', () => {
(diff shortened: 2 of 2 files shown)


src/govuk/i18n.mjs
```diff
@@ -15,19 +15,56 @@ function I18n (translations) {
  * returns the appropriate string.
  *
  * @param    {String}  lookupKey  - The lookup key of the string to use.
+ * @param    {Object}  options    - Any options passed with the translation string, e.g: for string interpolation.
  * @returns  {String}             - The appropriate translation string.
  */
-I18n.prototype.t = function (lookupKey) {
+I18n.prototype.t = function (lookupKey, options) {
```
36degrees, line 21: Are we calling this `options` to allow for future things to be passed in the object that are _not_ placeholders?

If so, are we happy that we'll have placeholders and other options mixed together in a single object with no real way to differentiate between them? Would we be better off either having a separate arg or having a nested object within options that has the placeholders?

If not, should we call this something else, like `placeholders`?

(I've said placeholders too many times now. It's stopped sounding like a word. I'm not even sure it's the right word. Is there a common term for this that's used by other i18n libraries, and if so should we adopt it?)

src/govuk/i18n.mjs
```diff
@@ -15,19 +15,56 @@ function I18n (translations) {
...
-    // Return the translation string if it exists in the object
-    return this.translations[lookupKey]
+    // Fetch the translation string for that lookup key
+    var translationString = this.translations[lookupKey]
+
+    // Check for ${} placeholders in the translation string
+    if (translationString.match(/\${(.\S+)}/)) {
+      if (!options) {
+        throw new Error('i18n: cannot replace placeholders in string if no option data provided')
+      }
```
36degrees, line 33: Is this needed if we're throwing errors when the placeholders aren't found?

src/govuk/i18n.unit.test.mjs
```diff
@@ -39,4 +41,46 @@ describe('I18n', () => {
       expect(() => i18n.t()).toThrow('i18n lookup key missing.')
     })
   })
+
+  describe('string interpolation', () => {
```
36degrees, line 45: I think it might be worth adding tests to cover some more of the 'stress cases', things like:

- Replacing placeholders with empty strings, zero or anything else that might be falsey (I think this might fail at the minute)
- Strings that contain things that look a bit like placeholders but aren't, like not having a closing brace
- Nested placeholders like `${name${foo}}`
- Placeholder keys that contain non alphanumeric characters
- Replacements that themselves contain placeholders, i.e. `i18n.t('nameString', { name: '${name}' }) }`
- Placeholders keys that match [function names found on objects](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object), i.e. `assign`, `create`

Thoughts?

src/govuk/i18n.mjs
```diff
@@ -15,19 +15,56 @@ function I18n (translations) {
...
+  var placeholderRegex = RegExp(/\${(.\S+)}/, 'g')
+  var placeholderMatch
+
+  // Use `exec` for fetching regex matches, as matchAll() is not supported in IE
+  while ((placeholderMatch = placeholderRegex.exec(translationString)) !== null) {
+    var placeholderKey = placeholderMatch[1]
+    if (!options[placeholderKey]) {
+      throw new Error('i18n: no data found to replace ' + placeholderMatch[0] + ' placeholder in string')
+    } else {
+      translationString = translationString.replace(placeholderMatch[0], options[placeholderKey])
```
36degrees, line 63: If you're not familiar with the output of `RegExp.match` then it's not 100% clear what this is doing.

Worth assigning `placeholderMatch[0]` to a variable so we can give it a useful name like we have with `placeholderKey`?

src/govuk/i18n.unit.test.mjs
```diff
@@ -39,4 +41,46 @@ describe('I18n', () => {
...
+    let config = {}
+
+    beforeEach(() => {
+      config = {
+        nameString: 'My name is ${name}'
+      }
+    })
```
36degrees, line 52: Is `config` ever changed?

Any reason this couldn't be

```suggestion
    const config = {
      nameString: 'My name is ${name}'
    }
```

src/govuk/i18n.mjs
```diff
@@ -15,19 +15,56 @@ function I18n (translations) {
...
+ * @returns  {String}             - The translation string to output, with ${} placeholders replaced
+ */
+I18n.prototype.replacePlaceholders = function (translationString, options) {
+  var placeholderRegex = RegExp(/\${(.\S+)}/, 'g')
+  var placeholderMatch
+
+  // Use `exec` for fetching regex matches, as matchAll() is not supported in IE
+  while ((placeholderMatch = placeholderRegex.exec(translationString)) !== null) {
+    var placeholderKey = placeholderMatch[1]
+    if (!options[placeholderKey]) {
```
romaricpascal, line 60: I think this will catch falsy values, especially `''`, as well, is that:

1. something we want to happen, in which case we might need to reword the error message as I'm not sure people will think of checking that the value they have in their hash is not falsy when reading "no data found"
2. something we don't want to happen (the only two cases I could imagine is passing an empty string or `false`), in which case we'll need to change the check to `placeholderKey in options`

I'd edge towards 2 as enabling empty values allows to interpolate things like "Some content ${optionalAdditionalContent}", but I'm obviously young to the issue so this may have been discusses ahead of time.

src/govuk/i18n.mjs
querkmachine: I noted this in my original spike but I don't think anyone was particularly opinionated on it. A reason to potentially keep it 'as is' is that some things that are passed are kind of both. e.g. The `count` option is both a placeholder AND determines which plural form is used.

src/govuk/i18n.mjs
36degrees: Good catch. I hinted at this in my comment about tests, but probably should have called it out explicitly 😄 

Agree we should support replacing with empty strings, and other falsey things like `0` 👍🏻 

> in which case we'll need to change the check to placeholderKey in options

I think this will match names of properties in the prototype chain (e.g. `'hasOwnProperty' in {}` is `true`), but using `hasOwnProperty` should work I think?

src/govuk/i18n.unit.test.mjs
vanitabarrett: I've added tests for the following, as I think it makes sense to 'stress test' a bit more, as you said:

- Replacing placeholders with empty strings, zero or anything else that might be falsey (I think this might fail at the minute)
- Strings that contain things that look a bit like placeholders but aren't, like not having a closing brace
- Placeholder keys that contain non alphanumeric characters
- Placeholders keys that match [function names found on objects](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object), i.e. assign, create

Let me know if you think those are covered enough now

For the other two (nested placeholders and replacements that themselves contain placeholders), that feels more like additional functionality? I definitely haven't added the logic with that in mind. Is that something we want to support? Related, have you got any examples of how that might be used?

src/govuk/i18n.mjs
vanitabarrett: You're right, the error will be flagged later on. However, it catches the error slightly earlier, so it doesn't have to do the regex lookup and while loop before it throws.

src/govuk/i18n.mjs
vanitabarrett: Thanks @romaricpascal @36degrees , really good catch!

src/govuk/i18n.mjs
vanitabarrett: Yep... any opinions on the naming? I'm thinking `placeholderMatchString`, `placeholderToReplace` - something that's enough to differentiate it from the placeholderKey 🤔

36degrees: Maybe something like `placeholderWithDelimiters`, `placeholderWithBraces` or `placeholderIncludingBraces`?

src/govuk/i18n.unit.test.mjs
36degrees: Looks great 👍🏻 

For the other two examples, I was thinking purely from the perspective of 'what sorts of inputs could break this?' rather than suggesting additional functionality.

So e.g. checking that we don't somehow cause an infinite loop with a placeholder that replaces with itself, and that the regex behaves 'as expected' when there's a placeholder that looks like `${name${foo}}` (I think it should just replace with a key `name${foo}` from the object?)

vanitabarrett: Ah I see, good point! I'll add some more tests for that.

src/govuk/i18n.mjs
36degrees: Happy to stick with `options`, basically because [it's consistent with what i18next are doing](https://www.i18next.com/overview/api#t) 😄

src/govuk/i18n.unit.test.mjs
36degrees: Sorry, that definitely wasn't clear from my original comment!

src/govuk/i18n.mjs
```diff
@@ -15,19 +15,62 @@ function I18n (translations) {
...
     // Print a console error if no lookup key has been provided
     throw new Error('i18n lookup key missing.')
   } else if (lookupKey in this.translations) {
-    // Return the translation string if it exists in the object
-    return this.translations[lookupKey]
+    // Fetch the translation string for that lookup key
+    var translationString = this.translations[lookupKey]
+
+    // Check for ${} placeholders in the translation string
+    if (translationString.match(/\${(.\S+)}/)) {
```
querkmachine, line 30: Something I've noticed is that the format for placeholders here is `${variableName}`, changed from `%{variableName}` in my previous spikes. 

I used `%{}` previously as this is the same format used in [govuk_publishing_components](https://github.com/alphagov/govuk_publishing_components/blob/main/config/locales/cy.yml#L31). 

My concern with using `${}` is it's the same format as JavaScript's own template literals, and that feels like it may create conflicts in the code implemented by services and potentially our own when components are refactored to take advantage of newer JavaScript features. 

For a bit of a contrived example:

```javascript
const prefix = 'You have'
new CharacterCount($element, {
  i18n: {
    charactersUnderLimitOne: `${prefix} ${count} character remaining`,
    charactersUnderLimitOther: `${prefix} ${count} characters remaining`,
    charactersOverLimitOne: `${prefix} ${count} character too many`,
    charactersOverLimitOther: `${prefix} ${count} characters too many`
  }
}).init();
```

In this instance, `${prefix}` is being used to interpolate a string into a JavaScript template literal, but `${count}` is expected to be replaced by our i18n function. Instead, this is likely to produce an error stating that the `count` variable doesn't exist in this context.

vanitabarrett: That's a good point, I'll change it 👍

src/govuk/i18n.unit.test.mjs
```diff
@@ -39,4 +39,91 @@ describe('I18n', () => {
...
+
+    it('can lookup a placeholder value with non-alphanumeric key', () => {
+      const i18n = new I18n({
+        age: 'My age is %{current-age}'
+      })
+      expect(i18n.t('age', { 'current-age': 55 })).toBe('My age is 55')
+    })
+
+    it('can lookup a placeholder value with reserved name as key', () => {
```
36degrees, line 77: I think it might be worth either modifying this test or adding another one where the options object does _not_ have a key matching the name of the function in the prototype, as otherwise we wouldn't go up the prototype chain anyway.

(I've realised that `assign` is a static method on the Object class, and so wouldn't work for these examples, but `valueOf` would)

Basically I think we want to check that the way we're using `Object.hasOwnProperty` is doing its job and we'll never try and replace a property with something from the prototype chain.

If we were using `if (placeholderKey in options)` instead, then we could end up doing something like this:

```
> config = new Object
{}
> config.name = 'Ollie'
'Ollie'
> 'valueOf' in config
true
> 'I have the value of %{valueOf}'.replace('%{valueOf}', config.valueOf)
Uncaught TypeError: Cannot convert undefined or null to object
    at valueOf (<anonymous>)
    at String.replace (<anonymous>)
```

Let me know if that still doesn't make sense – finding it hard to explain!

vanitabarrett: Added a new test here: https://github.com/alphagov/govuk-frontend/pull/2822/files#diff-ebbbe0f92d8344a9adf41d13a754cfddb236d06c1119a3b68226746800972cfdR84

36degrees (approved):
Great work 🎉

Issue #979, opened 2018-09-05
Conditional reveal radio component has aXe errors


nickcolley:
Originally raised by @injms on [alphagov/govuk-design-system](https://github.com/alphagov/govuk-design-system/issues/524)
> The aXe accessiblity tool flags two issues when using the conditional reveal radio component. Running aXe on the [example page](https://design-system.service.gov.uk/components/radios/conditional-reveal/index.html) shows the following issues:

---

#### Elements must only use allowed ARIA attributes

```markup
<input class="govuk-radios__input" id="how-contacted-conditional-1" name="how-contacted" type="radio" value="email" aria-controls="conditional-how-contacted-conditional-1" aria-expanded="false">
```

To solve this violation, you need to:

Fix the following:

 - ARIA attribute is not allowed: aria-expanded="false"

#### Radio inputs with the same name attribute value must be part of a group

```markup
<input class="govuk-radios__input" id="how-contacted-conditional-1" name="how-contacted" type="radio" value="email" aria-controls="conditional-how-contacted-conditional-1" aria-expanded="false">
```

To solve this violation, you need to:

Fix at least one (1) of these issues:
 - All elements with the name "how-contacted" do not reference the same element with aria-labelledby
 - Fieldset contains unrelated inputs

---

I think the expected behaviour would be to have no critical issues flagged by aXe - unless this is a false alarm or is found not to have any impact after testing in the empathy lab.

@36degrees:
> (For context, this is something that's come up before in alphagov/govuk_elements#455, but we don't currently have it recorded on the Frontend repo)

hannalaakso:
Results of my partial investigation:

### Error: "Elements must only use allowed ARIA attributes"

#### Underlying issue
`aria-expanded` is [not allowed](https://www.w3.org/WAI/PF/aria-1.1/states_and_properties#aria-expanded) on `<input>` elements, hence the error (although Siteimprove and W3C validator don't flag this).

###### Alternative approach  

Tested W3C's guidance about [alternative way of revealing content](https://www.w3.org/TR/wai-aria-practices/#accordion): "Optionally, each element that serves as a container for panel content has role region and aria-labelledby with a value that refers to the button that controls display of the panel" This made the experience worse for NVDA users as the "collapsed/expanded" was no longer announced on the radio

[Code](https://github.com/alphagov/govuk-frontend/compare/conditional-axe-violations?expand=1)

<details>

**NVDA / Firefox (worse experience)**

[...]

PR #2823, opened 2022-09-02
Remove aria-expanded from radios with conditional content when JS is enabled

ttbarnes:
This PR removes the `aria-expanded` attribute from radios with conditional content.

Without this change, accessibility tools such as AXE and Lighthouse flag a critical issue:

> ARIA attribute is not allowed: aria-expanded="false"

Currently, `aria-expanded` is not supported for radio buttons. This may change: https://github.com/w3c/aria/issues/1404

### DIFF
--- a/package/govuk-esm/components/radios/radios.mjs
+++ b/package/govuk-esm/components/radios/radios.mjs
@@ -82,7 +82,6 @@ Radios.prototype.syncConditionalRevealWithInputState = function ($input) {
   if ($target && $target.classList.contains('govuk-radios__conditional')) {
     var inputIsChecked = $input.checked
 
-    $input.setAttribute('aria-expanded', inputIsChecked)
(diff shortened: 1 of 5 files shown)


querkmachine:
Hi @ttbarnes, thanks for the contribution. 

We currently don't intend to remove `aria-expanded` from conditional reveals as—despite being an invalid use under the the current ARIA specification—we've found that some screen readers do expose this information to users, and we haven't noticed any detrimental affects from the attributes being present. 

In this instance, we believe the benefit to screen reader users outweighs being invalid according to the spec and, as you note, the spec may change in future to accommodate how we're using it.

Further information about our research and testing is in #979. The validation error in Axe is also listed as a [known issue](https://github.com/alphagov/govuk-frontend/issues/1280#issuecomment-509588851).

Issue #2584, opened 2022-03-30
Accessibility issue in the header


SalarGhotaslo:
## Description of the issue

The .govuk-skip-link in the template component is causing an accessibility issue that states **_page content is not contained by landmarks_**.  

## Steps to reproduce the issue

On the https://design-system.service.gov.uk/ if you inspect the page then --> go on Axe dev tools --> Scan page --> then you'll see something similar to the screenshot below. You can then highlight the issue by clicking on the _All page content should be contained by landmarks_ and then clicking the highlight.

![image](https://user-images.githubusercontent.com/74678444/160815471-7026898f-d836-406c-a55a-5c4cd9ebebd0.png)

## Actual vs expected behavior

I would expect there to be no accessibility issue.

## Environment (where applicable)

The issue can be seen on the https://design-system.service.gov.uk/ webpage so it is on the latest one. 

We found it happen when trying to add the cookie banner component into our project, it seemed to then expose this issue. We can see that this issue wouldn't happen if the govuk-skip-link was inside the header component rather than inside the template component.

- Operating system:
- Browser:
- Browser version:
- GOV.UK Frontend Version:

SalarGhotaslo:
Thanks for the prompt response! Will have a read

Issue #1949, opened 2020-09-09
All page content must be contained by landmarks


Dawdre:
We have been running weekly accessibility tests on all our services using govuk frontend. We use axe browser extension to test the DOM. The breadcrumbs component is consistently showing this issue...
![image](https://user-images.githubusercontent.com/11505395/92602661-5737f780-f2a6-11ea-934f-9404252df060.png)

One of our applications has recently had a 3rd party accessibility audit carried out. This was also raised as a problem. Is this an issue that is known about?

36degrees:
This is a duplicate of the (previously closed) #1604. I've re-opened that issue in favour of this one as it has more context.

Although ensuring that all elements are contained by landmarks is flagged as best practise, we're not aware of having elements outside of landmarks causing any issues (at least not seen in user research) and it's our understanding that it is not a WCAG failure.

Issue #1679, opened 2019-12-19
Improvements to the header component


36degrees:
The header needs to be more flexible – for example, it should allow for just having one navigational link without that link ending up in a menu on mobile; it should allow custom HTML so that you can include a sign out button (styled as a link).

There might be things we can do to make the header more useful in other contexts - eg, full-width, departmental, admin, caseworking systems, etc.

## Stories

- [ ] Header should support a single link without collapsing to a menu on mobile #1635 
- [ ] Header Component - Custom HTML option for links #1573
- [ ] Product name wraps onto new line with non full-width header #1502
- [ ] Review if `aria-label` in header navigation is necessary #1340
- [ ] 'Unlabelled graphic' read out within header using NVDA #1636
- [ ] Make it possible to override or remove the Crown logotype from the header #1639

nickcolley:
We caught up with Max and Ian from GOV.UK Publishing and agreed that they should review our proposed changes to the header (and footer) to ensure it makes sense for their use case.

vanitabarrett:
@36degrees Just wanted to check - does this include https://github.com/alphagov/govuk-frontend/issues/1904 ? It came up in the Friday call today and I wasn't sure if it was included in this work as it has been split out from https://github.com/alphagov/govuk-frontend/issues/1682

36degrees:
Closing this as having an epic for the component isn't currently useful for us.

PR #2824, opened 2022-09-02
Add complementary aria role to phase banner

ttbarnes:
This PR adds a `complementary` landmark role attribute to the phase banner.

> [landmark role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles#landmark_roles) is used to designate a supporting section that relates to the main content, yet can stand alone when separated. These sections are frequently presented as sidebars or call-out boxes

Without this, accessibility tools such as AXE flag a moderate issue:

> "All page content should be contained by landmarks"
> "Some page content is not contained by landmarks"

### DIFF
--- a/app/views/partials/banner.njk
+++ b/app/views/partials/banner.njk
@@ -1,5 +1,5 @@
 {% if shouldShowAppBanner %}
-<div class="app-banner app-banner--warning" data-module="app-banner">
+<div class="app-banner app-banner--warning" data-module="app-banner" role="complementary">
   <div class="govuk-width-container">
     <p class="govuk-body">
       This is an internal development app.
--- a/app/views/partials/exampleBanner.njk
+++ b/app/views/partials/exampleBanner.njk
@@ -1,5 +1,5 @@
 {% if shouldShowAppBanner %}
-<div class="app-banner app-banner--warning" data-module="app-banner">
+<div class="app-banner app-banner--warning" data-module="app-banner" role="complementary">
   <div class="govuk-width-container">
     <p class="govuk-body">
       This is not a real service, and is used for testing purposes only.
--- a/package/govuk/components/phase-banner/template.njk
+++ b/package/govuk/components/phase-banner/template.njk
@@ -1,7 +1,7 @@
 {% from "../tag/macro.njk" import govukTag -%}
 
 <div class="govuk-phase-banner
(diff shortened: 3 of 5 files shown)


querkmachine:
This seems something worth fixing, as the Axe failure gets flagged in support semi-frequently. I had two big questions around it, however: 

## Is `role="complementary"` the most appropriate role for this content?

It seems like it. 

`role="complementary"` is for elements that are semantically "related to the [primary] content, but not the [primary] content itself". 

One could argue that the phase banner is about the site or service as a whole, rather than the content of the current page, in which case `banner` would be a better fit (being "for globally repeating site-wide content"), however it is also recommended that each page only have a single element with the `banner` role, which we already use that for the sitewide header. The same recommendation exists for the `contentinfo` role on footer.

There have been some murmors of potentially making it possible to place phase banners within the header component, but those haven't been conclusive, to my knowledge. In lieu of that, using `complementary` seems the next best option.

## Why not just make them `<aside>` elements?

Adding `role="complementary"` is semantically equivalent to making these into `<aside>` elements, so there is an argument there to follow [the first rule of ARIA](https://www.w3.org/TR/using-aria/#rule1) and to convert the existing `<div>`s into `<aside>`s, rather than adding a `role`. 

This won't work for legacy IE versions, where we'd need to explicitly add a `role` anyway. I'm also not sure if changing the element entirely consitutes a larger (possibly breaking) change than adding a `role` attribute, which would affect when we can release the change. My gut says no, but second opinions are always welcome.

colinrotherham:
Hi @ttbarnes

Hope you don't mind but we're going to close this one

We're eventually planning to separate the `<header>` HTML element from the [**Header** component](https://design-system.service.gov.uk/components/header/).

By separating the two we can nest the Phase banner inside `<header>`

```
<header role="banner">
  <div class="govuk-header"></div>
  <div class="govuk-phase-banner"></div>
</header>
```

I'll link to these issues where it's come up before which might be useful:

* https://github.com/alphagov/govuk-frontend/issues/1604
* https://github.com/alphagov/govuk-frontend/issues/1949
* https://github.com/alphagov/govuk-frontend/issues/2584

We're also tracking various improvements in this one:

* https://github.com/alphagov/govuk-frontend/issues/1679

Thanks again Tony

ttbarnes:
Thanks @colinrotherham sounds great 👍

Issue #2802, opened 2022-08-22
Add ability to pass translation strings into accordion component via JS


vanitabarrett:
## What
Add ability for users to pass a translation object in via JavaScript (at the point of initialisation). This is an alternative method to passing translation strings via the Nunjucks macros (data-attributes).

Example:

```
new Accordion($element, {

  "i18n.showAllSections": "Show everything"

}).init()
```

## Why
Some users may prefer to pass translation strings via JavaScript, or it may be better suited to their service.

## Who needs to work on this
Developers

## Who needs to review this
Developers

## Done when

- [x] It’s possible to provide translations to the accordion component via JavaScript
- [x] Expanded logic which looks up translations passed via data attributes, to lookup translations provided by JavaScript
- [x] Data attribute translations take priority over JS translations, if both provided
- [x] The accordion component still works if no/only some translations are passed (falls back to defaults)
- [x] The accordion component has tests which cover the ability to translate via JavaScript

36degrees:
Does this also include being able to pass config via `initAll`, or should that be a separate story?

Issue #2698, opened 2022-07-15
Allow component JavaScript constructor to take a config object


vanitabarrett:
## What
Allow components to take a config object through their constructor, before being initialised. The config object should default to an empty object, and therefore a component must be able to continue working even if the config object is empty (e.g: by setting some defaults to fall back on).

## Why
To allow for options to be passed to our component JavaScript in future, e.g: translations

## Who needs to work on this
Developers

## Who needs to review this
Developers

## Done when

- [x] It's possible to pass a config object to a component via `init()` and `initAll()`
- [x] If that config object is empty, the component JavaScript should still work
- [x] Tests to handle different scenarios and that the JS behaves as expected, e.g: config object not passed, config object is empty, component sets defaults

querkmachine:
Would we want to add a config object to every component upfront (so that the interface exists, even if it's not being used yet), or only to the components as-and-when they have a reason to need it?

36degrees:
Does this also include being able to pass config via `initAll`, or should that be a separate story?

vanitabarrett:
@36degrees I think I imagined it all being part of this story... I don't think the work would be too big to do in one, but if it turned out that way we could split it up. I'll expand the "Done when" section a bit at least to make it clear it's intended to include both

36degrees:
> It's possible to pass a config object to a component via init() and initAll()

Based on [this discussion](https://github.com/alphagov/govuk-frontend/pull/2810#discussion_r955178143), I think we've agreed that config will be passed to the constructor rather than the init function, which also [matches the example in the proposal](https://github.com/alphagov/govuk-frontend/discussions/2740#:~:text=Example%20of%20passing%20translations%20to%20a%20specific%20component%3A).

PR #2826, opened 2022-09-02
Add support for localisation via JavaScript configuration to Accordion component

querkmachine:
This is related to two issues, #2802 and #2698, as I couldn't find a way to cleanly separate them.

- Modifies the `initAll` function to pass `options.accordion` into the accordion initialisation.
- Modifies the accordion component to accept a second `config` parameter (defaults to an empty object).
- Modifies the accordion component's config merging to accept the `config` parameter. 
- Updates the 'translated' full-page example to call `initAll` with a custom configuration.
- Adds integration tests checking that the accordion's text is modified according to the configuration.
- Adds unit tests to the existing `mergeConfig` function to ensure that it supports empty objects.

## Todo

#2802 includes a task to ensure that "Data attribute translations take priority over JS translations, if both provided"

This is partially covered by the configuration merging function's unit tests, though that function also doesn't make an explicit distinction between configuration passed from data-* attributes and those from JavaScript. Do we feel like we need to explicitly test for parameter ordering?

### DIFF
--- a/src/govuk/components/accordion/accordion.mjs
+++ b/src/govuk/components/accordion/accordion.mjs
@@ -20,7 +20,8 @@ import I18n from '../../i18n.mjs'
 import '../../vendor/polyfills/Function/prototype/bind.mjs'
 import '../../vendor/polyfills/Element/prototype/classList.mjs'
 
-function Accordion ($module) {
+function Accordion ($module, config) {
+  config = config || {}
   this.$module = $module
   this.$sections = $module.querySelectorAll('.govuk-accordion__section')
   this.$showAllButton = ''
@@ -34,7 +35,7 @@ function Accordion ($module) {
       showSection: 'Show<span class="govuk-visually-hidden"> this section</span>'
     }
   }
-  this.config = mergeConfigs(defaultConfig, $module.dataset)
+  this.config = mergeConfigs(defaultConfig, config, $module.dataset)
   this.i18n = new I18n(extractConfigByNamespace(this.config, 'i18n'))
 
   this.controlsClass = 'govuk-accordion__controls'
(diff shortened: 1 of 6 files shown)


src/govuk/components/accordion/accordion.mjs
```diff
@@ -34,7 +34,7 @@ function Accordion ($module) {
       showSection: 'Show<span class="govuk-visually-hidden"> this section</span>'
     }
   }
-  this.config = mergeConfigs(defaultConfig, $module.dataset)
+  this.config = mergeConfigs(defaultConfig, config || {}, $module.dataset)
```
vanitabarrett, line 37: Could we expand the existing config merging tests to make sure it handles empty objects ok?

querkmachine: I've added a new unit test which runs `mergeConfigs` with empty objects at various positions and compares those with the result of the same function sans-empty objects. Do you think that covers it?

vanitabarrett: I think that covers it, yep 👍

36degrees (approved):
LGTM! 👍🏻

src/govuk/components/accordion/accordion.test.js
```diff
@@ -283,6 +291,18 @@ describe('/components/accordion', () => {
           expect(allSectionsToggleText).toEqual(hideAllSectionsDataAttribute)
         })
 
+        it('should localise "Hide all sections" based on JavaScript configuration', async () => {
+          await page.goto(baseUrl + '/examples/translated', { waitUntil: 'load' })
+          await page.click('.govuk-accordion .govuk-accordion__section:nth-of-type(2) .govuk-accordion__section-header')
+          await page.click('.govuk-accordion .govuk-accordion__section:nth-of-type(3) .govuk-accordion__section-header')
+          await page.click('.govuk-accordion .govuk-accordion__section:nth-of-type(4) .govuk-accordion__section-header')
+          await page.click('.govuk-accordion .govuk-accordion__section:nth-of-type(5) .govuk-accordion__section-header')
```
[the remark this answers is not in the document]
querkmachine: Yes. Yes it would.

PR #2828, opened 2022-09-05
Validate the HTML output from all component examples

36degrees:
This adds a dependency on HTML Validate, an offline HTML validator, which we can use in our tests to validate the HTML output from every component example.

After adding tests for every component example, I disabled every rule that was failing until the tests were passing, and then. taking each rule in turn either:

- fixed the violations
- tweaked the rule config or the element definitions so that the rule passed whilst still catching other examples
- left the rule disabled and added a comment to explain why we should keep it disabled

I'd recommend reviewing commit-by-commit.


src/govuk/components/all.test.js
```diff
@@ -51,3 +55,75 @@ it.each(allComponents)('generate screenshots for Percy visual regression, with J
...
+  describe('examples output valid HTML', () => {
+    const examples = getComponentData(component).examples.map(function (example) {
+      return [example.name, example.data]
+    })
+
+    it.each(examples)('example "%s" outputs valid HTML', async (_, data) => {
+      const $ = render(component, data)
+
+      expect($.html()).toHTMLValidate({
+        rules: {
```
36degrees, line 69: This config could all go in to [a `.htmlvalidate.js` in the root of the repo](https://html-validate.org/usage/index.html#configuration), but my thinking is whilst we're only using this in one place we might as well keep it all together?

src/govuk/components/button/template.njk
```diff
@@ -37,7 +37,7 @@ treat it as an interactive element - without this it will be
 
 {#- Define common attributes we can use for both button and input types #}
 
-{%- set buttonAttributes %}{% if params.name %} name="{{ params.name }}"{% endif %}{% if params.disabled %} disabled="disabled" aria-disabled="true"{% endif %}{% if params.preventDoubleClick %} data-prevent-double-click="true"{% endif %}{% endset %}
+{%- set buttonAttributes %}{% if params.name %} name="{{ params.name }}"{% endif %}{% if params.disabled %} disabled aria-disabled="true"{% endif %}{% if params.preventDoubleClick %} data-prevent-double-click="true"{% endif %}{% endset %}
```
NickColley, line 40: potentially breaking change

button.disabled will still be true
but button.getAttribute('disabled') will change from "disabled" to and empty string

36degrees: Good catch Nick, thanks.

I'm going to split this change out into a follow up PR as it's the only change to the HTML output of one of the macros.

We're still discussing as a team how we want to treat the change, but at the very least it should go in the release notes, which will be easier once it's split out.

36degrees: Force-pushed to replace the changes in fcd04cd5a9bc2b8122cce5178d69ebf702104424 with 500a02fd2768270726c0658a1ac5608977445ade.

lib/jest-helpers.js
```diff
@@ -15,38 +15,57 @@ const sassRender = util.promisify(sass.render)
...
- * Render a component's macro for testing
- * @param {string} componentName
- * @param {string} params parameters that are used in the component macro
- * @param {any} children any child components or text, pass the children to the macro
- * @returns {function} returns cheerio (jQuery) instance of the macro for easy DOM querying
+ * Render the raw HTML for a component
+ *
+ * @param {String} componentName
+ * @param {Object} options options to pass to the component macro
+ * @param {String} callBlock if provided, the macro is called using the Nunjucks
```
romaricpascal, line 28: Not sure how tight we want to be on that and maybe a bit pedantic, but thought we may as well [show the parameter is optional in the JSDoc](https://jsdoc.app/tags-param.html#optional-parameters-and-default-values).

```suggestion
 * @param {String} [callBlock] if provided, the macro is called using the Nunjucks
```

lib/jest-helpers.js
```diff
@@ -15,38 +15,57 @@ const sassRender = util.promisify(sass.render)
...
-  return cheerio.load(output)
+  return nunjucksEnv.renderString(macroString)
+}
+
+/**
+ * Render component into Cheerio API
+ *
+ * Allows us to interrogate the output of the macro using a jQuery-like syntax
+ *
+* @param {String} componentName
```
romaricpascal, line 58: Just a little misalligned comment 😄 

```suggestion
 * @param {String} componentName
```

lib/jest-helpers.js
```diff
@@ -15,38 +15,57 @@ const sassRender = util.promisify(sass.render)
...
+ * @param {Object} options options to pass to the component macro
+ * @param {String} callBlock if provided, the macro is called using the Nunjucks
```
romaricpascal, line 60: Same as for the `renderHtml` docs :) 

```suggestion
 * @param {String} [callBlock] if provided, the macro is called using the Nunjucks
```

romaricpascal (commented):
Love the idea to validate the HTML we output 🙌🏻 Just caught a few details in the JSDocs that could be tweaked, but if we don't want to go that nitpicky, feel free to leave them out 😄

romaricpascal (approved):

PR #2830, opened 2022-09-07
Set the boolean `disabled` attribute consistently in the button component

36degrees:
Remove the value from the attribute so that we're setting boolean HTML attributes consistently across all components – matching what we're doing in the radios, checkboxes and select components.

Re-enable the 'attribute-boolean-style' rule so that we enforce this going forward.

Split out from #2828.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,18 @@
 
 ## Unreleased
 
+### Breaking changes
+
+#### Check that disabled buttons work as expected
+
+The `disabled` attribute on disabled buttons created using our Nunjucks macros no longer includes a value.
+
+If you are using `$button.getAttribute('disabled')` to check for the disabled attribute in JavaScript, this will now return an empty string. This may cause unexpected behaviour if you are relying on the return value being the string 'disabled' or being [truthy](https://developer.mozilla.org/en-US/docs/Glossary/Truthy).
+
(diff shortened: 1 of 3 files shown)


romaricpascal:
Another thing that the change may break besides app using JS to check if the button is disabled may be the fixtures for validating non JS implementations. 

Rails outputs boolean attributes in the disabled="disabled" form, I believe. [Through its `button_tag` helper](https://api.rubyonrails.org/v7.0.2/classes/ActionView/Helpers/FormTagHelper.html#method-i-button_tag), for example.

colinrotherham (approved):
Approving this based on a discussion we had yesterday 🙌 

But we also flagged the HTML living standard in reference to whether this is a breaking change or not from https://github.com/alphagov/govuk-frontend/pull/2830#issuecomment-1239524711

**2.3.2 Boolean attributes**
https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attributes

They give three examples considered valid:
```html
<label><input type=checkbox checked name=cheese disabled> Cheese</label>
```

```html
<label><input type=checkbox checked=checked name=cheese disabled=disabled> Cheese</label>
```

```html
<label><input type='checkbox' checked name=cheese disabled=""> Cheese</label>
```

Although it might be breaking, we've essentially switched  from "valid" to "also valid" ✅

querkmachine:
Presumably this can now be merged given 4.6.0 has gone out, conflict notwithstanding.

36degrees:
I need to add a changelog entry as well.

36degrees:
I've rebased to resolve the conflict and added a changelog entry.

colinrotherham (approved):
🎉

PR #2831, opened 2022-09-07
Cleanup the JavaScript for the button component

36degrees:
- Move the init function to the top
- Edit the comments for the other functions for brevity
- Fix indentation and wrap at 80 characters
- Remove redundant comments where the code is already self-documenting or covered by the JSDoc
- Remove a lie (that we listen for events at the document level)

### DIFF
--- a/src/govuk/components/button/button.mjs
+++ b/src/govuk/components/button/button.mjs
@@ -10,47 +10,50 @@ function Button ($module) {
 }
 
 /**
-* Initialise an event listener for keydown at document level
-* this will help listening for later inserted elements with a role="button"
-*/
+ * Initialise component
+ */
 Button.prototype.init = function () {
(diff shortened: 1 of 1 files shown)


owenatgov (commented):
Just needs a changelog entry then I'm happy with this

36degrees:
> Just needs a changelog entry then I'm happy with this

There aren't any changes to the code itself (other than moving a function) so I'm inclined to say this doesn't need a changelog entry – thoughts?

romaricpascal:
> There aren't any changes to the code itself (other than moving a function) so I'm inclined to say this doesn't need a changelog entry – thoughts?

I was surprised by the need for a changelog for that as well, as it doesn't really change things for people using the library (besides clarifying the JSDoc a bit).

owenatgov (approved):
Good point. In practical terms all these changes are inward facing. Approved!

PR #2840, opened 2022-09-12
Normalise booleans and numbers in data attributes

### LINKED ISSUE #2832: Allow for booleans and numbers to be passed via data attributes
## What

Ensure that config passed by data attributes is typed sensibly for the way we want to use it.

## Why

Attributes passed via data attributes will always be strings, but in some cases we want to be able to treat them as booleans or numbers.

For example:

- `data-prevent-double-click` on buttons which is a boolean
- `data-maxlength` on the character count which is a number

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [ ] Attributes that we expect to be booleans or numbers are typed as such when used as config in a component
- [ ] We have good test coverage of any function(s) required to implement this

36degrees:
Attributes passed via data attributes will always be strings, but in some cases we want to be able to treat them as booleans or numbers.

For example:

- `data-prevent-double-click` on buttons which is a boolean
- `data-maxlength` on the character count which is a number

Ensure that config passed by data attributes is typed sensibly for the way we want to use it.

This is heavily inspired by [Bootstrap's approach to dealing with data attributes][1].

Closes #2832 

[1]: https://github.com/twbs/bootstrap/blob/6e55fa929797a2d1d5250e1630615f515c45d0db/js/src/dom/manipulator.js#L8-L34

### DIFF
--- a/src/govuk/common.mjs
+++ b/src/govuk/common.mjs
@@ -129,3 +129,59 @@ export function extractConfigByNamespace (configObject, namespace) {
   }
   return newObject
 }
+
+/**
+ * Normalise string
+ *
+ * 'If it looks like a duck, and it quacks like a duck…' 🦆
+ *
+ * If the passed value looks like a boolean or a number, convert it to a boolean
+ * or number.
+ *
+ * Designed to be used to convert config passed via data attributes (which are
+ * always strings) into something sensible.
+ *
+ * @param {String} value - The value to normalise
+ * @returns {(String|Boolean|Number)} Normalised data
+ */
+export function normaliseString (value) {
+  if (typeof value !== 'string') {
+    return value
+  }
+
+  var trimmedValue = value.trim()
+
+  if (trimmedValue === 'true') {
+    return true
+  }
+
+  if (trimmedValue === 'false') {
+    return false
+  }
+
+  // Empty / whitespace-only strings are considered finite so we need to check
+  // the length of the trimmed string as well
+  if (trimmedValue.length > 0 && isFinite(trimmedValue)) {
+    return Number(trimmedValue)
+  }
+
+  return value
+}
+
+/**
+ * Normalise dataset
+ *
+ * Loop over an object and normalise each value using normaliseData function
+ *
+ * @param {DOMStringMap} dataset
+ * @returns {Object} Normalised dataset
+ */
+export function normaliseDataset (dataset) {
+  var out = {}
+
+  for (var key in dataset) {
+    out[key] = normaliseString(dataset[key])
+  }
+
+  return out
+}
--- a/src/govuk/common.unit.test.mjs
+++ b/src/govuk/common.unit.test.mjs
@@ -3,7 +3,7 @@
  */
 /* eslint-env jest */
 
-import { mergeConfigs, extractConfigByNamespace } from './common.mjs'
+import { mergeConfigs, extractConfigByNamespace, normaliseString, normaliseDataset } from './common.mjs'
 
 // TODO: Write unit tests for `nodeListForEach` and `generateUniqueID`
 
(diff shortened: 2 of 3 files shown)


src/govuk/common.mjs
```diff
@@ -129,3 +129,53 @@ export function extractConfigByNamespace (configObject, namespace) {
...
+  switch (value) {
+    case 'true':
+      return true
+
+    case 'false':
+      return false
+
+    // Matches if converting to a number and back to a string results in the
+    // same value
+    case Number(value).toString():
```
36degrees, line 157: This does mean that these tests all pass as well…

```js
      expect(normaliseData('Infinity')).toEqual(Infinity)
      expect(normaliseData('-Infinity')).toEqual(-Infinity)
      expect(normaliseData('NaN')).toEqual(NaN)
```

I think that's probably fine, but flagging in case we think need to do anything about this.

colinrotherham: Ahh decimals are going to fail the **string-to-number-and-back™** equality check

```html
<element data-amount="12.90">
```

```mjs
console.log(Number('12.90').toString() === '12.90') // false
console.log(Number('12.90').toString() === '12.9') // true
```

Regarding `NaN`/`Infinity` you've got [`Number.isFinite()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isFinite) or the older [`isFinite()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isFinite) global

>[Difference between Number.isFinite() and global isFinite()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isFinite#difference_between_number.isfinite_and_global_isfinite)
>In comparison to the global [isFinite()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isFinite) function, this method doesn't first convert the parameter to a number. This means only values of the type number and are finite return true, and non-numbers always return false.

The nerd in me would want to use [`Number.isSafeInteger()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isSafeInteger) via [core-js polyfill](https://github.com/zloirock/core-js/blob/master/packages/core-js/modules/es.number.is-safe-integer.js) 😆

36degrees: That's a great point. Hard to tell how likely it is that users will want to pass decimals as config.

I do think that Infinity / NaN are more likely to occur due to an error in a service team's code, rather than being passed intentionally, and I don't think typing them as numbers is going to break things any more than if they were kept as strings?

I think we need to try and find the right balance between building out something that's robust enough for our future components, and ideally something that's reusable for bespoke components built by teams, but without slowing us down too much.

Personally, I'm inclined to say 'if it's good enough for Bootstrap, it's good enough for us…'. We could add a comment to note that it won't work for decimals and wait for this to come up as an issue?

colinrotherham: Yep, emphasis on _nerd_ 🙈 

Your call on this, will affect this one too:

```mjs
console.log(Number('0.5').toString() === '.5') // false
console.log(Number('0.5').toString() === '0.5') // true
```

Not a blocker, but a quick fix?

```mjs
console.log(isFinite('.5')) // true
console.log(isFinite('0.5')) // true
console.log(isFinite('12.90')) // true
console.log(isFinite('12.9')) // true

console.log(isFinite('NaN')) // false
console.log(isFinite('Infinity')) // false
console.log(isFinite('-Infinity')) // false
```

```suggestion
    case isFinite(value) && Math.abs(Number(value)) <= 0x1FFFFFFFFFFFFF:
```

36degrees: Updated with an alternative approach. Had to switch (hah!) to an if statement as we're no longer doing an equality check.

Also needed to cater for the empty string as `isFinite('') === true` and `isFinite('   ') === true`!

colinrotherham: Fab, just seen it. I'll reply on that line number!

(Well spotted, knew there'd be a catch using the more-supported `isFinite()` vs `Math.isFinite()`)

romaricpascal (approved):
Looks pretty solid with the updated way to normalise numbers and the extra test cases 🙌🏻

src/govuk/common.mjs
```diff
@@ -129,3 +129,51 @@ export function extractConfigByNamespace (configObject, namespace) {
...
+ * If the passed value looks like a boolean or a number, convert it to a boolean
+ * or number.
+ *
+ * Designed to be used to convert config passed via data attributes (which are
+ * always strings) into something sensible.
+ *
+ * @param {String} value - The value to normalise
+ * @returns {(String|Boolean|Number)} Normalised data
+ */
+export function normaliseData (value) {
```
colinrotherham, line 147: Maybe we want to normalise unhandled values to an empty string?

(Guarding for undefined via `typeof value === 'string'`)

```suggestion
export function normaliseData (value) {
  if (typeof value !== 'string') {
    return value
  }

  // Allow TRUE, True, FALSE, False etc?
  value = value.trim().toLowerCase()
    
  if (['true', 'false'].indexOf(value) !== -1) {
    return value === 'true'
  }
  
  // Allow finite negatives without Math.abs()?
  if (isFinite(value)) {
    return Number(value)
  }
  
  return value
}
```

src/govuk/common.mjs
```diff
@@ -129,3 +129,51 @@ export function extractConfigByNamespace (configObject, namespace) {
...
+ *
+ * 'If it looks like a duck, and it quacks like a duck…' 🦆
+ *
+ * If the passed value looks like a boolean or a number, convert it to a boolean
+ * or number.
+ *
+ * Designed to be used to convert config passed via data attributes (which are
+ * always strings) into something sensible.
+ *
+ * @param {String} value - The value to normalise
```
colinrotherham, line 144: I know this is generally called from `normaliseDataset()` but [DOMStringMap](https://developer.mozilla.org/en-US/docs/Web/API/DOMStringMap) value might be [`undefined`](https://microsoft.github.io/PowerBI-JavaScript/interfaces/_node_modules_typedoc_node_modules_typescript_lib_lib_dom_d_.domstringmap.html)

**Element _without_ `data-missing`**
```mjs
normaliseDataset($module.dataset.missing)
```

It'll throw an error when `value.trim()` is run:
```console
TypeError: undefined is not an object (evaluating 'undefined.trim')
```

```suggestion
 * @param {String|undefined} value - The value to normalise
```

src/govuk/common.mjs
```diff
@@ -129,3 +129,51 @@ export function extractConfigByNamespace (configObject, namespace) {
...
+export function normaliseData (value) {
+  if (value === 'true') {
+    return true
+  }
+
+  if (value === 'false') {
+    return false
+  }
+
+  if (value.trim().length > 0 && isFinite(value)) {
```
colinrotherham, line 156: Love `.trim()` but this is also useful to have for the other checks?

src/govuk/common.mjs
36degrees: Maybe this function should be called `normaliseString`?

It's intended to be used by `normaliseDataset` which loops over the keys that are present in the `DOMStringMap` which I believe will always be strings.

Arguably it could be a private method that we don't export, or we integrate it into the `normaliseDataset` method, but that does make it slightly messier to test.

colinrotherham: Yeah that it really, if it's "public" code and accepts any `DOMStringMap` value I'd add the checks

Otherwise all good as it is, happy to get this merged

colinrotherham (approved):

PR #2842, opened 2022-09-13
Always run tests using latest LTS Node.js version

colinrotherham:
Quick PR just to contribute a few changes, figuring our how all the tasks/actions work etc

Includes:

1. **Cache npm packages in GitHub Actions**
Not huge, but our [Tests](https://github.com/alphagov/govuk-frontend/actions/workflows/tests.yml) action runs ~1 minute faster now

2. **Track Node.js LTS ‘Gallium’ (v16) release**
Following [**running-locally.md**](https://github.com/alphagov/govuk-frontend/blob/main/docs/contributing/running-locally.md#running-locally) the GitHub Action now tracks the latest LTS (currently v16.17.0)

I've kept Node.js v4 support via [./bin/check-nvmrc.js](https://github.com/alphagov/govuk-frontend/compare/test-speed-up?expand=1#diff-056722dd59fc3c64f52ba27400f18d5675f5d85f01f65c8f36ff0adcf0c283ac) (so older versions still get the unsupported message)


.github/workflows/diff-change-to-dist.yaml
```diff
@@ -14,20 +14,28 @@ jobs:
...
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
         with:
           fetch-depth: 0 # Need to also checkout the base branch to compare
-      - uses: actions/setup-node@v3
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v3
+        with:
+          cache: npm
```
colinrotherham, line 24: This caches the `~/.npm` directory which helps speed up `npm ci`

.github/workflows/tests.yml
```diff
@@ -8,17 +8,26 @@ jobs:
...
+
+      - name: Setup Node.js
         uses: actions/setup-node@v3
         with:
-          node-version-file: '.nvmrc'
+          cache: npm
+          node-version-file: .nvmrc
 
       - name: Install dependencies
+        if: steps.npm-cache.outputs.cache-hit != 'true'
```
colinrotherham, line 27: Exact `node_modules` match in the cache? Skip install 🎉

.github/workflows/tests.yml
```diff
@@ -8,17 +8,26 @@ jobs:
...
+        id: npm-cache
+        with:
+          key: npm-cache-${{ hashFiles('package-lock.json', 'package.json') }}
+          path: node_modules
+
+      - name: Setup Node.js
         uses: actions/setup-node@v3
         with:
-          node-version-file: '.nvmrc'
+          cache: npm
```
colinrotherham, line 23: This also caches the ~/.npm directory which helps speed up `npm ci` anyway

.github/workflows/diff-change-to-dist.yaml
36degrees, line 24: I don't know if it makes sense to use the cache here – as I understand it, this'll cause `setup-node` to use the hash of the lock file to restore the dependencies we'd need for `npm install`?

However, in this workflow we don't run `npm install` and we don't use any of the project's dependencies – we use npm to globally install `js-beautify` and use it to diff the `dist` folder against the main branch.

.github/workflows/tests.yml
```diff
@@ -8,17 +8,26 @@ jobs:
...
       - name: Checkout code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
 
-      - name: Setup node
+      - name: Cache dependencies
+        uses: actions/cache@v3
+        id: npm-cache
+        with:
+          key: npm-cache-${{ hashFiles('package-lock.json', 'package.json') }}
```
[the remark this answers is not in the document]
36degrees: Is this not already handled by `setup-node` using `cache: npm`? 🤔 

https://github.com/actions/setup-node#caching-global-packages-data

bin/check-nvmrc.js
```diff
@@ -2,22 +2,39 @@
 
 'use strict'
 
-var fs = require('fs')
-var path = require('path')
+const fs = require('fs')
```
36degrees, line 5: I'm least sure about the changes in this file. I think it's doing broadly the same thing but so much has changed I can't quite follow it.

To be honest, I'm not even really sure if we need this check any more.

From memory, I _think_ we introduced this when there was a lot of churn in the package lock file format between versions of node / npm.

The easiest way for us to avoid constantly checking in new versions of the lock file (or discarding the changes) was for the team to align on using the same version of node / npm.

That doesn't seem to be as much of an issue nowadays, so maybe we could relax the checks we're doing.

.github/workflows/diff-change-to-dist.yaml
colinrotherham: I used to think this too!

Using `cache: npm` via `actions/setup-node` will cache **~/.npm**

For example, when **./node_modules** is missing (or cleared via `npm ci`) it avoids costs network traffic to https://registry.npmjs.com again (which was slowing the tests down)

You can see your own location here:

```console
npm config get cache
```

.github/workflows/tests.yml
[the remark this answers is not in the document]
colinrotherham: Unlike `cache: npm` (only **~/.npm**) this one makes sure **./node_modules** is also cached

It means we can avoid `npm ci` entirely as we have an exact match in the cache. We'll only do this when we're 100% certain (via the `hashFiles()` match) but it's a huge performance win.

Should the hash not match due to package version changes, it's still good to cache **~/.npm** so that the 99% _unchanged_ packages can be reinstalled without hitting the network via https://registry.npmjs.com

bin/check-nvmrc.js
colinrotherham: Can I talk you through it tomorrow?

I've changed the `.nvmrc` file to an LTS release rather than Node.js v16.14.2 (18 Mar 2022)

I'm grabbing LTS information from `process.release.lts` rather than checking if major version numbers match

**.nvmrc**
```
lts/gallium
```

.github/workflows/tests.yml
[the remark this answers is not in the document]
colinrotherham: I've updated this line to remove `'package.json'` as we install using `npm ci` in this workflow

bin/check-nvmrc.js
colinrotherham: As discussed, I've removed **check-nvmrc.js** and replaced it with:

**.npmrc**
```properties
engine-strict=true
```

36degrees (approved):
👍🏻

PR #2843, opened 2022-09-13
Allow notification banner configuration to be passed at construction

### LINKED ISSUE #2839: Allow notification banner config set via data attributes to be passed when initialising the component in JavaScript
## What

The notification banner component currently uses a `data-disable-auto-focus` data attribute to prevent the component from taking focus on page load. (This is mostly designed to be used when the component is displayed in a different context, e.g. in the Design System itself)

Update the component so that the config can also be passed as an object when initialising the component in JavaScript.

## Why

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [x] The 'disable auto focus' functionality can be provided when instantiating a NotificationBanner object
- [x] The 'prevent double click' functionality can be provided when instantiating a NotificationBanner object via `initAll`
- [x] If a `data-disable-auto-focus` data attribute is present on the $module HTML element it overrides any config set via initialisation
- [x] Tests to cover the above

36degrees:
The intention of this PR is to ensure that the config that can currently be set via data attributes can also be set when instantiating the component. Closes #2839.

This builds on conventions already established during work on the accordion component in #2826.

As part of this, we realised that the Nunjucks macro only included the data attribute if the value of `disableAutoFocus` was truthy. To ensure that it's always possible to override the JavaScript config we've updated this so that it will output the value of `disableAutoFocus` as long as it is defined.

We've also introduced a new way to test the behaviour of a component whilst having control over how the component is inititialised, by:
- rendering the HTML for the component manually and then injecting it into an empty 'test boilerplate' page
- control the config passed when the component is instantiated on the page
- or, alternatively, use a custom function to initialise the component which allows you to test the behaviour of `initAll`

Suggest reviewing commit-by-commit.

### DIFF
--- a/lib/jest-helpers.js
+++ b/lib/jest-helpers.js
@@ -13,7 +13,7 @@ const sass = require('node-sass')
 const sassRender = util.promisify(sass.render)
 
 const configPaths = require('../config/paths.json')
-const { componentNameToMacroName } = require('./helper-functions.js')
+const { componentNameToMacroName, componentNameToPascalCase } = require('./helper-functions.js')
 
 const nunjucksEnv = nunjucks.configure([configPaths.src, configPaths.components], {
   trimBlocks: true,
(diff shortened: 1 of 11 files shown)


lib/jest-helpers.js
```diff
@@ -83,6 +83,53 @@ function renderTemplate (params = {}, blocks = {}) {
...
+/**
+ * Render and initialise a component within test boilerplate
+ *
+ * Renders a component's Nunjucks macro with the given params, injects it into
+ * the test boilerplate page, then initialises it with either:
+ *
+ * - the provided JavaScript configuration, used to instanciate the relevant
+ *   component class
+ * - a custom initialiser function (that'll let you run arbitrary code, like
+ *   `initAll` in the browser)
```
domoscargin, line 95: this is a bit unclear: does the custom initialiser function let me run arbitrary code:

* for example, `initAll`
* that looks like `initAll`, but in the browser

Or is the custom initialiser function similar to `initAll`, but in the browser?

lib/jest-helpers.js
```diff
@@ -83,6 +83,53 @@ function renderTemplate (params = {}, blocks = {}) {
...
+ *  passed configuration as `config` and the PascalCased component name as
+ *  `componentClassName`
+ * @returns {Promise}
+ */
+async function renderAndInitialise (componentName, options = {}) {
+  await page.goto(`${options.baseUrl}/tests/boilerplate`, { waitUntil: 'load' })
+
+  const html = renderHtml('notification-banner', options.nunjucksParams)
+
+  // Inject rendered HTML into the slot
```
domoscargin, line 114: First mention of "slot" so maybe it needs clarification?

> Inject rendered HTML into the slot on the boilerplate page

src/govuk/components/notification-banner/notification-banner.test.js
```diff
@@ -30,30 +33,104 @@ describe('/components/notification-banner/with-type-as-success', () => {
...
       const activeElement = await page.evaluate(() => document.activeElement.dataset.module)
 
       expect(activeElement).not.toBe('govuk-notification-banner')
     })
   })
-})
 
-describe('components/notification-banner/role=alert-overridden-to-role=region,-with-type-as-success', () => {
-  describe('when role is not alert', () => {
+  describe('and autofocus is disabled in JS but enabled in data attributes, attributes win', () => {
```
domoscargin, line 109: "attributes win" probably makes this suite easier to parse, but semantically probably belongs in the tests within the description?

Like, `it('is automatically focused when the page loads, as per data attributes')` or something.

I'm not particularly fussed either way.

domoscargin (commented):
Nice!

A few pedantic notes, but this is looking good.

romaricpascal:
Cheers for the review and suggestions @domoscargin. The major notable change in [`d45f013` (#2843)](https://github.com/alphagov/govuk-frontend/pull/2843/commits/d45f013aebe744fd328dc07f6620c37cbef5cc60) besides the bits of docs is the (maybe alittle heavy-handed) change to the `componentNameToPascalCase` change. 

I figured it was worth splitting things properly between:
- pure string manipulation, into a `kebabCaseToPascalCase` function as you described
- component name manipulation, into `componentNameTo...` functions delegating to the appropriate helper (either the kebabCase one or another `componentNameTo...`.

I didn't feel necessary to add extra tests for the `kebabCaseToPascalCase` one, though, as:
- it's not exported, just an internal helper of the module
- it's run by each of the tests of the `componentNameTo...` helpers anyway

domoscargin (approved):
🎉 

This looks great - nice work, folks.

PR #2844, opened 2022-09-14
Deprecate compatibility mode settings

owenatgov:
Fixes https://github.com/alphagov/govuk-frontend/issues/2787

## What/Why
Deprecates the public settings influenced by compatibility mode. This PR specifically targets instances when these settings are being used in user-authored sass outside the bounds of the compatibility mode variables. This scenario is unlikely but means we're covering all bases with our deprecation.

More details can be found in the original issue and descriptions of commits.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,6 +72,19 @@ This will enable screen reader users to have a better, more coherent experience 
 
 This was added in [pull request #2677: Amend error summary markup to fix page load focus bug in JAWS 2022](https://github.com/alphagov/govuk-frontend/pull/2677).
 
+### Deprecated features
+
+#### Stop using settings associated with legacy codebases
+
+In GOV.UK Frontend v5.0 we will stop supporting compatibility with legacy codebases. As part of this, we are deprecating settings controlled by compatibility mode variables. This includes the `govuk-compatibility` mixin and the following settings:
+
+- `$govuk-use-legacy-palette`
+- `$govuk-use-legacy-font`
+- `$govuk-typography-use-rem`
+- `$govuk-font-family-tabular`
+
+This was introduced in [pull request #2844: Deprecate compatibility mode settings](https://github.com/alphagov/govuk-frontend/pull/2844).
+
 ### Fixes
 
 We’ve made fixes to GOV.UK Frontend in the following pull requests:
--- a/src/govuk/settings/_colours-palette.scss
+++ b/src/govuk/settings/_colours-palette.scss
@@ -15,6 +15,8 @@
 ///
 /// @type Boolean
 /// @access public
+/// @deprecated Will be removed in v5.0 with the rest of the compatibility mode
+/// suite of tools and settings
 
 $govuk-use-legacy-palette: if(
   (
@@ -25,6 +27,16 @@ $govuk-use-legacy-palette: if(
   true,
   false
 ) !default;
+
+// Only show the deprecation warning if user is setting $govuk-use-legacy-palette
+// manually instead of automatically via compatibility variables
+@if $govuk-use-legacy-palette == true and
+  $govuk-compatibility-govukfrontendtoolkit == false and
+  $govuk-compatibility-govuktemplate == false and
+  $govuk-compatibility-govukelements == false {
+  @warn "$govuk-use-legacy-palette is deprecated. " +
+    "Only the modern colour palette will be supported from v5.0";
+}
 
 /// Modern colour palette
 ///
--- a/src/govuk/settings/_typography-font.scss
+++ b/src/govuk/settings/_typography-font.scss
@@ -56,12 +56,22 @@ $govuk-font-family: if(
 ///
 /// @type List
 /// @access public
+/// @deprecated Will be removed in v5.0 with the rest of the compatibility mode
+/// suite of tools and settings
 
 $govuk-font-family-tabular: if(
   $govuk-use-legacy-font,
   $govuk-font-family-nta-tabular,
   false
 ) !default;
+
+// Only show the deprecation warning if user is setting $govuk-font-family-tabular
+// manually instead of automatically via $govuk-use-legacy-font
+@if $govuk-font-family-tabular != false and $govuk-use-legacy-font == false {
+  @warn "$govuk-font-family-tabular is deprecated. " +
+    "From version 5.0, GOV.UK Frontend will not support using a separate " +
+    "font-face for tabular numbers";
+}
 
 /// Font families to use for print media
 ///
--- a/src/govuk/settings/_typography-font.scss
+++ b/src/govuk/settings/_typography-font.scss
@@ -13,6 +13,8 @@
 ///
 /// @type Boolean
 /// @access public
+/// @deprecated Will be removed in v5.0 with the rest of the compatibility mode
+/// suite of tools and settings
 
 $govuk-use-legacy-font: if(
   (
@@ -23,6 +25,17 @@ $govuk-use-legacy-font: if(
   true,
   false
 ) !default;
+
+// Only show the deprecation warning if user is setting $govuk-use-legacy-font
+// manually instead of automatically via compatibility variables
+@if $govuk-use-legacy-font == true and
+  $govuk-compatibility-govukfrontendtoolkit == false and
+  $govuk-compatibility-govuktemplate == false and
+  $govuk-compatibility-govukelements == false {
+  @warn "$govuk-use-legacy-font is deprecated. " +
+    "From version 5.0, GOV.UK Frontend will only support the included version " +
+    "of GDS Transport";
+}
 
 // =========================================================
 // Font families
--- a/src/govuk/settings/_typography-responsive.scss
+++ b/src/govuk/settings/_typography-responsive.scss
@@ -13,6 +13,8 @@
 ///
 /// @type Boolean
 /// @access public
+/// @deprecated Will be removed in v5.0 with the rest of the compatibility mode
+/// suite of tools and settings
 
 $govuk-typography-use-rem: if(
   (
@@ -23,6 +25,16 @@ $govuk-typography-use-rem: if(
   false,
   true
 ) !default;
+
+// Only show the deprecation warning if user is setting $govuk-typography-use-rem
+// manually instead of automatically via compatibility variables
+@if $govuk-typography-use-rem == false and
+  $govuk-compatibility-govukfrontendtoolkit == false and
+  $govuk-compatibility-govuktemplate == false and
+  $govuk-compatibility-govukelements == false {
+  @warn "$govuk-typography-use-rem is deprecated. " +
+    "From version 5.0, GOV.UK Frontend will not support disabling rem font sizes";
+}
 
 /// Root font size
 ///
--- a/src/govuk/tools/_compatibility.scss
+++ b/src/govuk/tools/_compatibility.scss
@@ -2,6 +2,20 @@
 /// @group tools/compatibility-mode
 ////
 
+/// Temporary private version of govuk-compatibility to avoid deprecation warnings
+///
+/// @access private
+
+@mixin _govuk-compatibility($product) {
+  @if map-has-key($_govuk-compatibility, $product) {
+    @if map-get($_govuk-compatibility, $product) == true {
+      @content;
+    }
+  } @else {
+    @error "Non existent product '#{$product}'";
+  }
+}
+
 /// Conditional Compatibility Mixin
 ///
 /// Selectively output a block (available to the mixin as @content) if a given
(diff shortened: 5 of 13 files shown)


src/govuk/tools/_compatibility.scss
```diff
@@ -34,3 +36,14 @@
     @error "Non existent product '#{$product}'";
   }
 }
+
+/// Temporary public version of _govuk-compatibility to suppress local warnings
+///
+/// @access public
+
+@mixin govuk-compatibility($product) {
```
NickColley, line 44: Love this approach.

src/govuk/tools/_compatibility.scss
```diff
@@ -34,3 +36,14 @@
     @error "Non existent product '#{$product}'";
   }
 }
+
+/// Temporary public version of _govuk-compatibility to suppress local warnings
+///
+/// @access public
```
36degrees, line 42: This is the Sass doc that will appear in the Sass API reference, so I think we need to maintain the existing docs here but add the deprecation notice (so that the deprecation notice appears in the API reference).

So I'd expect this'd get flipped to something like:

```scss
/// Temporary private version of govuk-compatibility to avoid deprecation warnings
///
/// @access private

@mixin _govuk-compatibility($product) {
  // ...
}

/// Conditional Compatibility Mixin
///
/// Selectively output a block (available to the mixin as @content) if a given
/// $product is also identified as being used in the project.
///
/// This can then be used to include styles that are only needed to override
/// styles provided by those other products (e.g. where govuk_template has a
/// very specific link selector that otherwise affects buttons).
///
/// @example scss
///   // Override .my-class if GOV.UK Template is also being used
///   @include govuk-compatibility(govuk_template) {
///     .my-class {
///       color: inherit;
///     }
///   }
///
/// @param {String} $product - Name of product that we are 'defending' against.
/// @content Passed content is outputted only if Frontend is being used with
///   this product
/// @throw Errors if product name is not recognised
/// @access public
/// @deprecated Will be removed in v5.0 with the rest of the compatibility mode
/// suite of tools and settings

@mixin govuk-compatibility($product) {
  // ...
}
```

src/govuk/tools/_compatibility.scss
```diff
@@ -34,3 +36,14 @@
...
+
+@mixin govuk-compatibility($product) {
+  @warn "govuk-compatibility is deprecated. Support for legacy codebases will be dropped in v5.0";
```
36degrees, line 45: Because the tests still call `govuk-compatibility` we now have warnings in our test output.

We've handled in the past by mocking Sass's `@warn` function, which also allows us to add tests that 'expect' the warning, which means we can be confident it's working as intended.

(We could also update the tests to call the private version of the mixin, **but** then we're not testing the public bit of the API and if we made a mistake in this wrapper function we might not spot it)

See https://github.com/alphagov/govuk-frontend/pull/2156 for an example.

CHANGELOG.md
```diff
@@ -72,6 +72,19 @@ This will enable screen reader users to have a better, more coherent experience
 
 This was added in [pull request #2677: Amend error summary markup to fix page load focus bug in JAWS 2022](https://github.com/alphagov/govuk-frontend/pull/2677).
 
+### Deprecated features
+
+#### Stop using settings associated with legacy codebases
+
+In govuk-frontend v5.0 we will stop supporting compatibility with legacy codebases. As part of this, we are deprecating settings that are controlled by compatibility mode variables. This includes the mixin `govuk-compatibility` and the following variables:
```
claireashworth, line 79:
```suggestion
In govuk-frontend v5.0 we will stop supporting compatibility with legacy codebases. As part of this, we are deprecating settings controlled by compatibility mode variables. This includes the mixin `govuk-compatibility` and the following variables:
```

src/govuk/settings/_typography-font.scss
```diff
@@ -24,6 +26,17 @@ $govuk-use-legacy-font: if(
...
 
+// Only show the deprecation warning if user is setting $govuk-use-legacy-font
+// manually instead of automatically via compatibility variables
+@if $govuk-use-legacy-font == true and
+  $govuk-compatibility-govukfrontendtoolkit == false and
+  $govuk-compatibility-govuktemplate == false and
+  $govuk-compatibility-govukelements == false {
+  @warn "$govuk-use-legacy-font is deprecated. " +
+    "Only the latest version of the GDS Transport font, included with govuk-frontend" +
+    "will be supported from v5.0";
```
claireashworth, line 37:
```suggestion
    "will be supported";
```

src/govuk/settings/_typography-font.scss
```diff
@@ -43,13 +56,22 @@ $govuk-font-family: if(
...
   $govuk-use-legacy-font,
   $govuk-font-family-nta-tabular,
   false
 ) !default;
 
+// Only show the deprecation warning if user is setting $govuk-font-family-tabular
+// manually instead of automatically via $govuk-use-legacy-font
+@if $govuk-font-family-tabular != false and $govuk-use-legacy-font == false {
+  @warn "$govuk-font-family-tabular is deprecated. " +
+    "Support for this method of specifying tabular data will be dropped in v5.0";
```
claireashworth, line 72:
```suggestion
    "From version 5.0, there will be no support for this method of specifying tabular data";
```

src/govuk/settings/_typography-responsive.scss
```diff
@@ -24,6 +26,16 @@ $govuk-typography-use-rem: if(
...
 ) !default;
 
+// Only show the deprecation warning if user is setting $govuk-typography-use-rem
+// manually instead of automatically via compatibility variables
+@if $govuk-typography-use-rem == false and
+  $govuk-compatibility-govukfrontendtoolkit == false and
+  $govuk-compatibility-govuktemplate == false and
+  $govuk-compatibility-govukelements == false {
+  @warn "$govuk-typography-use-rem is deprecated. " +
+    "Govuk-frontend will not support rem value suppression in typography from v5.0";
```
claireashworth, line 36:
```suggestion
    "From version 5.0, Govuk-frontend will not support rem value suppression in typography";
```

src/govuk/tools/_compatibility.scss
claireashworth, line 45:
```suggestion
  @warn "govuk-compatibility is deprecated. From version 5.0, Govuk-frontend will not support legacy codebases";
```

owenatgov:
I just realised I've changed all the warning text but didn't updated the associated tests 🤦🏻 Will sort out.

claireashworth (approved):

owenatgov:
Tests fixed, ready for re-review!

CHANGELOG.md
```diff
@@ -72,6 +72,19 @@ This will enable screen reader users to have a better, more coherent experience
 
 This was added in [pull request #2677: Amend error summary markup to fix page load focus bug in JAWS 2022](https://github.com/alphagov/govuk-frontend/pull/2677).
 
+### Deprecated features
+
+#### Stop using settings associated with legacy codebases
+
+In govuk-frontend v5.0 we will stop supporting compatibility with legacy codebases. As part of this, we are deprecating settings controlled by compatibility mode variables. This includes the mixin `govuk-compatibility` and the following variables:
```
36degrees, line 79:
```suggestion
In GOV.UK Frontend v5.0 we will stop supporting compatibility with legacy codebases. As part of this, we are deprecating settings controlled by compatibility mode variables. This includes the `govuk-compatibility` mixin and the following settings:
```

src/govuk/settings/_colours-palette.scss
```diff
@@ -26,6 +28,16 @@ $govuk-use-legacy-palette: if(
...
 ) !default;
 
+// Only show the deprecation warning if user is setting $govuk-use-legacy-palette
+// manually instead of automatically via compatibility variables
+@if $govuk-use-legacy-palette == true and
+  $govuk-compatibility-govukfrontendtoolkit == false and
+  $govuk-compatibility-govuktemplate == false and
+  $govuk-compatibility-govukelements == false {
+  @warn "$govuk-use-legacy-palette is deprecated. " +
+    "Only the govuk-frontend colour palette will be supported from v5.0";
```
36degrees, line 38:
```suggestion
    "Only the modern colour palette will be supported from v5.0";
```

src/govuk/settings/_typography-font.scss
```diff
@@ -43,13 +56,23 @@ $govuk-font-family: if(
...
   $govuk-font-family-nta-tabular,
   false
 ) !default;
 
+// Only show the deprecation warning if user is setting $govuk-font-family-tabular
+// manually instead of automatically via $govuk-use-legacy-font
+@if $govuk-font-family-tabular != false and $govuk-use-legacy-font == false {
+  @warn "$govuk-font-family-tabular is deprecated. " +
+    "From version 5.0, Govuk-frontend will not support this method of " +
+    "specifying tabular data";
```
36degrees, line 73: This isn't really anything to do with data, suggest something like:

```suggestion
    "From version 5.0, GOV.UK Frontend will not support using a separate " +
    "font-face for tabular numbers";
```

src/govuk/settings/_colours-palette.scss
```diff
@@ -26,6 +28,16 @@ $govuk-use-legacy-palette: if(
   false
 ) !default;
 
+// Only show the deprecation warning if user is setting $govuk-use-legacy-palette
+// manually instead of automatically via compatibility variables
+@if $govuk-use-legacy-palette == true and
+  $govuk-compatibility-govukfrontendtoolkit == false and
+  $govuk-compatibility-govuktemplate == false and
+  $govuk-compatibility-govukelements == false {
```
36degrees, line 36: If this was moved above the definition I think you could just do…

```scss
@if variable-exists($govuk-use-legacy-palette) {
```

As the only way the variable would exist at that point is if it was set in userspace.

(Same applies for the other settings)

src/govuk/settings/_typography-font.scss
```diff
@@ -24,6 +26,17 @@ $govuk-use-legacy-font: if(
...
 
+// Only show the deprecation warning if user is setting $govuk-use-legacy-font
+// manually instead of automatically via compatibility variables
+@if $govuk-use-legacy-font == true and
+  $govuk-compatibility-govukfrontendtoolkit == false and
+  $govuk-compatibility-govuktemplate == false and
+  $govuk-compatibility-govukelements == false {
+  @warn "$govuk-use-legacy-font is deprecated. " +
+    "From version 5.0, we will only support the latest version of the " +
+    "GDS Transport font, included with govuk-frontend";
```
36degrees, line 37: For brevity and clarity:

```suggestion
    "From version 5.0, GOV.UK Frontend will only support the included version " + 
    "of GDS Transport"
```

src/govuk/settings/_typography-responsive.scss
```diff
@@ -24,6 +26,16 @@ $govuk-typography-use-rem: if(
...
 ) !default;
 
+// Only show the deprecation warning if user is setting $govuk-typography-use-rem
+// manually instead of automatically via compatibility variables
+@if $govuk-typography-use-rem == false and
+  $govuk-compatibility-govukfrontendtoolkit == false and
+  $govuk-compatibility-govuktemplate == false and
+  $govuk-compatibility-govukelements == false {
+  @warn "$govuk-typography-use-rem is deprecated. " +
+    "From version 5.0, Govuk-frontend will not support rem value suppression in typography";
```
36degrees, line 36:
```suggestion
    "From version 5.0, GOV.UK Frontend will not support disabling rem font sizes";
```

src/govuk/tools/_compatibility.scss
```diff
@@ -24,13 +38,13 @@
...
 
 @mixin govuk-compatibility($product) {
-  @if map-has-key($_govuk-compatibility, $product) {
-    @if map-get($_govuk-compatibility, $product) == true {
-      @content;
-    }
-  } @else {
-    @error "Non existent product '#{$product}'";
+  @warn "govuk-compatibility is deprecated. " +
+    "From version 5.0, Govuk-frontend will not support legacy codebases";
```
36degrees, line 46: I think 'legacy codebases' is too ambiguous here. Can we just say we're getting rid of compatibility mode?

```suggestion
    "From version 5.0, GOV.UK Frontend will not support compatibility mode";
```

36degrees (commented):
Technical implementation looks solid – thanks @owenatgov :cool:

I think some of the deprecation notices could be clearer, so I've made a some suggestions. I'd also personally prefer we use the full product name ('GOV.UK Frontend') where possible.

src/govuk/settings/_colours-palette.scss
owenatgov: Sadly I haven't been able to get this to work. It seems that `variable-exists` and `global-variable-exists` both scope existing variables globally and doesn't cascade, meaning that even if I move this above the definition, it still exists in the sass as a whole so it's still true locally. This also applies if I add `and $govuk-use-legacy-palette == true` as this seems to be informed by the compatibility mode variables still, meaning that I still have to check that the compatibility mode variables are false or get bombarded by warnings when running locally. I don't know if I've done something wrong here or if `variable-exists` is too loose for this use case.

36degrees (approved):

src/govuk/tools/_compatibility.scss
```diff
@@ -24,13 +38,13 @@
...
 
 @mixin govuk-compatibility($product) {
-  @if map-has-key($_govuk-compatibility, $product) {
-    @if map-get($_govuk-compatibility, $product) == true {
-      @content;
-    }
-  } @else {
-    @error "Non existent product '#{$product}'";
+  @warn "govuk-compatibility is deprecated. " +
+    "From version 5.0, Govuk-frontend will not support compatibility mode";
```
36degrees, line 46:
```suggestion
    "From version 5.0, GOV.UK Frontend will not support compatibility mode";
```

PR #2847, opened 2022-09-14
Tidy up the `initAll` function

36degrees:
Tidy up the `initAll` function and add JSDoc. This shouldn't have any external-facing changes.

**Builds on #2843 which needs to be merged first.**


domoscargin (approved):
Looks good to me!

PR #2848, opened 2022-09-14
Add Git hook to lint JavaScript and Sass file `pre-commit`

romaricpascal:
## What

This PR adds automatic linting of code being commited before `git push`.

The relevant linters will run on any JavaScript and/or Scss being commited, and only what's being commited. The rest is stashed temporarily while the linting happens to ensure irrelevant files or [hunks][git-hunks] don't get in the way. The linters will also try to fix any issue they can fix automatically to make the commit only fail if a human intervention is needed.

The `pre-commit` hook get installed automatically on `npm install` (which we'll need to run after this gets merged in `main`). It [can be bypassed if really getting in the way of a commit that really needs to go][husky-bypass] (though like with disabling linters, best to have a good reason for it... and Github CI will fail anyway so might as well save yourself a round-trip). 

As a side bonus, you can also lint your staged files without commiting by running `npx lint-staged` after staging content.

[husky-bypass]: https://typicode.github.io/husky/#/?id=bypass-hooks
[git-hunks]: https://git-scm.com/docs/git-add#Documentation/git-add.txt---patch

## Why

Linting pre-commit reduces the chance of losing time to only detecting linting issues once Github CI run and is quick enough to not burden the time it takes to commit. Most our editors can likely run ESLint on save, but there's no guarantee everyone has something like that set up so that a little extra safety that gets automatically set up when installing the project dependencies.

Note that it won't run in case of a quick edit-and-commit on Github and we'll have to go fix things manually should the CI pick up a linting issue after such edit.

## How

It relies on two tools:
- [lint-staged](https://github.com/okonet/lint-staged) which handles detecting which files are being commited and running the linting on them. It stashes/unstashes any changes that are not staged so the linting runs on the files in the state they are actually commited, in case you're using `git add -p`.
   I thought there'd be some caution to have around the `package` and `dist` folder to avoid messing with compiled files, but `standard` is already configured to avoid those.
- [husky](https://typicode.github.io/husky) which handles the creation and installation of Git hooks, mostly pointing Git to the relevant scripts in `.husky` (configurable) through its `core.hooksPath` config.

## Possible future developments

With [Git hooks][git-hooks] available, there's potentially other automations we could look into, like:
- use [`post-checkout`][git-hooks-post-checkout] to detect if things like `npm install` need to run and warn
- use [`pre-push`]() to run tests locally before hitting CI (not sure if that's a great one though)
- possibly other stuff

As linters get run automatically, we could look into linting other kinds of files. I'm thinking Markdown specifically, both for its structure with something like [remark-lint] and for its content with something like [vale.sh] that Claire mentionned recently.

[git-hooks]: https://git-scm.com/docs/githooks#_pre_push
[git-hooks-post-checkout]: https://git-scm.com/docs/githooks#_post_checkout
[git-hooks-pre-push]: https://git-scm.com/docs/githooks#_pre_push
[remark-lint]: https://github.com/remarkjs/remark-lint
[vale.sh]: https://vale.sh/

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,4 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+
+npx --no-install lint-staged
--- a/.lintstagedrc.js
+++ b/.lintstagedrc.js
@@ -1,0 +1,6 @@
+module.exports = {
+  // ESLint's configuration will let it ignore built files
+  // in `dist` or `package` that we want left alone, as well as the polyfills
+  '*.{cjs,js,mjs}': ['npm run lint:js:cli -- --fix'],
+  '*.scss': ['npm run lint:scss:cli -- --fix']
+}
--- a/package.json
+++ b/package.json
@@ -34,7 +34,8 @@
     "lint:js": "npm run lint:js:cli -- \"**/*.{cjs,js,mjs}\"",
     "lint:js:cli": "eslint --cache --cache-location .cache/eslint --cache-strategy content --color --ignore-path .gitignore --max-warnings 0",
     "lint:scss": "npm run lint:scss:cli -- \"{app,src}/**/*.scss\"",
-    "lint:scss:cli": "stylelint --cache --cache-location .cache/stylelint --cache-strategy content --color --ignore-path .gitignore --max-warnings 0"
+    "lint:scss:cli": "stylelint --cache --cache-location .cache/stylelint --cache-strategy content --color --ignore-path .gitignore --max-warnings 0",
+    "prepare": "node -e \"try { require('husky').install() } catch (e) {if (e.code !== 'MODULE_NOT_FOUND') throw e}\""
   },
   "dependencies": {
     "@babel/core": "^7.20.2",
@@ -87,6 +88,7 @@
     "eslint-plugin-n": "^15.5.1",
     "eslint-plugin-promise": "^6.1.1",
     "html-validate": "7.10.0",
+    "husky": "^8.0.2",
     "jest": "^29.3.1",
     "jest-axe": "^7.0.0",
     "jest-dev-server": "^6.1.1",
--- a/package.json
+++ b/package.json
@@ -31,8 +31,10 @@
     "test": "jest --color --ignoreProjects \"Gulp tasks\" --maxWorkers=50%",
     "test:screenshots": "node ./tasks/screenshot-components.mjs",
     "lint": "npm run lint:js && npm run lint:scss",
-    "lint:js": "eslint --cache --cache-location .cache/eslint --cache-strategy content --color --ignore-path .gitignore --max-warnings 0 \"**/*.{cjs,js,mjs}\"",
-    "lint:scss": "stylelint --cache --cache-location .cache/stylelint --cache-strategy content --color --ignore-path .gitignore --max-warnings 0 \"{app,src}/**/*.scss\""
+    "lint:js": "npm run lint:js:cli -- \"**/*.{cjs,js,mjs}\"",
+    "lint:js:cli": "eslint --cache --cache-location .cache/eslint --cache-strategy content --color --ignore-path .gitignore --max-warnings 0",
+    "lint:scss": "npm run lint:scss:cli -- \"{app,src}/**/*.scss\"",
+    "lint:scss:cli": "stylelint --cache --cache-location .cache/stylelint --cache-strategy content --color --ignore-path .gitignore --max-warnings 0"
   },
   "dependencies": {
     "@babel/core": "^7.20.2",
@@ -92,6 +94,7 @@
     "jest-environment-node": "^29.3.1",
     "jest-environment-puppeteer": "^6.1.1",
     "jest-serializer-html": "^7.1.0",
+    "lint-staged": "^13.0.3",
     "outdent": "^0.8.0",
     "puppeteer": "^19.2.2",
     "sass-color-helpers": "^2.1.1",
--- a/package-lock.json
+++ b/package-lock.json
@@ -70,6 +70,7 @@
         "jest-environment-node": "^29.3.1",
         "jest-environment-puppeteer": "^6.1.1",
         "jest-serializer-html": "^7.1.0",
+        "lint-staged": "^13.0.3",
         "outdent": "^0.8.0",
         "puppeteer": "^19.2.2",
         "sass-color-helpers": "^2.1.1",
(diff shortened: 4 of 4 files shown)


.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
```
colinrotherham, line 2: Not sure if this is likely (yet) but we've got shell scripts that call `git commit`

Do we want to skip the hooks when `CI=true` is set in the GitHub workflows?

```suggestion
. "$(dirname -- "$0")/_/husky.sh"

# Ignore hook in GitHub workflows
[ -n "$CI" ] && exit 0
```

.lintstagedrc.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
```
colinrotherham, line 1: Love this

Will go really nicely with https://github.com/alphagov/govuk-frontend/pull/2849 with `standard --fix` → `eslint --fix`

Think we should stick to the ESM route (versus CommonJS) and make this **.lintstagedrc.mjs**?

.lintstagedrc.js
```diff
@@ -0,0 +1,6 @@
...
+  // standard's configuration will let it ignore built files
+  // in `dist` or `package` that we want left alone, as well as the polyfills
+  '*.{js,mjs}': ['standard --fix'],
```
colinrotherham, line 4: Thoughts on keeping things in one place, and only appending `-- --fix` to the existing npm scripts?

On other projects I've had "default" scripts, plus others ready to accept arguments:

```json
"lint:js": "npm run lint:js:cli -- \"**/*.{cjs,js,mjs}\"",
"lint:js:cli": "standard",
```

```suggestion
  '*.{cjs,js,mjs}': ['npm run lint:js -- --fix'],
```

.husky/pre-commit
romaricpascal: That'd likely save some execution time on those commits on the CI. I'd prefer doing that in a blanket way, either using the [`HUSKY=0` env variable](https://typicode.github.io/husky/#/?id=with-env-variables) to not have to manage each script individually. 

If we ever find some checks that need to run before CI Git operations, I think it'd be better that they're explicitely shown in what the CI does rather than happening through git hooks anyway.

.lintstagedrc.js
romaricpascal: I'll update now the switch to `eslint` is merged.

Regarding moving to ESM, though, I'm not 100% sure. Most our config files are CJS by the looks of it and it'd make more sense to me to have them all in the same format. I'm not strongly attached to it, though.

.husky/pre-commit
colinrotherham: Yeah don't mind how, just know we have `git config` and `git diff` calls here:
https://github.com/alphagov/govuk-frontend/blob/git-hook-linting/.github/workflows/diff-change-to-dist.yaml

They don't follow an npm install so just being safe. Cheers @romaricpascal

.lintstagedrc.js
colinrotherham: I tend to use ESM by default now, especially with lots of packages going "ESM only"

No rush to switch just for this one config though

.lintstagedrc.js
romaricpascal: I'm guessing you meant using `npm run lint:js:cli` in the `lintstagedrc`. No strong opinion on that one for me, just a bit warry of adding levels of indirection before a command actually runs. It does make updates more centralised, though, which is nice 😄 

Cheers for picking up the lack of `cjs` extension in the `lintstagedrc`, that'll future-proof that file (though unless we move to having`type="module"` in package.json, `.cjs` files shouldn't appear in the codebase, should they?) 😄

colinrotherham: I did indeed 🤦‍♂️ 

They shouldn't need to appear no, but if they do they won't be linted on commit 😉 

Mainly for completeness that "All JavaScript will be linted" (it's in our ESLint config too)

.husky/pre-commit
romaricpascal: I don't think those would trigger any hooks, [judging by the list in Git docs](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks). I'm still keep to deactivate Husky's hooks altogether on CI, though. Cheers for pointing the use of Git out 😄

.lintstagedrc.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+  // ESLint's configuration will let it ignore built files
+  // in `dist` or `package` that we want left alone, as well as the polyfills
+  // Not using `lint:js` or `lint:scss` as they ran slower than the raw linter
+  // (maybe because they use their cache while we only lint a handful of files)
+  '*.{cjs,js,mjs}': ['eslint --max-warnings 0 --fix'],
+  '*.scss': ['stylelint --max-warnings 0 --fix']
```
colinrotherham, line 7: Wondering if the config looks expressive enough to set `FILES=` environment variable

Which we could pass into the original npm scripts?

```console
FILES=abc.mjs npm run lint:js
FILES=def.scss npm run lint:scss
```

You'd add defaults to the npm script in **package.json** (scroll right)

```json
"lint:js": "eslint --cache --cache-location .cache/eslint --cache-strategy content --color --ignore-path .gitignore --max-warnings 0 ${FILES:-\"**/*.{cjs,js,mjs}\"}",
"lint:scss": "stylelint --cache --cache-location .cache/stylelint --cache-strategy content --color --ignore-path .gitignore --max-warnings 0 ${FILES:-\"{app,src}/**/*.scss\"}",
```

`${FILES:-\"**/*.{cjs,js,mjs}\"}`
`${FILES:-\"{app,src}/**/*.scss\"}`

Otherwise we're missing these flags making the hook a bit slow (no cache):

```console
--cache --cache-location .cache/eslint --cache-strategy content --color --ignore-path .gitignore
```

colinrotherham: It's a JavaScript config so you could read the **package.json** and grab the `lint:js`/`lint:scss` scripts?

```js
const pkg = require(join(__dirname, 'package.json'))
```

Take a look at their [JavaScript function example](https://github.com/okonet/lint-staged#example-ignore-files-from-match):

```js
// lint-staged.config.js
import micromatch from 'micromatch'

export default {
  '*.js': (files) => {
    // from `files` filter those _NOT_ matching `*test.js`
    const match = micromatch.not(files, '*test.js')
    return `eslint ${match.join(' ')}`
  },
}
```

Quick string replacement (watch out for escaping) and that does the job?

colinrotherham: Or revisit a separate "CLI" version of the npm script suggestion from https://github.com/alphagov/govuk-frontend/pull/2848#discussion_r971810349?

```json
"lint:js": "npm run lint:js:cli -- \"**/*.{cjs,js,mjs}\"",
"lint:js:cli": "eslint --cache --cache-location .cache/eslint --cache-strategy content --color --ignore-path .gitignore --max-warnings 0",
"lint:scss": "npm run lint:scss:cli -- \"{app,src}/**/*.scss\"",
"lint:scss:cli": "stylelint --cache --cache-location .cache/stylelint --cache-strategy content --color --ignore-path .gitignore --max-warnings 0"
```

So the hooks can run the CLI version, bypassing the default files arg

```suggestion
  '*.scss': ['npm run lint:scss:cli -- --fix']
```

package.json
```diff
@@ -31,8 +31,11 @@
     "test": "jest --color --ignoreProjects \"Gulp tasks\" --maxWorkers=50%",
     "test:screenshots": "node ./tasks/screenshot-components.mjs",
     "lint": "npm run lint:js && npm run lint:scss",
-    "lint:js": "eslint --cache --cache-location .cache/eslint --cache-strategy content --color --ignore-path .gitignore --max-warnings 0 \"**/*.{cjs,js,mjs}\"",
-    "lint:scss": "stylelint --cache --cache-location .cache/stylelint --cache-strategy content --color --ignore-path .gitignore --max-warnings 0 \"{app,src}/**/*.scss\""
+    "lint:js:cli": "eslint --cache --cache-location .cache/eslint --cache-strategy content --color --ignore-path .gitignore --max-warnings 0",
+    "lint:js": "npm run lint:js:cli -- \"**/*.{cjs,js,mjs}\"",
```
colinrotherham, line 35: @romaricpascal Sorry to be a pain, but could you move the `lint:js:cli` task below `lint:js`?

Same with `lint:scss` followed by `lint:scss:cli`

Can you tell I've been looking at [**eslint-plugin-json-files**](https://github.com/kellyselden/eslint-plugin-json-files#readme) for `package.json` linting? 🤦‍♂️

.lintstagedrc.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+  // ESLint's configuration will let it ignore built files
+  // in `dist` or `package` that we want left alone, as well as the polyfills
+  '*.{cjs,js,mjs}': ['npm run lint:js:cli -- --fix'],
+  '*.scss': ['npm run lint:scss:cli --fix']
```
colinrotherham, line 5: Ahh we've missed the extra `--` flag here

```suggestion
  '*.scss': ['npm run lint:scss:cli -- --fix']
```

colinrotherham (commented):
Approved with two little comment

One's a blocker but I'm sure you'll fix it before merging 😆 

Cheers @romaricpascal

package.json
```diff
@@ -31,8 +31,11 @@
     "test": "jest --color --ignoreProjects \"Gulp tasks\" --maxWorkers=50%",
     "test:screenshots": "node ./tasks/screenshot-components.mjs",
     "lint": "npm run lint:js && npm run lint:scss",
-    "lint:js": "eslint --cache --cache-location .cache/eslint --cache-strategy content --color --ignore-path .gitignore --max-warnings 0 \"**/*.{cjs,js,mjs}\"",
-    "lint:scss": "stylelint --cache --cache-location .cache/stylelint --cache-strategy content --color --ignore-path .gitignore --max-warnings 0 \"{app,src}/**/*.scss\""
+    "lint:js": "npm run lint:js:cli -- \"**/*.{cjs,js,mjs}\"",
+    "lint:js:cli": "eslint --cache --cache-location .cache/eslint --cache-strategy content --color --ignore-path .gitignore --max-warnings 0",
+    "lint:scss": "npm run lint:scss:cli -- \"{app,src}/**/*.scss\"",
+    "lint:scss:cli": "stylelint --cache --cache-location .cache/stylelint --cache-strategy content --color --ignore-path .gitignore --max-warnings 0",
+    "prepare": "node -e \"try { require('husky').install() } catch (e) {if (e.code !== 'MODULE_NOT_FOUND') throw e}\""
```
romaricpascal, line 38: This automatically installs the Git hooks [when you `npm install`](https://docs.npmjs.com/cli/v6/using-npm/scripts#life-cycle-scripts). This is [the lifecycle script recommended by Husky's documentation](https://typicode.github.io/husky/#/?id=install) (and we already have `postinstall` anyway).

We cannot run the recommended `husky install`, though. `husky` is a `devDependency`, as it's not needed to run the review app. Because of that, it won't be installed on Heroku, (or when running `npm install --omit=dev`) which means that `prepare` would fail if we were to run `husky install`. Instead [we can rely on the documented exception from Node](https://nodejs.org/api/modules.html#file-modules) to ensure a smooth installation, as proposed by [Husky's docs](https://typicode.github.io/husky/#/?id=with-a-custom-script).

colinrotherham (approved):
Approving again

You know what would be amazing?

Updating the `gulp.watch()` lint tasks so they _also_ only lint the "changed files", rather than every single JavaScript and Sass file

Issue #2009, opened 2020-11-06
Standard 16.0.0+ flags use of var instead of let/const


domoscargin:
## Problem
[Digital Marketplace's components repo](https://github.com/alphagov/digitalmarketplace-govuk-frontend) is built similarly to `govuk-frontend`. We recently bumped standard to 16.0.1 via dependabot and now get a series of warnings:

`Unexpected var, use let or const instead.`

Bumping standard on `govuk-frontend` results in a similar set of warnings.

## Cause
standard have implemented the `no-var` rule as of v16.0.0: https://standardjs.com/changelog.html#1601---2020-10-30
This is currently a warning, but will become an error in the future.

kevindew:
The approach we've taken on GOV.UK is to switch to [standardx](https://github.com/standard/standardx) and disabling the no-var rule.  Further details are available in https://github.com/alphagov/govuk_publishing_components/pull/1785

PR #2849, opened 2022-09-15
Add ESLint config for JavaScript Standard Style

colinrotherham:
This PR upgrades from `standard@14.3.4` to `standard@17` which closes https://github.com/alphagov/govuk-frontend/issues/2009

This means we can start using [Top-level `await`](https://v8.dev/features/top-level-await), [Nullish coalescing](https://v8.dev/features/nullish-coalescing) etc without [**standard**](https://standardjs.com) flagging errors/warnings. I've also added an **.eslintrc.js** config so IDEs and editors can see issues visually whilst coding

Plus you can turn on nice things like "auto fix on save" 😊

There have been minor rule changes to [`object-shorthand`](https://eslint.org/docs/latest/rules/object-shorthand) and [`no-var`](https://eslint.org/docs/latest/rules/no-var) but we can configure these per directory. We're also following the [eslint-config-standard **README.md**](https://github.com/standard/eslint-config-standard#readme) and adding the following ESLint plugins:

```
"eslint-config-standard": "^17.0.0",
"eslint-plugin-import": "^2.26.0",
"eslint-plugin-n": "^15.2.5",
"eslint-plugin-promise": "^6.0.1",
```


.eslintrc.js
```diff
@@ -0,0 +1,21 @@
+module.exports = {
+  extends: 'standard',
+  ignorePatterns: [
+    '!.*',
+    'dist/**/*',
+    'node_modules',
+    'node_modules/.*',
```
romaricpascal, line 7: I think this is unnnecessary as ESLint already ignores `node_modules` ([buried in this page of their doc](https://eslint.org/docs/latest/user-guide/configuring/ignoring-code)):

>In addition to any patterns in the .eslintignore file, ESLint always follows a couple of implicit ignore rules even if the --no-ignore flag is passed. The implicit rules are as follows:
>
>  -  node_modules/ is ignored.
>  -  dot-files (except for .eslintrc.*), as well as dot-folders and their contents, are ignored.

lib/.eslintrc.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+  env: {
+    jest: true
```
romaricpascal, line 3: I'm not sure the whole folder will only run in Jest. My understanding is that some methods in the `helper-functions.js` file will be run in a plain `Node` environment while `Gulp` runs. If we say they run in `jest` we risk allowing globals that may not be available when they run some otherway (though we'll likely uncover those when we try to build the project anyway).

romaricpascal (commented):
That'd be neat, I'd welcome the swap to ESLint for checking standard 😄 Just noticed two little things on the configuration (see comments).

lib/.eslintrc.js
domoscargin: Yeah, `file-helper.js` has some stuff that's used outside of `jest` as well.

.eslintrc.js
colinrotherham: Ahh sorry, I should explain

See a bit further up I've added `'!.*'`?

It's to override ESLint's "ignore dot-files by default" logic so we can lint files such as:

```
.eslintrc.js
.lintstagedrc.js
.babelrc.js
```

But then sadly it seems to then include "dot files inside node_modules" which made this combo necessary:

```
'node_modules',
'node_modules/.*',
```

Glad you spotted it though, I love a good code review 😊

lib/.eslintrc.js
colinrotherham: Good feedback, I'll narrow this down

.eslintrc.js
```diff
@@ -0,0 +1,21 @@
+module.exports = {
+  extends: 'standard',
+  ignorePatterns: [
```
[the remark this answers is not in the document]
colinrotherham: This is good, thank you. Watch this space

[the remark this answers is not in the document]
colinrotherham: @domoscargin I've instead added `--ignore-path .gitignore` to the npm script

Lots of good stuff in there we might as well share it

We can still keep `ignorePatterns` for "not ignored but don't lint" build output files

package.json
```diff
@@ -34,7 +34,11 @@
     "cross-env": "^7.0.3",
     "cssnano": "^4.1.11",
     "del": "^6.1.1",
-    "eslint": "^6.8.0",
+    "eslint": "^8.23.1",
+    "eslint-config-standard": "^17.0.0",
```
domoscargin, line 38: Could we reflect that we're also adding these new packages and give a brief explanation why in the commit message?

colinrotherham: ✅ PR description updated https://github.com/alphagov/govuk-frontend/pull/2849#issue-1374123200

lib/.eslintrc.js
```diff
@@ -0,0 +1,11 @@
+module.exports = {
+  overrides: [
+    {
+      files: ['jest-*'],
```
domoscargin, line 4: I think `axe-helper.js` is also only ever used with jest. We could potentially rename it `jest-axe-helper.js` to reflect this, and match this rule.

colinrotherham: Feels like we should move the `axe` function into **jest-helpers.js**

Because it's a jest helper (??)

colinrotherham: @domoscargin New commit added: https://github.com/alphagov/govuk-frontend/pull/2849/commits/a4956a62d7ee648297d5d85d956c787b3a70b6a4

package.json
```diff
@@ -20,7 +20,7 @@
     "test:build:package": "jest tasks/gulp/__tests__/after-build-package.test.js",
     "test:build:dist": "jest tasks/gulp/__tests__/after-build-dist.test.js",
     "lint": "npm run lint:js && npm run lint:scss",
-    "lint:js": "standard",
+    "lint:js": "eslint --cache --cache-location .cache/eslint --cache-strategy content --color --ignore-path .gitignore \"**/*.{cjs,js,mjs}\"",
```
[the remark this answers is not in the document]
colinrotherham: Yeah, it absolutely pains me too 😔 

Sadly, none of these flags can be configured via `.eslintrc.js`

**`--cache`**
Don't lint "already checked" files

**`--cache-strategy content`**
Determine "already checked" files when content (or rules) hasn't changed

**`--cache-location .cache/eslint`**
Unifies the cache location with other CLI tools

**`--color`**
Enables colour output for CI

**`--ignore-path .gitignore`**
Unifieds ignored files with other CLI tools

Most of these exist for stylelint too. What shall we do?

package.json
domoscargin: Thanks for that!

I think what I'm trying to guard against is the case we've hit a coupla times, where we've had an unneeded dependency lying around because we made a change and didn't realise it was no longer needed. These plugins feel a bit "hidden" because they're not being directly imported anywhere in our code. I'm not sure if this is a big deal, or me just being overly nervous.

lib/.eslintrc.js
domoscargin: I refuse to entertain something which MAKES SO MUCH SENSE.

Nice!

package.json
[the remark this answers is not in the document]
domoscargin: The only thing that comes to mind off the top of my head is to run it with a bash script (like `build-release` for example), 'cuz we could then at least document it a bit. But that doesn't feel super great.

Oh, actually, one thing we should do is start updating this file: https://github.com/alphagov/govuk-frontend/blob/main/docs/contributing/tasks.md

We could document this stuff there.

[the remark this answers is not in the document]
domoscargin: Yeah, I'd lean towards keeping it as is and updating those docs.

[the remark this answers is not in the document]
colinrotherham: Hey @domoscargin I had a look at [docs/contributing/tasks.md](https://github.com/alphagov/govuk-frontend/blob/main/docs/contributing/tasks.md) and it's pretty out of date

Mind if we wrap up this PR and I'll update those docs via https://github.com/orgs/alphagov/projects/41?

Think we need anything else?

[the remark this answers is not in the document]
domoscargin: Nope, that sounds good to me!

domoscargin (approved):
This looks good to me!

Issue #2643, opened 2022-05-24
4.1.0 breaks require.resolve to find all.js client javascript


HughePaul:
## Description of the issue
The new export block in the package.json breaks the ability to look for the main module using `require.resolve`

This should probably look more like:
```
  "exports": {
    "require": "./govuk/all.js",
    "import": "./govuk-esm/all.mjs"
  },
```
## Steps to reproduce the issue
```
require.resolve('govuk-frontend');
```

```
node:internal/modules/cjs/loader:488
      throw e;
      ^
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: No "exports" main defined in ./node_modules/govuk-frontend/package.json
    at new NodeError (node:internal/errors:371:5)
```

## Actual vs expected behaviour
Returns full path location of all.js compiled client javascript

## Environment (where applicable)

- Operating system: ubunu 18.04
- Browser: N/A
- Browser version: N/A
- GOV.UK Frontend Version: 4.1.0
- NodeJS Verson: 16.x.x

36degrees:
@HughePaul you mention `app.js` in the issue title and in the actual vs expected behaviour section, but `app.js` is part of the tooling in the repo and is not published as part of the package. Just to double check, I'm assuming you meant `all.js` – is that correct?

HughePaul:
yes that branch works as expected, eg:

```
$ node -e 'console.log(path.relative(".", require.resolve("govuk-frontend")))'
node_modules/govuk-frontend/govuk/all.js
```

PR #2850, opened 2022-09-15
Improvements to Jest project setup + test package updates

colinrotherham:
Lots of little fixes in this PR

1. Package updates for Jest, Percy, Puppeteer
2. Fixed Percy `window.fetch()` (newer versions assume a browser or jsdom environment)
3. Added caching for Jest rollup transforms (speeds up tests)
4. Only start Puppeteer (slow) for tests that actually need it
5. ~~Remove deprecated [`request`](https://github.com/request/request/issues/3142) package~~ https://github.com/alphagov/govuk-frontend/pull/2858

You can now reliably use the Jest CLI to run single tests, filters or target projects:

```console
npx jest --selectProjects 'Snapshot tests'
npx jest --selectProjects 'JavaScript unit tests'
npx jest --selectProjects 'JavaScript behaviour tests'
npx jest --selectProjects 'JavaScript component tests'
```

☝️ The test site (and puppeteer) will now start/stop automatically when needed

### Jest with `--watch` flag
For example, this PR resolves various Jest issues such as:

```console
Error: listen EADDRINUSE: address already in use :::8888
    at Server.setupListenHandle [as _listen2] (node:net:1432:16)
    at listenInCluster (node:net:1480:12)
    at Server.listen (node:net:1568:7)
    at Function.listen (/path/to/project/govuk-frontend/node_modules/express/lib/application.js:618:24)
    at Object.<anonymous> (/path/to/project/govuk-frontend/app/start.js:6:5)
    at Module._compile (node:internal/modules/cjs/loader:1126:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1180:10)
    at Module.load (node:internal/modules/cjs/loader:1004:32)
    at Function.Module._load (node:internal/modules/cjs/loader:839:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:81:12)
Emitted 'error' event on Server instance at:
    at emitErrorNT (node:net:1459:8)
    at processTicksAndRejections (node:internal/process/task_queues:83:21) {
  code: 'EADDRINUSE',
  errno: -48,
  syscall: 'listen',
  address: '::',
  port: 8888
}
```

These can be shown by adding `--detectOpenHandles` as suggested:

```
Jest did not exit one second after the test run has completed.

This usually means that there are asynchronous operations that weren't stopped in your tests. Consider running Jest with `--detectOpenHandles` to troubleshoot this issue.
```


jest.config.js
```diff
@@ -1,29 +1,70 @@
 const config = {
+  cacheDirectory: '<rootDir>/.cache/jest/',
+  testPathIgnorePatterns: [
+    '/node_modules/',
+    'after-*'
```
36degrees, line 5: I think this has broken the `test:build:package` and `test:build:dist` tasks:

```
govuk-frontend on  upgrade-test-jest-percy [$!] via ⬢ v16.17.0 took 20s 
➜ npm run test:build:package      

> test:build:package
> jest tasks/gulp/__tests__/after-build-package.test.js

No tests found, exiting with code 1
Run with `--passWithNoTests` to exit with code 0
In /Users/oliver.byford/Code/govuk-frontend
  1165 files checked across 4 projects. Run with `--verbose` for more details.
Pattern: tasks/gulp/__tests__/after-build-package.test.js - 0 matches

govuk-frontend on  upgrade-test-jest-percy [$!] via ⬢ v16.17.0 took 3s 
➜ npm run test:build:dist   

> test:build:dist
> jest tasks/gulp/__tests__/after-build-dist.test.js

No tests found, exiting with code 1
Run with `--passWithNoTests` to exit with code 0
In /Users/oliver.byford/Code/govuk-frontend
  1165 files checked across 4 projects. Run with `--verbose` for more details.
Pattern: tasks/gulp/__tests__/after-build-dist.test.js - 0 matches
```

The desired behaviour is that the tests in `tasks/gulp/__tests__/` are excluded from running as part of `npm test` but we can call them as part of the `build:package` and `build:dist` tasks to check that the package / dist are what we expect.

(I suspect there's a better way to do this! Maybe we can use the Jest projects config?)

36degrees (commented):
Thanks @colinrotherham!

I can generally understand the changes except for the last commit – I don't really understand what it's fixing or how the changes relate to each other. Would you mind adding a bit more detail to the commit message that explains what's going on, for future us?

jest.config.js
colinrotherham: Don't worry, this is what code reviews are for 😊 

I'll find a better way of grouping those excluded-but-not-excluded tests—probably Jest config yep!

src/govuk/components/all.test.js
```diff
@@ -43,6 +43,14 @@ it('_all.scss renders to CSS without errors', () => {
 })
 
 describe.each(allComponents)('%s', (component) => {
+  let percySnapshot
+
+  beforeAll(() => {
+    // Support fetch() detection, upload via WebSocket()
+    global.window = { fetch, WebSocket }
```
colinrotherham, line 50: @36degrees To answer the "Why do we need this?"

For context, Jest used to run with **jsdom** by default

**Performance improvements**
https://jestjs.io/blog/2022/04/25/jest-28
https://jestjs.io/blog/2021/05/25/jest-27

Jest now runs all tests via a Node.js environment. No browser globals

The problem? Percy code is required (by Jest) as a ["browserified" bundle](https://github.com/percy/cli/blob/90cd2f3a5d83f9835fcb975751cb828262fb95f0/packages/core/package.json#L22), so hits this code path:
https://github.com/percy/cli/blob/19f8be4903543544bdef197f2f7d5a6a648ed37d/packages/sdk-utils/src/logger.js#L74

So this _restores_ both browser globals and fixes:

1. **Fetch API** for Percy to see if it's CLI agent is running
2. **WebSocket** for Percy to upload screenshots securely

36degrees: I have been wondering about splitting the Percy snapshotting out as it's not really part of the 'tests' and doesn't rely on anything else from Jest.

I suspect we might also need to do this if we wanted to conditionally run the visual regression tests only if certain files have changed.

Until then, to stop us having to go down that rabbit hole too far, would another way to solve this be to split the snapshotting out into its own Jest test case?

colinrotherham: @36degrees Yeah, appreciate this is a quick fix. Did see the new projects list?

```console
npx jest --selectProjects "JavaScript component tests"
```

Likewise we can run everything except a certain project:

```console
npx jest --ignoreProjects "Snapshot tests"
```

jest.config.js
colinrotherham: @36degrees I've pushed this up again

The GitHub tests are now split out so Percy only monitors what it needs to:
https://github.com/alphagov/govuk-frontend/blob/upgrade-test-jest-percy/.github/workflows/tests.yml#L47

colinrotherham:
To help review this one I've split out two of the noisier changes:

https://github.com/alphagov/govuk-frontend/pull/2857
https://github.com/alphagov/govuk-frontend/pull/2858

colinrotherham:
Rebasing with https://github.com/alphagov/govuk-frontend/pull/2857 to pick up config path changes

.github/workflows/tests.yml
```diff
@@ -27,7 +34,16 @@ jobs:
         if: steps.npm-cache.outputs.cache-hit != 'true'
         run: npm ci
 
-      - name: Run tests
-        run: npx percy exec -- npm test
+      - name: Build assets
+        run: npm run build:assets
+
+      - name: Run test projects
```
36degrees, line 40: My only slight reservation about this is that it if we add new test projects we need to remember to come and update this.

But this might just be my own mental model that CI checks ~= npm test.

.github/workflows/tests.yml
```diff
@@ -27,7 +34,16 @@ jobs:
...
+        run: |
+          npx jest --color --selectProjects "Snapshot tests"
+          npx jest --color --selectProjects "JavaScript unit tests"
+          npx jest --color --selectProjects "JavaScript behaviour tests"
+
+      - name: Run visual regression
```
36degrees, line 46: This is also running the component tests as well, right? Also not sure if '_Run_ visual regression' is the right name as the actual regression testing is a separate status check – this just sends the screenshots over to Percy.

jest.config.js
```diff
@@ -1,29 +1,74 @@
...
     {
+      ...config,
+      displayName: 'Gulp tasks',
+      testMatch: [
+        '**/gulp/**/*.test.{js,mjs}'
+      ]
+    },
+    {
+      ...config,
       displayName: 'Snapshot tests',
```
36degrees, line 16: Appreciate this name is not new, but it's not what I'd naturally call these. Especially as Jest has its own meaning for 'snapshots'.

Maybe 'Nunjucks macro tests'?

src/govuk/components/all.test.js
36degrees: Does Percy inherently need both jsdom and puppeteer because we're running it in Jest? Or is this a symptom of the fact we have these tests inside a file with a lot of other ~unrelated tests and some of them need jsdom and some need puppeteer?

package.json
```diff
@@ -10,23 +10,26 @@
   "scripts": {
     "postinstall": "npm ls --depth=0",
     "start": "gulp dev",
-    "heroku": "gulp copy-assets && gulp sassdoc && node app/start.js",
+    "heroku": "npm run build:assets && node app/start.js",
     "build-release": "./bin/build-release.sh",
     "publish-release": "./bin/publish-release.sh",
     "pre-release": "./bin/pre-release.sh",
+    "build:assets": "gulp copy-assets && gulp sassdoc",
```
36degrees, line 17: Should this run `test:build:assets` similar to what the other two build tasks do?

package.json
```diff
@@ -10,23 +10,26 @@
...
     "build:package": "gulp build:package --destination 'package' && npm run test:build:package",
     "build:dist": "gulp build:dist --destination 'dist' && npm run test:build:dist",
-    "test": "gulp copy-assets && jest --testPathIgnorePatterns=after-*",
-    "test:build:package": "jest tasks/gulp/__tests__/after-build-package.test.js",
-    "test:build:dist": "jest tasks/gulp/__tests__/after-build-dist.test.js",
+    "pretest": "npm run build:assets",
+    "test": "jest --color --ignoreProjects \"Gulp tasks\" --maxWorkers=50%",
+    "test:build:assets": "npx jest --selectProjects \"Gulp tasks\" --testMatch \"**/*components*\"",
```
36degrees, line 22: Is this intended to run `check-individual-components-compile.test.js`?

I had no idea that test file existed…

1. I think it's running on the contents of the `src` directory, so don't think it makes sense to run it after `build:assets` as that sort of suggests it's running tests against the copied files in the `public` directory.
2. I think the test basically duplicates tests we already have in `components/all.test.js`:
   https://github.com/alphagov/govuk-frontend/blob/a1a1944b174910dc76032bc9e9318a7cfc0173f6/src/govuk/components/all.test.js#L40-L44

package.json
36degrees: Oh, would that cause an infinite loop because of `pretest`? 🤔

.github/workflows/tests.yml
36degrees:
```suggestion
      - name: Send screenshots to Percy
```

jest.config.js
36degrees:
```suggestion
      displayName: 'Nunjucks macro tests',
```

package.json
36degrees: If both of those are correct, I think we should delete `check-individual-components-compile.test.js` (because it's covered by the tests in `src/govuk/components/all.test.js`) and remove the `test:build:assets` script.

.github/workflows/tests.yml
36degrees: From conversations on Slack I think the intent is to use the different jobs to parallelise the tests, which depends on this change, so happy to continue as is.

package.json
36degrees: This might be irrelevant based on https://github.com/alphagov/govuk-frontend/pull/2850#discussion_r973188316

.github/workflows/tests.yml
colinrotherham: Yeah I agree with you. I'll share a screenshot though

Runners

Having separate (named) Jest projects means we can parallelise them 😮 

Our 2.5 minute test run goes down to about ~40 seconds total

package.json
colinrotherham: Yep that was the intention, agreed. I'll remove 👍

package.json
colinrotherham: Yeah think so, thanks

This line was a precursor to https://github.com/alphagov/govuk-frontend/issues/2855 so any more improvements can come after this PR if needed

.github/workflows/tests.yml
colinrotherham: Yeah Percy starts a little service on `http://localhost:5338/percy/snapshot` then runs your command

Our tests call `await percySnapshot()` which sends the screenshot if Percy started cleanly (with a token)

src/govuk/components/all.test.js
colinrotherham: It's because we use [@percy/puppeteer](https://www.npmjs.com/package/@percy/puppeteer) which crashes without `window.fetch` and `window.WebSocket`. For performance reasons, Jest CLI no longer defaults to a simulated browser (jsdom) environment.

Here's an extract of their code ([via browserify bundle](https://github.com/percy/cli/blob/master/packages/sdk-utils/package.json#L23)) showing `.call(window)`

```mjs
(function() {
  /* Percy internals */
}).call(window);

if (typeof define === "function" && define.amd) {
  define("@percy/sdk-utils", [], () => window.PercySDKUtils);
} else if (typeof module === "object" && module.exports) {
  module.exports = window.PercySDKUtils;
}
```

I'm happy with this fix until we can investigate calling [`@percy/client`](https://github.com/percy/cli/tree/master/packages/client#percyclient) directly

This PR is just to get `jest --watch` stable again (starting/stopping test site cleanly etc)

colinrotherham: To be honest, looks like a repeat of https://github.com/alphagov/govuk-frontend/issues/2643 again

>**[Jest v28 `package.json` `exports` support](https://jestjs.io/docs/28.x/upgrading-to-jest28#packagejson-exports)**
>…As a result, you might e.g. get browser code which assumes ESM, when Jest provides ['require', 'browser']. You can either report a bug to the library…

With Jest (correctly??) now loading the _default_ browser bundle for [@percy/puppeteer](https://www.npmjs.com/package/@percy/puppeteer) on this line:
https://github.com/percy/cli/blob/master/packages/sdk-utils/package.json#L27

```json
"exports": {
  ".": {
    "node": "./dist/index.js",
    "default": "./dist/bundle.js"
  }
}
```

We'll have to stick with this fix and consider a PR with Percy to add `import` and `require` exports

.github/workflows/tests.yml
colinrotherham: Thanks @36degrees I've updated this to your suggestion 👍

jest.config.js
colinrotherham: Updated this one too 👍

package.json
colinrotherham: Removed it now, well spotted. Don't need to call `renderSass()` on every component twice 😊

package.json
colinrotherham: I'll mark this one as resolved (answered on https://github.com/alphagov/govuk-frontend/issues/2855)

docs/contributing/tasks.md
```diff
@@ -28,6 +28,10 @@ NPM scripts are defined in `package.json`. These trigger a number of gulp tasks.
 - compiles CSS & JS
 - starts up Express
 
+**`npm run build:assets` will do the following:**
+- compiles CSS & JS
+- runs `npm run test:build:assets` (which will test the output is correct)
```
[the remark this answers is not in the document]
colinrotherham: Fixed, thanks @36degrees

36degrees (approved):
Fab work 👏🏻

PR #2852, opened 2022-09-15
Fix hardcoded component name in renderAndInitialise

romaricpascal:
[When refactoring boilerplate in a helper](https://github.com/alphagov/govuk-frontend/pull/2843/commits/e57cf9915729d0065c1ec6e55ef555821caa9e93) for #2843, we forgot to make used the passed `componentName` for picking which component to render 😬


36degrees (approved):
Whoops!

Issue #2568, opened 2022-03-09
Add thousands comma-seperation in character count messages


dalepotter:
## Related component
[Character count](https://design-system.service.gov.uk/components/character-count/)

## Context

The current [GDS content style guide](https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#numbers) states:
> For numerals over 999 - insert a comma for clarity: 9,000

...however the information/error message (when over 999 characters) does not include comma separation.  

Example entering 1200 characters in the Design System [documentation page](https://design-system.service.gov.uk/components/character-count/):
![image](https://user-images.githubusercontent.com/8247168/157439409-3b52b7ba-c2e3-44d1-a1b2-f1a190eaae7d.png)

The expected message in this case would be`You have 1,000 characters too many`

## Alternatives

None found yet, but we are considering adding some custom JS to our service. This would modify the comma-seperation message output for screens which use Character Count component.

domoscargin:
Hi @dalepotter,

Thanks for this!

We recommend you use [maximum word count](https://design-system.service.gov.uk/components/character-count/#consider-if-a-word-count-is-more-helpful) for longer answers which might help here.

We think it's a sensible change to consider, since there will be cases where somebody's copying and pasting, or there's just no other way to ask the question. But generally, if there's a situation where a user could enter that many characters/words over the limit, there's probably a different way to design the question.

36degrees:
We intend to look at this when we pick up the other localisation issues for the character count component.

kellylee-gds:
@querkmachine can you write a summary of work done so far, then update the done whens on this to reflect what we'll need to do next when we pick the other internationalisation work. Thanks!

vanitabarrett:
Another alternative (if there's nothing left to do on this other than release i18n) could be to attach this to a Github milestone and move the card to Ready To Release. I think it'll sit there for a while, but that might be ok?

querkmachine:
Formatting of the character count text has been included in the [internationalisation PR](https://github.com/alphagov/govuk-frontend/pull/2614) as a progressive enhancement for browsers that support Intl.NumberFormat.

querkmachine:
Moving this back to blocked whilst we do internationalisation pre-release testing.

36degrees:
[...]

PR #2853, opened 2022-09-15
Add pluralisation support to i18n functionality

### LINKED ISSUE #2804: Expand internationalisation logic to add pluralisation
## What
Expand our internationalisation logic to handle pluralisation. This will allow users to be able to translate hardcoded strings in the character count component.

For example: “You have 1 character remaining” vs “You have 20 characters remaining”

Needs to handle languages other than English, that may have different plural forms, see https://cldr.unicode.org/index/cldr-spec/plural-rules 

## Why
Some components will need to be able to switch between different strings based on the value of something, e.g: the character count component will need to switch the count message based on the count (see example above)

## Who needs to work on this
Developers

## Who needs to review this
Developers

## Done when

- [x] Internationalisation logic expanded to introduce support for pluralisation + unit tests
- [x] As noted in [feedback](https://docs.google.com/document/d/1B9ag7WPN0dDuBxVSKntVBxUM7Y7xVTKromKPcZwIjZc/edit#heading=h.qz76th513722), ensure there is only one key for 'zero’ translations
- [x] Internationalisation logic expanded to introduce support for number formatting, where relevant + tests
- [x] Nicely handles when a required plural form for a language is not provided, e.g: a clear error message (see [related comment from pre-release testing](https://github.com/alphagov/govuk-frontend/issues/2566#issuecomment-1222340758))

querkmachine:
Expands the i18n functionality to add pluralisation support, via passing the special `count` option through the `t()` method. Doing so will locate the appropriate plural form suffix for the current locale and append this to the lookup key. 

Suffixes are based on the [Unicode CLDR specification](https://cldr.unicode.org/index/cldr-spec/plural-rules) and align with the output of the [Intl.PluralRules implementation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/PluralRules/select).

For example, calling the method with `testKey` will instead use the following lookup keys in the configured locales:

```js
t('testKey', { count: 0 })
// locale: 'en' => testKeyOther
// locale: 'cy' => testKeyZero
// locale: 'fr' => testKeyOne

t('testKey', { count: 1 })
// locale: 'en' => testKeyOne
// locale: 'cy' => testKeyOne
// locale: 'zh' => testKeyOther

t('testKey', { count: 6 })
// locale: 'en' => testKeyOther
// locale: 'cy' => testKeyMany
// locale: 'ga' => testKeyFew
```

Closes #2804.

## Changes

- Added `config` parameter to the I18n constructor. Currently this only has one option, `locale`. Unlike the `locale` option in earlier spikes, this one is [only intended to be used internally](https://github.com/alphagov/govuk-frontend/issues/2804#issuecomment-1246451644) and isn't configurable by users. Falls back to the document language if not specified. 
- Modifies the `t` method to check for the presence of a `count` option and, if present, calls `getPluralSuffix` and modifies the lookup key to point to the appropriate plural form. 
- Modifies the `replacePlaceholders` method to check if the placeholder is of type 'number' and, if so, to format it according to the configured locale. The number will not be formatted if `Intl.NumberFormat` isn't supported.
- Added three of new functions to I18n:
   - `hasIntlPluralRulesSupport`: Checks for browser support for `Intl.PluralRules` and related dependencies. Used by `getPluralSuffix`.
   - `hasIntlNumberFormatSupport`: Checks for browser support for `Init.NumberFormat` and related dependencies. Used by `replacePlaceholders`.
   - `getPluralSuffix`: Gets the CLDR specification plural form type (`zero`, `one`, `two`, `few`, `many` or `other`) based on the given number and configured locale.  Falls back to the hardcoded plural rules if `Intl.PluralRules` is not supported.
- Added two lookup maps to I18n:
  - `pluralRules`: Provides a fallback set of functions for determining the correct plural suffix to use for a given number and language.
  - `pluralRulesMap`: Object mapping language codes to the appropriate function within `pluralRules`.
- Added unit tests.
  - Checks that placeholders of type 'number' are formatted as numbers.
  - Checks that placeholders that look like numbers, but are not (i.e. strings) are not formatted as numbers.
  - Checks that an error is thrown if the currently required plural form hasn't been passed in the `translations` object.
  - Checks that string interpolation of the `count` option is possible and uses the appropriate plural form for the value of `count`.
  - Checks that the fallback `pluralRules` functions produce values consistent with `Intl.PluralRules`.

## Supported fallback languages
Browsers that support `Intl.PluralRules` theoretically support any language the browser vendor chooses to support. For older browsers (primarily Internet Explorer), we provide a set of fallback functions for a limited set of languages. 

We don't have any definitive data on what languages are used by forms services on GOV.UK (the main expected use case for these changes). This list has been compiled by combining languages we *know* are used by GOV.UK services, some assumptions, and languages that, although not in known use, have plural rules identical to other supported languages and are 'quick wins' for support.

Known in use by GOV.UK services:
- Arabic (ar)
- Bangla (bn)
- Chinese (zh)
- English (en)
- French (fr)
- Gujarati (gu)
- Hindi (hi)
- Indonesian (id)
- Japanese (jp)
- Korean (ko)
- Russian (ru)
- Spanish (es)
- Tamil (ta)
- Thai (th)
- Turkish (tr)
- Urdu (ur)
- Vietnamese (vi)
- Welsh (cy)

Languages that are assumed may be in use:
- Irish Gaelic (ga)
- Scottish Gaelic (gd)

'Quick win' languages:
- Afrikaans (af)
- Albanian (sq)
- Armenian (hy)
- Azerbaijani (az)
- Basque (eu)
- Bulgarian (bg)
- Burmese (my)
- Catalan (ca)
- Danish (da)
- Dutch (nl)
- Estonian (et)
- Farsi (fa)
- Finnish (fi)
- Georgian (ka)
- German (de)
- Greek (el)
- Hungarian (hl)
- Italian (it)
- Javanese (jv)
- Luxembourgish (lb)
- Malay (ms)
- Norwegian (de)
- Portuguese (European) (pt-PT)
- Punjabi (pa)
- Somali (so)
- Swahili (sw)
- Swedish (sv)
- Telugu (te)
- Ukrainian (uk)
- Zulu (zu)

### DIFF
--- a/src/govuk/i18n.mjs
+++ b/src/govuk/i18n.mjs
@@ -27,7 +27,7 @@ function I18n (translations, config) {
 I18n.prototype.t = function (lookupKey, options) {
   if (!lookupKey) {
     // Print a console error if no lookup key has been provided
-    throw new Error('i18n lookup key missing.')
+    throw new Error('i18n: lookup key missing')
   }
 
   // If the `count` option is set, determine which plural suffix is needed and
@@ -43,6 +43,11 @@ I18n.prototype.t = function (lookupKey, options) {
 
     // Overwrite our existing lookupKey
     lookupKey = lookupKey + pluralSuffix
+
+    // Throw an error if this new key doesn't exist
+    if (!(lookupKey in this.translations)) {
+      throw new Error('i18n: Plural form "' + pluralSuffix + '" is required for "' + this.locale + '" locale')
+    }
   }
 
   if (lookupKey in this.translations) {
@@ -250,6 +255,8 @@ I18n.prototype.pluralRules = {
     if (last === 1 && lastTwo !== 11) { return 'one' }
     if (last >= 2 && last <= 4 && !(lastTwo >= 12 && lastTwo <= 14)) { return 'few' }
     if (last === 0 || (last >= 5 && last <= 9) || (lastTwo >= 11 && lastTwo <= 14)) { return 'many' }
+    // Note: The 'other' suffix is only used by decimal numbers in Russian.
+    // We don't anticipate it being used, but it's here for consistency.
     return 'other'
   },
   scottish: function (n) {
--- a/src/govuk/i18n.mjs
+++ b/src/govuk/i18n.mjs
@@ -151,7 +151,7 @@ I18n.prototype.hasIntlNumberFormatSupport = function () {
  */
 I18n.prototype.getPluralSuffix = function (count) {
   var locale = this.locale
-  var localeShort = (locale.length > 2) ? locale.substring(0, 2) : locale
+  var localeShort = locale.split('-')[0]
   var keySuffix = 'other'
 
   // Validate that the number is actually a number.
--- a/src/govuk/i18n.mjs
+++ b/src/govuk/i18n.mjs
@@ -2,12 +2,18 @@
  * i18n support initialisation function
  *
  * @constructor
- * @param    {Object}  translations  - Key-value pairs of the translation
-                                       strings to use.
+ * @param  {Object}  translations   - Key-value pairs of the translation strings to use.
+ * @param  {Object}  config         - Configuration options for the function.
+ * @param  {String}  config.locale  - An overriding locale for the PluralRules functionality.
  */
-function I18n (translations) {
+function I18n (translations, config) {
+  config = config || {}
+
   // Make list of translations available throughout function
   this.translations = translations || {}
+
+  // The locale to use for PluralRules and NumberFormat
+  this.locale = config.locale || document.documentElement.lang || 'en'
 }
 
 /**
@@ -22,7 +28,24 @@ I18n.prototype.t = function (lookupKey, options) {
   if (!lookupKey) {
     // Print a console error if no lookup key has been provided
     throw new Error('i18n lookup key missing.')
-  } else if (lookupKey in this.translations) {
+  }
+
+  // If the `count` option is set, determine which plural suffix is needed and
+  // change the lookupKey to match. We check to see if it's undefined instead of
+  // falsy, as this could legitimately be 0.
+  if (options && typeof options.count !== 'undefined') {
+    // Get the plural suffix
+    var pluralSuffix = this.getPluralSuffix(options.count)
+
+    // We need to transform this to have an initial uppercase letter,
+    // as our keys are stored in camelCase
+    pluralSuffix = pluralSuffix.charAt(0).toUpperCase() + pluralSuffix.slice(1)
+
+    // Overwrite our existing lookupKey
+    lookupKey = lookupKey + pluralSuffix
+  }
+
+  if (lookupKey in this.translations) {
     // Fetch the translation string for that lookup key
     var translationString = this.translations[lookupKey]
 
@@ -37,7 +60,8 @@ I18n.prototype.t = function (lookupKey, options) {
       return translationString
     }
   } else {
-    // Otherwise, return the lookup key itself as the fallback
+    // If the key wasn't found in our translations object,
+    // return the lookup key itself as the fallback
     return lookupKey
   }
 }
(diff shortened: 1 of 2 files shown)


src/govuk/i18n.mjs
```diff
@@ -74,4 +110,195 @@ I18n.prototype.replacePlaceholders = function (translationString, options) {
...
+ * usually care about this part, as pluralisation rules are usually the same
+ * regardless of region. There are exceptions, however, (e.g. Portuguese) so
+ * this searches by both the full and shortened locale codes, just to be sure.
+ *
+ * @param    {number}  count       - Number used to determine which pluralisation to use.
+ * @returns  {string}              - The suffix associated with the correct pluralisation for this locale.
+ */
+I18n.prototype.getPluralSuffix = function (count) {
+  var locale = this.locale
+  var localeShort = (locale.length > 2) ? locale.substring(0, 2) : locale
```
36degrees, line 154: BCP 47 [suggests that the primary language subtags could also be 3 characters](https://datatracker.ietf.org/doc/html/rfc5646#section-2.2.1) – do we want to play it safe and split on the separator (`-`) instead?

src/govuk/i18n.unit.test.mjs
```diff
@@ -132,5 +132,156 @@ describe('I18n', () => {
...
+
+      it('returns the correct plural form for a given count (Arabic rules)', () => {
+        const locale = 'ar'
+        const localeNumbers = [105, 125]
+
+        const i18n = new I18n({}, { locale: locale })
+        const intl = new Intl.PluralRules(locale)
+
+        testNumbers.concat(localeNumbers).forEach(num => {
+          expect(i18n.pluralRules.arabic(num)).toEqual(intl.select(num))
```
36degrees, line 186: This is a really neat way to test the fallback rules 👍🏻

src/govuk/i18n.unit.test.mjs
```diff
@@ -132,5 +132,156 @@ describe('I18n', () => {
...
+
+      expect(i18n.t('ageString', { age: 2000 })).toBe('I am 2,000 years old')
+    })
+
+    it('does not format number-like strings that are passed as placeholders', () => {
+      const i18n = new I18n({
+        yearString: 'Happy new year %{year}'
+      })
+
+      expect(i18n.t('yearString', { year: '2023' })).toBe('Happy new year 2023')
```
36degrees, line 149: Do we need a test to ensure that the number formatting changes depending on `locale`?

36degrees (commented):
This is very, very cool.

Well commented, tests seem to make sense.

is there anything that we've used as a reference, copied or adapted that should be credited either in the commit messages or comments? (And do we need to check the licensing on anything?)

36degrees:
I think this also closes #2568 – worth associate this PR with it and moving it into Needs Review so nobody else tries to pick it up?

src/govuk/i18n.unit.test.mjs
querkmachine: I considered it, but this struck me as more a test of the 'is this a number' logic rather than the number formatting functionality itself—as that functionality is being handled by the browser API—it didn't feel necessary to test more than once.

src/govuk/i18n.mjs
querkmachine: Good catch. Will update it to do that.

src/govuk/i18n.unit.test.mjs
36degrees: My thinking is that it checks that we're passing `locale` to `Intl.NumberFormat` correctly – we're not somehow hardcoding it to English or passing `undefined` which would actually still pass this test (at least in Chrome):

```js
new Intl.NumberFormat(undefined).format(1000)
// -> '1,000'
```

36degrees (approved):

PR #2854, opened 2022-09-15
Allow ErrorSummary to receive JavaScript config

### LINKED ISSUE #2838: Allow error summary config set via data attributes to be passed when initialising the component in JavaScript
## What

The error summary component currently uses a `data-disable-auto-focus` data attribute to prevent the component from taking focus on page load. (This is mostly designed to be used when the component is displayed in a different context, e.g. in the Design System itself)

Update the component so that the config can also be passed as an object when initialising the component in JavaScript.

## Why

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [x] The 'disable auto focus' functionality can be provided when instantiating an ErrorSummary object
- [x] The 'disable auto focus' functionality can be provided when instantiating an ErrorSummary object using `initAll`
- [x] If a `data-disable-auto-focus` data attribute is present on the $module HTML element it overrides any config set via initialisation
- [x] Tests to cover the above

romaricpascal:
Following the tracks of #2843 for the notification banner, and #2826 for the accordion, this PR adds the possibility to configure the `ErrorSummary` component through Javascript, alongside its existing data attributes.

Each commit takes care of one of the 3 things that needed to happen for that:
1. Allowing the component to receive JavaScript configuration
2. Letting `initAll` pass through any configuration when instantiating `ErrorSummary`
3. Making sure that the data attribute takes precedence over the JavaScript configuration

Closes #2838

### DIFF
--- a/src/govuk/all.mjs
+++ b/src/govuk/all.mjs
@@ -57,11 +57,15 @@ function initAll (config) {
 
   // Find first error summary module to enhance.
   var $errorSummary = $scope.querySelector('[data-module="govuk-error-summary"]')
-  new ErrorSummary($errorSummary, config.errorSummary).init()
+  if ($errorSummary) {
+    new ErrorSummary($errorSummary, config.errorSummary).init()
+  }
 
   // Find first header module to enhance.
   var $header = $scope.querySelector('[data-module="govuk-header"]')
-  new Header($header).init()
+  if ($header) {
+    new Header($header).init()
+  }
 
   var $notificationBanners = $scope.querySelectorAll('[data-module="govuk-notification-banner"]')
   nodeListForEach($notificationBanners, function ($notificationBanner) {
--- a/src/govuk/all.mjs
+++ b/src/govuk/all.mjs
@@ -21,6 +21,7 @@ import Tabs from './components/tabs/tabs.mjs'
  * @param {HTMLElement} [config.scope=document] - scope to query for components
  * @param {Object} [config.accordion] - accordion config
  * @param {Object} [config.notificationBanner] - notification banner config
+ * @param {Object} [config.errorSummary] - error summary config
  */
 function initAll (config) {
   config = typeof config !== 'undefined' ? config : {}
@@ -56,7 +57,7 @@ function initAll (config) {
 
   // Find first error summary module to enhance.
   var $errorSummary = $scope.querySelector('[data-module="govuk-error-summary"]')
-  new ErrorSummary($errorSummary).init()
+  new ErrorSummary($errorSummary, config.errorSummary).init()
 
   // Find first header module to enhance.
   var $header = $scope.querySelector('[data-module="govuk-header"]')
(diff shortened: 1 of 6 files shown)


src/govuk/components/error-summary/error-summary.yaml
```diff
@@ -196,3 +196,8 @@ examples:
   data:
     titleText: There is a problem
     disableAutoFocus: true
+- name: autofocus explicitely enabled
```
36degrees, line 199:
```suggestion
- name: autofocus explicitly enabled
```

src/govuk/all.mjs
```diff
@@ -56,7 +56,7 @@ function initAll (config) {
 
   // Find first error summary module to enhance.
   var $errorSummary = $scope.querySelector('[data-module="govuk-error-summary"]')
-  new ErrorSummary($errorSummary).init()
+  new ErrorSummary($errorSummary, config.errorSummary).init()
```
36degrees, line 59: Can you add `config.errorSummary` to the initAll doc block?

(I'm still not 100% sure this is going to scale very well, but let's be consistent)

src/govuk/components/error-summary/template.njk
```diff
@@ -1,6 +1,6 @@
 <div class="govuk-error-summary
   {%- if params.classes %} {{ params.classes }}{% endif %}"
-  {%- if params.disableAutoFocus %} data-disable-auto-focus="true"{% endif %}
+  {%- if params.disableAutoFocus !== undefined %} data-disable-auto-focus="{{ params.disableAutoFocus }}"{% endif %}
```
36degrees, line 3: Can we add a corresponding test to `template.test.js`?

romaricpascal:
Made the code updates, but still working on the failing tests. Narrowed it down to: when the code calls `mergeConfigs` or `normaliseDataset`, the other tests/suites fail. But obviously if they're not called, the test for this specific feature fails...

Trying to see if I can get some more explanation of the failures from Jest as it's a bit... light... on why all these tests fail 😢

src/govuk/components/error-summary/error-summary.mjs
```diff
@@ -2,8 +2,18 @@ import '../../vendor/polyfills/Function/prototype/bind.mjs'
...
+function ErrorSummary ($module, config) {
   this.$module = $module
+
+  var defaultConfig = {
+    disableAutoFocus: false
+  }
+  this.config = mergeConfigs(
+    defaultConfig,
+    config || {},
+    normaliseDataset($module.dataset)
```
36degrees, line 15: I think it's probably fine, but technically if someone was previously instantiating an ErrorSummary class with an `undefined` `$module` (not using `initAll`) it'd have failed silently, whereas now it throws an error.

We could add another guard inside the constructor if we wanted to play it extra safe?

src/govuk/components/error-summary/error-summary.mjs
```diff
@@ -2,8 +2,18 @@ import '../../vendor/polyfills/Function/prototype/bind.mjs'
 import '../../vendor/polyfills/Event.mjs' // addEventListener
 import '../../vendor/polyfills/Element/prototype/closest.mjs'
 
-function ErrorSummary ($module) {
+import { mergeConfigs, normaliseDataset } from '../../common.mjs'
+function ErrorSummary ($module, config) {
```
36degrees, line 6: Whilst we're here, Can we add JSDoc blocks for at least the init function, as its signature has changed?

36degrees (approved):
Changes look good.

Ideally I'd prefer the comments wrapped at 80 chars as I find the 'uneven' line length hard to read. Appreciate this is the sort of thing that should really be enforced by a linter rather than coming up at code review…

PR #2857, opened 2022-09-16
Make `config/paths.json` exportable for ESM import

colinrotherham:
This PR renames **./config/paths.json** → **./config/paths.js** as it's easier to import in ESM

## **Before**
Parses `import.meta.url` due to no `__dirname` in ESM
```mjs
import { dirname, resolve } from 'path'
import { fileURLToPath } from 'url'
import { readFile } from 'fs/promises'

const basePath = dirname(fileURLToPath(import.meta.url))
const configPaths = JSON.parse(await readFile(resolve(basePath, '../../paths.json')))
```

## **After** (via import)
Plain CommonJS import
```mjs
import configPaths '../../paths.js'
```

## **Future** (via import assertion from JSON)
Plain JSON import but needs `--experimental-json-modules` flag in Node.js 16
```mjs
import configPaths '../../paths.json' assert { type: 'json' }
```


36degrees (approved):
👍🏻

PR #2861, opened 2022-09-20
Use async file operations for helpers

colinrotherham:
This PR aims to remove IO blocking `readFileSync()` calls (including component directory listings)

Puppeteer tests are running super quickly 😊 

https://user-images.githubusercontent.com/415517/191203826-04de51dc-f34f-431e-8042-ab45bf0effeb.mp4


tasks/gulp/__tests__/after-build-package.test.js
```diff
@@ -129,71 +132,60 @@ describe('package/', () => {
...
+        const data = await readFile(path.join(configPaths.package, 'govuk', 'components', component, 'fixtures.json'), 'utf8')
+        const parsedData = JSON.parse(data)
+
+        // We expect the component JSON to contain "component" and an array of "fixtures" with "name", "options", and "html"
+        expect(parsedData).toEqual(
+          expect.objectContaining({
+            component,
+            fixtures: expect.any(Array)
+          })
+        )
```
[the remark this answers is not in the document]
colinrotherham: I'm glad you thought this too 😊 

Yeah you do lose the separate green tick per `it.each()` _but_ because the `expect.objectContaining()` assertion has a unique `component` property (per Promise) it still prints the component name in the failure output 🙌 

Used the same approach for HTML validation too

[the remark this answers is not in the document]
romaricpascal: Neat! That's my worries sorted then 😄

36degrees (commented):
Thanks for this – excited to see what performance improvements it yields 🚀 

This would be easier to review if it was broken down in to smaller commits – appreciate there are probably changes that need to happen together when the helper methods in `lib` change which might make this complicated, but broadly I think this could be broken down into:

- changes to the way we get components / examples
- changes to tests (moving things into beforeAl, refactoring to use promises)
- moving eslint environment setting from comments to `.eslintrc.js` files
- updating GitHub action workflows to use fs/promises

I've also called out a change in behaviour in the way the app works that I think needs consideration.

app/app.js
```diff
@@ -37,6 +33,23 @@ module.exports = (options) => {
     ...nunjucksOptions // merge any additional options and overwrite defaults above.
   })
 
+  // Cache common component listings
+  const [examples, components, fullPageExamples] = await Promise.all([
+    fileHelper.getDirectories(configPaths.examples),
+    fileHelper.getAllComponents(),
+    fileHelper.getFullPageExamples()
+  ])
```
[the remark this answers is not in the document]
colinrotherham: This is good feedback!

Problem came from seeing the app crawl all the component directories, for every page view 😮

What about updating the `gulp serve` (Nodemon) watch list instead?

Then we can still keep things cached until a file is modified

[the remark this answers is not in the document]
36degrees: As long as it reloads reasonably quickly (so there's minimal chance of reloading and getting a connection error) I'd be happy with that as a solution ⚡

If it takes a while to reload it might be worth living with the 'crawl on every page load' and writing up an issue to revisit in the future? I haven't personally noticed any issues with page load times in the review app – generally feels pretty responsive, even if it is doing lots of file system reads under the hood.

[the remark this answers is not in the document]
colinrotherham: Yeah they're just tiny lags

It's the multiple seconds that add up whilst the various tests send requests (100s?) to the review app

I'll try it out and see

app/app.test.js
```diff
@@ -51,10 +51,13 @@ describe(`http://localhost:${PORT}`, () => {
...
       const response = await fetchPath('/')
       const $ = cheerio.load(await response.text())
+
+      const componentsDirectory = await getDirectories(configPaths.components)
       const componentsList = $('li a[href^="/components/"]').get()
+
       // Since we have an 'all' component link that renders the default example of all
       // components, there will always be one more expected link.
-      const expectedComponentLinks = lib.allComponents.length + 1
+      const expectedComponentLinks = componentsDirectory.size + 1
```
colinrotherham, line 60: The file system helpers `getDirectories()`, `getFiles()` or the "everything" `getListing()` are now [Maps](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map)

```mjs
Map {
  'component-1' => {
    basename: 'component-1',
    path: 'src/govuk/components/component-1',
    stats: fs.Stats
  },
  'component-2' => {
    basename: 'component-2',
    path: 'src/govuk/components/component-2',
    stats: fs.Stats
  }
}
```

Which comes in handy for simplifying lots of other tests:

**`componentsDirectory.keys()`** → `['component-1', 'component-2']`
**`componentsDirectory.has('component-2')`** → `true`
**`componentsDirectory.has('component-3')`** → `false`

app/views/all-components.njk
```diff
@@ -19,9 +19,7 @@
       All components
     </h1>
   </div>
-  {% for data in componentData %}
-    {% set componentName = data.componentName %}
-    {% set componentNameHuman = componentName | replace("-", " ") %}
-    {{ showExamples(componentName, componentNameHuman, data, legacyQuery) }}
+  {% for component in components %}
+    {{ showExamples(component, legacyQuery) }}
```
colinrotherham, line 23: To remove a loop somewhere I added `.name` to each component. Was helpful here too

lib/file-helper.js
```diff
@@ -9,75 +9,148 @@ const configPaths = require('../config/paths.js')
...
   getFiles,
+  getFilesByDirectory,
+  getFullPageExamples,
   getListing
 }
+
+/**
+ * Component data from YAML
+ *
+ * @typedef {object} ComponentData
```
colinrotherham, line 150: For editors and IDEs that support JSDoc this is the bit where object types can be discovered

```mjs
const component = await getComponent(componentName)

component.name
component.params
component.previewLayout
component.accessibilityCriteria
```

JSDoc @typedef showing all available component properties

src/govuk/components/components.test.js
```diff
@@ -16,8 +22,8 @@ describe('Components', () => {
     })
 
     it('renders CSS for each component', () => {
-      const sassTasks = allComponents.map((component) => {
-        const file = `${configPaths.src}/components/${component}/_${component}.scss`
+      const sassTasks = [...componentsFiles].map(([componentName, files]) => {
```
colinrotherham, line 25: This is "Map-to-Array" shorthand to replace `Array.from(componentsFiles)`

src/govuk/components/components.test.js
```diff
@@ -16,8 +22,8 @@ describe('Components', () => {
...
+        const file = files.get(`_${componentName}.scss`)?.path
```
colinrotherham, line 26: Notice the `?.path` [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)

What if the `${componentName}.scss` file doesn't exist?

Here we continue with `undefined` so `renderSass()` has the opportunity to fail

tasks/gulp/__tests__/after-build-package.test.js
```diff
@@ -1,21 +1,30 @@
-const fs = require('fs')
+const { readFile } = require('fs/promises')
 const path = require('path')
-const util = require('util')
-
 const recursive = require('recursive-readdir')
```
colinrotherham, line 3: Didn't do this, but could probably replace this package with `getFilesByDirectory()`

tasks/gulp/__tests__/after-build-package.test.js
```diff
@@ -1,21 +1,30 @@
...
+  let componentsFilesSource
+  let componentsFilesPackage
+  let componentsFilesPackageESM
+
+  let componentNames
+
+  beforeAll(async () => {
+    componentsFilesSource = await getFilesByDirectory(configPaths.components)
+    componentsFilesPackage = await getFilesByDirectory(`${configPaths.package}govuk/components/`)
+    componentsFilesPackageESM = await getFilesByDirectory(`${configPaths.package}govuk-esm/components/`)
```
colinrotherham, line 22: Sorry, I couldn't help but add ESM file checks at the same time

romaricpascal:
Thanks for splitting the commits and going through all these changes. The move to async looks good to me, I like how it feels more parallel-y with the `Promise.all` 🙌🏻 . 

Only concern I have in that PR is the naming move from `componentData` to `component` that:

- looses the separation between the data about the component and the component itself
- makes it weird with the types of a lot of `component` variables being  `ComponentData` 

It feels a bit nitpicky and I'm not holding strongly to going back to `componentData`, but thought I'd mention it  😄

colinrotherham:
Ha, that's fine

Consistency wise, I'm more than happy to make everything `componentData`

Consistency was the aim, not the name 🙌

Or rename the `@typedef` to **Component**?

romaricpascal:
Making everything `componentData` makes more sense to me, as we keep the distinction from the component itself (in case we ever have tests that need to reference both).

colinrotherham:
> Making everything `componentData` makes more sense to me, as we keep the distinction from the component itself (in case we ever have tests that need to reference both).

All sorted, now using `componentData` instead 🎉

romaricpascal (approved):
Ace! Looks all good to me, thanks for that extra pass on it 😄

PR #2867, opened 2022-09-20
Allow Button to receive configuration via JavaScript

### LINKED ISSUE #2836: Allow button config set via data attributes to be passed when initialising the component in JavaScript
## What

The button component currently uses a `data-prevent-double-click` data attribute to enable the 'prevent double click' functionality.

Update the button component so that the config can also be passed as an object when initialising the component in JavaScript.

## Why

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [ ] The 'prevent double click' functionality can be provided when instantiating a Button object
- [ ] The 'prevent double click' functionality can be provided when instantiating a Button object via `initAll`
- [ ] If a `data-prevent-double-click` data attribute is present on the $module HTML element it overrides any config set via initialisation
- [ ] Tests to cover the above

romaricpascal:
The PRs for adding configuration via JavaScript carries on with the Button. 🎉 

Changes are similar to the [ErrorSummary](https://github.com/alphagov/govuk-frontend/pull/2854) and [NotificationBanner](https://github.com/alphagov/govuk-frontend/pull/2854) and the commits paint a step-by-step picture of the changes. 

The most notable thing is that it slightly changes when the value for whether to prevent double click is read. On `main` the value is read within the `click` listener but with this update, it'll be read at component instantiation. This has an effect on the following scenario:

1. the attribute could be set when the component is instantiated
2. some other piece of JS on the page may remove it while the user browses
3. at the time the use click, the attribute would then not be there, calling for a different behaviour than when the component was initialised

On `main` the `Button` will behave according to the value of the attribute at step 3, while after this PR it'll behave according to the value at step 1. 

Consistency with the other components has been prioritised here after discussing that updating the data-attribute at runtime was not an expected nor supported use case.

Closes #2836.

### DIFF
--- a/src/govuk/components/button/button.mjs
+++ b/src/govuk/components/button/button.mjs
@@ -4,7 +4,19 @@ import '../../vendor/polyfills/Function/prototype/bind.mjs'
 var KEY_SPACE = 32
 var DEBOUNCE_TIMEOUT_IN_SECONDS = 1
 
-function Button ($module) {
+/**
+ * JavaScript enhancements for the Button component
+ *
+ * @class
+ * @param {HTMLElement} $module - The element this component controls
+ * @param {Object} config
+ * @param {Boolean} [config.preventDoubleClick=false] - Whether the button should prevent double clicks
+ */
+function Button ($module, config) {
+  if (!$module) {
+    return this
+  }
+
   this.$module = $module
   this.debounceFormSubmitTimer = null
 }
@@ -13,6 +25,10 @@ function Button ($module) {
  * Initialise component
  */
 Button.prototype.init = function () {
+  if (!this.$module) {
+    return
+  }
+
   this.$module.addEventListener('keydown', this.handleKeyDown)
   this.$module.addEventListener('click', this.debounce)
 }
--- a/src/govuk/components/button/button.mjs
+++ b/src/govuk/components/button/button.mjs
@@ -1,3 +1,4 @@
+import { mergeConfigs, normaliseDataset } from '../../common.mjs'
 import '../../vendor/polyfills/Event.mjs' // addEventListener and event.target normaliziation
 import '../../vendor/polyfills/Function/prototype/bind.mjs'
 
@@ -19,6 +20,15 @@ function Button ($module, config) {
 
   this.$module = $module
   this.debounceFormSubmitTimer = null
+
+  var defaultConfig = {
+    preventDoubleClick: false
+  }
+  this.config = mergeConfigs(
+    defaultConfig,
+    config || {},
+    normaliseDataset($module.dataset)
+  )
 }
 
 /**
@@ -30,7 +40,7 @@ Button.prototype.init = function () {
   }
 
   this.$module.addEventListener('keydown', this.handleKeyDown)
-  this.$module.addEventListener('click', this.debounce)
+  this.$module.addEventListener('click', this.debounce.bind(this))
 }
 
 /**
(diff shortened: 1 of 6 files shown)


src/govuk/components/button/button.mjs
```diff
@@ -1,20 +1,46 @@
...
+    config || {},
+    normaliseDataset($module.dataset)
+  )
 }
 
 /**
  * Initialise component
  */
 Button.prototype.init = function () {
+  if (!this.$module) {
```
querkmachine, line 38: Asking for clarification: 

At the start of the module we're checking for the existence of `$module`, and only if it exists does the code to assign `$module` to `this.$module` run.

Here we are then checking for the existence of `this.$module`, however wouldn't the previous steps ensure that this statement would never evaluate to false?

romaricpascal: Unfortunately we need that extra check  in `init` because the first check only shortcuts the constructor, to safeguard the access ot the module's `dataset`. The `addEventListener` calls happen inside `init()` fail if the component was instantiated with an `undefined`/`null` element.

querkmachine: If the constructor was called with an `undefined` element, wouldn't we want to immediately halt execution anyway?

romaricpascal: Sorry Ollie, I was commenting on an old version of the page and hadn't seen you'd replied.

> I don't think there's anything we can do from the constructor to stop the `init` function from being called – that's happening outside of the object (`new govukButton($module).init()`).

We could always make `init()` a no-op in the constructor. That would keep everything in the same place. That'd actually make the thing quite portable from component to component. 

```js
if (!$module) {
   // Make `init` a no-op to avoid potentially breaking code at initialisation
   // from lack of HTML element 
   this.init = function(){}
   return this
}
```

romaricpascal: I'll actually update the PR for that as:
- it'll avoir the awkward second check that prompted @querkmachine's question
- it'll be easy to pass from one component to another

I'll open a second PR to amend the components we've already made configurable.

36degrees: I'm not 100% sure about this from a code readability perspective – appreciate at the minute the init function is near the constructor but I think it'd be easy to miss and I don't think I'd expect a class method to be redefined in this way…

If I was writing this I think I'd stick with the two guard clauses we had before, especially if we plan to get rid of the `init` function in the short-to-mid term.

romaricpascal: Happy to reset to the previous commit, it's not a solution I was particularly attached to and it'll become irrelevant when we remove the `init` function. Just thought it'd make sense to keep everything in a single place instead of the double check, especially as it's for handling an "error" scenario.

src/govuk/components/button/button.test.js
```diff
@@ -27,112 +47,240 @@ describe('/components/button', () => {
...
-          $button.parentNode.removeChild($button)
-          $form.appendChild($button)
-
-          window.__SUBMIT_EVENTS = 0
-          $form.addEventListener('submit', event => {
-            window.__SUBMIT_EVENTS++
-            // Don't refresh the page, which will destroy the context to test against.
-            event.preventDefault()
-          })
+          const $buttonPrime = $button.cloneNode(true)
```
36degrees, line 148: It took me a while to work out what this meant (and I'm still not 100% sure – guessing this is Is prime as in 'a derivative'?)

Wondering if `$secondButton` might be simpler and clearer?

36degrees (commented):
This is looking good – just a few minor formatting things that need tidying up

36degrees (approved):
👍🏻

PR #2869, opened 2022-09-21
Fix webpack docs example

colinrotherham:
Splitting out the "Fix webpack docs" commit from https://github.com/alphagov/govuk-frontend/pull/2863 into this one

I've made these changes by hand so needs a bit of scrutiny


romaricpascal:
Unrelated to the changes, but from giving it a quick go on my machine, I've had to make a few updates for `npm install` to run OK in the example folder:

- update webpack's version to "^4.6.0" to not have conflicting versions
- change `node-sass` to `"sass": "^1.55.0"` to have `sass` install OK with Node 16 (Node 14 was OK)

After that, however:

- `npm run dev` doesn't boot a server at all and just exits
- `npm run build` does build a `main.js` file in the `public` folder. Serving the folder with `npx serve` doesn't manage to load the styles (as it pulls them from `../../public/app.css`

Looks like there's a few more things to do to get the Webpack example back on its feet (and maybe up to date with the latest Webpack versions?)

colinrotherham:
@romaricpascal Thanks for digging into this. I'll make it a draft as I only really fixed paths and named imports

@36degrees Perhaps we could add tests for various bundlers?

36degrees:
I've added this to the agenda to discuss at tomorrow's dev catch up.

colinrotherham:
@36degrees @romaricpascal I've also fixed webpack (including the CSS) ready for catch up

docs/examples/webpack/webpack.config.js
```diff
@@ -1,51 +1,109 @@
...
+        },
+        test: /\.scss$/,
+        type: 'asset',
+        use: [
+          'postcss-loader',
+          {
+            loader: 'sass-loader',
+            options: {
+              sassOptions: {
+                quietDeps: true
```
colinrotherham, line 51: https://frontend.design-system.service.gov.uk/importing-css-assets-and-javascript/#silence-deprecation-warnings-from-dependencies-in-dart-sass

package.json
```diff
@@ -7,6 +7,10 @@
     "npm": "^8.5.0"
   },
   "license": "MIT",
+  "workspaces": [
+    "docs/examples/*",
```
colinrotherham, line 11: This line makes sure `npm install` and `npm ci` also install webpack's dependencies

romaricpascal: **quibble** Is that something we absolutely need for the example running? I'd rather we only brought example dependencies when trying to run the example rather than at every install.

romaricpascal (approved):
The demo is working 🎉  Thanks for putting it back on its feet 🙌🏻  

Only little quibble is with the installation for the examples dependency with the project's `npm install` (unless I misunderstood what it did), which I'd prefer to bring only when running the examples. It's not blocking for me if there's no other easy way 😄

package.json
colinrotherham: Thanks @romaricpascal

This was to stop the example(s) going stale or getting out of sync again.

Enables things like:

```console
npm run dev --workspace ./docs/examples/webpack
```

It's possible of course, but then we'd need an extra lockfile (per example) and have to rely on `file://` package paths to the parent `govuk-frontend`. We'd also opt out of:

1. Package updates + auditing
2. Dependabot dependency updates
3. Automatic package deduplication
4. Postinstall `"npm ls --depth=0"` check
5. Example gets a symlink to `govuk-frontend`

I'd recommend the same approach for:

```console
./app/package.json
./tasks/package.json
```

☝️ Not yet of course, but each "area" should manage its own packages and tests

romaricpascal: Makes sense. Thanks for clarifying 😄 I like the idea of properly separating the app and tasks as well, that sounds pretty neat 😄

colinrotherham:
Thanks @romaricpascal

I'll link up to the reply (to your concern) as it's useful https://github.com/alphagov/govuk-frontend/pull/2869#discussion_r998455361

package.json
colinrotherham: Yeah 100%

Deployment to Heroku doesn't need all our developer/test tooling etc 😊

Issue #2873, opened 2022-09-21
Documentation for `govuk-tint` perhaps wrong


klssmith:
## Description of the issue

I am not sure that the docs for the `govuk-tint` function are correct. The percentage is described as the [percentage of `$colour` in returned colour](https://github.com/alphagov/govuk-frontend/blob/651ad62d3a90c73b7a2e08b1729f831177d85e7d/src/govuk/helpers/_colour.scss#L89). When I've tried this out, a higher percentage adds more white to the colour, which seems to be the opposite of what the documentation suggests. `govuk-tint` is using `mix` under the hood, and the docs for that do suggest that the percentage is the percentage of the first colour, which is white in this case.

## Actual vs expected behaviour

A style of `govuk-tint(govuk-colour("blue"), 25)` looks like this:

<img width="80" alt="Screenshot 2022-09-21 at 16 53 51" src="https://user-images.githubusercontent.com/12881990/191552317-5248b51c-d2c1-4f11-b793-8ead4e39cd49.png">

A style of `govuk-tint(govuk-colour("blue"), 75)` looks like this:

<img width="80" alt="Screenshot 2022-09-21 at 16 53 26" src="https://user-images.githubusercontent.com/12881990/191552210-302f2e48-be6a-48d6-9c00-f540277d7ed7.png">

## Environment (where applicable)

- Operating system:
- Browser:
- Browser version:
- GOV.UK Frontend Version:

PR #2874, opened 2022-09-21
Reword $percentage sassdoc for govuk-tint and govuk-shade

owenatgov:
Fixes https://github.com/alphagov/govuk-frontend/issues/2873

As noted in the linked issue, we're providing incorrect information in the sassdoc for both `govuk-tint` and `govuk-shade` in that `$percentage` isn't actually a percentage of `$colour` but a percentage of the white and black being passed to each respective mixin. This change rewords the sassdoc for each mixin to be more accurate.

### DIFF
--- a/src/govuk/helpers/_colour.scss
+++ b/src/govuk/helpers/_colour.scss
@@ -75,7 +75,7 @@
 /// Make a colour darker by mixing it with black
 ///
 /// @param {Colour} $colour - colour to shade
-/// @param {Number} $percentage - percentage of `$colour` in returned colour
+/// @param {Number} $percentage - percentage of black to mix with $colour
 /// @return {Colour}
 /// @access public
 
@@ -86,7 +86,7 @@
 /// Make a colour lighter by mixing it with white
 ///
 /// @param {Colour} $colour - colour to tint
-/// @param {Number} $percentage - percentage of `$colour` in returned colour
+/// @param {Number} $percentage - percentage of white to mix with $colour
 /// @return {Colour}
 /// @access public
 


querkmachine:
Hmm, I'm not sure of the wording on this, myself. "alpha percentage applied to black/white on top of `$colour`" seems like a really obtuse way of explaining what this is doing, if more technically accurate to how the underlying function appears to work.

Would something like "percentage of black/white to mix with `$colour`" be clearer?

src/govuk/helpers/_colour.scss
```diff
@@ -75,7 +75,8 @@
 /// Make a colour darker by mixing it with black
 ///
 /// @param {Colour} $colour - colour to shade
-/// @param {Number} $percentage - percentage of `$colour` in returned colour
+/// @param {Number} $percentage - alpha percentage applied to black
+/// on top of `$colour`
```
36degrees, line 79: I'm not too hot on my colour theory but I'm not sure that it makes sense to talk about 'alpha' here as that tends to relate to transparency? Compared to the original version I'm not sure I really understand this what this means.

Does this need to be more complicated than just changing `$colour` to `black` or `white`?

```suggestion
/// @param {Number} $percentage - percentage of black in returned `$colour`
```

owenatgov: In my head the white/black was overlaying `$colour` which is why my head went to alpha, but that's not actually what's happening. The clues in the name that it's using `mix`.

Re: wording, I wonder if it's worth adding a smidge more detail. The way I read the original line, the "returned colour" is the colour that is output by these mixins so maybe specifying that the `$percentage` controls how much black/white is mixed with `$colour` to produced the returned colour.

I'm personally a fan of @querkmachine's suggestion above:

>percentage of black/white to mix with $colour

...or possibly an expanded version:

>percentage of black/white to mix with $colour in returned colour

36degrees: Might be worth talking to our designers to see how they think about tints and shades and if there's a common vocabulary we should be using here.

owenatgov: I bothered @christopherthomasdesign about this and he's concurred that my first suggestion, aka: a minor re-wording of @querkmachine's suggestion, is the winner. I'll therefore update this PR in a tic to reflect this.

36degrees (approved):

PR #2875, opened 2022-09-22
Ensure gulp tests return Promises

lfdebrux:
Fixes a couple of instances where Jest test functions were creating `Promise`s without returning them, so the test runner would get confused about which test was failing.


lfdebrux:
I looked to see if there were any other cases where we were creating a Promise and not returning it with `git grep '^\s*Promise'`, got not results.

colinrotherham (approved):

36degrees (approved):

PR #2876, opened 2022-09-22
Don’t copy *.js files into `govuk-esm`

colinrotherham:
In response to [comment regarding ***.js** files in `govuk-esm`](https://github.com/alphagov/govuk-frontend/pull/2851#issuecomment-1254698677)

>So I think we can just remove the `${configPaths.src}**/*.js` glob from the `js:copy-esm` task?

This PR includes:

1. Combine `js:copy-esm` task → `copy-files`
2. Remove `${configPaths.src}**/*.js` glob so only ***.mjs** are copied

### DIFF
--- a/tasks/gulp/copy-to-destination.js
+++ b/tasks/gulp/copy-to-destination.js
@@ -18,7 +18,6 @@ gulp.task('copy-files', () => {
     // Copy source JavaScript
     gulp.src([
       `${configPaths.src}**/*.mjs`,
-      `${configPaths.src}**/*.js`,
       `!${configPaths.src}**/*.test.*`
     ]).pipe(gulp.dest(`${taskArguments.destination}/govuk-esm/`)),
 
--- a/tasks/gulp/copy-to-destination.js
+++ b/tasks/gulp/copy-to-destination.js
@@ -15,54 +15,62 @@ const taskArguments = require('../task-arguments')
 
 gulp.task('copy-files', () => {
   return merge(
+    // Copy source JavaScript
     gulp.src([
-      `${configPaths.src}**/*`,
+      `${configPaths.src}**/*.mjs`,
+      `${configPaths.src}**/*.js`,
+      `!${configPaths.src}**/*.test.*`
+    ]).pipe(gulp.dest(`${taskArguments.destination}/govuk-esm/`)),
 
-      // Exclude files from copy
-      '!**/.DS_Store',
-      '!**/*.mjs',
-      '!**/*.test.*',
-      '!**/__snapshots__/',
-      '!**/__snapshots__/**',
+    merge(
+      gulp.src([
+        `${configPaths.src}**/*`,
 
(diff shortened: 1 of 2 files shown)


tasks/gulp/copy-to-destination.js
```diff
@@ -15,54 +15,61 @@ const taskArguments = require('../task-arguments')
 
 gulp.task('copy-files', () => {
   return merge(
+    // Copy source JavaScript
```
romaricpascal, line 18:
```suggestion
    // Copy ESM files to make them available in the package
```

tasks/gulp/copy-to-destination.js
```diff
@@ -15,54 +15,61 @@ const taskArguments = require('../task-arguments')
...
+        '!**/*.mjs',
+        '!**/*.test.*',
+        '!**/__snapshots__/',
+        '!**/__snapshots__/**',
+
+        // Preserve destination README when copying to ./package
+        // https://github.com/alphagov/govuk-frontend/tree/main/package#readme
+        `!${configPaths.src}README.md`,
+
+        // Exclude files from other streams
```
romaricpascal, line 39:
```suggestion
        // Exclude files from other streams (MJS and test files are already excluded a bit above)
```

romaricpascal (commented):
It looks good to me 🙌🏻  Just got two little comments to clarify what tripped me up while wrapping my head around things. Feel free to reword or disregard if they're not relevant 😄

tasks/gulp/copy-to-destination.js
colinrotherham: Can I suggest a tweak to align with the Gulpfile task?
https://github.com/alphagov/govuk-frontend/blob/542e7ec7bd2d047b19313d35fd9ba17e32cef8ac/gulpfile.js#L28

At the moment we use `--destination` to choose **./package** so the destination could vary

```suggestion
    // Copy ESM JavaScript to destination, ready for publishing
```

tasks/gulp/copy-to-destination.js
colinrotherham: @romaricpascal This one definitely needed a better comment

It was actually excluding files handled by the PostCSS and JSON (from YAML) streams a bit _below_

Hopefully this is clearer:

```mjs
// Exclude Sass files handled by PostCSS stream below
`!${configPaths.src}**/*.scss`,

// Exclude source YAML handled by JSON streams below
`!${configPaths.components}**/*.yaml`
```

colinrotherham:
> It looks good to me 🙌🏻 Just got two little comments to clarify what tripped me up while wrapping my head around things. Feel free to reword or disregard if they're not relevant 😄

Thanks, see what you think now

romaricpascal (approved):

colinrotherham:
This PR was related to https://github.com/alphagov/govuk-frontend/issues/2243

PR #2877, opened 2022-09-22
Ensure "clean" target matches `--destination` flag

colinrotherham:
I noticed when we build to a custom directory via `--destination` it's not "cleaned"

```
npx gulp build:package --destination 'custom/location/package'
npx gulp build:dist --destination 'custom/location/dist'
```

☝️ As those two commands only clean **./package** via `cleanPackage()` and **./dist** via `cleanDist()`

This PR includes:

1. Automatic destination based on task name `build:package` or `build:dist`
2. Automatic clean target based on destination
3. Tests for `--destination` if we want to keep it
4. ~Removal of `--destination` if we want to get rid of it~

This comment was really useful: https://github.com/alphagov/govuk-frontend/issues/2243#issuecomment-1214774873

### DIFF
--- a/tasks/task-arguments.js
+++ b/tasks/task-arguments.js
@@ -1,4 +1,38 @@
-const argv = require('yargs').argv
-const destination = argv.destination ? argv.destination : 'public'
+const { dirname, relative, resolve } = require('path')
 
-exports.destination = destination
+const slash = require('slash')
+const { argv } = require('yargs')
+
+// Defaults for known tasks
+const destinations = [
+  {
+    task: 'build:package',
+    destination: 'package'
+  },
+  {
+    task: 'build:dist',
+    destination: 'dist'
+  }
+]
+
+// Non-flag arguments
+const { _: tasks } = argv
+
+// Prefer `--destination`, default for known task, or 'public'
+const destination = argv.destination || (destinations
+  .filter(({ task }) => tasks.includes(task))[0]?.destination ?? 'public')
+
+const rootPath = dirname(__dirname)
+const destPath = resolve(rootPath, destination)
+
+module.exports = {
+  argv,
+
+  // Normalise slashes (Windows) for gulp.src
+  destination: slash(relative(rootPath, destPath)),
+
+  // Check destination flags
+  isPackage: destination === 'package',
+  isPublic: destination === 'public',
+  isDist: destination === 'dist'
+}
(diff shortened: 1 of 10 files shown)


tasks/task-arguments.unit.test.js
```diff
@@ -0,0 +1,57 @@
+describe('Task arguments', () => {
+  let args
+
+  beforeEach(() => {
+    jest.resetModules()
+
+    args = {
```
owenatgov, line 7: I feel like there's a jest reason but how come this can't be included in the declaration on line 2?

owenatgov (commented):
One tiny question. This looks great to me!

It's probs worth getting a js squad dev to look over this still as I'm rusty on these build scripts and conventions that we're trying to follow with these tooling updates.

tasks/task-arguments.unit.test.js
colinrotherham: It's a good point, and would probably work fine, but because objects are ["pass by reference"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Working_with_Objects#comparing_objects) in JavaScript there's a risk that one test will modify `args` affecting other tests

So I'm "resetting" `args` as a test fixture on every **beforeEach()**

owenatgov: Makes sense to me! Resolving.

tasks/task-arguments.js
```diff
@@ -1,4 +1,29 @@
-const argv = require('yargs').argv
-const destination = argv.destination ? argv.destination : 'public'
+const { argv } = require('yargs')
 
-exports.destination = destination
+// Defaults for known tasks
+const destinations = [
+  {
+    task: 'build:package',
```
domoscargin, line 6: Part of me wants to do this by parsing the task name, but this is way less fussy and prone to errors!

domoscargin (commented):
I think this looks grand!

Maybe at this stage it makes sense to keep the `--destination` flag until we've completed work to scope our JavaScript/Browser support changes, in case we make some big changes that require liberal use of it.

I think that's unlikely to happen, but it probably makes sense to keep our options open, and we can always come back and get rid of it down the line.

colinrotherham:
Thanks @domoscargin

Did you see the "remove" commit was 1 of 3?

I'm more than happy to drop that one, keeping the automatic `clean()` target + new test coverage?

domoscargin (approved):
I think this looks grand!

PR #2878, opened 2022-09-22
Reduce GitHub workflow runs (including Percy screenshot uploads)

colinrotherham:
Just adding some GitHub Action workflow tweaks here

### Ideas so far

**Cancel in-progress GitHub workflows when pushed again**
Forgot a change? Pushing up a second time with cancel the 1st test run

**Remove `push` event from GitHub Actions branch workflows**
Prevents duplicate runs as `pull_request` event already fires for PR pushes

~**Skip tests for documentation changes**~
~Don't need to run JavaScript tests when only `docs/**/*.md` files have changed (??)~
>☝️ Moved to branch https://github.com/alphagov/govuk-frontend/compare/main...sass-test-matrix

**Move “All components” template tests out of Puppeteer environment**
Don't lose time spinning up Chromium (headless) to run HTML validation checks etc

**Only take “JavaScript disabled” screenshots for components with JS**
Skip "JavaScript disabled" tests for components lacking `*.mjs` files as screenshots will be identical

## Before
1. Multiple workflow `[pull_request, push]` events for PRs
2. Screenshots all components with JavaScript enabled
3. Screenshots all components with JavaScript disabled

**GitHub Workflow**: 14 checks total (7 checks per run)
**Screenshots** (JavaScript enabled): 64 images (32 per run)
**Screenshots** (JavaScript disabled): 64 images (32 per run)

```console
Time:        89.322 s, estimated 96 s
```

## After
1. Single workflow `pull_request` events for PRs
2. Screenshots all components with JavaScript enabled
3. Screenshots "JS capable" components with JavaScript disabled

**GitHub Workflow**: 7 checks total
**Percy screenshots** (JavaScript enabled): 32 images
**Percy screenshots** (JavaScript disabled): 11 images

```console
Time:        58.233 s, estimated 96 s
```


colinrotherham:
@36degrees Morning! This [commit for **“JavaScript disabled” screenshots**](https://github.com/alphagov/govuk-frontend/pull/2878/commits/fba630d1e853382e7279be2ba9fe4008153dd837) follows your idea in https://github.com/alphagov/govuk-frontend/pull/2871

I've added a `getFiles()` directory listing helper to filter only components with `${component}.mjs`, but we can run it before all the tests so once Puppeteer opens the Chromium window it stays rapid

romaricpascal (commented):
That all looks pretty neat, nice!  🙌🏻 Just had 2 little questions on the Sass workflow, and maybe something that calls for a longer discussion:

From what I understood, we're splitting tests that require Puppeteer from those which don't so we only run puppeteer when necessary. However, we're keeping the simple `.test.js` naming convention for those Puppeteer test and I'm wondering if we wouldn't gain some clarity by:
- switching to a clear `.puppeteer.test.js` (or .behaviour.test.js, or whichever more generic naming we'd prefer) to clearly highlight that those files spin up Puppeteer and will be more costly to run
- match `.test.js` instead of `template.test.js` for non-puppeteer test to:
	1. reduce risks of a random new test spinning up Puppeteer
	2. not have `all.template.js` be marked as a template test while it also runs Sass compilation

.github/workflows/sass.yaml
```diff
@@ -1,94 +1,115 @@
 name: Sass
 
-on: [push, pull_request]
+on:
+  pull_request:
```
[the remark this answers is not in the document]
colinrotherham: @romaricpascal Yeah I did but I had to remove it 😆 

They're marked as mandatory checks via a **Branch protection rule** for the merge to be allowed. Can we discuss as a team, one for another PR rather than this one?

### Require status checks to pass before merging
Status checks that are required

.github/workflows/sass.yaml
```diff
@@ -1,94 +1,115 @@
...
+
+  push:
+    branches:
+      - main
+      - 'feature/**'
+      - 'support/**'
```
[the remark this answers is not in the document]
colinrotherham: @romaricpascal Sorry, I'll explain this

In a discussion with @36degrees and @injms we wanted to remove the `push` GitHub Action workflow event (so Percy screenshots weren't uploaded twice, also by `pull_request`) but _keep_ it when merging to **origin/main**

Sometimes two (passing) PRs get merged together into a single (failing) branch, so it's important to run the tests again on `push` to these recently "combined" branches.

For completeness I've added all the branches we have protection rules set up for:

Protection rules for main, feature and support

colinrotherham:
Thanks @romaricpascal, cheers for the feedback

Can we narrow down our Jest `projects: []` list (by directory/filename) in another PR?

I've made sure groups don't overlap and setup/teardown scripts only run where necessary in this change:
https://github.com/alphagov/govuk-frontend/commit/8bc8973bb136451bbf6369fed425951d8802b03b#diff-1e058ca1442e46581b13571fb8d261f9e1f5657e26c96634d4c1072f0f0347f1

.github/workflows/sass.yaml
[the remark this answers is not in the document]
romaricpascal: Oh, I see. It's probably worth a discussion and some background on why only the Sass tests are required (and not the Nunjucks or JavaScript ones 🤔 )

.github/workflows/sass.yaml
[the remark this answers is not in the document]
romaricpascal: Cheers for the explanation 😄

romaricpascal:
> Can we narrow down our Jest `projects: []` list (by directory/filename) in another PR?

Yeah, that'd be fine by me so we get this one moving and tidy up naming conventions separately 😄

colinrotherham:
Also a note about the **Add Sass test run matrix config** commit https://github.com/alphagov/govuk-frontend/pull/2878/commits/843ad1683ec0bcea7cded5b703976e30e82855dc

This removes "boilerplate" GitHub Action workflow code so new tests are only extra array items:

```yaml
- name: Dart Sass v1.0.0
  node-version: 8 # v8 required for sass v1.0.0

  run: |
    npm install -g sass@1.0.0
    sass --version
    time sass src/govuk/all.scss > /dev/null
```

.github/workflows/sass.yaml
```diff
@@ -1,94 +1,115 @@
...
+
+  workflow_dispatch:
+
+concurrency:
+   group: sass-${{ github.ref_name }}
```
injms, line 15: Does a fallback for `ref_name` need to be used for pushes without a pull request? [^1]

[^1]: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-a-fallback-value

colinrotherham: Hi @injms, thanks for watching out for us 😊 

Did you see I'd used `${{ github.ref_name }}` **refs/heads/main** instead of `${{ github.head_ref }}`?

Appreciate how the docs would give a nicer result though, happy to change

```suggestion
   group: sass-${{ github.head_ref || github.ref_name }}
```

injms: So many similarly-named-variables! Apologies, I didn't spot that you'd used `ref_name`, so clearly no need to change it 👍

colinrotherham: To be fair, was a good spot!

Two merges into `main` would share the same `${{ github.ref_name }}` so the 2nd push would cancel the 1st

I've used [your (and GitHub's) suggestion](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-a-fallback-value) to fall back to the (always-unique) `github.run_id` outside PRs

.github/workflows/tests.yml
```diff
@@ -52,4 +67,5 @@ jobs:
       - name: Send screenshots to Percy
         run: npx percy exec -- jest --color --selectProjects "JavaScript component tests"
         env:
+          PERCY_PARTIAL_BUILD: 1
```
colinrotherham, line 70: The [`PERCY_PARTIAL_BUILD` environment variable](https://docs.percy.io/docs/environment-variables#optional) lets Percy know we won't always be sending the full 64 snapshots as we're intentionally holding back various "no-js" screenshots

It'll only flag changed screenshots for review, rather than missing ones

jest.config.js
```diff
@@ -19,7 +19,7 @@ module.exports = {
         'jest-serializer-html'
       ],
       testMatch: [
-        '**/template.test.js'
+        '**/(*.)?template.test.js'
```
colinrotherham, line 22: Allows both:

```
template.test.js
example.template.test.js
```

lib/.eslintrc.js
```diff
@@ -1,7 +1,7 @@
 module.exports = {
   overrides: [
     {
-      files: ['jest-*'],
+      files: ['puppeteer-*'],
```
colinrotherham, line 4: Only the `puppeteer-*` file(s) need `page` and `browser` globals ignoring

lib/puppeteer-helpers.js
```diff
@@ -0,0 +1,56 @@
...
+ * @param {Object} options.nunjucksParams - Params passed to the Nunjucks macro
+ * @param {Object} [options.javascriptConfig] - The configuration hash passed to
+ *  the component's class for initialisation
+ * @param {Function} [options.initialiser] - A function that'll run in the
+ *  browser to execute arbitrary initialisation. Receives an object with the
+ *  passed configuration as `config` and the PascalCased component name as
+ *  `componentClassName`
+ * @returns {Promise<import('puppeteer').Page>}
+ */
+async function renderAndInitialise (componentName, options = {}) {
```
colinrotherham, line 28: Moved from [lib/jest-helpers.js](https://github.com/alphagov/govuk-frontend/pull/2878/files#diff-35cd93e336244402cc677971c753a8e36bc1ad944af90ef2b99369e1a5ba6158) due to missing `page` global

Tests vary by `@jest-environment`

```mjs
/**
 * @jest-environment puppeteer
 */
```

```mjs
/**
 * @jest-environment jsdom
 */
```

.github/workflows/tests.yml
36degrees: Assuming we don't plan to send 'no-js' screenshots for those components again, can we not remove them instead?

Without this I'm assuming it'd get flagged in Percy but we could still approve and they'd be removed from the set of baseline screenshots used for comparison?

src/govuk/components/all.template.test.js
```diff
@@ -0,0 +1,148 @@
...
+// We can't use the render function from jest-helpers, because we need control
+// over the nunjucks environment.
+const nunjucks = require('nunjucks')
+
+const { allComponents, getComponentData } = require('../../../lib/file-helper')
+const { nunjucksEnv, renderSass, renderHtml } = require('../../../lib/jest-helpers')
+
+const configPaths = require('../../../config/paths.js')
+
+describe('Components', () => {
```
colinrotherham, line 13: Separated Nunjucks and HTML tests from [src/govuk/components/all.test.js](https://github.com/alphagov/govuk-frontend/pull/2878/files#diff-e5806587e66dd93c9ab15b1591df5f8fbc6c631ec228c68518d68f93ad69e50c) to avoid loading the `@jest-environment puppeteer` (wasted a perfectly good Chromium window)

.github/workflows/tests.yml
colinrotherham: I don't know enough about Percy really, what do you think we should do? Future thinking for `PERCY_PARTIAL_BUILD` was so we could enable `jest --changedSince=origin/main`

We might not need to send any screenshots if nothing has changed, but one for another PR

Shall I remove `PERCY_PARTIAL_BUILD` for now?

36degrees: I think so – I am making assumptions about how Percy works too but I imagine we'll be able to approve it with the 'no-js' screenshots missing and then they'll be removed from the base build and forgotten about.

Otherwise they'll be hanging around forever as 'things we used to send'.

.github/workflows/sass.yaml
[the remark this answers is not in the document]
36degrees: All of the tests should be required… 'Run tests' is in that list of required status checks.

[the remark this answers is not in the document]
colinrotherham: Oh yeah, I'll mark this as resolved 😊

.github/workflows/tests.yml
colinrotherham: Thanks @36degrees I've removed `PERCY_PARTIAL_BUILD`

36degrees (commented):
Thanks for this, @colinrotherham 🙌🏻 

Generally looks good, but there is a lot going on in this PR – I might be missing something but the changes to the workflows feel pretty much unrelated to lots of the changes to the tests, for example.

The work is logically broken down into sensible commits though, so I don't think it's worth splitting it out at this stage.

It's really helpful to have extra context added through the review comments, but ideally I think at least some of it should be in the commit messages, which don't always explain _why_ a change is happening.

.github/workflows/sass.yaml
```diff
@@ -1,94 +1,115 @@
...
+on:
+  pull_request:
+
+  push:
+    branches:
+      - main
+      - 'feature/**'
+      - 'support/**'
+
+  workflow_dispatch:
```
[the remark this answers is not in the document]
colinrotherham: Yeah, with this PR we've lost the ability to test branches that aren't yet pull requests

So you can manually run **Sass** by picking a branch from a list

src/govuk/components/all.test.js
```diff
@@ -4,139 +4,53 @@
...
-  })
-})
-
-describe.each(allComponents)('%s', (component) => {
+describe('Components', () => {
   let percySnapshot
+  let componentsJavaScript
 
-  beforeAll(() => {
+  beforeAll(async () => {
```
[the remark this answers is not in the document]
colinrotherham: Yeah, agree with you about them not being tests (no assertions)

Mind if we solve that problem in another PR?

Chuck some alternative file naming ideas at me though as your naming conventions are new to me 😊

.github/workflows/sass.yaml
```diff
@@ -1,94 +1,115 @@
...
-      - run: |
-          npm install -g node-sass@v3.4.0
-          node-sass --version
-      - run: time node-sass src/govuk/all.scss > /dev/null
-
-  # Node Sass v7.x = LibSass v3 latest
-  lib-sass-latest:
-    name: LibSass v3 (latest, deprecated)
+  node:
+    name: ${{ matrix.name }}
```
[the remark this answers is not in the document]
colinrotherham: Ah that's fair, it's the opposite for me

Not so much in [Sass](https://github.com/alphagov/govuk-frontend/actions/workflows/sass.yaml) (this file) as none of the jobs have any dependencies, but imagine the approach in [Tests](https://github.com/alphagov/govuk-frontend/actions/workflows/tests.yml)

Once a project is 1) installed and 2) built you generally have three outputs:

1. Installed packages **`node_modules`**
1. Compiled files `./src` to **`./dist`**
2. Support files (or caches) such as [Import Maps](https://wicg.github.io/import-maps/) or [`.tsbuildinfo`](https://devblogs.microsoft.com/typescript/announcing-typescript-3-4/#faster-subsequent-builds-with-the---incremental-flag)

You can capture all these once, up front, then run all subsequent tasks in parallel via `matrix`

Each matrix job doesn't have to install/build/cache every time

For example an army of [Browserstack](https://github.com/marketplace/actions/browserstack-actions), [Cypress](https://github.com/marketplace/actions/cypress-io) or [Sauce Labs](https://github.com/marketplace/actions/sauce-labs-app-automate-action) browser test runners all sharing a unique `npm start` Express.js instance but from the same cached build

I'll move the commit to another branch though as it doesn't need to be in this PR

.github/workflows/tests.yml
```diff
@@ -1,6 +1,21 @@
 name: Tests
 
-on: [push, pull_request]
+on:
+  pull_request:
+    paths-ignore:
+      - 'docs/**'
```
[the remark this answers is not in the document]
colinrotherham: Ahh of course, will remove as it was only a bonus idea

src/govuk/components/all.template.test.js
```diff
@@ -0,0 +1,148 @@
...
+      allComponents.forEach((component) => {
+        expect(renderSass({ file: `${configPaths.src}/components/${component}/_${component}.scss` })).resolves.toEqual(
+          expect.objectContaining({
+            css: expect.any(Object),
+            stats: expect.any(Object)
+          })
+        )
+      })
+    })
+  })
```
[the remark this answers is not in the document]
colinrotherham: Yeah this is what @romaricpascal was mentioning via https://github.com/alphagov/govuk-frontend/pull/2878#pullrequestreview-1119840159

Are you happy with `describe('Nunjucks render')` and `describe('HTML validation')` staying in here? It's those two that gave me the steer to it being named **all.template.test.js**

Where should `describe('Sass render')` go?

Give us a file name 😆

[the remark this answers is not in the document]
36degrees: > Are you happy with `describe('Nunjucks render')` and `describe('HTML validation')` staying in here? It's those two that gave me the steer to it being named **all.template.test.js**

Yes, I think that broadly makes sense 👍🏻 

> Where should `describe('Sass render')` go?
> 
> Give us a file name 😆

This is what I was trying to flag by https://github.com/alphagov/govuk-frontend/pull/2878#discussion_r982253568 - I think `all.test.js` would be the most consistent place for these to live, except that name's already taken by the Percy screenshot 'tests'.

If we're open to renaming that file to something else (maybe `percy.test.js` or `visual-regression.test.js` then I think the Sass render tests should live back in `all.test.js`.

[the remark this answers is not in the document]
colinrotherham: @36degrees Yeah, it's tricky isn't it

Are you definitely happy with an exact match on `percy.test.js` or `visual-regression.test.js` or should we pick a new naming convention specifically for **"Needs test site"** and **"Needs Puppeteer"**?

For context, these are our current ones:

## **Gulp tasks**
Run with `npx jest --color --selectProjects "Gulp tasks"`

**Test site:** No
**Browser:** No
**Environment:** Node.js

```
'**/gulp/**/*.test.{js,mjs}'
```

## **Nunjucks macro tests**
Run with `npx jest --color --selectProjects "Nunjucks macro tests"`

**Test site:** No
**Browser:** No
**Environment:** jsdom (fake browser)

```
'!**/template.test.js'
```

## **JavaScript unit tests**
Run with `npx jest --color --selectProjects "JavaScript unit tests"`

**Test site:** No
**Browser:** No
**Environment:** Node.js

```
'**/*.unit.test.{js,mjs}'
```

## **JavaScript behaviour tests**
Run with `npx jest --color --selectProjects "JavaScript behaviour tests"`

**Test site:** Yes
**Browser:** No
**Environment:** Node.js
```
'**/*.test.js'
```

## **JavaScript component tests**
Run with `npx jest --color --selectProjects "JavaScript component tests"`

**Test site:** Yes
**Browser:** Yes
**Environment:** Puppeteer
```
'**/all.test.js',
'**/components/**/*.test.js'
```

colinrotherham:
> Thanks for this, @colinrotherham 🙌🏻
> 
> Generally looks good, but there is a lot going on in this PR – I might be missing something but the changes to the workflows feel pretty much unrelated to lots of the changes to the tests, for example.
> 
> The work is logically broken down into sensible commits though, so I don't think it's worth splitting it out at this stage.
> 
> It's really helpful to have extra context added through the review comments, but ideally I think at least some of it should be in the commit messages, which don't always explain _why_ a change is happening.

Thanks @36degrees

The Sass workflow `matrix` has gone, but I've [explained why we need `workflow_dispatch`](https://github.com/alphagov/govuk-frontend/pull/2878#discussion_r982290373)

I've pushed up again with commit comments too

Will take another look at naming conventions

src/govuk/components/all.template.test.js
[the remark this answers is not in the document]
36degrees: I think the closest equivalent we have would be:

- src/govuk/helpers/helpers.test.js
- src/govuk/settings/settings.test.js
- src/govuk/tools/tools.test.js

So this should probably be in `src/govuk/components/components.test.js`

.github/workflows/tests.yml
colinrotherham: Going to mark this resolved, mainly because it's gone 🚀

src/govuk/components/all.test.js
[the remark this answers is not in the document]
colinrotherham: Morning @36degrees

Sass tests now live in **`src/govuk/components/components.test.js`**

src/govuk/components/all.template.test.js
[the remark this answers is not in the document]
colinrotherham: I've gone with the following for now:

**`src/govuk/components/all.test.js`** — Percy
**`src/govuk/components/components.test.js`** — Sass ([via comment](https://github.com/alphagov/govuk-frontend/pull/2878#discussion_r983166068))
**`src/govuk/components/components.template.test.js`** — Nunjucks + HTML validation

Did you still want **`all.test.js`** to be either `percy.test.js` or `visual-regression.test.js`?

Let me know, think we're nearly there 😊

colinrotherham:
> Will take another look at naming conventions

@36degrees Ready for you to take a look at again, with one last question about **`all.test.js`** (Percy)

It's still the original filename (before this PR) but happy to change it if it's a good time to do so 👍

src/govuk/components/all.template.test.js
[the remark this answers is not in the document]
36degrees: As discussed on Slack, I think we can leave it as `all.test.js` for now 👍🏻

config/jest/browser/open.mjs
```diff
@@ -7,6 +7,14 @@ import { setup } from 'jest-environment-puppeteer'
  * @param {import('jest').Config} jestConfig
  */
 export default async function browserOpen (jestConfig) {
+  const { maxWorkers } = jestConfig
+
+  // Increase maximum Node.js listeners for browserPerWorker
```
36degrees, line 12: I can sort of follow this but I think this would be a more useful comment if it covered what the number of listeners is increased to, and why?

Is this basically making sure we have a corresponding listener for each worker, with the number of workers coming from the Jest config? But then we need to add 1 for some reason – the main thread, or just it 'just in case'?

src/govuk/components/all.test.js
```diff
@@ -4,139 +4,53 @@
...
-    await percySnapshot(page, 'no-js: ' + component)
-  })
-
-  it('generate screenshots for Percy visual regression, with JavaScript enabled', async () => {
-    await page.setJavaScriptEnabled(true)
-    await page.goto(baseUrl + '/components/' + component + '/preview', { waitUntil: 'load' })
-    await percySnapshot(page, 'js: ' + component)
-  })
+    // Filter "JavaScript enabled" components only
+    componentsJavaScript = allComponents
```
36degrees, line 26: For clarity:

```suggestion
    componentsWithJavaScript = allComponents
```

src/govuk/components/components.template.test.js
```diff
@@ -0,0 +1,124 @@
...
+      const componentTasks = allComponents.map(async (component) => {
+        const { examples } = await getComponentData(component)
+
+        // Loop through component examples
+        examples.forEach(({ name, data }) => {
+          const html = renderHtml(component, data)
+          const report = validator.validateString(html)
+
+          expect({ component, name, report })
+            .toEqual({ component, name, report: expect.objectContaining({ valid: true }) })
```
36degrees, line 117: As flagged on Slack, this doesn't appear to be working as intended – if I intentionally make a component output invalid HTML the tests are still passing.

config/jest/browser/open.mjs
colinrotherham: Morning @36degrees can you take another look?

Hopefully explained it a bit better in the comment and commit

src/govuk/components/components.template.test.js
colinrotherham: Thanks for spotting this one! A bit of digging revealed `new HtmlValidate()` can't just take a "config override", it also needs the default `extends: ['html-validate:recommended']` preset too which describes the test better

Fixed now 🙈 

It also looks like you can pass the **HtmlValidate** instance into `.toHTMLValidate()` so I might go back to doing that, as this will keep the nicely formatted Jest output

colinrotherham:
Morning @36degrees

I've gone through and split out all the commits, explaining what's happening and why

If you can take another look please 😊

src/govuk/components/components.template.test.js
36degrees: Ace, that sounds like a good improvement 👍🏻

colinrotherham: Ahh I'm afraid it doesn't support that. It's possible to pass in a config, but not a validator

Are you happy leaving it as a diff of the whole report object? It's definitely nerdier, and you get the full HTML source showing up in the results array (quite useful).

### Before
```
Result should be valid but had error "Unexpected close-tag, expected opening tag."
```

### After
Image showing HTML validator report object diff

PR #2879, opened 2022-09-22
Fix missing assertion in header tests

Top-Cat:
I do love it when my tests don't test anything

Also fixes the css selector used by changing from `a` to `li` as this element is not a link.

### DIFF
--- a/src/govuk/components/header/template.test.js
+++ b/src/govuk/components/header/template.test.js
@@ -165,8 +165,8 @@ describe('header', () => {
     it('renders navigation with active item', () => {
       const $ = render('header', examples['with navigation'])
 
-      const $activeItem = $('a.govuk-header__navigation-item:first-child')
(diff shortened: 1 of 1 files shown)


colinrotherham:
Well spotted @Top-Cat, would you mind updating the PR description showing that this PR fixes the missing assertion but also the [incorrect CSS selector for `--active` in **template.njk**](https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/components/header/template.njk#L71)?

I'll share this with the team, but I'll be happy to have this merged. Thank you

owenatgov (approved):
Looks good to me @Top-Cat! Just needs a 2nd approval fro mthe team and then we can merge.

I also think this could do with a fix changelog entry, but I or another team member can pick this up.

colinrotherham (approved):

owenatgov:
This actually doesn't need a changelog entry since we don't ship tests to users 😅 Merging!

PR #2880, opened 2022-09-26
Fix Jest setup/teardown for test server awaiting `SIGTERM`

colinrotherham:
This PR fixes an issue with Jest in watch mode:

```console
npx jest --watch src/govuk/components/button/button.test.js
```

When watching [**JavaScript component tests** project files](https://github.com/alphagov/govuk-frontend/blob/main/jest.config.js#L58) the array returned via [`getServers()` from **jest-dev-server**](https://github.com/smooth-code/jest-puppeteer/blob/6f16345622c91487b2704bf30350a0e57114c2be/packages/jest-dev-server/src/global.js#L231) appears to list processes that have terminated (or are currently awaiting termination).

We should check for `signalCode: 'SIGTERM'` processes before assuming the test site is running

### DIFF
--- a/config/jest/browser/open.mjs
+++ b/config/jest/browser/open.mjs
@@ -3,8 +3,10 @@ import { setup } from 'jest-environment-puppeteer'
 
 /**
  * Open browser
+ *
+ * @param {import('jest').Config} jestConfig
  */
-export default async function browserOpen () {
+export default async function browserOpen (jestConfig) {
   await serverStart() // Wait for web server
-  await setup() // Open browser
+  await setup(jestConfig) // Open browser
 }
--- a/config/jest/server/start.mjs
+++ b/config/jest/server/start.mjs
@@ -1,5 +1,6 @@
 import { getServers, setup } from 'jest-dev-server'
 import waitOn from 'wait-on'
+import serverStop from './stop.mjs'
 
 import configPaths from '../../paths.js'
 const { PORT = configPaths.ports.test } = process.env
@@ -8,22 +9,26 @@ const { PORT = configPaths.ports.test } = process.env
  * Start web server
  */
 export default async function serverStart () {
-  const [server] = await getServers()
+  const servers = await getServers()
 
-  // Server already running
-  if (server) {
-    return
+  // Server start timeout
+  let timeout = 1000
+
+  // Server stopping?
+  if (servers.some(({ signalCode }) => signalCode === 'SIGTERM')) {
+    await serverStop() // Wait for server to stop
+    timeout = 0 // No need to wait for start
   }
 
-  // Wait until ready
+  // Wait until ready check
   const ready = ({ timeout = 30000 } = {}) => waitOn({
     resources: [`tcp:${PORT}`],
     timeout
   })
 
-  // Start server
+  // Wait until ready (or start up)
   try {
-    await ready({ timeout: 1000 })
+    await ready({ timeout })
   } catch (error) {
     await setup({ command: `PORT=${PORT} node app/start.js`, port: PORT })
     await ready()
--- a/jest-puppeteer.config.js
+++ b/jest-puppeteer.config.js
@@ -1,6 +1,6 @@
 module.exports = {
   browserContext: 'incognito',
-  dumpio: true,
+  browserPerWorker: true,
   launch: {
     // we use --no-sandbox --disable-setuid-sandbox as a workaround for the
     // 'No usable sandbox! Update your kernel' error
--- a/src/govuk/components/button/button.test.js
+++ b/src/govuk/components/button/button.test.js
@@ -59,8 +59,8 @@ describe('/components/button', () => {
      *
      * Examples don't do this and we need it to have something to submit
      */
-    async function trackClicks () {
-      page.evaluate(() => {
+    function trackClicks (page) {
+      return page.evaluate(() => {
         const $button = document.querySelector('button')
         const $form = document.createElement('form')
         $button.parentNode.appendChild($form)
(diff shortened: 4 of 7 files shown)


config/jest/server/start.mjs
```diff
@@ -10,9 +10,12 @@ const { PORT = configPaths.ports.test } = process.env
 export default async function serverStart () {
   const [server] = await getServers()
```
romaricpascal, line 11: Is there ever a chance (race condition with tests re-running quickly in watch mode or something else) we could have more than one server up? I'm guessing thanks to the `await server.destroy()` that's been introduced there's only ever going to be one server in that list, just wanted to double-check 😄

colinrotherham: Multiple servers are supported but we only use one

I've had a look and calling [`server.destroy()`](https://github.com/smooth-code/jest-puppeteer/blob/master/packages/spawnd/src/index.js#L27) _does_ actually fully exit the process

So instead of checking if the first `[0]` process in **getServers()** is "exited but not removed", it's probably easier to remove all of them via **serverStop()** instead. I'll add it to the next push

(Just in case multiple servers ever get started)

config/jest/server/stop.mjs
```diff
@@ -1,8 +1,11 @@
...
+import { getServers, teardown } from 'jest-dev-server'
 
 /**
  * Stop web server
  */
 export default async function serverStop () {
   await teardown()
+
+  // Clear down server list
+  ;(await getServers()).length = 0
```
colinrotherham, line 10: Bit ugly that StandardJS's "no semicolons" prevents this looking nice

@romaricpascal Let me know if you want it reformatting:

```mjs
// Clear down server list
const servers = await getServers()
servers.length = 0
```

Uses the array `.length = 0` shorthand to delete all items

jest-puppeteer.config.js
```diff
@@ -1,6 +1,6 @@
 module.exports = {
   browserContext: 'incognito',
-  dumpio: true,
+  browserPerWorker: true,
```
colinrotherham, line 3: Found another useful speed up whilst debugging

We can change Puppeteer `setup()` → **`setup(jestConfig)`**

It'll then use a separate browser per test worker

jest-puppeteer.config.js
```diff
@@ -1,6 +1,6 @@
 module.exports = {
   browserContext: 'incognito',
-  dumpio: true,
```
colinrotherham, line 3: I've taken this out as it's supposed to be in `launch` so wasn't working. Can be useful for debugging but adds a lot of noise

config/jest/server/start.mjs
```diff
@@ -8,11 +9,15 @@ const { PORT = configPaths.ports.test } = process.env
...
-  // Server already running
-  if (server) {
-    return
+  // Connection timeout
+  let timeout = 1000
+
+  // Server already running? Wait for exit
+  if (servers.some(({ signalCode }) => signalCode === 'SIGTERM')) {
+    await serverStop()
+    timeout = 0
```
colinrotherham, line 20: We don't need to wait for the port to open when we know it's just closed

config/jest/browser/open.mjs
```diff
@@ -3,8 +3,10 @@ import { setup } from 'jest-environment-puppeteer'
...
 /**
  * Open browser
+ *
+ * @param {import('jest').Config} jestConfig
  */
-export default async function browserOpen () {
+export default async function browserOpen (jestConfig) {
   await serverStart() // Wait for web server
-  await setup() // Open browser
+  await setup(jestConfig) // Open browser
```
colinrotherham, line 11: This enables Jest to open multiple Chrome windows, one per worker

src/govuk/components/button/button.test.js
```diff
@@ -59,8 +59,8 @@ describe('/components/button', () => {
      *
      * Examples don't do this and we need it to have something to submit
      */
-    async function trackClicks () {
-      page.evaluate(() => {
+    function trackClicks (page) {
```
colinrotherham, line 62: With (potentially) multiple pages or tabs open at once, we need to pass in which one we use

colinrotherham (commented):
Morning @romaricpascal this PR is now ready for review again

config/jest/server/start.mjs
romaricpascal: That's probably worth putting a reminder in the code for the future

```suggestion
    // We don't need to wait for the port to open when we know it's just closed
    timeout = 0
```

jest-puppeteer.config.js
romaricpascal: Sounds like this option will get us a long way. So from what I gathered (let me know if I misunderstood), Jest:
- spins a worker per test file, and with this new option, attach a pristine browser tab to it
- inside the worker runs tests from the file sequentially (prevents issues when tests rely on DB queries for ex), unless explicitely using `it.concurrent` (which [is experimental](https://jestjs.io/docs/api#testconcurrentname-fn-timeout) anyway and probably shouldn't use because of the shared tab)

It seems to me that we shouldn't have any tests stepping on each others toes with that. If it means passing the `page` to helper functions, that sounds pretty good to me.

Only question is whether it also starts one server per worker of if it's still a unique server running for all the tests (we don't really need more than one). I saw the code checking for live servers was updated a bit further up but don't know if it's for "general safety" or because we would now have multiple servers.

romaricpascal (commented):
Sounds great that both the watch and test stepping on each other's tabs will be fixed by that 😄

jest-puppeteer.config.js
colinrotherham: Just one server

For any **JavaScript behaviour tests** project run (single, multiple, pattern, entire suite) It'll start once at the beginning of your `npx jest` run then close at the end 😊

This is because we've used "global setup" and "global teardown"

It's also true for `--watch` mode with the server starting up again when a file changes

I briefly looked into keeping the server running for the entire watch session until **Control + C** `SIGINT` [but there are bugs](https://github.com/smooth-code/jest-puppeteer/issues/305).

In terms of spinning up worker processes, I don't exactly know how it does it behind the scenes, but sounds about right 😆 

It'll vary sometimes as Jest reduces workers in `--watch` mode to keep CPU usage down, but now we're doing things the correct way we're all set up for any scenario, even `it.concurrent()` 🙌

config/jest/server/start.mjs
colinrotherham: I've put the comments together like this:

```mjs
await serverStop() // Wait for server to stop
timeout = 0 // No need to wait for start
```

romaricpascal (approved):

Issue #2469, opened 2021-12-09
Upgrade to postcss version 8 is blocked on oldie


lfdebrux:
### Cause

The package [oldie](https://www.npmjs.com/package/oldie) was last published 6 years ago and does not appear to be maintained. It and its dependents require postcss version 5. 

We currently use gulp-postcss for our Sass to minified CSS pipeline, which tries to arrange things so that the CSS is parsed only once, and so uses its own version of postcss to run any plugins.

When running `gulp scss:compile` with the latest version of gulp-postcss and postcss we get the following error:

```
Unknown error from PostCSS plugin. Your current PostCSS version is 8.4.4, but postcss-unmq uses 5.2.18. Perhaps this is the source of the error below.
Cannot read property '1' of null
```

### Consequences

We cannot upgrade to postcss 8 without finding a replacement for oldie. postcss 8 has a number of security fixes and advantages, including new best practice where plugins are asked to use it as a peer dependency rather than as a direct dependency, which results in a smaller dependency tree for us.

### Impact of debt

Low

### Effort to pay down

Medium

### Reason (effort to pay down)

~We could fork oldie, or use a vendored version, to get around the requirement for older dependencies. I think the code in oldie should still work, it's just tying a number of other plugins together.~ Looked into this, most of the dependencies for oldie are also unmaintained 😢.

### Overall rating

Low

36degrees:
I did a little investigation and tried to understand what effect each of [the transformations built into `oldie`](https://github.com/csstools/oldie#oldie-) is having on the outputted CSS.

I disabled each plugin in turn, re-compiled the dist folder and compared the outputted CSS to a version generated with the current plugin config.

## Salvage :nth-child (`nth`)

Disabling this plugin makes no difference to the compiled CSS.

## Replace :root (`root`)

Disabling this plugin makes no difference to the compiled CSS.

## Reduce calc() (`calc`)

Disabling this plugin makes no difference to the compiled CSS.

## Resolve rem (`rem`)

This plugin is [already disabled in the config](https://github.com/alphagov/govuk-frontend/blob/693ff651fb8e8a44395ffc790bb5a70ccea19db7/tasks/gulp/compile-assets.js#L96).

## Patch pseudo-elements (`pseudo`)

This plugin is [already disabled in the config](https://github.com/alphagov/govuk-frontend/blob/693ff651fb8e8a44395ffc790bb5a70ccea19db7/tasks/gulp/compile-assets.js#L98).

[...]

PR #2881, opened 2022-09-27
Add `postcss.config.js`/`.browserslistrc` configs

colinrotherham:
This PR adds two new config files `postcss.config.js` and `.browserslistrc`

### PostCSS config via `postcss.config.js`
To make it simpler to configure our CSS output the following `postcss()` calls now share the same config

* [Adding CSS prefixes to Sass files in **./package**](https://github.com/alphagov/govuk-frontend/blob/main/tasks/gulp/copy-to-destination.js#L55)
* [Autoprefixer + CSSNANO for **./dist** output](https://github.com/alphagov/govuk-frontend/tree/bfccd7e9c31e1a051f576c2d7f191d478bb7f9f6/tasks/gulp/compile-assets.js#L56)
* [Autoprefixer + CSSNANO + "Old IE" for **./dist** (IE 8 styles)](https://github.com/alphagov/govuk-frontend/tree/bfccd7e9c31e1a051f576c2d7f191d478bb7f9f6/tasks/gulp/compile-assets.js#L82)
* [Autoprefixer + 'companion' pseudo-selector classes for **./public**](https://github.com/alphagov/govuk-frontend/tree/bfccd7e9c31e1a051f576c2d7f191d478bb7f9f6/tasks/gulp/compile-assets.js#L60)
* [Autoprefixer + 'companion' pseudo-selector classes for **./public** (Legacy styles)](https://github.com/alphagov/govuk-frontend/tree/bfccd7e9c31e1a051f576c2d7f191d478bb7f9f6/tasks/gulp/compile-assets.js#L118)
* [Autoprefixer + 'companion' pseudo-selector classes + "Old IE" for **./public** (Legacy + IE8 styles)](https://github.com/alphagov/govuk-frontend/tree/bfccd7e9c31e1a051f576c2d7f191d478bb7f9f6/tasks/gulp/compile-assets.js#L133)
* [Autoprefixer + "Old IE" for **./public** (IE 8 styles)](https://github.com/alphagov/govuk-frontend/tree/bfccd7e9c31e1a051f576c2d7f191d478bb7f9f6/tasks/gulp/compile-assets.js#L93)

### Browser config via `.browserslistrc`
It's nice to have a separate shareable config which could also be used by Babel, Jest, ESLint, in future

Breaking changes have been moved into a second PR https://github.com/alphagov/govuk-frontend/pull/2924

### DIFF
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -1,0 +1,14 @@
+# This list builds on the GOV.UK service manual's browser testing recommendations
+# https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices
+
+> 0.1%
+last 2 Chrome versions
+last 2 Firefox versions
+last 2 Edge versions
+last 2 Samsung versions
+Safari >= 9
+ie 8-11
+iOS >= 9
+
+[node]
+node 16
--- a/package.json
+++ b/package.json
@@ -98,15 +98,5 @@
   },
   "optionalDependencies": {
     "fsevents": "*"
-  },
-  "browserslist": [
-    ">0.1%",
-    "last 2 Chrome versions",
-    "last 2 Firefox versions",
-    "last 2 Edge versions",
-    "last 2 Samsung versions",
-    "Safari >= 9",
-    "ie 8-11",
-    "iOS >= 9"
-  ]
+  }
 }
--- a/postcss.config.js
+++ b/postcss.config.js
@@ -1,0 +1,59 @@
+const { parse } = require('path')
+const autoprefixer = require('autoprefixer')
+const cssnano = require('cssnano')
+const oldie = require('oldie')
+const pseudoclasses = require('postcss-pseudo-classes')
+
+/**
+ * PostCSS config
+ *
+ * @param {object} context - PostCSS context
+ * @param {string} context.env - Browserslist environment
+ * @param {import('vinyl')} context.file - File object
+ */
+module.exports = ({ env, file }) => {
+  const { dir, name } = parse(file.path)
+
+  const plugins = [
+    autoprefixer({ env })
+  ]
+
+  // Minify CSS
+  if (name.endsWith('.min')) {
+    plugins.push(cssnano())
+  }
+
+  // Add review app auto-generated 'companion' classes for each pseudo-class
+  // For example ':hover' and ':focus' classes to simulate form label states
+  if (dir.endsWith('app/assets/scss') && (name === 'app' || name.startsWith('app-'))) {
+    plugins.push(pseudoclasses({
+      // Work around a bug in pseudo classes plugin that badly transforms
+      // :not(:whatever) pseudo selectors
+      blacklist: [
+        ':not(',
+        ':disabled)',
+        ':first-child)',
+        ':last-child)',
+        ':focus)',
+        ':active)',
+        ':hover)'
+      ]
+    }))
+  }
+
+  // Transpile CSS for Internet Explorer
+  if (name.endsWith('-ie8') || name.endsWith('-ie8.min')) {
+    plugins.push(
+      oldie({
+        rgba: { filter: true },
+        rem: { disable: true },
+        unmq: { disable: true },
+        pseudo: { disable: true }
+      })
+    )
+  }
+
+  return {
+    plugins
+  }
+}
--- a/postcss.config.js
+++ b/postcss.config.js
@@ -9,10 +9,10 @@ const pseudoclasses = require('postcss-pseudo-classes')
  *
  * @param {object} context - PostCSS context
  * @param {string} context.env - Browserslist environment
- * @param {import('vinyl')} context.file - File object
+ * @param {string | import('vinyl')} [context.file] - File path or object
  */
-module.exports = ({ env, file }) => {
-  const { dir, name } = parse(file.path)
+module.exports = ({ env, file = '' }) => {
+  const { dir, name } = parse(typeof file === 'object' ? file.path : file)
 
   const plugins = [
     autoprefixer({ env })
--- a/postcss.config.unit.test.js
+++ b/postcss.config.unit.test.js
@@ -1,0 +1,145 @@
+const Vinyl = require('vinyl')
+const configFn = require('./postcss.config.js')
+
+describe('PostCSS config', () => {
+  let env
+
+  function getPluginNames ({ plugins }) {
+    return plugins.map(({ postcssPlugin }) => postcssPlugin)
+  }
+
+  beforeAll(() => {
+    env = 'production'
+  })
+
+  describe('Default', () => {
+    it.each(
+      [
+        { path: 'example.css' },
+        { path: 'example.scss' }
+      ]
+    )('Adds plugins for $path', ({ path }) => {
+      const input = new Vinyl({ path })
+
+      // Confirm plugins for both file object and path
+      for (const file of [input, input.path]) {
+        const config = configFn({ env, file })
+
+        expect(getPluginNames(config))
+          .toEqual(['autoprefixer'])
+      }
+    })
+  })
+
+  describe('Default + IE8', () => {
+    it.each(
+      [
+        { path: 'example-ie8.css' },
+        { path: 'example-ie8.scss' }
+      ]
+    )('Adds plugins for $path', ({ path }) => {
+      const input = new Vinyl({ path })
+
+      // Confirm plugins for both file object and path
+      for (const file of [input, input.path]) {
+        const config = configFn({ env, file })
+
+        expect(getPluginNames(config))
+          .toEqual([
+            'autoprefixer',
+            'oldie'
+          ])
+      }
+    })
+  })
+
+  describe('Default + Minification', () => {
+    it.each(
+      [
+        { path: 'example.min.css' },
+        { path: 'example.min.scss' }
+      ]
+    )('Adds plugins for $path', ({ path }) => {
+      const input = new Vinyl({ path })
+
+      // Confirm plugins for both file object and path
+      for (const file of [input, input.path]) {
+        const config = configFn({ env, file })
+
+        expect(getPluginNames(config))
+          .toEqual([
+            'autoprefixer',
+            'cssnano'
+          ])
+      }
+    })
+  })
+
+  describe('Default + Minification + IE8', () => {
+    it.each(
+      [
+        { path: 'example-ie8.min.css' },
+        { path: 'example-ie8.min.scss' }
+      ]
+    )('Adds plugins for $path', ({ path }) => {
+      const input = new Vinyl({ path })
+
+      // Confirm plugins for both file object and path
+      for (const file of [input, input.path]) {
+        const config = configFn({ env, file })
+
+        expect(getPluginNames(config))
+          .toEqual([
+            'autoprefixer',
+            'cssnano',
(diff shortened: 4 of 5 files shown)


package.json
```diff
@@ -29,12 +29,12 @@
     "@percy/cli": "^1.10.2",
     "@percy/logger": "^1.10.2",
     "@percy/puppeteer": "^2.0.2",
-    "autoprefixer": "^9.8.6",
+    "autoprefixer": "^10.4.12",
     "body-parser": "^1.18.3",
-    "cheerio": "^1.0.0-rc.3",
+    "cheerio": "1.0.0-rc.3",
```
colinrotherham, line 34: I've had to pin this version

One of the package updates in this PR also allowed Cheerio to accept a compatible upgrade to **1.0.0-rc.12**

This broke lots of HTML escaping tests

36degrees (commented):
Did you see [the investigation I did into which plugins were actually doing something we needed](https://github.com/alphagov/govuk-frontend/issues/2469#issuecomment-1006658228)?

I think we can probably get rid of `calc`, `unroot`, `unnth` and potentially `unnot` as when I last tested they weren't doing anything.

We might also actually need `postcss-unmq` – although we have `unmq: { disable: true }` in the config, it's using the wrong namespace so that bit of the config isn't actually doing anything 🙃 

Generally it'd be good to compare the built `dist` and `package` folders with and without this change to see what effect it has.

colinrotherham:
> Did you see [the investigation I did into which plugins were actually doing something we needed](https://github.com/alphagov/govuk-frontend/issues/2469#issuecomment-1006658228)?

I did indeed, was very useful (stuck it in the description!)

This PR was more of a like-for-like "as is" spike, just to see if it was possible

Shall I add the `workflow_dispatch` events (run manually) to [**Diff changes to dist**](https://github.com/alphagov/govuk-frontend/actions/workflows/diff-change-to-dist.yaml) like we did here? https://github.com/alphagov/govuk-frontend/pull/2878/commits/396697a260a0e415b5c12dbd7aa94575b61abe75

colinrotherham:
Soon as I add `postcss-unmq` this error from @lfdebrux comes back:

```
Unknown error from PostCSS plugin. Your current PostCSS version is 8.4.16, but postcss-unmq uses 5.2.18. Perhaps this is the source of the error below.
Cannot read property '1' of null
```

But the error comes from [**css-mediaquery**](https://github.com/ericf/css-mediaquery) (See issue: https://github.com/ericf/css-mediaquery/issues/7)

If we tap into their code and add some logging it's caused by:

https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/helpers/_device-pixels.scss#L33
https://github.com/alphagov/govuk_frontend_toolkit/blob/master/stylesheets/_device-pixels.scss#L4

```mjs
{
  expression: '(-o-min-device-pixel-ratio: 20 / 10)',
  RE_MQ_EXPRESSION: /\(\s*([^\s\:\)]+)\s*(?:\:\s*([^\s\)]+))?\s*\)/,
  captures: null
}
```

☝️ It was never expecting media queries to be formatted that way

### But why??
The older `autoprefixer@9.8.6` is removing **-o-min-device-pixel-ratio**
The older `autoprefixer@10.4.12` is preserving **-o-min-device-pixel-ratio**

**Changes** via `npx browserslist`
Querying browserslist for changes only added **`firefox 105`** and modified:

`and_qq 10.4` → **`and_qq 13.1`**
`and_uc 12.12` → **`and_uc 13.4`**

Just a hunch (similar to Babel), but I wonder if `node_modules` Sass imports are excluded from processing now

package.json
colinrotherham: Not needed anymore, I've _only_ loaded the plugins mentioned via https://github.com/alphagov/govuk-frontend/issues/2469#issuecomment-1006658228

colinrotherham:
Solved it! Had another go at this during the Frontend Community meeting 😊 

We were running the `autoprefixer()` plugin with our usual browsers for the IE8 stylesheet

```
> 0.1%
last 2 Chrome versions
last 2 Firefox versions
last 2 Edge versions
last 2 Samsung versions
Safari >= 9
ie 8-11
iOS >= 9
```

But for the IE8 styles we guard with `<!--[if lte IE 8]>` yet still added `-webkit-`, `-ms` prefixes etc

For our Browserslist config I've added a lonely new environment:

```
[oldie]
ie 8
```

So this PR now introduces two "diffs" versus the previous CSS output:

1. Unnecessary prefixes for IE8 styles are gone
2. Selectors for `:not` rulesets are back [as mentioned](https://github.com/alphagov/govuk-frontend/issues/2469#issuecomment-1006658228) by @36degrees 

Ready for review 👍

postcss.config.js
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * PostCSS config
+ *
+ * @type {import('postcss-load-config').ConfigFn}
+ */
+module.exports = ({ env, file }) => {
+  const isIE8 = file.basename.includes('-ie8')
+
+  const plugins = [
+    autoprefixer({ env: isIE8 ? 'oldie' : env })
```
colinrotherham, line 17: The usual `env` is "**production**" or "**development**" based on `NODE_ENV`

colinrotherham:
This is another one linked with https://github.com/alphagov/govuk-frontend/issues/2243

colinrotherham:
Breaking changes moved to https://github.com/alphagov/govuk-frontend/pull/2924

colinrotherham:
With the PostCSS configs separated out the Gulp task can be simplified too https://github.com/alphagov/govuk-frontend/pull/2912

.browserslistrc
```diff
@@ -0,0 +1,13 @@
+# https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices
```
domoscargin, line 1: Nice little touch including this. Do we want to add a quick one-liner about how/why our compat list may differ from the service manual?

I'm thinking of the (exceedingly unlikely?) event that a future dev has a look at this without knowing the history and updates it to match the service manual completely.

postcss.config.js
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * @type {import('postcss-load-config').ConfigFn}
+ */
+module.exports = ({ env, file }) => {
+  const plugins = [
+    autoprefixer({ env })
+  ]
+
+  // Minify CSS
+  if (file.basename.endsWith('.min.css')) {
```
domoscargin, line 17: Neat!

postcss.config.js
```diff
@@ -0,0 +1,54 @@
...
+    plugins.push(cssnano())
+  }
+
+  // Auto-generate 'companion' classes for pseudo-selector states - e.g. a
+  // :hover class you can use to simulate the hover state in the review app
+  if (file.basename.startsWith('app-')) {
```
domoscargin, line 23: The original logic for this looks like it was "[anything that's not the `dist` folder](https://github.com/alphagov/govuk-frontend/blob/023c3bfa60470c9b614d0e34eaf23a2eba119529/tasks/gulp/compile-assets.js#L60)".

I get the other two `if` conditions, but for some reason I'm struggling to figure this change out, so I've probably missed something obvious - could you explain it for me?

.browserslistrc
colinrotherham: @domoscargin Ahh you'll have to teach me as well 😊 

Can you offer a suggestion? If not I can leave this line out for another PR (can add later)

Thanks

domoscargin: Something like "This list builds on the GOV.UK service manual's browser testing recommendations (https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices)"?

postcss.config.js
domoscargin: Ah, no, I should've read the surrounding code - so this stuff is applied to `app/assets/scss/app.scss` in `compileStyles` `app/assets/scss/app-legacy.scss` in `compileLegacy` and `app/assets/scss/app-legacy-ie8.scss` in `compileLegacyIE`.

So my revised query is why `app.scss` isn't included here?

colinrotherham: I've pushed again due to a slight complication

1. Plugin for Gulp `gulp-postcss` populates PostCSS **ctx.file** as type [`import('vinyl')`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/vinyl/index.d.ts#L75)
2. Plugin for webpack `postcss-loader` populates PostCSS **ctx.file** as a file path `string`

Found this out using the webpack example https://github.com/alphagov/govuk-frontend/pull/2869

Additionally the Gulp stream `ctx.file.basename` will be **.css** but other tools pass the original **.scss**

Take a look at the new version 👍 

Thanks @domoscargin

postcss.config.js
colinrotherham: Due to tooling file extension differences (source vs target) this is now `name.endsWith('.min')`

postcss.config.js
```diff
@@ -0,0 +1,59 @@
...
+        ':last-child)',
+        ':focus)',
+        ':active)',
+        ':hover)'
+      ]
+    }))
+  }
+
+  // Transpile CSS for Internet Explorer
+  if (name.endsWith('-ie8')) {
```
36degrees, line 45: I don't think this is doing what we want when running `npm run build:dist` because at that point we're generating a minified CSS file which ends with `-ie8.min`, so this plugin doesn't get added.

colinrotherham: Well spotted, this was a well-intentioned change from yesterday (previously `.includes()`) so I'll put it back!

colinrotherham: Done. For clarity I've matched on both:

```mjs
if (name.endsWith('-ie8') || name.endsWith('-ie8.min')) {
```

Plus I've added tests for each combination we rely on

36degrees (approved):

.browserslistrc
```diff
@@ -0,0 +1,14 @@
...
+last 2 Chrome versions
+last 2 Firefox versions
+last 2 Edge versions
+last 2 Samsung versions
+Safari >= 9
+ie 8-11
+iOS >= 9
+
+[node]
+node 16
```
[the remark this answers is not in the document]
colinrotherham: We can remove `engines` from **package.json** to be fair

They have different purposes:

1. **.browserslistrc** → Target environment
Tells our tooling what features are supported when building

2. **Package `engines`** → Development environment
Tells our local environment what versions are allowed for development

For example, when Node.js 18 LTS lands on the 25th October we'll likely want to increase our `engines` ranges, update tests, GitHub workflows etc, so our contributors can use the latest supported Node.js version.

But for compatibility, any compiled code (Node.js or browser) should still adhere to our build target environments

postcss.config.unit.test.js
```diff
@@ -0,0 +1,145 @@
...
+
+  function getPluginNames ({ plugins }) {
+    return plugins.map(({ postcssPlugin }) => postcssPlugin)
+  }
+
+  beforeAll(() => {
+    env = 'production'
+  })
+
+  describe('Default', () => {
```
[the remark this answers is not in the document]
colinrotherham: Thanks @36degrees

Yeah I didn't optimise by choice, but can come back and improve these in future?

Take a look in https://github.com/alphagov/govuk-frontend/pull/2924 (based on this branch) where we start to add in the environment switching tests for "Old IE"

Issue #2788, opened 2022-08-19
Deprecate compatibility mode variables


owenatgov:
## What
Deprecate the compatibility mode variables:

- `$govuk-compatibility-govukelements`
- `$govuk-compatibility-govuktemplate`
- `$govuk-compatibility-govukfrontendtoolkit`

## Why
We will be removing compatibility mode in 5.0 so we should deprecate these in order to communicate this to service teams.

Part of https://github.com/alphagov/govuk-frontend/issues/2769

## Who needs to work on this
Developers

PR #2882, opened 2022-09-27
Deprecate compatibility mode settings

owenatgov:
Fixes https://github.com/alphagov/govuk-frontend/issues/2788

## What/Why
Deprecates the compatibility mode variables.

Additionally adds a call to `$govuk-suppressed-warnings` in the legacy review app stylesheets to suppress warnings when developing locally.

## Notes
I've attempted to be consistent in my wording with the warnings introduced in https://github.com/alphagov/govuk-frontend/pull/2844.

This also includes a slight history rewrite in a change to the changelog entry for https://github.com/alphagov/govuk-frontend/pull/2844 to reflect a change to the PR's title to better describe the change and differentiate it from this one.

### DIFF
--- a/src/govuk/settings/_compatibility.scss
+++ b/src/govuk/settings/_compatibility.scss
@@ -23,8 +23,16 @@
 ///
 /// @type Boolean
 /// @access public
+/// @deprecated Will be removed in v5.0 with the rest of the compatibility mode
+/// suite of tools and settings
 
 $govuk-compatibility-govukfrontendtoolkit: false !default;
+
+@if $govuk-compatibility-govukfrontendtoolkit == true {
+  @include _warning("compatibility-mode", "$govuk-compatibility-govukfrontendtoolkit " +
+  "is deprecated. From version 5.0, GOV.UK Frontend will remove compatibility " +
+  "with the legacy library govuk_frontend_toolkit.");
+}
 
 /// Compatibility Mode: alphagov/govuk_template
 ///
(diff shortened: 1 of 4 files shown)


owenatgov:
After a chat with @36degrees and @colinrotherham, we've decided we're going to hold off on this until https://github.com/alphagov/govuk-frontend/issues/2890 is complete. The solution of creating a variable to suppress the warnings locally can be solved by the aforementioned issue and the aforementioned issue can also solve the wider problem of allowing users to suppress warnings.

owenatgov:
https://github.com/alphagov/govuk-frontend/issues/2890 is now complete so this has been updated and is ready to be reviewed again.

src/govuk/settings/_compatibility.scss
```diff
@@ -56,16 +72,26 @@ $govuk-compatibility-govuktemplate: false !default;
...
 /// @type Boolean
 /// @access public
+/// @deprecated Will be removed in v5.0 with the rest of the compatibility mode
+/// suite of tools and settings
 
 $govuk-compatibility-govukelements: false !default;
 
+@if $govuk-compatibility-govukelements == true {
+  @include _warning("compatibility-mode", "$govuk-compatibility-govukelements " +
+  "is deprecated. From version 5.0, GOV.UK Frontend will not support compatibility " +
```
colinrotherham, line 82: @owenatgov I think I preferred your wording here? https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/settings/_warnings.scss#L10

**Current**
>…will not support compatibility with the legacy codebase govuk_elements

**Suggested**
>…will remove compatibility with legacy libraries such as govuk_elements

colinrotherham (approved):
Looks fab to me, one comment on wording but not a blocker 😊

src/govuk/settings/_compatibility.scss
owenatgov: Thanks Colin! I made a slight tweak to the suggestion so that each warning roughly reads:

>{variable} is deprecated. From version 5.0, GOV.UK Frontend will remove compatibility with the legacy library {library}.

I've done this because i'd rather keep each warning separate so that users are clearer on what in their codebase is kicking off the warning and why.

PR #2883, opened 2022-09-27
Allow CharacterCount to receive configuration via JavaScript

### LINKED ISSUE #2837: Allow character count config set via data attributes to be passed when initialising the component in JavaScript
## What

The character count component currently uses data attributes to configure the component:

- `data-maxlength` to determine the character limit displayed to the user
- `data-maxwords` to determine the word count displayed to the user, if it makes more sense to count words than characters
- `data-threshold` to determine a cut-off point at which the count message should be displayed

Update the component so that this config can also be passed as an object when initialising the component in JavaScript.

## Why

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [ ] The component can be configured by provided a config object when instantiating a CharacterCount object
- [ ] If a `data-prevent-double-click` data attribute is present on the $module HTML element it overrides any config set via initialisation
- [ ] Tests to cover the above

romaricpascal:
The configuration spree continues! CharacterCount needs its own little behaviour when it comes to merging the data attributes and JavaScript configuration: 

`maxwords` takes precedence over `maxlength`. This means we need to clear some of the JavaScript configuration if any of these options are set through data attributes to avoid a `maxwords` set in JavaScript always winning against a `data-maxlength` attribute:

```js
new CharacterCount($element, {maxwords: 150}).init();
```

```html
<div class="govuk-character-count" data-module="govuk-character-count" data-maxlength="200">
  <!-- Omitted for clarity -->
</div>
```

Commits should paint a step-by-step picture of how things happened. 

~~Keeping this as draft for now as I'm on the fence of whether to rework the tests as unit tests checking that the right values end up in `this.config` rather than run in a browser.~~ Opening to review with puppeteer tests so further work on internationalisation can be carried on. Tests will likely be reworked soon when/if creating a base component in 5.0.

Closes #2837

### DIFF
--- a/src/govuk/all.mjs
+++ b/src/govuk/all.mjs
@@ -21,8 +21,9 @@ import Tabs from './components/tabs/tabs.mjs'
  * @param {HTMLElement} [config.scope=document] - scope to query for components
  * @param {Object} [config.accordion] - accordion config
  * @param {Object} [config.button] - button config
- * @param {Object} [config.notificationBanner] - notification banner config
+ * @param {Object} [config.characterCount] - character count config
  * @param {Object} [config.errorSummary] - error summary config
+ * @param {Object} [config.notificationBanner] - notification banner config
  */
 function initAll (config) {
   config = typeof config !== 'undefined' ? config : {}
--- a/src/govuk/all.mjs
+++ b/src/govuk/all.mjs
@@ -43,7 +43,7 @@ function initAll (config) {
 
   var $characterCounts = $scope.querySelectorAll('[data-module="govuk-character-count"]')
   nodeListForEach($characterCounts, function ($characterCount) {
-    new CharacterCount($characterCount).init()
+    new CharacterCount($characterCount, config.characterCount).init()
   })
 
   var $checkboxes = $scope.querySelectorAll('[data-module="govuk-checkboxes"]')
--- a/src/govuk/components/character-count/character-count.mjs
+++ b/src/govuk/components/character-count/character-count.mjs
@@ -318,33 +318,4 @@ CharacterCount.prototype.isOverThreshold = function () {
   return (thresholdValue <= currentLength)
 }
 
-/**
- * Get dataset
- *
- * Get all of the data-* attributes from a given $element as map of key-value
- * pairs, with the data- prefix removed from the keys.
- *
- * This is a bit like HTMLElement.dataset, but it does not convert the keys to
- * camel case (and it works in browsers that do not support HTMLElement.dataset)
- *
- * @todo Replace with HTMLElement.dataset
- *
- * @param {HTMLElement} $element - The element to read data attributes from
- * @returns {Object} Object of key-value pairs representing the data attributes
- */
-CharacterCount.prototype.getDataset = function ($element) {
-  var dataset = {}
-  var attributes = $element.attributes
-  if (attributes) {
-    for (var i = 0; i < attributes.length; i++) {
-      var attribute = attributes[i]
-      var match = attribute.name.match(/^data-(.+)/)
-      if (match) {
-        dataset[match[1]] = attribute.value
-      }
-    }
-  }
-  return dataset
-}
-
 export default CharacterCount
--- a/src/govuk/components/character-count/character-count.mjs
+++ b/src/govuk/components/character-count/character-count.mjs
@@ -49,14 +49,14 @@ CharacterCount.prototype.init = function () {
   // Hide the fallback limit message
   $fallbackLimitMessage.classList.add('govuk-visually-hidden')
 
-  // Read options set using dataset ('data-' values)
-  this.options = this.getDataset($module)
+  // Read config set using dataset ('data-' values)
+  this.config = this.getDataset($module)
 
   // Determine the limit attribute (characters or words)
-  if (this.options.maxwords) {
-    this.maxLength = this.options.maxwords
-  } else if (this.options.maxlength) {
-    this.maxLength = this.options.maxlength
+  if (this.config.maxwords) {
+    this.maxLength = this.config.maxwords
+  } else if (this.config.maxlength) {
+    this.maxLength = this.config.maxlength
   } else {
     return
   }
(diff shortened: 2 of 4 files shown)


src/govuk/components/character-count/character-count.mjs
```diff
@@ -1,8 +1,66 @@
...
+ * @class
+ * @param {HTMLElement} $module - The element this component controls
+ * @param {Object} config
+ * @param {Number} config.maxlength - If `maxwords` is set, this is not required.
+ * The maximum number of characters. If `maxwords` is provided, it will be ignored.
+ * @param {Number} config.maxwords - If `maxlength` is set, this is not required.
+ * The maximum number of words. If `maxwords` is provided, `maxlength` will be ignored.
+ * @param {Number} [config.threshold=0] - The percentage value of the limit at
+ * which point the count message is displayed. If this attribute is set, the
+ * count message will be hidden by default.
```
36degrees, line 21: Good idea to re-use the descriptions from the Nunjucks options 👍🏻

src/govuk/components/character-count/character-count.mjs
```diff
@@ -1,8 +1,66 @@
...
+  }
+
+  // Read config set using dataset ('data-' values)
+  var datasetConfig = normaliseDataset($module.dataset)
+
+  // To ensure data-attributes take complete precedence,
+  // even if they change the type of count, we need to reset
+  // the `maxlength` and `maxwords` from the JavaScript config
+  // We can't mutate `config`, though, as it may be shared across
+  // multiple components inside `initAll`
```
36degrees, line 39: Would be good to wrap at 80 characters:

```suggestion
  // To ensure data-attributes take complete precedence, even if they change the
  // type of count, we need to reset the `maxlength` and `maxwords` from the
  // JavaScript config.
  //
  // We can't mutate `config`, though, as it may be shared across multiple
  // components inside `initAll`.
```

src/govuk/components/character-count/character-count.mjs
```diff
@@ -229,13 +273,13 @@ CharacterCount.prototype.count = function (text) {
...
   var $textarea = this.$textarea
-  var options = this.options
+  var config = this.config
   var remainingNumber = this.maxLength - this.count($textarea.value)
 
   var charVerb = 'remaining'
   var charNoun = 'character'
   var displayNumber = remainingNumber
-  if (options.maxwords) {
+  if (config.maxwords) {
```
36degrees, line 282: As discussed on Slack, as the local `config` variable is only used once we can probably avoid it:

```suggestion
  if (this.config.maxwords) {
```

(Could do the same for `this.textarea`?)

src/govuk/components/character-count/character-count.mjs
```diff
@@ -253,51 +297,25 @@ CharacterCount.prototype.getCountMessage = function () {
  * If there is no configured threshold, it is set to 0 and this function will
  * always return true.
  *
- * @returns {Boolean} true if the current count is over the options.threshold
+ * @returns {Boolean} true if the current count is over the config.threshold
  *   (or no threshold is set)
  */
 CharacterCount.prototype.isOverThreshold = function () {
+  // No threshold means we're always above threshold
+  // so we can save some computation
```
36degrees, line 305: I like this optimisation – makes sense 👍🏻 

Suggest wrapping at 80 characters and rewording slightly to fit on one line

```suggestion
  // No threshold means we're always above threshold so save some computation
```

src/govuk/components/character-count/character-count.test.js
```diff
@@ -321,6 +322,207 @@ describe('Character count', () => {
...
       })
     })
+
+    describe('JavaScript configuration', () => {
+      let examples
+      beforeAll(() => {
+        examples = getExamples('character-count')
+      })
+
+      describe('at instanciation', () => {
```
36degrees, line 332: Typo (I struggle with this word too!)

```suggestion
      describe('at instantiation', () => {
```

src/govuk/components/character-count/character-count.mjs
```diff
@@ -253,51 +297,25 @@ CharacterCount.prototype.getCountMessage = function () {
...
-  // Set threshold if presented in options
-  var thresholdPercent = options.threshold ? options.threshold : 0
-  var thresholdValue = maxLength * thresholdPercent / 100
+  var thresholdValue = maxLength * this.config.threshold / 100
 
   return (thresholdValue <= currentLength)
 }
 
-/**
- * Get dataset
```
36degrees, line 275: 🗑️ 🎉

src/govuk/all.mjs
```diff
@@ -43,7 +43,7 @@ function initAll (config) {
 
   var $characterCounts = $scope.querySelectorAll('[data-module="govuk-character-count"]')
   nodeListForEach($characterCounts, function ($characterCount) {
-    new CharacterCount($characterCount).init()
+    new CharacterCount($characterCount, config.characterCount).init()
```
36degrees, line 46: This needs adding to the JSDoc for the `initAll` function

src/govuk/components/character-count/character-count.mjs
```diff
@@ -1,8 +1,66 @@
 import '../../vendor/polyfills/Function/prototype/bind.mjs'
 import '../../vendor/polyfills/Event.mjs' // addEventListener and event.target normalisation
 import '../../vendor/polyfills/Element/prototype/classList.mjs'
+import { mergeConfigs, normaliseDataset } from '../../common.mjs'
+
+/**
+ * JavaScript enhancements for the CharacterCount component
+ *
+ * Tracks the number of characters or words in the `.govuk-js-character-count`
+ * `<textarea>` inside the element. Displays a message when the
```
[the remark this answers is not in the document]
romaricpascal: Whoops 🤭

src/govuk/components/character-count/character-count.mjs
romaricpascal: This gets completely revamped in the I18n PR so I'll leave it as is in this one: https://github.com/alphagov/govuk-frontend/pull/2887/files#diff-604c43935cbbf4edb8c1209bc9b06ae7341fdaf64e208f777db723fd456bf9baR294

src/govuk/all.mjs
romaricpascal: Good catch, completely missed it 😬

36degrees (approved):
🎉

PR #2887, opened 2022-09-28
Allow CharacterCount component to receive i18n config via JS

### LINKED ISSUE #2568: Add thousands comma-seperation in character count messages

### LINKED ISSUE #2701: Add ability to pass translation strings into character count component via data-attributes
## What

## Why

## Who needs to work on this
Developers

## Who needs to review this
Developers

## Done when

### LINKED ISSUE #2805: Add ability to pass translation strings into character count component via JS
## What
Add ability for users to pass a translation object in via JavaScript (at the point of initialisation). This is an alternative method to passing translation strings via the Nunjucks macros (data-attributes).

Note: this may already have been covered by all the previous internationalisation tickets. Even if we don’t need to write additional code to get this working, we should ensure it’s tested.

## Why
Some users may prefer to pass translation strings via JavaScript, or it may be better suited to their service.

## Who needs to work on this
Developers

## Who needs to review this
Developers

## Done when

- [ ] It’s possible to provide translations to the character count component via JavaScript
- [ ] The character count component has tests which cover the ability to translate via JavaScript
- [ ] The character count component still works if no/only some translations are passed (falls back to defaults)
- [ ] Data attribute translations take priority over JS translations, if both provided

romaricpascal:
Builds upon the [configuration work for the CharacterCount component](https://github.com/alphagov/govuk-frontend/pull/2883) to allow translating the message displaying the count.

To avoid adding another flurry of Puppeteer tests, the PR starts with a little refactoring to isolate the formatting of the message so it can be unit tested. The Puppeteer tests remain in place to keep testing that the right thing appears based on user interactions.

Closes #2805, and closes #2568 as well, and closes #2701 also.


src/govuk/components/character-count/character-count.mjs
```diff
@@ -26,7 +27,21 @@ function CharacterCount ($module, config) {
   }
 
   var defaultConfig = {
-    threshold: 0
+    threshold: 0,
+    i18n: {
+      // Characters
+      charactersUnderLimitZero: 'You have %{count} characters remaining',
```
romaricpascal, line 33: Thought of renaming that one `charactersAtLimit` rather than piggyback on the `charactersUnderLimit`. This would have had impact on #2701 if it got started in parallel so stuck with the naming from the spike. Happy to add an extra commit to separate things.

src/govuk/components/character-count/character-count.mjs
```diff
@@ -52,6 +67,8 @@ function CharacterCount ($module, config) {
     datasetConfig
   )
 
+  this.i18n = new I18n(extractConfigByNamespace(this.config, 'i18n'), { locale: this.config.i18nLocale })
```
romaricpascal, line 70: Uses the `i18nLocale` key and not `i18n.locale` key to keep translations and the locale separated.

romaricpascal: Do we want to default the locale to `(this.$module.closest('[lang]') || {}).lang` (in a less one-liner fashion maybe)? It looks like we have a polyfill for `closest` in the codebase already. 

I'll probably update the accordion for consistency if that's the case.

colinrotherham: Feels like `.closest('[lang]')` would work

I'd worry about problematic nested multi-language content

For example a Welsh `cy-GB` page containing a nested "**This component is also available in English**" section with a `{{ govukCharacterCount() }}` count inside. Could it grab the wrong `[lang]`?

romaricpascal: > For example a Welsh cy-GB page containing a nested "This component is also available in English" section with a {{ govukCharacterCount() }} count inside. Could it grab the wrong [lang]?

Is that the structure that sentence describes? 

```html
<html lang="cy-GB">
   <!-- ... -->
   <body>
       <!-- Some preceeding content in Welsh -->
	   <!-- The announcement in English -->
       <p lang="en-GB">This component is also available in English</p>
       <div class="govuk-character-count" data-module="character-count" data-maxlength="200">
          <!-- The problematic character count component that would pull the Welsh locale instead of the English one -->
       </div>
    <!-- ... -->
```

Automatically pulling the locale from the closest `lang` attribute would indeed add a risk of showing misformated numbers in such scenario (we expect the component to [receive messages in the right locale](https://github.com/alphagov/govuk-frontend/blob/character-count-i18n/src/govuk/components/character-count/character-count.mjs#L33-L43), there's no picking messages based on the locale... – for now? ). 

The flip side of that is that it does remind people to use the `lang` attribute and make things accessible. And actually, having it impact the rendered content also gives a chance to catch that a `lang` is missing at QA (arguably, not so much for a Welsh/English combination of language). Especially if we imagine in the future the I18n feature evolving to maybe pick the messages based on the locale, either by having multiple locales provided to the component itself or in a centralised way.

Actually, this makes me wonder if we do even need that `data-i18nLocale` attribute and shouldn't rely on the closest `lang` at all times 🤔

36degrees: I think as long as the page is marked up correctly, finding the closest parent lang should match [the behaviour in the HTML spec](https://html.spec.whatwg.org/#the-lang-and-xml:lang-attributes):

> If these attributes are omitted from an element, then the language of this element is the same as the language of its parent element, if any.

We should allow for a lang attribute to be set on the `$module` itself – not sure if `HTMLElement.closest` can return the starting HTML element or not?

src/govuk/components/character-count/character-count.mjs
36degrees: `charactersAtLimit` makes sense to me 👍🏻 

Do we need to change anything about the implementation as that is effectively a separate translation that isn't pluralised?

src/govuk/components/character-count/character-count.mjs
36degrees: > We should allow for a lang attribute to be set on the $module itself – not sure if HTMLElement.closest can return the starting HTML element or not?

Incredibly lazy of me not to check this as it literally took 2 seconds – [`HTMLElement.closest` can return itself](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest#return_value).

Probably worth adding an explicit test for that case though.

romaricpascal: No worries, I would have added two test cases anyway and we'd have had the answer then 😆 (I'm always confused about `closest` as well).

Should we completely drop that `data-i18nLocale` attribute. Is there ever a scenario where we'd want that to be different from `lang`?

36degrees: > I'd be keen to drop that `data-i18nLocale` attribute, then. Is there ever a scenario where we'd want that to be different from `lang`?

Based on [this comment](https://github.com/alphagov/govuk-frontend/issues/2804#issuecomment-1246451644) from @querkmachine I think we want to keep it simple and just use the `lang` attribute.

We can always introduce a way to override it in the config later if we do find there's a valid need for it.

romaricpascal: Argh, I'd forgotten about this comment, thanks for the archeology. That's all neat then, I'll update to use `lang` and and update the tests accordingly.

src/govuk/components/character-count/character-count.mjs
romaricpascal: Yup, it requires a little update to `formatCountMessage` in the lines of:

```js
if (remainingNumber === 0) {
    var translationKey = type + 'AtLimit'
	return I18n.t(translationKey)
}

// Do things as it does now otherwise
```

So nothing too wild 😄

src/govuk/components/character-count/character-count.mjs
36degrees: > Is that the structure that sentence describes?

I think you must have added this example at the same time I was writing my reply as I've only just seen it.

I'd argue for that example the page is marked up incorrectly – if the character count is intended to be in English then it should be nested inside an element with an `lang="en-GB"` attribute.

src/govuk/components/character-count/character-count.mjs
romaricpascal: @claireashworth Because we're having a special option for when 0 words or characters are remaining now, we are no longer tied to the "You have <number> words/characters remaining" sentence structure when users hit the limit. 

Do you think it's worth changing the wording for that case, for ex to using only words "You have no words/characters remaining"? Thought behind that, besides it helping me with testing (😆), it that it'd also make that special situation stand out from when you're not quite there/past the limit because there'd be no longer a number in the sentence.

src/govuk/components/character-count/character-count.mjs
romaricpascal: > I'd argue for that example the page is marked up incorrectly – if the character count is intended to be in English then it should be nested inside an element with an `lang="en-GB"` attribute.

Agreed

> > I'd be keen to drop that `data-i18nLocale` attribute, then. Is there ever a scenario where we'd want that to be different from `lang`?
> 
> Based on [this comment](https://github.com/alphagov/govuk-frontend/issues/2804#issuecomment-1246451644) from @querkmachine I think we want to keep it simple and just use the `lang` attribute.
> 
> We can always introduce a way to override it in the config later if we do find there's a valid need for it.

As I started working on moving to just `lang`, realised that we do need that `i18nLocale` config option... for when we're passing config through JavaScript as there's no `lang` in that world. So I'll keep it and just default to the closest `lang` attribute when missing.

src/govuk/components/character-count/character-count.mjs
claireashworth: 'You have 0 words/characters remaining' keeps the sentence structure consistent.
I don't think 'You have no words/characters remaining' makes it stand out more. A 'more correct' way of phrasing it would be 'You do not have any words/characters remaining', but that is quite long to put under a text box.
Maybe swapping to 'zero' instead of '0' would work as it could not be mistaken for 'o' and is more characters which makes it more visible?

romaricpascal: Fair point on the 'standing out' not being massive with the 'no', that was a bit of a tenuous argument on my part 😬 . 'zero' is a much better option for that indeed, thanks for the idea 😄

Do we have any guidance on whether we'd prefer consistency (I can always turn my tests a different way to accomodate the sentences being the same) over disambiguating the '0'/switching to fully written number in that case (which helps my testing, but I don't want it to be at the detriment of users)?

lib/dom-helpers.mjs
```diff
@@ -0,0 +1,19 @@
+/**
+ * Creates an element with the given attributes
+ *
+ * TODO: Extend to allow passing props and children if necessary
+ * TODO: Make available to components once support for older browsers is dropped (add tests)
```
36degrees, line 5: Do we know for sure these are things we want to do, or are these more like ideas for the future? I'd be tempted to leave them out unless we know for sure.

src/govuk/common.mjs
```diff
@@ -185,3 +186,16 @@ export function normaliseDataset (dataset) {
 
   return out
 }
+
+/**
+ * Returns the value of the `lang` attribute closest to the given element (including itself)
```
36degrees, line 191: This needs to be generalised as this component can be used to get any attribute, not just `lang`.

src/govuk/common.unit.test.mjs
```diff
@@ -216,4 +217,32 @@ describe('Common JS utilities', () => {
...
+    })
+
+    it('returns the value of the closest parent with the attribute if it exists', () => {
+      const $greatGrandParent = createElement('div', { lang: 'cy-GB' }) // To check that we take the value up
+      const $grandparent = createElement('div', { lang: 'en-GB' })
+      const $parent = createElement('div') // To check that we walk up the tree
+      const $element = createElement('div')
+      $greatGrandParent.appendChild($grandparent)
+      $grandparent.appendChild($parent)
+      $parent.appendChild($element)
```
36degrees, line 235: I like the `createElement` helper function, but I do wonder if this would be easier to follow if this was written out in HTML rather than constructed by appending individual children. The naming does help, but it's a _little_ hard to follow.

lib/dom-helpers.mjs
```diff
@@ -0,0 +1,19 @@
+/**
+ * Creates an element with the given attributes
```
36degrees, line 2: I'm not 100% sure that we need this abstraction, and it could end up trying to do too much if we extend it to do everything we might want to do with a created element as the TODOs are hinting at.

Are we sure we want to go down this route rather than just using the native APIs directly in the tests?

src/govuk/common.mjs
```diff
@@ -185,3 +186,16 @@ export function normaliseDataset (dataset) {
...
+ *
+ * @param {HTMLElement} $element - The element to start walking the DOM tree up
+ * @returns {String|undefined}
+ */
+export function closestAttributeValue ($element, attributeName) {
```
36degrees, line 196: `attributeName` is missing from the JSDoc.

src/govuk/common.unit.test.mjs
```diff
@@ -216,4 +217,32 @@ describe('Common JS utilities', () => {
       })
     })
   })
+
+  describe('closestAttributeValue', () => {
```
36degrees, line 221: I like how you've split this logic out so we can unit test it separately 👍🏻

src/govuk/components/character-count/character-count.mjs
```diff
@@ -30,7 +31,21 @@ function CharacterCount ($module, config) {
...
+      charactersAtLimit: 'You have no characters remaining',
+      charactersOverLimitOne: 'You have %{count} character too many',
+      charactersOverLimitOther: 'You have %{count} characters too many',
+      // Words
+      wordsUnderLimitOne: 'You have %{count} word remaining',
+      wordsUnderLimitOther: 'You have %{count} words remaining',
+      wordsAtLimit: 'You have no words remaining',
+      wordsOverLimitOne: 'You have %{count} word too many',
+      wordsOverLimitOther: 'You have %{count} words too many'
+    }
```
36degrees, line 48: All of these now need to be in the JSDoc! 😱

src/govuk/components/character-count/character-count.mjs
```diff
@@ -57,6 +72,12 @@ function CharacterCount ($module, config) {
     datasetConfig
   )
 
+  console.log(this.$module)
```
36degrees, line 75:
```suggestion

```

src/govuk/components/character-count/character-count.mjs
```diff
@@ -57,6 +72,12 @@ function CharacterCount ($module, config) {
...
+  this.i18n = new I18n(extractConfigByNamespace(this.config, 'i18n'), {
+    // Read the fallback if necessary rather than have it set in the defaults
+    locale: this.config.i18nLocale || closestAttributeValue($module, 'lang')
```
36degrees, line 78: Picking up from the discussion in https://github.com/alphagov/govuk-frontend/pull/2887#discussion_r983423951, are we sure we need the `i18nLocale` config option? I think it's OK to always rely on the closest `lang` attribute – not 100% sure what the need is to be able to define it in JS.

src/govuk/components/character-count/character-count.mjs
```diff
@@ -277,22 +298,20 @@ CharacterCount.prototype.count = function (text) {
...
+  var countType = this.config.maxwords ? 'words' : 'characters'
+  return this.formatCountMessage(remainingNumber, countType)
+}
 
-  var charVerb = 'remaining'
-  var charNoun = 'character'
-  var displayNumber = remainingNumber
-  if (config.maxwords) {
-    charNoun = 'word'
+CharacterCount.prototype.formatCountMessage = function (remainingNumber, countType) {
```
36degrees, line 307: Can we add JSDoc for this method?

src/govuk/components/character-count/character-count.unit.test.mjs
```diff
@@ -0,0 +1,148 @@
...
+import { createElement } from '../../../../lib/dom-helpers.mjs'
+import CharacterCount from './character-count.mjs'
+
+describe('CharacterCount', () => {
+  describe('formatCountMessage', () => {
+    describe('default configuration', () => {
+      let component
+      beforeAll(() => {
+        // The component won't initialise if we don't pass it an element
+        component = new CharacterCount(createElement('div'))
```
36degrees, line 14: Slightly nervous about passing a div here, and imagine this could come back to haunt us in the future when e.g. we get rid of the `init` function (which at the minute we're not calling, so a lot of the 'setup' for the component isn't happening).

Don't have any immediate suggestions for a fix, and this might just be something we have to revisit later… but it feels a bit fragile.

src/govuk/components/character-count/character-count.unit.test.mjs
```diff
@@ -0,0 +1,148 @@
...
+        expect(component.formatCountMessage(-1, 'words')).toEqual('You have 1 word too many')
+      })
+      it('formats plural exceeding words', () => {
+        expect(component.formatCountMessage(-10, 'words')).toEqual('You have 10 words too many')
+      })
+      it('formats word limit being met', () => {
+        expect(component.formatCountMessage(0, 'words')).toEqual('You have no words remaining')
+      })
+
+      it('formats the count', () => {
```
36degrees, line 49: We're using the word format in the test descriptions to mean a few different things – can we be clearer about what it is we're testing?

src/govuk/components/character-count/character-count.unit.test.mjs
```diff
@@ -0,0 +1,148 @@
...
+describe('CharacterCount', () => {
+  describe('formatCountMessage', () => {
+    describe('default configuration', () => {
+      let component
+      beforeAll(() => {
+        // The component won't initialise if we don't pass it an element
+        component = new CharacterCount(createElement('div'))
+      })
+
+      it('formats singular remaining characters', () => {
```
36degrees, line 17: This might be [a good use case for `it.each`](https://elfi-y.medium.com/reduce-boilerplate-test-code-with-jest-it-each-30a0eec9776d) as it's quite repetitive ('given this number of remaining characters, the message should be this')

lib/dom-helpers.mjs
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * TODO: Extend to allow passing props and children if necessary
+ * TODO: Make available to components once support for older browsers is dropped (add tests)
+ *
+ * @param {String} tagName
+ * @param {Object} attributes
+ * @returns HTMLElement
+ */
+export function createElement (tagName, attributes = {}) {
+  const el = document.createElement(tagName)
```
36degrees, line 12: As flagged in Slack, although `document` seems to be part of the default globals for Standard, we might want to explicitly configure eslint so that it knows this is file is in the context of a browser / jsdom (for example by adding another override for this file in `.eslintrc.js`)

lib/dom-helpers.mjs
romaricpascal: Probably more ideas at that stage, but I think there's ground for having a helper function for creating DOM elements with specific attributes and children quickly. That would make some of the code of the Accordion and CharacterCount less verbose I believe.

I'll drop the comments and move the discussion about this into an issue to avoid confusion.

src/govuk/common.mjs
romaricpascal: Whoops, started with it only supporting `lang` and forgot to update 😬

src/govuk/common.unit.test.mjs
romaricpascal: Fair point, I'll see if I can rewrite this to work from an HTML snippet.

src/govuk/components/character-count/character-count.mjs
claireashworth: Nothing specific in guidance about consistency. It's generally better to be consistent with routine things and seeing a '0' would probably be accepted by the users as logical (I obviously can't say that with certainty as I haven't researched it). If using 'zero' for testing purposes helps, then in my opinion it shouldn't confuse users.

The only other point to note is that the government style guide specifically says "Use a 0 where there’s no digit before the decimal point." So if we stick to the government's chosen style we have to use '0'. If we use 'zero' we should be able to demonstrate that it's a better choice in these circumstances.

src/govuk/components/character-count/character-count.mjs
romaricpascal: Added, but I may want to dig a bit further into it as it makes massive lines 😬 @colinrotherham would you have any suggestions on how to make it a bit more manageable?

lib/dom-helpers.mjs
romaricpascal: Issue created for discussion there: https://github.com/alphagov/govuk-frontend/issues/2894

src/govuk/components/character-count/character-count.mjs
romaricpascal: I'll remove the `i18nLocale` config option then. It was more there for parity than anything.

src/govuk/components/character-count/character-count.mjs
```diff
@@ -23,14 +24,49 @@ import { mergeConfigs, normaliseDataset } from '../../common.mjs'
...
   }
 
   var defaultConfig = {
-    threshold: 0
+    threshold: 0,
+    i18n: {
+      // Characters
+      charactersUnderLimitOne: 'You have %{count} character remaining',
+      charactersUnderLimitOther: 'You have %{count} characters remaining',
+      charactersAtLimit: 'You have no characters remaining',
```
querkmachine, line 60: Just a note that the two 'AtLimit' strings here are different to what these components currently return ("You have 0 characters remaining" and "You have 0 words remaining"). Not sure if that needs to be flagged as a breaking change?

romaricpascal: I was discussing the spelling with Claire, and I think I'll revert to the 0 and find a different way to test.

From what I understood of what makes a breaking change or not (which I may well still be blurry about), this wouldn't qualify as users wouldn't have to change their code when upgrading, it's "only" the HTML we output that would change.

querkmachine: That makes sense to me. 

Does make me think more generally though: in future, if we change what a string says in English, we may want to flag it in the release notes in case it warrants a change to translations too.

src/govuk/components/character-count/character-count.unit.test.mjs
romaricpascal: Because we have the Puppeteer tests, I'm not too worried about mocking what the component runs onto and skipping its initialisation. 

If anything, having these tests only depend on the bare minimum that the feature needs (one element and no `init` call) can help us ensure we do update tests: if the formatting starts to depend on more than what's mocked, these tests will fail and force us to update them or to better isolate that method, and maybe prompt us to add new cases around the change that we may have missed through the end-to-end tests in Puppeteer (because we can't cover all combinations of everything there, while we can be more thorough in unit tests which run with less overhead). 

Is it the extra updates that worry you?

src/govuk/components/character-count/character-count.mjs
romaricpascal: That's a very good point for future changelogs indeed.

src/govuk/components/character-count/character-count.mjs
romaricpascal: I went and reverted to `0` in number in the end, keeping thing consistent with not only the rest of the messages but the current behaviour as well. Thanks for all the clarifications 😄

romaricpascal:
@36degrees Made the updates based on your last feedback. Happy to discuss test and linting concerns further but I feel they're two things that could be fixed in future PRs rather than completely blocking that one from moving forward?

colinrotherham:
> @36degrees Made the updates based on your last feedback. Happy to discuss test and linting concerns further but I feel they're two things that could be fixed in future PRs rather than completely blocking that one from moving forward?

@romaricpascal I've found it helps to ask if anything has changed "for the worse" in the PR

If it's something we agree we can live with temporarily, fine. Otherwise if the changes have carried on using existing conventions an techniques, code style etc, improvements can come later

romaricpascal:
> @romaricpascal I've found it helps to ask if anything has changed "for the worse" in the PR

Oh, that's a good one, I'll keep it in mind. If things are moving "for the better", we can always move them further "for the better" in future PRs indeed.

> If it's something we agree we can live with temporarily, fine. Otherwise if the changes have carried on using existing conventions an techniques, code style etc, improvements can come later

I think that words what I'm trying to get: a confirmation that we'd be fine shipping as is and revisiting in the future if we want to amend tests/linting

> Looking at the **dom-helpers.js** module (similar to [**puppeteer-helpers.js** that I added](https://github.com/alphagov/govuk-frontend/commit/7e4e0e20564012811072f0215db2f3f8e1b56795)) perhaps we should extend the Jest environments instead, not yet but soon?

I have a slightly different direction in mind for these helpers, where we could also use them to create elements in the components themselves. Unlike the querying, the DOM API is very verbose when it comes to creating elements with attributes and children and I think we'd gain some readability with a little abstraction there. I've opened an issue to discuss there: https://github.com/alphagov/govuk-frontend/issues/2894. 

> Otherwise could it grow into a "mini jQuery"?

Yeah, that's something we'd need to keep an eye on if we move these helpers further.

> We already use **cheerio** for this elsewhere. Faster to use a library that to write our own utilities.

Cheerio didn't quite fit the bill there as we needed to get an instance of `HTMLElement` to run the component onto without spinning up an actual browser. I didn't see any API in Cheerio to get something like that out. It seems to only have those jQuery-esque objects or text outputs.

> ### Jest environments
> 
> Since Jest doesn't implement `require.cache` you'll notice that any files we require/import in tests (versus environments) are executed _hundreds_ of times—this does start to add up when you're looping directory listings 😮
> 
>     1. **Environment `@jest-environment jsdom`**
>        Could we extend this to always `require()` **cheerio** `global.$` so **dom-helpers.js** isn't needed?
> 
>     2. **Environment `@jest-environment puppeteer`**
>        Could we extend this to always `require()` the various helpers like `renderAndInitialise`, `waitForHiddenSelector`, `waitForVisibleSelector` so they're also available in tests automatically?

Wasn't aware of the lack of require caching in Jest, cheers for pointing it out 😄 I'm wondering how much of a bottleneck it is vs how much we'd lose of which files rely on which helpers that `import` bring. I'd personally err on the cautious side of not putting too many globals if we don't have perfomance issues, but it may be an improvement worth tracking in an issue 😄

colinrotherham:
Ahh fair enough, I thought Cheerio had jQuery's `${'<p>hello</p>'}` quick creation

Should **dom-helpers.js** be moved inside `./src/govuk/`?

romaricpascal:
> Ahh fair enough, I thought Cheerio had jQuery's `${'<p>hello</p>'}` quick creation

It does, but the returned object doesn't have an equivalent of jQuery's `.el()` to return the corresponding HTML element from what I gathered of their docs. It's text -> jQuery-like object -> text/html/xml. They highlight in their docs that they are ["not a DOM emulation project like JSDom"](https://cheerio.js.org/#cheerio-is-not-a-web-browser)

> Should **dom-helpers.js** be moved inside `./src/govuk/`?

Not yet, it is deliberately in `lib` to highlight that it's tooling code rather than frontend code. It's also using syntax and element (template) not compatible with the browsers we supports 😄

colinrotherham:
See what @36degrees thinks?

We should do the hard work (+ tests) via a nice neat DOM helpers file, but only if it was going in `./src/govuk/`?

Until that happens, if it's just for some some quick ad-hoc test utilities we can use shorthand? Need to be aware of the subtle differences between properties/attributes, but these do the job:

**Adding `[lang]` to `<div>`**
```mjs
const $localised = Object.assign(document.createElement('div'), {
  lang: 'de'
})
```

**Adding content to `<div>`**
```mjs
const $content = Object.assign(document.createElement('div'), {
  innerHtml: 'Hello World'
})
```

**Finding closest `[lang]` attribute value**
```mjs
$closest = $element.closest('[lang]')?.getAttribute('lang')
```

colinrotherham:
**Another example for datasets**
```mjs
const $data = document.createElement('div')
Object.assign($data.dataset, { something1: 'abc', something2: 'def' })
```

It's not a blocker for me, but addresses @36degrees's [comment about a new abstraction](https://github.com/alphagov/govuk-frontend/pull/2887#discussion_r983699156) for tests

36degrees:
> We should do the hard work (+ tests) via a nice neat DOM helpers file, but only if it was going in `./src/govuk/`?

I'm not 100% sure why this is conditional on being in `src/govuk/`.

At the minute the src folder contains only the JS we want to include in the package and the tests for those JS files –  any 'supporting' functions that are used in tests live in the `lib` folder.

When we build the package, we then exclude the tests.

colinrotherham:
> I'm not 100% sure why this is conditional on being in `src/govuk/`.

Sorry @36degrees I had a chat about this with @romaricpascal so it's due an explanation

It was mentioned by that **dom-helpers.js** might [eventually be used in the components](https://github.com/alphagov/govuk-frontend/pull/2887#issuecomment-1268518911):

>I have a slightly different direction in mind for these helpers, where we could also use them to create elements in the components themselves. Unlike the querying, the DOM API is very verbose when it comes to creating elements with attributes and children and I think we'd gain some readability with a little abstraction there. I've opened an issue to discuss there: https://github.com/alphagov/govuk-frontend/issues/2894.

These points came up:

1. Are they for tests only?
2. Are they in the right location?
2. Can we use a 3rd party library instead?
3. Do we even need them?

In terms of location, if they're "battle-hardened" helpers intended for production, like `nodeListForEach`, `normaliseString` and now `normaliseDataset`, then let's keep it in the appropriate place (+ tests)

But because the **dom-helpers.js** file (right now) is only for tests, I'd [agree with you](https://github.com/alphagov/govuk-frontend/pull/2887#discussion_r983699156) that we might end up doing too much.

Those quick ad-hoc `document.createElement()` examples might be good enough?

36degrees:
Gotcha, that makes sense – I had missed some of that context.

My preference would be to do the minimum required to ship this and consider the other aspects as part of the upcoming architectural changes.

romaricpascal:
@36degrees @colinrotherham I've removed the abstractions, awaiting a longer discussion about helpers for creating markup in the issue created on that topic: https://github.com/alphagov/govuk-frontend/issues/2894. I'm fixing the failing tests from the move to named exports from `I18n` and I think it'll be all good after that, unless there's something I missed? 😄

src/govuk/components/character-count/character-count.mjs
```diff
@@ -1,7 +1,8 @@
 import '../../vendor/polyfills/Function/prototype/bind.mjs'
 import '../../vendor/polyfills/Event.mjs' // addEventListener and event.target normalisation
 import '../../vendor/polyfills/Element/prototype/classList.mjs'
```
colinrotherham, line 3: Do we put the polyfill here?

```suggestion
import '../../vendor/polyfills/Element/prototype/classList.mjs'
import '../../vendor/polyfills/Element/prototype/closest.mjs'
```

src/govuk/common.mjs
```diff
@@ -1,4 +1,5 @@
 import './vendor/polyfills/Element/prototype/dataset.mjs'
+import './vendor/polyfills/Element/prototype/closest.mjs'
```
colinrotherham, line 2: Does the polyfill definitely go here?

```suggestion
```

src/govuk/components/character-count/character-count.mjs
```diff
@@ -23,14 +24,49 @@ import { mergeConfigs, normaliseDataset } from '../../common.mjs'
...
+ *   Message notifying users they're any number of characters over the limit
+ * @param {String} [config.i18n.wordsUnderLimitOne="You have %{count} word remaining"]
+ *   Message notifying users they're 1 word under the limit
+ * @param {String} [config.i18n.wordsUnderLimitOther="You have %{count} words remaining"]
+ *   Message notifying users they're any number of words under the limit
+ * @param {String} [config.i18n.wordsAtLimit="You have no words remaining"]
+ *   Message notifying users they've reached the limit number of words
+ * @param {String} [config.i18n.wordsOverLimitOne="You have %{count} word too many"]
+ *   Message notifying users they're 1 word over the limit
+ * @param {String} [config.i18n.wordsOverLimitOther="You have %{count} words too many"]
```
colinrotherham, line 46: Thoughts on making large complicated JSDoc objects like **config.i18n** into a `@typedef`?

You can share them between files too, so uses of **config.i18n** match

Like the one we did for **ComponentData**?

```mjs
/**
 * Example
 *
 * @param {object} example
 * @param {import('../../lib/file-helper').ComponentData} example.componentData
 */
```

```mjs
/**
 * Component data from YAML
 *
 * @typedef {object} ComponentData
 * @property {string} name - Component name
 * @property {unknown[]} [params] - Nunjucks macro options
 * @property {unknown[]} [examples] - Example Nunjucks macro options
 * @property {string} [previewLayout] - Nunjucks layout for component preview
 * @property {string} [accessibilityCriteria] - Accessibility criteria
 */
```

src/govuk/components/character-count/character-count.mjs
romaricpascal: If we move it here, we'd have to move the `closestAttributeValue` function as well as it's the one requiring the polyfill. We need to export it for testing, though and I think Rollup will complain about mixing named and not named exports, I think. It could be worth extracting the method and its polyfill import in its own file (`src/closestAttributeValue.mjs` for ex.) to avoid burdening individual imports of other components with the polyfill, that's a good point.

romaricpascal: Oh, I see, good point about polyfills being side effects (thanks for the detailed explanation, as always).

I'm not a huge fan of putting them in each components as a blanket rule, but more in "whatever requires the polyfill". That'll limit the risks of a component loading, say `closestAttributeValue` (or `normaliseString`) and forgetting to polyfill. With this approach we would:

- make a separate module for `closestAttributeValue` that imports `Element.prototype.closest`
- make a separate module for `normaliseString` that imports `String.prototype.trim`
- import `Element.prototype.dataset` in all components requiring configuration as they access the `dataset`. Given that one will be an easy one to miss, I wonder if we shouldn't have a `getDataset` function that has side effect.

These could live in a `polyfilled` folder to be clear about why those methods are on their own and we could revisit at regular interval (starting with 5.0) the need for them.

All in all, I think that's something for a separate PR, though, as it's not really linked to the feature. The feature's implementation just carries on from having `dataset` loaded in `common.mjs` 😊

colinrotherham: Yep I'm happy with it not being a blocker, and can address whether both new imports (since the last release) should be moved or not in another PR too

```mjs
import './vendor/polyfills/Element/prototype/dataset.mjs'
import './vendor/polyfills/Element/prototype/closest.mjs'
````

File: src/govuk/common.mjs (hunk 19 above)
Comment (colinrotherham) [reply src/govuk/common.mjs:2]: Happy to merge with this line, we're going to address it before the next release under https://github.com/alphagov/govuk-frontend/issues/2907

File: src/govuk/components/character-count/character-count.mjs (hunk 20 above)
Comment (colinrotherham) [reply src/govuk/components/character-count/character-count.mjs:46]: Just commenting to say we'll leave this (JSDoc `@typedef`) for now, but would be nice to see one coming soon

File: src/govuk/components/character-count/character-count.mjs [hunk 21]
```diff
@@ -23,14 +24,49 @@ import { mergeConfigs, normaliseDataset } from '../../common.mjs'
...
+ *   Message notifying users they've reached the limit number of characters
+ * @param {String} [config.i18n.charactersOverLimitOne="You have %{count} character too many"]
+ *   Message notifying users they're 1 character over the limit
+ * @param {String} [config.i18n.charactersOverLimitOther="You have %{count} characters too many"]
+ *   Message notifying users they're any number of characters over the limit
+ * @param {String} [config.i18n.wordsUnderLimitOne="You have %{count} word remaining"]
+ *   Message notifying users they're 1 word under the limit
+ * @param {String} [config.i18n.wordsUnderLimitOther="You have %{count} words remaining"]
+ *   Message notifying users they're any number of words under the limit
+ * @param {String} [config.i18n.wordsAtLimit="You have no words remaining"]
```
colinrotherham, line 42: This one correct too?

```suggestion
 * @param {String} [config.i18n.wordsAtLimit="You have 0 words remaining"]
```

src/govuk/components/character-count/character-count.mjs
```diff
@@ -277,22 +318,28 @@ CharacterCount.prototype.count = function (text) {
...
+    return this.i18n.t(countType + 'AtLimit')
   }
-  charNoun = charNoun + ((remainingNumber === -1 || remainingNumber === 1) ? '' : 's')
 
-  charVerb = (remainingNumber < 0) ? 'too many' : 'remaining'
-  displayNumber = Math.abs(remainingNumber)
+  var translationKeySuffix = remainingNumber < 0 ? 'OverLimit' : 'UnderLimit'
 
-  return 'You have ' + displayNumber + ' ' + charNoun + ' ' + charVerb
+  return this.i18n.t(countType + translationKeySuffix, { count: Math.abs(remainingNumber) })
```
colinrotherham, line 342: We do `Math.abs(Math.floor(count))` in **i18n.mjs**

Is this enough? Probably can't have decimals can we

src/govuk/components/character-count/character-count.unit.test.mjs
```diff
@@ -0,0 +1,118 @@
...
+
+        it('overrides the locale when set on an ancestor', () => {
+          const $parent = document.createElement('div')
+          $parent.setAttribute('lang', 'de')
+          const $div = document.createElement('div')
+          $parent.appendChild($div)
+
+          const component = new CharacterCount($div)
+
+          expect(component.formatCountMessage(10000, 'words')).toEqual('You have 10.000 words remaining')
```
colinrotherham, line 84: Super minor one, but any thoughts on logically grouping what the test is doing?

```mjs
const $parent = document.createElement('div')
const $div = document.createElement('div')

$parent.setAttribute('lang', 'de')
$parent.appendChild($div)

const component = new CharacterCount($div)
const countMessage = component.formatCountMessage(10000, 'words')

expect(countMessage).toEqual('You have 10.000 words remaining')
```

1. Create things
2. Customise things
3. Run some code
4. Run assertion

Helps with readability slightly, not a blocker though

colinrotherham (approved):
Some minor comments, but looks fab so approving

I'll leave it with you 😊 

**Update:** Oh and the conflict, quick rebase and we'll see if everything passes again?

src/govuk/components/character-count/character-count.mjs
romaricpascal: That's a good catch, but I'd feel safer leaving the `Math.abs` in the CharacterCount:
- it explicits that the CharacterCount wants to display the absolute value in the number, rather than obscuring it by knowing that the I18n does it (which not everyone reading that line will expect)
- it prevents surprises the day I18n expands its support as the line you mention is led by `// Currently our custom code can only handle positive integers`

src/govuk/components/character-count/character-count.mjs
romaricpascal: They've been set up there: 
- https://github.com/alphagov/govuk-frontend/pull/2895/files#diff-cde937e9880b3f4528adf0fe09930dfdb170736d3cb092d42173c11c00995caaR1-R15
- https://github.com/alphagov/govuk-frontend/pull/2895/files#diff-604c43935cbbf4edb8c1209bc9b06ae7341fdaf64e208f777db723fd456bf9baR8-R16

src/govuk/components/character-count/character-count.mjs
colinrotherham: That's fine, just something that jumped out at me 😊

Issue #2717, opened 2022-07-20
Improve and simplify Gulp


domoscargin:
## What
Simplify our use of Gulp by reducing its use to task running. Write our own scripts for the actual tasks, which do not rely on Gulp.

## Why
Our gulp tasks are complicated and doing too many things: #2243

Additionally, we cannot update a lot of dependencies because of our reliance on `oldie`: #2469 

We had previously [made the decision to move away from Gulp](https://github.com/alphagov/govuk-frontend/issues/2558) and write our own build scripts.

We essentially removed all Gulp usage, then changed our mind with some exceptions

### Composed build tasks can use Gulp

1. Build targets in `tasks/build/**` for **./dist** and **./package**
2. Review app builds in `app/tasks/build/**` for **./app/dist**

But since every Gulp task such as `npx gulp build:package` has a corresponding `npm run build:package` we chose not to document these in [tasks.md](https://github.com/alphagov/govuk-frontend/blob/main/docs/contributing/tasks.md#gulp-tasks)

### Composed build tasks need logging

But then we realised we lost logging in the terminal so brought some Gulp usage back:

* Limiting `gulp.parallel()` and `gulp.series()` for composed build tasks only
* Limiting `gulp.task()` usage to **gulpfile.mjs** configs only

### Shared tasks without Gulp

But last of all, we split out all the Review app tasks in https://github.com/alphagov/govuk-frontend/pull/3384 which gave us the opportunity to make all shared tasks work without Gulp. We switched to the [export your tasks](https://gulpjs.com/docs/en/getting-started/creating-tasks) pattern vs [using `gulp.task()`](https://gulpjs.com/docs/en/api/task/)

>**Reminder:** This API isn't the recommended pattern anymore - [export your tasks](https://gulpjs.com/docs/en/getting-started/creating-tasks).

All our new tasks are now plain async functions:

```mjs
await files.copy('**/*', {
  srcPath: join(paths.src, 'govuk/assets'),
  destPath: join(paths.app, 'dist/assets')
})
```

Yet they can still be composed into tasks for Gulp's `gulp.series()` and `gulp.parallel()` etc

```mjs
task.name('copy:assets', () => files.copy('**/*', {
  srcPath: join(paths.src, 'govuk/assets'),
  destPath: join(paths.app, 'dist/assets')
}))
```

## Who needs to work on this
Developers

## Who needs to review this
Developers

## Done when

- #2708 
- #2709 
- #2715
- #2714
- #2779 
[...]

PR #2891, opened 2022-10-03
Replace `gulp serve` with `npm run serve`

colinrotherham:
This PR partly fell out of the https://github.com/alphagov/govuk-frontend/pull/2861 work, and makes a few changes:

1. Replaces the `gulp serve` task with `npm run serve`
2. Adds a **nodemon.json** config file
4. Prevent Nodemon restarts for test files

For example, we don't need the review application Node.js process to restart when `checkboxes.test.js` is saved

Related to https://github.com/alphagov/govuk-frontend/issues/2717


gulpfile.js
```diff
@@ -52,24 +51,15 @@ gulp.task('compile', gulp.series(
...
-
 /**
  * Dev task
  * Runs a sequence of tasks on start
  */
 gulp.task('dev', gulp.series(
   cleanPublic,
   'compile',
-  'serve'
+  'watch',
```
domoscargin, line 61: Are there any "gotchas" we might miss running these tasks in slightly different order?

nodemon.json
```diff
@@ -1,5 +1,7 @@
 {
   "exec": "node ./app/start.js",
   "watch": ["app", "src/govuk"],
+  "ignore": ["**/*.test.*"],
+  "ext": "js",
```
domoscargin, line 5: Removing json I think will not catch the json file we have here: https://github.com/alphagov/govuk-frontend/blob/main/app/views/full-page-examples/search/data.json

But that's the only example I can find.

colinrotherham: Ahh good detective work, thank you. Missed that one 👍

gulpfile.js
colinrotherham: Not that I could see (as Nodemon does its own watching) and no compile steps are required for Express to start

colinrotherham: We still have `'compile'` coming first, so browser assets served by Express will already be in place too

nodemon.json
colinrotherham: Added `.json` again 😊

domoscargin (dismissed):
LGTM!

domoscargin (commented):
Sorry for pulling my approval last second!

Just want to check a workflow here.

It looks like now the `watch` task recompiles `.mjs` files to whichever folder is being used, but the app doesn't get restarted, since none of that should trigger `nodemon` right now.

So if a dev edits an `.mjs` file, currently they'd need to at least refresh the page to see the changes, right?

colinrotherham:
@domoscargin Yeah that's it, totally separate now but same behaviour as before

Previously you would see erroneous Nodemon restarts for client-side JavaScript, but nothing server-side had actually changed

I did wonder if **livereload** or **Browser Sync** would be a handy improvement in future, but their JavaScript likely won't meet our browser support requirements

domoscargin:
Ah great.

Had a quick play around and the previous set up still requires a refresh if you edit an `.mjs` file in `src`.

I had vague memories that the server restart process somehow got around that, and we were losing a bit of convenience, but that doesn't appear to be the case, so I'll approve again.

colinrotherham:
This PR was related to https://github.com/alphagov/govuk-frontend/issues/2243

colinrotherham:
Another one included by https://github.com/alphagov/govuk-frontend/issues/2243

PR #2900, opened 2022-10-07
Configure ESM support for all test files via Babel

colinrotherham:
This PR configures Jest so we can write ESM `*.test.mjs` files whenever we like 🙌 

Previously we used two transpilers (Babel or Rollup) which causes a bit of config anxiety when things don't work

### What's changed?
1. All Jest test projects support `*.test.js` and `*.test.mjs` extensions
2. Configured Babel to ignore "browser" transforms that the test runner doesn't need
3. Removed `rollup-jest` as Jest supports ESM already

### Why use a transformer?
Jest has used [`babel-jest` by default](https://jestjs.io/docs/configuration#transform-objectstring-pathtotransformer--pathtotransformer-object) to transform ESM-to-CommonJS for the last few releases

Native ESM support isn't provided out of the box and requires the Node.js `--experimental-vm-modules` flag. I've followed the Jest guidance to improve ES module testing as we may want to use this in future

https://jestjs.io/docs/ecmascript-modules
https://jestjs.io/docs/manual-mocks#using-with-es-module-imports


tasks/clean.unit.test.js
```diff
@@ -28,8 +28,8 @@ describe('Clean task', () => {
         paths: ['custom/location/here/**/*']
       }
     ]
-  )('cleans destination "$destination"', async ({ destination, paths }) => {
-    jest.mock('./task-arguments.js', () => ({ destination }))
+  )('cleans destination "$destination"', async ({ destination: mockDestination, paths }) => {
+    jest.mock('./task-arguments.js', () => ({ destination: mockDestination }))
```
colinrotherham, line 32: We need to follow the `babel-jest` approach of prefixing imports with "mock" when we _don't_ want them automatically hoisting to the top of the _transpiled_ file

This doesn't need documenting as the CLI will explain this to you

The Jest guidance on [Using with ES module imports](https://jestjs.io/docs/manual-mocks#using-with-es-module-imports) was useful

romaricpascal:
It'd be nice to have consistent extensions for tests, indeed (though I personally don't mind having some `.js` and others `.mjs`).

I'm a bit concerned about applying babel to all test code, though. Won't that pre-process files during tests when they are not pre-processed when actually run, either by `govuk-frontend` users or the build process?

With NodeJS supporting `import` natively, do we need to go through a transform to get ESM support for now?

Tying our tests to babel feel more of a thing we'll need to do if we introduce babel in our build, to make sure that tests run the same code as built, but I'm worried it's a bit premature here.

colinrotherham:
In future we may want to run our test suite with ESM natively but it [requires a Node.js experimental flag](https://jestjs.io/docs/ecmascript-modules)

babel.config.js
```diff
@@ -0,0 +1,15 @@
+module.exports = function (api) {
+  const browserslistEnv = !api.env('test')
```
colinrotherham, line 2: For tests we want to skip the default **Browserlist** rules

We're taking ESM files, making as few modifications as possible for Node.js 16, then running them in Jest as CommonJS out the other end

We don't want Jest running browser transformations on our test assertions

jest.config.js
```diff
@@ -1,5 +1,8 @@
 const config = {
-  cacheDirectory: '<rootDir>/.cache/jest/'
+  cacheDirectory: '<rootDir>/.cache/jest/',
+  transform: {
+    '^.+\\.m?js$': ['babel-jest', { rootMode: 'upward' }]
```
colinrotherham, line 4: [Jest defaults to `babel-jest` anyway](https://jestjs.io/docs/configuration#transform-objectstring-pathtotransformer--pathtotransformer-object) but we want it to find our Babel config

colinrotherham:
@romaricpascal Ready for another look 👍

romaricpascal (approved):
Looks good to me 🙌🏻 Makes sense to get rid of `rollup-jest` if `babel-jest` already handles loading `.mjs` files and is used by Jest by default anyway.

PR #2901, opened 2022-10-07
Cache generation of `componentsData` and directory lookups for tests

colinrotherham:
For slower computers (GitHub containers too) you'll notice our tests are still a bit slow 😅 

We've got a [collection of helpers in **./lib**](https://github.com/alphagov/govuk-frontend/tree/main/lib) `file-helper`, `jest-helper` etc

But Jest doesn't implement `require.cache` so you'll notice that any files we require/import in tests are executed hundreds of times—this does start to add up when you're looping directory listings 😮

So every call to `getExamples()` or `getComponent()` starts from scratch

### Test environments
This PR makes three changes:

1. Creates a custom Jest environment for Node.js
2. Caches frequently uses directory lookups and `getComponentsData`
3. Adds cached globals to our custom Jest environments

This is the same approach as [**jest-puppeteer**](https://github.com/smooth-code/jest-puppeteer) adding `page` and `browser` globals

Follows on from the work in https://github.com/alphagov/govuk-frontend/pull/2933

### DIFF
--- a/config/jest/environment/jsdom.mjs
+++ b/config/jest/environment/jsdom.mjs
@@ -1,8 +1,10 @@
 import { TestEnvironment } from 'jest-environment-jsdom'
 
+import { globals } from '../globals.mjs'
+
 /**
  * Virtual browser environment
- * Adds jsdom window/document globals
+ * Adds jsdom window/document globals, shared test globals
  */
 class BrowserVirtualEnvironment extends TestEnvironment {
   async setup () {
@@ -13,6 +15,10 @@ class BrowserVirtualEnvironment extends TestEnvironment {
 
     // Ensure test fails for browser exceptions
     virtualConsole.on('jsdomError', (error) => process.emit('error', error))
+
+    // Add shared test globals
+    // componentsData, componentsDirectory, examplesDirectory
+    this.global.cache ??= await globals()
   }
 }
 
--- a/config/jest/globals.mjs
+++ b/config/jest/globals.mjs
@@ -1,0 +1,39 @@
+import configPaths from '../paths.js'
+import { getDirectories, getComponentsData } from '../../lib/file-helper.js'
+
+/**
+ * Cache store
+ */
+const store = new Map()
+
+/**
+ * Cache frequently used paths
+ */
+const paths = [
+  configPaths.components,
+  configPaths.examples,
+  configPaths.fullPageExamples
+]
+
+/**
+ * Cache test globals
+ */
+export async function globals () {
+  if (!store.has('globals')) {
+    const componentsData = await getComponentsData()
+
+    // Run directory lookups
+    const directoryEntries = await Promise.all(paths.map(
+      async (path) => [path, await getDirectories(path)]
+    ))
+
+    // Save to cache
+    store.set('globals', {
+      directories: new Map(directoryEntries),
+      componentsData
+    })
+  }
+
+  // Retrieve from cache
+  return store.get('globals')
+}
--- a/config/jest/environment/node.mjs
+++ b/config/jest/environment/node.mjs
@@ -1,11 +1,18 @@
 import TestEnvironment from 'jest-environment-node'
 
+import { globals } from '../globals.mjs'
+
 /**
  * Default Node.js environment
+ * Adds shared test globals
  */
 class NodeEnvironment extends TestEnvironment {
   async setup () {
     await super.setup()
+
+    // Add shared test globals
+    // componentsData, componentsDirectory, examplesDirectory
+    this.global.cache ??= await globals()
   }
 }
 
--- a/config/jest/environment/node.mjs
+++ b/config/jest/environment/node.mjs
@@ -1,0 +1,12 @@
+import TestEnvironment from 'jest-environment-node'
+
+/**
+ * Default Node.js environment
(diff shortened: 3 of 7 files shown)


config/jest/globals.mjs
```diff
@@ -0,0 +1,39 @@
+import configPaths from '../paths.js'
+import { getDirectories, getComponentsData } from '../../lib/file-helper.js'
+
+/**
+ * Cache store
+ */
+const store = new Map()
```
romaricpascal, line 7: **issue**: I'm not sure we need that extra level of indirection there as that map only has a single key and is never exposed.

**suggestion**: Use something like `let globals = null` (or `cache`) and check for it being truthy for generating the cache.

config/jest/environment/jsdom.mjs
```diff
@@ -13,6 +15,10 @@ class BrowserVirtualEnvironment extends TestEnvironment {
 
     // Ensure test fails for browser exceptions
     virtualConsole.on('jsdomError', (error) => process.emit('error', error))
+
+    // Add shared test globals
+    // componentsData, componentsDirectory, examplesDirectory
+    this.global.namespace ??= await globals()
```
romaricpascal, line 21: **nitpick**: That name confused me a bit, I didn't get what it was the namespace of. As it's effectively the cache we're reading from, could we name it just `cache` or something cache related?

romaricpascal (commented):
Makes a lot of sense to cache the data we load from files again and again 🙌🏻 

**question**: I'm a bit confused by the need to set up a global variable for saving the cache, and then only accessing it in a single module afterwards (`file-helper`). What do we gain from the global and using Jest environment, compared to a simpler approach of having `let` variables in the `file-helper` module for example? Does Jest loads and reloads that module so there's no gain from caching when it keeps its environment intact?

config/jest/environment/jsdom.mjs
colinrotherham: Yeah, it was a general "dumping ground" for shared things name! I'm happy with cache

```suggestion
    this.global.cache ??= await globals()
```

config/jest/globals.mjs
colinrotherham: Ahh sorry, it has to be at this scope 😬 

It's to make sure we're exploiting the Node.js module cache
https://nodejs.org/api/modules.html#caching

Whether it's `globals.mjs` being imported 1000s of times or `await globals()` being run 1000s of times, it just needs to be something at module scope to work correctly

Happy to rename it to match the other comment?

```suggestion
const cache = new Map()
```

colinrotherham:
> Makes a lot of sense to cache the data we load from files again and again 🙌🏻
> 
> **question**: I'm a bit confused by the need to set up a global variable for saving the cache, and then only accessing it in a single module afterwards (`file-helper`). What do we gain from the global and using Jest environment, compared to a simpler approach of having `let` variables in the `file-helper` module for example? Does Jest loads and reloads that module so there's no gain from caching when it keeps its environment intact?

I've replied 😊 

Yeah if we added caching inside `lib/file-helper.js` functions it would work per test, not per project 😭

Jest has no module cache (or require cache) so every `lib/file-helper.js` import gets it "fresh" every time

config/jest/globals.mjs
romaricpascal: Sorry for the lack of clarity there, I wasn't discussing the scope, but the need to use a `Map` as we're only using one key. Wouldn't this be a simpler path than doing `set()` and `get()` on the `Map`:

```js
let cache = null;

export async function globals() {
  if (!cache) {
     // Compute cache
  }
  return cache;
}
```

romaricpascal:
@colinrotherham Cheers for the explanation, that makes sense (shame we can't opt-in to Jest doing module caching or something 😭 ) 

Another question that poped to mind:

Do the environments get re-loaded in watch mode or are they kept from one run to another (first makes more sense to me to ensure runs are isolated, hoping it also did for the people building Jest)?

config/jest/globals.mjs
romaricpascal: Regarding renaming, I think it makes sense only if we move to that variable being the cache itself and not a `Map` ready to store multiple caches. Otherwise, the cache will actually live at `store.get('globals')` so it'll cause some confusion to have the store named `cache` in that scenario, I think (hopefully my sentence is not confusing 🤣 ).

colinrotherham: Yeah it's difficult. Shall we keep it the same then?

Was trying to think of a generic name `store` that represented "things this module wants to keep track of", without necessarily (as you say) representing the `cache` global. Cheers though!

romaricpascal: Fair enough. In that case, I'd say let's leave this module as it is, but rename the global in the tests so we have a clearer name than `namespace` there 😊

colinrotherham:
@romaricpascal All ready to review again

Yeah the cache gets cleared at the end of every test run (including `--watch` and `--watchAll`) as we've set our cache in environment `setup()`. Automatically on `super.teardown()` browsers close, review app stops, and cache clears 🎉

romaricpascal (approved):
@colinrotherham Sounds all good then, let's ship it 🎉

PR #2902, opened 2022-10-07
Reduce GitHub Test workflow from 4-5 mins to ~2 mins

colinrotherham:
I've had this code branched for a little while, but would be good to discuss

Along with https://github.com/alphagov/govuk-frontend/pull/2861, https://github.com/alphagov/govuk-frontend/pull/2878, https://github.com/alphagov/govuk-frontend/pull/2897, https://github.com/alphagov/govuk-frontend/pull/2901 we can run our tests nearly twice as fast

For our contributors, it's nice to see visually why (and where) the GitHub checks have failed

**Was it the install?
Was it code linting?
Was it build related?
Was it tests?**

Now you know 🙌 

Showing test jobs visually


colinrotherham:
>Expected — Waiting for status to be reported

This PR would need the **Branch protection rules** → **Require status checks to pass before merging** checks changing (see GitHub repository settings) from a single "Run tests" to each separate check:

1. **Lint Sass**
2. **Lint JavaScript**
3. **JavaScript unit tests**
4. **JavaScript behaviour tests**
5. **JavaScript component tests**
6. **Nunjucks macro tests**

Probably also:

1. **Verify built ./package** 
2. **Verify built ./dist**

These status checks are duplicated when also [running on Windows](https://github.com/alphagov/govuk-frontend/pull/2938)

All up for discussion

.github/workflows/actions/build/action.yaml
```diff
@@ -0,0 +1,30 @@
...
+    - name: Cache build
+      uses: actions/cache@v3
+      id: cache
+      with:
+        key: build-cache-${{ github.sha }}
+        path: |
+          dist
+          package
+          public
+          sassdoc
```
colinrotherham, line 16: Need to add jsdoc!

```suggestion
          sassdoc
          jsdoc
```

romaricpascal:
Thanks for running me through it in call, it all makes sense now 🙌🏻 Just putting in this comment of what I understood, as it may be helpful when looking back at the PR (might be worth adding description/comments directly in the files, though). Let me know if I missed or misundestood anything.

> The PR splits the `test` action into multiple actions, allowing to run some parts in parallel but with some caching so that we don't re-install or re-build things in each container.
> 
> 1. Install
> 
> Sets up node (if version has changed) and install the dependencies (if they have changed).
> 
> 2. Build
> 
> Builds the project and caches the build results against the commit hash.
> 
> 3. Lint
> 
> Runs the linting of SCSS and JavaScript files.
> 
> Caching restores any cache starting with `${{matrix.name}}-cache-`
> to allow ESlint/Stylelint to still get a cache if a change to `package-lock.json` happened that didn't update anything Stylelint/ESLint related. Both tools use their own caching mechanisme to ensure they re-run properly if the file,config,plugins have changed, allowing us to not worry about that and just save/restore the folder they use as cache.
> 
> 4. Test
> 
> Runs the JavaScript test not depending on the build results,
> allowing them to start early. 
> 
> Similarly to linting, caching will restore any cache starting with `${{matric.name}}-cache` to speed things up when it's not a test dependency that led to `package-lock.json` changing.
> 
> 5. Verify
> 
> Runs the JavaScript tests that rely on the built results.

Regarding the required check missing, just putting in the idea of a last "Run tests" step that would depend on the lint, test and verify ones and do nothing but allow us not to have to update which required steps we have at the moment. This would rely on that last step not running if any of the previous steps fail, though (which would make sense to me, but I'm not sure if that's how Github runs the different actions).

colinrotherham:
Thanks @romaricpascal

The run through in https://github.com/alphagov/govuk-frontend/pull/2902#issuecomment-1296992924 makes sense

I've added comments and split things up for further clarity. It's ready for review again now

Sadly for the "**Run tests**" status check we'll need to split this out further for Windows via https://github.com/alphagov/govuk-frontend/pull/2945. We decided recently that whilst most Linux status checks are required to merge, their Windows counterparts might not be

Accepting that the single "**Run tests**" status needs splitting out gives us this flexibility

romaricpascal:
Neat! Cheers for going through and commenting things 🙌🏻 The workflow looks good (and I can understand what's happening, cheers again for the walk through).

Guess the question would be how to coordinate the update of the checks, as changing them is for the whole repo and not on a per-branch basis, isn't it?

colinrotherham:
>Guess the question would be how to coordinate the update of the checks, as changing them is for the whole repo and not on a per-branch basis, isn't it?

@romaricpascal Yeah, whole repo

Can we see if the "Required" status check is merge-blocking or not, if you're happy to approve?

We can either:

1. Remove the "Run tests" status check and do nothing for now
2. Replace the "Run tests" status check with the new ones (all PRs to be rebased)

Don't have to add the new checks straight away

Even with the status check removed (temporarily) we'll still see the checks pass or fail

colinrotherham:
Yep, the old "**Run tests**" status check does disable the "Merge pull request" button

We can still take the two options above:

>1. Remove the "Run tests" status check and do nothing for now
>2. Replace the "Run tests" status check with the new ones (all PRs to be rebased)

I'd be happy if we rebased all active PR branches to use the new workflow + checks, but a short period (riskier) with the status check removed might be preferred. I'll gather feedback from Slack

PR #2906, opened 2022-10-10
Remove Cheerio from behaviour tests for radios and checkboxes

### LINKED ISSUE #2898: Remove Cheerio from behaviour tests for radios and checkboxes
## What

At the minute, the behaviour tests for radios and checkboxes load the examples in a headless browser using Puppeteer, then grab the HTML out of the page and parse it with Cheerio.

They then either make assertions about the makeup of the component based on the state of the Cheerio object, or grab attributes out of the Cheerio object to then make assertions about what's going on in Puppeteer (e.g. waiting for selectors)

## Why

There's no need to use Cheerio here – everything that we're doing with Cheerio could be done simpler using Puppeteer methods like [`page.$evaluate`](https://pptr.dev/api/puppeteer.page._eval/).

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [ ] Behaviour tests for Radios no longer use Cheerio
- [ ] Behaviour tests for Checkboxes no longer use Cheerio

colinrotherham:
This PR closes #2898

But also makes a few other changes:

1. Uses Puppeteer jQuery-style `$()` and `$$()` methods
2. Adds `beforeAll()` with **$components**, **$inputs** for each test
3. Adds new `getAttribute()`, `getProperty()`, `isVisible()` puppeteer helpers

Keeps each test block short and understandable

### DIFF
--- a/lib/puppeteer-helpers.js
+++ b/lib/puppeteer-helpers.js
@@ -111,10 +111,22 @@ async function getProperty ($element, propertyName) {
   return handle.jsonValue()
 }
 
+/**
+ * Get attribute value for element
+ *
+ * @param {import('puppeteer').ElementHandle} $element - Puppeteer element handle
+ * @param {string} attributeName - Attribute name to return value for
+ * @returns {Promise<string | null>} Attribute value
+ */
+function getAttribute ($element, attributeName) {
+  return $element.evaluate((el, name) => el.getAttribute(name), attributeName)
+}
+
 module.exports = {
   renderAndInitialise,
   goTo,
   goToComponent,
   goToExample,
+  getAttribute,
   getProperty
 }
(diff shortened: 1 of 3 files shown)


36degrees (approved):
Thanks for picking this up 🙌🏻  This looks great to me

Feels like there might be an opportunity to write some custom matchers that work with ElementHandles to make some of this even easier in the future?

It'd be neat to be able to do e.g. `expect($element).toBeVisible()` or `expect($element).toHaveProperty('checked', false)`. Sort of surprised this doesn't already exist – if it does, I can't find it!

(Not suggesting we do any of this as part of this PR)

colinrotherham:
I had to stop myself, I did have them! We might as well switch to Cypress or Playwright

(Always easier when someone has done it all for you already)

Issue #2907, opened 2022-10-10
Limit side effects of polyfills loaded for config/I18n


romaricpascal:
## What

Ensure only components needing the `dataset`, `closest` and possibly `trim` polyfills import them. This'll help reduce bundle size for non-configurable or non-internationalisable components.

Possible approach could be to:

- make a separate module for `closestAttributeValue` that imports `Element.prototype.closest`
- make a separate module for `normaliseString` that imports `String.prototype.trim`
- import `Element.prototype.dataset` in all components requiring configuration as they access the `dataset`. Given that one will be an easy one to miss, I wonder if we shouldn't have a `getDataset` function that has side effect.

These files could live in an `src/polyfilled` folder that we'll revisit for v5.

## Why

@colinrotherham wrote a neat explanation in this comment: https://github.com/alphagov/govuk-frontend/pull/2887#discussion_r991168049

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [ ] Polyfills don't burden components that don't require them when the components are imported individually

PR #2908, opened 2022-10-10
Add polyfill for `String.prototype.trim`

### LINKED ISSUE #2903: The `normaliseData` function uses `String.trim` which is not supported in IE8/IE9
## Description of the issue

In #2840 we introduced a new `normaliseString` helper method which calls `String.trim` which is [not supported in IE8 or IE9](https://caniuse.com/mdn-javascript_builtins_string_trim):

https://github.com/alphagov/govuk-frontend/blob/cf7b06ceb19236a9afeb5399158b35391219cf99/src/govuk/common.mjs#L152

This change has not yet been released.

This causes a script error when going to any page in the review app which contains a component that accepts config via data attributes.

Although we have plans to drop support for IE8-10 shortly, our [current browser support](https://github.com/alphagov/govuk-frontend#browser-and-assistive-technology-support) includes Internet Explorer 8, 9 and 10, 'although components may not look perfect'. A change in browser support would be considered a breaking change, so we should fix this before we do the next release.

1. Go to http://govuk-frontend-review.herokuapp.com/components/accordion/preview in IE8

Expected: The page should load without error

Actual: Script error ('Object doesn't support this property or method')

- Operating system: Windows 7 (via Browserstack)
- Browser: Internet Explorer
- Browser version: 8.0
- GOV.UK Frontend Version: [NEXT]

36degrees:
In bc843a3 (#2840) we introduced a new [`normaliseString` helper method][1] which relies on [`String.prototype.trim`][2], which is [not supported in IE8 or IE9][3].

This causes a script error when going to any page in the review app (in IE8) which contains a component that accepts config via data attributes.

Add a polyfill for `String.prototype.trim` based on [the polyfill that used to be on MDN][4] (before they removed content relating to IE8), adapted to match the format used for the existing polyfills.

Closes #2903.

[1]: https://github.com/alphagov/govuk-frontend/blob/cf7b06ceb19236a9afeb5399158b35391219cf99/src/govuk/common.mjs#L152
[2]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim
[3]: https://caniuse.com/mdn-javascript_builtins_string_trim
[4]: https://github.com/mdn/content/blob/cf607d68522cd35ee7670782d3ee3a361eaef2e4/files/en-us/web/javascript/reference/global_objects/string/trim/index.md#polyfill

### DIFF
--- a/src/govuk/common.mjs
+++ b/src/govuk/common.mjs
@@ -1,4 +1,5 @@
 import './vendor/polyfills/Element/prototype/dataset.mjs'
+import './vendor/polyfills/String/prototype/trim.mjs'
 
 /**
  * TODO: Ideally this would be a NodeList.prototype.forEach polyfill
--- a/src/govuk/vendor/polyfills/String/prototype/trim.mjs
+++ b/src/govuk/vendor/polyfills/String/prototype/trim.mjs
@@ -1,0 +1,13 @@
+(function(undefined) {
+
+    // Detection from https://github.com/mdn/content/blob/cf607d68522cd35ee7670782d3ee3a361eaef2e4/files/en-us/web/javascript/reference/global_objects/string/trim/index.md#polyfill
+    var detect = ('trim' in String.prototype)
+    
+    if (detect) return
+
+    // Polyfill from https://github.com/mdn/content/blob/cf607d68522cd35ee7670782d3ee3a361eaef2e4/files/en-us/web/javascript/reference/global_objects/string/trim/index.md#polyfill
+    String.prototype.trim = function () {
+        return this.replace(/^[\s\uFEFF\xA0]+|[\s\uFEFF\xA0]+$/g, '');
+    };
+
+}).call('object' === typeof window && window || 'object' === typeof self && self || 'object' === typeof global && global || {});


colinrotherham (approved):
Approving, but subject to us updating the [**sideEffects** package field](https://github.com/alphagov/govuk-frontend/blob/main/package/package.json#L16)

36degrees:
> Approving, but subject to us updating the [**sideEffects** package field](https://github.com/alphagov/govuk-frontend/blob/main/package/package.json#L16)

To summarise a conversation on Slack:

- the issue we're concerned about is that e.g. someone importing e.g. `{ nodeListForEach } from 'govuk-frontend/package/govuk-esm/common.mjs` gets the polyfills 'unsolicited' as side-effects when they don't want or need them
- we think that `govuk-esm/vendor` being listed in `sideEffects` _should_ mean that a bundler would understand that because `common.mjs` imports from `vendor` it also has side effects, but are concerned that not all bundlers would do this
- we think a better solution is to avoid importing polyfills from common.mjs, which we're planning to do explore as part of #2907 – I've attached that issue to the [NEXT] milestone to make sure that we actually do it!
- therefore we're going to merge this as-is and revisit as part of that issue before we release

Issue #2890, opened 2022-09-30
Allow users to suppress deprecation warnings


owenatgov:
## What
Add a sass settings map and accompanying function that does the following:

- Acts as a repository for all our currently active deprecation warnings
- Allow users to suppress one or more given warnings using a map of keys
- Prevent warnings from firing more than once

## Why
As we move towards v5, which includes deprecated settings, we don't want those why have to remain on v4.X to be stuck with deprecation warnings.

This was also inspired by https://github.com/alphagov/govuk-frontend/pull/2882 where the only issue with deprecating these settings were the fact that these settings are set in our review app and we could use the above warning suppression map to manage this more robustly.

Part of https://github.com/alphagov/govuk-frontend/issues/2769

## Who needs to work on this
Developers

PR #2911, opened 2022-10-10
Add warning suppression functionality

owenatgov:
Fixes https://github.com/alphagov/govuk-frontend/issues/2890

This PR does the following:

- Adds the `$govuk-suppressed-warnings` map
- Adds the `_warning` mixin
- Replaces all instances of `@warn` with the `_warning` mixin
- Updates the managing-change docs with why we allow warning suppression

The `settings/warnings` suite (`$govuk-suppressed-warnings` and `_warning`) are intended as a way to allow users to suppress deprecation warnings in instances where they're stuck on specific versions so they have the choice to keep their sass compile log clean. It also will allow us to suppress the compatibility mode deprecation warnings in our local legacy app between now and the release of v5.0.

More details are in the commit history.


src/govuk/settings/_warnings.scss
```diff
@@ -0,0 +1,50 @@
...
+/// You can add to this map and define which warnings to suppress by appending to
+/// it using the warning key, found in the warning message. For example:
+///
+/// @example scss:
+///   // warning message:
+///   //  $foobar is no longer supported. To silence this warning, update
+///   //  $govuk-suppressed-warnings with key: "foobar"
+///   $govuk-suppressed-warnings: append($govuk-suppressed-warnings, "foobar");
+///
+/// @type Map
```
romaricpascal, line 22: There is no key-value association by the description of how it's used, so it's more a [List](https://sass-lang.com/documentation/values/lists) than a[Map](https://sass-lang.com/documentation/values/maps)

romaricpascal (commented):
Looking pretty much there for me 🙌🏻 Just noticed we're missing a test case and nitpicked on the terminology for the `$govuk-suppressed-warnings` map 😊

src/govuk/settings/_warnings.scss
```diff
@@ -0,0 +1,50 @@
...
+/// @access public
+
+$govuk-suppressed-warnings: () !default;
+
+/// Warnings
+///
+/// A private mixin which act as a wrapper for the built in `@warn` sass function
```
36degrees, line 29: Given we have the `@access` annotation, I'm not sure it's necessary to flag it as private in the description as well.

docs/contributing/managing-change.md
```diff
@@ -171,3 +171,11 @@ Keep the old name in the selector list, and mark it as deprecated.
   foo: bar;
 }
 ```
+
+### The `_warning` mixin and allowing users to suppress warnings
```
36degrees, line 175: Thanks for adding this! 🙏🏻 📚

CHANGELOG.md
```diff
@@ -58,6 +58,16 @@ This was added in pull requests:
 - [#2818: Add support for localisation via data-* attributes to Accordion component](https://github.com/alphagov/govuk-frontend/pull/2818)
 - [#2826: Add support for localisation via JavaScript configuration to Accordion component](https://github.com/alphagov/govuk-frontend/pull/2826)
 
+#### Suppress deprecation warnings
+
+You can now suppress warnings from deprecations within GOV.UK Frontend by updating the `$govuk-suppressed-warnings` map in sass. Every deprecation warning will now include a warning "key" which you can use in the following code, placed at the root of your sass project:
+
+```
+$govuk-suppressed-warnings: append($govuk-suppressed-warnings, $key);
```
36degrees, line 66: I don't think this is what we want people to do for setting the initial state of the list (especially as I'm not sure you can append to a non-existent list).

In user land this could just be:

```scss
$govuk-suppressed-warnings: (
  legacy-palette,
  legacy-font
);
```

src/govuk/settings/warnings.test.js
```diff
@@ -0,0 +1,62 @@
...
+          'This is a warning. To silence this warning, update ' +
+          '$govuk-suppressed-warnings with key: "test"'
+        )
+    })
+  })
+
+  it('Only fires one @warn per warning key', async () => {
+    const sass = `
+    ${sassBootstrap}
+    @include _warning('test', 'This is a warning.');`
```
romaricpascal, line 42: I think for that test to test what it announces, we need to call the mixin twice. Otherwise, we'll get one call, but because there was only one call.

```suggestion
    @include _warning('test', 'This is a warning.');`
    @include _warning('test', 'This is a warning.');`
```

docs/contributing/managing-change.md
```diff
@@ -171,3 +171,11 @@ Keep the old name in the selector list, and mark it as deprecated.
...
+
+In the above examples we've used `@include _warning(...)` instead of the native sass `@warn` at-rule. We do this instead of using `@warn` because it gives users the option to suppress deprecation warnings by interacting with the `$govuk-suppressed-warnings` map.
+
+The `_warning` mixin takes 2 parameters: a warning key (`$key`) and a warning message (`$message`), checks if `$key` isn't already included in `$govuk-suppressed-warnings` and if it isn't, outputs a warning via `@warn` using `$message` plus a notice to add the `$key` to `$govuk-suppressed-warnings` if they want to silence the warning.
```
claireashworth, line 179:
```suggestion
If users want to silence the warning, the `_warning` mixin looks at 2 parameters, a warning key (`$key`) and a warning message (`$message`), and checks whether `$key` is included in `$govuk-suppressed-warnings`. If `$key` is not included, the `_warning` mixin outputs a warning through `@warn` using `$message` and adds a notice to add the `$key` to `$govuk-suppressed-warnings`.```

Comment (claireashworth) [reply docs/contributing/managing-change.md:179]: Please check whether I've changed the meaning of this.

File: docs/contributing/managing-change.md [hunk 7]
```diff
@@ -171,3 +171,11 @@ Keep the old name in the selector list, and mark it as deprecated.
...
+
+We allow users to do this because they don't always have the means to action deprecation warnings or upgrade their codebase beyond a specific version of GOV.UK Frontend eg: a legacy codebase that doesn't have the resource to upgrade to the latest breaking change where a deprecated feature will be removed. This feature allows those users to continue to operate their codebase without having to "tolerate" deprecation warnings in their testing.
```
claireashworth, line 181:
```suggestion
We make this option available for users because they can not always action deprecation warnings or upgrade their codebase beyond a specific version of GOV.UK Frontend. For example, a legacy codebase that does not have the resource to upgrade to the latest breaking change where a deprecated feature will be removed. This feature allows those users to continue to operate their codebase without having to repeatedly see non-actionable deprecation warnings in their testing.
```

src/govuk/settings/_warnings.scss
```diff
@@ -0,0 +1,53 @@
+////
+/// @group settings/warnings
+////
+
+/// Suppressed warnings map
+///
+/// This map is used to determine which deprecation warnings to **not** show
+/// to users when compiling sass. This is in place for codebases that don't
```
claireashworth, line 8:
```suggestion
/// to users when compiling sass. This is in place for codebases that do not
```

src/govuk/settings/_warnings.scss
```diff
@@ -0,0 +1,53 @@
...
+/// @type List
+/// @access public
+
+$govuk-suppressed-warnings: () !default;
+
+/// Warnings
+///
+/// Acts as a wrapper for the built in `@warn` sass function
+///
+/// We use this instead of just using `@warn` for 3 reasons:
```
claireashworth, line 33:
```suggestion
/// We use this instead of using `@warn` for 3 reasons:
```

src/govuk/settings/_warnings.scss
```diff
@@ -0,0 +1,53 @@
...
+///
+/// - To check if a warning is being suppressed via `$govuk-suppressed-warnings`,
```
claireashworth, line 35:
```suggestion
/// - To check if a warning is being suppressed through `$govuk-suppressed-warnings`,
```

docs/contributing/managing-change.md
owenatgov: I don't think this changes the meaning, however looking at it now that you've tidied it up I think the last sentence needs a bit more detail to make it clear that the `@warn` output is the passed `$message` with some instructions on how to silence it:

```
If `$key` is not included, the `_warning` mixin outputs a warning through `@warn` using `$message` with the addition of notice to add the `$key` to `$govuk-suppressed-warnings` as the warning text.
```


Let me know what you think.

querkmachine:
Code-wise this is looking pretty cushy to me, but I'll leave the formal review for the folks who've been looking at it for longer. 😉

docs/contributing/managing-change.md
claireashworth: yup, that is better.

claireashworth (approved):

docs/contributing/managing-change.md
36degrees: I think although it's mostly correct this has got a little muddled, as the mixin internally works the same whether the user wants to silence the warnings or not.

Try as I might, I also can't get my head around what this sentence means:

> If `$key` is not included, the `_warning` mixin outputs a warning through `@warn` using `$message` with the addition of notice to add the `$key` to `$govuk-suppressed-warnings` as the warning text.

What about something like:

> The `_warning` mixin accepts 2 parameters, a warning key (`$key`) and a warning message (`$message`). The warning message is only outputted if the `$key` is not already in the `$govuk-suppressed-warnings` list. Once the message has been outputted, `$key` is then added to the `$govuk-suppressed-warnings` list.
> 
> This means that only one warning be outputted for a given `$key`, and that users can suppress warnings by adding them to the `$govuk-suppressed-warnings` list before they include GOV.UK Frontend.

owenatgov: That's true actually I missed that. The `_warning` mixin has nothing to do with the user wanting to suppress warnings as its a private mixin 🙃  Users interact with this functionality via `$govuk-suppressed-warnings`.

This makes sense to me. @claireashworth Do you have an opinion?

owenatgov: As for that line you didn't understand, this is trying to explain that the `@warn` within `_warning` is going to output `$message` plus a little addendum telling users they can suppress warnings. It's also contextual to the previous sentence which explains how `$key` and `$govuk-suppressed-warnings` interact.

I'm starting to wonder if we include this paragraph at all since I feel like it's trying to do the job that the sassdoc for `$govuk-suppressed-warnings` and `_warning` are already doing. Do either of you object to me removing this paragraph altogether, potentially replacing it with something like:

>You can read more about how `$govuk-suppressed-warnings` and `_warning` work by reading their respective sassdocs.

Ideally I'd link out to the docs page but it's not live yet. Maybe something to do post release?

36degrees: The docs for `_warning` won't end up in Frontend Docs because we don't include private things in our API reference, but otherwise that sounds good to me.

owenatgov: I've scooped it out. Let me know what you reckon now.

owenatgov:
With 2 approvals I'm gonna go ahead and merge this. There was a slight change since Claire's approval to the managing change docs but this is intended as internal content and can be changed at any time so I'm not worried about this.

PR #2912, opened 2022-10-11
Use single `compileStyles` Sass helper function

colinrotherham:
This PR is a follow up to https://github.com/alphagov/govuk-frontend/pull/2881

Knowing we now have a **postcss.config.js** file that handles:

1. Adding prefixes (via Autoprefixer)
2. Adding pseudo-class companions for stylesheets starting `app-*`
3. Running plugins for IE8 compatibility (removing media queries etc)
4. Minifying stylesheets ending `*.min.css`

We can now combine the various task functions `compileOldIE()`, `compileLegacy()`, `compileLegacyIE()`, `compileFullPageStyles()` into a single `compileStyles()`

### DIFF
--- a/tasks/gulp/compile-assets.js
+++ b/tasks/gulp/compile-assets.js
@@ -15,109 +15,80 @@ const merge = require('merge-stream')
 const rename = require('gulp-rename')
 
 const configPaths = require('../../config/paths.js')
-const { destination, isDist, isPublic } = require('../task-arguments.js')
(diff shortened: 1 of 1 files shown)


colinrotherham:
Also comes under "does too much" so is also linked with https://github.com/alphagov/govuk-frontend/issues/2243

36degrees (approved):

Issue #2910, opened 2022-10-10
CharacterCount doesn't render or initialise properly if `maxlength` or `maxwords` is not passed to the macro


romaricpascal:
## What

It's a two part thing:

First is the HTML part, before JS kicks in: if omitting the `maxlength` or `maxwords` option when calling the `govukCharacterCount` macro, the HTML rendered displays a "You can enter up to undefined characters" message.

<img width="596" alt="CharacterCount component with faulty 'You can enter up to undefined characters' hint" src="https://user-images.githubusercontent.com/396367/194913911-c108079d-76e3-41c9-b55d-2a68cd3e08ed.png">

Second is the JavaScript part: When initialising the component with a JavaScript `maxlength` or `maxwords` config, the message disappears, but doesn't get replace by a count of the remaining character. 

<img width="611" alt="CharacterCount component without message after JavaScript kicks in" src="https://user-images.githubusercontent.com/396367/194914302-7a6b44d1-8f09-4374-940e-30e5a5b59e4f.png">

This will be problematic for people looking to use the JavaScript configuration to set a blanket setting accross their service. For that to work, we need to render the component without a `data-maxwords` or `data-maxlength` attribute and fix the HTML part.

And then once that's through, we need to make sure the component initialises OK with a missing length.

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [x] Library users can get a clean render of the count message when no `maxwords` or `maxlength` are provided
- [x] JavaScript initialises properly when running over a HTML generated by the macro without `maxwords` or `maxlength`

romaricpascal:
This happens with the current version of `govuk-frontend`, in Firefox. Latest version's JavaScript handles things OK in terms of replacing the message 🥳 

However the macro still outputs an untidy text and we could either hide it as I initially suggester (using `govuk-visually-hidden` like the JS does), or replace the default to "You can enter a limited number of characters".

36degrees:
Yeah, unfortunately I think this has been around a while – see also #1725. I think it'd be nice to fix this in v4.4.0, but I don't think this is a regression.

romaricpascal:
That was just something I noticed while preparing my demo yesterday juggling between the 4.3.1 and main versions, but I prefered flagging to keep track of it.

For the JavaScript bit, thinking about it, without any default for `maxlength` removing the message is probably fine. We can't really set a default anyway for that `maxlength` as a "generic number of charaters to limit a textarea" feels quite arbitrary.

[...]

PR #2915, opened 2022-10-13
Add non-broken fallback text when maxlength/maxwords is 0

romaricpascal:
Avoids displaying a "You can enter **undefined** characters" to users if neither `maxlength` nor `maxwords` is set.

This'll avoid services from getting an unsavory message in front of their users if they chose to use JavaScript to pass the `maxlength` or `maxcount` option. (Closes https://github.com/alphagov/govuk-frontend/issues/2910).

Unfortunately, before JavaScript kicks in, we can't tell if it'll be characters or words (nor the limit obviously), so the message is pretty vague (but still better than having a message that looks like there was a bug). @claireashworth, if you have a better wording, I'm open to suggestions.

Given we're expecting JavaScript to kick in and set a dynamic text based on the content of the component, we do need to have an element on the page taking some vertical space to avoid layout shifts, so some content is required. And as we're passing the option as `text`, `&nbsp;` or other kind of HTML entities will be escaped.

### DIFF
--- a/src/govuk/components/character-count/character-count.mjs
+++ b/src/govuk/components/character-count/character-count.mjs
@@ -121,18 +121,18 @@ CharacterCount.prototype.init = function () {
   }
 
   var $textarea = this.$textarea
-  var $fallbackLimitMessage = document.getElementById($textarea.id + '-info')
+  var $textareaDescription = document.getElementById($textarea.id + '-info')
 
   // Inject a decription for the textarea if none is present already
   // for when the component was rendered with no maxlength, maxwords
   // nor custom textareaDescriptionText
-  if ($fallbackLimitMessage.textContent.match(/^\s*$/)) {
-    $fallbackLimitMessage.textContent = this.i18n.t('textareaDescription', { count: this.maxLength })
+  if ($textareaDescription.textContent.match(/^\s*$/)) {
+    $textareaDescription.textContent = this.i18n.t('textareaDescription', { count: this.maxLength })
   }
 
-  // Move the fallback count message to be immediately after the textarea
+  // Move the textarea description to be immediately after the textarea
   // Kept for backwards compatibility
-  $textarea.insertAdjacentElement('afterend', $fallbackLimitMessage)
+  $textarea.insertAdjacentElement('afterend', $textareaDescription)
 
   // Create the *screen reader* specific live-updating counter
   // This doesn't need any styling classes, as it is never visible
@@ -140,20 +140,20 @@ CharacterCount.prototype.init = function () {
   $screenReaderCountMessage.className = 'govuk-character-count__sr-status govuk-visually-hidden'
   $screenReaderCountMessage.setAttribute('aria-live', 'polite')
   this.$screenReaderCountMessage = $screenReaderCountMessage
-  $fallbackLimitMessage.insertAdjacentElement('afterend', $screenReaderCountMessage)
+  $textareaDescription.insertAdjacentElement('afterend', $screenReaderCountMessage)
 
   // Create our live-updating counter element, copying the classes from the
-  // fallback element for backwards compatibility as these may have been
+  // textarea description for backwards compatibility as these may have been
   // configured
   var $visibleCountMessage = document.createElement('div')
-  $visibleCountMessage.className = $fallbackLimitMessage.className
+  $visibleCountMessage.className = $textareaDescription.className
   $visibleCountMessage.classList.add('govuk-character-count__status')
   $visibleCountMessage.setAttribute('aria-hidden', 'true')
   this.$visibleCountMessage = $visibleCountMessage
-  $fallbackLimitMessage.insertAdjacentElement('afterend', $visibleCountMessage)
+  $textareaDescription.insertAdjacentElement('afterend', $visibleCountMessage)
 
-  // Hide the fallback limit message
-  $fallbackLimitMessage.classList.add('govuk-visually-hidden')
+  // Hide the textarea description
+  $textareaDescription.classList.add('govuk-visually-hidden')
 
   // Remove hard limit if set
   $textarea.removeAttribute('maxlength')
--- a/src/govuk/components/character-count/character-count.mjs
+++ b/src/govuk/components/character-count/character-count.mjs
@@ -29,7 +29,7 @@ var TRANSLATIONS_DEFAULT = {
     one: 'You have %{count} word too many',
     other: 'You have %{count} words too many'
   },
-  fallbackHint: {
+  textareaDescription: {
     other: ''
   }
 }
@@ -125,9 +125,9 @@ CharacterCount.prototype.init = function () {
 
   // Inject a decription for the textarea if none is present already
   // for when the component was rendered with no maxlength, maxwords
-  // nor custom fallbackHintText
+  // nor custom textareaDescriptionText
   if ($fallbackLimitMessage.textContent.match(/^\s*$/)) {
-    $fallbackLimitMessage.textContent = this.i18n.t('fallbackHint', { count: this.maxLength })
+    $fallbackLimitMessage.textContent = this.i18n.t('textareaDescription', { count: this.maxLength })
   }
 
   // Move the fallback count message to be immediately after the textarea
@@ -383,7 +383,7 @@ export default CharacterCount
  * @property {PluralisedTranslation} [wordsUnderLimit] - Words under limit
  * @property {string} [wordsAtLimit] - Words at limit
  * @property {PluralisedTranslation} [wordsOverLimit] - Words over limit
- * @property {PluralisedTranslation} [fallbackHint] - Fallback hint
+ * @property {PluralisedTranslation} [textareaDescription] - Fallback hint
  */
 
 /**
--- a/src/govuk/components/character-count/character-count.mjs
+++ b/src/govuk/components/character-count/character-count.mjs
@@ -120,6 +120,13 @@ CharacterCount.prototype.init = function () {
   var $textarea = this.$textarea
   var $fallbackLimitMessage = document.getElementById($textarea.id + '-info')
 
+  // Inject a decription for the textarea if none is present already
+  // for when the component was rendered with no maxlength, maxwords
+  // nor custom fallbackHintText
+  if ($fallbackLimitMessage.textContent.match(/^\s*$/)) {
+    $fallbackLimitMessage.textContent = this.i18n.t('fallbackHint', { count: this.maxLength })
+  }
+
   // Move the fallback count message to be immediately after the textarea
   // Kept for backwards compatibility
   $textarea.insertAdjacentElement('afterend', $fallbackLimitMessage)
@@ -373,6 +380,7 @@ export default CharacterCount
  * @property {PluralisedTranslation} [wordsUnderLimit] - Words under limit
  * @property {string} [wordsAtLimit] - Words at limit
  * @property {PluralisedTranslation} [wordsOverLimit] - Words over limit
+ * @property {PluralisedTranslation} [fallbackHint] - Fallback hint
  */
 
 /**
--- a/src/govuk/components/character-count/character-count.mjs
+++ b/src/govuk/components/character-count/character-count.mjs
@@ -28,6 +28,9 @@ var TRANSLATIONS_DEFAULT = {
   wordsOverLimit: {
     one: 'You have %{count} word too many',
     other: 'You have %{count} words too many'
+  },
+  fallbackHint: {
+    other: ''
   }
 }
 
--- a/src/govuk/components/character-count/character-count.test.js
+++ b/src/govuk/components/character-count/character-count.test.js
@@ -429,6 +429,26 @@ describe('Character count', () => {
           const visibility = await page.$eval('.govuk-character-count__status', el => window.getComputedStyle(el).visibility)
           expect(visibility).toEqual('visible')
         })
+
+        it('fills the accessible description of the textarea once the maximum is known', async () => {
+          // This tests that any fallback hint provided through data-attributes
+          // (or the Nunjucks macro) waiting for a maximum to be provided in JavaScript config
+          // will lead to the message being injected in the element holding the textarea's accessible description
+          // (and interpolated to replace `%{count}` with the maximum)
+
+          await renderAndInitialise(page, 'character-count', {
+            nunjucksParams: examples['no maximum'],
+            javascriptConfig: {
+              maxlength: 10
+            }
+          })
+
+          const message = await page.$eval(
+            '.govuk-character-count__message',
+            (el) => el.innerHTML.trim()
+          )
+          expect(message).toEqual('No more than 10 characters')
+        })
       })
 
       describe('via `initAll`', () => {
--- a/src/govuk/components/character-count/character-count.test.js
+++ b/src/govuk/components/character-count/character-count.test.js
@@ -437,7 +437,8 @@ describe('Character count', () => {
           // (and interpolated to replace `%{count}` with the maximum)
 
           await renderAndInitialise(page, 'character-count', {
-            nunjucksParams: examples['no maximum'],
+            nunjucksParams:
+              examples['when neither maxlength nor maxwords are set'],
             javascriptConfig: {
               maxlength: 10
             }
--- a/src/govuk/components/character-count/character-count.test.js
+++ b/src/govuk/components/character-count/character-count.test.js
@@ -31,7 +31,7 @@ describe('Character count', () => {
       await page.setJavaScriptEnabled(true)
     })
 
-    it('shows the fallback message', async () => {
+    it('shows the textarea description', async () => {
       await goToComponent(page, 'character-count')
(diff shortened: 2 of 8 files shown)


claireashworth:
I agree. If there's no maxlength or maxword defined is there a default limit or does it become unlimited? If there's a default then we can refer to that. Is having nothing defined done for a particular use case? If so, we could possibly reference that. Otherwise, I agree it would be better if there was no message.

claireashworth:
In that case, no message sounds like the better option.

36degrees:
I think an unfortunate side effect of the config work that we didn't foresee is that the Character Count JS can now receive config which isn't available to the template, which I think needs to be factored in here.

I'd agree with omitting the count message entirely if neither `maxwords` nor `maxlength` is set **in the macro**, but if the component is then initialised in JS with `new CharacterCount($element, { maxlength: 200 })` it feels like that should still work – there just wouldn't be an existing 'fallback message' in that context, and the JS might need to allow for that?

romaricpascal:
That hint element does two nice little things for the component (besides serving as an anchor for the JavaScript, but we could easily default to the textarea being that anchor if there's no element):

1. it reserves space on the page, which means that when JavaScript will kick in, there won't be a shift of content. I'm not sure how critical that is to maintain that, it would provide a better experience for the users if buttons don't shift around as JavaScript loads.
2. it serves as `aria-describedby` for the textarea, letting users know that the length of the input is limited when they focus the field. We don't have a number until JavaScript kicks in and when JavaScript does kick in, we have no string to format to inject it. My understanding is that we set the `aria-describedby` to that element to avoid double announcements (based on what I gathered from the issues linked on https://github.com/alphagov/govuk-frontend/pull/2577. We could use try passing down the fallback hint down to JavaScript for use as a translation:
 - if no `maxwords` or `maxlength` is set, pass it as `data-i18n.fallback-hint`, JS picks it up and injects it into the element. We don't pluralise in Nunjucks so we could not worry about it there as well (and it'll likely be a large number anyway)
 - if `maxwords` or `maxlength` are set, interpolate like we do now and don't pass it to JS as it'll be on screen.
 
Feels like quite a bit of impact though, I may be missing a simpler route 😬

querkmachine:
I might be misunderstanding here, sorry.

The primary purpose of the fallback hint is so that guidance on character limits is still available in situations where JavaScript has been disabled or failed to load/initialise correctly. In that regard, anything that requires passing it to or processing it in JavaScript is going to fail that requirement.

romaricpascal:
No worries, I might be not explaining it super well as well 😊 .

The element does indeed act as the fallback visible text before JavaScript loads (I should have listed that as the first thing it does). And for that role, I agree with you, it doesn't make sense to be relying on JavaScript to inject any content.

However, the element also serves to provide guidance on the character limit for assistive technology, through being linked to the `textarea` via `aria-describedby`. If it's empty and we don't fill it in when the component is configured in JavaScript, users of services chosing that road for configuring their `CharacterCount` won't have an announcement that there's a limit to the number of characters when they focus the field, only announcements when they type and the number of remaining/over limit characters get updated. 

That's for this scenario that I propose to pass the `fallbackHintText` (or `fallbackHintDefault` if none is set) as a `data-i18n....` attribute so that the JavaScript can set that accessible description, replacing `%{count}` with whatever was configured at initialisation.

The lack of `maxlength`/`maxwords` in the macro options would trigger the template to switch from injecting `fallbackHintText`/`fallbackHintDefault` inside the HTML and instead pass it as an i18n message to be set at component initialisation.

querkmachine:
Ahh, I think I understand now.

It's not the way we would want service teams to be doing this (we'd expect one of `maxlength` or `maxwords` to be set in the Nunjucks in virtually all circumstances, for progressive enhancement as much as accessibility), but on the offchance that a JS configuration is present and a Nunjucks one is not, then populating the fallback dynamically makes sense. 👍

src/govuk/components/character-count/character-count.test.js
```diff
@@ -433,6 +433,28 @@ describe('Character count', () => {
           const visibility = await page.$eval('.govuk-character-count__status', el => window.getComputedStyle(el).visibility)
           expect(visibility).toEqual('visible')
         })
+
+        it.only('injects the description of the textarea', async () => {
```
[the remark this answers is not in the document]
romaricpascal: Whoops, had forgotten to clean that up 😬 Cheers for spotting it, should be all tidy now (also removed the `console.log`ging that was set up in the test).

src/govuk/components/character-count/character-count.mjs
```diff
@@ -120,6 +120,13 @@ CharacterCount.prototype.init = function () {
   var $textarea = this.$textarea
   var $fallbackLimitMessage = document.getElementById($textarea.id + '-info')
 
+  // Inject a decription for the textarea if none is present already
+  // for when the component was rendered with no maxlength, maxwords
+  // nor custom fallbackHintText
+  if ($fallbackLimitMessage.textContent.match(/^\s*$/)) {
+    $fallbackLimitMessage.textContent = this.i18n.t('fallbackHint', { count: this.maxLength })
```
36degrees, line 127: If folks aren't using the Nunjucks macro, and don't have the corresponding changes to pass us the fallback hint, this could result in an error being thrown.

Given we're already in a 'best efforts' context trying to do something sensible, I think we should avoid that by guarding this – either by checking for `fallbackHint` being set or just wrapping this in a try catch?

src/govuk/components/character-count/character-count.test.js
```diff
@@ -433,6 +433,21 @@ describe('Character count', () => {
           const visibility = await page.$eval('.govuk-character-count__status', el => window.getComputedStyle(el).visibility)
           expect(visibility).toEqual('visible')
         })
+
+        it('injects the description of the textarea', async () => {
```
36degrees, line 437: I think this test description needs more context – outside of the context of this PR if this test started failing I'm not sure I'd understand what it was trying to test.

src/govuk/components/character-count/character-count.mjs
```diff
@@ -367,14 +374,11 @@ export default CharacterCount
 
 /**
  * @typedef {object} CharacterCountTranslations
- * @property {PluralisedTranslation} [charactersUnderLimit] - Characters under limit
+ * @property {import('../../i18n.mjs').PluralisedTranslation} [charactersUnderLimit] - Characters under limit
```
36degrees, line 377: Are these changes related?

src/govuk/components/character-count/template.test.js
```diff
@@ -260,4 +260,23 @@ describe('Character count', () => {
       })
     })
   })
+
+  describe('no maximum', () => {
```
36degrees, line 264: Again this could do with being clearer, maybe something like:

```suggestion
  describe('when neither maxlength nor maxwords are set', () => {
```

src/govuk/components/character-count/template.test.js
```diff
@@ -260,4 +260,23 @@ describe('Character count', () => {
...
+
+      // Fallback hint is passed as data attribute
+      const $component = $('[data-module]')
+      expect($component.attr('data-i18n.fallback-hint.other')).toEqual('No more than %{count} characters')
+
+      // A non-breaking space is set as the count message to reserve space
+      // and prevent a layout shift when the JavaScript component
+      // initialises and fills the element with the relevant text
+      const $countMessage = $('.govuk-character-count__message')
+      expect($countMessage.html()).toContain('&#xA0;')
```
36degrees, line 279: Worth rebasing against main – suspect this might fail once rebased on top of 7bbc33234ad431e88e526f1b698355022e68e74d.

src/govuk/components/character-count/template.njk
```diff
@@ -3,10 +3,14 @@
 {% from "../textarea/macro.njk" import govukTextarea %}
 {% from "../hint/macro.njk" import govukHint %}
 
+{%- set hasNoLimit = (not params.maxwords and not params.maxlength) %}
+{%- set fallbackHintText = params.fallbackHintText or 'You can enter up to %{count} ' + ('words' if params.maxwords else 'characters') %}
```
36degrees, line 7: If neither maxwords nor maxlength are passed this will always end up talking about characters, but `maxwords` could then be used when the component is initialised in JS.

You can see this if you go to https://govuk-frontend-pr-2915.herokuapp.com/components/character-count/no-maximum/preview and then run `new GOVUKFrontend.CharacterCount($('[data-module]'), { maxwords: 10 }).init()` in the console. The visible message will correctly display the number of words remaining, but if you inspect the fallback hint will still be talking about characters.

Unfortunately I'm not sure that's really fixable with the current approach without making things quite a lot messier… 🤔

src/govuk/components/character-count/template.njk
```diff
@@ -35,9 +39,14 @@
...
-  {%- set fallbackHintDefault = 'You can enter up to %{count} ' + ('words' if params.maxwords else 'characters') %}
+  {#
+    Use the fact that `html` has precedence over `text` to override content
+    when no limit is set and reserve vertical space to avoid layout shifting
+    by injecting a non-breakable space
+  #}
   {{ govukHint({
-    text: (params.fallbackHintText or fallbackHintDefault) | replace('%{count}', fallbackHintLength),
+    text: ((fallbackHintText) | replace('%{count}', fallbackHintLength) if not hasNoLimit),
+    html: ('&nbsp;' if hasNoLimit),
```
36degrees, line 49: This does feel a little messy. Given this is (hopefully) something of an edge case, I'm wondering if we just live with the layout shift as the hint gets injected.

Alternatively, what if we added a class to make the hint `visibility: hidden` if we don't know the count, then remove it once the JS fires?

src/govuk/components/character-count/character-count.mjs
romaricpascal: I don't think anything will be thrown there:
- `$fallbackLimitMessage` was already used beforehand, it's the reference for where to insert the visible and screen-reader containers
- the I18n class will not throw if there's no entry for `fallbackHint` in its content, it [returns the lookup key if no translation was provided](https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/i18n.mjs#L43-L61). 

This would set a default of `fallbackHint` as a description for assistive technologies. How about we provide our own default of `{other: ''}` to ensure we have a smooth experience on that side? That'll set an empty string if no alternative is provided by data attributes (or Nunjucks) and warn people in the console if they needed to use any of the other keys.

Thinking about it, we should probably warn in I18n if no key is found for the translation as well.

36degrees: Because we're treating `fallbackHint` as a pluralised string, we'll call `getPluralSuffix` before we get to that code, and  `getPluralSuffix` will throw if `fallbackHint.other` is not defined.

https://github.com/alphagov/govuk-frontend/blob/5e0746848368760a1a263057b4957369e922a82d/src/govuk/i18n.mjs#L162-L179

> How about we provide our own default of {other: ''} to ensure we have a smooth experience on that side?

That could work 👍🏻

src/govuk/components/character-count/character-count.mjs
romaricpascal: Good spot, that was a poor conflict resolution from rebasing `main`.

src/govuk/components/character-count/template.test.js
romaricpascal: `main` is already rebased. Because this is a non breaking space, Cheerio still encodes it (though into its hexa decimal form rather than the initiaal `&nbsp;`).

36degrees: According to GitHub this branch is currently 26 commits behind main, including 7bbc33234ad431e88e526f1b698355022e68e74d.

But happy that it might not be an issue anyway 👍🏻

src/govuk/components/character-count/template.njk
romaricpascal: Yeah, we can only do so much without knowing if it's characters or words that will be counted. In the first take I had set a very generic "You can enter a limited amount of content". Discussing with Beeps and Claire, it doesn't bring anything to the table so left it out.

I may have a tidy way to sort that out: leave the attribute out if no fallback text is set and no maxwords/maxlength are (but keep injecting the default as we do in the HTML from Nunjucks if we have a `maxlength` or `maxwords`). With a default of `''` in the JS, that'd mean no description for the user when they first focus the field, then an announcement as they reach the threshold and the count message starts displaying. This sounds like the tidiest scenario for users.

src/govuk/components/character-count/template.test.js
romaricpascal: Yup, I'd forgotten to push the rebasing, so got confused as to why things were still marked as OK on Github 🤦🏻

romaricpascal:
@36degrees All your comments should have been adressed in that last push 😄 Main changes, as discussed, were:
- moving to not passing a text to JavaScript unless people provide an explicit fallback hint, as we cannot guess whether it'll be characters or words
- providing a default translation for the fallbackHint's other (hadn't thought that this would throw indeed, good spot)
- vertical space is now reserved in CSS rather than in the markup. Still with extra content, but a zero-width space rather than non-breaking one (that allow us to have it at all times, which is simpler).

romaricpascal:
How would we feel about renaming that option "textareaDescriptionText" (`data-i18n.textarea-description`, `i18n.textareaDescription`) so we don't miss that it provides the textarea description (and just mention that this description will be visible to users until the JavaScript component is initialised)?

36degrees:
> How would we feel about renaming that option "textareaDescriptionText" (`data-i18n.textarea-description`, `i18n.textareaDescription`) so we don't miss that it provides the textarea description (and just mention that this description will be visible to users until the JavaScript component is initialised)?

Sounds good to me.

Changes look good too – but if we're planning to rename then I'll hold off approving until that's done.

36degrees (commented):
Looks like there are still quite a lot of places where we talk about a 'fallback message' or 'fallback hint' - suspect some of them are fine but think it might be worth updating at least `$fallbackLimitMessage` and the related comments in the component JS?

```
$ ag fallback src/govuk/components/character-count 
src/govuk/components/character-count/character-count.yaml
43:  description: Text describing the maximum number of characters you can enter, which is announced to screen readers. The text is displayed as a fallback if the character count JavaScript does not run. Depending on how you configure this component and what parameters you add, instances of `%{count}` are replaced by the value of `maxwords`. If not configured, it uses `maxlength`. By default, fallback text is provided in English.
89:  - name: with custom fallback text
90:    description: with no-js fallback text translated into Welsh
92:      name: custom-fallback
93:      id: custom-fallback
300:  - name: when neither maxlength/maxwords nor fallback hint are set

src/govuk/components/character-count/template.njk
14:  This means we can't compute a default fallback message (also used as the textarea accessible

src/govuk/components/character-count/character-count.test.js
34:    it('shows the fallback message', async () => {
58:      it('hides the fallback hint', async () => {
440:                'when neither maxlength/maxwords nor fallback hint are set'
609:        it('interpolates the fallback hint in data attributes with the maximum set in JavaScript', async () => {
610:          // This tests that any fallback hint provided through data-attributes

src/govuk/components/character-count/character-count.mjs
94:    // Read the fallback if necessary rather than have it set in the defaults
124:  var $fallbackLimitMessage = document.getElementById($textarea.id + '-info')
129:  if ($fallbackLimitMessage.textContent.match(/^\s*$/)) {
130:    $fallbackLimitMessage.textContent = this.i18n.t('textareaDescription', { count: this.maxLength })
133:  // Move the fallback count message to be immediately after the textarea
135:  $textarea.insertAdjacentElement('afterend', $fallbackLimitMessage)
143:  $fallbackLimitMessage.insertAdjacentElement('afterend', $screenReaderCountMessage)
146:  // fallback element for backwards compatibility as these may have been
149:  $visibleCountMessage.className = $fallbackLimitMessage.className
153:  $fallbackLimitMessage.insertAdjacentElement('afterend', $visibleCountMessage)
155:  // Hide the fallback limit message
156:  $fallbackLimitMessage.classList.add('govuk-visually-hidden')
386: * @property {PluralisedTranslation} [textareaDescription] - Fallback hint

src/govuk/components/character-count/template.test.js
228:  describe('with custom fallback text', () => {
229:    it('allows customisation of the fallback message', () => {
230:      const $ = render('character-count', examples['with custom fallback text'])
261:    describe('with fallback hint set', () => {
263:      // it needs to pass down any fallback hint to the JavaScript
265:      it('renders the fallback hint as a data attribute', () => {
268:        // Fallback hint is passed as data attribute
279:    describe('without fallback hint', () => {
280:      it('does not render a fallback hint data attribute', () => {
283:          examples['when neither maxlength/maxwords nor fallback hint are set']
```

romaricpascal:
Yup, that last commit just updated the full variable name to its new one. I'm starting a round of the rest of the code to check for further references to "fallback" 😄

romaricpascal:
@36degrees  I _think_ I've shifted all references of "fallback" to "textarea description" when talking about the Character Count (appart from when saying that this description is a fallback when JS doesn't run) 🎉 I'll update the documentation draft as well 😄

36degrees (approved):

PR #2919, opened 2022-10-14
Move SassDoc generation to `npm run build:sassdoc`

colinrotherham:
This PR is a bit of a tidy up to move SassDoc out of Gulp

It also gets things in place for JSDoc generation (coming in another PR)

Relates to https://github.com/alphagov/govuk-frontend/issues/2717


36degrees (approved):
I reckon we could probably remove the group names from the config as not sure we really care about the naming of the groups in the review app.

(In frontend docs, we have [some logic to convert the group names to 'human friendly' versions](https://github.com/alphagov/govuk-frontend-docs/blob/fda3edff418cb166094a42d136d8d59292fcdbda/sassdocs_helpers.rb) and to sort them)

But doesn't have to be done as part of this PR.

PR #2921, opened 2022-10-14
Fall back to 'other' if a plural form is missing

### LINKED ISSUE #2918: Character Count component throws error with missing plural text
## Description of the issue
The Character Count component is undergoing some localisation changes:

1. #2887
2. #2895

After the v4.3.1 → v4.4.0 upgrade, Character Counts might run as a different locale than intended.

This is a problem because we `throw new Error()` when the "new" locale's text can't be found:
https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/i18n.mjs#L49

```mjs
// Throw an error if this new key doesn't exist
if (!(lookupKey in this.translations)) {
  throw new Error('i18n: Plural form "' + pluralSuffix + '" is required for "' + this.locale + '" locale')
}
```

Teams may miss this issue because the error is only thrown when approaching the character limit.

1. Upgrade to (future release) `govuk-frontend@4.4.0`
2. Open the JavaScript console to watch for error messages
3. Type into a Character Count component on a Welsh `cy-GB` (or other locale) page
4. See errors are logged for missing plural forms "Many", "Few" and "Two"

**Note:** This can be triggered on the [**Translated** page `/examples/translated`](https://govuk-frontend-review.herokuapp.com/examples/translated) by setting `<html lang="cy-GB">` before the JavaScript runs. You can do this with `{% set htmlLang = 'cy-GB' %}` in the [**Translated** page template](https://github.com/alphagov/govuk-frontend/blob/main/app/views/examples/translated/index.njk)

Added a video demonstration from @36degrees:

https://user-images.githubusercontent.com/415517/195649191-453e440a-310b-4774-b89a-8b3ef2128cb8.mp4

### Actual behaviour
1. Localised text doesn't display for plural forms "Many", "Few" and "Two"
2. Error is thrown to the JavaScript console

### Expected behaviour
1. Localised text stays in English as it did in `govuk-frontend@4.3.1` pre-upgrade
2. Error is _logged_ (but not thrown) to the JavaScript console for missing plural forms

>Where the minimum "One" and "Other" text strings are available in the newly detected locale, could we avoid an English fallback despite the full suite of "Many", "Few" and "Two" plural forms are missing?

36degrees:
If a service currently has a character count inside an element with a non-English `lang` tag and tried to use the current implementation, the Character Count may throw an error and not update the count message when the user has a certain number of characters remaining.

This is because we are only providing default translations for the `one` and `other` plural forms (the only forms required in English).

But if the character count was in a `cy-GB` locale for example it would expect plural forms for `two`, `few` and `many`, and when the user got down to 6-2 characters remaining an error would be thrown and the count message would not be updated.

In order to preserve the existing behaviour, falling back to `other` is a reasonable thing to do, even if it may lead to incorrect plural forms being used in some cases.

Closes #2918.

### DIFF
--- a/src/govuk/i18n.mjs
+++ b/src/govuk/i18n.mjs
@@ -35,26 +35,7 @@ I18n.prototype.t = function (lookupKey, options) {
   // falsy, as this could legitimately be 0.
   if (options && typeof options.count !== 'undefined') {
     // Get the plural suffix
-    var pluralSuffix = this.getPluralSuffix(options.count)
-
-    // Use the correct plural form if provided
-    if (lookupKey + '.' + pluralSuffix in this.translations) {
-      lookupKey = lookupKey + '.' + pluralSuffix
-    // Fall back to `other` if the plural form is missing, but log a warning
-    // to the console
-    } else if (lookupKey + '.other' in this.translations) {
-      if (console && 'warn' in console) {
-        console.warn('i18n: Missing plural form ".' + pluralSuffix + '" for "' +
-          this.locale + '" locale. Falling back to ".other".')
-      }
-
-      lookupKey = lookupKey + '.other'
-    // If the required `other` plural form is missing, all we can do is error
-    } else {
-      throw new Error(
-        'i18n: Plural form ".other" is required for "' + this.locale + '" locale'
-      )
-    }
+    lookupKey = lookupKey + '.' + this.getPluralSuffix(lookupKey, options.count)
   }
 
   if (lookupKey in this.translations) {
@@ -149,10 +130,18 @@ I18n.prototype.hasIntlNumberFormatSupport = function () {
 /**
  * Get the appropriate suffix for the plural form.
  *
+ * Uses Intl.PluralRules (or our own fallback implementation) to get the
+ * 'preferred' form to use for the given count.
+ *
+ * Checks that a translation has been provided for that plural form – if it
+ * hasn't, it'll fall back to the 'other' plural form (unless that doesn't exist
+ * either, in which case an error will be thrown)
+ *
+ * @param {string} lookupKey - The lookup key of the string to use.
  * @param {number} count - Number used to determine which pluralisation to use.
  * @returns {PluralRule} The suffix associated with the correct pluralisation for this locale.
  */
-I18n.prototype.getPluralSuffix = function (count) {
+I18n.prototype.getPluralSuffix = function (lookupKey, count) {
   // Validate that the number is actually a number.
   //
   // Number(count) will turn anything that can't be converted to a Number type
@@ -160,13 +149,34 @@ I18n.prototype.getPluralSuffix = function (count) {
   count = Number(count)
   if (!isFinite(count)) { return 'other' }
 
+  var preferredForm
+
   // Check to verify that all the requirements for Intl.PluralRules are met.
   // If so, we can use that instead of our custom implementation. Otherwise,
   // use the hardcoded fallback.
   if (this.hasIntlPluralRulesSupport()) {
-    return new Intl.PluralRules(this.locale).select(count)
+    preferredForm = new Intl.PluralRules(this.locale).select(count)
   } else {
-    return this.selectPluralRuleFromFallback(count)
+    preferredForm = this.selectPluralRuleFromFallback(count)
+  }
+
+  // Use the correct plural form if provided
+  if (lookupKey + '.' + preferredForm in this.translations) {
+    return preferredForm
+  // Fall back to `other` if the plural form is missing, but log a warning
+  // to the console
+  } else if (lookupKey + '.other' in this.translations) {
+    if (console && 'warn' in console) {
+      console.warn('i18n: Missing plural form ".' + preferredForm + '" for "' +
+        this.locale + '" locale. Falling back to ".other".')
+    }
+
+    return 'other'
+  // If the required `other` plural form is missing, all we can do is error
+  } else {
+    throw new Error(
+      'i18n: Plural form ".other" is required for "' + this.locale + '" locale'
+    )
   }
 }
 
--- a/src/govuk/i18n.mjs
+++ b/src/govuk/i18n.mjs
@@ -157,7 +157,7 @@ I18n.prototype.getPluralSuffix = function (lookupKey, count) {
   if (this.hasIntlPluralRulesSupport()) {
     preferredForm = new Intl.PluralRules(this.locale).select(count)
   } else {
-    preferredForm = this.selectPluralRuleFromFallback(count)
+    preferredForm = this.selectPluralFormUsingFallbackRules(count)
   }
 
   // Use the correct plural form if provided
@@ -181,15 +181,15 @@ I18n.prototype.getPluralSuffix = function (lookupKey, count) {
 }
 
 /**
- * Get the plural rule using our fallback implementation
+ * Get the plural form using our fallback implementation
  *
  * This is split out into a separate function to make it easier to test the
  * fallback behaviour in an environment where Intl.PluralRules exists.
  *
  * @param {number} count - Number used to determine which pluralisation to use.
- * @returns {PluralRule} The suffix associated with the correct pluralisation for this locale.
+ * @returns {PluralRule} The pluralisation form for count in this locale.
  */
-I18n.prototype.selectPluralRuleFromFallback = function (count) {
+I18n.prototype.selectPluralFormUsingFallbackRules = function (count) {
   // Currently our custom code can only handle positive integers, so let's
   // make sure our number is one of those.
   count = Math.abs(Math.floor(count))
--- a/src/govuk/i18n.mjs
+++ b/src/govuk/i18n.mjs
@@ -37,12 +37,23 @@ I18n.prototype.t = function (lookupKey, options) {
     // Get the plural suffix
     var pluralSuffix = this.getPluralSuffix(options.count)
 
-    // Overwrite our existing lookupKey
-    lookupKey = lookupKey + '.' + pluralSuffix
+    // Use the correct plural form if provided
+    if (lookupKey + '.' + pluralSuffix in this.translations) {
+      lookupKey = lookupKey + '.' + pluralSuffix
+    // Fall back to `other` if the plural form is missing, but log a warning
+    // to the console
+    } else if (lookupKey + '.other' in this.translations) {
+      if (console && 'warn' in console) {
+        console.warn('i18n: Missing plural form ".' + pluralSuffix + '" for "' +
+          this.locale + '" locale. Falling back to ".other".')
+      }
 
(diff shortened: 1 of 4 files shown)


36degrees:
I am not 100% sure about the warning – I think there are cases where users could omit plural forms and the translations would be fine.

Just because a language might use a different plural form, doesn't mean it will.

To give a slightly silly example, this would log a warning, but the localisation is perfectly correct:

```js
const i18n = new I18n({
    'test.other': 'Ollie has %{count} sheep'
}

i18n.t('test', { count: 1 }) // Ollie has 1 sheep
i18n.t('test', { count: 2 }) // Ollie has 2 sheep
```

To silence the warning, they'd need to pass `test.one` even though it works exactly the same as `other`. Should we be trying to enforce passing all plural forms in code when they might not all be needed?

src/govuk/i18n.mjs
```diff
@@ -37,12 +37,23 @@ I18n.prototype.t = function (lookupKey, options) {
...
-    // Overwrite our existing lookupKey
-    lookupKey = lookupKey + '.' + pluralSuffix
+    // Use the correct plural form if provided
+    if (lookupKey + '.' + pluralSuffix in this.translations) {
+      lookupKey = lookupKey + '.' + pluralSuffix
+    // Fall back to `other` if the plural form is missing, but log a warning
+    // to the console
+    } else if (lookupKey + '.other' in this.translations) {
+      if (console && 'warn' in console) {
+        console.warn('i18n: Missing plural form ".' + pluralSuffix + '" for "' +
```
querkmachine, line 47: The message formatting seems a bit inconsistent as it refers to the missing key as ".many" but then says it's falling back to the \`other\` key. I'd expect the missing key to probably also be written as \`many\`. 

Otherwise this all looks good to me. Seems to do the job.

querkmachine (approved):

src/govuk/i18n.mjs
```diff
@@ -37,12 +37,23 @@ I18n.prototype.t = function (lookupKey, options) {
     // Get the plural suffix
     var pluralSuffix = this.getPluralSuffix(options.count)
 
-    // Overwrite our existing lookupKey
-    lookupKey = lookupKey + '.' + pluralSuffix
+    // Use the correct plural form if provided
+    if (lookupKey + '.' + pluralSuffix in this.translations) {
```
colinrotherham, line 41: Nice work here, lots of plural wrangling

Since the new code is all "get the plural suffix" related, could we move it to `this.getPluralSuffix()`?

```js
var pluralSuffix = this.getPluralSuffix(options.count)
// '.other'
```

src/govuk/i18n.mjs
36degrees: Good catch! 👀

src/govuk/i18n.mjs
```diff
@@ -37,12 +37,23 @@ I18n.prototype.t = function (lookupKey, options) {
...
+      lookupKey = lookupKey + '.' + pluralSuffix
+    // Fall back to `other` if the plural form is missing, but log a warning
+    // to the console
+    } else if (lookupKey + '.other' in this.translations) {
+      if (console && 'warn' in console) {
```
colinrotherham, line 46: With the Developer Tools open does IE8 have `console.warn()` or only `console.log()`?

Sorry can't remember!

querkmachine:
> Should we be trying to enforce passing all plural forms in code when they might not actually be needed?

I'm gonna say yes, if just for consistency and error checking's sakes. In that example, it's more coincidental that the `one` and `other` forms of 'sheep' are the same than by the design of the language. 

A reverse argument might be that we shouldn't require `other` in all situations because Chinese, Malaysian, Indonesian, etc. don't have plural forms at all and all nouns are singular. Russian is similarly weird, with the `other` form only used for decimal numbers, with all whole numbers using one of the other plural forms.

At least by pushing a warning out for a missing form, we can assist developers (who may not be speakers of the language at hand) in knowing if a required string is likely to be missing or malformed.

src/govuk/i18n.mjs
```diff
@@ -37,12 +37,23 @@ I18n.prototype.t = function (lookupKey, options) {
...
+      }
 
-    // Throw an error if this new key doesn't exist
-    if (!(lookupKey in this.translations)) {
-      throw new Error('i18n: Plural form ".' + pluralSuffix + '" is required for "' + this.locale + '" locale')
+      lookupKey = lookupKey + '.other'
+    // If the required `other` plural form is missing, all we can do is error
+    } else {
+      throw new Error(
+        'i18n: Plural form ".other" is required for "' + this.locale + '" locale'
```
colinrotherham, line 55: Worth trying to share these messages between the `new Error()` and `console.warn()`?

```js
var pluralSuffixError = 'i18n: Plural form ".' + pluralSuffix + '" is required for "' + this.locale + '" locale'
var pluralSuffixWarn = pluralSuffixError + '. Falling back to `other`.'
```

```js
console.warn(pluralSuffixWarn)
```

```js
throw new Error(pluralSuffixError)
```

src/govuk/i18n.unit.test.mjs
```diff
@@ -151,13 +151,37 @@ describe('I18n', () => {
...
       const i18n = new I18n({
         'test.other': 'testing testing'
       }, {
         locale: 'en'
       })
-      expect(() => { i18n.t('test', { count: 1 }) }).toThrowError('i18n: Plural form ".one" is required for "en" locale')
+      expect(i18n.t('test', { count: 1 })).toBe('testing testing')
+    })
+
+    it('logs a console warning when falling back to `other`', () => {
```
colinrotherham, line 163: I'm seeing `.many`, `.few` and `.two` logged (not just `.one`)

Shall we do one of those fancy `it.each([])` and make sure each plural form has a test?

Looks good

colinrotherham (approved):
Cheers @36degrees looks good

Sheep example aside (deer, fish, moose… Emoji) it's all working. Perhaps good to consider the page weight increase having to duplicate lots of text for these cases?

Added some nice-to-have (but not blocking) comments

src/govuk/i18n.mjs
colinrotherham: Just means that `this.getPluralSuffix()` is always correct versus `this.getPluralSuffixThatMightNotBeUsed()`

src/govuk/i18n.unit.test.mjs
colinrotherham: This came up on Slack, so bringing over the points here

1. Do we suppress `.two`, `.few`, `.many` warnings, for all locales?
2. Do we suppress `.two`, `.few`, `.many` warnings, for English only?
3. Treating all locales equally, is it correct to need duplicate strings by design?

```js
var TRANSLATIONS_DEFAULT = {
  // Characters
  charactersUnderLimit: {
    one: 'You have %{count} character remaining',
    two: 'You have %{count} characters remaining',
    few: 'You have %{count} characters remaining',
    many: 'You have %{count} characters remaining',
    other: 'You have %{count} characters remaining'
  },
  charactersAtLimit: 'You have 0 characters remaining',
  charactersOverLimit: {
    one: 'You have %{count} character too many',
    two: 'You have %{count} characters too many',
    few: 'You have %{count} characters too many',
    many: 'You have %{count} characters too many',
    other: 'You have %{count} characters too many'
  },
  // Words
  wordsUnderLimit: {
    one: 'You have %{count} word remaining',
    two: 'You have %{count} words remaining',
    few: 'You have %{count} words remaining',
    many: 'You have %{count} words remaining',
    other: 'You have %{count} words remaining'
  },
  wordsAtLimit: 'You have 0 words remaining',
  wordsOverLimit: {
    one: 'You have %{count} word too many',
    two: 'You have %{count} words too many',
    few: 'You have %{count} words too many',
    many: 'You have %{count} words too many',
    other: 'You have %{count} words too many'
  }
}
```

>To silence the warning, they'd need to pass test.one even though it works exactly the same as other. Should we be trying to enforce passing all plural forms in code when they might not actually be needed?

👆 Above would be the Character Count defaults if we needed all strings

36degrees: Different plural forms are required depending on which locale the Character Count is in.

When in the default 'en' locale, only 'one' and 'other' are required.

So, I think the answers are:

1. We should warn when a plural form is missing that we try to use based on the locale, so 'it depends on the locale'
2. There shouldn't be any cases where we warn for a character count in the `en` locale, because we have all of the plural forms we need to localise in English ('one' and 'other')
3. I think that's what I'm asking with my question about the sheep. But I don't think we'd add plural forms for anything other than 'one' and 'other' to the defaults as the defaults are only designed to be used in an `en` locale – we'd expect if the locale is something other than `en` that we'd have a different set of translations passed with the correct plural forms for the locale. This PR is trying to address the edge case where that doesn't happen.

colinrotherham: @36degrees Perfect, thanks for the answers

>Shall we do one of those fancy it.each([]) and make sure each plural form has a test?

Do you want to add tests to confirm?

1. English: Warnings **do not** get logged for missing `.two`, `.few`, `.many`
2. Others: Warnings **do** get logged for missing `.two`, `.few`, `.many`

We've got this one already, but won't harm to document this decision about the others in test form

```js
expect(consoleWarn).toHaveBeenCalledWith(
  'i18n: Missing plural form ".one" for "en" locale. Falling back to ".other".'
)
```

colinrotherham: Yeah, makes sense

I like to think if a less experienced developer touches these lines in future, and they break something, the tests should be there to explain exactly why the change broke something—gives them a safety net.

Also, if we fix a Welsh bug, add a Welsh test case 😆

If it feels better in the Puppeteer tests, further away from I18n, can do that?

src/govuk/i18n.mjs
36degrees: Unsure – what's the thinking? I'm personally not sure it really matters if the warnings don't appear in IE8.

colinrotherham: Yep this! Just checking if it matters

querkmachine:
As a stray thought (sorry!), in my mind, the functionality to fall back to `other` is only present to avoid a breaking change in existing services where they upgrade Frontend, and have a situation where the page is set in one `lang` but the component has not been translated yet. 

However, falling back to `other` is a 'best guess' assumption and is not reliably true for all languages, nor is this something we actually want service developers to use or rely on as an intended feature. (Ideally, all plural forms needed should be explicitly set.)

Would it be a good idea to have a follow up card to remove the fallback in a future breaking release?

src/govuk/i18n.unit.test.mjs
36degrees: I've added a failing test to the Character Count component in c757025267f9df0f4d6dfcfb86f1943c7a30afb9.

colinrotherham (approved):

PR #2922, opened 2022-10-14
Improve test coverage of the I18n class

36degrees:
Because `Intl.PluralRules` exists in the test environment, we didn't have any coverage for these lines of code:

https://github.com/alphagov/govuk-frontend/blob/7fac761795d528ccec71f940af69b1c33342bbb1/src/govuk/i18n.mjs#L168-L182

This refactors the I18n class to split out and test the 'fallback' logic that: 

- determines which 'rules' to use to determine the plural rule (eurgh, too many things we're calling rules) based on locale
- uses the rules to determine the plural rule…

According to Jest, we now only have 2 lines of code that are not tested:

Line **158** (we could mock out `hasIntlPluralRulesSupport` to return false if we think it's needed?)

https://github.com/alphagov/govuk-frontend/blob/7f73c53c7dd94ebd8ed5c4e1f440dd25ca1aa9a3/src/govuk/i18n.mjs#L158

Line **302** (As far as I can tell this code is unreachable):

https://github.com/alphagov/govuk-frontend/blob/7f73c53c7dd94ebd8ed5c4e1f440dd25ca1aa9a3/src/govuk/i18n.mjs#L294-L303

I've also moved the `I18n.pluralRulesMap` and `I18n.pluralRules` from the prototype as they feel more like 'class constants' than properties that need to exist on every object.

Suggest reviewing commit by commit as I've moved some stuff around.

### DIFF
--- a/src/govuk/i18n.mjs
+++ b/src/govuk/i18n.mjs
@@ -138,17 +138,10 @@ I18n.prototype.hasIntlNumberFormatSupport = function () {
 /**
  * Get the appropriate suffix for the plural form.
  *
- * The locale may include a regional indicator (such as en-GB), but we don't
- * usually care about this part, as pluralisation rules are usually the same
- * regardless of region. There are exceptions, however, (e.g. Portuguese) so
- * this searches by both the full and shortened locale codes, just to be sure.
- *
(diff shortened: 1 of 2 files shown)


colinrotherham (approved):
Work of art 🙌

PR #2923, opened 2022-10-14
Enable unit test code coverage checks

colinrotherham:
Inspired by https://github.com/alphagov/govuk-frontend/pull/2922 this PR adds code coverage checks

**Note:** Coverage is turned off by default as it's quite noisy during `--watch` runs


jest.config.js
```diff
@@ -3,6 +3,14 @@ const config = {
 }
 
 module.exports = {
+  coverageThreshold: {
+    global: {
+      branches: 60,
+      functions: 60,
+      lines: 60,
+      statements: 60
+    }
```
romaricpascal, line 12: Are these numbers based on the current state of our project?

romaricpascal (commented):
I love some test coverage info 😍 

With the current set up, will the tests fail if we don't meet the coverage threshold or is it just informational? If they fail, I'm a bit worried we're going to shoot ourselves in the foot a bit with files like the character count: most of its tests are end-to-end tests, with only a few unit tests to avoid spinning puppeteer. This means unit test coverage will be pretty low for that file (even though the component's test coverage is actually solid thanks to the end-to-end tests).

Which brings nicely to my second question 😆: would we have a way to have some coverage info for the end to end tests? Guess that may be tricky with Puppeteer in the mix and us loading the built files in our example 😬

colinrotherham:
Thanks @romaricpascal

Yeah I've set a 60% "fail the build" threshold based on our current coverage figures

```console
File      | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s 
----------|---------|----------|---------|---------|-------------------
All files |    67.6 |    75.77 |   63.93 |   65.95 |                   
```

Want me to change anything?

For Puppeteer, we could grab the developer tools figures via `page.coverage.startJSCoverage()`?

Code coverage for Design System website

romaricpascal:
Sorry, I should have prefixed that these were more questions for understanding where we were going/feasability than things to change 😄 I think that'd be good to start with what you propose and adjust from there as we go, but I'll everyone else voice their thoughts rather than stamping the "Approve" button 😄 It's probably worth discussing how we want to use that metric as well, as there's a risk of chasing coverage for chasing coverage rather than improving shipping confidence.

For Puppeteer, I didn't know we could get coverage stats from it as well, that's neat! I think we'll want something more granular than on the whole `application` or even `govuk-frontend` file, though, and highlight if a specific component is short on tests. Sounds like it may not be straightforward, so that may not be for this PR.  

One thing that we may want is surfacing the results of the coverage on the PRs. I'd seen the numbers hunting them in the "Checks" page, but that requires a bit of digging. Is there anything you're aware of/may have used in the past for that?

colinrotherham:
@romaricpascal I've turned off the minimum thresholds so this isn't blocking

Our figures our quite low once all files are included (not just files that are tested)

```console
File      | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s 
----------|---------|----------|---------|---------|-------------------
All files |   19.95 |    37.55 |   17.68 |   18.09 |                   
```

Shall we consider this a starting point and in future:

1. Add minimum coverage threshold
2. Add Puppeteer coverage reporting
3. Add GitHub coverage reporting

colinrotherham:
I've attached the coverage report to the GitHub workflow too. Appreciate it's not nice and colourful, but it's a handy way to dig into the code coverage on someone else's PR when you're helping out 😊 

Code coverage report zip from GitHub workflow

jest.config.js
```diff
@@ -72,5 +79,6 @@ module.exports = {
       globalSetup: './config/jest/browser/open.mjs',
       globalTeardown: './config/jest/browser/close.mjs'
     }
-  ]
+  ],
+  reporters: ['default', 'github-actions']
```
36degrees, line 83: If I'm reading this right, this'll surface test failures as annotations on PRs? 🙌🏻 

If so, that sounds like an awesome change but it's 'snuck in' – it's not related to the other code coverage changes, and it's not mentioned in the PR description or the commit message.

Can we split it out into a separate commit and make sure it's mentioned in the PR description?

36degrees (commented):
Excited to see this! 🙌🏻

jest.config.js
colinrotherham: @36degrees Done 👍

36degrees (approved):

PR #2925, opened 2022-10-17
Remove history rewriting and new tab functionality from HtP button

querkmachine:
Removes the functionality to open a page in a new tab and to rewrite the current page's title in history, so we have a simpler base from which to produce new spikes from.

Closes #2384.


owenatgov (approved):
Looking sharp to me!

PR #2930, opened 2022-10-19
Move custom fallback text test

36degrees:
This is testing the output of the Nunjucks macro rather than the JavaScript, so there's no need to use Puppeteer and it makes sense for it to live in `template.test.js`.


colinrotherham (approved):

romaricpascal (approved):
🚀 🚀 🚀 🙌🏻

PR #2931, opened 2022-10-19
Summary card

owenatgov:
Add the summary card visual enhancement to the summary list component.

This is a draft PR as a reference for the summary card code whilst he guidance is being hashed out.


src/govuk/components/summary-list/template.njk
```diff
@@ -7,38 +7,76 @@
...
+    <div class="govuk-summary-card__title-wrapper">
+      {%- if params.title -%}
+        <h{{ headingLevel }} class="govuk-summary-card__title {%- if params.title.classes %} {{ params.title.classes }} {%- endif %}">
+          {{- params.title.html | safe if params.title.html else params.title.text -}}
+        </h{{ headingLevel }}>
+      {%- endif -%}
+
+      {%- if params.actions.items.length -%}
+        <ul class="govuk-summary-card__actions {%- if params.actions.classes %} {{ params.actions.classes }} {%- endif %}">
+          {%- for action in params.actions.items -%}
```
querkmachine, line 23: Real nitpicky thing, but action links in the 'content' part of the card (≈ line 57) have some funky logic where it only renders within a list element if there is more than one link present, otherwise the link is rendered standalone.

For consistency, would we want to reproduce that logic here too?

src/govuk/components/summary-list/template.njk
```diff
@@ -7,38 +7,76 @@
   </a>
 {% endmacro -%}
 
+{%- macro _summaryCard(params) %}
+  {%- set headingLevel = params.title.headingLevel if params.title.headingLevel else 2 -%}
+
+  <div class="govuk-summary-card {%- if params.classes %} {{ params.classes }}{% endif %}"{% for attribute, value in params.attributes %} {{attribute}}="{{value}}"{% endfor %}>
+    <div class="govuk-summary-card__title-wrapper">
+      {%- if params.title -%}
+        <h{{ headingLevel }} class="govuk-summary-card__title {%- if params.title.classes %} {{ params.title.classes }} {%- endif %}">
```
querkmachine, line 16:
```suggestion
        <h{{ headingLevel }} class="govuk-summary-card__title {%- if params.title.classes %} {{ params.title.classes }}{% endif %}">
```

For code consistency with how inline Nunjucks variables are written elsewhere.

src/govuk/components/summary-list/template.njk
```diff
@@ -7,38 +7,76 @@
...
+          {{- params.title.html | safe if params.title.html else params.title.text -}}
+        </h{{ headingLevel }}>
+      {%- endif -%}
+
+      {%- if params.actions.items.length -%}
+        <ul class="govuk-summary-card__actions {%- if params.actions.classes %} {{ params.actions.classes }} {%- endif %}">
```
querkmachine, line 22:
```suggestion
        <ul class="govuk-summary-card__actions {%- if params.actions.classes %} {{ params.actions.classes }}{% endif %}">
```

Same as above.

querkmachine (approved):
Barring my nitpicks, this is looking all sorts of excellent to me!

owenatgov:
@querkmachine Points addressed. Cheers for the review and approval 💪🏻

I'm gonna take that approve as given for when we're ready to release but if you spot anything else feel free to comment.

src/govuk/components/summary-list/_index.scss
```diff
@@ -166,4 +166,72 @@
...
+  }
+
+  .govuk-summary-card__action:last-child {
+    margin-right: 0;
+    padding-right: 0;
+  }
+
+  .govuk-summary-card__action:not(:last-child) {
+    border-right: 1px solid $govuk-border-colour;
+  }
```
36degrees, line 219: This is really nitpicky, and happy for this to be merged as is, but I find it a little bit weird that we've got two different ways of styling the last child here which means we need both `:last-child` and `:not(:last-child)`…

- we set margin and padding on all actions and then _unset_ them on the last child
- we add the border to everything _except_ the last child

Wondering if it's clearer if we just do one of them?

```suggestion
  .govuk-summary-card__action {
    display: inline;
    margin-right: govuk-spacing(2);
    padding-right: govuk-spacing(2);
    border-right: 1px solid $govuk-border-colour;
  }

  .govuk-summary-card__action:last-child {
    margin-right: 0;
    padding-right: 0;
    border-right: none;
  }
```

(I'm suggesting this way around because otherwise there'll be no spacing between actions in IE8 which doesn't support `:not()`)

src/govuk/components/summary-list/_index.scss
```diff
@@ -166,4 +166,72 @@
...
+  .govuk-summary-card {
+    @include govuk-responsive-margin(6, "bottom");
+    border: 1px solid $govuk-border-colour;
+  }
+
+  .govuk-summary-card__title-wrapper {
+    display: flex;
+    justify-content: space-between;
+    flex-wrap: wrap;
+    padding: govuk-spacing(3) govuk-spacing(4);
```
36degrees, line 185: @christopherthomasdesign do we want to align the left-edge with the summary card contents on mobile?

211529187-364d3614-dd3b-4824-858b-746825c16b24

src/govuk/components/summary-list/_index.scss
```diff
@@ -166,4 +166,72 @@
...
+  // because it's only interacted with via the summary list and it avoids awkwardness
+  // with trying to document a "private" component that's controlled via the
+  // component it's wrapping
+  .govuk-summary-card {
+    @include govuk-responsive-margin(6, "bottom");
+    border: 1px solid $govuk-border-colour;
+  }
+
+  .govuk-summary-card__title-wrapper {
+    display: flex;
```
36degrees, line 182: I like how we're using flexbox here, but it does mean that you can end up with the summary list actions floated to the right with the card actions on the left. I personally don't see this is an issue but wanted to flag it to make sure we're happy:

Screenshot 2023-01-11 at 09 14 44

src/govuk/components/summary-list/_index.scss
```diff
@@ -166,4 +166,72 @@
       }
     }
   }
+
+  // Additional block for the summary card
+  //
+  // We ship this with the summary list instead of making it it's own component
+  // because it's only interacted with via the summary list and it avoids awkwardness
+  // with trying to document a "private" component that's controlled via the
+  // component it's wrapping
```
36degrees, line 175: Is this important enough to include as a comment here? It's useful context, but not going to be relevant for most people. Wondering if it'd be better off living in the the commit message.

src/govuk/components/summary-list/_index.scss
```diff
@@ -166,4 +166,72 @@
...
+  }
+
+  .govuk-summary-card__title-wrapper {
+    display: flex;
+    justify-content: space-between;
+    flex-wrap: wrap;
+    padding: govuk-spacing(3) govuk-spacing(4);
+
+    // Ensures the card header appears separate to the summary list in forced colours mode
+    border-bottom: 1px solid transparent;
```
36degrees, line 188: 😍

src/govuk/components/_all.scss
```diff
@@ -29,6 +28,7 @@
 @import "radios/index";
 @import "select/index";
 @import "skip-link/index";
+@import "summary-list/index";
```
36degrees, line 31: Thanks for tidying this up! I was surprised this was out of order and has been for so long – I _reckon_ this is because this component was originally called 'data list' – when we renamed it during development I suspect we didn't change its order in this file.

36degrees (commented):
Love it, great work 👏🏻 

Have noticed a couple of minor things. If possible, it'd also be nice to add a new 'full page example' which shows how these are used.

src/govuk/components/summary-list/_index.scss
christopherthomasdesign: Yes we definitely do @36degrees, this is an oversight on my part!

src/govuk/components/summary-list/_index.scss
christopherthomasdesign: I suppose ideally it would all break at the same point so that the actions were placed consistently. Is that possible given the flexbox thingy above breaks in relation to the width of the content rather than any consistent measure?

src/govuk/components/summary-list/_index.scss
CharlotteDowns: I agree with this

src/govuk/components/summary-list/_index.scss
CharlotteDowns: If there's anything that can make them break at the same point that would be really neat but I don't feel it blocks publication if not :)

CharlotteDowns:
I'm not sure whether the touch target of the 'withdraw' and 'delete' actions is too small in the inspector preview of a Samsung Galaxy S8+ and Galaxy Fold. 

Samsung Galaxy S8+
Galaxy Fold inspector screenshot

src/govuk/components/summary-list/_index.scss
owenatgov: I'm not sure how this got here 🤔 Probs leftover prototype code, cheers for flagging.

src/govuk/components/summary-list/_index.scss
owenatgov: I'm trying to figure out where this gap is coming from but I'm a bit stumped. May confer with the channel later today.

src/govuk/components/summary-list/_index.scss
owenatgov: I'll move to the commit message 👍🏻

src/govuk/components/summary-list/_index.scss
36degrees: It's a difference in padding – `govuk-summary-card__title-wrapper` has `govuk-spacing(4)` for horizontal spacing at all breakpoints, but `govuk-summary-card__content` has `govuk-spacing(3)` on mobile and `govuk-spacing(4)` on tablet and above.

So `govuk-summary-card__title-wrapper` needs to be updated to match `govuk-summary-card__content`:

```suggestion
    padding: govuk-spacing(3);
    
    @include govuk-media-query($from: "tablet") {
      padding: govuk-spacing(3) govuk-spacing(4);
    }
```

owenatgov: Doh! Cheers for point that out. Gonna fix.

owenatgov:
Following a chat on slack @CharlotteDowns's point on touch size has been addressed. I'll update the PR description in a jiffy.

edwardhorsford:
A minor thing: is it intentional that the card actions are in bold?

owenatgov:
@edwardhorsford Yes! This is an intentional accessibility enhancement that we're using alongside content guidance to differentiate the action links associated with the card from the action links in the summary list as a way to communicate visually that the card actins will apply to the whole list.

CharlotteDowns:
The vertical spacing of the actions on mobile devices with smaller viewports looks great!

src/govuk/components/summary-list/summary-list.yaml
```diff
@@ -787,3 +954,52 @@ examples:
...
+              href: '#'
+      rows:
+        - key:
+            text: Name
+          value:
+            text: Firstname Lastname
+        - key:
+            text: Date of birth
+          value:
+            text: 13/08/1980
```
colinrotherham, line 1005: Sorry @owenatgov mind sticking a trailing new line here?

Saves the next person accidentally inserting a new line on save. If you fancy adding a plugin for [EditorConfig](https://editorconfig.org/#download) the project `.editorconfig` config will fix it for you

src/govuk/components/summary-list/template.njk
```diff
@@ -7,38 +7,82 @@
...
+{%- set summaryList -%}
+  <dl class="govuk-summary-list {%- if params.classes %} {{ params.classes }}{% endif %}"{% for attribute, value in params.attributes %} {{attribute}}="{{value}}"{% endfor %}>
+    {% for row in params.rows %}
+      {% if row %}
+        <div class="govuk-summary-list__row {%- if anyRowHasActions and not row.actions.items %} govuk-summary-list__row--no-actions{% endif %} {%- if row.classes %} {{ row.classes }}{% endif %}">
+          <dt class="govuk-summary-list__key {%- if row.key.classes %} {{ row.key.classes }}{% endif %}">
+            {{ row.key.html | safe if row.key.html else row.key.text }}
+          </dt>
+          <dd class="govuk-summary-list__value {%- if row.value.classes %} {{ row.value.classes }}{% endif %}">
+            {{ row.value.html | indent(8) | trim | safe if row.value.html else row.value.text }}
```
colinrotherham, line 59: @owenatgov Presume the `indent(8)` need bumping because we've indented another level?

But other lines don't have `indent` so worth a check

src/govuk/components/summary-list/template.njk
```diff
@@ -7,38 +7,82 @@
...
-    {% endif %}
-  {% endfor %}
-</dl>
+          {% if row.actions.items.length %}
+            <dd class="govuk-summary-list__actions {%- if row.actions.classes %} {{ row.actions.classes }}{% endif %}">
+              {% if row.actions.items.length == 1 %}
+                {{ _actionLink(row.actions.items[0]) | indent(12) | trim }}
+              {% else %}
+                <ul class="govuk-summary-list__actions-list">
+                  {%- for action in row.actions.items -%}
```
colinrotherham, line 67: @owenatgov Just a quick check, what happens if there are no actions—an empty `<ul>`?

querkmachine: In that case it wouldn't meet the `if` on L21 (length = 0 = false) and not render any part of the action list.

colinrotherham: Ahh @querkmachine, my eyes lied to me (false alarm). Thanks for checking

src/govuk/components/summary-list/template.njk
owenatgov: I completely missed this as something we were doing in the component and spent an interesting hour understanding what it was doing 😂  For anyone who's interested, [here's why we're doing this](https://github.com/alphagov/govuk-frontend/pull/1131/files#r247459722). My rough understanding is that this is to make the source easier to read.

I don't see any reason not to bump this for tidiness.

owenatgov: I'm going to double back as I'm debating the value of doing this. We'd need to add some logic to check if the `card` API was populated and add (or not add) the `indent` and it's value based on that. I worry that this is going to add a lot of chuff to the template with not that much reward. What do you think @colinrotherham?

colinrotherham: Might have to defer to @36degrees wisdom on this one

We do use `indent(n)` quite a lot, see this [comment on the **Button** component](https://github.com/alphagov/govuk-frontend/commit/d8fcb683b327bcd6387f5344994cb9fcbbd8349c#diff-b56c3509e935950b17bc531de661529e2025c837e4e40eceffc165aa4cebd9c8R51):

```njk
{# Indentation is intentional to output HTML nicely #}
{{- iconHtml | safe | trim | indent(2, true) if iconHtml -}}
```

Think I'd want all calls to `indent(n)` to either be correct or removed 😊

Bear in mind we might not see it in action as the first line (of multi-line HTML) isn't indented
https://mozilla.github.io/nunjucks/templating.html#indent

owenatgov: I'm doubling back on my double back! I've figured out the intent of these indents: they're for when CSS isn't available. Will action.

a184studio:
> @edwardhorsford Yes! This is an intentional accessibility enhancement that we're using alongside content guidance to differentiate the action links associated with the card from the action links in the summary list as a way to communicate visually that the card actins will apply to the whole list.

@owenatgov Building on Ed's question. 

Is there some inconstancies, I get the principles if you had something like this to have bold in the actions for when you also have change links in the body of the card. 
 
Screen Shot 2023-03-17 at 14 35 19

However when you just have 
Screen Shot 2023-03-17 at 14 36 49
Those actions arn't fighting with any other action. 

Style wise would match this. weight wise
Screen Shot 2023-03-17 at 14 38 11

Unless this is a higher visual contrast due to the grey background?

querkmachine:
@a184studio For my own idle musing: I understand the point that the header actions need not be in bold if there are no child actions, however I personally think that the header actions should stay in bold for consistency's sake.

For example, there might be multiple summary cards on one page, some with child actions, some without. It would be inconsistent, and potentially confusing to users, for some of cards to have header action links in bold and some not.

Consistency is also external. As we present GOV.UK as one website (although, architecturally, it isn't), we want summary cards to be implemented consistently and appear with the same visual styling across all services and all applications. That is easiest achieved here by using one style in all situations, rather than sometimes changing it based on a bevy of other conditions. 

Whilst the Design System isn't a rulebook and you are free to deviate if needed, we would ideally want a user-centred, user-researched justification for making those changes before folding anything back into our code or guidance. Does having the header links in bold actually cause confusion or present a barrier to a user in practice? 

As a note, because this PR has already been closed, further discussion should probably be in the backlog issue: https://github.com/alphagov/govuk-design-system-backlog/issues/210

Issue #2892, opened 2022-10-03
Prevent XSS from translation strings


romaricpascal:
## A bit of context

Components supporting internationalisation currently inject translated content using `innerHTML`. This is necessary to allow visually hidden text to be injected, like in the accordion "Show/Hide this section" messages. This content can reach them two ways, through: 
- their JavaScript configuration
- data-attributes on the element
The later also acts as a bridge between options passed to the Nunjucks macro and the JavaScript to make it easy to pass the translations.

While `innerHTML` won't [execute `<script>` tags](https://developer.mozilla.org/en-US/docs/Web/API/Element/innerHTML#security_considerations), there's a wide range of [malicious string](https://cheatsheetseries.owasp.org/cheatsheets/XSS_Filter_Evasion_Cheat_Sheet.html) that could be used for Cross Site scripting should they end up as is in the HTML.
With our current components, translations strings should not be user-generated, which limits the risks. However, this is not to say they'll always be, and that there won't be someone using our components with user-generated content for a translation string at some point in the future.

This calls for some thoughts and there are a few routes we could take (pooled from the Dev catch-up)

## 1. Leave it to the services

Do nothing and trust the services to pass sanitised strings for us to inject.
Maybe at some point in the future, introduce some sanitisation to limit HTML that can be injected

**Advantages**

- No effort on our part

**Impact**

- None

**Concerns**

- There's no way to guarantee that we'll receive sanitised HTML in all scenarios, and ultimately, we're the ones injecting it into the DOM
- Sanitising later means introducing breaking changes for people that were happily injecting HTML to solve their issues

## 2. Sanitise HTML before injection

Before injecting the content on the page, sanitise the strings to only allow a specific set of elements and attributes.

**Advantages**

- Extra safety as we only allow what we consider safe
- Sets the path for future attribute reads to be sanitised

**Impact**

- Extra work to either develop it or pick a library (npm has plenty, of various sizes). May be done in a central way [as part of the string normalisation](https://github.com/alphagov/govuk-frontend/blob/062946035cf9d105b6ce045f4866927605d42675/src/govuk/common.mjs#L147-L169)

**Cons**

- We'll need to allow the right tags/attributes to be safe, but not restrict too much that we don't meet users needs.
- Some more JS shipped to users (this can range from pretty small if we end up with a [solution like Bootstrap's](https://github.com/twbs/bootstrap/blob/main/js/src/util/sanitizer.js) to much larger with more full-fledged libraries) 

[...]

Issue #2327, opened 2021-08-26
Improve accordion with summary line semantics and presentation when JS loads but CSS doesn't 


hannalaakso:
## What
We should consider the semantics and presentation of the accordion with the summary line if the CSS fails to load and JS doesn't. As was pointed out at the [accessibility clinic](https://github.com/alphagov/govuk-frontend/issues/2295#issuecomment-906449543), this is what a screen reader user will hear when navigating the component. 

This is also unlikely to be considered a good arrangement of content:
>“If you showed this to a content designer and said ‘is this good content for a button’, I’m pretty sure they would say it’s not.”

This is related to https://github.com/alphagov/govuk-frontend/issues/2336 - changing the design so that the button doesn't contain the summary line is likely to fix this issue. 

If we don't change the design, the visual presentation when JS loads but CSS doesn't could be improved by not transforming the sections to buttons if CSS hasn't loaded - but this is will not help with the semantics when JS does load.

## Why

We should ensure that the component follows good practise and makes sense semantically. Also, in the somewhat edge case where JS loads but CSS doesn't, the buttons end up with a large amount of content that isn't visually ordered.

## Who needs to know about this
Developer, designer

## Done when

- [ ] We've either resolved https://github.com/alphagov/govuk-frontend/issues/2336 which will hopefully help with this card too, or considered (and created a card for) a potential alternative solution

hannalaakso:
## General improvements

We're going to mitigate against related issues raised at the accessibility clinic, see 'Next steps' https://github.com/alphagov/govuk-frontend/issues/2295#issuecomment-922840762 for full write up.

## Improve the semantics of the summary in the button by adding punctuation
Another thing we should try to do is to **insert a comma after the summary line in the markup**. This is because we had the slightly unexpected feedback from the accessibility clinic that the punctuation symbol which was originally introduced to improve NVDA announcements (https://github.com/alphagov/govuk-frontend/pull/2257/commits/83c8b570d4f2101ccc39126e1d35e3ed0ca59886 - if the link breaks with rebasing, it's the commit titled "Add pause") also improves the semantics of the button by dividing up its content thematically, as demonstrated here:
![Screenshot 2021-09-20 at 14 47 57](https://user-images.githubusercontent.com/5007934/134014521-c14d0546-8e6f-4c80-b344-77328e5f52b3.png)

## Next steps
I'm going to leave this issue open for now so that we can track the work to:
[...]

PR #2932, opened 2022-10-19
Cross Side Scripting prevention by use of textContent

romaricpascal:
Implements the approach outlined in [this comment](https://github.com/alphagov/govuk-frontend/issues/2892#issuecomment-1274980946) as a safety against Cross Side Scripting (closing #2892):

> 1. rename translation keys from `...Html` to `...Text` as they'll now be injected by text, both on the Accordion (4 keys) and CharacterCount (6 keys)
> 2. replace injection of content in the CharacterCount to use `textContent` (2 places, visible and visually hidden announcement)
> 3. replace injection of content in the Accordion to use `textContent` (2 places, `setExpanded` and `updateShowAllButton`
> 4. add 2 new `show/hideSectionAriaLabelText` (or just `...AriaLabel`) options to the macro and in the default JavaScript configuration and update `setExpanded` and `updateShowAllButton` to use them for labelling the buttons
> 5. check that the accessible name gets conveyed properly to assistive tech now we've changed it to use aria-label

## Notable things:

The commits should paint a good picture of the step-by-step changes. It's worth noting though that for implementing the swap to `aria-label` in the accordion:

- I added test checking the accessible name of the node rather than just the presence of hidden content (these may not have been available in Puppeteer at the time, see [their blog article](https://developer.chrome.com/blog/puppetaria/) for more info about their API)
- I've made the `aria-label` match the current accessible name, including a leading space before the commas
- I've kept the commas in the HTML, as [they help delimit the label of the button before CSS kicks in](https://github.com/alphagov/govuk-frontend/issues/2327#issuecomment-922957442)
- The change means the HTML of the button before CSS kicks in no longer reads "Show this section"/"Hide this section", but only "Show"/"Hide". That's an unfortunate side-effect of this approach, but seems minor enough.

I'll need to make a pass with AssistiveLabs to check 5, but given tests checked the accessible name, I'm hoping things will work out OK on that side 🤞🏻


romaricpascal:
> I'll need to make a pass with AssistiveLabs to check 5, but given tests checked the accessible name, I'm hoping things will work out OK on that side 🤞🏻

The results are in and looks like nothing broke from the change: https://docs.google.com/spreadsheets/d/1r1541gETGjavSxpjQQOLiVLvlNvinHHPQw7X8RKUnyE/edit#gid=0 🎉 🎉 🎉

src/govuk/components/accordion/accordion.test.js
```diff
@@ -230,32 +231,57 @@ describe('/components/accordion', () => {
...
-          const numberOfExampleSections = 2
-          const hiddenText = await page.evaluate(() => document.body.querySelectorAll('.govuk-accordion .govuk-accordion__section .govuk-accordion__section-toggle-text .govuk-visually-hidden').length)
+          const node = await page.$(
+            'aria/Section A , Show this section[role="button"]'
+          )
+          expect(node).not.toBeNull()
+
+          await node.click()
+
+          const updatedNode = await page.$(
```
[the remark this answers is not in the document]
colinrotherham: A lot of the other tests are using longer `document.body.querySelector` from before we knew about the `$` and `$$` methods so feel free to simplify them too.

Don't forget we have a new [`getAttribute()` helper](https://github.com/alphagov/govuk-frontend/blob/main/lib/puppeteer-helpers.js#L121) for **aria-label** as well 🎉

colinrotherham (commented):
Implementation looks fab 😊 

Offered a few suggestions on tests and selectors

Before we approve I'd like to be a bit more confident around accessibility testing, perhaps digging through some of the problems around dictation (Dragon) first time around

Thanks @romaricpascal

src/govuk/components/accordion/accordion.test.js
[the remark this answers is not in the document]
romaricpascal: It's important we rely on the `aria` selector for that test as the aim is to check what accessible name gets provided to assistive technology, and not take the responsibility of computing it ourselves depending on what we provide as [there are quite a few possibilities, like the inside of axe-core show](https://github.com/dequelabs/axe-core/blob/006a681395422bbd0603bab346dbdc6b38087d83/lib/commons/text/accessible-text-virtual.js#L29-L37) (no luck on that side finding an easy computation of the accessible name, though).

Unfortunately, in its current form, Puppeteer does not provide a `Element.getAccesibleTreeNode()` (invented name)
 that would let us cache the first found node to check that its corresponding bit in the accessibility tree has changed how we desire. ([fingers crossed that'll land in the future](https://github.com/puppeteer/puppeteer/issues/3641)). So we do need to re-query for that second ElementHandle[^1]
 
The risk of confusion is pretty limited, I think, given we're checking a distinct label "Section A" in the accessible name alongside the role `[role="button"]`. Given the example it runs onto and what we expect of our component, we know it's a unique thing. I could check for the absence of the other button but I don't think it adds much.

[^1]: Only other access to the accessibility tree information is through `page.accessibility.snapshot()`, but same issue we'd have to snapshot twice, and on top of that, do the same kind of walk Puppeteer must be doing when running its selectors.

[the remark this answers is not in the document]
colinrotherham: @romaricpascal I tried the same, to see if a computed name was available 😭 

But yeah still test for the accessible name, I like it 👍 

The concern was solely that the new test passes with duplicate buttons on the page:

**Button 1** `aria/Section A , Hide this section[role="button"]`
**Button 2** `aria/Section A , Show this section[role="button"]`

We're not testing that it's the same button with toggled text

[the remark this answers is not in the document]
romaricpascal: The examples we run on ensure the uniqueness, but I had a further dig to see if we could be safe in that respect anyway. 

Turns out you can get the accessible name from an `ElementHandle`, it's just not on the `ElementHandle` itself 🎉  `page.accessibility.snapshot()` accepts a `root` option to get a snapshot starting at a given node in the accessibility tree. Not all nodes will have an entry in the accessibility tree, so it may return null.

I've [updated the tests](https://github.com/alphagov/govuk-frontend/pull/2932/files#diff-074bb9b59bea515d21a2ba82e0e9f1cb6f0959302b6b9ea41c1c76cc80b31162R258-R284) which work more like you'd expect them to (check accessibility name, take action, check accessibility name) and [encapsulated the access to the tree and accessibility name in helpers](https://github.com/alphagov/govuk-frontend/pull/2932/files#diff-e8850c0e27106542fde2e15df546e04781301f4c63d9f887f56ac80c9e641a3bR125-R153) as it's something we'll likely want to use in other tests.

lib/puppeteer-helpers.js
```diff
@@ -122,6 +122,35 @@ function getAttribute ($element, attributeName) {
   return $element.evaluate((el, name) => el.getAttribute(name), attributeName)
 }
 
+/**
+ * Snapshots the node in the accessibility tree corresponding to the given element
+ *
+ * @param {import('puppeteer').ElementHandle} $element - Puppeteer element handle
+ * @returns {Promise<import('puppeteer').SerializedAXNode>})} -
```
colinrotherham, line 129: @romaricpascal Have we got a couple of extra characters here?

```suggestion
 * @returns {Promise<import('puppeteer').SerializedAXNode>} -
```

lib/puppeteer-helpers.js
```diff
@@ -122,6 +122,35 @@ function getAttribute ($element, attributeName) {
...
+}
+
+/**
+ * Gets the accessible name of the given element, if it exists in the accessibility tree
+ *
+ * @param {import('puppeteer').ElementHandle} $element - Puppeteer element handle
+ * @returns {string} - The element's accessible name
+ */
+async function getAccessibleName ($element) {
+  return (await getAccessibilityTreeNode($element)).name
```
colinrotherham, line 151: Instead of throwing from `getAccessibilityTreeNode` can we do this?

Then the `undefined` name will make it through to the Jest assertion, failing the test

```suggestion
  return (await getAccessibilityTreeNode($element))?.name
```

lib/puppeteer-helpers.js
```diff
@@ -122,6 +122,35 @@ function getAttribute ($element, attributeName) {
...
+ * @returns {Promise<import('puppeteer').SerializedAXNode>})} -
+ *   The accessibility tree node corresponding to the element
+ * @throws {Error} If no node corresponds to the element in the accessibility tree
+ *   (for ex. if you query an AT node for a `span` with no role)
+ */
+function getAccessibilityTreeNode ($element) {
+  const node = page.accessibility.snapshot({ root: $element })
+
+  if (!node) {
+    throw new Error('No accessibility tree node exist for this element')
```
colinrotherham, line 138: Adding a comment about throwing https://github.com/alphagov/govuk-frontend/pull/2932/files#r1003247281

src/govuk/components/accordion/accordion.test.js
```diff
@@ -230,32 +231,57 @@ describe('/components/accordion', () => {
...
 
-      describe('hidden text', () => {
-        it('should contain hidden text " this section" for assistive technology', async () => {
+      describe('accessible name', () => {
+        it('Should set the accessible name of the show/hide section buttons', async () => {
           await goToComponent(page, 'accordion')
 
-          const numberOfExampleSections = 2
-          const hiddenText = await page.evaluate(() => document.body.querySelectorAll('.govuk-accordion .govuk-accordion__section .govuk-accordion__section-toggle-text .govuk-visually-hidden').length)
+          const element = await page.$('.govuk-accordion__section-button')
```
colinrotherham, line 256: Think we try to stick to the jQuery `$` prefix for DOM elements

(Even though we also do it for Puppeteer handles)

```suggestion
          const $element = await page.$('.govuk-accordion__section-button')
```

src/govuk/components/accordion/accordion.test.js
```diff
@@ -230,32 +231,57 @@ describe('/components/accordion', () => {
...
+
+          expect(getAccessibleName(element)).resolves.toBe(
+            'Test , Additional description , Show this section'
+          )
 
-          expect(hiddenText).toEqual(numberOfExampleSections)
+          await element.click()
+
+          expect(getAccessibleName(element)).resolves.toBe(
```
colinrotherham, line 282: This is superb 🎉 

Proves that the SAME element has newly updated accessible text

src/govuk/components/accordion/accordion.test.js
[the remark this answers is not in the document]
colinrotherham: Thanks @romaricpascal, absolutely perfect

Means we're checking the same element so I'm happy

src/govuk/components/accordion/accordion.test.js
```diff
@@ -329,7 +355,41 @@ describe('/components/accordion', () => {
...
+
+        it('should localise "Show section" aria-label based on data attribute', async () => {
+          await goToComponent(page, 'accordion', {
+            exampleName: 'with-translations'
+          })
+
+          const showSectionDataAttribute = await page.evaluate(() =>
+            document.body
+              .querySelector('.govuk-accordion')
+              .getAttribute('data-i18n.show-section-aria-label')
```
colinrotherham, line 369: We alright using the [`getAttribute()`](https://github.com/alphagov/govuk-frontend/blob/main/lib/puppeteer-helpers.js#L121) helper?

lib/puppeteer-helpers.js
colinrotherham: Nah, that'd be like having `getAttribute()` or `getElementById()` crash if there's no attribute or element?

```js
// No accessible name – crashes
expect(await getAccessibleName($something)).toBeUndefined()
```

Not a blocker, make it crash for now if you like

src/govuk/components/accordion/accordion.test.js
romaricpascal: Actually, probably more a tidy up PR than a tidy up commit.  Most tests in the file use that pattern for reading attributes, so that'll muddy the changes quite a bit.

To be fair, the `getAttribute` helper may not make things better in its current shape as it'll involve 2 interractions with the browser:

1. find the element
2. get its attribute

instead of a single call.

I'd be keen to rework `getAttribute` into something like that to avoid the query the element than get steps.

```js
function getAttribute(pageOrElement, selectorOrAttributeName, attributeName) {
	if (pageOrElement instance of ElementHandle) {
		return pageOrElement.evaluate((el, name) => el.getAttribute('name'), selectorOrAttributeName)
	}
	
	return pageOrElement.$eval(selectorOrAttributeName, (el, name) => el.getAttribute('name'), attributeName)
}
```

colinrotherham: Or switch to Cypress where it does all this for us 🙃

lib/puppeteer-helpers.js
romaricpascal: Actually there was no need for that throwing 🤦🏻 The code would throw already if trying to access `name` on `undefined` in `getAccessibleName`, so I removed the `getAccessibilityTreeNode` function altogether in favour of a straight call to `puppeteer.accessibility.snapshot()`.

lib/puppeteer-helpers.js
```diff
@@ -122,6 +122,16 @@ function getAttribute ($element, attributeName) {
   return $element.evaluate((el, name) => el.getAttribute(name), attributeName)
 }
 
+/**
+ * Gets the accessible name of the given element, if it exists in the accessibility tree
+ *
+ * @param {import('puppeteer').ElementHandle} $element - Puppeteer element handle
+ * @returns {string} - The element's accessible name
```
colinrotherham, line 129: Looks like `page.accessibility.snapshot()` can return undefined

```suggestion
 * @returns {string | undefined} - The element's accessible name
```

lib/puppeteer-helpers.js
```diff
@@ -122,6 +122,16 @@ function getAttribute ($element, attributeName) {
...
+ */
+async function getAccessibleName ($element) {
+  return (await page.accessibility.snapshot({ root: $element })).name
```
colinrotherham, line 132: Because of a potential `Promise<null>` return we'll need a quick `?.` I guess?

```suggestion
  return (await page.accessibility.snapshot({ root: $element }))?.name
```

src/govuk/components/accordion/accordion.mjs
```diff
@@ -384,3 +401,19 @@ Accordion.prototype.getButtonPunctuationEl = function () {
 }
 
 export default Accordion
+
+/**
+ * @typedef {object} AccordionTranslations
```
colinrotherham, line 406: I know we don't have a rule for it but probably want a heading for JSDoc generation

```suggestion
 * Accordion translations
 *
 * @typedef {object} AccordionTranslations
```

romaricpascal:
Testing with Dragon done, and it's all OK on that side too: https://docs.google.com/spreadsheets/d/1r1541gETGjavSxpjQQOLiVLvlNvinHHPQw7X8RKUnyE/edit#gid=0 🎉 🎉 🎉

colinrotherham (approved):
Approving this, lots of hard work 🙌 

Some minor comments/suggestions added but not blockers

PR #2933, opened 2022-10-20
Ensure tests fail for browser exceptions

colinrotherham:
This PR ensures all browser tests (Puppeteer, jsdom) fail for unhandled browser exceptions

It does a few separate things:

1. Adds [Jest custom environments](https://jestjs.io/docs/configuration#testenvironment-string) for Puppeteer, jsdom
2. Adds `process.emit('error', error)` to ensure browser exceptions fail tests
3. Prevents Jest Puppeteer incorrectly emitting `uncaughtException`
4. Sets the unit test default environment to jsdom (not Node.js)

### DIFF
--- a/config/jest/environment/jsdom.mjs
+++ b/config/jest/environment/jsdom.mjs
@@ -7,6 +7,12 @@ import { TestEnvironment } from 'jest-environment-jsdom'
 class BrowserVirtualEnvironment extends TestEnvironment {
   async setup () {
     await super.setup()
+
+    // Access virtual console
+    const { virtualConsole } = this.dom
+
+    // Ensure test fails for browser exceptions
+    virtualConsole.on('jsdomError', (error) => process.emit('error', error))
   }
 }
 
--- a/config/jest/environment/jsdom.mjs
+++ b/config/jest/environment/jsdom.mjs
@@ -1,0 +1,13 @@
+import { TestEnvironment } from 'jest-environment-jsdom'
+
+/**
+ * Virtual browser environment
+ * Adds jsdom window/document globals
+ */
+class BrowserVirtualEnvironment extends TestEnvironment {
+  async setup () {
+    await super.setup()
+  }
+}
+
+export default BrowserVirtualEnvironment
--- a/jest-puppeteer.config.js
+++ b/jest-puppeteer.config.js
@@ -3,6 +3,12 @@ module.exports = {
   browserPerWorker: true,
 
   /**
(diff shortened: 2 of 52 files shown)


config/jest/environment/jsdom.mjs
```diff
@@ -0,0 +1,19 @@
...
+/**
+ * Virtual browser environment
+ * Adds jsdom window/document globals
+ */
+class BrowserVirtualEnvironment extends TestEnvironment {
+  async setup () {
+    await super.setup()
+
+    // Access virtual console
+    const { virtualConsole: console } = this.dom
```
36degrees, line 12: **nitpick:** Given the rename I think it'd be easier to read if we just used a normal assignment here, unless there's a particular reason to destructure that I'm missing:

```suggestion
    // Access virtual console
    const console = this.dom.virtualConsole
```

jest-puppeteer.config.js
```diff
@@ -1,6 +1,7 @@
 module.exports = {
   browserContext: 'incognito',
   browserPerWorker: true,
+  exitOnPageError: false,
```
36degrees, line 4: **suggestion:** worth adding a comment here that we handle this ourselves?

```suggestion
  // We handle errors ourselves in config/jest/environment/puppeteer.mjs
  exitOnPageError: false,
```

36degrees (approved):
Generally looks really good – have tested by recreating the issue we had in September (trying to access $module.dataset when module was undefined) and we now get actually useful messages surfaced in the test output 🙌🏻 

A minor nitpick and a suggestion for your consideration, but neither are blocking so happy to approve.

config/jest/environment/jsdom.mjs
colinrotherham: I'm happy to not rename?

```mjs
const { virtualConsole } = this.dom
```

36degrees: That works too 👍🏻

colinrotherham: Done

jest-puppeteer.config.js
colinrotherham: Done

colinrotherham:
Rebased since https://github.com/alphagov/govuk-frontend/pull/2937 merged and all passing still

I'll get this merged as it holds up https://github.com/alphagov/govuk-frontend/pull/2901 and https://github.com/alphagov/govuk-frontend/pull/2902

PR #2934, opened 2022-10-20
Update translated example for Character Count in the review app

### LINKED ISSUE #2701: Add ability to pass translation strings into character count component via data-attributes

romaricpascal:
- Updates the character counting component with translations
- Adds a word counting component, to illustrate variations between "one", "two" and "other"
- Adds a warning notification banner to tell not to use our translations (which I'd be very happy for the wording to be tweaked if necessary @claireashworth).

A shortcut to the resulting page: https://govuk-frontend-pr-2934.herokuapp.com/examples/translated

This should be the last step to close #2701 🎉

### DIFF
--- a/app/views/examples/translated/index.njk
+++ b/app/views/examples/translated/index.njk
@@ -110,7 +110,6 @@
     name: "with-hint",
     id: "with-hint",
     maxlength: 200,
-    fallbackHintText: "Gallwch ddefnyddio hyd at %{count} nod",
     label: {
       text: "Allwch chi roi mwy o fanylion?",
       classes: "govuk-label--l",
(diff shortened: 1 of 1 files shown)


app/views/examples/translated/index.njk
```diff
@@ -36,6 +36,21 @@
   {{ govukBackLink({
     "href": "/"
   }) }}
+
+  {% call govukNotificationBanner({
```
colinrotherham, line 40: Think we should go with the **Full page examples** approach instead?
https://github.com/alphagov/govuk-frontend/blob/main/app/views/full-page-examples/index.njk#L17

Don't think [Notification Banner](https://design-system.service.gov.uk/components/notification-banner/) fits the purpose

>Use a notification banner to tell the user about something they need to know about, but that’s not directly related to the page content.

>If the information is directly relevant to the thing the user is doing on that page, put the information in the main page content instead. Use [inset text](https://design-system.service.gov.uk/components/inset-text/) or [warning text](https://design-system.service.gov.uk/components/warning-text/) if it needs to stand out.

colinrotherham: Something plain and simple? @claireashworth

```html
<h1 class="govuk-heading-xl">
  Translated examples
</h1>

<p class="govuk-body-l">These examples are designed to test how patterns and components can be translated.</p>

<p class="govuk-body">Please do not use these translations directly in your application.</p>
```

Image showing the suggested text

claireashworth: If possible, I would swap the second 2 lines to make extra sure the user sees the warning. 
And how about this:

'These examples only illustrate how our components display translated text.'

romaricpascal:
Good spot on the misuse of the component, @colinrotherham. I've amended the way the warning is displayed (and used a [warning text](https://design-system.service.gov.uk/components/warning-text/) to make sure the warning stands out @claireashworth). 

Screenshot 2022-10-24 at 11 30 57

app/views/examples/translated/index.njk
```diff
@@ -32,14 +32,28 @@
 
 {% extends "layout.njk" %}
 
+{%block pageTitle%}Translated examples - GOV.UK Frontend{% endblock %}
```
colinrotherham, line 35:
```suggestion
{% block pageTitle %}Translated examples - GOV.UK Frontend{% endblock %}
```

app/views/examples/translated/index.njk
```diff
@@ -32,14 +32,28 @@
...
+    text: "Please do not use these translations directly in your application.",
+    iconFallbackText: "Warning"
+  }) }}
+
+  <p class="govuk-body">
+    These examples only illustrate how our components display translated text.
+    We do not guarantee that they are accurate.
+  </p>
+
+  <hr class="govuk-section-break govuk-section-break--l" />
```
colinrotherham, line 55: Had a quick look, don't think we tend to use the XHTML-style self-closing `<hr>`

```suggestion
  <hr class="govuk-section-break govuk-section-break--l">
```

app/views/examples/translated/index.njk
```diff
@@ -32,14 +32,27 @@
...
-  <h1 class="govuk-heading-xl">Translated</h1>
+  <h1 class="govuk-heading-xl">Translated examples</h1>
+
+  {{ govukWarningText({
+    text: "Please do not use these translations directly in your application.",
+    iconFallbackText: "Warning"
+  }) }}
+
+  <p class="govuk-body">
+    These examples only illustrate how our components display translated text.
```
colinrotherham, line 52: Thanks @romaricpascal

I can see pushes going up with different text changes being tried out

Think we need to apply the "two thirds" column layout, maybe keep `govuk-body-l` on the opening text, and go with the `--xl` break like some of the other pages do (compare the two visually?) but I'll take a 2nd opinion

Try this:

```html
<div class="govuk-grid-row">
  <div class="govuk-grid-column-two-thirds">
    <h1 class="govuk-heading-xl">Translated examples</h1>

    <p class="govuk-body-l">
      These examples only illustrate how our components display translated text.
      We do not guarantee that they are accurate.
    </p>

    {{ govukWarningText({
      text: "Please do not use these translations directly in your application.",
      iconFallbackText: "Warning"
    }) }}

    <hr class="govuk-section-break govuk-section-break--xl">
  </div>
</div>
```

☝️ @claireashworth suggested putting the warning first to make sure it's seen, but not sure if that's necessary now we're using `govukWarningText()` so I've swapped back

Screenshot showing HTML example above

romaricpascal:
Updated with Colin's latest proposal for the style of the warning, which had a much nicer hierarchy 🙌🏻 . 

At the end of the day, that warning only appears to people that would stumble upon one our review apps, so I'm not sure how much back-and-forth we should have about it given its very limited audience, as long as it's clear enough 😄

colinrotherham (approved):

claireashworth (approved):

PR #2935, opened 2022-10-21
IE8 fixes for I18n

36degrees:
- Avoid using `Array.prototype.indexOf` [which isn't supported by IE8](https://caniuse.com/mdn-javascript_builtins_array_indexof) – use a simple loop instead
- IE8 doesn't like us passing flags alongside a literal regex when calling RegExp (this [seems to be an ES6 feature](https://github.com/mdn/content/blob/ae586f9996f622b40379283f727febe94258b8c0/files/en-us/web/javascript/reference/global_objects/regexp/index.md#flags-in-constructor)). General best practise seems to be to avoid calling RegExp if we don't need to construct the expression dynamically – as we already have a literal regex, let's just use that and pass the flags as part of the literal regex.

  (I've manually tested in IE8 to verify that this works, and that the global flag is being applied and multiple placeholders are replaced if passed)


colinrotherham (approved):
Appreciate the faff involved in narrowing down such a simple change 🙌

PR #2936, opened 2022-10-21
Avoid trap where RegExp.exec misses placeholders

36degrees:
RegExp.exec uses a `lastIndex` to remember where in the string to start searching on the next iteration.

So given a string like this:

```
Hi, %{name}. %{name}, it sure is nice to meet you.
```

After the first call to RegExp.exec `lastIndex` will point here:

```
Hi, %{name}. %{name}, it sure is nice to meet you.
           ↑ lastIndex
```

If we then alter the string by replacing the placeholder with text that is shorter than the placeholder, we'll then start searching from _after_ the next placeholder, and so will never match it:

```
Hi, Bob. %{name}, it's nice to meet you.
           ↑ lastIndex
```

Avoid this issue by updating the `replacePlaceholders` function to use `String.replace` instead, passing a function to take care of determining the replacement value using the same logic as before.

Move the construction of the `Intl.NumberFormat` outside of the object so that we don't have to bind the replacement function to `this`, and to save constructing multiple instances of `Intl.NumberFormat` when we only need one.

Add a test which fails without the changes to the `replacePlaceholders` function.

### DIFF
--- a/src/govuk/i18n.mjs
+++ b/src/govuk/i18n.mjs
@@ -68,35 +68,32 @@ I18n.prototype.t = function (lookupKey, options) {
  * @returns {string} The translation string to output, with ${} placeholders replaced
  */
 I18n.prototype.replacePlaceholders = function (translationString, options) {
-  // eslint-disable-next-line prefer-regex-literals
-  var placeholderRegex = /%{(.\S+)}/g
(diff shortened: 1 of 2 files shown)


colinrotherham (approved):
🎉

PR #2572, opened 2022-03-13
Improving dev cross-platform support

TomBillingtonUK:
This PR:
- Fixes the lint:scss npm command so it runs cross-platform 
- Adds windows specific logic to the lint gulp tasks
- Introduces the npm package cross env to allow tests to work cross-platform

Changes have been tested on both a windows 11 machine and a MacBook running Monterey

PR #2938, opened 2022-10-21
Add `cross-env-shell` for Windows support

colinrotherham:
This PR adds `cross-env-shell` to restore Windows support

This ensures our process exit signals work correctly with `jest-dev-server`

https://www.npmjs.com/package/cross-env#cross-env-vs-cross-env-shell
>On Windows you need to use cross-env-shell, if you want to handle [signal events](https://nodejs.org/api/process.html#process_signal_events) inside of your program. A common case for that is when you want to capture a SIGINT event invoked by pressing Ctrl + C on the command-line interface.

For context, we accidentally removed `cross-env` in #2850 (kindly contributed as a fix in https://github.com/alphagov/govuk-frontend/pull/2572)

This PR is number 2) in a three-part special:

1. https://github.com/alphagov/govuk-frontend/pull/2946
2. https://github.com/alphagov/govuk-frontend/pull/2938
3. https://github.com/alphagov/govuk-frontend/pull/2945

### DIFF
--- a/config/jest/server/start.mjs
+++ b/config/jest/server/start.mjs
@@ -30,7 +30,11 @@ export default async function serverStart () {
   try {
     await ready({ timeout })
   } catch (error) {
-    await setup({ command: `PORT=${PORT} node app/start.js`, port: PORT })
+    await setup({
+      // Ensure PORT works (on Windows) + SIGINT/SIGTERM signal events
+      command: `cross-env-shell PORT=${PORT} node app/start.js`,
+      port: PORT
+    })
     await ready()
   }
 }


.github/workflows/tests.yml
```diff
@@ -72,6 +77,7 @@ jobs:
           npx jest --color --selectProjects "JavaScript behaviour tests"
 
       - name: Send screenshots to Percy
+        if: matrix.os == 'ubuntu-latest'
```
colinrotherham, line 80: Do probably want a better way of doing this as this excluded all **JavaScript component tests** on Windows (not just Percy screenshots)

config/jest/server/start.mjs
```diff
@@ -30,7 +30,10 @@ export default async function serverStart () {
   try {
     await ready({ timeout })
   } catch (error) {
-    await setup({ command: `PORT=${PORT} node app/start.js`, port: PORT })
+    await setup({
+      command: `cross-env-shell PORT=${PORT} node app/start.js`,
```
colinrotherham, line 34: Previously used `cross-env` but `cross-env-shell` (same package, different executable) ensures our process exit signals work correctly with [`jest-dev-server`](https://www.npmjs.com/package/jest-dev-server)

lib/file-helper.js
```diff
@@ -16,13 +17,22 @@ const getListing = async (directoryPath) => {
...
       basename,
-      path: normalize(entryPath),
+      path: entryPath,
       stats: await stat(entryPath)
-    }]
+    }
+
+    // Follow child directory listing
+    if (entry.stats.isDirectory()) {
+      entry.entries = await getListing(entryPath)
```
colinrotherham, line 32: This tiny change means we can remove [`recursive-readdir`](https://www.npmjs.com/package/recursive-readdir) and our `getFilesByDirectory()` helper function

package.json
```diff
@@ -22,12 +22,12 @@
     "build:compile": "gulp compile",
     "build:sassdoc": "sassdoc --config sassdoc.config.yaml ./src/govuk",
     "build:jsdoc": "jsdoc --configure jsdoc.config.js ./src/govuk",
-    "build:package": "gulp build:package --destination 'package' && npm run test:build:package",
-    "build:dist": "gulp build:dist --destination 'dist' && npm run test:build:dist",
+    "build:package": "gulp build:package --destination \"package\"",
+    "build:dist": "gulp build:dist --destination \"dist\"",
+    "postbuild:package": "jest --color --selectProjects \"Gulp tasks\" --testMatch \"**/*build-package*\"",
```
colinrotherham, line 27: By stitching npm commands together with `&&` we saw difficulty detecting "exits"
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#exit-codes-and-error-action-preference

colinrotherham:
>Expected — Waiting for status to be reported

This PR would need the **Branch protection rules** → **Require status checks to pass before merging** (GitHub repository settings) checks to change from "Run tests" to both:

1. **Run tests (windows-latest)**
2. **Run tests (ubuntu-latest)**

package.json
colinrotherham: This also allows us to run npm scripts _without_ other scripts (like tests) also running

### Build + test
Includes npm `pre` and `post` scripts

```console
npm run build:compile
npm run build:package
npm run build:dist
```

### Build only (no tests)
Excludes npm `pre` and `post` scripts

```console
npm run build:compile
npm run build:package --ignore-scripts
npm run build:dist --ignore-scripts
```

romaricpascal (approved):

config/jest/server/start.mjs
colinrotherham: We're using `cross-env-shell` primarily for the **PORT** env to be picked up correctly on Windows

Problem with the suggested comment is that behind the scenes (on Linux, macOS) `SIGINT` and `SIGTERM` signals already worked perfectly. Using `cross-env` would break this where `cross-env-shell` wouldn't

It's also on the [`cross-env-shell` README](https://github.com/kentcdodds/cross-env#cross-env-vs-cross-env-shell)

Think this comment does the trick?

```suggestion
      // Ensure PORT works (on Windows) + SIGINT/SIGTERM signal events
      command: `cross-env-shell PORT=${PORT} node app/start.js`,
```

PR #2942, opened 2022-10-24
Fix exit this page test failures

owenatgov:
Fixes 2 issues with tests in the Exit this Page prototype: https://github.com/alphagov/govuk-frontend/pull/2545

- We don't detect if there aren't any buttons on a page
- We're using the wrong keyboard key when using puppeteer to test for esc key press

This may interact in an awkward way with https://github.com/alphagov/govuk-frontend/pull/2940 as it's up in the air if we're likely to have more than one exit this page button on the page or how we handle this.


querkmachine (approved):
Tests failed before. They pass now. Job's a good 'un.

PR #2943, opened 2022-10-25
Add a full page example demonstrating Exit this Page

owenatgov:
Add a full page example to the review app that demonstrates the use of exit this page. This is a rough version of the proposed EtP pattern that is likely to be shipped with EtP and allows us to effectively test EtP without having to create pre-releases into other prototypes.

Likely to be additionally useful for integration testing later on.

### DIFF
--- a/app/views/full-page-examples/child-maintenance/index.njk
+++ b/app/views/full-page-examples/child-maintenance/index.njk
@@ -1,0 +1,133 @@
+---
+scenario: >-
+  You are a user looking for information on child maintenance.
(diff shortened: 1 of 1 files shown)


querkmachine (approved):

querkmachine:
Approved on the basis that this matches the example presented in the design system mockup, but I'm not sure if this is the 'ideal' presentation for this page. 

Specifically, the button appears too far right-aligned on large enough screens due to not having the inline padding of a surrounding column. However, adding a wrapping column prevents the use of `position: sticky`. Bit of a catch-22.

Issue #2637, opened 2022-05-20
Drop support for Ruby Sass and LibSass


36degrees:
## What

Drop support for Ruby Sass and LibSass, including:

- removing mentions of them from the dependencies section of [our install docs](https://frontend.design-system.service.gov.uk/installing-with-npm/)
- removing the tests for the Ruby Sass and LibSass compilers from [our Sass CI workflow](https://github.com/alphagov/govuk-frontend/blob/main/.github/workflows/sass.yaml)

Although there are no code changes, it's highly likely that we'll choose to make related breaking changes as part of this work. Regardless, it's a significant change for our users and so we should treat it as a breaking change.

## Why

Ruby Sass was deprecated in April 2018 and reached End of Life in April 2019. We're not aware of any of our users still using it, but as supporting it has involved no extra work on our part we've supported it up until now.

LibSass has also been deprecated since October 2020.

The new Sass module system (`@use`) is only available in Dart Sass, and there are no plans to introduce it to LibSass. Additionally, a future release of Dart Sass will drop support for `@import` which will make it impractical for us to continue to support both Dart Sass and LibSass.

Dropping support for LibSass is also [a recommendation made by the Sass maintainers themselves](https://sass-lang.com/blog/libsass-is-deprecated):

> "We recommend all existing LibSass users make plans to eventually move onto Dart Sass, **and that all Sass libraries make plans to eventually drop support for LibSass.**
> 
> We’re no longer planning to add any new features to LibSass, including compatibility with new CSS features.
> 
> LibSass and Node Sass will continue to be maintained indefinitely on a best-effort basis, including fixing major bugs and security issues and maintaining compatibility with the latest Node versions."

Dropping support for Ruby Sass and LibSass will allow us to resolve the following issues:

- https://github.com/alphagov/govuk-frontend/issues/1791
- https://github.com/alphagov/govuk-frontend/issues/2238
- https://github.com/alphagov/govuk-frontend/issues/2238
- https://github.com/alphagov/govuk-frontend/issues/2361

## Who needs to work on this

Developers, technical writer

## Who needs to review this

Developers, technical writer

## Done when

- [ ] Install docs make it clear that only Dart Sass is supported
- [ ] We no longer test compilation using Ruby Sass and LibSass

36degrees:
[...]

PR #2945, opened 2022-10-26
Add (optional) GitHub Action checkbox for runner OS

colinrotherham:
Split out from #2938 and enables a choice of GitHub test runner:

```
- macos-latest
- ubuntu-latest
- windows-latest
```

Choice of runner OS

This PR is number 3) in a three-part special:

1. https://github.com/alphagov/govuk-frontend/pull/2946
2. https://github.com/alphagov/govuk-frontend/pull/2938
3. https://github.com/alphagov/govuk-frontend/pull/2945

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -10,6 +10,15 @@ on:
       - 'support/**'
 
   workflow_dispatch:
+    inputs:
+      runner:
+        description: Run tests on
+        type: choice
+        default: ubuntu-latest
+        options:
+          - macos-latest
+          - ubuntu-latest
+          - windows-latest
 
 concurrency:
   group: tests-${{ github.head_ref || github.run_id }}
@@ -18,7 +27,7 @@ concurrency:
 jobs:
   install:
     name: Install
-    runs-on: ubuntu-latest
+    runs-on: ${{ github.event.inputs.runner || 'ubuntu-latest' }}
 
     env:
       PUPPETEER_SKIP_CHROMIUM_DOWNLOAD: true
@@ -32,7 +41,7 @@ jobs:
 
   build:
     name: Build
-    runs-on: ubuntu-latest
+    runs-on: ${{ github.event.inputs.runner || 'ubuntu-latest' }}
     needs: [install]
 
     steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/tests.yml
```diff
@@ -72,6 +77,7 @@ jobs:
           npx jest --color --selectProjects "JavaScript behaviour tests"
 
       - name: Send screenshots to Percy
+        if: matrix.os == 'ubuntu-latest'
```
colinrotherham, line 80: Do probably want a better way of doing this as this excluded all **JavaScript component tests** on Windows (not just Percy screenshots)

colinrotherham:
Look how slow everything runs on Windows 😭

.github/workflows/tests.yml
colinrotherham: Rebased with https://github.com/alphagov/govuk-frontend/pull/3009 so we can leave the Percy screenshot job as `ubuntu-latest` only

romaricpascal (approved):
The timings make me so sad 😭 But at least we check that things work OK on Windows and can look into further optimising if that's possible at all 😬

romaricpascal:
Would the merge of this require some update of our expected checks?

romaricpascal:
> Can we keep this PR open whilst we see where the Windows performance work on [GitHub **actions/cache**](https://github.com/actions/cache#readme) goes?

I don't see any rush in mergin it, but it was worth marking it as good to go. Maybe we should put it in blocker on the project board for now?

colinrotherham:
@romaricpascal @36degrees Not quite got these checks down to the 2 minute mark (that we've been used to recently) but the last run was below 5 mins thanks to some cache trimming:

* https://github.com/alphagov/govuk-frontend/pull/3154

I'll continue to park this PR but could see it getting merged with:

1. Single `npm ci` (for all platforms) using [`enableCrossOsArchive`](https://github.com/actions/cache#inputs)
2. Skipping non-essential checks on Windows

For the time savings from 1) we'd need [Dart Sass](https://github.com/alphagov/govuk-frontend/issues/2637) to avoid platform-specific `npm rebuild node-sass`

colinrotherham:
@romaricpascal As discussed the runner OS is now a workflow input but defaults to `ubuntu-latest`

PR #2937, opened 2022-10-21
Update dependencies + fix breaking changes

colinrotherham:
The following package updates have been applied:

## Compatible
Minor and patch semver-compatible changes

`@babel/core` → **`^7.19.6`**
`@percy/cli` → **`^1.12.0`**
`babel-jest` → **`^29.2.1`**
`cookie-parser` → **`^1.4.6`**
`eslint-plugin-jsdoc` → **`^39.3.14`**
`eslint-plugin-promise` → **`^6.1.1`**
`express` → **`^4.18.2`**
`express-validator` → **`^6.14.2`**
`govuk_frontend_toolkit` → **`^9.0.1`**
`gulp-task-listing` → **`^1.1.1`**
`jest` → **`^29.2.1`**
`jest-environment-jsdom` → **`^29.2.1`**
`map-stream` → **`^0.0.7`**
`marked` → **`^4.1.1`**
`nodemon` → **`^2.0.20`**
`outdent` → **`^0.8.0`**
`stylelint` → **`^14.14.0`**
`undici` → **`^5.11.0`**
`ws` → **`^8.9.0`**

## Major but compatible
No breaking changes

`glob` → **`^8.0.3`**
`jest-axe` → **`^6.0.0`**
`puppeteer` → **`^19.1.0`**
`yargs` → **`^17.6.0`**

## Breaking changes
Cheerio no longer encodes non-ASCII characters so `.html()` are true to the HTML source

`cheerio` → **`^1.0.0-rc.12`**

## Incompatible
Intentionally pinned back to older (not latest) versions

`html-validate` → **`7.5.0`**
`rollup` → **`0.56.5`**

## Added
Previously missing and only available as a child dependency

`nunjucks` → **`^3.2.3`**

```console
govuk-frontend-repository@ /path/to/project/govuk-frontend
└─┬ sassdoc@2.7.4
  └─┬ sassdoc-theme-default@2.8.3
    └── nunjucks@3.2.2
```

## Removed
These packages were either unused or not needed

```console
@percy/logger
expect-puppeteer
gulp-eol
fsevents
```

PR #2950, opened 2022-11-01
Update all dependencies using Dependabot

colinrotherham:
This PR enables Dependabot updates for npm and GitHub Actions

Previously only security-related updates were included

I've ignored intentionally pinned "do not update" packages flagged via https://github.com/alphagov/govuk-frontend/pull/2937

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,13 @@
+version: 2
+
+updates:
+
+  # Update npm packages
+  - package-ecosystem: npm
+    directory: /
+    reviewers:
+      - alphagov/design-system-developers
+    schedule:
+      interval: daily
+    allow:
+      - dependency-type: all
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,3 +11,11 @@ updates:
       interval: daily
     allow:
       - dependency-type: all
+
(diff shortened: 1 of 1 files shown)


domoscargin:
I think this looks good to me and is worth doing, but the potential noise might be an issue, so let's chat about it in the next squad catch up.

colinrotherham:
> I think this looks good to me and is worth doing, but the potential noise might be an issue, so let's chat about it in the next squad catch up.

Sounds good. If it helps there's a maximum [`open-pull-requests-limit`](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#open-pull-requests-limit) of 5 (by default)

Hopefully can become a time saver by doing it for us. Thanks @domoscargin

colinrotherham:
Questions:

1. Do we want `weekly` not `daily` checks?
2. Do other "worrying updates" need ignoring?
3. Do semver ranges in `package.json` need updating to match?
4. Do we want an ["Add to project" GitHub Action](https://github.com/actions/add-to-project) workflow before we merge?

For 3) this is assuming that Dependabot might respect our configured ranges

36degrees (approved):
Let's trial this for a bit and see how it works for us 👍🏻

PR #2962, opened 2022-11-04
Use absolute directory paths everywhere

colinrotherham:
This PR addresses a few code review comments that have come up recently

**Where are my files?**
Having to use the variable `process.cwd()` to find where the project root

```mjs
const pathToPackage = join(cwd(), configPaths.package, 'package.json')
```

**Where is the src directory?**
We suffix "govuk" into `configPaths.src` but (can) forget to join it to destination paths

```mjs
const pathToSrc = join(configPaths.src, 'all.scss') // Hmm, no `govuk`?
const pathToDest = join(configPaths.package, 'govuk', 'all.scss')
```

**Where is the _actual_ src directory?**
Remembering to add `../` to remove the "govuk" suffix

```mjs
const pathToSrcForReal = join(cwd(), configPaths.src, '../')
```

This PR makes all our config paths absolute and removes the "govuk" suffix by default

## Config paths
We currently used relative paths with trailing slashes

### Before
Relative ish file paths
```mjs
import configPaths from '../../config/paths.js'

configPaths.app     // relative 'app/'
configPaths.dist    // relative 'dist/'
configPaths.package // relative 'package/'
configPaths.public  // relative 'public/'
```

### After (macOS, Linux)
Absolute file paths

```mjs
import configPaths from '../../config/paths.js'

configPaths.app     // absolute '/path/to/project/govuk-frontend/app'
configPaths.dist    // absolute '/path/to/project/govuk-frontend/dist'
configPaths.package // absolute '/path/to/project/govuk-frontend/package'
configPaths.public  // absolute '/path/to/project/govuk-frontend/public'
```

### After (Windows)
Absolute file paths

```mjs
import configPaths from '../../config/paths.js'

configPaths.app     // absolute 'C:\\path\\to\\project\\govuk-frontend\\app'
configPaths.dist    // absolute 'C:\\path\\to\\project\\govuk-frontend\\dist'
configPaths.package // absolute 'C:\\path\\to\\project\\govuk-frontend\\package'
configPaths.public  // absolute 'C:\\path\\to\\project\\govuk-frontend\\public'
```

PR #2964, opened 2022-11-04
Update `copy:files` to glob “src” not “src/govuk”

colinrotherham:
This PR widens `gulp copy:files` to glob `./src/**/*` not just the "govuk" directory

Why? Since https://github.com/alphagov/govuk-frontend/pull/2851 was merged we're now running build tasks on files within:

```console
./src/govuk
./src/govuk-prototype-kit
```

As suggested by @36degrees in https://github.com/alphagov/govuk-frontend/pull/2851#pullrequestreview-1167005193 we could remove some unnecessary file operations:

> (I think it'd be neater if the `.js` and `.scss` files were handled by the existing copy-files task, but that looks like it's caught up in the fact that config.src is actually `src/govuk` so happy to revisit that later)

These changes were made possible by:

* https://github.com/alphagov/govuk-frontend/pull/2962

### DIFF
--- a/tasks/gulp/copy-to-destination.js
+++ b/tasks/gulp/copy-to-destination.js
@@ -31,7 +31,7 @@ gulp.task('copy:files', () => {
      */
     merge(
       gulp.src([
-        `${slash(configPaths.src)}/govuk/**/*`,
+        `${slash(configPaths.src)}/**/*`,
 
         // Exclude files we don't want to publish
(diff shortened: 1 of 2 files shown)


colinrotherham:
@36degrees Does this PR address your code review suggestion?

36degrees (approved):
👍🏻

Issue #2835, opened 2022-09-11
typescript support


jhackett1:
it would be great if this library declared its types, to make it easier to understand what i can import from it and how it can be used.

at the moment, when i import it in a typescript project, i get an error:

```
Could not find a declaration file for module 'govuk-frontend'. '/Users/jayehackett/Downloads/wp-fiddle/node_modules/govuk-frontend/govuk/all.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/govuk-frontend` if it exists or add a new declaration (.d.ts) file containing `declare module 'govuk-frontend';`
```

i could perhaps make a pull request for this if it's wanted?

## Alternatives
i could define some a declaration file in my own project, but that would only solve the problem for me!

## Additional information (if applicable)
for a library like this, including types is a useful form of self-documentation. it's nice to be able to import a library and see intellisense suggestions for what i can do with it without having to refer to a documentation website.

<img width="910" alt="Screenshot 2022-09-11 at 18 26 45" src="https://user-images.githubusercontent.com/14189497/189541004-79374de6-fc22-4489-bfa3-2a309258ebd4.png">

jhackett1:
here's an example @cloudratha did for the hackney fork of this lib. doing it here would probably look similar: 

https://github.com/LBHackney-IT/lbh-frontend/pull/177

colinrotherham:
Thanks @jhackett1 for raising this issue

I know things have been quiet (or silent) from our end on this but it's certainly been talked about.

For now, you're likely seeing problems because `"strict": true` is set in your TypeScript **compilerOptions**. For any regular JavaScript code without type declarations, ours included, you'll see the same.

Microsoft have some more tips here for regular JavaScript:
https://code.visualstudio.com/docs/languages/jsconfig

>Tip: `jsconfig.json` is a descendant of [tsconfig.json](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html), which is a configuration file for TypeScript. `jsconfig.json` is `tsconfig.json` with "allowJs" attribute set to `true`.

### Generated type declarations in future

That said, we appreciate type declarations are a brilliant way of bringing documentation straight to the code

What could we do in the future?

1. Manually write type declarations
2. Keeping our JavaScript source, but output type declarations via `tsc` + JSDoc
3. Converting to TypeScript source, but compiling to JavaScript with type declarations via `tsc`

[...]

PR #2971, opened 2022-11-07
Enable type declaration linting

colinrotherham:
We've recently changed **Character Count** config properties across various PRs—but it looks like we missed a config property change or our JSDoc comment needs a few tweaks

Adding a compiler check will spot missing references like this:

```console
src/govuk/components/character-count/character-count.unit.test.mjs:104:17 - error TS2322: Type '{ charactersUnderLimitOne: string; }' is not assignable to type 'CharacterCountTranslations'.
  Object literal may only specify known properties, but 'charactersUnderLimitOne' does not exist in type 'CharacterCountTranslations'. Did you mean to write 'charactersUnderLimit'?

104                 charactersUnderLimitOne: 'Different custom text. Count: %{count}'
                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

```console
  src/govuk/components/character-count/character-count.mjs:59:35
    59 function CharacterCount ($module, config) {
                                         ~~~~~~
    An argument for 'config' was not provided.
```

This work follows on from the recently added:

But this PR adds a bit more detail:

1. Using the TypeScript compiler to "type check" our code
2. Checking we're calling library code as intended
3. Checking we're calling DOM methods as intended
4. Checking we're adhering to our own JSDoc correctly

### Why TypeScript?

We're not actually writing any TypeScript code but this PR includes it to:

1. Check type information via ESLint using ~[`standard`](https://www.npmjs.com/package/eslint-config-standard)~ [`standard-with-typescript`](https://www.npmjs.com/package/eslint-config-standard-with-typescript)
2. Run the TypeScript compiler `tsc` to check if our code would build without errors
3. Run the TypeScript compiler `tsc` to generate type declarations (see https://github.com/alphagov/govuk-frontend/issues/2835)


tsconfig.base.json
```diff
@@ -0,0 +1,13 @@
+{
+  "extends": "@tsconfig/recommended/tsconfig.json",
+  "compilerOptions": {
+    "allowJs": true,
+    "checkJs": true,
+    "lib": ["ESNext"],
+    "module": "NodeNext",
+    "noEmit": true,
+    "strict": false,
+    "strictBindCallApply": true,
```
colinrotherham, line 10: We do lots of `.bind(this)` so wanted to opt-in to strict checking for these 👍

tsconfig.base.json
```diff
@@ -0,0 +1,13 @@
...
+    "strictFunctionTypes": true
```
colinrotherham, line 11: We want [`strictFunctionTypes`](https://www.typescriptlang.org/tsconfig#strictFunctionTypes) to confirm JSDoc `@param {string}` is not called with `number` etc

tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
+  "extends": "./tsconfig.base.json",
+  "files": [],
+  "references": [
```
colinrotherham, line 4: For now I'm only going to include `app` and `src` for type declaration checks

tsconfig.base.json
```diff
@@ -0,0 +1,13 @@
+{
+  "extends": "@tsconfig/recommended/tsconfig.json",
+  "compilerOptions": {
+    "allowJs": true,
+    "checkJs": true,
+    "lib": ["ESNext"],
+    "module": "NodeNext",
+    "noEmit": true,
```
colinrotherham, line 8: We don't want TypeScript to compile anything yet, including source maps `*.js.map` and [types `*.d.ts`](https://github.com/alphagov/govuk-frontend/issues/2835)

colinrotherham:
GitHub automatically marked this PR as merged whilst splitting it into another branch

Can't see a way to undo this so opened again:

PR #2972, opened 2022-11-07
Improve colour contrast of Crown Copyright logo

### LINKED ISSUE #2134: Improve colour contrast of Crown Copyright logo
## What
The crown copyright logo against the footer background has a low contrast ratio. However, it does meet the contrast requirement for [meaningful graphics](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html) of 3:1. We don't think this is a WCAG fail and it doesn't need to be listed in an accessibility statement, but there's potential work we can do to improve the contrast so it better matches that of the text/links around it.

current Crown copyright logo in the footer
The crown copyright logo has a hex code colour of `#6E777A`, which against the footer background of `#F3F2F1` has a AA Large 4.09:1 colour contrast ratio.

Normal Text
WCAG AA:Fail
WCAG AAA:Fail

Large Text
**WCAG AA:Pass**
WCAG AAA:Fail

Graphical Objects and User Interface Components
**WCAG AA:  Pass**

## Why
[A comment was added to the epic about an issue raised in the Axe Review accessibility tool](https://github.com/alphagov/govuk-frontend/issues/1680#issuecomment-761044742). The failure is only of the logo and not to the footer text.

Designers, Developers, Policy

## Further detail
A suggestion could be to change the colour of the logo but we would need to check with the owners of the Crown Copyright logo.

Crown copyright logo in the footer using the darker text colour
Here is a mock-up of the crown copyright logo using the footer text colour `#454A4C` which has a contrast ratio of AAA 8.03:1 against the footer background `#F3F2F1`.

## Done when

- [ ] Agreed with Crown Copyright on approach
- [ ] Decided on whether the logo should be an .svg or .png
- [ ] Updated file created
- [ ] Pull request created and merged

NickColley:
Increases the copyright crest logo contrast to match the surrounding text and match the OGL logo.

[mask-image is available in modern browsers (with -webkit- prefix for some which is generated automatically)](https://caniuse.com/?search=mask-image) and we fallback to regular background-image in all other browsers.

Tested as functional (same as before) in  Internet Explorer 9-11.

Tested to increase contrast in Firefox latest, Chrome latest, Safari latest, iOS Safari latest, Chrome for Android latest.

## Screenshots

| Before | After |
| - | - |
| Crest with light grey color | Crest with darker colour matching text |

Closes #2134

### DIFF
--- a/src/govuk/components/footer/_index.scss
+++ b/src/govuk/components/footer/_index.scss
@@ -121,7 +121,6 @@
     @include govuk-device-pixel-ratio {
       background-image: govuk-image-url("govuk-crest-2x.png");
     }
-    background-repeat: no-repeat;
     background-size: contain;
 
     @supports (mask-image: url("")) {
@@ -132,7 +131,6 @@
       @include govuk-device-pixel-ratio {
         mask-image: govuk-image-url("govuk-crest-2x.png");
       }
-      mask-repeat: no-repeat;
       mask-size: contain;
     }
   }
--- a/src/govuk/components/footer/_index.scss
+++ b/src/govuk/components/footer/_index.scss
@@ -106,17 +106,35 @@
 
   .govuk-footer__copyright-logo {
     display: inline-block;
-    min-width: $govuk-footer-crest-image-width;
-    padding-top: ($govuk-footer-crest-image-height + govuk-spacing(2));
+    text-align: center;
+    white-space: nowrap;
+  }
+
+  .govuk-footer__copyright-logo:before {
+    content: "";
+    display: block;
+    width: $govuk-footer-crest-image-width;
+    height: $govuk-footer-crest-image-height;
+    margin: auto;
+    margin-bottom: govuk-spacing(2);
     background-image: govuk-image-url("govuk-crest.png");
     @include govuk-device-pixel-ratio {
       background-image: govuk-image-url("govuk-crest-2x.png");
     }
     background-repeat: no-repeat;
-    background-position: 50% 0%;
-    background-size: $govuk-footer-crest-image-width $govuk-footer-crest-image-height;
-    text-align: center;
-    white-space: nowrap;
+    background-size: contain;
+
+    @supports (mask-image: url("")) {
+      // Use masks to change the colour of the crest to match the current text color
+      // to increase the contrast of the logo
+      background: currentcolor;
+      mask-image: govuk-image-url("govuk-crest.png");
+      @include govuk-device-pixel-ratio {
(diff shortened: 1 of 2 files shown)


src/govuk/components/footer/_index.scss
```diff
@@ -106,17 +106,35 @@
 
   .govuk-footer__copyright-logo {
     display: inline-block;
-    min-width: $govuk-footer-crest-image-width;
-    padding-top: ($govuk-footer-crest-image-height + govuk-spacing(2));
+    text-align: center;
+    white-space: nowrap;
+  }
+
+  .govuk-footer__copyright-logo:before {
```
NickColley, line 113: Rather than using padding we create a pseudo element that is before the text so we can avoid abusing padding-top and give a box the proper sizing, which apart from making this technique work feels a bit nicer.

36degrees:
Hey @NickColley – thanks for raising this 🙌🏻 

It's a pretty neat solution – `mask-image` is a new one for me! Unfortunately despite using `currentColor` it relies on the element having a background, and this means that the crest becomes invisible in forced colors modes such as Windows Contrast themes where backgrounds are removed.

I've tried sticking `forced-color-adjust: none` on the `:before` pseudo-element but that doesn't seem to fix it (and as far as I know Firefox still doesn't recognise `forced-color-adjust` anyway).

Because of that, I don't think we can accept this contribution as is.

## Before

Screenshot 2022-11-08 at 10 44 57

## After

Screenshot 2022-11-08 at 10 44 54

NickColley:
When I tested high contrast modes in Windows and Firefox it removed the background regardless so I figured that was a no goer anyway, what method did you use?

If we were to do the less clever approach of increasing the original image it'd be worse in high contrast.

So that'd leave updating the HTML, which feels like something we wouldnt do for something that is an improvement but not addressing a serious barrier...

36degrees:
> When I tested high contrast modes in Windows and Firefox it removed the background regardless so I figured that was a no goer anyway, what method did you use?

I'm not doing anything fancy – just comparing http://govuk-frontend-review.herokuapp.com/components/footer/preview and https://govuk-frontend-pr-2972.herokuapp.com/components/footer/preview in Firefox and in Edge with the 'High Contrast Black' theme enabled.

NickColley:
Ooo ar, I think I changed colours manually. Okay not the one then, thanks for reviewing appreciated.

NickColley:
For future people, the reason you wouldnt use an inline SVG for this is that the crest is very complicated so it'd end up very big in comparison to PNGs... So we wouldnt want to do the same approach as the OGL logo.

PR #2979, opened 2022-11-08
Import missing `Element.prototype.dataset` polyfill

36degrees:
At the minute browsers that do not support `HTMLElement.dataset` (including IE8-10) are not correctly reading config from their data attributes, as we are no longer importing the polyfill.

Looks like this was missed as part of the refactor in 0a87bbf1eb54a636e9fdfaaa7d7a5b992c06583e.

---

The `Element.prototype.dataset` polyfill is not imported on main:

```
$ ag prototype/dataset
src/govuk/vendor/polyfills/Element/prototype/dataset.mjs
6:  // Detection from https://raw.githubusercontent.com/Financial-Times/polyfill-library/13cf7c340974d128d557580b5e2dafcd1b1192d1/polyfills/Element/prototype/dataset/detect.js
18:  // Polyfill derived from  https://raw.githubusercontent.com/Financial-Times/polyfill-library/13cf7c340974d128d557580b5e2dafcd1b1192d1/polyfills/Element/prototype/dataset/polyfill.js
```

With this change:

```
$ ag prototype/dataset
src/govuk/common/normalise-dataset.mjs
1:import '../vendor/polyfills/Element/prototype/dataset.mjs'

src/govuk/vendor/polyfills/Element/prototype/dataset.mjs
6:  // Detection from https://raw.githubusercontent.com/Financial-Times/polyfill-library/13cf7c340974d128d557580b5e2dafcd1b1192d1/polyfills/Element/prototype/dataset/detect.js
18:  // Polyfill derived from  https://raw.githubusercontent.com/Financial-Times/polyfill-library/13cf7c340974d128d557580b5e2dafcd1b1192d1/polyfills/Element/prototype/dataset/polyfill.js
```

### DIFF
--- a/src/govuk/common/normalise-dataset.mjs
+++ b/src/govuk/common/normalise-dataset.mjs
@@ -1,3 +1,4 @@
+import '../vendor/polyfills/Element/prototype/dataset.mjs'
 import '../vendor/polyfills/String/prototype/trim.mjs'
 
 /**


colinrotherham (approved):
Nicely spotted

Wonder if we can be clever and test for this with an ESLint plugin?
https://www.npmjs.com/package/eslint-plugin-compat

PR #2982, opened 2022-11-08
Migrate Gulp and Node.js build tasks to ESM

colinrotherham:
This PR upgrades our build tooling to ESM using `*.mjs` files

Why? We have a couple of packages noted in https://github.com/alphagov/govuk-frontend/pull/2937 with "ESM only" upgrades

>Intentionally ignored until we use ESM by default
>
>`del` → **`^7.0.0`** – ESM export only, we need CommonJS
>`slash` → **`^5.0.0`** – ESM export only, we need CommonJS

Both of them are flagged by dependabot:

* https://github.com/alphagov/govuk-frontend/pull/2968
* https://github.com/alphagov/govuk-frontend/pull/3003

### Additional considerations

The following Gulp tasks have been removed by name but are still part of their npm script:

`gulp scss:compile` → **`npm run compile`**
`gulp js:compile` → **`npm run compile`**
`gulp copy:assets` → **`npm run build:dist`**
`gulp copy:files` → **`npm run build:package`**

For convenience I've kept `gulp watch` but perhaps `npm start` is enough?

I've also ensured new ESM imports have no "side effects" which means moving (some) `gulp.task()` calls to **`gulpfile.mjs`** which chips away a bit more at:


gulpfile.mjs
```diff
@@ -63,18 +50,24 @@ gulp.task('compile', gulp.series(
...
-  'watch',
+  watch,
   npmScriptTask('serve', ['--silent'])
 ))
 
+/**
+ * Watch task
+ * Runs tasks when files are modified
+ */
+gulp.task('watch', watch)
```
colinrotherham, line 61: I've added this so `gulp watch` still works, do we want this?

gulpfile.mjs
```diff
@@ -1,20 +1,16 @@
...
-const { destination } = require('./tasks/task-arguments.js')
-const slash = require('slash')
+import gulp from 'gulp'
+import taskListing from 'gulp-task-listing'
 
 // Gulp sub-tasks
-require('./tasks/gulp/compile-assets.js')
-require('./tasks/gulp/copy-to-destination.js')
-require('./tasks/gulp/watch.js')
+import { compileJavaScripts, compileStylesheets } from './tasks/gulp/compile-assets.mjs'
```
colinrotherham, line 5: Are we happy losing the `gulp js:compile` and `gulp scss:compile` tasks?

They're still available via `gulp compile` which includes **compileJavaScripts()** and **compileStylesheets()**

domoscargin: I can see edge cases where somebody is working on some finicky interaction between CSS and JS, but that feels such an unlikely case, and these are not part of the "public" package json scripts, so I think we're OK to toss 'em.

gulpfile.mjs
domoscargin: I say nah. Can't think of a good enough reason to use it outside of the dev task.

domoscargin (approved):
I don't have any blocking comments - I think this makes sense, especially since we're using ESM for most of our source code.  Ran through the `npm` scripts and they seemed fine.

gulpfile.mjs
colinrotherham: Thanks @domoscargin I'll remove it, rebase, and get this merged 😊

PR #2984, opened 2022-11-08
Upgrade to `del` and `slash` "ESM only" packages

colinrotherham:
Do we want to upgrade our pinned packages with "ESM only" updates?

There's a two-step process (unfortunately) which needs:

1. https://github.com/alphagov/govuk-frontend/pull/2982
2. Temporarily allow Babel to transpile "ESM only" dependencies for Jest

We need 2) until Node.js unflags [`--experimental-vm-modules`](https://nodejs.org/api/vm.html#class-vmmodule) for [Jest **ECMAScript Modules** support](https://jestjs.io/docs/ecmascript-modules)

**Update:** Various related links

* https://github.com/facebook/jest/issues/9430
* https://github.com/nodejs/node/issues/37648
* https://github.com/alphagov/govuk-frontend/pull/2968
* https://github.com/alphagov/govuk-frontend/pull/3003

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -51,7 +51,7 @@ module.exports = {
       ...config,
       displayName: 'JavaScript behaviour tests',
       testMatch: [
-        '**/*.test.js',
+        '**/*.test.{js,mjs}',
 
         // Exclude macro/unit tests
         '!**/(*.)?template.test.{js,mjs}',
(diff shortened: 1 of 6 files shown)


colinrotherham:
Thanks @romaricpascal 

>**question** Is there any reason we wouldn't want to configure Babel for all 4 projects? Worried that we one day add one of these packages to the projects and we miss that Babel is not set up for them.

Don't mind at all. Assumed we'd only add it (temporarily) where needed—but pushed again 👍 

>**suggestion**: Is it worth having a `ESM_ONLY_PACKAGES = ['del','slash']` constant we'd build the pattern from as a central point for listing packages we want Babel to process?

As the code reviewer, would you like that? 😊 

Might not be necessary now it's only defined once at the top?

romaricpascal (approved):

Issue #2944, opened 2022-10-26
Finalise Nunjucks options and JSDoc for new config introduced in v4.4.0


36degrees:
## What

Pull the Nunjucks macro options and JSDoc for new config introduced in v4.4.0 into [the draft doc](https://docs.google.com/document/d/1ztulyMp_pxi-uJimWrKQtnfSjdDDIN4Qr9rKmAHWO_0/edit#) so that it can be reviewed by our tech writer, and then 2i'ed.

Once the changes have been reviewed, update the component YAML files and the JS files to reflect any revisions.

## Why

As the Nunjucks macro options and the JSDoc are part of our published documentation, they should be reviewed by our tech writer before we do the release.

## Who needs to know about this

Tech writers

## Done when

- [x] The draft doc has a version of the Nunjucks macro options and JSDoc approximately as they would appear when published in the Design System / Frontend Docs
- [x] The Nunjucks macro options and JSDoc has been reviewed in the draft doc by the tech writer
- [x] The Nunjucks macro options and JSDoc has been 2i'ed
- [ ] Any revisions have been reflected in the component YAML / JS

romaricpascal:
@claireashworth I've pulled the current content we have in our JSDoc inside the document, split by components and with a little intro and some notes on why I left out some parts of the JSDoc to focus on what's strictly necessary for the next release.

romaricpascal:
And the Nunjucks macro descriptions should be split as they will appear in the table.

PR #2985, opened 2022-11-08
Updates documentation of Nunjucks options

romaricpascal:
Pulls documentation of the Nunjucks options related to I18n from the draft and into each component's YAML file (half the last step of #2944, other half being #2986).

These are the following components:
- Accordion - Default values have been set in quotes rather than bold for consistency with the rest of our options documentation. 
- Character Count - Missing options have been added
- Footer - Draft listed options that have not been implemented as part of this work, so that don't appear here. Unlike other components, note that options describe their default, and don't provide their actual value as they're a bit more than just text 
- Header - Default value has been set in quotes. The `menuButtonLabel` option has also be updated to match how the button was described.

### DIFF
--- a/src/govuk/components/character-count/character-count.yaml
+++ b/src/govuk/components/character-count/character-count.yaml
@@ -37,10 +37,6 @@ params:
   required: false
   description: Options for the hint component.
   isComponent: true
-- name: textareaDescriptionText
-  type: string
-  required: false
-  description: Text describing the maximum number of characters you can enter, which is announced to screen readers. The text is displayed as a fallback if the character count JavaScript does not run. Depending on how you configure this component and what parameters you add, instances of `%{count}` are replaced by the value of `maxwords`. If not configured, it uses `maxlength`. By default, textarea description is provided in English.
 - name: errorMessage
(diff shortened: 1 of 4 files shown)


src/govuk/components/character-count/character-count.yaml
```diff
@@ -76,6 +72,34 @@ params:
     type: string
     required: false
     description: Classes to add to the count message.
+- name: textareaDescriptionText
+  type: string
+  required: false
+  description: Message made available to assistive technologies to describe that the component accepts only a limited amount of content. It is visible on the page when JavaScript is unavailable. The component will replace the `%{count}` placeholder with the value of the `maxlength` or `maxwords` parameter.
+- name: charactersUnderLimitText
+  type: string
```
36degrees, line 80: Possibly misleading for pluralised lists – should this be object? (Same for all pluralised messages)

```suggestion
  type: object
```

romaricpascal: Indeed, good catch! Got too focused on the content of the messages and forgot to check the types 😬

PR #2992, opened 2022-11-09
Enable `package.json` updates for Dependabot

colinrotherham:
We think Dependabot is only making `package-lock.json` changes rather than `package.json` increases

For example:

https://github.com/alphagov/govuk-frontend/pull/2990 – Updating `@percy/sdk-utils` (child dependency) instead of **`@percy/cli`**
https://github.com/alphagov/govuk-frontend/pull/2973 – Updating `jest-runtime` (child dependency) instead of **`jest`**

This PR adds to the **dependabot.yml** config:

```yaml
- package-ecosystem: npm
  versioning-strategy: increase
```

Related `versioning-strategy` documentation:
https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#versioning-strategy

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,6 +9,8 @@ updates:
       - alphagov/design-system-developers
     schedule:
       interval: daily
+    versioning-strategy: increase
+
     allow:
       - dependency-type: all
 


colinrotherham:
Related: https://github.com/apache/superset/pull/20920

domoscargin (approved):
Looking at that linked PR, seems like something/anything could be interfering with dear old Dependabot here, so I'm going to approve and hopefully this'll force the package.json updates.

colinrotherham:
> I'm happy to give this a go, but is this not the default behaviour anyway? (`npm` increases stuff by default)

We install using `npm ci` in the workflows (installs exactly from `package-lock.json`) so won't get increases. Keeps things reproducible and easy to cache between workflow jobs.

It looks like for (mostly) `npm update` behaviour, with updates _not_ written to `package.json` we'd want:

```yaml
- package-ecosystem: npm
  versioning-strategy: increase-if-necessary
```

But I know we [like to see updates show up in `package.json`](https://github.com/alphagov/govuk-frontend/pull/2937#issuecomment-1289107715) so I picked the "always increase" option:

```yaml
- package-ecosystem: npm
  versioning-strategy: increase
```

>Looking at that linked PR, seems like something/anything could be interfering with dear old Dependabot here, so I'm going to approve and hopefully this'll force the package.json updates.

Yep, see what happens, and tweak again 😬

PR #2994, opened 2022-11-09
Update 'Translated' example to use updated API

36degrees:


romaricpascal (approved):
Thanks for spotting this 🙌🏻

PR #2997, opened 2022-11-10
Various JSDoc + type checking fixes (for v4.4.0 only)

colinrotherham:
This PR fixes various JSDoc (and related code) issues identified in https://github.com/alphagov/govuk-frontend/pull/2971

Including things like:

1. **I18n**, **Character Count** and **Accordion** `config` param are optional
2. **I18n** `plural form `other` is optional (with caveats)
3. **I18n** alternative `i18n.${lookupKey}` translation usage removed
4. **Character Count** `config.maxlength` and `config.maxwords` are mutually exclusive
5. Various minor type fixes, test configs checked against JSDoc

All working fab here.

Screenshot showing `wordsUnderLimit` (string) not matching the [**TranslationPluralForms** (object) JSDoc type](https://govuk-frontend-pr-2997.herokuapp.com/docs/javascript/global.html#TranslationPluralForms)

JSDoc comments for wordsUnderLimit mismatch as string

### DIFF
--- a/src/govuk/all.mjs
+++ b/src/govuk/all.mjs
@@ -17,13 +17,7 @@ import Tabs from './components/tabs/tabs.mjs'
  * Use the `data-module` attributes to find, instantiate and init all of the
  * components provided as part of GOV.UK Frontend.
  *
- * @param {object} [config] - Config
- * @param {HTMLElement} [config.scope=document] - Scope to query for components
- * @param {object} [config.accordion] - Accordion config
- * @param {object} [config.button] - Button config
- * @param {object} [config.characterCount] - Character Count config
- * @param {object} [config.errorSummary] - Error Summary config
- * @param {object} [config.notificationBanner] - Notification Banner config
+ * @param {Config} [config] - Config for all components
  */
 function initAll (config) {
   config = typeof config !== 'undefined' ? config : {}
@@ -103,3 +97,15 @@ export {
   SkipLink,
   Tabs
 }
+
+/**
+ * Config for all components
+ *
+ * @typedef {object} Config
+ * @property {HTMLElement} [scope=document] - Scope to query for components
+ * @property {import('./components/accordion/accordion.mjs').AccordionConfig} [accordion] - Accordion config
+ * @property {import('./components/button/button.mjs').ButtonConfig} [button] - Button config
+ * @property {import('./components/character-count/character-count.mjs').CharacterCountConfig} [characterCount] - Character Count config
+ * @property {import('./components/error-summary/error-summary.mjs').ErrorSummaryConfig} [errorSummary] - Error Summary config
+ * @property {import('./components/notification-banner/notification-banner.mjs').NotificationBannerConfig} [notificationBanner] - Notification Banner config
+ */
--- a/src/govuk/components/accordion/accordion.mjs
+++ b/src/govuk/components/accordion/accordion.mjs
@@ -6,9 +6,12 @@ import '../../vendor/polyfills/String/prototype/trim.mjs'
 import { normaliseDataset } from '../../common/normalise-dataset.mjs'
 
 /**
+ * @constant
  * @type {AccordionTranslations}
+ * @see Default value for {@link AccordionConfig.i18n}
+ * @default
  */
-var TRANSLATIONS_DEFAULT = {
+var ACCORDION_TRANSLATIONS = {
   hideAllSections: 'Hide all sections',
   hideSection: 'Hide',
   hideSectionAriaLabel: 'Hide this section',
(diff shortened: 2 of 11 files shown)


src/govuk/components/character-count/character-count.unit.test.mjs
```diff
@@ -36,22 +36,12 @@ describe('CharacterCount', () => {
...
-          const component = new CharacterCount(document.createElement('div'), {
-            i18n: { charactersUnderLimit: { one: 'Custom text. Count: %{count}' } },
-            'i18n.charactersOverLimit.other': 'Different custom text. Count: %{count}'
-          })
-
-          expect(component.formatCountMessage(1, 'characters')).toEqual('Custom text. Count: 1')
-          expect(component.formatCountMessage(-10, 'characters')).toEqual('Different custom text. Count: 10')
-          // Other keys remain untouched
-          expect(component.formatCountMessage(10, 'characters')).toEqual('You have 10 characters remaining')
-        })
```
romaricpascal, line 49: **issue** Argh, I think there was a misunderstanding about what I was proposing to drop about that test. It was just the part about 'i18n.charactersOverLimit.other' that could be dropped as we no longer use them.

The test in itself remains relevant to check how options get merged with our defaults, I _think_. 

```javascript
        it('overrides the default translation keys', () => {
          const component = new CharacterCount(document.createElement('div'), {
            i18n: { charactersUnderLimit: { one: 'Custom text. Count: %{count}' } },
          })

          expect(component.formatCountMessage(1, 'characters')).toEqual('Custom text. Count: 1')
          
          // Other keys remain untouched
          expect(component.formatCountMessage(10, 'characters')).toEqual('You have 10 characters remaining')
        })
```

romaricpascal (commented):
Thanks for splitting it, this feels much safer to merge ahead of the release 🙌🏻 Just noticed one possible misunderstanding about a test that got removed around flat keys for the character count JS config.

src/govuk/components/character-count/character-count.unit.test.mjs
colinrotherham: Hey that's fine, don't worry, I'll stick it back in 😊

colinrotherham: Sorted 👍

colinrotherham:
@romaricpascal Think we're all sorted now

I forgot we also accept `config.accordion` and `config.characterCount` on [**all.mjs**](https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/all.mjs#L15)

So I've had to link to the JSDoc config `@typedef` blocks from there too:
https://govuk-frontend-pr-2997.herokuapp.com/docs/javascript/global.html#initAll

src/govuk/all.mjs
```diff
@@ -103,3 +97,15 @@ export {
...
+ * Config for all components
+ *
+ * @typedef {object} Config
+ * @property {HTMLElement} [scope=document] - Scope to query for components
+ * @property {import('./components/accordion/accordion.mjs').AccordionConfig} [accordion] - Accordion config
+ * @property {import('./components/button/button.mjs').ButtonConfig} [button] - Button config
+ * @property {import('./components/character-count/character-count.mjs').CharacterCountConfig} [characterCount] - Character Count config
+ * @property {import('./components/error-summary/error-summary.mjs').ErrorSummaryConfig} [errorSummary] - Error Summary config
+ * @property {import('./components/notification-banner/notification-banner.mjs').NotificationBannerConfig} [notificationBanner] - Notification Banner config
+ */
```
romaricpascal, line 111: Oh yeah, we were so focused on the individual components that `all` was completely neglected 😬 Good spot! 🙌🏻

src/govuk/components/accordion/accordion.mjs
```diff
@@ -31,13 +31,11 @@ var TRANSLATIONS_DEFAULT = {
  *
  * @class
  * @param {HTMLElement} $module - HTML element to use for accordion
- * @param {object} config - Accordion config
- * @param {AccordionTranslations} config.i18n - Translations
+ * @param {AccordionTranslations} [config] - Accordion config
```
romaricpascal, line 34:
```suggestion
 * @param {AccordionConfig} [config] - Accordion config
```

romaricpascal (commented):
Little mix-up when replacing the AccordionConfig, but all good to go otherwise 😄

src/govuk/components/accordion/accordion.mjs
colinrotherham: Brilliant spot. All fixed and ready for you again 😊

colinrotherham:
See the [Generated JSDoc config](https://govuk-frontend-pr-2997.herokuapp.com/docs/javascript/) served by the review app

romaricpascal (approved):
🚀

PR #2998, opened 2022-11-10
Refactor back link and breadcrumb chevrons to use ems

### LINKED ISSUE #1717: Back link's arrow and border do not scale
## Problem

If you increase the browser's font size the Back Link component's arrow and border do not scale with the copy.

before

Screenshot from Chromium Edge v79 on Mac with font size set to large in edge://settings/fonts

image

This is caused by the css-arrow and border being set in px

## Solution

Replace/override px values with em values so they scale in proportion to the copy. 

after

querkmachine:
Currently the back link is measured in pixels. This prevents the arrow from changing size when text zooming is used. Changing this to em units allows the arrow to scale along with the text.

em units have been calculated by taking the previous pixel value and dividing it by the text's pixel size (16). 
e.g. 7px ÷ 16px = 0.4375em

Fixes #1717. 

||100% zoom|400% zoom|
|:-|:-:|:-:|
|Before|Screenshot 2022-11-10 at 11 51 45|Screenshot 2022-11-10 at 11 51 53|
|After|Screenshot 2022-11-10 at 11 51 19|Screenshot 2022-11-10 at 11 51 11|

A side effect of this is that the chevron now also scales _down_ on small screens, where text is displayed at 14px, which it didn't do previously.

||Narrow view|
|:-|:-:|
|Before|Screenshot 2022-11-10 at 16 26 34|
|After|Screenshot 2022-11-10 at 16 26 26|

Screenshots from Firefox 106 on macOS 12.6.

Also tested in Chrome 107, Safari 16.0, and Internet Explorers 9 through 11.

### DIFF
--- a/src/govuk/components/back-link/_index.scss
+++ b/src/govuk/components/back-link/_index.scss
@@ -7,8 +7,6 @@
   $chevron-size: govuk-em(7px, $font-size);
 
   // Size of chevron border
-  // Awkward syntax hack to ensure we're using the CSS max() function and not
-  // the Sass one. This can be called normally once we're using Sass modules.
   $chevron-border-width: govuk-em(1px, $font-size);
 
   // Colour of chevron
--- a/src/govuk/components/back-link/_index.scss
+++ b/src/govuk/components/back-link/_index.scss
@@ -1,19 +1,19 @@
 @include govuk-exports("govuk/component/back-link") {
   // Component font-size on the Frontend (used for calculations)
-  $font-scale-size: 16;
-  $font-size: $font-scale-size * 1px;
+  $font-size: 16;
 
   // Size of chevron (excluding border)
   $chevron-size: govuk-em(7px, $font-size);
 
   // Size of chevron border
-  $chevron-border-width: govuk-em(1px, $font-size);
+  $chevron-border-min-width: 1px;
+  $chevron-border-width: govuk-em($chevron-border-min-width, $font-size);
 
   // Colour of chevron
   $chevron-border-colour: $govuk-secondary-text-colour;
 
   .govuk-back-link {
-    @include govuk-typography-responsive($size: $font-scale-size);
+    @include govuk-typography-responsive($size: $font-size);
     @include govuk-link-common;
     @include govuk-link-style-text;
 
@@ -56,8 +56,12 @@
     transform: rotate(225deg);
 
     border: solid;
-    border-width: $chevron-border-width $chevron-border-width 0 0;
+    border-width: $chevron-border-min-width $chevron-border-min-width 0 0;
     border-color: $chevron-border-colour;
+
+    @supports (border-width: unquote("max(0px)")) {
+      border-width: unquote("max(#{$chevron-border-min-width}, #{$chevron-border-width}) max(#{$chevron-border-min-width}, #{$chevron-border-width})") 0 0;
+    }
 
     // Fall back to a less than sign for IE8
     @include govuk-if-ie8 {
--- a/src/govuk/components/back-link/_index.scss
+++ b/src/govuk/components/back-link/_index.scss
@@ -1,10 +1,10 @@
 @include govuk-exports("govuk/component/back-link") {
 
   // Size of chevron (excluding border)
-  $chevron-size: 7px;
+  $chevron-size: .4375em;
 
   // Size of chevron border
-  $chevron-border-width: 1px;
+  $chevron-border-width: .0625em;
 
   // Colour of chevron
   $chevron-border-colour: $govuk-secondary-text-colour;
@@ -21,7 +21,7 @@
     margin-bottom: govuk-spacing(3);
 
     // Allow space for the arrow
-    padding-left: 14px;
+    padding-left: .875em;
   }
 
   // Prepend left pointing chevron
--- a/src/govuk/components/back-link/_index.scss
+++ b/src/govuk/components/back-link/_index.scss
@@ -61,6 +61,9 @@
 
     @supports (border-width: unquote("max(0px)")) {
       border-width: unquote("max(#{$chevron-border-min-width}, #{$chevron-border-width}) max(#{$chevron-border-min-width}, #{$chevron-border-width})") 0 0;
+
+      // Ensure that the chevron never gets smaller than 16px
+      font-size: unquote("max(#{$font-size * 1px}, 1em)");
     }
 
     // Fall back to a less than sign for IE8
--- a/src/govuk/components/back-link/_index.scss
+++ b/src/govuk/components/back-link/_index.scss
@@ -1,16 +1,21 @@
 @include govuk-exports("govuk/component/back-link") {
+  // Component font-size on the Frontend (used for calculations)
+  $font-scale-size: 16;
+  $font-size: $font-scale-size * 1px;
 
   // Size of chevron (excluding border)
-  $chevron-size: .4375em;
+  $chevron-size: govuk-em(7px, $font-size);
 
   // Size of chevron border
-  $chevron-border-width: .0625em;
+  // Awkward syntax hack to ensure we're using the CSS max() function and not
+  // the Sass one. This can be called normally once we're using Sass modules.
+  $chevron-border-width: govuk-em(1px, $font-size);
 
   // Colour of chevron
   $chevron-border-colour: $govuk-secondary-text-colour;
 
   .govuk-back-link {
-    @include govuk-typography-responsive($size: 16);
+    @include govuk-typography-responsive($size: $font-scale-size);
     @include govuk-link-common;
     @include govuk-link-style-text;
 
@@ -21,7 +26,7 @@
     margin-bottom: govuk-spacing(3);
 
     // Allow space for the arrow
-    padding-left: .875em;
+    padding-left: govuk-em(14px, $font-size);
   }
 
   // Prepend left pointing chevron
@@ -35,14 +40,15 @@
     @if $govuk-use-legacy-font {
       // Begin adjustments for font baseline offset
       // These should be removed when legacy font support is dropped
-      top: -1px;
-      bottom: 1px;
+      $offset: govuk-em(1px, $font-size);
+      top: $offset * -1;
+      bottom: $offset;
     } @else {
       top: 0;
       bottom: 0;
     }
 
-    left: 3px;
+    left: govuk-em(3px, $font-size);
 
     width: $chevron-size;
     height: $chevron-size;
@@ -80,16 +86,5 @@
     right: 0;
     bottom: -14px;
     left: 0;
-  }
-
-  @if $govuk-use-legacy-font {
-    // Begin adjustments for font baseline offset
-    // These should be removed when legacy font support is dropped
-    .govuk-back-link:before {
-      $offset: 1px;
-
-      top: $offset * -1;
-      bottom: $offset;
-    }
   }
 }
--- a/src/govuk/components/breadcrumbs/_index.scss
+++ b/src/govuk/components/breadcrumbs/_index.scss
@@ -83,6 +83,9 @@
 
       @supports (border-width: unquote("max(0px)")) {
         border-width: unquote("max(#{$chevron-border-min-width}, #{$chevron-border-width}) max(#{$chevron-border-min-width}, #{$chevron-border-width})") 0 0;
+
+        // Ensure that the chevron never gets smaller than 16px
+        font-size: unquote("max(#{$font-size * 1px}, 1em)");
       }
 
       // Fall back to a greater than sign for IE8
--- a/src/govuk/components/breadcrumbs/_index.scss
+++ b/src/govuk/components/breadcrumbs/_index.scss
@@ -1,13 +1,13 @@
 @include govuk-exports("govuk/component/breadcrumbs") {
   // Component font-size on the Frontend (used for calculations)
-  $font-scale-size: 16;
-  $font-size: $font-scale-size * 1px;
+  $font-size: 16;
 
   // Size of chevron (excluding border)
   $chevron-size: govuk-em(7px, $font-size);
 
   // Size of chevron border
-  $chevron-border-width: govuk-em(1px, $font-size);
+  $chevron-border-min-width: 1px;
+  $chevron-border-width: govuk-em($chevron-border-min-width, $font-size);
 
   // Colour of chevron
   $chevron-border-colour: $govuk-secondary-text-colour;
@@ -19,7 +19,7 @@
   $chevron-altitude-calculated: govuk-em(5.655px, $font-size);
 
   .govuk-breadcrumbs {
-    @include govuk-font($size: $font-scale-size);
+    @include govuk-font($size: $font-size);
     @include govuk-text-colour;
 
     margin-top: govuk-spacing(3);
@@ -78,8 +78,12 @@
       transform: rotate(45deg);
 
       border: solid;
-      border-width: $chevron-border-width $chevron-border-width 0 0;
+      border-width: $chevron-border-min-width $chevron-border-min-width 0 0;
       border-color: $chevron-border-colour;
+
+      @supports (border-width: unquote("max(0px)")) {
+        border-width: unquote("max(#{$chevron-border-min-width}, #{$chevron-border-width}) max(#{$chevron-border-min-width}, #{$chevron-border-width})") 0 0;
+      }
 
       // Fall back to a greater than sign for IE8
       @include govuk-if-ie8 {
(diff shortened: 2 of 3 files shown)


src/govuk/components/back-link/_index.scss
```diff
@@ -1,10 +1,10 @@
 @include govuk-exports("govuk/component/back-link") {
 
   // Size of chevron (excluding border)
-  $chevron-size: 7px;
+  $chevron-size: .4375em;
```
colinrotherham, line 4: This looks brilliant 🙌 

Should we use `$govuk-root-font-size` just in case users have overridden the 16px root size?

Clearer that we're expressing 7px as ems too

```suggestion
  $chevron-size: govuk-em(7px, $govuk-root-font-size)
```

querkmachine: Clarifying how the em values were created is a good shout, but I'm not sure that `$govuk-root-font-size` is the right thing to derive it from.

The em values here are with respect to the font-size being defined by the mixin on L13, which on browsers that don't support rem is *always* 16px, regardless of the value of `$govuk-root-font-size`. In that situation, a `$govuk-root-font-size` that is larger than 16px would result in an arrow that is too large relative to the link text.

However, in terms of browsers we still support that's pretty much only an IE 8 problem, which doesn't even use this graphical chevron, so this is probably a safe enough route to take.

colinrotherham: No worries, it looked like the [L13](https://github.com/alphagov/govuk-frontend/blob/kg-back-link-text-zoom/src/govuk/components/back-link/_index.scss#L13) mixin calls in to `govuk-px-to-rem()`

Wanted to make sure the **Back link** text didn't scale up/down with `$govuk-root-font-size`—but not the arrow

Still great to see though 😊

querkmachine: In testing this out, I've noticed that at no point is `$govuk-root-font-size` actually applied to the root `<html>` element of the page. 

This means that users must manually set a matching `font-size` on `<html>` if they change `$govuk-root-font-size`, otherwise the resulting text will be sized incorrectly, demonstrated in the middle option in the table below, which is supposed to be the rem equivalent of 16px.

|`$govuk-root-font-size: 16px`<br>`html { font-size: 16px }`|`$govuk-root-font-size: 28px`<br>`html { font-size: 16px }`|`$govuk-root-font-size: 28px`<br>`html { font-size: 28px }`|
|:-:|:-:|:-:|
|Screenshot 2022-11-10 at 15 55 57|Screenshot 2022-11-10 at 15 56 06|Screenshot 2022-11-10 at 15 56 22|

Not really related to this PR, but also something that doesn't seem correct to my eyes. Is there any background to this, @36degrees?

36degrees: Have you seen the [Sass docs for `$govuk-root-font-size`](https://frontend.design-system.service.gov.uk/sass-api-reference/#govuk-root-font-size) already?

From memory, this is working as designed – we intentionally don't set the font size on the `html` element and rely on the browser default (which is nearly always 16px), but if the user _has_ for some reason done so (and set it to something other than 16px) then they might need to adjust `$govuk-root-font-size`.

romaricpascal: Is the root font-size only meant to be used for computation in Sass? Setting it (in px or other absolute units) on the `html` or `body` would mess with people configuring custom font-size in their browsers settings I believe 🤔

querkmachine: @36degrees: I hadn't. I suppose it makes more sense if it's more of a compatibility thing than a stylistic one (i.e. a user setting it to 19px because they want everything to be 20% larger than the 16px default).

@romaricpascal: It seems like it is only used for Sass computations, yeah. AFAIK the accessibility issues with setting absolute font-size values on `html` and `body` is confined to older browsers by now—newer ones will override or zoom them anyway.

36degrees: Effectively the setting is 'Tell GOV.UK Frontend what the root font size _is_' rather than 'Tell GOV.UK Frontend what you want the root font size to be', if that makes sense at all.

colinrotherham (commented):
Thanks @querkmachine

Looking good 🎉 

I've got some suggestions if they make sense?

1. **Keep the `border-width` at 1px** or set `min(1px, …)`
On small screens the 14px font size gives a faint 0.875px `0.0625em` border

2. **Use ems for the arrow position offset**
We still set a fixed `left: 3px`

3. **Use ems for the top/bottom pixel offset**
Only used with `$govuk-use-legacy-font`

Regarding 1) I've noticed we [use exact pixel border-widths](https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/settings/_measurements.scss#L53) everywhere.

Is this reason enough to continue with `$chevron-border-width: 1px` for now? Perhaps holding off until other borders are also ready to adopt ems too?

That said, the "Start now" chevron icon (albeit SVG) scales already

querkmachine:
Whilst it dipping below 1px equivalent on smaller screens is a concern (and something that should be fixed) I don't think enforcing a 1px border on all sizes makes sense here. The disparity between the visual weight of the arrow and the text starts to look pretty goofy at higher zoom levels, and the lack of relative contrast probably makes this a usability issue. 

image

I also think our existing use of absolute pixel borders doesn't apply in this instance—as here we're using borders to graphically hack together a chevron icon, rather than as an actual border around something. I've made it so that the arrow thickness is clamped to a minimum of 1px on browsers that support the CSS clamping functions.

I've implemented the two other suggestions. Notably, there seemed to be duplicated `top` and `bottom` declarations being made if `$govuk-use-legacy-font` was enabled, so I've removed that too.

colinrotherham:
Fab, thanks @querkmachine

> Whilst it dipping below 1px equivalent on smaller screens is a concern (and something that should be fixed) I don't think enforcing a 1px border on all sizes makes sense here. The disparity between the visual weight of the arrow and the text starts to look pretty goofy at higher zoom levels, and the lack of relative contrast probably makes this a usability issue.

Yep totally agree as an end goal, just wondered if we'd want to delay it until we fix all borders/outlines/shadows at the same time? Are we happy with the chevron border being scaled relative to the root font size but these ones not?

```scss
$govuk-page-width: 960px !default;
$govuk-gutter: 30px !default;
$govuk-border-width: 5px !default;
$govuk-border-width-wide: 10px !default;
$govuk-border-width-narrow: 4px !default;
$govuk-border-width-form-element: 2px !default;
$govuk-focus-width: 3px !default;
$govuk-hover-width: 10px !default;
```

Regarding `max()` escaping, do you think we should [stick with the `unquote("max(…)")` approach here](https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/objects/_width-container.scss#L36)?

```scss
// Use max() to pick largest margin, default or with safe area
// Escaped due to Sass max() vs. CSS native max()
margin-right: unquote("max(#{$govuk-gutter-half}, #{$gutter-safe-area-right})");
margin-left: unquote("max(#{$govuk-gutter-half}, #{$gutter-safe-area-left})");
```

src/govuk/components/back-link/_index.scss
```diff
@@ -35,14 +42,15 @@
     @if $govuk-use-legacy-font {
       // Begin adjustments for font baseline offset
       // These should be removed when legacy font support is dropped
-      top: -1px;
-      bottom: 1px;
+      $offset: govuk-em(1px, $govuk-root-font-size);
+      top: $offset * -1;
+      bottom: $offset;
     } @else {
       top: 0;
```
colinrotherham, line 49: @querkmachine You've got me looking at this more now 🙈 

Do you think we should let browsers calculate the vertical position instead of assuming `top: 0`?

For example, applying the old 50% vertical offset trick puts the chevron about 1px lower on mobile. I've added an example from macOS Safari 16.1 with 2x `window.devicePixelRatio`—looks better aligned?

1st animation frame — original `top: 0`
2nd animation frame — 50% offset (suggestion)

Back link animation showing 1px vertical shift

```suggestion
      top: 50%;
      margin-top: $chevron-size / -2
```

querkmachine:
> Yep totally agree as an end goal, just wondered if we'd want to delay it until we fix all borders/outlines/shadows at the same time?

I don't see any reason to delay it. The implementation here is pretty much entirely self-contained to the Back Link and Breadcrumbs components. Expanding the scope of the initial issue to cover changing every fixed pixel value in the project doesn't seem useful.

> Are we happy with the chevron border being scaled relative to the root font size but these ones not?

I'm kinda iffy with these chevrons being scaled based on the root font-size in general. It produces some undesirable outcomes when `$govuk-root-font-size` is changed, when compared to using em units local to the component. 

Using the methodology in my initial commit (though using `govuk-em` rather than hardcoding values) produces more consistent results in that instance.

||`$govuk-root-font-size: 16px`<br>`html { font-size: 16px }`|`$govuk-root-font-size: 28px`<br>`html { font-size: 28px }`|
|:-|:-:|:-:|
|ems using root font-size[^1]|Screenshot 2022-11-11 at 14 20 27|Screenshot 2022-11-11 at 14 21 25|
|ems using component font-size[^2]|Screenshot 2022-11-11 at 14 29 17|Screenshot 2022-11-11 at 14 29 01|

> Regarding `max()` escaping, do you think we should [stick with the `unquote("max(…)")` approach here](https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/objects/_width-container.scss#L36)?

Yeah, that makes more sense. Will update. 

[^1]: Using `govuk-em(pixel-value, $govuk-root-font-size)`.
[^2]: Equivalent to `govuk-em(pixel-value, 16px)`, based on 16px being the 'default' font-size of the Back Link component.

src/govuk/components/back-link/_index.scss
querkmachine: This might be a question for the designers, but I _think_ the intent is that the chevron is aligned with the text baseline, rather than being exactly vertically centred?

colinrotherham:
Yeah component font size is 100% what we want (using ems)

But the 16px is adjusted against `$govuk-root-font-size` in `govuk-px-to-rem()` via `govuk-typography-responsive()`

So you'll want a tiny tweak to your example

### Before
Component font size only
```scss
govuk-em(pixel-value, 16px)
```

### After
Component font size adhering to (potential) `$govuk-root-font-size` override
```scss
govuk-em(pixel-value, 16px / (16px / $govuk-root-font-size))
```

Feel free to tidy it up, plus 14px small screen variations too 🙈

src/govuk/components/back-link/_index.scss
colinrotherham: Yeah best check, Percy shows it well?

As in, small screens are now (correctly) showing a smaller arrow but it's too high:

Back link animation showing 1px vertical shift (Percy)

querkmachine:
Sorry, but I'm failing to see why we need to concern ourselves with what `$govuk-root-font-size`, `govuk-px-to-rem()` or `govuk-typography-responsive()` are doing.

The text of the back link text is _always_ 14px/16px. That number is converted into a rem-unit equivalent for browsers that support it, but no matter what `$govuk-root-font-size` is set to, the outcome is always a font-size that is 14px/16px in the browser's rendering engine.

We are then separately using em-units to scale the arrow relative to the back link text—we are always scaling relative to that computed 14px/16px, and (if we stick to basing it on the 16px scale) the resulting conversion will always be 1px = 0.0625em. What we're doing with em units is entirely independent of `$govuk-root-font-size`, from what I can gather.

colinrotherham:
Don't worry, I was thinking we were going down the mapped `$size: 16` typography scale route

Doubt many people override `$govuk-typography-scale` do they? If we're using `16px` literal pixels, all good!

Here's what I was chucking into the review app's `app.scss`

```scss
html {
  font-size: 62.5%;
}

$govuk-suppressed-warnings: (
  "allow-not-using-rem",
  "compatibility-mode"
);

$govuk-typography-use-rem: false;
$govuk-root-font-size: 10px;
```

All working nicely with `govuk-em(14px, 16px)` 👍

36degrees:
Please don't merge until after v4.4.0 has been released 🚀

src/govuk/components/back-link/_index.scss
```diff
@@ -1,16 +1,19 @@
 @include govuk-exports("govuk/component/back-link") {
+  // Component font-size on the Frontend (used for calculations)
+  $font-scale-size: 16;
+  $font-size: $font-scale-size * 1px;
```
colinrotherham, line 4: I think we can go "unitless" with `govuk-em(7, 16)` if that's nicer?

src/govuk/components/breadcrumbs/_index.scss
```diff
@@ -13,10 +16,10 @@
   // of length 8 (7px + 1px border):
   //
   // √(8² + 8²) * 0.5 ≅ 5.655
-  $chevron-altitude-calculated: 5.655px;
+  $chevron-altitude-calculated: govuk-em(5.655px, $font-size);
```
colinrotherham, line 19: Brilliant 🤓

src/govuk/components/back-link/_index.scss
```diff
@@ -1,16 +1,19 @@
...
 
   // Size of chevron (excluding border)
-  $chevron-size: 7px;
+  $chevron-size: govuk-em(7px, $font-size);
 
   // Size of chevron border
-  $chevron-border-width: 1px;
+  $chevron-border-width: govuk-em(1px, $font-size);
```
colinrotherham, line 10: Are we back to having the thinner-than-1px border now?

querkmachine: Unfortunately yes. The method using `max` didn't work out for the Breadcrumbs component, due to some of the ways it tries to use the value of 

As the chevron is intended to be visually identical between the Breadcrumbs and the Back Link, and how often these elements appear next to one another, being consistently too narrow seemed better than being inconsistent.

Granted, I may have just thought of a way around this, so I'm gonna give that a shot now 🤔

src/govuk/components/back-link/_index.scss
querkmachine: Good catch. `gov-em` internally converts it to a pixel value anyway, so saves us a step here, I guess!

colinrotherham:
@querkmachine Did we get design feedback on the smaller chevron appearing too high in https://github.com/alphagov/govuk-frontend/pull/2998#discussion_r1020302540?

Percy [shows the visual diff](https://percy.io/55ba5e1a/govuk-frontend/builds/22910651?utm_campaign=55ba5e1a&utm_content=govuk-frontend&utm_source=github_status_public)

Perceptual alignment

querkmachine:
@colinrotherham The chevon now maintains a minimum size on browsers that support `max()`, for consistency with the existing components. 

The spacing between breadcrumb items is still different to previous (being em-based), but the total difference is a couple of pixels at most.

colinrotherham:
> @colinrotherham The chevon now maintains a minimum size on browsers that support `max()`, for consistency with the existing components.

Brill, thanks 🎉 

> The spacing between breadcrumb items is still different to previous (being em-based), but the total difference is a couple of pixels at most.

On this one, the [Percy build](https://percy.io/55ba5e1a/govuk-frontend/builds/23263131/changed/1299051734?browser=edge&browser_ids=25%2C30&subcategories=unreviewed%2Cchanges_requested&viewLayout=overlay&viewMode=new&width=375&widths=375%2C1280) shows a bigger change than our [**Diff sensitivity** setting](https://docs.percy.io/docs/diff-sensitivity)

That said, I'm happy that the spacing is consistent with the font size. Let's check again with @Ciandelle as consistency here might apply to the horizontal spacing too

Spacing changes

Ciandelle:
Looks great to me, I don't believe that the horizontal spacing needs to be looked at but worth running past @christopherthomasdesign

christopherthomasdesign:
I mean I think it looks fine and doesn't make a huge difference either way. 

Sorry I know this is annoying cos I've only just swooped into this... just looking at the horizontal spacing in the preview, where does `padding-left: 0.97844em` come from? Seems a v specific number, and outputs 15.655px. Mostly just curious really – you could make an equally good argument for that spacing to be 16px (cos it's 1em) or 15px (cos it fits with our spacing scale).

querkmachine:
@christopherthomasdesign 

> Sorry I know this is annoying cos I've only just swooped into this... just looking at the horizontal spacing in the preview, where does `padding-left: 0.97844em` come from? Seems a v specific number, and outputs 15.655px. Mostly just curious really – you could make an equally good argument for that spacing to be 16px (cos it's 1em) or 15px (cos it fits with our spacing scale).

Maths 🙈

The padding calculated from 2 on the spacing scale (10px) plus the "chevron altitude" of 5.655px ([calculated with some funky trigonometry here](https://github.com/alphagov/govuk-frontend/blob/f721773f77b598b3ffcf30fb86cc56a0ac0900eb/src/govuk/components/breadcrumbs/_index.scss#L15-L19)), which totals 15.655px. This is then converted to ems with 16px as the base font-size, resulting in the highly-specific padding value we have currently.

christopherthomasdesign:
Gotcha, thanks @querkmachine 

Anyways, all looks good to me 👍

colinrotherham (approved):
Fab thorough work here 🎉

Issue #2886, opened 2022-09-28
Split project into separate app, config, lib and tasks packages


colinrotherham:
## What

We'd like to start including our JavaScript the same way our users do. We currently have no automated browser tests that load minified JavaScript (or CSS) from [`./dist`](https://github.com/alphagov/govuk-frontend/tree/main/dist) or include [**govuk-frontend**](https://www.npmjs.com/package/govuk-frontend) from [`./package`](https://github.com/alphagov/govuk-frontend/tree/main/package) like our users

Although our build tasks copy the same files:

1. We build things twice in both `./public` and `./package`
3. We don't load our [**govuk-frontend** `import` entry point](https://github.com/alphagov/govuk-frontend/blob/main/package/package.json#L10) via ESM
4. We don't load our [**govuk-frontend** `require` entry point](https://github.com/alphagov/govuk-frontend/blob/main/package/package.json#L11) via CommonJS

The "Review application" should run separately using [**govuk-frontend**](https://www.npmjs.com/package/govuk-frontend) from [`./package`](https://github.com/alphagov/govuk-frontend/tree/main/package) like our users

>**Reading material**
>Node.js [Package entry points](https://nodejs.org/api/packages.html#package-entry-points)
>(As used in [`./package/package.json`](https://github.com/alphagov/govuk-frontend/blob/main/package/package.json#L7))

### The Review application

For background, we currently have a Node.js [Express](https://expressjs.com) "Review application" in the [`./app`](https://github.com/alphagov/govuk-frontend/tree/main/app) directory
(Heroku deploys it to https://govuk-frontend-review.herokuapp.com)

* It loads components from [`./src/govuk/components`](https://github.com/alphagov/govuk-frontend/tree/main/src/govuk/components)
* It loads fonts and images [`./src/govuk/assets`](https://github.com/alphagov/govuk-frontend/tree/main/src/govuk/assets)
* It loads JavaScript and CSS from `./public` (build required, unminified only)

## Why

It's important to run our browser tests on the same files our users do (same exports, polyfills, minification). By enabling [source maps](https://firefox-source-docs.mozilla.org/devtools-user/debugger/how_to/use_a_source_map/index.html) developers can work on "the source" even when browsers are running a minified bundle:

* https://github.com/alphagov/govuk-frontend/pull/3023

[...]

PR #3004, opened 2022-11-11
Set up review app `package.json` (12MB install)

colinrotherham:
PR to split the review app into its own npm workspace:

### Dependencies
We currently have 200MB+ of `node_modules` via [**actions/cache@v3**](https://github.com/actions/cache#readme)—added in https://github.com/alphagov/govuk-frontend/commit/1cbc990803c0c27696b90ea96dd5f707fbcae910

```console
Received 234121571 of 234121571 (100.0%), 81.0 MBs/sec
Cache Size: ~223 MB (234121571 B)
```

Sadly due to `tar.exe` performance on Windows we've seen performance issues in:

* https://github.com/alphagov/govuk-frontend/pull/2945

What else do we know?

1. **All packages are deployed to Heroku due to `"NPM_CONFIG_PRODUCTION": "false"`**
This PR makes initial steps to split out `dependencies` from `devDependencies`

2. **Each GitHub Action job restores all 200MB+ of `node_modules`**
Does every test need every package?

## Review app improvements
Once merged, the review app only needs 12MB of `node_modules`:

```console
npm ci --workspace app --omit=dev
npm start --workspace app
```

Does this solve the large delay at Heroku start up? Maybe!

**Update:** Including the build tools for `npm run heroku` takes us up to ~12MB~ 62.8MB


app/middleware/vendor.js
```diff
@@ -0,0 +1,27 @@
+const { dirname, join } = require('path')
+const express = require('express')
+
+const router = express.Router()
+
+// Locate packages
+const pkgHTML5Shiv = dirname(require.resolve('html5shiv/package.json'))
```
romaricpascal, line 7: `require.resolve` make things pretty neat! 🙌🏻 Is is worth a little `packagePath(packageName)` helper?

app/app.js
```diff
@@ -78,22 +78,10 @@ module.exports = async (options) => {
...
-  app.use('/vendor/html5-shiv/', express.static('node_modules/html5shiv/dist/'))
-
-  // serve legacy code from node modules
-  app.use('/vendor/govuk_template/', express.static('node_modules/govuk_template_jinja/assets/'))
-  app.use('/vendor/govuk_frontend_toolkit/assets', express.static('node_modules/govuk_frontend_toolkit/images'))
-  app.use('/vendor/govuk_frontend_toolkit/', express.static('node_modules/govuk_frontend_toolkit/javascripts/govuk/'))
-  app.use('/vendor/jquery/', express.static('node_modules/jquery/dist'))
-
-  app.use('/assets', express.static(configPaths.assets))
+  app.use('/vendor', middleware.vendor)
```
romaricpascal, line 84: Tidy! 😍

romaricpascal (commented):
Looking tidy! Does the change speed up the heroku startup in the end (the question is left hanging in your description... suspense 😱 )?

Might be worth checking:
- with the team if anyone has muscle memory or personal scripts using `npm serve` from the root, but I suspect most of us `npm start` 😄 
- why we were not caching `node_modules` on Heroku before re-enabling it

app/middleware/vendor.js
colinrotherham: Done, just for you

colinrotherham:
Thanks @romaricpascal

To prevent any breaking changes I've put the `npm run serve` command back at root (but it calls the workspace)

```json
"serve": "npm run serve --workspace app"
```

Regarding Heroku time savings I'm not sure yet!

I've enabled both of these though as the **app.json** changes didn't make any difference. Perhaps it takes the defaults from the `govuk-frontend-review` app (or GitHub repo `main` integration) when they're not set:

```console
heroku config:set NODE_MODULES_CACHE=true -a govuk-frontend-pr-3004
heroku config:set NPM_CONFIG_PRODUCTION=true -a govuk-frontend-pr-3004
```

It must help as we've gone from 200MB+ down to 62.8MB (12MB without the build tools at project root)

```console
-----> Pruning devDependencies
       Skipping because NPM_CONFIG_PRODUCTION is 'true'
       
-----> Build succeeded!
-----> Discovering process types
       Procfile declares types -> web
-----> Compressing...
       Done: 62.8M
-----> Launching...
       Released v14
       https://govuk-frontend-pr-3004.herokuapp.com/ deployed to Heroku
```

colinrotherham:
Once this merges we may need to run (somewhere)

```console
heroku config:unset NODE_MODULES_CACHE
heroku config:unset NPM_CONFIG_PRODUCTION
```

If that brings us back to the defaults for our Node.js buildpack

colinrotherham:
Some quick probably-not-statistically-correct stats:

* https://github.com/alphagov/govuk-frontend/pull/3004
**Heroku build to deployment time:** 39 seconds 🚀 

Versus some recent PRs:

* https://github.com/alphagov/govuk-frontend/pull/3014
**Heroku build to deployment time:** 1 minute 28 seconds

* https://github.com/alphagov/govuk-frontend/pull/3015
**Heroku build to deployment time:** 1 minute 24 seconds

* https://github.com/alphagov/govuk-frontend/pull/3020
**Heroku build to deployment time:** 1 minute 14 seconds

Pretty good though?

romaricpascal (approved):

Issue #2864, opened 2022-09-20
Only run visual regression tests if there are code changes that might cause a visual change


36degrees:
## What

Only run Percy's visual regression tests if there are code changes that might cause a visual change.

As a (non-exhaustive) example, do not run visual regression tests if the only files changed in a PR are tests, docs, internal tooling, or in the dist and package folder.

## Why

We have a limited number of screenshots available to us in any given billing period. Only running the visual regression tests when they are useful helps us to burning through our included screenshots too quickly, allowing us to remain on a cheaper plan.

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [ ] Percy's visual regression tests are only run if there are code changes that might cause a visual change

colinrotherham:
Some suggestions from me:

1. **Adding Jest `--changedSince` flag** (via commit hash)
Percy has `PERCY_PARTIAL_BUILD` and we could skip components that were already tested

2. **Adding [GitHub workflow path filters](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-paths)**
So only component Nunjucks/JavaScript (or their tests) trigger Percy

colinrotherham:
We're also running the [**Tests** workflow](https://github.com/alphagov/govuk-frontend/actions/workflows/tests.yml) twice for every push to a pull request

[**.github/workflows/tests.yml**](https://github.com/alphagov/govuk-frontend/blob/main/.github/workflows/tests.yml#L3)
Triggers on pushes twice if a pull request is also open
```yaml
on: [push, pull_request]
```

36degrees:
I think this closed by #2878.

colinrotherham:
Going to reopen this one as we've still running Percy for code changes that don't affect component UI

PR #3009, opened 2022-11-14
Send Percy screenshots via npm script (not Jest)

colinrotherham:
This PR moves our Percy screenshots out of Jest following a recent conversation

>**Discussion regarding #2864**: “I was originally thinking of trying to split the Percy screenshotting out from Jest as the screenshot ‘tests’ aren’t really tests at all.”

This has added another big speed up to our checks

Regression checks via GitHub Actions workflow


.github/workflows/tests.yml
```diff
@@ -140,7 +140,13 @@ jobs:
 
           - description: JavaScript component tests
             name: test-component
-            run: npx percy exec -- jest --color --selectProjects "JavaScript component tests"
+            run: npx jest --color --maxWorkers=2 --selectProjects "JavaScript component tests"
```
colinrotherham, line 143: Without the CPU load from the [@percy/puppeteer](https://docs.percy.io/docs/puppeteer) Chromium tabs we can now run **JavaScript component tests** with 2x workers

We use the same flag for **JavaScript behaviour tests** above

.github/workflows/tests.yml
```diff
@@ -140,7 +140,13 @@ jobs:
...
+
+          - description: Send screenshots to Percy
+            name: test-component
+            run: |
+              npm run serve &
```
colinrotherham, line 148: Starts the review app in the background using `&`

.github/workflows/tests.yml
```diff
@@ -162,3 +168,4 @@ jobs:
         run: ${{ matrix.run }}
         env:
           PERCY_TOKEN: ${{ secrets.PERCY_TOKEN }}
+          PORT: 8888
```
colinrotherham, line 171: Review app uses port `3000` but our [Puppeteer helpers want `8888`](https://github.com/alphagov/govuk-frontend/blob/move-percy-screenshots/lib/puppeteer-helpers.js#L5)

.percy.js
```diff
@@ -0,0 +1,7 @@
+/** @type {import('@percy/config').PercyConfigObject} */
+module.exports = {
+  percy: {
+    deferUploads: true
```
colinrotherham, line 4: Reduces total screenshot collection time by uploading at the end

.github/workflows/tests.yml
```diff
@@ -154,11 +156,32 @@ jobs:
...
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Restore dependencies
+        uses: ./.github/workflows/actions/install-node
+
+      - name: Start review application
+        run: npm run serve &
```
colinrotherham, line 184: Starts the review app in the background using `&`

.github/workflows/tests.yml
```diff
@@ -154,11 +156,32 @@ jobs:
...
-          PERCY_TOKEN: ${{ secrets.PERCY_TOKEN }}
+
+  regression:
+    name: Regression checks
+    runs-on: ubuntu-latest
+    needs: [install, build]
+
+    env:
+      PERCY_TOKEN: ${{ secrets.PERCY_TOKEN }}
+      PORT: 8888
```
colinrotherham, line 174: Review app uses port `3000` but our [Puppeteer helpers want `8888`](https://github.com/alphagov/govuk-frontend/blob/move-percy-screenshots/lib/puppeteer-helpers.js#L5)

tasks/screenshot-components.mjs
```diff
@@ -0,0 +1,81 @@
...
+ * @returns {Promise<void>}
+ */
+export async function screenshotComponents () {
+  const browser = await launch(configPuppeteer.launch)
+  const componentNames = await getDirectories(configPaths.components)
+
+  // Screenshot each component
+  const input = [...componentNames]
+
+  // Screenshot 4x components concurrently
```
36degrees, line 25: How did you arrive at 4x? Might be useful to capture in the commit message in case we want to revisit in the future.

colinrotherham: Tried 5x initially but 4x was faster

Good point though, worth trying 2x and 3x as well (you never know!)

lib/puppeteer-helpers.js
```diff
@@ -2,7 +2,7 @@ const { componentNameToJavaScriptClassName } = require('./helper-functions.js')
 const { renderHtml } = require('./jest-helpers.js')
 
 const configPaths = require('../config/paths.js')
-const PORT = configPaths.ports.test
+const { PORT = configPaths.ports.test } = process.env
```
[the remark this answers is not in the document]
colinrotherham: Ah sorry you've said this before, I'll change it 👍 

We've got a couple of others:

https://github.com/alphagov/govuk-frontend/blob/main/config/jest/server/start.mjs#L6
https://github.com/alphagov/govuk-frontend/blob/main/lib/puppeteer-helpers.js#L5

It looks clearer with multiple envs perhaps?

```mjs
const {
  ENV_1 = 'Fallback value 1',
  ENV_2 = 'Fallback value 2',
  ENV_3 = 'Fallback value 3',
  ENV_4 = 'Fallback value 4',
  ENV_5 = 'Fallback value 5'
} = process.env
```

tasks/screenshot-components.mjs
```diff
@@ -0,0 +1,81 @@
...
+}
+
+/**
+ * Send screenshot to Percy
+ *
+ * @param {import('puppeteer').Page} page - Puppeteer page object
+ * @param {string} screenshotName - Screenshot name
+ * @returns {Promise<void>}
+ */
+export async function screenshot (page, screenshotName) {
```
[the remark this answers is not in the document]
colinrotherham: For JSDoc perhaps at the time, let's remove it

.github/workflows/tests.yml
```diff
@@ -137,10 +137,12 @@ jobs:
           - description: JavaScript behaviour tests
             name: test-behaviour
             run: npx jest --color --maxWorkers=2 --selectProjects "JavaScript behaviour tests"
+            cache: .cache/jest
```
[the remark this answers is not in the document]
colinrotherham: This was me noticing that we restore the `.cache/jest` directory for all four **Verify** jobs

But didn't necessarily need it for:

```console
npm run build:package
npm run build:dist
```

As we saw in the [Windows runners](https://github.com/alphagov/govuk-frontend/pull/2945) often "restoring the cache" was slower than not having it.

The code was simpler before so I'll put it back

.github/workflows/tests.yml
```diff
@@ -152,13 +154,37 @@ jobs:
...
-          path: .cache/jest
+          path: ${{ matrix.cache }}
 
       - name: Run verify task
         run: ${{ matrix.run }}
-        env:
-          PERCY_TOKEN: ${{ secrets.PERCY_TOKEN }}
+
+  regression:
+    name: Regression checks
```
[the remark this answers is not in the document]
colinrotherham: Ah so this is the job name "**Regression checks**" which might include others in future?

The step name on line 189 says "**Send screenshots to Percy**"

Happy to make them match though

tasks/screenshot-components.mjs
colinrotherham: If you click into the **Send screenshots to Percy** job to see the timings:

* 1x component in sequence [govuk-frontend/actions/runs/5854565851](https://github.com/alphagov/govuk-frontend/actions/runs/5854565851) – ⏱️ **36 seconds**
* 2x components concurrently [govuk-frontend/actions/runs/3496480973](https://github.com/alphagov/govuk-frontend/actions/runs/3496480973) – ⏱️ **31 seconds**
* 3x components concurrently [govuk-frontend/actions/runs/3496481529](https://github.com/alphagov/govuk-frontend/actions/runs/3496481529) – ⏱️ **28 seconds**
* 4x components concurrently [govuk-frontend/actions/runs/3496493071](https://github.com/alphagov/govuk-frontend/actions/runs/3496493071) – ⏱️ **22 seconds**
* 5x components concurrently [govuk-frontend/actions/runs/5854480472](https://github.com/alphagov/govuk-frontend/actions/runs/5854480472) – ⏱️ **22 seconds**

The total action time varies due to [GitHub cache retrieval](https://github.com/alphagov/govuk-frontend/actions/caches) API times being wobbly

I'll keep it at 4x but add a comment

docs/releasing/testing-and-linting.md
```diff
@@ -101,7 +101,7 @@ We generate 2 screenshots for each default example of every component. One examp
 
 The screenshots are public, so you can check them without logging in. A BrowserStack account is needed to approve or reject any changes (if you don't have access, ask your tech lead for help). If you're the reviewer of the pull request code, it's your responsibility to approve or request changes for any visual changes Percy highlights.
 
-When you run the tests locally (for example, using `npm test`), Percy commands are ignored and Percy does not generate any screenshots. You will see the following message in your command line output: `[percy] Percy is not running, disabling snapshots`.
+When you run the tests locally (for example, using `npm run test:screenshots`), Percy commands are ignored and Percy does not generate any screenshots. You will see the following message in your command line output: `[percy] Percy is not running, disabling snapshots`.
```
36degrees, line 104: Thanks for keeping the docs in sync 🙌🏻

src/govuk/components/all.test.js
```diff
@@ -1,51 +0,0 @@
-const { join } = require('path')
-const { fetch } = require('undici')
-const { WebSocket } = require('ws')
```
36degrees, line 3: Can we now remove `ws` from our dependencies? (Was introduced in 496cc386817e8360420bf9b590e08bbecabb4e60)

36degrees (approved):
I think we can tidy up a dependency we no longer need, but otherwise this looks good to me 👍🏻

src/govuk/components/all.test.js
colinrotherham: Very nice, yep let's do that 🎉

docs/releasing/testing-and-linting.md
colinrotherham: @36degrees No problem. I didn't quite know if we should keep the `test:*` prefix

Based on the "…‘tests’ aren’t really tests” feedback

But we can always revisit another day. Thanks again 😊

PR #3017, opened 2022-11-16
Use node-sass v8 when testing Sass compilation

36degrees:
See https://github.com/sass/node-sass/releases/tag/v8.0.0

### DIFF
--- a/.github/workflows/sass.yaml
+++ b/.github/workflows/sass.yaml
@@ -74,9 +74,9 @@ jobs:
       - uses: actions/setup-node@v3
         with:
           cache: npm
-          node-version: 14
+          node-version: 18
       - run: |
-          npm install -g node-sass@v7
+          npm install -g node-sass@v8
           node-sass --version
       - run: time node-sass src/govuk/all.scss > /dev/null
 


.github/workflows/sass.yaml
```diff
@@ -74,9 +74,9 @@ jobs:
       - uses: actions/setup-node@v3
         with:
           cache: npm
-          node-version: 14
+          node-version: 18
```
colinrotherham, line 77: I'd probably keep this PR solely as `node-sass@v8`?

Probably want to split out any "Node.js upgrade" related work as it touches lots of areas now. With [**Node.js v18** now in LTS](https://nodejs.org/en/blog/release/v18.12.0/) and [**npm v9** now `@latest`](https://github.blog/changelog/2022-11-09-npm-cli-v9-is-now-generally-available/) would supporting "latest two LTS releases" be sensible?

Other moving parts include:

1. **Testing against multiple Node.js versions**
See [`govuk-prototype-kit/blob/v13/.github/workflows/test.yaml`](https://github.com/alphagov/govuk-prototype-kit/blob/v13/.github/workflows/test.yaml#L16)

2. **Updating [.browserslistrc](https://github.com/alphagov/govuk-frontend/blob/main/.browserslistrc#L16)** for Node.js 18
So Babel is aware we need to support both `node 16` and `node 18`

3. **Updating [.nvmrc](https://github.com/alphagov/govuk-frontend/blob/main/.nvmrc) for Node.js 18**
So tools like [nvm](https://github.com/nvm-sh/nvm#readme) and [asdf](https://github.com/asdf-vm/asdf#readme) allow LTS releases

4. **Widen the [package.json `engines` range](https://github.com/alphagov/govuk-frontend/blob/main/package.json#L5)**
Ensures `npm install` and other scripts check the local Node.js version is allowed

Useful info:

* https://github.com/alphagov/govuk-prototype-kit/issues/1124
* https://github.com/alphagov/govuk-prototype-kit/pull/1700#discussion_r1004199604

colinrotherham: If this line got accidentally left behind (on Node.js v14) could match the repo for now?

```suggestion
          node-version-file: .nvmrc
```

colinrotherham (commented):
No blockers though, happy to approve 👍 

Have another look too, but I've rebased my [`sass-test-matrix` branch](https://github.com/alphagov/govuk-frontend/compare/sass-test-matrix)

I've added more commits for extra Node.js versions running here:
https://github.com/alphagov/govuk-frontend/actions/runs/3481861957

Similar to the Prototype Kit approach

colinrotherham (approved):

.github/workflows/sass.yaml
colinrotherham: @36degrees Tiny update on this

Remember all those **node-gyp** compilation steps on `npm install` when a binary wasn't available?

If you try and install `node-sass@7` on Node.js 18 it's super slow. This PR is fine, but something to bear in mind

36degrees: In my head these tests are relatively standalone from the rest of the repo, as they're meant to test that our Sass compiles using the oldest and latest versions of the Sass compilers we officially support. I don't personally see a reason to tie the Node version used here to the version we use for our tooling and tests.

Thoughts?

36degrees:
> Have another look too, but I've rebased my [`sass-test-matrix` branch](https://github.com/alphagov/govuk-frontend/compare/sass-test-matrix)
> 
> I've added more commits for extra Node.js versions running here: https://github.com/alphagov/govuk-frontend/actions/runs/3481861957

I don't think we need to care about multiple Node versions here – the Sass tests exist to give us confidence that our Sass compilers when used with the oldest and newest versions of [the compilers we say that we support](https://frontend.design-system.service.gov.uk/installing-with-npm/#install-with-node-js-package-manager-npm).

We should just test each compiler running in the latest version of Node / Ruby that is supported by the compiler.

.github/workflows/sass.yaml
colinrotherham: Yeah I think they're all correct answers

>I'd probably keep this PR solely as node-sass@v8?

What do we need the bump to Node.js 18 for?

36degrees: We don't need it; in my head it just makes sense to run the various compilers in the latest version of node that they support.

colinrotherham:
> I don't think we need to care about multiple Node versions here – the Sass tests exist to give us confidence that our Sass compilers when used with the oldest and newest versions of [the compilers we say that we support](https://frontend.design-system.service.gov.uk/installing-with-npm/#install-with-node-js-package-manager-npm).
> 
> We should just test each compiler running in the latest version of Node / Ruby that is supported by the compiler.

In my head (with Node.js 18 now in LTS) some contributors and our `govuk-frontend` users downstream might be stuck on older Node.js versions (14, 16) with various combinations too of `node-sass@7` and `node-sass@8`.

You hope it never happens, but a wider safety net sounded good.

Do we know how many haven't made the Dart Sass switch yet?

.github/workflows/sass.yaml
colinrotherham: Yeah same, Node.js 14 is the lowest `node-sass@8` supports 😆

36degrees:
> > I don't think we need to care about multiple Node versions here – the Sass tests exist to give us confidence that our Sass compilers when used with the oldest and newest versions of [the compilers we say that we support](https://frontend.design-system.service.gov.uk/installing-with-npm/#install-with-node-js-package-manager-npm).
> > We should just test each compiler running in the latest version of Node / Ruby that is supported by the compiler.
> 
> In my head (with Node.js 18 now in LTS) some contributors and our `govuk-frontend` users downstream might be stuck on older Node.js versions (14, 16) with various combinations too of `node-sass@7` and `node-sass@8`.
> 
> You hope it never happens, but a wider safety net sounded good.
> 
> Do we know how many haven't made the Dart Sass switch yet?

Feel like this is tangential to the changes in this PR, so let's pick this up as/when your changes turn into a PR or issue.

Given the rest of the repo is now running Node 18 I'm going to merge this as is.

PR #3021, opened 2022-11-17
Printing style for current page link in header

MalcolmVonMoJ:
Changing the blue back to the brand colour `#1d70b8` when printing.  The colour `#1d8feb` is presumably to contrast against the black `#0b0c0c` header, but when printing the header is white `#ffffff`, so it makes sense to use a darker blue which contrasts better against white.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Fixes
 
 - [#2998: Refactor back link and breadcrumb chevrons to use ems](https://github.com/alphagov/govuk-frontend/pull/2998)
+- [#3021 Printing style for current page link in header](https://github.com/alphagov/govuk-frontend/pull/3021) - thanks to [Malcolm Butler](https://github.com/MalcolmVonMoJ) for the contribution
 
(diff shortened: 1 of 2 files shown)


querkmachine (approved):
Does exactly what it says on the tin.

Needs a second approval before merging. Probably warrants a changelog entry too?

owenatgov (commented):
@MalcolmVonMoJ Looks good to me too 👍🏻 

Before I give this the official github tick, I'd like to ask that you add a changelog entry and squash the commit history. We're happy to handle this if you prefer.

MalcolmVonMoJ:
@owenatgov Can you do this please?  I don't know how to squash in github.  I went to add it to the changelog too, but mistook "feature release" for "future release" and put it in the wrong place.  Undid that error, but I think someone else is best to squash and update the changelog as I'm not sure where it is.  

Also, one of the automated tests was failing and I am not sure what to do about it.

owenatgov (approved):
Approving as the code is 👌🏻 I'll tidy up the commit history, add a changelog entry and merge in a jiffy.

PR #3023, opened 2022-11-18
Add source maps to compiled JavaScript + CSS

colinrotherham:
This PR adds source maps to our compiled code

For example:

**govuk-frontend-4.4.0.min.js**
```js
//# sourceMappingURL=govuk-frontend-4.4.0.min.js.map
```

**govuk-frontend-4.4.0.min.css**
```css
/*# sourceMappingURL=govuk-frontend-4.4.0.min.css.map */
```

When browser developer tools are open the associated source map files are downloaded

## JavaScript source maps
See log messages and stack traces mapped to the `*.mjs` source code (with line numbers)

JavaScript source maps

Click through to the original source via the browser

JavaScript source maps (code)

## CSS source maps
See styles mapped to the Sass `*.scss` source code (with line numbers)

CSS source maps

### DIFF
--- a/tasks/asset-version.mjs
+++ b/tasks/asset-version.mjs
@@ -1,41 +1,26 @@
-import { readFile, rename, writeFile } from 'fs/promises'
+import { readFile, writeFile } from 'fs/promises'
 import { EOL } from 'os'
-import { basename, join } from 'path'
+import { join } from 'path'
 
 import configPaths from '../config/paths.js'
 
 import { destination, isDist } from './task-arguments.mjs'
 
-// Update assets' version numbers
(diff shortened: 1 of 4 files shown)


romaricpascal:
Sourcemaps are great and the PR looks technically neat for adding them in. Just looking into what will consume them so we're clear on who will benefit and who will not (depending on how code is loaded) and maybe check if some guidance will need updating. More news on this space in a bit 😄

colinrotherham:
@romaricpascal Rebased and ready for review again 👍

romaricpascal (approved):
Looking neat! 🙌🏻

PR #3024, opened 2022-11-21
Fix Safari `TypeError: 'handleEvent'` iFrame resizer error

colinrotherham:
This PR removes code that logs a error in Safari

```console
TypeError: 'handleEvent' property of event listener should be callable
```

## The issue

When an event listener is not callable Safari logs an error where other browsers stay silent.

We've got an example of this in our review app below, where the `iFrameResize()` function call returns an array of `<iframe>` elements. Safari logs an error because we use this array as our `DOMContentLoaded` handler.

This also shows we don't need `domReady()` as [iFrame Resizer](http://davidjbradshaw.github.io/iframe-resizer/) runs anyway

```js
<script type="text/javascript">
  var domReady = function(a,b,c){b=document,c='addEventListener';b[c]?b[c]('DOMContentLoaded',a):window.attachEvent('onload',a)}
  domReady(iFrameResize({}, '.js-component-preview'))
</script>
```

TypeError 'handleEvent' in Safari

I've also included some other fixes:

1. **Remove `iframeResizer.min.js` **<script>** from `component-preview.njk`**
Preview iframes only need `iframeResizer.contentWindow.min.js`

2. ~**Upgrade to [iFrame Resizer v3.6.5](https://github.com/davidjbradshaw/iframe-resizer/releases/tag/v3.6.5)**~
~This is the last IE8 compatible release~

### DIFF
--- a/app/views/component-preview.njk
+++ b/app/views/component-preview.njk
@@ -16,5 +16,5 @@
 {% endblock %}
 
 {% block scripts %}
-  <script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/iframe-resizer/3.5.15/iframeResizer.contentWindow.min.js" integrity="sha256-4pHiLAYReL+uT1xGu9u8Afg9jkaV0vrdu/Dd0ax9Ak8=" crossorigin="anonymous"></script>
+  <script src="https://cdnjs.cloudflare.com/ajax/libs/iframe-resizer/3.5.15/iframeResizer.contentWindow.min.js" integrity="sha256-4pHiLAYReL+uT1xGu9u8Afg9jkaV0vrdu/Dd0ax9Ak8=" crossorigin="anonymous"></script>
 {% endblock %}
--- a/app/views/layouts/layout.njk
+++ b/app/views/layouts/layout.njk
@@ -10,12 +10,13 @@
 
 {% block bodyEnd %}
   {{ super() }}
-  <script src="https://cdnjs.cloudflare.com/ajax/libs/iframe-resizer/3.5.15/iframeResizer.min.js" integrity="sha256-NZjCYaMfryuJQRMgekHuC02c/Wv4sMRzHG2zyhrVwKU=" crossorigin="anonymous"></script>
-  <!--[if lte IE 8]>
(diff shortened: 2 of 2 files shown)


36degrees (commented):
Good catch, thanks for fixing this 🙌🏻 

It took me a little while to work out what was happening in the last commit, as it's not entirely clear that the `scripts` block is being defined in `layout.njk` and _overridden_ in `component-preview.njk`. I think it'd be worth explaining the changes in the commit message.

colinrotherham:
> Good catch, thanks for fixing this 🙌🏻
> 
> It took me a little while to work out what was happening in the last commit, as it's not entirely clear that the `scripts` block is being defined in `layout.njk` and _overridden_ in `component-preview.njk`. I think it'd be worth explaining the changes in the commit message.

Thanks @36degrees, yeah I was hoping we get a browser main thread performance boost by not loading `iframeResizer.min.js` in each iframe too.

I did try conditional IE comments to use [**iFrame Resizer v4.3.2**](https://github.com/davidjbradshaw/iframe-resizer/releases/tag/v4.3.2) but we'd still need IE10 on v3 😔 

I've pushed again though, how does that look?

36degrees (approved):

colinrotherham:
Thanks for approving this @36degrees

Sadly I've had to drop the "**Upgrade to iFrame Resizer v3.5.16**" commit

Turns out newer builds of [**iframe-resizer** v3](https://github.com/davidjbradshaw/iframe-resizer) v3 (IE8 compatible) aren't actually compatible. Partly due to a project-wide [Prettier](https://prettier.io) run that removed quotes from object keys `{ 'true': 1 }` → `{ true: 1 }` in [`9908ff9`](https://github.com/davidjbradshaw/iframe-resizer/commit/9908ff956cfc8d30141d5b86b98aa7c5c636c49d#diff-763d8de084e4b1e37760383c4d712caf3eb43eeaad2c9090bc192a03a8f68d55R251)

I'll merge without it so we still get the bug fix

36degrees:
Looks like that Prettier commit was introduced in 3.6.4, so is it worth us bumping up to 3.6.3?

colinrotherham:
> Looks like that Prettier commit was introduced in 3.6.4, so is it worth us bumping up to 3.6.3?

@36degrees It got worse. I tried that and we still had a script error in `iframeResizer.contentWindow.min.js` 🙈

PR #3028, opened 2022-11-21
Skip “Percy screenshots” when secrets are unavailable on forks

colinrotherham:
The Percy CLI requires `${{ secrets.PERCY_TOKEN }}` which forks won't necessarily have. We recently made our Percy npm script return an error code when the health check failed in:

* https://github.com/alphagov/govuk-frontend/pull/3009

This PR adds a condition to our **Send screenshots to Percy** job which makes it skippable

```yaml
# Skip when secrets are unavailable on forks
if: ${{ github.repository_owner == 'alphagov' }}
```

Where permissions allow you'll see we've now added a Dependabot `PERCY_TOKEN` secret too:
https://github.com/alphagov/govuk-frontend/settings/secrets/dependabot

### Reducing Percy runs in future
We may want to consider a ["paths" filter](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore) in future so this PR also moves the Percy uploads to a separate GitHub Actions [`.github/workflows/screenshots.yml`](https://github.com/alphagov/govuk-frontend/blob/percy-secret-permissions/.github/workflows/screenshots.yml) workflow. More information in:

* https://github.com/alphagov/govuk-frontend/issues/2864

### DIFF
--- a/.github/workflows/screenshots.yml
+++ b/.github/workflows/screenshots.yml
@@ -1,0 +1,34 @@
+name: Percy screenshots
+
+on:
+  workflow_call:
+  workflow_dispatch:
+
+concurrency:
+   group: screenshots-${{ github.head_ref || github.run_id }}
+   cancel-in-progress: true
+
+jobs:
+  screenshots:
+    name: Send screenshots
+    runs-on: ubuntu-latest
+
+    env:
+      PERCY_TOKEN: ${{ secrets.PERCY_TOKEN }}
+      PORT: 8888
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Install dependencies
+        uses: ./.github/workflows/actions/install-node
+
+      - name: Build
+        uses: ./.github/workflows/actions/build
+
+      - name: Start review application
+        run: npm run serve &
+
+      - name: Send screenshots to Percy
+        run: npx percy exec -- npm run test:screenshots
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -169,3 +169,6 @@ jobs:
     # (after install and build have been cached)
     uses: ./.github/workflows/screenshots.yml
     secrets: inherit
+
+    # Skip when secrets are unavailable on forks
+    if: ${{ github.repository_owner == 'alphagov' }}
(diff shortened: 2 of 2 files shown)


.github/workflows/screenshots.yml
```diff
@@ -0,0 +1,34 @@
+name: Percy screenshots
+
+on:
+  workflow_call:
+  workflow_dispatch:
+
+concurrency:
+   group: screenshots-${{ github.head_ref || github.run_id }}
+   cancel-in-progress: true
+
```
36degrees, line 10: This change has snuck in to the 'move to separate workflow commit' – can we split it out into a separate commit?

.github/workflows/tests.yml
```diff
@@ -162,26 +162,10 @@ jobs:
         run: ${{ matrix.run }}
 
   regression:
-    name: Send screenshots to Percy
-    runs-on: ubuntu-latest
+    name: Percy
```
36degrees, line 165: Is there any benefit to having this inside the tests workflow, rather than just having `screenshots.yml` trigger on the same `pull_request` and `push` events?

.github/workflows/screenshots.yml
colinrotherham: Ah which bit has snuck in, renaming the group?

.github/workflows/tests.yml
colinrotherham: Yeah, it gives the **Install** and **Build** steps chance to get cached

Then when we trigger the Percy workflow it can go super fast (already installed and built)

.github/workflows/screenshots.yml
colinrotherham: We added it in: https://github.com/alphagov/govuk-frontend/pull/2878/files#diff-1db27d93186e46d3b441ece35801b244db8ee144ff1405ca27a163bfe878957fR14

So just moved it over and made it unique

36degrees: Aah of course, the tests workflow already has concurrency set and this needs it applied as it's being split out.

.github/workflows/tests.yml
36degrees: Makes sense – but I think this might be worth adding as a comment somewhere as it's not obvious.

36degrees (approved):

PR #3033, opened 2022-11-23
Ensure npm scripts (via Gulp) emit errors

colinrotherham:
This PR ensure npm scripts (via Gulp) emit errors

We currently ignore npm script crashes even when they write to `stderr` (which wasn't intentional)

I've also prevented the [**npm run**](https://docs.npmjs.com/cli/v9/commands/npm-run-script) `--silent` flag being logged to reduce Gulp CLI "log spam"

### Before
```console
[15:04:48] Starting 'npm run build:jsdoc --silent'...
[15:04:49] Finished 'npm run build:jsdoc --silent' after 1.22 s
[15:04:49] Starting 'npm run build:sassdoc --silent'...
[15:04:51] Finished 'npm run build:sassdoc --silent' after 1.51 s
```

### After
```console
[15:04:48] Starting 'npm run build:jsdoc'...
[15:04:49] Finished 'npm run build:jsdoc' after 1.22 s
[15:04:49] Starting 'npm run build:sassdoc'...
[15:04:51] Finished 'npm run build:sassdoc' after 1.51 s
```

### DIFF
--- a/tasks/run.mjs
+++ b/tasks/run.mjs
@@ -14,14 +14,33 @@ export async function npmScript (name, args = []) {
   return new Promise((resolve, reject) => {
     const script = spawn(command, ['run', name, '--silent', ...args])
 
+    // Send output to console
     script.stdout.on('data', (data) => console.log(data.toString()))
-    script.stderr.on('data', (data) => console.error(data.toString()))
+
+    // Emit errors to error listener
+    script.stderr.on('data', (data) => {
+      script.emit('error', new Error(data.toString()))
+    })
 
     // Reject on actual script errors to exit `gulp dev`
     script.on('error', reject)
 
-    // Resolve all exit codes to continue `gulp dev`
-    script.on('close', resolve)
+    // Check for exit codes or continue `gulp dev`
+    script.on('close', (code) => {
+      let error
+
+      // Closed with errors
+      if (code > 0) {
+        error = new Error(`Task for npm script '${name}' exit code ${code}`)
+
+        // Hide error info (already written to `stderr`)
+        error.showProperties = false
+        error.showStack = false
+      }
+
+      // Reject on errors
+      error ? reject(error) : resolve(code)
+    })
   })
 }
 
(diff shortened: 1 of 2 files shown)


tasks/run.mjs
```diff
@@ -12,10 +12,15 @@ export async function npmScript (name, args = []) {
   const command = process.platform === 'win32' ? 'npm.cmd' : 'npm'
 
   return new Promise((resolve, reject) => {
-    const script = spawn(command, ['run', name, ...args])
+    const script = spawn(command, ['run', name, '--silent', ...args])
```
colinrotherham, line 15: We still add `--silent` behind the scenes because we don't want both **npm** AND **Gulp** logging progress to the console—one is enough

**Note:** _Output_ from the npm task (linting for example) is not silenced

tasks/run.mjs
```diff
@@ -12,10 +12,15 @@ export async function npmScript (name, args = []) {
...
 
+    // Send output to console
     script.stdout.on('data', (data) => console.log(data.toString()))
-    script.stderr.on('data', (data) => console.error(data.toString()))
+
+    // Emit errors to error listener
+    script.stderr.on('data', (data) => {
+      script.emit('error', new Error(data.toString()))
```
colinrotherham, line 22: By emitting an error here it ensures we hit `reject()` (below) via the error listener

romaricpascal: **question**: Won't that trigger multiple times if a script outputs to `stderr` multiple times? Furthermore, I'm worried:
- it'll catch false errors for a script that just outputs to stderr as a convenient second buffer. I thought the actual marker of a script erroring was a non-zero exit (covered by the `close`)
- it'll double up with the `close` handling later

**question**: Why are we emitting an error rather than directly rejecting? That feels an unnecessary extra step there.

colinrotherham: >Won't that trigger multiple times if a script outputs to stderr multiple times?

The first error written will hit `script.on('error', reject)` so you'll only see the first one

romaricpascal (approved):
It all makes sense now, cheers for the explanation 😄

Issue #2872, opened 2022-09-21
Add `gulp-plumber` to JavaScript compile task to log UglifyJS issues


colinrotherham:
Unlike how we compile Sass, we don't add [`gulp-plumber`](https://github.com/floatdrop/gulp-plumber#readme) to the JavaScript compile

**compileStyles ()** via `gulp scss:compile`
Notice how errors are caught by plumber and logged by `errorHandler`?
```js
return gulp.src(compileStylesheet)
  .pipe(plumber(errorHandler))
  .pipe(sass())
```

UglifyJS errors can provide error message, cause, line numbers etc

This will have caught https://github.com/alphagov/govuk-frontend/pull/2868 without manual logging

PR #3035, opened 2022-11-23
Allow JavaScript compilation to restart after `gulp dev` errors

colinrotherham:
This PR closes https://github.com/alphagov/govuk-frontend/issues/2872

The changes enable friendly errors for our JavaScript compile (+ uglify) task

### Before
Gulp logs a super long and confusing `Unhandled 'error' event`

```console
[15:39:33] The following tasks did not complete: compile, compile:scss
[15:39:33] Did you forget to signal async completion?

node:events:491
      throw er; // Unhandled 'error' event
      ^
PluginError: Unexpected token

[… more]
```

### After
We use [`gulp-plumber`](https://github.com/floatdrop/gulp-plumber) to log the friendly error message (showing the faulty code):

```console
[15:38:10] Starting 'compile:scss'...
Error in plugin "gulp-better-rollup"
Message:
    Unexpected token
Details:
    code: PARSE_ERROR
    pos: 1256
    loc: [object Object]
    frame: 25:   // Allow the user to initialise GOV.UK Frontend in only certain sections of the page
26:   // Defaults to the entire document if nothing is set.
27:   var $scope = typeof config.scope !== 'undefined' ? config.scope document
                                                                      ^
28:
29:   var $accordions = $scope.querySelectorAll('[data-module="govuk-accordion"]')
```

Follows the same approach as:

* https://github.com/alphagov/govuk-frontend/pull/3034

### DIFF
--- a/tasks/gulp/compile-assets.mjs
+++ b/tasks/gulp/compile-assets.mjs
@@ -129,30 +129,45 @@ export async function compileJavaScripts () {
       ? componentNameToJavaScriptModuleName(name)
       : 'GOVUKFrontend'
 
-    return gulp.src(slash(join(configPaths.src, file)))
-      .pipe(rollup({
-        // Used to set the `window` global and UMD/AMD export name
-        // Component JavaScript is given a unique name to aid individual imports, e.g GOVUKFrontend.Accordion
-        name: moduleName,
-        // Legacy mode is required for IE8 support
-        legacy: true,
-        // UMD allows the published bundle to work in CommonJS and in the browser.
-        format: 'umd'
-      }))
-      .pipe(gulpif(isDist, uglify({
-        ie8: true
-      })))
-      .pipe(gulpif(isDist,
-        rename({
-          basename: 'govuk-frontend',
-          extname: '.min.js'
-        })
-      ))
-      .pipe(rename({
-        extname: '.js'
-      }))
+    return compileJavaScript(gulp.src(slash(join(configPaths.src, file))), moduleName)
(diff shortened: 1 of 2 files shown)


colinrotherham:
Linking to https://github.com/alphagov/govuk-frontend/issues/2243

tasks/gulp/compile-assets.mjs
```diff
@@ -129,30 +129,61 @@ export async function compileJavaScripts () {
...
+      const error = new PluginError('compile:js', cause)
+      console.error(error.toString())
+
+      // Gulp continue (watch)
+      if (isDev) {
+        return this.emit('end')
+      }
+
+      // Gulp exit with error
+      return stream.emit('error', error)
```
romaricpascal, line 158: **question** This looks slightly different from the one used in #3034:
- we're building the `PluginError` with `cause` instead of `cause.messageFormatted`
- we're not doing an `error.showProperties = false`
- we're `console.error`ing the error in all cases, rather than just in dev

Are these differences intentional or just due to things having been written at different times?

romaricpascal (approved):
Cheers for clarifying once again 🙌🏻 Looks good to merge 🚀

colinrotherham:
Thanks @romaricpascal

I've taken more of your feedback and combined the two error handlers in https://github.com/alphagov/govuk-frontend/pull/3035/commits/374a8d7f5378d2b89ab209660039a3bda2d8bf7e

Found a neat way to work around the differences between Sass and Rollup stream errors 😊

PR #3018, opened 2022-11-17
Allow Node.js LTS ‘Hydrogen’ (v18) release version

colinrotherham:
This PR allows Node.js LTS ‘Hydrogen’ (v18) to install, run and build **GOV.UK Frontend**

Changes included:

1. **Updating [.browserslistrc](https://github.com/alphagov/govuk-frontend/blob/main/.browserslistrc#L16)** for Node.js 18
So Babel is aware we need to support only `node 18`

2. **Updating [.nvmrc](https://github.com/alphagov/govuk-frontend/blob/main/.nvmrc)** for Node.js 18
So tools like [nvm](https://github.com/nvm-sh/nvm#readme) and [asdf](https://github.com/asdf-vm/asdf#readme) allow the `lts/hydrogen` LTS release

3. **Bump the [package.json `engines.node` range](https://github.com/alphagov/govuk-frontend/blob/main/package.json#L6)**
Ensures `npm install` and other scripts check the local Node.js version is allowed

3. **Widen the [package.json `engines.npm` range](https://github.com/alphagov/govuk-frontend/blob/main/package.json#L7)**
So we're ready for `npm@9` as it's being [backported to Node.js 18 in December](https://github.com/npm/statusboard/issues/587)

PR #2858, opened 2022-09-16
Replace deprecated `request` package with Fetch API

colinrotherham:
I've split this change from https://github.com/alphagov/govuk-frontend/pull/2850 so it's easier to review

The changes [resolve 3x vulnerabilities in `node-sass`](https://github.com/sass/node-sass/releases/tag/v7.0.3) and removes this message on install:

>npm WARN deprecated request@2.88.2: request has been deprecated, see https://github.com/request/request/issues/3142

We'll see to check for CSS changes from [`node-sass@7.01`](https://github.com/sass/node-sass/releases/tag/v7.0.1) → [`node-sass@7.03`](https://github.com/sass/node-sass/releases/tag/v7.0.3)

I've replaced  `request()` with [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) following this approach (until native support arrives in Node.js 17):
https://blog.logrocket.com/fetch-api-node-js/#undici

PR #3045, opened 2022-11-28
Use native Fetch API in behaviour tests

colinrotherham:
We temporarily used [`undici`](https://github.com/nodejs/undici) to prepare for the [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) landing in Node.js 18:

* https://github.com/alphagov/govuk-frontend/pull/2858
* https://github.com/alphagov/govuk-frontend/pull/3018

Both PRs have now merged so we can remove [`undici`](https://github.com/nodejs/undici)

(Another package Dependabot doesn't need to worry about)

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -31,7 +31,6 @@
     "shuffle-seed": "^1.1.6"
   },
   "devDependencies": {
-    "cheerio": "^1.0.0-rc.12",
-    "undici": "^5.12.0"
+    "cheerio": "^1.0.0-rc.12"
   }
 }
--- a/app/package.json
+++ b/app/package.json
@@ -12,15 +12,15 @@
     "start": "node start.js"
   },
   "dependencies": {
-    "express": "^4.18.2",
-    "express-validator": "^6.14.2",
     "body-parser": "^1.18.3",
     "cookie-parser": "^1.4.6",
+    "express": "^4.18.2",
+    "express-validator": "^6.14.2",
     "front-matter": "^4.0.2",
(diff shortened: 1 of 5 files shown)


app/package.json
```diff
@@ -12,15 +12,15 @@
...
     "body-parser": "^1.18.3",
     "cookie-parser": "^1.4.6",
+    "express": "^4.18.2",
+    "express-validator": "^6.14.2",
     "front-matter": "^4.0.2",
     "glob": "^8.0.3",
-    "govuk-elements-sass": "3.1.3",
     "govuk_frontend_toolkit": "^9.0.1",
     "govuk_template_jinja": "^0.26.0",
+    "govuk-elements-sass": "3.1.3",
```
domoscargin, line 23: Nice little tidy up.

I would've expected `govuk-elements-sass` to stay where it was since hyphens come before underscores in ASCII/UTF (think it also just makes sense from a human-readable point of view - "e" comes before "f", and all that).

domoscargin (approved):
Looks good to me! Love to see native stuff being used in favour of packages.

My only comment is definitely non-blocking.

app/package.json
colinrotherham: @domoscargin I'll put my hand up and say I manually ordered it wrong last time (sorry), but this time it was definitely **npm** fixing my mistake so I've trusted it here 😬

There's an ESLint plugin for keeping `package.json` files in order, maybe we need it 🙈 

Thanks @domoscargin

Issue #2697, opened 2022-07-14
[Spike] Investigate updating the accordion to use the `hidden="until-found"` attribute


36degrees:
## What

Investigate whether we can update the accordion to use the `hidden="until-found"` attribute, as originally raised by @injms in https://github.com/alphagov/govuk-design-system-backlog/issues/1#issuecomment-1127769827:

> The `hidden="until-found"` attribute looks to be an interesting new feature that the accordion could benefit from in a backwards compatible manner:
> 
> > By adding hidden=until-found to the container for your hidden content, you make it possible for the browser to search text in that hidden region, and reveal the section if a match is found.
> 
> 
> https://developer.chrome.com/blog/hidden-until-found/

## Why

Using `hidden="until-found"` makes it possible to search the hidden content within the accordion using the browser's in-page search feature (in browsers that support it).

At the time of writing, [only very recent versions of Chromium-based browsers support the attribute](https://caniuse.com/mdn-html_global_attributes_hidden_until-found_value).

According to [Chrome Platform Status](https://chromestatus.com/feature/5400510406328320), Firefox [consider it worth prototyping](https://github.com/mozilla/standards-positions/issues/612) but there are no signals from Safari.

Chrome and Edge account for approximately 47% of visits to GOV.UK.

## Assumptions

We should firm this section up before starting the spike.

- Despite the lack of clear signals from other browsers, this attribute will continue to be supported and is worth adopting.

## Timebox

TBC

## Who is working on this?

Spike lead: TBC

Spike buddy: TBC

## Questions to answer

We should firm this up before starting the spike, but things we might want to look at include:

- behaviour in browsers that don't support the `hidden="until-found"` attribute
- how using ‘content-visibility: hidden’ instead of ‘display: none’ (because of the use of the attribute) affects how the content appears in the accessibility tree (or hopefully doesn't appear when the accordion is closed)
- whether this change can be made without breaking changes to the component

## Done when

- [ ] Questions have been answered or we have a clearer idea of how to get to our goal
- [ ] Findings have been reviewed and agreed with at least one other person
- [ ] Findings have been shared, e.g: via a write-up on the ticket, at a show & tell or team meeting

querkmachine:
I've made a very mild spike into this, essentially changing the Accordion component JS to add and remove `hidden="until-found"` on each sections content container (`.govuk-accordion__section-content`) instead of using class toggles and CSS to do so.

My answers to the above appear to be:

[...]

PR #3053, opened 2022-11-29
Enhance the Accordion component with `hidden='until-found'`

querkmachine:
`until-found` is a new value for the `hidden` attribute, that allows a user agent's find-in-page functionality to search the text inside of elements that are currently collapsed. It serves as a progressive enhancement to the existing `hidden` boolean attribute, falling back to the default behaviour in browsers that don't support the new value.

It was introduced at the same time as the `beforematch` JavaScript event, allowing us to see which element on the page the "found" text is contained within.

Currently the new attribute value and JavaScript event are only supported in Chromium-based browsers (including Google Chrome, Microsoft Edge, Opera, and Samsung Internet).

Linked to #2697. I previously did a small, hackier investigation into this at the start of August. See [my previous comment](https://github.com/alphagov/govuk-frontend/issues/2697#issuecomment-1203769281) for the outcome of that.

## Changes in this spike

- Replaces the hiding and showing of accordion sections via CSS, to instead conditionally add and remove the `hidden="until-found"` attribute using JavaScript.
  - This change means that the point that accordion content is hidden is slightly different. Formerly it was hidden (effectively) immediately, based on the presence of the `js-enabled` class. It now requires both the class and for the accordion's JavaScript to have been initialised before anything is hidden.
  - As a benefit, this is a little more resilient as it means the accordion's content is visible even when the accordion hasn't been initialised or errored whilst doing so, which wasn't the case previously. 
- Refactors the section content's padding to only apply when the section is expanded.
  - This is because an element with `hidden="until-found"` on it is not removed from the box model (it uses `content-visibility: hidden`, not `display: none` unlike boolean `hidden`) so the padding was always visible.
- Adds a `beforematch` event handler and function to correctly change the state, styles and button labels in the event of a match by the user agent.
  - Without this, the user agent would automatically remove the `hidden` attribute upon a match, but would not change any other parts of the accordion.

## Other thoughts

This would seem to be a non-breaking change in terms of service teams, unless a team has customised the styles of the section content box. As this element is no longer completely ignored when rendering the box model, any styles (such as borders, background colours, margins or padding) may become visible even when the section is collapsed.

~~This probably poses a larger issue for legacy Internet Explorer support. This is currently relying on user agent stylesheets to be clever and apply either `content-visibility: hidden` or `display: none` depending on the value (or lack) of the `hidden` attribute. IE 10 and below don't support the `hidden` attribute at all, so a `display: none` rule would need to be applied for those browsers without compromising those that support `hidden="until-found"`.~~ Resolved, see discussion below.

This functionality would seemingly want to be mirrored to the [Details component](https://design-system.service.gov.uk/components/details/), however we are relying solely on the user agent's own functionality to provide the expand/collapse feature in modern browsers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 ## Unreleased
 
+### New features
+
+#### Search within accordion content on supporting browsers
+
+We've updated the Accordion component to use the new [`hidden="until-found"` attribute value](https://developer.chrome.com/articles/hidden-until-found/).
+
+This allows the browser's native 'find in page' functionality to search within and automatically open sections of the accordion. Currently, this functionality is only supported by recent versions of Google Chrome, Microsoft Edge and Samsung Internet.
+
+This was added in [pull request #3053: Enhance the Accordion component with `hidden='until-found'`](https://github.com/alphagov/govuk-frontend/pull/3053). 
+
 ### Fixes
 
 - [#2998: Refactor back link and breadcrumb chevrons to use ems](https://github.com/alphagov/govuk-frontend/pull/2998)
--- a/src/govuk/components/accordion/_index.scss
+++ b/src/govuk/components/accordion/_index.scss
@@ -46,7 +46,18 @@
       padding-top: 0;
     }
 
-    // Hide the body of collapsed sections
+    // Manually apply display: none to browsers that don't have a user agent
+    // style for it, but override it with content-visibility if the browser
+    // supports that instead.
+    .govuk-accordion__section-content[hidden] {
+      display: none;
+      @supports (content-visibility: hidden) {
+        content-visibility: hidden;
+        display: inherit;
+      }
+    }
+
+    // Hide the padding of collapsed sections
     .govuk-accordion__section--expanded .govuk-accordion__section-content {
       @include govuk-responsive-padding(8, "bottom");
       @include govuk-responsive-padding(3, "top");
--- a/src/govuk/components/accordion/_index.scss
+++ b/src/govuk/components/accordion/_index.scss
@@ -47,15 +47,9 @@
     }
 
     // Hide the body of collapsed sections
-    .govuk-accordion__section-content {
-      display: none;
+    .govuk-accordion__section--expanded .govuk-accordion__section-content {
       @include govuk-responsive-padding(8, "bottom");
       @include govuk-responsive-padding(3, "top");
-    }
-
-    // Show the body of expanded sections
-    .govuk-accordion__section--expanded .govuk-accordion__section-content {
-      display: block;
     }
 
     .govuk-accordion__show-all {
--- a/src/govuk/components/accordion/accordion.mjs
+++ b/src/govuk/components/accordion/accordion.mjs
@@ -3,6 +3,7 @@ import { normaliseDataset } from '../../common/normalise-dataset.mjs'
 import { I18n } from '../../i18n.mjs'
 import '../../vendor/polyfills/Function/prototype/bind.mjs'
 import '../../vendor/polyfills/Element/prototype/classList.mjs'
+import '../../vendor/polyfills/Element/prototype/closest.mjs'
 import '../../vendor/polyfills/String/prototype/trim.mjs'
 
 /**
@@ -55,6 +56,7 @@ function Accordion ($module, config) {
   this.showAllClass = 'govuk-accordion__show-all'
   this.showAllTextClass = 'govuk-accordion__show-all-text'
 
+  this.sectionClass = 'govuk-accordion__section'
   this.sectionExpandedClass = 'govuk-accordion__section--expanded'
   this.sectionButtonClass = 'govuk-accordion__section-button'
   this.sectionHeaderClass = 'govuk-accordion__section-header'
@@ -70,6 +72,7 @@ function Accordion ($module, config) {
 
   this.sectionSummaryClass = 'govuk-accordion__section-summary'
   this.sectionSummaryFocusClass = 'govuk-accordion__section-summary-focus'
+  this.sectionContentClass = 'govuk-accordion__section-content'
 }
 
 // Initialize component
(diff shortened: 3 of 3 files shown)


colinrotherham:
Ah this is great, ties in nicely with the [`html-validate` (issue details)](https://gitlab.com/html-validate/html-validate/-/issues/193#note_1176783957) discussion on:

* https://github.com/alphagov/govuk-frontend/pull/3011#issuecomment-1315065073
* https://github.com/alphagov/govuk-frontend/pull/3014#issuecomment-1317583926
* https://github.com/alphagov/govuk-frontend/pull/3022#issuecomment-1319744218

Where HTML validation for `hidden="until-found"` was brought up too

36degrees:
> IE 10 and below don't support the `hidden` attribute at all, so a `display: none` rule would need to be applied for those browsers without compromising those that support `hidden="until-found"`.

I think we have prior art for this in the header component, and could do something similar here too?

https://github.com/alphagov/govuk-frontend/blob/14d3c645223dad050132a6635aaf0e0172bd1de2/src/govuk/components/header/_index.scss#L222-L225

querkmachine:
Unfortunately, seemingly not.

Applying `[hidden] { display: none }` overrides the UA's `content-visibility: hidden;`, breaking the `until-found` functionality in supporting browsers.

Trying something fancy with negation (like `[hidden]:not([hidden="until-found"])`) also doesn't work, as IE still sees the `until-found` value, even if it doesn't understand it. 

I think the solution here is to feature detect whether the browser supports `until-found` in JS and to apply different attribute values depending on whether it's supported or not. Going to experiment with that today.

36degrees:
Sounds sensible – from reading MDN it sounds like `hidden="until-found"` uses `content-visibility: hidden` so we might also be able to use `@supports`?

Something like

```scss
[hidden] {
  display :none;

  @supports (content-visibility: hidden) {
    content-visibility: hidden;  /* may not even be needed, as all browsers that support content-visibility: hidden probably have default UA styles? */
    display: inherit;
  }
}
```

We could also wait until 5.x when we no longer need to care about IE10!

querkmachine:
`@supports` seems to do it! 

It even words in "tweener" browsers—i.e. versions of Chrome released between `content-visibility` being added in version 85 and `hidden="until-found"` being added in 102—without any obvious side-effects.

36degrees:
Really nice improvement – It'd be great to get this shipped 🙌🏻 

Given this uses `content-visibility: hidden` we should just double check what impact (if any) this has on the accessibility tree / assistive tech. A quick Google suggests there _might_ be some gotchas to watch out for:

- https://marcysutton.com/content-visibility-accessible-semantics
- https://kittygiraudel.com/2021/02/17/hiding-content-responsibly/

So it might be worth comparing the behaviour with and without this change to check if anything's changed?

> This functionality would seemingly want to be mirrored to the [Details component](https://design-system.service.gov.uk/components/details/), however we are relying solely on the user agent's own functionality to provide the expand/collapse feature in modern browsers.

In my testing, the details element already takes care of this for us – if you search for terms hidden inside a details element, it'll expand automatically to reveal the content. Because our styling hooks on to the `open` attribute everything else just seems to work.

querkmachine:
@36degrees My earlier poke at this seemed to show that the accessibility tree was rendering and being updated as expected, though admittedly that was only checked in latest Chrome. The issues described in those articles were (seemingly) fixed in Chrome 90.

owenatgov:
Percy is being a bit odd. On edge it shows the accordions as default open. I'm unsure what could be causing this? Maybe some sort of version mish mash. I personally don't think this is a blocker for live but if we're able to figure out why this is happening that'd be neat.

querkmachine:
@owenatgov I think what I wrote in the initial content-y bit was my best guess as to why Percy is returning different:

> This change means that the point that accordion content is hidden is slightly different. Formerly it was hidden (effectively) immediately, based on the presence of the `js-enabled` class. It now requires both the class and for the accordion's JavaScript to have been initialised before anything is hidden.

Is Percy taking its screenshots before the bulk of Frontend's JS has kicked in? In the past, the accordion will already have been collapsed at this point (via CSS and JS inlined in the template), but now it won't until the accordion has been initialised.

owenatgov:
@querkmachine That'll be it. I don't think this is a showstopper but it probably is something for us to consider if we want to continue using percy and run into other cases where we move away from relying on `js-enabled`.

owenatgov (approved):
Code looking good.

Working across browsers.

Accessibility wise this seems to be operating as expected and from a bit of research, Chrome 90 is on the extreme low end of use.

Cracking work 💪🏻 

Something worth recording is that in IE9 the accordion looks a little odd with a big border at the bottom. See screenshot below. This was already there however so it's not a regression and it's not stopping users from reaching the content. It may even be captured somewhere else.

Screenshot 2022-12-07 at 14 20 07

colinrotherham:
> Is Percy taking its screenshots before the bulk of Frontend's JS has kicked in? In the past, the accordion will already have been collapsed at this point (via CSS and JS inlined in the template), but now it won't until the accordion has been initialised.

@querkmachine It's been fab reading through this PR 🎉 

Percy wise, we've just got the 100ms default `networkIdleTimeout` I believe

Can try dropping in a [**.percy.js** config](https://docs.percy.io/docs/cli-configuration) into the project root, seeing if a longer timeout helps settle things down? Does this mean users will see the same flicker through as sections close after init?

querkmachine:
@colinrotherham 

> Does this mean users will see the same flicker through as sections close after init?

Yes, but I personally think this isn't a problem and is actually an improvement.

Previously, if the stylesheet loaded but the network request for `all.js` (or equivalent) failed, the accordion's content would be hidden and inaccessible. You'd only have the section headings visible, and clicking on them doesn't do anything. You can emulate this by blocking the `all.js` request in browser devtools. 

Screenshot 2022-12-07 at 14 43 54

Now, the accordion's content remains visible up until the component JS has successfully downloaded and initialised.

colinrotherham:
Ah I see the benefit, I just know Google unfavourably ranks websites with high ["cumulative layout shift"](https://web.dev/cls/) as users prefer a stable layout when JavaScript kicks in. Having unclickable section headings is still a valid point though

PR #3059, opened 2022-11-30
[SPIKE] Experiment with alternate Exit this Page button positioning

querkmachine:
A happy little ***renegade*** draft PR so there's a deployed preview for folks to poke.

This updates the Exit this Page full-page example to place the EtP button in the area between the page header and main content area—similar to the typical placement of the Back Link, Breadcrumbs and Phase Banner components.

## Why?

This aims to explore a possible solution to some problems with the existing prototype:

### The use of `position: sticky` is incompatible with Frontend's grid system.

The existing EtP button was placed inside of the grid container, but couldn't be placed inside of a column as this limited the vertical space available to `position: sticky` to the height of the column (which would ostensibly be the same height as the button itself). 

Placement outside of a column led to the button being misaligned with other elements on the page, due to the grid container's use of negative left/right margins. 

This moves the EtP button outside of the grid container, allowing it to align with the page wrapper whilst maintaining the use of `position: sticky`.

### The use of `position: fixed` overlapped other page elements.

As the existing EtP button had to 'break out' of the grid container to appear at the top of the page using `position: fixed`, it had no awareness of other elements on the page. As a result, it covered up the page heading and GOV.UK logo, along with any other navigational elements present. 

By moving the EtP button to the top of the page, it no longer needs to 'break out' of its container and does not need to use `position: fixed`. 

### The 'two button' problem.

Currently there are two, separate EtP buttons on the page: A persistently visible one that sticks to the viewport, and a visible-on-focus 'skiplink' version. The skiplink version was introduced so that keyboard navigation users could access the EtP button by quickly navigating to the top of the page and accessing the skiplink. The existing button was located after the page's main content and was relatively time consuming to navigate to. 

This introduced some implementation concerns, as we would effectively need to create two visually and functionally separate EtP button components, and service teams would need to implement both in order to provide the desired accessibility coverage.

By moving the EtP button to be in this position, it is much closer to the start of the page (albeit not as close as a skiplink).  The placement is also consistently predictable, appearing directly prior to the page's unique `<main>` element.

## Other thoughts

No tests have been carried out yet to see how this plays alongside other components that use this space (e.g. Back Link, Breadcrumbs, Phase Banner).

We were using the old `position: fixed` behaviour to provide a fallback to Internet Explorer, which don't support `position: sticky`. This fallback no longer exists. However, reintroducing it would likely result in the overlapping issue reoccuring in IE, so is probably undesirable anyway. This cannot be polyfilled easily, as IE also does not support Intersection Observers.

It looks a bit weird, no longer being aligned with the content of the page. I also worry that this might make the button easier to miss, as it blends in more with the page header. We could probably do something with CSS to shift it around.


querkmachine:
Visual tests in combination with other components.

## Back link

Source order: EtP button → Back link.

Reversing the source order had no visual change on wide viewports.

|Narrow|Wide|
|:-:|:-:|
|Screen Shot 2022-12-01 at 11 05 12 Medium|Screen Shot 2022-12-01 at 11 05 04|

## Breadcrumbs

Source order: EtP button → Breadcrumbs.

Reversing the source order resulted in a larger spacing above the EtP button, pushing it to be below the bounding rectangle of the breadcrumbs component.

|Narrow|Wide|
|:-:|:-:|
|Screen Shot 2022-12-01 at 11 05 23 Medium|Screen Shot 2022-12-01 at 11 05 26|

## Phase banner

Source order: Phase banner → EtP button.

Reversing the source order resulted in the EtP button covering the right-hand side of the phase banner on wide viewports.

|Narrow|Wide|
|:-:|:-:|
|Screen Shot 2022-12-01 at 11 07 01 Medium|Screen Shot 2022-12-01 at 11 06 54|

christopherthomasdesign:
I think it broadly looks quite good, though I take the point on how it might stand out less because it blends with the header. It stands out to me, but 1) I can perceive red and blue pretty distinctly from each other, and 2) I spend all day looking at this website so notice any tiny difference to it.

Probably the main thing to work out with this idea is where we think the button should sit in the page structure, and if that affects the invisible ETP button at the very top of the page.

On a more detailed level, we might have to look at combinations of the components above, particularly phase banner + back button / breadcrumb. That's a fairly likely combo. On larger screens, the button ends up sort of inside the phase banner a bit, which looks a bit dodge:

Screenshot 2022-12-01 at 14 32 01


(That's if you put ETP before the phase banner. We'd need to work out the ideal order of these components too.)

I suspect positioning's gonna be tricky on this one whatever you do. We're pretty sure we want it to be towards the right, and to stick to the top as you scroll down. I think the ideal is for it to align vertically with the h1 of the page, like it does at the moment. But if we can solve other problems by moving the button higher up the page order, I think this is worth pursuing.

owenatgov:
Chris said the headlines for me. Position in the accessibility tree is going to impact our solution. Some other thoughts:

- I had a brief go at moving this after main so that we could keep the skip link but it obvs stuck to the bottom of main and we don't currently have an `afterContent` block in the template. It wouldn't be hard for us to add one but we would probably need to document it somewhere if we did.
- Something we could do is forgo sticky on desktop and have this just fixed on the right edge on desktop, ala https://enough.campaign.gov.uk/. It would look very odd on big screens but considering we're planning ot explain it, maybe it's not as detrimental. This would afford us more control over its location in the tree. We could even positioning it at the top of the body, solving the problem!
- Speaking of positioning the button at the top of the tree, I'm wondering if the skip link or some sort of more subtle button could instead be moved to the bottom of the tree. This would Mean users have sequential access to the button from the top or the bottom if they needed it. This of course isn't the intent of the skip link so maybe we need something else. If this is a solution we wanna explore, I'm not sure that extending the skip link is the right thing. I'll have a play around.
- Non-critical point that we're probably gonna get complaints from folks about axe failures now because the button isn't inside a region.

owenatgov (commented):
This is looking really good 💪🏻 

The only problem I've observed is that on IE it doesn't stick. That's to be expected as IE doesn't support `position: sticky` but until we drop IE support, I do think we need a backup. The only thing that comes to mind is `position: fixed` which will re-introduce the issue with the button covering the header on mobile, however my assessment is:

- We don't expect IE users to be on mobile
- Having access to the button is likely more important than having access to header controls

In terms of implementation, could you do a `@supports` trick?

querkmachine:
Good point on the IE/mobile thing, though there's still the risk that users are using a window or monitor narrow enough, a zoom level high enough, or some combination of both, that the mobile layout of the page is being seen on a desktop computer.

That said, the IE usage in GOV.UK stats are so low that we're—anecdotally—probably talking about one in the tens of thousands of people right now, so I'm happy to `@supports` it for modern browsers.

owenatgov (approved):

PR #3076, opened 2022-12-06
[Spike][WIP]Add size limit configuration

domoscargin:


domoscargin:
Transferring some random slack thoughts here:

* The associated Github issue is unverified, so we'd have some light hoop-jumping to do to get it approved by the @alphagov org
* Using the `webpack` and `webpack-why` plugins is probably the most interesting setup, as we can get an HTML report on things like included modules, duplicate modules, etc.
* In my examples, the govuk-esm/all.mjs is no longer working, for some reason (it did before)
* I think we'd have to write our own plugins for failing workflows on things like duplicate modules (anything other than file size/time to download)

Checking file size/time to download, and being able to dig into failures using `webpack-why` seems like a good improvement as a quick and dirty mvp. Things that may be trickier, but worthwhile:

* Long-term trends - we could save run datas to a spreadsheet or something in the roughest case
* Plugins for failing workflows for custom reasons (and actions to comment this on PRs?)

romaricpascal:
A couple of thoughts regarding tracking the size of our bundle:

- Which files should we run it in and what should we check regarding tree shaking? To be complete, feels like we should check `dist`, the compiled CJS files (to check if there's not a sudden bump in size due to surprise polyfills/transpiling) in the package and the "public" MJS files in package, both imported as a whole and each individual export. That feels like quite a lot though and may be complex to figure the exports for each MJS file (though we can always `import` in a node script)😬 
- do we want it to be failing checks or just informative for now (eg. [add a comment on the PR or something to help review](https://github.com/actions/github-script#comment-on-an-issue))?

I like the idea of running Statoscope on Webpack directly. In the end, we'll still have to handle failing the check/commenting whichever path we chose, so that'd be one less layer to run through than if we use `size-limit`, and maybe more control on options as well? I'm also quite intrigued by [their `--reference` option](https://github.com/statoscope/statoscope/tree/master/packages/cli) for comparing branches inside pull requests

romaricpascal:
Spike work is complete and perf work carries on in #3279

PR #3078, opened 2022-12-07
Allow nodemon to continue after `npm run serve` errors

colinrotherham:
This PR ensures Node.js errors (from Express) don't interrupt `gulp watch`

Follows the same approach as:

* https://github.com/alphagov/govuk-frontend/pull/3034
* https://github.com/alphagov/govuk-frontend/pull/3035

### DIFF
--- a/tasks/run.mjs
+++ b/tasks/run.mjs
@@ -1,5 +1,7 @@
 import { spawn } from 'child_process'
 
+import { isDev } from './task-arguments.mjs'
+
 /**
  * Spawns Node.js child process for npm script
  * rather than using Gulp
@@ -19,7 +21,12 @@ export async function npmScript (name, args = []) {
 
     // Emit errors to error listener
     script.stderr.on('data', (data) => {
-      script.emit('error', new Error(data.toString()))
+      const error = new Error(data.toString())
+
+      // Log errors (without rejection) in development
+      return isDev
+        ? console.error(error)
+        : script.emit('error', error)
     })
 
     // Reject on actual script errors to exit `gulp dev`


romaricpascal:
Do I undestand correctly that what was happening is `nodemon` logging to `stderr` and us turning that into an `error` event on the stream that closed everything? If that's the case, I'm wondering if that doesn't highlight that we're using output in `stderr` to be more than what they actually are. Maybe we should only pass them through like we do for `stdout` rather than interpret them as errors and leave it to non-zero exit codes to decide that something errored?

colinrotherham:
> Do I undestand correctly that what was happening is `nodemon` logging to `stderr` and us turning that into an `error` event on the stream that closed everything? If that's the case, I'm wondering if that doesn't highlight that we're using output in `stderr` to be more than what they actually are. Maybe we should only pass them through like we do for `stdout` rather than interpret them as errors and leave it to non-zero exit codes to decide that something errored?

Yeah it was a "no risk allowed" decision 😬 

I felt better knowing that every write to `stderr` will break the build unless:

1. Stream or task listens via `.on('error', handler)`
2. Stream has [`gulp-plumber`](https://github.com/floatdrop/gulp-plumber#readme) handling errors

I'm undecided though, maybe get a second opinion before we remove the fail-safe?

romaricpascal (approved):
I'm not really stuck on removing the check for `stderr`, it just felt like a workaround rather than what stderr is intended. If we feel safer with it, happy to keep it 😄 We can always discuss its removal in a future PR if that becomes an issue.

Issue #3063, opened 2022-12-01
Change the button activation combo for Exit this Page to 3 x shift


owenatgov:
## What
Change the keypresses that activate the EtP button from the esc key 3 times in a given timeframe to shift 3 times.

As part of this, investigate ways to "clear" the press 3 times only when shift is intentionally pressed 3 times in a row, not accidentally eg: if using it in tandem with a letter key to capitalise a word/letters.

## Why
This is a decision based on the outcomes of our spikes. See https://github.com/alphagov/govuk-design-system/issues/2402 for more details.

Part of https://github.com/alphagov/govuk-design-system/issues/1755

## Who needs to work on this
Developers

## Who needs to review this
Developers, Accessibility specialists

## Done when

- [x] We have a consistent way to activate the button using 3xshift that couldn't be activated accidentally
- [x] This solution has been implemented into the EtP branch

Issue #3061, opened 2022-12-01
Update screen reader announcement setup for Exit this Page


owenatgov:
## What
Make the following changes to the screen reader announcement behaviour for exit this page:

- Add an announcement for when the button keypress "timer" runs out
- Ensure that all announcements except the button activation use `role="status"`
- Ensure that the button activation announcement uses `role="alert"`
- Move the markup that acts as the source of the announcement to their respective locations
- Increase the timeout time for the keypress action. Times to try:
    - 5 seconds
    - 10 seconds

## Why
These are decisions based on the outcomes of our spikes. See https://github.com/alphagov/govuk-design-system/issues/2402 for more details.

Part of https://github.com/alphagov/govuk-design-system/issues/1755

## Who needs to work on this
Developers

## Who needs to review this
Accessibility specialists

## Done when

- [x] Keypress timeout announcement
- [x] `role="status"` on non-critical announcements
- [x] `role="alert"` on button activation
- [x] Markup for announcement located adjacent to the context of the announcement
- [x] Timeout time increase investigated

PR #3081, opened 2022-12-07
[SPIKE] Change 3x Escape detection to 3x Shift instead

querkmachine:
Changes the 3x Escape key functionality to use 3x Shift key instead, as per [#3053](https://github.com/alphagov/govuk-frontend/issues/3063).

This is intended to avoid problems we found where the Escape key was conflicting with other functions within the browser and operating system (including, but not limited to, exiting full page modes, halting page loading, closing dialogs and menus).

## Changes
The exit timer code now runs on instances of `Shift` and key code `16` instead of `Escape` and key code `27`.

The exit timer is immediately cancelled if a timer is currently running and the user presses a key that isn't Shift. This is to avoid accidental activation from the User Typing Normally.

The timer cancellation code has been abstracted into a new `resetEscTimer` function, as it is now called from multiple places. Variables and functions are still named with the `Esc` terminology to minimise code differences and make it easier to see what has actually been changed.

### Shift an alternative to Escape
Using Shift as an alternative to Escape appeared to work without side effects in the below browsers/OS combinations, and did not appear to accidentally activate any other browser or OS features when used.

* Chrome 109 / Windows 10
* Chrome 108 / macOS 12.6
* Chrome 105 / Android 13
* Firefox 107 / Windows 10
* Firefox 107 / macOS 12.6
* Safari 16.0 / macOS 12.6
* Safari 16.1 / iPadOS 16.1 (with attached keyboard)
* Internet Explorer 11 / Windows 7
* Internet Explorer 10 / Windows 7
* Internet Explorer 9 / Windows 7

### Shift in combination with other keys 
The Shift key still fires a keyup event when it's pressed in combination with another key. This makes sense, as the physical key is still being pressed and released, but means that `escCounter` is still incremented in these situations. 

Pressing 'Shift' alone fires one keyup event.
```
{ key: "Shift", keyCode: 16, shiftKey: false }
```

Pressing 'Shift' and 'A' fires two keyup events. The order these are fired depends on which order the user releases the keys, though the user must still be holding 'Shift' when they release 'A' if typing an uppercase letter, so they would typically appear in this order.

```
{ key: "A", keyCode: 65, shiftKey: true }
{ key: "Shift", keyCode: 16, shiftKey: false }
```

The chances of accidental activation of the page exit functionality remains low, due to the timer being reset on a non-Shift key subsequently being pressed, but it does result in the indicator lights flashing up unnecessarily.

#### VoiceOver

VoiceOver on Firefox 107 / macOS 12.6 did not announce when the Shift key was pressed.

## Todo
I've yet to test how this interacts with Sticky Keys in Windows, both with activation of Sticky Keys and how the Shift key works when it's in Sticky Keys mode. Sticky Keys is seemingly disabled in Browserstack. 

Haven't yet tested in other screen readers. 

Couldn't test in Internet Explorer 8 due to a JavaScript issue unrelated to this spike.


querkmachine:
**The tl;dr:**

* Shift works more reliably than Escape across the board and without side effects.
* The Shift keyup event is fired even if it's been pressed in combination with another key. This doesn't make accidental activation more likely, but it does introduce unnecessary screen reader spam.
* JAWS, for whatever reason, seems to reduce the number of key presses from 3 to 2. 
* Using the 3× Shift functionality with Slow Keys enabled is somewhat difficult in the current permitted timeframe, as there has to be a period of 'silence' between each key press.
* Using the 3× Shift functionality with Sticky Keys enabled is much more difficult in the current permitted timeframe, as they require the user to press Shift 2 to 3 times more than they would otherwise, and they may accidentally turn off assistive software functions in doing so.

owenatgov:
Incredible investigation @querkmachine. Some thoughts from my end:

- Technically, we no longer need to rely on `keyup` as the event since this was specifically to combat quirks with using esc. Would the behaviour be different if we used a different event like `keydown`? I would presume they're the same I'm saying that without investigating.
- Shift feels like an improvement over esc, however it still has problems. The other keys that comes to mind are ctrl or option/alt. We could apply the same check that they should only add to the counter when done so without another button being pressed.
- Screen reader noise is a bother, moreso with more functional keys like shift. I'm trying to think how we could also tie screen reader announcements to only one key being pressed at a time. I'm struggling to think of something concrete since keypress events only fire per key and you can't really detect when more than 1 key is being pressed without some hacky, unreliable code. The `KeyEvent` API does have [`shiftKey`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/shiftKey) and [`ctrlKey`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/ctrlKey) which could be useful here.
- RE: slow keys, https://github.com/alphagov/govuk-frontend/issues/3061 included a piece to increase the timeout time. Maybe it's worth investigating that here?

querkmachine:
My immediate thoughts, not done anything to spike or verify these yet:

1. I imagine `keydown` might make the activation faster for Sticky Keys users. Less sure for Slow Keys, as I think it fires both `keydown` and `keyup` only after the operating system controlled delay.
3. I don't think changing from Shift to another modifier key is likely to make a difference when it comes to detecting if pressed alone or if it's in combination with another key, as the keyboard events are being fired in response to physical key presses. I'd expect Option, Control, etc. to behave pretty much identically to Shift in that scenario.
4. I did consider trying to do something weird with `shiftKey`, such as keeping track of if the last keyboard event used `shiftKey`and ignoring the next subsequent Shift `keyup` event, but didn't persue it at the time. We might want to try that to counter the screen reader noise issue. This wouldn't be possible (I think?) if we changed to `keydown`, however.
5. I think increasing the timeout time is something we should probably do. We're building up quite a lot of cases where it doesn't seem to be long enough!

querkmachine:
I've changed the timeout from 2 seconds to 5 seconds. 

I've also added some code to record if the previous keyup event was `shiftKey: true` and, if so, to ignore the following Shift keyup event. This prevents the indicators from flashing up briefly, as well as avoiding unnecessary announcements by screen readers.

I've not changed the key away from Shift just yet, although that is probably something we want to explore in order to avoid conflicts with Sticky Keys.

querkmachine:
Trying it with Ctrl instead of Shift. 

The good stuff:

* **Ctrl seems to work in every browser without any obvious side effects.**
* **Sticky Keys and Slow Keys on macOS** work identically to before, except it is no longer possible to accidentally deactivate Sticky Keys by rapidly pressing the modifier key. Both are easier to accomplish now that the timeout has been increased to 5 seconds. 
* **Sticky Keys on Windows**: This works virtually identically to when using Shift, including the (somewhat frustrating) need to stagger keypresses so as to not 'lock' the Ctrl key when Sticky Keys is active. The increased timeout makes this easier to do, but it's still slow and easy to mess up.

The bad stuff:
* **JAWS** still immediately jumps to 'press 2 of 3' when it's running in any browser. Still no clue how or why this is the case. 
* **VoiceOver in Safari** apparently uses the Ctrl key to toggle VoiceOver announcement playback. Pressing Ctrl three times thus pauses, unpauses, then re-pauses announcements, in addition to activating the EtP button. It also (still) doesn't make the 'press n of 3' announcements. This wasn't exhibited in other browsers using VoiceOver.

Using the Ctrl key seems to be an improvement over Shift, with side effects either resolved or maintained, with only VoiceOver/Safari seeming to be worse-off from the change. 

Also, I failed to mention this before, but none of the screen readers I've tested have successfully announced the 
'Exit this page activated' announcement, instead moving directly onto announcing the browser state ('Loading...') and the title/content of the new page. NVDA—which shows a visual log of what has been and will be announced—doesn't appear to even queue it up.

owenatgov:
Thanks again @querkmachine!

From an off-github chat it looks like ctrl is an improvement but we're not out of the woods. Another key we could try is alt but that means we lose the easy way to track inflight key events using `ctrlKey` and `shiftKey`. I'm hoping there's a slicker way for us to fire the screen reader announcement that means we don't spam folks unnecessarily. If we erase the contents of the announcement spans after they're added to the announcement queue but before they're announced, does it "cancel" the announcement? I feel like we've discussed this somewhere but I can't remember.

That JAWS issue is also very weird and probably needs a bit of deeper investigation.

As for the button activation not being announced, I suspect part of this is that that we're not applying `role="alert"` to it on button activation which we plan to do as part of https://github.com/alphagov/govuk-frontend/issues/3061. I don't think we need to worry about that too much now.

querkmachine:
(Without having tried any code changes yet, but having done some minor experiments...) I think there are liable to be a few problems with Alt: 

## Alt codes
The Alt key is used (at least by Windows, I think also on some Linux distros) to insert characters that aren't usually accessible on the keyboard. This is done by holding down the Alt key whilst pressing multiple characters on the number pad. For example, holding <kbd>Alt</kbd> whilst typing <kbd>0</kbd><kbd>1</kbd><kbd>5</kbd><kbd>1</kbd> will insert the em dash (—) character, whilst holding <kbd>Alt</kbd> and typing <kbd>0</kbd><kbd>1</kbd><kbd>2</kbd><kbd>8</kbd> will return the Euro currency symbol: €.

Windows itself is delightfully weird in how it handles this in webpages. For example, although there is a simple keyboard shortcut for typing ß (<kbd>Alt</kbd> + <kbd>s</kbd>), pressing this key combination actually programatically 'types' <kbd>Alt</kbd> + <kbd>0</kbd><kbd>2</kbd><kbd>2</kbd><kbd>3</kbd>, and this is what our keyup event detects, _not_ the keys the user actually pressed. 

## Inserting diacritics
On macOS, the Alt key is used to insert characters featuring diacritic marks. Pressing <kbd>Alt</kbd> + <kbd>e</kbd> on macOS inserts the acute accent character: `´`.

The above two aren't blockers, and we can probably compensate for them in code, but they demonstrate that Alt is maybe a bit more tied into OS-level functions than we might otherwise expect. Now for the kicker:

## System menu shortcuts
On Windows, pressing the <kbd>Alt</kbd> key jumps the keyboard focus to the current application's menu. Pressing <kbd>Alt</kbd> in combination with a letter key jumps focus directly to a specific menu item. How browsers record and handle this behaviour is inconsistent. Based on my hurried investigations:

||Pressing <kbd>Alt</kbd> by itself|Pressing <kbd>Alt</kbd> and <kbd>F</kbd> (File menu)|
|:-|:-|:-|
|Chromium 108<br>(Chrome, Edge, etc.)|Registered the initial keyup event but no subsequent events.|Doesn't register any keyup event.|
|Firefox 108|Registers all keyup events.|Doesn't register any keyup event.|
|Internet Explorer 11|Registers all keyup events.|Doesn't register any keyup event.|

In all cases, the keyboard focus is moved outside of the webpage and to the browser UI.

In Chromium's case, the user cannot easily return to their previous focus position, and must navigate the webpage again from the top. In Firefox's case, pressing <kbd>Alt</kbd> toggles the keyboard focus between the menu and the webpage, restoring the previous focus position.

The behaviour in Chromium browsers probably makes using 3×Alt a non-starter on Windows systems.

## Documentation confusion

Unlike Esc, Shift, and the mostly similar Ctrl/Control keys, there isn't a lot of consistency in how the Alt key is labelled on keyboards. The physical key is labelled 'Alt' on Windows and some Mac keyboards, it's 'Option', 'Opt' or '⌥' on other Mac keyboards, some Mac keyboards show both 'Alt' _and_ 'Option', and what's up with this 'Alt Gr' key that some keyboards have? 

Alt feels like a fundamentally more difficult key to explain to users or to have be widely recognised because of this inconsistency in labelling.

querkmachine:
As for this:

> If we erase the contents of the announcement spans after they're added to the announcement queue but before they're announced, does it "cancel" the announcement?

I don't remember which screen reader it was, but this definitely doesn't happen in all of them. The announcement is still read out even after it's 'expired' and been removed from the DOM. The only thing that seems to definitively remove queued announcements is the webpage that spawned them being closed or navigated away from.

querkmachine:
In an attempt to really summarise the different keys we've tried so far:

||Esc|Shift|Ctrl|Alt[^Alt]|
|:-|:-|:-|:-|:-|
|Conflicts with browser/OS functionality|Yes[^EscConflict]|No|No|Yes[^AltConflict]|
|Affected by Slow Keys|Yes|Yes|Yes|Yes|
|Affected by Sticky Keys|No|Yes|Yes|Yes|
|Works in NVDA|Yes|Yes|Yes|Yes|
|Works in JAWS|Yes|Weird behaviour[^JAWS]|Weird behaviour[^JAWS]|?|
|Works in VoiceOver|Yes|Yes|Kinda[^VO]|?|
|Pressing counts as user activity|No[^EscActivity]|Yes|Yes|Yes|

[^Alt]: Use of Alt has not been as thoroughly researched as other options, as it was quickly deemed unfeasible due to its conflict with Windows system shortcuts. 

[^EscConflict]: Esc is often used to close modals and popovers, including those generated by system inputs like select elements. Esc is also used as a shortcut to stop a page from loading. On macOS, Esc is also used to exit full-page apps, a function which takes priority over JavaScript implementations.

[^AltConflict]: On Windows systems, pressing Alt is a shortcut to move keyboard focus to the current application’s menu bar. This cannot be prevented in JavaScript. As the user's focus is no longer within the webpage, subsequent key presses are not detectable by JavaScript.

[^JAWS]: When JAWS is active, for reasons unknown, the first keypress is not required. The user only needs to press the activation key twice to trigger EtP functionality. JAWS begins announcements at “press 2 of 3”.

[^VO]: By default, VoiceOver on macOS uses Ctrl as a shortcut to pause and resume VoiceOver announcements.

[^EscActivity]: Pressing Esc is not considered by the HTML spec to constitute user activity. Thus, trying to bind functionality to this key (especially an action like changing the page, which could be used in a user-hostile way) is not guaranteed to work, as it may be blocked by the browser or utilise different functionality.

querkmachine:
Following rebasing there is a new test that checks for the existence and validity of JSDoc comments, which is currently failing as the EtP work doesn't have these yet (this spike included). 

I'm going to make the executive decision to merge anyway, as adding JSDocs is not within the scope of this spike and should be done on the primary EtP branch instead.

PR #3087, opened 2022-12-09
Fix focus styles for links split over multiple lines in Chromium 108+ (Chrome 108+, Edge 108+, Opera 94+)

### LINKED ISSUE #3085: Links fail colour contrast on focus if they wrap to a new line
## Description of the issue
If a link is inline, and hits the end of the container forcing it to wrap to the next line, there is no border on the bottom when it gets focused using Chrome. By only having the yellow background it would fail WCAG 2.1 on colour contrast.

A screenshot of a link spanning multiple lines - it is highlighted in yellow.

Go to [Typography in the GOV.UK Design System](https://design-system.service.gov.uk/styles/typography/) use the keyboard to focus the link which spans multiple lines labelled: "read the ‘If your service is not on GOV.UK’ section on ‘Making your service look like GOV.UK’"

Expected it to have a black focus indicator like other links.

A screenshot of a link on a single line - it is highlighted in yellow and has a thick black border on the bottom.

- Operating system: Mac OS X 12.6
- Browser: Chrome
- Browser version: 108
- GOV.UK Frontend Version: 4.4.0

36degrees:
Chromium 108 includes [a change in behaviour for box decoration when a box is split over multiple lines, pages etc][1].

This change in theory brings Chromium in line with the [CSS Fragmentation Module Level 3 spec][2] but means that the box-shadow (which we rely on for colour contrast for the focus state) no longer appears on links that break over multiple lines.

This affects browsers based on Chromium 108 including Chrome 108, Edge 108 and Opera 94.

Add `box-decoration-break: clone` so that `box-shadow` is applied to each fragment independently, maintaining the existing behaviour in Chromium browsers.

## Before

A link with long enough text that it wraps over two lines. The link is focused, and has a yellow background. It is missing a black bottom edge (box-shadow) which is present on single-line links.

## After

A link with long enough text that it wraps over two lines. The link is focused, and has a yellow background, and has a black bottom edge (box-shadow) which matches the style used on single-line links.

Fixes #3085.

[1]: https://bugs.chromium.org/p/chromium/issues/detail?id=682173
[2]: https://www.w3.org/TR/css-break-3/#propdef-box-decoration-break

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,11 @@
 
 ### Fixes
 
+We've made fixes to GOV.UK Frontend in the following pull requests:
+
 - [#2998: Refactor back link and breadcrumb chevrons to use ems](https://github.com/alphagov/govuk-frontend/pull/2998)
-- [#3021 Printing style for current page link in header](https://github.com/alphagov/govuk-frontend/pull/3021) - thanks to [Malcolm Butler](https://github.com/MalcolmVonMoJ) for the contribution
+- [#3021: Printing style for current page link in header](https://github.com/alphagov/govuk-frontend/pull/3021) - thanks to [Malcolm Butler](https://github.com/MalcolmVonMoJ) for the contribution
+- [#3087: Fix focus styles for links split over multiple lines in Chromium 108+ (Chrome 108+, Edge 108+, Opera 94+)](https://github.com/alphagov/govuk-frontend/pull/3087)
 
 ## 4.4.0 (Feature release)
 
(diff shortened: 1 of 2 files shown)


colinrotherham (approved):
Brill, thanks for the co-author comment too 😊 

I checked the `autoprefixer` output too (just in case) and all good

Do we normally do a changelog entry too or one for another PR?

* https://github.com/postcss/autoprefixer/issues/344

abbott567:
Wow, ridiculously quick turn around! Bravo!

NickColley:
Great work!

Issue #2712, opened 2022-07-20
Refactor standard compile tasks to remove Gulp


domoscargin:
## What
Update `tasks/gulp/compile-assets.js` to replace the non-IE8 compile Gulp tasks with our own compile scripts. Updating the IE8 tasks depends on us removing that browser support, so has been separated into its own issue (#2713).

Currently, the `scss:compile` and `js:compile` tasks are building both app and dist files, depending on a flag. The pipelines for these files are similar, but not identical, and completely different files are built depending on the flag. We should split this work out and simplify this pipeline.

## Why
We have [made the decision to move away from Gulp](https://github.com/alphagov/govuk-frontend/issues/2558) and write our own build scripts.

## Who needs to work on this
Developers

## Who needs to review this
Developers

## Done when

- [x] IE8 tasks have been separated and run as is
- [x] App and dist compilations have been split into separate tasks
- [x] Non-IE8 compilation tasks no longer depend on Gulp
- [ ] Non-IE8 compilation tasks no longer depend on `taskArguments.destination`
- [x] docs/contributing/tasks.md has been updated
- [x] Dependencies are as up to date as is feasible (`oldie` is a blocker on a lot of these)
- [x] The tasks have been moved from `tasks/gulp` to `tasks`

domoscargin:
I'm going to close this in favour of https://github.com/alphagov/govuk-frontend/issues/2717 and I'll add some notes there about `taskArguments.destination`.

PR #3088, opened 2022-12-09
Use standalone Rollup task (no Gulp)

colinrotherham:
Friday tech debt PR

This PR moves our Rollup `compile:js` task into a separate file but also:

1. Removes all "gulp wrapper" legacy packages
2. Calls Rollup and UglifyJS via their JavaScript API (no Gulp)
3. Prepares us for Rollup v3 once we drop IE8

Also spotted for local development we were compiling all JavaScript files but only loading `all.js`

```html
```

That's now faster, plus we can remove some super old packages:

```
gulp-better-rollup
gulp-if
gulp-uglify
```

Ticks off some tasks from:

* https://github.com/alphagov/govuk-frontend/issues/2712


tasks/compile-javascripts.mjs
```diff
@@ -30,6 +30,11 @@ export async function compileJavaScripts () {
   } catch (cause) {
     throw new PluginError('compile:js', cause)
   }
+
+  // Write VERSION.txt file
+  if (isDist) {
+    await writeFile(join(destination, 'VERSION.txt'), pkg.version + EOL)
+  }
```
romaricpascal, line 37: **issue** I'm not sure it should be the responsibility of the JavaScript compilation to write the 'VERSION.txt' (nor the CSS compilation, or the compilation itself). We had a neat separation of concerns by having it as a separate task and I'm don't think there's any harm in keeping it that way (especially as it's a tangent matter to compiling rollup).

**suggestion** The existing asset-version task would benefit from the new way to access the version package that you added though 🙌🏻

tasks/compile-javascripts.mjs
```diff
@@ -126,10 +126,10 @@ export async function getModuleEntries () {
   const destPath = isPackage ? join(destination, 'govuk') : destination
 
   // Perform a search and return an array of matching file names
-  // but for 'dist' we only want compiled 'all.js'
-  const modulePaths = await getListing(srcPath, isDist
-    ? 'all.mjs'
-    : '**/!(*.test).mjs'
+  // but for 'dist' and 'public' we only want compiled 'all.js'
+  const modulePaths = await getListing(srcPath, isPackage
```
romaricpascal, line 135: Neat little update 🙌🏻

tasks/compile-javascripts.mjs
```diff
@@ -84,6 +90,32 @@ export async function compileJavaScript ([modulePath, { srcPath, destPath }]) {
   ])
 }
 
+/**
+ * Minify JavaScript ESM to CommonJS helper
+ *
+ * @param {string} modulePath - Relative path to module
+ * @param {object} result - Generated bundle
+ * @param {string} result.code - Source code
+ * @param {import('magic-string').SourceMap} result.map - Source map
```
romaricpascal, line 104: **nitpick** Looks like we're only importing `magic-string` to get that type. As we're looser with type checking on the tasks, couldn't we do without it and use `any` for now waiting for an upgrade to newer versions of rollup to bring their types (which hopefully they do). Feels like we're forcing a modern approach (typing) on legacy tools (rollup 0.56.x).

**thought** Wondering if there'd be an alternative route that'd save the typing by porting the task to a `rollup.config.js` file and using `rollup-plugin-uglify` and using Rollup's CLI directly? But maybe this has constraints due to our old version and some performance downsides (in case some stuff runs in parallel a bit in the current way?).

config/paths.js
```diff
@@ -1,5 +1,7 @@
 const { dirname, join } = require('path')
 
+const pkg = require('../package/package.json')
```
romaricpascal, line 3: Nice trick to not `readFile` and `JSON.parse` and all 🙌🏻

romaricpascal (commented):
Looking pretty neat! 🙌🏻 Was surprised to not see a `rollup.config.js` file popping up to be directly used by the CLI and instead have it in our own tasks. But it does do the job 😄 

I think it's been a little overreaching in removing the asset-version task though, the separation of concern by having it on its own was pretty clean.

tasks/compile-javascripts.mjs
colinrotherham: Yeah it was a safety thing from me

1. Rollup uses `SourceMap` via [`magic-string`](https://github.com/rollup/rollup/blob/v0.56.5/src/rollup/index.ts#LL14-L14C41)
2. UglifyJS uses plain `string` maps
3. Terser uses `EncodedSourceMap` via [`@jridgewell/gen-mapping`](https://github.com/jridgewell/gen-mapping/blob/main/src/types.ts#L12)

Wanted to be confident that the source map object had a `toString()` method

I'd love a Rollup config but thought we'd best wait for Rollup v3

tasks/compile-javascripts.mjs
colinrotherham: Ah don't worry, I can put back some of those 400+ removed lines of code 😉

tasks/compile-javascripts.mjs
colinrotherham: They were pretty advanced even for their old release (see [**TypeScript declarations**](https://github.com/rollup/rollupjs.org/blob/890687b8556b6134a0389284fb5fde71be8206a4/guide/en/04-javascript-api.md#typescript-declarations))

tasks/compile-javascripts.mjs
romaricpascal: Cheers 😄

colinrotherham: Moved it here:

* https://github.com/alphagov/govuk-frontend/pull/3114

colinrotherham:
@romaricpascal This is ready for review again

Split out a couple of unrelated changes:

* https://github.com/alphagov/govuk-frontend/pull/3113
* https://github.com/alphagov/govuk-frontend/pull/3114

config/paths.js
colinrotherham: Moved it here:

* https://github.com/alphagov/govuk-frontend/pull/3113

package.json
```diff
@@ -47,15 +47,12 @@
     "front-matter": "^4.0.2",
     "glob": "^8.0.3",
     "gulp": "^4.0.2",
-    "gulp-better-rollup": "3.1.0",
```
domoscargin, line 50: :tada:

package.json.unit.test.js
```diff
@@ -2,16 +2,13 @@ const packageLockJson = require('./package-lock.json')
...
-    it('should be pinned to 3.1.0 in package.json', () => {
-      expect(packageJson.dependencies['gulp-better-rollup']).toEqual('3.1.0')
+  describe('rollup', () => {
+    it('should be pinned to 0.56.5 in package.json', () => {
+      expect(packageJson.dependencies.rollup).toEqual('0.56.5')
     })
-  })
 
-  describe('rollup', () => {
     it('should be pinned to 0.56.5 in package-lock.json', () => {
```
domoscargin, line 10: Similar comment here in that we're probably able to bump this to 0.59.4.

colinrotherham: Consider it bumped 🎉

lib/helper-functions.js
```diff
@@ -39,19 +41,23 @@ function componentNameToJavaScriptClassName (componentName) {
...
- * Convert component name to JavaScript UMD module name
+ * Convert component path to JavaScript UMD module name
  *
- * Used by rollup to set the `window` global and UMD/AMD export name
+ * Used by Rollup to set the `window` global and UMD/AMD export name
  *
- * Component names are kebab-cased strings (button, date-input), whilst module
- * names have a `GOVUKFrontend.` prefix and are PascalCased
- * (GOVUKFrontend.Button, GOVUKFrontend.CharacterCount).
+ * Component paths have kebab-cased file names (button.mjs, date-input.mjs),
```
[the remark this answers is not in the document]
colinrotherham: I've just pushed some good path examples to the tests if that's alright? https://github.com/alphagov/govuk-frontend/pull/3088/commits/55e0ac0c854b25a0bbc92226a1b7ce235d158270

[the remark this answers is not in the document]
domoscargin: Nice!

domoscargin (approved):
🔥 

Thanks for popping in the tests - they're great! Getting rid of old gulp plugins is great! This is great!

PR #3094, opened 2022-12-12
Accordion margin/padding fixes

colinrotherham:
Just a quick PR to fix a few tiny Accordion issues we've had for a long time in:

1. **"Show all" button `margin-bottom` fix applied too wide**
Moves [legacy bug fix](https://github.com/alphagov/govuk-frontend/commit/879e37c85e4ff1156e1febb609afacc4df9de1bd) to 'tablet' (from 'desktop') which align with font size

2. **Section button `padding-bottom` jumps 10px taller on mobile**
Notice how the section button height appears to get taller (unlike tablet, desktop) 

3. **Trailing space in `class` attribute**
Now using `{%-` to trim whitespace as we do elsewhere
```html
<div class="govuk-accordion__section ">
```

Accordion height

See original Accordion design PR

* https://github.com/alphagov/govuk-frontend/pull/2257

Plus a minor fix where accordion sections lack padding (during slow page loads) since:

* https://github.com/alphagov/govuk-frontend/pull/3053

## Reducing layout shift
Whilst we know additional layout shifts take place when the accordion controls and section headers are injected, I've shifted `.govuk-accordion__section-content` content padding so it's applied whilst the JavaScript loads

Our Accordion sections are now open during page load, where previously we only added padding when `[hidden]` was added by component `init()` but in preparation for the `--expanded` modifier being toggled

Accordion section padding shown before and after component initialisation

### DIFF
--- a/src/govuk/components/accordion/_index.scss
+++ b/src/govuk/components/accordion/_index.scss
@@ -46,21 +46,25 @@
       padding-top: 0;
     }
 
+    .govuk-accordion__section-content {
+      @include govuk-responsive-padding(3, "top");
+      @include govuk-responsive-padding(8, "bottom");
+    }
+
     // Manually apply display: none to browsers that don't have a user agent
     // style for it, but override it with content-visibility if the browser
     // supports that instead.
     .govuk-accordion__section-content[hidden] {
       display: none;
+
       @supports (content-visibility: hidden) {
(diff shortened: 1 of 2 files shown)


colinrotherham:
Thanks @querkmachine 🙌

Issue #999, opened 2018-09-14
If GOV.UK Frontend fails to initialise but the end-user has JavaScript enabled components will fail


nickcolley:
<img width="1493" alt="js-blocked" src="https://user-images.githubusercontent.com/2445413/45561491-ce655d00-b83f-11e8-986c-0e7d1a35b547.png">

It is possible for components to be in a completely broken state if the user has JavaScript enabled but the GOV.UK Frontend bundle fails to download.

Previously we detected the presence of `GOV.UK` globally on window, but we removed this since we could not be sure if people would be using bundlers which will not output a global.

https://github.com/alphagov/govuk_template/blob/master/source/views/layouts/govuk_template.html.erb#L113

We could consider, injecting a global whenever a component is loaded that can remove js-enabled.

Potentially something like:

```js
if (not already injected) {
   window.___GOVUKFrontendComponentInitialised___
}
```

```js
// tabs.js
import 'inject-global-initialised'
// only injects if not already injected before
inject()

// rest of tabs code here
```

```html
// template.njk

<script>if(!window.___GOVUKFrontendComponentInitialised___) { // remove js-enabled }</script>
</body>
```

https://trello.com/c/oRDkiAz9/1448-frontend-components-appeared-styled-when-js-blocked

nickcolley:
Potentially by solving https://github.com/alphagov/govuk-frontend/issues/1127 we could then use that to solve this issue.

Since they both rely on knowing of instances of GOV.UK Frontend components on the page.

edwardhorsford:
We had a similar issue in my service this week. For _reasons_ our service's js file for IE failed to compile so our service's js wasn't available in IE - but we still had the `js-enabled` class. This caused the worst of both worlds - fields with conditional content were hidden and there was no way to get to them.

If there are recommended solutions - like testing for a variable I'd be keen to hear them.

shaquillaevelyn:
Hello all,

Currently experiencing this issue- what was the fix for this?

Thanks

> <img alt="js-blocked" width="1493" src="https://user-images.githubusercontent.com/2445413/45561491-ce655d00-b83f-11e8-986c-0e7d1a35b547.png">
> 
> It is possible for components to be in a completely broken state if the user has JavaScript enabled but the GOV.UK Frontend bundle fails to download.
> 
> Previously we detected the presence of `GOV.UK` globally on window, but we removed this since we could not be sure if people would be using bundlers which will not output a global.
> 
[...]

PR #3095, opened 2022-12-12
Hide Accordion content (again) during `.js-enabled` page load

colinrotherham:
This PR investigates an Accordion change to avoid a visual layout shift

During a slow page load (with `js-enabled` class on **`<body>`**) we now see Accordion sections jump from open to closed after component initialisation. We're not sure if this is preferred for users.

For example, shown when adding a 500ms delay [before our code runs](https://github.com/alphagov/govuk-frontend/blob/main/app/views/layouts/_generic.njk#L53-L54)

```html
<script>
  // Add 500ms delay before our code runs
  // (For example, analytics might run first !!)
  setTimeout(function () {
    window.GOVUKFrontend.initAll()
  }, 500)
</script>
```

For discussion, I've included a quick tweak which keeps all the enhancements from:

* https://github.com/alphagov/govuk-frontend/pull/3053

Accordion loading delay

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,10 @@ We've updated the Accordion component to use the new [`hidden="until-found"` att
 
 This allows the browser's native 'find in page' functionality to search within and automatically open sections of the accordion. Currently, this functionality is only supported by recent versions of Google Chrome, Microsoft Edge and Samsung Internet.
 
-This was added in [pull request #3053: Enhance the Accordion component with `hidden='until-found'`](https://github.com/alphagov/govuk-frontend/pull/3053). 
+This was added in pull requests:
+
+- [#3053: Enhance the Accordion component with `hidden='until-found'`](https://github.com/alphagov/govuk-frontend/pull/3053)
+- [#3095: Hide Accordion content (again) during `.js-enabled` page load](https://github.com/alphagov/govuk-frontend/pull/3095)
 
 #### Source maps for precompiled files
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ We've made fixes to GOV.UK Frontend in the following pull requests:
 - [#2998: Refactor back link and breadcrumb chevrons to use ems](https://github.com/alphagov/govuk-frontend/pull/2998)
 - [#3021: Printing style for current page link in header](https://github.com/alphagov/govuk-frontend/pull/3021) - thanks to [Malcolm Butler](https://github.com/MalcolmVonMoJ) for the contribution
 - [#3112: Remove unused `classList` polyfill from header component JavaScript](https://github.com/alphagov/govuk-frontend/pull/3112)
+- [#3094: Fix Accordion margin/padding inconsistencies](https://github.com/alphagov/govuk-frontend/pull/3094)
 - [#3150: Add missing `Event` polyfill to accordion component JavaScript](https://github.com/alphagov/govuk-frontend/pull/3150)
 
 ## 4.4.1 (Fix release)
--- a/src/govuk/components/accordion/_index.scss
+++ b/src/govuk/components/accordion/_index.scss
@@ -46,17 +46,18 @@
       padding-top: 0;
     }
 
+    // Hide the body of collapsed sections by default for browsers that lack
+    // support for `content-visibility` paired with [hidden=until-found]
     .govuk-accordion__section-content {
+      display: none;
+
       @include govuk-responsive-padding(3, "top");
       @include govuk-responsive-padding(8, "bottom");
     }
 
-    // Manually apply display: none to browsers that don't have a user agent
-    // style for it, but override it with content-visibility if the browser
-    // supports that instead.
+    // Hide the body of collapsed sections using `content-visibility` to enable
+    // page search within [hidden=until-found] regions where browser supported
     .govuk-accordion__section-content[hidden] {
-      display: none;
-
       @supports (content-visibility: hidden) {
         content-visibility: hidden;
         display: inherit;
@@ -65,6 +66,11 @@
       // Hide the padding of collapsed sections
       padding-top: 0;
       padding-bottom: 0;
+    }
+
+    // Show the body of expanded sections
+    .govuk-accordion__section--expanded .govuk-accordion__section-content {
+      display: block;
     }
 
     .govuk-accordion__show-all {


querkmachine:
Still antsy about this, really, as although it reduces the layout shift on load, it's still shipping something that doesn't work properly if JavaScript fails. Doesn't feel very progressive enhancement of us.

As that's a bigger problem to fix across multiple components though, I'll just sit on my hands for now.

owenatgov:
I agree with @querkmachine that this isn't ideal. The risk of users not being able to see the content is, IMO, more damaging than them seeing a flash of content. We do do this elsewhere where we interact with `js-enabled` and sort of assume that the component will initialise so there's precedence for it but in those instances eg: [the header](https://github.com/alphagov/govuk-frontend/blob/8a9e006b60327366d157d2d1a5e0f5c16a428391/src/govuk/components/header/_index.scss#L218-L226). In that instance though it defaults the header menu to be open on mobile, meaning users can still access content.

After ticking this over I unfortunately don't have any quick alternatives. It is technically the same as v4.4.0 and it allows us to ship a neat enhancement, so whilst it's a problem it's not explicitly a regression. I'm gonna wait and see if anyone else has opinions but the options I see right now are:

1. Merge this and make a note on https://github.com/alphagov/govuk-frontend/issues/999 as additional evidence
2. Close this and revert https://github.com/alphagov/govuk-frontend/pull/3053

Im leading towards 1. If nobody has any other feelings I'll approve this in the afternoon.

owenatgov (dismissed):

owenatgov (commented):
I'm not sure if this needs a changelog entry or not. I'm leaning towards "yes" unless somebody thinks otherwise.

colinrotherham:
Thanks @owenatgov

Knowing that this could go either way we did a minor padding fix in this one:

* https://github.com/alphagov/govuk-frontend/pull/3094

(So should we keep the flicker, at least it's a beautiful flicker 😊)

Had a quick run through [Tabs](https://design-system.service.gov.uk/components/tabs/) and there's another interesting middle ground. With slow-loading JS you can still click a tab and it'll (eventually) catch up via `window.location.hash`

Will be a shame to merge this PR but gives us more time to fully consider:

1. JavaScript off
2. JavaScript on but crashed
3. JavaScript on but skipped checks
4. JavaScript on but took too long

* https://github.com/alphagov/govuk-frontend/issues/999

owenatgov:
@colinrotherham That PR you've linked needs a changelog entry as well!

I initially wondered if this should have a change entry on the basis that this was technically a revision of an existing PR, however I'm not quite confident that it does.

colinrotherham:
> @colinrotherham That PR you've linked needs a changelog entry as well!

@owenatgov Ahh the padding fix will soon be hidden, can't do a change entry if the fix never gets rendered 😉 

I'll do a quick one though for the mobile height, button margin and (maybe) trailing attribute space?

querkmachine:
IMO this PR doesn't need a separate changelog entry as it's basically a continuation of the `hidden="until-found"` PR. Could reference both PRs against the same feature, I suppose?

colinrotherham:
@owenatgov @querkmachine @claireashworth Can you check these two new commits for me please?

* [`ec27262` **Add CHANGELOG note about accordion section visibility during load**](https://github.com/alphagov/govuk-frontend/pull/3095/commits/ec27262e788e935cc0e78bfe0a7e4af89c2d7a60)
* [`6d08bd` **Hide Accordion content (again) during `.js-enabled` page load**](https://github.com/alphagov/govuk-frontend/pull/3095/commits/6d08bd2498a6a8338a4850f711512755d08a4292)

Or see the full [**CHANGELOG.md** on this branch](https://github.com/alphagov/govuk-frontend/blob/accordion-default-visibility/CHANGELOG.md)

CHANGELOG.md
```diff
@@ -10,7 +10,14 @@ We've updated the Accordion component to use the new [`hidden="until-found"` att
...
 
-This was added in [pull request #3053: Enhance the Accordion component with `hidden='until-found'`](https://github.com/alphagov/govuk-frontend/pull/3053). 
+This was added in pull requests:
+
+- [#3053: Enhance the Accordion component with `hidden='until-found'`](https://github.com/alphagov/govuk-frontend/pull/3053)
+- [#3095: Hide Accordion content (again) during `.js-enabled` page load](https://github.com/alphagov/govuk-frontend/pull/3053)
+
+Note: To improve the reading experience on slow loading pages we initially removed `display: none` from each Accordion section. To avoid a flicker of content, we've chosen to restore the current behaviour in [pull request #3095](https://github.com/alphagov/govuk-frontend/pull/3053).
+
+We’d like to take another look at how components load in: [issue #999: Components failing initialisation break due to `.js-enabled` styles](https://github.com/alphagov/govuk-frontend/issues/999).
```
owenatgov, line 20: I don't think this line is necessary. I think you're being considerate in including it, however we do reference this issue in this PR and users can come here to see the discourse if they are interested in the issue.

CHANGELOG.md
```diff
@@ -10,7 +10,14 @@ We've updated the Accordion component to use the new [`hidden="until-found"` att
 
 This allows the browser's native 'find in page' functionality to search within and automatically open sections of the accordion. Currently, this functionality is only supported by recent versions of Google Chrome, Microsoft Edge and Samsung Internet.
 
-This was added in [pull request #3053: Enhance the Accordion component with `hidden='until-found'`](https://github.com/alphagov/govuk-frontend/pull/3053). 
+This was added in pull requests:
+
+- [#3053: Enhance the Accordion component with `hidden='until-found'`](https://github.com/alphagov/govuk-frontend/pull/3053)
+- [#3095: Hide Accordion content (again) during `.js-enabled` page load](https://github.com/alphagov/govuk-frontend/pull/3053)
+
+Note: To improve the reading experience on slow loading pages we initially removed `display: none` from each Accordion section. To avoid a flicker of content, we've chosen to restore the current behaviour in [pull request #3095](https://github.com/alphagov/govuk-frontend/pull/3053).
```
owenatgov, line 18: I don't think this line is necessary. You already reference this PR on line 16 and I think providing the link is enough.

CHANGELOG.md
```diff
@@ -10,7 +10,14 @@ We've updated the Accordion component to use the new [`hidden="until-found"` att
 
 This allows the browser's native 'find in page' functionality to search within and automatically open sections of the accordion. Currently, this functionality is only supported by recent versions of Google Chrome, Microsoft Edge and Samsung Internet.
 
-This was added in [pull request #3053: Enhance the Accordion component with `hidden='until-found'`](https://github.com/alphagov/govuk-frontend/pull/3053). 
+This was added in pull requests:
+
+- [#3053: Enhance the Accordion component with `hidden='until-found'`](https://github.com/alphagov/govuk-frontend/pull/3053)
+- [#3095: Hide Accordion content (again) during `.js-enabled` page load](https://github.com/alphagov/govuk-frontend/pull/3053)
```
owenatgov, line 16: Wonky link

```suggestion
- [#3095: Hide Accordion content (again) during `.js-enabled` page load](https://github.com/alphagov/govuk-frontend/pull/3095)
```

owenatgov (commented):
A few comments on the changelog, mostly making the content leaner in line with how we generally write up changes. I'll let @claireashworth come in with further thoughts on the content, although it might make sense to instead do this as part of the complete 4.4.1 changelog review.

CHANGELOG.md
colinrotherham: Good spot

owenatgov:
We're pausing this PR in favour of getting a fix release out in order to get a critical chrome bug released and give ourselves a bit more time to discuss the tradeoff's of this PR and it's interaction with https://github.com/alphagov/govuk-frontend/pull/3053.

36degrees:
> We're pausing this PR in favour of getting a fix release out in order to get a critical chrome bug released and give ourselves a bit more time to discuss the tradeoff's of this PR and it's interaction with #3053.

We discussed this team today and agreed that we want to take a proper look at #999, but in the meantime we're happy to merge this as it basically maintains the 'status quo'.

owenatgov (approved):
Does the job 👍🏻

PR #3096, opened 2022-12-13
[DO NOT MERGE] Set up Generics to type `nodeListForEach`

romaricpascal:
## What

This PRs demonstrates how to avoid unnecessary defensiveness by letting TypeScript properly follow the trail of types and match the type of element passed to `nodeListForEach` to the type expected by its iterator.

## Why

`nodeListForEach` uses only `Element` as a type for the content of its nodelist and the arguments of its iterator. This requires us to only pass functions expecting an `Element` as an iterator. If we try to pass something more specialised, like a function that'd expect an `HTMLInputElement`, TypeScript will complain at us as it cannot guarantee that that specialised iterator will receive the right type of `Element`, only that it's an `Element`.

The signature of these iterator is generic, but what they do inside will usually try to manipulate that specific kind of element (for ex, accessing the input's `form` property). This forces us to add a quick `instanceof` check inside the iterator as a safety that the element is of the right type before using it. Unnecessarily defensive, as we know for certain that these elements are the right type because of the selectors we use. 😭


src/govuk/common/index.mjs
```diff
@@ -162,9 +163,10 @@ export function extractConfigByNamespace (configObject, namespace) {
 }
 
 /**
+ * @template {Node} ElementType
```
romaricpascal, line 166: The same Generics system as for `nodeListForEach`allows to define iterators expecting specific kinds of elements as their first parameter (as well as the `NodeList` they receive as last parameter). For example,  a `nodeListIterator<HTMLInputElement>` that's meant to iterate on `<input>` elements.

That allows TypeScript to keep the trail of types when using `nodeListForEach`, and receive methods meant to iterate on sub-classes of `Element`. That's something it couldn't do before, as TypeScript would see `nodeListForEach` trying to run an iterator expecting `HTMLInputElement`, but could only acertain that it would pass it an `Element`, so bring that to our attention (and force us to use a wider type for the iterator and an `instanceof` to narrow down within the function).

src/govuk/components/checkboxes/checkboxes.mjs
```diff
@@ -17,6 +17,7 @@ function Checkboxes ($module) {
     return this
   }
 
+  /** @type {NodeListOf<HTMLInputElement>} */
```
romaricpascal, line 20: Thanks to the generic types, we can now set types on the `NodeList` being iterated on early, and know they'll carry through...

src/govuk/components/checkboxes/checkboxes.mjs
```diff
@@ -109,13 +110,9 @@ Checkboxes.prototype.syncAllConditionalReveals = function () {
  * Synchronise the visibility of the conditional reveal, and its accessible
  * state, with the input's checked state.
  *
- * @param {Element} $input - Checkbox input
+ * @param {HTMLInputElement} $input - Checkbox input
```
romaricpascal, line 113: ... allowing us to type the class methods more specifically ...

src/govuk/components/radios/radios.mjs
```diff
@@ -161,14 +162,10 @@ Radios.prototype.handleClick = function (event) {
     /**
      * Loop through radios
      *
-     * @param {Element} $input - Radio input
+     * @param {HTMLInputElement} $input - Radio input
```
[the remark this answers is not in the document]
colinrotherham: Yeah it's all a lot more transparent once we can use `Array.from()` on a **NodeList** 👍 

```mjs
var $inputs = document.querySelectorAll('input')

for (const $input of Array.from($inputs)) {
  // $input instanceof HTMLInputElement
}
```

src/govuk/common/index.mjs
romaricpascal: I wasn't aware it had to be the first line. Just put it there as it made sense to announce first that the block was using generics then use generics 😄

src/govuk/components/checkboxes/checkboxes.mjs
romaricpascal: I love this "typed-query-selector". Very happy to have that instead of the comments 😄 Why would you prefer waiting for the TypeScript issue to be resolved? 

If we want to wait though, I'd prefer the casting comments. Yes, they can have human errors (just like our tests can as well), but:
- them being close to the selector means gives them a chance to be picked on in code review
- it's one of those cases were we're actually certain of the type, so the `instanceof` is unnecessary defensiveness for the sake of TypeScript happiness.

colinrotherham: Yeah give the extended types a try then, might not even need `@type`

(Abandoned Gulp plugins have ruined me, all I'll say for avoiding extra packages 😆)

colinrotherham:
@romaricpascal To assist with code reviews I've split up the JSDoc and type linting changes into:

4. https://github.com/alphagov/govuk-frontend/pull/3104

You'll see this branch has been rebased too 👍

romaricpascal:
Cheers for the rebase 😄 I've added `typed-query-selector` and it worked pretty neatly so happy days, I'd say 😄

colinrotherham:
> Cheers for the rebase 😄 I've added `typed-query-selector` and it worked pretty neatly so happy days, I'd say 😄

@romaricpascal Amazing. Coming soon to a cherry pick near you

colinrotherham:
@romaricpascal Thanks again for this

I've added your commits (fixes in one, types in another) to:

Hope you didn't mind mind but I `fixup` a few more in 👍 

Bit frustrating that we can't remove more but [`typed-query-selector`](https://www.npmjs.com/package/typed-query-selector) won't understand this string (below) until we use [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) (backticks). Makes sense because `hash` could append another child selector 😬 

```js
'a.govuk-tabs__tab[href="' + hash + '"]'
```

Hit a gotcha though! For **ErrorSummary()** the `$input` element could be a select/textarea/upload etc 😮 

(So had to keep that one a bit more generic)

romaricpascal:
Closing now the commits have been cherry-picked in the actual PRs 😄

Issue #2513, opened 2022-01-20
Some component JavaScript is complex and difficult to follow


vanitabarrett:
### Cause

Some of our component JavaScript is quite complex, for example: there's a lot of DOM manipulation and managing state. We've also iterated component JavaScript over time and in some cases added code to already complex functions, which has led to some functions doing multiple things.

### Consequences

As a result, some of our JavaScript functions are difficult to follow and the naming of our functions can be misleading. This makes us less confident when debugging and making changes.

### Impact of debt

Low

### Effort to pay down

Medium

### Overall rating

Medium

### Reason (overall rating)

This doesn't cause us frequent issues, but it reduces our confidence working on our JavaScript code. I've rated this as a Medium to pay down because it's tied in with [adding unit testing/revisiting our testing strategy](https://github.com/alphagov/govuk-frontend/issues/2502) - ideally, we'd have smaller functions so that we can add unit tests for our JavaScript, but it would also be nice to have better tested JavaScript so we're confident doing our refactoring - a bit of a Catch-22.

There might also be some upskilling if the code is complex due to things like managing state. There may be some research to do into best practice here.

romaricpascal:
Part of solving the issue could be to add type annotations to our code, which is tracked by #3230.

36degrees:
We've made some good progress since this ticket was opened, including introducing JSDoc (as suggested by @querkmachine).

Next we want to take a proper look at the code with a view to refactoring it once we've moved to ES6 classes, which is covered by #2560.

On that basis I think we can close this.

PR #3103, opened 2022-12-15
Configure ESLint to require JSDoc `@params` etc

colinrotherham:
This PR follows up https://github.com/alphagov/govuk-frontend/pull/2913 (optional JSDoc) by making `@param` tags mandatory for methods that use JSDoc

It also configures ESLint excluded files correctly using `excludedFiles`

Which continues work we started for v4.4.0:

* https://github.com/alphagov/govuk-frontend/pull/2997

To assist with code reviews I've split these changes into:

4. https://github.com/alphagov/govuk-frontend/pull/3123
5. https://github.com/alphagov/govuk-frontend/pull/3104

Resolves https://github.com/alphagov/govuk-frontend/issues/2513

### DIFF
--- a/src/.eslintrc.js
+++ b/src/.eslintrc.js
@@ -7,7 +7,20 @@ module.exports = {
         browser: true
       },
       rules: {
-        'no-var': 'off'
+        'no-var': 'off',
+
+        // JSDoc blocks are mandatory
+        'jsdoc/require-jsdoc': [
+          'error', {
+            enableFixer: false,
+            require: {
+              ClassDeclaration: true,
+              ClassExpression: true,
+              FunctionExpression: true,
+              MethodDefinition: true
+            }
+          }
+        ]
       }
     },
     {
--- a/src/.eslintrc.js
+++ b/src/.eslintrc.js
@@ -1,10 +1,8 @@
 module.exports = {
   overrides: [
     {
-      files: [
-        '!**/*.test.{cjs,js,mjs}',
-        '**/*.{cjs,js,mjs}'
-      ],
+      files: ['**/*.{cjs,js,mjs}'],
+      excludedFiles: ['**/*.test.{cjs,js,mjs}'],
       env: {
         browser: true
       },
@@ -13,9 +11,15 @@ module.exports = {
       }
     },
     {
+      // Matches 'JavaScript component tests' in jest.config.mjs
+      // to ignore unknown 'page' and 'browser' Puppeteer globals
       files: [
-        '!**/template.test.js',
-        '**/*.test.js'
+        '**/all.test.{js,mjs}',
+        '**/components/*/*.test.{js,mjs}'
+      ],
+      excludedFiles: [
+        '**/(*.)?template.test.{js,mjs}',
+        '**/*.unit.test.{js,mjs}'
       ],
       globals: {
         page: true,
--- a/src/govuk/all.test.mjs
+++ b/src/govuk/all.test.mjs
@@ -47,12 +47,12 @@ describe('GOV.UK Frontend', () => {
     it('exported Components have an init function', async () => {
       await goTo(page, '/')
 
-      var componentsWithoutInitFunctions = await page.evaluate(() => {
+      const componentsWithoutInitFunctions = await page.evaluate(() => {
         const components = Object.keys(window.GOVUKFrontend)
           .filter(method => !['initAll'].includes(method))
 
         return components.filter(component => {
-          var prototype = window.GOVUKFrontend[component].prototype
+          const prototype = window.GOVUKFrontend[component].prototype
           return typeof prototype.init !== 'function'
         })
       })
--- a/src/govuk/i18n.mjs
+++ b/src/govuk/i18n.mjs
@@ -293,6 +293,7 @@ I18n.pluralRulesMap = {
  * @type {Object<string, function(number): PluralRule>}
  */
 I18n.pluralRules = {
+  /* eslint-disable jsdoc/require-jsdoc */
   arabic: function (n) {
     if (n === 0) { return 'zero' }
     if (n === 1) { return 'one' }
@@ -346,6 +347,7 @@ I18n.pluralRules = {
     if (n === 6) { return 'many' }
     return 'other'
   }
+  /* eslint-enable jsdoc/require-jsdoc */
 }
 
 /**
--- a/src/govuk/components/accordion/accordion.test.js
+++ b/src/govuk/components/accordion/accordion.test.js
@@ -17,8 +17,8 @@ describe('/components/accordion', () => {
 
         const numberOfExampleSections = 2
 
-        for (var i = 0; i < numberOfExampleSections; i++) {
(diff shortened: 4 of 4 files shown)


gulpfile.mjs
```diff
@@ -14,8 +14,10 @@ import { npmScriptTask } from './tasks/run.mjs'
  * Runs JavaScript code quality checks, documentation, compilation
  */
 gulp.task('scripts', gulp.series(
-  npmScriptTask('lint:js'),
-  compileJavaScripts,
+  gulp.parallel(
+    npmScriptTask('lint:js'),
+    compileJavaScripts
+  ),
```
romaricpascal, line 20: **question** Won't that mangle the output if anything logs? Feels like a neat little speed up though 🙌🏻

src/govuk/all.test.mjs
```diff
@@ -47,12 +47,12 @@ describe('GOV.UK Frontend', () => {
     it('exported Components have an init function', async () => {
       await goTo(page, '/')
 
-      var componentsWithoutInitFunctions = await page.evaluate(() => {
+      const componentsWithoutInitFunctions = await page.evaluate(() => {
```
romaricpascal, line 50: **question**: That doesn't feel JSDoc related, nor the change in `accordion.test.js`. Is it due to changes in the linting configuration?

src/govuk/i18n.mjs
```diff
@@ -294,6 +294,7 @@ I18n.pluralRulesMap = {
  * @type {Object<string, function(number): PluralRule>}
  */
 I18n.pluralRules = {
+  /* eslint-disable jsdoc/require-jsdoc */
```
romaricpascal, line 297: Great call there! 🙌🏻

romaricpascal (commented):
Makes all sense to me 🙌🏻  but I'm quite partial to the JSDoc work already. Might be worth getting a quick check with the team before we enforce adding JSDoc 😄

.eslintrc.js
```diff
@@ -41,8 +41,8 @@ module.exports = {
 
         // JSDoc blocks are optional
         'jsdoc/require-jsdoc': 'off',
-        'jsdoc/require-param-description': 'off',
-        'jsdoc/require-param': 'off',
+        'jsdoc/require-param-description': 'warn',
```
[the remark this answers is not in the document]
colinrotherham: Yeah don't mind at all, keeps our options open

(Reason for putting the warnings back on is that all our code passed the checks)

gulpfile.mjs
colinrotherham: Output is nice actually

For "linting only" messages everything is logged as it was but compilation continues
For invalid code you'll see the linting message AND a short Rollup or Sass error

It's mainly to remove an annoyance:

1. Open a component `*.mjs` or `*.scss` file
2. Update or add some new code
3. Save the changes… ⏳ 
4. **Linting error on invalid JSDoc or SassDoc etc** 🚨

It felt wrong to block developers from compiling valid code unless it 100% matched the style rules

colinrotherham: Like commenting out some code to do a quick browser refresh but being blocked on

>**'normaliseDataset' is defined but never used.** eslint([no-unused-vars](https://eslint.org/docs/rules/no-unused-vars))

colinrotherham: Now you see:

```console
[11:59:20] Starting 'compile:js'...
[11:59:20] Finished 'compile:js' after 86 ms

/path/to/project/govuk-frontend/src/govuk/components/accordion/accordion.mjs
  2:10  error  'normaliseDataset' is defined but never used  no-unused-vars

✖ 1 problem (1 error, 0 warnings)
```

☝️ `Finished 'compile:js'` 🎉

src/govuk/all.test.mjs
colinrotherham: Yeah it's a side effect of https://github.com/alphagov/govuk-frontend/pull/3103/commits/bebed90b207efb5843374fb49643da353ec87495

The new JSDoc rules weren't being applied to test files because we'd included/excluded them incorrectly

(Side effect being that `'no-var': 'off'` now correctly applies to source but not tests)

src/govuk/i18n.mjs
colinrotherham: Was a bit much to add JSDoc to all of those 😅

gulpfile.mjs
romaricpascal: That's brilliant then! Super neat! 🙌🏻

src/govuk/i18n.mjs
romaricpascal: At that point who's counting 🤣

.eslintrc.js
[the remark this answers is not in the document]
romaricpascal: Let's keep it off for now then to smoothen the transition and focus on useful descriptions. If we see we're missing out on descriptions, we can enable it later 😄 

```suggestion
        'jsdoc/require-param-description': 'off',
```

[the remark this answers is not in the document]
colinrotherham: All done 👍

src/.eslintrc.js
```diff
@@ -1,21 +1,39 @@
...
+          'error', {
+            require: {
+              ClassDeclaration: true,
+              ClassExpression: true,
+              FunctionExpression: true,
+              MethodDefinition: true
+            }
+          }
+        ],
+        'jsdoc/require-param-description': 'warn',
```
romaricpascal, line 23: Looks like the 'warn' value snuck back in or maybe you yere you looking to enable [`require-param-type`](https://github.com/gajus/eslint-plugin-jsdoc/blob/master/.README/rules/require-param-type.md) instead, as that's what we need for typing?

colinrotherham: @romaricpascal Oh I probably misunderstood, it's been moved from `.eslintrc.js` to `src/.eslintrc.js`

As in JSDoc params (and descriptions) are only needed for our user-facing source code for documentation

Think we should have it optional everywhere?

Thanks

colinrotherham: I've gone ahead and removed the warnings for `@param` + descriptions

It's something we can monitor during code reviews. Don't want to hassle ongoing spikes with linting issues

(Also set `enableFixer: false` so ESLint `--fix` won't keep adding placeholder `@param` either)

gulpfile.mjs
colinrotherham: Moved to:

* https://github.com/alphagov/govuk-frontend/pull/3127

domoscargin (commented):
Given that the code already meets these settings, I think it makes sense to formalise the jsdoc requirements.

domoscargin (approved):

colinrotherham:
> Given that the code already meets these settings, I think it makes sense to formalise the jsdoc requirements.

Thanks @domoscargin 👍 

@romaricpascal I know you were keen on adding warnings when we've missed type declarations (plus stricter checks for missing `@param` + descriptions) but only if the team agrees that's what we want

(Need to weigh up the benefits of valid types for publishing versus quick spikes where things change quickly)

So I'll go ahead and merge this one as-is and we can tighten further if wanted in future

PR #3104, opened 2022-12-15
Add (optional) type declaration checks

colinrotherham:
This PR adds a `tsconfig.json` so the TypeScript compiler can check our code and closes:

* https://github.com/alphagov/govuk-frontend/issues/3230

What does this mean for us?

1. Checks we're calling library code as intended
2. Checks we're calling DOM methods as intended
3. Checks we're adhering to our own JSDoc comments

### GitHub Actions
Checks are optional and can be run manually

GitHub Actions UI for check type declarations

For example, missing or invalid parameters will be spotted by the compiler:

```console
  src/govuk/components/character-count/character-count.mjs:59:35
    59 function CharacterCount ($module, config) {
                                         ~~~~~~
    An argument for 'config' was not provided.
```

### Related changes

This PR completes work we started for v4.4.0:

* https://github.com/alphagov/govuk-frontend/pull/2997

To assist with code reviews I've split these changes into:

5. https://github.com/alphagov/govuk-frontend/pull/3123
6. https://github.com/alphagov/govuk-frontend/pull/3104

## Why TypeScript?

We're not actually writing any TypeScript code but this PR includes it to:

1. Check type information via ESLint using ~[`standard`](https://www.npmjs.com/package/eslint-config-standard)~ [`standard-with-typescript`](https://www.npmjs.com/package/eslint-config-standard-with-typescript)
2. Run the TypeScript compiler `tsc` to check if our code would build without errors

Similar to [source maps](https://github.com/alphagov/govuk-frontend/pull/3023), we may want `tsc` to output type declarations in future:

* https://github.com/alphagov/govuk-frontend/issues/2835

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -104,5 +104,8 @@ module.exports = {
       }
     }
   ],
+  parserOptions: {
+    project: './tsconfig.json'
+  },
   root: true
 }
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -19,6 +19,9 @@ on:
           - macos-latest
           - ubuntu-latest
           - windows-latest
+      types:
+        description: Check type declarations
+        type: boolean
 
 concurrency:
   group: tests-${{ github.head_ref || github.run_id }}
@@ -50,6 +53,10 @@ jobs:
 
       - name: Build
         uses: ./.github/workflows/actions/build
+
+      - name: Type declarations (optional)
+        if: github.event.inputs.types == 'true'
+        run: npm run build:types
 
   lint:
     name: ${{ matrix.task.description }}
--- a/app/.eslintrc.js
+++ b/app/.eslintrc.js
@@ -1,0 +1,18 @@
+const { join } = require('path')
+
+module.exports = {
+  overrides: [
+    {
+      files: ['**/*.js'],
+      excludedFiles: [
+        '**/.eslintrc.js',
+        '**/*.test.js'
+      ],
+      parser: '@typescript-eslint/parser',
+      parserOptions: {
+        project: [join(__dirname, 'tsconfig.json')]
+      },
+      plugins: ['@typescript-eslint']
+    }
+  ]
+}
--- a/app/app.js
+++ b/app/app.js
@@ -149,6 +149,8 @@ module.exports = async (options) => {
     )
 
     let bodyClasses = ''
+
+    // @ts-expect-error `URLSearchParams` as 'query parser'
     if (req.query.has('iframe')) {
       bodyClasses = 'app-iframe-in-component-preview'
     }
--- a/app/tsconfig.json
+++ b/app/tsconfig.json
@@ -1,0 +1,8 @@
+{
+  "extends": "../tsconfig.base.json",
+  "include": ["**/*.js", "**/*.json"],
+  "exclude": ["**/*.test.*"],
+  "compilerOptions": {
+    "resolveJsonModule": true
+  }
+}
--- a/docs/examples/webpack/package.json
+++ b/docs/examples/webpack/package.json
@@ -13,7 +13,6 @@
     "@babel/core": "^7.21.0",
     "@babel/preset-env": "^7.20.2",
     "babel-loader": "^9.1.2",
-    "clean-webpack-plugin": "^4.0.0",
     "copy-webpack-plugin": "^11.0.0",
     "html5shiv": "^3.7.3",
     "postcss-loader": "^7.0.2",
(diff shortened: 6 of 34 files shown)


.eslintrc.js
```diff
@@ -1,5 +1,5 @@
 module.exports = {
-  extends: 'standard',
+  extends: 'standard-with-typescript',
```
romaricpascal, line 2: **question** I'm not sure that does what we hope it does. Looking at [the config it brings], it looks like it only adds TypeScript parsing for `.ts` and `.tsx` files. 

This won't match anything on our project, and will not add type information for the linter while parsing our JS files. 

Instead, don't we want to keep `standard` but add the `@typescript-eslint` plugin and set `@typescript-eslint/parser` to parse the files that will get bundled? 

I gave that a quick shot and ended with the following errors when linting ran:

```txt
However, that TSConfig does not include this file. Either:
- Change ESLint's list of included files to not include this file
- Change that TSConfig to include this file
- Create a new TSConfig that includes this file and include it in your parserOptions.project
See the typescript-eslint docs for more info: https://typescript-eslint.io/linting/troubleshooting##i-get-errors-telling-me-eslint-was-configured-to-run--however-that-tsconfig-does-not--none-of-those-tsconfigs-include-this-file
```

colinrotherham: Ah sorry @romaricpascal you're right we're not using `@typescript-eslint/parser`

But we are using the TypeScript ESLint plugin that comes with `standard-with-typescript`:

```js
plugins: ['@typescript-eslint']
```

We may want to opt in to the parser in future but likely still needs:

The ESLint TypeScript parser likes to have clear boundaries between child projects
https://typescript-eslint.io/linting/typed-linting/monorepos/

Thanks for taking a close look

colinrotherham: >Instead, don't we want to keep standard but add the @typescript-eslint plugin and set @typescript-eslint/parser to parse the files that will get bundled? Or [maybe follow @typescript-eslint's recommended setup](https://typescript-eslint.io/getting-started/#step-2-configuration)

@romaricpascal Pushed again, now following the recommended setup 👍

Also useful for prototype checks in your PR too:

* https://github.com/alphagov/govuk-frontend/pull/3153

.github/workflows/tests.yml
```diff
@@ -42,6 +46,10 @@ jobs:
       - name: Build
         uses: ./.github/workflows/actions/build
 
+      - name: Type declarations (optional)
+        if: github.event.inputs.types == 'true'
+        run: npm run build:types
```
romaricpascal, line 51: **question**: Because the step is in the `Build` job, that'll mark that the build failed, which sounds pretty alarming for an optional check. It will muddy things between bundling having failed or type checks having failed until you dig in the job logs. Does it necessarily have to be grouped with the `Build` (for performance, maybe? or was it just in anticipation that we'd use this to build a `.d.ts` file at some point in the future)? Could it be an entry in the linting matrix instead to highlight that it's specifically the types that failed in the list of checks?

colinrotherham: Yeah spot on! I did this on purpose

Do I need to rename the word "optional"?

**Don't tick it:** Types can't break the build
**Do tick it:** Types can break the build

In terms of grouping it with the **Build** step, it's because the TypeScript CLI command is `tsc --build` and would (eventually) output our type definitions. It's always skipped until you tick the box.

### Lint task matrix
Sadly it's not possible for a job-level `if:` to conditionally check values from "inputs" against matrix "include" context without using [`${{ fromJSON() }` to build the matrix dynamically](https://stackoverflow.com/questions/65384420/how-do-i-make-a-github-action-matrix-element-conditional) 😱 

The alternative is to drop to step-level `if:` but only EVERY matrix step 😬

```yaml
if: ${{ matrix.name !== 'lint-types' || github.event.inputs.types == 'true' }}
```

Which gets pretty repetitive:
```yaml
steps:
  - name: Checkout
    if: ${{ matrix.name !== 'lint-types' || github.event.inputs.types == 'true' }}
    uses: actions/checkout@v3

  - name: Restore dependencies
    if: ${{ matrix.name !== 'lint-types' || github.event.inputs.types == 'true' }}

  - name: Cache linter
    if: ${{ (matrix.name !== 'lint-types' || github.event.inputs.types == 'true') && matrix.cache }}
    uses: actions/cache@v3
    with:
      key: ${{ matrix.name }}-cache-${{ runner.os }}
      path: ${{ matrix.cache }}

  - name: Run lint task
    if: ${{ matrix.name !== 'lint-types' || github.event.inputs.types == 'true' }}
    run: ${{ matrix.run }}
```

romaricpascal (approved):
All ready to go! Thanks so much for your patience with this one for all the reviewing, amends and rebasing to keep it up to date 😄

PR #3105, opened 2022-12-15
Configure Heroku review app dyno size to “eco”

colinrotherham:
Ensures we're using the 1000 hours [Eco Dyno plan](https://devcenter.heroku.com/articles/dyno-types) for pull request "Review apps"

>⚠️ **Latest build failed**
>To create a review app, the pipeline owner must first authorize additional charges.
>Learn more at http://hrku.co/yzF95b.

### DIFF
--- a/app.json
+++ b/app.json
@@ -1,7 +1,5 @@
 {
   "name": "govuk-frontend",
-  "scripts": {
-  },
   "env": {
     "HEROKU_APP": "REVIEW"
   },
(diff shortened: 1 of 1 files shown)


owenatgov (dismissed):

colinrotherham:
Looks like this was a temporary blip with Heroku counting some of our PR apps as billable

Might not be a bad idea to merge anyway but what's in **app.json** is likely the same as the defaults

lfdebrux (approved):

colinrotherham:
@owenatgov @lfdebrux Added a comment to the commit

>These may represent the defaults but we had a billing blip today where some busy PRs were being charged as “basic” (not “eco”) size dynos
>
>Formation settings taken from:
>https://stackoverflow.com/a/72889265

PR #3110, opened 2022-12-16
Rename `target` to `$target` and other tiny code style changes

colinrotherham:
Separating out a few gotchas from from this PR:

They weren't type related but tidy up a few inconsistencies across components

`target` → **`$target`** – Sometimes "target" is an HTML element
`target` → **`targetId`** – Sometimes "target" is an ID attribute value

Additionally I've renamed Checkboxes `$target` to **`$clickedInput`** to match Radios

(The two components share a lot of code which are now back into sync)

Also related to https://github.com/alphagov/govuk-frontend/issues/2513

### DIFF
--- a/src/govuk/components/details/details.mjs
+++ b/src/govuk/components/details/details.mjs
@@ -85,7 +85,7 @@ Details.prototype.polyfillDetails = function () {
   }
 
   // Bind an event to handle summary elements
-  this.polyfillHandleInputs($summary, this.polyfillSetAttributes.bind(this))
+  this.polyfillHandleInputs(this.polyfillSetAttributes.bind(this))
 }
 
 /**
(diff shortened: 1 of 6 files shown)


src/govuk/components/details/details.mjs
```diff
@@ -110,21 +110,20 @@ Details.prototype.polyfillSetAttributes = function () {
...
  * Handle cross-modal click events
  *
- * @param {object} node - element
  * @param {polyfillHandleInputsCallback} callback - function
  */
-Details.prototype.polyfillHandleInputs = function (node, callback) {
-  node.addEventListener('keypress', function (event) {
-    var target = event.target
+Details.prototype.polyfillHandleInputs = function (callback) {
+  this.$summary.addEventListener('keypress', function (event) {
```
colinrotherham, line 116: I was a bit caught out by `node` as we already have `this.$summary`

PR #3112, opened 2022-12-16
Remove unused `classList` polyfill from header component JavaScript

36degrees:
The header imports the `Element.prototype.classList` polyfill but hasn’t actually called `classList` anywhere since we updated it to use the `hidden` attribute to toggle the menu visibility in 7217a7f (#2727).


owenatgov (approved):
Good spot!

PR #3113, opened 2022-12-16
Add version info from `package.json` to config

colinrotherham:
Quick PR to split up our `configPaths` into separate parts:

```mjs
import { paths, pkg, ports } from './config/index.js'
```

(Seemed a bit odd that "ports" was inside `configPaths.ports`)


36degrees (approved):
Neat 👍🏻

PR #3114, opened 2022-12-16
Ensure `VERSION.txt` task accepts `--destination` flag

colinrotherham:
Noticed our **build:dist** task accepts `--destination` but errors when trying to target another directory

It can now be varied like the others:

```console
npx gulp compile --destination '123'
npx gulp build:dist --destination '456'
npx gulp build:package --destination '789'
```

I've added a [check for `VERSION.txt`](https://github.com/alphagov/govuk-frontend/pull/3114/files#diff-24cf4c47d61ad26051ad863d660f1db5b9760eac9cd0e6f4d6e3a6a494b51ab5) in `./dist` for `npm run postbuild:dist`

### DIFF
--- a/tasks/asset-version.mjs
+++ b/tasks/asset-version.mjs
@@ -4,7 +4,7 @@ import { join } from 'path'
 
 import { pkg } from '../config/index.js'
 
-import { destination, isDist } from './task-arguments.mjs'
+import { destination } from './task-arguments.mjs'
 
 /**
  * Write VERSION.txt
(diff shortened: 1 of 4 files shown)


tasks/asset-version.mjs
```diff
@@ -4,21 +4,16 @@ import { join } from 'path'
 
 import { pkg } from '../config/index.js'
 
-import { destination, isDist } from './task-arguments.mjs'
```
domoscargin, line 7: Nice, looks like we now only use `isDist` in `compile-assets.mjs`.

domoscargin (approved):
This looks sensible to me!

PR #3115, opened 2022-12-16
[DO NOT MERGE] Provide a standalone way to run rollup for stats gathering

romaricpascal:
Just a late idea for running a more recent version of rollup for gathering stats, waiting for us to move to rollup 3, taking a different route than #3109:

Instead of running during our build, it bundles a one off entry, dynamically generated, that imports the content of our package. `rollup-plugin-visualizer` still serves as a stat collection tool and we'd need to look in the JSON stats it creates to highlight what we want.

We could have multiple entries pointing to each of the `govuk-esm` files to ensure that when imported they only load the modules we expect them to. Especially when importing parts of the exports (eg. `Accordion` from `all.mjs`), we could check that treeshaking works as intended.

This could be:
- a route to look into collecting stats with `rollup-plugin-visualizer` before we set it up on our build. Allowing us to put all the effort on it rather than a bit with size-limit now, and digging deeper once we've moved to rollup 3.
- a route to check treeshaking output (though it would only be valid for `rollup`'s way of tree-shaking)


romaricpascal:
Spike work is complete and perf work carries on in #3279

PR #3122, opened 2022-12-19
Accordion `$sections` should use `this.sectionClass` etc

colinrotherham:
Separating out a few more gotchas from from this PR:

Similar to https://github.com/alphagov/govuk-frontend/pull/3110 they weren't type related but fix various Accordion inconsistencies

`$header` → **`$headingText`** – We use "headingText" as the heading's inner span
`$headerWrapper` → **`$header`** – We use "header" as the wrapper for heading/button/icon
`$showToggle` → **`$showHideToggle`** – Uses the same name throughout the component
`$showToggleText` → **`$showHideText`** – Uses the same name throughout the component
`$icon` → **`$showHideIcon`** – Uses the same name throughout the component

I've picked the most frequent name but let me know if others are preferred.

Additionally we were re-selecting “Show all sections” icon `querySelector()` calls on every expanded check

Also related to https://github.com/alphagov/govuk-frontend/issues/2513

### DIFF
--- a/src/govuk/components/accordion/accordion.mjs
+++ b/src/govuk/components/accordion/accordion.mjs
@@ -61,6 +61,7 @@ function Accordion ($module, config) {
   this.sectionButtonClass = 'govuk-accordion__section-button'
   this.sectionHeaderClass = 'govuk-accordion__section-header'
   this.sectionHeadingClass = 'govuk-accordion__section-heading'
+  this.sectionHeadingDividerClass = 'govuk-accordion__section-heading-divider'
   this.sectionHeadingTextClass = 'govuk-accordion__section-heading-text'
   this.sectionHeadingTextFocusClass = 'govuk-accordion__section-heading-text-focus'
 
@@ -472,7 +473,7 @@ Accordion.prototype.setInitialState = function ($section) {
  */
 Accordion.prototype.getButtonPunctuationEl = function () {
   var $punctuationEl = document.createElement('span')
-  $punctuationEl.classList.add('govuk-visually-hidden', 'govuk-accordion__section-heading-divider')
+  $punctuationEl.classList.add('govuk-visually-hidden', this.sectionHeadingDividerClass)
   $punctuationEl.innerHTML = ', '
   return $punctuationEl
 }
--- a/src/govuk/components/accordion/accordion.mjs
+++ b/src/govuk/components/accordion/accordion.mjs
@@ -192,23 +192,23 @@ Accordion.prototype.constructHeaderMarkup = function ($header, index) {
   $headingTextFocus.innerHTML = $span.innerHTML
 
   // Create container for show / hide icons and text.
-  var $showToggle = document.createElement('span')
-  $showToggle.classList.add(this.sectionShowHideToggleClass)
+  var $showHideToggle = document.createElement('span')
+  $showHideToggle.classList.add(this.sectionShowHideToggleClass)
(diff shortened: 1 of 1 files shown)


src/govuk/components/accordion/accordion.mjs
```diff
@@ -39,17 +39,17 @@ var ACCORDION_TRANSLATIONS = {
  */
 function Accordion ($module, config) {
   this.$module = $module
-  this.$sections = $module.querySelectorAll('.govuk-accordion__section')
```
colinrotherham, line 42: Moved this line down so we can use `'.' + this.sectionClass` instead

src/govuk/components/accordion/accordion.mjs
```diff
@@ -61,6 +61,7 @@ function Accordion ($module, config) {
   this.sectionButtonClass = 'govuk-accordion__section-button'
   this.sectionHeaderClass = 'govuk-accordion__section-header'
   this.sectionHeadingClass = 'govuk-accordion__section-heading'
+  this.sectionHeadingDividerClass = 'govuk-accordion__section-heading-divider'
```
colinrotherham, line 64: Added this one so we don't need to hard code it

src/govuk/components/accordion/accordion.mjs
```diff
@@ -103,9 +107,9 @@ Accordion.prototype.initControls = function () {
   this.$showAllButton.setAttribute('aria-expanded', 'false')
 
   // Create icon, add to element
-  var $icon = document.createElement('span')
-  $icon.classList.add(this.upChevronIconClass)
-  this.$showAllButton.appendChild($icon)
+  this.$showAllIcon = document.createElement('span')
```
colinrotherham, line 110: By caching `this.$showAllIcon` here we don't need to query it again later

src/govuk/components/accordion/accordion.mjs
```diff
@@ -114,9 +118,9 @@ Accordion.prototype.initControls = function () {
   this.$module.insertBefore($accordionControls, this.$module.firstChild)
 
   // Build additional wrapper for Show all toggle text and place after icon
-  var $wrappershowAllText = document.createElement('span')
-  $wrappershowAllText.classList.add(this.showAllTextClass)
-  this.$showAllButton.appendChild($wrappershowAllText)
+  this.$showAllText = document.createElement('span')
```
colinrotherham, line 121: This one too as `this.$showAllText` saves another query later

src/govuk/components/accordion/accordion.mjs
```diff
@@ -150,13 +154,13 @@ Accordion.prototype.initSectionHeaders = function () {
 /**
  * Construct section header
  *
- * @param {HTMLDivElement} $headerWrapper - Section header wrapper
```
colinrotherham, line 153: We call this `$header` everywhere else

src/govuk/components/accordion/accordion.mjs
```diff
@@ -311,8 +315,12 @@ Accordion.prototype.setExpanded = function (expanded, $section) {
   $button.setAttribute('aria-expanded', expanded)
 
   // Update aria-label combining
-  var $header = $section.querySelector('.' + this.sectionHeadingTextClass)
```
colinrotherham, line 314: Changed this as it wasn't `$header` but actually `$headingText`

36degrees (approved):
Lots of tidy improvements – thanks for doing this 🙏🏻

PR #3123, opened 2022-12-19
Enable type declaration guards

colinrotherham:
Add `instanceof` checks where necessary to unblock type declaration checks in:

### DIFF
--- a/src/govuk/components/character-count/character-count.mjs
+++ b/src/govuk/components/character-count/character-count.mjs
@@ -88,6 +88,7 @@ function CharacterCount ($module, config) {
     }
   }
 
+  /** @type {CharacterCountConfig} */
   this.config = mergeConfigs(
     defaultConfig,
     config || {},
--- a/src/govuk/components/character-count/character-count.mjs
+++ b/src/govuk/components/character-count/character-count.mjs
@@ -57,12 +57,12 @@ var CHARACTER_COUNT_TRANSLATIONS = {
  * @param {CharacterCountConfig} [config] - Character count config
  */
 function CharacterCount ($module, config) {
-  if (!$module) {
+  if (!($module instanceof HTMLElement)) {
     return this
   }
 
   var $textarea = $module.querySelector('.govuk-js-character-count')
-  if (!$textarea) {
+  if (!($textarea instanceof HTMLTextAreaElement || $textarea instanceof HTMLInputElement)) {
     return this
   }
 
--- a/docs/contributing/coding-standards/js.md
+++ b/docs/contributing/coding-standards/js.md
@@ -22,7 +22,7 @@ import '../../vendor/polyfills/Element.mjs'
  * @param {Element} $module - HTML element to use for component
  */
 function Example ($module) {
-  if (!$module) {
+  if (!($module instanceof HTMLElement)) {
     return this
   }
 
--- a/src/govuk/common/index.mjs
+++ b/src/govuk/common/index.mjs
@@ -67,6 +67,7 @@ export function mergeConfigs (/* configObject1, configObject2, ...configObjects 
    */
   var flattenObject = function (configObject) {
     // Prepare an empty return object
+    /** @type {Object<string, unknown>} */
     var flattenedObject = {}
 
     /**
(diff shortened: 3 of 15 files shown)


src/govuk/components/character-count/character-count.mjs
```diff
@@ -57,12 +57,12 @@ var CHARACTER_COUNT_TRANSLATIONS = {
...
  */
 function CharacterCount ($module, config) {
-  if (!$module) {
+  if (!($module instanceof HTMLElement)) {
     return this
   }
 
   var $textarea = $module.querySelector('.govuk-js-character-count')
-  if (!$textarea) {
+  if (!($textarea instanceof HTMLTextAreaElement)) {
```
colinrotherham, line 65: Do we want to allow text inputs?

romaricpascal: I'd say yes, just as a safety to not break things if someone mounted the component on a markup with an `<input>` instead of `<textarea>`. I don't think we officially support it (given our template uses a textarea), but it does work.

I'm saying yes with the assumption that it's a matter of turning that `if` into `if (!($textarea instanceof HTMLTextAreaElement || $textarea instanceof HTMLInputElement))`, though. If it's more complex than that, I think we can stick to `<textarea>` and treat it in its own block of work for properly supporting `input` in the CharacterCount (both at JS and template level).

https://user-images.githubusercontent.com/396367/217499669-30cd1805-98dd-4fa8-8242-e7b7873eeecb.mp4

colinrotherham: Yeah this is where it gets tricky

File upload, checkbox and radio inputs are also `HTMLInputElement`

But in terms of guarding the `.value` property your example works great 👍

colinrotherham: These are the following places we've gone further than `HTMLElement`:

**Character count** needs `<textarea>` or `<input>`
[`lint-types` ./src/govuk/components/character-count/character-count.mjs#L65](https://github.com/alphagov/govuk-frontend/blob/lint-types/src/govuk/components/character-count/character-count.mjs#L65)

**Checkboxes** need `<input>` with `type="checkbox"`
[`lint-types` ./src/govuk/components/checkboxes/checkboxes.mjs#L179](https://github.com/alphagov/govuk-frontend/blob/lint-types/src/govuk/components/checkboxes/checkboxes.mjs#L179)

**Radios** need `<input>` with `type="radio"`
[`lint-types` ./src/govuk/components/radios/radios.mjs#L129](https://github.com/alphagov/govuk-frontend/blob/lint-types/src/govuk/components/radios/radios.mjs#L129)

**Details** needs `<details>` to avoid polyfill
[`lint-types` ./src/govuk/components/details/details.mjs#L40](https://github.com/alphagov/govuk-frontend/blob/lint-types/src/govuk/components/details/details.mjs#L40)

**Error summary** needs `<a>` for links and `<input>` (checkbox/radio) for legend `.scrollIntoView()`
[`lint-types` ./src/govuk/components/error-summary/error-summary.mjs#L114](https://github.com/alphagov/govuk-frontend/blob/lint-types/src/govuk/components/error-summary/error-summary.mjs#L114)
[`lint-types` ./src/govuk/components/error-summary/error-summary.mjs#L185](https://github.com/alphagov/govuk-frontend/blob/lint-types/src/govuk/components/error-summary/error-summary.mjs#L185)

**Skip link** needs `<a>`
[`lint-types` ./src/govuk/components/skip-link/skip-link.mjs#L14](https://github.com/alphagov/govuk-frontend/blob/lint-types/src/govuk/components/skip-link/skip-link.mjs#L14)

**Tabs** need `<a>`
[`lint-types` ./src/govuk/components/tabs/tabs.mjs#L270](https://github.com/alphagov/govuk-frontend/blob/lint-types/src/govuk/components/tabs/tabs.mjs#L270)

romaricpascal (approved):
I'd say we're good to go on that one, as I don't think any of these changes would be breaking. @36degrees I would appreciate a second look, though, in case we missed a potential breaking case when going through the changes with Colin.

36degrees (approved):

PR #3132, opened 2022-12-21
Configure Dependabot for direct updates only

colinrotherham:
This PR configures Dependabot to only allow updates to [`direct` dependencies](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#allow)

We regularly receive Dependabot PRs for “dependencies of dependencies” **package-lock.json** changes. It's great for [**Dependabot security updates** via Settings](https://github.com/alphagov/govuk-frontend/settings/security_analysis) it's but too noisy for [**Dependabot version updates** via `dependabot.yml`](https://github.com/alphagov/govuk-frontend/blob/main/.github/dependabot.yml)

For example updates to  [`indirect` dependencies](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#allow) don't appear in `npm outdated`:

1. `@babel/helper-*`
2. `@babel/plugin-*`
3. `@babel/types`
4. `core-js-compat`
5. `semver`

Yet they will still be included in a single PR update for [`@babel/preset-env`](https://www.npmjs.com/package/@babel/preset-env) by the author

Should we narrow down our non-security version updates to direct-only?

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,7 +12,7 @@ updates:
     versioning-strategy: increase
 
     allow:
-      - dependency-type: all
+      - dependency-type: direct
 
     ignore:
       # Ignore major updates (breaks IE8 PostCSS plugins)
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "govuk-frontend-repository",
-  "lockfileVersion": 3,
+  "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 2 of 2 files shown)


colinrotherham:
I've also had to downgrade **package-lock.json** to `"lockfileVersion": 2` until this bug is fixed:

* https://github.com/dependabot/dependabot-core/issues/6270

domoscargin (approved):
Pity about the lockfile, hopefully the bug gets sorted soon.

Other than that, though, I think it make a bundle o' sense to cut down the noise on these.

colinrotherham:
> Pity about the lockfile, hopefully the bug gets sorted soon.

Thanks @domoscargin we'll keep an eye on this

Shame to add those 19k+ lines back in!

> Other than that, though, I think it make a bundle o' sense to cut down the noise on these.

Yep. Should things go too quiet we may want to try another variation:

```yml
allow:
  - dependency-type: production
  - dependency-type: development
```

We'll monitor things and see how they go

Issue #3136, opened 2022-12-23
Minified JS in `dist` errors in IE8


36degrees:
## Description of the issue

The minified JavaScript we provide in the `dist` directory (which is also included in the zip file attached to each release) errors in IE8 – seemingly because of transformations made by Uglify JS.

Although we have plans to drop support for IE8-10 shortly, our [current browser support](https://github.com/alphagov/govuk-frontend#browser-and-assistive-technology-support) includes Internet Explorer 8, 9 and 10, 'although components may not look perfect'.

## Steps to reproduce the issue

Load the review app in IE8 (after #3050 has been merged)

## Actual vs expected behaviour

The page should load without errors. Instead we get:

> Object does not support this property or method

## Environment (where applicable)

- Operating system: Windows 7
- Browser: Internet Explorer
- Browser version: 8.0
- GOV.UK Frontend Version: N/A

36degrees:
Looks like this might be fixed by #3013.

colinrotherham:
We now have two fixes to pick:

1. https://github.com/alphagov/govuk-frontend/pull/3137
2. https://github.com/alphagov/govuk-frontend/pull/3013

But equally we could merge 1) and take on 2) at a later date

colinrotherham:
I've checked out previous releases, enabled UglifyJS, and tested in IE8 via Browserstack

### [GOV.UK Frontend releases](https://github.com/alphagov/govuk-frontend/releases)

* [**v4.3.1**](https://github.com/alphagov/govuk-frontend/releases/tag/v4.3.1) – using [`uglify-js@3.6.9`](https://github.com/mishoo/UglifyJS/releases/tag/v3.6.9) (works in IE8)
* [**v4.4.0**](https://github.com/alphagov/govuk-frontend/releases/tag/v4.4.0) – using [`uglify-js@3.17.4`](https://github.com/mishoo/UglifyJS/releases/tag/v3.17.4) (broken in IE8)
* [**v4.4.1**](https://github.com/alphagov/govuk-frontend/releases/tag/v4.4.1) – using [`uglify-js@3.17.4`](https://github.com/mishoo/UglifyJS/releases/tag/v3.17.4) (broken in IE8)

See full diff in [compare view](https://github.com/mishoo/UglifyJS/compare/v3.6.9...v3.17.4)

colinrotherham:
Closed by https://github.com/alphagov/govuk-frontend/pull/3137

PR #3013, opened 2022-11-15
Swap JavaScript minifier from UglifyJS to terser

colinrotherham:
Enable minification of ES6+ features

Swaps [UglifyJS](https://github.com/mishoo/UglifyJS#readme) with [terser](https://terser.org) (forked version) as explained on the [terser **README.md**](https://github.com/terser/terser#readme)

>`uglify-es` is [no longer maintained](https://github.com/mishoo/UglifyJS2/issues/3156#issuecomment-392943058) and `uglify-js` does not support ES6+.
>
>**`terser`** is a fork of `uglify-es` that mostly retains API and CLI compatibility with `uglify-es` and `uglify-js@3`.

Related:

* https://github.com/alphagov/govuk-frontend/pull/2870
* https://github.com/alphagov/govuk-frontend/issues/2872
* https://github.com/alphagov/govuk-frontend/pull/3050

PR #3137, opened 2022-12-23
Enable JavaScript minifier compatibility options

colinrotherham:
Investigation into https://github.com/alphagov/govuk-frontend/issues/3136


colinrotherham:
Whilst not related, I've also added a missing [**Event** polyfill](https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/vendor/polyfills/Event.mjs) to **Accordion**

36degrees (approved):

colinrotherham:
@36degrees I'm going to merge this in as initial testing looks good

Won't focus on it as we're nearly set to replace it with Terser once testing is complete:

* https://github.com/alphagov/govuk-frontend/pull/3013

PR #2868, opened 2022-09-20
Fix stray `let` in I18n preventing uglify

colinrotherham:
This PR fixes a tiny `let` → `var` which caused a fairly obscure **GulpUglifyError**

```console
[18:55:27] GulpUglifyError: unable to minify JavaScript
Caused by: SyntaxError: Unexpected token: name «placeholderValue», expected: punc «;»
File: /Users/colin/Sites/GDS/govuk-frontend/src/govuk/all.mjs
Line: 524
Col: 10
```

To catch this next time I've added the other npm `build:*` scripts to the GitHub workflow:

```yaml
- name: Build
  run: |
    npm run build:assets
    npm run build:package
    npm run build:dist
```

PR #3153, opened 2023-01-10
Set up `eslint-plugin-es-x` to check the syntax we use

romaricpascal:
[`eslint-plugin-es-x`](https://eslint-community.github.io/eslint-plugin-es-x/) validates the use of features brought by specific ES version. This should give us a safety net that we're not including the modern features it can detect in our code. 

The setup sets the restrictions to ES3 features, due to our IE8 support. The rules matching the features we polyfills are deactivated to allow us using them.

We'll still need a bit of caution as it only have rules [against EcmaScript features not Browser API features](https://eslint-community.github.io/eslint-plugin-es-x/rules/).

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -92,14 +92,6 @@ module.exports = {
       }
     },
     {
-      files: ['**/govuk/components/**/*.{cjs,js,mjs}'],
-      excludedFiles: ['**/govuk/components/**/*.test.{cjs,js,mjs}'],
-      parserOptions: {
-        // Note: Allow ES6 for import/export syntax (although our code is ES3 for legacy browsers)
-        ecmaVersion: '2015'
-      }
-    },
-    {
       files: ['**/*.test.{cjs,js,mjs}'],
       env: {
         jest: true
--- a/src/.eslintrc.js
+++ b/src/.eslintrc.js
@@ -24,6 +24,22 @@ module.exports = {
       }
     },
     {
+      files: ['govuk/**/*.mjs'],
+      excludedFiles: ['**/*.test.mjs'],
+      parserOptions: {
+        // Note: Allow ES6 for import/export syntax (although our code is ES3 for legacy browsers)
+        ecmaVersion: '2015'
+      },
+      plugins: ['es-x'],
+      extends: ['plugin:es-x/restrict-to-es3'],
+      rules: {
+        // Rollup transpiles modules to AMD export/define
+        'es-x/no-modules': 'off',
+        // Rollup plays fine with `export * from`
+        'es-x/no-export-ns-from': 'off'
+      }
+    },
+    {
       // Matches 'JavaScript component tests' in jest.config.mjs
       // to ignore unknown 'page' and 'browser' Puppeteer globals
       files: [
--- a/src/govuk/common/normalise-dataset.mjs
+++ b/src/govuk/common/normalise-dataset.mjs
@@ -1,3 +1,5 @@
+/* eslint-disable es-x/no-string-prototype-trim -- Polyfill imported */
+
 import '../vendor/polyfills/Element/prototype/dataset.mjs'
 import '../vendor/polyfills/String/prototype/trim.mjs'
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -57,6 +57,7 @@
         "cheerio": "^1.0.0-rc.12",
         "eslint": "^8.31.0",
         "eslint-config-standard": "^17.0.0",
+        "eslint-plugin-es-x": "^5.4.0",
         "eslint-plugin-import": "^2.26.0",
         "eslint-plugin-jsdoc": "^39.6.4",
         "eslint-plugin-n": "^15.6.0",
@@ -8211,6 +8212,25 @@
       },
       "funding": {
         "url": "https://github.com/sponsors/mysticatea"
+      },
+      "peerDependencies": {
+        "eslint": ">=4.19.1"
+      }
+    },
+    "node_modules/eslint-plugin-es-x": {
+      "version": "5.4.0",
+      "resolved": "https://registry.npmjs.org/eslint-plugin-es-x/-/eslint-plugin-es-x-5.4.0.tgz",
+      "integrity": "sha512-6Mniw760Nhd6brnDy+rz857LD+icZe5wXmsvXSuJ84svM0Q53ulJxpMhTJmpqHaLzYh7fuGAJ8V62ohbmqF+jA==",
+      "dev": true,
+      "dependencies": {
+        "eslint-utils": "^2.0.0 || ^3.0.0",
+        "regexpp": "^3.0.0"
+      },
+      "engines": {
+        "node": ">=8.10.0"
+      },
+      "funding": {
+        "url": "https://github.com/sponsors/ota-meshi"
       },
       "peerDependencies": {
         "eslint": ">=4.19.1"
@@ -31746,6 +31766,16 @@
           "integrity": "sha512-6J72N8UNa462wa/KFODt/PJ3IU60SDpC3QXC1Hjc1BXXpfL2C9R5+AU7jhe0F6GREqVMh4Juu+NY7xn+6dipUQ==",
           "dev": true
         }
+      }
+    },
+    "eslint-plugin-es-x": {
+      "version": "5.4.0",
+      "resolved": "https://registry.npmjs.org/eslint-plugin-es-x/-/eslint-plugin-es-x-5.4.0.tgz",
+      "integrity": "sha512-6Mniw760Nhd6brnDy+rz857LD+icZe5wXmsvXSuJ84svM0Q53ulJxpMhTJmpqHaLzYh7fuGAJ8V62ohbmqF+jA==",
+      "dev": true,
+      "requires": {
+        "eslint-utils": "^2.0.0 || ^3.0.0",
+        "regexpp": "^3.0.0"
       }
     },
     "eslint-plugin-import": {
--- a/package.json
+++ b/package.json
@@ -82,6 +82,7 @@
     "cheerio": "^1.0.0-rc.12",
     "eslint": "^8.31.0",
     "eslint-config-standard": "^17.0.0",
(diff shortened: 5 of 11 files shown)


.eslintrc.js
```diff
@@ -92,11 +89,24 @@ module.exports = {
       }
     },
     {
-      files: ['**/govuk/components/**/*.{cjs,js,mjs}'],
-      excludedFiles: ['**/govuk/components/**/*.test.{cjs,js,mjs}'],
+      files: ['src/govuk/**/*.{cjs,js,mjs}'],
```
romaricpascal, line 92: I expanded the downgrade of EcmaVersion to all files in `src/govuk`, meaning this now includes `all.mjs`, `common.mjs` and `i18n.mjs`. This should ensure all code ending up in the bundle (besides the polyfills) gets parsed as ES2015 by ESLint.

.eslintrc.js
```diff
@@ -92,11 +92,24 @@ module.exports = {
...
+        'plugin:es-x/restrict-to-es3'
+      ],
+      rules: {
+        // We use ES6 modules to structure our code
+        'es-x/no-modules': 'off',
+        // Rollup plays fine with `export * from`
+        'es-x/no-export-ns-from': 'off',
+        // These ES5+ features are polyfilled
+        'es-x/no-function-prototype-bind': 'off',
+        'es-x/no-date-now': 'off'
```
colinrotherham, line 112: @romaricpascal I'm aware we've got a much wider list of polyfills in our repo

>~Date.now()~ ✅
>Element.prototype.classList
>Element.prototype.closest()
>Element.prototype.dataset()
>Element.prototype.matches()
>Element.prototype.nextElementSibling
>Element.prototype.previousElementSibling
>~Function.prototype.bind()~ ✅
>Object.defineProperty()
>String.prototype.trim()

Plus `Document`, `DOMTokenList`, `Element`, `Event`, `Window` polyfills

I'm worried now that this plugin isn't catching those missed ones?

romaricpascal: Yeah, I'm not seeing it as an addition that'll catch everything, just automate some checks so we can focus on what actually requires a human eye. There's the caveat I added back in the description that the plugin only having rules for language features, not browser API features (so here go all the element ones). This rules out all the `Document`, `Window`, `DOMTokenList`, `Event`...

`Object.defineProperty` is only used in the polyfills so excluded from the linting (if you know of a way to add them back in the linting for this purpose, well happy to enable it). 

Good spot on `trim` though. It should be have been caught by the linting as [there's a rule for that](https://eslint-community.github.io/eslint-plugin-es-x/rules/no-string-prototype-trim.html) and `accordion.mjs` and `normalise-dataset.mjs` use it. I'll have a look.

colinrotherham: @romaricpascal Are there alternatives that will catch everything (or more than this)?

romaricpascal: Ah, reading the docs for `no-string-prototype-trim`, the rule doesn't do anything unless you set the plugin to be `aggressive`.

>This rule is silent by default because it's hard to know types. You need to configure [the aggressive mode](https://eslint-community.github.io/eslint-plugin-es-x/#the-aggressive-mode) or TypeScript in order to enable this rule.

The plugin mentions TypeScript, I'll have a quick try over the `fix-types` branch to see if it does pick it up then (that would be pretty neat).

romaricpascal:
> @romaricpascal Are there alternatives that will catch everything (or more than this)?

@colinrotherham There's [`eslint-plugin-compat`](https://www.npmjs.com/package/eslint-plugin-compat) as well, that may be worth exploring. [eslint-plugin-es5](https://www.npmjs.com/package/eslint-plugin-es5) seems focused on language features as well, and `eslint-plugin-es-x` looks sturdier (especially as it's under the [eslint-community](https://github.com/eslint-community/) org). Once we're out of ES3-land, [`eslint-plugin-ecmascript-compat`](https://www.npmjs.com/package/eslint-plugin-ecmascript-compat) may be relevant (but it lints only ES2016+ features).

There may be checks that could be done during compilation rather than at linting. For ex. we could try to detect which transformations Babel tries to apply when running on our code. Last I checked, though, I couldn't find anything to see when Babel actually transformed the AST to turn a modern syntax in older code, just [which plugins it enabled based on the browserlist](https://github.com/babel/babel/blob/main/packages/babel-compat-data/data/plugins.json). For polyfilling, we may get more information thanks to the [shouldInjectPolyfill](https://github.com/babel/babel-polyfills/pull/13) option, which would let us track what Babel actually tries to inject.

All in all, this PR is more a quick try to unlock the [terser PR](https://github.com/alphagov/govuk-frontend/pull/3013). Looks like we need more research if we want a tighter safety net on ES and browser features.

romaricpascal:
> Will it have caught this one?
>
> * https://github.com/alphagov/govuk-frontend/pull/2868

Yup, the [es-x/no-block-scoped-variables](https://eslint-community.github.io/eslint-plugin-es-x/rules/no-block-scoped-variables.html) rule catches that.

romaricpascal:
I think syntax-wise, the plugin will do the job fine (it's got all the information in the AST). Judging by the rules, it'll also catch any instanciation of new classes. Where it'll leave a gap is for detecting call to methods as it cannot infer things from types.

36degrees (commented):
Makes sense to me, but should these changes be in `src/.eslintrc.js` instead?

.eslintrc.js
colinrotherham: @romaricpascal Had a quick try and the prototype checks came to life 😊 

Turns out we don't need `aggressive: true` with [`@typescript-eslint/parser`](https://www.npmjs.com/package/@typescript-eslint/parser) 🎉 

For comparison: https://github.com/alphagov/govuk-frontend/compare/lint-types...eslint-plugin-es-x-opt-in

```console
/path/to/project/govuk-frontend/src/govuk/components/accordion/accordion.mjs
  145:49  error  ES5 'Function.prototype.bind' method is forbidden  es-x/no-function-prototype-bind
  149:46  error  ES5 'Function.prototype.bind' method is forbidden  es-x/no-function-prototype-bind
  167:5   error  ES5 'Function.prototype.bind' method is forbidden  es-x/no-function-prototype-bind
  178:41  error  ES5 'Function.prototype.bind' method is forbidden  es-x/no-function-prototype-bind
  342:5   error  ES5 'Function.prototype.bind' method is forbidden  es-x/no-function-prototype-bind
  383:25  error  ES5 'String.prototype.trim' method is forbidden    es-x/no-string-prototype-trim
  388:25  error  ES5 'String.prototype.trim' method is forbidden    es-x/no-string-prototype-trim
```

romaricpascal:
> Makes sense to me, but should these changes be in `src/.eslintrc.js` instead?

Yup, had missed that there was a separate config in `src`. I've tidied that up and also move the rule disabling next to the polyfill imports as @colinrotherham suggested (that should help us catch when we forget to import a polyfill for a feature that `eslint-plugin-es-x` can detect 😄

src/.eslintrc.js
```diff
@@ -23,6 +23,22 @@ module.exports = {
         ]
       }
     },
+    {
+      files: ['govuk/**/*.{cjs,js,mjs}'],
```
colinrotherham, line 27: Not sure if this speeds things up, but I think our source code is _only_ ESM now?

My mistake originally (appreciate you've just moved it)

```suggestion
      files: ['govuk/**/*.mjs'],
```

src/.eslintrc.js
```diff
@@ -23,6 +23,22 @@ module.exports = {
...
+      excludedFiles: ['govuk/**/*.test.{cjs,js,mjs}'],
```
colinrotherham, line 28: Which means we only need to exclude ESM tests

```suggestion
      excludedFiles: ['**/*.test.mjs'],
```

src/.eslintrc.js
```diff
@@ -23,6 +23,22 @@ module.exports = {
...
+      parserOptions: {
+        // Note: Allow ES6 for import/export syntax (although our code is ES3 for legacy browsers)
+        ecmaVersion: '2015'
+      },
+      plugins: ['es-x'],
+      extends: ['plugin:es-x/restrict-to-es3'],
+      rules: {
+        // We use ES6 modules to structure our code
```
colinrotherham, line 36: Want to mention Rollup on this line too?

```suggestion
        // Rollup transpiles modules to AMD export/define
```

src/govuk/common/normalise-dataset.mjs
```diff
@@ -1,5 +1,5 @@
 import '../vendor/polyfills/Element/prototype/dataset.mjs'
-import '../vendor/polyfills/String/prototype/trim.mjs'
+import '../vendor/polyfills/String/prototype/trim.mjs' /* eslint-disable es-x/no-string-prototype-trim */
```
36degrees, line 2: I'm surprised this works as my understanding is this [only disables the rule for this line](https://eslint.org/docs/latest/user-guide/configuring/rules#using-configuration-comments-1:~:text=To%20disable%20a%20specific%20rule%20on%20a%20specific%20line%3A) – and I'd expect the linting rule to be triggered where we're calling `String.prototype.trim` rather than at the point we're importing the polyfill? 🤔 

(Same for everywhere else we're doing this)

36degrees: Apparently I can't read – that example clearly uses `eslint-disable-line` 🤦🏻 

I think you can probably ignore this, but wonder if others might get tripped up on what this is doing.

romaricpascal:
@colinrotherham Should be all tidy based on your last comments 😄

src/govuk/common/normalise-dataset.mjs
```diff
@@ -1,3 +1,6 @@
+
```
colinrotherham, line 1:
```suggestion
```

colinrotherham (approved):
Approved! 🎉 

(One stray empty line you'll probably want to remove)

src/govuk/common/normalise-dataset.mjs
romaricpascal: Good spot! Thanks 😄

PR #3154, opened 2023-01-10
Remove Chromium from GitHub Actions cache

colinrotherham:
Whilst investigating the [`puppeteer@19.5.0` update PR](https://github.com/alphagov/govuk-frontend/pull/3152) I noticed Chromium was ~283MB on disk (macOS) and we were saving and restoring it to each GitHub Actions workflow step

This is one of the reasons we've seen Windows test runs perform slowly in:

* https://github.com/alphagov/govuk-frontend/pull/2945

This PR defers the Puppeteer download (~137MB compressed) until Jest "**JavaScript component tests**" run

### Before
```
Cache Size: ~180 MB (188782653 B)
```

### After
```
Cache Size: ~42 MB (44379878 B)
```

### DIFF
--- a/.github/workflows/actions/build/action.yml
+++ b/.github/workflows/actions/build/action.yml
@@ -10,7 +10,7 @@ runs:
 
       with:
         # Restore build cache (unless commit SHA changes)
-        key: build-cache-${{ github.sha }}
+        key: build-cache-${{ runner.os }}-${{ github.sha }}
         path: |
           jsdoc
           public
--- a/.github/workflows/actions/install-node/action.yml
+++ b/.github/workflows/actions/install-node/action.yml
@@ -10,7 +10,7 @@ runs:
 
       with:
         # Restore `node_modules` cache (unless packages change)
-        key: npm-install-cache-${{ hashFiles('package-lock.json', '**/package.json') }}
+        key: npm-install-cache-${{ runner.os }}-${{ hashFiles('package-lock.json', '**/package.json') }}
         path: node_modules
 
     - name: Setup Node.js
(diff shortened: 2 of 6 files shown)


tasks/browser.mjs
```diff
@@ -0,0 +1,25 @@
...
+  // Download latest browser (unless cached)
+  if (!versions.includes(puppeteer.browserRevision)) {
+    await fetcher.download(puppeteer.browserRevision)
+
+    // Remove outdated browser versions
+    for (const version of versions) {
+      await fetcher.remove(version)
+    }
+  }
+}
```
romaricpascal, line 25: *question*: Could we use Puppeteer's public installation script instead of maintaining our own? Looks like [`puppeteer/install.js`](https://github.com/puppeteer/puppeteer/blob/main/packages/puppeteer/install.js) is meant for doing the browser download? I'm a bit warry of having to keep our own script up to date when Puppeteer already maintains one 😄

romaricpascal (commented):
Good idea to move when we download Puppeteer to right before we need it 🙌🏻  Only got a quick question about having our own script rather than using [Puppeteer's](https://github.com/puppeteer/puppeteer/blob/main/packages/puppeteer/install.js) 😊

tasks/browser.mjs
colinrotherham: @romaricpascal Yeah this is what we were using on `npm ci` (behind the scenes) but it gets cached

Stopped it running with `{ skipDownload: true }` but yep happy to see if I can import it

Leave it with me

colinrotherham: Ahh it was a nice idea but it just [logs](https://github.com/puppeteer/puppeteer/blob/main/packages/puppeteer/src/node/install.ts#L40):

```console
**INFO** Skipping browser download as instructed.
```

Only other idea is to flip `PUPPETEER_SKIP_DOWNLOAD` **true** or **false** in the GitHub workflows?

romaricpascal (approved):
Ace! Thanks for updating 🙌🏻  I'd say it's good to go (not sure why Github shows its deployment errored, though. Heroku shows [the branch deployed OK and it's all up](https://govuk-frontend-pr-3154.herokuapp.com/)) 🚀

PR #3156, opened 2023-01-11
Correcting incorrect closing double quotes in pagination Nunjucks

JoPintoPaul:
Spotted an incorrectly placed closing double quote in the template for the new Pagination component.

PR #3157, opened 2023-01-11
Add custom class and attribute tests to pagination

owenatgov:
Adds test for the classes and attributes params for the pagination component.

It was revealed in https://github.com/alphagov/govuk-frontend/pull/3156 there there's been an error hidden in the pagination nunjucks macro since it's release. This would've been caught if we had proper testing for the attributes param, which we didn't 😱 This PR rectifies that gap in pagination's tests.


src/govuk/components/pagination/template.test.js
```diff
@@ -159,4 +159,20 @@ describe('Pagination', () => {
...
+  describe('custom classes and attributes', () => {
+    it('renders with custom additional classes', () => {
+      const $ = render('pagination', examples['with custom classes'])
+
+      expect($('.govuk-pagination').hasClass('my-custom-class')).toBeTruthy()
+    })
+
+    it('renders with custom attributes', () => {
+      const $ = render('pagination', examples['with custom attributes'])
+      const nav = $('.govuk-pagination')
```
colinrotherham, line 172: Not a blocker, but probably add the `$` prefix?

```suggestion
      const $nav = $('.govuk-pagination')
```

colinrotherham (approved):

PR #3162, opened 2023-01-12
[SPIKE] Webdriver.io set up

romaricpascal:
This branch is a quick set up of webdriver.io. Unlike Cypress or Playwright, this tool could let us run our test against older browsers to make sure we don't break anything in the lowest versions we're supporting. This feature would require us to have a SauceLabs or BrowserStack account where the tests would remotely run.

This branch only looks at a quick set up of the tool and the port of a few Character Count tests to get a feel for how the tests are organised and the syntax, which is a bit different than Cypress/Playwright and bit less abstracted. It does not set it up to run against old browsers. An example of such set up is available in the [Accessible Autocomplete repository](https://github.com/alphagov/accessible-autocomplete/blob/main/test/wdio.config.js). Algolia's instantsearch repository also offers [an example of such implementation](https://github.com/algolia/instantsearch/blob/772c918f47aec183af3f1aa78c65505f70dd0088/tests/e2e/wdio.saucelabs.conf.js).

Run `npm run wdio` to run the tests in Chrome with Webdriver.io. You can use `DEBUG=1 npm run wdio` to run the tests with a visible browser. Using `await browser.debug()` in the test code [will pause the test and let you inspect the page, as well as send it further webdriver commands](https://webdriver.io/docs/api/browser/debug).

You can use `DEBUG=inspect npm run wdio`, [connect Chrome's devtool to the test process](https://nodejs.org/en/docs/guides/debugging-getting-started/) (opening chrome://inspect/#devices in the browser) and [add the `govuk-frontend` folder to the workspace](https://developer.chrome.com/docs/devtools/workspaces/#devtools) to debug with actual breakpoints.

### Browser support

webdriver.io will control the browsers installed on the machine or pointed to a specific binary to run (which is how the Accessible Autocomplete points it to the browser from Puppeteer).

Note that to investigate a specific browser, we'd still be able to point it to SauceLabs or Browserstack just like CI would.

### Code abstractions we're provided for interacting with the browser

Tests are provided a `browser` global as well as a `$` and `$$` method for selecting elements. The API is very close to the WebDriver protocol, making it maybe a bit lower level than Playwright's or Cypress'. 

### Accessibility oriented API (eg. integration with axe, selection by accessibility tree)

[`@axe-core/webdriverio`](https://www.npmjs.com/package/@axe-core/webdriverio) should let us run axe through webdriver.io, similarly to how we do with Puppeteer.

The tool also offers [an `aria/` selector similar to Puppeteer's](https://webdriver.io/docs/selectors#accessibility-name-selector) allowing to select by accessible name. In practice, though, it is not able to [reliably return a form control when looking for its label's accessible text](https://github.com/webdriverio/webdriverio/issues/8826) 😢 It's also worth noting that the selector does an XPath query, so [does not follow the whole accessible name computation](https://www.w3.org/TR/accname/).

On the bright side, given the protocol follows Webdriver, we get access to [`getComputedLabel`](https://webdriver.io/docs/api/element/getComputedLabel) and [`getComputedRole`](https://webdriver.io/docs/api/element/getComputedRole) to access accessibility info.

### Watch mode

The `--watch` option will make `wdio` watch files and re-run tests as they change. The configuration needed a bit of fiddling, first with an exception breaking [re-runs when running chromedriver to control Chrome locally.](https://github.com/webdriverio/webdriverio/issues/9608). Workarounds are available, either [to keep using the WebDriver protocol](https://github.com/alphagov/govuk-frontend/pull/3162/commits/2689c5b606d1284d30e05333717dcf797a1f6368) or [switch to the devtools protocol](https://github.com/alphagov/govuk-frontend/pull/3162/commits/709a457ad22bb5962350d5568e2dd0624ec9f25c) when running locally. The later is not necessarily a bad thing as locally we'll likely want to run in a single browser to keep a snappy feedback loop.

### Debugging tests locally

The configuration is set to reveal the browser and lengthen jasmine's timeout when running with the `DEBUG=1` flag (or whatever value really). From there you can use `browser.debug()` to pause the script and get a REPL in your tests for inspecting the browser's content.

If the `DEBUG` value is `inspect` the configuration will set up Webdriver.io so you can [connect a debugger](https://nodejs.org/en/docs/guides/debugging-getting-started/) (opening Chrome and going to chrome://inspect/#devices) to the process running the tests. The tests are paused until the debugger connects in that case, so we don't have to rush.

### Running specific tests

The branch configures Webdriver to use jasmine behind the scenes. Tests can be isolated with `iit` or `ddescribe` (instead of the `it` or `describe` functions). It may not be the most intuitive while running alongside our regular tests, so we may want to run Mocha behind the scene, which supports [the same `it.only`/`describe.only` syntax as Jest](https://mochajs.org/#exclusive-tests).

I'll need to investigate how to configure to run only specific files as passing a partial name of the file didn't do the trick.

### Browser support

Besides being configured to run against Puppeteer's binary inside a Github action, WebDriver's appeal is that it can connect to SauceLabs or BrowserStack and give us access to a wide range of browsers, and more importantly, older versions. This is not implemented on this branch, but an example of such configuration is available on the [Accessible Autocomplete repository](https://github.com/alphagov/accessible-autocomplete/blob/main/test/wdio.config.js).

Note that if we want to look into [tracking code coverage](https://webdriver.io/docs/devtools-service/#capture-code-coverage), we may have to look at running both:
- a devtools protocol test piloting a local (Puppeteer's?) Chrome to gather coverage
- webdriver protocol tests against the range of browsers we care about (except Chrome, tested via the devtools) on SauceLabs/Browserstack

### Test reports

WebDriver would log its output inside the Github action. For example on the Accessible Autocomplete: https://github.com/alphagov/accessible-autocomplete/actions/runs/3739372148/jobs/6346502803

[SauceLabs can give us quite a lot of insights about failures](https://app.saucelabs.com/tests/343d26116a0244158e0135e8c4dae712#141) before checking the issue locally:
 - trace of the different commands sent to the browser, which screenshots
 - video re-playing the test


domoscargin:
It _feels_ like we kinda have to go with this for browser support reasons, but I'd love it to be higher level than it is.

romaricpascal:
Closing the spike PR. Future work will be tracked by this epic: https://github.com/alphagov/govuk-frontend/issues/3278

christian-bromann:
Hey 👋  WebdriverIO maintainer here! If I can help make a better decision I am happy to answer questions around WebdriverIO. Let me know 😉

PR #3176, opened 2023-01-16
Rename "hide this page" to "exit this page"

owenatgov:
## What
Rename instances of "hide" to "exit" when referencing exit this page in the exit this page branch.

## Why
As part of the Exit this Page work, we agreed to rename "Hide this Page" to "Exit this Page". This codifies that decision in our implementation. We'd have to do this anyway later, no reason not to sort it out now


querkmachine (approved):
The heading of the component's README file needs updating, but I think that can slide given it hasn't been seriously worked on yet (the file in general seems to be a copy/paste from the Inset Text component). 

The strings in the examples, tests and Nunjucks parameter docs also still say "Hide this Page", but this is similarly draft data so I'm not sure if those need updating yet either.

I'd be inclined to update these instances just for blanket consistency across everything, but ultimately they're not parts of the actual implementation, so aren't required changes IMO.

owenatgov:
Cheers @querkmachine for the spots. Updated now for brevity.

PR #3181, opened 2023-01-18
Change Phase Banner's `tag` parameter to required

### LINKED ISSUE #1930: Unclear if phase banner `tag` is required

querkmachine:
Updates the Phase Banner's params data to mark the `tag` parameter as required.

The tag parameter is not currently listed as being required, however the component does not render correctly if it's omitted—it still renders the tag, just without any text. 

Screenshot

Closes #1930.

### DIFF
--- a/src/govuk/components/phase-banner/phase-banner.yaml
+++ b/src/govuk/components/phase-banner/phase-banner.yaml
@@ -9,7 +9,7 @@ params:
   description: If `text` is set, this is not required. HTML to use within the phase banner. If `html` is provided, the `text` option will be ignored.
 - name: tag
   type: object
-  required: false
+  required: true
   description: Options for the tag component.
   isComponent: true
 - name: classes


colinrotherham (approved):
Nicely spotted

Do we need a CHANGELOG entry for this?

Suppose teams would already notice if they'd left out the tag text

PR #3183, opened 2023-01-18
Allow `Function.prototype.bind` when already polyfilled

colinrotherham:
Just a few more ESLint `/* eslint-disable es-x/no-function-prototype-bind */` additions to https://github.com/alphagov/govuk-frontend/commit/06f5561d0015a6e3aad3aa1e2d14ccb97c4112b1

@romaricpascal These ones started showing up in:


romaricpascal (approved):
Oh indeed! Wonder why they didn't get detected during the initial linting after I set up eslint-plugin-es-x 🤔 That's a bit worrying.

romaricpascal:
Thanks for spotting them! 🙌🏻

colinrotherham:
@romaricpascal Don't think any of the `.prototype` ones did without the TypeScript ESLint parser?

romaricpascal:
Oh yeah, for those, ESLint as no clue we're binding functions 😮‍💨 I think it had only picked up the
`function () {/** Do something */}.bind(this)` ones so far because ESLint detects the FunctionExpression on its own.

PR #3184, opened 2023-01-18
Fix PostCSS condition for Review app “pseudo-classes” plugin

colinrotherham:
This PR restores the [`postcss-pseudo-classes`](https://github.com/giuseppeg/postcss-pseudo-classes) PostCSS plugin for [review app](https://govuk-frontend-review.herokuapp.com) stylesheets

We use the plugin to generate classes (see below) for the following:

```
:link
:visited
:hover
:active
:focus
```

Unfortunately our config hasn't matched minified stylesheets (such as `app.min.css`) since merging:

* https://github.com/alphagov/govuk-frontend/pull/3050

Pseudo classes on form elements

This PR includes:

1. **Fix PostCSS condition for Review app “pseudo-classes” plugin**
We can match all `'**/app/assets/scss'` child stylesheets regardless of name or extension

2. **Adjust Sass task to run PostCSS first before renaming**
For some Gulp streams we were renaming/moving files before PostCSS could check the path

3. **Move PostCSS `cssnano` to default plugins**
We minify our CSS by default so can simplify our tests

### DIFF
--- a/tasks/gulp/compile-assets.mjs
+++ b/tasks/gulp/compile-assets.mjs
@@ -33,22 +33,25 @@ export function compileStylesheets () {
         gulp.src(`${slash(paths.src)}/govuk/all.scss`, {
           sourcemaps: true
         })
-          .pipe(rename({
-            basename: 'govuk-frontend',
-            suffix: `-${pkg.version}`
-          }))
-      ),
+      )
+        .pipe(rename({
+          basename: 'govuk-frontend'
+        })),
 
       compileStylesheet(
         gulp.src(`${slash(paths.src)}/govuk/all-ie8.scss`, {
           sourcemaps: true
         })
-          .pipe(rename({
-            basename: 'govuk-frontend-ie8',
-            suffix: `-${pkg.version}`
-          }))
       )
+        .pipe(rename({
+          basename: 'govuk-frontend-ie8'
+        }))
     )
+      .pipe(rename({
+        suffix: `-${pkg.version}`,
+        extname: '.min.css'
+      }))
+
       .pipe(gulp.dest(slash(destPath), {
         sourcemaps: '.'
       }))
@@ -75,12 +78,16 @@ export function compileStylesheets () {
       gulp.src(`${slash(paths.fullPageExamples)}/**/styles.scss`, {
         sourcemaps: true
       })
-        .pipe(rename((path) => {
-          path.basename = path.dirname
-          path.dirname = 'full-page-examples'
-        }))
     )
+      .pipe(rename((path) => {
+        path.basename = path.dirname
+        path.dirname = 'full-page-examples'
+      }))
   )
+    .pipe(rename({
+      extname: '.min.css'
+    }))
+
     .pipe(gulp.dest(slash(destPath), {
       sourcemaps: '.'
     }))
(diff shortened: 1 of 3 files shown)


tasks/gulp/compile-assets.mjs
```diff
@@ -75,12 +78,16 @@ export function compileStylesheets () {
...
         sourcemaps: true
       })
-        .pipe(rename((path) => {
-          path.basename = path.dirname
-          path.dirname = 'full-page-examples'
-        }))
     )
+      .pipe(rename((path) => {
+        path.basename = path.dirname
+        path.dirname = 'full-page-examples'
```
colinrotherham, line 84: After Sass had compiled from `*.scss` to `*.css`, PostCSS was seeing the path as:

~`app/views/full-page-examples/campaign-page/styles.css`~
`app/views/full-page-examples/full-page-examples/campaign-page.css`

To reduce the chance of error in future let's rename last (before writing to disk)

1. Read source file to stream
2. Compile Sass from `*.scss` to `*.css`
3. Transform `*.css` using PostCSS
5. Rename or modify path
6. Write file to destination

querkmachine:
Unfortunately it doesn't seem to be transforming all selectors as expected. (The :focus examples in your screenshot aren't styled correctly, for example.) 

Complex selectors seem to only be getting partially transformed — in the below case,`:hover` is getting transformed but `:focus` isn't. 

```css
/* Before */
.govuk-radios--small .govuk-radios__item:hover .govuk-radios__input:focus + .govuk-radios__label:before {}

/* After (other selector output omitted for brevity) */
.govuk-radios--small .govuk-radios__item.\:hover .govuk-radios__input:focus + .govuk-radios__label:before {}
```

I think this may be a bug with postcss-pseudo-classes' `restrictTo` option. Removing that part of the PostCSS configuration generates the expected selectors in the above example, but also results in the janked `:not` selectors that [the old configuration warned about](https://github.com/alphagov/govuk-frontend/blob/8fd787e636717bedcd069047d4bb4960d2fba0e4/tasks/gulp/compile-assets.js#L21-L25).

Simply swapping out for the old configuration doesn't seem to work in all cases either, as the `:not` selector jankiness seems to remain. Here the `:not(:disabled)` pseudo-selector shouldn't be transformed, as we've blacklisted it, but it's being transformed anyway.

```css
/* Before */
.govuk-radios--small .govuk-radios__item:hover .govuk-radios__input:not(:disabled) + .govuk-radios__label:before {}

/* After (other selector output omitted for brevity) */
.govuk-radios--small .govuk-radios__item.\:hover .govuk-radios__input.\:not\(.\:disabled\) + .govuk-radios__label:before {}
```
This happens with both the previous config (`blacklist: [':not(', ':disabled)']`) and with a more specific, targeted version (`blacklist: [':not(:disabled)']`).

I'm not sure why the messed up `:not` selectors keep happening despite being blacklisted, nor am I sure whether this was an existing issue or is a regression, but either way it's probably not the highest priority given this is code specific to the review app and not something end users are using. 

I'm going to try and make reduced test cases for each to see if they're bugs inherent in postcss-pseudo-classes and will report them if so.

colinrotherham:
Thanks @querkmachine, great spot—another bug!

Looks like we need to tread _very_ carefully with these legacy PostCSS plugins

I've fixed it by letting them run _before_ we minify 🙌 

(The consolidated selectors that CSSNANO outputs must be too complex for `postcss-pseudo-classes`)

querkmachine (approved):
Everything generated as expected now that cssnano has been moved down the order! 👍

PR #3186, opened 2023-01-19
Fix linting on commit for files ignored by ESLint

romaricpascal:
ESLint's configuration makes it ignore built files in `dist` or `package` that we want left alone, as well as the polyfills.

The glob used by lint-staged to trigger the linting on commit isn't aware of that ignore list, so will ask ESLint to lint those files.

This makes ESLint raise a warning for these files, which errors the linting because we use `--max-warnings 0`.

To avoid that, we need to filter out files ignored by ESLint, as recommended by lint-staged.

### DIFF
--- a/.lintstagedrc.js
+++ b/.lintstagedrc.js
@@ -1,6 +1,37 @@
+const { ESLint } = require('eslint')
+
 module.exports = {
-  // ESLint's configuration will let it ignore built files
-  // in `dist` or `package` that we want left alone, as well as the polyfills
-  '*.{cjs,js,mjs}': ['npm run lint:js:cli -- --fix'],
+  // ESLint's configuration makes it ignore built files in `dist` or `package`
+  // that we want left alone, as well as the polyfills.
+  // The glob used by lint-staged to trigger the linting on commit isn't aware
+  // of that ignore list, so will ask ESLint to lint those files.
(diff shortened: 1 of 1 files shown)


36degrees (approved):
👍🏻

.lintstagedrc.js
```diff
@@ -1,6 +1,37 @@
...
+  // that we want left alone, as well as the polyfills.
+  // The glob used by lint-staged to trigger the linting on commit isn't aware
+  // of that ignore list, so will ask ESLint to lint those files.
+  // This makes ESLint raise a warning for these files, which errors the linting
+  // because we use `--max-warnings 0`.
+  // To avoid that, we need to filter out files ignored by ESLint,
+  // as recommended by lint-staged.
+  //
+  // https://github.com/okonet/lint-staged#how-can-i-ignore-files-from-eslintignore
+  '*.{cjs,js,mjs}': async (files) => {
```
colinrotherham, line 14: Wonder if it would be any simpler to unwrap one of the promises and move `Promise.all()` up here?

```suggestion
  '*.{cjs,js,mjs}': (files) => Promise.all(files.map(async (file) => {
```

colinrotherham: Ahh too late, merged it

Issue #1713, opened 2020-01-24
Ensure input components have consistent disabled styling


NickColley:
## What
Some input components do not look disabled when tested cross-browser, we should try to add styles to input components to makes sure there is an intentional disabled state that works across browsers.

## Why
We don't recommend users disabled their inputs but if they do we want them to be as usable and accessible as possible.

## Who needs to know about this
Developers, Chris

## Further detail

If possible, contrasts should aim for WCAG guidelines to level AA but this is not a requirement.

See this related issue for more context: https://github.com/alphagov/govuk-frontend/issues/1692#issuecomment-574274312

## Done when

- [ ] File upload has disabled styling
- [ ] Select has disabled styling
- [ ] Text input has disabled styling
- [ ] Textarea has disabled styling

timpaul:
_"I wonder if styling the disabled button with `cursor: not-allowed` would provide more affordance to the user?"_

_Originally posted by @jesseyuen in https://github.com/alphagov/govuk-design-system-backlog/issues/34#issuecomment-528365796_

edwardhorsford:
Another to add: items in autocomplete have disabled styling (for when enhancing a select). It's common practice to have the first item be disabled and something like `Select your foo`.

querkmachine:
I've been looking into this as a random side thing. As we currently rely on user agent styles for disabled states, we have quite a lot of inconsistency in how they appear across different browsers. In some cases, disabled inputs don't appear visually distinct from enabled ones whatsoever. 

Based on checking the most recent browsers (and IE 11):

|Browser|Text input/textarea|Select|File upload|
|:-|:-|:-|:-|
|Chrome 109 (macOS and Windows)|Reduced contrast on text, light grey background added|Reduced contrast on text only|No visual differences|
|Safari 16.0 (macOS)|Reduced contrast on text only|No visual differences|Reduced contrast on “Choose file” button only|
|Safari 16.1 (iOS)|Reduced contrast on text and input border|Reduced contrast on arrowhead icon only|No visual differences|
|Firefox 108 (macOS and Windows)|Reduced contrast on text only|No visual differences|Reduced contrast on “Browse...” button only| 
|Edge 109 (Windows)|Reduced contrast on text, light grey background added|Reduced contrast on text only|No visual differences|
|Internet Explorer 11 (Windows)|Reduced contrast on text, light grey background added|Reduced contrast on text and arrowhead icon|Reduced contrast on “Browse...” button only|

[...]

Issue #1712, opened 2020-01-24
Add disabled option for input components consistently


NickColley:
## What
Some input components can be disabled but do not have an option in Nunjucks to do this, we should make sure all input components that can be disabled have this option.

## Why
Make sure our component APIs are predictable and consistent

## Who needs to know about this
Developers

## Further detail
We don't need to document this in the Design System guidance as an example as we don't recommend people do this, but it will be in the macro options.

See this related issue for more context: https://github.com/alphagov/govuk-frontend/issues/1692#issuecomment-574274312

## Done when

- [x] File upload has a top level 'disabled' option
- [x] Select has a top level 'disabled' option
- [x] Text input has a top level 'disabled' option
- [x] Textarea has a top level 'disabled' option

querkmachine:
This has been shipped as part of #3187. 🚀

Issue #1692, opened 2020-01-09
Disabled select text colour too faint?


kr8n3r:
this is the browser default style 
<img width="251" alt="Screenshot 2020-01-09 at 11 08 37" src="https://user-images.githubusercontent.com/3758555/72062841-7f62f700-32d0-11ea-8a83-954ec8c793e3.png">

nickcolley:
I've created a demonstration page of all our elements that can be disabled and screenshotted them across browsers.

Findings:

1. Some components have a styled disabled state
1. Some components dont have a 'disabled' option.
1. Some operating systems show a disabled state but others do not.

Recommendation:
1. create disabled states for elements that 
1. investigate if styling the disabled state for the select input would work across operating systems/browsers

<details>
<summary>Screenshots (generated by browserstack)</summary>

![win7_ie_9 0](https://user-images.githubusercontent.com/2445413/72365138-97cb7b00-36ef-11ea-868f-aa7f9cee3597.png)
![win7_ie_8 0](https://user-images.githubusercontent.com/2445413/72365140-98641180-36ef-11ea-8fde-3a864ef1bf34.png)
![win7_ie_11 0](https://user-images.githubusercontent.com/2445413/72365141-98641180-36ef-11ea-9d6e-29cb88d16eb5.png)
![win7_ie_10 0](https://user-images.githubusercontent.com/2445413/72365142-98fca800-36ef-11ea-9548-2a49661cb6fc.png)
![win10_edge_18 0](https://user-images.githubusercontent.com/2445413/72365143-98fca800-36ef-11ea-8960-4e951df607ad.png)
![win10_chrome_71 0](https://user-images.githubusercontent.com/2445413/72365144-98fca800-36ef-11ea-9d21-d6ab5b3cb0f7.png)
![macmo_safari_12 1](https://user-images.githubusercontent.com/2445413/72365147-99953e80-36ef-11ea-9a87-288b2e81e550.jpg)
![macmo_chrome_71 0](https://user-images.githubusercontent.com/2445413/72365148-99953e80-36ef-11ea-8239-38477e385c48.png)
![9 0_Samsung-Galaxy-S9-Plus_portrait_real-mobile](https://user-images.githubusercontent.com/2445413/72365149-99953e80-36ef-11ea-9554-cb1250d7c11b.jpg)
[...]

PR #3187, opened 2023-01-19
Add disabled styles for form controls

querkmachine:
Currently we rely on user agent styles to indicate the `disabled` state of all form controls except for radio buttons and checkboxes. For these, we reduce the opacity of both the input and associated label by 50%.

In my research, [the user agent styles for disabled controls are wildly inconsistent](https://github.com/alphagov/govuk-frontend/issues/1713#issuecomment-1397048516), both between browsers and different input types. Some user agents don't visually differentiate between enabled and disabled inputs at all! For this reason, I think we should explicitly define disabled styles rather than relying on UA styles. 

Whilst we don't tend to encourage the use of disabled inputs, this feels like one of those things we should accommodate within Frontend for when they are used or needed. 

This change was requested in #1713. I've also added `disabled` parameters to our Nunjucks macros as a linked change as was requested in #1712, though this can be pulled into a separate piece if desired.

## Changes
- Adds `:disabled` styles to the text input, textarea, select and file upload components, mirroring the 50% opacity reduction that we use for radio buttons and checkboxes. The actual CSS used for each differs slightly, as [explained in this comment](https://github.com/alphagov/govuk-frontend/issues/1713#issuecomment-1397048516). 
- Adds the `not-allowed` cursor style to disabled form controls and buttons when they are hovered over.
- Adds a `disabled` Nunjucks parameter to the text input, textarea, select and file upload components, with associated documentation.
- Expands the "Forms controls states" example in the Frontend review app to include more form controls and more states (and more bats).

### Text input / textarea

Textareas look pretty much identical to these, they're just taller and have a scrollbar, so I've not screenshotted them (sorry, there's just so many!) 

| |Before|After|
|:-|:-:|:-:|
|Chrome 109 (macOS)|image|image|
|Safari 16.0 (macOS)|image|image|
|Safari 16.1 (iOS)| image|image|
|Firefox 108 (macOS)|image|image|
|IE 11 (Windows 10)|image|image|

### Select

| |Before|After|
|:-|:-:|:-:|
|Chrome 109 (macOS)|image|image|
|Safari 16.0 (macOS)|image|image|
|Safari 16.1 (iOS)|image|image|
|Firefox 108 (macOS)|image|image|
|IE 11 (Windows 10)|image|image|

### File upload

| |Before|After|
|:-|:-:|:-:|
|Chrome 109 (macOS)|image|image|
|Safari 16.0 (macOS)|image|image|
|Safari 16.1 (iOS)|image|image|
|Firefox 108 (macOS)|image|image|
|IE 11 (Windows 10)|image|image|

### How well does this work in alternative colour modes (e.g. Windows High Contrast Mode)?
Windows HCM seems to universally apply a green border and text colour to disabled inputs, ignoring opacity-based styling entirely.

### How well does this work for users with low vision?
Disabled inputs do not need to meet the [3:1 non-text contrast ratio](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html) or [4.5:1 text contrast ratio](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum) required by WCAG, but we may want to ensure a minimum level of contrast regardless.

### How well does this work for sighted users?
Of particular concern to me is empty text inputs and textareas, where the lighter border colour is the only indicator that the control is disabled. This difference may not be clear if there are no other inputs to compare against.

Arguably it's unlikely that the user will be presented with a disabled input with no other options to compare it against. This doesn't mean the disabled styling is _obvious_, however.

### Do we want to add back in a grey background? If so, should that be mirrored in the radio button and checkbox styles?
This feels like it would convey the affordance of 'this input is disabled' a bit clearer, but it couldn't be applied consistently without some bigger changes—particularly on File Upload components and Selects on macOS. 

### Should we use `[disabled]` or `:disabled` for our CSS selectors?
We use `:disabled` virtually everywhere in Frontend, with the exception of the Button component, which uses `[disabled]`. Unsure if there's a specific reason for the difference, as `:disabled` seems to also work on `<button>`s. 

IE 8, however, doesn't support the `:disabled` pseudo-selector.

### DIFF
--- a/src/govuk/components/input/_index.scss
+++ b/src/govuk/components/input/_index.scss
@@ -44,6 +44,7 @@
       opacity: .5;
       color: inherit;
       background-color: transparent;
+      cursor: not-allowed;
     }
   }
 
--- a/src/govuk/components/input/_index.scss
+++ b/src/govuk/components/input/_index.scss
@@ -39,6 +39,12 @@
         border-width: $govuk-border-width-form-element * 2;
       }
     }
+
+    &:disabled {
+      opacity: .5;
+      color: inherit;
+      background-color: transparent;
+    }
   }
 
   .govuk-input::-webkit-outer-spin-button,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,35 @@
 
 ## Unreleased
 
+### New features
+
+#### Updated the appearance of disabled form controls
+
+The disabled state of Text Input, Textarea, Select and File Upload components has been updated to appear consistent across browsers and devices. They also now consistent with the existing disabled styles for Buttons, Checkboxes, and Radios.
+
+Disabled form controls appear with their opacity reduced by 50% and with an alternative cursor appearance when hovered over.
+
+This was added in [pull request #3187: Add disabled styles for form controls](https://github.com/alphagov/govuk-frontend/pull/3187).
+
+#### Added a top-level `disabled` parameter to form controls
+
+The Nunjucks macros for Text Input, Textarea, Select and File Upload components have been updated to include a top-level `disabled` parameter, making it easier to enable the disabled state for these controls.
+
+```nunjucks
+{{ govukInput({
+  id: "disabled-input",
+  name: "disabled-input",
+  value: "Unchangeable value",
+  disabled: true
+}) }}
+```
+
+Disabled form controls have poor contrast and can confuse some users, so avoid them if possible.
+
+Only use disabled form controls if research shows it makes the user interface easier to understand.
+
+This was added in [pull request #3187: Add disabled styles for form controls](https://github.com/alphagov/govuk-frontend/pull/3187).
+
 ### Fixes
 
 We’ve made fixes to GOV.UK Frontend in the following pull requests:
--- a/app/views/examples/form-controls-states/index.njk
+++ b/app/views/examples/form-controls-states/index.njk
@@ -1,8 +1,33 @@
 {% extends "layout.njk" %}
 
 {% from "back-link/macro.njk" import govukBackLink %}
-{% from "checkboxes/macro.njk" import govukCheckboxes %}
-{% from "radios/macro.njk" import govukRadios %}
+{% from "file-upload/macro.njk" import govukFileUpload %}
+{% from "input/macro.njk" import govukInput %}
+{% from "select/macro.njk" import govukSelect %}
+{% from "textarea/macro.njk" import govukTextarea %}
+
+{% set examples = [
+  { name: "Initial" },
+  { name: "Hover", classes: ":hover" },
+  { name: "Focus", classes: ":focus" },
+  { name: "Hover, Focus", classes: ":hover :focus" },
+  { name: "Disabled", attributes: { "disabled": "disabled" } },
+  { name: "Disabled, Hover", classes: ":hover", attributes: { "disabled": "disabled" } },
+  { name: "Error", error: true },
+  { name: "Error, Hover", error: true, classes: ":hover" },
+  { name: "Error, Focus", error: true, classes: ":focus" },
+  { name: "Error, Hover, Focus", error: true, classes: ":hover :focus" }
+] %}
+
+{% set examplesTextInputs = examples.concat([
+  { name: "Readonly",attributes: { "readonly": "readonly" } },
+  { name: "Readonly, Hover", classes: ":hover", attributes: { "readonly": "readonly" } },
+  { name: "Readonly, Focus", classes: ":focus", attributes: { "readonly": "readonly" } },
+  { name: "Readonly, Hover, Focus", classes: ":hover :focus", attributes: { "readonly": "readonly" } },
+  { name: "X-Ray", xray: true }
+]) %}
+
+{% set examples = examples.concat([{ name: "X-Ray", xray: true }]) %}
 
 {% block beforeContent %}
   {{ govukBackLink({
@@ -16,7 +41,9 @@
   box-shadow: 0 0 0 3px rgba(175, 175, 175, 0.5);
 }
 
-.x-ray input {
+.x-ray input,
+.x-ray select,
+.x-ray textarea {
   opacity: 1 !important;
   box-shadow: 0 0 0 3px rgba(0, 0, 255, 0.5);
   -webkit-appearance: none;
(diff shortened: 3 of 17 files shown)


owenatgov (commented):
Thanks for looking into this. Very thorough as usual. From a skim of the code I don't see any problems however I have some broad thoughts:

- As you noted in the description, the disabled styles are kinda tricky to see. I'm especially worried about the form upload component. We do need some kind of minimum contest or some way to alleviate the muted colours I reckon. I wonder if there's a more nuanced way to apply the disabled styling: for the form upload targeting the button so that the border has higher contrast but the text is still muted.
- Related to the above point, it feels like a risk for us to be relying on a colour change as the only indicator that a control is disabled. I think an extra visual indicator, like the error styling with the border, would be useful here. The only thing I can think of is applying a dotted border to controls or using `:before` to literally add the word "disabled" somewhere.
- I'm not worried about the use of `:disabled`. It doesn't break anything for IE8 users in the literal sense. They can still "use" the service even if they're getting a worse experience.
- Not one for now necessarily but we should probably expand on our content for inputs. We currently on purpose don't talk about disabled styling to dissuade teams from using them but this feels like a "meet them where they are" situation. Especially if we have these attributes in here now, we should say somewhere that teams shouldn't use it ideally. Maybe in the API docs?

I think this is gonna need a designer's brain, especially on the contest and visual indication points.

All of my thoughts are tempered by the fact that we don't want teams to use disabled form controls in the first place, so our solutions don't have to be perfect but I reckon we can try to mitigate risk of teams putting together inaccessible forms.

querkmachine:
I agree that the disabled style could benefit from a stronger visual indication. My goal here has been to not rock the boat too much and reuse the 50% opacity that we have on buttons, checkboxes and radio buttons. 

If we want to investigate other ways of indicating that something is disabled, I think that whatever styles we decide should be back-portable to the checkboxes and radios (at least) so that all form inputs maintain a consistent look. A dotted border sounds doable across most input types—maybe not file input—though we could try using `::file-selector-button` in supporting browsers to do something for it, perhaps?

colinrotherham:
Some context on other "disabled" discussion here:

* https://github.com/alphagov/govuk-frontend/pull/2830

Would be good to fill in the consistency cracks though

One of the `[disabled]` vs `:disabled` behaviours is this annoying one:
https://www.mehulkar.com/blog/2018/01/disabled-inputs/

```html
<form>
    <fieldset disabled="true">
        <input />
    </fieldset>
</form>
```

The input can be selected with CSS `input:disabled` but not `[disabled]` (CSS) or `.disabled` (JS)

src/govuk/components/input/_index.scss
```diff
@@ -39,6 +39,13 @@
         border-width: $govuk-border-width-form-element * 2;
       }
     }
+
+    &:disabled {
+      opacity: .5;
+      color: inherit;
+      background-color: transparent;
```
colinrotherham, line 46: What does the background colour do here out of interest?

With us not having one on `.govuk-input`

querkmachine: Chrome and Internet Explorer—and only those two browsers, for whatever reasons—apply a light grey background-color to disabled text inputs and textareas. Here I've opted to remove it so that the background-color is consistent across all text inputs on all browsers.

colinrotherham: Ahh that's good to know

In a world of "extended components" in the various organisational-level design systems, I'm wondering if this will make some department's (blue|green|red|whatever)-tinted input suddenly go transparent when disabled?

Should `background-color: transparent` be the component default?

I know it came up recently didn't it in this one:

* https://github.com/alphagov/govuk-frontend/issues/1625

CharlotteDowns:
I think this is a good addition to include labels into the change in styling of the associated form actions when `:disabled`. It is also good to acknowledge that [the browser native styling doesn't work on all components in all cases](https://github.com/alphagov/govuk-frontend/issues/1692#issuecomment-574274312).

I'm not sure how important legibility of `:disabled` components is when the intent is for them not to be used, I would hope there's some form of explanation on the page as to why something is no longer useable or active. This is I think why we discourage the use, it's hard to understand a context where a service is built for a user but doesn't allow them to do something.

I also agree that keeping the opacity consistent with radios and checkboxes is essential. Do we need to include anything about pointers or will these change my default?

querkmachine:
> I'm not sure how important legibility of `:disabled` components is when the intent is for them not to be used, I would hope there's some form of explanation on the page as to why something is no longer useable or active. 

It’s my understanding that it’s pretty much at our discretion. The two WCAG criteria relating to contrast of text and non-text content both have exceptions for ‘inactive’ components.

[SC 1.4.3: Contrast (Minimum)](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.html) says:

> The visual presentation of text and images of text has a contrast ratio of at least 4.5:1, except for […] Text or images of text that are part of an inactive user interface component

[SC 1.4.11: Non-text Contrast](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html) says:

> The visual presentation of the following have a contrast ratio of at least 3:1 against adjacent color(s): Visual information required to identify user interface components and states, except for inactive components

---

> This is I think why we discourage the use, it's hard to understand a context where a service is built for a user but doesn't allow them to do something.

We already account for individiual inputs or options within a set being unavailable, which is why it's already possible and we have styles for disabled Checkboxes, Radios and items within a Select. 

The likeliness of a disabled form controls within a service is admittedly quite low, however this is something that is more likely to be seen in a caseworking system or in a more complex user interface, such as a faceted search UI. 

Like the existing styles we have for disabled controls, I feel that this is more about making sure there is a consistent baseline already present for when disabled controls are used, rather than wanting to encourage their use unnecessarily. 

---

> I also agree that keeping the opacity consistent with radios and checkboxes is essential. Do we need to include anything about pointers or will these change my default?

Currently this PR makes it so that all disabled form controls and buttons show the 'not-allowed' cursor type on hover. This displays as some variation of a forbidden symbol (🚫) on all versions of Windows and macOS released in the last 15 years or so.

christopherthomasdesign:
Will try not to repeat what others have said... I basically think it makes sense to make the states more consistent and robust across components.

Like others, I feel uncomfortable with having UI components on a page that aren't legible to a lot of people (even if they can't be used). But I think it'd be a bigger and broader piece of work to look at how we style 'disabled' components in a way that all users can perceive.

src/govuk/components/input/_index.scss
CharlotteDowns: @colinrotherham that's a good point

owenatgov:
Putting on my product hat, this sounds like we're comfortable with putting out a version of disabled inputs which creates some consistency and exposes the option if teams absolutely need it with a view to make it extra good in the long term but are less comfortable with the levels of legibility in their current state. With that in mind, following some design/dev collaboration between @CharlotteDowns @querkmachine I'm happy to re-review

CharlotteDowns:
@querkmachine happy to share preparing this for the next GOV.UK Design System show and tell 28 Feb based on the discussion in the team retro this AM.

CharlotteDowns:
The understanding documentation from WCAG has a [section on contrast requirements for Inactive User Interface Components](https://www.w3.org/WAI/WCAG22/Understanding/non-text-contrast#inactive-controls).

> User Interface Components that are not available for user interaction (e.g., a disabled control in HTML) are not required to meet contrast requirements in WCAG 2.1.
...
> Although it would be benefitial to some people to discern inactive controls

In my opinion I think the change we are proposing here provides enough contrast to discern the component whilst also keeping it consistent with existing styling on components that we have already had to apply `:disabled`, in active, styling.

CharlotteDowns (approved):
[We strongly encourage teams not to use them](https://design-system.service.gov.uk/components/button/#disabled-buttons) but if a team needs to use them this pull request improves the 'in active' styling over the default browser styling that is currently used.

We've also made sure that these styles align to those seen in other components within the GOV.UK Design System which have had to have custom `:disabled `styling.

owenatgov (approved):
Giving a technical approval plus a few opinions:

1. I think that it's worth us closing https://github.com/alphagov/govuk-frontend/issues/1713 and https://github.com/alphagov/govuk-frontend/issues/1712 as part of this work and creating a new card to figure out what to do with disabled styles both in terms of guidance and design.
2. When we eventually revisit this and attempt to "solve" disabled styles I think it'd be valuable to distribute them as mixins the same way we do other states like link hover and focus. I'm not gonna split hairs about that now as the impact is pretty minimal.

querkmachine:
David's had a look and is happy to progress this as-is, so I've added a changelog and will merge in a moment!

PR #3164, opened 2023-01-13
Migrate GOV.UK Frontend build pipeline to use Dart Sass

colinrotherham:
This PR swaps "**Node Sass**" [`node-sass`](https://www.npmjs.com/package/node-sass) with "**Dart Sass**" ~[`sass`](https://www.npmjs.com/package/sass)~ [`sass-embedded`](https://www.npmjs.com/package/sass-embedded)

There are some minor output differences to be aware of:

1. No new lines between `@media` queries for **expanded** style
2. Number precision now at 10 decimal places ~`1.42857`~ `1.4285714286`
3. Legacy (LibSass) global functions renamed ~`to_upper_case`~ `to-upper-case`
4. Slightly re-ordered comma separated selectors

Other changes in preparation have been moved to:

* https://github.com/alphagov/govuk-frontend/pull/3191
* https://github.com/alphagov/govuk-frontend/pull/3192
* https://github.com/alphagov/govuk-frontend/pull/3193

Plus this swap has previously been looked at in an older PR:

* https://github.com/alphagov/govuk-frontend/pull/2583

Closes: https://github.com/alphagov/govuk-frontend/issues/2239

Issue #2238, opened 2021-05-28
Sass deprecation warning for slash as division


colinrotherham:
Hello 👋 

It's been a while since I was on **govuk-frontend** issues 😆 

## Description of the issue

We've just noticed Sass deprecation warnings popping up for latest `sass@1.34.0`:
https://sass-lang.com/d/slash-div

```
DEPRECATION WARNING: Using / for division is deprecated and will be removed in Dart Sass 2.0.0.

Recommendation: math.div($button-shadow-size, 2)

More info and automated migrator: https://sass-lang.com/d/slash-div

   ╷
35 │     padding: (govuk-spacing(2) - $govuk-border-width-form-element) govuk-spacing(2) (govuk-spacing(2) - $govuk-border-width-form-element - ($button-shadow-size / 2)); // s1
   │                                                                                                                                             ^^^^^^^^^^^^^^^^^^^^^^^
   ╵
    node_modules/govuk-frontend/govuk/components/button/_index.scss 35:141  @content
    node_modules/govuk-frontend/govuk/tools/_exports.scss 29:5              govuk-exports()
    node_modules/govuk-frontend/govuk/components/button/_index.scss 1:1     @import
    node_modules/govuk-frontend/govuk/components/_all.scss 6:9              @import
    node_modules/govuk-frontend/govuk/all.scss 6:9                          @import
```

With the "fix" being a breaking change for older versions of Sass, we might need to look into running the Sass migrator before we import any `*.scss` files: https://github.com/sass/migrator#readme

## Steps to reproduce the issue
Importing any GOV.UK Frontend `*.scss` files using `/` for division with `sass@1.33.0` and up.

See release notes: https://github.com/sass/dart-sass/releases/tag/1.33.0

36degrees:
Hey @colinrotherham, long time no see! Thanks for raising this.

This is a tricky one, as [we currently support LibSass and Ruby Sass alongside Dart Sass](https://frontend.design-system.service.gov.uk/installing-with-npm/#requirements), and `math.div` (as part of the wider module system) is not available in either. As such, I can't see any way to update the code to work with Dart Sass 2.0.0 (and to silence the deprecation warnings) without also removing support for compilers other than Dart Sass.

Ruby Sass has been deprecated for a while and I don't believe sees much use nowadays, but we wouldn't want to break support for LibSass without giving plenty of notice, and we'd need to treat it as a breaking change.

[...]

PR #3192, opened 2023-01-20
Fix or suppress Dart Sass deprecation warnings

colinrotherham:
This PR fixes or suppresses all deprecation warnings that show up in Dart Sass

For forwards compatibility (LibSass, Node Sass, Dart Sass) not all deprecations can be resolved

### $string: red is not a string.

Dart Sass will throw when trying to `quote()` colour identifiers as it no longer treats them as unquoted strings. We can use interpolation `"#{$colour}"` so Dart Sass takes `red` (color) as `"red"` (string)

See “Heads up!” box on Strings (Unquoted) docs:
https://sass-lang.com/documentation/values/strings#unquoted

```console
Error: $string: red is not a string.
   ╷
39 │   $colour: quote($colour);
   │            ^^^^^^^^^^^^^^
   ╵
    helpers/_colour.scss 39:12  govuk-colour()
    - 13:16                     root stylesheet]
```

### $weight: Passing a number without unit % (30) is deprecated.

The fix? Add units where needed

```console
Deprecation Warning: $weight: Passing a number without unit % (30) is deprecated.

To preserve current behavior: $weight * 1%

More info: https://sass-lang.com/d/function-units

   ╷
83 │   @return mix(#000000, $colour, $percentage);
   │           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   ╵
    src/govuk/helpers/_colour.scss 83:11               govuk-shade()
    src/govuk/components/tag/_index.scss 38:12         @content
    src/govuk/tools/_exports.scss 29:5                 govuk-exports()
    src/govuk/components/tag/_index.scss 1:1           @import
    src/govuk/components/phase-banner/_index.scss 1:9  @import
    src/govuk/components/_all.scss 26:9                @import
    src/govuk/all.scss 6:9                             @import
    src/govuk/all-ie8.scss 6:9                         @import
    app/assets/scss/app-ie8.scss 1:9                   root stylesheet
```

## Deprecations suppressed

This PR previously recreated [`math.div()`](https://sass-lang.com/documentation/modules/math#other-functions) ([see implementation](https://github.com/alphagov/govuk-frontend/compare/main...dart-sass-fix-deprecations-division)) but we've now suppressed warnings for:

1. Using / for division is deprecated and will be removed in Dart Sass 2.0.0.
2. Using / for division outside of calc() is deprecated and will be removed in Dart Sass 2.0.0.

These suppressions have been moved to:

* https://github.com/alphagov/govuk-frontend/pull/3164#pullrequestreview-1297214229

### DIFF
--- a/tasks/compile-stylesheets.mjs
+++ b/tasks/compile-stylesheets.mjs
@@ -66,6 +66,9 @@ export async function compileStylesheet ([modulePath, { srcPath, destPath }]) {
       file: moduleSrcPath,
       outFile: moduleDestPath,
 
+      // Turn off dependency warnings
+      quietDeps: true,
+
       // Enable source maps
       sourceMap: true,
       sourceMapContents: true,
--- a/src/govuk/components/tag/_index.scss
+++ b/src/govuk/components/tag/_index.scss
@@ -35,47 +35,47 @@
   }
 
   .govuk-tag--grey {
-    color: govuk-shade(govuk-colour("dark-grey", $legacy: "grey-1"), 30);
-    background: govuk-tint(govuk-colour("dark-grey", $legacy: "grey-1"), 90);
+    color: govuk-shade(govuk-colour("dark-grey", $legacy: "grey-1"), 30%);
+    background: govuk-tint(govuk-colour("dark-grey", $legacy: "grey-1"), 90%);
   }
 
   .govuk-tag--purple {
-    color: govuk-shade(govuk-colour("purple"), 20);
-    background: govuk-tint(govuk-colour("purple"), 80);
+    color: govuk-shade(govuk-colour("purple"), 20%);
+    background: govuk-tint(govuk-colour("purple"), 80%);
   }
 
   .govuk-tag--turquoise {
-    color: govuk-shade(govuk-colour("turquoise"), 60);
-    background: govuk-tint(govuk-colour("turquoise"), 70);
+    color: govuk-shade(govuk-colour("turquoise"), 60%);
+    background: govuk-tint(govuk-colour("turquoise"), 70%);
   }
 
   .govuk-tag--blue {
-    color: govuk-shade(govuk-colour("blue"), 30);
-    background: govuk-tint(govuk-colour("blue"), 80);
+    color: govuk-shade(govuk-colour("blue"), 30%);
+    background: govuk-tint(govuk-colour("blue"), 80%);
   }
 
   .govuk-tag--yellow {
-    color: govuk-shade(govuk-colour("yellow"), 65);
-    background: govuk-tint(govuk-colour("yellow"), 75);
+    color: govuk-shade(govuk-colour("yellow"), 65%);
+    background: govuk-tint(govuk-colour("yellow"), 75%);
   }
 
   .govuk-tag--orange {
-    color: govuk-shade(govuk-colour("orange"), 55);
-    background: govuk-tint(govuk-colour("orange"), 70);
+    color: govuk-shade(govuk-colour("orange"), 55%);
+    background: govuk-tint(govuk-colour("orange"), 70%);
   }
 
   .govuk-tag--red {
-    color: govuk-shade(govuk-colour("red"), 30);
-    background: govuk-tint(govuk-colour("red"), 80);
+    color: govuk-shade(govuk-colour("red"), 30%);
+    background: govuk-tint(govuk-colour("red"), 80%);
   }
 
   .govuk-tag--pink {
-    color: govuk-shade(govuk-colour("pink"), 40);
-    background: govuk-tint(govuk-colour("pink"), 80);
+    color: govuk-shade(govuk-colour("pink"), 40%);
+    background: govuk-tint(govuk-colour("pink"), 80%);
   }
 
   .govuk-tag--green {
-    color: govuk-shade(govuk-colour("green"), 20);
-    background: govuk-tint(govuk-colour("green"), 80);
+    color: govuk-shade(govuk-colour("green"), 20%);
+    background: govuk-tint(govuk-colour("green"), 80%);
   }
 }
--- a/src/govuk/helpers/_colour.scss
+++ b/src/govuk/helpers/_colour.scss
@@ -8,7 +8,7 @@
 
 /// Get colour
 ///
-/// @param {String} $colour - Name of colour from the colour palette
+/// @param {String | Colour} $colour - Name of colour from the colour palette
 ///   (`$govuk-colours`)
 /// @param {String} $legacy - Either the name of colour from the legacy palette
 ///   or a colour literal, to return instead when in 'legacy mode' - matching
@@ -36,7 +36,10 @@
     $colour: $legacy;
   }
 
-  $colour: quote($colour);
+  @if type-of($colour) == "color" {
+    // stylelint-disable-next-line scss/function-quote-no-quoted-strings-inside
+    $colour: quote("#{$colour}");
+  }
 
   @if not map-has-key($govuk-colours, $colour) {
     @error "Unknown colour `#{$colour}`";


lib/jest-helpers.js
```diff
@@ -142,6 +142,7 @@ async function getExamples (componentName) {
 function renderSass (options) {
   return sassRender({
     includePaths: sassPaths,
+    quietDeps: true,
```
colinrotherham, line 145: Whilst this PR fixes `govuk-frontend` deprecation warnings we add `quietDeps: true` to ignore:

```console
govuk_frontend_toolkit
govuk-elements-sass
```

See https://frontend.design-system.service.gov.uk/importing-css-assets-and-javascript/#silence-deprecation-warnings-from-dependencies-in-dart-sass

36degrees:
This is a really neat solution, but I'm still hesitant about this change because:

- we already have [a solution that silences the deprecation warnings for our users](https://frontend.design-system.service.gov.uk/importing-css-assets-and-javascript/#silence-deprecation-warnings-from-dependencies-in-dart-sass) which follows the approach [recommended by Sass](https://sass-lang.com/documentation/breaking-changes/slash-div#transition-period) for users of libraries like ours
- I expect that Dart Sass 2.0.0 (when the deprecated features are due to be removed) will also drop support for `@import`, which will force us to drop support for LibSass anyway – at which point having our own version of `math.div` wouldn't make any difference – though I haven't seen this documented anywhere so this is an assumption on my part
- I also expect there will be further deprecations before 2.0 is released, and I'm just not sure it's a good use of our time trying to work around them whilst we continue to support other Sass compilers

Would be keen to get other's thoughts on this. Is there a benefit to doing this that I'm missing?

colinrotherham:
Thanks @36degrees, appreciate this one's not crucial

The real good stuff like the Sass modern compilation API for 2.0 is in this PR instead:

* https://github.com/alphagov/govuk-frontend/pull/3164

This one does gives us a nice way to start using Dart Sass in spikes without hiding those warnings away. So when new deprecations arrive like **Passing a number without unit % is deprecated** we don't miss them in the terminal

Alternatively `{ quietDeps: true }` works but we'd have to move `govuk-frontend` to a "dep" (dependency)

## Before
Review app imports by resolving local files

```
@import "../../../src/govuk/all";
```

## After
Review app imports via `loadPaths` + `{ quietDeps: true }`

```
@import "govuk-frontend/govuk/all";
```

colinrotherham:
This bit from the [`quietDeps` documentation](https://sass-lang.com/documentation/js-api/interfaces/LegacyStringOptions#quietDeps) is good:

>This is useful for silencing deprecation warnings that you can’t fix on your own. However, please also notify your dependencies of the deprecations so that they can get fixed as soon as possible!

If we prefer `{ quietDeps: true }` should we try and surface them to ourselves another way instead?

romaricpascal:
> Would be keen to get other's thoughts on this. Is there a benefit to doing this that I'm missing?

Have teams reported those deprecation warnings being an issue or is it more for our own peace of mind/standard that we're trying to tidy them up? If the later (either because they don't care about the flurry or warning, or ignore them with `quietDeps`), I think we can avoid having a division workaround in our codebase and wait for adopting the new version of Sass/the warnings being enough of an issue.

If we were to add a division workaround, though, I think it should be unit tested, as it'd be something we'd easily reach for when building further features. I originally thought it would be worth making it public as well (should we have it), but projects where the warning would appear would us a Sass version that also offer the `math.div` function as well for teams to use in their own code.

The implementation also looks [much more complex than Bootstrap's](https://github.com/twbs/bootstrap/blob/main/scss/_functions.scss#L262). What's the reason for that extra complexity (esp. the units bit and separation into two functions)?

colinrotherham:
>Have teams reported those deprecation warnings being an issue or is it more for our own peace of mind/standard that we're trying to tidy them up?

@romaricpascal If you scroll down on this issue it's referenced fairly frequently:

* https://github.com/alphagov/govuk-frontend/issues/2238

### Deprecation warnings
Assuming we're going ahead with Dart Sass, we have three options:

1. **Log all warnings**
We'll see warnings our own Sass stylesheets but also dependencies via `loadPaths`

    ```scss
      // via `loadPaths`
      @import "govuk_frontend_toolkit/stylesheets/*";
      @import "govuk-elements-sass/public/sass/govuk-elements";
      
      // via local path resolution
      @import "../../../src/govuk/all";
    ```

2. **Log warnings for our stylesheets only** via [`{ quietDeps }`](https://sass-lang.com/documentation/js-api/interfaces/Options#quietDeps)
We'll see warnings for only our own Sass stylesheets

    ```console
      @import "../../../src/govuk/all";
    ```

2. **Silence all warnings** via [`{ quietDeps }`](https://sass-lang.com/documentation/js-api/interfaces/Options#quietDeps)
By replacing `@import "../../../src/govuk/all"` with `@import "govuk-frontend/govuk/all"`

colinrotherham:
>The implementation also looks [much more complex than Bootstrap's](https://github.com/twbs/bootstrap/blob/main/scss/_functions.scss#L262). Are we doing something they don't (esp. the units bit and separation into two functions)?

@romaricpascal I hadn't actually seen Bootstrap's version when I pushed this PR up, but we're both using the same type of [long division](https://en.wikipedia.org/wiki/Long_division) to count the number of subtractions 😊 

### Things they've done differently

1. Returning `0` early if the first number is zero
2. Throwing when trying to divide by `0`
3. Manually rounding the last digit in `66.66666%` → **`66.66667%`**

They've used this little snippet:

```scss
@if ($precision < 0 and $remainder >= $divisor * 5) {
  $result: $result + 1;
}
```

### Things I've done differently

1. Grabbed the remainder using [`%` modulo operator](https://en.wikipedia.org/wiki/Modulo)
2. Shifted the decimal remainder by 100 (not 10) to reduce loops
4. Automatically letting Sass round `66.66666%` → **`66.66667%`**

Could definitely combine my two functions into one if preferred

Or leave it out entirely!

36degrees:
I think we should fix the '$weight: Passing a number without unit % (30) is deprecated.' warnings now, but I don't think we should be trying to reimplement `math.div`.

My preference would be to go with option 2, but also see if there's a way to filter out the 'known' warnings – so we don't see warnings for `math.div` for example, but if we upgrade dart-sass and a new deprecation warning is introduced, we'd see it (and then either fix it or silence the new warning too).

tasks/compile-stylesheets.mjs
```diff
@@ -66,6 +66,9 @@ export async function compileStylesheet ([modulePath, { srcPath, destPath }]) {
       file: moduleSrcPath,
       outFile: moduleDestPath,
 
+      // Turn off dependency warnings
+      quietDeps: true,
```
colinrotherham, line 70: Whilst this PR fixes `govuk-frontend` deprecation warnings we add `quietDeps: true` to ignore:

```console
govuk_frontend_toolkit
govuk-elements-sass
```

See https://frontend.design-system.service.gov.uk/importing-css-assets-and-javascript/#silence-deprecation-warnings-from-dependencies-in-dart-sass

colinrotherham:
> I think we should fix the '$weight: Passing a number without unit % (30) is deprecated.' warnings now, but I don't think we should be trying to reimplement `math.div`.

@36degrees Updated the description to say we've fixed these two:

1. **$string: red is not a string.**
2. **$weight: Passing a number without unit % (30) is deprecated.**

> My preference would be to go with option 2, but also see if there's a way to filter out the 'known' warnings – so we don't see warnings for `math.div` for example, but if we upgrade dart-sass and a new deprecation warning is introduced, we'd see it (and then either fix it or silence the new warning too).

Added your suggestion in this PR:

* https://github.com/alphagov/govuk-frontend/pull/3164#pullrequestreview-1297214229

We needed access to the Dart Sass modern compilation API [`logger` property](https://sass-lang.com/blog/request-for-comments-new-js-api#loggers) so depends on:

* https://github.com/alphagov/govuk-frontend/pull/3258

domoscargin (commented):
These all feel like non-controversial changes now (having swooped in after everybody else dealt with the nitty gritty). 

I'm happy to approve, but appreciate @36degrees and @romaricpascal might want to weigh in?

romaricpascal (approved):
Changes look sound to me! 🚀

domoscargin (approved):

PR #3198, opened 2023-01-23
Use Prettier to check `*.{json,md,yaml,yml}` documents

colinrotherham:
This PR adds Prettier as a formatter for `*.{json,md,yaml,yml}` documents to fix and format everything automatically

**Note:** There's a plugin to [format Jinja](https://github.com/davidodenwald/prettier-plugin-jinja-template#readme) (+ Nunjucks) used by Ministry of Justice that we could add in future?

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,7 +1,6 @@
 version: 2
 
 updates:
-
   # Update npm packages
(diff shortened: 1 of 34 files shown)


colinrotherham:
Switched branches and it auto merged, will open again

PR #3201, opened 2023-01-23
Use Prettier to check `*.{json,md,yaml,yml}` documents

colinrotherham:
This PR adds Prettier as a formatter for `*.{json,md,yaml,yml}` documents to fix and format everything automatically

**Note:** There's a plugin to [format Jinja](https://github.com/davidodenwald/prettier-plugin-jinja-template#readme) (+ Nunjucks) used by Ministry of Justice that we could add in future?


docs/releasing/publishing.md
```diff
@@ -15,8 +15,9 @@ Developers should pair on releases. When remote working, it can be useful to be
 4. Create and check out a new branch (`release-[version-number]`). See the [versioning documentation](/docs/contributing/versioning.md) for more information.
 
 5. Update the [`CHANGELOG.md`](../../CHANGELOG.md) by:
-  - changing the 'Unreleased' heading to the new version-number and release-type - for example, '3.11.0 (Feature release)'
-  - adding a new 'Unreleased' heading above the new version-number and release-type, so users will know where to add PRs to the changelog
+
+- changing the 'Unreleased' heading to the new version-number and release-type - for example, '3.11.0 (Feature release)'
+- adding a new 'Unreleased' heading above the new version-number and release-type, so users will know where to add PRs to the changelog
```
romaricpascal, line 20: I think those are meant to be part of the 5. item:

```suggestion
   - changing the 'Unreleased' heading to the new version-number and release-type - for example, '3.11.0 (Feature release)'
   - adding a new 'Unreleased' heading above the new version-number and release-type, so users will know where to add PRs to the changelog
```

docs/releasing/publishing.md
```diff
@@ -43,17 +44,17 @@ Developers should pair on releases. When remote working, it can be useful to be
 
 3. Run `npm run publish-release`, which will prompt you to check whether the npm tag looks as expected.
 
-  If you're following these instructions, you're probably doing a sequential release, meaning
-  the tag should be 'latest'.
+If you're following these instructions, you're probably doing a sequential release, meaning
+the tag should be 'latest'.
```
romaricpascal, line 48: And those part of 3.

```suggestion
   If you're following these instructions, you're probably doing a sequential release, meaning the tag should be 'latest'.
```

docs/releasing/publishing.md
```diff
@@ -43,17 +44,17 @@ Developers should pair on releases. When remote working, it can be useful to be
...
 
 5. You will now be prompted to continue or cancel the release. Check the details and enter `y` to continue. If something does not look right, press `N` to cancel the release.
-  
-  This step will create a ZIP file containing the release in the root of your govuk-frontend git directory. You will need this file when creating the GitHub release.
 
-  It will also automatically create a tag in Github which you can use to create a Github release in the following section.
-  
+This step will create a ZIP file containing the release in the root of your govuk-frontend git directory. You will need this file when creating the GitHub release.
+
+It will also automatically create a tag in Github which you can use to create a Github release in the following section.
```
romaricpascal, line 56:
```suggestion
   This step will create a ZIP file containing the release in the root of your govuk-frontend git directory. You will need this file when creating the GitHub release.

   It will also automatically create a tag in Github which you can use to create a Github release in the following section.
```

romaricpascal (commented):
Looking so tidy 🤩 Spotted a few space changes in Markdown: list content that wasn't indented enough for Prettier to figure they were part of the list item.

In the YAML, I noticed this got rid of the following pattern where items in a list would start with just an empty `-`. 

```yaml
items:
  -
     key: 'first'
     other_key: 'other_value'
  -
     key: 'second'
```

Might be worth a quick check for some history behind it, though I'm happy to let Prettier just auto-format it and us not worry about it anymore 🥳

romaricpascal:
Another thought regarding YAML and Markdown linting. Do you think it'd be worth setting up some proper linting with [yaml-lint](https://www.npmjs.com/package/yaml-lint) for YAML, and [remark-lint](https://github.com/remarkjs/remark-lint) or [markdownlint](https://github.com/DavidAnson/markdownlint) for markdown (I'm partial to the first for its ecosystem of plugins, but won't die on that hill). This may help catch more than formatting errors (and may also do the formatting as well).

colinrotherham:
Thanks @romaricpascal, appreciate it. I'll take a close look at the changes 😊 

> Another thought regarding YAML and Markdown linting. Do you think it'd be worth setting up some proper linting with [yaml-lint](https://www.npmjs.com/package/yaml-lint) for YAML, and [remark-lint](https://github.com/remarkjs/remark-lint) or [markdownlint](https://github.com/DavidAnson/markdownlint) for markdown (I'm partial to the first for its ecosystem of plugins, but won't die on that hill). This may help catch more than formatting errors (and may also do the formatting as well).

I've left it as formatting-only because we largely convert `yaml` → JS → `json` for the Design System

Which gives us a bit of a type fight:
https://github.com/alphagov/govuk-frontend/blob/main/lib/file-helper.js#L185

**Question:** Do we still need YAML—could component data be exported from JS instead?

Otherwise we've got ESLint plugins available for both JSON and YAML

https://github.com/azeemba/eslint-plugin-json#readme
https://github.com/kellyselden/eslint-plugin-json-files#readme
https://github.com/ota-meshi/eslint-plugin-yml#readme

docs/releasing/publishing.md
colinrotherham: Thanks for this, fixed the bullet indenting 👍

docs/releasing/publishing.md
colinrotherham: Sorted 👍

docs/releasing/publishing.md
colinrotherham: Sorted 👍

colinrotherham:
Sadly Prettier doesn't support `.gitignore` (only `.prettierignore`)

* https://github.com/prettier/prettier/issues/8048
* https://github.com/prettier/prettier/issues/8506

So we're back to nearly duplicating `.gitignore` again 🙈

.prettierrc.js
```diff
@@ -0,0 +1,20 @@
...
+        embeddedLanguageFormatting: 'off',
+        singleQuote: false
+      }
+    },
+    {
+      files: '*.scss',
+      options: {
+        singleQuote: false
+      }
+    }
```
36degrees, line 18: If we're only running prettier against `json`, `md`, `yaml` and `yml` files will this override ever be applied?

colinrotherham: For context, this is for compatibility

Text editors and IDEs also pick up the Prettier config

Should anyone want to turn on auto formatting (by choice) this stops Sass quotes being flipped to "single"

36degrees (approved):

PR #3204, opened 2023-01-24
Use consistent ignore patterns for Git, ESLint and Stylelint

colinrotherham:
This PR includes:

1. Remove unused `.gitignore` entries
2. Ignore pattern review for Git, ESLint and Stylelint
3. Ignore “build-but-committed” directories such as **./dist** and **./package**

We're making these changes because committing `*.scss` files into **./package** throws:

```console
✖ npm run lint:scss:cli -- --fix:
Error: All input files were ignored because of the ignore pattern. Either change your input, ignore pattern or use "--allow-empty-input" to allow no inputs
```

We've fixed this by adding the `--allow-empty-input` flag to skip already-ignored files

This was a lot easier than the fix for ESLint:

* https://github.com/alphagov/govuk-frontend/pull/3186

Which we've now consolidated into `filterTask()` should other linters need to ignore files in future

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,14 +1,23 @@
+# Temporary only
+.cache/
+.tmp/
+
+# Node.js modules
 node_modules/
-npm-debug.log
-.DS_Store
-preview/
-tmp/
-axeReports/
+
+# Test coverage
 coverage/
-public/
-package/package-lock.json
-examples/**/package-lock.json
-*.zip
+
+# Build output
 jsdoc/
+public/
 sassdoc/
-.cache
+
+# Files to ignore
+.DS_Store
+*.zip
+*.log
+
+# Project lockfile only
+package-lock.json
+!/package-lock.json
--- a/.lintstagedrc.js
+++ b/.lintstagedrc.js
@@ -12,7 +12,7 @@ module.exports = {
   //
   // https://github.com/okonet/lint-staged#how-can-i-ignore-files-from-eslintignore
   '*.{cjs,js,mjs}': filterTask('npm run lint:js:cli -- --fix'),
-  '*.scss': filterTask('npm run lint:scss:cli -- --fix')
+  '*.scss': 'npm run lint:scss:cli -- --fix --allow-empty-input'
 }
 
 // Configure paths to ignore
--- a/.lintstagedrc.js
+++ b/.lintstagedrc.js
@@ -11,27 +11,32 @@ module.exports = {
   // as recommended by lint-staged.
   //
   // https://github.com/okonet/lint-staged#how-can-i-ignore-files-from-eslintignore
-  '*.{cjs,js,mjs}': async (files) => {
-    const filesToLint = await removeESLintIgnoredFiles(files)
-    return [`npm run lint:js:cli -- --fix ${filesToLint}`]
-  },
-  '*.scss': ['npm run lint:scss:cli -- --fix']
+  '*.{cjs,js,mjs}': filterTask('npm run lint:js:cli -- --fix'),
+  '*.scss': filterTask('npm run lint:scss:cli -- --fix')
 }
+
+// Configure paths to ignore
+const eslint = new ESLint()
 
 /**
  * Removes files ignored by ESLint from a list of files provided by lint-staged
  *
- * @param {Array<string>} files - The list of files lint-staged wants to lint
- * @returns {Array<string>} - The list without any of the files ignored by ESLint
+ * @param {string} task - The task `lint-staged` wants to execute
+ * @returns {Promise<function(string[]): string[]>} Tasks to run with files argument
  */
-async function removeESLintIgnoredFiles (files) {
-  const eslint = new ESLint()
-  const isIgnored = await Promise.all(
-    files.map((file) => {
-      return eslint.isPathIgnored(file)
-    })
-  )
-  const filteredFiles = files.filter((_, i) => !isIgnored[i])
-  // Wrap file in quotes in case it contains a space
-  return filteredFiles.map(file => `"${file}"`).join(' ')
+function filterTask (task) {
+  return async (files) => {
+    const isIgnored = await Promise.all(
+      files.map((file) => eslint.isPathIgnored(file))
+    )
+
+    // Wrap files in quotes in case they contains a space
+    const paths = files
+      .filter((_, i) => !isIgnored[i])
+      .map(file => `"${file}"`)
+
+    return paths.length
+      ? [`${task} ${paths.join(' ')}`]
+      : []
+  }
 }
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,14 +1,15 @@
 module.exports = {
   extends: 'standard',
   ignorePatterns: [
+    '**/dist/**',
+    '**/package/**',
+    '**/public/**',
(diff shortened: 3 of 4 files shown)


romaricpascal:
I've got a few concerns about this solution to the issue:

- re-using the `.gitignore` ensures we don't forget to move a value from one of the list to the other. Duplicating it in a separate list feel like a recipe for forgetting to port one of the entry to `.lintignore`. If we really want to factor commited-but-to-be-ignored files, we could always use a JS module to share them across both config files, but imho, it's fine that there's a little separation between the two tools.
- ESLint is used for filtering files linted at commit both for Stylelint and ESLint. Would that risk missing a file because a nested `.eslintrc.js` adds extra ignore patterns? It also feels a risk for the future in case the tools filtering diverges.

The error message makes it sound like Stylelint has a flag that would let us not worry about it trying to run on ignored files. Would it be fine to just use that instead rather than take more responsibilities on our side (it's a shame ESLint doesn't provide the same :'( )

colinrotherham:
@romaricpascal Shall I stick with a separate ignore file for each tool?

Will have the same problem with Prettier and EditorConfig

Not many repos commit their compiled code so I'm open to other ideas! 💡

romaricpascal:
Fair point about other linters, hadn't thought of that. 

As it's really a "commit-on-linting" concern, how about we keep it to `.lintstagedrc.js` with an explicit list of locations that shouldn't get linted on commit (`dist`, `package` and `src/vendor`), filtered with our own globbing rather than tie it to the linter's own configurations? 

That'd keep each concern (linting language X, linting on commit) isolated and feels a little safer to not tangle them all together?

romaricpascal:
And I realise I didn't answer your first question, but yeah, what I described would keep each linter in their current configuration (load `.gitignore` + have a couple of patterns in their config).

romaricpascal:
Oh sorry, I didn't mean to have a `.gitignore` + a `.<name-of-the-tool>ignore` file, but to follow the current convention of using `.gitignore` as the tool's ignore file and having extra paths in the `.<name-of-the-tool>rc.js` or `.<name-of-the-tool>.config.js` file.

colinrotherham:
> Oh sorry, I didn't mean to have a `.gitignore` + a `.<name-of-the-tool>ignore` file, but to follow the current convention of using `.gitignore` as the tool's ignore file and having extra paths in the `.<name-of-the-tool>rc.js` or `.<name-of-the-tool>.config.js` file.

Sorted 👍 

Prettier can only have "ignores" in `.prettierignore` but that's for another PR

.gitignore
```diff
@@ -1,14 +1,17 @@
...
-npm-debug.log
-.DS_Store
-preview/
-tmp/
-axeReports/
+
+# Test coverage
 coverage/
-public/
-package/package-lock.json
```
romaricpascal, line 9: **issue** I think we'd want to keep that as a safety for someone inadvertently running `npm install` inside `package` and then commiting it.

.gitignore
```diff
@@ -1,14 +1,17 @@
...
-examples/**/package-lock.json
-*.zip
```
romaricpascal, line 11: **issue** I think that one protects us from commiting the built archive we upload to Github releases. So we'd want to keep it.

.gitignore
```diff
@@ -1,14 +1,17 @@
+# Temporary cache
+.cache/
+
+# Node.js modules
 node_modules/
-npm-debug.log
-.DS_Store
-preview/
-tmp/
```
romaricpascal, line 5: **suggestion** It's a nice thing to know that you can use the `tmp` folder at the root of the project without worry. I'd be keen to keep it.

.lintstagedrc.js
```diff
@@ -11,27 +11,32 @@ module.exports = {
...
 
+// Configure paths to ignore
+const eslint = new ESLint()
+
 /**
  * Removes files ignored by ESLint from a list of files provided by lint-staged
  *
- * @param {Array<string>} files - The list of files lint-staged wants to lint
- * @returns {Array<string>} - The list without any of the files ignored by ESLint
+ * @param {string} task - The task `lint-staged` wants to lint
```
romaricpascal, line 24:
```suggestion
 * @param {string} task - The task `lint-staged` wants to execute
```

.lintstagedrc.js
```diff
@@ -11,27 +11,32 @@ module.exports = {
   // as recommended by lint-staged.
   //
   // https://github.com/okonet/lint-staged#how-can-i-ignore-files-from-eslintignore
-  '*.{cjs,js,mjs}': async (files) => {
-    const filesToLint = await removeESLintIgnoredFiles(files)
-    return [`npm run lint:js:cli -- --fix ${filesToLint}`]
-  },
-  '*.scss': ['npm run lint:scss:cli -- --fix']
+  '*.{cjs,js,mjs}': filterTask('npm run lint:js:cli -- --fix'),
+  '*.scss': 'npm run lint:scss:cli -- --fix'
```
romaricpascal, line 15: Shouldn't this set the `allow-empty-input` flag that would fix the error the PR was set to solve? Happy to have that flag as part ot `lint:scss:cli` so we don't have to think about it going forward :) 

```suggestion
  '*.scss': 'npm run lint:scss:cli -- --fix --allow-empty-input'
```

romaricpascal (commented):
The changes make sense, but I think the code stopped doing what it was set out to do as there's nothing filtering what gets ran through stylelint, nor extra argument passed to it. 🤔 

I've also noticed a few entries in the `.gitignore` file that got lost and we may want to keep 😊

.gitignore
colinrotherham: Yeah let's keep it if anyone's using it otherwise let's get rid

If you'd rather keep it shall we use `.tmp` so it stays obvious at the top (A-Z) along with `.cache`?

romaricpascal: Good idea, let's go with the leading dot 😄

.gitignore
colinrotherham: Would you mind a different approach?

I'm aware we've got lots of other npm workspace directories now

For safety, there are likely lots of other files we've missed:
```
app/package.json
app/package-lock.json
app/npm-debug.log

docs/examples/webpack/package.json
docs/examples/webpack/package-lock.json
docs/examples/webpack/npm-debug.log

package/package.json
package/package-lock.json
package/npm-debug.json
```

Perhaps if we want to stop accidents everywhere:

```.gitignore
.cache
.tmp
*.zip
*.log

package-lock.json
!/package-lock.json
```

.lintstagedrc.js
colinrotherham: Well spotted, I took this out when I ran it through `filterTask()`

But that's gone now

.gitignore
romaricpascal: Yeah, that's a neat solution for our growing number of workspaces.

.gitignore
colinrotherham: Ah don't worry, I'll add it back

Mind if I make it self-documenting?

```.gitignore
release-*.zip
```

romaricpascal:
Might be worth renaming the PR, though as it's now mostly fixing the Stylelint issue rather than setting a common ignore list 😆

PR #3209, opened 2023-01-25
Tie Exit this Page skiplink into ghost page functionality

querkmachine:
Adds a little JS so that skiplinks are tied into the EtP component's `exitPage` function, allowing the skiplink to activate the 'ghost page' function. `exitPage` has been updated so that it redirects to the href of the activated link, rather than being hardcoded to the href of the main EtP button.

Keyboard-based functionality operates on a page-wide basis, and thus does not need to be separately enabled for the skiplink. This leaves only the 'ghost page' overlay needing to work across both components. 

## Issues
- Writing Jest tests that play nicely with the `syncOverlayVisibility` function in our JS is proving elusive. This might be because we use `pageshow` as one of the event hooks for it, but this is never triggered when the tests are ran headlessly. 
  - That said, the test still fails when Jest/Puppeteer is ran in headed mode.
  - This uses the same event hooks as the functionality that restores the state of conditional checkboxes and radio buttons, which also lacks Jest tests. Maybe this is just a no go?

### DIFF
--- a/src/govuk/components/exit-this-page/exit-this-page.test.js
+++ b/src/govuk/components/exit-this-page/exit-this-page.test.js
@@ -116,19 +116,5 @@ describe('/components/exit-this-page', () => {
       const ghostOverlay = await page.evaluate((overlayClass) => document.body.querySelector(overlayClass), overlayClass)
       expect(ghostOverlay).not.toBeNull()
     })
-
-    it('hides the ghost page when the user navigates back in history', async () => {
-      await goToExample(page, 'exit-this-page-with-skiplink')
-
-      await Promise.all([
-        page.waitForNavigation(),
-        page.click(buttonClass)
-      ])
-
-      await page.goBack()
-
-      const ghostOverlay = await page.evaluate((overlayClass) => document.body.querySelector(overlayClass), overlayClass)
-      expect(ghostOverlay).toBeNull()
-    })
   })
 })
--- a/src/govuk/components/exit-this-page/exit-this-page.test.js
+++ b/src/govuk/components/exit-this-page/exit-this-page.test.js
@@ -116,5 +116,19 @@ describe('/components/exit-this-page', () => {
       const ghostOverlay = await page.evaluate((overlayClass) => document.body.querySelector(overlayClass), overlayClass)
       expect(ghostOverlay).not.toBeNull()
     })
+
+    it('hides the ghost page when the user navigates back in history', async () => {
+      await goToExample(page, 'exit-this-page-with-skiplink')
+
+      await Promise.all([
+        page.waitForNavigation(),
+        page.click(buttonClass)
+      ])
+
+      await page.goBack()
+
+      const ghostOverlay = await page.evaluate((overlayClass) => document.body.querySelector(overlayClass), overlayClass)
+      expect(ghostOverlay).toBeNull()
+    })
   })
 })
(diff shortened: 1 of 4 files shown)


querkmachine:
I _think_ this is good for a review now. 

The lack of any integration test for `syncOverlayVisibility` isn't great, but it doesn't seem like it's practical to write one.

I'm also not 100% on whether this approach is 'the best one', but it does fail safely in non-ideal conditions:
- If JavaScript is not available, both the EtP button and skiplink operate as normal links.
- If the EtP button is present and the skiplink isn't, the EtP button works as expected and no errors are returned from the skiplink being absent.
- If the skiplink is present and the EtP button isn't, the EtP JS doesn't run and the skiplink operates as a normal link.
- If the skiplink is missing the required class name, the skiplink operates as a normal link.

The use of a class name for binding JS to elements is according to the [JS authoring documentation](https://github.com/alphagov/govuk-frontend/blob/main/docs/contributing/coding-standards/js.md#use-classes-to-target-dom-elements). 

I had considered whether we'd want to change or supplement this with a `data-*` attribute so that extra information could be passed into it (i.e. configuration options to toggle features on/off), but as the ghost page is the only feature being extended currently, that felt like a premature development.

owenatgov (approved):
Great stuff 💪🏻 

I envisioned this working where the skip link was it's own initialised EtP instance and the init step would detect the context of what's initialising it, however I wonder if we necessarily need that. This way means we're asking teams less to get it working and encourages teams to implement EtP as a complete page pattern. There's the risk that the js could fail but the skip link still works since it's just a link!

The test gaff is a bummer and I think we should try to fix it but it feels like something we should try to understand at a wider level. I think you've done your level best here.

src/govuk/components/exit-this-page/exit-this-page.test.js
```diff
@@ -2,38 +2,133 @@
...
-      page.keyboard.press('Shift'),
-      page.keyboard.press('Shift'),
-      page.keyboard.press('Shift'),
-      page.waitForNavigation()
-    ])
+      await page.goBack()
 
-    const url = await page.url()
-    expect(url).toBe(href)
+      const ghostOverlay = await page.evaluate((overlayClass) => document.body.querySelector(overlayClass), overlayClass)
```
owenatgov, line 130: I'm not sure this has done the trick sadly. When I remove the contents of `syncOverlayVisibility` so that the overlay isn't removed this still reports `null`. I had a play around but I couldn't figure it out. It's returning the page successfully, just not retaining the overlay.

Don't feel that you have to solve this right now. I suspect this is something we need to figure out in our own testing infrastructure.

querkmachine: I think I have a new theory here: Puppeteer runs Chrome tests in incognito mode. Does that mode avoid using bfcache at all, always loaded anew rather than restoring the previous JS state? This would result in it passing the test without needing to run the actual function.

Alas I'm struggling to find anything concrete on this on t'interweb. If there are any weird interactions between private browsing and bfcache, it doesn't seem to be particularly well documented.

querkmachine:
We've agreed to exclude the state syncronisation integration test from this PR and to spin that out into it's own item of work.

This is because the same issues with testing we're seeing here exist on other components (particularly the checkbox and radio buttons), and it may require larger changes to our testing infrastructure that are well outside of the scope of this PR. 

As a result I've reverted the adding of the test. I chose to revert so that the test remains in source control in case we find use for it in the future, though I can rebase or squash it if desired.

PR #3221, opened 2023-01-30
Update screen reader announcements for exit this page

owenatgov:
## What/Why
Update exit this page to provide status updates to screen readers whilst interacting with the component. This is to ensure that exit this page complies with [WCAG 4.1.3: Status Messages](https://www.w3.org/WAI/WCAG22/Understanding/status-messages).

Closes https://github.com/alphagov/govuk-frontend/issues/3061

### Announcements
There are now announcements for the following interactions:

- Pressing the shift key. This announces "Exit this page key press x of 3" where x is the number of times the shift key has been pressed
- Pressing the shift key 3 times within 5 seconds in succession. This announces "Exit this page activated"
- Not pressing the shift key for 5 seconds after pressing it once or twice or pressing another key between shift key presses. This announces "Exit this page timed out"

The shift key announcement and timeout announcement will not activate if the shift key is pressed alongside another key eg: if you press shift+p to get a capitol P, in line with the intended functionality of the button.

## Testing
This has been tested against our standard testing matrix. [See the results here](https://docs.google.com/spreadsheets/d/1haNkUZ-iHHktVGuUegCuhFQUTHaxWwzuzS8JqndpD7w/edit?pli=1#gid=0).

### Move the announcement span out of the button
During testing, I found that on VoiceOver on safari that the updates from the announcement span  weren't being picked up. From investigating, this is because we were putting the span inside the link version of our button component, which was a link with `role="button"`. [ARIA forbids buttons or elements with button roles containing any meaningful accessible information](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role#all_descendants_are_presentational). Moving the announcement span to be a sibling of the button within the parent `govuk-exit-this-page` div fixes this.

### Update how we handle timeouts
We now store our timeouts (one for the window to press shift 3 times and another for clearing the timeout message) at the module level. This is so that we can clear the timeouts more easily if we want them to be interrupted. For example if the timeout message is announced and, before 5 seconds passes, the user presses shift again. This would mean we can prevent the timeout message timeout from ever completing, thus the span isn't unexpected cleared as a result of said timeout.

### Extend the overlay sync function to also reset the indication lights and timeouts
When building the ghost page feature we encountered a browser feature that would retain the state of the page, including the ghost page overlay, and so built a feature to manually remove the overlay on `pageload` if it was still there. I found during testing that the page would sometimes also retain the state of the announcement span, the indicator lights and the numerous timeouts. This adds onto the existing reset function to also reset those elements of the js.

## Analysis of testing
I'm generally happy with the state of the component following testing. I do however have some thoughts on the findings:

- The JAWS skipping an announcement is still weird and unclear. I haven't investigated this yet but I wondered if JAWS was, for some reason, registering 2 shift presses per single press. It may be worth expanding the number of required keypresses to verify this and then use that as a basis for investigation or raising a bug.
- I'm unsure how valuable the "exit this page activated" message is considering most screen readers defaulted to standard navigation behaviour. I may do some additional investigation with browser throttling.

### DIFF
--- a/src/govuk/components/exit-this-page/exit-this-page.mjs
+++ b/src/govuk/components/exit-this-page/exit-this-page.mjs
@@ -18,9 +18,14 @@ function ExitThisPage ($module) {
   this.$indicatorContainer = null
   this.$overlay = null
   this.escCounter = 0
-  this.escTimerActive = false
   this.lastKeyWasModified = false
-  this.timeout = 5000 // milliseconds
+  this.timeoutTime = 5000 // milliseconds
+
+  // Store the timeout events so that we can clear them to avoid user keypresses overlapping
+  // setTimeout returns an id that we can use to clear it with clearTimeout,
+  // hence the 'Id' suffix
+  this.keypressTimeoutId = null
+  this.timeoutMessageId = null
 }
 
 /**
@@ -28,10 +33,10 @@ function ExitThisPage ($module) {
  */
 ExitThisPage.prototype.initUpdateSpan = function () {
   this.$updateSpan = document.createElement('span')
-  this.$updateSpan.setAttribute('aria-live', 'polite')
-  this.$updateSpan.setAttribute('class', 'govuk-visually-hidden')
+  this.$updateSpan.setAttribute('role', 'status')
+  this.$updateSpan.className = 'govuk-visually-hidden'
 
-  this.$button.appendChild(this.$updateSpan)
+  this.$module.appendChild(this.$updateSpan)
 }
 
 /**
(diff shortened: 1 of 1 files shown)


src/govuk/components/exit-this-page/exit-this-page.mjs
```diff
@@ -18,20 +18,25 @@ function ExitThisPage ($module) {
...
 }
 
 /**
  * Create the <span> we use for screen reader announcements.
  */
 ExitThisPage.prototype.initUpdateSpan = function () {
   this.$updateSpan = document.createElement('span')
-  this.$updateSpan.setAttribute('aria-live', 'polite')
+  this.$updateSpan.setAttribute('role', 'status')
   this.$updateSpan.setAttribute('class', 'govuk-visually-hidden')
```
querkmachine, line 37: My nomination for the Award for the Most Nitpicky of Nitpickings: setting the `className` property is slightly more direct than changing the attribute, for those precious ten-thousands of a second. 

```suggestion
  this.$updateSpan.className = 'govuk-visually-hidden'
```

(`className` also seems to be the more common usage in our current JS, though we don't seem to have much consistency in that regard.) 

I look forward to the ceremony. 🏆

querkmachine (approved):
The code looks good to me! Nothing screaming red flags at me. 

I do also question how valuable the "exit this page activated" announcement is, as screen readers usually already conveys information about page navigation activities.

I also worry that, as the last announcement of the series, it's the one most likely to be heard by an undesired party (e.g. if EtP was activated because someone was about to enter the room). Similar to my thoughts on the ghost page, that something different to the default has been shown or said may be cause for suspicious in itself.

src/govuk/components/exit-this-page/exit-this-page.mjs
owenatgov: It'll be a white tie event 💅🏻

For real though I'm not sure why I didn't use `className` here. Makes much more sense. Good spot.

owenatgov:
# Update
I've updated the announcement content as per @calvin-lau-sig7's [microcopy update doc (only available internally)](https://docs.google.com/document/d/1h-loCCGRFyzAqMv6p_wXdwxvg4OBpduXVgRap0gj97E/edit). Having re-tested this, I'm happy with the results.

This now needs a final lookover from @davidc-gds before we can call it done.

davidc-gds:
Hey @owenatgov whatever's the best way to manage the development is good with me -- as long as we don't forget about revisiting the success announcement versus the loading overlay text announcement 👍

owenatgov:
@davidc-gds I reckon this is something we can progress in the discussion on https://github.com/alphagov/govuk-frontend/pull/3305

owenatgov:
After taking a look again at the microcopy doc and the discussion happening there, it does feel like this is done in it's current form. This will almost certainly be revisited as we progress the other spikes but I'm going to hit merge for now.

PR #3223, opened 2023-01-30
Organise review app middleware

colinrotherham:
We started organising our middleware into small exports for unit testing in https://github.com/alphagov/govuk-frontend/commit/807f5338dc972ffade1f1d0f62585f86c1c5be94

This PR adds a few more whilst I was working on the code


app/banner.js
```diff
@@ -1,10 +1,7 @@
-const cookieParser = require('cookie-parser')
-
 const BANNER_COOKIE_NAME = 'dismissed-app-banner'
 
 module.exports = function (app) {
   // Detect if banner should be shown based on cookies set
-  app.use(cookieParser())
```
36degrees, line 7: Can we split this change out into a separate commit?

app/common/middleware/request.js
```diff
@@ -0,0 +1,14 @@
...
+const express = require('express')
+
+/**
+ * Request handling
+ */
+const request = [
+  cookies(),
+
+  // Turn form POSTs into data that can be used for validation
+  express.urlencoded({ extended: true })
```
36degrees, line 11: Can we split this change out into a separate commit?

It looks like we no longer need `body-parser` because the functionality is now built into express – is that right? Would be good to capture in the commit message.

app/common/middleware/locals.js
```diff
@@ -0,0 +1,10 @@
+/**
+ * Legacy query handling
+ */
+const locals = (req, res, next) => {
+  res.locals.legacy = ['1', 'true'].includes(req.query.legacy)
```
36degrees, line 5: I'd also consider splitting this refactor out into a separate commit.

app/common/middleware/request.js
```diff
@@ -0,0 +1,14 @@
+const cookies = require('cookie-parser')
+const express = require('express')
+
+/**
+ * Request handling
+ */
+const request = [
```
36degrees, line 7: Not 100% sure about this file as it's just bringing in middleware from dependencies?

Could we just include these in `middleware/index.js` instead?

36degrees (commented):
The PR description mentions unit testing as a rationale for the change – would it make sense for us to add unit tests for these files as we go?

Generally when code is moved around I find it a lot easier to review if the moves happen in a separate commit with minimal other changes.

colinrotherham:
@36degrees Sounds good

One commit per middleware, with a basic test per commit?

Thanks for taking a look

app/common/middleware/request.js
colinrotherham: Yeah that's an option too

I'll explain my thinking in terms of testing GET or POST requests:

**HTTP GET request**

1. Cookies are added to `req.cookies`
2. Form submission empty `req.body`

**HTTP POST request**

1. Cookies are added to `req.cookies`
2. Form submission empty `req.body`

We're left with a nice way to test everything "HTTP request related" in one middleware

Hypothetically you'd put things like **Content-Encoding** `gzip` and `application/json` handling here too

```mjs
import compression from 'compression'

export const request = [
  compression(),
  express.json(),
  express.urlencoded({
    extended: true
  })
]
```

36degrees:
> One commit per middleware, with a basic test per commit?

Sounds sensible 👍🏻 Doesn't have to be strict though – the main thing for me is to try and avoid moving and changing code in the same commit as it makes it harder to review.

app/banner.js
colinrotherham: ✅ Done

app/common/middleware/request.js
colinrotherham: ✅ Done

>It looks like we no longer need body-parser because the functionality is now built into express – is that right? 

Yeah, the `express.urlencoded()` middleware was added to Express.js in v4.16.0
https://expressjs.com/en/api.html#express.urlencoded

app/common/middleware/locals.js
colinrotherham: ✅ Done

I've configured Express.js with the standard [`URLSearchParams()`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) too ready for Express.js v5

```mjs
const { query } = req

// Check for query
query.has('example')

// Get query value
query.get(example')

// Set query value
query.set('example', 'value')
```

app/common/middleware/request.js
colinrotherham: I've added tests for this

colinrotherham:
@36degrees Tests added which found a few broken things

Let me know if you want them on a separate PR

1. Component example back links don't append `?legacy=true`
2. Search example field names/values don't match JavaScript ~`sort`~ `order`
3. Search example field were only partially updated on submit (Organisation checkboxes)

app/views/full-page-examples/search/index.js
```diff
@@ -6,27 +6,30 @@ module.exports = (app) => {
...
-        total: formattedTotal,
         documents: shuffledDocuments,
-        order,
-        brexit,
-        values: request.query // In production this should sanitized.
+        order: query.get('order'),
+
+        // Make the total more readable
+        total: Number(randomizedTotal)
+          .toLocaleString('en', { useGrouping: true }),
```
colinrotherham, line 29: Where `useGrouping` (in **en** locale) adds commas to numbers

app/views/full-page-examples/search/index.njk
```diff
@@ -81,24 +82,22 @@ notes: >-
...
             },
             {
               value: "mod",
-              text: "Ministry of Defence",
-              checked: values["organisation"]|length and "mod" in values["organisation"]
+              text: "Ministry of Defence"
             }
-          ]
+          ],
+          values: values["organisation"]
```
colinrotherham, line 96: @36degrees Your nice feature definitely saved a lot of code

* https://github.com/alphagov/govuk-frontend/pull/2616

app/common/middleware/locals.js
36degrees: Is that because of the [change of the default query parser to `simple`](https://heynode.com/blog/2020-08/expressjs-5-nodes-latest-web-framework-release/#default-to-simple-query-parser-setting)?

I think the simple query parser would still work for our use case as we're not doing anything fancy with nested params etc. Do you think there's advantages to using `URLSearchParams` over it?

colinrotherham: Ah I remember that, there was a security vulnerability wasn't there

Primarily because we've now got the WHATWG [URL](https://url.spec.whatwg.org) and [Fetch](https://fetch.spec.whatwg.org) standards so it's all in Node.js too:

### Query to string
```mjs
// 'example1=value1&example2=value2'
req.query.toString()
```

### Query to object
```mjs
// { example1: 'value1', example2: 'value2' }
Object.fromEntries(req.query)
```

### Query via URL object
Where `window.location.search === '?example1=value1&example2=value2'`

```mjs
const url = new URL('http://localhost')

url.search = '?example1=value1&example2=value2'

url.searchParams.get('example1') // 'value1'
url.searchParams.get('example2') // 'value2'
```

Express.js has said they'll also switch in future
https://github.com/expressjs/express/issues/4990#issuecomment-1241976112

Happy to put it back to objects again

36degrees (approved):
Tidy 👍🏻 Great to have those tests!

PR #3224, opened 2023-01-30
Use `highlight.js` (server-side) code syntax highlighting

colinrotherham:
This PR replaces [**Prism** (client-side)](https://www.npmjs.com/package/prismjs) code highlighting with [**Highlight.js** (server-side)](https://www.npmjs.com/package/highlight.js)

The review app now matches the GOV.UK Design System code syntax highlighting:

* https://github.com/alphagov/govuk-design-system/pull/2562
* https://github.com/alphagov/govuk-design-system/pull/2566

Plus we've now got Nunjucks highlighting turned enabled where we previously didn't

Syntax highlighting enabled for both Nunjucks macro and HTML markup

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -1,39 +1,14 @@
-const { join } = require('path')
-
 const express = require('express')
-const app = express()
-const { marked } = require('marked')
-const nunjucks = require('nunjucks')
 
 const configPaths = require('../config/paths')
 const { getDirectories, getComponentsData, getFullPageExamples } = require('../lib/file-helper')
 const helperFunctions = require('../lib/helper-functions')
 
 const middleware = require('./common/middleware/index')
-
-const { HEROKU_APP } = process.env
-
-const appViews = [
-  configPaths.layouts,
-  configPaths.views,
-  configPaths.components,
-  join(configPaths.src, 'govuk'),
-  join(configPaths.node_modules, 'govuk_template_jinja')
-]
+const nunjucks = require('./common/nunjucks/index')
 
 module.exports = async (options) => {
-  const nunjucksOptions = options ? options.nunjucks : {}
-
-  // Configure nunjucks
-  const env = nunjucks.configure(appViews, {
-    autoescape: true, // output with dangerous characters are escaped automatically
-    express: app, // the express app that nunjucks should install to
-    noCache: true, // never use a cache and recompile templates each time
-    trimBlocks: true, // automatically remove trailing newlines from a block/tag
-    lstripBlocks: true, // automatically remove leading whitespace from a block/tag
-    watch: true, // reload templates when they are changed. needs chokidar dependency to be installed
-    ...nunjucksOptions // merge any additional options and overwrite defaults above.
-  })
+  const app = express()
 
   // Cache mapped components and examples
   const [componentsData, componentNames, exampleNames, fullPageExamples] = await Promise.all([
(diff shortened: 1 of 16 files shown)


domoscargin:
Given the Design System PR this mentions is back to draft, should this do the same, or is it ready for review?

colinrotherham:
> Given the Design System PR this mentions is back to draft, should this do the same, or is it ready for review?

Thanks @domoscargin that PR got split out so I've updated the description

Might want to discuss the code syntax theme though?

☝️ We've effectively swapped [`prism.css`](https://github.com/PrismJS/prism/blob/master/themes/prism.css) for [`github.css`](https://github.com/highlightjs/highlight.js/blob/main/src/styles/github.css)

romaricpascal:
The code changes look fine to me (and make the `app` file a bit lighter to go through) 🙌🏻 .

Regarding highlighting, could we start by vendoring [the same styles as on the design system repo](https://github.com/alphagov/govuk-design-system/blob/d1463ea6344b521bb69144700397e40e44e2d5e2/src/stylesheets/components/_highlight.scss#L4) so this PR gets through? Reasoning is that we're considering the highlighting good enough there, as it's deployed to production. If it needs revisiting we can do that in a separate PR. (for ex. do we want to look if we're missing classes, as there are [many more classes in highlight.js theme](https://github.com/highlightjs/highlight.js/blob/main/src/styles/github.css)).

colinrotherham:
Thanks @romaricpascal

Extra commit added to match the GOV.UK Design System highlight.js theme

app/app.js
```diff
@@ -45,20 +23,12 @@ module.exports = async (options) => {
 
   // Feature flags
   const flags = {
-    isDeployedToHeroku: !!HEROKU_APP
+    isDeployedToHeroku: !!process.env.HEROKU_APP
   }
 
-  // Share feature flags with middleware
-  env.addGlobal('flags', flags)
```
36degrees, line 52: We reference this in app/views/index.njk to avoid showing the link to the 'local' version of the Sass docs on the deployed review app:

https://github.com/alphagov/govuk-frontend/blob/56bd42d93ca653bfc6e62fac5dd5499f7e2aee10/app/views/index.njk#L45-L47

As `flags` is no longer a global in the Nunjucks environment, the link is now [showing on the preview app](https://govuk-frontend-pr-3224.herokuapp.com/).

app/common/nunjucks/index.js
```diff
@@ -0,0 +1,46 @@
...
+    watch: true // reload templates when they are changed. needs chokidar dependency to be installed
+  })
+
+  // Set view engine
+  app.set('view engine', 'njk')
+
+  // Custom filters
+  for (const params of Object.entries(filters)) {
+    env.addFilter(...params)
+  }
```
36degrees, line 36: This is a really neat use of `Object.entries`, but I must admit it took me a little while to work out what was going on. Wondering if it's worth simplifying slightly to make it clearer what's going on?

```suggestion
  for (key in filters) {
    env.addFilter(key, filters[key])
  }
```

36degrees (commented):
A couple of minor comments from me, but generally looking really good 👍🏻

app/app.js
colinrotherham: Well spotted, I've let Nunjucks grab it back from Express

```mjs
// Share feature flags with middleware
env.addGlobal('flags', app.get('flags'))
```

36degrees: Neat 🙌🏻

36degrees (approved):

PR #3247, opened 2023-02-03
Fix Exit this Page indicator colours not being suitable for hover/focus 

### LINKED ISSUE #3246: Fix hover styling for expanded Exit this Page button
## What
When the Exit this Page button is present on a page and the shift key is pressed once or twice, the button expands vertically to show indicators for how many times shift has been pressed.

When the Exit this Page button is expanded and `:hover` happens, the hover styling looks like it doesn't quite reach the centre area of the button.

Shift indicator with `:focus`: 
image

Shift indicator with `:hover`:
image

Shift indicator with `:focus` and `:hover`:
image

Live prototype where this can be tested, as of 3 February 2023: https://govuk-frontend-pr-3221.herokuapp.com/full-page-examples/child-maintenance

## Why

This is likely a bug due to the fact that buttons aren't generally expected to have extra height, unless inline content is being flowed to multiple lines. In this case the shift-press indicator isn't technically inline content, so the default button hover styling doesn't compensate for this scenario.

We should fix this broadly on the general button styles so that:
- Desktop users who press shift and then hover their pointer over the Exit this Page button don't see styling weirdness
- Future button implementations where a button contains non-text non-inline content don't encounter the same hover state styling issue

## Who needs to work on this
Developer(s)
Accessibility specialist

## Who needs to review this
Exit this Page epic lead, Accessibility specialist

## Done when

- [x] The Exit this Page hover styling looks correct when the shift keypress indicator is visible
- [ ] The implementation has been tested to ensure it doesn't negatively affect existing button styles
- [ ] The implementation is incorporated into our Exit this Page prototype

querkmachine:
Removes the background-color from the indicator dot container, as we want it to use whatever the button background colour is, including when hovering or focused.

Changes the indicator dots to use `inherit`/`currentcolor` so that the dots remain visible on the button's focus state. Still maintains the explicit colour definitions for IE8 as it doesn't support the `currentcolor` keyword.

Fixes #3246.

### DIFF
--- a/src/govuk/components/exit-this-page/_index.scss
+++ b/src/govuk/components/exit-this-page/_index.scss
@@ -28,8 +28,7 @@
     @include govuk-responsive-padding(2);
     display: none;
     padding-bottom: 0;
-    color: govuk-colour("white");
-    background-color: govuk-colour("red");
+    color: inherit;
     line-height: 0; // removes extra negative space below the indicators
     text-align: center;
     pointer-events: none;
@@ -44,12 +43,27 @@
     width: .5em;
     height: .5em;
     margin: 0 .125em;
-    border: 2px solid govuk-colour("white");
+    border-width: 2px;
+    border-style: solid;
     border-radius: 50%;
+
+    @include govuk-not-ie8 {
+      border-color: currentcolor;
+    }
+
+    @include govuk-if-ie8 {
+      border-color: govuk-colour("white");
+    }
   }
 
   .govuk-exit-this-page__indicator-light--on {
-    background-color: govuk-colour("white");
+    @include govuk-not-ie8 {
+      background-color: currentcolor;
+    }
+
+    @include govuk-if-ie8 {
+      background-color: govuk-colour("white");
+    }
   }
 
   @media only print {


owenatgov (approved):
Great stuff 💪🏻

src/govuk/components/exit-this-page/_index.scss
```diff
@@ -44,12 +43,27 @@
...
+
+    @include govuk-if-ie8 {
+      border-color: govuk-colour("white");
+    }
   }
 
   .govuk-exit-this-page__indicator-light--on {
-    background-color: govuk-colour("white");
+    @include govuk-not-ie8 {
+      background-color: currentcolor;
```
36degrees, line 61: Because backgrounds get removed in forced colors mode, the fill still doesn't show up even though we're using `currentcolor`.

It might be worth using a similar technique we use for radio buttons, where we make the entire thing out of border?

querkmachine: Good point! Will look at reworking how we light these up.

querkmachine: Actually, I'm going to progress this separately, as it isn't a new issue introduced by this PR and I'd rather not block the fixes with unrelated refactoring.

PR #3248, opened 2023-02-03
Tweak tab clicks to use `event.currentTarget`

colinrotherham:
A few minor tweaks to **Tabs** to simplify our code in places

1. **Find the tab link using `event.currentTarget`**
Prevents us having to check for child elements getting the click

2. ~**Move Tab bound handlers to constructor**~
~We were recreating handlers multiple times per tab~

**Update:** Commit for 2) has been merged via:

### DIFF
--- a/src/govuk/components/tabs/tabs.mjs
+++ b/src/govuk/components/tabs/tabs.mjs
@@ -260,19 +260,14 @@ Tabs.prototype.unsetAttributes = function ($tab) {
  * Handle tab link clicks
  *
  * @param {MouseEvent} event - Mouse click event
- * @returns {void | false} Returns void, or false within tab link
+ * @returns {void} Returns void
  */
 Tabs.prototype.onTabClick = function (event) {
   var $currentTab = this.getCurrentTab()
-  var $nextTab = event.target
+  var $nextTab = event.currentTarget
 
   if (!$currentTab || !$nextTab) {
     return
-  }
-
-  // Allow events on child DOM elements to bubble up to tab parent
-  if (!$nextTab.classList.contains('govuk-tabs__tab')) {
-    return false
   }
 
   event.preventDefault()


src/govuk/components/tabs/tabs.mjs
```diff
@@ -230,19 +230,17 @@ Tabs.prototype.unsetAttributes = function ($tab) {
...
  */
 Tabs.prototype.onTabClick = function (event) {
-  if (!event.target.classList.contains('govuk-tabs__tab')) {
-    // Allow events on child DOM elements to bubble up to tab parent
-    return false
-  }
   event.preventDefault()
-  var $newTab = event.target
+
+  var $nextTab = event.currentTarget
```
colinrotherham, line 238: Renamed to use `$nextTab` (not `$newTab`) for consistency

src/govuk/components/tabs/tabs.mjs
```diff
@@ -230,19 +230,17 @@ Tabs.prototype.unsetAttributes = function ($tab) {
  * Handle tab link clicks
  *
  * @param {MouseEvent} event - Mouse click event
- * @returns {void | false} Returns void, or false within tab link
+ * @returns {void} Returns void
  */
 Tabs.prototype.onTabClick = function (event) {
-  if (!event.target.classList.contains('govuk-tabs__tab')) {
```
colinrotherham, line 236: We've removed as `event.currentTarget` will be the tab link (not a child element)

We know the IE8 polyfill adds `event.currentTarget` support but just need to verify if it's clever enough to handle **target** vs **currentTarget** (might just alias them as the same thing)

colinrotherham: Yep from our [polyfill](https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/vendor/polyfills/Event.mjs#L140-L142)

```js
event.currentTarget = element;
event.relatedTarget = event.fromElement || null;
event.target = event.target || event.srcElement || element;
```

Where **element** is `this` (current target)

36degrees (approved):

36degrees:
Might be worth updating the PR title and description to avoid confusion.

Issue #3254, opened 2023-02-06
Including the JavaScript source map in the prototype kit config


nataliecarey:
## What

Add Javascript maps to the prototype kit config

## Why

1. Because at the moment there is a 404 when the console is open in the browser
2. So that prototype users get the benefit of the new source maps

## Who needs to work on this

 - Developer

## Who needs to review this

 - Developer

## Done when

- [ ] Source maps are used by the dev tools for prototype kit users

36degrees:
As discussed in dev catch up, we think that adding the source maps to the `assets` array should do what we want:

https://github.com/alphagov/govuk-frontend/blob/394c67abdd06895a7d05f70f5fc0dd549d178790/src/govuk-prototype-kit/govuk-prototype-kit.config.mjs#L29-L31

colinrotherham:
Related issue:

* https://github.com/alphagov/govuk-prototype-kit/issues/1950

colinrotherham:
Tested using `npm link` on a locally-built **GOV.UK Frontend** into a prototype 👍

PR #3255, opened 2023-02-06
Including the javascript source map in the prototype kit config.

nataliecarey:
A fix for https://github.com/alphagov/govuk-frontend/issues/3254


package/govuk-prototype-kit.config.json
```diff
@@ -1,6 +1,7 @@
 {
   "assets": [
-    "/govuk/assets"
+    "/govuk/assets",
+    "/govuk/all.js.map"
```
colinrotherham, line 4: Thanks @nataliecarey

Would you mind updating the `*.mjs` file instead of this one?
[**./src/govuk-prototype-kit/govuk-prototype-kit.config.mjs**#L30](https://github.com/alphagov/govuk-frontend/blob/main/src/govuk-prototype-kit/govuk-prototype-kit.config.mjs#L30)

We've generated it since this PR

* https://github.com/alphagov/govuk-frontend/pull/2851

@36degrees Will we want to add tests for "expected source maps" like we do in this one?
[**./tasks/gulp/__tests__/after-build-package.test.mjs**](https://github.com/alphagov/govuk-frontend/blob/main/tasks/gulp/__tests__/after-build-package.test.mjs)

For example we also have:

```
/govuk/common.js.map
/govuk/i18n.js.map
```

We can either choose to exclude these, or otherwise ensure we include them?

I'm not too familiar with the Prototype Kit but lots of other tooling has "load maps" options where source maps from other files can be included to supplement stack traces etc

**Gulp source maps** (example)
https://github.com/gulp-sourcemaps/gulp-sourcemaps#load-existing-source-maps

```js
gulp.src('src/**/*.js')
  .pipe(sourcemaps.init({ loadMaps: true }))
```

colinrotherham (changes requested):
Brilliant, thanks @nataliecarey

Before we approve would you mind dropping the `./package/**` changes? My fault for not explaining, we'll hold off any `dist` or `package` builds for a release commit:

(Here's the one from **v4.5.0** https://github.com/alphagov/govuk-frontend/commit/5bc602a85e6f7b890c93d989575996475cecacec)

nataliecarey:
Thanks Colin, re-reading your comment it was clear but I missed it.  I've reverted the package changes and force pushed to keep the package history in tact.

colinrotherham:
Thanks @nataliecarey

I'm thinking those paths need to be ~`/govuk-prototype-kit`~ `/govuk` and not forgetting [`all.js.map`](https://github.com/alphagov/govuk-prototype-kit/issues/1950#issue-1564699113)?

```js
'/govuk/assets',
'/govuk/all.js.map',
'/govuk/common.js.map',
'/govuk/i18n.js.map'
```

Sorry for causing all this hassle

colinrotherham:
@nataliecarey Do you want us to pick this up?

We've added https://github.com/alphagov/govuk-frontend/issues/3254 to our next release milestone

colinrotherham (approved):

PR #3257, opened 2023-02-07
Exit this Page - high contrast mode fixes

querkmachine:
@36degrees noted on #3247 that our current EtP indicators were not visible on forced colour modes (aka, high contrast mode), as the `background-color` is typically overridden in that scenario. This PR refactors the indicators so that the border is used to indicate 'lit' status, rather than a background colour. 

## Screenshots
|Browser / OS|Screenshot|
|:-|:-:|
|Chrome 109 / macOS 13.2|image|
|Firefox 109 / macOS 13.2|image|
|Safari 16.3 / macOS 13.2|image|
|IE 11 / Windows 10|image|
|IE 8 / Windows 7|image|
|Edge 108 / Windows 10<br>(High Contrast Mode)|image|

## Changes
- Changes the way we 'light' indicators. Rather than setting a background colour, the border width is increased in a manner that makes it appear to fill the indicator.

### DIFF
--- a/src/govuk/components/exit-this-page/_index.scss
+++ b/src/govuk/components/exit-this-page/_index.scss
@@ -1,6 +1,8 @@
 @import "../button/index";
 
 @include govuk-exports("govuk/component/exit-this-page") {
+  $indicator-size: .75em;
+
   .govuk-exit-this-page {
     z-index: 1000;
     top: 0;
@@ -39,9 +41,10 @@
   }
 
   .govuk-exit-this-page__indicator-light {
+    box-sizing: border-box;
     display: inline-block;
-    width: .5em;
-    height: .5em;
+    width: $indicator-size;
+    height: $indicator-size;
     margin: 0 .125em;
     border-width: 2px;
     border-style: solid;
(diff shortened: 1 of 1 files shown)


src/govuk/components/exit-this-page/_index.scss
```diff
@@ -56,13 +56,40 @@
...
+    width: $indicator-size;
+    height: $indicator-size;
+    margin: 0 .125em;
+    border-width: $indicator-border-width;
+  }
+
+  // High contrast/forced colour modes tend to remove background colours, so we
+  // create the 'lit' indicator  entirely out of 'border' so that they remain
+  // 'filled' even when colours are being overridden.
+  .govuk-exit-this-page__indicator-light:after {
```
36degrees, line 71: I think you could potentially avoid the after pseudo element by changing the size and border width of the indicator element instead?

Something like:

```scss
  .govuk-exit-this-page__indicator-light {
    display: inline-block;
    position: relative;
    width: $indicator-size;
    height: $indicator-size;
    margin: 0 .125em;
    border-width: $indicator-border-width;
  }

  .govuk-exit-this-page__indicator-light--on {
    width: 0;
    height: 0;
    border-width: ($indicator-size / 2) + $indicator-border-width;
  }
```

querkmachine: Happy to go that route instead! Would negate the need for an IE8-specific fix.

Could probably negate the need for fancy mathematics by using `box-sizing: border-box` instead too.

owenatgov (commented):
A quick comment to say I'm happy with this and think you've resolved Ollie's comment but I'm gonna re-request review from him to maintain review continuity.

PR #3258, opened 2023-02-07
Use standalone Sass task (no Gulp)

colinrotherham:
This PR addresses `gulp-sass` not supporting the Dart Sass modern compilation API

* https://github.com/dlmanning/gulp-sass/issues/837

It moves our Sass `compile:scss` task into a separate file but also:

1. Removes all "gulp wrapper" legacy packages
2. Calls Node Sass and PostCSS via their JavaScript API (no Gulp)
3. Creates a reusable `compileStylesheet()` helper ready for [deprecation log filtering](https://github.com/alphagov/govuk-frontend/pull/3192#issuecomment-1415640726)

It follows the same approach we took for Rollup:

* https://github.com/alphagov/govuk-frontend/pull/3088

Due to `gulp-postcss` we haven't been able to follow the PostCSS config options
https://github.com/postcss/postcss-load-config#options

These changes now show a clear `from` → `to` config for each stylesheet

Packages removed include:

```
gulp-plumber
gulp-postcss
gulp-sass
vinyl
```

Ticks off some tasks from:

* https://github.com/alphagov/govuk-frontend/issues/2712

But primarily helps unblock:

* https://github.com/alphagov/govuk-frontend/pull/3192


tasks/compile-stylesheets.mjs
```diff
@@ -1,130 +1,103 @@
...
+    file: moduleSrcPath,
+    outFile: moduleDestPath,
+
+    // Enable source maps
+    sourceMap: true,
+    sourceMapContents: true,
+
+    // Resolve @imports via
+    includePaths: [
+      join(paths.src, 'govuk'),
```
domoscargin: Let's take this line out, just in case it causes something unexpected to happen - can always review later.

tasks/compile-stylesheets.mjs
```diff
@@ -42,35 +51,48 @@ export async function compileStylesheet ([modulePath, { srcPath, destPath }]) {
...
+      file: moduleSrcPath,
+      outFile: moduleDestPath,
+
+      // Enable source maps
+      sourceMap: true,
+      sourceMapContents: true,
+
+      // Resolve @imports via
+      includePaths: [
+        join(paths.src, 'govuk'),
```
domoscargin, line 75: See previous comment.

domoscargin (approved):
LGTM

tasks/compile-stylesheets.mjs
colinrotherham: All sorted, yep happy to wait. Thank you 👍

PR #3268, opened 2023-02-08
Exit this page: Add attached keyboard help text

querkmachine:
Re-introduces having keyboard instructions "Press <kbd>Shift</kbd> 3 times" adjacent to the Exit this Page button, as per investigation #3267. 

This is done on the hypothesis that having that guidance next to the button is clearer to an end user than having the equivalent instructions on a separate guidance page. 

## Changes

image

- Adds help text displayed below the EtP button.
  - This text is injected as part of JavaScript initialisation and does not appear if JavaScript (and thus the keyboard-related functionality) is unavailable. 
  - This text remains 'stuck' to the viewport with the button when the user scrolls.
  - The help text is hidden on narrow screens, on the basis that screen real estate is more valuable and the vast majority of users are unlikely to be using a physical keyboard at this size.
- Implements our Frontend-wide localisation library into the EtP component's JavaScript for good measure.

## Things to consider

- The wording of the text. Currently this uses visually-hidden text to provide more context for screen readers. The full text being: "press <kbd>Shift</kbd> 3 times [to exit the page]"
  - Does it read better as "**or** press <kbd>Shift</kbd> 3 times"?
  - Is there liable to be confusion between pressing a key on a keyboard and 'pressing' something on a screen (aka, clicking)? 
- I've styled the word "Shift" differently to try and make it clearer that we mean Shift the key on a keyboard, not some other kind of Shift. This seems to be vaguely common practice, including on GitHub → <kbd>Shift</kbd>.
  - I'm not sure if there could be confusion here, as not all keyboards actually show the word "Shift" on the Shift key, using an upwards pointing arrow (⇧) instead, but I was worried this could be confused for the top arrow key.
  - Currently none of the help text, <kbd>Shift</kbd> included, is clickable. Should it be? 
- Hiding the help text on narrow screens also means that it disappears on high zoom levels. In this case, the point about screen real estate still applies, but the user is much more likely to be using a keyboard.

### DIFF
--- a/src/govuk/components/exit-this-page/_index.scss
+++ b/src/govuk/components/exit-this-page/_index.scss
@@ -82,6 +82,7 @@
       border-radius: 4px;
       background-color: govuk-colour("white");
       box-shadow: 0 1px 0 govuk-colour("white"), 0 2px 0 $govuk-border-colour;
+      white-space: nowrap;
     }
 
     @include govuk-media-query($from: tablet) {
--- a/src/govuk/components/exit-this-page/_index.scss
+++ b/src/govuk/components/exit-this-page/_index.scss
@@ -6,7 +6,6 @@
     z-index: 1000;
     top: 0;
     left: 0;
-    width: 100%;
 
     @include govuk-media-query($from: tablet) {
       right: 0;
@@ -78,7 +77,6 @@
 
     kbd {
       @include govuk-font($size: 16);
-      display: inline-block;
       padding: 0 govuk-spacing(1);
       border: 1px solid $govuk-border-colour;
       border-radius: 4px;
--- a/src/govuk/components/exit-this-page/_index.scss
+++ b/src/govuk/components/exit-this-page/_index.scss
@@ -13,6 +13,8 @@
 
     @include govuk-media-query($from: tablet) {
       display: inline-block;
+      display: inline-flex;
+      flex-direction: column;
       right: 0;
       left: auto;
       width: auto;
@@ -63,6 +65,28 @@
 
     @include govuk-if-ie8 {
       background-color: govuk-colour("white");
+    }
+  }
+
+  .govuk-exit-this-page__help {
+    @include govuk-responsive-padding(2);
+    @include govuk-font($size: 16);
+    display: none;
+    border: 1px solid $govuk-border-colour;
+    border-top: none;
+    background-color: govuk-colour("white");
+    text-align: center;
+
+    kbd {
+      @include govuk-font($size: 16);
+      display: inline-block;
+      padding: 0 govuk-spacing(1);
+      border: 1px solid $govuk-border-colour;
+      border-radius: 4px;
+    }
+
+    @include mq($from: tablet) {
+      display: block;
     }
   }
 
--- a/src/govuk/components/exit-this-page/_index.scss
+++ b/src/govuk/components/exit-this-page/_index.scss
@@ -82,6 +82,8 @@
       padding: 0 govuk-spacing(1);
       border: 1px solid $govuk-border-colour;
       border-radius: 4px;
+      background-color: govuk-colour("white");
+      box-shadow: 0 1px 0 govuk-colour("white"), 0 2px 0 $govuk-border-colour;
     }
 
     @include govuk-media-query($from: tablet) {
--- a/src/govuk/components/exit-this-page/_index.scss
+++ b/src/govuk/components/exit-this-page/_index.scss
@@ -2,27 +2,26 @@
 
 @include govuk-exports("govuk/component/exit-this-page") {
   .govuk-exit-this-page {
+    @include govuk-responsive-padding(8, "bottom");
     z-index: 1000;
     top: 0;
     left: 0;
     width: 100%;
-    @supports (position: sticky) {
-      @include govuk-responsive-padding(8, "bottom");
-      position: sticky;
-    }
 
     @include govuk-media-query($from: tablet) {
-      display: inline-block;
-      display: inline-flex;
-      flex-direction: column;
       right: 0;
       left: auto;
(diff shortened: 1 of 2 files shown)


src/govuk/components/exit-this-page/_index.scss
```diff
@@ -2,25 +2,26 @@
 
 @include govuk-exports("govuk/component/exit-this-page") {
   .govuk-exit-this-page {
+    @include govuk-responsive-padding(8, "bottom");
     z-index: 1000;
     top: 0;
     left: 0;
     width: 100%;
```
owenatgov, line 9: I don't think this is necessary anymore if we're applying `width: 100%` to the button on line 24.

src/govuk/components/exit-this-page/_index.scss
```diff
@@ -66,6 +67,30 @@
...
+    @include govuk-font($size: 16);
+    display: none;
+    border: 1px solid $govuk-border-colour;
+    border-top: none;
+    background-color: govuk-colour("white");
+    text-align: center;
+
+    kbd {
+      @include govuk-font($size: 16);
+      display: inline-block;
```
owenatgov, line 81: Is this `inline-block` necessary? From a very quick VO test the screen reader is splitting the announcement of the instructions into 3 parts:

- press
- Shift
- 3 times to exit this page

...which is coming from us sort of blocking the element. Removing this makes VO interpret the line as a single announcement. The result is that visually the padding get's a little squashed which isn't a dealbreaker for me personally.

This isn't a showstopper though so happy to negotiate.

src/govuk/components/exit-this-page/exit-this-page.mjs
```diff
@@ -1,21 +1,50 @@
...
+/**
+ * @constant
+ * @type {ExitThisPageTranslations}
+ * @see Default value for {@link ExitThisPageConfig.i18n}
+ * @default
+ */
+var EXIT_THIS_PAGE_TRANSLATIONS = {
+  keypressProgress: 'Exit this Page key press %{press} of 3',
+  keypressComplete: 'Exit this page activated',
+  helpText: 'press <kbd>Shift</kbd> 3 times<span class="govuk-visually-hidden"> to exit the page</span>'
```
owenatgov, line 19: Mega nitpick time: VO's text display is displaying the announcement from "3 times" like so:

>...3 timesto exit this page

It's interpreting it in the audio announcement fine but moving the space out of the `span` or putting a `&nbsp;` in there should sort this out:

```suggestion
  helpText: 'press <kbd>Shift</kbd> 3 times <span class="govuk-visually-hidden">to exit the page</span>'
```

owenatgov (commented):
This looks grand. I really like the use of `kbd` which has surprisingly decent support across browsers. I've added some small comments and have some general opinions:

- I'm personally ok with the instructions not being clickable. In the demo where we had it before it wasn't clickable but that's not necessarily a reason not to say it should be.
- I'm not super happy that we're not showing the instructions on small screens, mostly for the zoom case, but it feels like the correct call. Whilst I don't think it takes up that much space, the likelihood of a user using mobile and having a keyboard is lower (though not impossible) and might insight the worry that you mentioned of folks expecting there to be something to press on the screen itself.
- Your concerns about the wording here may be solved by @calvin-lau-sig7's microcopy work

As usual, it's worth getting others' opinions here, notably a designer and @davidc-gds if he can spare the time.

src/govuk/components/exit-this-page/exit-this-page.mjs
querkmachine: Have done so 👍

src/govuk/components/exit-this-page/_index.scss
querkmachine: The `inline-block` isn't needed, no. I think I added it with the expectation there'd be some vertical padding being applied, but I didn't end up using any in the end.

src/govuk/components/exit-this-page/_index.scss
querkmachine: Checked and it is indeed not doing much. Have removed.

Ciandelle:
I would feel more comfortable with the longer text here. I think that users are likely to think that they can click it before associating it with the keyboard itself. Though I do like the stylisation of shift, I'd want @davidc-gds opinion to make sure that we're comfortable that this would still pass accessibility tests. My only concern would be it being images/words which I know @calvin-lau-sig7 has been looking at too. 

Otherwise it looks really great. I think the additional text help it really useful to users. Great work :)

davidc-gds:
There appears to be a layout bug that happens between 641px and 757px on the current preview when the 'press shift' text is attached to the EtP button:

Here's a screenshot of how it looks at 641px, with shift pressed once:
image

The page title `<h1>` is pushed down in the layout, and the EtP button starts moving above it. This leads to a large vertical gap of white space above the page title.

src/govuk/components/exit-this-page/exit-this-page.mjs
```diff
@@ -1,21 +1,50 @@
...
 import '../../vendor/polyfills/Element/prototype/dataset.mjs'
 import '../../vendor/polyfills/Function/prototype/bind.mjs'
 
+/**
+ * @constant
+ * @type {ExitThisPageTranslations}
+ * @see Default value for {@link ExitThisPageConfig.i18n}
+ * @default
+ */
+var EXIT_THIS_PAGE_TRANSLATIONS = {
```
owenatgov, line 16: A general comment that even if we end up putting this work down for now, we should retain this as it's good for getting the final code up to scratch.

querkmachine:
Consensus has been [not to merge](https://github.com/alphagov/govuk-frontend/issues/3267#issuecomment-1452152807) for now.

PR #3272, opened 2023-02-10
Add empty alt attribute to logo IE8 fallback PNG

### LINKED ISSUE #3190: The logo in Header need to be decorative
In header the element actually have this code:

```html
<!--
<!--[if IE 8]>
          <img src="/assets/images/govuk-logotype-crown.png" class="govuk-header__logotype-crown-fallback-image" width="36" height="32">
          <![endif]-->
-->
```

As you have in svg logo w/ aria-hidden="true" and aria-focusable="false" to declare this element as decorative to assistive technologies you need something similar to the <img element. If not we got an image without alternative text.

So change to (add alt=""):

```html
<!--
<img src="/assets/images/govuk-logotype-crown.png" class="govuk-header__logotype-crown-fallback-image" width="36" height="32" 
                 alt="">

--> 
```

36degrees:
Closes #3190.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+### Fixes
+
+We’ve made fixes to GOV.UK Frontend in the following pull requests:
+
+- [#3272: Add empty alt attribute to logo IE8 fallback PNG](https://github.com/alphagov/govuk-frontend/pull/3272)
+
 ## 4.5.0 (Feature release)
 
 ### New features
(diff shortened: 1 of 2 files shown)


36degrees:
I've added an entry to the changelog.

Technically this is a markup change, but given the incredibly low number of users visiting from IE8 any more, I don’t think it’s worth instructing service teams to make this change.

At least it’ll be correct in the documentation and for anyone using our Nunjucks macros.

owenatgov (approved):
Does the job 👍🏻

PR #3277, opened 2023-02-13
Add PRs labelled `dependabot` to the project board

colinrotherham:
This PR configures Dependabot PRs to be added automatically to our project board

It uses the [Add to GitHub Projects](https://github.com/marketplace/actions/add-to-github-projects) action mentioned in:

* https://github.com/alphagov/govuk-frontend/pull/2950#issuecomment-1301931845

The official GitHub Action is quite limited but 3rd party alternatives are available should non-default columns be needed: https://docs.github.com/en/actions/managing-issues-and-pull-requests/moving-assigned-issues-on-project-boards

### DIFF
--- a/.github/workflows/triage-pull-requests.yml
+++ b/.github/workflows/triage-pull-requests.yml
@@ -1,0 +1,21 @@
+name: Triage pull requests
+
+on:
+  pull_request:
+    types:
+      - labeled
+      - opened
+
+jobs:
+  add-to-project:
+    name: Triage pull requests
+    runs-on: ubuntu-latest
+    steps:
+      - name: Add to project
+        uses: actions/add-to-project@v0.4.0
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+
+          # Add to 'Design System Sprint Board'
+          project-url: https://github.com/orgs/alphagov/projects/53
+          labeled: dependabot


domoscargin:
Yeeeesss!

I think we'd ideally want to:

* label the dependabot PR with "shared ownership" so it appears on each squad's views
* Assign the Design System developers group (if we don't do that already)
* Add the "Needs review" status (This depends on how quickly we'd like to triage dependabot stuff - we could decide that we'll pop 'em on the backlog and add some to a sprint, but it feels like we're likely to just want to continue dealing with them as they come in).

colinrotherham:
> Yeeeesss!
>
> I think we'd ideally want to:
> 

@domoscargin Brill. See what you think of these replies 😊 

Everything seems possible but it's intentionally MVP for now

Can pick it up if we think it's needed


## Edit issue project fields
> * label the dependabot PR with "shared ownership" so it appears on each squad's views

Looks like this one needs the GitHub Actions GraphQL API

**Automating projects**
[Updating a single select field](https://docs.github.com/en/issues/planning-and-tracking-with-projects/automating-your-project/using-the-api-to-manage-projects#updating-a-single-select-field)

## Edit issue project column

> * Add the "Needs review" status (This depends on how quickly we'd like to triage dependabot stuff - we could decide that we'll pop 'em on the backlog and add some to a sprint, but it feels like we're likely to just want to continue dealing with them as they come in).

Looks like we can choose the default column in the Project workflows but needs turning on:
https://github.com/orgs/alphagov/projects/53/workflows/6369989

## Edit issue assignees
> * Assign the Design System developers group (if we don't do that already)

Similar to how we [comment on diffs](https://github.com/alphagov/govuk-frontend/blob/main/.github/workflows/diff-change-to-dist.yaml#L54) (using the GitHub Actions API) we can edit issues and PRs with:

**Issue assignees**
[Add assignees to an issue](https://docs.github.com/en/rest/issues/assignees?apiVersion=2022-11-28#add-assignees-to-an-issue)

domoscargin (approved):
I've enabled the PR -> Needs Review and Issue -> Backlog workflows on the project which takes care of the Project Column. 

I think that makes this work for me as an MVP, as these will surface for us to label and assign properly.

.github/workflows/triage-pull-requests.yml
```diff
@@ -0,0 +1,21 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Add to project
+        uses: actions/add-to-project@v0.4.0
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+
+          # Add to 'Design System Sprint Board'
+          project-url: https://github.com/orgs/alphagov/projects/53
+          labeled: dependabot
```
colinrotherham, line 21: @domoscargin Are you happy with this only triaging Dependabot?

domoscargin: For now, yeah. I'd expect anyone on the team to handle their own stuff.

"awaiting triage" tickets are kinda in the same boat of potentially getting missed, but we probably need a brief team discussion about how best to handle these - I feel like adding 'em to the backlog probably risks they get just as ignored as they often do.

colinrotherham: Don't think it excludes drafts but we can!

Ahh so quite often we just put an issue on the board, with related PRs only linked to the issue?

domoscargin: Hmm, I personally like having all PRs in Needs Review with their description linking back to their Epic, but I'm not everyone else, so I'll pivot again (!) and say let's pop this in as just dependabot, and we can discuss other board management automation with the team.

PR #3300, opened 2023-02-16
Remove redundant Jest "port available" checks

colinrotherham:
The following Dependabot update includes [`jest-dev-server@7.0.1`](https://github.com/argos-ci/jest-puppeteer/blob/main/packages/jest-dev-server/CHANGELOG.md#701-2023-02-15)

* https://github.com/alphagov/govuk-frontend/pull/3298

With "dev server" port detection on macOS now fixed we can remove all our workaround code 🙌

Much easier to run training sessions on code that isn't there 😊 

#### Dependencies removed

```console
wait-on
```

### DIFF
--- a/config/jest/browser/close.mjs
+++ b/config/jest/browser/close.mjs
@@ -1,13 +1,8 @@
 import { teardown } from 'jest-environment-puppeteer'
 
-import serverStop from '../server/stop.mjs'
-
 /**
  * Close browser
  */
 export default async function browserClose () {
-  await Promise.all([
-    serverStop(), // Stop web server
-    teardown() // Close browser
-  ])
+  await teardown() // Close browser, stop server
 }
--- a/config/jest/browser/close.mjs
+++ b/config/jest/browser/close.mjs
@@ -1,8 +1,4 @@
 import { teardown } from 'jest-environment-puppeteer'
 
-/**
- * Close browser
- */
-export default async function browserClose () {
-  await teardown() // Close browser, stop server
-}
+// Close browser, stop server
+export default () => teardown()
--- a/config/jest/server/start.mjs
+++ b/config/jest/server/start.mjs
@@ -2,9 +2,5 @@ import { setup } from 'jest-dev-server'
 
 import config from '../../../jest-dev-server.config.js'
 
-/**
- * Start web server
- */
-export default async function serverStart () {
-  await setup(config)
-}
+// Start web server
+export default () => setup(config)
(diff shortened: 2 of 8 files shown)


colinrotherham:
Discussed with @romaricpascal and @domoscargin on Slack

config/jest/browser/close.mjs
```diff
@@ -1,13 +1,8 @@
...
-
 /**
  * Close browser
  */
-export default async function browserClose () {
-  await Promise.all([
-    serverStop(), // Stop web server
-    teardown() // Close browser
-  ])
+export default async function browserOpen () {
```
romaricpascal, line 6: **issue** I'm not sure why the named changed there, that function still closes the browser 😄 

```suggestion
export default async function browserClose () {
```

**suggestion** I don't think there's any need to `await` inside the function, we can just have it be a plain (as in non-`async`) function that returns the `Promise` returned by `teardown` 😄 Or even just export `teardown` as the default export.

config/jest/server/start.mjs
```diff
@@ -1,42 +1,10 @@
...
-    await ready({ timeout })
-  } catch (error) {
-    await setup({
-      // Ensure PORT works (on Windows) + SIGINT/SIGTERM signal events
-      command: `cross-env-shell PORT=${PORT} node app/start.js`,
-      port: PORT
-    })
-    await ready()
-  }
+  await setup(config)
```
romaricpascal, line 9: **suggestion** Similar suggestion to `browserClose`, we don't need to `await` and can just return the `Promise` from `setup` 😄

romaricpascal (commented):
Looking neat to me, appart from what looks like a typo/misnaming in the `close.mjs` file. Also dropped suggestions around `await`s that may not be necessary, but they're not blockers 😄

config/jest/browser/close.mjs
colinrotherham: Oh nice spot on that copy-paste-typo!

colinrotherham: Will mark this as resolved and reply about Promises in https://github.com/alphagov/govuk-frontend/pull/3300#discussion_r1111799019

config/jest/server/start.mjs
colinrotherham: Ahh so all these Jest hooks were previously typed with `Promise<void>`

We're not adding JSDoc types or extending anything anymore so shall we simplify even more?

```suggestion
  export default () => setup(config)
```

colinrotherham: Pushed up Jest hooks using the shorthand above 👍

romaricpascal (approved):
Looks good to go to me! thanks for the little edits! 🙌🏻

PR #3301, opened 2023-02-16
Remove redundant Jest alternative port

colinrotherham:
This PR lets Jest tests use the same Review app seen in browsers

It also speeds up local testing by closing the `Ctrl + c` and restart cycle

### Before

During local testing we currently launch two copies of the Review app:

1. Running `npm start` launches http://localhost:3000
2. Running `npx jest --watch XXX` launches http://localhost:8888

❌ Server-side changes in **./app** only appear on `:3000` but not `:8888`

### After

1. Running `npm start` launches http://localhost:3000
2. Running `npx jest --watch XXX` reuses http://localhost:3000

✅ Server-side changes in **./app** appear for both browser testing and Jest runs


jest-dev-server.config.js
```diff
@@ -1,14 +1,11 @@
...
-const PORT = process.env.PORT || ports.test
-
 /**
  * @type {import('jest-dev-server').JestDevServerOptions}
  */
 module.exports = {
-  // Ensure PORT works (on Windows) + SIGINT/SIGTERM signal events
-  command: `cross-env-shell PORT=${PORT} node app/start.js`,
-  port: PORT,
+  command: 'npm start --workspace app',
```
36degrees, line 7: We confident we don't need `cross-env` to handle signals still, as per the comment above it?

36degrees (approved):
Looks good – just want to double check we're definitely happy we don't need `cross-env` any more.

jest-dev-server.config.js
colinrotherham: Sorry it was badly worded by me originally!

It was to explain [`cross-env` vs `cross-env-shell`](https://github.com/kentcdodds/cross-env#cross-env-vs-cross-env-shell)

>On Windows you need to use cross-env-shell, if you want to handle [signal events](https://nodejs.org/api/process.html#process_signal_events) inside of your program. A common case for that is when you want to capture a SIGINT event invoked by pressing Ctrl + C on the command-line interface.

Neither of them are needed if we drop `PORT`

Issue #3266, opened 2023-02-08
Investigate adding text to the ghost page in the Exit this Page component


owenatgov:
## What
Build out a version of the Exit this Page component with transition text eg: "loading..." on the "ghost page" feature.

## Why
[When we originally built the ghost page](https://github.com/alphagov/govuk-frontend/pull/2949), we asserted that we wouldn't include text on the page under the assessment that it wasn't necessary when included in the context of the user activating the button, it would be challenging to brand and as it wasn't standard browser behaviour it could clue in observers that the user is looking at something suspicious.

This has since been challenged on the basis that text could give users a clearer message that they are navigating to a new page and could provide a means to avoid the user getting confused if the new page fails to load.

The intent of this spike is to build an in situ version with text on the ghost page to assess more clearly if this is valuable or not.

Part of https://github.com/alphagov/govuk-design-system/issues/1755

## Assumptions
- This is standard practice based on its use in Google docs products when clicking a link in a document that goes to a different domain

## Timebox
**1 working week (5 days) from start**

## Who is working on this?
@owenatgov @querkmachine @davidc-gds @Ciandelle @calvin-lau-sig7 

Spike lead:
@owenatgov 

## Questions to answer

- [x] Does unbranded text put users at risk?
- [x] Does the text make the button activation journey clearer for users?

## Done when

- [x] Questions have been answered or we have a clearer idea of how to get to our goal
- [x] Findings have been reviewed and agreed with at least one other person
- [x] Findings have been shared, e.g: via a write-up on the ticket, at a show & tell or team meeting

owenatgov:
# Summary of findings from the spike
- We'll be referring to this feature as the "loading overlay" instead of the "ghost page" as this has created confusion about what the feature actually is
- We don't feel that the text creates risk for folks on the basis that it's unbranded and unassuming
- The text is a sensible place from which to announce a version of the status message that the button has been activated to screen readers
- We've decided to use "Loading" as opposed to "Exiting page" for the activation announcement on the grounds that it'll disassociate navigation with the button and lines up with the user navigating to a new page
- We ran into a potential bug that users could interact with the content under the overlay between activating the button and navigating to a new page which we fixed via an extra class we're adding to the body to hide everything as well as use the overlay, which can act as a failsafe in case teams include anything that could conflict with the body class

PR #3305, opened 2023-02-17
Exit this Page: Add text to the "ghost page" button feature

owenatgov:
## What/Why
Adds text to the "ghost page" feature of the exit this page button so that we can revisit if this has a positive or negative impact on users. The text is unbranded and unstyled.

Additionally this change moves the button activation announcement for screen readers to the ghost page itself as opposed to the announcement `span` being used currently.

Closes https://github.com/alphagov/govuk-frontend/issues/3266

## Notes
From a quick screen reader test, as I've been finding whilst working on https://github.com/alphagov/govuk-frontend/issues/3061, there is an inconsistency in the button activation alert being announced before the screen reader "beats" the announcement and starts instead announcing navigation and the new page.

### DIFF
--- a/src/govuk/components/exit-this-page/_index.scss
+++ b/src/govuk/components/exit-this-page/_index.scss
@@ -67,7 +67,7 @@
     }
   }
 
-  .govuk-exit-this-page__overlay {
+  .govuk-exit-this-page-overlay {
     position: fixed;
     z-index: 9999;
     top: 0;
@@ -75,5 +75,22 @@
     bottom: 0;
     left: 0;
     background-color: govuk-colour("white");
+  }
+
+  // This class is added to the body when the Exit This Page button is activated
+  // in addition to the overlay to both block the entire screen and hide everything
+  // underneath it.
+  //
+  // We do this to ensure that users don't risk interacting with the page underneath
+  // the overlay between activating the button and navigating to the next page.
+  .govuk-exit-this-page-hide-content {
+    // stylelint-disable declaration-no-important
+    * {
+      display: none !important;
+    }
+
+    .govuk-exit-this-page-overlay {
+      display: block !important;
+    }
   }
 }
--- a/src/govuk/components/exit-this-page/exit-this-page.mjs
+++ b/src/govuk/components/exit-this-page/exit-this-page.mjs
@@ -14,7 +14,7 @@ import '../../vendor/polyfills/Function/prototype/bind.mjs'
  * @default
  */
 var EXIT_THIS_PAGE_TRANSLATIONS = {
-  activated: 'Exiting page.',
+  activated: 'Loading.',
   timedOut: 'Exit this page expired.',
   pressTwoMoreTimes: 'Shift, press 2 more times to exit.',
   pressOneMoreTime: 'Shift, press 1 more time to exit.'
@@ -132,7 +132,7 @@ ExitThisPage.prototype.updateIndicator = function () {
 
 /**
  * Initiates the redirection away from the current page.
- * Includes the 'ghost page' functionality, which covers the current page with a
+ * Includes the loading overlay functionality, which covers the current page with a
  * white overlay so that the contents are not visible during the loading
  * process. This is particularly important on slow network connections.
  *
@@ -146,10 +146,19 @@ ExitThisPage.prototype.exitPage = function (e) {
     redirectUrl = e.target.href
   }
 
+  this.$updateSpan.innerText = ''
+
   // Blank the page
+  // As well as creating an overlay with text, we also set the body to hidden
+  // to prevent screen reader and sequential navigation users potentially
+  // navigating through the page behind the overlay during loading
+  document.body.classList.add('govuk-exit-this-page-hide-content')
   this.$overlay = document.createElement('div')
-  this.$overlay.className = 'govuk-exit-this-page__overlay'
+  this.$overlay.className = 'govuk-exit-this-page-overlay'
+  this.$overlay.setAttribute('role', 'alert')
+
   document.body.appendChild(this.$overlay)
+  this.$overlay.innerText = this.i18n.t('activated')
 
   window.location.href = redirectUrl
 }
@@ -186,11 +195,9 @@ ExitThisPage.prototype.handleEscKeypress = function (e) {
 
     if (this.escCounter >= 3) {
       this.escCounter = 0
+
       clearTimeout(this.keypressTimeoutId)
       this.keypressTimeoutId = null
-
-      this.$updateSpan.setAttribute('role', 'alert')
-      this.$updateSpan.innerText = this.i18n.t('activated')
 
       this.exitPage()
     } else {
@@ -254,6 +261,8 @@ ExitThisPage.prototype.resetEscTimer = function () {
  */
 ExitThisPage.prototype.resetPage = function () {
   // If an overlay is set, remove it and reset the value
+  document.body.classList.remove('govuk-exit-this-page-hide-content')
+
   if (this.$overlay) {
     this.$overlay.remove()
     this.$overlay = null
--- a/src/govuk/components/exit-this-page/exit-this-page.test.js
+++ b/src/govuk/components/exit-this-page/exit-this-page.test.js
@@ -6,7 +6,7 @@ const { goToComponent, goToExample } = require('../../../../lib/puppeteer-helper
 
 const buttonClass = '.govuk-js-exit-this-page-button'
 const skiplinkClass = '.govuk-js-exit-this-page-skiplink'
-const overlayClass = '.govuk-exit-this-page__overlay'
+const overlayClass = '.govuk-exit-this-page-overlay'
 
 describe('/components/exit-this-page', () => {
   describe('when JavaScript is unavailable or fails', () => {
@@ -147,8 +147,8 @@ describe('/components/exit-this-page', () => {
         await page.keyboard.press('Shift')
         await page.keyboard.press('Shift')
 
-        const message = await page.evaluate((buttonClass) => document.querySelector(buttonClass).nextElementSibling.innerText, buttonClass)
-        expect(message).toBe('Exiting page.')
+        const message = await page.evaluate((overlayClass) => document.querySelector(overlayClass).innerText, overlayClass)
+        expect(message).toBe('Loading.')
       })
 
       it('announces when the keyboard shortcut has timed out', async () => {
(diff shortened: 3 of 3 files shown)


src/govuk/components/exit-this-page/exit-this-page.mjs
```diff
@@ -115,7 +115,9 @@ ExitThisPage.prototype.exitPage = function (e) {
   // Blank the page
   this.$overlay = document.createElement('div')
   this.$overlay.className = 'govuk-exit-this-page__overlay'
+  this.$overlay.setAttribute('role', 'alert')
   document.body.appendChild(this.$overlay)
+  this.$overlay.innerText = 'Exiting page'
```
querkmachine, line 120: Not sure how much of a difference this makes, but setting `innerText` before doing the `appendChild` feels neater to me—like we're doing all the set up and then adding the entire overlay element (text nodes and all) to the DOM, rather than adding the element, then updating the `innerText` separately.

querkmachine (approved):
Code looks good to me. 

I assume we're OK leaving `$updateSpan` reading "[...] press 2 of 3" on activation given we're leaving the page anyway at this point—with either the overlay or the browser hijacking any further screen reader announcements.

I see there's still a discussion ongoing about what the ghost page's text should say, so I won't weigh in on that just yet.

src/govuk/components/exit-this-page/exit-this-page.mjs
querkmachine: Ah, never mind! It just hit me that `role="alert"` is a live region, and will only read out content when it's _updated_, so setting `innerText` after-the-fact is kind of necessary. Y'all're good. 👍

owenatgov:
@querkmachine Re: clearing the span, I think that it doesn't really make a difference but it probably can't hurt to clear it. Cheers for pointing it out.

davidc-gds:
This generally looks good to me, and I'll add to the discussion of what the text should actually say in the internal microcopy doc. 👍

I've been one of the more vocal proponents for having some kind of status on the blank / loading / ghost page, especially as a way to assure people that the page didn't just crash when they're navigating on a slow connection.

After seeing it in action, I'm still in favour of it, with a couple extra notes of details:

- I think we should rename it to something like 'loading overlay', rather than ghost page. It is neither a ghost, nor a page.
- This loading overlay's text should communicate the current status, ideally without drawing attention to the EtP feature. I like Calvin's current suggestion of "Loading". It says what's it's doing and keeps it simple.
- I think this new "Loading" text should be announced by screen readers, but it's not wildly essential that it gets announced. But announcing it could be helpful for people on slow connections where it hands for a few seconds before the next page comes up.
- I think we might be able to fully replace the 'exit this page successful' announcement with announcing this 'loading overlay' text instead. It's a tradeoff as far as what info we're providing to screen reader users, but also screen readers aren't reading out the 'success' message anyhow, because there's no time.

querkmachine:
As another thing that just popped into my head during discussions on Slack: We haven't done anything to ensure that the content beneath the overlay _stops_ being accessible, via keyboard navigation, assistive tech or other means.

Whilst probably a minor issue given the user will be in the process of directing away from the page anyway, it could potentially lead to the accidental activation of a different control or announcement of different text after the EtP button has been activated.

Ciandelle:
100% agree with renaming the page to loading overlay, feels more in the style of calling it what is it.
Also like the idea of keeping things as simple as possible. In terms of the announcement to screen readers, wouldn't it be better from to announce 'loading complete', rather than 'exit this page successful' if they are in a dangerous situation anyway?

calvin-lau-sig7:
"Exiting this page" (telling the user they've successfully activated Exit this Page) and "Loading" (You are going to a new website) are 2 separate messages, but combining them does make sense if WCAGs allow.

Just "Loading" might be enough to assure the user that they've activated Exit this page correctly, but it'd be nice to reinforce somehow.

Is there any other way to give an audio indication that Exit this page has activated?

owenatgov:
Some thoughts from the discussion so far:

Firstly, after seeing this in action I also don't hate it. This isn't 100% informed but my instinct is that this isn't significant or unusual enough that going to put users at risk like we assessed (one for the community crit?).

Secondly, I agree that we should do something about moving focus and blocking the page for the user. I'll work on this next once I'm back from AL.

Secondly, on the announcement itself and the inconsistency in how it's announced across screen readers. I don't see a huge amount of harm in having a generic message like "loading" to signify button activation. I also don't think it only being announced some of the time is a major problem. As a reminder, this was found during the testing for https://github.com/alphagov/govuk-frontend/pull/3221 that SRs were rarely announcing the activation message over letting the user know that they were on a new page. I suspect this is the result of a race condition where the browser navigation is "beating" the activation message 99% of the time.

I don't think we should be trying to "hack" our screen reader setup to be more consistent with the exact announcement ecosystem we want, in part because I don't think there's a positive way to do so. `role="alert"` is intended as the most aggressive way to get an announcement out and all major screen readers know that text changing in an `alert` region means that they should stop their announcement and read this new status message right away. That's what we're doing and I think trying to stretch this artificially using tricks eg: adding a timeout to give the screen reader time to read the announcement would create technical risk.

I'm personally in favour of giving way to the opinion of screen readers ie: the screen readers think that telling a user they've navigated is more important than quickly announcing this message before navigation, therefore that's what we should accept.

owenatgov:
Okie dokie, I've fixed the navigation under the overlay issue by adding a class to the body which sets everything to display none. From some light testing, I don't think this requires an `!important` but I'm not 100% confident. Interested in specific takes on that by @querkmachine and @davidc-gds.

More broadly, I think we're in agreement that we're actually gonna implement this and that the current solution works nicely? Calling additionally on @calvin-lau-sig7 and @Ciandelle as folks in the convo so far for any additional thoughts.

src/govuk/components/exit-this-page/_index.scss
```diff
@@ -78,4 +78,8 @@
     left: 0;
     background-color: govuk-colour("white");
   }
+
+  .govuk-exit-this-page-hide-content *:not(.govuk-exit-this-page__overlay) {
```
querkmachine, line 82: `:not` won't work in IE8, but I'm not gonna lose sleep over it.

src/govuk/components/exit-this-page/_index.scss
```diff
@@ -78,4 +78,8 @@
...
+    display: none;
```
querkmachine, line 83: Also curious about whether to stick `!important` here or not. 

The specificity seems high enough to not need it, assuming pages are structured how we like 'em, but this does feel like one of those 'absolutely override this in all circumstances' situations that `!important` exists specifically for.

Might want to ensure that it works with our display override classes (`.govuk-!-display-display`, `.govuk-!-display-inline`, etc.), though the higher specificity makes me assume it does.

src/govuk/components/exit-this-page/exit-this-page.mjs
```diff
@@ -225,6 +232,8 @@ ExitThisPage.prototype.resetEscTimer = function () {
  */
 ExitThisPage.prototype.resetPage = function () {
   // If an overlay is set, remove it and reset the value
+  document.body.className = document.body.className.replace('govuk-exit-this-page-hide-content', '')
```
querkmachine, line 235: We're already polyfilling `classList` at the top of the file. Is there a reason not to use it here? 

```suggestion
  document.body.classList.remove('govuk-exit-this-page-hide-content')
```

querkmachine (approved):
A couple of non-blocking comments. Happy with everything else!

I'm inclined to say let's `!important` it.

src/govuk/components/exit-this-page/exit-this-page.mjs
```diff
@@ -117,10 +117,19 @@ ExitThisPage.prototype.exitPage = function (e) {
     redirectUrl = e.target.href
   }
 
+  this.$updateSpan.innerText = ''
+
   // Blank the page
+  // As well as creating an overlay with text, we also set the body to hidden
+  // to prevent screen reader and sequential navigation users potentially
+  // navigating through the page behind the overlay during loading
+  document.body.className = ((document.body.className) ? document.body.className + ' govuk-exit-this-page-hide-content' : 'govuk-exit-this-page-hide-content')
```
querkmachine, line 126: Ditto, re: `classList`.

```suggestion
  document.body.classList.add('govuk-exit-this-page-hide-content')
```

src/govuk/components/exit-this-page/exit-this-page.mjs
```diff
@@ -117,10 +117,19 @@ ExitThisPage.prototype.exitPage = function (e) {
...
   this.$overlay = document.createElement('div')
   this.$overlay.className = 'govuk-exit-this-page__overlay'
```
querkmachine, line 128: Do we want to change the name of this class to `govuk-exit-this-page-overlay`, as we seem to have started doing that for non-nested-but-related class names?

davidc-gds:
> Okie dokie, I've fixed the navigation under the overlay issue by adding a class to the body which sets everything to display none. From some light testing, I don't think this requires an `!important` but I'm not 100% confident. Interested in specific takes on that by @querkmachine and @davidc-gds.

Heya! I've read through the comments here and this is the only piece that seems unresolved: to `!important` or !`!important` (haha a code joke mixed with Hamlet).

I agree with @querkmachine that this is kind of the situation for which `!important` was designed.

`!important` feels appropriate for a few reasons:

- We definitely don't want anything getting through the loading overly and remaining active. Like, there's no exceptions that would possibly be a good idea to allow.
- If a specificity war breaks out accidentally on some service somewhere, it's going to be very hard to identify. The screen persists for only a moment on fast connections, so the likelihood of anyone noticing that portions of the content are still interactive after the overlay is applied is pretty much nil.
- Adding `!important` here provides contextual value to developers: don't override this thing, whether by accident or intentionally.

src/govuk/components/exit-this-page/exit-this-page.mjs
owenatgov: Agree. I remember I requested we BEMify that name when you first implemented it but I acknowledge that was silly and we should be consistent with how we treat non-nested-but-related stuff so that we're following BEM properly. I'll sort it out.

src/govuk/components/exit-this-page/_index.scss
owenatgov: I'm not either _but_ I've scooped it out for now until we move to v5. Good spot.

src/govuk/components/exit-this-page/_index.scss
owenatgov: From a bit of testing I've found the following I've found that we _do_ need `!important` when we're on a page with display override classes and that the specificity overrides the competing `!important` statements. It makes sense for it to be here and doesn't have any other costs that I can see so it's going in.

Katrina-Birch:
Going back to Ciandelle and Calvin's comments about the wording. Whilst 'exiting this page' and 'loading' are two different messages, and the former is a more accurate description of what's happening, I err on the side of 'loading page' to avoid any association with Exit this Page.

Issue #3294, opened 2023-02-15
Clean up old versions of Puppeteer's browsers on CI


romaricpascal:
## What

Remove old versions of Puppeteer's browsers after installing new ones.

## Why

We cache Puppeteer's browser installs between runs and automatically download new versions. The download code provided by Puppeteer does not clean up old versions when installing new ones, making the cache grow each time Puppeteer moves to the newest version of Chrome.

Detected in https://github.com/alphagov/govuk-frontend/pull/3290#pullrequestreview-1297258225

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [ ] Old puppeteer browsers get cleaned up on CI after installing a new one

colinrotherham:
Thanks @romaricpascal. Looks like this was caused by:

* https://github.com/puppeteer/puppeteer/pull/9568

PR #3307, opened 2023-02-17
Remove outdated Chromium downloads when updating

colinrotherham:
Thanks for reporting this @romaricpascal. Fixes issue:

* https://github.com/alphagov/govuk-frontend/issues/3294


romaricpascal (approved):
Ace! Thanks for putting in a unit test as well! 🙌🏻

Issue #2693, opened 2022-07-14
Consider a default value for warningText component's iconFallbackText


joelanman:
## Related component
https://design-system.service.gov.uk/components/warning-text/

iconFallbackText is required - but 'Warning' might be a good default value, [GOV.UK Publishing Components defaults to 'Warning'](https://components.publishing.service.gov.uk/component-guide/warning_text)

## Context
Reading through the options for components for the Component Editor firebreak idea.

PR #3312, opened 2023-02-20
Add default value for warning text icon fallback attribute

owenatgov:
## What/Why
Adds the default text 'Warning' to the `iconFallbackText` option for the [warning text component](https://design-system.service.gov.uk/components/warning-text/).

As we're now setting a default value, I've also set this to no longer be a required attribute. I couldn't find any evidence in commit history that there's a risk in setting a default and not setting this as required and we use the word 'Warning' for this value in all our review app examples and our code example on the website, therefore I don't think this change is controversial.

Fixes https://github.com/alphagov/govuk-frontend/issues/2693


colinrotherham (approved):
Nice work

Clue is in the name `govukWarningText()` that "Warning" makes a sensible default 🎉

querkmachine:
Does this need a changelog entry?

PR #3319, opened 2023-02-21
Add (mandatory) type declaration checks

colinrotherham:
Follow up PR to https://github.com/alphagov/govuk-frontend/pull/3104 which makes type checking mandatory


romaricpascal (approved):
Good to go for me! We just need to hold the merge until training around types has happened 😄

colinrotherham:
Going to merge this since our JSDoc parser expects correct types already:

* https://github.com/alphagov/govuk-frontend/pull/3512

Following recent Dependabot discussions I've also included:

1. Updates to type packages
2. Updates to linting packages (JavaScript only)

We'll continue to provide guidance for type-related issues that occur. To assist, I've replaced all `@ts-nocheck` comments with a short `@ts-expect-error` description explaining the issue (typically legacy or browser-specific methods)

.eslintrc.js
```diff
@@ -69,6 +69,14 @@ module.exports = {
           'error', {
             definedTypes: ['NodeListOf']
           }
+        ],
+
+        // Maintain new line after description
+        'jsdoc/tag-lines': [
+          'warn',
+          'never', {
+            startLines: 1
```
colinrotherham, line 78: For some reason [`eslint-plugin-jsdoc`](https://www.npmjs.com/package/eslint-plugin-jsdoc) is now warning us about the new lines it (previously) wanted between JSDoc descriptions and tags

We were happy with them before so let's keep them for now

PR #3326, opened 2023-02-22
Deprecate `govuk-button--disabled` class

### LINKED ISSUE #2271: Deprecate govuk-button--disabled - ability to style link buttons as disabled
## What

Mark the `govuk-button--disabled` modifier class as deprecated, and communicate the change to users.

## Why

Adding the `--disabled` modifier class to the button does not prevent users from interacting with it. Because of the way that our CSS is currently written, the disabled link button still has a change on focus state.

Because we're trying to remove states from an element that is still interactive and focusable, automated accessibility testing tools are likely to flag issues with it, as seen in #2127. We've also seen links with the disabled modifier flagged as failing color contrast requirements in Axe.

## Further detail

At the same time, we could consider applying the hover and active states using the `:enabled` pseudo-class, rather than the current approach of 're-setting' the non-hover state for each button variant to avoid the change on hover. This would allow us to simplify the CSS, but would mean that buttons would no longer have a hover or active state in IE8.

Designers, developers

## Done when

- [x] `disabled` modifier class marked as deprecated
- [x] Deprecation communicated in release notes
- [x] Created a card to remove the modifier class in a breaking release

querkmachine:
Notes the deprecation of the `govuk-button--disabled` class in Sass and the changelog. 

Closes #2271.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,18 @@ Only use disabled form controls if research shows it makes the user interface ea
 
 This was added in [pull request #3187: Add disabled styles for form controls](https://github.com/alphagov/govuk-frontend/pull/3187).
 
+### Deprecated features
+
+#### Stop using the `govuk-button--disabled` class on buttons
+
+The `govuk-button--disabled` class is deprecated and will be removed in the next major release.
+
+If a [Button](https://design-system.service.gov.uk/components/button/) uses a `<button>` or `<input>` element, use the `disabled` HTML attribute instead. You will not need to make any changes if you're using the `govukButton` Nunjucks macro.
+
+Disabling links that are styled to look like buttons will not be supported by future releases.
+
+This was added in [pull request #3326: Deprecate `govuk-button--disabled` class](https://github.com/alphagov/govuk-frontend/pull/3326).
+
 ### Fixes
 
 We’ve made fixes to GOV.UK Frontend in the following pull requests:
--- a/src/govuk/components/button/_index.scss
+++ b/src/govuk/components/button/_index.scss
@@ -166,6 +166,8 @@ $govuk-button-text-colour: govuk-colour("white") !default;
     }
   }
 
+  // @deprecated Disabling buttons using the .govuk-button--disabled class is
+  // deprecated and will be removed in the next major version.
   .govuk-button--disabled,
   .govuk-button[disabled="disabled"],
   .govuk-button[disabled] {


CHANGELOG.md
```diff
@@ -31,6 +31,18 @@ Only use disabled form controls if research shows it makes the user interface ea
 
 This was added in [pull request #3187: Add disabled styles for form controls](https://github.com/alphagov/govuk-frontend/pull/3187).
 
+### Deprecated features
+
+#### Stop using the `govuk-button--disabled` class on buttons
+
+The `govuk-button--disabled` class is deprecated and will be removed in the next major release.
+
+If a button uses a `<button>` or `<input>` element, use the `disabled` HTML attribute instead. You will not need to make any changes if you're using the Button Nunjucks macro.
```
colinrotherham, line 40: Had a quick skim through the changelog and we tend to link to or "code up" the macro name

```suggestion
If a button uses a `<button>` or `<input>` element, use the `disabled` HTML attribute instead. You will not need to make any changes if you're using the [button](https://design-system.service.gov.uk/components/button/) Nunjucks macro.
```

```suggestion
If a button uses a `<button>` or `<input>` element, use the `disabled` HTML attribute instead. You will not need to make any changes if you're using the `govukButton` Nunjucks macro.
```

package/govuk/components/button/_index.scss
```diff
@@ -166,6 +166,8 @@ $govuk-button-text-colour: govuk-colour("white") !default;
     }
   }
 
+  // @deprecated Disabling buttons using the .govuk-button--disabled class is 
+  // deprecated and will be removed in the next major version.
```
colinrotherham, line 170: Is it a SassDoc thing where we always have a space below `/// @deprecated` comments?

I've also only seen deprecated mixins or variables so far, how will this show up?

```suggestion
  // deprecated and will be removed in the next major version.

```

colinrotherham: Do we need to worry about the other selectors being deprecated too?

```css
.govuk-button[disabled="disabled"]
.govuk-button[disabled]
```

Shall we edit ~`package`~ `src` instead and see if the SassDoc output changes?

querkmachine: D'oh. Yes, this should be the `src` version of the file rather than `package`. Will resolve.

For deprecating the class, I've followed the format specified in [the managing change documentation](https://github.com/alphagov/govuk-frontend/blob/main/docs/contributing/managing-change.md#deprecating-a-css-class), which gives the `// @deprecated` format. 

I assume this is intending to copy SassDoc's style without it strictly being SassDoc, so it's easier to find deprecation comments of all types by only searching for '@‍deprecated'.

As this isn't SassDoc in the strict sense, I figured explicitly naming the class in the comment would be enough to indicate that the `[disabled]` variants shouldn't be removed.

CHANGELOG.md
querkmachine: I've modified this a little so that it's the Button documentation link in the first instance of the word, and the name of the macro in the second instance.

36degrees:
Yeah, I think so – this isn't SassDoc as SassDoc [isn't designed to be used to document CSS](http://sassdoc.com/frequently-asked-questions/#does-sassdoc-support-css-classes-and-ids-).

Appreciate it's confusing – open to suggestions if there's a better way of doing it.

colinrotherham:
> Yeah, I think so – this isn't SassDoc as SassDoc [isn't designed to be used to document CSS](http://sassdoc.com/frequently-asked-questions/#does-sassdoc-support-css-classes-and-ids-).
> 
> Appreciate it's confusing – open to suggestions if there's a better way of doing it.

@36degrees Brill, no just checking

Sass syntax highlighting kicks in at triple-slash so it seemed like the sensible answer. We don't have new lines below comments in those docs (but do across our Sass files) so perhaps that's the only inconsistency?

Issue #3317, opened 2023-02-21
Select Component invalid test fixture scenario for multiple selected options and size attribute support


mgladdish:
## Description of the issue

I'm implementing a test harness around component fixtures.jsons and run in to an issue with a scenario for the Select component.

## Actual vs expected behaviour

The 'with selected value' scenario has one item flagged as selected and also has another value, er, value. The expected select html fragment looks like this:

```
  <select class="govuk-select" id="select-1" name="select-1">
      <option value="1">GOV.UK frontend option 1</option>
      <option value="2" selected>GOV.UK frontend option 2</option>
      <option value="3" selected disabled>GOV.UK frontend option 3</option>
  </select>
```

But isn't a Select element only able to have more than one selected option if its `multiple` attribute is set?

I'd expect the select element to be more like:

```
<select class="govuk-select" id="select-1" name="select-1"  multiple size="2" >
...
```

But then a second issue appears in that the Select box seems to ignore the size attribute and renders single-height regardless of its value. This makes it pretty much unusable as there's no good way to see the choices to select any of them other than the first in the list. (Tested on Firefox and Chrome on MacOS and Firefox on Fedora)

## Environment (where applicable)

- Operating system:
- Browser:
- Browser version:
- GOV.UK Frontend Version: 4.5.0

querkmachine:
Thanks for raising this! This appears to be an issue with the data that generates this fixture. 

We don't support `select[multiple]` in Frontend due to there being a history of poor usability and assistive technology support for it, so only one of these should have the `selected` attribute.

mgladdish:
Thanks.

Any chance you could also add a note to the component's docs to say the multiple attribute is not supported? I spent a chunk of time trying to see what I'd done wrong when adding it didn't work.

querkmachine:
A fix for the fixtures has been merged and will go out in the next release. 

I've opened a separate issue for any documentation changes. Feel free to comment on it if I've missed anything you think should be included: https://github.com/alphagov/govuk-design-system/issues/2625

Thanks again for bringing this to our attention!

PR #3333, opened 2023-02-23
Fix select component fixture having two selected values

querkmachine:
The Select component's "with selected value" fixture/example had two values that were selected: one at the item level and one at the component level.

Based on [the related test](https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/components/select/template.test.js#L62-L67), only item 2 is expected to be selected, so I've changed the fixture to only select item 2.

This is based on the assumption that having two items selected wasn't a purposeful decision — i.e. to see if the two methods of selecting a value can be used in tandem, but given there isn't a test for this, it doesn't seem like that was the intention. 

Raised in #3317.

### DIFF
--- a/src/govuk/components/select/select.yaml
+++ b/src/govuk/components/select/select.yaml
@@ -111,11 +111,10 @@ examples:
           text: GOV.UK frontend option 1
         - value: 2
           text: GOV.UK frontend option 2
-          selected: true
         - value: 3
           text: GOV.UK frontend option 3
           disabled: true
-      value: 3
+      value: 2
   - name: with hint text and error message
     data:
       id: select-2


36degrees (approved):
Whoops! Good catch 👍🏻

colinrotherham:
@querkmachine @36degrees Not necessarily in scope, but should `selected` here be overridden by `value`?

querkmachine:
@colinrotherham Having `value` take priority makes sense to me if we want to Highlander it ("there can only be one!") 

`value` is a newer addition, which might be why it's more 'passive' than it could be, to avoid potentially breaking changes:?

Issue #3182, opened 2023-01-18
Accordion: Have the option to not restore accordion state


querkmachine:
## Related component
Accordion

## Context
Currently, in browsers that support `sessionStorage` and have it enabled, the Accordion component will remember the open and closed state of each section and will restore this state when the page is visited in future, provided the browser hasn't been closed and `sessionStorage` hasn't been otherwise cleared.

## Alternatives
Currently there are two ways around this:
- Programatically clear `sessionStorage` using JavaScript on page load.
- Randomise the `id` of the accordion, as this affects the key that the state is stored against.

Setting the `expanded` parameter on each section doesn't work, as this only affects the accordion's initial state on first visit, not the state stored in `sessionStorage`.

## Additional information
There have been a few occassions on Slack and in [the backlog issue](https://github.com/alphagov/govuk-design-system-backlog/issues/1) where developers have not wanted this functionality within their service. 

Some developers have also encountered issues where multiple accordions use the same `id`, and the state from one accordion is transposed onto another. 

Now that we have a relatively well-defined methodology for passing configuration options to components, we may want to provide an option to disable this state-remembering functionality.

querkmachine:
Whatever we do here might also want to be mirrored for the Tabs component, as that currently lacks state restoration despite some demand for it:

- #1976

CharlotteDowns:
I think it's also [mentioned here in the Accordion backlog issue](https://github.com/alphagov/govuk-design-system-backlog/issues/1#issuecomment-1110872575)

> Hi,
> Sorry if this is the wrong place to write this issue I am having. I am using this accordion component and having issues with it remembering expanding selection between different sessions.
> Example, select "show all sections" clicked on a deal. Insert new deal and go to page with accordion on a different url-> all sections are expanded. Press "hide all sections" and insert a new deal and navigate to it on a different url and all sections are closed. Repeat.
I have set expanded: false in items and have kept it as false to try and force it to be closed but it doesn't seem to work.

querkmachine:
@davidc-gds raised an interesting point today that this is basically allowing for two different behavioural modes for accordions, which might be confusing to users as no indication of whether their state will be remembered or not is given.

Whilst I understand this point (and consistency is king both in UX and for code maintenance purposes!), I don't think this is a blocker to introducing this change. 

[...]

PR #3342, opened 2023-02-27
Add option to disable sessionState in Accordion

querkmachine:
Adds a new configuration option to the Accordion component, allowing users to explicitly disable (or enable) the session storage and restoration functionality.

Some developers have expressed a desire to disable this function, which currently can only be done by separately clearing `sessionStorage` or randomising the `id` of accordions, as described in #3182.

## Option naming

My understanding of good naming is that boolean options, like checkboxes, should always be labelled using affirmative, opt-in kind of language: yes, remember my password; yes, allow analytics cookies; yes, I want disposable cutlery; and so forth.

As such, the parameter is named affirmatively—`rememberExpanded` in JS and Nunjucks APIs and `data-remember-expanded` in the HTML API.

I opted for this name as it's shorter and more self-explanatory (or maybe, less technical-sounding) than others I thought of (e.g. `rememberExpandedSections`, `rememberSectionState`, `enableStateRemembering`), but I'm not tied to it and am open to suggestions. It also feels more congruent with the existing `expanded` option available to each section.

However, this means that the default value for this setting is `true` (enabled) with the typical use case being developers setting it to `false` (disabled), which might be a bit unintuitive. 

## Usage

Like internationalisation options, the configuration can be provided via JavaScript initialisation, in HTML using a `data-*` attribute, or via a parameter on the `govukAccordion` Nunjucks macro.

JavaScript example:

```js
new window.GOVUKFrontend.Accordion(document.getElementById('accordion'), {
  rememberExpanded: false
}).init()
```

Nunjucks example:
```nunjucks
{{ govukAccordion({
  id: "accordion",
  rememberExpanded: false,
  items: […]
)} }}
```

HTML example:
```html
<div class="govuk-accordion" data-module="govuk-accordion" id="accordion" data-remember-expanded="false">
 […]
</div>
```

## Changes

- The Accordion's default configuration has a new default: `rememberExpanded: true`.
- The `setInitialState` and `storeState` methods of the Accordion's JS now check that `rememberExpanded` is `true` in addition to ensuring that `sessionState` is available.
- The Accordion template now looks for and uses `rememberExpanded` if it's provided. 
- Added documentation for the `rememberExpanded` Nunjucks parameter. Also tweaked the `id` parameter documentation.
- Added example and tests for the existence of the `data-*` attribute in the template and for the functionality itself.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,16 @@ Only use disabled form controls if research shows it makes the user interface ea
 
 This was added in [pull request #3187: Add disabled styles for form controls](https://github.com/alphagov/govuk-frontend/pull/3187).
 
+#### Configure whether the Accordion remembers and restores sessions
+
+By default, when a user leaves a page, the [Accordion](https://design-system.service.gov.uk/components/accordion/) will remember the layout of expanded and collapsed sections selected by the user. If the user returns to the page, this layout will be restored and override any sections manually set as `expanded` in code.
+
+You can now disable this functionality by using the `rememberExpanded` option in the `govukAccordion` Nunjucks macro.
+
+If you're not using the Nunjucks macro, you can disable it using the `data-remember-expanded` HTML attribute.
+
+This was added in [pull request #3342: Add option to disable sessionState in Accordion](https://github.com/alphagov/govuk-frontend/pull/3342).
+
 ### Deprecated features
 
 #### Stop using the `govuk-button--disabled` class on buttons
--- a/src/govuk/components/accordion/accordion.mjs
+++ b/src/govuk/components/accordion/accordion.mjs
@@ -49,8 +49,8 @@ function Accordion ($module, config) {
   this.$module = $module
 
   var defaultConfig = {
-    rememberExpanded: true,
-    i18n: ACCORDION_TRANSLATIONS
+    i18n: ACCORDION_TRANSLATIONS,
+    rememberExpanded: true
   }
 
   /** @type {AccordionConfig} */
@@ -531,8 +531,8 @@ export default Accordion
  *
  * @typedef {object} AccordionConfig
  * @property {AccordionTranslations} [i18n = ACCORDION_TRANSLATIONS] - See constant {@link ACCORDION_TRANSLATIONS}
- * @property {boolean} [rememberExpanded] - Whether to remember the expanded and
- *   collapsed state of each section is remembered and restored when navigating.
+ * @property {boolean} [rememberExpanded] - Whether the expanded and collapsed
+ *   state of each section is remembered and restored when navigating.
  */
 
 /**
--- a/src/govuk/components/accordion/accordion.mjs
+++ b/src/govuk/components/accordion/accordion.mjs
@@ -49,6 +49,7 @@ function Accordion ($module, config) {
   this.$module = $module
 
   var defaultConfig = {
+    rememberExpanded: true,
     i18n: ACCORDION_TRANSLATIONS
   }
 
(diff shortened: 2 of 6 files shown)


colinrotherham (approved):
This looks ace 🙌 

No code comments to add

I'm perhaps torn on whether `.restoreExpanded()`  would be better?

But I've had had `jest.restoreAllMocks()` on my mind

See what @claireashworth thinks

CHANGELOG.md
```diff
@@ -31,6 +31,16 @@ Only use disabled form controls if research shows it makes the user interface ea
 
 This was added in [pull request #3187: Add disabled styles for form controls](https://github.com/alphagov/govuk-frontend/pull/3187).
 
+#### Stop the Accordion from remembering and restoring sessions
+
+By default, the [Accordion](https://design-system.service.gov.uk/components/accordion/) will remember which sections a user has expanded and collapsed upon leaving a page. If the user later returns to the page, this layout will be restored, overriding any sections that have been manually set as `expanded` in code.
```
claireashworth, line 36:
```suggestion
By default, when a user leaves a page, the [Accordion](https://design-system.service.gov.uk/components/accordion/) will remember the layout of expanded and collapsed sections selected by the user. If the user returns to the page, this layout will be restored and override any sections manually set as `expanded` in code.
```

CHANGELOG.md
```diff
@@ -31,6 +31,16 @@ Only use disabled form controls if research shows it makes the user interface ea
...
+
+You can now disable this functionality by using the `rememberExpanded` option in the `govukAccordion` Nunjucks macro.
```
claireashworth, line 38: Do we need to add a link to the relevant nunjucks macro?

querkmachine: I don't think so. The user-facing version of the Nunjucks macro code is already part of the design system page linked in the previous paragraph.

claireashworth: ok great

CHANGELOG.md
```diff
@@ -31,6 +31,16 @@ Only use disabled form controls if research shows it makes the user interface ea
 
 This was added in [pull request #3187: Add disabled styles for form controls](https://github.com/alphagov/govuk-frontend/pull/3187).
 
+#### Stop the Accordion from remembering and restoring sessions
```
36degrees, line 34: A thought – we might want to reword this as we already have a few other entries in the changelog that start with stop where we _want_ people to stop doing something, but in this instance we're just _allowing_ people to disable it where it's not desired behaviour:

- [We want you to] stop using the deprecated IE8 mixins and settings
- [We want you to] stop using the `govuk-button--disabled` class on buttons
- [You now have the option to] stop the Accordion from remembering and restoring sessions

Maybe something like 'Configure whether the Accordion remembers and restores sessions'?

claireashworth (approved):

CHANGELOG.md
querkmachine: Makes sense to me! 

I'm working under the assumption that the changelog is going to get tidied up a bit pre-release anyway, but no harm in making things clearer now.

querkmachine:
There doesn't seem to be any strong opinions regarding the parameter name, so I'm going to keep it as-is for now. We still have a little bit of leeway before release to change it if we want to.

querkmachine:
I'm now having second thoughts on the parameter naming. Having noticed that many other options in Frontend don't follow the 'affirmative language' guideline—like the Button's`preventDoubleClick` and Error Summary's `disableAutoFocus`—it may be more consistent to reverse the logic of this one too.

Issue #3349, opened 2023-02-28
Reduce frequency of Dependabot PRs


romaricpascal:
## What

Reduce the frequency of Dependabot PRs to weekly (excluding security updates, which should still happen as quickly as possible)

## Why

The current frequency means we merge Dependabot PRs every day, sometimes for very minor amends to our tooling. These updates are not that urgent and take our attention each day, so we're going to try batching them once a week.

This may also help share the load across the team as these PRs often get reviewed and merged early in the morning by the same people.

## Who needs to work on this?

Developers

## Who needs to review this?

Developers

## Done when

- [ ] Dependabot frequency is set to 1 week

Issue #2960, opened 2022-11-03
Update all dependencies using Dependabot


colinrotherham:
## What

Consider using Dependabot's `dependabot.yml` config to set up:

1. Update PRs for all dependencies (npm, GitHub Actions, Ruby gems etc)
2. Document temporarily out-of-date packages we can't upgrade yet
3. Document legacy packages that we can't upgrade

## Why

An action to investigate automatic dependency updates was raised at our Design System retro

Primarily to avoid lengthy "catch up" PRs like https://github.com/alphagov/govuk-frontend/pull/2937, but also to catch breaking changes early (little and often) as it gets much more difficult as time passes like https://github.com/alphagov/govuk-frontend/pull/2924

Benefits include:

1. Saves time not managing updates
2. Raises awareness of available bug fixes, deprecations
4. Updates to child dependencies can [unblock security updates](https://github.com/alphagov/govuk-frontend/security/dependabot)
5. Maintains pace with the Frontend community 😩

Downsides include added code review time (although in theory we're still up on time saved)

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [x] Dependabot `dependabot.yml` config added to project
- [ ] Decision made on how often to check for updates (daily, weekly)
- [x] Known packages with "worrying updates" configured to be ignored
- [x] Known package acceptable semver ranges configured in `package.json`
- [ ] ~Pull requests labelled [`dependencies`](https://github.com/alphagov/govuk-frontend/labels/dependencies) added to [**Design System Sprint Board**](https://github.com/orgs/alphagov/projects/4)~

**Update:** We'll need to our "classic" project if we want to automate moving Dependabot PRs onto the board via: https://github.com/marketplace/actions/add-to-github-projects

Optionally we may want to expand the [**Diff changes to dist**](https://github.com/alphagov/govuk-frontend/actions/workflows/diff-change-to-dist.yaml) GitHub Action workflow to run against dependency PRs

colinrotherham:
Continuing to improve our config with:

* https://github.com/alphagov/govuk-frontend/pull/2992

colinrotherham:
@36degrees Think we just need to decide if we're going with "weekly" or "daily" then this is done?

36degrees:
Let's go with daily for now and we can always change it if we find it doesn't work for us.

PR #3350, opened 2023-02-28
Reduce Dependabot updates frequency to weekly

romaricpascal:
Reduces Dependabot updates frequency so it runs weekly on Monday (default) at 10:30 (some time in the day rather than before everyone is in, in the hopes to spread who picks these up a little more). 

This change of frequency shouldn't affect when Dependabot raises security updates (according to this [bit of Dependabot's documentation](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/about-dependabot-version-updates#frequency-of-dependabot-pull-requests))

Closes [#3349](https://github.com/alphagov/govuk-frontend/issues/3349)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,7 +7,11 @@ updates:
     reviewers:
       - alphagov/design-system-developers
     schedule:
-      interval: daily
+      # Defaults to weekly on Monday
+      interval: weekly
+      time: '10:30'
+      # Setting a timezone so we let dependabot worry about BST
+      timezone: 'Europe/London'
     versioning-strategy: increase
 
     allow:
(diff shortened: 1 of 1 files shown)


colinrotherham (approved):
Looks good ✅ 

Shall we see if it works then roll out to the other repos we configured? https://github.com/alphagov/govuk-frontend/issues/2960

romaricpascal:
Cheers for the quick review 😄 

> Shall we see if it works then roll out to the other repos we configured? https://github.com/alphagov/govuk-frontend/issues/2960

I think I'll quickly make an extra PR on those so it's easier to judge the impact of changing to weekly (otherwise we'd still have daily ones popping up, even though they seem slightly less frequent on the design-system and docs repo... but it may just be an impression).

PR #3353, opened 2023-03-02
Integrate I18n into Exit this Page component

querkmachine:
Integrates internationalisation functions into the Exit this Page component. Pulled out from #3268. 

## Changes
- Integrates the `I18n` JS functionality into EtP.
- Updates Nunjucks template to add support for new parameters and `data-*` attributes relating to i18n.
- Adds a test to ensure that `data-*` attributes are added.
- Adds a translated example of the EtP button to the review app (complete with Google Translated Welsh).
- Fixes a tiny bit of outdated parameter documentation. 

Does not include tests to ensure that the correct text is injected during each of the one press/two press/activated/expired stages, as they aren't strictly i18n related—these tests can be carried out against the English language strings.

### DIFF
--- a/src/govuk/all.mjs
+++ b/src/govuk/all.mjs
@@ -58,17 +58,17 @@ function initAll (config) {
     new ErrorSummary($errorSummary, config.errorSummary).init()
   }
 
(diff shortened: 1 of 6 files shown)


owenatgov (approved):
Looking good!

PR #3354, opened 2023-03-02
Move review app source into `./app/src`

colinrotherham:
Chipping away a bit more at:

Including:

1. Move review app to `*.mjs` ES modules
2. Move review app source into `./app/src`

**Context:** This is groundwork to keep the review app separate to "package" related tasks/configs/libs


romaricpascal (approved):
Number of files felt daunting but I think it's good to merge. Thanks for moving these around 🙌🏻

colinrotherham:
Thanks @romaricpascal

I've updated the comment on this PR:

* https://github.com/alphagov/govuk-frontend/issues/2886#issuecomment-1261027939

It's come up a few times that our build scripts do too much so this move is a first step to separating out the low-risk "Review app" things from high-risk `govuk-frontend` things

PR #3364, opened 2023-03-06
Fix conflict in ESLint rules about object shorthands

romaricpascal:
## What

Disables ESLint's object-shortand rule for our source files

## Why

[Standard's configuration requires object shorthands](https://github.com/standard/eslint-config-standard/blob/80b9734d817a9babc2d02bb30cfbc98265299a00/.eslintrc.json#L29) (via <https://eslint.org/docs/latest/rules/object-shorthand>) but eslint-plugin-x requires them unavailable as we're linting to limit to ES3 features (via <https://eslint-community.github.io/eslint-plugin-es-x/rules/no-property-shorthands.html>)

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [ ] Standard's rule has been disabled for our source files (but not the rest of the project that don't need to be ES3)

### DIFF
--- a/src/.eslintrc.js
+++ b/src/.eslintrc.js
@@ -40,6 +40,8 @@ module.exports = {
       ],
       extends: ['plugin:es-x/restrict-to-es3'],
       rules: {
+        // Object shorthands are not available in ES3
+        'object-shorthand': 'off',
         // Rollup transpiles modules to AMD export/define
         'es-x/no-modules': 'off'
       }


colinrotherham:
@romaricpascal Does this unblock another PR?

Ideally we don't need to add new rules at all

colinrotherham:
@romaricpascal Yep! I meant we can worry about adding it when we need it? Might not need it soon

romaricpascal:
I think we'll need it for any v4 development, including the support we'll have once v5 is shipped. I'd argue for bringing it in now so we don't wonder in the middle of a fix (or final new feature) for v4 why ESLint keeps complaining about both ways of writing the object.

We can happily remove the rule override when we start v5 as there'll no longer be `eslint-plugin-es-x`'s ES3 configuration.

romaricpascal:
Closing so it doesn't clog the "In review" log. We'll have a trace of it if we need the fix at some point in the future.

Issue #2711, opened 2022-07-20
Refactor file copying tasks to remove Gulp


domoscargin:
## What
Update `tasks/gulp/copy-to-destination.js` to replace Gulp with our own file copying script. There are currently two tasks in this file, one for copying source files to a destination, and one for copying the ESM source files to a destination.

## Why
We have [made the decision to move away from Gulp](https://github.com/alphagov/govuk-frontend/issues/2558) and write our own build scripts.

## Who needs to work on this
Developers

## Who needs to review this
Developers

## Done when

- [x] tasks/gulp/copy-to-destination.js no longer depends on Gulp
- [x] `heroku` and `test` npm scripts no longer rely on `taskArguments.destination`
- [x] file copying tasks no longer rely on `taskArguments.destination`
- [x] docs/contributing/tasks.md has been updated
- [x] All dependencies are as up to date as is feasible
- [x] copy-to-destination.js has been relocated to tasks/copy-to-destination.js (or similar)

colinrotherham:
Accidentally completed in: https://github.com/alphagov/govuk-frontend/commit/2051717b9c2341f5502e42cfd06abd0c0b021e63

PR #3370, opened 2023-03-09
Build tasks: Clearly show subdirectory paths

colinrotherham:
This PR helps to clearly show our build paths as part of:

* https://github.com/alphagov/govuk-frontend/issues/2711

Whilst pairing on https://github.com/alphagov/govuk-frontend/pull/3318 it wasn't obvious which areas of the repo are being copied, written or built from config paths

For example, what's the difference between these three?

1. `configPaths.components`
2. `join(configPaths.package, 'govuk/components')`
3. `join(configPaths.package, 'govuk-esm/components')`

I've ensured that only top-level paths are found in [config/**paths.js**](https://github.com/alphagov/govuk-frontend/blob/main/config/paths.js) so their locations are clear:

### Top level
```js
join(paths.root, 'config')
join(paths.root, 'node_modules')
```

### Source
```js
join(paths.src, 'govuk/assets')
join(paths.src, 'govuk/components')
```

### Review app
```js
join(paths.app, 'views')
join(paths.app, 'views/examples')
join(paths.app, 'views/full-page-examples')
join(paths.app, 'views/layouts')
```

### DIFF
--- a/src/govuk-prototype-kit/govuk-prototype-kit.config.mjs
+++ b/src/govuk-prototype-kit/govuk-prototype-kit.config.mjs
@@ -2,7 +2,7 @@ import { join } from 'path'
 
 import slash from 'slash'
 
-import configPaths from '../../config/paths.js'
+import { paths } from '../../config/index.js'
 import { filterPath, getDirectories, getListing } from '../../lib/file-helper.js'
 import { componentNameToMacroName } from '../../lib/helper-functions.js'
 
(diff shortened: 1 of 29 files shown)


src/govuk-prototype-kit/govuk-prototype-kit.config.mjs
```diff
@@ -20,7 +20,7 @@ export default async () => {
         .filter(filterPath([`${componentName}/macro.njk`]))
 
       return {
-        importFrom: slash(relative(configPaths.src, join(configPaths.components, macroPath))),
+        importFrom: slash(relative(paths.src, join(paths.src, 'govuk/components', macroPath))),
```
romaricpascal, line 23: **question** It's not your change, but isn't that equivalent to `join('govuk/components', macroPath)`?

romaricpascal (approved):
I think that makes things a little more obvious 🙌🏻 Wondering if we'd need helpers to remove the many `join(path.xyz,...)` that we do into a `path.xyz(...)` but that can be done in a further PR 😄

src/govuk-prototype-kit/govuk-prototype-kit.config.mjs
colinrotherham: @romaricpascal Spot on! Hidden away by `configPaths.components` until now

Will fix before merging, thanks again

PR #3372, opened 2023-03-09
Build tasks: Create tasks per target without `--destination`

colinrotherham:
This PR removes our build target `--destination` flag

We didn't actually use it and there's a lot of complexity including/moving/renaming files via:

```isPackage
isPackage
isPublic
isDist
isDev
```

This helps to unblock:

* https://github.com/alphagov/govuk-frontend/pull/3318

Where we'll need to build the JavaScript twice to include version numbers:

1x as Rollup AMD module bundles
1x as Rollup ES module bundles


tasks/task-arguments.mjs
```diff
@@ -1,44 +1,7 @@
...
+// Non-flag arguments as tasks
+const { _: tasks } = parser(process.argv)
 
-// Check destination flags
-export const isPackage = target === 'package'
-export const isPublic = target === 'public'
-export const isDist = target === 'dist'
-export const isDev = isPublic && tasks.includes('dev')
+// Check for development task
+export const isDev = tasks.includes('dev')
```
domoscargin: Ha, neat!

docs/contributing/tasks.md
```diff
@@ -34,7 +34,7 @@ npm scripts are defined in `package.json`. These trigger a number of Gulp tasks.
 **`npm run build:app` will do the following:**
 
 - clean the `./public` folder
-- output files into `./public`, or another location via the `--destination` flag
```
domoscargin, line 37: I don't do this personally, but is it worth checking whether anybody else makes use of `--destination` to, say, build to a temp directory while developing?

docs/contributing/tasks.md
```diff
@@ -63,9 +63,9 @@ npm scripts are defined in `package.json`. These trigger a number of Gulp tasks.
 
 ## Gulp tasks
 
-Gulp tasks are defined in `gulpfile.js` and .`/tasks/gulp/` folder.
+Gulp tasks are defined in `gulpfile.mjs` and .`/tasks/` folder.
```
domoscargin, line 66: Am I missing something about why the full stop is outside of the `<code>` element here?

Also, can we tweak it to either "and the .`/tasks/` folder" or "and .`/tasks/`"?

package.json
```diff
@@ -55,7 +55,6 @@
     "gulp": "^4.0.2",
     "gulp-cli": "^2.3.0",
     "gulp-rename": "^2.0.0",
-    "gulp-task-listing": "^1.1.1",
```
domoscargin, line 58: 🔥

gulpfile.mjs
```diff
@@ -47,130 +44,11 @@ gulp.task('styles', gulp.series(
...
-      return join(file.dir, `${file.name.replace(/^all/, pkg.name)}-${pkg.version}.min.css`)
-    }
-  }),
-
-  // Update GOV.UK Frontend version
-  version('VERSION.txt', {
-    destPath: paths.dist
-  })
-))
+gulp.task('build:app', build.app())
```
domoscargin, line 46: So satisfying!

domoscargin (commented):
Think this looks great - to my eye this simplifies these workflows a bunch. And thank you for a really easy-to-follow set of commits - despite quite a large overall codebase change.

Coupla questions in addition to a few small separate comments:
1. I haven't dug into the tests - are we confident they'd detect any issues with this approach? Anything we can take the opportunity to shore up?
2. This is a fairly big PR. Feels like we could maybe separate out the `gulp-task-listing` (and maybe the browser tasks?) as they don't seem to rely on `--destination`?

docs/contributing/tasks.md
colinrotherham: Hmm how strange, I didn't even notice. It's been in since 2017 https://github.com/alphagov/govuk-frontend/commit/2e64f5b7ca40df032b7d4611a2121553dadbca5e#diff-06cac645496a5082a65db61580c0d7b2d3890e8d164ca7f4d1193e8a28f76a68R92

To avoid lots of rebasing, mind if we add a follow up typo PR?

colinrotherham:
Thanks for the review @domoscargin

I've dropped a few commits from this PR branch and moved them to:

* https://github.com/alphagov/govuk-frontend/pull/3402
* https://github.com/alphagov/govuk-frontend/pull/3403
* https://github.com/alphagov/govuk-frontend/pull/3401

docs/contributing/tasks.md
colinrotherham: Looking good for removal so far

Split out from this PR now, but I'm happy with `destPath` being configurable per destination:

* https://github.com/alphagov/govuk-frontend/pull/3403

domoscargin: I think if nobody's doing anything other than standard builds anyway, let's not bother. We can always pop it back in quick enough if folks complain.

package.json
colinrotherham: Moved to PR:

* https://github.com/alphagov/govuk-frontend/pull/3402

gulpfile.mjs
colinrotherham: Split out into PR https://github.com/alphagov/govuk-frontend/pull/3403

docs/contributing/tasks.md
colinrotherham: @domoscargin Full stop removal included in https://github.com/alphagov/govuk-frontend/pull/3384/commits/edeb6d705ff0c6627a8d500e1f63c88d1c96e89b

PR #3373, opened 2023-03-10
Ensure Nunjucks global `markdown()` renders correctly

NickColley:
Go to https://govuk-frontend-review.herokuapp.com/full-page-examples, it blows up.

After this change, no blow up.

This broke when updating marked to version 4 where the parse method was introduced:

https://github.com/markedjs/marked/releases/tag/v4.0.0

### DIFF
--- a/app/app.test.js
+++ b/app/app.test.js
@@ -12,7 +12,9 @@ const expectedPages = [
   '/examples/links',
   '/examples/typography',
   '/examples/template-default',
-  '/examples/template-custom'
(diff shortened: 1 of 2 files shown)


colinrotherham (approved):
Well spotted 🙌

Thanks for adding the page tests, they would have stopped Dependabot sneaking through the breaking change in https://github.com/alphagov/govuk-frontend/pull/2499

owenatgov (approved):
Approval part 2. It not explode anymore!

PR #3380, opened 2023-03-13
Build tasks: Fix various JSDoc types

colinrotherham:
This PR fixes various JSDoc types that have drifted due to:

1. Dependabot updates (type changes were missed)
2. Build tool changes (e.g. `node-sass` now removed)
3. Build tool files and tests excluded from `tsconfig.json`

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,9 +80,9 @@
     "@percy/cli": "^1.20.0",
     "@percy/puppeteer": "^2.0.2",
     "@percy/sdk-utils": "^1.20.0",
-    "@types/node": "^18.14.2",
-    "@typescript-eslint/eslint-plugin": "^5.54.0",
-    "@typescript-eslint/parser": "^5.54.0",
+    "@types/node": "^18.15.3",
+    "@typescript-eslint/eslint-plugin": "^5.55.0",
+    "@typescript-eslint/parser": "^5.55.0",
     "cheerio": "^1.0.0-rc.12",
     "editorconfig-checker": "^5.0.1",
     "eslint": "^8.35.0",
@@ -110,8 +110,8 @@
     "stylelint": "^14.16.1",
     "stylelint-config-gds": "^0.2.0",
     "stylelint-order": "^6.0.2",
-    "typed-query-selector": "^2.9.1",
-    "typescript": "^4.9.5",
+    "typed-query-selector": "^2.9.2",
+    "typescript": "^5.0.2",
     "vinyl": "^3.0.0"
   },
   "overrides": {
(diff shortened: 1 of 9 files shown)


tasks/compile-assets.mjs
```diff
@@ -14,7 +14,7 @@ export async function writeAsset (filePath, result) {
   const writeTasks = []
 
   // Files to write
-  const code = result.code || result.css?.toString()
+  const code = 'css' in result ? result.css : result.code
```
colinrotherham, line 17: Since Dart Sass landed in https://github.com/alphagov/govuk-frontend/pull/3164 `result.css` no longer needs `toString()`

tasks/compile-assets.mjs
```diff
@@ -49,5 +49,5 @@ export async function writeAsset (filePath, result) {
  * 2. Terser minified bundle
  * 3. Sass compiler result
  *
- * @typedef {import('rollup').OutputChunk | import('terser').MinifyOutput | import('node-sass').Result} AssetOutput
+ * @typedef {import('rollup').OutputChunk | import('terser').MinifyOutput | import('postcss').Result} AssetOutput
```
colinrotherham, line 52: Missed reference to `node-sass` which should be the PostCSS transform result

package.json
```diff
@@ -111,7 +111,8 @@
     "stylelint-config-gds": "^0.2.0",
     "stylelint-order": "^6.0.2",
     "typed-query-selector": "^2.9.1",
-    "typescript": "^4.9.5"
+    "typescript": "^4.9.5",
+    "vinyl": "^3.0.0"
```
colinrotherham, line 115: We removed this one but forgot we still type component data with it:

```mjs
/**
 * @param {import('vinyl')} file
 * @returns {Promise<import('vinyl')>}
 */
```

tasks/gulp/copy-to-destination.mjs
```diff
@@ -170,7 +170,7 @@ async function generateMacroOptions (file) {
  * Parse YAML file content as JavaScript
  *
  * @param {import('vinyl')} file - Component data ${componentName}.yaml
- * @returns {Promise<{ examples?: unknown[]; params?: unknown[] }>} Component options
+ * @returns {Promise<{ examples?: Record<string, unknown>[]; params?: Record<string, unknown>[] }>} Component options
```
romaricpascal, line 173: **question** Non blocking, but what does the `Record` type represent?

colinrotherham: It's a [utility type](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type) for objects that are always keyed with the same name/value pair

romaricpascal: Oh didn't think to look in the built-in types, cheers 😄

romaricpascal (approved):

PR #3382, opened 2023-03-14
Build tasks: Split out package copy tasks

colinrotherham:
This PR removes the mystery behind the `copyFiles` task by clearly breaking it down into individual steps


romaricpascal (approved):
Nice to have the breakdown of what happens when building the package straight in the build:package task now  🙌🏻

colinrotherham:
> Nice to have the breakdown of what happens when building the package straight in the build:package task now 🙌🏻

Thanks @romaricpascal

Yeah I hadn't realised we only needed `*.{md,njk}`

Assets like fonts/images were being copied twice by two tasks

PR #3386, opened 2023-03-15
Compile ES modules using Rollup

colinrotherham:
This PR uses Rollup to compile (not copy) ES modules into **./package/govuk-esm**

It adds on to recent "Build tasks" work to unblock:

* https://github.com/alphagov/govuk-frontend/pull/3318

### DIFF
--- a/tasks/build/package.mjs
+++ b/tasks/build/package.mjs
@@ -56,9 +56,9 @@ export default gulp.series(
     })
   ),
 
-  // Copy GOV.UK Frontend JavaScript (ES modules)
-  task.name('copy:mjs', () =>
-    files.copy('**/!(*.test).mjs', {
+  // Compile GOV.UK Frontend JavaScript (ES modules)
+  task.name('compile:mjs', () =>
+    scripts.compile('!(*.test).mjs', {
       srcPath: join(paths.src, 'govuk'),
       destPath: join(paths.package, 'govuk-esm')
     })
--- a/tasks/scripts.mjs
+++ b/tasks/scripts.mjs
@@ -10,7 +10,7 @@ import { componentPathToModuleName } from '../lib/helper-functions.js'
 import { assets } from './index.mjs'
 
 /**
- * Compile JavaScript ESM to CommonJS task
+ * Compile JavaScript task
  *
  * @param {string} pattern - Minimatch pattern
  * @param {AssetEntry[1]} [options] - Asset options
@@ -30,7 +30,7 @@ export async function compile (pattern, options) {
 }
 
 /**
- * Compile JavaScript ESM to CommonJS helper
+ * Compile JavaScript helper
  *
  * @param {AssetEntry} assetEntry - Asset entry
  */
@@ -38,12 +38,26 @@ export async function compileJavaScript ([modulePath, { srcPath, destPath, fileP
   const moduleSrcPath = join(srcPath, modulePath)
   const moduleDestPath = join(destPath, filePath ? filePath(parse(modulePath)) : modulePath)
 
-  // Create Rollup bundle
-  const bundle = await rollup({
-    input: moduleSrcPath
-  })
+  // Option 1: Rollup bundle set (multiple files)
+  // - Module imports are preserved, not concatenated
+  if (moduleDestPath.endsWith('.mjs')) {
+    const bundle = await rollup({ input: [moduleSrcPath], experimentalPreserveModules: true })
 
-  // Compile JavaScript ESM to CommonJS
+    // Compile JavaScript to ES modules
+    await bundle.write({
+      dir: destPath,
+      format: 'es',
+      sourcemap: true
+    })
+
+    return
+  }
+
+  // Option 1: Rollup bundle (single file)
+  // - Universal Module Definition (UMD) bundle
+  const bundle = await rollup({ input: moduleSrcPath })
+
+  // Compile JavaScript to output format
   const bundled = await bundle[moduleDestPath.endsWith('.min.js') ? 'generate' : 'write']({
     file: moduleDestPath,
     sourcemapFile: moduleDestPath,
@@ -72,7 +86,7 @@ export async function compileJavaScript ([modulePath, { srcPath, destPath, fileP
 }
 
 /**
- * Minify JavaScript ESM to CommonJS helper
+ * Minify JavaScript helper
  *
  * @param {string} modulePath - Relative path to module
  * @param {import('rollup').OutputChunk} result - Generated bundle
--- a/docs/contributing/coding-standards/js.md
+++ b/docs/contributing/coding-standards/js.md
@@ -138,7 +138,7 @@ export function exampleHelper2 () {}
 
 You must specify the file extension when using the import keyword.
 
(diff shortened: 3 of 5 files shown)


tasks/build/package.mjs
```diff
@@ -47,8 +47,8 @@ export default async function build () {
     destPath: paths.package
   })
 
-  // Copy GOV.UK Frontend JavaScript (ES modules)
-  await files.copy('**/!(*.test).mjs', {
+  // Compile GOV.UK Frontend JavaScript (ES modules)
+  await scripts.compile('!(*.test).mjs', {
```
colinrotherham, line 51: Notice the glob pattern has the suffix `**/` removed

Rollup will automatically find `import` modules for us

tasks/scripts.mjs
```diff
@@ -41,28 +41,38 @@ export async function compileJavaScript ([modulePath, { srcPath, destPath, fileP
...
+    ? 'es'
+    : 'umd'
+
   // Create Rollup bundle
   const bundle = await rollup({
-    input: moduleSrcPath
+    input: moduleSrcPath,
+
+    // Preserve JavaScript ESM import statements
+    experimentalPreserveModules: format === 'es'
```
colinrotherham, line 54: Regarding the "experimental" option name, there are a few unsupported Rollup features that we may need to be aware of in future with modules preserved:

1. Dynamic `import()`
2. Multiple [`input` named entries](https://rollupjs.org/configuration-options/#input) in watch mode
3. Chunk optimisation
4. Tree shaking

We don't actually need these (yet) but they arrived in later Rollup releases

I've added some background info for context:

### Rollup v0.57.0
The option `{ experimentalPreserveModules }` was added in [**Rollup v0.57.0**](https://github.com/rollup/rollup/pull/1922)

This takes us from 15 March 2018 to the last IE8 release in 28 May 2018:

```console
rollup@0.57.0
rollup@0.57.1
rollup@0.58.0
rollup@0.58.1
rollup@0.58.2
rollup@0.59.0
rollup@0.59.1
rollup@0.59.2
rollup@0.59.3
rollup@0.59.4
```

### Rollup v0.58.0
It was stabilised in [Rollup v0.58.0](https://github.com/rollup/rollup/releases/tag/v0.58.0) with some notable PRs:

**Hoisting of deep imports for performance**
(e.g. Polyfills moved into all.mjs)
https://github.com/rollup/rollup/pull/2073

**Fixes for "single file" builds like ours**
https://github.com/rollup/rollup/pull/2123

**Warnings removed for tree-shaken imports** (that were exported)
https://github.com/rollup/rollup/pull/2124

### Rollup v0.59.0

With another fix in [Rollup v0.59.0](https://github.com/rollup/rollup/releases/tag/v0.59.0)

**Fixes for file paths in source maps**
https://github.com/rollup/rollup/pull/2161

### Rollup v1.0.0
The option was renamed to `{ preserveModules }` in Rollup v1.0.0

romaricpascal:
To track what was mentioned in the dev catch-up, the hoisting of deep imports may have impacts on tree-shaking, which need looking into.

In our current package, `import {Accordion} from 'govuk-frontend'` will only bring the polyfills necessary for the `Accordion`.

However with the polyfills of all components being hoisted when building `all.mjs`, it's likely polyfills that the `Accordion` doesn't need will be drop would end up in the final bundle (especially as our polyfills are not pure), so we'll need to check what happens there.

colinrotherham:
@romaricpascal That's a great point

Thankfully the hoisted imports in `all.mjs` (from this PR) continue to be tree shaken away 🎉 

Below is some testing from our [webpack example](https://github.com/alphagov/govuk-frontend/tree/main/docs/examples/webpack) using only `{ Button }` via import/require

## Bundling via `import` from copied files
File sizes for **main.min.js** with webpack resolving `package/govuk-esm/*` copied ES modules

**14KB** — Button polyfills only
```mjs
import { Button } from 'govuk-frontend'
```

```mjs
import { Button } from 'govuk-frontend/govuk-esm/all.mjs'
```

```mjs
import Button from 'govuk-frontend/govuk-esm/components/button/button.mjs'
```

## Bundling via `import` from Rollup `es` output
File sizes for **main.min.js** with webpack resolving `package/govuk-esm/*` compiled ES modules

**14KB** — Button polyfills only
```mjs
import { Button } from 'govuk-frontend'
```

```mjs
import { Button } from 'govuk-frontend/govuk-esm/all.mjs'
```

```mjs
import Button from 'govuk-frontend/govuk-esm/components/button/button.mjs'
```

## Bundling via `require()` from Rollup `umd` output
File sizes for **main.min.js** with webpack resolving `package/govuk/*` compiled UMD modules

**50KB** — All polyfills included
```cjs
const { Button } = require('govuk-frontend')
```

```cjs
const { Button } = require('govuk-frontend/govuk/all.js')
```

**12KB** — Button polyfills only
```cjs
const Button = require('govuk-frontend/govuk/components/button/button.js')
```

colinrotherham:
I've completed testing in Rollup with `npm link` on a local build of GOV.UK Frontend

Good news, JavaScript output is identical

Source map content appears to be the Rollup-transformed ES module code (with semicolons inserted etc) but this is likely the Design System `@rollup/*` plugins needing an "input source map" config to find the original source

tasks/scripts.mjs
```diff
@@ -30,20 +30,34 @@ export async function compile (pattern, options) {
...
   const moduleSrcPath = join(srcPath, modulePath)
   const moduleDestPath = join(destPath, filePath ? filePath(parse(modulePath)) : modulePath)
 
-  // Create Rollup bundle
-  const bundle = await rollup({
-    input: moduleSrcPath
-  })
+  // Option 1: Rollup bundle set (multiple files)
+  // - Module imports are preserved, not concatenated
+  if (moduleDestPath.endsWith('.mjs')) {
```
36degrees, line 43: This function seems to be doing 2 different jobs in two separate code paths – would it make sense to split it into two functions (one for generating a bundle and one that preserves modules) and decide which one to call in the `compile` function instead?

colinrotherham: Yeah could do. Only reason I didn't is that we'll just use the new code path in `rollup@3` and delete the old

In newer versions of Rollup the `{ preserveModules }` option can also output CJS/AMD

Is it a blocker?

36degrees: Gotcha, so once we've upgraded to v3 we'll be able to simplify this?

colinrotherham: Exactly. In this PR we process:

1. Entry point `*.mjs` files for ESM, top level only
2. Every non-test `*.mjs` file for UMD, recursively with `**`

```js
task.name('compile:ms', () => scripts.compile('!(*.test).mjs') // Entries all.mjs, common.mjs etc
task.name('compile:js', () => scripts.compile('**/!(*.test).mjs') // Every non-test *.mjs file
```

But in future we'll use entry points for both and let Rollup discover all child modules.

36degrees: OK, that makes sense 👍🏻 Happy to leave this for now then.

colinrotherham: We tried this in https://github.com/alphagov/govuk-frontend/issues/3462 but we'd have to output CommonJS rather than the hybrid Universal Module Definition (UMD) bundles for browsers ([see comment](https://github.com/alphagov/govuk-frontend/issues/3462#issuecomment-1544304034))

PR #3387, opened 2023-03-16
Add contribution docs on adding component configuration options

querkmachine:
I noticed yesterday that, although we've seemingly settled on a method of adding configuration options (including localisation strings) to components, we had only documented how to use the ones we added from an end user's perspective.

Here's some scratchy documentation on how to add new configuration options and guidelines for naming them.

### DIFF
--- a/docs/contributing/coding-standards/component-options.md
+++ b/docs/contributing/coding-standards/component-options.md
@@ -1,4 +1,4 @@
-# Adding JavaScript configuration options to components
+# Component configuration
 
 Configuration options allow developers to pass additional information into a component's JavaScript. This allows them to change how the component might work or appear without needing to create their own fork.
 
--- a/docs/contributing/coding-standards/component-options.md
+++ b/docs/contributing/coding-standards/component-options.md
@@ -9,9 +9,7 @@ Configuration options can be configured using Nunjucks, HTML or JavaScript, but 
 First, make sure the component class has a constructor parameter for passing in a configuration object.
 
 ```mjs
-import { GOVUKFrontendComponent } from '../../govuk-frontend-component.mjs'
-
-export class Accordion extends GOVUKFrontendComponent {
+export class Accordion {
   constructor($module, config = {}) {
     // ...
   }
@@ -21,9 +19,11 @@ export class Accordion extends GOVUKFrontendComponent {
 Within the entry class, you'll then create an object with the default configuration for the component. These are the options the component will use if no other options are provided.
 
 ```mjs
-static defaults = Object.freeze({
-  rememberExpanded: true
-})
+export class Accordion {
+  static defaults = Object.freeze({
+    rememberExpanded: true
+  })
+}
 ```
 
 Next, use the `mergeConfigs` helper to combine the default config and the configuration object. The result of the merge is assigned globally (using `this`) so that it's accessible anywhere within the component's JavaScript.
@@ -33,10 +33,14 @@ The order in which variables are written defines their priority, with objects pa
 There is no guarantee `config` will have any value at all, so it'll be `undefined`. We use an OR operator (`||`) as a safety check. If the value is `undefined`, we use an empty object instead.
 
 ```mjs
-this.config = mergeConfigs(
-  Accordion.defaults,
-  config
-)
+export class Accordion {
+  constructor($module, config = {}) {
+    this.config = mergeConfigs(
+      Accordion.defaults,
+      config
+    )
+  }
+}
 ```
 
 We can now reference the configuration option anywhere we need to in the component's JavaScript:
@@ -98,11 +102,15 @@ As we expect configuration-related `data-*` attributes to always be on the compo
 Using the `mergeConfigs` call discussed earlier in this document, update it to include `$module.dataset` as the highest priority.
 
 ```mjs
-this.config = mergeConfigs(
-  Accordion.defaults,
-  config,
-  normaliseDataset($module.dataset)
-)
+export class Accordion {
+  constructor($module, config = {}) {
+    this.config = mergeConfigs(
+      Accordion.defaults,
+      config,
+      normaliseDataset($module.dataset)
+    )
+  }
+}
 ```
 
 Here, we pass the value of `$module.dataset` through our `normaliseDataset` function. This is because attribute values in dataset are always interpreted as strings. `normaliseDataset` runs a few simple checks to convert values back to numbers or booleans where appropriate.
@@ -184,11 +192,15 @@ For example, we could namespace our `rememberExpanded` option under the `stateIn
 The `extractConfigByNamespace` JavaScript helper can be used to create an object containing _only_ the configuration options that belong to a certain namespace.
 
 ```mjs
-this.config = mergeConfigs(
-  Accordion.defaults,
-  config,
-  normaliseDataset($module.dataset)
-)
+export class Accordion {
+  constructor($module, config = {}) {
+    this.config = mergeConfigs(
+      Accordion.defaults,
+      config,
+      normaliseDataset($module.dataset)
+    )
 
-this.stateInfo = extractConfigByNamespace(this.config, 'stateInfo');
+    this.stateInfo = extractConfigByNamespace(this.config, 'stateInfo');
+  }
+}
 ```
--- a/docs/contributing/coding-standards/component-options.md
+++ b/docs/contributing/coding-standards/component-options.md
@@ -30,9 +30,11 @@ Next, use the `mergeConfigs` helper to combine the default config and the config
 
 The order in which variables are written defines their priority, with objects passed sooner being overwritten by those passed later. As we want the user's configuration to take precedence over our defaults, we list our default configuration object first.
 
-There is no guarantee `config` will have any value at all, so it'll be `undefined`. We use an OR operator (`||`) as a safety check. If the value is `undefined`, we use an empty object instead.
+There is no guarantee `config` will have any value at all, so we set the default to an empty object (`{}`) in the constructor parameters.
 
 ```mjs
+import { mergeConfigs } from '../../common/index.mjs'
+
 export class Accordion {
   constructor($module, config = {}) {
     this.config = mergeConfigs(
@@ -66,7 +68,7 @@ It's now possible to individually initialise the component with configuration op
 
 ## Allowing options to be passed through the `initAll` function
 
-Usually, teams will not be individually initialising components. Instead, GOV.UK Frontend ships with an `initAll` function, which searches the page for instances of components and automatically initialises them.
+Often, teams will not be individually initialising components. Instead, GOV.UK Frontend ships with an `initAll` function, which searches the page for instances of components and automatically initialises them.
 
 In `src/govuk/all.mjs`, update the component's `new` class to pass through a nested configuration object. The nested object should use the component's name converted to camelCase (for example, the 'Character Count' component becomes `characterCount`).
 
@@ -95,25 +97,30 @@ It's now possible to pass configuration options for your component, as well as m
 
 For convenience, we also allow for configuration options to be passed through HTML `data-*` attributes.
 
-You can find `data-*` attributes in JavaScript by looking at an element's `dataset` property. Browsers will convert the attribute names from HTML's kebab-case to more JavaScript-friendly camelCase when working with `dataset`. See ['Naming configuration options'](#naming-configuration-options) for exceptions.
+You can find `data-*` attributes in JavaScript by looking at an element's `dataset` property. Browsers will convert the attribute names from HTML's kebab-case to more JavaScript-friendly camelCase when working with `dataset`.
+
+See ['Naming configuration options'](#naming-configuration-options) for exceptions to how names are transformed.
 
 As we expect configuration-related `data-*` attributes to always be on the component's root element (the same element with the `data-module` attribute), we can access them all using `$module.dataset`.
 
 Using the `mergeConfigs` call discussed earlier in this document, update it to include `$module.dataset` as the highest priority.
 
 ```mjs
+import { mergeConfigs } from '../../common/index.mjs'
+import { normaliseDataset } from '../../common/normalise-dataset.mjs'
+
 export class Accordion {
   constructor($module, config = {}) {
     this.config = mergeConfigs(
       Accordion.defaults,
       config,
-      normaliseDataset($module.dataset)
+      normaliseDataset(Accordion, $module.dataset)
     )
   }
 }
 ```
 
-Here, we pass the value of `$module.dataset` through our `normaliseDataset` function. This is because attribute values in dataset are always interpreted as strings. `normaliseDataset` runs a few simple checks to convert values back to numbers or booleans where appropriate.
+Here, we pass the value of `$module.dataset` through our `normaliseDataset` function. This is because attribute values in dataset are always interpreted as strings. `normaliseDataset` looks at the component's configuration schema and converts values into numbers or booleans where needed.
 
 Now, in our HTML, we could pass configuration options by using the kebab-case version of the option's name.
 
@@ -127,6 +134,51 @@ Now, in our HTML, we could pass configuration options by using the kebab-case ve
 ```
 
 However, this only works for developers who are writing raw HTML. We include Nunjucks macros for each component with GOV.UK Frontend to make development easier and faster, but this also makes it harder for developers to manually alter the raw HTML. We'll add a new parameter to Nunjucks to help them out.
+
+### Adding a configuration schema
+
+Components that accept configuration using `data-*` attributes also require a schema. This schema documents what parameters a configuration object may contain and what types of value they're expected to be.
+
+Having a schema is required for the `normaliseDataset` function to work. A schema is also needed to use the `validateConfig` and `extractConfigByNamespace` functions we'll cover later on.
+
+```mjs
+export class Accordion {
+  static schema = Object.freeze({
+    properties: {
+      i18n: { type: 'object' },
+      rememberExpanded: { type: 'boolean' }
+    }
+  })
+}
+```
+
+### Validating a provided configuration against the schema
+
+You can use the `validateConfig` function to ensure that a configuration object matches the schema format.
+
+If it doesn't, you can return a `ConfigError`.
+
+```mjs
+import { mergeConfigs, validateConfig } from '../../common/index.mjs'
+import { normaliseDataset } from '../../common/normalise-dataset.mjs'
+import { ConfigError } from '../../errors/index.mjs'
+
+export class Accordion {
+  constructor($module, config = {}) {
+    this.config = mergeConfigs(
+      Accordion.defaults,
+      config,
+      normaliseDataset(Accordion, $module.dataset)
+    )
+
+    // Check that the configuration provided is valid
+    const errors = validateConfig(Accordion.schema, this.config)
+    if (errors[0]) {
+      throw new ConfigError(`Accordion: ${errors[0]}`)
+    }
+  }
+}
+```
 
 ## Adding a Nunjucks parameter
 
@@ -177,7 +229,7 @@ In our example, `rememberExpanded` becomes `data-remember-expanded`.
 
 Unlike the `data-*` attribute in HTML and our use of `dataset` in JavaScript, there is no intrinsic link between the Nunjucks parameter name and the names used elsewhere. The Nunjucks parameter can therefore be named differently, if convenient.
 
-A common case is specifying whether a parameter accepts HTML or only plain text. For example, if a configuration option's value is inserted into the page using `innerText`, you might want to name the Nunjucks parameter something like `sectionLabelText` to indicate that HTML will not be rendered.
+A common case is specifying whether a parameter accepts HTML or only plain text. For example, if a configuration option's value is inserted into the page using `innerText`, you might want to name the Nunjucks parameter something like `sectionLabelText` to indicate that HTML will not be parsed if provided.
 
 There is more guidance on naming Nunjucks parameters in [the Nunjucks API documentation](https://github.com/alphagov/govuk-frontend/blob/main/docs/contributing/coding-standards/nunjucks-api.md#naming-options).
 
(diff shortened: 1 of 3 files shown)


docs/contributing/component-js-configuration-options.md
```diff
@@ -0,0 +1,192 @@
+# Adding JavaScript configuration options to components
+
+Configuration options permit developers to pass additional information into a component's JavaScript, allowing them to change how the component might work or appear without needing to create their own fork.
```
claireashworth, line 3:
```suggestion
Configuration options allow developers to pass additional information into a component's JavaScript. This allows them to change how the component might work or appear without needing to create their own fork.
```

docs/contributing/component-js-configuration-options.md
```diff
@@ -0,0 +1,192 @@
...
+
+Configuration options straddle the line between Nunjucks, HTML and JavaScript.
```
claireashworth, line 5: What does 'straddle the line' mean in this context? Is it the Nunjucks, HTML and JavaScript all use the same configuration options or that they can recognise each others configuration options or something similar?

docs/contributing/component-js-configuration-options.md
```diff
@@ -0,0 +1,192 @@
...
+
+## Preparing the component's JavaScript
+
+First, ensure the entry function for component has a parameter for passing in a configuration object.
```
claireashworth, line 9:
```suggestion
First, make sure the entry function for the component has a parameter for passing in a configuration object.
```

docs/contributing/component-js-configuration-options.md
```diff
@@ -0,0 +1,192 @@
...
+
+Within the entry function, you'll then create an object with the default configuration for the component. These are what options will be used if no other options have been provided.
+
+```javascript
+var defaultConfig = {
+  rememberExpanded: true
+}
+```
+
+Next, use the `mergeConfigs` helper to combine the default config and the configuration object. This is assigned globally (using `this`) so that it's accessible anywhere within the component's JavaScript.
```
claireashworth, line 25: Would it work to structure this doc with numbered steps in each section?

claireashworth: Where it says 'This is assigned globally...' is it referring to the helper, the default config, the configuration object or the result of the merging done by the helper?

docs/contributing/component-js-configuration-options.md
```diff
@@ -0,0 +1,192 @@
...
+
+The order that variables are written defines their priority, with objects passed sooner being overwritten by those passed later. As we want the user's configuration to take precedence over our defaults, we list our default configuration object first.
```
claireashworth, line 27:
```suggestion
The order in which variables are written defines their priority, with objects passed sooner being overwritten by those passed later. As we want the user's configuration to take precedence over our defaults, we list our default configuration object first.
```

docs/contributing/component-js-configuration-options.md
```diff
@@ -0,0 +1,192 @@
...
+
+We also do a little safety check here. There is no guarantee that `config` will have any value at all, in which case it'll be `undefined`. We use an OR operator (`||`) to quickly check if this is the case and, if so, we use an empty object instead.
```
claireashworth, line 29:
```suggestion
There is no guarantee `config` will have any value at all, so it'll be `undefined`. We use an OR operator (`||`) as a safety check. If the value is `undefined`, we use an empty object instead.
```

docs/contributing/component-js-configuration-options.md
```diff
@@ -0,0 +1,192 @@
...
+```
+
+We can now reference the configuration option anywhere we need to in the component's JavaScript:
+
+```javascript
+this.config.rememberExpanded
+// => true
+```
+
+It's now possible to individually initialising the component with configuration options.
```
claireashworth, line 47:
```suggestion
It's now possible to individually initialise the component with configuration options.
```

docs/contributing/component-js-configuration-options.md
```diff
@@ -0,0 +1,192 @@
...
+
+Usually, teams will not be individually initialising components. Instead, GOV.UK Frontend ships with an `initAll` function, which searches the page for instances of components and automatically initialises them.
+
+In `src/govuk/all.mjs`, update the component's `new` function to pass through a nested configuration object. The nested object should use the component's name converted to camelCase (for example, the 'Character Count' component becomes `characterCount`).
+
+```javascript
+new Accordion($accordion, config.accordion).init()
+```
+
+It's now possible to pass configuration options for your component, as well as multiple other components, via the `initAll` function.
```
claireashworth, line 68:
```suggestion
It's now possible to pass configuration options for your component, as well as multiple other components, using the `initAll` function.
```

docs/contributing/component-js-configuration-options.md
```diff
@@ -0,0 +1,192 @@
...
+    characterCount: {
+      threshold: 90
+    }
+  })
+</script>
+```
+
+## Adding support for HTML `data-*` attributes
+
+For convenience, we allow for configuration options to be passed through HTML `data-*` attributes too.
```
claireashworth, line 85:
```suggestion
For convenience, we also allow for configuration options to be passed through HTML `data-*` attributes.
```

docs/contributing/component-js-configuration-options.md
```diff
@@ -0,0 +1,192 @@
...
+
+`data-*` attributes can be found in JavaScript by looking at an element's `dataset` property. Browsers will helpfully convert the attribute names from HTML's kebab-case to more JavaScript-friendly camelCase when working with `dataset`. See ['Naming configuration options'](#naming-configuration-options) for caveats.
```
claireashworth, line 87:
```suggestion
You can find `data-*` attributes in JavaScript by looking at an element's `dataset` property. Browsers will convert the attribute names from HTML's kebab-case to more JavaScript-friendly camelCase when working with `dataset`. See ['Naming configuration options'](#naming-configuration-options) for exceptions.
```

docs/contributing/component-js-configuration-options.md
```diff
@@ -0,0 +1,192 @@
...
+
+As we expect configuration-related `data-*` attributes to always be on the component's root element (the same element with the `data-module` attribute), we can access them all quickly using `$module.dataset`.
```
claireashworth, line 89:
```suggestion
As we expect configuration-related `data-*` attributes to always be on the component's root element (the same element with the `data-module` attribute), we can access them all using `$module.dataset`.
```

docs/contributing/component-js-configuration-options.md
```diff
@@ -0,0 +1,192 @@
...
+
+Let's update our `mergeConfigs` call from earlier to include `$module.dataset` as the highest priority.
```
claireashworth, line 91:
```suggestion
Using the `mergeConfigs` call discussed earlier in this document, update it to include `$module.dataset` as the highest priority.
```

docs/contributing/component-js-configuration-options.md
```diff
@@ -0,0 +1,192 @@
...
+import { normaliseDataset } from '../../common/normalise-dataset.mjs'
+
+this.config = mergeConfigs(
+  defaultConfig,
+  config || {},
+  normaliseDataset($module.dataset)
+)
+```
+
+Here, we pass it through our `normaliseDataset` function. This is because attribute values in dataset are always interpreted as being strings. `normaliseDataset` runs a few simple checks to convert values back to numbers or booleans where appropriate.
```
claireashworth, line 104:
```suggestion
Here, we pass `mergeConfigs` through our `normaliseDataset` function. This is because attribute values in dataset are always interpreted as strings. `normaliseDataset` runs a few simple checks to convert values back to numbers or booleans where appropriate.
```

docs/contributing/component-js-configuration-options.md
```diff
@@ -0,0 +1,192 @@
...
+```html
+<div
+  class="govuk-accordion"
+  data-module="govuk-accordion"
+  data-remember-expanded="false">
+  ...
+</div>
+```
+
+However, this only works for developers who are writing raw HTML. We include Nunjucks macros for each component with GOV.UK Frontend to make development easier and faster, but this also makes it harder for developers to manually alter the raw HTML. We'll need to add a new parameter to Nunjucks to help them out.
```
claireashworth, line 117:
```suggestion
However, this only works for developers who are writing raw HTML. We include Nunjucks macros for each component with GOV.UK Frontend to make development easier and faster, but this also makes it harder for developers to manually alter the raw HTML. We'll add a new parameter to Nunjucks to help them out.
```

docs/contributing/component-js-configuration-options.md
```diff
@@ -0,0 +1,192 @@
...
+
+## Adding a Nunjucks parameter
+
+Although most components support adding arbitrary attributes and values via the `attributes` parameter, not all of them do. This method is also more verbose compared to having a dedicated Nunjucks parameter.
```
claireashworth, line 121:
```suggestion
Most, but not all, components support adding arbitrary attributes and values through the `attributes` parameter. This method is also more verbose compared to having a dedicated Nunjucks parameter.
```

docs/contributing/component-js-configuration-options.md
```diff
@@ -0,0 +1,192 @@
...
+```nunjucks
+<div
+  class="govuk-accordion"
+  data-module="govuk-accordion"
+  {%- if params.rememberExpanded %} data-remember-expanded="{{ params.rememberExpanded | escape }}"{% endif %}>
+  ...
+</div>
+```
+
+The above checks for the existence of the `rememberExpanded` parameter and adds the `data-remember-expanded` attribute if it's present. It then passes in the developer's defined value, running it through [Nunjucks' native `escape` filter](https://mozilla.github.io/nunjucks/templating.html#escape-aliased-as-e) to ensure that values can't break our HTML.
```
claireashworth, line 134:
```suggestion
The above code checks for the existence of the `rememberExpanded` parameter. If the parameter is present it adds the `data-remember-expanded` attribute. It then passes in the developer's defined value, running it through [Nunjucks' native `escape` filter](https://mozilla.github.io/nunjucks/templating.html#escape-aliased-as-e) to make sure that values can't break our HTML.
```

docs/contributing/component-js-configuration-options.md
querkmachine: I guess what I meant here is that—although configuration options are only currently used for JavaScript—the manner in which they are set exists across all three of these contexts.

I was trying to set up the next few sections, basically 😅

docs/contributing/component-js-configuration-options.md
querkmachine: > Would it work to structure this doc with numbered steps in each section?

I considered this, but as some steps contain multiple paragraphs of information and code blocks, it isn't easy to semantically mark it up as a numbered list, so I didn't. 

> Where it says 'This is assigned globally...' is it referring to the helper, the default config, the configuration object or the result of the merging done by the helper?

The result of the merging done by the helper.

docs/contributing/component-js-configuration-options.md
querkmachine: This isn't quite accurate. We're passing the value of `$module.dataset` through the `normaliseDataset` function, not the value or result of `mergeConfigs`. 

The code example does this inside the `mergeConfigs` function essentially for brevity, but it doesn't strictly have to be nested in this way.

docs/contributing/component-js-configuration-options.md
```diff
@@ -0,0 +1,192 @@
...
+{% from "govuk/components/accordion/macro.njk" import govukAccordion %}
+
+{{ govukAccordion({
+  id: "accordion",
+  rememberExpanded: false,
+  items: [...]
+}) }}
+```
+
+And that's pretty much it!
```
claireashworth, line 148: Although I want this line to stay because it's cheery, I think there needs to be a 'wrap up' line. Something about the benefit of using this method perhaps?

docs/contributing/component-js-configuration-options.md
```diff
@@ -0,0 +1,192 @@
...
+
+### In JavaScript
+
+The option in JavaScript should be written in camelCase.
+
+If the option is a boolean option that accepts `true` or `false` as values, favour affirmative language when naming the option. This avoids the use of double-negatives being used to enable something. For example, `rememberExpanded: true` is easier to understand than `forgetExpanded: false`.
+
+### In HTML
+
+`data-*` attributes in HTML should use the same name as JavaScript, converted into kebab-case. This is because the `dataset` property in JavaScript automatically converts between these two formats for us.
```
claireashworth, line 160:
```suggestion
`data-*` attributes in HTML should use the same name as JavaScript, converted into kebab-case. This is because the `dataset` property in JavaScript automatically converts between these two formats.
```

docs/contributing/component-js-configuration-options.md
```diff
@@ -0,0 +1,192 @@
...
+
+> **Warning**: This automatic conversion doesn't happen for numbers, punctuation, or other non-Latin alphabet characters. For this reason, avoid using numbers, or spell them out as words instead, when naming configuration options.
```
claireashworth, line 162:
```suggestion
> **Warning**: This automatic conversion does not happen for numbers, punctuation, or other non-Latin alphabet characters. For this reason, when you're naming configuration options, avoid using numbers or spell them out as words.
```

docs/contributing/component-js-configuration-options.md
```diff
@@ -0,0 +1,192 @@
...
+
+In our example, `rememberExpanded` becomes `data-remember-expanded`.
+
+### In Nunjucks
+
+Unlike the `data-*` attribute in HTML and our use of `dataset` in JavaScript, there is no intrinsic link between the Nunjucks parameter name and the names used elsewhere. The Nunjucks parameter can therefore be named different, if convenient.
+
+A common case is specifying whether a parameter accepts HTML or only plain text. For example, if a configuration option's value is inserted into the page using `innerText`, you might want to name the Nunjucks parameter something like `sectionLabelText` to indicate that HTML won't be rendered.
```
claireashworth, line 170:
```suggestion
A common case is specifying whether a parameter accepts HTML or only plain text. For example, if a configuration option's value is inserted into the page using `innerText`, you might want to name the Nunjucks parameter something like `sectionLabelText` to indicate that HTML will not be rendered.
```

docs/contributing/component-js-configuration-options.md
```diff
@@ -0,0 +1,192 @@
...
+
+There is more guidance on naming Nunjucks parameters in [the Nunjucks API documentation](https://github.com/alphagov/govuk-frontend/blob/main/docs/contributing/coding-standards/nunjucks-api.md#naming-options).
+
+## Namespacing configuration options
+
+Configuration options in JavaScript and HTML can be grouped together by using namespaces; period-separated strings that prefix the configuration option name. Namespaces follow the same formats as other option names, being camelCase in JavaScript and kebab-case in HTML.
```
claireashworth, line 176:
```suggestion
You can group configuration options in JavaScript and HTML together by using namespaces; period-separated strings that prefix the configuration option name. Namespaces follow the same formats as other option names, being camelCase in JavaScript and kebab-case in HTML.
```

docs/contributing/component-js-configuration-options.md
```diff
@@ -0,0 +1,192 @@
...
+
+`data-*` attributes in HTML should use the same name as JavaScript, converted into kebab-case. This is because the `dataset` property in JavaScript automatically converts between these two formats for us.
+
+> **Warning**: This automatic conversion doesn't happen for numbers, punctuation, or other non-Latin alphabet characters. For this reason, avoid using numbers, or spell them out as words instead, when naming configuration options.
+
+In our example, `rememberExpanded` becomes `data-remember-expanded`.
+
+### In Nunjucks
+
+Unlike the `data-*` attribute in HTML and our use of `dataset` in JavaScript, there is no intrinsic link between the Nunjucks parameter name and the names used elsewhere. The Nunjucks parameter can therefore be named different, if convenient.
```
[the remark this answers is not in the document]
querkmachine: In our Nunjucks code and documentation, we encourage the names of prosaic Nunjucks parameters to include either 'text' or 'html' in their name to indicate how the value will be treated ('html' ones will go through an HTML parser, 'text' ones will not). 

When this new configuration API was introduced, we continued doing this for Nunjucks parameter naming. 

So for example, the Accordion uses `i18n.hideAllSections` in JavaScript, `data-i18n.hide-all-sections` in HTML, but then changes to `hideAllSectionsText` in Nunjucks.

docs/contributing/component-js-configuration-options.md
claireashworth: ok. So could we use your sentence something like this?
Currently, only JavaScript uses the configuration options, but due to the way they are written Nunjucks and HTML can use them as well.
Is that accurate?

querkmachine: It's more like the options can be configured using Nunjucks, HTML or JavaScript, but only JavaScript will use them?

docs/contributing/component-js-configuration-options.md
claireashworth: Yeah, as  worked my way down I could see that numbering would only work in some places making the whole doc rather inconsistent.

> The result of the merging done by the helper.

Thanks

claireashworth, line 25:
```suggestion
Next, use the `mergeConfigs` helper to combine the default config and the configuration object. The result of the merge is assigned globally (using `this`) so that it's accessible anywhere within the component's JavaScript.
```

docs/contributing/component-js-configuration-options.md
claireashworth: ok, so if we change 'it' to 'the value of `$module.dataset`', do we need to add further explanation or is that sufficient by itself?

querkmachine: I think that should be sufficient, yeah

claireashworth: Fab, thank you

docs/contributing/component-js-configuration-options.md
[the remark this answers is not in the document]
claireashworth: Thank you!

docs/contributing/component-js-configuration-options.md
claireashworth: Oh I see. I had it almost totally backwards! Yes, that sounds good.

docs/contributing/component-js-configuration-options.md
```diff
@@ -0,0 +1,192 @@
+# Adding JavaScript configuration options to components
+
+Configuration options allow developers to pass additional information into a component's JavaScript. This allows them to change how the component might work or appear without needing to create their own fork.
+
+Configuration options straddle the line between Nunjucks, HTML and JavaScript.
```
claireashworth, line 5:
```suggestion
Configuration options can be configured using Nunjucks, HTML or JavaScript, but only JavaScript will use the options.
```

docs/contributing/component-js-configuration-options.md
```diff
@@ -0,0 +1,192 @@
...
+      rememberExpanded: false
+    },
+    characterCount: {
+      threshold: 90
+    }
+  })
+</script>
+```
+
+## Adding support for HTML `data-*` attributes
```
colinrotherham, line 83: Mind if we link to this section from the JS coding standards?
https://github.com/alphagov/govuk-frontend/blob/main/docs/contributing/coding-standards/js.md#use-data-attributes-to-initialise-component-javascript

It's not been updated beyond `data-module`

docs/contributing/component-js-configuration-options.md
```diff
@@ -0,0 +1,192 @@
...
+
+As we expect configuration-related `data-*` attributes to always be on the component's root element (the same element with the `data-module` attribute), we can access them all using `$module.dataset`.
+
+Using the `mergeConfigs` call discussed earlier in this document, update it to include `$module.dataset` as the highest priority.
+
+```javascript
+import { mergeConfigs } from '../../common/index.mjs'
+import { normaliseDataset } from '../../common/normalise-dataset.mjs'
+
+this.config = mergeConfigs(
```
colinrotherham, line 97: Until we can use `Object.assign()` (or object spread) we should suggest types here 😩 

For example in the Character Count component, linting errors appear for:

```js
this.config.maxlength
this.config.threshold
```

Because without types, we appear to do calculations on unknown types

```suggestion
/** @type {CharacterCountConfig} */
this.config = mergeConfigs(
```

colinrotherham: Hard isn't it

Also the realisation that `normaliseDataset()` is a bit magic and will automatically picks types for you based on what the string value looks like. The compiler can't automatically "type" the return value.

Could potentially be solved by passing the defaults into `normaliseDataset()`

```js
this.config = normaliseDataset($module.dataset, {
  ...defaultConfig,
  ...config || {},
})
```

So rather than guessing types you could grab them from defaults

Not one for this PR, so yeah let's leave it

claireashworth (approved):

colinrotherham:
We should still re-visit this, it's useful

But will need to catch up with:

* https://github.com/alphagov/govuk-frontend/pull/4011

colinrotherham:
@querkmachine I've given it a rapid work-in-progress refresh, something along these lines?

romaricpascal (approved):
That's a neat piece of doc! 🙌🏻 I'd be keen to merge it as it is and add some extra information around typing the configuration in a separate PR (as we probably need to think if the info fits here or in a more generic `using-types` piece of docs).

@colinrotherham @querkmachine Do you think there's anything holding it from being merged? If not, let's press that green button 😊

colinrotherham:
@romaricpascal Just a double check over my quick edits?

Aware that lots of the examples given will fail due to missing JSDoc though, should we include it?

Will also need a big update once we fix how components set up their options after https://github.com/alphagov/govuk-frontend/issues/4230

querkmachine:
@romaricpascal It's been a while so I'm definitely not sure of how accurate parts of it are at this point, but better to have something that we can iterate upon rather than sitting on it for much longer.

docs/contributing/coding-standards/component-options.md
```diff
@@ -0,0 +1,261 @@
...
+</script>
+```
+
+## Adding support for HTML `data-*` attributes
+
+For convenience, we also allow for configuration options to be passed through HTML `data-*` attributes.
+
+You can find `data-*` attributes in JavaScript by looking at an element's `dataset` property. Browsers will convert the attribute names from HTML's kebab-case to more JavaScript-friendly camelCase when working with `dataset`.
+
+See ['Naming configuration options'](#naming-configuration-options) for exceptions to the how names are transformed.
```
claireashworth, line 102:
```suggestion
See ['Naming configuration options'](#naming-configuration-options) for exceptions to how names are transformed.
```

querkmachine:
Going to merge this on the basis that having something to iterate upon is probably better than letting this languish and grow outdated again.

PR #3390, opened 2023-03-16
Webpack example: Remove IE8 support

colinrotherham:
As an alternative to https://github.com/alphagov/govuk-frontend/pull/3389 it might be easier to drop IE8 from the webpack example to resolve:

* https://github.com/alphagov/govuk-frontend/pull/3343


colinrotherham:
Superseded by https://github.com/alphagov/govuk-frontend/pull/3390

PR #3392, opened 2023-03-17
Fix typos in overLimitText macro option names

matteason:
I noticed that the names of two options in the [character count documentation](https://design-system.service.gov.uk/components/character-count/#options-character-count-example) were wrong (probably a copy/paste error) - characters**Over**LimitText is shown as characters**Under**LimitText, and words**Over**LimitText as words**Under**LimitText:

Screenshot showing documentation of Nunjucks macro options. Two of them have the name charactersUnderLimitText

This change just fixes those typos. Didn't think it needed a changelog entry but let me know if it does.


querkmachine (approved):
Nice catch! Thanks for the fix!

romaricpascal (approved):
Indeed, cheers for spotting and fixing this 😊

PR #3401, opened 2023-03-23
Build tasks: Consolidate browser screenshot script as tasks

colinrotherham:
Split out from PR:

* https://github.com/alphagov/govuk-frontend/pull/3372

This PR consolidates the component screenshot and browser download functions into **Browser** tasks, so they can be imported as build task functions without side effects (no top-level await)

This change is in preparation for:

* https://github.com/alphagov/govuk-frontend/pull/3404

Which lets us reuse tasks in the Review app to support:


romaricpascal (approved):
Nice little tidy up as well. Makes sense to put both browser related tasks in the same file 🙌🏻

PR #3411, opened 2023-03-24
Fix for missing review app packages

colinrotherham:
This PR fixes `gulp screenshots` by moving the following packages to `dependencies`

```
@percy/puppeteer
@percy/sdk-utils
cheerio
jest-axe
outdent
puppeteer
```

The review app installs with `NODE_ENV=production npm ci` so can't use `devDependencies`

We'll aim to look at this again via:

* https://github.com/alphagov/govuk-frontend/pull/3384


lib/file-helper.js
```diff
@@ -98,6 +98,25 @@ const getComponentsData = async () => {
   return Promise.all(componentNames.map(getComponentData))
 }
 
+/**
+ * Get examples from a component's metadata file
+ *
+ * @param {string} componentName - Component name
+ * @returns {Promise<Object<string, ComponentExample['data']>>} returns object that includes all examples at once
+ */
+async function getExamples (componentName) {
```
colinrotherham, line 107: Not sure why this one wasn't alongside the other helpers like `getFullPageExamples()`

romaricpascal:
I'm a bit concerned by the need for refactoring on EtP from the changes of the nunjucks helper, separating the rendering from the Cheerio part, is it absolutely necessary for fixing the issue on the review app at the moment or does the first commit does fine on its own?

CharlotteDowns:
@romaricpascal Yep I'm happy to work with the preview :).

romaricpascal:
Oh, @CharlotteDowns sorry for the confusion, I was asking @colinrotherham if the PR could be split so we could quickly put the review app back on its feet and discuss the changes around the Nunjucks test with less urgency 😄

colinrotherham:
@romaricpascal I'm happy either way, don't mind. All of the same lines change for (future) **npm workspace** changes so can either do them now or later?

romaricpascal:
> @romaricpascal I'm happy either way, don't mind. All of the same lines change for (future) **npm workspace** changes so can either do them now or later?

In that case, I'd rather focus this PR on fixing the bug only so we can quickly ship it and then have the update of the helpers in a separate one that can be merged after we give a heads up to the Design System squad that they'll have to update their EtP spike, please 😊

colinrotherham:
Of course. I've got an EtP patch ready too 👍

I've amended this PR so it's just the dependencies moved (but a few more now)

romaricpascal (approved):

PR #3412, opened 2023-03-24
Reduce dependencies for review app

colinrotherham:
This PR removes the following `dependencies` from the Review app Heroku install:

```
cheerio
jest-axe
```

By creating a new [**lib/nunjucks-helpers.js**](https://github.com/alphagov/govuk-frontend/blob/app-test-packages/lib/nunjucks-helpers.js) from [**lib/jest-helpers.js**](https://github.com/alphagov/govuk-frontend/blob/app-test-packages/lib/jest-helpers.js)

Our Sass tests no longer initialise a Nunjucks environment too

Split from PR https://github.com/alphagov/govuk-frontend/pull/3411

PR #3413, opened 2023-03-24
Reduce dependencies for review app (EtP patch)

colinrotherham:
@querkmachine @owenatgov Just a heads up, we've split the test helpers up a bit in:

* https://github.com/alphagov/govuk-frontend/pull/3412

This PR is a patch for the EtP branch so everything carries on working


colinrotherham:
Replaced with https://github.com/alphagov/govuk-frontend/pull/3500 instead

Issue #2707, opened 2022-07-20
Refactor watch task to remove Gulp


domoscargin:
## What
Update `tasks/gulp/watch.js` to replace Gulp with our own watch script. This task is responsible for watching source code changes.

## Why
We have [made the decision to move away from Gulp](https://github.com/alphagov/govuk-frontend/issues/2558) and write our own build scripts.

## Who needs to work on this
Developers

## Who needs to review this
Developers

## Done when

- [x] tasks/gulp/watch.js no longer depends on Gulp
- [ ] ~docs/contributing/tasks.md has been updated~
- [x] watch.js has been relocated to tasks folder

domoscargin:
Since the watch task relies on the gulp tasks `styles`, `sassdoc` and `scripts`, I'm going to adjust its positioning in this work.

colinrotherham:
This issue was picked up in:

* https://github.com/alphagov/govuk-frontend/pull/3428

It's ready for review but we'll need to decide:

1. Do we _keep_ `gulp.watch()` and close this?
2. Do we _replace_ `gulp.watch()` but keep `gulp.parallel()` for logging (based on [#3384](https://github.com/alphagov/govuk-frontend/pull/3384) feedback)
2. Do we _replace_ `gulp.watch()` with a better tool (Browsersync?) versus a like-for-like replacement

domoscargin:
I'm inclined to do 1 or 3.

`gulp.watch()` is a pretty core part of Gulp's functionality so I think 2 (deftly shown in #3428) feels like a step too far.

I vote we go with 1 for now and open an issue to investigate an alternative which we can pick up if we find the solution holds us back from anything.

PR #3428, opened 2023-03-28
Refactor watch task to remove `gulp.watch()`

colinrotherham:
Closes https://github.com/alphagov/govuk-frontend/issues/2707 with most work coming from:

* https://github.com/alphagov/govuk-frontend/pull/3384

I've directly imported [chokidar](https://www.npmjs.com/package/chokidar) which `gulp.watch()` uses, but would [Browsersync](https://browsersync.io) be more useful?

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -16,6 +16,7 @@
     "start": "node src/start.mjs"
   },
   "dependencies": {
+    "chokidar": "^3.5.3",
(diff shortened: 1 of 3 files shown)


colinrotherham:
I've added another commit to this one to restore `gulp.parallel()` for CLI logging purposes like we did in:

* https://github.com/alphagov/govuk-frontend/pull/3384

PR #3439, opened 2023-03-29
Fix argument count in deprecated `this.mql.addListener()`

colinrotherham:
This PR partially reverts an unreleased change from https://github.com/alphagov/govuk-frontend/commit/0663ab3373736943e1b29fcf95473205ec382d17#diff-bcd43b95b8acddb0071fa9f557046160278bb06b55b12a58bf76b6c0a2bbb17aR57

Browser testing in Safari 9.1 correctly spotted some recent MediaQueryList argument changes:

```js
var mql = window.matchMedia('(min-width: 48.0625em)')

mql.addEventListener('change', handler)
mql.addListener(handler) // Legacy IE and Safari
```

### DIFF
--- a/src/govuk/components/header/header.mjs
+++ b/src/govuk/components/header/header.mjs
@@ -52,14 +52,15 @@ Header.prototype.init = function () {
     // Set the matchMedia to the govuk-frontend desktop breakpoint
     this.mql = window.matchMedia('(min-width: 48.0625em)')
 
-    var listenerMethod = 'addEventListener' in this.mql
-      ? 'addEventListener'
-      : 'addListener'
-
-    // addListener is a deprecated function, however addEventListener
-    // isn't supported by IE or Safari. We therefore add this in as
-    // a fallback for those browsers
-    this.mql[listenerMethod]('change', this.syncState.bind(this))
+    if ('addEventListener' in this.mql) {
+      this.mql.addEventListener('change', this.syncState.bind(this))
+    } else {
+      // addListener is a deprecated function, however addEventListener
+      // isn't supported by IE or Safari < 14. We therefore add this in as
+      // a fallback for those browsers
+      // @ts-expect-error Property 'addListener' does not exist
+      this.mql.addListener(this.syncState.bind(this))
+    }
 
     this.syncState()
     this.$menuButton.addEventListener('click', this.handleMenuButtonClick.bind(this))


src/govuk/components/header/header.mjs
```diff
@@ -52,6 +52,16 @@ Header.prototype.init = function () {
     // Set the matchMedia to the govuk-frontend desktop breakpoint
     this.mql = window.matchMedia('(min-width: 48.0625em)')
 
+    if ('addEventListener' in this.mql) {
+      this.mql.addEventListener('change', this.syncState.bind(this))
+    } else {
+      // addListener is a deprecated function, however addEventListener
+      // isn't supported by IE or Safari. We therefore add this in as
+      // a fallback for those browsers
+      // @ts-expect-error Property 'addListener' does not exist
```
colinrotherham, line 61: I've added `@ts-expect-error` as TypeScript isn't aware of any browsers that would fall into _else_

36degrees: Is that an issue with TypeScript’s data on `MediaQueryList`? Worth raising an issue or PR to fix it?

colinrotherham:
See related change in:

* https://github.com/alphagov/govuk-design-system/pull/2704

src/govuk/components/header/header.mjs
```diff
@@ -52,14 +52,15 @@ Header.prototype.init = function () {
...
-    // addListener is a deprecated function, however addEventListener
-    // isn't supported by IE or Safari. We therefore add this in as
-    // a fallback for those browsers
-    this.mql[listenerMethod]('change', this.syncState.bind(this))
+    if ('addEventListener' in this.mql) {
+      this.mql.addEventListener('change', this.syncState.bind(this))
+    } else {
+      // addListener is a deprecated function, however addEventListener
+      // isn't supported by IE or Safari. We therefore add this in as
+      // a fallback for those browsers
```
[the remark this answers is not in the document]
colinrotherham: Ahh nice, you liked the extra detail from the other repo. I'll update 🎉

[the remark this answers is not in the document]
36degrees: Hadn't spotted that yet! But yeah, think it's worth adding as currently suggests it's not supported by Safari _at all_, which felt slightly inaccurate.

src/govuk/components/header/header.mjs
36degrees: Should it not be using our browserslist, and therefore know that we need to support older versions of Safari that don't have `addEventListener`? Or is it not that clever?

src/govuk/components/header/header.mjs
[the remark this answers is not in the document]
colinrotherham: Fab, all done

src/govuk/components/header/header.mjs
36degrees: Looks like there's a related discussion here, but no activity for quite a while: https://github.com/microsoft/TypeScript/issues/32210

Wondering if there's anything we could do (maybe with ESLint?) to catch this – given how strongly TypeScript 'encourages' us to remove the `addListener` call, I can see this coming up again.

But possibly too much of a distraction for now, unless someone happens to think of something really simple we could try.

colinrotherham: Ah so the older methods aren't actually deprecated?

TypeScript gets its data from `addListener` via [`@deprecated` in **lib.dom.d.ts**](https://github.com/microsoft/TypeScript/blob/main/src/lib/dom.generated.d.ts#L10125) but added in:

* https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/550

Could just follow [**tabs.mjs** and use `addListener()` only](https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/components/tabs/tabs.mjs#L59)?

For ESLint though I've looked at [eslint-plugin-deprecation](https://github.com/gund/eslint-plugin-deprecation#readme) plugin before

36degrees: Let's stick with the current fix as is, and write up an issue to revisit how we use `MediaQueryList` later, work out which approach to standardise on then?

I think we'll want to revisit the tabs and header component JS at some point anyway, as all browsers that support `type="module"` support `window.matchMedia` so we should be able to simplify things slightly. Might be a good time to look at this then?

colinrotherham: Ah it's also marked as deprecated on [**Can I Use** "MediaQueryList API: `addListener()`"](https://caniuse.com/mdn-api_mediaquerylist_addlistener)

36degrees (approved):

Issue #2199, opened 2021-04-22
Significant line height differences between node-sass and dart-sass


andymantell:
Rendering the govuk-frontend scss with dart-sass results in surprisingly significant line height differences when compared to node-sass. This is caused by the default precision of dart-sass being 10, whereas it is 5 in node-sass. The problematic code lies here, in the calculation of line height.

```@function _govuk-line-height($line-height, $font-size) {
  @if not unitless($line-height) and unit($line-height) == unit($font-size) {
    $line-height: $line-height / $font-size;
  }

  @return $line-height;
}```

This is a screenshot of the differences registered by backstop (In this case, from nhsuk-frontend, but it's the same line height code).

![image](https://user-images.githubusercontent.com/134642/115725035-64a77b00-a379-11eb-980b-514573df800a.png)

Over on nhsuk-frontend, we are now forcibly rounding to 5 decimal places so that we get consistent rendering across node-sass and dart-sass (Sounds minor, but people are likely to be hopping between different bits of the NHS regularly so as much consistency as possible is nice). I will open a pull request shortly with our solution, for discussion...

gunjam:
Do you know if this issue affects both the dart-sass binary and the javascript version (npm sass), or just one or the other?

andymantell:
I've only tested this with the npm `sass` variant of dart-sass but I suspect it would be an issue in all variants of it. I see no reason to suspect it would only be the JavaScript compiled version?

36degrees:
From testing, it looks like 6 (or more) decimal places is required for the 'correct' computed line height in Chrome:

| Font size | Desired line height | 6+   | 5         | 4         | 3         | 2         | 1       |
|-----------|---------|------|-----------|-----------|-----------|-----------|---------|
| 80        | 80px    | 80px | 80px      | 80px      | 80px      | 80px      | 80px    |
| 48        | 50px    | 50px | 50.0002px | 50.0016px | 50.016px  | 49.92px   | 48px    |
| 36        | 40px    | 40px | 40px      | 39.9996px | 39.996px  | 39.96px   | 38.72px |
| 27        | 30px    | 30px | 30px      | 29.9997px | 30.0059px | 30.0144px | 29.92px |
| 24        | 30px    | 30px | 30px      | 30px      | 30px      | 30px      | 28.8px  |
| 19        | 25px    | 25px | 25px      | 25.0002px | 25.0145px | 25.1328px | 24.96px |
[...]

PR #3454, opened 2023-03-30
Update default link underline offset setting

### LINKED ISSUE #2743: Review the text decoration offset for links
## What

Revisit the default value for the text decoration offset on links, set by `$govuk-link-underline-offset`:

https://github.com/alphagov/govuk-frontend/blob/c9d331c2a155de6244c8fb240c22923e074a8099/src/govuk/settings/_links.scss#L39-L46

## Why

As noted in https://github.com/alphagov/govuk-frontend/pull/2183#issuecomment-831933308, when we introduced the new link styles we had to make a compromise on the offset of the underlines, as Chromium-based browsers measured the offset differently to other browsers, putting the underline further away.

This issue appears to have been fixed as of Chrome 104 (https://bugs.chromium.org/p/chromium/issues/detail?id=1251634). Once Chrome 104 reaches widespread adoption, we can revisit the value now that the behaviour in different browsers is consistent.

## Who needs to work on this

Designer

## Who needs to review this

Designer and developer?

## Done when

- [ ] Chrome 104 has decent adoption
- [ ] We've reviewed the default value for the underline offset (`$govuk-link-underline-offset`) and made any changes
- [ ] We've checked we're happy with the appearance in all supported browsers

querkmachine:
We introduced new link styles in #2183 ([GOV.UK Frontend v3.12.0](https://github.com/alphagov/govuk-frontend/releases/tag/v3.12.0)), however were forced to compromise on the underline offset due to [a bug](https://bugs.chromium.org/p/chromium/issues/detail?id=1251634) in the Chromium rendering engine. 

That bug was fixed in May 2022, was released in Chrome 104/Edge 104, and is now widely available. As such, we are updating the default link underline offset to achieve the originally intended appearance. 

Closes #2743.

### DIFF
--- a/src/govuk/settings/_links.scss
+++ b/src/govuk/settings/_links.scss
@@ -38,12 +38,16 @@ $govuk-link-underline-thickness: unquote("max(1px, .0625rem)") !default;
 
 /// Offset of link underlines from text baseline
 ///
+/// The default is 3px expressed as ems, as calculated against the default body
+/// font size (on desktop) of 19px.
+/// 3 ÷ 19 = 0.1578
+///
 /// Set this variable to `false` to avoid setting an offset.
 ///
 /// @type Number
 /// @access public
 
-$govuk-link-underline-offset: .1em !default;
+$govuk-link-underline-offset: .1578em !default;
 
 /// Thickness of link underlines in hover state
 ///
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,7 +54,7 @@ The [Button](https://design-system.service.gov.uk/components/button/) Nunjucks m
 
 This was added in [pull request #3344: Adding optional id attribute to button component](https://github.com/alphagov/govuk-frontend/pull/3344). Thanks to [Tom Billington](https://github.com/TomBillingtonUK) for this contribution.
 
-### Added a modifier for text input styles that accept sequences of digits
+#### Added a modifier for text input styles that accept sequences of digits
 
 We've added a new `.govuk-input--extra-letter-spacing` class for [Text Input](https://design-system.service.gov.uk/components/text-input/). This increases readability of text inputs that receive sequences of digits (like security codes, references or phone numbers).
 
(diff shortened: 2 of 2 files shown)


src/govuk/settings/_links.scss
```diff
@@ -43,7 +43,7 @@ $govuk-link-underline-thickness: unquote("max(1px, .0625rem)") !default;
 /// @type Number
 /// @access public
 
-$govuk-link-underline-offset: .1em !default;
+$govuk-link-underline-offset: .1578em !default;
```
colinrotherham, line 46: Shall we update the test in [**src/govuk/helpers/links.test.js**#L63](https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/helpers/links.test.js#L63)?

Probably want to follow the comment block below and explain this is 3px at font size 19

Lots of other magic numbers in this file but could show the calculation too? Not a blocker

```suggestion
$govuk-link-underline-offset: 3px / 19px * 1em !default;
```

querkmachine: I don't think the test needs updating. It seems to be testing that changes to the variable are being bubbled through to the related mixin, rather than testing what the value specifically is. 

I'm unsure of using an equation in place of a 'magic' number for a few reasons: 

- There is zero visual difference between how Chromium, WebKit or Gecko renders 0.1578em compared to 0.1578947368em—the [actual value](https://www.w3.org/TR/css-cascade-4/#actual) (sorta the 'normalised' computed value) seems to always be 3px. This remains the case even at high zoom levels. 
- As this equation doesn't evaluate to a round number, we'll end up having slightly different values depending on whether the Sass is compiled by libsass or Dart Sass, due to each compiler having different rounding precision. This could lead to incremental differences that become exacerbated in some circumstances (like #2199). Though admittedly this is unlikely due to the first bullet point, it feels like a complicating factor we could avoid. 
- The use of an equation, and the extra precision it results in, unnecessarily adds to compile times and compiled code weight (albeit, very minorly). 

Ultimately, to my mind, using an equation that always evaluates to the same result doesn't provide any tangible benefit over directly using the result, so I'd prefer to stick with what we have.

colinrotherham: Absolutely fine, it's not a blocker

You alright adding a comment to explain this is 3px at font size 19 for our future selves?

colinrotherham (approved):

PR #3481, opened 2023-04-03
Minor updates to pre-release documentation

querkmachine:
Some suggestions based on having used this part of the doc for the first time.

## Changes
- Updated link to sprint board.
- Added more info about what metadata each issue should have attached to it.
- Moved the info about metadata above the list of issues, so the reader is aware of them before starting to create new issues.

### DIFF
--- a/docs/releasing/before-publishing-a-release.md
+++ b/docs/releasing/before-publishing-a-release.md
@@ -12,7 +12,15 @@ Optionally, we can also let the prototype kit know that we are coordinating a re
 
 ## Raise release issues
 
-Those leading on the release should then raise new issues in the team GitHub repositories ([govuk-frontend](https://github.com/alphagov/govuk-frontend), [govuk-design-system](https://github.com/alphagov/govuk-design-system/issues/2024), [govuk-frontend-docs](https://github.com/alphagov/govuk-frontend-docs) to track the following:
+Release issues should be created in the team GitHub repositories: [govuk-frontend](https://github.com/alphagov/govuk-frontend), [govuk-design-system](https://github.com/alphagov/govuk-design-system/) and [govuk-frontend-docs](https://github.com/alphagov/govuk-frontend-docs).
+
+All these issues should be:
+
+- added to the [Design System sprint board](https://github.com/orgs/alphagov/projects/53) backlog
+- added to that release's milestone
+- tagged with the '🚀 release' label
+
+Those leading on the release should raise new issues to track the following:
 
 - draft comms for the new release (example issue: [#2507](https://github.com/alphagov/govuk-frontend/issues/2507))
 - create release notes for the new release (example issue: [#2508](https://github.com/alphagov/govuk-frontend/issues/2508))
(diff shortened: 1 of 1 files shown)


docs/releasing/before-publishing-a-release.md
```diff
@@ -12,7 +12,15 @@ Optionally, we can also let the prototype kit know that we are coordinating a re
 
 ## Raise release issues
 
-Those leading on the release should then raise new issues in the team GitHub repositories ([govuk-frontend](https://github.com/alphagov/govuk-frontend), [govuk-design-system](https://github.com/alphagov/govuk-design-system/issues/2024), [govuk-frontend-docs](https://github.com/alphagov/govuk-frontend-docs) to track the following:
+Release issues should be created in the team GitHub repositories: [govuk-frontend](https://github.com/alphagov/govuk-frontend), [govuk-design-system](https://github.com/alphagov/govuk-design-system/issues/2024) and [govuk-frontend-docs](https://github.com/alphagov/govuk-frontend-docs).
```
36degrees, line 15: Similar to #3405, this should probably link to the repo consistently:

```suggestion
Release issues should be created in the team GitHub repositories: [govuk-frontend](https://github.com/alphagov/govuk-frontend), [govuk-design-system](https://github.com/alphagov/govuk-design-system/) and [govuk-frontend-docs](https://github.com/alphagov/govuk-frontend-docs).
```

(Or, if we think it's useful to link to the 'example' issues, we should make that clear in the link text and do it consistently for all of the repos)

docs/releasing/before-publishing-a-release.md
```diff
@@ -12,7 +12,15 @@ Optionally, we can also let the prototype kit know that we are coordinating a re
...
+
+All these issues should be:
+
+- added to the [Design System sprint board](https://github.com/orgs/alphagov/projects/53) backlog
+- added to that release's milestone
+- tagged with the '🚀 release' tag
+
+Those leading on the release should then raise new issues to track the following:
```
36degrees, line 23: This feels slightly out of order – as I understand it, these _are_ the release issues that are mentioned above:

'Release issues should be created in the team GitHub repositories. […] Those leading on the release should then raise new [release issues]'

Maybe these additional bullets about the projects, milestones and tags should come after the list of issues that need creating?

36degrees (commented):
Thanks for doing this, beeps 🙌🏻

docs/releasing/before-publishing-a-release.md
querkmachine: Yeah, that's where they were originally. I moved them up because, in my mind, saying "each of the following issues requires these three things" makes more sense being said ahead of time rather than after-the-fact—especially if the document is being read as a set of sequential steps.

That could just be symptomatic of my failure to read ahead, though. 😛

36degrees: I think it might just be the word *then* that's throwing me off?

```suggestion
Those leading on the release should raise new issues to track the following:
```

36degrees (approved):
Thanks @querkmachine 👍🏻

PR #3492, opened 2023-04-05
Change EtP keyboard shortcut expiry methodology

querkmachine:
Changes the method that the EtP keyboard shortcut self-expires. 

Currently, the user must complete all 3 keypresses within a 5 second time period, starting with the first keypress. This can pose a problem for screen reader users, as the associated announcements can collectively take 5 seconds (or longer) to be fully read out. 

This changes the methodology so that the 5 second timer is restarted after each keypress, allowing up to 10 seconds to achieve the needed 3 keypresses. 

## Changes

- Removes the check that only ran `setKeypressTimer` if `keypressTimeoutId` was null, as this prevented any keypresses after the first from setting the timer. 
- Clears `keypressTimeoutId` when `setKeypressTimer` is invoked, so that only one timer is running at any one time. 

## Thoughts

- Now that the timer resets with each keypress, can the timer be shorter than 5 seconds? 
- https://github.com/alphagov/govuk-design-system/pull/2721 will need updating if this is merged.

### DIFF
--- a/src/govuk/components/exit-this-page/exit-this-page.mjs
+++ b/src/govuk/components/exit-this-page/exit-this-page.mjs
@@ -221,13 +221,21 @@ ExitThisPage.prototype.handleKeypress = function (e) {
 
 /**
  * Starts the 'quick escape' keyboard sequence timer.
+ *
+ * This can be invoked several times. We want this to be possible so that the
+ * timer is restarted each time the shortcut key is pressed (e.g. the user has
+ * up to n seconds between each keypress, rather than n seconds to invoke the
+ * entire sequence.)
  */
 ExitThisPage.prototype.setKeypressTimer = function () {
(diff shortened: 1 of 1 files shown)


davidc-gds (approved):
Seems good to me, I don't have anything to add from a functionality-change perspective. 

Others may have thoughts on the soundness of the code, but from my quick testing it seems to do what it says it'll do (on macOS Chrome, Safari, Edge and Firefox).

querkmachine:
There is currently a community consultation for EtP ongoing until 20th April, so we may want to avoid merging this until after then—just in case it causes issues.

stevenjmesser:
Moving to blocked so that we don't merge this, so that it doesn't impact the consultation. Can be merged once the consultation has been finished.

owenatgov (approved):
Looks good to me 👍🏻

PR #3493, opened 2023-04-06
Add (optional) type declaration checks in all files

colinrotherham:
This is a follow to:

This PR adds `tsconfig.json` files into **config**, **lib**, **tasks** with coverage now on `*.test.*` files

These checks are still optional but JSDoc type autocomplete (to helps spot issues) works all files

It was split out from https://github.com/alphagov/govuk-frontend/pull/3399


domoscargin (approved):
Looks grand to me!

PR #3499, opened 2023-04-12
Deprecate all JavaScript instance properties the except `init` method

### LINKED ISSUE #3473: Deprecate properties that will no longer be part of the public API
## What

Mark properties that will no longer be part of the public API in v5 as deprecated, so users are aware ahead of time.

## Why

We'll be clarifying our public API in v5, which means we will either not support or prevent access to some of the properties currently available on our components. We need to make people aware of this ahead of time to give them as much time as possible to anticipate the change.

## Who needs to work on this

Developers, Tech writers

## Who needs to review this

Developers, Tech writers

```[tasklist]
### Done when
1. [x] Confirm that any property that's not `init` should be deprecated
2. [x] Properties have been marked as `@deprecated` in their JSDocs
3. [x] We've drafted communications to make people aware of the deprecations
```

36degrees:
We'll be clarifying our public API in v5, which means we will either not support or prevent access to some of the properties currently available on our components.

Deprecate everything except the `init` method to make people aware of this ahead of time to give them as much time as possible to anticipate the change.

Closes #3473.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,14 @@ This was added in [pull request #2230: Add extra letter spacing modifier for inp
 
 ### Deprecated features
 
+#### Stop using JavaScript API properties other than the `init` method
+
+We have deprecated all of the JavaScript properties in the API, except for the `init` method for each component. We will make all of the deprecated JavaScript properties private in v5.0.
+
+Please [let us know if you're using parts of the API other than the `init` method by filling in this form](https://docs.google.com/forms/d/e/1FAIpQLSfmH2AitMeouXqB0FWC5p5e6y1TSiFCjmJ8DrVuwfmpRGCaWw/viewform?usp=sf_link). We'll use this information when prioritising future additions to the public API.
+
+This was added in [pull request #3499: Deprecate all JavaScript instance properties the except `init` method](https://github.com/alphagov/govuk-frontend/pull/3499).
+
 #### Stop using the `govuk-button--disabled` class on buttons
 
 The `govuk-button--disabled` class is deprecated and will be removed in the next major release.
--- a/src/govuk/common/closest-attribute-value.mjs
+++ b/src/govuk/common/closest-attribute-value.mjs
@@ -3,6 +3,7 @@ import '../vendor/polyfills/Element/prototype/closest.mjs'
 /**
  * Returns the value of the given attribute closest to the given element (including itself)
  *
+ * @deprecated Will be made private in v5.0
  * @param {Element} $element - The element to start walking the DOM tree up
  * @param {string} attributeName - The name of the attribute
  * @returns {string | null} Attribute value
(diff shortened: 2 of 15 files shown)


CHANGELOG.md
```diff
@@ -64,6 +64,14 @@ This was added in [pull request #2230: Add extra letter spacing modifier for inp
 
 ### Deprecated features
 
+#### Stop using JavaScript API methods other than the `init` method
+
+All JavaScript methods in the API for our JavaScript components except `init` are now deprecated, and will be made private in v5.0.
+
+If you're calling a method other than `init`, [please tell us by filling in this form](https://docs.google.com/forms/d/e/1FAIpQLSfmH2AitMeouXqB0FWC5p5e6y1TSiFCjmJ8DrVuwfmpRGCaWw/viewform?usp=sf_link). We'll use this information when prioritising future additions to the public API.
```
36degrees, line 71: Please review the content and structure of the linked Google Form as well. I've pinged @claireashworth to get her thoughts on it too.

CHANGELOG.md
```diff
@@ -64,6 +64,14 @@ This was added in [pull request #2230: Add extra letter spacing modifier for inp
 
 ### Deprecated features
 
+#### Stop using JavaScript API properties other than the `init` method
+
+All JavaScript properties in the API for our JavaScript components except for the `init` method are now deprecated, and will be made private in v5.0.
```
claireashworth, line 69:
```suggestion
We have deprecated all JavaScript properties in the API for our JavaScript components, except for the `init` method. We will make all deprecated JavaScript properties private in v5.0.
```

36degrees: What about this, to avoid some of the repetition? (Especially 'JavaScript' which was already in there twice…)

> We have deprecated all of the JavaScript properties in the API except for the `init` method. We will make them private in v5.0.

claireashworth: I repeated as the last thing mentioned in the previous sentence in the init method, which means grammatically 'them' would refer to the init method. I shall shuffle things...

claireashworth: I've amended my suggestion - does that work?

colinrotherham (approved):
Approving but feel free to find more things to deprecate 😊

CHANGELOG.md
36degrees: I think the version with the repetition was better – this one seems to slightly contradict itself. What about:

> We have deprecated all of the JavaScript properties in the API, except for the `init` method for each component. We will make all the deprecated JavaScript properties private in v5.0.

36degrees: Trying to break it down:

- We have deprecated all of the JavaScript properties in the API.
- The exception to this is the `init` method for each component, which remains public.

With the additional comma I think it would read as:

- We have deprecated all of the JavaScript properties in the API for every component.
- The exception to this is the `init` method, which remains public.

Which isn't quite right – every component has an `init` method, which will remain public; and there are also parts of the API that aren't part of any component.

claireashworth: So I've changed the suggestion to your 3 sentences.

romaricpascal (approved):
Looks good to me as well (both code and form), thanks for going through all these files Ollie 🙌🏻

PR #3500, opened 2023-04-14
Return `render()` and `renderTemplate()` helpers via cheerio

colinrotherham:
This PR addresses a code review comment in https://github.com/alphagov/govuk-frontend/pull/3412#pullrequestreview-1383787077 from @romaricpascal 

I've also added the `@types/nunjucks` package so we can type-check our Nunjucks code. Before we added **tsconfig.json** files it was downloaded through [automatic type acquisition](https://code.visualstudio.com/docs/nodejs/working-with-javascript#_typings-and-automatic-type-acquisition) by supported editors

### DIFF
--- a/app/src/app.mjs
+++ b/app/src/app.mjs
@@ -4,6 +4,7 @@ import express from 'express'
 import { paths } from 'govuk-frontend-config'
 import { getDirectories, getComponentsData, getFullPageExamples } from 'govuk-frontend-lib/files'
 import { componentNameToMacroName } from 'govuk-frontend-lib/names'
+import { outdent } from 'outdent'
 
 import * as middleware from './common/middleware/index.mjs'
 import * as nunjucks from './common/nunjucks/index.mjs'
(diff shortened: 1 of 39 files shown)


36degrees (approved):
Neat 🙌🏻

app/src/app.mjs
```diff
@@ -115,10 +116,10 @@ export default async () => {
...
     const macroParameters = JSON.stringify(exampleConfig.data, null, '\t')
 
-    res.locals.componentView = env.renderString(
-      `{% from '${componentName}/macro.njk' import ${macroName} %}
-      {{ ${macroName}(${macroParameters}) }}`
-    )
+    res.locals.componentView = env.renderString(outdent`
+      {% from '${componentName}/macro.njk' import ${macroName} %}
+      {{ ${macroName}(${macroParameters}) }}
+    `, {})
```
colinrotherham, line 122: Didn't appear to break but we'd missed the context param here

PR #3501, opened 2023-04-14
Include "Review app" in `npm run build:types` check

colinrotherham:
Our internal "Review app" imports source code with JSDoc comments but doesn't check for correct usage

Work on https://github.com/alphagov/govuk-frontend/pull/3500 revealed missing Nunjucks params against the community [`@types/nunjucks`](https://www.npmjs.com/package/@types/nunjucks) package

This PR adds enough JSDoc comments to the Review app that we can check our own usage:

```console
npm run build:types
```

We've had some failed deployments recently that these checks would have caught

Although they won't be switched on until we merge:

* https://github.com/alphagov/govuk-frontend/pull/3319

### DIFF
--- a/app/src/app.mjs
+++ b/app/src/app.mjs
@@ -27,7 +27,7 @@ export default async () => {
 
   // Set up Express.js
   app.set('flags', flags)
-  app.set('query parser', (query) => new URLSearchParams(query))
+  app.set('query parser', 'simple')
 
   // Set up middleware
   app.use('/docs', middleware.docs)
@@ -122,8 +122,7 @@ export default async () => {
 
     let bodyClasses = ''
 
-    // @ts-expect-error `URLSearchParams` as 'query parser'
-    if (req.query.has('iframe')) {
+    if ('iframe' in req.query) {
       bodyClasses = 'app-iframe-in-component-preview'
     }
 
--- a/app/src/routes/full-page-examples.test.mjs
+++ b/app/src/routes/full-page-examples.test.mjs
@@ -336,21 +336,21 @@ describe(`http://localhost:${ports.app}/full-page-examples/`, () => {
         const body = await response.text()
         const $ = load(body)
         // Check the results are correct
-        expect($.html()).toContain('121,842 results')
+        expect($.html()).toContain('482,211 results')
       })
       it('should show sorted results when selected', async () => {
         const response = await fetchPath('search?order=updated-newest')
         const body = await response.text()
         const $ = load(body)
         // Check the results are correct
-        expect($.html()).toContain('821,124 results')
+        expect($.html()).toContain('241,128 results')
       })
       it('should show organisation results when selected', async () => {
         const response = await fetchPath('search?order=updated-newest&organisation=hmrc')
         const body = await response.text()
         const $ = load(body)
         // Check the results are correct
-        expect($.html()).toContain('212,418 results')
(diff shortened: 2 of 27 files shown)


app/src/app.mjs
```diff
@@ -29,7 +29,7 @@ export default async () => {
 
   // Set up Express.js
   app.set('flags', flags)
-  app.set('query parser', (query) => new URLSearchParams(query))
+  app.set('query parser', 'simple')
```
colinrotherham, line 32: We recently switched to `URLSearchParams` (coming to Express in future) but the community types in [`@types/express`](https://www.npmjs.com/package/@types/express) don’t expect it to be overridden. See discussion:

* https://github.com/alphagov/govuk-frontend/pull/3223#discussion_r1102469094

app/src/views/full-page-examples/applicant-details/index.mjs
```diff
@@ -19,11 +22,17 @@ export default (app) => {
...
+     * @param {import('express').Request} request
+     * @param {import('express').Response} response
+     */
     (request, response) => {
       const errors = formatValidationErrors(validationResult(request))
 
       if (!errors) {
-        return response.render('./full-page-examples/applicant-details/confirm')
+        response.render('./full-page-examples/applicant-details/confirm')
+        return
```
romaricpascal, line 35: **question**: Not blocking, but just want to confirm that this is because the types expect the functions handling the requests to return `undefined`, is that it? 😊

app/src/routes/full-page-examples.test.mjs
```diff
@@ -336,21 +336,21 @@ describe(`http://localhost:${ports.app}/full-page-examples/`, () => {
         const body = await response.text()
         const $ = load(body)
         // Check the results are correct
-        expect($.html()).toContain('121,842 results')
+        expect($.html()).toContain('482,211 results')
```
romaricpascal, line 339: **question**: Why have the numbers in the tests changed? Is it due to the change in query parser?

romaricpascal (commented):
Overall looking good 🙌🏻  Just had two little questions as I got through the code. It's really the second one that's holding the approval 😊

app/src/views/full-page-examples/applicant-details/index.mjs
colinrotherham: Nah in this instance `response.render()` already does nothing

I'm just avoiding the ESLint plugin telling me off for missing a `@returns {void}` tag

Would you rather have the tag + return description?

app/src/routes/full-page-examples.test.mjs
colinrotherham: We pass the the query (as a string) into [`shuffle-seed`](https://www.npmjs.com/package/shuffle-seed)

So yeah it's gone from `URLSearchParams.prototype.toString()` to JSON which changes all the numbers

Same query name/values still

app/src/views/full-page-examples/applicant-details/index.mjs
romaricpascal: Given we're already typing the functions, I'd rather solve it with an extra JSDoc tag and keep the code simpler, yup 😄  Thanks!

romaricpascal (approved):
Cheers for using the `@returns` for typing the return of the route handlers 😊 

⛵

PR #3407, opened 2023-03-23
Update organisation colours

querkmachine:
A long-overdue update to our somewhat fairly outdated `_colours-organisations.scss` file. In addition to adding missing organisations and updating the colours, I've tried to tackle the comments in #1919.

The main colours for all existing organisations haven't changed: reds are still red, blues are still blues (with the exception of UK Export Finance, which has rebranded in the intervening years). The specific shades are different though, as are many of the contrast-safe equivalent colours and shades.

### Changes to code
- Adds a `$govuk-new-organisation-colours` feature flag to enable the new colour palette. 
- Adds a map of 'aliases' from old organisation keys to their new equivalents.
  - This is only intended for internal renames (e.g. if we typo'd a name in a release). Changes due to rebrands or machinery of government changes should be added as a new entry in the colour map, with any defunct predecessor organisations marked as deprecated. 
- Updates the `govuk-organisation-colour` function to produce a warning when a defunct organisation is referenced. 
- Updates all contrast-safe colours based on a fixed methodology (more about that below).
- Adds `contrast-safe` key to colour map and `$contrast-safe` parameter to the `govuk-organisation-colour` mixin, to [clearly indicate the intent behind them](https://github.com/alphagov/govuk-frontend/issues/1919).
  - These replace the now deprecated `colour-websafe` key and `$websafe` function parameter, respectively.

### Changes to organisations listed
- Adds current government departments that were missing from the colour map, with brand colours from Design102 and contrast-safe variants as determined by the process below:
  - `department-for-energy-security-net-zero`
  - `department-for-science-innovation-technology`
  - `department-of-health-social-care`
  - `foreign-commonwealth-development-office`
  - `ministry-of-housing-communities-local-government`
  - `office-of-the-secretary-of-state-for-scotland`
  - `office-of-the-secretary-of-state-for-wales`
  - `prime-ministers-office-10-downing-street`
- Adds government departments that have come and gone since the last significant update to the colour map for historical completeness.
  - `department-for-business-energy-industrial-strategy`
  - `department-for-digital-culture-media-sport`
  - `department-for-levelling-up-housing-communities`
- Adds deprecation messages to defunct organisations, with information on their replacements and current equivalents. [See this comment for a list of defunct organisations and their replacements](https://github.com/alphagov/govuk-frontend/pull/3407#issuecomment-1486590189). 
[...]

PR #3523, opened 2023-04-18
Add new brand colour for DSIT

beccapearce:


querkmachine:
Good morning! Thanks for the contribution!

Unfortunately, you've caught us right at the start of a code freeze, as we're preparing to release a new version of GOV.UK Frontend. As such, this PR is unlikely to be merged and released for quite a while as part of our substantial v5 release.

We're also currently reviewing what organisation colours we ship and the process by which we update them. That work is also likely to go out in v5. 

I've seen that you've opened [a similar change in GOV.UK Publishing Components](https://github.com/alphagov/govuk_publishing_components/pull/3349) too. Hopefully that will work for your needs in the meantime.

colinrotherham:
I've rebased this PR with [main](https://github.com/alphagov/govuk-frontend/blob/main) to resolve `src` → `packages/govuk-frontend/src` conflicts

//cc @querkmachine

querkmachine:
Hi there! Sorry for the long wait on this one.

We've decided we would prefer to batch organisation colour changes into our larger organisation colour update in #3407. In that PR, we're using the colours that (to the best of our knowledge) have been ascribed to each department by Design102 to try and enforce consistency, rather than having individual departments raise their own PRs. 

As such, I'm going to close this PR for now, as it will be largely superceded by this other work. Thank you!

PR #3527, opened 2023-04-20
Ensure Dart Sass `file://` sources use relative paths

colinrotherham:
In preparation for [**v4.6.0** release](https://github.com/alphagov/govuk-frontend/milestone/54) we noticed Dart Sass writes absolute file paths to source map sources:

* https://github.com/sass/dart-sass/pull/1021

This PR ensures we keep relative paths


colinrotherham:
The [**GOV.UK Design System** website](https://design-system.service.gov.uk) also uses Dart Sass but isn't built locally:

Source maps on the Design System website

shared/tasks/assets.mjs
```diff
@@ -14,15 +14,27 @@ export async function write (filePath, result) {
   const writeTasks = []
 
   // Files to write
+  /** @type {SourceMap | undefined} */
+  const map = result.map ? JSON.parse(result.map.toString()) : undefined
```
colinrotherham, line 18: We need `map.sources` a few lines down so parse to [`SourceMap`](https://github.com/source-map/source-map-spec) here

Otherwise across Terser, Rollup and PostCSS we see source maps as:

1. JavaScript `object` types that need `JSON.stringify()`
2. Class instances with `.toJSON()` and `.toString()` methods
3. Preprepared `string` type
4. Missing `undefined` type

This can be seen via the **AssetOutput** `@typedef` in this file

shared/tasks/assets.mjs
```diff
@@ -14,15 +14,27 @@ export async function write (filePath, result) {
...
-  if (map) {
-    writeTasks.push(writeFile(`${filePath}.map`, map))
+  if (map && 'sources' in map) {
+    map.sources = map.sources
+
+      /**
+       * Make source file:// paths relative (e.g. for Dart Sass)
+       * {@link https://sass-lang.com/documentation/js-api/interfaces/CompileResult#sourceMap}
+       */
+      .map((path) => path.startsWith('file://')
```
colinrotherham, line 32: PostCSS adds the Dart Sass entry point `../src/all.scss` as a relative path

But all the other Dart Sass sources are absolute path `file://` URLs

shared/tasks/build/package.test.mjs
```diff
@@ -70,6 +70,12 @@ describe('package/', () => {
         return output
       }))
 
+      // Add Autoprefixer prefixes to all source '*.scss' files
+      .flatMap(mapPathTo(['**/*.scss'], ({ dir: requirePath, name }) => [
+        join(requirePath, `${name}.scss`),
+        join(requirePath, `${name}.scss.map`) // with source map
```
colinrotherham, line 76: Service teams with source maps enabled will have been inspecting our Autoprefixer-transformed sources, not the original prefix-free `*.scss` source files

shared/tasks/styles.mjs
```diff
@@ -79,10 +94,8 @@ export async function compileStylesheet ([modulePath, { srcPath, destPath, fileP
     }))
 
     // Pass source maps to PostCSS
-    options.map = {
-      annotation: true,
-      inline: false,
-      prev: map
+    if (typeof options.map === 'object') {
+      options.map.prev = map
```
colinrotherham, line 98: PostCSS will automatically discover source maps via `sourceMappingURL` comments but we continue to pass them directly from Dart Sass here

romaricpascal (approved):
I think that did it and it's all very localised! Thanks for adding tests as well 🙌🏻 

Screenshot 2023-04-21 at 09 46 00
Screenshot 2023-04-21 at 09 49 00

PR #3530, opened 2023-04-22
Remove incorrect information about visually hidden text on previous/next

matteason:
The [pagination documentation](https://design-system.service.gov.uk/components/pagination/) says that by default, Previous and Next links will have some visually hidden text (eg `Previous<span class="govuk-visually-hidden"> page</span>`).

This doesn't actually seem to be the case - the link text is just 'Previous' or 'Next' (unless `labelText` is also provided, in which case there's a visually hidden colon, but that's it): https://github.com/alphagov/govuk-frontend/blob/6233eaa9cfb6c0dd44ed8341e0ac5e0afe339f84/src/govuk/components/pagination/template.njk#L10-L16

I can't see a visually hidden span or `aria-label` that would match the behaviour described in the documentation, and testing with NVDA on the example at https://design-system.service.gov.uk/components/pagination/default/index.html, those links just read 'Previous' and 'Next'

This PR just updates the documentation to match the current behaviour, but there's probably also a discussion to be had on whether the behaviour of the component should change to match what's described in the documentation. Some of the examples in this file also seem to assume that there are `previous.visuallyHiddenText` and `next.visuallyHiddenText` options: https://github.com/alphagov/govuk-frontend/blob/6233eaa9cfb6c0dd44ed8341e0ac5e0afe339f84/src/govuk/components/pagination/pagination.yaml#L223-L228

Again, those don't seem to actually exist, but maybe they should? (I'm guessing these examples are shown in the review app, which I can't find, because they're not in the Design System documentation)

### DIFF
--- a/src/govuk/components/pagination/pagination.yaml
+++ b/src/govuk/components/pagination/pagination.yaml
@@ -36,7 +36,7 @@ params:
       - name: text
         type: string
         required: false
-        description: The link text to the previous page. Defaults to 'Previous page', where 'page' is visually hidden.
+        description: The link text to the previous page. Defaults to 'Previous'.
       - name: labelText
         type: string
         required: false
@@ -57,7 +57,7 @@ params:
       - name: text
         type: string
         required: false
-        description: The link text to the next page. Defaults to 'Next page', where 'page' is visually hidden.
+        description: The link text to the next page. Defaults to 'Next'.
       - name: labelText
         type: string
         required: false
--- a/src/govuk/components/pagination/pagination.yaml
+++ b/src/govuk/components/pagination/pagination.yaml
@@ -225,11 +225,9 @@ examples:
       previous:
         text: 'précédente'
         href: '/previous'
-        visuallyHiddenText: 'Page précédente'
       next:
         text: 'suivante'
         href: '/next'
-        visuallyHiddenText: 'Page suivante'
   - name: with previous only
     data:
       previous:


36degrees:
Great spot, thanks @matteason 👍🏻 

From memory, I think this change came quite late in the development cycle. We were originally using `aria-label` on the previous and next links. But [when the previous and next links had labels we realised this meant the labels were not part of the accessible name](https://github.com/alphagov/govuk-frontend/pull/2610#discussion_r877999986) and therefore failed [WCAG 2.1 2.5.3 Label in Name](https://www.w3.org/WAI/WCAG21/Understanding/label-in-name.html).

After that review, there was a [force-push](https://github.com/alphagov/govuk-frontend/compare/5cfab20d27b72d51a10915abb92684e43d82a8c3..ea2931c30127a64d940f040df85b99dfcfe30919) that removed the `previous.visuallyHiddenText` and `next.visuallyHiddenText` params, but it looks like a couple of bits were missed.

I don't believe anything has changed – I think we're still broadly happy with this approach and we think that 'Previous' and 'Next' are descriptive enough labels.

On that basis I'm going to add an additional commit to update the examples that are still using the `previous.visuallyHiddenText` and `next.visuallyHiddenText` params.

I'm also going to tag @owenatgov just to check he's happy with this before merging.

owenatgov:
Excellent spot @matteason!

You're right @36degrees that this got lost between changes in accessible name implementation during the review process. I'll approve once you've added that extra commit.

owenatgov (approved):

Issue #3541, opened 2023-04-24
Switch to new Chrome headless mode


colinrotherham:
## What

We're getting lots of warnings about Chrome `{ headless: true }` related to the new headless mode in v112

**Chrome’s Headless mode gets an upgrade: introducing `--headless=new`**
https://developer.chrome.com/articles/new-headless/

We should investigate and switch to `{ headless: 'new' }` unless we hit issues:

>Technically, the old Headless was [a separate, alternate browser implementation](https://source.chromium.org/chromium/chromium/src/+/main:headless/;drc=c67febd82ae3e18ac8db1397f4ccfa87b0da2ffc) that happened to be shipped as part of the same Chrome binary. It doesn’t share any of the Chrome browser code in [//chrome](https://source.chromium.org/chromium/chromium/src/+/main:chrome/).

## Why

It's important that we

* Test using the same Chrome browser as our users
* Identify Chrome issues early before new headless mode is the default

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [ ] Switch to new Chrome headless mode

colinrotherham:
Can confirm it's rather noisy 🤦‍♂️ since we merged:

* https://github.com/alphagov/govuk-frontend/pull/3537

```console
Puppeteer old Headless deprecation warning:
    In the near feature `headless: true` will default to the new Headless mode
    for Chrome instead of the old Headless implementation. For more
    information, please see [https://developer.chrome.com/articles/new-headless/.](https://developer.chrome.com/articles/new-headless/)
    Consider opting in early by passing `headless: "new"` to `puppeteer.launch()`
    If you encounter any bugs, please report them to https://github.com/puppeteer/puppeteer/issues/new/choose.
```

<img width="1156" alt="Console spam from Puppeteer" src="https://user-images.githubusercontent.com/415517/234049571-baa99ae7-c19d-47b1-bf72-b496ffe429dc.png">

PR #3542, opened 2023-04-24
Prevent Chrome headless deprecation warning

colinrotherham:
This PR prevents the recent [Puppeteer `{ headless: true }` deprecation warning](https://github.com/puppeteer/puppeteer/pull/10039):

* https://github.com/puppeteer/puppeteer/pull/10039

We can revisit this in future with “new” headless mode via:

* https://github.com/alphagov/govuk-frontend/issues/3541

Until then it prevents all the console spam we're getting:

Console spam from Puppeteer

### DIFF
--- a/jest-puppeteer.config.js
+++ b/jest-puppeteer.config.js
@@ -30,7 +30,18 @@ module.exports = {
        */
       '--no-startup-window'
     ],
-    headless: process.env.HEADLESS !== 'false',
+
+    /**
+     * Allow headless mode switching using `HEADLESS=false`
+     * but default to `1` to skip deprecation warning
+     *
+     * {@link https://developer.chrome.com/articles/new-headless/}
+     */
+    headless: process.env.HEADLESS !== 'false'
+      ? 1
+      : false,
+
+    // See launch arg '--no-startup-window'
     waitForInitialPage: false
   },
 


jest-puppeteer.config.js
```diff
@@ -30,7 +30,18 @@ module.exports = {
...
-    headless: process.env.HEADLESS !== 'false',
+
+    /**
+     * Allow headless mode switching using `HEADLESS=false`
+     * but default to `undefined` to skip deprecation warning
+     *
+     * {@link https://developer.chrome.com/articles/new-headless/}
+     */
+    headless: process.env.HEADLESS !== 'false'
+      ? undefined
```
colinrotherham, line 41: By defaulting to `{ headless: undefined }` we skip the deprecation checks in:

* https://github.com/puppeteer/puppeteer/pull/10039

domoscargin (approved):
Makes sense to me!

colinrotherham:
I've raised this with the Puppeteer team as the intention probably wasn't to throw up 11x warnings per test run:

* https://github.com/puppeteer/puppeteer/issues/10070

jest-puppeteer.config.js
colinrotherham: @domoscargin Sadly they're using [`const headless = options.headless ?? true;`](https://github.com/puppeteer/puppeteer/pull/10039/files#diff-61c2acef7c3c76385e015ae62bb14ba8789889e19dc57f7993a5dec0cb253ac2R48) which compiles to:

```js
const headless = (_a = options.headless) !== null && _a !== void 0 ? _a : true;
```

But we can default to `1` instead of `undefined` to make this work

PR #3552, opened 2023-04-26
Clarify docs for radio and checkbox conditionals

### LINKED ISSUE #1903: Confused by the documentation for radios 'conditional'
I was trying to use the conditional option on radios.

I saw that the example used

```
conditional:{
    html:"<p> ...
```

I didn't need html, just a sentence so I thought I could use `text` instead of `html` as you can in other places in Frontend. However, I tried it and it didn't work - there's no text option here.

When I re-read the docs, I misread and thought the 'conditional' option, being a string, could just accept text.

However, I misread it, that doesn't work - the docs actually say:

> conditional | string | If true, content provided will be revealed when the item is checked.

But I don't think this is correct? In the examples, `conditional` is not a string, it's an object with only one property - `html`

I wonder if this might be better as two options:

`conditionalHTML` and `conditionalText`

This would avoid another object, and be more consistent with the use of HTML and text elsewhere?

36degrees:
The documentation for the conditional Nunjucks macro option for checkboxes and radios is confusing.

The checkboxes documentation lists `conditional` as a boolean, and the radios documentation lists it as a string.

Neither are correct.

The `conditional` option is an object with a single nested option, `html`. The nested `html` option is a string, used to specify the content that is revealed when the checkbox or radio is checked.

Update the component YAML so that the documentation reflects this.

### Before

Screenshot 2023-04-26 at 09 34 30

### After

Screenshot 2023-04-26 at 09 47 31
Screenshot 2023-04-26 at 09 47 38

### Before

Screenshot 2023-04-26 at 09 35 20

### After

Screenshot 2023-04-26 at 09 48 05
Screenshot 2023-04-26 at 09 48 12

Closes #1903

### DIFF
--- a/src/govuk/components/checkboxes/checkboxes.yaml
+++ b/src/govuk/components/checkboxes/checkboxes.yaml
@@ -79,13 +79,13 @@ params:
         required: false
         description: Whether the checkbox should be checked when the page loads. Takes precedence over the top-level `values` option.
       - name: conditional
-        type: boolean
+        type: object
         required: false
-        description: If `true`, content provided will be revealed when the item is checked.
-      - name: conditional.html
-        type: string
-        required: false
-        description: Provide content for the conditional reveal.
+        description: Provide additional content to reveal when the checkbox is checked.
+        params:
+          - name: html
+            type: string
+            description: The HTML to reveal when the checkbox is checked
       - name: behaviour
         type: string
         required: false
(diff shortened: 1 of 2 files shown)


colinrotherham:
Thanks @36degrees

You probably know I've written types for the YAML component data here:
https://github.com/alphagov/govuk-frontend/blob/main/shared/lib/files.js#L163

This is so we can get autocomplete (and checks) in tests using `getExamples()` etc

But the example data just has a generic `{object}` type `getComponentData()` isn't that helpful, but if we're bringing things in line it might be good to start typing it with the various flavours of component data

Might reveal other places where we've got the object structure wrong

querkmachine (approved):
Docs changes look good to me :shipit:

PR #3555, opened 2023-04-26
Update “caniuse” data with `npx update-browserslist-db@latest`

colinrotherham:
Just another console warning that popped up today

```console
Browserslist: caniuse-lite is outdated. Please run:
  npx update-browserslist-db@latest
  Why you should do it regularly: https://github.com/browserslist/update-db#readme
```

We apply these data-specific [browserslist](https://browsersl.ist) queries to [Autoprefixer](https://github.com/postcss/autoprefixer#readme) but our data is from November 2022

```
> 0.1%
last 2 Chrome versions
last 2 Firefox versions
last 2 Edge versions
last 2 Samsung versions
```

This PR runs `npx update-browserslist-db@latest` as suggested

Should browser data freshness be an automated test or release checklist item?

### DIFF
--- a/docs/releasing/publishing.md
+++ b/docs/releasing/publishing.md
@@ -30,7 +30,15 @@ Developers should pair on releases. When remote working, it can be useful to be 
 
    Do not commit the changes.
 
-7. Run `npm run build-release` to:
+7. Update browser data from ["Can I use"](https://caniuse.com) by running:
+
+   ```
+   npx update-browserslist-db@latest
+   ```
+
+   This step will update the project `package-lock.json` file if updates are found.
+
+8. Run `npm run build-release` to:
 
    - build GOV.UK Frontend into the `/package` and `/dist` directories
    - commit the changes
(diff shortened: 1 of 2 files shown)


domoscargin (approved):
Was just about to raise this myself! 

Thanks

colinrotherham:
> And yeah, if we could automate it easily, I don't see why not.

Not for now, but knowing that browser data will affect "the build" (polyfills dropped, prefixes removed, more agressive minification) it probably needs a scheduled GitHub Actions at some point—bit like Dependabot

For now, think we need docs changes?

For example we say to [run `nvm use`](https://github.com/search?q=repo%3Aalphagov%2Fgovuk-frontend%20Run%20%60nvm%20use%60&type=code) to make sure Node.js is up to date

But the browser data could be 5+ months old

domoscargin:
Yeah, let's put a line into the release docs - pretty easy to do manually till we automate.

colinrotherham:
> Yeah, let's put a line into the release docs - pretty easy to do manually till we automate.

@domoscargin I've added a line to the publishing instructions 😊

PR #3556, opened 2023-04-26
Add inverse button styles

### LINKED ISSUE #2267: Add inverse button modifier
## What

Add a modifier to the button component for situations where buttons appear against dark backgrounds (for example, in the blue masthead on product pages)

## Why

The [product page example](https://github.com/alphagov/product-page-example) repo currently contains [CSS for inverse buttons](https://github.com/alphagov/product-page-example/blob/cc4ee7f89693ec9d29b6f13833ea72986486c898/source/stylesheets/modules/_inverted-button.scss) for use on product pages across GaaP.

There are already various examples of inverse buttons across the various GaaP product pages, including the Design System:

Screenshot 2021-07-05 at 13 22 14
Screenshot 2021-07-05 at 13 22 23
Screenshot 2021-07-05 at 13 22 31
Screenshot 2021-07-05 at 13 22 37

## Further information

- How do the 'start button' and 'inverse' modifiers combine?
- Some of these buttons are as big as start buttons, but don't have the chevron. What are the rules around button sizes and chevrons?

Designers, developers

## Done when

- [ ] Inverse modifier for buttons added
- [ ] Example(s) added to review app

querkmachine:
Ports the inverse button styles from the [GaaP product page repository](https://github.com/alphagov/product-page-example/blob/master/source/stylesheets/modules/_inverted-button.scss) into Frontend.

I haven't added any additional automated or accessibility tests (not that other stylistic variants have them), but these styles have existed in the wild for a long time already—one would hope that any major issues with the styles had already been identified. 

Closes #2267.

### DIFF
--- a/src/govuk/components/button/button.yaml
+++ b/src/govuk/components/button/button.yaml
@@ -148,6 +148,33 @@ examples:
       text: Warning button
       href: '/'
       classes: govuk-button--warning
+  - name: Inverse
+    description: A button that appears on dark backgrounds
+    data:
+      name: Inverse
+      text: Inverse button
+      classes: govuk-button--inverse
+  - name: Inverse disabled
+    data:
+      name: Inverse
+      text: Inverse button disabled
+      classes: govuk-button--inverse
+      disabled: true
+  - name: Inverse link
+    description: A link button for actions that appear on dark backgrounds
+    data:
+      name: Inverse
+      text: Inverse button
+      href: '/'
+      classes: govuk-button--inverse
+  - name: Inverse start button
+    description: A start button that appears on dark backgrounds
+    data:
+      name: Inverse
+      text: Inverse start button
+      href: '/'
+      classes: govuk-button--inverse
+      isStartButton: true
 
   # Hidden examples are not shown in the review app, but are used for tests and HTML fixtures
   - name: attributes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 ## Unreleased
 
+### New features
+
+#### Added inverse modifier for buttons on dark backgrounds
+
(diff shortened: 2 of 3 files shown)


src/govuk/components/button/button.yaml
```diff
@@ -148,6 +148,33 @@ examples:
       text: Warning button
       href: '/'
       classes: govuk-button--warning
+  - name: Inverse
+    description: A button that appears on dark backgrounds
+    data:
+      name: Inverse
+      text: Inverse button
+      classes: govuk-button--inverse
```
owenatgov, line 156: I'm open to being disagreed with as I know this is an approach we take elsewhere but my personal preference would be to control this via an API attribute eg: `inverse: true`. Communicating features in this way feels clearer to me. It would then be more justified to write a test as you're testing that the API successfully applies the class.

owenatgov (commented):
Once comment that's up for debate. Everything else looking sharp.

src/govuk/components/button/button.yaml
querkmachine: As the stylistic alternatives are all mutually exclusive, I'd be inclined to combine them into a single option so that it's not possible to enable more than one (e.g. `appearance: "inverse"`), but then we'd probably have to add documentation what the allowed values are, as it may not be particularly obvious otherwise.

owenatgov: That's a good point @36degrees , dropping this point for now. @querkmachine I like this idea on principal but I also think this is something we should apply strategically to our whole codebase. I'll write up an issue to explore this.

querkmachine: Can I get a review if we're happy to leave as is? 🙏

owenatgov (approved):

paulrobertlloyd:
Will there be guidance for this component in the design system, or is this a hidden feature for now?

PR #3557, opened 2023-04-26
Remove deprecated `.govuk-button--disabled` class

### LINKED ISSUE #2681: Remove deprecated govuk-button--disabled class
## What

Relies on the class having been marked as deprecated in #2271.

It's currently possible to make 'link buttons' look disabled by using the `.govuk-button--disabled` modifier class. However, it doesn't really make sense to do so as [you can't easily – and almost certainly shouldn't – disable links](https://css-tricks.com/how-to-disable-links/).

Remove the `.govuk-button--disabled` modifier class and only apply disabled styles using the `[disabled]` attribute selector.

We'll also need to modify the macro so that the `disabled` option does nothing for link buttons, and update the macro documentation accordingly.

## Why

Adding the `--disabled` modifier class to the button does not prevent users from interacting with it. Because of the way that our CSS is currently written, the disabled link button still has a change on focus state.

Because we're trying to remove states from an element that is still interactive and focusable, automated accessibility testing tools are likely to flag issues with it, as seen in #2127. We've also seen links with the disabled modifier flagged as failing color contrast requirements in Axe.

## Further detail

At the same time, we could consider applying the hover and active states using the `:enabled` pseudo-class, rather than the current approach of 're-setting' the non-hover state for each button variant to avoid the change on hover. This would allow us to simplify the CSS, but would mean that buttons would no longer have a hover or active state in IE8.

Designers, developers

## Done when

- [x] CSS for disabled class removed
- [x] Macro updated to ignore the `disabled` option for link buttons
- [x] Macro documentation updated to make it clear the disabled option has no effect on links

querkmachine:
Removes the `.govuk-button--disabled` class that was deprecated in 4.6.0. Closes #2681. 

## Changes

- Removes the `.govuk-button--disabled` selector and the deprecation comment from the stylesheet.
- Removes the disabled link example from the review app.
- Removes the test associated with the disabled link example.
- Updates the Nunjucks macro to no longer add the `.govuk-button--disabled` class to disabled buttons.
- Updates the Nunjucks macro documentation to indicate that the `disabled` parameter has no affect on `a` elements.
- Updates tests to no longer check for the existence of the `.govuk-button--disabled` class.

## Thoughts

We also have a `.govuk-button[disabled="disabled"]` selector defined in there. I'm not sure why this exists alongside `.govuk-button[disabled]`, as they have the same specificity and `disabled` is a boolean attribute (the presence of a value is meaningless). 

`.govuk-button[disabled]` alone would seem to suffice. Should we also remove `.govuk-button[disabled="disabled"]`? I don't think doing so would require a deprecation or breaking change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,17 @@ If you are using `$button.getAttribute('disabled')` to check for the disabled at
 
 Instead we recommend checking for the disabled attribute using [`$button.hasAttribute('disabled')`](https://developer.mozilla.org/en-US/docs/Web/API/Element/hasAttribute) or the [`$button.disabled` IDL attribute](https://developer.mozilla.org/en-US/docs/Web/API/HTMLButtonElement/disabled).
 
-This change was made in [pull request #2830: Set the boolean disabled attribute consistently in the button component](https://github.com/alphagov/govuk-frontend/pull/2830).
+This change was introduced in [pull request #2830: Set the boolean disabled attribute consistently in the button component](https://github.com/alphagov/govuk-frontend/pull/2830).
+
+#### Remove deprecated `.govuk-button--disabled` class
+
+We've removed the `.govuk-button--disabled` class that we deprecated in [GOV.UK Frontend v4.6.0](https://github.com/alphagov/govuk-frontend/releases/tag/v4.6.0).
+
+Use the `disabled` attribute to mark `<button>` and `<input>` elements as being disabled instead.
+
+We no longer support link buttons being disabled or using disabled styles.
+
+This change was introduced in [pull request #3557: Remove deprecated `.govuk-button--disabled` class](https://github.com/alphagov/govuk-frontend/pull/3557).
 
 ## 4.6.0 (Feature release)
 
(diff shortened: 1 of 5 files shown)


querkmachine:
The changelog might be a bit awkward as there's already a section regarding changes to disabled buttons. I've tried to merge these changes into it rather than having a separate, similarly titled section.

querkmachine:
> My guess is this is a failsafe for incorrect implementations?

Unless I'm having a late afternoon brainfart (not unheard of!) I'm not aware of any implementation that `.govuk-button[disabled="disabled"]` would match and that `.govuk-button[disabled]` wouldn't also match.

Even if teams have added their own styles for either selector, the value-d option has no greater specificity than the value-less one, and a team's own styles will trump both of them on the cascade alone anyway.

I feel pretty safe taking it out, I'm pretty confident it can't break anything.

36degrees:
I agree with your conclusion – anything that's covered by `button[disabled="disabled"]` would already be covered by `button[disabled]`. So I think we can safely get rid of it.

I'm not 100% sure about merging the two entries under a single changelog entry – I'm mainly wondering if the second one might get missed.

I do think we should refer back to it's deprecation, like [we did for the previous major release](https://github.com/alphagov/govuk-frontend/blob/main/CHANGELOG.md#remove-deprecated-govuk-main-wrapper-and-govuk-main-wrapper--l-mixins). If users have been acting on our deprecations they shouldn't have to do anything – adding this would help them make that connection.

PR #3491, opened 2023-04-04
Update Review app to import `govuk-frontend` via local package

colinrotherham:
This PR makes the Review app a user of `govuk-frontend` as part of:

* https://github.com/alphagov/govuk-frontend/issues/3280

For the performance work all source code and build artifacts can now be resolved from a single package

Additionally, the Review app now follows our own documentation:
https://frontend.design-system.service.gov.uk/importing-css-assets-and-javascript/

These areas in particular:

## Import CSS, assets and JavaScript
1. [Simplify Sass import paths](https://frontend.design-system.service.gov.uk/importing-css-assets-and-javascript/#simplify-sass-import-paths)
2. [Import JavaScript using a bundler](https://frontend.design-system.service.gov.uk/importing-css-assets-and-javascript/#import-javascript-using-a-bundler)

Which means the Review app now imports `govuk-frontend` ESM JavaScript for the first time, without having to navigate to `../../../src` sources or rely on a separate review-specific UMD bundle

## Package directory changes
Previously `npm link` only resolved the pre-built packages files but they're now consolidated:

### Before

```console
├── app
│   ├── src
│   └── dist
└── package
    ├── govuk
    ├── govuk-esm
    ├── govuk-prototype-kit
    ├── package.json
    └── README.md
```

### After

```console
└── packages
    ├── govuk-frontend
    │   ├── src
    │   │   ├── govuk
    │   │   └── govuk-prototype-kit
    │   ├── dist
    │   │   ├── govuk
    │   │   ├── govuk-esm
    │   │   └── govuk-prototype-kit
    │   ├── package.json
    │   └── README.md
    └── govuk-frontend-review
        ├── src
        └── dist
```

~`src`~ → `packages/govuk-frontend/src`
~`package`~ → `packages/govuk-frontendpackage/dist`

To maintain backwards compatibility these directory changes aren't found in the [published `package.json`](https://github.com/alphagov/govuk-frontend/blob/review-app-user/package/dist/package.json)

### The problem with `dist`
Unfortunately the standard convention of building `src` to `dist` doesn't fit

We already reserve `./dist` for our GitHub tagged releases, do we need to move it?
Will it be confusing to have a committed `./dist` alongside `./package/src` → `./package/dist` build output?

Some options:

1. `dist` — Leave it where it is
2. ~`dist`~ → `package/release` — Maintain two build directories under `./package`
3. ~`dist`~ → `package/dist/release` — Consolidate single build directory under `./package`

This PR has currently picked 1) from the list above

PR #3560, opened 2023-04-27
Update packages to follow GOV.UK Frontend docs

colinrotherham:
Quick PR split out from:

* https://github.com/alphagov/govuk-frontend/pull/3491

Which updates the Review app (and webpack example) to follow our docs more closely


app/src/common/middleware/assets.mjs
```diff
@@ -8,7 +8,7 @@ const router = express.Router()
 /**
  * Add middleware to serve static assets
  */
-router.use('/assets', express.static(join(paths.app, 'dist/assets')))
+router.use('/assets', express.static(join(paths.src, 'govuk/assets')))
```
colinrotherham, line 11: We'll swap this with `node_modules/govuk-frontend` once we've changed our package structure

shared/tasks/styles.mjs
```diff
@@ -78,6 +78,7 @@ export async function compileStylesheet ([modulePath, { srcPath, destPath, fileP
 
       // Resolve @imports via
       loadPaths: [
+        paths.src,
```
colinrotherham, line 81: We'll swap this with `node_modules/govuk-frontend` once we've changed our package structure

domoscargin (approved):
Looks good to me, all working fine locally.

PR #3559, opened 2023-04-27
Remove IE8–10 related Sass and CSS build tasks

querkmachine:
A big one, probably easier to go through it one commit at a time.

Because trimming out Internet Explorer 8–10 support (especially IE8) is such a big chunk of work, I've chosen to only focus on the Sass/CSS elements here. Any JavaScript-related pieces haven't been touched, nor have any aspects related to compatibility mode.

Resolves https://github.com/alphagov/govuk-frontend/issues/3461 and resolves https://github.com/alphagov/govuk-frontend/issues/2622 — except for documentation changes.

## Changes
- Removes instances of the `govuk-if-ie8` mixin and any code within them.
- Removes instances of the `govuk-not-ie8` mixin, maintaining any code within them.
- Removes all Sass settings and mixins relating to IE8.
- Removes all IE8 related Sass/CSS entry files.
- Removes tests relating to IE8 CSS generation.
- Removes references to IE8-specific stylesheets from the review app.
- Removes IE8-specific build steps from PostCSS configuration.
- Removes PostCSS dependencies that were only needed for IE8.
- Removes IE8–10 from Browserslist configuration.
- Removes IE8 related sections from Sassdoc generation.
- Removes the block on Autoprefixer updates being made by Dependabot, as IE8 is no longer a blocker to updating.
- Updates single colon pseudo-element selectors (`:before`) to the standards compliant double colon (`::before`). Single colons were only needed for IE8 support.
- Adds a temporary override to the Stylelint configuration to not raise failures for using double colons. This can be removed when the upstream Stylelint configuration is updated: https://github.com/alphagov/stylelint-config-gds/pull/36

## Notes

This PR doesn't:
- change anything in Nunjucks/HTML or JavaScript files.
- touch any CSS relating to compatibility mode, even though the legacy frontend toolkits also ship IE8-compatible code.
- update any documentation relating to IE8–10.
- remove any CSS hacks needed for IE11, even if they also apply to older IE versions.

Issue #3466, opened 2023-03-31
Remove the `nodeListForEach` helper function


romaricpascal:
## What

Replace all calls to the `nodeListForEach` helper function with calls to `NodeList.prototype.forEach` (of `for...of` depending on what we support).

Remove the `nodeListForEach` function and related tests.

## Why

As of GOV.UK Frontend v5.0, JavaScript will only run in browsers that support ES6 modules. All browsers that support ES6 modules also support NodeList.prototype.forEach, so can get rid of this helper function.

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [ ] Calls to `nodeListForEach` have been replaced with vanilla API calls
- [ ] The `nodeListForEach` function has been removed

36degrees:
Looks like this was done as part of #3570.

Issue #3462, opened 2023-03-31
Upgrade Rollup


romaricpascal:
## What

Upgrade Rollup and any other related dependencies to the latest version(s).

## Why

We previously pinned Rollup to maintain support for IE8. As [we're dropping support for IE8](#2621), we can unpin the dependency and use the latest version.

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [ ] Rollup has been upgraded to its latest version

colinrotherham:
Successfully closed by Dependabot (with a little helping hand)

* https://github.com/alphagov/govuk-frontend/pull/3591

> We should go back and simplify the `compileJavaScript` function in `tasks/scripts.mjs` as part of this work – see [#3386 (comment)](https://github.com/alphagov/govuk-frontend/pull/3386#discussion_r1155949622) for details

☝️ Regarding this bit above we sadly couldn't use `{ preserveModules }` whilst we still build UMD bundles, presumably due to the way it uses a factory function to add window globals etc

It'll become dead code once we ship ESM-only (or alongside CommonJS for compatibility) so can be removed later

Issue #3279, opened 2023-02-13
Track performance impacts of built files


romaricpascal:
## What

Monitor metrics to judge how our built files may affect performance of projects using `govuk-frontend`. This includes the files in `dist` as well as those inside `package`.

This should be as automated as we can to save us the time of gathering and comparing these metrics, but could be implemented progressively. 

## Why

While we do a `diff` of the files in `dist` before releasing, it'd be good to keep an eye on the size and contents of both `dist` and `package`, so we can spot  unexpected variations closer to when we make the changes. 

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Tasks

> **Note** Issues for each of those coming soon

- [ ] Gather built file sizes and display them
- [ ] Gather breakdown of module contents
- [ ] Display performance metrics on PRs
- [ ] Keep tracking impact of polyfills once automated
- [ ] Compare file sizes between branches
- [ ] Compare breakdowns of modules between branches
- [ ] Break build when impact is too big
- [ ] Measure impact of transpiling
- [ ] Investigate having only one build
- [ ] Investigate not having `dist` and `package` in the repository

stevenjmesser:
Moving this epic to the backlog so the individual tasks can come through the board.

colinrotherham:
Going to move this back to the Backlog for post v5 tasks

PR #3570, opened 2023-05-02
Remove IE8-10 JavaScript polyfills, helpers, config

colinrotherham:
This PR unblocks https://github.com/alphagov/govuk-frontend/issues/3279 by removing Rollup `{ legacy: true }` so we can pick up:

* https://github.com/alphagov/govuk-frontend/issues/3462

Similar to https://github.com/alphagov/govuk-frontend/pull/3559 this PR also tackles IE8-10 JavaScript things:

1. https://github.com/alphagov/govuk-frontend/issues/3466
2. https://github.com/alphagov/govuk-frontend/issues/2506 (but excludes those for IE11 for now)
3. Removes IE8 legacy support for Rollup and Terser


docs/contributing/coding-standards/js.md
```diff
@@ -146,10 +144,10 @@ Prefer named exports over default exports to avoid compatibility issues with tra
 
 If you need to support older browsers, import the necessary [polyfills](/packages/govuk-frontend/src/govuk/vendor/polyfills) and they will be added to the environment when the feature is not supported.
 
-For example, if you want to polyfill `addEventListener` for IE8, import the Event polyfills.
+For example, if you want to polyfill `Element.closest()` for IE11, import the closest polyfill.
```
querkmachine, line 147: Given we also won't be shipping JS for IE11 in future, would it be better to use an example for a different browser?

Safari has a couple of potential candidates. I believe we're currently intending to support Safari 11+ in future (albeit at the lowest 'passing' grade). 

- [Intersection Observers](https://caniuse.com/intersectionobserver) support was added in Safari 12.1.
- [`String.prototype.replaceAll`](https://caniuse.com/mdn-javascript_builtins_string_replaceall) support was added in Safari 13.1.
- Some features of [`Intl.NumberFormat`](https://caniuse.com/mdn-javascript_builtins_intl_numberformat_formatrange) weren't implemented until Safari 15.4.

packages/govuk-frontend/src/govuk/all.mjs
```diff
@@ -1,5 +1,4 @@
 import { version } from './common/govuk-frontend-version.mjs'
-import { nodeListForEach } from './common/index.mjs'
```
querkmachine, line 2: Animated GIF of a young Asian girl saying "The evil is defeated."

packages/govuk-frontend/src/govuk/common/index.mjs
```diff
@@ -8,26 +8,6 @@
...
- */
-export function nodeListForEach (nodes, callback) {
-  if (window.NodeList.prototype.forEach) {
-    return nodes.forEach(callback)
-  }
-  for (var i = 0; i < nodes.length; i++) {
-    callback.call(window, nodes[i], i, nodes)
-  }
-}
-
```
querkmachine, line 30: Animated GIF of a young Asian girl saying "The evil is defeated."

docs/contributing/coding-standards/js.md
colinrotherham: Yeah I picked this one as we only have `classList`, `closest` and `matches` now to choose from

They're all primarily for IE11 until we flick the `<script type="module">` switch

Until we have new polyfill candidates (or remove the document entirely) shall we keep these changes?

packages/govuk-frontend/src/govuk/components/error-summary/error-summary.mjs
```diff
@@ -1,10 +1,6 @@
-/* eslint-disable es-x/no-function-prototype-bind -- Polyfill imported */
-
 import { mergeConfigs } from '../../common/index.mjs'
 import { normaliseDataset } from '../../common/normalise-dataset.mjs'
 import '../../vendor/polyfills/Element/prototype/closest.mjs'
```
querkmachine, line 3: Is this polyfill still needed if we're dropping JS support in IE11 as well?

colinrotherham: Yeah but since it's still a supported browser I've left it in until we pick up `<script type="module">`

Otherwise IE11 will load and run code that will crash

docs/contributing/coding-standards/js.md
querkmachine: Happy to keep it if you want, just conscious of it being something that'll need to be changed again in future.

colinrotherham: Yeah let's keep it. Until the day we have a new polyfilling approach, this remains our most up-to-date strategy

colinrotherham: We can pick it up as part of:

* https://github.com/alphagov/govuk-frontend/issues/2159

PR #3571, opened 2023-05-03
Add generic comment on PR script

### LINKED ISSUE #3545: Introduce reusable action/workflow/script for commenting on PRs reusing the same comment
## What

Create a reusable action, workflow or script that lets us post comments on Github PRs when they're created and update that initial comment when new commits are made to the PR.[^1]

## Why

We're looking to post more informations on our PRs to facilitate the reviews. This is the foundation that'll help us do so.

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

```[tasklist]
- [x] We've added a reusable piece of tooling we can use to comment on PRs
- [x] The diff of `dist` uses that reusable piece for commenting
```

[^1]: Besides our current `diff` action, [the spike for surfacing performance information](https://github.com/alphagov/govuk-frontend/pull/3239) also does some commenting (using a JavaScript function in the hopes to lead to something more testable)

domoscargin:
This is the MVP for a script to comment on PRs.

It can be passed the Github context, a marker string to identify which comment to target, and the content of the comment. The idea is to get this functionality into the codebase soon, then work on improvements such as:

- Creating the "stats" workspace to contain this script and other performance monitoring stuff
- Adding the ability to use workflow parameter inputs to define A and B inputs for diffing

Closes #3545

### DIFF
--- a/.github/workflows/diff-change-to-dist.yaml
+++ b/.github/workflows/diff-change-to-dist.yaml
@@ -52,13 +52,16 @@ jobs:
           github-token: ${{secrets.GITHUB_TOKEN}}
           script: |
             const { readFile } = require('node:fs/promises')
-            const { commentOnPr } = await import('${{ github.workspace }}/.github/workflows/scripts/comments.mjs')
+            const { comment } = await import('${{ github.workspace }}/.github/workflows/scripts/comments.mjs')
 
             // Read diff from previous step
             const diffText = await readFile('${{ github.workspace }}/dist.diff', 'utf8')
 
+            // PR information
+            const number = '${{ github.event.pull_request.number }}'
+
             // Add or update comment on PR
-            await commentOnPr({ github, context }, 'Diff release', '## Changes to dist\n' +
+            await comment({ github, context }, number, 'Diff release', '## Changes to dist\n' +
               '```diff\n' +
               `${diffText}\n` +
               '\nSHA: ${context.sha}\n' +
--- a/.github/workflows/diff-change-to-dist.yaml
+++ b/.github/workflows/diff-change-to-dist.yaml
@@ -59,11 +59,12 @@ jobs:
 
             // PR information
             const number = '${{ github.event.pull_request.number }}'
+            const commit = '${{ github.event.pull_request.head.sha }}'
 
             // Add or update comment on PR
             await comment({ github, context }, number, 'Diff release', '## Changes to dist\n' +
               '```diff\n' +
               `${diffText}\n` +
-              '\nSHA: ${context.sha}\n' +
-              '```'
+              '```\n\n' +
+              `SHA: ${commit}\n`
             )
--- a/.github/workflows/diff-change-to-dist.yaml
+++ b/.github/workflows/diff-change-to-dist.yaml
@@ -9,10 +9,12 @@ permissions:
 
 jobs:
   generate-diff:
-    name: Generate Diff
+    name: Generate diff
     runs-on: ubuntu-latest
-    # Abort if run from a fork, as token won't have write access to post comment
-    if: github.event.pull_request.head.repo.full_name == github.repository
+
+    # Skip when token write permissions are unavailable on forks
+    if: ${{ !github.event.pull_request.head.repo.fork }}
+
     steps:
       - uses: actions/checkout@v3
         with:
@@ -49,19 +51,16 @@ jobs:
         with:
           github-token: ${{secrets.GITHUB_TOKEN}}
           script: |
-            const fs = require('fs').promises
-            const diff = await fs.readFile(
-              process.env.GITHUB_WORKSPACE + '/dist.diff', 'utf8'
-            )
+            const { readFile } = require('node:fs/promises')
+            const { commentOnPr } = await import('${{ github.workspace }}/.github/workflows/scripts/comments.mjs')
 
-            const commentText = '## Changes to dist\n' +
+            // Read diff from previous step
+            const diffText = await readFile('${{ github.workspace }}/dist.diff', 'utf8')
+
+            // Add or update comment on PR
+            await commentOnPr({ github, context }, 'Diff release', '## Changes to dist\n' +
               '```diff\n' +
-              diff +
-              '\n```'
-
-            github.rest.issues.createComment({
-              issue_number: context.payload.pull_request.number,
-              owner: context.repo.owner,
-              repo: context.repo.repo,
-              body: commentText
-            })
+              `${diffText}\n` +
+              '\nSHA: ${context.sha}\n' +
+              '```'
+            )
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -3,6 +3,9 @@
   "files": [],
   "references": [
     {
+      "path": "./.github/workflows/scripts"
+    },
+    {
       "path": "./packages/govuk-frontend-review"
     },
     {
--- a/.github/workflows/scripts/comments.mjs
+++ b/.github/workflows/scripts/comments.mjs
@@ -1,0 +1,57 @@
+/**
+ * @param {object} githubContext - Github Action context
+ * @param {import('@octokit/rest').Octokit} githubContext.github
+ * @param {import('@actions/github').context} githubContext.context
+ * @param {string} markerText - A unique marker used to identify the correct comment
+ * @param {string} bodyText - The body of the comment
+ */
+export async function commentOnPr ({ github, context }, markerText, bodyText) {
+  if (!context.payload.pull_request) {
+    throw new Error('Workflow "comment on PR" requires a PR')
+  }
+
+  const marker = `<!-- ${markerText} -->\n`
+  const body = `${marker}${bodyText}`
+
+  let commentId
+
+  /**
+   * @satisfies {IssueCommentsListParams}
+   */
+  const issueParameters = {
+    issue_number: context.payload.pull_request.number,
+    owner: context.repo.owner,
+    repo: context.repo.repo
+  }
+
+  /**
+   * Finds the first issue comment for which the `matcher` function returns `true`
+   *
+   * {@link https://github.com/peter-evans/find-comment/blob/main/src/find.ts}
+   */
+  for await (const { data: comments } of github.paginate.iterator(
+    github.rest.issues.listComments,
+    issueParameters
+  )) {
+    const comment = comments.find((comment) => !!comment.body?.includes(marker))
+
+    if (comment) {
+      commentId = comment.id
+    }
+  }
+
+  /**
+   * Create GitHub issue comment
+   *
+   * Updates existing commentby ID if available
+   */
+  await (!commentId
+    ? github.rest.issues.createComment({ ...issueParameters, body })
+    : github.rest.issues.updateComment({ ...issueParameters, body, comment_id: commentId }))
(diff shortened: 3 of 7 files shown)


.github/workflows/diff-change-to-dist.yaml
```diff
@@ -51,9 +52,4 @@ jobs:
               diff +
               '\n```'
 
-            github.rest.issues.createComment({
-              issue_number: context.payload.pull_request.number,
-              owner: context.repo.owner,
-              repo: context.repo.repo,
-              body: commentText
-            })
+            await commentOnPr({github,context}, '## Changes to dist', commentText)
```
romaricpascal, line 55: Apologies if it's too early to comment or something you may have already tried. During the spike, [Ollie suggested using an HTML comment as a marker](https://github.com/alphagov/govuk-frontend/pull/3239#issuecomment-1422463273) rather than visible content from the comment. I think that's worth a shot as it'd allow to completely encapsulate what lets the comment be found again in `commentOnPr` and let us change the text at will 😊

.github/workflows/diff-change-to-dist.yaml
```diff
@@ -42,18 +42,15 @@ jobs:
...
           script: |
             const fs = require('fs').promises
+            const commentOnPr = require('./shared/stats/comment.js').commentOnPr
             const diff = await fs.readFile(
               process.env.GITHUB_WORKSPACE + '/dist.diff', 'utf8'
             )
 
-            const commentText = '## Changes to dist\n' +
+            const commentText = '<!-- Diff release -->' +
+              '## Changes to dist\n' +
```
romaricpascal, line 51: **issue** I think this needs a line break after the comment to make sure Github interprets the next line as Markdown. It displays as `## Changes to dist` in the [latest comment on this PR](https://github.com/alphagov/govuk-frontend/pull/3571#issuecomment-1539670369).

**suggestion** It's also be neat to move the comment part inside `commentOnPr` so that comment + line break is hidden from whatever script is trying to comment and that script just needs to provide a simple string ("Diff release" in this case) as a marker 😊

romaricpascal (commented):
Cheers for bringing that in and hunting the types to make it all proper 😊 Spotted a little issue with the [last comment automatically added on this PR](https://github.com/alphagov/govuk-frontend/pull/3571#issuecomment-1539670369) and got a little suggestion for the API.

From having dug a little more into this, do you think there could be a way to add automated tests to the `commentOnPr` as well or would we have to maintain our own mocks of Github's API and responses for that?

.github/workflows/diff-change-to-dist.yaml
domoscargin: Thanks for the suggestions!

I've shifted the comment into the function itself, so just pass in a simple string.

I've also popped a SHA on the comment to help us identify what the comment applies to.

I had a glance at testing - it's easy to mock the `context` object, but the Octokit/Github API is a bit trickier. There's a couple of ready-made mock Github APIs out there (https://github.com/kiegroup/mock-github), or we could try the fixtures route (https://github.com/octokit/fixtures) - don't see a premade scenario for comment actions, though.

romaricpascal (approved):
Looking neat 🙌🏻 The testing seems quite costly given there's no out of the box help. I'm fine if we leave it to later, especially as we're planning to have more regular comments on PRs which will let us spot quickly if something's off.

colinrotherham:
Looking good @domoscargin

Mind if we spend a bit of timing wiring up `shared/stats` with a **package.json** and **tsconfig.json**?

So we can move those new `optionalDependencies` only where needed and enable type checks

.github/workflows/diff-change-to-dist.yaml
```diff
@@ -42,6 +42,7 @@ jobs:
           github-token: ${{secrets.GITHUB_TOKEN}}
           script: |
             const fs = require('fs').promises
+            const commentOnPr = require('./shared/stats/comment.js').commentOnPr
```
colinrotherham, line 45: If you'd prefer to write the new script in ESM it's compatible with CommonJS via `await import()`

```suggestion
            const { commentOnPr } = await import('${{ github.workspace }}/shared/stats/comment.mjs')
```

The example came from [their docs](https://github.com/actions/github-script/tree/v6/#use-esm-import) but I'm assuming `process.env.GITHUB_WORKSPACE` would work too (as we've used for reading the file a bit further down)

.github/workflows/diff-change-to-dist.yaml
```diff
@@ -42,6 +42,7 @@ jobs:
           github-token: ${{secrets.GITHUB_TOKEN}}
           script: |
             const fs = require('fs').promises
```
colinrotherham, line 44: Not a big one but Node.js 14 added `fs/promises` directly

```suggestion
            const { readFile } = require('fs/promises')
```

Although we don't do it (optional) it's good to prefix core packages with `node:` in Node.js 16+ too

```suggestion
            const { readFile } = require('node:fs/promises')
```

.github/workflows/diff-change-to-dist.yaml
```diff
@@ -51,9 +52,4 @@ jobs:
               diff +
               '\n```'
 
-            github.rest.issues.createComment({
-              issue_number: context.payload.pull_request.number,
-              owner: context.repo.owner,
-              repo: context.repo.repo,
-              body: commentText
-            })
+            await commentOnPr({github,context}, 'Diff release', commentText)
```
colinrotherham, line 55: Love how this is all tucked away and tidy now 😊 

Worth running these snippets through ESLint?

```suggestion
            await commentOnPr({ github, context }, 'Diff release', commentText)
```

.github/workflows/diff-change-to-dist.yaml
```diff
@@ -42,6 +42,7 @@ jobs:
           github-token: ${{secrets.GITHUB_TOKEN}}
           script: |
             const fs = require('fs').promises
+            const commentOnPr = require('./shared/stats/comment.js').commentOnPr
             const diff = await fs.readFile(
```
colinrotherham, line 46: I doubt we'll ever run into it but good to have Windows compatibility in mind using `path.join()`

```suggestion
            const { GITHUB_WORKSPACE } = process.env

            const diff = await readFile(join(GITHUB_WORKSPACE, 'dist.diff', 'utf8'))
```

shared/stats/comment.js
```diff
@@ -0,0 +1,68 @@
...
+ * @param {import('@octokit/rest').Octokit} githubContext.github
+ * @param {import('@actions/github').context} githubContext.context
+ * @param {string} markerText - A unique marker used to identify the correct comment
+ * @param {string} bodyText - The body of the comment
+ * @returns {Promise<object>} Github request object
+ */
+const commentOnPr = async function ({ github, context }, markerText, bodyText) {
+  const marker = `<!-- ${markerText} -->\n`
+  const sha = `\nSHA: ${context.sha}`
+  const body = marker + bodyText + sha
```
colinrotherham, line 13: Would this look a bit more like code if we stole `outdent` from elsewhere?

```suggestion
  const body = outdent`
    <!-- ${markerText} -->
    ${bodyText}

    SHA: ${context.sha}
  `
```

shared/stats/comment.js
```diff
@@ -0,0 +1,68 @@
+/**
+ *
+ * @param {object} githubContext - Github Action context
+ * @param {import('@octokit/rest').Octokit} githubContext.github
+ * @param {import('@actions/github').context} githubContext.context
+ * @param {string} markerText - A unique marker used to identify the correct comment
+ * @param {string} bodyText - The body of the comment
+ * @returns {Promise<object>} Github request object
+ */
+const commentOnPr = async function ({ github, context }, markerText, bodyText) {
```
colinrotherham, line 10: Super tiny thing but named functions will appear in stack traces

```suggestion
export async function commentOnPr ({ github, context }, markerText, bodyText) {
```

colinrotherham: Using `export` here (versus `module.exports` later) assumes we'd rename to `*.mjs` too

.github/workflows/diff-change-to-dist.yaml
colinrotherham: Wonder if that's because they've faked `require()` via [wrap-require.ts](https://github.com/actions/github-script/blob/main/src/wrap-require.ts)?

Any better with `const { join } = require('node:path')`?

domoscargin: 🤦 My bad, the complaint about `path` was from a previous attempt to fix where I was using `path.join`.

domoscargin:
I've implemented some of your suggestions, @colinrotherham. Ready for review again!

shared/stats/package.json
```diff
@@ -0,0 +1,15 @@
...
+  "private": true,
+  "name": "govuk-frontend-stats",
+  "description": "GOV.UK Frontend shared stats",
+  "main": "index.mjs",
+  "engines": {
+    "node": "^18.12.0",
+    "npm": "^8.1.0 || ^9.1.0"
+  },
+  "license": "MIT",
+  "dependencies": {
```
colinrotherham, line 11: Does it break anything if we set these to `optionalDependencies`?

I try to split things into:

1. `dependencies` → Published or deployed (Heroku) packages
2. `devDependencies` → Development or test packages only
3. `optionalDependencies` → Nice to haves like JSDoc types

We added a [little workaround for Heroku](https://github.com/alphagov/govuk-frontend/commit/9a08a406359e8c761e23fd449b9c96ab097bff95) but without it these two `dependencies` would be deployed

```suggestion
  "optionalDependencies": {
```

tsconfig.json
```diff
@@ -19,6 +19,9 @@
     },
     {
       "path": "./shared/tasks"
+    },
+    {
+      "path": "./shared/stats"
```
colinrotherham, line 24: I'm still not 100% sold on this being in "shared"

Only because it's not shared, it's just used once in a single place

What do you think?

domoscargin: Are we only ever going to use the performance stuff in github actions? I was under the impression that we were going to use it elsewhere.

But if not, yup, makes total sense to move it. If we want to split out the github bits from the performance bit, we can take @romaricpascal's suggestion and pop it somewhere in `.github`?

Otherwise just pop the whole thing in the base directory?

romaricpascal: The way I imagined it we'd run the performance stuff:
- from Github actions, to comment on PR
- locally, if we want a glimpse before commiting

Not sure if that makes it "shared". As an other thought, "shared" may actually be more "internal" to widen its responsibilities without us having to worry about how shared a piece of code is.

The Github commenting itself will be shared for sure, though. Hence my previous suggestion of having it as part of `.github`. Alternatively, it may fit neatly into the `helpers` package (as `helpers/github.js` for example). It'd feel neater than having it in the base directory? Would that solve the "shared" aspect of it?

colinrotherham: @romaricpascal Only asking because I've got to find a home for several other performance packages

Helpers sounds good 😊

colinrotherham (approved):

domoscargin (commented):
Looks good to me, thanks @colinrotherham

romaricpascal (approved):
Neat one! Thanks both! ⛵

PR #3573, opened 2023-05-03
Update browser support in README

domoscargin:
Since the Service Manual [removed IE11 from their recommended browsers](https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices#browsers-to-test-in), our README has been a bit out of date, so I've explicitly added it back.

I'm not sure if I should be as explicit with Safari 9, 10, 11, since Safari 12.1+ is recommended by the Service Manual.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ files into your project](https://frontend.design-system.service.gov.uk/install-u
 
 GOV.UK Frontend supports:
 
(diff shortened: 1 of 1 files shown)


colinrotherham (approved):

colinrotherham:
Looks good for now, flagged lots of other docs in this comment to pick up soon:

* https://github.com/alphagov/govuk-frontend/pull/3559#issuecomment-1527577702

PR #3574, opened 2023-05-04
Remove legacy and tabular fonts support

### LINKED ISSUE #2791: Remove legacy font setting and tabular font separation setting from codebase
## What
Remove the setting [`$govuk-use-legacy-font`](https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/settings/_typography-font.scss#L17) and all its uses in govuk-frontend, as well as the settings it informs:

- [The font family definitions file](https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/settings/_typography-font-families.scss)
- [`$govuk-font-family-tabular`](https://github.com/alphagov/govuk-frontend/blob/c31c4bebba9a464116033ff5cd03772bf3aa79f5/src/govuk/helpers/_typography.scss#L186-L199)
 
## Why
Required to remove compatibility mode completely from govuk-frontend in 5.0.

## Who needs to work on this
Developers

## Done when

- [x] Remove from back link component
- [x] Remove from breadcrumbs component
- [x] Remove from button component
- [x] Remove from header component
- [x] Remove from tag component
- [x] Remove font family file and re-write font definition to always point to GDS Transport
- [x] Remove `$govuk-font-family-tabular` and reconfigure `govuk-font`
- [x] Remove setting completely

querkmachine:
Removes `$govuk-use-legacy-font` and `$govuk-font-family-tabular` settings. 

Resolves #2791.

## Changes
- Removes component code that's conditional on `$govuk-use-legacy-font` being true. 
- Removes code that's conditional on `$govuk-font-family-tabular` being true from `govuk-font` mixin.
- Removes tests relating to `$govuk-font-family-tabular`.
- Removes `$govuk-use-legacy-font`, `$govuk-font-family-tabular`, `$govuk-font-family-nta` and `$govuk-font-family-nta-tabular` settings.
- Updates `$govuk-font-family` setting to always use GDS Transport by default.
- Updates tests that used `$govuk-font-family-tabular` as dummy data (usually as a shorthand for a non-GDS Transport font) to use Helvetica instead.

### DIFF
--- a/packages/govuk-frontend/src/govuk/helpers/typography.test.js
+++ b/packages/govuk-frontend/src/govuk/helpers/typography.test.js
@@ -54,7 +54,7 @@ describe('@mixin govuk-typography-common', () => {
         @include govuk-typography-common;
       }
       :root {
-        @include govuk-typography-common($font-family: $govuk-font-family-tabular);
+        @include govuk-typography-common($font-family: Helvetica);
       }
     `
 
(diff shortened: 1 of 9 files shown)


packages/govuk-frontend/src/govuk/helpers/typography.test.js
```diff
@@ -54,7 +54,7 @@ describe('@mixin govuk-typography-common', () => {
         @include govuk-typography-common;
       }
       :root {
-        @include govuk-typography-common($font-family: $govuk-font-family-tabular);
+        @include govuk-typography-common($font-family: Helvetica);
```
colinrotherham, line 57: Feels odd to have `Helvetica` here without a matching check for the output?

I appreciate we didn't before this edit, but we did (further down) in the tabular test

```js
await expect(results).resolves.toMatchObject({
  css: expect.stringContaining('font-family: Helvetica')
})

await expect(results).resolves.toMatchObject({
  css: expect.stringContaining('font-family: "GDS Transport"')
})
```

colinrotherham (approved):
Looks brilliant

(I'm just going to rebase this btw, will let you know when it's done)

packages/govuk-frontend/src/govuk/helpers/typography.test.js
colinrotherham: Not a blocker

Issue #2790, opened 2022-08-19
Remove px instead of rem setting from codebase


owenatgov:
## What
Remove the setting [`$govuk-typography-use-rem`](https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/settings/_typography-responsive.scss#L17) and all its uses in govuk-frontend.
 
## Why
Required to remove compatibility mode completely from govuk-frontend in 5.0.

Part of https://github.com/alphagov/govuk-frontend/issues/2769

## Who needs to work on this
Developers

## Done when

- [x] Remove from input component
- [x] Remove from select component
- [x] Remove from typography helper
- [x] Remove setting completely

36degrees:
We'll need to be careful in how we communicate this – we've added deprecation warnings for it this flag is **set** in consumer code, but consumers might be using it in their own logic (and not setting it).

For example, there are components in GOV.UK Publishing Components that do things like:

```scss
  @if $govuk-typography-use-rem {
    font-size: govuk-px-to-rem(80px);
  }
```

In most cases `$govuk-typography-use-rem` will currently be set to `true`, but once we remove this flag the consumer code will **silently** stop outputting the font-size in rem – and this is unlikely to be noticed unless the person updating manually tests zooming text only (or they have visual regression tests that cover this).

(I think there's a lesson here in making future feature flags 'fail open' rather than 'fail closed')

PR #3575, opened 2023-05-04
Remove `$govuk-typography-use-rem` setting

owenatgov:
## What/Why
Removes all internal references to `$govuk-typography-use-rem`, references to it in internal documentation, tests that reference it and cover it and the setting itself.

Resolves https://github.com/alphagov/govuk-frontend/issues/2790


querkmachine:
Seems like some output tests need updating to account for the px values no longer being returned outside of print styles, but otherwise LGTM 👍

PR #3584, opened 2023-05-05
Fix URL `file://` path handling on Windows

colinrotherham:
This PR fixes two Windows regressions added in https://github.com/alphagov/govuk-frontend/pull/3527 and https://github.com/alphagov/govuk-frontend/pull/3372

URL file paths with drive letters `D:/` confuse [`path.relative()`](https://nodejs.org/docs/latest-v18.x/api/path.html#pathrelativefrom-to) and throw errors in [`await import()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import):

```console
Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only URLs with a scheme in: file and data are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'd:'
```

## File URLs with absolute paths

On Windows, Node.js `file://` URL paths look like:

```js
// Example file URL string
const file = 'file:///D:/path/to/asset'
```

### Converting URLs to paths incorrectly

But the property [`new URL(file).pathname`](https://nodejs.org/docs/latest-v18.x/api/url.html#urlpathname) includes a leading forward slash:

```js
// Example file URL pathname ❌
const pathname = '/D:/path/to/asset' 
```

We hadn't realised this breaks [`path.relative()`](https://nodejs.org/docs/latest-v18.x/api/path.html#pathrelativefrom-to) on Windows with paths joined and not resolved:

```js
'..\\..\\..\\..\\D:\\path\\to\\project\\packages\\govuk-frontend\\src\\govuk\\components\\character-count\\_index.scss'
```

### Converting URLs to paths correctly

After investigating this issue (also ["Converting between URLs and file paths"](https://exploringjs.com/nodejs-shell-scripting/ch_nodejs-path.html#converting-between-urls-and-file-paths)) we see that:

1. Compatible cross-platform “path to file URL” changes must use [`pathToFileURL()`](https://nodejs.org/docs/latest-v18.x/api/url.html#urlpathtofileurlpath)
2. Compatible cross-platform “file URL to path” changes must use [`fileURLToPath()`](https://nodejs.org/docs/latest-v18.x/api/url.html#urlfileurltopathurl)

```js
// Example file URL via fileURLToPath() ✅ 
const path = 'D:\\path\\to\\asset'
```

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -48,6 +48,9 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
 
+      - name: Restore dependencies
+        uses: ./.github/workflows/actions/install-node
+
       - name: Build
         uses: ./.github/workflows/actions/build
 
--- a/.github/workflows/actions/build/action.yml
+++ b/.github/workflows/actions/build/action.yml
@@ -13,12 +13,6 @@ runs:
         key: build-cache-${{ runner.os }}-${{ github.sha }}
         path: packages/*/dist
 
-    - name: Install dependencies
-      uses: ./.github/workflows/actions/install-node
-
-      # Skip install when build is already cached
-      if: steps.build-cache.outputs.cache-hit != 'true'
-
     - name: Build
       id: build
 
--- a/shared/tasks/assets.mjs
+++ b/shared/tasks/assets.mjs
@@ -1,4 +1,7 @@
 import { dirname, parse, relative } from 'path'
+import { fileURLToPath } from 'url'
+
+import slash from 'slash'
 
 import { files } from './index.mjs'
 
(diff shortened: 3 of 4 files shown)


colinrotherham:
Test runs on Windows must be run manually:
https://github.com/alphagov/govuk-frontend/actions/runs/4892616438

GitHub Actions UI to run on Windows

.github/workflows/tests.yml
```diff
@@ -48,6 +48,9 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
 
+      - name: Restore dependencies
+        uses: ./.github/workflows/actions/install-node
```
colinrotherham, line 52: Had to move this line here for Windows to find `node_modules/.bin/tsc`

Unlike our Linux runners its not in the global “npm get cache” location at `~/.npm`

```
> build:types
> tsc --build tsconfig.build.json

'tsc' is not recognized as an internal or external command,
operable program or batch file.
```

romaricpascal (approved):
Cheers for catching that! Looks good to me ⛵

Issue #3427, opened 2023-03-28
`_warning` mixin cannot be used within functions 


querkmachine:
## Description of the issue

We use [a `_warning` mixin](https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/settings/_warnings.scss#L47) in Sass to standardise our deprecation warnings and provide a mechanism for service teams to disable intrusive and repetative warnings. 

Our documentation for 'managing change' includes a section about [using the mixin within a function](https://github.com/alphagov/govuk-frontend/blob/main/docs/contributing/managing-change.md#deprecating-a-sass-function-or-mixin).

However, it isn't actually possible to call a Sass mixin from within a function. Doing so returns an error in all Sass compilers that we currently support in Frontend:

- Dart Sass — `This at-rule is not allowed here.`
- libsass — `Functions can only contain variable declarations and control directives.`
- Ruby Sass — `Functions can only contain variable declarations and control directives.`

## Steps to reproduce the issue

Try to include the `_warning` mixin (or any mixin) from inside of a function.

## Actual vs expected behaviour

Presumably the documentation is incorrect and needs to be amended.

We may need to devise a method of producing (and suppressing) warnings that are spawned by functions. 

## Environment

- GOV.UK Frontend version: 4.5.0
- Dart Sass version tested: 1.32.12
- libsass version tested: 3.5.5
- Ruby Sass version tested: 3.7.4

owenatgov:
Good catch. Really this should be a function, I'm not sure why it's not 🤔

querkmachine:
Probably because being a mixin is the only way that does it 'right'.

It seems like Sass only likes function calls to be used in certain ways. Changing the `_warning` mixin to a function and trying to call it stadalone like this produces a syntax error on compile:

```scss
@function get-colour {
  @if true {
    _warning("doo-wop", "woo-bop");
  }
  @return #786999;
}

body {
  color: get-colour();
}
```

Changing it to use the `@return` statement resolves the syntax error, but also negates its use as a warning, because the function will no longer return the colour — 'breaking' the output until the source of the warning has been fixed.

```scss
@function get-colour {
  @if true {
    @return _warning("doo-wop", "woo-bop");
  }
  @return #786999;
}

body {
  color: get-colour();
}
```

Trying to use it within a mixin will also only work if it's attached to a property (real or not).


```scss
@mixin nice-colours {
  warning: _warning("doo-wop", "woo-bop");
  color: white;
[...]

PR #3585, opened 2023-05-05
Add deprecation warning for `govuk-colour`'s `legacy` parameter

querkmachine:
We missed deprecating the `$legacy` parameter of the `govuk-colour` mixin when we deprecated the other legacy colour palette code in [v4.4.0](https://github.com/alphagov/govuk-frontend/releases/tag/v4.4.0). As we want to remove this parameter in v5, we should deprecate it beforehand. 

This doesn't use our usual `_warning` mixin due to [mixins not being allowed within functions](https://github.com/alphagov/govuk-frontend/issues/3427).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,18 @@
 
 ## Unreleased
 
+## 4.6.1 (Patch release)
+
+### Deprecated features
+
+#### Stop using legacy colour palette
+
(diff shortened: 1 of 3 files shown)


querkmachine:
Decided not to do a v4.6.x patch release at this time, as it would involve some refactoring to split `govuk-colour` into public and private variants so that we can suppress internal warnings. 

We'll revisit it if we find other reasons to do a v4 patch or feature release in future.

PR #3593, opened 2023-05-09
Remove deprecated static spacing classes

### LINKED ISSUE #2782: Remove deprecated `govuk-!-margin-static` and `govuk-!-padding-static` classes
## What
In [v4.3.1](https://github.com/alphagov/govuk-frontend/issues/2771) we added the `govuk-!-static` classes alongside the incorrectly named `govuk-!-margin-static` and `govuk-!-padding-static` classes and marked those incorrect ones as deprecated. We should remove the deprecated classes in our next breaking release (v5.0.0)

## Why
We said in the v4.3.1 release notes that we were deprecating the `govuk-!-margin-static` and `govuk-!-padding-static` classes and that we'd remove them in a future breaking release. We should stick to what we said and actually do it.

## Who needs to work on this
Developers

## Who needs to review this
Developers

## Done when

- [x] Remove the incorrectly ordered static spacing override classes [here](https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/overrides/_spacing.scss#L82) and [here](https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/overrides/_spacing.scss#L88)

querkmachine:
Removes the erroneously named `.govuk-!-margin-static` and `.govuk-!-padding-static` classes introduced in v4.3.0 and deprecated in v4.3.1.

Closes #2782.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,14 @@ We no longer support link buttons being disabled or using disabled styles.
 
 This change was introduced in [pull request #3557: Remove deprecated `.govuk-button--disabled` class](https://github.com/alphagov/govuk-frontend/pull/3557).
 
+#### Remove deprecated `govuk-!-margin-static` and `govuk-!-padding-static` classes
+
+We've removed the override classes starting with `.govuk-!-margin-static` and `.govuk-!-padding-static` which were deprecated in [GOV.UK Frontend v4.3.1](https://github.com/alphagov/govuk-frontend/releases/tag/v4.3.1).
+
+Use the classes starting with `.govuk-!-static-margin` and `.govuk-!-static-padding` instead.
+
+This change was introduced in [pull request #3593: Remove deprecated static spacing classes](https://github.com/alphagov/govuk-frontend/pull/3593).
+
 #### Update the HTML for warning text
 
 We've removed the `.govuk-warning-text__assistive` class and its styles from GOV.UK Frontend. This class is unnecessary, as it duplicates the functionality of the `.govuk-visually-hidden` class already present in Frontend.
--- a/packages/govuk-frontend/src/govuk/overrides/_spacing.scss
+++ b/packages/govuk-frontend/src/govuk/overrides/_spacing.scss
@@ -61,14 +61,6 @@ $_spacing-directions: (
 /// Generate spacing override classes for the given property (e.g. margin)
 /// for each point in the non-responsive spacing scale.
 ///
-/// The classes in the format govuk-#{$property}-static-#{$spacing-point}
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -40,6 +40,14 @@ We no longer support link buttons being disabled or using disabled styles.
 
 This change was introduced in [pull request #3557: Remove deprecated `.govuk-button--disabled` class](https://github.com/alphagov/govuk-frontend/pull/3557).
 
+#### Remove deprecated `govuk-!-margin-static` and `govuk-!-padding-static` classes
+
+We've removed `.govuk-!-margin-static`, `.govuk-!-padding-static` and their related classes that were deprecated in [GOV.UK Frontend v4.3.1](https://github.com/alphagov/govuk-frontend/releases/tag/v4.3.1).
+
+Use the correctly named `.govuk-!-static-margin` and `.govuk-!-static-padding` classes instead.
```
36degrees, line 47: Suggesting a few tweaks just to try and mirror the language used in the original deprecation:

```suggestion
We've removed the override classes starting with `.govuk-!-margin-static` and `.govuk-!-padding-static` which were deprecated in [GOV.UK Frontend v4.3.1](https://github.com/alphagov/govuk-frontend/releases/tag/v4.3.1).

Use the classes starting with `.govuk-!-static-margin` and `.govuk-!-static-padding` instead.
```

(We're not 100% consistent whether we include the leading dot on class names – can see examples of both. Might be something for us to look at when we come to review the 'finished' release notes for v5.)

36degrees (approved):
Some of the best 'deleting things' I've seen today 👏🏻 

Have suggested some minor tweaks to the release note but I don't consider them blocking.

CHANGELOG.md
querkmachine: Changelog tweaks are welcome! I was a umm-ing and aah-ing a bit over how to word it without just repeating everything from the v4.3.1 release notes.

querkmachine: With regards to the dot on class names, whilst I'm thinking about it, I think it makes sense to write the names of things similarly to how they'd be written in CSS, as the syntax helps disambugate them a bit.

For example, if something says "remove the `title`" without explicitly qualifying what 'title' is, that could be interpreted as the title element, the title attribute, something with 'title' as a class name or ID, a Nunjucks macro, a parameter, or any other kind of thing.

By comparison, the intentioned meanings of `<title>`, `[title]`, `.title` and `#title` are all fairly clear to someone with some familiarity of frontend development without the need for a qualifier, which (in my mind) is the primariy audience for documents like the changelog and release notes. 

The best option is probably to do both, though!

Issue #2204, opened 2021-04-26
Known issue: hover state is mis-painted on links within CSS multiple-column layouts


EoinShaughnessy:
## Description

When you hover your cursor over a link, only the thickness of that link's underline should change.

However, if you have links within a [multiple-column layout](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Columns) and your browser uses Chromium, then hovering your cursor over a link also changes the thickness of other, non-hovered link underlines.

![image](https://user-images.githubusercontent.com/72507742/116082284-2371dc80-a693-11eb-928e-4d4e925bf1ee.png)

Occasionally, the underline on some links may disappear entirely.

This issue will affect you if:
- your browser uses Chromium (for example, Chrome, Edge or Opera),
- you have links within a [multiple-column layout](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Columns)
- you have opted your service into the [GOV.UK Design System's new link styles](https://github.com/alphagov/govuk-frontend/pull/2183)

## How to fix this issue

- Choose not to opt in to the new link styles
- Remove the thicker underline from links in columns by turning off the hover state
- Change your layout so that it does not use a CSS multiple-column layout (albeit doing this could be very time-consuming)

We have [reported this issue to Chromium](https://bugs.chromium.org/p/chromium/issues/detail?id=1190987).

In the future, when Chromium fix the bug that's causing these issues, we'll turn on link styles by default.

36degrees:
It looks like this issue [may have been fixed in Chrome 102](https://bugs.chromium.org/p/chromium/issues/detail?id=1190987#c15) 🎉

36degrees:
Closing this issue as Chrome 102+ has reached 95%+ adoption.

PR #3599, opened 2023-05-10
Enable new link styles by default

### LINKED ISSUE #2350: Enable new link styles by default
At the minute, the new link styles introduced in 3.12.0 are disabled by default, and we require users to set a flag `$govuk-new-link-styles: true` before importing GOV.UK Frontend in order to opt-in to use them.

This is in case users:

- need to make changes to their service to introduce these new link styles consistently
- want to avoid opting in if they have links within a multiple-column layout in your CSS, due to #2204

In a future major release (probably after #2204 has been fixed in Chromium) we should change the setting to enable the links, making the new link styles opt-out.

We may then choose to deprecate the flag for removal in a following major release.

## What

Set `$govuk-new-link-styles` to `true` in `govuk/settings/_links.scss`.

Update any associated comments, documentation or tests.

## Why

We want to encourage as many services as possible to use the new link styles. We decided to make the link styles opt-in because of #2204.

Developers, tech writer

## Done when

- [ ] The new link styles are opt-in
- [ ] Comments, documentation and tests updated

36degrees:
In July 2021, we made [some improvements to new link styles][1] which were released as part of v3.12.0.

We introduced these changes behind a feature flag `$govuk-new-link-styles: true` which service teams needed to set before importing GOV.UK Frontend in order to opt-in.

This was in case users:

- needed to make changes to their service to introduce these new link styles consistently
- wanted to avoid opting in if they have links within a multiple-column layout in your CSS, due to Known issue: hover state is mis-painted on links within CSS multiple-column layouts #2204

Now that #2204 has been fixed in Chromium, we can make the new link styles ‘opt-out’.

[1]: https://github.com/alphagov/govuk-frontend/issues/2350

Closes #2350

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,17 @@ You can now style buttons on dark backgrounds to have a white background colour 
 
 This change was made in [pull request #3556: Add inverse button styles](https://github.com/alphagov/govuk-frontend/pull/3556).
 
+#### New link styles are now enabled by default
+
+In GOV.UK Frontend v3.12.0 we introduced new link styles which:
+
+- have underlines that are consistently thinner and a bit further away from the link text
+- have a clearer hover state, where the underline gets thicker to make the link stand out to users
+
+The new link styles are now enabled by default. If you are setting `$govuk-new-link-styles` to `true` in your Sass you can now remove this.
+
+This change was made in [pull request #3599: Enable new link styles by default](https://github.com/alphagov/govuk-frontend/pull/3599).
+
 ### Breaking changes
 
 #### Check that disabled buttons work as expected
--- a/packages/govuk-frontend-review/src/stylesheets/app.scss
+++ b/packages/govuk-frontend-review/src/stylesheets/app.scss
@@ -1,5 +1,4 @@
 $govuk-show-breakpoints: true;
-$govuk-new-link-styles: true;
 
 @import "govuk/all";
 @import "partials/app";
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,17 @@ You can now style buttons on dark backgrounds to have a white background colour
...
+
+In GOV.UK Frontend v3.12.0 we introduced new link styles which:
+
+- have underlines that are consistently thinner and a bit further away from the link text
+- have a clearer hover state, where the underline gets thicker to make the link stand out to users
+
+The new link styles are now enabled by default. If you are setting `$govuk-new-link-styles` to `true` in your Sass you can now remove this.
+
+This change was made in [pull request #3599: Enable new link styles by default](https://github.com/alphagov/govuk-frontend/pull/3599).
+
```
romaricpascal, line 23: **question** Did we say this setting would be removed in the next breaking change? If that's the case, do we want to add a little note that the setting is deprecated?

romaricpascal (approved):
Looks good for me, just thinking it's also an opportunity to mark deprecation and forget about the setting until the next breaking change if that's what we want 😊

CHANGELOG.md
36degrees: Would appreciate your thoughts on this in #3600! If we don't want to do #3600 then yes, we should probably deprecate it instead.

36degrees: Actually, I'll merge this and if we do decide to deprecate it we can do that in a follow up PR too.

PR #3600, opened 2023-05-10
Remove new link styles feature flag

36degrees:
This follows on from #3599 and removes the feature flag entirely. I think this makes sense as all of the CSS within the block guarded by the new link styles flag is also guarded by the other settings.

This means we can remove the flag as users can still disable the new link styles by setting:

```scss
$govuk-link-underline-offset: false;
$govuk-link-underline-thickness: false;
$govuk-link-hover-underline-thickness: false;
```

Would appreciate second opinions on whether this makes sense or whether we should do it 'properly' and deprecate it and remove it in v6 instead.


owenatgov (approved):
🔗

colinrotherham (approved):

Issue #2793, opened 2022-08-19
Remove `govuk-compatibility` mixin from codebase


owenatgov:
## What
Remove the mixin [`govuk-compatibility`](https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/tools/_compatibility.scss) and all its uses in govuk-frontend.
 
## Why
Required to remove compatibility mode completely from govuk-frontend in 5.0.

Part of https://github.com/alphagov/govuk-frontend/issues/2769

## Who needs to work on this
Developers

## Done when

- [x] Remove from button component
- [x] Remove from footer component
- [x] Remove from table component
- [x] Remove from section break style
- [x] Remove from links style
- [x] Remove setting completely

Issue #3583, opened 2023-05-04
Write up unified changelog entry for compatibility mode removal changes


owenatgov:
## What
Write up a unified changelog entry for the multiple changes in the v5 portion of the [compatibility mode removal work](https://github.com/alphagov/govuk-frontend/issues/2769).

## Why
To present the multiple removals around the same broad topic in a digestible way so that teams can better digest the changes.

## Who needs to work on this
Tech writer, developers

PR #3602, opened 2023-05-10
Remove `compatibility-mode` mixin

owenatgov:
## What/Why
Removes the public and private compatibility mode sass mixin (`compatibility-mode` and `_compatibility-mode` respectively), the private mixins use in the codebase and tests associated with the mixin.

Resolves https://github.com/alphagov/govuk-frontend/issues/2793


sassdoc.config.yaml
```diff
@@ -23,7 +23,6 @@ groups:
   settings/warnings: Settings / Warnings
   tools: Tools
   tools/assets: Tools / Assets
-  tools/compatibility-mode: Tools / Compatibility Mode
```
36degrees, line 26: Well done for spotting this – a thorough clean up

36degrees (approved):
Tidy! Assuming this will be covered by the changelog entry from #3583 and so doesn't need one now.

querkmachine (approved):
Was also just looking through this, so have a second approval for good luck.

Issue #2795, opened 2022-08-19
Remove compatibility mode variables


owenatgov:
## What
Remove the compatibility mode variables we use to provide support for legacy implementations of the govuk design system.
 
Details for why can be found in https://github.com/alphagov/govuk-frontend/issues/2769

## Who needs to work on this
Developers

## Done when

- [x] Remove `$_govuk-is-legacy-review-app`
- [x] Remove `$govuk-compatibility-govukelements`
- [x] Remove `$govuk-compatibility-govuktemplate`
- [x] Remove `$govuk-compatibility-govukfrontendtoolkit`
- [x] Check the repo for any other references to the legacy frameworks eg: in comments

PR #3608, opened 2023-05-11
Remove compatibility mode from review app

### LINKED ISSUE #2794: Remove legacy design system implementations from the review app
## What
Remove our legacy stylesheets which we use to mimic a service which uses legacy design system implementations.
 
## Why
We are removing compatibility mode in 5.0 and therefore it makes sense to also remove this from our own review app as they are no longer relevant.

## Who needs to work on this
Developers

## Done when

- [x] Decide which legacy examples we should keep and move over to govuk-frontend properly and which examples we should remove
- [x] Remove https://github.com/alphagov/govuk-frontend/blob/main/app/assets/scss/app-legacy-ie8.scss
- [x] Remove https://github.com/alphagov/govuk-frontend/blob/main/app/assets/scss/app-legacy.scss
- [x] Remove legacy implementations as npm dependancies

querkmachine:
Removes compatibility mode ("legacy") features, functionality and examples from the Frontend review app.

Closes #2794.


colinrotherham:
Thanks @querkmachine!

The build error is this line where we add `govuk_frontend_toolkit` to the Sass load paths:

https://github.com/alphagov/govuk-frontend/blob/b4ec518cc4ca23f651f6d6263b920066e0824c92/shared/tasks/styles.mjs#L83

The legacy packages are still in **package-lock.json** (bit odd) but npm can be a bit lazy during `npm install`

https://github.com/alphagov/govuk-frontend/blob/b4ec518cc4ca23f651f6d6263b920066e0824c92/package-lock.json#L84-L86

Bit annoying, but probably need to revert the change and run it again via the CLI instead:

```console
npm uninstall govuk_frontend_toolkit govuk_template_jinja govuk-elements-sass jquery --workspace govuk-frontend-review
```

colinrotherham:
> Also something I noticed. I've given that command a crack but they just don't seem to be getting removed from package-lock.json. Would deleting the lines manually be out of the question?

@querkmachine Try reinstalling them first then run the command? Worked for me locally

For a while now I've noticed `npm install` following a conflict or hand-crafted **package.json** edit misses things 😭

querkmachine:
> @querkmachine Try reinstalling them first then run the command? Worked for me locally

@colinrotherham Hrm, weird situation. The `package-lock.json` still includes an entry for `app`, as well as one for `packages/govuk-frontend-review`. The dependencies are being removed as expected from the latter but are sticking with the former. 

I'm not sure where the `app` entry is coming from. Might be a symlink thing?

colinrotherham:
@querkmachine Probably left over from an `npm install` after a previous **package-lock.json** conflict 🤦‍♂️

Not a major issue but we ideally want Dependabot to see that [jquery](https://www.npmjs.com/package/jquery) is finally gone! Try deleting **package-lock.json** entirely and running `npm install` to rebuild it from scratch?

You'll likely get some deep dependency patches creeping in but hopefully does the job

colinrotherham:
Think we should consider raising this as an npm bug 😬

To restore all those missing `optionalDependencies` for the Sass compiler:

```console
npm install sass-embedded --save-dev --workspace govuk-frontend
```

Just specifying one workspace is enough for them to be restored

packages/govuk-frontend-review/src/common/middleware/vendor.mjs
```diff
@@ -4,13 +4,9 @@ import { packageNameToPath } from 'govuk-frontend-lib/names'
...
 
 /**
- * Add middleware to serve legacy code
+ * Add middleware to serve dependencies
  * from node_modules
  */
-router.use('/govuk_template/', express.static(packageNameToPath('govuk_template_jinja', 'assets')))
-router.use('/govuk_frontend_toolkit/assets', express.static(packageNameToPath('govuk_frontend_toolkit', 'images')))
-router.use('/govuk_frontend_toolkit/', express.static(packageNameToPath('govuk_frontend_toolkit', 'javascripts/govuk')))
 router.use('/iframe-resizer/', express.static(packageNameToPath('iframe-resizer', 'js')))
```
colinrotherham, line 10: Should we treat `govuk-frontend` as "vendor" too?

We currently have this line in [middleware/assets.mjs](https://github.com/alphagov/govuk-frontend/blob/255263fa125198199ccf26eebf6be1e4ab14e7cd/packages/govuk-frontend-review/src/common/middleware/assets.mjs):

```js
router.use('/assets', express.static(packageNameToPath('govuk-frontend', 'src/govuk/assets')))
```

packages/govuk-frontend-review/src/views/macros/showExamples.njk
```diff
@@ -1,13 +1,13 @@
...
-{% macro showExamples(componentData, legacyQuery) %}
+{% macro showExamples(componentData) %}
 
 {% set componentSlug = componentData.name %}
 {% set componentName = componentData.name | replace("-", " ") %}
 
 {% for example in componentData.examples | rejectattr('hidden') %}
   {% set exampleSlug = (example.name | replace(" ", "-")) %}
-  {% set iframePathQuery = legacyQuery + '&iframe=true' if legacyQuery != '' else '?iframe=true' %}
+  {% set iframePathQuery = '?iframe=true' %}
```
colinrotherham, line 10: This variable is only used once, probably stick it straight into line 38?

```suggestion
```

colinrotherham (approved):
This looks ace 🙌 

Is there another issue for the `govuk-frontend` side of things?

Like removing various stray references to `govuk_elements`

https://github.com/alphagov/govuk-frontend/blob/2d85ce3e84fab73dd4dbd8042a8e1c2b6d4197fe/packages/govuk-frontend/src/govuk/settings/_warnings.scss#L11

https://github.com/alphagov/govuk-frontend/blob/2d85ce3e84fab73dd4dbd8042a8e1c2b6d4197fe/packages/govuk-frontend/src/govuk/components/radios/radios.yaml#L110

colinrotherham:
@querkmachine Nothing blocking, except maybe sorting out that **package-lock.json** before we merge

Fab work 😊

querkmachine:
> Is there another issue for the `govuk-frontend` side of things?

@colinrotherham We have some remaining code to strip out (#2795) but I don't think we have one specifically for random comments or documentation like those!

owenatgov:
Hopping in re:

>@colinrotherham We have some remaining code to strip out (https://github.com/alphagov/govuk-frontend/issues/2795) but I don't think we have one specifically for random comments or documentation like those!

I think I assumed that would be part of https://github.com/alphagov/govuk-frontend/issues/2795 but it's not explicitly noted. I'll add a todo to the issue.

colinrotherham:
Brill, thanks

Probably just checking for mentions of all the packages now removed:

>govuk_frontend_toolkit
>govuk_template_jinja
>govuk-elements-sass
>jquery

packages/govuk-frontend-review/src/common/middleware/vendor.mjs
querkmachine: I can't say I have a strong preference, but it probably makes sense given we're now consuming Frontend as an essentially separate package.

querkmachine: Well, that might be a reason. Seems that after moving the line between files, certain assets like the Transport font start 404-ing. Gonna keep them in-place for now, rather than dig into that.

colinrotherham:
@querkmachine I've fixed the **package-lock.json** issues in this PR:

* https://github.com/alphagov/govuk-frontend/pull/3619

Probably want to get that fixed separately versus the full rebuild in this PR:

>3,615 additions, 5,869 deletions not shown because the diff is too large. Please use a local Git client to view these changes.

colinrotherham (approved):
🙌 Approving again

PR #3614, opened 2023-05-11
Separate JS and CSS diff, and upload diffs as artifacts

romaricpascal:
Adds separate diffs for the JS and CSS when diffing `dist`, to prepare from extracting them completely from the main diff in order to limit the chances of the diff being too large to be posted as a comment on a PR (there's a limit of 65536 characters).

The resulting diffs are [uploaded as workflow artifacts](https://github.com/alphagov/govuk-frontend/issues/3229) to provide a fallback access in case the diff are too big for comment on the PR.

The PR extracts the diffing into a shell script, allowing it to be run locally as well.

Once #3571 is merged, we'll be able to post separate comments to the PR for each of the diffs, with links to the workflow artifacts when the diffs are too big.

Closes https://github.com/alphagov/govuk-frontend/issues/3229

> **Note**
>
> The diff comment was generated thanks to manual test changes to the `dist` files in [95c4e3f](https://github.com/alphagov/govuk-frontend/pull/3614/commits/95c4e3fa7f691fcb10a917cbbe8cdaa904a5147f), which has since been removed as we don't want to merge that on main.

### DIFF
--- a/.github/workflows/diff-change-to-dist.yaml
+++ b/.github/workflows/diff-change-to-dist.yaml
@@ -33,8 +33,9 @@ jobs:
       - name: Generate diff
         id: diff
         run: |
-          git diff -M05 origin/$GITHUB_BASE_REF -- dist \
-            > $GITHUB_WORKSPACE/dist.diff
+          # Using `origin/$GITHUB_BASE_REF` to avoid actually checking out the branch
+          # as all we need is to let Git diff the two references
+          bin/dist-diff.sh origin/$GITHUB_BASE_REF $GITHUB_WORKSPACE
 
       - name: Add comment to PR
         uses: actions/github-script@v6
--- a/.github/workflows/diff-change-to-dist.yaml
+++ b/.github/workflows/diff-change-to-dist.yaml
@@ -37,6 +37,13 @@ jobs:
           # as all we need is to let Git diff the two references
           bin/dist-diff.sh origin/$GITHUB_BASE_REF $GITHUB_WORKSPACE
 
+      - name: Save distribution diffs
+        uses: actions/upload-artifact@v3
+        with:
+          name: Dist diff
+          path: dist*.diff
+          if-no-files-found: ignore
+
       - name: Add comment to PR
         uses: actions/github-script@v6
         with:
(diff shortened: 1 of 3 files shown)


.github/workflows/diff-change-to-dist.yaml
```diff
@@ -33,8 +33,16 @@ jobs:
       - name: Generate diff
         id: diff
         run: |
-          git diff -M05 origin/$GITHUB_BASE_REF -- dist \
-            > $GITHUB_WORKSPACE/dist.diff
+          # Using `origin/$GITHUB_BASE_REF` to avoid actually checking out the branch
+          # as all we need is to let Git diff the two references
+          .github/workflows/scripts/dist-diff.sh origin/$GITHUB_BASE_REF $GITHUB_WORKSPACE
```
colinrotherham, line 38: Should we consolidate all scripts into [`./bin`](https://github.com/alphagov/govuk-frontend/tree/main/bin)?

Thinking ahead to when we do releases from GitHub Actions

.github/workflows/diff-change-to-dist.yaml
```diff
@@ -33,8 +33,16 @@ jobs:
...
+
+      - name: Save diffs as artifacts
```
colinrotherham, line 40: Probably don't need to mention that we're saving an artifact

Might use a different action in future

```
Verify distribution build
Save test coverage
Run test task
```

```suggestion
      - name: Save distribution diff
```

.github/workflows/scripts/dist-diff.sh
```diff
@@ -0,0 +1,31 @@
...
+base="${1:-main}"
+# And the output folder to the current working directory
+output_folder="${2:-`pwd`}"
+
+# Diff the minified JS file
+git diff \
+  $base:dist/govuk-frontend-$(git show $base:dist/VERSION.txt).min.js \
+  dist/govuk-frontend-$(cat dist/VERSION.txt).min.js \
+  > $output_folder/dist-js.diff
+# Diff the minified CSS file
```
colinrotherham, line 23: Mind if we keep the line breaks above comments (as above?)

Makes it a bit more readable

```suggestion

# Diff the minified CSS file
```

bin/dist-diff.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/sh
+set -e
```
romaricpascal, line 2: Little heads up that, I've udpdated this part to match the rest of the scripts in `bin`.

colinrotherham (approved):

PR #3621, opened 2023-05-12
Fix ESLint and Stylelint via `gulp.watch()`

colinrotherham:
This PR fixes a regression caused by https://github.com/alphagov/govuk-frontend/commit/cc0b77af1ea3cdd48fe75de6c49d532a3eed39e0

Our Gulp watch tasks now lint code from their respective npm workspace:

```console
npm run lint:scss --workspace govuk-frontend
npm run lint:js --workspace govuk-frontend
```

Sounds great, but the npm scripts above only exist at project level so fail with:

```console
npm ERR! Lifecycle script `lint:scss` failed with error:
npm ERR! Error: Missing script: "lint:scss"
```

We have two choices:

1) Duplicate `lint:scss` and `lint:js` into each workspace
2) Lint from project level (again) but target only the workspace

I've picked 2) for this fix


packages/govuk-frontend-review/nodemon.json
```diff
@@ -1,5 +1,5 @@
 {
-  "watch": ["./src", "../packages/govuk-frontend/src/govuk/**/*.{json,yaml}"],
+  "watch": ["./src", "../govuk-frontend/src/govuk/**/*.{json,yaml}"],
```
colinrotherham, line 2: This path was missed from when the Review app was moved from `./app`

packages/govuk-frontend/tasks/watch.mjs
```diff
@@ -18,7 +20,7 @@ export const watch = (options) => gulp.parallel(
     gulp.watch([
       `${slash(options.srcPath)}/govuk/**/*.scss`,
       `!${slash(options.srcPath)}/govuk/vendor/*`
-    ], npm.script('lint:scss', options))
+    ], npm.script('lint:scss:cli', [join(options.workspace, '**/*.scss')]))
```
romaricpascal, line 23: **question** Does this need `slash`ing (or use of `node:path/posix`) for Windows environments, as `join` will use the OS separator I believe?

colinrotherham: The arguments get passed straight into [`@npmcli/run-script`](https://www.npmjs.com/package/@npmcli/run-script) so I suppose the question is:

**Will ESLint and Stylelint accept OS-native slashes via CLI?**
I've just assumed they would do and not done any conversion

Have you got a Windows machine to hand??

romaricpascal: No Windows machine at hand unfortunately (would need to install a few things on the modern.ie VMs, which I haven't done). I was asking because we `slash` most of the other glob we pass around (like L21-22 right above) so figured we may need to do so there as well. 

For safety, would you mind adding a `slash` around it, please?

colinrotherham: You'd think CLI tools on Windows would rather accept Windows paths than POSIX ones?

I've got a Windows 10 partition so I'll boot into it tomorrow and check

colinrotherham: ✅ ESLint supports Windows paths on Windows
🚫 Stylelint does not support Windows paths on Windows

I've pushed up a [commit with `slash()`](https://github.com/alphagov/govuk-frontend/pull/3621/commits/6ba66c96e6e86a35c957cd996b4e2e6c0bd47fd1) 🤦‍♂️

colinrotherham:
✅ Ready for review again @romaricpascal

querkmachine (approved):
Looks good to me

shared/tasks/npm.mjs
```diff
@@ -9,14 +9,16 @@ import { task } from './index.mjs'
  * Spawns Node.js child process for npm script
  *
  * @param {string} name - npm script name
+ * @param {string[]} [args] - npm script CLI arguments
  * @param {import('./index.mjs').TaskOptions} [options] - Task options
  * @returns {Promise<void>} Script run
  */
-export async function run (name, options) {
+export async function run (name, args = [], options) {
```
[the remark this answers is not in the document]
colinrotherham: Thanks @querkmachine, appreciate it

Yeah was a bit awkward either way, sadly

I settled on this order as most of our tasks are at project level:

```js
// Run task at project level
await npm.run('task')

// Run task at project level with arguments
await npm.run('task', ['**/*.njk'])

// Run task at workspace level with arguments
await npm.run('task', ['**/*.njk'], { basePath: 'packages/govuk-frontend' })
```

I could have made the 2nd param as `argsOrOptions` but would have needed [`@overload`](https://typedoc.org/tags/overload/) faff

PR #3623, opened 2023-05-12
Compile GOV.UK Prototype Kit config to package top level

colinrotherham:
This PR moves our GOV.UK Frontend Prototype Kit config for backwards compatibility

1. Ensures config is exported from `govuk-frontend/govuk-prototype-kit.config.json`
2. Ensures config macros, assets and Nunjucks paths add `dist/` prefix

This was previously noted under [**Restructure govuk-frontend repository for v5** Implications](https://github.com/alphagov/govuk-design-system-architecture/blob/main/decision-records/005-repository-organisation-for-v5.md#implications)

>* We need to keep making sure our configuration for the GOV.UK Prototype Kit gets recognised by the Kit and picks up our files as necessary.

### DIFF
--- a/packages/govuk-frontend/package.json
+++ b/packages/govuk-frontend/package.json
@@ -6,6 +6,7 @@
   "files": [
     "src",
     "dist",
+    "govuk-prototype-kit.config.json",
     "package.json",
     "README.md"
   ],
--- a/bin/pre-release.sh
+++ b/bin/pre-release.sh
@@ -26,6 +26,7 @@ npm run build:package
 if [[ -n $(git status --porcelain) ]]; then
   echo "✍️ Commiting changed package"
   git add packages/govuk-frontend/dist/
+  git add --force packages/govuk-frontend/govuk-prototype-kit.config.json
 
   git commit -m "Release GOV.UK Frontend to '$BRANCH_NAME' for testing"
 
--- a/packages/govuk-frontend/.gitignore
+++ b/packages/govuk-frontend/.gitignore
@@ -1,2 +1,4 @@
+govuk-prototype-kit.config.json
+
 # Build output (committed)
 !/dist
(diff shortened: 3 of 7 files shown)


romaricpascal:
Things worked fined with `npm link` locally. However, the Prototype Kit config doesn't seem to be pushed to pre-release branches 😔 I've created a pre-release for this PR there: https://github.com/alphagov/govuk-frontend/tree/pre-release-move-config-prototype-kit 😊

colinrotherham:
> Things worked fined with `npm link` locally. However, the Prototype Kit config doesn't seem to be pushed to pre-release branches 😔 I've created a pre-release for this PR there: https://github.com/alphagov/govuk-frontend/tree/pre-release-move-config-prototype-kit 😊

Sadly we do two things which only https://github.com/alphagov/govuk-frontend/pull/3498 can unblock

1. Pre-releases check for unstaged changes
```sh
if [[ -n $(git status --porcelain) ]]; then
```

2. Pre-releases exclude anything outside `packages/govuk-frontend/dist`
```console
git filter-branch --force --subdirectory-filter packages/govuk-frontend
```

I've rebased https://github.com/alphagov/govuk-frontend/pull/3498 with this branch so let's test there?

romaricpascal (approved):
Tests with #3498 were OK, so let's merge them both one after the other (rather than going through some tedious rebasing).

packages/govuk-frontend/package.json
```diff
@@ -6,6 +6,7 @@
   "files": [
     "src",
     "dist",
+    "govuk-prototype-kit.config.json",
```
colinrotherham, line 9: This ensures GOV.UK Prototype Kit can find the config

But paths at their end are still hard coded: https://github.com/alphagov/govuk-prototype-kit/commit/953830afb1f2da8bdfac0a733f3a578a8fb1bbac

PR #3637, opened 2023-05-15
Increase Dependabot open PR limit

domoscargin:
We've decided to bump Dependabot's max open PR limit back to 10 and see how it goes, since we were falling a bit behind with the default limit of 5.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,6 +4,7 @@ updates:
   # Update npm packages
   - package-ecosystem: npm
     directory: /
+    open-pull-requests-limit: 10
     reviewers:
       - alphagov/design-system-developers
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -13,6 +13,7 @@ updates:
       # Setting a timezone so we let dependabot worry about BST
       timezone: 'Europe/London'
     versioning-strategy: increase
+    open-pull-requests-limit: 10
```
colinrotherham, line 16: Super minor thing (not blocking) is we've ordered these keys A-Z

(Except for `package-ecosystem` which is used as a title-ish)

colinrotherham (approved):

Issue #2769, opened 2022-08-15
Remove compatibility mode


owenatgov:
## What we'll do

We'll remove [compatibility mode](https://frontend.design-system.service.gov.uk/compatibility-mode/) from GOV.UK Frontend.

Compatibility mode exists to make it easier for users to [migrate from our old frameworks](https://frontend.design-system.service.gov.uk/migrating-from-legacy-products/#migrate-from-our-old-frameworks) (GOV.UK Elements, GOV.UK Frontend Toolkit and GOV.UK Template).

When compatibility mode is enabled, GOV.UK Frontend:

- uses the old colour palette
- uses the old GOV.UK font from GOV.UK Template
- overrides some of the CSS in the legacy frameworks
- no longer uses rem for font sizes

## Why we're doing it

We're [making changes to the typography scale](https://github.com/alphagov/govuk-design-system/issues/2289) used in GOV.UK Frontend.

Without significant effort from us (to support two different typography scales), users who have enabled compatibility mode would see a mix of different sized text in their services.

Given that the Design System is now a live service and it's been nearly 5 years since the last release of GOV.UK Elements, we think it's now time to retire compatibility mode.

## How this impacts our users

Teams that are still using our old frameworks will not be able to migrate directly to v5.0 or above. However, they will still  have a migration path – they'll just need to finish migrating to the latest version of GOV.UK Frontend v4 (which will still support compatibility mode) and then upgrade to v5 in a separate step.

## Epic lead
Owen

## Who needs to work on this
Developers

## Done when

In v4.x we will:

```[tasklist]
- [x] Investigate where compatibility mode is being used in govuk-frontend
- [x] Assess if there are any functions that compatibility mode facilitates that we want to keep
- [x] https://github.com/alphagov/govuk-frontend/issues/2787
- [x] https://github.com/alphagov/govuk-frontend/issues/2890
- [x] https://github.com/alphagov/govuk-frontend/issues/2788
- [ ] https://github.com/alphagov/govuk-frontend/issues/2789
- [ ] https://github.com/alphagov/govuk-frontend-docs/issues/207
```

In v5.0 we will:

```[tasklist]
- [ ] https://github.com/alphagov/govuk-frontend/issues/2790
- [ ] https://github.com/alphagov/govuk-frontend/issues/2791
- [ ] https://github.com/alphagov/govuk-frontend/issues/2792
- [ ] https://github.com/alphagov/govuk-frontend/issues/2793
[...]

PR #3643, opened 2023-05-17
Create a unified changelog entry for compatibility mode

owenatgov:
Creates a single changelog entry to cover all the 5.0 code removal tasks in https://github.com/alphagov/govuk-frontend/issues/2769

Closes https://github.com/alphagov/govuk-frontend/issues/3583

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,30 @@ This change was made in [pull request #3599: Enable new link styles by default](
 
 ### Breaking changes
 
+#### "Compatibility mode" features are no longer supported
+
+GOV.UK Frontend no longer supports compatibility with [our old frameworks](https://frontend.design-system.service.gov.uk/migrating-from-legacy-products/#migrate-from-our-old-frameworks): GOV.UK Elements, GOV.UK Template and GOV.UK Frontend Toolkit. You can no longer incrementally add GOV.UK Frontend to an existing service using one of these previous frameworks. We have removed the following Sass variables and one mixin which controlled compatibility mode:
+
+- `$govuk-compatibility-govukelements`
+- `$govuk-compatibility-govuktemplate`
+- `$govuk-compatibility-govukfrontendtoolkit`
+- The `compatibility-mode` mixin which automatically checked if any of the 3 control variables were set to `true`
+
+We have additionally removed features that were managed using the above variables. The following features and their corresponding variables have now been removed:
+
+- Access to the legacy colour palette using `$govuk-use-legacy-palette`
+- Access to the legacy font using `$govuk-use-legacy-font`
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -23,6 +23,33 @@ This change was made in [pull request #3599: Enable new link styles by default](
...
+#### "Compatibility mode" features are no longer supported
+
+GOV.UK Frontend no longer supports compatibility with legacy frameworks. You can no longer incrementally add GOV.UK Frontend to an existing service using a previous GOV.UK framework. We have removed the following Sass variables and one mixin which controlled compatibility mode:
+
+- `$govuk-compatibility-govukelements`
+- `$govuk-compatibility-govuktemplate`
+- `$govuk-compatibility-govukfrontendtoolkit`
+- The `compatibility-mode` mixin which automatically checked if any of the 3 control variables were set to `true`
+
+We have additionally removed features that were managed via the above variables. The following features and their corresponding variables have now been removed:
```
claireashworth, line 35:
```suggestion
We have additionally removed features that were managed using the above variables. The following features and their corresponding variables have now been removed:
```

claireashworth: not using 'via' is a style/accessibility thing - however I see no problem in keeping it, if using it provides the most accurate sentence.

CHANGELOG.md
```diff
@@ -23,6 +23,33 @@ This change was made in [pull request #3599: Enable new link styles by default](
 
 ### Breaking changes
 
+#### "Compatibility mode" features are no longer supported
+
+GOV.UK Frontend no longer supports compatibility with legacy frameworks. You can no longer incrementally add GOV.UK Frontend to an existing service using a previous GOV.UK framework. We have removed the following Sass variables and one mixin which controlled compatibility mode:
```
colinrotherham, line 28: What's a legacy framework? 🤷‍♂️

It's been so long since we told people to migrate, should we spell it out and link here?
https://frontend.design-system.service.gov.uk/migrating-from-legacy-products/

```suggestion
GOV.UK Frontend no longer supports compatibility with our old frameworks, GOV.UK Elements, GOV.UK Template and GOV.UK Frontend Toolkit. You can no longer incrementally add GOV.UK Frontend to an existing service using a previous GOV.UK framework. We have removed the following Sass variables and one mixin which controlled compatibility mode:
```

colinrotherham: We fondly call them "our old frameworks" rather than _legacy_

owenatgov: I was so sure we called them "legacy frameworks" 🤔 I'll amend. I think it's a good idea to link to that page but I'm not sure how the docs split work is going to impact the freshness of that URL. We should be cognisant of this and maybe return to this changelog item later.

claireashworth (approved):

PR #3644, opened 2023-05-17
[SPIKE] Modifications to pagination component

### LINKED ISSUE #3324: Pagination component numbering issue on small screens
## Description of the issue

For smaller screen sizes the pagination numbering stops obeying the rules as defined [here](https://design-system.service.gov.uk/components/pagination/#:~:text=Use%20ellipses%20(%E2%80%A6)%20to%20replace%20any%20skipped%20pages.%20For%20example%3A).

CleanShot 2023-02-22 at 09 44 50@2x

This issue can be reproduced by going to https://www.staging.tax.service.gov.uk/auth-login-stub/gg-sign-in?continue=/manage-transit-movements/draft-declarations using enrolment:
* Enrolment Key: HMRC-CTC-ORG
* Identifier Name: EORINumber
* Identifier Value: 1234567

For a smaller screen size you will see the behaviour as demonstrated in the above screenshot.

In the above screenshot I would expect to see
* 1 2 [3] 4 ... 7 for a normal screen size
* 1 ... [3] ... 7 for a smaller screen size

- Operating system: All
- Browser: All
- Browser version: All
- GOV.UK Frontend Version: play-frontend-hmrc 6.5.0-play-28

## Additional information

See the following thread in Slack for more info: https://hmrcdigital.slack.com/archives/CJMUM9AG3/p1676914677343619

querkmachine:
A little bit of a self-indulgent spike refactoring how the pagination component works, in an effort to resolve #3324 and (potentially) make it a little simpler for teams to use. 

Component page:
https://govuk-frontend-pr-3644.herokuapp.com/components/pagination

Temporary example page showing various pagination lengths in various states: 
https://govuk-frontend-pr-3644.herokuapp.com/examples/pagination

## Background

When investigating #3324 I came up with a couple of potential fixes, but ultimately it seemed (to me, at least) that the root of the problem is that the component lacks context about what information it’s handling. It currently only receives a limited set of explicitly categorised information and is told to format and spit it out again without interrogating it; so when it comes to processing the data more thoroughly (e.g. deciding which things should be hidden on mobile) it can only work off assumptions.

One solution is to allow service teams to provide that context by manually flagging what things are and in what contexts they should be visible, but this creates a burden on the implementor to get all the nuances of it right—if the service has multiple paginations, they'd need to get it right multiple times—and none of it is particularly easy to explain in documentation.

This is my first pass at a [“do the hard work to make it simple”](https://www.gov.uk/guidance/government-design-principles#do-the-hard-work-to-make-it-simple) approach that brings more of the logic into the Nunjucks.

Instead of a backend (or other code) needing to decide ahead of time which pages to render in the pagination, what labels they use, where and when ellipses should appear, etc., this rewrite now only requires implementors to provide a list of (ordered) URLs and what the current page is: the component handles everything else. 

### ...to how the component is used

- Users now list all pagination items within the `items` array, rather than just the subset that they wish to show on the page. (This may reduce complexity in backend code, as the same pagination code can be used in multiple pages and contexts?)
- The `item.ellipsis` parameter has been removed. Ellipsis placement and visibility is now determined automatically. 
- The `item.current` parameter has been replaced by a `currentPage` parameter.
- The `item.number` parameter is no longer required. Numbers are automatically generated where not defined. 
- A new `neighbouringPages` parameter has been added, allowing macro users to configure how many pages 'neighbouring' the current page to show. 

### ...to how the component works

- Rather than expecting only the subset of pages the user wants to show, the pagination component now expects _all_ possible pages to be provided. 
- Which pages are to be displayed (`visiblePagesLower` and `visiblePagesUpper`) are determined using a combination of `currentPage` and `neighbouringPages` parameters.
- Every item in the `items` array is now looped through and has various tests ran against it to determine whether it should be output or not.
  - The first and last items are always output.
  - If the first item and `currentPage` are more than 1 index apart, ellipses are appended after the first item.
    - If the _second_ item and `visiblePagesLower` are 1 or more indexes apart, the ellipses will only be shown on mobile. 
  - If the last item and `currentPage` are more than 1 index apart, ellipses are prepended before the last item. 
    - If the _penultimate_ item and `visiblePagesUpper` are 1 or more indexes apart, the ellipses will only be shown on mobile. 
  - If an item is within the range of `visiblePagesLower` and `visiblePagesUpper`, but is not the `currentPage`, it is rendered but will be hidden on mobile.
  - If an item's index matches `currentPage`, it is rendered with a 'current' class. 
  - Items that do not match any of the above conditions are not rendered. 
- As the code to render pagination items is now needed in multiple places, I've refactored it into a private macro within the template.
- If there are three or fewer items in the pagination, a special case is activated where all items are output on all viewports. This prevents the pagination being unnecessarily truncated on narrow viewports (it'll render `1 | 2 | 3` instead of `1 | … | 3`). 

## Thoughts

This is a bit of a compromise between the current method (implementors provide an array of page, with links, optional numbering and other attribute info) and a more fully-automated approach (an implementor doesn't even provide an array and simply say there are 40 pages of results).

Whilst the latter would be even simpler to use, it has the trade-off of removing or restricting a number of features that the current pagination component is capable of, such as being able to customise the labelling of items or use non-incremental URL schemes for paging. 

## Todo

- No tests have been written or updated for the new/altered functionality.
- The full gamut of pagination lengths, `currentPage` selections and `neighbouringPages` combinations hasn't been tested. There may still be edge cases. 
- I've left the previous/next link functionality completely untouched, but we could potentially automate aspects of them given we now consume all possible pages as a set and know our position within the set.
- Currently it's necessary for teams to provide overrides for `item.visuallyHiddenText` and `item.attributes` separately for each item. Defining a way to set these globally for all items would also seem like a nice enhancement.

### DIFF
--- a/packages/govuk-frontend-review/src/views/examples/pagination/index.njk
+++ b/packages/govuk-frontend-review/src/views/examples/pagination/index.njk
@@ -1,0 +1,92 @@
+{% from "back-link/macro.njk" import govukBackLink %}
+{% from "pagination/macro.njk" import govukPagination %}
+
+{% extends "layout.njk" %}
+
+{% block beforeContent %}
+  {{ govukBackLink({
+    href: "/"
+  }) }}
+{% endblock %}
+
+{% set veryShortPaginationItems = [
+  { href: '#' },
+  { href: '#' },
+  { href: '#' }
+] %}
+
+{% set shortPaginationItems = [
+  { href: '#' },
+  { href: '#' },
+  { href: '#' },
+  { href: '#' }
+] %}
+
+{% set mediumPaginationItems = [
+  { href: '#' },
+  { href: '#' },
+  { href: '#' },
+  { href: '#' },
+  { href: '#' },
+  { href: '#' },
+  { href: '#' }
+] %}
+
+{% set longPaginationItems = [
+  { href: '#' },
+  { href: '#' },
+  { href: '#' },
+  { href: '#' },
+  { href: '#' },
+  { href: '#' },
+  { href: '#' },
+  { href: '#' },
+  { href: '#' },
+  { href: '#' },
+  { href: '#' },
+  { href: '#' }
+] %}
+
+{% block content %}
+
+<h2 class="govuk-heading-l">Very short</h2>
+<p class="govuk-body">Very short paginations (3 or fewer items) are small enough to not need to hide anything on any breakpoint.</p>
+
+{%- for i in range(1, (veryShortPaginationItems | length) + 1) %}
+  {{ govukPagination({
+    items: veryShortPaginationItems,
+    currentPage: i
+  }) }}
+{%- endfor %}
+
+<h2 class="govuk-heading-l">Short</h2>
+<p class="govuk-body">Short paginations (4 items, with default settings) can show all items on wide breakpoints, but hide some for narrow breakpoints.</p>
+
+{%- for i in range(1, (shortPaginationItems | length) + 1) %}
+  {{ govukPagination({
+    items: shortPaginationItems,
+    currentPage: i
+  }) }}
+{%- endfor %}
+
+<h2 class="govuk-heading-l">Medium</h2>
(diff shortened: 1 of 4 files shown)


querkmachine:
Whilst I still think this is intriguing, there doesn't seem to be much appetite internally or externally to change how pagination currently works. I'm going to close this down for neatness sake.

PR #3645, opened 2023-05-18
Ensure component data types are correct

colinrotherham:
Just a few "component data" type fixes that showed up recently

### DIFF
--- a/shared/lib/files.js
+++ b/shared/lib/files.js
@@ -75,10 +75,12 @@ const mapPathTo = (patterns, callback) => (entryPath) => {
  * Load single component data
  *
  * @param {string} componentName - Component name
- * @returns {Promise<ComponentData>} Component data
+ * @returns {Promise<ComponentData & { name: string }>} Component data
  */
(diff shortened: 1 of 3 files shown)


shared/lib/files.js
```diff
@@ -161,7 +163,6 @@ module.exports = {
  * Component data from YAML
  *
  * @typedef {object} ComponentData
- * @property {string} name - Component name
```
colinrotherham, line 164: Component data YAML files don't have a `name` property

shared/lib/files.js
```diff
@@ -75,10 +75,12 @@ const mapPathTo = (patterns, callback) => (entryPath) => {
  * Load single component data
  *
  * @param {string} componentName - Component name
- * @returns {Promise<ComponentData>} Component data
+ * @returns {Promise<ComponentData & { name: string }>} Component data
```
colinrotherham, line 78: Component data from `getComponentData()` has a `name` property added

romaricpascal (approved):

PR #3646, opened 2023-05-18
Add warning functions

### LINKED ISSUE #3427: `_warning` mixin cannot be used within functions

querkmachine:
Splits the `_warning` mixin functionality into separate functions that can be used in contexts where mixins can't. Resolves #3427. 

## Changes

- Splits `_warning` mixin into separate `_should-warn` and `_warning-message` functions per @owenatgov's [comment on the issue](https://github.com/alphagov/govuk-frontend/issues/3427#issuecomment-1536484338). 
- Updates the contribution docs on managing change to mention the new functions, when to use them, and provide an example. 
- Updates the `govuk-colour` function to use the new functions. Previously this had to partially re-implement the mixin. 

## Thoughts

- Unlike the mixin, the functions do nothing to suppress future instances of the same warning type. Although technically possible, the [Sass documentation for functions](https://sass-lang.com/documentation/at-rules/function) actively discourages using functions to manipulate global variables, and there are situations where this lack of suppression might be useful.


docs/contributing/managing-change.md
```diff
@@ -55,9 +55,22 @@ In the annotation description include:
 - the suggested alternative, if there is one
 - a link to the GitHub issue for its removal
 
-If possible, update the mixin or function to output a warning using the `_warning` mixin (see section below on allowing users to suppress warnings).
+If possible, update the code to output a warning using the `_warning` mixin or functions (see section below on allowing users to suppress warnings).
```
owenatgov, line 58: A small personal content opinion.

```suggestion
If possible, update the code to output a warning using the`_warning` mixin or its associated functions (see section below on allowing users to suppress warnings).
```

owenatgov (approved):
One nitpick. Otherwise this looks great and deserves an approval anyway 👍🏻

PR #3647, opened 2023-05-18
Ensure shared task `options` are mandatory unless specified

colinrotherham:
This PR ensures all build task options are mandatory and type checked

I've also included some related changes:

1. Configure the Review app to use `govuk-frontend` JSDoc types again
2. Remove unnecessary options (like renaming `*.scss` to `*.scss`)
3. Adding comments to explain why files are renamed or moved

This work was split out whilst adding **postcss.config.mjs** and **rollup.config.mjs** files to each workspace

### Before
This will crash as no `destPath` has been provided

```js
import { scripts } from 'govuk-frontend-tasks'

// Compile GOV.UK Frontend JavaScript
scripts.compile('!(*.test).mjs', {
  destPath: '/custom/build/path'
})
```

### After
This works by extending the build options using `...options`

```js
import { scripts } from 'govuk-frontend-tasks'

// Default build options
import { options } from './build/options.mjs'

// Compile GOV.UK Frontend JavaScript
scripts.compile('!(*.test).mjs', {
  ...options,
  destPath: '/custom/build/path'
})
```

### DIFF
--- a/packages/govuk-frontend/tasks/build/release.mjs
+++ b/packages/govuk-frontend/tasks/build/release.mjs
@@ -12,9 +12,7 @@ import gulp from 'gulp'
  */
 export default (options) => gulp.series(
   task.name('clean', () =>
-    files.clean('*', {
-      destPath: options.destPath
-    })
+    files.clean('*', options)
   ),
 
   // Copy GOV.UK Frontend static assets
@@ -28,8 +26,9 @@ export default (options) => gulp.series(
   // Compile GOV.UK Frontend JavaScript
   task.name('compile:js', () =>
     scripts.compile('all.mjs', {
+      ...options,
+
       srcPath: join(options.srcPath, 'govuk'),
-      destPath: options.destPath,
 
       filePath (file) {
         return join(file.dir, `${file.name.replace(/^all/, pkg.name)}-${pkg.version}.min.js`)
@@ -40,8 +39,9 @@ export default (options) => gulp.series(
   // Compile GOV.UK Frontend Sass
   task.name('compile:scss', () =>
     styles.compile('**/[!_]*.scss', {
+      ...options,
+
       srcPath: join(options.srcPath, 'govuk'),
-      destPath: options.destPath,
 
       filePath (file) {
         return join(file.dir, `${file.name.replace(/^all/, pkg.name)}-${pkg.version}.min.css`)
@@ -51,8 +51,6 @@ export default (options) => gulp.series(
 
   // Update GOV.UK Frontend version
   task.name("file:version 'VERSION.txt'", () =>
-    files.version('VERSION.txt', {
-      destPath: options.destPath
-    })
+    files.version('VERSION.txt', options)
   )
 )
--- a/packages/govuk-frontend/tasks/build/release.mjs
+++ b/packages/govuk-frontend/tasks/build/release.mjs
@@ -30,8 +30,9 @@ export default (options) => gulp.series(
 
       srcPath: join(options.srcPath, 'govuk'),
 
-      filePath (file) {
-        return join(file.dir, `${file.name.replace(/^all/, pkg.name)}-${pkg.version}.min.js`)
+      // Rename using package name (versioned) and `*.min.js` extension
+      filePath ({ dir, name }) {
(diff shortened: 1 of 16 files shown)


romaricpascal:
Tidying up looks good, cheers! Can you explain the benefits of having `task.options` over doing the spread ourselves?

I'm expecting that with the types not being optional anymore, we'd catch situations like the one you mentioned in the first post. Would TypeScript warn OK on your first example (for lack of the necessary options)?

```js
import { scripts } from 'govuk-frontend-tasks'

// Compile GOV.UK Frontend JavaScript
scripts.compile('!(*.test).mjs', {
  destPath: '/custom/build/path'
})
```

If that's the case, I'd like to understand why we need a custom function for the merging. If it's only for freezing objects, I'm not sure it's that risky to let the objects be extended (especially as we'll be merging in the paths anyways so have a separate object per task call), but open to be convinced otherwise 😊

romaricpascal:
As we get type checking in any case, I'd reather we have less code and just use spread to start, cheers 😊

colinrotherham:
Swapped to using object spread, ready for review again

packages/govuk-frontend/tasks/build/release.mjs
```diff
@@ -28,31 +26,33 @@ export default (options) => gulp.series(
   // Compile GOV.UK Frontend JavaScript
   task.name('compile:js', () =>
     scripts.compile('all.mjs', {
+      ...options,
+
       srcPath: join(options.srcPath, 'govuk'),
-      destPath: options.destPath,
```
colinrotherham, line 32: Don't need to repeat ourselves. The destination is already in `options`

shared/tasks/assets.mjs
```diff
@@ -84,7 +86,6 @@ export async function write (filePath, result) {
  * @typedef {object} AssetFileOptions
  * @property {(file: import('path').ParsedPath) => string} [filePath] - File path formatter
  * @property {(contents?: string) => Promise<string>} [fileContents] - File contents formatter
- * @property {string[]} [ignore] - File path patterns to ignore
```
colinrotherham, line 87: We had to carefully ignore the published **package.json** and **README.md** in the past

We haven't used this option since merging:

* https://github.com/alphagov/govuk-frontend/pull/3498

shared/tasks/components.mjs
```diff
@@ -11,7 +11,7 @@ import { files } from './index.mjs'
  * Generate fixtures.json from component data
  *
  * @param {AssetEntry[0]} pattern - Path to ${componentName}.yaml
- * @param {AssetEntry[1]} options - Asset options
+ * @param {Pick<AssetEntry[1], "srcPath" | "destPath">} options - Asset options
```
colinrotherham, line 14: This task already provides its own `filePath` and `fileContents`

Here we're narrowing down the types to _warn_ when passing in options that aren't picked

shared/tasks/components.mjs
```diff
@@ -44,7 +43,7 @@ export async function generateFixtures (pattern, { srcPath, destPath }) {
  * Generate macro-options.json from component data
  *
  * @param {AssetEntry[0]} pattern - Path to ${componentName}.yaml
- * @param {AssetEntry[1]} options - Asset options
+ * @param {Pick<AssetEntry[1], "srcPath" | "destPath">} options - Asset options
```
colinrotherham, line 46: This task already provides its own `filePath` and `fileContents`

Here we're narrowing down the types to _warn_ when passing in options that aren't picked

shared/tasks/files.mjs
```diff
@@ -11,25 +11,25 @@ import slash from 'slash'
  * Delete path globs for a given destination
  *
  * @param {string} pattern - Pattern to remove
- * @param {AssetEntry[1]} options - Asset options
+ * @param {Pick<AssetEntry[1], "destPath">} options - Asset options
```
colinrotherham, line 14: The clean task only needs a `destPath` so we're being quite strict here

Other properties can be passed in when not TypeScript `"strict": true` but they'll be ignored

romaricpascal (approved):
Cheers for updating to using spread 🙌🏻 ⛵

PR #3648, opened 2023-05-19
Configure tasks to use `rollup.config.mjs` files

colinrotherham:
This PR updates our JavaScript tasks to use `rollup.config.mjs` files in various flavours

This means we've done the hard work to enable CLI builds:

```console
npx rollup --config rollup.config.mjs --input example.mjs
```


romaricpascal (approved):
Neat, thanks for moving the config files in each package ⛵

PR #3649, opened 2023-05-19
Configure paths to `postcss.config.mjs` files

colinrotherham:
This PR updates our Stylesheets tasks with paths to `postcss.config.mjs`

We've used [`postcss-load-config`](https://github.com/postcss/postcss-load-config) to find configs automatically but it's safer to follow the https://github.com/alphagov/govuk-frontend/pull/3648 approach

We now pass the exact config path to [`postcss-load-config`](https://github.com/postcss/postcss-load-config)


romaricpascal (approved):
⛵

PR #3650, opened 2023-05-19
Ensure GitHub Actions runs type declaration checks in all files

colinrotherham:
This PR closes a gap raised by @domoscargin when discussing https://github.com/alphagov/govuk-frontend/pull/3571

I've enabled [`tsc` skipLibCheck](https://www.typescriptlang.org/tsconfig#skipLibCheck) to prevent 3rd party type issues from showing up in `tsc --build` so we can now check types on tests, tasks and other shared packages in our [**Tests** GitHub Actions workflow](https://github.com/alphagov/govuk-frontend/actions/workflows/tests.yml)

Previously we've seen the "red squiggles" locally but only checked “build output” in GitHub

### DIFF
--- a/tsconfig.base.json
+++ b/tsconfig.base.json
@@ -7,6 +7,7 @@
     "moduleResolution": "Node",
     "noEmit": true,
     "resolveJsonModule": true,
+    "skipLibCheck": true,
     "strict": false,
     "strictBindCallApply": true,
     "strictFunctionTypes": true,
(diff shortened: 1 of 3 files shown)


tsconfig.base.json
```diff
@@ -7,6 +7,7 @@
     "moduleResolution": "Node",
     "noEmit": true,
     "resolveJsonModule": true,
+    "skipLibCheck": true,
```
colinrotherham, line 10: We're currently seeing errors raised in:

```console
@jest/environment
@types/glob-stream
@types/glob
@types/jest
path-scurry
```

Typically this is where packages have shipped their own types, leaving the community types out of sync in various unmaintained packages

romaricpascal (approved):

PR #3558, opened 2023-04-26
Use `npm version` for version bumps

colinrotherham:
Found a bug whilst testing docs changes for:

* https://github.com/alphagov/govuk-frontend/issues/3547

When bumping package versions we unintentionally leave **package-lock.json** and other workspaces behind

## Using `npm version`
We should use [`npm version`](https://docs.npmjs.com/cli/v9/commands/npm-version?v=true) to bump our package version:

```console
npm version patch --workspace package
npm version minor --workspace package
npm version major --workspace package
npm version 5.0.0 --workspace package
```

This lets npm keep **package-lock.json** and workspaces that install `govuk-frontend` in sync

PR #3671, opened 2023-05-22
Post separate comments for JS, CSS and other files diffs

### LINKED ISSUE #3546: Improve diffing between previous release and new `dist`
## What

Limit the chances of Git considering that the files in the previous release and the ones in the release PR are different files entirely, as it leads to the diff showing the old file being removed and the new one being created.

We're currently using Git's [`-M/--find-renames`](https://git-scm.com/docs/diff-options#Documentation/diff-options.txt--Mltngt) option for which we can adjust the similarity index. This may just be pushing the issue to a later point, so it could be worth considering other routes (explicitely naming the files to compare, for example, as our releases as pretty structured).

## Why

The last release's diff had to be done locally rather than through the action as Git decided the old and new files were entirely different.

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [ ] We've implemented a way of diffing that compares the actual contents of the new and old files more reliably

romaricpascal:
This PRs updates the Github action running the diff of `dist` to post each diff (JavaScript, CSS, and other files) as a separate comment. This will limit the risk of them being over the 64Kb limit allowed by the Github API for posting comment and close #3546.

In addition to the separate comments, the action will also post a fallback message signposting to the action's artifacts where the diffs are uploaded (added for #3229).

> **Note** The diff display was triggered by 3bd7c5814) which has since then been removed. Similarly, the fallback message was tested by forcing an error and pushing (212da24c5386583f2ae7bc79e30a8c34f9ccef5a being the latest iteration), which has since been overwritten.

### DIFF
--- a/.github/workflows/diff-change-to-dist.yaml
+++ b/.github/workflows/diff-change-to-dist.yaml
@@ -52,35 +52,11 @@ jobs:
           github-token: ${{secrets.GITHUB_TOKEN}}
           script: |
             const { readFile } = require('node:fs/promises')
-            const { comment } = await import('${{ github.workspace }}/.github/workflows/scripts/comments.mjs')
+            const { commentDistDiffs } = await import('${{ github.workspace }}/.github/workflows/scripts/comments.mjs')
 
             // PR information
-            const number = '${{ github.event.pull_request.number }}'
+            const issueNumber = '${{ github.event.pull_request.number }}'
             const commit = '${{ github.event.pull_request.head.sha }}'
+            const workspacePath = '${{github.workspace}}'
 
-            const diffs = [{
-              file: 'dist-js.diff',
-              title: 'JavaScript changes to `dist`',
-              marker: 'dist-diff-js'
-              },{
-              file: 'dist-css.diff',
-              title: 'CSS changes to `dist`',
-              marker: 'dist-diff-css'
-            }, {
-              file: 'dist-other.diff',
-              title: 'Other changes to `dist`',
-              marker: 'dist-diff-other'
-            }]
-
-            for(const {file, title, marker} of diffs) {
-              // Read diff from previous step
-              const diffText = await readFile(`${{ github.workspace }}/${file}`, 'utf8')
-
-              // Add or update comment on PR
-              await comment({ github, context }, number, marker, `## ${title}\n` +
-                '```diff\n' +
-                `${diffText}\n` +
-                '```\n\n' +
-                `SHA: ${commit}\n`
-              )
-            }
+            await commentDistDiffs({github, context, workspacePath, issueNumber, commit})
--- a/.github/workflows/diff-change-to-dist.yaml
+++ b/.github/workflows/diff-change-to-dist.yaml
@@ -51,7 +51,7 @@ jobs:
         with:
           github-token: ${{secrets.GITHUB_TOKEN}}
           script: |
-            const { commentDiff } = await import('${{ github.workspace }}/.github/workflows/scripts/comments.mjs')
+            const { commentDiffs } = await import('${{ github.workspace }}/.github/workflows/scripts/comments.mjs')
 
             // PR information
             const issueNumber = ${{ github.event.pull_request.number }}
@@ -75,6 +75,4 @@ jobs:
               }
             ]
 
-            for (const diff of diffs) {
-              await commentDiff({ github, context, commit }, issueNumber, diff)
-            }
+            await commentDiffs({ github, context, commit }, issueNumber, diffs)
--- a/.github/workflows/diff-change-to-dist.yaml
+++ b/.github/workflows/diff-change-to-dist.yaml
@@ -54,17 +54,33 @@ jobs:
             const { readFile } = require('node:fs/promises')
             const { comment } = await import('${{ github.workspace }}/.github/workflows/scripts/comments.mjs')
 
-            // Read diff from previous step
-            const diffText = await readFile('${{ github.workspace }}/dist.diff', 'utf8')
-
             // PR information
             const number = '${{ github.event.pull_request.number }}'
             const commit = '${{ github.event.pull_request.head.sha }}'
 
-            // Add or update comment on PR
-            await comment({ github, context }, number, 'Diff release', '## Changes to dist\n' +
-              '```diff\n' +
-              `${diffText}\n` +
-              '```\n\n' +
-              `SHA: ${commit}\n`
-            )
+            const diffs = [{
+              file: 'dist-js.diff',
+              title: 'JavaScript changes to `dist`',
+              marker: 'dist-diff-js'
+              },{
+              file: 'dist-css.diff',
+              title: 'CSS changes to `dist`',
+              marker: 'dist-diff-css'
+            }, {
+              file: 'dist-other.diff',
+              title: 'Other changes to `dist`',
+              marker: 'dist-diff-other'
+            }]
+
+            for(const {file, title, marker} of diffs) {
+              // Read diff from previous step
+              const diffText = await readFile(`${{ github.workspace }}/${file}`, 'utf8')
+
+              // Add or update comment on PR
+              await comment({ github, context }, number, marker, `## ${title}\n` +
+                '```diff\n' +
+                `${diffText}\n` +
+                '```\n\n' +
+                `SHA: ${commit}\n`
+              )
+            }
--- a/.github/workflows/diff-change-to-dist.yaml
+++ b/.github/workflows/diff-change-to-dist.yaml
@@ -60,17 +60,17 @@ jobs:
             const diffs = [
               {
                 path: '${{ github.workspace }}/dist-js.diff',
-                title: 'JavaScript changes to `dist`',
+                titleText: 'JavaScript changes to `dist`',
                 markerText: 'dist-diff-js'
               },
               {
                 path: '${{ github.workspace }}/dist-css.diff',
-                title: 'Stylesheets changes to `dist`',
+                titleText: 'Stylesheets changes to `dist`',
                 markerText: 'dist-diff-css'
               },
               {
                 path: '${{ github.workspace }}/dist-other.diff',
-                title: 'Other changes to `dist`',
+                titleText: 'Other changes to `dist`',
                 markerText: 'dist-diff-other'
               }
             ]
--- a/.github/workflows/diff-change-to-dist.yaml
+++ b/.github/workflows/diff-change-to-dist.yaml
@@ -51,12 +51,30 @@ jobs:
         with:
           github-token: ${{secrets.GITHUB_TOKEN}}
           script: |
-            const { readFile } = require('node:fs/promises')
-            const { commentDistDiffs } = await import('${{ github.workspace }}/.github/workflows/scripts/comments.mjs')
+            const { commentDiff } = await import('${{ github.workspace }}/.github/workflows/scripts/comments.mjs')
 
             // PR information
-            const issueNumber = '${{ github.event.pull_request.number }}'
+            const issueNumber = ${{ github.event.pull_request.number }}
             const commit = '${{ github.event.pull_request.head.sha }}'
-            const workspacePath = '${{github.workspace}}'
 
-            await commentDistDiffs({github, context, workspacePath, issueNumber, commit})
+            const diffs = [
+              {
+                path: '${{ github.workspace }}/dist-js.diff',
+                title: 'JavaScript changes to `dist`',
+                markerText: 'dist-diff-js'
+              },
+              {
+                path: '${{ github.workspace }}/dist-css.diff',
+                title: 'Stylesheets changes to `dist`',
+                markerText: 'dist-diff-css'
+              },
+              {
+                path: '${{ github.workspace }}/dist-other.diff',
+                title: 'Other changes to `dist`',
+                markerText: 'dist-diff-other'
+              }
+            ]
+
+            for (const diff of diffs) {
+              await commentDiff({ github, context, commit }, issueNumber, diff)
+            }
--- a/.github/workflows/scripts/comments.mjs
+++ b/.github/workflows/scripts/comments.mjs
@@ -1,71 +1,54 @@
 import { readFile } from 'node:fs/promises'
 
-const DIST_DIFFS = [
-  {
-    file: 'dist-js.diff',
-    title: 'JavaScript changes to `dist`',
-    marker: 'dist-diff-js'
-  },
-  {
-    file: 'dist-css.diff',
-    title: 'CSS changes to `dist`',
-    marker: 'dist-diff-css'
-  },
-  {
-    file: 'dist-other.diff',
-    title: 'Other changes to `dist`',
-    marker: 'dist-diff-other'
-  }
-]
-
 /**
- * @param {GithubActionContext} githubActionContext - Github Action context
+ * Posts the content of a diff as a comment on a Github issue
+ *
+ * @param {GithubActionContext} githubActionContext
+ * @param {number} issueNumber
+ * @param {DiffComment} diffComment
  */
-export async function commentDistDiffs (
-  { github, context, workspacePath, issueNumber, commit }
+export async function commentDiff (
+  githubActionContext,
+  issueNumber,
+  { path, title, markerText }
 ) {
-  for (const { file, title, marker } of DIST_DIFFS) {
-    // Read diff from previous step
-    const diffText = await readFile(`${workspacePath}/${file}`, 'utf8')
+  // Read diff from previous step
+  const diffText = await readFile(path, 'utf8')
 
-    // Add or update comment on PR
-    try {
-      await comment({ github, context, issueNumber }, marker, `## ${title}\n` +
+  // Add or update comment on PR
+  try {
+    await comment(
+      githubActionContext,
+      issueNumber,
+      markerText,
+      `## ${title}\n` +
         '```diff\n' +
         `${diffText}\n` +
         '```\n\n' +
-        `SHA: ${commit}\n`
-      )
-    } catch {
-      await comment({ github, context, issueNumber }, marker, `## ${title}\n` +
-        `The diff could not be posted as a comment. You can download it from the [workflow artifacts](${githubActionArtifactsUrl(context.runId)}).` +
+        `SHA: ${githubActionContext.commit}\n`
+    )
+  } catch {
+    await comment(
+      githubActionContext,
+      issueNumber,
+      markerText,
+      `## ${title}\n` +
+        `The diff could not be posted as a comment. You can download it from the [workflow artifacts](${githubActionArtifactsUrl(
+          githubActionContext.context.runId
+        )}).` +
         '\n\n' +
-        `SHA: ${commit}\n`
-      )
-    }
+        `SHA: ${githubActionContext.commit}\n`
+    )
   }
 }
 
 /**
- * Generates a URL to the "Artifacts" section of a Github action run
- *
- * Unfortunately the best we can do here is a link to the "Artifacts" section as
- * [the upload-artifact action doesn't provide
- * the public URL](https://github.com/actions/upload-artifact/issues/50) :'(
- *
- * @param {number} runId - - The ID of the Github action run
- * @returns {string} The URL to the "Artifacts" section of the given run
- */
-function githubActionArtifactsUrl (runId) {
-  return `https://github.com/alphagov/govuk-frontend/actions/runs/${runId}/#artifacts`
-}
-
-/**
- * @param {Pick<GithubActionContext, "github" | "context" | "issueNumber">} githubContext - Github Action context
+ * @param {GithubActionContext} githubContext - Github Action context
+ * @param {number} issueNumber - The number of the issue/PR on which to post the comment
  * @param {string} markerText - A unique marker used to identify the correct comment
  * @param {string} bodyText - The body of the comment
  */
-export async function comment ({ github, context, issueNumber }, markerText, bodyText) {
+export async function comment ({ github, context }, issueNumber, markerText, bodyText) {
   const marker = `<!-- ${markerText} -->\n`
   const body = `${marker}${bodyText}`
 
@@ -99,7 +82,7 @@ export async function comment ({ github, context, issueNumber }, markerText, bod
   /**
    * Create GitHub issue comment
    *
-   * Updates existing commentby ID if available
+   * Updates existing comment by ID if available
    */
   await (!commentId
     ? github.rest.issues.createComment({ ...issueParameters, body })
@@ -107,12 +90,31 @@ export async function comment ({ github, context, issueNumber }, markerText, bod
 }
 
 /**
+ * Generates a URL to the "Artifacts" section of a Github action run
+ *
+ * Unfortunately the best we can do here is a link to the "Artifacts" section as
+ * [the upload-artifact action doesn't provide
+ * the public URL](https://github.com/actions/upload-artifact/issues/50) :'(
+ *
+ * @param {number} runId - - The ID of the Github action run
+ * @returns {string} The URL to the "Artifacts" section of the given run
+ */
+function githubActionArtifactsUrl (runId) {
+  return `https://github.com/alphagov/govuk-frontend/actions/runs/${runId}/#artifacts`
+}
+
+/**
  * @typedef {object} GithubActionContext
  * @property {import('@octokit/rest').Octokit} github - The pre-authenticated Octokit provided by Github actions
  * @property {import('@actions/github').context} context - The context of the Github action
- * @property {string} workspacePath - The path to the root of the
- * @property {number} issueNumber - This number of the issue the action relates to
  * @property {string} commit - The SHA of the commit that triggered the action
+ */
+
+/**
+ * @typedef {object} DiffComment
+ * @property {string} path - The path of the file to post as a comment
+ * @property {string} title - The title of the comment
+ * @property {string} markerText - The marker to identify the comment
  */
 
 /**
--- a/.github/workflows/scripts/comments.mjs
+++ b/.github/workflows/scripts/comments.mjs
@@ -19,12 +19,7 @@ const DIST_DIFFS = [
 ]
 
 /**
- * @param {object} githubContext - Github Action context
- * @param {import('@octokit/rest').Octokit} githubContext.github
- * @param {import('@actions/github').context} githubContext.context
- * @param {string} githubContext.workspacePath
- * @param {number} githubContext.issueNumber
- * @param {string} githubContext.commit
+ * @param {GithubActionContext} githubActionContext - Github Action context
  */
 export async function commentDistDiffs (
   { github, context, workspacePath, issueNumber, commit }
@@ -66,10 +61,7 @@ function githubActionArtifactsUrl (runId) {
 }
 
 /**
- * @param {object} githubContext - Github Action context
- * @param {import('@octokit/rest').Octokit} githubContext.github
- * @param {import('@actions/github').context} githubContext.context
- * @param {number} githubContext.issueNumber - GitHub issue or PR number
+ * @param {Pick<GithubActionContext, "github" | "context" | "issueNumber">} githubContext - Github Action context
  * @param {string} markerText - A unique marker used to identify the correct comment
  * @param {string} bodyText - The body of the comment
  */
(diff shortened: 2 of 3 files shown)


.github/workflows/diff-change-to-dist.yaml
```diff
@@ -52,19 +52,11 @@ jobs:
...
             const { readFile } = require('node:fs/promises')
-            const { comment } = await import('${{ github.workspace }}/.github/workflows/scripts/comments.mjs')
-
-            // Read diff from previous step
-            const diffText = await readFile('${{ github.workspace }}/dist.diff', 'utf8')
+            const { commentDistDiffs } = await import('${{ github.workspace }}/.github/workflows/scripts/comments.mjs')
 
             // PR information
-            const number = '${{ github.event.pull_request.number }}'
+            const issueNumber = '${{ github.event.pull_request.number }}'
```
colinrotherham, line 58: I couldn't bring myself to call it `issueNumber` for a PR 😆

.github/workflows/scripts/comments.mjs
```diff
@@ -1,12 +1,71 @@
...
+
+const DIST_DIFFS = [
+  {
+    file: 'dist-js.diff',
+    title: 'JavaScript changes to `dist`',
+    marker: 'dist-diff-js'
+  },
+  {
+    file: 'dist-css.diff',
+    title: 'CSS changes to `dist`',
```
colinrotherham, line 11: Minor thing, but we seem to say styles or "Stylesheets" to match scripts or "JavaScripts"

.github/workflows/scripts/github.mjs
```diff
@@ -0,0 +1,82 @@
...
+    ? github.rest.issues.createComment({ ...issueParameters, body })
+    : github.rest.issues.updateComment({
+      ...issueParameters,
+      body,
+      comment_id: commentId
+    }))
+}
+
+/**
+ * @typedef {object} GithubActionContext
```
romaricpascal, line 70: This is an attempt at formalising the structure of functions we'd call from Github action's script into something of the shape of `function (githubActionContext, ...argumentsSpecificToTheFunction)`.

The idea is to have a one stop object to collect all the values from the github context in the action via `${{github.<xyz>}}` and clearly separate them from the parameters that are actually specific to the function's job.

This is only a starting point to set a direction, and can be adjusted as we work more with Github action. If we realise it needs splitting into different things, or smaller types or whatever, it can happily be refactored then.

romaricpascal (commented):
I think the code is pretty much there. I've moved the JS files a little to separate Github specific code from code that relates to what we need to do for GOV.UK Frontend specifically (especially as we're planning more commenting in the future). Hopefully it all makes sense.

bin/dist-diff.sh
```diff
@@ -14,16 +14,16 @@ output_folder="${2:-`pwd`}"
 # Diff the minified JS file
 git diff \
   $base:dist/govuk-frontend-$(git show $base:dist/VERSION.txt).min.js \
-  dist/govuk-frontend-$(cat dist/VERSION.txt).min.js \
+  dist/govuk-frontend-?.?.?.min.js \
```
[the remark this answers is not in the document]
colinrotherham: Will work great for simple [`npm version`](https://github.com/alphagov/govuk-frontend/pull/3558) calls:

```sh
npm version patch # v5.0.1
npm version minor # v5.1.0
npm version major # v6.0.0
```

But what about public prereleases etc?

```sh
npm version prepatch --preid rc #v5.0.1-rc.0
```

Can you grab the version via the npm CLI instead?

```console
npm pkg get version --workspace govuk-frontend
```

.github/workflows/scripts/github.mjs
colinrotherham: Sounds good

Mind if we narrow down `GitHubActionContext` to only what we get from GitHub?

Good to separate out our own params (either as actual params or an object)

```suggestion
 * @typedef {object} GitHubActionContext
```

bin/dist-diff.sh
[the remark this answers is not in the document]
romaricpascal: Ohh, hadn't thought of that, that's a good shout! I'll revert back to reading VERSION.txt then for consistency with what we do when grabbing the file from the Git history 😊

.github/workflows/scripts/github.mjs
```diff
@@ -0,0 +1,82 @@
...
+ * Generates a URL to the "Artifacts" section of a Github action run
+ *
+ * Unfortunately the best we can do here is a link to the "Artifacts" section as
+ * [the upload-artifact action doesn't provide
+ * the public URL](https://github.com/actions/upload-artifact/issues/50) :'(
+ *
+ * @param {number} runId - - The ID of the Github action run
+ * @returns {string} The URL to the "Artifacts" section of the given run
+ */
+export function githubActionArtifactsUrl (runId) {
```
colinrotherham, line 11: Would you mind keeping this file as `comments.mjs` for now?

For some reason it looks like we've deleted and totally rewritten it

Would be good for each comment to have predefined metadata like:

1. Commit SHA
2. Date updated
3. Link to workflow run

i.e. Don't need to hide `githubActionArtifactsUrl()` behind a try/catch, add the useful info to all comments?

.github/workflows/scripts/dist-diff.mjs
```diff
@@ -0,0 +1,49 @@
...
+
+const DIST_DIFFS = [
+  {
+    file: 'dist-js.diff',
+    title: 'JavaScript changes to `dist`',
+    marker: 'dist-diff-js'
+  },
+  {
+    file: 'dist-css.diff',
+    title: 'CSS changes to `dist`',
```
colinrotherham, line 13: https://github.com/alphagov/govuk-frontend/pull/3671#discussion_r1200744106

```suggestion
    title: 'Stylesheet changes to `dist`',
```

.github/workflows/scripts/dist-diff.mjs
```diff
@@ -0,0 +1,49 @@
+import { readFile } from 'node:fs/promises'
+
+import { comment, githubActionArtifactsUrl } from './github.mjs'
+
+const DIST_DIFFS = [
```
colinrotherham, line 5: Can we pass this in from the GitHub Action YAML?

Then we can keep everything in **comments.mjs** with maybe:

1. Comment as `comment()`
2. Comment single diff as `commentDiff()`
3. Comment multiple diffs as `commentDiffs()`

Preserves the change history from @domoscargin too

.github/workflows/scripts/github.mjs
colinrotherham: Argh sorry, wasn't clear was I? 🤦‍♂️ 

I mean the objects and tooling we get from [`actions/github-script`](https://github.com/actions/github-script/tree/v6/#readme)

>* `github` A pre-authenticated [octokit/rest.js](https://octokit.github.io/rest.js) client with pagination plugins
context An object containing the [context of the workflow run](https://github.com/actions/toolkit/blob/main/packages/github/src/context.ts)
>* `core` A reference to the [@actions/core](https://github.com/actions/toolkit/tree/main/packages/core) package
>* `glob` A reference to the [@actions/glob](https://github.com/actions/toolkit/tree/main/packages/glob) package
>* `io` A reference to the [@actions/io](https://github.com/actions/toolkit/tree/main/packages/io) package
>* `exec` A reference to the [@actions/exec](https://github.com/actions/toolkit/tree/main/packages/exec) package
>* `fetch` A reference to the [node-fetch](https://github.com/node-fetch/node-fetch) package

I liked this approach:

```js
export async function comment ({ github, context }, issueNumber, markerText, bodyText) {}
```

Unlike the shared `github` and `context` objects we're likely to vary these per function:

```
issueNumber
markerText
bodyText
```

.github/workflows/scripts/comments.mjs
```diff
@@ -1,14 +1,67 @@
...
 /**
- * @param {object} githubContext - Github Action context
- * @param {import('@octokit/rest').Octokit} githubContext.github
- * @param {import('@actions/github').context} githubContext.context
- * @param {number} issueNumber - GitHub issue or PR number
+ * @param {string} title - The title for the comment
+ * @param {string} diffText - The diff content
+ * @returns {string} The content for the comment body
+ */
+function renderDiffBodyText (title, diffText) {
```
romaricpascal, line 44: I've added a special case when the diff is empty so the comment does not look broken, making us wonder if something went wrong.

.github/workflows/scripts/comments.mjs
```diff
@@ -40,13 +93,49 @@ export async function comment ({ github, context }, issueNumber, markerText, bod
...
     ? github.rest.issues.createComment({ ...issueParameters, body })
     : github.rest.issues.updateComment({ ...issueParameters, body, comment_id: commentId }))
 }
 
+/**
+ * @param {Pick<GithubActionContext, "context" | "commit">} githubActionContext
+ * @returns {string} - The content for the footer
+ */
+function renderCommentFooter ({ context, commit }) {
+  return `Commit: ${commit} | [Action run](${githubActionRunUrl(context.runId)})`
```
romaricpascal, line 108: I've strayed away from dates. While we could pick a timezone and use the `Intl` API for formatting, it wouldn't be clear to anyone reading the comment. Plus the info is kind of available under the "edited" dropdown.

.github/workflows/scripts/comments.mjs
```diff
@@ -1,14 +1,67 @@
...
+      markerText,
+      renderDiffBodyText(title, diffText)
+    )
+  } catch {
+    await comment(
+      githubActionContext,
+      issueNumber,
+      markerText,
+      `## ${title}\n` +
+        `The diff could not be posted as a comment. You can download it from the [workflow artifacts](${githubActionRunUrl(
```
romaricpascal, line 32: I've kept this in addition to putting the link to the action run in the footer as:
- we want to replace any previous comment with something, so may as well be informative
- we want to point people to download the diffs, which are in the artifacts section of that page

romaricpascal (commented):
@colinrotherham I think I've addressed everything, let me know if I missed a comment. I've also handled empty diffs in a custom way to avoid confusion if that happens (should we start running on branches, the JS or CSS may not have had updates).

.github/workflows/diff-change-to-dist.yaml
```diff
@@ -52,19 +52,30 @@ jobs:
...
-            // Add or update comment on PR
-            await comment({ github, context }, number, 'Diff release', '## Changes to dist\n' +
-              '```diff\n' +
-              `${diffText}\n` +
-              '```\n\n' +
-              `SHA: ${commit}\n`
-            )
+            const diffs = [
+              {
+                path: '${{github.workspace}}/dist-js.diff',
```
colinrotherham, line 63: I know it's unlikely but I've developed an "oh no, Windows" reflex 😆

For some reason it only kicked in today so this isn't a blocker—but think we need `join()`?

Otherwise just double check all the spacing for consistency?

```suggestion
                path: '${{ github.workspace }}/dist-js.diff',
```

.github/workflows/diff-change-to-dist.yaml
```diff
@@ -52,19 +52,30 @@ jobs:
...
             const { readFile } = require('node:fs/promises')
-            const { comment } = await import('${{ github.workspace }}/.github/workflows/scripts/comments.mjs')
-
-            // Read diff from previous step
-            const diffText = await readFile('${{ github.workspace }}/dist.diff', 'utf8')
+            const { commentDiff } = await import('${{ github.workspace }}/.github/workflows/scripts/comments.mjs')
 
             // PR information
-            const number = '${{ github.event.pull_request.number }}'
+            const issueNumber = '${{ github.event.pull_request.number }}'
```
colinrotherham, line 58: It was me that used strings (sorry) but looks like we've now switched to type number?

```suggestion
            const issueNumber = ${{ github.event.pull_request.number }}
```

.github/workflows/diff-change-to-dist.yaml
```diff
@@ -52,19 +52,30 @@ jobs:
...
+              },
+              {
+                path: '${{github.workspace}}/dist-other.diff',
+                title: 'Other changes to `dist`',
+                markerText: 'dist-diff-other'
+              }
+            ]
+
+            for (const diff of diffs) {
+              await commentDiff({github, context, commit}, issueNumber, diff)
```
colinrotherham, line 80: Mind just running all this code through ESLint again? Just in case

```suggestion
              await commentDiff({ github, context, commit }, issueNumber, diff)
```

.github/workflows/scripts/comments.mjs
```diff
@@ -1,14 +1,70 @@
...
+  { path, title, markerText }
+) {
+  // Read diff from previous step
+  const diffText = await readFile(path, 'utf8')
+
+  // Add or update comment on PR
+  try {
+    await comment(
+      githubActionContext,
+      issueNumber,
```
colinrotherham, line 22: We've used a nice **DiffComment** object above `{ path, title, markerText }`

Should we add an optional `[bodyText]` to the type and merge it in? So we're being consistent

```js
await comment(githubActionContext, { ...diffComment, bodyText })
```

.github/workflows/scripts/comments.mjs
```diff
@@ -1,14 +1,70 @@
...
+ * @param {number} issueNumber
+ * @param {DiffComment} diffComment
+ */
+export async function commentDiff (
+  githubActionContext,
+  issueNumber,
+  { path, title, markerText }
+) {
+  // Read diff from previous step
+  const diffText = await readFile(path, 'utf8')
```
colinrotherham, line 16: Can we save a whole function by defaulting here?

```suggestion
  const diffText = (await readFile(path, 'utf8')) ?? 'The diff is empty.'
```

.github/workflows/scripts/comments.mjs
colinrotherham: I found it a bit hard to follow but [added some suggestions](https://github.com/alphagov/govuk-frontend/pull/3671/files#r1202654624)

.github/workflows/scripts/comments.mjs
```diff
@@ -1,14 +1,70 @@
...
+}
+
+/**
+ * @param {GithubActionContext} githubContext - Github Action context
+ * @param {number} issueNumber - The number of the issue/PR on which to post the comment
  * @param {string} markerText - A unique marker used to identify the correct comment
  * @param {string} bodyText - The body of the comment
  */
-export async function comment ({ github, context }, issueNumber, markerText, bodyText) {
+export async function comment ({ github, context, commit }, issueNumber, markerText, bodyText) {
```
colinrotherham, line 64: I quite like the obvious params:

```js
function ({ github, context, commit }, issueNumber, markerText, bodyText) {}
```

But if we're adopting objects on [#L13](https://github.com/alphagov/govuk-frontend/pull/3671/files#diff-e831ad02a0354c8c4b18febab2f20c25a2baa97f742c13b75499020dd29d58c9R13) let's be consistent:

```js
function ({ github, context, commit }, issueNumber, { title, bodyText }) {}
```

.github/workflows/scripts/comments.mjs
colinrotherham: Where workflow runs have been run again, are we alright linking to the latest?

The comment might have been written by a previous run:

https://github.com/alphagov/govuk-frontend/actions/runs/4894152384/attempts/1
https://github.com/alphagov/govuk-frontend/actions/runs/4894152384/attempts/2
https://github.com/alphagov/govuk-frontend/actions/runs/4894152384

Haven't thought too much about GitHub comments + screen readers but is `|` "Vertical Bar" alright?

```console
Commit: ${commit}
Workflow run: ${githubActionRunUrl(context.runId)}
```

.github/workflows/scripts/comments.mjs
```diff
@@ -1,14 +1,70 @@
...
+  // Read diff from previous step
+  const diffText = await readFile(path, 'utf8')
+
+  // Add or update comment on PR
+  try {
+    await comment(
+      githubActionContext,
+      issueNumber,
+      markerText,
+      renderDiffBodyText(title, diffText)
```
colinrotherham, line 24: I found this a bit hard to follow but [added some suggestions](https://github.com/alphagov/govuk-frontend/pull/3671/files#r1202654624)

.github/workflows/scripts/comments.mjs
romaricpascal: Good shout on the accessibility, I'll update, maybe to something smaller like "[Workflow run](url-of-fun) for [CommitSHA](url-of-commit)" so we eat as little characters as possible for that footer.

Good spot also for the attempts, I'll see if I can dig the URL to the specific attempt (hopefully it's somewhere in the `context` object or something else). If I cannot, we'll have to set it to the latest run.

.github/workflows/scripts/comments.mjs
romaricpascal: If the `await comment()` crashes for another reason, "The diff could not be posted as a comment" is still true and it still got uploaded to the artifacts by the previous step in the workflow. That gives us a second chance at having a useful message posted on the PR (as well as to reserve an early comment if the posting fails because of an unlucky technical glitch on the first comment attempt).

I can try to get a large diff happening by changing the commit that triggers the changes in `dist` using say 4.4.1 or something for the JS.

.github/workflows/diff-change-to-dist.yaml
romaricpascal: The action is explicitely set to run on Ubuntu, so I don't think paths are a worry (I don't see why we'd want to broaden the diffing Github action to other platforms, unlike the build our contributors).

Regarding the spacing, how would you feel about setting us up to have Github workflow running scripts always depend on a proper JS file (say `<name-of-the-workflow>-<name-of-the-step>.js`), and just require the file and run it passing `github` and `context` + whatever other options are needed). This may make typing a bit wonky due to the type for `github` not being available, but will solve all formatting/linting issues 😊

.github/workflows/scripts/comments.mjs
romaricpascal: I like that first one, hadn't thought I could solve it at that point, that'll simplify things greatly! 🙌🏻

.github/workflows/scripts/comments.mjs
```diff
@@ -40,13 +96,45 @@ export async function comment ({ github, context }, issueNumber, markerText, bod
...
+}
+
+/**
+ * Generates a URL to the Github action run
+ *
+ * @param {number} runId - - The ID of the Github action run
+ * @returns {string} The URL to the "Artifacts" section of the given run
+ */
+function githubActionRunUrl (runId) {
+  return `https://github.com/alphagov/govuk-frontend/actions/runs/${runId}`
```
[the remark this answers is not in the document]
romaricpascal: Good spot! I'll see where I can fish those from the data provided to Github scripts.

.github/workflows/scripts/comments.mjs
colinrotherham: Sorry @romaricpascal I mostly meant we call `await comment()` from inside the catch block

So if the 2nd `await` rejects there's nothing to catch it and the build will unintentionally fail

Could always let it crash but with [`continue-on-error: true`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error) and write a message?

1. [Adding a job summary](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#adding-a-job-summary)
2. [Setting a warning message](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-a-warning-message)

actions-job-summary-simple-example

colinrotherham: But the backup comment is a bonus feature we can live without

Footer links are probably enough?

Would like to see how we handle crashes though (so we don't block a release PR merge)

.github/workflows/diff-change-to-dist.yaml
colinrotherham: Yeah possibly, bit much for this one. Thanks for tidying it up

.github/workflows/scripts/comments.mjs
romaricpascal: I don't think we should merge the two things, `comment` doesn't really has a reason to accept a `path` (which we would pass by spreading `...diffComment`) given its role is just to post a comment on Github. TypeScript would likely complain about that as well, wouldn't it?

.github/workflows/scripts/comments.mjs
colinrotherham: That's good

Maybe remove the try/catch and we can address it when it happens?

.github/workflows/scripts/comments.mjs
colinrotherham: Excuse the TypeScript code but with two comment flavours:

```ts
type IssueComment = {
  issue: number
  title: string
  bodyText: string
}

type DiffComment = {
  path: string
  markerText: string
  bodyText?: string
} & IssueComment
```

That gives me a way of writing comments that already have `bodyText` populated

```mjs
/**
 * 
 * @param {GithubActionContext} context 
 * @param {IssueComment | DiffComment} comment
 */
async function comment(context, comment) {
  const { github } = context

  // Detect diff comments with paths
  if (!comment.bodyText && 'path' in comment) {
    comment.bodyText = 'Load from disk example'
  }

  github.rest.issues.createComment({
    /* Call API example */
  })
}
```

.github/workflows/scripts/comments.mjs
romaricpascal: Soo....

Turns out `continue-on-error` is a risky business, as it hides all errors from the workflow status (like [me forgetting to rename the function after introducting `commentDiffs`](https://github.com/alphagov/govuk-frontend/actions/runs/5082430875/jobs/9132138191?pr=3671)), so let's avoid it.

Simulating errors with `throw` in 2340312f5, this is the output we get, which is not too bad (only lists "Error" as a reason because I threw manually, not sure what would be logged for an actual errors from Github API), I think 😊 

```sh
Error: Posting diff comment failed
    at commentDiffs (file:///home/runner/work/govuk-frontend/govuk-frontend/.github/workflows/scripts/comments.mjs:26:11)
Error: Unhandled error: Error: Posting diff comment failed
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async eval (eval at callAsyncFunction (/home/runner/work/_actions/actions/github-script/v6/dist/index.js:15143:16), <anonymous>:27:1)
    at async main (/home/runner/work/_actions/actions/github-script/v6/dist/index.js:15236:20) {
  [cause]: [
    {
      status: 'rejected',
      reason: Error
          at commentDiff (file:///home/runner/work/govuk-frontend/govuk-frontend/.github/workflows/scripts/comments.mjs:75:47)
          at processTicksAndRejections (node:internal/process/task_queues:96:5)
    }
  ]
}
```

With the workflow failed status, the trace in the workflow logs and the the difference in the hashes in the comments (if a comment was posted at all), I think that's plenty of information to start looking into things. We can add more logging if we investigate and realise there's more we need to speed us up the next times 😊

.github/workflows/scripts/comments.mjs
```diff
@@ -1,14 +1,96 @@
...
+  // Run the comments in parallel to avoid that an early one failing prevents the others
+  // and use `allSettled` to avoid the promise rejecting on the first failure but wait
+  // for everything to complete
+  const results = await Promise.allSettled(
+    diffs.map((diff) =>
+      commentDiff(githubActionContext, issueNumber, diff)
+    )
+  )
+
+  const resultsWithErrors = results.filter(result => result.status === 'rejected')
```
colinrotherham, line 24: This is much nicer, thanks for passing `diffs` in 😊

For some reason TypeScript is typing `resultsWithErrors` incorrectly

```
PromiseSettledResult<void>[]
```

Might be similar to the [`.map()` inline typing](https://github.com/alphagov/govuk-frontend/blob/7fe712acc5f01da1a014f37b4c4a94ff17d0a6c9/shared/tasks/browser.mjs#L36) I had to do recently

I'll have a look tomorrow

.github/workflows/scripts/comments.mjs
```diff
@@ -1,14 +1,96 @@
...
+  if (resultsWithErrors.length) {
+    throw new Error('Posting diff comment failed', { cause: resultsWithErrors })
```
colinrotherham, line 26: This is nice too, love using error causes

Is there friendlier error info by grabbing the `{ reason }` property off all those rejected promises?

```js
if (results.some(({ status }) => status === 'rejected')) {
 // … Map each `{ reason }` property from `PromiseRejectedResult`
}

Comment (romaricpascal) [reply .github/workflows/scripts/comments.mjs:26]: It'll likely be less verbose as we don't really need to know the `status` is rejected in the logs. I'll do a quick mapping there 😊

Comment (romaricpascal) [reply .github/workflows/scripts/comments.mjs:26]: Actually no, I think we might need the verbosity there and my `filter` is getting in the way of us understanding what broke. The `filter` makes us lose the index in the array of rejected Promises, so we don't know which comment failed. We'll see from Github UI, but we may as well dump the whole `results` in the logs to avoid back and forth between the two views 😊

Comment (romaricpascal) [reply .github/workflows/scripts/comments.mjs]: Yeah, now we know which posting failed from the log:

```
Error: Posting diff comment failed
    at commentDiffs (file:///home/runner/work/govuk-frontend/govuk-frontend/.github/workflows/scripts/comments.mjs:25:11)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
Error: Unhandled error: Error: Posting diff comment failed
    at async eval (eval at callAsyncFunction (/home/runner/work/_actions/actions/github-script/v6/dist/index.js:15143:16), <anonymous>:27:1)
    at async main (/home/runner/work/_actions/actions/github-script/v6/dist/index.js:15236:20) {
  [cause]: [
    { status: 'fulfilled', value: undefined },
    {
      status: 'rejected',
      reason: Error
          at commentDiff (file:///home/runner/work/govuk-frontend/govuk-frontend/.github/workflows/scripts/comments.mjs:62:47)
    },
    { status: 'fulfilled', value: undefined }
  ]
}
```

That also solves any typing issue I believe as we're not manipulating the results, just dumping them 😊

File: .github/workflows/scripts/comments.mjs [hunk 21]
```diff
@@ -1,14 +1,97 @@
...
+}
+
 /**
- * @param {object} githubContext - Github Action context
- * @param {import('@octokit/rest').Octokit} githubContext.github
- * @param {import('@actions/github').context} githubContext.context
- * @param {number} issueNumber - GitHub issue or PR number
- * @param {string} markerText - A unique marker used to identify the correct comment
- * @param {string} bodyText - The body of the comment
+ * Posts the content of a diff as a comment on a Github issue
```
colinrotherham, line 32:
```suggestion
 * Posts the content of a diff as a comment on a GitHub issue
```

.github/workflows/scripts/comments.mjs
```diff
@@ -1,14 +1,97 @@
...
+        bodyText: `The diff could not be posted as a comment. You can download it from the [workflow artifacts](${githubActionRunUrl(
+          githubActionContext.context
+        )}#artifacts).`
+      }
+    )
+  }
+}
+
+/**
+ * @param {GithubActionContext} githubContext - Github Action context
```
colinrotherham, line 79:
```suggestion
 * @param {GithubActionContext} githubContext - GitHub Action context
```

.github/workflows/scripts/comments.mjs
```diff
@@ -40,13 +123,47 @@ export async function comment ({ github, context }, issueNumber, markerText, bod
...
+/**
+ * @param {Pick<GithubActionContext, "context" | "commit">} githubActionContext
+ * @returns {string} - The content for the footer
+ */
+function renderCommentFooter ({ context, commit }) {
+  return `[Action run](${githubActionRunUrl(context)}) for ${commit}`
+}
+
+/**
+ * Generates a URL to the Github action run
```
colinrotherham, line 142:
```suggestion
 * Generates a URL to the GitHub action run
```

.github/workflows/scripts/comments.mjs
```diff
@@ -40,13 +123,47 @@ export async function comment ({ github, context }, issueNumber, markerText, bod
...
+ */
+function githubActionRunUrl (context) {
+  const { runId, repo } = context
+
+  return `https://github.com/${repo.owner}/${repo.repo}/actions/runs/${runId}/attempts/${process.env.GITHUB_RUN_ATTEMPT}`
+}
+
+/**
+ * @typedef {object} GithubActionContext
+ * @property {import('@octokit/rest').Octokit} github - The pre-authenticated Octokit provided by Github actions
```
colinrotherham, line 155:
```suggestion
 * @property {import('@octokit/rest').Octokit} github - The pre-authenticated Octokit provided by GitHub actions
```

.github/workflows/scripts/comments.mjs
```diff
@@ -40,13 +123,47 @@ export async function comment ({ github, context }, issueNumber, markerText, bod
...
+ * @property {import('@actions/github').context} context - The context of the Github action
```
colinrotherham, line 156:
```suggestion
 * @property {import('@actions/github').context} context - The context of the GitHub Action
```

colinrotherham (approved):

Issue #3281, opened 2023-02-13
Gather breakdown of modules content


romaricpascal:
## What

Collect which modules get added to built bundles when importing our modules[^1]. This can be:

- importing individual components directly (`import Accordion from 'govuk-frontend/components/accordion.mjs'`)
- importing individual components from `all.mjs` (`import {Accordion} from 'govuk-frontend'`)
- importing all exports (`import * as GOVUKFrontend from 'govuk-frontend'`)

Display this breakdown inside the review app so it can be accessible for each pull-request.

> **Note**: The performance spike used the approach of building from a virtual file with a specific import for each scenario. We may be able to collect the same stats more efficiently at build time, so it'll be worth giving a quick comparison at both options.

## Why

This will let us check which sub-modules are part of each of our main modules, and spot when our changes make a component inadvertently bring in a sub-module they don't need.

This also lets us check that our modules tree-shake as expecting, bringing only the code they need in the final bundle.

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [ ] We collect breakdown of modules
- [ ] We serve this breakdown from the review app [^2]:
  - [ ] as a text breakdown for accessibility
  - [ ] as an interactive visualisation 
- [ ] We list all modules being monitored in an index page so we can navigate to them

[^1]: Using a similar approach to the performance spike, with `rollup-plugin-visualizer`
[^2]: rollup-plugin-visualizer provides 3 kinds of visualisation, but we only really need the treemap or sunburst. Neither are particularly accessible, though (information is only available on hover), hence the text alternative we need to have.

Issue #3280, opened 2023-02-13
Gather built files sizes and display them in review app


romaricpascal:
Part of https://github.com/alphagov/govuk-frontend/issues/3279 

## What

Collect the size of the files built in `dist` as well as `package/govuk`,
and display them in the review app [^1].

Ideally this would work as a two step thing so the sizes may be displayed in other ways (like a comment on a PR) or stored somewhere more permanent.

If quick enough to add, it could be neat to also collect Gzip and Brotli size (the [gzip-size](https://www.npmjs.com/package/gzip-size) and [brotli-size](https://www.npmjs.com/package/brotli-size) packages may be helpful there).

## Why

This is one of the steps towards keeping an eye on the size of the files we generate. It'll let us check if our changes introduce an unexpected variation of size in the built files.

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [ ] Collect size of files built in `dist` and `package/govuk` into JSON
- [ ] Render collected sizes on a page in the review app[^2]
- [ ] Either:
  - [ ] Collect Gzip and Brotli sizes
  - or
  - [ ] Create an issue for future collection of Gzip and Brotli sizes

[^1]: For humanizing the file sizes, the [filesize](https://www.npmjs.com/package/filesize) package may come handy.
[^2]: This will/may call for building `dist` and `package` on heroku, which we might be able to speed up by running in parallel, using something like [`npm-run-all`](https://www.npmjs.com/package/npm-run-all).

PR #3681, opened 2023-05-23
Add Rollup stats to review app

colinrotherham:
This PR adds a new package to generate Rollup build stats for:

* https://github.com/alphagov/govuk-frontend/issues/3281
* https://github.com/alphagov/govuk-frontend/issues/3280

Changes also include:

1. Enables JSDoc type discovery for `govuk-frontend` in any workspace
2. Restarts review app on shared package changes

JavaScript module stats

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,9 +9,9 @@
   "license": "MIT",
   "workspaces": [
     ".github/workflows/scripts",
-    "shared/*",
     "packages/govuk-frontend",
     "packages/govuk-frontend-review",
+    "shared/*",
     "docs/examples/*"
   ],
   "scripts": {
--- a/packages/govuk-frontend/tasks/scripts.mjs
+++ b/packages/govuk-frontend/tasks/scripts.mjs
@@ -1,6 +1,6 @@
 import { join, resolve } from 'path'
 
-import { configs, scripts, task } from 'govuk-frontend-tasks'
+import { configs, scripts, npm, task } from 'govuk-frontend-tasks'
 import gulp from 'gulp'
 
 /**
@@ -43,5 +43,8 @@ export const compile = (options) => gulp.series(
       srcPath: join(options.srcPath, 'govuk-prototype-kit'),
       destPath: resolve(options.destPath, '../') // Top level (not dist) for compatibility
     })
-  )
+  ),
+
+  // Compile GOV.UK Frontend build stats
+  npm.script('build:stats', [], options)
 )
--- a/shared/stats/package.json
+++ b/shared/stats/package.json
@@ -1,0 +1,24 @@
+{
+  "private": true,
+  "name": "govuk-frontend-stats",
+  "description": "GOV.UK Frontend build stats",
+  "main": "src/index.mjs",
+  "engines": {
+    "node": "^18.12.0",
+    "npm": "^8.1.0 || ^9.1.0"
+  },
+  "license": "MIT",
+  "scripts": {
+    "stats": "rollup --config rollup.config.mjs"
+  },
+  "dependencies": {
+    "govuk-frontend-config": "*",
+    "govuk-frontend-lib": "*"
+  },
+  "devDependencies": {
+    "@rollup/plugin-node-resolve": "^15.0.2",
+    "govuk-frontend": "*",
+    "rollup": "^3.23.0",
+    "rollup-plugin-visualizer": "^5.9.0"
+  }
+}
--- a/shared/stats/rollup.config.mjs
+++ b/shared/stats/rollup.config.mjs
@@ -27,10 +27,19 @@ export default defineConfig(modulePaths
      */
     plugins: [
       resolve(),
+
+      // Stats: File size
       visualizer({
         filename: join('dist', dirname(modulePath), `${parse(modulePath).name}.yaml`),
         projectRoot: packageNameToPath('govuk-frontend', 'dist/govuk-esm/'),
         template: 'list'
+      }),
+
+      // Stats: Module tree map
+      visualizer({
+        filename: join('dist', dirname(modulePath), `${parse(modulePath).name}.html`),
+        projectRoot: packageNameToPath('govuk-frontend', 'dist/govuk-esm/'),
+        template: 'treemap'
       })
     ]
   }))
--- a/shared/stats/rollup.config.mjs
+++ b/shared/stats/rollup.config.mjs
@@ -1,0 +1,37 @@
+import { dirname, join, parse } from 'path'
+
+import resolve from '@rollup/plugin-node-resolve'
+import { packageNameToPath } from 'govuk-frontend-lib/names'
+import { defineConfig } from 'rollup'
+import { visualizer } from 'rollup-plugin-visualizer'
+
+import { modulePaths } from './src/index.mjs'
+
+/**
+ * Rollup config with stats output
+ */
+export default defineConfig(modulePaths
+  .map((modulePath) => /** @satisfies {import('rollup').RollupOptions} */({
+    input: join('govuk-frontend/dist/govuk-esm', modulePath),
+
+    /**
+     * Output options
+     */
+    output: {
+      file: join('dist', modulePath),
+      format: 'es'
+    },
+
+    /**
+     * Input plugins
+     */
+    plugins: [
+      resolve(),
+      visualizer({
+        filename: join('dist', dirname(modulePath), `${parse(modulePath).name}.yaml`),
+        projectRoot: packageNameToPath('govuk-frontend', 'dist/govuk-esm/'),
+        template: 'list'
+      })
+    ]
+  }))
+)
--- a/.github/workflows/actions/build/action.yml
+++ b/.github/workflows/actions/build/action.yml
@@ -11,7 +11,9 @@ runs:
       with:
         # Restore build cache (unless commit SHA changes)
         key: build-cache-${{ runner.os }}-${{ github.sha }}
-        path: packages/*/dist
+        path: |
+          packages/*/dist
+          shared/*/dist
 
     - name: Build
       id: build
--- a/docs/contributing/tasks.md
+++ b/docs/contributing/tasks.md
@@ -41,6 +41,7 @@ npm scripts are defined in `package.json`. These trigger a number of Gulp tasks.
 - copy GOV.UK Prototype Kit config files
 - compile JavaScript to ECMAScript (ES) modules
 - compile JavaScript to Universal Module Definition (UMD)
+- compile Rollup build stats into `./shared/govuk-frontend-stats/dist`
 - runs `npm run postbuild:package` (which will test the output is correct)
 
 **`npm run build:release` will do the following:**
--- a/package-lock.json
+++ b/package-lock.json
@@ -27360,6 +27360,7 @@
         "govuk-frontend": "*",
         "govuk-frontend-config": "*",
         "govuk-frontend-lib": "*",
+        "govuk-frontend-stats": "*",
         "highlight.js": "^11.8.0",
         "iframe-resizer": "^4.3.6",
         "js-beautify": "^1.14.7",
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,9 +9,9 @@
       "license": "MIT",
       "workspaces": [
         ".github/workflows/scripts",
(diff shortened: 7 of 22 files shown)


package.json
```diff
@@ -9,9 +9,9 @@
   "license": "MIT",
   "workspaces": [
     ".github/workflows/scripts",
-    "shared/*",
     "packages/govuk-frontend",
     "packages/govuk-frontend-review",
+    "shared/*",
```
colinrotherham, line 14: For `npm run build --workspaces` we need **govuk-frontend** to build before shared packages

shared/stats/rollup.config.mjs
```diff
@@ -0,0 +1,36 @@
...
+ * Rollup config with stats output
+ *
+ * @param {Record<string, string | undefined>} args - Rollup CLI arguments
+ * @returns {import('rollup').MergedRollupOptions[]} Rollup options
+ */
+export default ({ i: input }) => [
+  configESM({
+    i: input,
+    d: join(paths.stats, 'dist/report/esm')
+  }),
```
romaricpascal, line 18: I'm not sure this sets us up to get the full picture. If I followed correctly, this will only build with `govuk-frontend` as an input, which will be resolved to `all.mjs` only. This means we may miss broken tree shaking like what we had to fix in https://github.com/alphagov/govuk-frontend/pull/2959.

Will this also get us when people import specific components from GOV.UK Frontend?

shared/stats/rollup.config.mjs
```diff
@@ -0,0 +1,36 @@
...
+  configUMD({
+    i: input,
+    d: join(paths.stats, 'dist/report/umd')
+  })
```
romaricpascal, line 22: How do the stats differ from the ESM build? The modules should be the same, with the same code built in and the only overhead being the UMD boilerplate, which will likely get dropped for v5.

shared/stats/rollup.config.mjs
```diff
@@ -0,0 +1,36 @@
+import { join } from 'path'
+
+import configESM from 'govuk-frontend/rollup.esm.config.mjs'
```
romaricpascal, line 3: Wondering if we do need to reuse the configs from govuk-frontend. It feels a risk to impact the stats when we update the config for building govuk-frontend. Do the stats need much more in terms of config than having the visualiser plugin?

packages/govuk-frontend/tasks/scripts.mjs
```diff
@@ -48,5 +48,8 @@ export const compile = (options) => gulp.series(
       srcPath: join(options.srcPath, 'govuk-prototype-kit'),
       destPath: resolve(options.destPath, '../') // Top level (not dist) for compatibility
     })
-  )
+  ),
+
+  // Compile GOV.UK Frontend build stats
+  npm.script('build:stats', [], options)
```
romaricpascal, line 54: Why does govuk-frontend need to run the stats itself? I was picturing the stats as completely independent and us coordinating its run in a similar  way we coordinate the run of the review app.

shared/stats/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "private": true,
+  "name": "govuk-frontend-stats",
+  "description": "GOV.UK Frontend build stats",
+  "main": "src/index.mjs",
```
romaricpascal, line 5: Really neat way to get the stats in the review app 🙌🏻

packages/govuk-frontend-review/src/common/nunjucks/filters/highlight.mjs
```diff
@@ -3,10 +3,10 @@ import hljs from 'highlight.js'
 /**
  * Format code with syntax highlighting
  *
- * @param {string} code - Code in plain text
+ * @param {string | undefined} code - Code in plain text
  * @param {string} [language] - Code programming language
  * @returns {string} Code with syntax highlighting
  */
-export function highlight (code, language) {
+export function highlight (code = '', language) {
```
romaricpascal, line 10: Do JS arguments use their defaults when passing `undefined`? 🤯 That's a bit arcane and it may be clearer to pass `(code ?? '').trim()` to `hljs.highlight` rather than lean on people knowing that bit of the specs 😊

shared/stats/rollup.config.mjs
colinrotherham: The UMD build is hugely different as it's an all-in-one bundle

I want to replace it with an ESM bundle though (rather than our default `{ preserveModules: true }`)

packages/govuk-frontend/tasks/scripts.mjs
colinrotherham: It's what keeps the review app stats up-to-date as you make changes

Otherwise they're only generated at the start of the session

It was an easy fix versus moving to `npx rollup --watch`

packages/govuk-frontend-review/src/common/nunjucks/filters/highlight.mjs
colinrotherham: Yeah both the param value `undefined` or optional attribute will default to `''`

It's a free pass for `undefined` that `null` doesn't get

Happy to swap it over!

shared/stats/rollup.config.mjs
colinrotherham: Ahh I figured this would reduce risk?

For example when we add `@rollup/plugin-babel` to **rollup.esm.config.mjs** the stats will change

In the PR preview deployment we'll instantly be able to compare them

shared/stats/rollup.config.mjs
romaricpascal: Makes sense. I think I'd lean towards updating the array of configs rather than multiply `npx` calls as that may be easier to review and maintain as we add more components 😊

packages/govuk-frontend/tasks/scripts.mjs
romaricpascal: Oh, I see. I thought we discussed that `watch` wasn't in scope, but that looks like a very cheap option to achieve it 🙌🏻

shared/stats/rollup.config.mjs
romaricpascal: Wouldn't we be running `babel` twice if that's the case: once to build `govuk-frontend/dist` and then that one going through `babel` again when building the stats, which is not what we're looking for?

packages/govuk-frontend-review/src/common/nunjucks/filters/highlight.mjs
romaricpascal: Yeah, feels definitely safer in the long run to swap rather than rely on intricate knowledge of the specs to understand what's going on. Cheers for confirming the behaviour 😊

shared/stats/rollup.config.mjs
colinrotherham: Each file is a custom bundle that can be imported standalone

Versus modules where import/require resolution is still needed

shared/stats/rollup.config.mjs
colinrotherham: Nah I'd rather we only run it once. It won't run on `node_modules` unless configured to

packages/govuk-frontend/tasks/scripts.mjs
colinrotherham: Yeah exactly, hope that explains it—bonus feature!

shared/stats/rollup.config.mjs
romaricpascal: Yeah, I'd rather it only runs once when we build `govuk-frontend` as well and sounds like by default it'll only run like that. That's just for Babel though, other plugins will likely work at all times. The reason I'm advocating for not reusing the `govuk-frontend` configs there it to avoid risks of a change in `govuk-frontend` configuration changing our stats computation. 

Framing the question differently: what do we gain by re-using the `govuk-frontend` rollup configs?

shared/stats/rollup.config.mjs
romaricpascal: There's something I must be misunderstanding there. Let's have a call tomorrow to clarify 😊

shared/stats/rollup.config.mjs
colinrotherham: Just went for the fastest option that didn't duplicate code. These configs are all ready to go

Shall we hold off setting up more configs until it becomes an issue?

shared/stats/rollup.config.mjs
colinrotherham: Yeah sounds good 😊

I'll add some background reading, compare these two:

```
scripts.compile('**/!(*.test).mjs', options) // compile:js
scripts.compile('!(*.test).mjs', options) // compile:mjs
```

See the globs?

For `compile:js` Rollup processes every file (recursively) into a separate UMD bundle
For `compile:mjs` Rollup processes only the entry points into ES modules

Which is why when you import multiple UMD files you get the polyfills multiple times

```js
require('govuk-frontend/dist/govuk/components/accordion/accordion')
require('govuk-frontend/dist/govuk/components/character-count/character-count')
require('govuk-frontend/dist/govuk/components/character-count/checkboxes')
```

If you change the format to `cjs` and enable `{ preserveModules }` you'll see the difference

shared/stats/rollup.config.mjs
colinrotherham: All sorted, I've pushed up an update with [multiple entry examples](https://github.com/alphagov/govuk-frontend/tree/performance-stats/shared/stats/src/javascripts) similar to [your spike options](https://github.com/alphagov/govuk-frontend/pull/3239/files#diff-6b7a7f6278023134eca8e8fa3d757c01528122c05cb8702e834ad9b2ff688a32)

Had to turn off `{ preserveModules }` (now bundling by default) and use `--file` not `--dir` but it works

packages/govuk-frontend-review/src/common/nunjucks/filters/highlight.mjs
colinrotherham: I've removed this edit and added a new `inspect` filter to format JavaScript objects before highlighting

```njk
{{- stats | inspect | highlight("js") | safe if stats else "No build stats available" -}}
```

colinrotherham:
@romaricpascal Pushed up a fairly awful design but shows what's possible 😊
https://govuk-frontend-pr-3681.herokuapp.com

Component names with file sizes

Need to show the different total bundle sizes too

shared/stats/rollup.config.mjs
colinrotherham: ✅ UMD stats build has now been removed

shared/stats/rollup.config.mjs
romaricpascal: I think in this case we do need the duplication to make sure things are independent. By sharing the configs, we're mixing, in the computation of stats, stuff that we run for the build of the package itself (like the replacement of the version). While it doesn't do affect it much for now, that's a risk in the future to double up some transformations without noticing or maybe make multiple outputs for our build (which I'm not sure how the stats plugin will cope with, thinking of it).

At the end of the day, the only part we're getting from that sharing is the config, so I'd rather we start with a clean separation from the get go there. I think this will also avoid the need for updating the current Rollup configs as well which is that less changes 😊

romaricpascal:
That list is starting to look neat! Nice one! 🙌🏻

shared/stats/rollup.config.mjs
colinrotherham: Pushed up an update. All configs are now standalone

romaricpascal (approved):
Works super neatly! Really nice one! 🙌🏻 Thanks for making the updates from our discussions, as well as setting a really tidy layout on the review app 😊 

It's not even too much of a pain to run it against an older version and see all the good removing polyfills brought by:

- updating `govuk-frontend`’s package version to 5.0 so it’s different from the 4.6.0 release (I’d advocate for updating the timing we change our version numbers)
- installing `govuk-frontend` 4.6.0 explicitely in the stats package
- updating the path to the modules in the stats rollup config to remove `dist` (we may want to work some magic there).

<details>
<summary>v4.6.0</summary>
Screenshot 2023-06-02 at 12 38 39
</details>


<details>
<summary>PR branch</summary>
Screenshot 2023-06-02 at 11 26 38
</details>

PR #3683, opened 2023-05-24
Silence markdown deprecation warnings

domoscargin:
Since we bumped marked to v5.0.0, we've been getting deprecation warnings:

```
marked(): headerIds and headerPrefix parameters enabled by default, but are deprecated since version 5.0.0, and will be removed in the future. To clear this warning, install  https://www.npmjs.com/package/marked-gfm-heading-id, or disable by setting `{headerIds: false}`.

marked(): mangle parameter is enabled by default, but is deprecated since version 5.0.0, and will be removed in the future. To clear this warning, install https://www.npmjs.com/package/marked-mangle, or disable by setting `{mangle: false}`.
```

We use marked in one place: on our [full page examples index page](https://govuk-frontend-review.herokuapp.com/full-page-examples), to render the "scenarios" under each heading.

We don't mangle anything, and we don't render any headings that might be affected by `headerIds` and `headerPrefix`.

Nevertheless, I've followed `marked`'s advice and installed the `marked-gfm-heading-id` plugin, to maintain expected behaviour if we include headings in the future.

There is no diff in the rendered page.

This introduces the small chore of removing the explicit `mangle: false` once that property defaults to `false`. But since it'll just keep working in that instance anyway, I think not a particularly bad temporary fix.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -59,6 +59,7 @@
         "@types/gulp": "^4.0.10",
         "@types/jest": "^29.5.1",
         "@types/jest-axe": "^3.5.5",
+        "@types/marked": "^5.0.0",
         "@types/node": "^20.2.3",
         "@types/nunjucks": "^3.2.2",
         "puppeteer": "^20.3.0"
@@ -4530,6 +4531,12 @@
       "dependencies": {
         "@types/node": "*"
       }
+    },
+    "node_modules/@types/marked": {
+      "version": "5.0.0",
+      "resolved": "https://registry.npmjs.org/@types/marked/-/marked-5.0.0.tgz",
+      "integrity": "sha512-YcZe50jhltsCq7rc9MNZC/4QB/OnA2Pd6hrOSTOFajtabN+38slqgDDCeE/0F83SjkKBQcsZUj7VLWR0H5cKRA==",
+      "optional": true
     },
     "node_modules/@types/mime": {
       "version": "3.0.1",
--- a/packages/govuk-frontend-review/src/common/nunjucks/globals/markdown.mjs
+++ b/packages/govuk-frontend-review/src/common/nunjucks/globals/markdown.mjs
@@ -7,5 +7,9 @@ import { marked } from 'marked'
  * @returns {string} Rendered Markdown
  */
 export function markdown (content) {
-  return marked.parse(content)
+  // Explicitly set deprecated properties to false to avoid deprecation warnings
+  return marked.parse(content, {
+    mangle: false,
+    headerIds: false
+  })
 }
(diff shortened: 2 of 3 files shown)


colinrotherham (approved):
Thanks @domoscargin 🎉

Happy to approve but got a couple of suggestions??

First one, can we let `tsc` detect when the deprecated properties are removed? It'll be useful to see this happen in a future Dependabot 🤖 update [by adding types](https://github.com/markedjs/marked/pull/2577):

```console
npm install @types/marked --save-optional
```

But then the type made me realise `marked.parsed()` has a 2nd param 😮

```js
return marked.parse(content, {
  mangle: false
})
```

If that's shorter and sweeter?

packages/govuk-frontend-review/src/common/nunjucks/globals/markdown.mjs
```diff
@@ -7,5 +8,14 @@ import { marked } from 'marked'
...
  */
 export function markdown (content) {
+  // mangle is enabled by default, but deprecated since marked 5.0.0.
+  // Explicitly set it to false to avoid deprecation warnings
+  marked.use({
+    mangle: false
+  })
+  // headerIds and headerPrefix are enabled by default, but deprecated since
+  // marked 5.0.0. Use recommended plugin instead
+  marked.use(gfmHeadingId())
```
colinrotherham, line 18: Sudden thought. Do we actually have any headings in `example.scenario` fields?

Maybe we can get away with `{ headerIds: false }`

domoscargin: Though it does feel nicer to not install a needless plugin!

domoscargin:
Thanks for the suggestions, @colinrotherham 

I've gone with the non-plugin option to disable `headerId` and added the types as insurance.

packages/govuk-frontend-review/package.json
```diff
@@ -51,5 +51,8 @@
     "slash": "^5.1.0",
     "supertest": "^6.3.3",
     "typedoc": "^0.24.7"
+  },
+  "optionalDependencies": {
+    "@types/marked": "^5.0.0"
```
colinrotherham, line 56: Thanks @domoscargin

Would you mind moving this one to the project **package.json** for now?

Seems to work more reliably, especially for dependencies of dependencies

We can move `@types/*` packages into relevant workspaces in future, but probably need [`typeRoots`](https://www.typescriptlang.org/tsconfig#typeRoots)

package-lock.json
```diff
@@ -14,6 +14,9 @@
         "packages/govuk-frontend-review",
         "docs/examples/*"
       ],
+      "dependencies": {
+        "@types/marked": "*"
```
colinrotherham, line 18: Sorry @domoscargin can you run `npm install` again?

It's left a fragment of a previous change in the **package-lock.json**

Installing again should do a tidy up

domoscargin: Done!

Issue #3506, opened 2023-04-14
Migrate JavaScript to ES2015 classes


romaricpascal:
## What

We're updating our components to use JavaScript classes rather than constructor functions, and internally to use `let` and `const` instead of `var`. The methods currently set on the component's `prototype` will become actual class methods. All but the `init` one will be annotated `@private` in their JSDoc, though.

> **Warning**: This issue is limited to migrating our components to using classes. **We should avoid any further refactoring** to help the migration. As we're defining a clear public API for the components, we'll be able to refactor and improve in future minor versions.

> **Note**: Some IDEs provide code intelligence features that might be able to automate a lot of this. For example, in VS Code, you can hit Cmd + . with the cursor on the constructor and it should allow you to 'Convert function to ES2015 class'.

## Why

Migrating to classes allows for a more idiomatic and compact syntax, as well as opening the door to factoring shared components into a base class to be inherited by the components

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

```[tasklist]
### Components have been migrated to modern syntax (ES classes, `let`/`const`)
1. [ ] Accordion
2. [ ] Button
3. [ ] Character Count
4. [ ] Checkboxes
5. [ ] Details
6. [ ] Error Summary
7. [ ] Header
8. [ ] Notification Banner
9. [ ] Radios
10. [ ] Skip Link
11. [ ] Tabs
```

Issue #3463, opened 2023-03-31
Upgrade ESLint configuration


romaricpascal:
## What

Update our ESLint config to reflect the way that we write JavaScript now that we no longer support IE8-11.

For example, re-enable the no-var rule and update the rules for the es-x plugin.

Fix any linting errors that occur. For example, replace var with let or const as appropriate.

## Why

With our new strategy for browser support and us dropping IE, we can write more modern JavaScript

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [ ] ESLint configuration has been updated to match how we write JavaScript
- [ ] JavaScript errors have been fixed

PR #3698, opened 2023-05-26
Enable ESLint + Stylelint for code inside Markdown

colinrotherham:
Noticed some of the code blocks in our documentation had drifted a bit in terms of code style

But knowing we've got these two coming up:

* https://github.com/alphagov/govuk-frontend/issues/3506
* https://github.com/alphagov/govuk-frontend/issues/3463

I've added ESLint [eslint-plugin-markdown](https://www.npmjs.com/package/eslint-plugin-markdown) to check code blocks in `*.md` files get migrated too

Similarly Stylelint can use [postcss-markdown](https://www.npmjs.com/package/postcss-markdown) to check Sass code blocks


.eslintrc.js
```diff
@@ -18,7 +18,10 @@ module.exports = {
         'plugin:n/recommended',
         'plugin:promise/recommended'
       ],
-      files: ['**/*.{cjs,js,mjs}'],
+      files: [
+        '**/*.{cjs,js,mjs}',
+        '**/*.md/*.{cjs,js,mjs}'
```
romaricpascal, line 23: **question**: That's an odd glob with `*.md` being a "directory" when they're actually files. What does it represent? Is it to match the codeblocks within markdown?

.lintstagedrc.js
```diff
@@ -11,7 +11,7 @@ module.exports = {
   // as recommended by lint-staged.
   //
   // https://github.com/okonet/lint-staged#how-can-i-ignore-files-from-eslintignore
-  '*.{cjs,js,mjs}': filterTask('npm run lint:js:cli -- --fix'),
+  '*.{cjs,js,md,mjs}': filterTask('npm run lint:js:cli -- --fix'),
   '*.{json,md,yaml,yml}': 'npm run lint:prettier:cli -- --write',
   '*.scss': 'npm run lint:scss:cli -- --fix --allow-empty-input'
```
romaricpascal, line 16: **question**: Are we missing the `.md` extension here to automatically lint Markdown on commit?

romaricpascal (commented):
Looks neat and useful going forward to keep our docs tidy (might be worth seeing if it's useful on the Design System site as well). There are a few things that make me a bit wary to merge it as is for now:

1. Updating the CHANGELOG, which has been published already. I've commented on two bits that have quite heavy edits, but I'm also worried about the linters automatically updating code (say turning `var` to `const`) on older entries and us having to track what to enable and disable. I'm wondering if it's worth leaving the CHANGELOG out of linting to avoid changes sneaking in without us noticing 🤔 
2. I'm not sure I understand the change to the JSDoc linting, would you be able to explain it a little more (either in the commit message or on this PR) please?
3. We need to handle lint-staged running tasks concurrently to avoid Prettier possibly overriding the formatting done by the linters (unless it is compatible)

.eslintrc.js
colinrotherham: Yeah it matches all `cjs`, `js` and `mjs` blocks in `*.md` files
https://github.com/eslint/eslint-plugin-markdown#advanced-configuration

.lintstagedrc.js
colinrotherham: Yeah we were indeed. I've put an updated example in https://github.com/alphagov/govuk-frontend/pull/3698#discussion_r1213530634

CHANGELOG.md
```diff
@@ -3303,16 +3303,28 @@ If you're using your own components that rely on the overflow state of the docum
...
+    margin-bottom: $govuk-spacing-scale-1;
+    margin-bottom: $govuk-spacing-scale-2;
+    margin-bottom: $govuk-spacing-scale-3;
+    margin-bottom: $govuk-spacing-scale-4;
+    margin-bottom: $govuk-spacing-scale-5;
+    margin-bottom: $govuk-spacing-scale-6;
+    margin-bottom: $govuk-spacing-scale-7;
+    margin-bottom: $govuk-spacing-scale-8;
+    margin-bottom: $govuk-spacing-scale-9;
+  }
```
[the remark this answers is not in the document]
colinrotherham: That makes sense. I've ignored **CHANGELOG.md** so we don't change history

In future we may want to enable it again but add [ESLint skip comments](https://github.com/eslint/eslint-plugin-markdown#skipping-blocks):

```html
<!-- eslint-skip -->
```

CHANGELOG.md
```diff
@@ -978,10 +980,10 @@ We've changed the naming of our components' JavaScript modules so individual imp
 You can now import multiple modules without overwriting the previous one, for example:
 
 ```js
-//= require govuk/components/accordion/accordion.js
-//= require govuk/components/button/button.js
+require('govuk-frontend/govuk/components/accordion/accordion')
+require('govuk-frontend/govuk/components/button/button')
 
-# These modules are available under window.GOVUKFrontend.Accordion and window.GOVUKFrontend.Button respectively
+// These modules are available under window.GOVUKFrontend.Accordion and window.GOVUKFrontend.Button respectively
```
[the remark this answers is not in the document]
colinrotherham: Don't worry, I've ignored **CHANGELOG.md** and put an idea in https://github.com/alphagov/govuk-frontend/pull/3698#discussion_r1213532508

colinrotherham:
Thanks @romaricpascal

Ready for review again now. Answers below:

> 1. Updating the CHANGELOG, which has been published already. I've commented on two bits that have quite heavy edits, but I'm also worried about the linters automatically updating code (say turning `var` to `const`) on older entries and us having to track what to enable and disable. I'm wondering if it's worth leaving the CHANGELOG out of linting to avoid changes sneaking in without us noticing 🤔

Yeah course. I've ignored **CHANGELOG.md** from both `.eslintrc.js` and `stylelint.config.js`

> 2. I'm not sure I understand the change to the JSDoc linting, would you be able to explain it a little more (either in the commit message or on this PR) please?

JSDoc linting is now the default so we can include Markdown code blocks too

Where previously only [`'src/govuk/**/*.mjs'` files required JSDoc](https://github.com/alphagov/govuk-frontend/blob/5e2bcd1292b91cf3b6bc6d02c22fb4f79b2d75d4/packages/govuk-frontend/.eslintrc.js#L40)

BUT you'll see I've carefully disabled it in all the places we don't need it with:

**.eslintrc.js**
```js
module.exports = {
  rules: {
    // JSDoc blocks and descriptions are optional
    'jsdoc/require-jsdoc': 'off',
    'jsdoc/require-param-description': 'off'
  }
}
```

> 3. We need to handle lint-staged running tasks concurrently to avoid Prettier possibly overriding the formatting done by the linters (unless it is compatible)

Good thinking, I've set up a [separate array of commands](https://github.com/alphagov/govuk-frontend/pull/3698#discussion_r1213530634) for Markdown:

```js
module.exports = {
  '*.{cjs,js,mjs}': commands.eslint,
  '*.{json,yaml,yml}': commands.prettier,
  '*.md': [commands.eslint, commands.stylelint, commands.prettier],
  '*.scss': commands.stylelint
}
```

CHANGELOG.md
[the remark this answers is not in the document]
romaricpascal: Ace thanks! 

Regarding linting the changelog and skipping, I thought of that, but I'm not sure how that would work in the long run. 

We'd want to only lint the changelog bits that are for the next release and keep any examples from previous release as published. That would mean adding `<-- eslint-skip -->` comments at release time, which feels quite cumbersome/easy to forget. It's something we could ultimately automate with something like [remark](https://remark.js.org/), but I'm not sure it's a massive gain compared to the build/maintenance required.

.eslintrc.js
romaricpascal: Could we add a little comment in the config to link us there so everyone can quickly find about it, please?
```suggestion
        '**/*.md/*.{cjs,js,mjs}' // https://github.com/eslint/eslint-plugin-markdown#advanced-configuration
```

CHANGELOG.md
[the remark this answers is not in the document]
colinrotherham: As long as you're still happy to skip it entirely? Cheers

.eslintrc.js
colinrotherham: Are there any other plugin-related rules or settings that should have comments too?

Aware there's probably quite a few things that don't make immediate sense

romaricpascal (approved):
I think this can go in this new state (maybe after updating the `docs` config to be more generic). We can always adjust/amend configurations if it becomes cumbersome to have the Markdown linted (eg. if we have too many adjustments of the config to make for the Markdown).

Centralising the JSDoc linting config highlights (with the `.eslintrc` files) all the places we should review and update our JSDoc to avoid overriding. This can happen in its own time, I'll add an issue for it.

docs/.eslintrc.js
```diff
@@ -0,0 +1,11 @@
+module.exports = {
+  overrides: [
+    {
+      files: ['**/coding-standards/js.md/*.{cjs,js,mjs}'],
```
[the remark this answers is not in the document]
colinrotherham: I've dropped the commit to require JSDoc in Markdown code blocks

We can come back and review it another time

docs/contributing/coding-standards/css.md
```diff
@@ -167,144 +167,124 @@ Good:
...
 Good:
 
-```
+```scss
 .govuk-wrapper {
-  ...
-}
-```
-
-### Use single colons for pseudo-element selectors
```
colinrotherham, line 198: Removed this guidance as it's no longer necessary since https://github.com/alphagov/govuk-frontend/commit/f9c7ff0cfa5466076870275827d96e0554e4337d

Issue #2506, opened 2022-01-19
Remove polyfills for IE


vanitabarrett:
## What
At the moment, we ship polyfills to all browsers even when they're not needed. We think around 50% of the JavaScript we ship is polyfills (this might say more about the amount of JavaScript we have, rather than the size of the polyfills).

Having to include polyfills also means we tend to restrict ourselves to features that we know are either 1) well supported, or 2) don't require large polyfills to work in older browsers.

## Why
There is a performance impact if we're shipping unnecessary JavaScript to browsers that don't need it.
Our current approach puts some restrictions on the JavaScript features we use. This isn't necessarily causing problems right now, but might do in future when we come to build more complex components. (also depends on https://github.com/alphagov/govuk-frontend/issues/2503 )

## Who needs to work on this
Developers

## Who needs to review this
Developers

## Done when

Depends on our [browser support proposal](https://github.com/alphagov/govuk-frontend/issues/2519) being approved by the community. If we go down the route of dropping IE support to 'functional':

- [x] https://github.com/alphagov/govuk-frontend/pull/3570
- [x] https://github.com/alphagov/govuk-frontend/pull/3720
- [x] https://github.com/alphagov/govuk-frontend/pull/3723
- [x] Update/remove [polyfill documentation](https://github.com/alphagov/govuk-frontend/blob/main/docs/contributing/polyfilling.md) and pick up https://github.com/alphagov/govuk-frontend/issues/2159 as part of it (making sure it meets style guide etc)
- [x] If we remove all polyfills and won't be adding any in future, we can close https://github.com/alphagov/govuk-frontend/issues/674

NickColley:
On a project I removed the polyfills in the build step, sharing here if anyone is Googling finds this issue and wants to do the same:

Rollup Plugin: 
```javascript
// file: rollup-remove-polyfills.js
// Based on https://github.com/proteriax/rollup-plugin-ignore

// Replace polyfills with empty files that'll get removed with tree shaking.
function removePolyfills() {
  const emptyFile = "export default {}";
  const emptyFileName = "\0rollup_plugin_ignore_empty_module_placeholder";
  return {
    resolveId(importee) {
      return importee.includes("vendor/polyfills") ? emptyFileName : null;
    },
    load(id) {
      return id === emptyFileName ? emptyFile : null;
    },
  };
}

module.exports = removePolyfills;
```

```javascript
// file: index.mjs
[...]

PR #3720, opened 2023-05-30
Remove IE11 vendor polyfills

domoscargin:
Partially addresses #2506


domoscargin:
A few thoughts:

## Intl API
There's some mentions of IE11 in the i18n's check for [Plural Rules](https://github.com/alphagov/govuk-frontend/blob/2c96f0e8d72f4d6332b0f9f75268920baa651488/packages/govuk-frontend/src/govuk/i18n.mjs#L119-L120) and [Number Format](https://github.com/alphagov/govuk-frontend/blob/2c96f0e8d72f4d6332b0f9f75268920baa651488/packages/govuk-frontend/src/govuk/i18n.mjs#L133-L134) support. 

Can we simplify these checks to just look for `.supportedLocalesOf` and assume we're good on `Intl` and `PluralRules/NumberFormat`? Or do we need to keep these checks as is? @querkmachine - you probably have thoughts?

## Bits and bobs
There's a coupla things that come up when searching for "IE" (stuff like [IE rendering in the template](https://github.com/alphagov/govuk-frontend/blob/2c96f0e8d72f4d6332b0f9f75268920baa651488/packages/govuk-frontend/src/govuk/template.njk#L11-L12), potentially some [html-validate rules](https://github.com/alphagov/govuk-frontend/blob/2c96f0e8d72f4d6332b0f9f75268920baa651488/packages/govuk-frontend/src/govuk/components/components.template.test.js#L61-L63), some specific fixes for components, etc.

Not sure we have anywhere to start logging these things to tackle as we go about the work of updating things?

colinrotherham (approved):

domoscargin:
@colinrotherham I've updated polyfilling documentation and linked to #3722

colinrotherham:
> @colinrotherham I've updated polyfilling documentation and linked to #3722

Nice, thanks @domoscargin. Keeps the `git blame` history going

PR #3723, opened 2023-05-31
Remove checks for newer JavaScript features

domoscargin:
## `onpageshow`
[All ES6 module supporting browsers support PageTransitionEvents](https://browsersl.ist/#q=supports+es6-module+and+not+supports+page-transition-events), so we can remove the check for `window.onpageshow`.

Potentially we can go further and revert to checking `document.readyState`, which seems to have been affected by a bug in IE9-10, but this feels cleaner.

## `window.matchMedia`
[All ES6 module supporting browsers support window.matchMedia](https://caniuse.com/matchmedia)

We can probably set up the responsive checks as part of `Tabs.prototype.setup` now, but I didn't do that as I didn't want to mess with the deprecation step of setting those methods to private in v5.0

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/character-count/character-count.mjs
+++ b/packages/govuk-frontend/src/govuk/components/character-count/character-count.mjs
@@ -195,11 +195,8 @@ CharacterCount.prototype.init = function () {
   // When the page is restored after navigating 'back' in some browsers the
   // state of the character count is not restored until *after* the
   // DOMContentLoaded event is fired, so we need to manually update it after the
-  // pageshow event in browsers that support it.
-  window.addEventListener(
-    'onpageshow' in window ? 'pageshow' : 'DOMContentLoaded',
-    this.updateCountMessage.bind(this)
-  )
+  // pageshow event.
+  window.addEventListener('pageshow', this.updateCountMessage.bind(this))
 
   this.updateCountMessage()
 }
--- a/packages/govuk-frontend/src/govuk/components/tabs/tabs.mjs
+++ b/packages/govuk-frontend/src/govuk/components/tabs/tabs.mjs
@@ -51,11 +51,7 @@ Tabs.prototype.init = function () {
     return
   }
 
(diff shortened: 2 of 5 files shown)


packages/govuk-frontend/src/govuk/components/character-count/character-count.mjs
```diff
@@ -195,9 +195,9 @@ CharacterCount.prototype.init = function () {
   // When the page is restored after navigating 'back' in some browsers the
   // state of the character count is not restored until *after* the
   // DOMContentLoaded event is fired, so we need to manually update it after the
-  // pageshow event in browsers that support it.
+  // pageshow event.
   window.addEventListener(
```
colinrotherham, line 199: Not a blocker, but these can all go on one line now?

```suggestion
  window.addEventListener('pageshow', this.updateCountMessage.bind(this))
```

packages/govuk-frontend/src/govuk/components/tabs/tabs.mjs
```diff
@@ -51,11 +51,7 @@ Tabs.prototype.init = function () {
     return
   }
 
-  if (typeof window.matchMedia === 'function') {
-    this.setupResponsiveChecks()
-  } else {
-    this.setup()
-  }
+  this.setupResponsiveChecks()
```
colinrotherham, line 54: Another non-blocker, but shall we delete `setupResponsiveChecks()` and move its code into `init()`?

There's only one "init path" through the code now, like the other components

colinrotherham (approved):

packages/govuk-frontend/src/govuk/components/tabs/tabs.mjs
domoscargin: I decided not to do that because we're saying these methods will go private in v5, so it felt like moving 'em outright is either a "later" thing, or one for discussion when we review the code of the relevant components.

PR #3724, opened 2023-05-31
Add shared `getYaml()` lib function

colinrotherham:
This PR adds a new `getYaml()` function to read Rollup performance data, split out from:

* https://github.com/alphagov/govuk-frontend/pull/3681

Now used by `generateFixture()` and `generateMacroOption()` which both read `${componentName}.yaml` files

```js
const json = await getYaml(join(options.srcPath, componentDataPath))
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -27529,7 +27529,6 @@
         "govuk-frontend-lib": "*",
         "gulp": "^4.0.2",
         "gulp-cli": "^2.3.0",
-        "js-yaml": "^4.1.0",
         "nunjucks": "^3.2.4",
         "plugin-error": "^2.0.1",
         "postcss": "^8.4.23",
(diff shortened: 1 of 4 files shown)


domoscargin (approved):
Looks good, thanks (and TIL about `Pick`)

colinrotherham:
Thanks @domoscargin

For now, I'm happy our component tests all call `getExamples()` → `getComponentData()` → **`getYaml()`**

```js
let examples

beforeAll(async () => {
  examples = await getExamples('accordion')
})
```

Definitely should start dropping in more tests

For reference, we still have missing coverage for the Design System JSON files:

>govuk/components/${componentName}/fixtures.json
>govuk/components/${componentName}/macro-options.json

But we'd likely be better [consolidating library code](https://github.com/alphagov/govuk-design-system/blob/a51471521ecab03a8774bee2eefe506bbf809b7d/lib/get-macro-options/index.js#LL12C47-L12C47) and tests between the two projects

PR #3725, opened 2023-05-31
Add default “npm run build” for project (with docs updates)

colinrotherham:
Our **tasks.md** docs talk about serving the review app directly with:

```shell
npm run serve --workspace govuk-frontend-review
```

But we didn't explain that an MVB™ (minimum viable build 😆) was needed first:

- build GOV.UK Frontend using `npm run build:package`
- build Rollup stats YAML using `npm run build:stats` in [#3681](https://github.com/alphagov/govuk-frontend/pull/3681)
- build Express.js "Review app" using `npm run build:app`

This PR adds a convenient `npm run build` script to solve that


package.json
```diff
@@ -20,17 +20,18 @@
     "build-release": "./bin/build-release.sh",
     "publish-release": "./bin/publish-release.sh",
     "pre-release": "./bin/pre-release.sh",
-    "predev": "npm run build --workspace govuk-frontend",
+    "predev": "npm run build",
     "dev": "concurrently \"npm run dev --workspace govuk-frontend-review\" \"npm run dev --workspace govuk-frontend\" --kill-others --names \"app,pkg\" --prefix-colors \"red.dim,blue.dim\"",
+    "build": "npm run build --workspace govuk-frontend --workspace govuk-frontend-review",
```
colinrotherham, line 25: Our default build is to prepare the review app for launch

1. Build `govuk-frontend` package
2. Build `govuk-frontend-review` assets

But notice that the webpack example isn't included? For [GitHub Actions **Tests**](https://github.com/alphagov/govuk-frontend/actions/workflows/tests.yml) we instead run:

https://github.com/alphagov/govuk-frontend/blob/01169199015a28fe9779bb814db36bc6f3bb1139/.github/workflows/actions/build/action.yml#L23

By adding a new `npm run build` we can run it before key scripts

E.g. Ensuring we have the `govuk-frontend` package built before we import it

romaricpascal (approved):
Looks good, especially in anticipation of building the stats. Just a little suggestion to reuse the existing tasks for avoiding future divergence, but I'm fine if it goes as is 😊

package.json
colinrotherham: @romaricpascal Ah we can't I'm afraid, review app relies on `govuk-frontend/dist/**` things

Any thoughts on using `npm run build` (default) versus `npm run build:package` (named)?

colinrotherham: Let's narrow it down when we add the stats package (also needs pre-building) in:

* https://github.com/alphagov/govuk-frontend/pull/3681

PR #3727, opened 2023-06-01
Add shared `getComponentNames()` and other lib functions

colinrotherham:
For the performance stats in https://github.com/alphagov/govuk-frontend/pull/3681 I've split out some library functions for filtering components with JavaScript

### Component names
All components by their directory names

```mjs
await getComponentNames()
```

### Component names
Only JavaScript-enabled components by their directory names

```mjs
await getComponentNames((componentName, componentFiles) =>
  componentFiles.every(filterPath([`**/${componentName}.mjs`])))
```

### Component files
All component files by directory listing (files only)

```mjs
await getComponentFiles()
```

### Review app library functions

I've also split out library functions with [`front-matter`](https://www.npmjs.com/package/front-matter) that weren't shared:

```mjs
await getExampleNames()
await getFullPageExamples()
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -27317,6 +27317,7 @@
         "cookie-parser": "^1.4.6",
         "express": "^4.18.2",
         "express-validator": "^7.0.1",
+        "front-matter": "^4.0.2",
         "govuk-frontend": "*",
         "govuk-frontend-config": "*",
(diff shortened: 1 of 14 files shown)


romaricpascal (approved):
Nice to have clear name on these calls and that non-shared code got extracted 🙌🏻 ⛵

PR #3730, opened 2023-06-01
Split Rollup plugins into input and output etc

colinrotherham:
Following some code reviews on https://github.com/alphagov/govuk-frontend/pull/3681 this PR splits out a few Rollup changes:

1. **Each Rollup config should be standalone**
We no longer extend configs to reduce risk from accidental edits

2. **Ensure Rollup task preserves config `file` or `dir` options**
Removes surprises when editing configs or using [`--file`](https://rollupjs.org/configuration-options/#output-file) and [`--dir`](https://rollupjs.org/configuration-options/#output-dir) CLI flags

3. **Ensure Rollup task matches extension to output format**
Automatically output `*.js` by default or `*.mjs` for ES modules

4. **Split Rollup plugins into input (bundling) and output (generating)**
Matches what we did in [rollup.config.js](https://github.com/alphagov/govuk-design-system/blob/main/rollup.config.js) on the Design System

I've also clarified our docs for ECMAScript (ES) modules versus UMD bundles

### DIFF
--- a/docs/contributing/coding-standards/js.md
+++ b/docs/contributing/coding-standards/js.md
@@ -124,7 +124,7 @@ var myExample = new Example()
 
 ## Modules
 
-Use ECMAScript modules (`import`/`export`) over CommonJS and other formats. You can always transpile to your preferred module system.
+Use ECMAScript (ES) modules (`import`/`export`) over CommonJS and other formats. You can always transpile to your preferred module system.
 
 ```js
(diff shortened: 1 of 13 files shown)


romaricpascal (approved):
Neat tidy-up ahead of future config changes 🙌🏻 And thanks for the thorough editing of our docs to make our mentions of modules more precise 😊

PR #3732, opened 2023-06-04
Update references from back link to inset text in inset text options list

matteason:
Just fixes a couple of copy/paste errors that have resulted in the inset text guidance on the design system containing references to the back link component by mistake


domoscargin (approved):
Nice spot!

Needs a second review.

colinrotherham (approved):

PR #3756, opened 2023-06-06
Allow globs in paths for `getListing()` and `getDirectories()`

colinrotherham:
This PR is a first step in making our directory listings work with `govuk-frontend@4` again

For example using `**/govuk` instead of hard coding `src/govuk` for gathering legacy stats

```mjs
await getListing('/path/to/govuk-frontend/**/govuk/components/accordion')
```

We can still pass in `option.cwd` where needed:

## Before

```mjs
await getListing(srcPath, pattern, {
  ignore: []
})
```

## After

```mjs
await getListing(pattern, {
  cwd: srcPath,
  ignore: []
})
```


romaricpascal (approved):
Neat! Good catch on the subtleties of how we handled these paths to allow us to keep loading v4 😊

Issue #3755, opened 2023-06-06
Error resolving GOV.UK Frontend subpath exports in Node.js 17+


romaricpascal:
## Description of the issue

We've identifying two scenarios when trying to `require()` or `import` GOV.UK Frontend

1. Node.js >= 17 errors with GOV.UK Frontend v4 [folder mapping in `"exports"`](https://nodejs.org/docs/latest-v18.x/api/deprecations.html#dep0148-folder-mappings-in-exports-trailing-)
2. Node.js <= 12.18 errors with GOV.UK Frontend v5 (unreleased) [subpath patterns in `"exports"`](https://nodejs.org/docs/latest-v18.x/api/packages.html#subpath-patterns)

```console
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './govuk/all.js' is not defined by "exports" in /Users/romaric.pascal/code/package-exports/node_modules/govuk-frontend/package.json
```

This can happen if extending our components and having a test suite run with a tool like [Mocha](https://mochajs.org/) that uses native Node `require` or `import` to load modules.

## Steps to reproduce the issue

1. Create a new prototype GOV.UK Frontend v4.6.0
    ```shell
    npx govuk-prototype-kit create package-exports
    cd package-exports
    ```
2. Run Node.js `require.resolve()` to locate GOV.UK Frontend's default exports
   ```shell
   node --eval "console.log(require.resolve('govuk-frontend'))" --conditions require
   node --eval "console.log(require.resolve('govuk-frontend'))" --conditions import
   ```

3. Run Node.js `require.resolve()` to locate a subpath export
   ```shell
   node --eval "console.log(require.resolve('govuk-frontend/govuk/all.js'))"
   ```

   Using Node.js 16 you'll see the following warning:
   
   ```console
   [DEP0148] DeprecationWarning: Use of deprecated folder mapping "./govuk/" in the "exports" field module resolution of the package at /node_modules/govuk-frontend/package.json.
   Update this package.json to use a subpath pattern like "./govuk/*"
   ```

   Using Node.js 17+ you'll see the following error:

   ```console
   Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './govuk/all.js' is not defined by "exports" in /node_modules/govuk-frontend/package.json
   ```

## Actual vs expected behaviour

Instead of receiving an error, the subpath export should should be resolved properly.

@colinrotherham suggested the following update to our exports to achieve so, using both :

```
[...]

PR #3757, opened 2023-06-06
Update package helpers to support `govuk-frontend@4`

colinrotherham:
This PR updates our `require.resolve()` package resolution to work with GOV.UK Frontend v4

As a workaround for https://github.com/alphagov/govuk-frontend/issues/3755 **packageTypeToPath()** can resolve paths via the `main` or `module` fields instead

## Rollup stats for old releases
Package stats can now be built for `govuk-frontend@4` as shown on branch [`package-resolver-v4`](https://github.com/alphagov/govuk-frontend/tree/package-resolver-v4)

```shell
npm version 5.0.0 --no-git-tag-version --workspace govuk-frontend
npm install govuk-frontend@4.6.0 --save --workspace govuk-frontend-stats
```

**Note:** It's necessary to bump our workspace `govuk-frontend` to v5.0.0 to avoid conflict

### DIFF
--- a/packages/govuk-frontend-review/rollup.config.mjs
+++ b/packages/govuk-frontend-review/rollup.config.mjs
@@ -1,5 +1,6 @@
 import resolve from '@rollup/plugin-node-resolve'
 import terser from '@rollup/plugin-terser'
+import { paths } from 'govuk-frontend-config'
 import { defineConfig } from 'rollup'
 
 /**
@@ -38,6 +39,6 @@ export default defineConfig(({ i: input }) => ({
    * Input plugins
    */
   plugins: [
-    resolve()
+    resolve({ rootDir: paths.app })
   ]
 }))
--- a/shared/lib/names.js
+++ b/shared/lib/names.js
@@ -1,4 +1,4 @@
-const { dirname, parse } = require('path')
+const { dirname, join, parse } = require('path')
 
 const { minimatch } = require('minimatch')
 
@@ -62,6 +62,23 @@ function componentPathToModuleName (componentPath) {
 }
 
 /**
+ * Resolve path to package entry from any npm workspace
+ *
+ * Once the package entry is resolved, the option `modulePath` can be used to
+ * append a new path relative to the package entry, for example `i18n.mjs`
+ *
+ * @param {string} packageEntry - Installed npm package entry, for example `govuk-frontend/src/govuk/all.mjs`
+ * @param {Pick<PackageOptions, "modulePath">} [options] - Package resolution options
+ * @returns {string} Path to installed npm package entry
+ */
+function packageResolveToPath (packageEntry, { modulePath } = {}) {
+  const packagePath = require.resolve(packageEntry)
+
+  // Append optional module path
+  return modulePath !== undefined ? join(dirname(packagePath), modulePath) : packagePath
+}
+
+/**
  * Resolve path to package from any npm workspace
  *
  * Used to find npm workspace packages that might be hoisted to
@@ -71,12 +88,21 @@ function componentPathToModuleName (componentPath) {
  * @returns {string} Path to installed npm package
  */
 function packageNameToPath (packageName) {
-  return dirname(require.resolve(`${packageName}/package.json`))
+  return packageResolveToPath(`${packageName}/package.json`, {
+    modulePath: ''
+  })
 }
 
 module.exports = {
   componentNameToClassName,
   componentNameToMacroName,
   componentPathToModuleName,
+  packageResolveToPath,
   packageNameToPath
 }
+
+/**
+ * @typedef {object} PackageOptions
+ * @property {string} [field] - Package field name from package.json, for example `module`
+ * @property {string} [modulePath] - Module path (optional, relative to package entry), for example `i18n.mjs`
+ */
--- a/shared/lib/names.js
+++ b/shared/lib/names.js
@@ -1,4 +1,4 @@
-const { dirname, join, parse } = require('path')
+const { dirname, parse } = require('path')
 
 const { minimatch } = require('minimatch')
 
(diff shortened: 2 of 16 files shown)


packages/govuk-frontend-review/rollup.config.mjs
```diff
@@ -39,6 +40,8 @@ export default defineConfig(({ i: input }) => ({
    * Input plugins
    */
   plugins: [
-    resolve()
+    resolve({
+      modulePaths: resolver(paths.app).modules
+    })
```
romaricpascal, line 45: **question** How come Rollup needs a nudge to look up in the workspace's `node_modules` first before jumping up to the root one?

shared/config/resolver.js
```diff
@@ -0,0 +1,50 @@
...
+// Repository root directory
+const paths = require('./paths')
+
+/**
+ * Default lookup resolver paths
+ *
+ * @param {string} basePath - Base directory, for example `/path/to/package`
+ * @returns {{ modules: string[], stylesheets: string[], views: string[] }} Resolver paths
+ */
+module.exports = (basePath) => {
```
romaricpascal, line 12: **question** Do I understand correctly that the need for that resolver function is because the `require` that we use to resolve in packageEntryToPath would resolve things from the point of view of the `shared/lib` workspace (and as such miss the local install of `govuk-frontend` in favour of the the local one hoisted at root)?

**suggestion** If that's the case, would it be less code for us to maintain (if at all possible) to have `packageEntryToPath` receive the "local" `require` function from the workspace it's being ran in? I'm a bit worried about us having to keep an eye on that list of paths?

romaricpascal (commented):
Nice to be able to pick our version of GOV.UK frontend and good path for resolving things like `import.meta.resolve` 🙌🏻 Just a couple of suggestions that I think would help us touch this area of the code more easily going forward (one for naming and one for the resolving)? No worries if they miss the mark completely though.

shared/lib/names.js
```diff
@@ -75,6 +75,26 @@ function packageEntryToPath (packageEntry, { modulePath } = {}) {
...
 }
 
+/**
+ * Resolve path to package field from any npm workspace
+ *
+ * @param {string} packageName - Installed npm package name
+ * @param {PackageOptions} [options] - Package resolution options
+ * @returns {string} Path to installed npm package field
+ */
+function packageFieldToPath (packageName, { modulePath, field = 'main' } = {}) {
```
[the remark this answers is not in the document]
colinrotherham: Makes sense, but can you suggest a better name than `resolve()`?

I've used `packageEntryToPath()` for packages that support `require.resolve()`
But used `packageFieldToPath()` for packages that aren't built yet or otherwise unresolvable

For example `packageEntryToPath('govuk-frontend/src/govuk/all.mjs')` crashes due to this bug:

* https://github.com/alphagov/govuk-frontend/issues/3755

So we have to use `packageFieldToPath()` instead

packages/govuk-frontend-review/rollup.config.mjs
colinrotherham: Bit of a stretch goal. You can install `govuk-frontend@4` into the review app too 😉

A few of the paths will need replacing with globs (to avoid the dist or no-dist prefix)

It's one of the things I stopped working on the other day

shared/lib/names.js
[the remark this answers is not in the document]
romaricpascal: How about `packageModulePath` to make explicit that it finds the path to a specific module in the package?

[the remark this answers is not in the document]
colinrotherham: Will that be confusing as you could pick the `module`, `main` or `sass` fields?

Doesn't actually have to be a module, just a package field

shared/config/resolver.js
romaricpascal: Creating a `require` sounds a good solution to help keep the calls to `packageEntryToPath` closer to where they're used. Actually, all that function needs is the path from `import.meta.url` isn't it, it can centralise that `createRequire` all, maybe? 🤔

colinrotherham: Yeah that can work just for Node.js

Shall I remove all the changes for Sass and Nunjucks and just pass `import.meta.url` (ES) or `__filename` (CommonJS) to every `packageEntryToPath()` call that needs a local `govuk-frontend`?

I picked directories (versus module paths) so we have something that works for Rollup, Sass and Nunjucks

So we'll still need directories for the `@rollup/plugin-node-resolve` plugin:

```mjs
plugins: [
  resolve({
    modulePaths: ['???']
  })
]
```

shared/lib/names.js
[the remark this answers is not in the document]
colinrotherham: I'm going to use `type` instead of `field`

Where the [**package.json** `"type"` field](https://nodejs.org/api/packages.html#type) can be `module` or `commonjs`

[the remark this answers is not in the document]
colinrotherham: How's this? Just pushed up


```mjs
import { paths } from 'govuk-frontend-config'
import { packageTypeToPath } from 'govuk-frontend-lib/names'

// Locate GOV.UK Frontend
const packagePath = packageTypeToPath('govuk-frontend', {
  moduleRoot: paths.stats,
  type: 'module'
})
```

shared/config/resolver.js
colinrotherham: Removed, have a look at https://github.com/alphagov/govuk-frontend/pull/3757#discussion_r1224505955

shared/lib/names.js
```diff
@@ -78,6 +78,43 @@ function packageResolveToPath (packageEntry, { modulePath } = {}) {
...
+ *
+ * ```mjs
+ * const templatePath = packageResolveToPath('govuk-frontend', {
+ *   modulePath: `components/accordion/accordion.mjs`
+ * })
+ * ```
+ * @param {string} packageName - Installed npm package name
+ * @param {PackageOptions} [options] - Package resolution options
+ * @returns {string} Path to installed npm package field
+ */
```
romaricpascal, line 106: Super neat and nice solution for the naming as well 😍

romaricpascal (approved):
Cheers for the changes, looks good to go for me! ⛵

PR #3763, opened 2023-06-07
Change hierarchy of security section in README

36degrees:
The security section is currently an h3 'under' the h2 'Getting updates', but isn't related to getting updates.

Change it to an h2 so that it's a sibling of 'Getting updates' rather than a child, thereby starting a new section.


domoscargin (approved):
🔐

PR #3768, opened 2023-06-07
Change autocomplete example to street-address

matteason:
The textarea documentation currently uses `postal-code` and `username` as examples for the `autocomplete` option. These are good examples for the text input component, but I wouldn't expect a textarea to be used for those fields. 

I've updated the example for textarea to `street-address`. From reading the [spec](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofilling-form-controls:-the-autocomplete-attribute) that's the only autocomplete value I'd really expect to use on a textarea (though technically I think they're all valid) and it's the only value that allows newlines in the canonical format.

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/textarea/textarea.yaml
+++ b/packages/govuk-frontend/src/govuk/components/textarea/textarea.yaml
@@ -58,7 +58,7 @@ params:
   - name: autocomplete
     type: string
     required: false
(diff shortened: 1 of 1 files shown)


36degrees (approved):
Makes sense to me, thanks! Just needs a second review and then we can get this merged.

querkmachine (approved):
Seems logical to me!

Issue #3508, opened 2023-04-14
Update review app to use new JavaScript loading strategy


romaricpascal:
## What

Update the snippet setting a class on the on the document's `body` to mark where GOV.UK Frontend is supported with a `govuk-frontend-supported` class, in complement of the `js-enabled` one marking where JavaScript is available.

That class will be set on the detection of `noModule` in `HTMLScriptElement.prototype` as it's the best way we've found to identify the browsers we support. 

## Why

We currently set a `js-enabled` class on the body to signpost that JavaScript is available in the browser and hook the styles of our components to it. It lets us style our components as early as possible with the assumption that they’ll be enhanced by JavaScript. This avoids a flicker of components showing their “no JavaScript styles” before JavaScript loads.

We’ll keep that to avoid breaking styles of services that may still want to run their own components on good old `<script>` tags like they do now. 

We’ll complement that class with a new `govuk-frontend-supported` one that highlights that GOVUK Frontend can be initialised.

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

```[tasklist]
- [x] `template.njk` has been updated with a new snippet
- [x] Add GOV.UK Frontend using `<script type="module">` in review app
- [x] Ensure GOV.UK frontend is exported as an ES module in review app
- [x] [The SHA hash for the snippet](https://frontend.design-system.service.gov.uk/importing-css-assets-and-javascript/#use-a-hash-to-unblock-inline-javascript) has been updated in the Frontend docs
```

colinrotherham:
These PRs are in review to close this issue off:

* https://github.com/alphagov/govuk-frontend/pull/3801
* https://github.com/alphagov/govuk-frontend-docs/pull/317

colinrotherham:
Aware that we shimmed the `window.GOVUKFrontend` global (only found in UMD bundles) for ES modules in:

https://github.com/alphagov/govuk-frontend/blob/945e8a25463967f3c0a244c5f18259932545e975/packages/govuk-frontend-review/src/javascripts/all.mjs#L4

Do we want to continue doing this for modules?

```html
<script type="module" src="/javascripts/all.min.js"></script>
<script type="module">window.GOVUKFrontend.initAll()</script>
```

Or can we decide to use modules as intended?

```html
<script type="module">
  import { initAll } from '/javascripts/all.min.js'
  initAll()
</script>
```

[...]

PR #3769, opened 2023-06-08
Migrate to use `<script type="module">`

colinrotherham:
This PR migrates the review app + webpack example to `<script type="module">` for https://github.com/alphagov/govuk-frontend/issues/3508

```html
<script type="module" src="/javascripts/all.min.js"></script>
<script type="module">window.GOVUKFrontend.initAll()</script>
```

### Components styled with `.js-enabled`

Once this PR merges there will be a disconnect between browsers that apply `.js-enabled` styling and those that can actually initialise `<script type="module">` functionality but we’ll address that in a future commit

### JavaScript load order

We ideally want GOV.UK Frontend to load _before_ 3rd party scripts so that component initialisation isn't delayed

Since module scripts have [`defer` loading behaviour](https://v8.dev/features/modules#defer) we've added `DOMContentLoaded` to 3rd party scripts:

```html
<script src="/vendor/iframe-resizer/iframeResizer.min.js" defer></script>
<script>
  document.addEventListener('DOMContentLoaded', function () {
    window.iFrameResize({}, '.js-component-preview')
  })
</script>
```


36degrees (approved):
Nice 🎉

colinrotherham:
Linking the follow up PR to this one:

PR #3770, opened 2023-06-08
Upgrade ESLint configuration for ES2015 (without classes)

colinrotherham:
This PR updates our ESLint config as part of https://github.com/alphagov/govuk-frontend/issues/3463 to:

1. Use ES2015 (or ES6) as our target
2. Use [arrow functions](https://caniuse.com/arrow-functions) to preserve `this` in components
3. Use [template strings](https://caniuse.com/template-literals) rather than `A + 'to' + B` concatenation
4. Use [`let`](https://caniuse.com/mdn-javascript_statements_let) or [`const`](https://caniuse.com/const) with block scoping rather than `var`

I've left ESLint rules in place for `.prototype` inheritance as we'll move to ES2015 classes in another PR

### Browser support
We may go higher than ES2015 once we confirm the browser support guarded by our “cut the mustard” checks

We know that browsers supporting `<script type="module">` already [support the new features](https://browsersl.ist/#q=supports+es6-module%2Cnot+supports+arrow-functions%2Cnot+supports+template-literals%2Cnot+supports+let%2Cnot+supports+const) above

```console
supports es6-module
not supports arrow-functions
not supports template-literals
not supports let
not supports const
```

### DIFF
--- a/packages/govuk-frontend/src/govuk/all.mjs
+++ b/packages/govuk-frontend/src/govuk/all.mjs
@@ -24,62 +24,62 @@ function initAll (config) {
 
   // Allow the user to initialise GOV.UK Frontend in only certain sections of the page
   // Defaults to the entire document if nothing is set.
-  var $scope = config.scope instanceof HTMLElement ? config.scope : document
+  const $scope = config.scope instanceof HTMLElement ? config.scope : document
 
-  var $accordions = $scope.querySelectorAll('[data-module="govuk-accordion"]')
+  const $accordions = $scope.querySelectorAll('[data-module="govuk-accordion"]')
   $accordions.forEach(function ($accordion) {
     new Accordion($accordion, config.accordion).init()
   })
 
-  var $buttons = $scope.querySelectorAll('[data-module="govuk-button"]')
+  const $buttons = $scope.querySelectorAll('[data-module="govuk-button"]')
   $buttons.forEach(function ($button) {
     new Button($button, config.button).init()
   })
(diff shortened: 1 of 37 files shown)


packages/govuk-frontend/src/govuk/all.mjs
```diff
@@ -24,63 +24,63 @@ function initAll (config) {
 
   // Allow the user to initialise GOV.UK Frontend in only certain sections of the page
   // Defaults to the entire document if nothing is set.
-  var $scope = config.scope instanceof HTMLElement ? config.scope : document
+  const $scope = config.scope instanceof HTMLElement ? config.scope : document
 
-  var $accordions = $scope.querySelectorAll('[data-module="govuk-accordion"]')
-  $accordions.forEach(function ($accordion) {
+  const $accordions = $scope.querySelectorAll('[data-module="govuk-accordion"]')
+  $accordions.forEach(($accordion) => {
```
colinrotherham, line 30: For all these arrow function changes I'm aware they're split into:

1. Arrow functions to preserve `this` as the component instance
2. Arrow functions to avoid `.bind(this)` in various event handlers
3. Arrow functions simply for "code style" or consistency

I'm aware that there are plenty of event listeners and `.forEach()` loops (like this one) that never actually refer to `this` so may not necessary to change those outside of component scope

Terser is going to convert all that it can to arrow functions anyway to save KBs

36degrees (approved):

PR #3773, opened 2023-06-09
Omit the value attribute from select options with no value

### LINKED ISSUE #3440: Possibly unexpected behaviour if you provide text but no value for select items
## Description of the issue

possible foot gun when using govukSelect component and not passing an explicit value to your select items

https://ukgovernmentdigital.slack.com/archives/C6DMEH5R6/p1680092442731649

https://github.com/alphagov/govuk-design-system-backlog/issues/60

pass only the text parameter to select items when using govukSelect

If you don't supply a value to a select option `<option>text</option>` then the value of that form field will be "text"

However, with govukSelect, [if you don't supply a value for one of the select items then you'll get an empty value attribute](https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/components/select/template.njk#L47) `<option value="">text</option>` which will cause the value of that form field to (maybe unexpectedly) be an empty string ""

I can see the value of being able to pass an empty string as a value, but it seems like it should be an explicit choice rather than the default

N/A

- Operating system:
- Browser:
- Browser version:
- GOV.UK Frontend Version:

36degrees:
In HTML, if you omit the `value` attribute from an `<option>` [the value defaults to the text contained inside the element][1].

However if you omit the value from an option when using the `govukSelect` macro, we currently include an empty `value` attribute on the `<option>` element. This would mean that when submitting a form using a select with options with no value set, the value submitted would be the empty string.

Instead, omit the `value` attribute if no value option has been provided. We then mimic the default HTML behaviour, as the option value will then naturally fall back to the text content.

We still need to ensure that the `value` attribute is included when the value option is passed but has a falsey value (like an empty string, 0, or boolean false).

Also allow the top-level `value` option on a `govukSelect` to work with options that have no explicit value set by comparing it to the text content instead.

Fixes #3440.

[1]: https://html.spec.whatwg.org/multipage/form-elements.html#the-option-element:~:text=The%20value%20attribute%20provides%20a%20value%20for%20element.%20The%20value%20of%20an%20option%20element%20is%20the%20value%20of%20the%20value%20content%20attribute%2C%20if%20there%20is%20one%2C%20or%2C%20if%20there%20is%20not%2C%20the%20value%20of%20the%20element%27s%20text%20IDL%20attribute.

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/select/template.test.js
+++ b/packages/govuk-frontend/src/govuk/components/select/template.test.js
@@ -33,11 +33,41 @@ describe('Select', () => {
       expect($items.length).toEqual(3)
     })
 
-    it('renders item with value', () => {
+    it('includes the value attribute', () => {
       const $ = render('select', examples.default)
 
       const $firstItem = $('.govuk-select option:first-child')
       expect($firstItem.attr('value')).toEqual('1')
+    })
+
+    it('includes the value attribute when the value option is an empty string', () => {
+      const $ = render('select', examples['with falsey values'])
+
+      const $firstItem = $('.govuk-select option:nth(0)')
+      expect($firstItem.attr('value')).toEqual('')
+    })
+
+    it('includes the value attribute when the value option is false', () => {
+      const $ = render('select', examples['with falsey values'])
+
+      const $secondItem = $('.govuk-select option:nth(1)')
+      expect($secondItem.attr('value')).toEqual('false')
+    })
+
+    it('includes the value attribute when the value option is 0', () => {
+      const $ = render('select', examples['with falsey values'])
+
+      const $thirdItem = $('.govuk-select option:nth(2)')
+      expect($thirdItem.attr('value')).toEqual('0')
+    })
+
+    it('omits the value attribute if no value option is provided', () => {
+      const $ = render('select', examples['without values'])
+
+      const $firstItem = $('.govuk-select option:first-child')
+      // Ideally we'd test for $firstItem.attr('value') == undefined but it's
+      // broken in Cheerio – https://github.com/cheeriojs/cheerio/issues/3237
+      expect($firstItem.toString()).not.toContain('value')
     })
 
     it('renders item with text', () => {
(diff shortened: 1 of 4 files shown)


packages/govuk-frontend/src/govuk/components/select/template.test.js
```diff
@@ -33,13 +33,43 @@ describe('Select', () => {
...
+      const $thirdItem = $('.govuk-select option:nth(2)')
+      expect($thirdItem.attr('value')).toEqual('0')
+    })
+
+    it('omits the value attribute if no value option is provided', () => {
+      const $ = render('select', examples['without values'])
+
+      const $firstItem = $('.govuk-select option:first-child')
+      // Ideally we'd test for $firstItem.attr('value') == undefined but it's
+      // broken in Cheerio – https://github.com/cheeriojs/cheerio/issues/3237
```
colinrotherham, line 69: ❤️

packages/govuk-frontend/src/govuk/components/select/template.test.js
```diff
@@ -61,6 +91,13 @@ describe('Select', () => {
       expect($selectedItem.attr('selected')).toBeTruthy()
     })
 
+    it('selects options with implicit value using selected value', () => {
+      const $ = render('select', examples['without values with selected value'])
+
+      const $selectedItem = $('option:contains(\'Green\')')
```
colinrotherham, line 97: ESLint should be happy with double quotes when they're intentionally nesting single quotes if you like

```suggestion
      const $selectedItem = $("option:contains('Green')")
```

colinrotherham (approved):

colinrotherham:
@36degrees Do we need a CHANGELOG entry for this should anyone rely on the old behaviour?

36degrees:
> Do we need a CHANGELOG entry for this should anyone rely on the old behaviour?

D'oh! I'd added it but not pushed. Have applied your suggestion and updated.

colinrotherham (approved):
Looking good still 😊

PR #3774, opened 2023-06-09
Add inverse breadcrumb and back link modifiers and styles

### LINKED ISSUE #2266: Add inverse breadcrumb and back link modifiers
## What

Add a modifier to the breadcrumbs component to invert the colours, for situations where breadcrumbs appear against dark backgrounds (for example, in the blue masthead on product pages)

## Why

The [product page example](https://github.com/alphagov/product-page-example) repo currently contains [CSS to invert breadcrumbs](https://github.com/alphagov/product-page-example/blob/master/source/stylesheets/modules/_breadcrumbs.scss) for use on product pages across GaaP.

There are already various examples of inverse breadcrumbs on GOV.UK, for example on the Coronavirus page, on topic pages and across the various GaaP product pages:

Screenshot 2021-07-05 at 11 57 46
Screenshot 2021-07-05 at 11 57 52
Screenshot 2021-07-05 at 11 57 54
Screenshot 2021-07-05 at 11 58 00
Screenshot 2021-07-05 at 12 00 55

It would make sense to add a modifier for back links at the same time, as they occupy the same position on the page and are conceptually similar.

Designers, developers

## Further detail

Many of these examples also add a border to the bottom of the breadcrumb. Is that needed? Is that a part of the breadcrumb?

## Done when

- [x] Inverse breadcrumb modifier added
- [x] Inverse back link modifier added
- [x] Examples added to the review app

querkmachine:
Because everyone loves a small story.

Adds inverted colour scheme variants of the Breadcrumb and Back Link components, for use on darker backgrounds, as is used on product landing pages and some sections of GOV.UK.

Related to #2266.

## Thoughts

- We don't have a mechanism for showing examples of these inverse coloured variants in the review app—currently it's white text on a white background. We should probably do something about that!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,18 @@ You can now style buttons on dark backgrounds to have a white background colour 
 
 This change was made in [pull request #3556: Add inverse button styles](https://github.com/alphagov/govuk-frontend/pull/3556).
 
+#### Added inverse modifier for breadcrumbs on dark backgrounds
+
+You can now style breadcrumbs on dark backgrounds to use white text, links and arrows by adding the `govuk-breadcrumbs--inverse` class.
+
+This change was made in [pull request #3774: Add inverse breadcrumb and back link modifiers and styles](https://github.com/alphagov/govuk-frontend/pull/3774).
+
+#### Added inverse modifier for back links on dark backgrounds
+
+You can now style back links on dark backgrounds to have a white link and arrow colour by adding the `govuk-back-link--inverse` class.
+
+This change was made in [pull request #3774: Add inverse breadcrumb and back link modifiers and styles](https://github.com/alphagov/govuk-frontend/pull/3774).
+
 #### New link styles are now enabled by default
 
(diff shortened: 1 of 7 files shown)


owenatgov (approved):
Does the job. Tested by applying the class and adding a black background locally.

You make a good point on not being able to see examples of this in the review app and I think we can get around this relatively easily by updating the [showExamples macro](https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend-review/src/views/macros/showExamples.njk). I'll put an issue together.

owenatgov:
An extra comment: I wonder if we should keep the linked issue open after this gets merged for 2 reasons:

1. Technically speaking, the issue requires these examples get added to the review app and as we've illustrated that's not possible yet. Hopefully this can be solved with https://github.com/alphagov/govuk-frontend/issues/3780
2. Ollie makes a comment on the border at the bottom of the breadcrumbs in the govuk examples. I think it's worth using that issue to explore this more in future

querkmachine:
Good points, I've removed the closing comment.

PR #3776, opened 2023-06-09
Simplify logic in `govuk-font` mixin

36degrees:
Both flows through this if statement `@include govuk-typography-common` so we can move it out of the if statement and disappear the else clause 🪄


romaricpascal (approved):
Good little catch ⛵

Issue #2293, opened 2021-08-02
Improve support for Windows High Contrast Mode and Internet Explorer 11


lfdebrux:
This builds on work by @chris-gds for the accordion component refresh in #2257.

## What

Add media feature `-ms-high-contrast` to media queries that change styles for Windows High Contrast Mode.

For example, see draft PR #2290 

## Why

We can port fixes for High Contrast Mode such as improved focus states for radios and checkboxes (https://github.com/alphagov/govuk-frontend/pull/2264) that currently work in Edge / Chrome / Opera but not Internet Explorer.

According to the 2018 WebAIM survey of users with low vision, a higher percentage use Internet Explorer than Edge.

## Who needs to know about this

Devs

## Done when

- [ ] GOV.UK Frontend has support for Windows High Contrast Mode and Internet Explorer 11
- [ ] Documented how to add tweaks for Windows High Contrast Mode and other forced colour modes in future

lfdebrux:
It would be nice if `autoprefixer` dealt with this for us, however it looks like this functionality won't be added: https://github.com/postcss/autoprefixer/issues/1351 

We could add a mixin for Windows High Contrast Mode (as suggested by @chris-gds), however even in that case we would need to be careful when using it to make sure we include prefixed CSS properties such as `forced-color-adjust`/`-ms-high-contrast-adjust`.

Also, there may be scenarios where we want to use the `forced-color` media query but not the `-ms-high-contrast` media query.

PR #3777, opened 2023-06-09
Fix hover style on small checkboxes and radio buttons in High Contrast Mode

### LINKED ISSUE #3695: Hover states for checkboxes and radios aren't visible in Windows High Contrast Mode

**Usability issue**
A component did not display a change of state to indicate mouse hover.

**Issue ID:**
DAC_Usability_Mouse_Hover_01

**URLs:**
Example page link: https://govuk-frontend-pr-3503.herokuapp.com/full-page-examples/search

Relevant guidance pages: [Checkboxes](https://design-system.service.gov.uk/components/checkboxes/) and [Radios](https://design-system.service.gov.uk/components/radios/).

## Screen shots
A set of radio options within a search filter module, shown in Windows High Contrast Mode. The mouse pointer is hovering over the first radio input and there is no visual change to indicate a hover state.

A set of checkboxes within a search filter module, shown in Windows High Contrast Mode. The mouse pointer is hovering over the first checkbox input and there is no visual change to indicate a hover state.

The checkboxes and radio buttons do not display a visible change when receiving mouse hover in high contrast mode. This means that some users may not be able to distinguish when they are able to activate the component and form.

## Current code ref(s)
`#main-content > form > div.govuk-grid-row > div.govuk-grid-column-one-third > ul > li:nth-child(4) > div > fieldset > div > div:nth-child(1)`

```
<div class="govuk-checkboxes__item">
    <input class="govuk-checkboxes__input" id="organisation" name="organisation" type="checkbox" value="hmrc">
    <label class="govuk-label govuk-checkboxes__label" for="organisation">HM Revenue and Customs (HMRC)</label>
</div>
```

## Solution
Consider adding a clear indication for users when hovering over inputs and forms with a mouse to indicate that they are interactive.
Example: 
```
.govuk-radios__input {
    cursor: pointer;
}
```

querkmachine:
Hover styles are currently not visible on 'small' checkbox and radio buttons in Windows High Contrast Mode, because the `box-shadow` property is ignored in High Contrast Mode. 

This PR aims to fix that by also applying an `outline` attribute at the same time: transparent normally, but coloured in High Contrast Mode, so that a visual change still takes place on hover.

As `outline` is also used by the focus state, there is some possible overlap here, but given we additionally thicken the input border on focus, a visual distinction between hovered, focused, and simultaneously hovered and focused is retained.

Raised as a low vision usability issue by DAC in their recent audit of Frontend. Resolves #3695.

||Before|After|
|:-|:-|:-|
|Checkbox (hovered)|image|image|
|Checkbox (focused)|image|image|
|Checkbox (hovered and focused)|image|image|
|Checkbox (hovered, focused and checked)|image|image|
|Radio (hovered)|image|image|
|Radio (focused)|image|image|
|Radio (hovered and focused)|image|image|
|Radio (hovered, focused and checked)|image|image|

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/checkboxes/_index.scss
+++ b/packages/govuk-frontend/src/govuk/components/checkboxes/_index.scss
@@ -273,6 +273,10 @@
     // feedback to the user as to which checkbox they will select when their
     // cursor is outside of the visible area.
     .govuk-checkboxes__item:hover .govuk-checkboxes__input:not(:disabled) + .govuk-checkboxes__label::before {
+      // Forced colours modes tend to ignore box-shadow.
+      // Apply an outline for those modes to use instead.
+      outline: $govuk-focus-width dashed transparent;
+      outline-offset: 1px;
       box-shadow: 0 0 0 $govuk-hover-width $govuk-hover-colour;
     }
 
@@ -282,6 +286,10 @@
     // We use two box shadows, one that restores the original focus state [1]
     // and another that then applies the hover state [2].
     .govuk-checkboxes__item:hover .govuk-checkboxes__input:focus + .govuk-checkboxes__label::before {
+      // Set different HCM colour when we have both hover/focus applied at once
+      @media screen and (forced-colors: active), (-ms-high-contrast: active) {
+        outline-color: Highlight;
+      }
       box-shadow:
         0 0 0 $govuk-focus-width $govuk-focus-colour, // 1
         0 0 0 $govuk-hover-width $govuk-hover-colour; // 2
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -252,6 +252,7 @@ This change was introduced in [pull request #3773: Omit the value attribute from
 
 We’ve made fixes to GOV.UK Frontend in the following pull requests:
 
+- [#3777: Fix hover style on small checkboxes and radio buttons in High Contrast Mode](https://github.com/alphagov/govuk-frontend/pull/3777)
 - [#3791: Refactor mobile menu button label/text handling](https://github.com/alphagov/govuk-frontend/pull/3791)
 - [#3862: Fix focus style being overlapped by summary action links](https://github.com/alphagov/govuk-frontend/pull/3862)
 
--- a/packages/govuk-frontend/src/govuk/components/radios/_index.scss
+++ b/packages/govuk-frontend/src/govuk/components/radios/_index.scss
@@ -293,6 +293,10 @@
     // feedback to the user as to which radio they will select when their
     // cursor is outside of the visible area.
     .govuk-radios__item:hover .govuk-radios__input:not(:disabled) + .govuk-radios__label::before {
+      // Forced colours modes tend to ignore box-shadow.
+      // Apply an outline for those modes to use instead.
+      outline: $govuk-radios-focus-width dashed transparent;
+      outline-offset: 1px;
       box-shadow: 0 0 0 $govuk-hover-width $govuk-hover-colour;
     }
 
@@ -302,6 +306,10 @@
     // We use two box shadows, one that restores the original focus state [1]
     // and another that then applies the hover state [2].
     .govuk-radios__item:hover .govuk-radios__input:focus + .govuk-radios__label::before {
+      // Set different HCM colour when we have both hover/focus applied at once
+      @media screen and (forced-colors: active), (-ms-high-contrast: active) {
+        outline-color: Highlight;
+      }
       box-shadow:
         0 0 0 $govuk-radios-focus-width $govuk-focus-colour, // 1
         0 0 0 $govuk-hover-width        $govuk-hover-colour; // 2


36degrees:
Thanks for picking this up!

I do have some mild reservations about using Highlight as a colour for hover states. Similar to how we try to reserve yellow only for focus states, I think it would make sense to do the same for Highlight.

I'm wondering if we could explore a few alternative options, possibly using a thinner dotted or dashed outline could work? Given the limited palette and reduced feature set of forced colours mode, I'd also consider simplifying things by not trying to cover the focused + hovered state, and instead just letting the focus state 'win'.

I think it would be worth having this reviewed by @davidc-gds and a designer too to make sure they're happy with the changes.

querkmachine:
> I do have some mild reservations about using Highlight as a colour for hover states. Similar to how we try to reserve yellow only for focus states, I think it would make sense to do the same for Highlight.

In looking at the prior work (aka, Windows itself) in High Contrast Mode, it seems to be standard to use the Highlight colour for hovering. This is at least the case when hovering over checkboxes, radios, buttons, menu items, selects and toggles too. 

A sample checkbox from Windows 10 High Contrast Mode. The label is being hovered, turning the checkbox's border blue.

A sample radio button from Windows 10 High Contrast Mode. The label is being hovered, turning the radio's border blue.

Interestingly, this means our usage of the colours is inverted compared to the Windows default, which uses white for focus states and Highlight for hover.

davidc-gds:
> this means our usage of the colours is inverted compared to the Windows default, which uses white for focus states and Highlight for hover.

Heya! Chiming in to give some thoughts that might help us make a decision, while not rethinking our entire high-contrast-mode focus and hover implementation.

In my view, a visible hover when using contrast themes (like Windows High Contrast Mode) in any format is better than no visible hover.

I am also thinking that some sort of visual change (dotted, dashed, thickness change) might satisfy that view though, as Ollie suggested.

querkmachine:
Yep! Not planning to reverse everything we have currently, just an interesting thing I noticed whilst experimenting 😅

querkmachine:
Updated the initial comment with a dashed line approach instead.

Note that I'm not applying any specific styles for the 'hovered and focused' states, this is just how they're turning out with the cascade the way it currently is, and documented for completion's sake.

querkmachine:
@CharlotteDowns @christopherthomasdesign @Ciandelle Sorry for another ask, but any particular views on the dashed border look in the top comment?

christopherthomasdesign:
Hey @querkmachine 

I think I'm ok with this in theory, it's quite a neat solution to the problem and it maintains a feeling that moving from hover to focus adds more visual weight to the thing.

Are there any other components with hover states that should use this convention on high contrast mode too?

querkmachine:
@christopherthomasdesign I'm not sure, myself. It was flagged in this specific instance because we provide a hover style normally to compensate for the radios and checkboxes having a small visual target, but a large actual target area. This affordance didn't exist in High Contrast Mode.

There are a few other elements that have a hover style normally but don't in High Contrast Mode, but they also don't have an invisible target area, so it's not precisely the same circumstances as here.

davidc-gds:
Based on this thread, it seems like there's agreement that this is a good solution that doesn't interfere with anything els. So I'd recommend we go for it! 👍 

On @christopherthomasdesign's question about whether we have other hover-able interactive components where the hover isn't carrying over to the contrast-themed version (for example, when using Windows High Contrast Mode):

- I've tested a few other components which might be worth considering, shown in the headings below
- This feels like a larger exploration of hover states that might be better suited for after we wrap up this main radios / checkboxes hover piece

### Start buttons
Taken from this page: https://design-system.service.gov.uk/components/button/
(This is likely the case for all links styled as buttons)
On regular colours mode (not contrast themed), both the 'Start' button and the regular button get focus states. When contrast themed, the 'Start' button doesn't get a hover state change.

**Regular button un-hovered:**
image

**Regular button hovered:**
image

**Start button un-hovered:**
image

**Start button hovered:**
image

### Accordion 'Show all' button
Taken from this page: https://design-system.service.gov.uk/components/accordion/
On regular colours mode (not contrast themed), both the 'Show all' button and the individual accordion items get focus states. When contrast themed, the 'Show all' doesn't get a hover state change.

**Accordion area un-hovered:**
image

**Accordion item hovered:**
image

**Accordion 'Show all sections' hovered:**
image

### Checkboxes
We already know this one (it's what this issues is about)

### File upload
(This doesn't have a hover stat when contrast themed, but it's also a native browser control at the moment)

### Radios
We already know this one (it's what this issues is about)

### Select
Oddly, this one has a hover state when contrast themed with Windows High Contrast Mode, but it doesn't normally have a hover state.

**Select un-hovered in high-contrast mode:**
image

**Select hovered in high-contrast mode:**
image

**Select un-hovered regularly:**
image

**Select hovered regularly:**
image

### Text input
We know these don't have hover states, in both contrast theme and regular theme.

### Text area
We know these don't have hover states, in both contrast theme and regular theme.

davidc-gds:
@querkmachine sorry to add another thought right after that previous one.

For the states where there's multiple things happening at once, would it make sense to combine the focus/active and hover state properties?

So make the dotted 'hover' outline use the 'highlight' blue colour when:
- Checkbox (hovered and focused)
- Checkbox (hovered, focused and checked)
- Radio (hovered and focused)
- Radio (hovered, focused and checked)

querkmachine:
@davidc-gds That seems a sensible change to me! Don't think it's too controversial either (I hope!)

christopherthomasdesign:
Yeah all looks good to me 👍 thanks for the suggestions @davidc-gds 🙌

davidc-gds (approved):
Now that we've switched to the dotted outline concept for the hover enhancements, this feels like an overall positive improvement for accessibility when using contrast themes. ✅ 

I can't speak fully to the code portion of this and would recommend getting a developer to review as well, but from a glance it looks to change what we want to change and doesn't do anything extra.

packages/govuk-frontend/src/govuk/components/checkboxes/_index.scss
```diff
@@ -282,6 +286,10 @@
     // We use two box shadows, one that restores the original focus state [1]
     // and another that then applies the hover state [2].
     .govuk-checkboxes__item:hover .govuk-checkboxes__input:focus + .govuk-checkboxes__label::before {
+      // Set different HCM colour when we have both hover/focus applied at once
+      @media screen and (forced-colors: active), (-ms-high-contrast: active) {
```
colinrotherham, line 290: Double checked to see if Autoprefixer does it now (just in case) but no.

Thanks for adding prefixes still. Adding another mention for:

* https://github.com/alphagov/govuk-frontend/issues/2293

packages/govuk-frontend/src/govuk/components/checkboxes/_index.scss
```diff
@@ -282,6 +286,10 @@
     // We use two box shadows, one that restores the original focus state [1]
     // and another that then applies the hover state [2].
     .govuk-checkboxes__item:hover .govuk-checkboxes__input:focus + .govuk-checkboxes__label::before {
+      // Set different HCM colour when we have both hover/focus applied at once
```
colinrotherham, line 289: Do we need this boilerplate comment from other places `Hightlight` is used?

```scss
// When in an explicit forced-color mode, we can use the Highlight system
// color for the outline to better match focus states of native controls
```

colinrotherham (approved):
No blockers just some minor comments, looks brill 🙌

PR #3543, opened 2023-04-25
Switch to new Chrome headless mode

colinrotherham:
Following https://github.com/alphagov/govuk-frontend/pull/3542 just a quick PR for:

* https://github.com/alphagov/govuk-frontend/issues/3541

For more info on our headless mode deprecation issues see:

* https://github.com/puppeteer/puppeteer/issues/10070

PR #3807, opened 2023-06-13
Fix flaky tests with debounce timeouts

colinrotherham:
This PR updates our button tests to use click handling without a `<form>` wrapper

Instead of counting events that may (or may not) bubble up, we now count every click + debounce

I've also been able to remove some legacy `{ headless: true }` workarounds since:

* https://github.com/alphagov/govuk-frontend/pull/3543

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/button/button.test.js
+++ b/packages/govuk-frontend/src/govuk/components/button/button.test.js
@@ -4,6 +4,12 @@ const { getExamples } = require('govuk-frontend-lib/files')
 describe('/components/button', () => {
   let examples
 
+  const clickTimeoutTime = 1000 // ms
+
+  // The longest possible time a button will ignore unintentional clicks for
+  // until it can be clicked again (+ 50ms to stay outside the total wait time)
+  const debouncedWaitTime = clickTimeoutTime + 50
+
   beforeAll(async () => {
     examples = await getExamples('button')
   })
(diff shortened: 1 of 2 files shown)


packages/govuk-frontend/src/govuk/components/button/button.test.js
```diff
@@ -55,105 +55,121 @@ describe('/components/button', () => {
...
+    async function trackClicks ($button) {
+      const counts = {
+        click: 0,
+        debounce: 0
+      }
+
+      // Track number of button clicks
+      return $button.evaluate((el, counts) => el.addEventListener('click', (event) => {
+        counts.click++
+        el.dataset.clickCount = `${counts.click}`
```
colinrotherham, line 73: Counts every click to make sure Puppeteer isn't throttling them

packages/govuk-frontend/src/govuk/components/button/button.test.js
```diff
@@ -55,105 +55,121 @@ describe('/components/button', () => {
...
+
+        // Track number of button clicks that debounced
+        event.preventDefault = () => {
```
colinrotherham, line 76: Replaces `event.preventDefault()` with a stub that counts every time we debounce a click

packages/govuk-frontend/src/govuk/components/character-count/character-count.test.js
```diff
@@ -1,20 +1,14 @@
...
-// Detect when browser has been launched headless
-const { headless = 'new' } = jestPuppeteerConfig.launch
-
 // The longest possible time from a keyboard user ending input and the screen
 // reader counter being updated: handleFocus interval time + last input wait time
-// but raised to the nearest whole second when browser is not headless as timers
-// in background pages will be aligned, clamped or throttled
-const debouncedWaitTime = headless ? 1500 : 2000
+// but raised by 50ms to avoid the timeout window
+const debouncedWaitTime = 1550
```
colinrotherham, line 7: Since https://github.com/alphagov/govuk-frontend/pull/3543 (or perhaps in recent Puppeteer updates) we're seeing much more accurate timers

Running assertions at the exact `setTimeout()` period can fail the test so I've added 50ms on

packages/govuk-frontend/src/govuk/components/character-count/character-count.test.js
```diff
@@ -1,24 +1,16 @@
...
-
 describe('Character count', () => {
   let examples
 
+  const focusIntervalTime = 1000 // ms
+  const lastInputOffsetTime = 500 // ms
+
+  // The longest possible time from a keyboard user ending input and the screen
+  // reader counter being updated (+ 50ms to stay outside the total wait time)
+  const debouncedWaitTime = focusIntervalTime + lastInputOffsetTime + 50
```
colinrotherham, line 12: Since https://github.com/alphagov/govuk-frontend/pull/3543 (or perhaps in recent Puppeteer updates) we're seeing much more accurate timers

Running assertions at the exact `setTimeout()` period can fail the test so I've added 50ms on

packages/govuk-frontend/src/govuk/components/button/button.test.js
```diff
@@ -55,105 +61,121 @@ describe('/components/button', () => {
...
         })
 
         await page.click('button:nth-child(1)')
         await page.click('button:nth-child(2)')
 
-        const clicksCount = await getClicksCount(page)
+        const counts = await getButtonTracking($button)
 
-        expect(clicksCount).toBe(2)
+        expect(counts.click).toBe(1)
```
domoscargin, line 170: I'm a bit confused by this one - through the `beforeEach`, we're tracking `$button1`, so that's all good, but with this test we don't seem to know if `$button2` has actually been clicked?

domoscargin (commented):
Looks great - love the splitting of click and debounce info - gives a really clear picture.

I have one query, but I suspect I've just missed something obvious there.

packages/govuk-frontend/src/govuk/components/button/button.test.js
colinrotherham: I did wonder about that. The old tests didn't either (when it was named `$secondButton`)

Shall we add tracking to both buttons?

domoscargin: Definitely - to satisfy the test description we need to ensure the second button was clicked successfully. 

The previous test looks like it was at least aiming to test that there were two valid clicks after clicking each button once.

colinrotherham: Thanks @domoscargin, can you take a look now?

packages/govuk-frontend/src/govuk/components/button/button.test.js
```diff
@@ -55,105 +61,126 @@ describe('/components/button', () => {
...
+        const button2Counts = await getButtonTracking($button2)
 
-        const clicksCount = await getClicksCount(page)
+        // 2nd click on button 1 prevented
+        expect(button1Counts.click).toBe(2)
+        expect(button1Counts.debounce).toBe(1)
 
-        expect(clicksCount).toBe(2)
+        // 3rd click on button 2 not prevented
+        expect(button2Counts.click).toBe(1)
```
domoscargin, line 175: 👍

packages/govuk-frontend/src/govuk/components/button/button.test.js
domoscargin: Nice, looks sound!

domoscargin (approved):
The only flakes we accept here are made of chocolate. 🍫

PR #3808, opened 2023-06-14
Add ability to invert background colour in review app examples

### LINKED ISSUE #3780: Allow for viewing of inverse modifications of components in the review app
## What
Add functionality to the review app to apply a dark background on examples that are intended to show inverse modifiers of components.

This can probably done quite easily by updating [the `showExamples` macro](https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend-review/src/views/macros/showExamples.njk) to check for a param like `inverse` in the component example list.

## Why
So that we can more easily view and manually test these components and their modifiers.

Originally raised in https://github.com/alphagov/govuk-frontend/pull/3774

## Who needs to work on this
Developers

querkmachine:
Adds support for adding a metadata flag (`invertLayoutColours`) to review app examples. If the flag exists, the example is rendered with a blue coloured background. 

Resolves #3780. 

## Changes
- Updates app JS so that it looks for a `inverseLayoutColours` flag in each example's metadata. If it finds it, it adds an additional inverting class to the `bodyClasses` string.
- Adds style for the inverting class. It makes the body background blue.
- Updates Button, Breadcrumbs and Back Link component examples that use inverse modifiers to have the flag. 

## Thoughts
- There's almost certainly a better name for the flag than `invertLayoutColours`.
- The class is currently added to the `<body>` when it might make more sense being on `<html>`. However, the `htmlClasses` Nunjucks variable doesn't seem to be accessible at the same point that `bodyClasses` is being set.
- The inverted background colour being used is blue. This is the colour we use for 'dark background' examples in the Design System documentation and is what many of the inverted-colour modifiers were originally designed to be used against (being ported from [the Product Page template](https://github.com/alphagov/product-page-example)). There's a potential "Why not on black?" question here, but I'm not particularly passionate one way or another.


packages/govuk-frontend-review/src/app.mjs
```diff
@@ -133,8 +133,12 @@ export default async () => {
 
     let bodyClasses = ''
 
+    if (exampleConfig.invertLayoutColours) {
+      bodyClasses = `${bodyClasses} app-inverse-example-background`
```
colinrotherham, line 137: Feels like we've now got three battling ways of setting the background colour?

Added to **`<body>`** is `.govuk-template__body` and `.app-inverse-example-background`
Added to **`<html>`** is `.app-no-canvas-background` which already [sets a white background](https://github.com/alphagov/govuk-frontend/commit/8a30d6eaed4215c88525902a0ef0ed3ad384699f)

Can we find a way to unify all our `.app` classes maybe with the `--inverse` modifier?

```css
.app-template {}
.app-template__body {}
.app-template__body--inverse {}
```

colinrotherham (commented):
This is brilliant 🙌

Can you fix the "inverse" colour for `$app-preview-border-colour` too?

We're using `#e4f2ff` at the moment which works with a white background, maybe:

```scss
$app-preview-border-colour: mix(govuk-colour("blue"), govuk-colour("white"), 10%);
$app-preview-border-colour-inverse: mix(govuk-colour("white"), govuk-colour("blue"), 10%);
```

Added a comment about background colour classes too

packages/govuk-frontend/src/govuk/components/button/button.yaml
```diff
@@ -145,25 +145,29 @@ examples:
...
     description: A link button for actions that appear on dark backgrounds
+    invertLayoutColours: true
     data:
       name: Inverse
       text: Inverse button
       href: '/'
       classes: govuk-button--inverse
   - name: Inverse start button
     description: A start button that appears on dark backgrounds
+    invertLayoutColours: true
```
colinrotherham, line 170: We already use `previewLayout: full-width`

How about `previewLayoutInverse: true`?

```suggestion
    previewLayoutInverse: true
```

colinrotherham (approved):
I'm going to approve this because it's looking amazing since the last push

Added some thoughts on `invertLayoutColours` naming but not a blocker 😊

packages/govuk-frontend/src/govuk/components/button/button.yaml
querkmachine: I opted against using `inverse` on the basis that it might not be descriptive enough ("Inverse what? The component? The text direction?"), but `previewLayoutInverse` sounds like a good candidate.

I'm also pondering whether using the word 'dark' might be better than 'inverse' as that (probably?) better describes what it is—it's the component that is inverted, rather than the layout. Will have a ponder.

colinrotherham: Yeah, all valid points but worked for `--inverse` 🤷‍♂️

colinrotherham: Or another idea?

```suggestion
    previewLayoutModifiers:
      - inverse
```

colinrotherham (approved):
Still superb, approving again 🙌

PR #3812, opened 2023-06-15
[WIP] Migrate component data to ES modules

colinrotherham:
This PR is a work in progress to migrate our component data YAML into ES modules

This means we can:

1. Run type checks using TypeScript
2. Run code quality checks using ESLint
3. Import examples into tests without YAML conversion

This is important because we've accidentally missed things in the past:

* https://github.com/alphagov/govuk-frontend/issues/1903

But makes it a lot easier to add automated tests for:

* https://github.com/alphagov/govuk-frontend/issues/1139


colinrotherham:
Lots of tests will fail as I've only made a proof of concept using:

1. Accordion
2. Back link
3. Breadcrumbs
4. Button

colinrotherham:
Just rebasing since #2261 and `tsc` has caught [another `descrption` typo in **task-list.yaml**](https://github.com/alphagov/govuk-frontend/blob/fbd82b327a4203dae8ac5871eb9dbacee8936911/packages/govuk-frontend/src/govuk/components/task-list/task-list.yaml#L45)

romaricpascal (commented):
Sounds a neat way to catch errors, and gives us some nice way to catch which component options use specific fields (like `isComponent`). 😍 

Screenshot 2023-06-23 at 13 14 21

No qualms on the naming which sounds fine to me (and nice to separate the params from the examples in different files as well).

Besides fixing the tests, Ollie suggested it could be good to present this to the whole team to take the temperature on not using YAML anymore, as it's a quite different authoring experience (especially for examples) 😊

Izabela-16:
@colinrotherham to add to the next show & tell

jest.config.mjs
```diff
@@ -48,7 +48,7 @@ const config = {
 
     // Transform some `*.js` to compatible CommonJS
     ...Object.fromEntries(
-      ['del', 'slash'].map((packagePath) => [
+      ['govuk-frontend-lib/components', 'del', 'slash'].map((packagePath) => [
```
colinrotherham, line 51: Since `components.js` is CommonJS, we need Jest to Babel transform `import()`

Izabela-16:
@colinrotherham Book a slot to talk about it at the next show & tell (sprint one).

PR #3816, opened 2023-06-16
Fix Node.js <= 12.18 require + import

colinrotherham:
As described in https://github.com/alphagov/govuk-frontend/issues/3755 this PR fixes point 2)

> We've identifying two scenarios when trying to `require()` or `import` GOV.UK Frontend
> 
> 1. Node.js >= 17 errors with GOV.UK Frontend v4 [folder mapping in `"exports"`](https://nodejs.org/docs/latest-v18.x/api/deprecations.html#dep0148-folder-mappings-in-exports-trailing-)
> 2. Node.js <= 12.18 errors with GOV.UK Frontend v5 (unreleased) [subpath patterns in `"exports"`](https://nodejs.org/docs/latest-v18.x/api/packages.html#subpath-patterns)

This may also fix issues with bundlers that also follow Node.js package exports resolution


.github/workflows/tests.yml
```diff
@@ -204,6 +204,58 @@ jobs:
...
+        env:
+          # Node.js conditions override from "require" to "import" etc
+          # https://nodejs.org/api/cli.html#-c-condition---conditionscondition
+          FLAGS: ${{ matrix.conditions != 'require' && format(' --conditions {0}', matrix.conditions) || '' }}
+
+        run: |
+          node --eval "console.log(require.resolve('govuk-frontend'))"${{ env.FLAGS }}
+          node --eval "console.log(require.resolve('govuk-frontend/package.json'))"${{ env.FLAGS }}
+          node --eval "console.log(require.resolve('govuk-frontend/dist/govuk/i18n.js'))"${{ env.FLAGS }}
+          node --eval "console.log(require.resolve('govuk-frontend/dist/govuk-esm/i18n.mjs'))"${{ env.FLAGS }}
```
romaricpascal, line 257: Brilliant! That's going to be neat to check we're not breaking things by changing `package.json` 😍

packages/govuk-frontend/package.json
```diff
@@ -16,12 +17,13 @@
       "import": "./dist/govuk-esm/all.mjs",
       "require": "./dist/govuk/all.js"
     },
-    "./*": "./*"
+    "./*": "./*",
+    "./dist/": "./dist/",
+    "./package.json": "./package.json"
```
romaricpascal, line 22: **question**: Is there a way we can allow "./" in a similar fashion as "./*"? That'd let us not worry about which folders are specifically accessible 😊

romaricpascal (commented):
Neat! And nice one for the testing. If the answer to my question is "no", then all good to go 🙌🏻

packages/govuk-frontend/package.json
colinrotherham: Ah sorry, I started with that but Visual Studio Code gives me this:

>Property ./ is not allowed.

Are there any new paths in v5 we should add for legacy Node.js?

colinrotherham: This was my thinking process by the way:

1. Take the compatibility subpath exports from [v4.6.0](https://github.com/alphagov/govuk-frontend/releases/tag/v4.6.0):

```json
"./govuk/": "./govuk/",
"./govuk-esm/": "./govuk-esm/",
"./package.json": "./package.json"
```

2. Convert compatibility subpath exports for v5

```json
"./dist/govuk/": "./dist/govuk/",
"./dist/govuk-esm/": "./dist/govuk-esm/",
"./package.json": "./package.json"
```

3. Simplify compatibility subpath exports:

```json
"./dist/": "./dist/",
"./package.json": "./package.json"
```

colinrotherham: Mind if I just do `govuk-prototype-kit.config.json` as Node.js can't require/import Markdown?

colinrotherham: Done it, but let me know

romaricpascal (approved):

PR #3817, opened 2023-06-16
Fix package resolution in Node.js 17+ for `support/4.x`

colinrotherham:
As described in https://github.com/alphagov/govuk-frontend/issues/3755 this PR fixes point 1)

> We've identified two scenarios when trying to `require()` or `import` GOV.UK Frontend
> 
> 1. Node.js >= 17 errors with GOV.UK Frontend v4 [folder mapping in `"exports"`](https://nodejs.org/docs/latest-v18.x/api/deprecations.html#dep0148-folder-mappings-in-exports-trailing-)
> 2. Node.js <= 12.18 errors with GOV.UK Frontend v5 (unreleased) [subpath patterns in `"exports"`](https://nodejs.org/docs/latest-v18.x/api/packages.html#subpath-patterns)

This may also fix issues with bundlers that also follow Node.js package exports resolution

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+### Fixes
+
+We’ve made fixes to GOV.UK Frontend in the following pull requests:
+
+- [#3817: Fix package resolution in Node.js 17+](https://github.com/alphagov/govuk-frontend/pull/3817)
+
 ## 4.6.0 (Feature release)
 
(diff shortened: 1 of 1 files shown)


colinrotherham:
> Does this need changing to target `support/4.x`?

I've held off doing it to save time but if we're 100% sure then I'll proceed 🚀

colinrotherham:
Branch changed

PR #3818, opened 2023-06-16
Add GitHub Actions Accessibility alt text bot

colinrotherham:
Suggestion from @davidc-gds 🙌 

See [**Make your GitHub projects more accessible with accessibility-alt-text-bot**](https://github.blog/2023-06-12-make-your-github-projects-more-accessible-with-accessibility-alt-text-bot/) from the GitHub Blog

### DIFF
--- a/.github/workflows/accessibility-bot.yml
+++ b/.github/workflows/accessibility-bot.yml
@@ -1,0 +1,31 @@
+name: Accessibility bot
+
+on:
+  issues:
+    types: [opened, edited]
+
+  pull_request:
+    types: [opened, edited]
+
+  issue_comment:
+    types: [created, edited]
+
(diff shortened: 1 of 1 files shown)


davidc-gds:
@colinrotherham do you know if we need to make this same update to each of our repos to get it to work? In other words, should we do this for the ['govuk-design-system' repo](https://github.com/alphagov/govuk-design-system) as well?

colinrotherham:
> @colinrotherham do you know if we need to make this same update to each of our repos to get it to work? In other words, should we do this for the ['govuk-design-system' repo](https://github.com/alphagov/govuk-design-system) as well?

Yeah we will do. Good to try it out her first and see what happens

PR #3829, opened 2023-06-19
Set up webpack example to handle assets itself

colinrotherham:
This PR updates our webpack example to add:

1. Source maps for Sass in CSS output
2. Assets (images, fonts) automatically extracted by usage (not copied)
3. Directory structure consistency `assets/stylesheets/javascripts`

It was sparked by recent Dependabot activity for webpack so we keep our example up to date:

* https://github.com/alphagov/govuk-frontend/pull/3565
* https://github.com/alphagov/govuk-frontend/pull/3667
* https://github.com/alphagov/govuk-frontend/pull/3706
* https://github.com/alphagov/govuk-frontend/pull/3750
* https://github.com/alphagov/govuk-frontend/pull/3796
* https://github.com/alphagov/govuk-frontend/pull/3822


docs/examples/webpack/webpack.config.js
```diff
@@ -83,25 +100,29 @@ module.exports = ({ WEBPACK_SERVE }, { mode }) => ({
...
   plugins: [
+    new MiniCssExtractPlugin({
+      filename: 'stylesheets/[name].min.css'
+    }),
     new CopyPlugin({
       patterns: [
-        join(srcPath, 'index.html'),
-        {
-          context: join(frontendPath, './src/govuk/assets'),
-          from: '{fonts,images}/**',
```
colinrotherham, line 98: Webpack `css-loader` now automatically scans `url()` etc to extract the necessary assets

Previously we copied everything

docs/examples/webpack/webpack.config.js
```diff
@@ -17,20 +18,19 @@ module.exports = ({ WEBPACK_SERVE }, { mode }) => ({
   devServer: {
     watchFiles: {
       paths: ['**/*.html']
-    },
-    static: {
```
colinrotherham, line 21: By letting webpack handle assets itself they never need writing to disk during development

docs/examples/webpack/webpack.config.js
```diff
@@ -83,25 +100,29 @@ module.exports = ({ WEBPACK_SERVE }, { mode }) => ({
...
+        join(srcPath, 'index.html')
       ]
     })
   ],
 
+  resolve: {
+    alias: {
+      '/assets/': join(frontendPath, './src/govuk/assets/')
+    }
+  },
```
romaricpascal, line 124: *question*: What's the role of this `alias`?

colinrotherham: It's a webpack thing so it knows where assets (like fonts) are:

```css
src: url(/assets/fonts/light-94a07e06a1-v2.woff2)
```

It treats CSS as a module, just like JavaScript

colinrotherham: So you can turn on things like "asset fingerprinting" or removed unused assets from build output

romaricpascal: Sounds good, cheers for clarifying 😊 If I follow:
- sass-loader compiles the URL of the assets to `/assets/...`, because of our default `$govuk-assets-path`
- css-loader picks up these URLs, but needs to understand what `/assets/` mean so we need the alias 😊 

Is that it?

colinrotherham: Yep! We then have to extract it into a file, as the default will be CSS-in-JS otherwise

Issue #3685, opened 2023-05-24
Upload file component has no visual target area when dragging and dropping a file


CharlotteDowns:
This issue is from May 2023 external accessibility audit report.

**Usability issue**

**Issue ID:**
DAC_Usability_Upload_File_01

**URLs:**
[https://govuk-frontend-pr-3503.herokuapp.com/full-page-examples/upload-your-photo-success](https://govuk-frontend-pr-3503.herokuapp.com/full-page-examples/upload-your-photo-success)

## Screen shots
<img width="553" alt="Upload file component" src="https://github.com/alphagov/govuk-frontend/assets/56260216/f5c7b7f6-56ef-404f-ad39-4fd2f974d171">

There is no indication provided for users to know that they can drag and drop files onto the page, as well as when doing so, any indication of where they can drop the file, possibly resulting in the file being dropped in an incorrect location and not being successfully added for uploading.

## Current code ref(s)
`#photo`

## Cognitive user comments
> “Drag and drop works for photo upload. However, there is no text information or box that appears to tell me where to drop it while I am in the process of dragging a photo on the page. I would normally expect that, as this is what usually tells me that a photo can be uploaded in this way. Without it, this functionality would go overlooked by myself and other users unless it had something to tell me this functionality was present”

## Solution
Some users rely on visual indication to convey functionality of components to aid them in understanding what is available to them presently on the page. Consider providing some kind of indication on the page that the drag and drop feature is available, and when doing so that there is also an indication of where to drop the file.

querkmachine:
This is the user agent's native form control. Pretty much all UA's support dragging and dropping of files into inputs, though I'm not aware of any that provide a visual affordance when doing so.

We could try rolling our own version of a file upload, but that's not a small ask as we'd likely want to reproduce a fair amount of the native functionality—a visual button, drag and drop, showing the name of the selected file, the option to select multiple files simultaneously, and potentially the `accept` and `capture` attributes.

We would be investing time and resources into producing something that will, optimistically, equal the functionality of what already exists, and has the potential to be worse. Although I've made a very basic spike to provide a visual affordance on top of the native input, I don't think there's that much value in changing the current component.

querkmachine:
Comment from @davidc-gds on Slack, copied here for posterity: 

> I feel like this one is more of a “add evidence to our backlog item on file upload, and consider making a better file uploader in the future”?

querkmachine:
[...]

PR #3832, opened 2023-06-19
[SPIKE] Dropzone interface for file upload component

querkmachine:
One of the items flagged in the recent audit of Frontend was the lack of a user interface indicating that files could be dragged and dropped into the [File Upload component](https://design-system.service.gov.uk/components/file-upload/).

- #3685

Putting aside that our file upload component is the browser's native control with only light style customisations (and the usability issue is probably the onus of browser vendors, not us), I figured I'd do a technical spike of how we could do something about it in Frontend. 

This is principally a technical prototype. It is intentionally very basic and ugly, and probably not as accessible as it could be, but it does (basically) what it says on the tin.

### Initial state

Appears as a standard, native file input.

image

### With a file dragged over the viewport

A previously hidden div with a dashed grey border and instructions that the user can drop a file here are revealed. The native file input is hidden.

image

### With a file dragged over the dropzone

The div with instructions gains a more prominently coloured border when the user's cursor is within the bounds of it, indicating that it is safe to drop the file.

image

### After a file is dropped

The component returns to only showing the native file input, with the value correctly set.

image

### DIFF
--- a/packages/govuk-frontend/src/govuk/all.mjs
+++ b/packages/govuk-frontend/src/govuk/all.mjs
@@ -5,6 +5,7 @@ import { CharacterCount } from './components/character-count/character-count.mjs
 import { Checkboxes } from './components/checkboxes/checkboxes.mjs'
 import { Details } from './components/details/details.mjs'
 import { ErrorSummary } from './components/error-summary/error-summary.mjs'
+import { FileUpload } from './components/file-upload/file-upload.mjs'
 import { Header } from './components/header/header.mjs'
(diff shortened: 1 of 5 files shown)


owenatgov:
I'm aware this is in draft however an initial thought: Could the "drag here" UI be on the component by default? That would save the need for js at all. Probably needs a designer to sus this out but this is still just a spike so it's not urgent.

querkmachine:
@owenatgov It absolutely could, that's how I've built them in the past when on greenfield projects. In this case, I was looking to avoid potential breaking changes, but I suppose that's a practicality I shouldn't worry about too much in a spike.

PR #3771, opened 2023-06-08
Migrate JavaScript to ES2015 classes

colinrotherham:
Now we've added `<script type="module">` we can use ES2015 (or ES6) classes for components in https://github.com/alphagov/govuk-frontend/issues/3506

Make sure you view [Files changed](https://github.com/alphagov/govuk-frontend/pull/3771/files?w=1) with "Hide whitespace" ticked as class methods are mostly only indented

Files changed with hide whitespace

PR #3833, opened 2023-06-20
Migrate to `import { initAll }` from ES module bundle

colinrotherham:
This PR follows our switch to ES modules:

* https://github.com/alphagov/govuk-frontend/pull/3769

It changes our review app to use ES module `import` not `window.GOVUKFrontend.*` for #3508 

### Before
We include our JavaScript as an IIFE via `<script type="module">`

```html
<script type="module" src="/javascripts/all.min.js"></script>
<script type="module">window.GOVUKFrontend.initAll()</script>
```

### After
We import our JavaScript as an ES module bundle via `<script type="module">`

```html
<script type="module" src="/javascripts/all.bundle.min.mjs"></script>
<script type="module">
  import { initAll } from '/javascripts/all.bundle.min.mjs'
  initAll()
</script>
```

Whilst the first `<script>` is optional we've kept it for the browser preload scanner ([see comment](https://github.com/alphagov/govuk-frontend/pull/3833#discussion_r1234843270))

### DIFF
--- a/packages/govuk-frontend-review/src/views/layouts/_generic.njk
+++ b/packages/govuk-frontend-review/src/views/layouts/_generic.njk
@@ -15,6 +15,9 @@
 {% set mainClasses = 'govuk-main-wrapper--auto-spacing' %}
 
 {% block bodyEnd %}
-  <script type="module" src="/javascripts/all.min.js"></script>
-  <script type="module">window.GOVUKFrontend.initAll()</script>
+  <script type="module" src="/javascripts/all.bundle.min.mjs"></script>
+  <script type="module">
+    import { initAll } from '/javascripts/all.bundle.min.mjs'
+    initAll()
+  </script>
 {% endblock %}
--- a/babel.config.js
+++ b/babel.config.js
@@ -8,6 +8,9 @@ module.exports = function (api) {
 
   const presets = [
     ['@babel/preset-env', {
+      exclude: [
+        'transform-dynamic-import'
+      ],
       targets: {
         node: 'current'
       }
--- a/package-lock.json
+++ b/package-lock.json
@@ -27789,6 +27789,7 @@
       "devDependencies": {
         "@axe-core/puppeteer": "^4.7.3",
         "cheerio": "^1.0.0-rc.12",
+        "govuk-frontend": "*",
         "govuk-frontend-config": "*",
         "govuk-frontend-lib": "*",
         "jest-axe": "^7.0.1",
--- a/packages/govuk-frontend-review/rollup.config.mjs
+++ b/packages/govuk-frontend-review/rollup.config.mjs
@@ -14,7 +14,7 @@ export default defineConfig(({ i: input }) => ({
    * Output options
    */
   output: {
-    format: 'iife',
+    format: 'es',
 
     /**
      * Output plugins
--- a/packages/govuk-frontend-review/src/app.mjs
+++ b/packages/govuk-frontend-review/src/app.mjs
@@ -112,7 +112,7 @@ export default async () => {
     const componentName = req.params.componentName
     const exampleName = req.params.exampleName || 'default'
 
-    const previewLayout = res.locals.componentData?.previewLayout || 'layout'
+    const previewLayout = res.locals.componentData?.previewLayout
 
     const exampleConfig = res.locals.componentData?.examples.find(
       example => example.name.replace(/ /g, '-') === exampleName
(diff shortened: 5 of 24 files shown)


packages/govuk-frontend-review/src/views/layouts/_generic.njk
```diff
@@ -4,6 +4,10 @@
   <link rel="stylesheet" href="/stylesheets/app.min.css">
 
   {% block styles %}{% endblock %}
+
+  <script type="importmap">
+    { "imports": { "govuk-frontend": "/javascripts/all.bundle.min.mjs" } }
```
colinrotherham, line 9: An import map has been added to resolve `govuk-frontend` to `/javascripts/all.bundle.min.mjs` when run inside Puppeteer `page.evaluate()`

packages/govuk-frontend-review/src/views/layouts/_generic.njk
```diff
@@ -15,6 +19,9 @@
 {% set mainClasses = 'govuk-main-wrapper--auto-spacing' %}
 
 {% block bodyEnd %}
-  <script type="module" src="/javascripts/all.min.js"></script>
-  <script type="module">window.GOVUKFrontend.initAll()</script>
+  <script type="module" src="/javascripts/all.bundle.min.mjs"></script>
```
colinrotherham, line 22: Keeping the `<script>` tag allows [browser preload scanners](https://web.dev/preload-scanner/) to download our ES module bundle early

It's optional and can be replaced with [`modulepreload`](https://developer.chrome.com/blog/modulepreload/) in future (Safari 17+)

```html
<link rel="modulepreload" href="/javascripts/all.bundle.min.mjs">
```

colinrotherham:
@romaricpascal We're 100% "double download" free in:

Safari 11.1
Safari 12.1

☝️ Both browsers show concurrent download alongside the CSS

Additional checks in Internet Explorer 11 and Safari 10.1 show component JavaScript `init()` is skipped

packages/govuk-frontend-review/src/views/layouts/_generic.njk
romaricpascal: **question** Is that importmap used outside of the test boilerplate? If not, I'd suggest moving it to that template specifically (even if it lives at the top of the `<body>` tag) as it's a feature to help with the tests 😊

colinrotherham: It's used by `renderAndInitialise()` via [/tests/boilerplate](http://govuk-frontend-review.herokuapp.com/tests/boilerplate)

https://github.com/alphagov/govuk-frontend/blob/bf728fc3876fe942e61356815323d37c9770717b/shared/helpers/puppeteer.js#L101

(Few other tests also use boilerplate)

But all the globals tests (now "exports" rather than globals) use the [home page `/`](http://govuk-frontend-review.herokuapp.com/)

https://github.com/alphagov/govuk-frontend/blob/bf728fc3876fe942e61356815323d37c9770717b/packages/govuk-frontend/src/govuk/components/globals.test.mjs#L52

https://github.com/alphagov/govuk-frontend/blob/bf728fc3876fe942e61356815323d37c9770717b/packages/govuk-frontend/src/govuk/components/globals.test.mjs#L20

https://github.com/alphagov/govuk-frontend/blob/bf728fc3876fe942e61356815323d37c9770717b/packages/govuk-frontend/src/govuk/components/globals.test.mjs#L12

Let me just double check they'd all work from the boilerplate too

colinrotherham: @romaricpascal Yeah all fine, I've moved it to the boilerplate only

romaricpascal (approved):
Nice to totally take advantage of the ES modules for `import`ing our API 🥳  

I think the review app and its tests are in a bit of a wonky state, though, as it neither follows:

1. an `import` version of https://frontend.design-system.service.gov.uk/importing-css-assets-and-javascript/#add-the-javascript-file-to-your-html, where we would serve `node_modules/govuk-frontend/dist/govuk/all.mjs` without bundling
2. https://frontend.design-system.service.gov.uk/importing-css-assets-and-javascript/#import-javascript-using-a-bundler, where the bundled file would be the one initialising the components

So we're not quite testing that GOV.UK Frontend does what it's meant to in our tests, more that our review app bundle does what GOV.UK Frontend is meant to do (in a closer way than before so that's a win, though 🥳 ).

Let's merge this step forward and I'll add a note on https://github.com/alphagov/govuk-frontend/issues/3508 to update how the review app serves/bundles govuk-frontend so we do test govuk-frontend and not the review app bundle (I'd say in favour of 1 so we actually test `govuk-frontend` through the review app).

PR #3834, opened 2023-06-20
[SPIKE] Babel with ES2018 async `init()` error handling

colinrotherham:
Spike to consider async `init()` and `initAll()` with "catchable" error messages

### DIFF
--- a/packages/govuk-frontend/src/govuk/all.mjs
+++ b/packages/govuk-frontend/src/govuk/all.mjs
@@ -17,8 +17,9 @@ import { Tabs } from './components/tabs/tabs.mjs'
  * components provided as part of GOV.UK Frontend.
  *
  * @param {Config} [config] - Config for all components
+ * @returns {Promise<void>}
  */
-function initAll (config) {
+async function initAll (config) {
   config = typeof config !== 'undefined' ? config : {}
 
   // Skip initialisation when GOV.UK Frontend is not supported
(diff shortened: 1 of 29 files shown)


packages/govuk-frontend/src/govuk/all.mjs
```diff
@@ -17,71 +17,79 @@ import { Tabs } from './components/tabs/tabs.mjs'
...
+    initList.push(new SkipLink($skipLink).init())
   }
 
   const $tabs = $scope.querySelectorAll('[data-module="govuk-tabs"]')
   $tabs.forEach(($tabs) => {
-    new Tabs($tabs).init()
+    initList.push(new Tabs($tabs).init())
   })
+
+  await Promise.all(initList)
```
romaricpascal, line 92: This will resolve at the first component breaking, potentially hiding other component breakages (well there'll still be some logging in the console). Given [the support for `allSettled`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled), we could have a little wrapper of our own there to properly catch the breakages of all components. We'll probably want some way to make sure users can access which element and kind of component was trying to be initialised (but that's likely down to the Error object that would be thrown to hold that).

colinrotherham: I love how you added this comment 😆

I'm still waiting for your PR to `allSettled` all the things 🙌 

(But yep!)

colinrotherham: Post-spike, if we make this production ready, we'd return something useful

We talked about a component registry at some point

Probably a good thing to return?

PR #3835, opened 2023-06-20
Add colour value for Prime Minister's office

chao-xian:
Currently this is being hard coded on GOV.UK[1] and needs to be consumed from GOV.UK Frontend as we have for all other departments.

1. https://github.com/alphagov/govuk_publishing_components/pull/3143


chao-xian:
From talking to @owenatgov while this is purely for GOV.UK's use, we might as well just leave it out of GOV.UK Frontend for the time being.

PR #3838, opened 2023-06-21
Update post publish release docs with note on updating aliases

owenatgov:
Done to account for the addition of aliases in https://github.com/alphagov/govuk-design-system/pull/2524 as a way to remind team members to review this following releases.


colinrotherham:
Possibly related but see this page:

**Confirm a phone number**
https://design-system.service.gov.uk/patterns/confirm-a-phone-number/

It includes content for:

* Check your **phone**
* I do not have access to the **phone**
* …no longer have access to the **phone** used to sign up
* …partially show the user the **phone** number

But lacks aliases for "Telephone" like we did in reverse:

* https://github.com/alphagov/govuk-design-system/pull/584

owenatgov:
@colinrotherham I don't think that's related as this PR is specific to the process of updating the aliases for Upcoming components and patterns as a procedure proposal. However it feels like a gap in our search and something we could get fixed quickly separate to this.

colinrotherham (approved):

PR #3847, opened 2023-06-23
Fix Browsersync review app watch extension

colinrotherham:
Forgot to update Browsersync to watch `*.mjs` in https://github.com/alphagov/govuk-frontend/pull/3833


romaricpascal (approved):
Nice catch! 

**question** Would using `{js,cjs,mjs}` help us not have to worry about things in the long run?

colinrotherham:
Cheers @romaricpascal. Maybe 🤷‍♂️

Issue #3101, opened 2022-12-14
Expand code coverage collection


romaricpascal:
## What

Improve test coverage collection across our tests to get a better view of how much our components are tested. This includes merging coverage of unit tests and in-browser tests together to get a complete view. Ideally, once collected, we'd have a way to surface it in the PRs so the results are clearly visible rather than having to download an archive.

Besides collecting coverage of the JavaScript component there is also a few other areas we may want to look into:
- coverage of our Nujucks macro: how much of their `if` is actually being ran by the tests
- coverage of our CSS: how much of our classes is actually used by our components, do we have some rogue classes in the CSS that are actually no longer used
- coverage of our SCSS variables & mixins, if at all possible?

## Why

We're currently [collecting some code coverage from Jest which gets [attached to the github action](https://github.com/alphagov/govuk-frontend/pull/2923). However, it doesn't cover (🥁) all the code being run during our tests. Code running inside puppeteer, for example, is not collected. Managing to collect coverage more accurately would give us a sense of how much code is actually tested and which areas of our codebase we may need to address.

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [ ] Investigate coverage collection for in-browser tests
- [ ] Investigate coverage collection for Nunjucks (using Istanbul and relying on source-maps maybe?)
- [ ] Investigate coverage collection for CSS
- [ ] Investigate coverage collection for SCSS

romaricpascal:
A few rough notes around code coverage:

## General coverage tooling

Istanbul: oldie but sturdy, it was one of the earliest tool for collecting code coverage. [instruments code to collect which lines and branches get ran while code executes](https://github.com/istanbuljs/nyc#how-istanbul-works) (usually test code). It provides a CLI, [nyc](https://github.com/istanbuljs/nyc), which you use to prefix the command you want to execute: `nyc mocha` for example. Istanbul is [built into Jest](https://istanbul.js.org/integrations/). It can also be used at transpilation time with [babel-plugin-istanbul](https://www.npmjs.com/package/babel-plugin-istanbul). Istanbul not only provides the instrumentation of the code, but [reporters to output the results.](https://istanbul.js.org/docs/advanced/alternative-reporters/)
[...]

PR #3851, opened 2023-06-26
Babel transform only `*.mjs` files for Jest

colinrotherham:
Quick spike to solve to blockers in https://github.com/alphagov/govuk-frontend/pull/3812

1. Puppeteer needs to _keep_ [dynamic `import()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import) to load in a browser
2. Jest needs to _replace_ [dynamic `import()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import) with `require()` for [virtual machine contexts](https://nodejs.org/docs/latest-v18.x/api/vm.html#vm-executing-javascript)

By running Babel only on `*.mjs` ES modules we can use [dynamic `import()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import) in `*.js` CommonJS

**Update:** Gave it a couple of runs on Windows to confirm it works:

* https://github.com/alphagov/govuk-frontend/actions/runs/5379997972
* https://github.com/alphagov/govuk-frontend/actions/runs/5380112102

### DIFF
--- a/jest.config.mjs
+++ b/jest.config.mjs
@@ -32,10 +32,10 @@ const config = {
    */
   testEnvironment: 'jest-environment-node-single-context',
 
-  // Enable Babel transforms until Jest supports ESM
+  // Enable Babel transforms until Jest supports ESM and `import()`
   // See: https://jestjs.io/docs/ecmascript-modules
   transform: {
-    '^.+\\.m?js$': ['babel-jest', { rootMode: 'upward' }]
+    '^.+\\.mjs$': ['babel-jest', { rootMode: 'upward' }]
   },
 
   // Enable Babel transforms for ESM-only node_modules
--- a/jest.config.mjs
+++ b/jest.config.mjs
@@ -104,7 +104,7 @@ export default {
         '!**/*.unit.test.{js,mjs}',
 
         // Exclude other tests
-        '!**/components/globals.test.mjs',
+        '!**/components/globals.test.js',
         '!**/components/*/**',
         '!**/tasks/build/**'
       ],
@@ -118,7 +118,7 @@ export default {
       displayName: 'JavaScript component tests',
       testEnvironment: 'govuk-frontend-helpers/jest/environment/puppeteer.mjs',
       testMatch: [
-        '**/components/globals.test.mjs',
+        '**/components/globals.test.js',
         '**/components/*/*.test.{js,mjs}',
 
         // Exclude accessibility/macro/unit tests
--- a/jest.config.mjs
+++ b/jest.config.mjs
@@ -60,6 +60,7 @@ const config = {
  */
 export default {
   collectCoverageFrom: ['./packages/govuk-frontend/src/**/*.{js,mjs}'],
+  coverageProvider: 'v8',
 
   // Reduce CPU usage during project test runs
   maxWorkers: headless
--- a/jest.config.mjs
+++ b/jest.config.mjs
@@ -1,3 +1,6 @@
+import { packageResolveToPath } from 'govuk-frontend-lib/names'
+import { replacePathSepForRegex } from 'jest-regex-util'
+
 import jestPuppeteerConfig from './jest-puppeteer.config.js'
 
 // Detect when browser has been launched headless
@@ -35,7 +38,20 @@ const config = {
   // Enable Babel transforms until Jest supports ESM and `import()`
   // See: https://jestjs.io/docs/ecmascript-modules
   transform: {
-    '^.+\\.mjs$': ['babel-jest', { rootMode: 'upward' }]
+    // Transform all `*.mjs` to compatible CommonJS
+    '^.+\\.mjs$': ['babel-jest', {
+      rootMode: 'upward'
+    }],
+
+    // Transform some `*.js` to compatible CommonJS
+    ...Object.fromEntries([
+      'del',
+      'slash'
+    ].map((packagePath) => [
+      replacePathSepForRegex(`${packageResolveToPath(packagePath)}$`), ['babel-jest', {
+        rootMode: 'upward'
+      }]
+    ]))
   },
 
   // Enable Babel transforms for ESM-only node_modules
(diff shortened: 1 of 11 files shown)


romaricpascal:
I like that, that sounds a very “working with how Node works” way to sort ourselves out 🙌🏻 Two little questions:

1. Is it worth linting for it, or would Jest fail with enough clarity that we'd know that it's because it transpiled an `import` to `require` that our tests failed
2. Do `del` and `slash` need to be treated specially because they're dependencies with `"type":"module"` in their package? Would there be other situation we'd want to transpile `.js` files (for ex. some we do author)?

colinrotherham:
Thanks @romaricpascal

## Using ESLint

> 1. Is it worth linting for it, or would Jest fail with enough clarity that we'd know that it's because it transpiled an `import` to `require` that our tests failed

Yeah we could add ESLint rules but you can't identify Puppeteer tests by filename

**Puppeteer tests with `import()` in `*.mjs` would fail** with missing Babel helpers

```console
ReferenceError: _interopRequireWildcard is not defined
pptr:evaluate;Object.%3Canonymous%3E%20(%2FUsers%2Fcolin%2FSites%2FGDS%2Fgovuk-frontend%2Fpackages%2Fgovuk-frontend%2Fsrc%2Fgovuk%2Fcomponents%2Fglobals.test.mjs%3A13%3A29):1:61
```

**Other tests with `import()` in `*.js` would fail** with Jest runtime syntax errors

```console
Jest encountered an unexpected token

Jest failed to parse a file. This happens e.g. when your code or its dependencies use non-standard JavaScript syntax, or when Jest is not configured to support such syntax.
```

## ESM-only packages

> 2. Do `del` and `slash` need to be treated specially because they're dependencies with `"type":"module"` in their package? Would there be other situation we'd want to transpile `.js` files (for ex. some we do author)?

Babel doesn't transpile code in `node_modules` by default so we:

1. Update `transformIgnorePatterns` to allow paths to `del` and `slash`
2. Update `transform` to transpile the full `*.js` path to `del` and `slash`

Yeah they use `"type":"module"` so their `*.js` paths would be skipped if we only transpiling `*.mjs`

colinrotherham:
There are also Jest environments that avoid Node.js VM contexts like:

* https://github.com/nicolo-ribaudo/jest-light-runner#readme

docs/examples/webpack/babel.config.js
```diff
@@ -0,0 +1,18 @@
+/**
+ * Babel config
+ *
+ * @type {import('@babel/core').ConfigFunction}
+ */
+module.exports = function (api) {
+  const browserslistEnv = !api.env('test')
+    ? 'production'
```
colinrotherham, line 8: Just a note to say you'd typically see projects use `development` to run fewer transforms locally (for example, targeting modern browsers only) but we're using `production` here by default

Until we have an [environment named `production`](https://github.com/browserslist/browserslist#configuring-for-different-environments) this will return our default browsers in **.browserslistrc**

jest.config.mjs
```diff
@@ -60,6 +60,7 @@ const config = {
  */
 export default {
   collectCoverageFrom: ['./packages/govuk-frontend/src/**/*.{js,mjs}'],
+  coverageProvider: 'v8',
```
romaricpascal, line 79: **question** Does that affect the coverage of the tests that do run through Babel? I think what I'm worried about is the Babel helpers being considered for the coverage if it is collected via `v8`, but excluded with default setting. I wouldn't block the PR as we're not looking much at coverage at the moment, but it'll be something to dig into if/when we get to https://github.com/alphagov/govuk-frontend/issues/3101 😊

romaricpascal (commented):
Looking good! 🙌🏻 There's a reference to `globals.test.mjs` in `testing-and-linting.md` wich could use updating as part of 84f3263f196586463b14a2a9abc6b603c7e10c3c. And I think it'd be worth highlighting the reason for picking a `.js` file over a `.mjs` (`.mjs` by default, unless you need to do a dynamic `import()` in the file, is that it?). Would you mind adding that, please?

jest.config.mjs
colinrotherham: Yeah a little bit

Coverage from [`coverageProvider: 'v8'`](https://v8.dev/blog/javascript-code-coverage) is generated from transpiled code but it's mapped back to the original source using source maps and will exclude sneaky helpers. [Read about when it launched in Jest v25](https://jestjs.io/blog/2020/01/21/jest-25#v8-code-coverage)

I'm sure we'll see minor differences in coverage percentages. Turning off Babel transforms for `*.js` stops [`babel-plugin-istanbul`](https://github.com/istanbuljs/babel-plugin-istanbul) from instrumenting the code so it's better than losing it entirely?

colinrotherham:
> Looking good! 🙌🏻

Thanks @romaricpascal

>There's a reference to `globals.test.mjs` in `testing-and-linting.md` which could use updating as part of [84f3263](https://github.com/alphagov/govuk-frontend/commit/84f3263f196586463b14a2a9abc6b603c7e10c3c).

✅ Done

>And I think it'd be worth highlighting the reason for picking a `.js` file over a `.mjs` (`.mjs` by default, unless you need to do a dynamic `import()` in the file, is that it?). Would you mind adding that, please?

✅ Done

I've gone with the text below as it's only dynamic `import()` in browsers (e.g. Puppeteer) that we need to call natively without Babel transforms. All other tests using dynamic `import()` have been reverted and work great still

>Tests should be written using ES modules (`*.mjs`) by default, but use CommonJS modules (`*.js`) for tests using browser [`import()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import) in Puppeteer. This avoids Babel transforms until [Jest supports `import()` and ES modules](https://jestjs.io/docs/ecmascript-modules) in future.

jest.config.mjs
romaricpascal: The situation doesn't sound too bad if the `v8` thing takes sourcemaps into account. Like I said, not a blocker, just wanted to understand a bit and flag for #3101 😊

Issue #2060, opened 2020-12-09
Consider larger elements for a bigger touch/target size


joelanman:
## Related component
Anything interactive:
 - Back link
 - Checkboxes
 - Radios
 - Text input

## Context

Larger elements are easy to use, whether that's on a touch device like a phone, or using a mouse or other input device.

GOVUK Frontend elements have always been nice and big, but they don't quite meet the WCAG standard of at least 44px x 44px. Apple and Android also have 44px 44px as their standard.

https://adrianroselli.com/2019/06/target-size-and-2-5-5.html

In WCAG this is an AAA requirement, and we only aim for AA normally, but I think this would be a good one to consider.

36degrees:
Overlaps a little with #1817

36degrees:
If we were going to change the visual height of some form controls I think we'd also want to include the button component in that list.

However, I'd suggest that we consider the target size of back links as part of #1817 instead as we'd also need to look at at least breadcrumbs, if not also the other standalone links in the header and footer.

Worth noting also that although checkboxes and radios have a 'visual' size of 40x40px, they do have a target size of 44x44px, including the smaller checkboxes and radios.

![Screenshot 2020-12-14 at 15 40 14](https://user-images.githubusercontent.com/121939/102101571-b451c500-3e22-11eb-8d21-4df9bd52ac8c.png)
![Screenshot 2020-12-14 at 15 40 25](https://user-images.githubusercontent.com/121939/102101569-b451c500-3e22-11eb-976a-f7c3dcc3997a.png)

Finally, there's some confusion around [Apple's Human Interface guidelines specifying 44**pt** rather than 44**px**](https://developer.apple.com/design/human-interface-guidelines/ios/visual-design/adaptivity-and-layout/) but I believe that's their way of specifying sizes in a resolution-independent way (e.g. 1024x1366 pt = 2048x2732 px @2x) and broadly equivalent (?) to CSS pixels as defined in WCAG.

Ayesha-S-DWP:
Following the inclusion of the 'Back' links for the AAA consideration, it could also make sense to include the 'Change' links on the 'Check your answers' summary. 

<img width="592" alt="Screen Shot 2020-11-06 at 15 00 58" src="https://user-images.githubusercontent.com/78350425/106445406-e4a6fe80-6476-11eb-851d-63f4a47766b4.png">

querkmachine:
Some relevant asides: 

[...]

PR #3862, opened 2023-06-26
Fix focus style being overlapped by summary action links

### LINKED ISSUE #3841: WCAG 2.2: Summary list ‖ Focus Not Obscured (Enhanced)
## What

In what might be the most pedantic failure of the entire audit: When a summary list row contains enough action links that they wrap onto multiple lines, the dividing line between links can slightly overlap the focus styles of the links on the line above. 

Zoomed-in screenshot of a focused summary list action link. The grey dividing line from the row below is slightly overlapping the bottom of the summary action link.

### Resources

- [Summary list component](https://design-system.service.gov.uk/components/summary-list/)
- [WCAG 2.2 criterion for 'Focus Not Obscured (Enhanced)'](https://www.w3.org/WAI/WCAG22/Understanding/focus-not-obscured-enhanced.html)

## Why

The new WCAG 2.2 Level AAA criterion 'Focus Not Obscured (Enhanced)' says: 

> When a [user interface component](https://www.w3.org/WAI/WCAG22/Understanding/focus-not-obscured-enhanced.html#dfn-user-interface-component) receives keyboard focus, no part of the component is hidden by author-created content.

This slight overlap is a very technical failure of 'Focus Not Obscured (Enhanced)', which does not permit **any** part of the focus style to be hidden.

It seems like a easy fix, but if it's not, this is probably minor enough to safely ignore. 

## Who needs to work on this
- Developers

## Who needs to review this
- Developers
- Designers (potentially)

## Done when
- [x] The summary list has been amended so that the overlap does not take place

querkmachine:
Fixes the WCAG 2.2 issue of subsequent action links in the summary list component slightly overlapping the currently focused link.

Resolves #3841. 

## Screenshots

||Before|After|
|:-|:-:|:-:|
|Narrow layout|A long list of action links. The second, 'use', is focused. A small grey line is partially overlapping it from a link on the line below.|A long list of action links. The second, 'use', is focused. The link is unobscured.|
|Wide layout|A long list of action links. The eighth, 'upgrade', is focused. A small grey line is partially overlapping it from a link on the line below.|A long list of action links. The eighth, 'upgrade', is focused. The link is unobscured.|

## Changes
- Summary list action links now receive the `isolation: isolate` style when focused. This creates a new [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_positioned_layout/Understanding_z-index/Stacking_context), making it appear 'above' the surrounding content.

## Thoughts
- There are a few ways that creating a new stacking context could have been done. `position: relative`, `transform: translate(0, 0)`, and others. I've used the `isolation` property as it is intended specifically for this purpose, on the assumption this fix doesn't need to support IE. ([Browser support for `isolation`](https://caniuse.com/mdn-css_properties_isolation))
- This issue doesn't affect the action links on summary cards, as they are already separated using `row-gap`, preventing the overlap from occuring.
  - Presumably we don't want to do the same thing here, as the action link line spacing is consistent with the content next to it and we probably don't want to change that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -253,6 +253,7 @@ This change was introduced in [pull request #3773: Omit the value attribute from
 We’ve made fixes to GOV.UK Frontend in the following pull requests:
 
 - [#3791: Refactor mobile menu button label/text handling](https://github.com/alphagov/govuk-frontend/pull/3791)
+- [#3862: Fix focus style being overlapped by summary action links](https://github.com/alphagov/govuk-frontend/pull/3862)
 
 ## 4.6.0 (Feature release)
 
--- a/packages/govuk-frontend/src/govuk/components/summary-list/_index.scss
+++ b/packages/govuk-frontend/src/govuk/components/summary-list/_index.scss
@@ -131,6 +131,15 @@
     }
   }
 
+  // Large groups of action links may wrap onto multiple lines. Because the link
+  // focus styles are applied outside of the link's bounding box, there are
+  // situations where the focus style on a link can be overlapped by subsequent
+  // links. We don't want this, so let's create a new stacking context on focus
+  // so the link always appears to be 'on top'.
+  .govuk-summary-list__actions-list-item .govuk-link:focus {
+    isolation: isolate;
+  }
+
   // No border on entire summary list
   .govuk-summary-list--no-border {
     .govuk-summary-list__row {


colinrotherham:
This is great, hadn't seen `isolation` before

Could it have been solved with ["target size"](https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum.html) instead as those link heights are quite low on mobile?

* https://github.com/alphagov/govuk-frontend/issues/2060

querkmachine:
@colinrotherham My presumption is that the current spacing is intentional, as it means the links have a vertical rhythm consistent with the text next to it.

image

If we increase the size or spacing of the links, we lose that bit of visual consistency. 

image

As there aren't usually more than a 2 or 3 action links (this being a contrived, extreme example), the target size issue probably doesn't manifest itself in a problematic way very often. 

The links are also (just barely) large enough to not fall foul of the WCAG 2.2 minimum target size in their current form.

dav-idc:
Cleanup Q: should this issue be in the govuk-frontend 5.0.0 milestone?

dav-idc:
Ah never mind, I see the [issue associated with this pull request](https://github.com/alphagov/govuk-frontend/issues/3841) is in the 5.0.0 milestone 👍 

Here's a full list of the accessibility issues addressed in 5.0, from what I can tell: https://github.com/alphagov/govuk-frontend/issues?q=milestone%3Av5.0+label%3Aaccessibility

Issue #3839, opened 2023-06-22
WCAG 2.2: Investigate the use of focus styles on large target areas


querkmachine:
## What

Some components—notably the accordion and step-by-step navigation—include large clickable regions to expand and collapse each section. 

Currently, these areas turn light grey on hover.

![Screenshot of the accordion component with the first section heading hovered, giving it a light grey background that extends across the entire heading.](https://github.com/alphagov/govuk-frontend/assets/1253214/ad23361a-0e02-4f0e-aa06-221a2cab5f57)

However, the focus style used on these elements does not cover the same area.

![Screenshot of the accordion component with the first section heading focused, giving only the text on the left of the heading a yellow background with a black underline.](https://github.com/alphagov/govuk-frontend/assets/1253214/0508ebea-b29a-4c39-8b5c-af5afdb517d8)

## Why

The WCAG 2.2 Level AAA ['Focus Appearance'](https://www.w3.org/WAI/WCAG22/Understanding/focus-appearance.html) criterion states:

> When the keyboard [focus indicator](https://www.w3.org/WAI/WCAG22/Understanding/focus-appearance.html#dfn-focus-indicator) is visible, an area of the focus indicator meets all the following:
> 
> - is at least as large as the area of a 2 [CSS pixel](https://www.w3.org/WAI/WCAG22/Understanding/focus-appearance.html#dfn-css-pixel) thick [perimeter](https://www.w3.org/WAI/WCAG22/Understanding/focus-appearance.html#dfn-perimeter) of the unfocused component or sub-component, and
> - has a contrast ratio of at least 3:1 between the same pixels in the focused and unfocused states.

The focus style not encompassing the entire clickable area may constitute a failure under this criterion.

Additionally, the current styles may fail the Level AA ['Focus Not Obscured (Minimum)'](https://www.w3.org/WAI/WCAG22/Understanding/focus-not-obscured-minimum.html) and Level AAA ['Focus Not Obscured (Enhanced)'](https://www.w3.org/WAI/WCAG22/Understanding/focus-not-obscured-enhanced) criteria when used in combination with certain assistive technologies, such as screen magnifiers, as the focus-styled elements may not be visible on the user's screen even if other parts of the clickable region are.

## Assumptions

- A screen magnification user may not see the focus style if their magnifier is centred to the right of the heading, causing the text and focus style to not be visible on screen.

## Timebox

3 days

[...]

PR #3866, opened 2023-06-27
[SPIKE] Modify accordion header focus so it includes entire clickable area

querkmachine:
Modifies the accordion component's section header focus styles so that they are applied to the entire clickable area of the header, not just the text inside of it. 

For this investigation:
- #3839

---

The modified component: 
https://govuk-frontend-pr-3866.herokuapp.com/components/accordion

image

---

The unmodified component for comparison:
https://govuk-frontend-review.herokuapp.com/components/accordion

image


querkmachine:
This is now being covered as part of a wider investigation into focus styles, including the accordion.

PR #3867, opened 2023-06-28
Simplify i18n macro to handle single or plural text

colinrotherham:
Split out from the https://github.com/alphagov/govuk-frontend/pull/3812 spike

This PR ensures all component and utility macros can be given a single `params` argument

During component data testing I found it hard to understand how **Character Count** data attributes were rendered, so it also renames `govukPluralisedI18nAttributes()` to `govukI18nAttributes()` to handle single or plural translation text


romaricpascal (approved):
The move to `params` with specific keys and the macro handling both singular and plural makes things clearer than the sometimes call a macro, sometimes not. Nice one! 🙌🏻

PR #3869, opened 2023-06-28
Fix component data typos, missing fields etc

colinrotherham:
This PR fixes various component data typos and missing fields from the https://github.com/alphagov/govuk-frontend/pull/3812 spike

I've added a Gulp watch task to build YAML component data into `fixtures.json` + `macro-options.json` during development. This is useful when running `npm link` with the Design System


romaricpascal (approved):
That's a good few little catches! 🙌🏻 ⛵

PR #3873, opened 2023-06-29
Require only 1 image for royal coat of arms

paulrobertlloyd:
Currently there are 2 versions of the royal coat of arms for single and double pixel density screens, the different images served based on a resolution media query. My belief is that this implementation was chosen as IE8 didn’t support the `background-size` property in CSS.

With IE8 support being dropped, it’s now possible to have a 2x image, and use `background-size` to scale it down by half.

As this is the last place the `govuk-device-pixel-ratio()` mixin is used, I have deprecated this helper.

I’ve also further optimised the large `govuk-crest.png` image. Currently it is 8,884 bytes, the version included in this PR is ~~8,365~~ 7,624 bytes.

Sidenote: I tried using an SVG, but given the complexity of the image, this results in a much larger file size.


packages/govuk-frontend/src/govuk/components/footer/_index.scss
```diff
@@ -73,15 +70,12 @@
 
   .govuk-footer__copyright-logo {
     display: inline-block;
-    min-width: $govuk-footer-crest-image-width;
-    padding-top: ($govuk-footer-crest-image-height + govuk-spacing(2));
+    min-width: $govuk-footer-crest-width;
+    padding-top: ($govuk-footer-crest-height + govuk-spacing(2));
     background-image: govuk-image-url("govuk-crest.png");
```
colinrotherham, line 75: Could you do a quick ImageOptim (or similar) run on the PNG image?

I'm seeing another 8.9% file size saving (which is great)

packages/govuk-frontend/src/govuk/components/footer/_index.scss
```diff
@@ -73,15 +70,12 @@
...
-    @include govuk-device-pixel-ratio {
```
colinrotherham, line 79: I've had a reminder from @36degrees that other service teams likely use `govuk-device-pixel-ratio()` so we'll need to stick with our [**Deprecating a Sass function or mixin**](https://github.com/alphagov/govuk-frontend/blob/main/docs/contributing/managing-change.md#deprecating-a-sass-function-or-mixin) notes to avoid a surprise breaking change

But love that we don't need this now 😊

To help manage that change we can either:

1. Keep the mixin but [add `@deprecated`](https://github.com/alphagov/govuk-frontend/blob/main/docs/contributing/managing-change.md#deprecating-a-sass-function-or-mixin) with a CHANGELOG entry in this PR
2. Keep the mixin but write up a separate GitHub issue to deprecate it in future

colinrotherham:
Thanks @paulrobertlloyd 🙌

Love a good opportunity to remove things

Would you mind splitting out the [Allow Crown copyright notice to be removed](https://github.com/alphagov/govuk-frontend/pull/3873/commits/d0d42978cbef87ad53ed90be777401b4ef18c135) side of things into another PR? Just to let us consider anything copyright/legal related separately

I've leave the Design System squad to have a closer look

packages/govuk-frontend/src/govuk/components/footer/_index.scss
paulrobertlloyd: Oh wow. Hadn’t thought to do that also (I used ImageMagick to reduce colour palette only). Have updated the image, it’s now 7,624 bytes. 😄

paulrobertlloyd:
> Would you mind splitting out the [Allow Crown copyright notice to be removed](https://github.com/alphagov/govuk-frontend/pull/3873/commits/d0d42978cbef87ad53ed90be777401b4ef18c135) side of things into another PR? Just to let us consider anything copyright/legal related separately

Split into a separate PR, #3876

packages/govuk-frontend/src/govuk/components/footer/_index.scss
paulrobertlloyd: Happy to! Given impending v5.0 release, is there an opportunity to deprecate in a prior 4.x release, and remove in v5.0?

If so, I can create a separate issue to remove this mixin and attach it to the 5.0 milestone.

colinrotherham: We've done a "last" (v4.6) release already and next "last last" (v4.7) we'd like to freeze by tomorrow at 5pm. But if you're up for it we can try and take a quick deprecation-only PR against [`support/4.x`](https://github.com/alphagov/govuk-frontend/tree/support/4.x) branch?

It's a bit of a stretch as time's against us but we massively appreciate the contribution

Alternatively we can deprecate in v5.0 and remove in v5.1 instead?

colinrotherham: As an example, you can see how we deprecated [`govuk-not-ie8()` on `support/4.x`](https://github.com/alphagov/govuk-frontend/blob/support/4.x/package/govuk/tools/_ie8.scss#L36) but added new private versions too. Means we could continue using them (internally) until ready to remove entirely:

* https://github.com/alphagov/govuk-frontend/pull/3343

paulrobertlloyd: Challenge accepted! #3877

querkmachine:
To be a little contrarian, I think there's still value in us having 1x and 2x versions of the graphic.

The different scales exist for more than just legacy IE support—although that certainly helped—but to account for screens of different pixel densities. Although 2x (or even 3x and 4x) pixel densities are common on mid- and high-end smartphones and laptops nowadays, the 1x density is still pretty common, especially on desktop PCs. 

Even the not-particularly-cheap, not-particularly-old external monitor I use at home is low enough pixel density to still use the 1x image. 

If we remove the 1x image, we're forcing all devices to download and use the larger image, even if there is no benefit to it doing so. Even though this is only a few kilobytes per load (caching notwithstanding), that feels like it would build up to a lot of unnecessary data usage quite quickly. (Unfortunately device pixel density isn't something GOV.UK collects as part of its analytics, so it's only possible to estimate based on generalisations.)

Personally, I don't think this change serves a need for users, for the environment, or for the data bills of our servers and users.

colinrotherham:
As a bit of related info, we could still remove `govuk-device-pixel-ratio()` in a future PR

[Autoprefixer](https://github.com/postcss/autoprefixer#readme) handles `min-resolution` now so we wouldn't need a mixin for browser support prefixes

### Input
```css
@media (min-resolution: 2dppx) {}
```

### Autoprefixer defaults
```css
@media (-webkit-min-device-pixel-ratio: 2),
  (-o-min-device-pixel-ratio: 2/1),
  (min-resolution: 2dppx) {}
```

### Autoprefixer using [`.browserslistrc`](https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/.browserslistrc)
```css
@media (-webkit-min-device-pixel-ratio: 2),
  (min-resolution: 2dppx) {}
```

querkmachine:
Hi @paulrobertlloyd,

I'm going to close this PR. There are whispers coming through the grapevine that the Royal Arms here might be receiving an update soon(ish, maybe), so we might revisit this work at that time.

Sorry for having this sit around for so long just to say 'not yet'!

paulrobertlloyd:
@beeps Makes sense, and thanks for the update. Quietly hoping the revised coat of arms are simplified to the extent that they can be saved as an SVG file that’s smaller than a PNG. Probably not, but that’d solve a whole number of issues! 🤞

PR #3877, opened 2023-06-29
Deprecate `govuk-device-pixel-ratio` mixin and make private

paulrobertlloyd:
There’s a single place where the `govuk-device-pixel-ratio` mixin is required, and that can be removed in #3873.

Given the breaking 5.0 release, and impending v4.7 release, this PR deprecates this mixin prior to its removal in 5.0.

In the meantime, it also makes this mixin private… not sure if this is the right approach, so happy to be corrected on how best to manage this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+### Deprecated `govuk-device-pixel-ratio` mixin
+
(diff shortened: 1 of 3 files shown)


querkmachine:
Note that we ultimately opted not to include this change in 4.7. 

Instead, it'll likely be pushed into 5.0 (with the removal happening in 6.0), or it may be included in a future 4.x feature release if we end up doing one.

Issue #3688, opened 2023-05-24
The non-visual experience of Exit this Page button doesn't make it clear it's a safety feature


CharlotteDowns:
This issue is from May 2023 external accessibility audit report.

**Usability issue**

**Issue ID:**
DAC_Usability_Exit_Button_01

**URLs:**
[https://govuk-frontend-pr-3503.herokuapp.com/full-page-examples/child-maintenance](https://govuk-frontend-pr-3503.herokuapp.com/full-page-examples/child-maintenance)

## Screen shots
<img width="215" alt="Exit this page red button" src="https://github.com/alphagov/govuk-frontend/assets/56260216/dd2adaac-629c-45d8-b106-318ad674cc84">

The ‘Exit this page’ button did not inform users of screen reading assistive technologies of its emergency purpose with the label. Although this may already be conveyed on a previous page, it does not mean that the user will always remember this information.

## Current code ref(s)
`#main-content > div.govuk-exit-this-page > a`

## Screen reader comments
> “I was able to clearly navigate to and select the ‘Exit this page’ button. The button appeared at the beginning of the main content section making it easy to locate and use. 

> However, although I was aware from the label of the button that I would exit the page it was not immediately clear to me that the item was an emergency exit and that my information would be removed from the page. Placing the phrase ‘Emergency’ within the label for example ‘Emergency exit from this page’ with additional hint information advising me that the page will be removed from browsing history would ensure that all screen reader users are aware of the value of this feature.”

## Solution
Consider adding the term ‘Emergency’ to the button. This can be visually hidden to provide that extra information for users of screen reading assistive technologies, while keeping the current aesthetics.

PR #3878, opened 2023-06-29
Add `html` parameter to Exit this Page button

querkmachine:
In relation to #3688.

## Changes
- Adds a `html` parameter to the EtP component, to complement the existing `text` parameter. This is to allow service teams to provide their own copy containing visually-hidden text.
- Updates tests and parameter documentation to align with these changes.

### DIFF
--- a/src/govuk/components/exit-this-page/exit-this-page.yaml
+++ b/src/govuk/components/exit-this-page/exit-this-page.yaml
@@ -2,7 +2,11 @@ params:
   - name: text
     type: string
     required: false
-    description: Text for the link. Defaults to `Exit this page`.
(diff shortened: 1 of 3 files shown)


owenatgov (commented):
Looks good! Could we please squash the commits? Whilst they more accurately reflect the history of the work at present, I think it might be simpler to reflect the change we're making at this point in time.

owenatgov (approved):
💪🏻

querkmachine:
Have done so. The original version of the commit is here, for future reference around how I handled the default HTML/`html`/`text` parameter logic.
https://github.com/alphagov/govuk-frontend/commit/a5983a06c4ddb13aa0b5548e93f4ec8431733979

PR #3880, opened 2023-06-30
Fix error with child maintenance example

owenatgov:
A minor bug that got missed in a previous review. Swaps out quotes in the phase banner in the example so that it doesn't break the html string.

### DIFF
--- a/app/src/views/full-page-examples/child-maintenance/index.njk
+++ b/app/src/views/full-page-examples/child-maintenance/index.njk
@@ -43,7 +43,7 @@ scenario: >-
     tag: {
       text: "beta"
     },
-    html: "This is a new service – your <a class="govuk-link" href="#">feedback</a> will help us to improve it."
+    html: "This is a new service – your <a class='govuk-link' href='#'>feedback</a> will help us to improve it."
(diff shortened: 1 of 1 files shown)


app/src/views/full-page-examples/child-maintenance/index.njk
```diff
@@ -43,7 +43,7 @@ scenario: >-
     tag: {
       text: "beta"
     },
-    html: "This is a new service – your <a class="govuk-link" href="#">feedback</a> will help us to improve it."
+    html: "This is a new service – your <a class='govuk-link' href='#'>feedback</a> will help us to improve it."
```
colinrotherham, line 46: Ahh nice spot

Would you mind if we always used double quotes in HTML?

Nunjucks doesn't mind single quotes, and we tend to use them when they _enclose_ double quotes

```suggestion
    html: 'This is a new service – your <a class="govuk-link" href="#">feedback</a> will help us to improve it.'
```

owenatgov:
Closing as this is superseded by https://github.com/alphagov/govuk-frontend/pull/3953

PR #3883, opened 2023-07-03
Group Dependabot package updates into shared PRs

colinrotherham:
Sets up Dependabot groups for packages that can be grouped together

Some groups like `@babel/*` and `@percy/*` will avoid compatibility issues
Others are to reduce the noise of all the separate PRs on Mondays


36degrees (approved):
Let's give it a go! 👍🏻

PR #3899, opened 2023-07-03
Update default `aria-label` in Pagination component

### LINKED ISSUE #3684: Pagination `aria-label="results"` does not inform the user of the purpose of the region

**Usability issue**

**Issue ID:**
DAC_Usability_Pagination_02

**URLs:**
[https://govuk-frontend-pr-3503.herokuapp.com/full-page-examples/search-three-pages](https://govuk-frontend-pr-3503.herokuapp.com/full-page-examples/search-three-pages)

## Screen shots
Pagination component options

The pagination region has been marked up within a `<nav>` element. However, this `<nav>` has been provided with an accessible name of ‘results’. This does not inform the user of the purpose of the region and does not convey that it contains the pagination options. As screen reader users will often navigate via regions, it is important that they are labelled adequately.

## Current code ref(s)
`#main-content > form > div.govuk-grid-row > div.govuk-grid-column-two-thirds > nav`

## Solution
All labels should be descriptive and inform users of screen reading assistive technologies of the purpose of the region, allowing them to easily discern what content is contained within the said region. Consider changing the value of the ‘aria-label’ to ‘Results pagination’.

querkmachine:
In the recent audit of GOV.UK Frontend, it was flagged that the default `aria-label` of 'results' did not describe the purpose of the region adequately. The component is not a list of results, but a list of pages. The audit suggested we change the label to 'Results pagination'.

Taking into mind that the pagination component is designed for use in contexts that are not related to showing results—such as the previous/next variation, which is intended for navigating between content pages—I have further simplified this to 'Pagination'.

Although this may now be too generic, something 'generic' strikes me as being a more appropriate default than something specific enough to be incorrect or misleading. I still worry that it might be too much of a technical term to be well understood. Happy to hear other views on this.

Resolves #3684.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -248,6 +248,16 @@ If you need to maintain the existing behaviour, you can set the value to an empt
 
 This change was introduced in [pull request #3773: Omit the value attribute from select options with no value](https://github.com/alphagov/govuk-frontend/pull/3773).
 
+### Suggested changes
+
+#### Update the Pagination component's default `aria-label`
+
+The default value of the Pagination component's `aria-label` has been updated to be more descriptive of the contents of the region. If you are using the component's default label, you may wish to update it to the new value.
+
+You don't need to change anything if you're using the `govukPagination` Nunjucks macro.
+
+This change was introduced in [pull request #3899: Update default `aria-label` in Pagination component](https://github.com/alphagov/govuk-frontend/pull/3899).
+
 ### Fixes
 
 We’ve made fixes to GOV.UK Frontend in the following pull requests:
--- a/packages/govuk-frontend/src/govuk/components/pagination/pagination.yaml
+++ b/packages/govuk-frontend/src/govuk/components/pagination/pagination.yaml
@@ -73,7 +73,7 @@ params:
   - name: landmarkLabel
     type: string
     required: false
-    description: The label for the navigation landmark that wraps the pagination. Defaults to 'results'.
+    description: The label for the navigation landmark that wraps the pagination. Defaults to 'Pagination'.
   - name: classes
     type: string
     required: false
--- a/packages/govuk-frontend/src/govuk/components/pagination/template.njk
+++ b/packages/govuk-frontend/src/govuk/components/pagination/template.njk
@@ -1,6 +1,6 @@
 {% set blockLevel = not params.items and (params.next or params.previous) %}
 
-<nav class="govuk-pagination {{- ' ' + params.classes if params.classes }} {{- ' govuk-pagination--block' if blockLevel }}" role="navigation" aria-label="{{ params.landmarkLabel | default("results") }}" {%- for attribute, value in params.attributes %} {{ attribute }}="{{ value }}" {%- endfor -%}>
+<nav class="govuk-pagination {{- ' ' + params.classes if params.classes }} {{- ' govuk-pagination--block' if blockLevel }}" role="navigation" aria-label="{{ params.landmarkLabel | default("Pagination") }}" {%- for attribute, value in params.attributes %} {{ attribute }}="{{ value }}" {%- endfor -%}>
   {%- if params.previous and params.previous.href -%}
     <div class="govuk-pagination__prev">
       <a class="govuk-link govuk-pagination__link" href="{{ params.previous.href }}" rel="prev" {%- for attribute, value in params.previous.attributes %} {{ attribute }}="{{ value }}" {%- endfor -%}>


36degrees:
> I still worry that it might be too much of a technical term to be well understood. Happy to hear other views on this.

I share your worry, though I've got no data to back it up. Wondering if 'pages' might be better?

Interesting that [Bootstrap suggest 'Search results pages'](https://getbootstrap.com/docs/4.0/components/pagination/#:~:text=an%20appropriate%20label%20could%20be%20aria%2Dlabel%3D%22Search%20results%20pages%22) as an example (introduced in https://github.com/twbs/bootstrap/pull/19704), though as you say it's dependent on context.

davidc-gds:
How about a label thats a bit more of a phrase, rather than "results pagination"?

Something like "Navigation for results pages".

querkmachine:
@davidc-gds Would you still expect that to be used on the [previous/next variant](https://govuk-frontend-review.herokuapp.com/components/pagination/with-prev-and-next-only-and-labels/preview) of the pagination too?

davidc-gds:
> @davidc-gds Would you still expect that to be used on the [previous/next variant](https://govuk-frontend-review.herokuapp.com/components/pagination/with-prev-and-next-only-and-labels/preview) of the pagination too?

@querkmachine I might be missing some context as to why it wouldn't work for the 'previous / next' variant, but at first glance, I feel like the 'previous / next' variant is still a form of navigation for results pages. You use the previous and next links to navigate between results pages, yeah? 🤷 

So a non-confident 'yes' from me on it applying to both variants.

36degrees:
> Something like "Navigation for results pages".

Bear in mind that it's a navigation landmark, so it may be announced as 'Navigation for results pages, navigation'?

'Pages of results'? 🤷🏻‍♂️

querkmachine:
> You use the previous and next links to navigate between results pages, yeah? 🤷

@davidc-gds I didn't think so? If the intended use is the same as demonstrated in [the Publishing Components version](https://components.publishing.service.gov.uk/component-guide/previous_and_next_navigation) then the intent is that it's used to navigate between related pages in a set. You can see it in use [at the bottom of travel advice pages](https://www.gov.uk/foreign-travel-advice/canada/entry-requirements).

In this context it's unrelated to searching or results, it's basically just a faster way of navigating between pages in a table of contents.

davidc-gds:
Ahh thanks @querkmachine  I misinterpreted the use-case for the 'previous and next' pagination, yep. My bad!

After thinking on the above comments from you and @36degrees, I can think of only a few alternatives to 'pagination'.

It might be worth looping in @claireashworth or @calvin-lau-sig7 for a second opinion on the wording.

The ones I've come up with, which avoid the word 'navigation' and also should work okay for the 'next / previous page' variant are:

- pagination (AKA go with what we have)
- page picker
- page selector

I'm personally thinking we just stick with 'pagination' even though it's not an ideal descriptor. Both 'picker' and 'selector' have other meanings or uses in web development.

claireashworth:
Hmm
So this description needs to work for both navigating through pages and a return showing a list of pages? If so, that makes it harder to be accurate and succinct.
Other options which might work are:

- page sequence
- page series
- page order
- page results
- page count
- page list results

But these don't work as well for page navigation.

querkmachine:
I'm personally happy to stick with 'pagination', as although it's a bit of a technical term, it's one that seems to be supported by a number of sources. 

- [W3C Design System](https://design-system.w3.org/components/pagination.html) uses 'pagination'.
- [Accessibility Matters](https://www.a11ymatters.com/pattern/pagination/) uses 'pagination navigation'.
- The next/previous navigation on [GOV.UK search pages](https://www.gov.uk/search/all?keywords=bat&order=relevance) uses 'pagination'.
- [Adam Silver's blog](https://adamsilver.io/blog/) uses 'pagination'.

Although this is a small subset of sites, it certainly seems like 'pagination' is what is recommended by some fairly authoritative groups and people, and it's what existing screen reader users of GOV.UK will have encountered in the past.

davidc-gds:
I'd like to get a thought from @claireashworth to see if just 'pagination' seems fine.

From my perspective, it seems we've done a thorough search for alternative phrases, and nothing seems to work for both the list of page numbers version and the 'next / previous' version.

So I'm personally ok with moving forward with 'pagination', based on that work we did.

colinrotherham (approved):

claireashworth:
Yes, let's try pagination. We can revisit another time if it's causing issues.

PR #3902, opened 2023-07-04
Add visually-hidden text to clarify pagination links

### LINKED ISSUE #3679: Pagination links for 'next' and 'previous' pages should be understandable even when out of context

**WCAG Reference:**
2.4.9 Link Purpose (Link Only) (Level AAA)
[Understanding Link Purpose (Link Only) ](https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-link-only.html)| [How to Meet Link Purpose (Link Only)](https://www.w3.org/WAI/WCAG21/quickref/#link-purpose-link-only)

**Issue ID:**
DAC_Link_Purpose_Out_Of_Context_02

**URLs:**
[https://govuk-frontend-pr-3503.herokuapp.com/full-page-examples/search](https://govuk-frontend-pr-3503.herokuapp.com/full-page-examples/search)

## Screen shots
Pagination example

The ‘Next’ and ‘Previous’ links in the pagination do not possess additional link text of ‘page’ or similar to inform users of screen reading assistive technologies of their purpose when viewing the links out of context using the link text alone.

## Current code ref(s)
`#main-content > form > div.govuk-grid-row > div.govuk-grid-column-two-thirds > nav > div.govuk-pagination__prev > a`

> Adding the phrase ‘Page’ to the next and previous links will enable all screen reader users to make a quick and easy selection.” 

> “When navigating in context and out of context, I noticed two links ‘previous’ and ‘next’ as I did not know what this referred to. I found this to be confusing because of the two links which are not clear as this would make it difficult for users to navigate in context or out of context may not pick up what the links are. It would be helpful if the links previous and next could be labelled as previous page and next page.”

## Solution
Ensure that links are descriptive enough when navigating out of context, allowing for users of screen reading assistive technologies to discern their purpose and functionality when doing so.
Consider adding the text ‘page’ to the ‘next’ and ‘previous’ pagination links ot clearly identify that that is what they are used for.

querkmachine:
Adds visually-hidden text to previous and next links to try and clarify their purpose out of context. Resolves #3679.

## Changes
- Updates the default previous and next link text to include the word 'page' in a visually-hidden span next to them.
- Adds support for `previous.html` and `next.html` so that teams can localise these buttons with visually-hidden text, if needed.

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/pagination/pagination.yaml
+++ b/packages/govuk-frontend/src/govuk/components/pagination/pagination.yaml
@@ -36,7 +36,11 @@ params:
       - name: text
         type: string
         required: false
-        description: The link text to the previous page. Defaults to 'Previous'.
+        description: The text content of the link to the previous page. Defaults to 'Previous page', with 'page' being visually hidden. If `html` is provided, the `text` option will be ignored.
+      - name: html
+        type: string
+        required: false
+        description: The HTML content of the link to the previous page. Defaults to 'Previous page', with 'page' being visually hidden. If `html` is provided, the `text` option will be ignored.
       - name: labelText
         type: string
         required: false
@@ -57,7 +61,11 @@ params:
       - name: text
         type: string
         required: false
-        description: The link text to the next page. Defaults to 'Next'.
+        description: The text content of the link to the next page. Defaults to 'Next page', with 'page' being visually hidden. If `html` is provided, the `text` option will be ignored.
+      - name: html
+        type: string
+        required: false
+        description: The HTML content of the link to the next page. Defaults to 'Next page', with 'page' being visually hidden. If `html` is provided, the `text` option will be ignored.
       - name: labelText
         type: string
         required: false
--- a/packages/govuk-frontend/src/govuk/components/pagination/template.njk
+++ b/packages/govuk-frontend/src/govuk/components/pagination/template.njk
@@ -8,7 +8,11 @@
           <path d="m6.5938-0.0078125-6.7266 6.7266 6.7441 6.4062 1.377-1.449-4.1856-3.9768h12.896v-2h-12.984l4.2931-4.293-1.414-1.414z"></path>
         </svg>
         <span class="govuk-pagination__link-title {{- ' govuk-pagination__link-title--decorated' if blockLevel and not params.previous.labelText }}">
-          {{- params.previous.text | default("Previous") -}}
+          {%- if params.previous.html or params.previous.text -%}
+            {{ params.previous.html | safe if params.previous.html else params.previous.text }}
+          {%- else -%}
+            Previous<span class="govuk-visually-hidden"> page</span>
+          {%- endif -%}
         </span>
         {%- if params.previous.labelText and blockLevel -%}
           <span class="govuk-visually-hidden">:</span>
@@ -47,7 +51,11 @@
           {{- nextArrow | safe -}}
         {%- endif %}
         <span class="govuk-pagination__link-title {{- ' govuk-pagination__link-title--decorated' if blockLevel and not params.next.labelText }}">
-          {{- params.next.text | default("Next") -}}
+          {%- if params.next.html or params.next.text -%}
+            {{ params.next.html | safe if params.next.html else params.next.text }}
+          {%- else -%}
+            Next<span class="govuk-visually-hidden"> page</span>
+          {%- endif -%}
         </span>
         {%- if params.next.labelText and blockLevel -%}
           <span class="govuk-visually-hidden">:</span>


packages/govuk-frontend/src/govuk/components/pagination/pagination.yaml
```diff
@@ -36,7 +36,11 @@ params:
       - name: text
         type: string
         required: false
-        description: The link text to the previous page. Defaults to 'Previous'.
+        description: The text content of the link to the previous page. If `previous.html` is supplied, this is ignored.
```
colinrotherham, line 39: We seem to add the "Defaults to" text in both `html` and `text`?

```suggestion
        description: The text content of the link to the previous page. Defaults to 'Previous page', with 'page' being visually hidden. If `previous.html` is supplied, this is ignored.
```

packages/govuk-frontend/src/govuk/components/pagination/pagination.yaml
```diff
@@ -36,7 +36,11 @@ params:
...
+      - name: html
+        type: string
+        required: false
+        description: The HTML content of the link to the previous page. Defaults to 'Previous page', with 'page' being visually hidden.
```
colinrotherham, line 43: Similar with this one, think we always add the "this is ignored" caveat to both `text` and `html`?

```suggestion
        description: The HTML content of the link to the previous page. Defaults to 'Previous page', with 'page' being visually hidden. If `previous.text` is supplied, this is ignored.
```

packages/govuk-frontend/src/govuk/components/pagination/pagination.yaml
```diff
@@ -57,7 +61,11 @@ params:
       - name: text
         type: string
         required: false
-        description: The link text to the next page. Defaults to 'Next'.
+        description: The text content of the link to the next page. If `next.html` is supplied, this is ignored.
```
colinrotherham, line 64:
```suggestion
        description: The text content of the link to the next page. Defaults to 'Next page', with 'page' being visually hidden. If `next.html` is supplied, this is ignored.
```

packages/govuk-frontend/src/govuk/components/pagination/pagination.yaml
```diff
@@ -57,7 +61,11 @@ params:
...
+      - name: html
+        type: string
+        required: false
+        description: The HTML content of the link to the next page. Defaults to 'Next page', with 'page' being visually hidden.
```
colinrotherham, line 68:
```suggestion
        description: The HTML content of the link to the next page. Defaults to 'Next page', with 'page' being visually hidden. If `next.text` is supplied, this is ignored.
```

colinrotherham: Do we need to clarify that:

1. The `text` default is `'Previous page'`
2. The `html` default is `'Previous<span class="govuk-visually-hidden"> page</span>'`

Same for the **Next page** options

packages/govuk-frontend/src/govuk/components/pagination/pagination.yaml
querkmachine: I'm not sure that we have a convention, personally. We have very few instances where we provide default text **and** have both a `text` and a `html` option. From what I can see skimming through:

* Back Link documents the default (which is plain text) in both `text` and `html` parameters.
* Notification Banner documents the plain text default(s) in `titleText`, omitting them from `titleHtml`. 
* Footer documents the (HTML-containing) defaults in both text and HTML parameters of `contentLicence` and `copyright`.

Personally I think the documentation on what is default should be stuck with whatever parameter is most relevant—`text` if it's plain text, `html` if it contains HTML—especially in cases where it can be quite long, like on the Notification Banner, but it's not something I have strong feelings over.

packages/govuk-frontend/src/govuk/components/pagination/pagination.yaml
querkmachine: I'm not sure if that's useful because it's not technically true, I guess?

If a user managed to override `html` in a way that it registered as being undefined but still bypassed the default (which, to be fair, I don't think is really possible with the current Nunjucks logic), there is no `text` default that would be fallen back upon. The only default is the one containing HTML.

packages/govuk-frontend/src/govuk/components/pagination/pagination.yaml
colinrotherham: I'd stick it in both, knowing that this text might go into editor/IDE autocomplete in future

But yep, we definitely have gaps where we should _also_ put it in both

Probably better to repeat ourselves that assume it's obvious from the context?

packages/govuk-frontend/src/govuk/components/pagination/pagination.yaml
colinrotherham: Not technically, it's hard to explain the behaviour isn't it 😣

For users that want to add `html` we should show the correct HTML "default" so to speak

But it's harder to justify `text` as yep there isn't a default

querkmachine: I'm gonna keep the description of the default prosaic for now, rather than showing inline HTML, as that seems to be the more common way of describing these things (presumably if a person is comfy with code, they could just look at it already). Something else to hash out for the documentation style guide.

packages/govuk-frontend/src/govuk/components/pagination/pagination.yaml
querkmachine: I already added this to the batch, but thinking on it, this probably isn't accurate? 

The `text` parameter states that 'If `previous.html` is supplied, this is ignored.' to indicate that the `html` parameter takes priority over the `text` parameter in situations if both are defined. The only situation where `html` is ignored is if it's empty.

Does that need to be worded more clearly? 🤔

colinrotherham: Yeah weighing everything up, shall we copy what **Summary List** does?

```suggestion
        description: The HTML content of the link to the previous page. Defaults to 'Previous page', with 'page' being visually hidden. If `html` is provided, the `text` option will be ignored.
```

Seems to be the most consistent

packages/govuk-frontend/src/govuk/components/pagination/pagination.yaml
```diff
@@ -57,7 +61,11 @@ params:
       - name: text
         type: string
         required: false
-        description: The link text to the next page. Defaults to 'Next'.
+        description: The text content of the link to the next page. Defaults to 'Next page', with 'page' being visually hidden. If `next.html` is supplied, this is ignored.
+      - name: html
+        type: string
+        required: false
+        description: The HTML content of the link to the next page. Defaults to 'Next page', with 'page' being visually hidden. If this is provided, the `next.text` option will be ignored.
```
colinrotherham, line 68: I think I prefer how **Summary list** calls out "If `html`" rather than "If this"?

```suggestion
        description: The HTML content of the link to the next page. Defaults to 'Next page', with 'page' being visually hidden. If `html` is provided, the `text` option will be ignored.
```

colinrotherham: Using the full "`next.html`" (with parent prefix) versus just "`html`" is definitely more readable but doesn't match the name in the table column—shall we go with just `html`?

(I've looked at auto parent prefixes in https://github.com/alphagov/govuk-design-system/pull/2892)

packages/govuk-frontend/src/govuk/components/pagination/pagination.yaml
```diff
@@ -36,7 +36,11 @@ params:
       - name: text
         type: string
         required: false
-        description: The link text to the previous page. Defaults to 'Previous'.
+        description: The text content of the link to the previous page. Defaults to 'Previous page', with 'page' being visually hidden. If `previous.html` is supplied, this is ignored.
```
colinrotherham, line 39: Sorry, similar comment to https://github.com/alphagov/govuk-frontend/pull/3902/files#r1255539276

Shall we stick to _provided_ wording in both fields, rather than _supplied_?

```suggestion
        description: The text content of the link to the previous page. Defaults to 'Previous page', with 'page' being visually hidden. If `html` is provided, the `text` option will be ignored.
```

PR #3906, opened 2023-07-05
Split up output for the export tests

36degrees:
Run each ‘test’ in a separate step, and promote the setting of the environment to the job so it’s shared between all steps.

Putting each statement onto its own step means that the output from each command is grouped alongside the command that was run, aiding debugging.

## Before

```
[+] Export import, Node.js 12
  ▶ Run node --eval "console.log(require.resolve('govuk-frontend'))" --conditions import
  /home/runner/work/govuk-frontend/govuk-frontend/packages/govuk-frontend/dist/govuk/all.mjs
  /home/runner/work/govuk-frontend/govuk-frontend/packages/govuk-frontend/package.json
  /home/runner/work/govuk-frontend/govuk-frontend/packages/govuk-frontend/dist/govuk/components/accordion/accordion.bundle.js
  /home/runner/work/govuk-frontend/govuk-frontend/packages/govuk-frontend/dist/govuk/components/accordion/accordion.bundle.mjs
  /home/runner/work/govuk-frontend/govuk-frontend/packages/govuk-frontend/dist/govuk/i18n.mjs
```

You can only see all of the commands run if you expand the collapsed run statement:

```
[+] Export import, Node.js 12
  ▼ Run node --eval "console.log(require.resolve('govuk-frontend'))" --conditions import
    node --eval "console.log(require.resolve('govuk-frontend'))" --conditions import
    node --eval "console.log(require.resolve('govuk-frontend/package.json'))" --conditions import
    node --eval "console.log(require.resolve('govuk-frontend/dist/govuk/components/accordion/accordion.bundle.js'))" --conditions import
    node --eval "console.log(require.resolve('govuk-frontend/dist/govuk/components/accordion/accordion.bundle.mjs'))" --conditions import
    node --eval "console.log(require.resolve('govuk-frontend/dist/govuk/i18n.mjs'))" --conditions import
    shell: /usr/bin/bash -e {0}
    env:
      FLAGS:  --conditions import
  /home/runner/work/govuk-frontend/govuk-frontend/packages/govuk-frontend/dist/govuk/all.mjs
  /home/runner/work/govuk-frontend/govuk-frontend/packages/govuk-frontend/package.json
  /home/runner/work/govuk-frontend/govuk-frontend/packages/govuk-frontend/dist/govuk/components/accordion/accordion.bundle.js
  /home/runner/work/govuk-frontend/govuk-frontend/packages/govuk-frontend/dist/govuk/components/accordion/accordion.bundle.mjs
  /home/runner/work/govuk-frontend/govuk-frontend/packages/govuk-frontend/dist/govuk/i18n.mjs
```

## After

```
[+] Run node --eval "console.log(require.resolve('govuk-frontend'))" --conditions import
  ▶ Run node --eval "console.log(require.resolve('govuk-frontend'))" --conditions import
  /home/runner/work/govuk-frontend/govuk-frontend/packages/govuk-frontend/dist/govuk/all.mjs

[+] Run node --eval "console.log(require.resolve('govuk-frontend/package.json'))" --conditions import
  ▶ Run node --eval "console.log(require.resolve('govuk-frontend/package.json'))" --conditions import
  /home/runner/work/govuk-frontend/govuk-frontend/packages/govuk-frontend/package.json

[+] Run node --eval "console.log(require.resolve('govuk-frontend/dist/govuk/components/accordion/accordion.bundle.js'))" --conditions import
  ▶ Run node --eval "console.log(require.resolve('govuk-frontend/dist/govuk/components/accordion/accordion.bundle.js'))" --conditions import
  /home/runner/work/govuk-frontend/govuk-frontend/packages/govuk-frontend/dist/govuk/components/accordion/accordion.bundle.js

[+] Run node --eval "console.log(require.resolve('govuk-frontend/dist/govuk/components/accordion/accordion.bundle.mjs'))" --conditions import
  ▶ Run node --eval "console.log(require.resolve('govuk-frontend/dist/govuk/components/accordion/accordion.bundle.mjs'))" --conditions import
  /home/runner/work/govuk-frontend/govuk-frontend/packages/govuk-frontend/dist/govuk/components/accordion/accordion.bundle.mjs

[+] Run node --eval "console.log(require.resolve('govuk-frontend/dist/govuk/i18n.mjs'))" --conditions import
  ▶ Run node --eval "console.log(require.resolve('govuk-frontend/dist/govuk/i18n.mjs'))" --conditions import
  /home/runner/work/govuk-frontend/govuk-frontend/packages/govuk-frontend/dist/govuk/i18n.mjs
```

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -240,6 +240,11 @@ jobs:
           - conditions: import
             node-version: 12.18
 
+    env:
+      # Node.js conditions override from "require" to "import" etc
+      # https://nodejs.org/api/cli.html#-c-condition---conditionscondition
+      FLAGS: ${{ matrix.conditions != 'require' && format(' --conditions {0}', matrix.conditions) || '' }}
+
     steps:
       - name: Checkout
         uses: actions/checkout@v3
@@ -255,18 +260,11 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
 
-      - name: Export ${{ matrix.conditions }}, Node.js ${{ matrix.node-version }}
-        env:
-          # Node.js conditions override from "require" to "import" etc
-          # https://nodejs.org/api/cli.html#-c-condition---conditionscondition
-          FLAGS: ${{ matrix.conditions != 'require' && format(' --conditions {0}', matrix.conditions) || '' }}
-
-        run: |
-          node --eval "console.log(require.resolve('govuk-frontend'))"${{ env.FLAGS }}
-          node --eval "console.log(require.resolve('govuk-frontend/package.json'))"${{ env.FLAGS }}
-          node --eval "console.log(require.resolve('govuk-frontend/dist/govuk/components/accordion/accordion.bundle.js'))"${{ env.FLAGS }}
-          node --eval "console.log(require.resolve('govuk-frontend/dist/govuk/components/accordion/accordion.bundle.mjs'))"${{ env.FLAGS }}
-          node --eval "console.log(require.resolve('govuk-frontend/dist/govuk/i18n.mjs'))"${{ env.FLAGS }}
+      - run: node --eval "console.log(require.resolve('govuk-frontend'))"${{ env.FLAGS }}
+      - run: node --eval "console.log(require.resolve('govuk-frontend/package.json'))"${{ env.FLAGS }}
+      - run: node --eval "console.log(require.resolve('govuk-frontend/dist/govuk/components/accordion/accordion.bundle.js'))"${{ env.FLAGS }}
+      - run: node --eval "console.log(require.resolve('govuk-frontend/dist/govuk/components/accordion/accordion.bundle.mjs'))"${{ env.FLAGS }}
+      - run: node --eval "console.log(require.resolve('govuk-frontend/dist/govuk/i18n.mjs'))"${{ env.FLAGS }}
 
   regression:
     name: Percy


colinrotherham:
Thanks for doing this, makes it much more obvious what failed

Is it worth letting GitHub Actions generate all the possible combinations for us?

```yaml
      matrix:
        node-version:
          - 12 # Node.js 12.20+ uses package exports with subpath patterns
          - 18 # Node.js 17+ cannot use package exports with trailing slashes

        conditions:
          - require
          - import

        export:
          - govuk-frontend
          - govuk-frontend/package.json
          - govuk-frontend/dist/govuk/components/accordion/accordion.bundle.js
          - govuk-frontend/dist/govuk/components/accordion/accordion.bundle.mjs
          - govuk-frontend/dist/govuk/i18n.mjs

        include:
          - conditions: require
            node-version: 12.18 # Node.js 12.18 uses package exports with trailing slashes
```

Can have a single `run` line and it'll swap in `${{ matrix.export }}` for each combo:

```yaml
        run: node --eval "console.log(require.resolve('${{ matrix.export }}'))"${{ env.FLAGS }}
```

36degrees:
Won't that run each command as a separate job?

colinrotherham (approved):
Sorry, forgot to approve this

We use separate jobs in [`sass.yaml`](https://github.com/alphagov/govuk-frontend/actions/workflows/sass.yaml) but the suggestion above isn't a blocker

Thinking of Windows users though, shall we delete the super slow step?

```yaml
- name: Restore dependencies
```

We don't _really_ need to restore the entire `node_modules` directory per job

Instead, we can suffix the command `npx --workspace govuk-frontend -- ` to run the Node.js CLI from the `./packages/govuk-frontend` directory:

```yaml
run: npx --workspace govuk-frontend -- node --eval "console.log(require.resolve('govuk-frontend'))"${{ env.FLAGS }}
```

36degrees:
I think we've still got parity with the Sass tests in terms of job structure – we have 1 job for each combination of node version and conditions, which is roughly equivalent to each Node / Ruby version and Sass compiler.

If for example we decided to test the compilation of multiple Sass files with the same compiler, I would expect that to be done in separate steps rather than having a separate job for every file we want to test.

I don't quite follow what you're suggesting, but do we need `node_modules` at all, if we're just running a single command with node directly?

Why do we need to run it in the govuk-frontend workspace if what we're testing is resolving govuk-frontend from outside of the package?

colinrotherham:
That's fine, just an idea—not a blocker

> I don't quite follow what you're suggesting, but do we need `node_modules` at all, if we're just running a single command with node directly?

Nah we don't need it at all really, it was just an easy way to add `govuk-frontend` to the require paths at the time

Let's delete the **Restore dependencies** step

> Why do we need to run it in the govuk-frontend workspace if what we're testing is resolving govuk-frontend from outside of the package?

The project root is too high up to locate `govuk-frontend`

Using `require.resolve()` will only walk upwards, but does include some global directories too

You can see all the lookup paths with `require.resolve.paths('govuk-frontend')` with something like:

```console
[
  '/path/to/project/govuk-frontend/packages/govuk-frontend/node_modules',
  '/path/to/project/govuk-frontend/packages/node_modules',
  '/path/to/project/govuk-frontend/node_modules',
  '/path/to/project/node_modules',
  '/path/to/node_modules',
  '/path/node_modules',
  '/node_modules',
  '/Users/example/.node_modules',
  '/Users/example/.node_libraries',
  '/Users/example/.asdf/installs/nodejs/18.14.0/lib/node'
]
```

So we can either:

1. Add GitHub Action step [`working-directory`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsrun) to run from `./packages/govuk-frontend`
2. Use `npx --workspace govuk-frontend` to run from `./packages/govuk-frontend`

Once Node.js is in a directory with **package.json** it can resolve against the exports without `node_modules`

36degrees:
Looks like without `node_modules` npx can't find the workspace?

Or I've done something wrong…!

Shall I drop that last commit and keep the restore dependencies step?

colinrotherham:
Famous last words. Looks like `npx` that comes with Node.js 12 isn't as capable 😭

Shall we try GitHub's step [`working-directory`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsrun) instead or save these optimisations for another day?

It's certainly faster now though in Node.js 18

36degrees:
Force-pushed to remove that last commit, keeping the 'restore dependencies' step.

colinrotherham (approved):

PR #3907, opened 2023-07-05
Remove side effects from package.json

36degrees:
Since we’ve now removed the last of the polyfills in the vendor directory, I believe we can remove this from our `package.json`.


colinrotherham (approved):
Looks good

Might come back as the [`@rollup/plugin-babel`](https://www.npmjs.com/package/@rollup/plugin-babel) "virtual" directory

```json
"sideEffects": [
   "**/_virtual/**"
 ],
 ```

But depends if we use "pure" polyfills that don't modify globals

PR #3910, opened 2023-07-06
Watch for changes to assets

36degrees:
I was taking a look at #3873 and noticed that changes to assets were not reflected in the review app unless I manually ran `npm run build:package`.

Watch the `govuk/assets` folder and re-run the assets task when there are changes.

This means that changes to the assets folder are applied to the dist folder (and therefore are reflected in the review app) without having to manually run the `build:package` task.

### DIFF
--- a/packages/govuk-frontend/tasks/build/package.mjs
+++ b/packages/govuk-frontend/tasks/build/package.mjs
@@ -3,7 +3,7 @@ import { join } from 'path'
 import { files, task } from 'govuk-frontend-tasks'
 import gulp from 'gulp'
 
-import { fixtures, scripts, styles, templates } from '../index.mjs'
+import { assets, fixtures, scripts, styles, templates } from '../index.mjs'
 
 /**
  * Build package task
@@ -16,18 +16,11 @@ export default (options) => gulp.series(
     files.clean('*', options)
   ),
 
+  assets(options),
   fixtures(options),
   scripts(options),
   styles(options),
   templates(options),
-
-  // Copy GOV.UK Frontend static assets
-  task.name('copy:assets', () =>
-    files.copy('**/*', {
-      srcPath: join(options.srcPath, 'govuk/assets'),
-      destPath: join(options.destPath, 'govuk/assets')
-    })
-  ),
 
   // Copy GOV.UK Prototype Kit JavaScript
   task.name("copy:files 'govuk-prototype-kit'", () =>
--- a/packages/govuk-frontend/tasks/watch.mjs
+++ b/packages/govuk-frontend/tasks/watch.mjs
@@ -4,7 +4,7 @@ import { npm, task } from 'govuk-frontend-tasks'
 import gulp from 'gulp'
 import slash from 'slash'
 
-import { fixtures, scripts, styles, templates } from './index.mjs'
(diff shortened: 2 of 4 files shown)


packages/govuk-frontend/tasks/watch.mjs
```diff
@@ -57,5 +57,12 @@ export const watch = (options) => gulp.parallel(
     gulp.watch([
       `${slash(options.srcPath)}/govuk/**/*.{md,njk}`
     ], templates(options))
+  ),
+
+  // Copy GOV.UK Frontend static assets
+  task.name('copy:assets watch', () =>
+    gulp.watch([
+      `${slash(options.srcPath)}/govuk/assets/`
```
colinrotherham, line 65: Mind if we use globs to target files, not directories?

Might be fine for assets, but we avoid a potential race condition where **Nodemon** (server-side) or **Browsersync** (client-side) might reload before the contents has finished copying

```suggestion
      `${slash(options.srcPath)}/govuk/assets/**`
```

colinrotherham:
Thanks @36degrees this is brilliant 🎉

For context, we spotted a list of things we didn't watch in this PR:

* https://github.com/alphagov/govuk-frontend/pull/3498#issuecomment-1535033842

But last week we added [Gulp `.watch()` for component data](https://github.com/alphagov/govuk-frontend/commit/df210908b1f710c6fbb239f74213160343fe6397) so this PR might complete that list?

packages/govuk-frontend/tasks/watch.mjs
colinrotherham: For context, it's largely because the macOS [File System Events (FSEvents) API](https://developer.apple.com/documentation/coreservices/file_system_events) used by the [`chokidar` watcher](https://github.com/paulmillr/chokidar) is replaced with basic "polling" on Windows

So when directories are removed by `clean()` we still need to wait for all the files to come back

packages/govuk-frontend/tasks/watch.mjs
```diff
@@ -57,5 +57,12 @@ export const watch = (options) => gulp.parallel(
...
+    ], assets(options))
```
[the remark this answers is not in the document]
36degrees: I had noticed that when deleting a file from `src` it wasn't deleted from `package`, presumably because we're only running `copy:files` and not `clean`.

Removing the `unlink` wouldn't fix that but would prevent running the task when a file's deleted when it's not going to do anything anyway?

packages/govuk-frontend/tasks/build/package.mjs
```diff
@@ -16,19 +16,12 @@ export default (options) => gulp.series(
     files.clean('*', options)
   ),
 
+  assets(options),
```
colinrotherham, line 19: Should we share this new task with `release.mjs` too?

It's mostly identical between `package.mjs` and `release.mjs`

Would just need `destPath` overriding without the `govuk/` prefix

colinrotherham (approved):
Approved! One minor comment about reusing tasks but not a blocker

packages/govuk-frontend/tasks/build/package.mjs
36degrees: Will leave this as a future improvement if that's alright.

Issue #3827, opened 2023-06-19
Backport inverse breadcrumbs, back links and button modifier classes to v4.7.0


36degrees:
## What

Backport the new modifier classes for inverse breadcrumbs, back links and buttons to the `support/4.x` branch so that we can release them as part of v4.7.0

These changes were made to the `main` branch in:

- https://github.com/alphagov/govuk-frontend/pull/3556
- https://github.com/alphagov/govuk-frontend/pull/3774

## Why

Releasing these changes as part of v4.7.0 means users can benefit from the new features earlier, and allows us to reduce the length of the release notes for v5.0.

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [x] Pull request raised to backport the new inverse modifiers for breadcrumbs, back links and button to the `support/4.x` branch
- [x] Changes have been documented in the changelog on the `support/4.x` branch

PR #3850, opened 2023-06-23
Backport: Add inverse modifiers for Buttons, breadcrumbs and back links

owenatgov:
This is a backport of the following changes onto `support/4.x` for release in 4.7.0:

- https://github.com/alphagov/govuk-frontend/pull/3556
- https://github.com/alphagov/govuk-frontend/pull/3774

Resolves https://github.com/alphagov/govuk-frontend/issues/3827

PR #3915, opened 2023-07-06
re-add inverse breadcrumbs sass

owenatgov:
Somehow during the backporting of inverse modifiers for buttons, breadcrumbs and back links in https://github.com/alphagov/govuk-frontend/issues/3827, the actual sass code for the inverse breadcrumbs was missed.

### DIFF
--- a/src/govuk/components/breadcrumbs/_index.scss
+++ b/src/govuk/components/breadcrumbs/_index.scss
@@ -139,4 +139,16 @@
       }
     }
   }
+
+  .govuk-breadcrumbs--inverse {
+    color: govuk-colour("white");
+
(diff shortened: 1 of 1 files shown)


colinrotherham (approved):

36degrees:
Looks like those changes got made against the `package` folder rather than `src`:
https://github.com/alphagov/govuk-frontend/pull/3850/files#diff-e56ff8e57885f1320a693de4328ff561a2f11055144805e68b205f0398c83518

owenatgov:
That explains why it got removed from the dist in https://github.com/alphagov/govuk-frontend/pull/3912

PR #3920, opened 2023-07-06
[SPIKE] Add bundled-only `"browser"` package exports

colinrotherham:
This spike adds "pre-bundled" **package.json** `"browser"` exports whilst defaulting to separate modules

Whilst not included in this PR, we could resolve UMD for browser-based bundlers alongside CommonJS

(For example, Browserify and webpack v4)

```json
".": {
  "browser": {
    "sass": "./dist/govuk/all.scss",
    "import": "./dist/govuk/all.bundle.mjs",
    "require": "./dist/govuk/all.bundle.js",
    "default": "./dist/govuk/all.bundle.js"
  },
  "sass": "./dist/govuk/all.scss",
  "import": "./dist/govuk/all.mjs",
  "require": "./dist/govuk/all.cjs",
  "default": "./dist/govuk/all.bundle.js"
}
```

Split out from:

* https://github.com/alphagov/govuk-frontend/pull/3726#discussion_r1252818553

Includes a commit from https://github.com/alphagov/govuk-frontend/pull/3906 to improve our export test output

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -232,7 +232,7 @@ jobs:
         run: ${{ matrix.task.run }}
 
   package:
-    name: Export ${{ matrix.conditions }}, Node.js ${{ matrix.node-version }}
+    name: Export ${{ join(matrix.conditions, ' ') || 'require' }}, Node.js ${{ matrix.node-version }}
     runs-on: ubuntu-latest
     needs: [install, build]
 
@@ -245,17 +245,17 @@ jobs:
           - 18 # Node.js 17+ cannot use package exports with trailing slashes
 
         conditions:
-          - require
-          - import
+          - [require]
+          - [import]
 
         include:
-          - conditions: require
+          - conditions: []
             node-version: 12.18 # Node.js 12.18 uses package exports with trailing slashes
 
     env:
       # Node.js conditions override from "require" to "import" etc
       # https://nodejs.org/api/cli.html#-c-condition---conditionscondition
-      FLAGS: ${{ matrix.conditions != 'require' && format(' --conditions {0}', matrix.conditions) || '' }}
+      FLAGS: ${{ matrix.conditions[0] != null && format(' --conditions {0}', join(matrix.conditions, ' --conditions ')) || '' }}
 
     steps:
       - name: Checkout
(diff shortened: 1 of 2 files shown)


romaricpascal:
Digging around a bit, it seems that neither Browserify[^1] nor Webpack 4[^2] would use that field. Is this field meant for other tools/versions? Or is it mean to be `browser` at top level, which both would understand (but double up with `main` which they already understand).

[^1]: Judging by [its documentation](https://browserify.readthedocs.io/en/stable/readme/?q=package.json&check_keywords=yes&area=default#packagejson) Browserify looks at `main` and `browser` at the top level of `package.json` rather than inside `exports`
[^2]: Support for exports in `package.json` seems to only [have landed in Webpack 5](https://github.com/webpack/webpack/issues/9509#issuecomment-1381896299) so wouldn't be used by Webpack 4

colinrotherham:
@romaricpascal Sorry I didn't explain very well 😆

Remember we tried compiling separate CommonJS modules?

This would let unpkg, webpack 4 and Browserify stay on top-level `"browser"` as UMD

But modern bundlers get `require()` or `import` (with tree-shaking) via package exports as CommonJS, but we'd still offer a route for those applying a `"browser"` condition to get bundles if needed

romaricpascal:
None of these tools mention a `browser` field inside `exports` for getting their dependency (webpack and Browserify talk about one at the top level of `package.json`, and Unpkg mentions [its own `unpkg` field](https://unpkg.com/)). So what I was trying to clarify is whether the configuration was in the right place.

Thinking on it some more, I think until we get a clear demand for automatically linking to an UMD build for these old bundlers, we should avoid adding more complexity to our exports. People can `require` the relevant path in their code, so wouldn't be blocked. That's also one less path that `govuk-frontend` may resolve to, which simplifies support. 

If demand comes in, though, happy to look into it further. The change wouldn't be breaking anyway as people using the full path wouldn't be forced to make the update.

colinrotherham:
Yeah, don't worry, hope it's clearer now

Definitely, we split this PR out when CommonJS was removed in https://github.com/alphagov/govuk-frontend/pull/3726#discussion_r1252818553

Won't need looking at again unless we drop UMD as the default for `require()`

Issue #3918, opened 2023-07-06
Text colour for inverse button should use colour variable that indicates design intent


paulrobertlloyd:
## Related component

Version v4.7.0 introduced an inverse button modifier (yay!), but it's hard coded to use `govuk-colour("Blue")`:

https://github.com/alphagov/govuk-frontend/blob/635bb3cb65f83ea4056353e276270595b4443add/packages/govuk-frontend/src/govuk/components/button/_index.scss#L39

However, it’s not clear if this is blue because `govuk-brand-colour` is Blue, or because `govuk-link-colour` is Blue. In most cases, this doesn't matter because backgrounds and links are both blue. But when the brand or link colours differ from the default, what should an inverse button look like? Should it:

* Have a text colour that is the same as other links
* Have a text colour that is the same as the background it sits on (essentially giving it a ‘see-through’ appearance)

## Context

* On the Mod Design System site, the inverse button shares the same colour as the background **and** the link colour, which are both `#532a45`:

  <img width="300" alt="MoD masthead with inverse button" src="https://github.com/alphagov/govuk-frontend/assets/813383/848ef051-1e5b-4f5c-8118-fd10ecae491e">

* On X-GOVUK sites, the inverse button text colour is the same as the turquoise background (brand) colour (`#2288aa`), as links are Blue (`#1d70b8`):

  <img width="300" alt="X-GOVUK masthead with inverse button" src="https://github.com/alphagov/govuk-frontend/assets/813383/529f0b6f-40fd-49ee-a248-223c39e7dce6">

## Alternatives

On the prototype masthead component, am considering whether to override the hardcoded Blue colour to share that of the masthead’s background: https://github.com/x-govuk/govuk-prototype-components/pull/134

## Additional information (if applicable)

Happy with whatever design decision the Design System team come up with, but the colour in the codebase should indicate the design intent behind the colour chosen for inverse buttons; either `govuk-brand-colour` or `govuk-link-colour` (but not `govuk-colour("Blue")`).

querkmachine:
Hmm, my concern is also that the brand colour has no guarantee of contrast when used against white—although the [Notification Banner component](https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/src/govuk/components/notification-banner/_index.scss) appears to have already assumed it will, using `$govuk-brand-colour` on line 8 and hardcoding white text on top of it on line 33.

In my mind, we might want to do something like this:
[...]

PR #3925, opened 2023-07-07
Improvements to inverse components

querkmachine:
Makes some improvements to how we handle colours for 'inverse' components, which are intended to be used on dark background colours. 

Includes the changes discussed in #3918. 

## Changes
- Adds a global variable to define the inverse text colour: `$govuk-inverse-text-colour`.
- Changes the `govuk-link-style-inverse` mixin to use the new variable.
  - This bubbles the change to the Back Link and Breadcrumbs components, which use the mixin.
- Changes the Breadcrumbs component to use the new variable (for arrows and plain text breadcrumb items).
- Changes the default colours used by the inverse Button:
  - The text colour has been changed from `govuk-colour("blue")` to `$govuk-brand-colour`. 
  - The background colour has been changed from `govuk-colour("white")` to `$govuk-inverse-text-colour`.
- Refactors the Button component so that the inverse button colours can be changed manually if desired, e.g. if the inverse text colour and brand colour do not have sufficient contrast with one another.
  - These are exposed as `$govuk-inverse-button-text-colour` and `$govuk-inverse-button-background-colour`.

### Not changed
- There are several other components where we use white text on a coloured background, but I have not changed them to use the new variable. This is because the background colours are hardcoded in many of these instances, so a user could not easily change them to complement a modified `$govuk-inverse-text-colour`. This includes:
  - Header — hardcoded `white` on `black`
  - Panel (confirmation) — hardcoded `white` on `green`
  - Notification Banner (success) — hardcoded `white` on `green`
  - Warning Text — icon is hardcoded `white` on `black`

## Thoughts
- Interestingly, the default Notification Banner is hardcoded to use `white` on `$govuk-brand-colour`, despite there being no guarantee that the brand colour provides sufficient contrast. This seems like it could be an oversight. Does it need looking into?

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/breadcrumbs/_index.scss
+++ b/packages/govuk-frontend/src/govuk/components/breadcrumbs/_index.scss
@@ -119,7 +119,7 @@
   }
 
   .govuk-breadcrumbs--inverse {
-    color: govuk-colour("white");
+    color: $govuk-inverse-text-colour;
 
     .govuk-breadcrumbs__link {
       @include govuk-link-style-inverse;
--- a/packages/govuk-frontend/src/govuk/components/button/_index.scss
+++ b/packages/govuk-frontend/src/govuk/components/button/_index.scss
@@ -16,6 +16,20 @@ $govuk-button-background-colour: govuk-colour("green") !default;
 
 $govuk-button-text-colour: govuk-colour("white") !default;
 
+/// Inverted button component background colour
+///
+/// @type Colour
+/// @access public
+
+$govuk-inverse-button-background-colour: $govuk-inverse-text-colour !default;
+
+/// Inverted button component text colour
+///
+/// @type Colour
+/// @access public
+
+$govuk-inverse-button-text-colour: $govuk-brand-colour !default;
+
 @include govuk-exports("govuk/component/button") {
   $govuk-button-colour: $govuk-button-background-colour;
   $govuk-button-text-colour: $govuk-button-text-colour;
@@ -35,8 +49,8 @@ $govuk-button-text-colour: govuk-colour("white") !default;
   $govuk-warning-button-shadow-colour: govuk-shade($govuk-warning-button-colour, 60%);
 
   // Inverse button variables
-  $govuk-inverse-button-colour: govuk-colour("white");
-  $govuk-inverse-button-text-colour: govuk-colour("blue");
+  $govuk-inverse-button-colour: $govuk-inverse-button-background-colour;
+  $govuk-inverse-button-text-colour: $govuk-inverse-button-text-colour;
   $govuk-inverse-button-hover-colour: govuk-tint($govuk-inverse-button-text-colour, 90%);
   $govuk-inverse-button-shadow-colour: govuk-shade($govuk-inverse-button-text-colour, 30%);
 
--- a/packages/govuk-frontend/src/govuk/components/header/_index.scss
+++ b/packages/govuk-frontend/src/govuk/components/header/_index.scss
@@ -172,7 +172,7 @@
     margin: 0;
     padding: 0;
     border: 0;
-    color: govuk-colour("white");
+    color: $govuk-header-text;
     background: none;
     cursor: pointer;
 
--- a/packages/govuk-frontend/src/govuk/helpers/_links.scss
+++ b/packages/govuk-frontend/src/govuk/helpers/_links.scss
@@ -259,14 +259,14 @@
 @mixin govuk-link-style-inverse {
   &:link,
   &:visited {
-    color: govuk-colour("white");
+    color: $govuk-inverse-text-colour;
   }
 
   // Force a colour change on hover to work around a bug in Safari
   // https://bugs.webkit.org/show_bug.cgi?id=224483
   &:hover,
   &:active {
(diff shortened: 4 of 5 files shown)


paulrobertlloyd:
> Interestingly, the default Notification Banner is hardcoded to use white on `$govuk-brand-colour`, despite there being no guarantee that the brand colour provides sufficient contrast.

I noticed this too. It may also cause issues if the brand colour is the same or similar to another notification banner or error colour (blue, green or red). For example, in the MoD Design System, the information banner also has a reddish colour:

Notification banner component for the MoD Design System

colinrotherham:
Would you mind giving this a review please @paulrobertlloyd?

Especially with your colour variations suggestion in mind https://github.com/alphagov/govuk-frontend/issues/3918#issuecomment-1625185884

@querkmachine Shall we get designer feedback on the hard-coded things?

paulrobertlloyd:
Thanks for the prompt @colinrotherham. After plentiful yak shaving, I have updated my [prototype masthead component](https://github.com/x-govuk/govuk-prototype-components/pull/134), and added an option to [set a value for `$govuk-inverse-text-colour` in the GOV.UK Eleventy Plugin](https://github.com/x-govuk/govuk-eleventy-plugin/pull/178).

Playing with these 2 values, I foresee an issue, namely this:

Screenshot of the header and masthead components. The dark grey inverse text is visible when set above the light grey brand colour, but not when set on the black header background colour.

What does ‘inverse’ mean? Does it mean:

> text that appears on a known dark background colour

or does it mean:

> text that appears on the brand colour

I believe the intention is the former, but in the examples provided so far, the later.

Ah themeing, it's a big ’ol bag of worms! I foresee 2 options:

1. `$govuk-inverse-text-colour` is intended to be used for text whose background colour is known to be dark given the value is defined within `govuk-frontend`.

    It’s up to consuming libraries to decide how and when to use this value (I think the masthead component should probably not use it directly, but maybe have it as a default value instead)
    
    TL;DR: Shit it!

2. Add an explicit `govuk-brand-text-colour` (or similarly named) variable in addition or instead of the above, to be used when text is intended to appear upon the unknowable brand colour. Right now, I’m not sure any components in `govuk-frontend` have that concept.

    TL;DR: Reconsider approach, deciding whether to add another global variable or change the behaviour/intention of `$govuk-inverse-text-colour`.

querkmachine:
"Text on a known background colour" is probably closest to how I've been imagining it, though not necessarily a known _dark_ background colour. If the brand colour is quite light, then the inverse colour will probably be darker.

But then that makes me think that you're right in that it's more practically an inverse of the brand colour, other than a universal constant. That might be the safer assumption to make, here.

querkmachine:
My thinking currently is to:
1. Maintain `$govuk-inverse-text-colour` as being for contrast on dark backgrounds. (Default: white).
2. Create a new global variable called `$govuk-brand-text-colour`, which is explicitly for contrast with the brand colour. (Default: white).
3. Update the inverse button's default styles to use `$govuk-brand-text-colour`.
4. Do not change the inverse link helper, breadcrumbs or back link (continue using `$govuk-inverse-text-colour`).

This might feel a little redundant, as `$govuk-brand-text-colour` will only be used in one place, but the semantic distinction feels kind of important. Even if we don't use it much internally, it'll probably be useful to users of Frontend who are customising colours.

The helper, breadcrumbs and back link are a bit harder to pin down. Although they are _likely_ to be used on the brand colour, there isn't a guarantee that they will. We use the inverse link style on red in our review app, for example.

If I may be so bold as to nullify my own work, there may be more value in only making the changes to allow the button colour customisation and to save the rest until we have a more comprehensive idea on how we want to implement theming.

paulrobertlloyd:
A few different use cases and concepts are getting combined here:

* Text that appears on dark background
* Text that appears on the brand colour

…with an added complication that the inverse button is itself inversely coloured 🤪

For the former, perhaps it makes sense to continue using private variables or hard-coded values (i.e. the header text should be hard coded to use white because the header background has been hard-coded to use black). Does `$govuk-inverse-text-colour` need to be a global variable?

I think the later is the use case these inverse components best address, for which there are plentiful examples in the wild, and where this formalisation would help. Having a `$govuk-brand-text-colour` global variable to compliment `$govuk-brand-colour` makes a lot of sense, but less clear is how it is applied wiring different components. You’re right, there may be cases where ‘inverse’ !== ‘on the brand colour’.

This is making my head hurt a bit. There’s quite a few hypotheticals being thrown into the mix, too - helpful to test robustness of approach, but perhaps a case of great being the enemy of good?

Maybe this is something to get some other (fresher) eyes on? 👀

querkmachine:
Having given time for others on the team to contribute if they so wished, and having mulled this over a little, I think this PR is going to be closed in favour of #4043 for the time being. 

Over the last few weeks we've had a lot of questions around theming, implementing dark modes, and deeper customisation of Frontend. In my mind it makes sense to tackle all of those together, but the team is already in the midst of radical architecture changes and refactoring for our upcoming v5 release, and I don't think we want to add this to the bucket at this time, even if this PR represents only a small part of a greater vision. 

This isn't a 'no', just a 'not yet'. Hopefully #4043 resolves the most pressing issue with the inverse button component!

PR #3949, opened 2023-07-11
Simplify font family settings

36degrees:
Before we removed compatibility mode, the font-families file included multiple ‘settings’ (realistically, more like constants) that defined the font to be used depending on whether compatibility mode was enabled or not.

Now that we only have one possible default font stack, we can remove the unnecessary abstraction.

We also need to update the default value for the `$govuk-include-default-font-face` setting. Previously it checked whether the chosen font family was the default stack based on GDS Transport, but we can instead just check whether “GDS Transport” is in the font list. This also makes it more robust if somebody for example wants to update the setting to change the fallback fonts.


packages/govuk-frontend/src/govuk/settings/typography.test.js
```diff
@@ -0,0 +1,44 @@
...
+    `
+
+    await expect(compileSassString(sass)).resolves.toMatchObject({
+      css: expect.stringContaining('--result: true')
+    })
+  })
+
+  it('is true if $govuk-font-family includes GDS Transport', async () => {
+    const sass = `
+    $govuk-font-family: "GDS Transport", "Comic Sans MS", "Comic Sans", cursive;
```
colinrotherham, line 19:
```suggestion
      $govuk-font-family: "GDS Transport", "Comic Sans MS", "Comic Sans", cursive;
```

packages/govuk-frontend/src/govuk/settings/typography.test.js
```diff
@@ -0,0 +1,44 @@
...
+    `
+
+    await expect(compileSassString(sass)).resolves.toMatchObject({
+      css: expect.stringContaining('--result: true')
+    })
+  })
+
+  it('is false if $govuk-font-family does not include GDS Transport', async () => {
+    const sass = `
+    $govuk-font-family: "Comic Sans MS", "Comic Sans", cursive;
```
colinrotherham, line 33:
```suggestion
      $govuk-font-family: "Comic Sans MS", "Comic Sans", cursive;
```

packages/govuk-frontend/src/govuk/settings/_typography-font.scss
```diff
@@ -27,15 +25,13 @@ $govuk-font-family-print: sans-serif !default;
...
+/// Defaults to true if "GDS Transport" appears in the $govuk-font-family
+/// setting.
 ///
 /// @type Boolean
 /// @access public
 
-$govuk-include-default-font-face: (
-  $govuk-font-family == $govuk-font-family-gds-transport
-) !default;
+$govuk-include-default-font-face: if(index($govuk-font-family, "GDS Transport"), true, false) !default;
```
colinrotherham, line 34: Can we check for the double-quoted `"GDS Transport"` rather than just `GDS Transport`?

You never know, don't want to match future fonts:

```
GDS Transport
GDS Transport New
```

```suggestion
$govuk-include-default-font-face: if(index($govuk-font-family, '"GDS Transport"'), true, false) !default;
```

36degrees: This is using `list.index` rather than string matching ([`str-index`](https://sass-lang.com/documentation/modules/string/#index)), so should only match complete entries.

Double checked by changing to `if(index($govuk-font-family, '"GDS"')` which caused the tests to fail (always `false`).

colinrotherham: Nice. Good to know 🚀

colinrotherham (approved):
Happy to approve

Should we have deprecated the class first?

Probably want to add a CHANGELOG entry to add:

1. Removed public variable `$govuk-font-family-gds-transport`
2. Action to use variable `$govuk-font-family` instead

36degrees:
> Should we have deprecated the class first?

Hmm, yeah, I think ideally we would have – it looks like [the other font family 'settings' were deprecated](https://github.com/alphagov/govuk-frontend/commit/88f28d949daedcf11d2cac2f2ff323d7c23a51c9) at the point the new font was introduced back in 2019!

Not sure it's worth doing an extra 4.x release just for this, but if we do another before v5 maybe we should include this deprecation then?

Alternatively we hold off on this change until v6?

> Probably want to add a CHANGELOG entry to add:
> 
> 1. Removed public variable `$govuk-font-family-gds-transport`
> 2. Action to use variable `$govuk-font-family` instead

Yep, we also haven't mentioned the removal of the settings for the old version of the font which have already been removed. 

A quick search of Google suggests that people are most often using `$govuk-font-family-gds-transport` to set the font on something, where really they should be using the `govuk-font` mixin instead. I'll try and write a changelog entry to nudge people towards that.

colinrotherham:
Thanks @36degrees

I keep forgetting about GitHub code search too:
https://github.com/search?q=%22%24govuk-font-family-gds-transport%22&type=code

Perhaps see if issues like https://github.com/alphagov/govuk-frontend/issues/3918 will become another v4 release? Good to deprecate it before v5

36degrees:
Rebased and added a changelog entry

colinrotherham (approved):
🎉

paulrobertlloyd:
If there is another 4.x release to deprecate things, please don’t forget https://github.com/alphagov/govuk-frontend/pull/3877

PR #3951, opened 2023-07-11
Additional tests for task list Nunjucks macro, allow attributes on tags and tidy up

### LINKED ISSUE #3828: Improve template test coverage for the task list component
## What

Add additional template tests that cover the various options that can be passed to the macro.

## Why

Improving the coverage will improve our confidence that the component can be used without error – for example, that adding new classes or attributes works as expected.

As a side effect, the new examples that get added as part of writing the tests improve the coverage of the macro in our test fixtures which other teams can use to ensure their ports of GOV.UK Frontend output the same HTML.

As discussed with the original contributor on Slack, they're happy for us to take on the work of adding additional tests on their behalf.

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [ ] Tests have been added that cover all of the flows of execution through the Nunjucks template

36degrees:
Add additional template tests that cover the various options that can be passed to the macro.

Closes #3828.

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/task-list/task-list.yaml
+++ b/packages/govuk-frontend/src/govuk/components/task-list/task-list.yaml
@@ -41,21 +41,9 @@ params:
         params:
           - name: tag
             type: object
+            isComponent: true
             required: false
-            description: Object containing the options for a tag that acts as the status for the task.
-            params:
-              - name: text
-                type: string
-                required: true
-                description: Text to use within the tag. If `html` is provided, the `text` argument will be ignored.
-              - name: html
-                type: string
-                required: true
-                description: HTML to use within the tag. If `html` is provided, the `text` argument will be ignored.
-              - name: classes
-                type: string
-                required: false
-                description: Classes to add to the tag.
+            description: Options for the tag component.
           - name: text
             required: false
             type: string
--- a/packages/govuk-frontend/src/govuk/components/task-list/task-list.yaml
+++ b/packages/govuk-frontend/src/govuk/components/task-list/task-list.yaml
@@ -186,8 +186,7 @@ examples:
             tag:
               text: Error
               classes: govuk-tag--red
-        - classes: app-task-list__item--no-link
-          title:
+        - title:
             text: Payment
           hint:
             text: It will cost between £15 and £75
--- a/packages/govuk-frontend/src/govuk/components/task-list/task-list.yaml
+++ b/packages/govuk-frontend/src/govuk/components/task-list/task-list.yaml
@@ -92,7 +92,6 @@ params:
 examples:
   - name: default
     data:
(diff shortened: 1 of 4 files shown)


packages/govuk-frontend/src/govuk/components/task-list/task-list.yaml
```diff
@@ -41,21 +41,9 @@ params:
...
-                description: Text to use within the tag. If `html` is provided, the `text` argument will be ignored.
-              - name: html
-                type: string
-                required: true
-                description: HTML to use within the tag. If `html` is provided, the `text` argument will be ignored.
-              - name: classes
-                type: string
-                required: false
-                description: Classes to add to the tag.
+            description: Options for the tag component, if the status should use a tag.
```
colinrotherham, line 46: Nice spot. Do you mind adding this to the macro options style guide Trello?

I've only checked briefly, but other `required: false` components don't qualify if they should be used

```suggestion
            description: Options for the tag component.
```

packages/govuk-frontend/src/govuk/components/task-list/template.njk
```diff
@@ -1,31 +1,30 @@
...
       </div>
-      <div class="govuk-task-list__status {{ item.status.classes }}" id="{{ statusId }}">
+      <div class="govuk-task-list__status {%- if item.status.classes %} {{ item.status.classes }}{% endif %}" id="{{ statusId }}">
         {% if item.status.tag %}
           {{ govukTag({
             text: item.status.tag.text,
             html: item.status.tag.html,
-            classes: item.status.tag.classes
+            classes: item.status.tag.classes,
+            attributes: item.status.tag.attributes
```
colinrotherham, line 27: Other nested components usually override `id` or `for` params

But we don't here. Should we just pass all the options in one?

```mjs
{{ govukTag(item.status.tag) }}
```

colinrotherham (commented):
Looks brilliant. Every single commit

Mind if we update the PR title + description to say we've added fixes and features rather than just better test coverage?

packages/govuk-frontend/src/govuk/components/task-list/template.njk
36degrees: That is a great shout 👍🏻

packages/govuk-frontend/src/govuk/components/task-list/task-list.yaml
36degrees: > Nice spot. Do you mind adding this to the macro options style guide Trello?

https://trello.com/c/smPF3ePN/18-nested-components-components-used-by-other-components-should-not-have-their-params-listed-in-the-top-level-component-instead-sho

colinrotherham (approved):

PR #3911, opened 2023-07-06
[main] Pin GitHub `actions/setup-node` to v3.6.0

36degrees:
Cherry-picking @colinrotherham's fix from #3909 to make the same changes on the `main` branch.

---
Workaround for an npm cache path issue throwing the error:

```
Post job cleanup.
Error: Cache folder paths are not retrieved for npm with cache-dependency-path =
```

PR #3954, opened 2023-07-13
Use exact GitHub Actions versions

colinrotherham:
This change adds exact version numbers to all GitHub Actions

Dependabot will now raise PRs for minor/patch updates

## Node.js setup

I've also bumped GitHub `actions/setup-node` back to v3.7.0 (see [rollback PR for more info](https://github.com/alphagov/govuk-frontend/pull/3911))

Version v3.7.0 now runs cache checks earlier, [causing issues when `~/.npm` (npm cache) didn't exist](https://github.com/actions/setup-node/issues/797). The update revealed that we hadn't anticipated `null` "Boolean values" (booleans as strings) as shown in:

* https://github.com/actions/runner/issues/2238

Fixed in https://github.com/alphagov/govuk-frontend/commit/7874ca487592e2ed2545df6f452e8e8d622f7cd2

### DIFF
--- a/.github/workflows/actions/build/action.yml
+++ b/.github/workflows/actions/build/action.yml
@@ -5,7 +5,7 @@ runs:
 
   steps:
     - name: Cache build
-      uses: actions/cache@v3
+      uses: actions/cache@v3.3.1
       id: build-cache
 
       with:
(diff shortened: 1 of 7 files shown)


romaricpascal (approved):
Nice! Sounds much safer to have a Dependabot PR to test newer action versions for any version bump rather than having them used by Github automatically 🥳

Issue #3682, opened 2023-05-24
Pagination for smaller numbers of pages does not show current page number


CharlotteDowns:
This issue is from May 2023 external accessibility audit report.

**Usability issue**

**Issue ID:**
DAC_Usability_Pagination_01

**URLs:**
[https://govuk-frontend-pr-3503.herokuapp.com/full-page-examples/search-three-pages](https://govuk-frontend-pr-3503.herokuapp.com/full-page-examples/search-three-pages)

## Screen shots
<img width="182" alt="Pagination with text 'Previous 1 of 3' then 'Next 3 of 3'" src="https://github.com/alphagov/govuk-frontend/assets/56260216/bf4076c2-b33a-44c2-b67c-e357d8c4feb5">

The pagination method used does not indicate to users which page they are currently on and relies on the user to decipher this information from the context of the previous and next links, which some users may not be able to do with ease.

## Current code ref(s)
`#main-content > form > div.govuk-grid-row > div.govuk-grid-column-two-thirds > nav`

## Low vision user comments
> “While viewing this page I found it difficult to determine which page I am currently viewing because there is no number being displayed at the bottom of the page therefore, I cannot determine which page I am on. I found the ‘Navigational page’ buttons that is displayed on journey 5 to be clearer for me to determine which page I am currently viewing; this would also help me to navigate between each page with ease.”

## Cognitive user comments
> "It was difficult for me to understand exactly what page I’m on with this style of pagination, as it doesn’t explicitly state what page the user is on. Others encountering this may find it just as confusing to use and understand. The pagination in Journey 5 is a much better design choice for finding pages you’re on and what ones you can navigate to.”

## Solution
Consider using the pagination method which indicates the current page for users and not just the next and previous pages.

querkmachine:
I think the ismplest way of doing this would be a visually-hidden element with info on the current page in it and `aria-current="page"`. Existing implementations would need updating to add this previously unnecessary information, however.

davidc-gds:
@querkmachine do you know if this would be a 'breaking change' for existing implementations of the 'previous / next' pagination variant?

querkmachine:
@davidc-gds Assuming we would only be adding in new 'stuff', rather than changing or removing what's already there, then this wouldn't be a breaking change, no.

calvin-lau-sig7:
Just weighing in that @36degrees's suggestions sound sensible to slightly adjusting the use case for prev/next pagination.

[...]

PR #3957, opened 2023-07-13
Update pagination examples in review app

querkmachine:
Updates some of the full page examples in the review app to show both versions of the pagination component _in situ_ with their expected use cases. 

This is intended to complement a proposed guidance change (https://github.com/alphagov/govuk-design-system/issues/2944), based on a suggestion by Ollie made here: https://github.com/alphagov/govuk-frontend/issues/3682#issuecomment-1621643850

## Changes
- Changes the labels on the 'pagination with labels' example as we no longer recommend using these labels to number pages.
- Changes the 'search' page example to use numbered pagination.
- Adds a new 'travel guidance' page example to demonstrate block pagination.

### DIFF
--- a/packages/govuk-frontend-review/src/views/full-page-examples/search/index.njk
+++ b/packages/govuk-frontend-review/src/views/full-page-examples/search/index.njk
@@ -184,10 +184,15 @@ notes: >-
 
       {{ govukPagination({
         next: {
-          text: 'Next page',
-          href: '#',
-          labelText: '2 of 2939'
-        }
+          href: "#"
+        },
+        items: [
+          { href: "#", number: 1, current: true },
+          { href: "#", number: 2 },
+          { href: "#", number: 3 },
+          { ellipsis: true },
+          { href: "#", number: 2939 }
+        ]
       }) }}
     </div>
   </div>
--- a/packages/govuk-frontend-review/src/views/full-page-examples/travel-guidance/index.njk
+++ b/packages/govuk-frontend-review/src/views/full-page-examples/travel-guidance/index.njk
@@ -1,0 +1,235 @@
+---
+scenario: You want to read guidance about the entry requirements for travelling to Narnia.
+notes: Based on https://www.gov.uk/foreign-travel-advice/iceland/entry-requirements
+---
+{% extends "layouts/full-page-example.njk" %}
+
+{% from "govuk/components/breadcrumbs/macro.njk" import govukBreadcrumbs %}
+{% from "govuk/components/pagination/macro.njk" import govukPagination %}
+
+{% set pageTitle = "Entry requirements - Narnia travel advice" %}
+{% block pageTitle %}{{ pageTitle }} - GOV.UK{% endblock %}
+
+{% block beforeContent %}
+  {{ govukBreadcrumbs({
+    items: [
+      {
+        text: "Home",
+        href: "#/"
+      },
+      {
+        text: "Passports, travel and living abroad",
+        href: "#/browse/abroad"
+      },
+      {
+        text: "Travel abroad",
+        href: "#/browse/abroad/travel-abroad"
+      },
+      {
+        text: "Foreign travel advice",
+        href: "#/foreign-travel-advice"
+      }
+    ]
+  }) }}
+{% endblock %}
+{% block content %}
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds-from-desktop">
+
+    <div class="govuk-!-margin-bottom-8">
+      <span class="govuk-caption-xl">Foreign travel advice</span>
+      <h1 class="govuk-heading-xl">Narnia entry requirements</h1>
+    </div>
+
+    <aside class="govuk-!-padding-bottom-3" role="complementary">
+      <nav aria-label="Travel advice pages" role="navigation">
+        <h2 class="govuk-heading-s">Contents</h2>
+        <ol class="govuk-list govuk-list--bullet govuk-!-font-size-16">
+          <li><a class="govuk-link" href="#/foreign-travel-advice/narnia">Summary</a></li>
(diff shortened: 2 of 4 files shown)


packages/govuk-frontend-review/src/views/full-page-examples/travel-guidance/index.njk
```diff
@@ -0,0 +1,396 @@
...
+  <div class="govuk-grid-column-two-thirds-from-desktop">
+
+    <div class="govuk-!-margin-bottom-8">
+      <span class="govuk-caption-xl">Foreign travel advice</span>
+      <h1 class="govuk-heading-xl">Narnia entry requirements</h1>
+    </div>
+    
+    <aside class="govuk-!-padding-bottom-3" role="complementary">
+      <nav aria-label="Travel advice pages" class="gem-c-contents-list" role="navigation">
+        <h2 class="gem-c-contents-list__title">Contents</h2>
```
colinrotherham, line 46: Is the `gem-c-contents-list` component similar to `gem-c-related-navigation` in that we should add a disclaimer comment above it?

https://github.com/alphagov/govuk-frontend/blob/087d6807c96c14908b5933d3fa6936060d9197d7/packages/govuk-frontend-review/src/views/full-page-examples/bank-holidays/index.njk#L3180

packages/govuk-frontend-review/src/views/full-page-examples/travel-guidance/index.njk
```diff
@@ -0,0 +1,396 @@
...
+        <ol class="gem-c-contents-list__list">
+          <li class="gem-c-contents-list__list-item gem-c-contents-list__list-item--dashed">
+            <a class="gem-c-contents-list__link govuk-link" href="#/">Warnings and insurance</a>
```
colinrotherham, line 49: I hadn't realised we use this fake link pattern

But the other examples all populate a fake path too

```suggestion
            <a class="gem-c-contents-list__link govuk-link" href="#/warnings-and-insurance">Warnings and insurance</a>
```

packages/govuk-frontend-review/src/views/full-page-examples/travel-guidance/index.njk
```diff
@@ -0,0 +1,396 @@
...
+    ],
+    html: 'Built by the <a href="#" class="govuk-footer__link">Government Digital Service</a>'
+  }
+}) }}
+
+{% endblock %}
+
+{% block styles %}
+<style>
+/* Styles for the GOV.UK Publishing Components contents list. */
```
colinrotherham, line 251: Mind if we move these styles into [`./stylesheets/full-page-examples`](https://github.com/alphagov/govuk-frontend/tree/main/packages/govuk-frontend-review/src/stylesheets/full-page-examples) as `travel-guidance.scss`?

packages/govuk-frontend-review/src/views/full-page-examples/search/index.njk
```diff
@@ -184,10 +184,15 @@ notes: >-
 
       {{ govukPagination({
         next: {
-          text: 'Next page',
-          href: '#',
-          labelText: '2 of 2939'
-        }
+          href: '#'
```
colinrotherham, line 187: We're slowly switching our Nunjucks to double quotes everywhere

Can we fix these at the same time? Thanks @querkmachine

packages/govuk-frontend-review/src/views/full-page-examples/travel-guidance/index.njk
querkmachine: I can add a comment noting it's not part of GOV.UK Frontend, but the Contents List styles aren't bundled with the Prototype Kit, so I'll have to trim off that bit.

packages/govuk-frontend-review/src/views/full-page-examples/travel-guidance/index.njk
colinrotherham: Trying to pick a consistent approach we're all happy with is more important

Typically it'd be to tap into our setup so it gets free upgrades as we go:

1. Stylelint for code quality checks
2. Sass tools, helpers (colours, font stacks, type scale, breakpoints)
3. PostCSS transforms (Autoprefixer, CSSNANO)

But if the lift-and-shift examples should have a separate `<style>` tag I'm happy with that

I don't know how far we typically go to adhere them to GOV.UK Frontend, avoiding tech debt

colinrotherham: Just bringing over a conversation from Slack

Accepting that this is tech debt that we'll need to support during the new type scale work

But in reply from @36degrees:

>This has been at the back of my mind for a while. I think we want the full-page examples as ‘realistic’ as possible but ideally use as little extra CSS as possible. I don’t think we need to precisely recreate the layout as it is on the pages we’re mimicking, just get close enough that e.g. DAC can understand what it is they’re looking at.
>
>I think on that basis my preference for the new travel advice page would be to just use a standard H2 and a bulleted list. I would quite like to at some point simplify the search example. The campaign page may be trickier, but also I’m not sure how useful it is any more.
>
>I don’t think we need to eradicate extra CSS entirely, but I think we should try and keep it to a minimum. Inevitably they diverge from the original design over time anyway.
Loosely held opinion, so definitely open to hearing alternative views.

Do we need to tweak anything @querkmachine?

querkmachine: It sounds like the intent going forward is to try and be 'similar but not the same' with regard to non-Frontend components in examples, so I can strip this down to a basic heading and list, sure.

colinrotherham (approved):
Looks more maintainable now, thanks @querkmachine

I'll approve now but feel free to make any little markup tweaks if you're still testing

PR #3820, opened 2023-06-19
[SPIKE] Babel with ES2020 class fields

colinrotherham:
This PR uses Babel to transform two ES2020 features back to ES2015 without polyfills or helpers

## Public static fields
After merging [Migrate JavaScript to ES2015 classes](https://github.com/alphagov/govuk-frontend/pull/3771) we were left with "optional" class properties appended in the constructor that would be clearly defined as [`static`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/static) in ES2020

### Before

```mjs
export class I18n {
  constructor() {
    // ...
  }
}

I18n.pluralRulesMap = {}
I18n.pluralRules = {}
```

### After

```mjs
export class I18n {
  constructor() {
    // ...
  }
  
  static pluralRulesMap = {}
  static pluralRules = {}
}
```

## Public fields
We also have "optional" class properties appended in the constructor that would be [public class fields](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Public_class_fields)

### Before

```mjs
export class Radios {
  constructor($module) {
    if (!($module instanceof HTMLElement)) {
      return this
    }
  
    this.$module = $module
    this.$inputs = $module.querySelectorAll('input[type="radio"]')
  }
}
```

### After

Both `$module` and `$inputs` can be declared as class fields (with default values if required)

```mjs
export class Radios {
  $module
  $inputs

  constructor($module) {
    if (!($module instanceof HTMLElement)) {
      return this
    }
  
    this.$module = $module
    this.$inputs = $module.querySelectorAll('input[type="radio"]')
  }
}
```

PR #3958, opened 2023-07-13
Configure Babel to remove non-public comments

colinrotherham:
This PR takes a look at component KB file size without comments

It configures Babel to remove all non-public comments unless tagged with `@preserve` or containing JSDoc tags

# Before

## `v5.0.0-preview`

File sizes from the [`main` branch](https://govuk-frontend-review.herokuapp.com)

| Component module    | File size |
| :---                |      ---: |
| all.mjs             | 114.7 KB  |
| accordion.mjs       | 40.31 KB  |
| character-count.mjs | 39.06 KB  |

## `v4.7.0`

File sizes from the **review app** on the `v4.7.0` tag

| Component module    | File size |
| :---                |      ---: |
| all.mjs             | 161.40 KB  |
| accordion.mjs       | 82.67 KB  |
| character-count.mjs | 76.34 KB  |

# After

## Only non-public comments removed

File sizes from [`rollup-babel-comments` branch](https://govuk-frontend-pr-3958.herokuapp.com/) with all non-public comments removed versus `v5.0.0-preview`

| Component module    | File size | Change  |
| :---                | :---      | :---    |
| all.mjs             | 66.35 KB  | -42.15% |
| accordion.mjs       | 20.77 KB  | -48.47% |
| character-count.mjs | 20.34 KB  | -47.93% |

## Only private and internal comments removed

File sizes from a previous spike with private and internal comments removed versus `v5.0.0-preview`

| Component module    | File size | Change  |
| :---                | :---      | :---    |
| all.mjs             | 94.19 KB  | -17.88% |
| accordion.mjs       | 36 KB  | -10.69% |
| character-count.mjs | 34.13 KB  | -12.62% |

## All comments removed

File sizes from a previous spike with ALL comments removed versus `v5.0.0-preview`

| Component module    | File size | Change  |
| :---                | :---      | :---    |
| all.mjs             | 55.57 KB  | -51.55% |
| accordion.mjs       | 19.98 KB  | -50.43% |
| character-count.mjs | 15.24 KB  | -60.98% |

### DIFF
--- a/packages/govuk-frontend/babel.config.js
+++ b/packages/govuk-frontend/babel.config.js
@@ -19,9 +19,17 @@ module.exports = function (api) {
 
         // Assume all comments are public unless
         // tagged with `@private` or `@internal`
-        return ['* @internal', '* @private'].every(
-          (tag) => !comment.includes(tag)
+        const isPrivate = ['* @internal', '* @private'].some((tag) =>
+          comment.includes(tag)
         )
+
+        // Flag any JSDoc comments worth keeping
+        const isDocumentation = ['* @param', '* @returns', '* @typedef'].some(
+          (tag) => comment.includes(tag)
+        )
+
+        // Print only public JSDoc comments
+        return !isPrivate && isDocumentation
       }
     },
     presets: [
--- a/packages/govuk-frontend/babel.config.js
+++ b/packages/govuk-frontend/babel.config.js
@@ -11,6 +11,19 @@ module.exports = function (api) {
   const browserslistEnv = isBrowser ? 'javascripts' : 'node'
 
   return {
+    generatorOpts: {
+      shouldPrintComment(comment) {
+        if (!isBrowser || comment.includes('* @preserve')) {
+          return true
+        }
+
+        // Assume all comments are public unless
+        // tagged with `@private` or `@internal`
+        return ['* @internal', '* @private'].every(
+          (tag) => !comment.includes(tag)
+        )
+      }
+    },
     presets: [
       [
         '@babel/preset-env',
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -43,11 +43,6 @@ module.exports = {
           }
         ],
 
-        // JSDoc blocks are optional
-        'jsdoc/require-jsdoc': 'off',
-        'jsdoc/require-param-description': 'off',
-        'jsdoc/require-param': 'off',
-
         // Check for valid formatting
         'jsdoc/check-line-alignment': [
           'warn',
@@ -56,6 +51,19 @@ module.exports = {
             wrapIndent: '  '
           }
         ],
+
+        // JSDoc blocks can use `@preserve` to prevent removal
+        'jsdoc/check-tag-names': [
+          'warn',
+          {
+            definedTags: ['preserve']
+          }
+        ],
+
+        // JSDoc blocks are optional
+        'jsdoc/require-jsdoc': 'off',
+        'jsdoc/require-param-description': 'off',
+        'jsdoc/require-param': 'off',
 
         // Require hyphens before param description
         // Aligns with TSDoc style: https://tsdoc.org/pages/tags/param/
--- a/docs/contributing/coding-standards/js.md
+++ b/docs/contributing/coding-standards/js.md
@@ -15,6 +15,8 @@ component
 ```mjs
 /**
  * Component name
+ *
+ * @preserve
  */
 export class Example {
   /**
--- a/packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
+++ b/packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
@@ -13,6 +13,8 @@ import { I18n } from '../../i18n.mjs'
  *
  * The state of each section is saved to the DOM via the `aria-expanded`
  * attribute, which also provides accessibility.
+ *
+ * @preserve
  */
 export class Accordion {
   /** @private */
--- a/packages/govuk-frontend/src/govuk/components/button/button.mjs
+++ b/packages/govuk-frontend/src/govuk/components/button/button.mjs
@@ -6,6 +6,8 @@ const DEBOUNCE_TIMEOUT_IN_SECONDS = 1
 
 /**
  * JavaScript enhancements for the Button component
+ *
+ * @preserve
  */
 export class Button {
   /** @private */
@@ -24,7 +26,6 @@ export class Button {
   debounceFormSubmitTimer = null
 
   /**
-   *
    * @param {Element} $module - HTML element to use for button
    * @param {ButtonConfig} [config] - Button config
    */
--- a/packages/govuk-frontend/src/govuk/components/character-count/character-count.mjs
+++ b/packages/govuk-frontend/src/govuk/components/character-count/character-count.mjs
@@ -12,6 +12,8 @@ import { I18n } from '../../i18n.mjs'
  *
  * You can configure the message to only appear after a certain percentage
  * of the available characters/words has been entered.
+ *
+ * @preserve
  */
 export class CharacterCount {
   /** @private */
--- a/packages/govuk-frontend/src/govuk/components/checkboxes/checkboxes.mjs
+++ b/packages/govuk-frontend/src/govuk/components/checkboxes/checkboxes.mjs
@@ -1,5 +1,7 @@
 /**
  * Checkboxes component
+ *
+ * @preserve
  */
 export class Checkboxes {
   /** @private */
--- a/packages/govuk-frontend/src/govuk/components/error-summary/error-summary.mjs
+++ b/packages/govuk-frontend/src/govuk/components/error-summary/error-summary.mjs
@@ -5,6 +5,8 @@ import { normaliseDataset } from '../../common/normalise-dataset.mjs'
  * Error summary component
  *
  * Takes focus on initialisation for accessible announcement, unless disabled in configuration.
+ *
+ * @preserve
  */
 export class ErrorSummary {
   /** @private */
@@ -17,7 +19,6 @@ export class ErrorSummary {
   config
 
   /**
-   *
    * @param {Element} $module - HTML element to use for error summary
    * @param {ErrorSummaryConfig} [config] - Error summary config
    */
--- a/packages/govuk-frontend/src/govuk/components/exit-this-page/exit-this-page.mjs
+++ b/packages/govuk-frontend/src/govuk/components/exit-this-page/exit-this-page.mjs
@@ -4,6 +4,8 @@ import { I18n } from '../../i18n.mjs'
 
 /**
  * Exit This Page component
+ *
+ * @preserve
  */
 export class ExitThisPage {
   /** @private */
--- a/packages/govuk-frontend/src/govuk/components/header/header.mjs
+++ b/packages/govuk-frontend/src/govuk/components/header/header.mjs
@@ -1,5 +1,7 @@
 /**
  * Header component
+ *
+ * @preserve
  */
 export class Header {
   /** @private */
--- a/packages/govuk-frontend/src/govuk/components/notification-banner/notification-banner.mjs
+++ b/packages/govuk-frontend/src/govuk/components/notification-banner/notification-banner.mjs
@@ -3,6 +3,8 @@ import { normaliseDataset } from '../../common/normalise-dataset.mjs'
 
 /**
  * Notification Banner component
+ *
+ * @preserve
  */
 export class NotificationBanner {
   /** @private */
--- a/packages/govuk-frontend/src/govuk/components/radios/radios.mjs
+++ b/packages/govuk-frontend/src/govuk/components/radios/radios.mjs
@@ -1,5 +1,7 @@
 /**
  * Radios component
+ *
+ * @preserve
  */
 export class Radios {
   /** @private */
--- a/packages/govuk-frontend/src/govuk/components/skip-link/skip-link.mjs
+++ b/packages/govuk-frontend/src/govuk/components/skip-link/skip-link.mjs
@@ -1,5 +1,7 @@
 /**
  * Skip link component
+ *
+ * @preserve
  */
 export class SkipLink {
   /** @private */
@@ -15,7 +17,6 @@ export class SkipLink {
   linkedElementListener = false
 
   /**
-   *
    * @param {Element} $module - HTML element to use for skip link
    */
   constructor($module) {
--- a/packages/govuk-frontend/src/govuk/components/tabs/tabs.mjs
+++ b/packages/govuk-frontend/src/govuk/components/tabs/tabs.mjs
@@ -1,5 +1,7 @@
 /**
  * Tabs component
+ *
+ * @preserve
  */
 export class Tabs {
   /** @private */


romaricpascal:
Really like the ideas of only keeping the comments that we need. A couple of questions as I looked into it:
 - why do we need to preserve the comments for `I18n`? And more generally, what would be the rule of thumbs for using `@preserve`? 
 - do you think it'd be worth/possible to automate the addition of `@preserve`? I was thinking of something like adding it to any JSDoc block that doesn't have an `@private` using a Babel transform, but that likely depends on the rules for what needs that `@preserve` tag.

colinrotherham:
Thanks @romaricpascal

> * why do we need to preserve the comments for `I18n`? And more generally, what would be the rule of thumbs for using `@preserve`?

There are a few things flagged `@private` that can't really be 😣

For example, the JSDoc compiler sees `I18n` and `TranslationPluralForms` are both used in public documentation. We're having to [suppress warnings about public-but-private things in **typedoc.config.js**](https://github.com/alphagov/govuk-frontend/pull/3820/commits/4bd6f819fe340d93a5d30c4b799ec829a72d66e4)

So it made sense to preserve all class, constructor and public function JSDoc comments

>* do you think it'd be worth/possible to automate the addition of `@preserve`? I was thinking of something like adding it to any JSDoc block that doesn't have an `@private` using a Babel transform, but that likely depends on the rules for what needs that `@preserve` tag.

We don't have many, I'd probably keep it manual but your idea would work with [`{ shouldPrintComment }`](https://babeljs.io/docs/options#shouldprintcomment)

Pretty sure it keeps `@license` too by default

colinrotherham:
@romaricpascal After we discussed `@internal` the other day I've updated this PR

I've taken your idea in https://github.com/alphagov/govuk-frontend/pull/3958#issuecomment-1635904949 to do things automatically:

1. Remove `@internal` comments
1. Remove `@private` comments
1. But keep `@preserve` comments

It's actually quite nice to see our descriptive developer comments preserved 😊

```patch
--- A/packages/govuk-frontend/dist/govuk/components/character-count/character-count.mjs
+++ B/packages/govuk-frontend/dist/govuk/components/character-count/character-count.mjs
@@ -19,40 +19,15 @@
    * @param {CharacterCountConfig} [config] - Character count config
    */
   constructor($module, config) {
-    /** @private */
     this.$module = void 0;
-    /** @private */
     this.$textarea = void 0;
-    /**
-     * @private
-     * @type {HTMLElement | null}
-     */
     this.$visibleCountMessage = null;
-    /**
-     * @private
-     * @type {HTMLElement | null}
-     */
     this.$screenReaderCountMessage = null;
-    /**
-     * @private
-     * @type {number | null}
-     */
     this.lastInputTimestamp = null;
-    /** @private */
     this.lastInputValue = '';
-    /**
-     * @private
-     * @type {number | null}
-     */
     this.valueChecker = null;
-    /**
-     * @private
-     * @type {CharacterCountConfig}
-     */
     this.config = void 0;
-    /** @private */
     this.i18n = void 0;
-    /** @private */
     this.maxLength = Infinity;
     if (!($module instanceof HTMLElement) || !document.body.classList.contains('govuk-frontend-supported')) {
       return this;
@@ -148,15 +123,6 @@
     // added to the page dynamically, so update now too.
     this.updateCountMessage();
   }
-
-  /**
-   * Bind change events
-   *
-   * Set up event listeners on the $textarea so that the count messages update
-   * when the user types.
-   *
-   * @private
-   */
   bindChangeEvents() {
     this.$textarea.addEventListener('keyup', () => this.handleKeyUp());
 
@@ -164,35 +130,10 @@
     this.$textarea.addEventListener('focus', () => this.handleFocus());
     this.$textarea.addEventListener('blur', () => this.handleBlur());
   }
-
-  /**
-   * Handle key up event
-   *
-   * Update the visible character counter and keep track of when the last update
-   * happened for each keypress
-   *
-   * @private
-   */
   handleKeyUp() {
     this.updateVisibleCountMessage();
     this.lastInputTimestamp = Date.now();
   }
-
-  /**
-   * Handle focus event
-   *
-   * Speech recognition software such as Dragon NaturallySpeaking will modify the
-   * fields by directly changing its `value`. These changes don't trigger events
-   * in JavaScript, so we need to poll to handle when and if they occur.
-   *
-   * Once the keyup event hasn't been detected for at least 1000 ms (1s), check if
-   * the textarea value has changed and update the count message if it has.
-   *
-   * This is so that the update triggered by the manual comparison doesn't
-   * conflict with debounced KeyboardEvent updates.
-   *
-   * @private
-   */
   handleFocus() {
     this.valueChecker = window.setInterval(() => {
       if (!this.lastInputTimestamp || Date.now() - 500 >= this.lastInputTimestamp) {
@@ -200,24 +141,10 @@
       }
     }, 1000);
   }
-
-  /**
-   * Handle blur event
-   *
-   * Stop checking the textarea value once the textarea no longer has focus
-   *
-   * @private
-   */
   handleBlur() {
     // Cancel value checking on blur
     clearInterval(this.valueChecker);
   }
-
-  /**
-   * Update count message if textarea value has changed
-   *
-   * @private
-   */
   updateIfValueChanged() {
     if (this.$textarea.value !== this.lastInputValue) {
       this.lastInputValue = this.$textarea.value;
@@ -224,25 +151,10 @@
       this.updateCountMessage();
     }
   }
-
-  /**
-   * Update count message
-   *
-   * Helper function to update both the visible and screen reader-specific
-   * counters simultaneously (e.g. on init)
-   *
-   * @private
-   */
   updateCountMessage() {
     this.updateVisibleCountMessage();
     this.updateScreenReaderCountMessage();
   }
-
-  /**
-   * Update visible count message
-   *
-   * @private
-   */
   updateVisibleCountMessage() {
     const remainingNumber = this.maxLength - this.count(this.$textarea.value);
 
@@ -268,12 +180,6 @@
     // Update message
     this.$visibleCountMessage.innerText = this.getCountMessage();
   }
-
-  /**
-   * Update screen reader count message
-   *
-   * @private
-   */
   updateScreenReaderCountMessage() {
     // If over the threshold, remove the aria-hidden attribute, allowing screen
     // readers to announce the content of the element.
@@ -286,15 +192,6 @@
     // Update message
     this.$screenReaderCountMessage.innerText = this.getCountMessage();
   }
-
-  /**
-   * Count the number of characters (or words, if `config.maxwords` is set)
-   * in the given text
-   *
-   * @private
-   * @param {string} text - The text to count the characters of
-   * @returns {number} the number of characters (or words) in the text
-   */
   count(text) {
     if ('maxwords' in this.config && this.config.maxwords) {
       const tokens = text.match(/\S+/g) || []; // Matches consecutive non-whitespace chars
@@ -303,28 +200,11 @@
       return text.length;
     }
   }
-
-  /**
-   * Get count message
-   *
-   * @private
-   * @returns {string} Status message
-   */
   getCountMessage() {
     const remainingNumber = this.maxLength - this.count(this.$textarea.value);
     const countType = 'maxwords' in this.config && this.config.maxwords ? 'words' : 'characters';
     return this.formatCountMessage(remainingNumber, countType);
   }
-
-  /**
-   * Formats the message shown to users according to what's counted
-   * and how many remain
-   *
-   * @private
-   * @param {number} remainingNumber - The number of words/characaters remaining
-   * @param {string} countType - "words" or "characters"
-   * @returns {string} Status message
-   */
   formatCountMessage(remainingNumber, countType) {
     if (remainingNumber === 0) {
       return this.i18n.t(`${countType}AtLimit`);
@@ -334,18 +214,6 @@
       count: Math.abs(remainingNumber)
     });
   }
-
-  /**
-   * Check if count is over threshold
-   *
-   * Checks whether the value is over the configured threshold for the input.
-   * If there is no configured threshold, it is set to 0 and this function will
-   * always return true.
-   *
-   * @private
-   * @returns {boolean} true if the current count is over the config.threshold
-   *   (or no threshold is set)
-   */
   isOverThreshold() {
     // No threshold means we're always above threshold so save some computation
     if (!this.config.threshold) {
```

colinrotherham:
I'll add another example with all comments removed except JSDoc

```patch
--- A/packages/govuk-frontend/dist/govuk/components/character-count/character-count.mjs
+++ B/packages/govuk-frontend/dist/govuk/components/character-count/character-count.mjs
@@ -12,6 +12,8 @@
  *
  * You can configure the message to only appear after a certain percentage
  * of the available characters/words has been entered.
+ *
+ * @preserve
  */
 class CharacterCount {
   /**
@@ -19,40 +21,15 @@
    * @param {CharacterCountConfig} [config] - Character count config
    */
   constructor($module, config) {
-    /** @private */
     this.$module = void 0;
-    /** @private */
     this.$textarea = void 0;
-    /**
-     * @private
-     * @type {HTMLElement | null}
-     */
     this.$visibleCountMessage = null;
-    /**
-     * @private
-     * @type {HTMLElement | null}
-     */
     this.$screenReaderCountMessage = null;
-    /**
-     * @private
-     * @type {number | null}
-     */
     this.lastInputTimestamp = null;
-    /** @private */
     this.lastInputValue = '';
-    /**
-     * @private
-     * @type {number | null}
-     */
     this.valueChecker = null;
-    /**
-     * @private
-     * @type {CharacterCountConfig}
-     */
     this.config = void 0;
-    /** @private */
     this.i18n = void 0;
-    /** @private */
     this.maxLength = Infinity;
     if (!($module instanceof HTMLElement) || !document.body.classList.contains('govuk-frontend-supported')) {
       return this;
@@ -61,17 +38,7 @@
     if (!($textarea instanceof HTMLTextAreaElement || $textarea instanceof HTMLInputElement)) {
       return this;
     }
-
-    // Read config set using dataset ('data-' values)
     const datasetConfig = normaliseDataset($module.dataset);
-
-    // To ensure data-attributes take complete precedence, even if they change the
-    // type of count, we need to reset the `maxlength` and `maxwords` from the
-    // JavaScript config.
-    //
-    // We can't mutate `config`, though, as it may be shared across multiple
-    // components inside `initAll`.
-    /** @type {CharacterCountConfig} */
     let configOverrides = {};
     if ('maxwords' in datasetConfig || 'maxlength' in datasetConfig) {
       configOverrides = {
@@ -81,11 +48,8 @@
     }
     this.config = mergeConfigs(CharacterCount.defaults, config || {}, configOverrides, datasetConfig);
     this.i18n = new I18n(extractConfigByNamespace(this.config, 'i18n'), {
-      // Read the fallback if necessary rather than have it set in the defaults
       locale: closestAttributeValue($module, 'lang')
     });
-
-    // Determine the limit attribute (characters or words)
     if ('maxwords' in this.config && this.config.maxwords) {
       this.maxLength = this.config.maxwords;
     } else if ('maxlength' in this.config && this.config.maxlength) {
@@ -99,31 +63,17 @@
     if (!$textareaDescription) {
       return;
     }
-
-    // Inject a description for the textarea if none is present already
-    // for when the component was rendered with no maxlength, maxwords
-    // nor custom textareaDescriptionText
     if ($textareaDescription.innerText.match(/^\s*$/)) {
       $textareaDescription.innerText = this.i18n.t('textareaDescription', {
         count: this.maxLength
       });
     }
-
-    // Move the textarea description to be immediately after the textarea
-    // Kept for backwards compatibility
     this.$textarea.insertAdjacentElement('afterend', $textareaDescription);
-
-    // Create the *screen reader* specific live-updating counter
-    // This doesn't need any styling classes, as it is never visible
     const $screenReaderCountMessage = document.createElement('div');
     $screenReaderCountMessage.className = 'govuk-character-count__sr-status govuk-visually-hidden';
     $screenReaderCountMessage.setAttribute('aria-live', 'polite');
     this.$screenReaderCountMessage = $screenReaderCountMessage;
     $textareaDescription.insertAdjacentElement('afterend', $screenReaderCountMessage);
-
-    // Create our live-updating counter element, copying the classes from the
-    // textarea description for backwards compatibility as these may have been
-    // configured
     const $visibleCountMessage = document.createElement('div');
     $visibleCountMessage.className = $textareaDescription.className;
     $visibleCountMessage.classList.add('govuk-character-count__status');
@@ -130,69 +80,21 @@
     $visibleCountMessage.setAttribute('aria-hidden', 'true');
     this.$visibleCountMessage = $visibleCountMessage;
     $textareaDescription.insertAdjacentElement('afterend', $visibleCountMessage);
-
-    // Hide the textarea description
     $textareaDescription.classList.add('govuk-visually-hidden');
-
-    // Remove hard limit if set
     this.$textarea.removeAttribute('maxlength');
     this.bindChangeEvents();
-
-    // When the page is restored after navigating 'back' in some browsers the
-    // state of form controls is not restored until *after* the DOMContentLoaded
-    // event is fired, so we need to sync after the pageshow event.
     window.addEventListener('pageshow', () => this.updateCountMessage());
-
-    // Although we've set up handlers to sync state on the pageshow event, init
-    // could be called after those events have fired, for example if they are
-    // added to the page dynamically, so update now too.
     this.updateCountMessage();
   }
-
-  /**
-   * Bind change events
-   *
-   * Set up event listeners on the $textarea so that the count messages update
-   * when the user types.
-   *
-   * @private
-   */
   bindChangeEvents() {
     this.$textarea.addEventListener('keyup', () => this.handleKeyUp());
-
-    // Bind focus/blur events to start/stop polling
     this.$textarea.addEventListener('focus', () => this.handleFocus());
     this.$textarea.addEventListener('blur', () => this.handleBlur());
   }
-
-  /**
-   * Handle key up event
-   *
-   * Update the visible character counter and keep track of when the last update
-   * happened for each keypress
-   *
-   * @private
-   */
   handleKeyUp() {
     this.updateVisibleCountMessage();
     this.lastInputTimestamp = Date.now();
   }
-
-  /**
-   * Handle focus event
-   *
-   * Speech recognition software such as Dragon NaturallySpeaking will modify the
-   * fields by directly changing its `value`. These changes don't trigger events
-   * in JavaScript, so we need to poll to handle when and if they occur.
-   *
-   * Once the keyup event hasn't been detected for at least 1000 ms (1s), check if
-   * the textarea value has changed and update the count message if it has.
-   *
-   * This is so that the update triggered by the manual comparison doesn't
-   * conflict with debounced KeyboardEvent updates.
-   *
-   * @private
-   */
   handleFocus() {
     this.valueChecker = window.setInterval(() => {
       if (!this.lastInputTimestamp || Date.now() - 500 >= this.lastInputTimestamp) {
@@ -200,24 +102,9 @@
       }
     }, 1000);
   }
-
-  /**
-   * Handle blur event
-   *
-   * Stop checking the textarea value once the textarea no longer has focus
-   *
-   * @private
-   */
   handleBlur() {
-    // Cancel value checking on blur
     clearInterval(this.valueChecker);
   }
-
-  /**
-   * Update count message if textarea value has changed
-   *
-   * @private
-   */
   updateIfValueChanged() {
     if (this.$textarea.value !== this.lastInputValue) {
       this.lastInputValue = this.$textarea.value;
@@ -224,37 +111,17 @@
       this.updateCountMessage();
     }
   }
-
-  /**
-   * Update count message
-   *
-   * Helper function to update both the visible and screen reader-specific
-   * counters simultaneously (e.g. on init)
-   *
-   * @private
-   */
   updateCountMessage() {
     this.updateVisibleCountMessage();
     this.updateScreenReaderCountMessage();
   }
-
-  /**
-   * Update visible count message
-   *
-   * @private
-   */
   updateVisibleCountMessage() {
     const remainingNumber = this.maxLength - this.count(this.$textarea.value);
-
-    // If input is over the threshold, remove the disabled class which renders the
-    // counter invisible.
     if (this.isOverThreshold()) {
       this.$visibleCountMessage.classList.remove('govuk-character-count__message--disabled');
     } else {
       this.$visibleCountMessage.classList.add('govuk-character-count__message--disabled');
     }
-
-    // Update styles
     if (remainingNumber < 0) {
       this.$textarea.classList.add('govuk-textarea--error');
       this.$visibleCountMessage.classList.remove('govuk-hint');
@@ -264,26 +131,14 @@
       this.$visibleCountMessage.classList.remove('govuk-error-message');
       this.$visibleCountMessage.classList.add('govuk-hint');
     }
-
-    // Update message
     this.$visibleCountMessage.innerText = this.getCountMessage();
   }
-
-  /**
-   * Update screen reader count message
-   *
-   * @private
-   */
   updateScreenReaderCountMessage() {
-    // If over the threshold, remove the aria-hidden attribute, allowing screen
-    // readers to announce the content of the element.
     if (this.isOverThreshold()) {
       this.$screenReaderCountMessage.removeAttribute('aria-hidden');
     } else {
       this.$screenReaderCountMessage.setAttribute('aria-hidden', 'true');
     }
-
-    // Update message
     this.$screenReaderCountMessage.innerText = this.getCountMessage();
   }
 
@@ -297,7 +152,7 @@
    */
   count(text) {
     if ('maxwords' in this.config && this.config.maxwords) {
+      const tokens = text.match(/\S+/g) || [];
-      const tokens = text.match(/\S+/g) || []; // Matches consecutive non-whitespace chars
       return tokens.length;
     } else {
       return text.length;
@@ -347,26 +202,14 @@
    *   (or no threshold is set)
    */
   isOverThreshold() {
-    // No threshold means we're always above threshold so save some computation
     if (!this.config.threshold) {
       return true;
     }
-
-    // Determine the remaining number of characters/words
     const currentLength = this.count(this.$textarea.value);
     const maxLength = this.maxLength;
     const thresholdValue = maxLength * this.config.threshold / 100;
     return thresholdValue <= currentLength;
   }
-
-  /**
-   * Character count default config
-   *
-   * @see {@link CharacterCountConfig}
-   * @constant
-   * @default
-   * @type {CharacterCountConfig}
-   */
 }
 
 /**
@@ -455,7 +298,6 @@
 CharacterCount.defaults = Object.freeze({
   threshold: 0,
   i18n: {
-    // Characters
     charactersUnderLimit: {
       one: 'You have %{count} character remaining',
       other: 'You have %{count} characters remaining'
@@ -465,7 +307,6 @@
       one: 'You have %{count} character too many',
       other: 'You have %{count} characters too many'
     },
-    // Words
     wordsUnderLimit: {
       one: 'You have %{count} word remaining',
       other: 'You have %{count} words remaining'
```

packages/govuk-frontend/babel.config.js
```diff
@@ -11,6 +11,25 @@ module.exports = function (api) {
...
+          return true
+        }
+
+        // Assume all comments are public unless
+        // tagged with `@private` or `@internal`
+        const isPrivate = ['* @internal', '* @private']
+          .some((tag) => comment.includes(tag))
+
+        // Flag any JSDoc comments worth keeping
+        const isDocumentation = ['* @param', '* @returns', '* @typedef']
```
romaricpascal, line 26: **question**: Would matching `/**` help us not have to add `@preserve` tags and limit the manual  flagging of comments we want to keep (like all the classes JSDoc blocks)?

packages/govuk-frontend/src/govuk/common/index.mjs
```diff
@@ -4,8 +4,6 @@
  * IMPORTANT: If a helper require a polyfill, please isolate it in its own module
  * so that the polyfill can be properly tree-shaken and does not burden
  * the components that do not need that helper
- *
- * @module common/index
```
romaricpascal, line 8: Good shout, that wasn't super useful 🙌🏻

packages/govuk-frontend/src/govuk/i18n.mjs
```diff
@@ -70,6 +72,7 @@ export class I18n {
    * Takes a translation string with placeholders, and replaces the placeholders
    * with the provided data
    *
+   * @internal
```
romaricpascal, line 75: **question** I think outside of `t`, `I18n`'s methods are `@private` (to the class) rather than `@internal` (to our package). Does flagging them as internal help with something else (TypeScript maybe)?

romaricpascal (commented):
Just two little questions, one on the implementation of what picks what we remove, the other on the tagging of the methods of I18n. Other than that, keen to get the heavy reduction of size in our package 😊

packages/govuk-frontend/babel.config.js
colinrotherham: Sometimes we just have non-documentation comments like:

```mjs
/**
 * Hello Romaric
 */
```

So they get removed too

packages/govuk-frontend/src/govuk/i18n.mjs
colinrotherham: No not really, we'd just need to watch out for these ones:

1. `getPluralSuffix()`
1. `getPluralRulesForLocale()`
1. `selectPluralFormUsingFallbackRules()`

We call them externally from tests so it's wrong to make them `@private`

romaricpascal: Is that something that TypeScript would flag? In which case we can go with the process of trying to lock things as `@private`, if TypeScript flags (for tests, for ex) `@internal`.

colinrotherham: Sadly we also add mocks/spies to private methods:

```mjs
jest.spyOn(i18n, 'hasIntlPluralRulesSupport')
  .mockImplementation(() => false)
```

I'll keep them all as `@internal` and we can assess how we test "non-public" class interfaces another day!

romaricpascal: Yeah, may be more efficient to go for `@internal` all the way + some convention (the commonplace leading `_`) for things that are meant to only be accessed from within the class. I've scheduled something in dev catch up to discuss the pains of marking things properly `@private` so we can chose knowingly whether to embrace it.

PR #3959, opened 2023-07-13
Add Percy screenshots for “inverse” components

colinrotherham:
We only run Percy visual regression tests on component default examples

Should we consider "inverse" examples just as important?

Flagging this because we nearly missed some inverse styles in the last release:

* https://github.com/alphagov/govuk-frontend/pull/3915


shared/tasks/browser.mjs
```diff
@@ -30,7 +30,21 @@ export async function screenshots () {
 
   // Screenshot components
   for (const componentName of componentNames) {
+    const componentExamples = await getExamples(componentName)
+
+    // Check for optional "inverse" example
+    const exampleInverted = Object.entries(componentExamples)
+      .find(([exampleName, data]) => typeof data.classes === 'string' && data.classes.endsWith('--inverse'))
```
querkmachine, line 37: I'm a little antsy about this check being what determines whether Percy considers it inverse or not. 

Given the `classes` parameter is a generic one, there is no guarantee that `--inverse` be the last part of the string—there could be other classes after it. Maybe we have something like `govuk-button--inverse govuk-button--small` at some point? 

Not going to block on that basis alone, as it works for our current needs, but doing something more future-proof now might save us some confusion/debugging/refactoring down the line.

querkmachine (approved):

shared/tasks/browser.mjs
colinrotherham: Yeah you're spot on, that extra class example is a good one 😣

Have a look now? I've edited the component data so we always have both:

```yaml
- name: default
- name: inverse
```

Can just match on the example name `inverse` now

querkmachine: Yeah, that seems a better approach to me.

This has made me wonder why until now we've only done the 'default' versions of components. There seems to be a few places where having Percy keeping track of prominent variants would be useful (the two different designs of the Pagination component, an open Details component, start Buttons, etc.) but that's outside of the immediate scope of this PR. We can add those in later if we think it's valuable.

colinrotherham: Thanks @querkmachine

They're really quick to add to moved your suggestions to:

* https://github.com/alphagov/govuk-frontend/pull/3962

Issue #3674, opened 2023-05-23
Summary list row actions within summary cards must specify which summary card they're part of


CharlotteDowns:
This issue is from May 2023 external accessibility audit report.

**WCAG Reference:**
2.4.4 Link Purpose (In Context) (Level A)
[Understanding Link Purpose (In Context)](https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-in-context.html) | [How to Meet Link Purpose (In Context)](https://www.w3.org/WAI/WCAG21/quickref/#link-purpose-in-context)

2.4.9 Link Purpose (Link Only) (Level AAA)
[Understanding Link Purpose (Link Only)](https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-link-only.html) | [How to Meet Link Purpose (Link Only)](https://www.w3.org/WAI/WCAG21/quickref/#link-purpose-link-only)

**Issue ID:**
DAC_Link_Purpose_02

**URLs:**
[https://govuk-frontend-pr-3503.herokuapp.com/full-page-examples/work-history#/delete](https://govuk-frontend-pr-3503.herokuapp.com/full-page-examples/work-history#/delete)

## Screen shots
<img width="637" alt="Summary card component with 'change' link" src="https://github.com/alphagov/govuk-frontend/assets/56260216/2b31d608-754b-48e2-aa3a-0b3bb71276fc">

The ‘change’ links have been provided with additional link text to inform users of screen reading assistive technologies that they relate to the different questions i.e., ‘job title’. However, additional context is required to understand their whole purpose as to which section and work they relate to. For example, a user would need to navigate backwards to locate the heading to which they are sectioned under, and as there are multiple ‘change job title’ links, this can be confusing for users.

## Current code ref(s)
`#main-content > div > div > div:nth-child(3) > div.govuk-summary-card__content > dl > div:nth-child(1) > dd.govuk-summary-list__actions > a`

## Screen-reader comments
> “I located multiple links on the page which read to me as ‘Delete work history’. These links related to different sections of the page and would delete different areas although at the time of testing did not perform a function. I was not aware when browsing out of context that the links would perform a different action or what would be deleted as the links did not contain enough description within the link text to make this obvious. Giving each link text clear and unique information that can be understood either in or out of context will prevent any confusion and will ensure that I can select easily.”

## Solution
Ensure that all links are descriptive of their purpose and functionality, allowing users of screen reading assistive technologies to confidently interact and understand the link.
[...]

PR #3961, opened 2023-07-14
Append card title to action links inside of a Summary Card

### LINKED ISSUE #3673: Summary card actions must specify which card they are for

**WCAG Reference:** 
2.4.4 Link Purpose (In Context) (Level A)
[Understanding Link Purpose (In Context)](https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-in-context.html) | [How to Meet Link Purpose (In Context)](https://www.w3.org/WAI/WCAG21/quickref/#link-purpose-in-context)

2.4.9 Link Purpose (Link Only) (Level AAA)
[Understanding Link Purpose (Link Only)](https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-link-only.html) | [How to Meet Link Purpose (Link Only)](https://www.w3.org/WAI/WCAG21/quickref/#link-purpose-link-only)

**Issue ID:**
DAC_Link_Purpose_01

**URLs:**
[https://govuk-frontend-pr-3503.herokuapp.com/full-page-examples/work-history#/delete](https://govuk-frontend-pr-3503.herokuapp.com/full-page-examples/work-history#/delete)

## Screen shots
Summary card component with 'delete work history' link

The ‘Delete work history’ links for each section, rely on surrounding content to understand their context and which work they will be deleting. As the link is not marked up in the same list item, table cell, or using different methods of aria, the link requires a user to move their focus to understand which work history they will be deleting.

## Current code ref(s)
`#main-content > div > div > div:nth-child(2) > div.govuk-summary-card__title-wrapper > div > a`

## Solution
Ensure that all links are descriptive of their purpose and functionality, allowing users of screen reading assistive technologies to confidently interact and understand the link.
The best way to make sure of this, is to locate links where some additional link context is needed to understand the purpose of the link. For each link:

1. Check whether the context is contained in the same sentence, paragraph, list item, table cell, or associated table headers.
2. Check whether the link context can be programmatically determined in some other way, for example, by using a WAI-ARIA property such as aria-label, aria-labelledby or aria-describedby on the link to provide sufficient context.

Consider adding additional text to the link which can be visually hidden to maintain the visual aesthetic.

querkmachine:
It was raised in the recent DAC audit that pages with multiple Summary Cards had the potential to fail the Level A criterion [Link Purpose (In Context)](https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-in-context.html) and the Level AAA criterion [Link Purpose (Link Only)](https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-link-only), due to the repetition of links with identical labels—unless the implementor went out of their way to manually provide the required distinctions.

This PR modifies the Nunjucks template for Summary Lists so that, if a card title has been provided, the title is automatically appended to any action links within the card. This is applied to both the card-level and individual row's action links. 

This PR resolves #3673 add #3674.

## Changes
- Updated the component's internal `_actionLink` macro now takes a second, optional `cardTitle` parameter.
- Updated instances of `_actionLink` to pass through the `params.card.title` object, if it exists. 

## Thoughts
- This is currently hardcoding a content order (and parentheses) that might not be preferable in all localisations. I'm hoping that, as two distinct pieces of information that are intended as separate phrases, this might be okay, but I'm not sure.
- Piping `params.card.title.html` into a visually-hidden span feels a bit weird and like it may have adverse side effects depending on what HTML is provided, however we have no guarantee that `params.card.title.text` is set in all situations.

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/summary-list/template.njk
+++ b/packages/govuk-frontend/src/govuk/components/summary-list/template.njk
@@ -1,9 +1,12 @@
-{%- macro _actionLink(action) %}
+{%- macro _actionLink(action, cardTitle) %}
   <a class="govuk-link {%- if action.classes %} {{ action.classes }}{% endif %}" href="{{ action.href }}" {%- for attribute, value in action.attributes %} {{attribute}}="{{value}}"{% endfor %}>
-    {{ action.html | safe if action.html else action.text }}
-    {%- if action.visuallyHiddenText -%}
-      <span class="govuk-visually-hidden"> {{ action.visuallyHiddenText }}</span>
-    {% endif -%}
+    {{- action.html | safe if action.html else action.text -}}
+    {%- if action.visuallyHiddenText or cardTitle -%}
+      <span class="govuk-visually-hidden">
+        {%- if action.visuallyHiddenText %} {{ action.visuallyHiddenText }}{% endif -%}
+        {%- if cardTitle %} ({{ cardTitle.html | safe if cardTitle.html else cardTitle.text }}){% endif -%}
+      </span>
+    {%- endif -%}
   </a>
 {% endmacro -%}
 
@@ -21,13 +24,13 @@
       {%- if params.actions.items.length -%}
         {%- if params.actions.items.length == 1 -%}
           <div class="govuk-summary-card__actions {%- if params.actions.classes %} {{ params.actions.classes }}{% endif %}">
-            {{- _actionLink(params.actions.items[0]) -}}
+            {{- _actionLink(params.actions.items[0], params.title) -}}
           </div>
         {%- else -%}
           <ul class="govuk-summary-card__actions {%- if params.actions.classes %} {{ params.actions.classes }}{% endif %}">
             {%- for action in params.actions.items -%}
               <li class="govuk-summary-card__action">
-                {{- _actionLink(action) -}}
+                {{- _actionLink(action, params.title) -}}
               </li>
             {%- endfor -%}
           </ul>
@@ -61,12 +64,12 @@
           {% if row.actions.items.length %}
             <dd class="govuk-summary-list__actions {%- if row.actions.classes %} {{ row.actions.classes }}{% endif %}">
               {% if row.actions.items.length == 1 %}
-                {{ _actionLink(row.actions.items[0]) | indent(16 if params.card else 12) | trim }}
+                {{ _actionLink(row.actions.items[0], params.card.title) | indent(16 if params.card else 12) | trim }}
               {% else %}
                 <ul class="govuk-summary-list__actions-list">
                   {%- for action in row.actions.items -%}
                     <li class="govuk-summary-list__actions-list-item">
-                      {{- _actionLink(action) | indent(22 if params.card else 18) | trim -}}
+                      {{- _actionLink(action, params.card.title) | indent(22 if params.card else 18) | trim -}}
                     </li>
                   {%- endfor -%}
                 </ul>
--- a/packages/govuk-frontend/src/govuk/components/summary-list/template.test.js
+++ b/packages/govuk-frontend/src/govuk/components/summary-list/template.test.js
@@ -268,7 +268,7 @@ describe('Summary list', () => {
         const $singleAction = $('.govuk-summary-card__actions > a')
         const $actionItems = $('.govuk-summary-card__action')
         expect($actionItems.length).toBe(0)
-        expect($singleAction.text().trim()).toBe('My lonely action')
+        expect($singleAction.text().trim()).toBe('My lonely action (Undergraduate teaching assistant)')
       })
     })
 
--- a/packages/govuk-frontend/src/govuk/helpers/_visually-hidden.scss
+++ b/packages/govuk-frontend/src/govuk/helpers/_visually-hidden.scss
@@ -33,6 +33,11 @@
   // causes content to wrap 1 word per line:
   // https://medium.com/@jessebeach/beware-smushed-off-screen-accessible-text-5952a4c2cbfe
   white-space: nowrap if($important, !important, null);
+
+  // Prevent users from selecting or copying visually-hidden text. This prevents
+  // a user unintentionally copying more text than they intended and needing to
+  // manually trim it down again.
+  user-select: none;
 }
 
 /// Hide an element visually, but have it available for screen readers whilst
@@ -64,6 +69,11 @@
   // https://medium.com/@jessebeach/beware-smushed-off-screen-accessible-text-5952a4c2cbfe
   white-space: nowrap if($important, !important, null);
 
+  // Prevent users from selecting or copying visually-hidden text. This prevents
+  // a user unintentionally copying more text than they intended and needing to
+  // manually trim it down again.
+  user-select: none;
+
   &:active,
   &:focus {
     position: static if($important, !important, null);
@@ -77,5 +87,8 @@
     clip-path: none if($important, !important, null);
 
     white-space: inherit if($important, !important, null);
+
+    // Allow the text to be selectable now it's visible
+    user-select: text;
   }
 }


packages/govuk-frontend/src/govuk/components/summary-list/template.njk
```diff
@@ -1,6 +1,9 @@
-{%- macro _actionLink(action) %}
+{%- macro _actionLink(action, cardTitle = null) %}
```
colinrotherham, line 1: Can probably leave the Nunjucks `undefined` default?

Had a quick check and tend to use `!== undefined` elsewhere for optional things

Although any falsy value will work with `{%- if cardTitle %}`

```suggestion
{%- macro _actionLink(action, cardTitle) %}
```

colinrotherham (approved):
Looks brill from a code point of view

Do we know if this is best approach?

Versus using the **`<h2>`** `govuk-summary-card__title` with `aria-describedby`?

querkmachine:
That certainly seems like a possibility that might work in practice (assuming the AT announces `aria-describedby`, though I'm not sure if doing that is compliant with the listed WCAG criteria, as the link text itself would not be unique and would appear identical in contexts like VoiceOver's links view.

CC @davidc-gds in case you have a view on this.

davidc-gds:
Hi @querkmachine from a information hierarchy standpoint, would it make sense to put the unique element (the action) first, and then the card title second?

We can take the '[Summary list as a summary card with actions plus summary list actions](https://govuk-frontend-pr-3961.herokuapp.com/components/summary-list/as-a-summary-card-with-actions-plus-summary-list-actions/preview)' PR preview example as a case study.

Right now it looks to be read out as:
> "Undergraduate teaching assistant: Delete job history"
> "Undergraduate teaching assistant: Withdraw job history"

I'm suggesting that since the unique element is the action name, we should reverse the order of the items, like so:
> "Delete job history: Undergraduate teaching assistant"
> "Withdraw job history: Undergraduate teaching assistant"

Thoughts , possibly from @claireashworth?

davidc-gds:
> That certainly seems like a possibility that might work in practice (assuming the AT consistently announces `aria-describedby`), though I'm not sure if doing that is compliant with the listed WCAG criteria, as the link text itself would not be unique and would appear identical in contexts like VoiceOver's links view.
> 
> CC @davidc-gds in case you have a view on this.

I tested the `ariadescribedby` option in a few screen readers, and generally it looked like it worked for some and didn't for others. I think to be robust we can go ahead with the `govuk-visually-hidden` currently in use here.

I tested an `aria-describedby` version with the VoiceOver 'rotor' list of links, and it didn't show the describedby piece in the link label. I couldn't test the 'list of links' features in NVDA or JAWS quick enough, mainly because my computer doesn't have an `insert` key, and I couldn't troubleshoot how to open the list in the time I had.

So in conclusion:
- this current `govuk-visually-hidden` method works in all 3 screen readers I tested
- the `aria-describedby` was hit and miss, but we'd be fine using it as its totally valid use of that ARIA attribute

The only consistent downside for `govuk-visually-hidden` is that it always makes copy/pasting the content a real mess:

Copy/paste using `govuk-visually-hidden`:
> Undergraduate teaching assistant
> Undergraduate teaching assistant:Delete job history
> Undergraduate teaching assistant:Withdraw job history
> Name	Firstname Lastname	
> Undergraduate teaching assistant:EditnameUndergraduate teaching assistant:Deletename
> Date of birth	13/08/1980	Undergraduate teaching assistant:Changedate of birth

Without using `govuk-visually-hidden`:
> Undergraduate teaching assistant
> Delete job history
> Withdraw job history
> Name	Firstname Lastname	
> Edit Delete
> Date of birth	13/08/1980	Change

querkmachine:
@davidc-gds Swapping the order makes sense to me.

The text-selection problem makes me wonder whether we want to set `user-select: none` on visually hidden text, but that's probably a topic for wider discussion.

claireashworth:
Having the action first makes sense to me as well.

querkmachine:
I've pushed changes to place the card title after the action name.

I've also added `user-select: none` to visually-hidden elements for the time being, so that it can be AT tested in situ.

davidc-gds:
@querkmachine I may be doing it wrong, but I can still copy/paste the hidden text, even with `user-select: none` applied to `govuk-visually-hidden` on [the PR deployment](https://govuk-frontend-pr-3961.herokuapp.com/components/summary-list).

I'm using cmd + c and cmd+ v on a Mac, on Chrome.

querkmachine:
^ We found that David's macOS is configured is to paste with formatting by default, which seemingly always includes the visually-hidden text. This doesn't appear to be the default behaviour in macOS, needing to be configured in Settings → Keyboard → Keyboard Shortcuts... → App Shortcuts. 

This does not happen in Firefox or Safari, so it seems to be a case of Chromium purposefully ignoring the CSS rule. As this is a configuration of the user's environment, we probably can't practically do anything about it.

Pasting without formatting omits the visually-hidden text as expected.

packages/govuk-frontend/src/govuk/components/summary-list/template.njk
```diff
@@ -1,9 +1,12 @@
-{%- macro _actionLink(action) %}
+{%- macro _actionLink(action, cardTitle) %}
   <a class="govuk-link {%- if action.classes %} {{ action.classes }}{% endif %}" href="{{ action.href }}" {%- for attribute, value in action.attributes %} {{attribute}}="{{value}}"{% endfor %}>
-    {{ action.html | safe if action.html else action.text }}
+    {{- action.html | safe if action.html else action.text -}}
     {%- if action.visuallyHiddenText -%}
       <span class="govuk-visually-hidden"> {{ action.visuallyHiddenText }}</span>
     {% endif -%}
+    {%- if cardTitle -%}
+      <span class="govuk-visually-hidden"> ({{- cardTitle.html | safe if cardTitle.html else cardTitle.text }})</span>
```
colinrotherham, line 8: Not a blocker, but don't think this `-` does anything?

```suggestion
      <span class="govuk-visually-hidden"> ({{ cardTitle.html | safe if cardTitle.html else cardTitle.text }})</span>
```

packages/govuk-frontend/src/govuk/components/summary-list/template.njk
```diff
@@ -1,9 +1,12 @@
-{%- macro _actionLink(action) %}
+{%- macro _actionLink(action, cardTitle) %}
   <a class="govuk-link {%- if action.classes %} {{ action.classes }}{% endif %}" href="{{ action.href }}" {%- for attribute, value in action.attributes %} {{attribute}}="{{value}}"{% endfor %}>
-    {{ action.html | safe if action.html else action.text }}
+    {{- action.html | safe if action.html else action.text -}}
     {%- if action.visuallyHiddenText -%}
       <span class="govuk-visually-hidden"> {{ action.visuallyHiddenText }}</span>
```
colinrotherham, line 5: To keep HTML file size down, should combine these two `govuk-visually-hidden` spans where possible?

```html
<a class="govuk-link" href="#">
  Delete
  
  <!-- Hidden text #1 -->
  <span class="govuk-visually-hidden"> name</span>
  
  <!-- Hidden text #2 -->
  <span class="govuk-visually-hidden"> (Undergraduate teaching assistant)</span>
</a>
```

packages/govuk-frontend/src/govuk/components/summary-list/template.njk
```diff
@@ -61,12 +64,12 @@
...
             <dd class="govuk-summary-list__actions {%- if row.actions.classes %} {{ row.actions.classes }}{% endif %}">
               {% if row.actions.items.length == 1 %}
-                {{ _actionLink(row.actions.items[0]) | indent(16 if params.card else 12) | trim }}
+                {{ _actionLink(row.actions.items[0], params.card.title) | indent(16 if params.card else 12) | trim }}
               {% else %}
                 <ul class="govuk-summary-list__actions-list">
                   {%- for action in row.actions.items -%}
                     <li class="govuk-summary-list__actions-list-item">
-                      {{- _actionLink(action) | indent(22 if params.card else 18) | trim -}}
+                      {{- _actionLink(action, params.card.title) | indent(22 if params.card else 18) | trim -}}
```
colinrotherham, line 72: Mind if we re-assess these indents and empty space removal?

Rendered HTML indentation is a bit strange

```html
<dd class="govuk-summary-list__actions">
    <ul class="govuk-summary-list__actions-list"><li class="govuk-summary-list__actions-list-item"><a class="govuk-link" href="#">Edit<span class="govuk-visually-hidden"> name</span>
            <span class="govuk-visually-hidden"> (Undergraduate teaching assistant)</span>
            </a></li><li class="govuk-summary-list__actions-list-item"><a class="govuk-link" href="#">Delete<span class="govuk-visually-hidden"> name</span>
            <span class="govuk-visually-hidden"> (Undergraduate teaching assistant)</span>
            </a></li></ul>
</dd>
```

colinrotherham (approved):
Nice work, thanks for checking we picked the best solution 😊

Some comments on HTML output but see what you think

packages/govuk-frontend/src/govuk/components/summary-list/template.njk
querkmachine: I've ended up crunching the entire `<ul>` and its contents onto a single line at the moment. It's not great, but as the links contain semantically significant whitespace, it seemed like the simplest 'tidy' solution.

The whole component is a jumble of weird indentation (it switches from 2 spaces to 4 at various points) so it's not something I want to focus too much on trying to fix in this PR, but certainly a nice thing to try and achieve here and in other components at some point.

colinrotherham: Thanks @querkmachine

For reference the example above is now:

```html
<dd class="govuk-summary-list__actions">
    <ul class="govuk-summary-list__actions-list"><li class="govuk-summary-list__actions-list-item"><a class="govuk-link" href="#">Edit<span class="govuk-visually-hidden"> name (Undergraduate teaching assistant)</span></a></li><li class="govuk-summary-list__actions-list-item"><a class="govuk-link" href="#">Delete<span class="govuk-visually-hidden"> name (Undergraduate teaching assistant)</span></a></li></ul>
</dd>
```

Not related to this comment but worth a mention, the title wrapper similarly is:

```html
    <div class="govuk-summary-card__title-wrapper"><h2 class="govuk-summary-card__title">Undergraduate teaching assistant</h2><ul class="govuk-summary-card__actions"><li class="govuk-summary-card__action">  <a class="govuk-link" href="#">Delete job history<span class="govuk-visually-hidden"> (Undergraduate teaching assistant)</span></a>
</li><li class="govuk-summary-card__action">  <a class="govuk-link" href="#">Withdraw job history<span class="govuk-visually-hidden"> (Undergraduate teaching assistant)</span></a>
</li></ul>
</div>
```

PR #3964, opened 2023-07-17
Fix review app `SocketError: other side closed` issue

colinrotherham:
We keep having issues with `fetch()` failing in tests:

https://github.com/alphagov/govuk-frontend/actions/runs/5573254772/attempts/1
https://github.com/alphagov/govuk-frontend/actions/runs/5573282581/attempts/1

This is likely related to Review app setup/teardown from Jest dev server

### TypeError: fetch failed

```console
Cause:
    SocketError: other side closed
```

Since we had `keepalive` enabled previously in https://github.com/alphagov/govuk-frontend/pull/3045 this PR brings it back

If this doesn't help, we can look at automatic retries to allow for setup/teardown delays

### DIFF
--- a/shared/helpers/tests.js
+++ b/shared/helpers/tests.js
@@ -48,7 +48,10 @@ async function compileSassString (source, options = {}) {
  * @returns {Promise<Response>} Fetch response
  */
 const fetchPath = (path, options = {}) => {
-  return fetch(`http://localhost:${ports.app}${path}`, options)
+  return fetch(new URL(path, `http://localhost:${ports.app}`), {
+    keepalive: true,
+    ...options
+  })
 }
 
 /**
--- a/shared/helpers/tests.js
+++ b/shared/helpers/tests.js
@@ -1,6 +1,6 @@
 const { join } = require('path')
 
-const { paths } = require('govuk-frontend-config')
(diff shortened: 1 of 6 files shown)


shared/helpers/tests.js
```diff
@@ -40,6 +40,20 @@ async function compileSassString (source, options = {}) {
...
+/**
+ * Fetch by URL request path
+ *
+ * @param {string} path - URL request path
+ * @param {RequestInit} [options] - Fetch options
+ * @returns {Promise<Response>} Fetch response
+ */
+const fetchPath = (path, options = {}) => {
+  return fetch(new URL(path, `http://localhost:${ports.app}`), {
+    keepalive: true,
```
colinrotherham, line 52: Using `{ keepalive: true }` appears to avoid [`other side closed` errors in Node.js](https://github.com/nodejs/node/blob/7196946d7f59652c19dec278186165806b1e28ad/deps/undici/src/lib/client.js#L989-L995)

36degrees (approved):
Worth a try!

colinrotherham:
@36degrees Annoyingly we had one today in https://github.com/alphagov/govuk-frontend/actions/runs/5596668879/jobs/10234104679

Has it helped? Maybe

Issue #3639, opened 2023-05-15
Make sure the `docs` folder is up to date with new structure


stevenjmesser:
## What

Check that the documentation inside the repository (in `docs`) are up to date with the new repository structure, especially:

- the docs for releasing to npm
- the docs for pushing a pre-release as a branch on Github
- but also all the other pieces of documentation in there

## Why

We're making [substantial changes to our repository structure](https://github.com/alphagov/govuk-frontend/issues/3291) ahead of v5. We need to ensure that our docs reflect these changes so they can help future us and contributors appropriately.

Pushing a pre-release branch, especially, is really helpful for contributors as it allows them to install their changes as an npm package (without releasing to npm)

## Who needs to work on this

Developers, Tech writers?

## Who needs to review this

Developers, Tech writers?

## Done when

```[tasklist]
### We've reviewed
- [ ] The npm release docs (both from main and from support branch)
- [x] The docs for making a pre-release on Github
- [ ] The rest of the docs
```

romaricpascal:
Closed by #3975 (the PR was referencing the wrong issue).

PR #3977, opened 2023-07-19
Add pre-release documentation with version numbers

colinrotherham:
This PR adds pre-release documentation and sets a version number using `npm version`

* https://github.com/alphagov/govuk-frontend/issues/3639

### Pre-release version numbers

For example, for early preview of `v5.0.0` from branch `main` we could:

- Use version `premajor` to bump from `v4.7.0` to `v5.0.0-beta.0`
- Use version `preminor` to bump from `v4.7.0` to `v4.8.0-beta.0`
- Use version `prepatch` to bump from `v4.7.0` to `v4.7.1-beta.0`
- Use version `prerelease` to bump from `v5.0.0-beta.0` to `v5.0.0-beta.1`

This determines the `npm version` command to update **package.json** and **package-lock.json**:

```shell
npm version <VERSION NUMBER> --preid beta --no-git-tag-version --workspace govuk-frontend
```

Unlike previews, pre-releases must pick patch/minor/major bumps so it's hard to automatically version

### DIFF
--- a/docs/releasing/publishing-a-pre-release.md
+++ b/docs/releasing/publishing-a-pre-release.md
@@ -42,19 +42,13 @@ Developers should pair on pre-releases. When remote working, it can be useful to
    - Use `preminor` to bump from `v4.7.0` to `v4.8.0-beta.0`
    - Use `prepatch` to bump from `v4.7.0` to `v4.7.1-beta.0`
 
+   See the [versioning documentation](/docs/contributing/versioning.md) for more information.
+
    Alternatively, when publishing an update to an existing pre-release:
 
    - Use `prerelease` to bump from `v5.0.0-beta.0` to `v5.0.0-beta.1`
 
-6. Create and check out a new branch (`release-[version-number]`). See the [versioning documentation](/docs/contributing/versioning.md) for more information.
-
-7. If you're publishing a beta pre-release, update the [`CHANGELOG.md`](/CHANGELOG.md) by:
-
-   - changing the 'Unreleased' heading to the new version number and release type. For example, '5.0.0-beta.0 (Pre-release)'
-   - adding a new 'Unreleased' heading above the new version number and release type, so users will know where to add PRs to the changelog
-   - saving your changes
-
-8. Apply the new pre-release version number by running:
+6. Apply the new pre-release version number by running:
 
    ```shell
    npm version <PRE-RELEASE TYPE> --preid <PRE-RELEASE IDENTIFIER> --no-git-tag-version --workspace govuk-frontend
@@ -63,6 +57,18 @@ Developers should pair on pre-releases. When remote working, it can be useful to
    This step will update [`govuk-frontend`'s `package.json`](/packages/govuk-frontend/package.json) and project [`package-lock.json`](/package-lock.json) files.
 
    Do not commit the changes.
+
+7. Create and check out a new branch (`release-[version-number]`)
+
+   ```shell
+   git switch -c "release-$(npm run version --silent --workspace govuk-frontend)"
+   ```
+
+8. If you're publishing a beta pre-release, update the [`CHANGELOG.md`](/CHANGELOG.md) by:
+
+   - changing the 'Unreleased' heading to the new version number and release type. For example, '5.0.0-beta.0 (Pre-release)'
+   - adding a new 'Unreleased' heading above the new version number and release type, so users will know where to add PRs to the changelog
+   - saving your changes
 
 9. Run `npm run build-release` to:
 
--- a/docs/releasing/publishing-a-pre-release.md
+++ b/docs/releasing/publishing-a-pre-release.md
@@ -69,14 +69,14 @@ Developers should pair on pre-releases. When remote working, it can be useful to
 
     You will now be prompted to continue or cancel.
 
-10. If you're publishing a beta pre-release, create a pull request
+10. Create a pull request.
     When reviewing the PR, check that the version numbers have been updated and that the compiled assets use this version number.
-
-11. If you're publishing a beta pre-release, once a reviewer approves the pull request, merge it to **main**.
+    
+11. Once a reviewer approves the pull request, merge it to **main**.
 
 ## Publish a release to npm
 
-1. If you're publishing a beta pre-release, check out the **main** branch and pull the latest changes.
+1. Check out the **main** branch and pull the latest changes.
 
 2. Sign in to npm (`npm login`), using the credentials for the govuk-patterns-and-tools npm user from Bitwarden.
 
--- a/docs/releasing/publishing-a-pre-release.md
+++ b/docs/releasing/publishing-a-pre-release.md
@@ -1,15 +1,27 @@
-# Before you publish GOV.UK Frontend
+# Before you publish a pre-release of GOV.UK Frontend
 
-Read the docs for [what to do before publishing a release](/docs/releasing/before-publishing-a-release.md) to ensure you are prepared to publish.
+Decide the kind of pre-release you're publishing. All pre-releases are public, but we distinguish:
+
+- `internal` pre-releases for internal use
+- `beta` pre-releases aimed at external users
+
+We'll use the kind of pre-release as [the pre-release identifier](https://docs.npmjs.com/cli/v8/commands/npm-version#preid) in the package version.
+
+## If you're publishing a beta pre-release
+
+Besides publishing the code itself, beta releases will likely involve
+documentation updates linked to the code, as well as communications, similarly to an actual release.
+
+Review the docs for [what to do before publishing a release](/docs/releasing/before-publishing-a-release.md) to assess which steps you need to follow for your specific pre-release and ensure you are prepared to publish.
 
 See the [documentation on support branches](https://govuk-design-system-team-docs.netlify.app/how-we-work/version-control/support-branches.html#support-branches) if you need to:
 
-- publish a new release of previous major versions of GOV.UK Frontend
+- publish a new pre-release of previous major versions of GOV.UK Frontend
 - publish a ‘hotfix’ release of GOV.UK Frontend without including other unreleased changes on the `main` branch
 
 # Publish a new version of GOV.UK Frontend
 
-Developers should pair on releases. When remote working, it can be useful to be on a call together.
+Developers should pair on pre-releases. When remote working, it can be useful to be on a call together.
 
 1. Check out the **main** branch and pull the latest changes.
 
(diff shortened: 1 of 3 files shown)


36degrees:
Thanks for picking this up!

These existing prerelease docs are focused on releasing 'to GitHub' – in step 6 when we run `npm run pre-release` it creates a 'filtered' orphaned branch with the contents of the package we would publish to npm.

Is there value in prereleases to GitHub having a pre-release version number?

I'm wondering if we're conflating two things that should be kept separate? Maybe we need different terms for 'pre-releases pushed to GitHub' (which I think are still useful for e.g. exploratory testing) and 'pre-releases to npm' which are more like release candidates?

colinrotherham:
>Is there value in prereleases to GitHub having a pre-release version number?

Thought we should put @querkmachine's versioning work into action

When we run `npm run pre-release` it picks up the new `"version"` from **package.json**

It'll be confusing if our first pre-release identifies the version as `4.6.0` won't it?

### CSS version
```js
getComputedStyle(document.body)
  .getPropertyValue('--govuk-frontend-version')
```

### JS version
```js
await import('/javascripts/govuk-frontend.min.js')
  .then(({ version }) => console.log(version))
```

So instead of seeing `4.6.0` we'll see `5.0.1-rc.0` 🙌

>I'm wondering if we're conflating two things that should be kept separate? Maybe we need different terms for 'pre-releases pushed to GitHub' (which I think are still useful for e.g. exploratory testing) and 'pre-releases to npm' which are more like release candidates?

Is it the `-rc` suffix from `--preid rc` that worries you? We can change it

Have a look down the [`typescript` versions](https://www.npmjs.com/package/typescript?activeTab=versions) "Tag" column for other examples:

```
@latest
@next
@dev
@beta
@rc
@insiders
```

Not for this PR, but publishing to npm is just one script tweak away

36degrees:
I think it makes sense to use incremental version numbers (`rc.0`, `rc.1`) when we're working towards a release and doing pre-releases from the main branch.

But, we also sometimes create GitHub-based pre-releases from feature branches just to test something. Effectively, these are 'throw-away' releases just to spike something, try something out in the Design System repo or Prototype Kit etc.

I can see how it'd be useful to have a different version identifier in those instances, but I don't think it should be incremental or tied to any specific version for a couple of reasons. 

1. When I'm testing something out on a feature branch, I may not yet know how it's going to be released. Following this process, I have to make a decision about something that I don't care about right now.
2. Because we 'throw it away', the incremented version number never hits the main branch. So if we create multiple releases from feature branches, they will **all** be called `v5.0.0-rc.0` despite being distinct. The eventual npm pre-release will also be called `v5.0.0-rc.0` but will be different yet again.

36degrees:
For 'actual' pre-releases to npm, I have no issue using the `rc.x` suffix.

This is just about pre-releases from feature branches. Using the branch name for the version number might be a good option?

Given it seems like we're getting in a muddle I'm starting to think we do need distinct terms and documentation…

colinrotherham:
Yeah that works, I've updated and pushed the docs

We need the branch name as a suffix like `5.0.0-ollie` if that's alright?

I've added incremental pre-releases following the [**Update a pre-release** steps](https://github.com/alphagov/govuk-frontend/blob/main/docs/releasing/publishing-a-pre-release.md#update-a-pre-release) as an alternative. Would certainly align with "working in the open" if every GitHub release was published for easy `npm install` too

Let's ask @domoscargin for feedback as well

colinrotherham:
For our v5 pre-releases we’re probably going to have at least three? No doubt more

* `v5.0.0-rc.0` — Initial pre-release, this week?
* `v5.0.0-rc.1` — Incremental pre-release, removal of init() methods
* `v5.0.0-rc.2` — Incremental pre-release, API changes to throw errors

But for Design System previews we'll likely do the same for [new typography scale](https://github.com/alphagov/govuk-design-system/issues/2289) pre-releases

Can we think of a naming convention that works with `npm version` for both of these?

romaricpascal:
Following this morning's dev catch-up and further slack discussions, we're going to clearly separate the concept of "pre-release" (to npm) from "preview branches" (on a branch on Github, which we currently call pre-release 😱 ).

We need to do a little more research on how to handle npm pre-releases (especially on how to handle the version bump, suffix or suffixes to use, changelog, docs...).

We can already start updating our documentation for releasing to a preview branch, though.

36degrees:
I think this needs updating so that any changes to accommodate pre-releases are made to our existing release documentation instead? Is that right?

docs/releasing/publishing-a-pre-release.md
```diff
@@ -0,0 +1,125 @@
...
+
+10. Run `npm run build-release` to:
+
+    - build GOV.UK Frontend into [the package's `/dist`](/packages/govuk-frontend/dist) and [root `/dist`](/dist) directories
+    - commit the changes
+    - push a branch to GitHub
+
+    You will now be prompted to continue or cancel.
+
+11. If you're publishing a beta pre-release, create a pull request
```
romaricpascal, line 80: **note** Figured we don't really need to reverse `internal` pre-releases to `main`, only `beta` ones. Reason for that is that if we alternate `beta` and `internal` pre-releases (for example, to check a bugfix before the next pre-release), that will make the version bump rocky for `beta` release as we'd go from `-beta.X` to `-internal.0` then the next `prerelease` bump would go to `-beta.0` instead of `-beta.X+1`.

docs/releasing/publishing-a-pre-release.md
```diff
@@ -0,0 +1,125 @@
...
+
+12. If you're publishing a beta pre-release, once a reviewer approves the pull request, merge it to **main**.
+
+## Publish a release to npm
+
+1. If you're publishing a beta pre-release, check out the **main** branch and pull the latest changes.
+
+2. Sign in to npm (`npm login`), using the credentials for the govuk-patterns-and-tools npm user from Bitwarden.
+
+3. Run `npm run publish-release`, which will prompt you to check whether the npm tag looks as expected.
```
romaricpascal, line 91: **note** The script generates a default tag based on the package version thanks to `bin/generate-npm-tag.sh`. Pre-release version numbers don't break the script (yay!), however, the script could likely use some upgrading later on to create tags for pre-releases automatically.

docs/releasing/publishing-a-pre-release.md
```diff
@@ -0,0 +1,125 @@
...
+
+4. Enter `N` to continue to set the npm tag corresponding to the kind of release you're publishing:
+
+   - `internal` for internal pre-releases
+   - `next` for beta pre-releases
```
romaricpascal, line 96: **suggestion** Followed a convention that seemed followed by a good few packages publishing their pre-releases under the `@next` tag. Picking `next` instead of `beta` would allow to extract the tag directly from the version number 🤔

docs/releasing/publishing-a-pre-release.md
```diff
@@ -0,0 +1,125 @@
...
+5. You will now be prompted to continue or cancel the release. Check the details and enter `y` to continue. If something does not look right, press `N` to cancel the release.
+
+   This step will create a ZIP file containing the release in the root of your govuk-frontend git directory. You will need this file when creating the GitHub release.
+
+   It will also automatically create a tag in Github which you can use to create a Github release in the following section.
+
+6. Verify the presence of the pre-release and its tag on [npm](https://www.npmjs.com/package/govuk-frontend?activeTab=versions)
+
+   If the pre-release has been assigned the wrong tag (mistakes happen),
+   you can use [`npm dist-tag`](https://docs.npmjs.com/cli/v8/commands/npm-dist-tag) to quickly correct.
```
romaricpascal, line 107: **note** Figured a little final check wouldn't hurt. Also, as suggested by Ollie, dug how we can correct any mistakes that may have happened in tagging the package, while still logged in to npm. Also updated `publishing.md`.

romaricpascal (commented):
Pushed an update to makes the pre-release process follow the release one, rather than the preview process, as we're aiming to publish to npm.

It carries over the explanations for picking the right version bump and distinguishes steps that don't need to be taken if simply publishing an internal pre-release.

Ran the bits pushing to a branch locally and things worked fine. Also checked that `bin/generate-npm-tag.sh` didn't choke on pre-release version numbers, which it doesn't (we'll likely want to update it at some point so it picks the default tag from the version number for pre-releases).

docs/releasing/publishing-a-pre-release.md
```diff
@@ -0,0 +1,125 @@
...
+
+   - Use `premajor` to bump from `v4.7.0` to `v5.0.0-beta.0`
+   - Use `preminor` to bump from `v4.7.0` to `v4.8.0-beta.0`
+   - Use `prepatch` to bump from `v4.7.0` to `v4.7.1-beta.0`
+
+   Alternatively, when publishing an update to an existing pre-release:
+
+   - Use `prerelease` to bump from `v5.0.0-beta.0` to `v5.0.0-beta.1`
+
+6. Create and check out a new branch (`release-[version-number]`). See the [versioning documentation](/docs/contributing/versioning.md) for more information.
```
colinrotherham, line 46: The "version number" might be the same but we could be going from `.0` to `.1`

Do we need to widen `release-[version-number]`?

docs/releasing/publishing-a-pre-release.md
```diff
@@ -0,0 +1,125 @@
...
+3. Run `npm ci` to make sure you have the exact dependencies installed.
+
+4. Determine the pre-release identifier
+
+   - Use `internal` for internal pre-releases
+   - Use `beta` for beta pre-releases
+
+5. Determine the pre-release version type
+
+   Given the `beta` pre-release identifier:
```
colinrotherham, line 36: This might indicate that the following options are only for `beta` pre-releases

With enough "for beta, do this" or "for internal, do this" variables, should we split the two paths out more?

romaricpascal:
Updated with the following changes, based on our last pre-release runs:
- Removing a few "if you're publishing a beta" statements no longer necessary as we merge both internal and beta on main
- Generalise how we check we're running the right node version, as not everyone runs nvm
- Move the update of the version to before creating the branch. We don't commit until we run the build-package script, so we may as well let npm compute the new version number. It also allows to get a copy-pastable command for creating the branch, as we can read the new version number
- Added a statement to clarify that `prerelease` would also move the version from `-internal.X` to `-beta.0` and a note that it'd reset the last number to `.0`
- Added a note to log out from npm in the browser if the command line sent you there for logging in.

romaricpascal:
@colinrotherham I've updated based on what you spotted during this morning's pre-release, and tried to address your two comments 😊

docs/releasing/publishing-a-pre-release.md
```diff
@@ -0,0 +1,135 @@
+# Before you publish a pre-release of GOV.UK Frontend
+
+Decide the kind of pre-release you're publishing. All pre-releases are public, but we distinguish:
```
domoscargin, line 3:
```suggestion
Decide the kind of pre-release you're publishing. All pre-releases are public, but we distinguish between:
```

docs/releasing/publishing-a-pre-release.md
```diff
@@ -0,0 +1,135 @@
...
+
+- `internal` pre-releases for internal use
+- `beta` pre-releases aimed at external users
+
+We'll use the kind of pre-release as [the pre-release identifier](https://docs.npmjs.com/cli/v8/commands/npm-version#preid) in the package version.
+
+## If you're publishing a beta pre-release
+
+Besides publishing the code itself, beta releases will likely involve
```
domoscargin, line 12: Can we pin this down so we can say something like:

```md
Much like an actual release, a beta release involves:

- publishing the code itself
- documentation updates linked to the code
- user communication
```

If we're not sure about pinning it down, the `similarly to an actual release` is a bit awkward. Maybe something like:

```
Besides publishing the code itself, beta releases will likely involve code-related documentation updates and user communications, much like an actual release.

File: docs/releasing/publishing-a-pre-release.md [hunk 9]
```diff
@@ -0,0 +1,135 @@
...
+documentation updates linked to the code, as well as communications, similarly to an actual release.
+
+Review the docs for [what to do before publishing a release](/docs/releasing/before-publishing-a-release.md) to assess which steps you need to follow for your specific pre-release and ensure you are prepared to publish.
```
domoscargin, line 15: Again, it'd be great if we could just be sure that for beta releases, we follow a normal release process rather than leaving it for individuals to assess. But no worries if that's not possible right now.

docs/releasing/publishing-a-pre-release.md
```diff
@@ -0,0 +1,135 @@
...
+
+See the [documentation on support branches](https://govuk-design-system-team-docs.netlify.app/how-we-work/version-control/support-branches.html#support-branches) if you need to publish a new pre-release of previous major versions of GOV.UK Frontend.
+
+# Publish a new version of GOV.UK Frontend
+
+Developers should pair on pre-releases. When remote working, it can be useful to be on a call together.
```
domoscargin, line 21:
```suggestion
Developers should pair on pre-releases. When working remotely, it can be useful to be on a call together.
```

docs/releasing/publishing-a-pre-release.md
```diff
@@ -0,0 +1,135 @@
...
+1. Check out the **main** branch and pull the latest changes.
+
+2. Ensure you're running the version of NodeJS matching [`.nvmrc`](/.nvmrc).
+
+   - If you use NVM, run `nvm use` to set up the right version
+   - If you use another management system (like [`asdf`](https://asdf-vm.com/guide/getting-started.html)), compare the output of `node --version` and install the right one if necessary
+
+3. Run `npm ci` to make sure you have the exact dependencies installed.
+
+4. Determine the pre-release identifier
```
domoscargin, line 32: Can we somehow flag this and link it better to step 6's code?

Would it be insane to make this `Determine the <PRE-RELEASE IDENTIFIER>`, so it's easy to connect it to step 6?

docs/releasing/publishing-a-pre-release.md
```diff
@@ -0,0 +1,135 @@
...
+
+   - Use `internal` for internal pre-releases
+   - Use `beta` for beta pre-releases
+
+5. Determine the pre-release version type
```
domoscargin, line 37: Similar to my comment above, is there a way to flag this more? There's a lot of commands, version numbers, identifiers in this section.

docs/releasing/publishing-a-pre-release.md
```diff
@@ -0,0 +1,135 @@
...
+   - adding a new 'Unreleased' heading above the new version number and release type, so users will know where to add PRs to the changelog
+   - saving your changes
+
+9. Run `npm run build-release` to:
+
+   - build GOV.UK Frontend into [the package's `/dist`](/packages/govuk-frontend/dist) and [root `/dist`](/dist) directories
+   - commit the changes
+   - push a branch to GitHub
+
+   You will now be prompted to continue or cancel.
```
domoscargin, line 81: It might seem pedantic, but should we give the user an action here? Similar to step 5 of "Publish a release to npm"

docs/releasing/publishing-a-pre-release.md
```diff
@@ -0,0 +1,135 @@
...
+4. Enter `N` to continue to set the npm tag corresponding to the kind of release you're publishing:
+
+   - `internal` for internal pre-releases
+   - `next` for beta pre-releases
+
+5. You will now be prompted to continue or cancel the release. Check the details and enter `y` to continue. If something does not look right, press `N` to cancel the release.
+
+   This step will create a ZIP file containing the release in the root of your govuk-frontend git directory. You will need this file when creating the GitHub release.
+
+   It will also automatically create a tag in GitHub which you can use to create a GitHub release in the following section.
```
domoscargin, line 112: Pedantic: it's not _super_ clear what "it" is referring to here, since we're talking about a file just before, instead of "this step".

docs/releasing/publishing.md
```diff
@@ -13,31 +13,39 @@ Developers should pair on releases. When remote working, it can be useful to be
...
-2. Run `nvm use` to make sure you're using the right version of Node.js and npm.
+2. Ensure you're running the version of NodeJS matching [`.nvmrc`](/.nvmrc).
+
+   - If you use NVM, run `nvm use` to set up the right version
+   - If you use another management system (like [`asdf`](https://asdf-vm.com/guide/getting-started.html)), compare the output of `node --version` and install the right one if necessary
 
 3. Run `npm ci` to make sure you have the exact dependencies installed.
 
-4. Create and check out a new branch (`release-[version-number]`). See the [versioning documentation](/docs/contributing/versioning.md) for more information.
+4. Pick a new version number according to the [versioning documentation](/docs/contributing/versioning.md) and apply it running:
```
domoscargin, line 23:
```suggestion
4. Pick a new version number according to the [versioning documentation](/docs/contributing/versioning.md) and apply it by running:
```

docs/releasing/publishing-a-pre-release.md
romaricpascal: @domoscargin I think that's something we'll only be able to pin down after a couple of betas. We can clarify once we've got a few under our belt, for different kinds of releases and see what's common and what's flexible from one to another.

docs/releasing/publishing-a-pre-release.md
romaricpascal: Good shout, I'll add the placeholder to the heading, and in step 5 as well.

docs/releasing/publishing-a-pre-release.md
romaricpascal: I'm not sure we can here. In the npm part, step 4, we can guide the users towards pressing `N` as we know they'll have to change from 'latest' to 'next' or 'internal'. This is more like the step 5 of npm where we need the person driving to have a pause and consider whether to continue, rather than automatically go "'y', let's go!"

docs/releasing/publishing-a-pre-release.md
romaricpascal: Tracked by https://github.com/alphagov/govuk-frontend/issues/4396

romaricpascal:
@domoscargin I've addressed your suggestions and replied when I didn't make an update. Ready for another review 😊

domoscargin (approved):
@romaricpascal Looking good. I've left one comment where I think we can still be clearer, but I wouldn't consider it blocking - we can always tweak this as we do more pre-releases.

docs/releasing/publishing-a-pre-release.md
romaricpascal: Oh, I see, yeah, I can add the same clarification here, good shout 😊

PR #3978, opened 2023-07-20
Add support for multiple ids in radio aria-controls

sobakasu:


querkmachine:
Good morning @sobakasu, thanks for submitting a PR. 

I'm curious as to what the intended use case for this is. Are you allowing multiple radio buttons to conditionally show and hide a single element further down the page? 

If so, I don't think this is a use case we intend to support in Frontend. 

Support for conditionally revealed content by assistive technologies (especially on radio buttons) is inconsistent, and part of our affordance for that is to have the conditionally revealed content always appear directly after the toggled input, so that it's always the next thing a user will find on the page. In that scenario there is always a 1-to-1 relationship between an input and the content it toggles. 

That way, even if the user is not notified by their assistive tech that checking the radio revealed anything, they will immediately encounter the follow-up question. This is also why we advise to [only conditionally reveal questions](https://design-system.service.gov.uk/components/radios/#conditionally-revealing-a-related-question), rather than plain text content, as plain text may be skipped by ATs in form contexts. 

If that isn't (or is!) the use case you're trying to cover, more information and perspectives are always welcome!

sobakasu:
> Good morning @sobakasu, thanks for submitting a PR.
> 
> I'm curious as to what the intended use case for this is. Are you allowing multiple radio buttons to conditionally show and hide a single element further down the page?
> 
> If so, I don't think this is a use case we intend to support in Frontend.
> 
> Support for conditionally revealed content by assistive technologies (especially on radio buttons) is inconsistent, and part of our affordance for that is to have the conditionally revealed content always appear directly after the toggled input, so that it's always the next thing a user will find on the page. In that scenario there is always a 1-to-1 relationship between an input and the content it toggles.
> 
> That way, even if the user is not notified by their assistive tech that checking the radio revealed anything, they will immediately encounter the follow-up question. This is also why we advise to [only conditionally reveal questions](https://design-system.service.gov.uk/components/radios/#conditionally-revealing-a-related-question), rather than plain text content, as plain text may be skipped by ATs in form contexts.
> 
> If that isn't (or is!) the use case you're trying to cover, more information and perspectives are always welcome!

Our use case is to conditionally reveal content further down the page (in addition to hint text that's revealed under the radio button) which are separated in the dom. Your explanation about why frontend doesn't support this makes sense to me. I think we can just close this PR

PR #3988, opened 2023-07-20
Update review app iframes to resize after `load`

colinrotherham:
Should we take any of [`iframe-resizer` changes](https://github.com/alphagov/govuk-design-system/pull/2941) into the review app @domoscargin?


packages/govuk-frontend-review/src/views/layouts/layout.njk
```diff
@@ -14,7 +14,14 @@
     <script src="/vendor/iframe-resizer/iframeResizer.min.js" defer></script>
```
domoscargin, line 14: Hmm, maybe we should defer this on the Design System site as well. We don't currently.

packages/govuk-frontend-review/src/views/layouts/layout.njk
```diff
@@ -14,7 +14,14 @@
...
     <script>
       document.addEventListener('DOMContentLoaded', function () {
-        window.iFrameResize({}, '.js-component-preview')
+        var $examples = document.querySelectorAll('.js-component-preview')
```
domoscargin, line 17: Since this is an internal development app, are we able to start using modern syntax (`const`, etc)? I'm not sure how the review app fits into our plans! Non-blocking, since I guess the default assumption is we don't change things till we do so elsewhere.

packages/govuk-frontend-review/src/views/layouts/layout.njk
```diff
@@ -14,7 +14,14 @@
...
+
+        // Resize component preview iframes when loaded
+        for (var i = 0; i < $examples.length; i++) {
+          $examples[i].addEventListener('load', function () {
```
domoscargin, line 21: There's a lot more layout shift on the review app than the DS website - we've set some height classes there so that the unloaded examples take up roughly the right amount of space. Probably not a biggie for an internal app, and it looks like it'd be trickier to target above-fold iframes and make 'em eager here than on the DS site, so not a blocker for this PR.

domoscargin (approved):
A few non-blocking comments, but otherwise looking good!

packages/govuk-frontend-review/src/views/layouts/layout.njk
colinrotherham: I was thinking it'd be nice to see iframes resize in IE 11 and other older browsers

But I think you've convinced me, plus we always have "open in a new window" as backup

packages/govuk-frontend-review/src/views/layouts/layout.njk
colinrotherham: Yeah it's a bit jarring isn't it

Can we keep the first 3x components with `loading="eager"` to keep the old behaviour? Pushed 👍

We have a lot of internal margin/padding inside our iframes, but works for smaller components like "Button"

Originally [previews just had `margin: 15px 0;`](https://github.com/alphagov/govuk-frontend/blob/80e1b90caed2f91f1c2f34540ee606409924483b/packages/govuk-frontend-review/src/stylesheets/partials/_app.scss#L7) but then we added `.govuk-main-wrapper--auto-spacing` and the "**Code**" `<details>` got pushed much further down and the shift is quite prominent

packages/govuk-frontend-review/src/views/layouts/layout.njk
colinrotherham: Ah we added this when ES modules were introduced, since they have `defer` behaviour by default

Otherwise we saw regular `<script>` run ahead of GOV.UK Frontend `<script type="module">` code

I've pushed up an ES modules version instead 👍

domoscargin (approved):
Looks good!

PR #3989, opened 2023-07-20
Update default Exit This Page button text

### LINKED ISSUE #3688: The non-visual experience of Exit this Page button doesn't make it clear it's a safety feature

querkmachine:
Updates the default Exit This Page button text to include visually hidden text, so the button reads 'Emergency Exit this Page' to screen reader users, to clarify the button's intended use as a safety feature. 

Split out from #3878. Intended to resolve #3688. 

I assume this would need a 'Suggested change' in the changelog now that EtP has shipped in 4.7.0?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -287,6 +287,21 @@ You don't need to change anything if you're using the `govukPagination` Nunjucks
 
 This change was introduced in [pull request #3899: Update default `aria-label` in Pagination component](https://github.com/alphagov/govuk-frontend/pull/3899).
 
+#### Update the Exit this Page button's default text
+
+The default text of the Exit this Page button has been updated to indicate that the button is a safety tool and not a generic method of leaving the current page. If you are using the component's default label, you may wish to update it to the new value.
+
+```diff
+<a href="..." role="button" draggable="false" class="govuk-button govuk-button--warning govuk-exit-this-page__button govuk-js-exit-this-page-button" data-module="govuk-button">
+-   Exit this page
++   <span class="govuk-visually-hidden">Emergency</span> Exit this page
+</a>
+```
+
+You don't need to change anything if you're using the `govukExitThisPage` Nunjucks macro.
+
+This change was introduced in [pull request #3989: Update default Exit This Page button text](https://github.com/alphagov/govuk-frontend/pull/3989).
+
 ### Fixes
 
 We’ve made fixes to GOV.UK Frontend in the following pull requests:
--- a/packages/govuk-frontend/src/govuk/components/exit-this-page/exit-this-page.yaml
+++ b/packages/govuk-frontend/src/govuk/components/exit-this-page/exit-this-page.yaml
@@ -2,11 +2,11 @@ params:
   - name: text
     type: string
     required: false
-    description: Text for the link. If `html` is provided, the `text` option will be ignored. Defaults to 'Exit this page'.
+    description: Text for the link. If `html` is provided, the `text` option will be ignored. Defaults to 'Emergency Exit this page' with 'Emergency' visually hidden.
   - name: html
     type: string
     required: false
-    description: HTML for the link. If `html` is provided, the `text` option will be ignored.
+    description: HTML for the link. If `html` is provided, the `text` option will be ignored. Defaults to 'Emergency Exit this page' with 'Emergency' visually hidden.
   - name: redirectUrl
     type: string
     required: false
--- a/packages/govuk-frontend/src/govuk/components/exit-this-page/template.njk
+++ b/packages/govuk-frontend/src/govuk/components/exit-this-page/template.njk
@@ -1,5 +1,9 @@
 {% from "../button/macro.njk" import govukButton -%}
 
+{% set defaultHtml -%}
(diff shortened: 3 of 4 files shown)


packages/govuk-frontend/src/govuk/components/exit-this-page/template.njk
```diff
@@ -1,5 +1,10 @@
 {% from "../button/macro.njk" import govukButton -%}
 
+{% set hasSetCustomText = params.html !== undefined or params.text !== undefined -%}
```
colinrotherham, line 3: There's a gotcha in this as `{ html: null }` will pass the checks

Probably too far to filter empty space with ` | trim` but will this do?

```suggestion
{% set hasSetCustomText = params.html or params.text -%}
```

packages/govuk-frontend/src/govuk/components/exit-this-page/exit-this-page.yaml
```diff
@@ -2,11 +2,11 @@ params:
   - name: text
     type: string
     required: false
-    description: Text for the link. If `html` is provided, the `text` option will be ignored. Defaults to 'Exit this page'.
+    description: Text for the link. If `html` is provided, the `text` option will be ignored. Defaults to 'Emergency Exit this page' with 'Emergency' visually hidden.
```
colinrotherham, line 5: Should it be `"Emergency: Exit this page"`?

I can't help read "Emergency Exit…" as this sign

Emergency exit sign

packages/govuk-frontend/src/govuk/components/exit-this-page/template.njk
```diff
@@ -8,8 +13,8 @@
   {%- if params.pressOneMoreTimeText %} data-i18n.press-one-more-time="{{ params.pressOneMoreTimeText | escape }}"{% endif %}
 >
   {{- govukButton({
-    html: params.html,
-    text: params.text | default("Exit this page"),
+    html: params.html if hasSetCustomText else defaultHtml,
```
colinrotherham, line 16: Or keep it short and sweet here? Parentheses optional

```suggestion
    html: params.html if (params.html or params.text) else defaultHtml,
```

packages/govuk-frontend/src/govuk/components/exit-this-page/template.njk
querkmachine: I may be misremembering (I originally wrote this about a month ago!) but I seem to recall it not working as expected without the `!== undefined`. I can give it a re-test just to be sure.

querkmachine: Seems like it works fine without it, actually. I might be remembering a previous iteration, then.

CHANGELOG.md
```diff
@@ -287,6 +287,21 @@ You don't need to change anything if you're using the `govukPagination` Nunjucks
 
 This change was introduced in [pull request #3899: Update default `aria-label` in Pagination component](https://github.com/alphagov/govuk-frontend/pull/3899).
 
+#### Update the Exit this Page button's default text
+
+The default text of the Exit this Page button has been updated to indicate that the button is a safety tool and not a generic method of leaving the current page. If you are using the component's default label, you may wish to update it to the new value.
+
+```diff
```
colinrotherham, line 294: 😮

CHANGELOG.md
```diff
@@ -287,6 +287,21 @@ You don't need to change anything if you're using the `govukPagination` Nunjucks
...
+<a href="..." role="button" draggable="false" class="govuk-button govuk-button--warning govuk-exit-this-page__button govuk-js-exit-this-page-button" data-module="govuk-button">
+-   Exit this page
```
colinrotherham, line 296: ❤️

CHANGELOG.md
```diff
@@ -287,6 +287,21 @@ You don't need to change anything if you're using the `govukPagination` Nunjucks
...
++   <span class="govuk-visually-hidden">Emergency</span> Exit this page
```
colinrotherham, line 297: 🙌

PR #4002, opened 2023-07-24
Have Dependabot ignore Stylelint major version updates

romaricpascal:
stylelint-config-gds is preventing the update to Stylelint 15 for now. Until a new version is released, we'll ignore the next major version bumps.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -79,6 +79,12 @@ updates:
     allow:
       - dependency-type: direct
 
+    ignore:
+      # Ignore major updates (causes `stylelint-config-gds` peer dependency conflict)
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -79,6 +79,11 @@ updates:
     allow:
       - dependency-type: direct
 
+    ignore:
+      # We need to wait for stylelint-config-gds to be ready for Stylelint 15
```
colinrotherham, line 83: Shall we document the PR so anyone knows when to review this `ignore` rule?

```suggestion
      # Ignore major updates (causes `stylelint-config-gds` peer dependency conflict)
      # See pull request: https://github.com/alphagov/stylelint-config-gds/pull/32
```

colinrotherham (approved):

PR #4006, opened 2023-07-24
[support/4.x] Backport: GitHub Actions + Jest fixes for Windows

colinrotherham:
Partially backports https://github.com/alphagov/govuk-frontend/pull/3816 and https://github.com/alphagov/govuk-frontend/pull/3863 for **v4.x** for Windows test reliability

**Note:** Changes related to path handling https://github.com/alphagov/govuk-frontend/pull/3756 (v5 only) were excluded

### DIFF
--- a/.github/workflows/actions/build/action.yml
+++ b/.github/workflows/actions/build/action.yml
@@ -9,8 +9,10 @@ runs:
       id: build-cache
 
       with:
+        enableCrossOsArchive: true
+
         # Restore build cache (unless commit SHA changes)
-        key: build-cache-${{ runner.os }}-${{ github.sha }}
+        key: build-${{ runner.os }}-${{ github.sha }}
         path: app/dist
 
     - name: Build
--- a/.github/workflows/actions/install-node/action.yml
+++ b/.github/workflows/actions/install-node/action.yml
@@ -9,8 +9,10 @@ runs:
       id: npm-install-cache
 
       with:
+        enableCrossOsArchive: true
+
         # Restore `node_modules` cache (unless packages change)
-        key: npm-install-cache-${{ runner.os }}-${{ hashFiles('package-lock.json', '**/package.json') }}
+        key: npm-install-${{ runner.os }}-${{ hashFiles('package-lock.json', '**/package.json') }}
         path: |
           node_modules
           app/node_modules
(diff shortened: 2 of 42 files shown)


.github/workflows/actions/setup-node/action.yml
```diff
@@ -3,16 +3,17 @@ name: Setup
...
 
   steps:
     - name: Setup Node.js
-      uses: actions/setup-node@v3.6.0
+      uses: actions/setup-node@v3.7.0
       id: setup-node
 
       with:
-        cache: ${{ inputs.use-cache != 'false' && 'npm' || '' }}
+        cache: ${{ inputs.use-cache == 'true' && 'npm' || '' }}
```
colinrotherham, line 18: This was [the fix](https://github.com/alphagov/govuk-frontend/pull/3954/commits/b8165f9c53bfb8303cdc506df2e424ac03d89980) that unblocked https://github.com/alphagov/govuk-frontend/pull/3954/commits/65965b2968e678cdadbe3ca8f8f486c815b4c560 but speeds up Windows cache restoration times

domoscargin:
Equally, pinning leaves us vulnerable to uncaught security warnings when `main` gets more and more out of sync with the `v4` branch, so could end up releasing something dangerous.

I think it probably makes sense to configure security bumps for the branch? Security fixes are pretty much what we're committing to anyway.

colinrotherham:
Great minds! I had the exact same thoughts but kept it solely for stability

For example, when releasing v4.8.0 we know what worked 3 months ago still works now, avoiding any surprise GitHub Action updates that slipped in a breaking change whilst we were busy on v5

Could look into [Dependabot `target-branch`](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#target-branch) and update `support/4.x` too?

36degrees:
> I think it probably makes sense to pin everything, then configure security bumps for the branch? Security fixes are pretty much what we're committing to anyway.

Unfortunately, as far as I can tell, [Dependabot doesn't support security updates against multiple branches](https://github.com/orgs/community/discussions/15027) 😞

colinrotherham:
So we can discuss further + unblock, I've split out the commits that pin GitHub Actions versions:

* https://github.com/alphagov/govuk-frontend/pull/4014

.github/workflows/actions/setup-node/action.yml
colinrotherham: Moved to https://github.com/alphagov/govuk-frontend/pull/4014

romaricpascal (approved):
Backport looks fine, let's merge it ⛵

PR #4011, opened 2023-07-26
Remove component init() methods and initialise in constructor

### LINKED ISSUE #4008: Remove `init()` method from our JavaScript components
## What

Remove the `init()` method from our JavaScript components and move their initialisation to the bottom `constructor()`.

## Why

The `init()` method adds an unnecessary step for our components' initialisation,
without tangible benefits for what our components do. Without it we can simplify both how our components are initialised, as well as their internal implementation as we're looking to [throw errors at instantiation](./009-throw-errors-component-constructors.md).

Removing it entirely and moving its code to the `constructor` will ensure that components don't get a double initialisation by services that may have missed a `.init()` call during their migration. Instead, they'll get a clear error telling them that they're calling a method that no longer exists.

You can find more explanations of that decision in [our decision record](https://github.com/alphagov/govuk-design-system-architecture/pull/33/files).

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [ ] All components have had their `init` method removed and initialisation moved to the `constructor()`

domoscargin:
Closes #4008 

## What

Remove `init()` method from each component and move initialisation into constructor.

## Why
The init() method adds an unnecessary step for our components' initialisation,
without tangible benefits for what our components do. Without it we can simplify both how our components are initialised, as well as their internal implementation as we're looking to [throw errors at instantiation](https://github.com/alphagov/govuk-frontend/issues/009-throw-errors-component-constructors.md).

Removing it entirely and moving its code to the constructor will ensure that components don't get a double initialisation by services that may have missed a .init() call during their migration. Instead, they'll get a clear error telling them that they're calling a method that no longer exists.

You can find more explanations of that decision in [our decision record](https://github.com/alphagov/govuk-design-system-architecture/pull/33/files).

### eslint `no-new` rule fails
eslint flags the fact that we don't store a reference to the constructed component when we instantiate it. We've decided to ignore this temporarily until we look at adding a component registry down the line.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,22 @@ This change was made in [pull request #3819: Add linked image focus style](https
 
 You must make the following changes when you migrate to this release, or your service might break.
 
+#### Update component initialisation
+
+Remove `.init()` from individually instantiated components as initialisation now happens automatically:
+
+```mjs
+new Radios($radio).init()
+```
+
+```mjs
+new Radios($radio)
+```
+
+If you import the JavaScript using `window.GOVUKFrontend.initAll()`, you will not need to make any changes.
+
+This change was introduced in [pull request #4011: Remove component init() methods and initialise in constructor](https://github.com/alphagov/govuk-frontend/pull/4011).
+
 #### Check that details components work as expected
 
 The Details component no longer uses JavaScript, and is no longer polyfilled in older browsers.
--- a/packages/govuk-frontend/src/govuk/all.mjs
+++ b/packages/govuk-frontend/src/govuk/all.mjs
@@ -1,3 +1,5 @@
+/* eslint-disable no-new */
+
 import { version } from './common/govuk-frontend-version.mjs'
 import { Accordion } from './components/accordion/accordion.mjs'
 import { Button } from './components/button/button.mjs'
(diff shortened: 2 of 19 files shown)


CHANGELOG.md
```diff
@@ -59,6 +59,25 @@ This change was made in [pull request #3819: Add linked image focus style](https
...
+#### Update component initialisation
+
+We have removed the `init()` method from all components which include JavaScript. Initialisation now happens when the
+component is created.
+
+If you import the JavaScript using `window.GOVUKFrontend.initAll()`, you will not need to make any changes.
+
+If you initialise any component individually, you will need to remove the `.init()`. For example:
+
+```js
```
colinrotherham, line 71: Super subtle, but can we use `mjs` by default just in case we turn ESLint back on for CHANGELOG?

```suggestion
```mjs
```

packages/govuk-frontend/src/govuk/components/checkboxes/checkboxes.mjs
```diff
@@ -9,6 +9,18 @@ export class Checkboxes {
   $inputs
 
   /**
+   *
```
colinrotherham, line 12:
```suggestion
```

packages/govuk-frontend/src/govuk/components/globals.test.js
```diff
@@ -45,22 +45,6 @@ describe('GOV.UK Frontend', () => {
       ])
     })
 
-    it('exported Components have an init function', async () => {
```
colinrotherham, line 48: 🙌

packages/govuk-frontend/src/govuk/components/header/header.mjs
```diff
@@ -44,20 +46,14 @@ export class Header {
...
-   * Apply a matchMedia for desktop which will trigger a state sync if the browser
-   * viewport moves between states.
-   */
-  init () {
-    // Check that required elements are present
-    if (!this.$module || !this.$menuButton || !this.$menu) {
-      return
+    if (
+      !(
+        this.$menuButton instanceof HTMLButtonElement ||
```
colinrotherham, line 52: We tend to not be specific with these checks so `HTMLElement` is good enough usually, but if we need the compiler to guard specific browser APIs then go for it

For example:

**Character count** checks for `HTMLTextAreaElement` or `HTMLInputElement`
**Radios** and **Checkboxes** check for `HTMLInputElement`
**Skip link** checks for `HTMLAnchorElement`

```suggestion
        this.$menuButton instanceof HTMLElement ||
```

packages/govuk-frontend/src/govuk/all.mjs
```diff
@@ -33,60 +33,71 @@ function initAll (config) {
 
   const $accordions = $scope.querySelectorAll('[data-module="govuk-accordion"]')
   $accordions.forEach(($accordion) => {
-    new Accordion($accordion, config.accordion).init()
+    /* eslint-disable-next-line no-new */
```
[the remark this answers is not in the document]
domoscargin: I was going for being really explicit about what might need changing in the future, but I guess removing a top-of-file disable will flag everything in the linting anyway, so I'll do this instead.

colinrotherham (approved):
A work of wonder 🎉

PR #4013, opened 2023-07-26
[SPIKE] Share installed `node_modules` between runners

colinrotherham:
This PR checks how close we are to sharing `node_modules` between runners

Although we use the [`enableCrossOsArchive` option](https://github.com/alphagov/govuk-frontend/pull/4012) we typically hit problems with:

1. **Executable compatibility in `node_modules/.bin`**
   Restoring previously linked `node_modules/.bin` executables into Windows lacks [`--bin-links` `.cmd` shims](https://docs.npmjs.com/cli/v9/commands/npm-install?v=true#bin-links)

2. **Executable compatibility in Dart Sass**
   Restoring previously installed `node_modules` into Windows provides Linux Sass binaries

```console
Error: Embedded Dart Sass couldn't find the embedded compiler executable. Please make sure the optional dependency sass-embedded-win32-x64 is installed in node_modules.
```

### DIFF
--- a/.github/workflows/actions/install-node/action.yml
+++ b/.github/workflows/actions/install-node/action.yml
@@ -13,9 +13,10 @@ runs:
         enableCrossOsArchive: true
 
         # Restore `node_modules` cache (unless packages change)
-        key: npm-install-${{ runner.os }}-${{ hashFiles('package-lock.json', '**/package.json') }}
+        key: npm-install-${{ hashFiles('package-lock.json', '**/package.json') }}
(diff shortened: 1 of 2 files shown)


colinrotherham:
Looks like the initial `npm ci` keeps only executables for Linux/x64

```json
{
  "name": "sass-embedded-linux-x64",
  "version": "1.64.1",
  "os": [
    "linux"
  ],
  "cpu": [
    "x64"
  ]
}
```

PR #4014, opened 2023-07-27
[support/4.x] Backport: Use exact GitHub Actions versions

colinrotherham:
Backports https://github.com/alphagov/govuk-frontend/pull/3954 to pin GitHub Actions versions


romaricpascal (approved):
We agreed to pin the versions during dev catch-up yesterday, so good to merge 🥳

PR #4030, opened 2023-07-31
Throw error when instantiating components where GOV.UK Frontend is not supported

### LINKED ISSUE #4009: Throw errors if GOV.UK Frontend is not supported
## What

Throw a `GOVUKFrontendNotSupportedError` (extending a base `GOVUKFrontendError`) if any of our JavaScript component is instanciated while GOV.UK Frontend is not supported (`document.body.classList.contains('govuk-frontend-supported')`).

> **Note**: Components throwing because of this means we'll likely want an early return in `initAll` to avoid riddling the console with exceptions logs

## Why

This avoids users getting a reference to a component in an invalid state, as well as helps debugging with a clear error (rather than nothing happening).

You can find more information about this decision in [our decision record](https://github.com/alphagov/govuk-design-system-architecture/pull/32/files).

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [ ] All components throw an error if instanciated where GOV.UK Frontend is not supported

romaricpascal:
This PR updates all our JavaScript components to throw a `GOVUKFrontendNotSupportedError` if they get instantiated where GOV.UK Frontend is not supported.

Besides updating the components to throw this new error the PR:

- introduces a `GOVUKFrontendError` for our errors to subclass. It allows users to easily [separate our custom errors from regular JavaScript errors using `instanceof`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error#handling_a_specific_error_type).
- updates `renderAndInitialise` to make running code before the component's initialisation more reliable (had issues where the `initialiser` – now renamed `beforeInitialisation` for clarity – didn't execute when run in the middle of the block instantiating the component).

Closes #4009

### DIFF
--- a/docs/contributing/coding-standards/js.md
+++ b/docs/contributing/coding-standards/js.md
@@ -136,6 +136,40 @@ Avoid using namespace imports (`import * as namespace`) in code bundled for Comm
 
 Prefer named exports over default exports to avoid compatibility issues with transpiler "synthetic default" as discussed in: https://github.com/alphagov/govuk-frontend/issues/2829
 
+## Throwing errors
+
+### Error types
+
+First, check if one of the [native errors provides](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error#error_types) the right semantic for the error you're looking to throw, if so, use this class.
+
+If none of the native errors have the right semantics, create a new class for this error. This class should:
+
+- have a name ending in `Error` to match the native conventions
+- extend `GOVUKFrontendError`, so users can separate our custom errors from native ones using `instanceof`
+- have a `name` property set to its class name, as extending classes doesn't set this automatically and grabbing the constructor's name risks being affected by mangling during minification
+
+```js
+class CustomError extends GOVUKFrontendError {
+  name = 'CustomError'
+}
+```
+
+### Error message
+
+Keep the message to the point, but provide the users the information they need to identify the cause of the error.
+
+If the message is the same whatever the situation, you may use the constructor of our custom error to centralise that message, rather than passing it each time an error is thrown.
+
+```js
+class SupportError extends GOVUKFrontendError {
+  name = 'SupportError'
+
+  constructor () {
+    super('GOV.UK Frontend is not supported in this browser')
+  }
+}
+```
+
 ## Polyfilling
 
 If you need polyfills for features that are not yet included in this project, please see the following guide on [how to add polyfills](../polyfilling.md).
--- a/packages/govuk-frontend/src/govuk/common/index.mjs
+++ b/packages/govuk-frontend/src/govuk/common/index.mjs
@@ -133,3 +133,17 @@ export function extractConfigByNamespace(configObject, namespace) {
   }
   return newObject
 }
+
+/**
+ * Checks if GOV.UK Frontend is supported on this page
+ *
+ * Some browsers will load and run our JavaScript but GOV.UK Frontend
+ * won't be supported.
+ *
+ * @internal
+ * @param {HTMLElement} [$scope] - The `<body>` element of the document to check for support
+ * @returns {boolean} Whether GOV.UK Frontend is supported on this page
+ */
+export function isSupported($scope = document.body) {
+  return $scope.classList.contains('govuk-frontend-supported')
+}
--- a/packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
+++ b/packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
@@ -1,5 +1,6 @@
 import { mergeConfigs, extractConfigByNamespace } from '../../common/index.mjs'
 import { normaliseDataset } from '../../common/normalise-dataset.mjs'
+import { GOVUKFrontendSupportError } from '../../errors/index.mjs'
 import { I18n } from '../../i18n.mjs'
 
 /**
@@ -113,10 +114,11 @@ export class Accordion {
    * @param {AccordionConfig} [config] - Accordion config
    */
   constructor($module, config) {
-    if (
-      !($module instanceof HTMLElement) ||
-      !document.body.classList.contains('govuk-frontend-supported')
-    ) {
+    if (!document.body.classList.contains('govuk-frontend-supported')) {
+      throw new GOVUKFrontendSupportError()
+    }
+
+    if (!($module instanceof HTMLElement)) {
       return this
     }
 
--- a/packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
+++ b/packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
@@ -1,6 +1,6 @@
 import { mergeConfigs, extractConfigByNamespace } from '../../common/index.mjs'
 import { normaliseDataset } from '../../common/normalise-dataset.mjs'
-import { GOVUKFrontendSupportError } from '../../errors/index.mjs'
+import { GOVUKFrontendComponent } from '../../govuk-frontend-component.mjs'
 import { I18n } from '../../i18n.mjs'
 
 /**
@@ -15,7 +15,7 @@ import { I18n } from '../../i18n.mjs'
  * The state of each section is saved to the DOM via the `aria-expanded`
  * attribute, which also provides accessibility.
  */
-export class Accordion {
+export class Accordion extends GOVUKFrontendComponent {
   /** @private */
   $module
 
@@ -114,9 +114,7 @@ export class Accordion {
    * @param {AccordionConfig} [config] - Accordion config
    */
   constructor($module, config) {
-    if (!document.body.classList.contains('govuk-frontend-supported')) {
-      throw new GOVUKFrontendSupportError()
-    }
+    super()
 
     if (!($module instanceof HTMLElement)) {
       return this
--- a/packages/govuk-frontend/src/govuk/components/accordion/accordion.test.js
+++ b/packages/govuk-frontend/src/govuk/components/accordion/accordion.test.js
@@ -712,6 +712,28 @@ describe('/components/accordion', () => {
             )
           })
         })
+
+        describe('errors at instantiation', () => {
+          let examples
+
+          beforeAll(async () => {
+            examples = await getExamples('accordion')
+          })
+
+          it('throws when GOV.UK Frontend is not supported', async () => {
+            await expect(
+              renderAndInitialise(page, 'accordion', {
+                params: examples.default,
+                beforeInitialisation() {
+                  document.body.classList.remove('govuk-frontend-supported')
+                }
+              })
+            ).rejects.toEqual({
+              name: 'SupportError',
+              message: 'GOV.UK Frontend is not supported in this browser'
+            })
+          })
+        })
       })
     })
   })
--- a/packages/govuk-frontend/src/govuk/components/character-count/character-count.test.js
+++ b/packages/govuk-frontend/src/govuk/components/character-count/character-count.test.js
@@ -763,6 +763,28 @@ describe('Character count', () => {
         )
       })
     })
+
+    describe('errors at instantiation', () => {
+      let examples
+
+      beforeAll(async () => {
+        examples = await getExamples('character-count')
+      })
+
+      it('throws when GOV.UK Frontend is not supported', async () => {
+        await expect(
+          renderAndInitialise(page, 'character-count', {
+            params: examples.default,
+            beforeInitialisation() {
+              document.body.classList.remove('govuk-frontend-supported')
+            }
+          })
+        ).rejects.toEqual({
+          name: 'SupportError',
+          message: 'GOV.UK Frontend is not supported in this browser'
+        })
+      })
+    })
   })
 
   describe('in mismatched locale', () => {
--- a/packages/govuk-frontend/src/govuk/components/character-count/character-count.test.js
+++ b/packages/govuk-frontend/src/govuk/components/character-count/character-count.test.js
@@ -777,7 +777,7 @@ describe('Character count', () => {
           // Override maxlength to 10
           maxlength: 10
         },
-        initialiser($module) {
+        beforeInitialisation($module) {
           // Set locale to Welsh, which expects translations for 'one', 'two',
           // 'few' 'many' and 'other' forms – with the default English strings
           // provided we only have translations for 'one' and 'other'.
--- a/packages/govuk-frontend/src/govuk/components/error-summary/error-summary.mjs
+++ b/packages/govuk-frontend/src/govuk/components/error-summary/error-summary.mjs
@@ -1,5 +1,6 @@
 import { mergeConfigs } from '../../common/index.mjs'
 import { normaliseDataset } from '../../common/normalise-dataset.mjs'
+import { GOVUKFrontendSupportError } from '../../errors/index.mjs'
 
 /**
  * Error summary component
@@ -22,17 +23,11 @@ export class ErrorSummary {
    * @param {ErrorSummaryConfig} [config] - Error summary config
    */
   constructor($module, config) {
-    // Some consuming code may not be passing a module,
-    // for example if they initialise the component
-    // on their own by directly passing the result
-    // of `document.querySelector`.
-    // To avoid breaking further JavaScript initialisation
-    // we need to safeguard against this so things keep
-    // working the same now we read the elements data attributes
-    if (
-      !($module instanceof HTMLElement) ||
-      !document.body.classList.contains('govuk-frontend-supported')
-    ) {
+    if (!document.body.classList.contains('govuk-frontend-supported')) {
+      throw new GOVUKFrontendSupportError()
+    }
+
+    if (!($module instanceof HTMLElement)) {
       return this
     }
 
--- a/packages/govuk-frontend/src/govuk/components/error-summary/error-summary.mjs
+++ b/packages/govuk-frontend/src/govuk/components/error-summary/error-summary.mjs
@@ -1,13 +1,13 @@
 import { mergeConfigs } from '../../common/index.mjs'
 import { normaliseDataset } from '../../common/normalise-dataset.mjs'
-import { GOVUKFrontendSupportError } from '../../errors/index.mjs'
+import { GOVUKFrontendComponent } from '../../govuk-frontend-component.mjs'
 
 /**
  * Error summary component
  *
  * Takes focus on initialisation for accessible announcement, unless disabled in configuration.
  */
-export class ErrorSummary {
+export class ErrorSummary extends GOVUKFrontendComponent {
   /** @private */
   $module
 
@@ -23,9 +23,7 @@ export class ErrorSummary {
    * @param {ErrorSummaryConfig} [config] - Error summary config
    */
   constructor($module, config) {
-    if (!document.body.classList.contains('govuk-frontend-supported')) {
-      throw new GOVUKFrontendSupportError()
-    }
+    super()
 
     if (!($module instanceof HTMLElement)) {
       return this
--- a/packages/govuk-frontend/src/govuk/components/notification-banner/notification-banner.test.js
+++ b/packages/govuk-frontend/src/govuk/components/notification-banner/notification-banner.test.js
@@ -203,4 +203,20 @@ describe('Notification banner', () => {
       })
     })
   })
+
+  describe('errors at instantiation', () => {
+    it('throws when GOV.UK Frontend is not supported', async () => {
+      await expect(
+        renderAndInitialise(page, 'notification-banner', {
+          params: examples.default,
+          beforeInitialisation() {
+            document.body.classList.remove('govuk-frontend-supported')
+          }
+        })
+      ).rejects.toEqual({
+        name: 'SupportError',
+        message: 'GOV.UK Frontend is not supported in this browser'
+      })
+    })
+  })
 })
--- a/packages/govuk-frontend/src/govuk/components/notification-banner/notification-banner.test.js
+++ b/packages/govuk-frontend/src/govuk/components/notification-banner/notification-banner.test.js
@@ -4,142 +4,19 @@ const {
 } = require('govuk-frontend-helpers/puppeteer')
 const { getExamples } = require('govuk-frontend-lib/files')
 
-describe('Notification banner, when type is set to "success"', () => {
+describe('Notification banner', () => {
   let examples
 
   beforeAll(async () => {
     examples = await getExamples('notification-banner')
   })
 
-  it('has the correct tabindex attribute to be focused with JavaScript', async () => {
-    await goToComponent(page, 'notification-banner', {
-      exampleName: 'with-type-as-success'
-    })
-
-    const tabindex = await page.$eval('.govuk-notification-banner', (el) =>
-      el.getAttribute('tabindex')
-    )
-
-    expect(tabindex).toEqual('-1')
-  })
-
-  it('is automatically focused when the page loads', async () => {
-    await goToComponent(page, 'notification-banner', {
-      exampleName: 'with-type-as-success'
-    })
-
-    const activeElement = await page.evaluate(() =>
-      document.activeElement.getAttribute('data-module')
-    )
-
-    expect(activeElement).toBe('govuk-notification-banner')
-  })
-
-  it('removes the tabindex attribute on blur', async () => {
-    await goToComponent(page, 'notification-banner', {
-      exampleName: 'with-type-as-success'
-    })
-
-    await page.$eval(
-      '.govuk-notification-banner',
-      (el) => el instanceof window.HTMLElement && el.blur()
-    )
-
-    const tabindex = await page.$eval('.govuk-notification-banner', (el) =>
-      el.getAttribute('tabindex')
-    )
-    expect(tabindex).toBeNull()
-  })
-
-  describe('and auto-focus is disabled using data attributes', () => {
-    beforeAll(async () => {
-      await goToComponent(page, 'notification-banner', {
-        exampleName: 'auto-focus-disabled,-with-type-as-success'
-      })
-    })
-
-    it('does not have a tabindex attribute', async () => {
-      const tabindex = await page.$eval('.govuk-notification-banner', (el) =>
-        el.getAttribute('tabindex')
-      )
-
-      expect(tabindex).toBeNull()
-    })
-
-    it('does not focus the notification banner', async () => {
-      const activeElement = await page.evaluate(() =>
-        document.activeElement.getAttribute('data-module')
-      )
-
-      expect(activeElement).not.toBe('govuk-notification-banner')
-    })
-  })
-
-  describe('and auto-focus is disabled using JavaScript configuration', () => {
-    beforeAll(async () => {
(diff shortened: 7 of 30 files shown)


packages/govuk-frontend/src/govuk/all.mjs
```diff
@@ -103,6 +104,7 @@ export {
   Checkboxes,
   ErrorSummary,
   ExitThisPage,
+  GOVUKFrontendError,
```
romaricpascal, line 107: Only export our base `GOVUKFrontendError` rather than all error classes with the aim of:

- allowing separating our own errors from native ones in bulk, which users can do running `instanceof GOVUKFrontendError`
- but not crowding our export with each error we throw, as users can check the type of error through [their `name` property](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/name)

packages/govuk-frontend/src/govuk/errors/index.mjs
```diff
@@ -0,0 +1,33 @@
...
+ *   // set a new `name` on the subclass. The `name` property is important
+ *   // to ensure intelligible error names even if the class name gets
+ *   // mangled by a minifier
+ *   name = "MissingRootError"
+ * }
+ * ```
+ * @abstract
+ */
+export class GOVUKFrontendError extends Error {
+  name = 'GOVUKFrontendError'
```
romaricpascal, line 20: Annoying, but necessary. Simply extending `Error` doesn't give the error a new name, unfortunately. 

And going for something automated like `this.name = this.constructor.name` in `GOVUKFrontendError`'s `constructor` means the error gets a mangled name once minifier have run over the code and rename the class to say `n`. While we may prevent the mangling in our files, we don't control how services minify theirs, so hardcoding the name limits risks of having an error with the name of `n`.

docs/contributing/coding-standards/js.md
```diff
@@ -133,6 +133,40 @@ Avoid using namespace imports (`import * as namespace`) in code bundled for Comm
 
 Prefer named exports over default exports to avoid compatibility issues with transpiler "synthetic default" as discussed in: https://github.com/alphagov/govuk-frontend/issues/2829
 
+## Throwing errors
```
romaricpascal, line 136: This is just a starter for 10 listing the spirit in which the errors have been implemented in this PR. Happy to discuss and update.

packages/govuk-frontend/src/govuk/all.mjs
colinrotherham: Fab work in this PR Romaric

Do we want to discuss if we're ready for errors to be part of our "public" API?

Bit like the **I18n** class, we chose not to export it

(Gives us freedom to keep iterating without breaking changes)

packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
```diff
@@ -113,7 +114,11 @@ export class Accordion {
    * @param {AccordionConfig} [config] - Accordion config
    */
   constructor ($module, config) {
-    if (!($module instanceof HTMLElement) || !document.body.classList.contains('govuk-frontend-supported')) {
+    if (!document.body.classList.contains('govuk-frontend-supported')) {
+      throw new GOVUKFrontendNotSupportedError()
```
colinrotherham, line 118: Just a random thought

Similar to how `TypeError` isn't `NotCorrectTypeError`, should we flip out the `Not`?

```suggestion
      throw new GOVUKFrontendSupportError()
```

packages/govuk-frontend/src/govuk/components/accordion/accordion.test.js
```diff
@@ -508,6 +508,28 @@ describe('/components/accordion', () => {
...
+          beforeAll(async () => {
+            examples = await getExamples('accordion')
+          })
+
+          it('throws when GOV.UK Frontend is not supported', async () => {
+            await expect(
+              renderAndInitialise(page, 'accordion', {
+                params: examples.default,
+                beforeInitialisation () {
+                  document.body.classList.remove('govuk-frontend-supported')
```
colinrotherham, line 524: 🎉

packages/govuk-frontend/src/govuk/components/character-count/character-count.test.js
```diff
@@ -648,7 +670,7 @@ describe('Character count', () => {
           // Override maxlength to 10
           maxlength: 10
         },
-        initialiser ($module) {
+        beforeInitialisation ($module) {
```
colinrotherham, line 673: As it's no longer with us, as a sign of respect for `.init()` shall we rename this? 😆

Suppose it matches `initAll()` a bit better too

```suggestion
        beforeInit ($module) {
```

packages/govuk-frontend/src/govuk/components/error-summary/error-summary.mjs
```diff
@@ -22,14 +23,11 @@ export class ErrorSummary {
    * @param {ErrorSummaryConfig} [config] - Error summary config
    */
   constructor ($module, config) {
-    // Some consuming code may not be passing a module,
-    // for example if they initialise the component
-    // on their own by directly passing the result
-    // of `document.querySelector`.
-    // To avoid breaking further JavaScript initialisation
```
colinrotherham, line 29: I was quite fond of this message, all on its own in **ErrorSummary**

Thoughts on adding a little disclaimer to every component?

1. Error throws
2. Error console shows stack trace
3. Stack trace uses original source via source maps
4. Developer clicks to these lines and reads helpful message

packages/govuk-frontend/src/govuk/components/globals.test.js
```diff
@@ -37,6 +37,7 @@ describe('GOV.UK Frontend', () => {
         'Checkboxes',
         'ErrorSummary',
         'ExitThisPage',
+        'GOVUKFrontendError',
```
colinrotherham, line 40: Related to https://github.com/alphagov/govuk-frontend/pull/4030/files#r1284215663

Should we not export our errors in the default export, we'll need to manage Terser mangling:

https://github.com/alphagov/govuk-frontend/blob/91d0a5d8b694875b34eb4be52ecf155f8b3701d0/packages/govuk-frontend/rollup.release.config.mjs#L31-L33

packages/govuk-frontend/src/govuk/components/notification-banner/notification-banner.test.js
```diff
@@ -1,165 +1,219 @@
...
   let examples
 
   beforeAll(async () => {
     examples = await getExamples('notification-banner')
   })
 
-  it('has the correct tabindex attribute to be focused with JavaScript', async () => {
-    await goToComponent(page, 'notification-banner', {
-      exampleName: 'with-type-as-success'
+  describe('when type is set to "success"', () => {
```
colinrotherham, line 11: Nice little test refactor with the `describe()` nesting, mind if we split it into another commit?

packages/govuk-frontend/src/govuk/components/radios/radios.test.js
```diff
@@ -1,236 +1,261 @@
...
+const { goToComponent, goToExample, getProperty, getAttribute, isVisible, renderAndInitialise } = require('govuk-frontend-helpers/puppeteer')
+const { getExamples } = require('govuk-frontend-lib/files.js')
 
-describe('Radios with conditional reveals', () => {
-  describe('when JavaScript is unavailable or fails', () => {
-    beforeAll(async () => {
-      await page.setJavaScriptEnabled(false)
-    })
+describe('Radios', () => {
+  describe('with conditional reveals', () => {
```
colinrotherham, line 5: Same with this one, separate commit?

Jest output is improved too ⭐

packages/govuk-frontend/src/govuk/errors/index.mjs
```diff
@@ -0,0 +1,33 @@
+/**
+ * A base class for `Error`s thrown by GOV.UK Frontend.
```
colinrotherham, line 2: Mind if we keep a short and sweet "title" at the top for any generated docs?

```suggestion
 * GOV.UK Frontend error
 *
 * A base class for errors thrown by GOV.UK Frontend.
```

packages/govuk-frontend/src/govuk/errors/index.mjs
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * It is meant to be extended into specific types of errors
+ * to be thrown by our code.
+ *
+ * @example
+ * ```js
```
colinrotherham, line 8: Just a reminder to use `mjs` over `js` for ES modules code blocks

Can't remember if ESLint is set up for these (like it is for Markdown) but it might matter one day

packages/govuk-frontend/src/govuk/errors/index.mjs
```diff
@@ -0,0 +1,33 @@
...
+  name = 'GOVUKFrontendError'
+}
+
+/**
+ * Indicates that GOV.UK Frontend is not supported
+ */
+export class GOVUKFrontendNotSupportedError extends GOVUKFrontendError {
+  name = 'GOVUKFrontendNotSupportedError'
+
+  /** */
```
colinrotherham, line 29:
```suggestion
```

shared/helpers/puppeteer.js
```diff
@@ -91,19 +90,42 @@ async function renderAndInitialise (page, componentName, options) {
...
+    await page.$eval('[data-module]', options.beforeInitialisation)
+  }
+
   // Run a script to init the JavaScript component
-  await page.evaluate(async (exportName, options) => {
+  //
+  // Use `evaluate` to ensure we run `document.querySelector` inside the
+  // browser, like users would, rather than rely on Puppeteer looking for the
+  // element which would cause an error in Jest-land rather than within the
+  // browser if the element is missingß
```
colinrotherham, line 105:
```suggestion
  // browser if the element is missing
```

shared/helpers/puppeteer.js
```diff
@@ -91,19 +90,42 @@ async function renderAndInitialise (page, componentName, options) {
...
-
     const namespace = await import('govuk-frontend')
-    /* eslint-disable-next-line no-new */
-    new namespace[exportName]($module, options.config)
+
+    try {
+      /* eslint-disable-next-line no-new */
+      new namespace[exportName]($module, options.config)
+    } catch ({ name, message }) {
+      return { name, message }
```
colinrotherham, line 121: So even if we re-throw (not return) on this line it doesn't work? 😣

packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
romaricpascal: Fair, and that makes it shorter, I'll rename 😊

packages/govuk-frontend/src/govuk/components/error-summary/error-summary.mjs
romaricpascal: I think most of that info should be summarised in the error message when implementing `https://github.com/alphagov/govuk-frontend/issues/4035`.

Regarding adding things to each components, I think this would rather be the role of a base Component class. I think with the error throwing for GOV.UK Frontend support and the type check of `$module` there's enough ground for creating it already (only to add new features to it, not refactor existing code into it) and centralise these checks and possible comments.

packages/govuk-frontend/src/govuk/components/globals.test.js
romaricpascal: Can you expand on what we'll need to manage on the Terser side if we don't export the errors internally? 

If it's about the error naming after mangling, [the hardcoded `name` property is there to handle the situation](https://github.com/alphagov/govuk-frontend/pull/4030#discussion_r1280921668) (especially as we can't guarantee that library users will take the same step as us towards not mangling the name of the errors/our exports).

packages/govuk-frontend/src/govuk/components/notification-banner/notification-banner.test.js
romaricpascal: No probs, I'll split the extra `describe`s in their own commit to separate them from the implementation of the error throwing.

shared/helpers/puppeteer.js
romaricpascal: Indeed, the information was still lost when re-throwing 😣 I didn't think of checking what lands in the `pageerror` event, though so I'll give that a quick go if it can avoid us that juggling.

packages/govuk-frontend/src/govuk/errors/index.mjs
romaricpascal: The example doesn't use anything module specific (`import`/`export`) so plain `js` is fine there 😊 

Regarding ESLint, it doesn't seem we've enabled [jsdoc/check-examples](https://github.com/gajus/eslint-plugin-jsdoc/blob/HEAD/docs/rules/check-examples.md#readme). We can discuss its enabling in a separate part as it may have some tricky side effects (eg. requiring this class to have its own JSDoc block inside the example).

shared/helpers/puppeteer.js
romaricpascal: No more luck looking at the `pageerror` and `error` events on the page. If I remove the `try/catch` in the code that's run by `page.evaluate`, no event gets triggered. The promise for `page.evaluate` simply rejects, with only a `name` property to the mangled class name of the error (see the following Jest output) 😞 

```diff
- Object {
-   "message": "GOV.UK Frontend is not supported in this browser",
-   "name": "GOVUKFrontendNotSupportedError",
+ Error {
+   "name": "n",
}
```

packages/govuk-frontend/src/govuk/errors/index.mjs
```diff
@@ -0,0 +1,33 @@
...
+ * ```js
+ * class MissingRootError extends GOVUKFrontendError {
+ *   // Setting an explicit name is important as extending the class will not
+ *   // set a new `name` on the subclass. The `name` property is important
+ *   // to ensure intelligible error names even if the class name gets
+ *   // mangled by a minifier
+ *   name = "MissingRootError"
+ * }
+ * ```
+ * @abstract
```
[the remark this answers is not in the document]
36degrees: Worth enforcing this in the constructor too?

```js
export class GOVUKFrontendError extends Error {
  name = 'GOVUKFrontendError'

  constructor () {
    if (this.constructor == GOVUKFrontendError) {
      throw new Error("GOVUKFrontendError is an abstract class and can't be instantiated");
    }
  }
}
```

Or is errors throwing errors too confusing?

shared/helpers/puppeteer.js
romaricpascal: What's weird is that puppeteer on its own seems to be happy with errors that have a custom name thrown from `page.evaluate`.

Taking the following script:

```mjs
import puppeteer from 'puppeteer';

// Launch the browser and open a new blank page
const browser = await puppeteer.launch({
  headless: 'new'
});
const page = await browser.newPage();

await page.goto('about:blank')

try {
  await page.evaluate(() => {
    class CustomError extends Error {
      name = 'CustomError'
    }

    throw new CustomError('Custom message in custom error')
  })
} catch (error) {
  console.log('Error class:', error.constructor.name)
  const {name, message, cause } = error
  console.log('Contents:', {name, message, cause})
  console.log(error)
}
```

The logs end up as expected (the class on the Node side can only be `Error` as the `CustomError` only exists inside the browser's page).

```sh
Error class: Error
Contents: {
  name: 'CustomError',
  message: 'Custom message in custom error',
  cause: undefined
}
Error [CustomError]: Custom message in custom error
```

So there's something in between that messes with our `page.evaluate` and requires us to add that workaround with `throw`.

I had a quick look in the sources of argos-ci/jest-puppeteer (especially jest-environment-puppeteer), but didn't notice anything that would wrap `page.evaluate` to mess with the error thrown. The page seems to be a regular Puppeteer page 😔

colinrotherham: Is the name-shortening to `n` [from Terser](https://github.com/alphagov/govuk-frontend/pull/4030#discussion_r1284234033)?

packages/govuk-frontend/src/govuk/components/globals.test.js
colinrotherham: Yeah that `mangle.reserved` option

But if the mangled error class name doesn't appear in stack traces then all good

shared/helpers/puppeteer.js
romaricpascal: Yup, that's what ends up in the `govuk-frontend.min.js` that's being built in `dist` and then loaded by the review app 😊

packages/govuk-frontend/src/govuk/errors/index.mjs
colinrotherham: Yeah not so much the module specific stuff

We tend to use [`*.mjs` as the default now](https://github.com/alphagov/govuk-frontend/commit/4f6a747ce3b25e7a5323b772fd4ca47a8049780b), so some rules like the TypeScript parser are only for `*.mjs`

https://github.com/alphagov/govuk-frontend/blob/e1ef6d2df52908e1b87a8c692bc894adf015f57c/packages/govuk-frontend/.eslintrc.js#L11-L13

packages/govuk-frontend/src/govuk/errors/index.mjs
romaricpascal: Oh, no, that wasn't whoops, it's making ESLint happy as there's not much to document on that constructor 🤔 I'll replace with a more explicit disabling of ESLint for that line 😊

romaricpascal:
Deleted comments were from the dist diff action, which I had triggered by mistakenly commiting a test build of `dist` 🤦🏻 

@colinrotherham I think I've addressed all the code bits of your feedback (discussions on which bits of the errors are public API to be had). I've also added an extra commit to refactor the validation into a central class as we know that's the architecture we're going towards, so there's not much sense spreading new features all across the components when we can already have them centralised from the get go 😊 I'll try centralising the testing tomorrow so hold your review for a bit please (though the class is tested via all the class inherinting... really wish we could gather coverage in the browser 😔 )

packages/govuk-frontend/src/govuk/component.mjs
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * Base Component class
+ *
+ * Centralises the behaviours shared by our components
+ *
+ * @internal
+ * @abstract
+ */
+export class Component {
```
colinrotherham, line 11: 🎉

packages/govuk-frontend/src/govuk/component.mjs
```diff
@@ -0,0 +1,31 @@
...
+  /**
+   * Constructs a new component, validating that GOV.UK Frontend is supported
+   *
+   * @internal
+   */
+  constructor () {
+    this.validateGOVUKFrontendSupport()
```
romaricpascal, line 18: Idea is to keep that constructor a series of method calls to give an easy overview of what's happening. It'll also allow fine grained overriding should someone need to reach to the internals to amend one of these behaviours 😊

colinrotherham, line 18: Probably pair a shorter `checkSupport()` with an `isSupported` getter?

Based on our current method names:

```
checkMode()
checkIfAllSectionsOpen()
checkForSessionStorage()
```

colinrotherham: Just seen your comment, holding off!

romaricpascal: Sounds fine with me, I'll update tomorrow.

Regarding `isSupported()`, I'd be keen to have it as a utility function in `govuk/common/index.mjs` rather than on the class itself as there's no `this` involved in its execution.

packages/govuk-frontend/src/govuk/common/index.mjs
```diff
@@ -133,3 +133,17 @@ export function extractConfigByNamespace(configObject, namespace) {
...
+ *
+ * Some pages may be able to load our JavaScript, but GOV.UK Frontend may not be
+ * supported due to our check relying on support for `nomodule` rather than
+ * `type="module"` (as we can't detect the latter)
+ *
+ * @internal
+ * @returns {boolean} Whether GOV.UK Frontend is supported on this page
+ */
+export function isSupported() {
+  return document.body.classList.contains('govuk-frontend-supported')
```
colinrotherham, line 148: Possibly want to pass in `document` as a parameter here?

For use in Puppeteer helpers or Cheerio tests where `document` might not be global

Did something similar in Design System's [`new Navigation(document)`](https://github.com/alphagov/govuk-design-system/blob/e8224f2f1bbd4dae82c11b73b71495a07c120d93/src/javascripts/components/navigation.mjs#L8-L14)

romaricpascal:
> I'll try centralising the testing tomorrow so hold your review for a bit please (though the class is tested via all the class inherinting... really wish we could gather coverage in the browser 😔 )

So... centralising a test for `Component` would actually require that class to be exported. Otherwise, it can't be accessed in the scripts that run in the browser to be exercised. Until we make it part of our public API, we'll have to exercise it through the components that extend it. I think that's fine though, as we do need to test these components anyways.

I think it's good for review now @colinrotherham 😊

packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
```diff
@@ -14,7 +15,7 @@ import { I18n } from '../../i18n.mjs'
  * The state of each section is saved to the DOM via the `aria-expanded`
  * attribute, which also provides accessibility.
  */
-export class Accordion {
+export class Accordion extends Component {
```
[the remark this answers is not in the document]
romaricpascal: Oh, that's not something I had considered too much, just went with what felt right. `GOVUKComponent` feels very verbose, as well as misnamed (I'd think a `GOVUKComponent` would be part of the GOV.UK project). `GOVUKFrontendComponent` even more verbose. I'd be keen to keep `Component` for shortness.

For the `Error`, though, I think a little verbosity is useful to avoid confusion of what `throw new Error()` does (is it ours or a native error). If we shorten to `Error`, people would have to `import {Error as GOVUKFrontendError} from 'govuk-frontend'` to make their code clear, so we may as well take that step already 😊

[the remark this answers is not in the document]
colinrotherham: Feels strange having two approaches

How about `BaseError` and `BaseComponent`?

packages/govuk-frontend/src/govuk/errors/index.mjs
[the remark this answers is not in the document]
romaricpascal: I'm not sure it's worth enforcing the `@abstract` in code. I'd rather lean towards shipping less code and if we get evidence that people misuse it, clamp down on it rather than be pre-emptively defensive 😊

packages/govuk-frontend/src/govuk/common/index.mjs
romaricpascal: Ohhh, good spot. Missed a unit test for this method, which will require passing JSDom's document 😊 I'll add a `$scope` argument defaulted to `document.body` (to avoid people having to set it themselves).

packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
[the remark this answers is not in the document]
romaricpascal: I think it's important the error gets prefixed with `GOVUKFrontend`. It makes `instanceof` checks clearer: `instanceof GOVUKFrontendError` is more explicit than `instanceof BaseError`. This prefix would hopefully also make people think twice before extending it for their own purposes: we'd want them to throw their own kind of error if possible so there's a separation between service errors and our own library errors.

For the component bit, I think on the contrary it's important that its not labelled `GOVUKFrontend`. Down the road, we're imagining encouraging people to extend it. Creating `GOVUKFrontendComponent`s reads like creating "components of GOVUKFrontend", while it'd be just `Components` for their service.

As we haven't released yet, we can still tweak namings. Would you be fine moving naming discussion in its own issue while we implement the other errors? 😊

packages/govuk-frontend/src/govuk/all.mjs
colinrotherham: Let's default to not exporting it yet, can always add it in another PR

packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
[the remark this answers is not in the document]
colinrotherham: Quick Slack decision?

Your point is great for `instanceof GOVUKFrontendComponent` too

packages/govuk-frontend/src/govuk/common/index.mjs
```diff
@@ -133,3 +133,17 @@ export function extractConfigByNamespace(configObject, namespace) {
...
+ * Checks if GOV.UK Frontend is supported on this page
+ *
+ * Some browsers will load and run our JavaScript but GOV.UK Frontend
+ * won't be supported.
+ *
+ * @internal
+ * @param {HTMLElement} [$scope] - The `<body>` element of the document to check for support
+ * @returns {boolean} Whether GOV.UK Frontend is supported on this page
+ */
+export function isSupported($scope = document.body) {
```
colinrotherham, line 147: Do we have browser support for parameter defaults without Babel polyfills?

romaricpascal: Yup, Babel leaves that line as is in `dist`. Support looks [a bit older than what we're configured to support](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters#browser_compatibility). And if transpilation happened, there's [no polyfills involved](https://babeljs.io/repl#?browsers=IE%2011&build=&builtIns=false&corejs=false&spec=false&loose=true&code_lz=GYVwdgxgLglg9mABDAzgZRAB03ATlAUwBMAKAEhQjkwMQF5Ei4IQBbAsKAOgCM4iAngEpEAbwBQiRLgJQQuJBSo0uEADYBDFCgAyqblU4aYYFCQDkAczgA3EAGsAtMFwJCYIo5RYc-YuaEAbnEAXyA&debug=false&forceAllTransforms=false&modules=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=true&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Cstage-2%2Cstage-3&prettier=false&targets=&version=7.22.10&externalPlugins=&assumptions=%7B%7D). So all good there 🎉

colinrotherham (approved):

packages/govuk-frontend/src/govuk/components/globals.test.js
romaricpascal: Added a comment about mangling on this PR: https://github.com/alphagov/govuk-frontend/pull/4075.

The name is clear in the logged message, but mangled in the stacktrace. I think mangling is worth its own discussion, as it may be worth disabling it altogether rather than keep maintaining an evergrowing list of names to not mangle for our `min.js` file.

romaricpascal: And opened https://github.com/alphagov/govuk-frontend/issues/4076 to explore our options.

PR #4039, opened 2023-08-03
Add stats comment to Pull Requests

### LINKED ISSUE #3282: Display file sizes and module breakdown on pull requests
Part of https://github.com/alphagov/govuk-frontend/issues/3279

## What

Add a comment on PR that summarises the statistics collected from the content of our package. This may include (to keep below the 65536 character limit of comments):

- file size inside dist
- file size of package/govuk/all.js
- link to review app for file sizes of other files in package/govuk
- size and number of modules for each file going through rollup-plugin-visualizer, with a link to the review app for the actual breakdown

The comment should include the SHA of the commit the stats relate to, so we can check it is up to date with the tip of the PR branch. To cater for times where this may get out of sync, the workflow collecting the stats should be [runnable manually](https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/).

If possible, the code for posting comments on the PR should be made generic so we can re-use it to add other comments on the PR (for example the diff of `dist`).[^1]

## Why

Having these information accessible in the review app is handy, but hidden away. Commenting on the pull requests will give us more immediate access to the information.

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [ ] An existing or new workflow comments on our PRs with the stats[^2]
- [ ] Either:
  - [ ] The code for commenting on the PR is generic
  - or
  - [ ] An issue was created to make the code generic

[^1]: The approach could be the same as in the spike, but using a comment as marker to recognise the comment rather than actual content in the comment.
[^2]: Little note of caution around trying not to increase the build time on CI too much

domoscargin:
Closes #3282


.github/workflows/scripts/comments.mjs
```diff
@@ -75,6 +77,59 @@ export async function commentDiff (
...
+  const fileSizes = await getFileSizes()
+  const fileSizeRows = Object.entries(fileSizes).map(([key, value]) => [key, String(value)])
+  const fileSizeHeaders = ['File', 'Size']
+  const fileSizeTable = renderTable(fileSizeHeaders, fileSizeRows)
+  const fileSizeText = [fileSizeTitle, fileSizeTable].join('\n')
+
+  // Modules
+  const modulesTitle = '### Modules'
+  const modules = Object.fromEntries(await Promise.all(modulePaths
+    .filter((modulePath) => typeof modulePath === 'string')
```
domoscargin, line 105: Stole this from https://github.com/alphagov/govuk-frontend/blob/47db9411a00c49f0d05ead84fb178a97600895d4/packages/govuk-frontend-review/src/app.mjs#L47

But it requires extra guards here to satisfy TypeScript.

shared/lib/package.json
```diff
@@ -18,5 +18,8 @@
     "js-yaml": "^4.1.0",
     "minimatch": "^9.0.3",
     "slash": "^3.0.0"
+  },
+  "devDependencies": {
+    "@types/js-yaml": "^4.0.5"
```
domoscargin, line 23: When adding `paths` to `tsconfig.base.json`, this gets flagged

shared/stats/src/index.mjs
```diff
@@ -9,6 +11,21 @@ import { getComponentNames, filterPath, getYaml } from 'govuk-frontend-lib/files
 const componentNamesWithJavaScript = await getComponentNames((componentName, componentFiles) =>
   componentFiles.some(filterPath([`**/${componentName}.mjs`])))
 
+/**
+ * Target files for file size analysis
+ */
+const filesForAnalysis = {
```
domoscargin, line 17: Considered doing this a bit more automagically, but I think for now just defining the files we want is fine. We can always finesse later.

shared/stats/src/index.mjs
```diff
@@ -44,14 +61,50 @@ export async function getStats (modulePath) {
...
     .reduce((total, rendered) => total + rendered, 0)
 
-  return { total, modules }
+  return { total, modules, moduleCount }
+}
+
+/**
+ * Returns file sizes of key files
+ *
+ * @returns {Promise<{[key: string]: string | number | any[] | FileSizeObject}>} - File names and size
```
domoscargin, line 77: My best attempt at a return type. `filesize` exports the `filesize` function, but doesn't export what that function returns, so I've hacked it in here.

shared/lib/package.json
colinrotherham: Ahh probably because we have strict mode turned on?

Would you mind adding it to the root of the project inside `optionalDependencies`?

Keeps all our optional types together

colinrotherham:
@domoscargin Love seeing these stats appearing 😍

domoscargin:
Hmm, application error on the deployment...

shared/stats/src/index.mjs
```diff
@@ -44,12 +56,31 @@ export async function getStats (modulePath) {
...
+  return { total, modules, moduleCount }
+}
+
+/**
+ * Returns file sizes of key files
+ *
+ * @returns {Promise<{[key: string]: import('fs').Stats}>} - File names and size
+ */
+export async function getFileSizes () {
+  /** @type { { [key: string]: import('fs').Stats } } */
```
domoscargin, line 75: Anyway to rewrite this without this messy type annotation?

shared/stats/src/index.mjs
```diff
@@ -9,6 +10,17 @@ import { getComponentNames, filterPath, getYaml } from 'govuk-frontend-lib/files
 const componentNamesWithJavaScript = await getComponentNames((componentName, componentFiles) =>
   componentFiles.some(filterPath([`**/${componentName}.mjs`])))
 
+/**
+ * Target files for file size analysis
+ */
+const filesForAnalysis = [
+  `dist/govuk-frontend-${pkg.version}.min.js`,
```
domoscargin, line 17: Could potentially move this out to a separate file.

shared/stats/src/index.mjs
```diff
@@ -44,12 +56,31 @@ export async function getStats (modulePath) {
...
+ * Returns file sizes of key files
+ *
+ * @returns {Promise<{[key: string]: import('fs').Stats}>} - File names and size
+ */
+export async function getFileSizes () {
+  /** @type { { [key: string]: import('fs').Stats } } */
+  const result = {}
+
+  for (const filename of filesForAnalysis) {
+    const stats = await stat(join(paths.root, filename))
```
domoscargin, line 79: Might need a tweak to account for different versions.

.github/workflows/scripts/comments.mjs
colinrotherham: The extra guards will be because TypeScript knows arrays can be pushed, popped, changed etc

For example, mapping a list of items:

```mjs
list.map((listItem) => [listItem, '123'])
```

By adding `/** @type {const} */` we say "this array is constant, it won't change" (parentheses are key)

```mjs
list.map((listItem) => /** @type {const} */(
  [listItem, '123']
))
```

Which avoids the need for those extra safety checks on later lines

.github/workflows/scripts/comments.mjs
```diff
@@ -75,6 +78,62 @@ export async function commentDiff (
...
+
+  // File sizes
+  const fileSizeTitle = '### File sizes'
+  const fileSizes = await getFileSizes()
+  const fileSizeRows = Object.entries(fileSizes)
+    .map(([key, value]) => {
+      return [key, String(filesize(value.size, { base: 2 }))]
+    })
+  const fileSizeHeaders = ['File', 'Size']
+  const fileSizeTable = renderTable(fileSizeHeaders, fileSizeRows)
```
colinrotherham, line 103: We similarly loop `componentNamesWithJavaScript` in the review app to build a summary list

Should we save a job and build the rows for both review app and PR stats in `'govuk-frontend-stats'`?

https://github.com/alphagov/govuk-frontend/blob/5804103ad5d417222f3b114c8c2db4e694ba2801/packages/govuk-frontend-review/src/views/index.njk#L116-L117

Saves maintaining two sets of table building logic

shared/stats/src/index.mjs
colinrotherham: Yeah, using the same array mapping idea to preserve types?

```mjs
const fileTasks = filesForAnalysis
  .map(async (filename) => /** @type {const} */(
    [filename, await stat(join(paths.root, filename))]
  ))

return Promise.all(fileTasks)
  .then(Object.fromEntries)
```

shared/stats/src/index.mjs
colinrotherham: Yeah it's a bit of a worry that we're using `pkg.version` and `paths.root`

For example, we could have [GOV.UK Frontend v4.7.0 installed in the review app + stats packages](https://github.com/alphagov/govuk-frontend/blob/main/shared/stats/README.md#stats-for-previous-version):

```shell
npm install govuk-frontend@4.7.0 --workspace packages/govuk-frontend-review
npm install govuk-frontend@4.7.0 --workspace shared/stats
```

The include `<script type="module">` wouldn't work but the names, files, data would all be correct

We'd see stats coming from `shared/stats/node_modules/govuk-frontend` not `packages/govuk-frontend`. Makes it difficult to maintain `filesForAnalysis` doesn't it 😣 

Could we just stick to component module file sizes for now?

Then we only need to loop `componentNamesWithJavaScript` regardless of version:

https://github.com/alphagov/govuk-frontend/blob/5804103ad5d417222f3b114c8c2db4e694ba2801/packages/govuk-frontend-review/src/views/index.njk#L116-L117

colinrotherham: I've made sure all helpers work with any GOV.UK Frontend version in:

* https://github.com/alphagov/govuk-frontend/pull/4041

So it might be nice to move the filesize helpers to `govuk-frontend-lib/files` too

tsconfig.base.json
```diff
@@ -14,7 +14,13 @@
     "target": "ESNext",
     "types": ["jest", "jest-axe", "jest-puppeteer", "node"],
     "paths": {
-      "govuk-frontend": ["./packages/govuk-frontend/src/govuk/all.mjs"]
+      "govuk-frontend": ["./packages/govuk-frontend/src/govuk/all.mjs"],
```
colinrotherham, line 17: This is great 🙌

Had a good check and probably need a few more with `/*` wildcards?

```suggestion
      "govuk-frontend": ["./packages/govuk-frontend/src/govuk/all.mjs"],
      "govuk-frontend-stats": ["./shared/stats/src/index.mjs"],
      "govuk-frontend-config": ["./shared/config/index.js"],
      "govuk-frontend-helpers": ["./shared/helpers/index.js"],
      "govuk-frontend-helpers/*": ["./shared/helpers/*.js"],
      "govuk-frontend-helpers/jest/*": ["./shared/helpers/jest/*"],
      "govuk-frontend-lib": ["./shared/lib/index.js"],
      "govuk-frontend-lib/*": ["./shared/lib/*.js"]
```

colinrotherham: We'll finally be able to remove all the places we include `govuk-frontend`

```json
  "include": ["**/*.js", "**/*.mjs", "../../packages/govuk-frontend"],
  "exclude": ["../../packages/govuk-frontend/dist/**"]
```

Leaving it like:

```json
  "include": ["**/*.js", "**/*.mjs"]
```

colinrotherham: Or where we have packages that include others:

```json
  "include": [
    "**/*.js",
    "**/*.mjs",
    "../config",
    "../lib",
    "../../packages/govuk-frontend"
  ],
  "exclude": ["./dist/**", "../../packages/govuk-frontend/dist/**"]
```

Won't need to include `"../config"` or `"../lib"` etc

```json
  "include": ["**/*.js", "**/*.mjs"],
  "exclude": ["./dist/**"]
```

Since we'll have them in **tsconfig.base.json** instead 😊

.github/workflows/scripts/comments.mjs
domoscargin: I'm a bit confused since [this](https://github.com/alphagov/govuk-frontend/blob/47db9411a00c49f0d05ead84fb178a97600895d4/packages/govuk-frontend-review/src/app.mjs#L47) imports `modulePaths` and maps it to something and `modulePaths` is correctly listed as `string[]`.

If I import `modulePaths` [here](https://github.com/alphagov/govuk-frontend/blob/5804103ad5d417222f3b114c8c2db4e694ba2801/.github/workflows/scripts/comments.mjs#L4), it lists its type as `(string | undefined)[]`, even if all I do is assign it to a const: `const test = modulePaths`.

I was assuming it was something to do with the scope of govuk-frontend-review's `app.mjs` vs a function in the comments script.

colinrotherham: If you mouse over `modulePath` in the filter params what type does it give you?

domoscargin: `modulePaths` is `(string | undefined)[]`.
`modulePath` is `string | undefined`

colinrotherham: Sorry, you said didn't you

Probably just need that inline comment over in the stats package

Extra checks are caused by `strict: true` in:

https://github.com/alphagov/govuk-frontend/blob/main/.github/workflows/scripts/tsconfig.json

Can turn it off if it's being annoying, or consider replacing with `noImplicitAny` which at least stops "untyped" things from being let through unchecked

domoscargin: Ah, tracked it down.

In stats `index.mjs`, [this fella](https://github.com/alphagov/govuk-frontend/blob/5804103ad5d417222f3b114c8c2db4e694ba2801/shared/stats/src/index.mjs#L29-L33) is used as such:

https://github.com/alphagov/govuk-frontend/blob/5804103ad5d417222f3b114c8c2db4e694ba2801/shared/stats/src/index.mjs#L38-L40

and `packageOptions.modulePath` is `modulePath?: string`

colinrotherham: @domoscargin Nice spot 😊

In that case, we know `modulePath` definitely _isn't_ `undefined`

We can swap this `@type` to `@satisfies` to take the populated values into account too

https://github.com/alphagov/govuk-frontend/blob/5804103ad5d417222f3b114c8c2db4e694ba2801/shared/stats/src/index.mjs#L27

colinrotherham: You'll still need the `/** @type {const} */` on your `.map()` but can at least drop the `.filter()` now

```mjs
  const modules = Object.fromEntries(await Promise.all(modulePaths
    .map(async (modulePath) => /** @type {const} */(
      [modulePath, await getStats(String(modulePath))]
    ))))
```

shared/stats/src/index.mjs
domoscargin: `getFileSizes()` now takes a glob as input, which I'm passing from the `comments` script. It should be sizing everything in `dist` and the files directly in `packages/govuk-frontend/dist/govuk`. In both cases, ignoring `.map` files.

.github/workflows/scripts/comments.mjs
domoscargin: I'd like to get this one merged as is, preferably. I'll add an issue to do some optimised table building.

domoscargin:
@colinrotherham, I've implemented your changes and all looks good.

I've also cleaned up the commit list, because it was getting messy.

.github/workflows/scripts/comments.mjs
```diff
@@ -157,6 +264,12 @@ function githubActionRunUrl(context) {
  * @property {string} markerText - The marker to identify the comment
  */
 
+/**
+ * @typedef {object} StatComment
+ * @property {string} titleText - The title of the comment
+ * @property {string} markerText - The marker to identify the comment
+ */
```
colinrotherham, line 271: Not a blocker, but we could share (or rename) the `DiffComment` type

The workspace path could satisfy `path` to stop any warnings

```mjs
await commentStats({ github, context, commit }, issueNumber, {
  path: '${{ github.workspace }}',
  titleText: ':clipboard: Stats',
  markerText: 'stats'
})
```

colinrotherham (approved):
Looks ace, good to have this merged

Can you double check package versions? Think some had updated

PR #4043, opened 2023-08-04
Add ability to customise inverse button colours

querkmachine:
A "quick-fix" for the inverse button styles using hardcoded blue text and hover colours, where non-GOV.UK branded services probably want this to match the `$govuk-brand-colour` variable.

Related to #3918. Extracted from PR #3925, which includes more significant changes covering all inverse components.

## Changes
- Added public variable `$govuk-inverse-button-background-colour`, which defaults to `govuk-colour("white")`.
- Added public variable `$govuk-inverse-button-text-colour`, which defaults to `$govuk-brand-colour`.
  - This is probably technically a breaking change, as the inverse button's text colour was previously `govuk-colour("blue")`. This will look the same if the brand colour has not been changed, but will be different if it has.
- Mapped the public variables to their scoped equivalents.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,24 @@ These styles can be used in your own code by including the `govuk-focused-box` S
 
 This change was made in [pull request #3819: Add linked image focus style](https://github.com/alphagov/govuk-frontend/pull/3819).
 
+#### Customise inverse button colours
+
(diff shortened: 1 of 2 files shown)


owenatgov (approved):
💪🏻

owenatgov:
@querkmachine So, based on your reasoning I reckon lets be cautious and call it a breaking change. We want teams to check that it's not going to impact their buttons, even if it's unlikely.

PR #3875, opened 2023-06-29
Format JavaScript using Prettier

colinrotherham:
This PR configures Prettier to format `*.{cjs,js,mjs}` files

I've also added [`eslint-config-prettier`](https://github.com/prettier/eslint-config-prettier#readme) to turn off ESLint rules handled by Prettier, since:

* [StandardJS](https://standardjs.com) removes semicolons almost everywhere, [Prettier](https://prettier.io) inserts them
* [StandardJS](https://standardjs.com) adds spaces before function parentheses, [Prettier](https://prettier.io) removes them

We also have a PR to switch stylesheet formatting to Prettier:

* https://github.com/alphagov/govuk-frontend/pull/3799

PR #4044, opened 2023-08-04
Update testing and linting documentation

colinrotherham:
This PR makes a few testing and linting documentation tweaks

1. Adding EditorConfig info for https://github.com/alphagov/govuk-frontend/pull/3197
2. Added Prettier info for https://github.com/alphagov/govuk-frontend/pull/3201
3. Added auto fixing commands for StyleLint and ESLint

I've included more Prettier info (including auto formatting) where relevant in:

* https://github.com/alphagov/govuk-frontend/pull/3799
* https://github.com/alphagov/govuk-frontend/pull/3875

### DIFF
--- a/docs/releasing/testing-and-linting.md
+++ b/docs/releasing/testing-and-linting.md
@@ -1,6 +1,8 @@
 # Testing and linting
 
-GitHub Actions lints Sass and JavaScript, runs unit and functional tests with Node, and generates snapshots for [visual regression testing](https://www.browserstack.com/percy/visual-testing).
+GitHub Actions lints Sass and JavaScript, runs unit and functional tests with Node.js, and generates snapshots for [visual regression testing](https://www.browserstack.com/percy/visual-testing).
+
+See the [GitHub Actions **Tests** workflow](https://github.com/alphagov/govuk-frontend/actions/workflows/tests.yml) for more information.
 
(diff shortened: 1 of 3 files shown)


docs/releasing/testing-and-linting.md
```diff
@@ -46,9 +52,32 @@ To lint the whole codebase, run:
...
+npm run lint:editorconfig
+```
+
+See [.editorconfig](/.editorconfig) for details.
+
+### Running only Prettier linting
+
+```shell
+npm run lint:prettier
+```
```
romaricpascal, line 78: **issue** I think it's worth mentioning the `--write` option for Prettier similarly to how we mention the `--fix` options for Stylelint and Eslint. I'm not sure if it's compatible with `--check`, though, so we may have to move that flag in the `lint:prettier` script rather than `lint:prettier:cli` 😊

romaricpascal (commented):
Nice additions. I think it's worth mentioning Prettier in both the CSS and JS parts (even if just a link to the `testing-and-linting.md` docs) to signpost that Prettier is used on these files as well.

I'm also thinking it could be worth mentioning the `:cli` scripts to lint specific files or groups of files if people need to lint and `--fix` (or `--write` a single file).

PR #4047, opened 2023-08-07
Mark internal things as `@internal` and add access annotations to the I18n class

### LINKED ISSUE #4017: Mark internal classes and functions as `@internal`
## What

Annotate all of our 'internal' classes or functions (like the I18n class and the mergeConfigs function) as `@internal`.

Remove any other access annotations (like `@private`) from classes and functions. (Individual properties within an `@internal` class can still be marked as private)

## Why

This is part of our decision to [reset the public JavaScript API](https://github.com/alphagov/govuk-design-system-architecture/blob/main/decision-records/008-reset-the-javascript-api.md).

When using access related annotations we want to be clearer about whether the annotation refers to:

- access in context of the class – whether a particular method can be called from outside the class, including by other classes in GOV.UK Frontend
- access in context of the library – whether a particular method is considered to be available for use outside of GOV.UK Frontend

Using `@private`, `@protected` and `@public` only for describing class-level access helps with this.

We can then use `@internal` (and potentially in the future [the other release tags](https://api-extractor.com/pages/tsdoc/doc_comment_syntax/#release-tags)) for library level access annotations.

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [x] All 'internal' classes and functions are marked as `@internal`
- [x] Other access annotations are removed from the class JSDoc and any internal functions

36degrees:
When using access related annotations we want to be clearer about whether the annotation refers to:

- access in context of the class – whether a particular method can be called from outside the class, including by other classes in GOV.UK Frontend
- access in context of the library – whether a particular method is considered to be available for use outside of GOV.UK Frontend

Annotate all of our 'internal' classes or functions (like the I18n class and the mergeConfigs function) as https://github.com/internal.

Remove any other access annotations (like `@private`) from classes and functions.

Add access annotations to the I18n class.

Closes #4017

### DIFF
--- a/packages/govuk-frontend/src/govuk/common/closest-attribute-value.mjs
+++ b/packages/govuk-frontend/src/govuk/common/closest-attribute-value.mjs
@@ -1,7 +1,7 @@
 /**
  * Returns the value of the given attribute closest to the given element (including itself)
  *
- * @private
+ * @internal
  * @param {Element} $element - The element to start walking the DOM tree up
  * @param {string} attributeName - The name of the attribute
  * @returns {string | null} Attribute value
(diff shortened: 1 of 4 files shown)


36degrees:
Looks like this has already been done as part of #3958.

PR #4048, opened 2023-08-07
Document removal of polyfills in changelog

36degrees:
Because our polyfills created or extended global objects ('polluting the global namespace') service teams may have other code that was unintentionally relying on them being included.

Flag this in the changelog so that users are aware and can take action if required.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -293,6 +293,25 @@ If you were using `$govuk-font-family-gds-transport` to set the font on an eleme
 
 This change was introduced in [pull request #3949: Simplify font family settings](https://github.com/alphagov/govuk-frontend/pull/3949).
 
+#### Removal of Javascript polyfills for Internet Explorer (IE) 8-11
+
+We've removed polyfills for:
+
+- `DOMTokenList`, `Element.prototype.classList`, `Element.prototype.closest`,`Element.prototype.matches`, and `Event`, required for IE 11 and below
+- `Element.prototype.dataset`, required for IE 10 and below
+- `Date.now`, `Document`, `Element`, `Element.prototype.nextElementSibling`, `Element.prototype.previousElementSibling`, `Function.prototype.bind`, `Object.defineProperty`, `String.prototype.trim` and `Window`, required for IE 8
+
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -293,6 +293,25 @@ If you were using `$govuk-font-family-gds-transport` to set the font on an eleme
 
 This change was introduced in [pull request #3949: Simplify font family settings](https://github.com/alphagov/govuk-frontend/pull/3949).
 
+#### Removal of Javascript polyfills for IE8-11
+
+We've removed polyfills for:
+
+- `DOMTokenList`, `Element.prototype.classList`, `Element.prototype.closest`,`Element.prototype.matches`, and `Event`, required for IE 11 and below
+- `Element.prototype.dataset`, required for IE 10 and below
+- `Date.now`, `Document`, `Element`, `Element.prototype.nextElementSibling`, `Element.prototype.previousElementSibling`, `Function.prototype.bind`, `Object.defineProperty`, `String.prototype.trim` and `Window`, required for IE8
```
colinrotherham, line 302: Don't know why, but "IE 8" feels weird with a space, unlike IE 10 and IE 11 🙃

```suggestion
- `Date.now`, `Document`, `Element`, `Element.prototype.nextElementSibling`, `Element.prototype.previousElementSibling`, `Function.prototype.bind`, `Object.defineProperty`, `String.prototype.trim` and `Window`, required for IE 8
```

CHANGELOG.md
```diff
@@ -293,6 +293,25 @@ If you were using `$govuk-font-family-gds-transport` to set the font on an eleme
...
+
+GOV.UK Frontend no longer needs these polyfills because Internet Explorer will no longer try to run our JavaScript, as long as you are including it in your service according to our instructions.
```
colinrotherham, line 304: I might be overly worried, but feels like we should call this out

```suggestion
GOV.UK Frontend no longer needs these polyfills because Internet Explorer will no longer try to run our JavaScript, as long as you are including it in your service according to our instructions using `<script type="module">`.
```

colinrotherham (approved):

colinrotherham (approved):

PR #4054, opened 2023-08-07
Add `rel` attribute to the Exit this Page button

### LINKED ISSUE #3921: Add rel=noreferrer to Exit this page link
## Related component

[Exit this page](https://design-system.service.gov.uk/components/exit-this-page/)

## Context

When exiting the service, the target website by default has knowledge of the referring page within the service. The target website therefore _could_ show a message like "Thank you for visiting us from X", which would undermine the intention of the exit route and potentially put the user at risk.

This can be somewhat mitigated by choosing sensible exit sites, but since we have no control over external website's content their behaviour cannot be guaranteed.

By adding the [rel="noreferrer"](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/noreferrer) attribute to the link, the target website will not be passed the referrer's information by the browser and it will instead behave as if the user has visited the target website directly.

## Alternatives

There is also [rel="noopener"](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/noopener), which prevents setting the `window.opener` property on the target site. rel=noreferrer provides this functionality _and_ strips out the "Referer" header.

## Additional information (if applicable)

- [Original slack thread](https://ukgovernmentdigital.slack.com/archives/C6DMEH5R6/p1688718443379409)
- NB: the `Referer` is knowingly misspelt in the HTML spec and all implementations
- The attribute has [wide support](https://caniuse.com/rel-noreferrer). Where unsupported, it is just ignored and acts as if not provided.

querkmachine:
Adds the `rel="nofollow noreferrer"` attribute to the Exit this Page button, so that the button:
- Does not unduly endorse the external webpage in search rankings.
- Does not provide a referrer header or information about the current browsing context to the external webpage, as this could be used by the target webpage to reveal information about the originating page.

Resolves #3921.

## Changes
- Adds `{ 'rel': 'nofollow noreferrer' }` the `attributes` parameter of the `govukButton` macro.
- Adds tests for the new default functionality.
- Adds changelog entry.
  - Also tweaks the wording of some of the other 'suggested change' entries.

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/exit-this-page/template.njk
+++ b/packages/govuk-frontend/src/govuk/components/exit-this-page/template.njk
@@ -15,6 +15,9 @@
     html: params.html if (params.html or params.text) else defaultHtml,
     text: params.text,
     classes: "govuk-button--warning govuk-exit-this-page__button govuk-js-exit-this-page-button",
-    href: params.redirectUrl | default("https://www.bbc.co.uk/weather")
+    href: params.redirectUrl | default("https://www.bbc.co.uk/weather"),
+    attributes: {
+      rel: "nofollow noreferrer"
+    }
   }) -}}
 </div>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -297,7 +297,7 @@ This change was introduced in [pull request #3949: Simplify font family settings
 
 #### Update the Pagination component's default `aria-label`
 
-The default value of the Pagination component's `aria-label` has been updated to be more descriptive of the contents of the region. If you are using the component's default label, you may wish to update it to the new value.
+The default value of the Pagination component's `aria-label` has been updated to be more descriptive of the contents of the region. If you are using the component's default label, you may wish to update it to the new default of 'Pagination'.
 
 You don't need to change anything if you're using the `govukPagination` Nunjucks macro.
 
@@ -305,18 +305,28 @@ This change was introduced in [pull request #3899: Update default `aria-label` i
 
 #### Update the Exit this Page button's default text
 
-The default text of the Exit this Page button has been updated to indicate that the button is a safety tool and not a generic method of leaving the current page. If you are using the component's default label, you may wish to update it to the new value.
+The default text of the Exit this Page button has been updated. It now includes visually-hidden text to clarify that the button is a safety tool and not a generic method of leaving the current page.
 
-```diff
-<a href="..." role="button" draggable="false" class="govuk-button govuk-button--warning govuk-exit-this-page__button govuk-js-exit-this-page-button" data-module="govuk-button">
--   Exit this page
-+   <span class="govuk-visually-hidden">Emergency</span> Exit this page
-</a>
-```
+If you are using the component's default text, you may wish to update it to the new value: `<span class="govuk-visually-hidden">Emergency</span> Exit this page`
 
 You don't need to change anything if you're using the `govukExitThisPage` Nunjucks macro.
 
 This change was introduced in [pull request #3989: Update default Exit This Page button text](https://github.com/alphagov/govuk-frontend/pull/3989).
+
+#### Add the `rel` attribute to the Exit this Page button and secondary link
+
+Update the Exit this Page button and secondary link to include a new attribute and value: `rel="nofollow noreferrer"`.
+
+Adding this attribute does two things:
+
+1. It instructs search engines that your service does not endorse the external website for the purposes of determining search engine rankings.
+2. It instructs web browsers to not send information about your service to the external website.
+
+This fixes a potential risk where the external website could detect that a user had visited from a GOV.UK page and play that information back to the user, which could risk a user's personal safety in some contexts.
+
+You don't need to change the Exit this Page button if you're using the `govukExitThisPage` Nunjucks macro. You will still have to update the secondary link manually.
+
+This change was introduced in [pull request #4054: Add `rel` attribute to the Exit this Page button](https://github.com/alphagov/govuk-frontend/pull/4054). Thanks to [Greg Tyler](https://github.com/gregtyler) for reporting this issue.
 
 ### Fixes
 
--- a/packages/govuk-frontend-review/src/views/examples/exit-this-page-with-skiplink/index.njk
+++ b/packages/govuk-frontend-review/src/views/examples/exit-this-page-with-skiplink/index.njk
@@ -8,7 +8,10 @@
   {{ govukSkipLink({
     href: "https://www.gov.uk/",
     text: "Exit this page",
-    classes: "govuk-js-exit-this-page-skiplink"
+    classes: "govuk-js-exit-this-page-skiplink",
+    attributes: {
+      rel: "nofollow noreferrer"
+    }
   }) }}
 {% endblock %}
 
--- a/packages/govuk-frontend-review/src/views/full-page-examples/child-maintenance/index.njk
+++ b/packages/govuk-frontend-review/src/views/full-page-examples/child-maintenance/index.njk
@@ -34,7 +34,11 @@ scenario: >-
   {{ super() }}
   {{ govukSkipLink({
     href: "https://bbc.co.uk/weather/",
-    classes: "govuk-js-exit-this-page-skiplink"
+    text: "Exit this page",
+    classes: "govuk-js-exit-this-page-skiplink",
+    attributes: {
+      rel: "nofollow noreferrer"
+    }
   }) }}
 {% endblock %}
 
--- a/packages/govuk-frontend/src/govuk/components/exit-this-page/template.test.js
+++ b/packages/govuk-frontend/src/govuk/components/exit-this-page/template.test.js
@@ -16,6 +16,7 @@ describe('Exit this page', () => {
       expect($button.hasClass('govuk-button--warning')).toBeTruthy()
       expect($button.html()).toContain('<span class="govuk-visually-hidden">Emergency</span> Exit this page')
       expect($button.attr('href')).toBe('/full-page-examples/announcements')
+      expect($button.attr('rel')).toBe('nofollow noreferrer')
     })
   })
 


packages/govuk-frontend/src/govuk/components/exit-this-page/template.njk
```diff
@@ -15,6 +15,7 @@
     html: params.html if (params.html or params.text) else defaultHtml,
     text: params.text,
     classes: "govuk-button--warning govuk-exit-this-page__button govuk-js-exit-this-page-button",
-    href: params.redirectUrl | default("https://www.bbc.co.uk/weather")
+    href: params.redirectUrl | default("https://www.bbc.co.uk/weather"),
+    attributes: params.button.attributes | default({ 'rel': 'nofollow noreferrer' })
```
owenatgov, line 19: These values would be overwritten if a user passes any attributes to the button. I wonder if we should set the value to dynamically include `attributes` if it isn't empty.

owenatgov (commented):
One comment. Looking good.

I wonder if @claireashworth should sus out the changelog entry now or later. Claire, have you been reviewing other changelog entries for 5.0 before they get merged into `main`?

packages/govuk-frontend/src/govuk/components/exit-this-page/template.njk
claireashworth: It's on my list, but not yet.

querkmachine: I mentioned it in my thoughts section, yeah. 

I'm not really sure how to deal with that, though. To my knowledge there isn't any easy way to manipulate, merge or add new keys to objects within out-the-box Nunjuck templating, it requires either a custom filter or global function.

We have custom filters in the review app, but as far as I'm aware we've avoided using custom Nunjucks filters in components as it'd require services to have the same filters configured in their own Nunjucks set up.

packages/govuk-frontend/src/govuk/components/exit-this-page/exit-this-page.yaml
```diff
@@ -23,6 +23,15 @@ params:
...
     description: HTML attributes (for example data attributes) to add to the exit this page container.
+  - name: button
+    type: object
+    required: false
+    description: Options for the exit this page button.
+    params:
+      - name: attributes
+        type: object
+        required: false
+        description: HTML attributes (for example data attributes) to add to the exit this page button. The default disables referrer information being provided to the `redirectUrl` and tells search engines to ignore the page being linked to.
```
colinrotherham, line 34: I feel like we should use `isComponent` but it becomes a worse example of [**Radio and checkbox item nested label components list options that can't be used**](https://trello.com/c/RxuFrKQS/19-radio-and-checkbox-item-nested-label-components-list-options-that-cant-be-used) 🤦‍♂️

```suggestion
        description: HTML attributes (for example data attributes) to add to the exit this page button. The default disables referrer information being provided to the `redirectUrl` and tells search engines to ignore the page being linked to.
        isComponent: true
```

querkmachine: That issue is pretty much why I didn't use it. We're only allowing customisation of a single one of the component's thirteen parameters, so linking off to the full component docs feels iffy.

I'm interested in your opinion on the feasibility of a merge function or filter, as I've mentioned above. That would create the possibility of a user customising every parameter without letting them override the bits required for the component to function. 

For a contrived example:

```nunjucks
{# Our preferred parameter values, but ones the user can override if they wish. #}
{% set defaultParams = {
  redirectUrl: 'https://www.bbc.co.uk/weather',
  button: {
    classes: 'govuk-button--warning',
    attributes: { 'rel': 'nofollow noreferrer' }
  }
} %}

{# The parameter values required for the component to work properly. #}
{% set requiredParams = {
  button: {
    classes: 'govuk-exit-this-page__button govuk-js-exit-this-page-button'
  }
} %}

{{ govukExitThisPage(mergeParams(
  defaultParams,
  params,
  requiredParams
)) }}
```

There's definitely some quirks there that need thought (like how user-defined classes would override `defaultParams` classes but be merged with `requiredParams` classes), but it would seem to make situations like this a bit tidier, perhaps!

packages/govuk-frontend/src/govuk/components/exit-this-page/template.njk
36degrees: > We have custom filters in the review app, but as far as I'm aware we've avoided using custom Nunjucks filters in components as it'd require services to have the same filters configured in their own Nunjucks set up.

Exactly this. We also want to make it possible for someone to recreate the templates with the same 'API' in other templating languages, which may not support object merging (either natively or by some sort of extension). So we try and restrict ourselves to simple logic and looping functionality that's likely to be universally available.

Has anyone requested being able to pass attributes to the button? Wondering if we should do the simplest thing and just hard code the attributes we need, and not provide macro options for them?

```js
attributes: {
  "rel": "nofollow noreferrer"
}
```

packages/govuk-frontend/src/govuk/components/exit-this-page/exit-this-page.yaml
colinrotherham: I'd love a merge function, but yeah probably a filter

We've had [little mentions of merging](https://github.com/alphagov/govuk-frontend/issues/995) for a while [on different issues](https://github.com/alphagov/govuk-frontend/pull/1172#issuecomment-460257465)

Every service team seems to maintain their own to merge POST form data etc:

```js
{{ govukInput(form.context.email | govukParams({
  type: "email",
  autocomplete: "username",
  spellcheck: false
})) }}
```

packages/govuk-frontend/src/govuk/components/exit-this-page/template.njk
querkmachine: It's not been a request from a user, no, but the lack of user-definable attributes was noted as a barrier to users patching the issue manually—it wouldn't have required a new code release to fix and we could have issued guidance in the interim.

Making the attribute configurable is a bit of pre-emptive futureproofing in that way, but I suppose there is no requirement we actually do it if we don't think it'll be useful.

packages/govuk-frontend/src/govuk/components/exit-this-page/exit-this-page.yaml
owenatgov: I'm quite adamant that we shouldn't use `isComponent` here. A huge part of ETP's design success is specifically how we use the button component ie: it's a link that looks like a button and it's red. Allowing users to change that introduces risk. Putting my foot down on this suggestion not getting included.

packages/govuk-frontend/src/govuk/components/exit-this-page/template.njk
owenatgov: It was my suggestion off of the linked issue to include attributes. As we're finding out, this might not be feasible. I've been running this over for a day and can't come up with a way to get this to work within our constraints. I'm leaning towards hard coding the attributes as well.

The one hypothetical use case for allowing attributes is click tracking, which would be useful for us in terms of gathering usage within services. I don't think that's a show stopper right now until we solve this thorny issue of how to manipulate existing objects though. On top of that, nobody's asked for it yet.

I can see there's some merge discourse going on in the thread below.

packages/govuk-frontend/src/govuk/components/exit-this-page/exit-this-page.yaml
colinrotherham: Foot down, agreed, suggestion wasn't to enable button options.

Rather, we have the `isComponent` flag to enable the "this is a component option" UI

<br>
Button nested option

packages/govuk-frontend/src/govuk/components/exit-this-page/template.njk
querkmachine: I've opted to hardcode the attributes on the basis that it leaves room for us to make them configurable at a later date without introducing a breaking change. If we committed to one merge/override strategy now, and changed it later down the line, that would be much more troublesome.

packages/govuk-frontend/src/govuk/components/exit-this-page/exit-this-page.yaml
querkmachine: I've opted to hardcode the attribute for now, so this discussion is moot — or one for later, at least.

colinrotherham: Yeah definitely, would you mind expanding the [Trello card](https://trello.com/c/RxuFrKQS/19-radio-and-checkbox-item-nested-label-components-list-options-that-cant-be-used) to include Exit this Page buttons too?

Or adding a new one for nested components that we choose not to identify

packages/govuk-frontend/src/govuk/components/exit-this-page/template.njk
```diff
@@ -15,6 +15,9 @@
     html: params.html if (params.html or params.text) else defaultHtml,
     text: params.text,
     classes: "govuk-button--warning govuk-exit-this-page__button govuk-js-exit-this-page-button",
-    href: params.redirectUrl | default("https://www.bbc.co.uk/weather")
+    href: params.redirectUrl | default("https://www.bbc.co.uk/weather"),
+    attributes: {
+      "rel": "nofollow noreferrer"
```
colinrotherham, line 20:
```suggestion
      rel: "nofollow noreferrer"
```

colinrotherham (approved):
One tiny minor thing

Although we often quote `attributes` keys like `"data-attribute"` so not a blocker

packages/govuk-frontend/src/govuk/components/exit-this-page/exit-this-page.yaml
owenatgov: Aha gotcha. I'm on the fence about this as it's not really the component API we're exposing, just the attributes bit.

As it happens, it looks like this might not be relevant anymore from the above chat.

owenatgov (commented):
We're looking good. One final thing I forgot in my previous review: could we also update the child maintenance and ETP with skiplink examples so that the secondary links on those pages include the `rel` improvement?

packages/govuk-frontend/src/govuk/components/exit-this-page/exit-this-page.yaml
colinrotherham: Yeah no worries. What we really need is a way to document:

1. The component has a nested component option
2. The nested component option can configure all/some/one params
3. The nested component links to its page or shows options in a table (like `label` and `hint`)

But not for this PR

querkmachine:
@owenatgov It is done! 

One of the skiplinks was missing any text, too. Fixed that while I was at it. Ooerr.

owenatgov (approved):
💪🏻

packages/govuk-frontend/src/govuk/components/exit-this-page/exit-this-page.yaml
owenatgov: I agree. I am far from against finding a way to communicate "we are using one of our components here and are letting you manipulate XYZ part of it".

Issue #3895, opened 2023-07-03
New typography scale: update footer link font size to 19


owenatgov:
## What
Update the font size of links in the footer component to be 19 instead of 16.

Should be merged into https://github.com/alphagov/govuk-frontend/pull/2421 and not `main`.

## Why
[During testing of the govuk-frontend components against the new typography scale](https://github.com/alphagov/govuk-design-system/issues/2289), it was queried why the footer needed to use 16 font size for its links.

## Who needs to work on this
Developers

## Who needs to review this
Developers, designers

PR #4064, opened 2023-08-08
Update footer font size from 16 to 19

owenatgov:
Updates footer link font size to be 19 on the type scale instead of 16.

Fixes https://github.com/alphagov/govuk-frontend/issues/3895

## Visual changes
| Screen size | Before | After |
| - | - | - |
| desktop | Screenshot of footer on desktop before font change |  Screenshot of footer on desktop after font change |
| mobile | Screenshot of footer on mobile before font change | Screenshot of footer on mobile after font change |

## Thoughts
Could this change be merged directly into `main` instead of into `type-scale`? The risk there is that it could count as a "significant" visual change, although in theory nobody would need to change their code in response.


christopherthomasdesign (approved):
Looks good to me 👍 

Especially helps with footers that have lots of links and sections.

The thing you lose is a bit of typographic hierarchy on the page, I reckon the footer links are smaller atm because they're less important than the main body text above. But I think the other stylistic elements like the border and grey background emphasise that enough. Be interested to see what our GOV.‌UK colleagues think, they're the service with by far the biggest footer.

Issue #3896, opened 2023-07-03
Review position of the menu button in the header


owenatgov:
## What
Adjust the position of the menu button in the header component on small screens.

~Should be merged into https://github.com/alphagov/govuk-frontend/pull/2421 and not `main`.~

## Why
[During testing of the govuk-frontend components against the new typography scale](https://github.com/alphagov/govuk-design-system/issues/2289), we noticed that the menu button now looked too low as it was inline with the baseline govuk logo.

## Who needs to work on this
Developers, designers

## Who needs to review this
Developers, designers

owenatgov:
From starting this work and doing some light investigation, the recommended approach is to refactor the header component to use flexbox to manage it's positioning rather than manually adjust the position of the button.

PR #4065, opened 2023-08-09
Adjust the positioning of the header mobile menu button

owenatgov:
Adjust the positioning of the header mobile menu button so that it sits in the middle of the header bar on mobile rather than at the baseline of the GOV.UK logo.

Fixes https://github.com/alphagov/govuk-frontend/issues/3896

### Before
Screenshot of header on mobile before change

### After
Screenshot of header on mobile after change

## Thoughts
I'm wondering if there's a way to flexbox the header so that we can make this positioning dynamic. Will explore this as part of this PR.

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/header/_index.scss
+++ b/packages/govuk-frontend/src/govuk/components/header/_index.scss
@@ -166,7 +166,7 @@
   .govuk-header__menu-button {
     @include govuk-font($size: 16);
     position: absolute;
-    top: govuk-spacing(4);
+    top: govuk-spacing(3);
     right: 0;
     margin: 0;
     padding: 0;


querkmachine (approved):
LGTM.

Unrelated to this, but I noticed when reviewing that there is 50 pixels of horizontal space reserved for the absolutely positioned toggle, but the toggle itself is wider than this (≈54 pixels). 

Flagging as it might need to be accounted for if the toggle text is going to be scaled up at some point.

owenatgov:
@querkmachine Good spot. I'm gonna investigate it a little more but I expect this can be solved by just increasing that padding to `govuk-spacing(9)` rather than `8`. I wonder if there's a way we can make the positioning make more sense and avoid then eed for absolute positioning hacks at all.

PR #4067, opened 2023-08-09
Catch errors when instantiating components with `initAll`

### LINKED ISSUE #4010: Prevent `initAll` from failing when one of the components' instantiation errors
## What

Prevent `initAll()` from failing at the first component that may error during `instantiation` and let it carry on to initialise all the other components. 

For components that do error, `initAll` should log the error to the console so users know something went wrong.

As [components would throw if GOV.UK Frontend is not supported](https://github.com/alphagov/govuk-frontend/issues/4009), we'll want to not clutter the logs with an error for each components. Instead, we'll want to abort early and only log once that GOV.UK Frontend is not supported.

## Why

With components [throwing errors](https://github.com/alphagov/govuk-frontend/issues/4009) when instantiated over invalid markup, we want to keep `initAll()` as a quick "fire and forget" way to initialise all components on the page.

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [ ] We've agreed on an implementation (among [those offered in comment](https://github.com/alphagov/govuk-frontend/issues/4010#issuecomment-1650193531) or new propositions)
- [ ] `initAll()` logs any error happening during components' initialisation and initialises properly components that do not error.

36degrees:
Initialise all of the components using a loop in initAll, catching any errors thrown by the components.

Closes #4010

### DIFF
--- a/packages/govuk-frontend/src/govuk/all.mjs
+++ b/packages/govuk-frontend/src/govuk/all.mjs
@@ -29,76 +29,40 @@ function initAll(config) {
     return
   }
 
+  const components = [
+    Accordion,
+    Button,
+    CharacterCount,
+    Checkboxes,
+    ErrorSummary,
+    ExitThisPage,
+    Header,
+    NotificationBanner,
+    Radios,
+    SkipLink,
+    Tabs
+  ]
+
   // Allow the user to initialise GOV.UK Frontend in only certain sections of the page
   // Defaults to the entire document if nothing is set.
   const $scope = config.scope instanceof HTMLElement ? config.scope : document
 
-  const $accordions = $scope.querySelectorAll('[data-module="govuk-accordion"]')
-  $accordions.forEach(($accordion) => {
-    new Accordion($accordion, config.accordion)
-  })
+  components.forEach((Component) => {
+    // configKey is the component name converted from PascalCase to camelCase
+    const configKey = Component.name.replace(/^./, (str) => str.toLowerCase())
 
-  const $buttons = $scope.querySelectorAll('[data-module="govuk-button"]')
-  $buttons.forEach(($button) => {
-    new Button($button, config.button)
-  })
+    const $elements = $scope.querySelectorAll(
+      `[data-module="${Component.moduleName}"]`
+    )
 
-  const $characterCounts = $scope.querySelectorAll(
-    '[data-module="govuk-character-count"]'
-  )
-  $characterCounts.forEach(($characterCount) => {
-    new CharacterCount($characterCount, config.characterCount)
-  })
-
-  const $checkboxes = $scope.querySelectorAll(
-    '[data-module="govuk-checkboxes"]'
-  )
-  $checkboxes.forEach(($checkbox) => {
-    new Checkboxes($checkbox)
-  })
-
-  // Find first error summary module to enhance.
-  const $errorSummary = $scope.querySelector(
-    '[data-module="govuk-error-summary"]'
-  )
-  if ($errorSummary) {
-    new ErrorSummary($errorSummary, config.errorSummary)
-  }
-
-  const $exitThisPageButtons = $scope.querySelectorAll(
-    '[data-module="govuk-exit-this-page"]'
-  )
-  $exitThisPageButtons.forEach(($button) => {
-    new ExitThisPage($button, config.exitThisPage)
-  })
-
-  // Find first header module to enhance.
-  const $header = $scope.querySelector('[data-module="govuk-header"]')
-  if ($header) {
-    new Header($header)
-  }
-
-  const $notificationBanners = $scope.querySelectorAll(
-    '[data-module="govuk-notification-banner"]'
-  )
-  $notificationBanners.forEach(($notificationBanner) => {
-    new NotificationBanner($notificationBanner, config.notificationBanner)
-  })
-
-  const $radios = $scope.querySelectorAll('[data-module="govuk-radios"]')
-  $radios.forEach(($radio) => {
-    new Radios($radio)
-  })
-
-  // Find first skip link module to enhance.
-  const $skipLink = $scope.querySelector('[data-module="govuk-skip-link"]')
-  if ($skipLink) {
-    new SkipLink($skipLink)
-  }
-
-  const $tabs = $scope.querySelectorAll('[data-module="govuk-tabs"]')
-  $tabs.forEach(($tabs) => {
-    new Tabs($tabs)
+    $elements.forEach(($element) => {
+      // Only pass config to components that accept it
+      if ('defaults' in Component) {
+        new Component($element, configKey in config ? config[configKey] : {})
+      } else {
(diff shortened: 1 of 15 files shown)


packages/govuk-frontend/src/govuk/all.mjs
```diff
@@ -29,76 +29,39 @@ function initAll(config) {
...
-  // Find first skip link module to enhance.
-  const $skipLink = $scope.querySelector('[data-module="govuk-skip-link"]')
-  if ($skipLink) {
-    new SkipLink($skipLink)
-  }
-
-  const $tabs = $scope.querySelectorAll('[data-module="govuk-tabs"]')
-  $tabs.forEach(($tabs) => {
-    new Tabs($tabs)
+  components.forEach(([Component, config]) => {
```
colinrotherham, line 50: Need to double check browser support for this line

romaricpascal: Isn't that the kind of things that Babel abstracts for us? It seems a transform that's pretty straightforward for it anyway.

packages/govuk-frontend/src/govuk/all.mjs
```diff
@@ -29,76 +29,39 @@ function initAll(config) {
...
+    const $elements = $scope.querySelectorAll(
+      `[data-module="${Component.moduleName}"]`
+    )
+
+    $elements.forEach(($element) => {
+      try {
+        // Only pass config to components that accept it
+        'defaults' in Component
```
romaricpascal, line 58: **suggestion** This line should probably rely on the presence of `config` now it's here rather than some internals of `Component`?

colinrotherham: Give it a try if you like? You'll see that it's flagged `@private` 😉

But every component with static `defaults` _also_ has a config

packages/govuk-frontend/src/govuk/all.mjs
colinrotherham: Sometimes a polyfill is needed, but we haven't enabled them. Just need to double check

packages/govuk-frontend/src/govuk/all.mjs
36degrees: I think @romaricpascal was suggesting checking whether the local `config` variable was defined, rather than looking at `Component.config`?

I've tried switching this out for 

```js
typeof config !== 'undefined'
```

But the TypeScript compiler doesn't like it:

```
../govuk-frontend/src/govuk/all.mjs:60:37 - error TS2554: Expected 1 arguments, but got 2.

60           ? new Component($element, config)
                                       ~~~~~~
```

colinrotherham: Yeah, that too

Adding another local config to `components` doesn't mean `Component.config` exists

But for `'defaults' in Component` the compiler knows "all components with defaults also have a config"

colinrotherham: Otherwise we could change that to `'config' in Component` but we'd need to remove `@private`

romaricpascal: > I think @romaricpascal was suggesting checking whether the local config variable was defined, rather than looking at Component.config?

Yup, that's what I was suggesting, to rely on something local rather than mix responsibilities for `defaults` (providing a default configuration *and* being a marker of the availability of a config from the array of components)  😊

colinrotherham: Yeah, would be nice

Sadly it's _our_ marker and doesn't tell the compiler anything about the component

E.g. We could add 3x fake configs but doesn't mean they exist on the component

```js
[CharacterCount, config.fake123],
[Checkboxes, config.fake123],
[ErrorSummary, config.fake123],
```

packages/govuk-frontend/src/govuk/all.unit.test.mjs
```diff
@@ -0,0 +1,132 @@
...
+      .spyOn(global.console, 'log')
+      .mockImplementation(() => {
+        /* noop */
+      })
+
+    expect(() => {
+      GOVUKFrontend.initAll()
+    }).not.toThrow()
+
+    expect(consoleLog).toHaveBeenCalledWith(
```
colinrotherham, line 126: Suppose we want to check an actual `Error` instance has been logged

```suggestion
    expect(global.console.log).toHaveBeenCalledWith(expect.any(Error))
    expect(global.console.log).toHaveBeenCalledWith(
```

colinrotherham (commented):
Looks brill! Nothing particularly blocking, mainly Jest tips etc

36degrees:
@colinrotherham I think I've addressed all your comments – can you take another look?

colinrotherham:
@36degrees Nice, thanks. Looking now

shared/lib/names.js
```diff
@@ -20,6 +20,25 @@ function kebabCaseToPascalCase(value) {
...
+  return (
+    value
+      .toLowerCase()
+      .split('-')
+      // capitalize each 'word'
+      .map((word, index) => {
+        return index > 0 ? word.charAt(0).toUpperCase() + word.slice(1) : word
+      })
+      .join('')
+  )
```
colinrotherham, line 39: Do you mind me saying I liked your original idea?

Where you just lowercased the first **PascalCase** letter back into **camelCase**

```suggestion
  return kebabCaseToPascalCase(value).replace(/^./, (str) =>
    str.toLowerCase()
  )
```

packages/govuk-frontend/src/govuk/components/exit-this-page/exit-this-page.mjs
```diff
@@ -391,6 +391,13 @@ export class ExitThisPage extends GOVUKFrontendComponent {
     }
   }
 
+  /**
+   * Name for the component used when initialising using data-module attributes.
+   *
+   * @type {string}
```
colinrotherham, line 397: Only other suggestion, but optional, would be removing these types

Where default values are obvious to the compiler already, don't need them

```suggestion
```

colinrotherham (approved):
Brilliant to see all this new coverage 🎉

One suggestion about unnecessary types

E.g. We don't typically type strings with default values:

```js
  /** @private */
  controlsClass = 'govuk-accordion__controls'

  /** @private */
  showAllClass = 'govuk-accordion__show-all'

  /** @private */
  showAllTextClass = 'govuk-accordion__show-all-text'
```

But have a look at https://github.com/alphagov/govuk-frontend/pull/4067#discussion_r1295850711. Otherwise approved ✅

packages/govuk-frontend/src/govuk/all.mjs
36degrees: I think I prefer the `defaults` check as it's that much simpler. To avoid blocking the PR whilst @romaricpascal is off, going to merge this for now but happy to revisit later.

PR #4069, opened 2023-08-10
Update TypeScript paths to minimise included files

colinrotherham:
TypeScript config related changes to minimise the number of included files

Should improve the speed of `tsc` and editor autocomplete etc

### DIFF
--- a/.github/workflows/scripts/tsconfig.json
+++ b/.github/workflows/scripts/tsconfig.json
@@ -2,6 +2,7 @@
   "extends": "../../../tsconfig.base.json",
   "include": ["**/*.mjs"],
   "compilerOptions": {
-    "strict": true
+    "strict": true,
+    "types": ["@actions/github", "@octokit/rest", "node"]
   }
 }
--- a/tsconfig.base.json
+++ b/tsconfig.base.json
@@ -6,6 +6,7 @@
     "module": "ESNext",
     "moduleResolution": "Node",
     "noEmit": true,
+    "noImplicitThis": true,
     "resolveJsonModule": true,
     "skipLibCheck": true,
     "strict": false,
--- a/tsconfig.base.json
+++ b/tsconfig.base.json
@@ -14,7 +14,14 @@
     "target": "ESNext",
     "types": ["jest", "jest-axe", "jest-puppeteer", "node"],
     "paths": {
-      "govuk-frontend": ["./packages/govuk-frontend/src/govuk/all.mjs"]
+      "govuk-frontend": ["./packages/govuk-frontend/src/govuk/all.mjs"],
+      "govuk-frontend-stats": ["./shared/stats/src/index.mjs"],
+      "govuk-frontend-config": ["./shared/config/index.js"],
+      "govuk-frontend-helpers": ["./shared/helpers/index.js"],
+      "govuk-frontend-helpers/*": ["./shared/helpers/*.js"],
+      "govuk-frontend-helpers/jest/*": ["./shared/helpers/jest/*"],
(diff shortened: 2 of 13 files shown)


tsconfig.json
```diff
@@ -3,28 +3,28 @@
   "files": [],
   "references": [
     {
-      "path": "./.github/workflows/scripts"
+      "path": "./.github/workflows/scripts/tsconfig.json"
```
domoscargin, line 6: Did this work before? Absolutely good to be specific here, just curious if there's any other reason for doing so!

tsconfig.base.json
```diff
@@ -6,6 +6,7 @@
     "module": "ESNext",
     "moduleResolution": "Node",
     "noEmit": true,
+    "noImplicitThis": true,
```
domoscargin, line 9: 👍

.github/workflows/scripts/tsconfig.json
```diff
@@ -2,6 +2,7 @@
   "extends": "../../../tsconfig.base.json",
   "include": ["**/*.mjs"],
   "compilerOptions": {
-    "strict": true
+    "strict": true,
+    "types": ["@actions/github", "@octokit/rest", "node"]
```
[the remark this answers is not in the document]
colinrotherham: Don't have to do this, it's handy to let TypeScript know exactly where the `jest-axe` extensions can be used etc

Let me know which commits might go too far

tsconfig.json
colinrotherham: It did, but a recent Visual Studio update broke CMD + click on the paths 🤷‍♂️

.github/workflows/scripts/tsconfig.json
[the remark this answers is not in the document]
domoscargin: I don't think this is too far, but maybe we need to start documenting some of the specific typing stuff we do on the repo. Whether in the team docs or in the dev docs here somewhere.

I'm thinking things like "Adding new types". But maybe _that's_ too far and this is pretty standard TypeScript stuff so we should just refer to TypeScript docs?

domoscargin (approved):
Looks good!

.github/workflows/scripts/tsconfig.json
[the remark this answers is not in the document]
colinrotherham: We don't turn the `noImplicitAny` option on so almost all types are optional

Forget one? You'll miss out on code autocomplete and type checking, but I'll still work

Some types though like `"jest"` add globals `expect()`, `describe()`, `it()` etc

You'll see compiler errors for those if you forget to add the types

tsconfig.base.json
```diff
@@ -6,15 +6,26 @@
...
+      "govuk-frontend-config": ["./shared/config/index.js"],
+      "govuk-frontend-helpers": ["./shared/helpers/index.js"],
+      "govuk-frontend-helpers/*": ["./shared/helpers/*.js"],
+      "govuk-frontend-helpers/jest/*": ["./shared/helpers/jest/*"],
+      "govuk-frontend-lib": ["./shared/lib/index.js"],
+      "govuk-frontend-lib/*": ["./shared/lib/*.js"]
     }
+  },
+  "typeAcquisition": {
+    "enable": false
```
colinrotherham, line 29: Added one extra flag

This prevents editors from automatically downloading `@types/*` packages to speed up first startup time. We already declare all the types we'd like anyway

It's highlighted that a few packages are missing, but other than `@types/js-yaml` they're out of date and cause issues with our code accessing things that aren't typed

```console
@types/js-yaml
@types/js-beautify
@types/shuffle-seed
```

PR #4075, opened 2023-08-11
Add example demonstrating the `SupportError` in the browser

romaricpascal:
Little follow up on https://github.com/alphagov/govuk-frontend/pull/4030 to add [an example for JavaScript errors to the review app](https://govuk-frontend-pr-4075.herokuapp.com/examples/javascript-errors). As we add more errors, we may reshape this example to show multiple kinds of errors or add extra examples for the other kinds of errors.

The example uses a CharacterCount component to trigger the errors as the component should be able to emit all the types of errors we raise, especially the one for invalid configuration that it is the only one to throw.

Screenshot of the JavaScript errors page showing the SupportError and its stacktrace

> **Note**: The name of the error appears clearly in the logs. Its class name is mangled in the stacktrace (`n`), but I don't think that's a massive issue given:
> 
> - the name of the error is clearly apparent
> - the component raising the error is clear (`Accordion`)
> - sourcemaps show the full sources
>
> I've opened a [separate issue to discuss mangling](https://github.com/alphagov/govuk-frontend/issues/4076) as I think it's worth its own little exploration to keep things maintainable as we add new classes, helpers...

### DIFF
--- a/packages/govuk-frontend-review/src/views/examples/javascript-errors/index.njk
+++ b/packages/govuk-frontend-review/src/views/examples/javascript-errors/index.njk
@@ -1,0 +1,55 @@
+{% extends "layouts/layout.njk" %}
+
+{% from "govuk/components/back-link/macro.njk" import govukBackLink %}
+{% from "govuk/components/character-count/macro.njk" import govukCharacterCount %}
+
+{% block beforeContent %}
+  {{ govukBackLink({
+    href: "/"
+  }) }}
+{% endblock %}
+
+{% block content %}
+  <h1 class="govuk-heading-l">JavaScript errors</h1>
+  <p class="govuk-body-lead">
+    Open your <a class="govuk-link" href="https://developer.mozilla.org/en-US/docs/Web/API/console#see_also">brower's console</a>
+    to witness the error thrown in browsers where GOV.UK Frontend is not supported.
+  </p>
+  {{ govukCharacterCount({
+    label: {
+      text: "Any other comment"
+    },
+    id: "some-content",
+    name: "some-content"
+  })}}
+{% endblock %}
+
+{% block bodyEnd %}
+  <script type="module" src="/javascripts/govuk-frontend.min.js"></script>
+
+  <!-- SupportError -->
+  <script type="module">
+    // Calling `initAll` would result in errors being logged rather than thrown
+    import { CharacterCount } from '/javascripts/govuk-frontend.min.js'
+
+    // Simulate GOV.UK Frontend not being supported
+    document.body.classList.remove('govuk-frontend-supported')
+
+    const $element = document.querySelector('[data-module="govuk-character-count"]')
+
+    try {
+      // Instantiate the component directly so errors are thrown
+      new CharacterCount($element)
+      // Use `finally` to ensure we tidy up regardless of the error
+    } finally {
+      // Add back the class marking GOV.UK Frontend support for future examples
+      document.body.classList.add('govuk-frontend-supported')
+    }
+  </script>
+
+  <!--
+    Add examples of other types of errors in their own `<script>` tag.
+    This separates each error example, as well as isolate their code
+    from the errors thrown by the other `<script>` tags.
+  -->
+{% endblock %}


packages/govuk-frontend-review/src/views/examples/js-support-error/index.njk
```diff
@@ -0,0 +1,34 @@
...
+    href: "/"
+  }) }}
+{% endblock %}
+
+{% block content %}
+  <h1 class="govuk-heading-l">Example: JavaScript error - Support Error</h1>
+  <p class="govuk-body-lead">
+      Open your <a href="https://developer.mozilla.org/en-US/docs/Web/API/console#see_also">brower's console</a>
+      to witness the error thrown in browsers where GOV.UK Frontend is not supported.
+    </p>
```
colinrotherham, line 17:
```suggestion
  <p class="govuk-body-lead">
    Open your <a href="https://developer.mozilla.org/en-US/docs/Web/API/console#see_also">brower's console</a>
    to witness the error thrown in browsers where GOV.UK Frontend is not supported.
  </p>
```

packages/govuk-frontend-review/src/views/examples/js-support-error/index.njk
```diff
@@ -0,0 +1,34 @@
...
+{% from "govuk/components/accordion/macro.njk" import govukAccordion %}
+
+{% block beforeContent %}
+  {{ govukBackLink({
+    href: "/"
+  }) }}
+{% endblock %}
+
+{% block content %}
+  <h1 class="govuk-heading-l">Example: JavaScript error - Support Error</h1>
```
colinrotherham, line 13: Probably good to match the example link text to the heading?

```suggestion
  <h1 class="govuk-heading-l">Example: JavaScript support error</h1>
```

packages/govuk-frontend-review/src/views/examples/js-support-error/index.njk
```diff
@@ -0,0 +1,34 @@
...
+  {{ govukAccordion()}}
+{% endblock %}
+
+{% block bodyEnd %}
+  <!-- Simulate GOV.UK Frontend not being supported -->
+  <script>
+    document.body.classList.remove('govuk-frontend-supported')
+  </script>
+  <script type="module" src="/javascripts/govuk-frontend.min.js"></script>
+  <!-- We don't want to use `initAll` as it returns early if govuk-frontend is not supported -->
```
colinrotherham, line 27: Good thinking. Mind if we move this into a JavaScript comment?

packages/govuk-frontend-review/src/views/examples/js-support-error/index.njk
```diff
@@ -0,0 +1,34 @@
...
+  <script type="module">
+    import { Accordion } from '/javascripts/govuk-frontend.min.js'
+
+    const $element = document.querySelector('[data-module="govuk-accordion"]');
```
colinrotherham, line 31: Probably don't need `$element` as it's `null` anyway

The support checks run before the "instanceof" check

romaricpascal: That's technically what happens, but I'd rather keep the initialisation code as close as possible to what may happen in actual code for clarity.

colinrotherham: I think I'd wonder where the Accordion was?

packages/govuk-frontend-review/src/views/examples/javascript-errors/index.njk
```diff
@@ -0,0 +1,40 @@
...
+
+  <script type="module" src="/javascripts/govuk-frontend.min.js"></script>
+
+  <script type="module">
+    import { Accordion } from '/javascripts/govuk-frontend.min.js'
+    // Simulate GOV.UK Frontend not being supported
+    document.body.classList.remove('govuk-frontend-supported')
+
+    // We don't want to use `initAll` as it returns early if govuk-frontend is not supported
+    const $element = document.querySelector('[data-module="govuk-character-count"]')
```
colinrotherham, line 37: Mind if we clarify that "return early" actually "will catch and log errors" in https://github.com/alphagov/govuk-frontend/pull/4067?

```suggestion

    // Simulate element not found
    $element = null

    // Enable `SupportError` by removing GOV.UK Frontend support
    document.body.classList.remove('govuk-frontend-supported')
    
    // Throw `SupportError` by instantiating component directly
    // since calling `initAll()` will catch and log errors only
```

packages/govuk-frontend-review/src/views/examples/javascript-errors/index.njk
```diff
@@ -0,0 +1,40 @@
...
+
+{% block content %}
+  <h1 class="govuk-heading-l">JavaScript errors</h1>
+  <p class="govuk-body-lead">
+    Open your <a class="govuk-link" href="https://developer.mozilla.org/en-US/docs/Web/API/console#see_also">brower's console</a>
+    to witness the error thrown in browsers where GOV.UK Frontend is not supported.
+  </p>
+  {{ govukCharacterCount({
+    label: {
+      text: 'Any other comment'
```
colinrotherham, line 20: We don't have Prettier for Nunjucks, but would you mind using double quotes `"` for strings?

packages/govuk-frontend-review/src/views/examples/javascript-errors/index.njk
36degrees: At the minute, we still have an early return in #4067 which I _think_ is what we want…?

https://github.com/alphagov/govuk-frontend/blob/e6a32b64550eef033600b4229296ad8c66ddbb49/packages/govuk-frontend/src/govuk/all.mjs#L27-L30

colinrotherham: Ahh here's me thinking we had a helper for this, but `initAll()` caught + logged it 🤦‍♂️

```js
function checkSupport() {
  if (!isSupported()) {
    throw new SupportError()
  }
}
```

Perhaps we do want `checkSupport()` as a helper so we can log something useful for `initAll()` too?

```js
// Skip initialisation when GOV.UK Frontend is not supported
try {
  checkSupport()
} catch (error) {
  console.error(error)
  return
}
``

Comment (romaricpascal) [reply packages/govuk-frontend-review/src/views/examples/javascript-errors/index.njk:37]: I [don't think we need to make that early return be triggered by an error being thrown](https://github.com/alphagov/govuk-frontend/pull/4117#issuecomment-1686057627), so I'd keep the comment and `initAll`'s behaviour as is.

**Comment by romaricpascal** [comment]
Looks like I need to rebase onto main

File: packages/govuk-frontend-review/src/views/examples/javascript-errors/index.njk (hunk 5 above)
Comment (colinrotherham) [reply packages/govuk-frontend-review/src/views/examples/javascript-errors/index.njk:37]: That's nicely cleared up now in https://github.com/alphagov/govuk-frontend/pull/4117

Other lines clarified that it's the class removal that makes `SupportError` possible

But I've used `$element = null` as no Character Count exists on the page, intentionally

```suggestion
    // Simulate element not found
    $element = null

    // Enable `SupportError` by removing GOV.UK Frontend support
    document.body.classList.remove('govuk-frontend-supported')
    
    // Throw `SupportError` by instantiating component directly
    // since calling `initAll()` logs only without throwing
```

packages/govuk-frontend-review/src/views/examples/javascript-errors/index.njk
```diff
@@ -0,0 +1,40 @@
...
+  </p>
+  {{ govukCharacterCount({
+    label: {
+      text: "Any other comment"
+    },
+    id: "some-content",
+    name: "some-content"
+  })}}
+{% endblock %}
+
```
colinrotherham, line 26:
```suggestion
```

packages/govuk-frontend-review/src/views/examples/javascript-errors/index.njk
```diff
@@ -0,0 +1,40 @@
...
+{% block bodyEnd %}
+
```
colinrotherham, line 28:
```suggestion
```

packages/govuk-frontend-review/src/views/examples/javascript-errors/index.njk
```diff
@@ -0,0 +1,40 @@
...
+  <script type="module" src="/javascripts/govuk-frontend.min.js"></script>
+
```
colinrotherham, line 30:
```suggestion
```

packages/govuk-frontend-review/src/views/examples/js-support-error/index.njk
colinrotherham: We're still importing `Accordion` and running `new Accordion($element)`?

I'd prefer we used `$element = null` until we expand this page with a component

See https://github.com/alphagov/govuk-frontend/pull/4075#discussion_r1308667562

packages/govuk-frontend-review/src/views/examples/javascript-errors/index.njk
romaricpascal: Update on its way to describe adding new examples in separate script tags.

packages/govuk-frontend-review/src/views/examples/js-support-error/index.njk
colinrotherham: Thanks for fixing this

packages/govuk-frontend-review/src/views/examples/javascript-errors/index.njk
colinrotherham: Thanks. The line below this one was removed instead

packages/govuk-frontend-review/src/views/examples/javascript-errors/index.njk
```diff
@@ -0,0 +1,55 @@
...
+    // Simulate GOV.UK Frontend not being supported
+    document.body.classList.remove('govuk-frontend-supported')
+
+    const $element = document.querySelector('[data-module="govuk-character-count"]')
+
+    try {
+      // Instantiate the component directly so errors are thrown
+      new CharacterCount($element)
+      // Use `finally` to ensure we tidy up regardless of the error
+    } finally {
```
colinrotherham, line 44: Do we have try/finally support in our supported browsers?

Or even try-without-catch for future browser testing?

romaricpascal: I'm not aware of any support issue with `try` blocks when running synchronous code. It's a feature that's been there [since pretty much the start](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch). [Using it without `catch` was there as early as ES5 in the specs as well](https://262.ecma-international.org/5.1/#sec-12.14) (maybe even earlier, but I did not find specs for ES3 quickly)

It's a different matter for asynchronous code, though, as [`Promise.finally` landed more recently](https://caniuse.com/?search=finally), so that's where we'll need to be a bit more cautious. Regarding `catch`, the more recent development is the ability to omit grabbing the `(error)`, for when you don't want to do anything with it (the ["Optional catch binding" in the MDN support table](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch#browser_compatibility)), is that what you refer to by "try-without-catch"?

colinrotherham: >…is that what you refer to by "try-without-catch"?

Yeah that's the one https://caniuse.com/mdn-javascript_statements_try_catch_optional_catch_binding

Got a bit of a gap from [`<script type="module">`](https://caniuse.com/es6-module) to [`try...catch`: Optional catch binding](https://caniuse.com/mdn-javascript_statements_try_catch_optional_catch_binding)

Best adding an empty catch and running the tidy up below

romaricpascal: Lack of a `catch` clause has been well supported since ES5, so we're good here. The optional catch binding is about omitting the "argument" after `catch`. Here's a screenshot of the page in Chrome 61 (that doesn't support Optional catch binding):
- the error is logged in the console (showing the component was instantiated with the `govuk-frontend-supported` class removed)
- the `govuk-frontend-supported` class is on the `<body>` (after having been added back in the `finally` block)

colinrotherham: That's good enough for me. Thanks for testing it @romaricpascal

colinrotherham (approved):

PR #4089, opened 2023-08-15
Add type discovery to `src` symlink directories

colinrotherham:
We currently don't show TypeScript `tsc` compiler issues for our previous source code symbolic links:

* `src/govuk` → `packages/govuk-frontend/src/govuk`
* `src/govuk-prototype-kit` → `packages/govuk-frontend/src/govuk-prototype-kit`

This PR adds a new **tsconfig.json** file to fix that

But to avoid running our checks twice we can exclude it from `references` by default:

https://github.com/alphagov/govuk-frontend/blob/f22b592608332549b8587a602e577b6fa43ca8b4/tsconfig.json#L4

### DIFF
--- a/src/tsconfig.json
+++ b/src/tsconfig.json
@@ -1,0 +1,4 @@
+{
+  "extends": "../packages/govuk-frontend/tsconfig.json",
+  "include": ["**/*.js", "**/*.mjs"]
+}


colinrotherham:
Alternatively we may want to delete the symbolic links entirely but see https://github.com/alphagov/govuk-frontend/pull/3491#pullrequestreview-1404367423

>* we should add a way, if only for a while, to direct people to the new places, as they're used to looking at the sources in `src` and the app in `app`. We mentioned README.md files in the old locations, which sound an easy way to sort this out, or symlinks. Does any have particular drawbacks (maybe in terms of rebasing)? I think these pointers should be part of this PR so all the info is available once merged.

36degrees (approved):
Thanks for coming up with a fix 🙌🏻 Let's go with this for now. If we see other issues caused by the symlinks we can revisit.

PR #4091, opened 2023-08-15
Scope npm workspaces packages under `@govuk-frontend`

romaricpascal:


tsconfig.base.json
```diff
@@ -16,14 +16,14 @@
     "types": [],
     "paths": {
       "govuk-frontend": ["./packages/govuk-frontend/src/govuk/all.mjs"],
```
colinrotherham, line 18: Probably super picky for such nice work

But shove this one to the end to keep it ordered A-Z

colinrotherham (approved):

PR #4092, opened 2023-08-15
Use new “Open this example in a new tab” link text

colinrotherham:
This PR adds:

1. Improved “Open this example in a new tab” link text [from the Design System](https://github.com/alphagov/govuk-design-system/pull/2994)
1. Markdown formatting on **Full page example** `notes:`
1. Markdown formatting on **Full page example** `scenario:`
1. Automatic "smart" typography and URL linking

# Before

All formatting inside `notes` is removed

Example with notes, before

# After

All formatting inside `notes` is now passed to Markdown with a **Notes** subheading

Example with notes, after

### DIFF
--- a/packages/govuk-frontend-review/src/views/full-page-examples/index.njk
+++ b/packages/govuk-frontend-review/src/views/full-page-examples/index.njk
@@ -24,14 +24,15 @@
 
       <h2 class="govuk-heading-m"><a href="/full-page-examples/{{ example.path }}" class="govuk-link">{{ example.name | unslugify }}</a></h2>
 
-      {% if example.scenario %}
-      <div class="app-prose-scope">
-        {{ example.scenario | markdown }}
-      </div>
-      {% endif %}
+      {% if example.scenario or example.notes %}
+        <div class="app-prose-scope">
+          {{ example.scenario | markdown if example.scenario }}
 
-      {% if example.notes %}
-        <p class="govuk-body govuk-!-font-size-16">{{ example.notes }}</p>
+          {% if example.notes %}
+            <h3>Notes</h3>
+            {{ example.notes | markdown }}
+          {% endif %}
+        </div>
       {% endif %}
 
       <hr class="govuk-section-break govuk-section-break--l" />
--- a/packages/govuk-frontend-review/src/views/full-page-examples/index.njk
+++ b/packages/govuk-frontend-review/src/views/full-page-examples/index.njk
@@ -21,22 +21,29 @@
     <hr class="govuk-section-break govuk-section-break--l" />
 
     {% for example in fullPageExamples %}
+    <section aria-labelledby="heading-{{ example.name }}">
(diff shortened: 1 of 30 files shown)


CharlotteDowns:
Oh this looks like an improvement 👍

packages/govuk-frontend-review/src/views/full-page-examples/index.njk
```diff
@@ -21,21 +21,29 @@
     <hr class="govuk-section-break govuk-section-break--l" />
 
     {% for example in fullPageExamples %}
+    <section aria-labelledby="heading-{{ example.name }}">
+      <h2 id="heading-{{ example.name }}" class="govuk-heading-l govuk-!-margin-bottom-2">{{ example.name | unslugify }}</h2>
 
-      <h2 class="govuk-heading-m"><a href="/full-page-examples/{{ example.path }}" class="govuk-link">{{ example.name | unslugify }}</a></h2>
+      <p class="govuk-body">
+        <a href="/full-page-examples/{{ example.path }}" target="_blank" class="govuk-link">
```
owenatgov, line 28: Worth adding `rel="noopener noreferrer"`?

packages/govuk-frontend-review/src/views/macros/showExamples.njk
```diff
@@ -17,12 +17,12 @@
 
   <section aria-labelledby="heading-{{ exampleName }}" class="govuk-!-margin-bottom-9">
     <div class="govuk-width-container">
-      <div class="govuk-heading-m">
-        <h3 id="heading-{{ exampleName }}" class="govuk-!-display-inline">{{ heading | safe }}</h3>
-        <a href="{{ path }}" class="govuk-link govuk-!-margin-left-1 govuk-!-font-size-16">
-          (open in a new window)
+      <h3 id="heading-{{ exampleName }}" class="govuk-heading-m govuk-!-margin-bottom-2">{{ heading | safe }}</h3>
+      <p class="govuk-body">
+        <a href="{{ path }}" target="_blank" class="govuk-link">
```
owenatgov, line 22: Same comment on `rel="noopener noreferrer"`

owenatgov (commented):
One comment about new tab links. Looking really nice.

packages/govuk-frontend-review/src/views/full-page-examples/index.njk
colinrotherham: @owenatgov Yeah course, will do

Just a heads up, I copied it from here where we forgot:

https://github.com/alphagov/govuk-design-system/blob/7980f36135b37f8bfb73dd949dc397948f0fd8e2/views/partials/_example.njk#L33

owenatgov: Oooo good spot. If I get a sec I'll patch that up as well.

owenatgov (approved):
Love a tidy up

PR #4093, opened 2023-08-16
Refactor positioning of radios and checkboxes

owenatgov:
## What
Refactor radios and checkboxes to use flexbox to handle positioning. Specific changes:

- Remove all float positioning from radios and checkboxes
- Use flexbox on labels so that their content is always centre aligned against the input
- Calculate position of radio rings and checkbox boxes using the "gutter" of the input touch target aka the extra 2px around the 40px width inputs
- Set labels to have a max width of the container minus the input touch target width and the padding on either side of the label
- Set hints to always be 100% width to prevent them appearing inline against small labels, now that `__item`'s are using `flex-wrap: wrap`
- Calculate the radio button position automatically based on the width of the touch target and the border width of the radio button

## Why
To make the positioning of radios and checkboxes dynamic in preparation for the new typography scale.

Closes https://github.com/alphagov/govuk-frontend/issues/3898

Based off work done in https://github.com/alphagov/govuk-frontend/pull/4040

## Notes
Since this is only a refactor of the underlying positioning of the components, I'm expecting the Percy diff to be next to nothing, no more than a few pixels.

Whilst this is a significant change for us, it is intended to make little to no difference to users and therefore should be treated as a bug fix change.


36degrees:
The small radios and checkboxes have changed in that they input is no longer [left-aligned with the label](https://github.com/alphagov/govuk-frontend/blob/9bc8c33d637ea112ff4bfc944ecdf28249a3fd6d/packages/govuk-frontend/src/govuk/components/checkboxes/_index.scss#L197-L205):

| Before | After |
| -- | -- |
| Small checkbox aligned with the label, with the extra touch area pushed into the margin | Small checkbox not aligned with the label, with the extra touch area included in the form group area |

Looks like this hasn't been caught by Percy either – worth adding small radios and checkboxes, and maybe examples with hints, to our [list of extra visual regression examples](https://github.com/alphagov/govuk-frontend/blob/15cfdb2f0a46a9d72ea23b30c1811dc25c2770d5/shared/tasks/browser.mjs#L50-L59)? Could be worth adding them with a separate PR to main and then rebasing so we can see the diff here.

owenatgov:
@36degrees Good idea, I'll make a separate PR for this. I've addressed this comment in the mean time.

owenatgov:
Percy PR spun off from this one https://github.com/alphagov/govuk-frontend/pull/4108

packages/govuk-frontend/src/govuk/components/checkboxes/_index.scss
```diff
@@ -4,23 +4,18 @@
 @import "../label/index";
 
 @include govuk-exports("govuk/component/checkboxes") {
-  $govuk-touch-target-size: 44px;
+  $govuk-touch-target-gutter: 4px;
```
colinrotherham, line 7: This PR looks brill. Thanks for clarifying the gutter

packages/govuk-frontend/src/govuk/components/checkboxes/_index.scss
```diff
@@ -75,29 +73,26 @@
   .govuk-checkboxes__label::after {
     content: "";
     box-sizing: border-box;
-
     position: absolute;
-    top: 11px;
-    left: 9px;
+    top: 13px;
+    left: 10px;
```
colinrotherham, line 78: Is it a good time to explain what these number mean?

packages/govuk-frontend/src/govuk/components/checkboxes/_index.scss
```diff
@@ -29,39 +24,42 @@
...
     opacity: 0;
-
     cursor: pointer;
   }
 
   .govuk-checkboxes__label {
-    display: inline-block;
+    display: flex;
+    flex-wrap: wrap;
+    align-items: center;
```
colinrotherham, line 37: Have we got an example set up for this?

I couldn't replicate the alignment issues without it:

>* Use flexbox on labels so that their content is always centre aligned against the input

packages/govuk-frontend/src/govuk/components/checkboxes/_index.scss
```diff
@@ -29,39 +24,42 @@
...
     margin-bottom: 0;
-    padding: 8px $govuk-checkboxes-label-padding-left-right govuk-spacing(1);
+    padding: (govuk-spacing(1) + $govuk-border-width-form-element) govuk-spacing(3);
     cursor: pointer;
     // remove 300ms pause on mobile
     touch-action: manipulation;
   }
 
+  // to account for when there are multiple children in a label
+  .govuk-checkboxes__label * {
```
colinrotherham, line 52: Mainly mentioned [labels with `flex-wrap` above](https://github.com/alphagov/govuk-frontend/pull/4093/files#r1302864343) since `label.html` with child `<strong>` would show:

Checkbox label children issue

packages/govuk-frontend/src/govuk/components/checkboxes/_index.scss
```diff
@@ -236,8 +217,8 @@
     //
     // Reduce the size of the check mark and re-align within the checkbox
     .govuk-checkboxes__label::after {
-      top: 15px;
-      left: 6px;
+      top: 17px;
+      left: (16px - $input-offset);
```
colinrotherham, line 221: Are we able to explain these offsets in relation to "container size" minus "checkbox width" etc?

packages/govuk-frontend/src/govuk/components/checkboxes/_index.scss
```diff
@@ -252,16 +233,14 @@
     // (If you do use them, they won't look completely broken... but seriously,
     // don't use them)
     .govuk-checkboxes__hint {
-      padding: 0;
-      clear: both;
+      padding-left: $govuk-touch-target-size;
```
colinrotherham, line 236: Mind checking hint alignment for small sizes?

Hint alignment issue

packages/govuk-frontend/src/govuk/components/checkboxes/_index.scss
owenatgov: Bugger. If you remove the `width: 100%` you'll see this most prominently in [this example](https://govuk-frontend-pr-4093.herokuapp.com/components/radios/with-legend-as-page-heading/preview). We probably need to retain the inline/block level styling of the label's children, but that makes it difficult to automatically centrally align. I'll have a muck around.

packages/govuk-frontend/src/govuk/components/checkboxes/_index.scss
owenatgov: I'll write something up. Basically it's not feasible to automate the positioning of the check mark because it's too complex a shape, but we can determine a few things like it's left position based on the small input offset.

packages/govuk-frontend/src/govuk/components/checkboxes/_index.scss
colinrotherham: Yeah, was thinking for safety it might be back to tweaking `padding-top` 😬

owenatgov: I believe I've solved this by removing `flex` from the label and using `align-self: center`

colinrotherham:
This one's got some (good) conflicts to de-duplicate `govuk-font()` since https://github.com/alphagov/govuk-frontend/pull/4267 merged

If you don't mind doing another rebase please 🙏

owenatgov:
@colinrotherham I remember where that extra space comes from. The hint now needs to be below the bottom bound of the visual element of the checkbox or radio. A knock-on effect of trying to de-floatify everything.

checkbox component with hint and devtools overlay showing flex bounding boxes

It's a good question to ask. I'm biased but I think that's ok. I don't think there's any harm in things being more spaced out and the space between the individual checkboxes indicates that  I had a go and trying to squeeze the label and the hint together but got nowhere. Throwing it back, do you think it's gonna be an issue?

colinrotherham:
@owenatgov Ah maybe, but it's only an issue by accident 😣

We've lost a collapsed margin so now both `.govuk-hint` + `.govuk-radios__item` are doubled up, compare:

* https://govuk-frontend-review.herokuapp.com/components/radios/with-hints-on-items/preview
* https://govuk-frontend-pr-4093.herokuapp.com/components/radios/with-hints-on-items/preview

Or you can see the margins now stacked here, but previously were collapsed:

No more collapsed margin

owenatgov:
Aha, problem identified.

I'm presuming that the lost margin collapse is a result of the flex things which is a pain. It turns out removing this is a bit of a hassle because of [this ultra-specific line in `govuk-hint`](https://github.com/alphagov/govuk-frontend/blob/d9f24216e964ce53355143b4687cca0f478924f0/packages/govuk-frontend/src/govuk/components/hint/_index.scss#L35-L37). The specificity is preventing this being removed simply without some raucous styling. For now I've removed the bottom margin from `govuk-checkboxes__item`, which I don't really like but it does the job.

Whilst I'm here I also tried nudging hints up by 5px so they're close to their original design which it turns out works!

owenatgov:
Now also applied to radios 👀

colinrotherham:
>…The specificity is preventing this being removed simply without some raucous styling. For now I've removed the bottom margin from govuk-checkboxes__item, which I don't really like but it does the job.

@owenatgov Bit annoying, cheers for that

Hmm, might have to tackle it though as it's made checkboxes (without hints) really close together 😣

owenatgov:
🤦🏻 🤦🏻 🤦🏻 Yeah that'll do that won't it...

Let's have another go.

owenatgov:
@colinrotherham I cracked it by just replicating the hint's specificity. Probably the best solution for this for the moment, unless you can figure out how to collapse margins in flexbox (I couldn't!).

colinrotherham (approved):
Thanks @owenatgov, looking good

Can you double check small radios/checkboxes for me?

Percy is showing wider spacing everywhere (good) but for small ones only the labels have shifted a tiny bit closer to their inputs

If design people are happy, go for it

owenatgov:
@colinrotherham I think I've figured it out. On the live version, small variants have a 1px right padding. I couldn't find any concrete reasoning from a quick history skim but my suspicion is that it's to create a little bit of space between the small hover state and the label. For ease I've reintroduced it. Whatcha reckon?

PR #4094, opened 2023-08-16
[WIP] Export type declarations for component source code

colinrotherham:
Closes https://github.com/alphagov/govuk-frontend/issues/2835 and up for discussion

PR #4096, opened 2023-08-17
Fix compiler issues suppressed by `skipLibCheck`

colinrotherham:
@domoscargin has been seeing some compiler checks fail only in GitHub Actions workflows:
https://github.com/alphagov/govuk-frontend/actions/runs/5881160187/job/15949158296

```console
Error: shared/lib/components.js(179,43): error TS2769: No overload matches this call.
  Overload 1 of 2, '(name: string, context: object): string', gave the following error.
    Argument of type 'object | undefined' is not assignable to parameter of type 'object'.
      Type 'undefined' is not assignable to type 'object'.
```

But due to `skipLibCheck: true` errors in `node_modules/@govuk-frontend/lib/components.js` won't appear locally

To make sure these checks always appear in editors and GitHub Actions workflows I've included:

1. Move type checks to `npm run lint:types` task
2. Keep "build only" type checks on `npm run build:types`
3. Ensure “build only” type checks also appear in editors
4. Restore type checks on external libraries (removes `skipLibCheck`)


tsconfig.json
```diff
@@ -6,10 +6,10 @@
       "path": "./.github/workflows/scripts/tsconfig.json"
     },
     {
-      "path": "./packages/govuk-frontend-review/tsconfig.json"
+      "path": "./packages/govuk-frontend-review/tsconfig.dev.json"
```
domoscargin: Going through commit by commit, so this may be explained later, but does the `.dev` in the filename allow TypeScript to do anything automagically, or is it only for flagging its purpose to us?

domoscargin (approved):
Think this all looks good - great detective work figuring it all out!

A coupla questions more to satisfy my curiosity/learn - nothing blocking.

tsconfig.base.json
```diff
@@ -8,7 +8,6 @@
     "noEmit": true,
     "noImplicitThis": true,
     "resolveJsonModule": true,
-    "skipLibCheck": true,
```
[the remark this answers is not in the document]
colinrotherham: Yeah, there was bump to `@types/vinyl-fs` back in May

It switched from [`@types/glob`](https://www.npmjs.com/package/@types/glob) to the now-native [`glob`](https://www.npmjs.com/package/glob) which fixed our main type mismatch issues

* https://github.com/DefinitelyTyped/DefinitelyTyped/pull/65258

tsconfig.json
colinrotherham: Nothing magic, I'll explain

Editors like Visual Studio Code "find the nearest tsconfig.json" so they missed out checks in `tsconfig.build.json`

Consider these two commits optional, but I'll explain further:

* **Move “development only” type checks to `tsconfig.dev.json`**
* **Ensure “build only” type checks also appear in editors**

The comment above shows us move `tsconfig.json` → `tsconfig.dev.json`

But later recreates a new wrapper `tsconfig.json` to enable checks across both:

```json
{
  "extends": "../../tsconfig.base.json",
  "files": [],
  "references": [
    {
      "path": "./tsconfig.build.json"
    },
    {
      "path": "./tsconfig.dev.json"
    }
  ]
}
```

We'd use each config as follows:

### tsconfig.dev.json
* Development types like `"jest", "jest-puppeteer"` so globals like `page` and `describe()` get types
* Type checks tasks, tests and other build tools

### tsconfig.build.json
* No development types like `"jest", "jest-puppeteer"` so we can flag accidental globals outside tests
* Optionally locks in `rootDir` and `outDir` to ensure we're not including development packages
* Optionally enables `noEmit: false` to build types, so checks verify [module](https://www.typescriptlang.org/tsconfig#module) + [target](https://www.typescriptlang.org/tsconfig#target) flavours
* Optionally enables strict mode which we wouldn't necessarily use in tests
* Excludes `"**/*.test.*"` files

But see all the issues "live" in the editor, not just in GitHub Actions workflows

colinrotherham: Oh and I've split out type checks into:

1. `npm run lint:types`
2. `npm run build:types`

So we see "build only" checks during the build step, but test + development things in the linting step

I've added a spike here on what publishing types might look like:

* https://github.com/alphagov/govuk-frontend/pull/4094

You quite quickly forget what it's like without autocomplete, types, descriptions in other projects

domoscargin: All good changes - thanks for thinking this stuff through.

colinrotherham:
Thanks @domoscargin

Should mention too that npm automatically fixed https://github.com/alphagov/govuk-frontend/pull/4096/commits/33316d643ff5f6670a88360701929fb2e1c65d45 following https://github.com/alphagov/govuk-frontend/pull/4091

PR #4056, opened 2023-08-08
[SPIKE] Run the GOV.UK Prototype Kit locally

colinrotherham:
This PR creates and installs an instance of GOV.UK Prototype Kit locally as part of https://github.com/alphagov/govuk-frontend/issues/3759

After install, we use `npm link` to replace GOV.UK Frontend with [**./packages/govuk-frontend**](https://github.com/alphagov/govuk-frontend/tree/main/packages/govuk-frontend)

I'd like to spike using Percy to:

1. Start the prototype
2. Screenshot the prototype home page
3. Screenshot the **Manage your prototype** page

This spots issues like missing fonts, assets etc

PR #4101, opened 2023-08-17
Add metadata to Prototype Kit plugin config

nataliecarey:
We're working on a new plugin page, we'll be making use of new metadata that comes from plugin config.  This PR adds that metadata.

I've also added the validator to the post-build tests, this uses `npx` and `@latest` to always look at the latest requirements.  I recommend this approach but it does erode the purity of repeatable builds, you could hardcode a version number in there or install `govuk-prototype-kit` as a (dev) dependency and remove both `npx` and `@latest`.

I've not added any tests or a strict type definition because I've added the validator which imposes the rules more strictly than any test I'd be contributing here.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4299,6 +4299,15 @@
         "@types/responselike": "^1.0.0"
       }
     },
+    "node_modules/@types/concat-stream": {
+      "version": "1.6.1",
+      "resolved": "https://registry.npmjs.org/@types/concat-stream/-/concat-stream-1.6.1.tgz",
+      "integrity": "sha512-eHE4cQPoj6ngxBZMvVf6Hw7Mh4jMW4U9lpGmS5GBPB9RYxlFg+CHaVN7ErNY4W9XfLIEn20b4VDYaIrbq0q4uA==",
+      "dev": true,
+      "dependencies": {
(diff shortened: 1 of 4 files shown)


colinrotherham:
Ahh this is brilliant, thank you @nataliecarey

We've got a spike open at the moment which will eventually include GOV.UK Prototype Kit in our automated tests:

* https://github.com/alphagov/govuk-frontend/pull/4056

Would you mind if I included these changes in that PR?

Once v5 testing is complete we'll switch to managing `govuk-prototype-kit` versions with Dependabot

colinrotherham:
Ahh I've tried it with the [`frontend-v5-compatibility`](https://github.com/alphagov/govuk-prototype-kit/compare/frontend-v5-compatibility) branch and it says:

```
Error: The following invalid keys exist in your config: constants
```

So we'll hold off on my last suggestion until `constants` is the approach for us

But this is still really useful so I've pushed up some config tweaks [from the spike](https://github.com/alphagov/govuk-frontend/pull/4056) too

colinrotherham (approved):

nataliecarey:
@colinrotherham looks good to me.

PR #4102, opened 2023-08-17
Fix: remove SASS warnings for govuk-frontend

guilhem-fry:
Hi, I'm from the the Data Products Team at the Cabinet Office, and I work on the GovSearch app.

SCSS decided to flag lots of deprecated divisions. This PR fixes it.

Thanks for your consideration

image


querkmachine:
Morning @guilhem-fry,

Unfortunately, I'm going to have to reject this PR. The CSS `calc` function is not a replacement for the `/` operator in this case, as it is unable to work with Sass variables.

The replacement for `/` is the [`math.div` function](https://sass-lang.com/documentation/modules/math/#div), however using it would require us to migrate Frontend to the Sass module system, [which we aren't doing just yet](https://github.com/alphagov/govuk-frontend/issues/1791).

You can silence this and other Sass deprecation warnings in your project by using the [`--quiet-deps` command line flag](https://sass-lang.com/documentation/cli/dart-sass/#quiet-deps) or the [`quietDeps` option](https://sass-lang.com/documentation/js-api/interfaces/stringoptionswithoutimporter/#quietDeps) if you're using the JavaScript API.

Issue #4072, opened 2023-08-11
Review content of thrown errors


romaricpascal:
## What

Review the wording of our error message and their names to make sure they're as useful as possible to our users

## Why

The clearer these messages, the easier it will be for users for fix the errors that we raise.

## Who needs to work on this

Developers, Tech writers

## Who needs to review this

Developers, Tech writers

## Done when

- [ ] We've gathered all the messages of all errors we'll be throwing in v5
- [ ] We've decided on the wording of the error messages
- [ ] We've implemented any changes necessary in the codebase

Vayras:
@romaricpascal I would like to contribute towards this issue.

colinrotherham:
Just a few per-developer things that have come up across error PRs so far

1. Adding a component name suffix to error messages
1. Explaining _why_ something is wrong (selector, instance type, missed config key)
1. Enclosing variable names in back ticks
1. Enclosing strings in single quotes
1. Enclosing strings in double quotes when they contain single quotes

Most important is to explain why something is wrong, where possible

romaricpascal:
Worth a quick look at how many of our components validate IDs set in `aria-controls` to ensure we're consistent in checking (or not checking) across the board.

colinrotherham:
Another note to say we've begun scoping using CSS selectors

>Checkboxes: [data-module="govuk-checkboxes"] not found

```mjs
throw new ElementError($module, {
  componentName: 'Checkboxes',
  identifier: `[data-module="${Checkboxes.moduleName}"]`
})
```

But others still use `$module`

```
throw new ElementError($module, {
  componentName: 'Exit this page',
  identifier: '$module'
})
```

Either way, do we need to worry when `$module` is overridden making our identifier incorrect?

domoscargin:
Following on from [this doc](https://docs.google.com/document/d/1CIHxeSW7sjUA7GbP01v9xWrANeQeReCw1N5-d8U2dFI/edit#heading=h.6vmoib5b9l5x) and some further discussions, we're aiming to format all our error messages with a mixture of plain English and selectors/variable names in backticks. 

In general, this might look like:

```
<COMPONENT>: <PLAIN_ENGLISH_ELEMENT> (`<ELEMENT_IDENTIFIER>`) [<PLAIN_ENGLISH_ELEMENT_PROPERTY> (`<PROPERTY_IDENTIFIER`)] <ERROR>
```

Where:

**COMPONENT**: Plain English component name (eg: "Character count")
[...]

PR #4104, opened 2023-08-18
Throw `TypeError` when the module is not of the expected type

romaricpascal:
Work in progress


packages/govuk-frontend/src/govuk/components/skip-link/skip-link.mjs
```diff
@@ -7,7 +7,7 @@ import { GOVUKFrontendComponent } from '../../govuk-frontend-component.mjs'
  */
 export class SkipLink extends GOVUKFrontendComponent {
   /**
-   * @protected
+   * @override
    * @type {HTMLAnchorElement}
    */
   $module = this.$module
```
romaricpascal: Odd, but that's how class fields work. Without that initialisation, `this.$module` would be set to `undefined` after the `super()` call in the `constructor()`... which would kind of defeat the purpose of encapsulating its setting in the base class.

More info [on what happens around these class fields on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Public_class_fields#examples).

packages/govuk-frontend/src/govuk/govuk-frontend-component.mjs
```diff
@@ -8,11 +8,29 @@ import { SupportError } from './errors/index.mjs'
...
  * @abstract
+ * @template {Element} [ModuleType=HTMLElement]
  */
 export class GOVUKFrontendComponent {
+  /**
+   * Defines the type of `Element` the components expects at instantiation
+   *
+   * If the component requires a more specialised type:
+   * - document the child class using `@extends`,
+   *   setting the required type in the generic part of `GOVUKFrontendComponent`.
```
domoscargin, line 19: Can we be explicit about using the `@augments` annotation to do this?

I'm also thinking it'd be good to think about what's in our contributing docs before v5. For example, I can see it being helpful having a "creating a new component" doc which takes a user step by step through creating a component and includes tips on specific stuff like this.

Or maybe a more generic "working with types" doc, since I'm guessing a lot of potential contributors are not going to be super familiar with them (we could also document "how to turn off type checking" to streamline the process for contributors, then take the actual typing work on ourselves, so we don't block folks).

Not necessarily something to do now, but maybe worth getting around to. @claireashworth - dunno if the [contributing docs](https://github.com/alphagov/govuk-frontend/tree/main/docs/contributing) are on the radar at all?

romaricpascal: Do you mean providing a clear `@extends GOVUKFrontendComponent<HTMLAnchorElement>` example?

Agreed on the need for some docs describing how we use TypeScript, links to interesting places and maybe recurring patterns in our code.

domoscargin: Yeah, it's `@augments GOVUKFrontendComponent<HTMLAnchorElement>` on the Skip Link, but good to have it here to show how its done without having to go look there.

romaricpascal: Ohh, I don't remember typing `@augments`, maybe something linted it that way.

romaricpascal:
@colinrotherham Ready for re-review, simplified implementation by replacing the `return this` with throwing errors 😊

romaricpascal:
Actually no, let me drop that first commit that's not relevant anymore. Sorry 😔

packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
```diff
@@ -119,7 +119,9 @@ export class Accordion extends GOVUKFrontendComponent {
     super()
 
     if (!($module instanceof HTMLElement)) {
-      return this
+      throw new TypeError(
+        'Expected `$module` to be an instance of `HTMLElement`'
```
colinrotherham, line 123: Shall we add component name prefixes everywhere and double quotes from https://github.com/alphagov/govuk-frontend/pull/4177?

```suggestion
        `Accordion: $module is not an instance of "HTMLElement"'
```

Or even better, tell the user what the incorrect instance is (e.g. `HTMLTextElement`)

```suggestion
        'Expected `$module` "${$module && 'constructor' in $module ? $module.constructor.name : $module}" is not an instance of `HTMLElement`'
```

colinrotherham:
Thanks @romaricpascal, would you mind updating the description please?

I'm really keen to make sure we tell users _why_ an error has been thrown

That could mean:

1. Using the `initAll()` try/catch to log the selector that was used
2. Logging `$module.constructor.name` where available (e.g. Character count on `<button>`)

packages/govuk-frontend/src/govuk/components/skip-link/skip-link.mjs
```diff
@@ -25,7 +25,9 @@ export class SkipLink extends GOVUKFrontendComponent {
     super()
 
     if (!($module instanceof HTMLAnchorElement)) {
-      return this
+      throw new TypeError(
+        'Skip Link: $module is not an instance of "HTMLAnchorElement"'
+      )
```
colinrotherham, line 30: Mind rebasing with the `MissingElementError` changes?

For each of these checks we can error for both:

1. Parameter `$module` is not provided
2. Parameter `$module` is not the expected type

Where 1) could be from a `null` selector or other falsy values

Would we need something like this?

```suggestion
      throw !$module
        ? new MissingElementError('Skip Link: $module is not found')
        : new TypeError(
            'Skip Link: $module is not an instance of "HTMLAnchorElement"'
          )
```

packages/govuk-frontend/src/govuk/components/skip-link/skip-link.test.js
```diff
@@ -81,5 +81,21 @@ describe('Skip Link', () => {
...
     })
+
+    it('throws when receiving the wrong type for $module', async () => {
+      await expect(
+        renderAndInitialise(page, 'skip-link', {
+          params: examples.default,
+          beforeInitialisation() {
+            // Remove the root of the components as a way
+            // for the constructor to receive the wrong type for `$module`
+            document.querySelector('[data-module]').remove()
```
colinrotherham, line 92: Should users with `null` selectors see `MissingElementError` instead?

More informative for `TypeError` to throw for the wrong selector element

E.g. Trying to pass a `<button>` into `new CharacterCount()`

romaricpascal:
Differentiated between the two errors 😊 Didn't try to factor in a `createModuleError` function to avoid spending time figuring types to make TypeScript happy as I wasn't quite sure what to put for the `HTMLElement` that would need passing.

packages/govuk-frontend/src/govuk/components/accordion/accordion.test.js
```diff
@@ -733,6 +733,45 @@ describe('/components/accordion', () => {
...
+
+          it('throws when $module is not set', async () => {
+            await expect(
+              renderAndInitialise(page, 'accordion', {
+                params: examples.default,
+                beforeInitialisation() {
+                  // Remove the root of the components as a way
+                  // for the constructor to receive the wrong type for `$module`
+                  document.querySelector('[data-module]').remove()
+                }
```
colinrotherham, line 745: Mind if we try an keep the boilerplate down?

Hopefully obvious that `$module.remove()` throws `MissingElementError`

```suggestion
                beforeInitialisation: ($module) => $module.remove()
```

packages/govuk-frontend/src/govuk/components/accordion/accordion.test.js
```diff
@@ -733,6 +733,45 @@ describe('/components/accordion', () => {
...
+                  document.querySelector('[data-module]').remove()
+
+                  // And replace with an `<svg>` element which doesn't have the right type
+                  const $svg = document.createElementNS(
+                    'http://www.w3.org/2000/svg',
+                    'svg'
+                  )
+                  $svg.setAttribute('data-module', 'accordion')
+                  document.body.appendChild($svg)
+                }
```
colinrotherham, line 768: Perhaps the same here?

Just a quick `outerHTML` over the top of `$module` would be great for tests

```suggestion
                beforeInitialisation: ($module) {
                  $module.outerHTML = '<svg data-module="accordion"></svg>'
                }
```

colinrotherham (commented):
It's doing a great job @romaricpascal 

We probably forget the `beforeInitialisation()` function accepts `$module`?

With that in mind, can we please reduce all the new boilerplate code?

### Swap `$module` with SVG to throw TypeError

```mjs
beforeInitialisation($module) {
  $module.outerHTML = '<svg data-module="accordion"></svg>'
}
```

### Remove $module entirely to throw MissingElementError

```mjs
beforeInitialisation($module) {
  $module.remove()
}
```

packages/govuk-frontend/src/govuk/components/accordion/accordion.test.js
colinrotherham: If it's helpful, can always add another `@param` like `moduleName`?

```suggestion
                beforeInitialisation($module, moduleName) {
                  $module.outerHTML = `<svg data-module="${moduleName}"></svg>`
                }
```

packages/govuk-frontend/src/govuk/components/button/button.test.js
```diff
@@ -359,6 +338,45 @@ describe('/components/button', () => {
...
+            beforeInitialisation() {
+              // Remove the root of the components for replacement
+              document.querySelector('[data-module]').remove()
+
+              // And replace with an `<svg>` element which doesn't have the right type
+              const $svg = document.createElementNS(
+                'http://www.w3.org/2000/svg',
+                'svg'
+              )
+              $svg.setAttribute('data-module', 'button')
```
colinrotherham, line 371: Probably want to lock this down but `[data-module]` works with any value:

https://github.com/alphagov/govuk-frontend/blob/0443c375979978fb2321267774151b4806a81fb5/shared/helpers/puppeteer.js#L115

But shall we stay consistent and always follow ~`button`~ `govuk-button` which `initAll()` uses?

https://github.com/alphagov/govuk-frontend/blob/0443c375979978fb2321267774151b4806a81fb5/packages/govuk-frontend/src/govuk/all.mjs#L54-L56

packages/govuk-frontend/src/govuk/components/character-count/character-count.mjs
```diff
@@ -70,7 +71,11 @@ export class CharacterCount extends GOVUKFrontendComponent {
     super()
 
     if (!($module instanceof HTMLElement)) {
-      return this
+      throw !$module
+        ? new MissingElementError('Character Count: $module not found')
+        : new TypeError(
+            'Character Count: $module is not an instance of "HTMLElement"'
+          )
```
colinrotherham, line 78: How about a non-blocking base class idea to solve these suffixes?

E.g. New getter `.log` to automatically suffix `Character count: `

```mjs
this.log = new Logger('Character count')

this.log.debug()
this.log.warn()
this.log.error()
```

Except it can also put its log messages into `this.log.throw()` for neater things like:

```suggestion
      this.logger.throw(
        $module ? TypeError : MissingElementError,
        $module
          ? '$module is not an instance of "HTMLElement"'
          : '$module not found'
      )
```

packages/govuk-frontend/src/govuk/components/accordion/accordion.test.js
romaricpascal: Oooh, hadn't thought of that, yeah, that's shorter :D

packages/govuk-frontend/src/govuk/components/accordion/accordion.test.js
romaricpascal: Good shout, didn't think of `outerHTML` for a less tedious replacement, I'll update 😊

packages/govuk-frontend/src/govuk/components/accordion/accordion.test.js
colinrotherham: Single line arrow functions work too 👍

shared/helpers/puppeteer.js
```diff
@@ -95,7 +95,11 @@ async function renderAndInitialise(page, componentName, options) {
   // as running it inside the body of the next `evaluate`
   // didn't provide a reliable execution
   if (options.beforeInitialisation) {
-    await page.$eval('[data-module]', options.beforeInitialisation)
+    await page.$eval(
+      '[data-module]',
+      options.beforeInitialisation,
+      componentName
```
colinrotherham, line 101: @romaricpascal Would you mind using what we call `moduleName`

Each class has a static property like `Accordion.moduleName` as `'govuk-accordion'`

But it can be formed from `govuk-${componentName}` to maintain compatibility with `initAll()`

```suggestion
      `govuk-${componentName}`
```

romaricpascal: Actually, no need for this extra param, `$module` already has al the information we need: we can `.getAttribute('data-module')` (not `.dataset.module` because of typing), and that gets us the value we want when replacing with a dummy element. 

Makes the replacement a little clearer as it shows the attribute value came from the original element, and also avoid having the template string in both places 😊

colinrotherham: That's fine too. Can you check you're happy keeping the loose `[data-module]` selector in https://github.com/alphagov/govuk-frontend/pull/4104#pullrequestreview-1617358822?

Was hoping to prevent other tests from selecting the wrong module:

### Before
```mjs
// ✅ <svg data-module="colin"></svg>
await page.$eval('[data-module]', options.beforeInitialisation)
```

### After
```mjs
// ❌ <svg data-module="colin"></svg>
await page.$eval(`[data-module="${moduleName}"]`, options.beforeInitialisation)
```

packages/govuk-frontend/src/govuk/components/accordion/accordion.test.js
```diff
@@ -733,6 +733,37 @@ describe('/components/accordion', () => {
...
+
+          it('throws when receiving the wrong type for $module', async () => {
+            await expect(
+              renderAndInitialise(page, 'accordion', {
+                params: examples.default,
+                beforeInitialisation($module) {
+                  // Replace with an `<svg>` element which is not an `HTMLElement` in the DOM (but an `SVGElement`)
+                  $module.outerHTML = `<svg data-module="${$module.getAttribute(
+                    'data-module'
+                  )}"></svg>`
```
colinrotherham, line 759: Yo know how we clearly put `accordion` on line 753?

Can we help newcomers out and save them a click by doing the same here?

```suggestion
                  $module.outerHTML = '<svg data-module="govuk-accordion"></svg>'
```

shared/helpers/puppeteer.js
romaricpascal: I'd say we're fine for that PR: `renderAndInitialise` only has one component on the page, so the topmost element with `[data-module]` is what we're after.

If we were to use the `moduleName` (which I'm not against, just think it's not in the scope of this PR), I'd rather have that in its own PR, so we can rework the helper a little and:
- grab that `moduleName` directly from the component class passed to the helper rather than re-implement that it's `govuk-${componentName}`
- take advantage of that to change the following `evaluate` into `$eval` as well, as it's no longer necessary to do `document.querySelector` (an opportunity missed in https://github.com/alphagov/govuk-frontend/commit/b5528761a799967b4d8aaa7fcab12e9e94cc6855)

packages/govuk-frontend/src/govuk/components/accordion/accordion.test.js
colinrotherham: Keeps the [simplicity of how you did it initially](https://github.com/alphagov/govuk-frontend/pull/4104#pullrequestreview-1617358822) above? I loved that

```js
$svg.setAttribute('data-module', 'accordion')
```

Together with `$module` we can be both simple and obvious

shared/helpers/puppeteer.js
colinrotherham: Brill. Aware it was unanswered until now, so I'm happy if you're happy

packages/govuk-frontend/src/govuk/components/accordion/accordion.test.js
romaricpascal: I appreciate that it's more all in plain sight to have an hardcoded `data-module="govuk-accordion"`. However given this test is the same from one component to another, I think there's a benefit for grabbing it from the existing element [^1]. That quickly allows porting it to a new component without risking to put the wrong module name in the attribute, one place less to worry about.

As mentionned before, I also think it clarifies what happens: we're not replacing with an element with an arbitrary `data-module`, we're injecting an element with the same `data-module` as the one that was there. 



I think there are other solutions if we're worried about that line being clearer to newcommers:
1. Add extra comment to clarify that we're pulling the component's attribute to ensure the initialisation will happen on the replacement element:

```suggestion
                  // Replace with an `<svg>` element which is not an `HTMLElement` in the DOM (but an `SVGElement`)
				  // reusing the `data-module` attribute of the original element to ensure the component 
                  // will be instantiated on this replacement element
                  $module.outerHTML = `<svg data-module="${$module.getAttribute(
                    'data-module'
                  )}"></svg>`
```

2. That line is a bit unwieldly being on two line, but it could easily be made more readable by getting the value earlier:

```suggestion
                  // Grab `data-module` from the existing element for reuse in the replacement element
                  // to ensure the component will initialise on it
                  const moduleName = $module.getAttribute(
                    'data-module'
                  );
                  
                  // Replace with an `<svg>` element which is not an `HTMLElement` in the DOM (but an `SVGElement`)
                  $module.outerHTML = `<svg data-module="${moduleName}"></svg>`
```

@alphagov/design-system-developers From discussing on Slack, we don't seem to be able to agree on a way to go there, would anyone be able to weigh in, please?

[^1]: I also acknowledge that the matter would be less of a question should that value have been the selector we use to grab the element under test, or a value compared in an expectation, where putting that hardcoded string straight in the tests would be much safer in the long run.

36degrees: I don't really have a strong opinion either way, but given we seem to be at a bit of an impasse, I'd lean towards hardcoding the attribute.

There are good reasons on both sides, but my thinking is:

- there are two other places in the block of code that would be copied where the component name is already hardcoded, so when porting to a new component you already have to change them anyway
- there are fewer things to go wrong (e.g. `data-module` being `undefined` for some unanticipated reason – with a hardcoded attribute there are fewer uncertainties you have to reason about in the code)
- we should prefer self-documenting code where possible, and the fact that we think this might need a comment is a bit of a 'code smell' for me

Having said all of that, we should bear in mind that this is entirely internal and easy to change – I don't think I would personally make this a blocking thing if I was reviewing.

I do think it'd be neat to abstract the tests for common component functionality into a reusable set of tests at some point – is it worth creating an issue to do that once we've got v5 out the way?

romaricpascal: Cheers for weighing on this. I've hardcoded the attributes in my last push so that should be all clear for a last review. I'll raise an issue for abstracting tests for common components functionnality as well.

packages/govuk-frontend/src/govuk/components/character-count/character-count.mjs
romaricpascal: Can we track that option as part of #4072 as this relates more to this aspect of the errors. One thing that comes to mind here is whether any logging through that class would be able to be dropped by minifiers similarly to how they can drop `console.<logLevel>`. It does open a line of thought where the prefixing does not happen in the error classes, though, so that's definitely worth tracking 😊

packages/govuk-frontend/src/govuk/components/button/button.test.js
romaricpascal: I like the idea of referring to `Component.moduleName`, but I think it'd make sense to do that as part of a work that'd make our test helpers (`goToComponent`, `renderAndInitialise`,...) all rely on the component's class to get that kind of information (would ensure things are systemised from one component to another as well). If that's fine with you, I'd be keen to track that as a separate issue from this PR 😊

colinrotherham: Yeah sounds good. It might catch us out with nested components one day 👍

Thanks Romaric

packages/govuk-frontend/src/govuk/components/character-count/character-count.mjs
colinrotherham: Yeah not a blocker

We'd have the same [`drop_console`](https://terser.org/docs/options/#compress-options) problem anywhere but the error "throw" would still be kept

Just an idea to save a few bytes with `new Logger('Character count')` keeping the string safe

packages/govuk-frontend/src/govuk/errors/index.mjs
```diff
@@ -47,3 +47,25 @@ export class SupportError extends GOVUKFrontendError {
...
+  name = 'ElementError'
+
+  /**
+   * @param {Element} element - The element in error
+   * @param {object} options - options
+   * @param {string} options.componentName - The name of the component throwing the error
+   * @param {string} options.identifier - An identifier that'll let the user understand which element has an error (variable name, CSS selector)
+   * @param {typeof HTMLElement} [options.expectedType] - The type that was expected for the element
+   */
+  constructor(element, { componentName, identifier, expectedType }) {
```
colinrotherham, line 64: Thanks for the brill work in this PR @romaricpascal

One super tiny thing and we're good to go! If you turn on [`"strict": true`](https://www.typescriptlang.org/tsconfig#strict) it'll spot that we've got `@param` `[options.expectedType]` marked as optional but line 67 can crash when accessing `.name`

Optional is good though

If we default `expectedType = HTMLElement` we can remove all these lines:

```patch
--- A/packages/govuk-frontend/src/govuk/components/radios/radios.mjs
+++ B/packages/govuk-frontend/src/govuk/components/radios/radios.mjs
@@ -33,8 +33,7 @@
       throw new ElementError($module, {
         componentName: 'Radios',
+        identifier: '$module'
-        identifier: '$module',
-        expectedType: HTMLElement
       })
     }
```

Makes sense that `HTMLElement` is always the default expected type:

```suggestion
  constructor(element, { componentName, identifier, expectedType = HTMLElement }) {
```

colinrotherham (approved):

packages/govuk-frontend/src/govuk/errors/index.mjs
romaricpascal: Updated it in the last push 🥳

packages/govuk-frontend/src/govuk/errors/index.mjs
```diff
@@ -47,3 +47,28 @@ export class SupportError extends GOVUKFrontendError {
...
+   * @param {string} options.identifier - An identifier that'll let the user understand which element has an error (variable name, CSS selector)
+   * @param {typeof HTMLElement} [options.expectedType] - The type that was expected for the element
+   */
+  constructor(
+    element,
+    { componentName, identifier, expectedType = HTMLElement }
+  ) {
+    const reason = !element
+      ? `${identifier} not found`
+      : `${identifier} is not an instance of "${expectedType.name}"`
```
colinrotherham, line 70: Prettier really isn't helping these parameter lines 🙈

Reminds me of a recent discussion on destructuring with default values too

Think you and @36degrees would prefer this?

```suggestion
  constructor(element, options) {
    const expectedType = options.expectedType || HTMLElement
    
    const reason = !element
      ? `${options.identifier} not found`
      : `${options.identifier} is not an instance of "${expectedType.name}"`
```

packages/govuk-frontend/src/govuk/errors/index.mjs
colinrotherham: Cheers, approved it but remembered earlier feedback on destructuring in https://github.com/alphagov/govuk-frontend/pull/4104#discussion_r1327291883

packages/govuk-frontend/src/govuk/errors/index.mjs
romaricpascal: I'll update after lunch 😊

colinrotherham: Whilst [fixing those conflicts](https://github.com/alphagov/govuk-frontend/pull/4104#issuecomment-1723436569) for you I've pushed up a suggestion that keeps your original destructuring with:

```js
  constructor(element, { componentName, identifier, expectedType }) {
    expectedType = options.expectedType || window.HTMLElement
```

But we have Babel now [kept another branch](https://github.com/alphagov/govuk-frontend/compare/components-type-error...components-type-error-nullish#diff-e17a722e3c0cbe12ed4c41bcb10900a1cd5b8d13c0e10825e222baaedd046bdb) if you'd prefer:

```js
  constructor(element, { componentName, identifier, expectedType }) {
    expectedType ??= window.HTMLElement
```

colinrotherham (approved):

Issue #3478, opened 2023-04-03
Clarify our JavaScript components' public API


romaricpascal:
## What

We're drawing a clear line between the properties and methods of our JavaScript components that are meant to be publicly accessed and those that are only for internal use.

## Why

Currently, there's no clear distinction about what's for internal us and what's meant to be used by service code. This makes a very wide area for breaking changes, that we don't intentionally control, slowing down our capacity to make changes in places that should be private but are exposed. Also, while accessible to service code, some of these properties and methods are also not designed for outside use in their API.

With the opportunity for breaking changes in v5, this is a chance to clearly separate public and private APIs. This provides to services clarity on what our components are designed to offer, and to us safety when deciding what we can change without it being a breaking change.

## Who needs to work on this

Developers, Tech writers

## Who needs to review this

Developers, Tech writers

## Done when

```[tasklist]
### v4
- [ ] https://github.com/alphagov/govuk-frontend/issues/3473
```

```[tasklist]
### v5
- [ ] https://github.com/alphagov/govuk-frontend/issues/3479
- [ ] https://github.com/alphagov/govuk-frontend/issues/4008
- [ ] https://github.com/alphagov/govuk-frontend/issues/4009
- [ ] https://github.com/alphagov/govuk-frontend/issues/4035
- [ ] https://github.com/alphagov/govuk-frontend/issues/4128
- [ ] https://github.com/alphagov/govuk-frontend/issues/4129
- [ ] https://github.com/alphagov/govuk-frontend/issues/4130
- [ ] https://github.com/alphagov/govuk-frontend/issues/4131
- [ ] https://github.com/alphagov/govuk-frontend/issues/4125
- [ ] https://github.com/alphagov/govuk-frontend/issues/4126
- [ ] https://github.com/alphagov/govuk-frontend/issues/4127
- [ ] https://github.com/alphagov/govuk-frontend/issues/4144
- [ ] https://github.com/alphagov/govuk-frontend/issues/4036
- [ ] https://github.com/alphagov/govuk-frontend/issues/4037
- [ ] https://github.com/alphagov/govuk-frontend/issues/4074
- [ ] https://github.com/alphagov/govuk-frontend/issues/4073
- [ ] https://github.com/alphagov/govuk-frontend/issues/4010
- [ ] https://github.com/alphagov/govuk-frontend/issues/3904
- [ ] https://github.com/alphagov/govuk-frontend/issues/4017
[...]

PR #4106, opened 2023-08-18
Enable ESLint stylistic typed linting in source code

colinrotherham:
This PR enables [ESLint stylistic typed linting](https://typescript-eslint.io/linting/configs/#projects-with-type-checking)

It highlights opportunities to replace older JavaScript with new, but also uses [TypeScript `{ "strict": true }`](https://www.typescriptlang.org/tsconfig#strict) to add warnings when our GOV.UK Frontend source code is too loose or unnecessarily defensive

As part of https://github.com/alphagov/govuk-frontend/pull/4323 we've already fixed:

* Class fields with incorrect `null` default
* Selector `null` return values not handled
* Timeout and interval IDs cleared when `null`
* Confg allows `null` as `typeof value === 'object'`

But this PR includes a few more:

1. Potentially `null` attribute `NamedNodeMap` values
1. Potentially `null` element `.parentNode` values
1. Potentially `null` element `.textContent` values

Whilst quite pedantic for internal code, it's a huge safety net as we [clarify our public API](https://github.com/alphagov/govuk-frontend/issues/3478) and catch issues like:

* https://github.com/alphagov/govuk-frontend/pull/4394

## Stylistic typed linting

This change lets ESLint use types to reduce unnecessary code with [suggestions to use modern syntax](https://github.com/alphagov/govuk-frontend/pull/4286)

Unnecessary conditional

## Strict mode checks (existing)

Here's a list of the compiler options we already enabled in the past:

* [`strictBindCallApply`](https://www.typescriptlang.org/tsconfig#strictBindCallApply) – Function `bind`, `call` and `apply` must be checked (added in https://github.com/alphagov/govuk-frontend/commit/73a706a07b70bdf3c31b876dcb0fbfd6f1bc31e4)
* [`strictFunctionTypes`](https://www.typescriptlang.org/tsconfig#strictFunctionTypes) – Function parameters must be checked (added in https://github.com/alphagov/govuk-frontend/commit/73a706a07b70bdf3c31b876dcb0fbfd6f1bc31e4)
* [`noImplicitThis`](https://www.typescriptlang.org/tsconfig#noImplicitThis) – Context for `this` must be checked (added in https://github.com/alphagov/govuk-frontend/commit/8b3d6f35b1c202ecb1808788e84df93ed56596ed)

## Strict mode checks (new)

Here's a list of the compiler options enabled in this PR:

* [`alwaysStrict`](https://www.typescriptlang.org/tsconfig#alwaysStrict) – Assume code has `"use strict"`
* [`strictNullChecks`](https://www.typescriptlang.org/tsconfig#strictNullChecks) – Possible `null` and `undefined` must be checked
* [`strictPropertyInitialization`](https://www.typescriptlang.org/tsconfig#strictPropertyInitialization) – Class field values must be checked
* [`noImplicitAny`](https://www.typescriptlang.org/tsconfig#noImplicitAny) – All types must be added (e.g. in JSDoc)
* [`noUncheckedIndexedAccess`](https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess) – Array and object key/values must be checked
* [`useUnknownInCatchVariables`](https://www.typescriptlang.org/tsconfig#useUnknownInCatchVariables) – Assume errors are unknown until `error instanceof`

### DIFF
--- a/packages/govuk-frontend/.eslintrc.js
+++ b/packages/govuk-frontend/.eslintrc.js
@@ -18,8 +18,8 @@ module.exports = {
       },
       plugins: ['@typescript-eslint', 'es-x'],
       extends: [
-        'plugin:@typescript-eslint/recommended',
-        'plugin:@typescript-eslint/recommended-requiring-type-checking',
+        'plugin:@typescript-eslint/strict-type-checked',
+        'plugin:@typescript-eslint/stylistic-type-checked',
         'plugin:es-x/restrict-to-es2015',
         'prettier'
       ],
--- a/packages/govuk-frontend/.eslintrc.js
+++ b/packages/govuk-frontend/.eslintrc.js
@@ -27,6 +27,14 @@ module.exports = {
         browser: true
       },
       rules: {
+        // Allow void return shorthand in arrow functions
+        '@typescript-eslint/no-confusing-void-expression': [
+          'error',
+          {
+            ignoreArrowShorthand: true
+          }
+        ],
+
         // Check type support for template string implicit `.toString()`
         '@typescript-eslint/restrict-template-expressions': [
           'error',
--- a/packages/govuk-frontend/.eslintrc.js
+++ b/packages/govuk-frontend/.eslintrc.js
@@ -1,4 +1,4 @@
-const { resolve } = require('path')
+const { join } = require('path')
 
 module.exports = {
   settings: {
@@ -14,7 +14,7 @@ module.exports = {
       parserOptions: {
         // Note: Allow ES2015 for import/export syntax
         ecmaVersion: '2015',
-        project: [resolve(__dirname, 'tsconfig.dev.json')]
+        project: [join(__dirname, 'tsconfig.build.json')]
       },
       plugins: ['@typescript-eslint', 'es-x'],
       extends: [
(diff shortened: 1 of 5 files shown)


romaricpascal:
Cheers! Not sure it's one we'll merge straight away, but it's good to have the steps to enable it. I'm also wondering if us throwing (rather than returning early) during initialisation will not remove a couple of these "watch out it may be `null`" warnings, which would be nice and may make the list of things to fix less daunting 😊

domoscargin:
Just rebasing to keep this branch up to date

colinrotherham:
Going to remove **[SPIKE]** from the title

Everything passes if https://github.com/alphagov/govuk-frontend/pull/4323 is merged

colinrotherham:
Following dev catch-up this week we agreed to get this PR merged to test things out

I've assumed we're undecided on ES2015+ syntax suggestions from https://github.com/alphagov/govuk-frontend/pull/4286 so they're turned off in https://github.com/alphagov/govuk-frontend/pull/4106/commits/331d7f0bd1718303d43a8823bbec54a4eff2ac9d

@domoscargin @romaricpascal Mind giving this the green tick please?

Thanks 🙌

domoscargin (approved):
Makes sense, let's give it a go.

PR #4117, opened 2023-08-21
Make `initAll` log a `SupportError`

romaricpascal:
`initAll` was still returning early silently when GOV.UK Frontend is not supported. This makes it log a `SupportError` to align the message with components that throw `SupportError`s in that situation.


colinrotherham (approved):
Thanks @romaricpascal

[This came up in a previous comment](https://github.com/alphagov/govuk-frontend/pull/4075#discussion_r1295535776) where I thought we already logged!

Rather than two places that throw, do we need a `checkSupport()` helper?

romaricpascal:
Oh, had completely missed those. I'm not sure we need a `checkSupport` as the two places do slightly different things:

- the base component `throw`s
- `initAll` logs

Don't think there's a point in making the `initAll` catch a potential error to then log it, simpler to just have it run the shared `isSupported` and log directly 😊

romaricpascal:
I'll rebase (when I get a more stable connection 😆) and see if it fixes Percy's failure.

colinrotherham:
Mainly so stack traces go through the same throwing code, easier to mock etc. Not a blocker!

PR #4118, opened 2023-08-22
Fix “dev only” type checks in `./src` symlink directories

colinrotherham:
This commit brings changes from 8b20a4d to `./src` symlink directories

When accidentally opening a Jest test or Gulp task file inside `./src` globals for Jest etc weren't supported:

```
describe()
beforeAll()
beforeEach()
it()
```

This is now resolved. We previously enabled type discovery in:

* https://github.com/alphagov/govuk-frontend/pull/4089


romaricpascal (approved):
Nice! Hopefully that's the last adjustment we need to make for these symlinks 🥳

PR #4121, opened 2023-08-24
Ensure stat comments only run on PRs

domoscargin:
Stat comments are part of the `tests` workflow, which runs on pulls and pushes, but should only run on PRs.

This should hopefully make sure that happens.


colinrotherham (approved):
Looks good. This manual run was correctly skipped:
https://github.com/alphagov/govuk-frontend/actions/runs/5962662757

Would be nice for non-PRs to write to GitHub workflow status but one for another day!

PR #3919, opened 2023-07-06
Clarify docs for all nested options

colinrotherham:
I noticed last week that we sometimes "dot-prefix" nested option names with the associated parent or component

The Prototype Kit team are looking into generating documentation from component `macro-options.json` so this change copies what we did for [**Radios** and **Checkboxes** `conditional`](https://github.com/alphagov/govuk-frontend/commit/b815522894e1f841f019ef61d91e6e690eb768af) to structure all nested options correctly

### Accordion

1. ~`heading.text`~, ~`heading.html`~ → `heading`
1. ~`summary.text`~, ~`summary.html`~ → `summary`
1. ~`content.text`~, ~`content.html`~ → `content`

### Summary list

1. ~`key.text`~, ~`key.html`~ → `key`
1. ~`value.text`~, ~`value.html`~ → `value`
1. ~`actions.text`~, ~`actions.html`~ → `actions`

Like other nested options, these now show as separate tables (with links) on the Design System website:

Accordion items showing nested options

PR #4122, opened 2023-08-24
Ensure Summary list `actions` for both rows and cards are consistent

colinrotherham:
Macro options for [**Summary list**](https://design-system.service.gov.uk/components/summary-list/) `rows[].actions` and `card.actions` are shared

This PR ensures they match in description and option order

As mentioned at Show & Tell this morning, we've realised that Design System HTML will include duplicate IDs for nested options with the same key names when updating to GOV.UK Frontend v5, but we can fix this separately

## Nested options
For context, we expanded older `key`, `value` and `actions` nested options recently:

* https://github.com/alphagov/govuk-frontend/pull/3919

Which matches how newer options like `card` are nested:

* https://github.com/alphagov/govuk-frontend/pull/2931


colinrotherham:
>As mentioned at Show & Tell this morning, we've realised that Design System HTML will include duplicate IDs for nested options with the same key names when updating to GOV.UK Frontend v5, but we can fix this separately

A quick fix would be to include parent key names in ID attributes ~`--actions`~ `--card-actions`

But with two identical tables named **Options for actions** (rows + cards) we should write up an issue

Example showing parent names included in nested options names

querkmachine:
> But with two identical tables named Options for actions (rows + cards) we should write up an issue

There's an existing issue regarding the duplicate IDs here: https://github.com/alphagov/govuk-design-system/issues/2260

romaricpascal (approved):

Issue #1739, opened 2020-02-19
Use SVG lockup for GOV.UK in header


36degrees:
## What
Use an SVG lockup of the GOV.UK logotype in the header.

This will help ensure consistent alignment between the crown emblem and the text, and ensure that the correct typeface is used regardless of whether the webfont loads.

This was previously discussed in https://github.com/alphagov/govuk-frontend/issues/1575#issuecomment-532689552.

## Why

At the minute, our font loading strategy is to use `display: fallback` which gives a short (3s) swap period during which the font can be swapped out to Transport. If the font takes longer than 3s to load (e.g. on a slow 3G connection) the font swap will not occur and the page will use the next font in the stack (usually Arial).

This means 'GOV.UK' in the logo in the header is rendered incorrectly. As this forms a key part of the brand, we should take care to ensure that it is rendered correctly wherever possible.

@Nooshu has been looking at #1437 again and we might consider changing the font-face `display` attribute to:

- `swap` which gives an infinite swap time, but makes it more likely that Arial will be displayed briefly before being swapped out (because, at least as implemented in Chrome, there is no block period)
- `optional` which means the browser can choose not to download the font at all, for example if it knows it is on a slow connection.

Separately, we've seen issues with alignment between the crown and the logotype before on a few occasions, and using a lockup should fix this.

## Further details

We'll need to consider:

- the fallback PNG, used by browsers that do not support SVG
- how the logo is announced to screen readers

## Who needs to know

?

nickcolley:
Think this would be a breaking change since you'd end up with [crown] GOV.UK GOV.UK if users had not updated their markup.

nickcolley:
I've made a very highly optimised version of this.

Just the crown came to 1kb, with the GOV.UK and optimisation it comes to 1.56k.

```svg
<svg  role="presentation" focusable="false" class="govuk-header__logotype-crown" 
 xmlns="http://www.w3.org/2000/svg" viewBox="0 0 557.3 84.1" height="26" width="152">
[...]

PR #4150, opened 2023-08-30
Header menu button position refactor

owenatgov:
## What/Why
Refactor the header component's navigation menu button styling:

- Move specific, repeated values into variables
- Apply a min hight and a max width to the menu button
- Calculates the top offset of the menu button based on the height of the button against the height of the header
- Increases the expected width of the menu button
- Increases the height of the menu button to 24px, allowing us to additionally resolve https://github.com/alphagov/govuk-frontend/issues/4357

Additional details can be found in comments within commits.

Resolves https://github.com/alphagov/govuk-frontend/issues/3896

Reimplementation off of https://github.com/alphagov/govuk-frontend/pull/4065

## Thoughts
I couldn't figure out a good alternative to using right hand padding on the logo and an absolute positioned menu button. Ideally we would dynamically calculate the padding but sass doesn't have the means to fetch dynamic width. This way at least there's a relationship between the padding and the button width and it's easier to change the padding/button width. Open to ideas.

I'm wondering if there's a more explicit way to define the height of the header or if our assumption that the crown is always 30px is enough. We want it to stay at 50px but that's entirely human memory driven. There's nothing in our code to reflect that. Nothing to lose sleep over.

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/header/_index.scss
+++ b/packages/govuk-frontend/src/govuk/components/header/_index.scss
@@ -6,6 +6,7 @@
   $govuk-header-link-active: #1d8feb;
   $govuk-header-nav-item-border-color: #2e3133;
   $govuk-header-link-underline-thickness: 3px;
+  $govuk-header-vertical-spacing-value: 2;
 
   .govuk-header {
     @include govuk-font($size: 16, $line-height: 1);
@@ -28,7 +29,7 @@
     @include govuk-clearfix;
     position: relative;
     margin-bottom: -$govuk-header-border-width;
-    padding-top: govuk-spacing(2);
+    padding-top: govuk-spacing($govuk-header-vertical-spacing-value);
     border-bottom: $govuk-header-border-width solid $govuk-header-border-color;
   }
 
@@ -160,8 +161,8 @@
   }
 
   .govuk-header__logo {
-    @include govuk-responsive-margin(2, "bottom");
     padding-right: govuk-spacing(8);
+    @include govuk-responsive-margin($govuk-header-vertical-spacing-value, "bottom");
 
     @include govuk-media-query($from: desktop) {
       width: 33.33%;
--- a/packages/govuk-frontend/src/govuk/components/header/_index.scss
+++ b/packages/govuk-frontend/src/govuk/components/header/_index.scss
@@ -212,6 +212,7 @@
     border: 0;
     color: govuk-colour("white");
     background: none;
+    word-break: break-all;
     cursor: pointer;
 
     &:hover {
--- a/packages/govuk-frontend/src/govuk/components/header/_index.scss
+++ b/packages/govuk-frontend/src/govuk/components/header/_index.scss
@@ -11,6 +11,7 @@
   // as the crown svg height is the only thing that controls the height of the header
   $govuk-header-crown-height: 30px;
   $govuk-header-menu-button-height: 20px;
+  $govuk-header-menu-button-width: 80px;
 
   .govuk-header {
     @include govuk-font($size: 16, $line-height: 1);
@@ -165,8 +166,10 @@
   }
 
   .govuk-header__logo {
-    padding-right: govuk-spacing(8);
     @include govuk-responsive-margin($govuk-header-vertical-spacing-value, "bottom");
+    // Protect the absolute positioned menu button from overlapping with the
+    // logo with right padding using the button's width
+    padding-right: $govuk-header-menu-button-width;
 
     @include govuk-media-query($from: desktop) {
       width: 33.33%;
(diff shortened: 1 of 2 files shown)


packages/govuk-frontend/src/govuk/components/header/_index.scss
```diff
@@ -166,8 +174,16 @@
...
-    top: govuk-spacing(4);
+    // calculate top offset by:
+    // - getting the vertical spacing for the top and the bottom of the header
+    // - adding that to the crown height
+    // - dividing it by 2 so you have the vertical centre of the header
+    // - subtracting half the height of the menu button
+    top: (((govuk-spacing($govuk-header-vertical-spacing-value) * 2) + $govuk-header-crown-height) / 2) -
+      ($govuk-header-menu-button-height / 2);
     right: 0;
+    max-width: $govuk-header-menu-button-width;
```
colinrotherham, line 185: Looks good @owenatgov

My only hesitation would be around the new `max-width: 80px` restriction.

Teams can customise `params.menuButtonLabel` so full width with a logo-sized safe area would be better? Probably a lot easier to do once https://github.com/alphagov/govuk-frontend/issues/1739 lands

Menu button with long name

owenatgov: I made a comment in https://github.com/alphagov/govuk-frontend/pull/4150/commits/e9d62b4d19f5a7554eb948e067d75a66f3db6798 that the only instance of a service not using the text "menu" that's publicly available is the Welsh for menu: "Dewislen". However, that's a good point, we could just turn the calculation the other way around. I had a go at "shrink wrapping" the logo's width but was having trouble with it. I also need to have a think about how we could set the padding to the rest of the width container; I'm sure I've managed to do this before? Let me muck around with this. If you have any ideas I am very open to them.

colinrotherham: Here's what I was thinking

Including margins, the logo occupies 183px width (or 191px using Arial)

Will need to adjust for potential text-only zoom, but given a 200px "safe area" we could:

```patch
--- A/packages/govuk-frontend/src/govuk/components/header/_index.scss
+++ B/packages/govuk-frontend/src/govuk/components/header/_index.scss
@@ -10,8 +10,8 @@
   // This crown height is only used to calculate top offset of mobile menu button
   // as the crown svg height is the only thing that controls the height of the header
   $govuk-header-crown-height: 30px;
+  $govuk-header-logo-width: 200px;
   $govuk-header-menu-button-height: 20px;
-  $govuk-header-menu-button-width: 80px;
 
   .govuk-header {
     @include govuk-font($size: 16);
@@ -83,6 +83,7 @@
     @include govuk-link-style-inverse;
 
+    max-width: $govuk-header-logo-width;
     text-decoration: none;
 
     &:hover {
@@ -151,10 +152,6 @@
 
   .govuk-header__logo {
     @include govuk-responsive-margin($govuk-header-vertical-spacing-value, "bottom");
-    // Protect the absolute positioned menu button from overlapping with the
-    // logo with right padding using the button's width
-    padding-right: $govuk-header-menu-button-width;
-
     @include govuk-media-query($from: desktop) {
       width: 33.33%;
       padding-right: $govuk-gutter-half;
@@ -182,13 +179,16 @@
     top: (((govuk-spacing($govuk-header-vertical-spacing-value) * 2) + $govuk-header-crown-height) / 2) -
       ($govuk-header-menu-button-height / 2);
     right: 0;
+    width: 100%;
-    max-width: $govuk-header-menu-button-width;
     min-height: $govuk-header-menu-button-height;
     margin: 0;
+    // Protect the absolute positioned menu button from overlapping with the
+    // logo with left padding using the logo's width
+    padding: 0 0 0 $govuk-header-logo-width;
-    padding: 0;
     border: 0;
     color: govuk-colour("white");
     background: none;
+    text-align: right;
     cursor: pointer;
 
     &:hover {
```

colinrotherham: Downsides are that the button's clickable area will grow to fill non-logo space

Not a bad thing at mobile size, but could fix the width at tablet size perhaps

colinrotherham: Pushed up a branch [header-menu-button-position-refactor-colin](https://github.com/alphagov/govuk-frontend/compare/header-menu-button-position-refactor-colin) with commit https://github.com/alphagov/govuk-frontend/commit/4b0273849b36a881bf590f55491b0b0c43b30316

colinrotherham:
@owenatgov Mind rebasing this with `origin/main`?

Mainly to pick up these two:

* https://github.com/alphagov/govuk-frontend/pull/4449
* https://github.com/alphagov/govuk-frontend/pull/4498

packages/govuk-frontend/src/govuk/components/header/_index.scss
```diff
@@ -6,6 +6,12 @@
   $govuk-header-link-active: #1d8feb;
   $govuk-header-nav-item-border-color: #2e3133;
   $govuk-header-link-underline-thickness: 3px;
+  $govuk-header-vertical-spacing-value: 2;
+  // This crown height is only used to calculate top offset of mobile menu button
+  // as the crown svg height is the only thing that controls the height of the header
+  $govuk-header-crown-height: 30px;
+  $govuk-header-menu-button-height: 24px;
+  $govuk-header-menu-button-width: 80px;
```
colinrotherham, line 14: Think I'm happy to approve, is this width definitely enough?

E.g. "Menu" in Welsh, quick double check?

colinrotherham (approved):

owenatgov:
@colinrotherham Good shout. Testing with "Dewislen" (menu in Welsh) and increase the font size to 16px which ti will be with the new typography scale, get a total with of 78.359px width on the button. I'm feeling positive so I'm gonna be cheeky and merge.

Issue #3924, opened 2023-07-07
Increase margins between form groups for new typography scale


owenatgov:
## What
Update the bottom between form groups.

Should be merged into https://github.com/alphagov/govuk-frontend/pull/2421 and not `main`.

## Why
[During testing of the govuk-frontend components against the new typography scale](https://github.com/alphagov/govuk-design-system/issues/2289), we noticed that on small screens, form groups now looked quite tight next to each other due to the new larger font size on smaller screens.

## Who needs to work on this
Developers

## Who needs to review this
Developers, designers

owenatgov:
During review of this work, we discovered that a `govuk-responsive-margin` of 6 is being used by a number of components and elements:

- Accordion
- Button
- Button group
- Character count
- Details
- Inset text
- Pagination
- Summary list
- Table
- Tabs
- Task list
- Textarea
- Warning Text
- Our own code windows in the review app

This is our unspoken standard bottom margin and helps to define vertical visual rhythm through govuk-frontend pages. It would be risky to only change one of these without changing the others but this could be quite significant. I'll be seeking input from our designers on how best to proceed.

owenatgov:
Moving this into blocked just for now until there are more designers on hand on Monday.

owenatgov:
Following a chat amongst the team we're now not going to do this. The risk to the disruption of the responsive scale is greater than the perceived impact of things looking squashed.

PR #4156, opened 2023-08-31
Remove responsive spacing from form group

owenatgov:
Remove the responsive margin from form group so that the bottom margin is 30px across all breakpoints.

Fixes https://github.com/alphagov/govuk-frontend/issues/3924. More details in the issue.

### DIFF
--- a/packages/govuk-frontend/src/govuk/objects/_form-group.scss
+++ b/packages/govuk-frontend/src/govuk/objects/_form-group.scss
@@ -3,7 +3,7 @@
 @include govuk-exports("govuk/objects/form-group") {
   .govuk-form-group {
     @include govuk-clearfix;
-    @include govuk-responsive-margin(6, "bottom");
+    margin-bottom: govuk-spacing(6);
 
     .govuk-form-group:last-of-type {
       margin-bottom: 0; // Remove margin from last item in nested groups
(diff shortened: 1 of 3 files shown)


packages/govuk-frontend/src/govuk/objects/_form-group.scss
```diff
@@ -3,7 +3,7 @@
 @include govuk-exports("govuk/objects/form-group") {
   .govuk-form-group {
     @include govuk-clearfix;
-    @include govuk-responsive-margin(6, "bottom");
+    margin-bottom: govuk-spacing(6);
```
colinrotherham, line 6: It's quite typical to see `.govuk-form-group`, `.govuk-button-group` and `.govuk-button` together

This change leaves two of those with a "responsive" margin still

Can the vertical rhythm drift off-beat for a little while? 🥁

colinrotherham:
Thanks for those tweaks @owenatgov

I've just noticed `.govuk-textarea` which also switches between `20px` and `30px` bottom margin

Although if it's in a `.form-group` you can't tell (margin collapsing)

Worth checking for any others?

owenatgov:
@colinrotherham I did a sweep of components using a `govuk-responsive-margin` value of 6 and it's pretty widespread. I believe responsive 6 this is the standard bottom margin across our components, with a couple of specific exceptions. This feels like we should apply this change to all of them but this feels quite significant. I'm gonna ask the designers about this.

owenatgov:
We're not going to do this. Will update the linked issue with more details momentarily.

Issue #3900, opened 2023-07-03
Review line height of prefix and suffixes for text inputs


owenatgov:
## What
Review the custom line height applied to prefixes and suffixes for text inputs with an aim to remove it if possible.

## Why
[During testing of the govuk-frontend components against the new typography scale](https://github.com/alphagov/govuk-design-system/issues/2289), we noticed that prefixes and suffixes were both now slightly vertically misaligned on small screens. This is due to the CSS rule: `line-height: 1.6` on prefixes and suffixes on small screens. This can be removed once the new typography scale is merged, however there might be a way to future proof this further and refactor prefixes and suffixes to be more robust.

## Who needs to work on this
Developers

## Who needs to review this
Developers, designers

PR #4157, opened 2023-08-31
Dynamically position text within input prefixes and suffixes

owenatgov:
## What/Why
Use flexbox to vertically and horizontally position the text within the input prefix and suffix elements and remove the custom line height being used to nudge the text on smaller screens.

Fixes https://github.com/alphagov/govuk-frontend/issues/3900. Details in the issue.

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/input/_index.scss
+++ b/packages/govuk-frontend/src/govuk/components/input/_index.scss
@@ -118,26 +118,21 @@
   .govuk-input__prefix,
   .govuk-input__suffix {
     @include govuk-font($size: 19);
-
     box-sizing: border-box;
-    display: inline-block;
+    // Use flexbox to align text within the prefix and suffix
+    display: flex;
+    align-items: center;
+    justify-content: center;
     min-width: govuk-px-to-rem(40px);
     height: govuk-px-to-rem(40px);
     padding: govuk-spacing(1);
     border: $govuk-border-width-form-element solid $govuk-input-border-colour;
     background-color: govuk-colour("light-grey");
-
     text-align: center;
-    @include govuk-media-query($until: tablet) {
-      line-height: 1.6;
-    }
     white-space: nowrap;
-
     // Emphasise non-editable status of prefixes and suffixes
     cursor: default;
-
     flex: 0 0 auto;
-
     // Split prefix/suffix onto separate lines on narrow screens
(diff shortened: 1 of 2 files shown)


packages/govuk-frontend/src/govuk/components/input/_index.scss
```diff
@@ -118,26 +118,21 @@
   .govuk-input__prefix,
   .govuk-input__suffix {
     @include govuk-font($size: 19);
-
     box-sizing: border-box;
-    display: inline-block;
+    // Use flexbox to align text within the prefix and suffix
+    display: flex;
+    align-items: center;
```
colinrotherham, line 124: Works great 👍

Will it be an issue that the input still has `line-height: 1.25`?

We've got too vertical alignment methods to keep in sync now

owenatgov: Good spot. From a quick play around this doesn't seem to make a difference for me. I'll do some digging and try to understand what's going on...

owenatgov: Ah silly me. This is coming from `govuk-font`. I'm personally not worried about this as I'm viewing the prefixes and suffixes as purely visual aids rather than necessarily elements of the typography. I'm also not totally sure it makes a difference for the inputs since they're already driven by padding and flexbox.

colinrotherham: If it's not a worry that's fine. Wondered if you wanted the same approach on both?

colinrotherham:
@owenatgov Review wise, are we accepting there's still a very slight misalignment?

>[During testing of the govuk-frontend components against the new typography scale](https://github.com/alphagov/govuk-design-system/issues/2289), we noticed that prefixes and suffixes were both now slightly vertically misaligned on small screens.

It's better than what the issue describes, but close enough to call it done?

Safari macOS

iOS Simulator

querkmachine (approved):
Looks rockin' to me 🤘

owenatgov:
@colinrotherham So it's on record, I think that this alignment is ok. I'm open to being disagreed with if the feelings are strong.

colinrotherham:
@owenatgov Fine by me, was only checking against the "Why" in the issue to see if it was complete

colinrotherham (approved):

PR #3641, opened 2023-05-15
Remove fallback GOV.UK logo for IE8

querkmachine:
Removes the PNG image fallback for the SVG GOV.UK logo. 

This was needed for versions of IE 8 and older, as these browsers do not support SVG images. As we're dropping support for IE8, we can now remove this.

Closes #3624.

PR #4158, opened 2023-09-01
Remove unused `assetsPath` for IE8 fallback GOV.UK logo

colinrotherham:
We no longer use [**Header** `params.assetsPath`](https://design-system.service.gov.uk/components/header/#options-default-1--primary) since we merged:

* https://github.com/alphagov/govuk-frontend/pull/3641

We can remove it from the Nunjucks macro options


36degrees (approved):
Good catch 👍🏻

PR #4170, opened 2023-09-04
Remove incorrect `params.text` Button description etc

colinrotherham:
Documentation for the [Button component](https://design-system.service.gov.uk/components/button/) is quite complicated

As a quick fix I've removed the Button description stating `params.text` has no effect on inputs

We actually use `params.text` in `<input value="{{ params.text }}">`

Plus some other suggested tweaks for where:

1. We sometimes say **"button or link"** to mean the element
1. We sometimes say **"button tag"** versus **"button element"**
1. We sometimes say **"button"** to mean **"button component"**

It's still complicated, but correct

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/button/button.yaml
+++ b/packages/govuk-frontend/src/govuk/components/button/button.yaml
@@ -6,11 +6,11 @@ params:
   - name: text
     type: string
     required: true
-    description: If `html` is set, this is not required. Text for the button or link. If `html` is provided, the `text` option will be ignored and `element` will be automatically set to `button` unless `href` is also set, or it has already been defined.
+    description: If `html` is set, this is not required. Text for the `input`, `button` or `a` element. If `html` is provided, the `text` option will be ignored and `element` will be automatically set to `button` unless `href` is also set, or it has already been defined.
   - name: html
     type: string
     required: true
-    description: If `text` is set, this is not required. HTML for the button or link. If `html` is provided, the `text` option will be ignored and `element` will be automatically set to `button` unless `href` is also set, or it has already been defined. This option has no effect if `element` is set to `input`.
+    description: If `text` is set, this is not required. HTML for the `button` or `a` element only. If `html` is provided, the `text` option will be ignored and `element` will be automatically set to `button` unless `href` is also set, or it has already been defined. This option has no effect if `element` is set to `input`.
   - name: name
     type: string
     required: false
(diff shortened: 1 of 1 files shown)


packages/govuk-frontend/src/govuk/components/button/button.yaml
```diff
@@ -2,15 +2,15 @@ params:
   - name: element
     type: string
     required: false
-    description: Whether to use an `input`, `button` or `a` element to create the button. In most cases you will not need to set this as it will be configured automatically if you use `href` or `html`.
+    description: Whether to use an `input`, `button` or `a` element to create the button component. In most cases you will not need to set this as it will be configured automatically if you use `href` or `html`.
```
domoscargin, line 5: "Whether to use" indicates a boolean to me. Can this be something like

```suggestion
    description: Element to use to create the button component. It can be  `input`, `button` or `a`. In most cases you will not need to set this as it will be configured automatically if you use `href` or `html`.
```

packages/govuk-frontend/src/govuk/components/button/button.yaml
```diff
@@ -2,15 +2,15 @@ params:
...
   - name: text
     type: string
     required: true
-    description: If `html` is set, this is not required. Text for the button or link. If `html` is provided, the `text` option will be ignored and `element` will be automatically set to `button` unless `href` is also set, or it has already been defined. This option has no effect if `element` is set to `input`.
+    description: If `html` is set, this is not required. Text for the `input`, `button` or `a` elements. If `html` is provided, the `text` option will be ignored and `element` will be automatically set to `button` unless `href` is also set, or it has already been defined.
```
domoscargin, line 9: `element` singular? Since it's only one of those three?

domoscargin:
Eish, yeah, just reading all those "applies to this, but not to that" was fiddly. This looks good to me, just a coupla queries.

packages/govuk-frontend/src/govuk/components/button/button.yaml
colinrotherham: Ah I tried to steer clear of changing what was already in there unless it was wrong, but good point

For consistency, shall we follow how the `type` choices are shown?

```suggestion
    description: HTML element for the button component – `input`, `button` or `a`. In most cases you will not need to set this as it will be configured automatically if you use `href` or `html`.
```

packages/govuk-frontend/src/govuk/components/button/button.yaml
colinrotherham: I wasn't totally sure as we have a mix of:

* "For input and button **elements**"
* "…no effect on a or input **elements**"

But sounds better where "for the…" is used now

colinrotherham:
Thanks @domoscargin

Updated with your feedback

Added another fix for something we added in https://github.com/alphagov/govuk-frontend/pull/306 but removed in https://github.com/alphagov/govuk-frontend/pull/683

### Before
>…configured automatically if you use `href` or `html`

### After
>…configured automatically if you use `href`

domoscargin (approved):
🥇

PR #4176, opened 2023-09-05
Throw `ConfigError` when component configs are invalid

### LINKED ISSUE #4037: Throw errors if JavaScript components receive an illegal configuration
## What

Have components receiving an illegal configuration error throw an `IllegalConfigurationError` (extending `GOVUKFrontendError`).

### Affected components

- Character Count – Missing both `maxlength` and `maxwords`

## Why

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [ ] **Review the list of affected components for completeness**
- [ ] Components have been updated to throw errors

domoscargin:
Closes #4037 

## What

I've got two approaches here:

### Validate within the component class
At the moment, we only need this validation on Character Count, so we _could_ just do a specific validation in that class. Might be good enough for now.

### Use a schema
It'd be nice to just use something like AJV and validate a schema, but that'd add weight to our code, so I've implemented a custom validation method on the base component that we can build on as necessary.

It'd also be nice to somehow use our YAML (or compiled JSON fixtures) here, but that's a lot trickier to unpick than is required for this work (for example: how to separate out nunjucks and JS properties, how to document more complex rules like `anyOf`, etc).

---

**Update:** We've gone with the schema approach and loosely followed:
https://ajv.js.org/packages/ajv-errors.html#single-message

### DIFF
--- a/packages/govuk-frontend/src/govuk/all.mjs
+++ b/packages/govuk-frontend/src/govuk/all.mjs
@@ -111,3 +111,9 @@ export {
  * @typedef {import('./components/exit-this-page/exit-this-page.mjs').ExitThisPageTranslations} ExitThisPageTranslations
  * @typedef {import('./components/notification-banner/notification-banner.mjs').NotificationBannerConfig} NotificationBannerConfig
  */
+
+/**
+ * Component config keys, e.g. `accordion` and `characterCount`
+ *
+ * @typedef {keyof Config} ConfigKey
+ */
--- a/packages/govuk-frontend/src/govuk/common/index.mjs
+++ b/packages/govuk-frontend/src/govuk/common/index.mjs
@@ -180,7 +180,21 @@ export function validateConfig(schema, config) {
 }
 
 /**
+ * Schema for component config
+ *
+ * @typedef {object} Schema
+ * @property {SchemaCondition[]} [anyOf] - List of schema conditions
+ */
+
+/**
+ * Schema condition for component config
+ *
+ * @typedef {object} SchemaCondition
+ * @property {string[]} required - List of required config fields
+ * @property {string} errorMessage - Error message when required config fields not provided
+ */
+
+/**
  * @typedef {import('govuk-frontend').Config} Config - Config for all components via `initAll()`
  * @typedef {import('govuk-frontend').ConfigKey} ConfigKey - Component config keys, e.g. `accordion` and `characterCount`
- * @typedef {import('../govuk-frontend-component.mjs').Schema} Schema - Schema for component config
  */
--- a/packages/govuk-frontend/src/govuk/common/index.mjs
+++ b/packages/govuk-frontend/src/govuk/common/index.mjs
@@ -149,12 +149,12 @@ export function isSupported($scope = document.body) {
 }
 
 /**
- * Validates Component config against a schema
+ * Validate component config by schema
  *
- * @protected
- * @param {Schema} schema
- * @param {Config[ConfigKey]} config
- * @returns {string[]} An array of errors, or an empty array if there are no errors
+ * @internal
+ * @param {Schema} schema - Config schema
+ * @param {Config[ConfigKey]} config - Component config
+ * @returns {string[]} List of validation errors
  */
 export function validateConfig(schema, config) {
   /**
@@ -180,7 +180,7 @@ export function validateConfig(schema, config) {
 /**
  * @typedef {import('govuk-frontend').Config} Config - Config for all components via `initAll()`
  * @typedef {import('govuk-frontend').ConfigKey} ConfigKey - Component config keys, e.g. `accordion` and `characterCount`
- * @typedef {import('../govuk-frontend-component.mjs').Schema} Schema - Component config schema
- * @typedef {import('../govuk-frontend-component.mjs').SchemaAnyOf} SchemaAnyOf - Component "Any Of" schema rule
- * @typedef {import('../govuk-frontend-component.mjs').SchemaCondition} SchemaCondition - Schema required array
+ * @typedef {import('../govuk-frontend-component.mjs').Schema} Schema - Schema for component config
+ * @typedef {import('../govuk-frontend-component.mjs').SchemaAnyOf} SchemaAnyOf - Schema rule "Any of" for component config
+ * @typedef {import('../govuk-frontend-component.mjs').SchemaCondition} SchemaCondition - Schema condition for component config
  */
--- a/packages/govuk-frontend/src/govuk/common/index.mjs
+++ b/packages/govuk-frontend/src/govuk/common/index.mjs
@@ -147,3 +147,40 @@ export function extractConfigByNamespace(configObject, namespace) {
 export function isSupported($scope = document.body) {
   return $scope.classList.contains('govuk-frontend-supported')
 }
+
+/**
+ * Validates Component config against a schema
+ *
+ * @protected
+ * @param {Schema} schema
+ * @param {Config[ConfigKey]} config
+ * @returns {string[]} An array of errors, or an empty array if there are no errors
+ */
+export function validateConfig(schema, config) {
+  /**
+   * @type {string[]}
+   */
+  const errors = []
+
+  if (!schema.anyOf) {
+    return errors
+  }
+
+  const satisfiesAnyOf = schema.anyOf.conditions.some(({ required }) =>
+    required.every((prop) => prop in config && config[prop])
+  )
+
+  if (!satisfiesAnyOf) {
+    errors.push(schema.anyOf.errorMessage)
+  }
+
+  return errors
+}
+
+/**
+ * @typedef {import('govuk-frontend').Config} Config - Config for all components via `initAll()`
+ * @typedef {import('govuk-frontend').ConfigKey} ConfigKey - Component config keys, e.g. `accordion` and `characterCount`
+ * @typedef {import('../govuk-frontend-component.mjs').Schema} Schema - Component config schema
+ * @typedef {import('../govuk-frontend-component.mjs').SchemaAnyOf} SchemaAnyOf - Component "Any Of" schema rule
+ * @typedef {import('../govuk-frontend-component.mjs').SchemaCondition} SchemaCondition - Schema required array
+ */
--- a/packages/govuk-frontend/src/govuk/common/index.mjs
+++ b/packages/govuk-frontend/src/govuk/common/index.mjs
@@ -157,30 +157,30 @@ export function isSupported($scope = document.body) {
  * @returns {string[]} List of validation errors
  */
 export function validateConfig(schema, config) {
-  /**
-   * @type {string[]}
-   */
-  const errors = []
+  const validationErrors = []
 
-  if (!schema.anyOf) {
-    return errors
+  // Check errors for each schema
+  for (const [name, conditions] of Object.entries(schema)) {
+    const errors = []
+
+    // Check errors for each schema condition
+    for (const { required, errorMessage } of conditions) {
+      if (!required.every((key) => !!config[key])) {
+        errors.push(errorMessage) // Missing config key value
+      }
+    }
+
+    // Check one condition passes or add errors
+    if (name === 'anyOf' && !(conditions.length - errors.length >= 1)) {
+      validationErrors.push(...errors)
+    }
   }
 
-  const satisfiesAnyOf = schema.anyOf.conditions.some(({ required }) =>
-    required.every((prop) => prop in config && config[prop])
-  )
-
-  if (!satisfiesAnyOf) {
-    errors.push(schema.anyOf.errorMessage)
-  }
-
-  return errors
+  return validationErrors
 }
 
 /**
  * @typedef {import('govuk-frontend').Config} Config - Config for all components via `initAll()`
  * @typedef {import('govuk-frontend').ConfigKey} ConfigKey - Component config keys, e.g. `accordion` and `characterCount`
  * @typedef {import('../govuk-frontend-component.mjs').Schema} Schema - Schema for component config
- * @typedef {import('../govuk-frontend-component.mjs').SchemaAnyOf} SchemaAnyOf - Schema rule "Any of" for component config
- * @typedef {import('../govuk-frontend-component.mjs').SchemaCondition} SchemaCondition - Schema condition for component config
  */
--- a/packages/govuk-frontend/src/govuk/components/character-count/character-count.mjs
+++ b/packages/govuk-frontend/src/govuk/components/character-count/character-count.mjs
@@ -1,6 +1,11 @@
 import { closestAttributeValue } from '../../common/closest-attribute-value.mjs'
-import { extractConfigByNamespace, mergeConfigs } from '../../common/index.mjs'
+import {
+  extractConfigByNamespace,
+  mergeConfigs,
+  validateConfig
+} from '../../common/index.mjs'
 import { normaliseDataset } from '../../common/normalise-dataset.mjs'
+import { ConfigError } from '../../errors/index.mjs'
 import { GOVUKFrontendComponent } from '../../govuk-frontend-component.mjs'
 import { I18n } from '../../i18n.mjs'
 
@@ -108,7 +113,11 @@ export class CharacterCount extends GOVUKFrontendComponent {
       datasetConfig
     )
 
-    this.checkConfig(CharacterCount.schema, this.config)
+    // Check for valid config
+    const errors = validateConfig(CharacterCount.schema, this.config)
+    if (errors[0]) {
+      throw new ConfigError(`Character count: ${errors[0]}`)
+    }
 
     this.i18n = new I18n(extractConfigByNamespace(this.config, 'i18n'), {
       // Read the fallback if necessary rather than have it set in the defaults
@@ -529,6 +538,6 @@ export class CharacterCount extends GOVUKFrontendComponent {
  */
 
 /**
- * @typedef {import('../../govuk-frontend-component.mjs').Schema} Schema
+ * @typedef {import('../../common/index.mjs').Schema} Schema
  * @typedef {import('../../i18n.mjs').TranslationPluralForms} TranslationPluralForms
  */
--- a/packages/govuk-frontend/src/govuk/components/character-count/character-count.mjs
+++ b/packages/govuk-frontend/src/govuk/components/character-count/character-count.mjs
@@ -62,6 +62,21 @@ export class CharacterCount extends GOVUKFrontendComponent {
   /** @private */
   maxLength = Infinity
 
+  /** @private */
+  configSchema = {
+    anyOf: {
+      conditions: [
+        {
+          required: ['maxwords']
+        },
+        {
+          required: ['maxlength']
+        }
+      ],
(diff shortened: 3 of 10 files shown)


packages/govuk-frontend/src/govuk/govuk-frontend-component.mjs
```diff
@@ -29,4 +29,37 @@ export class GOVUKFrontendComponent {
...
   }
+
+  /**
+   *
+   * @protected
+   * @param {Schema} schema
+   * @param {object} config
+   * @returns {boolean} Whether the config is valid
+   */
+  validateConfig(schema, config) {
```
romaricpascal, line 40: I like the direction of this 😍

colinrotherham: The idea for `super.validateConfig()` is brill

Probably in the same way we've suggested `@augments` and generics in https://github.com/alphagov/govuk-frontend/pull/4104 I'd prefer to type what goes in/out of these new base methods?

Trying to avoid abstracting away type-safe things that spot errors for us into loose wildcard `{object}` or `{*}` types which may let errors go through undetected

I had a go at setting up JSDoc comments for component options + config in https://github.com/alphagov/govuk-frontend/commit/5b0a3d3294f401b17d48fb4e0124a5c65e898a16, but couldn't find a nice way to avoid duplicating all the descriptions from the macro options

@romaricpascal Didn't you find a utility that could extract types into JavaScript objects?

domoscargin: The stumbling block I'm hitting (which is probably more of a "look at this later" thing) is I'd like to avoid having something like our YAML, some type definitions, and a schema.

Would be nice if all of that was wrapped up in a single source of truth.

colinrotherham: Yeah, the commit was from the "Component options as ES Modules" work I started

We could definitely keep schemas in there

packages/govuk-frontend/src/govuk/govuk-frontend-component.mjs
```diff
@@ -29,4 +29,37 @@ export class GOVUKFrontendComponent {
...
+    if (schema.anyOf) {
+      // Check if at least one of the conditions in anyOf is met
+      for (const condition of schema.anyOf) {
+        const requiredProps = condition.required || []
+        if (requiredProps.every((prop) => prop in config)) {
```
colinrotherham, line 45: We do "truthy" checks in the component

```js
'maxwords' in this.config && this.config.maxwords
```

So would that be this?

```suggestion
        if (requiredProps.every((prop) => prop in config && config[prop])) {
```

packages/govuk-frontend/src/govuk/govuk-frontend-component.mjs
```diff
@@ -29,4 +29,37 @@ export class GOVUKFrontendComponent {
...
+      }
+      return false
+    }
+    return true
+  }
 }
+
+/**
+ * @typedef {object} Schema
+ * @property {Array<SchemaItem>} [anyOf] - An array of conditions
```
colinrotherham, line 57: I'm sure ESLint used to ask for shorthand at some point?

Would be this if you prefer it:

```suggestion
 * @property {SchemaItem[]} [anyOf] - An array of conditions
```

packages/govuk-frontend/src/govuk/govuk-frontend-component.mjs
domoscargin: Has there been any movement on https://github.com/alphagov/govuk-frontend/pull/3812 since June? I see @romaricpascal suggested running it past the team. What was the reaction? I like the general direction at a glance

packages/govuk-frontend/src/govuk/govuk-frontend-component.mjs
domoscargin: Ugh, I do prefer it, I need to change my snippets. But eslint not complaining for me.

packages/govuk-frontend/src/govuk/govuk-frontend-component.mjs
colinrotherham: Yeah it's what powered the JavaScript autocomplete for `govukAccordion()` etc in the demo this week. I've also been using it to find lots of the macro options bugs and issues 😎

Feedback wise, a couple of people liked the simplicity of the YAML

Some of it was covered in the previous sprint's Show & Tell but I'll feature it more in the next one

packages/govuk-frontend/src/govuk/govuk-frontend-component.mjs
```diff
@@ -29,4 +29,37 @@ export class GOVUKFrontendComponent {
...
+      return false
+    }
+    return true
+  }
 }
+
+/**
+ * @typedef {object} Schema
+ * @property {SchemaItem[]} [anyOf] - An array of conditions
+ * @property {*} [otherProperties] - Other properties that may exist in the schema.
```
colinrotherham, line 58: Mind if we drop `otherProperties` if it's unused? Save it for the PR that adds more properties

packages/govuk-frontend/src/govuk/govuk-frontend-component.mjs
```diff
@@ -29,4 +29,37 @@ export class GOVUKFrontendComponent {
       throw new SupportError()
     }
   }
+
+  /**
+   *
+   * @protected
+   * @param {Schema} schema
+   * @param {{ [key: string]: number | string | object | undefined}} config
```
colinrotherham, line 37: TypeScript won't help here as you'll see that `config` has become type `any`

Problem comes from adding the type `object` which is always awful in JavaScript 😭

```
typeof [] === 'object'
typeof null === 'object'
typeof new Map() === 'object'
```

Thankfully we have a "component configs" type already, just needs moving

I’d suggest we [move `ConfigKey`](https://github.com/alphagov/govuk-frontend/blob/main/shared/helpers/puppeteer.js#L257) into [**packages/govuk-frontend/src/govuk/all.mjs**](https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/src/govuk/all.mjs#L88) giving us:

```mjs
/**
 * @typedef {import('govuk-frontend').Config} Config - Config for all components via `initAll()`
 * @typedef {import('govuk-frontend').ConfigKey} ConfigKey - Component config keys, e.g. `accordion` and `characterCount`
 */
```

As an example, you'll see from the [`ConfigKey`](https://github.com/alphagov/govuk-frontend/blob/main/shared/helpers/puppeteer.js#L257) tooltip the available configs:

Tooltip for component config types

<br>Which we can use here too:

```suggestion
   * @param {Config[ConfigKey]} config
```

packages/govuk-frontend/src/govuk/govuk-frontend-component.mjs
```diff
@@ -29,4 +29,37 @@ export class GOVUKFrontendComponent {
       throw new SupportError()
     }
   }
+
+  /**
+   *
```
colinrotherham, line 34:
```suggestion
```

packages/govuk-frontend/src/govuk/govuk-frontend-component.mjs
colinrotherham: Hopefully we can get generics working with JSDoc and narrow those configs down by component

packages/govuk-frontend/src/govuk/govuk-frontend-component.mjs
```diff
@@ -29,4 +29,37 @@ export class GOVUKFrontendComponent {
...
+      // Check if at least one of the conditions in anyOf is met
+      for (const condition of schema.anyOf) {
+        const requiredProps = condition.required
+        if (requiredProps.every((prop) => prop in config && config[prop])) {
+          return true
+        }
+      }
+      return false
+    }
+    return true
```
colinrotherham, line 51: Are we able to reduce these 3x returns for readability?

Maybe using `.some()` for the "at least one" instead of a loop

Happy to help look

packages/govuk-frontend/src/govuk/components/character-count/character-count.mjs
```diff
@@ -108,6 +121,12 @@ export class CharacterCount extends GOVUKFrontendComponent {
       datasetConfig
     )
 
+    if (!this.validateConfig(this.configSchema, this.config)) {
```
colinrotherham, line 124: This line is brill. Do we need to error on unknown properties too?

I'm assuming `{ maxwords: 20, colin: true }` will be valid

domoscargin: Ideally, yes - but was holding off till we've figured out how to do our config schema/types.

Or do you think it's best to define the entire schema now within the component and move it later?

colinrotherham: I'm happy to leave it for another PR

packages/govuk-frontend/src/govuk/govuk-frontend-component.mjs
```diff
@@ -29,4 +29,35 @@ export class GOVUKFrontendComponent {
...
     }
   }
+
+  /**
+   * @protected
+   * @param {Schema} schema
+   * @param {Config[ConfigKey]} config
+   * @returns {boolean} Whether the config is valid
+   */
+  validateConfig(schema, config) {
```
colinrotherham, line 39: @domoscargin Do we need to follow `checkSupport()` here?

For example:

1. Add an `isConfigValid()` utility to return true/false
2. Use method `this.checkConfig()` to throw an error

packages/govuk-frontend/src/govuk/errors/index.mjs
```diff
@@ -40,3 +40,10 @@ export class SupportError extends GOVUKFrontendError {
     super('GOV.UK Frontend is not supported in this browser')
   }
 }
+
+/**
+ * Indicates that a component has received an illegal configuration
+ */
+export class IllegalConfigurationError extends GOVUKFrontendError {
```
colinrotherham, line 47: Similar to how `TypeError` isn't `IllegalTypeError`, should we flip out the negative?

Some ideas:

```
ConfigError
ConfigValidationError
```

Probably use "Config" versus "Configuration" too to keep it consistent?

packages/govuk-frontend/src/govuk/govuk-frontend-component.mjs
```diff
@@ -29,4 +29,35 @@ export class GOVUKFrontendComponent {
...
 }
+
+/**
+ * @typedef {object} Schema
+ * @property {SchemaItem[]} [anyOf] - An array of conditions
+ */
+
+/**
+ * @typedef {object} SchemaItem
+ * @property {string[]} [required] - Whether the current item is required
```
colinrotherham, line 57: Think we best make `required` mandatory

Otherwise we'll potentially crash on `required.every()` above

```suggestion
 * @property {string[]} required - Whether the current item is required
```

packages/govuk-frontend/src/govuk/components/character-count/character-count.mjs
```diff
@@ -62,6 +63,18 @@ export class CharacterCount extends GOVUKFrontendComponent {
   /** @private */
   maxLength = Infinity
 
+  /** @private */
+  configSchema = {
+    anyOf: [
+      {
+        required: ['maxwords']
```
colinrotherham, line 70: Probably need to put the error message alongside the validation rule?

Otherwise from a boolean `this.validateConfig()` how do we know we rule was broken?

```suggestion
        required: ['maxwords'],
        errorMessage: 'Either `maxlength` or `maxwords` must be provided'
```

packages/govuk-frontend/src/govuk/components/character-count/character-count.mjs
```diff
@@ -108,6 +121,12 @@ export class CharacterCount extends GOVUKFrontendComponent {
       datasetConfig
     )
 
+    if (!this.validateConfig(this.configSchema, this.config)) {
+      throw new IllegalConfigurationError(
+        'Character Count: either `maxlength` or `maxwords` must be provided'
```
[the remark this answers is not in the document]
romaricpascal: If they're both missing, we have no way to know which the user intended, I think. We can only refer them to the docs, maybe? Another one to remember when discussing #4072.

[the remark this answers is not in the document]
colinrotherham: Sorry @romaricpascal I mean when we expand the schema to include other rules

What if the validator finds 2x rules fail, which one do we display?

packages/govuk-frontend/src/govuk/components/character-count/character-count.mjs
```diff
@@ -62,6 +63,18 @@ export class CharacterCount extends GOVUKFrontendComponent {
   /** @private */
   maxLength = Infinity
 
+  /** @private */
+  configSchema = {
+    anyOf: [
```
colinrotherham, line 68: To honour "at least one, but not both" I think this should be `oneOf`?

```suggestion
    oneOf: [
```

domoscargin: The way I read it, it IS currently valid to provide both - one will just be ignored.

packages/govuk-frontend/src/govuk/components/character-count/character-count.mjs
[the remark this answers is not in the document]
domoscargin: > Sorry @romaricpascal I mean when we expand the schema to include other rules
> 
> What if the validator finds 2x rules fail, which one do we display?

That's for the next chump to figure out...

packages/govuk-frontend/src/govuk/components/character-count/character-count.mjs
colinrotherham: Yeah in the code it is, but did we mean that?? 😬

domoscargin: Yours is definitely the sensible take, but I think to enforce it would be a breaking change, since services might have both properties right now, and we should obviously change the property descriptions.

Basically, do we need eyes from a non-dev? @romaricpascal, whaddya think?

packages/govuk-frontend/src/govuk/components/character-count/character-count.mjs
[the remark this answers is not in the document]
domoscargin: A sound plan!

packages/govuk-frontend/src/govuk/components/character-count/character-count.mjs
colinrotherham: Looking at the documentation you're right, breaking change:
https://frontend.design-system.service.gov.uk/javascript-api-reference/#charactercount

### maxlength
> The maximum number of characters. If `maxwords` is provided, the `maxlength` option will be ignored.

### maxwords
> The maximum number of words. If `maxwords` is provided, the `maxlength` option will be ignored.

packages/govuk-frontend/src/govuk/all.mjs
```diff
@@ -119,3 +119,17 @@ export {
...
+ * @typedef {object} Schema
+ * @property {SchemaItem[]} [anyOf] - An array of conditions
+ */
+
+/**
+ * Component schema fields
+ *
+ * @typedef {object} SchemaItem
+ * @property {string[]} required - Whether the current item is required
+ */
```
colinrotherham, line 135: We may want to keep these types in the base class if we consider them private or internal

Works great where they are but aware they'll go into the JSDoc HTML output:

[/docs/javascript/interfaces/**Schema.html**](https://govuk-frontend-pr-4176.herokuapp.com/docs/javascript/interfaces/Schema.html)
[/docs/javascript/interfaces/**SchemaItem.html**](https://govuk-frontend-pr-4176.herokuapp.com/docs/javascript/interfaces/SchemaItem.html)

**Tip:** To purposely hide things that TypeDoc has discovered we can use **typedoc.config.js** like:

https://github.com/alphagov/govuk-frontend/blob/1d1fb0a535062128556a669b7ae391e509f030d6/packages/govuk-frontend-review/typedoc.config.js#L29-L30

packages/govuk-frontend/src/govuk/components/character-count/character-count.mjs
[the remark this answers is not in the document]
domoscargin: Popped an implementation in the latest commit.

packages/govuk-frontend/src/govuk/common/index.mjs
```diff
@@ -149,19 +149,32 @@ export function isSupported($scope = document.body) {
...
+    return errors
   }
 
-  return schema.anyOf.some(({ required }) =>
+  const satisfiesAnyOf = schema.anyOf.some(({ required }) =>
     required.every((prop) => prop in config && config[prop])
   )
+
+  if (!satisfiesAnyOf) {
+    errors.push('Either `maxlength` or `maxwords` must be provided')
```
colinrotherham, line 174: Great minds, this is where I'd pluck that example

```suggestion
    errors.push(schema.anyOf.errorMessage)
```

domoscargin: Pushed up this change.

packages/govuk-frontend/src/govuk/components/character-count/character-count.mjs
colinrotherham: Resolved in https://github.com/alphagov/govuk-frontend/pull/4183

romaricpascal: Yup, let's match the docs and keep `anyOf` as both could be set and we deal with it OK 😊 Nice little tidy up in #4183, makes both types and code simpler.

romaricpascal (approved):
It does a bit more than we initialy planned, but it shows a direction we may want to take towards validating more generally in the future. As the validations are internal anyway, we can rework them to our heart's content should we need to adjust later on (as long as we throw the appropriate error) 😊

packages/govuk-frontend/src/govuk/common/index.mjs
```diff
@@ -147,3 +147,54 @@ export function extractConfigByNamespace(configObject, namespace) {
...
+ * @param {Schema} schema - Config schema
+ * @param {Config[ConfigKey]} config - Component config
+ * @returns {string[]} List of validation errors
+ */
+export function validateConfig(schema, config) {
+  const errors = []
+
+  // Check errors for each schema
+  for (const [name, conditions] of Object.entries(schema)) {
+    const schemaErrors = []
```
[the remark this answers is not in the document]
colinrotherham: I did contemplate `conditionErrors`?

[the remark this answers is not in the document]
romaricpascal: Works as well 😊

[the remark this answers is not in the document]
colinrotherham: I've changed my mind

In the loop we have `errorMessage` so I'll pair that with `errors`

```mjs
errors.push(errorMessage) // Missing config key value
```

Easier then to rename the top-level to ~`errors`~ `validationError` to pair with `validateConfig`

PR #4177, opened 2023-09-05
Add MissingElementError and use it within the Skip Link

### LINKED ISSUE #4128: Throw errors during Skip Link initialisation if key HTML elements are missing
## What

- The linked element

## Why

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

domoscargin:
Closes #4128 

Throws an error instead of returning if the linked element is missing.

I considered implementing a separate method to check for this, but that feels a bit needless for Skip Link - it may be necessary for more complex components, though.

There's also an argument for throwing errors _within_ `getLinkedElement` and `getFragmentFromUrl` (our error message could potentially be more helpful if we could provide the `linkedElementId`, for example), but since these get caught by the existing check, I kept things simple.

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/skip-link/skip-link.mjs
+++ b/packages/govuk-frontend/src/govuk/components/skip-link/skip-link.mjs
@@ -1,3 +1,4 @@
+import { MissingElementError } from '../../errors/index.mjs'
 import { GOVUKFrontendComponent } from '../../govuk-frontend-component.mjs'
 
 /**
@@ -33,7 +34,7 @@ export class SkipLink extends GOVUKFrontendComponent {
     // Check for linked element
     const $linkedElement = this.getLinkedElement()
     if (!$linkedElement) {
-      return
+      throw new MissingElementError('The linked HTML element does not exist')
     }
 
     this.$linkedElement = $linkedElement
--- a/packages/govuk-frontend/src/govuk/components/skip-link/skip-link.mjs
+++ b/packages/govuk-frontend/src/govuk/components/skip-link/skip-link.mjs
@@ -21,6 +21,7 @@ export class SkipLink extends GOVUKFrontendComponent {
 
   /**
    * @param {Element} $module - HTML element to use for skip link
+   * @throws {MissingElementError} If the element with the specified ID is not found
    */
   constructor($module) {
     super()
@@ -35,12 +36,11 @@ export class SkipLink extends GOVUKFrontendComponent {
     try {
       const $linkedElement = this.getLinkedElement()
       this.$linkedElement = $linkedElement
-    } catch (cause) {
+    } catch (error) {
       throw new MissingElementError(
-        'Skip link: the linked HTML element does not exist',
-        {
-          cause: cause instanceof Error ? cause : undefined
-        }
+        `Skip link: ${
+          error instanceof Error ? error.message : 'Linked element not found'
+        }`
       )
     }
 
@@ -52,22 +52,20 @@ export class SkipLink extends GOVUKFrontendComponent {
    *
    * @private
    * @throws {Error} If the "href" attribute does not contain a hash
-   * @throws {Error} If the element with the specified ID does not exist
+   * @throws {TypeError} If the element with the specified ID is not found
    * @returns {HTMLElement} $linkedElement - DOM element linked to from the skip link
    */
   getLinkedElement() {
     const linkedElementId = this.getFragmentFromUrl()
     if (!linkedElementId) {
-      throw new Error(
-        `Skip link: $module "href" attribute does not contain a hash`
-      )
+      throw new Error(`$module "href" attribute does not contain a hash`)
     }
 
     const linkedElement = document.getElementById(linkedElementId)
 
     if (!linkedElement) {
-      throw new Error(
-        `Skip link: Target selector "#${linkedElementId}" not found`
+      throw new TypeError(
+        `Linked element selector "#${linkedElementId}" not found`
       )
     }
 
--- a/packages/govuk-frontend/src/govuk/components/skip-link/skip-link.mjs
+++ b/packages/govuk-frontend/src/govuk/components/skip-link/skip-link.mjs
@@ -32,12 +32,18 @@ export class SkipLink extends GOVUKFrontendComponent {
     this.$module = $module
 
     // Check for linked element
-    const $linkedElement = this.getLinkedElement()
-    if (!$linkedElement) {
-      throw new MissingElementError('The linked HTML element does not exist')
+    try {
+      const $linkedElement = this.getLinkedElement()
+      this.$linkedElement = $linkedElement
+    } catch (cause) {
+      throw new MissingElementError(
+        'Skip link: the linked HTML element does not exist',
+        {
+          cause: cause instanceof Error ? cause : undefined
+        }
+      )
     }
 
-    this.$linkedElement = $linkedElement
     this.$module.addEventListener('click', () => this.focusLinkedElement())
   }
 
@@ -45,15 +51,27 @@ export class SkipLink extends GOVUKFrontendComponent {
    * Get linked element
    *
    * @private
-   * @returns {HTMLElement | null} $linkedElement - DOM element linked to from the skip link
+   * @throws {Error} If the "href" attribute does not contain a hash
+   * @throws {Error} If the element with the specified ID does not exist
+   * @returns {HTMLElement} $linkedElement - DOM element linked to from the skip link
    */
   getLinkedElement() {
     const linkedElementId = this.getFragmentFromUrl()
     if (!linkedElementId) {
-      return null
+      throw new Error(
+        `Skip link: $module "href" attribute does not contain a hash`
+      )
     }
 
-    return document.getElementById(linkedElementId)
+    const linkedElement = document.getElementById(linkedElementId)
+
+    if (!linkedElement) {
+      throw new Error(
+        `Skip link: Target selector "#${linkedElementId}" not found`
+      )
+    }
+
+    return linkedElement
   }
 
   /**
--- a/packages/govuk-frontend/src/govuk/errors/index.mjs
+++ b/packages/govuk-frontend/src/govuk/errors/index.mjs
@@ -23,6 +23,13 @@ export class GOVUKFrontendError extends Error {
 }
 
 /**
+ * Indicates that a Component's required HTML element is missing
+ */
+export class MissingElementError extends GOVUKFrontendError {
+  name = 'MissingElementError'
+}
+
+/**
  * Indicates that GOV.UK Frontend is not supported
  */
 export class SupportError extends GOVUKFrontendError {
--- a/packages/govuk-frontend/src/govuk/components/skip-link/skip-link.test.js
+++ b/packages/govuk-frontend/src/govuk/components/skip-link/skip-link.test.js
@@ -92,7 +92,22 @@ describe('Skip Link', () => {
         })
       ).rejects.toEqual({
         name: 'MissingElementError',
-        message: 'Skip link: the linked HTML element does not exist'
+        message:
+          'Skip link: Linked element selector "#this-element-does-not-exist" not found'
+      })
+    })
+
+    it('throws when the href does not contain a hash', async () => {
+      await expect(
+        renderAndInitialise(page, 'skip-link', {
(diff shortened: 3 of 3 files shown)


colinrotherham:
> There's also an argument for throwing errors _within_ `getLinkedElement` and `getFragmentFromUrl` (our error message could potentially be more helpful if we could provide the `linkedElementId`, for example), but since these get caught by the existing check, I kept things simple.

Not a blocker in the slightest, but knowing what's missing is kinda useful!

Could throw (and catch) an error in `this.getLinkedElement()` but add to [`{ cause }`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause) for the developer tools?

```mjs
try {
  this.$linkedElement = this.getLinkedElement()
} catch (cause) {
  throw new MissingElementError('The linked HTML element does not exist', {
    cause
  })
}
```

Would be helpful to see what `linkedElementId` was for sure

colinrotherham (approved):
Approved! Added a comment about an error cause

For any error thrown it's good to explain what caused it

domoscargin:
@colinrotherham Added a commit that addresses your comment. As discussed, I've had to check the type of `cause`.

packages/govuk-frontend/src/govuk/components/skip-link/skip-link.mjs
```diff
@@ -31,28 +32,43 @@ export class SkipLink extends GOVUKFrontendComponent {
...
-      return
+    try {
+      const $linkedElement = this.getLinkedElement()
+      this.$linkedElement = $linkedElement
+    } catch (errorCause) {
+      const cause =
+        errorCause instanceof Error ? errorCause : new Error(String(errorCause))
+      throw new MissingElementError('The linked HTML element does not exist', {
+        cause
+      })
```
colinrotherham, line 43: Mind if we simplify this a bit? TypeScript doesn't know we only throw `Error` but we do

So can either leave the right hand side as `undefined` because it's never used

```suggestion
    } catch (cause) {
      throw new MissingElementError('The linked HTML element does not exist', {
        cause: cause instanceof Error ? cause : undefined
      })
```

colinrotherham: Feels a shame to add extra code so could ignore `@typescript-eslint/no-unsafe-assignment` too

packages/govuk-frontend/src/govuk/components/skip-link/skip-link.mjs
```diff
@@ -31,28 +32,43 @@ export class SkipLink extends GOVUKFrontendComponent {
...
+   * @throws {Error} If the "href" attribute does not contain a hash
+   * @throws {Error} If the element with the specified ID does not exist
+   * @returns {HTMLElement} $linkedElement - DOM element linked to from the skip link
    */
   getLinkedElement() {
     const linkedElementId = this.getFragmentFromUrl()
     if (!linkedElementId) {
-      return null
+      throw new Error(
+        `The Skip Link's "href" attribute does not contain a hash`
```
colinrotherham, line 61: Should we match "The link" to ["The linked" wording above](https://github.com/alphagov/govuk-frontend/pull/4177/files#diff-fb3a25d42cbb3621d375cc34dc960447e84fa2a8729fd16e97e4c34edd9b17aeR41)

Or should we always add a component name?

```suggestion
        `The link "href" attribute does not contain a hash`
```

```suggestion
        `Skip link: $module "href" attribute does not contain a hash`
```

domoscargin: A bit of bike shedding, since we'll do a content review of these, but I think the second option is more immediately understandable

packages/govuk-frontend/src/govuk/components/skip-link/skip-link.mjs
```diff
@@ -31,28 +32,43 @@ export class SkipLink extends GOVUKFrontendComponent {
...
+      )
+    }
+
+    const linkedElement = document.getElementById(linkedElementId)
+
+    if (!linkedElement) {
+      throw new Error(`Element with ID "${linkedElementId}" does not exist`)
```
colinrotherham, line 68: Might be worth making these generic?

Mainly so monitor tools that scrape errors can contain `Target selector "${XXX}" not found`

```suggestion
      throw new Error(`Skip link: Target selector "#${linkedElementId}" not found`)
```

packages/govuk-frontend/src/govuk/errors/index.mjs
```diff
@@ -22,6 +22,13 @@ export class GOVUKFrontendError extends Error {
   name = 'GOVUKFrontendError'
 }
 
+/**
+ * Indicates that a Component's required HTML element is missing
+ */
+export class MissingElementError extends GOVUKFrontendError {
+  name = 'MissingElementError'
+}
```
colinrotherham, line 30: Not a blocker, but it's triggered a discussion

To keep file size down, could we use Error `{ cause }` to keep our custom errors to a minimum? Otherwise our published package will grow as custom errors are added

E.g. Using [**TypeError** "…not of the expected type"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypeError) for `null` or missing things

### Component errors
Here's an example with nested `{ cause }` to make the error more informative:

```mjs
throw new ComponentError('I has died', {
  cause: new TypeError("Argh, 'ere be no link target #pirates")
}
```

But using the nested `{ cause }` approach to surface all that useful information. It's handy for monitoring tools like [Sentry](https://sentry.io/welcome/) which links chained errors + stack traces

For extra user friendliness a custom `toString()` (see below) could log the component name:

>ComponentError: "Skip link: I has died"
> → Cause: TypeError: "Argh, 'ere be no link target #pirates"

```suggestion
export class ComponentError extends GOVUKFrontendError {
  name = 'ComponentError'
  component = 'Component name'

  toString() {
    return `${this.component}: ${this.message}`;
  }
}
```

colinrotherham: There's also a gotcha

We'd need to confirm Babel transforms or a polyfill for [JavaScript built-in: Error: cause](https://caniuse.com/mdn-javascript_builtins_error_cause)

colinrotherham (approved):

packages/govuk-frontend/src/govuk/components/skip-link/skip-link.mjs
```diff
@@ -31,28 +32,46 @@ export class SkipLink extends GOVUKFrontendComponent {
...
-    if (!$linkedElement) {
-      return
+    try {
+      const $linkedElement = this.getLinkedElement()
+      this.$linkedElement = $linkedElement
+    } catch (cause) {
+      throw new MissingElementError(
+        'Skip link: the linked HTML element does not exist',
+        {
+          cause: cause instanceof Error ? cause : undefined
```
colinrotherham, line 42: For [JavaScript built-in: Error: cause](https://caniuse.com/mdn-javascript_builtins_error_cause) browser support are we happy treating this as optional?

If so, do you mind adding a bit of safety to the `GOVUKFrontendError` constructor?

i.e. Use feature detection, not a polyfill

```mjs
export class GOVUKFrontendError extends Error {
  name = 'GOVUKFrontendError'

  /**
   * @param {string} message 
   * @param {ErrorOptions} options
   */
  constructor(message, { cause, ...options }) {
    super(message, options)

    // Add cause property where supported
    if (cause && new Error('test', { cause }).cause === cause) {
      this.cause = cause
    }
  }
}
```

We'll need to:

1. ESLint ignore "ES2022 Error Cause is forbidden" for feature detection purposes
2. ESLint ignore "ES2018 rest/spread properties are forbidden" if Babel handles it already
3. Triple check my feature detection idea, maybe it needs a try/catch?

For 3) it's mainly because `'cause' in Error.prototype` didn't work 🙈

colinrotherham: Or we remove `{ cause }` for another day?

packages/govuk-frontend/src/govuk/components/skip-link/skip-link.mjs
```diff
@@ -31,28 +32,46 @@ export class SkipLink extends GOVUKFrontendComponent {
...
-      return null
+      throw new Error(
+        `Skip link: $module "href" attribute does not contain a hash`
+      )
+    }
+
+    const linkedElement = document.getElementById(linkedElementId)
+
+    if (!linkedElement) {
+      throw new Error(
```
colinrotherham, line 69: Shall we try and match up the browser errors?

```suggestion
      throw new TypeError(
```

packages/govuk-frontend/src/govuk/components/skip-link/skip-link.mjs
```diff
@@ -31,28 +32,46 @@ export class SkipLink extends GOVUKFrontendComponent {
...
 
     // Check for linked element
-    const $linkedElement = this.getLinkedElement()
-    if (!$linkedElement) {
-      return
+    try {
+      const $linkedElement = this.getLinkedElement()
+      this.$linkedElement = $linkedElement
+    } catch (cause) {
+      throw new MissingElementError(
```
colinrotherham, line 39: Do we need to document `@throws` in the constructor?

Unlike for `getLinkedElement()` (private method) it'll appear in our JSDoc HTML for this one

packages/govuk-frontend/src/govuk/components/skip-link/skip-link.mjs
domoscargin: I'm leaning toward leaving `{ cause }` for now - feels like p'raps we need a bit more thinking about how we make it work and can do that as part of iterating on these errors.

I've pushed a change to use the error messages to convey the error to the user, whaddya reckon?

packages/govuk-frontend/src/govuk/components/skip-link/skip-link.mjs
```diff
@@ -31,28 +33,41 @@ export class SkipLink extends GOVUKFrontendComponent {
...
     const linkedElementId = this.getFragmentFromUrl()
     if (!linkedElementId) {
-      return null
+      throw new Error(`$module "href" attribute does not contain a hash`)
+    }
+
+    const linkedElement = document.getElementById(linkedElementId)
+
+    if (!linkedElement) {
+      throw new TypeError(`Target selector "#${linkedElementId}" not found`)
```
colinrotherham, line 67: Not a blocker, but we now say "linked element" in the fallback error but "target selector" here

```suggestion
      throw new TypeError(`Linked element selector "#${linkedElementId}" not found`)
```

packages/govuk-frontend/src/govuk/components/skip-link/skip-link.mjs
```diff
@@ -31,28 +33,41 @@ export class SkipLink extends GOVUKFrontendComponent {
     this.$module = $module
 
     // Check for linked element
-    const $linkedElement = this.getLinkedElement()
-    if (!$linkedElement) {
-      return
+    try {
+      const $linkedElement = this.getLinkedElement()
+      this.$linkedElement = $linkedElement
+    } catch (err) {
```
colinrotherham, line 39: Another non-blocker, but we use `error` everywhere else

```suggestion
    } catch (error) {
```

colinrotherham (approved):
Brilliant @domoscargin

Some non-blocking things since the last push but approved! Thanks again for making the errors actionable, explaining why and error was throw

Appreciate we've used double quotes here and back ticks in [#4104](https://github.com/alphagov/govuk-frontend/pull/4104#discussion_r1318597483) so I've written up some of the content feedback for later:

* https://github.com/alphagov/govuk-frontend/pull/4177

## Error causes
We had some good chats about error causes, especially with monitoring tools like [Sentry](https://sentry.io/welcome/) linking chained errors + stack traces together

So we should pull out some actions/discussions such as:

1. Using [Error `cause`](https://caniuse.com/mdn-javascript_builtins_error_cause) to reduce custom errors ([see comment](https://github.com/alphagov/govuk-frontend/pull/4177#discussion_r1318526224))
2. Spike [Error `cause`](https://caniuse.com/mdn-javascript_builtins_error_cause) feature detection ([see comment](https://github.com/alphagov/govuk-frontend/pull/4177#pullrequestreview-1615338937))

PR #4182, opened 2023-09-08
Remove stray `.only()` from PR #4030

colinrotherham:
Accidentally added in:

* https://github.com/alphagov/govuk-frontend/pull/4030


romaricpascal (approved):
Argh, sorry 😔

PR #4183, opened 2023-09-08
Update character count types to show `maxlength` can be ignored

colinrotherham:
We’d previously assumed Character count `maxlength` and `maxwords` were mutually exclusive

But in our [**Character count** JavaScript API reference](https://frontend.design-system.service.gov.uk/javascript-api-reference/#charactercount) we say both can be provided:

> If maxwords is provided, the maxlength option will be ignored.

This PR updates the config types to show that both can be provided to resolve https://github.com/alphagov/govuk-frontend/pull/4176#discussion_r1318846115


romaricpascal (approved):
That's making it all simpler 🥳

PR #4184, opened 2023-09-08
Configure ESLint for required JSDoc `@param` descriptions

colinrotherham:
This PR ensures we don't forget JSDoc `@param` descriptions during our API work https://github.com/alphagov/govuk-frontend/issues/3478

We try to stick to the following rules:

1. JSDoc blocks are optional by default for the project
2. JSDoc blocks are mandatory in `govuk-frontend` source code

But when either optional or mandatory JSDoc blocks are used, they must:

* have `@param` by default for the project
* have `@param` descriptions in `govuk-frontend` source code

Which ensures our JSDoc JavaScript API documentation is populated:  
https://govuk-frontend-review.herokuapp.com/docs/javascript/

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -60,10 +60,13 @@ module.exports = {
           }
         ],
 
-        // JSDoc blocks are optional
+        // JSDoc blocks are optional by default
         'jsdoc/require-jsdoc': 'off',
+
+        // JSDoc @param required in (optional) blocks but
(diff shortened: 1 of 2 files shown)


romaricpascal (approved):
Happy to switch it on, but let's keep an eye on where we just write `@param {Type} <paramName> - The <paramName>` to make sure we get some value from the descriptions 😊

Issue #4175, opened 2023-09-05
Decide how much we want to change the favicon


christopherthomasdesign:
## What

As we're updating the crown, now would be a good time to revisit the favicon. This should be a short investigation to work out the best way forward.

The simplest version of this is to update the existing graphic to use the new crown. But there are other possibilities – some browsers now support SVG for favicons, and we could generally do some work to make sure our favicon implementation is up to date with modern standards. 

We could also simplify the design of the crown to work better at very small sizes (to use only in the favicon). But we're not currently sure if that's something we can do.

## Why

* To create consistency with the new crown
* Potentially, to improve performance and visual clarity over the current solution

## Who needs to be involved

Developer, designer

querkmachine:
I've trawled through our existing favicon and sharing graphics and what's shakin' in the space currently and come up with the following thoughts.

These are strictly for v5 onwards. 
v4.x and older will use updated versions of the existing favicons they shipped with. 

## Keep these

|Current filename|Filetype|Dimensions|Notes|
|:-|:-|-:|:-|
|favicon.ico|Icon|(up to) 32×32|Used by iOS/iPadOS Safari in the browser. Also used by legacy browsers. This is still used in enough contexts that we're probably best keeping it. We may need to [change how we're using the `sizes` attribute](https://twitter.com/subzey/status/1417099064949235712), however.|
|govuk-mask-icon.svg|SVG|[square]|Used by macOS Safari. Expected to be a flat colour, background-less, black icon that is coloured according to a `theme-color` meta tag in HTML. Colours can be different between dark and light mode.|
|govuk-apple-touch-icon-180x180.png|PNG|180×180|Used by iOS and iPadOS. Has unchangingly been the standard size for iOS app icons since around 2017. Could be renamed with removal of other Apple icons.|
|govuk-opengraph-image.png|PNG|1200×630|Used for OpenGraph-based sharing. Different services crop the image differently so practically speaking only the centre square is a safe zone.|

## Bin these

|Current filename|Filetype|Dimensions|Notes|
|:-|:-|-:|:-|
|govuk-apple-touch-icon.png|PNG|120×120|Outdated. Lower density displays will scale down the 180x180 version automatically.|
|govuk-apple-touch-icon-152x152.png|PNG|152×152|As above.|
|govuk-apple-touch-icon-167x167.png|PNG|167×167|As above.|

## Add these

|Potential filename|Filetype|Dimensions|Notes|
|:-|:-|-:|:-|
[...]

Issue #4058, opened 2023-08-08
Create the assets for the crown


stevenjmesser:
## What

Create the SVG for the logo used in the header, as well as other applications like the favicon. Relies on us having a clear idea of [where we need to update it](https://github.com/alphagov/govuk-frontend/issues/4057).

```[tasklist]
### Assets shared across v3, v4 and v5.x (Tudor crown)
- [x] Favicon (.ico)(16x16, 32x32, 48x48)
- [x] Apple touch icon (.png) (180x180)
- [x] Mask icon (.svg)
- [x] Open Graph image (.png) (1200x630)
```

```[tasklist]
### v3.x & v4.x assets (Tudor crown)
- [x] Standalone crown (.svg)
- [x] Standalone crown fallback (.png)
- [x] Legacy Apple touch icons (.png) (120x120, 152x152, 167x167)
```

```[tasklist]
### v5.x assets (Tudor crown)
- [x] Logo lockup (.svg)
- [x] Favicon (.svg)
- [x] Android touch icons (.png) (192x192, 512x512)
```

```[tasklist]
### v5.0 transitional assets (St. Edward's crown)
- [x] Logo lockup (.svg)
- [x] Favicon (.svg)
- [x] Android touch icons (.png) (192x192, 512x512)
```

For more info on this list, see notes in https://github.com/alphagov/govuk-frontend/issues/4057#issuecomment-1706617935

## Why

So we can roll out the new crown across GOV.UK.

## Who needs to work on this 

Designer, developer

## Who needs to review this

The crown is already signed off by the Garter and Cabinet Office comms. We'll run versions of the full lockup GOV.‌UK logo by members of the design community and senior people on GOV.UK.

colinrotherham:
Worked with @christopherthomasdesign yesterday to create some new SVG assets

For comparison, a few options:

## Old crown
>File type: SVG with single compound path
>File size: 2,263 bytes

![crown-old](https://github.com/alphagov/govuk-frontend/assets/415517/52114ac2-9369-4af7-9108-c50e215fbdc6)

## New crown
>File type: SVG with single compound path
>File size: 2,423 bytes

![crown-new](https://github.com/alphagov/govuk-frontend/assets/415517/3d77ae64-f47e-4ff1-ad22-08bd3226dc40)

## New crown
>File type: SVG with single compound path
>File size: 2,423 bytes

## New crown, mirrored horizontally
>File type: SVG with single symbol
>File size: 1,511 bytes

[...]

PR #4185, opened 2023-09-08
[SPIKE] Update favicons (v5)

querkmachine:
Provisionally update the favicons based on my thoughts in this comment (https://github.com/alphagov/govuk-frontend/issues/4175#issuecomment-1709975127), implementing all except for the 'maskable' version of the Android icon, as that would likely require a unique graphical treatment to ensure legibility.

> [!WARNING]
> The graphical assets used in this PR are **not final**. The PR, whilst in draft, is intended to test changes to what icons we provide and how we implement them from a technical perspective. Final iconography will be produced as part of #4058. 

## Changes
- Replaces icons using the St. Edward's Crown with equivalents using the Tudor Crown. These are **not final** assets, as described above.
- Renames all of the icons to be largely agnostic to their usage, instead being named for the image's dimensions.
- Changes the way the ICO favicon is referenced in HTML: 
  - Fixes a potential issue with the `sizes` attribute where Chrome would always use the ICO even when an SVG favicon was available ([`sizes` issue detailed here in section 2.3](https://dev.to/masakudamatsu/favicon-nightmare-how-to-maintain-sanity-3al7)).
  - Changes `rel="shortcut icon"` to `rel="icon"`. The `shortcut` keyword was only needed for supporting Internet Explorer 8 and earlier, which Frontend no longer does.
- Removes `apple-touch-icon` sizes smaller than 180x180. These aren't used by Apple devices released in the last few years and lower resolution Apple devices can scale down the larger image automatically.
- Adds an SVG favicon. 
  - This favicon deviates from the .ico version by not containing the crown within a box, instead utilising the `prefers-color-scheme` CSS query to change the colour of the crown depending on the user's colour settings. 
  - The removal of the containing box also allows the crown to be larger, as there's no need to provide padding around it.
- Adds 192x192 PNG icon. This is most commonly used by Android devices.
- Adds a 512x512 PNG icon. This is a bit of a future-proofing 'catch-all' unspecific to any specific device or OS, but present in case an icon larger than 192x192 is needed. This may also be used by macOS Sonoma. 
- Adds a `manifest.json` file listing all of the available PNG icon sizes.
- Adds Jest tests for the icons and manifest file. 

## Screenshots

All of the below display the icons from this PR alongside the current favicon that GOV.UK serves to that browser, so that the differences between their appearance can be clearly seen,

### Chrome 116, light system theme
SVG favicon.

Screenshot of two tabs in Chrome, but light

### Chrome 116, dark system theme
SVG favicon.

Screenshot of two tabs in Chrome, but dark

### Edge 116, light system theme
SVG favicon.

Screenshot of two tabs in Edge, but light

### Edge 116, dark system theme
SVG favicon.

Screenshot of two tabs in Edge, but dark

### Firefox 115, light system theme
SVG favicon.

Screenshot of two tabs in Firefox, but light

### Firefox 115, dark system theme
SVG favicon.

Screenshot of two tabs in Firefox, but dark

### Safari 16.6 on macOS, light system theme with separate tab bar
Safari mask icon.

Screenshot of two tabs in Safari, but light

### Safari 16.6 on macOS, dark system theme with separate tab bar
Safari mask icon.

A white background is added to the icon as Safari surmises that the configured mask colour is too dark to be visible against the dark system theme.

Screenshot of two tabs in Safari, but dark

### Safari 16.6 on macOS, compact tab bar

Safari mask icon.

When Safari is configured to use the compact tab bar layout, the tab bar takes on a colour from the current page (either automatically, or the one defined by `theme-color` meta tag). 

As there is no guarantee that the favicon will have contrast against those colours, a white background is added around it.

Screenshot of two tabs in Safari, but in rounded boxes against a purple background

### Internet Explorer 11

ICO favicon. Captured through Browserstack, which may have distorted the appearance of the icons due to video compression.

Screenshot of two tabs in Internet Explorer 11

### Safari 17.0 (beta) on iOS, tab switcher

ICO favicon.

Screenshot of the tab switcher on an iPhone, showing previews of two webpages with their favicons and page titles written below them

### iOS 17.0 (beta) home screen

Apple touch icons.

Two app-like icons on an iPhone homescreen, the first renamed 'v4' and the second named 'v5'

### macOS 14.0 (beta) dock

Uses the largest icon defined in the manifest file, if any are defined. If not, falls back to using `apple-touch-icon`. If neither are available, the favicon is used. If none of them are available, macOS generates an icon using the first letter of the page's title with a background colour.

Two app-like icons on the macOS dock

### Android 13

Android 192 PNG favicon.

Both the homescreen and Android version of Chrome use the Android icon defined in the manifest. I couldn't get screenshots of this because the device security policy disallowed it, but they work. 

The `theme-color` HTML attribute still works too, and has not been affected by the addition of a manifest file.


packages/govuk-frontend/src/govuk/assets/manifest.webmanifest
```diff
@@ -0,0 +1,6 @@
+{
```
colinrotherham, line 1: Great to see this 😊

Probably want to confirm if this overrides **template.njk** so we don't lose:

1. `<meta name="theme-color">`
2. `<link rel="mask-icon">`

For compatibility with web servers, it's often named **manifest.json**

Worth a check?

packages/govuk-frontend/src/govuk/assets/manifest.webmanifest
```diff
@@ -0,0 +1,6 @@
...
+  "icons": [
+    { "src": "images/govuk-android-icon.png", "type": "image/png", "sizes": "192x192" },
+    { "src": "images/govuk-android-icon-512.png", "type": "image/png", "sizes": "512x512" }
```
colinrotherham, line 4: Coming through on Chrome macOS

Icons from manifest

colinrotherham: Although this is where it gets tricky as a "copy and paste" manifest, dynamic fields:

Validation errors

packages/govuk-frontend/src/govuk/assets/manifest.webmanifest
querkmachine: Currently checking 'em! I think the HTML tags are safe, as neither are being overridden by the manifest.

Noted the .json file thing in my comment on the issue. It seems like `.webmanifest` is the preferred option these days, but I figure we might have to ask around on that to see how severely it might impact folks. Given this would be new for v5 I don't think it'll be a massive barrier to ask folks to check their content-type configurations?

packages/govuk-frontend/src/govuk/assets/manifest.webmanifest
querkmachine: Likewise in Firefox's inspector 👍

packages/govuk-frontend/src/govuk/assets/manifest.webmanifest
colinrotherham: As an optional thing we might be alright?

We're shipping our default ESM bundle as `govuk-frontend.bundle.js` not `*.mjs` for content-types though!

packages/govuk-frontend/src/govuk/assets/manifest.webmanifest
colinrotherham: Not checked the Safari 17 beta yet, but "Add to Home Screen" will work on desktop too
https://webkit.org/blog/14205/news-from-wwdc23-webkit-features-in-safari-17-beta/

Assuming it's using manifest support, the name might need a tweak?

Or consolidating our "touch icons" for multiple platform sizes?

```suggestion
    { "src": "images/govuk-touch-icon.png", "type": "image/png", "sizes": "192x192" },
    { "src": "images/govuk-touch-icon-512.png", "type": "image/png", "sizes": "512x512" }
```

querkmachine: The warnings under 'Installability' are only relevant to PWAs and aren't required if we're simply supplying icons for home screen and bookmarking purposes. 

I'd forgotten about PWAs coming to desktop Safari. I like to live dangerously so I was already running Safari 17, and have added a screenshot of the icons it used above. I'm not 100% sure (though I assume based on the documentation) that it uses he manifest icons if they're defined, so there's good reason to potentially rename them.

querkmachine: Weirdly, since updating to the newest Sonoma beta over the weekend (presumably clearing out Safari caches in the process) it now appears to be favouring the apple-touch-icon over the manifest file. 

Not sure if that's an intentional change in the update, or whether caching fudged with the results before or after. May well be the latter—Safari is slightly notorious for aggressive caching of favicons.

packages/govuk-frontend/src/govuk/assets/manifest.webmanifest
querkmachine: Yeah, I don't think this is worth causing a hubbub about and JSON is the 'safer' of the two options, so will go with that unless we have compelling reasons to do otherwise.

packages/govuk-frontend/src/govuk/assets/images/govuk-icon-mask.svg
colinrotherham, line 1: Just as I'm looking at the code, do we need `version="1.1"` set?

packages/govuk-frontend/src/govuk/assets/images/govuk-icon-512.png
colinrotherham, line 1: Good to chuck aware the alpha channel from these PNGs or try limiting to 256 colours not full sRGB

[ImageOptim](https://imageoptim.com/mac) will nearly save 70% on this big one, down to just over 5KB

packages/govuk-frontend/src/govuk/assets/images/favicon.svg
colinrotherham, line 1: Is the XML line optional at the top? Especially the encoding which web servers may mismatch

```svg
<?xml version="1.0" encoding="UTF-8"?>
```

packages/govuk-frontend/src/govuk/assets/images/govuk-icon-mask.svg
querkmachine: I'm not sure. This is a carry over from the previous mask icon, which also included it and doesn't include the XML doctype (`<?xml version="1.0" encoding="UTF-8"?>`), which the SVG favicon _does_ have.

There doesn't seem to be any consensus online—some examples and tutorials include it, others do not—but to me that reads as it not being a requirement for it to work.

packages/govuk-frontend/src/govuk/assets/manifest.json
```diff
@@ -0,0 +1,19 @@
+{
+  "icons": [
```
colinrotherham, line 2: How much do we need to match these up with **template.njk**

E.g. We've not got the 48x48 icon here (maybe that's fine)

packages/govuk-frontend/src/govuk/assets/images/govuk-icon-180.png
colinrotherham, line 1: >Removes apple-touch-icon sizes smaller than 180x180. These aren't used by Apple devices released in the last few years and lower resolution Apple devices can scale down the larger image automatically.

Based on your findings, wondering if the 192px icon is enough to delete this one?

packages/govuk-frontend/src/govuk/assets/images/favicon.ico
colinrotherham, line 1: >**Internet Explorer 11**
>
> ICO favicon. Captured through Browserstack, which may have distorted the appearance of the icons due to video compression.

Love the `sizes="48x48"` trick but presume we'd keep 16x16 and 32x32 sizes for older browsers?

querkmachine: Older browsers (like IE 11 here) don't understand the sizes attribute so they just ignore the attribute. The .ico itself is 48x48, so it seems like IE is sizing it down automatically without us needing to supply seperately sized versions.

colinrotherham: Yeah that's what I mean sorry

The trick is great, but we did we definitely need to remove 16x16 and 32x32 icons?

Downsampling from 48x48 will add blur in IE11 (not just Browserstack compression) but that might be fine

packages/govuk-frontend/src/govuk/assets/manifest.json
querkmachine: I don't think they practically need to match at all. If the _only_ reason we're providing a manifest is for Android devices, we could get away with only defining the 192x192 image. 

I included the 180x180 and 512x512 as options in case a requesting service or device might find them useful, as we already have them on hand, and to provide a little bit of futureproofing in case iOS does start using the manifest (as we're seeing with macOS Sonoma), or some other usage comes along that prefers an even larger graphic to be available.

packages/govuk-frontend/src/govuk/assets/images/govuk-icon-180.png
querkmachine: This is more of a design preference question. We have 180x180 for iOS because their icons are 60x60 <abbr title="Device Independent Pixels">dips</abbr>, rendered on all recent devices at 3x resolution. We can be confident that a 180x180 image will scale perfectly and remain crisp.

We're providing a 192x192 image for Android for the same reasons, except their icons are rendered at 64x64 dips.

I couldn't see any evidence that either functionally needs the icons to be those sizes—we could toss a 3000x3000 monster at both and be done with it—but supplying the sizes that look the best on each feels like the nicer thing to do.

packages/govuk-frontend/src/govuk/assets/images/govuk-icon-512.png
querkmachine: As these aren't the final assets, I'm not gonna do this just yet, but optimisation is definitely on the cards for later. ✌️

packages/govuk-frontend/src/govuk/assets/images/favicon.ico
querkmachine: I could remake the .ico to contain 16x16 and 32x32 versions, but even if we did, I don't think the `sizes` attribute needs changing to account for it—every browser that would use those sizes would either ignore the attribute or prioritise using one of the other icons we're supplying.

As another 'not final assets' point, I'm probably not gonna add more sizes to the .ico yet, but will intend to do so for the release.

packages/govuk-frontend/src/govuk/assets/manifest.json
colinrotherham: >…provide a little bit of futureproofing…

Yeah for this bit

i.e. For future proofing, and as a safety net, should we match **template.njk** to prepare for the day other browsers also ship support for **manifest.json**?

Making sure all the familiar sizes they support are still in there

I've double checked the [**What’s New in Safari** documentation](https://developer.apple.com/library/archive/releasenotes/General/WhatsNewInSafari/Articles/Safari_11_1.html#//apple_ref/doc/uid/TP40014305-CH14-SW6) to find when manifest support was added (Safari 11.1) but they default to `apple-touch-icon` in the HTML

Future browsers might default to `icons` in the manifest instead 🤷‍♂️

colinrotherham: Sizes wise, we're all set for the [Google docs](https://web.dev/add-manifest/)

> For Chromium, you must provide at least a 192x192 pixel icon, and a 512x512 pixel icon. If only those two icon sizes are provided

Except for Android we likely want an SVG touch icons too given their examples?

>Chromium-based browsers also support SVG icons which can be scaled arbitrarily without looking pixelated and that support advanced features like [responsiveness to `prefers-color-scheme`](https://blog.tomayac.com/2021/07/21/dark-mode-web-app-manifest-app-icons/), with the caveat that the icons do not update live, but remain in the state they were in at install time.

```suggestion
  "icons": [
    {
      "src": "/images/govuk-icon.svg",
      "type": "image/svg+xml",
      "sizes": "512x512"
    },
```

colinrotherham: To add, [Apple's documentation](https://developer.apple.com/design/human-interface-guidelines/app-icons) has widened since https://github.com/alphagov/govuk_template/pull/300

They suggest 1024×1024 for auto scaling, but for exact pixel sizes they list:

### iPhone
* 180×180 `@3x` — Home screen
* 120×120 `@2x` — Home screen

### iPad
* 167×167 `@1x` — Home screen (iPad Pro)
* 152×152 `@1x` — Home screen (iPad, iPad mini)

### Mac
* 512×512 `@1x` — Desktop, Dock
* 1024×1024 `@2x` — Desktop, Dock

packages/govuk-frontend/src/govuk/assets/images/favicon.ico
colinrotherham: Yup. Definitely keep `sizes="48x48"` as it's a fab hack 🙌

packages/govuk-frontend/src/govuk/assets/images/govuk-icon-180.png
colinrotherham: Thanks @querkmachine makes total sense to keep it

I've reviewed the [Apple documentation](https://developer.apple.com/design/human-interface-guidelines/app-icons) in https://github.com/alphagov/govuk-frontend/pull/4185#discussion_r1324571351 and found a few gaps

Mainly 2018, 2019 devices (iPhone X) at `@2x` or the previous non-retina iPad Pro and iPad mini

Seems fair to [follow what we did back in 2017](https://github.com/alphagov/govuk_template/pull/300) and review them all again

packages/govuk-frontend/src/govuk/assets/manifest.json
querkmachine: We could supply all of those if we really want to, it's just not really necessary.

I've stuck with a single 180x180 that seems to be the sole size used in many modern implementations and recommended by folks dedicated enough to write about the subject.[^1][^2][^3][^4]

I'm not too concerned with re-defining every single icon from the template in the manifest either. The backwards-compatible nature of the web means that even if Safari, et al. update to _prefer_ the manifest, they're unlikely to _only_ use the manifest, and letting them choose the icon that fits best is why we specify size information in the first place.

[^1]: https://www.leereamsnyder.com/blog/favicons-in-2021
[^2]: https://evilmartians.com/chronicles/how-to-favicon-in-2021-six-files-that-fit-most-needs
[^3]: https://dev.to/masakudamatsu/favicon-nightmare-how-to-maintain-sanity-3al7
[^4]: realfavicongenerator.net changed to only supply a single 180x180 icon way back in 2016 — https://github.com/RealFaviconGenerator/realfavicongenerator/issues/211

colinrotherham: There's only one size missing if we say the desktop sizes are out of scope?

* 120×120 `@2x` — Home screen

Don't have to add the "desktop" sizes, maybe the SVG though? https://github.com/alphagov/govuk-frontend/pull/4185#discussion_r1324567044

colinrotherham:
@querkmachine If you give this branch a rebase, we fixed the Rollup build failures in https://github.com/alphagov/govuk-frontend/pull/4208

packages/govuk-frontend/src/govuk/assets/manifest.json
querkmachine: I've opted to relent and add all of the icons to the manifest. Seemingly this isn't unexpected (MDN has examples that include even the .ico being defined within the manifest), but I'll be surprised if at least some of these aren't superfluous, being unused by any current browser/device.

I'm not sure whether the current SVG supplied is really suitable for the intended use cases either. It's not the same design as the PNGs offered and doesn't come with any included margins, so if it were to be selected for an app icon on a device it may end up looking a bit weird. Something to test.

querkmachine:
Closing this spike for now as I think it's accomplished what it needed to. A new PR has been created for the implementation of new icon formats using the Tudor crown: #4354

Depending on the scheduling of the v5 release and GOV.UK comms, we may also have a 'transitional' PR that implements the new formats but continues using the St. Edward's crown (but that doesn't exist yet!)

PR #4199, opened 2023-09-11
Throw `ElementError` when "Exit this page" button is missing

### LINKED ISSUE #4129: Throw errors during Exit This Page initialisation if key HTML elements are missing
## What

- The button

## Why

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

romaricpascal:
~Wording to match that of the Skip link component, but to be reviewed in #4072.~ Wording is now handled by the `ElementError` class and will match the other `ElementError` being thrown.

Closes #4129

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/exit-this-page/exit-this-page.mjs
+++ b/packages/govuk-frontend/src/govuk/components/exit-this-page/exit-this-page.mjs
@@ -90,7 +90,11 @@ export class ExitThisPage extends GOVUKFrontendComponent {
 
     const $button = $module.querySelector('.govuk-exit-this-page__button')
     if (!($button instanceof HTMLElement)) {
-      return this
+      throw new ElementError($button, {
+        componentName: 'Exit this page',
+        identifier: 'Button',
+        expectedType: HTMLElement
+      })
     }
 
     this.config = mergeConfigs(
--- a/packages/govuk-frontend/src/govuk/components/exit-this-page/exit-this-page.puppeteer.test.js
+++ b/packages/govuk-frontend/src/govuk/components/exit-this-page/exit-this-page.puppeteer.test.js
@@ -237,6 +237,20 @@ describe('/components/exit-this-page', () => {
           message: 'Exit this page: $module is not an instance of "HTMLElement"'
         })
       })
+
+      it('throws when the button is missing', async () => {
+        await expect(
+          renderAndInitialise(page, 'exit-this-page', {
+            params: examples.default,
+            beforeInitialisation($module) {
+              $module.querySelector('.govuk-exit-this-page__button').remove()
+            }
+          })
+        ).rejects.toEqual({
+          name: 'ElementError',
+          message: 'Exit this page: Button not found'
+        })
+      })
     })
   })
 })


packages/govuk-frontend/src/govuk/components/exit-this-page/exit-this-page.test.js
```diff
@@ -208,6 +208,20 @@ describe('/components/exit-this-page', () => {
           message: 'GOV.UK Frontend is not supported in this browser'
         })
       })
+
+      it('throws when the button is missing', async () => {
+        await expect(
+          renderAndInitialise(page, 'exit-this-page', {
+            params: examples.default,
+            beforeInitialisation($module) {
+              $module.querySelector('.govuk-exit-this-page__button').remove()
```
colinrotherham, line 217: 🚀

packages/govuk-frontend/src/govuk/components/exit-this-page/exit-this-page.mjs
```diff
@@ -86,7 +87,7 @@ export class ExitThisPage extends GOVUKFrontendComponent {
 
     const $button = $module.querySelector('.govuk-exit-this-page__button')
     if (!($button instanceof HTMLElement)) {
-      return this
+      throw new MissingElementError('Exit this page: Button not found')
```
colinrotherham, line 90: Still not sure whether we can simplify this?

This PR differs to [#4104](https://github.com/alphagov/govuk-frontend/pull/4104/files#diff-f67b9a6c95fac32fd6400f00f9aa875ac18ee16b38e305f925ee2bb0f377d4d1) and [#4177](https://github.com/alphagov/govuk-frontend/pull/4177/files#diff-fb3a25d42cbb3621d375cc34dc960447e84fa2a8729fd16e97e4c34edd9b17aeR40) as _truthy_ buttons get an incorrect "not found" error

Knowing that we have two incorrect types to deal with:

* Incorrect _truthy_ type `$button && !($button instanceof HTMLElement)`
* Incorrect `null` type "not found" is `!$button` or exactly `$button === null`

Can we narrow down our approach?

### 1. Keep the dual-error approach from [#4104](https://github.com/alphagov/govuk-frontend/pull/4104/files#diff-f67b9a6c95fac32fd6400f00f9aa875ac18ee16b38e305f925ee2bb0f377d4d1)

```suggestion
      throw !module
        ? new MissingElementError('Exit this page: $button not found')
        : new TypeError(
            'Exit this page: $button not an instance of "HTMLElement"'
          )
```

### 2. Use TypeError for both incorrect types

```suggestion
      throw new TypeError(
        `Exit this page: ${
          !$module
            ? '$button not found' // Type is `null` for "not found"
            : '$button not an instance of "HTMLElement"'
        }`
      )
```

romaricpascal: Yeah, that's fair. The simplification was because all these `instanceof` checks are really existence checks in their intent, but you're right there's different situation. How about a variation of 2 where we throw the following error to avoid ternaries all over the codebase when we do these `instanceof`:

```js
class ElementError extends GOVUKFrontendError {
   constructor(element, {componentName, variableName, expectedType}) {
       const reason = element
         ? `${variableName} not found'
         : `${variableName} not an instance of ${expectedType.name}`
       
       // To be fair, I'd put that second bit of formatting in `GOVUKFrontendError`
       super(`${componentName: $variableName}`)
   }
}
```

romaricpascal: That would pair well with an `AttributeError` when there's a missing hash in the skip link 😊

romaricpascal: And could be introduced as part of: https://github.com/alphagov/govuk-frontend/pull/4104

colinrotherham: Brilliant, this is like [`ElementTypeError` we discussed on Slack](https://gds.slack.com/archives/C0447D0STV0/p1694106177971369?thread_ts=1694084487.172989&cid=C0447D0STV0)

```mjs
throw new ElementTypeError('$module', {
  component: 'Accordion',
  expected: HTMLElement
})
```

@domoscargin and I discussed passing in the `componentName` in https://github.com/alphagov/govuk-frontend/pull/4177#discussion_r1318526224

Good to bring it all together

romaricpascal: Cool, I'll rework #4104 to introduce that new error instead of throwing `TypeError`. Once it's merged, we can rebase this one and get to implementing the rest of the errors 😊

romaricpascal: Implemented and rebased on the [components-type-error](https://github.com/alphagov/govuk-frontend/tree/components-type-error) branch from #4104 so this can be reviewed while we sort the last piece of it.

colinrotherham: Thanks @romaricpascal

I don't mind which you choose, but we've [iterated our wording a few times](https://github.com/alphagov/govuk-frontend/pull/4206#discussion_r1322902397)

Code looks super tidy though

romaricpascal: Aligned both to `throws when ... is missing`. I'll comment on the issues for the other components so we carry the wording across.

colinrotherham (approved):

PR #4202, opened 2023-09-11
Fix Review app code examples using `example.options`

colinrotherham:
Fixes Markup and Macro code rendering in the review app

Macro options were renamed to ~`example.data`~ `example.options` in https://github.com/alphagov/govuk-frontend/pull/4042/commits/7a2d1386e10c5e11969b7f3b9fdc4bda7d6e4dcc via https://github.com/alphagov/govuk-frontend/pull/4042

We missed this because the `{% for example in componentFixtures.fixtures %}` loop was in a separate file

Code rendering has missing component options

To help avoid Nunjucks issues like this in future I've included a few changes such as:

1. Move code partial into `showExamples.njk` for easier debugging
2. Configure Nunjucks with `dev: true` to log stack traces etc


colinrotherham:
@romaricpascal This change was missed from https://github.com/alphagov/govuk-frontend/pull/4042/commits/7a2d1386e10c5e11969b7f3b9fdc4bda7d6e4dcc 🤦‍♂️

packages/govuk-frontend-review/src/common/nunjucks/index.mjs
```diff
@@ -20,6 +20,7 @@ export function renderer(app) {
   const env = nunjucksEnv(
     [join(paths.app, 'src/views')],
     {
+      dev: true, // log stack traces
```
romaricpascal, line 23: Ohhh, that's going to be handy for debugging! 🥳

romaricpascal (commented):
Whoops for missing that at the time, didn't occur to me when reviewing to check for leftover offcurences of `example.data` or even `\.data(?!set)` (many `.dataset` that are not relevant).

Cheers for tidying it up! Don't think there's much risk on moving the code partial inside the macro, unless anyone in @alphagov/design-system-developers was planning to reuse that bit specifically?

packages/govuk-frontend-review/src/common/nunjucks/index.mjs
colinrotherham: I discovered it recently and I'm getting bored of adding it 😆

You get to see Nunjucks stack traces right into global/filter code too

romaricpascal (approved):
Plenty of time to raise concerns about the `code.njk` disappearance, nothing raised, let's merge! 🥳

Issue #4169, opened 2023-09-04
Fix GitHub Actions workflow flaky tests


colinrotherham:
## What

Our GitHub Actions workflows only have two flaky tests

They typically pass when re-run. We should fix them

1. [**Banner** in JavaScript behaviour tests](https://github.com/alphagov/govuk-frontend/blob/7a33714d5a0d48a7805877330cce96954dddcfcf/packages/govuk-frontend-review/src/routes/banner.test.mjs#L5-L8) throws `TypeError: fetch failed` calling `fetchPath()`
    
    > https://github.com/alphagov/govuk-frontend/actions/runs/6072389732/job/16472409676
    > https://github.com/alphagov/govuk-frontend/actions/runs/6072198194/job/16471867044
    > https://github.com/alphagov/govuk-frontend/actions/runs/6072787178/job/16473539140
    > https://github.com/alphagov/govuk-frontend/actions/runs/6072960509/job/16474064674
    > https://github.com/alphagov/govuk-frontend/actions/runs/6039910759/job/16389701125
    > https://github.com/alphagov/govuk-frontend/actions/runs/6039836317/job/16389477864
    > https://github.com/alphagov/govuk-frontend/actions/runs/6039338891/job/16387907168
    > https://github.com/alphagov/govuk-frontend/actions/runs/6037941434/job/16383398031

2. [**Rollup** stats build](https://github.com/alphagov/govuk-frontend/blob/01981fa9f05b59c5bca16672c4c688984e62c35d/packages/govuk-frontend/tasks/scripts.mjs#L56-L57) throws `RollupError: Unexpected token`

    > https://github.com/alphagov/govuk-frontend/actions/runs/6122493682/job/16618794894#step:5:40
    > https://github.com/alphagov/govuk-frontend/actions/runs/6144727581/job/16670866996#step:5:40
    > https://github.com/alphagov/govuk-frontend/actions/runs/6144759438/job/16670871062#step:4:55
    > https://github.com/alphagov/govuk-frontend/actions/runs/6147512336/job/16679295401#step:4:55
    > https://github.com/alphagov/govuk-frontend/actions/runs/6145962649/job/16674439248#step:4:55

Flaky test 2) [also affects webpack](https://github.com/alphagov/govuk-frontend/actions/runs/6144759438/job/16670871062#step:4:463) with "missing semicolon" logged

Perhaps `dist/govuk/common/index.mjs` isn't valid due to Babel not updating the Rollup source maps?

## Why

[...]

PR #4205, opened 2023-09-12
Match Jest environment by filename `*.puppeteer.test.js` etc

colinrotherham:
This PR does a spring clean to simplify our Jest filenames [as discussed](https://github.com/alphagov/govuk-frontend/pull/2878#pullrequestreview-1119840159) previously by @romaricpascal

1. Use filename `*.puppeteer.test.{js,mjs}` for Puppeteer tests
2. Use filename `*.jsdom.test.{js,mjs}` for JSDOM tests
3. Use filename `*.unit.test.{js,mjs}` for unit tests

Lots of our tests were actually "unit tests" (by content) but matched other Jest environments (by filename)

Similarly, running JSDOM setup/teardown for every unit test was slow when only these files actually needed it:

```
packages/govuk-frontend/src/govuk/all.unit.test.mjs
packages/govuk-frontend/src/govuk/i18n.unit.test.mjs
packages/govuk-frontend/src/govuk/components/character-count/character-count.unit.test.mjs
packages/govuk-frontend/src/govuk/common/index.jsdom.unit.mjs
packages/govuk-frontend/src/govuk/common/closest-attribute-value.jsdom.unit.mjs
```

## Legacy Cheerio tests

We typically use [Cheerio](https://cheerio.js.org) for template rendering tests and snapshots

But instead of [Puppeteer](https://pptr.dev) these three tests made their own [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) HTTP requests:

```
packages/govuk-frontend-review/src/app.test.mjs
packages/govuk-frontend-review/src/routes/banner.test.mjs
packages/govuk-frontend-review/src/routes/full-page-examples.test.mjs
```

We've now migrated them to [Puppeteer](https://pptr.dev) so the flaky `fetchPath()` helper can be removed:

* https://github.com/alphagov/govuk-frontend/issues/4169

### Before
We load HTML from `fetchPath()` into Cheerio selectors

```mjs
it('should have theme-color overriden', async () => {
  const response = await fetchPath('/examples/template-custom')
  const $ = load(await response.text())

  const $linkMaskIcon = $('link[rel="mask-icon"]')
  const $metaThemeColor = $('meta[name="theme-color"]')

  expect($linkMaskIcon.attr('color')).toBe('blue')
  expect($metaThemeColor.attr('content')).toBe('blue')
})
```

### After
We use Puppeteer selectors like in our other tests

```mjs
beforeAll(async () => {
  await goTo(page, '/examples/template-custom')
})

it('should have theme-color overridden', async () => {
  const $linkMaskIcon = await page.$('link[rel="mask-icon"]')
  const $metaThemeColor = await page.$('meta[name="theme-color"]')

  expect(getAttribute($linkMaskIcon, 'color')).resolves.toBe('blue')
  expect(getAttribute($metaThemeColor, 'content')).resolves.toBe('blue')
})
```


colinrotherham:
Mostly looking good

Only issue seems to be `packages/govuk-frontend/postcss.config.test.mjs`

Some of our checks aren't working without the JSDOM environment

.eslintrc.js
```diff
@@ -121,6 +121,16 @@ module.exports = {
         jest: true
       }
     },
+    {
+      // Matches Puppeteer environment in jest.config.mjs
+      // to ignore unknown Jest Puppeteer globals
+      files: ['**/*.puppeteer.test.{js,mjs}'],
```
colinrotherham, line 127: Moved from [**packages/govuk-frontend/.eslintrc.js**](https://github.com/alphagov/govuk-frontend/pull/4205/files#diff-5522c69942e4ae29ac022153cddf7837010848662f1ad2afd987ea9d6643bbbf) to allow new Puppeteer tests in the Review app to use `page` globals etc

colinrotherham:
> Only issue seems to be `packages/govuk-frontend/postcss.config.test.mjs`
> 
> Some of our checks aren't working without the JSDOM environment

This was very interesting actually 😬

Under the JSDOM environment it hid `NODE_ENV` from PostCSS causing plugins to run in "development mode"

Glad we found that, fix has now been pushed

packages/govuk-frontend-review/src/app.puppeteer.test.mjs
```diff
@@ -0,0 +1,173 @@
...
+
+    it('should load correctly', async () => {
+      const $title = await page.$('title')
+
+      // Check the page responded correctly
+      expect(getProperty($title, 'textContent')).resolves.toContain('GOV.UK')
+    })
+  })
+
+  describe('/examples/template-custom', () => {
```
36degrees, line 30: I think these tests pre-date us (cccb7d3352da08ef03312709cf629992a567542c) introducing separate tests for the template itself (3095e165ebb5d2fc945711cbd9c5393ab040b0ec).

There does seem to be a lot of overlap, and these are really testing the template rather than the review app. I do wonder if we should just delete this whole `describe` block instead?

Happy to create an issue to review this later if you prefer.

package.json
```diff
@@ -23,9 +23,9 @@
     "predev": "npm run build",
     "dev": "concurrently \"npm run dev --workspace @govuk-frontend/review\" \"npm run dev --workspace govuk-frontend\" --kill-others --names \"app,pkg\" --prefix-colors \"red.dim,blue.dim\"",
     "build": "npm run build --workspace govuk-frontend --workspace @govuk-frontend/review",
-    "build:app": "npm run build --workspace @govuk-frontend/review",
-    "build:package": "npm run build:package --workspace govuk-frontend",
-    "build:release": "npm run build:release --workspace govuk-frontend",
+    "build:app": "cross-env-shell NODE_ENV=production npm run build --workspace @govuk-frontend/review",
```
36degrees, line 26: I think this is probably fine as when e.g. doing the build for a release we _should_ be following the process in our release docs, but wondering if it's worth trying to flip this so that we 'default to' creating a production build and the review app opts out?

Would that be the safer behaviour?

36degrees (commented):
I think the changes broadly make sense, but I think the bigger change in this PR is the 'spring clean'. Can we either update the PR title to make it clearer what's changing, or split this into 2 PRs?

A couple of minor comments, both are just things to consider rather than blocking.

package.json
colinrotherham: I was very torn by adding this

Problem came from PostCSS following this Node.js principle:
https://nodejs.dev/en/learn/nodejs-the-difference-between-development-and-production/

>Node.js assumes it's always running in a development environment. You can signal Node.js that you are running in production by setting the `NODE_ENV=production` environment variable.

So it would be safer for sure, but we're not in control of the defaults our dependencies us 😣

E.g. If we assume `production` as the default, what if CSSNANO defaults to `development`?

packages/govuk-frontend-review/src/app.puppeteer.test.mjs
colinrotherham: Let's remove it

On the basis that I removed all the HTTP header checks too

We already have [tests for `x-robots-tag`](https://github.com/alphagov/govuk-frontend/commit/4b1a3695f2eaac3f3ef6ed8448440281a988ee38#diff-ddb795bb152f353e17602ff9179ab6f6a97532b3608013f57accad032bb1c2d5) and Puppeteer will tell us should `text/html` go wrong

package.json
36degrees: Yep, that's fair. We'll just have to be careful to make sure we're using the top-level build scripts (or setting the env) when doing anything where it's important we have the version number etc.

packages/govuk-frontend-review/src/app.puppeteer.test.mjs
colinrotherham: Added an extra commit for the removal https://github.com/alphagov/govuk-frontend/pull/4205/commits/f26d60d06ec3442a76122a11cfe7073e6c8ebfd4

colinrotherham:
> I think the changes broadly make sense, but I think the bigger change in this PR is the 'spring clean'. Can we either update the PR title to make it clearer what's changing, or split this into 2 PRs?

Thanks @36degrees

All feedback addressed and I've updated the PR title and description

36degrees (approved):

PR #4206, opened 2023-09-12
Throw `ElementError` when the menu of the header is missing but a toggle is present

### LINKED ISSUE #4130: Throw errors during Header initialisation if key HTML elements are missing
## What

- Menu

## Why

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

romaricpascal:
We need to keep the early return if the toggle is absent, as it's not mandatory for the header to have a navigation.

Closes #4130

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/header/header.mjs
+++ b/packages/govuk-frontend/src/govuk/components/header/header.mjs
@@ -61,9 +61,16 @@ export class Header extends GOVUKFrontendComponent {
       return this
     }
 
-    const $menu = $module.querySelector(
-      `#${$menuButton.getAttribute('aria-controls')}`
-    )
+    const menuId = $menuButton.getAttribute('aria-controls')
+    if (!menuId) {
+      throw new ElementError($menuButton, {
+        componentName: 'Header',
+        identifier: '$menuButton["aria-controls"]',
+        expectedType: 'string'
+      })
+    }
+
+    const $menu = document.getElementById(menuId)
 
     if (!($menuButton instanceof HTMLElement)) {
       throw new ElementError($menu, {
--- a/packages/govuk-frontend/src/govuk/components/header/header.mjs
+++ b/packages/govuk-frontend/src/govuk/components/header/header.mjs
@@ -61,28 +61,27 @@ export class Header extends GOVUKFrontendComponent {
       return this
     }
 
-    const menuId = $menuButton.getAttribute('aria-controls')
-    if (!menuId) {
+    if (!($menuButton instanceof HTMLElement)) {
       throw new ElementError($menuButton, {
         componentName: 'Header',
-        identifier: '$menuButton["aria-controls"]',
-        expectedType: 'string'
+        identifier: '.govuk-js-header-toggle'
+      })
+    }
+
+    const menuId = $menuButton.getAttribute('aria-controls')
+    if (!menuId) {
+      throw new ElementError(null, {
+        componentName: 'Header',
+        identifier: '.govuk-js-header-toggle[aria-controls]'
       })
     }
 
     const $menu = document.getElementById(menuId)
 
-    if (!($menuButton instanceof HTMLElement)) {
-      throw new ElementError($menu, {
-        componentName: 'Header',
-        identifier: 'Menu'
-      })
-    }
-
     if (!($menu instanceof HTMLElement)) {
       throw new ElementError($menu, {
         componentName: 'Header',
-        identifier: 'Menu'
+        identifier: `#${menuId}`
       })
     }
 
--- a/packages/govuk-frontend/src/govuk/components/header/header.mjs
+++ b/packages/govuk-frontend/src/govuk/components/header/header.mjs
@@ -52,21 +52,35 @@ export class Header extends GOVUKFrontendComponent {
     }
 
     this.$module = $module
-    this.$menuButton = $module.querySelector('.govuk-js-header-toggle')
-    this.$menu =
-      this.$menuButton &&
-      $module.querySelector(
-        `#${this.$menuButton.getAttribute('aria-controls')}`
-      )
+    const $menuButton = $module.querySelector('.govuk-js-header-toggle')
 
-    if (
-      !(
-        this.$menuButton instanceof HTMLElement ||
-        this.$menu instanceof HTMLElement
-      )
-    ) {
+    // Headers don't necessarily have a navigation.
+    // When they don't, the menu toggle won't be rendered
+    // by our macro (or may be omitted when writing plain HTML)
+    if (!$menuButton) {
       return this
     }
+
+    const $menu = $module.querySelector(
+      `#${$menuButton.getAttribute('aria-controls')}`
+    )
+
+    if (!($menuButton instanceof HTMLElement)) {
+      throw new ElementError($menu, {
+        componentName: 'Header',
+        identifier: 'Menu'
+      })
+    }
+
+    if (!($menu instanceof HTMLElement)) {
+      throw new ElementError($menu, {
+        componentName: 'Header',
+        identifier: 'Menu'
+      })
+    }
+
+    this.$menu = $menu
+    this.$menuButton = $menuButton
 
     // Set the matchMedia to the govuk-frontend desktop breakpoint
     this.mql = window.matchMedia('(min-width: 48.0625em)')
--- a/packages/govuk-frontend/src/govuk/components/header/header.puppeteer.test.js
+++ b/packages/govuk-frontend/src/govuk/components/header/header.puppeteer.test.js
@@ -230,6 +230,23 @@ describe('Header navigation', () => {
         })
       })
 
+      it("throws when the toggle's aria-control attribute is missing", async () => {
+        await expect(
+          renderAndInitialise(page, 'header', {
+            params: examples['with navigation'],
+            beforeInitialisation($module) {
+              $module
+                .querySelector('.govuk-js-header-toggle')
+                .removeAttribute('aria-controls')
+            }
+          })
+        ).rejects.toEqual({
+          name: 'ElementError',
+          message:
+            'Header: $menuButton["aria-controls"] is not of type "string"'
+        })
+      })
+
       it('throws when the menu is missing, but a toggle is present', async () => {
         await expect(
           renderAndInitialise(page, 'header', {
--- a/packages/govuk-frontend/src/govuk/components/header/header.puppeteer.test.js
+++ b/packages/govuk-frontend/src/govuk/components/header/header.puppeteer.test.js
@@ -219,6 +219,31 @@ describe('Header navigation', () => {
           message: 'Header: $module is not an instance of "HTMLElement"'
         })
       })
+
+      it('does not throw if the toggle is absent', async () => {
+        // The default example is rendered without navigation
+        // and should keep rendering. No expectations as the JavaScript
+        // will just return early. All we ask of that test is for it not
+        // to throw during the initialisation
+        await renderAndInitialise(page, 'header', {
+          params: examples.default
+        })
+      })
+
+      it('throws when the menu is missing, but a toggle is present', async () => {
+        await expect(
+          renderAndInitialise(page, 'header', {
+            params: examples['with navigation'],
+            beforeInitialisation($module) {
+              // Remove the menu `<ul>` referenced by $menuButton's `aria-controls`
+              $module.querySelector('#navigation').remove()
+            }
+          })
+        ).rejects.toEqual({
+          name: 'ElementError',
+          message: 'Header: Menu not found'
+        })
+      })
     })
   })
 })
--- a/packages/govuk-frontend/src/govuk/components/header/header.puppeteer.test.js
+++ b/packages/govuk-frontend/src/govuk/components/header/header.puppeteer.test.js
@@ -230,6 +230,24 @@ describe('Header navigation', () => {
         })
       })
 
+      it('throws when the toggle is of the wrong type', async () => {
+        await expect(
+          renderAndInitialise(page, 'header', {
+            params: examples['with navigation'],
+            beforeInitialisation($module) {
+              // Replace with an `<svg>` element which is not an `HTMLElement` in the DOM (but an `SVGElement`)
+              $module.querySelector(
+                '.govuk-js-header-toggle'
+              ).outerHTML = `<svg class="govuk-js-header-toggle" ></svg>`
+            }
+          })
+        ).rejects.toEqual({
+          name: 'ElementError',
+          message:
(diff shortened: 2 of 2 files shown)


packages/govuk-frontend/src/govuk/components/header/header.test.js
```diff
@@ -221,6 +221,30 @@ describe('Header navigation', () => {
...
+        // The default example is rendered without navigation
+        // and should keep rendering. No expectations as the JavaScript
+        // will just return early. All we ask of that test is for it not
+        // to throw during the initialisation
+        await renderAndInitialise(page, 'header', {
+          params: examples.default
+        })
+      })
+
+      it('throws if the toggle is present, but no menu is there', async () => {
```
colinrotherham, line 235: Only a minor one, but can you pick your favourite wording across the PRs?

Helps when searching the project for "is not set" or "throws when" etc

### Suggestion from [#4104](https://github.com/alphagov/govuk-frontend/pull/4104/files#diff-4579399f94ca47997fb8c34cf74e61fe87a1ca2ffadb5d6050302ae285af6112R342)

```suggestion
      it('throws when #navigation is not set', async () => {
```

<br>

### Suggestion from [#4199](https://github.com/alphagov/govuk-frontend/pull/4199/files#diff-24879075b6d7335064bd2d9c2919edcb111443bf3f52549b048e4a16cbd9dde4R243)

```suggestion
      it('throws when the navigation is missing', async () => {
```

packages/govuk-frontend/src/govuk/components/header/header.mjs
```diff
@@ -53,22 +53,30 @@ export class Header extends GOVUKFrontendComponent {
...
-
-    if (
-      !(
-        this.$menuButton instanceof HTMLElement ||
-        this.$menu instanceof HTMLElement
-      )
-    ) {
+    const $menuButton = $module.querySelector('.govuk-js-header-toggle')
+
+    // Header don't necessarily have a navigation, in which case
```
colinrotherham, line 58: Do we need to watch our wording/naming here or in the component?

1. **`<nav>`** for `.govuk-header__navigation`
2. **`<ul>`** for `govuk-header__navigation-list`

Confusingly in the JavaScript `this.$menu` is the **`<ul>`** but has `id="navigation"

```suggestion
    // Headers don't necessarily have a navigation, in which case
```

packages/govuk-frontend/src/govuk/components/header/header.test.js
romaricpascal: Happy to settle for `throws when ... is missing`. I still want to complement that last one, as it'll only throw if it's missing *and* there is a toggle in the markup.

colinrotherham: You choose 😊

packages/govuk-frontend/src/govuk/components/header/header.mjs
romaricpascal: Whoops, sorry for the typo.

Regarding the wording, I was referring to what's toggled by the `navigation` option of the Nunjucks component, which would be the `<nav>`, in the comment. Agreed it's a bit unfortunate that the `id` is `navigation` for the `__navigation-list` element. Would be a breaking change for people not using the Nunjucks macro, so worth its own discussion if we want to change it (not sure the current situation hurts much, but not fussed if the team prefers updating).

colinrotherham: > Do we need to watch our wording/naming here **or in the component?**

Cheers for fixing that. Probably more the last bit?

I'll link another comment to this thread

packages/govuk-frontend/src/govuk/components/header/header.mjs
```diff
@@ -52,22 +52,30 @@ export class Header extends GOVUKFrontendComponent {
...
-      )
-    ) {
+    const $menuButton = $module.querySelector('.govuk-js-header-toggle')
+
+    // Headers don't necessarily have a navigation.
+    // When they don't, the menu toggle won't be rendered
+    // by our macro (or may be omitted when writing plain HTML)
+    if (!($menuButton instanceof HTMLElement)) {
       return this
     }
```
colinrotherham, line 62: Hmm, the plain HTML comment has caught my eye

Due to progressive enhancement we fall back to the menu being open

But remove the button and invalid navigation menus get a `return this` free pass

(We should still check them)

romaricpascal: Do you mean a Header having a menu but not a toggle, which would fall into this `return this`?

romaricpascal: Looked a bit more into it: I think with a missing button, we would not have been checking the type of the `$menu` anyway due to it begin `false` because `this.$menuButton` being falsey anyway (first line of the table).

Slightly different matter if `$menuButton` was just not an instance of `HTMLElement`. If `$menu` was an instance of `HTMLElement` we wouldn't be return early while we should have (1).

Digging a bit further, I think there was a few instance where the type of `$menu` wasn't checked where it should have been because of the `||` not running the second clause if the first is truthy (2).

$menuButton | $menu | $menuButton instanceof HTMLElement | $menu instanceof HTMLElement | returning early
-- | -- | -- | -- | --
`null` from not being on the page | Can only be `false` from `this.$menuButton` being falsey | FALSE | FALSE | Yes
an `HTMLElement` | an `HTMLElement` | TRUE | Not run | No (2)
not an `HTMLElement` | an `HTMLElement` | FALSE | TRUE | No (1)
an `HTMLElement` | not an `HTMLElement` | TRUE | Not run | No (2)
not an `HTMLElement` | not an `HTMLElement` | FALSE | FALSE | Yes

I _think_ the following implementation limits the free pass for `$menu` to when `$menuButton` is `null` (like in `v4`) while making sure its type gets checked when `$menuButton` is an `HTMLElement` (2). No checks when `$menuButton` is not an HTMLElement, but the component would not carry further as it mistakenly does at the moment.

```
    // Headers don't necessarily have a navigation.
    // When they don't, the menu toggle won't be rendered
    // by our macro (or may be omitted when writing plain HTML)
    if (!$menuButton) {
      return this
    }

    const $menu = $module.querySelector(
      `#${$menuButton.getAttribute('aria-controls')}`
    )

    if (!($menuButton instanceof HTMLElement)) {
      throw new ElementError($menuButton, {
        componentName: 'Header',
        identifier: 'Menu toggle'
      })
    }

    if (!($menu instanceof HTMLElement)) {
      throw new ElementError($menu, {
        componentName: 'Header',
        identifier: 'Menu'
      })
    }

```

colinrotherham: Thanks for the write up 😎

I'm afraid the free pass is still there, but as:

```mjs
if (!$menuButton) {
  return this
}
```

i.e. By removing the `$menuButton` I can fall back to an always-open `<svg>` menu without errors

Can we stick with the usual approach of:

1. Run a selector
2. Check if missing or wrong type
3. Add to class field after checks pass

```mjs
// 1. Run selector
const $menu = $module.querySelector('#navigation')
    
// 2. Check if missing or wrong type
if (!($menu instanceof HTMLElement)) {
  throw new ElementError($menu, {
    componentName: 'Header',
    identifier: 'Menu'
  })
}
```

Another error would be `$menuButton.getAttribute('aria-controls')` not matching `navigation`

romaricpascal: Yeah, that would work if we add a check for the link of the `$menu.id` and `$menuButton.getAttribute('aria-controls')`.

colinrotherham: @romaricpascal Brill, glad that did the trick

Yeah please, it's commented out in https://github.com/alphagov/govuk-frontend/pull/4206/commits/7660f032b83e54d741ab4cb9e707c65648c456ba (commit after) but would be crucial. Hopefully easier to string a nice error message together since we merged https://github.com/alphagov/govuk-frontend/pull/4219/commits/8965564d8c087ada039adf008276da99668bd8c0 too

```mjs
if (!$menu.id || $menu.id !== $menuButton.getAttribute('aria-controls')) {
  throw new ElementError($menuButton, {
    componentName: 'Header',
    identifier: '$menuButton["aria-controls"]',
    expectedType: '???'
  })
}
```

packages/govuk-frontend/src/govuk/components/header/header.mjs
```diff
@@ -52,22 +52,36 @@ export class Header extends GOVUKFrontendComponent {
...
+    // Headers don't necessarily have a navigation.
+    // When they don't, the menu toggle won't be rendered
+    // by our macro (or may be omitted when writing plain HTML)
+    if (!$menuButton) {
       return this
     }
 
+    const $menu = $module.querySelector(
+      `#${$menuButton.getAttribute('aria-controls')}`
+    )
```
colinrotherham, line 66: Saw this piece of history yesterday in https://github.com/alphagov/govuk-frontend/commit/266f8772857ca9afc60ded16279ea4b2c9595462 whilst reviewing **Checkboxes** errors

i.e. Should we trust `$menuButton.getAttribute('aria-controls')` is a valid ID before running it?

Not a blocker since we had this issue before, but we typically guard it elsewhere:

```mjs
const targetId = $input.getAttribute('aria-controls')
if (!targetId) {
  return
}
```

packages/govuk-frontend/src/govuk/components/header/header.mjs
```diff
@@ -52,22 +52,36 @@ export class Header extends GOVUKFrontendComponent {
...
+    // Headers don't necessarily have a navigation.
+    // When they don't, the menu toggle won't be rendered
+    // by our macro (or may be omitted when writing plain HTML)
+    if (!$menuButton) {
       return this
     }
 
+    const $menu = document.getElementById(
+      $menuButton.getAttribute('aria-controls')
+    )
```
colinrotherham, line 66: Would you mind if we added the guard in https://github.com/alphagov/govuk-frontend/pull/4206#discussion_r1334120152?

Matches the approach we recently added to **Skip link**

```suggestion
    const menuId = $menuButton.getAttribute('aria-controls')

    // Check for menu button target
    if (!menuId) {
      throw new ElementError(this.$module, {
        componentName: 'Header',
        identifier: '$menuButton["aria-controls"]',
        expectedType: 'string'
      })
    }

    const $menu = document.getElementById(menuId)
```

packages/govuk-frontend/src/govuk/components/header/header.mjs
colinrotherham: See https://github.com/alphagov/govuk-frontend/pull/4206#discussion_r1334196561

packages/govuk-frontend/src/govuk/components/header/header.mjs
```diff
@@ -52,22 +52,43 @@ export class Header extends GOVUKFrontendComponent {
...
+      })
+    }
+
+    const $menu = document.getElementById(menuId)
+
+    if (!($menuButton instanceof HTMLElement)) {
+      throw new ElementError($menu, {
+        componentName: 'Header',
+        identifier: 'Menu'
+      })
```
colinrotherham, line 79: Sorry for not spotting this earlier

We'll need to customise this one for the button not the menu:

```suggestion
      throw new ElementError($menuButton, {
        componentName: 'Header',
        identifier: '.govuk-js-header-toggle'
      })
```

packages/govuk-frontend/src/govuk/components/header/header.mjs
```diff
@@ -52,22 +52,43 @@ export class Header extends GOVUKFrontendComponent {
...
+    }
+
+    if (!($menu instanceof HTMLElement)) {
+      throw new ElementError($menu, {
+        componentName: 'Header',
+        identifier: 'Menu'
```
colinrotherham, line 85:
```suggestion
        identifier: `#${menuId}`
```

packages/govuk-frontend/src/govuk/components/header/header.mjs
```diff
@@ -52,22 +52,43 @@ export class Header extends GOVUKFrontendComponent {
...
+    const menuId = $menuButton.getAttribute('aria-controls')
+    if (!menuId) {
+      throw new ElementError($menuButton, {
+        componentName: 'Header',
+        identifier: '$menuButton["aria-controls"]',
+        expectedType: 'string'
+      })
+    }
+
+    const $menu = document.getElementById(menuId)
```
colinrotherham, line 73: Only a tiny suggestion, but can we complete all the `$menuButton` checks first and move this down?

Gives the `$menuButton` checks chance to throw before we run selectors we might not need yet

packages/govuk-frontend/src/govuk/components/header/header.mjs
```diff
@@ -52,22 +52,43 @@ export class Header extends GOVUKFrontendComponent {
...
+    // by our macro (or may be omitted when writing plain HTML)
+    if (!$menuButton) {
       return this
     }
 
+    const menuId = $menuButton.getAttribute('aria-controls')
+    if (!menuId) {
+      throw new ElementError($menuButton, {
+        componentName: 'Header',
+        identifier: '$menuButton["aria-controls"]',
```
colinrotherham, line 68: 🙌

packages/govuk-frontend/src/govuk/components/header/header.mjs
romaricpascal: Whoops, forgot to update when I copy-pasted, sorry.

colinrotherham: Don't worry, do you want to add tests to make sure we've caught them all?

packages/govuk-frontend/src/govuk/components/header/header.mjs
```diff
@@ -52,22 +52,43 @@ export class Header extends GOVUKFrontendComponent {
...
+
+    if (!($menu instanceof HTMLElement)) {
+      throw new ElementError($menu, {
+        componentName: 'Header',
+        identifier: `#${menuId}`
+      })
+    }
+
+    this.$menu = $menu
+    this.$menuButton = $menuButton
```
colinrotherham, line 90: Thanks for doing this

Helps the compiler out a lot (narrows the types down) when adding to class fields _after_ the checks

packages/govuk-frontend/src/govuk/components/header/header.mjs
colinrotherham: In fact, my suggestion was daft wasn't it. Mind fixing it? 🤦‍♂️

```suggestion
        identifier: '.govuk-js-header-toggle["aria-controls"]',
```

colinrotherham (approved):
Approved, but can you fix my daft previous suggestion? Sorry

Issue #4209, opened 2023-09-12
Replace `MissingElementError` with `ElementError`


romaricpascal:
## What

Update `SkipLink` to throw an `ElementError` instead of a `MissingElementError` and remove that error from the codebase.

As the `ElementError` already handles the prefixing with the component name, this is also the occasion to just let the errors through from where the missing bits are detected rather than throw, catch, wrap and re-throw. 

## Why

Due to a lot of checks for the presence of element being `instanceof` checks, we introduced an `ElementError` to format the message appropriately depending on whether the element was actually missing or the wrong type. This error overlaps a lot what `MissingElementError` was doing so we need only one of those.

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [x] The `SkipLink` component no longer throws a `MissingElementError`
- [x] The error has been removed from the codebase

romaricpascal:
Blocked until we discuss #4072/#4074 as how we decide to finalise our errors may impact what exactly the replacement error will be.

colinrotherham:
Closed by https://github.com/alphagov/govuk-frontend/pull/4219

colinrotherham:
@36degrees Yeah, should have said "Will be closed by" and it's closed now 🎉

Thanks

PR #4219, opened 2023-09-18
Replace Skip link errors with `ElementError`

colinrotherham:
This PR replaces **Skip link** `$module` and `$module.hash` errors with `ElementError` to close:

* https://github.com/alphagov/govuk-frontend/issues/4209


packages/govuk-frontend/src/govuk/components/skip-link/skip-link.mjs
```diff
@@ -10,18 +10,17 @@ export class SkipLink extends GOVUKFrontendComponent {
...
 
   /** @private */
   linkedElementListener = false
 
   /**
    * @param {Element} $module - HTML element to use for skip link
-   * @throws {MissingElementError} If the element with the specified ID is not found
+   * @throws {ElementError} when $module is not set or the wrong type
+   * @throws {ElementError} when $module.hash does not contain a hash
+   * @throws {ElementError} when the linked element is missing or the wrong type
```
colinrotherham, line 23: @romaricpascal Just flagging that @domoscargin has documented `@throw` for [JSDoc output](https://govuk-frontend-review.herokuapp.com/docs/javascript/classes/SkipLink.html#constructor)

We only need them on public methods but I've listed them based on the test names

packages/govuk-frontend/src/govuk/components/skip-link/skip-link.mjs
```diff
@@ -30,23 +29,12 @@ export class SkipLink extends GOVUKFrontendComponent {
...
       })
     }
 
     this.$module = $module
-
-    // Check for linked element
-    try {
-      const $linkedElement = this.getLinkedElement()
-      this.$linkedElement = $linkedElement
-    } catch (error) {
```
colinrotherham, line 43: @domoscargin It got a bit tricky to handle all the error variations in a try/catch

Now that we have `ElementError` we can throw the same error type

packages/govuk-frontend/src/govuk/errors/index.mjs
```diff
@@ -59,7 +52,7 @@ export class ElementError extends GOVUKFrontendError {
    * @param {object} options - Element error options
    * @param {string} options.componentName - The name of the component throwing the error
    * @param {string} options.identifier - An identifier that'll let the user understand which element has an error (variable name, CSS selector)
-   * @param {typeof HTMLElement} [options.expectedType] - The type that was expected for the element
+   * @param {typeof HTMLElement | StringConstructor} [options.expectedType] - The type that was expected for the identifier
```
colinrotherham, line 55: Hope this makes sense

We have errors on the **Skip link** `$module` element [`.hash` property](https://developer.mozilla.org/en-US/docs/Web/API/HTMLAnchorElement/hash) which now output:

* `'Skip link: $module.hash is not an instance of "String"'`
* `'Skip link: $module.hash target #this-element-does-not-exist not found'`

domoscargin: Is `StringConstructor` purely a TypeScript thing? I can't find vanilla JS docs for it.

And the second question: do we care? I haven't been on top of all the parameter typing, but have we included things elsewhere that aren't either native or explicitly imported?

To be clear: I think this is fine and understandable, just curious about these queries.

domoscargin (approved):
Looks good!

I have a query, but it's non-blocking.

packages/govuk-frontend/src/govuk/errors/index.mjs
colinrotherham: Yeah `StringConstructor` is in their library types with `BooleanConstructor`, `NumberConstructor` etc

https://github.com/microsoft/TypeScript/blob/main/src/lib/es5.d.ts#L526
https://github.com/Microsoft/TypeScript/blob/main/src/lib/es2015.core.d.ts#L522

I'm happy with them since we we heavily rely on their DOM libs already

Context wise, I made the switch from `typeof String` to `StringConstructor` to stop ESLint fighting:

> `Invalid JSDoc @param "options.expectedType" type "String"; prefer: "string".`
> eslint([jsdoc/check-types](https://github.com/gajus/eslint-plugin-jsdoc/blob/main/docs/rules/check-types.md#repos-sticky-header))

But this morning I'm having second thoughts since it's a primitive `'string'` not a `new String('string')`

colinrotherham: Should probably make `expectedType` a string since it's only for building a message

colinrotherham: @domoscargin How's this? https://github.com/alphagov/govuk-frontend/compare/f825dbacee1080a6088c165fee749b609c6a359f..70265f15802a4a9f599e4f7b71e15632d10d62ee

domoscargin: Looks good to me! Tests will probably need an update for the slightly different messages.

PR #4233, opened 2023-09-20
Add changelog entry for #4113

36degrees:
As part of a Dependabot update for `html-validate`, we made a change so that the button `type` attribute is always set (defaulting to `submit`, which is the same behaviour as if the `type` attribute is omittted)

Although this isn’t a breaking change, I think it’s worth mentioning in the release notes.

I’ve also updated the PR title to mention the change. This means the fix and the PR title are not identical, as I’ve omitted the `html-validate` bump here, but I think that’s OK.


colinrotherham (approved):
Looks good to me

Do we need to say it reverses entry https://github.com/alphagov/govuk-frontend/commit/f7c8917b9b757b228878def03596b347b9eb8e4e to remove button `type`?

36degrees:
I don't think so – it's linked from the PR if anyone wants to go digging 🕳️

PR #4235, opened 2023-09-20
[support/3.x] Add note on how support branches work

36degrees:
We’re adding docs on using support branches to our team docs as part of alphagov/design-system-team-docs#21.

Update our release docs for v3.x to reflect the fact that this branch is a support branch and to link to the new docs.


36degrees:
I'm also going to use this branch as a bit of a test and to configure the required status checks for the `support/3.x` branch.

colinrotherham (approved):

PR #4236, opened 2023-09-20
Throw ElementError (not found) if checkboxes or radios have no input items

### LINKED ISSUE #4131: Throw errors during Radios and Checkboxes initialisation if key HTML elements are missing
## What

- The `<input>` elements

## Why

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

domoscargin:
Closes #4131


domoscargin:
Suspect we also want to throw an error here, for the case where the element with `targetId` is not found, but not for the case where the input has no `data-aria-controls` attribute.

https://github.com/alphagov/govuk-frontend/blob/df9c59a4f6487a0445f4374d0322371403582747/packages/govuk-frontend/src/govuk/components/checkboxes/checkboxes.mjs#L56-L60

packages/govuk-frontend/src/govuk/components/checkboxes/checkboxes.mjs
```diff
@@ -40,7 +40,11 @@ export class Checkboxes extends GOVUKFrontendComponent {
     /** @satisfies {NodeListOf<HTMLInputElement>} */
     const $inputs = $module.querySelectorAll('input[type="checkbox"]')
     if (!$inputs.length) {
-      return this
+      throw new ElementError(null, {
+        componentName: 'Checkboxes',
+        identifier: '<input type="checkbox">',
+        expectedType: HTMLElement
```
colinrotherham, line 46: Love seeing this!

I suppose it would be `HTMLInputElement` but `null` **element** skips `expectedType` 🤷‍♂️

```suggestion
        identifier: '<input type="checkbox">'
```

colinrotherham (commented):
Am I allowed to approve it yet??

domoscargin:
> Am I allowed to approve it yet??

Haha, I'm just looking at the stuff in my comment - which wouldn't be caused by a user-created parameter, except I guess in the case where they're using HTML and make a mistake.

domoscargin:
@colinrotherham Should hopefully be good for review now.

packages/govuk-frontend/src/govuk/components/checkboxes/checkboxes.mjs
```diff
@@ -51,13 +51,21 @@ export class Checkboxes extends GOVUKFrontendComponent {
 
     this.$inputs.forEach(($input) => {
       const targetId = $input.getAttribute('data-aria-controls')
```
domoscargin, line 53: We create item and conditional IDs internally in the template: 

https://github.com/alphagov/govuk-frontend/blob/704cf6d1a29a5478a25542575a2197ebaf5e1c66/packages/govuk-frontend/src/govuk/components/checkboxes/template.njk#L62

However, this check is still useful for folks using HTML instead of the Nunjucks macros, as they'd be inputting their own `data-aria-controls` and ids.

packages/govuk-frontend/src/govuk/components/checkboxes/checkboxes.mjs
```diff
@@ -40,21 +40,32 @@ export class Checkboxes extends GOVUKFrontendComponent {
     /** @satisfies {NodeListOf<HTMLInputElement>} */
     const $inputs = $module.querySelectorAll('input[type="checkbox"]')
     if (!$inputs.length) {
-      return this
+      throw new ElementError(null, {
+        componentName: 'Checkboxes',
+        identifier: '<input type="checkbox">'
```
colinrotherham, line 45: No problem with this at all, I think it's much friendlier

But now all the other JavaScript variable `identifier: '$module'` options jump out at me

Thinking ahead then, wonder if we should stick to CSS selectors?

### Before
```mjs
// Checkboxes: $module not found
identifier: `$module`
```

```mjs
// Checkboxes: <input type="checkbox"> not found
identifier: `<input type="checkbox">`
```

### After
```mjs
// Checkboxes: [data-module="govuk-checkboxes"] not found
identifier: `[data-module="${Checkboxes.moduleName}"]`
```

```mjs
// Checkboxes: input[type="checkbox"] not found
identifier: 'input[type="checkbox"]'
```

packages/govuk-frontend/src/govuk/components/checkboxes/checkboxes.mjs
```diff
@@ -40,21 +40,32 @@ export class Checkboxes extends GOVUKFrontendComponent {
...
+      // Skip radios without data-aria-controls attributes
+      if (!targetId) {
         return
       }
 
+      // Throw if target conditional element does not exist.
+      if (!targetElement) {
+        throw new ElementError(targetElement, {
+          componentName: 'Checkboxes',
+          identifier: `.govuk-checkboxes__conditional #${targetId}`
```
colinrotherham, line 65: I like how this is clearer, but realised we only select by ID?

```suggestion
          identifier: `#${targetId}`
```

colinrotherham: I do prefer CSS selectors though, see https://github.com/alphagov/govuk-frontend/pull/4236/files#r1331775855

domoscargin: I'm happy to revert this for this PR, but something to discuss when we review the content - is `targetId` user-friendly enough when there may be some kind of issue with it? Ideally having the parent radios item info in there would be great, but realise I may be going too far.

colinrotherham: Did some digging, and we used to scope it to `$module` before https://github.com/alphagov/govuk-frontend/commit/266f8772857ca9afc60ded16279ea4b2c9595462

I suppose since then the target could genuinely be anywhere on the page?

E.g. Tick a box in a left column and a question is revealed on the right?

packages/govuk-frontend/src/govuk/components/checkboxes/checkboxes.mjs
```diff
@@ -49,12 +52,21 @@ export class Checkboxes extends GOVUKFrontendComponent {
...
+      // Skip radios without data-aria-controls attributes
+      if (!targetId) {
         return
       }
 
+      const targetElement = document.getElementById(targetId)
+
+      // Throw if target conditional element does not exist.
+      if (!targetElement) {
+        throw new ElementError(targetElement, {
```
colinrotherham, line 64: Sorry Brett now it's moved I've noticed we only use it when it's `null`

```suggestion
      // Throw if target conditional element does not exist.
      if (!document.getElementById(targetId)) {
        throw new ElementError(null, {
```

colinrotherham (approved):

packages/govuk-frontend/src/govuk/components/checkboxes/checkboxes.mjs
domoscargin: Oh nice catch!

Issue #4237, opened 2023-09-20
The review app stops compiling Sass after encountering an error


36degrees:
## Description of the issue

If you make a change that causes an error in Sass compliation whilst using the review app, the build process for the review app crashes and is not restarted.

This means that even if you fix the change, you need to stop the process and restart it (Ctrl + C then re-run `npm start`).

## Steps to reproduce the issue

1. Run npm start
2. Edit *.scss file to cause a Sass or PostCSS error

## Actual vs expected behaviour

Expected behaviour:

The error is outputted on the console and the watch task for the review app build continues to retry compilation when further changes are made

Actual behaviour:

The error is outputted on the console but the watch task for the review app build crashes.

## Environment (where applicable)

- Operating system: macOS 13.5.1
- Browser: N/A
- Browser version: N/A
- GOV.UK Frontend Version: N/A (happening on `main` branch)

colinrotherham:
Will be fixed by https://github.com/alphagov/govuk-frontend/pull/4239 and tested in https://github.com/alphagov/govuk-frontend/pull/4240

PR #4239, opened 2023-09-20
Fix Review app Sass compile after error

colinrotherham:
This PR fixes Gulp "Did you forget to signal async completion?" when Sass compilation errors occur in:

* https://github.com/alphagov/govuk-frontend/issues/4237

### The issue

Spot my missing `await` in https://github.com/alphagov/govuk-frontend/commit/8f74d0265903b199b91f21425e7bb357d55bee39 🤦‍♂️

### Error logging

I've made sure all "watch + reload" tasks use Gulp's [`new PluginError()`](https://github.com/gulpjs/plugin-error#readme) for consistent output, as currently:

1. Sass tasks via `scripts.compile()` throw `new PluginError()`
2. Rollup tasks via `styles.compile()` throw `new PluginError()`
3. npm tasks via `npm.script()` log [`console.error(error.message)` without error](https://github.com/alphagov/govuk-frontend/blob/2ad331924779ed1ff17cc19c261e299410ae7671/shared/tasks/npm.mjs#L33-L41)

But Gulp is now aware of npm task errors too, without interrupting watch tasks:

ESLint errors now known to Gulp

### DIFF
--- a/shared/tasks/npm.mjs
+++ b/shared/tasks/npm.mjs
@@ -28,6 +28,12 @@ export async function run(name, args = [], options) {
       throw new Error(`Task '${name}' not found in '${pkgPath}'`)
     }
   } catch (cause) {
+    // Skip Nodemon (SIGINT) exit or aborted task error codes
+    // https://github.com/open-cli-tools/concurrently/pull/359/files
+    if (cause.signal === 'SIGINT' || [130, 3221225786].includes(cause.code)) {
+      return
+    }
+
     throw new PluginError(`npm run ${name}`, cause, {
       // Hide error properties already formatted by npm
       showProperties: false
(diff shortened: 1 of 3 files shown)


colinrotherham:
Thanks @romaricpascal

Just need to confirm those error codes etc in Windows before merging

36degrees (approved):
🙌🏻

colinrotherham:
To avoid unnecessary error logging on Windows I've silenced `{ code: 3221225786 }` as shown:

* https://github.com/microsoft/vscode/issues/107501
* https://github.com/open-cli-tools/concurrently/pull/359

Don't want to make Ctrl + C too noisy on Windows if we can avoid it there too

PR #4240, opened 2023-09-21
Add new `govuk-frontend.min.css` package bundle

colinrotherham:
To make sure we catch Sass compilation problems early, our npm package has a new minified bundle

Previously we ran PostCSS-only tasks during `npm run build:package` and missed https://github.com/alphagov/govuk-frontend/pull/4239

### Package minified bundles

Output using `npm run build:package`

* **`govuk-frontend.min.js`**
* **`govuk-frontend.min.css`** ✅ Added

### GitHub Release minified bundles

**Using** `npm run build:release`

* **`govuk-frontend-X.Y.Z.min.js`**
* **`govuk-frontend-X-Y-Z.min.css`**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,15 @@ For advice on how to use these release notes see [our guidance on staying up to 
 
 ### New features
 
+#### Precompiled CSS and JavaScript
+
+The precompiled CSS and JavaScript files found in our [GitHub releases](https://github.com/alphagov/govuk-frontend/releases) are now also published to [`govuk-frontend` on npm](https://www.npmjs.com/package/govuk-frontend).
+
+These changes were introduced in:
+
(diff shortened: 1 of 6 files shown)


36degrees:
I've split this from #4237 as it's related but I think we can call it done – will add this to the board separately.

romaricpascal (commented):
Neat! Checked the output and it's the same as the one in `dist` (except the source map URL, obviously) 🙌🏻 

Could you add a little entry to the CHANGELOG, please? then all good to go 🥳

colinrotherham:
Thanks @romaricpascal

>Could you add a little entry to the CHANGELOG, please? then all good to go 🥳

✅ Done

romaricpascal (approved):
Cheers for adding the changelog entry! 😊 Looks good to merge on the tech side. @claireashworth let us know if the CHANGELOG entry looks fine.

PR #4241, opened 2023-09-21
Skip JSDoc, SassDoc, Rollup stats during dev, add ESLint

colinrotherham:
This PR does a few things to speed up development:

1. Skips JSDoc, SassDoc, and Rollup stats
2. Splits watch tasks into separate “lint” and “compile”
3. Moves TypeScript compiler `build:types` to a watch task
4. Adds new `npm run clean` task for workspaces without Gulp

### Where have my type checks gone?

Type checks [moved to `npm run lint` last month](https://github.com/alphagov/govuk-frontend/pull/4096/commits/904ec06698541a5710290810dea371fd4b187ea6) so we can skip `npm run build:types`

In development, ESLint, StyleLint and the TypeScript compiler now run via `gulp.watch()` instead

### Where have my stats gone?

JSDoc, SassDoc and Rollup stats are now considered optional in development

That said, the Review app will turn on these optional sections when they're built using:

```console
npm run build:jsdoc --workspace @govuk-frontend/review
npm run build:sassdoc --workspace @govuk-frontend/review
npm run build:stats --workspace govuk-frontend
```


domoscargin:
I don't have any major problem with this - takes my rebuild time from 4+ seconds to 1-2 seconds.

One slight concern is it makes it more annoying to develop things that focus on the stuff that's skipped. I can't just run `npm run build:sassdoc`, for example - I have to remember how to access the review app workspace and run it from there.

Relatedly, does this mean we'll get all the type complaints only when we run the Github action workflow?

colinrotherham:
```

**Comment by colinrotherham** [comment]
Sometimes you might want to skip the slow `predev` build entirely and run these in separate tabs:

```shell
npm run dev --workspace govuk-frontend
npm run dev --workspace @govuk-frontend/review
```

Another top tip

domoscargin:
Yeah, I think running the doc stuff manually will be absolutely fine 99.5% of the time so that's grand.

Completely on the fence about whether it's nicer to develop without getting type failures, or whether it'd be worse to get a bunch of 'em at once and potentially have to do a bigger refactor. Hopefully somebody else chimes in and leans one way or the other!

colinrotherham:
@domoscargin I'd quite like them turned on too, but it's usually just me with that opinion 😆

If you don't mind [TypeScript `*.tsbuildinfo` cache files](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#faster-subsequent-builds-with-the---incremental-flag) it goes from ~7 seconds to 500ms (see below)

We do this in GitHub Actions already:

https://github.com/alphagov/govuk-frontend/blob/c397f768655167337f93d09f320a194adede6f78/.github/workflows/tests.yml#L99

### Type check

>6.63s user 0.41s system 208% cpu 3.376 total

```console
time npm run build:types --workspace govuk-frontend
```

### Type check with cache

>0.52s user 0.13s system 79% cpu 0.821 total

```console
time npm run build:types --workspace govuk-frontend -- --incremental
```

domoscargin:
Any way to

colinrotherham:
@domoscargin I've added the TypeScript compiler `build:types` back as a watch task this morning

Ready for review again

36degrees:
I think this makes sense most of the time but that there are scenarios where it'd be useful to see the stats or docs update as you make changes.

How hard would it be to:

- provide an easy way to re-enable the stats / docs (using a command line flag or environment variable)
- hide the stats and docs in the review app when they're not being 'live updated' (but always show them on Heroku deploys)

That way you won't end up in situation where you can forget you're looking at stale stats / docs. Thoughts?

colinrotherham:
@36degrees Sounds like a nice addition, can decide the best approach today?

In readiness, I've pushed `flags.hasDocs` and `flags.hasStats` to the Review app to hide them for `npm start`

colinrotherham:
We talked about this in today's dev catch-up

Optional flags to "opt in" to JSDoc, SassDoc or Rollup stats can be added in future

In the meantime we can use existing npm scripts when optional features are required:

```console
npm run build:jsdoc --workspace @govuk-frontend/review
npm run build:sassdoc --workspace @govuk-frontend/review
npm run build:stats --workspace govuk-frontend
```

These will automatically appear in the Review app when available

domoscargin (approved):
This is great! Can definitely feel the difference when developing locally, and there's not as much gumph in my terminal to scroll through to find any relevant errors.

And the type checking seems really speedy too!

Issue #3923, opened 2023-07-07
Investigate long tags in task lists


owenatgov:
## What
Investigate the impact of long tags in task list components and apply affordances we feel are useful or necessary.

## Why
During the audit of our components against the new typography scale (https://github.com/alphagov/govuk-design-system/issues/2299) we noticed that larger tags start to look a little squashed with the larger font sizes on smaller screens. It was unclear what the impact of very large tags would be on the task list and if we should force them to not break eg: using `nowrap`.

## Timebox
half a sprint ie: 1 week

## Who is working on this?
Developers, designers

Spike lead:
TBD

Spike buddy:
TBD

## Questions to answer

- [x] How large a tag can the task list handle before it "breaks"?
- [x] Should we update the guidance to address use of long tags?

## Done when

- [x] Questions have been answered or we have a clearer idea of how to get to our goal
- [x] Findings have been reviewed and agreed with at least one other person
- [x] Findings have been shared, e.g: via a write-up on the ticket, at a show & tell or team meeting

CharlotteDowns:
Today (06/09/2023) we looked at the effects of long content in the [tag component](https://design-system.service.gov.uk/components/tag/) and task items within the new task list component.

## First experiment

Add longer than normal text to the tag component, observing how it wraps on desktop and mobile resolutions.

![On a standard desktop resolution width screen the tag component flows over two lines](https://github.com/alphagov/govuk-frontend/assets/56260216/60fb68c6-589f-4b50-b065-c80fd8862218)
![On a 320px resolution width screen the tag component flows over three lines](https://github.com/alphagov/govuk-frontend/assets/56260216/a58d1dfa-dd99-4ea7-af73-f287512191f0)

Although a bit clunky visually, it did not 'break' anything else within the page.

### Real service tag examples

> Waiting for induction to be recorded

(36 characters) _this tag has only been used in a table view and not a task list._

## Second experiment

Add longer than normal words to the tag component, observing what happens on desktop and mobile resolutions.

![On a 320px resolution width screen when a 23 character word is used in as the task list item, the tag component cannot be anymore than 5 characters without distortion occurring](https://github.com/alphagov/govuk-frontend/assets/56260216/630c585b-498b-4fd3-930e-2271c0bdaa64)

In this experiment the column container width became larger than expected. Although on desktop words of this length probably don't exist to cause distortion, on mobile (320px resolution) long words used in both the task and the tag may cause issue.

### Suggestions

[...]

Issue #1009, opened 2018-09-24
Select element max-width


davidjamesstone:
The `<select>` element's width is governed by the widths of its `<option>`'s.
When the text in the options grows large, the <select> element can grow to > 100% of the width of its container and break the layout.

Could I suggest a width: 100% or max-width: 100% to fix this.

Many thanks

nickcolley:
We have intentionally removed `width: 100%` for accessibility reasons but I think `max-width: 100%` would work fine to fix this, thanks for raising David!

PR #4259, opened 2023-09-25
Add word breaking for long single word tags

owenatgov:
Adds `overflow-wrap: break-word` to tags to split them in instances where we have very long tags in small paces like the task list.

More details can be found in the comments of https://github.com/alphagov/govuk-frontend/issues/3923

Some extra details on what else I tried can be found [in this comment](https://github.com/alphagov/govuk-frontend/issues/3923#issuecomment-1715690396).

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/tag/_index.scss
+++ b/packages/govuk-frontend/src/govuk/components/tag/_index.scss
@@ -1,8 +1,16 @@
 @include govuk-exports("govuk/component/tag") {
+  $govuk-tag-max-width: if(map-has-key($govuk-breakpoints, "mobile"), map-get($govuk-breakpoints, "mobile") / 2, 160px);
+
   .govuk-tag {
     @include govuk-font($size: 19);
 
     display: inline-block;
+
+    // set a max-width along with overflow-wrap: break-word below for instances
+    // where a tag has a single long word and could overflow its boundaries.
+    // The max-width is necessary as break-word requires a bounding box to base
+    // where to break off of.
+    max-width: $govuk-tag-max-width;
 
     // These negative margins make sure that the tag component doesn’t increase the
     // size of its container. Otherwise, for example, a table row containing a tag
@@ -16,11 +24,10 @@
     padding-right: 8px;
     padding-bottom: 3px;
     padding-left: 8px;
-
     color: govuk-shade(govuk-colour("blue"), 60%);
     background-color: govuk-tint(govuk-colour("blue"), 70%);
-
     text-decoration: none;
+    overflow-wrap: break-word;
 
     // When forced colour mode is active, for example to provide high contrast,
     // the background colour of the tag is the same as the rest of the page. To ensure
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ This change was made in:
 
 - [pull request #3502: Tag design changes](https://github.com/alphagov/govuk-frontend/pull/3502).
 - [pull request #3731: Remove the first letter modifier from the tag component](https://github.com/alphagov/govuk-frontend/pull/3731)
+- [pull request #4259: Add word breaking for long single word tags](https://github.com/alphagov/govuk-frontend/pull/4259)
 
 #### New link styles are now enabled by default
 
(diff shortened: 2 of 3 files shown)


owenatgov:
That percy screenshot is pretty bad 😬 I'm going to go back to the drawing board on this. Ideas welcome.

owenatgov:
I've managed to fix this with the following:
- applying a `max-width` of `160px` to tags
- using `overflow-wrap: break-word` instead of `word-break: break-all`

This should prevent instances where we don't want the tag to break because the word itself isn't very long eg: on the phase banner. The next step if this isn't suitable would be to only apply word breaking CSS to the task list tag.

packages/govuk-frontend/src/govuk/components/tag/_index.scss
```diff
@@ -4,6 +4,12 @@
 
     display: inline-block;
 
+    // set a max-width along with overflow-wrap: break-word below for instances
+    // where a tag has a single long word and could overflow it's boundaries.
```
frankieroberto, line 8:
```suggestion
    // where a tag has a single long word and could overflow its boundaries.
```

frankieroberto (approved):

packages/govuk-frontend/src/govuk/components/tag/_index.scss
```diff
@@ -4,6 +4,12 @@
 
     display: inline-block;
 
+    // set a max-width along with overflow-wrap: break-word below for instances
+    // where a tag has a single long word and could overflow its boundaries.
+    // The max-width is necessary as break-word requires a bounding box to base
+    // where to break off of.
+    max-width: 160px;
```
colinrotherham, line 11: If this comes from `$govuk-breakpoints` "mobile" (320px ÷ 2) should we add the calculation?

packages/govuk-frontend/src/govuk/components/tag/_index.scss
```diff
@@ -16,11 +22,10 @@
     padding-right: 8px;
     padding-bottom: 3px;
     padding-left: 8px;
-
     color: govuk-shade(govuk-colour("blue"), 60%);
     background-color: govuk-tint(govuk-colour("blue"), 70%);
-
     text-decoration: none;
+    overflow-wrap: break-word;
```
colinrotherham, line 28: Can we add a Review app example to show long tags in action?

Bit like [Summary list extreme](https://govuk-frontend-review.herokuapp.com/components/summary-list/extreme/preview) 😆

packages/govuk-frontend/src/govuk/components/tag/_index.scss
owenatgov: It doesn't but that's a neat coincidence. Maybe valuable to make it dynamic based on the breakpoints map.

colinrotherham (approved):
Looks good @owenatgov 🎉

Not a blocker for this PR but we probably need a per-component "overflow scroll" issue raising. Like in other components, Task list items with long links also break the layout 😣

Task list causing extra scrolling to the right

owenatgov:
@colinrotherham Thanks! Do you think that Task List overflow scroll isn't worth doing in this PR? It feels simple enough to add to this PR.

colinrotherham:
> @colinrotherham Thanks! Do you think that Task List overflow scroll isn't worth doing in this PR? It feels simple enough to add to this PR.

Not at all, entirely up to you if you want to add it

packages/govuk-frontend/src/govuk/components/task-list/_index.scss
```diff
@@ -8,6 +8,7 @@
     margin-top: 0;
     @include govuk-responsive-margin(6, "bottom");
     padding: 0;
+    overflow: scroll;
```
colinrotherham, line 11: Isn't this the rule that puts always-visible scrollbars on Windows? 😬

Both horizontal and vertical

colinrotherham: I think `<select>` was the best example in:

* https://github.com/alphagov/govuk-frontend/issues/1009

Users could add long `<option>` items with long widths, breaking the layout

Ideally we try our best to never add scroll bars?

colinrotherham: You can configure macOS to show scroll bars too

Scroll bars on macOS

<br>
<br>

At the moment both horizontal + vertical scroll bars show up?

Example showing scroll bars

Could narrow it down to `overflow-x` but `max-width` + text wrapping would be better

Shall we drop the commit for another PR?

owenatgov: I think so. I'll drop the commit, merge this and make an issue.

PR #4260, opened 2023-09-25
Fix double redirect to HTML specification

colinrotherham:
Just a few specification links that are double-redirecting at the moment

[**https://www.w3.org/TR/html51**/sec-forms.html](https://www.w3.org/TR/html51/sec-forms.html)
→ [**https://www.w3.org/TR/html**/sec-forms.html](https://www.w3.org/TR/html/sec-forms.html)
→ [**https://html.spec.whatwg.org/multipage**/sec-forms.html](https://html.spec.whatwg.org/multipage/sec-forms.html)


domoscargin (approved):
Neat!

PR #4261, opened 2023-09-25
Throw `ElementError` for missing elements during Character count instantiation

### LINKED ISSUE #4125: Throw errors during Character Count initialisation if key HTML elements are missing
## What

- The input/textarea
- The textarea description

## Why

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

romaricpascal:
Adds two errors thrown when:
- The textarea is missing or not an `HTMLTextareaElement` or an `HTMLInputElement`
- The textarea description is missing (preventing the component to replace its content with the live count)

Closes #4125

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/character-count/character-count.jsdom.test.mjs
+++ b/packages/govuk-frontend/src/govuk/components/character-count/character-count.jsdom.test.mjs
@@ -1,38 +1,22 @@
+import { getExamples, renderComponent } from '@govuk-frontend/lib/components'
+
 import { CharacterCount } from './character-count.mjs'
 
 describe('CharacterCount', () => {
-  let $container
-  let $textarea
-  let $textareaDescription
+  let html
 
-  function cloneContainerAndAddToDocument() {
-    const $clone = $container.cloneNode(true)
-    document.body.appendChild($clone)
-    return $clone
-  }
-
-  // Jest JSDOM does not clear the body between tests in the same file
-  // so we need to do it ourselves
-  // https://github.com/jestjs/jest/issues/1224#issuecomment-443661330
-  beforeEach(() => {
-    document.body.innerHTML = ''
+  beforeAll(async () => {
+    const examples = await getExamples('character-count')
+    html = renderComponent(
+      'character-count',
+      examples['to configure in JavaScript']
+    )
   })
 
-  beforeAll(() => {
-    $container = document.createElement('div')
-    $textarea = document.createElement('textarea')
-    $textareaDescription = document.createElement('span')
-
-    // Component checks that GOV.UK Frontend is supported
-    document.body.classList.add('govuk-frontend-supported')
-
-    // Component checks that required elements are present
-    $textarea.classList.add('govuk-js-character-count')
-    $textarea.id = 'textarea-id'
-    $container.appendChild($textarea)
-
-    $textareaDescription.id = 'textarea-id-info'
-    $container.appendChild($textareaDescription)
+  beforeEach(async () => {
+    // Some tests add attributes to `document.body` so we need
+    // to reset it alongside the component's markup
+    document.body.outerHTML = `<body class="govuk-frontend-supported">${html}</body>`
   })
 
   describe('formatCountMessage', () => {
@@ -40,8 +24,8 @@ describe('CharacterCount', () => {
       let componentWithMaxLength
       let componentWithMaxWords
 
-      beforeAll(() => {
-        const $div = cloneContainerAndAddToDocument()
+      beforeEach(() => {
+        const $div = document.querySelector('[data-module]')
         componentWithMaxLength = new CharacterCount($div, { maxlength: 100 })
         componentWithMaxWords = new CharacterCount($div, { maxwords: 100 })
       })
@@ -106,7 +90,7 @@ describe('CharacterCount', () => {
     describe('i18n', () => {
       describe('JavaScript configuration', () => {
         it('overrides the default translation keys', () => {
-          const $div = cloneContainerAndAddToDocument()
+          const $div = document.querySelector('[data-module]')
           const component = new CharacterCount($div, {
             maxlength: 100,
             i18n: {
@@ -127,7 +111,7 @@ describe('CharacterCount', () => {
         })
 
         it('uses specific keys for when limit is reached', () => {
-          const $div = cloneContainerAndAddToDocument()
+          const $div = document.querySelector('[data-module]')
           const componentWithMaxLength = new CharacterCount($div, {
             maxlength: 100,
             i18n: {
@@ -155,7 +139,7 @@ describe('CharacterCount', () => {
 
       describe('lang attribute configuration', () => {
         it('overrides the locale when set on the element', () => {
-          const $div = cloneContainerAndAddToDocument()
+          const $div = document.querySelector('[data-module]')
           $div.setAttribute('lang', 'de')
 
           const component = new CharacterCount($div, { maxwords: 20000 })
@@ -167,13 +151,9 @@ describe('CharacterCount', () => {
         })
 
         it('overrides the locale when set on an ancestor', () => {
-          const $parent = document.createElement('div')
-          $parent.setAttribute('lang', 'de')
+          document.body.setAttribute('lang', 'de')
 
-          const $div = $container.cloneNode(true)
-          $parent.appendChild($div)
-
-          document.body.appendChild($parent)
+          const $div = document.querySelector('[data-module]')
 
           const component = new CharacterCount($div, { maxwords: 20000 })
 
@@ -186,7 +166,7 @@ describe('CharacterCount', () => {
 
       describe('Data attribute configuration', () => {
         it('overrides the default translation keys', () => {
-          const $div = cloneContainerAndAddToDocument()
+          const $div = document.querySelector('[data-module]')
           $div.setAttribute(
             'data-i18n.characters-under-limit.one',
             'Custom text. Count: %{count}'
@@ -208,13 +188,11 @@ describe('CharacterCount', () => {
 
         describe('precedence over JavaScript configuration', () => {
           it('overrides translation keys', () => {
-            const $div = cloneContainerAndAddToDocument()
+            const $div = document.querySelector('[data-module]')
             $div.setAttribute(
               'data-i18n.characters-under-limit.one',
               'Custom text. Count: %{count}'
             )
-
-            console.log($div.outerHTML)
 
             const component = new CharacterCount($div, {
               maxlength: 100,
--- a/packages/govuk-frontend/src/govuk/components/character-count/character-count.jsdom.test.mjs
+++ b/packages/govuk-frontend/src/govuk/components/character-count/character-count.jsdom.test.mjs
@@ -3,17 +3,36 @@ import { CharacterCount } from './character-count.mjs'
 describe('CharacterCount', () => {
   let $container
   let $textarea
+  let $textareaDescription
+
+  function cloneContainerAndAddToDocument() {
+    const $clone = $container.cloneNode(true)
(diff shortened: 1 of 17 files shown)


packages/govuk-frontend/src/govuk/components/character-count/character-count.mjs
```diff
@@ -88,7 +88,11 @@ export class CharacterCount extends GOVUKFrontendComponent {
         $textarea instanceof HTMLInputElement
       )
     ) {
-      return this
+      throw new ElementError($textarea, {
+        componentName: 'Character count',
+        identifier: '.govuk-js-character-count',
+        expectedType: 'HTMLTextareaElement" or "HTMLInputElement'
```
romaricpascal, line 94: **nitpick**: That's a bit meh to have these closing/opening quotes in the middle of the string (option 1). Equally, it felt a bit meh to expand the types of the `expectedType` to `Array<typeof HTMLElement>` as it starts to make a bit too many situations to handle (option 2). 

Two alternatives to these two "meh" options:
3. drop the quotes. Are they really helpful? we're not quoting the identifier, for example
4. limit the `expectedType` to two types: `string|string[]`, focusing the constructor on working with strings, either injecting as is or joining with `,` and a final `or`.

1 works but is a bit of a sore sight, 2 looks like more complexity than we need for the formatting of an error message, 4 looks a bit less complex (but hit by proper handling of 1, 2, and more than 2 cases) so 3 looks like the more straightforward path.

@colinrotherham @domoscargin how does that sound? @domoscargin would updating anything on `expectedType` affect any work on the Tabs?

domoscargin: No impact on Tabs as far as I've got.

Think (3) sounds OK to me.

colinrotherham: Happy to drop the quotes, we don't use them for selectors

There are sneaky `string-but-maybe-already-array-to-array` things you can do to allow arrays too:

```
[expectedType].flat().join()
```

romaricpascal: Cool, I'll drop the quotes then, that'll make things simpler for now :) 

I think we're not quite geared for using `.flat()` yet as [we'd need a polyfill](https://browsersl.ist/#q=supports+es6-module+and+not+supports+array-flat%0A). We could work around it with a `Array.isArray(expectedType) ? expectedType.join(', ) : expectedType` (also more explicit, I think). Open for discussion in #4072.

romaricpascal:
Looks like we'll have to port the tests currently run in JSDom to Puppeteer. Even after adding an element for the textarea description, appending the markup to the `document` (so `document.getElementById` matches), [cleaning it up between tests](https://github.com/jestjs/jest/issues/1224#issuecomment-444586798), we run into the lack of [implementation of `innerText` from JSDom (hadn't expected that one 😱). ](https://github.com/jsdom/jsdom/issues/1245).

colinrotherham:
@romaricpascal Any quicker fixes to avoid `innerText`?

romaricpascal:
I think it's properly ready for review 😊 

@domoscargin The removal of the quotes has a tiny bit of overlap with the tests you may be writing for the Tabs. It should be fairly easy to update with a search and replace for `an instance of "(.*)`" and `of type "(.*)"` once merged. Equally, we could be leave it until we've got all the errors and tidy up their wording, keeping the quotes for now. Not fussed either way 😊

packages/govuk-frontend/src/govuk/components/character-count/character-count.jsdom.test.mjs
```diff
@@ -3,17 +3,36 @@ import { CharacterCount } from './character-count.mjs'
 describe('CharacterCount', () => {
   let $container
   let $textarea
+  let $textareaDescription
+
+  function cloneContainerAndAddToDocument() {
```
colinrotherham, line 8: Aware that we didn't need it previously, but with some tests continuing to do their own appending do we need a helper?

romaricpascal: It felt worth abstracting as all but one test use it straight away. The one that does not use it requires a parent wrapping the element. 

Now you mention it, though, wondering if injecting the HTML is not something that can be sorted in the global `beforeEach` that got introduce to reset the `document`. Each test could then do a `querySelector` similar to what we do in Puppeteer. Let me see if I can rework things in that direction and remove that helper :)

packages/govuk-frontend/src/govuk/components/character-count/character-count.mjs
```diff
@@ -88,7 +88,11 @@ export class CharacterCount extends GOVUKFrontendComponent {
         $textarea instanceof HTMLInputElement
       )
     ) {
-      return this
+      throw new ElementError($textarea, {
+        componentName: 'Character count',
+        identifier: '.govuk-js-character-count',
```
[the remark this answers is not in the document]
colinrotherham: Linking with https://github.com/alphagov/govuk-frontend/pull/4263#pullrequestreview-1644001470

For these "module child" selectors it could be brilliant to show a scope that the browser can inspect

```suggestion
        identifier: '.govuk-js-character-count',
        scope: $module,
```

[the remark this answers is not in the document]
romaricpascal: 🤔 Yeah, that's probably the component most likely to be on a page multiple times. We recommend to [start with a single question per page](https://design-system.service.gov.uk/patterns/question-pages/#start-by-asking-one-question-per-page) so that limits the risks. It doesn't completely avoids the possibility of multiple of those on the page (especially if we consider the radios/checkboxes potentially revealing fields). That said, there shouldn't be too many of them on a given page that it's tricky for developers to find back the one that's wonky (either by looking carefully at the page or breaking on error in the devtools). I would say it's not something we should not worry too hard about at the start. We can always dig deeper if we receive feedback.

If we have a simple solution, though, not against adding extra information, but it might be tricky to add that consistently. The element with `data-module` does not really have anything identifiable usually. Here, the `$textarea` could provide an `id` (but only if it is on the page so nothing if it doesn't exist).

packages/govuk-frontend/src/govuk/components/character-count/character-count.jsdom.test.mjs
```diff
@@ -187,12 +208,14 @@ describe('CharacterCount', () => {
...
         describe('precedence over JavaScript configuration', () => {
           it('overrides translation keys', () => {
-            const $div = $container.cloneNode(true)
+            const $div = cloneContainerAndAddToDocument()
             $div.setAttribute(
               'data-i18n.characters-under-limit.one',
               'Custom text. Count: %{count}'
             )
 
+            console.log($div.outerHTML)
```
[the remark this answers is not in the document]
romaricpascal: Whoops, missed that one, sorry 😔

packages/govuk-frontend/src/govuk/components/character-count/character-count.mjs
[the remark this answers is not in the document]
romaricpascal: Whoops, there wasn't your second message when I started writing mine, sorry. How would users access the `$module` you propose passing as `$scope` (it's fine if that part still needs looking into)? Do errors log their extra fields? the `cause` maybe, in browsers that support it (that'd be neat)?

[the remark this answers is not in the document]
colinrotherham: One to investigate. Would be amazing if you could hover over `scope` and it highlighted the element

Can leave them off for now if we don't need it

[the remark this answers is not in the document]
romaricpascal: That'd speed debugging indeed, but let's see if people come back needing more pointers to where things happen as I think that'd need a good bit of investigation 😊 Had a quick glance at `cause`, Firefox logs it (only if the element itself, if an object, you get `Object {...}`, not expandable), not Chrome, nor Safari (at the time of this message). We could always override `toString` to log before the message gets computed for display in the console, but seems quite hacky anyway 😆

packages/govuk-frontend/src/govuk/components/character-count/character-count.jsdom.test.mjs
```diff
@@ -1,28 +1,32 @@
...
-
-    // Component checks that GOV.UK Frontend is supported
-    document.body.classList.add('govuk-frontend-supported')
+  let html
+
+  // Reset to a clean component within a clean `<body>` tag
+  beforeEach(async () => {
+    if (!html) {
+      const examples = await getExamples('character-count')
+      html = renderComponent(
```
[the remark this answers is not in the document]
colinrotherham: Assign `html` in the `beforeAll()` if you like?

Saves the conditional `!html`?

```mjs
let examples
let html

beforeAll(async () => {
  examples = await getExamples('character-count')

  html = renderComponent(
    'character-count',
    examples['to configure in JavaScript']
  )
})
```

packages/govuk-frontend/src/govuk/components/character-count/character-count.jsdom.test.mjs
```diff
@@ -1,28 +1,32 @@
...
+        examples['to configure in JavaScript']
+      )
+    }
 
-    // Component checks that required elements are present
-    $textarea.classList.add('govuk-js-character-count')
-    $container.appendChild($textarea)
+    document.documentElement.innerHTML = `
+      <body class="govuk-frontend-supported">${html}</body>
+    `
```
colinrotherham, line 20: Ah was it too risky using `document.body.innerHTML = html`?

romaricpascal: No, it's done to reset the `<body>` tag to a clean state after a test adds a `lang` attribute to it. I've added a comment so we know why it's that way in the future 😊

colinrotherham: Yeah that makes sense. Shall we keep `outerHTML` and `document.body` for familiarity?

```suggestion
    document.body.outerHTML = `<body class="govuk-frontend-supported">${html}</body>`
```

packages/govuk-frontend/src/govuk/components/character-count/character-count.mjs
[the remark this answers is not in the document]
romaricpascal: Sorry, not the clearest blurb. I'll (try to) clarify:

- I'd prefer leaving it off until we have the need for it as it'd require more investigation to make it work
- Got curious and gave a quick check at where Error's `cause` are logged (no luck there) and tried a quick hack by overriding `toString` to log before the error message gets computed (can log the element, but a bit hacky).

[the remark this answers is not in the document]
colinrotherham: No don't worry, the confused face was wondering why `cause` wasn't logged

One for another day (and don't want to be Chrome specific) but some nice ideas in:
https://developer.chrome.com/docs/devtools/console/log/

packages/govuk-frontend/src/govuk/components/character-count/character-count.jsdom.test.mjs
colinrotherham: Cheers. If you want to keep the new lines (but without trailing spaces) wrap the HTML in `outdent`?

romaricpascal: Not fussy about the whitespace, I just had edited the left side of the `=`.

colinrotherham (approved):

PR #4263, opened 2023-09-26
Throw errors during Tabs initialisation if key HTML elements are missing

### LINKED ISSUE #4127: Throw errors during Tabs initialisation if key HTML elements are missing
## What

- The tabs (`<a>`)
- The tablist and their items
- The tab panel associated to a tab

## Why

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

domoscargin:
Closes #4127 

A coupla notes:

- So far, I've limited this PR to replacing empty returns in the initialisation of the component. There are 17 more empty returns in the component, some of which are bound during initialisation. These are in event handlers, however, which we can refactor as and when we add things to the API.
- I've removed a check in `setup()` as unnecessary. Some checks will need to replace it if `setup()` becomes an API method that can be called separately from initialisation, but I figure we'd probably do a slight code refactor then anyway.

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/tabs/tabs.mjs
+++ b/packages/govuk-frontend/src/govuk/components/tabs/tabs.mjs
@@ -141,9 +141,6 @@ export class Tabs extends GOVUKFrontendComponent {
 
     // Show either the active tab according to the URL's hash or the first tab
     const $activeTab = this.getTab(window.location.hash) || this.$tabs[0]
-    if (!$activeTab) {
-      return
-    }
 
     this.showTab($activeTab)
 
--- a/packages/govuk-frontend/src/govuk/components/tabs/tabs.mjs
+++ b/packages/govuk-frontend/src/govuk/components/tabs/tabs.mjs
@@ -81,11 +81,15 @@ export class Tabs extends GOVUKFrontendComponent {
     // MediaQueryList.addEventListener isn't supported by Safari < 14 so we need
     // to be able to fall back to the deprecated MediaQueryList.addListener
     if ('addEventListener' in this.mql) {
-      this.mql.addEventListener('change', () => this.checkMode())
+      this.mql.addEventListener('change', () => {
+        this.checkMode()
+      })
     } else {
       // @ts-expect-error Property 'addListener' does not exist
       // eslint-disable-next-line @typescript-eslint/no-unsafe-call
-      this.mql.addListener(() => this.checkMode())
+      this.mql.addListener(() => {
+        this.checkMode()
+      })
     }
 
     this.checkMode()
@@ -112,12 +116,10 @@ export class Tabs extends GOVUKFrontendComponent {
   setup() {
     const $tabList = this.$module.querySelector('.govuk-tabs__list')
     const $tabListItems = this.$module.querySelectorAll(
-      '.govuk-tabs__list-item'
+      'li.govuk-tabs__list-item'
     )
 
-    if (!this.$tabs || !$tabList || !$tabListItems) {
-      return
-    }
+    this.checkTabList($tabList, $tabListItems)
 
     $tabList.setAttribute('role', 'tablist')
 
@@ -157,12 +159,10 @@ export class Tabs extends GOVUKFrontendComponent {
   teardown() {
     const $tabList = this.$module.querySelector('.govuk-tabs__list')
     const $tabListItems = this.$module.querySelectorAll(
-      'a.govuk-tabs__list-item'
+      'li.govuk-tabs__list-item'
     )
 
-    if (!this.$tabs || !$tabList || !$tabListItems) {
-      return
-    }
+    this.checkTabList($tabList, $tabListItems)
 
     $tabList.removeAttribute('role')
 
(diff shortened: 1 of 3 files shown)


packages/govuk-frontend/src/govuk/components/tabs/tabs.mjs
```diff
@@ -53,7 +53,10 @@ export class Tabs extends GOVUKFrontendComponent {
     /** @satisfies {NodeListOf<HTMLAnchorElement>} */
     const $tabs = $module.querySelectorAll('a.govuk-tabs__tab')
     if (!$tabs.length) {
-      return this
+      throw new ElementError(null, {
+        componentName: 'Tabs',
+        identifier: 'a.govuk-tabs__tab'
```
colinrotherham, line 58: Wondering if we need to show this _scoped_ via `$module` as it's not from `document` here?

```suggestion
        identifier: `[data-module="${Tabs.moduleName}"] a.govuk-tabs__tab`
```

packages/govuk-frontend/src/govuk/components/tabs/tabs.mjs
```diff
@@ -64,6 +67,28 @@ export class Tabs extends GOVUKFrontendComponent {
...
+
+    if (!$tabListItems.length) {
+      throw new ElementError(null, {
+        componentName: 'Tabs',
+        identifier: '.govuk-tabs__list-item'
+      })
+    }
+
+    this.$tabList = $tabList
+    this.$tabListItems = $tabListItems
```
colinrotherham, line 90: Love it, thanks @domoscargin

If you can add them as class fields too?

```mjs
  /** @private */
  $tabs

  /** @private */
  $tabList

  /** @private */
  $tabListItems
```

Won't need types as they can be inferred in the constructor

packages/govuk-frontend/src/govuk/components/tabs/tabs.mjs
```diff
@@ -107,18 +136,9 @@ export class Tabs extends GOVUKFrontendComponent {
    * @private
    */
   setup() {
-    const $tabList = this.$module.querySelector('.govuk-tabs__list')
-    const $tabListItems = this.$module.querySelectorAll(
-      '.govuk-tabs__list-item'
-    )
-
-    if (!this.$tabs || !$tabList || !$tabListItems) {
```
colinrotherham, line 115: 🙌

packages/govuk-frontend/src/govuk/components/tabs/tabs.mjs
domoscargin: Went with the first one since we've been doing it that way so far, but we can always revisit when we review these.

packages/govuk-frontend/src/govuk/components/tabs/tabs.mjs
domoscargin: Donezos

packages/govuk-frontend/src/govuk/components/tabs/tabs.mjs
colinrotherham: I asked @romaricpascal the same question on this thread here: https://github.com/alphagov/govuk-frontend/pull/4261#discussion_r1337169400

Will leave you two to decide what's best

romaricpascal: I think going for the `scope` thing will require quite a bit of investigation so I'd prefer we only look at it if needs arise.

Regarding the addition of `[data-module="${Tabs.moduleName}"]`, which new information does it bring in addition to:
- the name of the component that starts the error messages
- our classnames which usually refer to the component as well

colinrotherham: It came up [whilst discussing Checkboxes](https://github.com/alphagov/govuk-frontend/pull/4236#discussion_r1331778120) since realising it's only **Error summary**, **Radios**, **Checkboxes** and **Exit this page** that use `document.querySelector()` versus `$module.querySelector()` etc

So wondered if we should explain that or not, or undo previous scoping as part of https://github.com/alphagov/govuk-frontend/issues/4072

E.g. In previous PRs we've [scoped via `$module.hash target #${linkedElementId}`](https://github.com/alphagov/govuk-frontend/commit/fd7a8ed7e10d0dc8f4a362f0e7482839d0a43534#diff-fb3a25d42cbb3621d375cc34dc960447e84fa2a8729fd16e97e4c34edd9b17aeR66)

romaricpascal: Let's see once we have the whole list of errors for #4072. That'll give us a clear view of which component parts are referenced. And we can decide on which extra prefix we may need to better locate the faulty element on top of the name of the component that couldn't initialise.

colinrotherham: So drop the prefix in this PR like you've done in https://github.com/alphagov/govuk-frontend/pull/4261?

colinrotherham (commented):
This is looking ace @domoscargin

Just merged https://github.com/alphagov/govuk-frontend/pull/4261 so let's give this one a rebase and [drop the `[data-module="${Tabs.moduleName}"]` prefix](https://github.com/alphagov/govuk-frontend/pull/4263#discussion_r1337536858) then think we're good to go 🚀

packages/govuk-frontend/src/govuk/components/tabs/tabs.mjs
```diff
@@ -64,6 +73,28 @@ export class Tabs extends GOVUKFrontendComponent {
     this.boundTabKeydown = this.onTabKeydown.bind(this)
     this.boundOnHashChange = this.onHashChange.bind(this)
 
+    const $tabList = this.$module.querySelector('.govuk-tabs__list')
+    const $tabListItems = this.$module.querySelectorAll(
+      'li.govuk-tabs__list-item'
+    )
+
+    if (!$tabList) {
+      throw new ElementError($tabList, {
```
colinrotherham, line 82: Sorry, this old chestnut again 😆

```suggestion
      throw new ElementError(null, {
```

colinrotherham (approved):
✅ Approved

(One little comment, but only a quick tweak)

packages/govuk-frontend/src/govuk/components/tabs/tabs.mjs
colinrotherham: Not caused by this PR, but should really match up the JSDoc at the other end

i.e. Adding `| null` to the allowed types

```mjs
/**
 * @param {Element | null} element - The element in error
 */
```

domoscargin: Implemented the suggestion and add `| null` to the ElementError JSDoc.

Issue #4230, opened 2023-09-19
Prevent JavaScript component config being flattened


colinrotherham:
## Description of the issue

Our JavaScript component configs are being flattened in `mergeConfigs()` and don't match [JSDoc output](https://govuk-frontend-review.herokuapp.com/docs/javascript/classes/Accordion.html#defaults)

```mjs
this.config = mergeConfigs(
  Accordion.defaults,
  config || {},
  normaliseDataset($module.dataset)
)
```

## Steps to reproduce the issue

Instantiate a component (e.g. Accordion) using the JavaScript API

### Before
Console output for `Accordion.defaults` shows an unflattened config

```mjs
{
  i18n: {
    hideAllSections: 'Hide all sections',
    hideSection: 'Hide',
    hideSectionAriaLabel: 'Hide this section',
    showAllSections: 'Show all sections',
    showSection: 'Show',
    showSectionAriaLabel: 'Show this section'
  },
  rememberExpanded: true
}
```

### After
Console output for `new Accordion().config` shows an unexpectedly flattened config

```mjs
{
  'i18n.hideAllSections': 'Hide all sections',
  'i18n.hideSection': 'Hide',
  'i18n.hideSectionAriaLabel': 'Hide this section',
  'i18n.showAllSections': 'Show all sections',
  'i18n.showSection': 'Show',
  'i18n.showSectionAriaLabel': 'Show this section'
  module: 'govuk-accordion',
  rememberExpanded: true
}
```

Both are accidentally typed with `AccordionConfig` even though they're different

## Actual vs expected behaviour

Whilst we've picked "flattening" to our method to normalise `i18n` text between `data-*` attributes and the JavaScript API, the final merged `new Accordion().config` is expected to match the JavaScript API format in [JSDoc output](https://govuk-frontend-review.herokuapp.com/docs/javascript/classes/Accordion.html#defaults).

We hold our JavaScript component configs in two places:

1. Using static fields for defaults, e.g. `Accordion.defaults`
2. Using class fields once merged, e.g. `new Accordion().config`

We merge component config `data-*` attributes over JavaScript config, using the defaults to fill the gaps

colinrotherham:
To avoid confusion, we should distinguish `mergeConfigs()` from a new `flattenConfigs()` helper

Then we may want to clearly flatten and extract only **i18n** key/values as shown:

```mjs
this.config = mergeConfigs(Accordion.defaults, config)

// Component data attributes
[...]

PR #4266, opened 2023-09-27
Throw errors during accordion initialisation

### LINKED ISSUE #4126: Throw errors during Accordion initialisation if key HTML elements are missing
## What

- The sections of the accordion
- Header in the section
- `<span>` header
- Heading
- Toggle icon or text

## Why

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

domoscargin:
Closes #4126

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
+++ b/packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
@@ -139,7 +139,10 @@ export class Accordion extends GOVUKFrontendComponent {
 
     const $sections = this.$module.querySelectorAll(`.${this.sectionClass}`)
     if (!$sections.length) {
-      return this
+      throw new ElementError({
+        componentName: 'Accordion',
+        identifier: `Sections (\`.${this.sectionClass}\`)`
+      })
     }
 
     this.$sections = $sections
--- a/packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
+++ b/packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
@@ -411,7 +411,22 @@ export class Accordion extends GOVUKFrontendComponent {
     const $button = $section.querySelector(`.${this.sectionButtonClass}`)
     const $content = $section.querySelector(`.${this.sectionContentClass}`)
 
-    if (!$showHideIcon || !$showHideText || !$button || !$content) {
+    if (!$button) {
+      throw new ElementError({
+        componentName: 'Accordion',
+        identifier: `Section button (\`.${this.sectionButtonClass}\`)`
+      })
+    }
+
+    if (!$content) {
+      throw new ElementError({
+        componentName: 'Accordion',
+        identifier: `Section content (\`.${this.sectionContentClass}\`)`
+      })
+    }
+
+    if (!$showHideIcon || !$showHideText) {
+      // Return early for elements we create
       return
     }
 
--- a/packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
+++ b/packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
@@ -237,8 +237,18 @@ export class Accordion extends GOVUKFrontendComponent {
     const $heading = $header.querySelector(`.${this.sectionHeadingClass}`)
     const $summary = $header.querySelector(`.${this.sectionSummaryClass}`)
 
-    if (!$span || !$heading) {
-      return
+    if (!$heading) {
+      throw new ElementError({
+        componentName: 'Accordion',
+        identifier: `Section heading (\`.${this.sectionHeadingClass}\`)`
+      })
+    }
+
+    if (!$span) {
+      throw new ElementError({
+        componentName: 'Accordion',
+        identifier: `Section button (\`.${this.sectionButtonClass}\`)`
+      })
     }
 
     // Create a button element that will replace the
--- a/packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
+++ b/packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
@@ -206,7 +206,10 @@ export class Accordion extends GOVUKFrontendComponent {
     this.$sections.forEach(($section, i) => {
       const $header = $section.querySelector(`.${this.sectionHeaderClass}`)
       if (!$header) {
-        return
+        throw new ElementError({
+          componentName: 'Accordion',
+          identifier: `Section headers (\`.${this.sectionHeaderClass}\`)`
+        })
       }
 
       // Set header attributes
--- a/packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
+++ b/packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
@@ -141,7 +141,7 @@ export class Accordion extends GOVUKFrontendComponent {
     if (!$sections.length) {
       throw new ElementError({
         componentName: 'Accordion',
-        identifier: `Sections (\`.${this.sectionClass}\`)`
+        identifier: `Sections (\`<div class="${this.sectionClass}">\`)`
       })
     }
 
@@ -208,7 +208,7 @@ export class Accordion extends GOVUKFrontendComponent {
       if (!$header) {
         throw new ElementError({
           componentName: 'Accordion',
-          identifier: `Section headers (\`.${this.sectionHeaderClass}\`)`
+          identifier: `Section headers (\`<div class="${this.sectionHeaderClass}">\`)`
         })
       }
 
@@ -247,7 +247,7 @@ export class Accordion extends GOVUKFrontendComponent {
     if (!$span) {
       throw new ElementError({
         componentName: 'Accordion',
-        identifier: `Section button (\`.${this.sectionButtonClass}\`)`
+        identifier: `Section button (\`<span class="${this.sectionButtonClass}">\`)`
       })
     }
 
@@ -414,14 +414,14 @@ export class Accordion extends GOVUKFrontendComponent {
     if (!$button) {
       throw new ElementError({
         componentName: 'Accordion',
-        identifier: `Section button (\`.${this.sectionButtonClass}\`)`
+        identifier: `Section button (\`<span class="${this.sectionButtonClass}">\`)`
       })
     }
 
     if (!$content) {
       throw new ElementError({
         componentName: 'Accordion',
-        identifier: `Section content (\`.${this.sectionContentClass}\`)`
+        identifier: `Section content (\`<div class="${this.sectionContentClass}">\`)`
       })
     }
 
--- a/packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
+++ b/packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
@@ -247,7 +247,7 @@ export class Accordion extends GOVUKFrontendComponent {
     if (!$span) {
       throw new ElementError({
         componentName: 'Accordion',
-        identifier: `Section button (\`<span class="${this.sectionButtonClass}">\`)`
+        identifier: `Section button placeholder (\`<span class="${this.sectionButtonClass}">\`)`
       })
     }
 
@@ -411,13 +411,6 @@ export class Accordion extends GOVUKFrontendComponent {
     const $button = $section.querySelector(`.${this.sectionButtonClass}`)
     const $content = $section.querySelector(`.${this.sectionContentClass}`)
 
-    if (!$button) {
-      throw new ElementError({
-        componentName: 'Accordion',
-        identifier: `Section button (\`<span class="${this.sectionButtonClass}">\`)`
-      })
-    }
-
     if (!$content) {
       throw new ElementError({
         componentName: 'Accordion',
@@ -425,7 +418,7 @@ export class Accordion extends GOVUKFrontendComponent {
       })
     }
 
-    if (!$showHideIcon || !$showHideText) {
+    if (!$showHideIcon || !$showHideText || !$button) {
       // Return early for elements we create
       return
     }
--- a/packages/govuk-frontend/src/govuk/components/accordion/accordion.puppeteer.test.js
+++ b/packages/govuk-frontend/src/govuk/components/accordion/accordion.puppeteer.test.js
@@ -783,7 +783,7 @@ describe('/components/accordion', () => {
             cause: {
               name: 'ElementError',
               message:
-                'Accordion: Sections (`.govuk-accordion__section`) not found'
+                'Accordion: Sections (`<div class="govuk-accordion__section">`) not found'
             }
           })
         })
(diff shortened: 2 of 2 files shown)


packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
```diff
@@ -394,18 +410,30 @@ export class Accordion extends GOVUKFrontendComponent {
...
-      !($showHideText instanceof HTMLElement) ||
-      !$button ||
-      !$content
-    ) {
+    // Return early for elements created during set up
+    if (!$showHideIcon || !$showHideText) {
       return
     }
 
+    if (!$button || !$content) {
```
colinrotherham, line 422: What do think about merging all these into one?

```suggestion
    if (!$showHideIcon || $showHideText || !$button || !$content) {
```

packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
```diff
@@ -394,18 +410,30 @@ export class Accordion extends GOVUKFrontendComponent {
...
+        [$content, this.sectionContentClass]
+      ])
+        // Filter the list by which checks failed
+        .filter(([$element]) => !($element instanceof HTMLElement))
+
+      throw new ElementError(check[0], {
+        componentName: 'Accordion',
+        identifier: `.${check[1]}`,
+        expectedType: 'HTMLElement'
+      })
```
colinrotherham, line 434: You'd have to guard against an `undefined` check when only two selectors are the array

```suggestion
      if (check) {
        throw new ElementError(check[0], {
          componentName: 'Accordion',
          identifier: `.${check[1]}`,
          expectedType: 'HTMLElement'
        })
      }
      
      // Return early for elements created during set up
      return
```

packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
domoscargin: I thought we'd made the call to silently return for the first two? Do you mean have 'em all in one check, then only throw for the relevant ones?

packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
```diff
@@ -394,18 +410,30 @@ export class Accordion extends GOVUKFrontendComponent {
...
-    ) {
+    // Return early for elements created during set up
+    if (!$showHideIcon || !$showHideText) {
       return
     }
 
+    if (!$button || !$content) {
+      const [check] = /** @type {const} */ ([
+        [$button, this.sectionButtonClass],
+        [$content, this.sectionContentClass]
```
colinrotherham, line 425: Alternatively, with only two selectors, is it worth having an array + filter?

Was a better idea when we had 4x elements

domoscargin: Yeah, this is a great point. I'll do some rejigging.

packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
domoscargin: Ah, right! Sorry, didn't click that the comments were connected. It makes sense now, thanks!

packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
domoscargin: I've split this out into a check for `$button`, a check for `$content`, and a check for the two internally created classes.

It's a similar amount of code, and I think makes things super easy to parse.

packages/govuk-frontend/src/govuk/components/accordion/accordion.puppeteer.test.js
```diff
@@ -712,148 +712,160 @@ describe('/components/accordion', () => {
             )
           })
         })
+      })
 
-        describe('errors at instantiation', () => {
-          let examples
+      describe('errors at instantiation', () => {
```
domoscargin, line 717: Noticed we accidentally had our `errors at instantiation` wrapped within the `localisation` block.

packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
```diff
@@ -122,7 +122,7 @@ export class Accordion extends GOVUKFrontendComponent {
     if (!($module instanceof HTMLElement)) {
       throw new ElementError($module, {
         componentName: 'Accordion',
-        identifier: '$module'
+        identifier: 'Root element (`[data-module=govuk-accordion]`)'
```
domoscargin, line 125: I've tried to get all the identifiers looking as described in #4299

packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
```diff
@@ -140,7 +140,7 @@ export class Accordion extends GOVUKFrontendComponent {
     if (!$sections.length) {
       throw new ElementError(null, {
         componentName: 'Accordion',
-        identifier: `.${this.sectionClass}`
+        identifier: `Element (\`.${this.sectionClass}\`)`
```
domoscargin, line 143: Wrinkle here: 

I've just gone with using `Element` any time the element in question is a `div`, but do we want to be specific? Or add a less generic qualifier?

For example, here would we prefer something like `Section element`? (Though not _that_, because `<section>` is a thing).

colinrotherham: Happy to drop "element" and just use the plain English name?

1. **Accordion:** Module (`[data-module=govuk-accordion]`)
1. **Accordion:** Section (`.govuk-accordion__section`)
1. **Accordion:** Section header (`.govuk-accordion__section-header`)
1. **Accordion:** Section heading (`.govuk-accordion__section-heading`)
1. **Accordion:** Section button (`.govuk-accordion__section-button`)
1. etc 

The word "Section" then doesn't have to mean `<section>` element

domoscargin:
Rebased - allows this PR to be simplified, since all the `beforeInitialisation` wrangling has already been merged.

packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
```diff
@@ -411,7 +411,22 @@ export class Accordion extends GOVUKFrontendComponent {
     const $button = $section.querySelector(`.${this.sectionButtonClass}`)
     const $content = $section.querySelector(`.${this.sectionContentClass}`)
 
-    if (!$showHideIcon || !$showHideText || !$button || !$content) {
+    if (!$button) {
```
domoscargin, line 414: Note this is our second check for `$button`, so I haven't written another test for it. I guess technically you could trigger this slightly differently from the other instance, but the end result is an error. This will be neater if we figure out some kind of caching solution.

packages/govuk-frontend/src/govuk/components/accordion/accordion.puppeteer.test.js
```diff
@@ -760,6 +760,25 @@ describe('/components/accordion', () => {
...
+                  .forEach((item) => item.remove())
+              },
+              context: {
+                selector: '.govuk-accordion__section'
+              }
+            })
+          ).rejects.toEqual({
+            name: 'ElementError',
+            message:
+              'Accordion: Sections (`.govuk-accordion__section`) not found'
```
domoscargin, line 779: I'm tempted to suggest we have a new kind of message in `ElementError` of this sort:

`No sections (<IDENTIFIER>) were found`

For when a group of things is empty.

But I wouldn't consider this blocking.

packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
```diff
@@ -139,7 +139,10 @@ export class Accordion extends GOVUKFrontendComponent {
 
     const $sections = this.$module.querySelectorAll(`.${this.sectionClass}`)
     if (!$sections.length) {
-      return this
+      throw new ElementError({
+        componentName: 'Accordion',
+        identifier: `Sections (\`.${this.sectionClass}\`)`
```
colinrotherham, line 144: Are we alright moving all these to HTML still?

```suggestion
        identifier: `Sections (\`<div class="govuk-accordion__section">\`)`
```

romaricpascal: I'd be enclined to use HTML too (as well as for the other elements that may be missing) 😊

packages/govuk-frontend/src/govuk/components/accordion/accordion.puppeteer.test.js
romaricpascal: Yeah, it does sound better, indeed, when multiple things are missing. Agreed this wouldn't block the PR and can be dealt with separately (wouldn't even block the pre-release, as the content is not part of the public API).

packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
domoscargin: Switched things to HTML where possible.

Section headings change tag depending on the `headingLevel` param (and I don't have time before I clock off to check how easy that'd be to use in generating the message).

colinrotherham: Yeah happy to add it, but guessing that promotes it to "JavaScript config" land so will cause issues 😣

What if the Nunjucks is rendered with `headingLevel: 2` but JS initialised with `headingLevel: 3`??

We could use [`Omit`](https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys) to exclude it from the JavaScript config so it's available post-merge with the `data-*` attributes but can't be passed in, but that's something I've been looking at in https://github.com/alphagov/govuk-frontend/issues/4230

Or go ahead and update the JSDoc like you said but use `validateConfig()` to ban it

romaricpascal: I'm not sure we should forward Nunjucks config all the way down to the JavaScript component just for formatting an error message. That's also asking people using raw HTML to have that attribute set correctly, same for libraries implementing our templates in other languages that would need to update. I find it quite and ask for the benefit it brings.

A couple of other options:

1. Use a generic `hX` or `hN` name for the tag... might lead people to have actual `hX` tag in their markup
    ```
    `Section heading (\`<hX class="${this.sectionHeadingClass}">\`)`
    ```

2. Default to `h2`, at least they'll have a valid tag, but it may mess their heading hierarchy if they copy-paste
   ```
   `Section heading (\`<h2 class="${this.sectionHeadingClass}">\`)`
   ```
3. List the possible tag names, people will need to edit anyway
   ```
   `Section heading (\`<{h2,h3,h4,h5,h6} class="${this.sectionHeadingClass}">\`)`
   ```
4. Keep CSS for that one, 
    ```
    `Section heading (\`.${this.sectionHeadingClass}\`)`
    ```
    We could even highlight the name of the tag using the `:is()` selector
    ```
    `Section heading (\`:is(h2,h3,h4,h5,h6).${this.sectionHeadingClass}\`)`
    ```

1,2 and 3 seem quite prone to copy-pasting errors, so I'd lean towards 4 (in its simplest form to avoid one extra update), with 1 seeming the least worst of the HTML options if we prefer HTML.

packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
```diff
@@ -231,8 +237,18 @@ export class Accordion extends GOVUKFrontendComponent {
...
+      throw new ElementError({
+        componentName: 'Accordion',
+        identifier: `Section heading (\`.${this.sectionHeadingClass}\`)`
+      })
+    }
+
+    if (!$span) {
+      throw new ElementError({
+        componentName: 'Accordion',
+        identifier: `Section button (\`<span class="${this.sectionButtonClass}">\`)`
```
romaricpascal, line 250: **nitpick** Maybe a little pedantic, but it looked weird to have "button" followed by a `span` tag.

```suggestion
        identifier: `Section button placeholder (\`<span class="${this.sectionButtonClass}">\`)`
```

packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
```diff
@@ -395,7 +411,22 @@ export class Accordion extends GOVUKFrontendComponent {
     const $button = $section.querySelector(`.${this.sectionButtonClass}`)
     const $content = $section.querySelector(`.${this.sectionContentClass}`)
 
-    if (!$showHideIcon || !$showHideText || !$button || !$content) {
+    if (!$button) {
+      throw new ElementError({
+        componentName: 'Accordion',
+        identifier: `Section button (\`<span class="${this.sectionButtonClass}">\`)`
```
romaricpascal, line 417: **nitpick** Technically at that stage, it's an actual `button` tag, so we could either:

```suggestion
        identifier: `Section button (\`<button class="${this.sectionButtonClass}">\`)`
```

Which would help people look for a button, though what they likely need to do for fixing is making sure the placeholder is in so we turn it into a button with:

```suggestion
        identifier: `Section button placeholder (\`<span class="${this.sectionButtonClass}">\`)`
```

romaricpascal (commented):
Spotted two little updates we could make to avoid "button" issues mentioning a `<span>` element, but happy for them to happen later on 😊

packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
colinrotherham: Thanks for spotting this

It's a quirk of Accordion that's caught me out too, that `this.sectionButtonClass` matches different things

Should we instead return early for `!$button` here, knowing we've thrown already for the placeholder?

romaricpascal: Good idea! At this stage, we're the ones that created the `<button>` so we can early return like in other listener for now 😊

colinrotherham:
@domoscargin I've rebased this PR with https://github.com/alphagov/govuk-frontend/pull/4329 and https://github.com/alphagov/govuk-frontend/pull/4345

We've renamed `renderAndInitialise()` to `render()` as not all components have JavaScript (to initialise)

packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
domoscargin: Yeah, I'd also favour just keeping it as is rather than wrangle with trying to get the exact right heading level. Will keep it as

````Section heading (\`.${this.sectionHeadingClass}\`)````

packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
domoscargin: Thanks for taking a look, folks. I've now made these tweaks in the latest commit.

colinrotherham (approved):

Issue #4242, opened 2023-09-21
Check improper use of `govuk-font` within GOV.UK Frontend


owenatgov:
## What
Check GOV.UK Frontend for any improper use of `govuk-font` ie: where it's producing repeated code and could instead be replaced with a different typography mixin like `govuk-typography-responsive`, and fix it.

## Why
It emerged during https://github.com/alphagov/govuk-design-system/issues/2300 that users may misunderstand how `govuk-font` works and end up using it twice across different context, resulting in repeated code when they only wanted to change the font size or weight. We reckon that we are also misusing `govuk-font` in places.

## Who needs to work on this
Developers

## Who needs to review this
Developers

## Done when

- [x] The codebase has been searched for any improper use of `govuk-font` and findings are written up
- [ ] A PR is raised to fix improper uses, if any

owenatgov:
From static searching, 8 components are using `govuk-font` more than once when they should instead be using `govuk-typography-responsive` or one of the font weight mixins:

- Fieldset
- Header
- Input
- Label
- Notification banner
- Phase banner
- Summary list
- Table

There are also 7 components where `gvuk-font` isn't being used improperly but we can move the declaration earlier up the chain to make the component more efficient:

- Checkboxes
- Error summary
- Pagination
- Panel
- Radios
- Tabs
- Warning text

So in total I'm anticipating **15 changes** to our sass.

PR #4267, opened 2023-09-27
Remove unnecessary duplicated use of govuk-font

owenatgov:
## What/Why

Remove instances of when our components are using `govuk-font` more than once when they don't need to.

Fixes https://github.com/alphagov/govuk-frontend/issues/4242. More details in the issue.

## Notes
At time of writing, building `main` locally produces minified CSS of 123kb and a CSS map of 408kb. Building this branch locally produces minified CSS of 117kb and a CSS map of 397kb.

The only 2 components that still have more than one use of `govuk-font` are the header and input component. Both of these are because of user agent styles that override the inherited font size, weight and font family. Specifically the `button` element that `govuk-header__menu-button` is using and the `input` for `govuk-input`.

This change makes the following classes redundant:

- `govuk-panel__body`
- `govuk-table__caption--s`

Should we deprecate these classes?

This change additionally adds our default text colour to `govuk-summary-card` which was missed when we initially launched this.

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/checkboxes/_index.scss
+++ b/packages/govuk-frontend/src/govuk/components/checkboxes/_index.scss
@@ -10,8 +10,6 @@
   $govuk-checkboxes-label-padding-left-right: govuk-spacing(3);
 
   .govuk-checkboxes__item {
-    @include govuk-font($size: 19);
-
     display: block;
     position: relative;
 
@@ -141,9 +139,9 @@
   // =========================================================
 
   .govuk-checkboxes__divider {
-    $govuk-divider-size: $govuk-checkboxes-size !default;
     @include govuk-font($size: 19);
     @include govuk-text-colour;
+    $govuk-divider-size: $govuk-checkboxes-size !default;
     width: $govuk-divider-size;
     margin-bottom: govuk-spacing(2);
     text-align: center;
--- a/packages/govuk-frontend/src/govuk/components/error-summary/_index.scss
+++ b/packages/govuk-frontend/src/govuk/components/error-summary/_index.scss
@@ -2,6 +2,7 @@
 
 @include govuk-exports("govuk/component/error-summary") {
   .govuk-error-summary {
+    @include govuk-font($size: 19);
     @include govuk-text-colour;
     @include govuk-responsive-padding(4);
     @include govuk-responsive-margin(8, "bottom");
@@ -14,15 +15,14 @@
   }
 
   .govuk-error-summary__title {
-    @include govuk-font($size: 24, $weight: bold);
+    @include govuk-typography-responsive($size: 24);
+    @include govuk-typography-weight-bold;
 
     margin-top: 0;
     @include govuk-responsive-margin(4, "bottom");
   }
 
   .govuk-error-summary__body {
-    @include govuk-font($size: 19);
-
     p {
       margin-top: 0;
       @include govuk-responsive-margin(4, "bottom");
--- a/packages/govuk-frontend/src/govuk/components/fieldset/_index.scss
+++ b/packages/govuk-frontend/src/govuk/components/fieldset/_index.scss
@@ -35,24 +35,27 @@
   }
 
   // Modifiers that make legends look more like their equivalent headings
+  .govuk-fieldset__legend--xl,
+  .govuk-fieldset__legend--l,
+  .govuk-fieldset__legend--m {
+    @include govuk-typography-weight-bold;
+    margin-bottom: govuk-spacing(3);
+  }
 
   .govuk-fieldset__legend--xl {
-    @include govuk-font($size: 48, $weight: bold);
-    margin-bottom: govuk-spacing(3);
+    @include govuk-typography-responsive($size: 48);
   }
 
   .govuk-fieldset__legend--l {
-    @include govuk-font($size: 36, $weight: bold);
-    margin-bottom: govuk-spacing(3);
+    @include govuk-typography-responsive($size: 36);
   }
 
   .govuk-fieldset__legend--m {
-    @include govuk-font($size: 24, $weight: bold);
-    margin-bottom: govuk-spacing(3);
+    @include govuk-typography-responsive($size: 24);
   }
 
   .govuk-fieldset__legend--s {
-    @include govuk-font($size: 19, $weight: bold);
+    @include govuk-typography-weight-bold;
   }
 
   // When the legend contains an H1, we want the H1 to inherit all styles from
--- a/packages/govuk-frontend/src/govuk/components/label/_index.scss
+++ b/packages/govuk-frontend/src/govuk/components/label/_index.scss
@@ -9,24 +9,27 @@
   }
 
   // Modifiers that make labels look more like their equivalent headings
+  .govuk-label--xl,
+  .govuk-label--l,
+  .govuk-label--m {
+    @include govuk-typography-weight-bold;
+    margin-bottom: govuk-spacing(3);
+  }
 
   .govuk-label--xl {
-    @include govuk-font($size: 48, $weight: bold);
-    margin-bottom: govuk-spacing(3);
+    @include govuk-typography-responsive($size: 48);
   }
 
   .govuk-label--l {
-    @include govuk-font($size: 36, $weight: bold);
-    margin-bottom: govuk-spacing(3);
+    @include govuk-typography-responsive($size: 36);
   }
 
   .govuk-label--m {
-    @include govuk-font($size: 24, $weight: bold);
-    margin-bottom: govuk-spacing(2);
+    @include govuk-typography-responsive($size: 24);
   }
 
   .govuk-label--s {
-    @include govuk-font($size: 19, $weight: bold);
+    @include govuk-typography-weight-bold;
   }
 
   // When the label is nested inside a heading, override the heading so that it
--- a/packages/govuk-frontend/src/govuk/components/summary-list/_index.scss
+++ b/packages/govuk-frontend/src/govuk/components/summary-list/_index.scss
@@ -201,7 +201,8 @@
   }
 
   .govuk-summary-card__actions {
-    @include govuk-font($size: 19, $weight: bold);
+    @include govuk-typography-responsive($size: 19);
+    @include govuk-typography-weight-bold;
     display: flex;
     flex-wrap: wrap;
     row-gap: 10px;
--- a/packages/govuk-frontend/src/govuk/components/summary-list/_index.scss
+++ b/packages/govuk-frontend/src/govuk/components/summary-list/_index.scss
@@ -193,6 +193,7 @@
 
   .govuk-summary-card__title {
     @include govuk-font($size: 19, $weight: bold);
+    @include govuk-text-colour;
     margin: govuk-spacing(1) govuk-spacing(4) govuk-spacing(2) 0;
 
     @include govuk-media-query($from: "tablet") {
--- a/packages/govuk-frontend/src/govuk/components/table/_index.scss
+++ b/packages/govuk-frontend/src/govuk/components/table/_index.scss
@@ -43,23 +43,21 @@
   }
 
   // Modifiers that make captions look more like their equivalent headings
+  .govuk-table__caption--xl,
+  .govuk-table__caption--l,
+  .govuk-table__caption--m {
+    margin-bottom: govuk-spacing(3);
+  }
 
   .govuk-table__caption--xl {
-    @include govuk-font($size: 48, $weight: bold);
-    margin-bottom: govuk-spacing(3);
+    @include govuk-typography-responsive($size: 48);
   }
 
   .govuk-table__caption--l {
-    @include govuk-font($size: 36, $weight: bold);
-    margin-bottom: govuk-spacing(3);
+    @include govuk-typography-responsive($size: 36);
   }
 
   .govuk-table__caption--m {
-    @include govuk-font($size: 24, $weight: bold);
-    margin-bottom: govuk-spacing(3);
-  }
-
-  .govuk-table__caption--s {
-    @include govuk-font($size: 19, $weight: bold);
+    @include govuk-typography-responsive($size: 24);
   }
 }
--- a/packages/govuk-frontend/src/govuk/components/tabs/_index.scss
+++ b/packages/govuk-frontend/src/govuk/components/tabs/_index.scss
@@ -2,10 +2,14 @@
   .govuk-tabs {
     @include govuk-responsive-margin(1, "top");
     @include govuk-responsive-margin(6, "bottom");
+    @include govuk-font($size: 19);
   }
 
   .govuk-tabs__title {
-    @include govuk-font($size: 19);
+    // Set the size and weight again because this element is a heading and the
+    // user agent font size overrides the inherited font size
+    @include govuk-typography-responsive($size: 19);
+    @include govuk-typography-weight-regular;
     @include govuk-text-colour;
     margin-bottom: govuk-spacing(2);
   }
@@ -18,7 +22,6 @@
   }
 
   .govuk-tabs__list-item {
-    @include govuk-font($size: 19);
     margin-left: govuk-spacing(5);
 
     &::before {
--- a/packages/govuk-frontend/src/govuk/components/warning-text/_index.scss
+++ b/packages/govuk-frontend/src/govuk/components/warning-text/_index.scss
@@ -1,13 +1,15 @@
 @include govuk-exports("govuk/component/warning-text") {
   .govuk-warning-text {
-    position: relative;
+    @include govuk-font($size: 19);
     @include govuk-responsive-margin(6, "bottom");
+    position: relative;
     padding: govuk-spacing(2) 0;
   }
 
   .govuk-warning-text__icon {
-    @include govuk-font($size: false, $weight: bold);
-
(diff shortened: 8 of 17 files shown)


packages/govuk-frontend/src/govuk/components/checkboxes/_index.scss
```diff
@@ -9,9 +9,11 @@
   $govuk-small-checkboxes-size: 24px;
   $govuk-checkboxes-label-padding-left-right: govuk-spacing(3);
 
-  .govuk-checkboxes__item {
+  .govuk-checkboxes {
     @include govuk-font($size: 19);
+  }
```
colinrotherham, line 14: Hear me out @owenatgov, I'm being controversial

We already set `font-size` on the Checkbox label and hint, so can we ditch this? 😱

I guess it's only needed because we use a `<div>` for the divider text?

```html
<div class="govuk-checkboxes__divider">or</div>
```

```suggestion
```

packages/govuk-frontend/src/govuk/components/error-summary/_index.scss
```diff
@@ -2,6 +2,7 @@
 
 @include govuk-exports("govuk/component/error-summary") {
   .govuk-error-summary {
+    @include govuk-font($size: 19);
```
colinrotherham, line 5: Same with this one, could it go??

We only need it because we haven't put a heading class on the error summary title?

```html
<h2 class="govuk-error-summary__title"></h2>
```

```suggestion
```

packages/govuk-frontend/src/govuk/components/checkboxes/_index.scss
owenatgov: Good spot! This also applies to radios. I tried removing the `govuk-text-colour` as well and `govuk-font` doesn't apply this 😬 I'm gonna do a second pass to double check none of these have lost our off-black.

packages/govuk-frontend/src/govuk/components/error-summary/_index.scss
owenatgov: This one's tricky. We'd need to apply this to the template, which would be a suggested change which would require a bit of comms. For the sake of getting this PR out as a fix, would you be willing to leave this for now, I make an issue, in 5.X we make this a suggested change and scoop it out wholesale in 6.0?

packages/govuk-frontend/src/govuk/components/checkboxes/_index.scss
owenatgov: Only impact was on the summary card, which I'm pretty sure has never had the proper colour 😱 Fixed as part of this.

packages/govuk-frontend/src/govuk/components/error-summary/_index.scss
colinrotherham: Sounds perfect

E.g. Search for `__title` in `*.njk`

See all the templates where a heading class could make more `govuk-font()` lines unnecessary:

```
.govuk-error-summary__title
.govuk-notification-banner__title
.govuk-panel__title
.govuk-summary-card__title
.govuk-tabs__title
```

Search again with `__heading` and you get the idea 🎉

Definitely one for another issue

owenatgov: Making this a wider heading clearup issue is a cracking idea. I'll write up an issue in a bit.

packages/govuk-frontend/src/govuk/components/fieldset/_index.scss
```diff
@@ -35,24 +35,27 @@
   }
 
   // Modifiers that make legends look more like their equivalent headings
+  .govuk-fieldset__legend--xl,
+  .govuk-fieldset__legend--l,
+  .govuk-fieldset__legend--m {
+    @include govuk-typography-weight-bold;
+    margin-bottom: govuk-spacing(3);
+  }
```
colinrotherham, line 43: 🎉

packages/govuk-frontend/src/govuk/components/error-summary/_index.scss
```diff
@@ -14,15 +15,14 @@
   }
 
   .govuk-error-summary__title {
-    @include govuk-font($size: 24, $weight: bold);
+    @include govuk-typography-responsive($size: 24);
+    @include govuk-typography-weight-bold;
```
colinrotherham, line 19: Can we drop all these `font-weight: 700` lines for every heading?

Unless the browser default `font-weight: bold` (same as 700) has issues?

```suggestion
```

packages/govuk-frontend/src/govuk/components/summary-list/_index.scss
```diff
@@ -172,6 +172,8 @@
 
   // Additional block for the summary card
   .govuk-summary-card {
+    @include govuk-font($size: 19, $weight: bold);
```
colinrotherham, line 175: Can we drop this too?

We style `.govuk-summary-card__title` and `.govuk-summary-list` etc separately

```suggestion
```

packages/govuk-frontend/src/govuk/components/table/_index.scss
```diff
@@ -43,23 +43,21 @@
   }
 
   // Modifiers that make captions look more like their equivalent headings
+  .govuk-table__caption--xl,
+  .govuk-table__caption--l,
+  .govuk-table__caption--m {
+    margin-bottom: govuk-spacing(3);
+  }
```
colinrotherham, line 50: Thanks for combining all these @owenatgov

Out of interest, how come we don't use the heading classes?

We should fix this example in the YAML where we've used one by accident:

https://github.com/alphagov/govuk-frontend/blob/16d0085e072c995df5107a6055a1328a57631b99/packages/govuk-frontend/src/govuk/components/table/table.yaml#L135

packages/govuk-frontend/src/govuk/components/tabs/_index.scss
```diff
@@ -2,10 +2,14 @@
   .govuk-tabs {
     @include govuk-responsive-margin(1, "top");
     @include govuk-responsive-margin(6, "bottom");
+    @include govuk-font($size: 19);
```
colinrotherham, line 5: Our tab links `.govuk-tabs__tab` already use `@mixin govuk-link-common` so might not need this?

Can override the `font-size` only on `.govuk-tabs__tab` where we bump it up?

```suggestion
```

packages/govuk-frontend/src/govuk/components/warning-text/_index.scss
```diff
@@ -1,13 +1,12 @@
 @include govuk-exports("govuk/component/warning-text") {
   .govuk-warning-text {
-    position: relative;
+    @include govuk-font($size: 19, $weight: bold);
```
colinrotherham, line 3: Can we rely on `<strong>` in the template instead?

```suggestion
    @include govuk-font($size: 19);
```

packages/govuk-frontend/src/govuk/components/error-summary/_index.scss
owenatgov: The weight mixins are tied to the variables `$govuk-font-weight-bold` and `$govuk-font-weight-regular`. I'd rather maintain that relationship.

There's a question this raises for me on the value of those mixins but that's a not today problem.

packages/govuk-frontend/src/govuk/components/summary-list/_index.scss
owenatgov: To clarify, are you suggesting moving the `govuk-font` declaration to the `__title` element?

packages/govuk-frontend/src/govuk/components/table/_index.scss
owenatgov: I'll fix that example but the question itself feels like one for https://github.com/alphagov/govuk-frontend/issues/4277

packages/govuk-frontend/src/govuk/components/summary-list/_index.scss
colinrotherham: Yeah, keeping it on `__title` elements (no `font-family`) and remove it everywhere else

Then when https://github.com/alphagov/govuk-frontend/issues/4277 is picked up we'd naturally remove it from `__title` elements too?

Otherwise I'm totally happy moving it to the component block for safety but it's not really necessary

packages/govuk-frontend/src/govuk/components/table/_index.scss
colinrotherham: Brill, thank you

packages/govuk-frontend/src/govuk/components/summary-list/_index.scss
owenatgov: Nope, makes sense!

packages/govuk-frontend/src/govuk/components/tabs/_index.scss
owenatgov: I'm going to keep this if that's ok because without it, the heading and the list dashes on mobile don't get typography:

| With `govuk-font` | Without `govuk-font`|
| - | - |
| Screenshot of tabs on mobile with typography |  Screenshot of tabs on mobile without typography |

Once again, we've got an example where https://github.com/alphagov/govuk-frontend/issues/4277 could solve at least some of this.

packages/govuk-frontend/src/govuk/components/warning-text/_index.scss
owenatgov: This would mean it didn't apply to the "!" because the icon itself is secretly typography, but we can just use a font weight mixin for that. Will fix.

packages/govuk-frontend/src/govuk/components/tabs/_index.scss
colinrotherham: Sorry I should have joined the dots here

Same as https://github.com/alphagov/govuk-frontend/pull/4267#discussion_r1341203283 can we style only the elements without `font-family`?

* Add it to `.govuk-tabs__title` until we extend it with `.govuk-heading-s`
* Add it to `.govuk-tabs__list` until we extend it with `.govuk-list`

Happy to move to the component block though and document these instead?

packages/govuk-frontend/src/govuk/components/warning-text/_index.scss
colinrotherham: Appreciate we're squeezing the life out the CSS file size, so don't have to go this far 🙈

packages/govuk-frontend/src/govuk/components/tabs/_index.scss
owenatgov: I'm trying to remember my sass compilation arithmetic here. Would moving it out of the top level mean we end up with more less generated CSS? In this case it's also stopover until we get to #4277 and also we're probably talking individual kilobytes so I'm happy to do this for now if you reckon it's better.

colinrotherham: Tell you what, with the amount of code that `govuk-font()` adds let's keep it at block level for now

Better to remove it entirely (later) when suggestions 1) and 2) can be picked up

owenatgov: Thank you for compromising 😂

packages/govuk-frontend/src/govuk/components/table/_index.scss
owenatgov: I just spotted the following in our docs for `captionClass` on tables:

>Classes for caption text size. Classes should correspond to the available typography heading classes.

Oh dear. More mounting evidence for https://github.com/alphagov/govuk-frontend/issues/4277 I'm not gonna let that stop me removing the heading class but I'm instead going to replace it with one of the caption modifiers for consistency _for now_.

colinrotherham: Ahh so the YAML docs said ["typography heading classes" since 2018](https://github.com/alphagov/govuk-frontend/pull/998) but must have forgotten to update that line when [caption classes came in 2021](https://github.com/alphagov/govuk-design-system/pull/1478). Will leave it with you

packages/govuk-frontend/src/govuk/components/label/_index.scss
```diff
@@ -9,24 +9,27 @@
...
 
   .govuk-label--m {
-    @include govuk-font($size: 24, $weight: bold);
-    margin-bottom: govuk-spacing(2);
+    @include govuk-typography-responsive($size: 24);
   }
 
   .govuk-label--s {
-    @include govuk-font($size: 19, $weight: bold);
+    @include govuk-typography-weight-bold;
```
colinrotherham, line 32: Should the whole rule be removed and put `govuk-typography-weight-bold` in `.govuk-label`?

Means you can remove it further up too

```patch
  .govuk-label--xl,
  .govuk-label--l,
  .govuk-label--m {
-   @include govuk-typography-weight-bold;
    margin-bottom: govuk-spacing(3);
  }
```  

```suggestion
```

packages/govuk-frontend/src/govuk/components/fieldset/_index.scss
```diff
@@ -35,24 +35,27 @@
   }
 
   // Modifiers that make legends look more like their equivalent headings
+  .govuk-fieldset__legend--xl,
+  .govuk-fieldset__legend--l,
+  .govuk-fieldset__legend--m {
+    @include govuk-typography-weight-bold;
```
colinrotherham, line 41: Can we save some more code and stick `$weight: bold` in `.govuk-fieldset__legend` instead?

packages/govuk-frontend/src/govuk/components/fieldset/_index.scss
```diff
@@ -35,24 +35,27 @@
...
 
   .govuk-fieldset__legend--m {
-    @include govuk-font($size: 24, $weight: bold);
-    margin-bottom: govuk-spacing(3);
+    @include govuk-typography-responsive($size: 24);
   }
 
   .govuk-fieldset__legend--s {
-    @include govuk-font($size: 19, $weight: bold);
+    @include govuk-typography-weight-bold;
```
colinrotherham, line 58: Can drop this line if we do https://github.com/alphagov/govuk-frontend/pull/4267/files#r1341511854

```suggestion
```

colinrotherham:
Think I'm happy with my final review 🙌

Shall we update the description to add `@include govuk-text-colour;` has been added where missed too?

Or split that out into another issue? I've not checked if we've caught them all

owenatgov:
@colinrotherham I've added it to the bottom of the description of this PR. I'm fairly confident we're using `govuk-text-colour` across our library so I don't think it needs a new issue right now.

packages/govuk-frontend/src/govuk/components/label/_index.scss
owenatgov: Default labels aren't bold so I think this is the best we can do right now

packages/govuk-frontend/src/govuk/components/fieldset/_index.scss
owenatgov: Same thing as https://github.com/alphagov/govuk-frontend/pull/4267#discussion_r1341530600

packages/govuk-frontend/src/govuk/components/fieldset/_index.scss
owenatgov: Same again https://github.com/alphagov/govuk-frontend/pull/4267#discussion_r1341530600

packages/govuk-frontend/src/govuk/components/label/_index.scss
colinrotherham: Ahh you learn something new every day (just found [Labels legends and headings](https://govuk-frontend-review.herokuapp.com/examples/labels-legends-and-headings))

Thanks for checking

packages/govuk-frontend/src/govuk/components/error-summary/_index.scss
36degrees: IIRC, there are a few things we've explored in the past that, if we ever did them, would benefit from using the mixins:

- only enabling font smoothing for bold text
- using `font-size-adjust`, which has [different values for bold vs normal text](https://github.com/alphagov/govuk_frontend_toolkit/pull/271)

It'd also allow us to e.g. introduce a mid weight if we wanted to.

Arguably all of these are premature optimisations but I don't think there's enough justification to remove them now.

packages/govuk-frontend/src/govuk/components/table/_index.scss
36degrees: > Out of interest, how come we don't use the heading classes?

We've generally tried to avoid calling things headings if they aren't semantically headings.

colinrotherham: Thanks @36degrees

Squeeze in a `captionClass` YAML docs fix if you like @owenatgov?

owenatgov:
@colinrotherham Interestingly, the minified CSS hasn't changed but the map has increased by one kb 😮 I wonder how much of this is impacted by parallel work and rebasing. Not an analysis worth stressing over right now. Merging!

colinrotherham:
Thanks @owenatgov

We've added more Sass than we've removed I guess? All those source changes needs mapping

They're are only downloaded when developer tools are open so it's fine by me

Issue #4179, opened 2023-09-05
Implement the new crown assets in Frontend (v4)


querkmachine:
## What

Implement the new crown assets in new releases of GOV.UK Frontend v4.x. 

v4 will support both the original (St. Edward's) and new (Tudor) crown graphics. v4 will continue to use the St Edward's crown by default to avoid unexpected breaking changes, with teams having to explicitly opt-in to using the Tudor crown graphics.

## Why

To ensure that teams can easily impement the new GOV.UK branding into existing services.

## Who needs to work on this

- Developers

## Who needs to review this

- Developers
- Designers
- Accessibility specialist

## Done when

```[tasklist]
### Tasks
- [x] Inline SVG crown has been implemented in Header component
- [x] Fallback PNG crown has been implemented in Header component
- [x] Nunjucks and Sass feature flags to enable Tudor crown graphics have been implemented
- [x] Tudor crown favicon and web sharing graphics have been updated
```

stevenjmesser:
Merging and releasing this is dependent on https://github.com/alphagov/govuk-frontend/issues/4059.

PR #4278, opened 2023-09-29
Implement the Tudor crown in the Header component (v4.x)

querkmachine:
Implement the Tudor crown iconography in the Header component. Part of #4179.

See #4297 for the equivalent changes for Frontend v5.

> [!WARNING]
> This PR should not be merged until we have approval to do so. 

## Changes
- Adds the Tudor crown SVG to the Header component.
  - Both crowns (and their fallbacks) have been moved out into their own private variables to try and keep the bulk of the HTML easy to read.
- Adds a new Tudor crown fallback graphic for IE8.
- Adds a new Nunjucks parameter to the Header component, `useTudorCrown`, to switch to using the new Tudor crown.
- Adds some styles adjusting the positioning of the Tudor crown, if it's in use. 
- Adds tests ensuring that the Tudor crown is being rendered if the flag is set, and repeats the same tests carried out for the St. Edward's crown.

## Todo
- When complete, this PR can (probably) be wholesale copied for v3.x with little to no alteration.


owenatgov (approved):
Looks good to me! I ummed and ahhed about the `width="32"` targeting over using a modifier class but it's probably the simplest way to do it considering we'd ask users to change that value anyway, and IMO the impact of a user missing setting width to 32 if they're not using our macros is pretty low.

We should be explicit in our release notes that users not using our macro should set the width when updating the svg, or ideally just re-copy a new svg.

querkmachine:
Merging in preparation for release w/c 15th January.

Issue #4243, opened 2023-09-21
Rename `govuk-typography-responsive`


owenatgov:
## What
Rename the mixin [`govuk-typoraphy-responsive`](https://frontend.design-system.service.gov.uk/sass-api-reference/#govuk-typography-responsive) to something that more clearly dictates what it does.

## Why
It emerged during https://github.com/alphagov/govuk-frontend/issues/3837 that users may not understand how to use `govuk-font` is it is intended. Part of this is driven by the fact that users don't know about `govuk-typography-responsive` as the mixin that handles the retrieval of `font-size` and `line-height` from the typography scale. Part of this can be handled by documenting it better however we have a reckon that a more appropriate name would help users understand it at a glance and recall it easier.

A name that emerged during the linked spike is `govuk-font-size` however this isn't set in stone.

## Who needs to work on this
Developers

## Who needs to review this
Developers

## Done when

- [x] A new name is decided
- [x] The new name is applied
- [x] `govuk-typography-responsive` is deprecated

36degrees:
It looks like it's already covered in the done whens (thanks! 🙌🏻 ) but we should follow our [process for renaming Sass mixins](https://github.com/alphagov/govuk-frontend/blob/main/docs/contributing/managing-change.md#renaming-sass-mixins-and-functions) for this.

querkmachine:
`govuk-font-size` feels like a nice, self-explanatory name. Helps that it's aligned with the name of the existing `font-size` property, which is already something we aim to do with the naming of components. 

That does make me wonder if there needs to be some formality introduced as to how all the mixins are named though. Some begin with `govuk-typography` and others use `govuk-font`. The likes of `govuk-typography-weight-regular` would probably be more intuitively named `govuk-font-weight-regular`, etc.

That might be spinning off a whole other bunch of changes, though 😅

owenatgov:
@querkmachine Excellent point. I do wonder if there's an argument to changing all our typography mixins to use a `govuk-font-xyz` prefix. To me, that feels easier to parse than "typography" although there are arguments for both eg: when we use `govuk-typography-common` we aren't just applying a font, we're applying some settings that are central to our _typography_. 

[...]

PR #4291, opened 2023-10-03
Rename `govuk-typography-responsive` to `govuk-font-size`

owenatgov:
Renames `govuk-typography-responsive` to `govuk-font-size`, deprecates `govuk-typography-responsive`, removes internal use of the old name and adds a test to check for the deprecation waring.

Fixes https://github.com/alphagov/govuk-frontend/issues/4243. More details in the issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@ For advice on how to use these release notes see [our guidance on staying up to 
 
 ## Unreleased
 
+### Replace instances of `govuk-typography-responsive` with `govuk-font-size`
+
+We've renamed the Sass mixin `govuk-typography-responsive` to `govuk-font-size` and have deprecated `govuk-typography-responsive`. You can still use `govuk-typography-responsive` but we'll be removing it in a future breaking release (6.0.0).
+
+This is an experimental change, based on our hypothesis that `govuk-font-size` as a name better communicates the Sass mixin's intended use than `govuk-typography-responsive`. We're interested in feedback from the community on this name change so please let us know what you think.
+
+This change was introduced in [pull request #4291: Rename `govuk-typography-responsive` to `govuk-font-size`](https://github.com/alphagov/govuk-frontend/pull/4291)
+
 ### Fixes
 
 We've made fixes to GOV.UK Frontend in the following pull requests:
--- a/packages/govuk-frontend/src/govuk/helpers/_typography.scss
+++ b/packages/govuk-frontend/src/govuk/helpers/_typography.scss
@@ -129,7 +129,7 @@
 ///
 /// @param {Number} $size - Point from the spacing scale (the size as it would
 ///   appear on tablet and above)
-/// @param {Number} $override-line-height [false] - Non responsive custom line
+/// @param {Number} $line-height [false] - Non responsive custom line
 ///   height. Omit to use the line height from the font map.
 /// @param {Boolean} $important [false] - Whether to mark declarations as
 ///   `!important`.
@@ -138,7 +138,7 @@
 ///
 /// @access public
 
-@mixin govuk-font-size($size, $override-line-height: false, $important: false) {
+@mixin govuk-font-size($size, $line-height: false, $important: false) {
   @if not map-has-key($govuk-typography-scale, $size) {
     @error "Unknown font size `#{$size}` - expected a point from the typography scale.";
   }
@@ -149,8 +149,14 @@
     $font-size: map-get($breakpoint-map, "font-size");
     $font-size-rem: govuk-px-to-rem($font-size);
 
-    $line-height: _govuk-line-height(
-      $line-height: if($override-line-height, $override-line-height, map-get($breakpoint-map, "line-height")),
+    // $calculated-line-height is a separate variable from $line-height,
+    // as otherwise the value would get redefined with each loop and
+    // eventually break _govuk-line-height.
+    //
+    // We continue to call the param $line-height to stay consistent with the
+    // naming with govuk-font.
+    $calculated-line-height: _govuk-line-height(
+      $line-height: if($line-height, $line-height, map-get($breakpoint-map, "line-height")),
       $font-size: $font-size
     );
 
@@ -159,20 +165,20 @@
     // are used in calculations
     $font-size: $font-size if($important, !important, null);
     $font-size-rem: $font-size-rem if($important, !important, null);
-    $line-height: $line-height if($important, !important, null);
+    $calculated-line-height: $calculated-line-height if($important, !important, null);
 
     @if not $breakpoint {
       font-size: $font-size-rem;
-      line-height: $line-height;
+      line-height: $calculated-line-height;
     } @else if $breakpoint == "print" {
       @include govuk-media-query($media-type: print) {
         font-size: $font-size;
-        line-height: $line-height;
+        line-height: $calculated-line-height;
       }
     } @else {
       @include govuk-media-query($from: $breakpoint) {
         font-size: $font-size-rem;
-        line-height: $line-height;
+        line-height: $calculated-line-height;
       }
     }
   }
@@ -210,6 +216,6 @@
   }
 
   @if $size {
-    @include govuk-font-size($size, $override-line-height: $line-height);
+    @include govuk-font-size($size, $line-height);
   }
 }
--- a/packages/govuk-frontend/src/govuk/helpers/_typography.scss
+++ b/packages/govuk-frontend/src/govuk/helpers/_typography.scss
@@ -210,6 +210,6 @@
   }
 
   @if $size {
-    @include govuk-typography-responsive($size, $override-line-height: $line-height);
+    @include govuk-font-size($size, $override-line-height: $line-height);
   }
 }
--- a/packages/govuk-frontend/src/govuk/helpers/_typography.scss
+++ b/packages/govuk-frontend/src/govuk/helpers/_typography.scss
@@ -80,7 +80,31 @@
   @return $line-height;
 }
 
-/// Responsive typography helper
+/// Font size and line height helper
+///
+/// @param {Number} $size - Point from the spacing scale (the size as it would
+///   appear on tablet and above)
+/// @param {Number} $override-line-height [false] - Non responsive custom line
+///   height. Omit to use the line height from the font map.
+/// @param {Boolean} $important [false] - Whether to mark declarations as
+///   `!important`.
+///
+/// @throw if `$size` is not a valid point from the spacing scale
+///
+/// @access public
+///
+/// @alias govuk-font-size
+/// @deprecated Use `govuk-font-size` instead
+
+@mixin govuk-typography-responsive($size, $override-line-height: false, $important: false) {
+  @include _warning(
+    "govuk-typography-responsive",
+    "govuk-typography-responsive is deprecated. Use govuk-font-size instead."
+  );
+  @include govuk-font-size($size, $override-line-height, $important);
+}
+
+/// Font size and line height helper
 ///
 /// Takes a point from the responsive 'font map' as an argument (the size as it
 /// would appear on tablet and above), and uses it to create font-size and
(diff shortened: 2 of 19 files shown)


packages/govuk-frontend/src/govuk/helpers/_typography.scss
```diff
@@ -114,7 +138,7 @@
 ///
 /// @access public
 
-@mixin govuk-typography-responsive($size, $override-line-height: false, $important: false) {
+@mixin govuk-font-size($size, $override-line-height: false, $important: false) {
```
colinrotherham, line 141: Fab work in this PR 🙌

Do we align `govuk-font()` and `govuk-font-size()` and use **$line-height** in both?

```suggestion
@mixin govuk-font-size($size, $line-height: false, $important: false) {
```

owenatgov: I think yes. This is going to be released with the initial deprecation alongside changes to our docs, so it's potentially more likely that, if we release these holistically, there will be fewer instances of users changing `govuk-typography-responsive` to `govuk-font-size` and more instances of changing an incorrect use of `govuk-font` to `govuk-font-size`. Aligning these values would make that part of the update a little simpler.

owenatgov: Done. I ran into some interesting problems with doing this. We already use a var called `$line-height` in the mixin and it was being renamed on every loop which led to `_govuk-line-height` breaking:
https://github.com/alphagov/govuk-frontend/blob/f837f9c003a50f9e64a2fe3476be91e6e228f318/packages/govuk-frontend/src/govuk/helpers/_typography.scss#L128-L131
I fixed it by renaming `$line-height` to `$calculated-line-height`. Not my finest work naming wise so I'm open to alternatives.

Additionally I'm not sure if I should also deprecate this in `govuk-typography-responsive`. My instinct says no.

packages/govuk-frontend/src/govuk/helpers/_typography.scss
```diff
@@ -114,19 +139,34 @@
 ///
 /// @access public
 
-@mixin govuk-typography-responsive($size, $override-line-height: false, $important: false) {
+@mixin govuk-font-size($size, $line-height: false, $important: false, $override-line-height: false) {
   @if not map-has-key($govuk-typography-scale, $size) {
     @error "Unknown font size `#{$size}` - expected a point from the typography scale.";
   }
 
+  @if $override-line-height != false {
```
36degrees, line 147: I don't think we need to do this, as long as we make it clear what's changed in the release notes. We can change the mixin name and the name of the params at the same time.

As long as the `govuk-typography-responsive` mixin keeps the same params as before, this won't break anything.

packages/govuk-frontend/src/govuk/helpers/typography.unit.test.js
```diff
@@ -241,13 +241,36 @@ describe('@mixin govuk-typography-responsive', () => {
     )
   })
 
+  // govuk-typography-responsive is the previous name for govuk-font-size
+  it('throws a deprecation warning if govuk-typography-responsive is used', async () => {
```
36degrees, line 245: Can we also test that `govuk-typography-responsive` still does what we expect? i.e. that when all arguments are passed it generates the expected CSS output?

(Could potentially call `govuk-typography-responsive` and `govuk-font` with the same arguments and expect they match?)

36degrees (commented):
Would you mind if we hold off merging this until 5.0 has gone out?

A couple of minor comments but this is looking pretty good.

owenatgov:
@36degrees Before I address review, to be clear I'm not planning on releasing this until after 5.0 is out. It's currently categorised in [the epic](https://github.com/alphagov/govuk-design-system/issues/2289) under the heading "5.X release before the initial release of the new scale". I think I need to find a way to be clearer about how these things are categorised and when we plan to put them out.

packages/govuk-frontend/src/govuk/helpers/typography.unit.test.js
owenatgov: I've reorganised the `govuk-typography-responsive` testing to cover this. I only tested against the basic `govuk-font-size` test as I was finding testing against `govuk-font` tricky to do without a lot of messy string manipulation. This might be my own misunderstanding of our embedded sass strings.

I don't know if this needs further testing against `govuk-font-size` like if `$line-height` or `$important` are set.

36degrees: Like how you've organised that, makes sense 👍🏻

I do think it'd be most valuable to test with all three arguments so we know they're all being passed as expected. For example, if we just forgot to pass the `$important` param to `govuk-font-size` all our tests would still pass at the mo.

From a quick play I think something like this would work:

```
it('outputs the same CSS as govuk-font-size', async () => {
  const sass = `
    ${sassBootstrap}

    .foo {
      @include govuk-typography-responsive(
        $size: 14,
        $override-line-height: 40px,
        $important: true
      )
    }
  `

  const expectedSass = `
    ${sassBootstrap}

    .foo {
      @include govuk-font-size(
        $size: 14,
        $line-height: 40px,
        $important: true
      )
    }
  `

  await expect(compileSassString(sass)).resolves.toMatchObject({
    css: (await compileSassString(expectedSass)).css
  })
})
```

But I'm not sure if that's the most idiomatic way of comparing the results of two promises – perhaps @colinrotherham can advise?

owenatgov: Aha that's a much better idea. I'll have a go at implementing that for the mo.

colinrotherham: Multiple correct answers!

If concurrent compiling isn't a must there's always:

```mjs
const { css: result1 } = await compileSassString('Example 1')
const { css: result2 } = await compileSassString('Example 2')

expect(result1).toEqual(result2)
```

Or start the promises running early but await them later?

Still got to faff around with that nested CSS property

```mjs
const result1 = compileSassString('Example 1')
const result2 = compileSassString('Example 2')

await expect(result1).resolves.toMatchObject({
  css: (await result2).css
})
```

Or chain them in advance so each promise awaits to CSS in readiness?

But Jest has no [`expectAsync()` like Jasmine](https://jasmine.github.io/api/5.1/global.html#expectAsync) so odd to `.resolves` the 1st promise and `await` the 2nd

```mjs
const css1 = compileSassString('Example 1')
  .then(({ css }) => css)

const css2 = compileSassString('Example 2')
  .then(({ css }) => css)

await expect(css1).resolves.toEqual(await css2)
```

Or combine them into a single `await Promise.all()` and extracting the CSS after?

```mjs
const [css1, css2] = await Promise.all([
  compileSassString('Example 1'),
  compileSassString('Example 2')
]).then((results) => results.map(({ css }) => css))

expect(css1).toEqual(css2)
```

Can use as much or little shorthand as you like:

```patch
- ]).then((results) => results.map(({ css }) => css))
+ ]).then((results) => results.map((result) => result.css))
```

Or split out the `.map()` onto another line if it's clearer?

```mjs
const results = await Promise.all([
  compileSassString('Example 1'),
  compileSassString('Example 2')
])

const [result1, result2] = results.map(({ css }) => css)

expect(result1).toEqual(result2)
```

colinrotherham: Can't leave out [`for await...of`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of) but shame we only have one result for comparison:

```mjs
const result = await compileSassString('Example 1')

const results = [
  compileSassString('Example 2'),
  compileSassString('Example 3'),
  compileSassString('Example 4')
]

for await (const { css } of results) {
  expect(css).toEqual(result.css)
}
```

owenatgov: Thanks for this Colin! As you said, lots of correct answers. I'm going to take the lazy way out and say lets figure this out closer to the point at which this is going out. I get the impression as a pre-review we're fairly happy with this in prospect.

colinrotherham:
@owenatgov Mind fixing the conflict and rebasing so we can weigh up the diff changes?

It's going to be good one

colinrotherham (approved):

querkmachine (approved):
Double approval!

CHANGELOG.md
```diff
@@ -4,6 +4,14 @@ For advice on how to use these release notes see [our guidance on staying up to
 
 ## Unreleased
 
+### Replace instances of `govuk-typography-responsive` with `govuk-font-size`
+
+We have renamed the Sass mixin `govuk-typography-responsive` to `govuk-font-size` and have deprecated `govuk-typography-responsive`. You can still use `govuk-typography-responsive` but we will be removing it in the next breaking release (6.0.0).
+
+This is an experimental change based on our hypothesis that `govuk-font-size` as a name better communicates what the Sass mixin's intended use is better than `govuk-typography-responsive`. We're interested in feedback from the community on this change so please let us know what you think of this name change.
+
+This change was introduced in [pull request #4291: Rename `govuk-typography-responsive` to `govuk-font-size`](https://github.com/alphagov/govuk-frontend/pull/4291)
```
calvin-lau-sig7, line 13: Suggestion to slightly simplify the sentences.

```suggestion
### Replace instances of `govuk-typography-responsive` with `govuk-font-size`

We've renamed the Sass mixin `govuk-typography-responsive` to `govuk-font-size` and have deprecated `govuk-typography-responsive`. You can still use `govuk-typography-responsive` but we'll be removing it in a future breaking release (6.0.0).

This is an experimental change, based on our hypothesis that `govuk-font-size` as a name better communicates the Sass mixin's intended use than `govuk-typography-responsive`. We're interested in feedback from the community on this name change so please let us know what you think.

This change was introduced in [pull request #4291: Rename `govuk-typography-responsive` to `govuk-font-size`](https://github.com/alphagov/govuk-frontend/pull/4291)
```

packages/govuk-frontend/src/govuk/helpers/_typography.scss
```diff
@@ -125,8 +149,14 @@
...
     $font-size-rem: govuk-px-to-rem($font-size);
 
-    $line-height: _govuk-line-height(
-      $line-height: if($override-line-height, $override-line-height, map-get($breakpoint-map, "line-height")),
+    // We give this variable a separate name to the $line-height param because
+    // otherwise the variable's value is redefined per loop and eventually
+    // _govuk-line-height breaks. We retain the param being called $line-height
+    // instead of this one because $line-height is consistent with the param
+    // naming for govuk-font, meaning it's easier for users to use them both
+    // with minimal friction.
```
calvin-lau-sig7, line 157: Can we simplify this sentence with something like this:

> This variable is separate from the $line-height param, as otherwise this variable will get redefined with each loop and eventually break  `_govuk-line-height`.
> 
> We continue to call the param $line-height to keep consistent naming with govuk-font, and make both easier to use.

Also, do can we give the name of the variable instead of calling it "this variable"?

calvin-lau-sig7: Another small go at this based on a brief Slack chat with @owenatgov 

> `$calculated-line-height` is a separate variable from the `$line-height` param, as otherwise the value would get redefined with each loop and eventually break  `_govuk-line-height`.
> 
> We continue to call the param $line-height to stay consistent with the naming with govuk-font, and make both easier to use.

owenatgov: I'd like to suggest some light amendments:

>`$calculated-line-height` is a separate variable from `$line-height`, as otherwise the value would get redefined with each loop and eventually break `_govuk-line-height`.
>
>We continue to call the variable `$line-height` to stay consistent with the naming in govuk-font.

Why:

- This carries over from the original comment so it's my bad but it's probably confusing referring to "variable" and "param" in the same sentence when they mean the same thing
- It doesn't feel quite right to me to say that it makes both easier to use. It makes it so that users don't have to remember to use different variable names when swapping between the different mixins. I think just saying it's to keep it consistent does all the heavy lifting of the sentence.

calvin-lau-sig7: Yep looks good to me, much simpler! Thanks for helping me muddle through the terminology.

owenatgov: DONE

calvin-lau-sig7 (approved):

Issue #4178, opened 2023-09-05
Implement the new crown assets in Frontend (v5.x)


querkmachine:
## What

Implement the new crown assets in GOV.UK Frontend v5.x.

## Why

To ensure that teams are using the new GOV.UK branding on any new services.

## Who needs to work on this

- Developers

## Who needs to review this

- Developers
- Designers
- Accessibility specialist

## Done when

```[tasklist]
- [x] Inline SVG graphic has been implemented in Header component
- [x] Sass has been updated to account for the new lockup (link hover styles, product name integration, etc.)
- [x] Favicon and web sharing graphics have been replaced
- [x] A feature flag to activate the new header graphic has been implemented
```

stevenjmesser:
Merging and releasing this is dependent on https://github.com/alphagov/govuk-frontend/issues/4059.

querkmachine:
Shifting this into blocked for the time being as this might become work for a 5.x release instead, with a transitional piece being merged into 5.0 instead.

PR #4297, opened 2023-10-03
 Implement the Tudor crown in the Header component (v5.x)

querkmachine:
Implements the Tudor crown change and new GOV.UK logo lockup in the Header component. Part of #4178. 

The bulk of changes originally in this PR have now been merged in other PRs:
- #4449
- #4497 

See #4278 for the equivalent changes for Frontend v4 and earlier.

> [!WARNING]
> This PR should not be merged until we have approval to do so.

## Changes
- Adds the new GOV.UK logo lockup (with Tudor crown) to the Header component, behind a `useTudorCrown` parameter, similar to the v3 and v4 releases.
- Adds parameter to Nunjucks documentation.
- Adds a new test for the parameter.
- Updates changelog.


36degrees:
Will also close #1739.

colinrotherham (commented):
Looking ace @querkmachine

You know this bit about the 2px offset?

>    * The top offset has been increased to 2px as this appears to better match the text baseline when the logo is paired with a product name.

It's fixed Chrome, but Safari is out by 1px still on `@1x`, `@1.5x` and `@2x` displays

Weirdly for in-betweens like 4K → 3K both Safari and Chrome are misaligned

Product name baseline

It's going to be quite noticeable as a vertical text shift between services:

https://github.com/alphagov/govuk-frontend/assets/415517/fb6148d5-82ea-40a6-a766-e7a5f06273f3

querkmachine:
@colinrotherham I'm not _too_ concerned about misalignment between old crown and new. From the sounds of it there's gonna be a push for a rapid, widescale roll out of the new crown, so the two will ideally only coexist for a short time.

I'm not totally sure what to do about differences between browsers. The current offset seems to work for Chromium, Firefox and IE 11, and is only one pixel off in Safari, so barring implementing some Safari-specific hack I'd be inclined to say it's a 'good enough' fit. 

For the tweener screen resolutions it likely just comes down to how a user agent tries to round subpixels to device pixels. I doubt there's too much we can really do to influence that.

colinrotherham:
@querkmachine I think we're happy with a minor horizontal text jump but the same baseline?

Could you restore the logo ~`top: -2px`~ `top: -1px` but adjust the product name instead?

Regarding Safari, the product name's 24px height and lack of `vertical-align` causes the main difference

Bit of a hammer, but try this and Chrome and Safari are pixel perfect again

```patch
.govuk-header__product-name {
-  line-height: 1;
+  position: relative.
+  height: 30px;
+  top: 4px;
+  line-height: 30px;
+  vertical-align: top;
}
```

owenatgov:
Weighing in on @colinrotherham's comment above: I'll leave it up to @querkmachine, I'm personally agnostic. I don't think the impact of the problem is big enough that it'll negatively impact any users or even be noticeable in a way that disrupts a user or takes them out of a journey.

If you wanted a decision, let's give Colin's suggestion a try. I did some very light testing of it and it seems to do the trick. I'm a little worried about the suggested `line-height` override as this is currently being powered by `govuk-typography-responsive`. I haven't tried passing it a px value as an override so it's worth checking that that's not going to create a mess.

owenatgov:
Also: once that comment is resolved I'm personally happy with this PR 💪🏻 I won't step on Colin's reviewer toes anymore though.

colinrotherham:
@owenatgov @querkmachine I'm a Safari user so my legacy will be seeing the 1px alignment issue forever more 😆

I'm happy to help with a fix?

colinrotherham:
@owenatgov @querkmachine I've had a go in https://github.com/alphagov/govuk-frontend/pull/4423

querkmachine:
Popping this back into draft for now as it will require some changes to work alongside the transitional logo that is likely to go out with 5.0—namely adding a feature flag to switch logo. 

I have been wondering if this one actually _needs_ a feature flag. In the v3/v4 branches, changing the crown SVG also meant changing the fallback PNG and some of the CSS. Here, those surrounding changes will have happened in 5.0, so all 5.x has to do is replace the embedded SVG, and I'm not sure we technically need a flag to pull that off.

packages/govuk-frontend/src/govuk/components/header/template.njk
```diff
@@ -4,38 +4,24 @@
...
-            ></path>
-          </svg>
-          <span class="govuk-header__logotype-text">
-            GOV.UK
-          </span>
-        </span>
+        We use a single compound path for the logo to uneven prevent subpixel
+        rounding shifting elements around relative to one another. #}
```
colinrotherham, line 13: Just an extra word to take out, but do we need an explanation at all?

The original SVG was also a compound path (with no explanation)

```suggestion
        We use a single compound path for the logo to prevent subpixel
        rounding shifting elements around relative to one another. #}
```

colinrotherham (approved):

querkmachine:
Sticking this back in review as I've made a few changes to it. Notably, I somehow blanked on adding the Nunjucks flag functionality mentioned in https://github.com/alphagov/govuk-frontend/pull/4297#issuecomment-1803737483. That is present now. 

I've added a changelog to this PR intending to cover all of the crown changes included in 5.1, not just the ones in this PR. It's structurally similar to the v3/v4 version but with some different bits of code.

colinrotherham:
Thanks @querkmachine

Do we definitely consider it a breaking change to switch to the correct crown?

Would we want `useStEdwardsCrown: true` to "opt out" rather than `useTudorCrown: true` to "opt in"?

querkmachine:
@colinrotherham I'm still somewhat undecided on whether it is, personally, however I'd mentioned the existence of a feature flag in update documentation shared with the cross-team crown change group and consistency with those expectations is probably more valuable. 

Similarly, it allows the update instructions for 5.1, 4.8 and 3.15 to be essentially identical to one another, hopefully simplifying comms around the crown change.

Agree that it should be opt-out going forward. To my mind that would mean having it be opt-in for 5.1 as a 'transition period' and then change it be opt-out from 5.2 onward, deprecating the flag in the process. Does that sound sensible?

CHANGELOG.md
```diff
@@ -6,6 +6,62 @@ For advice on how to use these release notes see [our guidance on staying up to
...
+- favicon.svg
+- govuk-icon-180.png
+- govuk-icon-192.png
+- govuk-icon-512.png
+- govuk-icon-mask.svg
+- govuk-opengraph-image.png
+
+##### Update the logo in the header of your page
+
+If you are using the `govukHeader` Nunjucks macro in your service, add the `useTudorCrown` parameter to the macro instantiation.
```
colinrotherham, line 38: Shall we note the future default change here?

```suggestion
If you are using the `govukHeader` Nunjucks macro in your service, add the `useTudorCrown` parameter to the macro instantiation. This will become the default in a future version of GOV.UK Frontend.
```

colinrotherham (approved):
Makes sense @querkmachine

If we choose not to have a transitional period, the default can be flipped in another PR so it's not a blocker 🙌

PR #4301, opened 2023-10-04
[SPIKE] Configure Babel to add ES shims polyfills

colinrotherham:
This PR adds the Babel polyfill provider ES shims via:
https://www.npmjs.com/package/babel-plugin-polyfill-es-shims

This is part of a two spike series:

* https://github.com/alphagov/govuk-frontend/pull/4301
* https://github.com/alphagov/govuk-frontend/pull/4557

## Findings

Lots of polyfills are missing in ES shims when compared to core-js in https://github.com/alphagov/govuk-frontend/pull/4557

We also have some unwanted bug fixes:

* ES shims adds `Array.prototype.includes` polyfill [to avoid bugs](https://bugzilla.mozilla.org/show_bug.cgi?id=1767541)
* ES shims adds `Error.prototype.cause` polyfill [to avoid bugs](https://bugs.chromium.org/p/v8/issues/detail?id=12006)

### Remaining issues

* Browserslist matches Opera Mobile 73 under `es6-module` but [Babel disagrees](https://github.com/alphagov/govuk-frontend/pull/4301/commits/41468a62818d6c5d57e219a149690b9bdb62a5d8)
* Rollup plugin [`@rollup/plugin-commonjs`](https://www.npmjs.com/package/@rollup/plugin-commonjs) adds ES module wrappers to polyfills
* Rollup plugin [`@rollup/plugin-babel`](https://www.npmjs.com/package/@rollup/plugin-babel) adds helpers to avoid ES module wrappers
* Rollup outputs `dist/govuk/node_modules` not `dist/govuk/vendor`

---

### Automatic polyfills

When Babel was added we chose "transforms only" without polyfills (see [decision record](https://github.com/alphagov/govuk-design-system-architecture/blob/main/decision-records/006-javascript-compatibility.md))

* https://github.com/alphagov/govuk-frontend/pull/3820

This PR ensures early investigations aren't lost for:

* https://github.com/alphagov/govuk-frontend/issues/3082
* https://github.com/alphagov/govuk-frontend/issues/3287
* https://github.com/alphagov/govuk-frontend/issues/3948
* https://github.com/alphagov/govuk-frontend/issues/4287

### Decision record implications

Since Babel with "transforms only" silently skips polyfills, we noted that [care must be taken](https://github.com/alphagov/govuk-design-system-architecture/blob/main/decision-records/006-javascript-compatibility.md#implications):

[...]

PR #4306, opened 2023-10-05
Log formatted warnings from Rollup

colinrotherham:
This PR replaces our custom [`onwarn` handler](https://rollupjs.org/configuration-options/#onwarn) with the [Rollup config `warnings.flush()`](https://rollupjs.org/javascript-api/#programmatically-loading-a-config-file)

We still break the build on logged warnings, but now includes:

1. Warnings are formatted by Rollup (exclamation marks, yellow text)
2. Warnings are included for config plugins like Babel, Terser

Probably a great way to test Rollup v4 in https://github.com/alphagov/govuk-frontend/pull/4304 and split out from https://github.com/alphagov/govuk-frontend/pull/4301

### DIFF
--- a/shared/tasks/scripts.mjs
+++ b/shared/tasks/scripts.mjs
@@ -43,14 +43,15 @@ export async function compileJavaScript([
 
   // Create Rollup bundle(s)
   for (const options of config.options) {
-    const bundle = await rollup({
-      ...options,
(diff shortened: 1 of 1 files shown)


romaricpascal (approved):
Looks like a nice little improvement! 🥳

PR #4307, opened 2023-10-05
Refactor tabular number activation into their own mixin

### LINKED ISSUE #3778: Split tabular numbers functionality out from the `govuk-font` mixin
## What

Introduce a new mixin that can be used to turn on typographic numbers without needing to call the `govuk-font` mixin.

Update any components that currently call `govuk-font` with `$tabular: true` to use the new mixin.

Update the `govuk-font` mixin to use the new mixin whilst maintaining the existing behaviour.

Optionally, we could also decide to deprecate (and in the future remove) the `$tabular` param on the `govuk-font` mixin and tell people to use the new mixin instead.

## Why

The current implementation is an artefact of the way tabular numbers used to work when compatibility mode was enabled, when we had to switch out the font entirely rather than just switching some font variant settings:

https://github.com/alphagov/govuk-frontend/blob/3f137650cc89923882a6f1af9ee153619516df0f/packages/govuk-frontend/src/govuk/helpers/_typography.scss#L189

Because `govuk-font` includes `govuk-typography-common`, our current approach means that switching on tabular numbers also brings in all of the common typography declarations including the font-face, even when they are not required.

For example, the CSS for the `govuk-table__cell--numeric` modifier class currently looks like this:

```css
.govuk-table__cell--numeric {
    -webkit-font-smoothing: antialiased;
    -moz-osx-font-smoothing: grayscale;
    -webkit-font-feature-settings: "tnum" 1;
    font-feature-settings: "tnum" 1;
    font-family: GDS Transport,arial,sans-serif;
    font-weight: 400
}

@media print {
    .govuk-table__cell--numeric {
        font-family: sans-serif
    }
}

@supports (font-variant-numeric: tabular-nums) {
    .govuk-table__cell--numeric {
        -webkit-font-feature-settings:normal;
        font-feature-settings: normal;
        font-variant-numeric: tabular-nums
    }
}
```

When it could just look like this:

```css
.govuk-table__cell--numeric {
    -webkit-font-feature-settings: "tnum" 1;
    font-feature-settings: "tnum" 1;
}

@supports (font-variant-numeric: tabular-nums) {
    .govuk-table__cell--numeric {
        -webkit-font-feature-settings:normal;
        font-feature-settings: normal;
        font-variant-numeric: tabular-nums
    }
}
```

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [x] We've introduced a new mixin that can be used to turn on typographic numbers
- [x] We've updated any components that currently call `govuk-font` with `$tabular: true` to use the new mixin
- [x] We've updated the `govuk-font` mixin to use the new mixin whilst maintaining the existing behaviour
- [x] We've decided whether to deprecate the `$tabular` param on the `govuk-font` mixin

querkmachine:
Splits the code that activates tabular numbers out of the `govuk-font` mixin into its own standalone mixin. Resolves #3778.

> [!WARNING]
> This PR is for future v5.x work. It should not be merged until after v5.0 has been released. 

## Changes
- Creates a new mixin, `govuk-font-tabular-numbers`, to contain the existing CSS rules.
  - This name of the mixin is intended to be 'self explanatory', insomuch as it closely matches the CSS properties and values it's setting, similar to my suggestion for the `font-weight` mixins [in this comment](https://github.com/alphagov/govuk-frontend/issues/4243#issuecomment-1740680964). 
- Simplifies the CSS rules for enabling tabular numbers.
  - Removes the use of `font-feature-settings` to define tabular numbers. This was only still needed for IE 11 support, and as it's only a stylistic flag, IE having proportional figures seems an acceptable fallback.
  - I'm not sure exactly why we were setting `font-feature-settings: normal` in the `@supports` query. This will override any customisations that a user may have made (see the thoughts below), and resetting the value doesn't do anything for our use case here. I think we can remove both it and the `@supports` query without causing any issues. 
- Changes `govuk-font` to call the new mixin instead.
- Updates all instances where `govuk-font` was being called solely to activate tabular numbers to call the new mixin instead.

## Thoughts
- I've opted to keep and not deprecate `govuk-font`'s `$tabular` parameter. I think it's Mostly Harmless and am inclined to leave it be unless a compelling reason to deprecate it is presented.
- I've made this new mixin accept a parameter that sets `!important` to be consistent with the others and to potentially facilitiate a future utility or override class, which could help implement https://github.com/alphagov/govuk-design-system/issues/1233.
- I'm not a fan of`font-feature-settings`. It's a very 'broad' property that is prone to overreach and the way we were using it would override or will be overridden by any other use of `font-feature-settings` that a user may set. As we now only had `font-feature-settings` for IE 11 support, it's not functionally critical in any way, and as we don't make any guarantee that IE will render identically, I've removed the property entirely.

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/character-count/_index.scss
+++ b/packages/govuk-frontend/src/govuk/components/character-count/_index.scss
@@ -14,7 +14,7 @@
   }
 
   .govuk-character-count__message {
-    @include govuk-font($size: false, $tabular: true);
+    @include govuk-font-tabular-numbers;
     margin-top: 0;
     margin-bottom: 0;
 
--- a/packages/govuk-frontend/src/govuk/helpers/_typography.scss
+++ b/packages/govuk-frontend/src/govuk/helpers/_typography.scss
@@ -62,6 +62,23 @@
   font-weight: $govuk-font-weight-bold if($important, !important, null);
 }
 
+/// Tabular number helper
+///
+/// Switches numerical glyphs (0–9) to use alternative forms with a
+/// monospaced bounding box. This ensures that columns of numbers, such
+/// as those in tables, remain horizontally aligned with one another.
+/// This also has the useful side effect of making numbers more legible
+/// in some situations, such as reference codes, as the numbers are more
+/// distinct and visually separated from one another.
+///
+/// @param {Boolean} $important [false] - Whether to mark declarations as
+///   `!important`. Generally Used to create override classes.
+/// @access public
+
+@mixin govuk-font-tabular-numbers($important: false) {
+  font-variant-numeric: tabular-nums if($important, !important, null);
+}
+
 /// Convert line-heights specified in pixels into a relative value, unless
 /// they are already unit-less (and thus already treated as relative values)
 /// or the units do not match the units used for the font size.
@@ -201,12 +218,7 @@
   @include govuk-typography-common;
 
   @if $tabular {
-    font-feature-settings: "tnum" 1;
-
-    @supports (font-variant-numeric: tabular-nums) {
-      font-feature-settings: normal;
-      font-variant-numeric: tabular-nums;
-    }
+    @include govuk-font-tabular-numbers;
   }
 
   @if $weight == regular {
--- a/packages/govuk-frontend/src/govuk/helpers/typography.unit.test.js
+++ b/packages/govuk-frontend/src/govuk/helpers/typography.unit.test.js
@@ -120,6 +120,42 @@ describe('@mixin govuk-typography-common', () => {
   })
 })
 
+describe('@mixin govuk-font-tabular-numbers', () => {
+  it('enables tabular numbers opentype feature flags', async () => {
+    const sass = `
+      ${sassBootstrap}
+
+      .foo {
+        @include govuk-font-tabular-numbers;
+      }
+    `
+
+    const results = compileSassString(sass)
+
+    await expect(results).resolves.toMatchObject({
+      css: expect.stringContaining('font-variant-numeric: tabular-nums;')
+    })
+  })
+
+  it('marks font-variant-numeric as important if $important is set to true', async () => {
+    const sass = `
+      ${sassBootstrap}
+
+      .foo {
+        @include govuk-font-tabular-numbers($important: true);
+      }
+    `
+
+    const results = compileSassString(sass)
+
+    await expect(results).resolves.toMatchObject({
+      css: expect.stringContaining(
+        'font-variant-numeric: tabular-nums !important;'
+      )
+    })
+  })
+})
+
 describe('@function _govuk-line-height', () => {
   it('preserves line-height if already unitless', async () => {
     const sass = `
@@ -369,18 +405,7 @@ describe('@mixin govuk-font-size', () => {
       const results = compileSassString(sass)
 
       await expect(results).resolves.toMatchObject({
-        css: expect.stringContaining('font-feature-settings: "tnum" 1;')
-      })
-
-      await expect(results).resolves.toMatchObject({
-        css: expect.stringContaining(outdent`
-          @supports (font-variant-numeric: tabular-nums) {
-            .foo {
-              font-feature-settings: normal;
-              font-variant-numeric: tabular-nums;
-            }
-          }
-        `)
+        css: expect.stringContaining('font-variant-numeric: tabular-nums;')
       })
     })
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,21 @@ For advice on how to use these release notes see [our guidance on staying up to 
 
 ## Unreleased
 
-### Replace instances of `govuk-typography-responsive` with `govuk-font-size`
+### New features
+
+#### Use tabular numbers with the `govuk-font-tabular-numbers` mixin
+
+You can now use tabular numbers in your authored Sass by including the new `govuk-font-tabular-numbers` mixin.
+
+Previously, you would use the `govuk-font` mixin with the `$tabular` parameter. However, the `govuk-font` mixin includes styles that are unrelated to tabular numbers, which are unnecessary in some contexts.
+
+These additional styles are not included if you use `govuk-font-tabular-numbers`. Switching to the new mixin can reduce the size of your compiled CSS without affecting the appearance of pages.
+
+This change was introduced in [pull request #4307: Refactor tabular number activation into their own mixin](https://github.com/alphagov/govuk-frontend/pull/4307)
+
+### Recommended changes
+
+#### Replace instances of `govuk-typography-responsive` with `govuk-font-size`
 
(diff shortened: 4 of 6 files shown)


packages/govuk-frontend/src/govuk/helpers/typography.unit.test.js
```diff
@@ -373,14 +373,7 @@ describe('@mixin govuk-typography-responsive', () => {
...
       await expect(results).resolves.toMatchObject({
-        css: expect.stringContaining(outdent`
-          @supports (font-variant-numeric: tabular-nums) {
-            .foo {
-              font-feature-settings: normal;
-              font-variant-numeric: tabular-nums;
-            }
-          }
-        `)
+        css: expect.stringContaining('font-variant-numeric: tabular-nums;')
```
owenatgov, line 376: Thanks for tidying this up. Could we also add some tests just for `govuk-typography-tabular-numbers` that test it plain and with the `$important` param?

owenatgov (commented):
This is looking sharp. Here are some thoughts off of your thoughts:

- I think we can keep `$tabular` on `govuk-font` for instances where someone wants all the typography things at once.
- The naming is a tricky one. I'd like to do some data gathering at the end of the typography work to see what names the community would prefer, however it feels to me like we're starting to move towards `govuk-font-x` rather than `govuk-typography-x`. I'm fine with us calling it something like `govuk-font-variant-tabular`.
- Looking at the original code again, I read it as `font-variant-numeric` being the priority with `font-feature-settings` being the backup, so my instinct is to sack it. I feel even more confident cross referencing our [5.0 browser support grading](https://docs.google.com/document/d/1tEd4UsCmXH55OqD0wJUg5nNQj5pZyrhVzc2TsmppVww/edit#heading=h.hx5s99t7hce) against [support for `font-variant-numeric`](https://caniuse.com/font-variant-numeric).

packages/govuk-frontend/src/govuk/helpers/_typography.scss
```diff
@@ -62,6 +62,22 @@
...
+/// using font-feature-settings as the default and then removing it again if
+/// font-variant-numeric is supported.
+///
+/// @param {Boolean} $important [false] - Whether to mark declarations as
+///   `!important`. Generally Used to create override classes.
+/// @access public
+
+@mixin govuk-typography-tabular-numbers($important: false) {
+  font-feature-settings: "tnum" 1 if($important, !important, null);
+  font-variant-numeric: tabular-nums if($important, !important, null);
```
colinrotherham, line 78: Was there a risk at the time that `font-feature-settings` would be dropped?

Only checked Safari and Firefox but `font-feature-settings` without `font-variant-numeric` still works

```suggestion
```

colinrotherham: Or drop `font-feature-settings` and use ~`font-variant-numeric`~ `font-variant` for IE11 support?

```suggestion
  font-variant: tabular-nums if($important, !important, null);
```

packages/govuk-frontend/src/govuk/components/character-count/_index.scss
```diff
@@ -14,7 +14,7 @@
   }
 
   .govuk-character-count__message {
-    @include govuk-font($size: false, $tabular: true);
+    @include govuk-typography-tabular-numbers;
```
colinrotherham, line 17: Quite like your `govuk-font-variant-tabular-nums` suggestion

If we use `font-variant-numeric` we might want `-numeric` too:

```scss
@include govuk-font-variant-numeric-tabular-nums;
@include govuk-font-variant-numeric-oldstyle-nums; // Maybe one day
```

But made me realise that `font-variant` already works in IE11

```scss
@include govuk-font-variant-tabular-nums;
@include govuk-font-variant-oldstyle-nums; // Maybe one day
```

packages/govuk-frontend/src/govuk/helpers/_typography.scss
querkmachine: `font-feature-settings` isn't at risk of being dropped, but using it is discouraged these days due to its broadness and tendency to overreach, as I described in the thoughts section, as well as the sometimes obtuse syntax. 

Newer, more bespoke properties like `font-variant-numeric` were introduced to solve those problems, but IE doesn't support them as a result.

CanIUse doesn't seem to mention this, but IE 11 only supports the `small-caps` keyword for `font-variant`. `font-variant: tabular-nums` unfortunately doesn't do anything there.

colinrotherham: Ah that's a shame it's only `small-caps`, was hoping we could have the best of both worlds

I'd be happy dropping IE11 support since we'd still default to proportional not monospaced numbers

Character count text updates would jump a bit in IE11 (variable number width) but I think that's acceptable

owenatgov (commented):
One final comment. This is basically good to go as far as I'm concerned.

Would you like an approval tick now, pending my last comment, or once 5.0 is out?

packages/govuk-frontend/src/govuk/helpers/_typography.scss
```diff
@@ -62,6 +62,21 @@
   font-weight: $govuk-font-weight-bold if($important, !important, null);
 }
 
+/// Tabular number helper
+///
+/// Using font-feature-settings is generally discouraged, but unfortunately
```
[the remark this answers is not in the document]
querkmachine: D'oh! Replaced it with a little explainer on why tabular numbers exist and what they're good for.

colinrotherham (approved):

36degrees (commented):
I think this broadly makes sense, but wondering if we should simplify the name slightly and just call it `govuk-tabular-numbers`?

I think it's obvious from what it does that it relates to fonts, and it feels a tad more memorable / guessable?

querkmachine:
@36degrees As per [Owen's comment](https://github.com/alphagov/govuk-frontend/pull/4307#pullrequestreview-1661456283) and [https://github.com/alphagov/govuk-frontend/issues/4243](the naming issue), I think we're kinda leaning towards typography-related mixins being namespaced under `govuk-font-` to match how most of the CSS properties they're setting begin with `font-`. 

I don't much mind if it's something like `govuk-font-tabular-numbers` though!

packages/govuk-frontend/src/govuk/helpers/_typography.scss
```diff
@@ -62,6 +62,23 @@
   font-weight: $govuk-font-weight-bold if($important, !important, null);
 }
 
+/// Tabular number helper
+///
+/// Switches numerical glyphs (0–9) to use alternate forms with a
```
colinrotherham, line 67: Sorry me again with my British words 🎩

```suggestion
/// Switches numerical glyphs (0–9) to use alternative forms with a
```

querkmachine: Pip pip ☕

owenatgov (commented):
This just needs a changelog entry and we're in the clear

querkmachine:
@owenatgov Added a changelog entry. I wasn't totally sure how to describe the change or why people should use it over `govuk-font`. I opted to describe it as a new feature that can be used to slightly optimise CSS file sizes, but there might be a better take on it.

owenatgov (approved):
I think the way you've described the change works well! Efficiency and neatness is what it's about really. Double approved!

PR #4317, opened 2023-10-09
Comment gardening 🪴

36degrees:
The JSDoc for most of our functions is ever so slightly too wide as all of the comments were indented by 2 characters when we moved to native ES6 classes.

- Wrap comments at 80 characters where possible
- Remove self-explanatory comments (e.g. ‘loop over X’)
- Small edits for clarity / conciseness

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
+++ b/packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
@@ -199,7 +199,6 @@ export class Accordion extends GOVUKFrontendComponent {
    * @private
    */
   initSectionHeaders() {
-    // Loop through sections
     this.$sections.forEach(($section, i) => {
       const $header = $section.querySelector(`.${this.sectionHeaderClass}`)
       if (!$header) {
@@ -213,8 +212,8 @@ export class Accordion extends GOVUKFrontendComponent {
       // Handle events
       $header.addEventListener('click', () => this.onSectionToggle($section))
 
-      // See if there is any state stored in sessionStorage and set the sections to
-      // open or closed.
+      // See if there is any state stored in sessionStorage and set the sections
+      // to open or closed.
       this.setInitialState($section)
     })
   }
@@ -235,7 +234,8 @@ export class Accordion extends GOVUKFrontendComponent {
       return
     }
 
-    // Create a button element that will replace the '.govuk-accordion__section-button' span
+    // Create a button element that will replace the
+    // '.govuk-accordion__section-button' span
     const $button = document.createElement('button')
     $button.setAttribute('type', 'button')
     $button.setAttribute(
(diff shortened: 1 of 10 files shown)


packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
```diff
@@ -235,19 +234,19 @@ export class Accordion extends GOVUKFrontendComponent {
...
       'aria-controls',
       `${this.$module.id}-content-${index + 1}`
     )
 
-    // Copy all attributes (https://developer.mozilla.org/en-US/docs/Web/API/Element/attributes) from $span to $button
+    // Copy all attributes from $span to $button (except `id`, which gets added
+    // to the `$headingText` element)
     for (let i = 0; i < $span.attributes.length; i++) {
       const attr = $span.attributes.item(i)
-      // Add all attributes but not ID as this is being added to
```
romaricpascal, line 249: *nitpick* Wondering if that comment is not actually useful because it explains that ID will get its own special treatment with the heading text, so may help not miss the `$headingText.id = $span.id`.

packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
```diff
@@ -256,23 +255,25 @@ export class Accordion extends GOVUKFrontendComponent {
...
+    // (see https://www.w3.org/TR/2011/WD-html5-20110525/content-models.html#phrasing-content)
     $headingTextFocus.innerHTML = $span.innerHTML
 
     // Create container for show / hide icons and text.
     const $showHideToggle = document.createElement('span')
     $showHideToggle.classList.add(this.sectionShowHideToggleClass)
-    // Tell Google not to index the 'show' text as part of the heading
-    // For the snippet to work with JavaScript, it must be added before adding the page element to the
-    // page's DOM. See https://developers.google.com/search/docs/advanced/robots/robots_meta_tag#data-nosnippet-attr
+    // Tell search engines not to index the 'show' text as part of the heading.
```
romaricpascal, line 274: *issue* I think that data-nosnippet is actually a Google thing. All I can find regarding Bing support is [a tweet from last year saying they don't support it, through a plethora of SEO articles](https://duckduckgo.com/?q=bing+data-nosnippet&t=newext&atb=v341-1). Has there been a confirmation of support since?

romaricpascal (commented):
Cheers for the tidy up 😄 Noticed two little things we may want to look at again in the Accordion, otherwise, all neat 🥳

packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
36degrees: I've moved it to the comment before the loop – think that's too far away from the code?

romaricpascal: Oh, I'd missed it, sorry. Yeah, all good 🥳

romaricpascal (approved):

PR #4318, opened 2023-10-09
Prefer `classList.toggle` where possible

36degrees:


romaricpascal (approved):
Damn, that shortens things nicely 🙌🏻

PR #4320, opened 2023-10-10
Move `getFragmentFromUrl()` to common utilities

colinrotherham:
This PR moves duplicate `getFragmentFromUrl()` methods into a common utility + tests

This has highlighted a few areas where we hadn't checked for missing hash fragments:

```patch
- const panelId = this.getHref($tab).slice(1)
+ const panelId = getFragmentFromUrl($tab.href)
+ if (!panelId) {
+   return
+ }
```

### Total bundle size
`all.mjs` ~68.5 KiB~ 68.33 KiB

### DIFF
--- a/packages/govuk-frontend/src/govuk/common/index.jsdom.test.mjs
+++ b/packages/govuk-frontend/src/govuk/common/index.jsdom.test.mjs
@@ -1,7 +1,8 @@
 import {
   mergeConfigs,
   extractConfigByNamespace,
-  isSupported
+  isSupported,
+  getFragmentFromUrl
 } from './index.mjs'
 
 describe('Common JS utilities', () => {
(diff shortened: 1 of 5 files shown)


romaricpascal (commented):
I like that we're moving code that's got no link to `this` out of classes 🥳 

Doesn't seem like the extra empty returns make users lose any information as it doesn't look like the previous implementation was throwing in case of missing hash in the URL (would have returned the last character of the URL). What do you think of throwing an `ElementError` to make things more visible?

colinrotherham:
>What do you think of throwing an ElementError to make things more visible?

Yes please, that's the reason 🙌

Would you mind if we limited this one to "move the thing" only?

We can ensure **Tabs** throws like **Error summary** and **Skip link** in another PR

romaricpascal (approved):
> Would you mind if we limited this one to "move the thing" only?

Fine by me 😊

PR #4321, opened 2023-10-10
Assert Exit this page `HTMLAnchorElement` button and clearly guard all dynamic fields

colinrotherham:
This PR is a follow up to https://github.com/alphagov/govuk-frontend/issues/4129 (see https://github.com/alphagov/govuk-frontend/pull/4199) to fix the "link button" check:

```patch
- if (!($button instanceof HTMLElement)) {
+ if (!($button instanceof HTMLAnchorElement)) {
```

We should catch this early to ensure `location.href` changes work correctly

```patch
- window.location.href = this.$button.getAttribute('href')
+ window.location.href = this.$button.href
```

I've also added some extra guards for dynamic elements so it's clear which ones we chose to throw errors for

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/exit-this-page/exit-this-page.mjs
+++ b/packages/govuk-frontend/src/govuk/components/exit-this-page/exit-this-page.mjs
@@ -292,7 +292,7 @@ export class ExitThisPage extends GOVUKFrontendComponent {
       this.updateIndicator()
 
       // Clear the timeout for the keypress timeout message clearing itself
-      if (this.timeoutMessageId !== null) {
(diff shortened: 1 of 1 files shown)


domoscargin (commented):
Looks good to me, though I've had to get my head out of the Error messages space to allow it to accept early returns again 😆 

Could we sneak in a coupla words to cover the guards in the PR title? Or give it a more general message about tightening up checks?

domoscargin (approved):

colinrotherham:
@domoscargin Thanks for reviewing, all updated and rebased

domoscargin (approved):
🚀

PR #4322, opened 2023-10-10
Reduce file size using `for...of` loops + other optimisations

colinrotherham:
With IE11 not running JavaScript `type="module"` we can now:

1. Loop using [`for...of`](https://caniuse.com/mdn-javascript_statements_for_of)
2. Map attributes using `Array.from()`
3. Map objects with `Object.entries()`
4. Add default and [`...` rest parameters](https://browsersl.ist/#q=supports+es6-module+and+not+supports+rest-parameters)
5. [Remove `Intl.NumberFormat` guards](https://browsersl.ist/#q=supports+es6-module+and+not+supports+internationalization)

Saves just over 1KB from `all.mjs` when bundled

### Total bundle size
`all.mjs` ~68.5 KiB~ 67.33 KiB

### DIFF
--- a/packages/govuk-frontend/src/govuk/common/index.mjs
+++ b/packages/govuk-frontend/src/govuk/common/index.mjs
@@ -49,11 +49,10 @@ export function mergeConfigs(...configObjects) {
         // If the value is a nested object, recurse over that too
         if (typeof value === 'object') {
           flattenLoop(value, prefixedKey)
-          continue
+        } else {
+          // Otherwise, add this value to our return object
+          flattenedObject[prefixedKey] = value
         }
-
-        // Otherwise, add this value to our return object
-        flattenedObject[prefixedKey] = value
       }
     }
 
--- a/packages/govuk-frontend/src/govuk/common/index.mjs
+++ b/packages/govuk-frontend/src/govuk/common/index.mjs
@@ -88,21 +88,8 @@ export function mergeConfigs(...configObjects) {
  * @param {{ [key: string]: unknown }} configObject - The object to extract key-value pairs from.
  * @param {string} namespace - The namespace to filter keys with.
  * @returns {{ [key: string]: unknown }} Flattened object with dot-separated key namespace removed
- * @throws {Error} Config object required
- * @throws {Error} Namespace string required
  */
 export function extractConfigByNamespace(configObject, namespace) {
-  // Check we have what we need
-  if (!configObject || typeof configObject !== 'object') {
-    throw new Error('Provide a `configObject` of type "object".')
-  }
-
-  if (!namespace || typeof namespace !== 'string') {
-    throw new Error(
-      'Provide a `namespace` of type "string" to filter the `configObject` by.'
-    )
-  }
-
   /** @type {{ [key: string]: unknown }} */
   const newObject = {}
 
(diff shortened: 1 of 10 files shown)


packages/govuk-frontend/src/govuk/common/index.mjs
```diff
@@ -41,23 +42,18 @@ export function mergeConfigs(/* configObject1, configObject2, ...configObjects *
...
         const prefixedKey = prefix ? `${prefix}.${key}` : key
+
+        // If the value is a nested object, recurse over that too
         if (typeof value === 'object') {
-          // If the value is a nested object, recurse over that too
           flattenLoop(value, prefixedKey)
-        } else {
-          // Otherwise, add this value to our return object
-          flattenedObject[prefixedKey] = value
+          continue
```
36degrees, line 52: I think it's clearer to keep the `if/else` here, unless there are other benefits to using continue I'm missing.

36degrees (commented):
Broadly all makes sense, though there's one minor change I'm not 100% sold on.

The PR title is a bit misleading as there a number of other optimisations being made (thanks for listing them in the PR description!) – can you update the title so it better represents what the PR does?

packages/govuk-frontend/src/govuk/common/index.mjs
colinrotherham: I've matched the "early return" pattern we use elsewhere, except here as an "early continue"

When there's only one option left the `else { }` becomes redundant:

```mjs
for (loop me please) {
  if (somethingElse) {
    continue
  }
  
  // You're still here? Best do the default thing
  flattenedObject[prefixedKey] = value
}
```

36degrees: Personally, I generally try and avoid `continue` – my thinking pretty much aligns with what's in [this blog post](https://www.teamten.com/lawrence/programming/avoid-continue.html).

I'm not sure I'd go as far as to [disallow it entirely](https://eslint.org/docs/latest/rules/no-continue), but where there's a simple alternative (like `if/else`) I'd generally prefer that as I think it's easier to understand at a glance.

I don't want to block this PR on it, so if you feel strongly about preferring `continue` we can continue (hah…) with it for now, but it'd be good to discuss it at a dev catch up soon? Keen to see what others think.

romaricpascal: I think `if/else` makes things clearer here as the code is about doing two different things depending on the situation, rather than just skipping when the condition is met. 

I think I also lean toward avoiding `continue` if you can (eg. by negating conditions in an `if` to only run code if necessary, rather than having an early `continue`). It's one less branch to consider when trying to understand what happens.

Wouldn't be averse to set `no-continue` to clarify that and discuss punctual lifts with `eslint-disable` for it if really `continue` is the only option to get us through. That would communicate the preference for avoiding it clearly 😊

colinrotherham: Fine by me, good to clarify things over time

E.g. We added our first `continue` in [`99d6135`](https://github.com/alphagov/govuk-frontend/commit/99d6135318bcdd0223f81d8de9400f379b98dff7#diff-d4f18af2a63bd7e3e0e9530109542122a62903104a76d3dc5e9ab376cef398bbR57-R61) but both are gone now

```mjs
// Check to see if this is a prototypical key/value,
// if it is, skip it.
if (!Object.prototype.hasOwnProperty.call(obj, key)) {
  continue
}
```

romaricpascal (approved):
Cheers for updating! All good to go ⛵

Issue #4036, opened 2023-08-01
Review ElementErrors and tie up any loose ends


romaricpascal:
## What

Review the MissingElementError work (tracked here: https://github.com/alphagov/govuk-frontend/issues/3478) and tie up any loose ends before a content review.

- [Skip Link](https://github.com/alphagov/govuk-frontend/issues/4128)
- [Exit this page](https://github.com/alphagov/govuk-frontend/issues/4129)
- [Header](https://github.com/alphagov/govuk-frontend/issues/4130)
- [Radios & Checkboxes](https://github.com/alphagov/govuk-frontend/issues/4131)
- [Character Count](https://github.com/alphagov/govuk-frontend/issues/4125)
- [Tabs](https://github.com/alphagov/govuk-frontend/issues/4127)
- [Accordion](https://github.com/alphagov/govuk-frontend/issues/4126)

## Why

This is part of our work to [have our components throw rather than silently return during initialisation](https://github.com/alphagov/govuk-design-system-architecture/blob/main/decision-records/009-throw-errors-component-constructors.md).

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [x] Review the list of affected components for consistency, functionality and any other concerns
- [x] Ensure errors are ready for content review
- [ ] #4323

romaricpascal:
Let's divide this issue into smaller ones, in order of complexity:
 
- Skip Link
- Exit this page
- Header
- Radios & Checkboxes
- Character Count
- Tabs
- Accordion
- Review any loose ends (keep this issue)

domoscargin:
From a discussion on slack:

We may want to look at removing the `element` param from the `ElementError` and just tell the error what's missing or invalid. We're using it like a boolean, and often just pass in `null`.

romaricpascal:
Maybe it's a sign that we need to have the `ElementError` just accept a generic string message and move the formatting to separate helper functions rather than pass that boolean-ish first param: `missingElementMessage` and `missingOrMistypedElementMessage`.

I'd like to propose something like the following (the passing of the `componentName` being a parallel concern as it affects multiple errors, we can always add in an extra `options` parameter to the constructor anyway or with a `componentErrorMessage` function to wrap other message).

```js
class ElementError extends Error {
  name = 'ElementError'
}

function missingElementMessage(identifier) {
  return `${identifier} is missing`  
}

// Two little updates here:
// - identifier as first param to keep in line with the `missingElementMessage
//
[...]

PR #4286, opened 2023-10-02
Allow ES2015+ supported features

colinrotherham:
This PR pulls together a few ES2015+ code changes with their supporting ESLint config

These were recommendations from the [ESLint stylistic typed linting](https://typescript-eslint.io/linting/configs/#projects-with-type-checking) plugin in https://github.com/alphagov/govuk-frontend/pull/4106

## Already supported by ES modules

* ~ES2015 `Array.from()`~ Done in [#4322](https://github.com/alphagov/govuk-frontend/pull/4322)
* ES2015 `''.includes()`
* ES2016 `[].includes()`
* ~ES2017 `Object.entries()`~ Done in [#4322](https://github.com/alphagov/govuk-frontend/pull/4322)

### Total bundle size
`all.mjs` ~69.3 KiB~ 68.66 KiB

## Babel transforms without polyfills

* ES2020 `?.` optional chaining (+ 0.09 KiB)
* ES2020 `??` nullish coalescing (+ 0.55 KiB)

### Total bundle size
`all.mjs` ~68.66 KiB~ 69.3 KiB

### Transform examples

Note the following size increases using `!=` versus `!==` to check for nullish rather than falsy values

#### Before

```mjs
if ($summary?.parentNode) {
  // ...
}

if (config.scope ?? document) {
  // ...
}
```

#### After

```mjs
if ($summary != null && $summary.parentNode) {
  // ...
}

var _config$scope
if ((_config$scope = config.scope) != null ? _config$scope : document)) {
  // ...
}
```

PR #4323, opened 2023-10-10
Review and guard all missing fields/attributes

colinrotherham:
Ahead of https://github.com/alphagov/govuk-frontend/issues/4036 I've spent some time reviewing all optional things accessed without guards

We should follow this PR up with any extra errors we'd like to throw, although I've already split out:

1. https://github.com/alphagov/govuk-frontend/pull/4321
2. https://github.com/alphagov/govuk-frontend/pull/4320

For example, in PR https://github.com/alphagov/govuk-frontend/pull/4320 we've identified tab links without hash fragments ([already handled in **Skip link**](https://github.com/alphagov/govuk-frontend/pull/4177))

### Missing elements, attributes etc

* Class fields with incorrect `null` default
* Selector `null` return values not handled
* Timeout and interval IDs cleared when `null`
* Confg allows `null` as `typeof value === 'object'`

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
+++ b/packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
@@ -312,7 +312,7 @@ export class Accordion extends GOVUKFrontendComponent {
     $button.appendChild(this.getButtonPunctuationEl())
 
     // If summary content exists add to DOM in correct order
-    if ($summary) {
+    if ($summary && $summary.parentNode) {
       // Create a new `span` element and copy the summary line content from the
       // original `div` to the new `span`. This is because the summary line text
       // is now inside a button element, which can only contain phrasing
@@ -507,6 +507,10 @@ export class Accordion extends GOVUKFrontendComponent {
    * @param {boolean} expanded - Section expanded
    */
   updateShowAllButton(expanded) {
+    if (!this.$showAllButton || !this.$showAllText || !this.$showAllIcon) {
+      return
+    }
+
     this.$showAllButton.setAttribute('aria-expanded', expanded.toString())
     this.$showAllText.textContent = expanded
       ? this.i18n.t('hideAllSections')
--- a/packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
+++ b/packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
@@ -113,7 +113,7 @@ export class Accordion extends GOVUKFrontendComponent {
   $showAllText = null
 
   /**
-   * @param {Element} $module - HTML element to use for accordion
+   * @param {Element | null} $module - HTML element to use for accordion
    * @param {AccordionConfig} [config] - Accordion config
    */
   constructor($module, config = {}) {
--- a/packages/govuk-frontend/src/govuk/components/character-count/character-count.mjs
+++ b/packages/govuk-frontend/src/govuk/components/character-count/character-count.mjs
@@ -261,7 +261,9 @@ export class CharacterCount extends GOVUKFrontendComponent {
    */
   handleBlur() {
     // Cancel value checking on blur
-    clearInterval(this.valueChecker)
+    if (this.valueChecker) {
+      window.clearInterval(this.valueChecker)
+    }
   }
 
   /**
--- a/packages/govuk-frontend/src/govuk/components/character-count/character-count.mjs
+++ b/packages/govuk-frontend/src/govuk/components/character-count/character-count.mjs
@@ -62,7 +62,7 @@ export class CharacterCount extends GOVUKFrontendComponent {
   maxLength = Infinity
 
   /**
-   * @param {Element} $module - HTML element to use for character count
+   * @param {Element | null} $module - HTML element to use for character count
    * @param {CharacterCountConfig} [config] - Character count config
    */
   constructor($module, config = {}) {
--- a/packages/govuk-frontend/src/govuk/components/character-count/character-count.mjs
+++ b/packages/govuk-frontend/src/govuk/components/character-count/character-count.mjs
@@ -59,7 +59,7 @@ export class CharacterCount extends GOVUKFrontendComponent {
   i18n
 
   /** @private */
-  maxLength = Infinity
+  maxLength
 
   /**
    * @param {Element | null} $module - HTML element to use for character count
@@ -128,7 +128,7 @@ export class CharacterCount extends GOVUKFrontendComponent {
     })
 
     // Determine the limit attribute (characters or words)
-    this.maxLength = this.config.maxwords || this.config.maxlength
+    this.maxLength = this.config.maxwords || this.config.maxlength || Infinity
 
     this.$module = $module
     this.$textarea = $textarea
--- a/packages/govuk-frontend/src/govuk/components/character-count/character-count.mjs
+++ b/packages/govuk-frontend/src/govuk/components/character-count/character-count.mjs
@@ -28,17 +28,11 @@ export class CharacterCount extends GOVUKFrontendComponent {
   /** @private */
   $textarea
 
-  /**
-   * @private
-   * @type {HTMLElement | null}
-   */
-  $visibleCountMessage = null
+  /** @private */
+  $visibleCountMessage
 
-  /**
-   * @private
-   * @type {HTMLElement | null}
-   */
-  $screenReaderCountMessage = null
+  /** @private */
+  $screenReaderCountMessage
 
   /**
    * @private
--- a/packages/govuk-frontend/src/govuk/components/header/header.mjs
+++ b/packages/govuk-frontend/src/govuk/components/header/header.mjs
@@ -39,7 +39,7 @@ export class Header extends GOVUKFrontendComponent {
    * Apply a matchMedia for desktop which will trigger a state sync if the
    * browser viewport moves between states.
    *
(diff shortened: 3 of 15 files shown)


packages/govuk-frontend/src/govuk/components/character-count/character-count.mjs
```diff
@@ -132,7 +132,7 @@ export class CharacterCount extends GOVUKFrontendComponent {
     })
 
     // Determine the limit attribute (characters or words)
-    this.maxLength = this.config.maxwords || this.config.maxlength
+    this.maxLength = this.config.maxwords || this.config.maxlength || Infinity
```
romaricpascal, line 129: **question** Given how we validate the configuration, we'll never reach the `|| Infinity`, I think. What's requiring us to keep it?

packages/govuk-frontend/src/govuk/components/tabs/tabs.mjs
```diff
@@ -94,6 +91,7 @@ export class Tabs extends GOVUKFrontendComponent {
 
     this.$tabList = $tabList
     this.$tabListItems = $tabListItems
+    this.mql = window.matchMedia('(min-width: 40.0625em)')
```
romaricpascal, line 94: **question** What's prompting the move of that line out of `setupResponsiveChecks`, the method was neatly encapsulating all the responsive setup before?

packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
```diff
@@ -486,6 +486,10 @@ export class Accordion extends GOVUKFrontendComponent {
    * @param {boolean} expanded - Section expanded
    */
   updateShowAllButton(expanded) {
+    if (!this.$showAllButton || !this.$showAllText || !this.$showAllIcon) {
+      return
+    }
```
romaricpascal, line 491: **question** We're creating these elements ourselves, what's requiring us to add this early return?

packages/govuk-frontend/src/govuk/components/header/header.mjs
```diff
@@ -118,6 +118,10 @@ export class Header extends GOVUKFrontendComponent {
    * @private
    */
   syncState() {
+    if (!this.mql || !this.$menu || !this.$menuButton) {
+      return
+    }
```
romaricpascal, line 123: **question** I _think_ this will never happen:
* We early return if `this.$menu` is falsy
* We throw if `this.$menu` exists and `this.$menuButton` is falsy
* And if we haven't thrown, we create `this.mql` ourselves

What's requiring us to add this early return?

romaricpascal (commented):
Most of the changes look good to me 🙌🏻 In a few places it looks like we're adding code branches that'll never run, though, so I've added a few questions to try and understand why they're needed. I'm worried that to please TypeScript's strict mode (in #4106), we're having to add `if`s (or `||`) of which we know the result for sure (but TypeScript doesn't) 😬

packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
colinrotherham: Ah we [discussed via Slack thread](https://gds.slack.com/archives/C05P6FJ1YH2/p1696235949838579?thread_ts=1695976340.212499&cid=C05P6FJ1YH2) to use silent returns (not errors) for self-created elements

I've been running [ESLint `@typescript-eslint/stylistic-type-checked`](https://typescript-eslint.io/linting/configs/#projects-with-type-checking) with @domoscargin in https://github.com/alphagov/govuk-frontend/pull/4106 so we can teach ESLint to be aware of types. These three are all `null` at instantiation

packages/govuk-frontend/src/govuk/components/character-count/character-count.mjs
colinrotherham: Ah it's only moved it from the class field to here to make the compiler happy in strict mode

Perhaps when we pick up https://github.com/alphagov/govuk-frontend/issues/4230 we can look at `CharacterCountConfig` becoming `Required<CharacterCountConfig>` when all fields are validated?

Then the compiler will know a fallback is no longer needed

packages/govuk-frontend/src/govuk/components/header/header.mjs
colinrotherham: Same as we [discussed via Slack thread again](https://gds.slack.com/archives/C05P6FJ1YH2/p1696235949838579?thread_ts=1695976340.212499&cid=C05P6FJ1YH2)

Don't forget that `this.$menu` being falsy causes `this.mql` to stay as `null`

Unlike in **Tabs**, the compiler spotted that `this.mql` isn't always set in the constructor

Can see all these with strict mode turned on in **packages/govuk-frontend/tsconfig.build.json**

```patch
{
  "extends": "../../tsconfig.base.json",
  "include": ["./src/govuk/**/*.mjs"],
  "exclude": ["**/*.test.*"],
  "compilerOptions": {
    "lib": ["ESNext", "DOM"],
+  "strict": true,
    "target": "ES2015",
    "types": ["node"]
  }
}
```

packages/govuk-frontend/src/govuk/components/tabs/tabs.mjs
colinrotherham: The compiler probably thinks that `!$tabListItems.length` causes it to stay as `mql = null`

By moving it into the constructor we don't need to add extra `null` checks in `setupResponsiveChecks()`

packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
romaricpascal: Returning rather than throwing is fine as we're creating these elements, as discussed on the Slack thread. 

Shame TypeScript needs an extra nudge leading to us checking things we do know exist 😔

colinrotherham: Yeah it's a shame, but good to be consistent

Alternatively we could consider getters and avoid `null` defaults entirely?

Here's a diff of what that would look like:
https://github.com/alphagov/govuk-frontend/compare/null-checks...null-checks-getters

packages/govuk-frontend/src/govuk/components/header/header.mjs
colinrotherham: As an alternative, I've mentioned [getters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get) in https://github.com/alphagov/govuk-frontend/pull/4323#discussion_r1357003285

Changing `this.mql` to a [getter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get) (without a [setter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/set)) ensures it only runs when first accessed:

```mjs
get mql() {
  if (!this._mql) {
    this._mql = window.matchMedia('(min-width: 48.0625em)')
  }

  return this._mql
}
```

colinrotherham:
@romaricpascal I've split this PR into separate commits to show what each change is for

Agree that `textContent` being `null` is frustrating so I've suggested an alternative below:

```patch
- ariaLabelParts.push(($summary.textContent || '').trim())
+ ariaLabelParts.push(`${$summary.textContent}`.trim())
```

But this is only necessary when running the TypeScript compiler with [`strict`](https://www.typescriptlang.org/tsconfig#strict) or [`strictNullChecks`](https://www.typescriptlang.org/tsconfig#strictNullChecks) enabled since the specification shows all ([**Node** interface `textContent` properties](https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent)) can be `null`

From the [mdn **Node** page](https://developer.mozilla.org/en-US/docs/Web/API/Node):

> In addition, every kind of DOM node is represented by an interface based on Node. These include [Attr](https://developer.mozilla.org/en-US/docs/Web/API/Attr), [CharacterData](https://developer.mozilla.org/en-US/docs/Web/API/CharacterData) (which [Text](https://developer.mozilla.org/en-US/docs/Web/API/Text), [Comment](https://developer.mozilla.org/en-US/docs/Web/API/Comment), [CDATASection](https://developer.mozilla.org/en-US/docs/Web/API/CDATASection) and [ProcessingInstruction](https://developer.mozilla.org/en-US/docs/Web/API/ProcessingInstruction) are all based on), and [DocumentType](https://developer.mozilla.org/en-US/docs/Web/API/DocumentType).

It's a shame the compiler can't filter out other Node types that aren't possible

colinrotherham:
@romaricpascal @domoscargin I've updated this PR to move anything "new" or for discussion into:

* https://github.com/alphagov/govuk-frontend/pull/4286
* https://github.com/alphagov/govuk-frontend/pull/4106

This PR now includes only guards we've used elsewhere

packages/govuk-frontend/src/govuk/components/tabs/tabs.mjs
colinrotherham: Moved to https://github.com/alphagov/govuk-frontend/pull/4286

romaricpascal (approved):
Cheers for reducing the scope of it to the less controversial changes. Happy with that ⛵ . 

I'm warming up to that getter pattern as well, it's quite neat. On your branch, it make the build of the elements of the accordion neatly encapsulated, on top of fixing the `null` issue.

Issue #4305, opened 2023-10-05
Review Nunjucks `default()` fallbacks for "falsey" values


JoPintoPaul:
## Description of the issue

When using the tabs template, it used to be the case the an empty string ("") being passed as title would result in the default string "Contents" being displayed:
https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/src/govuk/components/tabs/template.njk#L7

However, since Nunjucks 2.0, an empty string will NOT trigger the default, as Nunjucks now only uses default when a value is undefined rather than falsey:
https://mozilla.github.io/nunjucks/templating.html#default-value-default-boolean

If the template is passed an empty string for the title parameter, should the page render as:
```
<h2 class="govuk-tabs__title"></h2>
```
or as:
```
<h2 class="govuk-tabs__title">Contents</h2>
```
## Steps to reproduce the issue
Pass an empty string ("") for the title to the tabs template:
https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/src/govuk/components/tabs/template.njk

## Actual vs expected behaviour
Expected: Passing an empty string for title renders as `<h2 class="govuk-tabs__title">Contents</h2>`
Actual: `<h2 class="govuk-tabs__title"></h2>`

colinrotherham:
Thanks for opening this @JoPintoPaul

I did wonder if our `default()` usage needed tweaking so I've update the title for a wider review

Your example is spot on though, we always need a title so "falsey" values should be skipped:

```patch
<h2 class="govuk-tabs__title">
- {{ params.title | default("Contents") }}
+ {{ params.title | default("Contents", true) }}
</h2>
```

But in contrast we'd take care with other "falsey" values such as:

1. [**Checkboxes** input](https://design-system.service.gov.uk/components/checkboxes/) `item.checked` should preserve `false` values
2. [**Select** menu](https://design-system.service.gov.uk/components/select/) `item.selected` should preserve `false` values

We also use `params.value` without `default()` but maybe they should keep `""` values too

Will be good to tighten things up, thanks again

JoPintoPaul:
@colinrotherham Great stuff, thanks for the update!

colinrotherham:
Thanks for raising this @JoPintoPaul

We'll have it fixed in our next release

PR #4327, opened 2023-10-11
Fix Nunjucks `default()` values when `null`, `false` or `""` is provided

colinrotherham:
This PR ensures Nunjucks `default()` handles all possible "falsey" values

It closes https://github.com/alphagov/govuk-frontend/issues/4305 which shows that `undefined` values fall back to `defaults()` but not `null`, `false`, `""` etc

### Nunjucks default example

For example, passing `redirectUrl: null` into [**Exit this page**](https://design-system.service.gov.uk/components/exit-this-page/) renders a default submit button 😮

```njk
{{ govukExitThisPage({
  redirectUrl: null
}) }}
```

The fallback `href` value isn't used:

```html
<button type="submit">…</button>
```

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/select/template.njk
+++ b/packages/govuk-frontend/src/govuk/components/select/template.njk
@@ -47,7 +47,7 @@
       {# Allow selecting by text content (the value for an option when no value attribute is specified) #}
       {% set effectiveValue = item.value | default(item.text) %}
       <option {%- if item.value !== undefined %} value="{{ item.value }}"{% endif %}
-        {{-" selected" if item.selected | default(effectiveValue == params.value if params.value else false, true) }}
+        {{-" selected" if item.selected | default((effectiveValue == params.value and item.selected != false) if params.value else false, true) }}
         {{-" disabled" if item.disabled }}
         {%- for attribute, value in item.attributes %} {{ attribute }}="{{ value }}"{% endfor -%}>{{ item.text }}</option>
     {% endif %}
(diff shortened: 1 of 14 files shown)


packages/govuk-frontend/src/govuk/components/select/template.njk
```diff
@@ -47,7 +47,7 @@
       {# Allow selecting by text content (the value for an option when no value attribute is specified) #}
       {% set effectiveValue = item.value | default(item.text) %}
       <option {%- if item.value !== undefined %} value="{{ item.value }}"{% endif %}
-        {{-" selected" if item.selected | default(effectiveValue == params.value if params.value else false, true) }}
+        {{-" selected" if item.selected | default((effectiveValue == params.value and item.selected != false) if params.value else false, true) }}
```
domoscargin, line 50: When will this trigger? My brain's spinning in circles...

domoscargin:
Looks good, I just need an explainer on the `item.selected` stuff because it's still too early for my brain.

packages/govuk-frontend/src/govuk/components/select/template.njk
colinrotherham: It's a weird case where `govukSelect()` can have a _selected_ value but the item itself is _unselected_ 🤯

So we want to keep the falsey value as long as it's actually `false` and not `null` or `""`

https://github.com/alphagov/govuk-frontend/blob/72c543a0447c15ea61040cf2fbf901005fe1a0c2/packages/govuk-frontend/src/govuk/components/select/select.yaml#L350-L358

It's covered by this test case:

https://github.com/alphagov/govuk-frontend/blob/72c543a0447c15ea61040cf2fbf901005fe1a0c2/packages/govuk-frontend/src/govuk/components/select/template.test.js#L101-L106

domoscargin (approved):
I was going to do a foot emoji and a bug emoji for squashing the bug, but let's go with releasing the bug into the garden instead.

🐛 🏞️

colinrotherham:
Thanks @domoscargin just added a CHANGELOG entry too

PR #4329, opened 2023-10-13
Add lib functions `renderTemplate()` and `renderPreview()` etc

colinrotherham:
We chatted on Slack yesterday that we could bypass the Review app and render HTML using:

```mjs
// We have no "full template" renderer
const html = '???'

// Inject rendered HTML into the page
await page.setContent(html)
```

This PR adds new lib functions `renderTemplate()` and `renderPreview()` to make it possible

✅ Pre-built styles `govuk-frontend.min.css`
✅ Pre-built script `govuk-frontend.min.js`
✅ Lib functions to render components
❌ Lib functions to [render templates and previews](http://govuk-frontend-review.herokuapp.com/tests/boilerplate)

```mjs
// We have no "full template" renderer
const html = renderPreview(componentName, renderOptions)

// Inject rendered HTML into the page
await page.setContent(html)
```

### Shared test helpers and library functions

This PR also aligns options between render functions and helpers, matching up their names too

* **Test helpers** in `@govuk-frontend/helpers` import test and dev dependencies (local use only)
* **Library functions** in `@govuk-frontend/lib` import production dependencies (deployed to Heroku)

```mjs
import { render } from '@govuk-frontend/lib/components' // Returns strings
import { render } from '@govuk-frontend/helpers/nunjucks' // Returns Cheerio objects
import { render } from '@govuk-frontend/helpers/puppeteer' // Returns Puppeteer pages
```


packages/govuk-frontend-review/src/common/middleware/assets.mjs
```diff
@@ -19,6 +19,9 @@ const frontendPath = packageTypeToPath('govuk-frontend', {
 
 router.use('/assets', express.static(join(frontendPath, 'assets')))
 router.use('/javascripts', express.static(frontendPath))
-router.use('/stylesheets', express.static(join(paths.app, 'dist/stylesheets')))
+router.use('/stylesheets', [
+  express.static(frontendPath),
```
romaricpascal, line 23: **question** Just to confirm that I understand this change properly: this leads Express to also look in `govuk-frontend` (however resolved earlier in the file) alongside looking at the built stylesheets from the review app. In turn, this enables it to find the `/stylesheets/govuk-frontend.min.css` from the boilerplate markup, is that it?

romaricpascal (approved):
Looks good. I like the generic rendering to string from `lib` then specialised in `nunjucks` to get a cheerio object and in `puppeteer` to run on the page. Only a little question, but it's not a blocker, just a confirmation that I understood that change correctly.

packages/govuk-frontend-review/src/common/middleware/assets.mjs
colinrotherham: Yeah Express.js accepts an array of routes to try in order

Makes sure we can resolve CSS files from both locations:

```
/stylesheets/govuk-frontend.min.css
/stylesheets/app.min.css
```

PR #4330, opened 2023-10-13
Support `beforeInitialisation()` options, multiple modules

colinrotherham:
This PR is split out from https://github.com/alphagov/govuk-frontend/pull/4266 and includes:

1. Support custom `beforeInitialisation()` browser options
1. Support multiple `$modules` being initialised in `renderAndInitialise()`

This follows a "Custom Puppeteer browser options" [discussion on Slack](https://gds.slack.com/archives/C05P6FJ1YH2/p1695976340212499) and [first suggested](https://gds.slack.com/archives/C05P6FJ1YH2/p1695898993278089?thread_ts=1695827768.160929&cid=C05P6FJ1YH2) by @romaricpascal

### Custom initialisation example
Render a single button but clone it before both are initialised together

```mjs
const examples = await getExamples('button')

// Render single button
renderAndInitialise(page, 'button', {
  params: examples.default,

  // Duplicate button before initialisation
  beforeInitialisation($module, { className }) {
    const $button1 = $module
    
    // Create 2nd button
    const $button2 = el.cloneNode(true)

    // Append 2nd button, configure
    $button1.parentNode.appendChild($button2)
    $button2.setAttribute('class', className)
  },
  
  // Pass custom options to browser
  beforeInitialisationOptions: {
    className: 'govuk-button--dodgy'
  }
})
```

### DIFF
--- a/shared/helpers/puppeteer.js
+++ b/shared/helpers/puppeteer.js
@@ -119,13 +119,17 @@ async function renderAndInitialise(
   // them when back in Jest (to keep them triggering a Promise rejection)
   const error = await page.evaluate(
     async (exportName, config) => {
-      const $module = document.querySelector('[data-module]')
-
       const namespace = await import('govuk-frontend')
 
+      // Find all matching modules
+      const $modules = document.querySelectorAll('[data-module]')
+
       try {
-        /* eslint-disable-next-line no-new */
-        new namespace[exportName]($module, config)
+        // Loop and initialise all $modules or use default
+        // selector `null` return value when none found
+        ;($modules.length ? $modules : [null]).forEach(
+          ($module) => new namespace[exportName]($module, config)
+        )
       } catch ({ name, message }) {
         return { name, message }
       }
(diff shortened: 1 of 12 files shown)


shared/helpers/puppeteer.js
```diff
@@ -111,20 +117,25 @@ async function renderAndInitialise(page, componentName, options) {
...
 
+      // Find all matching modules
+      const $modules = document.querySelectorAll(selector)
+
       try {
-        /* eslint-disable-next-line no-new */
-        new namespace[exportName]($module, options.config)
+        // Loop and initialise all $modules or use default
+        // selector `null` return value when none found
+        ;($modules.length ? $modules : [null]).forEach(
```
domoscargin, line 129: Is this semi-colon doing something fancy?

colinrotherham: Yeah caught me out too

It's a Prettier protection against those parentheses  `()` becoming a function call
https://prettier.io/docs/en/rationale#semicolons

[ESLint does it too](https://eslint.org/docs/latest/rules/semi) but not nearly as much

Issue #4103, opened 2023-08-18
Restore ability for tests to fail when component initialisation throws


romaricpascal:
## What

Find a way to keep Puppeteer tests failing when they run `initAll` and an error happen on the page.

## Why

We [have set up Jest and Puppeteer to fail if an error happen on the page](https://github.com/alphagov/govuk-frontend/pull/2933). However, recently, [we've updated `initAll` to catch component's initialisation errors](https://github.com/alphagov/govuk-frontend/pull/4067).

A lot of our tests rely on accessing review app pages, which call `initAll`... so won't throw errors and fail tests if one of the components fail to initialise.

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [x] Devise a way to keep tests failing when `initAll` is called and a component throws during initialisation

romaricpascal:
I think it's another symptom of our tests [relying heavily on the review app](https://github.com/alphagov/govuk-frontend/issues/2565). Wondering if the long term fix wouldn't be to move them all to `renderAndInitialise`. Short term, it's probably fine checking for stacktraces in the logs or something similar.

colinrotherham:
Just an update on this as we've hit a snag

Some tests will instantly start to throw errors 😣

For example, **Skip link** errors without an element ID matching the hash fragment

This is because our examples are single-component by design, but we could:

1. Add multi-component previews
2. Add dummy elements to pass the checks
4. Add and document known failures (updating tests accordingly)
5. Prefer `initAll()` unless testing a custom config or `beforeInitialisation()` change
6. Sneakily use different code vs. preview examples like we do for [**Page template** guidance](https://design-system.service.gov.uk/styles/page-template/)

Ideas 1) and 2) impact both the Review app and Design System previews so we should involve a designer

We discussed this at dev catch-up and will leave it on the agenda

Issue #2565, opened 2022-03-01
Tests are reliant on the review app


vanitabarrett:
### Cause

Frontend Puppeteer tests rely on the review app in order to run correctly. The tests are tightly coupled to the review app's structure and the examples we use in the review app.

### Consequences

This causes us a few different issues. It overlaps with the issues described in https://github.com/alphagov/govuk-frontend/issues/2242 which is an issue we experience now. 

In the future, this may trip us up more when we want to add tests for configurable components. At the moment, components are already initialised by the time the tests are run. This wouldn't work for components that we want to configure and test.

It also generally doesn't feel ideal to have the tests and the review app so closely coupled. Ideally, we would have more control over the environment the tests run in without having to rely on there being an existing review app page that does what you want. This could be something like having an empty page which we can then add HTML into within the test suite.

When working on this, we need to consider the impact to debugging. Having tests associated with the review app does make it easier for us to debug if a test fails because we can go straight to that page to figure out what's going on. We might lose this if we decouple the tests and review app.

Note: I've split this tech debt card out separately from https://github.com/alphagov/govuk-frontend/issues/2242 because it may be possible to address #2242 without picking this wider piece of work up if we need a short term solution. However, it's possible we may be forced to address this as part of the JavaScript work if we're introducing configurable components.

### Impact of debt

Medium

### Effort to pay down

Medium

### Overall rating

Medium

romaricpascal:
A couple of thoughts on the topic, so I don't forget them:

1. wondering how much of the tests running through Puppeteer to run against a little static server rendering a blank boilerplate page (or populating `about:blank` or a blank page using a data-URL) and using our `renderAndInitialise()` helper to render examples. That would allow us to not be coupled with the review app anymore.
2. wondering if the app used for testing (that light one or the review app) could be made to run against the sources rather than our built package during development. Similarly to how you'd run an app for development locally and for production when deployed, we could have quick local tests that skip building and more secure testing on CI that run against the built package.

colinrotherham:
We used the following PRs to break the `govuk-frontend` package away from the Review app:

* https://github.com/alphagov/govuk-frontend/pull/4329
* https://github.com/alphagov/govuk-frontend/pull/4345
* https://github.com/alphagov/govuk-frontend/pull/4346
* https://github.com/alphagov/govuk-frontend/pull/4359

[...]

PR #4345, opened 2023-10-16
Update Puppeteer `render()` to bypass Review app

colinrotherham:
Part of https://github.com/alphagov/govuk-frontend/issues/4103

### DIFF
--- a/jest-puppeteer.config.js
+++ b/jest-puppeteer.config.js
@@ -18,13 +18,6 @@ module.exports = {
   launch: {
     args: [
       /**
-       * Workaround for 'No usable sandbox! Update your kernel' error
-       * see more https://github.com/Googlechrome/puppeteer/issues/290
-       */
-      '--no-sandbox',
-      '--disable-setuid-sandbox',
-
-      /**
        * Prevent empty Chrome startup window
        * Tests use their own `browser.newPage()` instead
        */
--- a/jest-puppeteer.config.js
+++ b/jest-puppeteer.config.js
@@ -18,6 +18,11 @@ module.exports = {
   launch: {
     args: [
       /**
+       * Allow tests to use file:// paths
+       */
+      '--allow-file-access-from-files',
+
+      /**
        * Prevent empty Chrome startup window
        * Tests use their own `browser.newPage()` instead
        */
--- a/shared/helpers/puppeteer.js
+++ b/shared/helpers/puppeteer.js
@@ -1,5 +1,5 @@
 const { AxePuppeteer } = require('@axe-core/puppeteer')
-const { ports } = require('@govuk-frontend/config')
+const { urls } = require('@govuk-frontend/config')
 const components = require('@govuk-frontend/lib/components')
 const { componentNameToClassName } = require('@govuk-frontend/lib/names')
 const slug = require('slug')
@@ -153,7 +153,7 @@ async function render(page, componentName, renderOptions, browserOptions) {
  * @returns {Promise<import('puppeteer').Page>} Puppeteer page object
  */
 async function goTo(page, path) {
-  const { href, pathname } = new URL(path, `http://localhost:${ports.app}`)
+  const { href, pathname } = new URL(path, urls.app)
 
   const response = await page.goto(href)
   const code = response.status()
--- a/shared/helpers/puppeteer.js
+++ b/shared/helpers/puppeteer.js
@@ -1,7 +1,11 @@
+const { readFileSync } = require('fs')
+const { join } = require('path')
+
 const { AxePuppeteer } = require('@axe-core/puppeteer')
-const { urls } = require('@govuk-frontend/config')
+const { paths, urls } = require('@govuk-frontend/config')
 const { renderPreview } = require('@govuk-frontend/lib/components')
 const { componentNameToClassName } = require('@govuk-frontend/lib/names')
+const mime = require('mime-types')
 const slug = require('slug')
 
 /**
@@ -76,6 +80,8 @@ async function axe(page, overrides = {}) {
  * @returns {Promise<import('puppeteer').Page>} Puppeteer page object
  */
 async function render(page, componentName, renderOptions, browserOptions) {
+  await page.setRequestInterception(true)
+
   const exampleName = renderOptions.fixture?.name ?? 'default'
   const exportName = componentNameToClassName(componentName)
   const selector = `[data-module="govuk-${componentName}"]`
@@ -85,8 +91,16 @@ async function render(page, componentName, renderOptions, browserOptions) {
     exampleName
   })
 
-  // Inject rendered HTML into the page
-  await page.setContent(renderPreview(componentName, renderOptions))
+  // Mock preview HTTP response
+  page.once('request', (request) =>
+    request.respond({
+      body: renderPreview(componentName, renderOptions),
+      contentType: 'text/html'
+    })
+  )
+
+  // Mock assets HTTP responses
+  page.on('request', requestHandler)
 
   // Navigate to component preview
   await goTo(page, route)
@@ -144,7 +158,41 @@ async function render(page, componentName, renderOptions, browserOptions) {
     )
   }
 
+  // Disable middleware
+  page.off('request', requestHandler)
+
+  await page.setRequestInterception(false)
+
   return page
+}
+
+/**
+ * Component preview HTTP request handler
+ *
+ * - Returns early for requests already handled
+ * - Responds to file:// asset requests (web fonts etc)
+ * - Skips unknown requests
+ *
+ * @param {import('puppeteer').HTTPRequest} request - Puppeteer HTTP request
+ * @returns {Promise<void>}
+ */
+async function requestHandler(request) {
+  if (request.isInterceptResolutionHandled()) {
+    return
+  }
+
+  const { pathname, protocol } = new URL(request.url())
+
+  // Return static assets
+  if (protocol === 'file:' && pathname.startsWith('/assets/')) {
+    return request.respond({
+      body: readFileSync(join(paths.package, `dist/govuk/${pathname}`)),
+      contentType: mime.lookup(pathname) || 'text/plain'
+    })
+  }
+
+  // Skip unknown requests
+  return request.continue()
 }
 
 /**
--- a/shared/helpers/puppeteer.js
+++ b/shared/helpers/puppeteer.js
@@ -1,6 +1,6 @@
 const { AxePuppeteer } = require('@axe-core/puppeteer')
 const { urls } = require('@govuk-frontend/config')
-const components = require('@govuk-frontend/lib/components')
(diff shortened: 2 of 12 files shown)


shared/helpers/puppeteer.js
```diff
@@ -111,49 +125,94 @@ async function render(page, componentName, renderOptions, browserOptions) {
...
+    page.off('request', middleware)
+
+    await page.setRequestInterception(false)
   }
 
   return page
 }
 
+/**
+ * Component preview HTTP request middleware
```
romaricpascal, line 172: **suggestion** I think it's worth describing what this method does in its docs. This can be infered from the code, but `middleware` may muddy things a little there (as well as leading with component preview, as the middleware deals with both serving the component preview and assets)

```suggestion
 * HTTP request middleware for component preview and assets
 *
 * Creates a handler for Puppeteer's request event that will:
 * - respond with the component preview if requested the provided `route`
 * - respond with the content for the relevant assets if requested an asset
 * - or let the request carry through
```

Alternatively, maybe renaming the function to `createRequestHandler` would help drop that blurb of doc:

```
/**
 * Create a handler for Puppeteer `request` event to serve component preview and assets
 * 
 * @param ...
 * @returns ...
 */
function createRequestHandler({ route, body, contentType = 'text/html' }) {
 // ...
}
```

shared/helpers/puppeteer.js
```diff
@@ -111,49 +125,94 @@ async function render(page, componentName, renderOptions, browserOptions) {
...
+      return request.respond({ body, contentType })
+    }
+
+    // Return static assets
+    if (protocol === 'file:' && pathname.startsWith('/assets/')) {
+      return request.respond({
+        body: readFileSync(join(paths.package, `dist/govuk/${pathname}`)),
+        contentType: mime.lookup(pathname) || 'text/plain'
+      })
+    }
```
romaricpascal, line 197: **question** With the paths to our JavaScript, CSS and assets folder being generated by [`pathToFileURL`](https://nodejs.org/api/url.html#urlpathtofileurlpath) in `components.js`, which URLs will have a path starting with `/assets/`?

shared/helpers/puppeteer.js
```diff
@@ -71,23 +75,33 @@ async function axe(page, overrides = {}) {
...
+  const middleware = renderMiddleware({
+    body: renderPreview(componentName, renderOptions),
+    route
+  })
+
+  // Enable middleware
+  page.on('request', middleware)
+
+  // Navigate to component preview
+  await goTo(page, route)
```
romaricpascal, line 104: **question** I follow what's happening, but would have thought that we'd just have to fill the content of the `page` with `page.setContent` and have our assets referenced with `file:` URLs for things to work. Is this dance with URLs and `goTo` so we benefit of the centralised things like bringing Puppeteer to front and waiting for fonts to load?

shared/helpers/puppeteer.js
```diff
@@ -163,6 +222,7 @@ async function goTo(page, path) {
     throw new Error(`HTTP ${code} for '${pathname}'`)
```
romaricpascal, line 222: **question** That's something for a separate PR, I think, but I'm wondering if by not hitting the review app, we're not undoing the work from [this PR](https://github.com/alphagov/govuk-frontend/pull/3963).

romaricpascal (commented):
That's a neat step towards adressing https://github.com/alphagov/govuk-frontend/issues/2565 🥳 Overall looks good and I think I followed OK, just want to confirm why a few things are happening. Don't think there'll be any blocker for the answers, but feels useful to have a trace of why things changed in a couple of places 😊

shared/helpers/puppeteer.js
colinrotherham: I think I'll split it out so it's only assets, we don't have to handle the preview here

jest-puppeteer.config.js
```diff
@@ -18,11 +18,10 @@ module.exports = {
   launch: {
     args: [
       /**
-       * Workaround for 'No usable sandbox! Update your kernel' error
-       * see more https://github.com/Googlechrome/puppeteer/issues/290
+       * Allow tests to use file:// paths
```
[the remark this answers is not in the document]
colinrotherham: To be honest, I've just checked and don't need `--disable-web-security` anymore

Now we've got the assets middleware we're running the same-origin via `file://` for everything

Puppeteer switched from Chromium to "Chrome for Testing" recently but can add a link in future

shared/helpers/puppeteer.js
```diff
@@ -111,49 +125,94 @@ async function render(page, componentName, renderOptions, browserOptions) {
...
+    // Throw Puppeteer errors back to Jest
+    if (error) {
+      throw new Error(
+        `Initialising \`new ${exportName}()\` with example '${exampleName}' threw:` +
+          `\n\t${error.name}: ${error.message}`,
+        { cause: error }
+      )
+    }
+
+    await page.evaluateHandle('document.fonts.ready')
```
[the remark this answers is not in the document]
colinrotherham: I've got this comment in the commit message:

>Since fonts are downloaded asynchronously, we should await the `document.fonts.ready` promise since further network requests may not be available once the page has loads

This is because we turn the assets middleware on and off using:

```
page.on('request', middleware)
page.off('request', middleware)
```

Network requests can continue after the `render()` Promise resolves

>Similarly, some components are made visible after initialising JavaScript (or new elements created and inserted) so we should wait for additional font downloads to complete there too
>
>E.g. GDS Transport Light may have loaded already, but GDS Transport Bold may be requested on demand as required

You'll have noticed this running with `HEADLESS=false` on the Accordion examples. Since Accordion sections are not visible during page load only the **GDS Transport** Bold font is loaded for the headings

Accordion sections opened by tests appeared in Arial (with an HTTP 404 for **GDS Transport** Light)

shared/helpers/puppeteer.js
```diff
@@ -111,49 +125,94 @@ async function render(page, componentName, renderOptions, browserOptions) {
...
   return page
 }
 
+/**
+ * Component preview HTTP request middleware
+ *
+ * @param {{ route: URL; body: string; contentType?: string }} options - Route options
+ * @returns {(request: import('puppeteer').HTTPRequest) => Promise<void>} Route middleware
+ */
+const renderMiddleware =
```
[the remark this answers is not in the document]
colinrotherham: Just picked the shortest one

What's your favourite?

```mjs
// 1. Arrow function to async arrow function
const createThing = (options) => async (request) => {
  // Return thing
}

// 2. Arrow function to async function
const createThing = (options) =>
  async function (request) {
    // Return thing
  }

// 3. Function to async arrow function
function createThing(options) {
  return async (request) => {
    // Return thing
  }
}

// 4. Function to async function
function createThing(options) {
  return async function (request) {
    // Return thing
  }
}
```

shared/helpers/puppeteer.js
colinrotherham: Yeah it's a lot tidier for sure 😊

There was the realisation that we'd want to keep:

1. Events like `pageshow`
2. Ability to reload pages
3. Ability to intercept page reloads
4. Ability to intercept page navigation

We only mock a single page request at the moment but there's room to grow

It also means page navigations can go back to the Review app again quite easily:

```patch
const route = getComponentURL(componentName, {
- baseURL: new URL('file://'),
  exampleName
})
```

☝️ But the middleware will need turning off too

shared/helpers/puppeteer.js
colinrotherham: We're using the same `goToComponent()` method still which runs `slug()`

So all the URL escaping is still there, we just mock the response now:

```mjs
const exampleName = slug(options?.exampleName ?? '', { lower: true })
```

jest-puppeteer.config.js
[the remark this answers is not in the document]
romaricpascal: Even better if we don't need the scary flag 🥳

shared/helpers/puppeteer.js
romaricpascal: Cool! Makes sense! And good to know that it's not much to get back to hitting the review app if we really need for some tests 😊

shared/helpers/puppeteer.js
romaricpascal: Ah good point, didn't think of those!

shared/helpers/puppeteer.js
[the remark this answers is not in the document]
romaricpascal: Cool, I think that makes things clearer for me, thanks for the extra detail. Should have mentioned that I had seen the comit but was really puzzled as to why it affected us more now than in our previous tests. Sounds like it's to ensure that `page.off` happens only after fonts are requested so that their request go through our request handler, which is very fine 👍🏻

shared/helpers/puppeteer.js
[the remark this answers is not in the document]
romaricpascal: My preference would go to 3 to match at least the other top level functions 😊

shared/helpers/puppeteer.js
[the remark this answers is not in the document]
colinrotherham: Brill. Similarly it will have happened when Express.js was torn down and test pages were still loading

shared/helpers/puppeteer.js
colinrotherham: Done ✅

shared/helpers/puppeteer.js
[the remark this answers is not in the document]
colinrotherham: Shall we mark this as resolved?

I no longer need to create a custom handler after doing https://github.com/alphagov/govuk-frontend/pull/4345#discussion_r1367054489

shared/helpers/puppeteer.js
colinrotherham: Don't forget too it's only the `/tests/boilerplate` route + `#slot` that this PR removes

All the component previews are still there

romaricpascal: Cool! That wasn't blocking, just wanted to flag it so we don't forget that at the moment `goTo` was both used for testing our components, as well as checking that the review app responded to the right URLs for their preview. But you likely had it in mind anyway given you did the work to add that check for 400 statused 😊

shared/helpers/puppeteer.js
```diff
@@ -71,24 +75,36 @@ async function axe(page, overrides = {}) {
...
+    baseURL: new URL('file://'),
+    exampleName
+  })
+
+  // Mock preview HTTP response
+  page.once('request', (request) =>
+    request.respond({
+      body: renderPreview(componentName, renderOptions),
+      contentType: 'text/html'
+    })
```
romaricpascal, line 99: Nice one, that makes things tidier for the handler serving the assets 😊

romaricpascal (approved):

PR #4346, opened 2023-10-16
Update accessibility tests to bypass Review app

colinrotherham:
Part of https://github.com/alphagov/govuk-frontend/issues/4103

This allows Accessibility tests to run without the Review app

For these tests, we've [chosen to log errors as warnings](https://github.com/alphagov/govuk-frontend/actions/runs/6610937177/job/17954246976#step:6:15) (rather than throw) due to invalid component examples:

1. Skip link with `#custom` target
2. Character count without "maxlength" or "maxwords"
3. Tabs with missing items

### Test performance

Worth mentioning that we went to 11-12 minutes per PR run by adding Windows tests

We’re down to [6 minutes without the Review app](https://github.com/alphagov/govuk-frontend/actions/runs/6724671614) once https://github.com/alphagov/govuk-frontend/pull/4359 merges :raised_hands:

### DIFF
--- a/shared/helpers/puppeteer.js
+++ b/shared/helpers/puppeteer.js
@@ -55,7 +55,10 @@ async function axe(page, overrides = {}) {
 
   // Add preview URL to report violations
   report.violations.forEach((violation) => {
-    violation.helpUrl = `${violation.helpUrl}\n${page.url()}`
+    const { pathname } = new URL(page.url())
+
+    // Replace file:// URLs with Review app
+    violation.helpUrl = `${violation.helpUrl}\n${new URL(pathname, urls.app)}`
   })
 
   return report
--- a/shared/helpers/puppeteer.js
+++ b/shared/helpers/puppeteer.js
@@ -67,9 +67,8 @@ async function axe(page, overrides = {}) {
 /**
  * Render component HTML with browser preview
  *
- * Renders a component's Nunjucks macro with the given params, injects it into
- * the test boilerplate page, and instantiates the component class, passing the
- * provided JavaScript configuration.
+ * Uses Nunjucks component {@link renderPreview} for HTML output, but runs
+ * component JavaScript (where available) or `initAll()` in the browser
  *
  * It runs an optional `beforeInitialisation` function before initialising the
  * components, allowing to tweak the state of the page before the component gets
--- a/shared/helpers/puppeteer.js
+++ b/shared/helpers/puppeteer.js
@@ -132,6 +132,11 @@ async function render(page, componentName, renderOptions, browserOptions) {
       async (selector, exportName, config) => {
         const namespace = await import('govuk-frontend')
 
+        // Skip custom initialisation without export
+        if (!namespace[exportName]) {
+          return namespace.initAll()
+        }
+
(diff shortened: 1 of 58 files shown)


shared/helpers/puppeteer.js
```diff
@@ -55,7 +55,10 @@ async function axe(page, overrides = {}) {
 
   // Add preview URL to report violations
   report.violations.forEach((violation) => {
-    violation.helpUrl = `${violation.helpUrl}\n${page.url()}`
+    const { pathname } = new URL(page.url())
+
+    // Replace file:// URLs with Review app
```
domoscargin, line 60: Quite a complicated string - might be helpful to include an example of what a `violation.helpUrl` looks like?

shared/helpers/puppeteer.js
```diff
@@ -132,6 +132,11 @@ async function render(page, componentName, renderOptions, browserOptions) {
       async (selector, exportName, config) => {
         const namespace = await import('govuk-frontend')
 
+        // Skip custom initialisation without export
```
domoscargin, line 142: Nice little time saver!

shared/helpers/puppeteer.js
```diff
@@ -128,6 +128,10 @@ async function render(page, componentName, renderOptions, browserOptions) {
   // property, which means we get a mangled value). As a workaround, we can
   // gather and `return` the values we need from inside the browser
   try {
+    if (!page.isJavaScriptEnabled()) {
```
domoscargin, line 134: Another nice time saver!

shared/lib/components.js
```diff
@@ -206,7 +206,7 @@ function renderPreview(componentName, options) {
       footer: '',
 
       main: outdent`
-        <div class="govuk-width-container">
```
domoscargin, line 209: I wonder if we can link to this function in the `helpers/puppeteer.render` JSDoc when we mention  the `test boilerplate page`. Would've made it a bit easier for this addled brain to figure out where that was coming from.

domoscargin (commented):
This is looking grand. Coupla little comments.

I think you mentioned on Slack the speed benefits to these changes. Nice to summarise those on the PR description if relevant!

shared/helpers/puppeteer.js
colinrotherham: Good shout, added a JSDoc `@example` to show our Review app example URL added on:

```console
You can find more information on this issue here:
https://dequeuniversity.com/rules/axe/4.8/aria-allowed-attr?application=axe-puppeteer
http://localhost:8080/components/radios/with-conditional-items/preview
```

shared/helpers/puppeteer.js
colinrotherham: Kinda, it crashes trying to run the JS without JS otherwise 🤦‍♂️

shared/lib/components.js
colinrotherham: Yeah another good idea. I've added some JSDoc `@link` references

domoscargin (approved):
⏳ 

Lovely stuff!

PR #4349, opened 2023-10-18
Refactor ElementError

### LINKED ISSUE #4300: Refactor ElementError class
## What
Potential changes include:

- Ensuring `ElementError.options.expectedType` is always a string, when present
- Simplifying expected type error messages to something like `${identifier} is not of type ${expectedType || 'HTMLElement'}`
- Make `identifier` the mandatory first param and move `element` into options:

```constructor(identifier, {componentName, element, expectedType})```

## Why
Making `expectedType` a string allows us to avoid having different cases for when it's a string vs when it's an element, reducing our checks and allowing us to simplify the outputted message.

Using `identifier` as our mandatory param allows us to avoid the hack of passing in `null` as the element to get the right error messages.

## Who needs to work on this
Developer

## Who needs to review this
Developer

## Done when

- [x] A decision has been made on which changes to make
- [ ] Changes have been implemented
- [ ] All existing tests and throws have been refactored to work with the changes

### PR DESCRIPTION
**Pull request by domoscargin**
Closes #4300 

- rearranges params in the constructor:
`constructor(identifier, {componentName, element, expectedType})`
- Expected type is now always a string when present
- Expected type errors have been simplified

### DIFF
--- a/packages/govuk-frontend/src/govuk/errors/index.mjs
+++ b/packages/govuk-frontend/src/govuk/errors/index.mjs
@@ -48,23 +48,18 @@ export class ElementError extends GOVUKFrontendError {
   name = 'ElementError'
 
   /**
-   * @param {Element | null} element - The element in error
+   * @param {string} identifier - An identifier that'll let the user understand which element has an error (variable name, CSS selector)
    * @param {object} options - Element error options
    * @param {string} options.componentName - The name of the component throwing the error
-   * @param {string} options.identifier - An identifier that'll let the user understand which element has an error (variable name, CSS selector)
-   * @param {string | typeof HTMLElement} [options.expectedType] - The type that was expected for the identifier
+   * @param {Element | null} [options.element] - The element in error
+   * @param {string} [options.expectedType] - The type that was expected for the identifier
    */
-  constructor(element, { componentName, identifier, expectedType }) {
+  constructor(identifier, { componentName, element, expectedType }) {
     let reason = `${identifier} not found`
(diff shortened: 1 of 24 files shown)

### CODE REVIEW

File: packages/govuk-frontend/src/govuk/errors/index.mjs [hunk 1]
```diff
@@ -48,23 +48,18 @@ export class ElementError extends GOVUKFrontendError {
   name = 'ElementError'
 
   /**
-   * @param {Element | null} element - The element in error
+   * @param {string} identifier - An identifier that'll let the user understand which element has an error (variable name, CSS selector)
    * @param {object} options - Element error options
    * @param {string} options.componentName - The name of the component throwing the error
-   * @param {string} options.identifier - An identifier that'll let the user understand which element has an error (variable name, CSS selector)
-   * @param {string | typeof HTMLElement} [options.expectedType] - The type that was expected for the identifier
+   * @param {Element} [options.element] - The element in error
```
colinrotherham, line 54: Mind if we keep your `Element | null` change from https://github.com/alphagov/govuk-frontend/pull/4263/commits/1377822a64c3b3eb92d32290f3edacc76bf9dc21?

Selectors `document.querySelector()` and `document.getElementById()` return `null` when not found 

```suggestion
   * @param {Element | null} [options.element] - The element in error
```

colinrotherham (approved):
Instant approval from me 🙌

Added one bonus comment regarding selectors returning `null` but have a look

Thanks @domoscargin

PR #4350, opened 2023-10-18
Log component example with Puppeteer JavaScript errors

colinrotherham:
Part of https://github.com/alphagov/govuk-frontend/issues/4103 and includes:

1. Puppeteer render errors output component and example information
1. Test helper `getExamples()` now includes component fixture data too

### Error log output before

Error log output from browser with unknown example

### Error log output after

Error log output from browser with component and example information

### DIFF
--- a/shared/helpers/puppeteer.js
+++ b/shared/helpers/puppeteer.js
@@ -78,6 +78,7 @@ async function axe(page, overrides = {}) {
 async function render(page, componentName, renderOptions, browserOptions) {
   await goTo(page, '/tests/boilerplate')
 
+  const exampleName = renderOptions.fixture?.name ?? 'default'
   const exportName = componentNameToClassName(componentName)
   const selector = `[data-module="govuk-${componentName}"]`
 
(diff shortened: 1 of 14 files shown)


shared/helpers/puppeteer.js
```diff
@@ -134,7 +135,11 @@ async function render(page, componentName, renderOptions, browserOptions) {
   )
 
   if (error) {
-    throw error
+    throw new Error(
+      `Initialising \`new ${exportName}()\` with example '${exampleName}' threw:` +
+        `\n\t${error.name}: ${error.message}`,
+      { cause: error }
+    )
```
romaricpascal, line 143: Nice way to surface all the info we need 🙌🏻

colinrotherham:
@romaricpascal The other commits are all related except the last (which moves a Jest spy)

1. **Add component fixture to `getExamples()` render options**
   3c248aeebabe1694a71b13659f32b6d6a79ccebc  adds YAML-to-JSON `fixture` to the Nunjucks options so we can log "Example name"

2. **Flag known examples with JavaScript errors**
   44b155f8ce382959469afabe0af9464602c50471 adds YAML `errors: true` to later suppress errors via Puppeteer

3. **Ensure Puppeteer errors log component and example names**
   882291fc187f86706946e88f89ca40bbc05dc79e adds the new logging shown in the description screenshots

4. **Skip Puppeteer errors for known examples with JavaScript errors**
   566501eaf9b1440352686f5eaed789309c525759 uses `errors: true` to conditionally skip `throw new Error()` via Puppeteer

5. **Move Jest `console.log()` spy to hide browser log spam**
   918753252a3e7e3a9500f7fbed9fb97b4664a43b mocks `console.log()` via `initAll()` so a test that logs an error no longer makes scary output

Issue #4299, opened 2023-10-04
Error message review: Use plain English and a selector/variable name in backticks as an identifier


domoscargin:
## What

Following on from [this doc](https://docs.google.com/document/d/1CIHxeSW7sjUA7GbP01v9xWrANeQeReCw1N5-d8U2dFI/edit#heading=h.6vmoib5b9l5x), we're aiming to format all our error messages with a mixture of plain English and selectors/variable names in backticks. 

In general, this might look like:

`<COMPONENT>: <PLAIN_ENGLISH_TARGET_NAME> (<CSS_SELECTOR>) <ERROR>`

For example:

```
Checkboxes: Input elements (`input[type="checkbox"]`) not found

Header: Button attribute (`.govuk-js-header-toggle[aria-controls]`) is not of type string

Character count: Form field (`.govuk-js-character-count`) is not of type `HTMLTextareaElement` or `HTMLInputElement`

Error summary: Root element (`[data-module=govuk-accordion]`) not found

Tabs: Link elements (`a.govuk-tabs__tab`) not found
```

## Why

We want users to be able to quickly identify exactly what's caused the error. Using plain English in addition to a specific selector should hopefully provide all the context the user needs.

## Who needs to work on this
Developer, Content Designer/Technical Writer

## Who needs to review this
Content Designer/Technical Writer, Developer

## Done when

- [ ] All identifiers match the above format

colinrotherham:
Looks great in plain English 🙌

Got a few little inconsistencies we could tweak?

For example, keeping the named element at the start:

```patch
- aria-controls attribute on menu toggle (`.govuk-js-header-toggle`) is not of type string
+ Menu toggle (`.govuk-js-header-toggle`) attribute (`aria-controls`) is not of type string
```

Or instead of names like "Menu toggle" we sometimes suffix "element", "form control" or "form field" too

Could we solely use the name?

```patch
- Checkbox form controls (`input[type="checkbox"]`) not found
+ Checkboxes (`input[type="checkbox"]`) not found
```

Or if we want a suffix, can we narrow them down?

```patch
- Checkbox form controls (`input[type="checkbox"]`) not found
+ Checkboxes elements (`input[type="checkbox"]`) not found
```

Last of all, shall we swap all JavaScript names to CSS selectors?

```patch
- Root element (`$module`) not found
+ Root element (`[data-module=govuk-accordion]`) not found
```

colinrotherham:
[...]

PR #4351, opened 2023-10-18
[WIP] Edit content of error messages

### LINKED ISSUE #4072: Review content of thrown errors

domoscargin:
Closes #4072 

Popping this one up so peeps can chime in early if they want to.

Mostly following this comment: https://github.com/alphagov/govuk-frontend/issues/4072#issuecomment-1766486106

## Custom messages
I've added a flag (`identifierIsMessage`) to allow passing a message directly through and ignore all the processing. I think possibly the better option here is to make `options` completely optional and skip all processing if it's not present.


packages/govuk-frontend/src/govuk/components/skip-link/skip-link.mjs
```diff
@@ -51,20 +52,19 @@ export class SkipLink extends GOVUKFrontendComponent {
 
     // Check for link hash fragment
     if (!linkedElementId) {
-      throw new ElementError('$module.hash', {
-        componentName: 'Skip link',
-        element: this.$module,
-        expectedType: 'string'
-      })
+      throw new ElementError(
```
domoscargin, line 55: Here's an example of passing a message instead of options.

packages/govuk-frontend/src/govuk/components/character-count/character-count.mjs
```diff
@@ -145,7 +145,7 @@ export class CharacterCount extends GOVUKFrontendComponent {
       throw new ElementError({
         componentName: 'Character count',
         element: $textareaDescription,
-        identifier: `#${textareaDescriptionId}`
+        identifier: `Hint element #${textareaDescriptionId}`
```
domoscargin, line 148: @claireashworth I'm not sure about the descriptor here.

In this case, the element that causes the error is the container for the count message at the bottom of a Character Count
Screenshot 2023-10-19 at 11 44 06

I've called it "Hint element" because it's using our Hint component, but the text above the textarea ("Do not include...") is more properly the "Hint". Other options might be "Description element" or "Count message element", but wanted to get your thoughts.

colinrotherham: We call it "count message" (8 times) in the [**Character count** guidance](https://design-system.service.gov.uk/components/character-count/)

>The count message appears below the textarea so that:
>
>* it’s clearly separate from any hint text or error message above the textarea

But we do say "character count message" once and call it a "static message" without JavaScript:

>This component uses JavaScript. If JavaScript is not available, users will see a static message in place of the count message

domoscargin: The "count message"'s have it!

packages/govuk-frontend/src/govuk/components/header/header.mjs
```diff
@@ -75,7 +76,7 @@ export class Header extends GOVUKFrontendComponent {
       throw new ElementError({
         componentName: 'Header',
         element: $menu,
-        identifier: `#${menuId}`
+        identifier: `List element #${menuId}`
```
domoscargin, line 79: A more helpful descriptor might be something like "Nav menu" or "Menu element", but I've gone with this in the meantime since this is a `<ul>`

packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
```diff
@@ -120,9 +120,10 @@ export class Accordion extends GOVUKFrontendComponent {
     super()
 
     if (!($module instanceof HTMLElement)) {
-      throw new ElementError('$module', {
+      throw new ElementError({
         componentName: 'Accordion',
-        element: $module
+        element: $module,
+        identifier: 'Root element (`$module`)'
```
colinrotherham, line 126: Can't remember if we said to keep `$module` versus @romaricpascal's HTML preference?

I'm happy with `$module` though as only `initAll()` is guaranteed to use this:

```suggestion
        identifier: 'Root element (`<div data-module="govuk-accordion">`)'
```

packages/govuk-frontend/src/govuk/components/character-count/character-count.mjs
```diff
@@ -140,9 +142,10 @@ export class CharacterCount extends GOVUKFrontendComponent {
     const textareaDescriptionId = `${this.$textarea.id}-info`
     const $textareaDescription = document.getElementById(textareaDescriptionId)
     if (!$textareaDescription) {
-      throw new ElementError(`#${textareaDescriptionId}`, {
+      throw new ElementError({
         componentName: 'Character count',
-        element: $textareaDescription
+        element: $textareaDescription,
+        identifier: `Count message #${textareaDescriptionId}`
```
colinrotherham, line 148:
```suggestion
        identifier: `Count message (\`#${textareaDescriptionId}\`)`
```

packages/govuk-frontend/src/govuk/components/checkboxes/checkboxes.mjs
```diff
@@ -57,8 +59,9 @@ export class Checkboxes extends GOVUKFrontendComponent {
 
       // Throw if target conditional element does not exist.
       if (!document.getElementById(targetId)) {
-        throw new ElementError(`#${targetId}`, {
-          componentName: 'Checkboxes'
+        throw new ElementError({
+          componentName: 'Checkboxes',
+          identifier: `Conditional element #${targetId}`
```
colinrotherham, line 64: I'm torn on "element" creeping back in, but does everyone call these "Conditional reveals"?

```suggestion
          identifier: `Conditional reveal (\`#${targetId}\`)`
```

packages/govuk-frontend/src/govuk/components/checkboxes/checkboxes.mjs
```diff
@@ -31,16 +31,18 @@ export class Checkboxes extends GOVUKFrontendComponent {
...
       })
     }
 
     const $inputs = $module.querySelectorAll('input[type="checkbox"]')
     if (!$inputs.length) {
-      throw new ElementError('input[type="checkbox"]', {
-        componentName: 'Checkboxes'
+      throw new ElementError({
+        componentName: 'Checkboxes',
+        identifier: 'Input elements <input type="checkbox">'
```
colinrotherham, line 45: Just putting this out there, but we say "Form field" when it's of unknown type

Should we say "Form input" when we know it's an input?

```suggestion
        identifier: 'Form input (\`<input type="checkbox">\`)'
```

packages/govuk-frontend/src/govuk/components/header/header.mjs
colinrotherham: From the **Header** guidance we have elements "Service name" and "Navigation"

Shall we use HTML to help here?

```suggestion
        identifier: `Navigation (\`<ul id="${menuId}">\`)`
```

packages/govuk-frontend/src/govuk/components/header/header.mjs
```diff
@@ -63,16 +64,19 @@ export class Header extends GOVUKFrontendComponent {
 
     const menuId = $menuButton.getAttribute('aria-controls')
     if (!menuId) {
-      throw new ElementError('.govuk-js-header-toggle[aria-controls]', {
-        componentName: 'Header'
+      throw new ElementError({
+        componentName: 'Header',
+        identifier:
+          'Button (`.govuk-js-header-toggle`) attribute (`aria-controls`)'
```
colinrotherham, line 70: Do we move as many identifiers as we can to HTML?

```suggestion
          'Navigation button (`<button class="govuk-js-header-toggle">`) attribute (`aria-controls`)'
```

packages/govuk-frontend/src/govuk/components/skip-link/skip-link.mjs
```diff
@@ -51,20 +52,19 @@ export class SkipLink extends GOVUKFrontendComponent {
...
     const $linkedElement = document.getElementById(linkedElementId)
 
     // Check for link target element
     if (!$linkedElement) {
-      throw new ElementError(`$module.hash target #${linkedElementId}`, {
+      throw new ElementError({
         componentName: 'Skip link',
-        element: $linkedElement
+        element: $linkedElement,
+        identifier: `Linked element #${linkedElementId}`
```
colinrotherham, line 67: Knowing that the default is to "skip to the main content" should we give this a tweak?

Keeping to HTML is the `id=""` format helpful?

```suggestion
        identifier: `Target content `id="${linkedElementId}"`
```

packages/govuk-frontend/src/govuk/components/tabs/tabs.mjs
```diff
@@ -51,16 +51,18 @@ export class Tabs extends GOVUKFrontendComponent {
...
       })
     }
 
     const $tabs = $module.querySelectorAll('a.govuk-tabs__tab')
     if (!$tabs.length) {
-      throw new ElementError(`a.govuk-tabs__tab`, {
-        componentName: 'Tabs'
+      throw new ElementError({
+        componentName: 'Tabs',
+        identifier: 'Link elements (`a.govuk-tabs__tab`)'
```
colinrotherham, line 65: Another one for HTML?

```suggestion
        identifier: 'Link (`<a class="govuk-tabs__tab">`)'
```

colinrotherham (commented):
Looking good @domoscargin

My comments are mainly thoughts for discussion really, see what you think

packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
domoscargin: On this one, I was going by the comment here: https://github.com/alphagov/govuk-frontend/issues/4299#issuecomment-1753245288, which points out the same sort of issue with using the HTML.

packages/govuk-frontend/src/govuk/components/character-count/character-count.mjs
domoscargin: 🤦

packages/govuk-frontend/src/govuk/components/checkboxes/checkboxes.mjs
domoscargin: Yeah, let's go for the wider use rather than trying to shimmy everything into a certain format.

packages/govuk-frontend/src/govuk/components/accordion/accordion.mjs
colinrotherham: Great minds, thanks for confirming

packages/govuk-frontend/src/govuk/components/checkboxes/checkboxes.mjs
colinrotherham: Brill. Then there's the HTML suggestion from below too:

```suggestion
          identifier: `Conditional reveal (\`id="${targetId}"\`)`
```

packages/govuk-frontend/src/govuk/components/checkboxes/checkboxes.mjs
colinrotherham: Ah it's plural, so `Form inputs`?

domoscargin:
Thanks @colinrotherham - agreed with everything and have pushed some changes up.

packages/govuk-frontend/src/govuk/components/tabs/tabs.mjs
```diff
@@ -51,16 +51,18 @@ export class Tabs extends GOVUKFrontendComponent {
...
       })
     }
 
     const $tabs = $module.querySelectorAll('a.govuk-tabs__tab')
     if (!$tabs.length) {
-      throw new ElementError(`a.govuk-tabs__tab`, {
-        componentName: 'Tabs'
+      throw new ElementError({
+        componentName: 'Tabs',
+        identifier: 'Link (`<a class="govuk-tabs__tab">`)'
```
colinrotherham, line 65: This is my fault, forgot this one was plural ~`Link`~ `Links` too

packages/govuk-frontend/src/govuk/components/tabs/tabs.mjs
```diff
@@ -78,14 +80,16 @@ export class Tabs extends GOVUKFrontendComponent {
     )
 
     if (!$tabList) {
-      throw new ElementError(`.govuk-tabs__list`, {
-        componentName: 'Tabs'
+      throw new ElementError({
+        componentName: 'Tabs',
+        identifier: 'Tab list element (`.govuk-tabs__list`)'
```
colinrotherham, line 85: HTML all the things?

```suggestion
        identifier: 'Tab list (`<ul class="govuk-tabs__list">`)'
```

packages/govuk-frontend/src/govuk/components/tabs/tabs.mjs
```diff
@@ -78,14 +80,16 @@ export class Tabs extends GOVUKFrontendComponent {
...
       })
     }
 
     if (!$tabListItems.length) {
-      throw new ElementError(`.govuk-tabs__list-item`, {
-        componentName: 'Tabs'
+      throw new ElementError({
+        componentName: 'Tabs',
+        identifier: 'Tab list item elements (`.govuk-tabs__list-item`)'
```
colinrotherham, line 92:
```suggestion
        identifier: 'Tab list items (`<li class="govuk-tabs__list-item">`)'
```

packages/govuk-frontend/src/govuk/components/checkboxes/checkboxes.mjs
```diff
@@ -31,16 +31,18 @@ export class Checkboxes extends GOVUKFrontendComponent {
     super()
 
     if (!($module instanceof HTMLElement)) {
-      throw new ElementError(`[data-module="${Checkboxes.moduleName}"]`, {
+      throw new ElementError({
         componentName: 'Checkboxes',
-        element: $module
+        element: $module,
+        identifier: `Root element (\`$module\`)`
```
colinrotherham, line 37: Not a blocker, and we should configure ESLint to do this, but can avoid the escaping?

```suggestion
        identifier: 'Root element (`$module`)'
```

colinrotherham (approved):
Fab work @domoscargin

Let's bring @romaricpascal in for HTML identifier feedback too

PR #4370, opened 2023-10-23
Base pre-release documentation on release documentation

romaricpascal:
Updates https://github.com/alphagov/govuk-frontend/pull/3977 to makes the pre-release process follow the release one, rather than the preview process, as we're aiming to publish to npm.

It carries over the explanations for picking the right version bump from #3977 and distinguish steps that don't need to be taken if simply publishing an internal pre-release.

### DIFF
--- a/docs/releasing/publishing-a-pre-release.md
+++ b/docs/releasing/publishing-a-pre-release.md
@@ -14,10 +14,7 @@ documentation updates linked to the code, as well as communications, similarly t
 
 Review the docs for [what to do before publishing a release](/docs/releasing/before-publishing-a-release.md) to assess which steps you need to follow for your specific pre-release and ensure you are prepared to publish.
 
-See the [documentation on support branches](https://govuk-design-system-team-docs.netlify.app/how-we-work/version-control/support-branches.html#support-branches) if you need to:
-
-- publish a new pre-release of previous major versions of GOV.UK Frontend
-- publish a ‘hotfix’ release of GOV.UK Frontend without including other unreleased changes on the `main` branch
+See the [documentation on support branches](https://govuk-design-system-team-docs.netlify.app/how-we-work/version-control/support-branches.html#support-branches) if you need to publish a new pre-release of previous major versions of GOV.UK Frontend.
 
 # Publish a new version of GOV.UK Frontend
 
@@ -60,7 +57,7 @@ Developers should pair on pre-releases. When remote working, it can be useful to
    npm version <PRE-RELEASE TYPE> --preid <PRE-RELEASE IDENTIFIER> --no-git-tag-version --workspace govuk-frontend
    ```
 
-   This step will update the [`package.json`](/package.json) and project [`package-lock.json`](/package-lock.json) files.
+   This step will update [`govuk-frontend`'s `package.json`](/packages/govuk-frontend/package.json) and project [`package-lock.json`](/package-lock.json) files.
 
    Do not commit the changes.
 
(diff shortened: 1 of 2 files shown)


docs/releasing/publishing-a-pre-release.md
```diff
@@ -46,20 +46,78 @@ Projects can install `govuk-frontend@beta` in their package.json alongside other
...
+
+10. Run `npm run build-release` to:
+
+    - build GOV.UK Frontend into the [`/package`](/package) and [`/dist`](/dist) directories
+    - commit the changes
+    - push a branch to GitHub
+
+    You will now be prompted to continue or cancel.
+
+11. If you're publishing a beta pre-release, create a pull request
```
romaricpascal, line 80: **note** Figured we don't really need to reverse `internal` pre-releases to `main`, only `beta` ones. Reason for that is that if we alternate `beta` and `internal` pre-releases (for example, to check a bugfix before the next pre-release), that will make the version bump rocky for `beta` release as we'd go from `-beta.X` to `-internal.0` then the next `prerelease` bump would go to `-beta.0` instead of `-beta.X+1`.

docs/releasing/publishing-a-pre-release.md
```diff
@@ -46,20 +46,78 @@ Projects can install `govuk-frontend@beta` in their package.json alongside other
...
+1. If you're publishing a beta pre-release, check out the **main** branch and pull the latest changes.
+
+2. Sign in to npm (`npm login`), using the credentials for the govuk-patterns-and-tools npm user from Bitwarden.
+
+3. Run `npm run publish-release`, which will prompt you to check whether the npm tag looks as expected.
+
+4. Enter `N` to continue to set the npm tag corresponding to the kind of release you're publishing:
+
+   - `internal` for internal pre-releases
+   - `next` for beta pre-releases
```
romaricpascal, line 96: **suggestion** Followed a convention that seemed followed by a good few packages publishing their pre-releases under the `@next` tag. Picking `next` instead of `beta` would allow to extract the tag directly from the version number 🤔

docs/releasing/publishing-a-pre-release.md
```diff
@@ -46,20 +46,78 @@ Projects can install `govuk-frontend@beta` in their package.json alongside other
...
+
+12. If you're publishing a beta pre-release, once a reviewer approves the pull request, merge it to **main**.
+
+## Publish a release to npm
+
+1. If you're publishing a beta pre-release, check out the **main** branch and pull the latest changes.
+
+2. Sign in to npm (`npm login`), using the credentials for the govuk-patterns-and-tools npm user from Bitwarden.
+
+3. Run `npm run publish-release`, which will prompt you to check whether the npm tag looks as expected.
```
romaricpascal, line 91: **note** The script generates a default tag based on the package version thanks to `bin/generate-npm-tag.sh`. Pre-release version numbers don't break the script (yay!), however, the script could likely use some upgrading later on to create tags for pre-releases automatically.

docs/releasing/publishing-a-pre-release.md
```diff
@@ -46,20 +43,83 @@ Projects can install `govuk-frontend@beta` in their package.json alongside other
...
+5. You will now be prompted to continue or cancel the release. Check the details and enter `y` to continue. If something does not look right, press `N` to cancel the release.
+
+   This step will create a ZIP file containing the release in the root of your govuk-frontend git directory. You will need this file when creating the GitHub release.
+
+   It will also automatically create a tag in Github which you can use to create a Github release in the following section.
+
+6. Verify the presence of the pre-release and its tag on [npm](https://www.npmjs.com/package/govuk-frontend?activeTab=versions)
+
+   If the pre-release has been assigned the wrong tag (mistakes happen),
+   you can use [`npm dist-tag`](https://docs.npmjs.com/cli/v8/commands/npm-dist-tag) to quickly correct.
```
romaricpascal, line 107: **note** Figured a little final check wouldn't hurt. Also, as suggested by Ollie, dug how we can correct any mistakes that may have happened in tagging the package, while still logged in to npm. Also updated `publishing.md`.

PR #4373, opened 2023-10-24
🚀 Pre-release v5.0.0-internal.0

36degrees:


colinrotherham (approved):
Approved 🎉

Not given as much scrutiny as a full release since the preview build was used quite widely already, including the Design System Day workshop: "Using task lists in your services"

Any snags or bug fixes can always go in `5.0.0-internal.1`

PR #4377, opened 2023-10-24
Add Dependabot `caniuse-lite` updates

colinrotherham:
Browser data from ["Can I use"](https://caniuse.com) is used by Autoprefixer, Babel and CSSNANO but it's not updated automatically

This PR configures Dependabot to update the [`caniuse-lite`](https://www.npmjs.com/package/caniuse-lite) nested **package-lock.json** dependency

We currently only allow [direct **package.json** dependencies](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#allow)

### Release instructions

By checking for [`caniuse-lite`](https://www.npmjs.com/package/caniuse-lite) updates using Dependabot weekly checks, we can review [potentially breaking changes](https://github.com/alphagov/govuk-design-system/pull/3051#issuecomment-1680267867) before merging, yet also remove the update step from our release instructions:

>   ```shell
>   npx update-browserslist-db@latest
>   ```
>
>   This step will update the project [`package-lock.json`](/package-lock.json) file if updates are found.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -79,12 +79,6 @@ updates:
     allow:
       - dependency-type: direct
 
-    ignore:
-      # Ignore major updates (causes `stylelint-config-gds` peer dependency conflict)
-      # See pull request: https://github.com/alphagov/stylelint-config-gds/pull/32
-      - dependency-name: stylelint
-        update-types: ['version-update:semver-major']
-
   # Update GitHub Actions
   - package-ecosystem: github-actions
     directory: /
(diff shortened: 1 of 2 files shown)


.github/dependabot.yml
```diff
@@ -79,12 +79,6 @@ updates:
     allow:
       - dependency-type: direct
 
-    ignore:
```
domoscargin, line 82: 👍

domoscargin:
What about the case where a breaking update gets released on a Wednesday and we released on a Thursday? Do we need some kind of in the documentation or can we just rely on the log output to clue us in to what to do?

domoscargin:
Yup, sounds good to me!

PR #4391, opened 2023-10-27
Move private workspace packages to `devDependencies`

colinrotherham:
Unlike npm, yarn is not happy unless `optionalDependencies` are published

Addresses user feedback from https://github.com/alphagov/govuk-frontend/discussions/4389#discussioncomment-7401425


romaricpascal (approved):
Neat! ⛵

PR #4392, opened 2023-10-27
Add CHANGELOG entry about `<script>` snippet

romaricpascal:
The snippet setting the `govuk-frontend-supported` class is described in the Frontend Docs, but not the release notes.

Addresses [feedback from #4389](https://github.com/alphagov/govuk-frontend/discussions/4389#discussioncomment-7396486)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -204,6 +204,36 @@ These changes were made in the following pull requests:
 - [#3491: Update Review app to import `govuk-frontend` via local package](https://github.com/alphagov/govuk-frontend/pull/3491)
 - [#3498: Remove built `dist` and `package` from source](https://github.com/alphagov/govuk-frontend/pull/3498)
 
+#### Update the `<script>` snippet at the top of your `<body>` tag
+
+Page templates now include a new `govuk-frontend-supported` class on the `<body>` tag when GOV.UK Frontend JavaScript components are fully supported.
+
+If you are not using our [Nunjucks page template](https://design-system.service.gov.uk/styles/page-template/), replace the existing snippet:
+
+```html
+<script>document.body.className = ((document.body.className) ? document.body.className + ' js-enabled' : 'js-enabled');</script>
+```
+
+with:
+
+```html
+<script>document.body.className += ' js-enabled' + ('noModule' in HTMLScriptElement.prototype ? ' govuk-frontend-supported' : '');</script>
+```
+
+#### Update the hash used by your Content Security Policy
+
+If your Content Security Policy uses a hash to let the snippet above run, you'll need to update it from:
+
+```
+sha256-+6WnXIl4mbFTCARd8N3COQmT3bJJmo32N8q8ZSQAIcU=
+```
+
+to:
+
+```
+sha256-GUQ5ad8JK5KmEWmROf3LZd9ge94daqNvd8xy9YS1iDw=
+```
+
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -204,6 +204,22 @@ These changes were made in the following pull requests:
 - [#3491: Update Review app to import `govuk-frontend` via local package](https://github.com/alphagov/govuk-frontend/pull/3491)
 - [#3498: Remove built `dist` and `package` from source](https://github.com/alphagov/govuk-frontend/pull/3498)
 
+#### Update the `<script>` snippet at the top of your `<body>` tag
+
+We've introduced a new `govuk-frontend-supported` class marking that GOV.UK Frontend is supported in the browser showing the page. It complements the existing `js-enabled` class and is necessary for components to initialise.
```
claireashworth, line 209: is this - 
We've introduced a new `govuk-frontend-supported` class, marking that...
or
We've introduced a new `govuk-frontend-supported` class marking, showing that...
?
And I'm not sure what you mean by 'showing the page' in this context

romaricpascal: It's the first one. Guess it's a sign that I should split it into two sentences, maybe?

Regarding "showing the page", I was trying to clarify "the browser" by adding that it's the one currently rendering the page (as "GOV.UK Frontend is supported in the browser" sounded a bit of a short end).

colinrotherham: We'll need to put this on the migration page too?
https://release-5-0--govuk-frontend-docs-preview.netlify.app/changes-to-govuk-frontend-v5/

I'd probably state that we handle this change automatically too

Only users that implement their own page templates need to take action

```suggestion
Page templates now include a new `govuk-frontend-supported` class on the `<body>` tag when GOV.UK Frontend JavaScript components are fully supported.

If you are not using our [Nunjucks page template](https://design-system.service.gov.uk/styles/page-template/), replace the existing snippet:
```

colinrotherham: Very important to note that good CSP (Content Security Policy) will block inline scripts so there will very likely be a [hash to update](https://frontend.design-system.service.gov.uk/importing-css-assets-and-javascript/#use-a-hash-to-unblock-inline-javascript) in production services:

```patch
- sha256-+6WnXIl4mbFTCARd8N3COQmT3bJJmo32N8q8ZSQAIcU=
- sha256-GUQ5ad8JK5KmEWmROf3LZd9ge94daqNvd8xy9YS1iDw=
```

claireashworth: Probably yes to the migration page. And the suggested change looks good to me as well.

romaricpascal: Updated and added another heading for the CSP :)

Agree it'll be good to mention on the migration page as well, good shout.

claireashworth (approved):

CHANGELOG.md
colinrotherham: @romaricpascal Mind including both a "from" and a "to" for the CSP hash?

It's far easier for users to search their code for the string to replace

colinrotherham (approved):

romaricpascal:
Added the content to [the release notes](https://github.com/alphagov/govuk-frontend/releases/tag/v5.0.0-beta.0) and [the discussion](https://github.com/alphagov/govuk-frontend/discussions/4389).

Issue #4393, opened 2023-10-27
Remove table formatting from the Graded Browser Support document in docs>contributing


claireashworth:
## What
Change the table formatting in the graded browser support (full version) in docs>contributing to bullet lists for each grade.
## Why
The table did not work well in a mobile format and made the information harder to read.
## Who needs to work on this
Tech writer
## Who needs to review this
Developer
## Done when

- [ ] Updates are added to the document

claireashworth:
This is linked to PR #4395

PR #4395, opened 2023-10-27
Update browser-support.md to remove table format

claireashworth:
Replace the table format in browser-support.md with bullet points. This is to correct how the information looks in mobile formats.
This closes issue #4393


colinrotherham (approved):
Thanks @claireashworth, looking good

Did a [minor tweak for Prettier](https://github.com/alphagov/govuk-frontend/compare/3a4bfedc83b59dcc17993f37def36389e5e10446..8979b41b88a13cd8c7529b96b6dd25f27d68f6f9) but checks should all pass now

PR #4415, opened 2023-10-31
Remove custom attribute on status in task list example

frankieroberto:
Custom attributes aren't currently supported on the status within the task list, only for the overall task list itself.

We could chose to support custom attributes on other elements within the task list (such as the status, title and hint) – but I'm not sure what the convention is for this across govuk-frontend?

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/task-list/task-list.yaml
+++ b/packages/govuk-frontend/src/govuk/components/task-list/task-list.yaml
@@ -324,8 +324,6 @@ examples:
           status:
             tag:
               text: Status
-              attributes:
-                data-tag-attribute: tag-value
 
   - name: custom id prefix
     hidden: true


frankieroberto:
@36degrees spotted this when working on the ruby implementation of the task list. Any thoughts on if/how custom attributes should be supported on lower-level elements within components?

colinrotherham:
Ah that's odd, Is it only the Ruby implementation that's affected?

>Custom attributes aren't currently supported on the status within the task list…

I'm seeing the attributes being rendered here:
http://govuk-frontend-review.herokuapp.com/components/task-list/custom-attributes/preview

colinrotherham:
Forgot to add, it came up during code review to [pass everything into `govukTag(item.status.tag)`](https://github.com/alphagov/govuk-frontend/blob/33027ebf89a5168c6d9312e3b6df3f49ba1773f4/packages/govuk-frontend/src/govuk/components/task-list/template.njk#L23)

* https://github.com/alphagov/govuk-frontend/pull/3951#discussion_r1260926518

### Custom attributes

For comparison, form elements and hints always add the exact params including `attributes`

```njk
{{ govukHint({
  id: hintId,
  classes: params.hint.classes,
  attributes: params.hint.attributes,
  html: params.hint.html,
  text: params.hint.text
}) | indent(2) | trim }}
```

But the [phase banner `govukTag()` only supports `text`, `html` and `classes`](https://github.com/alphagov/govuk-frontend/blob/33027ebf89a5168c6d9312e3b6df3f49ba1773f4/packages/govuk-frontend/src/govuk/components/phase-banner/template.njk#L6-L10) not `attributes`

frankieroberto:
@colinrotherham:
> Ah that's odd, Is it only the Ruby implementation that's affected?

No, this is about the Nunjucks implementation.

> > Custom attributes aren't currently supported on the status within the task list…
> 
> I'm seeing the attributes being rendered here: http://govuk-frontend-review.herokuapp.com/components/task-list/custom-attributes/preview

That's only on overall list container, not the item status?

If we want to support custom attributes (as well as custom classes) on some or all of the item, status or title, then we'll have to update the Nunjucks macro, I think, and add some extra tests to cover it.

I think generally across the Design System, it's not super consistent whether sub-components will accept attributes in the nunjucks macros or not. I guess they’ve only been added where there’s been call for it?

The Summary List seems to accept attributes on the `items`, `card` and `item`, but not on `rows`, `action.items`, `title` or `actions` (all of which accept `classes`). 🤷

frankieroberto:
Opps I made a mistake - the attribute is being added to the tag, not the status, and there’s an existing test for it! 🤦   I somehow misread that completely.

colinrotherham:
Ah sorry @frankieroberto

Good points though regarding inconsistencies in passing params to nested components

Probably another reason a Nunjucks "merge objects" filter or global would be handy

* https://github.com/alphagov/govuk-frontend/issues/995#issue-359364686
* https://github.com/alphagov/govuk-frontend/pull/1172#issuecomment-460257465

PR #4417, opened 2023-11-01
Add HTML validation rules for missing elements

colinrotherham:
This PR enables extra HTML validation rules:

1. [`input-missing-label`](https://html-validate.org/rules/input-missing-label.html) – Require input to have a label
2. [`no-missing-references`](https://html-validate.org/rules/no-missing-references.html) – Require all form field and ARIA references to exist

But also adds the HTML validation preset [`html-validate:document`](https://html-validate.org/rules/presets.html#html-validate-document) for testing **govuk/template.njk**

The good news is we’re fully valid already, but we’d be extra safe :sweat_smile:

## Missing elements

To resolve missing elements, we could use ~`render()`~ `renderPreview()` to render the entire page

But instead I've taken the `id="content"` wrapper like we did for **Skip link** accessibility tests in https://github.com/alphagov/govuk-frontend/pull/4346 and updated it with @36degrees's idea to use a dummy element instead:

```patch
  const html = outdent`
    ${render(componentName, {
      context: fixture.options,
      fixture
    })}
+
+   <!--
+     Target for references in examples (e.g. aria-controls)
+     https://html-validate.org/rules/no-missing-references.html
+   -->
+   <div id="test-target-element"></div>
  `
```

This passes validation for examples that add `aria-describedby` etc such as in **fieldset.yaml**

```patch
  - name: with describedBy
    hidden: true
    options:
-     describedBy: content
+     describedBy: test-target-element
      legend:
        text: Which option?
```

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/components.template.test.js
+++ b/packages/govuk-frontend/src/govuk/components/components.template.test.js
@@ -9,6 +9,7 @@ const {
 } = require('@govuk-frontend/lib/components')
 const validatorConfig = require('govuk-frontend/.htmlvalidate.js')
 const { HtmlValidate } = require('html-validate')
+const { outdent } = require('outdent')
 
 describe('Components', () => {
   let nunjucksEnvCustom
@@ -65,10 +66,18 @@ describe('Components', () => {
       // Validate component examples
       for (const { component: componentName, fixtures } of componentsFixtures) {
         const fixtureTasks = fixtures.map(async (fixture) => {
-          const html = render(componentName, {
-            context: fixture.options,
-            fixture
-          })
+          const html = outdent`
+            ${render(componentName, {
+              context: fixture.options,
+              fixture
+            })}
+
+            <!--
+              Target for references in examples (e.g. aria-controls)
+              https://html-validate.org/rules/no-missing-references.html
+            -->
+            <div id="test-target-element"></div>
+          `
 
           // Validate HTML
           return expect({
--- a/packages/govuk-frontend/.htmlvalidate.js
+++ b/packages/govuk-frontend/.htmlvalidate.js
@@ -15,8 +15,14 @@ module.exports = {
     // Allow pattern attribute on input type="number"
     'input-attributes': 'off',
 
+    // Require input to have a label
+    'input-missing-label': 'error',
+
     // Allow inline styles for testing purposes
     'no-inline-style': 'off',
+
+    // Require all form field and ARIA references to exist
+    'no-missing-references': 'error',
 
     // Allow for explicit roles on regions that have implict roles
     // We do this to better support AT with older versions of IE that
--- a/packages/govuk-frontend/src/govuk/components/button/button.yaml
+++ b/packages/govuk-frontend/src/govuk/components/button/button.yaml
@@ -186,7 +186,7 @@ examples:
       element: button
       text: Submit
       attributes:
-        aria-controls: content
+        aria-controls: test-target-element
         data-tracking-dimension: 123
   - name: link attributes
     hidden: true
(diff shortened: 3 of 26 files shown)


packages/govuk-frontend/src/govuk/components/components.template.test.js
```diff
@@ -65,10 +66,14 @@ describe('Components', () => {
...
-            fixture
-          })
+          const html = outdent`
+            <div id="content" class="govuk-width-container">
+              ${render(componentName, {
+                context: fixture.options,
+                fixture
+              })}
+            </div>
+          `
```
36degrees, line 76: I think this could be slightly misleading because this looks very much like one of our main wrapper elements, which I don't think is something you'd ever associate with a component using `aria-describedby`.

Can we make it clearer what this is for by:

- adding the extra element _after_ the rendered component
- changing the ID to something like `test-target-element` (and in the corresponding examples)
- adding a comment to explain what it's for

```suggestion

          const html = outdent`
            ${render(componentName, {
              context: fixture.options,
              fixture
            })}
            <div id="test-target-element">
              <!-- Target for references in examples (e.g. aria-describedby) so no-missing-references can pass -->
            </div>
          `
```

colinrotherham: @36degrees Mind if I tag @romaricpascal in?

I thought the same as you but we agreed it was acceptable for the Skip link accessibility tests

I'd like to make your suggestion there too

36degrees: Not sure I follow, sorry – how is the skip link involved here?

colinrotherham: Ah I'll explain, it was covered in dev catch-up after your holiday

E.g. **Skip link** failed Accessibility tests (JavaScript error) without a target element

So we decided that a dummy element ID was acceptable for now, but due to hard-coded defaults like `default("#content", true)` we [added `id="content"` to the component preview wrapper](https://github.com/alphagov/govuk-frontend/commit/3ed9d1f418ded73ae4e76cc88835b5f1f258db72) just like in this PR

I like your suggestion so perhaps we can annotate the test ID in both wrappers?

colinrotherham: Here's the existing component preview wrapper:

https://github.com/alphagov/govuk-frontend/blob/3ed9d1f418ded73ae4e76cc88835b5f1f258db72/shared/lib/components.js#L209-L211

36degrees: Thanks, that's helpful context 👍🏻 

I think annotating the test ID in both wrappers sounds like a good solution.

colinrotherham: Morning 👋

I've swapped all the `#content` references in 3ed9d1f418ded73ae4e76cc88835b5f1f258db72 to `test-target-element`

Whilst the HTML validation tests lack the `<div id="content">` wrapper, it's still needed by:

```console
npx jest --selectProjects "Accessibility tests"
npx jest --selectProjects "JavaScript component tests"
```

But that's only because it's hard coded in [`href="{{ params.href | default("#content", true) }}"`](https://github.com/alphagov/govuk-frontend/blob/545fde1dda58d2b502d3a67618ebd0dc9567ee51/packages/govuk-frontend/src/govuk/components/skip-link/template.njk#L1)

Without it we'll see JavaScript warnings for every Skip link example:

```console
  console.warn
    Initialising `new SkipLink()` with example 'no href' threw:
     ElementError: Skip link: Target content (`id="content"`) not found
```

Think we're looking good now, ready for review again 👍

36degrees (approved):
Looks great, thanks for making that change 👍🏻 Think it's a lot clearer what's going on now.

PR #4423, opened 2023-11-03
[SPIKE] Tudor crown Safari position tweaks

colinrotherham:
Spike for https://github.com/alphagov/govuk-frontend/pull/4297 to fix vertical alignment (Header product name) in Safari https://github.com/alphagov/govuk-frontend/pull/4297#pullrequestreview-1704517206

This PR aligns Chrome and Safari again

* **Current:** [header/full-width-with-navigation/preview](https://govuk-frontend-review.herokuapp.com/components/header/full-width-with-navigation/preview)
* **Before:** [header/full-width-with-navigation/preview](https://govuk-frontend-pr-4297.herokuapp.com/components/header/full-width-with-navigation/preview) from [#4297](https://github.com/alphagov/govuk-frontend/pull/4297)
* **After:** [header/full-width-with-navigation/preview](https://govuk-frontend-pr-4423.herokuapp.com/components/header/full-width-with-navigation/preview) from [#4423](https://github.com/alphagov/govuk-frontend/pull/4423)

## Original GOV.UK text

Branch `main` renders the GOV.UK text and product name 1px too low in Firefox
https://govuk-frontend-review.herokuapp.com/components/header/full-width-with-navigation/preview

### Using DPI `@1x`

Original 1x

### Using DPI `@2x`

**Note:** We see Safari and Chrome baseline differences using `@2x` versus `@1x`

Original 2x

## GOV.UK text as SVG

PR https://github.com/alphagov/govuk-frontend/pull/4297 renders the product name 1px too high in Chrome and Safari:
https://govuk-frontend-pr-4297.herokuapp.com/components/header/full-width-with-navigation/preview

### Using DPI `@1x`

Before 1x

### Using DPI `@2x`

**Note:** We see Safari and Chrome baseline differences using `@2x` versus `@1x` but the SVG text is unaffected

Before 2x

## GOV.UK text as SVG with position tweaks

PR https://github.com/alphagov/govuk-frontend/pull/4423 (this spike) is correctly aligned:
https://govuk-frontend-pr-4423.herokuapp.com/components/header/full-width-with-navigation/preview

After

There's also a workaround included for Firefox where the text baseline is calculated 1px lower (or 0.5px at `@2x` DPI) than other browsers, breaking alignment with the product name. This could be "fixed" in Firefox with:

```patch
  // Maintain space below logo when wrapped
  margin-top: $product-name-offset;
+ @-moz-document url-prefix() {
+   margin-top: $product-name-offset - 0.5px;
+ }
```

Firefox also appeared to have issues with some HiDPI resolutions where the top and bottom parts of the Crown (and the GOV.UK text) could subtly move apart as they round to pixel values but this was fixed in https://github.com/alphagov/govuk-frontend/pull/4297/commits/fe6f75b30e0aef9efa53bbbeb233eb3c6cc8005c

Related Firefox issues regarding SVG rendering accuracy:

* https://bugzilla.mozilla.org/show_bug.cgi?id=1395350
* https://bugzilla.mozilla.org/show_bug.cgi?id=1674681


packages/govuk-frontend/src/govuk/components/header/_index.scss
```diff
@@ -98,7 +113,7 @@
     font-size: 30px; // We don't have a mixin that produces 30px font size
     line-height: 1;
 
-    @include govuk-media-query($from: tablet) {
+    @include govuk-media-query($from: desktop) {
```
colinrotherham, line 116: @owenatgov Having an parent link with `display: inline` didn't help child alignment, but think it's a bug from https://github.com/alphagov/govuk-frontend/pull/2549 where `display: inline` was added at tablet size but the product name didn't wrap until desktop

querkmachine:
This seems to have done the job in desktop Safari, though—just to add some more pixelpushing fun to the equation—it now appears to be slightly misaligned in iOS Safari where it wasn't before. Only by a pixel, so might need to click through to the full size image. 

Composite screenshot of three Safari on iPad screens, showing the header as it appears on the main branch, in PR 4297 and PR 4423. The two PRs use a different crown. The logo in PR 4423 is slightly shifted up compared to the other two.

Outside of that, I think this could use some code comments explaining what's going on. Why is padding and alignment being changed between mobile and tablet? Why does it become inline on desktop? Why is a Firefox 'hack' needed and why does a fractional pixel value fix it?

More generally, and not specific to this PR, was the product name always positioned like this regardless of viewport size? I'm not sure it was like this in v4? 

Screenshot of the GOV.UK header with a product name, which is rendering below the logo across all screen sizes.

colinrotherham:
@querkmachine Ah that's fair, I only checked tablet and desktop

Hopefully enough of a spike to demonstrate what's possible, or that the SVG points needs snapping to pixels otherwise Firefox is going to go ahead and do it anyway 😣

See https://github.com/alphagov/govuk-frontend/pull/4423#discussion_r1381958130 for "Why does it become inline on desktop?"

Do you want me to hand back over to you or shall I answer the other questions?

querkmachine:
@colinrotherham I've updated #4297 with a tweaked SVG that snaps some of those errant .005px coordinates to their full number. 

At a glance, the lockup appears to be rendering in the right position on Firefox and iOS Safari and 1px lower than the product name in Chrome and macOS Safari. Can you corroborate that that's what you see too?

colinrotherham:
Sorry for the delay @querkmachine

Yeah, for the GOV.UK part it's exactly as you said, 1px too low in Chrome and macOS Safari now. Taking just the **Product Name** text though, it's perfectly v-aligned in Chrome, Firefox and Safari

I think we should go back to it just being Firefox misaligned and apply the CSS hack

colinrotherham:
@querkmachine Mind taking another look?

The new SVGs have definitely helped as has using a single compound path

querkmachine:
I've now merged the code changes from this spike into the branches for v5.0 (#4449) and v5.x (#4297). If everything in those branches is looking good, this spike can be shut down. 

Thanks for spending the time to look at this!

PR #4424, opened 2023-11-03
Make review app page titles meaningful

owenatgov:
## What
Gives our examples meaningful page titles beyond just "GOV.UK Frontend". I've gone with the following format:

- Component pages: {{ Component name }} - GOV.UK Frontend
- Component example previews: {{ Component name }}, {{ Example name }} - GOV.UK Frontend
- Non-full page example pages: {{ Example name }} - GOV.UK Frontend
- All components: All components - GOV.UK Frontend
- List of full page examples: Full page examples - GOV.UK Frontend

The homepage is still just "GOV.UK Fronted".

## Why
It's an accessibility best practice to have pages with different meaningful titles across a website. This in theory isn't as critical for us since the review app is mostly internal, however it also has a general productivity benefit of enabling folks working with the review app to clearly see differences between tabs when working across multiple examples.

To illustrate this, here's what main looks like now if you are working across multiple different pages:
List of browser tabs that all have the page title GOV.UK Frontend

 Here's what those same tabs look like with this change:
List of browser tabs with different page titles

### DIFF
--- a/packages/govuk-frontend-review/src/app.mjs
+++ b/packages/govuk-frontend-review/src/app.mjs
@@ -185,6 +185,7 @@ export default async () => {
 
   // Example view
   app.get('/examples/:exampleName', function (req, res, next) {
+    res.locals.exampleName = req.params.exampleName
     // Passing a random number used for the links so that they will be unique and not display as "visited"
     const randomPageHash = (Math.random() * 1000000).toFixed()
     res.render(
--- a/packages/govuk-frontend-review/src/app.mjs
+++ b/packages/govuk-frontend-review/src/app.mjs
@@ -143,14 +143,15 @@ export default async () => {
     '/components/:componentName/:exampleName?/preview',
     function (req, res, next) {
       // Find the data for the specified example (or the default example)
-      const componentName = req.params.componentName
-      const exampleName = req.params.exampleName || 'default'
+      res.locals.componentName = req.params.componentName
+      res.locals.exampleName = req.params.exampleName || 'default'
 
       /** @type {ComponentFixtures | undefined} */
       const componentFixtures = res.locals.componentFixtures
 
       const fixture = componentFixtures?.fixtures.find(
-        (fixture) => nunjucks.filters.slugify(fixture.name) === exampleName
+        (fixture) =>
+          nunjucks.filters.slugify(fixture.name) === res.locals.exampleName
       )
 
       if (!fixture) {
@@ -158,7 +159,7 @@ export default async () => {
       }
 
       // Construct and evaluate the component with the data for this example
-      res.locals.componentView = render(componentName, {
+      res.locals.componentView = render(res.locals.componentName, {
         context: fixture.options,
         env,
         fixture
--- a/packages/govuk-frontend-review/src/views/component-preview.njk
+++ b/packages/govuk-frontend-review/src/views/component-preview.njk
@@ -1,5 +1,9 @@
 {% extends "layouts/" + previewLayout | default('layout') + ".njk" %}
 
+{% set templateTitle = ("Default " + componentName if exampleName == "default" else componentName + " " + exampleName) | unslugify %}
+
+{% block pageTitle %}{{ templateTitle }} - GOV.UK Frontend{% endblock %}
+
 {% set bodyClasses %}
   {{ bodyClasses }}
 {% endset %}
--- a/packages/govuk-frontend-review/src/views/all-components.njk
+++ b/packages/govuk-frontend-review/src/views/all-components.njk
@@ -3,6 +3,8 @@
 {% from "govuk/components/back-link/macro.njk" import govukBackLink %}
 {% from "macros/showExamples.njk" import showExamples %}
 
+{% block pageTitle %}All components - GOV.UK Frontend{% endblock %}
+
 {% set bodyClasses %}
   language-markup
 {% endset %}
--- a/packages/govuk-frontend-review/src/views/component.njk
+++ b/packages/govuk-frontend-review/src/views/component.njk
@@ -3,6 +3,8 @@
 
 {% extends "layouts/full-width-landmarks.njk" %}
 
+{% block pageTitle %}{{ componentName | unslugify }} - GOV.UK Frontend{% endblock %}
+
 {% set bodyClasses %}
   language-markup
 {% endset %}
--- a/packages/govuk-frontend-review/src/views/examples/button-groups/index.njk
+++ b/packages/govuk-frontend-review/src/views/examples/button-groups/index.njk
@@ -1,4 +1,4 @@
-{% extends "layouts/layout.njk" %}
+{% extends "layouts/examples.njk" %}
 
 {% from "govuk/components/back-link/macro.njk" import govukBackLink %}
 {% from "govuk/components/button/macro.njk" import govukButton %}
--- a/packages/govuk-frontend-review/src/views/examples/conditional-reveals/index.njk
+++ b/packages/govuk-frontend-review/src/views/examples/conditional-reveals/index.njk
@@ -1,4 +1,4 @@
-{% extends "layouts/layout.njk" %}
+{% extends "layouts/examples.njk" %}
 
 {% from "govuk/components/back-link/macro.njk" import govukBackLink %}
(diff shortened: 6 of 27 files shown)


packages/govuk-frontend-review/src/views/examples/template-default/index.njk
```diff
@@ -1,4 +1,4 @@
-{% extends "govuk/template.njk" %}
+{% extends "layouts/examples.njk" %}
```
colinrotherham, line 1: Looking good this @owenatgov 🙌

Probably seen from the tests the [template-default](https://govuk-frontend-pr-4424.herokuapp.com/examples/template-default) example has broken?

https://govuk-frontend-pr-4424.herokuapp.com/examples/template-default

owenatgov: Ahaaaaaa that's where my test are going. Cheers Colin I'll sus this out.

colinrotherham: Thanks @owenatgov, let me know if you need a hand with anything

packages/govuk-frontend-review/src/views/component-preview.njk
```diff
@@ -1,5 +1,11 @@
 {% extends "layouts/" + previewLayout | default('layout') + ".njk" %}
 
+{% block pageTitle %}
+  {{- componentName | unslugify -}}
+  {{ ", " + exampleName | unslugify | lower if exampleName }}
+  - GOV.UK Frontend
+{%- endblock %}
+
```
colinrotherham, line 8: Not looked too closely but might have a couple of minor conflicts with https://github.com/alphagov/govuk-frontend/pull/4416 in this PR

Similar to https://github.com/alphagov/govuk-frontend/pull/4416/commits/c7d87769e24e725a820b00b40407867cadab37c8 probably watch out for new lines creeping in here?

```suggestion
{% set pageTitle = (componentName + ", " + exampleName | lower if exampleName else componentName) | unslugify %}
{% block pageTitle %}{{ pageTitle }} - GOV.UK Frontend{% endblock %}
```

packages/govuk-frontend-review/src/app.mjs
```diff
@@ -143,22 +143,23 @@ export default async () => {
     '/components/:componentName/:exampleName?/preview',
     function (req, res, next) {
       // Find the data for the specified example (or the default example)
-      const componentName = req.params.componentName
-      const exampleName = req.params.exampleName || 'default'
+      res.locals.componentName = req.params.componentName
+      res.locals.exampleName = req.params.exampleName || 'default'
```
colinrotherham, line 147: @owenatgov Not a blocker, but just a reminder for us to revisit this

So we're not assigning these in multiple places, I'll do a follow up PR to move parameter parsing into `app.param()` then other routes like `'/full-page-examples/:exampleName'` in **full-page-examples.mjs** pick it up

#### Handler for `app.param('componentName')`

```mjs
app.param('componentName', function (req, res, next, componentName) {
  const exampleName = 'default'

  // Find all fixtures for component
  const componentFixtures = componentsFixtures.find(
    ({ component }) => component === componentName
  )

  // Find default fixture for component
  const componentFixture = componentFixtures?.fixtures.find(
    ({ name }) => name === exampleName
  )

  // Add response locals
  res.locals.componentName = componentName
  res.locals.componentFixtures = componentFixtures
  res.locals.componentFixture = componentFixture
  res.locals.exampleName = 'default'

  next()
})
```

#### Handler for `app.param('exampleName')`

```mjs
app.param('exampleName', function (req, res, next, exampleName) {
  const { componentFixtures } = res.locals

  // Replace default fixture with named example
  const componentFixture = componentFixtures?.fixtures.find(
    ({ name }) => nunjucks.filters.slugify(name) === exampleName
  )

  // Update response locals
  res.locals.componentFixture = componentFixture
  res.locals.exampleName = exampleName

  next()
})
```

packages/govuk-frontend-review/src/views/examples/template-custom/index.njk
```diff
@@ -6,7 +6,7 @@
 {% from "govuk/components/phase-banner/macro.njk" import govukPhaseBanner %}
 {% from "govuk/components/footer/macro.njk" import govukFooter %}
 
-{% extends "govuk/template.njk" %}
+{% extends "layouts/examples-govuk-template.njk" %}
```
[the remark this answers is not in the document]
owenatgov: I just realised that `template-custom` already has a specified `pageTitle` because that's what the example is testing 🤦🏻 I'm going to leave both `template-custom` and `template-default` out of this change as both of them are individually intended to test defaults and customisation coming from our govuk template.

packages/govuk-frontend-review/src/views/component-preview.njk
```diff
@@ -1,5 +1,9 @@
 {% extends "layouts/" + previewLayout | default('layout') + ".njk" %}
 
+{% set templateTitle = (componentName + ", " + exampleName | lower if exampleName else componentName) | unslugify %}
```
colinrotherham, line 3: Quick one, but `exampleName` falls back to `|| 'default'` so `if exampleName` will always be truthy?

colinrotherham: Yeah, just checked. Nearly there though!

```
<title>Character count, default - GOV.UK Frontend</title>
```

CharlotteDowns:
I love this!!

packages/govuk-frontend-review/src/views/component-preview.njk
owenatgov: Good idea! 2 tics...

packages/govuk-frontend-review/src/views/component-preview.njk
```diff
@@ -1,5 +1,9 @@
 {% extends "layouts/" + previewLayout | default('layout') + ".njk" %}
 
+{% set templateTitle = (componentName + " " + exampleName | lower) | unslugify %}
```
colinrotherham, line 3: Ah did you definitely want "Accordion default" rather than "Accordion"?

```suggestion
{% set templateTitle = (componentName + (" " + exampleName if exampleName != "default")) | unslugify %}
```

We also don't have `| lower` on the example iframe headings, did you definitely want it?

https://github.com/alphagov/govuk-frontend/blob/545fde1dda58d2b502d3a67618ebd0dc9567ee51/packages/govuk-frontend-review/src/views/macros/showExamples.njk#L10-L16

owenatgov: So on the first comment: I'd ideally like some identifier other than the component name because then you've got 2 things with the same title:

1. root component page: {{Component}} - GOV.UK Frontend
2. default example page: {{Component}} - GOV.UK Frontend

Maybe if we checked if it was "default" and then put it at the front instead? So "Default accordion" for the first example and "Accordion blah blah blah" for everything else. What do you think?

Re: the `lower` thing, I need to check again and we no longer need it! I'll remove.

colinrotherham: You choose, just think it'll be good to match [showExamples.njk](https://github.com/alphagov/govuk-frontend/blob/545fde1dda58d2b502d3a67618ebd0dc9567ee51/packages/govuk-frontend-review/src/views/macros/showExamples.njk#L10-L16) 😊

owenatgov: I've implemented it! Let me know what you think. Maybe there's a neater way to format it.

packages/govuk-frontend-review/src/views/component-preview.njk
```diff
@@ -1,5 +1,9 @@
 {% extends "layouts/" + previewLayout | default('layout') + ".njk" %}
 
+{% set templateTitle = ("Default " + componentName if exampleName == "default" else componentName + " " +exampleName) | unslugify %}
```
colinrotherham, line 3: If you're happy, I'm happy

Just a tiny missing space and let's merge

```suggestion
{% set templateTitle = ("Default " + componentName if exampleName == "default" else componentName + " " + exampleName) | unslugify %}
```

colinrotherham (approved):

PR #4434, opened 2023-11-06
Remove X-UA-Compatible meta tag

querkmachine:
This meta tag ensured that older versions of Internet Explorer (specifically IE8 to IE10) used the most recent renderer and rendering rules available to them, rather than relying on the browser's in-built heuristics that could silently 'downgrade' the renderer. For example: if IE10 thought you were viewing a Sharepoint intranet site, it might automatically switch itself to use IE9's rendering rules.

IE11 does not require this meta tag. It deprecates the X-UA-Compatible meta tag and now uses the most recent renderer automatically if the HTML5 doctype (`<!DOCTYPE html>`) is being used.

Both the EdgeHTML and Chromium versions of Edge ignore the meta tag entirely. No non-Microsoft browsers ever supported it, as far as I can find. 

As Frontend now only supports Internet Explorer 11, I think we can safely remove this meta tag. 

Info from Microsoft regarding the deprecation in IE11: https://learn.microsoft.com/en-us/previous-versions/windows/internet-explorer/ie-developer/dev-guides/bg182625(v=vs.85)#document-mode-changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,22 @@ For advice on how to use these release notes see [our guidance on staying up to 
 
 ## Unreleased
 
+### Recommended changes
+
+We've recently made some non-breaking changes to GOV.UK Frontend. Implementing these changes will make your service work better.
+
+#### Remove the X-UA-Compatible meta tag
+
+Remove the `<meta http-equiv="X-UA-Compatible" content="IE=edge">` meta tag from your page template.
+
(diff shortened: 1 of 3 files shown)


colinrotherham (approved):

36degrees (commented):
I think this probably needs a changelog entry.

querkmachine:
@36degrees Added a changelog 👍

36degrees (approved):

Issue #4225, opened 2023-09-19
Create prototype show/hide password component


querkmachine:
## What

Create a prototype show/hide component based on [the contribution from GOV.UK Publishing Components](https://components.publishing.service.gov.uk/component-guide/show_password).

In doing so, we will tweak aspects of the component to be internally consistent with the Design System, most notably, changing the toggle button to resemble [the Button component](https://design-system.service.gov.uk/components/button/) instead of using a bespoke version of the link styles.

## Why

So we have a version of the component that adheres to Frontend's coding standards and architecture. 

So we have a 'baseline' component we can test against and make iterative changes to. 

## Who needs to work on this

- Designers
- Developers

## Who needs to review this

- Designers
- Developers

## Done when

- [x] The show/hide password component from Publishing Components has been ported to Frontend

PR #4442, opened 2023-11-07
Create password input component

querkmachine:
Prototypical password input component for #4225.

[See it in the review app](https://govuk-frontend-pr-4442.herokuapp.com/components/password-input). 

## Changes

- Adds a new component: Password input.
  - This is a (mostly) straight port of [the component in GOV.UK Publishing Components](https://components.publishing.service.gov.uk/component-guide/show_password) with some minor alterations:
    - JavaScript has been modified to fit the Design System's coding and documentation conventions.
    - Support for configuration and localisation strings has been added.
    - The option to disable automatically switching the input back to a `password` type upon form submission has been removed. We couldn't find any evidence that this was being used by any live service and it doesn't serve a clear user need, so we've chosen to omit it unless a need arises. 
    - The [function to move data-attributes from the input to the toggle button](https://components.publishing.service.gov.uk/component-guide/show_password#with_data_attributes) has been omitted.
      - I think this code primarily existed for use with analytics event tracking, which isn't normally a use case we account for in Design System components, but I'm interested if there's other uses where we might want this. 
      - This has probably been superseded anyway by the button existing as a configurable sub-component rather than being purely injected by JS.
    - Custom button styles have not been ported, instead using the button from the Design System.
- Updates `all.mjs` to export the Password input and automatically use it if `data-module="govuk-password-input"` is found on a page.
- Updates the Prototype Kit configuration to export the Password input.
- Adds tests for new component and altered functionality elsewhere.

### Changes to other components

- The Text input component has had a new `autocapitalize` attribute and parameter added.
- The text input component now has an `inputWrapper` parameter that allows for setting classes and attributes on the `govuk-input__wrapper` element. 

### Changes to the review app
- Adds a new "Sign in to a service" full-page example with the password input component with the `current-password` autocomplete value.
- Updates the "Update your account details" full-page example to use the password input component with the `new-password` autocomplete value.

## Thoughts

- The input `type` is hardcoded to `password` in the initial HTML, as I don't envisage this practically being anything else.
- For [accessible authentication](https://www.w3.org/WAI/WCAG22/Understanding/accessible-authentication-minimum) reasons the input should virtually always have an `autocomplete` attribute. As such, I've made it a required attribute that defaults to `current-password`.
- The `spellcheck` and `autocapitalize` attributes are both set to off/disabled. This is to prevent the user's device attempting to 'correct' the user's input if they are typing a password whilst the input is displaying as plain text. 
- Various parameters that are available on the text input component—such as `pattern`, `inputmode` and prefix/suffixes—are not accessible via the Password input. I think these have limited application when it comes to passwords and could only be used in ways we would typically advise against (e.g. client-side validation, number-only passwords).


Ciandelle (approved):

owenatgov (commented):
I don't have loads to add! I haven't looked at the code in great depth as this is just a prototype but it all seems to work as expected.

The button thing is tricky but not one for right now, I agree. It looks disturbing enough that it's a problem so we should solve it eventually.

From memory, the reason the original was designed the way it was was to distinguish between other potential buttons in a journey. A secondary button that's adjusted to line up with the input is probably fine. From poking around in DI's journeies, it looks like this is only used on pages where there's a primary button as a CTA.

Personally I wonder if we should consider our button ecosystem and see if we should add to it. I wonder if this is something we ask from the working group during the first round.

querkmachine:
I've also heard rumours that the appearance of the secondary button in the Design System isn't favoured because it looks like it might be disabled (given it's all light grey / monochrome). I can't be sure of the veracity of said rumours, but it seems like a fair criticism of the current design.

owenatgov:
A couple of extra things from chatting to @andysellick:

1. We reckon that my hypothesis was correct that this was to avoid visual clashes with primary buttons on a given view. Why it looks like a link inside a box is still unclear and probably a best effort to solve the problem by designers at the time. I think we have licence to change it.
2. Is there value in us putting together a login page full page example to demo this component in situ to assess it better? @wilsond-gds shared some example auth journey views with us on slack, we could probably pick one of them with his guidance.

36degrees:
> For [accessible authentication](https://www.w3.org/WAI/WCAG22/Understanding/accessible-authentication-minimum) reasons the input should virtually always have an `autocomplete` attribute. As such, I've made it a required attribute that defaults to `current-password`. I've added a `newPassword` boolean parameter to switch it to `new-password` when needed.

I think the exception to this would be when setting the password for another user, e.g. as part of user management in a case working or admin system – in which case the `autocomplete` attribute should not be used.

I think it does make sense to default to `current-password`, but I think we should just make `autocomplete` an option and ensure it can be omitted entirely.

> The `spellcheck` and `autocapitalize` attributes are both set to off/disabled. This is to prevent the user's device attempting to 'correct' the user's input if they are typing a password whilst the input is displaying as plain text.

The [`autocapitalize` attribute is explicitly disabled for password inputs](https://html.spec.whatwg.org/multipage/interaction.html#autocapitalization) in the spec.

I’d say spellcheck is implicitly off too, based on [this part of the spec](https://html.spec.whatwg.org/multipage/interaction.html#spelling-and-grammar-checking) (emphasis mine):

> User agents must **only** consider the following pieces of text as checkable for the purposes of this feature:
> 
> - The [value](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#concept-fe-value) of [`input`](https://html.spec.whatwg.org/multipage/input.html#the-input-element) elements whose [`type`](https://html.spec.whatwg.org/multipage/input.html#attr-input-type) attributes are in the [Text](https://html.spec.whatwg.org/multipage/input.html#text-(type=text)-state-and-search-state-(type=search)), [Search](https://html.spec.whatwg.org/multipage/input.html#text-(type=text)-state-and-search-state-(type=search)), [URL](https://html.spec.whatwg.org/multipage/input.html#url-state-(type=url)), or [Email](https://html.spec.whatwg.org/multipage/input.html#email-state-(type=email)) states and that are [mutable](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#concept-fe-mutable) (i.e. that do not have the [`readonly`](https://html.spec.whatwg.org/multipage/input.html#attr-input-readonly) attribute specified and that are not [disabled](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#concept-fe-disabled)).

If we **know** there are browsers that don’t follow the spec correctly on either of those attributes (i.e. they incorrectly autocapitalize or spellcheck them) then we could consider adding them. Otherwise, we shouldn't omit them as they shouldn't be necessary.

querkmachine:
> I think the exception to this would be when setting the password for another user, e.g. as part of user management in a case working or admin system – in which case the `autocomplete` attribute should not be used.
> 
> I think it does make sense to default to `current-password`, but I think we should just make `autocomplete` an option and ensure it can be omitted entirely.

User agents use their own heuristics to decide that a password input should be automatically fillable (e.g. based on the input type or the associated label) and [many will still present the option to autofill](https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion#the_autocomplete_attribute_and_login_fields) even if the `autocomplete` attribute is omitted or set to `off`. 

Given the autofill behaviour is almost always going to be presented to a user anyway, my thinking is that it's better to be explicit, rather than omitting it and potentially presenting the wrong thing (e.g. the option to autofill a new password field). 

> The [`autocapitalize` attribute is explicitly disabled for password inputs](https://html.spec.whatwg.org/multipage/interaction.html#autocapitalization) in the spec.
> 
> I’d say spellcheck is implicitly off too, based on [this part of the spec](https://html.spec.whatwg.org/multipage/interaction.html#spelling-and-grammar-checking) (emphasis mine):

The attributes are primarily to avoid those issues when the password has been made visible, such as if the toggle feature is being used for data entry and not just reviewing input after the fact.

36degrees:
> > I think the exception to this would be when setting the password for another user, e.g. as part of user management in a case working or admin system – in which case the `autocomplete` attribute should not be used.
> > I think it does make sense to default to `current-password`, but I think we should just make `autocomplete` an option and ensure it can be omitted entirely.
> 
> User agents use their own heuristics to decide that a password input should be automatically fillable (e.g. based on the input type or the associated label) and [many will still present the option to autofill](https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion#the_autocomplete_attribute_and_login_fields) even if the `autocomplete` attribute is omitted or set to `off`.
> 
> Given the autofill behaviour is almost always going to be presented to a user anyway, my thinking is that it's better to be explicit, rather than omitting it and potentially presenting the wrong thing (e.g. the option to autofill a new password field).

Just to make sure we're on the same page, happy with defaulting to `current-password`, but I do think users should be able to set the `autocomplete` attribute to anything, rather than just having the `newPassword` boolean.

If we _really_ think constraining to only `current-password` / `new-password` is the right thing to do, my preference would be to have an `autocomplete` option that only accepts those two options, rather than a boolean. Boolean options can become problematic when you want to introduce additional options in the future, as you end up with competing booleans where the behaviour is hard to define.

> > The [`autocapitalize` attribute is explicitly disabled for password inputs](https://html.spec.whatwg.org/multipage/interaction.html#autocapitalization) in the spec.
> > I’d say spellcheck is implicitly off too, based on [this part of the spec](https://html.spec.whatwg.org/multipage/interaction.html#spelling-and-grammar-checking) (emphasis mine):
> 
> The attributes are primarily to avoid those issues when the password has been made visible, such as if the toggle feature is being used for data entry and not just reviewing input after the fact.

Ah, of course! 🤦🏻 That makes sense, thanks for explaining. It might be worth adding a comment along those lines?

querkmachine:
> Just to make sure we're on the same page, happy with defaulting to `current-password`, but I do think users should be able to set the `autocomplete` attribute to anything, rather than just having the `newPassword` boolean.
> 
> If we _really_ think constraining to only `current-password` / `new-password` is the right thing to do, my preference would be to have an `autocomplete` option that only accepts those two options, rather than a boolean. Boolean options can become problematic when you want to introduce additional options in the future, as you end up with competing booleans where the behaviour is hard to define.

Yeah, fair point on the boolean parameter. 

Regarding constraints, I guess the question is what other `autocomplete` values would we expect? For the purposes of the Accessible Authentication criterion the only compliant ones that currently exist would seem to be `current-password` and `new-password`. Any other value would either bypass or undermine the goals of the component.

`one-time-code` is a maybe, I guess, though [the guidance](https://github.com/alphagov/govuk-design-system/pull/3355) I've been drafting explicitly warns against using the password input for OTPs, as there isn't a significant practical or security benefit to them being hidden. 

`webauthn` is another possibility for services with passkeys, though the guidance I've seen says this value is better placed on the username or email field rather than the password field. (The goal of passkeys is to replace passwords, after all.)

On that basis I personally wouldn't have a problem with constraining the value to being one of `current-password` or `new-password`.

owenatgov:
Revisiting the design of this, specifically the little bottom "box shadow" flowing below the baseline of the input. We should make a call on re: if that bottom border belongs in the bounds of the button or outside. We have this [form alignment page](https://govuk-frontend-review.herokuapp.com/examples/form-alignment) where we opt for the border being inline with text inputs. For my part I find it being inline more pleasing. The bottom border is intended as a box shadow but also it feels to me like it exists as a whole thing, not a button and a box shadow, if that makes sense. Interested in other takes.

querkmachine:
I've rebased and updated this spike to tick off the stylistic todos: 

- Changed `align-items` from the default `stretch` to `flex-start` so that the already-thought-about alignment gubbins @owenatgov mentioned above can work its magic.
- Added a small gap between the input and button so that the input's focus ring doesn't get obscured by the button.
- Mitigated the button changing width when the text inside of it is switched by setting a `flex-basis` that works for the default text, though translations/alternate text can still cause this to happen.

packages/govuk-frontend/src/govuk/components/password-input/password-input.mjs
```diff
@@ -0,0 +1,265 @@
...
+    // moving through the page content linearly:
+    // [password input] -> [your password is visible/hidden] -> [show/hide password]
+    this.$statusText = document.createElement('span')
+    this.$statusText.className = 'govuk-visually-hidden'
+    this.$statusText.innerText = this.i18n.t('passwordHiddenAnnouncement')
+    this.$statusText.setAttribute('aria-live', 'polite')
+    this.$input.parentNode?.insertBefore(
+      this.$statusText,
+      this.$input.nextSibling
+    )
```
colinrotherham, line 117: Wondering if we should align with Character count here?

1. Naming convention `$screenReaderXXXMessage`
2. Updates via `updateScreenReaderXXXMessage()`
3. Element class `__sr-status` used in test selectors
4. Element as `<div>` rather than `<span>`
5. Injected via `insertAdjacentElement()`

https://github.com/alphagov/govuk-frontend/blob/56fc909f395caede98440e061488bddc940fef81/packages/govuk-frontend/src/govuk/components/character-count/character-count.mjs#L159-L169

colinrotherham: Or whether we should abstract both away into a utility, base class or Nunjucks macro?

querkmachine: Probably sensible to align variable names and classes closer to how the character count has it, agreed. 

I don't think 2 is needed in this case, character count only does has a separate method because it additionally has to account for the threshold functionality and to differentiate it from the separate visible counter. 

Two mostly-similar cases doesn't constitute a trend in my mind, I think abstracting them might be a premature optimisation.

owenatgov (approved):
Looking sharp. Done some browser testing and screen reader testing. All looking good to me. Code looking good as well. I've left a single very pithy comment that's far from critical.

I haven't looked at the puppeteer tests very much as it looks like this is being handled in https://github.com/alphagov/govuk-frontend/pull/4763.

querkmachine:
Anika has concluded testing https://github.com/alphagov/govuk-frontend/pull/4442/commits/80d33b49d589ea1ffea15f9365e8ee1bc988a6ca and found that it indeed resolves the issues with VoiceOver status announcements without negatively affecting other screen readers.

Reverting when the `aria-live` attribute is added does mean that JAWS and NVDA will re-announce the status in the event a user returns to the page via the bfcache, after having previously entered a value and toggled it to be visible before leaving. 

I would consider this issue to be secondary to ensuring that announcements occur consistently on the 'happy path' in VoiceOver, however, so the reversion feels sensible.

With that, I think this is mergeable! 🎉

Issue #4443, opened 2023-11-07
UMD component exports have duplicate names


colinrotherham:
## Description of the issue

GOV.UK Frontend v5 beta releases include Universal Module Definition (UMD) bundles for:

1. Entire package as **all.bundle.js**
2. Individual components as **accordion.bundle.js**, **button.bundle.js** etc

This appears to be related to workarounds added (see https://github.com/alphagov/govuk-frontend/pull/2426, https://github.com/alphagov/govuk-frontend/issues/2829, https://github.com/alphagov/govuk-frontend/pull/2899) to rename default (unnamed) exports when older Rollup releases didn't support named exports in the same bundle

See the following `componentPathToModuleName()` helper for context:

https://github.com/alphagov/govuk-frontend/blob/4d13a759e4ffdea97309fde322eeb82087ac6e23/packages/govuk-frontend/rollup.publish.config.mjs#L49-L51

Since we moved to named exports in https://github.com/alphagov/govuk-frontend/commit/b94e8127dfcb8f3363092ecbd9d995d89eb12314 as part of https://github.com/alphagov/govuk-frontend/pull/3771 we can likely remove the helper to fix the issue

## Steps to reproduce the issue

When using individual component bundles, code using existing `require()` calls (CommonJS, AMD) or window global properties no longer work and must be updated to use duplicate names instead:

```patch
- const { Accordion } = window.GOVUKFrontend
+ const { Accordion } = window.GOVUKFrontend.Accordion
```

PR #4444, opened 2023-11-08
Fix UMD component exports with duplicate names

colinrotherham:
Closes https://github.com/alphagov/govuk-frontend/issues/4443 by removing the `componentPathToModuleName()` helper

Rollup is now configured to automatically add named exports to `window.GOVUKFrontend`


domoscargin (approved):
Love a diff with way more red than green.

Let's get this in a beta release and check it works in https://github.com/alphagov/govuk_publishing_components/pull/3700

Issue #4440, opened 2023-11-07
Implement 'transitional' crown assets in Frontend (v5.0)


querkmachine:
## What

Implement a set of 'transitional' crown assets in GOV.UK Frontend v5.0.

These assets will use the new formats decided in #4175 but continue to show the St. Edward's crown.

## Why

There is a reasonable chance that we will not have permission to launch the Tudor crown until after we intend to launch GOV.UK Frontend 5.0. 

To mitigate this, we will make as many changes to the code and assets as possible in 5.0 whilst maintaining the St. Edward's crown. We will then switch to the Tudor crown in a subsequent release. See #4178.

## Who needs to work on this

- Developers

## Who needs to review this

- Developers
- Designers
- Accessibility specialist

## Done when

```[tasklist]
- [x] Inline SVG graphic has been implemented in Header component
- [x] Accessibility considerations for logo's change in have been made (e.g. removal of `role="presentation"`)
- [x] Sass has been updated to account for the new lockup (link hover styles, product name integration, etc.)
- [x] Favicon and web sharing graphics have been replaced
```

PR #4354, opened 2023-10-19
[DO NOT MERGE] Implement the Tudor crown favicons (v5.x)

querkmachine:
Implements the Tudor crown for browser 'favourite' icons, homescreen icons and OpenGraph sharing graphics. Part of #4178. This follows the spike work in #4185 using the finalised set of icons and production-ready assets.

> [!WARNING]
> This PR should not be merged until we have approval to do so.

## Changes

- Replaces icons using the St. Edward's crown with equivalents using the Tudor crown. 
- Renames all of the icons to be largely agnostic to their usage, instead being named for the image's dimensions.
- Changes the way the ICO favicon is referenced in HTML: 
  - Fixes a potential issue with the `sizes` attribute where Chrome would always use the ICO even when an SVG favicon was available ([`sizes` issue detailed here in section 2.3](https://dev.to/masakudamatsu/favicon-nightmare-how-to-maintain-sanity-3al7)).
  - Changes `rel="shortcut icon"` to `rel="icon"`. The `shortcut` keyword was only needed for supporting Internet Explorer 8 and earlier, which Frontend no longer does.
- Removes `apple-touch-icon` sizes smaller than 180×180. These aren't used by Apple devices released in the last few years and lower resolution Apple devices can scale down the larger image automatically.
- Redesigns the smallest icon sizes (48px and below) to use a simplified version of the Tudor crown icon. 
- Adds an SVG version of the favicon graphic.
- Adds 192×192 and 512×512 PNG icons. 
- Inverts the design of the mask icon to be consistent with how the icon is presented elsewhere. The crown is now transparent and presented within a coloured box.
- Adds a `manifest.json` file listing all of the available icon sizes and their expected uses.
- Adds Jest tests for the icons and manifest file. 

## Assets included
|Name|Type|Dimensions|Notes|
|:-|:-|-:|:-|
|favicon.ico|Icon container|16×16, 32×32, 48×48|Multiple PNGs bundled into a single container. All images use simplified crown design. Primarily used by Safari, IE and some third-parties.|
|favicon.svg|SVG|variable|Uses simplified crown design. Primarily used by Chromium and Firefox.|
|govuk-icon-180.png|PNG|180×180|Primarily used on the iOS/iPadOS homescreen.|
|govuk-icon-192.png|PNG|192×192|Primarily used on the Android homescreen.|
[...]

PR #4445, opened 2023-11-09
Implement transitional crown favicons (v5.0)

querkmachine:
Implements transitional crown assets for browser 'favourite' icons, homescreen icons and OpenGraph sharing graphics. That is, assets that utilise the formats agreed for Frontend v5 (#4175), but using the St. Edward's crown. Part of #4440. 

This is intended to be a transitional step prior to launching the Tudor crown in a future version: #4354

## Changes

- Carries over the ICO, mask icon and OpenGraph image from GOV.UK Frontend v4 unchanged.
- Renames all of the icons to be largely agnostic to their usage, instead being named for the image's dimensions.
- Changes the way the ICO favicon is referenced in HTML: 
  - Fixes a potential issue with the `sizes` attribute where Chrome would always use the ICO even when an SVG favicon was available ([`sizes` issue detailed here in section 2.3](https://dev.to/masakudamatsu/favicon-nightmare-how-to-maintain-sanity-3al7)).
  - Changes `rel="shortcut icon"` to `rel="icon"`. The `shortcut` keyword was only needed for supporting Internet Explorer 8 and earlier, which Frontend no longer does.
- Removes `apple-touch-icon` sizes smaller than 180×180. These aren't used by Apple devices released in the last few years and lower resolution Apple devices can scale down the larger image automatically.
- Adds an SVG version of the favicon graphic. We've opted not to have this icon change depending on the user's dark or light mode setting.
- Adds 192×192 and 512×512 PNG icons. 
- Redesigns the 180×180 PNG icon to match the larger versions.
- Adds a `manifest.json` file listing all of the available icon sizes and their expected uses.
- Adds Jest tests for the icons and manifest file. 

## Assets included
|Name|Type|Dimensions|Notes|
|:-|:-|-:|:-|
|**Icon container**<br>favicon.ico|Icon container|16×16, 32×32, 48×48|Multiple PNGs bundled into a single container. Primarily used by Safari (iOS/iPadOS), IE and some third-parties. This has been carried over from v4.|
|**SVG icon**<br>favicon.svg|SVG|variable|Primarily used by Chromium and Firefox. This is new for v5.|
|**iOS icon**<br>govuk-icon-180.png|PNG|180×180|Primarily used on the iOS/iPadOS homescreen. This has been redesigned for v5.|
|**Android icon**<br>govuk-icon-192.png|PNG|192×192|Primarily used on the Android homescreen. This is new for v5.|
|**Large icon**<br>govuk-icon-512.png|PNG|512×512|Fallback for any applications that request a larger size but cannot use SVG. May also be used by Safari 17's "Add to dock" feature on macOS. This is new for v5.|
|**Mask icon**<br>govuk-icon-mask.svg|SVG|variable|Primarily used by Safari (macOS) and MacBooks featuring the Touch Bar. This has been carried over from v4.|
|**OpenGraph image**<br>govuk-opengraph-image.png|PNG|1200×630|Primarily used when sharing on third-party websites and services. This has been carried over from v4.|


packages/govuk-frontend/src/govuk/template.njk
```diff
@@ -10,12 +10,11 @@
     <meta name="theme-color" content="{{ themeColor | default("#0b0c0c", true) }}"> {# Hardcoded value of $govuk-black #}
```
colinrotherham, line 10: Mind if we match this one up with the ~`{#`~ `{#-` change on line 15?

```suggestion
    <meta name="theme-color" content="{{ themeColor | default("#0b0c0c", true) }}"> {#- Hardcoded value of $govuk-black #}
```

colinrotherham (changes requested):
Looking good @querkmachine 🙌

Mind if you and @claireashworth work together to add a new **Breaking changes** entry to the CHANGELOG specifically about:

1. Removing old `govuk-apple-touch-*` icons
2. Removing old `govuk-mask-*` icons
3. Adding new `govuk-icon-*` icons
4. Adding new `manifest.json` file

Perhaps consolidating it with other removed images changes?

```md
#### Remove the fallback GOV.UK crown logo from your HTML
```

As lots of services may have these files committed, probably want them to confirm the new files are reachable from `assetPath` (or `/assets`) without 404s

querkmachine:
I've added a changelog entry, though I have not consolidated it with the fallback crown entry as they feel conceptually separate and have fairly different associated HTML changes.

CHANGELOG.md
```diff
@@ -4,6 +4,43 @@ For advice on how to use these release notes see [our guidance on staying up to
 
 ## Unreleased
 
+### Breaking changes
+
+#### Check your favicons, app icons and OpenGraph image still work
+
+We've changed the names, formats and sizes of icon assets that we distribute in Frontend. You will want to check that the correct files are in place.
```
colinrotherham, line 11:
```suggestion
We've changed the names, formats and sizes of icon assets that we distribute in GOV.UK Frontend. You will want to check that the correct files are in place.
```

CHANGELOG.md
```diff
@@ -4,6 +4,43 @@ For advice on how to use these release notes see [our guidance on staying up to
...
+
+```html
+<link rel="icon" sizes="48x48" href="/assets/images/favicon.ico">
+<link rel="icon" sizes="any" href="/assets/images/favicon.svg" type="image/svg+xml">
+<link rel="mask-icon" href="/assets/images/govuk-icon-mask.svg" color="#0b0c0c">
+<link rel="apple-touch-icon" href="/assets/images/govuk-icon-180.png">
+<link rel="manifest" href="/assets/manifest.json">
+```
+
+You will need to update the file path to match your assets folder if it's not at the default location.
```
colinrotherham, line 40: I've pinched this from **Remove the fallback GOV.UK crown logo from your HTML**

Think it's helpful?

```suggestion
You will need to update the file paths to match your assets folder depending on [how you are serving the font and image assets](https://frontend.design-system.service.gov.uk/importing-css-assets-and-javascript/#font-and-image-assets).
```

CHANGELOG.md
```diff
@@ -4,6 +4,43 @@ For advice on how to use these release notes see [our guidance on staying up to
 
 ## Unreleased
 
+### Breaking changes
+
+#### Check your favicons, app icons and OpenGraph image still work
```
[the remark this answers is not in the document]
claireashworth: I think make 'image' plural instead of 'work'.

CHANGELOG.md
```diff
@@ -4,6 +4,43 @@ For advice on how to use these release notes see [our guidance on staying up to
...
+
+We've changed the names, formats and sizes of icon assets that we distribute in Frontend. You will want to check that the correct files are in place.
+
+The following files have been added to the assets folder:
```
colinrotherham, line 13: I think we need to call out `assetPath` (which defaults to `/assets`), `assetUrl` and `opengraphImageUrl`

Could be helpful to refer to the link in https://github.com/alphagov/govuk-frontend/pull/4445/files#r1389317268 earlier but see what @claireashworth recommends

CHANGELOG.md
claireashworth: I made a couple of tweaks to Colin's suggestion

CHANGELOG.md
claireashworth, line 13:
```suggestion
We've added the following files to the assets folder:
```

CHANGELOG.md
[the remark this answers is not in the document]
colinrotherham: My brain said to do this but there's only a single OpenGraph image

Would you make _images_ apply the other types too?

```suggestion
#### Check your favicon, app icon and OpenGraph images still work
```

claireashworth:
Just a couple of minor suggestions from me. Also, should this be moved within the rest of the v5 information or is that done for the release rather than the pre-release?

CHANGELOG.md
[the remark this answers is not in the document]
claireashworth: Ah ok, another option is to put 'OpenGraph image' at the start of the list, the brain ignores the inconsistency

CHANGELOG.md
claireashworth: Can do this. Another option is to move the file path line up between current lines 11 and 13, as it's an action users need to do

colinrotherham:
@claireashworth It's useful for beta participants for now (separately) until the beta's over, then combine?

CHANGELOG.md
colinrotherham: Was thinking more for context as lots of services use project-specific paths. When you read "assets folder" it doesn't necessarily help identify where the images are copied or deployed

>If a user is already using the Nunjucks template (and thus `assetPath`, et al.), then they are will already have these variables configured. They don't need to be changed if the images are being replaced in-place.

Even the Nunjucks templates can override them elsewhere:

```mjs
{
  assetPath: "/specific/directory/that/other/team/manages",
  assetUrl: "http://cdn-assets-here.domain.com/assets",
  opengraphImageUrl: "http://social-media-contract-assets-here.domain.com/assets/specific.png"
}
```

querkmachine: Do we still think specifically calling it out is needed now that documentation relating to it is linked from the changelog entry?

claireashworth: I think it should be ok as is. We can always amend it if we get feedback to the contrary.

colinrotherham: Nice, thanks for sticking the link in. Yeah we're all set now 🙌

colinrotherham (approved):

PR #3766, opened 2023-06-07
Remove JavaScript from Details component

domoscargin:
Part of #3464

PR #4446, opened 2023-11-09
Remove ‘details polyfill’ example from review app

36degrees:
This example was used to test the JavaScript polyfill for the details component on a `<div>` to ensure that it added the same functionality as the native `<details>` element.

We removed that JavaScript polyfill in #3766.

This example no longer functions correctly (it renders something that looks like the details component, but clicking the summary does not open or close) and serves no purpose, so we can remove it.


domoscargin (approved):
🔥

PR #4450, opened 2023-11-10
Update descriptions for Nunjucks macro options + fixes

colinrotherham:
Adds new descriptions from https://github.com/alphagov/govuk-design-system/pull/2929#issuecomment-1798183406

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/character-count/character-count.yaml
+++ b/packages/govuk-frontend/src/govuk/components/character-count/character-count.yaml
@@ -40,7 +40,7 @@ params:
   - name: errorMessage
     type: object
     required: false
-    description: Options for the error message component. The error message component will not display if you use a falsy value for `errorMessage`, for example `false` or `null`.
+    description: Can be used to add an error message to the character count component. The error message component will not display if you use a falsy value for `errorMessage`, for example `false` or `null`.
     isComponent: true
   - name: formGroup
     type: object
--- a/packages/govuk-frontend/src/govuk/components/character-count/character-count.yaml
+++ b/packages/govuk-frontend/src/govuk/components/character-count/character-count.yaml
@@ -30,7 +30,7 @@ params:
   - name: label
     type: object
     required: true
-    description: Additional options for the label used by the character count component.
+    description: The label used by the character count component.
     isComponent: true
   - name: hint
     type: object
--- a/packages/govuk-frontend/src/govuk/components/character-count/character-count.yaml
+++ b/packages/govuk-frontend/src/govuk/components/character-count/character-count.yaml
@@ -30,12 +30,12 @@ params:
   - name: label
     type: object
     required: true
-    description: Options for the label component.
+    description: Can be used to add a label to a character count component.
     isComponent: true
   - name: hint
     type: object
     required: false
-    description: Options for the hint component.
+    description: Can be used to add a hint to the character count component.
     isComponent: true
   - name: errorMessage
     type: object
@@ -45,7 +45,7 @@ params:
   - name: formGroup
     type: object
     required: false
-    description: Options for the form-group wrapper.
+    description: Additional options for the form group containing the character count component.
     params:
       - name: classes
         type: string
@@ -66,7 +66,7 @@ params:
   - name: countMessage
     type: object
     required: false
-    description: Options for the count message.
+    description: The message for the character count component.
     params:
       - name: classes
         type: string
--- a/packages/govuk-frontend/src/govuk/components/character-count/character-count.yaml
+++ b/packages/govuk-frontend/src/govuk/components/character-count/character-count.yaml
@@ -30,7 +30,7 @@ params:
   - name: label
     type: object
     required: true
-    description: Can be used to add a label to a character count component.
+    description: Additional options for the label used by the character count component.
     isComponent: true
   - name: hint
     type: object
@@ -66,7 +66,7 @@ params:
   - name: countMessage
     type: object
     required: false
-    description: The message for the character count component.
+    description: Additional options for the count message used by the character count component.
     params:
       - name: classes
         type: string
--- a/packages/govuk-frontend/src/govuk/components/cookie-banner/cookie-banner.yaml
+++ b/packages/govuk-frontend/src/govuk/components/cookie-banner/cookie-banner.yaml
@@ -2,7 +2,7 @@ params:
   - name: ariaLabel
     type: string
     required: false
-    description: The text for the `aria-label` which labels the cookie banner region. This region applies to all messages that the cookie banner includes. For example, the cookie message and the confirmation message. Defaults to 'Cookie banner'.
+    description: The text for the `aria-label` which labels the cookie banner region. This region applies to all messages that the cookie banner includes. For example, the cookie message and the confirmation message. Defaults to `"Cookie banner"`.
   - name: hidden
     type: boolean
     required: false
@@ -39,7 +39,7 @@ params:
       - name: actions
         type: array
         required: false
-        description: The buttons and links that you want to display in the message. `actions` defaults to `button` unless you set `href`, which renders the action as a link.
+        description: The buttons and links that you want to display in the message. `actions` defaults to `"button"` unless you set `href`, which renders the action as a link.
         params:
           - name: text
             type: string
@@ -48,11 +48,11 @@ params:
           - name: type
             type: string
             required: false
-            description: The type of button. You can set `button` or `submit`. Set `button` and `href` to render a link styled as a button. If you set `href`, it overrides `submit`.
+            description: The type of button. You can set `"button"` or `"submit"`. Set `type` to `"button"` and set `href` to render a link styled as a button. If you set `href`, it overrides `type` `"submit"`.
           - name: href
             type: string
             required: false
-            description: The `href` for a link. Set `button` and `href` to render a link styled as a button.
+            description: The `href` for a link. Set `type` to `"button"` and set `href` to render a link styled as a button.
           - name: name
             type: string
             required: false
@@ -76,7 +76,7 @@ params:
       - name: role
         type: string
         required: false
-        description: Set `role` to `alert` on confirmation messages to allow assistive tech to automatically read the message. You will also need to move focus to the confirmation message using JavaScript you have written yourself.
+        description: Set `role` to `"alert"` on confirmation messages to allow assistive tech to automatically read the message. You will also need to move focus to the confirmation message using JavaScript you have written yourself.
       - name: classes
         type: string
         required: false
--- a/packages/govuk-frontend/src/govuk/components/cookie-banner/cookie-banner.yaml
+++ b/packages/govuk-frontend/src/govuk/components/cookie-banner/cookie-banner.yaml
@@ -48,7 +48,7 @@ params:
           - name: type
             type: string
             required: false
-            description: The type of button. You can set `"button"` or `"submit"`. Set `type` to `"button"` and set `href` to render a link styled as a button. If you set `href`, it overrides `type` `"submit"`.
+            description: The type of button – `"button"` or `"submit"`. If `href` is provided, set `type` to `"button"` render a link styled as a button.
           - name: href
             type: string
             required: false
--- a/packages/govuk-frontend/src/govuk/components/date-input/date-input.yaml
+++ b/packages/govuk-frontend/src/govuk/components/date-input/date-input.yaml
@@ -10,7 +10,7 @@ params:
   - name: items
     type: array
     required: false
-    description: An array of input objects with name, value and classes.
+    description: The items of each date input component.
     params:
       - name: id
         type: string
@@ -47,7 +47,7 @@ params:
   - name: hint
     type: object
     required: false
-    description: Options for the hint component.
+    description: Can be used to add a hint to a date input component.
     isComponent: true
   - name: errorMessage
     type: object
@@ -57,7 +57,7 @@ params:
   - name: formGroup
     type: object
     required: false
-    description: Options for the form-group wrapper.
+    description: Additional options for the form group containing the date input component.
     params:
       - name: classes
         type: string
--- a/packages/govuk-frontend/src/govuk/components/date-input/date-input.yaml
+++ b/packages/govuk-frontend/src/govuk/components/date-input/date-input.yaml
@@ -6,11 +6,11 @@ params:
   - name: namePrefix
     type: string
     required: false
-    description: Optional prefix. This is used to prefix each item `name` using `-`.
+    description: Optional prefix. This is used to prefix each item `name`, separated by `-`.
   - name: items
     type: array
     required: false
-    description: The items of each date input component.
+    description: The inputs within the date input component.
     params:
       - name: id
         type: string
--- a/packages/govuk-frontend/src/govuk/components/date-input/date-input.yaml
+++ b/packages/govuk-frontend/src/govuk/components/date-input/date-input.yaml
@@ -6,7 +6,7 @@ params:
   - name: namePrefix
     type: string
     required: false
-    description: Optional prefix. This is used to prefix each `item.name` using `-`.
+    description: Optional prefix. This is used to prefix each item `name` using `-`.
   - name: items
     type: array
     required: false
@@ -31,7 +31,7 @@ params:
       - name: autocomplete
         type: string
         required: false
-        description: Attribute to [identify input purpose](https://www.w3.org/WAI/WCAG21/Understanding/identify-input-purpose.html), for instance `bday-day`. See [autofill](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofill) for full list of attributes that can be used.
+        description: Attribute to [identify input purpose](https://www.w3.org/WAI/WCAG21/Understanding/identify-input-purpose.html), for instance `"bday-day"`. See [autofill](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofill) for full list of attributes that can be used.
       - name: pattern
         type: string
         required: false
--- a/packages/govuk-frontend/src/govuk/components/date-input/date-input.yaml
+++ b/packages/govuk-frontend/src/govuk/components/date-input/date-input.yaml
@@ -52,7 +52,7 @@ params:
   - name: errorMessage
     type: object
     required: false
-    description: Options for the error message component. The error message component will not display if you use a falsy value for `errorMessage`, for example `false` or `null`.
+    description: Can be used to add an error message to the date input component. The error message component will not display if you use a falsy value for `errorMessage`, for example `false` or `null`.
     isComponent: true
   - name: formGroup
     type: object
--- a/packages/govuk-frontend/src/govuk/components/date-input/date-input.yaml
+++ b/packages/govuk-frontend/src/govuk/components/date-input/date-input.yaml
@@ -66,7 +66,7 @@ params:
   - name: fieldset
     type: object
     required: false
-    description: Options for the fieldset component (for example legend).
+    description: Can be used to add a fieldset to the date input component.
     isComponent: true
   - name: classes
     type: string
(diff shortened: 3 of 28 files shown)


colinrotherham:
Mind taking a look at some suggestions of mine please @CharlotteDowns?

Mainly to swap "Can be used to add a…" with "Additional options for…" when defaults are already provided

I didn't include `fieldset` or nested `hint`/`label` for Checkboxes/Radios but take a look, can keep tweaking

CharlotteDowns (approved):
Looks good to me

colinrotherham:
@36degrees Shall we add in some of your suggestions from https://github.com/alphagov/govuk-frontend/pull/3919 before merging this?

One of them was moving the "this is not required" text to the end:

```patch
- description: If `text` is set, this is not required. The heading HTML content of each section. The header is inside the HTML `<button>` element, so you can only add [phrasing content](https://html.spec.whatwg.org/#phrasing-content) to it. If `html` is provided, the `text` option will be ignored.
+ description: The heading HTML content of each section. The header is inside the HTML `<button>` element, so you can only add [phrasing content](https://html.spec.whatwg.org/#phrasing-content) to it. If `html` is provided, the `text` option will be ignored. If `text` is provided, this is not required.
```

packages/govuk-frontend/src/govuk/components/character-count/character-count.yaml
```diff
@@ -30,12 +30,12 @@ params:
   - name: label
     type: object
     required: true
-    description: Options for the label component.
+    description: Additional options for the label used by the character count component.
```
36degrees, line 33: I'm not sure about the use of 'additional' here, because to it suggests that this is optional, whereas the character count must have a label of some sort to be accessible.

(It arguably does makes sense in other places… for example in the checkboxes component where we 'promote' the `text` and `html` options, allowing the user to set the label for a checkbox using `items[].text` but provide e.g. additional classes for the label using `items[].label.classes`… API design is hard 😕)

packages/govuk-frontend/src/govuk/components/cookie-banner/cookie-banner.yaml
```diff
@@ -48,11 +48,11 @@ params:
           - name: type
             type: string
             required: false
-            description: The type of button. You can set `button` or `submit`. Set `button` and `href` to render a link styled as a button. If you set `href`, it overrides `submit`.
+            description: The type of button. You can set `"button"` or `"submit"`. Set `type` to `"button"` and set `href` to render a link styled as a button. If you set `href`, it overrides `type` `"submit"`.
```
36degrees, line 51: This isn't quite right (it wasn't before either).

https://github.com/alphagov/govuk-frontend/blob/3cd8054a7ca9abadedae6283a4da0d0b0c73b792/packages/govuk-frontend/src/govuk/components/cookie-banner/template.njk#L42-L58

It works more like this:

```mermaid
flowchart TD
    href{"Is `href` set?"}
    href-- No -->o1["Make a button of the specified `type`"]
    href-- Yes -->hb{"Is `type` set to 'button'?"}
    hb-- No -->o2["Make a link, ignore `type` entirely"]
    hb-- Yes -->o3["Make a link that looks like a button"]
```

Maybe something like?

```suggestion
            description: The type of button – `"button"` or `"submit"`. If `href` is provided, set `type` to `"button"` render a link styled as a button.
```

packages/govuk-frontend/src/govuk/components/date-input/date-input.yaml
```diff
@@ -6,11 +6,11 @@ params:
   - name: namePrefix
     type: string
     required: false
-    description: Optional prefix. This is used to prefix each `item.name` using `-`.
+    description: Optional prefix. This is used to prefix each item `name` using `-`.
```
36degrees, line 9:
```suggestion
    description: Optional prefix. This is used to prefix each item `name`, separated by `-`.
```

packages/govuk-frontend/src/govuk/components/date-input/date-input.yaml
```diff
@@ -6,11 +6,11 @@ params:
...
   - name: items
     type: array
     required: false
-    description: An array of input objects with name, value and classes.
+    description: The items of each date input component.
```
36degrees, line 13: There is only 1 date input, so it doesn't make sense to say 'each date input component'. I'm also not sure 'items' is descriptive enough to help the user understand what this is.

Maybe:

```suggestion
    description: The inputs within the date input component.
```

packages/govuk-frontend/src/govuk/components/footer/footer.yaml
```diff
@@ -2,24 +2,24 @@ params:
   - name: meta
     type: object
     required: false
-    description: Object containing options for the meta navigation.
+    description: The section to add content before the standard footer component.
```
36degrees, line 5: I think I understand the intent here – that the 'standard footer' is the copyright and license information? But adding the word _component_ implies it's the entire thing (i.e. the meta section is before the grey block that makes up the footer).

I think people's mental model of what the 'standard footer' is will differ, and some will include any navigation etc that appears within it.

Can we better describe this? Maybe something similar to what we have for some of the other sections, like 'after any navigation, before the copyright and license information'?

packages/govuk-frontend/src/govuk/components/footer/footer.yaml
```diff
@@ -2,24 +2,24 @@ params:
...
       - name: text
         type: string
         required: false
-        description: Text to add to the meta section of the footer, which will appear below any links specified using meta.items. If meta.html is specified, this option is ignored.
+        description: Text to add to the meta section of the footer, which will appear below any links specified using meta `items`. If meta `html` is specified, this option is ignored.
       - name: items
         type: array
         required: false
-        description: Array of items for use in the meta section of the footer.
+        description: The meta `items` add content within a unordered list to the meta section of the footer component. These appear above any text or custom html in the meta section.
```
36degrees, line 22: Nice 👍🏻

packages/govuk-frontend/src/govuk/components/footer/footer.yaml
```diff
@@ -49,11 +49,11 @@ params:
       - name: width
         type: string
         required: false
-        description: Width of each navigation section in the footer. Defaults to full width. You can pass any design system grid width here, for example, 'one-third'; 'two-thirds'; 'one-half'.
+        description: Width of each navigation section in the footer. Defaults to `"full"`. You can pass any design system grid width here, for example, `"one-third"`; `"two-thirds"`; `"one-half"`.
```
36degrees, line 52: Is there a reason for using semi-colons here, when we use a standard comma separated list for other sets of options? Is it because it's non-exhaustive?

If not, can we change it?

```suggestion
        description: Width of each navigation section in the footer. Defaults to `"full"`. You can pass any design system grid width here, for example, `"one-third"`, `"two-thirds"` or `"one-half"`.
```

packages/govuk-frontend/src/govuk/components/input/input.yaml
```diff
@@ -30,12 +30,12 @@ params:
   - name: label
     type: object
     required: true
-    description: Options for the label component.
+    description: Additional options for the label used by the text input component.
```
36degrees, line 33: Again, having a `label.text` or `label.html` is pretty important, so unsure about 'Additional'.

packages/govuk-frontend/src/govuk/components/radios/radios.yaml
```diff
@@ -2,12 +2,12 @@ params:
   - name: fieldset
     type: object
     required: false
-    description: Options for the fieldset component (for example legend).
+    description: Can be used to add a fieldset to the radios component.
```
36degrees, line 5: Radios always need a fieldset (unlike checkboxes where there can be only one, like 'I accept the terms and conditions') so maybe reword this to make it sound less optional?

packages/govuk-frontend/src/govuk/components/radios/radios.yaml
```diff
@@ -30,11 +30,11 @@ params:
   - name: name
     type: string
     required: true
-    description: Name attribute for each radio item.
+    description: Name attribute for all radio items.
```
36degrees, line 33: Nice attention to detail 👍🏻

packages/govuk-frontend/src/govuk/components/summary-list/summary-list.yaml
```diff
@@ -45,12 +45,12 @@ params:
...
         type: object
         required: false
-        description: The actions of each row.
+        description: The action link content for each row item in the summary list component.
         params:
           - name: items
             type: array
             required: false
-            description: Array of action item objects.
+            description: The items within the summary list component.
```
36degrees, line 53: This isn't quite right – this is the list of links to add to the row.

packages/govuk-frontend/src/govuk/components/summary-list/summary-list.yaml
```diff
@@ -109,12 +109,12 @@ params:
       - name: actions
         type: object
         required: false
-        description: The actions of each summary card.
+        description: The action link content for each summary card variant of the summary list component.
```
36degrees, line 112: I think this could be clearer in how it differs from the other `actions` option – it does relate to the summary card, but the difference is that the actions are added to the card (within the grey area at the top) rather than to the individual rows.

The way this is currently written suggests it's an alternative way of doing the same thing, if you're using the variant, and doesn't really help the user understand what the difference is.

packages/govuk-frontend/src/govuk/components/table/table.yaml
```diff
@@ -35,7 +35,7 @@ params:
   - name: head
     type: array
     required: false
-    description: Array of table head cells.
+    description: The head cells within the table component.
```
36degrees, line 38: This macro isn't great – reality is that [table header cells can appear anywhere in a table](https://www.w3.org/WAI/tutorials/tables/) but our `head` option just adds a row at the very top of the table and makes all the cells within it a header, so it's really only useful for very simple tables.

I don't think we can solve this completely with content, but we could follow the pattern we've used elsewhere to try and explain what it does?

```suggestion
    description: Can be used to add a row of table header cells (`<th>`) at the top of the table component.
```

We might also want to tweak the description used for `firstCellIsHeader` to talk about 'table headers' instead:

> If set to true, the first cell in each row will be a table header (`<th>`).

packages/govuk-frontend/src/govuk/components/tabs/tabs.yaml
```diff
@@ -14,7 +14,7 @@ params:
   - name: items
     type: array
     required: true
-    description: Array of tab items.
+    description: The items within the tabs component.
```
36degrees, line 17:
```suggestion
    description: The individual tabs within the tabs component.
```

packages/govuk-frontend/src/govuk/components/task-list/task-list.yaml
```diff
@@ -24,7 +24,7 @@ params:
       - name: hint
         type: object
         required: false
-        description: Object containing a hint for the task.
+        description: The hint for each task within the task list component.
```
36degrees, line 27:
```suggestion
        description: Can be used to add a hint to each task within the task list component.
```

36degrees (commented):
This generally looks like a great improvement, but there are a few places where there are a few inaccuracies – a few of them are existing issues.

There's also a few places where we've aligned the description to the name of the option (like talking about 'items' more) but in doing so made it less useful in terms of explaining what `items` actually means.

packages/govuk-frontend/src/govuk/components/character-count/character-count.yaml
colinrotherham: Ah we had two variations suggested in https://github.com/alphagov/govuk-design-system/pull/2929#issuecomment-1798183406

* "**Can be used to add a label…**" to set up an optional label
* "**Additional options for the label…**" to configure a mandatory label

i.e. When mandatory, it means "Additional [user] options" on top of our defaults

colinrotherham: Good point regarding the radios/checkboxes label options, there's a good [Trello card](https://trello.com/c/RxuFrKQS/19-radio-and-checkbox-item-nested-label-components-list-options-that-cant-be-used) for it

packages/govuk-frontend/src/govuk/components/input/input.yaml
colinrotherham: Good spot, this one shouldn't say "Additional…"

```suggestion
    description: Can be used to add a label to the text input component.
```

packages/govuk-frontend/src/govuk/components/cookie-banner/cookie-banner.yaml
colinrotherham: Nice one, best of both

packages/govuk-frontend/src/govuk/components/radios/radios.yaml
colinrotherham: Ah this is my fault because the template has `{% if params.fieldset %}`

Do we need to double check?

colinrotherham: It was the [**Radios** without fieldset](https://govuk-frontend-review.herokuapp.com/components/radios/without-fieldset/preview) example added in https://github.com/alphagov/govuk-frontend/commit/e060420ac7cb44b245db8fedefbbe5342d64a797

Do we need to flip it to `required: true` at the same time?

https://github.com/alphagov/govuk-frontend/blob/3cd8054a7ca9abadedae6283a4da0d0b0c73b792/packages/govuk-frontend/src/govuk/components/radios/radios.yaml#L242-L252

36degrees: This is a tricky one because there are edge cases where a user might want to manually manage the fieldset (like https://github.com/alphagov/govuk-frontend/issues/1079)

So e.g.

```
{% call govukFieldset %}
    <h2>Group of radios</h2>
    {{ govukRadios({
      // ...
    ) }}
    <h2>Another group of radios</h2>
    {{ govukRadios({
      // ...
    ) }}
{% endcall %}
```

I think it makes sense to optimise the description for the general case? In the vast majority of cases, you should provide a fieldset.

packages/govuk-frontend/src/govuk/components/character-count/character-count.yaml
36degrees: Fair enough, 'Additional' just sounds like a thing you can ignore to me 🤷🏻‍♂️. But it might just be how I'm interpreting it.

packages/govuk-frontend/src/govuk/components/radios/radios.yaml
colinrotherham: Will flip the text but keep `required: false`?

```suggestion
    description: Additional options for the fieldset used by the radios component.
```

packages/govuk-frontend/src/govuk/components/character-count/character-count.yaml
colinrotherham: You sure? We can drop the word and it looks good still

```suggestion
    description: Options for the label used by the character count component.
```

packages/govuk-frontend/src/govuk/components/footer/footer.yaml
colinrotherham: Sounds like a good suggestion for `meta`

I'll do the same with the "standard footer" mention in `navigation` too

```suggestion
    description: The meta section of the footer after any navigation, before the copyright and license information.
```

packages/govuk-frontend/src/govuk/components/footer/footer.yaml
colinrotherham: Mind if I put the default at the end like we do elsewhere?

Also adding the `–` before the values like your suggestion in https://github.com/alphagov/govuk-frontend/pull/4450#discussion_r1397385071

```suggestion
        description: Width of each navigation section in the footer. You can pass any Design System grid width here – for example, `"one-third"`, `"two-thirds"` or `"one-half"`. Defaults to `"full"`.
```

packages/govuk-frontend/src/govuk/components/input/input.yaml
colinrotherham: On second thoughts, this one has `required: true`

For mandatory fields we've added "Additional options…" to show it was in addition to the defaults

Linking back to https://github.com/alphagov/govuk-frontend/pull/4450#discussion_r1397583810 where it was also confusing

packages/govuk-frontend/src/govuk/components/summary-list/summary-list.yaml
colinrotherham: Good spot, I'll fix this one

```suggestion
            description: The action link items within the row item of the summary list component.
```

packages/govuk-frontend/src/govuk/components/summary-list/summary-list.yaml
colinrotherham: The `card` description says the "…summary card will wrap around the summary list" so how about this?

```suggestion
        description: The action link content shown in the header of each summary card wrapped around the summary list component.
```

colinrotherham:
@36degrees I've pushed up all those great suggestions

Not urgent, but we jump between "The thing" versus "Thing" which we could tackle at some point:

* **description:** Name attribute…
* **description:** The name attribute…
* **description:** The `name` attribute…

Possibly made worse now we're (mostly) code-wrapping attribute names, values etc

packages/govuk-frontend/src/govuk/components/character-count/character-count.yaml
36degrees: I think this also applies to https://github.com/alphagov/govuk-frontend/pull/4450#discussion_r1397403872 and https://github.com/alphagov/govuk-frontend/pull/4450#discussion_r1397580463 so let's either change all of them to drop 'Additional' or leave it.

Appreciate some thought's gone into it already and it might just be how I'm interpreting it.

colinrotherham: I've dropped the entire "Additional options for…" to be consistent with other options

Think this is a winner

```yaml
    description: The label used by the character count component.
```

packages/govuk-frontend/src/govuk/components/radios/radios.yaml
colinrotherham: I've dropped the entire "Additional options for…" as we've done in https://github.com/alphagov/govuk-frontend/pull/4450#discussion_r1401799402

```yaml
    description: The fieldset used by the radios component.
```

packages/govuk-frontend/src/govuk/components/input/input.yaml
colinrotherham: I've dropped the entire "Additional options for…" as we've done in https://github.com/alphagov/govuk-frontend/pull/4450#discussion_r1401799402

```yaml
    description: The label used by the text input component.
```

packages/govuk-frontend/src/govuk/components/radios/radios.yaml
colinrotherham: I've dropped the entire "Additional options for…" as we've done in https://github.com/alphagov/govuk-frontend/pull/4450#discussion_r1401799402

Noted in the commit that this one is special (worded as mandatory but not) and linked to this thread

```yaml
    description: The fieldset used by the radios component.
```

colinrotherham:
Thanks @36degrees for all the comments, this is ready to review now

* See commit https://github.com/alphagov/govuk-frontend/pull/4450/commits/849cacb197d57a75a35f4fd67a11777f6bf61581 to remove “**Additional options for…**” from mandatory params
* See commit https://github.com/alphagov/govuk-frontend/pull/4450/commits/8113c4f3b95047862ef7be44268262480f93eb24 with added “**Can be used to…**” to more optional nested components

36degrees (approved):
Some really nice improvements here, thanks @colinrotherham and @CharlotteDowns! 👏🏻

PR #4462, opened 2023-11-14
Fix various bugs in Review app

colinrotherham:
Fixes a few bugs and issues such as:

### Component template `*.njk` watch + reload is broken

Rendered components are cached in JSON and since fca508c3f5e678c76b320e281caa86985a5bd5c4 only `*.yaml` changes will rebuild them

### Routes throw errors for unknown params

We'd missed a ~`next()`~ `return next()` and didn't check Express.js parameters before use as shown in:

https://govuk-frontend-review.herokuapp.com/components/fdsfdsfds
https://govuk-frontend-review.herokuapp.com/components/accordion/gdfgdfgd/preview
https://govuk-frontend-review.herokuapp.com/examples/fdsfdsfsd
https://govuk-frontend-review.herokuapp.com/full-page-examples/fdhgdfga

### Removing unnecessary error handling

This is handled automatically already in the Nunjucks view engine

```patch
- res.render('all-components', function (error, html) {
-   if (error) {
-     next(error)
-   } else {
-     res.send(html)
-   }
- })
+ res.render('all-components')
```

### Removing unnecessary component filtering

We already loop examples in Nunjucks and don't need to keep assigning `res.locals.componentsFixtures`

```patch
- res.locals.componentsFixtures = componentsFixtures.map(
-   (componentFixtures) => {
-     const defaultFixture = componentFixtures.fixtures.find(
-       ({ name }) => name === 'default'
-     )
-     return {
-       ...componentFixtures,
-       fixtures: [defaultFixture]
-     }
-   }
- )

- res.render('all-components')
+ res.render('all-components', {
+  componentsFixtures
+ })
```

### Moving the [View all components](https://govuk-frontend-review.herokuapp.com/components/all) page to `/components`

I've moved and redirected ~`/components/all`~ `/components` to ensure the Express.js `:componentName` parameter is never populated as `res.locals.componentName === 'all'`

PR #4463, opened 2023-11-14
Add error pages to Review app

colinrotherham:
This PR adds Express.js default routes to handle errors

This ensures our own Review app tests correctly catch issues since a few URLs currently crash:

>https://govuk-frontend-review.herokuapp.com/components/fdsfdsfds
>https://govuk-frontend-review.herokuapp.com/components/accordion/gdfgdfgd/preview
>https://govuk-frontend-review.herokuapp.com/examples/fdsfdsfsd
>https://govuk-frontend-review.herokuapp.com/full-page-examples/fdhgdfga

These are handled in the following PR but lacked error pages:

* https://github.com/alphagov/govuk-frontend/pull/4462

Error 500 page

### DIFF
--- a/packages/govuk-frontend-review/src/app.mjs
+++ b/packages/govuk-frontend-review/src/app.mjs
@@ -80,6 +80,55 @@ export default async () => {
   // Configure nunjucks
   const env = nunjucks.renderer(app)
 
+  // Define parameters
+
+  /**
+   * Handle parameter :componentName
+   *
+   * Finds all component fixtures and default example
+   */
+  app.param('componentName', (req, res, next, componentName) => {
+    const exampleName = 'default'
+
+    // Find all fixtures for component
+    const componentFixtures = componentsFixtures.find(
+      ({ component }) => component === componentName
+    )
+
+    // Find default fixture for component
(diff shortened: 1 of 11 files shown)


packages/govuk-frontend-review/src/app.mjs
```diff
@@ -80,10 +80,81 @@ export default async () => {
...
 
+  // Define parameters
+
+  /**
+   * Handle parameter :componentName
+   *
+   * Finds all component fixtures and default example
+   */
+  app.param(
+    'componentName',
```
colinrotherham, line 91: I've added Express.js `app.param()` to follow up code review feedback in https://github.com/alphagov/govuk-frontend/pull/4424#discussion_r1385127501

romaricpascal (approved):
That'll be helpful for debugging 🙌🏻 Only little nitpick would be with the naming of the error templates. With the Design System site and Frontend docs using `404.html`, for example, thinking it'd be good to align and just have `404.njk` and `500.njk`. Happy to approve in this state to speed up merging after renaming (and if there's a technical blocker to having them named with just the HTTP status, very fine to keep them as they are now).

PR #4491, opened 2023-11-23
[WIP][SPIKE] SupportError updates

### LINKED ISSUE #4477: Investigate improvements for `SupportError` message
## What

Run a technical spike to see if we can improve the messaging of our `SupportError` message so it better helps users in browsers where it is supported, but the `govuk-frontend-supported` class has not been added to the body (for example, because the snippet adding it is still the one from v4).

## Why

Reading that GOV.UK Frontend is not supported in browsers where it is actually supported is misleading and does not help users fix their issues.

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [ ] We've identified a way to improve the `SupportError` message

domoscargin:
Closes #4477

## What and why

We want to improve our `SupportError` messages, since plenty of users in the research session were getting a "GOV.UK Frontend is not supported" message when the only issue was they hadn't added the `.govuk-frontend-supported` class to the body.

### What SHOULD trigger the "not supported" message?
It's actually very rare that we want to trigger a "not supported" message. Any browser which doesn't support ES-module simply won't run the JavaScript, so the only folks who see this error are:

1. Services that skip adding `.govuk-frontend-supported` to the `<body>`
2. Safari 10.1 (because it supports es-module and runs our code, but doesn't support `noModule`, which we use in a check and thus fails)

So it looks like only Safari 10.1 users should see "GOV.UK Frontend is not supported".

### getSupportedLevelMessage

For this spike, I've added a `getSupportedLevelMessage` in place of `isSupported`. Where `isSupported` returned true or false depending on whether there was support, `getSupportedLevelMessage` runs checks to return an appropriate message by these cases:

- `OK`, by default
- `NO_MODULE`: when GOV.UK Frontend is initialised without `<script type="module">`
- `MISSING_CLASS`: when the `.govuk-frontend-supported` class has not been added to `<body>`
- `NO_SUPPORT`: for all other cases (ie: Safari 10.1) where there is no support

### SupportError

This has been simplified to just take a message like any other error.

## TO DO
- I'm not sure how to test in jest for browsers which don't support `noModule`
- We might want to protect browsers which don't support `noModule` and let them run GOV.UK Frontend anyway? In which case, we wouldn't even need the "not supported" message at all.
- Would we prefer to `warn` rather than `error` in some cases? If no other conditions trigger except that the `.govuk-frontend-supported` class is missing, for example, then everything should function just fine. If so, we can do a little logic based on `getSupportedLevelMessage`'s return value.
- Tests need updating - we're currently only testing for the case where `.govuk-frontend-supported` is removed, and there are a few errors in other tests I haven't had time to look at.
- Do we want to give a nice console message to services which don't load the JavaScript saying "You're fine, everything works, you just don't have our JavaScript enhancements"?

### DIFF
--- a/packages/govuk-frontend/src/govuk/common/index.mjs
+++ b/packages/govuk-frontend/src/govuk/common/index.mjs
@@ -134,21 +134,33 @@ export function getFragmentFromUrl(url) {
 }
 
 /**
- * Checks if GOV.UK Frontend is supported on this page
- *
- * Some browsers will load and run our JavaScript but GOV.UK Frontend
- * won't be supported.
  *
  * @internal
  * @param {HTMLElement | null} [$scope] - HTML element `<body>` checked for browser support
- * @returns {boolean} Whether GOV.UK Frontend is supported on this page
+ * @returns {string} Support level
  */
-export function isSupported($scope = document.body) {
-  if (!$scope) {
-    return false
+export function getSupportedLevelMessage($scope = document.body) {
+  const LEVELS = {
+    OK: 'OK',
+    NO_SUPPORT: 'GOV.UK Frontend is not supported in this browser',
+    NO_MODULE: 'GOV.UK Frontend initialised without `<script type="module">`',
+    MISSING_CLASS: '<body> tag is missing the `govuk-frontend-supported` class'
   }
+  let supportedLevel = LEVELS.OK
 
-  return $scope.classList.contains('govuk-frontend-supported')
+  // The currentScript property only exists on <script> tags with no module.
+  // $scope will be null if `initAll` was called from the <head> with no module
+  if (document.currentScript ?? !$scope) {
+    supportedLevel = LEVELS.NO_MODULE
+  } else if (!document.body.classList.contains('govuk-frontend-supported')) {
+    supportedLevel = LEVELS.MISSING_CLASS
+  } else if (!('noModule' in HTMLScriptElement.prototype)) {
+    // For browsers that support es-module, but not noModule
+    supportedLevel = LEVELS.NO_SUPPORT
+  }
+  // All other browsers either don't load the JavaScript, or fully support GOV.UK Frontend
+
+  return supportedLevel
 }
 
 /**
--- a/packages/govuk-frontend-review/src/views/examples/javascript-head-error/index.njk
+++ b/packages/govuk-frontend-review/src/views/examples/javascript-head-error/index.njk
@@ -1,0 +1,39 @@
+{% extends "layouts/examples.njk" %}
+
+{% from "govuk/components/back-link/macro.njk" import govukBackLink %}
+{% from "govuk/components/character-count/macro.njk" import govukCharacterCount %}
+
+{% block head %}
+  <link rel="stylesheet" href="/stylesheets/app.min.css">
+
+  {% block styles %}{% endblock %}
+  <script src="/javascripts/all.bundle.js"></script>
+  <script>
+    window.GOVUKFrontend.initAll()
+  </script>
+{% endblock %}
+
+{% block beforeContent %}
+  {{ govukBackLink({
+    href: "/"
+  }) }}
+{% endblock %}
+
+{% block content %}
+  <h1 class="govuk-heading-l">JavaScript in &lt;head&gt; error</h1>
+  <p class="govuk-body-lead">
+    Open your <a class="govuk-link" href="https://developer.mozilla.org/en-US/docs/Web/API/console#see_also">brower's console</a>
+    to witness the error thrown in browsers where GOV.UK Frontend is not supported.
+  </p>
+  {{ govukCharacterCount({
+    label: {
+      text: "Any other comment"
+    },
+    id: "some-content",
+    name: "some-content",
+    maxlength: "100"
+  })}}
+{% endblock %}
+
+{% block bodyEnd %}
+{% endblock %}
--- a/packages/govuk-frontend/src/govuk/all.mjs
+++ b/packages/govuk-frontend/src/govuk/all.mjs
@@ -1,7 +1,7 @@
 /* eslint-disable no-new */
 
 import { version } from './common/govuk-frontend-version.mjs'
-import { isSupported } from './common/index.mjs'
+import { getSupportedLevelMessage } from './common/index.mjs'
 import { Accordion } from './components/accordion/accordion.mjs'
 import { Button } from './components/button/button.mjs'
 import { CharacterCount } from './components/character-count/character-count.mjs'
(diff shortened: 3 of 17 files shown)


packages/govuk-frontend/src/govuk/common/index.mjs
```diff
@@ -134,21 +134,33 @@ export function getFragmentFromUrl(url) {
...
+
+  // The currentScript property only exists on <script> tags with no module.
+  // $scope will be null if `initAll` was called from the <head> with no module
+  if (document.currentScript ?? !$scope) {
+    supportedLevel = LEVELS.NO_MODULE
+  } else if (!document.body.classList.contains('govuk-frontend-supported')) {
+    supportedLevel = LEVELS.MISSING_CLASS
+  } else if (!('noModule' in HTMLScriptElement.prototype)) {
+    // For browsers that support es-module, but not noModule
+    supportedLevel = LEVELS.NO_SUPPORT
```
romaricpascal, line 159: **issue** I think this needs to go before the check for `govuk-frontend-supported`, as if the `noModule` check fails, the class won't be there. So in the current order, the previous `if` would catch both the lack of support and the lack of having added the `<script>` snippet to the `<body>`.

packages/govuk-frontend/src/govuk/common/index.mjs
```diff
@@ -134,21 +134,33 @@ export function getFragmentFromUrl(url) {
...
+    OK: 'OK',
+    NO_SUPPORT: 'GOV.UK Frontend is not supported in this browser',
+    NO_MODULE: 'GOV.UK Frontend initialised without `<script type="module">`',
+    MISSING_CLASS: '<body> tag is missing the `govuk-frontend-supported` class'
+  }
+  let supportedLevel = LEVELS.OK
+
+  // The currentScript property only exists on <script> tags with no module.
+  // $scope will be null if `initAll` was called from the <head> with no module
+  if (document.currentScript ?? !$scope) {
```
36degrees, line 153: I think this means that GOV.UK Frontend will now 'refuse to run' if it's loaded without using `type="module"`?

That might be desired, but we should make that change consciously.

romaricpascal:
Overall looks like a neat path to get some more info on why our components are not initialising 🙌🏻 

> * I'm not sure how to test in jest for browsers which don't support `noModule`

Yeah, with Jest running Chrome through Puppeteer, which supports `noModule`,  that's tricky. And we don't even have access to Jest mocking capabilities as those run in the test code (Node) while the test for `noModule` runs in the browser 😔 Only thing I'm thinking of is a blunt `delete HTMLScriptElement.prototype.noModule` before that particular test but: 
1. It's super blunt
2. Browsers might not even allow deleting a property from the prototype of a native class.
 
> * We might want to protect browsers which don't support `noModule` and let them run GOV.UK Frontend anyway? In which case, we wouldn't even need the "not supported" message at all.

Do you mean that as long as GOV.UK Frontend loads as a `module`, we'd be good to let it run through? I don't think it's desirable as the CSS is configured to be set up for JavaScript based on the presence of `noModule`. If we let it run for all browsers running it as a module, that'd mean either:
1. JavaScript running but the CSS not expecting it
2. JavaScript adding the `govuk-frontend-supported` class to make the CSS aware it's running, which would mean a flash of the components styled for not having the JavaScript while the JavaScript loads. 

> * Would we prefer to `warn` rather than `error` in some cases? If no other conditions trigger except that the `.govuk-frontend-supported` class is missing, for example, then everything should function just fine. If so, we can do a little logic based on `getSupportedLevelMessage`'s return value.

We could just `warn` when not loaded as a `module`, which is a bit softer than halting everything by throwing an error. My reasoning is that we have error that prevent our code to run if we're in a browser that's actually not supporting us. Using `type="module"` is more of a performance optimisation to not load in the browsers that do not support us, things will still run fine if the script is loaded as a classic `<script>`, hence why we could be softer on that point.

> * Tests need updating - we're currently only testing for the case where `.govuk-frontend-supported` is removed, and there are a few errors in other tests I haven't had time to look at.

Good spot. Looks like JSDom doesn't support `noModule`... could be an out for your first point: unit test the `getSupportedLevelMessage` with JSDom to check it returns the right message for each situation as it might be easier to mock things there and just check in Puppeteer that the method gets exercised by running without the `govuk-frontend-supported` class.

> * Do we want to give a nice console message to services which don't load the JavaScript saying "You're fine, everything works, you just don't have our JavaScript enhancements"?

What do you mean by "services that don't load JavaScript"? Is it for people whose browsers don't support `noModule` or is it for services that would actually not add our JavaScript at all? Overall, seems more like a documentation job (reassuring that things are fine for users not running JavaScript) than having some code for logging.

packages/govuk-frontend/src/govuk/common/index.mjs
domoscargin: Good spot! Need to figure out a better test for the logic here.

packages/govuk-frontend/src/govuk/common/index.mjs
colinrotherham: Older browsers that _could_ potentially see "**GOV.UK Frontend is not supported in this browser**" will probably crash on ES2015+ syntax anyway and won't reach the `isSupported()` check 😭

(Other than Safari 10.1)

But I love the idea of also running the `!('noModule' in HTMLScriptElement.prototype)` check in our JavaScript so we can swap out the error message for services that skip the new snippet

Could we just do the error message switching in **src/govuk/errors/index.mjs**?

```mjs
const supportMessage =
  'noModule' in HTMLScriptElement.prototype
    ? 'GOV.UK Frontend is missing `<body class="govuk-frontend-supported">` via inline `<script>`'
    : 'GOV.UK Frontend is not supported in this browser'

super(
  $scope
    ? supportMessage
    : 'GOV.UK Frontend initialised without `<script type="module">`'
)
```

domoscargin: It's a great point - think this would mess with folks who use UMD, for example?

domoscargin:
> Only thing I'm thinking of is a blunt `delete HTMLScriptElement.prototype.noModule` before that particular test but:
> 
> 1. It's super blunt
> 2. Browsers might not even allow deleting a property from the prototype of a native class.

Worth giving a go! We only want to test the behaviour, after all, not the context leading to that behaviour.

> I don't think it's desirable as the CSS is configured to be set up for JavaScript based on the presence of `noModule`. If we let it run for all browsers running it as a module, that'd mean either:

Good shout - hadn't considered the CSS.

> We could just `warn` when not loaded as a `module`

Agreed - similar point from @36degrees as well. Can tweak this.

> unit test the `getSupportedLevelMessage` with JSDom to check it returns the right message for each situation as it might be easier to mock things there and just check in Puppeteer that the method gets exercised by running without the `govuk-frontend-supported` class.

Yeah, this is the approach I would lean towards.

> What do you mean by "services that don't load JavaScript"? 

In this case, services which just don't add the JavaScript at all. @colinrotherham and I thought it would be nice to assure folks in certain older browsers (IE11, for example) that they're all good, functionally.

domoscargin:
Closing in favour of #4492 - we can revisit this spike if/when components also need more granular support levels.

PR #4492, opened 2023-11-23
Improve SupportError when `<body class="govuk-frontend-supported">` is not set

### LINKED ISSUE #4477: Investigate improvements for `SupportError` message

domoscargin:
Closes #4477

## What and why

We want to improve our `SupportError` messages, since plenty of users in the research session were getting a "GOV.UK Frontend is not supported" message when the only issue was they hadn't added the `.govuk-frontend-supported` class to the body.

### What SHOULD trigger the "not supported" message?
It's actually very rare that we want to trigger a "not supported" message. Any browser which doesn't support ES-module simply won't run the JavaScript, so the only folks who see this error are:

1. Services that skip adding `.govuk-frontend-supported` to the `<body>`
2. Safari 10.1 (because it supports es-module and runs our code, but doesn't support `noModule`, which we use in a check and thus fails)

So it looks like only Safari 10.1 users should see "GOV.UK Frontend is not supported".

This PR takes @colinrotherham's suggestion of sticking with the simpler option of more granular error messages, revisiting https://github.com/alphagov/govuk-frontend/pull/4491 if/when components need different levels of support.

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/accordion/accordion.puppeteer.test.js
+++ b/packages/govuk-frontend/src/govuk/components/accordion/accordion.puppeteer.test.js
@@ -716,7 +716,7 @@ describe('/components/accordion', () => {
           examples = await getExamples('accordion')
         })
 
-        it('throws when GOV.UK Frontend is not supported', async () => {
+        it('can throw a SupportError if appropriate', async () => {
           await expect(
             render(page, 'accordion', examples.default, {
               beforeInitialisation() {
@@ -726,7 +726,8 @@ describe('/components/accordion', () => {
           ).rejects.toMatchObject({
             cause: {
               name: 'SupportError',
-              message: 'GOV.UK Frontend is not supported in this browser'
+              message:
+                'GOV.UK Frontend initialised without `<body class="govuk-frontend-supported">` from template `<script>` snippet'
             }
           })
         })
--- a/packages/govuk-frontend/src/govuk/errors/index.jsdom.test.mjs
+++ b/packages/govuk-frontend/src/govuk/errors/index.jsdom.test.mjs
@@ -12,6 +12,11 @@ describe('errors', () => {
   })
 
   describe('SupportError', () => {
+    beforeEach(() => {
+      // JSDOM hasn't yet implemented `noModule`, so we have to mock this
+      window.HTMLScriptElement.prototype.noModule = true
+    })
+
     it('is an instance of GOVUKFrontendError', () => {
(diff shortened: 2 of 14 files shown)


packages/govuk-frontend/src/govuk/components/accordion/accordion.puppeteer.test.js
```diff
@@ -726,7 +726,8 @@ describe('/components/accordion', () => {
           ).rejects.toMatchObject({
             cause: {
               name: 'SupportError',
-              message: 'GOV.UK Frontend is not supported in this browser'
+              message:
+                'GOV.UK Frontend is missing `<body class="govuk-frontend-supported">` via inline `<script>`'
```
colinrotherham, line 730: We should get feedback from @claireashworth too but I'm changing my own suggestion 🙈

```suggestion
                'GOV.UK Frontend initialised without `<body class="govuk-frontend-supported">` from template `<script>` snippet'
```

packages/govuk-frontend/src/govuk/errors/index.jsdom.test.mjs
```diff
@@ -21,11 +21,23 @@ describe('errors', () => {
...
+      // something like this
+      delete window.HTMLScriptElement.prototype.noModule
       expect(new SupportError(document.body).message).toBe(
         'GOV.UK Frontend is not supported in this browser'
       )
     })
 
+    it('provides feedback when <body> class is missing', () => {
+      // JSDOM hasn't yet implemented `noModule`, so we have to mock this
+      window.HTMLScriptElement.prototype.noModule = true
```
colinrotherham, line 35: Worth putting these lines into a `beforeEach()` so it's always reset anyway?

colinrotherham (approved):
Looking good! Some minor comments and should confirm error wording

colinrotherham:
Just going to give this a rebase with the new logo SVG (so Percy checks pass)

romaricpascal (approved):
Code is looking good 🙌🏻 @claireashworth does the wording read alright?

claireashworth (approved):

PR #4494, opened 2023-11-24
Fix missed `actions/setup-node` update

colinrotherham:
Quick PR to update an action missed by Dependabot:

https://github.com/alphagov/govuk-frontend/blob/6b93b3667b67a0932c9959ca054ae6fc77d63d9f/.github/workflows/actions/setup-node/action.yml#L13-L14

Possibly because we ignored it in the past?

* https://github.com/alphagov/govuk-frontend/pull/3936


romaricpascal (approved):
Good spot! 🦅

PR #4498, opened 2023-11-24
Fix header product name wrapping onto new line

colinrotherham:
This PR fixes https://github.com/alphagov/govuk-frontend/issues/1502

See the "Service Manual" topic page for an updated example:
https://govuk-frontend-pr-4498.herokuapp.com/full-page-examples/service-manual-topic

## Before

We float the logo and apply `width: 33.33%`

Before

## After

When the logo matches `:last-child` (no navigation or service name) we can reset the float

After


romaricpascal (approved):
Examples look fine [on the review app](https://govuk-frontend-pr-4498.herokuapp.com/components/header/) 🙌🏻 ⛵

PR #4500, opened 2023-11-24
Organise and add tests for Review app "Other examples"

colinrotherham:
This PR adds the same tests for "[Other examples](https://govuk-frontend-review.herokuapp.com)" as done in https://github.com/alphagov/govuk-frontend/pull/4499

Makes sure none of our examples stop working unexpectedly


romaricpascal (approved):
Well, that was much simpler to review than the Full page ones 😆 All good to go ⛵

Issue #4147, opened 2023-08-30
File sizes commented on pull requests didn't vary between two PRs


romaricpascal:
The comment of the file stats reported the same file sizes on [the PR that disabled mangling of our class and function names](https://github.com/alphagov/govuk-frontend/pull/4145#issuecomment-1697609663) and [the last one merged on main](https://github.com/alphagov/govuk-frontend/pull/4142#issuecomment-1695392549).

Running build:release on main and my branch locally does give me different file sizes (and a 10K smaller number than on the stats PR):

```
Terser options updated
-rw-r--r--   1 romaric.pascal  staff    35K Aug 30 09:31 govuk-frontend-4.6.0.min.js

main
-rw-r--r--   1 romaric.pascal  staff    34K Aug 30 09:56 govuk-frontend-4.6.0.min.js
```

romaricpascal:
One lead could be the `checkout` action using `main` instead of the actual branch the PR relates to: https://github.com/actions/checkout/issues/1418.

romaricpascal:
Started [a little branch](https://github.com/alphagov/govuk-frontend/pull/4149) for investigation, logging the SHA that are checked out in the different scripts.

Looking at the logs (expanding "Checking out the ref" in the "Checkout code" step), [the stats job checks out the same ref](https://github.com/alphagov/govuk-frontend/actions/runs/6025532455/job/16346558695?pr=4149) as [the ref that the build checks](https://github.com/alphagov/govuk-frontend/actions/runs/6025532455/job/16346483833?pr=4149) out and the rest of out actions (the SHA referenced by `github.sha` which is the merge of the PR head with the branch it'll be merged into).

To be continued...

romaricpascal:
Looks like our workflows are only building the npm package and not the files for our release archive. That would explain the sameness of the stats, as well as the 10K difference between local and CI.

Two potential fixes:
1. Updating [the files being built](https://github.com/alphagov/govuk-frontend/blob/d71e45f3438868017fddd85fdec97116fbce1f5e/.github/workflows/actions/build/action.yml#L30-L31) and [the cache paths](https://github.com/alphagov/govuk-frontend/blob/d71e45f3438868017fddd85fdec97116fbce1f5e/.github/workflows/actions/build/action.yml#L17-L20) so we build `dist` as well.

  ```
  npm run build --workspaces --if-present --ignore-scripts
  npm run build:release --workspaces --if-present --ignore-scripts
  ```
[...]

PR #4510, opened 2023-11-27
Build GitHub release in “Generate stats” workflow

colinrotherham:
Fixes https://github.com/alphagov/govuk-frontend/issues/4147 by running `npm run build:release` in GitHub Actions

This now means we see the PR stats impact on [committed files in **./dist**](https://github.com/alphagov/govuk-frontend/tree/main/dist) too

## Development build output
Because [Node.js assumes a development environment by default](https://nodejs.org/en/learn/getting-started/nodejs-the-difference-between-development-and-production) I've ensured Babel, PostCSS and Rollup make their `NODE_ENV=development` version numbers and filenames obvious in the PR stats comment:

```patch
- const version="5.0.0-beta.2";function mergeConfigs(...t){function flattenObject(t){const e={};return function flattenLoop(t,i){
+ const version="development";function mergeConfigs(...t){function flattenObject(t){const e={};return function flattenLoop(t,i){
```

### Before

GitHub release committed assets use version numbers

```console
dist/govuk-frontend-5.0.0-beta.2.min.css
dist/govuk-frontend-5.0.0-beta.2.min.js
```

### After

GitHub release development builds use `NODE_ENV`

```console
dist/govuk-frontend-development.min.css
dist/govuk-frontend-development.min.js
```

### DIFF
--- a/.github/workflows/stats-comment.yml
+++ b/.github/workflows/stats-comment.yml
@@ -20,8 +20,10 @@ jobs:
       - name: Restore dependencies
         uses: ./.github/workflows/actions/install-node
 
-      - name: Build
-        uses: ./.github/workflows/actions/build
+      - name: Build npm package and GitHub release
+        run: |
+          npm run build:package --workspace govuk-frontend
+          npm run build:release --workspace govuk-frontend
 
       - name: Add comment to PR
         uses: actions/github-script@v7.0.1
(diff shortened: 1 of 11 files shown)


domoscargin (approved):
Looks good - doesn't seem to affect the build time much, if at all.

.github/workflows/stats-comment.yml
```diff
@@ -20,21 +20,26 @@ jobs:
       - name: Restore dependencies
         uses: ./.github/workflows/actions/install-node
 
-      - name: Build
-        uses: ./.github/workflows/actions/build
+      - name: Build npm package and GitHub release
+        run: |
+          npm run build:package --workspace govuk-frontend
+          npm run build:release --workspace govuk-frontend
```
domoscargin, line 26: 👍

domoscargin (approved):

PR #4516, opened 2023-11-29
Fix shared task `styles.compile()` when loadPaths don’t exist yet

colinrotherham:
This PR fixes a bug where a newly cloned repo can't run:

```
npm install
npm run build:release
```

I've tweaked a bit of awful Prettier formatting and added some better comments too

## Problem

For our [Sass `loadPaths`](https://sass-lang.com/documentation/js-api/interfaces/options/#loadPaths) we use `require.resolve()` via [`packageResolveToPath()`](https://github.com/alphagov/govuk-frontend/blob/2cdcb65570aba21e9babb2b692e3585aa606fac0/shared/lib/names.js#L75) to locate `govuk-frontend`

But Node.js throws “Cannot find module” since `npm run build:package` hasn't been run yet

## Solution

We can use the ~`packageResolveToPath()`~ [`packageTypeToPath()`](https://github.com/alphagov/govuk-frontend/blob/2cdcb65570aba21e9babb2b692e3585aa606fac0/shared/lib/names.js#L109) helper instead

This helper was added when Node.js v17 couldn't use `require.resolve()` in https://github.com/alphagov/govuk-frontend/issues/3755

### DIFF
--- a/shared/lib/names.js
+++ b/shared/lib/names.js
@@ -72,7 +72,9 @@ function componentNameToConfigName(componentName) {
  * @param {Pick<PackageOptions, "modulePath" | "moduleRoot">} [options] - Package resolution options
  * @returns {string} Path to installed npm package entry
  */
-function packageResolveToPath(packageEntry, { modulePath, moduleRoot } = {}) {
+function packageResolveToPath(packageEntry, options = {}) {
+  const { modulePath, moduleRoot } = options
+
   const packagePath = require.resolve(packageEntry, {
     paths: [moduleRoot ?? paths.root]
   })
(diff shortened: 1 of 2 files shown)


shared/lib/names.js
```diff
@@ -91,34 +105,29 @@ function packageResolveToPath(packageEntry, { modulePath, moduleRoot } = {}) {
...
  * @returns {string} Path to installed npm package field
  */
-function packageTypeToPath(
-  packageName,
-  { modulePath, moduleRoot, type = 'commonjs' } = {}
-) {
+function packageTypeToPath(packageName, options = {}) {
+  const { modulePath, moduleRoot, type = 'commonjs' } = options
+
+  // Assume package.json is always resolvable
```
romaricpascal, line 115: **question** Can a package prevent access to its `package.json`? If not, can we drop the "Assume", please, to avoid confusion when we read back this code in the future.

romaricpascal (approved):
Good spot, neat we can work around not having built yet with `packageTypetoPath` ⛵

shared/lib/names.js
colinrotherham: Yeah sadly it can, so this is the correct comment 😔

With package exports you can quite easily [lock yourself out from **package.json** access](https://github.com/isaacs/node-glob/pull/503#issuecomment-1450874010)

So you'll often see it as a separate entry point:
https://github.com/isaacs/tshy?tab=readme-ov-file#exports

PR #4520, opened 2023-11-30
Skip or delete PR diff comments when empty

colinrotherham:
This PR addresses some of the review comments in https://github.com/alphagov/govuk-frontend/pull/4511#pullrequestreview-1754871951

## Skipping empty GitHub diff comments

We no longer post "No diff changes found." comments, except for:

1. **JavaScript changes to GitHub release**
2. **Stylesheets changes to GitHub release**

This reduces comment noise for day-to-day PRs but keeps the important two for GitHub releases

Empty comments

## Deleting outdated GitHub diff comments

What if a PR had a diff and a comment was added

But another push results in no diff, so we'd be left with an outdated comment that was already posted

With the comment now out of date, we now remove it via [GitHub REST API `issues.deleteComment`](https://octokit.github.io/rest.js/v20#issues-delete-comment)

### DIFF
--- a/.github/workflows/diff-change-to-dist.yaml
+++ b/.github/workflows/diff-change-to-dist.yaml
@@ -68,7 +68,8 @@ jobs:
               {
                 path: '${{ github.workspace }}/.cache/diff/dist/other.diff',
                 titleText: 'Other changes to GitHub release',
-                markerText: 'dist/other.diff'
+                markerText: 'dist/other.diff',
+                skipEmpty: true
               }
             ]
 
--- a/.github/workflows/diff-change-to-package.yaml
+++ b/.github/workflows/diff-change-to-package.yaml
@@ -58,17 +58,20 @@ jobs:
               {
                 path: '${{ github.workspace }}/.cache/diff/package/js.diff',
                 titleText: 'JavaScript changes to npm package',
-                markerText: 'package/js.diff'
+                markerText: 'package/js.diff',
+                skipEmpty: true
               },
               {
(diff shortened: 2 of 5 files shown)


romaricpascal (commented):
Cheers for digging into solutions for the comments on that PR 😊  

I think there's a mix of things in this PR and we may want to split it into two:
1. Adding comments in a consistent order sounds fairly uncontroversial and could go on its own
2. Skipping or deleting when empty could use a bigger discussion, as we may still want a "no diff to output" comment that would reassure us that the diff did actually run (and that there wasn't a bug despite the workflow completing OK, for example).

In terms of code, the steps taken look generally fine. I'd be keen to keep the `<!--` and `-->` wrapping the marker text inside the functions creating the comments (and now searching/deleting it), as it provides some safety that the marker won't be visible in the comment once posted.

colinrotherham:
@romaricpascal Thanks

* Diff comment order moved to https://github.com/alphagov/govuk-frontend/pull/4523
* Marker text `<!--` and `-->` abstraction is hidden away again (pushed)

>…Skipping or deleting when empty could use a bigger discussion, as we may still want a "no diff to output" comment that would reassure us that the diff did actually run (and that there wasn't a bug despite the workflow completing OK, for example).

☝️ Sounds good, let me know. I was happy with the noise at first but now they're everywhere 😱

romaricpascal (approved):
Let's reduce the noise for now, and think about how to get some certainty that things have ran OK in a future PR. I'll create an issue to track that.

Thanks for updating our set up 😊 ⛵

PR #4523, opened 2023-12-01
Add PR diff comments in order

colinrotherham:
This PR addresses one of the review comments in https://github.com/alphagov/govuk-frontend/pull/4511#pullrequestreview-1754871951

## Posting GitHub diff comments in order

We previously used [`Promise.allSettled()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled) but the least useful "Other changes" comment often comes first

>- I wonder if it's worth rethinking the grouping down the line - feels _slightly_ strange to see CSS and SCSS changes split up on #4498 , especially since they directly relate to each other. That could be a whole spiral of work though.

This PR now adds comments in the order they're configured

### DIFF
--- a/.github/workflows/scripts/comments.mjs
+++ b/.github/workflows/scripts/comments.mjs
@@ -1,5 +1,5 @@
 import { readFile } from 'node:fs/promises'
-import { join } from 'path'
+import { basename, join } from 'path'
 
 import { getFileSizes } from '@govuk-frontend/lib/files'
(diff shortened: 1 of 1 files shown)


romaricpascal (approved):
Neat! Thanks for splitting 😊 ⛵

romaricpascal:
Whoops, little linting issue before we can merge it seems. Strange that Husky didn't pick it up, but maybe it's because the commits were moved with rebases 🤔

Issue #2501, opened 2022-01-17
Duplicated JavaScript code across components


vanitabarrett:
### Cause

Some of our components have very similar/the same JavaScript functionality. A recent example is when we copied some of the error summary JavaScript to add to the Skip Link component (https://github.com/alphagov/govuk-frontend/pull/2450/commits/291066469f248b696d26e6baea02ff5f76e17e93).

### Consequences

There are some positives and negatives of this.

Negatives:

- If we want to change the way a function works or fix a bug, we might make the change in one place and forget to make it in the other places it's been copied to
- It's a slight maintenance overhead
- Slight performance implications
- Causes some visibility issues - it's not clear where functions have been implemented/used before, so sometimes we end up building things from scratch without realising one component already does it

Positives:

- We have 3 times as much CSS as JavaScript at the moment, so it's not a big performance issue
- Makes it easier for us to judge the result of any changes to these functions because we know exactly where they're being used

### Impact of debt

Low

### Effort to pay down

Low

### Reason (effort to pay down)

We think we'd want to have more confidence in our JavaScript testing strategy before making these changes. Sharing functions between components means we might be more likely to introduce unintended changes to component behaviour without realising. We should make sure our testing coverage is good enough to pick up on these.

### Overall rating

Low

colinrotherham:
We only really have had big duplicated things:

1. Different `getFragmentFromUrl()` implementations across **Error summary**, **Skip link** and **Tabs**
2. Different `setFocus()` implementations across **Error summary**, **Notification banner**, **Skip link**

We fixed 1) in https://github.com/alphagov/govuk-frontend/pull/4320 and I'll open quick PR for 2)

Issue #4465, opened 2023-11-14
Add content to Skip link guidance about problem caused by preservation of fragment identifiers between pages


tombye:
The skip link component adds a `#main-content` fragment identifier to the current URL. This can persist across pages causing an unexpected jump to the main content area on load. I'd like to add a warning about this to the guidance, along with tips about ways teams could write their code to avoid this. 

## Related component

https://design-system.service.gov.uk/components/skip-link/

## Context

This problem has been seen to happen in the following scenario:
- the skip link was used, thereby adding `#main-content` to the current URL
- you leave the page via a successful form submission
- the form doesn't set the `action` attribute and so POSTs to the current URL
- the server uses [the POST -> Redirect -> GET pattern](https://en.wikipedia.org/wiki/Post/Redirect/Get) to navigate to the next page

Given all that, the browser will preserve `#main-content` in the new URL[^1] causing the next page to jump to the start of the main content area when it loads.

This was picked up as an issue by the Digital Accessibility Centre (DAC) in a recent audit of GOV.UK Notify[^2], who stated that it failed [the WCAG 2.4.3 (Focus Order) success criteria](https://www.w3.org/WAI/WCAG20/Understanding/focus-order).

Given [the POST -> Redirect -> GET pattern](https://en.wikipedia.org/wiki/Post/Redirect/Get) is so commonly used in web apps, I think there's some value in making teams aware of this problem and helping them write their code in a way that stops it happening.

## Alternatives

N/A

## Additional information (if applicable)

Suggested changes, to stop the problem occurring are:
- setting the `action` attribute on the `<form>` tag to match the current URL, without `#main-content`
- setting the URL for the next page explicitly using the [Content-Location](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Location) header

## References

[^1]: This is correct behaviour, as defined in [Handling of fragment identifiers in redirected URLs](https://www.w3.org/Protocols/HTTP/Fragment/draft-bos-http-redirect-00.txt)
[^2]: Issue 'ID: DAC_Focus_order_01' on page 19 of [the audit report](https://drive.google.com/file/d/1jRkkL3AZj-7Q0uo7yfHUv_JF9q0Ecw6g/view?usp=sharing) (needs a GDS Google account for access)

calvin-lau-sig7:
Will investigate if this can be fixed in code before deciding on any content. Also if this affects other components.

jordan-carlton:
[...]

PR #4535, opened 2023-12-06
Move `setFocus()` to common utilities

colinrotherham:
Closes https://github.com/alphagov/govuk-frontend/issues/2501 by moving our last bit of duplicated code to a common utility

* https://github.com/alphagov/govuk-frontend/issues/2501

>We only really had two big duplicated things:
>
>1. Different `getFragmentFromUrl()` implementations across **Error summary**, **Skip link** and **Tabs**
>2. Different `setFocus()` implementations across **Error summary**, **Notification banner**, **Skip link**

## Component `setFocus()` method

This PR picked up item 2) above and moved `setFocus()` to a common utility for:

1. Error summary
2. Notification banner
3. Skip link

Their implementations did vary slightly:

1. Skip link target focus is configured on every click, but skipped if `tabindex` is present
1. Notification banner focus is configured once, but skipped if `tabindex` is present
1. Error summary focus is configured once once, but never skipped
1. All three components add a `blur` listener to remove `tabindex`
1. Only Skip link prevents multiple `blur` listeners being added
1. Only Skip link toggles class `govuk-skip-link-focused-element` (see below)
1. Only Skip link calls it ~`setFocus()`~ `focusLinkedElement()`

### Skip link class toggle

Unlike other components that had `setFocus()` duplicated, **Skip link** toggles the `govuk-skip-link-focused-element` class on the target element in addition to managing the `tabindex` attribute

You'll see I've added `onFocus` and `onBlur` callbacks to the `setFocus()` utility to support this:

```mjs
/**
 * Focus the linked element on click
 *
 * Adds a helper CSS class to hide native focus styles,
 * but removes it on blur to restore native focus styles
 */
this.$module.addEventListener('click', () =>
  setFocus($linkedElement, {
    onBeforeFocus() {
      $linkedElement.classList.add('govuk-skip-link-focused-element')
    },
    onBlur() {
      $linkedElement.classList.remove('govuk-skip-link-focused-element')
    }
  })
)
```

### Total bundle size
`all.mjs` ~69.9 KiB~ 69.5 KiB

### DIFF
--- a/packages/govuk-frontend/src/govuk/common/index.mjs
+++ b/packages/govuk-frontend/src/govuk/common/index.mjs
@@ -134,6 +134,52 @@ export function getFragmentFromUrl(url) {
 }
 
 /**
+ * Move focus to element
+ *
+ * Sets tabindex to -1 to make the element programmatically focusable,
+ * but removes it on blur as the element doesn't need to be focused again.
+ *
+ * @private
+ * @template {HTMLElement} FocusElement
+ * @param {FocusElement} $element - HTML element
+ * @param {object} [options] - Handler options
+ * @param {function(this: FocusElement): void} [options.onFocus] - Callback on focus
+ * @param {function(this: FocusElement): void} [options.onBlur] - Callback on blur
+ */
+export function setFocus($element, options = {}) {
+  const isFocusable = $element.getAttribute('tabindex')
+
+  if (!isFocusable) {
+    $element.setAttribute('tabindex', '-1')
+  }
+
+  /**
+   * Handle element focus
+   */
+  function onFocus() {
+    $element.addEventListener('blur', onBlur, { once: true })
+  }
+
+  /**
+   * Handle element blur
+   */
+  function onBlur() {
+    options.onBlur?.call($element)
+
+    if (!isFocusable) {
+      $element.removeAttribute('tabindex')
+    }
+  }
+
+  // Add listener to reset element on blur, after focus
+  $element.addEventListener('focus', onFocus, { once: true })
+
+  // Focus element
+  options.onFocus?.call($element)
+  $element.focus()
+}
+
+/**
  * Checks if GOV.UK Frontend is supported on this page
  *
  * Some browsers will load and run our JavaScript but GOV.UK Frontend
--- a/packages/govuk-frontend/src/govuk/common/index.mjs
+++ b/packages/govuk-frontend/src/govuk/common/index.mjs
@@ -143,7 +143,7 @@ export function getFragmentFromUrl(url) {
  * @template {HTMLElement} FocusElement
  * @param {FocusElement} $element - HTML element
  * @param {object} [options] - Handler options
- * @param {function(this: FocusElement): void} [options.onFocus] - Callback on focus
+ * @param {function(this: FocusElement): void} [options.onBeforeFocus] - Callback before focus
  * @param {function(this: FocusElement): void} [options.onBlur] - Callback on blur
  */
 export function setFocus($element, options = {}) {
@@ -175,7 +175,7 @@ export function setFocus($element, options = {}) {
   $element.addEventListener('focus', onFocus, { once: true })
 
   // Focus element
-  options.onFocus?.call($element)
+  options.onBeforeFocus?.call($element)
   $element.focus()
 }
 
--- a/packages/govuk-frontend/src/govuk/components/skip-link/skip-link.mjs
+++ b/packages/govuk-frontend/src/govuk/components/skip-link/skip-link.mjs
@@ -88,7 +88,7 @@ export class SkipLink extends GOVUKFrontendComponent {
      */
     this.$module.addEventListener('click', () =>
       setFocus($linkedElement, {
-        onFocus() {
+        onBeforeFocus() {
           $linkedElement.classList.add('govuk-skip-link-focused-element')
         },
         onBlur() {
--- a/packages/govuk-frontend/src/govuk/components/skip-link/skip-link.mjs
+++ b/packages/govuk-frontend/src/govuk/components/skip-link/skip-link.mjs
@@ -89,10 +89,10 @@ export class SkipLink extends GOVUKFrontendComponent {
     this.$module.addEventListener('click', () =>
       setFocus($linkedElement, {
         onFocus() {
-          this.classList.add('govuk-skip-link-focused-element')
+          $linkedElement.classList.add('govuk-skip-link-focused-element')
         },
         onBlur() {
-          this.classList.remove('govuk-skip-link-focused-element')
+          $linkedElement.classList.remove('govuk-skip-link-focused-element')
         }
       })
     )
--- a/packages/govuk-frontend/src/govuk/components/skip-link/skip-link.mjs
+++ b/packages/govuk-frontend/src/govuk/components/skip-link/skip-link.mjs
@@ -1,4 +1,4 @@
-import { getFragmentFromUrl } from '../../common/index.mjs'
+import { getFragmentFromUrl, setFocus } from '../../common/index.mjs'
 import { ElementError } from '../../errors/index.mjs'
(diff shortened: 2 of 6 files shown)


romaricpascal (commented):
I like that the helper is extracted 🙌🏻  

Thinking as well taht it would be less complexity (and more de-coupling) to not have `setFocus` receive callbacks, and instead set `focus` and `blur` event listeners on `$linkedElement` in the SkipLink to adjust the classes (they don't bubble so we'd know that's when that specific element is focused, not one of its children). Anyone frome @alphagov/design-system-developers has any thoughts on the matter?

packages/govuk-frontend/src/govuk/common/index.mjs
```diff
@@ -133,6 +133,46 @@ export function getFragmentFromUrl(url) {
...
+      $element.removeAttribute('tabindex')
+    }
+  }
+
+  // Add listener to reset element on blur
+  $element.removeEventListener('blur', resetFocus)
+  $element.addEventListener('blur', resetFocus)
+
+  // Focus element
+  options.onFocus?.call($element)
```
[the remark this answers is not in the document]
colinrotherham: @romaricpascal I previously had it as `onBeforeFocus` if that's alright?

packages/govuk-frontend/src/govuk/components/skip-link/skip-link.mjs
```diff
@@ -86,57 +80,22 @@ export class SkipLink extends GOVUKFrontendComponent {
...
+    /**
+     * Focus the linked element on click
+     *
+     * Adds a helper CSS class to hide native focus styles,
+     * but removes it on blur to restore native focus styles
+     */
+    this.$module.addEventListener('click', () =>
+      setFocus($linkedElement, {
+        onFocus() {
+          this.classList.add('govuk-skip-link-focused-element')
```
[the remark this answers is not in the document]
colinrotherham: Just keeping the KB file size down, but fine by me

colinrotherham:
> I like that the helper is extracted 🙌🏻

Thanks @romaricpascal

> Thinking as well taht it would be less complexity (and more de-coupling) to not have `setFocus` receive callbacks, and instead set `focus` and `blur` event listeners on `$linkedElement` in the SkipLink to adjust the classes (they don't bubble so we'd know that's when that specific element is focused, not one of its children).

Ah this was to support a single "on" and "off" approach:

```mjs
// Add listener to reset element on blur
$element.removeEventListener('blur', resetFocus)
$element.addEventListener('blur', resetFocus)
```

With the approach above, resetting the listeners each time becomes like `clearTimeout()` and meant we could drop `this.linkedElementListener` management inside the **Skip link** component

packages/govuk-frontend/src/govuk/components/skip-link/skip-link.mjs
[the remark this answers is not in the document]
colinrotherham: I've updated it in 6ef903fb78efd41110013604e584664c0a2145f7 but I'm going to keep the `.call($element)` to bind `this` as `$element` because it's what you'd expect from an event handler that isn't an arrow function

packages/govuk-frontend/src/govuk/common/index.mjs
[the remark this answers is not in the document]
colinrotherham: Renamed in f885d67945d21a446717b72d4bad8aa43b9d60da

packages/govuk-frontend/src/govuk/components/skip-link/skip-link.mjs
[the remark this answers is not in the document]
colinrotherham: File size has increased from ~69.9 KiB~ 69.39 KiB → 69.44 KiB which is still a saving

packages/govuk-frontend/src/govuk/common/index.mjs
```diff
@@ -133,6 +133,46 @@ export function getFragmentFromUrl(url) {
...
+  function resetFocus() {
+    options.onBlur?.call($element)
+
+    if (!isFocusable) {
+      $element.removeAttribute('tabindex')
+    }
+  }
+
+  // Add listener to reset element on blur
+  $element.removeEventListener('blur', resetFocus)
```
romaricpascal, line 168: **issue** Reading your last comment on the listeners, I looked a bit more closely at this part and noticed something I missed. 

I think this may not do what we intend it to do actually. The `resetFocus` that gets passed to `removeEventListener` is not the same reference as the one that would have been added by a previous `addEventListener` inside `setFocus`, like in [this Codepen demo](https://codepen.io/romaricpascal/pen/yLZdojK?editors=1011).

romaricpascal: **thought** This makes me wonder if we actually do need all that cleanup:
* Keeping the `tabindex` in doesn't hurt anything, the element will remain programmatically focusable, but not focusable through other means (clicks, keyboard...) so won't impact users
* For the SkipLink, if we constantly remove the class on `blur`, I don't think that's a big deal that a listener remains on the linked element all the time. We can also have it registered during the `focus` listener with `{once: true}` so it's only running on the next blur ([support looks OK for browsers with ES Modules](https://browsersl.ist/#q=supports+es6-module+and+not+supports+once-event-listener)). Could do the same for the `focus` if we want to and only run it on the next `focus` after `setFocus` is called.

romaricpascal (commented):
> Ah this was to support a single "off" and "on" approach:
> 
> ```js
> // Add listener to reset element on blur
> $element.removeEventListener('blur', resetFocus)
> $element.addEventListener('blur', resetFocus)
> ```
> 
> With the approach above, resetting the listeners each time becomes like `clearTimeout()` and meant we could drop `this.linkedElementListener` management inside the **Skip link** component

This prompted me to look at that part more closely and spotted a couple of things we may want to discuss. Happy to jump on a call this afternoon 😊

packages/govuk-frontend/src/govuk/common/index.mjs
colinrotherham: @romaricpascal Cheers for the demo, yeah I understand references and it's totally my fault for last minute changes to keep file size down 🙈

I'm very glad you like `{ once: true }` though, let's do that instead

Otherwise it's quite noticeable without the cleanup when Error summary boxes are click-focusable

colinrotherham: I've revised commit 1a56bfd31ca59022c87bedeb665217f88217ca62 with `{ once: true }` so `removeEventListener()` has gone

In terms of "teardown" it would be nice to fully clean up after ourselves, but there's one other difference between the components we could perhaps discuss from https://github.com/alphagov/govuk-frontend/issues/4465?

1. Skip links preserve the hash fragment on click
2. Tab links preserve the hash fragment on click
3. Error summary links [prevent the hash fragment on click](https://github.com/alphagov/govuk-frontend/blob/ed08beadef553575a24eefcda41f0eed85c53b36/packages/govuk-frontend/src/govuk/components/error-summary/error-summary.mjs#L67-L72)

The issue highlights that hash fragments stick around after redirects:

* https://github.com/alphagov/govuk-frontend/issues/4465

colinrotherham: File size has increased from ~69.9 KiB~ 69.44 KiB → 69.5 KiB which is still a saving

romaricpascal (approved):
Looks good to go now! Thanks for updating 🙌🏻 ⛵

packages/govuk-frontend/src/govuk/common/index.mjs
romaricpascal: > In terms of "teardown" it would be nice to fully clean up after ourselves, but there's one other difference between the components we could perhaps discuss from #4465?

I think we can look at #4465 on its own, so we can focus just on the hash fragments.

PR #4536, opened 2023-12-06
Fix Skip link hash fragment check for Exit this page

colinrotherham:
This PR ensures we ignore Skip link focus handling when linking to other pages

I noticed we throw an error on [Skip link `href: "https://www.gov.uk/"` for **Exit this page**](https://govuk-frontend-review.herokuapp.com/examples/exit-this-page-with-skiplink)

```console
Skip link: Root element (`$module`) attribute (`href`) has no URL fragment
```

For the guidance see [Skip link "**Adding the secondary link**" guidance](https://design-system.service.gov.uk/components/exit-this-page/#adding-the-secondary-link)

### Total bundle size
`all.mjs` ~69.41 KiB~ 69.9 KiB

It's good to note the file size increase due to the new errors, but see #4535 for ideas

## Hash fragment is missing

Since multiple skip links can exist on a page, the error message now includes `getAttribute('href')` from the source HTML attribute to help developer users identify which link is missing a hash fragment

### Before

```console
Skip link: Root element (`$module`) attribute (`href`) has no URL fragment
```

### After

```console
Skip link: Target link (`href="https://example.service.gov.uk"`) has no hash fragment
```

## Invalid URL error message

To skip external URLs we need to compare `location.origin` but this requires a valid URL

I've added a new error message should URL parsing fail:

```console
Skip link: Target link (`href="example.service.gov.uk"`) is invalid
```

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/skip-link/skip-link.mjs
+++ b/packages/govuk-frontend/src/govuk/components/skip-link/skip-link.mjs
@@ -36,18 +36,7 @@ export class SkipLink extends GOVUKFrontendComponent {
     }
 
     this.$module = $module
-    this.$linkedElement = this.getLinkedElement()
 
-    this.$module.addEventListener('click', () => this.focusLinkedElement())
-  }
-
-  /**
-   * Get linked element
-   *
-   * @private
-   * @returns {HTMLElement} $linkedElement - Target of the skip link
-   */
-  getLinkedElement() {
     const linkedElementId = getFragmentFromUrl(this.$module.hash)
 
     // Check for link hash fragment
@@ -68,7 +57,9 @@ export class SkipLink extends GOVUKFrontendComponent {
       })
     }
 
-    return $linkedElement
+    this.$linkedElement = $linkedElement
+
+    this.$module.addEventListener('click', () => this.focusLinkedElement())
   }
 
   /**
--- a/packages/govuk-frontend/src/govuk/components/skip-link/skip-link.mjs
+++ b/packages/govuk-frontend/src/govuk/components/skip-link/skip-link.mjs
@@ -39,6 +39,12 @@ export class SkipLink extends GOVUKFrontendComponent {
 
     const hash = this.$module.hash
     const href = this.$module.getAttribute('href') ?? ''
+    const url = new window.URL(this.$module.href)
+
+    // Check for external link URL and return early
+    if (url.origin !== window.location.origin) {
+      return
+    }
 
     const linkedElementId = getFragmentFromUrl(hash)
 
@@ -73,6 +79,10 @@ export class SkipLink extends GOVUKFrontendComponent {
    * @private
    */
   focusLinkedElement() {
+    if (!this.$linkedElement) {
+      return
+    }
+
     if (!this.$linkedElement.getAttribute('tabindex')) {
       // Set the element tabindex to -1 so it can be focused with JavaScript.
       this.$linkedElement.setAttribute('tabindex', '-1')
@@ -101,6 +111,10 @@ export class SkipLink extends GOVUKFrontendComponent {
    * @private
    */
   removeFocusProperties() {
+    if (!this.$linkedElement) {
+      return
+    }
+
     this.$linkedElement.removeAttribute('tabindex')
     this.$linkedElement.classList.remove('govuk-skip-link-focused-element')
   }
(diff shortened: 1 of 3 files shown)


packages/govuk-frontend/src/govuk/components/skip-link/skip-link.mjs
```diff
@@ -31,12 +31,40 @@ export class SkipLink extends GOVUKFrontendComponent {
...
+    // Check for external link URL and return early
+    if (url.origin !== window.location.origin) {
+      return
+    }
+
+    const linkedElementId = getFragmentFromUrl(hash)
 
     // Check for link hash fragment
-    if (!linkedElementId) {
+    if (!linkedElementId || url.pathname !== window.location.pathname) {
```
36degrees, line 63: Given we switch which error we throw based on `linkedElementId` anyway, I think it might make be more readable to split this into two separate throws?

```mjs
if (url.pathname !== window.location.pathname) {
  throw new ElementError(
    `Skip link: Target link (\`href="${href}"\`) must stay on page (\`${window.location.pathname}\`)`
  )
}

const linkedElementId = getFragmentFromUrl(hash)

if (!linkedElementId) {
  throw new ElementError(
    `Skip link: Target link (\`href="${href}"\`) has no hash fragment`
  )
}
```

36degrees: Actually, I think we might need to take another look at this because of the reasons you mentioned on Slack earlier – from [our guidance](https://design-system.service.gov.uk/components/exit-this-page/#consider-what-to-do-with-user-session-data):

> You’ll need to decide if your service will store the user’s session data after they’ve activated the Exit this page component and left your service.
> 
> For example, you could do this in 2 steps:
>
> 1. replace the default website link to a URL in your service, that when visited, will clear the user’s session data
> 2. redirect the user to another website, such as BBC Weather

In the case of option 1, the URL used could be a different pathname on the same origin – e.g. `/clear-session-data`. In that instance I think we want this to return silently without throwing, the same way we do if the origin doesn't match.

packages/govuk-frontend/src/govuk/components/skip-link/skip-link.puppeteer.test.js
```diff
@@ -67,6 +67,15 @@ describe('Skip Link', () => {
   })
 
   describe('errors at instantiation', () => {
+    it('can return early without errors for external href', async () => {
```
36degrees, line 70: Thanks for adding these 🙌🏻

packages/govuk-frontend/src/govuk/components/skip-link/skip-link.mjs
```diff
@@ -31,12 +31,40 @@ export class SkipLink extends GOVUKFrontendComponent {
...
+
+    /**
+     * Check for valid link URL
+     *
+     * {@link https://caniuse.com/url}
+     * {@link https://url.spec.whatwg.org}
+     *
+     */
+    try {
+      url = new window.URL(this.$module.href)
```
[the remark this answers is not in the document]
colinrotherham: Ah I've used `this.$module.href` here for the full URL:

```console
http://localhost:3000/components/skip-link/preview#content
```

But I'm using `href` to show what's in the actual HTML for helpful error messages:

```mjs
const href = this.$module.getAttribute('href') ?? ''
```

Which most of time will default to:

```console
#content
```

packages/govuk-frontend/src/govuk/components/skip-link/skip-link.mjs
colinrotherham: What if we check for the presence of a hash fragment and keep the error?

But for paths like `/clear-session-data` without a hash fragment we return early

Just tried it with https://github.com/alphagov/govuk-frontend/pull/4536/commits/f93a017aba26ed1ed5823399e7c74989e77df7f1

colinrotherham:
I've unstacked PR https://github.com/alphagov/govuk-frontend/pull/4535 from this one so it's not a blocker

packages/govuk-frontend/src/govuk/components/skip-link/skip-link.mjs
colinrotherham: Just caught up with @romaricpascal

>…we’re thinking it’d be less risky to go with only throwing if it’s a same page link and there’s no hash, otherwise early returning and not worrying, rather than trying to be too tight around expecting a link to the same page specifically.

So I've pushed 5cf5c3289a4ab85b2360bcff763fa9404cc58062 with a single early return

```mjs
// Return early for external URLs or links to other pages
if (
  url.origin !== window.location.origin ||
  url.pathname !== window.location.pathname
) {
  return
}
```

Which means this error message is no longer included too:

```console
Target link (`href="/somewhere-else#main-content"`) must stay on page (`/components/skip-link/preview`)
```

36degrees: Sounds good to me 👍🏻

romaricpascal (approved):

Issue #4512, opened 2023-11-27
Decide how we want the show/hide password component to be implemented


querkmachine:
There are a few different ways that we could implement the show/hide password toggle component (herein, just 'password component' for brevity) within the current structure of Frontend. Each method has pros, cons, and repercussions on how we manage other components and the documentation going forward.

## Comparison of methods

On a high level, the choice is between an entirely new standalone component, new "hybrid" component that inherits existing components, or creating a variant of the existing text input component.

### Standalone component

As a **standalone component**, it would reimplement some existing features where we need them to be significantly different—such as the password input itself—as well as providing unique wrapping elements and styles. 

This gives the component the flexibility to diverge as much as it needs to, at the expense of duplicating some code that already exists. 

However, this can also be beneficial, as it allows the password component to be more focused and intentional, rather than sharing parameters and features that the text input has that wouldn't make sense on a password toggle. 

The duplication can be partially mitigated on the CSS front by importing the text input's styles, creating a stylistic dependency between them, which is something we already do elsewhere. 

In terms of documentation, standalone components usually receive similarly standalone guidance in the Design System, linking to related guidance where necessary.

#### Existing standalone components in Frontend

- Most components are standalone, which makes sense as most of them are distinct in function, code, and appearance.
- The [checkboxes](https://design-system.service.gov.uk/components/checkboxes/) and [radios](https://design-system.service.gov.uk/components/radios/) are functionally and stylistically alike enough that they _could_ be the same component, but we've decided to keep them separate as it's ultimately easier to manage the duplication than to maintain and document them as a single thing. 

### "Hybrid" component

As a **hybrid component**, the component would directly use the text input (and, potentially, button) macros within itself, immediately inheriting the capabilities of each of them and combining them with new code and styles.

This reduces duplication within the Design System code at the expense of more tightly coupling these components to one another. 

Whilst this theoretically reduces maintenance, by reducing duplication, it also adds a maintenance task as we would need to ensure that future updates to the text input and button components do not have an adverse effect on the password component.

There is a new form of duplication introduced in how developers interface with the component too, as we would need to 'forward' parameters from the password component down to the child components. In some cases we won't want this to happen, or we would want the default value of a parameter to be different to what the child component otherwise uses, which can create more complex logic. 

[...]

PR #4539, opened 2023-12-08
[SPIKE] Password input component - hybrid approach

querkmachine:
Alternative approach to a password input component using the 'hybrid' approach documented in #4512. Like the 'standalone' spike (#4442), there are no tests or documentation included. 

As noted in #4512, this demonstrates a few things:
- There is less code duplication compared to entirely standalone.
- There is slightly less flexibility in how HTML can be used. In this instance, the `govuk-password-input` div has had to move to be outside of `govuk-form-group` and new JavaScript added to create the necessary 'wrapper' around the input/button. 
- A lot of parameters need to be manually forwarded down to the child macro, similar to the character count.
  - This is actually neater than the character count's case, as that has to override some nested parameters. 
  - It could be neatened up still if we decide that some features of text input are just not useful in this context. i.e. Do we realistically want the `pattern`, `inputmode`, `prefix` or `suffix` options to be available for password inputs?
  - I think the DOM changes needed for `prefix` and `suffix` are incompatible with the DOM changes needed for password input anyway.
- The input component has had to be updated to add new parameters that the password input needs, as they cannot be applied directly in the password input component. 
  - Or rather, they could, if we hijacked the `attributes` parameter, but then end users couldn't use it and we probably don't want that.

### DIFF
--- a/packages/govuk-frontend/src/govuk-prototype-kit/govuk-prototype-kit.config.unit.test.mjs
+++ b/packages/govuk-frontend/src/govuk-prototype-kit/govuk-prototype-kit.config.unit.test.mjs
@@ -137,6 +137,10 @@ describe('GOV.UK Prototype Kit config', () => {
           macroName: 'govukPanel'
         },
         {
+          importFrom: 'govuk/components/password-input/macro.njk',
+          macroName: 'govukPasswordInput'
+        },
+        {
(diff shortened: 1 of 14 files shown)


querkmachine:
In https://github.com/alphagov/govuk-frontend/issues/4512 we determined we will probably follow the standalone route for now. Closing this down as it's served its purpose.

Issue #4287, opened 2023-10-02
Investigate strengthening checks that transpiled code matches browser support


romaricpascal:
## What

See if we can improve our confidence that the code we ship, which is transpiled by [Babel](https://babeljs.io/), matches our browser support (ie. doesn't require any surprise polyfill or leftover syntax that wouldn't run everywhere).

There are various times this could happen:
- during review of PRs enabling new syntaxes ([review of browserlist](https://github.com/alphagov/govuk-frontend/blob/be9639391cc7974622a2995c351b1ceee6053553/packages/govuk-frontend/.eslintrc.js#L43), [review of code compiled by Babel in its online REPL](https://babeljs.io/repl#?browsers=Chrome%2066&build=&builtIns=false&corejs=false&spec=false&loose=true&debug=false&forceAllTransforms=false&modules=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=true&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Cstage-2%2Cstage-3&prettier=false&targets=&version=7.23.1&externalPlugins=&assumptions=%7B%7D), manual test with tail end of browser support)
- through some automation (diff display on pull requests, linting of output code with [ESLint's es-x plugin](https://eslint-community.github.io/eslint-plugin-es-x/), [test with tail end of browser support](https://github.com/alphagov/govuk-frontend/issues/3278), [having Babel log the polyfills it'd be keen to inject](https://github.com/babel/babel-polyfills/blob/main/docs/usage.md#shouldinjectpolyfill)...)

## Why

This would give us more confidence that our configuration of , to limit the language features we use, and Babel, to transpile new features to older syntax, doesn't have gaps.

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [ ] Plan investigation (some options in our [dev catch-up metting notes (internal)](https://docs.google.com/document/d/1_ckzZw_V8RQgYlVV65bCun87Uw7v7MCZVlMaKgmjDxM/edit#heading=h.wzkmk3eqz8ms))

Issue #3948, opened 2023-07-11
Set up Babel polyfills for browser compatibility


colinrotherham:
## What

Set up Babel to automatically add polyfills for unsupported browser APIs.

Ideally, unsupported code should be replaced with "pure" polyfills (ponyfills) that do not modify browser globals.

## Why

This issue follows on from https://github.com/alphagov/govuk-frontend/issues/3504 and is only needed when we choose to use a browser API unavailable in [`.browserlistrc`](https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/.browserslistrc) supported browsers

Polyfill code should not cause parsing or syntax errors. Since feature detection can guard against code we choose not to run, unsupported browser APIs don't have to be polyfilled.

This will be done with caution while [keeping an eye on the impact of the polyfills on the size of the built files](https://github.com/alphagov/govuk-frontend/issues/3279).

## Who needs to work on this

Developers

## Who needs to review this

Developers

### Depends on
- [ ] #3504 

```[tasklist]
### Done when
- [ ] We've reviewed the relevance of #3082
- [ ] https://github.com/alphagov/govuk-frontend/issues/3283
- [ ] https://github.com/alphagov/govuk-frontend/issues/3722
```

PR #4557, opened 2023-12-11
[SPIKE] Configure Babel to add core-js polyfills

colinrotherham:
This PR adds the Babel polyfill provider core-js via:
https://www.npmjs.com/package/babel-plugin-polyfill-corejs3

This is part of a two spike series:

* https://github.com/alphagov/govuk-frontend/pull/4301
* https://github.com/alphagov/govuk-frontend/pull/4557

## Findings

Lots more polyfills are added by core-js when compared to ES shims in https://github.com/alphagov/govuk-frontend/pull/4301

We also have lots more unwanted bug fixes:

* core-js adds `Array.prototype.from` polyfill because the [type being iterated is unknown](https://github.com/zloirock/core-js/issues/1003)
* core-js adds `Array.prototype.includes` polyfill [to avoid bugs](https://bugzilla.mozilla.org/show_bug.cgi?id=1767541)
* core-js adds `Error.prototype.cause` polyfill [to avoid bugs](https://bugs.chromium.org/p/v8/issues/detail?id=12006)
* core-js adds `URL` constructor polyfill [to avoid bugs](https://bugs.webkit.org/show_bug.cgi?id=216841)

Including an overwhelming amount of polyfills to [patch various browser APIs](https://github.com/zloirock/core-js/blob/v3/packages/core-js-compat/src/data.mjs) in already-supported features:

```console
'es.array.includes',
'es.array.iterator',
'es.regexp.exec',
'es.regexp.to-string',
'es.string.includes',
'es.string.match',
'es.string.replace',
'es.string.trim',
```

For example, Safari 12.1 updated `''.trim()` to include revised empty space characters:
https://github.com/paulmillr/es6-shim/pull/354#issuecomment-115565060

Similarly, versions below Firefox 102 handled `[].includes()` with sparse arrays incorrectly:
https://bugzilla.mozilla.org/show_bug.cgi?id=1767541

### Global polyfills

This PR uses `core-js` rather than `core-js-pure` since lots of polyfills are missed due to:

* https://github.com/babel/babel-polyfills/issues/166

### Remaining issues

* Browserslist matches Opera Mobile 73 under `es6-module` but [Babel disagrees](https://github.com/alphagov/govuk-frontend/pull/4301/commits/41468a62818d6c5d57e219a149690b9bdb62a5d8)
* Rollup plugin [`@rollup/plugin-commonjs`](https://www.npmjs.com/package/@rollup/plugin-commonjs) adds ES module wrappers to polyfills
* Rollup plugin [`@rollup/plugin-babel`](https://www.npmjs.com/package/@rollup/plugin-babel) adds helpers to avoid ES module wrappers
* Rollup outputs `dist/govuk/node_modules` not `dist/govuk/vendor`

---

### Automatic polyfills

When Babel was added we chose "transforms only" without polyfills (see [decision record](https://github.com/alphagov/govuk-design-system-architecture/blob/main/decision-records/006-javascript-compatibility.md))

This PR ensures early investigations aren't lost for:

* https://github.com/alphagov/govuk-frontend/issues/3082
* https://github.com/alphagov/govuk-frontend/issues/3287
* https://github.com/alphagov/govuk-frontend/issues/3948
* https://github.com/alphagov/govuk-frontend/issues/4287

### Decision record implications

Since Babel with "transforms only" silently skips polyfills, we noted that [care must be taken](https://github.com/alphagov/govuk-design-system-architecture/blob/main/decision-records/006-javascript-compatibility.md#implications):

>When writing or changing JavaScript we currently have to remember not to use features that aren't available in the oldest browsers we support, or to manually add a polyfill. We also rely heavily on manual testing across multiple browsers.

This spike PR ensures:

1. Babel logger lists required polyfills
2. Babel injects imports for required polyfills
3. Rollup builds fail when imports are missing

i.e. Support for both polyfills and protection from missed polyfills all-in-one


querkmachine:

The related issue remains open for comments. https://github.com/alphagov/govuk-frontend/issues/3948

Issue #4439, opened 2023-11-07
Replace dependency: cpy


domoscargin:
## What
Replace `cpy` with a safer solution. This may just be a case of [reverting to using Gulp](https://github.com/alphagov/govuk-frontend/pull/4430#issuecomment-1795516873), or we could [roll our own function](https://github.com/alphagov/govuk-frontend/pull/4430#pullrequestreview-1715773255).

## Why
From the developer of `cpy`:

> IMPORTANT: This package has a lot of problems and I unfortunately don't have time to fix them. I would recommend against using this package until these problems are resolved. Help welcome (see the issue tracker) 🙏

Some of the open issues on the package mention things like data loss. We only use simple combinations of '**/*' and '**/*.js', so we're not likely to be affected by glob-related malarkey, but still worth making sure we're using something safer.

## Who needs to work on this
Developer

## Who needs to review this
Developer

## Done when

- [ ] We're no longer using `cpy`

PR #4560, opened 2023-12-12
Use Gulp to copy files instead of `cpy`

colinrotherham:
Adding a dash of Gulp back again 😶

Closes https://github.com/alphagov/govuk-frontend/issues/4439 by replacing our `files.copy()` task with `gulp.src()` + `gulp.dest()`

## Removing the `cpy` package

This lets us remove the [`cpy` package](https://www.npmjs.com/package/cpy) added in https://github.com/alphagov/govuk-frontend/pull/3382/commits/8edda314842332f4c11200b379f4de516ce9d45d due to the maintainer stating:

>**IMPORTANT:** This package has a lot of problems and I unfortunately don't have time to fix them. I would recommend against using this package until these problems are resolved. Help welcome (see the issue tracker) 🙏

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8779,22 +8779,6 @@
         "node": ">=0.10.0"
       }
     },
-    "node_modules/copy-file": {
-      "version": "11.0.0",
-      "resolved": "https://registry.npmjs.org/copy-file/-/copy-file-11.0.0.tgz",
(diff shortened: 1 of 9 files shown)


romaricpascal (approved):
Cheers for sorting this out 🙌🏻

PR #4561, opened 2023-12-12
[SPIKE] Prevent components being initialised more than once

colinrotherham:
Quick spike to discuss adding `data-module-init` to components when they're initialised

This can be used to prevent components being initialised more than once

Closes: https://github.com/alphagov/govuk-frontend/issues/1127

### Checking if previously initialised

I've added a new `checkInitialised($module)` method on the base class

When the data attribute `data-module-init` is already set we can throw an error

```patch
  constructor($module) {
    this.checkSupport()
+   this.checkInitialised($module)
+
+   // Mark component as initialised in HTML
+   $module?.setAttribute('data-module-init', 'true')
  }
```

### DIFF
--- a/packages/govuk-frontend/src/govuk/common/index.mjs
+++ b/packages/govuk-frontend/src/govuk/common/index.mjs
@@ -201,6 +201,17 @@ export function setFocus($element, options = {}) {
 }
 
 /**
+ * Checks if component is already initialised
+ *
+ * @internal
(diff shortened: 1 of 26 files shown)


querkmachine:

Fixing the multiple initialisation issue is still on our radar and is something we would like to resolve in future. 
- https://github.com/alphagov/govuk-frontend/issues/1127

Issue #3764, opened 2023-06-07
Remove hardcoded media query breakpoints from the JavaScript for the header and tabs components


36degrees:
## What

We have hardcoded the default desktop and tablet breakpoints into the JavaScript for both the header and tabs components:

https://github.com/alphagov/govuk-frontend/blob/a4ae21949293cdb86eda32c1383fd5ec78835e70/packages/govuk-frontend/src/govuk/components/header/header.mjs#L57-L58

https://github.com/alphagov/govuk-frontend/blob/a4ae21949293cdb86eda32c1383fd5ec78835e70/packages/govuk-frontend/src/govuk/components/tabs/tabs.mjs#L63-L64

Find another way to link the JavaScript behaviour to the styles applied at different breakpoints without needing to hardcode them.

## Why

Having the breakpoints hardcoded is likely cause issues if users override the default breakpoints.

There's also a risk that they may get out of sync if we ever make changes to the default breakpoints.

## Further details

Once we've switched to loading our JavaScript using `script type="module"`[^1] we might be able to expose the breakpoint as a CSS variable in the CSS for each component (or on the document root) and read it out using `window.getComputedStyle().getPropertyValue`.

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [x] Hardcoded desktop breakpoint removed from the JavaScript for the header component
- [x] Hardcoded tablet breakpoint removed from the JavaScript for the header component

[^1]: because [all browsers that support `script type="module"` also support CSS variables](https://browsersl.ist/#q=supports+es6-module+and+not+supports+css-variables)

PR #4562, opened 2023-12-13
Use CSS custom properties for component `matchMedia()`

colinrotherham:
This PR adds CSS custom properties for all our breakpoints alongside `--govuk-frontend-version`

```css
--govuk-frontend-breakpoint-mobile: 20rem;
--govuk-frontend-breakpoint-tablet: 40.0625rem;
--govuk-frontend-breakpoint-desktop: 48.0625rem;
--govuk-frontend-version: "development";
```

Then uses them in `window.matchMedia()` instead of hard coded values in JavaScript

Closes https://github.com/alphagov/govuk-frontend/issues/3764

### Missing breakpoints

Throws an error if the CSS custom properties are missing:

```console
ElementError: Tabs: CSS custom property (`--govuk-frontend-breakpoint-tablet`) on pseudo-class `:root` not found
```


packages/govuk-frontend/src/govuk/components/header/header.mjs
```diff
@@ -83,23 +83,45 @@ export class Header extends GOVUKFrontendComponent {
...
+  /**
+   * Setup viewport resize check
+   *
+   * @private
+   */
+  setupResponsiveChecks() {
+    const breakpointProperty = '--govuk-frontend-breakpoint-desktop'
+    const breakpointValue = getComputedStyle(document.body).getPropertyValue(
+      breakpointProperty
+    )
```
romaricpascal, line 102: **question** How would you feel about encapsulating that into a `getBreakpointValue(breakpointName)`. Both to wrap what's needed to access the value out of each class and to see if services maybe start picking up the pattern by digging into our private code?

packages/govuk-frontend/src/govuk/components/header/header.mjs
```diff
@@ -83,23 +83,45 @@ export class Header extends GOVUKFrontendComponent {
...
+  }
+
+  /**
+   * Setup viewport resize check
+   *
+   * @private
+   */
+  setupResponsiveChecks() {
+    const breakpointProperty = '--govuk-frontend-breakpoint-desktop'
+    const breakpointValue = getComputedStyle(document.body).getPropertyValue(
```
romaricpascal, line 100: **nitpick**: `:root` corresponds to `document.documentElement`. It would make the mapping between CSS and JS easier if they both reference the same element, even though the custom property value will cascade down to `document.body`.

packages/govuk-frontend/src/govuk/components/header/header.mjs
```diff
@@ -83,23 +83,45 @@ export class Header extends GOVUKFrontendComponent {
...
+      breakpointProperty
+    )
+
+    if (!breakpointValue) {
+      throw new ElementError({
+        componentName: 'Header',
+        identifier: `CSS custom property (\`${breakpointProperty}\`)`
```
romaricpascal, line 107: **issue** Thinking the message may lead to people looking to add a custom property on the root of the component's markup, while we look it up on the body 😊 

```suggestion
        identifier: `CSS custom property (\`${breakpointProperty}\`) on document's \`body\``
```

romaricpascal (commented):
I like the pattern for passing values from CSS to JavaScript that's [enabled by our new browser support](https://browsersl.ist/#q=supports+es6-module+and+not+supports+css-variables)! 

Listed a couple of suggestions (only the message wording is really blocking). Also thinking it's a pattern worth sharing with the team at dev catch-up as it may be useful in other occasions 😊

packages/govuk-frontend/src/govuk/components/header/header.mjs
colinrotherham: Absolutely fine by me

packages/govuk-frontend/src/govuk/components/header/header.mjs
colinrotherham: Ah the issue said `window.getComputedStyle()` and I grabbed [`document.body` from Slack](https://gds.slack.com/archives/C0447D0STV0/p1688038334540579?thread_ts=1688038122.738779&cid=C0447D0STV0)

Shall I use `document.documentElement` for `<html>` instead?

You've suggested `<body>` below in https://github.com/alphagov/govuk-frontend/pull/4562#discussion_r1427030841

packages/govuk-frontend/src/govuk/components/header/header.mjs
colinrotherham: How's this? Just pushed it up in 21a519d4fe484ce1b9acb8dc6563b8c165c9e46e

```console
ElementError: Tabs: CSS custom property (`--govuk-frontend-breakpoint-tablet`) on pseudo-class :root not found
```

packages/govuk-frontend/src/govuk/components/header/header.mjs
colinrotherham: I've gone with `document.documentElement` to represent `:root`

packages/govuk-frontend/src/govuk/components/header/header.mjs
colinrotherham: I've added `getBreakpoint()` in bf557114a4a40c2afd5c150f986297bcd5050ae2 as we'll still need the CSS property name for the error

```mjs
const breakpoint = getBreakpoint('tablet')

// breakpoint.property
// breakpoint.value
```

packages/govuk-frontend/src/govuk/components/header/header.mjs
romaricpascal: Oh damn sorry for the confusion in the following comment, I did indeed mean to align CSS and JS on the `:root`/`document.documentElement`.

packages/govuk-frontend/src/govuk/components/header/header.mjs
```diff
@@ -83,23 +84,42 @@ export class Header extends GOVUKFrontendComponent {
...
+   *
+   * @private
+   */
+  setupResponsiveChecks() {
+    const breakpoint = getBreakpoint('desktop')
+
+    if (!breakpoint.value) {
+      throw new ElementError({
+        componentName: 'Header',
+        identifier: `CSS custom property (\`${breakpoint.property}\`) on pseudo-class :root`
```
romaricpascal, line 105: **nitpick** I think there's a couple of backticks missing.

```suggestion
        identifier: `CSS custom property (\`${breakpoint.property}\`) on pseudo-class \`:root\``
```

romaricpascal (approved):
Neat! Thanks for introducing the helper and picking `document.documentElement`. Approving, but I think there's a couple of backticks that are missing in the error messages around `:root` (won't need another review).

packages/govuk-frontend/src/govuk/components/header/header.mjs
colinrotherham: Yeah, left them off but will add 👍

Is it just types that we don't wrap in backticks?

```console
Radios: Root element (`$module`) is not of type HTMLElement
```

PR #4564, opened 2023-12-13
[SPIKE] Undo Tabs JS styling when initialisation fails

colinrotherham:
Another thing worth a quick spike

We've added extra reasons for the **Tabs** component to fail initialisation now `ElementError` is thrown

Since all of these are in control, could we add a class like `.govuk-tabs--error` to "undo" JS styling?

```patch
  // GOV.UK Frontend JavaScript enabled
- .govuk-frontend-supported {
+ .govuk-frontend-supported .govuk-tabs:not(.govuk-tabs--error) {
```

## Failing sensibly

Accidents happen. Errors can handled, and we can fall back to the non-JS experience:

Tabs component without JS styling

Might close https://github.com/alphagov/govuk-frontend/issues/999

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/tabs/_index.scss
+++ b/packages/govuk-frontend/src/govuk/components/tabs/_index.scss
@@ -45,7 +45,7 @@
   }
 
   // GOV.UK Frontend JavaScript enabled
-  .govuk-frontend-supported {
+  .govuk-frontend-supported .govuk-tabs:not(.govuk-tabs--error) {
     @include govuk-media-query($from: tablet) {
       .govuk-tabs__list {
         @include govuk-clearfix;
--- a/packages/govuk-frontend/src/govuk/components/tabs/tabs.mjs
+++ b/packages/govuk-frontend/src/govuk/components/tabs/tabs.mjs
@@ -55,15 +55,18 @@ export class Tabs extends GOVUKFrontendComponent {
       })
     }
 
+    this.$module = $module
+
     const $tabs = $module.querySelectorAll('a.govuk-tabs__tab')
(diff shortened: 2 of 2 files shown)


36degrees:
Thanks for looking into this!

I think the more likely failure state is where something outside of our control prevents the JavaScript for a component from running entirely. This could be a JavaScript error elsewhere in a bundled script, or a network error that prevents the file from loading.

I think we might want to flip the logic a little so we can 'fail open' even if the component JS never runs at all. Prompted by a suggestion from yourself (a year ago!) about a 'smoke and mirrors' approach I've had something like this in my head for a while: https://github.com/alphagov/govuk-frontend/compare/main...spike-fail-better

Thoughts?

colinrotherham:
No problem 😊

I think both approaches are covered by [my initial thoughts](https://github.com/alphagov/govuk-frontend/issues/999#issuecomment-1348961804) and you paraphrased the "smoke and mirrors" idea into https://github.com/alphagov/govuk-frontend/issues/999#issuecomment-1387289185 if it's easier to link to those comments?

E.g. We could update your spike to move `govuk-frontend-ready` to a per-component `--loading` modifier because components could be set up at any time, e.g. when Tabs kicks in at wider viewports

But equally, there's a benefit of styling a component _before_ it's ready to avoid flickers of UI enhancement. This could be waiting for an error (my spike) or when the time from `--loading` → `--ready` exceeds a maximum timeout

Best of both worlds?

querkmachine:

Fixing the issue with components 'failing closed' is still on our radar and is something we would like to resolve in future. 
- https://github.com/alphagov/govuk-frontend/issues/999

Issue #2893, opened 2022-10-03
Make CharacterCount use of `attributes` option consistent with other components


romaricpascal:
## What

Update CharacterCount component so its `attributes` option sets attributes on the root tag of the component, instead of the textarea. To still allow setting attributes on the `textarea`, introduce a `textareaAttributes`.

This'll require a two step approach where:
1. [Non breaking] We introduce the new `textareaAttributes` option and:
   - if present, assign `attributes` to the root tag
   - if absent, assign `attributes` to the textarea like we do now and warn users that the meaning of the `attributes` option will soon change and encourage using `textareaAttributes`
2. [Breaking] Remove the current behaviour and always assign `attributes` to the root tag

## Why

This'll make the component consistent with the rest of the other components which set attributes on their root tag, for ex. allowing them to set a `lang` attribute without having to wrap the component.

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [ ] New `textareaAttributes` option is introduced and used for picking on which tag `attributes` apply
- [ ] Deprecation message is shown if only using the `attributes`
- [ ] New behaviour is documented in the component options
- [ ] Issue is created to remove the current behaviour and deprecation message

36degrees:
Related: #1649

querkmachine:
Given this is an issue with multiple Nunjucks parameters, would it be sensible to have a nested set of params rather than adding new top-level ones? 

i.e. Instead of adding `textareaAttributes`, `textareaClasses`, `textareaId`, and so on, we do this? 
```nunjucks
{{ govukCharacterCount({

  // These apply to the textarea
  attributes: {},
  classes: '',
  id: '',
  ...

  // These apply to the container
  formGroup: {
    attributes: {},
    classes: '',
    id: '',
    ...
  }

}) }}
```

That would still be internally consistent (we use `formGroup.classes` to apply classes on many form input types) and—at first glance—would not require a breaking change.

colinrotherham:
This is now ready for review as part of https://github.com/alphagov/govuk-frontend/pull/4566

I've gone with @querkmachine approach from https://github.com/alphagov/govuk-frontend/issues/2893#issuecomment-1269613931

PR #4565, opened 2023-12-14
Allow `attributes` option on form groups

colinrotherham:
This PR adds support for `formGroup: { attributes: {} }`

It's part of a fix suggested by @querkmachine in https://github.com/alphagov/govuk-frontend/issues/2893#issuecomment-1269613931 and follows the thinking in:

* https://github.com/alphagov/govuk-frontend/pull/1043#issuecomment-434289395

>Are there other things that we can foresee users wanting to add form-groups in the future, like attributes? Just wondering if we should opt for `formGroup: { classes: '' }` instead.

This is a supporting change to enable https://github.com/alphagov/govuk-frontend/pull/4566

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,22 @@ For advice on how to use these release notes see [our guidance on staying up to 
 
 ### New features
 
+#### Add attributes to component form group wrappers
+
+You can now add attributes to the form group wrapper for all components with form fields.
+
+```njk
+govukRadios({
+  formGroup: {
+    attributes: {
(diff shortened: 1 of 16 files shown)


romaricpascal (commented):
Looks neat to me. 🙌🏻 Thinking it's worth adding a CHANGELOG entry before we merge, do you mind adding one, please? 😊

owenatgov (approved):
💪🏻

colinrotherham:
Thanks @owenatgov

Just going to rebase to pick up the new Percy diff changes

Issue #1649, opened 2019-11-18
Can't add custom class to character count


LBHELewis:
Currently it's not possible to add a custom class to the character count wrapper (`.govuk-character-count`) because the `classes` property is being used for the nested textarea instead. This goes against the convention in other components. 

It would be helpful for us to have `classes` apply to the `.govuk-character-count` (or if this would be a problem because of introducing a breaking change then it would be helpful to have a new property that gives the ability to add custom classes to `.govuk-character-count`). In our use case we are re-skinning `govuk-frontend` for Hackney Council (see https://github.com/lbhackney-IT/lbh-frontend) and would like to alter the spacing around the character count component and to style the hint inside it without targeting solely `.govuk-*` classes (a convention that we're trying to stick to). There are ways of getting round this (not least breaking our convention in this case) so it isn't absolutely crucial to us, but would be nice if it could work the same way as all other components.

nickcolley:
Hey Emma,

We recognise that the api for this component is inconsistent with the rest of our components.

To make it consistent we'd need to make a breaking change.

For this reason we'll see if more of our users run into this issue, and then consider changing this.

Let us know if this causes blockers for you, and we can reconsider this decision.

vanitabarrett:
As discussed in the v4 prioritisation, this component does work differently to our other components, so maybe it's ok that this is reflected in the macro options. This problem is wider than just classes, e.g: the macro options also let you pass in `id` and other HTML attributes that apply to the textarea, not the parent. There are probably 3 options here:

1. We rename `classes`, `id` etc to `textareaClasses`, `textareaId` etc and introduce new macro options to apply to the parent that follow the macro options for our other components
2. Introduce new macro options that apply to the parent, e.g: `parentClasses`, `parentId`. This would not be a breaking change
3. Revisit the character count component and consider rolling it into the textarea component

Probably needs more investigation. It would also be good to know if this is something that causes an issue for a lot of our users, or if it behaves as expected despite being different to other design system components.

colinrotherham:
This is now ready for review as part of https://github.com/alphagov/govuk-frontend/pull/4566

Issue #1602, opened 2019-10-07
Character count message should be placed correctly in markup without Javascript


hannalaakso:
The character count Javascript [moves the count message](https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/components/character-count/character-count.js#L31) ("You have x characters remaining") in the DOM to be inside `govuk-form-group`. This was probably useful when the message markup was created dynamically. However this functionality is now redundant. The JS should be removed and the count message markup placed inside `govuk-form-group` in the component markup. This is so that we don't unnecessarily rely on Javascript for the correct order of DOM elements.

It’s important that the count message markup is inside `govuk-form-group` so that it gets wrapped within the error block (see screen grab below).

![65713384-fc0af600-e090-11e9-90d3-33230813d538](https://user-images.githubusercontent.com/5007934/66323584-ebbe0b00-e91b-11e9-9bbd-c367ed85530b.png)

We put off making this change this in https://github.com/alphagov/govuk-frontend/pull/1594 as it's a breaking change. 

## Done when

- [ ] TBC
- [ ] "Details of breaking change" section below completed to help with writing release notes

## Details of breaking change

- which users are affected: TBC
- the change that’s been made: TBC
- changes users will have to make: TBC
- impact of users not making those changes: TBC

36degrees:
I think this would require us to add something to the textarea API that allows us to insert this inside the form-group.

hannalaakso:
Yes, we’d need to add it to the textarea API and tweak the textarea macro template so that it can accept a hint which would render below `<textarea>` but inside `.govuk-form-group`. But that is actually quite a strange requirement since the character count component is the only thing that needs to do this and most of the time inserting a hint below the textarea for any other purposes is not something we'd recommend teams do.

In conclusion, I don't think we should do this - it's a bit of an irregularity (ideally the JS wouldn’t need to move the markup around) but it's not causing any problems and if anything, adding it into the textarea API could actually lead to it being misused in other contexts. 

Tempted to just close this issue but leave it open for a bit longer for any other comments.

lfdebrux:
If we made this change, would that mean that the hint would appear below the text area for users who have turned off JavaScript? That would seem like surprising/wrong behaviour to me.

36degrees:
> If we made this change, would that mean that the hint would appear below the text area for users who have turned off JavaScript? That would seem like surprising/wrong behaviour to me.

[...]

Issue #2917, opened 2022-10-13
Introduce Nunjucks macros abstracting data attribute rendering


romaricpascal:
## What

Abstract rendering of data-attributes for passing configuration into a `govukDataAttributes` with a call of the following shape:

```njk
{{
	govukDataAttributes({
	   'attribute-name': value,
	   'other-attribute-name': {
		  value: otherValue, 
	      testUndefined: true,
	      escape: false # Only if there are instances we don't escape the attributes
	   }
	})
}}
```

Expand the concept with a helper for internationalisation attributes that relies on `plurals` or `string` key instead of `value` and delegates to `govukDataAttributes`, building the `i18n.<attribute-name>[.plural-category]` attribute name. Based on [the ideas in this comment](https://github.com/alphagov/govuk-frontend/pull/2895#discussion_r994400894).

## Why

Setting the attributes is very repetitive in the templates and always the same shape:
1. if {{value}} or {{value !== undefined}} to condition the rendering of the attribute itself
2. output " data-{{name}}="{{value | escape}}"

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [ ] Abstraction macros have been introduced and tested
- [ ] Component templates have been updated to use the abstractions

querkmachine:
I notice that an [i18n related macro](https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/macros/i18n.njk) was added a couple of months ago, that seems related to this idea. 

Is this issue still relevant or in need of updates in response to that?

colinrotherham:
Hadn't realised this issue existed. We nearly implemented this in:

* https://github.com/alphagov/govuk-frontend/pull/3867

Fairly quick to pull out another macro (for data- or regular attributes in general) from `govukI18nAttributes()`

PR #4566, opened 2023-12-14
Use Character count `formGroup` as module wrapper

colinrotherham:
This is a non-breaking change to remove an unnecessary **Character count** wrapper

It uses the same `afterInput` approach adopted by **Password input** in https://github.com/alphagov/govuk-frontend/pull/4442

Closes https://github.com/alphagov/govuk-frontend/issues/1602, closes https://github.com/alphagov/govuk-frontend/issues/1649 and closes https://github.com/alphagov/govuk-frontend/issues/2893

## Before

The **Character count** component includes both a `data-module` wrapper and **Textarea** form group

```html
<div class="govuk-character-count" data-module="govuk-character-count" data-maxlength="100">
  <div class="govuk-form-group govuk-form-group--error">
    <!-- // Label, hint, error and textarea -->
  </div>
  <!-- // Count message -->
</div>
```

## After

The **Character count** component adds `data-module` directly onto the **Textarea** form group

```html
<div class="govuk-form-group govuk-form-group--error govuk-character-count" data-module="govuk-character-count" data-maxlength="100">
  <!-- // Label, hint, error, textarea and count message -->
</div>
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,43 @@ For advice on how to use these release notes see [our guidance on staying up to 
 
 ## Unreleased
 
+#### Update the HTML for the character count
+
+We've updated the HTML for the character count.
+
+The component wrapper `data-module="govuk-character-count"` and its form group `class="govuk-form-group"` are now combined as the same `<div>`. The hint text used as the count message is now placed directly after the `<textarea>`.
+
+If you're not using Nunjucks macros, then you should:
+
+- move all classes and attributes from the form group `<div>` to the component wrapper `<div>`
+- remove the opening `<div>` and closing `</div>` tags used by the form group
+- check the count message is now directly after the `<textarea>`
+
+The following example shows some HTML and the difference once it is updated.
+
+HTML before:
+
+```html
+<div class="govuk-character-count" data-module="govuk-character-count" data-maxlength="100">
+  <div class="govuk-form-group">
+    <!-- // Label, hint, error and textarea -->
+  </div>
+  <!-- // Count message -->
+</div>
+```
+
+HTML after:
+
+```html
+<div class="govuk-form-group govuk-character-count" data-module="govuk-character-count" data-maxlength="100">
+  <!-- // Label, hint, error, textarea and count message -->
+</div>
+```
+
+Check your changes against [the character count example in the Design System](https://design-system.service.gov.uk/components/character-count/#character-count-example) to make sure you have correctly implemented them.
+
+This change was introduced in [pull request #4566: Use Character count `formGroup` as module wrapper](https://github.com/alphagov/govuk-frontend/pull/4566).
+
 ### Fixes
 
 - [#4811: Use `KeyboardEvent.key` over deprecated `KeyboardEvent.keyCode` in the Tabs component](https://github.com/alphagov/govuk-frontend/pull/4811)
--- a/packages/govuk-frontend/src/govuk/components/character-count/template.njk
+++ b/packages/govuk-frontend/src/govuk/components/character-count/template.njk
@@ -85,7 +85,12 @@
   }) -}}
 {% endset -%}
 
-<div class="govuk-character-count" {{- attributesHtml | safe }}>
+{#- Append form group attributes onto attributes set above #}
+{%- for name, value in params.formGroup.attributes %}
+  {% set attributesHtml = attributesHtml + " " + name | escape + '="' + value | escape + '"' %}
+{% endfor -%}
+
+<div class="govuk-character-count">
   {{ govukTextarea({
     id: params.id,
     name: params.name,
@@ -95,7 +100,7 @@
     value: params.value,
     formGroup: {
       classes: params.formGroup.classes,
-      attributes: params.formGroup.attributes,
+      attributes: attributesHtml,
       beforeInput: params.formGroup.beforeInput,
       afterInput: {
         html: countMessageHtml
--- a/packages/govuk-frontend/src/govuk/components/character-count/template.njk
+++ b/packages/govuk-frontend/src/govuk/components/character-count/template.njk
@@ -1,3 +1,4 @@
+{% from "../../macros/attributes.njk" import govukAttributes %}
 {% from "../../macros/i18n.njk" import govukI18nAttributes %}
 {% from "../textarea/macro.njk" import govukTextarea %}
 {% from "../hint/macro.njk" import govukHint %}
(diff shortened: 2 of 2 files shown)


packages/govuk-frontend/src/govuk/components/character-count/template.njk
```diff
@@ -2,85 +2,100 @@
...
-  {%- if params.threshold %} data-threshold="{{ params.threshold }}"{% endif %}
-  {%- if params.maxwords %} data-maxwords="{{ params.maxwords }}"{% endif %}
-  {#-
-    Without maxlength or maxwords, we can't guess if the component will count words or characters.
-    We can't guess a default textarea description to be interpolated in JavaScript
-    once the maximum gets configured there.
-    So we only add the attribute if a textarea description was explicitely provided.
-  #}
-  {%- if hasNoLimit and params.textareaDescriptionText %}
-    {{- govukI18nAttributes({
```
colinrotherham, line 18: By looping params and appending to `attributes` we can't use `govukI18nAttributes()` anymore

But since Nunjucks can both push to and loop arrays (as if they were objects) using `Object.entries()` format, it could be a great way to merge default `attributes` using https://github.com/alphagov/govuk-frontend/issues/2917

This comes from a little known Nunjucks [**for** loop](https://mozilla.github.io/nunjucks/templating.html#for) feature where these are interchangeable:

## Attributes as plain object

```njk
{{ govukButton({
  text: "Press me",
  attributes: {
    data-example: "value"
  }
}) }}
```

## Attributes as `Object.entries()` array

```njk
{{ govukButton({
  text: "Press me",
  attributes: [
    ["data-example": "value"]
  ]
}) }}
```

36degrees: There's some really clever stuff going on here – not only treating 'arrays of arrays' as objects but also the later re-assignment of `attributes` and the (ab)use of [the comma operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Comma_operator), which was new to me!

However, I think the features we're (ab)using are just a little too obscure. I think it would be hard for others who aren't familiar with them to understand what's going on, which makes it hard to read and to make changes.

It did however get me thinking about an alternative approach – what if we allowed for `formGroup.attributes` to be either an object _or a string_ in the textarea component? Removing the requirement for us to pass an object allows the character count to effectively 'take ownership' of the attributes on the form group. It also means we can continue to use the `govukI18nAttributes` macro.

I've pushed a branch to try and demonstrate what this might look like – take a look at https://github.com/alphagov/govuk-frontend/compare/character-count-form-group...spike-pass-attributes-as-string?expand=1 and see what you think?

colinrotherham: Ha, that's fair. Every Nunjucks service I've ever worked on (ab)uses the comma operator 😆

Will take a look 👀

colinrotherham: Yep that's a nice enhancement that gets around my "appending to objects" blocker 🙌

Got my support, how do you want to proceed?

36degrees: I'm not sure – I think we will want to eventually make this consistent, so anywhere you can pass `attributes` as an object you can also pass a string? I don't think we have to do that to unblock this PR, but we should probably raise an issue and see where else we can apply the same 'fix'.

I'm also wondering how it should be documented given we only allow for a single 'type' in the YAML – possibly we keep it as a `type: object` but mention you can pass a string in the description?

Happy for you to pick this up and run with it if you're happy to update this PR to adopt that approach?

36degrees: I've put the associated issues back into in progress for now.

colinrotherham: Issue raised for this in https://github.com/alphagov/govuk-frontend/issues/4666

colinrotherham: This was a great suggestion by the way 🙌

Made possible by https://github.com/alphagov/govuk-frontend/pull/4742 and added in https://github.com/alphagov/govuk-frontend/pull/4566/commits/779004c4fa1f11a2af517e79fea55fce565f442d

querkmachine (approved):
This change makes sense to me. (Probably helps that having random wrapping elements outside of `govuk-form-group`s was a minor bugbear of mine.)

I vaguely ponder if this is a breaking change, but I don't think it necessarily is. Regardless, it probably needs a "Suggested change" changelog entry to avoid any surprises if a team has written anything that anticipates a specific HTML structure.

CHANGELOG.md
```diff
@@ -4,6 +4,41 @@ For advice on how to use these release notes see [our guidance on staying up to
...
+- move the `govuk-form-group` class from the form group to the component wrapper `<div>`
+- move the `govuk-form-group--error` class (optional) from the form group to the component wrapper `<div>`
+- move any other custom classes and attributes from the form group to the component wrapper `<div>`
+- remove the opening `<div>` and closing `</div>` tags previously used by the form group
+
+For example, see how the HTML below:
+
+```html
+<div class="govuk-character-count" data-module="govuk-character-count" data-maxlength="100">
+  <div class="form-group form-group--error">
```
domoscargin, line 24: Should we make these classes match the steps above? (ie: `govuk-form-group` and `govuk-form-group--error`)

colinrotherham: 🤦‍♂️

CHANGELOG.md
```diff
@@ -4,6 +4,41 @@ For advice on how to use these release notes see [our guidance on staying up to
...
+
+would now be:
+
+```html
+<div class="form-group form-group--error govuk-character-count" data-module="govuk-character-count" data-maxlength="100">
+  <!-- // -->
+</div>
+```
+
+[Check your changes against the character count example in the Design System](https://design-system.service.gov.uk/components/character-count/#character-count-example) to make sure you have correctly implemented them.
```
claireashworth, line 38: I'd make this link a little shorter
```suggestion
Check your changes against [the character count example in the Design System](https://design-system.service.gov.uk/components/character-count/#character-count-example) to make sure you have correctly implemented them.
```

CHANGELOG.md
```diff
@@ -4,6 +4,41 @@ For advice on how to use these release notes see [our guidance on staying up to
...
+The component wrapper `data-module="govuk-character-count"` and its form group `class="govuk-form-group"` are now combined as the same `<div>`. The hint text used as the count message is now placed directly after the `<textarea>`.
+
+If you're not using Nunjucks macros, then you should:
+
+- move the `govuk-form-group` class from the form group to the component wrapper `<div>`
+- move the `govuk-form-group--error` class (optional) from the form group to the component wrapper `<div>`
+- move any other custom classes and attributes from the form group to the component wrapper `<div>`
+- remove the opening `<div>` and closing `</div>` tags previously used by the form group
+
+For example, see how the HTML below:
```
claireashworth, line 20: I'd change the explanatory text so it isn't split between the code examples. Maybe say something like 'The following example shows some HTML and the difference once it is updated.' Then you can always add something like 'HTML before:' and 'HTML after:' to clearly divide them.

claireashworth (approved):

Issue #924, opened 2018-07-23
Add an icon next to an input using the govukInput macro


dashouse:
I was trying to replicate this design with the Input macro:

<img width="374" alt="screen shot 2018-07-23 at 12 49 09" src="https://user-images.githubusercontent.com/23356842/43075365-821238e2-8e78-11e8-9849-977ac2c20c11.png">

In the end I did this with HTML like this

```
<div class="govuk-form-group">
  <label class="govuk-label govuk-label--s" for="cvv">
    Card security code
  </label>
  <span id="cvv-hint" class="govuk-hint">
    The last 3 digits on the back of the card
  </span>
  <input class="govuk-input govuk-input--width-4 govuk-!-display-inline-block" id="cvv" name="post-code" type="text">
  <span class="govuk-!-display-inline-block" style="vertical-align: middle; height: 40px;">
    <img alt="image of the back of a bank card" src="https://png.icons8.com/ios/1600/bank-card-back-side.png" style="height: 40px;">
  </span>
</div>
```

My expectation was that I would be able to import the label, hint, error and input components  separately, add in an additional image and wrap it in `govuk-form-group` - This works, however I was unable to remove the second `govuk-form-group` that comes with the govukInput macro.

igloosi:
if we added `caller` to input the macro would look like that:
```
{% call govukInput({
    label: {
      text: 'Card security code',
      classes: 'govuk-label--s'
    },
    hint: {
      text: 'The last 3 digits on the back of the card'
    },
    classes: 'govuk-input--width-4 govuk-!-display-inline-block',
    id: 'cvv'
  }) %}
   <span class="govuk-!-display-inline-block" style="vertical-align: middle; height: 40px;">
    <img alt="image of the back of a bank card" src="https://png.icons8.com/ios/1600/bank-card-back-side.png" style="height: 40px;">
  </span>
  {% endcall %}
```

nickcolley:
I guess one pseudo code could be:

```jsx
{% call fieldset() %}
    {{ label() }}
    {{ hint() }}
    {{ input() }}
    <span class="govuk-!-display-inline-block" style="vertical-align: middle; height: 40px;">
        <img alt="image of the back of a bank card" src="https://png.icons8.com/ios/1600/bank-card-back-side.png" style="height: 40px;">
     </span>
{% endcall %}
```

igloosi:
yeah that's the alternative.
[...]

PR #4567, opened 2023-12-14
Add `beforeInput(s)` and `afterInput(s)` options to form groups

colinrotherham:
This PR adds support for component before/after content

This work is based on a fix suggested by @36degrees in https://github.com/alphagov/govuk-frontend/issues/1602#issuecomment-913740373 to solve these issues:

* https://github.com/alphagov/govuk-frontend/issues/924
* https://github.com/alphagov/govuk-frontend/issues/1602

>I don't think it'd have to be specific to hints though – we could allow e.g. an 'arbitrary' `formGroup.beforeHtml` / `formGroup.afterHtml` to be passed, which _might_ also be useful for scenarios like those described in #924

It is a supporting change to enable "components within components" as used by:

* https://github.com/alphagov/govuk-frontend/pull/4566
* https://github.com/alphagov/govuk-design-system/issues/3479

# Nunjucks options

I've picked "input" for all form controls (input, select, textarea) as we do with [**Error summary** `$input`](https://github.com/alphagov/govuk-frontend/blob/396996fba09f8ac6d08276f9b7c28c9739ce753d/packages/govuk-frontend/src/govuk/components/error-summary/error-summary.mjs#L104-L112)

All options considered might be:

1. `beforeInput`/`afterInput`
2. `beforeControl`/`afterControl`
3. `beforeField`/`afterField`

Nunjucks options have also been varied for components with single or multiple fields:

## Components with single form field

Use options `beforeInput` and `afterInput` for components with a single field

```js
govukInput({
  formGroup: {
    beforeInput: {
      html: "example HTML",
      text: "example text"
    },
    afterInput: {
      html: "example HTML",
      text: "example text"
    },
  }
})
```

## Components with single form field

Use options `beforeInputs` and `afterInputs` for components with multiple fields

```js
govukRadios({
  formGroup: {
    beforeInputs: {
      html: "example HTML",
      text: "example text"
    },
    afterInputs: {
      html: "example HTML",
      text: "example text"
    },
  }
})
```

☝️ Note that **Password input** will have an input + button combo, is the option name still suitable?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,27 @@ For advice on how to use these release notes see [our guidance on staying up to 
 
 ## Unreleased
 
+### New features
+
+#### Insert custom HTML into component form group wrappers
+
+You can now insert custom HTML into form group wrappers for all components with form fields.
+
+```njk
+govukInput({
+  formGroup: {
+    beforeInput: {
+      html: "example"
+    },
+    afterInput: {
+      html: "example"
+    },
+  }
+})
+```
+
+This change was introduced in [pull request #4567: Add `beforeInput(s)` and `beforeInput(s)` options to form groups](https://github.com/alphagov/govuk-frontend/pull/4567).
+
 ## 5.1.0 (Feature release)
 
 To install this version with npm, run `npm install govuk-frontend@5.1.0`. You can also find more information about [how to stay up to date in our documentation](https://frontend.design-system.service.gov.uk/staying-up-to-date/#updating-to-the-latest-version).
--- a/packages/govuk-frontend/src/govuk/components/character-count/character-count.yaml
+++ b/packages/govuk-frontend/src/govuk/components/character-count/character-count.yaml
@@ -55,6 +55,32 @@ params:
         type: object
         required: false
         description: HTML attributes (for example data attributes) to add to the form group.
+      - name: beforeInput
+        type: object
+        required: false
+        description: Content to add before the textarea used by the character count component.
+        params:
+          - name: text
+            type: string
+            required: true
+            description: Text to add before the textarea. If `html` is provided, the `text` option will be ignored.
+          - name: html
+            type: string
+            required: true
+            description: HTML to add before the textarea. If `html` is provided, the `text` option will be ignored.
+      - name: afterInput
+        type: object
+        required: false
+        description: Content to add after the textarea used by the character count component.
+        params:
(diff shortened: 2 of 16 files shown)


querkmachine:
Given recent discussions about the limitations of components that inherit other components (i.e. character count and EtP), I'm quite interested in this! 

I'm less sure how much benefit there is to doing it this universally, though.

Being able to insert custom HTML after an input/before the form group closes seems like it will have some utility; but I can't really see much benefit in putting something at the start of the group before the label, hint and error message (or fieldset). 

I don't think doing that's a behaviour we'd expect, and having it as an option doesn't seem like it'd help users do anything they can't usually already do by putting their custom HTML prior to the form group and potentially futzing with margins. 

It feels like what we might actually want is the ability to inject HTML before and after the `input` element, rather than the form group as a whole. I'm not sure how well that might work in cases of radios or checkboxes, or text inputs using prefixes/suffixes, but it seems closer to what is desired in the linked issues.

colinrotherham:
That's great @querkmachine

To solve https://github.com/alphagov/govuk-frontend/issues/1602 we only really need to add content after the `<textarea>` 

So I'm open to narrowing it down and/or chuck me some ideas for renaming the options

querkmachine:
I agree that this could solve #1602, but I don't think it's necessarily a fix for #924, and I'm also not sure there are many situations outside of #1602 where having these parameters available would facilitate solving a problem. 

For example, if I were wanting to add a icon before a text input, I wouldn't be able to do that here because `beforeHtml` comes before everything (label, hint and error message included) and `afterHtml` is after the input. The content order could be reversed with flexbox (although WCAG might preclude that), but the icon and input still wouldn't share a wrapper and we can't insert one ourselves with those parameters, making it similarly difficult.

This similarly wouldn't really help us solve other issues with nested components, like in the password input we're currently developing.

Code-wise I don't really see anything harmful with having the options, I just also don't see much use in them when compared to something akin to a `beforeInput` and `afterInput`—though I realise those may not work so well for controls like radios and checkboxes.

Would be interested in having this shopped around to other devs for their thoughts.

colinrotherham:
@querkmachine Using `beforeInput` and `afterInput` works for me, would that help?

When inserting Character count's `<textarea>` your suggestion would work identically

For radios and checkboxes we'd see the custom HTML move within the `<fieldset>` but that's alright?

PR #4568, opened 2023-12-15
Bug fixes for Sass `gulp watch` tasks

colinrotherham:
Couple of bug fixes with more info in the commit messages

* Prevent `gulp watch` early Sass compile after `clean` script runs
* Ensure `gulp watch` task waits for PostCSS to complete task

See https://github.com/alphagov/govuk-frontend/pull/3910#discussion_r1254347310 for previous discussion

### DIFF
--- a/packages/govuk-frontend-review/tasks/watch.mjs
+++ b/packages/govuk-frontend-review/tasks/watch.mjs
@@ -43,7 +43,13 @@ export const watch = (options) =>
     task.name('compile:scss watch', () =>
       gulp.watch(
         ['**/*.scss', join(paths.package, 'dist/govuk/all.scss')],
-        { cwd: options.srcPath },
+        {
+          cwd: options.srcPath,
+
+          // Prevent early Sass compile by ignoring delete (unlink) event
+          // when GOV.UK Frontend runs the `clean` script before build
+          events: ['add', 'change']
+        },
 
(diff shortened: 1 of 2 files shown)


packages/govuk-frontend-review/tasks/watch.mjs
```diff
@@ -38,7 +43,14 @@ export const watch = (options) =>
...
       gulp.watch(
         ['**/*.scss', join(paths.package, 'dist/govuk/all.scss')],
-        { cwd: options.srcPath },
+        {
+          awaitWriteFinish: true,
+          cwd: options.srcPath,
+
+          // Prevent early Sass compile by ignoring delete (unlink) event
+          // when GOV.UK Frontend runs the `clean` script before build
+          events: ['add', 'change']
```
romaricpascal, line 52: **note** Thinking this will also skip `unlink` events for files matching `**/*.scss`. Brings a slim risk of not recompiling straigh after having mistakenly deleted a `.scss` file from the review app but not the corresponding `@import`, which would delay the realisation of that deletion. This would hopefully be caught at commit time (if adding file by file) or by CI so that skiping of all `unlink` seems our least worst option there. Shame the [`ignored` option](https://gulpjs.com/docs/en/api/watch/#options) doesn't get us enough info 😢

PR #4603, opened 2024-01-04
Remove v5 warning from readme

querkmachine:
All of the code-related work for Frontend v5.0.0 has now been released, as such, it feels like we can now remove this warning from the project readme.


owenatgov (approved):
We did the thing!

PR #4624, opened 2024-01-09
Add changelog entry for #4562

owenatgov:
Adds a changelog entry for https://github.com/alphagov/govuk-frontend/pull/4562


colinrotherham:
Thanks @owenatgov, we didn't document [`--govuk-frontend-version`](https://github.com/alphagov/govuk-frontend/pull/3318) either. Do we need to?

For this entry though, do we need to explain that custom Sass `$govuk-breakpoints` edits will be reflected?

owenatgov:
@colinrotherham On the premise of making this a feature changelog, it sorta depends what we want. Are we bothered about users not knowing that we have CSS custom properties? Personally I'm not gonna kick off if we convert this to just a fix and call it a day.

colinrotherham:
@owenatgov Yeah, it's a fix but probably still needs an entry 😆

E.g. Users that override Sass `$govuk-breakpoints` will now see **Tabs** and **Header** components use them

owenatgov:
@colinrotherham I'm fine with keeping the custom props private. Shall we just make it a fix for ease? Just throw in the title of the PR and add it to the fixes list?

colinrotherham (approved):

PR #4628, opened 2024-01-12
Remove `aria-labelledby` from accordion sections

### LINKED ISSUE #2472: Accordion - Elements must only use allowed ARIA attributes
## Description of the issue
Axe DevTools reports "ARIA attribute: aria-labelledby is not well supported. Use a different role attribute or element."

Element location 
`#accordion-default-content-1`

Need to ensure ARIA attributes are allowed for an element's role

Run the accessibility checker with an expanded accordion

Error should not be reported for the expanded accordion

GOV.UK Frontend v3.14.0
https://design-system.service.gov.uk/components/accordion/default/index.html

36degrees:
Adding `aria-labelledby` to a `<div>` with no explicit role is [explicitly disallowed by the 'ARIA in HTML' specification][1] and in our testing [provides little or no benefit to users, including users of assistive technologies][2].

We’ve explored using `<section>` elements for each accordion section (for which a label would make sense), however best practise is to limit the use of region landmarks [‘as too many landmarks can dilute their usefulness in quickly navigating to important areas of a page’][3].

We don’t have clear evidence that marking accordion sections up as regions is useful, so more research is needed in this area.

There has also been [a suggestion that we could use `<article>` elements][4], but it would depend on how the accordion is being used.

We may still want to pursue these options in the future, but until we make a decision on whether the sections should have a semantic role remove the `aria-labelledby` attribute – it’s trivial to reintroduce it later.

Closes #2472.

[1]: https://www.w3.org/TR/html-aria/#docconformance-naming
[2]: https://github.com/alphagov/govuk-frontend/issues/2472#issuecomment-1766282547
[3]: https://www.scottohara.me/blog/2018/03/03/landmarks.html#region
[4]: https://github.com/alphagov/govuk-frontend/issues/2472#issuecomment-1878807875


36degrees:
Merging (and releasing) this should allow us to remove the related issue from our accessibility statement.

owenatgov (commented):
You just need to update the tests that look for the `aria-labelledby` and then this is good to go.

36degrees:
All these pesky tests, getting in the way of me ripping out bits of code – what have they ever done for us, eh? 😉

packages/govuk-frontend/src/govuk/components/accordion/accessibility.puppeteer.test.mjs
```diff
@@ -4,18 +4,6 @@ import { getExamples } from '@govuk-frontend/lib/components'
 describe('/components/accordion', () => {
   let axeRules
```
owenatgov, line 5: We're not setting this anymore, does it still need to be here?

owenatgov (commented):
One little query then I'm happy to call this done

owenatgov (approved):
Lovely

PR #4632, opened 2024-01-12
Omit nav element and menu button when header `navigation` is an empty array

36degrees:
Ensure that the `navigation` param of the header is non-empty rather than truthy. This avoids a scenario where the wrapper, nav and menu button are outputted with an empty navigation `<ul>` when the `navigation` param is an empty array (`[]`):

```html
<div class="govuk-header__content">
  <nav aria-label="Menu" class="govuk-header__navigation">
    <button type="button" class="govuk-header__menu-button govuk-js-header-toggle" aria-controls="navigation" hidden>Menu</button>
    <ul id="navigation" class="govuk-header__navigation-list"></ul>
  </nav>
</div>
```

This issue is hard to spot as there is no visual change on tablet breakpoints and above – only on the mobile breakpoint where the menu button is visible (but clicking it only changes the arrow direction).

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/header/template.test.js
+++ b/packages/govuk-frontend/src/govuk/components/header/template.test.js
@@ -249,6 +249,29 @@ describe('header', () => {
     })
   })
 
+  describe.each([
+    { message: 'without navigation', exampleName: 'default' },
+    { message: 'with empty navigation', exampleName: 'empty navigation array' }
+  ])('$message', ({ exampleName }) => {
+    let $
+
+    beforeAll(() => {
+      $ = render('summary-list', examples[exampleName])
+    })
+
+    it('does not include a menu button', async () => {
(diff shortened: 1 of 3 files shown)


packages/govuk-frontend/src/govuk/components/header/template.test.js
```diff
@@ -249,6 +249,29 @@ describe('header', () => {
     })
   })
 
+  describe.each([
+    { message: 'without navigation', exampleName: 'default' },
+    { message: 'with empty navigation', exampleName: 'empty navigation array' }
+  ])('$message', ({ exampleName }) => {
```
romaricpascal, line 255: That worked nicely to factor both tests 😍

romaricpascal (approved):
Nice one! Thanks for making the fix 😊 ⛵

Issue #4168, opened 2023-09-04
Remove `input type="button"` functionality from button component


romaricpascal:

colinrotherham:
For context, typically seen for server-side or progressively enhanced buttons

We'd want to inform teams that that `type="button"` can also have name/value

But also that `<input type="submit">` is no longer required for legacy browser support reasons

```js
govukButton({
  text: "Accept analytics cookies",
  element: "input",
  name: "cookies[analytics]",
  value: "yes"
})

govukButton({
  text: "Reject analytics cookies",
  element: "input",
  name: "cookies[analytics]",
  value: "no"
})
```

querkmachine:
I'm a bit iffy on removing the `type` option entirely. I think there are legitimate situations in which a team wants to insert a button that has `type="button"` using the Nunjucks macro, without necessarily having to breach the ideal of progressive enhancement. 

---

For a hypothetical example, maybe a team has developed a progressively enhanced postcode lookup form. 

If JavaScript is not available, it only shows a country and post input and the results are rendered on a separate page. If JS is available, both the search fields and results are displayed within the same page. 

Because the HTML for the latter could be quite complex, they've decided not to construct and inject the HTML from within JavaScript, but to instead use a `<template>` element and have the enhanced form structure sent from the server. When JS initialises, it takes the contents of the `<template>` and replaces the non-enhanced form. This approach lets them use Nunjucks for all of their components too, so it's a win-win for maintainability and separation of concerns. 

```
Select country of residence
[United Kingdom            ↓]

Enter postcode
[TE57 1NG   ] [Search]

Select your address
[Flat 1, 23 Fake Street    ↓]

[Save and continue]
```

But... they now have no choice but for the 'Search' button to be `type="submit"`, and that's causing problems. By default, it will perform a postback to the server, even though that's not what they want it to do. 

They could use JavaScript to stop it posting the form and to perform the asyncronous search instead, but this still leaves a problem...

---

When focused on most input types, the keyboard's <kbd>Enter</kbd> key will, when pressed, submit the form. If there is more than one submit button within a form, it will use the first one it encounters. In this situation, this is now the 'Search' button, not the 'Save and continue' button. 

In this case, a keyboard-heavy or keyboard-only user may tab from the postcode field to the address field, select their address from the list, and seeing it as the last input on the page, press <kbd>Enter</kbd>… and nothing will happen. Or worse, the search will get performed again, resetting the result field and requiring them to go and select their address again.

[...]

PR #4646, opened 2024-01-15
Deprecate `element` parameter on button component

querkmachine:
As per #4168, mark the `element` parameter as deprecated in the Nunjucks parameter docs and in the changelog.


owenatgov (approved):
Does the job!

Issue #2786, opened 2022-08-19
Deprecate 14 on the font scale


owenatgov:
## What
Deprecate 14 as an option in our font sass mixins as well as the override class `govuk-!-font-size-14`.

## Why
So that we are communicating that we will be removing these in 6.0.

Part of https://github.com/alphagov/govuk-design-system/issues/2289

## Who needs to work on this
Developers

PR #4649, opened 2024-01-16
Deprecate 14 on the type scale

owenatgov:
## What/Why
Deprecates 14 on the responsive typography scale along with the classes `govuk-body-xs` and `govuk-!-font-size-14` as we're removing this option from the typography scale in 6.0.

Fixes https://github.com/alphagov/govuk-frontend/issues/2786.

## Thoughts
Deprecating a key in a map is tricky. The solution I've gone for here is that `govuk-font-size` looks for a deprecation key and a deprecation message in the responsive scale. If they're there, it outputs a warning and removes those keys from the font map in the mixin so that the deprecation keys don't break the loop. To get around this internally, I've added a `_14` type scale point which replicates 14.

I've done this so that the deprecation is intrinsically tied to the scale value that we're removing. The downside is that users can override `$govuk-typography-scale` to get around the warning, however they could do that anyway with warning suppression.

The other options I explored are:

- Throw a warning if `$size == 14`. This feels a bit fragile to me as it's just looking at the map name. There's a theoretical, very unlikely case where a user has a custom type scale map where one of the keys is 14 but the actual returned font sizes are higher than 14.
- Throw a warning if the returned size from a given font-map is below 16px (and rem equivalent). This solution also infers that at 6.0 we'd throw an error if the same thing happened to block any user using font sizes below 16px regardless of custom type scales. Like the last option, this feels like it's disrupting the customisation of govuk-frontend and doesn't tie the actual scale value we're removing to the deprecation. I don't know how I feel about the error idea when we reach 6.0. I'm interested in views on this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -114,6 +114,17 @@ In the future, the component will automatically use the use the `<a>` element if
 
 This change was introduced in [pull request #4646: Deprecate `element` parameter on button component](https://github.com/alphagov/govuk-frontend/pull/4646)
 
+#### Stop using `govuk-body-xs`, `govuk-!-font-size-14` and '14' on the type scale
+
+We have deprecated point 14 (14px large screens, 12px small screens) on the GOV.UK Frontend responsive type scale, including font override classes that use point 14:
+
+- `govuk-body-xs`
+- `govuk-!-font-size-14`
+
+We will be removing these classes and point 14 on the type scale in GOV.UK Frontend release v6.0.0. From the next major release, you will no longer be able to call the Sass mixins `govuk-font` or `govuk-font-size` with `$size` set to '14'.
+
+This change was introduced in [#4649: Deprecate 14 on the type scale](https://github.com/alphagov/govuk-frontend/pull/4649)
+
 ### Fixes
 
 We've made fixes to GOV.UK Frontend in the following pull requests:
--- a/packages/govuk-frontend/src/govuk/helpers/_typography.scss
+++ b/packages/govuk-frontend/src/govuk/helpers/_typography.scss
@@ -157,10 +157,19 @@
 
 @mixin govuk-font-size($size, $line-height: false, $important: false) {
   @if not map-has-key($govuk-typography-scale, $size) {
-    @error "Unknown font size `#{$size}` - expected a point from the typography scale.";
+    @error "Unknown font size `#{$size}` - expected a point from the type scale.";
   }
 
   $font-map: map-get($govuk-typography-scale, $size);
+
+  // Check for a deprecation within the type scale
+  $deprecation: map-get($font-map, "deprecation");
+
+  @if $deprecation {
+    @include _warning(map-get($deprecation, "key"), map-get($deprecation, "message"));
+    // remove the deprecation map keys so they do not break the breakpoint loop
+    $font-map: map-remove($font-map, "deprecation");
+  }
 
   @each $breakpoint, $breakpoint-map in $font-map {
     $font-size: map-get($breakpoint-map, "font-size");
--- a/packages/govuk-frontend/src/govuk/helpers/_typography.scss
+++ b/packages/govuk-frontend/src/govuk/helpers/_typography.scss
@@ -99,14 +99,14 @@
 
 /// Font size and line height helper
 ///
-/// @param {Number} $size - Point from the spacing scale (the size as it would
+/// @param {Number} $size - Point from the type scale (the size as it would
 ///   appear on tablet and above)
 /// @param {Number} $override-line-height [false] - Non responsive custom line
 ///   height. Omit to use the line height from the font map.
 /// @param {Boolean} $important [false] - Whether to mark declarations as
 ///   `!important`.
 ///
-/// @throw if `$size` is not a valid point from the spacing scale
+/// @throw if `$size` is not a valid point from the type scale
 ///
 /// @access public
 ///
@@ -144,14 +144,14 @@
 ///   )
 /// );
 ///
-/// @param {Number} $size - Point from the spacing scale (the size as it would
-///   appear on tablet and above)
+/// @param {Number | String} $size - Point from the type scale (the size as
+///   it would appear on tablet and above)
 /// @param {Number} $line-height [false] - Non responsive custom line
 ///   height. Omit to use the line height from the font map.
 /// @param {Boolean} $important [false] - Whether to mark declarations as
 ///   `!important`.
 ///
-/// @throw if `$size` is not a valid point from the spacing scale
+/// @throw if `$size` is not a valid point from the type scale
 ///
 /// @access public
 
@@ -212,14 +212,15 @@
 
 /// Font helper
 ///
-/// @param {Number | Boolean} $size Point from the spacing scale (the size as it
-///   would appear on tablet and above). Use `false` to avoid setting a size.
+/// @param {Number | Boolean | String} $size Point from the type scale (the
+///   size as it would appear on tablet and above). Use `false` to avoid setting
+///   a size.
 /// @param {String} $weight [regular] - Weight: `bold` or `regular`
 /// @param {Boolean} $tabular [false] - Whether to use tabular numbers or not
 /// @param {Number} $line-height [false] - Line-height, if overriding the
 ///   default
 ///
-/// @throw if `$size` is not a valid point from the spacing scale (or false)
+/// @throw if `$size` is not a valid point from the type scale (or false)
 ///
 /// @access public
 
--- a/packages/govuk-frontend/src/govuk/overrides/_typography.scss
+++ b/packages/govuk-frontend/src/govuk/overrides/_typography.scss
@@ -5,9 +5,20 @@
   // typography scale eg .govuk-\!-font-size-80
   //
   // govuk-!-font-size-14 is deprecated
-  @each $size in map-keys($govuk-typography-scale) {
-    .govuk-\!-font-size-#{$size} {
-      @include govuk-font-size($size, $important: true);
+  @each $size, $font-map in $govuk-typography-scale {
+    // Suppress class if the map key is a string with an underscore eg: _14 to
+    // avoid classes like govuk-!-font-size-_14 getting generated
+    @if type-of($size) == "number" or str-slice(#{$size}, 1, 1) != "_" {
+      .govuk-\!-font-size-#{$size} {
+        $font-map: map-get($govuk-typography-scale, $size);
+
+        // Add underscore to deprecated typography scale keys
+        @if map-has-key($font-map, "deprecation") {
+          $size: _#{$size};
+        }
+
+        @include govuk-font-size($size, $important: true);
+      }
     }
   }
 
--- a/packages/govuk-frontend/src/govuk/overrides/_typography.scss
+++ b/packages/govuk-frontend/src/govuk/overrides/_typography.scss
@@ -3,6 +3,8 @@
 
   // Generate typography override classes for each responsive font map in the
   // typography scale eg .govuk-\!-font-size-80
+  //
+  // govuk-!-font-size-14 is deprecated
   @each $size in map-keys($govuk-typography-scale) {
     .govuk-\!-font-size-#{$size} {
       @include govuk-font-size($size, $important: true);
--- a/packages/govuk-frontend/src/govuk/settings/_typography-responsive.scss
+++ b/packages/govuk-frontend/src/govuk/settings/_typography-responsive.scss
@@ -151,6 +151,11 @@ $govuk-typography-scale: (
     print: (
       font-size: 12pt,
       line-height: 1.2
+    ),
+    deprecation: (
+      key: "govuk-typography-scale-14",
+      message: "14 on the type scale is deprecated and will be removed as " +
+        "a possible option in the next major version."
     )
   )
 ) !default;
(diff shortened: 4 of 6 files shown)


owenatgov:
Oh dear, that percy test is pretty damning. Will fix.

owenatgov:
Fixed 😎

colinrotherham (commented):
What a nice clever way to deprecate Sass map values

If everyone's happy with the approach, I've only got a few content/naming suggestions

CHANGELOG.md
```diff
@@ -114,6 +114,17 @@ In the future, the component will automatically use the use the `<a>` element if
 
 This change was introduced in [pull request #4646: Deprecate `element` parameter on button component](https://github.com/alphagov/govuk-frontend/pull/4646)
 
+#### Stop using `govuk-body-xs`, `govuk-!-font-size-14` and '14' on the typography scale
+
+We have deprecated '14' on the responsive typography scale as well as classes within GOV.UK Frontend which use 14:
```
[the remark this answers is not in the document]
owenatgov: I'm gonna have a think about this. I want to retain a flat `14` somewhere in there to communicate that it's the 14 point on the scale rather than _necessarily_ 14px (cause it drops down to 12px on small screens). I don't know how actually valuable that is though. Aside: this would be simpler if we had t-shirt sizing aliases set up.

packages/govuk-frontend/src/govuk/settings/_typography-responsive.scss
```diff
@@ -140,6 +140,23 @@ $govuk-typography-scale: (
...
+    ),
+    tablet: (
+      font-size: 14px,
+      line-height: 20px
+    ),
+    print: (
+      font-size: 12pt,
+      line-height: 1.2
+    ),
+    deprecation-key: "type-scale-14",
```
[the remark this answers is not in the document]
owenatgov: Both excellent ideas.

packages/govuk-frontend/src/govuk/helpers/_typography.scss
```diff
@@ -214,7 +224,7 @@
 ///
 /// @access public
 
-@mixin govuk-font($size, $weight: regular, $tabular: false, $line-height: false) {
+@mixin govuk-font($size, $weight: regular, $tabular: false, $line-height: false, $ignore-warning: false) {
```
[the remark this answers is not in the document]
owenatgov: Ah it was! This is a carry over from early ideas to deprecate the thing. I'll get rid.

CHANGELOG.md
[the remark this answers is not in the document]
owenatgov: I've added something to try and clarify that 14 the scale value ~= 14px the font size. Let me know what you think.

packages/govuk-frontend/src/govuk/overrides/_typography.scss
```diff
@@ -3,8 +3,13 @@
 
   // Generate typography override classes for each responsive font map in the
   // typography scale eg .govuk-\!-font-size-80
+  //
+  // govuk-!-font-size-14 is deprecated
   @each $size in map-keys($govuk-typography-scale) {
     .govuk-\!-font-size-#{$size} {
+      @if $size == 14 {
+        $size: _14;
+      }
```
colinrotherham, line 12: Do we need to update the `govuk-font-size()` SassDoc type to allow strings?

```scss
/// @param {String | Number} $size - Point from the spacing scale (the size as
///   it would appear on tablet and above)
```

colinrotherham: We should probably check for the `"deprecation"` key here too?

Other sizes could be deprecated in future so should avoid hard coding them

```suggestion
  @each $size, $font-map in $govuk-typography-scale {
    $font-map: map-get($govuk-typography-scale, $size);

    // Add underscore to deprecated typography scale keys
    @if map-has-key($font-map, "deprecation") {
      $size: _#{$size};
    }
```

packages/govuk-frontend/src/govuk/settings/_typography-responsive.scss
```diff
@@ -140,6 +140,26 @@ $govuk-typography-scale: (
...
+      line-height: 1.2
+    ),
+
+    deprecation: (
+      key: "govuk-typography-scale-14",
+      message: "14 on the typography scale is deprecated and will be removed as " +
+        "a possible option in the next major version."
+    )
+  ),
+  _14: (
```
colinrotherham, line 162: Wondering whether a duplicate `_14` key/map pair is 100% necessary?

For example, we could automatically remove the underscores:

```scss
// Remove underscore from deprecated typography scale keys
@if type-of($size) == "string" and str-slice($size, 1, 1) == "_" {

  // Remove underscore, but still a string
  $size: str-slice($size, 2);
}
```

But it seemed to complicated to convert `_14` (string) into `14` (number) for the lookup to work:

```scss
// The $size argument type must match the map key
$font-map: map-get($govuk-typography-scale, $size);
```

CHANGELOG.md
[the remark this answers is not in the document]
colinrotherham: Ah I know it doesn't, but we kinda say it does with that [guidance example above](https://github.com/alphagov/govuk-frontend/pull/4649#discussion_r1469308411)

Then in our Sass documentation we call them points?

>Point from the spacing scale (the size as it would appear on tablet and above)

Maybe we embrace that and spell it out as the guidance does?

>We have deprecated 14 point (14px large screens, 12px small screens) on the GOV.UK Frontend responsive typography scale, including font override classes that use 14 point:

packages/govuk-frontend/src/govuk/settings/_typography-responsive.scss
owenatgov: I had a go at implementing it but can't get past the variable type at the end without it getting just a bit too complicated. Do you have some other ideas for removing the extra scale value? Otherwise I think this is probably the simplest way to get the thing done, even if it isn't the best.

colinrotherham: Nah, leave it the . If the keys were already strings we'd be fine

colinrotherham (changes requested):
Looking good Owen

Presume we need to suppress `.govuk-!-font-size-_14` being output in the CSS?

The old (deprecated) class `.govuk-!-font-size-14` is still there

packages/govuk-frontend/src/govuk/overrides/_typography.scss
colinrotherham: Going to unresolve this one based on https://github.com/alphagov/govuk-frontend/pull/4649#pullrequestreview-1856265873

Both approaches renamed `14` to `_14`, is that where the extra class `.govuk-!-font-size-_14` comes from?

Or because we're looping both `14` and `_14` in `$govuk-typography-scale`?

owenatgov: It's the latter. Just checking a solution now.

CHANGELOG.md
```diff
@@ -114,6 +114,17 @@ In the future, the component will automatically use the use the `<a>` element if
 
 This change was introduced in [pull request #4646: Deprecate `element` parameter on button component](https://github.com/alphagov/govuk-frontend/pull/4646)
 
+#### Stop using `govuk-body-xs`, `govuk-!-font-size-14` and '14' on the typography scale
```
claireashworth, line 117: we're trialling using 'type scale' instead of 'typography scale', so I'm changing for consistency across docs.

```suggestion
#### Stop using `govuk-body-xs`, `govuk-!-font-size-14` and '14' on the type scale
```

CHANGELOG.md
```diff
@@ -114,6 +114,17 @@ In the future, the component will automatically use the use the `<a>` element if
...
+
+We have deprecated 14 on the responsive typography scale as well as classes within GOV.UK Frontend which set text to 14px:
```
claireashworth, line 119:
```suggestion
We have deprecated 14 on the responsive type scale, as well as classes within GOV.UK Frontend which set text to 14px:
```

CHANGELOG.md
```diff
@@ -114,6 +114,17 @@ In the future, the component will automatically use the use the `<a>` element if
...
+
+- `govuk-body-xs`
+- `govuk-!-font-size-14`
+
+We will be removing these classes and 14 on the typography scale in a future major release (6.0.0). From the next major release, you will no longer be able to call the Sass mixins `govuk-font` or `govuk-font-size` with `$size` set to '14'.
```
claireashworth, line 124:
```suggestion
We will be removing these classes and 14 on the type scale in GOV.UK Frontend release v6.0.0. From the next major release, you will no longer be able to call the Sass mixins `govuk-font` or `govuk-font-size` with `$size` set to '14'.
```

CHANGELOG.md
```diff
@@ -114,6 +114,17 @@ In the future, the component will automatically use the use the `<a>` element if
...
+
+This change was introduced in [#4649: Deprecate 14 on the typography scale](https://github.com/alphagov/govuk-frontend/pull/4649)
```
claireashworth, line 126: I'm not suggesting we change the 'typography scale' here because that would potentially mean changing the PR title as well.

packages/govuk-frontend/src/govuk/overrides/_typography.scss
owenatgov: I'm going to re-resolve this! CSS diff is now zero.

packages/govuk-frontend/src/govuk/overrides/_typography.scss
```diff
@@ -3,9 +3,22 @@
...
   // typography scale eg .govuk-\!-font-size-80
-  @each $size in map-keys($govuk-typography-scale) {
-    .govuk-\!-font-size-#{$size} {
-      @include govuk-font-size($size, $important: true);
+  //
+  // govuk-!-font-size-14 is deprecated
+  @each $size, $font-map in $govuk-typography-scale {
+    // suppress class if the map key is a string eg: _14 to avoid classes like
+    // govuk-!-font-size-_14 getting generated
+    @if type-of($size) == "number" {
```
colinrotherham, line 11: Sorry @owenatgov, but what if a fork of GOV.UK Frontend has switched to t-shirt sizing?

This change will stop their `"string"` key names from working

How about we still allow strings but only if the first character is NOT an underscore?

```suggestion
    @if type-of($size) == "number" or (type-of($size) == "string" and str-slice($size, 1, 1) != "_") {
```

CHANGELOG.md
owenatgov: I changed the PR title anyway 😎

packages/govuk-frontend/src/govuk/overrides/_typography.scss
colinrotherham: I've found a way to avoid the duplicate `_14` entry in `$govuk-typography-scale`

I've changed all the keys to strings but updating `$size` to `#{$size}` in our utilities allows that

Probably considered a breaking change but was a nice exercise to try:
https://github.com/alphagov/govuk-frontend/compare/deprecate-14...deprecate-14-strings

CHANGELOG.md
claireashworth: Ah ok!

packages/govuk-frontend/src/govuk/helpers/_typography.scss
```diff
@@ -162,6 +162,15 @@
 
   $font-map: map-get($govuk-typography-scale, $size);
 
+  // Check for a deprecation within the typography scale
```
claireashworth, line 165:
```suggestion
  // Check for a deprecation within the type scale
```

packages/govuk-frontend/src/govuk/helpers/_typography.scss
```diff
@@ -162,6 +162,15 @@
...
+  $deprecation: map-get($font-map, "deprecation");
+
+  @if $deprecation {
+    @include _warning(map-get($deprecation, "key"), map-get($deprecation, "message"));
+    // remove the deprecation map keys so that they don't break the breakpoint loop
```
claireashworth, line 170:
```suggestion
    // remove the deprecation map keys so they do not break the breakpoint loop
```

claireashworth:
Just some minor tweak suggestions from me.

CHANGELOG.md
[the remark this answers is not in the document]
owenatgov: I've made a light amend to refer to the 14 point as "point 14" just because it feels more natural to me. I'm open to disagreements.

packages/govuk-frontend/src/govuk/overrides/_typography.scss
colinrotherham: Sorry @owenatgov I must have [changed my suggestion above](https://github.com/alphagov/govuk-frontend/pull/4649#discussion_r1474385602) too late

I remembered we don't need `type-of()` if we use `#{$size}` to let Sass do the interpolation

owenatgov: Captured the update! I'm not sure if that string fix would be a breaking change or not. It's really cool but would it be ok if we tackled it separately to this?

colinrotherham: Of course. I've split it out into another PR:

* https://github.com/alphagov/govuk-frontend/pull/4713

PR #4650, opened 2024-01-16
Only use fixture HTML 'cache' when running on CI

36degrees:
Using `options?.fixture?.html` provides a neat speed optimisation when running the tests on CI (where we already rendered all of the fixtures to HTML and they will not change).

However, for local development, we currently only update the fixtures based on watching the YAML files. This means they are not updated when the Nunjucks template changes, so tests can run on outdated versions of templates, which is confusing and potentially misleading.

We discussed additionally watching for changes to the Nunjucks templates as well, however there is still a race condition when running tests (especially as the fixture generation currently takes ~1.6-1.8s).

For now, only use this optimisation on CI where we can guarantee we’re testing against an up to date build.

Also update the review app to try and keep it in sync and avoid having multiple different ways of switching on and off the various optimisations.

### DIFF
--- a/packages/govuk-frontend-review/src/app.mjs
+++ b/packages/govuk-frontend-review/src/app.mjs
@@ -48,8 +48,7 @@ export default async () => {
 
   // Feature flags
   const flags = /** @type {FeatureFlags} */ ({
-    isDeployedToHeroku: !!process.env.HEROKU_APP,
-    isDevelopment: !['test', 'production'].includes(process.env.NODE_ENV),
+    isDevelopment: !process.env.HEROKU_APP && !process.env.CI,
 
     // Check for JSDoc, SassDoc and Rollup stats
     hasDocsScripts: await hasPath(join(paths.app, 'dist/docs/jsdoc')),
@@ -221,7 +220,7 @@ export default async () => {
         env,
 
         // Skip Nunjucks render from cache in development
-        fixture: !flags.isDevelopment ? fixture : undefined
+        fixture: flags.isDevelopment ? undefined : fixture
       })
 
       let bodyClasses = 'app-template__body'
@@ -278,8 +277,7 @@ export default async () => {
 
 /**
  * @typedef {object} FeatureFlags
- * @property {boolean} isDeployedToHeroku - Review app using `HEROKU_APP`
- * @property {boolean} isDevelopment - Review app not using `NODE_ENV` production or test
+ * @property {boolean} isDevelopment - Review app not in CI or on Heroku
  * @property {boolean} hasDocsStyles - Stylesheets documentation (SassDoc) is available
  * @property {boolean} hasDocsScripts - JavaScripts documentation (JSDoc) is available
  * @property {boolean} hasStats - Rollup stats are available
--- a/packages/govuk-frontend-review/src/app.mjs
+++ b/packages/govuk-frontend-review/src/app.mjs
@@ -218,9 +218,7 @@ export default async () => {
       const componentView = render(componentName, {
         context: fixture.options,
         env,
-
-        // Skip Nunjucks render from cache in development
-        fixture: flags.isDevelopment ? undefined : fixture
+        fixture
       })
 
       let bodyClasses = 'app-template__body'
--- a/packages/govuk-frontend-review/src/common/nunjucks/index.mjs
+++ b/packages/govuk-frontend-review/src/common/nunjucks/index.mjs
@@ -22,7 +22,7 @@ export function renderer(app) {
     {
       dev: true, // log stack traces
       express: app, // the Express.js review app that nunjucks should install to
-      noCache: !flags.isDeployedToHeroku, // use cache when deployed only
+      noCache: flags.isDevelopment, // use cache in development only
       watch: flags.isDevelopment // reload templates in development only
     },
     {
--- a/shared/lib/components.js
+++ b/shared/lib/components.js
@@ -155,8 +155,10 @@ function render(componentName, options) {
   const macroName = componentNameToMacroName(componentName)
   const macroPath = `govuk/components/${componentName}/macro.njk`
 
-  // Use built fixtures (if they exist) on CI to optimise for speed
-  if (process.env.CI === 'true' && options?.fixture?.html) {
+  // On Heroku / CI we know we're running against an up-to-date build so we can
+  // use the generated HTML from the component JSON (where it exists) to make
+  // things faster
+  if ((process.env.HEROKU_APP || process.env.CI) && options?.fixture?.html) {
     return options.fixture.html
   }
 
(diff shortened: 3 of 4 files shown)


packages/govuk-frontend-review/src/app.mjs
```diff
@@ -48,8 +48,7 @@ export default async () => {
 
   // Feature flags
   const flags = /** @type {FeatureFlags} */ ({
-    isDeployedToHeroku: !!process.env.HEROKU_APP,
-    isDevelopment: !['test', 'production'].includes(process.env.NODE_ENV),
+    isDevelopment: !process.env.HEROKU_APP && !process.env.CI,
```
colinrotherham, line 51: Shall we make this so either can be set independently?

packages/govuk-frontend-review/src/common/nunjucks/index.mjs
```diff
@@ -22,8 +22,8 @@ export function renderer(app) {
     {
       dev: true, // log stack traces
       express: app, // the Express.js review app that nunjucks should install to
-      noCache: !flags.isDeployedToHeroku, // use cache when deployed only
-      watch: flags.isDevelopment // reload templates in development only
+      noCache: flags.isDevelopment, // use cache when deployed only
+      watch: flags.isDevelopment // reload templates unless deployed
```
colinrotherham, line 26: Shall we revert this line and keep it how it was?

```suggestion
      watch: flags.isDevelopment // reload templates in development only
```

packages/govuk-frontend-review/src/common/nunjucks/index.mjs
```diff
@@ -22,8 +22,8 @@ export function renderer(app) {
     {
       dev: true, // log stack traces
       express: app, // the Express.js review app that nunjucks should install to
-      noCache: !flags.isDeployedToHeroku, // use cache when deployed only
-      watch: flags.isDevelopment // reload templates in development only
+      noCache: flags.isDevelopment, // use cache when deployed only
```
colinrotherham, line 25: Shall we say "in development only" like we do for `watch`?

```suggestion
      noCache: flags.isDevelopment, // use cache in development only
```

colinrotherham (approved):
Approved (it works 🙌) but added some optional comments

shared/lib/components.js
```diff
@@ -155,8 +155,12 @@ function render(componentName, options) {
   const macroName = componentNameToMacroName(componentName)
   const macroPath = `govuk/components/${componentName}/macro.njk`
 
-  // Return built fixture or render
-  return options?.fixture?.html ?? renderMacro(macroName, macroPath, options)
+  // Use built fixtures (if they exist) on CI to optimise for speed
+  if (process.env.CI === 'true' && options?.fixture?.html) {
```
colinrotherham, line 159: Sorry to realise this a bit late, but we'll want the HEROKU check here too

We optionally pass through the fixture from **app.mjs** with our new check in:

https://github.com/alphagov/govuk-frontend/blob/4f1d252f2f8dbc305fab6910d70f4ee120a0b7b2/packages/govuk-frontend-review/src/app.mjs#L223

But then chuck it away again here on line 159 when running in Heroku

colinrotherham:
Jest run times from my once-fast Intel Mac after `npm t` completes

It's nice to see we've only losing 3+ seconds, likely now that we (mostly) [bypass the Review app](https://github.com/alphagov/govuk-frontend/pull/4345)

## With cache

Run 1:    62.255 s
Run 2:    62.607 s

## Without cache

Run 1:    66.571 s
Run 2:    65.255 s

---

Just added one extra comment above https://github.com/alphagov/govuk-frontend/pull/4650#discussion_r1454874376

I've realised we'll need to check both `CI` and `HEROKU_APP` in **lib/components.js**

packages/govuk-frontend-review/src/app.mjs
36degrees: I don't quite follow – do you mean have separate flags for Heroku and for CI?

colinrotherham: Sorry, no, but that really wasn't very well explained from me

For example, to start the Review app "from cache" locally you can do:

```console
HEROKU_APP=review npm start --workspace @govuk-frontend/review
```

But previously you could use `NODE_ENV=development` to flip to `isDevelopment: true`

```console
NODE_ENV=development HEROKU_APP=review npm start --workspace @govuk-frontend/review
```

Think it matters?

```suggestion
    isDevelopment:
      process.env.NODE_ENV === 'development' ||
      (process.env.CI !== 'true' && !process.env.HEROKU_APP),
```

36degrees: At the minute we'll always default `isDevelopment` to true unless either `HEROKU_APP` or `CI` are set.

What are the scenarios where those environment variables might be set but we need to override them? I can't think of anything obvious off the top of my head…

colinrotherham: Only what we briefly talked about when pairing

E.g. To run a Heroku preview in "development mode" for extra debugging or log output etc

You've answered the "Shall we…?" and it's not a blocker so can leave it

shared/lib/components.js
colinrotherham: E.g. To prevent re-rendering on Heroku page reloads we'd need to add `HEROKU_APP`

```suggestion
  if (options?.fixture?.html && (process.env.CI || process.env.HEROKU_APP)) {
```

colinrotherham:
@36degrees What do you want to do about https://github.com/alphagov/govuk-frontend/pull/4650#discussion_r1454874376?

Because Heroku deploys lack `CI=true` each preview page refresh re-renders the component

36degrees:
> @36degrees What do you want to do about [#4650 (comment)](https://github.com/alphagov/govuk-frontend/pull/4650#discussion_r1454874376)?
> 
> Because Heroku deploys lack `CI=true` each preview page refresh re-renders the component

I'm just about to push another commit which I think fixes this (if I've understood it correctly)

shared/lib/components.js
36degrees: I believe I've addressed this in https://github.com/alphagov/govuk-frontend/commit/e3651dc8ca9d2163a4eca49df8e3323373390270 – happy with that?

colinrotherham: Amazing, thank you

colinrotherham (approved):
Looking great, let's get this merged 🚀

Unrelated to this change, but turns out [`default` examples skip the cache](https://github.com/alphagov/govuk-frontend/blob/b2cf8f5f163d8b3a6e8d7f3c2ed9172052106cbe/packages/govuk-frontend-review/src/app.mjs#L123-L146) anyway since they don't have an `:exampleName` in the URL 🤦‍♂️

One for another PR

PR #4667, opened 2024-01-23
Fix GitHub Actions updates in nested directories

colinrotherham:
Looks like our GitHub Actions composite workflow YAML files aren't being updated:

This PR adds extra Dependabot config entries for each `directory` following other suggestions in: 

* https://github.com/dependabot/dependabot-core/issues/6704

## Missed update to `actions/setup-node`

We're currently seeing warnings that `actions/setup-node@3.8.2` still uses Node.js 16

Warning shows GitHub Actions using Node js 16 are deprecated

<br>
<br>

This action (line 14 below) should have been updated in https://github.com/alphagov/govuk-frontend/pull/4585 but each directory needs a Dependabot config

https://github.com/alphagov/govuk-frontend/blob/9da6e947afdc5844bd0caaecbe9f61cde76ee9f4/.github/workflows/actions/setup-node/action.yml#L13-L15

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -95,3 +95,39 @@ updates:
       interval: weekly
       time: '10:30'
       timezone: 'Europe/London'
+
+  # Update GitHub Actions (Build)
+  - package-ecosystem: github-actions
+    directory: /.github/workflows/actions/build
+    reviewers:
+      - alphagov/design-system-developers
+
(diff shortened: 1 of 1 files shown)


romaricpascal (approved):
Nice catch! Hopefully we'll remember when we create future actions 🤞🏻 

⛵

Issue #3211, opened 2023-01-26
Review use `indent` in our templates


owenatgov:
## What
We use [`indent`](https://mozilla.github.io/nunjucks/templating.html#indent) across our component templates. Do we still need to do this?

## Why
The primary intent of using `indent` in our context is to make it look tidy when a developer is viewing the source for a page with our components on. There's a question now about how valuable this is based on the reckon that most devs will use dev tools over view source and that we pretty print our HTML.

We should investigate if this is still valuable. If so, we should see if there are things we can do to improve and/or document our use of it. If not, we should remove it.

### Places we use indent
In many of the below examples there are more than one use of `indent` so look at the whole file as well as the referenced line.

- https://github.com/alphagov/govuk-frontend/blob/853403db206aa2fad1daef33e517bdc1c4a61d9a/src/govuk/components/button/template.njk#L48
-  https://github.com/alphagov/govuk-frontend/blob/853403db206aa2fad1daef33e517bdc1c4a61d9a/src/govuk/components/checkboxes/template.njk#L85
- https://github.com/alphagov/govuk-frontend/blob/853403db206aa2fad1daef33e517bdc1c4a61d9a/src/govuk/components/cookie-banner/template.njk#L50
- https://github.com/alphagov/govuk-frontend/blob/853403db206aa2fad1daef33e517bdc1c4a61d9a/src/govuk/components/date-input/template.njk#L40
- https://github.com/alphagov/govuk-frontend/blob/853403db206aa2fad1daef33e517bdc1c4a61d9a/src/govuk/components/file-upload/template.njk#L16
- https://github.com/alphagov/govuk-frontend/blob/853403db206aa2fad1daef33e517bdc1c4a61d9a/src/govuk/components/input/template.njk#L16
- https://github.com/alphagov/govuk-frontend/blob/853403db206aa2fad1daef33e517bdc1c4a61d9a/src/govuk/components/label/template.njk#L11
- https://github.com/alphagov/govuk-frontend/blob/853403db206aa2fad1daef33e517bdc1c4a61d9a/src/govuk/components/phase-banner/template.njk#L10
- https://github.com/alphagov/govuk-frontend/blob/853403db206aa2fad1daef33e517bdc1c4a61d9a/src/govuk/components/radios/template.njk#L25
- https://github.com/alphagov/govuk-frontend/blob/853403db206aa2fad1daef33e517bdc1c4a61d9a/src/govuk/components/select/template.njk#L16 

[...]

PR #4670, opened 2024-01-25
Consistent Nunjucks `safe` and `trim` filter order

colinrotherham:
Split out from https://github.com/alphagov/govuk-frontend/pull/4448 to partially resolve #3211

### Prefer filter `trim` before `indent()`

This change is in preparation for later Nunjucks indentation tweaks to prevent `trim` from removing the first line’s indentation. For example, when using `indent(2, true)` first we must not use `trim` second

### Prefer filter `safe` before `trim`

Ensure HTML output is trimmed first to prevent escaped whitespace characters remaining un-trimmed


36degrees (approved):
Makes sense to me 👍🏻

PR #4671, opened 2024-01-25
Fix Nunjucks HTML indentation: Accordion

colinrotherham:
Accordion HTML indentation changes, split out from https://github.com/alphagov/govuk-frontend/pull/4448 to partially resolve #3211

Similar to the `_actionLink()` and `_summaryCard()` macros in **Summary list**, I've added `_accordionItem()` to flatten the HTML indent level back to 2 spaces

For components with further nesting this approach lets us add `| indent(n)` to the macro output when needed


owenatgov (approved):
Sorted. That diff comment is really nice for stuff like this innit.

PR #4676, opened 2024-01-26
Fix Nunjucks HTML indentation: Radios and Checkboxes

colinrotherham:
Radios and Checkboxes changes, split out from https://github.com/alphagov/govuk-frontend/pull/4448 to partially resolve #3211

Similar to the `_actionLink()` and `_summaryCard()` macros in **Summary list**, I've added `_radioItem()` and `_checkboxItem()` macros to flatten the HTML indent level back to 2 spaces

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/checkboxes/template.njk
+++ b/packages/govuk-frontend/src/govuk/components/checkboxes/template.njk
@@ -26,7 +26,7 @@
   {% set conditionalId = "conditional-" + itemId %}
   {%- if item.divider %}
     <div class="govuk-checkboxes__divider">{{ item.divider }}</div>
-  {%- else %}
+  {% else %}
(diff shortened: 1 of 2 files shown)


romaricpascal (approved):
Nice! ⛵

RichardBradley:
Adding indents to the app-supplied html in radio conditional items breaks our app, because multi-line values inside our `<textarea>`s inside conditional expandos now have extra whitespace added to them, corrupting user input.

The line in question is here, but I suspect this may cause similar issues elsewhere:
https://github.com/alphagov/govuk-frontend/commit/6955db326a2e560c9862d29b28e39bd8f4020829#r139105666

What was the need that this change solved? 
Can this part (or more) be reverted? It seems risky to be changing app supplied HTML in this way

Issue #4629, opened 2024-01-12
Apply typography feature flag to the type scale branch


owenatgov:
## What/Why
Add a new sass variable, `$govuk-new-typography-scale`, and apply it as a feature flag to the new type scale and all changes made as part of the testing of the new type scale. This will stop necessary changes appearing for users who haven't opted into the new type scale and govuk-frontend appearing potentially broken.

## Who needs to work on this
Developers

## Who needs to review this
Developers

PR #2421, opened 2021-11-04
Adjust the responsive type scale

christopherthomasdesign:
## What/Why
Make the following adjustments to the responsive type scale:

- point 16 is now 16 across all screen sizes
- point 19 is now 19 across all screen sizes
- point 24 on small screens is now font size 21px instead of 18px and line height 25px instead of 20px
- point 27 on small screens is now font size 21px instead of 18px and line height 25px instead of 20px
- point 36 on small screens is now font size 27px instead of 24px and line height 30px instead of 25px

Additionally applies the following changes to components in response to the above type scale changes:

- #4064
- #4152 
- #4155 
- #4630 

All of the above is behind the feature flag `$govuk-new-typography-scale` (#4701).

See the epic https://github.com/alphagov/govuk-design-system/issues/2289 and the original investigation https://github.com/alphagov/govuk-frontend/issues/2418 for further details about this change.

PR #4701, opened 2024-01-30
Add new typography scale feature flag

owenatgov:
- adds the feature flag `$govuk-new-typography-scale`
- splits out the new and old type scale maps
- applies the new feature flag across the codebase inline with changes in `type-scale`
- sets the new flag to on in the review app

Closes https://github.com/alphagov/govuk-frontend/issues/4629

### DIFF
--- a/packages/govuk-frontend-review/src/stylesheets/app.scss
+++ b/packages/govuk-frontend-review/src/stylesheets/app.scss
@@ -1,4 +1,5 @@
 $govuk-show-breakpoints: true;
+$govuk-new-typography-scale: true;
 
 @import "govuk/all";
 @import "partials/app";
--- a/packages/govuk-frontend/src/govuk/components/table/_index.scss
+++ b/packages/govuk-frontend/src/govuk/components/table/_index.scss
@@ -9,13 +9,15 @@
     border-collapse: collapse;
   }
 
-  // Modifier for tables with a lot of data. Tables with lots of data benefit
-  // from a smaller font size on small screens.
-  .govuk-table--small-text-until-tablet {
-    @include govuk-font-size($size: 16);
+  @if $govuk-new-typography-scale {
+    // Modifier for tables with a lot of data. Tables with lots of data benefit
+    // from a smaller font size on small screens.
+    .govuk-table--small-text-until-tablet {
+      @include govuk-font-size($size: 16);
 
-    @include govuk-media-query($from: tablet) {
-      @include govuk-font-size($size: 19);
+      @include govuk-media-query($from: tablet) {
+        @include govuk-font-size($size: 19);
+      }
     }
   }
 
--- a/packages/govuk-frontend/src/govuk/settings/_typography-responsive.scss
+++ b/packages/govuk-frontend/src/govuk/settings/_typography-responsive.scss
@@ -16,20 +16,20 @@
 
 $govuk-root-font-size: 16px !default;
 
-/// Responsive typography font map
+/// Feature flag for new typography scale
 ///
-/// This is used to generate responsive typography that adapts according to the
-/// breakpoints.
+/// When set to true, $govuk-typography-scale will use the new font map instead
+/// of the current/old one as well as apply changes in components to account
+/// for the updated scale.
 ///
-/// Font size and font weight can be defined for each breakpoint. You can define
-/// different behaviour on tablet and desktop. The 'null' breakpoint is for
-/// mobile.
+/// Should be removed in 6.0.
 ///
-/// Line-heights will automatically be converted from pixel measurements into
-/// relative values. For example, with a font-size of 16px and a line-height of
-/// 24px, the line-height will be converted to 1.5 before output.
-///
-/// You can also specify a separate font size and line height for print media.
+/// @type Boolean
+/// @access public
+
+$govuk-new-typography-scale: false !default;
+
+/// The font map for the old typography scale
 ///
 /// @type Map
 ///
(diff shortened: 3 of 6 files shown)


packages/govuk-frontend/src/govuk/settings/_typography-responsive.scss
```diff
@@ -151,4 +277,30 @@ $govuk-typography-scale: (
...
+/// @type Map
+///
+/// @prop {Number} $point.$breakpoint.font-size - Font size for `$point` at `$breakpoint`
+/// @prop {Number} $point.$breakpoint.line-height - Line height for `$point` at `$breakpoint`
+/// @prop {Number} $point.print.font-size - Font size for `$point` when printing
+/// @prop {Number} $point.print.line-height - Line height for `$point` when printing
+///
+/// @access public
+
+$govuk-typography-scale: if($govuk-new-typography-scale, $_new-typography-scale, $_old-typography-scale) !default;
```
colinrotherham, line 306: This PR looks pretty ace

Wondered if we should tweak these variable names to follow the colour palette switch?

Bit like with CSS, adding the `-legacy` or `-modern` modifiers at the end, but `$_govuk-` at the start

```scss
$govuk-colours: if(
  $govuk-use-legacy-palette,
  $_govuk-colour-palette-legacy,
  $_govuk-colour-palette-modern
) !default;
```

You can see when we removed it in https://github.com/alphagov/govuk-frontend/commit/2a0a53908682d2b04f1e3660f63e766f67bd6515

packages/govuk-frontend-review/src/stylesheets/app.scss
```diff
@@ -1,4 +1,5 @@
 $govuk-show-breakpoints: true;
+$govuk-new-typography-scale: true;
```
colinrotherham, line 2: If you want a nice CHANGELOG entry, the one for `$govuk-new-link-styles` was pretty good:
https://github.com/alphagov/govuk-frontend/releases/tag/v3.12.0

Maybe flag the Review app to `true` in a separate commit?

36degrees (commented):
Something doesn't seem quite right here – the flag defaults to `false`, so why are we seeing so many changes to font sizes and line-heights in the built CSS in the package? 🤔

36degrees:
I should have thought for an extra thirty seconds before posting that – this PR is targeting the `type-scale` branch so the changes we're seeing are effectively _reverting_ the built package back to what's on `main`? Is that correct?

I think it might be worth doing a manual diff against what's in `main` just to check everything is as we expect.

colinrotherham:
@36degrees Yeah pretty much!

The diff from `main` → `type-scale` in https://github.com/alphagov/govuk-frontend/pull/2421#issuecomment-1889266533 over in PR #2421
The diff from `type-scale` → `new-typography-scale-feature-flag` in https://github.com/alphagov/govuk-frontend/pull/4701#issuecomment-1916703149 above

It's this command that uses `$GITHUB_BASE_REF` for the target:

https://github.com/alphagov/govuk-frontend/blob/2337e95da551b19fc3218f7de9c561a4f207025a/.github/workflows/diff-change-to-package.yaml#L37

owenatgov:
@36degrees @colinrotherham So, if I diff this branch against main it's looking good. CSS and JS diffs are empty. For HTML this is all we've got:

```diff
diff --git a/packages/govuk-frontend/dist/govuk/components/table/template-with-small-text-modifier-for-tables-with-a-lot-of-data.html b/packages/govuk-frontend/dist/govuk/components/table/template-with-small-text-modifier-for-tables-with-a-lot-of-data.html
new file mode 100644
index 000000000..377abdff6
--- /dev/null
+++ b/packages/govuk-frontend/dist/govuk/components/table/template-with-small-text-modifier-for-tables-with-a-lot-of-data.html
@@ -0,0 +1,19 @@
+<table class="govuk-table govuk-table--small-text-until-tablet">
+  <tbody class="govuk-table__body">
+        <tr class="govuk-table__row">
+          <td class="govuk-table__cell">January</td>
+          <td class="govuk-table__cell govuk-table__cell--numeric">£85</td>
+          <td class="govuk-table__cell govuk-table__cell--numeric">£95</td>
+        </tr>
+        <tr class="govuk-table__row">
+          <td class="govuk-table__cell">February</td>
+          <td class="govuk-table__cell govuk-table__cell--numeric">£75</td>
+          <td class="govuk-table__cell govuk-table__cell--numeric">£55</td>
+        </tr>
+        <tr class="govuk-table__row">
+          <td class="govuk-table__cell">March</td>
+          <td class="govuk-table__cell govuk-table__cell--numeric">£165</td>
+          <td class="govuk-table__cell govuk-table__cell--numeric">£125</td>
+        </tr>
+  </tbody>
+</table>

```

That's there because of the new example I added to demo the table class modifier.

The "other" diff is identical to the comment above. I feel confident about this.

packages/govuk-frontend-review/src/stylesheets/app.scss
owenatgov: Whoops I meant to do this but forgot!

packages/govuk-frontend/src/govuk/components/table/_index.scss
```diff
@@ -9,13 +9,15 @@
     border-collapse: collapse;
   }
 
-  // Modifier for tables with a lot of data. Tables with lots of data benefit
-  // from a smaller font size on small screens.
-  .govuk-table--small-text-until-tablet {
-    @include govuk-font-size($size: 16);
+  @if $govuk-new-typography-scale {
```
colinrotherham, line 12: The guidance PR introduces the `--small-text-until-tablet` modifier

But we haven't explained that it's opt-in only via the `$govuk-new-typography-scale` flag, should we?

owenatgov: I'm aware that we don't clarify this and we should. I've got a plan to hold off on looking into this because we're planning on looking at how to do this more broadly via [the typography section restructure work](https://github.com/alphagov/govuk-design-system/pull/3343). Whatever we decide there I'll just copy for that guidance PR.

colinrotherham (approved):

PR #1281, opened 2019-04-23
Make it easier to add a hint or an error message to a fieldset

colinrotherham:
This feature adds a new macro `govukFormGroup()`.

I've addressed what looks like a bit of technical debt.

Duplicated inside almost all form control macros is Nunjucks code to generate a `govukLabel()` label, a `govukHint()` hint an a `govukErrorMessage()` error message:

```nunjucks
  {{ govukLabel({
    html: params.label.html,
    text: params.label.text,
    classes: params.label.classes,
    isPageHeading: params.label.isPageHeading,
    attributes: params.label.attributes,
    for: params.id
  }) | indent(2) | trim }}
{% if params.hint %}
  {% set hintId = params.id + '-hint' %}
  {% set describedBy = describedBy + ' ' + hintId if describedBy else hintId %}
  {{ govukHint({
    id: hintId,
    classes: params.hint.classes,
    attributes: params.hint.attributes,
    html: params.hint.html,
    text: params.hint.text
  }) | indent(2) | trim }}
{% endif %}
{% if params.errorMessage %}
  {% set errorId = params.id + '-error' %}
  {% set describedBy = describedBy + ' ' + errorId if describedBy else errorId %}
  {{ govukErrorMessage({
    id: errorId,
    classes: params.errorMessage.classes,
    attributes: params.errorMessage.attributes,
    html: params.errorMessage.html,
    text: params.errorMessage.text,
    visuallyHiddenText: params.errorMessage.visuallyHiddenText
  }) | indent(2) | trim }}
{% endif %}
```

It's used by textarea, select, radios, checkboxes, text input, file upload and date input.

All this boilerplate code is now handled by `govukFormGroup()`.

This allows us to port existing services with more complex multi-field pages (often with some fields nested inside other fieldsets) like this:

Example

With little markup like:

```js
{% call govukFormGroup({
  fieldset: {
    legend: {
      text: "Pick one of the following options:",
      classes: "govuk-fieldset__legend--m"
    }
  }
}) %}

{{ govukRadios({
  "fieldset": {
    "legend": {
      text: "Group 1",
      classes: "govuk-fieldset__legend--s"
    }
  },
  "idPrefix": "radio-stacked-sub-headings-group-1",
  "name": "radio-stacked-sub-headings",
  "items": [
    {
      "value": "a",
      "text": "Option A"
    },
    {
      "value": "b",
      "text": "Option B"
    },
    {
      "value": "c",
      "text": "Option C"
    }
  ]
}) }}

[...]

PR #4702, opened 2024-01-30
[SPIKE] Investigate the utility of an input macro separate from `govukInput` 

querkmachine:
## Background

Currently, the Date input component directly imports and uses the `govukInput` macro to provide the date fields; and the Character count component directly imports and uses the `govukTextarea` macro for the textarea. 

In both cases, importing the macro carries with it the entire HTML of that component, including the `govuk-form-group` wrapper, label, hint, error message, and the input/textarea itself. This HTML is non-permeable to the parent component and cannot be altered, unless we specifically extend those macros with new parameters to do so.

### The Character count component

In the case of the Character count, this limitation has led to a few 'hacky' workarounds:

- The inability to alter what's inside `govukTextarea` means that the 'info' hint below the textarea has to live outside of `govuk-form-group`, which has a large bottom margin that pushes the 'info' hint away from the textarea. 
- Custom CSS is then used to remove the bottom margin from `govuk-form-group` and instead apply an identical margin to `govuk-character-count`, a new element that wraps around the entire form group.
- JavaScript is used to move the 'info' hint up to within the `govuk-form-group` where it's available to do so. 

None of these are ideal, and all of them constitutes extra code loaded by browsers that doesn't need to exist.

### The Password input component

In developing the Password input component, we also saw us having to deal with similar frustrations when it came to applying a wrapping element around only the input of `govukInput` and placing a button within that new wrapper. 

There was also some trepidation about how `govukInput` could conflict with or compromise the effectiveness of the Password input; such as how input's prefix/suffix HTML may interfere with the Password input's own wrapper and button, or that some of the accessibility and security features of the password input could be overridden through modification of the inherited input.

These issues led us to initially decide to not use the macro — #4512.

## Exploring solutions

Reckoning that this may be a recurring problem in future, we are considering a few ways to modify how these components are composed to avoid these problems.

One option being explored is to add 'injection' points within the components where arbitrary HTML can be added, similar to how we use Nunjucks `block`s in templating — #4567.

### The idea behind this spike

This spike seeks to minimise the use of directly importing and extending `govukInput` and `govukTextarea` in favour of two 'internal' macros. 

These macros only include the base `input` and `textarea` elements, and do not include the other parts (group wrapper, label, etc.) with the expectation that:

- Components become more 'standalone', implementing their own label/hint/error code rather than reusing them from the Text input and Textarea components.
- In doing so, we can 'liberalise' the component HTML and allow it to more readily diverge where it needs to, such as for Character count and Password input.
- These macros define the system-wide defaults and are extended with whatever parameter options become necessary, instead of having those options unnecessarily bloat the `govukInput` and `govukTextarea` components. 
- Components can choose to only expose the relevant parameters of the underlying macros. 

## Changes

- Creates `govukInputElement` and `govukTextareaElement` Nunjucks macros, each containing the base `input` and `textarea` elements. 
- Additionally creates a `_govukTextCommonAttributes` macro for attributes that are common to both `input` and `textarea`.
- Reworks Character Count, Date input, (Text) Input, Password input and Textarea to use the new macros.
- Additionally reworks Character count to:
  - move the wrapping element to within the form group
  - move the 'info' hint to within the form group
  - remove the CSS styles to adjust the bottom margins
  - remove the JavaScript to programatically move the 'info' hint 

## Thoughts

- I am not at all tied to the current names for the internal macros. They can probably be a lot clearer.
- Not much care put to how the HTML is output either. This can definitely be improved too.
- No tests as it's a spike.

### DIFF
--- a/packages/govuk-frontend/src/govuk-prototype-kit/govuk-prototype-kit.config.unit.test.mjs
+++ b/packages/govuk-frontend/src/govuk-prototype-kit/govuk-prototype-kit.config.unit.test.mjs
@@ -137,6 +137,10 @@ describe('GOV.UK Prototype Kit config', () => {
           macroName: 'govukPanel'
         },
         {
+          importFrom: 'govuk/components/password-input/macro.njk',
+          macroName: 'govukPasswordInput'
+        },
+        {
           importFrom: 'govuk/components/phase-banner/macro.njk',
           macroName: 'govukPhaseBanner'
         },
--- a/packages/govuk-frontend/src/govuk/all.jsdom.test.mjs
+++ b/packages/govuk-frontend/src/govuk/all.jsdom.test.mjs
@@ -14,6 +14,7 @@ jest.mock(`./components/error-summary/error-summary.mjs`)
 jest.mock(`./components/exit-this-page/exit-this-page.mjs`)
 jest.mock(`./components/header/header.mjs`)
 jest.mock(`./components/notification-banner/notification-banner.mjs`)
+jest.mock(`./components/password-input/password-input.mjs`)
 jest.mock(`./components/radios/radios.mjs`)
 jest.mock(`./components/skip-link/skip-link.mjs`)
 jest.mock(`./components/tabs/tabs.mjs`)
@@ -27,7 +28,8 @@ describe('initAll', () => {
     'character-count',
     'error-summary',
     'exit-this-page',
-    'notification-banner'
+    'notification-banner',
+    'password-input'
   ]
 
   afterEach(() => {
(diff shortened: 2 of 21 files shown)


colinrotherham:
@querkmachine I'll give this a proper look tomorrow 👀

Whilst I remember, this previous work in https://github.com/alphagov/govuk-frontend/pull/1281 could be an alternative

It abstracted away the form group wrapper, label, hint, error message into `govukFormGroup()`

That PR could be made to make `govukFormGroup()` optional, letting every form field component render "wrapper-less". Similar to how Checkboxes have optional fieldsets

colinrotherham:
Ah I've still got a separate input macro for the Nunjucks indenting work

I've moved the Text input changes into a separate PR where an `_inputElement()` macro is used:

* https://github.com/alphagov/govuk-frontend/pull/4707

PR #4707, opened 2024-01-31
Fix Nunjucks HTML indentation: Text input

colinrotherham:
Text input changes, split out from https://github.com/alphagov/govuk-frontend/pull/4448 to partially resolve #3211

Similar to the `_actionLink()` and `_summaryCard()` macros in **Summary list**, I've added `_inputElement()` and `_affixItem()` macros to flatten the HTML indent level back to 2 spaces

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/input/template.njk
+++ b/packages/govuk-frontend/src/govuk/components/input/template.njk
@@ -6,6 +6,9 @@
   aria-describedby – for example hints or error messages -#}
 {% set describedBy = params.describedBy if params.describedBy else "" -%}
 
+{%- set hasPrefix = true if params.prefix and (params.prefix.text or params.prefix.html) else false %}
+{%- set hasSuffix = true if params.suffix and (params.suffix.text or params.suffix.html) else false %}
+
 <div class="govuk-form-group {%- if params.errorMessage %} govuk-form-group--error{% endif %} {%- if params.formGroup.classes %} {{ params.formGroup.classes }}{% endif %}" {%- for attribute, value in params.formGroup.attributes %} {{ attribute }}="{{ value }}"{% endfor %}>
   {{ govukLabel({
     html: params.label.html,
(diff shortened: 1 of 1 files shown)


packages/govuk-frontend/src/govuk/components/input/template.njk
```diff
@@ -39,30 +60,17 @@
...
+      {{- _affixItem(params.prefix, "prefix") | indent(2, true) }}
+    {% endif %}
+    {{ _inputElement(params) }}
+    {% if hasSuffix %}
+      {{- _affixItem(params.suffix, "suffix") | indent(2, true) }}
+    {% endif %}
+  </div>
+{% else %}
+  {{ _inputElement(params) }}
+{% endif %}
```
romaricpascal, line 75: Linking how the encapsulation makes the structure very apparent and easier to follow than the `{%- if}` triggering the rendering of opening/closing tags 😍

packages/govuk-frontend/src/govuk/components/input/template.njk
```diff
@@ -6,6 +6,27 @@
...
+    {%- if params.pattern %} pattern="{{ params.pattern }}"{% endif %}
+    {%- if params.inputmode %} inputmode="{{ params.inputmode }}"{% endif %}
+    {%- for attribute, value in params.attributes %} {{ attribute }}="{{ value }}"{% endfor %}>
+{%- endmacro -%}
+
+{%- macro _affixItem(affix, type) %}
+  <div class="govuk-input__{{ type }} {%- if affix.classes %} {{ affix.classes }}{% endif %}" aria-hidden="true" {%- for attribute, value in affix.attributes %} {{ attribute }}="{{ value }}"{% endfor %}>
+    {{- affix.html | safe | trim | indent(4) if affix.html else affix.text -}}
+  </div>
+{%- endmacro -%}
```
romaricpascal, line 28: Nunjucks is not where I'm the most familiar, are there any side effects to defining macros each time an input is rendered?

colinrotherham: Very little I imagine

It's the approach we started taking with Summary list, and now Button and Accordion:

https://github.com/alphagov/govuk-frontend/blob/8eeca8e8a7980dfaceb6619ca97084fdb63634f6/packages/govuk-frontend/src/govuk/components/summary-list/template.njk#L1

https://github.com/alphagov/govuk-frontend/blob/8eeca8e8a7980dfaceb6619ca97084fdb63634f6/packages/govuk-frontend/src/govuk/components/summary-list/template.njk#L13

https://github.com/alphagov/govuk-frontend/blob/8eeca8e8a7980dfaceb6619ca97084fdb63634f6/packages/govuk-frontend/src/govuk/components/button/template.njk#L23

https://github.com/alphagov/govuk-frontend/blob/8eeca8e8a7980dfaceb6619ca97084fdb63634f6/packages/govuk-frontend/src/govuk/components/accordion/template.njk#L3

packages/govuk-frontend/src/govuk/components/input/template.njk
colinrotherham: Thanks, same! It helped me understand it too

PR #4713, opened 2024-02-01
Ensure `govuk-font-size()` handles string keys

colinrotherham:
Quick spike to split out some code review ideas in https://github.com/alphagov/govuk-frontend/pull/4649#discussion_r1474385602

### Add tests for "T-Shirt Sizing" font scales

This PR adds new tests for services that extend `$govuk-typography-scale` with string sizes:

```scss
@include govuk-font-size($size: "small")
@include govuk-font-size($size: "medium")
@include govuk-font-size($size: "large")
```

### Remove duplicate font declaration for "internal use" point 14

This PR automatically maps `"_14"` to `14` as part of `govuk-font-size()` by removing the underscore:

```scss
@include govuk-font-size($size: "_14")
@include govuk-font-size($size: "14")
@include govuk-font-size($size: 14)
```

We can now remove the duplicated `_14` declaration:

https://github.com/alphagov/govuk-frontend/blob/eb3c18d8e75cb752fce368ad2c3b84ed11037b9d/packages/govuk-frontend/src/govuk/settings/_typography-responsive.scss#L161-L174


packages/govuk-frontend/src/govuk/helpers/_typography.scss
```diff
@@ -156,17 +156,37 @@
 /// @access public
 
 @mixin govuk-font-size($size, $line-height: false, $important: false) {
-  @if not map-has-key($govuk-typography-scale, $size) {
-    @error "Unknown font size `#{$size}` - expected a point from the type scale.";
+  $internal-use: str-slice(#{$size}, 1, 1) == "_";
```
owenatgov, line 159: Could we add a comment just to clarify what this is? You and me both know obviously, but for future folks.

Aside: Should we keep this in as a means to deprecate other parts of the type scale in future or remove it in 6.0?

owenatgov (approved):
I love this, thanks Colin. I've left one comment but approving anyway as it's non-critical.

I think this deserves a changelog entry but it doesn't need a big one. I reckon either under fixes or you can add it to the 14 deprecation changelog entry ie: "this change was done in [original PR] and [this PR]".

packages/govuk-frontend/src/govuk/helpers/_typography.scss
colinrotherham: Yeah course, will do. Thanks @owenatgov

Keep it? We've built ourselves a mechanism to deprecate sizes that other teams could use too

owenatgov: Aha good point. Let's leave it in.

colinrotherham: Any thoughts on a better variable than `$internal-use`?

I thought about `$size-is-deprecated` but that's only when the map _also_ has a `deprecated` property

colinrotherham: I've gone with `$size-internal-use-only` and will push up this comment:

```suggestion
  // Flag font sizes that start with underscores so we can suppress warnings on
  // deprecated sizes used internally, for example `govuk-font($size: "_14")`
  $size-internal-use-only: str-slice(#{$size}, 1, 1) == "_";
```

owenatgov: I think that's fine. This particular variable feels really tricky to name without getting trite and what you've gone with is probably as good as we can get it.

colinrotherham:
@owenatgov Also realised, we no longer need to suppress `.govuk-!-font-size-_14` output

```scss
// Suppress class if the map key is a string with an underscore eg: _14 to
// avoid classes like govuk-!-font-size-_14 getting generated
```

Updated in https://github.com/alphagov/govuk-frontend/pull/4713/commits/220a4c07f753b3d94535ba0543e752a90643b251 where `_14` is removed 🙌

PR #4717, opened 2024-02-02
Fix Nunjucks HTML indentation: Table

colinrotherham:
Table changes, split out from https://github.com/alphagov/govuk-frontend/pull/4448 to partially resolve #3211


romaricpascal (approved):
Diff looks neat! ⛵

PR #4718, opened 2024-02-03
Update `@prettier/sync`

fisker:
Issue was reported that @prettier/sync@0.5.0 doesn't work, https://github.com/fisker/make-synchronized/issues/17

Creating this to try reproduce on CI.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3990,17 +3990,6 @@
         "node": ">=14"
       }
     },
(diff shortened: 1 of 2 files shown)


colinrotherham:
Thanks for looking the issue for us @fisker

Will close this PR in favour of https://github.com/alphagov/govuk-frontend/pull/4737 with the update

PR #4740, opened 2024-02-07
Make Tudor Crown logo the default

querkmachine:
Now that the versions of Frontend featuring the new Tudor crown logo have been released for the February–March transition period, we can start thinking about making the new logo the default in a future releases.

This is so that new projects and prototypes using Frontend will automatically use the new logo, and users will no longer have to manually enable it.

## Changes 

- Reverses the Header component logic so that the Tudor Crown is the default and can be overridden by setting `useTudorCrown: false`
- Updates Nunjucks documentation to indicate the new default state.
- Updates tests to account for reversed logic.
- Marks `useTudorCrown` as deprecated in the changelog and Nunjucks documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,14 @@ govukInput({
 
 This change was introduced in [pull request #4567: Add `beforeInput(s)` and `beforeInput(s)` options to form groups](https://github.com/alphagov/govuk-frontend/pull/4567).
 
+### Deprecated features
+
+#### Stop using the `useTudorCrown` parameter in the Heading component
+
+As the beginning of the GOV.UK logo rollout has passed, the Tudor crown logo is now shown by default and the `useTudorCrown` parameter has been deprecated. We will remove it in the next major release.
+
+The `useTudorCrown` parameter, along with any other adjustments made to display the Tudor crown logo in your service, can now be removed.
+
 ## 5.1.0 (Feature release)
 
 To install this version with npm, run `npm install govuk-frontend@5.1.0`. You can also find more information about [how to stay up to date in our documentation](https://frontend.design-system.service.gov.uk/staying-up-to-date/#updating-to-the-latest-version).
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,14 @@ For advice on how to use these release notes see [our guidance on staying up to
 
 ## Unreleased
 
+### Deprecated features
+
+#### Stop using the `useTudorCrown` parameter in the Heading component
+
+As the GOV.UK logo rollout has ended, the Tudor crown logo is now shown by default and the `useTudorCrown` parameter has been deprecated. We will remove it in the next major release.
```
owenatgov, line 11: As we're planning on releasing this on the 23rd, I suggest we soften this language.

```suggestion
As the beginning of the GOV.UK logo rollout has past, the Tudor crown logo is now shown by default and the `useTudorCrown` parameter has been deprecated. We will remove it in the next major release.
```

packages/govuk-frontend/src/govuk/components/header/template.test.js
```diff
@@ -301,11 +301,11 @@ describe('header', () => {
       expect($svg.html()).toContain('<title>GOV.UK</title>')
     })
 
-    it('uses the Tudor Crown if useTudorCrown is set', () => {
-      $ = render('header', examples['with Tudor crown'])
+    it("uses the St Edward's Crown if useTudorCrown is false", () => {
```
owenatgov, line 304: Extremely pithy comment that we're using double quotes here when the pattern for test cases are single quotes

owenatgov (approved):
A couple of tiny comments that are non-critical. Great stuff.

CHANGELOG.md
querkmachine: Done but for swapping `past` for `passed`, which my spellchecker says is the word that should be there.

Issue #1529, opened 2019-08-12
Improve documentation for attributes option


edwardhorsford:
The `attributes` option on some macros is seemingly very powerful - but I'm not sure the expected format is documented anywhere.

The docs say:
`"HTML attributes (for example data attributes) to add to the hint span tag."`

As a new user, it would have been helpful to see something that showed the data structure needed - I had to guess it would be an object with keys.

edwardhorsford:
I was hoping to use `attributes` to set a class on a checkbox item - but I don't think I can because there's already one. Should attributes be merged in? should it mention that there's some stuff you can't set?

nickcolley:
@edwardhorsford I think at this point it'd be easier to consider adding 'classes' to individual items than merging attributes (as this would require Nunjucks global functions), so if you need that feel free to raise a separate issue.

PR #4743, opened 2024-02-08
Update macro `govukAttributes()` to support optional attributes

colinrotherham:
This PR was split out from https://github.com/alphagov/govuk-frontend/pull/4742 and closes https://github.com/alphagov/govuk-frontend/issues/2917

It extends `govukAttributes()` with a new object config to support:

1. Optional attributes that render only when `value` is set
2. Boolean attributes that render only a name, e.g. `<input type="radio" checked>`

For **Password input**, we can now render a show/hide button `hidden` by default, without failing on [**html-validate** `attribute-boolean-style`](https://html-validate.org/rules/attribute-boolean-style.html) linting errors due to unnecessary empty values

### Before

Renders attribute `'hidden="hidden"'` or `'hidden=""'`

```njk
{{ govukButton({
  attributes: {
    hidden: "hidden" if params.hidden
  })
}}
```

### After

Renders attribute `'hidden="hidden"'` or `''`

```njk
{{ govukButton({
  attributes: {
    hidden: {
      value: params.hidden,
      optional: true
    }
  })
}}
```

## No breaking changes

This PR continues to render `true`, `"true"`, `false` and `"false"` as strings for attributes such as:

* `aria-hidden="false"`
* `focusable="false"`

But adds a new object config with `optional: true` to "opt in" to make attributes and their values optional, where empty values such as `null`, `undefined` or `false` will affect how the attribute is rendered

This gives us the following new behaviours:

* By default or when using `optional: false`, attributes render as ` name="value"`
* When using `optional: true`, attributes with empty (`null`, `undefined` or `false`) values are omitted
* When using `optional: true`, attributes with `true` (boolean) values render `name` only without value

## Boolean attributes

Why do we need this PR?

The previous Nunjucks attribute loop couldn't add true/false [boolean attributes](https://developer.mozilla.org/en-US/docs/Glossary/Boolean/HTML) without them being stringified:

```html
<button hidden="hidden"> ❌
<button hidden="true"> ❌
<button hidden=""> ❌
<button hidden> ✅
```

But by using the example above, we now can 🙌

See https://github.com/alphagov/govuk-frontend/pull/1665 for previous work on this

## Other optional attributes

Not all optional attributes are [boolean attributes](https://developer.mozilla.org/en-US/docs/Glossary/Boolean/HTML)

For example we add `data-nosnippet` to the [**Cookie banner** component](https://design-system.service.gov.uk/components/cookie-banner/) and could be used as shown:

### Nunjucks input

```html
<p class="govuk-body" {{- govukAttributes({
  "data-nosnippet": {
    value: true,
    optional: true
  }
}) }}>
  Exclude me from search engine result snippets
</p>
```

### HTML output

```html
<p class="govuk-body" data-nosnippet>
  Exclude me from search engine result snippets
</p>
```

This approach will close https://github.com/alphagov/govuk-frontend/issues/2917

### DIFF
--- a/packages/govuk-frontend/src/govuk/macros/attributes.njk
+++ b/packages/govuk-frontend/src/govuk/macros/attributes.njk
@@ -1,8 +1,14 @@
 {#
   Renders component attributes as string
 
+  * By default or using `optional: false`, attributes render as ` name="value"`
+  * Using `optional: true`, attributes with empty (`null`, `undefined` or `false`) values are omitted
+  * Using `optional: true`, attributes with `true` (boolean) values render `name` only without value
+
+  {@link https://developer.mozilla.org/en-US/docs/Glossary/Boolean/HTML}
+
   @private
-  @param {{ [attribute: string]: string | { value: string } } | string} attributes - Component attributes param
+  @param {{ [attribute: string]: string | { value: string, optional?: boolean } } | string} attributes - Component attributes param
 #}
 {% macro govukAttributes(attributes) -%}
   {#- Default attributes output -#}
@@ -13,11 +19,17 @@
     {% for name, value in attributes %}
       {#- Set default attribute options -#}
       {% set options = value if value is mapping else {
-        value: value
+        value: value,
+        optional: false
       } %}
 
-      {#- Always render ` name="value"` pair -#}
-      {% set attributesHtml = attributesHtml + " " + name | escape + '="' + options.value | escape + '"' %}
+      {#- Output ` name` only (no value) for boolean attributes -#}
+      {% if options.optional === true and options.value === true %}
+        {% set attributesHtml = attributesHtml + " " + name | escape %}
+      {#- Skip optional empty attributes or output ` name="value"` pair by default -#}
+      {% elif (options.optional === true and not [undefined, null, false].includes(options.value)) or options.optional !== true %}
+        {% set attributesHtml = attributesHtml + " " + name | escape + '="' + options.value | escape + '"' %}
+      {% endif %}
     {% endfor %}
   {% endif -%}
 
--- a/packages/govuk-frontend/src/govuk/macros/attributes.njk
+++ b/packages/govuk-frontend/src/govuk/macros/attributes.njk
@@ -2,7 +2,7 @@
   Renders component attributes as string
 
   @private
-  @param {{ [attribute: string]: string } | string} attributes - Component attributes param
+  @param {{ [attribute: string]: string | { value: string } } | string} attributes - Component attributes param
 #}
 {% macro govukAttributes(attributes) -%}
   {#- Default attributes output -#}
@@ -11,7 +11,13 @@
   {#- Append attribute name/value pairs -#}
   {%- if attributes is mapping %}
     {% for name, value in attributes %}
-      {% set attributesHtml = attributesHtml + " " + name | escape + '="' + value | escape + '"' %}
+      {#- Set default attribute options -#}
+      {% set options = value if value is mapping else {
+        value: value
+      } %}
+
+      {#- Always render ` name="value"` pair -#}
+      {% set attributesHtml = attributesHtml + " " + name | escape + '="' + options.value | escape + '"' %}
     {% endfor %}
   {% endif -%}
 
--- a/packages/govuk-frontend/src/govuk/macros/attributes.njk
+++ b/packages/govuk-frontend/src/govuk/macros/attributes.njk
@@ -7,6 +7,57 @@
 
   {@link https://developer.mozilla.org/en-US/docs/Glossary/Boolean/HTML}
 
+  @example
+  Output attribute ` aria-hidden="true"` when `true` (boolean) or `"true"` (string)
+
+  ```njk
+  govukAttributes({
+    "aria-hidden": true
+  })
+  ```
+
+  @example
+  Output attribute ` aria-hidden="false"` when `false` (boolean) or `"false"` (string)
+
+  ```njk
+  govukAttributes({
+    "aria-hidden": false
+  })
+  ```
+
+  @example
+  Output attribute ` hidden=""` when `null`, `undefined` or empty `""` (string)
+
+  ```njk
+  govukAttributes({
+    "hidden": undefined
+  })
+  ```
+
+  @example
+  Output attribute ` hidden` as boolean attribute when optional and `true`
+
+  ```njk
+  govukAttributes({
+    hidden: {
+      value: true,
+      optional: true
+    },
+  })
+  ```
+
+  @example
+  Output empty string when optional and `null`, `undefined` or `false`
+
+  ```njk
+  govukAttributes({
+    hidden: {
+      value: undefined,
+      optional: true
+    },
+  })
+  ```
+
   @private
   @param {{ [attribute: string]: string | { value: string, optional?: boolean } } | string} attributes - Component attributes param
 #}
--- a/packages/govuk-frontend/src/govuk/macros/attributes.unit.test.mjs
+++ b/packages/govuk-frontend/src/govuk/macros/attributes.unit.test.mjs
@@ -57,6 +57,188 @@ describe('attributes.njk', () => {
       )
     })
 
+    it('renders attribute values as strings when `optional` not set', () => {
+      const attributes = renderMacro(
+        'govukAttributes',
+        'govuk/macros/attributes.njk',
+        {
+          context: {
+            viewBox: '0 0 15 13',
+            focusable: {
+              value: false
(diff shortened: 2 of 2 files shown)


querkmachine:
> **After:** Attributes with `false`, `null` or `undefined` values are removed entirely

This gives me cause for concern, as there are many legitimate cases where one of these (though, particularly `false`) is a valid value to pass into an attribute. 

* Many ARIA attributes work using `true`/`false` values to indicate state. Though we usually inject these attributes via JavaScript than have them in our HTML, other people using the macros may be including them straight in their HTML.
* The `focusable` attribute we include on SVGs is normally set to `false`. 
* Custom `data-*` attributes within a service may be both expected to exist and to have a value of `false`. 

I imagine checking for `undefined`-ness rather than `false`-ness would cause too many issues with valueless attributes appearing. 

Would it be more foolproof for it to check against a list of 'known good' boolean attributes (`hidden`, `checked`, etc.) and only coerce those into outputting no value?

colinrotherham:
@querkmachine Definitely an idea, here's the [original comment](https://github.com/alphagov/govuk-frontend/pull/4742#discussion_r1482570163)

It's the always needing to be "type aware" that makes this a breaking change

Great example by the way, two attributes side-by-side but one would need replacing ~`true`~ → `"true"`

```patch
<svg {{- govukAttribute({
  "data-nosnippet": true
-  "aria-hidden": true
+  "aria-hidden": "true"
})>
```

Or similar to https://github.com/alphagov/govuk-frontend/issues/2917 we could opt-in to boolean attribute behaviour using configs:

```patch
<svg {{- govukAttribute({
- "data-nosnippet": true
+ "data-nosnippet": {
+   type: "boolean",
+   value: true
+ },
  "aria-hidden": true
})>
```

querkmachine:
@colinrotherham Thanks for the context. If this is intending to be a breaking change, is there a workaround for using boolean attributes in the interim? Just not using them within `govukAttributes`? Repeating attribute name for the value, XHTML style?

colinrotherham:
@querkmachine We'd be safer making it opt-in (attribute config shown above) especially with your example

Caught up with @romaricpascal today since attribute configs were his idea originally, one for dev catch up and I'll get this PR updated so it's no longer a breaking change

In the meantime, go ahead with `hidden="hidden"` as it matches our use of `disabled="disabled"` before https://github.com/alphagov/govuk-frontend/commit/269634b78597bc31edb2e888b0c69811de9d128f. You should be able to use this with `govukAttributes()` but see b1bb9e81135cc5ff0d35c118d03d7d180c766e8a on ignoring the HTML validation rule

colinrotherham:
@querkmachine PR updated to avoid breaking changes

querkmachine (approved):
This seems good to me.

It feels like there's a little redundancy in requiring boolean attributes to provide both a type and a value in all circumstances. The value is going to be `true` if the type is `boolean` the vast majority of the time, but for circumstances where we can't be as sure (i.e. the value comes from a Nunjucks parameter) it makes sense to have the option of the value being falsy. Not a big deal, IMO.

colinrotherham:
Thanks @querkmachine, glad the options idea is a good one

Would be good to handle optional non-boolean attributes too, like these:

```njk
{%- if params.maxlength %} data-maxlength="{{ params.maxlength }}"{% endif %}
{%- if params.threshold %} data-threshold="{{ params.threshold }}"{% endif %}
{%- if params.maxwords %} data-maxwords="{{ params.maxwords }}"{% endif %}
```

Will catch up with @romaricpascal and see if we can close https://github.com/alphagov/govuk-frontend/issues/2917 too

36degrees:
This is neat but I think we're going to struggle to effectively document how it works within the 'macro options' for the components that use it.

I'm wondering if we should add some generic guidance on 'passing attributes to components' with some examples to our ['Use Nunjucks' documentation](https://frontend.design-system.service.gov.uk/use-nunjucks) that we could link to from the 'attributes' macro option rows?

May be able to close #1529 at the same time.

colinrotherham:
Thanks @36degrees, yeah I've wondered exactly how public to make these new options

Similarly we added support for string attributes in https://github.com/alphagov/govuk-frontend/pull/4742 but we're not using it yet until https://github.com/alphagov/govuk-frontend/pull/4566 or https://github.com/alphagov/govuk-frontend/pull/4442 are merged

## Redundant options

Doing some more thinking about @querkmachine's thoughts in https://github.com/alphagov/govuk-frontend/pull/4743#pullrequestreview-1878515743 which says:

>It feels like there's a little redundancy in requiring boolean attributes to provide both a type and a value…

How about we trust that `true` (boolean) and `"true"` (string) are separate without a type?

We could rename the option to `optional: true` and handle `null`, `undefined` in addition to `false`?

### Before

```njk
<div class="govuk-character-count" data-module="govuk-character-count"
  {%- if params.maxlength %} data-maxlength="{{ params.maxlength }}"{% endif %}
  {%- if params.threshold %} data-threshold="{{ params.threshold }}"{% endif %}
  {%- if params.maxwords %} data-maxwords="{{ params.maxwords }}"{% endif %}
```

### After

We can prevent empty `maxlength=""`, `maxwords=""` etc by setting `optional: true` for each:

```njk
<div class="govuk-character-count" data-module="govuk-character-count"  
  {{- govukAttributes({
    maxlength: {
      value: params.maxlength,
      optional: true
    },
    maxwords: {
      value: params.maxwords,
      optional: true
    },
    threshold: {
      value: params.threshold,
      optional: true
    }
  }) -}}
```

This would solve https://github.com/alphagov/govuk-frontend/issues/2917

romaricpascal:
For what we're trying to achieve right now, would having only the `value` be enough? If so, would the following be sufficient to unblock us:
- limiting it to only `value` without `type` to limit the area for future breaking changes
- keeping that API private for now, as it seems the initial need stemmed from internal use and that the existing behaviour of our `attributes` option is unchanged

Thinking we could have a later cycle to properly tighten the implementation, with proper time to discuss documentation, potential for breaking changes and deprecation. If a cycle feels too large for this, thinking there could be further thoughts to have around the `attributes` option in terms of how it interacts with the `class` attribute (and others, like `id`...) that the component may already set.

romaricpascal:
> How about we trust that true (boolean) and "true" (string) are separate without a type?

I think that would be very fine for the `value` in that new object to take into account the `type` if it saves us adding an extra option.

colinrotherham:
Thanks @romaricpascal

>> How about we trust that true (boolean) and "true" (string) are separate without a type?
>
>I think that would be very fine for the `value` in that new object to take into account whether it receives a `string` or `boolean` if it saves us adding an extra option.

We'd still need an option like `optional: true` in https://github.com/alphagov/govuk-frontend/pull/4743#issuecomment-1944019911 to distinguish between:

1. Attributes like `focusable="false"` where the false state is still rendered
2. Attributes like `checked` where only the true state is rendered, false is removed

I've pushed up again, what do you think?

packages/govuk-frontend/src/govuk/macros/attributes.njk
```diff
@@ -1,8 +1,8 @@
 {#
-  Renders component attributes as string
+  Renders component attributes as strings unless configured as a boolean attribute
```
querkmachine, line 2: This comment might need updating after the most recent changes.

querkmachine (approved):
As I'm understanding it, if `optional: false` (the default):
- passing in an attribute (boolean or not) with any value will output both the attribute name and value in all circumstances.

If `optional: true`:
- passing an string attribute with a truthy[^1] value will output both the attribute name and value.
- passing an string attribute with a falsy[^1] value will output nothing.
- passing an boolean attribute with a truthy value will output only the attribute name.
- passing an boolean attribute with a falsey value will output nothing.

To me, this makes sense after having it explained a bit. I wonder if it could use some documentation or code comments making each scenario explicit (maybe like above) could help some.

I agree with the sentiment that the use of `govukAttributes` be better kept internal for the time being to allay the risk of breaking other people's macro usage whilst we feel out the edge cases. 

[^1]: A falsy value in this context only includes `false`, `null` and `undefined`. Values of `""` and `0` are considered truthy.

colinrotherham:
>To me, this makes sense after having it explained a bit. I wonder if it could use some documentation or code comments making each scenario explicit (maybe like above) could help some.

@querkmachine Perfect, thank you

Good thinking, I've pushed up five new examples in cc046ec1c74b31a8852a28eb7667613fdf5343a9

romaricpascal (approved):
That'll work nicely 🥳 And if we keep it internal for now, the risks of misuse are limited. Thanks for adding all the examples in the docs as well ⛵ ⛵ ⛵

Issue #3479, opened 2023-04-03
Define a public API for our components in v5


romaricpascal:
## What

Write a proposal defining the public API of our components, especially:

### The properties available on their instances

We'll likely want to start with a very narrow public API (eg. only `init()` is public) and expand it in future versions.

### Errors components may throw during initialisation

These may include instanciating a component:

- where [GOV.UK Frontend is not supported](https://github.com/alphagov/govuk-frontend/issues/3509)
- which has already been initialised
- which has already been initialised as a different component
- by passing something which is not an HTML element
- by passing something which does not have the expected DOM
- by passing a config object which is not what we expect
 
### Methods they offer for their initialisation

We may want to offer custom methods of initialisation that allow to re-use component instances or are a bit softer on error handling. For example ([following a similar naming as on this branch](https://github.com/alphagov/govuk-frontend/compare/wip-classes)):
- `new <ComponentName>($element)` to throw if invalid
- `<ComponentName>.create($element)` not to throw
- `<ComponentName>.findOrCreate($element) not to throw and reuse existing instance`
- `<ComponentName>.findOrCreateAll($root) to find all components under `root` and initialise them all

> **Note**: While we're not necessarily planning to implement reuse of components straight in 5.0, planning our API for it and encouraging users to use these new methods could be a neat way to solve issues of double component initialisation https://github.com/alphagov/govuk-frontend/issues/1127

## Why

First step in clarifying the public API of our JavaScript components is defining what that public API will look like. Looking at the components' initialisation, especially, would let us be more intentional with what consumers end up with when creating components and limit them receiving components that may break at different points due or in different ways.

At the minute, we tend to 'return early' when trying to instantiate or initialise a component and we encounter something we don't expect. This means that the user ends up with an instantiated object, but without the expected underlying DOM or config. This may mask underlying errors in the consumers code, and calling other public methods on this object might cause other errors or unexpected behaviour.

## Who needs to work on it

Developers

## Who needs to review it

Developers

## Done when

```[tasklist]
1. [ ] Define which properties will be available on components
1. [ ] Propose a way to handle errors during components initialisation
1. [ ] Define methods available for components initialisation
```

romaricpascal:
Dumping a little though about a possible base API involving consistent errors when the element is invalid.

```js
/**
 * Base class for components that offers a clear lifecycle around component's initialisation
 * and validation that the component will run properly:
 * 
 * - sub-classes define a `configure` method responsible for storing the component
[...]

PR #4746, opened 2024-02-09
[SPIKE] Password input component - hybrid using new params

querkmachine:
A remix of #4442 using the `govukInput` macro alongside the new additions of `beforeInput`, `afterInput`, form group level `attributes`, and `attribute` values as strings.

Generally speaking, the reduction in duplication and greater dependency on our own macros is a good thing for long-term maintainability (assuming that there are guards preventing changing a thing in the parent macros immediately breaking their dependents...)

But there are a few things that I'm not so keen on:
- Duplication of default configuration as some parts are needed in Nunjucks now, whereas previously they could be entirely confined to the JavaScript module.
- Personally I'm not a fan of the "hide this in the HTML in case JavaScript needs it" way of doing things, but that's inevitable with this approach.
- Needing higher specificity CSS selector as we're targeting and overriding parts of `govuk-input` from within `govuk-password-input`. 
- It involves adding more new parameters to `govukInput`, which is something I tried to avoid previously.

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/input/template.njk
+++ b/packages/govuk-frontend/src/govuk/components/input/template.njk
@@ -21,6 +21,7 @@
     {%- if params.autocomplete %} autocomplete="{{ params.autocomplete }}"{% endif %}
     {%- if params.pattern %} pattern="{{ params.pattern }}"{% endif %}
     {%- if params.inputmode %} inputmode="{{ params.inputmode }}"{% endif %}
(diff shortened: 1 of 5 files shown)


querkmachine:
Tests are currently failing as the linter doesn't like the format for boolean attributes created by the `govukAttributes` macro. A potential fix for this is in #4743.

Regardless, I'm going to merge this down into the working branch for password input, off the back of the #3479.

PR #4760, opened 2024-02-12
Move Review app confirmation pages to `/confirm` URL paths etc

colinrotherham:
This PR includes another roundup of Review app improvements

1. Fix return value for `formatValidationErrors()` without errors
2. Fix missing confirmation page on "Check your answers" example
3. Full page example confirmation pages now append `/confirm` URL paths

For item 3) this helps test the [**Password input** component](https://github.com/alphagov/govuk-design-system/issues/3454) with back/forward page navigation

# Back links and Home navigation

Full page example "Back" and "Home" links now go to the Review app home page, just like component examples

Links to home

# Full page example grouping

Examples are now grouped for single or multiple pages. For example, forms with separate confirmation pages

You'll see each “Upload your photo” example is now shown in separate groups:

1. Upload your photo (with success message)
2. Upload your photo (with confirmation page)

Full page example grouping


querkmachine (approved):
Approval on the basis that this all seems to work and has no obvious code smells, but I'm not super experienced with Express, and there may be other problems lurking I'm not able to spot so easily. 

Would be nice to have someone with more relevant knowledge check this out too.

colinrotherham:
Thanks @querkmachine

Let's merge it, can always tweak further in another PR

Techniques wise, I've taken our existing `res.locals` approach for [`componentFixtures`](https://github.com/alphagov/govuk-frontend/blob/8a72480521c5211d36b7b55f89b21a5d815c95ba/packages/govuk-frontend-review/src/views/component.njk#L41-L43), [`componentName`](https://github.com/alphagov/govuk-frontend/blob/8a72480521c5211d36b7b55f89b21a5d815c95ba/packages/govuk-frontend-review/src/views/component-preview.njk#L3) and [`example`](https://github.com/alphagov/govuk-frontend/blob/8a72480521c5211d36b7b55f89b21a5d815c95ba/packages/govuk-frontend-review/src/views/full-page-examples/index.njk#L43-L45) in Nunjucks context and used them in URLs too, for example for redirects:

https://github.com/alphagov/govuk-frontend/blob/08bb6adc0a21c9d819fe9f2f9817e17d94e5c7fe/packages/govuk-frontend-review/src/views/full-page-examples/have-you-changed-your-name/index.mjs#L19-L23

### Redirect after POST

To explain why I've used [HTTP 303](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/303) for the confirmation page redirect, it's typically used instead of [HTTP 301](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/301) or [302](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/302) to prevent POST being retried on refresh/back where the [mdn docs for HTTP 303](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/303) use confirmation pages as an example:

>The HyperText Transfer Protocol (HTTP) `303 See Other` redirect status response code indicates that the redirects don't link to the requested resource itself, but to another page (such as a confirmation page, a representation of a real-world object

PR #4763, opened 2024-02-13
Add Puppeteer tests for Password input

querkmachine:
Adds Puppeteer tests for the [Password input component](https://github.com/alphagov/govuk-frontend/pull/4442).

Separated out as I'm not most accustomed to writing tests or Puppeteer, so open to suggestions for improvements or other tests that could be added.

## Changes
- Adds Puppeteer tests for Password input's JavaScript functionality.

## Thoughts

The form submission test is a bit weird as it uses a full-page example as the test bed, which no other component test currently does, however the full-page examples are the only examples in the review app that currently seem to support form submission and routing properly.

There are no tests for configuration or localisation options provided via JavaScript. Other components didn't seem to have tests for this, so I wasn't sure if or how best to include them.

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/password-input/password-input.mjs
+++ b/packages/govuk-frontend/src/govuk/components/password-input/password-input.mjs
@@ -63,6 +63,12 @@ export class PasswordInput extends GOVUKFrontendComponent {
       })
     }
 
+    if ($input.type !== 'password') {
+      throw new ElementError(
+        'Password input: Form field (`.govuk-js-password-input-input`) must be of type `password`.'
+      )
+    }
+
     const $showHideButton = $module.querySelector(
       '.govuk-js-password-input-toggle'
     )
@@ -73,6 +79,12 @@ export class PasswordInput extends GOVUKFrontendComponent {
         expectedType: 'HTMLButtonElement',
         identifier: 'Button (`.govuk-js-password-input-toggle`)'
       })
+    }
+
+    if ($showHideButton.type !== 'button') {
+      throw new ElementError(
+        'Password input: Button (`.govuk-js-password-input-toggle`) must be of type `button`.'
+      )
     }
 
     this.$module = $module
--- a/packages/govuk-frontend/src/govuk/components/password-input/password-input.puppeteer.test.js
+++ b/packages/govuk-frontend/src/govuk/components/password-input/password-input.puppeteer.test.js
@@ -307,7 +307,7 @@ describe('/components/password-input', () => {
           })
         })
 
-        it('throws when the input element is not the right type', async () => {
+        it('throws when the input is not an <input> element', async () => {
           await expect(
             render(page, 'password-input', examples.default, {
               beforeInitialisation($module, { selector }) {
@@ -324,6 +324,26 @@ describe('/components/password-input', () => {
               name: 'ElementError',
               message:
                 'Password input: Form field (`.govuk-js-password-input-input`) is not of type HTMLInputElement'
+            }
+          })
+        })
+
+        it('throws when the input is not a `password` type', async () => {
+          await expect(
+            render(page, 'password-input', examples.default, {
+              beforeInitialisation($module, { selector }) {
+                // Make the input a number input instead
+                $module.querySelector(selector).setAttribute('type', 'number')
+              },
+              context: {
+                selector: inputSelector
+              }
+            })
+          ).rejects.toMatchObject({
+            cause: {
+              name: 'ElementError',
+              message:
+                'Password input: Form field (`.govuk-js-password-input-input`) must be of type `password`.'
             }
           })
         })
@@ -347,7 +367,7 @@ describe('/components/password-input', () => {
           })
         })
 
-        it('throws when the button is not the right type', async () => {
+        it('throws when the button is not a <button> element', async () => {
           await expect(
             render(page, 'password-input', examples.default, {
               beforeInitialisation($module, { selector }) {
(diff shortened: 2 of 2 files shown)


packages/govuk-frontend/src/govuk/components/password-input/password-input.puppeteer.test.js
```diff
@@ -0,0 +1,363 @@
...
+      })
+
+      it('still renders an unmodified password input', async () => {
+        await render(page, 'password-input', examples.default)
+
+        const inputType = await page.evaluate(
+          (inputClass) =>
+            document.querySelector(inputClass).getAttribute('type'),
+          inputClass
+        )
```
romaricpascal, line 32: **suggestion** When looking at evaluating a function against a specific element, I think we should favour [`page.$eval`](https://pptr.dev/api/puppeteer.page._eval) rather than running `document.querySelector` inside the evaluated function.

It makes what we're evaluating against more visible and limits the function to what's actually happening to that element, so it's a little clearer.

packages/govuk-frontend/src/govuk/components/password-input/password-input.puppeteer.test.js
```diff
@@ -0,0 +1,363 @@
...
+            (buttonClass) =>
+              document.querySelector(buttonClass).getAttribute('aria-label'),
+            buttonClass
+          )
+
+          expect(buttonAriaLabel).toEqual('Hide password')
+        })
+      })
+
+      describe('when the toggle button is clicked twice', () => {
```
romaricpascal, line 153: **question** Would it be worth testing that the password component can be made visible again after having been reverted to 'password', thinking of how some bugs sometimes happen when a component seemingly reverts back to its original state after interactions, but listeners or whatnot are not quite set up properly and future interaction breaks.

romaricpascal: If useful, this could be sorted by:

- gathering the `it` calls for the 'visible' and 'hidden' state of the password into two functions: `itShowsThePassword` and `itHidesThePassword`
- using [`describe.each`](https://jestjs.io/docs/api#describeeachtablename-fn-timeout) to run the tests against a given number of clicks and check that the component is in the right state

Something along the lines of (untested, so may not work).

```js
function itShowsThePassword() {
  // Tests for each part of the state (input type, button label,...)
	it(...) {
	}
	
	it(...) {
  }
}

function itHidesThePassword() {
	// Same but for when the password is hidden
}

describe.each([
	[1, itShowsThePassword],
  [2, itHidesThePassword],
  [3, itShowsThePassword]
])('when clicked %i time(s)', (clicks, expectation)  => {
	beforeAll(async () => {
    await render(page, 'password-input', examples.default)
    for (i = 0; i< clicks; i++) {
      await page.click(buttonClass)
    }
  })

  expectation()
}) 
```

packages/govuk-frontend/src/govuk/components/password-input/password-input.puppeteer.test.js
```diff
@@ -0,0 +1,363 @@
...
+          ).rejects.toMatchObject({
+            cause: {
+              name: 'ElementError',
+              message:
+                'Password input: Form field (`.govuk-js-password-input-input`) not found'
+            }
+          })
+        })
+
+        it('throws when the input element is not the right type', async () => {
```
romaricpascal, line 310: **question** This test checking against what class the element is an instance of made me wonder if it should also check if the `type` attribute of the input is appropriate (ie. `password`, or at least not something weird like `button`, `reset` or `image`). It might:
- have already been discussed (in which case, do ignore me)
- be a bit over the top (in which case, same)

romaricpascal (commented):
Looking pretty solid in terms of the scope of the test (from what I gathered of what the component was doing). Possibly some i18n tests missing (I _think_ there was a commit for it at some point, so it could be they got lost in a rebase)

Just suggested a couple of changes and questions, but nothing blocking that I can see.

packages/govuk-frontend/src/govuk/components/password-input/password-input.mjs
```diff
@@ -63,6 +63,12 @@ export class PasswordInput extends GOVUKFrontendComponent {
       })
     }
 
+    if ($input.type !== 'password') {
+      throw new ElementError(
```
romaricpascal, line 67: Yeah, that would be the right one for that case. 🙌🏻

packages/govuk-frontend/src/govuk/components/password-input/password-input.mjs
```diff
@@ -75,6 +81,12 @@ export class PasswordInput extends GOVUKFrontendComponent {
       })
     }
 
+    if ($showHideButton.type !== 'button') {
```
romaricpascal, line 84: Oh, nice catch on making sure that button has the right type as well! 🙌🏻

romaricpascal (commented):
Looking nice! Thanks for updating after my suggestions. 😊  

I think we're missing tests for how the JavaScript uses the translations, though. I found back this commit https://github.com/alphagov/govuk-frontend/commit/ca304eb54148046ff7135c13c91b49faac02491a which tests that Nunjucks passes the translation strings down to the template, but couldn't find something that checks that the JavaScript component makes good use of those strings.

romaricpascal:
We might be able to piggyback on that existing `describe.each` to implement the test for the translations with something like (untested):

```js
describe.each([
  [
    "default",
    {
      buttonTextWhenShown: "Hide",
      buttonLabelWhenShown: "Hide password",
      statusWhenShown: "Your password is shown",
      buttonTextWhenHidden: "Show",
      buttonLabelWhenHidden: "Show password",
      statusWhenHidden: "Your password is hidden",
    },
  ],
  [
    "with translations",
    {
      buttonTextWhenShown: "Cuddio",
      buttonLabelWhenShown: "Cuddio cyfrinair",
      statusWhenShown: "Mae eich cyfrinair yn weladwy.",
      buttonTextWhenHidden: "Datguddia",
      buttonLabelWhenHidden: "Datgelu cyfrinair",
      statusWhenHidden: "Mae eich cyfrinair wedi'i guddio.",
    },
  ],
])('%s', (exampleName, expectedTranslations) => {
  describe.each([
    [1, itShowsThePassword],
    [2, itHidesThePassword],
    [3, itShowsThePassword],
  ])("when clicked %i time(s)", (clicks, expectation) => {
    beforeAll(async () => {
      await render(page, "password-input", examples[exampleName]);
      for (let i = 0; i < clicks; i++) {
        await page.click(buttonSelector);
      }
    });

    expectation(expectedTranslations);
  });
});

// And update the `itShowsThePassword`/`itHidesThePassword` to receive translations
function itShowsThePassword(expectedTranslations) {
	//...

  it('changes the button text to "hide"', async () => {
    const buttonText = await page.$eval(buttonSelector, (el) =>
      el.innerHTML.trim()
    )
	
    expect(buttonText).toBe(expectedTranslations.buttonTextWhenShown)
  })
  
  	//...
}
```

querkmachine:
Discussion in Slack indicated that the above suggestion would probably make the tests more complex than they needed to be. Added simpler test for translation strings instead.

PR #4764, opened 2024-02-13
Fix Nunjucks HTML indentation: Tabs

colinrotherham:
Tabs changes, split out from https://github.com/alphagov/govuk-frontend/pull/4448 to partially resolve #3211

This PR adds `| trim | indent(4)` etc to tab items and panels to ensure nested HTML lines are indented, using:

* `_tabListItem()`
* `_tabPanel()`


domoscargin (approved):
Nice - output looks real neat.

I like the macro approach - it makes the general gist of the template easier to read beyond just being helpful for indenting. Plus we seem to have already been using it on Summary List, so good precedent.

PR #4766, opened 2024-02-14
Fix: automated axe testing custom options overridden by later configuration

domoscargin:
## Bug
When setting up our axe reporter, we define some default rules (using `.withRules()`), and we allow rule overrides.

Later, however, we run the reporter like this:

`const report = await reporter.options({ rules }).analyze()`

This call to `.options()` [overwrites all our previous rules](https://github.com/dequelabs/axe-core-npm/tree/develop/packages/puppeteer#axepuppeteeroptionsoptions-axerunoptions), resulting in us unknowingly not running the correct checks.

## Solution
Instead of using the AxePuppeteer chained methods, we define an AxeOptions object, which we can then merge with passed-in overrides.

At the moment, we only override `options.rules`, so I've kept that behaviour. However, it would be very simple to change this behaviour so that we can pass in a complete set of custom axe options for a particular test.

## Future
In the future, it'd be really nice to be able to pass in some kind of filter option, so we can get really specific about what cases we want to disable (eg: "Ignore the `aria-allowed-attr` for radios with `aria-controls="conditional-[id]""). At the moment it's quite brute force: "Disable this rule entirely for every radio example".

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/accordion/accessibility.puppeteer.test.mjs
+++ b/packages/govuk-frontend/src/govuk/components/accordion/accessibility.puppeteer.test.mjs
@@ -2,13 +2,26 @@ import { axe, render } from '@govuk-frontend/helpers/puppeteer'
 import { getExamples } from '@govuk-frontend/lib/components'
 
 describe('/components/accordion', () => {
+  let axeRules
+
+  beforeAll(() => {
+    axeRules = {
+      /**
+       * Ignore 'Element has insufficient color contrast' for WCAG Level AAA
+       *
+       * Affects 'Show all sections', 'Show' and 'Hide' toggle links
+       */
+      'color-contrast-enhanced': { enabled: false }
+    }
+  })
+
   describe('component examples', () => {
     it('passes accessibility tests', async () => {
       const examples = await getExamples('accordion')
 
       for (const exampleName in examples) {
         await render(page, 'accordion', examples[exampleName])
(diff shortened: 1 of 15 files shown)


packages/govuk-frontend/src/govuk/components/accordion/accessibility.puppeteer.test.mjs
```diff
@@ -2,13 +2,22 @@ import { axe, render } from '@govuk-frontend/helpers/puppeteer'
 import { getExamples } from '@govuk-frontend/lib/components'
 
 describe('/components/accordion', () => {
+  let axeRules
+
+  beforeAll(() => {
+    axeRules = {
+      // Ignore colour contrast (enhanced) from WCAG Level AAA
```
colinrotherham, line 9: Thanks for adding these @domoscargin

Shall we be nice to our future selves and say what it was ignored for?

```suggestion
      /**
       * Ignore 'Element has insufficient color contrast' for WCAG Level AAA
       * 
       * Affects 'Show all sections', 'Show' and 'Hide' toggle links
       */
```

domoscargin: Done!

colinrotherham (approved):
Looks great to me 🙌

Glad we hadn't missed anything, but we're all set should new tags/rules be added in future

PR #4767, opened 2024-02-15
Fix Nunjucks HTML indentation: Task list, Tag

colinrotherham:
Task list and Tag changes, split out from https://github.com/alphagov/govuk-frontend/pull/4448 to partially resolve #3211

This PR adds `| trim | indent(8)` etc to task list titles, hints and tags to ensure nested HTML lines are indented, using:

* `_taskListItem()`


domoscargin (approved):
Diff looks good!

PR #4768, opened 2024-02-16
Fix `z-index` of inputs in Radios and Checkboxes component

### LINKED ISSUE #4762: ElementClickInterceptedError using Selenium in 5.1.0
## Description of the issue

5.1.0 contains some refactoring to the CSS for radio buttons and checkboxes resulting in these form controls no longer able to be selected / checked by ID using Selenium. This really should have been a breaking change in my opinion, as it will break a lot of acceptance test suites(including ours).

```
const { Builder, By } = require('selenium-webdriver');

(async function helloSelenium() {
  try {
    let driver = await new Builder().forBrowser('chrome').build();

    await driver.get('https://design-system.service.gov.uk/components/radios/default/');

    const wales = await driver.findElement(By.id('whereDoYouLive-3'));
    wales.click();
  } catch (e) {
    console.log(e);
  }
})();
```

```
ElementClickInterceptedError: element click intercepted: Element <input class="govuk-radios__input" id="whereDoYouLive-3" name="whereDoYouLive" type="radio" value="wales"> is not clickable at point (52, 215). Other element would receive the click: <label class="govuk-label govuk-radios__label" for="whereDoYouLive-3">...</label>
```

- Actual: Selenium throws the above error as the pseudo elements are positioned over the top of the input
- Expected: The radio to be selected

romaricpascal:
The `z-index: 1` declaration laying out the input above their corresponding label had disappeared when we updated how the elements were laid out to Flexbox. This broke Selenium testing suites trying to click the input (now covered by its label).

This PR reinstates the declaration, and adds automated tests for both components to ensure the order in which the elements are overlaid is checked so we won't miss this in the future.

Closes #4762

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,12 @@ As the beginning of the GOV.UK logo rollout has passed, the Tudor crown logo is 
 
 The `useTudorCrown` parameter, along with any other adjustments made to display the Tudor crown logo in your service, can now be removed.
 
+### Fixes
+
+We've made fixes to GOV.UK Frontend in the following pull requests:
+
+- [#4768 : Fix z-index of inputs in Radios and Checkboxes component](https://github.com/alphagov/govuk-frontend/pull/4768)
+
 ## 5.1.0 (Feature release)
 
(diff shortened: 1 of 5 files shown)


owenatgov (approved):
Brill!

Issue #4783, opened 2024-02-20
Improve our Sass workflow to check for compilation of variables and functions with LibSass and Ruby Sass


romaricpascal:
## What

Improve [our Sass workflow](https://github.com/alphagov/govuk-frontend/blob/main/.github/workflows/sass.yaml) to verify that Libsass and Ruby Sass correctly transform all the Sass variables and functions in our code.

## Why

Our Sass code can still be imported with Libsass and Ruby Sass (as we've yet to [move to Sass module system](https://github.com/alphagov/govuk-frontend/issues/1791)). However these [older versions do not transform some newer syntaxes supported by dart Sass, outputting invalid CSS](https://github.com/alphagov/govuk-frontend/issues/4782).

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [x] Our Sass workflow checks for the presence of Sass variables in the built ouput
- [ ] Our Sass workflow checks for the presence of function calls starting with `govuk-` in the built output.

romaricpascal:
A possible way to achieve this could be to run a custom Stylelint plugin with two rules:
- one failing at the presence of a Sass variable
- one failing at the presence of a function starting with `govuk-` (as there may be legitimate calls to CSS functions and we don't want to keep an eye on all the functions in CSS as of today or introduced in the future).

romaricpascal:
#4784 nicely finds the Sass variables using `grep`. If we can use `grep` to match the function calls as well, that would be ace. This may require us to use the [`-E`](https://explainshell.com/explain?cmd=grep+-E+%27govuk-%5B%5Cw-%5D%2B%5C%28%27) or [`-P`](https://explainshell.com/explain?cmd=grep+-P+%27govuk-[\w-]%2B\(%27) flag for more complex regexp.

colinrotherham:
Bit late now, but this would do it:

```shell
! grep --regexp "\$govuk-" --regexp "govuk\-\([a-z0-9-]\+\)(" .tmp/all.css
```

36degrees:
This is no longer relevant as we're dropping support for LibSass and Ruby Sass in v6.

Issue #4782, opened 2024-02-20
Interpolate sass variables inside calc() to maintain support for compiling with libsass


oscarduignan:
## Description of the issue

[In 5.1.0 you introduced some styles that mean that your sass is no longer compatible with libsass due to lack of interpolation within some uses of "calc()"](https://github.com/alphagov/govuk-frontend/commit/d6e38b14828f997f243654c66c29ddee531343ea#r138836052)

## Steps to reproduce the issue

1. grab the latest from main
2. `cat dist/*.css | npx prettier --stdin-filepath .css | grep "calc(100% -"`
3. you will see this:
   ```
   max-width: calc(100% - 74px);
   max-width: calc(100% - 74px);
   ```
2. `npm install node-sass`
3. `cd src/govuk`
4. `npx node-sass all.scss | grep "calc(100% -"`
5. and now you will see:
   ```
   max-width: calc(100% - (($govuk-checkboxes-label-padding-left-right * 2) + $govuk-touch-target-size));
   max-width: calc(100% - ($govuk-radios-label-padding-left-right + $govuk-touch-target-size + govuk-spacing(3)));
   ```

because the contents of `calc()` are treated as an unquoted string in libsass, it outputs everything literally

in hmrc-frontend I think we must have something that's throwing an exception on this - possibly stylelint or something like this when we build - but node-sass won't through an error itself, it will just output it literally, so it's possible that 5.1.0 might break some people's usages of govuk-frontend unknowingly if they are using node-sass to compile your scss without any other checks that it generated valid css

## Actual vs expected behaviour

libsass is deprecated but you could maintain compatibility with it by interpolating the sass variables like this:
```
   max-width: calc(100% - #{(($govuk-checkboxes-label-padding-left-right * 2) + $govuk-touch-target-size)});
   max-width: calc(100% - #{($govuk-radios-label-padding-left-right + $govuk-touch-target-size + govuk-spacing(3))});
```
the before and after output with dart-sass and libsass is:
```
   max-width: calc(100% - 74px);
   max-width: calc(100% - 74px);
```

([for an example see this branch in a fork](https://github.com/alphagov/govuk-frontend/compare/main...oscarduignan:govuk-frontend:interpolate-vars-in-calc-to-maintain-node-sass-support))

## Environment (where applicable)

we're using node-sass@7 (sorry) but I tried updating hmrc-frontend through all the newer major versions (8 and 9) and still got the issue with compilation.

- Operating system: N/A
[...]

PR #4784, opened 2024-02-20
Fix LibSass `calc()` compatibility in Radios and Checkboxes

colinrotherham:
This PR interpolates Sass `calc()` contents for compatibility with LibSass

It fixes https://github.com/alphagov/govuk-frontend/issues/4782 using the suggestion from @oscarduignan

## LibSass compatibility

Our Sass tests didn't catch LibSass `calc()` differences in https://github.com/alphagov/govuk-frontend/pull/4093

This was unintentional and [compatibility with LibSass and Ruby Sass](https://frontend.design-system.service.gov.uk/installing-with-npm/#install-with-node-js-package-manager-npm) should be maintained

>Although GOV.UK Frontend currently supports LibSass (version 3.3.0 and above) and Ruby Sass (version 3.4.0 and above), we will remove support in future. If you’re using either of these Sass compilers, you should [migrate to Dart Sass](https://sass-lang.com/blog/libsass-is-deprecated#how-do-i-migrate) as soon as you reasonably can.

## Updated tests

I've added new checks for `$govuk-*` in Sass output, using [grep](https://www.gnu.org/software/grep/manual/grep.html) with `!` to throw when matches are ~not~ found:

```shell
! grep "\$govuk-" .tmp/all.css
```

This ensures we break the build and would output the following when matches are found:

```console
  max-width: calc(100% - (($govuk-checkboxes-label-padding-left-right * 2) + $govuk-touch-target-size));
  max-width: calc(100% - ($govuk-radios-label-padding-left-right + $govuk-touch-target-size + govuk-spacing(3)));
```

This partially closes https://github.com/alphagov/govuk-frontend/issues/4783 but does not check for `govuk-spacing()` etc

### DIFF
--- a/.github/workflows/sass.yaml
+++ b/.github/workflows/sass.yaml
@@ -36,7 +36,14 @@ jobs:
           sass --version
 
       - name: Run command
-        run: time sass packages/govuk-frontend/src/govuk/all.scss > /dev/null
+        run: |
+          mkdir -p .tmp
+          time sass packages/govuk-frontend/src/govuk/all.scss > .tmp/all.css
+
+      # Check output for uncompiled Sass
+      - name: Check output
+        run: |
+          ! grep "\$govuk-" .tmp/all.css
 
   dart-sass-latest:
     name: Dart Sass v1 (latest)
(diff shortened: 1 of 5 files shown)


romaricpascal (approved):
Glad we ended with a nice output in the end by splitting the grep in its own step 🥳 ⛵

PR #4791, opened 2024-02-23
Add new config schema `allOf` condition support

colinrotherham:
To offload some old code here's support for `allOf` schema conditions (alongside `anyOf`)

Might never need this, but to prevent `i18n` or `rememberExpanded` being accidentally overridden by null/undefined:

```mjs
/**
  * Accordion config schema
  *
  * @constant
  * @satisfies {Schema}
  */
static schema = Object.freeze({
  allOf: [
    {
      required: ['i18n'],
      errorMessage: '"i18n" must be provided'
    },
    {
      required: ['rememberExpanded'],
      errorMessage: '"rememberExpanded" must be provided'
    }
  ]
})
```


colinrotherham:
All the commits in this PR (except for one) got merged in:

* https://github.com/alphagov/govuk-frontend/pull/4792

Once the [branch is rebased](https://github.com/alphagov/govuk-frontend/compare/main...colinrotherham:govuk-frontend:schema-conditions) these are the only lines that remain:

```mjs
// Check all conditions pass or add errors
if (name === 'allOf' && errors.length) {
  validationErrors.push(...errors)
}
```

Izabela-16:
We'll be waiting for a component to require an allOf validation, so we'll close this for now. We will likely do something like this when it happens.

PR #4806, opened 2024-02-27
Add Rollup stats for minified files

colinrotherham:
This PR splits Rollup stats into "bundled" versus "minified" file sizes

It adds another call to [rollup-plugin-visualizer](https://www.npmjs.com/package/rollup-plugin-visualizer) with minified sizes as the default for tree maps

Rollup minified file stats

### DIFF
--- a/.github/workflows/scripts/comments.mjs
+++ b/.github/workflows/scripts/comments.mjs
@@ -122,7 +122,10 @@ export async function commentStats(
       const statsPath = `docs/stats/${modulePath.replace('mjs', 'html')}`
       const statsURL = new URL(statsPath, reviewAppURL)
 
-      return [`[${modulePath}](${statsURL})`, moduleSize]
+      return [
+        `[${modulePath}](${statsURL})`,
+        `${moduleSize.bundled} (bundled)<br>${moduleSize.minified} (minified)`
+      ]
     }
   )
 
--- a/.github/workflows/scripts/comments.mjs
+++ b/.github/workflows/scripts/comments.mjs
@@ -1,5 +1,5 @@
 import { readFile } from 'node:fs/promises'
-import { basename, join } from 'path'
+import { basename, join, parse } from 'path'
 
 import { getFileSizes } from '@govuk-frontend/lib/files'
 import { getStats, modulePaths } from '@govuk-frontend/stats'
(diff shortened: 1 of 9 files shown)


shared/stats/rollup.config.mjs
```diff
@@ -39,24 +50,35 @@ export default defineConfig(
...
-              `${parse(modulePath).name}.html`
-            ),
+            emitFile: true,
+            filename: `${parse(modulePath).name}.html`,
             projectRoot: dirname(packagePath),
-            template: 'treemap'
+            template: 'treemap',
+
+            // Use sourcemaps to calculate minified sizes
+            sourcemap: true
```
domoscargin, line 81: I'm not opposed to this at all, but good to get your rationale for using minified sizes for the tree map.

colinrotherham: We use Babel to [preserve JSDoc blocks with `@param`, `@returns` and `@typdef`](https://github.com/alphagov/govuk-frontend/blob/f8aa28e2c34fc6803c59f124747d675f5568a9ac/packages/govuk-frontend/babel.config.js#L29-L32) (unless private/internal)

Browsing the map points out "which module is most documented" rather than the largest, so by defaulting to the minified size it better represents what ends up in browsers

Gzip/Brotli next one day?

domoscargin (approved):
Very cool!

shared/stats/rollup.config.mjs
```diff
@@ -39,24 +50,35 @@ export default defineConfig(
...
-              dirname(modulePath),
-              `${parse(modulePath).name}.yaml`
-            ),
+            emitFile: true,
+            filename: `${parse(modulePath).name}.yaml`,
+            projectRoot: dirname(packagePath),
+            template: 'list',
+
+            // Skip sourcemaps to use original file sizes
+            sourcemap: false
```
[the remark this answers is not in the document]
colinrotherham: We do the same in other Rollup configs to explain the defaults

For example, these two options:

https://github.com/alphagov/govuk-frontend/blob/f8aa28e2c34fc6803c59f124747d675f5568a9ac/packages/govuk-frontend/rollup.publish.config.mjs#L27-L36

.github/workflows/scripts/comments.mjs
```diff
@@ -124,12 +124,13 @@ export async function commentStats(
 
       return [
         `[${modulePath}](${statsURL})`,
-        `${moduleSize.bundled} (bundled)<br>${moduleSize.minified} (minified)`
```
domoscargin: 👍

PR #4808, opened 2024-02-29
Prevent non-object values from being merged into config keys already set to an object

romaricpascal:
This limits the risk of mistaken JavaScript keys or data attributes completely wiping some config values that use an object (`i18n` messages and plural forms).

It takes advantage of the fact that everything is merged onto the default config, so object values will already be set, allowing us to detect that we'd be replacing them with a non-object one.

Similarly, when going through the data attributes, going in order allows deeper keys (eg. `data-i18n.messageKey`) to properly override a non-object value (eg. `data-i18n`), ensuring an object is created.

This bring the behaviour of the new merging closer to the one with flatten objects, which guaranteed access to nested keys set by the default when the config or dataset would set shallower keys to a non-object value.

## Why

In its current form #4792 doesn't quite behave like the flattened attributes. 

Because of the flattening, merging the following together:

```js
mergeConfigs({
  a: 'antelope',
  'c.a': 'camel'
},
{
  b: 'bee',
  c: 'whoops!'
})
```
Would create the following:

```js
({
  a: 'antelope'
  b: 'bee',
  'c.a': 'camel',
  c: 'whoops!'
})
```

But #4792 creates:

```js
({
  a: 'antelope',
  b: 'bee',
  c: 'whoops!'
})
```

We'd lost the access the nested value `c.a`.

### DIFF
--- a/packages/govuk-frontend/src/govuk/common/index.jsdom.test.mjs
+++ b/packages/govuk-frontend/src/govuk/common/index.jsdom.test.mjs
@@ -160,6 +160,22 @@ describe('Common JS utilities', () => {
       const result = extractConfigByNamespace($element.dataset, 'i18n')
       expect(result).toEqual({ key1: 'One', key2: 'Two' })
     })
+
+    it('can handle multiple levels of nesting', () => {
+      document.body.outerHTML = outdent`
+        <div id="app-example2"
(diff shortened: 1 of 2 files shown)


romaricpascal:
Closing as it's been taken care of by #4792

PR #4809, opened 2024-02-29
[SPIKE] Throw attempting to override an object value with a non-object 

romaricpascal:
Make both `mergeConfig` and `extractConfigByNamespace` throw if they try to merge a non-object value on a key that already stores an object value.

The errors would let users know that they are (likely unintentionally) passing a property or data-attribute that conflicts with deeper values (for example setting `data-i18n` which would default with the configuration for the default message). 

While this sounds a handy information:
- #4808 ignores merges of values for 'shallow' configuration keys when a deeper one is already set
- these error may cause some people to have to edit their code, are they cause for being a breaking change?

## Implementation note

Was a bit shy to pass `componentName` to the helper functions, but that could be an alternative solution that avoids catching and re-throwing the errors. Doesn't feel super tidy from a separation of responsibility point of view.

### DIFF
--- a/packages/govuk-frontend/src/govuk/common/index.jsdom.test.mjs
+++ b/packages/govuk-frontend/src/govuk/common/index.jsdom.test.mjs
@@ -31,7 +31,7 @@ describe('Common JS utilities', () => {
     }
 
     it('ignores a single object', () => {
-      const config = mergeConfigs(config1)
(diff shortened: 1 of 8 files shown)


romaricpascal:
Closing to avoid noise in the list of PRs now that #4792 is merged.

Issue #4709, opened 2024-01-31
Replace calls to deprecated `KeyboardEvent.keyCode`


36degrees:
## What

Replace calls to the deprecated `KeyboardEvent.keyCode` from our components, including Button, Exit this Page and Tabs.

> [!NOTE]  
> We probably want to switch to using [`KeyboardEvent.key`](https://browsersl.ist/#q=supports+es6-module+and+not+supports+keyboardevent-key%0A) which has better support than  [`KeyboardEvent.code`](https://browsersl.ist/#q=supports+es6-module+and+not+supports+keyboardevent-code). Noteably, `KeyboardEvent.code` is not supported by Edge 15-18 and will be `undefined`.

The Exit this Page component is more complex and uses both `KeyboardEvent.keyCode` and `KeyboardEvent.key` – this may need more investigation and testing than the other components.

## Why

`KeyboardEvent.keyCode` is deprecated.

Using the modern `KeyboardEvent.key` will allow to simplify our code – for example, we can get rid of the hardcoded ASCII codes for the arrow and space keys.

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [x] We no longer use the deprecated `KeyboardEvent.keyCode` property in our code

PR #4811, opened 2024-03-01
Use `KeyboardEvent.key` over deprecated `KeyboardEvent.keyCode` in the Tabs component

36degrees:
`KeyboardEvent.keyCode` is deprecated. All of the browsers that now run our JavaScript support the modern `KeyboardEvent.key` property.

Update the button component to using `KeyboardEvent.key` instead, removing hardcoded ASCII values in the `keys` map.

Unfortunately Edge 16 uses e.g. “Left" instead of “ArrowLeft”, so we need to add those possible values as well if we want keyboard navigation for tabs to work in that browser (see the question I've left further down as to whether that's sensible or not).

[1]: https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_key_values#navigation_keys

Part of #4709

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/tabs/tabs.mjs
+++ b/packages/govuk-frontend/src/govuk/components/tabs/tabs.mjs
@@ -21,9 +21,6 @@ export class Tabs extends GOVUKFrontendComponent {
   $tabListItems
 
   /** @private */
-  keys = { left: 37, right: 39, up: 38, down: 40 }
-
-  /** @private */
   jsHiddenClass = 'govuk-tabs__panel--hidden'
 
   /** @private */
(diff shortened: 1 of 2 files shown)


packages/govuk-frontend/src/govuk/components/tabs/tabs.mjs
```diff
@@ -375,14 +372,19 @@ export class Tabs extends GOVUKFrontendComponent {
    * @param {KeyboardEvent} event - Keydown event
    */
   onTabKeydown(event) {
-    switch (event.keyCode) {
-      case this.keys.left:
-      case this.keys.up:
+    switch (event.key) {
+      // 'Left', 'Right', 'Up' and 'Down' required for Edge 16 support.
```
36degrees, line 376: If we wanted to save a few bytes, we could ignore this and just go with the values in the spec. It shouldn't cause errors – it'd just mean that navigating between the tabs using the keyboard wouldn't work in Edge 16.

The number of users visiting GOV.UK (based on those who've opted in to analytics) are _vanishingly_ small – between 20 to 50 sessions per day in February (out of 95,335,104 – so about 0.00004% of our sessions).

Thoughts?

colinrotherham (approved):

packages/govuk-frontend/src/govuk/components/tabs/tabs.mjs
36degrees: We discussed this in our dev catch up and decided that because the only way to move between tabs using the keyboard is using the arrow keys we should keep this in to properly support Edge 16.

PR #4842, opened 2024-03-11
Add eslint-plugin-jest and enable the recommended rules

### LINKED ISSUE #4238: Investigate using eslint-plugin-jest
## What

Investigate whether we should install and use `eslint-plugin-jest` as part of our ESLint setup. 

## Why

We had a single instance of a `.only()` (used to run only a single test in a test suite during development) being left in a merged PR, which we think `eslint-plugin-jest` would have caught.

It looks as though [the other rules](https://www.npmjs.com/package/eslint-plugin-jest#Rules) included with `eslint-plugin-jest` would also help us to write better tests and to avoid other mistakes sneaking in to our codebase.

## Assumptions

- We're still using ES Lint by the time we're working on this…

## Timebox

> [!NOTE]
> We need to agree a timebox for this spike

## Who is working on this?

> [!NOTE]
> We need to agree who will work on this spike.

Spike lead: TBC.
Spike buddy: TBC.

## Questions to answer

- [x] Does the plugin catch any other inconsistencies or errors in our tests?
- [x] Are the error messages clear and instructive when rules match? Would a contributor understand what needs to be changed to fix it?
- [x] Does the plugin interfere with the development process? (for example, making it harder to temporarily use `.only` during development)
- [x] Should we install the plugin?
- [x] Are there any changes to the default config that we might want to make? Including optional rules that we might want to consider enabling.

## Done when

- [x] Questions have been answered or we have a clearer idea of how to get to our goal
- [ ] Findings have been reviewed and agreed with at least one other person
- [ ] Findings have been shared, e.g: via a write-up on the ticket, at a show & tell or team meeting

36degrees:
Suggest reviewing commit-by-commit.

Closes #4238


36degrees:
Putting this back into draft for now as I realise I hadn't checked how this works during local development (one of the points listed on the issue)

36degrees:
Looks like it's fine when developing locally – have added more detail to the issue.

packages/govuk-frontend/src/govuk/components/checkboxes/checkboxes.puppeteer.test.js
```diff
@@ -201,11 +201,13 @@ describe('Checkboxes', () => {
       describe('with conditional items with special characters', () => {
         it('does not error when ID of revealed content contains special characters', async () => {
           // Errors logged to the console will cause this test to fail
-          await render(
-            page,
-            'checkboxes',
-            examples['with conditional items with special characters']
-          )
+          return expect(
```
romaricpascal, line 204: **nitpick** If we no longer `await` inside a function and `return` the promise from `expect`, the test function no longer needs to be `async` and can just be a plain function.

romaricpascal: Looks like there's an ESLint rule for that: https://eslint.org/docs/latest/rules/require-await. It can probably be enabled as its own PR, though.

romaricpascal (approved):
That should make things more consistent in the testing (and avoid the `.only`). Like it! 🙌🏻  

Noticed a few further tidy-ups we could make around `async` functions, but they're probably the scope of their own PR more than this one, so happy to merge like that.

packages/govuk-frontend/src/govuk/components/checkboxes/checkboxes.puppeteer.test.js
36degrees: From a quick test it looks like it does flag this and another instance on line 75. (57 across the entire repo).

Happy for that to be a follow up thing?

romaricpascal: Yup, I thought there'd be more across the repo, so let's not worry about it for this PR and raise a separate one for `require-await`.

domoscargin (approved):
I think this all looks sensible to me.

PR #4843, opened 2024-03-11
Enable 'style' rule configuration for eslint-plugin-jest

36degrees:
Using the `style` configuration over the `recommended` config enables three more opinionated rules:

- `prefer-to-be` – Suggest using `toBe()` for primitive literals
- `prefer-to-contain` – Suggest using `toContain()`
- `prefer-to-have-length` - Suggest using `toHaveLength()`

These rules are all automagically fixable using `--fix`.

Raised for discussion as to whether these would be useful in addition to the recommended rules enabled in #4842.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -17,7 +17,7 @@ module.exports = {
       extends: [
         'eslint:recommended',
         'plugin:import/recommended',
-        'plugin:jest/recommended',
+        'plugin:jest/style',
         'plugin:jsdoc/recommended-typescript-flavor',
         'plugin:n/recommended',
         'plugin:promise/recommended',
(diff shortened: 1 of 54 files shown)


romaricpascal:
Given all these rules [are automatically fixable](https://www.npmjs.com/package/eslint-plugin-jest#rules), I think it's pretty safe to merge and have consistent matchers in the code base. The linting pre-commit (or your editor if it lints on save) will put the right matcher when necessary, as well, so the burden should be quite light if you're used to writing the disallowed version.

domoscargin (approved):
That's a lotta `toBe`'s!

But yeah, this looks good, very light burden, better consistency, and I personally agree with the rules, so all good from me.

PR #4854, opened 2024-03-12
Remove redundant `role` attributes

### LINKED ISSUE #4708: Remove redundant explicit role attributes from HTML5 elements
## What

Remove the explicit `role` attribute on HTML elements that have the same role implicitly, including:

- the `<header>` element in the Header component, which has a `banner` role
- the `<footer>` element in the Footer component, which has a `contentinfo` role
- the `<nav>` element in the Pagination component, which has a `navigation` role

In some places we include an explicit `group` role on `<fieldset>` elements (e.g. the date input) to ensure that JAWS announces the accessible description – we should **leave this in place** unless we're able to do further investigation and understand whether this is still needed.

Update our [HTML validation rules](https://github.com/alphagov/govuk-frontend/blob/8eeca8e8a7980dfaceb6619ca97084fdb63634f6/packages/govuk-frontend/.htmlvalidate.js#L27-L30) accordingly.

## Why

These have historically been set because older browsers ([noteably Internet Explorer](https://web.archive.org/web/20180103181655/https://connect.microsoft.com/IE/feedback/details/804723/html5-structural-elements-not-exposed-via-accessibility-api)) do not [map these implicit roles correctly](https://www.html5accessibility.com/).

As our browser support model has evolved, we can now remove these implicit roles.

## Who needs to work on this

Developers

## Who needs to review this

Developers, Accessibility Specialist

## Done when

- [x] We are no longer adding redundant role attributes to elements that have the same implicit role (except `<fieldset>`)
- [x] HTML validation rules are updated to reflect this change

querkmachine:
Closes #4708.

## Changes
- Removes `role` attributes in components and review app examples for each of the following:
  - `banner` role on `header`
  - `contentinfo` role on `footer`
  - `main` role on `main`
  - `navigation` role on `nav`
  - `complementary` role on `aside`
  - `region` role on `section`
  - `article` role on `article`
- Removes component tests checking for the existence of these roles.
- Updates the html-validate config to:
  - Remove the `no-redundant-role` rule override. This now defaults back to not allowing redundant `role` attributes.
  - Add an exception for the use of the `group` role on `fieldset` elements, which we're maintaining for assistive technology support.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,23 @@ Check your changes against [the character count example in the Design System](ht
 
 This change was introduced in [pull request #4566: Use Character count `formGroup` as module wrapper](https://github.com/alphagov/govuk-frontend/pull/4566).
 
+#### Remove redundant `role` attributes from elements
+
+We've made minor changes to the HTML of the page template, as well as the header, footer and pagination components.
+
+You can update your HTML to remove the `role` attribute from some elements. These include:
+
+- the `main` role on the `main` element in the template.
+- the `banner` role on the `header` element in the header component.
+- the `contentinfo` role on the `footer` element in the footer component.
+- the `navigation` role on the `nav` element in the pagination component.
+
+These roles were present to support legacy browsers, such as older versions of Internet Explorer. GOV.UK Frontend no longer supports these browsers, so the roles can now be removed.
+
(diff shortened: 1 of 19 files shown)


owenatgov (approved):
By my eye you got them all. It feels a bit weird that there are now no tests for the default header but that's not a today problem.

Issue #4841, opened 2024-03-08
Header alignment issue using new type scale


colinrotherham:
## Description of the issue

When using the [new type scale in v5.2.0](https://github.com/alphagov/govuk-frontend/releases/tag/v5.2.0) there's an alignment issue affecting the header product name

```scss
$govuk-new-typography-scale: true;
```

<img src="https://github.com/alphagov/govuk-frontend/assets/415517/97d65ad6-37e4-4e93-8b84-6f3ab61177e4" alt="Header bug" width="50%">

<br>
<br>

<img src="https://github.com/alphagov/govuk-frontend/assets/415517/22bd97bd-1574-4bfe-8936-a7c4c901d62e" alt="Firefox at 2x" width="75%">

## Steps to reproduce the issue

Using a mobile viewport:

* Open the [**Header** with product name](https://govuk-frontend-review.herokuapp.com/components/header/with-product-name/preview) Review app example
* Open the [**GOV.UK Design System** website](https://design-system.service.gov.uk)

## Actual vs expected behaviour

It's actually not aligned but it's expected to be aligned

Ciao for now 👋

PR #4855, opened 2024-03-12
Fix mobile product name being misaligned in new type scale

querkmachine:
Intended to be a 'quick fix' of the most visible issue mentioned in #4841. There are still potential alignment issues documented below.

## Changes
- Modifies the `margin-top` of the product name on the mobile breakpoint, if the new type scale is being used.

## Screenshots

(All screenshots are in Chromium.)

||Before|After|
|:-|:-:|:-:|
|Narrow mobile|before-narrow-mobile|after-narrow-mobile|
|Mobile|before-mobile|after-mobile|
|Tablet+|before-tablet|after-tablet|

This change principally aims to fix the middle row here ("Mobile") although it does have repercussions on the narrow mobile view too, as the product name appears slightly higher than previously, which may be undesirable. 

(Note that the logo and product name always link to the same place, so proximity to one another shouldn't be a problem regarding accessibility.)

## Thoughts
In typically frustrating fashion, each browser and screen has a slightly different interpretation of how to align the product name with respect to the logo depending on our dreaded foe, **subpixel rounding**.

The baseline misalignment between GOV.UK logo and product name on my screens appears to be:

|Browser|MBP display<br>(3024×1964, 254 ppi)|External display<br>(2560×1440, 123 ppi)|
|:-|:-:|:-:|
|Chromium|0|-1px|
|Firefox (w/ -0.5px hack)|-1px|0|
|Firefox (w/o -0.5px hack)|0|0|
|Safari|+2px|-1px|

I'm not sure how much pixel nudging we're up to doing to try and minimise the difference between browsers. Experience would tell me that getting it dead-on everywhere is liable to become a caper. That's why this PR only really aims to fix the immediate "this is just _way_ off" alignment issue, rather than fix it in all circumstances.

## Todo
Add changelog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,7 @@ We've made fixes to GOV.UK Frontend in the following pull requests:
 - [#4811: Use `KeyboardEvent.key` over deprecated `KeyboardEvent.keyCode` in the Tabs component](https://github.com/alphagov/govuk-frontend/pull/4811)
 - [#4812: Use `KeyboardEvent.key` over deprecated `KeyboardEvent.keyCode` in the Button component](https://github.com/alphagov/govuk-frontend/pull/4812)
 - [#4813: Remove deprecated `KeyboardEvent` properties from the Exit this Page component](https://github.com/alphagov/govuk-frontend/pull/4813)
+- [#4855: Fix mobile product name being misaligned in new type scale](https://github.com/alphagov/govuk-frontend/pull/4855)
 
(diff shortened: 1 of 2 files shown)


owenatgov (approved):
Great stuff. I had a go at trying to make this dynamic but it seems non-trivial enough to not be worth it right now. There probably is a way using the height of the text and the header but because this isn't at the baseline of the SVG itself but a little above to align with the text, it's a pain. This is probably the least complex solution.

PR #4872, opened 2024-03-18
Remove use of `innerHTML` from password input

querkmachine:
Removes the use of `innerHTML` from the Password input JavaScript, replacing it with `innerText`. 

- We want to avoid using `innerHTML` where possible as it presents a cross-site scripting (XSS) risk.
- i18n translation strings can only be provided as plain text anyway.
- The component's Nunjucks API already only allows plain text.

Does we think this warrants a changelog entry? It's a very minor change that undoubtedly makes little functional difference as the way `innerHTML` was being used meant it could never actually have HTML passed to it.


romaricpascal (approved):
⛵ ⛵ ⛵

Issue #4807, opened 2024-02-27
"conditional.html" value on radio item has whitespace inserted into it, breaking <textarea> elements


RichardBradley:
## Description of the issue

In our public facing app, we have some radio items which conditionally reveal a text area, for users to input their overseas address, only if they answer the radio question a certain way.

(We are aware that this isn't ideal from an accessibility point of view, and we have recently been unpacking some of these into separate pages, but I don't think that invalidates this bug.)

Since v5.2.0 / commit [6955db326a2e560c9862d29b28e39bd8f4020829](https://github.com/alphagov/govuk-frontend/commit/6955db326a2e560c9862d29b28e39bd8f4020829), whitespace is added to the html in here by govuk-frontend, which changes the value inside a textarea, corrupting user input. 

## Steps to reproduce the issue

Here is an extract from the njk template on our page that has caught this bug.
I have marked the key line with "HERE".
The `localisedCharacterCount` macro is our wrapper around https://design-system.service.gov.uk/components/character-count/ and contains a `<textarea>`

```
{% set bfpoAddressHtml %}
  {% call govukFieldset({
    classes: "govuk-!-width-two-thirds"
  }) %}
    {{ localisedCharacterCount({
      errorMessage: errors['bfpo-address'],
      name: "bfpo-address",
      id: "bfpo-address",
      classes: 'address-textarea',
      value: (form['bfpo-address'] or session.contactAddress.address) if bfpoInitiallySelected,
      label: {
        text: 'Address'
      },
      autocomplete: "street-address",
      hint: {
        text: 'Maximum 5 lines'
      },
      maxlength: 1200,
      threshold: 75,
      isWelshLocale: isWelshLocale
    }) }}

    {{ govukInput({
      label: {
        text: "BFPO"
      },
      id: "bfpo",
      name: "bfpo",
      classes: "govuk-input--width-10",
      value: (form['bfpo'] or session.contactAddress.postcode) if bfpoInitiallySelected,
      errorMessage: errors['bfpo']
    }) }}
  {% endcall %}
{% endset %}

...

{% block articleContent %}
  <form method="post" id="contact-address-form">
    {{ govukRadios({
      idPrefix: "type",
      name: "type",
      fieldset: {
        legend: {
          text: "Where should we write to you about your registration?",
          isPageHeading: true,
          classes: "govuk-fieldset__legend--xl"
        }
      },
      errorMessage: errors["type"],
      items: [
        {
          value: "uk",
[...]

PR #4899, opened 2024-03-26
Remove indents from conditional reveals in radios and checkboxes

owenatgov:
Removes `indent` from the rendering of `conditional.html` in both radios and checkboxes. Fixes https://github.com/alphagov/govuk-frontend/issues/4807

## Notes
There's a further feeling that `indent` might be a general risk to govuk-frontend and there's a question if we should simply remove _all_ instances of `indent`. This PR is the minimum required work to solve the problem in the linked issue but I think it's valuable to explore separately.

Are there any other places where textareas could be present? In theory every `html` param _could_ contain a textarea but that doesn't mean we should necessarily account for them in instances where a textarea shouldn't be eg: `button.html`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ We've made fixes to GOV.UK Frontend in the following pull requests:
 
 - [#4906: Update the icon in the warning text component to match the defined text colour and background colour, rather than always being white on black](https://github.com/alphagov/govuk-frontend/pull/4906)
 - [#4919: Use canvas colour for cookie banner over hardcoded grey](https://github.com/alphagov/govuk-frontend/pull/4919)
+- [#4899: Remove indents from conditional reveals in radios and checkboxes](https://github.com/alphagov/govuk-frontend/pull/4899)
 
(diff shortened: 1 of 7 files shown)


36degrees:
I think it might be worthwhile adding an example with a textarea for both components with a pre-defined `value`, and adding a test for it that fails without the removal of the `indent`?

Would hopefully stop some well-intended person re-introducing the issue in the future.

36degrees (approved):

nathdolan:
Hi guys, this is still an issue where you have a textarea within the conditional block of a Radios selection item (probably Checkboxes too I guess).  If I go into the Radios template and remove all the indent filters then that fixes the problem.  Needs fixing as a priority really as it is corrupting textarea values that have newlines (which obvs is pretty common in a textarea) where they're within conditionals.

36degrees:
@nathdolan thanks for flagging this – looks like this happens when a `fieldset` is provided (which is obviously pretty common with radios and checkboxes). This was previously flagged in #4807 but hadn't been picked up by our team.

I've raised #6232 which should handle this case.

PR #4910, opened 2024-03-28
Set appropriate value for `color-scheme` in dark mode

romaricpascal:
Setting [`color-scheme: dark`](https://developer.mozilla.org/en-US/docs/Web/CSS/color-scheme) swaps the native UI elements (like the file input) to their dark styles, enabling better contrast.

We'll probably want to dig into the support of that property inside webviews. This may vary depending of the OS versions app are meant to support.

For example on the file input:

| | `color-scheme: normal` |`color-scheme: dark`|
| --- | --- |---|
| Chrome | Screenshot of the file upload component showing light text on a light (default) background | Screenshot of the file uploadcomponent showing light text on a dark (default) background |
| Firefox | Screenshot of the file upload component showing dark (unstyled) text on a light (default) background | Screenshot of the file uploadcomponent showing light text on a dark (default) background | 
| Safari | Screenshot of the file upload component showing light text on a light (default) background | Screenshot of the file uploadcomponent showing light text on a dark (default) background|

### DIFF
--- a/packages/govuk-frontend/src/govuk/objects/_template.scss
+++ b/packages/govuk-frontend/src/govuk/objects/_template.scss
@@ -50,6 +50,8 @@
 
   @include govuk-media-query($media-type: screen) {
     .govuk-template--dark {
+      color-scheme: dark;
+
       --govuk-canvas-background-colour: #333;
       --govuk-body-background-colour: #222;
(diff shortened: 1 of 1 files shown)


owenatgov (approved):
Happy with it!

PR #4917, opened 2024-04-02
[WIP][SPIKE] Experiment with modifying header component with new top-level elements

querkmachine:
Spike of https://github.com/alphagov/govuk-design-system/issues/3714. Still a work in progress but wanted to PR it for visibility and so I could write out thoughts. 

**[Look at the thing.](https://govuk-frontend-pr-4917.herokuapp.com/components/header)**

This PR aims to update the header component to make the following changes:

- Remove the navigation.
- Add the ability to show account management options.
- Add the ability to show language selection options.
- Add the ability to show a search field. 

It must continue to show the GOV.UK logo and product or service name. 

> [!NOTE]
> This is an experiment and is not production-ready code. Many elements are hardcoded for the purposes of the spike. Some hacky CSS overrides are used, as well as components utilised in vaguely non-standard ways. Tests and documentation will not be written for it.

## Changes

- Basically completely refactors the thing. 

## Todo 
- Toolbox dropdowns are jank on browsers that don't support `details`/`summary`, like IE 11. 
- This currently uses [the `name` attribute on `details` elements](https://caniuse.com/mdn-html_elements_details_name) to ensure that only one toolbox is open at once. This is a relatively new feature of the HTML spec and is not yet supported by Firefox, nor likely the older versions of our other supported browsers.

### General thoughts
I've split the header into two rows. The top row shows the GOV.UK logo and service or product name. The bottom row shows the search and what I have nicknamed 'toolboxes', little (intended to be) dropdowns that contain additional links inside of them. The thought behind this separation is that 'top row' elements will virtually always be present, whereas the bottom row could be entirely removed if unneeded. 

I wonder how suitable it is to have the search function in the header. In my mind, search is hierarchically on the same level as primary navigation, however we've already decided to demote primary nav to outside of the header. Personally, it feels a little weird for search to still be here. 

### Product and service names
The service name has been moved to the same position as the product name. To differentiate them, I've swapped their text styling, so that service names are more visually distinct compared to the 'GOV.UK' logo, whereas product names are more similar.

Product names are now separate from the GOV.UK logo link. This would allow us to potentially enforce the GOV.UK logo always linking to the GOV.UK homepage, rather than the current different approaches depending on whether it's a service or product.

### Toolboxes 
I'm wondering how best to do the 'toolbox' links. The megamenu approach that GOV.UK uses doesn't feel like it makes sense here, as a toolbox can contain as few as two links in some scenarios. GOV.UK depends on having dedicated pages to fall back to if JavaScript is unavailable, which we cannot guarantee services will want or be able to do. Any solution here must thus degrade gracefully.

As the toolboxes currently use `details`/`summary` elements to work, they will look totally jank on IE 11.

An approach that I've seen before (on GitHub, in fact) is to style the content of `details` like a modal. This could be one method of resolving exactly how these 'dropdowns' should work and avoids the pains that shifting alignment and space can cause across viewport sizes, but it would open the whole can of modal worms again.

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/header/_index.scss
+++ b/packages/govuk-frontend/src/govuk/components/header/_index.scss
@@ -127,7 +127,7 @@
 
     @include govuk-media-query($from: tablet) {
       margin-top: 0;
-      padding-left: govuk-spacing(2);
+      padding-left: govuk-spacing(4);
       margin-left: govuk-spacing(2);
       border-left: 1px solid $govuk-header-separator-color;
     }
--- a/packages/govuk-frontend/src/govuk/components/header/_index.scss
+++ b/packages/govuk-frontend/src/govuk/components/header/_index.scss
@@ -4,7 +4,7 @@
   $govuk-header-border-color: $govuk-brand-colour;
   $govuk-header-border-width: govuk-spacing(2);
   $govuk-header-link-underline-thickness: 3px;
-  $govuk-header-separator-color: #ffffff;
+  $govuk-header-separator-color: #2e3133;
   $govuk-header-toolbox-arrow-size: 12px;
 
   .govuk-header {
@@ -94,6 +94,28 @@
     }
   }
 
+  // TEMPORARY HACKY CODE TO POSITION ONE LOGIN ICON
+  .govuk-header__one-login-icon {
+    margin-top: -1px;
+    margin-left: govuk-spacing(1);
+    vertical-align: text-bottom;
+  }
+
+  // One Login icon focus code
+  .govuk-header__link--one-login:focus .govuk-header__one-login-icon {
+    .background {
+      fill: govuk-colour("black");
+    }
+
+    .border {
+      stroke: govuk-colour("black");
+    }
+
+    .person {
+      fill: govuk-colour("white");
+    }
+  }
+
   // Logo container
   .govuk-header__logo {
     display: block;
@@ -127,8 +149,8 @@
 
     @include govuk-media-query($from: tablet) {
       margin-top: 0;
-      padding-left: govuk-spacing(4);
       margin-left: govuk-spacing(2);
+      padding-left: govuk-spacing(4);
       border-left: 1px solid $govuk-header-separator-color;
     }
   }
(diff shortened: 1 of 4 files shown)


mia-allers-gds:
This is fantastic @querkmachine !
Some thoughts on what you've done:

**Logo and product/ service name**
- I think we need to consider both the size of the product/ service name and also how it would wrap onto multiple lines if it's longer than one or two words
- I think we may also need to differentiate between the logo and the domain (product/ service) name more, as they are technically two links. I've used a pipe between them on campaigns, which you can see here: [Help for household campaign website](https://helpforhouseholds.campaign.gov.uk/)

**Search**
- I think we could potentially discuss where to put search (primary header or nav), personally it feels 'domain level' to me but open to ideas
- The search is very prominent and I'm wondering if we could also make it a dropdown? I'm not sure how well the search function works on most of our domains, or how necessary it often is, and its taking up a lot of prime real estate 
- Makes sense for the 'toolboxes' to be taken out if not needed!

querkmachine:
> * I think we need to consider both the size of the product/ service name and also how it would wrap onto multiple lines if it's longer than one or two words

At the moment the entire product/service name wraps to appear below the logo, which I agree isn't the prettiest of things. I'll see about changing this so that the name wraps onto multiple lines but they remain to the right of the logo, more like how the current header component does it. 

> * I think we may also need to differentiate between the logo and the domain (product/ service) name more, as they are technically two links. I've used a pipe between them on campaigns, which you can see here: [Help for household campaign website](https://helpforhouseholds.campaign.gov.uk/)

I've added a separator line between them.

How are we feeling about the current colour and weight of the separators? The appearance is taken directly from the current header, where it was being used to separate navigation items on mobile. It's not particularly high contrast, but I'm not sure this is a context where contrast is too important when compared to spacing. 

> * I think we could potentially discuss where to put search (primary header or nav), personally it feels 'domain level' to me but open to ideas
> * The search is very prominent and I'm wondering if we could also make it a dropdown? I'm not sure how well the search function works on most of our domains, or how necessary it often is, and its taking up a lot of prime real estate

My thoughts on the search are roughly:

- Search forms are very service-specific, insomuch as each service will likely implement them in radically different ways. For example, requiring a user to select a search type from a list before a search can be made, or the search being entirely JS-powered with results in their own dropdown, as on the Design System website. I'm not too confident we can build a 'one size fits all' search form that fits neatly into the header component—at least not how we typically would provide something like this, it feels like there's too many possible variations. 
- Regardless, if we're intending to build this simplified input + button combo, in my mind it makes sense to build it as a separate component that doesn't rely on the header as there is already some demand for it. https://github.com/alphagov/govuk-design-system-backlog/issues/132
- Making search a toolbox item could be difficult. Toolboxes are currently quite simple (heading text, arrow icon, list of links), and there's the possibility of letting users add their own toolboxes following this pattern. A search form within a toolbox would need to exist as an exception to those rules.
- Relatedly, having the search field hide itself behind a dropdown upon loading a new page might be frustrating if a user is using it repeatedly to refine their query.

querkmachine:
Closing this spike as it's now served its purpose.

PR #4918, opened 2024-04-02
Fix browser grade listing in README

peteryates:
The angle brackets in the script example snippet break markdown rendering. Wrapping them in backticks solves it.

Also continue listing the 'main' desktop browsers within the brackets as it stops rather abruptly.


36degrees (approved):
Great spot, thanks! This just needs another review from the team and then we can get this merged.

domoscargin (approved):
Nice catch!

PR #4919, opened 2024-04-03
Use canvas colour for cookie banner over hardcoded grey

36degrees:
The cookie banner is hardcoded to use light-grey as a background colour. Wherever possible, we should use the ‘applied’ colours rather than specific colours from the palette.

We think (although this is somewhat post-rationalising) that the cookie banner uses light-grey for the same reason the footer does. Light grey is the 'canvas' colour used for the `<html>` element and everything outside of the <body>, including any viewport after the footer and the overscroll area. As the cookie banner appears above the header, it's conceptually outside of the body of the page.

Update the cookie banner to reference the canvas background colour rather than the hardcoded light-grey.


owenatgov (approved):
I think these test failures are coming from github having a wobble but the code looks dandy to me

PR #4922, opened 2024-04-04
[WIP][SPIKE]: Address component

ttbarnes:
Initial setup for a new address component.

Keen to see if there is interest for this component before continuing - any help or feedback is greatly appreciated.

## Why?

Currently, the only way to render address fields is to render the text input component several times. I.e one text input for each address field.

Instead, it would be very useful to have a single component that:
- Automatically renders address field text inputs, ensuring a consistent pattern.
- Consumes an object of values that pre-populate all text input values.
- Consumes an object of errors that pre-populate all text input error messages.
- Has the ability to consume custom field IDs for each field.
  - For example, data wise, one service might have a "Line 1" field named as `line1`, whereas in another service, it could be e.g `addressLine1`.
- Has the ability to render "line 1" and "line 2", with only "line 1" having a visible label. E.g "Building and street" could have 2 lines.

This should therefore handle the recommended [address pattern](https://design-system.service.gov.uk/patterns/addresses/#multiple-text-inputs) with a single component, whilst also reducing lines of code in individual services.

## Example usage

```
{% from "govuk/components/address/macro.njk" import govukAddress %}

{{ govukAddress({
  legend: {
    text: "What is your address?"
  },
  line1: {
    id: "line1"
  },
  line2: {
    id: "line2"
  },
  townOrCity: {
    id: "townOrCity"
  },
  county: {
    id: "county"
  },
  postcode: {
    id: "postcode"
  },
  values: {
    line1: "Line 1!",
    county: "London"
  },
  errorMessages: {
    line1: {
      text: "Address line 1 cannot contain special characters"
    },
    townOrCity: {
      text: "Enter town or city"
    },
    postcode: {
      text: "Enter postcode"
    }
  }
}) }}
```

### Rendering

Screenshot 2024-04-04 at 16 58 27

### Alternative rendering
Screenshot 2024-04-04 at 16 58 54

## Rendering with error messages and pre-populated values

Screenshot 2024-04-04 at 16 59 04

## TODO
- Unit tests (having some trouble understanding how to test with the nunjucks test config, doesn't seem to render anything past the 1st DOM level).
- Review examples.
- Determine if any other updates are required.

## Additional information
This component idea stemmed from [UKEF - Export Insurance Policies](https://github.com/UK-Export-Finance/exip/), where we were rendering several address components. We ended up creating our own address component that handles the above functionality; Instead of having to render 5x text inputs and various conditions in individual forms.

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/address/README.md
+++ b/packages/govuk-frontend/src/govuk/components/address/README.md
@@ -1,0 +1,15 @@
+# Address
+
+## Installation
+
+See the [main README quick start guide](https://github.com/alphagov/govuk-frontend#quick-start) for how to install this component.
+
+## Guidance and Examples
+
+Find out when to use the address component in your service in the [GOV.UK Design System](https://design-system.service.gov.uk/components/address).
+
+## Component options
+
+Use options to customise the appearance, content and behaviour of a component when using a macro, for example, changing the text.
+
+See [options table](https://design-system.service.gov.uk/components/address/#options-address-example) for details.
--- a/packages/govuk-frontend/src/govuk/components/address/address.yaml
+++ b/packages/govuk-frontend/src/govuk/components/address/address.yaml
@@ -1,0 +1,243 @@
+params:
+  - name: legend
+    type: object
+    required: true
+    description: The legend for the address component.
+    params:
+      - name: text
+        type: string
+        required: true
+        description: If `html` is set, this is not required. Text to use within the legend. If `html` is provided, the `text` option will be ignored.
+      - name: html
+        type: string
+        required: true
+        description: If `text` is set, this is not required. HTML to use within the legend. If `html` is provided, the `text` option will be ignored.
+      - name: classes
+        type: string
+        required: false
+        description: Classes to add to the legend.
+      - name: isPageHeading
+        type: boolean
+        required: false
+        description: Whether the legend also acts as the heading for the page.
+  - name: line1
+    type: object
+    required: true
+    description: Line 1 of the address
+    params:
+      - name: id
+        type: string
+        required: true
+        description: The ID of the address line 1 field.
+      - name: label
+        type: object
+        required: false
+        description: The label used by address line 1. Defaults to "Line 1".
+        isComponent: true
+  - name: line2
+    type: object
+    required: true
+    description: Line 2 of the address
+    params:
+      - name: id
+        type: string
+        required: true
+        description: The ID of the address line 2 field.
+      - name: label
+        type: object
+        required: false
+        description: The label used by address line 2. Defaults to "Line 2".
+        isComponent: true
+      - name: visuallyHiddenText
+        type: boolean
+        required: false
+        description: Optional field to hide the label of line 2 and use hidden text.
+  - name: townOrCity
+    type: object
+    required: true
+    description: Town/city of the address
+    params:
+      - name: id
+        type: string
+        required: true
+        description: The ID of town/city field.
+      - name: label
+        type: object
+        required: false
+        description: The label used by town/city. Defaults to "Town or city".
+        isComponent: true
+  - name: county
+    type: object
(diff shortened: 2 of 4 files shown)


querkmachine:
Some random, disjointed, personal thoughts about this:

I think the main reason we don't have any singular component for this is because the format of addresses is too variable. Although the format used here works for most UK addresses, it isn't necessarily suitable for overseas addresses or those from other countries (including Ireland).

As such, what formats and validations that need carried out on addresses can be heavily dependent on the audience for the service. We might risk codifying one technique as being 'the correct way' in people's minds by making it a component, when really there is no single 'correct' choice.

Even within UK addresses there is some variation as to what might be the best way of gathering this information. 'County', for example, hasn't been needed for a long time now, and a team may want to omit it. Strictly speaking, we don't need a town or city either if we ask for the postcode. 

Having a dedicated address component may also open a Pandora's box of why we don't also provide components for postcode lookup, card or bank account details, or other repeatable form design patterns.

ttbarnes:
> Some random, disjointed, personal thoughts about this:
> 
> I think the main reason we don't have any singular component for this is because the format of addresses is too variable. Although the format used here works for most UK addresses, it isn't necessarily suitable for overseas addresses or those from other countries (including Ireland).
> 
> As such, what formats and validations that need carried out on addresses can be heavily dependent on the audience for the service. We might risk codifying one technique as being 'the correct way' in people's minds by making it a component, when really there is no single 'correct' choice.
> 
> Even within UK addresses there is some variation as to what might be the best way of gathering this information. 'County', for example, hasn't been needed for a long time now, and a team may want to omit it. Strictly speaking, we don't need a town or city either if we ask for the postcode.
> 
> Having a dedicated address component may also open a Pandora's box of why we don't also provide components for postcode lookup, card or bank account details, or other repeatable form design patterns.

Thank you, all understood and makes sense. What would your thoughts be on a UK Address component with all fields being optional?

For example, address fields could only render if they are passed to the component. The following example would only render 2x inputs:

```
{{ govukAddressUk({
  legend: {
    text: "What is your address?"
  },
  line1: {
    id: "line1"
  },
  townOrCity: {
    id: "townOrCity"
  },
  values: {
    line1: "Line 1!",
    townOrCity: "London"
  },
  errorMessages: {
    line1: {
      text: "Address line 1 cannot contain special characters"
    },
    townOrCity: {
      text: "Enter town or city"
    },
  }
}) }}
```

I appreciate there are many variants and this could potentially open a pandora's box, however this could allow for multiple configurations for a UK address - in theory, I imagine, all combinations. For example, if only "line 1" is desired, it would be supported in the component by only passing `line1`.

owenatgov:
@ttbarnes Just checking if it's still useful for you to have this component open as a reference point or if we can close it for now?

For transparency, it doesn't look like this is something we'll be able to work on, at least not in the immediate future. We would be interested in you writing up any findings you had on our [community backlog](https://github.com/alphagov/govuk-design-system-backlog/issues). I couldn't find an issue for an addresses component, although there is one for an [addresses **pattern**](https://github.com/alphagov/govuk-design-system-backlog/issues/31), so I'm not sure if you've written this up already. If you don't have the time to, we can try to summarise something.

ttbarnes:
> @ttbarnes Just checking if it's still useful for you to have this component open as a reference point or if we can close it for now?
> 
> For transparency, it doesn't look like this is something we'll be able to work on, at least not in the immediate future. We would be interested in you writing up any findings you had on our [community backlog](https://github.com/alphagov/govuk-design-system-backlog/issues). I couldn't find an issue for an addresses component, although there is one for an [addresses **pattern**](https://github.com/alphagov/govuk-design-system-backlog/issues/31), so I'm not sure if you've written this up already. If you don't have the time to, we can try to summarise something.

Thank you @owenatgov - happy to close.

It doesn't seem wise to create such a component, from the previous comment, as there are too many variants.

PR #4908, opened 2024-03-28
[SPIKE] Navigation - create navigation component

domoscargin:
This PR adds a navigation component to govuk-frontend, wholesale copied from the One Login header, then tweaked slightly.

It's a foundation to investigate https://github.com/alphagov/govuk-design-system/issues/3715

See the thing here: https://govuk-frontend-pr-4908.herokuapp.com/components/navigation

PR #4923, opened 2024-04-04
[WIP][SPIKE] Combine header redesign and primary navigation spikes 

querkmachine:
Merges together the content of the #4908 and #4917 spikes.

## Changes
- The header component now inherits and uses the navigation component, using the same navigation parameters of the existing Design System header. 
- Updates the navigation macro to use the same parameter names that the Design System header used.


querkmachine:
Closing this spike as it's now served its purpose.

PR #4928, opened 2024-04-08
New idsk file structure

Sobanim:
New idsk file structure.
Removed all unnecessary components. Changed paths to files to make the project start


Sobanim:
Error. Created a PR in the wrong place

PR #4929, opened 2024-04-08
Update Sass GitHub action to save built files

romaricpascal:
This will allow inspecting the output should an older version of Sass fail. 

Each job gets the output of the Sass compilation [uploaded as artifact](https://github.com/actions/upload-artifact), which we can then download for investigation from the Summary page of our GitHub Sass workflow (accessible by clicking 'Details' next to any of the Sass checks and then 'Summary').

The artifact upload is set up as a final step for each job, so that checking there's no more Sass variables in the output runs first. It is set up with `if: ${{ !cancelled() }}` so that it runs even if an error happens beforehand (but [respects the job having been cancelled, unlike `always()`](https://docs.github.com/en/actions/learn-github-actions/expressions#always)).

### DIFF
--- a/.github/workflows/sass.yaml
+++ b/.github/workflows/sass.yaml
@@ -45,6 +45,14 @@ jobs:
         run: |
           ! grep "\$govuk-" .tmp/index.css
 
+      - name: Save compiled Sass
+        uses: actions/upload-artifact@v4.3.1
+        if: ${{ !cancelled() }}
+        with:
+          name: Dart Sass v1.0.0 output
+          path: .tmp/index.css
+          if-no-files-found: ignore
+
   dart-sass-latest:
     name: Dart Sass v1 (latest)
     runs-on: ubuntu-22.04
@@ -80,7 +88,15 @@ jobs:
         run: |
           ! grep "\$govuk-" .tmp/index.css
 
-  # Node Sass v5.0.0 = LibSass v3.5.5
+      - name: Save compiled Sass
+        uses: actions/upload-artifact@v4.3.1
+        if: ${{ !cancelled() }}
+        with:
+          name: Dart Sass v1 (latest) output
+          path: .tmp/index.css
+          if-no-files-found: ignore
+
+  # Node Sass v3.4.0 = LibSass v3.3.0
   lib-sass:
     name: LibSass v3.5.5 (deprecated)
     runs-on: ubuntu-22.04
@@ -114,6 +130,14 @@ jobs:
         run: |
           ! grep "\$govuk-" .tmp/index.css
 
+      - name: Save compiled Sass
+        uses: actions/upload-artifact@v4.3.1
+        if: ${{ !cancelled() }}
+        with:
+          name: LibSass v3.5.5 (deprecated) output
+          path: .tmp/index.css
+          if-no-files-found: ignore
+
   # Node Sass v8.x = LibSass v3 latest
   lib-sass-latest:
     name: LibSass v3 (latest, deprecated)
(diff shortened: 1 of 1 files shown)


.github/workflows/sass.yaml
```diff
@@ -45,6 +45,14 @@ jobs:
         run: |
           ! grep "\$govuk-" .tmp/all.css
 
+      - name: Save compiled Sass
+        uses: actions/upload-artifact@v4.3.1
+        if: ${{ !cancelled() }}
+        with:
+          name: Dart Sass v1.0.0 output
```
romaricpascal, line 52: **nitpick** Would have wanted to re-use the name of the job by looking in one of the [contexts](https://docs.github.com/en/actions/learn-github-actions/contexts) provided to the step... but couldn't find anything. If anyone knows how, would be keen to update. 

Don't think that's a blocker, though, more of a nice-to-have-if-we-know-how.

.github/workflows/sass.yaml
```diff
@@ -45,6 +45,14 @@ jobs:
         run: |
           ! grep "\$govuk-" .tmp/all.css
 
+      - name: Save compiled Sass
+        uses: actions/upload-artifact@v4.3.1
+        if: ${{ !cancelled() }}
```
romaricpascal, line 50: **nitpick** Thinking we could add a `continue-on-error` to avoid that step failing the job if the upload has an issue. This `continue-on-error` would have to be conditionned to the previous step being a success, though. Otherwise, I think it'd make test failures not fail the job.

.github/workflows/sass.yaml
36degrees: Uploading the artefact is also very repetitive – we briefly explored running these using a matrix in #3030 but at the time it felt like it made the code too complicated. Wondering if that's worth revisiting.

Equally with us dropping support for Ruby Sass and LibSass 'soon' maybe it's not worth spending any more time on this? 😕

.github/workflows/sass.yaml
36degrees: Why do we need to guard against the job being cancelled?

romaricpascal: Following GitHub's recommendations to prefer `!cancelled()` to `always()` to ensure a step runs regardless of the previous steps. Makes sense to me as if the job was cancelled, we don't care about uploading artifacts anyways. Happy to drop a comment on the first one to remind of [GitHub's recommendations](https://docs.github.com/en/actions/learn-github-actions/expressions#always).

.github/workflows/sass.yaml
romaricpascal: Yeah, thought there was a fair amount of repetitions and remembered Colin's attempt at having a matrix. Still not convinced we need to DRY things up with a matrix there given it's not that costly to add something 6 times and we don't touch those that often. Especially with the change of Sass compiler support at some point in the future.

If for whatever reason we start bulking these workflows with more shared steps, we could revisit then.

36degrees (approved):

.github/workflows/sass.yaml
```diff
@@ -114,6 +130,14 @@ jobs:
         run: |
           ! grep "\$govuk-" .tmp/index.css
 
+      - name: Save compiled Sass
+        uses: actions/upload-artifact@v4.3.1
+        if: ${{ !cancelled() }}
+        with:
+          name: LibSass v3.3.0 (deprecated) output
```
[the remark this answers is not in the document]
romaricpascal: It's even 5.0.0 after https://github.com/alphagov/govuk-frontend/pull/5523

PR #4932, opened 2024-04-11
Allow values for the default date input

### LINKED ISSUE #1154: Pass values to date-input when using default items
At the moment there is no way to pass values or field-level error states (CSS classes) to the date-input when using the default item set. This means you have to redefine the items param, including CSS classes for widths, whenever it's used. This encourages repeated code and is arguably more prone to error.

Is it worth having a values param and an errors param on date-input? This could fall back to values/errors passed via the `items` param for backward compatibility, but favouring the new `items`/`errors` props.

e.g.
```nunjucks
{{ govukDateInput({
  id: "dob",
  namePrefix: "dob",
  values: {
    day: 4,
    month: 5,
    year: 20t19
  },
  errorMessage: 'Please enter a valid year',
  errors: {
    year: true
  }
}) }}
```

I don't have an immediate need for this but it's something I noticed and thought worth suggesting in case others agree.

joelanman:
There are two ways to use `govukDateInput`, by passing `items` or leaving that blank and getting the default Day, Month and Year that most services need.

However you nearly always need to set the values, at which point you can't use the default mode, you have to set all the items individually.

This PR adds a `values` option similar to the checkboxes component, so you can keep using the default mode. This is particularly useful in prototyping, where the current approach is much more complex for non-developers.

In usage this looks like:

```
{{ govukDateInput({
  id: "passport-issued",
  namePrefix: "passport-issued",
  fieldset: {
    legend: {
      text: "When was your passport issued?",
      isPageHeading: true,
      classes: "govuk-fieldset__legend--l"
    }
  },
  hint: {
    text: "For example, 27 3 2007"
  },
  values: [27, 3, 2007]
}) }}
```

compared to the existing approach:

```
{{ govukDateInput({
  id: "passport-issued",
  namePrefix: "passport-issued",
  fieldset: {
    legend: {
      text: "When was your passport issued?",
      isPageHeading: true,
      classes: "govuk-fieldset__legend--l"
    }
  },
  hint: {
    text: "For example, 27 3 2007"
  },
  items: [
    {
      classes: "govuk-input--width-2",
      name: "day",
      value: "27"
    },
    {
      classes: "govuk-input--width-2",
      name: "month",
      value: "3"
    },
    {
      classes: "govuk-input--width-4",
      name: "year",
      value: "2007"
    }
  ]
}) }}
```

It also supports errors:

```
{{ govukDateInput({
  id: "passport-issued",
  namePrefix: "passport-issued",
  fieldset: {
    legend: {
      text: "When was your passport issued?",
      isPageHeading: true,
      classes: "govuk-fieldset__legend--l"
    }
  },
  hint: {
    text: "For example, 27 3 2007"
  },
  errorMessage: {
    text: "The date your passport was issued must be in the past"
  },
  values: [27, 3, 2047],
  errors: [true, true, true]
}) }}
```

compared to

```
{{ govukDateInput({
  id: "passport-issued",
  namePrefix: "passport-issued",
  fieldset: {
    legend: {
      text: "When was your passport issued?",
      isPageHeading: true,
      classes: "govuk-fieldset__legend--l"
    }
  },
  hint: {
    text: "For example, 27 3 2007"
  },
  errorMessage: {
    text: "The date your passport was issued must be in the past"
  },
  items: [
    {
      classes: "govuk-input--width-2 govuk-input--error",
      name: "day",
      value: "27"
    },
    {
      classes: "govuk-input--width-2 govuk-input--error",
      name: "month",
      value: "3"
    },
    {
      classes: "govuk-input--width-4 govuk-input--error",
      name: "year",
      value: "2076"
    }
  ]
}) }}
```

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/date-input/template.njk
+++ b/packages/govuk-frontend/src/govuk/components/date-input/template.njk
@@ -15,19 +15,22 @@
   {% set dateInputItems = params.items %}
 {% else %}
   {% set dateInputItems = [
-    {
-      name: "day",
-      classes: "govuk-input--width-2"
-    },
-    {
-      name: "month",
-      classes: "govuk-input--width-2"
-    },
-    {
-      name: "year",
-      classes: "govuk-input--width-4"
-    }
-  ] %}
+  {
+    name: "day",
+    classes: "govuk-input--width-2" + (" govuk-input--error" if params.errors[0] else ""),
+    value: params.values[0]
+  },
+  {
+    name: "month",
+    classes: "govuk-input--width-2" + (" govuk-input--error" if params.errors[1] else ""),
+    value: params.values[1]
+  },
+  {
+    name: "year",
+    classes: "govuk-input--width-4" + (" govuk-input--error" if params.errors[2] else ""),
+    value: params.values[2]
+  }
+] %}
 {% endif %}
 
 {#- Capture the HTML so we can optionally nest it in a fieldset -#}


joelanman:
thinking some more, if the main value here is for prototyping, it might be better to look into an option to automatically set all `value`s based on `name`s in the kit, I think @36degrees did an investigation into that a while back

joelanman:
you could add another property to handle errors for the default case:

```
{{ govukDateInput({
    values: [data['passport-issued-day'],
             data['passport-issued-month'],
             data['passport-issued-year']],
    errors: [false, false, true]
...
}) }}
```

frankieroberto:
This looks good, and would resolve #1154.

I wonder if the values key would be slightly more intuitive as an object, eg:

{{ govukDateInput({
    values: {
      day: data['passport-issued-day',
      data['passport-issued-month'],
      data['passport-issued-year']
    },
}) }}  
```             

That would save people having to know which order it's in, and would also allow you to just set one or two of the date part values if needed (although this seems less likely).

The key could also be `value` for consistency with checkboxes and radios, where the key is singular even though multiple values can be set (in the case of checkboxes). Although I think having it plural as `values` is more intuitive.

**Comment by joelanman** [comment]
I've update to latest Frontend and added the `errors` params

**Comment by frankieroberto** [comment]
@joelanman Thinking about this some more, having the `values` key accept an object should make it even easier to set all 3 date parts at once using the prototype kit, as you could do this:

```nunjucks
{{ govukDateInput({
   namePrefix: "passportIssued",
    values: {
      day: data.passportIssued
    },
...    
}) }}
```

The only other change needed would be to change the default for the `name` attribute on each of the fields to be `{{params.namePrefix}}[day]` instead of `{{params.namePrefix}}-day`.

Feels like that’d be worthwhile?

joelanman:
It is simpler and I can see how `{{params.namePrefix}}[day]` would be beneficial for prototyping, but it would be a breaking change, and not follow the naming conventions in Frontend so I don't think it would be accepted

frankieroberto:
> @frankieroberto Receiving a `{day:..., month:..., year:...}` object and updating the field names to use brackets for the name feel better suited for their own piece of work. The latter especially will require a thorough review of the impact of updating the default names, as not all backends will parse the brackets in submitted names (while a common convention, it requires [enabling the `extended` option for express' body-parser](https://www.npmjs.com/package/body-parser#options-3), I _think_). It's not necessarily a breaking change, though, as we could provide the new name formatting through a `nameFormat: 'brackets'` option to trigger the bracket notation, leaving the default behaviour as it is now.

I _think_ the bracket syntax has always worked in the GOV.UK prototype kit, at least for as long as I can remember. The `body-parser` documentation seems like it defaults to `true`?

@romaricpascal The change in the default name could come later, and still use the object format for values, which would mean having to do something like this initially, which isn't much more code over an array:

```
{{ govukDateInput({
    values: {
      day: data['passport-issued-day'],
      month: data['passport-issued-month'],
      year: data['passport-issued-year']
    },
 ...
}) }} 
```

Alternatively, I guess it could make sense to support _both_ an array and an object? (Although possibly that’s over-complicated?)

One other advantage of accepting an object is that this could then set the values even if you used a custom `items` array (eg just month and year), which @edwardhorsford [has suggested](https://github.com/nhsuk/nhsuk-frontend/pull/994#discussion_r1721676440) on our version of this PR for NHS frontend.

romaricpascal:
> I think the bracket syntax has always worked in the GOV.UK prototype kit, at least for as long as I can remember. The body-parser documentation seems like it defaults to true?

Oh, good spot, had missed that `true` was the default for that option of body-parser in that last sentence. Changing the names wholesale would be a breaking change in any case, so using an option for the macro could speed up its release without waiting for the next major release of GOV.UK Frontend.

> One other advantage of accepting an object is that this could then set the values even if you used a custom items array (eg just month and year), which @edwardhorsford https://github.com/nhsuk/nhsuk-frontend/pull/994#discussion_r1721676440 on our version of this PR for NHS frontend.

I believe that @edwardhorsford's suggestion could also work for arrays, [using `loop.index0`](https://mozilla.github.io/nunjucks/templating.html#for) to pull the value at the index of the item that's being rendered.

> Alternatively, I guess it could make sense to support both an array and an object? (Although possibly that’s over-complicated?)

Untested (and unsure of the syntax), like Ed's suggestion, but I guess it wouldn't be too terrible to look at both array and object access when rendering the item with something like:

```
item.value or ((values[loop.index0] or values[item.name]) if values)
```

That'd require some thorough testing around false-y values to ensure that an unexpected 0 (submitted by mistake, for example) gets handled correctly.

joelanman:
It's still true that "not all backends will parse the brackets in submitted names" - the kit is only one place Frontend is used, other people could be using the components in other backends and frameworks

romaricpascal:
> It's still true that "not all backends will parse the brackets in submitted names" - the kit is only one place Frontend is used, other people could be using the components in other backends and frameworks

Yup, we'd certainly want a review of what Node backends do in any case before changing the name of the field (Ports being free to adapted the naming of the fields to their language conventions, like [what govuk-form-builder does](https://github.com/x-govuk/govuk-form-builder/blob/main/lib/govuk_design_system_formbuilder/elements/date.rb#L143-L150)).

frankieroberto:
I agree it should be considered a breaking change, even if we don't know of any services it’d affect.

Using a option for the macro seems like a nice way of avoiding it being a breaking change, and then it could always become the default in the next breaking change release?

36degrees:
Closed as this has been resolved by #6971.

PR #4935, opened 2024-04-12
Fix password input button unexpectedly stretching

### LINKED ISSUE #4934: When using a width modifier on the password input component, it makes the show/hide button wide (fill out space)
## Description of the issue

When using a width modifier on the password input component, it makes the show/hide button wide (fill out space).
I'm not sure if this is really a bug but I thought it was worth bringing to your attention as I was not expecting this behaviour.

In the below example I've added the `govuk-!-width-one-half` width modifier to the password input.

Password input component with a govuk-!-width-one-half width modifier on the input

```HTML
<div class="govuk-form-group govuk-password-input" data-module="govuk-password-input">
  <label class="govuk-label" for="password-input">
    Password
  </label>
  <div class="govuk-input__wrapper govuk-password-input__wrapper">
    <input class="govuk-input govuk-password-input__input govuk-js-password-input-input  govuk-!-width-one-half" id="password-input" name="password" type="password" spellcheck="false" autocomplete="current-password" autocapitalize="none">
    <button type="button" class="govuk-button govuk-button--secondary govuk-password-input__toggle govuk-js-password-input-toggle" data-module="govuk-button" aria-controls="password-input" aria-label="Show password" hidden>
      Show
    </button>
  </div>
</div>
```

I would expect the show/hide button size to be unaffected by the width of the input.

- Browser: Chrome
- Browser version: 123.0.6312.107
- GOV.UK Frontend Version: 5.3.0

querkmachine:
Fixes issue where the Password input toggle button unexpectedly stretched to fill any space not used by the text input, such as if the input has a restricted maximum width.

I think the `flex-grow` property may have been a leftover from earlier in the component's development, as it doesn't appear to serve any function in the current version of the component.

Fixes #4934. 

## Changes
- Removed the `flex-grow` property from the Password input toggle button.
- Added a review app example of a restricted width password input to help keep this in check in future.
- Tweaked wording of a few comments/documentation pieces.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ We've made fixes to GOV.UK Frontend in the following pull requests:
 - [#4906: Update the icon in the warning text component to match the defined text colour and background colour, rather than always being white on black](https://github.com/alphagov/govuk-frontend/pull/4906)
 - [#4919: Use canvas colour for cookie banner over hardcoded grey](https://github.com/alphagov/govuk-frontend/pull/4919)
 - [#4899: Remove indents from conditional reveals in radios and checkboxes](https://github.com/alphagov/govuk-frontend/pull/4899)
+- [#4935: Fix password input button unexpectedly stretching](https://github.com/alphagov/govuk-frontend/pull/4935)
 
 ## GOV.UK Frontend v5.3.0 (Feature release)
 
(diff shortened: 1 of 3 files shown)


owenatgov (approved):
Easy peasy

PR #4936, opened 2024-04-15
Fix skip link underline being removed when global styles are enabled 

### LINKED ISSUE #4930: Global styles conflict with skip link component
## Description of the issue

When `$govuk-global-styles` is set to true in Sass, the underline is missing from skip links.

`$govuk-global-styles` applies [govuk-focus-text](https://github.com/alphagov/govuk-frontend/blob/bd465bda9bea6b46aee8ac8f15b607a3b51c2b3b/packages/govuk-frontend/src/govuk/helpers/_focused.scss#L25) to all links, which removes underline and adds `box-shadow`.

But the skip link component [removes box shadow](https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/src/govuk/components/skip-link/_index.scss#L30) in this case, without adding back in the underline.

We are using global styles because our service contains some content that is rendered from markdown, and we cannot easily apply the proper classes to the generated HTML.

To get the underline I need to manually add some extra CSS.

- Compile the GOV.UK frontend Sass with $govuk-global-styles set to true
- View a page with a skip link
- Navigate to the skip link by keyboard

Actual - no underline
Screenshot 2024-04-10 at 10 24 49

Expected
Screenshot 2024-04-10 at 10 23 31

- Operating system: MacOS
- Browser: Chrome
- Browser version: 123.0.6312.59
- GOV.UK Frontend Version: 5.1.0

querkmachine:
Fixes #4930.

## Changes
- Adds the `govuk-link-decoration` mixin (which sets the `text-decoration`, `text-decoration-thickness` and `text-decoration-offset` properties) to the list of styles re-added to the Skip link component when global styles are enabled.
- Update changelog.

## Thoughts
- Simply re-adding the underline with `text-decoration: underline` doesn't have the desired effect, as the `text-decoration: none` inherited from the global styles also resets the thickness and offset properties. We do need all three.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ We've made fixes to GOV.UK Frontend in the following pull requests:
 - [#4919: Use canvas colour for cookie banner over hardcoded grey](https://github.com/alphagov/govuk-frontend/pull/4919)
 - [#4899: Remove indents from conditional reveals in radios and checkboxes](https://github.com/alphagov/govuk-frontend/pull/4899)
 - [#4935: Fix password input button unexpectedly stretching](https://github.com/alphagov/govuk-frontend/pull/4935)
+- [#4936: Fix skip link underline being removed when global styles are enabled](https://github.com/alphagov/govuk-frontend/pull/4936)
 
 ## GOV.UK Frontend v5.3.0 (Feature release)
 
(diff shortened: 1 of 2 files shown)


romaricpascal (approved):
Works neatly 🙌🏻

Issue #4937, opened 2024-04-18
`govukAttributes` macro doesn't output values that use Nunjucks' `safe` filter


querkmachine:
Raised and investigated via support.

## Description of the issue

The `govukAttributes` macro, when provided a map of key-value pairs, will fail to output values that have previously passed through the [Nunjucks `safe` filter](https://mozilla.github.io/nunjucks/templating.html#safe).

This appears to be because the output of the `safe` filter is (if not `null`) actually an object, which the `govukAttributes` macro is interpreting to be one of our first-party attribute configuration objects.

## Steps to reproduce the issue

Use this code in a project using GOV.UK Frontend's Nunjucks macros. The use of Back Link component here is illustrative, this issue is present in all components that use `govukAttributes`. 

```nunjucks
{{ govukBackLink({
  attributes: {
    'data-text': 'Testing',
    'data-html': '<i>Testing</i>',
    'data-number': 123.45,
    'data-boolean': true,
    'data-safe-text': 'Testing' | safe,
    'data-safe-html': '<i>Testing</i>' | safe,
    'data-safe-number': 123.45 | safe,
    'data-safe-boolean': true | safe
  }
}) }}
```

Each of the `safe` filtered values as they're exposed by the `dump` filter. The `safe` filter appears to convert anything passed to it into a string.

```js
// data-safe-text
{ "val": "Testing", "length": 7 }

// data-safe-html
{ "val": "<i>Testing</i>", "length": 14 }

// data-safe-number
{ "val": "123.45", "length": 6 }

// data-safe-boolean
{ "val": "true", "length": 4 }

```

## Actual vs expected behaviour

(HTML output has been reformatted for easier reading.)

### Actual output

```html
<a 
  href="#"
  class="govuk-back-link"
  data-text="Testing"
  data-html="&lt;i&gt;Testing&lt;/i&gt;"
  data-number="123.45"
  data-boolean="true"
  data-safe-text=""
  data-safe-html=""
  data-safe-number=""
  data-safe-boolean="">
  Back
</a>
```

### Expected output

```html
<a 
  href="#"
  class="govuk-back-link"
  data-text="Testing"
  data-html="&lt;i&gt;Testing&lt;/i&gt;"
  data-number="123.45"
  data-boolean="true"
  data-safe-text="Testing"
  data-safe-html="<i>Testing</i>" <!-- maybe undesirable, but would be logically consistent -->
  data-safe-number="123.45"
  data-safe-boolean="true">
[...]

PR #4938, opened 2024-04-18
Fix `govukAttributes` macro not outputting values passed from the `safe` filter

querkmachine:
Updates the `govukAttributes` macro to, when passed a set of key-value pairs, check if the value is an object that matches the shape of those returned by Nunjucks' `safe` filter (i.e. an object with `val` and `length` keys). 

If found, that object's `val` is used in place of the object. Previously, this would just be lost and the macro would return an empty string.

If the value is an object that doesn't have those specific keys, it is instead treated as one of our attribute configuration objects (which use the keys `value` and `optional` instead). 

Resolves https://github.com/alphagov/govuk-frontend/issues/4937.

## Changes
- Adds a check at the start of the `govukAttributes` attribute loop to look for objects resembling the output of the `safe` filter. If matched, the object's `val` value is mapped to the macro's working value.
- Adds a test to ensure that values passed via the `safe` filter are output by the macro as expected. 

## Thoughts
Checking for the existence of two keys is definitely not a foolproof way of ensuring that the object is actually one returned by the `safe` filter, however I'm not sure there is a more obvious way of detecting its use.

This does **not** create output that perfectly matches the expected output in #4937, as HTML is still returned escaped even if `safe` is used by the author. This is because the macro indiscriminately escapes all of the attributes and values it outputs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ We've made fixes to GOV.UK Frontend in the following pull requests:
 - [#4899: Remove indents from conditional reveals in radios and checkboxes](https://github.com/alphagov/govuk-frontend/pull/4899)
 - [#4935: Fix password input button unexpectedly stretching](https://github.com/alphagov/govuk-frontend/pull/4935)
 - [#4936: Fix skip link underline being removed when global styles are enabled](https://github.com/alphagov/govuk-frontend/pull/4936)
+- [#4938: Fix `govukAttributes` macro not outputting values passed from the `safe` filter](https://github.com/alphagov/govuk-frontend/pull/4938)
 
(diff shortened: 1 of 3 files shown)


romaricpascal (approved):
Thanks for the speedy fix. The solution looks very fine to me, I'm not sure if we'd have access to the `SafeString` class to check types anyway (and trying to check the type could lead to [subtle issues](https://github.com/mozilla/nunjucks/issues/980), which would complicate our testing).

Thinking we can explore the forced escaping in another PR if we get feedback it causes issues.

I think we can look into the values being escaped

querkmachine:
I'll see about raising another issue specifically about using `safe` to return unescaped values, but that's definitely a much lower priority than this fix.

PR #4941, opened 2024-04-19
Release v5.3.1

owenatgov:
### Fixes

We've made fixes to GOV.UK Frontend in the following pull requests:

- [#4906: Update the icon in the warning text component to match the defined text colour and background colour, rather than always being white on black](https://github.com/alphagov/govuk-frontend/pull/4906)
- [#4919: Use canvas colour for cookie banner over hardcoded grey](https://github.com/alphagov/govuk-frontend/pull/4919)
- [#4899: Remove indents from conditional reveals in radios and checkboxes](https://github.com/alphagov/govuk-frontend/pull/4899)
- [#4935: Fix password input button unexpectedly stretching](https://github.com/alphagov/govuk-frontend/pull/4935)
- [#4936: Fix skip link underline being removed when global styles are enabled](https://github.com/alphagov/govuk-frontend/pull/4936)
- [#4938: Fix `attributes` option ignoring values passed from the `safe` filter ](https://github.com/alphagov/govuk-frontend/pull/4938)


querkmachine (approved):
On the face of it it seems like stuff is missing, but a lot of changes made here are refactors that don't change the output, or changes that only apply to different-from-default Sass configurations, so they're not going to be reflected in the default CSS output.

This looks correct. 👍

PR #4942, opened 2024-04-19
Remove duplicate `errorMessage` argument for the password input component

tim-s-ccs:
Fairly simple change for something I noticed while looking at the source code.

Remove duplicate `errorMessage` argument for the password input component.


querkmachine (approved):
A simple fix for a simple oversight. Thanks for raising this @tim-s-ccs!

PR #4943, opened 2024-04-19
Various changelog updates for consistency

querkmachine:
Some of these were noticed during the 5.3.1 release process. The others are bits that I noticed of my own accord.

## Changes
- Adds a changelog entry for #4942. 
- Adds the missing npm installation instructions to the 5.3.0 and 5.3.1 releases.
- Changes the heading for the 5.3.0 release to be consistent with previous releases: changed from `GOV.UK Frontend v5.3.0 (Feature release)` to `5.3.0 (Feature release)`.
- Fixes inconsistent capitalisation of `Feature release` and `Fix release` in multiple headings. 

## Thoughts
I noticed that 5.0.0 and 5.2.0 included opening paragraphs providing an overview of the changes in that version, but 5.1.0 and 5.3.0 onwards do not. Is this something we want to keep having or is it additional to requirements?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,15 @@ For advice on how to use these release notes see [our guidance on staying up to 
 
 ## Unreleased
 
+### Fixes
+
+We've made fixes to GOV.UK Frontend in the following pull requests:
+
+- [#4942: Remove duplicate `errorMessage` argument for the password input component](https://github.com/alphagov/govuk-frontend/pull/4942) - thanks to [Tim South](https://github.com/tim-s-ccs) for contributing this change
+
 ## 5.3.1 (Fix release)
+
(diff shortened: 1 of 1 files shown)


owenatgov (approved):
Great stuff! Should we update the releases themselves as well?

Re: us not including a description at the beginning of 5.3.1 and 5.3.0, I think the frank reason is that this was just missed. Probably we need to make a call on this and document it somewhere. It's certainly something we should be conscious of if we think we'll be doing more little releases with less effort required from tech writers.

CHANGELOG.md
```diff
@@ -4,8 +4,16 @@ For advice on how to use these release notes see [our guidance on staying up to
 
 ## Unreleased
 
+### Fixes
```
claireashworth, line 7: I think I'm not understanding something. What release is this? Not 5.3.1?

querkmachine: Nah, this one just missed out and was merged after the release.

claireashworth: Ah! A rogue

claireashworth (approved):

PR #4951, opened 2024-04-22
Add result of calculation

Nepomuceno:
Helps with 
#2238

Putting the result of the calculation using a quite close approximation for results that endup not being a whole number.

### DIFF
--- a/packages/govuk-frontend/src/govuk/settings/_measurements.scss
+++ b/packages/govuk-frontend/src/govuk/settings/_measurements.scss
@@ -18,21 +18,22 @@ $govuk-page-width: 960px !default;
 /// @type Map
 /// @access public
 
-$govuk-grid-widths: (
+$govuk-grid-widths: ( 
+  // using direct values instead of divisions to keep compatibility with previous frameworks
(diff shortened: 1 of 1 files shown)


owenatgov:
Hi @Nepomuceno. Sorry we've taken ages to repond to this.

The transition path for this is to move to the Dart Sass module system, which we cannot do yet as Frontend still supports teams that are using Ruby Sass and LibSass—for which there is no alternative syntax.

Although the examples changed here don't see to have any side effects, there are about 40 more instances of / for division elsewhere in our Sass. I appreciate this is just trying to help, but we'd rather solve the problem holistically if we're able to. We're planning to move to the Dart Sass module system fairly soon, so our intent is to resolve all of them at once when we're able.

In the meantime, you can silence any warnings you're receiving from govuk-frontend by using the [quietDeps option in your Sass build configuration](https://sass-lang.com/documentation/js-api/interfaces/stringoptions/#quietDeps) or [using the --quiet-deps command line flag](https://sass-lang.com/documentation/cli/dart-sass/#quiet-deps).

PR #4955, opened 2024-04-24
Rename `all` files to `index` for our Sass entrypoints

### LINKED ISSUE #4960: Rename any `_all.scss` files to `_index.scss` so they work as index files
## What

Rename the `all` files that act as entrypoints to each of our Sass layers.

Retain the `all` files as aliases, but add deprecation warnings when importing  via them so we can remove those files in the next major release.

Update any documentation referring to the `all` files to reference the new `index` files.

## Why

Use `index` as the entrypoint for each layer as it:

- matches how we’re structuring the Sass for components
- allows users to omit ‘index’ from the filename where their Sass compiler supports index files
- is a small stepping stone towards [moving to the Sass module system](https://tannerdolby.com/writing/using-index-files-in-sass/)

## Who needs to work on this

Developers

## Who needs to review this

Developers, tech writer

## Done when

- [x] We have index files for each of our Sass layer entrypoints
- [x] We've updated any internal imports to use the new `index` files
- [x] We've retained the existing `all` files as aliases for the index files, and they output warnings when imported
- [x] We've updated any documentation that refers to the `all` files to reference the `index` files instead.

36degrees:
Use `index` as the entrypoint for each layer as it:

- matches how we’re structuring the Sass for components
- allows users to omit ‘index’ from the filename where their Sass compiler supports index files (Dart Sass, LibSass 3.6.0 or higher, or Ruby Sass 3.6.0 or higher [^1])
- is a small stepping stone towards moving to the Sass module system [^2]

Keep the existing `all` files as aliases, but add deprecation warnings when importing  via them so we can remove those files in the next major release.

Fixes #4960 

[^1]: https://sass-lang.com/documentation/at-rules/import/#index-files
[^2]: https://tannerdolby.com/writing/using-index-files-in-sass/


packages/govuk-frontend/src/govuk/components/_all.scss
```diff
@@ -1,37 +1,2 @@
...
-@import "select/index";
-@import "skip-link/index";
-@import "summary-list/index";
-@import "table/index";
-@import "tabs/index";
-@import "tag/index";
-@import "task-list/index";
-@import "textarea/index";
-@import "warning-text/index";
+@warn "Importing using 'govuk/components/all' is deprecated. Update your import statement to import 'govuk/components/index'";
```
owenatgov, line 1: What do you think about using the `warning` API for these warnings? I was thinking the warnings themselves could be different but the key could be the same.

36degrees: Potentially! My only reservation is that it might be helpful in this instance to see warnings for all instances of importing from `all`?

I could look at modifying the `_warning` mixin to make appending to `$govuk-suppressed-warnings` optional?

owenatgov: I think if we wanted to make sure that you _had_ to see the warning then in that case we might as well use `@warn`. The intent of `warning` though was to allow teams to have clean build outputs if they're not equipped to handle a deprecation, however on their own heads be it if they can head the warning but choose not to. All the warnings can still have different text and different keys if we don't want to jsut turn them all off.

36degrees: I think we can still let people suppress the warnings if they want to, but they'd see all the warnings initially rather than just the one for the first thing they were trying to import?

owenatgov: Aha, I'd forgotten that `warning` suppressed at the pint of delivery. Makes perfect sense to me.

owenatgov (approved):

PR #4961, opened 2024-04-30
Fix tree-shaking when importing `govuk-frontend`

### LINKED ISSUE #4957: Fix tree-shaking when importing from `'govuk-frontend'`
## What

Ensure that when importing individual components from `'govuk-frontend'`, only the code from these components get included in the final bundle, and not the entirety of GOV.UK Frontend.

## Why

This will make it easier for people to reduce the size of their JavaScript bundle by only including the code of the components they actually use.

## Who needs to work on this 

Developers

## Who needs to review this

Developers

## Done when

- [ ] Importing individual components from `'govuk-frontend'` includes only the code necessary for these components

romaricpascal:
This PR fixes the ability for our package to be [tree-shaken] by bundlers without extra configuration on the users part, allowing code exported by `govuk-frontend` but not imported by the user's code to be appropriately dropped from their compiled file.

[tree-shaken]: https://developer.mozilla.org/en-US/docs/Glossary/Tree_shaking

## The fix

The fix consists of adding `"sideEffects": true` to our `package.json` file. This announces to bundlers that the files of our project are free of side effects (code that modifies JavaScript globals or the DOM) on import. In turn, this lets them know that it's safe to remove code that wouldn't be imported by the user's project.

While not a field [officially documented by npm](https://docs.npmjs.com/cli/v6/configuring-npm/package-json), this is what [Webpack](https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free) and [Rollup](https://github.com/rollup/rollup/issues/2593) look at when deciding whether they can drop code from within `node_modules`.

### Unit tests

The PR adds a `package.json.unit.test.mjs` to verify the presence of that field.

### Integration tests

This work highlighted the need for us to at least have a little check that our library works OK with major bundlers. Without setting up a complex configuration, this PR adds integration tests with Rollup, Webpack and Vite to verify that tree-shaking works as expected with them.

This required the creation of a new npm workspace in the project: `@govuk-frontend/bundler-integrations` (in `.github/workflows/bundler-integrations` as this check should primarily happen on CI). This use of a workspace is necessary so that `govuk-frontend` is consumed as an external module by the bundlers. If we were to just run the bundlers in the `govuk-frontend` workspace, bundlers would have no trouble tree-shaking unused code as it would all be code from the same project.

That workspace contains light configurations for each of the bundlers (mostly to ensure either tree-shaking is enabled when it needs explicit enabling like for Webpack and/or that minifying is disabled so we can look up the names of our components in the output and not mangled variables).

This workspace is then used on CI by the `bundlers-integration.yml` workflow, automatically called as a last step by our `test.yml` (so we benefit from the caching of the dependencies and build).

### Freedom from side effect

The package is currently free from side effects thanks to our code only affecting the global scope from inside functions or classes. If functions are not called or classes not instanciated, nothing happens from only importing our files. This may change if we start including polyfills. At that point, we'll need to be cautious of either:

- remaining side-effect free by using 'pure' polyfills, themselves free of side-effects (like ['core-js-pure'](https://www.npmjs.com/package/core-js-pure))
- adjusting our 'sideEffects' field accordingly (you can list files that do have side-effects).

### Tree-shaking in file stats

While investigating this issue, we noticed that the scripts computing the stats [imports from each component's file](https://github.com/alphagov/govuk-frontend/blob/cf576d96326f81ee48a1535bf889dcfcec8e4682/shared/stats/src/index.mjs#L31) rather than from `govuk-frontend`. We may want to expand the stats to compare the file sizes when importing a component from its file and from `govuk-frontend` (which may highlight that some unnecessary code doesn't get tree-shaken).

------
Closes #4957


.github/workflows/bundler-integrations.yml
```diff
@@ -0,0 +1,41 @@
...
+        uses: ./.github/workflows/actions/build
+
+      - name: Build with bundler
+        run: npm run ${{matrix.bundler}} -w @govuk-frontend/bundler-integrations
+
+      # Check output for modules that should not be included
+      - name: Check output
+        working-directory: ./.github/workflows/bundler-integrations
+        run: |
+          ! grep "Accordion" dist/${{matrix.bundler}}/*.js
```
romaricpascal, line 41: **comment** Using `*.js` to avoid having to dig into how to configure the filename in Vite, as that bundler takes care of adding a hash and replacing the filename in `index.html`.

packages/govuk-frontend/tasks/build/package.unit.test.mjs
```diff
@@ -151,6 +151,7 @@ describe('packages/govuk-frontend/dist/', () => {
       'govuk-prototype-kit.config.json',
       'gulpfile.mjs',
       'package.json',
+      'package.json.unit.test.mjs',
```
romaricpascal, line 154: **comment** Something for another PR, but there are a couple of `.mjs` files that end up in the final package when they shouldn't (see the couple of `.test.mjs` and `.config.mjs` files after this line). There's a `*.test.*` entry in `.npmignore`, but it doesn't seem to do much.

owenatgov:
Not one we need to figure out in this PR necessarily but I think we should have a think as a dev team about which bundler examples to include and maybe establish some criteria. The inclusion of vite makes sense due to it's skyrocketing popularity and rollup is useful because we use it on the website, but should we also include examples for esbuild and parcel (I appreciate the example for parcel would hypothetically be nothing)?

My sources for what's popular _at the moment_:
- https://byby.dev/web-bundlers
- https://dev.to/olwiba/js-toolbox-2024-bundlers-and-test-frameworks-30f9

owenatgov:
I've adjusted the bundler integration tests to also account for importing via `initAll`. Specifically what I've done is:
- added an `initAll.mjs` entry point to `src`
- renamed `index.mjs` to `default.mjs` so that it's no longer the 'index' of the entry point
- adjusted the bundler scripts and configs so that they all output both a `default.js` and an `initAll.js` in their respective `dist` folders
- adjusted the test itself to also test for a successful `grep` of 'Accordion' in `initAll.js`

I've added this so that we are sure that this change, and any further changes to how we do imports, don't negatively impact users who are using `initAll` to import instead of importing specific components.

The leftover question for me are firstly, how comprehensive we want this testing to be? Right now both tests are just `grep`ing for 'Accordion' and reporting the success. Should we also `grep` for 'Button' in `default`? Should we `grep` for all components in `initAll`? In theory it wouldn't negatively impact the testing time too significantly but it's also more test code to maintain eg: if we release a new component we ahve to remember to add it to this.

Secondly, should we get rid of the webpack example in docs/examples/webpack? From a skim of this repo's docs, we don't reference it anywhere but we do now reference the bundler testing in our docs with this PR, which could act as a set of examples.

.github/workflows/bundler-integrations.yml
```diff
@@ -0,0 +1,42 @@
...
+
+      - name: Build with bundler
+        run: npm run ${{matrix.bundler}} -w @govuk-frontend/bundler-integrations
+
+      # Check output for modules that should not be included
+      - name: Check output
+        working-directory: ./.github/workflows/bundler-integrations
+        run: |
+          ! grep "Accordion" dist/${{matrix.bundler}}/default.js
+          grep "Accordion" dist/${{matrix.bundler}}/initAll.js
```
36degrees, line 42: **suggestion** The [output from the grep](https://github.com/alphagov/govuk-frontend/actions/runs/8921612225/job/24502174396?pr=4961#step:6:8) isn't particularly useful so we could consider silencing it with `--quiet` (or `-q`)

packages/govuk-frontend/tasks/build/package.unit.test.mjs
36degrees: **note** We only publish a subset of what's in `package/govuk-frontend` to npm because we're filtering it using the [`files` field](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#files) of `packages/govuk-frontend/package.json`:

https://github.com/alphagov/govuk-frontend/blob/f04749907cbab1fddc3dc488b42319c7022ca1a2/packages/govuk-frontend/package.json#L8-L13

Agree it's a little confusing as we have a combination of exclusions at build time, exclusions using `.npmignore` and a list of inclusions in that `files` field that all intersect with each other, but I think what we're publishing is broadly 'correct'.

.github/workflows/bundler-integrations/src/default.mjs
```diff
@@ -0,0 +1,8 @@
+import { Button } from 'govuk-frontend'
```
36degrees, line 1: **suggestion** I am wondering if it'd make sense to rename this file to make it clearer what the difference is? Either 'button' or 'single-component' rather than just 'default'?

May be more effort than it's worth at this point.

36degrees (approved):
I think this is looking great – @owenatgov thanks for picking this up and adding the tests for `initAll` 🙌🏻 

I only have a couple of suggestions, neither of which are blocking.

PR #4770, opened 2024-02-16
[WIP] Move component attributes into `govukAttributes()`

colinrotherham:
Work in progress PR to move key attributes into `govukAttributes()`

Due to improved checks for `null`, `undefined` or empty `""` this closes https://github.com/alphagov/govuk-frontend/issues/4669

To help discuss, I've picked two components so far:

* Button
* Input

Issue #4669, opened 2024-01-23
Text input ignoring 0 as value


dwp-dmitri-algazin:
## Description of the issue

If 0 (zero) is a valid field value, govukInput() ignores it, because of this if condition:  

https://github.com/alphagov/govuk-frontend/blob/87d7633848e14183a655bb06006cbc3f99b466d2/packages/govuk-frontend/src/govuk/components/input/template.njk#L52

As we know from JavaScript, if ( 0  ) returns false, not true

## Steps to reproduce the issue

Create govukInput with 0 value

## Actual vs expected behaviour

I would expect to see mu input with value 0 in it

## Environment (where applicable)

we use DWP Casa framework with latest GOV.UK Frontend Version: 5.0.0

- Operating system:  any
- Browser: any
- Browser version: any
- GOV.UK Frontend Version: 5.0.0

## when zero is a number ( 0 )

<img width="699" alt="Screenshot 2024-01-23 at 17 14 21" src="https://github.com/alphagov/govuk-frontend/assets/144342072/bd8600f2-6626-41b2-80bb-37c3f916a22f">

## when zero is a string ( '0' )

<img width="635" alt="Screenshot 2024-01-23 at 17 10 50" src="https://github.com/alphagov/govuk-frontend/assets/144342072/09b4aa52-ba03-46b6-be68-dbb7e9c0e62a">

PR #4963, opened 2024-05-02
Fix input value not being set if the value was '0'

patrickpatrickpatrick:
## What

Input component now uses the govukAttributes macro to validate and format attributes. Test and example added for edge-case outlined in issue https://github.com/alphagov/govuk-frontend/issues/4669.

## Why

Issue https://github.com/alphagov/govuk-frontend/issues/4669 outlines a case where setting 'value' of the Input component to 0 (integer) would not set the value of the input component. This is because the template logic would see '0' as false-y and so not set the value attribute. The govukAttributes takes this case into account and so using it means that this no longer occurs. 

This is based from https://github.com/alphagov/govuk-frontend/pull/4770. It adds tests and the input wrapper classes and attributes back.

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/input/input.yaml
+++ b/packages/govuk-frontend/src/govuk/components/input/input.yaml
@@ -399,6 +399,14 @@ examples:
       label:
         text: With value
       value: QQ 12 34 56 C
+  - name: zero value
+    hidden: true
+    options:
+      id: with-zero-value
+      name: with-zero-value
+      label:
+        text: With zero value
+      value: 0
   - name: with describedBy
     hidden: true
     options:
--- a/packages/govuk-frontend/src/govuk/components/input/template.njk
+++ b/packages/govuk-frontend/src/govuk/components/input/template.njk
@@ -3,9 +3,20 @@
 {% from "../hint/macro.njk" import govukHint %}
 {% from "../label/macro.njk" import govukLabel %}
 
+{#- Set classes for this component #}
+{%- set classNames = "govuk-input" -%}
+
+{%- if params.classes %}
+  {% set classNames = classNames + " " + params.classes %}
+{% endif %}
+
+{%- if params.errorMessage %}
+  {% set classNames = classNames + " govuk-input--error" %}
+{% endif %}
+
 {#- a record of other elements that we need to associate with the input using
   aria-describedby – for example hints or error messages -#}
-{% set describedBy = params.describedBy if params.describedBy else "" -%}
+{% set describedBy = params.describedBy if params.describedBy else undefined -%}
 
 {%- set hasPrefix = true if params.prefix and (params.prefix.text or params.prefix.html) else false %}
 {%- set hasSuffix = true if params.suffix and (params.suffix.text or params.suffix.html) else false %}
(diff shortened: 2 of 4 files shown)


packages/govuk-frontend/src/govuk/components/input/template.njk
```diff
@@ -3,25 +3,64 @@
...
+      disabled: {
+        value: params.disabled,
+        optional: true
+      },
+      "aria-describedby": {
+        value: describedBy,
+        optional: true
+      },
+      autocomplete: {
+        value: params.autocomplete | default("none", true)
```
36degrees, line 52: It looks like this has changed from #4770 and changes the behaviour slightly.

Previously, if `params.autocomplete` wasn't set, we wouldn't output an `autocomplete` attribute.

Now, if `params.autocomplete` isn't set, we output a default `autocomplete="none"` [which isn't valid](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete#values).

This is one of the reasons the tests are failing – we run an html validator against all of the examples and [it's erroring with '"none" is not a valid autocomplete token or field name'](https://github.com/alphagov/govuk-frontend/actions/runs/8922309070/job/24504321827?pr=4963#step:6:483).

packages/govuk-frontend/src/govuk/components/input/template.njk
```diff
@@ -3,25 +3,64 @@
...
 {%- macro _inputElement(params) -%}
-  <input class="govuk-input {%- if params.classes %} {{ params.classes }}{% endif %} {%- if params.errorMessage %} govuk-input--error{% endif %}" id="{{ params.id }}" name="{{ params.name }}" type="{{ params.type | default("text", true) }}"
-    {%- if (params.spellcheck === false) or (params.spellcheck === true) %} spellcheck="{{ params.spellcheck }}"{% endif %}
-    {%- if params.value %} value="{{ params.value }}"{% endif %}
-    {%- if params.disabled %} disabled{% endif %}
-    {%- if describedBy %} aria-describedby="{{ describedBy }}"{% endif %}
-    {%- if params.autocomplete %} autocomplete="{{ params.autocomplete }}"{% endif %}
-    {%- if params.pattern %} pattern="{{ params.pattern }}"{% endif %}
-    {%- if params.inputmode %} inputmode="{{ params.inputmode }}"{% endif %}
-    {%- if params.autocapitalize %} autocapitalize="{{ params.autocapitalize }}"{% endif %}
```
36degrees, line 24: **issue** It looks like the `autocapitalize` attribute was added in #4442, after the work in #4770 was done, so this will need adding to the new `govukAttributes` call.

(There's a corresponding test for this which we need to keep too)

packages/govuk-frontend/src/govuk/components/input/template.test.js
```diff
@@ -455,22 +446,4 @@ describe('Input', () => {
       expect($prefixBeforeSuffix.length).toBeTruthy()
     })
   })
-
-  describe('when it includes the input wrapper', () => {
```
36degrees, line 459: **issue** Why have these tests been removed?

packages/govuk-frontend/src/govuk/components/input/input.yaml
```diff
@@ -399,6 +399,14 @@ examples:
...
         text: With value
       value: QQ 12 34 56 C
+  - name: zero value
+    hidden: true
+    options:
+      id: with-zero-value
+      name: with-zero-value
+      label:
+        text: With zero value
+      value: !!int 0
```
36degrees, line 409: **question** Is the casting definitely necessary here? My understanding was that YAML should treat this is an integer already.

packages/govuk-frontend/src/govuk/components/input/template.test.js
```diff
@@ -71,6 +71,13 @@ describe('Input', () => {
       expect($component.val()).toBe('QQ 12 34 56 C')
     })
 
+    it('renders with zero value', () => {
```
36degrees, line 74: **praise** Thanks for adding a test for this 🙌🏻

36degrees (commented):
Thanks for picking this up and congrats on your first contribution to GOV.UK Frontend on the team 😎 🎉 

There's a couple of things that aren't quite right at the minute – in particular it looks like the changes in #4770 might have been out of date with some other changes to add the `autocapitalize` attribute, sorry we'd missed that!

Let me know if it'd be helpful to chat through any of this or if you've got any questions!

36degrees (approved):
Looks good to me! 🙌🏻

PR #4971, opened 2024-05-07
Fix Error Summary component outputting list HTML when no `errorList` is provided

### LINKED ISSUE #3864: Should errorList be a required field in ErrorSummary component?
I have a use case where I'm showing a single generic error that isnt tied to a specific form input.

In https://design-system.service.gov.uk/components/notification-banner/ guidance it says to use the Error Summary component.

However the error list is marked as required and if you do not submit an error list you'll get an empty list.

Should this be marked as optional and only output the error list if there's an errorList supplied?

querkmachine:
Modifies the Error Summary component and documentation to no longer require that the `errorList` parameter be set, and to not render associated HTML if it's empty.

This is intended to benefit situations where an error may not be associated with user input or any particular form field (e.g. a fault with the service), and where being able to provide `descriptionText`/`descriptionHtml` may suffice to explain the issue to a user.

Based on the conclusions reached in #3864. Closes #3864.

## Changes
- Adds a new condition in the Nunjucks template to not output the `ul.govuk-error-summary__list` element if the `errorList` parameter is empty.
- Marks `errorList` as not being required in Nunjucks parameter documentation.
- Adds a review app example of an Error Summary with a title and description, but no error list.
- Adds tests to ensure that `ul.govuk-error-summary__list` is output if there are error list items, and not if there isn't. 

## Thoughts
I wasn't 100% sure on where this would live in the changelog. It's ultimately a very minor logic and documentation update that could be sold as both a new feature ("You can now use the Error Summary without providing a list of errors") and a simple bugfix ("The Error Summary no longer outputs a list if there isn't anything to put in the list"). 

I ended up listing it as a bugfix, as it's ultimately a very inconsequential change (people could already not pass any `errorList` items, it'd just return the 'wrong' HTML), and renamed this PR appropriately.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,7 +42,7 @@ It was previously only possible to use tabular numbers by using the `govuk-font-
 
 This change was introduced in [pull request #4973: Add override class for tabular numbers](https://github.com/alphagov/govuk-frontend/pull/4973).
 
-### Deprecations
+### Deprecated features
 
 #### Importing layers using `all` files
 
@@ -70,6 +70,7 @@ We've made fixes to GOV.UK Frontend in the following pull requests:
 - [#4942: Remove duplicate `errorMessage` argument for the password input component](https://github.com/alphagov/govuk-frontend/pull/4942) - thanks to [Tim South](https://github.com/tim-s-ccs) for contributing this change
 - [#4961: Fix tree-shaking when importing `govuk-frontend`](https://github.com/alphagov/govuk-frontend/pull/4961)
 - [#4963: Fix input value not being set if the value was '0'](https://github.com/alphagov/govuk-frontend/pull/4963) – thanks to [@dwp-dmitri-algazin](https://github.com/dwp-dmitri-algazin) for reporting this issue
+- [#4971: Fix Error Summary component outputting list HTML when no `errorList` is provided](https://github.com/alphagov/govuk-frontend/pull/4971)
 
 ## 5.3.1 (Fix release)
 
--- a/packages/govuk-frontend/src/govuk/components/error-summary/_index.scss
+++ b/packages/govuk-frontend/src/govuk/components/error-summary/_index.scss
@@ -24,14 +24,21 @@
 
   .govuk-error-summary__body {
     p {
-      margin-top: 0;
-      @include govuk-responsive-margin(4, "bottom");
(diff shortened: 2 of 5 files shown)


romaricpascal (commented):
Thanks for looking into that one. The changes to the templates look neat and sounds fine to have the change as a fix. 🙌🏻 

I think it might need a little CSS update as well. With only the description, there's some extra spacing from the `<p>` that wraps the description. 

<a href="https://govuk-frontend-pr-4971.herokuapp.com/components/error-summary/with-description-only/preview">Screenshot of the description only example for the Error summary, showing extra space</a>

Maybe something like:

```scss
.govuk-error-summary__body {
  > * {
  	margin-bottom: 0;
  	margin-top: 0;
  }
  
  > * + * {
  	@include govuk-responsive-margin(4, "top");
  }
}
```

We might also want a little design look and/or some guidance regarding when to use a description or a list with a single item, as the output is visually the same (spacing aside) but the underlying semantics are different.

<a href="https://govuk-frontend-pr-4971.herokuapp.com/components/error-summary/without-links/preview">Screenshot of the without links example for the Error summary</a>

querkmachine:
@romaricpascal I've updated the PR to use a variant of your suggestion, though I note that it didn't completely solve the issue of uneven negative space between description and list variants, as list items also have some uncollapsed `margin-bottom` applied to them! I added a rule to remove the margin from the last list item too.

In future it'd probably be a good idea to refactor this component a little to potentially use a vertical flexbox and `gap` instead.

romaricpascal (approved):
Good spot on these 5px of blank space after that last `<li>` 🦅  All good to go for me 😊

mia-allers-gds:
I'm late to this but I have some thoughts. Basically I think we should roll back to the previous version with the +5px bottom margin.

The last line item in a paragraph or a banner or a box or a button should, in my opinion, have a very slightly bigger bottom margin. The reason for this is that there are often (though not always) descenders in the type (a descender is a letter that descends below the baseline, for example a `y` or a `g` )

These descenders will then sometimes make the last line appear closer to whatever sits below it. Additionally, looking at this particular change, the underline also makes the space look much smaller. And finally, the margin at the top reached the highest point of the type, the capital letters. The bottom margin extends from the baseline. This will also give create more whitespace at the top, which needs to be balanced by increasing the bottom margin.

PR #4973, opened 2024-05-07
Add override class for tabular numbers

querkmachine:
Adds an override class that invokes the `govuk-font-tabular-numbers` mixin. 

I think this is a necessary feature to help meet https://github.com/alphagov/govuk-design-system/issues/1233, as users are currently unable to use tabular numbers without using Sass—which in turn requires having a Sass build pipeline set up and authoring bespoke code—which is a bit of a faff, really.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,18 @@ For advice on how to use these release notes see [our guidance on staying up to 
 
 ## Unreleased
 
+### New features
+
+#### Use tabular numbers easily with `govuk-!-font-tabular-numbers`
+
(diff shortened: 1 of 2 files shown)


owenatgov (approved):
Seems reasonable to me 👍🏻

There's a tricky thing here around what to name it. We've currently got the convention with our overrides that they follow the same rule written out in css as a kebab case class eg: `font-weight: bold` becomes `font-weight-bold`. Therefore should this be `govuk-!-font-variant-numeric-tabular-nums` or some variant? I personally think this is a reasonable tradeoff as the full class name looks a bit yucky. We could experiment with variants but I still think this communicates what it is.

PR #4975, opened 2024-05-09
Add `createAll` function to initialise individual components

romaricpascal:


36degrees (approved):

packages/govuk-frontend/src/govuk/all.mjs
```diff
@@ -1,124 +1,19 @@
...
+export { Checkboxes } from './components/checkboxes/checkboxes.mjs'
+export { ErrorSummary } from './components/error-summary/error-summary.mjs'
+export { ExitThisPage } from './components/exit-this-page/exit-this-page.mjs'
+export { Header } from './components/header/header.mjs'
+export { NotificationBanner } from './components/notification-banner/notification-banner.mjs'
+export { PasswordInput } from './components/password-input/password-input.mjs'
+export { Radios } from './components/radios/radios.mjs'
+export { SkipLink } from './components/skip-link/skip-link.mjs'
+export { Tabs } from './components/tabs/tabs.mjs'
+export { initAll, createAll } from './init.mjs'
```
romaricpascal, line 14: **note** With the removal of `initAll` from `all.mjs`, all the file does is re-exporting the difference pieces of our library that constitute our public API, so we can switch to the `export { NAMED_IMPORT } from 'module'` syntax rather than have a bunch of `import`s followed by an `export`.

packages/govuk-frontend/tasks/build/package.unit.test.mjs
```diff
@@ -181,24 +181,20 @@ describe('packages/govuk-frontend/dist/', () => {
...
+          export { Checkboxes } from './components/checkboxes/checkboxes.mjs';
+          export { ErrorSummary } from './components/error-summary/error-summary.mjs';
+          export { ExitThisPage } from './components/exit-this-page/exit-this-page.mjs';
+          export { Header } from './components/header/header.mjs';
+          export { NotificationBanner } from './components/notification-banner/notification-banner.mjs';
+          export { PasswordInput } from './components/password-input/password-input.mjs';
+          export { Radios } from './components/radios/radios.mjs';
+          export { SkipLink } from './components/skip-link/skip-link.mjs';
+          export { Tabs } from './components/tabs/tabs.mjs';
+          export { createAll, initAll } from './init.mjs';
```
romaricpascal, line 196: **note** This change is unrelated to our move to using `export` in `all.mjs`. Even keeping `import`s + a final `export`, Rollup would rewrite to the `export { NAMED_EXPORT } from 'module'` syntax.

packages/govuk-frontend/tasks/build/release.unit.test.mjs
```diff
@@ -126,7 +126,7 @@ describe('dist/', () => {
 
     it('should contain relative paths to sources', () => {
       expect(sourcemap.sources).toContain(
-        '../packages/govuk-frontend/src/govuk/all.mjs'
+        '../packages/govuk-frontend/src/govuk/init.mjs'
```
romaricpascal, line 129: **note** `sourcemap.sources` lists many other modules, which was already the case alongside `all.mjs`, so we can just swap to another one of our files being imported.

PR #4980, opened 2024-05-13
Fix tree-shaking of inner modules for the Webpack integration

### LINKED ISSUE #4964: Investigate tree shaking of common js utilities
## What
Investigate ways to ensure that our [common js utilities](https://github.com/alphagov/govuk-frontend/tree/main/packages/govuk-frontend/src/govuk/common), primarily the multiple modules in [common/index/mjs](https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/src/govuk/common/index.mjs), are correctly tree-shaken when bundled. This includes but is not limited to splitting the modules in index.mjs into their own files.

## Why
We found during https://github.com/alphagov/govuk-frontend/issues/4957 that whilst we were able to effectively get tree shaking to work as expected for components, bundlers were importing all of common/index.mjs when not all of common was being used. It would be good to further optimise our js and understand why our `"sideEffects": true` solution in https://github.com/alphagov/govuk-frontend/pull/4961 didn't work for common.

## Who needs to work on this
Developers

## Who needs to review this
Developers

## Done when

- [ ] Test if splitting out common/index.mjs fixes tree shaking
- [ ] Optional: Investigate alternative methods to fix the tree shaking

romaricpascal:
We were not quite following the right step to optimise the build as it would be in an actual production set up while still retaining the possibility to inspect the content. This PR amends the Webpack build so tree-shaking is set up properly and internal utilities not used by `single-component.mjs` get excluded.

Tree-shaking in Webpack is not only a matter of setting `usedExports` (which still [includes unused code, just doesn't export it](https://webpack.js.org/guides/tree-shaking/#add-a-utility)), but also of having the `TerserPlugin` set up, as it's the one [effectively removing the unused code](https://webpack.js.org/guides/tree-shaking/#minify-the-output).

Instead, tree-shaking is properly set up [by using the `production` mode](https://webpack.js.org/guides/tree-shaking/#minify-the-output). We can then tweak the configuration of TerserPlugin, disabling mangling so we can look at class, function and variable names in the output. 

Closes #4964

### DIFF
--- a/.github/workflows/bundler-integrations.yml
+++ b/.github/workflows/bundler-integrations.yml
@@ -35,8 +35,17 @@ jobs:
         run: npm run ${{matrix.bundler}} -w @govuk-frontend/bundler-integrations
 
       # Check output for modules that should not be included
-      - name: Check output
+      - name: Check absence of unused modules in `single-component.js`
         working-directory: ./.github/workflows/bundler-integrations
         run: |
           ! grep "Accordion" dist/${{matrix.bundler}}/single-component.js -q
+
+      - name: Check absence of unused utility functions in `single-component.js`
+        working-directory: ./.github/workflows/bundler-integrations
+        run: |
+          ! grep "getFragmentFromUrl" dist/${{matrix.bundler}}/single-component.js -q
+
+      - name: Check presence of modules in `initAll.js`
+        working-directory: ./.github/workflows/bundler-integrations
+        run: |
           grep "Accordion" dist/${{matrix.bundler}}/initAll.js -q
--- a/.github/workflows/bundler-integrations/package.json
+++ b/.github/workflows/bundler-integrations/package.json
@@ -18,6 +18,7 @@
     "del-cli": "^5.1.0",
     "govuk-frontend": "*",
     "rollup": "^4.17.2",
+    "terser-webpack-plugin": "^5.3.10",
     "vite": "^5.2.10",
     "webpack": "^5.91.0"
   }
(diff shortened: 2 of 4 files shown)


.github/workflows/bundler-integrations.yml
```diff
@@ -35,8 +35,17 @@ jobs:
...
+
+      - name: Check absence of unused utility functions in `single-component.js`
+        working-directory: ./.github/workflows/bundler-integrations
+        run: |
+          ! grep "getFragmentFromUrl" dist/${{matrix.bundler}}/single-component.js -q
+
+      - name: Check presence of modules in `initAll.js`
+        working-directory: ./.github/workflows/bundler-integrations
+        run: |
+          grep "initAll" dist/${{matrix.bundler}}/initAll.js -q
```
romaricpascal, line 51: **note** Splitting those on their own command. Despite [GitHub using the `-e` flag for its default shell](https://github.com/github/docs/issues/23853), which should fail fast, the grep for `getFragmentFromUrl` failing didn't lead to the failure of the check.

.github/workflows/bundler-integrations.yml
```diff
@@ -35,8 +35,17 @@ jobs:
...
           ! grep "Accordion" dist/${{matrix.bundler}}/single-component.js -q
+
+      - name: Check absence of unused utility functions in `single-component.js`
+        working-directory: ./.github/workflows/bundler-integrations
+        run: |
+          ! grep "getFragmentFromUrl" dist/${{matrix.bundler}}/single-component.js -q
+
+      - name: Check presence of modules in `initAll.js`
+        working-directory: ./.github/workflows/bundler-integrations
+        run: |
```
romaricpascal, line 50: **note** Splitting those on their own command. Despite [GitHub using the -e flag for its default shell](https://github.com/github/docs/issues/23853), which should fail fast, the grep for `getFragmentFromUrl` failing didn't lead to the failure of the check.

owenatgov (approved):

PR #4981, opened 2024-05-13
Build each entry independently when testing Vite

### LINKED ISSUE #4978: Review Vite integration tests
## What

Review Vite integration test to ensure each entry is built independently. If this means having two separate scripts that each build an entry, that's fine.

## Why

Vite seems to treat [the two entries we configured for building](https://github.com/alphagov/govuk-frontend/blob/6c9d8d7191edb18e8a46d2a383e311f397367946/.github/workflows/bundler-integrations/vite.config.mjs#L12) as part of a single thing, and extracts shared code in its own module. This would make sense if we were building a single project with multiple entry points. However, what we want to do is check that each of our entry points, when build on its own, includes what we expect. Because Vite splits the shared code in its own file, we can't verify that tree-shaking works by only looking in the build `initAll.mjs` or `single-component.mjs`, as some code in the shared module may not have been tree-shaken, or may be there because `initAll` imports it.

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [ ] Vite builds the two entries independently

romaricpascal:
Builds `initAll.mjs` and `single-component.mjs` independently during Vite integration tests, to work around Vite cleverly optimising a build with multiple entries by splitting shared code in a separate file. [After investigation](https://github.com/alphagov/govuk-frontend/issues/4978#issuecomment-2107863308), it seems we can only make separate builds to circumvent that cleverness.

Simplest option looked to allow to pass the name of the file through environment variable, so we can run the build command multiple times targeting a separate entry. As we want to have the same build folder structure as for Webpack and Rollup, both builds will need to output to the same folder, for which we need to disable Vite automatically cleaning up its build folder.

Closes #4978.

### DIFF
--- a/.github/workflows/bundler-integrations/vite.config.mjs
+++ b/.github/workflows/bundler-integrations/vite.config.mjs
@@ -1,6 +1,10 @@
-/** @type {import('vite').UserConfig} */
-// the default vite config used by the different test case configs
+// Allows to run the same configuration against different files
+// as Vite would split shared code in a separate chunk if trying
+// to build an array of entries in `build.rollupOptions.input`
+// or `build.lib.entry`
+const entryName = process.env.ENTRY_NAME ?? 'single-component'
 
+/** @type {import('vite').UserConfig} */
 export default {
   build: {
     // Align output with other bundlers to facilitate testing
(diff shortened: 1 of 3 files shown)


.github/workflows/bundler-integrations/vite.config.mjs
```diff
@@ -1,15 +1,22 @@
-/** @type {import('vite').UserConfig} */
-// the default vite config used by the different test case configs
+// Allows to run the same configuration against different files
+// as Vite would split shared code in a separate chunk if trying
+// to build an array of entries in `build.rollupOptions.input`
+// or `build.lib.entry`
+const entryName = process.env.ENTRY_NAME ?? 'single-component'
```
romaricpascal, line 5: **note** Set a default here to avoid things breaking badly when runing Vite outside of `npm run vite -w @govuk-frontend/bundler-integrations`

36degrees (approved):
Makes sense to me 👍🏻

PR #4983, opened 2024-05-16
Stop warnings/errors from raw SCSS calculations

mark-roberts-ho:
There are a few calculations which are not wrapped in `calc() `and thus generate warnings.

The effect of this change is to stop these errors/warnings being reported in the console

### DIFF
--- a/packages/govuk-frontend/src/govuk/settings/_measurements.scss
+++ b/packages/govuk-frontend/src/govuk/settings/_measurements.scss
@@ -20,19 +20,19 @@ $govuk-page-width: 960px !default;
 
 $govuk-grid-widths: (
   one-quarter: (
-    100% / 4
+    calc(100% / 4)
   ),
   one-third: (
-    100% / 3
+    calc(100% / 3)
   ),
   one-half: (
-    100% / 2
+    calc(100% / 2)
   ),
   two-thirds: (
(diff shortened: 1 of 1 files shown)


querkmachine:
Hi @mark-roberts-ho,

The calculations you've changed here are Sass calculations. They do not need to be wrapped in `calc` as they are intended to be resolved during the Sass compilation process, rather than on the client. 

The warnings are due to [the deprecation of using `/` for division](https://sass-lang.com/documentation/breaking-changes/slash-div/). The transition path for this is to move to the Dart Sass module system, which we cannot do yet as Frontend still supports teams that are using Ruby Sass and LibSass—for which there is no alternative syntax.

Although the examples changed here don't see to have any side effects, there are about 40 more instances of `/` for division elsewhere in our Sass. We're planning to move to the Dart Sass module system fairly soon, so our intent is to resolve all of them at once when we're able. 

In the meantime, you can silence any warnings you're receiving from govuk-frontend by using the [`quietDeps` option in your Sass build configuration](https://sass-lang.com/documentation/js-api/interfaces/stringoptions/#quietDeps) or [using the `--quiet-deps` command line flag](https://sass-lang.com/documentation/cli/dart-sass/#quiet-deps).

PR #4985, opened 2024-05-17
Selectively silence deprecations and fix missed warnings

36degrees:
Silencing all logs means that we may miss:

- any deprecations from Sass that we can fix now without breaking support for LibSass and Ruby Sass
- tests that need updating because of our own deprecations (like the recent deprecation of importing using the `all` entrypoints!)

Instead we can use the new [`silenceDeprecations` option](https://sass-lang.com/documentation/js-api/interfaces/options/#silenceDeprecations) introduced in Sass 1.74.0 and silence just the deprecations we know we can’t do anything about.

Fix the deprecation warnings that we can now see 🎉

### DIFF
--- a/shared/helpers/tests.js
+++ b/shared/helpers/tests.js
@@ -1,7 +1,7 @@
 const { join } = require('path')
 
 const { paths } = require('@govuk-frontend/config')
-const { compileAsync, compileStringAsync, Logger } = require('sass-embedded')
+const { compileAsync, compileStringAsync } = require('sass-embedded')
 
 const sassPaths = [
   join(paths.package, 'src/govuk'),
@@ -18,7 +18,7 @@ const sassPaths = [
 async function compileSassFile(path, options = {}) {
   return compileAsync(path, {
     loadPaths: sassPaths,
-    logger: Logger.silent,
+    silenceDeprecations: ['slash-div'],
     quietDeps: true,
     ...options
   })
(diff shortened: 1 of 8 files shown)


shared/helpers/tests.js
```diff
@@ -18,7 +18,7 @@ const sassPaths = [
 async function compileSassFile(path, options = {}) {
   return compileAsync(path, {
     loadPaths: sassPaths,
-    logger: Logger.silent,
+    silenceDeprecations: ['slash-div'],
```
colinrotherham, line 21: @36degrees Really glad this is out now, I had my eyes on this 🙌

Remember I had to [write my own logger](https://github.com/alphagov/govuk-frontend/blob/adfa0abc33b54262a08d9b2218cd8123a4b0ecf1/shared/tasks/styles.mjs#L137-L143) for Sass builds? Can probably whip that out now too

It's suppressing these two currently:

```js
suppressed: [
  'Using / for division is deprecated and will be removed in Dart Sass 2.0.0.',
  'Using / for division outside of calc() is deprecated and will be removed in Dart Sass 2.0.0.'
]
```

36degrees: Thanks for the heads up! 🙌🏻

I'll raise an issue as something to do in a follow up PR.

romaricpascal (approved):
That `silenceDeprecations` is quite handy 🙌🏻 Thanks for tidying up the tests 😊

Issue #4451, opened 2023-11-10
Breadcrumb component accessibility


jon-kirwan:
## Related component

[Breadcrumbs](https://design-system.service.gov.uk/components/breadcrumbs/)

## Context

We have a stale [pull request - improve a11y of breadcrumbs](https://github.com/alphagov/govuk_publishing_components/pull/2045) in the GOV.UK Publishing Components repo aimed at enhancing breadcrumb accessibility by adding a `nav` landmark. We recently discussed whether to close the PR or continue the work.

The comments on the pull request go into both the advantages and disadvantages of implementing the suggested change, as well as the historical context of the component and the rationale behind not using a landmark. Ultimately, no changes were made, probably due to the lack of evidence and / absence of user research on the benefits of the proposed change.

From recently reviewing the issue again, a couple of ideas were mentioned that could improve component accessibility. For example, including a hidden heading, adding a `nav` landmark or incorporating visible text such as "**you are here:**" to make it easier to understand. (See [comments](https://github.com/alphagov/govuk_publishing_components/pull/2045#issuecomment-1803596065) from @selfthinker).

## Alternatives

None.

## Additional information (if applicable)

None.

querkmachine:
Yesterday we merged some changes to the breadcrumb component (https://github.com/alphagov/govuk-frontend/pull/4995), which included making it a `nav` element and adding an `aria-label` attribute to it, making it a named landmark.

This was prioritised due to our ongoing navigation work, where it was identified that when a navigation component and a breadcrumb component appeared next to each other, there was insufficient affordances for screen reader users to tell the two apart.

It's not the most perfect solution. For example, we didn't add a heading, so the two elements still appear visually identical in text-only browsers or if CSS fails to load. Hopefully this is still an improvement, though!

PR #4995, opened 2024-05-20
Update Breadcrumb component to improve screen reader accessibility

querkmachine:
An accessibility issue raised during development of #4950 was that, when a Breadcrumbs component is placed following a Header or Service Header that contains navigation elements, they are both announced (by screen readers) or displayed (in no CSS/reader modes) as a bulleted lists of links, with little affordance to users that they are two distinct navigational elements. 

This PR aims to resolve the screen reader issue by adding an `aria-label` of "Breadcrumb". It does not solve the display issue. Doing so would require the inclusion of some visible or visually-hidden text, judging by @selfthinker's comment on the GOV.UK Publishing Components version: https://github.com/alphagov/govuk_publishing_components/pull/2045#issuecomment-1803596065, though we could choose to modify this PR to do that as well.

As `aria-label` is only supported in practice by interactive elements and landmarks, the breadcrumb component wrapper has been changed to a `nav` to make it a landmark element. Doing so also resolves a common accessibility audit complaint that this component does not exist within a landmark (e.g. https://github.com/alphagov/govuk-frontend/issues/1604). This is not strictly required to meet WCAG, but may be considered good practice. 

## Background

The breadcrumb navigation was previously contained within a `nav` element, but this was removed in 2013 in an effort to reduce the number of navigation landmarks that appear on the page, as documented [in this blog post](https://insidegovuk.blog.gov.uk/2013/07/03/rethinking-navigation/) by @tombye. 

However, I think that the Breadcrumbs do not meet the criteria for removing `nav` as given in the post—the individual links do not provide sufficient labelling on their own and the components' role as a navigational aid can only be inferred through visual context which is not available to users who are blind or with low vision, nor assistive technologies they may use, and for which no fallback is provided.

Compare Breadcrumbs with other `nav` elements it's common to find on GOV.UK:
- Primary navigation has an accessible label and can have context inferred by being within the `header` (`banner`) landmark and near the top of the page.
- Links in the page footer have visible headings and can have context inferred by being within the `footer` (`contentinfo`) landmark.
- Most other `nav` elements that appear on GOV.UK have visible headings (e.g. related content lists), accessible labels (e.g. pagination), or can have some context inferred from their parents (e.g. table of contents has a `complementary` role and is one of the first things inside of the `main` landmark). Most provide multiple of these affordances.
- Skip links and back links don't typically exist in groups and can have their contexts largely by their labelling. 

Breadcrumbs currently have none of these: they lack the `nav` element, they lack headings (visible or not), lack accessible labelling, their context cannot be inferred from a parent landmark because they don't have one, and it cannot be inferred from the links themselves as (apart from 'Home') they are disembodied section names.

With these changes, our component is much more in line with the example in the [WAI ARIA Authoring Guide](https://www.w3.org/WAI/ARIA/apg/patterns/breadcrumb/examples/breadcrumb/) and the [W3C's design system component](https://design-system.w3.org/components/breadcrumbs.html) (which is heavily inspired by ours, their deviation implying they felt we were in the wrong here). The same approach is also used by our friends at [ONS](https://service-manual.ons.gov.uk/design-system/components/breadcrumbs#example.example-breadcrumbs.html), [Scottish Government Design System](https://designsystem.gov.scot/components/breadcrumbs) and [USWDS](https://designsystem.digital.gov/components/breadcrumb/), really putting us in the minority with our current approach. 

## Changes
- Changes the wrapping element of the Breadcrumbs component from a `div` to a `nav`.
- Adds a required `aria-label` to the Breadcrumb component wrapper.
- Adds a new Nunjucks parameter to set the `aria-label` named `labelText`. This defaults to `"Breadcrumb"`.
- Adds Nunjucks parameter documentation for `labelText`.
- Adds tests to ensure the element is a `nav`, renders the default `aria-label`, and that the Nunjucks parameter can override the `aria-label`. 
- Updates one existing example and test that set `role="navigation"`, as this was being flagged by the HTML linter as redundant role usage.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,16 @@ For advice on how to use these release notes see [our guidance on staying up to 
 
 ## Unreleased
 
+### Recommended changes
+
+#### Update Breadcrumbs to use `nav` and `aria-label`
+
+We've made changes to the Breadcrumbs component to improve how it appears to screen readers.
+
+We've changed the wrapping element to use the `nav` tag to expose it as a navigational landmark, and added an `aria-label` attribute to differentiate it as breadcrumb navigation.
+
+This change was introduced in [pull request #4995: Update Breadcrumb component to improve screen reader accessibility](https://github.com/alphagov/govuk-frontend/pull/4995).
+
 ### Fixes
 
 We've made fixes to GOV.UK Frontend in the following pull requests:
--- a/packages/govuk-frontend/src/govuk/components/breadcrumbs/breadcrumbs.yaml
+++ b/packages/govuk-frontend/src/govuk/components/breadcrumbs/breadcrumbs.yaml
@@ -32,6 +32,10 @@ params:
     type: object
     required: false
     description: HTML attributes (for example data attributes) to add to the breadcrumbs container.
+  - name: labelText
+    type: string
+    required: false
+    description: Plain text label identifying the landmark to screen readers. Defaults to "Breadcrumb".
 
 examples:
   - name: default
@@ -108,7 +112,7 @@ examples:
     options:
       attributes:
         id: my-navigation
-        role: navigation
+        'data-foo': 'bar'
       items:
         - text: Home
   - name: item attributes
@@ -134,3 +138,13 @@ examples:
         - html: <em>Section 1</em>
           href: /section-1
         - html: <em>Section 2</em>
+          href: /section-2
+  - name: custom label
+    hidden: true
+    options:
+      labelText: Briwsion bara
+      items:
+        - text: Hafan
+          href: '/'
+        - text: Sefydliadau
+          href: '/organisations'
--- a/packages/govuk-frontend/src/govuk/components/breadcrumbs/template.jsdom.test.js
+++ b/packages/govuk-frontend/src/govuk/components/breadcrumbs/template.jsdom.test.js
@@ -18,6 +18,14 @@ describe('Breadcrumbs', () => {
       $listItems = document.querySelectorAll('li.govuk-breadcrumbs__list-item')
     })
 
+    it('renders as a nav element', () => {
+      expect($component.tagName.toLowerCase()).toBe('nav')
+    })
+
+    it('renders with default aria-label', () => {
+      expect($component).toHaveAttribute('aria-label', 'Breadcrumb')
+    })
+
     it('includes an ordered list', () => {
       expect($component).toContainElement($list)
     })
@@ -152,7 +160,14 @@ describe('Breadcrumbs', () => {
 
       const $component = document.querySelector('.govuk-breadcrumbs')
       expect($component).toHaveAttribute('id', 'my-navigation')
-      expect($component).toHaveAttribute('role', 'navigation')
+      expect($component).toHaveAttribute('data-foo', 'bar')
+    })
+
+    it('renders with a custom aria-label', () => {
+      document.body.innerHTML = render('breadcrumbs', examples['custom label'])
+
+      const $component = document.querySelector('.govuk-breadcrumbs')
+      expect($component).toHaveAttribute('aria-label', 'Briwsion bara')
     })
   })
 })
--- a/packages/govuk-frontend/src/govuk/components/breadcrumbs/template.njk
+++ b/packages/govuk-frontend/src/govuk/components/breadcrumbs/template.njk
@@ -11,7 +11,7 @@
   {% set classNames = classNames + " govuk-breadcrumbs--collapse-on-mobile" %}
 {% endif -%}
 
-<div class="{{ classNames }}" {{- govukAttributes(params.attributes) }}>
+<nav class="{{ classNames }}" {{- govukAttributes(params.attributes) }} aria-label="{{ params.labelText | default("Breadcrumb") }}">
   <ol class="govuk-breadcrumbs__list">
 {% for item in params.items %}
   {% if item.href %}
@@ -27,4 +27,4 @@
   {% endif %}
 {% endfor %}
   </ol>
-</div>
+</nav>


selfthinker:
I generally agree with Tom's initial assessment back in 2013 that there were way too many navigation landmarks on the page. Although they had worked on removing some, I assume some have moved back in or were added later as www.gov.uk still uses way too many, lots of which should perhaps be removed. (For example, [tax codes](https://www.gov.uk/tax-codes) has 6 navigation landmarks, at least half of which are probably not necessary.)

> Breadcrumbs currently have none of these: they lack the `nav` element, they lack headings (visible or not), lack accessible labelling, their context cannot be inferred from a parent landmark because they don't have one, and it cannot be inferred from the links themselves as (apart from 'Home') they are disembodied section names.

That is a very good point. A point that makes me think maybe this is a good idea.

Something that is a bit problematic is that when breadcrumbs are used on www.gov.uk, they are often not very useful as the website's hierarchy is presented as pretty flat. Therefore the breadcrumbs often only have a "Home" link and then a link to the page's general topic.
I don't know enough about how breadcrumbs are used on services to know if they are more useful there. But I can imagine simple services won't need them at all.
But I guess that is not the problem of the Design System. Or if it is, it's something for guidance, not code.

Overall I'm happy with this change.

The only thing I would like to get input from a content designer for is the label "Breadcrumb".
I would assume that the majority of people won't know what breadcrumbs are, including screen reader users. Or maybe screen reader users know the meaning of the word better than most other users because it is used on lots of websites to label them?
Unless "breadcrumb" explains it better to people (in this case specifically screen reader users), I would try to avoid jargon and explain in plain English what it is or what it does. I like "You are here", but a content designer might have a better idea.

selfthinker:
Also, directly related to #4451.

calvin-lau-sig7:
Thanks @selfthinker. I think I've got all the context. I’m not sure I can do better than W3C’s example that just calls it “Breadcrumbs”

Something like "You are here" is tricky because 'here' could be easily misinterpreted to mean any of the breadcrumb links.

If anything, there's the idea of "Page levels" or "Page levels for this page", but that's only slightly less jargon-y but also imprecise as it might be mistaken for something concerning the page itself and overemphasise its importance.

selfthinker:
@calvin-lau-sig7, I've tried to google alternative names and also if there has been any research about the name with screen reader users but couldn't find anything for the latter and only very little for the former.
Other than the previously mentioned "You are here" which is used in a couple of places, I also found "Navigation Path". But that's it, nothing else.
If I had to describe it with simpler words I would say something like "path to where you are on this website" (which is obviously too long).
I will ask the community, maybe someone knows of any user research in this area.

calvin-lau-sig7:
Thanks for the investigtion @selfthinker!

Continuing our musings, something like "Page level path" could be a bit more descriptive, as long as it's not mistaken for the user's back history.

I might need to better understand the semantics for "You are here"... would the ARIA tag be understood to refer to the whole list, without too much risk of being mistaken for one of the links in the list?

querkmachine:
We had similar thoughts when we introduced the pagination component and how we should accessibly label that. 

In that instance, we stuck with 'pagination' because, even though it's probably more niche terminology than necessary, it was the labelling that the overwhelming amount of other examples we could find also used for their accessible labels, so it seemed likely that assistive technology users would have encountered it many times before and probably understood what it referred to. 

That's just anec-data though, and we didn't specifically test that assumption.

selfthinker:
> I might need to better understand the semantics for "You are here"... would the ARIA tag be understood to refer to the whole list, without too much risk of being mistaken for one of the links in the list?

Yes, it would refer to the whole list, or rather to the navigation landmark the list is in. That can be discovered by navigating by landmark or arrowing through the page. It would then announce something like "Breadcrumb, navigation" before it gets into the list, so clearly not a link.
Although when directly navigating to a link in the breadcrumbs, it also announces the context, including the landmark. But it would do that in a way the context is clearly announced. Not sure if it might be possible for a less proficient screen reader user who is not used to landmarks and how they present to confuse them with being part of a link in that case. I don't think that would be very common.

selfthinker:
> It would then announce something like "Breadcrumb, navigation" before it gets into the list

Just clarifying... It would read "[whatever the aria-label says], navigation", so "You are here, navigation" in the other case. (And some screen readers also add "region" or "landmark" at the end.)

querkmachine:
The way that Breadcrumbs is currently often used has it only list the ancestors of the current page, but omits the current page (as that is given by the `h1`). That makes me unsure whether "You are here" would necessarily be an accurate description of the landmark's content.

romaricpascal (commented):
The component code looks good, as do the tests 🙌🏻 However, we've updated the breadcrumbs tests to use JSDOM in #5027 (sorry for not thinking of this PR while reviewing it), so the tests will need a little update before we can merge 😊

romaricpascal (approved):
> I've rebased from main and updated the new tests to use JSDOM 🎉

Ace! Thanks for rebasing. All good to go (code-wise, that is) ⛵

calvin-lau-sig7:
> The way that Breadcrumbs is currently often used has it only list the ancestors of the current page, but omits the current page (as that is given by the `h1`). That makes me unsure whether "You are here" would necessarily be an accurate description of the landmark's content.

Okay, @querkmachine's point here is a strong one – breadcrumb content often does not actually indicate the current page, instead only showing the earlier or higher-level pages.

Based on this I'd agree we should keep it to "Breadcrumb", as we can't rely on the breadcrumb being a reliable indication of "here". 

I'm open to figuring out a more plain English label amongst us that'd work through. E.g. How you got here (if that were accurate).

querkmachine:
I'm going to merge this given it's had approval and the consensus seems to be that current default labelling is sensible, even if not ideal. We can still update it later in the cycle without issue.

PR #4996, opened 2024-05-20
Replace custom Sass logger w/ silenceDeprecations

### LINKED ISSUE #4993: Replace custom Sass logger in our build pipeline with `silenceDeprecations` option
## What

As flagged by @colinrotherham in https://github.com/alphagov/govuk-frontend/pull/4985#discussion_r1605033243, we currently have a custom Sass logger that's designed to silence specific deprecation messages coming from Sass (introduced as part of #3164):

https://github.com/alphagov/govuk-frontend/blob/adfa0abc33b54262a08d9b2218cd8123a4b0ecf1/shared/tasks/styles.mjs#L137-L143

Sass introduced a new [silenceDeprecations option](https://sass-lang.com/documentation/js-api/interfaces/options/#silenceDeprecations) introduced in 1.74.0 which should allow us to remove (or at least simplify) the custom logger.

## Why

Using the new Sass feature will allow us to remove some custom code and provides a more robust solution (for example, matching correctly even if deprecation messages change).

## Who needs to work on this

Developers, developers, developers!

## Who needs to review this

Developers

## Done when

- [x] We've updated the Sass build pipeline to use the new `silenceDeprecations` option
- [x] The logger has been removed or simplified to no longer take care of silencing deprecations

36degrees:
Use the new `silenceDeprecations` feature introduced in Sass 1.74.0 and remove the custom logger that was implemented to filter them out.

The logger was introduced in 81f59d9 and ec547be as part of #3164.

As mentioned in the first commit message, the output otherwise exactly matches the log output from Dart Sass. I’ve also manually verified this by comparing the output from the custom logger with an empty array for `supressed` and using the default logged with an empty array for `silenceDeprecations`.

There’s nothing else using `chalk` which was introduced as a dependency in 81f59d9 so uninstall that too.

Closes #4993

### DIFF
--- a/shared/tasks/package.json
+++ b/shared/tasks/package.json
@@ -14,7 +14,6 @@
     "@govuk-frontend/lib": "*",
     "@npmcli/run-script": "^7.0.4",
     "@percy/puppeteer": "^2.0.2",
-    "chalk": "^5.3.0",
     "gulp": "^4.0.2",
     "gulp-cli": "^3.0.0",
     "nunjucks": "^3.2.4",
--- a/package-lock.json
+++ b/package-lock.json
@@ -31590,7 +31590,6 @@
         "@govuk-frontend/lib": "*",
         "@npmcli/run-script": "^7.0.4",
         "@percy/puppeteer": "^2.0.2",
-        "chalk": "^5.3.0",
         "gulp": "^4.0.2",
         "gulp-cli": "^3.0.0",
         "nunjucks": "^3.2.4",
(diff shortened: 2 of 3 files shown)


shared/tasks/package.json
```diff
@@ -14,7 +14,6 @@
     "@govuk-frontend/lib": "*",
     "@npmcli/run-script": "^7.0.4",
     "@percy/puppeteer": "^2.0.2",
-    "chalk": "^5.3.0",
```
romaricpascal, line 17: Good spot! 🦅

romaricpascal (approved):
Nice that the logging is back in the hands of Sass ⛵

PR #4999, opened 2024-05-23
ignore eslint for now

patrickpatrickpatrick:


romaricpascal (approved):
Let's give this a spin. ⛵ 

After merge, let's try and [trigger a dependabot run from the web UI](https://github.com/alphagov/govuk-frontend/network/updates/4430578/jobs).

Issue #5010, opened 2024-05-23
Migrate template tests from Cheerio to jsdom


36degrees:
## What

Update the existing template tests to use jsdom and the [matchers provided by `@testing-library/jsdom`](@testing-library/jest-do) rather than loading the HTML into a Cheerio instance.

Remove the dependency on Cheerio.

## Why

We currently render the template to HTML and then load it into a Cheerio instance which we can use to check the state of the HTML by using Cheerio methods like `hasClass`.

However, if we instead used jsdom (through jest-environment-jsdom) we can make use of the [matchers provided by `@testing-library/jsdom`](@testing-library/jest-do)).

Before:

```js
    it('renders with classes', () => {
      const $ = render('button', examples.classes)

      const $component = $('.govuk-button')
      expect($component.hasClass('app-button--custom-modifier')).toBeTruthy()
    })
```

After: 

```js
    it('renders with classes', () => { 
      document.body.innerHTML = render('button', examples.grey)

      const $component = document.querySelector('.govuk-button')
      expect($component).toHaveClass('app-button--custom-modifier')
    })
```

It'd also allow us to use other more useful matchers like `toHaveAccessibleDescription`, `toHaveName`, `toHaveRole` which are simpler and better convey intention than checking the association as we currently do:

Before:
```js
    it('associates the input as "described by" the hint', () => {
      const $ = render('input', examples['with hint text'])

      const $input = $('.govuk-input')
      const hintId = $('.govuk-hint').attr('id')

      const describedBy = new RegExp(
        `${WORD_BOUNDARY}${hintId}${WORD_BOUNDARY}`
      )

      expect($input.attr('aria-describedby')).toMatch(describedBy)
    })
```

After:
```js
    it('associates the input as "described by" the hint', () => {
      document.body.innerHTML = render('input', examples['with hint text'])

      const $input = document.querySelector('.govuk-input')
      expect($input).toHaveAccessibleDescription(
        "It’s on your National Insurance card, benefit letter, payslip or P60. For example, ‘QQ 12 34 56 C’."
      )
    })
```

Finally, given we aren't using jQuery anywhere else in the codebase, this also means that we only need to remember a single syntax, rather than switching between 'vanilla JavaScript' in our JavaScript and the jQuery-esque Cheerio syntax.

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [ ] **Breadcrumbs** template tests updated
- [ ] **Button** template tests updated
[...]

PR #5011, opened 2024-05-24
Add Jest test matchers for jsdom

36degrees:
Include the [@testing-library/jest-dom library](https://www.npmjs.com/package/@testing-library/jest-dom) which provides custom Jest matchers for asserting things about the state of the DOM, including things like what class an element has (`toHaveClass`), or the accessible description for an element (`toHaveAccessibleDescription`).

This as the first step in migrating our template tests from Cheerio to jsdom.

Part of #5010

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -18,6 +18,7 @@ module.exports = {
         'eslint:recommended',
         'plugin:import/recommended',
         'plugin:jest/style',
+        'plugin:jest-dom/recommended',
         'plugin:jsdoc/recommended-typescript-flavor',
         'plugin:n/recommended',
         'plugin:promise/recommended',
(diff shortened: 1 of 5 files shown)


shared/helpers/package.json
```diff
@@ -17,6 +17,8 @@
     "@axe-core/puppeteer": "^4.8.5",
     "@govuk-frontend/config": "*",
     "@govuk-frontend/lib": "*",
+    "@testing-library/jest-dom": "^6.4.5",
+    "@types/testing-library__jest-dom": "^6.0.0",
```
romaricpascal, line 21: **nitpick** Given the `jest.config.mjs` is at the root of the repository, I think those two should go in the root `package.json`. That's also where `jest` and its plugins are listed as dependencies. I think things work at the moment because Node [hoists packages to the `root` node_modules when it installs dependencies from workspaces](https://leticia-mirelly.medium.com/a-comprehensive-guide-to-npm-workspaces-and-monorepos-ce0cdfe1c625#979b).

romaricpascal (approved):
Neat! And nice find on the ESLint plugin as well  ⛵

PR #5017, opened 2024-05-28
Update input template tests to use jsdom

36degrees:
> [!NOTE]
> This PR is stacked on top of #5011 which must be merged first.

Also refactor the tests to:

- improve the test descriptions and grouping
- move duplicated code (like rendering the same example of a component) into `beforeAll` blocks
- use `toHaveAccessibleDescription` and `toHaveAccessibleName` where appropriate
- not use snapshots

Part of #5010.


packages/govuk-frontend/src/govuk/components/input/input.yaml
```diff
@@ -180,8 +180,6 @@ examples:
     options:
       label:
         text: National Insurance number
-      hint:
-        text: It’s on your National Insurance card, benefit letter, payslip or P60. For example, ‘QQ 12 34 56 C’.
```
romaricpascal, line 184: **question** Thinking that hint might have been there to check where the error would be rendered relative to the hint. By removing it, and not having a test for it, we're no longer checking that it goes 'label, hint, error, field', which I guess is important from a design perspective.

romaricpascal: I'd be keen to add that hint back and add a test that checks the order of elements inside the `govuk-form-group`, what do you say?

36degrees: There's a separate example which includes both an error message and a hint further down:

https://github.com/alphagov/govuk-frontend/pull/5017/files#diff-64dabf8e89e534dc5f047bb2159938a010e94c221cc75b9b7196d04dcb8a5ceeR451-R461

With a test for the ordering of aria-describedby here:

https://github.com/alphagov/govuk-frontend/pull/5017/files#diff-54a0f41b55758264a220e2eb6e3fba6d181769d8541a05acbe05a1c0963b46bcR241-R261

I think it would make sense to make that example visible though, I'll update it to match this example before I removed the hint.

36degrees: > I'd be keen to add that hint back and add a test that checks the order of elements inside the govuk-form-group, what do you say?

Do you mean the order they appear in the description, or in the DOM?

packages/govuk-frontend/src/govuk/components/input/template.jsdom.test.js
```diff
@@ -0,0 +1,496 @@
...
+    })
+
+    it('sets the `value` attribute based on the `value` option', () => {
+      document.body.innerHTML = render('input', examples.value)
+
+      const $component = document.querySelector('.govuk-input')
+      expect($component).toHaveValue('QQ 12 34 56 C')
+    })
+
+    it('sets the `value` attribute based on a `value` of "0"', () => {
```
romaricpascal, line 99: **nitpick** I think the issue was with a `0` value (falsey, made the attribute not render) rather than a string of `"0"` (truthey).

```suggestion
    it('sets the `value` attribute based on a `value` of `0`', () => {
```

If there's an issue with both, we should probably have two separate tests 😊

packages/govuk-frontend/src/govuk/components/input/template.jsdom.test.js
```diff
@@ -0,0 +1,496 @@
...
+    it('associates the input as described by both the hint and the `describedBy` option', () => {
+      document.body.innerHTML = render(
+        'input',
+        examples['hint with describedBy']
+      )
+
+      const $additionalDescription = document.createElement('p')
+      $additionalDescription.id = 'test-target-element'
+      $additionalDescription.textContent = 'Additional description'
+      document.body.appendChild($additionalDescription)
```
romaricpascal, line 162: **question** Is this necessary for `toHaveAccessibleDescription` to function properly (finding the element with the relevant ID and computing its accessible name for use as a description, which it cannot do if the element is not in the DOM)?

romaricpascal (approved):
Neat, seems all test are there 🙌🏻 I'd say we can add tests for the DOM order at a later stage, I can add an issue for that.

Thinking we could update the tests like `it('includes the hint')` and similar tests to become tests checking the DOM order with something like `expect($component.matches('.govuk-form-group .govuk-hint ~ .govuk-input')).toBe(true)`, a bit like what happens for checking the wrapper is here when prefix and suffix are used.

Nothing blocking for merging, but would be good if you could fix the quotes in the test for the 0 value (unless I'm mistaken) and document the reason for the addition of the paragraph hosting the description (a thumbs up on my question if my supposition was right should be enough of a trace for the future 😊 ), cheers!.

packages/govuk-frontend/src/govuk/components/input/template.jsdom.test.js
36degrees: Yes, if we want to be able to use `toHaveAccessibleDescription` and be confident that the external element referenced by the `describedBy` option is included.

The alternative would be to assert against the value of the `aria-describedby` attribute itself (like the old tests did) but I felt it was better to have just one way of doing it?

romaricpascal: I agree, it's much nicer to use the 'toHaveAccessibleDescription' rather than look at the `aria-describedby. Just wanted to check I understood why that paragraph was there 😊

36degrees:
[Force-pushed a change](https://github.com/alphagov/govuk-frontend/compare/3be210a8c5809dc3a88af2365c19d0073c37fde4..f751c3be242d4a1e13e3852ade72dd0f70e5fa45) to remove the quotes around zero to resolve https://github.com/alphagov/govuk-frontend/pull/5017#discussion_r1620820751.

romaricpascal:
@36degrees Looks good to go ⛵

PR #5040, opened 2024-06-03
Update README.md

Nooshu:
Minimal change to the README.md. Currently, it mentions WCAG 2.1 but when you click on the link it goes to a WCAG 2.2 page.


36degrees (approved):
Good spot, thanks @Nooshu! This just needs a second review from the team and then we can get it merged.

romaricpascal (approved):

PR #5041, opened 2024-06-03
Define additional screenshots in component YAML

36degrees:
Rather than hardcoding the list of additional examples to screenshot in the task, add a new `screenshot` boolean to the component YAML (and generated fixtures.yaml) which determines whether the example should be included in the screenshots or not.

Preserve the existing list of screenshotted examples by setting `screenshot: true` in the YAML for those examples.

Remove special treatment of examples named ‘inverse’ but retain their screenshots by also setting `screenshot: true`.

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/back-link/back-link.yaml
+++ b/packages/govuk-frontend/src/govuk/components/back-link/back-link.yaml
@@ -29,6 +29,7 @@ examples:
       href: '#'
       text: Back to home
   - name: inverse
+    screenshot: true
     previewLayoutModifiers:
       - inverse
     options:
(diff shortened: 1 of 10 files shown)


36degrees:
The tests are failing because the snapshot names are different – the screenshots used to use the 'kebab case' example names hard coded in the task, and now they're using the examples names as they appear in the YAML.

So Percy is showing this as 6 missing screenshots and 6 new screenshots.

Do we think it's important to maintain the existing screenshot names? It *might* cause some PRs to fail if they're not up to date with the changes introduced in this PR.

If not, we can simply approve the changes in Percy and move on.

owenatgov (approved):

PR #5043, opened 2024-06-04
Refactor the accordion JavaScript

36degrees:
I've given the accordion JavaScript a bit of TLC and fixed a few things that have been bugging me ✨ 

(There are a few more things I'd like to do that are a bit 🌶️ spicy so I'll raise them as separate PRs stacked on this one.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ For advice on how to use these release notes see [our guidance on staying up to 
 
 ## Unreleased
 
+### Fixes
+
+We've made fixes to GOV.UK Frontend in the following pull requests:
+
+- [#5043: Refactor the accordion JavaScript](https://github.com/alphagov/govuk-frontend/pull/5043)
+
 ## 5.4.0 (Feature release)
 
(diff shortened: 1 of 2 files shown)


owenatgov (approved):

owenatgov:
I approved but I had a question I forgot to include: Do we think this change warrants a changelog entry? It does technically impact user-facing code but it's just a tidyup rather than an explicit bugfix.

romaricpascal:
Looking at the CHANGELOG, we do have a few refactoring mentioned in 'Fixes' sections (like [the first item in 4.4.0](https://github.com/alphagov/govuk-frontend/blob/main/CHANGELOG.md#fixes-10)). Thinking it's helpful to mention it in the changelog: in case an unforeseen bug crops up in the accordion, the entry may give pointers to the person looking into what's happening. @36degrees Would you mind adding a 'Fixes' entry, please ? 😊

romaricpascal (approved):
@36degrees Cheers! I think it'll need a little rebase for Percy and then good to go!

PR #5044, opened 2024-06-04
Remove session storage checks from accordion JavaScript

36degrees:
We don’t do anything different if we detect the browser doesn’t support session storage, so we don’t really need to care whether it works or not.

We do need to wrap attempts to set *or get* from sessions storage in try blocks because setItem [can throw `QuotaExceededError` exceptions][1] and some [older browsers will throw other exceptions if session storage isn’t available (disabled, or using private browsing)][3]

However we don’t need to do anything different if exceptions are thrown – we can safely catch those exceptions and swallow them.

[1]: https://html.spec.whatwg.org/multipage/webstorage.html#dom-storage-setitem
[2]: https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API/Using_the_Web_Storage_API#feature-detecting_localstorage
[3]: https://gist.github.com/paulirish/5558557

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ For advice on how to use these release notes see [our guidance on staying up to 
 We've made fixes to GOV.UK Frontend in the following pull requests:
 
 - [#5043: Refactor the accordion JavaScript](https://github.com/alphagov/govuk-frontend/pull/5043)
+- [##5044: Remove session storage checks from accordion JavaScript](https://github.com/alphagov/govuk-frontend/pull/5044)
 
 ## 5.4.0 (Feature release)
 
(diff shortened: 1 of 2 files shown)


querkmachine (approved):
This looks good to me.

I have a mild recollection, somewhere in the recesses of the mind, that `localStorage`/`sessionStorage` can—in some circumstances—appear to be available APIs, but they will refuse to store or retrieve anything if you actually try to use it. 

That was probably due to some sort of private browsing mode. I imagine this is why the helper function explicitly tested for this scenario.

However, I'm okay with that being removed in this situation as the fail state would be that the expanded sections don't get remembered, which feels acceptably still-functional.

PR #5046, opened 2024-06-05
Skip ‘empty’ tasks in the task list

### LINKED ISSUE #4968: Task list component should check if item is empty before rendering.
https://github.com/alphagov/govuk-frontend/blob/8729af48207d7513218840669aefcc0271db56c2/packages/govuk-frontend/src/govuk/components/task-list/template.njk#L39

The task list component currently renders a line for each item - it should ideally be checking for empty / falsy items and skipping if so. This is relevant as when prototyping you might often do an `if` statement to decide whether to show a row - and currently the component renders weirdly in this case as you are rendering an empty row.

I think this is handled by something like `{% if foo %}` in [other components](https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/src/govuk/components/summary-list/template.njk#L59).

36degrees:
> [!NOTE]
> This PR is stacked on top of #5045 which must be merged first.

This allows users to e.g. guard individual tasks with an `if` statement without empty rows displaying.

Closes #4968

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ For advice on how to use these release notes see [our guidance on staying up to 
 
 We've made fixes to GOV.UK Frontend in the following pull requests:
 
+- [#5046: Skip ‘empty’ tasks in the task list](https://github.com/alphagov/govuk-frontend/pull/5046)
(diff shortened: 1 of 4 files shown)


edwardhorsford:
Looking good! thanks.

romaricpascal (approved):
⛵

PR #5050, opened 2024-06-07
Update hint template tests to use jsdom

36degrees:
Part of #5010.


romaricpascal (approved):
Looks all good to me ⛵

PR #5060, opened 2024-06-10
Reintroduce additional bottom margin to Error Summary content

### LINKED ISSUE #4997: Add margin below final element in the Error Summary
## What

Add back a margin after the final element of the content in the Error Summary (either the final error in the list or the description).

Note: it might be that the margin does not necessarily need to be added back on the `<li>` it was removed from. The content is enclosed in the `govuk-error-summary__body`. 

## Why

We recently removed a 5px margin under the final `<li>` element in #4971, but it turns out that t[he extra margin can be helpful to account for descenders of letters like 'g', 'y', 'p'](https://github.com/alphagov/govuk-frontend/pull/4971#issuecomment-2119941900)...

## Who needs to work on this

Developers, designers

## Who needs to review this

Developers, designers

## Done when

- [ ] Margins have been added back at the end of Error summary's content

querkmachine:
Reverses a visual change introduced in #4971 where the negative space at the bottom of the Error Summary component was reduced to create equal padding on all sides of the container.

Although this was originally approved of by a designer on the team, further discussion after the 5.4.0 release concluded that having additional negative space below the Error Summary's content was desirable, as it gave extra space for text descenders to occupy (if present).

This PR reintroduces that space, but does so in a way that compatible with the new configuration variants made possible by #4971. (Originally, this additional spacing only existed if the `errorList` parameter was being used, whereas this is no longer a requirement.) 

Closes #4997.


romaricpascal (commented):
Code change looks good to me! @mia-allers-gds (or @Ciandelle, @CharlotteDowns) would you be able to give a quick 'designer approved' stamp on this one as well (the value added back is the same as the one that was there before, which was the [bottom margin of the list](https://github.com/alphagov/govuk-frontend/blob/e2e0d9a93d1a813b71eaa29fc798a5d6fe74e496/packages/govuk-frontend/src/govuk/core/_lists.scss#L20))  😊

mia-allers-gds:
@romaricpascal I *think* it looks good!

romaricpascal (approved):
> @romaricpascal I think it looks good!

Let's merge it! ⛵

PR #5061, opened 2024-06-10
Standardise spelling to "falsy"

domoscargin:
I noticed a typo in some of our tests using `falsely` instead of `falsey`. Then I noticed we use both falsy and falsey across the codebase.

There were more examples of `falsy`, and we have methods using this spelling. It also seems to be [generally preferred](https://developer.mozilla.org/en-US/docs/Glossary/Falsy).


owenatgov:
Before I review, I wanted to add some [extra anecdotal evidence that "falsy" is the way to go](https://english.stackexchange.com/questions/109996/is-it-falsy-or-falsey)

owenatgov (approved):
EZ

PR #5063, opened 2024-06-11
Refactor `govuk-visually-hidden` Sass

### LINKED ISSUE #1597: govuk-visually-hidden-focusable mixin incorrectly sets margin to inherit when focused
https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/helpers/_visually-hidden.scss#L73

I think this is trying to 'undo' the `margin: 0` applied when not focused.

However, I don't think this makes sense – this means that if you have a margin on your `<body>` then e.g. the skip link will inherit that margin too, as seen in https://github.com/alphagov/paas-admin/issues/426

querkmachine:
Review app: https://govuk-frontend-pr-5063.herokuapp.com/

A neat little optimisation that we can probably make now that we've dropped support for older IE versions.

Currently, we have two mixins that share mostly identical styles: `govuk-visually-hidden` and `govuk-visually-hidden-focusable`, with a few of differences: 

1. `govuk-visually-hidden` uses pseudo-elements to add spaces before and after the visually-hidden content.
2. `govuk-visually-hidden-focusable` applies the visually-hidden styles by default, but then attempts to undo these styles when the element receives keyboard focus or is in the active state.
3. `govuk-visually-hidden-focusable` additionally omits a couple of the visually-hidden styles on the basis of not wanting to undo them upon focus or active states.

Having two almost-identical mixins isn't ideal when it comes to code duplication. Trying to manually revert styles upon a particular state is also quite untidy and evidently is something we've had problems doing before, given we've needed to explicitly omit some styles to not unintentionally revert the styles we _do_ want. 

Now that we've dropped support for Internet Explorer 8 and 9, we have the option of using the `:not` selector to only apply the visually-hidden styles when the element is not focused or active, allowing both mixins to use the same set of core styles and allowing for the complete removal of the manual revert code. 🛹

This would close #1597. 

## Changes
* Moves styles common to both mixins into an internal helper mixin.
* Changes `govuk-visually-hidden-focusable` to only apply the visually-hidden styles when the target element is `:not` focused or active, rather than setting the styles by default and then overriding them.

## Thoughts
This uses two chained `:not` selectors to achieve the desired result: `:not(:active):not(:focus)`. This could be just one selector, `:not(:active, :focus)`, but IE 11 doesn't support this syntax and PostCSS doesn't appear to be converting it to the syntax that IE 11 understands. 

I've named the internal mixin `_govuk-visually-hidden-text`. This is technically a misnomer as it could hide any element or set of elements, but simply calling it `_govuk-visually-hidden` seemed too likely to cause confusion. Alternative naming suggestions welcome.

### DIFF
--- a/packages/govuk-frontend/src/govuk/helpers/_visually-hidden.scss
+++ b/packages/govuk-frontend/src/govuk/helpers/_visually-hidden.scss
@@ -2,7 +2,7 @@
 /// @group helpers/accessibility
 ////
 
-/// Hide an element visually, but have it available for screen readers
+/// Helper function containing the common code for the following two mixins
 ///
 /// @link https://snook.ca/archives/html_and_css/hiding-content-for-accessibility
 ///   - Hiding Content for Accessibility, Jonathan Snook, February 2011
@@ -11,22 +11,10 @@
 ///
 /// @param {Boolean} $important [true] - Whether to mark as `!important`
 ///
-/// @access public
+/// @access private
 
-@mixin govuk-visually-hidden($important: true) {
+@mixin _govuk-visually-hide-content($important: true) {
   position: absolute if($important, !important, null);
-
-  // Absolute positioning has the unintended consequence of removing any
-  // whitespace surrounding visually hidden text from the accessibility tree.
-  // Insert a space character before and after visually hidden text to separate
-  // it from any visible text surrounding it.
-  &::before {
-    content: "\00a0";
-  }
-
-  &::after {
-    content: "\00a0";
-  }
 
   width: 1px if($important, !important, null);
   height: 1px if($important, !important, null);
@@ -36,6 +24,8 @@
   padding: 0 if($important, !important, null);
 
   overflow: hidden if($important, !important, null);
+
+  // `clip` is needed for IE11 support
   clip: rect(0 0 0 0) if($important, !important, null);
   clip-path: inset(50%) if($important, !important, null);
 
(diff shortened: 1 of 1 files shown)


romaricpascal (commented):
Looking good to me, and a welcome refactoring on top of fixing the issue 🙌🏻 For a more generic suggestion `_govuk-visually-hidden-content` might do, but I don't think there's a big risk of confusion in using `text` there.

Had a look with Safari and VoiceOver and the Skip link is showing OK when getting focused, so I'd say things look OK in terms of the new selector doing its job. @selfthinker would you prefer having further testing with a wider range of browser/assistive tech combinations? As @querkmachine detailed, the new code is not using anything wild CSS wise, only plain `:not()` selectors which have been around for [a long while now](https://caniuse.com/?search=%3Anot), so this feels fairly safe.

querkmachine:
I've renamed the internal mixin to `_govuk-visually-hide-content`, to make it a bit more distinct from the others. Helps to understand it by making it more verb-y too, I guess?

selfthinker:
> @selfthinker would you prefer having further testing with a wider range of browser/assistive tech combinations?

I would prefer to test more. That's mostly because I know we had issues with the visually hidden class before that would only happen in a specific screen reader / browser combination. Better to be safe than sorry.

Where can I best test this?

romaricpascal:
> Where can I best test this? I can do that tomorrow.

@querkmachine may have better options

The `govuk-visually-hidden-focusable` mixin is used by [the Skip Link](https://govuk-frontend-pr-5063.herokuapp.com/components/skip-link) (for comparison, here's the [Skip Link example for `main`](https://govuk-frontend-review.herokuapp.com/components/skip-link))

We have quite a few [applications of the `govuk-visually-hidden` class in the different components and examples](https://github.com/search?q=repo%3Aalphagov%2Fgovuk-frontend+govuk-visually-hidden+language%3ANunjucks&type=code&l=Nunjucks). 
- Thinking something like the [Bank Holidays example](https://govuk-frontend-pr-5063.herokuapp.com/full-page-examples/bank-holidays) may be interesting as it's used to hide the table headings (which would be announced when users navigate the table). 
- For a simpler example where the content is not referenced by anything, the [Error Message component](https://govuk-frontend-pr-5063.herokuapp.com/components/error-message) has a visually hidden 'Error' before the actual message. Again for comparison here are the [Bank Holidays example for `main`](https://govuk-frontend-review.herokuapp.com/full-page-examples/bank-holidays) and the [Error Message component for `main`](https://govuk-frontend-review.herokuapp.com/components/error-message)

selfthinker:
>> Where can I best test this? I can do that tomorrow.

@romaricpascal, sorry, I meant where on the internet. Is this PR deployed somewhere?
I could potentially also copy the CSS and create an example page that uses it in several different scenarios.

querkmachine:
https://govuk-frontend-pr-5063.herokuapp.com/

There's an automated deployment under the 'View deployment' button (currently) a few comments above this one. GitHub's interface be at it again.

selfthinker:
> There's an automated deployment under the 'View deployment' button

I thought it would be deployed automatically, and I was looking for such a link or button. But either I didn't see it or I didn't understand that that is what it did. Thanks for pointing that out.

selfthinker:
I've just tested the three examples suggested by Romaric and they behave exactly like before, so everything looks good to me. :+1:

(And while I was doing that I noticed that it's not a good idea to visually hide table headers as most screen readers ignore them when navigating data cells. But that was already an issue before and doesn't have anything to do with this implementation.)

romaricpascal (approved):
With the accessibility testing being OK, I think we're good to merge that one 🙌🏻 ⛵

Issue #3554, opened 2023-04-26
Pagination in stacked mode relies on whitespace for correct text alignment


matteason:
## Description of the issue

When in stacked (block) mode (no page numbers displayed), the pagination component relies on whitespace in the HTML to align the 'Previous' and 'Next' text with the labels below.

If there's no whitespace between the `govuk-pagination__icon` SVG and the `govuk-pagination__link-title` span, the 'Previous' or 'Next' link is displayed about 4 pixels too far to the left.

With whitespace:

![Previous and Next links stacked on top of each other, each with an example label below them and an arrow to the left. The text and labels are almost aligned vertically, demonstrated with a red line](https://user-images.githubusercontent.com/1935173/234560418-fd1fb657-711b-4a63-8f90-1c7e5cd28ce5.png)

Without whitespace:

![Previous and Next links stacked on top of each other, each with an example label below them and an arrow to the left. The text and labels are unaligned, demonstrated with a red line](https://user-images.githubusercontent.com/1935173/234560463-ab953204-9b55-4626-a917-acc3fc84e6a9.png)

(The text is actually slightly misaligned even in the first version - 'Previous' aligns with the end of the underline on the label rather than the first letter - I'm not sure if this is intentional)

## Steps to reproduce the issue

Open [this Design System pagination example](https://design-system.service.gov.uk/components/pagination/labels-2/index.html) in Firefox, right-click > Inspect the link, expand the `<a>` tag, click the whitespace node and press Delete

![HTML code from GOV.UK pagination component](https://user-images.githubusercontent.com/1935173/234561589-64a7dcd4-ce1d-4ee7-b691-55eaf488c52c.png)

## Actual vs expected behaviour

I would expect the text alignment to not rely on whitespace in the HTML. In my particular case (a [Vue component library](https://github.com/govuk-vue/govuk-vue)) I have no control over whether consumers of the library will have [the whitespace compression option](https://vuejs.org/api/application.html#app-config-compileroptions-whitespace) enabled or not, though by default it's enabled

There's been [a previous issue with whitespace causing layout changes in the header component](https://github.com/alphagov/govuk-frontend/issues/828), and there appears to be a fix for a similar issue when pagination is used in list mode:

https://github.com/alphagov/govuk-frontend/blob/eed94f8b29072325542f2ca4c4d0ffec6c4c543d/src/govuk/components/pagination/_index.scss#L57-L61

## Environment (where applicable)

- Operating system: Windows 10
[...]

PR #5066, opened 2024-06-13
Fix whitespace affecting text alignment in pagination block variant

querkmachine:
An attempt at fixing the issue where whitespace (or the lack of it) changes the alignment of the title text in the Pagination component's 'block' variant, as described in #3554.

## Changes
- Changes block pagination links to be `inline-block`, allowing for the addition of `padding-left` that affects all text within the component.
- Changes the block pagination arrows to be floated and offset using a negative `margin-left`—preventing them from directly influencing text flow calculations.

### Focus style difference
This PR does slightly change the focus state on these links, with the focus state now stretching to fit the longest line of text, rather than separately 'shrinkwrapping' the title and label lines. 

|Before|After|
|:-:|:-:|
|Screenshot of some links, with the first one focused. The focused text has a yellow background that follows the length of each line of text and a thick black border under the last line.|Screenshot of some links, with the first one focused. The top link has a rectangular yellow background that is sized according to the longer second line, with a thick black border underneath it.|

I was unable to find a simple way around this using only CSS and wanted to avoid a breaking HTML change. 

I considered removing the focus style from the actual link and re-applying it to the two child elements, but doing so introduced more complexity than I'd like, so I'm gonna 🤞 and hope this is okay until we're content with making a breaking change.

### 0.326 is the magic number

This also introduces a magic number, the `margin-top` on the arrows themselves. Being floated means they no longer automatically align with the text baseline, so some manual offsetting was required.

This could potentially be worked out programmatically: it's $(lineHeight-arrowHeight)\div2$, specifically $(25-13)\div2=6px$ with numbers added, which is ≅0.326em in this context.

However, we don't presently have a function for getting just the `line-height` for a specific type size and breakpoint, and I was still trying to avoid adding unnecessary complexity, and the pagination's `line-height` doesn't change per breakpoint anyway, so I've hardcoded it. This value is precise enough to visually match the arrow's previous position on even high zoom levels. 

A two-frame animation showing before and after screenshots. In the before, the first line of title text is slightly aligned too far to the left and the second line is entirely to the left. In the after, both lines are aligned with the label text below it. Neither the arrow nor label text move between the before/after.

These comparison screenshots were made in Firefox at 300% zoom on a 72 DPI display.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ For advice on how to use these release notes see [our guidance on staying up to 
 
 ## Unreleased
 
+### Fixes
+
+We've made fixes to GOV.UK Frontend in the following pull requests:
+
+- [#5066: Fix whitespace affecting text alignment in pagination block variant](https://github.com/alphagov/govuk-frontend/pull/5066)
+
 ## 5.4.1 (Fix release)
 
 ### Recommended changes
--- a/packages/govuk-frontend/src/govuk/components/pagination/_index.scss
+++ b/packages/govuk-frontend/src/govuk/components/pagination/_index.scss
@@ -151,7 +151,6 @@
     @include govuk-typography-weight-regular;
     @include govuk-link-decoration;
     display: inline-block;
-    padding-left: govuk-spacing(6);
   }
 
   .govuk-pagination__icon {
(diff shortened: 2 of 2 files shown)


owenatgov (commented):
Code is looking good.

If memory serves, all of these dodgy attributes were in service of trying to shrink wrap our focus state. Given that it appears to be causing problems elsewhere, I'm perfectly happy to drop it.

I'm gonna pause at approving as I think it's worth getting a designer to cast one final eye over the focus state change to verify that it's ok.

A final thought is that I wonder if there's an opportunity here to refactor pagination, get rid of the annoying floats etc. I'm very aware that you [started this](https://github.com/alphagov/govuk-frontend/pull/3644) but it didn't get lots of traction. Maybe it's something we can revisit and hopefully actually get done 🤞🏻

mia-allers-gds (approved):
Visually it looks much better!

querkmachine:
> A final thought is that I wonder if there's an opportunity here to refactor pagination, get rid of the annoying floats etc. I'm very aware that you [started this](https://github.com/alphagov/govuk-frontend/pull/3644) but it didn't get lots of traction. Maybe it's something we can revisit and hopefully actually get done 🤞🏻

That refactor was more concerned with altering the functionality, in part to try and reduce complexity regarding the behaviour of ellipses and options hiding on mobile should behave. It's fairly standalone from what is happening in this PR.

owenatgov (approved):
Percy approved as well 👍🏻

PR #5067, opened 2024-06-14
[SPIKE] Add JS property observer function; allow for programatic updating of character count

querkmachine:
Some experimental code dabbling into one way we could resolve the old, old problem of form controls with associated JS not responding to programatic value changes. 

This is only one possible approach. In the past we've discussed creating an explicit JavaScript API for changing those values, but this hasn't yet come to fruition.

## Changes
- Introduces a new helper function, `observeElementProperty`, which overrides an HTML element's native property setter function to inject a customisable callback function, allowing us to invoke custom code when the property's value is changed.
- Add some (currently very minimal) tests for this function.
- Updates the Character Count component to use this function, updating the counters in response to programatic changes to the textarea's `value` property.

I've devtested this in all contemporary major browser engines and back to Safari 11 without issue.

## Limitations

As this works by overriding the property's native setter function, this solution only works for direct manipulation of a property value. It does not work if the property value is passively updated by some other code.

```js
observeElementProperty($input, 'disabled', () => {…})
observeElementProperty($input, 'value', () => {…})

// These will invoke the callback function
$input.disabled = true
$input.value = 'hello world'

// These will not invoke the callback function, even though changing these attributes also updates the underlying properties
$input.removeAttribute('disabled') 
$input.setAttribute('value', 'hello world')
```

Property setters and getters come as a pair. Because we're overriding the setter, we must also redefine the getter. In this case, I've just manually re-invoked the native getter. 

I've not done extensive testing on this whatsoever. There is a possibility that by overriding the native property handling that something somewhere else may become broken as a result.

### DIFF
--- a/packages/govuk-frontend/src/govuk/common/observe-element-property.jsdom.test.mjs
+++ b/packages/govuk-frontend/src/govuk/common/observe-element-property.jsdom.test.mjs
@@ -1,0 +1,31 @@
+import { observeElementProperty } from './observe-element-property.mjs'
+
+describe('observeElementProperty', () => {
+  it('fires a callback function when a property value is updated', () => {
+    const $element = document.createElement('button')
+    let callbackCalled = false
+
+    observeElementProperty($element, 'disabled', () => (callbackCalled = true))
+    $element.disabled = true
+
+    expect(callbackCalled).toBeTruthy()
+  })
+
+  it('returns the values of properties unaltered', () => {
+    // This test is not directly related to the function, but as we have to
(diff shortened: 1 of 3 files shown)


romaricpascal:
> This was an interesting experiment but I think our plan for the future is to create public methods specifically for cases like this, rather than relying on passive observers. Closing.

I don't think both are exclusive: even if we have a public method for setting fields values and recomputing we may still find the need to observe the element for people that'd plainly do `document.querySelector(...).value = 'something'` (possibly because of a library they're using), so what you proposed here may still be of use 😊

PR #5068, opened 2024-06-14
Update _grid.scss to fix typo

edwardhorsford:
From reading [_grid.scss](https://github.com/alphagov/govuk-frontend/blob/520f2345aed725b2c29f14af2cdffcd75a5e130a/packages/govuk-frontend/src/govuk/objects/_grid.scss#L20), I think the generated class name is 'from', not 'at'.


36degrees (approved):
I think this is just mean to be an example of how to use the mixin, but I agree it'd probably make sense to make it consistent anyway.

romaricpascal (approved):

edwardhorsford:
Thanks both!

PR #5070, opened 2024-06-14
Fix alignment of content in conditional checkboxes and radio buttons

### LINKED ISSUE #5069: Content of conditionally revealed checkboxes and radio buttons is misaligned
## Description of the issue

The border and text of conditionally revealed content in Checkboxes and Radios is misaligned with respect to the input and label, appearing slightly too far to the left. 

This is a visual regression that seems to have crept into one of the 5.x releases of Frontend. This issue was not present in older versions. 

Reported by @MartinJJones via Slack.

View [expanded checkboxes](https://govuk-frontend-review.herokuapp.com/components/checkboxes/with-conditional-item-checked/preview) or [expanded radios](https://govuk-frontend-review.herokuapp.com/components/radios/with-conditional-item-checked/preview) in any browser or screen size.

Screenshot of some checkboxes. The first one is checked, revealing an additional label and text input below it. The border, label and text input are all aligned a few pixels further left than the input and label above.

Screenshot of some radio buttons. The first one is checked, revealing an additional label and text input below it. The border, label and text input are all aligned a few pixels further left than the input and label above.

The border and text should be aligned.

- Operating system: All
- Browser: All
- Browser version: All
- GOV.UK Frontend Version: 5.4.0

querkmachine:
Checkboxes and radio buttons are visually 40 pixels wide, however their actual width on the page is 44 pixels, with an invisible 2 pixel 'border' around the input that the conditional area's margin-left calculation didn't account for.

Fixes #5069.

## Changes
- Amends the `$conditional-border-padding` calculation in both components to use `$govuk-touch-target-size` instead of `$govuk-checkboxes-size` or `$govuk-radios-size`.
- Adds conditional checkboxes and radios with expanded content to the list of things Percy should screenshot, to help avoid future regressions of this kind.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ We've made fixes to GOV.UK Frontend in the following pull requests:
 - [#5043: Refactor the accordion JavaScript](https://github.com/alphagov/govuk-frontend/pull/5043)
 - [#5044: Remove session storage checks from accordion JavaScript](https://github.com/alphagov/govuk-frontend/pull/5044)
 - [#5060: Reintroduce additional bottom margin to Error Summary content](https://github.com/alphagov/govuk-frontend/pull/5060)
+- [#5070: Fix alignment of content in conditional checkboxes and radio buttons](https://github.com/alphagov/govuk-frontend/pull/5070)
 
 ## 5.4.0 (Feature release)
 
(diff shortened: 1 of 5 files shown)


36degrees (approved):
Needs a changelog entry, but otherwise makes sense to me. Good shout to add visual regression tests – thanks for doing that.

Does make me wonder when this changed 🤔

PR #5082, opened 2024-06-17
Change Safari 10.1 to Safari 11 in Grade C browser example

querkmachine:
The summary currently includes Safari 10.1 for macOS as an example of a Grade C browser.

However, in the Grade C section of the document, it's mentioned that Safari 10.1 will not run our JavaScript as (although it technically supports `type="module"`) it fails the prerequisite support tests present in Frontend. That section lists Safari 11 as being Grade C instead.

This PR updates the summary on the presumption that Safari 10.1 is actually Grade X due to not receiving any JavaScript enhancements, and Safari 11 is the oldest macOS version that classifies as Grade C.


domoscargin:
The problem is we've defined Grade C as everything not in A or B that supports `<script type="module">`, which includes Safari 10.1, even if we do the `noModule` check before it can do anything.

Because the `type` attribute is present, older (Grade X) browsers [should not process](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#any_other_value) the `<script>` block. Safari 10.1, however, will process the `<script>` block, and thus we have to ensure the script doesn't break it.

querkmachine:
Truuuuue. I suppose Safari 10.1 is just an annoying special case. From our perspective, it's JS enabled but subsequently has everything turned off (including "necessary enhancements"), whereas from a user perspective it just doesn't appear to run JS at all.

I still think the documentation is a little confusing as-is, as the [Grade C section](https://github.com/alphagov/govuk-frontend/blob/f4913af03e6210dc8a264b57f5b7ed4f44c81038/docs/contributing/browser-support.md#grade-c) appears to list Safari 11 as the minimum version that's in Grade C, contrary to the summary at the top of the page. It tripped me up, at least!

domoscargin:
I wonder if we should add some kind of link to documentation in this error message: https://github.com/alphagov/govuk-frontend/blob/56094f0efbcf316dce42f377cecb08463b6ea002/packages/govuk-frontend/src/govuk/errors/index.mjs#L40

But not sure how many people who get into this specific case will be checking the console.

As for the documentation, maybe we can add a "special cases" heading or something in the Grade C section so that it's clearer there's some niche case?

romaricpascal:
Thinking the issue might be with the definition of what we're supporting. We document that we run in browsers that support `<script type="module">` where actually, we make the cut at browser that support `<script nomodule>`. I _think_ that if we clarify the feature we base the support on, we could clarify to Safari 11 without ambiguity.

Maybe with either:
1. 'browsers that support `<script nomodule>`', but that's a bit cryptic
2. 'browsers that [fully support ES6 Modules](https://caniuse.com/es6-module)'

domoscargin:
I like the second option for simplicity. Then we could simplify the Grade C section documentation:

```
## Grade C

This grade covers browsers not in Grade A or B which [fully support <script type="module">](https://caniuse.com/es6-module). These are:

- Chrome 61+
- Edge 16-18
- Edge 79+
- Safari 11
- Firefox 60+
- Opera 48+
- Samsung Internet 8.2+
```

querkmachine:
Coincidentally, this was brought up on support today. It was questioned why [the Frontend Docs currently state](https://frontend.design-system.service.gov.uk/browser-support/#browser-support) Safari 10.3 as the minimum supported version when Safari 10.3 appears to behave the same as 10.1 — JS is loaded but no components are initialised.

In my mind, bumping the 'headline' version of Safari supported in Grade C to 11.0 would make the documentation easier to understand at a glance. 10.1–10.3 can be in the footnote. 

I'll update my suggested edits shortly.

romaricpascal (approved):
Sounds good to merge for me. Let's make sure we co-ordinate with the merge of the same change on the frontend docs so we don't have two different wordings in different places.

Issue #5085, opened 2024-06-18
Add new review app example pages in preparation for external audit


querkmachine:
## What

Add new full page examples to the Frontend review app so that external auditors can test our styles, components and patterns in the context of full, functional pages.

## Why

We produced lots of new full page examples for our last external audit (https://github.com/alphagov/govuk-frontend/pull/3503), however these remained separated from the main project and are now outdated, and likely incompatible with the architectural and tooling changes made in subsequent versions of Frontend.

This time around we intend to make these examples permanent additions to the review app, so that they can be remain contemporary with Frontend. 

We will only be recreating a subset of the examples, based on which ones were [flagged for improvement in the last audit](https://docs.google.com/spreadsheets/d/1qiWCRwH7QofQgJ9iDLRSsfETLKlZCyrv7z04qjZurhM/edit?pli=1&gid=0#gid=0) [internal document]. 

## Who needs to work on this

- Developers

## Who needs to review this

- Developers
- Content writers(?)
- Designers(?)
- Accessibility specialist(?)

## Done when

- [ ] Examples from the previous audit have been recreated in the current review app
  - [ ] ~~'Visit the UK on a business trip' (step-by-step navigation)~~[^1]
  - [x] 'Make a payment' (inputs with prefixes/suffixes and error validation)
- [x] Existing examples have been updated 
  - [x] Start page / 'Apply for a UK passport': Add headings to tab panels for mobile
  - [x] 'Update your account details': Add second password input
- [x] An example using the prototype service header component has been created (this will be within the service header PR and provided as a separate URL)

[^1]: We decided not to re-audit the step-by-step navigation, as this component is owned by the GOV.UK team rather than the Design System and hasn't been substantially altered since the last audit.

querkmachine:
The service header component example is located here: https://govuk-frontend-pr-4950.herokuapp.com/full-page-examples/manage-company-information

PR #5088, opened 2024-06-18
Update review app examples

querkmachine:
We want to merge these changes into our main branch so that we can use them for ongoing testing and future audits, without having to reproduce them anew.

For https://github.com/alphagov/govuk-frontend/issues/5085.

## Changes
- Adds headings to the Tabs panels in 'Start page' example.
- Adds a second Password input to the 'Update your account details' example.
  - Each has customised ARIA labels and screen reader text unique to each input.
  - Adds validation that checks that both inputs have the same value.
- Adds 'Make a payment' example for testing Text inputs with prefixes and related validation styles.
- Modify the 'Search' example to make it appear to be the second page of results, so that the previous link is visible in Pagination. 
- Remove Exit This Page component configuration from Child Maintenance example, as these were repeating the default values.

### DIFF
--- a/packages/govuk-frontend-review/src/views/full-page-examples/make-a-payment/index.mjs
+++ b/packages/govuk-frontend-review/src/views/full-page-examples/make-a-payment/index.mjs
@@ -1,0 +1,37 @@
+import express from 'express'
+import { body, matchedData, validationResult } from 'express-validator'
+
+import { formatValidationErrors } from '../../../utils.mjs'
+
+const router = express.Router()
+
+router.post(
+  '/make-a-payment',
+
+  body('amount')
+    .notEmpty()
+    .withMessage('Enter an amount')
+    .isNumeric()
+    .withMessage('Enter a number')
+    .isFloat({ gt: 0 })
+    .withMessage('Enter an amount greater than 0'),
+
+  (req, res) => {
+    const { example } = res.locals
+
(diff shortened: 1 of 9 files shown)


packages/govuk-frontend-review/src/views/full-page-examples/make-a-payment/index.mjs
```diff
@@ -0,0 +1,37 @@
...
+    const errors = formatValidationErrors(validationResult(req))
+
+    if (!errors) {
+      return res.redirect(303, `./${example.path}/confirm`)
+    }
+
+    res.render(`${viewPath}/index`, {
+      errors,
+      errorSummary: Object.values(errors),
+      values: matchedData(req, { onlyValidData: false }) // In production this should sanitized.
```
owenatgov, line 32:
```suggestion
      values: matchedData(req, { onlyValidData: false }) // In production this should be sanitized.
```

owenatgov (approved):
One pithy comment about the comment. Everything else looks grand. I don't see why we can't merge this.

packages/govuk-frontend-review/src/views/full-page-examples/make-a-payment/index.mjs
querkmachine: I'm going to be a meanie and reject this change, if just that this is a copy-pasted piece of code that has the same typo in about a dozen other places.

PR #5089, opened 2024-06-18
Fix details expanded state not announced on iOS

### LINKED ISSUE #2349: Details expanded state is not announced by iOS VoiceOver
## Description of the issue

As flagged by @mcheung-nhs, the expanded state of the details component is not announced by VoiceOver when the element is opened on iOS

1. Open https://design-system.service.gov.uk/components/details/default/index.html on an iPhone
2. Enable VoiceOver (VO)
3. Focus 'Help with nationality'
4. Double tap to expand the element
5. Swipe forward into the content
6. Swipe backwards to the summary

## Expected behaviour

- In step 4, after expanding the element , VO should announce 'Help with nationality, expanded'
- In step 6, when swiping back to the summary, VO should announce 'Help with nationality, expanded'

## Actual behaviour

Although VO announces the element as collapsed in step 3, it does not announce the expanded state.

- In step 4, after expanding the element , VO announces 'Help with nationality'
- In step 6, when swiping back to the summary, VO announces 'Help with nationality'

- Operating system: iOS 14.7.1
- Browser: Safari
- Browser version: N/A
- GOV.UK Frontend Version: 3.13.1

### LINKED ISSUE #4972: Is it worth adding aria expanded to details?
## Description of the issue

This issue is now very old, is it worth adding `aria-expanded` to details to fix it in Frontend instead?

 - https://github.com/alphagov/reported-bugs/issues/29

Came up recently here:

 - https://github.com/alphagov/govuk-design-system-backlog/issues/44#issuecomment-2098352959

36degrees:
For some reason, setting `display: inline-block` on the summary means that [VoiceOver on iOS no longer announces the expanded state of the `<details>` element][1].

We were using `display: inline-block` so that the interactive area and the focus state (when visible) were constrained to the text within the `<summary>`.

We can achieve the same thing by using `display: block` with `width: fit-content`.

However, now that we’re no longer using `display: inline-block` [^fn1], the the 5px bottom margin on the summary and the 30px bottom margin on the details now collapse when the `<details>` is closed *except* in Chrome which [has been updated to use `content-visibility`][2].

Preserve the existing margin on the component and make the behaviour across browsers consistent by establishing a new block formatting context using `display: flow-root`, preventing the margins from collapsing.

Safari < 13 does not support `display: flow-root` and so the margins will collapse, which means these older versions of Safari will have 5px less margin than when we were still using `display: inline-block`.

**EDIT:** We've dropped the commit that added `display: flow-root` to the root `details` element as we removed the extra spacing in https://github.com/alphagov/govuk-frontend/pull/5352, bypassing this issue. Therefore the previous 3 paragraphs are out of date. I'm preserving Ollie's description if we ever wanted to bring the spacing back in.

Fixes #2349
Fixes #4972 (whilst this PR doesn't address 4972 directly, we did explore this option and it unfortuantely doesn't make a difference. This solution has been our most successful so far)

[1]: https://bugs.webkit.org/show_bug.cgi?id=230408
[2]: https://issues.chromium.org/issues/40815464

[^fn1]: ‘margins of inline-block boxes do not collapse (not even with their in-flow children)’ – https://www.w3.org/TR/CSS21/box.html#collapsing-margins

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ We've made fixes to GOV.UK Frontend in the following pull requests:
 - [#5278: Fix service navigation mobile toggle spacing](https://github.com/alphagov/govuk-frontend/pull/5278)
 - [#5331: Fix Warning Text font weight when <strong> styles are reset](https://github.com/alphagov/govuk-frontend/pull/5331)
 - [#5352: Only apply margin to details summary when open](https://github.com/alphagov/govuk-frontend/pull/5352)
+- [#5089: Fix details expanded state not announced on iOS](https://github.com/alphagov/govuk-frontend/pull/5089)
 
 ## v5.6.0 (Feature release)
 
--- a/packages/govuk-frontend/src/govuk/components/details/_index.scss
+++ b/packages/govuk-frontend/src/govuk/components/details/_index.scss
@@ -8,8 +8,7 @@
   }
 
   .govuk-details__summary {
-    // Make the focus outline shrink-wrap the text content of the summary
-    display: inline-block;
+    display: block;
   }
 
   .govuk-details[open] .govuk-details__summary {
@@ -73,6 +72,9 @@
     .govuk-details__summary {
       // Absolutely position the marker against this element
       position: relative;
+
+      // Make the focus outline shrink-wrap the text content of the summary
+      width: fit-content;
 
       // Allow for absolutely positioned marker and align with disclosed text
       padding-left: govuk-spacing(4) + $govuk-border-width;


36degrees:
Spreadsheet for testing:
https://docs.google.com/spreadsheets/d/1Z7SxQcK-HcwScI_dbSDLwidL7mDctsN0aFPLHaMPmK8/edit?gid=0#gid=0

So far I've not observed any changes other than in Safari on iOS 17.5.1 (21F90) where we see the desired improvement, with the expanded state announced on toggle and when returning to the summary:

| Scenario | [Before](https://govuk-frontend-review.herokuapp.com/full-page-examples/what-is-your-nationality?hide-banner) | [After](https://govuk-frontend-pr-5089.herokuapp.com/full-page-examples/what-is-your-nationality?hide-banner) |
|---|---|---|
| When summary is focused (closed) | Help with nationality, button, collapsed. Double-tap to expand. | Help with nationality, button, collapsed. Double-tap to expand. |
| When toggling open | Help with nationality. | Help with nationality, expanded. |
| When summary is focused (open) | Help with nationality, button. | Help with nationality, button, expanded. Double-tap to collapse. |

I don't have access to test with TalkBack on Android – it'd be fab if someone else was able to help with that.

owenatgov:
I've had a quick look at the margin issue mentioned in the description and I'm going to make the bold case that this isn't worth worrying about. As Ollie mentions, Safari <13, so 12 and lower, for both mac and iOS will functionally have 5px less margin. You can see this illustrated in the below screenshot of Safari mac 12.1 where the margin collapses into the focus state bottom border:

screenshot of the details component on Safari mac 12.1 where the margin collapses into the focus state bottom border

Safari 11 and 12 fall under [Grade C in our browser support model](https://frontend.design-system.service.gov.uk/browser-support/#grade-c) which states the following about bugs in those browsers:

> For grade C browsers we:
> 
> - might remove support for individual features in these browsers at any time without considering them a breaking change
> - will not regularly test in these browsers
> - will not fix bugs affecting these browsers unless they prevent a user from being able to complete their task, or we can fix the bug by disabling a feature

I'm using this policy and the low impact to visual rhythm for Safari 12 and below users to justify not implementing a fix for this.

romaricpascal (commented):
Code changes look good to me. If I understood correctly, the margin issue is that we now have 30px of margin in [browsers that don't support `flow-root`](https://caniuse.com/flow-root) rather than 35px? 

Given the large margin of 30px and the limited browsers it applies to, I think it'd be fine, but it would be worth confirming with a designer.

Another thought on the matter is whether the margins not collapsing was intentional or accidental, in the first place. Looks like the summary requiring 5px of margin is really for when the detail is open, so there's some space between it and the grey bar, is it 'leaking' when the detail is closed? If that's the case we may want to condition that to `[open] .govuk-details__summary` or add extra `margin-top`/`padding-top` to the `govuk-details__text`, which would both clarify the reason for that extra space. We could look into that in a separate small story, though 😊 

**With margin**
Details element open with margin between arrow and grey bar
**Without margin**
Screen Shot 2024-09-23 at 10 42 21

Ciandelle:
I definitely think its worth fixing for all versions of the browser, but to answer to main question yes the answer sounds right to be, scrapping the extra 5px sounds like it should work. Without looking at the reasoning for it, I can’t really see why it would be there in the first place.

owenatgov:
Added a PR to tackle the spacing of the details element separately https://github.com/alphagov/govuk-frontend/pull/5352

owenatgov (approved):
Romaric had planned to approve this but he ran out of time yesterday and it's now his non-working day. Outside of github he's commented that he's happy with this change and he's given me permission to approve and merge this 🎉

selfthinker:
Just to give a bit more context that I had shared with the team internally last week...

I have tested this in other iOS versions, and I've also tested the native version. Results from that are in the spreadsheet.
(Back in June I had also tested on TalkBack and updated the spreadsheet. The result was that there was no difference.)

## VoiceOver iOS 17.5.1

Both the native version and the version from this PR work as expected.
While our old version doesn't.
The fact that the native version works as expected also makes clear that it was our CSS changes to `display` that caused the issue (together with the bug in iOS and/or VO that makes this behaviour possible).

## VoiceOver iOS 15.8

None of the versions work.
Even the native version only says the text of the summary with nothing else announced.

## VoiceOver iOS 12.5.7

Interestingly, the native version as well as the version from this PR worked okay.
And our old version doesn't.
That's an indication that that was a regression in iOS, either in version 13, 14 or 15.

## Other iOS versions

I don't have access to other iOS versions.
I tested with a simulator in iOS 16.
And although that cannot be relied upon, it seems to be as bad as iOS 15.

PR #5090, opened 2024-06-18
Remove redundant tag CSS from phase banner

36degrees:
The phase banner uses the tag component, which takes care of doing this for us since 976535e (part of #3502):

https://github.com/alphagov/govuk-frontend/blob/3822ac1bd000d0f965d3cece2716df91641b82cd/packages/govuk-frontend/src/govuk/components/tag/_index.scss#L32-L40


romaricpascal (approved):
Good spot! 🦅

PR #5091, opened 2024-06-18
🪴 Comment Gardening Part 2: Return of the 80 character limit

36degrees:
Follow up to last year's smash hit [Comment gardening 🪴](https://github.com/alphagov/govuk-frontend/pull/4317) this is another pass at trying to consistently wrap Sass comments at 80 characters, which we mostly do, except for when we don't.

I've rewritten a few comments, mainly where we were really close to fitting on one line but not quite. I've also opted to use short URLs for bug trackers where possible (`bugzil.la`, `webkit.org/b`) for brevity.

Finally, I've removed some ancient comments where we recorded what padding selects used to have in 2017 and pondered whether we should use `box-sizing: border-box` everywhere 🤔

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/pagination/_index.scss
+++ b/packages/govuk-frontend/src/govuk/components/pagination/_index.scss
@@ -54,8 +54,8 @@
   .govuk-pagination__next {
     @include govuk-typography-weight-bold;
 
-    // Use flex to get around a whitespace issue between the arrow svg and the link text
-    // without having to rely on whitespace control from backend tooling
+    // Use flex to avoid extra space being introduced when whitespace exists
+    // between the arrow svg and the link text
     .govuk-pagination__link {
       display: flex;
       align-items: center;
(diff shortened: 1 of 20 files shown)


packages/govuk-frontend/src/govuk/components/pagination/_index.scss
```diff
@@ -199,17 +199,17 @@
       border-top: 1px solid $govuk-border-colour;
     }
 
-    // Reset both these elements to their inline default, both to ensure that the focus state
-    // for block mode "shrink wraps" text as expected
+    // Reset both these elements to their inline default, both to ensure that
```
domoscargin, line 202: That second "both" looks a little odd, like there used to be a second reason in the past.

domoscargin (approved):
Looks good to me, just one comment.

Some of the more overt edits made me do a chef's kiss - turns out I enjoy English being optimised just as much as I enjoy code being optimised...

PR #5092, opened 2024-06-19
Move service name and navigation into section

domoscargin:
This is a possible fix for the 'unclear service name' accessibility issue from https://github.com/alphagov/govuk-design-system/issues/3811

## Links
- [Service header component](https://govuk-frontend-pr-5092.herokuapp.com/components/service-header)
- [Example with a header, service header (with service name and navigation) and phase banner](https://govuk-frontend-pr-5092.herokuapp.com/full-page-examples/upload-your-photo)
- [Example page with header and service header (service name only)](https://govuk-frontend-pr-5092.herokuapp.com/full-page-examples/work-history)
- [Example page with header and service header (navigation only)](https://govuk-frontend-pr-5092.herokuapp.com/full-page-examples/announcements)
- [Example page with cookie banner, header, phase banner, service header (with service name and navigation) and breadcrumbs](https://govuk-frontend-pr-5092.herokuapp.com/full-page-examples/manage-company-information), to emulate a potential 'worse case scenario'
- [Preview Branch](https://github.com/alphagov/govuk-frontend/tree/preview-bk-service-information-section). To install, use `npm install --save "alphagov/govuk-frontend#dc8ff97b4"`

## Before:

```html
<div class="govuk-service-header">
  <div class="govuk-service_header__container govuk-width-container">
    <!-- start slot -->
    <span>service name</span>
    <nav>service navigation</nav>
    <!-- end slot ->>
  </div>
</div>
```

Manage company information before

Slotted content before

After:

```html
<div class="govuk-service-header" aria-label="Service information and navigation">
    <!-- start slot -->
   <section class="govuk-service_header__container">
      <span>service name</span>
      <nav>service navigation</nav>
    </section>
    <!-- end slot ->>
  </div>
</div>
```

Manage company information after

Slotted content after

## Potential issues
- I've limited the background colour and bottom border to the service name and navigation section. This means folks would have to do some extra work to make the `start` and `end` slots look of a piece, but there's not really a guarantee they want them to look the same, and it enables a Phase Banner to be inserted without mucking about with that component's styles. I'm not sure it looks great with a border on the grey service header and then a border on the phase banner, though.
- Feels a bit weird not using the main `.govuk-service-header` class for anything. We could probably finagle some positioning and naming of classes.

### DIFF
--- a/packages/govuk-frontend-review/src/views/examples/template-custom/index.njk
+++ b/packages/govuk-frontend-review/src/views/examples/template-custom/index.njk
@@ -51,12 +51,14 @@
     containerClasses: "app-width-container--wide"
   }) }}
 
-  {{ govukPhaseBanner({
-    tag: {
-      text: "Alpha"
-    },
-    html: 'C\'est un nouveau service - vos <a class="govuk-link" href="#">commentaires</a> nous aideront à l\'améliorer.'
-  }) }}
+  {% set phaseBanner %}
+    {{ govukPhaseBanner({
+      tag: {
+        text: "Alpha"
+      },
+      html: 'C\'est un nouveau service - vos <a class="govuk-link" href="#">commentaires</a> nous aideront à l\'améliorer.'
+    }) }}
+  {% endset %}
 
   {{ govukServiceHeader({
     serviceName: "Nom du service",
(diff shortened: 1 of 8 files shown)


domoscargin:
I'm going to merge this into the component branch to save time and enable previews, but am still very open to it being changed.

PR #5102, opened 2024-06-26
Replace polyfill.io docs with notice

owenatgov:
Replace our polyfilling docs with a notice that we can't provide polyfilling advice in the wake of [the polyfill.io supply chain attack](https://sansec.io/research/polyfill-supply-chain-attack)


docs/contributing/polyfilling.md
```diff
@@ -1,39 +1,8 @@
...
-    )
-
-    if (detect) return
-
-    // Polyfill from https://cdn.polyfill.io/v2/polyfill.js?features=Event&flags=always
-    // code goes here
-
-}).call('object' === typeof window && window || 'object' === typeof self && self || 'object' === typeof global && global || {});
-```
+GOV.UK Frontend v4.8.0 and below makes reference to polyfill.io and uses polyfills generated by polyfill.io, however they were generated before polyfill.io came under malicious ownership and don't contain any malicious code.
```
domoscargin, line 8:
```suggestion
GOV.UK Frontend v4.8.0 and below makes reference to polyfill.io and uses polyfills generated by polyfill.io, however they were generated before polyfill.io came under malicious ownership and don't contain any malicious code.

If you have used this doc to generate a polyfill in the last year, you should check for any malicious code.
```

owenatgov:
Percy isn't passing but the changes themselves are fairly trivial. We think that this is the result of a percy version bump and not enough of a problem to block merging.

PR #5103, opened 2024-06-26
Replace polyfill.io docs with notice

owenatgov:
Repeat of https://github.com/alphagov/govuk-frontend/pull/5102 for 3.x


domoscargin (approved):
🔥

PR #5127, opened 2024-07-10
Remove direct links to Polyfill[dot]io

### LINKED ISSUE #5105: Replace links to polyfill.io in comments on `support/v4.x` branch
## What

On the `support/4.x` branch, replace links to the polyfill.io website in the comments of our vendored polyfills with the following note at the top of the file:

```js
/**
 * NOTE
 * 
 * These polyfills were generated using polyfill.io, which was reported as compromised on 25th June 2024.
 * 
 * We generated this code well before the compromise, and it is free of malicious code.
 * However, we recommend checking any polyfills you have generated in a similar way.
 */
```

## Why

polyfill.io was reported as compromised on 25th June 2024. While our code doesn't load scripts directly from the live service, the polyfills in govuk-frontend had been extracted from this service while it was free of malicious code. These extracts have comments pointing to the polyfill.io website, which would lead our users to a malicious site.

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [ ] Comments linking to polyfill.io in our polyfills have been removed in favour of the note above.

domoscargin:
Closes #5105 

The included note bulked out the package code considerably (see [#dcf4810](https://github.com/alphagov/govuk-frontend/commit/dcf481045b4aba63390ad672bea2a368f086e693)), so I've gone with a single README in the polyfill folder.

If we wanted to include the note in the code itself, we could add a version of it to `src/all.mjs`. Then it'd only be included once. But that felt a bit misplaced.

To propagate the changes to the `dist` and `package` folders, we'd need to do a patch release. I considered just manually updating these and skipping the CHANGELOG entry, but this seems more transparent, despite being more hassle.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+We’ve made fixes to GOV.UK Frontend in the following pull requests:
+
+- [#5127: Remove direct links to Polyfill[dot]io](https://github.com/alphagov/govuk-frontend/pull/5127)
+
 ## GOV.UK Frontend v4.8.0 (Feature release)
 
 This release includes the ability to update the crown logo. You must do this between 19 February and 1 March 2024.
--- a/src/govuk/vendor/README.md
+++ b/src/govuk/vendor/README.md
@@ -1,0 +1,7 @@
+# NOTE
+
+These polyfills were generated using polyfill.io, which was reported as compromised on 25th June 2024.
+
+We generated this code well before the compromise, and it is free of malicious code.
+
+However, we recommend checking any polyfills you have generated in a similar way.
--- a/src/govuk/vendor/polyfills/DOMTokenList.mjs
+++ b/src/govuk/vendor/polyfills/DOMTokenList.mjs
@@ -1,7 +1,6 @@
 // @ts-nocheck
 (function (undefined) {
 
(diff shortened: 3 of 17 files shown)


package/govuk-esm/vendor/polyfills/Date/now.mjs
```diff
@@ -1,12 +1,12 @@
 // @ts-nocheck
+
 (function (undefined) {
 
     // Detection from https://github.com/Financial-Times/polyfill-library/blob/v3.111.0/polyfills/Date/now/detect.js
```
owenatgov, line 5: Do we also want to remove references to the FT polyfill repo? If you go to the repo it's been erased for containing malicious code. Github seems to be handling it but maybe good hygine? Whatcha reckon?

owenatgov (commented):
One little comment/question. This is looking solid otherwise 💪🏻

package/govuk-esm/vendor/polyfills/Date/now.mjs
domoscargin: Yeah, go on.

domoscargin:
@owenatgov I've smooshed ALL related comments now, and added a CHANGELOG entry rather than manually edit the `dist` and `package` folders. Just makes everything more obvious, even if it means the hassle of a support branch release at some point.

Note for future releasers: the diff to the `dist` folder was big enough that our validation checks fail because we can't generate a comment that large.

owenatgov (approved):
polyfill.io? Never heard if it.

PR #5156, opened 2024-07-19
Document private fields in dev JSDoc

domoscargin:
Enable private fields in our generated documentation by default.

You still need to select the "Private" checkbox in Settings > Member Visibility on the JSDoc page (`/docs/javascript/`).

Since the review app is meant for internal development, it makes sense to expose everything by default.


romaricpascal (approved):
That's going to be nice! ⛵

Issue #5154, opened 2024-07-18
Tabs cannot be accessed via arrowing with NVDA/Firefox


selfthinker:
## Description of the issue

When using NVDA with Firefox, the individual tabs in the [Tabs component](https://design-system.service.gov.uk/components/tabs/) cannot be accessed via arrowing.
(They cannot be accessed via tabbing either, but the that is expected behaviour.)

This is the [same issue we encountered with the navigation component](https://github.com/alphagov/govuk-design-system/issues/3811#issuecomment-2183273602). It's due to a vendor bug and can be fixed by not using `display: inline-block` on the tabs.

This issue also happens with older versions of Chrome. It was fixed in Chrome version 125.

This can be overcome by using the tab key to get to the individual tabs. But when NVDA users do that they are faced with [another problem](https://github.com/alphagov/govuk-frontend/issues/860) that means they cannot easily get out of the tabs again.

## Steps to reproduce the issue

1. Use Firefox and NVDA
2. Open https://design-system.service.gov.uk/components/tabs/default/
3. Use the down arrow key to get to the first tab
4. Use the down arrow key again

### Actual behaviour

When using the down arrow key (as per step 3 above), it reads out all of the tabs:

> Tab selected past day, tab past day, tab past month, tab past year

And when using it again (as per step 4), it skips all the other tabs straight to the heading and reads out:

> Heading level 2, past day

### Expected behaviour

When using NVDA with Chrome (from version 125 onwards) or JAWS with Firefox or Chrome it shows the expected behaviour.

When using the down arrow key (as per step 3), it reads out only the first tab:

> Tab selected past day

And when using it again (as per step 4), it reads out the next tab:

> Tab past week

## Environment (where applicable)

- Operating system: Windows 10 Pro
- Browser: Firefox
- Browser version: 127 (latest), although it's in older versions as well and in older Chrome versions
- Screen reader: NVDA
- Screen reader version: 2024.2
- GOV.UK Frontend Version: latest (the one that's deployed on the GOV.UK Design System website)

Issue #860, opened 2018-07-02
Tabs interface needs better way to get to panel content


kr8n3r:
Moving over from [Frontend Toolkit](https://github.com/alphagov/govuk_frontend_toolkit/issues/464) 
issue raised by  @detlevhfischer

When used with NVDA it is difficult to get from the tabs on https://design-system.service.gov.uk/components/tabs/ to the corresponding panel. Both left / right as well as up / down arrows traverse the tabs in the tablist. I suggest to consider implementing arrow down as a shortcut to move focus to the right panel content. See https://inclusive-components.design/tabbed-interfaces/ , section "A problem reading panels".

@alex-ju commented
@detlevhfischer: hi, thanks for adding your suggestions, we took that article into consideration while building the component. have you done any user research around this?

@detlevhfischer commented
I just tried it using NVDA and Firefox and noticed that left/right arrow did the same as top/down arrow, so that latter is not used (as I believe recommended by Heydon) to move the focus to the panel. Leaving forms mode and trying reading mode to progress to the panel often lands in the wrong one because tab and its panel are not adjacent in the source code, I guess.
I intend o include the gov.uk example in the user tests I am currently conducting and will report back on issues, if there are any.

@alex-ju commented
Feedback from research would be extremely valuable for us and would help us shape the component, thank you!

hannalaakso:
Hi @detlevhfischer,

Thanks for raising this issue related to the tabs component. 

We'd be keen to know if any related findings came out of the user tests you conducted?

Thanks,
Hanna

detlevhfischer:
It's holiday time - there will be more tests after the summer break. I'll be in touch!
[...]

PR #5158, opened 2024-07-19
Remove ↑ up and ↓ down arrow key bindings from tabs

querkmachine:
Currently the Tabs component binds the up and down arrow keys to the same functionality as the left and right arrow keys, using them to navigate between the previous and next tabs within a tablist. 

Last week, [we received a comment stating that these bindings were causing problems in NVDA](https://github.com/alphagov/govuk-design-system-backlog/issues/100#issuecomment-2217329666), as the user became 'trapped' within the tablist and could not arrow out of it—up and down arrows being a frequent control mechanism for moving between adjacent elements.

This morning, [we received another comment stating the opposite](https://github.com/alphagov/govuk-design-system-backlog/issues/100#issuecomment-2238652113), and that the up/down bindings weren't necessarily problematic and that a user could 'break out' by switching to NVDA's document mode.

The [example pattern in the WAI-ARIA Authoring Practices](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/#keyboardinteraction) explicitly states that only the arrow keys matching the physical direction of the tablist should be bound to tab changes (left/right for a horizontal list, up/down for a vertical list). 

On this basis, I think we should remove the existing key bindings for the up/down arrows.

## Changes

- Removes up and down arrow `keydown` bindings in the Tabs component.

## Todo 

There's an open-ended question about whether this constitutes a breaking change or not. Is it removing functionality or altering it? Or is it just fixing something that's unexpected? I've avoided adding a changelog just yet so we can ponder that. 

## Thoughts

I initially experimented with rebinding the up and down arrows to move focus between the tablist and the tab panel, as suggested [in this issue](https://github.com/alphagov/govuk-frontend/issues/860), however this still felt like it had a high risk of being incongruous with how an assistive technology user expects their tools to behave and it isn't a behaviour suggested by WAI-ARIA, so I backtracked on doing that in this PR. [You can see that work in a separate branch.](https://github.com/alphagov/govuk-frontend/compare/tabs-remove-up-down-keybinding...tabs-down-arrow-focuses-panel?expand=1)

This does not resolve the other recent issue with the Tabs component in NVDA/Firefox. https://github.com/alphagov/govuk-frontend/issues/5154

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ We've made fixes to GOV.UK Frontend in the following pull requests:
 
 - [#5046: Skip ‘empty’ tasks in the task list](https://github.com/alphagov/govuk-frontend/pull/5046)
 - [#5066: Fix whitespace affecting text alignment in pagination block variant](https://github.com/alphagov/govuk-frontend/pull/5066)
+- [#5158: Remove ↑ up and ↓ down arrow key bindings from tabs](https://github.com/alphagov/govuk-frontend/pull/5158)
 
 ## 5.4.1 (Fix release)
 
--- a/packages/govuk-frontend/src/govuk/components/tabs/tabs.mjs
+++ b/packages/govuk-frontend/src/govuk/components/tabs/tabs.mjs
@@ -365,26 +365,22 @@ export class Tabs extends GOVUKFrontendComponent {
   /**
    * Handle tab keydown event
    *
-   * - Press right/down arrow for next tab
-   * - Press left/up arrow for previous tab
+   * - Press right arrow for next tab
+   * - Press left arrow for previous tab
    *
    * @private
    * @param {KeyboardEvent} event - Keydown event
    */
   onTabKeydown(event) {
     switch (event.key) {
-      // 'Left', 'Right', 'Up' and 'Down' required for Edge 16 support.
+      // 'Left' and 'Right' required for Edge 16 support.
       case 'ArrowLeft':
-      case 'ArrowUp':
       case 'Left':
-      case 'Up':
         this.activatePreviousTab()
         event.preventDefault()
         break
       case 'ArrowRight':
-      case 'ArrowDown':
       case 'Right':
-      case 'Down':
         this.activateNextTab()
         event.preventDefault()
         break


selfthinker:
I don't have any concerns from an accessibility perspective.

But I wonder if it could be a problem if users got used to using the up/down arrow keys and would now be missing the functionality. It might be difficult to find out if that is the case or not. And even if that turns out to be true, it might not be a reason to not remove them. There are certainly lots of tab components out there in other design systems and other websites that only support left/right, so users should generally be used to the that.
What do others think?

romaricpascal:
> But I wonder if it could be a problem if users got used to using the up/down arrow keys and would now be missing the functionality. It might be difficult to find out if that is the case or not. And even if that turns out to be true, it might not be a reason to not remove them. There are certainly lots of tab components out there in other design systems and other websites that only support left/right, so users should generally be used to the that.
What do others think?

Looking at the change, it's a fairly easy one to reverse should we discover we're in the wrong and release a patch.

The change looks good to merge. Interesting that there's no tests for these keys (looks like we're only testing for 'ArrowRight' in [tabs.puppeteer.test.js](https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/src/govuk/components/tabs/tabs.puppeteer.test.js). I'll create an issue to test add tests for ArrowLeft at least (if there's none already) so we close that gap in a separate PR.

That leaves the question of statuating on whether it's a breaking change or not before we merge, which we can discuss on Slack or at dev catch-up.

romaricpascal:
Digging a bit actually, looks like Bootstrap is also having some proposed changes by a user on their Tabs to remove the support for up/down arrows. This comment makes a good point about [the need for both up/down and left/right when the tab list can be scrolled](https://github.com/twbs/bootstrap/issues/39969#issuecomment-2241035192). However, our tabs do not scroll, but reflow when too many tabs are present, so I think it's safe for us to remove.

selfthinker:
> Bootstrap is also having some proposed changes by a user on their Tabs to remove the support for up/down arrows.

They also make a good point that the standard behaviour of up/down arrow keys to scroll the page breaks for non-SR keyboard users. Another reason to remove it.

owenatgov (commented):
I agree with Romaric and Anika's takes in the convo of this PR. Reckon this is hot to go.

As for the type of change, I'd like to make a case that it's _not_ a breaking change on the basis that we're not making any changes to the 'API' of govuk-frontend, even if it impacts functionality. The only way I can see in which this could be breaking for someone is if they have an automated test that checks for the up down arrow interaction. That's possible but IMO not likely and not critical. Generally speaking, 'functionality' is on the other side of the line we've drawn on what is and isn't breaking and I don't think this is a reason to change our minds on that.

Not expecting us to come to a conclusion on this now. Just airing my thoughts on this.

querkmachine:
I've added it as a fix entry in the changelog.

romaricpascal (approved):
Looks all good with the fix entry, thanks ⛵

PR #5159, opened 2024-07-19
Add break-word typography helper

### LINKED ISSUE #2233: Long, unbroken link text (such as email addresses) can break out of page container
## Summary
The standard wrapping on body content can create a scenario where long email addresses can cause a service to fail [WCAG 2.1 - 1.4.10 Reflow](https://www.w3.org/WAI/WCAG21/Understanding/reflow.html).

Reflow states that content must not force horizontal scrolling in a viewport of 320px wide unless it's a map, a table, a game or something else where it is necessary. For regular content it needs to reflow and maintain vertical scrolling only.

If you provide an email address which does not contain regular breaking characters such as hyphens, then the email address refuses to break and forces the viewport to horizontal scroll.

By adding the CSS property: `overflow-wrap:anywhere` the issue is resolved.

1. Add a long email address to some body content which does not contain any standard breaking characters. For example:
```html
<a href="#" class="govuk-link">
  anExampleOfAReallyLongEmailAddress@WhichPushesTheContentOver320PixelsWide.com
</a>
```

2. Reduce the viewport width to 320px wide and observe horizontal scrolling.

image

- Operating system: OS X Catalina + Windows 10
- Browser: Chrome (Mac), Chrome (Win), Edge, Safari
- GOV.UK Frontend Version: 3.12.0

querkmachine:
Teeny little PR to add a new typography helper for breaking long words, such as email addresses or 50% of the German language, across multiple lines when a container is likely to be too narrow to contain it.

This suggestion was originally raised in #2233. It felt useful to create it as a mixin, rather than have service teams implement it as-needed, due to older Microsoft browsers only supporting a non-standard, unprefixed version of the property.

Comments on the issue noted that we probably cannot apply a 'one size fits all' CSS rule to text generally, as this would interfere with components that are intrinsically sized according to their contents, such as the Tag component.

Resolves #2233.

## Changes
- Adds a new `govuk-text-break-word` mixin, which sets both the nonstandard `word-wrap` and standard `overflow-wrap` CSS properties.
- Adds tests for the mixin functionality.
- Adds a new `govuk-!-text-break-word` override class.

## Todo
- Addition of the new override class to the [font override classes guidance](https://design-system.service.gov.uk/styles/font-override-classes/).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,22 @@ For advice on how to use these release notes see [our guidance on staying up to 
 
 ## Unreleased
 
+### New features
+
+#### Stop long words breaking out of components with `govuk-!-text-break-word`
+
+We've added a new override class to help display long or unpredictable words on narrow screens, such as an email address entered by a user.
+
+Wrapping the content with the `govuk-!-text-break-word` class forces words that are too long for the parent element to break onto a new line.
+
+```html
+A confirmation email will be sent to <span class="govuk-!-text-break-word">arthur_phillip_dent.42@peoplepersonalitydivision.siriuscyberneticscorporation.corp</span>.
+```
+
+Sass users can also use the `govuk-text-break-word` mixin.
+
+This change was introduced in [pull request #5159: Add break-word typography helper](https://github.com/alphagov/govuk-frontend/pull/5159).
+
 ### Fixes
 
 We've made fixes to GOV.UK Frontend in the following pull requests:
--- a/packages/govuk-frontend/src/govuk/helpers/_typography.scss
+++ b/packages/govuk-frontend/src/govuk/helpers/_typography.scss
@@ -79,6 +79,23 @@
   font-variant-numeric: tabular-nums if($important, !important, null);
 }
 
+/// Word break helper
+///
+/// Forcibly breaks long words that lack spaces, such as email addresses,
+/// across multiple lines when they wouldn't otherwise fit.
+///
+/// @param {Boolean} $important [false] - Whether to mark declarations as
+///   `!important`. Generally used to create override classes.
+/// @access public
+
+@mixin govuk-text-break-word($important: false) {
+  // IE 11 and Edge 16–17 only support the non-standard `word-wrap` property
+  word-wrap: break-word if($important, !important, null);
+
+  // All other browsers support `overflow-wrap`
+  overflow-wrap: break-word if($important, !important, null);
+}
+
 /// Convert line-heights specified in pixels into a relative value, unless
 /// they are already unit-less (and thus already treated as relative values)
 /// or the units do not match the units used for the font size.
--- a/packages/govuk-frontend/src/govuk/helpers/typography.unit.test.js
+++ b/packages/govuk-frontend/src/govuk/helpers/typography.unit.test.js
@@ -170,6 +170,50 @@ describe('@mixin govuk-font-tabular-numbers', () => {
   })
 })
 
+describe('@mixin govuk-text-break-word', () => {
+  it('adds the word-wrap and overflow-wrap properties', async () => {
+    const sass = `
+      ${sassBootstrap}
+
+      .foo {
+        @include govuk-text-break-word;
+      }
+    `
+
+    const results = compileSassString(sass)
+
+    await expect(results).resolves.toMatchObject({
+      css: outdent`
+        .foo {
+          word-wrap: break-word;
+          overflow-wrap: break-word;
+        }
+      `
+    })
+  })
+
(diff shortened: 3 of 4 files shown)


packages/govuk-frontend/src/govuk/helpers/_typography.scss
```diff
@@ -79,6 +79,23 @@
   font-variant-numeric: tabular-nums if($important, !important, null);
 }
 
+/// Word break helper
+///
+/// Forcibly breaks long words that lack spaces, such as email addresses,
+/// across multiple lines when they wouldn't otherwise fit.
+///
+/// @param {Boolean} $important [false] - Whether to mark declarations as
+///   `!important`. Generally Used to create override classes.
```
owenatgov, line 88:
```suggestion
///   `!important`. Generally used to create override classes.
```

CHANGELOG.md
```diff
@@ -4,6 +4,20 @@ For advice on how to use these release notes see [our guidance on staying up to
 
 ## Unreleased
 
+### New features
+
+#### Stop long words breaking out of components with `govuk-!-font-break-word`
+
+We've added a new override class to help display long or unpredictable words on narrow screens, such as an email address entered by a user.
```
owenatgov, line 11: Can we mention the mixin here as well?

owenatgov (approved):
A couple of little comments, nothing blocking. Looks grand!

Another non-blocking question: what do we reckon to filtering `govuk-font-break-word` into the parent `govuk-font` mixin?

romaricpascal:
> Another non-blocking question: what do we reckon to filtering `govuk-font-break-word` into the parent `govuk-font` mixin?

Not sure I understand what you mean by this, are you talking about having `govuk-font` have an option to also apply `govuk-font-break-word`?

This brough to my mind that I'd also associate breaking words with 'text' setting rather than 'font' setting, so the `govuk-font' prefix feels a bit odd to me, but this may be my brain being too picky.

romaricpascal:
> This brough to my mind that I'd also associate breaking words with 'text' setting rather than 'font' setting, so the `govuk-font' prefix feels a bit odd to me, but this may be my brain being too picky.

Had a quick check and that distinction goes all the way down to CSS specs:
- font stuff (like `font-size`) are in the [CSS Fonts module](https://drafts.csswg.org/css-fonts/#font-size-prop)
- text stuff (like `overflow-wrap` and `word-wrap`) are in the [CSS Text Module](https://drafts.csswg.org/css-text/#overflow-wrap-property).

I'm not sure it's a distinction that'll be in the mind of people looking for the feature, though. I'm wondering if we need the `font` mention at all and could just go with:
- `govuk-!-break-word` (in a similar way as we have `govuk-!-text-align-left` which is close to the property name)
- `govuk-break-word` as a mixin

@querkmachine what drove the `govuk-font` and `govuk-!-font` prefix?

querkmachine:
@romaricpascal Typically I would try to name a mixin/override to be the same as the CSS property and value being defined, but didn't do that here as multiple properties are being defined and `overflow-wrap` just doesn't strike me as being a very intuitive name for what the styles do (it sounds more like something to do with the `overflow` property.)

Otherwise it's close to what you thought: There doesn't seem to be a clear distinction between 'font', 'type' and 'typography' in the minds of most people, and we've generally seemed to gravitate towards using the word 'font' for all mixins relating to text in recent changes. 

I'm open to changing the names. From my perspective, having some word to faux namespace the mixin seems like a nice-to-have, just so it's obvious it's in the font/text group of mixins and classes—though the mixins will all be namespaced under `typography` once Sass modules come into play, anyway.

romaricpascal:
@querkmachine cheers for the explanation. Agree that namespacing sound nice and that `overflow` wouldn't make a great for the reason you mention.

`font` doesn't seem quite right a namespace either (for example, we don't force a `font-text-align` either to namespace the `text-align` utilities). How about:
- `govuk-text-break-word` for the mixin
- `govuk-!-text-break-word` for the utility

querkmachine:
Renamed 🎉

querkmachine:
> Another non-blocking question: what do we reckon to filtering `govuk-font-break-word` into the parent `govuk-font` mixin?

I don't have any particular reasons against making it available via `govuk-font` other than that my _assumption_ is that the override class is going to be used very specifically and contextually—either on pre-composed content where a developer knows a long, unbroken string will appear, or when playing back particular user-provided content that has the potential to contain long, unbroken strings.

In my mind, that means that only single words (in prose) or elements (in a component) will have the mixin/class applied, rather than it being part of a wider reaching font configuration. That's entirely assumptive though—it (probably) doesn't really make any difference where the break-word styles are placed.

Issue #3497, opened 2023-04-11
Consider standardising on default id


joelanman:
Consider standardising on making a default id from the components name if an id is not given.

## Related component

Radios make a default id from the name, other components like textarea do not

## Context

Training people in using the Prototype Kit, where it was confusing why some components need id and others do not

joelanman:
great work! I'd suggest defaults from `name` for all the required ones

joelanman:
if this is something we're interested in I'm happy to try a PR

querkmachine:
@colinrotherham has brought up that `name` may not always be a valid value for an `id`, for example, in the Date Input, the fields may have names like `dateOfBirth[day]`, `dateOfBirth[month]` and `dateOfBirth[year]`. These are not valid `id`s as square brackets aren't permitted in `id` values.

Simiarly to this, there is no guarantee that a `name` be unique, as is required for `id`s, but the cases where this is most likely to happen (checkboxes and radio buttons) already have this accounted for within the Nunjucks templates.

joelanman:
the default dateInput example takes a `namePrefix` which we could use the same way to make a default `id`, it then builds the default day month and year from that

frankieroberto:
This just tripped me up slightly - I was confused why the label of a text input didn't focus the input on clicking it, which turned out to be because I'd forgotten to add an id using the Nunjucks macro.

I think I had wrongly assumed that the id would default to the `name` value, as that's what happens with radios and checkboxes.

joelanman:
I think thats a good point - if missed it creates a significant accessibility problem, the label and input won't be associated for screen readers and other assistive tech

querkmachine:
> [@colinrotherham](https://github.com/colinrotherham) has brought up that `name` may not always be a valid value for an `id`. For example, in the Date Input, the fields may have names like `dateOfBirth[day]`, `dateOfBirth[month]` and `dateOfBirth[year]`. These are not valid `id`s as square brackets aren't permitted in `id` values.

In a wonderful example of "you never stop learning", it turns out this hasn't been the case for nearly 18 years! HTML5 liberalised what characters are allowed in `id` attributes to include every character that's allowed by other HTML attributes, so square brackets are a-okay being in `id`s.

PR #5171, opened 2024-07-24
[WIP] Define default `id` parameter values for various components

querkmachine:
Many components in the Design System have a functional requirement for an `id` to be defined in order to work as expected. As it stands, some components provide a default `id` value if one isn't specified, whereas other components don't. We already do this ourselves for the Checkboxes and Radios components, by having the `idPrefix` to match the `name` parameter by default.

This has led to [instances of developers forgetting to set `id`s](https://github.com/alphagov/govuk-frontend/issues/3497#issuecomment-1697688942) and for some ports of govuk-frontend to implement their own methods of setting default `id`s. The [ASP.NET Core](https://github.com/gunndabad/govuk-frontend-aspnetcore) and [Jinja](https://github.com/LandRegistry/govuk-frontend-jinja) ports both fall back to `name`, and the [Vue port](https://govukvue.org/) automatically generates unique IDs if one isn't specified. 

Related to #3497. 

## Changes
- Everywhere an `id` is required, but lacked a default or fallback value, now has a default or fallback defined.
  - On form controls, this fallback is to the `name` parameter. Otherwise, it is to a hardcoded string (the component name). 
- Update Nunjucks parameter documentation for modified components.
- Update default examples to remove explicitly defined `id` parameters.

### Components changed
Previously, all of these components required an `id` attribute to work as expected and/or to render valid and accessible HTML.

- **Accordion**: ID defaults to string `"accordion"`. Used for the root element and as a prefix for child element IDs. 
- **Character count**: ID falls back to `name` parameter. Used for the `textarea` and for related labels/hint IDs. 
- **Date input**: ID falls back to the `namePrefix` parameter if it's defined, otherwise defaults to the string `"date-input"`. Used for the root element and to prefix the IDs of child inputs.
- **File upload**: ID falls back to `name` parameter. Used for the `input` and for related labels/hint IDs. 
- **Select**: ID falls back to `name` parameter. Used for the `select` and for related labels/hint IDs. 
- **Text input**: ID falls back to `name` parameter. Used for the `input` and for related labels/hint IDs. 
- **Textarea**: ID falls back to `name` parameter. Used for the `textarea` and for related labels/hint IDs. 

**Tabs** is a bit of a weird case in that its `id` and `idPrefix` parameters were completely detached from one another, both needing setting individually even if they were the same value.

This was incongruent with similar components, where the `idPrefix` always falls back to the `id` or `name`. In this case, I've made it so that `idPrefix` falls back to match the `id`, and the `id` defaults to `"tabs"`. 

## Todo
- There are no new or modified tests for this change.
- Changelog entry.

## Thoughts
- There feels like there's greater value to fallbacks existing for form controls (where we can match the `name`) than non-form components like the Accordion and Tabs. I'm open to not setting defaults for non-form components, but having defaults for everything feels more consistent.
- Cases where the fallback is the `name` parameter can introduce a potential issue — `name`s can contain characters that `id`s cannot, such as square brackets. There are no attempts to protect against the creation of invalid `id`s here, though one could potentially be introduced, such as 'slugifying' the `name`.

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/accordion/accordion.yaml
+++ b/packages/govuk-frontend/src/govuk/components/accordion/accordion.yaml
@@ -234,6 +234,15 @@ examples:
               </ul>
 
   # Hidden examples are not shown in the review app, but are used for tests and HTML fixtures
+  - name: id
+    hidden: true
+    options:
+      id: custom-id
+      items:
+        - heading:
+            text: Section A
+          content:
+            text: Some content
   - name: classes
     hidden: true
     options:
--- a/packages/govuk-frontend/src/govuk/components/accordion/accordion.yaml
+++ b/packages/govuk-frontend/src/govuk/components/accordion/accordion.yaml
@@ -1,7 +1,7 @@
 params:
   - name: id
     type: string
-    required: true
+    required: false
     description: Must be unique across the domain of your service if `rememberExpanded` is `true` (as the expanded state of individual instances of the component persists across page loads using [session storage](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage)). Used as an `id` in the HTML for the accordion as a whole, and also as a prefix for the `id`s of the section contents and the buttons that open them, so that those `id`s can be the target of `aria-control` attributes.
   - name: headingLevel
     type: integer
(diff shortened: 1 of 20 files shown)


querkmachine:
Closing this PR as many of its changes have been merged via #5658, and there doesn't seem to be as much benefit to settings defaults for the others.

PR #5187, opened 2024-08-01
Add documentation on how to deploy the review app

romaricpascal:
Forks will not get an automatic deployment of the review app, so it's worth having a little bit of doc to explain how to set things up.

### DIFF
--- a/docs/contributing/deploying-the-review-app.md
+++ b/docs/contributing/deploying-the-review-app.md
@@ -1,0 +1,48 @@
+# Deploying the review app
+
+GOV.UK Frontend review app is automatically deployed on Heroku for PRs raised from branches of the `alphagov/govuk-frontend` repository.
+If you've forked GOV.UK Frontend in your own repository (for contributing a bug fix or a feature, for example),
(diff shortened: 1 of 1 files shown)


docs/contributing/deploying-the-review-app.md
```diff
@@ -0,0 +1,48 @@
...
+This means it does not run on 'static' hosting services like GitHub Pages or Netlify.
+
+Your organisation may already use such hosting service (for the Prototype Kit, for example).
+Check with your IT or digital team about which platform to use.
+
+Some hosting services automatically publish every time you push to GitHub. For example:
+
+- [Railway](https://railway.app/new/github)
+- [Render](https://render.com/docs/github)
+- [Heroku](https://devcenter.heroku.com/articles/github-integration) (requires payment)
```
romaricpascal, line 20: **note** Taken from [the Prototype Kit docs](https://prototype-kit.service.gov.uk/docs/publishing#hosting-services). I don't think we need to put much effort in maintaining that list, just give pointers for people to start doing their own research.

patrickpatrickpatrick (approved):

PR #5188, opened 2024-08-01
Remove deprecated Husky setup code

querkmachine:
govuk-frontend repo duplicate of this PR: https://github.com/alphagov/govuk-design-system/pull/3970

As this is solely related to internal tooling, it's my understanding that this doesn't require a changelog entry.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,1 @@
-#!/usr/bin/env sh
-. "$(dirname -- "$0")/_/husky.sh"
-
-npx --no-install lint-staged
+npx --no lint-staged


domoscargin (approved):
Looks good locally.

I guess since we're specifying `--no-install`, we can't jettison `npx` as per the 9.1.1 release notes? https://github.com/typicode/husky/releases/tag/v9.1.1

querkmachine:
> I guess since we're specifying `--no-install`, we can't jettison `npx` as per the 9.1.1 release notes? https://github.com/typicode/husky/releases/tag/v9.1.1

@domoscargin It's a bit of a poorly named flag.[^1] It doesn't run `lint-staged` without installing it, but instead automatically aborts running `lint-staged` if the CLI prompts to install it. Thus it requires `lint-staged` to already be installed if Husky is to do anything. 

My assumption is that's intentional? I'm not totally sure why (maybe release scripts don't necessarily have `lint-staged`, maybe it's a developer consideration). Either way, it seems harmless to keep it.

[^1]: It's also the deprecated name for the flag. It's just `--no` in more recent versions of npm.

querkmachine:
I've updated the PR to swap [the deprecated `--no-install` flag](https://docs.npmjs.com/cli/v9/commands/npx?v=true#compatibility-with-older-npx-versions) for the replacement `--no` flag as well.

PR #5191, opened 2024-08-02
Fix rendering of Back link's `href` and `text` for falsy values

### LINKED ISSUE #5189: Restore behaviour of Back Link component for falsy `text` and `href` options
## What

Make `text` and `href` correctly render default values when a falsy value is passed to the component, and add some tests checking this for `undefined`/`null` and different kinds of falsy values `0`,`''`, `false`.

## Why

The use of the `default` filter means only `undefined` values are replaced by the default value provided to the filter. This means a `''` string as `text` would make the button render with no content, causing both visual and accessibility issues.

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [ ] Default values for `text` and `href` render when passed a 'falsy' value
- [ ] Tests have been added for `undefined`/`null` and different kinds of falsy values `0`,`''`, `false`

romaricpascal:
Reverts to the behaviour of 5.4.0, which ensured consistency across falsy values, and importantly that a 'Back' text was displayed in case of empty strings (and a '#' value for the `href` attribute, but there's less difference with a '').

Adds tests to ensure we keep an eye on the behaviour in future template changes, or at least make sure that any deviation is intentional (and tested).

Instead of reverting the commit, I used the `boolean` option from Nunjucks' `default` filter to keep the tidy up of the templates introduced in https://github.com/alphagov/govuk-frontend/commit/fc704d1efc0fc15a69ff84eb18050dd7ba7ce4b3.

Fixes #5189

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/back-link/template.jsdom.test.js
+++ b/packages/govuk-frontend/src/govuk/components/back-link/template.jsdom.test.js
@@ -35,32 +35,71 @@ describe('back-link component', () => {
     expect($component).toHaveClass('app-back-link--custom-class')
   })
 
-  it('allows the link to be customised using the `href` option', () => {
-    document.body.innerHTML = render('back-link', examples['with custom link'])
+  describe('the `href` option', () => {
+    it('allows the link to be customised', () => {
+      document.body.innerHTML = render(
+        'back-link',
+        examples['with custom link']
+      )
 
-    const $component = document.querySelector('.govuk-back-link')
-    expect($component).toHaveAttribute('href', '/home')
+      const $component = document.querySelector('.govuk-back-link')
+      expect($component).toHaveAttribute('href', '/home')
+    })
+
+    it.each(['', 0, false, null, undefined])('uses `#` for `%s`', (href) => {
+      document.body.innerHTML = render('back-link', { context: { href } })
+      const $component = document.querySelector('.govuk-back-link')
+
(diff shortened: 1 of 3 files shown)


packages/govuk-frontend/src/govuk/components/back-link/template.jsdom.test.js
```diff
@@ -35,32 +35,74 @@ describe('back-link component', () => {
...
+      )
+
+      const $component = document.querySelector('.govuk-back-link')
+      expect($component).toHaveAttribute('href', '/home')
+    })
 
-    const $component = document.querySelector('.govuk-back-link')
-    expect($component).toHaveAttribute('href', '/home')
+    it.each(['', 0, false, null, undefined])(
+      'displays `Back` for `%s`',
```
querkmachine, line 50: I think this test is misnamed, as it appears to be testing whether `href` is `#` when a value is falsy or undefined, not the link text.

The test for the link text is on line 78.

querkmachine (approved):
One note about a test having the wrong name, but otherwise this seems to do what it sets out to do, and it has tests now! Hurray!

packages/govuk-frontend/src/govuk/components/back-link/template.jsdom.test.js
romaricpascal: Whoops, good catch, I'll rename.

Issue #5157, opened 2024-07-19
Spike how we can integrate information from Typedoc in our own docs


romaricpascal:
## What

Spike to discover:
- which informations on our API [Typedoc](https://typedoc.org/) can provide us in a format we can then integrate in our docs (hopefully JSON or similar machine parseable format)
- how we can integrate that information in the frontend docs
- what this automation would mean for authoring/reviewing the content of the API docs

## Why

So we can automatically generate part of our API documentation in our docs

## Who needs to work on this

Developers, Tech Writers

## Who needs to review this

Developers, Tech Writers

## Done when

We've identified:
- [x] Information Typedoc gives us access to
- [x] How we could integrate it in the frontend docs
- [ ] The impact on authoring/reviewing the API docs

romaricpascal:
We currently integrate our Sass API doc automatically, so there may be interesting bits to take away from there (thinking of the authoring/reviewing).

owenatgov:
I looked at this very briefly today. The hope is that we can export our js API via typedoc the same way we do with sassdoc.

## How our sass API is generated via sassdoc

This is mainly driven by the `build:sassdoc` and `build:sassdocv4` commands which are both run by the `postinstall` command. [These can all be found in the package.json](https://github.com/alphagov/govuk-frontend-docs/blob/main/package.json#L6-L8). The build commands are just running sassdoc but specifying the entry point as within the govuk-frontend package and the output as a json file.
[
In the index for the Sass API reference page](https://github.com/alphagov/govuk-frontend-docs/blob/main/source/sass-api-reference/index.html.md.erb) it interprets this json data, with the help of [the sassdocs_helpers module](https://github.com/alphagov/govuk-frontend-docs/blob/main/lib/sassdocs_helpers.rb). At this stage it's data interpolation and feels very possible.

## How we might generate our JS API via typedoc

Typedoc has similar functions for [specifying an entry point](https://typedoc.org/options/input/#entrypoints) and [an output location as json](https://typedoc.org/options/output/#out). It looks like it might be a bit more involved though. The cli won't run without a valid config and we don't distribute our one in govuk-frontend. We can either replicate it in the frontend docs or try exporting it. There may be a need to redo some of our paths so the former option might be better, if a bit of a drag.

owenatgov:
# Update
[...]

PR #5239, opened 2024-08-16
Consider JSDoc `@import` for non-exported types

colinrotherham:
Hello 👋

Thought I'd share a little improvement added in [TypeScript 5.5: JSDoc `@import` tag](https://devblogs.microsoft.com/typescript/announcing-typescript-5-5/#the-jsdoc-import-tag)

Should prevent `@typedef` re-exporting private/internal definitions in https://github.com/alphagov/govuk-frontend/issues/5157 by using `@import` instead

### DIFF
--- a/.github/workflows/scripts/comments.mjs
+++ b/.github/workflows/scripts/comments.mjs
@@ -333,7 +333,8 @@ function getReviewAppUrl(prNumber, path = '/') {
  */
 
 /**
- * @typedef {import('@octokit/plugin-rest-endpoint-methods').RestEndpointMethodTypes["issues"]} IssuesEndpoint
+ * @import {RestEndpointMethodTypes} from '@octokit/plugin-rest-endpoint-methods'
+ * @typedef {RestEndpointMethodTypes["issues"]} IssuesEndpoint
  * @typedef {IssuesEndpoint["listComments"]["parameters"]} IssueCommentsListParams
  * @typedef {IssuesEndpoint["getComment"]["response"]["data"]} IssueCommentData
  */
--- a/package-lock.json
+++ b/package-lock.json
@@ -37,7 +37,7 @@
         "eslint-plugin-import": "^2.29.1",
         "eslint-plugin-jest": "^27.9.0",
         "eslint-plugin-jest-dom": "^5.4.0",
-        "eslint-plugin-jsdoc": "^48.2.2",
+        "eslint-plugin-jsdoc": "^50.6.3",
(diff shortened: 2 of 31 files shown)


romaricpascal:
Oh, that's nice, thanks for pointing that out and updating the existing `@typedef` 🙌🏻 . Is the PR in draft because you're expecting to make further changes?

colinrotherham:
@romaricpascal No problem 😊

Nah only in "draft" because I've only rolled out `@import` within source code, it's good to go if you're happy. I was being nosey and hoped it might ease up the [missing exports in `75e6e14`](https://github.com/alphagov/govuk-frontend/commit/75e6e14b641ae86108501987fd2eabe658a8c4d8#diff-3a5267728719bcf0c66ff6965f911852b7ba3776350fde6acaa7c9af41450127)

I'd recommend widely using `@import` as it preserves all the generics too, so no need for `@template`

Just pushed an extra commit to demonstrate

colinrotherham:
>Happy to take over updating the rest of the `import()` in types, as well. Would it be fine if they were pushed on your branch so we merged them all in one go?

Yeah go for it ✅

>Just to confirm, is it [this part of the commit](https://github.com/alphagov/govuk-frontend/pull/5239/commits/4f599cdd9bad67c01937095f66b4837b778093d9#diff-39a7ecb7b97bdab58870b05feecedac4450eb52978881aeec18b12a56f7675a9L412-L415) that statement refers to, where you were able to drop the `@typedef` for HandlerFunction and the `@template` it required to keep it being a generics because we were now able to directly use `EvaluateFuncWith` via `@import`?

That's the one

Previously all generics needed to be re-declared via `@template`, now they just work

colinrotherham:
Think you're all set

IDEs and editors will discover [`@types/govuk-frontend`](https://www.npmjs.com/package/@types/govuk-frontend) and this repo uses TypeScript 5.5+ already

Thanks @romaricpascal

domoscargin:
> Made another push to tweak that last commit ([diff](https://github.com/alphagov/govuk-frontend/compare/e620b8b1a3e1f9878bdab512cc88e99d3ff5a6ea..3494a621d150c3c45fe94f394a7ab97f46aa3209))
> 
> Had forgotten `TaskOptions` (alias of `AssetPathOptions`) needed to stay as `@typedef`
> 
> Since there's some `@import` added in #5607 I'll rebase this after it merges

Thanks @colinrotherham - we were just about to take a look at that failure, so thanks for jumping on it!

domoscargin (approved):
Nice!

PR #5240, opened 2024-08-17
Add heading level option to label component called `headingLevel`

aliuk2012:
This optional setting only works if the label is a page heading i.e `isPageHeading` is set to true.

A HMPPS service had an external accessibility audit raised the following point on a few of it pages:

> The form labels are styled to look like headings and stand out visually. While sighted users can identify headings based on larger font size and bold formatting, screen reader users rely on proper HTML markup to understand the structure of a page.

> Screen reader users sometimes jump between the headings of the page to gain a better understanding of the structure and content on the page. The [WebAIM screen reader survey](https://webaim.org/projects/) has found that [headings are the number one means by which screen reader users get a sense of a complex page](https://webaim.org/projects/screenreadersurvey9/#finding). When headings are not available programmatically to users, it affects the time and effort needed to understand how the content is presented on the page. Furthermore, this creates a discrepancy between information available to a sighted individual and users relying on assistive technology.

They suggested something like
```
<h2>
    <label class="govuk-label govuk-label--m" for="myInputField">
        This is a h2 label
     </label>
</h2>
```

The problem is that the existing label component:

1. It does support custom html, but that would place the headings inside the label means Inline elements would contain block level elements which is not valid
2. Using isPageHeading, uses a hard coded h1 so cannot use that options

These changes are backwards compatible by setting the heading level to 1.

### DIFF
--- a/packages/govuk-frontend/src/govuk/components/label/label.yaml
+++ b/packages/govuk-frontend/src/govuk/components/label/label.yaml
@@ -72,6 +72,11 @@ examples:
     options:
       text: National Insurance number
       isPageHeading: true
+  - name: as page heading with defined heading level
+    options:
+      text: National Insurance number
+      isPageHeading: true
+      headingLevel: 2
 
   # Hidden examples are not shown in the review app, but are used for tests and HTML fixtures
   - name: empty
--- a/packages/govuk-frontend/src/govuk/components/label/template.njk
+++ b/packages/govuk-frontend/src/govuk/components/label/template.njk
@@ -10,9 +10,12 @@
 {% endset -%}
 
 {% if params.isPageHeading %}
-<h1 class="govuk-label-wrapper">
+{#- Optional. The page heading level(1-6) to be used with ``isPageHeading` defaults to heading level 1-#}
+{% set headingLevel = params.headingLevel if params.headingLevel else 1 %}
+
+<h{{ headingLevel }} class="govuk-label-wrapper">
   {{ labelHtml | safe | trim | indent(2) }}
-</h1>
+</h{{ headingLevel }}>
 {% else %}
 {{ labelHtml | safe | trim }}
 {% endif %}
--- a/packages/govuk-frontend/src/govuk/components/label/template.test.js
+++ b/packages/govuk-frontend/src/govuk/components/label/template.test.js
@@ -68,6 +68,13 @@ describe('Label', () => {
       expect($selector.length).toBeTruthy()
     })
 
+    it('can be nested inside an heading tag using isPageHeading and headingLevel', () => {
+      const $ = render('label', examples['as page heading with defined level'])
+
+      const $selector = $('h2 > .govuk-label')
+      expect($selector.length).toBeTruthy()
+    })
+
     it('allows additional attributes to be added to the component', () => {
       const $ = render('label', examples.attributes)
 


romaricpascal:
Hi Alistair! 

Thanks for the suggestion, this feels useful, but before moving further with it, we'll likely want some more input from our accessibility specialist and research to better understand when and when not form labels need wrapping in headings so we could issue the appropriate guidance with the option. 

Releasing the option would also have an effect on our [Fieldset component](https://design-system.service.gov.uk/components/fieldset/) which has a similar `isPageHeading` option for its legend as the Label.

In terms of API, it may be preferable to have it replace the `isPageHeading` option altogether, which would let people set a single option to control the heading level (`isPageHeading` being an equivalent to `headingLevel=1` while being [deprecated until its removal](https://github.com/alphagov/govuk-frontend/blob/main/docs/contributing/managing-change.md#deprecating-and-removing-features)).

As an aside for when we look further into this, I've tracked the introduction of the `isPageHeading` option to [this pull request](https://github.com/alphagov/govuk-frontend/pull/684) which may contain extra context.

aliuk2012:
Hi @romaricpascal 

Thanks for your replying. A very good point about fieldset component would need to behave in exactly the same way. Do you want me to make those changes as part of this PR?

I agree, replacing isPageHeading might be a better option but I was trying to suggest a smaller change that wouldn't break existing functionality and mean that the change would have to wait to be part of a major version bump.

It terms of this change, I'm not 100% clear on what is required of me to move this forward.  Happy to help out and contribute as much as I can.

colinrotherham:
Did you get any further with this @aliuk2012 or @romaricpascal?

The GOV.UK Component Guide includes [form input `heading_level` and `heading_size`](https://components.publishing.service.gov.uk/component-guide/input/with_label_as_heading)

```erb
<%= render "govuk_publishing_components/components/input", {
  label: {
    text: "This is a heading 1 and a label"
  },
  name: "name",
  heading_level: 1,
  heading_size: "l"
} %>
```

They do the same for legends in [radios](https://components.publishing.service.gov.uk/component-guide/radio/with_custom_heading_size) and [checkboxes](https://components.publishing.service.gov.uk/component-guide/checkboxes/with_custom_heading_size)

It would certainly solve the mixup between `govuk-fieldset__legend--l` and `govuk-label--l`

>This allows the size of the legend to be changed. Valid options are `s`, `m`, `l`, `xl`, defaulting to `m` if no option is passed.
>
>If the `is_page_heading` option is `true` and `heading_size` is not set, the text size will be `xl`.

We're hoping to do something similar on NHS.UK frontend but consensus that this is a good thing would be great—especially for staff-facing services that don't fit the 'one question per page' model

It was raised again today where legends that look like headings weren't included in the VoiceOver headings rotor 😔

colinrotherham:
Copying over from a [Slack thread](https://ukgovernmentdigital.slack.com/archives/C6DMEH5R6/p1762258993240699) for context:

>Regarding the audit failure and `headingLevel`, I’m reading it as:
>
>* Multiple “medium” size `<h2>` elements are on the page
>* Some “medium” size `<label>` elements are on the page
>* All “medium” size headings and labels are styled visually the same—but the labels are not headings
>
>i.e. We have a consistency issue: Sighted users can skim through all the headings (including labels that look like headings) but screen reader users cannot
>
>Yet those “medium” size headings could have been `<h3>` or `<h4>` elements and the issue still stands

---

For extra context, some staff facing services can have multiple `<form>` elements on a page:

* Main content (e.g. comment box) `<form>` with `<h1>`
* Aside content (e.g. search box) `<form>` with `<h2>`

The current macros assume one-thing-per-page and prevent this

Adding a `headingLevel` option would be incredibly helpful

selfthinker:
We have an issue about this (#6471) and will investigate further. We need further evidence that this is a need that is impacting a lot of users. And we need to check if this is something that can or should be fixed by code or design. If this is going to be changed, we would also need to have additional guidance on how to use it. When that happens, we will implement our own version, so we're going to close this.

Issue #5210, opened 2024-08-08
Make `isSupported()` part of our public API


romaricpascal:
## What

Make [`isSupported()`](https://github.com/alphagov/govuk-frontend/issues/5205) public and export it in `all.mjs`

## Why

So services can check whether GOV.UK Frontend is supported the same way our code does, reducing duplication.

## Who needs to work on this

Developers, Content writers

## Who needs to review this

Developers, Content writers

## Done when

- [x] `isSupported()` is made available in `all.mjs`
- [x] `isSupported()` is documented as public in the JSDoc
- [ ] The feature is documented in the Frontend Docs

romaricpascal:
Added an item to the Docs tasklist of the epic (#5205) so we can close this one.

PR #5250, opened 2024-08-19
Add `isSupported` to `all.mjs`

patrickpatrickpatrick:
## What

- adds `isSupported` as an export for import from `govuk-frontend`
- removes `internal` from JSDoc to mark function as public
- adds `exports isSupported function` test
- updates `exports Components` test

## Why

`isSupported` is a method that will be very useful for developers creating their own components that dependent on the `govuk-frontend`. `isSupported` returns a boolean to indicate if the browser is supported by `govuk-frontend` by checking for presence of the `govuk-frontend-supported` on the body of the page.

Fixes https://github.com/alphagov/govuk-frontend/issues/5210

### DIFF
--- a/packages/govuk-frontend/src/govuk/common/index.mjs
+++ b/packages/govuk-frontend/src/govuk/common/index.mjs
@@ -194,8 +194,7 @@ export function setFocus($element, options = {}) {
  * Some browsers will load and run our JavaScript but GOV.UK Frontend
  * won't be supported.
  *
- * @internal
- * @param {HTMLElement | null} [$scope] - HTML element `<body>` checked for browser support
+ * @param {HTMLElement | null} [$scope] - (internal) `<body>` HTML element checked for browser support
  * @returns {boolean} Whether GOV.UK Frontend is supported on this page
  */
 export function isSupported($scope = document.body) {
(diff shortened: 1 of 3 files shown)


packages/govuk-frontend/src/govuk/common/index.mjs
```diff
@@ -194,7 +194,6 @@ export function setFocus($element, options = {}) {
  * Some browsers will load and run our JavaScript but GOV.UK Frontend
  * won't be supported.
  *
- * @internal
  * @param {HTMLElement | null} [$scope] - HTML element `<body>` checked for browser support
```
romaricpascal, line 197:
```suggestion
 * @param {HTMLElement | null} [$scope] - (internal) `<body>` HTML element checked for browser support
```

That parameter is only really used for testing, especially as [the snippet in our docs](https://frontend.design-system.service.gov.uk/import-javascript/#before-you-start) adds the class to the `<body>` element in any case. Best to not encourage people to play with it, especially as you might misinterpret that you can pass any element and know if GOV.UK Frontend is supported for it (which is not what would happen).

Just documenting it is lighter than wrapping the function before export in `all.mjs` 😊

romaricpascal (approved):
🙌🏻

PR #5251, opened 2024-08-19
Add `isSupported` to `createAll`

### LINKED ISSUE #5213: Add leading check to verify that GOV.UK Frontend is supported in `createAll`
## What

Similarly to what `initAll` does, add a leading check to `createAll` to verify that GOV.UK Frontend is supported before looping on components. 

## Why

To avoid unnecessary looping

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [x] `createAll` bails early if GOV.UK Frontend is not supported

patrickpatrickpatrick:
## What

- `createAll` returns empty array and throws error is `isSupported` false
- update existing `createAll` tests and add test for `isSupported` false

## Why

Prevents `createAll` running if `govuk-frontend` not supported.

Fixes #5213

### DIFF
--- a/packages/govuk-frontend/src/govuk/init.jsdom.test.mjs
+++ b/packages/govuk-frontend/src/govuk/init.jsdom.test.mjs
@@ -146,8 +146,12 @@ describe('initAll', () => {
 })
 
 describe('createAll', () => {
+  beforeEach(() => {
+    document.body.classList.add('govuk-frontend-supported')
+  })
+
   afterEach(() => {
-    document.body.innerHTML = ''
+    document.body.outerHTML = '<body></body>'
   })
 
   class MockComponent {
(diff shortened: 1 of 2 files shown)


romaricpascal:
@patrickpatrickpatrick I've change the base of the PR to not point to main, so we can work on the JS API without disturbing the Navigation component release.

packages/govuk-frontend/src/govuk/init.jsdom.test.mjs
```diff
@@ -146,6 +146,10 @@ describe('initAll', () => {
 })
 
 describe('createAll', () => {
+  beforeEach(() => {
+    document.body.classList.add('govuk-frontend-supported')
+  })
+
   afterEach(() => {
     document.body.innerHTML = ''
```
romaricpascal, line 154: **suggestion** [As JSDOM doesn't get cleaned between the same tests of a given file](https://github.com/jestjs/jest/issues/1224#issuecomment-444586798) we should ensure the class gets removed as well.

```suggestion
    document.body.outerHTML = '<body></body>'
```

romaricpascal (approved):
Looks all neat, I'd just suggest a little more safety in the tests cleanup now we add a class before each of the `createAll` tests.

PR #5252, opened 2024-08-21
Add `errorCallback` to `createAll`

### LINKED ISSUE #5212: Add callback to respond to errors when a component does not initialise in `createAll`
## What

Add an option to `createAll` that allows services to pass a callback function to `createAll` that will be called if a component fails to initialise (or GOV.UK Frontend is not supported).

## Why

So services can react to errors happening during component initialisation with `createAll`, for example sending a report to an error monitoring service

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [x] We've decided on the arguments and name of that function
- [x] We've implemented the function
- [x] The feature is documented

patrickpatrickpatrick:
## What

- Add support for `onError` callback in `createAll` which is called if error occurs on component initialisation.
- New parameter for `createAll`, `createAllOptions` which allows user to specify a `scope`, `onError` or an object that contains both.
- `isSupported` will execute `onError` callback if specified
- New tests added for `onError` callback and `createAllOptions`.

## Why

Provide better support for defining custom components by allowing users to execute custom callback if the components fail to initialise. 

Fixes #5212

### DIFF
--- a/packages/govuk-frontend/src/govuk/init.mjs
+++ b/packages/govuk-frontend/src/govuk/init.mjs
@@ -65,18 +65,46 @@ function initAll(config) {
  *
  * @template {CompatibleClass} T
  * @param {T} Component - class of the component to create
- * @param {T["defaults"]} [config] - config for the component
- * @param {Element|Document} [$scope] - scope of the document to search within
+ * @param {T["defaults"]} [config] - Config supplied to component
+ * @param {OnErrorCallback<T> | Element | Document | CreateAllOptions<T> } [createAllOptions] - options for createAll including scope of the document to search within and callback function if error throw by component on init
  * @returns {Array<InstanceType<T>>} - array of instantiated components
  */
-function createAll(Component, config, $scope = document) {
+function createAll(Component, config, createAllOptions) {
+  let /** @type {Element | Document} */ $scope = document
+  let /** @type {OnErrorCallback<Component> | undefined} */ onError
+
+  if (typeof createAllOptions === 'object') {
+    createAllOptions = /** @type {CreateAllOptions<Component>} */ (
+      // eslint-disable-next-line no-self-assign
+      createAllOptions
+    )
+
+    $scope = createAllOptions.scope ?? $scope
+    onError = createAllOptions.onError
+  }
+
+  if (typeof createAllOptions === 'function') {
+    onError = createAllOptions
+  }
+
+  if (createAllOptions instanceof HTMLElement) {
+    $scope = createAllOptions
+  }
+
   const $elements = $scope.querySelectorAll(
     `[data-module="${Component.moduleName}"]`
   )
 
   // Skip initialisation when GOV.UK Frontend is not supported
   if (!isSupported()) {
-    console.log(new SupportError())
+    if (onError) {
+      onError(new SupportError(), {
+        component: Component,
+        config
+      })
+    } else {
+      console.log(new SupportError())
+    }
     return []
   }
 
@@ -96,7 +124,16 @@ function createAll(Component, config, $scope = document) {
           ? new Component($element, config)
           : new Component($element)
       } catch (error) {
-        console.log(error)
+        if (onError && error instanceof Error) {
+          onError(error, {
+            element: $element,
+            component: Component,
+            config
+          })
+        } else {
+          console.log(error)
+        }
+
         return null
       }
     })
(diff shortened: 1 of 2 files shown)


packages/govuk-frontend/src/govuk/init.mjs
```diff
@@ -67,9 +74,10 @@ function initAll(config) {
  * @param {T} Component - class of the component to create
  * @param {T["defaults"]} [config] - config for the component
  * @param {Element|Document} [$scope] - scope of the document to search within
+ * @param {errorCallback<T>} [errorCallback] - scope of the document to search within
  * @returns {Array<InstanceType<T>>} - array of instantiated components
  */
-function createAll(Component, config, $scope = document) {
+function createAll(Component, config, $scope = document, errorCallback) {
```
romaricpascal, line 80: **suggestion** As discussed during our call, let's try to limit the number of arguments to 3 and regroup the scope and error callback in a param that can be either:
- an `Element|Document` representing the scope
- a `Function` representing the error callback
- an `{scope: Element|Document, errorCallback: Function}` if both are needed.

packages/govuk-frontend/src/govuk/init.mjs
```diff
@@ -65,11 +79,30 @@ function initAll(config) {
...
+function createAll(Component, config, createAllOptions) {
+  let /** @type {Element | Document} */ $scope = document
+  let /** @type {onErrorCallback<Component> | undefined} */ onError
+
+  if (typeof createAllOptions === 'object') {
+    $scope =
+      /** @type {createAllOptions<Component>} */ (createAllOptions).scope ??
+      $scope
+    onError = /** @type {createAllOptions<Component>} */ (createAllOptions)
+      .onError
```
romaricpascal, line 95: **suggestion** Would it work to do something like the following to put the casting in its own line and make things a little more readable by separating the assignments of the options from the casting?

```suggestion
		createAllOptions = /** @type {createAllOptions<Component>} */ createAllOptions
		
    $scope = createAllOptions.scope ?? $scope
    onError = createAllOptions.onError
```

packages/govuk-frontend/src/govuk/init.mjs
```diff
@@ -90,7 +123,16 @@ function createAll(Component, config, $scope = document) {
           ? new Component($element, config)
           : new Component($element)
       } catch (error) {
-        console.log(error)
+        if (onError) {
+          onError(/** @type {Error} */ (error), {
```
romaricpascal, line 127: **suggestion** We've usually ran `typeof`/`instanceof` calls rather than casting to make typescript aware of types. Would the following work for that?

```suggestion
        if (onError && error instanceof Error) {
          onError(error, {
```

packages/govuk-frontend/src/govuk/init.mjs
```diff
@@ -54,6 +54,20 @@ function initAll(config) {
...
+ * @param {Error} error - Thrown error
+ * @param {ErrorContext<T>} context - Object containing the element, component class and configuration
+ */
+
+/**
+ * @template {CompatibleClass} T
+ * @typedef {object} createAllOptions
+ * @property {Element | Document} [scope] - scope of the document to search within
+ * @property {onErrorCallback<T>} [onError] - callback function if error throw by component on init
+ */
```
romaricpascal, line 69: **question** We've tended to put our types at the bottom of the files across the project, is there something forcing these to be early on? If not, would be good to align with the rest of the files for consistency.

**suggestion** We've generally been PascalCasing the types rather than camelCasing them.

romaricpascal (commented):
Looking neat! 🙌🏻 Just little remarks about undocumented conventions in the rest of the files for how we do type annotations for TypeScript (which we should probably start documenting in the docs).

romaricpascal:
@patrickpatrickpatrick I've changed the PR base to `public-js-api` on GitHub so we don't merge on `main` accidentally, but I think it'll need a little rebase to get the check for `isSupported` inside `createAll`. 

What are your thoughts on whether the check should also trigger a call to `onError`, as well?

packages/govuk-frontend/src/govuk/init.mjs
patrickpatrickpatrick: That makes sense, I have disable eslint rule "no-self-assign" but it does look neater

patrickpatrickpatrick:
> @patrickpatrickpatrick I've changed the PR base to `public-js-api` on GitHub so we don't merge on `main` accidentally, but I think it'll need a little rebase to get the check for `isSupported` inside `createAll`.
> 
> What are your thoughts on whether the check should also trigger a call to `onError`, as well? For consistency, it feels like we should trigger a call as well, and if people initialise multiple components with multiple `createAll` calls likely receiving the same error handler, they can use `isSupported` to not call `createAll` and avoid the noise 🤔

That makes sense, have added this!

romaricpascal (approved):
Looks good to go! Thanks for making the updates 🙌🏻

